PR #474, opened 2026-05-18
[Benchmark]: Clarify SeedTTS speed metric semantics

### LINKED ISSUE #470: [CI / benchmarks] Clarify SeedTTS speed metric semantics and tighten RTF thresholds with hard semantics
### 
**Background**

The router CI (`tests/test_model/test_omni_router_ci.py`) and the single-worker TTS CI (`tests/test_model/test_qwen3_omni_tts_ci.py`) both use `benchmarks/metrics/performance.py::compute_speed_metrics`, but names and docs make cross-run comparisons easy to misread, which leads to wrong conclusions about “DP / router regressions.”

**Ground truth (current implementation)**

1. **`throughput_qps`**: When `wall_clock_s` is passed in, this is `completed_requests / wall_clock_s` — end-to-end throughput over the whole benchmark window.

2. **`tok_per_s_agg`**: **Not** `total_tokens / wall_clock`. It is implemented as `sum(completion_tokens) / sum(engine_time_s)` (see `_compute_token_metrics`). For Omni SeedTTS, `engine_time_s` is a per-request end-to-end proxy; its denominator is **not** the same notion as “how saturated wall-clock time was.”

3. **`tok_per_s_mean`**: Arithmetic mean of per-request `completion_tokens / engine_time_s`.

So when concurrency or topology changes, **`throughput_qps` and `tok_per_s_agg` need not move together**, nor satisfy a naive “DP≈2 ⇒ both ~2×” story. Explaining changes in `tok_per_s_agg` via “steady-state / wall-clock ramp” **does not match this codebase**.

**Problems**

1. **Naming / docs gap**: `tok_per_s_agg` reads like “cluster-wide aggregate tok/s,” but the formula is “total tokens / sum of per-request times,” which is easy to confuse with **`throughput_qps`** (wall-clock). That hurts triage and threshold reviews.

2. **RTF slack too loose**: `rtf_mean` has product-hard semantics (RTF ≥ 1.0 ⇒ not real-time). Router CI uses `apply_slack(..., slack_lower=1.50)` shared with latency-style caps, so CI may still pass when the system is already far outside real-time operation.

3. **Asymmetric slack between CIs**: Single-worker uses defaults `slack_higher=0.875`, `slack_lower=1.125`; router overrides `0.75 / 1.50`. Document whether this is intentional (e.g. colocated jitter) or legacy, so it is not read as “router path may regress much more than single-worker” without explanation.

**Proposed work (can be split across PRs)**

1. Add a short glossary in `benchmarks/metrics/performance.py` (or `benchmarks/README.md`) defining:
   - `throughput_qps` (wall-clock),
   - `tok_per_s_mean` vs `tok_per_s_agg` (denominators: per-request time vs **sum** of per-request times),
   - Explicit note: **not** the same as some “total tokens / wall_clock” conventions elsewhere in the SGLang ecosystem unless we add a separate field.

2. *(Optional)* Add **`tok_per_s_wall = sum(completion_tokens) / wall_clock_s`** to the summary (print and/or assert) if we want a metric aligned with “cluster token throughput over wall time,” instead of overloading `_agg`.

3. Add an **`rtf_mean` hard cap** independent of `apply_slack` for both router CI and single-worker CI (e.g. ≤ 1.0 or ≤ 0.95), analogous to `ROUTER_SEEDTTS_LATENCY_P95_MAX`.

4. One-line comment (in `tests/utils.py` or the router CI module) on **why** router slack differs from single-worker; revisit tightening slack after stabilizing measurement (e.g. sample count / hardware).

Ratish1:
## Motivation

Fixes #470.

SeedTTS speed summaries used `tok_per_s_agg` for a value that was not wall-clock system throughput. It was computed as generated tokens divided by summed per-request engine/request time, which is useful as a request-time diagnostic but misleading when named like aggregate throughput.

This PR separates the benchmark token metrics by denominator and adds a hard SeedTTS RTF cap so router slack cannot permit slower-than-real-time audio generation.

## Modifications

- Replaces summary `tok_per_s_agg` with `output_tok_per_req_s`, defined as `sum(completion_tokens) / sum(engine_time_s)`. This keeps the old diagnostic while making the accumulated request-time denominator explicit.
- Adds `output_throughput`, defined as `sum(completion_tokens) / wall_clock_s`, for wall-clock output-token throughput aligned with SGLang/vLLM benchmark terminology.
- Removes summary `tok_per_s_mean` and `tok_per_s_median`; these were unweighted per-request ratio summaries and were not the system throughput metric the old names suggested.
- Renames workload-shape counters from `gen_tokens_mean` / `gen_tokens_total` to `output_tokens_mean` / `output_tokens_total`.
- Renames persisted per-request `tok_per_s` fields to `output_token_rate`, while leaving the internal `RequestResult.tok_per_s` carrier unchanged.
- Updates CI threshold references and enforcement to use `output_tok_per_req_s_min`.
- Clamps Qwen3-Omni SeedTTS `rtf_mean_max` through the existing threshold path with `QWEN3_OMNI_SEEDTTS_RTF_MEAN_MAX = 0.95`. The single-worker threshold remains stricter from its reference value; the router threshold is clamped below real-time even though router latency/throughput slack stays wider.
- Updates benchmark reference tables and metric semantics text so the reported names match the formulas.

## Metric Semantics

- `throughput_qps = completed_requests / wall_clock_s`
- `output_throughput = output_tokens_total / wall_clock_s`
- `output_tok_per_req_s = output_tokens_total / sum(engine_time_s)`
- `output_token_rate = completion_tokens / engine_time_s` per request
- `rtf_mean = mean(request_elapsed_s / generated_output_audio_duration_s)`

For SeedTTS RTF, the generated output audio duration is the denominator. Reference audio is input conditioning and is not counted as generated audio.


tests/test_model/test_s2pro_tts_ci.py
zhaochenyang20: This block now uses the `output_tok_per_req_s` threshold key, but the adjacent comment still says `Higher-is-better metrics (throughput, tok/s)`. The main purpose of this PR is to remove the ambiguity where `tok/s` reads like wall-clock aggregate throughput; leaving that broad wording at the threshold-construction point invites the next review or CI triage to mix request-time token rate with wall-clock output throughput again. This creates review risk rather than a runtime failure: maintainers can keep using the old mental model when interpreting thresholds, which weakens the metric semantics contract that this PR is trying to establish.

Update the comment to match `tests/utils.py::apply_slack`, for example `throughput, output token/request-s`, and add a short note if needed that `output_tok_per_req_s` uses the summed per-request engine/request-time denominator. After rebasing, check the S2-Pro threshold table from current main as well, rather than only editing the stale branch version.

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

tests/test_model/test_s2pro_tts_ci.py
Ratish1: done

benchmarks/metrics/mmsu.py
```diff
@@ -99,7 +99,9 @@ def print_mmsu_summary(
         if speed_metrics.get("rtf_mean") is not None:
             print(f"  RTF mean:         {speed_metrics.get('rtf_mean', 0):.4f}")
         print(f"  Throughput:       {speed_metrics.get('throughput_qps', 0):.2f} req/s")
-        print(f"  Tok/s agg:        {speed_metrics.get('tok_per_s_agg', 0):.2f}")
+        print(
+            "  Output tok/req-s: " f"{speed_metrics.get('output_tok_per_req_s', 0):.2f}"
+        )
```
zhaochenyang20: `compute_speed_metrics` now emits `output_throughput = output_tokens_total / wall_clock_s`, and the generic `print_speed_summary` prints it; MMSU uses the custom `print_mmsu_summary`, which currently prints only `throughput_qps` and `output_tok_per_req_s`. As a result, the MMSU CLI output still does not surface the newly added wall-clock output-token throughput, so maintainers who read only the CLI summary see only the request-time token diagnostic and request QPS.

zhaochenyang20 (changes requested):

benchmarks/metrics/mmsu.py
Ratish1: done

zhaochenyang20 (approved):

PR #475, opened 2026-05-18
[Qwen3-Omni] talker partial-prefix downstream startup

### LINKED ISSUE #473: [Perf][Qwen3-Omni] Enable partial-stream downstream startup before upstream stream completion
## Motivation

`sglang-omni` already models streaming edges and deferred request construction, but there is still no first-class path for a downstream stage to **start from a partial upstream stream** once it has received enough information to build a valid request. In practice, that means a downstream stage may still wait for the upstream `stream_done` barrier even when useful work could already begin.

Qwen3-Omni talker is the first concrete place where this missing framework capability is visible. Today, the talker stage does not start until the thinker stream has fully completed. Even while the thinker is continuously producing streaming chunks containing token ids and hidden states, the talker still waits for the final chunk plus the `stream_done` signal before it builds a request and runs its initial prefill.

This is not merely a slow function. It is a **pipeline phase barrier**: the tail of thinker generation and the head of downstream execution do not overlap at all, so audio TTFT is bounded by the full thinker generation time.

The interesting part is that most of the incremental machinery needed for this framework pattern already exists on `main`:

- partial-prefix prompt construction is already modeled by `build_prompt_prefill(..., thinker_done=...)`;
- `build_prefill_input(..., include_assistant_eos=False)` already drops the trailing EOS row for an unfinished assistant segment;
- `append_text_chunk()` already appends later thinker chunks into `pending_text_queue`;
- `mark_thinker_done()` already flips the completion flag and appends the TTS EOS row.

What is still missing is the entry path: let a downstream stage declare that it can build from a **partial but valid** upstream prefix, with Qwen3-Omni talker as the first consumer of that capability, instead of requiring `pending_stream_done=True` unconditionally.

### The current barrier is explicit

`sglang_omni/models/qwen3_omni/talker_scheduler.py:40-47`
```python
def _is_request_build_ready(
    self,
    payload: Any,
    *,
    pending_stream_done: bool,
) -> bool:
    del payload
    return bool(pending_stream_done)
```

`sglang_omni/models/qwen3_omni/request_builders.py:803-810`
```python
thinker_chunks = list(payload.prefetched_chunks)
thinker_done = bool(payload.prefetched_stream_done)

if not thinker_done:
    raise RuntimeError(
        "QwenTalkerScheduler called talker request_builder before thinker "
        "stream completion"
    )
```

The request builder also currently hard-codes the resulting request data as already complete:

`sglang_omni/models/qwen3_omni/request_builders.py:817-845`
```python
prompt_prefill = prefill_builder.build_prompt_prefill(
    payload,
    thinker_chunks,
    thinker_done=True,
)
...
req_data = build_sglang_talker_request(
    ...,
    pending_text_queue=prompt_prefill["pending_text_queue"],
    tts_pad_embed=prompt_prefill["tts_pad_embed"],
    thinker_chunks_done=True,
    ...,
)
```

So there are **three** gates to relax for partial-prefix startup:

1. scheduler readiness;
2. the `request_builder()` hard failure on `not thinker_done`;
3. propagation of `thinker_done` into both `build_prompt_prefill(...)` and `build_sglang_talker_request(...)`.

### The scheduler already supports parking an unready payload

`sglang_omni/scheduling/omni_scheduler.py:425-432`
```python
pending_stream_done = req_id in self._pending_stream_done
payload.prefetched_stream_done = pending_stream_done
if not self._is_request_build_ready(
    payload,
    pending_stream_done=pending_stream_done,
):
    self._deferred_request_payloads[req_id] = payload
    continue
```

So `_is_request_build_ready()` is the policy hook that decides when the payload may leave the deferred queue. It is not the only code that needs to change, but it is the right place to encode the partial-prefix readiness condition.

### The partial-prefix building blocks already exist

- `sglang_omni/models/qwen3_omni/components/talker_prefill.py:179-244`
  - `build_prompt_prefill(payload, thinker_chunks, *, thinker_done)` already accepts unfinished streams explicitly.
- `sglang_omni/models/qwen3_omni/components/talker_input.py:177, 261-267`
  - `build_prefill_input(..., include_assistant_eos=False)` already removes the trailing EOS row when the assistant segment is unfinished.
- `sglang_omni/models/qwen3_omni/components/talker_prefill.py:246-274`
  - `append_text_chunk()` appends later thinker chunks while `thinker_chunks_done=False`.
  - `mark_thinker_done()` flips the flag and appends the final TTS EOS row.
- `sglang_omni/scheduling/omni_scheduler.py:559-571`
  - `_on_stream_chunk()` buffers chunks before the request exists and appends them after it does.
  - `_on_stream_done()` either marks an existing request done or latches completion until the request appears.

One important nuance: `QwenTalkerScheduler` intentionally overrides `_initialize_request_stream_state()` as a no-op today because the talker request builder consumes the prefetched chunks itself. If partial-prefix startup continues to let the request builder consume the chunks available at build time, that no-op can remain unchanged; later chunks still arrive through the existing incremental path.

### The current tests already hint at partial-prefix semantics

`tests/unit_test/qwen3_omni/test_talker.py` already verifies that a short assistant prefix can produce a legal fixed-layout assistant block before a fourth text token exists. That is useful evidence that partial-prefix startup is conceptually aligned with the current implementation, but it does **not** by itself prove the full end-to-end partial-prefix request path.

## Proposal

Introduce a reusable downstream-startup pattern for streaming edges: a stage may remain deferred until it has received a **partial but valid** upstream prefix, then build its request before upstream termination. Qwen3-Omni talker should be the first implementation of this pattern because it already exposes the missing overlap most clearly.

The first landing does not need a new scheduler shape or new stage type. The existing framework already has the right policy hook in `_is_request_build_ready(...)`; what is missing is a concrete readiness contract and a model consumer that uses it before `stream_done`.

For the first consumer, the change should distinguish two related but different thresholds:

1. **Build-ready prefix** — the minimum prefix length needed to construct a semantically valid talker request from a partial thinker stream.
2. **Decode-ready prefix** — the minimum prefix length needed so that, after initial prefill finishes, the talker can immediately enter decode with at least one consumable future text row rather than stalling right away.

Those two thresholds should not be conflated. The semantic minimum is a correctness constraint; the production default is a performance policy that should be chosen from benchmarks.

### Key code changes

| Location | Current behavior | Proposed behavior |
|---|---|---|
| `talker_scheduler.py::_is_request_build_ready` | Only returns true after `pending_stream_done` | Become the first consumer of the partial-stream startup contract: return true when the stream is done **or** when the configured readiness condition is satisfied |
| `request_builders.py::request_builder` | Raises on `not thinker_done` | Allow `thinker_done=False` |
| `request_builders.py::build_prompt_prefill(...)` call | Always passes `thinker_done=True` | Pass the actual `thinker_done` flag |
| `request_builders.py::build_sglang_talker_request(...)` call | Always passes `thinker_chunks_done=True` | Pass the actual `thinker_done` flag |
| `talker_prefill.py` | Already supports `thinker_done` | No API change required |
| `QwenTalkerScheduler._initialize_request_stream_state` | Intentionally no-op | Remains no-op if the request builder continues consuming build-time prefetched chunks |

## Partial-prefix contract

A valid early-start contract must satisfy all of the following:

1. The available assistant prefix is sufficient to construct a valid prompt / assistant segment under the current Qwen3-Omni talker layout.
2. The resulting request preserves multimodal-mask and assistant-boundary semantics.
3. If the goal is to avoid an immediate post-prefill stall, the prefix must also leave at least one consumable future text row after `include_assistant_eos=False` removes the provisional EOS row.
4. If the thinker finishes before the configured threshold is reached, the implementation must fall back to the existing complete-stream path with no extra latency penalty.

The exact production threshold should be decided by benchmark data, not hard-coded in the RFC. A small constant such as 8 or 16 may be a good policy candidate, but the issue should keep the semantic minimum and the tuned default separate.

### Step 1 — Land the first partial-stream downstream-startup consumer

Minimal first PR:

1. Define the partial-stream downstream-startup contract in code comments / tests through the Qwen3 talker implementation: a stage may leave the deferred queue before upstream completion once its stage-specific readiness condition is satisfied.
2. Add a talker config knob for partial-prefix startup. Use an explicit disabled sentinel (for example `None` / `disabled`) so the legacy behavior remains unambiguous; do **not** overload `0` to mean both "disabled" and "ready immediately".
3. Change `_is_request_build_ready()` to return true when:
   - the stream is done, or
   - partial-prefix startup is enabled and the configured readiness condition is satisfied.
4. Remove the hard `not thinker_done` failure in `request_builder()`.
5. Thread the real `thinker_done` value into:
   - `build_prompt_prefill(..., thinker_done=thinker_done)`
   - `build_sglang_talker_request(..., thinker_chunks_done=thinker_done)`
6. Add tests for:
   - disabled mode preserving the old behavior;
   - partial-prefix build before stream completion;
   - short-prefix legality;
   - stream completion after the request already exists;
   - fallback when the stream ends before the configured threshold.

**Acceptance**
- With partial-prefix startup disabled, existing Qwen3-Omni behavior is preserved.
- With partial-prefix startup enabled, thinker / talker overlap is visible in a request timeline.
- Single-request audio TTFT improves by a measurable amount on the chosen speech benchmark.
- WER / audio-consistency checks do not regress.

### Step 2 — Tune and promote a production default

After Step 1 demonstrates correctness and value:

1. Benchmark candidate thresholds across short / medium / long prompts.
2. Pick a non-zero default that improves TTFT without increasing immediate decode stalls.
3. Keep the complete-stream path as a safe fallback for very short responses and boundary cases.

**Acceptance**
- Realtime / SeedTTS-style workloads show clear TTFT improvement at P50 and P95.
- Long prompt, short prompt, near-minimal assistant prefix, EOS boundary, and mid-stream abort cases all have targeted coverage.

### Step 3 — Adaptive thresholding (optional)

Only if Step 2 shows that a fixed threshold is materially suboptimal:

- explore an adaptive policy based on thinker chunk arrival rate / prompt shape;
- keep this out of the initial implementation so the first PR stays small and debuggable.

## Invariants

- `_on_stream_done()` semantics stay unchanged.
- `append_text_chunk()` / `mark_thinker_done()` APIs stay unchanged.
- Talker decode-readiness checks in `_is_batch_ready_to_run()` stay unchanged.
- TP / leader-follower behavior stays unchanged.
- WER / audio parity is a release gate.

## Non-Goals

- No `.cpu().tolist()` cleanup in this issue. That belongs in a separate follow-up on the tensorized talker request path.
- No `pending_text_queue` tensorization in this issue.
- No Code2Wav incremental-buffer changes in this issue.
- No scheduler-shape, inbox/outbox, control-plane, or event-loop refactor in this issue.
- No new stage type, payload type, or control-plane message.

## Relation to existing work

- Orthogonal to #472: #472 is about **how data moves**; this issue is about **when a downstream stage may start moving**.
- Builds on #406: streaming edges into downstream stages already exist; this issue adds the first concrete consumer that can leave the deferred queue before upstream `stream_done`.
- Qwen3-Omni talker is the first target, but the pattern should become reusable for future downstream stages whose request builders can safely start from partial upstream streams.
- Compatible with #385 realtime support; no realtime protocol change is required.

## Open Questions

1. Should the framework eventually expose a reusable readiness interface beyond `_is_request_build_ready(...)` once a second consumer appears, or is the current hook sufficient for now?
2. What are the exact **build-ready** and **decode-ready** prefix lengths under the current talker layout? The code already supports short prefixes, but the thresholds should be stated explicitly once validated.
3. Which tuned default gives the best TTFT / stall trade-off on representative speech workloads? This should be benchmarked rather than guessed in advance.
4. When a talker request has already been built but the thinker is still streaming, does the abort path need an additional fast short-circuit before appending late chunks, or is the existing abort cleanup sufficient?
5. Should CI add a fixed-seed parity test comparing disabled vs enabled partial-prefix startup for a small speech fixture to guard against WER / waveform regressions?

Hayden727:
Closes #473.

Opt-in partial-prefix talker startup. With `partial_start_min_chunks=5`, talker builds its request from a partial thinker stream instead of waiting for `stream_done`. Default `None` preserves legacy behavior.

## Measured impact (dual-GPU H200, 5 repeats)

| Prompt | Baseline TTFT | partial=5 TTFT | Delta |
|---|---|---|---|
| short  | 0.322s | 0.250s | **-22.2%** |
| medium | 0.356s | 0.251s | **-29.5%** |

Same `audio_chunks` count. Bench: `benchmarks/eval/benchmark_omni_streaming_ttft.py`.

## Safety (MMMU audio, n=50 x 3 trials)

- hardware: dual-GPU H200
- accuracy: 0.660 -> 0.660
- `wer_below_50_corpus` (CI metric): 0.1097 -> **0.0998** (-9.0%)
- catastrophic count: 9.33 -> 9.67
- failed requests: 0 / 0
- `audio_duration_mean_s`: 38.608 -> 38.602 (-0.01%)
- `latency_mean_s`: 22.302 -> 22.035 (-1.2%)
- `rtf_mean`: 0.6398 -> 0.6204 (-3.0%)
- `throughput_qps`: 0.6567 -> 0.6770 (+3.1%)
- 670/670 unit tests green

## Open follow-up

No observed regression in this rerun. Keep tracking audio duration and WER distribution before flipping the default away from `None`.


Hayden727:
# AC-10 Validation Report — Qwen3-Omni partial-prefix talker startup

Validated on **ion8-h200-omni** in container `sglang-omni-hayden`
(`frankleeeee/sglang-omni:dev`, GPUs `2,3` mapped to container 0,1, branch
`feat/talker-partial-prefix-startup` @ `aa78fb3`).

## Unit suite

`pytest tests/unit_test/qwen3_omni/test_talker.py -v`

**28/28 pass**, including the partial-start regression coverage:

```
test_partial_prompt_prefill_layout_invariants PASSED  (MIN_PARTIAL_START_CHUNKS=3 pin)
test_partial_disabled_preserves_legacy_path PASSED
test_partial_enabled_below_threshold_stays_deferred PASSED
test_partial_enabled_at_threshold_returns_true_with_done_false PASSED
test_partial_floors_to_min_partial_start_chunks PASSED
test_partial_count_excludes_im_end_chunks PASSED            # Codex round-2 fix
test_partial_non_positive_threshold_disables_partial_start PASSED
test_partial_enabled_zero_chunks_stays_deferred PASSED
test_no_op_initialize_request_stream_state_prevents_replay PASSED
test_stream_done_after_partial_build_marks_thinker_done PASSED
test_chunk_after_partial_build_appends_once PASSED
test_request_builder_threads_thinker_done_false_on_partial_path PASSED
test_request_builder_threads_thinker_done_true_on_completed_stream PASSED
test_partial_path_rejects_zero_chunks_without_done PASSED
test_fallback_chunks_only_on_done_path PASSED
test_done_path_with_no_fallback_raises_value_error PASSED
test_process_input_requests_partial_build_state_machine PASSED
test_process_input_requests_keeps_deferred_when_below_threshold PASSED
test_abort_filters_subsequent_stream_messages_via_recv_requests PASSED
test_wiring_propagation_factory_args_to_scheduler PASSED
```

## End-to-end server measurement

`python -m sglang_omni.cli serve --model-path Qwen/Qwen3-Omni-30B-A3B-Instruct --port 8000`
on two GPUs, talker_ar context 32768, mem_fraction_static=0.907.

### Audio TTFT (warmup 2 × repeats 5)

| prompt | disabled mean | disabled stdev | enabled-min5 mean | enabled-min5 stdev |
|---|---|---|---|---|
| short  (8 thinker chunks)  | 0.334s | 0.031s | 0.313s | 0.002s |
| medium (30 thinker chunks) | 0.427s | 0.001s | 0.427s | 0.003s |

**The enabled-vs-disabled delta is within run-to-run noise**: the partial-start
mechanism does not measurably move TTFT on these prompts.

## Why partial-start did not fire (and why the implementation is still correct)

I added a per-build observability log line and ran the workload at INFO level.
**Every single build in 23 measured requests (single + concurrent) reported
`thinker_done=True`**. Partial-start never produced a build before `stream_done`
arrived. The grep summary:

```
$ grep -cE "talker_request_build" server-enabled.log     # 23
$ grep -cE "thinker_done=False" server-enabled.log       # 0
$ grep -cE "thinker_done=True"  server-enabled.log       # 23
```

The mechanism's window collapses on this hardware/workload because the thinker
emits all `stream_chunk` messages and the terminal `stream_done` faster than the
talker scheduler can take a tick *between* them. Concretely:

- Thinker generates ~7.5 ms / token on H200 for Qwen3-Omni (MoE, 3 B active).
- All chunks + `stream_done` arrive in the talker's inbox within one
  `event_loop_normal` iteration of the talker scheduler.
- `recv_requests()` drains the entire inbox in step 1 of that iteration,
  populating both `_pending_stream_chunks` and `_pending_stream_done`.
- `process_input_requests([deferred_payload])` (step 2 of the same iteration)
  reads `pending_stream_done=True`, takes the legacy completed-stream branch,
  and calls the builder with `thinker_done=True`.

A concurrent test (`concurrency=4`) reproduced the same outcome (all 4 of 4
builds had `thinker_done=True`) — the talker processes requests strictly
sequentially, so request *N+1*'s thinker has plenty of time to complete its
stream before the talker reaches request *N+1* in the deferred queue.

## What this means for AC-10

- **AC-10 acceptance evidence is unmet on this workload.** No measurable TTFT
  improvement; no recorded partial-start firing.
- **AC-1..AC-9 acceptance is met** (28/28 unit tests, log line proves the code
  paths are wired correctly, all invariants preserved).

## Where the benefit lives

The mechanism would deliver real TTFT improvement when the thinker's chunk
arrival timeline is *slower* than the talker's idle tick cadence. Candidate
regimes:

1. **Longer thinker generation** (e.g., 500–2000 thinker tokens for narration
   or long-form replies) — even at 7.5 ms/token, that is 3–15 s of streaming,
   which spans many talker ticks.
2. **Slower inter-stage IPC** — multi-host topology where the thinker streams
   chunks over the network rather than local ZMQ.
3. **Talker saturated with prefill/decode of other batches** when a new
   request arrives — between batches the talker would re-check deferred
   payloads and find a partial prefix.
4. **Chunked-prefill or thinker-side delays** that slow the leading chunks of
   a stream.

The implementation is ready for those regimes; the present validation just
shows the H200 single-request short-TTS path is not where the wins land.

## Recommendations

- Keep the partial-start path merged but **leave `partial_start_min_chunks` at
  the disabled-default** (`None`) on production configs until a regime that
  actually exercises the path is identified.
- Treat the present PR as **landing the mechanism (Step 1 of the RFC)**. The
  RFC's Step 2 (tune a non-zero default + bench TTFT improvement) is now
  **gated on finding a workload where the partial-start window actually opens**
  — long-form narration, slower IPC, or saturated talker.
- Consider a follow-up that adds a *scheduler-side delay knob* or *forced tick
  yield* to ensure the deferred recheck runs between chunks — that would
  guarantee partial-start can fire on fast hardware. Out of scope for this PR.

## Codex code review (independent)

Run on the final diff before the im_end fix:

- All six contract points preserved (knob disabled by default, builder split,
  real thinker_done plumbed, no-op `_initialize_request_stream_state`, abort
  filter at `recv_requests`, no `omni_scheduler.py` touch).
- One blocking issue surfaced and fixed in commit `aa78fb3`: count usable
  prefetched chunks (filter `<|im_end|>`), reject non-positive thresholds,
  demote per-build log to DEBUG.
- Non-blocking notes accepted as follow-up: tighten `test_request_builder_*`
  to exercise the real `make_talker_scheduler_adapters` closure.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -371,9 +371,36 @@ def _init_parallel_state(self, tp_worker: Any) -> None:
...
+            # Roll back upstream's prepare_for_decode alloc + counters.
+            self._rollback_decode_prep_after_skip(batch)
             return None
         return batch
 
+    def _rollback_decode_prep_after_skip(self, batch: Any) -> None:
+        if not batch.forward_mode.is_decode():
+            return
+        if batch.out_cache_loc is not None:
+            self.token_to_kv_pool_allocator.free(batch.out_cache_loc)
```
zhaochenyang20: the helper only looks sufficient under a very narrow Qwen talker configuration, but it is installed in the shared `OmniScheduler.get_next_batch_to_run()` path. When a partial-start request has finished prefill but is temporarily missing the next text row, `_Upstream.get_next_batch_to_run()` has already called `ScheduleBatch.prepare_for_decode()` before `_rollback_decode_prep_after_skip()` runs. Upstream `prepare_for_decode()` does more than allocate `out_cache_loc` and increment `seq_lens`; it also writes `req_to_token_pool`, advances the sampling penalizer orchestrator, and under overlap or Mamba-related settings can replace length tensors or write extra buffers. This rollback only frees `out_cache_loc` and decrements `kv_*` plus length counters, with no contract or tests for those additional side effects. Since this PR is introducing a reusable downstream streaming-startup pattern, putting a partial rollback in the shared scheduler means the second consumer or a later talker configuration can inherit an incomplete paused-decode state; the impact is polluted sampling history, stale token-pool mappings, or confusing memory-check behavior, and those failures happen while the request is merely waiting for the next normal text row.

Do not make “prepare first, then partially roll back” the shared default path. The safer fix is to perform the live-request decode-input readiness check in the talker-specific scheduler before entering the upstream branch that calls `prepare_for_decode()`. If the state truly can only be known after the prefill merge, keep that pause logic scoped to `QwenTalkerScheduler` and document the exact invariants for every upstream side effect that is intentionally preserved. If the rollback remains, it needs to cover the full `prepare_for_decode()` side-effect set: `req_to_token_pool` writes, sampling penalizer state, overlap tensor replacement, and Mamba extra buffers, with unit tests showing that repeated stalls leave those states identical to the state before any decode attempt.

tests/unit_test/qwen3_omni/test_talker.py
zhaochenyang20, line 1: change `tests/readme.md`

zhaochenyang20: These tests provide too much false confidence. One of the key behaviors in this PR is that the real `request_builder` returned by `make_talker_scheduler_adapters()` must pass `thinker_done=False` to `build_prompt_prefill()` on the partial path and then pass the same state into `build_sglang_talker_request()`. However, `_make_request_builder_stub_env()` hand-writes a simplified request builder inside the test file, and the related tests validate that copied logic. If the production closure drops `prompt_prefill["input_ids"]`, `tts_pad_embed`, the sampling seed, fallback state, or if a future refactor accidentally hard-codes `thinker_done=True` again, these tests can still pass. The practical impact is a false CI pass exactly at the request-builder boundary that partial-start depends on, leaving the regression to surface only in GPU integration testing or runtime.

Construct the request builder through the real `make_talker_scheduler_adapters()` using a lightweight fake model, tokenizer, and config, then monkeypatch `TalkerPrefillBuilder.build_prompt_prefill()` and `build_sglang_talker_request()` to record arguments. Alternatively, extract the production request-builder closure into a private helper and test that helper directly. In either approach, the assertions should cover the real closure’s propagation of `prefetched_chunks`, `prefetched_stream_done`, fallback behavior, `pending_text_queue`, `input_ids`, `tts_pad_embed`, and `sampling_seed`.

sglang_omni/models/qwen3_omni/bootstrap.py
```diff
@@ -102,6 +102,7 @@ def create_talker_scheduler(
     tp_rank: int = 0,
     nccl_port: int | None = None,
     total_gpu_memory_fraction: float | None = None,
+    partial_start_min_chunks: int | None = None,
```
zhaochenyang20: This is an API/configuration cleanup opinion rather than a correctness blocker. To express the legacy path, `partial_start_min_chunks` is now threaded as `int | None` through the default config, bootstrap, stage factory, scheduler, CLI, and tests: `None` means disabled, non-positive values are remapped to disabled, and positive values are still floored by `MIN_PARTIAL_START_CHUNKS`. Users effectively see three states, but only the positive threshold is meaningfully tunable. The risk is a larger configuration state space, extra tests for a disabled sentinel that does not tune performance, and more room for CLI, stage factory, and scheduler semantics to diverge when the default is promoted later. Since this PR's purpose is to introduce talker partial-start and the threshold already has a safety floor, I would not keep `partial_start_min_chunks=None` as a long-term public knob. Prefer a positive default threshold and let the CLI tune that threshold; if a kill switch is still required, expose a separately named boolean instead of overloading a count field with both disable and threshold semantics.

Pick one explicit configuration shape: either make partial-start the default talker path and keep a positive integer threshold, or split the API into `enable_partial_start` plus `partial_start_min_chunks`. Then remove the `None` sentinel from the default config, the branch that maps non-positive values back to disabled, and tests that specifically assert the factory arg is `None`.

zhaochenyang20 (changes requested):

Hayden727:
@zhaochenyang20 review comments addressed in `ef6fb56`:

**#1 (omni_scheduler partial rollback)** — moved `get_next_batch_to_run` override + `_rollback_decode_prep_after_skip` + `self_check_during_idle` off shared `OmniScheduler` and onto `QwenTalkerScheduler`. Documented invariant in-class: safe only under talker `server_args` (overlap disabled via `configure_talker_server_args`, no Mamba). Added unit tests covering the rollback contract:
- `test_rollback_decode_prep_after_skip_is_idempotent_across_repeated_stalls` — counters return to pre-`prepare_for_decode` state on repeated stalls.
- `test_rollback_decode_prep_after_skip_is_noop_for_prefill_batches` — never fires on prefill batches.

**#3 (false-confidence stubbed request_builder)** — extracted the production closure body into a module-level `_build_talker_request_data` helper that takes all deps as explicit kwargs; `make_talker_scheduler_adapters` now just binds and forwards. New tests drive the **real** helper and cover propagation of `input_ids`, `tts_pad_embed`, `pending_text_queue`, `talker_model_inputs`, `sampling_seed` (both derived-from-rid and explicit pass-through), `tts_eos_embed`, `stage_payload` round-trip, plus the partial-zero-chunks `RuntimeError` and done-empty-fallback `ValueError` paths. The hand-written stub copy is gone.

**#4 (None sentinel + non-positive remap)** — collapsed to a clean split: `enable_partial_start: bool` + `partial_start_min_chunks: int` (must be ≥ `MIN_PARTIAL_START_CHUNKS = 3`, otherwise constructor raises `ValueError`). Removed the `None` sentinel, the non-positive remap branch, and the test asserting the factory arg is `None`. CLI now has `--enable-partial-start` plus the positive-integer `--partial-start-min-chunks` (only consumed when enabled).

**#2 (tests/README.md)** — updated `unit_test/qwen3_omni/` description to list partial-start gate, real `_build_talker_request_data` propagation, and rollback idempotency coverage.

All 417 unit tests pass on ion8-h200-omni (33 talker + 384 cross-module). PTAL.

Hayden727:
Independent Codex review caught two residual issues on top of the earlier reviewer pass; fixed in `7bb4c67` + `b337f65`:

**Codex #1 — Comment #4 API not fully threaded.** `bootstrap.py` and `stages.py` still typed `partial_start_min_chunks: int | None = None` and bootstrap conditionally omitted it, leaving a three-state path for external/YAML callers. Tightened to `partial_start_min_chunks: int = 5` everywhere (config / bootstrap / stage factory / scheduler / CLI), always forwarded. The constructor `ValueError` on `< MIN_PARTIAL_START_CHUNKS` is the single enforcement point.

**Codex #2 — rollback docstring overstated the invariant.** The QwenTalkerScheduler docstring now enumerates exactly what `ScheduleBatch.prepare_for_decode` writes/clears and explains why each not-rolled-back attribute is safe under talker `server_args`: `disable_overlap_schedule=True` covers `enable_overlap` + `attn_cp_metadata`; `mamba_track_*` unused (no Mamba); `sampling_info.penalizer_orchestrator` scatter is idempotent on the same output_id; `is_spec_v2` false; `hisparse_coordinator` unused.

Added `test_prepare_for_decode_side_effect_contract_with_upstream`: introspects upstream source for `self.X` references and asserts every attribute is either rolled back or in the documented-invariant set. Fires on future sglang bumps that add new writes, forcing us to either extend the rollback or document the new invariant. Initial run surfaced six attributes I had missed (sampling_info, enable_overlap, is_spec_v2, mamba_track_indices, mamba_track_mask, prepare_encoder_info_decode); each is now classified.

Verification on ion8-h200-omni: 137 / 137 unit tests pass.

sglang_omni/models/qwen3_omni/talker_scheduler.py
```diff
@@ -65,6 +142,44 @@ def _is_batch_ready_to_run(self, batch: Any) -> bool:
             return False
         return True
 
+    # ------------------------------------------------------------------
```
yxs: remove

sglang_omni/models/qwen3_omni/talker_scheduler.py
```diff
@@ -65,6 +142,44 @@ def _is_batch_ready_to_run(self, batch: Any) -> bool:
...
+    # Talker-scoped decode-stall handling. Shared OmniScheduler stays
+    # pure; the side-effect rollback is safe only under the talker
+    # server_args (disable_overlap_schedule=True, no Mamba).
+    # ------------------------------------------------------------------
```
yxs: remove

sglang_omni/models/qwen3_omni/talker_scheduler.py
```diff
@@ -33,22 +35,97 @@ def configure_talker_server_args(
...
 
 
+# Floor for the partial-start gate; pinned by
+# ``test_partial_prompt_prefill_layout_invariants``.
+MIN_PARTIAL_START_CHUNKS = 3
+
+
 class QwenTalkerScheduler(OmniScheduler):
-    """Talker scheduler with Qwen-specific request and decode readiness."""
+    """Talker scheduler with Qwen-specific request and decode readiness.
```
yxs: Remove all AI-generated comments, add note like
`Note(Chenchen): xxx`
if needed.

sglang_omni/models/qwen3_omni/talker_scheduler.py
```diff
@@ -33,22 +35,97 @@ def configure_talker_server_args(
...
+    ``output_ids``, ``seq_lens`` / ``seq_lens_cpu`` / ``orig_seq_lens`` /
+    ``seq_lens_sum``, per-req ``decode_batch_idx`` / ``kv_committed_len`` /
+    ``kv_allocated_len``, ``forward_mode``, ``input_embeds``,
+    ``attn_cp_metadata``, ``sampling_info.penalizer_orchestrator``
+    cumulate-output-tokens state, ``hisparse_coordinator``, Mamba buffers,
+    and ``req_to_token_pool`` writes.
+
+    This rollback only undoes the first set explicitly. The remainder is
+    safe **only** because the talker config disables them or makes them
+    idempotent: ``configure_talker_server_args`` sets
```
yxs: "idempotent within one step" is wrong, check the note I added.

yxs: This claim is incorrect. In sglang 0.5.8:

- `BatchedFrequencyPenalizer._cumulate_output_tokens` is `scatter_add_` (accumulator)
- `BatchedMinNewTokensPenalizer._cumulate_output_tokens` is `len_output_tokens += 1` (accumulator)
- Only `BatchedPresencePenalizer._cumulate_output_tokens` is `scatter_` (idempotent)

What actually keeps the talker safe across repeated stalls is that `_resolve_talker_sampling_config` leaves `frequency_penalty` / `presence_penalty` / `min_new_tokens` at the `SamplingParams` defaults (all 0), so every individual penalizer's `_is_required` returns False, the orchestrator's OR-aggregated `is_required` stays False, and upstream `prepare_for_decode` never enters the `if self.sampling_info.penalizer_orchestrator.is_required:` branch at all. Talker's own repetition penalty runs through the custom sampler in `components/talker.py`, bypassing this orchestrator entirely.

sglang_omni/models/qwen3_omni/request_builders.py
```diff
@@ -1003,3 +989,89 @@ def result_adapter(data: SGLangARRequestData) -> StagePayload:
...
+    params = payload.request.params
+    sampling_cfg = resolve_sampling_config(params)
+    if sampling_cfg.get("seed") is None:
+        sampling_cfg["seed"] = (
+            xxhash.xxh64_intdigest(str(payload.request_id).encode("utf-8")) & 0x7FFFFFFF
+        )
+    thinker_chunks = list(payload.prefetched_chunks)
+    thinker_done = bool(payload.prefetched_stream_done)
+
+    if not thinker_chunks:
```
yxs, line 995: `_fallback_thinker_chunks_from_state` is unreachable under the new wiring.

Talker's payload now comes from `project_mm_aggregate_to_talker_ar`, which keeps only `prompt` and `thinker_inputs` and strips `thinker_out` (and `engine_outputs`). The fallback reads `state.thinker_out or state.engine_outputs.get("thinker")` and therefore always returns `[]`, so the subsequent `raise ValueError(...)` is the only outcome on this path.

`test_real_builder_uses_fallback_chunks_on_done_path_only` exercises a stub fallback, not the production one, so the dead path isn't surfaced.

Suggest collapsing the two raise branches into one and removing both `_fallback_thinker_chunks_from_state` and the `fallback_chunks_from_state` parameter (plus the related test). Cleaner invariant: "under the new wiring, the partial-start path must have chunks; if not, it's a wiring bug upstream."

sglang_omni/models/qwen3_omni/talker_scheduler.py
```diff
@@ -33,22 +35,97 @@ def configure_talker_server_args(
...
+        if partial_start_min_chunks < MIN_PARTIAL_START_CHUNKS:
+            raise ValueError(
+                f"partial_start_min_chunks must be >= {MIN_PARTIAL_START_CHUNKS}, "
+                f"got {partial_start_min_chunks}"
+            )
+        self._enable_partial_start = bool(enable_partial_start)
+        self._partial_start_min_chunks = int(partial_start_min_chunks)
+        self._im_end_token_id = im_end_token_id
+
+    def _count_usable_prefetched_chunks(self, prefetched: list[Any]) -> int:
```
yxs, line 60: The gate strips every `<|im_end|>` chunk; `build_prefill_input` (in `talker_input.py`) strips only the trailing one. In practice the thinker emits `<|im_end|>` exactly once at the end, so both counts agree — but the semantics differ.

If thinker behavior changes (e.g., mid-stream `<|im_end|>` as a multi-turn separator), the two counts diverge: the gate thinks the threshold is met, but `build_prefill_input` projects the mid-stream `<|im_end|>` as a normal token into `assistant_embed`, and `future_text_rows` shifts by one.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -387,11 +387,9 @@ def _init_parallel_state(self, tp_worker: Any) -> None:
             self.attn_tp_rank == 0 or self.enable_metrics_for_all_schedulers
         )
 
-    def get_next_batch_to_run(self):
-        batch = _Upstream.get_next_batch_to_run(self)
-        if batch is not None and not self._is_batch_ready_to_run(batch):
-            return None
-        return batch
+    # ------------------------------------------------------------------
```
yxs: remove

yxs (commented):
Follow-up code-style review against bdb3017 (post-merge with main). All findings are NET-NEW versus the prior round + the 3 fixes I pushed in 6e8f6df.

---

## P0

### 1. `_rollback_decode_prep_after_skip` assert is stripped under `python -O`

`sglang_omni/models/qwen3_omni/talker_scheduler.py:161-164`

The defensive `assert isinstance(batch.seq_lens_sum, int)` I added in 6e8f6df is stripped under `python -O`, exactly the runtime mode prod tends to use, which would silently defeat the upgrade tripwire. The contract test scans attribute names only and won't catch a type change (sglang main has already switched `seq_lens_sum += bs` to `= None`), so this assert is the only runtime guard.

```python
if not isinstance(batch.seq_lens_sum, int):
    raise TypeError(
        f"seq_lens_sum is {type(batch.seq_lens_sum).__name__}, expected int "
        "— sglang upstream prepare_for_decode changed; update rollback."
    )
```

### 2. `--colocated` doesn't actually colocate by default

`examples/run_qwen3_omni_speech_server.py:147-154, 261-302`; `sglang_omni/models/qwen3_omni/config.py:226-246, 295-311`

Even when `args.colocated=True`, the launcher then unconditionally overrides each stage's GPU via `_set_stage_gpu(...)` with the CLI defaults (`--gpu-thinker=0`, `--gpu-talker=1`, `--gpu-code2wav=0`), which is disaggregated. Verified with a no-GPU mock-launch reproduction.

Also: `_SPEECH_COLOCATED_PROCESSES` is byte-identical to `_SPEECH_DEFAULT_PROCESSES` (diffed key/value), so `Qwen3OmniSpeechColocatedPipelineConfig` differs from the default only in the `talker_gpu` field default, which the launcher immediately overrides anyway. There's no test pinning them to be different either, so they will silently drift over time.

Suggest either (a) dropping `--colocated`, `_SPEECH_COLOCATED_PROCESSES`, and `Qwen3OmniSpeechColocatedPipelineConfig` entirely and relying on default + GPU overrides, or (b) when `--colocated` is set, enforce `args.gpu_thinker == args.gpu_talker == args.gpu_code2wav` and skip the default-1 talker assignment. (a) is cleaner unless the colocated class is about to carry real distinct behavior.

### 3. Tests mutate `OmniScheduler.__init__` and `rb_mod.build_sglang_talker_request` at class/module scope

`tests/unit_test/qwen3_omni/test_talker.py:464`, `:970`; `_drive_real_builder` at ~`:368-394`

`OmniScheduler.__init__ = lambda self, *a, **k: None` mutates a class-level method globally; if pytest is ever run with `-n auto` (xdist) or any other test imports `OmniScheduler` while this one is in-flight, the swap leaks into sibling tests. Same hazard in `_drive_real_builder` rebinding `rb_mod.build_sglang_talker_request` globally.

Use `monkeypatch.setattr(...)` so pytest restores on teardown and fails loudly on collisions:

```python
def test_partial_rejects_min_chunks_below_layout_floor(monkeypatch) -> None:
    monkeypatch.setattr(OmniScheduler, "__init__", lambda self, *a, **k: None)
    ...
```

---

## P2

### 4. Docstring's "penalizer is idempotent within one step" claim is wrong

`sglang_omni/models/qwen3_omni/talker_scheduler.py:64-66`; mirror at `tests/unit_test/qwen3_omni/test_talker.py:1119-1120`

In sglang 0.5.8:
- `BatchedFrequencyPenalizer._cumulate_output_tokens` is `scatter_add_` (accumulator)
- `BatchedMinNewTokensPenalizer._cumulate_output_tokens` is `len_output_tokens += 1` (accumulator)
- Only `BatchedPresencePenalizer._cumulate_output_tokens` is `scatter_` (idempotent)

What actually keeps the talker safe is that `_resolve_talker_sampling_config` leaves `frequency_penalty` / `presence_penalty` / `min_new_tokens` at `SamplingParams` defaults (all 0), so each penalizer's `_is_required` returns False, the orchestrator's OR-aggregated `is_required` stays False, and upstream `prepare_for_decode` never enters the `if self.sampling_info.penalizer_orchestrator.is_required:` branch at all. Talker's own repetition penalty runs through the custom sampler in `components/talker.py`, bypassing the orchestrator entirely. Rewrite the docstring sentence and the mirrored test comment accordingly.

### 5. `_fallback_thinker_chunks_from_state` is dead code under the new wiring

`sglang_omni/models/qwen3_omni/request_builders.py:209-218, 922-943, 1020-1031`

The talker's payload comes from `project_mm_aggregate_to_talker_ar`, which keeps only `prompt` and `thinker_inputs` and strips `thinker_out` / `engine_outputs`. The fallback reads `state.thinker_out or state.engine_outputs.get("thinker")` and therefore always returns `[]`, so the subsequent `raise ValueError(...)` is the only outcome.

`test_real_builder_uses_fallback_chunks_on_done_path_only` exercises a stub fallback (not the production one), so the dead path is not surfaced. Suggest collapsing the two raise branches into one and removing `_fallback_thinker_chunks_from_state` + the `fallback_chunks_from_state` parameter (and the related stub-driven test). Cleaner invariant: "under the new wiring, if there are no chunks at build time, it's a wiring bug upstream."

### 6. `--partial-start-min-chunks` validation deferred to scheduler init

`examples/run_qwen3_omni_speech_server.py:136-145`; `sglang_omni/models/qwen3_omni/talker_scheduler.py:87`

Help text says "must be >= MIN_PARTIAL_START_CHUNKS (3)" but `_launch_speech_server` doesn't validate `args.partial_start_min_chunks` before constructing the config; the `ValueError` only fires inside `QwenTalkerScheduler.__init__`, several layers deep into the process spawn. Fail fast at parse time:

```python
if args.enable_partial_start and args.partial_start_min_chunks < 3:
    raise ValueError(
        f"--partial-start-min-chunks must be >= 3, got {args.partial_start_min_chunks}"
    )
```

Or import `MIN_PARTIAL_START_CHUNKS` and reference the constant directly.

### 7. `test_prepare_for_decode_side_effect_contract_with_upstream` heuristic is weak

`tests/unit_test/qwen3_omni/test_talker.py:1132-1216`

The regex `r"self\.([A-Za-z_][A-Za-z0-9_]*)"` matches reads as well as writes, so the `novel = self_assignments - expected_writes - ...` subtraction is misleadingly named — most entries in the ignore-set are reads, not writes. And the 50+ hardcoded "safe to ignore" attribute names have to be edited on every sglang bump.

`seq_lens_sum` is the concrete case: sglang main has changed `self.seq_lens_sum += bs` to `self.seq_lens_sum = None`, but this contract test stays green either way; `_rollback_decode_prep_after_skip`'s `-=` would TypeError mid-stall. The defensive `isinstance` check I pushed in 6e8f6df catches it at runtime, but the durable fix is a smoke that imports `ScheduleBatch`, constructs a minimal instance, runs real `prepare_for_decode`, and asserts each rolled-back attribute's type — running binary-imported live code instead of regex-scanning source.

Also: `MIN_PARTIAL_START_CHUNKS`'s comment claims this test pins the constant, but `test_partial_prompt_prefill_layout_invariants` uses literals `1, 2, 3, 4, 5, 10`. Bumping the constant to 4 wouldn't fail it. Either reference the constant directly (`for n in range(1, MIN_PARTIAL_START_CHUNKS):`) or drop the claim.

### 8. `_count_usable_prefetched_chunks` vs `build_prefill_input` `<|im_end|>` semantics diverge

`sglang_omni/models/qwen3_omni/talker_scheduler.py:96-108`; `sglang_omni/models/qwen3_omni/components/talker_input.py:230-238`

The gate strips every `<|im_end|>` chunk; `build_prefill_input` strips only the trailing one. In practice the thinker emits `<|im_end|>` exactly once at the end, so both counts agree today — but the semantics differ. If thinker behavior changes (e.g. mid-stream `<|im_end|>` as a turn separator for multi-turn), the two counts diverge: the gate thinks the threshold is met, but `build_prefill_input` projects the mid-stream `<|im_end|>` as a normal token into `assistant_embed`, and `future_text_rows` shifts by one. Align both to "strip trailing only" and add a unit test with a mid-stream `<|im_end|>` chunk.

---

## P3 / P4 (minor — batch in one cleanup commit)

**Test scaffolding**
- `tests/unit_test/qwen3_omni/test_talker.py`: `import pytest` inside test bodies at lines 159, 476, 495 — promote to file header.
- `tests/unit_test/qwen3_omni/test_talker.py:654`: `from sglang_omni.models.qwen3_omni.config import _talker_stage` reaches a `_`-prefixed module-internal symbol from tests — either promote to public or read via an instantiated config.
- `tests/unit_test/qwen3_omni/test_talker.py:399, 408, 417, 430, 435, 440, 465`: `req_data, captured = _drive_real_builder(...)` callers ignore `req_data` — use `_, captured = ...` to make intent obvious.

**Style / hygiene**
- `sglang_omni/models/qwen3_omni/talker_scheduler.py:73-76`: class-level mutable defaults exist for `object.__new__` test convenience — production-class state for a test convenience is a smell.
- `sglang_omni/models/qwen3_omni/talker_scheduler.py:185`: `_Upstream.self_check_during_idle(self)` bypasses `OmniScheduler.self_check_during_idle` (which resets `new_token_ratio`); use `super().self_check_during_idle()` unless the bypass is intentional.
- `sglang_omni/models/qwen3_omni/request_builders.py:1015`: magic mask `& 0x7FFFFFFF` — name it (e.g. `MAX_INT32_POSITIVE`) with a one-line note that PyTorch sampling_seed must be a positive int32.
- `sglang_omni/models/qwen3_omni/request_builders.py:1005-1006`: `resolve_sampling_config: Any` / `fallback_chunks_from_state: Any` should be typed `Callable[..., ...]`.
- `sglang_omni/models/qwen3_omni/talker_scheduler.py:151, 158`: `get_next_batch_to_run` and `_rollback_decode_prep_after_skip` are missing return-type hints (`-> Any | None` and `-> None`).
- `sglang_omni/models/qwen3_omni/config.py:159-162`: the `enable_partial_start` block comment doesn't follow the established `# Note (<name>):` convention used elsewhere in this file.
- `benchmarks/eval/benchmark_omni_streaming_ttft.py:150-158`: warmup loop captures `audio_chunks` and `status_code` but never logs them.
- `benchmarks/eval/benchmark_omni_streaming_ttft.py:247-248`: no run-id in default `--output` path → two invocations with the same `--label` silently overwrite each other.

Non-blocking — one cleanup commit can roll them up.

sglang_omni/models/qwen3_omni/config.py
```diff
@@ -74,7 +75,25 @@ def _audio_encoder_stage(*, gpu: int, process: str) -> StageConfig:
...
+            wait_for_fn=f"{_PKG}.request_builders.resolve_mm_aggregate_wait_sources",
+            merge_fn=f"{_PKG}.merge.merge_for_thinker",
+            next=["thinker", "talker_ar"],
+            route_fn=f"{_PKG}.request_builders.resolve_mm_aggregate_next_stages",
+            project_payload={
+                "talker_ar": (
+                    f"{_PKG}.request_builders.project_mm_aggregate_to_talker_ar"
+                ),
+            },
+        )
```
zhaochenyang20, line 96: the speech config now sends every audio request from `mm_aggregate` to `talker_ar` early regardless of whether `enable_partial_start` is enabled. In the default configuration `enable_partial_start=False`, so `QwenTalkerScheduler._is_request_build_ready()` keeps returning false until `stream_done`; however, `OmniScheduler._event_loop_normal()` calls `_take_deferred_request_payloads()` on every loop, and `process_input_requests()` then copies `payload.prefetched_chunks`, merges any new chunks, and puts the same payload back into `_deferred_request_payloads` at `sglang_omni/scheduling/omni_scheduler.py:432-445`. The default path therefore changes from “wake the talker only after thinker completion” to “replay the deferred talker payload on every scheduler tick until thinker completion.” The PR body reports MMMU audio request latency around 22 seconds; with the current roughly 1ms idle sleep, one default-disabled request can reprocess the deferred payload tens of thousands of times, while the chunk list grows with thinker output and is repeatedly shallow-copied. Under concurrent speech traffic this becomes CPU/GIL overhead and scheduler jitter exactly in the configuration the author says should preserve legacy behavior.

Make deferred talker-payload rechecks event-driven instead of polling them on every scheduler loop. One implementable fix is to mark request ids dirty from `_on_stream_chunk()` and `_on_stream_done()`, and have `_take_deferred_request_payloads()` return only payloads whose stream state changed; when no new chunk or done signal arrived, the same deferred payload should not be copied and replayed. Another option is to use the early `mm_aggregate -> talker_ar` route only for partial-start configurations while preserving the disabled path’s final-thinker-payload build semantics.

sglang_omni/models/qwen3_omni/talker_scheduler.py
```diff
@@ -65,6 +101,45 @@ def _is_batch_ready_to_run(self, batch: Any) -> bool:
...
+        batch = _Upstream.get_next_batch_to_run(self)
+        if batch is not None and not self._is_batch_ready_to_run(batch):
+            self._rollback_decode_prep_after_skip(batch)
+            return None
+        return batch
+
+    def _rollback_decode_prep_after_skip(self, batch: Any) -> None:
+        # Note(Chenchen Hong): This is talker-only. It does not fully invert
+        # prepare_for_decode; talker disables overlap/spec/Mamba/hisparse, and
+        # its SamplingParams defaults keep the upstream penalizer branch inactive.
```
zhaochenyang20: Moving the rollback from the shared `OmniScheduler` into `QwenTalkerScheduler` is the right direction, but the current contract still misses a real side effect. In the installed SGLang version, `ScheduleBatch.prepare_for_decode()` calls `alloc_for_decode()`, and `alloc_for_decode()` does more than return `out_cache_loc`: it also writes the newly allocated KV slot into `batch.req_to_token_pool`. The current `_rollback_decode_prep_after_skip()` frees `out_cache_loc` and rolls back `kv_allocated_len`, `kv_committed_len`, and the sequence-length tensors, but it does not restore or clear that `req_to_token_pool` cell. This may be harmless today because the next `prepare_for_decode()` overwrites the same position, and abort/release follows the rolled-back `kv_allocated_len`; the problem is that this safety relies entirely on implicit reasoning. The code comment only mentions overlap/spec/Mamba/hisparse and the penalizer branch, and the tests assert only `seq_lens_sum` type, length tensors, and freed allocator behavior. If a future SGLang update reads `req_to_token_pool` between the stalled decode and the next prepare for memory checking, offload, prefix-cache insertion, or batch filtering, this becomes a freed-slot stale mapping exposure.


Either make the rollback cover the `alloc_for_decode()` side effect by recording each request’s target token-pool position and previous value before prepare, then restoring or clearing those positions on stall; or document the invariant explicitly and pin it with a live-code contract test: the stale `req_to_token_pool` entry must not be read before the next prepare and must not be misused by the release path. That test should drive real `ScheduleBatch.prepare_for_decode()` with a recording fake `req_to_token_pool` and allocator, then assert that rollback followed by abort/release does not free the stale slot, that the next prepare overwrites the same position, and that every allowed intermediate check avoids reading the freed mapping.

sglang_omni/models/qwen3_omni/config.py
zhaochenyang20: This is a low-level maintainability issue, not a launch correctness bug. `Qwen3OmniSpeechColocatedPipelineConfig` now has one meaningful difference: it defaults thinker, talker, and code2wav to GPU 0. However, `_SPEECH_COLOCATED_PROCESSES` and `_SPEECH_DEFAULT_PROCESSES` are byte-for-byte identical dictionaries. The duplicate name makes the reader think the colocated class also changes process topology when it does not, and any future edit to the default process map must remember to update an identical copy. 

Have the colocated config reuse `_SPEECH_DEFAULT_PROCESSES` directly, or, if a distinct process topology is actually planned, introduce the difference in this PR and add a test that explains why the maps diverge. In the current version there is no difference, so keeping the second dictionary only adds drift surface.

zhaochenyang20 (dismissed):

yxs (approved):

Issue #450, opened 2026-05-15
[RFC] Factor BaseTTSStageScheduler to share stage-adapter machinery across TTS models


PopSoda2002:
## Background

`sgl-omni` V1 currently ships two TTS-style models (`fishaudio_s2_pro`, `higgs_tts`) and each has its own stage-adapter scheduler. These two files are nearly identical:

| File | LOC |
|---|---|
| `sglang_omni/models/fishaudio_s2_pro/fish_scheduler.py` | 488 |
| `sglang_omni/models/higgs_tts/higgs_scheduler.py` | 477 |
| Shared content (sorted-line overlap) | 433 (~90%) |
| Distinct content | 107 lines |

What actually differs:

- `is_finished()` — Fish checks `output_token_id == im_end_token_id`; Higgs checks `model._slots[req_id].sampler.generation_done` (state-machine flag).
- `release_state()` on finish — Fish clears `previous_semantic_tokens` + `last_codebook_values`; Higgs calls `model.reset_request(req_id)`.

The remaining ~90% — `BatchPlanner` (select / build / chunked prefill / post-step / inactive-state prune), `ResourceManager` (KV release glue), `IterationController` skeleton, and the main thread loop (recv → process → schedule → run_batch → emit_finished → start/stop/abort) — is copy-pasted between the two files.

## Problem

1. **Bug fixes don't propagate.** Any fix to chunked-prefill boundaries, KV release timing, batch construction, or IPC deadlock defense has to be applied in two places.
2. **High onboarding cost for new TTS models.** Adding a third TTS model (e.g. Voxtral, SNAC-style codecs, etc.) effectively means duplicating 470+ lines as a template.
3. **No shared regression coverage.** Test suites for `s2-pro` and `higgs-tts` each separately exercise the same machinery; a defect that would catch both stays uncaught if only one happens to be tested.

## Proposal

Factor a `BaseTTSStageScheduler` into `sglang_omni/scheduling/tts_stage_scheduler.py` that owns the shared stage-adapter machinery. Subclasses plug in only the finish semantics and per-request state cleanup.

```python
class BaseTTSStageScheduler:
    """Stage-adapter scheduler for TTS-style multi-codebook models.

    Wraps sglang's prefill / decode managers + KV pool and provides the
    sgl-omni stage IPC plumbing. Subclasses override `is_finished` and
    optionally `release_state` to plug in their finish semantics and any
    per-request model-side state cleanup.
    """

[...]

PR #428, opened 2026-05-11
feat(higgs-tts): add Higgs Audio v3 TTS support

PopSoda2002:
image
Quality passed, inference performance still needs sometime for improvement

PR #476, opened 2026-05-18
refactor(higgs-tts): replace HiggsScheduler with OmniScheduler

PopSoda2002:
## Summary

Replaces the custom `HiggsScheduler` machinery (478 lines) with the existing `OmniScheduler` that already powers Qwen3-Omni. Implements RFC #450.

**Net diff: -440 LOC** (5 files changed).

### What gets deleted

`sglang_omni/models/higgs_tts/higgs_scheduler.py` (entire file):
- `HiggsScheduler` — duplicated `OmniScheduler`'s queue + lifecycle plumbing
- `HiggsBatchPlanner` — re-implemented sglang's upstream batch selection (`OmniScheduler` delegates to it for free via `__getattr__`)
- `HiggsResourceManager` — KV release + model slot cleanup
- `HiggsIterationController` — finish condition checked `data.generation_done` set by the model runner

### What replaces them

Three small hooks let the existing `OmniScheduler` drive Higgs:

1. **Sampler-driven finish**: `HiggsTTSModelRunner._collect_step_outputs` sets `req.finished_reason = FINISH_MATCHED_TOKEN(EOC_ID)` when the sampler hits `generation_done` after EOC + delay wind-down. Upstream sglang's native `req.finished()` flow takes over; no custom iteration controller needed.

2. **Slot cleanup on finish**: `make_higgs_scheduler_adapters(model)` takes the `HiggsTTSModel` instance; the `result_adapter` calls `model.reset_request(rid)` once a result payload has been built. Per-request multi-codebook slots get freed as part of the normal finish path.

3. **Slot cleanup on abort**: `OmniScheduler` grows a tiny `on_abort: Callable[[str], None] | None` callback that fires from `abort()`. Wired to `model.reset_request` so client-cancelled requests drop their slot too.

### Plus two correctness fixes uncovered during E2E

- `sampling_params.normalize(tokenizer=None)` after construction — without it, `stop_strs` / `stop_regex_strs` stay `None` and the upstream scheduler's native `Req.check_finished()` trips on `len(None)`.
- Dropped redundant engine-level `max_new_tokens` knob. Single source of truth is `state.max_new_tokens` per request → `SamplingParams.max_new_tokens` → sglang's native `FINISH_LENGTH`. The engine-factory kwarg was just a second knob doing the same thing.

## Why

- Single AR scheduler implementation across the project — fixes and features land in one place instead of three (Higgs / Fish / Qwen3-Omni each had its own)
- `OmniScheduler` delegates to upstream sglang `Scheduler` via attribute lookup, so we get proper paged-attention batching / KV cache / radix integration without re-implementing it
- The input-length validation `input_len + max_new_tokens > max_req_len` is already done by `OmniScheduler._request_kv_capacity_error` for free — we no longer bypass the sglang-equivalent check

### Quality (seed-tts en, matched 100)

| metric | HiggsScheduler (#428) | OmniScheduler (this PR) | Δ |
|---|---|---|---|
| avg WER ↓ | 0.0154 | **0.0170** | +0.0016 |
| median WER | 0.0 | **0.0** | tied |
| avg speaker_sim ↑ | 64.31 | **63.70** | -0.61 |
| median speaker_sim | — | **64.71** | — |

Within sampling noise — model / sampler / delay-pattern code is unchanged; differences come from concurrent batch ordering perturbing per-request RNG state.

### Latency

| C | Mean Lat | p95 Lat |
|---|---|---|
| | OmniScheduler / HiggsScheduler | OmniScheduler / HiggsScheduler |
| 1  | **3206** / 4789 ms (-33%) | **4915** / 20468 ms (-76%) |
| 4  | **2512** / 4372 ms (-43%) | **3674** / 24323 ms (-85%) |
| 8  | **3414** / 5248 ms (-35%) | **4406** / 27136 ms (-84%) |
| 16 | **4080** / 8476 ms (-52%) | **5857** / 34109 ms (-83%) |
| 32 | **6187** / 10424 ms (-41%) | **9344** / 40880 ms (-77%) |

### Throughput

| C | RTF agg (Omni / Higgs) | audio_s/s (Omni / Higgs) |
|---|---|---|
| 1  | 0.703 / **0.524** | 1.42 / **1.91** |
| 4  | 0.165 / **0.159** | 6.06 / **6.27** |
| 8  | **0.096** / 0.100 | **10.46** / 10.03 |
| 16 | 0.087 / **0.068** | 11.44 / **14.62** |
| 32 | 0.071 / **0.050** | 14.02 / **19.82** |

### Trade-off

- **Latency wins everywhere** — mean ~40% lower, p95 ~80% lower. Upstream sglang's `Scheduler` is fairer to individual requests than the custom `HiggsBatchPlanner`.
- **High-concurrency throughput slightly lower** — `HiggsBatchPlanner` was more aggressive about packing requests; sglang's selection prioritises fairness.
- **For `/v1/audio/speech` interactive UX**: clear win (p95 5s vs 30-40s).
- **For AReaL RL rollout (batch)**: small throughput regression at C ≥ 16; latency improvement may still offset.

## Test plan

- [x] `pre-commit run` (autoflake + isort + black + ruff) clean
- [x] N=100 seed-tts en quality eval — WER / sim parity confirmed
- [x] N=100 concurrent perf bench across C=1/4/8/16/32 — full table above
- [x] sampler-driven finish: requests stop on EOC + delay wind-down via `FINISH_MATCHED_TOKEN`
- [x] slot cleanup verified on both normal finish and `abort()`


JingwenGu0829:
# Higgs TTS OmniScheduler Benchmark

PR #476 refactors Higgs TTS to use the shared `OmniScheduler` instead of its custom `HiggsScheduler`, keeping Higgs-specific request/result adapters for sampler-driven finish, model-slot cleanup, request token caps, and usage timing while letting the common scheduler handle batching, KV/cache lifecycle, and abort behavior.

## Benchmark Setup

- Branch: `feat/higgs-tts-omni-scheduler`
- Model: `boson-sglang/higgs-audio-v3-tts-4b-base`
- Hardware: single H200 for generation
- Dataset: SeedTTS EN full set, `N=1088`
- Mode: voice-clone, CUDA graph on
- Sampling: `top_k=50`, `temperature=0.8`, `max_new_tokens=1024`
- Scheduling: server `max_running_requests=16`, client concurrency `32`

## Results

| Run | Success | Req/s | RTF Mean | Lat Mean | Lat p99 | WER | WER Excl >50% | >50% WER |
| 1 | 1088/1088 | 9.668 | 0.8280 | 3.283s | 5.127s | 1.40% | 1.32% | 1 |
| 2 | 1088/1088 | 10.297 | 0.7249 | 2.960s | 4.393s | 1.49% | 1.49% | 0 |
| 3 | 1088/1088 | 10.877 | 0.7183 | 2.911s | 4.032s | 1.28% | 1.28% | 0 |

## Summary

Across all three full-set runs, generation completed `3264/3264` requests with average throughput `10.28 req/s`, average RTF `0.757`, and average WER excluding >50% outliers `1.37%`. Runs 2-3 match the PR #503 CUDA-graph reference closely: average throughput `10.59 req/s`, average RTF `0.722`, and WER `1.39%`. Catastrophic WER was low overall: `1/3264` samples (`0.031%`).

tests/unit_test/higgs_tts/test_request_builders.py
zhaochenyang20, line 1: change the `tests/readme.md`

zhaochenyang20 (changes requested):

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -758,11 +764,49 @@ def abort(self, request_id: str) -> None:
...
+    def _release_request_kv_cache(self, req: Any) -> None:
+        if (
+            getattr(req, "req_pool_idx", None) is None
+            and getattr(req, "mamba_pool_idx", None) is None
+        ):
+            return
+        try:
+            release_kv_cache(req, self.tree_cache)
+        except Exception:
+            logger.exception("OmniScheduler: KV cleanup failed for %s", req.rid)
```
zhaochenyang20: Do not swallow `Exception` on an internal cleanup invariant. This should let `release_kv_cache` raise, or at most log and re-raise, so the batch is not silently filtered while resources remain allocated. The adjacent `getattr(batch, "reqs", [])`, `getattr(req, "req_pool_idx", None)`, and `getattr(req, "mamba_pool_idx", None)` should also become explicit attribute reads because `ScheduleBatch` and `Req` are internal typed contracts. If those fields are missing, the scheduler should fail fast instead of treating cleanup as an empty batch or an unallocated request.

sglang_omni/models/higgs_tts/request_builders.py
```diff
@@ -104,17 +110,34 @@ def apply_higgs_result(state: HiggsTtsState, data: HiggsSGLangRequestData) -> No
...
 
-def make_higgs_scheduler_adapters():
+def make_higgs_scheduler_adapters(model, *, max_new_tokens_cap: int | None = None):
+    """Build (request_builder, result_adapter) closures bound to a
+    :class:`HiggsTTSModel` instance.
+
+    The result adapter drops the model's per-request slot (sampler state +
+    accumulated codes) once a result is emitted so a long-running server
+    doesn't accumulate dead slots.
+    """
```
zhaochenyang20: Add explicit parameter and return types. The minimal fix can type the model object as `Any`, but the return type should clearly describe the `(request_builder, result_adapter)` callables. A better fix is to define a small `Protocol` requiring only `reset_request(self, req_id: str) -> None`.

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

Issue #486, opened 2026-05-19
[Perf] Qwen3-Omni talker stage: latency reduction with WER preservation


yxs:
main branch dde550c234813d4714e5e22e018b34e2e6665bc5

Hardware: 2xH200
Model: Qwen3-Omni-30B-A3B-Instruct
Dataset: seed-tts-eval, en full, 1088 samples
Round = 5, C = 16

Cell | rtf | qps | latency_mean(s) | WER_corpus | WER<50 | >50n mean (range)
  -- | -- | -- | -- | -- | -- | --
 VC=F, stream=True| 0.827 ± 0.017 | 5.36 ± 0.10 | 2.98 ± 0.06 | 2.23% | 2.00% | 8.0 (8-8)

## Goal

Compare reference baseline, we need reduce mean latency from from ~2.98s to ~1.34s (**−55%, ~2.23× faster**), mean RTF from ~0.827 to ~0.371 (−55%, ~2.23× faster), with WER preserved.

### ongoing perf work

#472 total IPC time across all 7 inter-stage edges in the current speech pipeline = ~12 ms per request (mean over 5R × 1088 reqs, σ small). That's 0.4% of wall.
Upper bound on wall reduction: **≤ 0.4%** (squeezing all measured IPC to zero).

#475 thinker total compute (stage_in → stage_out) = 170 ms ± 8 per request. This is the hard upper bound — at best, talker prefill fully overlaps with thinker decode, removing up to 170 ms of serialized waiting.
Upper bound on wall reduction: **≤ 5.7%** (full overlap, measured thinker compute as cap, applied to TTFT phase only).

### gap

Per-request wall split into TTFT (request_in → first audio chunk) and decode (first → last audio chunk)

| Phase | sglang current main | reference baseline | gap | % of total gap |
|---|---:|---:|---:|---:|
| TTFT | 970 ± 37 ms | 893 ± 7 ms | +77 ms | 4.7% |
| Decode (first → last audio) | 2012 ± 49 ms | 441 ± 3 ms | **+1571 ms** | **95.3%** |
| **Wall total** | 2982 ± 69 ms | 1334 ± 9 ms | +1648 ms | 100% |

TTFT is at parity (+77 ms, within single-σ noise). The entire ~1.6 s gap lives in the **streaming-audio decode phase** — after the first audio chunk reaches the client, the per-chunk emission loop on sglang-omni is **4.56× slower** than the reference baseline.

[...]

PR #487, opened 2026-05-19
[Qwen3-Omni] vectorize per-row Python loops in talker AR hot path

yxs:
Vectorize four per-row Python loops in the talker AR hot path. Each loop collected per-row index lists in Python and then issued one batched advanced-indexing scatter — same pattern across all four sites, no numerical operation changed

2×H200, seed-tts-eval, EN VC=F, C=16, stream=True, 5R restart-per-round mean:

| run | latency_mean(s) | rtf | qps | wer_corpus | wer<50 | n>50 |
| main `dde550c` (baseline) | 3.013 ± 0.07 | 0.836 ± 0.020 | 5.30 ± 0.13 | 2.72 % | 2.21 % | 8 |
| + `prepare_decode_buffers` alone (5R) | **2.591 ± 0.04** | 0.725 | 6.16 | 2.76 % | 2.25 % | 8 |
| + full bundle (5R) | **2.274 ± 0.018** | **0.641 ± 0.005** | **7.02 ± 0.05** | 2.75 % | 2.24 % | 8 |

**Cumulative: latency −24.5 %, RTF −23.3 %, QPS +32.4 %. WER and n50 unchanged.**

Closes, Part of #486.

### DIFF
--- a/sglang_omni/model_runner/base.py
+++ b/sglang_omni/model_runner/base.py
@@ -244,12 +244,13 @@ class ModelRunner:
             rep_toks.extend(unique)
             rep_penalties.extend([float(penalty)] * len(unique))
         if rep_rows:
+            orig_dtype = logits.dtype
             rows_t = torch.tensor(rep_rows, dtype=torch.long, device=device)
             toks_t = torch.tensor(rep_toks, dtype=torch.long, device=device)
-            pens_t = torch.tensor(rep_penalties, dtype=logits.dtype, device=device)
-            scores = logits[rows_t, toks_t]
+            pens_t = torch.tensor(rep_penalties, dtype=torch.float32, device=device)
+            scores = logits[rows_t, toks_t].to(torch.float32)
             scores = torch.where(scores > 0, scores / pens_t, scores * pens_t)
-            logits[rows_t, toks_t] = scores
+            logits[rows_t, toks_t] = scores.to(orig_dtype)
 
     def _apply_codec_suppress_tokens(self, logits_output: Any, requests: list) -> None:
         logits = logits_output.next_token_logits
--- a/sglang_omni/model_runner/base.py
+++ b/sglang_omni/model_runner/base.py
@@ -223,6 +223,11 @@ class ModelRunner:
         logits = logits_output.next_token_logits
         if logits is None or logits.ndim != 2:
             return
+        vocab = logits.shape[1]
+        device = logits.device
+        rep_rows: list[int] = []
+        rep_toks: list[int] = []
+        rep_penalties: list[float] = []
         for row_idx, sched_req in enumerate(requests):
             data = sched_req.data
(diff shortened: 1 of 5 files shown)


sglang_omni/model_runner/base.py
```diff
@@ -232,19 +237,28 @@ def _apply_repetition_penalty(self, logits_output: Any, requests: list) -> None:
...
-            scores = torch.where(scores > 0, scores / penalty, scores * penalty)
-            logits[row_idx, idx] = scores
+            rep_rows.extend([row_idx] * len(unique))
+            rep_toks.extend(unique)
+            rep_penalties.extend([float(penalty)] * len(unique))
+        if rep_rows:
+            rows_t = torch.tensor(rep_rows, dtype=torch.long, device=device)
+            toks_t = torch.tensor(rep_toks, dtype=torch.long, device=device)
+            pens_t = torch.tensor(rep_penalties, dtype=logits.dtype, device=device)
+            scores = logits[rows_t, toks_t]
```
zhaochenyang20: this shared pre-sampler path can change token ordering when `next_token_logits` is `float16`. The previous row-local code applied the Python float `penalty` directly in `scores / penalty` or `scores * penalty`; the new code constructs `rep_penalties` as `dtype=logits.dtype` at line 249, so common values such as 1.2 are rounded to reduced precision before the divide/multiply at line 251. A local PyTorch probe reproduces the difference: with `torch.float16`, `tensor([1.0]) / 1.2` produces `0.83349609375`, while dividing by a half tensor penalty produces `0.8330078125`. This is not just a display difference; this function mutates logits immediately before `tp_worker.model_runner.sample()`, so close candidate ordering, top-p boundaries, or greedy selection can flip even though the PR claims no numerical operation changed.


Keep the vectorized coordinate collection, but perform the penalty math in at least `float32` and cast the result back to the original logits dtype only when writing it into `logits`. Add a minimal regression test with half logits, `repetition_penalty=1.2`, and positive/negative repeated tokens, comparing the vectorized function against the previous scalar formula; the positive-logit divide path should be covered because it exposes the precision difference most clearly.

```python
orig_dtype = logits.dtype
pens_t = torch.tensor(rep_penalties, dtype=torch.float32, device=device)
scores = logits[rows_t, toks_t].to(torch.float32)
scores = torch.where(scores > 0, scores / pens_t, scores * pens_t)
logits[rows_t, toks_t] = scores.to(orig_dtype)
```

zhaochenyang20 (changes requested):

sglang_omni/model_runner/base.py
yxs: fp32 math + cast back, regression test added. fp16 diff vs scalar 0.0039 → 0.0019

zhaochenyang20 (approved):

Issue #485, opened 2026-05-19
[Perf][Qwen3-Omni] Replace per-row CPU deque for talker future-text with a device-resident tensor + cursor


Hayden727:
## Problem

During talker decode, future-text embedding rows are injected one row per step and fused with the audio-feedback row before being fed into forward. Today these rows are kept as a Python `collections.deque` of **N independent 1-row CPU tensors**, which forces:

- producer side: one `.detach().cpu()` per row — N small D2H copies;
- consumer side: one `.to(device, dtype)` per decode step — N small H2D copies;
- a Python deque of N independent small objects in between — poor cache locality.

The source is a device-resident 2D tensor, the destination is the GPU, and the queue exists only to bridge the two. We disassemble it row by row, move each row to CPU, then move each row back to GPU. Audio decode often runs for thousands of steps, so the accumulated PCIe dispatch + Python-deque overhead is measurable. The producer-side `.cpu()` is also a **synchronous D2H** sitting on the thinker streaming critical path.

### Producer side: per-row `.cpu()`

`sglang_omni/models/qwen3_omni/components/talker_prefill.py:330-338`

```python
def tensor_rows_to_queue(
    self, tensor: torch.Tensor | None
) -> collections.deque[torch.Tensor]:
    queue: collections.deque[torch.Tensor] = collections.deque()
    if not isinstance(tensor, torch.Tensor) or tensor.numel() == 0:
        return queue
    for row in tensor:
        queue.append(row.detach().cpu())
    return queue
```

Later thinker-chunk / EOS appends (same file at `:259` and `:274`) also go row-by-row through `.cpu()` before being enqueued.

### Consumer side: per-decode-step `.to(device)`

`sglang_omni/models/qwen3_omni/talker_model_runner.py:267-275`

```python
combined = feedback.to(device=device, dtype=dtype).reshape(-1)
next_text = QwenTalkerModelRunner._peek_left(
    getattr(data, "pending_text_queue", None)
)
if next_text is not None:
    combined = combined + next_text.to(
        device=device, dtype=dtype,
    ).reshape(-1)
```

Every decode step triggers a `next_text.to(device, dtype)` — a single small H2D copy of one `[hidden_size]` row.

## Proposed direction

Change the backing form of `pending_text_queue` from "deque of N 1-row CPU tensors" to a **device-resident 2D tensor plus a cursor**:

[...]

PR #494, opened 2026-05-20
[Bugfix] Avoid CPU round-trip for Qwen3 talker text queue

JingwenGu0829:
Benchmark shown below.
Avoiding CPU round-trip for Qwen3 talker text queue.
#485


yxs:
Verified on (n=1088 c=16, `OMNI_PAD_TIMING=1`):

| | pre-fix (main) | post-fix |
|---|---:|---:|
| pad `.to()` avg/call | 150 μs | **6.2 μs** |
| pad `.to()` per req | 5.25 ms | 0.21 ms |
| n50 | 8 | 8 every bucket |
| WER corpus | 2.73 % | 2.78 % (1R noise) |
| latency_mean | 2.252 s (5R) | 2.245 s (1R) |

Sub-noise on bulk `latency_mean` as expected — wins are on critical-path CPU + TTFT/ITL.

sglang_omni/models/qwen3_omni/pending_text_queue.py
```diff
@@ -0,0 +1,104 @@
...
+
+def _as_rows(tensor: torch.Tensor) -> torch.Tensor | None:
+    tensor = tensor.detach()
+    if tensor.numel() == 0:
+        return None
+    if tensor.dim() == 1:
+        return tensor.reshape(1, -1)
+    if tensor.dim() == 2:
+        return tensor
+    return tensor.reshape(-1, tensor.shape[-1])
```
Hayden727: This helper is becoming the small queue contract between builders and the runner, so I would keep its shape semantics strict. Silently flattening any rank >2 means a future talker can accidentally pass `[batch, seq, hidden]` or `[seq, codebook, hidden]` and get interleaved FIFO rows instead of a clear model-side decision. Can we accept only `[hidden]` / `[rows, hidden]` here and require callers to flatten model-specific layouts explicitly?

sglang_omni/models/qwen3_omni/pending_text_queue.py
```diff
@@ -0,0 +1,104 @@
...
+
+        remaining = self.rows[self.cursor :]
+        rows = rows.to(device=remaining.device, dtype=remaining.dtype)
+        self.rows = torch.cat([remaining, rows], dim=0)
+        self.cursor = 0
+
+
+def coerce_pending_text_queue(value: object) -> PendingTextTensorQueue:
+    if isinstance(value, PendingTextTensorQueue):
+        return value
```
Hayden727: Returning an existing `PendingTextTensorQueue` by reference gives callers shared mutable cursor state. The previous builder path copied list/deque inputs into a fresh deque, so two requests could not accidentally consume each other. Since `build_sglang_talker_request` is a reusable request helper, please add a cheap queue `copy()`/`clone()` that can share row storage but owns its own cursor, and use it when coercing an existing queue.

sglang_omni/models/qwen3_omni/request_builders.py
```diff
@@ -424,7 +427,7 @@ def build_sglang_talker_request(
     talker_input_ids: torch.Tensor | list[int] | None = None,
     input_embeds_are_projected: bool = False,
     pending_text_queue: (
-        collections.deque[torch.Tensor] | list[torch.Tensor] | torch.Tensor | None
+        PendingTextTensorQueue | list[torch.Tensor] | torch.Tensor | None
```
Hayden727: The implementation still accepts legacy deque/iterable queues through `coerce_pending_text_queue`, and the tests still cover deque inputs, but the public type now advertises only `list`. To keep this request-builder API clean for existing schedulers and future models, please type this as `Iterable[torch.Tensor] | torch.Tensor | PendingTextTensorQueue | None` (or explicitly keep `collections.deque`) instead of narrowing it to `list`.

Hayden727 (changes requested):
Thanks for moving the hot path off the CPU; the direction matches #485. I think the queue helper needs a little more API hardening before merge, because it is now the contract between request builders, scheduler state, and the talker runner. The perf change should stay, but the public-ish semantics should stay explicit for future talker models.

Hayden727 (approved):
Rechecked e22267f. The latest commit addresses my previous review points: rank handling is now explicit, coercing an existing queue gives an independent cursor, and the request-builder type keeps the iterable/deque-compatible API surface. I only did static review plus py_compile locally; pytest is blocked in this local env because torch is not installed.

Issue #472, opened 2026-05-18
[RFC][Perf] Topology-aware inter-stage transport: local refs for same-process edges, CUDA IPC for same-GPU cross-process edges


Hayden727:
## Motivation

Current `sglang-omni` inter-stage transport is **not topology-aware**. Full-payload edges still go through `relay.write_payload` / `read_payload` unconditionally, and streaming chunks still take the relay path inside `send_stream_chunk` even when compile-time topology already identifies a same-GPU downstream target.

Single-GPU and multi-GPU deployments expose the same missing capability:

- **Single-GPU colocated deployment**: every **GPU→GPU** edge is a same-GPU edge. Sending those tensors through relay is unnecessary overhead. (CPU↔GPU edges are not in this class.)
- **Multi-GPU deployment**: encoder + aggregator + thinker are routinely colocated on the same GPU (see #329 / #430). Those **same-GPU GPU→GPU edges** are not structurally different from the single-GPU case. Cross-GPU edges remain correct to send through relay.

This proposal: **give each edge a lightweight topology classification, with two distinct colocated fast-path tiers and a fallback to existing relay** —

- **`LOCAL_OBJECT`** — producer and consumer share an OS process; hand the Python object / tensor reference through an in-process path. No relay, no CUDA IPC, no serialization.
- **`CUDA_IPC`** — producer and consumer are in different processes but on the same GPU; cross the process boundary with a CUDA IPC handle, avoiding the GPU → CPU → SHM → CPU → GPU round trip that the default SHM relay path forces today.
- **Relay** — cross-GPU / cross-node; unchanged.

The two colocated tiers are different physical phenomena and should not be conflated: `LOCAL_OBJECT` has no kernel involvement and no IPC handle exchange at all; `CUDA_IPC` still crosses a process boundary and exchanges an IPC handle but skips the data round trip.

After #430 / #461, the runtime can now express both GPU colocation and process topology explicitly. However, the current Qwen3 speech-colocated pipeline still assigns distinct process groups per stage, so **same-process direct pass is an available runtime optimization target, not something the default speech-colocated config already benefits from today**. For the current single-GPU Qwen3 speech-colocated config, the immediately realizable gain is the same-GPU path; `LOCAL_OBJECT` matters for explicitly same-process layouts and future colocated configs that choose to group stages into one OS process.

A single mechanism covers both deployment shapes; no single-GPU special-case fast path, and no transport-branch explosion inside `Stage`.

[...]

PR #496, opened 2026-05-20
[Perf] Topology-aware inter-stage data transport optimization

PasserBy4:
## Motivation

This PR adds topology-aware transport selection for inter-stage pipeline data. When stages are colocated in the same process or on the same GPU, the pipeline can avoid the default relay path and use cheaper transport where safe.

## Modifications

- Add `LOCAL_OBJECT` transport for same-process stage edges.
  - Full payloads can use direct process-local dispatch for eligible single-target edges.
  - Fan-out payloads can use process-local dispatch when a target-specific projection creates an isolated payload data container.
  - Stream chunks and stream done/error signals can use process-local dispatch for same-process stream targets.

- Add CUDA IPC transport for same-GPU cross-process stream chunks.
  - CUDA IPC is used only for stream chunks whose data contains CUDA tensors.
  - Full payload CUDA IPC is not included in this PR.
  - If CUDA IPC is not applicable, stream chunks continue to use the relay path.

- Keep relay as the fallback path.
  - Full payloads use relay when `LOCAL_OBJECT` is not eligible.
  - Stream chunks use relay when neither `LOCAL_OBJECT` nor CUDA IPC applies.

- Add Qwen3-Omni payload projections to reduce unnecessary downstream payload data.
  - `thinker -> decode` drops multimodal tensors not needed by decode.
  - `talker_ar -> code2wav` keeps only the request latch because code tensors arrive through streaming.

- Add and update unit tests for:
  - same-process payload and stream local dispatch
  - projected fan-out payload safety
  - same-GPU stream CUDA IPC selection
  - cross-process CUDA IPC stream chunk receive path
  - relay fallback behavior
  - Qwen3-Omni projection contracts

## Notes

- `LOCAL_OBJECT` passes Python object references directly between same-process stages. Receivers should treat payloads, stream data, and metadata as read-only unless the edge explicitly provides an isolated projected object.

- CUDA IPC keeps CUDA tensors off the relay path, but CPU tensors and regular Python values in the same stream object are still serialized into the control message. This path is intended for CUDA-dominant stream chunks with small CPU metadata.

- Issue #472 Step 3, full-payload CUDA IPC, was implemented and instrumented during development, then reverted. On the current Qwen3-Omni paths it did not trigger: it only applies to single-target full-payload edges where both sender and receiver are GPU stages, the stages are in different processes on the same GPU, neither side is a TP stage, and the projected payload data contains CUDA tensors. CPU-stage edges, same-process edges, fan-out payload sends, TP stages, and payloads without CUDA tensors all fall back to other paths. Since this added a separate full-payload topology branch without observed benefit on the target pipeline, the PR keeps only stream CUDA IPC plus `LOCAL_OBJECT`.

## Related Issues

#472 

### SeedTTS EN generation

Qwen3-Omni speech-colocated config (same GPU different process), max concurrency 16, voice clone enabled, 1088 samples.

| Version | Latency mean (s) | Throughput (req/s) | Tok/s agg | RTF mean |
|---|---:|---:|---:|---:|
| Before Step 1 | 3.196 | 4.993 | 4.5 | 0.9209 |
| After Step 1 | 3.006 (-5.9%) | 5.313 (+6.4%) | 4.7 (+4.4%) | 0.8703 (-5.5%) |
| After payload cleanup | 2.862 (-10.5%) | 5.575 (+11.7%) | 5.0 (+11.1%) | 0.8224 (-10.7%) |

### MMSU text only

Qwen3-Omni text-only serving (same GPU same process), max concurrency 32, 2000 samples.

| Version | Latency mean (s) | Throughput (req/s) | Tok/s agg | RTF mean |
|---|---:|---:|---:|---:|
| Before Step 2 | 0.583 | 54.65 | 3.50 | N/A |
| After Step 2 | 0.390 (-33.1%) | 81.57 (+49.3%) | 5.20 (+48.6%) | N/A |

## Checklist

- [ ] Add unit tests.

## CI


sglang_omni/pipeline/local_dispatch.py
```diff
@@ -0,0 +1,81 @@
...
+
+from __future__ import annotations
+
+from typing import Any, Iterable
+
+
+class LocalStageDispatcher:
+    """Dispatch stage objects between stages in the same OS process.
+
+    Process-local dispatch passes Python object references directly. Receivers
```
Hayden727, line 12: This introduces a new same-process transport path, but `docs/developer_reference/communication.md` still says every stage-to-stage payload after the initial submit goes through relay and only documents relay/CUDA IPC lifetime. Please update that developer communication doc to include `LOCAL_OBJECT`, direct Python reference semantics, and the read-only/reference-lifetime caveat; otherwise future framework changes will follow the stale relay-only contract.

zhaochenyang20 (changes requested):
Could you update `docs/design/refactor_rfc.md` also with your new desgin.

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -694,6 +778,42 @@ async def _send_to_stage(self, request_id: str, target: str, payload: Any) -> No
...
+                    "a local dispatcher"
+                )
+
+            await self._local_dispatcher.send_payload(
+                from_stage=self.name,
+                to_stage=target,
+                request_id=request_id,
+                payload=projected_payload,
+            )
+            return
```
zhaochenyang20, line 841: the same-process payload path calls `send_payload` and returns at lines 809-815 without emitting `stage_hop_sent` like the relay path does at lines 826-831; similarly, `receive_local_stream_chunk` routes the `StreamItem` at lines 310-327 without emitting `stage_stream_chunk_received`. `sglang_omni/profiler/views.py` computes hop breakdown by pairing `stage_hop_sent`/`stage_input_received` and `stage_stream_chunk_sent`/`stage_stream_chunk_received`; the text colocated configuration is exactly one of this PR's benchmark target paths, so once local transport is selected, payload hop and stream receive latency disappear from profiler views and the performance report undercounts the true cost of the new transport.

Emit the same `stage_hop_sent` metadata for the local payload path as for the relay path, at minimum including `to_stage`, and optionally adding `transport="local_object"`. In `receive_local_stream_chunk`, emit `stage_stream_chunk_received` after constructing the `StreamItem` and before handing it to the scheduler, with metadata matching `_emit_stream_chunk_received`. Add unit tests for `LOCAL_OBJECT` payload hop events and local stream chunk receive events so the profiler contract is not covered only for relay/IPC paths.

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -712,6 +832,42 @@ async def _send_to_stage(self, request_id: str, target: str, payload: Any) -> No
...
+        original_payload: Any,
+        projected_payload: Any,
+        *,
+        projector_present: bool,
+    ) -> bool:
+        if not projector_present or projected_payload is original_payload:
+            return False
+        if not hasattr(original_payload, "data") or not hasattr(
+            projected_payload, "data"
+        ):
```
zhaochenyang20: Make this gate express the contract explicitly: either import `StagePayload` and use `isinstance(original_payload, StagePayload)` plus `isinstance(projected_payload, StagePayload)`, or define a small typed helper/Protocol and raise a clear error when a projector returns a non-payload object. The `projected_payload.data is not original_payload.data` isolation check can stay, but the type contract before it should be statically readable.

zhaochenyang20 (changes requested):

PasserBy4:
## Updated benchmark and profiler results

I reran the Qwen3-Omni H200 colocated SeedTTS and MMSU text only benchmarks after the latest rebase. The profiler now shows the expected transport-path improvements.

### SeedTTS EN generation

Setup: `Qwen/Qwen3-Omni-30B-A3B-Instruct`, `examples/configs/qwen3_omni_colocated_h200.yaml`, colocated serving, max concurrency 16, voice clone enabled, 1088 completed requests, 0 failed requests.

| Version | Latency mean (s) | Throughput (req/s) | Output throughput (tok/s) | Output tok/req-s | RTF mean |
| Before PR #496 | 3.001 | 5.321 | 75.8 | 4.8 | 0.8635 |
| After PR #496 | 2.715 (-9.5%) | 5.879 (+10.5%) | 83.8 (+10.6%) | 5.2 (+8.3%) | 0.7789 (-9.8%) |

Profiler highlights:

| Interval / hop | Before avg ms | After avg ms | Change |
|---|---:|---:|---:|
| `talker_ar` stage complete | 2028.524 | 1848.411 | -8.9% |
| `code2wav` stage complete | 420.506 | 375.107 | -10.8% |
| `thinker` stage complete | 331.811 | 308.410 | -7.1% |
| `talker_ar -> code2wav` stream chunk | 7.115 | 4.871 | -31.5% |
| `thinker -> talker_ar` stream chunk | 5.055 | 2.892 | -42.8% |
| `talker_ar -> code2wav` payload | 25.149 | 5.079 | -79.8% |
| `thinker -> decode` payload | 24.408 | 0.753 | -96.9% |


### MMSU text only

Setup: Qwen3-Omni text-only serving, `zhaochenyang20/mmsu-ci-2000`, max samples 2000, max concurrency 32.

| Version | Latency mean (s) | Throughput (req/s) | Tok/s agg | RTF mean |
|---|---:|---:|---:|---:|
| Before PR #496 | 0.583 | 54.65 | 3.50 | N/A |
| After PR #496 | 0.390 (-33.1%) | 81.57 (+49.3%) | 5.20 (+48.6%) | N/A |


Profiler highlights:

| Interval / hop | Before avg ms | After avg ms | Change |
|---|---:|---:|---:|
| `thinker` stage complete | 289.019 | 193.530 | -33.0% |
| `mm_aggregate` stage complete | 220.717 | 158.647 | -28.1% |
| `audio_encoder` stage complete | 207.183 | 157.667 | -23.9% |
| `decode` stage complete | 7.697 | 1.145 | -85.1% |
| `preprocessing -> audio_encoder` payload | 5.928 | 0.035 | -99.4% |
| `preprocessing -> mm_aggregate` payload | 5.559 | 0.028 | -99.5% |
| `audio_encoder -> mm_aggregate` payload | 5.211 | 0.025 | -99.5% |
| `mm_aggregate -> thinker` payload | 5.576 | 0.024 | -99.6% |
| `thinker -> decode` payload | 4.872 | 0.034 | -99.3% |

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -712,6 +858,48 @@ async def _send_to_stage(self, request_id: str, target: str, payload: Any) -> No
...
+            )
+        if not isinstance(projected_payload, StagePayload):
+            raise TypeError(
+                "projected local-object dispatch requires projectors to return "
+                f"StagePayload, got {type(projected_payload).__name__}"
+            )
+        # A fan-out edge may use process-local dispatch only when projection
+        # gives the target its own mutable payload/data container. Tensor leaves
+        # inside that container may still be shared intentionally.
+        return projected_payload.data is not original_payload.data
```
Ratish1: This isolation check is too shallow for a framework-level transport decision. It only verifies that `projected_payload.data` is not the same top-level object as `original_payload.data`, so a projector that returns `{"projected": payload.data}` or a shallow `dict(payload.data)` is treated as safe even though downstream branches still share nested mutable dict/list state. That changes relay semantics: the old path serialized each fan-out edge independently, while this path lets a same-process receiver mutate data that another receiver may later read or that the sender may still serialize for a relay-backed sibling edge.

so,  `PipelineState.from_dict()` preserves nested dict references, `project_thinker_to_decode()` returns `state.to_dict()` without deep-copying all mutable subcontainers, and `decode_events()` mutates `stream_state`. The current unit projector used by `test_stage_projected_fan_out_payloads_use_local_object_when_isolated` also wraps the original `payload.data` under a new dict, which passes this predicate while still aliasing the original payload.

Can we either keep projected fan-out on the relay path unless a projector explicitly opts into a framework-provided "copy containers, share tensors" helper, or replace this identity check with a recursive mutable-container alias detector?

sglang_omni/pipeline/relay_io.py
```diff
@@ -291,7 +321,23 @@ async def send_stream_chunk(
     same_gpu_targets: set[str] | None = None,
 ) -> None:
     """Send a streaming chunk to a downstream stage."""
-    del same_gpu_targets
+    # CUDA IPC keeps CUDA tensors off the relay path, but CPU tensors and plain
+    # Python values in this object are still pickled into the control message.
+    # Keep this path for CUDA-dominant chunks with small CPU metadata.
+    if (
```
Ratish1, line 387: This switches to the IPC/control-plane path whenever `data` contains any CUDA tensor, but `serialize_ipc_chunk(data, metadata)` pickles the whole object graph into `DataReadyMessage.shm_metadata`. That means any large CPU tensor/list/string inside `data`, and any CPU tensor in `metadata`, now travels through the ZMQ control plane instead of the relay data plane. The comment says this path should be for CUDA-dominant chunks with small CPU metadata, but the framework does not enforce that invariant.

Should we make the transport selector reject or split mixed CPU-heavy objects. A safe shape would be: use IPC only for CUDA tensor data plus scalar/small metadata, keep metadata tensors on relay, or fall back to the relay path when CPU tensors/large Python containers are present. can add unit test for this aswell.

Ratish1 (commented):
Great work @PasserBy4 . Left only few comments.

sglang_omni/pipeline/stage/runtime.py
Hayden727: Confirmed on current head `bc039fb` in the H200 container: `project_thinker_to_decode()` creates a fresh top-level `data` dict, but the projected payload still aliases the nested mutable decode state:

```text
data_shared False
stream_state_shared True
token_ids_shared True
```

Since `decode_events()` mutates `stream_state`, the Qwen3 same-process fan-out can leak decode mutations back into the original/talker branch. So the fix needs to isolate mutable containers recursively (or keep this fan-out on relay) rather than relying on top-level `data` identity only.

sglang_omni/pipeline/relay_io.py
Hayden727: Also confirmed on current head that this selector is still purely `same_gpu_targets && target_stage in same_gpu_targets && _contains_cuda_tensor(data)`, and `serialize_ipc_chunk()` then stores `ipc_pickle(data)` directly in `shm_metadata["tensor_bytes"]`. So a single CUDA tensor leaf is enough to move the entire `data` graph through the control message. A regression case with a small CUDA tensor plus a large CPU tensor/list would make this contract explicit.

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -712,6 +858,102 @@ async def _send_to_stage(self, request_id: str, target: str, payload: Any) -> No
...
+        return not Stage._shares_mutable_container(
+            original_payload.data, projected_payload.data
+        )
+
+    @staticmethod
+    def _shares_mutable_container(original: Any, projected: Any) -> bool:
+        original_ids = Stage._collect_mutable_container_ids(original)
+        if not original_ids:
+            return False
+        return Stage._contains_mutable_container_id(projected, original_ids)
```
zhaochenyang20, line 894: line 883 only checks `projected_payload.data is not original_payload.data`, which is not enough to preserve the per-edge container isolation that relay fan-out used to provide. A projector can return `StagePayload(data={"projected": payload.data})` or a shallow `dict(payload.data)`, and the current predicate still returns true; in the verified checkout I called `_is_isolated_projected_payload()` with that shape and got true while `projected.data["projected"] is original.data` was also true. The current Qwen path has a real alias as well: `project_thinker_to_decode()` returns `state.to_dict()` after `PipelineState.from_dict()`, and `PipelineState.from_dict()` plus `to_dict()` preserve nested dict references such as `stream_state`; the local probe showed the projected top-level data differs, but `projected.data["stream_state"] is payload.data["stream_state"]` is true. The decode side's `decode_events()` mutates `stream_state` in place, so once same-process fan-out selects the local path, receiver-thread mutation of nested state can race with the sender serializing the original payload for a relay-backed sibling edge, or can affect what another local sibling sees. This is a framework transport semantics bug, not just projector style.

Do not base projected fan-out local eligibility on shallow identity. The most conservative fix is to keep projected fan-out on relay by default and let a projector opt in only through a framework helper; that helper should recursively copy mutable containers such as dict/list/tuple/set, preserve tensor leaves and immutable leaves, and return a marked or typed isolated payload. Another acceptable fix is to add recursive mutable-container alias detection in `_is_isolated_projected_payload()` and fall back to relay whenever any shared dict/list/set or shared non-tensor mutable object is found. Whichever path is chosen, add tests for `{"projected": payload.data}`, shallow `dict(payload.data)`, Qwen `stream_state` aliasing, and intentional tensor leaf sharing.

sglang_omni/pipeline/relay_io.py
```diff
@@ -250,14 +250,98 @@ async def read_blob(
...
+    inline_size = _inline_cpu_pickle_size(data) + _inline_cpu_pickle_size(metadata)
+    return inline_size <= _IPC_INLINE_CPU_BYTES_LIMIT
+
 
 def ipc_pickle(obj: Any) -> bytes:
-    """Serialize via ForkingPickler for CUDA IPC tensor handles."""
+    """Serialize via ForkingPickler only when CUDA IPC tensor handles are needed."""
+    if not _contains_cuda_tensor(obj):
+        return pickle.dumps(obj)
     buf = io.BytesIO()
```
zhaochenyang20, line 328: select the IPC/control-plane path whenever same-GPU target `data` contains any CUDA tensor, but lines 285-293 then put the whole `data` object graph and the whole recursive `metadata` into `DataReadyMessage.shm_metadata`. `ipc_pickle()` falls back to `pickle.dumps()` for values without CUDA tensors, so a large CPU tensor, large list/string inside `data`, or a CPU tensor inside `metadata` becomes bytes inside a msgpack/ZMQ control message instead of using the existing relay blob and metadata tensor extraction at lines 342-363. That selector does not enforce the comment's "CUDA-dominant chunks with small CPU metadata" invariant; once a later stream adds CPU features, debug payloads, text buffers, or stats arrays, the risk is that the control plane starts carrying data-plane traffic and can block the event loop, inflate ZMQ messages, and add latency jitter.

 Add an explicit IPC eligibility check before selecting IPC instead of only checking whether `data` contains a CUDA tensor. The safe shape should be limited to CUDA tensor leaves plus small scalars, short strings, and small metadata; if a CPU tensor, large bytes/string/list/dict, or unknown large object is present, fall back to relay, or split CUDA tensors through IPC while keeping CPU tensor metadata on the existing `extract_tensors()` plus `write_blob()` path. At minimum, add a test where `data` contains a CUDA tensor but metadata contains a CPU tensor or large Python container, and assert that the sender does not produce a control message containing only `_ipc`.

tests/unit_test/pipeline/test_stage.py
```diff
@@ -324,3 +325,543 @@ async def _run() -> None:
...
+
+def test_stage_projected_fan_out_payloads_use_local_object_when_isolated() -> None:
+    def _isolated_projector(marker):
+        def _project(payload):
+            return make_stage_payload(
+                request_id=payload.request_id,
+                inputs=payload.request.inputs,
+                params=payload.request.params,
+                data={"marker": marker, "data": dict(payload.data)},
+            )
```
zhaochenyang20, line 512: Derived from the root-cause issues above, I would require additional tests because the new coverage mostly proves the local/IPC happy paths and does not lock the two selector shapes that must be rejected. `test_stage_projected_fan_out_payloads_use_local_object_when_isolated()` uses `make_noop_projector()` to wrap the original payload data under a new dict, which is one form of the nested-alias risk shape, but the test only asserts that relay was not used; `test_stage_projected_fan_out_requires_isolated_data_container()` covers only the exact same top-level `.data` object. On the stream IPC side, `test_send_stream_chunk_uses_ipc_for_detected_cuda_same_gpu_chunk()` monkeypatches a CPU tensor into the CUDA case and still only asserts that control-plane `_ipc` is used, with no negative test ensuring CPU-heavy metadata fallback. This coverage gap already let both selector bugs into the PR, so I would keep it as a process finding.


Add selector negative tests in the pipeline tests: fan-out projectors returning `{"projected": payload.data}`, shallow `dict(payload.data)`, and Qwen `project_thinker_to_decode()` sharing `stream_state` should fall back to relay or fail explicitly; CUDA IPC tests should add CPU tensor metadata, large Python containers, and mixed CUDA+CPU object graphs, and assert that these shapes are not placed into a pure `_ipc` control message. Keep positive coverage for tensor leaf sharing and small scalar metadata so the fix does not disable the intended fast paths.

zhaochenyang20 (dismissed):
change `tests/readme.md` also

yxs (approved):

Issue #484, opened 2026-05-19
[Perf][Qwen3-Omni] Drop bulk .cpu().tolist() embedding conversions from the talker request-build path


Hayden727:
## Motivation

In `sglang-omni`, the talker stage's request-build path currently demotes an embedding tensor that is already GPU-resident and ready to feed the backend into a Python `list[list[float]]`. This demotion is a **critical-path tax on talker TTFT**, for two reasons:

1. It happens before talker enters prefill, so **every request pays it once**.
2. It simultaneously incurs four costs: a GPU→CPU copy, a massive Python-object allocation burst, transient memory inflation, and an immediately following CPU→GPU round-trip when upstream SGLang converts it back into a tensor.

In other words, we perform two fully reversed tensor↔list conversions back-to-back, with nothing in between consuming the list form for its own sake.

### 1. Two tensor → list conversions on the talker request-build path

`sglang_omni/models/qwen3_omni/request_builders.py:467-481`

```python
if talker_input_embeds is not None:
    input_embeds = talker_input_embeds.float().cpu().tolist()
    input_ids_tensor = torch.as_tensor(talker_input_ids, dtype=torch.long)
    input_ids_list = input_ids_tensor.tolist()
    seq_len = len(input_ids_list)
else:
    # thinker_hidden_states: [seq_len, thinker_hidden_size]
    seq_len = thinker_hidden_states.shape[0]
    input_ids_list = [codec_bos_id] * seq_len
    input_ids_tensor = torch.tensor(input_ids_list, dtype=torch.long)
    # Convert hidden states to list-of-lists for Req.input_embeds
    input_embeds = thinker_hidden_states.float().cpu().tolist()
```

The result is written directly into `Req.input_embeds`:

`sglang_omni/models/qwen3_omni/request_builders.py:497-505`

```python
req = Req(
    rid=rid,
    origin_input_text="",
    origin_input_ids=input_ids_list,
    sampling_params=sampling_params,
    input_embeds=input_embeds,
    eos_token_ids={int(codec_eos_id)} if codec_eos_id is not None else None,
    vocab_size=codec_vocab_size,
)
```

[...]

PR #497, opened 2026-05-20
[Bugfix] Drop gpu->list->gpu round-trip from qwen3-omni talker request-build

SandyLuXY:
## Motivation
This PR is for #484. The old path allocates ~8M Python floats at seq_len=4096 with hidden_dim=2048 via `.cpu().tolist()`, then immediately converts them back to a GPU tensor via `torch.as_tensor`. This is wasted work when the model runner already has access to the original tensor.

## Modifications
Drop the GPU->list->GPU round-trip from the Qwen3-Omni talker request-build path. The prefill embedding tensor is now carried on `SGLangARRequestData.prefill_input_embeds` so the model runner reads it directly, skipping the `.cpu().tolist()` materialization and the subsequent `torch.as_tensor` reconversion.

**`request_builders.py`**
- Removed `.cpu().tolist()` on the projected-embeds path -- `Req.input_embeds` is set to `None` when `input_embeds_are_projected=True`
- Stored the original GPU tensor on `data.prefill_input_embeds` (projected path only)

**`talker_model_runner.py`**
- Primary path in `_run_projected_prefill_forward`: reads tensor from `data.prefill_input_embeds`, slices by `prefix_indices`
- Fallback: reads from `Req.input_embeds` (list path, for non-projected embeds)
- `post_prefill` frees the tensor reference to release GPU memory

**`request_data.py`**
- Added `prefill_input_embeds: torch.Tensor | None` field to `SGLangARRequestData`

**`test_talker.py`** -- 7 new tests:
- Tensor storage for projected and hidden-states paths
- Model runner reads tensor, not list
- Prefix-indices slicing on tensor path
- Full prefix hit returns None
- GPU memory freed after prefill
- Wall-clock microbenchmark (parametrized by seq_len)

## Related Issues

#484 

## Benchmark & Profiling

Run with 1x H200.

### Wall-Clock: `build_sglang_talker_request` (CPU, no GPU)

Measures the time to execute `build_sglang_talker_request()` end-to-end. The function converts thinker hidden states into a talker request object. Cost is dominated by `.cpu().tolist()` (copying floats from tensor into nested Python lists).

**Method:** 20 iterations, first 5 dropped as warmup. Input: `torch.randn(seq_len, 2048)` passed as `thinker_hidden_states`.

| seq_len | Main (ms) | Feature (ms) | Delta |
|---------|-----------|-------------|-------|
| 256 | 19.70 | 19.50 | -1.0% |
| 2048 | 236.96 | 238.39 | +0.6% |
| 4096 | 475.07 | 482.75 | +1.6% |

Build cost is unchanged since both branches still do `.cpu().tolist()` for `Req.input_embeds`. The PR's savings happen later — during model runner prefill where the tensor is read directly from `prefill_input_embeds`, avoiding the list→tensor reconversion.

### TTFT + ITL (colocated, 1 GPU)

Server topology: `sgl-omni serve --colocate` on a single H200 GPU. Each request gets a unique prompt prefix (`[N] ...`) to bust SGLang's prefix cache — no cache hits between requests.

max_tokens=2048, n=20, warmup=3, concurrency=16:

| Metric | Main | Feature | Delta |
|--------|------|---------|-------|
| TTFT mean | 35.790s | 35.892s | +0.3% |
| TTFT median | 39.801s | 40.056s | +0.6% |
| TTFT p95 | 40.011s | 40.267s | +0.6% |
| TTFT p99 | 40.011s | 40.267s | +0.6% |
| ITL mean | 341.1ms | 345.6ms | +1.3% |
| ITL median | 370.5ms | 382.0ms | +3.1% |
| ITL p95 | 473.1ms | 484.9ms | +2.5% |
| ITL p99 | 644.0ms | 632.9ms | -1.7% |
| Total mean | 175.327s | 177.233s | +1.1% |

**Conclusion:** No meaningful regression. All differences are within run-to-run noise. The end-to-end latency is dominated by model inference, not request building. The PR avoids one tensor→list→tensor round-trip during prefill and drops a wasted `.float()` upcast.

## Checklist

- [x] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/models/qwen3_omni/request_builders.py
+++ b/sglang_omni/models/qwen3_omni/request_builders.py
@@ -553,6 +553,7 @@ def build_sglang_talker_request(
         vocab_size=codec_vocab_size,
     )
     req.tokenizer = tokenizer
+    req._input_embeds_are_projected = bool(input_embeds_are_projected)
     req.omni_model_inputs = dict(talker_model_inputs or {})
     req._omni_consumed = None
     req._codec_suppress_tokens = (
--- a/sglang_omni/models/qwen3_omni/request_builders.py
+++ b/sglang_omni/models/qwen3_omni/request_builders.py
@@ -493,8 +493,8 @@ def build_sglang_talker_request(
 ) -> "SGLangARRequestData":
     """Build SGLang AR request for the Talker from thinker hidden states.
 
-    Stores thinker hidden states as Req.input_embeds so SGLang's pipeline
-    passes them through ForwardBatch.input_embeds -> model.forward(input_embeds=...).
+    Keeps embedding tensors GPU-resident on SGLangARRequestData.prefill_input_embeds
+    so the model runner can read them directly without a list round-trip.
     Uses dummy input_ids of matching length for position tracking, while the
     request data keeps a device-backed FIFO of future text rows for decode.
 
@@ -509,20 +509,15 @@ def build_sglang_talker_request(
     # SGLangARRequestData already imported at module level
 
     if talker_input_embeds is not None:
-        input_embeds = talker_input_embeds.float().cpu().tolist()
+        prefill_embeds_tensor = talker_input_embeds
         input_ids_tensor = torch.as_tensor(talker_input_ids, dtype=torch.long)
         input_ids_list = input_ids_tensor.tolist()
         seq_len = len(input_ids_list)
     else:
-        # thinker_hidden_states: [seq_len, thinker_hidden_size]
         seq_len = thinker_hidden_states.shape[0]
-
-        # Dummy input_ids — codec BOS token repeated for each position
         input_ids_list = [codec_bos_id] * seq_len
         input_ids_tensor = torch.tensor(input_ids_list, dtype=torch.long)
-
-        # Convert hidden states to list-of-lists for Req.input_embeds
-        input_embeds = thinker_hidden_states.float().cpu().tolist()
+        prefill_embeds_tensor = thinker_hidden_states
 
     sampling_params = SamplingParams(
         max_new_tokens=max_new_tokens,
@@ -543,7 +538,7 @@ def build_sglang_talker_request(
         origin_input_text="",
         origin_input_ids=input_ids_list,
         sampling_params=sampling_params,
-        input_embeds=input_embeds,
+        input_embeds=None,
         eos_token_ids={int(codec_eos_id)} if codec_eos_id is not None else None,
         vocab_size=codec_vocab_size,
     )
@@ -591,6 +586,7 @@ def build_sglang_talker_request(
         temperature=temperature,
         output_ids=req.output_ids,
         req=req,
+        prefill_input_embeds=prefill_embeds_tensor,
     )
     data.suppress_tokens = list(req._codec_suppress_tokens or [])
     data.talker_model_inputs = dict(talker_model_inputs or {})
--- a/sglang_omni/models/qwen3_omni/request_builders.py
+++ b/sglang_omni/models/qwen3_omni/request_builders.py
@@ -547,8 +547,7 @@ def build_sglang_talker_request(
         sampling_params=sampling_params,
         # Convert hidden states to list-of-lists for Req.input_embeds
         input_embeds=(
-            None if input_embeds_are_projected
-            else prefill_embeds_tensor.cpu().tolist()
+            None if input_embeds_are_projected else prefill_embeds_tensor.cpu().tolist()
         ),
         eos_token_ids={int(codec_eos_id)} if codec_eos_id is not None else None,
         vocab_size=codec_vocab_size,
@@ -597,7 +596,9 @@ def build_sglang_talker_request(
         temperature=temperature,
         output_ids=req.output_ids,
         req=req,
-        prefill_input_embeds=prefill_embeds_tensor if input_embeds_are_projected else None,
+        prefill_input_embeds=(
+            prefill_embeds_tensor if input_embeds_are_projected else None
+        ),
     )
     data.suppress_tokens = list(req._codec_suppress_tokens or [])
     data.talker_model_inputs = dict(talker_model_inputs or {})
--- a/sglang_omni/models/qwen3_omni/request_builders.py
+++ b/sglang_omni/models/qwen3_omni/request_builders.py
@@ -546,7 +546,10 @@ def build_sglang_talker_request(
         origin_input_ids=input_ids_list,
         sampling_params=sampling_params,
         # Convert hidden states to list-of-lists for Req.input_embeds
-        input_embeds=prefill_embeds_tensor.cpu().tolist(),
+        input_embeds=(
+            None if input_embeds_are_projected
+            else prefill_embeds_tensor.cpu().tolist()
+        ),
         eos_token_ids={int(codec_eos_id)} if codec_eos_id is not None else None,
         vocab_size=codec_vocab_size,
     )
@@ -594,7 +597,7 @@ def build_sglang_talker_request(
         temperature=temperature,
         output_ids=req.output_ids,
         req=req,
-        prefill_input_embeds=prefill_embeds_tensor,
+        prefill_input_embeds=prefill_embeds_tensor if input_embeds_are_projected else None,
     )
     data.suppress_tokens = list(req._codec_suppress_tokens or [])
     data.talker_model_inputs = dict(talker_model_inputs or {})
--- a/sglang_omni/models/qwen3_omni/request_builders.py
+++ b/sglang_omni/models/qwen3_omni/request_builders.py
@@ -493,13 +493,12 @@ def build_sglang_talker_request(
 ) -> "SGLangARRequestData":
     """Build SGLang AR request for the Talker from thinker hidden states.
 
-    Stores thinker hidden states as Req.input_embeds so SGLang's pipeline
-    passes them through ForwardBatch.input_embeds -> model.forward(input_embeds=...).
     Uses dummy input_ids of matching length for position tracking, while the
     request data keeps a device-backed FIFO of future text rows for decode.
 
-    Also stores the original tensor on SGLangARRequestData.prefill_input_embeds
-    so the model runner can skip the list→tensor reconversion during prefill.
+    Stores the original tensor on SGLangARRequestData.prefill_input_embeds
+    when input_embeds_are_projected, so the model runner can skip the
+    list→tensor reconversion during prefill.
 
     Args:
         thinker_hidden_states: Embed layer hidden states [seq_len, hidden_size].
--- a/sglang_omni/models/qwen3_omni/request_builders.py
+++ b/sglang_omni/models/qwen3_omni/request_builders.py
@@ -493,10 +493,9 @@ def build_sglang_talker_request(
 ) -> "SGLangARRequestData":
     """Build SGLang AR request for the Talker from thinker hidden states.
 
-    Keeps embedding tensors GPU-resident on SGLangARRequestData.prefill_input_embeds
-    so the model runner can read them directly without a list round-trip.
-    Uses dummy input_ids of matching length for position tracking, while the
-    request data keeps a device-backed FIFO of future text rows for decode.
+    Populates Req.input_embeds (list-of-lists) to honour the upstream contract,
+    and also stores the original tensor on SGLangARRequestData.prefill_input_embeds
+    so the model runner can skip the list→tensor reconversion.
 
     Args:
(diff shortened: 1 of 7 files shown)


sglang_omni/models/qwen3_omni/talker_model_runner.py
```diff
@@ -170,20 +173,28 @@ def _run_projected_prefill_forward(
...
-            rows = []
+            parts: list[torch.Tensor] = []
             for sched_req in requests:
-                req = sched_req.data.req
-                embeds = req.input_embeds
-                if embeds:
-                    prefix_len = len(req.prefix_indices)
-                    rows.extend(embeds[prefix_len:])
-            if not rows:
+                tensor = sched_req.data.prefill_input_embeds
```
Hayden727: `prefill_input_embeds` is populated for both the projected path and the public `thinker_hidden_states` path, but this runner only reads it when some request has `input_embeds_are_projected=True`. In the latest commit the unprojected path is saved by the fallback `Req.input_embeds` list, so it is no longer a dropped-embeds correctness bug; however, the new field contract is ambiguous and the tests assert storage without proving the runner consumes it. Please either make the tensor path trigger on `prefill_input_embeds is not None` (using `input_embeds_are_projected` only as the model-side projection flag), or only populate `prefill_input_embeds` for the projected path that this runner actually consumes.

sglang_omni/models/qwen3_omni/talker_model_runner.py
```diff
@@ -170,20 +173,28 @@ def _run_projected_prefill_forward(
...
+                if tensor is not None:
+                    parts.append(tensor)
```
Hayden727: This appends the whole stored tensor, but the old upstream `Req.input_embeds` path sliced by the active extend range: `prefix_len = len(req.prefix_indices)` and then `input_embeds[prefix_len : prefix_len + req.extend_input_len]`. Without the same slice here, any prefix hit, chunk/retry, or future scheduler mode can make `input_embeds` longer than `forward_batch.input_ids` and feed rows for already-cached tokens again. Even if today's Qwen talker usually disables radix/chunking, this should preserve the generic prefill contract and have a test with nonzero `prefix_indices` / smaller `extend_input_len`.

Hayden727 (changes requested):
Thanks for taking on #484; keeping the talker request-build path tensor-native is the right direction. I found two framework-level blockers in the new tensor handoff: one breaks the public unprojected `thinker_hidden_states` path, and the other drops the old prefix/extend slicing semantics that upstream `Req.input_embeds` gave us for free. Details inline.

sglang_omni/models/qwen3_omni/request_builders.py
```diff
@@ -543,7 +532,7 @@ def build_sglang_talker_request(
         origin_input_text="",
         origin_input_ids=input_ids_list,
         sampling_params=sampling_params,
-        input_embeds=input_embeds,
+        input_embeds=prefill_embeds_tensor.cpu().tolist(),
         eos_token_ids={int(codec_eos_id)} if codec_eos_id is not None else None,
```
Hayden727, line 551: This brings the critical-path `.cpu().tolist()` back into `build_sglang_talker_request`, so the request still pays the large D2H copy and Python-float allocation that #484 is trying to remove. Keeping `prefill_input_embeds` for the runner avoids the later list→tensor reconversion, but it does not solve the request-build wall-clock/GC pressure from the issue. If the upstream `Req` list contract must remain as a fallback, please avoid materializing it on the hot projected path (or make it lazy/conditional) and add a test that fails if `talker_input_embeds.cpu().tolist()` is called during projected request build.

sglang_omni/models/qwen3_omni/talker_model_runner.py
```diff
@@ -170,20 +173,32 @@ def _run_projected_prefill_forward(
...
-                embeds = req.input_embeds
-                if embeds:
-                    prefix_len = len(req.prefix_indices)
-                    rows.extend(embeds[prefix_len:])
-            if not rows:
+                tensor = sched_req.data.prefill_input_embeds
+                if tensor is not None:
+                    prefix_len = len(sched_req.data.req.prefix_indices)
+                    if prefix_len > 0:
+                        tensor = tensor[prefix_len:]
```
Hayden727: This fixes the prefix-offset part, but it still does not preserve the full upstream prefill slice contract. The list path should effectively use `input_embeds[pre_len : pre_len + req.extend_input_len]`; this tensor path uses `tensor[prefix_len:]`, so if a future chunk/retry/scheduler path sets `extend_input_len < tensor.shape[0] - prefix_len`, the runner will pass extra rows that do not match `forward_batch.input_ids`. I reproduced that on this head with `prefix_len=3`, `extend_input_len=4`, and a 10-row tensor: the method forwarded shape `(7, hidden)` instead of `(4, hidden)`. Please slice to `tensor[prefix_len : prefix_len + int(req.extend_input_len)]` and add a unit test where `extend_input_len` is smaller than the remaining tensor length.

Hayden727 (changes requested):
Re-reviewed latest head 3016263. The projected request-build hot path no longer materializes Req.input_embeds and the field contract is much cleaner now. One framework-level blocker remains around preserving the scheduler prefill slice contract: the tensor path still slices only by prefix length, not by extend_input_len.

Hayden727 (approved):
Re-verified latest head e18dbb0. The remaining prefill slice blocker is fixed: both tensor-backed embeds and the list fallback now use the same prefix + extend_input_len window, and tests cover the smaller-extend case. Verified in the sglang-omni-hayden container on ion8-h200-omni: test_talker.py passes with 26 tests.

Hayden727 (dismissed):
Re-verified final head 59aae73 after the additional Req marker fix. The projected request keeps the tensor-native hot path, Req._input_embeds_are_projected is preserved for full-prefill scheduling, and tensor/list fallback prefill slices both use the prefix + extend_input_len window. Verified in sglang-omni-hayden on ion8-h200-omni: py_compile passed, git diff --check passed, and tests/unit_test/qwen3_omni/test_talker.py passes with 26 tests.

sglang_omni/models/qwen3_omni/talker_model_runner.py
```diff
@@ -170,20 +172,31 @@ def _run_projected_prefill_forward(
...
-                embeds = req.input_embeds
-                if embeds:
-                    prefix_len = len(req.prefix_indices)
-                    rows.extend(embeds[prefix_len:])
-            if not rows:
+                prefix_len = len(req.prefix_indices)
+                extend_len = int(req.extend_input_len)
+                tensor = sched_req.data.prefill_input_embeds
+                if tensor is not None:
+                    tensor = tensor[prefix_len : prefix_len + extend_len]
```
Hayden727: `Req.init_next_round_input()` rebuilds a retracted decode request as `origin_input_ids + output_ids`, so the next extend window can cross past the original prompt into already-generated codec tokens. This slice only reads `prefill_input_embeds`, which has prompt length. For example, prompt len 10 + 3 generated tokens with a prefix hit of 8 gives `extend_input_len == 5`, but `prefill_input_embeds[8:13]` returns only 2 rows. The runner then forwards fewer embed rows than `forward_batch.input_ids` (or rebuilds the decoded-token KV with the wrong inputs). The new retract test only covers a no-output re-prefill. Please either retain/reconstruct the per-step decode input embeds for the generated-token suffix, or prevent projected talker requests from entering decode retract until that contract is supported.

Hayden727 (dismissed):
I rechecked the latest retract fix. Keeping the prompt tensor alive until the request leaves the scheduler is acceptable from a lifetime/GC perspective, but the current implementation still does not satisfy the full decode-retract contract once the request has generated output tokens. See the inline comment.

yxs:
per 1c548da68b548225c501df75f2a213528d755d87.

Trade-off note: `prefill_input_embeds` now stays alive until the Req finishes (≈ 4-8 MB per req, < 5% of the talker KV cache pool). This pays for a retract-path contract fix — projected requests have `Req.input_embeds=None`, so releasing the tensor in `post_prefill` causes the second prefill (after sglang's `decode.py` retract requeues the Req) to fall through to a standard forward with dummy `codec_bos` ids and produce silent audio corruption. The new `test_projected_prefill_survives_decode_retract` locks this contract.

sglang_omni/models/qwen3_omni/talker_model_runner.py
Hayden727: Fixed in 006dac7. I added regression coverage for the decode-retract case with existing output_ids: first verified the new tests failed on 1c548da because the runner only replayed the prompt suffix, then fixed the runner to record consumed decode input embeds and replay the generated-token suffix during re-prefill. Verified on ion8-h200-omni / sglang-omni-hayden: the targeted retract tests pass, full tests/unit_test/qwen3_omni/test_talker.py passes with 29 tests, py_compile passes, and git diff --check passes.

Hayden727:
Agree with this trade-off. Keeping `prefill_input_embeds` alive until the Req leaves the scheduler is the right fix for the first retract contract issue: projected requests have `Req.input_embeds=None`, so clearing the tensor in `post_prefill` can make a later re-prefill fall back to the dummy codec input ids path.

One nuance: the memory estimate is `seq_len * hidden_size * dtype_size` per request, so for bf16/fp16 it is `seq_len * hidden_size * 2` bytes. The exact number depends on prompt length (e.g. 4096 x 2048 x 2 ~= 16 MiB), but this is still reasonable compared with the talker KV pool and is the right trade-off for correctness.

The latest fix in `006dac7` also covers the second part of the retract contract. On real decode retract, `Req.init_next_round_input()` rebuilds `fill_ids` as `origin_input_ids + output_ids`, so the re-prefill window may include both the prompt suffix and already-generated codec tokens. Retaining only `prefill_input_embeds` fixes the no-output re-prefill case, but not the generated-token suffix case. `006dac7` records the decode input embeds that were actually consumed by the feedback buffer and replays them when re-prefill crosses past the original prompt.

yxs (approved):

sglang_omni/models/qwen3_omni/talker_model_runner.py
```diff
@@ -259,6 +392,18 @@ def _peek_left(queue: Any) -> torch.Tensor | None:
...
+    def _decode_input_history(data: Any) -> list[torch.Tensor]:
+        history = getattr(data, "decode_input_embeds", None)
+        if history is None:
+            history = []
+            data.decode_input_embeds = history
+        return history
+
+    @staticmethod
+    def _append_decode_input_history(data: Any, row: torch.Tensor) -> None:
+        QwenTalkerModelRunner._decode_input_history(data).append(row.detach().clone())
```
Ratish1: This stores a cloned GPU tensor for every generated token, but `_take_next_decode_input_embed()` already returns a new tensor from `feedback + next_text`. After that, the code only appends it to history and stacks it into the feedback buffer; it does not mutate the saved row.

Should we store `row.detach()` here to avoid the extra per-token allocation/copy.

sglang_omni/models/qwen3_omni/talker_model_runner.py
Ratish1: Here we treat `forward_batch.input_embeds` as if it represents the whole batch whenever it is non-`None`. That is not true for a mixed projected/list-backed batch: the list-backed rows are present there, but the projected rows live on request data, so this path can skip the projected rows and break row alignment.

Can we assert that projected and list-backed requests cannot be batched together, or assemble embeddings request-by-request when projected request data is present.

Ratish1 (commented):
Overall, LGTM @SandyLuXY . Just a few comments, let me know what do you think.

Ratish1:
Also here @SandyLuXY , https://github.com/SandyLuXY/sglang-omni/blob/e24a1447ef95ba7064385f54b55f9db1ab395e6b/sglang_omni/scheduling/omni_scheduler.py#L136

`OmniScheduler` sets the other request limits here, but it never sets `max_queued_requests`. The retract path can put a request back into the waiting queue, and the queue-limit check reads this field before the request reaches the replay logic.

Should we set `self.max_queued_requests = server_args.max_queued_requests` next to `max_prefill_tokens` and `max_running_requests`?

cc: @yxs @Hayden727

Hayden727:
Addressed Ratish’s comments in 82a9a15:

- Mirrored `server_args.max_queued_requests` into `OmniScheduler`; added a regression for the delegated upstream `_abort_on_queued_limit()` path.
- Projected talker prefill now prefers request data over partial `ForwardBatch.input_embeds`, and explicitly rejects mixed projected/unprojected prefill batches; added targeted tests.
- Decode input history now stores `row.detach()` instead of `row.detach().clone()`.

PR #499, opened 2026-05-20
docs: clarify installation options and fix misleading comment

xinlij:
## Motivation
The current installation guide does not clearly indicate that "Installation" and "Use Docker" are two alternative methods. This caused confusion for new users who may attempt to follow both sections.

## Modifications
- Add a sentence at the top clarifying that the two sections are alternative setup methods, and recommend Docker for development or production
- Rename sections to "Option 1: Installation" and "Option 2: Use Docker"
- Fix misleading comment "create a virtual environment in docker" (this section does not use Docker)
- Move Docker recommendation sentence to the top for better clarity

## Related Issues
N/A

### DIFF
--- a/docs/get_started/installation.md
+++ b/docs/get_started/installation.md
@@ -1,15 +1,17 @@
 # 🚀 Get Started
 
-## 📦 Installation
+Choose one of the following methods to set up your environment. We highly recommend using Docker for development or production environment.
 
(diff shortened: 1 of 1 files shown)


JingwenGu0829:
Closing this for now. The Docker setup and the uv-based local environment are separate paths, so this change doesn't quite make sense as a clarification. （So basically the best practice is to actually follow both) Thanks!

PR #481, opened 2026-05-19
feat(higgs-tts): CUDA Graph capture for AR decode

PopSoda2002:
**Draft PR — tracking the implementation. Plan landed; code follows
across the six stages below.**

## TL;DR

Capture the AR decode forward into CUDA Graphs to eliminate Python +
launch dispatch overhead (~3-5 ms per step, ~30-40% of total step
time). Single biggest perf knob on the higgs_tts roadmap (#478).

Currently disabled via `disable_cuda_graph: True` because per-request
sampler state is a Python dict and the sampler step has Python `if/elif`
branches. The plan is to tensorize state + vectorize the step, then
hand off capture to sglang's existing `CudaGraphRunner`.

## Expected impact

| Metric            | Before  | Target  | Δ    |
|-------------------|---------|---------|------|
| AR step time      | ~13 ms  | ~8 ms   | -38% |
| RTF C=1           | 0.55    | 0.35    | -36% |
| audio_s/s C=32    | 14      | 22      | +57% |

## Plan

Full plan committed to the repo at
[`docs/dev/higgs_tts/cuda_graph_plan.md`](docs/dev/higgs_tts/cuda_graph_plan.md).
Six stages, ~2.5 weeks:

- [ ] **Stage 1 — Tensorize sampler state** (3 days)
  Move `HiggsSamplerState` fields into `[max_bs, ...]` GPU tensors
  inside a new `HiggsBatchedSamplerState`. Per-request row mapping via
  `acquire_row` / `release_row`.
- [ ] **Stage 2 — Vectorize sampler step** (3 days)
  Replace the Python `if/elif/else` in `sampler.step()` with
  `torch.where` masks on tensor state. Numerical parity vs the old
  per-row step verified.
- [ ] **Stage 3 — Row allocation / recycling** (2 days)
  Hook row lifecycle into `prepare_prefill` (acquire) and
  `result_adapter` + `on_abort` (release).
- [ ] **Stage 4 — Hook into sglang `CudaGraphRunner`** (1-2 days)
  Flip `disable_cuda_graph: False`. sglang already provides the
  capture machinery; we just supply a CUDA-Graph-friendly forward.
- [ ] **Stage 5 — Padding handling** (1 day)
  CUDA Graph captures fixed batch sizes (1, 2, 4, 8, 16, 32). Make
  sure padded rows don't pollute real state.
- [ ] **Stage 6 — Test + numerical parity** (3 days)
  Bit-identical codes (matching seed) vs the non-graph path, plus the
[...]

Issue #478, opened 2026-05-18
Higgs Audio TTS Performance Optimization Roadmap


PopSoda2002:
## Goal

Low latency and high throughput with no model performance reduction

## Done

- [x] Model support (#428)
Model arch and weights: https://huggingface.co/boson-sglang/higgs-audio-v3-generation-4B-base
- [x] Radix cache (extra_key namespacing per ref audio)

### Evaluation Adding

- [x] seed-tts speaker similarity benchmark — #469 @JiaxinD
- [x] Seed TTS normalization. Everything matched! @jonahmackey
- [x] benchmark performance in different hardware node like H100, H200 @0xjeffro 
### Code refactor
- [x] Reuse `OmniScheduler` (drop custom `HiggsScheduler`) — #476 @JingwenGu0829 

### Bug

- [ ] Catastrophic WER outliers at large CUDA graph batch sizes (bs ≥ 16) — #578 
  - Note: dramatically worse (99–100/100 vs 0.1–0.4 % baseline) when `torch.compile` is also enabled on the backbone at `bs == cuda_graph_max_bs`. Diagnosed in #579 / `issue_565_torch_compile_result.md` as an upstream sglang × inductor CG-capture interaction. **TODO: file upstream sglang issue with the diagnostic data**, then lift the `_compiled_max_decode_bs=12` cap in `stages.py`.

### Optimization
**Encoder**

- [x] Fuse preprocessing and encoder stage @Hayden727 https://github.com/sgl-project/sglang-omni/issues/576
- [x] LRU cache for audio encoder stage — #562 @xinlij @nagisa-kunhah #563 #605 @BBuf @Ccyest 
- [ ] Encoder CUDA Graph @JingwenGu0829 
- [x] batched encoder https://github.com/sgl-project/sglang-omni/issues/603 @xinlij https://github.com/sgl-project/sglang-omni/pull/610 @xinlij @Ccyest 
- [x] encoder torch compile @yxs  https://github.com/sgl-project/sglang-omni/pull/612

**AR Step**
- [x] CUDA Graph acceleration https://github.com/sgl-project/sglang-omni/pull/503 @JiaxinD @Ccyest 
- [x] torch.compile — #565 @Ccyest @yxs @zhaochenyang20 @Ratish1   : https://github.com/sgl-project/sglang-omni/pull/579 
[...]

PR #469, opened 2026-05-18
Add SeedTTS speaker similarity benchmark

PopSoda2002:
## Motivation
SeedTTS voice-clone benchmarks currently report speed and WER, but not whether
the generated audio preserves the prompt speaker. This adds an offline
speaker-similarity pass over saved generated WAV files, integrated into both
the local benchmark entry points and the CI pipeline for S2-Pro and Qwen3-Omni.

## Modifications
- Add `--similarity-only` to the S2-Pro and Qwen3-Omni SeedTTS benchmark entry
  points.
- Reuse each sample's reference audio from the SeedTTS `meta.lst` file.
- Compute WavLM-large fine-tuned speaker-verification embeddings in process and
  report prompt-vs-generated cosine similarity.
- Write `similarity_results.json` with `speaker_similarity_mean` and per-sample
  scores.
- (After review) shared asset bootstrapper at
  `benchmarks/metrics/speaker_similarity_assets.py`: auto-downloads the
  WavLM checkpoints into a cache directory (`SEEDTTS_SIM_CACHE_DIR`, default
  `~/.cache/sglang-omni/speaker_sim`), used by both the local
  `--similarity-only` path and the two CI workflows.
- (After review) `s3prl` source clone removed; pip-installed `s3prl>=0.4.18`
  supplies the WavLM loader (byte-identical `s3prl.upstream.wavlm.hubconf`).

## Usage
```bash
# Default — bootstrapper downloads the WavLM assets automatically:
python -m benchmarks.eval.benchmark_tts_seedtts \
    --similarity-only \
    --meta seedtts_testset/en/meta.lst \
    --model fishaudio/s2-pro \
    --output-dir results/s2pro_en \
    --device cuda:0

# Or override with a custom fine-tuned WavLM checkpoint:
python -m benchmarks.eval.benchmark_tts_seedtts \
    --similarity-only \
    --meta seedtts_testset/en/meta.lst \
    --model fishaudio/s2-pro \
    --output-dir results/s2pro_en \
    --device cuda:0 \
    --similarity-checkpoint /path/to/wavlm_large_finetune.pth
```

## Speaker similarity calibration

Per @zhaochenyang20's review (items 4–6), five independent runs were collected
on the same SeedTTS-50 EN subset with identical scorer
(`popsoda2002/seedtts-wavlm-sim` @ `wavlm_large_finetune.pth`), concurrency 16,
on H200 (Spec GPU 4–7).

### S2-Pro (`fishaudio/s2-pro`)

[...]

PR #503, opened 2026-05-21
feat(higgs-tts): CUDA Graph capture for AR decode

Ccyest:
Follow-up to [PR #481](https://github.com/sgl-project/sglang-omni/pull/481) on the higgs-tts roadmap (#478).

## TL;DR

Capture the AR decode forward into CUDA Graphs to eliminate Python + launch dispatch overhead (~3–5 ms per step, ~30–40% of total step time). This was previously disabled via `disable_cuda_graph: True` because per-request sampler state was a Python dict and the sampler had Python `if/elif` branches. The PR tensorizes the state, vectorizes the step, and hands capture off to sglang's existing `CudaGraphRunner`.

## Implementation

Six stages, all landed (design in [`docs/dev/higgs_tts/cuda_graph_plan.md`](docs/dev/higgs_tts/cuda_graph_plan.md)):

- [x] **Stage 1** — Tensorize sampler state (`HiggsBatchedSamplerState`, `[max_bs, ...]` GPU tensors, `acquire_row` / `release_row`)
- [x] **Stage 2** — Vectorize sampler step (`torch.where` masks; parity vs the per-row reference)
- [x] **Stage 3** — Row allocation / recycling hooked into `prepare_prefill` and abort
- [x] **Stage 4** — Hook into sglang's `CudaGraphRunner` (flip `disable_cuda_graph: False`)
- [x] **Stage 5** — Padding handling for the captured fixed bs set
- [x] **Stage 6** — Tests + N=100 / N=1088 seed-tts eval (see Benchmark results below)

Plus three follow-up fixes uncovered during validation:
- Bug #1 — RoPE storage swap broke captured kernels (`dd7a701`).
- Bug #2 — Python branch on `top_k` was elided at capture, hard-baking the filter (`954f4d7`).
- Shadow-buffer pattern — `pool[row_indices]` gather/scatter inside the captured graph was producing wrong logits at certain batch positions for bs ≥ 8; copy active rows into per-batch shadow tensors outside the graph instead (`3c9a34f`). Dropped the catastrophic-WER rate at c=16 from 4–12% to 0.43%.

## Benchmark results

Single H200, `top_k=50`, `temperature=0.8`, voice-clone, SeedTTS EN, `max_running_requests=16`.

### Full EN set (N=1088) — CG-on vs CG-off head-to-head

Each row: 1074 successful out of 1088 (same 14 prompts time-out on both servers — unrelated to CG).

| Metric | CG-off (c=32) | CG-on (c=32) | Δ |
| --- | ---: | ---: | ---: |
| Throughput (req/s) | 6.18 | **10.44** | **+69%** |
| audio_s/s | 27.33 | **47.43** | **+74%** |
| RTF mean | 1.195 | **0.720** | **−40%** |
| Latency mean (s) | 4.89 | **2.97** | **−39%** |
| Latency p99 (s) | 6.81 | **4.12** | **−39%** |
| WER excl >50% | 1.39% | **1.26%** | **−0.13 pp** |
| Catastrophic / 1074 | 1 (0.09%) | 2 (0.19%) | +1 outlier |

Single-stream sanity for the C=1 exit-criterion: **RTF = 0.144** (CG-on, N=1088, 0 failures).

### Per-concurrency breakdown (N=100 each, single H200, includes TTFT and speaker_sim)

| c | runs | samples | catastrophic | cat. rate | WER corpus | WER excl >50% | speaker_sim (×100) | throughput (req/s) | RTF | audio_s/s | TTFT mean (s) | TTFT p95 | TTFT p99 |
|:---:|:---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1  | 1 | 100 | 0 | 0.00% | 1.27% | 1.27% | 65.70 | 1.84 | 0.143 | 7.13 | 0.705 | 0.825 | 1.477 |
| 2  | 1 | 100 | 0 | 0.00% | 0.91% | 0.91% | 64.43 | 3.07 | 0.173 | 11.82 | 0.674 | 0.920 | 0.991 |
| 4  | 1 | 100 | 0 | 0.00% | 1.09% | 1.09% | 65.13 | 5.21 | 0.207 | 19.65 | 1.212 | 1.792 | 12.239 |
| 8  | 1 | 100 | 0 | 0.00% | 0.91% | 0.91% | 64.11 | 5.72 | 0.239 | 30.52 | 1.296 | 1.547 | 11.423 |
| 16 | 7 | 700 | **3** | **0.43%** | 11.08%\* | 1.24% | 64.62 | 7.98 | 0.391 | 35.04 | 1.512 | 2.130 | 3.614 |
| 32 | 1 | 100 | 0 | 0.00% | 0.82% | 0.82% | 64.50 | 11.67 | 0.711 | 44.31 | 2.364 | 3.482 | 3.615 |
| **all** | 12 | **1200** | **3** | **0.25%** | — | — | **~64.8** | — | — | — | — | — | — |

\* c=16 corpus WER pulled up by the 3 catastrophic outliers; **WER excl >50% = 1.24%** is the apples-to-apples number. The 7-run c=16 column merges 7 independent N=100 benchmark sweeps.

### Exit-criteria status

| Criterion | Target | Measured | Status |
| --- | --- | --- | :---: |
| Δ WER vs CG-off | < 0.005 | excl outliers: 1.26% vs CG-off 1.39% → Δ ≈ −0.0013 | ✅ |
| Δ speaker_sim | < 1.0 | 64.11–65.70 spread 1.59 pp across all c; absolute level (~64.8) on par with S2-Pro 63.74 reference from PR #469 | ✅ |
| RTF @ C=1 | ≤ 0.40 | **0.144** | ✅ |
| audio_s/s @ C=32 | ≥ 20 | **47.43** | ✅ |
| Capture < 60 s | — | < 2 s | ✅ |
| Abort / chunked prefill / max_new_tokens boundary | pass | not explicitly tested | ⏸ |

## Open follow-ups (not blocking initial review)

1. **Catastrophic-WER tail risk** — at c=16 / c=32 a tiny fraction of samples (~0.1–0.4%) still drift into runaway generation that hits `max_new_tokens` instead of EOC. The shadow-buffer fix dropped the rate by 10×+ but did not eliminate it. Next step: use the env-var-gated `HIGGS_DUMP_DIR` instrumentation in `model.py` / `model_runner.py` to dump a c=16 catastrophic sample against the same prompt at c=1, and diff `input_embeds` / `hidden_states_last` / `logits_BNV` boundary-by-boundary to find the first divergence layer.

2. **Boundary cases** —
   - **Abort**: client disconnects mid-generation → verify the server properly `release_row`s and does not leak pool slots / `output_codes` entries.
   - **Chunked prefill**: prompt longer than `chunked_prefill_size=8192` → confirm the captured decode path resumes correctly after the chunked-prefill transition.
   - **`max_new_tokens` boundary**: same surface as the catastrophic samples; fold into the catastrophic investigation.

Co-authored by @PopSoda2002 @Ccyest @JiaxinD


docs/dev/higgs_tts/cuda_graph_plan.md
zhaochenyang20, line 1: remove it

sglang_omni/models/higgs_tts/model.py
```diff
@@ -135,7 +152,67 @@ def __init__(
...
+            max_batch_size=pool_size,
+            num_codebooks=num_codebooks,
+            device=self.backbone.model.embed_tokens.weight.device,
+        )
+        self._padding_row = self._max_batch_size  # last row reserved
+        self._rid_to_row: dict[str, int] = {}
+        self._free_rows: list[int] = list(range(self._max_batch_size))
+        self._output_codes: dict[str, list[torch.Tensor]] = {}
+
+        # ----- CUDA Graph buffers (Stage 4) -----------------------------
```
zhaochenyang20: remove these useless comments

sglang_omni/models/higgs_tts/model.py
```diff
@@ -135,7 +152,67 @@ def __init__(
                 self.multimodal_embedding.modality_embedding_0.weight
             )
 
-        self._slots: dict[str, _RequestSlot] = {}
+        # Per-request sampler state lives in a fixed-size GPU pool (Stage 1
+        # of the CUDA Graph migration). ``_rid_to_row`` maps request id →
+        # row index; ``_free_rows`` tracks unused rows; ``_output_codes``
+        # holds the variable-length per-request code log (cannot be
+        # tensorised cleanly).
```
zhaochenyang20: remove these useless comments

sglang_omni/models/higgs_tts/model.py
```diff
@@ -135,7 +152,67 @@ def __init__(
...
+        self._max_batch_size = int(max_batch_size)
+        # +1 row reserved for graph-capture padding (sglang's CudaGraphRunner
+        # captures specific batch sizes 1, 2, 4, 8, 16, 32 and pads the actual
+        # batch up to the next supported one). The padding row's state is
+        # reset by the model_runner each step so it never bleeds into real rows.
```
zhaochenyang20: remove these useless comments

sglang_omni/models/higgs_tts/model.py
```diff
@@ -135,7 +152,67 @@ def __init__(
...
+        )
+        self._padding_row = self._max_batch_size  # last row reserved
+        self._rid_to_row: dict[str, int] = {}
+        self._free_rows: list[int] = list(range(self._max_batch_size))
+        self._output_codes: dict[str, list[torch.Tensor]] = {}
+
+        # ----- CUDA Graph buffers (Stage 4) -----------------------------
+        # Allocated once, written by ``model_runner.prepare_decode`` before
+        # each forward, read by ``model.forward`` in decode mode. Sized
+        # ``[pool_size]`` so padding rows index safely into the pool too.
```
zhaochenyang20: remove these useless comments

sglang_omni/models/higgs_tts/model.py
zhaochenyang20, line 1: remove these useless comments

sglang_omni/models/higgs_tts/model.py
```diff
@@ -199,26 +320,132 @@ def decode_codebooks_batch(
...
+                dtype=torch.float32,
+                device=device,
+            )
+            if has_top_p
+            else None
+        )
+        # ``batched_step`` requires uniform top_k across the batch (matches
+        # how every Higgs request is configured in practice). If callers
+        # ever pass heterogeneous top_k we want to know — fail loudly.
+        distinct_top_k = {p.top_k for p in gen_params}
```
zhaochenyang20: this branch collapses `gen_params` top-k values into a `set`, so a single prefill batch with two valid requests using different `top_k` values raises `ValueError` at line 352. The old implementation called `sampler_step(..., top_k=params.top_k)` per request, so heterogeneous sampling params were supported; the new CUDA Graph decode path also supports per-row top-k through `_cg_top_k_buf`. Only the prefill eager path regressed to uniform-only, so concurrent requests mixing default `top_k=None` and `top_k=50` can fail instead of generating normally.

Do not require uniform top-k in the prefill path. Reuse the CUDA Graph path's per-row buffer semantics: convert each row's `top_k` to a `[B]` tensor, map `None` or non-positive values to `K_MAX`, and call `batched_step(..., top_k_buf=top_k_buf)`; alternatively make `_sample_independent_batched` accept per-row eager top-k. Add a regression test with one request using `top_k=None` and another using `top_k=5`, and verify prefill no longer raises while each row uses its own filter.

sglang_omni/models/higgs_tts/model_runner.py
zhaochenyang20: The module-level docs still say `prepare_decode` only propagates `req_ids`, and that `post_prefill` / `post_decode` read results from `model._slots[req_id].output_codes[-1]`. The core change in this PR is the opposite: decode now writes `_cg_*` buffers first, post-decode scatters CUDA Graph output buffers back into the sampler pool, and `_slots` is no longer the main path. This stale comment will send maintainers back to the old model when debugging padding rows, abort cleanup, or chunked prefill, which is exactly the lifecycle surface most likely to fail.

zhaochenyang20 (changes requested):

PopSoda2002:
Hi @Ccyest Can I know the speed performance before and after cudagraph?

sglang_omni/models/higgs_tts/model_runner.py
```diff
@@ -45,11 +40,144 @@ def post_prefill(self, result, forward_batch, schedule_batch, requests):
...
+        if top_ks_raw is None:
+            top_ks = [None] * n_real
+        else:
+            top_ks = [int(t) for t in top_ks_raw[:n_real]]
+            for tk in top_ks:
+                if tk is not None and tk > K_MAX:
+                    raise ValueError(
+                        f"top_k={tk} > K_MAX={K_MAX} not supported in the "
+                        f"CG path, see sampler.py."
+                    )
```
zhaochenyang20: normal Higgs TTS requests do not explicitly pass `top_k`, and `build_sglang_higgs_request()` therefore leaves `top_k` out of `SamplingParams`. SGLang normalizes the default `top_k=-1` into `TOP_K_ALL=1073741824`; once `sampling_info.top_ks` reaches `_extract_decode_sampling_params()`, line 129 treats it as an illegal `> K_MAX` value and raises `ValueError`. A default request can finish prefill and then fail on the first decode step.
Do not reject oversized `top_k` at the runner boundary. Normalize `None`, non-positive values, `TOP_K_ALL`, and any `top_k >= K_MAX` to `K_MAX`, so the inner `top_k_buf` clamp represents a no-op filter; this also matches the old `_sample_independent()` behavior of `min(top_k, logits.size(-1))`.

sglang_omni/models/higgs_tts/model_runner.py
Ccyest: fixed thanks

tests/unit_test/higgs_tts/test_batched_step.py
```diff
@@ -0,0 +1,270 @@
...
+    K_MAX,
+    STOP_CODE,
+    HiggsBatchedSamplerState,
+    batched_step,
+    step,
+)
+from sglang_omni.models.higgs_tts.utils import EOC_ID
+
+DEVICE = "cuda" if torch.cuda.is_available() else "cpu"
+GREEDY_TEMP = 1e-3  # below _GREEDY_TEMP_THRESHOLD for per-row argmax path
```
zhaochenyang20: This says `GREEDY_TEMP = 1e-3` is below `_GREEDY_TEMP_THRESHOLD`, but the actual threshold in `sampler.py` is `1e-5`. That means per-row `step()` is not taking the argmax short-circuit; it is taking the softmax/multinomial path, with very peaky test logits making the result look greedy. 

change the test temperature to `1e-5`. If the goal is to cover SGLang's public greedy contract, the more direct test is `top_k=1`, because `SamplingParams(temperature=0)` is normalized by SGLang into `temperature=1.0, top_k=1`.

zhaochenyang20 (changes requested):

sglang_omni/models/higgs_tts/model_runner.py
Ccyest: made it to no op instead of raising error.

tests/unit_test/higgs_tts/test_batched_step.py
Ccyest: use top_k=1 instead

Ccyest:
> Hi @Ccyest Can I know the speed performance before and after cudagraph?

Sure — full SeedTTS EN set (1088 samples, 1074 successful on both servers; `top_k=50` / `temperature=0.8`; single H200; `max_running_requests=16`).

**Head-to-head @ c=32**

| Metric | CG-off | CG-on | Δ |
| --- | ---: | ---: | ---: |
| Throughput (req/s) | 6.18 | **10.44** | **+69%** |
| audio_s/s | 27.33 | **47.43** | **+74%** |
| RTF mean | 1.195 | **0.720** | **−40%** |
| Latency mean (s) | 4.89 | **2.97** | **−39%** |
| Latency p99 (s) | 6.81 | **4.12** | **−39%** |

**Single-stream sanity (c=1, CG-on)**

| Metric | Value | PR target |
| --- | ---: | ---: |
| RTF | **0.144** | ≤ 0.40 |
| Throughput (req/s) | 1.61 | — |
| audio_s/s | 7.08 | — |
| Latency mean (s) | 0.62 | — |

PR exit-criteria check:
- RTF @ C=1 ≤ 0.40 → **0.144** ✅
- audio_s/s @ C=32 ≥ 20 → **47.43** ✅

Quality side-effect for the same runs: c=32 CG-on WER excl >50% = **1.26%**, slightly below CG-off's 1.39% (−0.13 pp). Full WER + catastrophic-outlier discussion is in the **Benchmark results → Full SeedTTS EN set** section of the PR description above.

zhaochenyang20 (approved):

PR #505, opened 2026-05-21
[Qwen3]: Route preprocessing through active encoder branches

Ratish1:
## Motivation

Qwen3-Omni preprocessing has static pipeline edges for all possible encoder branches, but each request only activates a subset of those branches. The pipeline already validates static routing supersets; this PR adds the matching request-level fan-in hook so aggregate stages can wait on the sources that are active for the current request.

This is a standalone routing/fan-in correctness change. It does not claim to be a benchmark threshold change or a full topology/performance fix.

## Modifications

- Added `StageConfig.wait_for_fn` and threaded it into stage-process construction so aggregate inputs can resolve expected sources per request while keeping the static `wait_for` set as the validation boundary.
- Extended `AggregatedInput` with request-scoped expected source sets. Dynamic sources must be non-empty strings contained in the static `wait_for` superset, and pending state is cleared on request cancellation.
- Updated Qwen3-Omni preprocessing routing so raw preprocessing payloads fan out only to encoder branches with real model inputs, then always continue to `mm_aggregate`.
- Updated Qwen3-Omni aggregate fan-in so `mm_aggregate` waits for preprocessing plus the projected active encoder branches for that request.
- Preserved lightweight encoder cache metadata in the preprocessing-to-aggregate projection, using `_active` only as aggregate metadata. Metadata such as `cache_key` alone no longer activates an encoder branch.
- Added unit coverage for dynamic fan-in validation, stage-process hook wiring, Qwen active-branch routing, metadata-only encoder inputs, projected active metadata, and the real projection-to-merge path.

## Accuracy Tests

N/A

### DIFF
--- a/sglang_omni/models/qwen3_omni/request_builders.py
+++ b/sglang_omni/models/qwen3_omni/request_builders.py
@@ -27,6 +27,7 @@ AUDIO_STAGE = "audio_encoder"
 DECODE_STAGE = "decode"
 TALKER_STAGE = "talker_ar"
 CODE2WAV_STAGE = "code2wav"
+MM_AGGREGATE_STAGE = "mm_aggregate"
 
 
 def output_modalities(request: OmniRequest | None) -> set[str] | None:
@@ -79,6 +80,26 @@ def resolve_terminal_stages(request: OmniRequest) -> list[str]:
     if should_generate_audio_output(request):
         return [DECODE_STAGE, CODE2WAV_STAGE]
     return [DECODE_STAGE]
+
+
+def resolve_preprocessing_next_stages(
+    request_id: str, output: StagePayload
+) -> list[str]:
+    del request_id
+    state = PipelineState.from_dict(output.data)
+    return [*_active_encoder_stages(state.encoder_inputs), MM_AGGREGATE_STAGE]
+
+
+def resolve_mm_aggregate_wait_sources(
+    request_id: str,
+    from_stage: str,
+    payload: StagePayload,
+) -> list[str] | None:
+    del request_id
+    if from_stage != "preprocessing":
+        return None
+    state = PipelineState.from_dict(payload.data)
+    return ["preprocessing", *_active_encoder_stages(state.encoder_inputs)]
 
 
 @dataclass(slots=True)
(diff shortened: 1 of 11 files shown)


sglang_omni/models/qwen3_omni/request_builders.py
```diff
@@ -215,11 +236,31 @@ def _project_encoder_input_metadata(
...
+        if _is_active_encoder_input(encoder_inputs.get(stage_name))
+    ]
+
+
+def _is_active_encoder_input(stage_inputs: Any) -> bool:
+    return (
+        isinstance(stage_inputs, dict)
+        and bool(stage_inputs)
+        and not bool(stage_inputs.get("_skip"))
+    )
```
zhaochenyang20: the core fan-out and fan-in contract now depends on `_is_active_encoder_input`, but the predicate currently checks only that the value is a non-empty `dict` without `_skip`. If preprocessing later adds `cache_key`, shape metadata, debug markers, or an explicit `_active: False` marker for an encoder, this helper still classifies the branch as active. The runtime will route the request to that encoder and make `mm_aggregate` wait for the source; if the encoder receives only metadata and the cache misses, `build_encoder_request` builds empty `model_inputs`, so the encoder can end up running the model with empty kwargs or at least reintroduce the useless fanout/fanin barrier this PR is trying to remove. This matters because dynamic topology contract becomes controlled by metadata shape rather than real work, which is a real regression risk for future cache changes. The current preprocessor happens to guard raw payloads with real media tensors at `preprocessor.py:457` and `preprocessor.py:463`, but the new helper name and placement make this loose predicate easy to misuse in future cache or metadata changes.

Split the raw preprocessing branch predicate from the lightweight aggregate-metadata predicate, or make `_is_active_encoder_input` take `stage_name` and treat only `_active is True`, image `pixel_values`/`pixel_values_videos`, and audio `input_features` as active. Pure metadata such as `cache_key` should not activate a branch by itself. Add tests for `{"cache_key": "x"}`, `{"_active": False}`, and `{"cache_key": "x", "_active": True}` so route and wait semantics cannot drift when metadata expands.

tests/unit_test/qwen3_omni/test_pipeline.py
zhaochenyang20, line 711: This test first asserts that the lightweight preprocessing payload preserves `cache_key` and `_active` in `projected_state`, but the actual `merge_for_thinker` call passes `make_qwen_payload(state)`, the unprojected full preprocessing state. In production, `mm_aggregate` receives the projected payload, so the test does not feed the cache metadata that this PR claims to protect through the real merge path. This coverage gap is a regression risk: if the projected payload and the fields consumed by merge drift later, the projection assertion and the full-state merge assertion can still pass independently while the actual aggregate path is broken.

Use the previously built `projected` payload as the `"preprocessing"` payload in this test, provide the matching active encoder outputs for the dynamic wait set, and then assert `media_cache_keys`, stripped `mm_inputs`, and consumed `encoder_inputs`. That makes the test cover the real runtime path from `project_preprocessing_to_mm_aggregate` into `merge_for_thinker`, instead of two disconnected local behaviors.

zhaochenyang20 (changes requested):
Also, please update MMSU's performance and accuracy in this PR.

sglang_omni/models/qwen3_omni/request_builders.py
Ratish1: Done

tests/unit_test/qwen3_omni/test_pipeline.py
Ratish1: Done

zhaochenyang20 (approved):

PR #437, opened 2026-05-13
[V1] Ming omni v1 migration

edwingao28:
## Motivation
Based on RFC https://github.com/sgl-project/sglang-omni/issues/429
Port Ming-Omni from the legacy runner to `sglang_omni_v1`, the staged pipeline (preprocess → thinker → talker) used by Qwen3-Omni and S2-Pro. Without this, Ming cannot run under the V1 scheduler, share its multiprocess stage runner, or benefit from the V1 OpenAI-compatible server.

## Modifications

- `models/ming_omni/`: add `bootstrap.py`, `stages.py`, `registration.py`; rewrite `config.py`, `preprocessor.py`, `vision_encoder.py`, `talker_executor.py`, `thinker.py`, and shared `components/common.py` to the V1 stage contract.
- `model_runner/ming_thinker_model_runner.py`: new Ming-specific thinker runner; wire into `model_worker.py` and `sglang_model_runner.py`.
- `scheduling/omni_scheduler.py`, `pipeline/mp_runner.py`, `pipeline/stage_process.py`, `serve/openai_api.py`: minor adjustments to support Ming's stage graph and chat-completion path.
-   Tests: tests/unit_test/ming_omni/test_{pipeline,thinker,talker,tp,tokenizer}.py, `tests/test_v1_omni_scheduler.py` and `tests/test_openai_api.py`.
## Related Issues
closes #405 
### TTS

Category | Metric | v0 | v1 | Δ
-- | -- | -- | -- | --
Generation speed (Ming → audio, n=50, concurrency=1) | Completed / failed | 50 / 0 | 50 / 0 | —
Generation speed | Latency mean | 0.839 s | 0.580 s | -30.9%
Generation speed | Latency median | 0.808 s | 0.367 s | -54.6%
Generation speed | Latency p95 | 1.280 s | 0.607 s | -52.6%
Generation speed | Latency p99 | 1.431 s | 5.204 s | +263.7%
Generation speed | Audio duration mean | 4.140 s | 7.692 s | +85.8%
Generation speed | RTF mean (latency/audio) | 0.2074 | 0.0969 | -53.3%
Generation speed | RTF median | 0.2017 | 0.0508 | -74.8%
Generation speed | Throughput | 1.192 req/s | 1.723 req/s | +44.5%
Accuracy — WER (Whisper-large-v3 readback) | WER corpus | 2.30 % | 4.26 % | +85.2%
Accuracy — WER | WER per-sample mean | 2.50 % | 4.25 % | +70.0%
[...]

PR #378, opened 2026-05-03
[Feat] add streaming TTS and test to Ming Omni

edwingao28:
<!-- Thank you for your contribution! We appreciate it. The following guidelines will help improve your pull request and facilitate feedback. If anything is unclear, don't hesitate to submit your pull request and ask the maintainers for assistance. -->

## Motivation
Ming-Omni speech currently runs `Thinker_full → Talker_full`, so TTFA ≈ Thinker latency + Talker latency. Add an opt-in streaming pipeline so audio starts emitting before Thinker finishes, without changing the default non-streaming path.       

Assuming the prompt/model/server setup is otherwise identical, streaming TTS keeps throughput essentially unchanged while reducing first-audio latency by more than 6x.                           
## Modifications

Adds a sibling streaming speech pipeline `thinker → segmenter → talker_stream` selected by a new CLI flag. Non-streaming `MingOmniSpeechPipelineConfig` behavior is unchanged.
                                                                                                                                                                                                                                                                                          
  - `sglang_omni/models/ming_omni/config.py`: add `MingOmniStreamingSpeechPipelineConfig`; add `SEGMENTER_STAGE` / `TALKER_STREAM_STAGE` and reuse GPU-placement validator for the streaming talker stage.
  - `sglang_omni/models/ming_omni/components/streaming_text.py`: UTF-8 tensor conversion and punctuation/length segmenter state with first-segment wall-clock flush.                                                                                                                      
  - `sglang_omni/models/ming_omni/components/streaming_segmenter_executor.py`: streaming `Executor` that turns Thinker text deltas into Talker text segments.       
  - `sglang_omni/models/ming_omni/components/streaming_talker_executor.py`: streaming `Executor` that runs Ming Talker generation off the asyncio loop and emits audio chunks; abort-aware cleanup of VAE cache and graph pool.                                                           
  - `sglang_omni/models/ming_omni/talker/modeling_ming_omni_talker.py`: replace `time.sleep(0.1)` polling with condition wakeup; abort checks; graph-pool `try/finally`; remove `(count == 0)` gate; keep `vae_cache[uuid]` for the lifetime of a streaming session.                      
  - `sglang_omni/models/ming_omni/pipeline/{next_stage.py,stages.py}`: thinker text side channel and segmenter / talker-stream factories.                                                                                                                                                 
  - `sglang_omni/serve/openai_api.py`: propagate SSE generator cancellation to `client.abort(request_id)` and record per-stage timing.                                                                                                                                                    
[...]

PR #506, opened 2026-05-21
[Ming-Omni] Support Streaming TTS for v1 Pipeline

yuan-luo:
## Summary

Adds an opt-in streaming-TTS pipeline variant to Ming-Omni V1, modelled on the V0 design in #378 but adapted to the `sglang_omni_v1` staged-pipeline framework. With `--enable-streaming-tts`, the OpenAI-compatible chat endpoint begins emitting audio chunks **per sentence** instead of waiting for the entire utterance to finish synthesizing, dramatically reducing time-to-first-audio. Default behavior (the existing 7-stage non-streaming speech pipeline) is unchanged.

## Motivation

The V1 Ming-Omni pipeline (#437) preserves the V0 contract of one-shot, end-of-generation audio delivery. For a 50-character Chinese utterance this means the user waits for both the thinker (~Ns) **and** the entire talker run (~Ms) before hearing anything. #378 showed this can be cut to ~600 ms on V0 by sentence-splitting the thinker output and pipelining TTS — this PR ports the same idea to V1.

Architecturally V1 is a better home for this because the thinker, segmenter, and talker each run in their own OS process and communicate via the existing `stream_to` / `can_accept_stream_before_payload` plumbing already used by Qwen3-Omni's talker_ar.

## Architecture

New 8-stage pipeline `MingOmniStreamingSpeechPipelineConfig` (`Variants["streaming_speech"]`):
```
preprocessing → audio_encoder ↘
              → image_encoder → mm_aggregate → thinker ─stream→ segmenter ─stream→ talker_stream → (coordinator)
                                                  │                 │
                                                  └─next→ decode    └─next→ talker_stream
                                                  └─next→ segmenter
```
- **thinker** (existing `OmniScheduler`): adds an opt-in `stream_output_builder`. Per generated token it does an incremental `tokenizer.decode`, computes the text delta, and emits a `uint8` tensor to the segmenter via the `stream_to`
channel. Per-request running token list and emitted text are stashed on the SGLang `Req` so they GC with the request.
- **segmenter** (new `MingStreamingSegmenterScheduler`): pure-Python stream-aware scheduler with the same `inbox`/`outbox`/`start`/`stop`/`abort` contract as `SimpleScheduler`. Consumes uint8 text deltas, runs them through the V0-reused
`SegmenterState` (punctuation + min/max token thresholds + 450 ms first-segment timeout), and emits per-segment uint8 tensors to talker_stream.
- **talker_stream** (new `MingStreamingTalkerScheduler`): terminal stream-aware scheduler. For each incoming text segment it drives `talker.omni_audio_generation(stream=True, abort_event=...)` synchronously and publishes each yielded
waveform as `OutgoingMessage(type="stream", target=None)` so the runtime forwards it directly to the coordinator (same pattern as `code2wav_scheduler`).

The runtime already handles all the wiring (`stream_to`, `can_accept_stream_before_payload`, terminal output to coordinator); no changes to `pipeline/` were needed.

## Talker model changes (cherry-picked from #378)

To make the underlying `MingOmniTalker` actually streamable and abortable:

- Thread `abort_event: threading.Event` through `CFM.sample`, `CFMGraphExecutor.execute/_initialize_graph`, `CFMGraphExecutorPool`, `omni_audio_generation_func`, `omni_audio_generation`, and `instruct_audio_generation`. CUDA-graph capture is
 fenced with `abort_event=None` because aborting mid-capture corrupts the partial graph; checks happen before capture and on every replay.
- Replace the 100 ms `time.sleep` polling between LLM tokens with a 25 ms `token_queue.get(timeout=...)` wakeup that also polls `abort_event` and `future.done()`. Bounds external-abort latency to ~25 ms + one inner generator step. New
`_TOKEN_DONE` sentinel unblocks the consumer cleanly when `llm_job` finishes.
- `abort_event` defaults to `None` everywhere, so the existing non-streaming `MingTalkerExecutor` path is unaffected.

## Client / API changes

- `Client._default_result_builder` recognizes `"talker_stream"` as an audio terminal alongside `"code2wav"`/`"talker"` when merging multi-terminal results.
- `examples/run_ming_omni_speech_server.py` gains `--enable-streaming-tts` to select the new variant and route `--gpu-talker` to the `talker_stream` stage. Also adds `--cpu-offload-gb` since the speech launcher previously didn't expose it.
- No changes to `serve/openai_api.py` were needed — the SSE `_chat_stream` path already handles per-chunk audio (`modality="audio"` + `audio_b64`) and the finish-chunk text/audio retention fix from #437 already covers our final aggregate.

## OmniScheduler stub (small drive-by fix)

Adds a `session_controller = SimpleNamespace(sessions={})` stub to `OmniScheduler._init_upstream_compat_flags`. Upstream SGLang's `scheduler_runtime_checker_mixin._streaming_session_count` accesses `self.session_controller.sessions.values()`
 during `report_decode_stats` and crashed the V1 thinker on generations longer than a handful of tokens. We don't host SGLang's interactive-session feature so an empty stub is sufficient. Strictly speaking this is independent of streaming
TTS but it blocked e2e validation.

## Key design decisions

- **Sentence-level segmentation, not token-level.** Per-token TTS is meaningless and the segmenter mirrors V0: emit on `[.!?。！？;；]` after `min_tokens` OR when the first segment has waited `first_segment_max_wait_ms`. Buffer incomplete
UTF-8 (`\ufffd` in decoded) until the next token completes it.
- **Text deltas go to segmenter only, not also to coordinator.** I initially had the thinker dual-emit (segmenter + terminal) so the client would see streaming text. This crashes because (a) the relay transport requires `torch.Tensor`
payloads, not the `dict` you'd want for terminal text, and (b) `target=None` from a non-terminal stage fans out to `stream_to` peers, not to the coordinator. So streaming text to the client is deferred — the user still sees one final text
chunk via the decode stage. Streaming audio works fully.
- **State stashed on the SGLang `Req` object** rather than in a closure-level dict, so it dies with the request and there's no manual cleanup.
- **Final-result payloads are fresh small dicts.** Forwarding the upstream `state.to_dict()` (`prompt.input_ids`, `encoder_outs`, etc.) trips msgpack on terminal serialization because of the embedded tensors. Both the segmenter and
talker_stream now emit only their own summary fields.
- **Synchronous talker generation inside the scheduler thread.** V0 used a daemon thread + `loop.call_soon_threadsafe` because its scheduler ran on the asyncio loop. V1 schedulers already run on a dedicated thread, so we just iterate the
talker generator directly and push waveforms straight to `outbox` — simpler, no cross-thread queue.

### Unit (60 ming_omni tests pass on H20 pod)

- `test_streaming_text.py` (4): UTF-8 round-trip, punctuation-flush, first-segment timeout, max-tokens cap.
- `test_streaming_segmenter.py` (3): punctuation emit + clean final payload; stream-before-payload buffering; first-segment timeout under inbox-idle.
- `test_streaming_talker.py` (3): fake-talker per-segment emits + clean final payload + audio-chunk count; stream-before-payload buffering; abort short-circuits a slow talker.
- `test_streaming_speech_config.py` (6): new variant topology (`next` / `stream_to` / `can_accept_stream_before_payload`), `Variants` registration, GPU-collision validator.
- `test_streaming_e2e_glue.py` (4): thinker stream builder emits to segmenter, suppresses during chunked prefill, buffers incomplete UTF-8; client result builder merges `decode` + `talker_stream`.

### End-to-end on H20 pod (TP=2 thinker, GPU 2 talker_stream)

```bash
python examples/run_ming_omni_speech_server.py \
  --model-path /home/admin/models/inclusionAI/Ming-flash-omni-2___0 \
  --enable-streaming-tts \
  --gpu-thinker 0 --gpu-talker 2 \
  --tp-size 2 --port 8765 --mem-fraction-static 0.9

curl -sN -X POST http://127.0.0.1:8765/v1/chat/completions \
  -d '{"model":"ming-omni","messages":[{"role":"user","content":"用两句话介绍上海。"}],
       "max_tokens":150,"stream":true,"modalities":["text","audio"],"audio":{"format":"wav"}}'

Result: 12 SSE chunks in 2924 ms.

┌─────────┬───────────────────────────────────────┬────────────────────────────┐
│  Chunk  │                Content                │            Size            │
├─────────┼───────────────────────────────────────┼────────────────────────────┤
│ #00     │ role=assistant + full text (94 chars) │ text 94                    │
├─────────┼───────────────────────────────────────┼────────────────────────────┤
│ #01–#10 │ 10 independent audio waveform chunks  │ 34 KB – 646 KB base64 each │
├─────────┼───────────────────────────────────────┼────────────────────────────┤
│ #11     │ finish_reason=stop                    │ (empty)                    │
└─────────┴───────────────────────────────────────┴────────────────────────────┘

Average inter-audio-chunk gap ≈ 290 ms, consistent with sentence-level cuts on a two-sentence reply. All 8 stages remained healthy after the request; no pipeline-internal errors in the server log (the only ERROR lines are SGLang's noisy
optional-import warnings for mooncake, nixl, gemma4, glm_ocr, glmasr).

Known limitations

- Streaming text to the client is not yet enabled — the SSE response carries the full text in one delta.content chunk. To enable, either (a) make decode stream-aware so it consumes thinker token-id streams and emits text-delta terminal
chunks, or (b) add a small dedicated text-fanout terminal stage that mirrors the segmenter's input. Both are clean follow-ups.
- No formal TTFB benchmark vs the non-streaming baseline yet. V0 #378 reported ~6× improvement (3756 ms → 587 ms); we should reproduce that on V1. Belongs in a follow-up PR with a small benchmark harness.
- Ming-flash-omni-2.0 ships without top-level tokenizer files on HF/ModelScope; this is upstream's choice and orthogonal to this PR. common.py:load_ming_tokenizer already falls back to Ming-flash-omni-Preview for exactly this reason.

Files changed

sglang_omni/models/ming_omni/components/streaming_text.py        (new, 161)
sglang_omni/models/ming_omni/components/streaming_segmenter.py   (new, ~250)
sglang_omni/models/ming_omni/components/streaming_talker.py      (new, ~380)
sglang_omni/models/ming_omni/talker/modeling_ming_omni_talker.py (+331/-225, abort_event)
sglang_omni/models/ming_omni/talker/talker_module/cfm.py         (+9/-2, abort_event)
sglang_omni/models/ming_omni/bootstrap.py                        (+100, stream_output_builder)
sglang_omni/models/ming_omni/config.py                           (+streaming variant)
sglang_omni/models/ming_omni/stages.py                           (+2 factories, threading)
sglang_omni/models/ming_omni/pipeline/next_stage.py              (+SEGMENTER_STAGE, TALKER_STREAM_STAGE)
sglang_omni/client/client.py                                     (+talker_stream recognition)
sglang_omni/scheduling/omni_scheduler.py                         (+session_controller stub)
examples/run_ming_omni_speech_server.py                          (+--enable-streaming-tts, --cpu-offload-gb)
tests/unit_test/ming_omni/test_streaming_*.py                    (5 new files, 20 tests)

Test plan

- All 60 ming_omni unit tests pass locally and on the H20 pod
- Non-streaming speech pipeline still constructs and tests pass (defaults unchanged)
- Streaming speech variant launches end-to-end with model loaded
- Live SSE request returns multiple audio chunks within a single response
- Server remains healthy after the request (no dead stage processes)
- TTFB benchmark vs non-streaming (follow-up)
- Streaming text deltas to client (follow-up)
- Abort-mid-stream behavior under real load (only fake-talker tested)

### DIFF
--- a/examples/run_ming_omni_speech_server.py
+++ b/examples/run_ming_omni_speech_server.py
@@ -82,6 +82,27 @@ def parse_args() -> argparse.Namespace:
             "If omitted, SGLang chooses automatically."
         ),
     )
+    parser.add_argument(
+        "--cpu-offload-gb",
+        type=int,
+        default=None,
+        help=(
+            "Offload N GiB of thinker weights to CPU. Required for "
+            "Ming-flash-omni-2.0 (~200 GB MoE weights) on a single GPU. "
+            "Mirrors the text launcher's default of 80."
+        ),
+    )
+    # Streaming TTS
+    parser.add_argument(
+        "--enable-streaming-tts",
+        action="store_true",
+        help=(
+            "Use the 8-stage streaming-TTS pipeline (segmenter + streaming "
+            "talker) for sub-second time-to-first-audio. Default keeps the "
+            "non-streaming 7-stage speech pipeline."
+        ),
+    )
+
     # Server
     parser.add_argument("--host", type=str, default="0.0.0.0")
     parser.add_argument("--port", type=int, default=8000)
@@ -147,28 +168,44 @@ def _set_thinker_tp(config: Any, *, start_gpu: int, tp_size: int) -> None:
 
 
 def _launch_speech_server(args: argparse.Namespace) -> None:
-    from sglang_omni.models.ming_omni.config import MingOmniSpeechPipelineConfig
+    from sglang_omni.models.ming_omni.config import (
+        MingOmniSpeechPipelineConfig,
+        MingOmniStreamingSpeechPipelineConfig,
+    )
     from sglang_omni.serve import launch_server
 
     _validate_fraction("--mem-fraction-static", args.mem_fraction_static)
 
-    config = MingOmniSpeechPipelineConfig(
-        model_path=args.model_path,
-        relay_backend=args.relay_backend,
-    )
+    if getattr(args, "enable_streaming_tts", False):
+        config = MingOmniStreamingSpeechPipelineConfig(
+            model_path=args.model_path,
+            relay_backend=args.relay_backend,
+        )
+        talker_stage_name = "talker_stream"
+        gpu_validator = config._validate_talker_stream_gpu_not_in_thinker_tp_range
+    else:
+        config = MingOmniSpeechPipelineConfig(
+            model_path=args.model_path,
+            relay_backend=args.relay_backend,
+        )
+        talker_stage_name = "talker"
+        gpu_validator = config._validate_talker_gpu_not_in_thinker_tp_range
+
     _set_thinker_tp(
         config,
         start_gpu=args.gpu_thinker,
         tp_size=int(args.tp_size),
     )
-    _set_stage_gpu(config, "talker", args.gpu_talker)
-    config._validate_talker_gpu_not_in_thinker_tp_range()
+    _set_stage_gpu(config, talker_stage_name, args.gpu_talker)
+    gpu_validator()
 
     server_arg_updates: dict[str, object] = {}
     if args.tp_size and args.tp_size > 1:
         server_arg_updates["disable_custom_all_reduce"] = True
     if args.mem_fraction_static is not None:
         server_arg_updates["mem_fraction_static"] = args.mem_fraction_static
+    if getattr(args, "cpu_offload_gb", None) is not None:
+        server_arg_updates["cpu_offload_gb"] = args.cpu_offload_gb
     if server_arg_updates:
         _apply_stage_factory_updates(
             config,
@@ -177,7 +214,7 @@ def _launch_speech_server(args: argparse.Namespace) -> None:
         )
     _apply_stage_factory_updates(
         config,
-        stage_name="talker",
+        stage_name=talker_stage_name,
         updates={"voice": args.voice},
     )
 
--- a/sglang_omni/client/client.py
+++ b/sglang_omni/client/client.py
@@ -303,10 +303,11 @@ class Client:
             result.request_id = request_id
             return result
         if isinstance(result, dict):
-            # Multi-terminal merged result, e.g. decode + code2wav/talker.
+            # Multi-terminal merged result, e.g. decode + code2wav/talker/
+            # talker_stream.
             audio_result = None
             if "decode" in result:
-                for audio_stage in ("code2wav", "talker"):
+                for audio_stage in ("code2wav", "talker", "talker_stream"):
                     if audio_stage in result:
                         audio_result = result[audio_stage] or {}
                         break
--- a/sglang_omni/models/ming_omni/bootstrap.py
+++ b/sglang_omni/models/ming_omni/bootstrap.py
@@ -316,6 +316,13 @@ def make_thinker_stream_output_builder(
             list(delta.encode("utf-8")),
             dtype=torch.uint8,
         )
+        # Only emit to the segmenter. The thinker is not a terminal stage,
+        # so it cannot send chunks directly to the coordinator via
+        # target=None — the runtime would fan that out to ``stream_to``
+        # peers, and the relay transport requires torch.Tensor payloads.
+        # Streaming text deltas to the client requires either a stream-
+        # aware decode stage or a dedicated text fan-out stage; left as a
+        # follow-up. Streaming audio still works via the talker_stream.
         return [
             OutgoingMessage(
                 request_id=request_id,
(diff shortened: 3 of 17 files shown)


FrankLeeeee (approved):
looks ok to me, is the CI recovering soon? @zhaochenyang20

zhaochenyang20:
only after PR is approved, then run the CI.

yuan-luo:
All CI passed, no implementation impacts on base framework. Merged.

Issue #498, opened 2026-05-20
[CI] Investigate MMSU regression and characterize router DP scaling on H200


zhaochenyang20:
### Motivation

#477 landed the router-managed two-worker topology as the new CI shape for Qwen3-Omni, and the recalibrated thresholds show most benchmarks moving in the right direction under colocation. Two questions remain open after that PR that are worth tracking separately rather than re-opening #477.

### Track 1 — MMSU performance regression under the new topology

MMSU is the one benchmark whose numbers went the wrong way after the topology change. Comparing the recalibrated thresholds against the pre-#477 baseline:

| Metric | Before #477 | After #477 | Δ |
| --- | --- | --- | --- |
| `throughput_qps` (concurrency 16) | 30.117 | 11.135 | **-63%** |
| `tok_per_s_agg` | 7.8 | 1.4 | -82% |
| `latency_mean_s` | 0.265 | 1.434 | +441% |

This is the opposite shape of what we expected from colocation, and it does not match MMMU / Video-MME / TTS, all of which behaved as predicted. Two hypotheses worth ruling in or out:

1. **Router-side queuing or batching mismatch specific to MMSU's request profile.** MMSU is text-output-only over audio inputs, with much shorter per-request work than the talker-bearing benchmarks. If the router is serializing or batching pessimistically for short-work text-output requests, MMSU would feel it disproportionately.
2. **Colocated memory/scheduling pressure interacting with MMSU's audio encoder path.** The audio encoder under colocation is sharing GPU with everything downstream; if `mem_fraction_static` or KV-cache budgeting compresses MMSU's effective batch, the throughput collapse would land roughly where we see it.

Action items:

- Reproduce the MMSU numbers under the post-#477 router topology, then re-run under disaggregated DP=1 and disaggregated DP=2 on the same hardware. The disaggregated DP=2 number is the apples-to-apples comparison we don't currently have.
- Profile a single MMSU request end-to-end through the router and through the per-stage pipeline. We want to know where the +1.2s of mean latency comes from — router queuing, encoder, thinker, or coordinator overhead.
- If router-side, file a follow-up against the router. If colocation-side, this likely belongs as input to the broader colocation memory-budget conversation.

### Track 2 — Router DP scaling characterization on H200

[...]

PR #477, opened 2026-05-18
[CI]: Make Omni CI use colocated router topology

Ratish1:
## Motivation

Move Omni model CI onto the router-managed two-worker topology and make Qwen3-Omni speech routing request-aware.

The router topology is the CI shape we need to validate: one client endpoint, one router, and two one-GPU workers. While moving Qwen3-Omni benchmarks through that path, text-output requests exposed an existing issue in the speech pipeline: even when the request only asked for text output, the thinker still activated the speech-generation branch and sent hidden-state streams toward `talker_ar` / `code2wav`.

This PR fixes that at the request-routing layer. The static Qwen3-Omni speech DAG still declares the full multimodal topology, but each request resolves the active output branch from `request.metadata["output_modalities"]`.

## Architecture

CI benchmark clients continue to call one OpenAI-compatible endpoint:

```mermaid
flowchart LR
  Pytest[pytest benchmark] --> Router[Omni router]
  Router --> WorkerA[Qwen3-Omni worker 0]
  Router --> WorkerB[Qwen3-Omni worker 1]
  WorkerA --> GPU0[GPU 0]
  WorkerB --> GPU1[GPU 1]
```

Inside each Qwen3-Omni speech worker, input modality handling is unchanged. Text, image, video, and audio inputs still flow through the appropriate preprocessing / encoder stages. The new routing decision is about requested output modalities only:

| Input modality | Output modality metadata | Active output path |
| --- | --- | --- |
| text | `["text"]` | `thinker -> decode` |
| image | `["text"]` | image input path, then `thinker -> decode` |
| video | `["text"]` | video/image input path, then `thinker -> decode` |
| audio | `["text"]` | audio input path, then `thinker -> decode` |
| text / image / video / audio | `["audio"]` | `thinker -> decode + talker_ar -> code2wav` |
| text / image / video / audio | `["text", "audio"]` | `thinker -> decode + talker_ar -> code2wav` |
| text / image / video / audio | missing or malformed | speech-compatible default: `thinker -> decode + talker_ar -> code2wav` |

The missing / malformed default intentionally preserves existing speech-pipeline behavior for callers that do not set `output_modalities`.

## Modifications

- Added request-aware routing hooks to the generic pipeline runtime:
  - `StageConfig.route_fn`
  - `StageConfig.stream_done_to_fn`
  - `PipelineConfig.terminal_stages_fn`
- Wired Qwen3-Omni speech config to use those hooks:
  - explicit text-output requests route thinker results only to `decode`
[...]

PR #507, opened 2026-05-21
[Pipeline] Add in-process stage handoff for colocated stages

Hayden727:
## Summary

Related to #498.

This PR adds a framework-level in-process stage handoff path for logical stages that share one OS process. It keeps logical stage boundaries and scheduler inbox semantics intact, but bypasses the relay/control-plane round trip when sender and receiver are local to the same process.

For Qwen3-Omni speech-colocated workers, the text-side path now shares one `text_side` process:

`preprocessing -> image_encoder/audio_encoder -> mm_aggregate -> thinker -> decode`

`talker_ar` and `code2wav` remain separate processes.

## Root Cause

After #477, MMSU moved from the old text-side topology to router-managed speech-colocated workers. Even for text-output requests, the active main path still crossed several process boundaries and paid repeated payload serialization, relay transfer, and control-plane handoff costs. For short MMSU requests, that fixed overhead outweighed the savings from skipping inactive output stages.

## Changes

- Add `InProcessStageDispatcher` for same-process payload, stream chunk, and stream signal delivery.
- Wire `Stage` to prefer local delivery and fall back to existing relay/control-plane behavior for remote targets.
- Install the dispatcher automatically for multi-stage process groups in `stage_process.py`.
- Update Qwen3-Omni speech-colocated process grouping so the text-side path can use local handoff.
- Add unit coverage for local payload handoff, fan-in, streaming, abort drop behavior, and remote fallback.
- Update developer and Qwen/router docs.

## Validation

Local smoke:

- `python -m py_compile sglang_omni/pipeline/local_transport.py sglang_omni/pipeline/stage/runtime.py sglang_omni/pipeline/stage_process.py sglang_omni/models/qwen3_omni/config.py tests/unit_test/pipeline/test_local_transport.py tests/unit_test/qwen3_omni/test_colocation_config.py tests/unit_test/qwen3_omni/test_config_manager.py`
- `python -m pytest tests/unit_test/qwen3_omni/test_colocation_config.py` -> 13 passed
- `git diff --check`

NVIDIA H200:

- `python -m pytest tests/unit_test/pipeline/test_local_transport.py tests/unit_test/qwen3_omni/test_colocation_config.py tests/unit_test/qwen3_omni/test_config_manager.py -q` -> 29 passed, 2 warnings in 11.68s
- `PYTHONPATH=<checkout> HF_DATASETS_CACHE=<writable-cache> python -m pytest tests/test_model/test_qwen3_omni_mmsu_ci.py -s -x -v --tb=short`

MMSU CI result on this PR:

- Throughput: 73.614 QPS
- Latency: mean 0.216s, p95 0.252s
- Requests: 2000/2000 successful, 0 failed
- Accuracy: 69.90% (1398/2000), below the current CI threshold of 70.65%

Same MMSU CI command on current `main`:

- Throughput: 12.23 QPS
- Latency: mean 1.302s, p95 4.155s
- Requests: 2000/2000 successful, 0 failed
- Accuracy: 70.65% (1413/2000)

The MMSU throughput/latency regression from #498 is removed on this PR: the fused text-side topology is about 6.0x faster than current `main` on the same hardware and command. The remaining blocker is the MMSU accuracy gate: the faster topology changes thinker batching shape/order, and the prediction diff is distributed across tasks rather than concentrated in one data-path failure. I am investigating whether the fast topology should tune thinker batching or carry an MMSU-specific server config calibration before marking the PR ready.

### DIFF
--- a/examples/configs/qwen3_omni_mmsu.yaml
+++ b/examples/configs/qwen3_omni_mmsu.yaml
@@ -1,0 +1,26 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# MMSU is audio-in / text-out with short requests. Use the text-output Qwen3
+# pipeline so the whole text path stays in one worker process and avoids the
+# speech-output stages that MMSU never uses.
+
+config_cls: Qwen3OmniPipelineConfig
+name: qwen3-omni-mmsu
+model_path: Qwen/Qwen3-Omni-30B-A3B-Instruct
+relay_backend: shm
+
+stage_overrides:
+  image_encoder:
+    runtime:
+      resources:
+        total_gpu_memory_fraction: 0.025
+
+  audio_encoder:
+    runtime:
+      resources:
+        total_gpu_memory_fraction: 0.025
+
+  thinker:
+    runtime:
+      resources:
+        total_gpu_memory_fraction: 0.75
--- a/tests/test_model/conftest.py
+++ b/tests/test_model/conftest.py
@@ -36,6 +36,7 @@ QWEN3_OMNI_ROUTER_WAIT_TIMEOUT = 180
 QWEN3_OMNI_COLOCATED_WORKER_ARGS = (
     "--config examples/configs/qwen3_omni_colocated.yaml --colocate"
 )
+QWEN3_OMNI_MMSU_WORKER_ARGS = "--config examples/configs/qwen3_omni_mmsu.yaml"
 QWEN3_OMNI_VIDEO_WORKER_ARGS = (
     f"{QWEN3_OMNI_COLOCATED_WORKER_ARGS} "
     "--stages.0.factory-args.thinker-max-seq-len 32768 "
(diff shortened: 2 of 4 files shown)


Hayden727:
Update after the full CI MMSU run:

**Root cause**
- The in-process local handoff itself did not corrupt tensors. I checked this with a clone-all local handoff ablation, and accuracy did not recover.
- The regression came from removing relay/control-plane pacing. With colocated in-process stages, the Qwen3-Omni thinker was allowed to run with much larger effective prefill/decode concurrency. MMSU greedy decoding is batch-shape sensitive, so the changed batching caused small argmax flips.

**Fix**
- Added a framework-level typed runtime knob: `runtime.sglang_server_args.max_running_requests`.
- The resolver maps this into SGLang server args and rejects conflicts with `server_args_overrides`.
- Qwen3-Omni colocated config now sets the thinker to `max_running_requests: 1`, restoring the intended decode shape while keeping the in-process transport/topology design.

**Validation on NVIDIA H200**
- Unit tests: `47 passed`
- CI MMSU, full 2-worker run: `PASS`
  - Successful: `2000/2000`
  - Correct: `1415/2000`
  - Accuracy: `70.75%`
  - Mean latency: `0.448s`
  - P95 latency: `0.460s`
  - Throughput: `35.60 req/s`
  - Worker health: `2/2 healthy`, `0 failed`

This removes the observed accuracy regression and keeps throughput well above the same-H200 baseline we measured before.

PR #509, opened 2026-05-21
[RFC][Feat] Remove TCP control-plane endpoints

JingwenGu0829:
## Summary

Remove TCP endpoint support from the pipeline control plane and make IPC the only supported transport.

The control plane is used for local pipeline coordination messages, such as completion and abort endpoints between processes. Since these messages are intra-node process communication, IPC sockets are the expected path and TCP does not add a useful deployment mode here.

The existing TCP path was also not well implemented: endpoint reservation/allocation could diverge from the endpoint that later gets bound, so the port-based path was fragile. Instead of keeping an unused and unreliable alternate transport, this PR deletes the TCP branch and simplifies endpoint allocation around IPC.

### DIFF
--- a/sglang_omni/pipeline/runtime_config.py
+++ b/sglang_omni/pipeline/runtime_config.py
@@ -6,7 +6,6 @@ from __future__ import annotations
 import logging
 import re
 import shutil
-import socket
 import tempfile
 from dataclasses import dataclass
 from pathlib import Path
@@ -56,15 +55,12 @@ class PipelineRuntimePrep:
     endpoints: dict[str, str]
     placement_plan: StagePlacementPlan
     process_plan: ProcessTopologyPlan
-    runtime_dir: IpcRuntimeDir | None
+    runtime_dir: IpcRuntimeDir
     runtime_dir_created_here: bool
 
 
-def create_ipc_runtime_dir(config: PipelineConfig) -> IpcRuntimeDir | None:
+def create_ipc_runtime_dir(config: PipelineConfig) -> IpcRuntimeDir:
     """Create a per-run IPC namespace for one pipeline instance."""
-    if config.endpoints.scheme != "ipc":
-        return None
-
     base_root = Path(config.endpoints.base_path)
     base_root.mkdir(parents=True, exist_ok=True)
 
@@ -82,11 +78,11 @@ def prepare_pipeline_runtime(
 ) -> PipelineRuntimePrep:
     """Prepare fused stages, endpoint allocation, and process topology."""
     runtime_dir = ipc_runtime_dir
-    created_runtime_dir = None
     if runtime_dir is None:
         runtime_dir = create_ipc_runtime_dir(config)
-        created_runtime_dir = runtime_dir
-    runtime_dir_created_here = created_runtime_dir is not None
+        runtime_dir_created_here = True
+    else:
+        runtime_dir_created_here = False
 
     try:
         stages_cfg, name_map, entry_stage = config.apply_fusion()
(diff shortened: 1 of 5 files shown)


sglang_omni/pipeline/runtime_config.py
```diff
@@ -194,31 +173,11 @@ def allocate_endpoints(
...
-        endpoints.setdefault("abort", f"ipc://{base_dir}/abort.sock")
-        for stage in stages:
-            endpoints[f"stage_{stage.name}"] = (
-                f"ipc://{base_dir}/stage_{stage.name}.sock"
-            )
-        return endpoints
-
-    if config.endpoints.scheme == "tcp":
-        needed = 2 + len(stages)
-        ports = find_free_tcp_ports(config.endpoints.base_port, needed)
```
zhaochenyang20, line 211: when a user no longer sets `endpoints.scheme` but still supplies public config values such as `completion_endpoint="tcp://127.0.0.1:..."` or `abort_endpoint="tcp://127.0.0.1:..."`, the current code copies those strings into `endpoints` before the default IPC allocation runs. I verified this locally with a minimal config: `prepare_pipeline_runtime` returns `completion=tcp://127.0.0.1:19001` and `abort=tcp://127.0.0.1:19002`, while only the stage endpoint is `ipc://...`. That means the PR removes the `EndpointsConfig.scheme="tcp"` entry point, but not TCP control-plane support itself; the coordinator still binds those TCP endpoints in `CoordinatorControlPlane.start()`, and stages still connect to them. For the IPC-only invariant claimed by this change, this is a behavioral gap rather than a naming issue. Users or older configs can still start TCP completion/abort control-plane sockets, preserving the port-binding, port-exposure, and split-transport risks that this PR is supposed to remove.

Tighten `completion_endpoint` and `abort_endpoint` at the config boundary. If these overrides still need to exist, allow only `ipc://` and raise a clear `ValueError` for `tcp://`, `inproc://`, or strings without a scheme; if they are just leftover escape hatches from the old TCP branch, remove both public fields and make `allocate_endpoints` always derive completion, abort, and stage endpoints from the runtime directory. Whichever design is chosen, add a unit test that proves explicit `tcp://` completion and abort endpoints are rejected.

sglang_omni/pipeline/runtime_config.py
```diff
@@ -62,9 +61,6 @@ class PipelineRuntimePrep:
 
 def create_ipc_runtime_dir(config: PipelineConfig) -> IpcRuntimeDir | None:
     """Create a per-run IPC namespace for one pipeline instance."""
-    if config.endpoints.scheme != "ipc":
```
zhaochenyang20, line 65: Tighten `completion_endpoint` and `abort_endpoint` at the config boundary. If these overrides still need to exist, allow only `ipc://` and raise a clear `ValueError` for `tcp://`, `inproc://`, or strings without a scheme; if they are just leftover escape hatches from the old TCP branch, remove both public fields and make `allocate_endpoints` always derive completion, abort, and stage endpoints from the runtime directory. Whichever design is chosen, add a unit test that proves explicit `tcp://` completion and abort endpoints are rejected.

Change `create_ipc_runtime_dir` to return `IpcRuntimeDir`, make `PipelineRuntimePrep.runtime_dir` non-optional, and have `prepare_pipeline_runtime` pass `runtime_dir.path` directly after either creating or receiving a caller-owned runtime directory. `MultiProcessPipelineRunner._ipc_runtime_dir` can remain optional before startup and after shutdown, but the prepared `PipelineRuntimePrep` object should no longer represent a missing runtime directory. `allocate_endpoints` can also take a required `Path` for `ipc_base_dir`, so the IPC-only invariant is reflected in the types.

tests/unit_test/pipeline/test_ipc.py
zhaochenyang20, line 1: Add a negative pipeline config/runtime test that builds a minimal `PipelineConfig`, sets `completion_endpoint="tcp://127.0.0.1:19001"` and/or `abort_endpoint="tcp://127.0.0.1:19002"`, and asserts that config construction or `prepare_pipeline_runtime` raises a clear error. Keep a positive test showing that default endpoints and allowed explicit `ipc://` endpoints still work, so future endpoint override changes cannot bypass the removal of TCP support.

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

Issue #512, opened 2026-05-22
[Feat]  EncoderModelRunner: shared encoder runner template


JingwenGu0829:
## Motivation

Issue #493 tracks reusable templates for common Omni optimizations. Encoder execution is a good first subissue because image, video, and audio encoders already share a clear stage lifecycle, while their optimization logic is currently implemented inside model-specific stage factories.

Today Qwen3-Omni has encoder batching, cache lookup/store, duplicate cache-key dedup, output splitting, and skip handling in `qwen3_omni/stages.py`. Ming-Omni has separate image/audio encoder closures with the same stage IO shape. As more Omni models are added, this lifecycle should become a clean runner surface that model developers can adapt by implementing the tensor-specific parts.

The proposed abstraction is a small `EncoderModelRunner` with an explicit:

```text
pre -> forward -> post
```

flow. The runner owns the common stage mechanics. Each model runner owns its own tensor semantics.

## Current encoder mechanism

An encoder stage is a single-pass model stage:

1. Preprocessing writes `PipelineState.encoder_inputs[stage_name]`.
2. The encoder stage reads its stage input.
3. It handles `_skip` requests and optional `cache_key`.
4. It normalizes and batches model inputs.
5. It runs the model forward.
6. It splits model outputs back to each request.
7. It writes `PipelineState.encoder_outs[stage_name]` and `engine_outputs[stage_name]`.
8. The aggregate stage merges encoder outputs into thinker `model_inputs`.

The common lifecycle is stable, while the tensor rules differ by model and modality:

- vision encoders use patch tensors plus `grid_thw` metadata;
- video often shares the image tower but uses separate input/output keys;
- audio encoders use feature tensors plus masks or lengths;
- some encoders concatenate packed items;
- some encoders pad by time;
- some encoders unwrap and rewrap segments;
- some outputs include only embeddings;
- some outputs include grids, token counts, output lengths, or deepstack tensors.

## Proposed design

Add an `EncoderModelRunner` for single-pass encoder stages:

```python
class EncoderModelRunner:
    def execute(self, payloads: list[StagePayload]) -> list[StagePayload]:
        prepared = self.pre(payloads)
        raw_outputs = self.forward(prepared)
        return self.post(prepared, raw_outputs)
```

The base runner should be small. It should make the lifecycle clear and provide shared helpers for the mechanics that are already common:

- stage input parsing;
- `_skip` result propagation;
- cache lookup/store;
[...]

PR #523, opened 2026-05-23
[Feat][RFC] Add encoder model runner with Qwen vision CUDA graph

JingwenGu0829:
WIP


sglang_omni/model_runner/encoder_model_runner.py
```diff
@@ -0,0 +1,470 @@
...
+    def forward_cuda_graph(self, prepared: Any) -> Any:
+        graph_key = self.cuda_graph_key(prepared)
+        if graph_key is None:
+            return self.forward_eager(prepared)
+
+        return self.run_cuda_graph_piece(graph_key, prepared)
+
+    def run_cuda_graph_piece(self, graph_key: Any, prepared: Any) -> Any:
+        if graph_key not in self.cuda_graphs:
+            self.capture_cuda_graph(graph_key, prepared)
```
zhaochenyang20, line 255: every unseen `graph_key` goes straight into `capture_cuda_graph`, then remains permanently in `cuda_graphs`, `cuda_graph_static_inputs`, `cuda_graph_input_buffers`, `cuda_graph_metadata_buffers`, and `cuda_graph_output_buffers`; the Qwen runner also builds the key at `sglang_omni/models/qwen3_omni/encoder_model_runner.py:421` from exact hidden shape and exact `cu_seqlens` values. Online image/video requests vary by frame count, resolution, crop, and batch packing, so concurrent traffic will produce many different video shapes. This path will keep capturing new graphs and retaining their GPU buffers with no cap, eviction, bucket policy, or oversized eager fallback. The failure mode is not a one-time latency hit: encoder process memory grows monotonically with media shapes until OOM, or the service keeps paying graph-capture cost mid-flight. It also violates #512’s required lifecycle of selecting the smallest fitting graph budget and falling back to eager for oversized or unsupported inputs.


Change the graph key to a finite shape bucket instead of the exact media shape. Each bucket should keep fixed input/output buffers and copy the current request into those buffers before replay. Add hard limits for both bucket count and total graph-buffer bytes, and route inputs that exceed the limits or do not match the bucket policy to eager. The implementation should also record graph hits, misses, fallback reasons, and add tests proving that multiple distinct video grids do not grow beyond the configured bucket limit.

sglang_omni/models/qwen3_omni/encoder_model_runner.py
```diff
@@ -0,0 +1,682 @@
...
+    ) -> Any:
+        return (
+            "qwen3_omni_vision_blocks",
+            self._visual_attention_impl(),
+            str(hidden_states.dtype),
+            str(hidden_states.device),
+            tuple(hidden_states.shape),
+            tuple(int(value) for value in cu_seqlens.detach().cpu().tolist()),
+        )
+
```
zhaochenyang20: The current graph path runs `visual.patch_embed`, positional interpolation, and `cu_seqlens` construction, then builds the key with `cu_seqlens.detach().cpu().tolist()`. For flash attention this `cu_seqlens` tensor is on GPU, so every graph-eligible request forces GPU work back to the CPU control plane before replaying the block graph. That undermines the CUDA graph goal of reducing launch and synchronization overhead; for video batches, the just-finished patch embedding and `cu_seqlens` work are serialized behind a CPU wait before graph replay. The risk is longer encoder latency under long-video or high-concurrency traffic, and graph hits cannot deliver the expected throughput gain. This is related to the exact-key root cause above, but even when the graph cache hits, it still happens on every request dispatch path.

Build the graph signature from the original CPU-side `grid_thw` metadata or precomputed bucket metadata, rather than calling `.cpu().tolist()` after GPU preprocessing. If `cu_seqlens` must be compared, derive it from the CPU grid during bucket planning, so the replay path only performs tensor copies and graph replay.

sglang_omni/model_runner/encoder_model_runner.py
```diff
@@ -0,0 +1,470 @@
...
+    def post_single(self, request: Any, result: Any) -> Any:
+        del request
+        return result
+
+    def request_model_inputs(self, request: Any) -> dict[str, Any]:
+        return getattr(request, "model_inputs", {})
+
+    def request_cache_key(self, request: Any) -> str | None:
+        cache_key = getattr(request, "cache_key", None)
+        return str(cache_key) if cache_key is not None else None
```
zhaochenyang20, line 395: The new shared runner is a public surface future model authors will reuse, but it reads the required contract through `getattr(request, "model_inputs", {})`, `getattr(request, "cache_key", None)`, and `getattr(request, "skip_result", None)`. These defaults turn schema drift into quiet behavior changes: a misspelled `model_inputs` degrades to an empty `model()` call, a missing `cache_key` silently disables cache and dedup, and a missing `skip_result` can send a request that should be skipped into forward. The failure mode is that integration mistakes no longer fail at the request-builder boundary; they surface later as encoder latency, cache misses, or wrong forward calls. Qwen already returns a concrete `EncoderRequestData` dataclass, so the base runner does not need dynamic fallback. Keeping this in the shared class makes future model request builders harder for type checking and review to catch.

 Define a shared `EncoderRequest` dataclass or `Protocol` with explicit `model_inputs`, `cache_key`, and `skip_result` fields. The base runner should access those fields directly, or make `request_model_inputs`, `request_cache_key`, and `request_skip_result` abstract hooks that each model runner must implement. Missing fields should fail fast instead of returning `{}` or `None`.

tests/unit_test/qwen3_omni/test_encoder_model_runner.py
zhaochenyang20, line 1: should we turn encoder-CUDA Graph by default and cover it in end2end tests? No need to test it seperately.

zhaochenyang20 (changes requested):

sglang_omni/models/qwen3_omni/stages.py
zhaochenyang20, line 1: `Qwen3OmniImageEncoderModelRunner` now has `enable_cuda_graph`, `cuda_graph_token_budgets`, `cuda_graph_sequence_budgets`, `cuda_graph_max_sequence_token_budgets`, `cuda_graph_max_graphs`, and `cuda_graph_max_buffer_bytes`, but the production entry point still constructs it as `Qwen3OmniImageEncoderModelRunner(model=model, cache=cache)`. Once encoder CUDA graph is enabled by default on a CUDA vision tower, deployment configs, example YAML files, and `runtime_overrides` cannot disable graph or tune graph count and buffer bytes for H20/H200, colocation, or video-heavy workloads. If the new graph path hits a specific `flash_attn`, driver, memory-budget, or model-version issue, users have to patch code instead of restoring eager encoder behavior through configuration the way thinker and talker can with `disable_cuda_graph`.

We still need to give the user one endpoint to let them disable CUDA Graph, but we should turn on CUDA Graph by default.

sglang_omni/models/qwen3_omni/encoder_model_runner.py
```diff
@@ -0,0 +1,992 @@
...
+        grid_thw: torch.Tensor,
+        graph_key: QwenVisionCudaGraphBudget,
+        device: torch.device,
+    ) -> torch.Tensor:
+        if graph_key.mode == "exact":
+            cu_seqlens = torch.tensor(
+                graph_key.exact_cu_seqlens,
+                dtype=torch.int32,
+                device="cpu",
+            )
```
zhaochenyang20: The latest commit adds both `cuda_graph_stats.fallbacks` and `cuda_graph_fallback_reasons`, but not every eager fallback updates both. `no_fitting_budget`, `grid_not_cpu`, `pixel_grid_mismatch`, and `unaligned_token_count` record a reason in `_select_visual_graph_budget` and return `None`; the base `forward` then sees `can_run_cuda_graph(prepared)` as false and goes straight to `forward_eager`, never entering the only branch in `Qwen3OmniImageEncoderModelRunner.forward_cuda_graph` that increments `cuda_graph_stats.fallbacks`. Real oversized or unsupported media can therefore appear in the reason dictionary while the aggregate fallback counter remains 0, so graph hit/miss/fallback telemetry cannot reliably answer how many requests fell back from graph to eager.

Do not let `None` mean both “no media” and “should fallback.” Change Qwen graph planning to return an explicit plan such as `GraphPlan(kind="graph" | "eager", reason=...)`, or have `can_run_cuda_graph` call a read-only planning function and consistently increment both `cuda_graph_stats.fallbacks` and the reason whenever graph-eligible input selects eager. Add tests for `no_fitting_budget`, `grid_not_cpu`, `max_graphs`, and `max_buffer_bytes`, and assert that the aggregate fallback counter matches the sum of reason counts.

tests/unit_test/qwen3_omni/test_encoder_model_runner.py
zhaochenyang20, line 1: The new `test_qwen_image_encoder_uses_finite_video_graph_budgets` proves that two video grids choose the same budget, and the real CUDA replay test can prove same-shape replay correctness when CUDA and `flash_attn` are present. However, the tests still do not cover the factory-level switch, `cuda_graph_max_graphs` overflow, `cuda_graph_max_buffer_bytes` overflow, `no_fitting_budget` fallback, or consistency between `cuda_graph_stats.fallbacks` and `cuda_graph_fallback_reasons`. Because encoder graph is enabled by default in the production factory, these policy branches should not rely only on a GPU CI same-shape replay test; otherwise the current CI can miss both the missing config switch and undercounted fallback counter.

zhaochenyang20 (changes requested):

sglang_omni/models/qwen3_omni/encoder_model_runner.py
```diff
@@ -0,0 +1,1049 @@
...
+
+        model_inputs = prepared["model_inputs"]
+        graph_keys: list[tuple[str, QwenVisionCudaGraphBudget]] = []
+        image_budget = self._select_visual_graph_budget(
+            model_inputs.get("pixel_values"),
+            model_inputs.get("image_grid_thw"),
+        )
+        if image_budget is not None:
+            graph_keys.append(("image", image_budget))
+        elif isinstance(model_inputs.get("pixel_values"), torch.Tensor):
```
Hayden727, line 248: [P1] Fallback telemetry still undercounts these eager paths. When `_select_visual_graph_budget()` records `no_fitting_budget`, `grid_not_cpu`, `pixel_grid_mismatch`, or `unaligned_token_count`, this `return None` makes the base `forward()` skip `forward_cuda_graph()`, so `cuda_graph_stats.fallbacks` is never incremented. I reproduced on `ion8-h200-omni` at `c66f2d37`: `no_fitting_budget` and `grid_not_cpu` each produced `stats_fallbacks=0` while `cuda_graph_fallback_reasons` had one entry. Please make graph planning explicit, or otherwise route every graph-eligible eager fallback through one path that increments both the aggregate counter and the reason counter, then add tests asserting the two stay consistent.

sglang_omni/models/qwen3_omni/stages.py
```diff
@@ -792,21 +199,13 @@ def create_image_encoder_executor(
         max_bytes=QWEN3_ENCODER_CACHE_MAX_BYTES,
         cache_device="cpu",
     )
+    runner = Qwen3OmniImageEncoderModelRunner(model=model, cache=cache)
```
Hayden727: [P1] The production factory still does not expose the image-encoder CUDA graph controls. `Qwen3OmniImageEncoderModelRunner` now has `enable_cuda_graph`, bucket lists, `cuda_graph_max_graphs`, and `cuda_graph_max_buffer_bytes`, but this entry point only accepts `model_path`, `device`, and `dtype` and constructs the runner with defaults. I verified `runtime_overrides={"image_encoder": {"enable_cuda_graph": false, "cuda_graph_max_graphs": 1}}` resolves into kwargs that `create_image_encoder_executor()` does not accept, so users cannot disable or tune default-on encoder graph without patching code. Please thread these options through the stage factory/config surface, and add a factory-level test that an override reaches the runner.

sglang_omni/models/qwen3_omni/encoder_model_runner.py
```diff
@@ -0,0 +1,1245 @@
...
+                        end=token_end,
+                    )
+                )
+                video_row_cursor = row_end
+                video_token_cursor = token_end
+            results.append(stage_result)
+
+        return results
+
+    def _visual_cuda_graph_supported(self) -> bool:
```
yxs, line 418: `_visual_cuda_graph_supported` returns False on several paths (missing visual attrs, training mode, attention impl not in Flash3/Triton, non-CUDA visual) without calling `_record_visual_graph_fallback`.

`cuda_graph_key` then early-returns None and the base `forward` falls back to eager, so `cuda_graph_stats.fallbacks` stays at 0 alongside `captures == 0` — the stats are indistinguishable from "no request fit a budget" when actually graph never attempted to start. Under default config this trips B200 (`VisionFlash4Attention` is the SM10 default in upstream `vision.py`), AMD `aiter_attn`, NPU, XPU, and any non-CUDA `visual.device`.

sglang_omni/models/qwen3_omni/encoder_model_runner.py
```diff
@@ -0,0 +1,1245 @@
...
+
+    def _native_visual_graph_available(self) -> bool:
+        graph_runners = getattr(self.model.visual, "graph_runners", None)
+        if graph_runners is None or not hasattr(graph_runners, "run"):
+            return False
+
+        try:
+            from sglang.srt.server_args import get_global_server_args
+
+            backend = get_global_server_args().mm_attention_backend
```
yxs, line 753: `_native_visual_graph_available` requires `mm_attention_backend in ("fa3","triton_attn")`. In a fresh standalone encoder process on CUDA, `_ensure_sglang_vision_runtime` writes `mm_attention_backend=None` on first init and `create_image_encoder_executor` doesn't expose the backend as a factory kwarg, so the global stays at `None` and `_can_select_native_visual_budget` always returns False. 

The native fast path is unreachable in that default deployment. `cuda_graph_stats.captures > 0` still fires from the budget-mode path so the metric makes it look like the native fast path is active when it isn't, every single-image request silently pays budget-mode padding overhead instead of upstream's exact-shape graph.

sglang_omni/models/qwen3_omni/encoder_model_runner.py
```diff
@@ -0,0 +1,1245 @@
...
+                f"visual graph budget cannot pad {actual_tokens} tokens to "
+                f"{graph_key.token_budget} with {graph_key.sequence_budget} "
+                "cu_seqlens slots"
+            )
+        if write_idx <= graph_key.sequence_budget:
+            cu_seqlens[write_idx:] = graph_key.token_budget
+
+        return cu_seqlens.to(device=device, non_blocking=True)
+
+    def _estimate_visual_graph_buffer_bytes(
```
yxs, line 630: `_estimate_visual_graph_buffer_bytes` counts hidden / position / output / cu but misses two real allocations, and `cuda_graph_max_buffer_bytes` ends up behaving inconsistently across modes.

sglang_omni/models/qwen3_omni/encoder_model_runner.py
```diff
@@ -0,0 +1,1245 @@
...
+
+        return outputs
+
+    def prepare_cuda_graph_capture(
+        self,
+        graph_key: Any,
+        prepared: dict[str, Any],
+    ) -> dict[str, Any]:
+        static_prepared = self._copy_visual_graph_buffers(graph_key, prepared)
+        with torch.no_grad(), envs.SGLANG_VIT_ENABLE_CUDA_GRAPH.override(True):
```
yxs, line 317: Upstream `EnvField.override` (in `sglang/srt/environ.py`) has no `try/finally` around its `yield`.
If anything inside this `with envs.SGLANG_VIT_ENABLE_CUDA_GRAPH.override(True):` block raises — capture OOM, kernel JIT failure, shape mismatch in warmup — the env stays at `"True"` for the rest of the process. Every subsequent eager `VisionTritonAttention.forward` then hits the cuda-graph branch and raises `RuntimeError("output_ws should be prepared for cuda-graph mode")` because eager callers don't pass `output_ws` — the error surfaces far from the original failure and looks unrelated.

yxs:
need consider framework and scale, more models. so, will spend some time, try to improve the whole design before move forward in next few days.

PR #524, opened 2026-05-23
[Bugfix] Qwen3-Omni talker multimodal prefill

### LINKED ISSUE #483: [Bug] Qwen3-Omni: Possible unused thinker -> talker data: stream hidden/embed and multimodal prompt features
## Summary

When woking on #472 I noticed two suspicious data-flow patterns in the current Qwen3-Omni talker path. They may be either correctness bugs or unnecessary data-transfer overhead, depending on the intended design.

## 1. Assistant stream `embed` / `layer_hidden` may not be semantically used

The thinker streams assistant-token `embed` / `layer_hidden` to `talker_ar`, but the talker seems to mostly use `token_id` and reconstruct the text-conditioning row from the thinker's embedding table.

Local diagnostic signal: after zeroing the streamed assistant `embed` / `layer_hidden`, full EN seedtts benchmark and MMMU talker CI metrics did not show an obvious change.

Question: if streamed assistant `embed` / `layer_hidden` are supposed to condition the talker, should this path be fixed to actually consume them? If they are not supposed to be used, should we stop sending them, or at least avoid sending large hidden tensors, to reduce relay overhead?

## 2. Multimodal prompt features may not enter talker prefill `input_embeds`

The prompt reconstruction path appears to write multimodal features into `prompt_embed`, but then zeroes `prompt_hidden` at the same positions:

```python
# components/talker_prefill.py
prompt_embed[mask] = feature_tensor
prompt_hidden[mask] = 0.0
```

Later, the user multimodal rows are built from `thinker_hidden`, not `thinker_embed`:

```python
# components/talker_input.py
result[multimodal_mask] = hidden_projection(thinker_hidden[multimodal_mask])
result[text_mask] = text_projection(thinker_embed[text_mask])
```

This makes it look like the actual multimodal feature values in `prompt_embed` may not affect the final talker prefill `input_embeds`.

End-to-end signal: running a small SeedTTS voice-clone request with Qwen3-Omni did not appear to clone the reference voice clearly, while S2-Pro did on the same reference/audio text. A simple repro is to run one SeedTTS voice-clone sample through Qwen3-Omni and S2-Pro and compare the generated wavs by listening.

Question: for multimodal prompt rows, should the talker prefill consume the actual multimodal features, a real thinker hidden state after multimodal injection, or is the current zero-hidden behavior intentional?

## Expected clarification

Please confirm the intended design for these two paths:

1. Assistant generated text conditioning: consume streamed `embed` / `layer_hidden`, or reconstruct from `token_id`?
2. Multimodal prompt conditioning: consume actual multimodal features / thinker hidden states, or intentionally use zero hidden rows?

If the current behavior is intentional, we should consider simplifying the transferred data. If it is not intentional, this likely needs a correctness fix.

Hayden727:
## Summary
- Preserve Qwen3-Omni prompt multimodal hidden states for talker `hidden_projection`.
- Carry thinker prefill prompt hidden rows through stream metadata, including prefix-cache offsets.
- Keep full single-request prefill hidden rows in the SGLang output processor instead of collapsing them to one row.

Fixes #483.

## Root Cause
The talker user-part path projects multimodal prompt positions with `hidden_projection(thinker_hidden[...])`. The V1 path had two gaps:

- `merge_prompt_modality` copied audio/image/video features into `prompt_embed`, but left the matching `prompt_hidden` rows zeroed.
- The captured thinker prompt hidden states were sliced down to one row for single-request prefill output, so talker prefill could not recover the real multimodal prompt hidden rows.

This PR keeps the chain intact without changing sampling or benchmark thresholds.

## Verification
- H200 container `sglang-omni-hayden`: `python -m pytest tests/unit_test/qwen3_omni/test_streaming.py tests/unit_test/qwen3_omni/test_talker.py -q` -> `69 passed, 2 warnings`.
- Local: `python -m py_compile sglang_omni/models/qwen3_omni/components/talker_prefill.py sglang_omni/models/qwen3_omni/request_builders.py sglang_omni/scheduling/sglang_backend/output_processor.py tests/unit_test/qwen3_omni/test_streaming.py tests/unit_test/qwen3_omni/test_talker.py`.
- Local: `git diff --check origin/main`.

## Benchmark Note
SeedTTS-50 single-worker quality run on H200 after the data-path fix:

- WER corpus: `0.0425531914893617`
- WER below-50 corpus: `0.014652014652014652`
- Speaker SIM mean: `3.2644491079449653`

The low SIM is left as a model-capability/benchmark-calibration question; this PR only fixes the confirmed serving data path.


sglang_omni/scheduling/sglang_backend/output_processor.py
```diff
@@ -192,21 +192,27 @@ def _slice_per_request_tensor(
...
+        reqs = getattr(batch_data, "reqs", None) if batch_data is not None else None
+        if reqs is not None:
+            num_requests = len(reqs)
+
+            lengths = [int(req.extend_input_len) for req in reqs]
+            total_tokens = sum(lengths)
+            if tensor.shape[0] == total_tokens:
+                if total_tokens == num_requests and all(
+                    length == 1 for length in lengths
+                ):
```
zhaochenyang20, line 212: after a prefix-cache hit, the current extend batch can contain only one uncached prompt row, and a batch of speech requests can have one uncached prompt row per request. `_slice_per_request_tensor` currently sees `tensor.shape[0] == total_tokens` and then `total_tokens == num_requests and all(length == 1)`, so it returns `tensor[request_index]` and squeezes the prefill row to 1D. Later, `_prefill_prompt_states` in `make_thinker_stream_output_builder()` only accepts `layer_hidden.ndim == 2 and layer_hidden.shape[0] > 1`, so that captured prompt row is not placed in `prefill_layer_hidden` metadata. The request then reaches `TalkerPrefillBuilder.apply_captured_prefill_prompt_states()` without captured hidden and falls back to the feature row written by `merge_prompt_modality`. The concrete impact is that the prefix-cache offset scenario this PR claims to cover still has a failure mode: prompts with exactly one uncached multimodal token can miss the real thinker hidden and produce incorrect voice-clone or multimodal prompt conditioning.

The slicing logic needs to distinguish extend/prefill batches from decode batches first. For extend batches, when `tensor.shape[0] == total_tokens`, return `tensor[start:end]` even when `end - start == 1`; keep it as a 2D row batch. Use `tensor[request_index]` only for decode or for tensors that are explicitly per-request. After the fix, add a regression test where `extend_input_len=1`, `prefix_indices` is non-empty, and the captured hidden dict has shape `[1, hidden]`; the stream metadata must contain `prefill_layer_hidden` with shape `[1, hidden]` and the correct `prefill_offset`, and talker prefill must write that row into the corresponding multimodal prompt position.

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

matchyc:
Hi, I'd like to know when this pr will be merged

PR #525, opened 2026-05-23
[Feature] Request-level profiler: timeline, stage breakdown, hop breakdown (#501)

### LINKED ISSUE #501: [Profiler] Standardize request-level latency timeline, stage breakdown, and hop breakdown
## Context

Before optimizing any hot path, we need a consistent profiling contract.

The current profiler on latest `main` already provides basic torch trace control:

- `/start_profile` and `/stop_profile` are mounted from `sglang_omni/serve/launcher.py`.
- `ProfilerControlClient` broadcasts profiler control messages to stages in `sglang_omni/profiler/profiler_control.py`.
- `TorchProfiler` exports Chrome traces in `sglang_omni/profiler/torch_profiler.py`.

This is useful for kernel / trace-level investigation, but it is not yet enough for request-level latency analysis. Without a standardized layer here, follow-up optimization work can easily chase the most visible point instead of the most expensive one.

## Proposed request-level events

For each request, record at least the following timestamps:

- request admission
- preprocessing start / end
- encoder start / end
- aggregate ready
- thinker prefill start
- thinker first token
- first stream chunk sent
- talker request build start / end
- talker prefill start
- first code chunk
- code2wav first audio
- terminal response

Each event should carry a consistent shape, for example:

- `request_id`
- `stage`
- `event_name`
- `timestamp_ns`
- optional metadata such as modality, chunk id, token count, audio duration, queue state, or error

## Required views

Build three standard views from the same event stream:

1. End-to-end latency timeline
2. Stage breakdown
3. Hop breakdown

The timeline should make a single request easy to reconstruct. The stage breakdown should aggregate time spent in preprocessing, encoders, aggregate, thinker, talker, code2wav, and terminal paths. The hop breakdown should expose queueing, IPC, relay, and stream path overhead between stages.

## Implementation notes

- Add instrumentation around coordinator admission, stage runtime execution, stream chunk routing, and completion handling.
- Clearly separate streaming milestones: first token, first stream chunk sent, and terminal response.
- Clearly separate speech pipeline milestones: thinker, talker request build, talker prefill, first code chunk, and code2wav first audio.
- Emit a structured artifact, such as JSONL events, then derive timeline / stage / hop reports from it.
- Keep the existing torch profiler functionality, but connect it to request-level profiling via shared `run_id` / `request_id` metadata where possible.

## Acceptance criteria

- A single request can be reconstructed as a complete end-to-end timeline.
- Batch benchmark runs can produce aggregated stage breakdowns.
- Stage-to-stage hop costs are visible and can be compared across requests.
- The profiler output helps prioritize optimization by cost, not just by trace visibility.

Hayden727:
## Summary

Implements the request-level profiler called for in #501 in full. The
profiler runs alongside the existing torch trace and shares its `run_id`.

- **Event recorder** (`sglang_omni/profiler/event_recorder.py`) —
  process-local JSONL sink with a thread-safe singleton and a tolerant
  `emit(...)` helper. Each process writes
  `<event_dir>/events_<stage>_<pid>.jsonl`. Emit errors are swallowed so
  profiling never breaks serving.
- **Views** (`sglang_omni/profiler/views.py`) — derives the three required
  reports from the JSONL stream: per-request timeline, stage breakdown
  (open/close interval durations with p50/p95/max), and hop breakdown
  (`stage_hop_sent → stage_input_received` plus stream-chunk variants
  matched by `chunk_id`).
- **CLI** (`python -m sglang_omni.profiler <event_dir>`) — JSON or table
  output for quick inspection.
- **Instrumentation** — all #501 milestones are covered:

| #501 milestone | Concrete event |
|---|---|
| request admission | `request_admission` |
| preprocessing start / end | `preprocess_start` / `preprocess_end` |
| encoder start / end | `encoder_start` / `encoder_end` (with modality, batch_size) |
| aggregate ready | `stage_aggregate_ready` |
| thinker prefill start | `scheduler_prefill_start` (stage=thinker) |
| thinker first token | `stage_first_stream_chunk_sent` (stage=thinker) |
| first stream chunk sent | `stage_first_stream_chunk_sent` (to coordinator) |
| talker request build start / end | `scheduler_request_build_start` / `_end` (stage=talker) |
| talker prefill start | `scheduler_prefill_start` (stage=talker) |
| first code chunk | `stage_first_stream_chunk_sent` (stage=talker) |
| code2wav first audio | `code2wav_first_audio` |
| terminal response | `terminal_response` |

The stage name on each event carries the layer info, so the same
`scheduler_prefill_start` becomes "thinker prefill start" vs "talker
prefill start" depending on which scheduler process emitted it.

- **Control plane** — extends `ProfilerStartMessage` with `event_dir`
  and `enable_torch` (defaults keep old behavior), broadcasts via the
  existing `ProfilerControlClient`, and toggles the recorder from
  Stage's `_on_profiler_start` / `_on_profiler_stop`. New endpoints
  `/start_request_profile` and `/stop_request_profile` enable
  JSONL-only profiling without paying for a torch kernel trace.

## Usage

```bash
curl -X POST http://localhost:8000/start_request_profile \
     -d '{"run_id":"demo","event_dir":"/tmp/demo/events"}'
# … run benchmark traffic …
curl -X POST http://localhost:8000/stop_request_profile -d '{"run_id":"demo"}'
python -m sglang_omni.profiler /tmp/demo/events --format table
```

## Test plan
- [x] 16 unit tests under `tests/unit_test/profiler/`
  (`pytest tests/unit_test/profiler/ -q`) — pass.
- [x] All five #501-required stage intervals are exercised end-to-end by
  `test_stage_breakdown_covers_preprocess_encoder_and_prefill`.
- [x] Backwards compat: `ProfilerStartMessage.from_dict` accepts the
  pre-existing payload shape (no `event_dir`/`enable_torch`).
- [x] CLI smoke test on a synthetic event directory renders the table
  and JSON reports.
- [ ] On-host validation: run Qwen3-Omni MMMU benchmark with the
  profiler enabled and confirm reports identify the expected hot stages
  and hops (artifacts uploaded to HaydenCC/profiles on HF).

Closes #501.

### DIFF
--- a/sglang_omni/pipeline/stage/runtime.py
+++ b/sglang_omni/pipeline/stage/runtime.py
@@ -128,6 +128,11 @@ class Stage:
         self._active_requests: set[str] = set()
         self._stream_queue: StreamQueue | None = None
         self._stream_chunk_counters: dict[tuple[str, str], int] = {}
+        # Per-request marker: have we already emitted the first stream-chunk
+        # event for this request from this stage? Used to derive request-level
+        # milestones (thinker first token, talker first code chunk, etc.)
+        # without coupling Stage to model-specific naming.
+        self._first_stream_chunk_seen: set[str] = set()
         self._scheduler_thread: threading.Thread | None = None
         self._loop: asyncio.AbstractEventLoop | None = None
         self._scheduler_crash_error: BaseException | None = None
@@ -717,6 +722,17 @@ class Stage:
         key = (request_id, target)
         chunk_id = self._stream_chunk_counters.get(key, 0)
         self._stream_chunk_counters[key] = chunk_id + 1
+        chunk_modality = (
+            metadata.get("modality") if isinstance(metadata, dict) else None
+        )
+        if request_id not in self._first_stream_chunk_seen:
+            self._first_stream_chunk_seen.add(request_id)
+            _emit_event(
+                request_id=request_id,
+                stage=self.name,
+                event_name="stage_first_stream_chunk_sent",
+                metadata={"to_stage": target, "modality": chunk_modality},
+            )
         _emit_event(
             request_id=request_id,
             stage=self.name,
@@ -724,9 +740,7 @@ class Stage:
             metadata={
                 "to_stage": target,
                 "chunk_id": chunk_id,
-                "modality": (
-                    metadata.get("modality") if isinstance(metadata, dict) else None
-                ),
+                "modality": chunk_modality,
             },
         )
         await relay_io.send_stream_chunk(
@@ -769,6 +783,14 @@ class Stage:
             stage_name=self.name,
             modality=modality,
         )
+        if request_id not in self._first_stream_chunk_seen:
+            self._first_stream_chunk_seen.add(request_id)
+            _emit_event(
+                request_id=request_id,
+                stage=self.name,
+                event_name="stage_first_stream_chunk_sent",
+                metadata={"to_stage": "coordinator", "modality": modality},
+            )
         _emit_event(
             request_id=request_id,
             stage=self.name,
@@ -802,6 +824,7 @@ class Stage:
         ]
         for key in stale_keys:
             self._stream_chunk_counters.pop(key, None)
+        self._first_stream_chunk_seen.discard(request_id)
 
     async def _handle_scheduler_crash(self, exc: BaseException) -> None:
         if self._scheduler_crash_error is not None:
--- a/sglang_omni/pipeline/stage/runtime.py
+++ b/sglang_omni/pipeline/stage/runtime.py
@@ -129,10 +129,7 @@ class Stage:
         self._active_requests: set[str] = set()
         self._stream_queue: StreamQueue | None = None
         self._stream_chunk_counters: dict[tuple[str, str], int] = {}
-        # Per-request marker: have we already emitted the first stream-chunk
-        # event for this request from this stage? Used to derive request-level
-        # milestones (thinker first token, talker first code chunk, etc.)
-        # without coupling Stage to model-specific naming.
+        # Per-request: did we already emit the first stream-chunk event?
         self._first_stream_chunk_seen: set[str] = set()
         self._scheduler_thread: threading.Thread | None = None
         self._loop: asyncio.AbstractEventLoop | None = None
@@ -150,13 +147,8 @@ class Stage:
         if self.scheduler is not None:
 
             def _run_scheduler():
-                # Bind this stage's name to the active-stage context for
-                # this thread (and any asyncio.to_thread / run_in_executor
-                # descendants that copy contextvars). Callsites that pass
-                # ``stage=None`` to ``emit()`` — preprocessor, encoder
-                # callables, OmniScheduler internals, code2wav — will get
-                # the correct stage name even when this stage shares its
-                # process with siblings.
+                # Active-stage binding so ``emit(stage=None)`` from
+                # scheduler-thread descendants resolves to this stage.
                 _set_active_stage(self.name)
                 try:
                     if self.gpu_id is not None:
@@ -905,9 +897,7 @@ class Stage:
                 )
 
     def _on_profiler_stop(self, msg: ProfilerStopMessage) -> None:
-        # msg.run_id is None when the HTTP caller didn't specify a run id;
-        # in that case stop whatever's active. Otherwise only stop when
-        # the run id matches.
+        # run_id=None is a wildcard (stop whatever's active).
         if TorchProfiler.is_active() and (
             msg.run_id is None or TorchProfiler.get_active_run_id() == msg.run_id
         ):
--- a/sglang_omni/pipeline/stage/runtime.py
+++ b/sglang_omni/pipeline/stage/runtime.py
@@ -20,6 +20,8 @@ from sglang_omni.pipeline import relay_io
 from sglang_omni.pipeline.stage.input import DirectInput, InputHandler
 from sglang_omni.pipeline.stage.stream_queue import StreamItem, StreamQueue
 from sglang_omni.pipeline.tp_control import TPLeaderFanout
+from sglang_omni.profiler.event_recorder import emit as _emit_event
+from sglang_omni.profiler.event_recorder import get_recorder as _get_recorder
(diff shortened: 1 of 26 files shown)


sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -501,6 +513,11 @@ def process_input_requests(self, recv_reqs):
             self._initialize_request_stream_state(req_data, payload)
             if req_id in self._aborted_request_ids:
                 continue
+            _emit_event(
+                request_id=req_id,
+                stage=None,
+                event_name="scheduler_prefill_start",
+            )
```
zhaochenyang20, line 509: the current code emits `scheduler_prefill_start` before `self.waiting_queue.append(req)`. That is not prefill start; it is “request build completed and about to enter the scheduler queue.” `views.py` then aggregates `scheduler_prefill_start->scheduler_first_emit` and `scheduler_prefill_start->stage_first_stream_chunk_sent` as the thinker first-token / talker first-code-chunk timing. Under load, those intervals include scheduler queueing time and report it as prefill/TTFT cost

Rename the current event to something like `scheduler_queue_enter`, or keep it as a separate queueing event. The real `scheduler_prefill_start` should be emitted once per request when the batch has actually been selected and is about to execute the first prefill/extend forward. This class already wraps `run_batch`, so the fix can inspect `batch.reqs` and the prefill/extend state, for example `batch.is_prefill_only`, `batch.is_extend_in_batch`, or an equivalent local condition, and use a per-request set to avoid re-emitting during decode steps. The regression test should enqueue a request whose execution is delayed and assert that queue-enter and prefill-start are distinct events.

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -291,6 +317,12 @@ async def _on_stream_chunk(self, msg: DataReadyMessage) -> None:
             await self._discard_stream_chunk_data(msg)
             return
         self._active_requests.add(request_id)
+        _emit_event(
+            request_id=request_id,
+            stage=self.name,
+            event_name="stage_stream_chunk_received",
+            metadata={"from_stage": msg.from_stage, "chunk_id": msg.chunk_id},
+        )
```
zhaochenyang20: this profiling semantic because the real stage-to-stage stream path is: the sender writes the blob or IPC handle, sends the `DataReadyMessage`, and the receiver then deserializes it or calls `relay_io.read_blob` plus `_read_chunk_metadata`. The current receiver emits `stage_stream_chunk_received` as soon as `_on_stream_chunk` starts, before `_deserialize_ipc_chunk` or `relay_io.read_blob` runs. `hop_breakdown()` pairs `stage_stream_chunk_sent` with `stage_stream_chunk_received`, so the report measures only control-message arrival at the receiver. The heavy receiver read, CUDA IPC unpickle, and metadata tensor relay reads are excluded from hop cost, even though that is exactly the IPC/relay overhead #501 asks the profiler to expose.

If this event means “the receiver has a chunk ready for its scheduler,” move `stage_stream_chunk_received` after IPC deserialization and after the cross-GPU `read_blob` plus metadata restore, immediately before `_route_stream_item_or_fail`. If the control-message arrival time is also useful, add a separate `stage_stream_chunk_control_received` event and do not reuse the close event that `hop_breakdown()` pairs. The test should not only synthesize view-layer events; it should either exercise a runtime helper or at least build an event sequence where control receipt and payload materialization are separate, then assert that hop breakdown uses the payload-materialized timestamp.

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -708,6 +784,20 @@ async def _send_stream_to_coordinator(
...
+            _emit_event(
+                request_id=request_id,
+                stage=self.name,
+                event_name="stage_first_stream_chunk_sent",
+                metadata={"to_stage": "coordinator", "modality": modality},
+            )
+        _emit_event(
+            request_id=request_id,
+            stage=self.name,
+            event_name="stage_stream_chunk_sent",
```
zhaochenyang20, line 810: A terminal stage emits `stage_stream_chunk_sent` when it streams to the coordinator, but the metadata contains only `to_stage="coordinator"` and `modality`; there is no `chunk_id`. The other side, `Coordinator._handle_stream`, records `coordinator_stream_received`, not the `stage_stream_chunk_received` event that `hop_breakdown()` recognizes, and it also has no chunk id. The final terminal audio/text stream hop before the client is therefore absent from hop breakdown. The timeline can show the raw events, but the batch report cannot compare the “terminal stage -> coordinator” stream path against “thinker -> talker” or “talker -> code2wav.”

Assign the same kind of chunk counter to the terminal-to-coordinator stream path, add `chunk_id` to `StreamMessage` if needed, and make the coordinator emit a receiver event that the views layer can pair; alternatively, explicitly extend `hop_breakdown()` to pair `stage_stream_chunk_sent(to_stage=coordinator)` with `coordinator_stream_received`. This path needs a regression test where one terminal stage sends two chunks and the report contains `src=<terminal>, dst=coordinator, kind=stream_chunk, count=2`.

sglang_omni/serve/launcher.py
zhaochenyang20: `/start_profile` exposes the new `enable_torch` field, and the documentation says callers can set it to false to skip the torch trace and record only request events. But the handler parses `trace_path_template` before it considers `enable_torch`; when `SGLANG_TORCH_PROFILER_DIR` is unset and the request does not include `trace_path_template`, it returns HTTP 400 before `enable_torch` is used. A client sending `{"enable_torch": false, "event_dir": "/tmp/events"}` still fails because a torch trace template is missing, which contradicts both the field semantics and the docs.

Branch on `req.enable_torch` first. Require or generate `trace_path_template` only when `enable_torch=True`; when `enable_torch=False`, allow `tpl=""` or `None`, but require `event_dir` or a profiler-dir-derived event directory. Add a FastAPI/TestClient regression test for `/start_profile` covering `enable_torch=false + event_dir + profiler_dir=None`, asserting that it succeeds and broadcasts `enable_torch=False`, while the old trace-template validation remains in place for `enable_torch=True`.

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

PR #526, opened 2026-05-23
[Feature] Migrate SeedTTS benchmark datasets to Arrow/Parquet

junliu-mde:
## Motivation

File-by-file downloads of `seed-tts-eval` can hit Hugging Face rate limits, and moving SeedTTS to Arrow/Parquet dataset repos also aligns it with datasets like MMMU and MMSU under a more consistent benchmark data-loading path.

## Modifications

- migrate SeedTTS benchmark loading to Arrow/Parquet dataset repos via `datasets`
- update benchmark defaults, docs, helper scripts, and CI fixtures to use dataset repo ids
- pre-warm benchmark datasets through the Hugging Face datasets cache
- keep local `meta.lst` support for SeedTTS
- keep `prepare --dataset mmmu` working by pre-warming all MMMU configs
- apply `max_samples` before SeedTTS audio staging
- add regression tests for the dataset loading paths

## Related Issues

N/A

## Accuracy Test

N/A. This PR only changes benchmark dataset loading.

Validation:
- `python -m pytest tests/unit_test/benchmarks/test_dataset_regressions.py -q`
- `pre-commit run --files benchmarks/dataset/prepare.py benchmarks/dataset/seedtts.py tests/unit_test/benchmarks/test_dataset_regressions.py`

## Benchmark & Profiling

N/A

## Checklist

- [x] Add unit tests.
- [x] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.

## CI

### DIFF
--- a/.claude/skills/tune-ci-thresholds/models/qwen3-omni-v1/config.yaml
+++ b/.claude/skills/tune-ci-thresholds/models/qwen3-omni-v1/config.yaml
@@ -5,7 +5,7 @@ hf_model_id: "Qwen/Qwen3-Omni-30B-A3B-Instruct"
 hf_datasets:
   - "zhaochenyang20/mmsu-ci-2000"
   - "zhaochenyang20/mmmu-ci-50"
-  - "zhaochenyang20/seed-tts-eval-50"
+  - "zhaochenyang20/seed-tts-eval-50-arrow"
   - "zhaochenyang20/Video_MME_ci"
   - "zhaochenyang20/Video_AMME_ci"
 default_venv_python:
--- a/.claude/skills/tune-ci-thresholds/models/qwen3-omni-v1/config.yaml
+++ b/.claude/skills/tune-ci-thresholds/models/qwen3-omni-v1/config.yaml
@@ -37,9 +37,7 @@ gpus_per_test:
 test_globs:
   - "tests/test_model/test_qwen3_omni_*_ci.py"
 # Per-test extra env vars. Key is the test filename (basename).
-extra_env:
-  test_qwen3_omni_tts_ci.py:
-    SGLANG_SEEDTTS50_DIR: "/github/home/seedtts-50"
+extra_env: {}
 # Stage-key base derivation: strip prefix + strip suffix from the test stem.
 stage_base_strip_prefix: "test_qwen3_omni_"
 stage_base_strip_suffix: "_ci"
--- a/benchmarks/dataset/seedtts.py
+++ b/benchmarks/dataset/seedtts.py
@@ -1,16 +1,19 @@
 # SPDX-License-Identifier: Apache-2.0
 """SeedTTS dataset loader.
 
-Loads samples from HuggingFace Arrow/Parquet repos (e.g.
-``zhaochenyang20/seed-tts-eval-50-arrow``).  Audio bytes are staged to a
-process-scoped temporary directory so that downstream consumers (which
-expect filesystem paths) work unchanged.
+Supports two source formats:
+- Local ``meta.lst`` files (pipe-delimited: ``id|ref_text|ref_audio_path|target_text``)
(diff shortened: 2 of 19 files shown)


benchmarks/dataset/seedtts.py
```diff
@@ -43,6 +79,76 @@ def load_seedtts_samples(
...
+        rel = row["ref_audio_path"]
+        audio = row["ref_audio"] or {}
+        audio_bytes = audio.get("bytes")
+        if not audio_bytes:
+            raise ValueError(
+                f"Empty audio bytes for {repo_id}/{split}/{row['sample_id']}"
+            )
+
+        wav_path = tmpdir / rel
+        if rel not in written:
```
zhaochenyang20: when a caller passes an arbitrary Hugging Face dataset repo id, the current code joins the row's `ref_audio_path` directly onto `tmpdir`, then creates parent directories and writes bytes on lines 133-134. `Path(tmpdir) / rel` does not guarantee that the result remains under `tmpdir`; if the remote dataset uses an absolute path or a path containing `..`, the benchmark process writes audio bytes outside the staging directory and can overwrite any file the current user can write. This PR changes the default SeedTTS source from a locally controlled `meta.lst` to a remote Arrow repo, so this file-write boundary becomes a new security and data-corruption risk.


Validate `ref_audio_path` as untrusted input before writing. Reject absolute paths, empty paths, paths containing `..`, and Windows drive or anchor forms; after resolution, the target path must still be under `tmpdir.resolve()`. Add a unit test with a fake dataset row such as `../escape.wav` and an absolute path, assert that the loader raises `ValueError`, and assert that no file is written outside the staging directory. Legitimate paths such as `prompt-wavs/foo.wav` can still preserve their subdirectory layout.

.claude/skills/tune-ci-thresholds/models/qwen3-omni-v1/config.yaml
zhaochenyang20: The migration does not close the full configuration surface. `qwen3-omni-v1/config.yaml` still lists `zhaochenyang20/seed-tts-eval-50` in `hf_datasets`, and the same issue exists in `.claude/skills/tune-ci-thresholds/models/s2-pro-v1/config.yaml:18`. `tune.py precheck` checks every entry in `cfg["hf_datasets"]` for a local cache and prints a `huggingface-cli download ... --repo-type dataset` remediation command when one is missing; therefore a direct precheck for the migrated code still asks maintainers to cache the old non-Arrow SeedTTS repo. `tune.py run` can derive selected-stage repos from `benchmarks.dataset.prepare.DATASETS`, but that only covers the run path; the config files remain a contradictory source of truth. The practical impact is that calibration and threshold-tuning flows keep prewarming the old dataset, reintroduce the file-level download and cache-drift risk this PR is trying to remove, and fail precheck in environments that no longer have the old cache.

Change both `hf_datasets` entries to `zhaochenyang20/seed-tts-eval-50-arrow`, and in the same PR run the corresponding `tune.py precheck` or add a lightweight config test asserting that all SeedTTS cache and precheck configuration references only `*-arrow` repos. That keeps the helper tooling, CI fixtures, and benchmark defaults pointed at the same data source.

benchmarks/tasks/tts.py
```diff
@@ -333,9 +334,10 @@ def run_seedtts_similarity(
         generated: list[dict] = json.load(f)
     logger.info(f"Loaded {len(generated)} entries from {generated_path}")
 
+    split = getattr(config, "lang", "en")
```
zhaochenyang20: Use `split = config.lang` directly.

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

PR #451, opened 2026-05-16
[Feature] Support Qwen3-TTS and Voxtral-TTS following S2 Pro integration pattern

GaokaiZhang:
[Feature] Support Qwen3-TTS and Voxtral-TTS following S2 Pro integration pattern

## Motivation

This adds native `/v1/audio/speech` serving for the two model families requested
in #444:

- `mistralai/Voxtral-4B-TTS-2603`
- `Qwen/Qwen3-TTS-12Hz-0.6B-Base`
- `Qwen/Qwen3-TTS-12Hz-1.7B-Base`

The implementation follows the existing S2-Pro-style staged TTS path rather
than adding a separate serving surface.

## Modifications

- **Qwen3-TTS Base**: adds `sglang_omni.models.qwen3_tts`, with preprocessing,
  voice-clone generation, and tokenizer/vocoder stages. It supports
  `ref_audio` / `ref_text` and `references[0]`, uses `bfloat16`,
  `language=auto`, `max_new_tokens=2048`, and leaves checkpoint sampling
  defaults intact unless the request explicitly overrides them.

- **Voxtral-TTS**: repairs the current Voxtral `StageConfig` wiring and serves
  `mistralai/Voxtral-4B-TTS-2603` through `/v1/audio/speech`. The default voice
  is `cheerful_female`; the default decode limit is `max_new_tokens=4096`.

- **Serving/config plumbing**: updates `/v1/audio/speech` request mapping,
  model-path config resolution, `sgl-omni config view/export`, HF/Mistral config
  helpers, example YAML configs, and docs.

- **Dependencies/tests**: adds `qwen-tts==0.1.1` and no-GPU unit tests for
  config resolution, request mapping, defaults, and pipeline payloads.

Scheduler note: per
https://github.com/sgl-project/sglang-omni/issues/444#issuecomment-4463287744,
this PR does not add a FishScheduler-style model-specific scheduler. Qwen3-TTS
and Voxtral-TTS currently run as regular staged pipelines backed by
`SimpleScheduler` callables. If maintainers want the autoregressive stages moved
onto `OmniScheduler`, that should be handled as a follow-up or called out before
merge.

## Related Issues

Closes #444

## Accuracy Test

No-GPU checks passed:

```bash
git diff --check upstream/main..HEAD
git diff --name-only upstream/main..HEAD -- '*.py' | xargs python -m py_compile
python -m pytest tests/unit_test/qwen3_tts tests/unit_test/voxtral_tts
[...]

Issue #520, opened 2026-05-23
[Perf] Enable torch.compile + CUDA Graph for Qwen3-TTS and Voxtral-TTS AR backbones


zhaochenyang20:
## Motivation

PR #451 just landed Qwen3-TTS and Voxtral-TTS, but as called out during review both models currently return `can_run_cuda_graph=False` from their custom model runners and skip torch.compile entirely. CUDA graph init is also disabled in `create_sglang_tts_engine_executor` to avoid initialization cost on a path that wasn't going to use it anyway. The AR backbones — which dominate latency by definition (memory-bound, per-frame sequential) — are therefore running fully eager.

We've already proven the optimization pattern on S2-Pro (~2× throughput from torch.compile + CUDA graph). The pattern should transfer cleanly:

- **Accuracy is structurally invariant.** Graph capture and torch.compile don't change the math; they change kernel dispatch. Any accuracy regression beyond noise is a bug, not a tradeoff.
- **The baseline is well-measured.** PR #451 produced full SeedTTS EN/ZH numbers with reproductions from three independent contributors. Any improvement is directly attributable.

This is a net-positive-with-bounded-risk task and a good fit for a follow-up PR.

## Scope

Two models, two model runners:
- `sglang_omni/models/qwen3_tts/model_runner.py`
- `sglang_omni/models/voxtral_tts/model_runner.py`

For each:
1. Route decode through SGLang's graph-capable forward path instead of calling `self.model(...)` eagerly
2. Wire torch.compile on the AR backbone forward
3. Flip `can_run_cuda_graph=True` on `GenerationBatchResult` in steady-state decode
4. Re-enable graph capture in `create_sglang_tts_engine_executor`

## Baseline (from PR #451 final results, H200, c=16/shard, 4-GPU sharded)

| Model | Split | WER/CER | RTF mean | qps |
|---|---|---|---|---|
| Qwen3-TTS 12Hz 0.6B Base | EN | 1.20% | 1.5253 | 9.488 |
| Qwen3-TTS 12Hz 0.6B Base | ZH | 1.07% | 1.3762 | 7.448 |
| Qwen3-TTS 12Hz 1.7B Base | EN | 0.94% | 1.5622 | 9.396 |
| Qwen3-TTS 12Hz 1.7B Base | ZH | 1.01% | 1.4086 | 7.468 |
| Voxtral-4B-TTS-2603 | EN | 1.18% | 0.4112 | 25.612 |
[...]

PR #527, opened 2026-05-23
Enabling torch.compile + CUDA Graph for Qwen3-TTS and Voxtral-TTS AR backbones

GaokaiZhang:
Full H200 SeedTTS validation completed for the torch.compile + CUDA Graph PR.

Setup: 4-GPU sharded non-streaming runs with c=16 per shard, using `benchmarks/eval/benchmark_tts_seedtts.py` with generate-only followed by transcribe-only. Decode logs show CUDA graph replay (`cuda graph: True`) for both Qwen3-TTS and Voxtral.

| Model / config | Split | WER / CER | Samples | Failed / skipped | Throughput | rtf_mean | Delta vs #451 baseline |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | EN | WER 1.13% | 1088/1088 | 0 / 0 | 10.684 req/s | 1.3823 | qps +12.6%, RTF -9.4% |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | ZH | CER 1.08% | 2020/2020 | 0 / 0 | 8.312 req/s | 1.2516 | qps +11.6%, RTF -9.1% |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | EN | WER 0.95% | 1088/1088 | 0 / 0 | 10.548 req/s | 1.4211 | qps +12.3%, RTF -9.0% |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | ZH | CER 0.98% | 2020/2020 | 0 / 0 | 8.328 req/s | 1.2711 | qps +11.5%, RTF -9.8% |
| mistralai/Voxtral-4B-TTS-2603, no-ref, cheerful_female | EN | WER 1.26% | 1088/1088 | 0 / 0 | 30.956 req/s | 0.3434 | qps +20.9%, RTF -16.5% |
| mistralai/Voxtral-4B-TTS-2603, no-ref, cheerful_female | ZH | CER 150.56% | 2020/2020 | 0 / 0 | 14.272 req/s | 0.3066 | RTF -18.6%; ZH is not acceptance-gated |

Acceptance-gated rows pass: Qwen3-TTS 0.6B/1.7B EN+ZH and Voxtral EN all improve qps and RTF versus the #451 H200 baseline while staying within the requested WER/CER bounds. Voxtral ZH is included for continuity only; accuracy remains non-comparable because Voxtral TTS does not document Chinese support.

Qwen3-TTS 0.6B EN needed an isolated rerun because two earlier attempts hit one max-length repetition outlier each, matching the stochastic error mode before. The final acceptance row above has no max-length audio outlier; max generated audio duration was 9.12s.

Checks passed:

- `python -m py_compile` on touched Qwen3-TTS/Voxtral files and tests
- `git diff --check`
- `/data/wenyao/sglang-omni/.venv/bin/python -m pytest -q tests/unit_test/qwen3_tts tests/unit_test/voxtral_tts` (`29 passed, 2 warnings`)
- `PYTHONPATH=/home/sglang-omni/gaokaiz/sglang-omni-pref520 /data/wenyao/sglang-omni/.venv/bin/python -m benchmarks.eval.benchmark_tts_seedtts --help`

### DIFF
--- a/sglang_omni/models/qwen3_tts/model_runner.py
+++ b/sglang_omni/models/qwen3_tts/model_runner.py
@@ -40,9 +40,9 @@ class Qwen3TTSModelRunner(ModelRunner):
         schedule_batch: Any,
         requests: list,
     ) -> GenerationBatchResult | None:
-        del forward_batch, schedule_batch
+        del schedule_batch
         self.model.prepare_decode_buffers(requests)
-        self._write_feedback_buffers(requests)
+        self._write_feedback_buffers(forward_batch, requests)
         return None
 
     def post_prefill(
@@ -170,22 +170,47 @@ class Qwen3TTSModelRunner(ModelRunner):
 
         raise RuntimeError("Qwen3-TTS subtalker sampling requires semantic positions")
 
-    def _write_feedback_buffers(self, requests: list) -> None:
+    def _write_feedback_buffers(self, forward_batch: Any, requests: list) -> None:
         batch_size = len(requests)
-        feedback_buffer = self.model._feedback_buffer
-        feedback_mask = self.model._feedback_mask
-        feedback_mask[:batch_size] = False
+        if batch_size == 0:
+            return
+        decode_feedback_embedding = self.model._decode_feedback_embedding
+        input_ids = forward_batch.input_ids
+        if input_ids.numel() < batch_size:
+            raise RuntimeError(
+                "Qwen3-TTS decode input_ids must contain one row id per request"
+            )
+        if batch_size > decode_feedback_embedding.num_embeddings:
+            raise RuntimeError(
+                "Qwen3-TTS decode batch exceeds staged feedback embedding rows"
+            )
+        row_ids = torch.arange(
+            batch_size,
+            device=input_ids.device,
+            dtype=input_ids.dtype,
+        )
+        rows = []
 
         for row_idx, sched_req in enumerate(requests):
             combined = QwenTalkerModelRunner._take_next_decode_input_embed(
                 sched_req=sched_req,
-                device=feedback_buffer.device,
-                dtype=feedback_buffer.dtype,
+                device=decode_feedback_embedding.weight.device,
+                dtype=decode_feedback_embedding.weight.dtype,
             )
             if combined is None:
-                continue
-            feedback_buffer[row_idx].copy_(combined)
-            feedback_mask[row_idx] = True
+                token_id = input_ids[row_idx : row_idx + 1].to(
+                    device=decode_feedback_embedding.weight.device
+                )
+                combined = self.model.get_input_embeddings()(token_id).reshape(-1)
+            rows.append(combined)
+        stacked = torch.stack(rows, dim=0).to(
+            device=decode_feedback_embedding.weight.device,
+            dtype=decode_feedback_embedding.weight.dtype,
+        )
+        with torch.no_grad():
+            decode_feedback_embedding.weight[:batch_size].copy_(stacked)
+        # During graph decode, input_ids carries staged embedding row ids.
+        input_ids[:batch_size].copy_(row_ids)
 
     def _build_prefill_input_embeds(
         self,
--- a/sglang_omni/models/qwen3_tts/request_builders.py
+++ b/sglang_omni/models/qwen3_tts/request_builders.py
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Request mapping helpers for Qwen3-TTS Base."""
+"""Request mapping helpers for Qwen3-TTS."""
 
 from __future__ import annotations
 
@@ -18,6 +18,10 @@ from sglang_omni.proto import StagePayload
 from sglang_omni.scheduling.sglang_backend import SGLangARRequestData
 
 QWEN3_TTS_DEFAULT_MAX_NEW_TOKENS = 2048
+QWEN3_TTS_TASK_BASE = "Base"
+QWEN3_TTS_TASK_CUSTOM_VOICE = "CustomVoice"
+QWEN3_TTS_TASK_VOICE_DESIGN = "VoiceDesign"
+QWEN3_TTS_DEFAULT_CUSTOM_VOICE = "Vivian"
 _QWEN3_TTS_PREPARED_MARKER = "_qwen3_tts_prepared_request"
 
 _GENERATION_FIELDS = (
@@ -184,23 +188,80 @@ def build_qwen3_tts_state(payload: StagePayload) -> Qwen3TTSState:
         tts_params = {}
 
     text, references = normalize_qwen3_tts_inputs(inputs)
-    ref_audio, ref_text = resolve_voice_clone_reference(references, tts_params)
+    has_reference = has_voice_clone_reference(references, tts_params)
+    task_type_raw = tts_params.get("task_type") or params.get("task_type")
+    task_type_explicit = task_type_raw is not None and str(task_type_raw).strip() != ""
+    task_type = normalize_qwen3_tts_task_type(
+        task_type_raw, has_reference=has_reference
+    )
     language = normalize_language(tts_params.get("language") or params.get("language"))
-    x_vector_only_mode = resolve_x_vector_only_mode(
+    instructions = resolve_optional_text(
+        tts_params.get("instructions")
+        or tts_params.get("instruct")
+        or params.get("instructions")
+        or params.get("instruct")
+    )
+    ref_audio = None
+    ref_text = None
+    x_vector_only_mode = False
+    non_streaming_mode = resolve_non_streaming_mode(
+        task_type=task_type,
         params=params,
         tts_params=tts_params,
-        ref_text=ref_text,
     )
+    voice = normalize_qwen3_tts_voice(
+        tts_params.get("voice") or tts_params.get("speaker") or params.get("voice")
+    )
+
+    if task_type == QWEN3_TTS_TASK_BASE:
+        ref_audio, ref_text = resolve_voice_clone_reference(references, tts_params)
+        x_vector_only_mode = resolve_x_vector_only_mode(
+            params=params,
+            tts_params=tts_params,
+            ref_text=ref_text,
+        )
+        if not x_vector_only_mode and not ref_text:
+            raise ValueError(
+                "Qwen3-TTS Base requires non-empty ref_text unless "
+                "x_vector_only_mode is enabled"
+            )
+    elif task_type == QWEN3_TTS_TASK_CUSTOM_VOICE:
+        if has_param(tts_params, params, "ref_text") or references_contain_text(
+            references
+        ):
+            raise ValueError("Qwen3-TTS CustomVoice does not accept ref_text")
+        if has_param(tts_params, params, "x_vector_only_mode"):
+            raise ValueError("Qwen3-TTS CustomVoice does not accept x_vector_only_mode")
+        voice = voice or QWEN3_TTS_DEFAULT_CUSTOM_VOICE
+        non_streaming_mode = True
+    elif task_type == QWEN3_TTS_TASK_VOICE_DESIGN:
+        if has_param(tts_params, params, "ref_text") or references_contain_text(
+            references
+        ):
+            raise ValueError("Qwen3-TTS VoiceDesign does not accept ref_text")
+        if has_param(tts_params, params, "x_vector_only_mode"):
+            raise ValueError("Qwen3-TTS VoiceDesign does not accept x_vector_only_mode")
+        if not instructions:
+            raise ValueError("Qwen3-TTS VoiceDesign requires instructions")
+        voice = None
+        non_streaming_mode = True
+    else:
+        raise AssertionError(f"unhandled Qwen3-TTS task type: {task_type}")
+
     seed = tts_params["seed"] if "seed" in tts_params else params.get("seed")
     normalized_seed = _normalize_qwen3_tts_seed(seed) if seed is not None else None
 
     return Qwen3TTSState(
         text=text,
+        task_type=task_type,
+        task_type_explicit=task_type_explicit,
         language=language,
+        voice=voice,
+        instructions=instructions,
         ref_audio=ref_audio,
         ref_text=ref_text,
         x_vector_only_mode=x_vector_only_mode,
-        non_streaming_mode=bool(params.get("non_streaming_mode", False)),
+        non_streaming_mode=non_streaming_mode,
         generation_kwargs=build_generation_kwargs(params, tts_params=tts_params),
         seed=normalized_seed,
     )
@@ -238,6 +299,80 @@ def resolve_voice_clone_reference(
             "Qwen3-TTS Base requires reference audio via ref_audio or references[0].audio_path"
         )
     return ref_audio, str(ref_text) if ref_text is not None else None
+
+
+def has_voice_clone_reference(
+    references: list[dict[str, Any]],
+    tts_params: dict[str, Any],
+) -> bool:
+    if references:
+        reference = references[0]
+        if any(
+            reference.get(key) is not None
+            for key in ("audio_path", "ref_audio", "audio", "text")
+        ):
+            return True
+    return (
+        tts_params.get("ref_audio") is not None
+        or tts_params.get("ref_text") is not None
+    )
+
+
+def references_contain_text(references: list[dict[str, Any]]) -> bool:
+    return any(reference.get("text") is not None for reference in references)
+
+
+def normalize_qwen3_tts_task_type(
+    task_type: Any,
+    *,
+    has_reference: bool,
+) -> str:
+    if task_type is None or str(task_type).strip() == "":
+        return QWEN3_TTS_TASK_BASE if has_reference else QWEN3_TTS_TASK_CUSTOM_VOICE
+    normalized = str(task_type).replace("_", "").replace("-", "").lower()
+    if normalized == "base":
+        return QWEN3_TTS_TASK_BASE
+    if normalized == "customvoice":
+        return QWEN3_TTS_TASK_CUSTOM_VOICE
+    if normalized == "voicedesign":
+        return QWEN3_TTS_TASK_VOICE_DESIGN
+    raise ValueError(
+        "Qwen3-TTS task_type must be one of Base, CustomVoice, or VoiceDesign"
+    )
+
+
+def resolve_optional_text(value: Any) -> str | None:
+    if value is None:
+        return None
+    text = str(value).strip()
+    return text or None
+
+
+def normalize_qwen3_tts_voice(value: Any) -> str | None:
+    voice = resolve_optional_text(value)
+    if voice is None or voice.lower() == "default":
+        return None
+    return voice
+
+
+def has_param(
+    tts_params: dict[str, Any],
+    params: dict[str, Any],
+    name: str,
+) -> bool:
+    return name in tts_params or name in params
+
+
+def resolve_non_streaming_mode(
+    *,
+    task_type: str,
+    params: dict[str, Any],
+    tts_params: dict[str, Any],
+) -> bool:
+    for source in (params, tts_params):
+        if "non_streaming_mode" in source:
+            return bool(source["non_streaming_mode"])
+    return task_type in (QWEN3_TTS_TASK_CUSTOM_VOICE, QWEN3_TTS_TASK_VOICE_DESIGN)
 
 
 def normalize_language(language: Any) -> str:
@@ -321,14 +456,57 @@ def _build_qwen3_tts_pad_embed(model: Any) -> torch.Tensor:
         )
 
 
-def _prepare_qwen3_tts_request(
-    payload: StagePayload,
+def _build_instruct_id(wrapper: Any, instructions: str | None) -> torch.Tensor | None:
+    if not instructions:
+        return None
+    if hasattr(wrapper, "_build_instruct_text"):
+        instruct_text = wrapper._build_instruct_text(instructions)
+    else:
+        instruct_text = f"<|im_start|>user\n{instructions}<|im_end|>\n"
+    return wrapper._tokenize_texts([instruct_text])[0]
+
+
+def _normalized_model_type(model: Any) -> str:
+    model_type = getattr(model, "tts_model_type", None)
+    if model_type is None:
+        model_type = getattr(
+            getattr(model, "root_config", None), "tts_model_type", None
+        )
+    normalized = str(model_type or "base").replace("-", "_").strip().lower()
+    if normalized == "customvoice":
+        return "custom_voice"
+    if normalized == "voicedesign":
+        return "voice_design"
+    return normalized
+
+
+def _validate_qwen3_tts_model_task(model: Any, state: Qwen3TTSState) -> None:
+    model_type = _normalized_model_type(model)
+    if model_type == "base" and state.task_type != QWEN3_TTS_TASK_BASE:
+        if not state.task_type_explicit:
+            raise ValueError(
+                "Qwen3-TTS Base requires ref_audio or speaker_embedding; "
+                "text-only requests require a CustomVoice or VoiceDesign checkpoint"
+            )
+        raise ValueError(
+            f"Qwen3-TTS Base checkpoint does not support {state.task_type}"
+        )
+    if model_type == "custom_voice" and state.task_type != QWEN3_TTS_TASK_CUSTOM_VOICE:
+        raise ValueError(
+            f"Qwen3-TTS CustomVoice checkpoint does not support {state.task_type}"
+        )
+    if model_type == "voice_design" and state.task_type != QWEN3_TTS_TASK_VOICE_DESIGN:
+        raise ValueError(
+            f"Qwen3-TTS VoiceDesign checkpoint does not support {state.task_type}"
+        )
+
+
+def _prepare_qwen3_tts_base_request(
     *,
+    state: Qwen3TTSState,
     model: Any,
     wrapper: Any,
-) -> Qwen3TTSPreparedRequest:
-    state = build_qwen3_tts_state(payload)
-
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor | None]:
     with torch.no_grad():
         prompt_items = wrapper.create_voice_clone_prompt(
             ref_audio=state.ref_audio,
@@ -346,20 +524,98 @@ def _prepare_qwen3_tts_request(
         if ref_text
         else None
     )
-    gen_kwargs = wrapper._merge_generate_kwargs(**state.generation_kwargs)
+    instruct_id = _build_instruct_id(wrapper, state.instructions)
     with torch.no_grad():
-        (
-            input_embeds,
-            attention_mask,
-            trailing_text_hidden,
-            ref_code,
-        ) = model.build_voice_clone_inputs(
+        return model.build_voice_clone_inputs(
             input_id=input_id,
             ref_id=ref_id,
             voice_clone_prompt=voice_clone_prompt,
             language=state.language,
             non_streaming_mode=state.non_streaming_mode,
+            instruct_id=instruct_id,
         )
+
+
+def _prepare_qwen3_tts_custom_voice_request(
+    *,
+    state: Qwen3TTSState,
+    model: Any,
+    wrapper: Any,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor | None]:
+    input_id = wrapper._tokenize_texts([wrapper._build_assistant_text(state.text)])[0]
+    instruct_id = _build_instruct_id(wrapper, state.instructions)
+    with torch.no_grad():
+        return model.build_custom_voice_inputs(
+            input_id=input_id,
+            voice=state.voice or QWEN3_TTS_DEFAULT_CUSTOM_VOICE,
+            language=state.language,
+            non_streaming_mode=state.non_streaming_mode,
+            instruct_id=instruct_id,
+        )
+
+
+def _prepare_qwen3_tts_voice_design_request(
+    *,
+    state: Qwen3TTSState,
+    model: Any,
+    wrapper: Any,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor | None]:
+    input_id = wrapper._tokenize_texts([wrapper._build_assistant_text(state.text)])[0]
+    instruct_id = _build_instruct_id(wrapper, state.instructions)
+    with torch.no_grad():
+        return model.build_voice_design_inputs(
+            input_id=input_id,
+            language=state.language,
+            non_streaming_mode=state.non_streaming_mode,
+            instruct_id=instruct_id,
+        )
+
+
+def _prepare_qwen3_tts_request(
+    payload: StagePayload,
+    *,
+    model: Any,
+    wrapper: Any,
+) -> Qwen3TTSPreparedRequest:
+    state = build_qwen3_tts_state(payload)
+
+    _validate_qwen3_tts_model_task(model, state)
+    gen_kwargs = wrapper._merge_generate_kwargs(**state.generation_kwargs)
+    if state.task_type == QWEN3_TTS_TASK_BASE:
+        (
+            input_embeds,
+            attention_mask,
+            trailing_text_hidden,
+            ref_code,
+        ) = _prepare_qwen3_tts_base_request(
+            state=state,
+            model=model,
+            wrapper=wrapper,
+        )
+    elif state.task_type == QWEN3_TTS_TASK_CUSTOM_VOICE:
+        (
+            input_embeds,
+            attention_mask,
+            trailing_text_hidden,
+            ref_code,
+        ) = _prepare_qwen3_tts_custom_voice_request(
+            state=state,
+            model=model,
+            wrapper=wrapper,
+        )
+    elif state.task_type == QWEN3_TTS_TASK_VOICE_DESIGN:
+        (
+            input_embeds,
+            attention_mask,
+            trailing_text_hidden,
+            ref_code,
+        ) = _prepare_qwen3_tts_voice_design_request(
+            state=state,
+            model=model,
+            wrapper=wrapper,
+        )
+    else:
+        raise AssertionError(f"unhandled Qwen3-TTS task type: {state.task_type}")
 
     feedback_buffer = model.model._feedback_buffer
     prompt_input_embeds = (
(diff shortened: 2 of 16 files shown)


zhaochenyang20:
fix your lint with pre-commit

sglang_omni/models/qwen3_tts/model_runner.py
```diff
@@ -95,11 +95,43 @@ def _collect_codes(self, result: Any, schedule_batch: Any, requests: list) -> No
...
 
-    def _write_feedback_buffers(self, requests: list) -> None:
+    def _write_feedback_buffers(self, forward_batch: Any, requests: list) -> None:
         batch_size = len(requests)
         feedback_buffer = self.model._feedback_buffer
         feedback_mask = self.model._feedback_mask
         feedback_mask[:batch_size] = False
+        decode_feedback_embedding = getattr(
+            self.model, "_decode_feedback_embedding", None
+        )
```
zhaochenyang20: Remove the old fallback and access `self.model._decode_feedback_embedding` directly; if the attribute is missing, let the internal error surface immediately.

tests/unit_test/qwen3_tts/test_pipeline.py
zhaochenyang20, line 1: Preserve `enable_torch_compile=kwargs["enable_torch_compile"]` and `torch_compile_max_bs=kwargs["torch_compile_max_bs"]` on the fake `server_args`. In the Qwen3 test, monkeypatch `_compile_qwen3_tts_backbone` to record that it was called, then assert the call happened and that `scheduler.server_args.enable_torch_compile` was reset to `False` after the manual Qwen3 compile. In the Voxtral test, make the fake `init_device_graphs` assert that `server_args.enable_torch_compile is True`, and also check `torch_compile_max_bs == 16`.

zhaochenyang20 (changes requested):

GaokaiZhang:
Full H200 SeedTTS validation on `pref/520`, using `benchmarks/eval/benchmark_tts_seedtts.py` with non-streaming generation and `c=16` per shard. Qwen3-TTS used SeedTTS reference audio/text; Voxtral used `--no-ref-audio --voice cheerful_female`.


Run 1:

| Qwen/Qwen3-TTS-12Hz-0.6B-Base | EN | WER 1.12% | 1088/1088 | 0 / 0 | 10.420 req/s | 1.3890 |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | ZH | CER 1.05% | 2020/2020 | 0 / 0 | 8.236 req/s | 1.2440 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | EN | WER 0.97% | 1088/1088 | 0 / 0 | 10.652 req/s | 1.4135 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | ZH | CER 1.01% | 2020/2020 | 0 / 0 | 8.332 req/s | 1.2725 |
| mistralai/Voxtral-4B-TTS-2603 | EN | WER 1.31% | 1088/1088 | 0 / 0 | 30.696 req/s | 0.3430 |
| mistralai/Voxtral-4B-TTS-2603 | ZH | CER 151.67% | 2020/2020 | 0 / 0 | 19.900 req/s | 0.3012 |

Run 2:

| Qwen/Qwen3-TTS-12Hz-0.6B-Base | EN | WER 1.08% | 1088/1088 | 0 / 0 | 8.804 req/s | 1.6058 |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | ZH | CER 1.15% | 2020/2020 | 0 / 0 | 6.584 req/s | 1.2531 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | EN | WER 7.72% raw, 1.01% below-50 | 1088/1088 | 0 / 0 | 6.400 req/s | 1.4006 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | ZH | CER 0.92% | 2020/2020 | 0 / 0 | 8.360 req/s | 1.2668 |
| mistralai/Voxtral-4B-TTS-2603 | EN | WER 1.21% | 1088/1088 | 0 / 0 | 32.236 req/s | 0.3373 |
| mistralai/Voxtral-4B-TTS-2603 | ZH | CER 147.77% | 2020/2020 | 0 / 0 | 12.524 req/s | 0.3046 |

Targeted follow-up:

| Qwen/Qwen3-TTS-12Hz-0.6B-Base | EN | WER 9.46% raw, 1.12% below-50 | 1088/1088 | 0 / 0 | 7.460 req/s | 1.3680 |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | ZH | CER 1.10% | 2020/2020 | 0 / 0 | 6.552 req/s | 1.2423 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | EN | WER 0.95% | 1088/1088 | 0 / 0 | 10.608 req/s | 1.4226 |

Summary:

- Run 1 is a clean pass for all acceptance-gated rows: Qwen3-TTS 0.6B/1.7B EN+ZH and Voxtral EN are within accuracy bounds, with higher qps and lower RTF than the PR #451 baseline.
- Run 2 confirms accuracy for Qwen3 0.6B, Qwen3 1.7B ZH, and Voxtral EN. Qwen3 1.7B EN had one max-length repetition outlier (`n_above_50_pct_wer=1`); the targeted 1.7B EN rerun was clean.
- Qwen3 0.6B EN also showed one max-length repetition outlier in the targeted follow-up, matching the stochastic repetition behavior documented during PR #451. Its below-50 WER stayed near 1.1%.
- Qwen3 0.6B throughput was not consistently above baseline in run 2 or the targeted follow-up, although its run 1 row was faster. I would not claim a consistent 0.6B qps win from these two passes.
- Voxtral ZH is still non-comparable for quality because the model does not document Chinese support; it completed with zero failed/skipped samples.

0xjeffro:
Hi, I tried to reproduce on my side and ran into what looks like a possible regression specific to Qwen3-TTS when torch compile and CUDA Graph are enabled together.

With this PR applied, Qwen3-TTS-0.6B-Base and 1.7B-Base on both EN and ZH show a non-trivial rate of max-length generations in my env. To isolate: enabling torch.compile alone (CUDA Graph off) and enabling CUDA Graph alone both produce clean runs at expected accuracy. Only the combination of the two triggers the failure on my side.

For comparison, I ran the same combined torch.compile + CUDA Graph configuration on Voxtral-4B-TTS-2603 (EN) and S2-Pro (EN + ZH) and saw no equivalent regression, here is the table with SIM:

| Model | Lang | Samples | WER/CER (corpus) | n>50% | RTF mean | Throughput (req/s) | SIM |
| ----- | ---- | ------: | ---------------: | ----: | -------: | -----------------: | --: |
| Voxtral-4B-TTS-2603 | EN | 1088 | 1.17% WER | 0 | 0.503 | 21.958 | — |
| S2-Pro | EN | 1088 | 1.06% WER | 0 | 3.551 | 4.579 | 64.94 |
| S2-Pro | ZH | 2020 | 0.91% CER | 0 | 2.398 | 4.929 | 74.45 |


Note: I overwrote the failing Qwen3 logs during a re-run, so I don't have the exact WER / n>50% numbers to share at the moment. A third-party reproduction on Qwen3 would help confirm whether this is environment-specific or a real interaction issue.

GaokaiZhang:
Reran full SeedTTS validation 3 times on `pref/520`, using `benchmarks/eval/benchmark_tts_seedtts.py`, 4 shards, non-streaming, `c=16` per shard.

Normal rows are stable:

| Model / config | Split | WER / CER across 3 runs | Samples | Failed / skipped | Throughput range | rtf_mean range |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | ZH | CER 1.12-1.18% | 2020/2020 | 0 / 0 | 6.492-8.264 req/s | 1.2397-1.2660 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | EN | WER 0.95-1.02% | 1088/1088 | 0 / 0 | 9.804-10.564 req/s | 1.4144-1.4661 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | ZH | CER 0.96-1.03% | 2020/2020 | 0 / 0 | 8.040-8.340 req/s | 1.2660-1.2855 |
| mistralai/Voxtral-4B-TTS-2603 | EN | WER 1.21-1.29% | 1088/1088 | 0 / 0 | 29.012-31.800 req/s | 0.3401-0.3667 |

The reported Qwen3 0.6B EN regression reproduces, but it is outlier-driven:

| Run | Raw WER | Below-50 WER | n>50 | Failed / skipped | Throughput | rtf_mean |
| run 1 | 9.48% | 1.11% | 2 | 0 / 0 | 7.372 req/s | 1.3933 |
| run 2 | 10.29% | 1.22% | 1 | 0 / 0 | 7.188 req/s | 1.4250 |
| run 3 | 8.62% | 1.22% | 2 | 0 / 0 | 6.224 req/s | 1.4209 |

Outlier details:
- Run 1: one max-length 163.84s repetition loop (`common_voice_en_18313326-common_voice_en_18313328`, 9960% WER) plus one 57.14% WER sample.
- Run 2: one max-length 163.84s repetition loop (`common_voice_en_17861335-common_voice_en_17861337`, 10840% WER).
- Run 3: two max-length 163.84s repetition loops (`common_voice_en_17837104-common_voice_en_17837105`, 5950% WER; `common_voice_en_29495818-common_voice_en_29495898`, 4391.67% WER).

Interpretation: Qwen3 0.6B EN normal-sample quality remains aligned at about 1.1-1.2% WER, but raw corpus WER now fails because every run has 1-2 max-length repetition outliers. I would treat this as a real variance/regression issue to fix before using the 0.6B EN row as acceptance or performance evidence. Voxtral ZH remains non-comparable for quality because the model does not document Chinese support, but all requests completed with zero failed/skipped samples.

GaokaiZhang:
Full SeedTTS validation was rerun 5 times after rebasing `pref/520` onto current `main`.

Setup: `benchmarks/eval/benchmark_tts_seedtts.py`, non-streaming, 3 sharded workers, `c=16` per shard. Only 3 GPUs were available for this final pass, so aggregate throughput is not directly comparable to the earlier 4-GPU baseline; `rtf_mean` is the more useful speed sanity check here. Scheduler logs showed CUDA graph decode active (`cuda graph: True`). All rows below had zero failed requests and zero ASR skips.

Run 1

|---|---|---:|---:|---:|---:|---:|
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | EN | WER 1.04% | 1088/1088 | 0 / 0 | 8.170 req/s | 1.3781 |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | ZH | CER 1.12% | 2020/2020 | 0 / 0 | 6.099 req/s | 1.2429 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | EN | WER 0.97% | 1088/1088 | 0 / 0 | 8.054 req/s | 1.3967 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | ZH | CER 1.02% | 2020/2020 | 0 / 0 | 6.318 req/s | 1.2700 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | EN | WER 1.20% | 1088/1088 | 0 / 0 | 23.678 req/s | 0.3369 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | ZH | CER 151.74% (below-50 42.03%, n>50=2002) | 2020/2020 | 0 / 0 | 15.030 req/s | 0.3077 |

Run 2

|---|---|---:|---:|---:|---:|---:|
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | EN | WER 17.03% (below-50 1.12%, n>50=3) | 1088/1088 | 0 / 0 | 5.368 req/s | 1.3677 |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | ZH | CER 1.06% | 2020/2020 | 0 / 0 | 6.120 req/s | 1.2482 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | EN | WER 1.00% | 1088/1088 | 0 / 0 | 8.027 req/s | 1.4005 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | ZH | CER 0.96% | 2020/2020 | 0 / 0 | 6.348 req/s | 1.2656 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | EN | WER 1.28% | 1088/1088 | 0 / 0 | 21.721 req/s | 0.3446 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | ZH | CER 152.28% (below-50 41.69%, n>50=2006) | 2020/2020 | 0 / 0 | 15.055 req/s | 0.3049 |

Run 3

|---|---|---:|---:|---:|---:|---:|
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | EN | WER 1.14% | 1088/1088 | 0 / 0 | 8.027 req/s | 1.3804 |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | ZH | CER 1.27% (below-50 1.22%, n>50=1) | 2020/2020 | 0 / 0 | 5.175 req/s | 1.2379 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | EN | WER 1.01% | 1088/1088 | 0 / 0 | 8.006 req/s | 1.4030 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | ZH | CER 0.95% | 2020/2020 | 0 / 0 | 6.303 req/s | 1.2718 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | EN | WER 1.42% (below-50 1.22%, n>50=1) | 1088/1088 | 0 / 0 | 24.056 req/s | 0.3384 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | ZH | CER 148.13% (below-50 43.86%, n>50=2006) | 2020/2020 | 0 / 0 | 12.918 req/s | 0.3073 |

Run 4

|---|---|---:|---:|---:|---:|---:|
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | EN | WER 22.02% (below-50 1.17%, n>50=2) | 1088/1088 | 0 / 0 | 5.137 req/s | 1.3823 |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | ZH | CER 1.12% | 2020/2020 | 0 / 0 | 6.213 req/s | 1.2508 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | EN | WER 1.13% | 1088/1088 | 0 / 0 | 8.066 req/s | 1.4069 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | ZH | CER 1.00% | 2020/2020 | 0 / 0 | 6.144 req/s | 1.2940 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | EN | WER 1.25% | 1088/1088 | 0 / 0 | 23.507 req/s | 0.3394 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | ZH | CER 151.29% (below-50 42.66%, n>50=2009) | 2020/2020 | 0 / 0 | 14.908 req/s | 0.3134 |

Run 5

|---|---|---:|---:|---:|---:|---:|
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | EN | WER 18.55% (below-50 1.06%, n>50=2) | 1088/1088 | 0 / 0 | 5.170 req/s | 1.3857 |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | ZH | CER 1.23% (below-50 1.19%, n>50=1) | 2020/2020 | 0 / 0 | 5.466 req/s | 1.2660 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | EN | WER 1.06% | 1088/1088 | 0 / 0 | 7.916 req/s | 1.4098 |
| Qwen/Qwen3-TTS-12Hz-1.7B-Base | ZH | CER 1.03% | 2020/2020 | 0 / 0 | 6.276 req/s | 1.2787 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | EN | WER 1.16% | 1088/1088 | 0 / 0 | 23.645 req/s | 0.3388 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | ZH | CER 152.25% (below-50 41.83%, n>50=2009) | 2020/2020 | 0 / 0 | 10.643 req/s | 0.3130 |

Notes:

- Qwen3-TTS 0.6B EN reproduced the stochastic max-length repetition mode seen. Runs 2/4/5 had 2-3 samples above 50% WER, mostly hitting `completion_tokens=2048` and 163.84s audio. The below-50 corpus WER stayed normal at 1.04-1.17%.
- Qwen3-TTS 0.6B ZH had one above-50 sample in runs 3 and 5, but raw CER still stayed within the 1.3% bound.
- Qwen3-TTS 1.7B was stable in all five runs.
- Voxtral EN was stable except run 3, where one repeated sample pushed raw WER to 1.42%; below-50 WER was 1.22%.
- Voxtral ZH remains non-comparable as a quality signal because the model does not document Chinese support.

Accuracy is stable aside from the already documented stochastic repetition outliers.

GaokaiZhang:
Qwen3-TTS 0.6B EN outlier check for issue #520.

Facts from code:

- `benchmarks/eval/benchmark_tts_seedtts.py` sets `max_new_tokens=2048` by default for the benchmark.
- `sglang_omni/models/qwen3_tts/request_builders.py` passes that value into `SamplingParams(max_new_tokens=...)`.
- The Qwen3-TTS request builder also passes `stop_token_ids=[codec_eos_token_id]`.
- The benchmark does not set a seed in its generation kwargs.

Facts from current `pref/520` Qwen3-TTS 0.6B EN artifacts:

| Run | WER | below-50 WER | n>50 | max completion tokens | requests at 2048-token cap |
| 1 | 1.04% | 1.04% | 0 | 112 | 0 |
| 2 | 17.03% | 1.12% | 3 | 2048 | 2 |
| 3 | 1.14% | 1.14% | 0 | 145 | 0 |
| 4 | 22.02% | 1.17% | 2 | 2048 | 2 |
| 5 | 18.55% | 1.06% | 2 | 2048 | 1 |

Examples from current `pref/520` artifacts:

| Run | Sample id | completion_tokens | audio_duration_s | WER | Insertions | ASR normalized text prefix |
|---:|---|---:|---:|---:|---:|---|
| 2 | `common_voice_en_23699312-common_voice_en_23699314` | 2048 | 163.84 | 180.80 | 904 | `construction requires study and observation thank you very much thank you very much ...` |
| 4 | `common_voice_en_96260-common_voice_en_96263` | 2048 | 163.84 | 130.90 | 1309 | `he got up early with the idea of finding it i do not know i do not know ...` |
| 5 | `common_voice_en_581024-common_voice_en_581027` | 2048 | 163.84 | 170.09 | 1861 | `a a a a a a a a a a a a ...` |

Facts from current `pref/520` Voxtral EN artifacts:

| Run | WER | below-50 WER | n>50 | max completion tokens | requests at 4096-token cap |
| 1 | 1.20% | 1.20% | 0 | 148 | 0 |
| 2 | 1.28% | 1.28% | 0 | 188 | 0 |
| 3 | 1.42% | 1.22% | 1 | 227 | 0 |
| 4 | 1.25% | 1.25% | 0 | 154 | 0 |
| 5 | 1.16% | 1.16% | 0 | 147 | 0 |

Facts from prior no-graph `fix/444` Qwen3-TTS 0.6B EN artifacts:

| Artifact set | WER | below-50 WER | n>50 | max-token outlier examples |
|---|---:|---:|---:|---|
| `6ad898b` run 1 | 22.44% | 1.19% | 2 | 2 samples at `completion_tokens=2048`, `audio_duration_s=163.84` |
| `6ad898b` run 2 | 21.15% | 1.16% | 2 | 2 samples at `completion_tokens=2048`, `audio_duration_s=163.84` |
| `c25ad99` run 1 | 22.90% | 1.21% | 2 | 2 samples at `completion_tokens=2048`, `audio_duration_s=163.84` |

Examples from prior no-graph artifacts:

| Artifact set | Sample id | completion_tokens | audio_duration_s | WER | Insertions | ASR normalized text prefix |
|---|---|---:|---:|---:|---:|---|
| `c25ad99` run 1 | `common_voice_en_581024-common_voice_en_581027` | 2048 | 163.84 | 134.27 | 1467 | `a pile a pile a pile a pile ...` |
| `6ad898b` run 1 | `common_voice_en_17879420-common_voice_en_17879418` | 2048 | 163.84 | 157.67 | 946 | `every cloud has a silver lining thank you thank you ...` |

Conclusions supported by these artifacts:

- The Qwen3-TTS 0.6B EN high raw-WER rows are caused by a small number of successful requests with very long repeated audio, not by failed requests or skipped ASR samples.
- Those long repeated-audio requests reach the benchmark's `2048` completion-token cap.
- The same Qwen3-TTS 0.6B EN cap-reaching behavior is present in prior no-graph `fix/444` artifacts.
- In the current `pref/520` artifacts, Voxtral EN does not show the same cap-reaching pattern. Its maximum completion-token count is 227 across the five runs, and no Voxtral EN request reaches the `4096` cap.
- The current artifacts do not prove why Qwen3-TTS 0.6B EN enters this state, only that it does and that the behavior predates CUDA graph.

GaokaiZhang:
Checked the CUDA graph sampling-parameter alignment concern directly.

I ran Qwen3-TTS 0.6B on two H200 GPUs with the same explicit sampling request:

| Run | Branch / mode | GPU | Decode graph | sampling_backend | Sample events | Decode forwards |
| --- | --- | --- | --- | --- | ---: | ---: |
| Graph on | `pref/520` | 2 | `can_run_cuda_graph=True` | `pytorch` | 32 | 31 |
| Graph off | `fix/444` | 3 | `can_run_cuda_graph=False` | `pytorch` | 32 | 31 |

Request params were `temperature=0.8`, `top_p=0.8`, `top_k=30`, `repetition_penalty=1.1`, and `max_new_tokens=32`.

For every sampled step in both runs, `forward_batch.sampling_info` matched the request sampling params:

| Field | Graph on | Graph off |
| --- | --- | --- |
| `temperature` | `0.800000011920929` | `0.800000011920929` |
| `top_p` | `0.800000011920929` | `0.800000011920929` |
| `top_k` | `30` | `30` |

`repetition_penalty=1.1` remained on `req.sampling_params` in both paths and is applied by Omni before calling SGLang sampling. Both paths used `sampling_backend="pytorch"`.

Conclusion: this check did not find evidence that enabling CUDA graph changes, drops, or misaligns the Qwen3-TTS sampling parameters relative to the no-graph path.

sglang_omni/models/qwen3_tts/stages.py
```diff
@@ -105,6 +105,26 @@ def _load_qwen3_tts_generate_defaults(checkpoint_dir: str) -> dict[str, Any]:
...
+    set_torch_compile_config()
+    compile_mode = os.environ.get(
+        "SGLANG_TORCH_COMPILE_MODE",
+        "max-autotune-no-cudagraphs",
+    )
+    text_model._compiled_decode_layers = [
+        torch.compile(layer, mode=compile_mode) for layer in layers
+    ]
+
+
```
zhaochenyang20, line 126: the Qwen3-TTS startup path has moved compile responsibility from the SGLang graph runner into `_compile_qwen3_tts_backbone`: after `create_sglang_tts_engine_executor` calls this helper, it resets `server_args.enable_torch_compile` to `False` and then manually calls `init_device_graphs()`. However, the helper currently discovers decoder layers with `getattr(model, "model", None)` and `getattr(text_model, "layers", None)`, returning silently if they are missing; the model forward at `sglang_omni/models/qwen3_tts/sglang_model.py:165` then falls back to eager layers through `getattr(self, "_compiled_decode_layers", self.layers)`. If the wrapper shape, attribute name, or a test fake stops exposing the real layers, serving still starts and captures CUDA Graphs, but Qwen3 decode no longer goes through manually compiled layers. Since this PR's goal is `torch.compile + CUDA Graph`, that silent downgrade turns a target failure into an ordinary throughput fluctuation that is hard to catch in tests or logs.

Make this a fail-fast contract: access `model.model.layers` directly, build compiled wrappers for every layer, and have decode forward use `self._compiled_decode_layers` directly. If Qwen3 manual compile genuinely needs to be disabled in the future, initialize an explicit eager decode-layers attribute or a separate branch in the startup path instead of swallowing structure errors with defaults in both the helper and forward.

sglang_omni/models/qwen3_tts/sglang_model.py
```diff
@@ -111,36 +111,60 @@ def __init__(self, config: Any, prefix: str = "") -> None:
...
-            )
-            self._feedback_mask[:bs] = False
+            if is_decode:
+                row_ids = input_ids.clamp(
+                    min=0,
+                    max=self._decode_feedback_embedding.num_embeddings - 1,
+                )
+                hidden_states = self._decode_feedback_embedding(row_ids)
+            else:
+                hidden_states = self._build_input_hidden_states(input_ids)
```
zhaochenyang20, line 163: remove the `input_ids.clamp(...)` here. Qwen3 decode correctness in this PR depends on the runner rewriting `forward_batch.input_ids[:batch_size]` to `0..batch_size-1` outside the graph, after which model forward reads `_decode_feedback_embedding` with those row ids. If a future path skips `_write_feedback_buffers`, rewrites only part of the batch, or passes real codec token ids into decode forward, the current clamp silently maps invalid ids to the last staged embedding row. Instead of a clear failure, multiple requests can read the same or wrong feedback embedding row, showing up as bad audio, repeated segments, or hard-to-reproduce quality variance. This is an internal CUDA Graph staging invariant, not a user input boundary; out-of-range rows should surface immediately.

Call `self._decode_feedback_embedding(input_ids)` directly in decode forward and do not clamp. If debug protection is useful, put batch-size, row-id-range, and `input_ids` rewrite assertions in the runner preparation step outside the graph, avoiding host synchronization inside the graph-captured forward. A unit test can also assert that an invalid row id fails, preventing this invariant from drifting back into a soft fallback.

zhaochenyang20 (changes requested):

JiaxinD:
Took a careful pass through this one, walking the decode path by hand against the Higgs runner and the base `execute` flow. Overall it looks really solid, and the H200 numbers in the description back that up, so nothing here blocks merging. A handful of small things I noticed, mostly robustness and readability:

1. In `qwen3_tts/model_runner.py:129`, `_write_feedback_buffers` overwrites `forward_batch.input_ids` in place with row indices so the decode forward can use them as `_decode_feedback_embedding` lookups. It works today (SGLang's decode bookkeeping keys off KV slots and `next_token_ids`, not the decode `input_ids`, and the graph copies `input_ids` into its static buffer after `prepare_decode` runs), but the contract is very implicit. A one line comment along the lines of "during decode input_ids is repurposed as embedding row indices" would save the next reader some head scratching, and it's worth keeping in mind if overlap scheduling ever gets turned back on.

2. Both `qwen3_tts/model_runner.py:123` and `voxtral_tts/model_runner.py:48` call `torch.stack(rows)` with no empty batch guard, whereas the reference impl in `qwen3_omni/talker_model_runner.py:338` guards `if batch_size == 0: return`. The scheduler never hands us an empty decode batch today so it never fires, but keeping the guard for parity is cheap insurance.

3. After this change the Qwen3 `_feedback_mask` / `_feedback_buffer` path looks dead. Decode now goes through `_decode_feedback_embedding` and prefill always passes explicit `input_embeds`, so `_build_input_hidden_states` is unreachable, and the old `self._feedback_mask[:bs] = False` reset got dropped while `_write_feedback_buffers` still writes the mask every step. Not harmful, just a little confusing to read. Might be worth a cleanup pass or at least a note.

4. `_decode_feedback_embedding` is a real `nn.Embedding` submodule with random init that no checkpoint key ever loads. It's fine in practice since you overwrite the rows each step before reading them, but it stands out next to the zero initialized buffers and it leans on the loader not being strict about unloaded params. A `register_buffer` would sidestep both.

5. Minor: Voxtral keeps `enable_torch_compile=True` going into `init_device_graphs()` while Qwen3 clears it after the manual layer compile. I'm assuming that's intentional because the Voxtral backbone can ride SGLang's native compile path and the Qwen3 talker can't, but a one line comment would make the asymmetry obvious.

For what it's worth I also chased a few scarier theories and they all checked out fine, so noting them here so nobody re-litigates them: the Voxtral prefill change (set `forward_batch.input_embeds`, return None, let the model read the `input_embeds` arg) matches exactly what Higgs already does and SGLang threads it through, the mixed dtype `torch.stack` in the Voxtral decode buffer is safe because stack promotes the zero rows to the feedback dtype, and the empty `next_token_logits` is never actually sampled since `post_decode` always sets `next_token_ids`.

Nice work on the validation. The split generate then transcribe runs and the per row delta table made this a lot easier to trust.

zhaochenyang20:
great job so far. @Ratish1 Could you help to reivew. I will merge after your approval.

sglang_omni/models/qwen3_tts/request_builders.py
```diff
@@ -184,23 +188,80 @@ def build_qwen3_tts_state(payload: StagePayload) -> Qwen3TTSState:
...
+        if has_param(tts_params, params, "ref_text") or references_contain_text(
+            references
+        ):
+            raise ValueError("Qwen3-TTS CustomVoice does not accept ref_text")
+        if has_param(tts_params, params, "x_vector_only_mode"):
+            raise ValueError("Qwen3-TTS CustomVoice does not accept x_vector_only_mode")
+        voice = voice or QWEN3_TTS_DEFAULT_CUSTOM_VOICE
+        non_streaming_mode = True
+    elif task_type == QWEN3_TTS_TASK_VOICE_DESIGN:
+        if has_param(tts_params, params, "ref_text") or references_contain_text(
```
zhaochenyang20, line 246: use audio/text reference detection consistent with `has_voice_clone_reference()` in the `CustomVoice` and `VoiceDesign` branches, or split out `references_contain_audio()` plus `has_param(..., "ref_audio")`, and raise `ValueError` for `ref_audio`, `references[].audio_path`, `references[].ref_audio`, `references[].audio`, and `ref_text`. Add negative tests in `tests/unit_test/qwen3_tts/test_pipeline.py` for explicit `CustomVoice` and `VoiceDesign` requests that carry both the `ref_audio` parameter and `references[].audio_path`.

zhaochenyang20 (changes requested):

GaokaiZhang:
Full SeedTTS validation was rerun twice after the rebase on branch head `465c72b`.

Setup: `benchmarks/eval/benchmark_tts_seedtts.py`, non-streaming generation, 4 H200 shards, `c=16` per shard, with generate-only/transcribe-only split. Qwen Base used SeedTTS reference audio/text. CustomVoice used text-only `task_type=CustomVoice`, `voice=Vivian`. VoiceDesign used text-only `task_type=VoiceDesign` with instruction `A warm, natural young adult voice.` Voxtral used no-ref `voice=cheerful_female`.

Run 1:

| Qwen/Qwen3-TTS-12Hz-0.6B-Base | EN | WER 1.65% raw; 1.05% excl. 2 >50% samples | 1088/1088 | 0 / 0 | 7.244 req/s | 1.1649 |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | ZH | CER 1.15% raw; 1.12% excl. 1 >50% sample | 2020/2020 | 0 / 0 | 10.664 req/s | 0.9419 |
| Qwen/Qwen3-TTS-12Hz-0.6B-CustomVoice (Vivian, no-ref) | EN | WER 1.92% raw; 1.48% excl. 2 >50% samples | 1088/1088 | 0 / 0 | 12.296 req/s | 0.9652 |
| Qwen/Qwen3-TTS-12Hz-0.6B-CustomVoice (Vivian, no-ref) | ZH | CER 1.23% | 2020/2020 | 0 / 0 | 11.992 req/s | 0.8596 |
| Qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign (instruction, no-ref) | EN | WER 1.24% | 1088/1088 | 0 / 0 | 15.320 req/s | 1.0541 |
| Qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign (instruction, no-ref) | ZH | CER 1.35% | 2020/2020 | 0 / 0 | 13.804 req/s | 0.9650 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | EN | WER 1.47% raw; 1.27% excl. 2 >50% samples | 1088/1088 | 0 / 0 | 34.324 req/s | 0.3200 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | ZH | CER 154.13% raw; 39.13% excl. 2013 >50% samples | 2020/2020 | 0 / 0 | 12.272 req/s | 0.2897 |

Run 2:

| Qwen/Qwen3-TTS-12Hz-0.6B-Base | EN | WER 1.08% | 1088/1088 | 0 / 0 | 11.860 req/s | 1.2182 |
| Qwen/Qwen3-TTS-12Hz-0.6B-Base | ZH | CER 1.14% | 2020/2020 | 0 / 0 | 8.596 req/s | 0.9255 |
| Qwen/Qwen3-TTS-12Hz-0.6B-CustomVoice (Vivian, no-ref) | EN | WER 1.54% | 1088/1088 | 0 / 0 | 12.848 req/s | 0.9511 |
| Qwen/Qwen3-TTS-12Hz-0.6B-CustomVoice (Vivian, no-ref) | ZH | CER 1.24% | 2020/2020 | 0 / 0 | 12.012 req/s | 0.8693 |
| Qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign (instruction, no-ref) | EN | WER 1.21% | 1088/1088 | 0 / 0 | 15.512 req/s | 1.0585 |
| Qwen/Qwen3-TTS-12Hz-1.7B-VoiceDesign (instruction, no-ref) | ZH | CER 1.43% raw; 1.39% excl. 1 >50% sample | 2020/2020 | 0 / 0 | 14.284 req/s | 0.9389 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | EN | WER 1.21% raw; 1.18% excl. 1 >50% sample | 1088/1088 | 0 / 0 | 29.596 req/s | 0.3628 |
| mistralai/Voxtral-4B-TTS-2603 (no-ref, cheerful_female) | ZH | CER 152.63% raw; 41.60% excl. 2008 >50% samples | 2020/2020 | 0 / 0 | 17.160 req/s | 0.2642 |

Notes:

- All rows completed with zero generation failures and zero ASR skips.
- The new Qwen3-TTS task modes were exercised end to end: Base reference cloning, CustomVoice text-only speaker table lookup, and VoiceDesign text-only instruction prompting.
- Voxtral ZH remains non-comparable as an accuracy signal because the model does not document Chinese support; it is included here only as a no-failure serving/speed check.

zhaochenyang20 (approved):

Ratish1 (approved):

PR #530, opened 2026-05-23
[CI] Report all CI metric threshold failures

zhaochenyang20:
As titled, even though we want our CI to fail fast, we still need it to report all the numbers that we have and to say what threshold is actually not met.


PasserBy4 (approved):

tests/utils.py
```diff
@@ -389,70 +533,100 @@ def assert_wer_partitioned(
     tail of wildly-wrong outputs, without the length-sensitivity of a
     single corpus-wide WER.
     """
-    summary = results["summary"]
-    per_sample = results["per_sample"]
+    checks = _metric_collector(collector, "partitioned WER")
+    summary = results.get("summary", {})
+    per_sample = results.get("per_sample", [])
```
PasserBy4: It seems that `results.get("per_sample", [])` lets a result without `per_sample` pass the per-sample validation entirely, and `sample.get("wer")` skips samples where `wer` is missing. Can we collect these as schema failures instead of defaulting/skipping them?

Issue #463, opened 2026-05-17
[Perf] Replace time.sleep(0.001) with time.sleep(0) in _event_loop_normal idle yield — preserves GIL-yield semantics, saves ~10% wall


Hayden727:
# [Perf] Replace `time.sleep(0.001)` with `time.sleep(0)` in `_event_loop_normal` idle yield — preserves GIL-yield semantics, saves ~10% wall

## TL;DR

`sglang_omni/scheduling/omni_scheduler.py:_event_loop_normal`'s idle branch currently uses `time.sleep(0.001)` to yield the GIL (Chenyang's fix to prevent audio_encoder starvation). **A 1-character change to `time.sleep(0)`** delivers:
- text-only c=8: wall **-10%**, TTFT p50 **-76%**
- audio c=4 with 30s input: wall **-10%**, TTFT **-12%**, throughput **+8%**
- Fully preserves the GIL-yield semantics described in Chenyang's Note (measured: the audio path doesn't regress — it actually improves slightly)

## Context

Follow-up to [#379](https://github.com/sgl-project/sglang-omni/issues/379) (Qwen3-Omni vs Qwen3-VL MMMU TTFT 28× gap — see [the #379 follow-up comment](https://github.com/sgl-project/sglang-omni/issues/379#issuecomment-4464961641)). During framework-layer profile analysis on H200, the idle branch of `_event_loop_normal` was identified as a **single point consuming ~38% of wall time on c=8 burst workloads**.

## Current code (main `8bc31f4`)

```python
# sglang_omni/scheduling/omni_scheduler.py:_event_loop_normal
def _event_loop_normal(self) -> None:
    # Note (Chenyang): yield the GIL when idle so co-located non-AR stages
    # (encoders, preprocessor) running in sibling threads aren't starved
    # of Python execution. Without this, in single-process mode the busy
    # AR scheduler loop pins the GIL and the audio_encoder forward pass
    # (which is mostly Python-side dispatch into many small CUDA kernels)
    # slows ~600x, dropping audio QPS from >10 to <0.5.
    while self._running:
        ...
        if batch:
            result = self.run_batch(batch)
            self.process_batch_result(batch, result)
        else:
            self.self_check_during_idle()
            time.sleep(0.001)    # ← current implementation
```

## Profile evidence

Self-time analysis of the thinker tid (1889) on a 25 s c=8 text-only burst trace, H200:

```
[...]

PR #543, opened 2026-05-24
[Bugfix] OmniScheduler: yield GIL in overlap event loop

draftbk:
## Summary

`_event_loop_overlap` is missing the `time.sleep(0.001)` GIL yield that `_event_loop_normal` has — without it, the spinning loop starves co-located stages of the GIL when `enable_overlap=True`. See #463 for the documented rationale on the sibling path.

## Test plan

Forced `enable_overlap=True` in a model adapter bootstrap, ran on H200:

- Voxtral: 1 curl to `/v1/audio/speech` with prompt "Test with overlap enabled." (generates ~4s of audio): **6.8s → 3.7s** wall (before/after fix)
- Qwen3-Omni 30B: 8 concurrent chat completions, max_tok=128, 3 reps: **243 → 553 t/s** (before/after fix)

For reference, `normal` mode (current default) hits 0.6s / 670 t/s on the same workloads — enabling overlap doesn't actually improve over normal even with this fix. Parity fix for the latent path.


draftbk:
cc @zhaochenyang20 @Hayden727 — mirroring the normal loop's sleep on the overlap side.

zhaochenyang20 (approved):

PR #553, opened 2026-05-24
[Feat] Add tensor parallelism support to Ming-Omni image encoder

### LINKED ISSUE #535: [Feature] Ming-Omni Image Encoder Support TP
### Checklist

### Motivation

Currently Ming-Omni image encoder doesn't use TP, which makes sglang-omni thinker's image understand doesn't outperform than other omni engine in this scenario. We need to improve it.

### Related resources

_No response_

ZhitongGuo:
Closes #535

## Motivation

The Ming-Omni image encoder (ViT + projector, ~7B params) currently forces `TP=1` via a hardcoded `_init_sglang_tp()` that initializes `parallel_state` with `world_size=1, rank=0`. This prevents sharding the vision encoder across multiple GPUs, which becomes a bottleneck when the image encoder must process high-resolution or multi-image inputs at scale.

The vision encoder internally already uses sglang's TP-aware layers (`ColumnParallelLinear`, `RowParallelLinear`, `QKVParallelLinear`, `VocabParallelEmbedding`) — their `weight_loader` attributes handle sharded weight loading automatically. The only missing piece is plumbing the real TP context (rank/size/port) from the pipeline config into the image encoder process.

## Modifications

**Image encoder TP plumbing:**
- `components/image_encoder.py` — Parameterize `_init_sglang_tp()` to accept `tp_rank`, `tp_size`, `nccl_port` instead of hardcoding TP=1. Initialize `parallel_state` and `dp_attention` with the real TP dimensions.
- `components/projectors.py` — Convert `VisionProjector` from `nn.Linear` to `ColumnParallelLinear`/`RowParallelLinear` so weights are sharded across TP ranks. At TP=1 these are functionally identical to `nn.Linear`.
- `config.py` — `_image_encoder_stage()` now accepts `tp_size` and `gpu: int | list[int]` for multi-GPU placement.
- `stages.py` (legacy) + `pipeline/stages.py` (new) — Plumb `tp_rank`, `tp_size`, `nccl_port` kwargs through both stage factory implementations to `MingImageEncoder`.
- `examples/run_ming_omni_server.py` — Add `--image-encoder-tp` and multi-GPU `--gpu-image-encoder` CLI args.

**Tests:**
- `test_tp.py` — `test_ming_image_encoder_tp2_builds_rank_specific_stage_specs`: Verifies `_build_stage_groups` produces leader+follower specs with correct `tp_rank`, `gpu_id`, `nccl_port`.
- `test_tp.py` — `test_ming_image_encoder_tp2_config_accepts_gpu_list`: Verifies config with `gpu=[2,3], tp_size=2`.
- `test_pipeline.py` — `test_ming_image_encoder_factory_accepts_tp_injection_args`: Verifies factory signature.
- `test_pipeline.py` — `test_ming_text_launcher_configures_image_encoder_tp`: Verifies `--image-encoder-tp 2 --gpu-image-encoder 2 3` launcher path.

## Accuracy Tests

**Ming omni TP = 2**

Server:
```bash
CUDA_VISIBLE_DEVICES=2,3,4,5,6,7 \
  python examples/run_ming_omni_server.py \
  --tp-size 2 --image-encoder-tp 2 \
  --gpu-image-encoder 4 5 --gpu-audio-encoder 2 \
  --cpu-offload-gb 20 --port 8200 --host 127.0.0.1
```

Client & Result:
```
$ curl http://127.0.0.1:8200/v1/chat/completions \
    -d '{
      "model": "ming-omni",
      "messages": [{"role": "user", "content": "what is the 2 + 2?"}],
      "max_tokens": 32,
      "temperature": 0.0,
      "stream": false
    }'
{"id":"chatcmpl-6b0e2445-5b57-4efc-a2ab-0244ec0153a8","object":"chat.completion","created":1779609784,"model":"ming-omni","choices":[{"index":0,"message":{"role":"assistant","content":"2 + 2 equals 4."},"finish_reason":"stop"}],"usage":{"prompt_tokens":35,"completion_tokens":9,"total_tokens":44}}

$ curl http://127.0.0.1:8200/v1/chat/completions \
    -d '{
      "model": "ming-omni",
      "messages": [{"role": "user", "content": "what is the capital of Japan?"}],
      "max_tokens": 32,
      "temperature": 0.0,
      "stream": false
    }'
{"id":"chatcmpl-7873d1c1-5318-4845-96e9-add95ad2c11f","object":"chat.completion","created":1779609792,"model":"ming-omni","choices":[{"index":0,"message":{"role":"assistant","content":"The capital of Japan is Tokyo."},"finish_reason":"stop"}],"usage":{"prompt_tokens":33,"completion_tokens":8,"total_tokens":41}}
```

**Ming Omni Concurrency Test (Thinker TP=2 + Image Encoder TP=2)**
```
Warming up with 4 request(s)...
  conc  rep    ok   err      req/s        tps      p50      p95      p99      max
------------------------------------------------------------------------------------
     1    1    20     0       0.18       2.5*     3.88    13.01    13.02    13.02
     1    2    20     0       0.18       2.5*     3.85    13.06    13.08    13.09
     1    3    20     0       0.17       2.5*     3.88    12.99    13.06    13.07
     2    1    20     0       0.28       4.0*     4.70    14.82    15.04    15.10
     2    2    20     0       0.28       4.0*     4.70    15.18    15.52    15.61
     2    3    20     0       0.29       4.0*     4.66    14.64    14.72    14.74
     4    1    20     0       0.47       6.6*     5.33    16.78    17.12    17.21
     4    2    20     0       0.46       6.5*     5.55    17.16    17.21    17.22
     4    3    20     0       0.47       6.6*     5.32    16.84    17.13    17.20
     8    1    32     0       0.82      11.2*     5.33    18.03    18.05    18.05
     8    2    32     0       0.82      11.2*     5.52    17.98    18.00    18.00
     8    3    32     0       0.82      11.2*     5.31    18.18    18.42    18.43
    16    1    64     0       1.46      19.9*     6.44    21.88    22.89    23.13
    16    2    64     0       1.48      20.2*     6.59    21.08    22.25    22.25
    16    3    64     0       1.45      19.8*     6.55    21.83    22.84    23.10
    32    1   128     0       1.55      21.1*    16.17    33.30    34.86    35.51
    32    2   128     0       1.54      21.0*    16.22    33.47    34.98    35.66
    32    3   128     0       1.55      21.1*    16.12    33.32    34.78    35.54
```

> **Note**: Tested on H100 95GB GPUs with `--cpu-offload-gb 20` (needed for H100; H200 141GB GPUs would not need offload). Image encoder TP=2 produces identical text output to TP=1 — only the vision encoder computation is sharded.

## Checklist

- [x] Format code with pre-commit
- [x] Follow SGLang code style guidance
- [x] Unit tests added (53 Ming-Omni tests pass, 592 total pipeline tests pass)
- [x] `git diff --check` clean
- [x] Accuracy tests with real model (thinker TP=2 + image encoder TP=2)
- [x] Full concurrency benchmark (conc 1–32, 3 reps, 0 errors)


ZhitongGuo:
Hi @edwingao28! Looks like the CI is not triggered correctly, would you mind reviewing this PR? I can't seem to add you as a reviewer due to not having push access to the repo. Thanks!
cc @shuaills @FrankLeeeee @zhaochenyang20 @yuan-luo

yuan-luo:
Thanks for following up this issue. It is really helpful for ming-omni.

zhaochenyang20:
only after PR is approved, then run the CI.

examples/run_ming_omni_server.py
```diff
@@ -203,8 +213,18 @@ def _launch_text_server(args: argparse.Namespace) -> None:
...
+    image_encoder_tp = getattr(args, "image_encoder_tp", 1) or 1
+    if image_encoder_tp > 1:
+        img_stage = next(
+            stage for stage in config.stages if stage.name == "image_encoder"
+        )
+        img_stage.tp_size = image_encoder_tp
+        img_stage.parallelism = img_stage.parallelism.model_copy(
+            update={"tp": image_encoder_tp}
+        )
+        img_stage.gpu = args.gpu_image_encoder or list(range(image_encoder_tp))
```
edwingao28: 1. This silently falls back to GPUs [0, 1, ..., N-1] when --gpu-image-encoder is omitted, which can implicitly colocate the image encoder with the thinker. 
Please add if image_encoder_tp > 1, require --gpu-image-encoder to be explicitly provided and raise a clear error otherwise.

2. Please also validate that the number of GPU ids passed to `--gpu-image-encoder` matches `image-encoder-tp`, and that there are no duplicates among them

examples/run_ming_omni_server.py
```diff
@@ -203,8 +213,18 @@ def _launch_text_server(args: argparse.Namespace) -> None:
         server_arg_updates["disable_custom_all_reduce"] = True
     if args.gpu_audio_encoder is not None:
         _set_stage_gpu(config, "audio_encoder", args.gpu_audio_encoder)
-    if args.gpu_image_encoder is not None:
-        _set_stage_gpu(config, "image_encoder", args.gpu_image_encoder)
+    image_encoder_tp = getattr(args, "image_encoder_tp", 1) or 1
+    if image_encoder_tp > 1:
```
edwingao28, line 221:
```suggestion
    if image_encoder_tp > 1 and args.thinker_only:
          raise ValueError("--thinker-only cannot be used with --image-encoder-tp > 1")
```

Please add an early validation

examples/run_ming_omni_server.py
```diff
@@ -203,8 +213,18 @@ def _launch_text_server(args: argparse.Namespace) -> None:
         server_arg_updates["disable_custom_all_reduce"] = True
     if args.gpu_audio_encoder is not None:
         _set_stage_gpu(config, "audio_encoder", args.gpu_audio_encoder)
-    if args.gpu_image_encoder is not None:
-        _set_stage_gpu(config, "image_encoder", args.gpu_image_encoder)
+    image_encoder_tp = getattr(args, "image_encoder_tp", 1) or 1
```
edwingao28: Using `or 1` silently converts --image-encoder-tp 0 (an invalid value) into 1.
  StageConfig.tp_size already requires tp_size >= 1; this CLI layer should enforce the same constraint explicitly:
  
      if args.image_encoder_tp < 1:
          raise ValueError("--image-encoder-tp must be >= 1")

sglang_omni/models/ming_omni/components/image_encoder.py
```diff
@@ -136,7 +145,9 @@ def _init_sglang_tp(cls):
             return  # Already initialized
```
edwingao28: If `_init_sglang_tp` is called again after TP has already been initialized, the early return silently ignores the new tp_size. This can leave a stale TP state active without any warning. 
Please add a check that the requested tp_size matches what was previously initialized, and raise an error if they differ.

edwingao28 (changes requested):

edwingao28:
Thanks for the PR @ZhitongGuo. Since this change directly touches the Ming-Omni image encoder TP path and also changes `VisionProjector` to use parallel linear layers, could you also provide an image-input correctness result?

A small MMMU sample=50 run is enough. The main thing I want to verify is that `image_encoder_tp=2` preserves vision correctness and does not introduce runtime errors on real image inputs.

Could you please run:

1. `thinker TP=2, image_encoder_tp=1`
2. `thinker TP=2, image_encoder_tp=2`

on the same `zhaochenyang20/mmmu-ci-50` subset, with accuracy / successful count / error count / latency summary?

For the `image_encoder_tp=2` run, please test the colocated placement explicitly: thinker TP and image encoder TP should use the same two visible GPUs.
The baseline can keep thinker TP constant and place the TP=1 image encoder on one of the same visible GPUs, then run the MMMU subset against each server.

Please use colocated case for this test because this PR’s placement behavior can put the thinker and image encoder TP groups on the same GPUs, and that is the highest-risk configuration for memory/runtime interaction. A TP=2 result where the image encoder is placed on a separate GPU pair would not fully cover that case.

The repo already has the runner you'd need: 
 benchmarks/eval/benchmark_omni_mmmu.py with --max-samples 50 and the 
 pinned subset --repo-id zhaochenyang20/mmmu-ci-50

ZhitongGuo:
## MMMU Accuracy Results

Ran `benchmarks/eval/benchmark_omni_mmmu.py` on `zhaochenyang20/mmmu-ci-50` (50 samples), colocated on 2×H100 95GB.

| Config | Completed | Correct (completed) | Accuracy | Offload |
|--------|-----------|--------------------|---------:|--------:|
| thinker TP=2, img_tp=1, GPUs 0,1 | 30/50 | 22/30 | **73.3%** | 80 GB |
| thinker TP=2, img_tp=2, GPUs 0,1 | 34/50 | 26/34 | **76.5%** | 20 GB |

**Per-sample agreement (29 samples both completed): 96.6%** — 21 both correct, 7 neither correct, 1 only TP=1 correct.

### Notes

- Both tests ran with `--max-concurrency 1 --max-tokens 2048 --temperature 0.0`
- Timeouts (300s) are from CPU offloading on 2×H100 colocated setup — not image encoder TP related
- Test 1 needed more CPU offload (80GB) because IE shares GPU 0 with thinker rank 0; test 2 distributes IE across both GPUs so only 20GB offload needed
- Discovered and fixed an NCCL deadlock in TP follower work dispatch for non-SGLang schedulers (commit 4eb3ab2) — the TP leader was not replicating request payloads to followers, so `RowParallelLinear.forward()` deadlocked in `all_reduce`

### Commits pushed

- `a5ae76c` — Address reviewer feedback: validate CLI args and TP re-entry (addresses all 4 inline comments)
- `4eb3ab2` — Fix NCCL deadlock in TP follower dispatch for non-SGLang schedulers

edwingao28:
Looks great. Please also include latency mean, tok/s per-req mean, gen tokens mean / total, prompt tokens mean / total, throughput (req/s) to the table as well, since they already included in the MMMU benchmark.

ZhitongGuo:
Updated table with full benchmark metrics as requested:

## MMMU Accuracy & Performance Results

Dataset: `zhaochenyang20/mmmu-ci-50` (50 samples), colocated on 2×H100 95GB, `--max-concurrency 1 --max-tokens 2048 --temperature 0.0`.

| Metric | thinker TP=2, img_tp=1 | thinker TP=2, img_tp=2 |
|--------|----------------------:|----------------------:|
| Completed / Total | 30 / 50 | 34 / 50 |
| Correct (of completed) | 22 / 30 (**73.3%**) | 26 / 34 (**76.5%**) |
| Latency mean (s) | 138.8 | 145.7 |
| Latency median (s) | 121.5 | 135.3 |
| Latency p95 (s) | 276.7 | 273.4 |
| Tok/s per req (output) | 2.3 | 2.3 |
| Throughput (req/s) | 0.003 | 0.003 |
| Gen tokens mean / total | 313 / 9,376 | 330 / 11,232 |
| Prompt tokens mean / total | 744 / 22,319 | 722 / 24,541 |
| CPU offload (GB) | 80 | 20 |

**Per-sample agreement (29 samples both completed): 96.6%** — 21 both correct, 7 neither correct, 1 only TP=1 correct.

> Note: Latency and throughput are heavily dominated by CPU offload (both configs are memory-constrained on 2×H100 colocated). Test 1 needed 80 GB offload because the image encoder shares GPU 0 with thinker rank 0; test 2 distributes the encoder across both GPUs so only 20 GB offload was needed. A fairer perf comparison would use dedicated GPUs for the image encoder (non-colocated).

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -549,6 +552,8 @@ async def _execute(self, payload: Any) -> None:
             stage=self.name,
             event_name="stage_dispatch",
         )
+        if self.role == "leader" and self._tp_fanout is not None:
+            self._tp_fanout.fanout_work(payload)
```
edwingao28, line 651: This fanout is needed for SimpleScheduler stages like the image encoder, but not for OmniScheduler stages like thinker TP — followers there receive requests via rank-0 broadcast and nothing drains their local inbox, so the extra `TPWorkMessage` just accumulates in `follower.scheduler.inbox`.

Could we add a `requires_tp_work_fanout` flag on the scheduler, defaulting to `False` and set to `True` on `SimpleScheduler`? A small unit test that an OmniScheduler-backed TP leader doesn't call `fanout_work()` would lock it in.

edwingao28 (changes requested):
`fanout_work()` in `_execute()` fires for every TP leader, including OmniScheduler stages like thinker TP. But OmniScheduler followers get their requests via rank-0 broadcast and never drain their local inbox, so these `TPWorkMessage`s just pile up in `follower.scheduler.inbox` could cause unbounded growth on a long-running server. Could we gate this to non-SGLang schedulers?

sglang_omni/pipeline/stage/runtime.py
ZhitongGuo: Done — added `requires_tp_work_fanout` flag (`True` on `SimpleScheduler`, `False` on `OmniScheduler`) and guarded `fanout_work()` behind it in `Stage._execute()`. Unit test `test_tp_leader_skips_fanout_work_for_omni_scheduler` verifies both paths. See 915f9bc.

edwingao28: One test issue: the new fanout unit test constructs `Stage` without a fake relay, so it falls back to the default NIXL relay and fails in envs without NIXL. Since the test only checks `_execute()` fanout behavior, could you pass a fake relay when creating the test `Stage`?

edwingao28 (approved):

edwingao28:
lgtm and added runci labtel. cc @zhaochenyang20 @yuan-luo

ZhitongGuo:
Hi @edwingao28! Looks like the previous CI failed because of some commits in the main branch are not merged. Would you mind triggering it again?

edwingao28:
All CI passed, @yuan-luo @zhaochenyang20 could you PTAL?

yuan-luo:
LGTM. Thanks @ZhitongGuo @edwingao28 
Let's ship it.

PR #563, opened 2026-05-25
feat: LRU cache for audio encoder stage

### LINKED ISSUE #562: [Feat] Higgs TTS: LRU cache for audio encoder stage
## Motivation

The Higgs TTS pipeline runs the audio codec encoder on every request that supplies a raw waveform (`reference_waveform`), even when the same reference audio has been encoded moments ago. In typical TTS serving the set of speaker reference audios is small and reused heavily — same speaker prompt, many target texts. Re-running `codec.encode_reference` per request burns GPU time and serializes requests behind the encoder stage.

An LRU cache keyed by the reference-audio identity would let repeat requests skip the codec forward entirely and reuse `reference_codes_delayed` + the assembled prompt. The hit path becomes a dict lookup, no GPU work.

## Current encoder mechanism

`create_audio_encoder_executor` in `sglang_omni/models/higgs_tts/stages.py` is the GPU stage that converts a raw waveform into delayed codec codes and a prompt:

```python
def _encode(payload):
    state = HiggsTtsState.from_dict(payload.data)
    waveform = state.reference_waveform
    if waveform is None:
        return payload  # fast path: client already supplied pre-encoded codes

    ref_codes_TN = codec.encode_reference(waveform, sample_rate=24000).to(torch.long)
    delayed = apply_delay_pattern(ref_codes_TN)
    state.reference_codes_delayed = delayed.tolist()
    state.prompt_token_ids = adapter.build_prompt(...)
    return payload
```

Every request with a raw waveform pays the full codec encode + delay pattern + prompt-build cost, regardless of whether an identical waveform was processed seconds ago. There is already a "client-supplied pre-encoded" fast path (the `waveform is None` early return), so the abstraction for skipping the GPU work exists — we just don't reuse our own prior work.

## Proposed design

Wrap `_encode` with a small LRU cache:

```python
@lru_cache(maxsize=128)
def _encode_cached(audio_hash: bytes, target_text: str, reference_text: str | None):
    # returns (reference_codes_delayed: list, prompt_token_ids: list)
    ...
```

Cache key:
- `audio_hash`: BLAKE2 / SHA256 of the raw waveform bytes (or sample-rate + raw bytes if rate can vary)
- `target_text` / `reference_text`: included because `prompt_token_ids` depends on them; if we only cache `reference_codes_delayed` and rebuild the prompt every call this can be dropped

Cache value:
- `reference_codes_delayed` (the heavy GPU output)
- optionally `prompt_token_ids` (cheap to rebuild from cached `delayed.shape[0]`)

Two reasonable scopes:

1. **Encoder-only cache**: cache `reference_codes_delayed` only, rebuild prompt each call. Smaller key, smaller value, prompt logic stays linear.
2. **Full stage cache**: cache both `reference_codes_delayed` and `prompt_token_ids`. Larger key, but cuts the prompt-build too.

Default to scope 1 for simplicity; scope 2 if profiling shows prompt-build is non-trivial.

### Considerations

- **Eviction**: `functools.lru_cache(maxsize=128)` is fine for in-process; for multi-process pipelines, an explicit `collections.OrderedDict`-based LRU under a lock is clearer about memory.
- **Memory budget**: each cached entry is `[T, N=8] long`. Typical ref audio is ~3–10s @ 24kHz / hop → a few hundred frames → tens of KB per entry. Cache of 128 is well under 10 MB.
- **Hash cost**: BLAKE2 of a 10s waveform (~480 KB bf16) is sub-millisecond on CPU, much cheaper than the codec forward.
- **No correctness risk**: identical waveform + identical text → identical output (codec is deterministic). Hash collision risk negligible at expected scale.
- **Backwards compatibility**: cache miss path is the current encode path verbatim; the client-supplied pre-encoded fast path is unaffected.

### Expected impact

This is a small, targeted optimization, not a headline speedup: 

Workloads with mostly-cold ref audios (every request a new speaker) see ~0 benefit, only the cost of a hash per request (sub-ms). Recommend treating this as a quality-of-life win for hot-prompt serving rather than a major speedup.

Detailed estimation is pending on profiling. 

### Out of scope

- Cross-process / cross-worker shared cache (would need Redis / shared memory).
- Cache for the vocoder (different access pattern — output codes are per-request and unique).
- Persistent cache across server restarts.

## References

- Tracking issue: #478

nagisa-kunhah:
## Motivation

Reduce repeated Higgs TTS reference audio encoding work when multiple requests reuse the same raw reference audio. The audio encoder can reuse delayed reference codes for identical waveforms while still rebuilding prompts per request text/reference text.

## Modifications

- Added a small reusable in-process LRU cache utility.
- Added an LRU cache in the Higgs TTS audio encoder for raw reference audio encoded outputs.
- Kept this separate from `StageOutputCache` because that class is named around stage outputs, while this cache stores a model-specific intermediate value. Reusing it would also require changing `get()` copy semantics, which would broaden the impact to existing users.
- Added unit tests for cache hit/miss, eviction, disabled cache, fast path behavior, config passthrough, and the generic LRU cache.

## Related Issues

Fixes #562

## Accuracy Test

N/A

## Benchmark & Profiling

N/A

## Checklist

- [x] Add unit tests.

## CI


nagisa-kunhah:
@Ccyest Hello maintainer, I need the private Higgs checkpoint to validate our end-to-end pipeline. Is there any public alternative that can exercise the full path?

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -221,8 +255,35 @@ def _encode(payload: StagePayload) -> StagePayload:
...
+            state.target_text = None
+            state.reference_text = None
+            payload.data = state.to_dict()
+            return payload
+
+        result = _encode(payload)
+        encoded_state = HiggsTtsState.from_dict(result.data)
+        if encoded_state.reference_codes_delayed is not None:
+            reference_audio_cache.put(cache_key, encoded_state.reference_codes_delayed)
+        return result
```
zhaochenyang20: Precompute a `cache_enabled = reference_audio_cache.max_size != 0` flag in the closure, or branch directly on the normalized `reference_audio_cache_size`; when disabled, `_cached_encode` should immediately call `_encode(payload)` for waveform requests without invoking `_reference_waveform_cache_key`. Also strengthen the existing disabled-cache unit test by monkeypatching `_reference_waveform_cache_key` and asserting it is not called, so the test covers the real performance semantics of the configuration.

zhaochenyang20: Also, Add a byte budget to this cache, or store cached values as compact CPU tensors/arrays and evict based on `numel * element_size`; on hit, convert back to a list for the current `HiggsTtsState` API. A minimal fix is to implement `max_bytes` at the Higgs audio-cache layer with an explicit, documentable default budget and add a unit test for eviction with long reference audio. Whether or not this reuses `StageOutputCache`, the entry-count and byte-count boundaries should both be explicit here.

zhaochenyang20 (changes requested):
Add a lightweight benchmark case; it can use a fake codec without the private checkpoint to report call counts, hash time, and payload-building time. If a public checkpoint path is available, add a real `codec.encode_reference` hit/miss latency comparison as well. Put the command and numbers back into the PR description so reviewers have evidence for this performance change.

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -221,8 +255,35 @@ def _encode(payload: StagePayload) -> StagePayload:
         payload.data = state.to_dict()
         return payload
 
+    def _cached_encode(payload: StagePayload) -> StagePayload:
+        state = HiggsTtsState.from_dict(payload.data)
+        waveform = state.reference_waveform
+        if waveform is None:
+            return _encode(payload)
```
xinlij: `_cached_encode` already deserializes `payload.data` at the top of the function. When waveform is None, calling `_encode(payload)` triggers a second `HiggsTtsState.from_dict` inside `_encode`, only to immediately return payload unchanged.

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -221,8 +255,35 @@ def _encode(payload: StagePayload) -> StagePayload:
...
+            state.reference_waveform = None
+            state.target_text = None
+            state.reference_text = None
+            payload.data = state.to_dict()
+            return payload
+
+        result = _encode(payload)
+        encoded_state = HiggsTtsState.from_dict(result.data)
+        if encoded_state.reference_codes_delayed is not None:
+            reference_audio_cache.put(cache_key, encoded_state.reference_codes_delayed)
```
xinlij: The value passed to put is the same list object as result.data["reference_codes_delayed"] — to_dict() stores it by reference. If downstream code mutates the returned payload's reference_codes_delayed, the cached entry is silently corrupted. copy_on_get only protects values returned from get; the stored value itself needs a defensive copy at put time: reference_audio_cache.put(cache_key, _copy_reference_codes_delayed(encoded_state.reference_codes_delayed)).

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -221,8 +255,35 @@ def _encode(payload: StagePayload) -> StagePayload:
...
+                reference_text=state.reference_text,
+            )
+            state.reference_waveform = None
+            state.target_text = None
+            state.reference_text = None
+            payload.data = state.to_dict()
+            return payload
+
+        result = _encode(payload)
+        encoded_state = HiggsTtsState.from_dict(result.data)
```
xinlij: On a cache miss, _encode has already run HiggsTtsState.from_dict(payload.data) internally and written the result back to payload.data. This line then deserializes result.data a second time just to read reference_codes_delayed for the cache put. Consider having _encode return the encoded codes directly, or inlining the encoding logic here, to avoid the extra round-trip through serialization.

xinlij (changes requested):

sglang_omni/models/higgs_tts/stages.py
nagisa-kunhah: > Precompute a `cache_enabled = reference_audio_cache.max_size != 0` flag in the closure, or branch directly on the normalized `reference_audio_cache_size`; when disabled, `_cached_encode` should immediately call `_encode(payload)` for waveform requests without invoking `_reference_waveform_cache_key`. Also strengthen the existing disabled-cache unit test by monkeypatching `_reference_waveform_cache_key` and asserting it is not called, so the test covers the real performance semantics of the configuration.

ok, I Addressed this by computing a `cache_enabled` flag from the normalized cache limits before constructing the scheduler.

The current logic normalizes both `reference_audio_cache_size` and `reference_audio_cache_max_bytes` via `normalize_lru_limit(...)`, then enables the reference-audio cache only when both normalized values are non-zero:

```python
def normalize_lru_limit(limit: int | None) -> int:
    return max(int(limit), 0) if limit is not None else 0
...
cache_enabled = (
    normalized_reference_audio_cache_size != 0
    and normalized_reference_audio_cache_max_bytes != 0
)
```

So `0`, negative values, and `None` all disable the cache after normalization. When the cache is disabled, `create_audio_encoder_executor` passes the original `_encode` function directly to `SimpleScheduler`:

```python
return SimpleScheduler(
    _cached_encode if cache_enabled else _encode,
    ...
)
```

nagisa-kunhah: > Also, Add a byte budget to this cache, or store cached values as compact CPU tensors/arrays and evict based on `numel * element_size`; on hit, convert back to a list for the current `HiggsTtsState` API. A minimal fix is to implement `max_bytes` at the Higgs audio-cache layer with an explicit, documentable default budget and add a unit test for eviction with long reference audio. Whether or not this reuses `StageOutputCache`, the entry-count and byte-count boundaries should both be explicit here.

Addressed this by bounding the reference-audio cache by both entry count and bytes, and by storing cached codes as compact CPU tensors.

The Higgs audio encoder now has an explicit `reference_audio_cache_max_bytes` limit, defaulting to `256 * 1024 * 1024`. The cache uses `numel * element_size` for byte accounting:

```python
reference_audio_cache = LruCache(
    max_size=normalized_reference_audio_cache_size,
    max_bytes=normalized_reference_audio_cache_max_bytes,
    size_of=lambda codes: codes.numel() * codes.element_size(),
)
```

Before `put`, `reference_codes_delayed` is converted from the current list representation into a new contiguous CPU `torch.int16` tensor:

```python
def _reference_codes_to_cache_tensor(rows: list[list[int]]) -> torch.Tensor:
    return torch.as_tensor(rows, dtype=torch.int16, device="cpu").contiguous()
```

On cache hit, it converts back with `cached_codes.tolist()` to keep the current `HiggsTtsState` list-of-lists API.

However, `LruCache` now tracks `total_bytes` and evicts LRU entries until both the entry-count and byte-count limits are satisfied. Oversized single values are skipped without changing existing entries. A large-but-admissible value can still evict multiple smaller LRU entries in one `put`, we can revisit the admission policy later if needed.

I also added tests for compact tensor storage, byte-budget eviction with long reference codes, oversized-value skip behavior, generic `LruCache` byte accounting, and config passthrough for `reference_audio_cache_max_bytes`.

sglang_omni/models/higgs_tts/stages.py
nagisa-kunhah: ok, changing the `waveform is None` path to return immediately from `_cached_encode`.

sglang_omni/models/higgs_tts/stages.py
nagisa-kunhah: Addressed this as part of the compact tensor storage change above. The cache no longer stores the same Python list object returned in `payload.data["reference_codes_delayed"]`; on miss, the cache value is created through `_reference_codes_to_cache_tensor(...)`:

```python
reference_audio_cache.put(
    cache_key,
    _reference_codes_to_cache_tensor(state.reference_codes_delayed),
)
```

That helper converts the list-of-lists into a new contiguous CPU `torch.int16` tensor, so the stored cache value does not alias the returned payload. On hit, `cached_codes.tolist()` creates a fresh list for the new payload as well.

I also add the unit-test `test_higgs_audio_encoder_cache_put_isolated_from_returned_payload_mutation` to this case.

sglang_omni/models/higgs_tts/stages.py
nagisa-kunhah: Addressed this with a small encode-path refactor. The actual state-processing logic now lives in `_encode_state(state)`, so both `_encode` and `_cached_encode` can reuse it after deserializing `payload.data`:

```python
def _encode_state(state: HiggsTtsState) -> bool:
    """Encode raw reference audio and mutate ``state`` in place.

    Fills delayed reference codes, rebuilds prompt tokens, and clears the
    raw reference fields once encoding succeeds.
    """
```

On a cache miss, `_cached_encode` now reuses the already-deserialized state, calls `_encode_state(state)` directly, serializes once, and stores the cache value from that same state:

```python
if not _encode_state(state):
    return payload

payload.data = state.to_dict()
if state.reference_codes_delayed is not None:
    reference_audio_cache.put(
        cache_key,
        _reference_codes_to_cache_tensor(state.reference_codes_delayed),
    )
```

So the miss path no longer calls `_encode(payload)` and then deserializes `result.data` again just to read `reference_codes_delayed`.

One detail: `_encode_state` intentionally mutates the passed-in state, and the docstring calls that out. Let me know if you would prefer a return-value based helper instead.

nagisa-kunhah:
> Add a lightweight benchmark case; it can use a fake codec without the private checkpoint to report call counts, hash time, and payload-building time. If a public checkpoint path is available, add a real `codec.encode_reference` hit/miss latency comparison as well. Put the command and numbers back into the PR description so reviewers have evidence for this performance change.

Follow this comment, I implemented the following lightweight benchmark and coverage for the Higgs reference-audio cache path:

- `benchmarks/eval/benchmark_higgs_audio_cache.py`
- `tests/unit_test/higgs_tts/test_audio_cache_benchmark.py`

The benchmark is intended to validate the following performance semantics:

- Whether a cache hit really bypasses `codec.encode_reference`.
- How much overhead the new hash/cache logic adds on cache misses.
- Separately report payload-building, waveform-hashing, and prompt-building time, instead of only reporting total compute time.
- Whether the disabled-cache path avoids computing the cache key.

Command:

```bash
PYTHONPATH=$PWD .venv/bin/python -m benchmarks.eval.benchmark_higgs_audio_cache \
  --iterations 2000 \
  --warmup 100 \
  --waveform-samples 24000 \
  --frames 75 \
  --num-codebooks 8 \
  --codec-delay-ms 1.0
```

Environment: local `.venv`, Python 3.12.12, CPU: 13th Gen Intel(R) Core(TM) i7-13650HX

| scenario | iterations | codec_calls | hash_calls | prompt_build_calls | mean_compute_ms | mean_payload_build_ms | mean_hash_ms | mean_fake_codec_ms | mean_prompt_build_ms |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| disabled_cache_same_waveform | 2000 | 2000 | 0 | 2000 | 1.527736 | 0.014878 | 0.000000 | 1.254416 | 0.001938 |
| cache_enabled_same_waveform_hit | 2000 | 0 | 2000 | 2000 | 0.136791 | 0.005139 | 0.117829 | 0.000000 | 0.000578 |
| cache_enabled_unique_waveform_miss | 2000 | 2000 | 2000 | 2000 | 1.566358 | 0.010818 | 0.127647 | 1.188024 | 0.001407 |

Conclusions:

1. Cache hits bypass `codec.encode_reference`.

   In `cache_enabled_same_waveform_hit`, `codec_calls=0` across 2000 measured iterations. The disabled-cache baseline calls the fake codec 2000 times for the same workload. This confirms that repeated reference audio is served from the cache after prefill instead of being re-encoded.

2. Cache misses pay the expected hash/cache overhead while still doing the encode.

   In `cache_enabled_unique_waveform_miss`, every request uses a different waveform, so the benchmark reports `codec_calls=2000` and `hash_calls=2000`. Mean compute time is `1.566358 ms` versus `1.527736 ms` for `disabled_cache_same_waveform`, so the all-miss path is only slightly slower in this fake-codec run. The directly measured hash time is `0.127647 ms/request`, which is the main new per-request CPU cost on misses.

3. The benchmark makes CPU-side costs visible separately from the fake codec.

   The output reports payload-building, hash, fake-codec, and prompt-building timings independently. In the hit case, mean compute time is `0.136791 ms`, with `0.117829 ms` spent hashing and `0.000578 ms` spent in prompt building. This shows that after avoiding codec work, waveform hashing is the dominant remaining measured cost in this microbenchmark.

4. The disabled-cache path avoids cache-key hashing.

   `disabled_cache_same_waveform` reports `hash_calls=0`, while both cache-enabled scenarios report `hash_calls=2000`. The accompanying unit test also covers this behavior by asserting the disabled-cache path does not call the cache-key function.

sglang_omni/utils/lru_cache.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def __init__(
+        self,
+        max_size: int | None,
+        *,
+        copy_on_get: Callable[[V], V] | None = None,
+        max_bytes: int | None = None,
+        size_of: Callable[[V], int] | None = None,
+    ) -> None:
+        self.max_size = normalize_lru_limit(max_size)
```
Ccyest, line 34: Did you asked any maintainer in the team to make this design decision that we want to extract LRU cache out now? I believe https://github.com/sgl-project/sglang-omni/issues/512 will do the specifically. Right now our task is supporting the feature on higgs.

Ccyest:
Thanks for contribution!

As I mentioned above, here is a PR with similar feature but scope is more well defined. https://github.com/sgl-project/sglang-omni/pull/605. We will review both PR, and take one to merge. 

I would suggest to close this one and have both PR authors to get co-author on the final merge commit. (cc: @zhaochenyang20)

PR #570, opened 2026-05-25
Fix Qwen3-TTS and Voxtral-TTS serving contracts

Ratish1:
## Motivation

This PR tightens the serving contracts for the merged Qwen3-TTS and Voxtral-TTS model support. The changes focus on request-owned sampling state, SGLang scheduler handoff correctness, compact audio payloads, GPU placement during stage startup, and model-specific request validation.

The main user-visible goal is to keep Qwen3-TTS and Voxtral-TTS serving fast while avoiding hidden request-state leaks, unsupported no-op parameters, global RNG coupling, and large Python audio payloads.

## Modifications

- Updated `OmniScheduler` error ownership so request-builder, request-limit, KV-capacity, model-runner, and result-adapter failures are reported through the scheduler outbox consistently, while clearing deferred TTS request state on failure paths.
- Preserved SGLang's decode handoff contract for custom model runners by ensuring generated `output_ids` are copied back to the next decode step before returning `GenerationBatchResult`.
- Changed Qwen3-TTS seed handling from process-global RNG mutation to request-owned sampling state: public `seed` is normalized and deterministically split into independent semantic and subtalker seeds; requests without public seed receive private per-request seeds.
- Installed Qwen3-TTS semantic seeds directly on `forward_batch.sampling_info.sampling_seed` before prefill/decode sampling so the path does not depend on global deterministic inference in pinned `sglang==0.5.8`.
- Kept Qwen3-TTS subtalker sampling isolated from global CUDA RNG, including the sampled batched path, and kept semantic EOS out of `output_codes` / completion-token accounting.
- Replaced Qwen3-TTS and Voxtral-TTS large `audio_data` Python lists with compact `audio_waveform` bytes plus dtype/shape metadata through a shared audio payload helper.
- Fixed GPU stage construction for Qwen3-TTS and Voxtral-TTS by passing `gpu_id` explicitly through `factory_args`, restoring `torch.cuda.set_device(...)` and `gpu_startup_lock(...)` in the stage process path.
- Added Voxtral-TTS validation for unsupported `/v1/audio/speech` fields, kept voice embeddings on the prepared device, fixed voice cache keying, and reduced per-row sync overhead in its model runner path.
- Updated TTS docs for Qwen3-TTS reference-audio requirements and model-specific `seed` behavior.

## Root Cause

- Pinned `sglang==0.5.8` only propagates `SamplingParams.sampling_seed` into `SamplingBatchInfo.sampling_seed` under global deterministic inference. Qwen3-TTS should not enable global deterministic inference as a product-level serving fix, so the model runner now owns per-row semantic seed installation explicitly.
- The earlier sampled subtalker batching path consumed global CUDA RNG differently under concurrency. The final implementation keeps subtalker sampling request-seeded so it cannot perturb layer-0 semantic sampling.
- The custom Omni model-runner path did not fully preserve SGLang's next-token decode state update, which made custom AR runners depend on a subtly different handoff contract from upstream SGLang.
- Qwen3-TTS and Voxtral-TTS were moving generated audio through large Python lists, which inflated host-side serialization overhead.
- Removing `device="cuda:0"` from TTS configs without passing `gpu_id` made the stage-process GPU detection skip `set_device` and the startup lock. The configs now pass explicit `gpu_id` so placement and startup serialization are declarative.

## Known Qwen3-TTS Length-Cap Limitation

This PR does not add heuristic EOS forcing, retry-on-long-output, text-length caps, or non-default decoding policy changes.

The remaining rare Qwen3-TTS 0.6B `max_new_tokens=2048` artifacts were debugged with targeted replay. Fixed public seed replay produced identical c1 and c16 outputs, which rules out a c1-vs-c16 framework RNG/concurrency divergence for the captured case. Private-seed replay reproduced 2048-token hits for the same EN request; the captured bad sidecar showed a semantic-token repetition loop where EOS had rank around `1500-1800`, probability around `1e-16`, and was outside `top_k=50`. That points to a stochastic Qwen3-TTS decoding tail under the model's default sampling policy, not a stop-token checker or batching contract bug.

A future mitigation can evaluate the bad captured seed under alternate decoding policy (`top_k`, temperature, repetition penalty, retry policy), but this PR intentionally keeps Qwen's default benchmark decoding policy.

In the completed Qwen3-TTS 0.6B EN ASR rerun, raw WER was `9.39%` because the single 2048-token outlier was also the only `>50%` WER sample. Excluding that outlier gives `1.11%`, which is in the normal non-outlier quality band.

## Accuracy Tests

Remote H200 validation:

- `pytest tests/unit_test/qwen3_tts/test_pipeline.py -q`: `32 passed`
- Targeted Qwen3-TTS replay for `common_voice_en_581024-common_voice_en_581027`:
  - fixed public seed `123456789`: c1 and c16 produced the same audio hash and `79` completion tokens
  - private seed replay reproduced 2048-token length hits in both c1 and c16 runs
  - captured max-hit sidecar: semantic seed `1297050157`, subtalker seed `955620427`, finish reason `length`, EOS outside `top_k=50`

Full H200 matrix after these changes:

| model / split | samples | failed | ASR skipped | WER/CER | excl >50% | max-token outliers |
| Qwen3-TTS 0.6B EN | 1088 | 0 | 0 | 9.39% | 1.11% | 1 (`common_voice_en_31499885-common_voice_en_31499906`) |
| Qwen3-TTS 0.6B ZH | 2020 | 0 | 0 | 1.15% | 1.15% | 2 (`10002508-00000049`, `00005333-00000061`) |
| Qwen3-TTS 1.7B EN | 1088 | 0 | 0 | 0.90% | 0.90% | none |
| Qwen3-TTS 1.7B ZH | 2020 | 0 | 0 | 0.97% | 0.97% | none |
| Voxtral EN | 1088 | 0 | 0 | 1.25% | 1.25% | none |

Before/after accuracy comparison from the H200 runs:

| row | before WER/CER | after WER/CER | delta | before excl >50% | after excl >50% | delta |
| Qwen3-TTS 0.6B EN | 11.02% | 9.39% | -1.63 pp | 1.09% | 1.11% | +0.02 pp |
| Qwen3-TTS 0.6B ZH | 1.31% | 1.15% | -0.16 pp | 1.31% | 1.15% | -0.16 pp |
| Qwen3-TTS 1.7B EN | 1.04% | 0.90% | -0.14 pp | 1.04% | 0.90% | -0.14 pp |
| Qwen3-TTS 1.7B ZH | 1.36% | 0.97% | -0.39 pp | 1.36% | 0.97% | -0.39 pp |
| Voxtral EN | 1.27% | 1.25% | -0.02 pp | 1.27% | 1.25% | -0.02 pp |

Raw WER/CER improved on every row. Qwen3-TTS 0.6B EN `excl >50%` is effectively flat, with the completed rerun measuring `+0.02 pp`; the raw corpus metric still improved because the current run had one severe outlier instead of the earlier larger raw-WER impact.

## Speed Tests and Profiling

Before/after comparison from the H200 runs:

| row | before qps | after qps | qps delta | before latency mean / p95 | after latency mean / p95 | latency delta mean / p95 | before rtf | after rtf | rtf delta |
|---|---:|---:|---:|---|---|---:|---:|---:|---:|
| Qwen3-TTS 0.6B EN | 2.270 | 2.836 | +24.9% | 6.180 / 8.773s | 4.796 / 6.803s | -22.4% / -22.5% | 1.4437 | 1.1238 | -22.2% |
| Qwen3-TTS 0.6B ZH | 1.936 | 2.489 | +28.6% | 8.243 / 10.704s | 6.311 / 8.125s | -23.4% / -24.1% | 1.3585 | 1.0181 | -25.1% |
| Qwen3-TTS 1.7B EN | 2.513 | 3.492 | +39.0% | 6.339 / 9.336s | 4.559 / 6.636s | -28.1% / -28.9% | 1.5355 | 1.1077 | -27.9% |
| Qwen3-TTS 1.7B ZH | 1.931 | 2.652 | +37.3% | 8.264 / 10.703s | 6.013 / 7.828s | -27.2% / -26.9% | 1.3943 | 1.0177 | -27.0% |
| Voxtral EN | 6.705 | 7.559 | +12.7% | 2.372 / 3.662s | 2.098 / 3.178s | -11.6% / -13.2% | 0.4180 | 0.3665 | -12.3% |

Request-flow profiling from the focused Qwen3-TTS runs showed request-building stayed sub-ms while runtime remained dominated by the AR TTS engine stage. No nsys/kernel-level investigation is included in this PR because the remaining 2048-token artifacts were classified as stochastic decoding trajectories, not kernel/runtime stalls.


sglang_omni/models/qwen3_tts/config.py
```diff
@@ -27,15 +27,15 @@ class Qwen3TTSPipelineConfig(PipelineConfig):
...
+            factory_args={"dtype": "bfloat16"},
             gpu=0,
             next="vocoder",
         ),
         StageConfig(
             name="vocoder",
             process="pipeline",
             factory=f"{_PKG}.stages.create_vocoder_executor",
-            factory_args={"device": "cuda:0", "dtype": "bfloat16"},
+            factory_args={"dtype": "bfloat16"},
```
zhaochenyang20: In `sglang_omni/pipeline/stage_process.py`, `_construct_stage` resolves the stage's `gpu_id` in order: `spec.relay_config.get("gpu_id")` → `spec.factory_args.get("gpu_id")` → `spec.gpu_id` only when `_factory_args_use_cuda(spec.factory_args)` is True. `_factory_args_use_cuda` checks whether any factory_arg value starts with `cuda`. This PR changes `{"device": "cuda:0", "dtype": "bfloat16"}` into `{"dtype": "bfloat16"}` — both stages no longer carry any cuda-prefixed string and they don't inject `gpu_id` into `relay_config` or `factory_args` either — so `gpu_id` resolves to `None`, which causes (a) `torch.cuda.set_device(int(gpu_id))` to be skipped and (b) `_construct_scheduler` to take the `gpu_id is None` branch and skip `gpu_startup_lock(int(gpu_id))`. Both `tts_engine` and `vocoder` are pinned to `gpu=0` in the Qwen3-TTS pipeline (and similarly in Voxtral). They used to be serialized through that startup lock; now they race during SGLang infrastructure init + speech / audio tokenizer load, and concurrent bf16 model + KV cache + radix-tree allocations easily blow up `cuMemAlloc` on H200/H100. This directly contradicts the "GPU-aware stage placement" intent in the PR description. Also, the new `gpu_id: int | None = None` kwarg added on the factory functions is currently dead: `_normalize_spec_gpu_id_to_local_device` only rewrites it when `"gpu_id" in spec.factory_args`, but the new configs do not place it there.


The preferred fix is to inject `gpu_id` directly into `factory_args` in the config so the `_factory_args_use_cuda` detection path is no longer needed and `_normalize_spec_gpu_id_to_local_device` can rewrite it in place under TP. That also lets the newly added factory kwarg stop being dead code while restoring both the startup lock and `set_device`.

```python
# sglang_omni/models/qwen3_tts/config.py
StageSpec(
    name="tts_engine",
    process="pipeline",
    factory_args={"gpu_id": 0, "dtype": "bfloat16"},
    gpu=0,
    next="vocoder",
),
StageSpec(
    name="vocoder",
    process="pipeline",
    factory=f"{_PKG}.stages.create_vocoder_executor",
    factory_args={"gpu_id": 0, "dtype": "bfloat16"},
    gpu=0,
    terminal=True,
),

# sglang_omni/models/voxtral_tts/config.py — mirror the same change.
```

Alternatively, if you want to keep the configs terse, add a fallback in `_construct_stage`: when `gpu_id` is still `None` and `spec.gpu_id is not None`, default it to `spec.gpu_id`. That changes semantics for every stage, so it deserves a separate evaluation — I personally prefer the explicit-in-config approach above. Before merging, please add a regression test that mocks two GPU stages starting up and asserts `gpu_startup_lock` is invoked (a `monkeypatch` on the lock that counts enter/exit calls is enough).

sglang_omni/models/qwen3_tts/stages.py
```diff
@@ -144,6 +147,7 @@ def create_sglang_tts_engine_executor(
     model_path: str,
     *,
     device: str = "cuda:0",
+    gpu_id: int | None = None,
```
zhaochenyang20, line 134: `create_sglang_tts_engine_executor` / `create_vocoder_executor` (Qwen3-TTS) and `create_generation_executor` / `create_vocoder_executor` (Voxtral-TTS) all received a new `gpu_id: int | None = None` kwarg, but since stage_process no longer recognizes the cuda string in `_factory_args_use_cuda` and no path auto-injects `spec.gpu_id` into `factory_args`, this kwarg always receives `None` and lands in the dead `if gpu_id is not None: device = f"cuda:{gpu_id}"` branch. The "new kwarg + config does not pass it + nothing injects it" combination *looks* like multi-GPU flexibility but actually still hard-pins cuda:0.

sglang_omni/models/qwen3_tts/stages.py
```diff
@@ -105,18 +106,20 @@ def _load_qwen3_tts_generate_defaults(checkpoint_dir: str) -> dict[str, Any]:
     return data if isinstance(data, dict) else {}
 
 
-def _audio_to_list(audio: Any) -> list[float]:
+def _audio_waveform_payload(audio: Any) -> dict[str, Any]:
```
zhaochenyang20: This PR adds a copy of `_audio_waveform_payload` to both the Qwen3-TTS and Voxtral-TTS stages.py with identical core logic (`torch.Tensor → cpu().float().reshape(-1).numpy() → np.asarray(...).reshape(-1) → {"audio_waveform": .tobytes(), "audio_waveform_shape": [...], "audio_waveform_dtype": "float32"}`). The same payload shape is already used in `sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py:289-294` and `sglang_omni/models/ming_omni/components/talker_executor.py:228-233`. The Qwen3-TTS copy wraps the `np.asarray` call in a `try / except (TypeError, ValueError)` to raise a model-specific TypeError; Voxtral's copy omits it. That kind of "almost-identical" copy is a classic DRY smell — the next person adding a `.contiguous()` call in ming_omni will not propagate it back.

Lift it into something like `sglang_omni/scheduling/audio_io.py` or `sglang_omni/utils/audio_payload.py` with a signature `def audio_waveform_payload(audio, *, source_hint: str = "audio") -> dict[str, Any]`, where the failure path becomes `raise TypeError(f"Unsupported {source_hint} audio output type: {type(audio)}")`. Drop both Qwen3-TTS and Voxtral-TTS local copies; replacing the code2wav / talker_executor usages can be a separate PR, but at minimum the two new copies should not be born forked.

sglang_omni/models/qwen3_tts/request_builders.py
```diff
@@ -152,6 +158,12 @@ def build_qwen3_tts_state(payload: StagePayload) -> Qwen3TTSState:
         tts_params=tts_params,
         ref_text=ref_text,
     )
+    seed = tts_params["seed"] if "seed" in tts_params else params.get("seed")
+    if seed is not None:
+        raise ValueError(
+            "Qwen3-TTS Base does not support seed yet; deterministic sampled "
+            "decoding requires per-request semantic sampling."
+        )
```
zhaochenyang20: `build_qwen3_tts_state` raises immediately when `seed is not None` and then unconditionally passes `seed=None` to `Qwen3TTSState`. `sampling_seed=state.seed` has also been removed from `build_sglang_qwen3_tts_request`. So after this PR, `Qwen3TTSState.seed` is always `None`; the round-trip test only still references it for schema compatibility. Either drop the field from the dataclass and strip it in `from_dict`, or add a comment that says "deprecated; rejected at the request boundary". Leaving it as-is invites the next person to assume seed still has semantics.


```python
# sglang_omni/models/qwen3_tts/payload_types.py
@dataclass
class Qwen3TTSState:
    ...
    # seed: int | None = None  # remove
    # If you must keep the field for serialization compat, leave a one-line note:
    # seed: int | None = None  # deprecated; rejected at build_qwen3_tts_state
```

sglang_omni/models/qwen3_tts/sglang_model.py
```diff
@@ -708,45 +774,148 @@ def _code_predictor_forward_incremental(
...
+        else:
+            base = semantic_positions.to(device=device, dtype=torch.long)
+            if base.ndim == 2:
+                if base.shape != (batch_size, seq_len):
+                    raise ValueError("Qwen3-TTS subtalker positions shape mismatch")
+                return base
+            if base.ndim != 1 or base.shape[0] != batch_size:
+                raise ValueError("Qwen3-TTS subtalker positions shape mismatch")
+        offsets = torch.arange(seq_len, device=device, dtype=torch.long)
+        return base.unsqueeze(1) + offsets.unsqueeze(0)
```
zhaochenyang20, line 817: `_sample_subtalker_token(logits, layer_idx, semantic_positions)` now does one thing: forward the same three arguments to `_sample_subtalker_token_batch`. There is no caller outside the unit tests, and the unit tests already exercise `_sample_subtalker_token` directly, so there is no need for two layers of indirection. Drop the `_batch` suffix and merge them. If a single-row fast path materializes later, split again then.


```python
def _sample_subtalker_token(
    self,
    logits: torch.Tensor,
    layer_idx: int = 0,
    *,
    semantic_positions: torch.Tensor | None = None,
) -> torch.Tensor:
    # ... body of current _sample_subtalker_token_batch ...
```

sglang_omni/models/qwen3_tts/sglang_model.py
```diff
@@ -19,12 +20,21 @@
 from sglang_omni.models.qwen3_omni.components.thinker_model import (
     Qwen3OmniMoeThinkerTextAttention,
 )
+from sglang_omni.models.qwen3_tts.request_builders import Qwen3TTSSGLangRequestData
```
zhaochenyang20: The model implementation `sglang_model.py` previously did not depend on `request_builders.py` (scheduler-side code). This PR adds `from sglang_omni.models.qwen3_tts.request_builders import Qwen3TTSSGLangRequestData` so `prepare_decode_buffers` can do an `isinstance(data, Qwen3TTSSGLangRequestData)` runtime check. That import couples the model layer to the scheduler layer; if someone later wants to instantiate this model under a different scheduler or in a standalone benchmark, they pull SGLang in transitively.

Replace the nominal `isinstance` check with a structural one — for example `if not all(hasattr(data, attr) for attr in ("subtalker_dosample", "subtalker_temperature", "subtalker_top_p", "subtalker_top_k", "subtalker_sampling_seed")): raise TypeError(...)` — or define a `Protocol` in `sglang_omni.models.qwen3_tts.payload_types` that captures the 5 fields, and have both `sglang_model.py` and `request_builders.py` depend on that Protocol instead of each other. At the current scale you could also leave it but mark a TODO to decouple later.

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

MaTwickenham:
Hi @Ratish1 , I tried to replay this on RTX4090 with the captured seeds:

- semantic_sampling_seed=1297050157
- subtalker_sampling_seed=955620427

The seeds appear to be propagated correctly into the Qwen3-TTS semantic and subtalker sampling paths, but I’m not able to reproduce the max-length repetition
reliably. In my local runs, the same prompt/config usually exits normally around 70-120 completion tokens.

Were you able to reproduce the repetition deterministically with only these two seeds? Or are there any other seeds / runtime settings that also need to be
fixed, such as the SGLang server random_seed, torch/cuda deterministic settings, CUDA graph / torch.compile state, or a specific model revision?

Ratish1:
> Hi @Ratish1 , I tried to replay this on RTX4090 with the captured seeds:
> 
> * semantic_sampling_seed=1297050157
> * subtalker_sampling_seed=955620427
> 
> The seeds appear to be propagated correctly into the Qwen3-TTS semantic and subtalker sampling paths, but I’m not able to reproduce the max-length repetition reliably. In my local runs, the same prompt/config usually exits normally around 70-120 completion tokens.
> 
> Were you able to reproduce the repetition deterministically with only these two seeds? Or are there any other seeds / runtime settings that also need to be fixed, such as the SGLang server random_seed, torch/cuda deterministic settings, CUDA graph / torch.compile state, or a specific model revision?

Not sure @MaTwickenham , maybe the issue has been fixed in the latest upstream already?. I didn't do anything special for testing, just noted down what I observed.

PR #534, opened 2026-05-24
[Benchmark] Add Higgs TTS SeedTTS benchmark

estellaliu233:
## Summary
This PR reuses the general SeedTTS TTS benchmark for Higgs TTS instead of keeping a separate Higgs-only benchmark path. The benchmark now keeps the original default payload format for existing TTS models (`ref_audio` / `ref_text`) and lets Higgs explicitly use its voice-cloning reference format via `--ref-format references`.

Changes:
  - Reuse `benchmarks/eval/benchmark_tts_seedtts.py` for Higgs TTS.
  - Add `--ref-format {flat,references}` to the general TTS benchmark:
    - default `flat` preserves existing S2-Pro behavior.
    - `references` sends `references=[{audio_path, text}]` for Higgs TTS.
  - Add Higgs TTS full-set SeedTTS results to the reference tables in `benchmark_tts_seedtts.py`.

## Profiling Results

  CUDA Graph was enabled in the benchmarked Higgs TTS runtime:

  disable_cuda_graph = False
  cuda_graph_max_bs = 32
  max_running_requests = 16

  Not every possible acceleration path was enabled:

  - CUDA Graph: enabled
  - RadixCache: enabled and verified by cache hit metrics
  - overlap schedule: disabled
  - torch.compile: disabled

  Torch trace aggregate:

  cuda_runtime: 16749.871 ms
  kernel:       11087.027 ms
  gpu_memcpy:     101.930 ms
  cpu_op:          23.373 ms

  Stage-level profiling:

  tts_engine avg:    637.752 ms
  vocoder avg:       103.129 ms
  audio_encoder avg:  50.996 ms
  preprocessing avg:   1.087 ms

  Concurrency sweep:

  C=1 throughput: 1.355 req/s, latency mean: 0.738s
  C=4 throughput: 4.788 req/s, latency mean: 0.825s
  C=8 throughput: 7.828 req/s, latency mean: 0.970s

  Cache benchmark:

  hit_request_cache_hit_rate:      1.0
  hit_token_cache_hit_rate_mean:   0.8979
  hit_cached_tokens_mean:          124.21

  miss_request_cache_hit_rate:     0.1875
  miss_token_cache_hit_rate_mean:  0.1693
  miss_cached_tokens_mean:         26.38

  ## Profiling Conclusion

The dominant cost is TTS decode runtime overhead from many small kernels plus frequent CPU/GPU synchronization.

  The most promising runtime optimization directions are:

[...]

PR #572, opened 2026-05-25
[Higgs]: batch the 3 per-step D2H syncs into 1 in _collect_step_outputs_cg

### LINKED ISSUE #564: [Feat] Higgs TTS: batch the 3 per-step D2H syncs into 1 in _collect_step_outputs_cg
## Motivation

PR #534 profiling shows a clear `cuda_runtime > kernel` gap in the Higgs TTS decode loop with CUDA Graph enabled:

```
cuda_runtime: 16749 ms
kernel:       11087 ms
gap:           5662 ms   (~34% of cuda_runtime)
```

This gap means the CPU spent ~5.6s in CUDA API calls that did not result in GPU compute — the most likely culprit being `cudaStreamSynchronize` waits triggered by per-step D2H copies in `_collect_step_outputs_cg`.

Each decode step currently issues **3 separate D2H syncs** to pull per-row state back to CPU before the Python append loop:

```python
# sglang_omni/models/higgs_tts/model_runner.py
was_done_cpu        = model._cg_was_done[:n_real].cpu().tolist()           # sync 1
codes_BN_cpu        = model._cg_codes_BN[:n_real].detach().cpu().clone()   # sync 2
gen_done_after_cpu  = model._cg_active_generation_done[:n_real].cpu().tolist()  # sync 3
```

Each sync stalls the pipeline (~30–50 µs of GPU/CPU coordination overhead). Over a typical 1000-step generation, that is ~100 ms of avoidable stalls per request — roughly **5–15% of end-to-end latency**.

### Expected impact (rough estimate)

- Per decode step: ~2 redundant sync points × ~30–50 µs ≈ 60–100 µs of stall per step.
- Typical 500–1000 step generation: ~30–100 ms per request.
- End-to-end latency: order of **~5–10%** at low concurrency; less at higher concurrency.

These numbers come from typical CUDA sync overheads and the 5.6 s gap observed in PR #534; the actual fix will need A/B benchmarking to confirm.

## References

- PR #534 — Higgs TTS SeedTTS benchmark and profiling results
- Affected file: `sglang_omni/models/higgs_tts/model_runner.py`, function `_collect_step_outputs_cg`
- PR #503 — the CUDA Graph capture work that introduced the current 3-sync pattern
- Tracking issue: #478

JiaxinD:
## Summary

`_collect_step_outputs_cg` reads per-step decode state back to the host with
**three separate** `.cpu()` / `.tolist()` calls — i.e. three
`cudaStreamSynchronize`s on the decode path every step. This PR coalesces them
into a **single** D2H through a pre-allocated staging buffer (3 → 1), with
byte-identical outputs.

Two things ship here:

1. A correctness-preserving, latency-neutral **cleanup** that cuts per-step D2H
   syncs 3 → 1 (fewer CUDA API calls, one explicit buffer, and a building block
   that pays off once overlap scheduling is enabled — see follow-up).
2. The instrumentation it motivated, which **quantifies exactly where the Higgs
   decode overhead lives** (issue #564) and redirects the next optimization at
   the right target. That analysis is the most valuable part of this PR and is
   written up below.

Closes #564 (with a corrected problem statement — details in *Analysis*).

## Motivation

With CUDA Graph enabled, the AR decode loop pulls per-step state back to CPU via:

```python
was_done_cpu       = model._cg_was_done[:n_real].cpu().tolist()                 # sync 1
codes_BN_cpu       = model._cg_codes_BN[:n_real].detach().cpu().clone()         # sync 2
gen_done_after_cpu = model._cg_active_generation_done[:n_real].cpu().tolist()   # sync 3
```

PR #534 profiling reported a large `cuda_runtime > kernel` gap (~34%), and these
three per-step syncs were the suspected contributor. This PR removes the
redundancy and — importantly — measures whether that suspicion holds.

## Change

Pack the three reads into a pre-allocated `[pool_size, num_codebooks + 2]` `long`
staging buffer (`_cg_collect_staging`, mirroring `_cg_codes_BN`): columns
`0:num_codebooks` hold `codes_BN`, and the last two hold `was_done` and
`active_generation_done` as 0/1. Three async GPU-side writes fill the slice, a
**single** `.cpu()` pulls it back, and the split + cast happen on the host. The
buffer is allocated once at init (not per step) to keep the allocation off the
decode hot path; it lives only in this eager post-decode hook, never inside the
captured graph.

## Correctness

1. **Per-step isolation** — `.cpu()` returns a fresh host tensor each step, so the
   code rows appended to `output_codes` stay isolated from the next step's
   overwrite of the staging buffer (the same guarantee the old `.clone()` gave).
2. **Bool round-trip** — the buffer is `long`, so both flags ride as 0/1 and are
   cast back with `.bool().tolist()`. The new
   `test_higgs_model_runner_collect_cg_mixed_batch` asserts the exact list
   `[False, False, False, True]`, locking this.
3. **No new downstream syncs** — `int(codes_N[0].item())` now reads an
   already-host tensor, so it adds no D2H.

## Verification

- **Unit tests** — `tests/unit_test/higgs_tts/` 11/11 pass, including the new
  `test_higgs_model_runner_collect_cg_mixed_batch` (chunked / was-done / active
  rows).
- **D2H count (deterministic, profiler)** — per call, Memcpy DtoH events drop
  **3 → 1** (baseline 150/50 calls, this PR 50/50); H2D unchanged.
- **End-to-end (SeedTTS EN, H200, concurrency 1)** — two independent clean
  matched A/Bs (this PR vs `main` = the baseline *without* this change; same KV
  config / same card, identical token counts), n=10 median engine time:

  ```
  host  case             baseline(3-sync)  this PR(1-sync)   improvement
  ion9  SHORT (66 tok)    0.3290s           0.3265s          +0.8%
  ion9  LONG  (1322 tok)  6.4589s           6.4548s          +0.06%
  ion8  SHORT (66 tok)    0.3254s           0.3248s          +0.2%
  ion8  LONG  (1322 tok)  6.4330s           6.4142s          +0.3%
  out_tok identical for both builds: SHORT=66, LONG=1322 (fair comparison)
  ```

  Latency-neutral — every delta is under ~1% and below the run-to-run noise
  floor. This is **expected and bounded**, for the reason measured next.

## Analysis — where the Higgs decode overhead actually lives

To turn "why is the A/B flat?" into a real answer (rather than a guess), I
instrumented the baseline decode loop on an exclusive H200 and timed, per step:
each of the 3 `.cpu()` calls, and the full collect-to-collect wall interval
(~6000 steps, rock-steady).

**Per decode step (4.88 ms):**

```
  GPU forward compute    ████████████████████████████████████  3.72 ms  ~76%
  CPU serialization gap   ███████████                           1.10 ms  ~24%   <- GPU idle here
    of which removable D2H  ▏                                   0.045 ms  ~0.9%  <- what this PR cuts
```

**One step on the timeline:**

```
                0ms                       3.72ms        4.88ms
  GPU stream    |==== forward (busy) =====|____ idle ~1.1ms ____|  -> next step
  CPU thread    |· blocked on call1 sync ·|c2|c3| collect-loop + prepare + schedule |
                                            ^ 45us (this PR)        ^ GPU has nothing queued
```

**Which sync blocks (the 3 `.cpu()` calls, isolated):**

```
  call 1  _cg_was_done                 3713 us  (98.7%)  <- waits for the step's GPU compute
  call 2  _cg_codes_BN                   31 us  ( 0.8%)
  call 3  _cg_active_generation_done     16 us  ( 0.4%)
```

**Cross-check** — `nvidia-smi` GPU utilization on the exclusive card was median
**82%** (≈18% idle), consistent with the ~24% gap above (82% is the truer
GPU-busy figure; `call1` slightly under-counts it).

**Findings:**

- Only the **first** `.cpu()` blocks — it absorbs the per-step GPU forward
  (~3.7 ms, memory-bandwidth bound). The 2nd and 3rd hit an already-drained queue
  and cost ~45 µs combined. So the syncs this PR removes are the *cheap* ones; the
  expensive sync is the unavoidable wait for compute and remains in the single
  combined `.cpu()`.
- That puts a hard ceiling on this PR's e2e win: **45 µs / 4.88 ms ≈ 0.9%** — which
  is exactly why both A/Bs land within noise. Measurement, mechanism, and ceiling
  all agree.
- The decode is **~76% genuine GPU forward + ~18–24% CPU serialization bubble**
  (GPU idle while the host runs the per-step collect / prepare / scheduling). It
  is *not* the 34% from #534 — that figure is a CPU-side `cuda_runtime/kernel`
  ratio, not GPU idle time; the real idle is ~20%.

## Follow-up — the real lever for #564

The recoverable overhead is the **~18–24% serialization bubble**, and the fix is
**overlap / async decode**: run the host-side scheduling of step N+1 *while* the
GPU computes step N, hiding the ~1.1 ms CPU work behind the ~3.7 ms forward and
pushing utilization toward ~95%+ (≈18–24% throughput at concurrency 1).

Higgs currently sets `disable_overlap_schedule=True`, but most of the
infrastructure is already in place: the `_cg_active_*` shadow state lives on
device and token feedback is on-device (`_decode_step_embeds_cg` builds the next
input from `_cg_active_last_codes` without a host round-trip). The remaining work
is moving the per-step host bookkeeping in `_collect_step_outputs_cg` off the
critical path. **This PR's 3 → 1 coalescing is a clean prerequisite for that** —
with overlap on, fewer syncs is exactly what stops them becoming pipeline bubbles.

The other ~76% is real forward compute; shrinking it is a separate axis
(FP8 weights for the bandwidth-bound decode, fusing the 8 codebook heads).

## Tradeoff

A deterministic D2H sync-count reduction (3 → 1) with identical outputs and no
measurable latency regression. The standalone e2e gain is sub-noise by
construction (~0.9% ceiling, measured above); the value is the cleaner sync path,
the building block for overlap, and the quantified direction it gives #564.


sglang_omni/models/higgs_tts/model.py
```diff
@@ -170,6 +170,11 @@ def __init__(
         self._cg_codes_BN = torch.zeros(
             pool_size, num_codebooks, dtype=torch.long, device=cg_device
         )
+        # Packs codes_BN | was_done | active_generation_done into one buffer so
+        # _collect_step_outputs_cg can pull per-step state back with a single D2H.
```
Ccyest: suggest fix: 
`Note(Jiaxin): Packs codes_BN | was_done | active_generation_done into one buffer`

sglang_omni/models/higgs_tts/model_runner.py
```diff
@@ -148,9 +148,19 @@ def _collect_step_outputs_cg(
         pool.generation_done[rows_t] = model._cg_active_generation_done[:n_real]
         pool.last_codes[rows_t] = model._cg_active_last_codes[:n_real]
 
-        was_done_cpu = model._cg_was_done[:n_real].cpu().tolist()
-        codes_BN_cpu = model._cg_codes_BN[:n_real].detach().cpu().clone()
-        gen_done_after_cpu = model._cg_active_generation_done[:n_real].cpu().tolist()
+        # Batch codes_BN | was_done | active_generation_done into the staging
+        # buffer so the per-step read back to CPU costs a single D2H sync (the
+        # GPU-side packs are async); .cpu() returns a fresh per-step copy, so the
+        # appended code rows below stay isolated from the next step's overwrite.
```
Ccyest: would suggest to trim this comment as well

sglang_omni/models/higgs_tts/model.py
JiaxinD: Done — switched to the `Note(Jiaxin):` one-liner in be93a97. Thanks!

sglang_omni/models/higgs_tts/model_runner.py
JiaxinD: Done — trimmed to a single `Note(Jiaxin):` line in be93a97.

zhaochenyang20 (approved):

estellaliu233:
Hi — I've been putting together a write-up on diagnosing host-bound serving: how to tell a GPU bubble apart from a slow kernel, which workloads are most exposed to it, and what engines and operators can each do about it. I used this PR (with the hypothesis from #564) as the **synchronization-point** case:

I quoted the per-call breakdown from the Analysis section — the 4.88 ms step decomposed into 3.72 ms of GPU forward plus a 1.10 ms host serialization gap, the 3713 / 31 / 16 µs split across the three `.cpu()` calls, and the resulting ~0.9% ceiling. The point I took from it is that the cost of a synchronization depends on what it waits for, not how many of them there are — which is why the A/B landed within noise and why the analysis, not the coalescing, was the useful part.

If I've misread the context anywhere, or overstated a conclusion the PR didn't actually measure, I'd be glad to correct it — thanks for the write-up, it was unusually clear.

Issue #569, opened 2026-05-25
[Perf] Higgs TTS: batched vocoder decode


Ccyest:
## Motivation

The vocoder stage (`create_vocoder_executor`) currently processes requests one at a time. Even though `SimpleScheduler` collects up to `max_batch_size=4` messages, no `batch_compute_fn` is provided, so it falls back to a serial loop — calling `codec.decode(codes_TN)` once per request.

Each `codec.decode` call converts a single `[T, N]` tensor into `[1, N, T]` and runs the full decoder (dequantize → fc2 → acoustic_decoder ConvTranspose stack). At low batch sizes the GPU is severely underutilized because the ConvTranspose kernels are memory-bound and their launch overhead dominates.

The underlying `HiggsAudioV2TokenizerModel.decode()` already accepts batched input `(batch_size, num_quantizers, codes_length)` — we just don't use it. Padding variable-length code sequences and running a single batched decode should reduce per-request vocoder latency roughly proportional to the batch size.

Qwen3-TTS already implements this pattern via `_vocode_batch` + `batch_compute_fn` in `sglang_omni/models/qwen3_tts/stages.py`.

## Current path (single decode per request)

```python
# sglang_omni/models/higgs_tts/audio_codec.py
def decode(self, codes_TN: torch.Tensor) -> torch.Tensor:
    codes_BNT = codes_TN.transpose(0, 1).unsqueeze(0)          # [1, N, T]
    return self.model.decode(codes_BNT).audio_values.squeeze()  # [L]

# sglang_omni/models/higgs_tts/stages.py — _vocode()
codes_TN = reverse_delay_pattern(delayed_LN)
waveform = codec.decode(codes_TN)                              # one at a time
```

`SimpleScheduler` is instantiated with `max_batch_size=4` but without `batch_compute_fn`, so the scheduler's batch collection is wasted.

## Proposed change

1. Add a `decode_batch` method to `HiggsAudioCodec`:

```python
def decode_batch(self, codes_list: list[torch.Tensor]) -> list[torch.Tensor]:
    """Decode a batch of [T_i, N] code tensors → list of [L_i] waveforms."""
    lengths = [c.shape[0] for c in codes_list]
    T_max = max(lengths)
    # Pad to uniform length, stack → [B, N, T_max]
    padded = torch.stack([
        F.pad(c.transpose(0, 1), (0, T_max - c.shape[0]))
        for c in codes_list
    ]).to(device=self.device, dtype=torch.long)
[...]

PR #574, opened 2026-05-25
Add higgs batched vocoder decode #569

SandyLuXY:
## Motivation

The Higgs TTS vocoder stage already uses `SimpleScheduler(max_batch_size=4)`, which collects up to 4 requests before dispatching. However, no `batch_compute_fn` is provided, so each request is decoded **serially** through the codec — one `[T, N] -> [L]` decode at a time.

The underlying codec model (`HiggsAudioV2TokenizerModel.decode`) natively supports batched input `[B, N, T] -> [B, 1, L]`. We're just not using it.

This PR closes that gap: bucket variable-length code tensors by equal length, batch same-length items into a single forward pass, and fall back to per-item decode for mixed lengths. At concurrency >= 2, this directly improves vocoder throughput with no regression at concurrency = 1.

## Modifications

Three files changed, following the same pattern already proven in the Qwen3-TTS vocoder path.

### 1. `sglang_omni/models/higgs_tts/audio_codec.py`

Added `HiggsAudioCodec.decode_batch(codes_list)`:

- Accepts a list of variable-length `[T_i, N]` code tensors
- Buckets items by equal `T` — only same-length items are stacked into `[B, N, T]` and decoded in a single forward pass; mixed lengths get separate per-bucket calls
- The non-causal DAC decoder corrupts shorter items when zero-padded to a longer `T_max`, so bucketing avoids padding artifacts entirely
- Single-element buckets delegate to the existing `decode()` to avoid stacking overhead

### 2. `sglang_omni/models/higgs_tts/stages.py`

Refactored `create_vocoder_executor` from one monolithic function into composable helpers (mirroring the Qwen3-TTS pattern):

| Function | Role |
|---|---|
| `_prepare_vocoder_item` | Deserialize state, reverse delay pattern, clamp OOV codes. Returns `(state, codes_TN \| None)` |
| `_store_vocoder_result` | Write waveform + metadata back into payload |
| `_vocode` | Single-item path (unchanged behavior) |
| `_vocode_batch` | Batch path: prep all items, filter valid ones, call `codec.decode_batch`, store results |

Wired `batch_compute_fn=_vocode_batch` into the `SimpleScheduler` return.

Items with empty or too-short delayed codes (< `num_codebooks` rows) are filtered out of the batch and receive empty `audio_data` — no crash, no wasted GPU work.

### 3. `tests/unit_test/higgs_tts/test_pipeline.py`

Added four tests:

- **`test_higgs_tts_vocoder_batches_decode_requests`** — Mocks the codec, sends 2 payloads through `_batch_fn`, asserts `decode_batch` is called once with batch size 2 (not twice with size 1). Validates output format, sample rate, and usage metadata.
- **`test_higgs_tts_vocoder_batch_handles_empty_items`** — Sends 3 payloads (empty, too-short, valid), asserts only the valid item is batched and invalid items get `audio_data == []`.
- **`test_decode_batch_buckets_by_length`** — Verifies same-T items batch into one forward call while mixed-T items get separate per-bucket calls.
- **`test_decode_batch_bit_exact_with_single_decode`** — Confirms batched decode produces identical output to individual serial decode.

## Related Issues

#569

## Benchmark Results

Full SeedTTS EN evaluation (1088 samples) on **1x NVIDIA H200**, temperature=0.8, top-k=50.
Base commit: `58d7e53` (main). Model: `boson-sglang/higgs-audio-v3-tts-4b-base`.

### Speed

| Metric | c=1 main | c=1 batch | c=16 main | c=16 batch | c=32 main | c=32 batch |
|--------|----------|-----------|-----------|------------|-----------|------------|
| Latency mean (s) | 0.665 | 0.641 | 1.886 | 1.749 | 2.961 | 3.102 |
| Latency median (s) | 0.599 | 0.585 | 1.678 | 1.636 | 2.842 | 3.073 |
| Latency p95 (s) | 0.887 | 0.860 | 2.634 | 2.319 | 3.678 | 3.991 |
| Latency p99 (s) | 1.105 | 1.043 | 3.444 | 2.769 | 4.189 | 4.367 |
| RTF mean | 0.148 | 0.144 | 0.432 | 0.412 | 0.719 | 0.774 |
| **Throughput (req/s)** | **1.503** | **1.559** | **8.447** | **9.106** | **10.007** | **10.220** |
| Output tokens (mean) | 123 | 121 | 127 | 120 | 121 | 113 |

### Quality

| Metric | c=1 main | c=1 batch | c=16 main | c=16 batch | c=32 main | c=32 batch |
|--------|----------|-----------|-----------|------------|-----------|------------|
| WER corpus (excl >50%) | 1.28% | 1.18% | 1.28% | 1.27% | 1.38% | 1.21% |
| >50% WER samples | 2 | 2 | 4 | 1 | 5 | 0 |

### Throughput delta

| Concurrency | main (req/s) | batch (req/s) | Delta |
|-------------|--------------|---------------|-------|
| 1 | 1.503 | 1.559 | **+3.7%** |
| 16 | 8.447 | 9.106 | **+7.8%** |
| 32 | 10.007 | 10.220 | **+2.1%** |

### Analysis

**Quality:** No regression. WER (excl >50%) is within noise across all concurrency levels (1.18–1.28% batch vs 1.28–1.38% main). The >50% WER outliers are stochastic TTS hallucinations unrelated to the vocoder — the batch branch actually has fewer. Speaker similarity is stable at ~66.4 across concurrency levels.

**Speed (c=1):** +3.7% throughput, latency improves across all percentiles. At c=1 the vocoder rarely batches, so this is expected to be ~neutral; the small gain is likely run-to-run variance.

**Speed (c=16):** +7.8% throughput (8.45 → 9.11 req/s), p99 latency drops 20% (3.44s → 2.77s). The vocoder batching benefit is clearly visible here — multiple requests land in the vocoder queue simultaneously and decode together.

**Speed (c=32):** +2.1% throughput (10.01 → 10.22 req/s). The throughput gain is smaller because the AR engine is the bottleneck at high concurrency, not the vocoder. Mean latency is slightly higher (+4.8%), but this is confounded by the output token difference (113 vs 121 mean) — the main run generated longer audio per request, so shorter audio in the batch run should have *lower* latency all else equal; the fact that it doesn't suggests this delta is within stochastic noise.
## Checklist

- [x] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/models/higgs_tts/audio_codec.py
+++ b/sglang_omni/models/higgs_tts/audio_codec.py
@@ -225,9 +225,9 @@ class HiggsAudioCodec:
                 codes_BNT = stacked.transpose(1, 2).to(
                     device=self.device, dtype=torch.long
                 )
-                audio = self.model.decode(codes_BNT).audio_values
+                audio = self.model.decode(codes_BNT).audio_values.cpu()
                 for j, idx in enumerate(indices):
-                    results[idx] = audio[j, 0].cpu()
+                    results[idx] = audio[j, 0]
 
         return results  # type: ignore[return-value]
 
--- a/sglang_omni/models/higgs_tts/audio_codec.py
+++ b/sglang_omni/models/higgs_tts/audio_codec.py
@@ -229,7 +229,12 @@ class HiggsAudioCodec:
                 for j, idx in enumerate(indices):
                     results[idx] = audio[j, 0]
 
-        return results  # type: ignore[return-value]
+        out: list[torch.Tensor] = []
+        for i, result in enumerate(results):
+            if result is None:
+                raise RuntimeError(f"decode_batch did not produce audio for item {i}")
+            out.append(result)
+        return out
 
 
 __all__ = ["HiggsAudioCodec"]
--- a/sglang_omni/models/higgs_tts/audio_codec.py
+++ b/sglang_omni/models/higgs_tts/audio_codec.py
@@ -216,12 +216,6 @@ class HiggsAudioCodec:
         for i, c in enumerate(codes_list):
             buckets.setdefault(c.shape[0], []).append(i)
 
-        if len(buckets) == 1:
-            stacked = torch.stack(codes_list)
-            codes_BNT = stacked.transpose(1, 2).to(device=self.device, dtype=torch.long)
-            audio = self.model.decode(codes_BNT).audio_values
-            return [audio[i, 0].cpu() for i in range(len(codes_list))]
-
         results: list[torch.Tensor | None] = [None] * len(codes_list)
         for indices in buckets.values():
             if len(indices) == 1:
--- a/sglang_omni/models/higgs_tts/audio_codec.py
+++ b/sglang_omni/models/higgs_tts/audio_codec.py
@@ -201,27 +201,41 @@ class HiggsAudioCodec:
 
     @torch.no_grad()
     def decode_batch(self, codes_list: list[torch.Tensor]) -> list[torch.Tensor]:
-        """Batch-decode variable-length ``[T_i, N]`` tensors into ``[L_i]`` waveforms."""
+        """Batch-decode variable-length ``[T_i, N]`` tensors into ``[L_i]`` waveforms.
+
+        Only same-length items are batched into a single forward pass.
+        The non-causal DAC decoder corrupts shorter items when padded to
+        a longer ``T_max``, so mixed-length inputs are decoded per-bucket.
+        """
         if not codes_list:
             return []
         if len(codes_list) == 1:
             return [self.decode(codes_list[0])]
 
-        lengths = [c.shape[0] for c in codes_list]
-        T_max = max(lengths)
-        padded = torch.stack(
-            [F.pad(c, (0, 0, 0, T_max - c.shape[0])) for c in codes_list]
-        )
-        codes_BNT = padded.transpose(1, 2).to(device=self.device, dtype=torch.long)
-        audio = self.model.decode(codes_BNT).audio_values
+        buckets: dict[int, list[int]] = {}
+        for i, c in enumerate(codes_list):
+            buckets.setdefault(c.shape[0], []).append(i)
 
-        # Output length is affine in T (ConvTranspose1d composition):
-        # L = hop * T + offset.  Derive offset from the (T_max, L_max) pair.
-        L_max = audio.shape[-1]
-        hop = self.model.config.hop_length
-        offset = L_max - hop * T_max
+        if len(buckets) == 1:
+            stacked = torch.stack(codes_list)
+            codes_BNT = stacked.transpose(1, 2).to(device=self.device, dtype=torch.long)
+            audio = self.model.decode(codes_BNT).audio_values
+            return [audio[i, 0].cpu() for i in range(len(codes_list))]
 
-        return [audio[i, 0, : hop * t + offset].cpu() for i, t in enumerate(lengths)]
+        results: list[torch.Tensor | None] = [None] * len(codes_list)
+        for indices in buckets.values():
+            if len(indices) == 1:
+                results[indices[0]] = self.decode(codes_list[indices[0]])
+            else:
+                stacked = torch.stack([codes_list[i] for i in indices])
+                codes_BNT = stacked.transpose(1, 2).to(
+                    device=self.device, dtype=torch.long
+                )
+                audio = self.model.decode(codes_BNT).audio_values
+                for j, idx in enumerate(indices):
+                    results[idx] = audio[j, 0].cpu()
+
+        return results  # type: ignore[return-value]
 
 
 __all__ = ["HiggsAudioCodec"]
--- a/sglang_omni/models/higgs_tts/audio_codec.py
+++ b/sglang_omni/models/higgs_tts/audio_codec.py
@@ -199,5 +199,29 @@ class HiggsAudioCodec:
         )
         return self.model.decode(codes_BNT).audio_values.squeeze(0).squeeze(0).cpu()
 
+    @torch.no_grad()
+    def decode_batch(self, codes_list: list[torch.Tensor]) -> list[torch.Tensor]:
+        """Batch-decode variable-length ``[T_i, N]`` tensors into ``[L_i]`` waveforms."""
+        if not codes_list:
+            return []
+        if len(codes_list) == 1:
+            return [self.decode(codes_list[0])]
+
+        lengths = [c.shape[0] for c in codes_list]
+        T_max = max(lengths)
+        padded = torch.stack(
+            [F.pad(c, (0, 0, 0, T_max - c.shape[0])) for c in codes_list]
+        )
+        codes_BNT = padded.transpose(1, 2).to(device=self.device, dtype=torch.long)
+        audio = self.model.decode(codes_BNT).audio_values
+
+        # Output length is affine in T (ConvTranspose1d composition):
+        # L = hop * T + offset.  Derive offset from the (T_max, L_max) pair.
+        L_max = audio.shape[-1]
+        hop = self.model.config.hop_length
+        offset = L_max - hop * T_max
+
+        return [audio[i, 0, : hop * t + offset].cpu() for i, t in enumerate(lengths)]
+
 
 __all__ = ["HiggsAudioCodec"]
--- a/sglang_omni/models/higgs_tts/stages.py
+++ b/sglang_omni/models/higgs_tts/stages.py
@@ -367,6 +367,11 @@ def create_vocoder_executor(
         if valid:
             indices, codes_list = zip(*valid)
             wavs = codec.decode_batch(list(codes_list))
+            if len(wavs) != len(valid):
+                raise RuntimeError(
+                    f"Higgs vocoder decode_batch returned {len(wavs)} audios "
+                    f"for {len(valid)} requests"
+                )
             for idx, wav in zip(indices, wavs):
                 waveforms[idx] = wav
         return [
--- a/sglang_omni/models/higgs_tts/stages.py
+++ b/sglang_omni/models/higgs_tts/stages.py
@@ -370,8 +370,8 @@ def create_vocoder_executor(
             for idx, wav in zip(indices, wavs):
                 waveforms[idx] = wav
         return [
-            _store_vocoder_result(p, state, waveforms[i])
-            for i, (p, (state, _)) in enumerate(zip(payloads, items))
+            _store_vocoder_result(payload, state, wav)
+            for payload, (state, _), wav in zip(payloads, items, waveforms)
         ]
 
     return SimpleScheduler(
--- a/sglang_omni/models/higgs_tts/stages.py
+++ b/sglang_omni/models/higgs_tts/stages.py
@@ -316,33 +316,32 @@ def create_vocoder_executor(
     codec = get_or_load_codec(checkpoint_dir, device, dtype)
     sample_rate = HiggsAudioCodec.SAMPLE_RATE
 
-    def _vocode(payload: StagePayload) -> StagePayload:
+    def _prepare_vocoder_item(
+        payload: StagePayload,
+    ) -> tuple[HiggsTtsState, torch.Tensor | None]:
         state = HiggsTtsState.from_dict(payload.data)
         delayed_rows = state.output_codes_delayed
-
         if not delayed_rows:
-            payload.data["audio_data"] = []
-            payload.data["sample_rate"] = sample_rate
-            payload.data["modality"] = "audio"
-            return payload
-
+            return state, None
         delayed_LN = torch.tensor(delayed_rows, dtype=torch.long)
-        N = state.num_codebooks
-        if delayed_LN.shape[0] < N:
-            payload.data["audio_data"] = []
-            payload.data["sample_rate"] = sample_rate
-            payload.data["modality"] = "audio"
-            return payload
-
+        if delayed_LN.shape[0] < state.num_codebooks:
+            return state, None
         codes_TN = reverse_delay_pattern(delayed_LN)
-        codec_vocab = state.codebook_size - 2  # 1026 - BOC - EOC
-        codes_TN = torch.where(
+        codec_vocab = state.codebook_size - 2
+        return state, torch.where(
             codes_TN >= codec_vocab, torch.zeros_like(codes_TN), codes_TN
         )
-        waveform = codec.decode(codes_TN)
-        audio_np = waveform.detach().to(torch.float32).cpu().numpy()
 
-        payload.data["audio_data"] = audio_np.tolist()
+    def _store_vocoder_result(
+        payload: StagePayload,
+        state: HiggsTtsState,
+        waveform: torch.Tensor | None,
+    ) -> StagePayload:
+        if waveform is not None:
+            audio_np = waveform.detach().to(torch.float32).cpu().numpy()
+            payload.data["audio_data"] = audio_np.tolist()
+        else:
+            payload.data["audio_data"] = []
         payload.data["sample_rate"] = sample_rate
         payload.data["modality"] = "audio"
         if state.prompt_tokens or state.completion_tokens or state.engine_time_s:
@@ -356,8 +355,28 @@ def create_vocoder_executor(
             payload.data["usage"] = usage
         return payload
 
+    def _vocode(payload: StagePayload) -> StagePayload:
+        state, codes_TN = _prepare_vocoder_item(payload)
+        waveform = codec.decode(codes_TN) if codes_TN is not None else None
+        return _store_vocoder_result(payload, state, waveform)
+
+    def _vocode_batch(payloads: list[StagePayload]) -> list[StagePayload]:
+        items = [_prepare_vocoder_item(p) for p in payloads]
+        valid = [(i, codes) for i, (_, codes) in enumerate(items) if codes is not None]
+        waveforms: list[torch.Tensor | None] = [None] * len(items)
+        if valid:
+            indices, codes_list = zip(*valid)
+            wavs = codec.decode_batch(list(codes_list))
+            for idx, wav in zip(indices, wavs):
+                waveforms[idx] = wav
+        return [
+            _store_vocoder_result(p, state, waveforms[i])
+            for i, (p, (state, _)) in enumerate(zip(payloads, items))
+        ]
+
     return SimpleScheduler(
         _vocode,
+        batch_compute_fn=_vocode_batch,
         max_batch_size=max_batch_size,
         max_batch_wait_ms=max_batch_wait_ms,
     )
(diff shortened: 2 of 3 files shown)


yxs:
@SandyLuXY 
English 1088-sample table, latency 2.039s → 2.155s (+5.7%), QPS 7.81 → 7.18 (−8.1%), and WER 1.29% → 1.82% (+0.53pp) — all three regress. The Chinese row is essentially a tie at noise-floor scale (latency −0.6% / QPS +0.5% / WER −0.05pp).

My hypothesis is that two mechanisms stack:
1. at c=16 / max_batch_size=4 / ~8 AR steps/sec, the vocoder is already not the bottleneck (the AR + KV cache is), so batching saves sub-percent;
2. but the padding overhead introduced by batching (`F.pad` × 4 + `torch.stack` + running ConvTranspose over an extra `T_max - T_i` codebook-0 frames per item) exceeds that saving, netting a regression. English SeedTTS samples are on average longer and higher-variance than Chinese, which gives a higher padding ratio and explains why English regresses more visibly.

sglang_omni/models/higgs_tts/audio_codec.py
```diff
@@ -199,5 +199,29 @@ def decode(self, codes_TN: torch.Tensor) -> torch.Tensor:
...
+        """Batch-decode variable-length ``[T_i, N]`` tensors into ``[L_i]`` waveforms."""
+        if not codes_list:
+            return []
+        if len(codes_list) == 1:
+            return [self.decode(codes_list[0])]
+
+        lengths = [c.shape[0] for c in codes_list]
+        T_max = max(lengths)
+        padded = torch.stack(
+            [F.pad(c, (0, 0, 0, T_max - c.shape[0])) for c in codes_list]
```
yxs: `decode_batch` pads variable-length code sequences with token id 0, then decodes all positions without any mask. In this codec, id 0 is a real codebook entry, not a neutral pad symbol.
That means shorter requests can become batch-dependent (their waveform changes when batched with longer requests), so output is no longer equivalent to per-request `decode()`.
Please avoid cross-length padding with real token IDs: bucket by equal T for batched decode, and fallback to single decode for mixed lengths (or add explicit masked decoding support).

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -356,8 +355,28 @@ def _vocode(payload: StagePayload) -> StagePayload:
...
+    def _vocode_batch(payloads: list[StagePayload]) -> list[StagePayload]:
+        items = [_prepare_vocoder_item(p) for p in payloads]
+        valid = [(i, codes) for i, (_, codes) in enumerate(items) if codes is not None]
+        waveforms: list[torch.Tensor | None] = [None] * len(items)
+        if valid:
+            indices, codes_list = zip(*valid)
+            wavs = codec.decode_batch(list(codes_list))
+            for idx, wav in zip(indices, wavs):
+                waveforms[idx] = wav
+        return [
```
yxs, line 377:
```python
return [
    _store_vocoder_result(payload, state, wav)
    for payload, (state, _), wav in zip(payloads, items, waveforms)
]
```

SandyLuXY:
@yxs Thanks for catching the cross-length padding issue! Please find the latest benchmarking results below.

## Benchmark Results — Run 2 (bucketed decode, rebased main)

As response to the feedback, the implementation was changed from zero-padding to **bucketing by equal T**. Variable-length sequences are no longer padded with real token IDs — only same-length items are batched together, and mixed lengths fall back to per-bucket decode. This avoids the non-causal DAC decoder corruption issue entirely.

Full SeedTTS evaluation on **1x H200**, concurrency={1, 16, 32}, temperature=0.8, top-k=50.
Main branch rebased to `origin/main` (commit `24d4786`). Model: `boson-sglang/higgs-audio-v3-tts-4b-base`.

### English (200 samples, concurrency=1)

> 2 requests per branch hit the 2048-token generation cap (~10 s each). Excluded below so percentiles reflect typical-length requests.

| Metric | main | batch-vocoder |
|--------|------|---------------|
| Latency mean (s) | 0.635 | 0.684 |
| Latency median (s) | 0.610 | 0.619 |
| Latency p99 (s) | 1.127 | 1.776 |
| RTF | 0.169 | 0.182 |
| Throughput (QPS) | 1.575 | 1.463 |
| WER corpus | 1.33% | 1.33% |
| Speaker Similarity | 64.16 | 64.16 |

### English (200 samples, concurrency=16)

| Metric | main | batch-vocoder |
|--------|------|---------------|
| Latency mean (s) | 1.640 | 1.519 |
| Latency median (s) | 1.560 | 1.445 |
| Latency p99 (s) | 2.812 | 2.514 |
| RTF | 0.413 | 0.407 |
| Throughput (QPS) | 7.657 | 10.104 |
| WER corpus | 1.38% | 1.00% |
| Speaker Similarity | 64.47 | 64.42 |

### English (200 samples, concurrency=32)

| Metric | main | batch-vocoder |
|--------|------|---------------|
| Latency mean (s) | 3.013 | 2.771 |
| Latency median (s) | 2.859 | 2.761 |
| Latency p99 (s) | 4.696 | 4.402 |
| RTF | 0.787 | 0.768 |
| Throughput (QPS) | 8.318 | 10.826 |
| WER corpus | 1.23% | 1.23% |
| Speaker Similarity | 63.94 | 64.34 |

### Analysis

**Quality (WER, Speaker Similarity):** No regression. At c=1, output is bit-exact — identical WER, speaker similarity, and total output tokens (24740 = 24740). At c=16 and c=32, both metrics are within normal run-to-run variance.

**Speed (c=1):** The scheduler dispatches single items directly to the serial `_vocode` path — the code path is identical to main. The median (0.622 vs 0.613) is the fair comparison. p99 latency is higher for the batched decode due to outliers.

**Speed (c=16):** QPS jumps from 7.66 to 10.10 (**+32%**), latency drops 7%. The vocoder was the pipeline bottleneck at this concurrency; batching removes it.

**Speed (c=32):** QPS jumps from 8.32 to 10.83 (**+30%**), latency drops 8%. The AR engine starts to dominate latency, but throughput still benefits from faster vocoder queue drain.

yxs:
Can we go with at least one full benchmark?

Ccyest:
Looking at the throughput increase, I am wondering why we can see such a huge improvement on throughput? Doesn't look too reasonable to me, for two reason:
1. in our current solution # "Only same-length items are batched into a single forward pass." --> i am concerned that the throughput increase is only on benchmark, not real world use case.
2. according to the profiling, the entire vocder stage only takes 14% of the runtime, now it increases ~30% throughput, which is impossible.

Based on the above comment, I suggest to:
run full size N=1088 benchmark with batched vocoder, and compare to running N=1088 on current main branch (not benchmark branch)

Ccyest:
Actually, I took a second look at PR https://github.com/sgl-project/sglang-omni/pull/534
Looks like our existing QPS is: 9.104 on EN, wondering how we tested and got QPS 7.66 on main? (bs=16) 

> Looking at the throughput increase, I am wondering why we can see such a huge improvement on throughput? Doesn't look too reasonable to me, for two reason:
> 
> 1. in our current solution # "Only same-length items are batched into a single forward pass." --> i am concerned that the throughput increase is only on benchmark, not real world use case.
> 2. according to the profiling, the entire vocder stage only takes 14% of the runtime, now it increases ~30% throughput, which is impossible.
> 
> Based on the above comment, I suggest to: run full size N=1088 benchmark with batched vocoder, and compare to running N=1088 on current main branch (not benchmark branch)

sglang_omni/models/higgs_tts/audio_codec.py
```diff
@@ -199,5 +199,43 @@ def decode(self, codes_TN: torch.Tensor) -> torch.Tensor:
...
+        """Batch-decode variable-length ``[T_i, N]`` tensors into ``[L_i]`` waveforms.
+
+        Only same-length items are batched into a single forward pass.
+        The non-causal DAC decoder corrupts shorter items when padded to
+        a longer ``T_max``, so mixed-length inputs are decoded per-bucket.
+        """
+        if not codes_list:
+            return []
+        if len(codes_list) == 1:
+            return [self.decode(codes_list[0])]
```
Ccyest, line 213: > RFC, not 100% correct but worth considering: Further space for improvement: can we try it to see if any improvement:
> Exact-T match probability is too low even with integer T (uniform over ~900 values means most requests land in single-element buckets). 
> 
> 1. Pad with BOC_ID (1024) instead of 0. BOC is the codec's "begin / silence" token, in-distribution for the decoder, far less likely to produce audible artifacts than out-of-distribution 0.
> 2. Round T to a step (e.g. K =  32) → bucket by rounded T.
> 
> Any input will be appreciated @SandyLuXY @yxs 

Update: we can keep fixed T considering we are going to have stream vocoder.

Ccyest:
cross-validated main, throughput on main (before batched vocoder) is:
bs=16 8.5 QPS
bs=32 9.2 QPS 
on N=1088, EN test.

zhaochenyang20:
good job!

sglang_omni/models/higgs_tts/audio_codec.py
```diff
@@ -199,5 +199,43 @@ def decode(self, codes_TN: torch.Tensor) -> torch.Tensor:
...
+        for i, c in enumerate(codes_list):
+            buckets.setdefault(c.shape[0], []).append(i)
+
+        if len(buckets) == 1:
+            stacked = torch.stack(codes_list)
+            codes_BNT = stacked.transpose(1, 2).to(device=self.device, dtype=torch.long)
+            audio = self.model.decode(codes_BNT).audio_values
+            return [audio[i, 0].cpu() for i in range(len(codes_list))]
+
+        results: list[torch.Tensor | None] = [None] * len(codes_list)
```
Ccyest, line 219: The all-same-length fast path at lines (len(buckets) == 1) and the multi-element bucket path at lines 230-236 perform identical logic. The all-same-length case is just the degenerate case of the mixed-length path (one bucket with multiple elements), so the fast path can be removed and the mixed path handles it naturally. 
Fix would be just remove it.

SandyLuXY: Totally make sense, updated.

SandyLuXY:
Thanks @Ccyest @yxs for the thorough review! Sharing the latest benchmarking results on the full EN seedTTS data:

## Benchmark: Batched Vocoder (bucketed decode)

Full SeedTTS EN evaluation (1088 samples) on **1x NVIDIA H200**, temperature=0.8, top-k=50.
Base commit: `58d7e53` (main). Model: `boson-sglang/higgs-audio-v3-tts-4b-base`.

### Speed

| Metric | c=1 main | c=1 batch | c=16 main | c=16 batch | c=32 main | c=32 batch |
|--------|----------|-----------|-----------|------------|-----------|------------|
| Latency mean (s) | 0.665 | 0.641 | 1.886 | 1.749 | 2.961 | 3.102 |
| Latency median (s) | 0.599 | 0.585 | 1.678 | 1.636 | 2.842 | 3.073 |
| Latency p95 (s) | 0.887 | 0.860 | 2.634 | 2.319 | 3.678 | 3.991 |
| Latency p99 (s) | 1.105 | 1.043 | 3.444 | 2.769 | 4.189 | 4.367 |
| RTF mean | 0.148 | 0.144 | 0.432 | 0.412 | 0.719 | 0.774 |
| **Throughput (req/s)** | **1.503** | **1.559** | **8.447** | **9.106** | **10.007** | **10.220** |
| Output tokens (mean) | 123 | 121 | 127 | 120 | 121 | 113 |

### Quality

| Metric | c=1 main | c=1 batch | c=16 main | c=16 batch | c=32 main | c=32 batch |
|--------|----------|-----------|-----------|------------|-----------|------------|
| WER corpus (excl >50%) | 1.28% | 1.18% | 1.28% | 1.27% | 1.38% | 1.21% |
| >50% WER samples | 2 | 2 | 4 | 1 | 5 | 0 |

### Throughput delta

| Concurrency | main (req/s) | batch (req/s) | Delta |
|-------------|--------------|---------------|-------|
| 1 | 1.503 | 1.559 | **+3.7%** |
| 16 | 8.447 | 9.106 | **+7.8%** |
| 32 | 10.007 | 10.220 | **+2.1%** |

### Analysis

**Quality:** No regression. WER (excl >50%) is within noise across all concurrency levels (1.18–1.28% batch vs 1.28–1.38% main). The >50% WER outliers are stochastic TTS hallucinations unrelated to the vocoder — the batch branch actually has fewer. Speaker similarity is stable at ~66.4 across concurrency levels.

**Speed (c=1):** +3.7% throughput, latency improves across all percentiles. At c=1 the vocoder rarely batches, so this is expected to be ~neutral; the small gain is likely run-to-run variance.

**Speed (c=16):** +7.8% throughput (8.45 → 9.11 req/s), p99 latency drops 20% (3.44s → 2.77s). The vocoder batching benefit is clearly visible here — multiple requests land in the vocoder queue simultaneously and decode together.

**Speed (c=32):** +2.1% throughput (10.01 → 10.22 req/s). The throughput gain is smaller because the AR engine is the bottleneck at high concurrency, not the vocoder. Mean latency is slightly higher (+4.8%), but this is confounded by the output token difference (113 vs 121 mean) — the main run generated longer audio per request, so shorter audio in the batch run should have *lower* latency all else equal; the fact that it doesn't suggests this delta is within stochastic noise.

yxs:
looks good, lets update PR description.

sglang_omni/models/higgs_tts/audio_codec.py
```diff
@@ -199,5 +199,37 @@ def decode(self, codes_TN: torch.Tensor) -> torch.Tensor:
...
+        for indices in buckets.values():
+            if len(indices) == 1:
+                results[indices[0]] = self.decode(codes_list[indices[0]])
+            else:
+                stacked = torch.stack([codes_list[i] for i in indices])
+                codes_BNT = stacked.transpose(1, 2).to(
+                    device=self.device, dtype=torch.long
+                )
+                audio = self.model.decode(codes_BNT).audio_values
```
yxs: `audio[j, 0]` slices `[L]` on device, then `.cpu()` issues one `cudaMemcpyAsync` D2H — repeated B times in the loop, all on the default stream, so the driver serializes them. Each per-item `.cpu()` is an implicit sync point, so Python blocks on each transfer before launching the next. For Higgs the bucket-level result is just `[B, 1, L]` contiguous; one `audio.cpu()` followed by host-side indexing is strictly fewer launches and one synchronization point.

```python
audio = self.model.decode(codes_BNT).audio_values.cpu()  # one D2H for [B, 1, L]
for j, idx in enumerate(indices):
    results[idx] = audio[j, 0]
```

sglang_omni/models/higgs_tts/audio_codec.py
```diff
@@ -199,5 +199,37 @@ def decode(self, codes_TN: torch.Tensor) -> torch.Tensor:
...
+                for j, idx in enumerate(indices):
+                    results[idx] = audio[j, 0].cpu()
```
zhaochenyang20: Run `.detach().cpu()` once for each multi-item bucket, then assign CPU rows back to the corresponding requests. This preserves the `decode_batch` return shape without leaking codec layout details into the stage layer.

```python
audio_cpu = self.model.decode(codes_BNT).audio_values[:, 0].detach().cpu()
for j, idx in enumerate(indices):
    results[idx] = audio_cpu[j]
```

sglang_omni/models/higgs_tts/audio_codec.py
SandyLuXY: Good call! Updated

sglang_omni/models/higgs_tts/stages.py
zhaochenyang20, line 377: Validate the returned count before filling `waveforms`, and raise `RuntimeError` on mismatch. This is an internal invariant, so it should not be swallowed or turned into a fallback; letting `SimpleScheduler` emit errors for the batch is clearer.

```python
wavs = codec.decode_batch(list(codes_list))
if len(wavs) != len(valid):
    raise RuntimeError(
        f"Higgs vocoder decode_batch returned {len(wavs)} audios for {len(valid)} requests"
    )
```

sglang_omni/models/higgs_tts/audio_codec.py
```diff
@@ -199,5 +199,37 @@ def decode(self, codes_TN: torch.Tensor) -> torch.Tensor:
...
+
+        return results  # type: ignore[return-value]
```
zhaochenyang20: The `# type: ignore[return-value]` here turns `list[torch.Tensor | None]` into `list[torch.Tensor]`. The current bucket iteration should fill every slot, but that is the most important ordering and completeness invariant in `decode_batch`; once a future change touches bucket construction, empty-input filtering, or masked padding, a missed index will no longer be visible to static checking, and callers may fail much later after receiving `None`. This is a new codec API, so the invariant should be encoded explicitly instead of suppressed with a type ignore.

zhaochenyang20: Build the final `list[torch.Tensor]` explicitly and raise an internal error for any unfilled slot. That removes the `# type: ignore` while keeping the input-order and one-output-per-request constraints at the return boundary.

```python
final: list[torch.Tensor] = []
for i, result in enumerate(results):
    if result is None:
        raise RuntimeError(f"decode_batch did not produce audio for item {i}")
    final.append(result)
return final
```

tests/unit_test/higgs_tts/test_pipeline.py
```diff
@@ -220,3 +222,161 @@ def test_higgs_model_runner_skips_already_finished_eager_request() -> None:
...
+    codec = _make_fake_codec(call_log)
+
+    codes_a = torch.randint(0, 100, (10, 8))
+    codes_b = torch.randint(0, 100, (15, 8))
+
+    single_a = codec.decode(codes_a)
+    single_b = codec.decode(codes_b)
+    call_log.clear()
+
+    batch_results = codec.decode_batch([codes_a, codes_b])
```
zhaochenyang20, line 375: Make the bit-exact test include at least two same-length inputs and assert that `call_log` contains `(2, T)`. A separate mixed-length test can still verify singleton fallback, but the bit-exact assertion must cover the multi-item bucket.

```python
codes_a = torch.randint(0, 100, (10, 8))
codes_b = torch.randint(0, 100, (10, 8))
single_a = codec.decode(codes_a)
single_b = codec.decode(codes_b)
call_log.clear()

batch_results = codec.decode_batch([codes_a, codes_b])

assert call_log == [(2, 10)]
assert torch.equal(single_a, batch_results[0])
assert torch.equal(single_b, batch_results[1])
```

zhaochenyang20 (changes requested):

sglang_omni/models/higgs_tts/audio_codec.py
SandyLuXY: Resolved in https://github.com/sgl-project/sglang-omni/pull/574#discussion_r3314786687

zhaochenyang20 (approved):

Issue #565, opened 2026-05-25
[Perf] Enable torch.compile for Higgs TTS AR backbone


Ccyest:
## Motivation

CUDA graph capture for the AR decode loop is already landed and enabled by default (`disable_cuda_graph: False` in `create_sglang_tts_engine_executor`). The sampler state is fully tensorized (`HiggsBatchedSamplerState`) and the step function vectorized (`batched_step_direct` with `torch.where` masks), so the graph captures cleanly.

However, the Qwen3 backbone (4B) still runs fully eager kernels *inside* the captured graph — attention, MLP, RMSNorm, the fused multi-codebook embedding lookup, and the multi-head sampler step are all dispatched as individual CUDA kernels with no cross-operator fusion. torch.compile would fuse these, improving compute utilization and reducing kernel launch overhead within each captured graph replay.

We've proven this pattern on S2-Pro (~2× throughput from torch.compile + CUDA graph). Higgs TTS is structurally clean for this optimization — pure AR per frame, no flow-matching head, CUDA graph infrastructure already in place — making it a low-risk follow-up.

## Scope

One model, one stage:
- Model: `sglang_omni/models/higgs_tts/model.py` (`HiggsTTSModel`, Qwen3 backbone)
- Model runner: `sglang_omni/models/higgs_tts/model_runner.py` (`HiggsTTSModelRunner`)
- Stage factory: `sglang_omni/models/higgs_tts/stages.py` (`create_sglang_tts_engine_executor`)
- Config: `sglang_omni/models/higgs_tts/config.py` (`HiggsTtsPipelineConfig`)

What needs to happen:
1. Pass `enable_torch_compile: true` (and `torch_compile_max_bs`) through `server_args_overrides` in `create_sglang_tts_engine_executor`
2. Verify sglang's `ModelRunner` picks up the flag and applies `torch.compile` to the backbone forward on the decode path
3. Expose the toggle in `HiggsTtsPipelineConfig` default stages or via CLI override (currently the CLI only exposes `--thinker-torch-compile` / `--talker-torch-compile` for Qwen3-Omni; Higgs uses a `tts_engine` stage name)
4. Validate numerical parity and measure perf delta

## Baseline

Baseline should be measured with the current CUDA-graph-enabled configuration (`examples/configs/higgs_tts.yaml`, which defaults to `disable_cuda_graph: False`).

Measure on a single GPU (H100 or H200) using `benchmarks/eval/benchmark_tts_seedtts.py`:

[...]

PR #579, opened 2026-05-26
[Higgs] Enable torch.compile on AR backbone (decode-only fork)

Ccyest:
## Summary

Compile sglang's Qwen3 decoder layers via a thin fork (`HiggsQwen3Model`). Each layer is wrapped with `torch.compile(layer, mode="max-autotune-no-cudagraphs")` and stored on `model._compiled_decode_layers`. The fork's `forward` dispatches eager (prefill) vs compiled (decode).

CG capture is deferred until after compile so captured graphs record the compiled kernels.

**`_warmup_eager_pre_cg()`** — one eager forward at `bs=1` after `_compile_higgs_backbone` and before `init_device_graphs()` — forces lazy CUDA init (cuBLAS handle, allocator pool, …) to happen **outside** the CG capture window. Without it, the first dynamo trace inside sglang's capture loop pulls that lazy init *into* the captured graph and replay returns corrupted logits → audio runs to `max_new_tokens` → catastrophic outliers.

With the pre-warmup, the previous `bs ≤ 12` cap is no longer needed: `_compiled_max_decode_bs = 16` and `torch_compile_max_bs = 16`.

### c=16, max_running=16, cap=16

| Metric | Baseline | This PR | Δ |
|---|---:|---:|---:|
| Latency p99 | 3.14 s | 2.98 s | −5 % |
| QPS | 8.41 | 8.49 | +1 % |
| **WER excl > 50 %** | 1.29 % | 1.31 % | +0.02 pp (noise) |
| **Outliers (> 50 %)** | **50** | **2** | **−96 %** |

### c=32, max_running=32, cap=32

| Metric | Baseline | This PR | Δ |
|---|---:|---:|---:|
| Latency p99 | 5.38 s | 5.56 s | +3 % |
| QPS | 9.68 | 9.42 | −3 % |
| **WER excl > 50 %** | 1.28 % | **1.24 %** | **−3 %** rel |
| **Outliers (> 50 %)** | **50** | **4** | **−92 %** |
| WER per-sample max | 7 600 % | 2 793 % | −63 % |

Side-effect: the eager pre-warmup also cuts most of Higgs's inherent c ≥ 16 outliers (50 → ~3), so aggregate WER actually matches or beats baseline.

### Honest caveat

`torch.compile` itself doesn't deliver throughput on this workload (4 B decode-bound, bs ≤ 32). QPS lands within ±3 % of baseline. The PR's value is now:
1. compile is **safe** to enable across the full bs range, and
2. as a side effect, ~95 % of Higgs's inherent c ≥ 16 outliers go away.

## The bug, briefly

When `torch.compile` covers the first bs in sglang's capture loop, the resulting CG returns corrupted logits on replay (~6 % L2-norm drop in `hidden_states`, no NaN). Truth-tabled across 10 combinations — fires iff `compile=on AND CG=on AND that bs is the first one compiled inside the capture loop`. Six fixes attempted before V14 landed: `sync+empty_cache`, two-phase warmup, `dynamic=True`, skipping `set_torch_compile_config`, `mode=default`, ascending capture order — none rooted the bug cleanly. V13 (ascending order) was close but just moved the corruption from bs=largest to bs=1 (catastrophic at c=1).

Full investigation (truth table, hidden-states numerical signature, ruled-out hypotheses, all 7 fix attempts incl. V14) is in `issue_565_torch_compile_result.md` on this branch.

## Requested changes from sglang

V14 is a workaround in our wrapper, not a real fix. Things we'd want upstream sglang to do:

1. **Pin the actual root cause.** Why does the FIRST compile dynamo trace inside `CUDAGraphRunner.capture()` pull lazy CUDA state into the captured graph? Likely candidate: cuBLAS handle init / inductor scratch allocator first-touch happens during `run_once` warmup at iter 1, with `is_capture_mode` already true on the parent stream. A deeper inductor + CUDA-graph owner could nail this with the diagnostic data in `issue_565_torch_compile_result.md`.
2. **Force an eager warmup inside sglang's CG capture path** when `enable_torch_compile=True` (or per-layer compile is detected on the model). One internal `_dummy_run`-style call before the for-loop in `CUDAGraphRunner.capture()` would make the entire ecosystem immune to this class of bug — no model-specific patches needed downstream.
3. **Document the requirement.** Until (2) lands, the docstring on `init_device_graphs()` / `CUDAGraphRunner.capture()` should call out that the caller is responsible for pre-warming CUDA state if `torch.compile` is in play; otherwise the first captured bs is silently corrupted.

## Test plan

- [x] N=1088 SeedTTS-EN @ c=16 + c=32 — perf tables above
- [x] N=50 c=1 smoke test — matches baseline (lat 0.79 s, tokens 144, audio 5.48 s)
- [ ] `speaker_sim` evaluation
- [ ] Upstream sglang issue with the V14 evidence

### DIFF
--- a/issue_565_torch_compile_result.md
+++ b/issue_565_torch_compile_result.md
@@ -1,164 +1,144 @@
 # Issue #565 — Enable `torch.compile` for Higgs TTS AR backbone
 
 **Branch:** `feat/yichi/higgs-tts/torch_compile`
-**Status:** Production patch caps `_compiled_max_decode_bs=12` as the workaround. Root cause of the bs=16/32 catastrophic localized to an upstream sglang × inductor interaction. Multiple `sglang`-side fix candidates tried — one (V13: ascending capture order) eliminates the runaway and matches baseline quality, but **does not deliver throughput**. Holding the workaround pending an upstream owner.
+**Status:** Production patch ready. The previous `bs ≤ 12` cap is no longer needed — a one-shot eager forward before sglang's CG capture loop side-steps the upstream `sglang × inductor × CG` corruption at the first compiled+captured bs.
 
 ---
 
 ## TL;DR
 
-| Metric (single H200, SeedTTS EN) | Baseline (CG-on, no compile) | This PR (compile up to bs=12, eager bs≥13) |
-|---|---:|---:|
-| N=100, c=1   throughput (req/s) | 1.15 | **1.68**  (+46 %) |
-| N=100, c=1   RTF | 0.158 | 0.160 |
-| N=100, c=32  throughput | 5.96 | **9.22**  (+55 %) |
-| N=100, c=32  audio_s/s | 32.4 | **34.8** |
-| N=1088, c=16 WER excl > 50 % | 1.36 % (PR #534 report) | 1.32 % |
-| Catastrophic samples | 0 | 0 |
+`torch.compile`'s first dynamo trace during sglang's CUDA-graph capture loop allocates / initializes lazy CUDA state (cuBLAS handles, allocator pool, etc.) **inside the captured graph**. Replay of that graph then reads garbage at the slot the lazy init laid down → model emits broken logits → audio runs to `max_new_tokens` → catastrophic outliers.
 
-At c=16 we ship `_compiled_max_decode_bs=12`, so `bs==16` decode falls back to eager and dodges the upstream bug. Throughput at low concurrency is +46 %, but at c=16 with `max_running_requests=16` (and thus a fully compiled CG range up to bs=16) the bug fires; without the bs=12 cap the WER explodes.
+**The fix:** force all lazy CUDA init *outside* the capture window by running ONE eager forward at `bs=1` after `_compile_higgs_backbone` but before `init_device_graphs()`. ~12 lines in `stages.py`, no sglang patch required.
+
+Result vs no-compile baseline (SeedTTS-EN N=1088, single H200):
+
+| Config | Δ vs baseline |
+|---|---|
+| c=16 / mr=16 / cap=16 | outliers **50 → 2** (−96 %), WER excl 1.29 → 1.31 %, QPS 8.41 → 8.49 (+1 %) |
+| c=32 / mr=32 / cap=32 | outliers **50 → 4** (−92 %), WER excl 1.28 → 1.24 %, QPS 9.68 → 9.42 (−3 %) |
+
+Note: `torch.compile` itself doesn't deliver throughput on this workload (4 B decode-bound, bs ≤ 32). The PR's value is now (a) compile is **safe** to enable across the full bs range, and (b) the eager pre-warmup as a side-effect cuts most of Higgs's inherent c≥16 outliers from ~50/run to ~2–4/run.
 
 ---
 
 ## What the patch does
 
-Three files (one new), ~100 lines of code:
+Three files (one new), ~110 lines of code:
 
-1. **`sglang_omni/models/higgs_tts/sglang_qwen3_backbone.py`** (new) — `HiggsQwen3Model` subclasses sglang's `Qwen3Model` and overrides `forward` to pick between eager `self.layers` (prefill, bs > `_compiled_max_decode_bs`) and compiled `self._compiled_decode_layers` (decode, bs ≤ cap). `HiggsQwen3ForCausalLM` subclasses `Qwen3ForCausalLM` and swaps the inner `self.model`.
+1. **`sglang_omni/models/higgs_tts/sglang_qwen3_backbone.py`** (new) — `HiggsQwen3Model` subclasses sglang's `Qwen3Model` and overrides `forward` to pick between eager `self.layers` (prefill) and compiled `self._compiled_decode_layers` (decode). `HiggsQwen3ForCausalLM` subclasses `Qwen3ForCausalLM` and swaps the inner `self.model`.
 2. **`sglang_omni/models/higgs_tts/model.py`** — import the local fork; instantiate `HiggsQwen3ForCausalLM` instead of `Qwen3ForCausalLM`.
-3. **`sglang_omni/models/higgs_tts/stages.py`** — `_compile_higgs_backbone` populates the compiled-layers list and the bs cap. `create_sglang_tts_engine_executor` defers CG capture (`disable_cuda_graph=True` until after compile, then `init_device_graphs()`) so capture records compiled kernels rather than eager ones.
+3. **`sglang_omni/models/higgs_tts/stages.py`** —
+   - `_compile_higgs_backbone`: populate `_compiled_decode_layers`.
+   - `_warmup_eager_pre_cg` (NEW, ~12 lines): temporarily nulls `_compiled_decode_layers` (so dispatch falls back to eager `self.layers`), calls `model_runner._dummy_run(batch_size=1)`, restores the compiled list. Runs after `_compile_higgs_backbone` and before `init_device_graphs()`.
+   - `create_sglang_tts_engine_executor`: defers CG capture (`disable_cuda_graph=True` until after compile, then `init_device_graphs()`) so capture records compiled kernels.
 
 ---
 
-## Root-cause investigation — bs=16 catastrophic
+## Bug — what happens without the fix
 
-Initial attempt let bs=16 captured graphs run through compiled layers; that path produced 99–100/100 catastrophic outputs at c=32 (WER > 50 % on essentially every sample, audio runs to `max_new_tokens` instead of EOC).
+When `torch.compile` covers the largest bs in sglang's capture set, the CUDA graph for that bs returns corrupted logits on replay. Model never samples EOC → audio runs to `max_new_tokens=2048` (~80 s of audio per outlier). c=16 p99 latency goes from ~3 s (baseline) to ~19 s (no fix); c=32 p99 goes from ~5 s to ~27 s.
 
 ### Truth table (✓ = clean, ✗ = catastrophic 50–100/100)
 
-| # | bs | compile | CG capture | model | extra knob | Result |
+| # | bs | compile | CG | model | extra knob | Result |
 |:---:|:---:|:---:|:---:|:---:|:---|:---:|
-| 1 | 16 | off | off | Higgs | (eager baseline) | ✓ |
-| 2 | 16 | off | on  | Higgs | shipped path (CG-only at bs=16) | ✓ |
+| 1 | 16 | off | off | Higgs | eager baseline | ✓ |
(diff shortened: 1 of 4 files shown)


yxs:
**TL;DR:**

Across 5 N=1088 `compile+CG` runs (this PR), 4 N=1088 `eager+CG` baseline runs (current `main` HEAD b8da4fa), and 3 N=1088 `pre-warmup only` ablation runs (no compile), this PR does **not strictly improve QPS or RTF** over the current `main` baseline. Mean values trend slightly worse than baseline on all three measured axes (QPS, outlier rate, per-sample WER mean), though largely within run-to-run variance. The cause is mechanical: `torch.compile` and `CUDA Graph` attack the same bottleneck (per-kernel host launch overhead), and CG (default-on since #481) already eliminates it. Compile's only remaining theoretical edge — kernel selection — is empirically absent: inductor's autotune log on this branch picks the same cuBLAS kernel `eager` calls. The 50→2 outlier improvement reported here was measured against a pre-#503/#572 `main` snapshot; on current `main` the outlier baseline is already ~0–4 per 1088, so the marginal improvement is smaller and comes at a ~7% throughput cost (isolated via a no-compile pre-warmup-only ablation).

The acceptance criterion *"QPS strictly higher than CG-only baseline at same concurrency"* is **not satisfiable through `enable_torch_compile=True`** as currently formulated.

### 1. Data (all SeedTTS-EN, N=1088, c=16, single H200)

**Baseline (current `main` HEAD `b8da4fa`, eager + immediate-init CG)**: 4 runs

| Run | Throughput (req/s) | Latency mean (s) | Latency p99 (s) | RTF mean | Outliers (>50% WER) | WER per-sample mean |
| R1 | 8.616 | 1.949 | 3.280 | 0.4185 | 4 | 4.08% |
| R2 | 9.488 | 1.680 | 2.667 | 0.4025 | 0 | 1.27% |
| R3 | 8.804 | 1.811 | 3.223 | 0.4206 | 3 | 5.85% |
| R4 | 9.302 | 1.658 | 2.653 | 0.3969 | 1 | 1.36% |
| **mean** | **9.053** | 1.775 | — | 0.410 | **2.0** | **3.14%** |

**This PR (`compile + deferred-CG + pre-warmup`)**: 5 runs

| Run | Throughput (req/s) | Latency mean (s) | RTF mean | Outliers | WER per-sample mean | Notes |
| R1 | 5.051 | 3.153 | 0.7551 | 2 | 1.35% | inductor cold cache |
| R2 | 5.220 | 3.046 | 0.7197 | 2 | 1.54% | inductor cold cache |
| R3 | 8.947 | 1.729 | 0.4130 | 2 | 3.61% | warm |
| R4 | 8.626 | 1.843 | 0.4354 | 3 | 6.85% | warm |
| R5 | 8.361 | 1.906 | 0.4577 | 4 | 6.13% | warm |
| **mean (all 5)** | **7.241** | 2.336 | 0.5562 | 2.6 | 3.90% | |
| **mean (warm R3–R5)** | **8.645** | 1.826 | 0.4354 | **3.0** | **5.53%** | |

**Ablation: pre-warmup only (current `main` + `_warmup_eager_pre_cg`, no compile)**: 3 runs

| Run | Throughput (req/s) | Latency mean (s) | RTF mean | Outliers | WER per-sample mean |
| R1 | 8.280 | 1.869 | 0.4548 | 0 | 1.23% |
| R2 | 8.647 | 1.842 | 0.4341 | 1 | 3.67% |
| R3 | 8.306 | 1.919 | 0.4422 | 1 | 1.51% |
| **mean** | **8.411** | 1.877 | 0.4437 | **0.67** | **2.14%** |

**Side ablation: compile + no CG** (this PR + `disable_cuda_graph: True`, 1 run): **2.382 req/s** — ~4× slower than baseline. Confirms compile-alone cannot reach CG-only.

### 2. Bottom line vs current `main`

| Config | Throughput mean | Δ vs baseline | Outliers mean | Δ vs baseline | WER per-sample mean | Δ vs baseline |
| baseline (4 runs) | 9.053 | — | 2.0 | — | 3.14% | — |
| this PR (warm R3-R5) | 8.645 | **-4.5%** | 3.0 | **+50%** | 5.53% | **+2.4pp** |
| pre-warmup only (3 runs) | 8.411 | -7.1% | **0.67** | **-67%** | **2.14%** | **-1.0pp** |

- This PR (`compile + pre-warmup`) is **net-negative** on all three measured axes vs current `main` (means trend worse; ranges overlap due to run-to-run variance ~±10%).
- Pre-warmup-only is a **quality vs throughput trade-off**: ~7% throughput hit, ~3× outlier reduction, ~1pp better per-sample WER. The throughput hit is also present in this PR's warm runs (8.645) and appears traceable to the deferred-CG-init reordering itself (the eager `_dummy_run` call is a one-time boot cost, not per-step).
- No PR run, warm or cold, strictly exceeds any baseline run on throughput. The acceptance criterion *"QPS strictly higher"* is not met.

### 3. Why

`torch.compile`'s primary throughput contribution on decode-shape workloads is **reducing kernel launch overhead via fusion** — fewer kernels means less Python-side dispatch per step. `CUDA Graph` solves the same problem more thoroughly: the entire decoder forward is recorded once into a graph, and replay is a single CUDA driver call with zero Python participation. Empirical anchor: #481 measured CG-on vs CG-off at +69% QPS at c=32 — that headroom is exactly the launch-overhead band that CG eliminates and compile would otherwise target.

The two mechanisms attack the same bottleneck. Once CG is on (default in `main` since #481), compile's fusion benefit has nothing left to reduce; the only theoretical remaining lever is **kernel selection** — inductor producing a Triton kernel faster than the cuBLAS kernel eager would call.

For Higgs's decode MLP shape `mm(16x9728, 9728x2560)`, the inductor autotune log from this branch shows:

```
AUTOTUNE mm(16x9728, 9728x2560)
strides: [9728, 1], [1, 9728]
dtypes: torch.bfloat16, torch.bfloat16
  mm 0.0248 ms 100.0%
  triton_mm_38 0.0311 ms 79.7%  ACC_TYPE='tl.float32', ALLOW_TF32=False, BLOCK_K=128, BLOCK_M=16, BLOCK_N=32, ...
  triton_mm_42 0.0332 ms 74.8%
  triton_mm_46 0.0398 ms 62.3%
  triton_mm_50 0.0504 ms 49.2%
  triton_mm_37 0.0678 ms 36.6%
  triton_mm_36 0.0724 ms 34.2%
  triton_mm_41 0.0731 ms 33.9%
  triton_mm_35 0.0732 ms 33.9%
  triton_mm_45 0.0874 ms 28.4%
  [17 Triton variants total, all slower than cuBLAS]
SingleProcess AUTOTUNE benchmarking takes 0.2944 seconds for 18 choices
```
Inductor picks `mm` (cuBLAS) over all 17 Triton variants. The captured graph contains the same cuBLAS kernels eager would call. **Compile's kernel-selection edge is empirically empty for this shape.**

Combined: with CG default-on, compile's fusion benefit has nothing to reduce (CG already does it), and compile's kernel-selection benefit is rejected by its own autotune. Net result: compile adds per-batch autotune cost and occasional recompiles without offsetting throughput, observed as the cold-cache hit in R1/R2 and the slight gap below baseline in R3/R4/R5.

### 4. Cross-checks

- **S2-Pro reference cited in #565 motivation** ("~2× throughput from torch.compile + CUDA graph") is misattributed. The actual S2-Pro compile contribution is `+33% tok/s` per commit 4e5a1e6 (*"partial compile on the codebook decoder, +33% tok/s on S2-Pro TTS with fullgraph=True and zero graph breaks."*), applied to a structurally different code path: S2-Pro Fast AR, 4 layers × 9-step inner codebook loop, `dim ≈ 1024`. Higgs's 36-layer Qwen3-4B forward at `dim=2560, intermediate=9728` lives in a different shape regime where cuBLAS is already optimal (per the autotune log above). The mechanism that makes compile effective on S2-Pro (fusion of many tiny ops in a tight inner loop) does not apply to Higgs's structurally larger and shallower-loop decode.
- **vllm-omni** ships its Higgs path with `enforce_eager: true` (no compile) and `max_num_seqs: 4` at `vllm_omni/deploy/higgs_audio_v2.yaml`. Independent corroboration that compile isn't the throughput lever for this model. (They implement a different Higgs variant — Llama-3.2-3B based `HiggsAudioV2`, not the Qwen3-4B `HiggsMultimodalQwen3` used here. The architectural finding still transfers.)
- **Standalone full-layer microbench** (one Qwen3-style decoder layer at `bs=16`, `seq=1`, outside sglang/CG): compile shows `1.69×` (`mode=default`) and `1.89×` (`max-autotune-no-cudagraphs`) over eager — this is the launch-overhead reduction in a pure-Python execution context. The gain fully closes when CG is enabled, consistent with the mechanism above.

### 5. Why the 50→2 outlier figure has eroded

The PR description's 50→2 outlier delta measures the V14 fix's effect *on the compile-enabled path*: pre-V14, the compile+CG capture-loop bug (truth-tabled in `issue_565_torch_compile_result.md`) fires at the first compiled bs and produces ~50 catastrophic outliers; V14 sidesteps it, bringing the compile path's outlier count to ~2. This delta is correctly attributed to the V14 fix.

What it does *not* tell us is the comparison vs current `main` without compile. Measuring that separately:

- current `main` baseline (no compile): outliers mean 2.0 (range 0–4) across 4 runs
- pre-warmup only (current `main` + the `_warmup_eager_pre_cg` change, no compile): outliers mean **0.67** (range 0–1) across 3 runs
- this PR (`compile + pre-warmup`, warm R3–R5): outliers mean **3.0** (range 2–4) across 3 runs

So the pre-warmup mechanism in isolation does reduce outliers below current `main`. The PR's combined config (compile on top) does not — adding compile back brings outliers back up to slightly above current `main`. This reframes the trade-off: V14 is correctly necessary to *unbreak the compile path*, but turning compile on (even with V14) does not net-improve outlier rate vs the simpler current-`main`-plus-pre-warmup configuration.

### 6. What this means for #565

Two clean resolutions:

1. **Close #565** with the finding above. `enable_torch_compile` can be wired through as a config toggle (the code-health criterion is satisfiable in isolation), but turning it on does not deliver the perf criteria. If a future sglang version reduces CG's launch-overhead coverage, the analysis can be revisited.
2. **Rescope #565 to "improve Higgs decode RTF/QPS by any means"** (not compile-specific) and pursue host-side optimizations against the ~23 ms/step host-side gap (43% of wall, per torch.profiler trace on `main`). 

**On the pre-warmup mechanism specifically:** if outlier reduction is independently valuable, the `_warmup_eager_pre_cg` change can be considered as a small standalone PR — but the ~7% throughput cost vs current `main`, so I do not think it is a option for us.

cc: @Ccyest @zhaochenyang20

Ccyest:
close for finding out it is not improving throughput, and will be extra server launch overhead.

PR #560, opened 2026-05-25
feat(docs): add Higgs Audio v3 TTS cookbook

xinlij:
## Motivation

Adds a Higgs Audio v3 TTS Cookbook and fixes a packaging bug that caused `FileNotFoundError` at server startup when the package is installed via pip/uv.

## Modifications

- **`docs/cookbook/higgs_tts.md`** (new): Higgs Audio v3 TTS Cookbook covering model highlights, architecture overview with diagram, server launch, codec override, zero-shot TTS, voice cloning (curl + Python), streaming, inline control tokens (emotion/style/sfx/prosody with full token tables and audio demo samples), per-benchmark detailed WER/SIM results (Seed-TTS, CV3, MiniMax-Multilingual), and known limitations.
- **`docs/_static/image/`**: added `higgs-architecture.png` (architecture diagram), `logo.png`, and `logo.ico` to replace the external logo URL in `conf.py`.
- **`docs/_static/audio/`**: added `higgs-1.wav`, `control-tokens-test1.wav`, and `control-tokens-test2.wav` as inline audio demo samples embedded in the cookbook page.
- **`docs/conf.py`**: switched `html_logo` and `html_favicon` from external URLs to local `_static/image/` paths so the site logo renders correctly.
- **`docs/index.rst`**: added `higgs_tts` Cookbook entry to toctree.
- **`pyproject.toml`**: added `sglang_omni.models.higgs_tts/_vendored/*.json` to `package_data`; fixes `FileNotFoundError` for `higgs_audio_v2_tokenizer_config.json` at server startup when installed via pip/uv.

### navigation bar and table of contents:
image

### Here's the full page of Higgs cookbook:
image

Issue #488, opened 2026-05-19
[Docs] SGLang Omni docs overhaul: feature matrix coverage, router one-shot launch, RFC history consolidation


zhaochenyang20:
### Background

SGLang Omni's docs have accumulated meaningful gaps as the project iterates. With Qwen3 Omni in particular, the configuration space exploded along several orthogonal axes (thinker-only vs thinker-talker, FP8, colocated topology, input/output modal combinations), and it's hard for a user to land on the right launch command and request schema for their scenario from the current docs. Meanwhile a lot of early design discussion still lives in our internal Lark wiki and in PR comment threads, which raises the bar for new external contributors.

This issue breaks the work into four chunks that can be picked up independently.

---

### 1. Qwen3 Omni feature matrix doc (highest priority)

Qwen3 Omni's configuration space is a Cartesian product of several orthogonal dimensions. Today the information is scattered across the README, blog posts, and a handful of example scripts. We should add one cookbook-style page that flattens the matrix into **tables**, rather than describing each combination in prose.

Orthogonal dimensions:

- **Architecture mode**: thinker-only / thinker-talker
- **Precision**: BF16 / FP8
- **Topology**: disaggregated / colocated
- **Input modals**: text / audio / video / image, in any combination
- **Output modals**: text-only / text + audio

Proposed structure (modeled on the [DeepSeek-V4 cookbook](https://lmsysorg.mintlify.app/cookbook/autoregressive/DeepSeek/DeepSeek-V4#3-1-basic-configuration)):

- **Server launch table**: rows are configuration combinations (e.g. thinker-only + FP8 + colocated), columns are the full `python -m sglang_omni.launch_server ...` command, key flag explanations, and recommended hardware.
- **Request schema table**: rows are input/output modal combinations, columns are request body examples, supported sampling params, and known limitations.
- **Compatibility matrix**: which combinations are supported today, which are WIP, which are explicitly unsupported (so users stop stepping on landmines).

Coverage goal: any user scenario should be findable as a row in one of these tables, not via issue archaeology.

---

### 2. TTS docs incremental update

The TTS docs are relatively stable and don't need restructuring — mostly we need to fill in recent model and feature additions:

- New model coverage: Voxtral TTS, Higgs TTS, Qwen3 TTS
- Streaming I/O: client-side examples for streaming input and streaming output (current docs only show non-streaming)
- A small table of voice/prompt parameter differences across TTS models, if there are meaningful divergences

---

[...]

PR #585, opened 2026-05-26
docs(tts): add Voxtral TTS and Qwen3 TTS cookbooks (#488)

JiaxinD:
## Motivation

Part two (chunk 2) of #488 is a TTS docs task whose model-coverage list is *Voxtral TTS, Higgs TTS, Qwen3 TTS*. #560 delivered the Higgs piece as a per-model cookbook; this PR adds the two remaining cookbooks, mirroring that structure. The model code already exists and is registered, so this is docs-only.
image
image

## Modifications

- **`docs/cookbook/voxtral_tts.md`** (new): Voxtral-4B-TTS cookbook — prerequisites (incl. `mistral-common`), server launch, zero-shot, named voices (`voice`, with how to list the 20 presets), streaming, request parameters, known limitations. Voxtral uses preset named voices only and fixes `temperature=0.0` (no reference-clip cloning).
- **`docs/cookbook/qwen3_tts.md`** (new): Qwen3-TTS Base cookbook covering both 0.6B and 1.7B — prerequisites (`qwen-tts`), reference voice cloning (`references`), `language` hint, streaming, generation defaults, variants table.
- **`docs/index.rst`**: add a Cookbook toctree section.
- **`docs/basic_usage/tts.md`**: link the Voxtral / Qwen3-TTS rows to their cookbooks.

## Preview the docs locally

```bash
cd docs
uv pip install -r requirements.txt
python -m sphinx -b html . _build/html
# then open _build/html/cookbook/voxtral_tts.html and _build/html/cookbook/qwen3_tts.html
```

(`make html` works too.) The two cookbooks appear under the new **Cookbook** section in the sidebar.

## Bug fixed along the way

The Qwen3 prerequisite that the existing TTS docs carried used `uv pip install --upgrade transformers==4.57.3 ...`. The `--upgrade` flag drags the whole CUDA stack forward (torch 2.9.1 → 2.12, NumPy → 2.4, plus a parallel set of cu13 NVIDIA libs), which then breaks inference at request time:

- `RuntimeError: cuDNN Frontend error: [cudnn_frontend] Error: No valid execution plans built.`
- `numba` also refuses to import: `Numba needs NumPy 2.3 or less. Got NumPy 2.4.`

The new cookbook drops `--upgrade` and pins only the required packages, leaving the image's existing `torch` build untouched.

## Verification

Served both models on an H200 via `sgl-omni serve` and ran every documented curl example:

- **Voxtral**: zero-shot + named voice (`casual_male`) → HTTP 200, 24 kHz WAV; confirmed the 20 voice presets; `references` is silently ignored (named-voice only).
- **Qwen3-TTS 0.6B**: reference cloning + `language: Chinese` + streaming (SSE chunks + `[DONE]`) → HTTP 200, 24 kHz WAV.

> Note: #560 also introduces the `Cookbook` toctree section. If it merges first, rebase to keep `cookbook/higgs_tts.md` alongside the two new entries.


docs/cookbook/voxtral_tts.md
```diff
@@ -0,0 +1,142 @@
...
+> Voxtral generation is **deterministic**: the engine fixes `temperature` to `0.0`, so sampling
+> parameters such as `top_p`, `top_k`, and `temperature` are not used. Reference-clip voice
+> cloning (`references`) is **not** supported for Voxtral — use a preset `voice` instead.
+
+## Performance
+
+Per the [model card](https://huggingface.co/mistralai/Voxtral-4B-TTS-2603), Voxtral targets
+production voice agents: ~70 ms first-audio latency at concurrency 1 (RTF ≈ 0.10) and high
+throughput under load on a single GPU with ≥16 GB of memory. Output is 24 kHz.
+
```
0xjeffro, line 150:
```suggestion

## Benchmark Results

On seed-tts EN (full set, 1088 utterances), bf16, `max_new_tokens=4096`,
`--no-ref-audio --voice cheerful_female`, c=16, scored with HF Whisper-large-v3
for WER.

Hardware: 1 x H100 SXM

| metric | value |
|---|---|
| WER (corpus, micro-avg) | 1.15% |
| WER per-sample mean | 1.14% |
| WER per-sample median | 0.00% |
| WER per-sample std | 3.66% |
| WER per-sample p95 | 9.09% |
| WER per-sample max | 42.86% |
| >50% WER samples | 0 (0.0%) |
| Evaluated / Skipped | 1088/1088 / 0 |
| Concurrency | 16 |
| Completed requests | 1088 |
| Failed requests | 0 |
| Latency mean (s) | 3.321 |
| Latency median (s) | 3.252 |
| Latency p95 (s) | 5.003 |
| Latency p99 (s) | 6.17 |
| RTF mean | 0.5834 |
| RTF median | 0.5761 |
| Audio duration mean (s) | 5.711 |
| Output throughput (tok/s) | 341.5 |
| Output tokens (mean) | 71 |
| Output tokens (total) | 77675 |
| Prompt tokens (mean) | 150 |
| Prompt tokens (total) | 163577 |
| Throughput (req/s) | 4.784 |

```

Hi @JiaxinD , I had a parallel PR #586  doing the same thing, happy to converge on yours. 
One thing I can add from my side: I ran the SeedTTS benchmark for Voxtral on H100, so we can include a Benchmark Results section. The `higgs_tts.md` guide already has one, so adding it here keeps the model guides consistent in structure. If this is acceptable, I'll add Qwen3-TTS results the same way :)

JiaxinD: Reproduced this on an H200 and added a `## Benchmark Results` section in b3fc522 (SeedTTS EN full set, 1088 utts, c=16, `--no-ref-audio --voice cheerful_female`, WER via Whisper-large-v3). The numbers line up well with yours: WER 1.20% corpus micro-avg (vs your 1.15%), per-sample p95/max identical at 9.09%/42.86%, RTF mean 0.519 and 5.40 req/s (a little faster than the H100 run), 1088/0 completed.

I reworded the latency note while I was there: the model card's ~70 ms / low-RTF figure is a concurrency-1 number, so I labeled the table explicitly as the c=16 throughput run to avoid mixing the two.

Same approach works for Qwen3-TTS — I can reproduce a SeedTTS run and add a matching section. Closing #586 in favor of this one works for me.

JiaxinD: Added the Qwen3-TTS section too (0.6B-Base, same Seed-TTS EN setup at c=16, reference voice cloning) in 30e074c. Headline numbers: per-sample median WER 0.00%, p95 9.09%, RTF mean 1.51, 1088/0 completed.

One finding worth flagging: the **raw corpus micro-avg is 18.29%**, but that's entirely from 2/1088 utterances (0.2%) that ran away into a repetition loop and kept generating ~164 s of looping audio up to `max_new_tokens` (Whisper transcribes endless "Thank you very much…" / "a pile of a pile of…"). Excluding those two, corpus WER is 1.07%. I documented the runaway mode + mitigation (`repetition_penalty` / `max_new_tokens`) under Known Limitations rather than hide it behind the excl-outlier number. RTF > 1 is the 0.6B codec pipeline at c=16, not single-stream latency.

I ran 0.6B since it's the more failure-prone variant; happy to add a 1.7B row later if useful.

zhaochenyang20:
image

no need to mention this in our docs

zhaochenyang20:
image

delete this

zhaochenyang20:
image

[TTS usage guide](../basic_usage/tts.md) will be removed. So directly have all the content here as self-contained.

zhaochenyang20:
image

Is this the model intent to do?

zhaochenyang20:
image

The performance should be updated after #527 @GaokaiZhang and:

> Reproduce with the SeedTTS command in the [TTS usage guide](http://127.0.0.1:8040/basic_usage/tts.html). The Voxtral model card also quotes ~70 ms first-audio latency at concurrency 1; the table above is a throughput-oriented run at concurrency 16, so its RTF reflects batched load rather than the latency-optimized single-stream figure. Output is 24 kHz.


This should be self-contained.

zhaochenyang20:
image

running failed for this block. 

image

JiaxinD:
All addressed in 022c1c9: dropped the license and token notes, inlined the streaming consumer and reproduce command so each cookbook is self contained, and reworded the Voxtral note as this engine's behavior. Removed the Qwen3 zero shot example since the model is cloning only and 500s without a reference, and documented that requirement instead (also fixed the language example, its zh ref URL was 404). Benchmark numbers can refresh after #527.

0xjeffro:
Hi @JiaxinD , may be we should also update the table at the top of https://github.com/sgl-project/sglang-omni/blob/24d478692573c24b37f79be7f36cf940faacf103/benchmarks/eval/benchmark_tts_seedtts.py#L105-L110
which currently only has S2-Pro and Higgs rows with the ‘Source’ column referencing this PR

zhaochenyang20:
image

image

we should remove zero-shot part.

zhaochenyang20 (approved):

JiaxinD:
Added these to the reference tables in `benchmark_tts_seedtts.py` (Source #585): Voxtral EN, Qwen3-0.6B EN/ZH, and Qwen3-1.7B EN, in both the accuracy and speed tables. Qwen3-0.6B EN raw WER has 2 runaway outliers so I noted the outlier-excluded 1.07% next to the Higgs note. Voxtral is English only so there's no ZH row, and 1.7B ZH didn't finish cleanly on the shared cluster (co-tenant OOM) so I left it out for now.

Issue #580, opened 2026-05-26
[Qwen3-Omni] tracking board


yxs:
Under the current topology and design constraints, conduct relatively fair comparisons and prioritize optimizing the most important metrics to achieve leading performance.

## Current status

c=4, WER and latency are in a good position

### comprehensive profiling with broader metric coverage

- [x] #582 
- [x] #587

### Large Profiling

evaluate whether achieving leading performance across all levels is necessary and realistically achievable under the current architecture

### further profile and optimize additional metrics

**Optimize TTFT**
- [x] #475

**Bug fix**
- [ ] #524

### Encoder CUDA Graph

- [ ] #523

yijiangt:
Hi, I'd like to work on adding UTMOS evaluation to the benchmark pipeline. Will open a separate issue and submit a PR. References #580.

yxs:
## Profiling


<details>
<summary>May 28, 2026</summary>

**Date**: 2026-05-28 12:11, 5R
**Host**: 2×H200 engine (cuda:5,6) + 1 ASR GPU (cuda:7)
**Workload**: 1088 EN SeedTTS samples, streaming, c=4, audio-only modality
**Sampler (overridden via request payload, both sides)**: T=0.7, top_p=0.9, max_tokens=256
**ASR**: openai/whisper-large-v3

**lat −20.8%、QPS +26.3%、TTFT −56.6%**

| Metric | sglang-omni e3e9413a3899ab3db5ddc4dcd22baed8089904ab  | baseline | gap (sg vs baseline) |
|---|---:|---:|---|
| **lat_mean ± σ (s)** | **0.881 ± 0.015** | 1.113 ± 0.011 | -20.8% |
| lat p95 (s) | 1.289 | 1.558 | -17.3% |
| lat p99 (s) | 1.576 | 1.751 | -10.0% |
| **QPS mean ± σ** | **4.535 ± 0.073** | 3.591 ± 0.037 | +26.3% |
| audio throughput (s/s) | 16.313 | 15.249 | +7.0% |
| RTF mean ± σ | 0.2487 ± 0.0033 | 0.2682 ± 0.0023 | n/a (audio-length asymmetry) |
| RTF p95 | 0.3009 | 0.3189 | — |
| RTF p99 | 0.3426 | 0.3582 | — |
| **TTFC mean ± σ (s)** (audio_ttfp_mean) | **0.3471 ± 0.0063** | 0.7187 ± 0.0061 | -51.7% |
[...]

PR #587, opened 2026-05-26
[Benchmark] Add UTMOS speech quality evaluation metric

### LINKED ISSUE #584: [Benchmark] Add UTMOS speech quality evaluation metric
## Motivation

The SeedTTS benchmark currently reports WER, speaker similarity, and latency metrics, but does not measure perceptual audio quality. UTMOS (UTokyo-SaruLab MOS) is a neural MOS predictor that estimates human-perceived naturalness scores (1–5) for synthesized speech without requiring human annotators, providing a complementary quality signal.

## Proposed Modifications

- Add `benchmarks/metrics/utmos.py` with a `UTMOSScorer` class
- Add `benchmarks/metrics/utmos_assets.py` as a shared bootstrapper for auto-downloading model weights
- Add `run_seedtts_utmos()` in `benchmarks/tasks/tts.py`, following the same pattern as `run_seedtts_similarity()`
- Add `--with-utmos` flag to `benchmark_omni_seedtts.py` and `benchmark_tts_seedtts.py`
- Output `utmos_results.json` with per-sample scores and summary (mean, median, p5, p95)

## References

- Tracking issue: #580
- Pattern reference: speaker similarity implementation in #469

yijiangt:
## Motivation

The SeedTTS benchmark reports WER, speaker similarity, and latency metrics, but does not measure perceptual audio quality. UTMOS22 is a neural MOS predictor that estimates human-perceived naturalness scores (1–5) for synthesized speech without requiring human annotators, providing a complementary quality signal.

## Modifications

- Add `benchmarks/metrics/utmos.py` with a `UTMOSScorer` class and `ensure_utmos_assets()` bootstrapper. Model weights (`balacoon/utmos`) are downloaded from HuggingFace automatically; cache directory is controlled by `UTMOS_CACHE_DIR` env var (default `~/.cache/sglang-omni/utmos`). Cache integrity is verified via a JSON marker file recording `repo_id` and file size.
- Add `SeedttsUTMOSConfig` Protocol and `run_seedtts_utmos()` in `benchmarks/tasks/tts.py`, following the same pattern as `run_seedtts_similarity()`.
- Add `--utmos-only` flag to `benchmark_omni_seedtts.py` and `benchmark_tts_seedtts.py`.
- Output `utmos_results.json` with per-sample scores and summary (mean, median, p5, p95).

## Usage

```bash
# Score existing generated audio:
python -m benchmarks.eval.benchmark_omni_seedtts \
    --utmos-only \
    --output-dir results/qwen3_omni_en \
    --model qwen3-omni \
    --device cuda:0
```

## UTMOS Calibration

5 independent full generate+score runs on SeedTTS-50 EN, S2-Pro, H200 SXM:

| run | utmos_mean |
|-----|------------|
| 1   | 4.1638     |
| 2   | 4.1749     |
| 3   | 4.1549     |
| 4   | 4.1538     |
| 5   | 4.1617     |

worst-of-5 = 4.1538 · mean = 4.1618 · stdev = 0.0079 · threshold = apply_mos_slack(4.1538, slack=0.97) = 4.0292

5 independent full generate+score runs on SeedTTS-50 EN, Qwen3-Omni, H200 SXM:

| run | utmos_mean |
|-----|------------|
| 1   | 4.2346     |
| 2   | 4.2536     |
| 3   | 4.2866     |
| 4   | 4.1924     |
| 5   | 4.3202     |

worst-of-5 = 4.1924 · mean = 4.2575 · stdev = 0.0487 · threshold = apply_mos_slack(4.1924, slack=0.97) = 4.0667

## Related Issues

- Tracking issue: #580
- Closes #584
- Pattern reference: #469
- Model: [balacoon/utmos](https://huggingface.co/balacoon/utmos)

## Checklist

- [ ] Add unit tests.

## CI


benchmarks/metrics/utmos.py
```diff
@@ -0,0 +1,199 @@
...
+    Returns the path to the validated model file.
+    """
+    resolved = _resolve_utmos_cache_dir(cache_dir)
+    resolved.mkdir(parents=True, exist_ok=True)
+    logger.info("[utmos-assets] cache dir: %s", resolved)
+
+    model_path = resolved / _HF_FILENAME
+    marker = resolved / _MARKER_FILENAME
+
+    # ---------- cache HIT path ----------
```
zhaochenyang20: delete this useless AI generated comments.

benchmarks/metrics/utmos.py
```diff
@@ -0,0 +1,199 @@
...
+            logger.info("[utmos-assets] cache HIT at %s", resolved)
+            return model_path
+        logger.warning(
+            "[utmos-assets] cache invalidation: size=%s recorded=%s repo_id=%s",
+            model_path.stat().st_size if model_path.is_file() else "missing",
+            recorded_size,
+            recorded_repo_id,
+        )
+
+    # ---------- cache MISS path ----------
```
zhaochenyang20: delete all useless AI generated comments.

benchmarks/metrics/utmos.py
```diff
@@ -0,0 +1,199 @@
...
+        hit = (
+            model_path.is_file()
+            and model_path.stat().st_size >= _MIN_ASSET_SIZE_BYTES
+            and (recorded_size is None or model_path.stat().st_size == recorded_size)
+            and recorded_repo_id == _HF_REPO_ID
+        )
+        if hit:
+            logger.info("[utmos-assets] cache HIT at %s", resolved)
+            return model_path
+        logger.warning(
```
zhaochenyang20: use f-string overall.

benchmarks/metrics/utmos.py
```diff
@@ -0,0 +1,199 @@
...
+    if marker_data is not None:
+        entry = marker_data.get("files", {}).get(_HF_FILENAME, {})
+        recorded_size = entry.get("size") if isinstance(entry, dict) else None
+        recorded_repo_id = entry.get("repo_id") if isinstance(entry, dict) else None
+        hit = (
+            model_path.is_file()
+            and model_path.stat().st_size >= _MIN_ASSET_SIZE_BYTES
+            and (recorded_size is None or model_path.stat().st_size == recorded_size)
+            and recorded_repo_id == _HF_REPO_ID
+        )
```
zhaochenyang20: Make marker parsing a strict schema validation step: `files` must be a dict, `files[_HF_FILENAME]` must be a dict, `repo_id` must equal `_HF_REPO_ID`, `size` must be a positive integer, and the local file size must equal that integer. Any missing field should invalidate the cache; do not treat `recorded_size is None` as a hit. A small `_validate_asset(path, expected_size, expected_repo_id, recorded_repo_id)` helper would also align this code with the speaker-similarity asset bootstrapper.

benchmarks/metrics/utmos.py
```diff
@@ -0,0 +1,199 @@
...
+            return model_path
+        logger.warning(
+            "[utmos-assets] cache invalidation: size=%s recorded=%s repo_id=%s",
+            model_path.stat().st_size if model_path.is_file() else "missing",
+            recorded_size,
+            recorded_repo_id,
+        )
+
+    # ---------- cache MISS path ----------
+    logger.info("[utmos-assets] cache MISS at %s — fetching", resolved)
```
zhaochenyang20: f string

benchmarks/metrics/utmos.py
```diff
@@ -0,0 +1,199 @@
...
+    if marker.exists():
+        marker.unlink()
+    if model_path.exists():
+        ok = model_path.is_file() and model_path.stat().st_size >= _MIN_ASSET_SIZE_BYTES
+        if not ok:
```
zhaochenyang20: Wrap the miss/download/marker-write section in a lock such as `filelock.FileLock(resolved / ".utmos.lock")`; after acquiring the lock, re-run the hit validation so the second process does not redownload. The marker write should also use a process-unique tmp file, or keep the current fixed tmp path only while under that lock. This keeps the fast cache-hit path lock-free while protecting the slow path that mutates cache state.

benchmarks/tasks/tts.py
```diff
@@ -504,6 +506,132 @@ def run_seedtts_similarity(
...
+            scores.append(score)
+            scored_rows.append(
+                {
+                    "id": entry["sample_id"],
+                    "wav_path": entry["wav_path"],
+                    "utmos_score": round(score, 4),
+                    "is_success": True,
+                    "error": None,
+                }
+            )
```
zhaochenyang20: Include `sample_id = entry.get("sample_id")` in the `scoreable` predicate and require it to be a non-empty string; when it is missing, emit a skipped row with a reason such as `"sample_id missing from generated.json entry"`. If the project wants to support external audio manifests without IDs, then use `entry.get("sample_id")` and fall back to `wav_path` as the output ID, but that contract should be explicit in the same partitioning logic.

zhaochenyang20 (changes requested):
Submit the results with UTMOS. And, should we turn this evalution default on for TTS models and in the CI? @yxs

yijiangt:
Hi, all review comments addressed in the latest commit:

- Removed generated section comments and switched to f-strings overall.
- Added `_validate_asset()` with strict schema validation; removed `recorded_size is None` as hit
- Wrapped miss/download/marker-write in `FileLock` with double-checked hit validation post-lock
- `sample_id` validation moved into `scoreable` predicate; skipped rows fall back to `wav_path` as id

yijiangt:
> Submit the results with UTMOS. And, should we turn this evalution default on for TTS models and in the CI? @yxs

How many samples and runs do you need for UTMOS?

yxs:
@zhaochenyang20 Yes, but staged. like
 - land it as a metric first
 - collect N runs to characterize its σ
 - set a worst-of-N regression threshold, then flip it to a gate

zhaochenyang20:
> @zhaochenyang20 Yes, but staged. like
> 
> * land it as a metric first
> * collect N runs to characterize its σ
> * set a worst-of-N regression threshold, then flip it to a gate

@yijiangt Thanks for your contribution, and I think, yeah, we should have this in our CI. 

https://github.com/sgl-project/sglang-omni/blob/main/tests/test_model/test_s2pro_tts_ci.py

https://github.com/sgl-project/sglang-omni/blob/main/tests/test_model/test_qwen3_omni_tts_ci.py

You can try to modify these and run it locally 5 times and get the UTMOS result, and put all the results in your report, and set the worst of the 5 as the accuracy threshold.

> Reuse the existing speed-stage WAVs: add `_run_utmos(output_dir, device="cuda:0")` that invokes `python -m benchmarks.eval.benchmark_tts_seedtts --utmos-only ...`, add `_assert_utmos_results()` that checks non-empty `per_sample`, `skipped == 0`, presence of `utmos_mean`, and `utmos_mean >= threshold`. Then add a S2-Pro non-streaming UTMOS test next to `test_voice_cloning_similarity`, submit the five local run results, and set `VC_UTMOS_MEAN_MIN` from the worst-of-five value. If you want a soft landing first, at least add the structural CI check and persist the result artifact, with a constant comment explaining when the hard gate will be enabled.

benchmarks/tasks/tts.py
```diff
@@ -504,6 +506,137 @@ def run_seedtts_similarity(
...
+            "config": {
+                "model": config.model,
+                "device": device,
+            },
+            "per_sample": per_sample,
+        },
+        config.output_dir,
+        "utmos_results.json",
+    )
+    return {"summary": metrics, "per_sample": per_sample}
```
zhaochenyang20: Keep an output slot for every input row while iterating over `generated`, and store the original index together with each scoreable entry; after inference, write the scored result back to that original index. The final `per_sample` should be emitted in `generated` order, not as `scored_rows + skipped_rows`. If you want to keep the two-list structure, another option is to build an `id -> result` map and then emit by iterating over `generated`, but indexes are simpler and also handle duplicate or missing IDs in skipped rows.

zhaochenyang20 (dismissed):

zhaochenyang20:
有兴趣可以加我微信 LoveDeathAndLLM

benchmarks/tasks/tts.py
yijiangt: I see. Thanks for pointing out!

yijiangt:
Hi, all review comments addressed:
- Fixed `per_sample` ordering using index-based approach
- Added `_run_utmos` and `_assert_utmos_results` to both CI test files  
- Ran 5 calibration runs on SeedTTS-50 EN for S2-Pro and Qwen3-Omni on H200 SXM, results in PR description
- Set `VC_UTMOS_MEAN_MIN = 4.1535` (S2-Pro) and `VC_UTMOS_MEAN_MIN = 4.2876` (Qwen3-Omni)

tests/test_model/test_qwen3_omni_tts_ci.py
```diff
@@ -70,6 +71,7 @@
 # five with the standard slack margin. See the "Speaker similarity
 # calibration" section of the PR description for the per-run numbers.
 VC_SIMILARITY_MEAN_MIN = 60.0
+VC_UTMOS_MEAN_MIN = 4.2876
```
yxs: the PR did "run it 5 times", but those 5 runs were `--utmos-only` **re-scoring the same fixed audio** (that is exactly where `stdev=0.0000` comes from — 5 runs over the identical wavs are trivially identical). 

**So the 5 runs got the count right but the subject wrong.** Both thresholds are pinned to that fixed-audio mean with no downward margin, while the **CI test gates a different batch of audio**: `test_voice_cloning_utmos` scores audio generated *fresh in that run* (`benchmark_omni_seedtts` has no `--seed`, `temperature=0.8`, re-sampled every run). qwen3's `wer_audio_dir` fixture reuses `speed_artifacts.output_dir` (live `_run_benchmark` generation); s2pro's `wer_input_dirs` come from the stage-1 speed run's freshly generated `vc_nonstream_c{c}` dirs. TTS generation is non-deterministic (exactly why WER carries `apply_wer_slack` 1.25× and `VC_N_ABOVE_50_MAX=0` relies on run-to-run tolerance). Gating a quantity that fluctuates around 4.2876 with `>= 4.2876` is a coin flip: ~half of legitimate runs FAIL. s2pro is worse — it loops over `selected_s2pro_tts_concurrencies`, each concurrency must independently clear 4.1535, so the compound failure probability climbs further.

To fix it, you need re-calibrate over full generate+score runs (not --utmos-only on fixed audio), then apply a downward slack like the other metrics (e.g. apply_slack's ×0.875) instead of pinning to the observed mean.

something like

```python
# tests/utils.py — a higher-is-better mean-slack helper alongside apply_wer_slack
def apply_mos_slack(reference: float, slack: float = 0.97) -> float:
    """Derive a min-MOS threshold from a reference with downward slack."""
    return round(reference * slack, 4)

# value must come from worst-of-5 of FULL generate+score runs, not --utmos-only on fixed audio
VC_UTMOS_MEAN_MIN = apply_mos_slack(<worst_of_5_full_pipeline>)
```

yijiangt: Hi, the new commit contains re-calibrated UTMOS thresholds from full generate+score runs (not --utmos-only on fixed audio):

- Added `apply_mos_slack` helper to `tests/utils.py` (slack=0.97)
- S2-Pro: worst-of-5 = 4.1538, threshold = apply_mos_slack(4.1538) = 4.0292
- Qwen3-Omni: worst-of-5 = 4.1924, threshold = apply_mos_slack(4.1924) = 4.0667
- Updated PR description with full calibration tables

yxs (approved):

yxs:
@yijiangt If some stages fail, please fix the CI issue.

zhaochenyang20:
@yijiangt rebase please

yxs (approved):

PR #590, opened 2026-05-26
[Higgs] Async decode (one-step lookahead) for the omni AR loop

JiaxinD:
## Summary
Overlaps the per-step host collect with the next step's GPU forward in the omni
AR decode loop (single stream + CUDA `event.query()`, launch-first one-step
lookahead). Off by default (`--enable-async-decode` /
`SGLANG_OMNI_ENABLE_ASYNC_DECODE=1`). Rebased onto `main` after the Higgs
CG-batch-d2h base (#572) merged, so this is a clean async-decode-only diff.

## Design
- `ModelRunner.execute()` split into `execute_launch` (enqueue forward + on-GPU
  sample + async D2H of the collect snapshot into a pinned ping-pong buffer +
  record event) and `execute_resolve` (query event, host collect). The sync
  `execute()` is a pure extraction over the same sub-steps (byte-identical off).
- `OmniScheduler._event_loop_async_decode`: launch(N) then resolve(N-1), so
  step N-1's CPU collect overlaps step N's forward.
- **Fast path**: decode batches below `async_decode_min_batch_size` (default 2)
  bypass the lookahead and run the plain sync step — at bs=1 the collect is too
  small to overlap, so this avoids the lookahead's fixed overhead.
- New per-model hooks `post_decode_launch` / `post_decode_resolve`; CUDA event
  kept outside graph capture; `event.query()` with `synchronize()` fallback.

## Applicability — why Higgs first, what other models would get
This is shared infrastructure: the `execute_launch`/`execute_resolve` split and
the lookahead loop live in the base `ModelRunner` / `OmniScheduler`. But a model
only benefits if two conditions hold:

1. **It routes through `OmniScheduler`.** Plain LLMs (e.g. Llama-3-8B) are served
   by upstream sglang directly, which already has its own overlap scheduler — they
   never touch this path, so the flag is inert for them.
2. **It implements `post_decode_launch` / `post_decode_resolve`** — the hooks that
   take the per-step host collect off the critical path. The base default raises
   `NotImplementedError`; only async-safe models opt in (off by default).

**Where the win comes from, and why it's model-dependent.** The lookahead hides
step N−1's *host-side collect* behind step N's GPU forward. The recoverable time
is bounded by how much *serial CPU* that collect is relative to the forward, and
it grows with batch size (the collect scales with the number of requests). So the
benefit is large exactly when the per-step host collect is heavy — and ≈0 when it
isn't.

**Why Higgs is wired up first:** its decode is a multi-codebook AR loop whose
per-step collect is unusually heavy — per-request sampler-state scatter,
delay-pattern bookkeeping, EOC/finish handling, and a D2H of the codes snapshot,
in Python, every step. That is exactly the work this overlap hides, making Higgs
both the most natural beneficiary and a strict correctness exercise for the
machinery.

**What other models would get if enabled later:**
- Other multi-codebook TTS models (Qwen3-TTS, Voxtral-TTS, FishAudio) have a
  similarly heavy per-step collect → **similar bs>1 gains expected, scaling with
  concurrency** (direction expected; magnitude must be measured per model).
- Lighter, text-centric decode paths (e.g. the Qwen3-Omni thinker) have a smaller
  per-step collect → **proportionally smaller gains, possibly within noise** — the
  same reason Higgs is break-even at bs=1.
- In every case it's a **bs>1 throughput win**; bs=1 is break-even by design (fast
  path), and it ships off by default.

Bringing up a new model is just: implement the two hooks, then gate it behind the
bit-identical `verify_correctness` check and measure — no scheduler changes needed.

## Correctness
`output_codes` **bit-identical OFF vs ON, bs=1 and bs=4, 100/100 each** (greedy),
via `scripts/verify_correctness.py`. Also fixes the batched sampler to
short-circuit greedy → argmax so `temperature=0` is deterministic.

## Performance (Higgs TTS, greedy, olen=128)

Measured under controlled conditions (dedicated GPU, NUMA-pinned server, ON vs OFF on the byte-identical-off build; query_hit 100% throughout):

| Batch | Latency, OFF -> ON | Throughput, OFF -> ON | Output tok/s |
|---|---|---|---|
| bs=16 | 1228 -> 1113 ms (-9.3%) | 10.27 -> 10.77 req/s (+4.9%) | 1314 -> 1379 |
| bs=4 | 751 -> 717 ms (-4.5%) | 4.91 -> 5.23 req/s (+6.4%) | 629 -> 669 |
| bs=1 | break even (fast path -> sync) | break even | - |

The win grows with batch size, as expected (the overlapped per-step collect scales with the number of requests). An earlier uncontrolled run reported larger bs=16 deltas (about -17% latency / +12% throughput); it did not isolate host-CPU contention and is superseded by these.

**Full-set realistic workload** (SeedTTS-EN, N=1088, concurrency 16; 1088/1088 completed, 0 failures, clean run):

| Metric | OFF | ON | Δ |
|---|---:|---:|---:|
| Throughput | 8.29 req/s | 9.35 req/s | +12.7% |
| Latency mean | 1.893 s | 1.589 s | -16.1% |
| Latency p99 | 3.291 s | 2.638 s | -19.8% |
| RTF mean | 0.437 | 0.372 | -14.9% |
| RTF p99 | 0.972 | 0.592 | -39.1% |

The win is larger than the synthetic paired run above because variable-length requests with a heavy tail are exactly where overlapping the per-step host collect pays off most (tail RTF p99 drops about 40%).

- A decode-isolated nsys profile shows the residual per-step GPU idle is CPU
  scheduler/prepare work, not D2H — so a multi-stream follow-up would not add to
  this on Higgs. Decode is CPU-dispatch-bound (GPU ~1.3% busy); reducing the
  per-step CPU path is the larger lever and is left as separate follow-up.

## Test plan
- `tests/unit_test/.../test_async_decode.py` + batched-sampler determinism:
  41 unit tests pass.
- `scripts/verify_correctness.py` bit-identical gate (bs=1 & bs=4).

## Acknowledgments

Thanks to @JingwenGu0829 for the prior `_event_loop_overlap` exploration that 
confirmed the per-step CPU bubble can't be addressed at the scheduler-loop 
level alone — that finding shaped this PR's approach of restructuring 
`ModelRunner.execute()` itself.


sglang_omni/models/higgs_tts/model_runner.py
```diff
@@ -73,6 +116,30 @@ def _populate_cg_buffers(self, forward_batch, requests) -> None:
             rows_py, dtype=torch.long, device=model._cg_row_indices.device
         )
 
+        if (
+            self._async_enabled
+            and getattr(forward_batch, "_is_lookahead", False)
```
zhaochenyang20: `prepare_decode` already attaches `forward_batch.req_ids = ...`. But a cleaner alternative is to add an `is_lookahead: bool` field to the `_PendingStep` dataclass, or to pass it as a kwarg through `_prepare_and_forward`, rather than mutating an upstream framework class. If upstream `ForwardBatch` later becomes a frozen dataclass / `__slots__` class this breaks immediately. Not a hard request, but worth considering.

sglang_omni/model_runner/base.py
```diff
@@ -30,8 +57,169 @@ def __init__(self, tp_worker: Any, output_processor: Any):
...
+            batch_result,
+            forward_batch,
+            schedule_batch,
+            model_worker_batch,
+            scheduler_output,
+        )
+
+    # ------------------------------------------------------------------
+    # Async decode: launch / resolve split (one-step lookahead, decode only)
+    # ------------------------------------------------------------------
```
zhaochenyang20: delete these

sglang_omni/model_runner/base.py
```diff
@@ -30,8 +57,169 @@ def __init__(self, tp_worker: Any, output_processor: Any):
...
+            pending.model_worker_batch,
+            pending.scheduler_output,
+            set_output_ids=False,
+        )
+
+    # ------------------------------------------------------------------
+    # Shared execute sub-steps — one definition routed through by both the
+    # synchronous execute() and the async launch/resolve, so behavior can't
+    # drift between the two paths.
+    # ------------------------------------------------------------------
```
zhaochenyang20: delete these

sglang_omni/model_runner/base.py
```diff
@@ -133,9 +320,9 @@ def execute(self, scheduler_output: Any) -> ModelRunnerOutput:
...
                 scheduler_output.requests,
             )
-        schedule_batch.output_ids = batch_result.next_token_ids
+        if set_output_ids:
+            schedule_batch.output_ids = batch_result.next_token_ids
 
-        # Output extraction
         outputs = self.output_processor.process(batch_result, scheduler_output)
         for sched_req in scheduler_output.requests:
             data = sched_req.data
```
zhaochenyang20, line 339: In the async resolve path, `scheduler_output.requests` includes the overrun req (one that already finished in the prior step — `_resolve_and_process` drops it from `batch.reqs` but not from `scheduler_output.requests`). So its `data.generation_steps` is incremented by an extra 1, and `extra_model_outputs.update(extra)` also overwrites the real result with the overrun step's extras.

This has no observable side effect on current Higgs behavior (the finished req is not re-sampled and its output_codes append is skipped), but the reported step counter ends up off by one, and if `extra_model_outputs` later carries error-sensitive fields this will leak. Recommend `_finalize` also accept a `skip_rids` (or similar mask), kept in sync with `_emit_stream_output`.

hen have `execute_resolve` thread `skip_rids` through — but that set is currently computed by `OmniScheduler._resolve_and_process`, which is a different layer. The cleaner option is to have `execute_resolve` recompute it internally via `req.finished()` (mirroring `_decode_collect_host`), so it does not need to be passed as a parameter.

zhaochenyang20 (changes requested):

JingwenGu0829:
The BS=16 run

## Setup

- Main repo: `/data/sglang-omni`, `b8da4fa` (`origin/main`, latest fetched)
- PR repo: `/data/sglang-omni-590`, `e62e7de` (`pr-590`)
- Python/env: `/data/sglang-omni/.venv/bin/python`
- GPU: H200 GPU 0, verified empty before each server run with `nvidia-smi`
- Server: `sglang_omni.cli serve`, `examples/configs/higgs_tts.yaml`
- PR env: `SGLANG_OMNI_ENABLE_ASYNC_DECODE=1`, `SGLANG_OMNI_ASYNC_DECODE_MIN_BS=2`
- Benchmark: `benchmark_tts_seedtts --generate-only --ref-format references`
- Workload: 9 trials x 20 requests, concurrency/bs 16, `max_new_tokens=128`, `temperature=0`, warmup 0
- Result dirs:
  - Main: `/data/higgs_perf_bs16_main`
  - PR: `/data/higgs_perf_bs16_pr590`

## Result

| Metric | Main mean | PR mean | Delta, paired 95% CI |
|---|---:|---:|---:|
| Mean latency | 1.448 s | 1.198 s | -16.9% [11.9%, 21.8%] |
| Request throughput | 9.31 req/s | 10.40 req/s | +12.0% [6.8%, 17.1%] |
| Output token throughput | 991.8 tok/s | 1142.9 tok/s | +15.5% [10.8%, 20.2%] |
| Mean RTF | 0.370 | 0.295 | -19.8% [15.4%, 24.2%] |

sglang_omni/models/higgs_tts/model_runner.py
JiaxinD: Done in `0026f6f` — replaced the `forward_batch._is_lookahead` mutation with an `is_lookahead: bool = False` kwarg threaded through `prepare_decode` → `_prepare_and_forward` → `_populate_cg_buffers`, so the lookahead flag no longer mutates an upstream framework class. Thanks!

sglang_omni/model_runner/base.py
JiaxinD: Removed in `0026f6f`.

sglang_omni/model_runner/base.py
JiaxinD: Removed in `0026f6f`.

sglang_omni/model_runner/base.py
JiaxinD: Addressed in `0026f6f`: `execute_resolve` now recomputes `skip_rids` internally via `req.data.req.finished()` (mirroring `_decode_collect_host`) and threads it into `_finalize`, which skips those rids before `generation_steps += 1` and the `extra_model_outputs.update(extra)` overwrite — so no off-by-one step count and no extras leak from the overrun step. (Separately, found a fast-path overrun while testing — fix in `8e7563b`, see the comment below.)

JiaxinD:
🐛 Pushed `8e7563b` for a KV-cache double-free in the async-decode **fast path** (distinct from the resolve-path overrun above).

The fast path reads `batch` from `get_next_batch_to_run` at the top of the loop, then drains the in-flight lookahead step (`_resolve_pending_async`) before running `batch`. The drain can finish a req still present in that pre-drain `batch`; running it again re-frees its committed KV →
`process_batch_result_decode → release_kv_cache → cache_finished_req → pop_committed_kv_cache: AssertionError "Committed KV cache already freed"`.

Fix: after the drain, `filter_batch()` drops the just-finished reqs (fast-path analogue of the `_resolve_and_process` pre-finished drop). It's in the **shared** loop, so it also covers Higgs **length-finishes** — Higgs dodges most cases by marking EOC finishes early, but a length-finish at a bs→bs-1 transition hits the same double-free. Regression test `test_fast_path_does_not_double_free_req_finished_by_drain` drives the real `_event_loop_async_decode` (fails before / passes after); async-decode suite 13/13.

JiaxinD:
📋 FYI on extending this async-decode work to the **Qwen3-Omni talker** (the 2nd AR model). Sharing the attempt, the numbers, and a recommendation — TL;DR: **I don't think it's worth landing for the talker right now.**

**Attempt.** Wired `post_decode_launch`/`post_decode_resolve` for the talker (opt-in via `enable_async_decode`), profile-first, same gate as this PR.

**e2e A/B** (greedy, fixed olen, same card, OFF then ON sequentially, 7 trials, paired Student-t CI):

| bs | Metric | OFF | ON | Improvement, paired 95% CI |
|---|---|---:|---:|---:|
| 4 | Mean latency (ms) | 1163 | 1119 | +3.8% [−4.9%, +12.5%] — crosses 0 |
| 4 | Throughput (req/s) | 3.27 | 3.35 | +2.4% [−3.7%, +8.5%] — crosses 0 |
| 4 | Mean RTF | 0.355 | 0.339 | +4.6% [−4.3%, +13.6%] — crosses 0 |
| 16 | Mean latency (ms) | 2170 | 2185 | −0.7% [−2.6%, +1.2%] — crosses 0 |
| 16 | Throughput (req/s) | 6.45 | 6.38 | −1.1% [−3.2%, +0.9%] — crosses 0 |
| 16 | Mean RTF | 0.686 | 0.689 | −0.5% [−2.5%, +1.5%] — crosses 0 |

Every CI crosses 0 — no significant effect either way.

**Isolated talker AR-decode per-step *wall* A/B** (apples-to-apples with Higgs, which benchmarks its AR model directly — instrumented the talker_ar stage only, median per-step wall by batch size):

| bs | OFF (ms) | ON (ms) |
|---:|---:|---:|
| 2 | 19.56 | 19.53 |
| 4 | 19.48 | 17.26 |
| 8 | 21.29 | 22.00 |
| 16 | 23.01 | 23.05 |

Identical OFF vs ON across the whole bs 2–16 range (±2 ms noise, not consistently favoring ON). So even measuring *only the optimized AR loop* in isolation, lookahead moves nothing.

**Why no gain (the interesting part):**
1. **Only `post` is overlappable, and it's small.** Phase-0 per-step (isolated): `prepare` 2.25 + `forward` 2.57 + `post` 2.14 = **6.96 ms**. One-step lookahead overlaps `post` (~2.14 ms) behind the next forward; `prepare` runs *before* the forward and isn't hidden by this scheme.
2. **The real per-step wall is ~20 ms, not 6.96 ms.** In the colocated pipeline the talker shares the GPU with the thinker + code2wav, plus per-iteration loop/IPC overhead. So the ~2.14 ms overlappable slice is ~10% of the real step — below the measurement noise.
3. **The talker isn't the e2e bottleneck.** Qwen3-Omni e2e = thinker (autoregressive text, dominates) → talker → code2wav; the talker is a short, diluted tail. Even a perfect talker decode speedup barely moves e2e. Higgs is the opposite: a standalone AR TTS where the AR model *is* the whole workload, so the bubble dominates and the +12% @ bs=16 shows up.
4. **The Phase-0 gate was over-optimistic** — it compared the recoverable bubble against *summed components* (6.96 ms), not against the real per-step wall (which includes loop/IPC + contention). Lesson for future gates: compare the bubble to the real wall, not the component sum.

**Recommendation.** I'd suggest **not landing the talker async-decode wiring** for now — no measurable upside, and it adds an opt-in path to maintain. The one durable win from this exploration is the shared-loop fix (`8e7563b`) above, which is worth keeping regardless. Happy to revisit if the talker ever becomes the bottleneck (e.g. a talker-bound / talker-only serving mode), where the picture could change.

JiaxinD:
Final controlled numbers (async decode ON vs OFF, Higgs greedy, olen=128):

| Batch | Latency | Throughput | tok/s, OFF -> ON |
|---|---:|---:|---|
| bs=16 | -9.3% | +4.9% | 1314 -> 1379 |
| bs=4 | -4.5% | +6.4% | 629 -> 669 |
| bs=1 | break even | - | - |

Clean on every run (100% query-hit, no KV double free, no crash), so no regression from the fix. The win grows with batch size, and it only helps models that route through OmniScheduler and implement the launch/resolve hooks with a heavy per-step CPU collect (multi-codebook TTS); bs=1 is break even by design.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -896,6 +966,156 @@ def pop_and_process():
...
+                time.sleep(0.001)
+                continue
+
+            batch = self.get_next_batch_to_run()
+            self.cur_batch = batch
+
+            # Order matters for the bs=1 hot path: the batch-size check is the
+            # cheapest and most often decisive at low concurrency, so it
+            # short-circuits before the _batch_is_decode call (avoiding it
+            # entirely every bs=1 step).
```
zhaochenyang20: remove unsed comments. If needed, use `note (xingli): `

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -896,6 +966,156 @@ def pop_and_process():
...
+            )
+
+            if use_lookahead:
+                # launch current decode step, then resolve the previous one
+                sched_output, pending_step = self._run_batch_launch(batch)
+                prev_pending = self._async_pending
+                self._async_pending = (batch.copy(), sched_output, pending_step)
+                if prev_pending is not None:
+                    pb, ps, pstep = prev_pending
+                    self._resolve_and_process(pb, ps, pstep)
```
zhaochenyang20: synchronous decode and the existing overlap decode execute batches through `run_batch()`, and `run_batch()` at `sglang_omni/scheduling/omni_scheduler.py:601` catches batch execution exceptions, calls `_handle_batch_failure()`, and returns `_FAILED_BATCH_RESULT`. The async lookahead branch directly calls `_run_batch_launch(batch)` when `use_lookahead` is true, then directly calls `_resolve_and_process(pb, ps, pstep)` for the previous pending step; neither call is inside the same failure wrapper. I ran a minimal ad hoc reproducer on the current head: when `_run_batch_launch()` raises `RuntimeError("launch boom")`, `_event_loop_async_decode()` propagates the exception and `_handle_batch_failure()` is called 0 times; when `_run_batch_resolve()` raises `RuntimeError("resolve boom")`, `_resolve_and_process()` also propagates and `_handle_batch_failure()` is called 0 times. If the model runner, CUDA event, host collect, or stream emission raises in production on the async path, the request will not receive an error, `abort(..., defer_running_cleanup=False)` and KV cleanup will not run under the existing serving contract, and the scheduler thread can exit with unfinished requests left behind.

zhaochenyang20: Add safe wrappers for launch and resolve that are equivalent to `run_batch()`, or make `_event_loop_async_decode()` explicitly reuse `_handle_batch_failure()` at both call sites. The fix needs a clear pending-step policy: if the current launch fails, the previous pending step must not be left unreachable; if resolve fails, the pending batch's requests should go through the same error, abort, and resource-release path. This should not be broad exception swallowing; it should reuse the scheduler's already existing public serving-boundary failure contract.

sglang_omni/model_runner/base.py
```diff
@@ -30,8 +58,164 @@ def __init__(self, tp_worker: Any, output_processor: Any):
...
+            pending.event.synchronize()
+            self._async_query_miss += 1
+        skip_rids = {
+            req.request_id
+            for req in pending.scheduler_output.requests
+            if getattr(getattr(req, "data", None), "req", None) is not None
+            and req.data.req.finished()
+        }
+        self.post_decode_resolve(
+            pending.host_buf,
```
zhaochenyang20, line 196: Do not use `getattr` 

Use `req.data.req.finished()` directly in production code so schema drift surfaces early. The tests can make `_sched_output(n)` return realistic `SchedulerRequest`-shaped objects, or provide a minimal `data=SimpleNamespace(req=SimpleNamespace(finished=lambda: False))` fixture for tests that do not care about finished state. That preserves the overrun fix without swallowing internal contract errors.

sglang_omni/models/higgs_tts/stages.py
zhaochenyang20, line 1: where do we have this `--enable-async-decode` parameter?

zhaochenyang20: I think we should add a parameter, not ENV Args.

zhaochenyang20 (changes requested):

sglang_omni/scheduling/omni_scheduler.py
JiaxinD: Deleted in `c955c9a`.

sglang_omni/scheduling/omni_scheduler.py
JiaxinD: Addressed in `c955c9a` — `_event_loop_async_decode()` now reuses `_handle_batch_failure()` at both call sites (matches the `run_batch()` contract; no broad swallowing).

Pending-step policy:
- **Launch fail:** `try / except / else` — `_async_pending` is only rotated in the `else` branch, so `prev_pending` is left intact and the next iteration drains it normally. Failed `batch` goes through `_handle_batch_failure(batch, exc)`.
- **Resolve fail (lookahead branch):** the just-launched step is already stored in `_async_pending`; the failed prev batch `pb` goes through `_handle_batch_failure(pb, exc)`.
- **Drain (`_resolve_pending_async`, used by fast-path / prefill / pause / shutdown):** also wrapped — was the same exposure under the fast-path drain.

3 new tests mirror the reproducer (`launch boom` / `resolve boom` / `drain boom`):
- `test_async_path_launch_failure_calls_handle_batch_failure`
- `test_async_path_resolve_failure_calls_handle_batch_failure`
- `test_drain_resolve_failure_calls_handle_batch_failure`

Full suite: 16/16 in `tests/unit_test/pipeline/test_async_decode.py` pass on H200.

sglang_omni/model_runner/base.py
JiaxinD: Done in `c955c9a` — `execute_resolve` now calls `req.data.req.finished()` directly. The test helper `_sched_output(n)` now returns `SimpleNamespace`-shaped requests (`request_id=...`, `data=SimpleNamespace(req=SimpleNamespace(finished=lambda: False))`) so the schema-drift signal isn't swallowed.

sglang_omni/models/higgs_tts/stages.py
JiaxinD: Done in `c955c9a`:
- Added `--enable-async-decode` and `--async-decode-min-batch-size` typer flags in `sglang_omni/cli/serve.py` (mirrors `--enable-realtime`), with `apply_async_decode_cli_overrides` threading them into the `tts_engine` stage's `factory_args`.
- Promoted `enable_async_decode` and `async_decode_min_batch_size` to factory kwargs on `create_sglang_tts_engine_executor`.
- Dropped the `SGLANG_OMNI_ENABLE_ASYNC_DECODE` / `SGLANG_OMNI_ASYNC_DECODE_MIN_BS` env reads entirely.

So the canonical opt-in is now `--enable-async-decode` (or `enable_async_decode: true` in the stage `factory_args` of a config), and the env flags no longer exist.

JiaxinD:
SeedTTS-EN full set (N=1088, Higgs, concurrency 16, async decode ON vs OFF), 1088/1088 completed, 0 failures, clean run (no KV double free, no crash):

| Metric | OFF | ON | Δ |
|---|---:|---:|---:|
| Throughput | 8.29 req/s | 9.35 req/s | **+12.7%** |
| Output tok/s | 1030.6 | 1099.3 | +6.7% |
| Audio throughput | 38.9 s/s | 41.4 s/s | +6.3% |
| Latency mean | 1.893 s | 1.589 s | **-16.1%** |
| Latency p95 / p99 | 2.732 / 3.291 s | 2.242 / 2.638 s | -17.9% / -19.8% |
| RTF mean | 0.437 | 0.372 | -14.9% |
| RTF p95 / p99 | 0.730 / 0.972 | 0.488 / 0.592 | **-33.2% / -39.1%** |

The win is larger here than the synthetic paired benchmark above (uniform olen=128), because a real workload has variable-length requests and a heavy tail, which is exactly where overlapping the per-step host collect with the GPU forward pays off most. Tail RTF (p99) drops about 40%.

sglang_omni/cli/serve.py
```diff
@@ -624,6 +624,52 @@ def apply_cuda_graph_cli_overrides(
...
+        updates["async_decode_min_batch_size"] = int(async_decode_min_batch_size)
+    if not updates:
+        return pipeline_config
+    _apply_stage_factory_args_override(
+        pipeline_config,
+        stage_name="tts_engine",
+        updates=updates,
+        reason="async decode override",
+    )
+    return pipeline_config
```
zhaochenyang20, line 683: Validate the matched stage factory before mutating factory args. The direct fix is to allow only `sglang_omni.models.higgs_tts.stages.create_sglang_tts_engine_executor`, and otherwise raise `typer.BadParameter` with a clear message that async decode currently supports Higgs TTS only. A pipeline-config capability marker would also work, but model-specific parameters should not be injected by a generic stage name. Passing `--async-decode-min-batch-size` by itself needs the same support check because it also writes a parameter that other factories do not accept.

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

zhaochenyang20 (approved):

PR #592, opened 2026-05-27
[Playground] restructure into qwen-omni/s2pro/higgs and add Higgs TTS UI

JingwenGu0829:
- Rename playground/web → playground/qwen-omni and playground/tts → playground/s2pro so each subdirectory is named after its target model. Updates internal imports, start.sh comments/module paths, the realtime sub-README, and docs/basic_usage/tts.md.

- Add playground/higgs/ — a vanilla HTML/CSS/JS playground for Higgs Audio v3 TTS served by a small FastAPI proxy. Mirrors the s2pro layout (text + reference audio + sampling params on the left, output on the right). Single Synthesize button with a Streaming on/off toggle (default on); during streaming the hidden live <audio> autoplays the accumulating WAV and an animated equalizer indicator shows playback activity, with an SVG mute toggle. Final audio bar appears on completion.

- Inline control-token picker (per #560): clickable chips for emotion / style / sfx / prosody, each inserting the literal <|category:name|> at the cursor.

- Reference audio supports either file upload (forwarded as a local path via a tempdir on the playground host) or a public URL.

- Lab/instrument aesthetic: IBM Plex Sans + Mono, phosphor-mint accent on near-black, with a light-mode toggle persisted via localStorage and matched to prefers-color-scheme on first visit.

- Rewrite playground/README.md to cover only qwen-omni / s2pro / higgs, with both one-command (start.sh) and split-process (sgl-omni serve + UI) launch examples.


xinlij:
Please confirm whether the streaming functionality is working as intended. Testing indicates that while it does auto-play, users still have to wait for the voice audio to be fully generated before playback begins.

xinlij:
The UI is beautiful! We could add "sgl-omni" in a suitable spot—for instance, in the top-left corner.

xinlij:
Regarding voice cloning, please clarify which input takes precedence when a user uploads a file **and** provides a reference audio URL.

xinlij:
When the user is not playing audio, make this icon stop bouncing. Or, even more simply, remove the icon entirely.
image

xinlij:
For each generated audio, it would be even better if we could provide a download button for users (this would facilitate wider dissemination of the model's audio; however, it is not a problem if we don't provide it).

xinlij:
> For each generated audio, it would be even better if we could provide a download button for users (this would facilitate wider dissemination of the model's audio; however, it is not a problem if we don't provide it).

I’ve located the download button—it works fine as is! Alternatively, if time permits, we could remove the three-dot menu and place the "Adjust Playback Speed" and "Download" buttons directly in a spot where users can see them at a glance.

xinlij:
The "Download" and "Playback Speed" features appear in Chinese in my browser; I am unsure if this is due to my operating system. Please keep them in English.

xinlij (approved):
lgtm!

PR #597, opened 2026-05-27
Add real streaming for Higgs TTS

JingwenGu0829:
# Higgs TTS Streaming Benchmark (bs=16 full EN)

- Branch/commit: `higgs-real-streaming` @ `55914c2`
- Results root: `results/pr_stream_full_bs16_20260527`
- Model: `boson-sglang/higgs-audio-v3-tts-4b-base`
- Dataset: `zhaochenyang20/seed-tts-eval-arrow`, EN split, full run, 1088 samples
- Request shape: SeedTTS voice cloning, `--ref-format references`, request rate `inf`, `max_new_tokens=2048`
- `bs=16` means benchmark `--max-concurrency 16`
- Generation ran in parallel on two H100 80GB GPUs:
  - Non-streaming: `CUDA_VISIBLE_DEVICES=6`, server port `18106`
  - Streaming: `CUDA_VISIBLE_DEVICES=7`, server port `18107`
- WER/ASR ran after stopping both servers, also in parallel on GPUs 6 and 7.
- `Done/Fail` means completed generation requests / failed generation requests. WER metrics were evaluated on saved generated WAVs.
- `WER excl >50%` is corpus WER after excluding samples with per-sample WER > 50%.

## bs=16 Full EN

| Mode | Done/Fail | Lat mean (s) | Lat p95 (s) | Lat p99 (s) | RTF mean | RTF p95 | RTF p99 | Audio mean (s) | Audio p95 (s) | Audio p99 (s) | Audio max (s) | Audio thr (s/s) | Req/s | Out tok mean | Out tok p99/max | Out tok/s | WER corpus | WER excl >50% | WER p95/max | >50% WER samples | ASR lat mean/p95 (s) | ASR RTF mean |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| Non-streaming | 1088/0 | 1.782 | 2.629 | 3.228 | 0.4214 | 0.6758 | 0.9877 | 4.588 | 6.320 | 8.137 | 81.640 | 40.978 | 8.931 | 122.0 | 210.4 / 2048 | 1087.0 | 1.68% | 1.32% | 9.68% / 250.00% | 4 | 0.283 / 0.366 | 0.0684 |
| Streaming | 1088/0 | 2.291 | 2.966 | 17.896 | 0.4939 | 0.6566 | 0.8441 | 5.013 | 6.386 | 62.500 | 81.640 | 33.386 | 6.659 | 132.0 | 1569.5 / 2048 | 881.3 | 1.79% | 1.66% | 10.00% / 142.86% | 2 | 0.318 / 0.388 | 0.0727 |

## Notes

- Both generation runs completed all 1088 requests successfully.
- Full-dataset WER is similar between modes: non-streaming 1.68%, streaming 1.79%.
- Streaming is slower in this run: mean RTF 0.4939 vs 0.4214, and request throughput 6.659 req/s vs 8.931 req/s.
- The long-output tail is not streaming-only in this full run. Non-streaming had 5/1088 samples hit 2048 output tokens and 81.64s generated audio. Streaming had 11/1088 samples hit the same cap.
- Streaming has the larger long-tail impact: latency p99 17.896s vs 3.228s, audio p99 62.500s vs 8.137s, and output-token p99 1569.5 vs 210.4.
- Only one max-token/81.64s sample overlapped across both modes: `common_voice_en_167249-common_voice_en_167246`.
- Benchmark streaming-specific chunk timing fields (`audio_ttfp_s`, `text_ttft_s`, `inter_chunk_s`) were null for these saved results, so first-audio-chunk latency is not included in the table.

Max-token / 81.64s samples:

- Non-streaming: `common_voice_en_167249-common_voice_en_167246`, `common_voice_en_19770519-common_voice_en_19770517`, `common_voice_en_24009789-common_voice_en_24009790`, `common_voice_en_26554168-common_voice_en_26554170`, `common_voice_en_28638628-common_voice_en_28638717`
- Streaming: `common_voice_en_167249-common_voice_en_167246`, `common_voice_en_17271507-common_voice_en_17271509`, `common_voice_en_17352169-common_voice_en_17352173`, `common_voice_en_21659259-common_voice_en_21659260`, `common_voice_en_23589975-common_voice_en_23589976`, `common_voice_en_25035745-common_voice_en_25035744`, `common_voice_en_25035745-common_voice_en_25035747`, `common_voice_en_25693607-common_voice_en_25693609`, `common_voice_en_28003405-common_voice_en_28003408`, `common_voice_en_32144882-common_voice_en_32144885`, `common_voice_en_603760-common_voice_en_603759`

## Commands

Non-streaming generation:

```bash
CUDA_VISIBLE_DEVICES=6 HF_HUB_OFFLINE=1 .venv/bin/python -m benchmarks.eval.benchmark_tts_seedtts \
  --generate-only --base-url http://localhost:18106 \
  --model boson-sglang/higgs-audio-v3-tts-4b-base \
  --meta zhaochenyang20/seed-tts-eval-arrow --ref-format references \
  --max-concurrency 16 \
  --output-dir results/pr_stream_full_bs16_20260527/bs16_nonstream_full_en \
  --lang en --disable-tqdm
```

Streaming generation:

```bash
CUDA_VISIBLE_DEVICES=7 HF_HUB_OFFLINE=1 .venv/bin/python -m benchmarks.eval.benchmark_tts_seedtts \
  --generate-only --base-url http://localhost:18107 \
  --model boson-sglang/higgs-audio-v3-tts-4b-base \
  --meta zhaochenyang20/seed-tts-eval-arrow --ref-format references \
  --max-concurrency 16 \
  --output-dir results/pr_stream_full_bs16_20260527/bs16_stream_full_en \
  --lang en --stream --disable-tqdm
```

WER / ASR:

```bash
CUDA_VISIBLE_DEVICES=<6|7> HF_HUB_OFFLINE=1 .venv/bin/python -m benchmarks.eval.benchmark_tts_seedtts \
  --transcribe-only \
  --model boson-sglang/higgs-audio-v3-tts-4b-base \
  --meta zhaochenyang20/seed-tts-eval-arrow --ref-format references \
  --max-concurrency 16 \
  --output-dir <matching-output-dir> \
  --lang en --device cuda:0 --disable-tqdm [--stream]
```


yxs:
Pure IPC overhead + multiple `codec.decode` calls + per-chunk `.cpu().numpy().tolist()` could explain some of the mean throughput regression, but they **cannot explain the 7.5× p99 out-token gap and the near-disjoint max-cap-hit sample sets**. With the same model, same dataset, temperature=1.0 (random sampling), both runs in parallel on GPU 6 vs GPU 7, even accounting for sampling randomness the overall length distribution should be similar — p99 should not diverge from 210 to 1569. This points to either the streaming path or its host-side timing affecting generation (e.g. sampling RNG, sync timing), or the engine failing to emit EOC reliably for certain samples in streaming mode.

sglang_omni/models/higgs_tts/streaming_vocoder.py
```diff
@@ -0,0 +1,294 @@
...
+
+@dataclass
+class _HiggsStreamState:
+    delayed_rows: list[torch.Tensor] = field(default_factory=list)
+    emitted_raw_frames: int = 0
+    next_decode_rows: int = 0
+    has_emitted: bool = False
+
+
+class HiggsStreamingVocoderScheduler:
```
yxs, line 30: This is the third streaming vocoder scheduler, we can consider do the base-class extraction.
Not In this PR

sglang_omni/models/higgs_tts/streaming_vocoder.py
```diff
@@ -0,0 +1,294 @@
...
+    def _decode_state_to_audio(self, state: HiggsTtsState) -> torch.Tensor | None:
+        delayed_rows = state.output_codes_delayed
+        if not delayed_rows:
+            return None
+        rows = [torch.tensor(row, dtype=torch.long) for row in delayed_rows]
+        if len(rows) < int(state.num_codebooks):
+            return None
+        return self._decode_delayed_rows(rows, codebook_size=int(state.codebook_size))
+
+    def _decode_delayed_rows(
```
yxs, line 257: `_decode_delayed_rows(rows, *, codebook_size: int = 1026)` is called from `_decode_state_to_audio` with `codebook_size=int(state.codebook_size)` on the non-streaming path, but `_decode_delta` (streaming path) calls `_decode_delayed_rows(rows)` without passing it, so it silently uses the default 1026.

Higgs's `codebook_size` is always 1026 today, so this is not an active bug. But `codec_vocab = codebook_size - 2` determines the mask threshold `codes >= codec_vocab → 0` (silence), and the mask's correctness depends on `codebook_size` matching the engine's true vocab. If a future Higgs variant uses `codebook_size != 1026` (e.g. larger audio vocab), the streaming path will silently mask valid codes in `[1024, codec_vocab)` to silence (producing artificial silence) while the non-streaming path will be correct.

xinlij:
Please ensure that the content in the Higgs TTS Cookbook is updated synchronously.

sglang_omni/models/higgs_tts/streaming_vocoder.py
```diff
@@ -0,0 +1,294 @@
...
+            codes_TN >= codec_vocab, torch.zeros_like(codes_TN), codes_TN
+        )
+        return self._codec.decode(codes_TN).detach().to(torch.float32)
+
+    def _build_audio_payload(self, audio: torch.Tensor) -> dict[str, Any]:
+        return {
+            "audio_data": audio.cpu().numpy().tolist(),
+            "sample_rate": self._sample_rate,
+            "modality": "audio",
+        }
```
zhaochenyang20, line 278: every streaming vocoder audio chunk goes through `_build_audio_payload`, and the current implementation converts the waveform to CPU, then numpy, then a Python list.

Reuse the binary waveform shape that the client already supports: emit `audio_waveform` bytes, shape, dtype, and `sample_rate`, or add a shared audio payload helper so Higgs, Qwen3 code2wav, and S2-Pro streaming vocoders use the same compact format. The non-streaming path may still do one host serialization for the final HTTP WAV response, but streaming chunks should not build nested Python lists in the scheduler/vocoder hot path.

sglang_omni/models/higgs_tts/streaming_vocoder.py
```diff
@@ -0,0 +1,294 @@
...
+
+        window_start_raw = max(
+            0, state.emitted_raw_frames - self._stream_overlap_tokens
+        )
+        rows_end = emit_until_raw + num_codebooks - 1
+        rows = state.delayed_rows[window_start_raw:rows_end]
+        audio = self._decode_delayed_rows(rows)
+
+        decoded_raw_frames = emit_until_raw - window_start_raw
+        samples_per_frame = max(int(audio.shape[-1]) // max(decoded_raw_frames, 1), 1)
```
zhaochenyang20, line 210: Do not let `_decode_delayed_rows` carry an implicit model contract default. Store `codebook_size` in `_HiggsStreamState` and fill it when the payload arrives; if chunks may arrive before the payload and decoding must happen before then, have `HiggsTTSModelRunner` include `num_codebooks` and `codebook_size` in stream metadata and let the vocoder latch them from the first chunk. Either way, streaming and non-streaming should derive the mask threshold from the same state contract.

sglang_omni/models/higgs_tts/model_runner.py
```diff
@@ -258,5 +270,26 @@ def _mark_sampler_finished(req: Any, generation_done: bool) -> None:
...
+        # codes_N is already CPU here; clone breaks aliasing with the producer's staging buffer.
+        self._outbox.put(
+            OutgoingMessage(
+                request_id=sched_req.request_id,
+                type="stream",
+                target=self._vocoder_target,
+                data=codes_N.clone(),
+                metadata={"modality": "audio_codes"},
+            )
+        )
```
zhaochenyang20, line 292: Do not use `getattr`.

Read `sched_req.data.stage_payload` directly in `HiggsTTSModelRunner`, and expose a missing payload or malformed `params` as an internal contract error. On the vocoder side, explicitly accept `StreamItem` or a small typed helper instead of accepting a wrapper, raw tensor, and arbitrary list at the same time; tests can construct the real `StreamItem` or a minimal dataclass. This makes later pipeline contract drift fail in the first test or request instead of masquerading as a successful non-streaming response.

sglang_omni/models/higgs_tts/stages.py
zhaochenyang20, line 22: Update the top-level `stages.py` description so it names the streaming-aware vocoder scheduler, and fix the `_decode_delayed_rows` comment to state that `HiggsAudioCodec.decode` currently returns a CPU tensor and callers choose the compact serialization format. If this later changes to a device tensor, update it together with the payload helper format.

zhaochenyang20 (changes requested):

JingwenGu0829:
Closed for #614

PR #602, opened 2026-05-28
[Bugfix] Fix ming omni talker voice preset loading and add generation duration guard

edwingao28:
## Motivation

When `talker/data/voice_name.json` is missing from the model mount the server logged a warning and kept serving with an empty `voice_json_dict`. The default voice `"DB30"` then never resolved to a prompt wav, so CFM ran with `prompt_wav_lat=None` — no speaker anchor at all — and output drifted away from DB30 within seconds.

## Modifications

  - `MingTalkerExecutor` and `MingStreamingTalkerScheduler` now fail fast at load time when:
    - a default voice is configured but `voice_name.json` is missing → `FileNotFoundError`,
    - the default-voice key is not present in the manifest → `ValueError`,
    - any preset entry is missing `prompt_wav_path` → `ValueError` (not a bare `KeyError` from `dict[…]` lookup),
    - any referenced wav file does not exist on disk → `FileNotFoundError`.
  - `_validate_voice_presets` owns both the relative→absolute path resolution and the existence check, so malformed manifests surface with a clear message at the right layer.
  - `MingOmniTalker.omni_audio_generation` resolves speaker conditioning in this order:
    1. explicit `prompt_wav_path` wins (per-request override of preset),
    2. else `voice_name` is looked up in `voice_json_dict`,
    3. else raise. `voice_name=None` AND `prompt_wav_path=None` still raises.
    Resolution is hoisted out of the CUDA stream context so bad requests don't hold CUDA resources.
## Checklist

- [x] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/models/ming_omni/components/streaming_talker.py
+++ b/sglang_omni/models/ming_omni/components/streaming_talker.py
@@ -371,6 +371,35 @@ class MingStreamingTalkerScheduler:
             sr = getattr(owner, "sample_rate", None)
         return int(sr) if sr is not None else None
 
+    def _validate_voice_presets(
+        self, voice_dict: dict, manifest_path: str, talker_dir: str
+    ) -> None:
+        """Resolve relative prompt-wav paths and validate the manifest.
+
+        Mutates ``voice_dict`` in place so each entry's ``prompt_wav_path``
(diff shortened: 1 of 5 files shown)


yuan-luo (approved):
LGTM.

PR #604, opened 2026-05-28
[Perf] Optimize Qwen3-TTS reference prompt path

BBuf:
## Summary

This PR keeps the Qwen3-TTS 0.6B optimization focused on behavior-preserving serving paths:

- `SGLANG_OMNI_QWEN3_TTS_REF_PROMPT_CACHE=1`: cache repeated voice-clone reference prompt items.
- Keep CUDA graph enabled by default on top of the current torch.compile + CUDA graph bootstrap path.
- CUDA graph can still be explicitly disabled through `disable_cuda_graph=True` or `server_args_overrides={"disable_cuda_graph": True}`.

No greedy subtalker path is included. The original subtalker sampling behavior is preserved.

## Upstream readiness

Rebased onto current `sgl-project/sglang-omni:main` (`0a8da7a`, after #527) and preserved the upstream torch.compile/CUDA graph setup while adding the explicit opt-out knobs.

Checked locally:

```bash
git diff --check
python3 -m py_compile \
  sglang_omni/models/qwen3_tts/request_builders.py \
  sglang_omni/models/qwen3_tts/stages.py \
  tests/unit_test/qwen3_tts/test_pipeline.py
```

Both passed. Local pytest collection is blocked by the local `transformers` version missing `Qwen2_5_VLProcessor`, so the functional test result below is from the H200 validation environment.

## Correctness

Environment:
- Host/container: `ion-h200-8`, `sglang_omni_bbuf`
- GPU: H200, `CUDA_VISIBLE_DEVICES=1`
- Base commit: `e3e9413a3899ab3db5ddc4dcd22baed8089904ab`
- Runtime `PYTHONPATH`: worktree plus `/sgl-workspace/sglang/python`

Command:

```bash
CUDA_VISIBLE_DEVICES=1 \
PYTHONPATH=/tmp/sglang_omni_qwen_pr2_fix:/sgl-workspace/sglang/python \
SGLANG_OMNI_QWEN3_TTS_REF_PROMPT_CACHE=1 \
pytest -q tests/unit_test/qwen3_tts -q
```

Result:

```text
37 passed
```

## Performance

Benchmark workload:

```bash
CUDA_VISIBLE_DEVICES=1 \
PYTHONPATH=DIR:/sgl-workspace/sglang/python \
  --generate-only \
  --ref-format references \
  --max-samples 16 \
  --concurrency 16 \
  --warmup 1 \
  --max-new-tokens 2048 \
  --lang en \
  --disable-tqdm \
  --port PORT \
  --output-dir REPDIR
```

Median over 3 repetitions on GPU 1:

| metric | baseline | optimized | delta |
| --- | ---: | ---: | ---: |
| throughput_qps | 1.896 | 2.227 | +17.458% |
| latency_mean_s | 4.601 | 3.704 | -19.496% |
| latency_p95_s | 6.735 | 5.831 | -13.422% |
| rtf_mean | 1.0828 | 0.9183 | -15.192% |
| audio_throughput_s_per_s | 8.248 | 9.14 | +10.815% |

Artifact:
`/home/sglang-omni/bbuf/experiments/omni_h200_goal_20260527_1452/artifacts_latest/benchmarks_qwen_pr2_nogreedy_cgdefault_20260528_072124/summary.md`

### DIFF
--- a/sglang_omni/models/qwen3_tts/request_builders.py
+++ b/sglang_omni/models/qwen3_tts/request_builders.py
@@ -3,21 +3,22 @@
 
 from __future__ import annotations
 
-import copy
 import hashlib
 import os
 import threading
 import time
-from collections import OrderedDict
 from dataclasses import dataclass, field
+from pathlib import Path
 from typing import Any
 
 import torch
 
 from sglang_omni.models.qwen3_omni.pending_text_queue import PendingTextTensorQueue
 from sglang_omni.models.qwen3_tts.payload_types import Qwen3TTSState
+from sglang_omni.preprocessing.cache_key import hash_bytes
 from sglang_omni.proto import StagePayload
 from sglang_omni.scheduling.sglang_backend import SGLangARRequestData
+from sglang_omni.scheduling.stage_cache import StageOutputCache
 
 QWEN3_TTS_DEFAULT_MAX_NEW_TOKENS = 2048
 QWEN3_TTS_TASK_BASE = "Base"
@@ -125,8 +126,17 @@ class Qwen3TTSPreprocessingContext:
 _PREPROCESSING_CONTEXT: Qwen3TTSPreprocessingContext | None = None
 _PREPARED_REQUESTS: dict[str, Qwen3TTSPreparedRequest] = {}
 _PREPARED_REQUESTS_LOCK = threading.Lock()
-_REFERENCE_PROMPT_CACHE: OrderedDict[str, list[Any]] = OrderedDict()
+# Long-lived server state, so bound it by bytes (not just item count) and keep
+# the cached reference tensors on the host so all-miss / low-reuse traffic cannot
+# pin GPU memory without bound. Mirrors the Higgs ref-cache and Qwen3 Omni encoder
+# cache patterns.
 _REFERENCE_PROMPT_CACHE_MAX_SIZE = 256
+_REFERENCE_PROMPT_CACHE_MAX_BYTES = 512 * 1024 * 1024
+_REFERENCE_PROMPT_CACHE = StageOutputCache(
+    max_size=_REFERENCE_PROMPT_CACHE_MAX_SIZE,
+    max_bytes=_REFERENCE_PROMPT_CACHE_MAX_BYTES,
+    cache_device="cpu",
+)
 
 
 def set_qwen3_tts_preprocessing_context(*, model: Any, wrapper: Any) -> None:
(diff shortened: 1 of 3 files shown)


GaokaiZhang:
The direction and implementation looks correct and neat! I just have a glance and have some thoughts (I could be wrong).

### 1. Standalone `disable_cuda_graph` param is redundant / inconsistent with siblings
`qwen3_omni` and `fishaudio_s2_pro` flip this flag **solely** through `server_args_overrides={"disable_cuda_graph": ...}`. Here you add *both* a dedicated `disable_cuda_graph` param and the override. Since the override already covers it (and wins on conflict, because `.update()` runs last), consider dropping the standalone param to match the codebase pattern — or at least document the precedence.

### 2. Clone path is fragile to future schema changes
`_clone_cached_reference_prompt_item` reconstructs `VoiceClonePromptItem(**kwargs)` from a hardcoded 5-field set. If the upstream dataclass gains a field, the primary path **silently drops it** (only the deepest `copy.copy` fallback preserves it). Prefer `dataclasses.replace(item, ref_code=<clone>, ref_spk_embedding=<clone>)` (or `copy.copy(item)` + overwrite the two tensors) as the primary path so new fields are carried automatically.

BBuf:
Closing this PR.

Re-benchmarked the reference-prompt cache on H200 (Qwen3-TTS-12Hz-0.6B-Base, GPU isolated, 0 failures), comparing `SGLANG_OMNI_QWEN3_TTS_REF_PROMPT_CACHE` off vs on, median of 3 reps:

| workload | metric | cache off | cache on | delta |
|---|---|---:|---:|---:|
| full SeedTTS-EN (1088), conc=16 | throughput_qps | 0.690 | 0.695 | +0.7% |
| full SeedTTS-EN (1088), conc=16 | latency_mean_s | 22.371 | 22.146 | -1.0% |
| 16-sample, conc=16 | throughput_qps | 0.168 | 0.171 | +1.8% |

The speedup is within noise on both methodologies; the originally reported +17% (16-sample) does not reproduce. The cache does hit (~38.8% repeated refs in SeedTTS, content-addressed key), but this workload is AR-decode-bound (~22s/req for long audio), so `create_voice_clone_prompt` (tens of ms) is <1% of latency and caching it cannot move throughput.

The cache is correct and harmless (default-off), but it doesn't deliver a measurable serving gain here, so closing rather than merging. The CUDA-graph-default-on change is a separate optimization that can be pursued independently if desired.

PR #605, opened 2026-05-28
[Perf] Optimize Higgs TTS reference code caching

BBuf:
## Summary

This PR adds an opt-in Higgs TTS reference cache guarded by `SGLANG_OMNI_HIGGS_REF_CODE_CACHE=1`.

The cache avoids repeated work for repeated reference audio by:
- deriving stable cache keys for path, raw bytes, and base64/data audio inputs;
- including local path `size` and `mtime_ns` in the key so replacing a file at the same path does not hit stale cached data;
- caching the loaded/resampled reference waveform in preprocessing;
- caching delayed reference codec codes in the audio encoder stage;
- leaving default behavior unchanged when the env var is unset.

## Upstream readiness

Rebased onto current `sgl-project/sglang-omni:main` (`0a8da7a`). Checked locally:

```bash
git diff --check origin/main..HEAD
python3 -m py_compile \
  sglang_omni/models/higgs_tts/payload_types.py \
  sglang_omni/models/higgs_tts/stages.py \
  tests/unit_test/higgs_tts/test_pipeline.py
```

Both passed. Local pytest collection is blocked by the local macOS Python environment: `torchaudio` cannot load `libtorchaudio.so`, and the local `transformers` version is missing `Qwen2_5_VLProcessor`. The functional test result below is from the H200 validation environment.

## Correctness

Environment:
- Host/container: `ion-h200-8`, `sglang_omni_bbuf`
- GPU: H200, `CUDA_VISIBLE_DEVICES=1`
- Base commit: `e3e9413a3899ab3db5ddc4dcd22baed8089904ab`
- Runtime `PYTHONPATH`: worktree plus `/sgl-workspace/sglang/python`

Command:

```bash
CUDA_VISIBLE_DEVICES=1 \
PYTHONPATH=/home/sglang-omni/bbuf/experiments/omni_h200_goal_20260527_1452/sglang-omni-latest-e3e9413:/sgl-workspace/sglang/python \
SGLANG_OMNI_HIGGS_REF_CODE_CACHE=1 \
pytest -q tests/unit_test/higgs_tts -q
```

Result:

```text
10 passed, 2 warnings in 7.59s
```

Artifact:
`/home/sglang-omni/bbuf/experiments/omni_h200_goal_20260527_1452/artifacts_latest/correctness/higgs_after_waveform_cache_rerun_20260527_204134`

## Performance

Benchmark workload:

```bash
CUDA_VISIBLE_DEVICES=1 \
PYTHONPATH=/home/sglang-omni/bbuf/experiments/omni_h200_goal_20260527_1452/sglang-omni-latest-e3e9413:/sgl-workspace/sglang/python \
  --generate-only \
  --model boson-sglang/higgs-audio-v3-tts-4b-base \
  --ref-format references \
  --max-samples 16 \
  --concurrency 16 \
  --warmup 1 \
  --max-new-tokens 2048 \
  --lang en \
  --disable-tqdm \
  --port 30181 \
  --output-dir REPDIR
```

Median over 3 repetitions on GPU 1:

| metric | baseline | optimized | delta |
| --- | ---: | ---: | ---: |
| throughput_qps | 7.514 | 8.527 | +13.482% |
| latency_mean_s | 1.684 | 1.465 | -13.005% |
| latency_p95_s | 2.087 | 1.833 | -12.171% |
| rtf_mean | 0.4239 | 0.3657 | -13.730% |
| audio_throughput_s_per_s | 31.558 | 35.922 | +13.829% |

Artifact:
`/home/sglang-omni/bbuf/experiments/omni_h200_goal_20260527_1452/artifacts_latest/benchmarks_higgs_waveform_cache_20260527_204539/summary.md`

### DIFF
--- a/sglang_omni/models/higgs_tts/stages.py
+++ b/sglang_omni/models/higgs_tts/stages.py
@@ -22,8 +22,13 @@ Pipeline shape::
 
 from __future__ import annotations
 
+import base64
+import hashlib
 import logging
 import os
+import threading
+from collections import OrderedDict
+from pathlib import Path
 from typing import Any
 
 import torch
@@ -61,6 +66,102 @@ logger = logging.getLogger(__name__)
 # Codec runs at 75 Hz; chunked prefill of the multi-codebook prompt is unsafe
 # (sampler state machine has no rollback) so reject inputs past chunked_prefill_size.
 _MAX_REF_AUDIO_SEC = 100
+_REF_CODE_CACHE_MAX_ITEMS = 256
+_REF_CODE_CACHE_ENV = "SGLANG_OMNI_HIGGS_REF_CODE_CACHE"
+_REF_WAVEFORM_CACHE_MAX_ITEMS = 256
+
+
+def _ref_code_cache_enabled() -> bool:
+    return os.getenv(_REF_CODE_CACHE_ENV) == "1"
+
+
+def _reference_path_cache_key(path_like: str | Path) -> str | None:
+    path = Path(path_like).expanduser().resolve()
+    try:
+        stat = path.stat()
+    except OSError:
+        return None
+    return f"path:{path}:{stat.st_size}:{stat.st_mtime_ns}"
+
+
+def _reference_audio_cache_key(reference_audio: Any) -> str | None:
+    if isinstance(reference_audio, (str, Path)):
+        return _reference_path_cache_key(reference_audio)
+    if not isinstance(reference_audio, dict):
+        return None
+    path = reference_audio.get("audio_path") or reference_audio.get("path")
+    if path:
+        return _reference_path_cache_key(path)
+    if "bytes" in reference_audio:
+        data = reference_audio["bytes"]
+        if isinstance(data, str):
+            data = data.encode()
+        return "bytes:" + hashlib.blake2b(data, digest_size=16).hexdigest()
+    encoded = reference_audio.get("base64") or reference_audio.get("data")
+    if encoded is None:
+        return None
+    if isinstance(encoded, str):
+        raw = base64.b64decode(encoded)
+    else:
+        raw = bytes(encoded)
+    media_type = reference_audio.get("media_type", "audio/wav")
+    digest = hashlib.blake2b(raw, digest_size=16).hexdigest()
+    return f"base64:{media_type}:{digest}"
+
+
+def _waveform_cache_get(
+    cache: OrderedDict[str, torch.Tensor],
+    lock: threading.Lock,
+    key: str | None,
+) -> torch.Tensor | None:
+    if key is None:
+        return None
+    with lock:
+        cached = cache.get(key)
+        if cached is None:
+            return None
+        cache.move_to_end(key)
+        return cached.clone()
+
+
+def _waveform_cache_put(
+    cache: OrderedDict[str, torch.Tensor],
+    lock: threading.Lock,
+    key: str | None,
+    value: torch.Tensor,
+) -> None:
+    if key is None:
+        return
+    with lock:
+        cache[key] = value.detach().clone()
+        cache.move_to_end(key)
+        while len(cache) > _REF_WAVEFORM_CACHE_MAX_ITEMS:
+            cache.popitem(last=False)
+
+
+def _cache_get(
+    cache: OrderedDict[str, list[list[int]]], key: str | None
+) -> list[list[int]] | None:
+    if key is None:
+        return None
+    cached = cache.get(key)
+    if cached is None:
+        return None
+    cache.move_to_end(key)
+    return [list(row) for row in cached]
+
+
+def _cache_put(
+    cache: OrderedDict[str, list[list[int]]],
+    key: str | None,
+    value: list[list[int]],
+) -> None:
+    if key is None:
+        return
+    cache[key] = [list(row) for row in value]
+    cache.move_to_end(key)
+    while len(cache) > _REF_CODE_CACHE_MAX_ITEMS:
+        cache.popitem(last=False)
 
 
 def create_preprocessing_executor(
@@ -84,6 +185,8 @@ def create_preprocessing_executor(
     raw = Tokenizer.from_file(os.path.join(checkpoint_dir, "tokenizer.json"))
     tokenizer = PreTrainedTokenizerFast(tokenizer_object=raw)
     adapter = HiggsTokenizerAdapter(tokenizer)
+    reference_waveform_cache: OrderedDict[str, torch.Tensor] = OrderedDict()
+    reference_waveform_cache_lock = threading.Lock()
 
     def _preprocess(payload: StagePayload) -> StagePayload:
         inputs = payload.request.inputs or {}
@@ -117,17 +220,34 @@ def create_preprocessing_executor(
             )
 
         waveform_tensor = None
+        reference_cache_key = None
         if ref_codes_TN is None and inputs.get("reference_audio") is not None:
-            waveform_np, sample_rate = load_audio_to_24k(inputs["reference_audio"])
-            wav = torch.from_numpy(waveform_np)
-            if sample_rate != 24000:
-                wav = F_audio.resample(wav, sample_rate, 24000)
-            if wav.shape[-1] > _MAX_REF_AUDIO_SEC * 24000:
-                raise ValueError(
-                    f"reference_audio is too long "
-                    f"({wav.shape[-1] / 24000:.1f}s); cap at {_MAX_REF_AUDIO_SEC}s."
+            reference_audio = inputs["reference_audio"]
+            reference_cache_key = _reference_audio_cache_key(reference_audio)
+            if _ref_code_cache_enabled():
+                waveform_tensor = _waveform_cache_get(
+                    reference_waveform_cache,
+                    reference_waveform_cache_lock,
+                    reference_cache_key,
                 )
-            waveform_tensor = wav.view(1, 1, -1).contiguous().float()
+            if waveform_tensor is None:
+                waveform_np, sample_rate = load_audio_to_24k(reference_audio)
+                wav = torch.from_numpy(waveform_np)
+                if sample_rate != 24000:
+                    wav = F_audio.resample(wav, sample_rate, 24000)
+                if wav.shape[-1] > _MAX_REF_AUDIO_SEC * 24000:
+                    raise ValueError(
+                        f"reference_audio is too long "
+                        f"({wav.shape[-1] / 24000:.1f}s); cap at {_MAX_REF_AUDIO_SEC}s."
+                    )
+                waveform_tensor = wav.view(1, 1, -1).contiguous().float()
+                if _ref_code_cache_enabled():
+                    _waveform_cache_put(
+                        reference_waveform_cache,
+                        reference_waveform_cache_lock,
+                        reference_cache_key,
+                        waveform_tensor,
+                    )
 
         if ref_codes_TN is not None:
             delayed = apply_delay_pattern(ref_codes_TN)
@@ -156,6 +276,7 @@ def create_preprocessing_executor(
             prompt_token_ids=prompt_ids,
             reference_codes_delayed=ref_codes_delayed,
             reference_waveform=waveform_tensor,
+            reference_cache_key=reference_cache_key,
             target_text=target_text_for_encoder,
             reference_text=reference_text_for_encoder,
             num_codebooks=num_codebooks,
@@ -199,6 +320,7 @@ def create_audio_encoder_executor(
     codec.encode_reference(
         torch.zeros(codec.SAMPLE_RATE), sample_rate=codec.SAMPLE_RATE
     )
+    reference_code_cache: OrderedDict[str, list[list[int]]] = OrderedDict()
 
     def _encode(payload: StagePayload) -> StagePayload:
         state = HiggsTtsState.from_dict(payload.data)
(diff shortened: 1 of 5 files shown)


Hayden727 (approved):

Ccyest:
## Benchmark: SeedTTS EN full set (1088), rebased on `main`

Re-ran on this PR **rebased onto `upstream/main`** (tip `18a33ce27`, includes #590 async-decode) so the cache effect is isolated from the build delta. Clean venv (`uv pip install -e .`; sglang 0.5.8 / torch 2.9.1 / flashinfer 0.6.1), single uncontended H200, bs=16, full 1088-sample set, **3 repetitions per setting** (cache OFF vs `SGLANG_OMNI_HIGGS_REF_CODE_CACHE=1`).

### Workload

```bash
# server (cache ON adds SGLANG_OMNI_HIGGS_REF_CODE_CACHE=1)
CUDA_VISIBLE_DEVICES=7 [SGLANG_OMNI_HIGGS_REF_CODE_CACHE=1] \
sgl-omni serve \
  --model-path boson-sglang/higgs-audio-v3-tts-4b-base \
  --config examples/configs/higgs_tts.yaml \
  --host 0.0.0.0 --port 8007

# benchmark
PYTHONPATH=$(pwd) python -m benchmarks.eval.benchmark_tts_seedtts \
  --generate-only \
  --model boson-sglang/higgs-audio-v3-tts-4b-base \
  --ref-format references \
  --concurrency 16 \
  --warmup 1 \
  --max-new-tokens 2048 \
  --lang en \
  --port 8007 \
  --output-dir REPDIR
```

### Median over 3 repetitions (full 1088 set), GPU 7

| metric | baseline (cache off) | optimized (cache on) | delta |
|---|---|---|---|
| throughput_qps | 8.946 | 10.052 | **+12.363%** |
| latency_mean_s | 1.689 | 1.460 | −13.558% |
| latency_p95_s | 2.150 | 2.010 | −6.512% |
| rtf_mean | 0.3718 | 0.3478 | −6.455% |
| audio_throughput_s_per_s | 44.055 | 45.594 | +3.493% |

All 6 runs completed 1088/1088, 0 failures. Per-run throughput: cache-off `{8.946, 8.870, 8.952}` (very tight, std 0.04); cache-on `{10.052, 10.018, 10.943}`. Conservative floor (on-min vs off-max) is still **+11.9%**; the gain is never negative.

> Note: SeedTTS EN has 666 unique ref audios over 1088 samples (≈38.8% repeat), so cache hit rate here is a property of the eval set. Production voice-clone traffic with unique refs per request will see less; on all-miss traffic the cache is net overhead (key compute + clone).

### Why the gain (~12%) is larger than the encoder compute it removes

Per-stage profiling (in-tree request profiler, `SGLANG_TORCH_PROFILER_DIR` + `/start_request_profile`) at **concurrency=1** (no cross-request queue → residence ≈ pure compute):

| stage | pure compute (conc=1) | residence (conc=16) |
|---|---:|---:|
| tts_engine (AR decode) | 565 ms | 869 ms |
| audio_encoder (codec) | 55 ms | 686 ms |
| vocoder | 42 ms | 261 ms |
| preprocessing | 1 ms | 3 ms |

- The encoder's **pure compute is ~55 ms** (matches the original profiling); the AR decode at ~565 ms is the real compute bottleneck. The large `audio_encoder` residence at bs=16 is queueing, not compute.
- The cache removes only `hit_rate × enc_compute ≈ 0.388 × 55 ms ≈ 21 ms/req`, yet mean latency drops ~216 ms → **~10× amplification**.
- Mechanism: the 4 stages run in one process / one CUDA context on one GPU, and the encoder is a serial stage whose codec GPU work + GIL-held CPU work contends with the continuously-running bs=16 AR decode. Removing 38.8% of encodes relieves that contention — turning the cache on (encoder-only code) also dropped **tts_engine** conc=16 residence from **997 → 869 ms**, which is only explainable by reduced encoder↔decode contention on the shared GPU.

Ccyest:
## Re: fixed-seed request

Added a `--seed` flag to `benchmark_tts_seedtts.py` (flows through `gen_kwargs` → payload `seed` → server `sampling_seed`) and re-ran cache off vs on with `--seed 42`, full 1088 set, single uncontended H200 (GPU 0), bs=16.

### Fixed seed = 42 (single run each)

| metric | baseline (off) | optimized (on) | delta |
|---|---|---|---|
| throughput_qps | 8.652 | 9.309 | **+7.59%** |
| latency_mean_s | 1.783 | 1.650 | −7.46% |
| latency_p95_s | 2.561 | 2.344 | −8.47% |
| rtf_mean | 0.4308 | 0.3795 | −11.91% |

1088/1088 completed, 0 failures both sides.

### ⚠️ Important caveat: fixing the seed does **not** make this pipeline deterministic

I checked per-request completion-token counts between the two seeded runs: **only 52 / 1088 requests (4.8%) generated the same number of tokens**; mean |Δtok| = 33, max = 1961. Total generated tokens still differed by ~8% (off 129,191 vs on 139,343).

Reason: at bs=16 with continuous batching, the seed only fixes the sampling draw — it does **not** fix batch composition. Each request lands in a different batch shape run-to-run, which changes the floating-point reduction order in the GEMMs, perturbs the logits, and flips argmax/sampling → sequences diverge. So a fixed seed reduces but cannot remove generation variance here.

Practical consequence: this particular seeded pair happens to have cache-ON generating **more** tokens (128 vs 119 per request), i.e. ON did *more* work and was still faster — so the +7.6% here is on the conservative side. The cleanest apples-to-apples measure remains **median over N repetitions**:

> n=3 median (reported above): **+12.4% throughput**, −13.6% latency.

Net: the speedup is robust across both methodologies (seeded single-run +7.6%, n=3 median +12.4%, never negative). Recommend keeping the n=3 median as the headline since fixed-seed determinism isn't achievable in this batched-serving setup.

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -61,6 +66,34 @@
...
+        return hash_media_item(path)
+    if "bytes" in reference_audio:
+        data = reference_audio["bytes"]
+        if isinstance(data, str):
+            data = data.encode()
+        return hash_media_item(data)
+    encoded = reference_audio.get("base64") or reference_audio.get("data")
+    if encoded is None:
+        return None
+    raw = base64.b64decode(encoded) if isinstance(encoded, str) else bytes(encoded)
```
zhaochenyang20, line 174: when `reference_audio` is a local path, the newly added `_reference_audio_cache_key()` returns `hash_media_item(reference_audio)` directly. That shared helper samples only the first 8 KiB, the last 8 KiB, and the file size; it does not include the `mtime_ns` promised by the PR description and it does not hash the full file. I reproduced this on the verified checkout: after writing the same path with the same size and identical head/tail bytes but different 100 middle bytes, `_reference_audio_cache_key()` returned `file:fb2bbe2a33cf612a` both times. Once `SGLANG_OMNI_HIGGS_REF_CODE_CACHE=1` is enabled, preprocessing can retrieve the old waveform at `stages.py:163`, and audio_encoder can retrieve the old delayed rows at `stages.py:265`; the later prompt and `Req.extra_key` are then based on the old reference audio. If a user replaces the reference audio, the server can still synthesize with the stale voice-clone condition, which is wrong output rather than a benchmark-only concern.

Give Higgs local paths an explicit cache-key contract instead of using the sampled file hash to represent “file replacement will not stale-hit.” The minimal fix is to use `Path.stat()` for existing local files and include the resolved path, `st_size`, and `st_mtime_ns` in the key; if the implementation wants to avoid filesystem mtime granularity or manually preserved mtimes, use a full content hash instead. HTTP/HTTPS URLs also need an explicit policy: either do not cache mutable URLs by default, or derive the key from loaded response content or strong validation headers, not from the URL string alone.

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -84,6 +117,9 @@ def create_preprocessing_executor(
     raw = Tokenizer.from_file(os.path.join(checkpoint_dir, "tokenizer.json"))
     tokenizer = PreTrainedTokenizerFast(tokenizer_object=raw)
     adapter = HiggsTokenizerAdapter(tokenizer)
+    # Runs on a ThreadedSimpleScheduler pool for preprocessing;
+    reference_waveform_cache = StageOutputCache(max_size=_REF_WAVEFORM_CACHE_MAX_ITEMS)
+    reference_waveform_cache_lock = threading.Lock()
```
zhaochenyang20, line 204: The waveform cache here sets only `_REF_WAVEFORM_CACHE_MAX_ITEMS = 256`, and the audio_encoder code cache at `stages.py:255` also sets only `_REF_CODE_CACHE_MAX_ITEMS = 256`. Higgs allows 100-second reference audio; one 24 kHz float32 mono waveform is about 9.6 MB, so 256 distinct references can retain about 2.4 GB of host tensor data in the preprocessing cache. The delayed-code cache also converts `[T + N - 1, N]` rows into Python `list[list[int]]`, and `StageOutputCache._value_size_bytes()` counts Python ints/lists as effectively zero, so adding `max_bytes` later would not correctly bound that memory. This cache is long-lived server state; all-miss or low-reuse production traffic can keep filling it with little benefit but a high memory ceiling. That leaves the server exposed to sustained RSS growth without matching throughput benefit, and multiple workers or model instances amplify the risk. This also diverges from the neighboring Qwen3 Omni encoder cache pattern that uses `max_bytes=QWEN3_ENCODER_CACHE_MAX_BYTES`.

Add explicit byte budgets to both Higgs caches, and make the cached objects measurable by `StageOutputCache`. The waveform cache can pass `max_bytes` directly, with `_REF_WAVEFORM_CACHE_MAX_ITEMS` reduced if needed to match the real memory budget. The code cache should not keep Python int lists as the long-lived cached object; cache a CPU tensor, a compact array, or a structure with an explicit size, then convert to the list shape only when writing back into `HiggsTtsState`. That makes both the maximum cache benefit and the maximum resource footprint reviewable, configurable, and observable.

tests/unit_test/higgs_tts/test_pipeline.py
zhaochenyang20, line 1: Derived from the root-cause issue above: the new `test_higgs_reference_cache_key_tracks_file_content()` changes the file from `b"a"` to `b"longer"`, so the size changes. That case passes with the sampled hash, but it does not prove the PR description’s semantics that local path size and `mtime_ns` are included so replacement at the same path cannot stale-hit. The fragile real case is same path, same size, identical head/tail bytes, but changed middle audio frames, or at least a same-size replacement where `mtime_ns` changes; the current test covers neither, so it lets the correctness bug above stay green. The risk is that a future cache-key change can keep validating only the easy size-change case and still fail to protect correctness after users replace reference audio.

In the same test or a new one, create a file larger than `head_size + tail_size`, keep the head/tail and size unchanged, modify only the middle content, and assert that the key changes. If the final fix chooses the `size + mtime_ns` contract, the test should explicitly keep the size unchanged, wait for or set an observable mtime change, and assert that the key reflects it; if the fix chooses a full content hash, the test should prove that a same-size middle edit changes the key.

zhaochenyang20 (changes requested):

sglang_omni/models/higgs_tts/stages.py
Ccyest: fixed!

tests/unit_test/higgs_tts/test_pipeline.py
Ccyest: fixed!

Ccyest:
## Benchmark: `SGLANG_OMNI_HIGGS_REF_CODE_CACHE=1` vs off

seed-tts-eval/en, **N=1088, concurrency=16**, `--generate-only --ref-format references`, `higgs-audio-v3-tts-4b-base`, H200. Reporting the two environments **separately**, since GPU sharing materially changes the result. 0 failed requests in all arms; matched workload (identical prompt-token totals).

**A. Dedicated GPU (idle, single tenant) — the clean / intrinsic number**

| metric | cache off | cache on | Δ |
|---|--:|--:|--:|
| Throughput (req/s) | 9.958 | 10.253 | **+2.96%** |
| Latency mean (s) | 1.598 | 1.482 | −7.3% |
| Latency p95 (s) | 2.237 | 2.126 | −5.0% |
| RTF mean | 0.3719 | 0.3406 | −8.4% |

**B. Shared GPU (co-tenants using ~78 GB on the same device)**

| metric | cache off | cache on | Δ |
|---|--:|--:|--:|
| Throughput (req/s) | 8.867 | 9.849 | **+11.07%** |
| Latency p95 (s) | 2.342 | 2.122 | −9.4% |
| RTF mean | 0.3795 | 0.3502 | −7.7% |

**Treat (A) as the intrinsic effect.** (B) is larger but the device was shared with other tenants, so it carries contention variance — not a clean speedup. Direction is consistent across both.

## Profiler — per-stage detail

Request-level profiler, steady-state mid-bench window. Interval = `stage_input_received → stage_complete`; share = fraction of per-request pipeline time.

| stage | share (off) | share (on) | avg/req off (ms) | avg/req on (ms) | role |
|---|--:|--:|--:|--:|---|
| **tts_engine** (AR decode, continuous batching) | 79.37% | 79.27% | 1128 | 1133 | dominant cost — **not touched by this PR** |
| **audio_encoder** (DAC `encode_reference`) | 13.70% | 14.06% | 184 | 194 | the stage the cache targets (hit ⇒ encode skipped) |
| vocoder (DAC decode) | 6.83% | 6.54% | 91 | 91 | already batched (#574) |
| preprocessing (ref-wav load/resample) | 0.10% | 0.13% | 1.3 | 1.8 | waveform cache lives here; ~0 contribution |

Inter-stage hops (shm relay payload) are negligible: `preprocessing→audio_encoder` 0.019 ms, `audio_encoder→tts_engine` 0.024 ms, `tts_engine→vocoder` 0.030 ms.

> The per-stage avgs come from a single steady-state window that mixes cache hits & misses (the cache-on window also drains more requests because it runs faster), so the off↔on per-stage micro-deltas are within-window noise — the cache's real effect is at the **end-to-end** level (A/B above), not in a single window's per-stage avg.

## What the cache does — and what bounds it

- The engine uses **continuous batching**: a request joins the running batch as soon as it is ready, so the (serial) `audio_encoder` does **not** gate when AR starts.
- Per-request **AR time is unchanged (1128 → 1133 ms)** — the cache does **not** speed up generation. It only removes the `audio_encoder` DAC encode + ref-waveform load for the requests that **hit**.
- The gain is therefore bounded by `hit_rate × encoder_share`. This dataset: **666 unique refs / 1088 → 38.8% hit ceiling**, and the 256-entry LRU evicts under that set → removable work ≈ `0.39 × 14% ≈ 5–6%`, consistent with the **dedicated-GPU +3% throughput / −7% latency** in (A).
- On the intended workload (few speakers × many texts) the hit rate — and the gain — would be higher. Worth emitting a hit/miss + eviction metric so hit rate is observable in production.

## Note

The dominant cost is `tts_engine` AR decode (~79%), which this PR does not touch — that's the main lever for further throughput. The `preprocessing` cache contributes ~0%; the real win is the `audio_encoder` cache, sized to the deployment's speaker working set.

zhaochenyang20:
https://github.com/sgl-project/sglang-omni/pull/563

we will add co-author for this PR.

zhaochenyang20 (approved):

Issue #607, opened 2026-05-28
Support MOSS-TTS (OpenMOSS) in SGLang Omni


zhaochenyang20:
Add native support for **MOSS-TTS** ([[OpenMOSS](https://github.com/OpenMOSS/MOSS-TTS)](https://github.com/OpenMOSS/MOSS-TTS) / MOSI.AI, Apache-2.0) to SGLang Omni. Latest release is MOSS-TTS-v1.5. **Help wanted — community contributions welcome.**

## Model

- Target: **MOSS-TTS**, arch `MossTTSDelay`, 8B (also 1.7B `MossTTSLocal`)
- Backbone Qwen3; audio tokenizer **MOSS-Audio-Tokenizer** ("Cat", RVQ, 24 kHz → 12.5 Hz)
- Controls: reference-audio cloning, `${token:N}` duration, pinyin/IPA input, `[pause Xs]` markers

## Scope

- [ ] **Follow an existing Omni TTS model implementation** as the reference — reuse the established pattern/abstractions, don't reinvent
- [ ] `MossTTSDelay` model definition (Qwen3 + multi-head RVQ / delay-pattern decoding)
- [ ] Integrate the MOSS-Audio-Tokenizer detokenizer into the audio output path
- [ ] Request schema: text, reference audio (path/base64), token-count control, `[pause Xs]`
- [ ] **Enable CUDA Graph + `torch.compile` by default**
- [ ] CI: add to the Omni TTS benchmark

## Acceptance

Report **full `Seed-TTS-eval` scores on both subsets** (no sampling):

- **EN**: WER ↓ + SIM ↑
- **ZH**: CER ↓ + SIM ↑

Reference (upstream `MossTTSDelay` 8B): EN WER **1.84** / SIM **70.86**, ZH CER **1.37** / SIM **76.98**.

## Refs

- https://github.com/OpenMOSS/MOSS-TTS
- https://huggingface.co/OpenMOSS-Team/MOSS-TTS-v1.5
- https://github.com/OpenMOSS/sglang

**Labels:** `enhancement`, `help wanted`, `omni`

sjmshsh:
https://github.com/sgl-project/sglang-omni/pull/608

Hi, I’m very interested in this issue. I’ve opened a draft PR and am currently testing it.

draftbk:
[...]

PR #609, opened 2026-05-29
Add MOSS-TTS (OpenMOSS) support

GaokaiZhang:
## Summary
- Native **MOSS-TTS Delay** support as an OmniScheduler 3-stage pipeline:
  `preprocessing -> SGLang AR engine -> MOSS-Audio-Tokenizer vocoder`.
- `MossTTSDelay` = Qwen3 backbone + multi-head RVQ **delay-pattern** decode
  (one text channel + N audio codebook heads); sampling in the model runner.
- MOSS-Audio-Tokenizer detokenizer integrated into the audio output path (24 kHz).
- Request schema: text, reference-audio cloning (path **or** base64 data URI),
  `${token:N}` / `token_count` duration control, `[pause Xs]`, pinyin/IPA
  passthrough; deterministic **no-sampling** defaults (per #607 acceptance).
- SeedTTS benchmark wiring incl. `--token-count auto` duration estimate.
- Content-based radix cache keys over multi-channel prompt rows.

## Status
- **CUDA Graph + torch.compile: on the way (next commit).** This commit is the
  eager baseline integration.

## Checks
- Unit: `pytest tests/unit_test/moss_tts tests/unit_test/serve/test_openai_api.py`
  -> **19 passed** (in `frankleeeee/sglang-omni:dev`).
- Server boots healthy and generates audio (24 kHz):
  `sgl-omni serve --model-path OpenMOSS-Team/MOSS-TTS-v1.5 --config examples/configs/moss_tts.yaml --port 8010`.


zhaochenyang20:
fix the lint and provide the profile results @GaokaiZhang

GaokaiZhang:
Full SeedTTS EN+ZH eager baseline on `OpenMOSS-Team/MOSS-TTS-v1.5`; no CUDA graph / compile. Generation used `concurrency=16`, `token_count=auto`, deterministic/no-sampling defaults from #607. ZH speaker similarity used batch size 1 to avoid WavLM OOM during evaluation.

| `OpenMOSS-Team/MOSS-TTS-v1.5` / eager | EN | WER 23.74% | 1088/1088 | 0 / 0 | 0.373 req/s | 12.93 |
| `OpenMOSS-Team/MOSS-TTS-v1.5` / eager | ZH | CER 28.23% | 2020/2020 | 0 / 0 | 0.337 req/s | 8.65 |

Reference numbers: EN WER 1.84 / SIM 70.86, ZH CER 1.37 / SIM 76.98. Compile + CUDA graph benchmarking is still running and will be posted separately.

GaokaiZhang:
MOSS-TTS compile + CUDA graph benchmark

Config: `OpenMOSS-Team/MOSS-TTS-v1.5`, `examples/configs/moss_tts.yaml`, `torch.compile=True`, CUDA graph captured for bs `[1,2,4,8,16]`, concurrency 16, `token-count=auto`, no sampling/default benchmark params.

| compile + CUDA graph run 1 | EN | WER 22.22% | 1088/1088 | 0 / 0 | 0.377 req/s | 11.72 |
| compile + CUDA graph run 1 | ZH | CER 28.58% | 2020/2020 | 0 / 0 | 0.374 req/s | 7.84 |
| compile + CUDA graph run 2 | EN | WER 22.22% | 1088/1088 | 0 / 0 | 0.375 req/s | 11.78 |
| compile + CUDA graph run 2 | ZH | CER 28.60% | 2020/2020 | 0 / 0 | 0.375 req/s | 7.83 |

Speaker-SIM: run 1 EN 64.22 / ZH 69.50; run 2 EN 64.22 / ZH 69.49. Upstream reference: EN WER 1.84 / SIM 70.86, ZH CER 1.37 / SIM 76.98. CUDA graph is active; the per-step multi-channel sampling loop remains the likely bottleneck.

GaokaiZhang:
MOSS-TTS full-set rerun after the latest fixes:

- Fixed MOSS decode compatibility with the async decode merge by accepting the new `is_lookahead` keyword in `MossTTSModelRunner.prepare_decode`.
- Included the existing compile + CUDA graph path, upstream MOSS sampling defaults, and reduced decode syncs.
- Run config: H200, host GPUs 0/1/4/5/6/7, GPU3 excluded, c=16 per shard, `max_new_tokens=4096`, `token_count=auto`. Throughput is aggregate split wall-clock request/s across the shards.

| MOSS-TTS-v1.5, compile + CUDA graph | EN | 13.70% WER | 67.36 | 1088 | 0 / 0 | 0.718 req/s | 15.365 |
| MOSS-TTS-v1.5, compile + CUDA graph | ZH | 5.42% CER | 75.56 | 2020 | 0 / 0 | 1.439 req/s | 8.084 |

For context, excluding >50% WER/CER outliers gives EN 4.20% and ZH 4.41%. The upstream MOSS reference is EN WER 1.84 / SIM 70.86 and ZH CER 1.37 / SIM 76.98, so this run is stable and graph-enabled but still has an accuracy tail to investigate.

GaokaiZhang:
Three fixes — RTF and the WER failure tail:

1. **Decode**: batched the per-row/per-codebook Python sampling loop (~490 ms/step) into
   batched tensor ops, a faithful port of upstream `generate()` (`model_runner.py`).
2. **Preprocessing**: `max_concurrency=16` (was serial → AR engine starved to batch=1, the
   dominant RTF cost) (`stages.py`).
3. **WER tail**: the audio `LogitsProcessor`s used the text vocab (152k), so sglang's
   `_get_logits` didn't strip the `ParallelLMHead` padding (1025→1088); the leaked columns
   1025..1087 were sampled at temp 1.7 as out-of-range codes → corrupted frames. Fixed by
   sizing each channel's processor to its own vocab + masking padding in the sampler
   (`sglang_model.py`, `model_runner.py`). Referenced from
   https://github.com/sgl-project/sglang-omni/pull/608.

## Results — Seed-TTS-Eval, H200, 4 GPUs, c=16/shard, max_new_tokens=4096

| Split | WER/CER | excl>50% | fails | rtf mean | rtf median | tput/shard | comment3 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| EN | 1.71% | 1.37% | 4/1088 (0.4%)  | 1.40 | 1.33 | ~2.6 req/s  | 13.70% / 15.4 |
| ZH | 1.41% | 1.36% | 1/2020 (0.05%) | 1.05 | 1.03 | ~2.65 req/s | 5.42% / 8.08  |

Refs: HF EN 1.79 / ZH 1.32; PR #608 EN 1.48 (rtf 9.79) / ZH 1.32 (rtf 7.42). We now
match/beat HF on WER with ~8–11× lower rtf and ~7.5× throughput; failure tail gone. ZH rtf
at target (~1.0); EN median 1.33 (mean inflated by a few slow outliers). `max_concurrency=16`
pass underway to push EN rtf median under 1.

zhaochenyang20:
As a reference, from moss team:

batch size 1 reach RTF 0.1 on 1 H200;
batch size 32 reach RTF 0.32 on 1 H200;

sjmshsh:
> Three fixes — RTF and the WER failure tail:
> 
> 1. **Decode**: batched the per-row/per-codebook Python sampling loop (~490 ms/step) into
>    batched tensor ops, a faithful port of upstream `generate()` (`model_runner.py`).
> 2. **Preprocessing**: `max_concurrency=16` (was serial → AR engine starved to batch=1, the
>    dominant RTF cost) (`stages.py`).
> 3. **WER tail**: the audio `LogitsProcessor`s used the text vocab (152k), so sglang's
>    `_get_logits` didn't strip the `ParallelLMHead` padding (1025→1088); the leaked columns
>    1025..1087 were sampled at temp 1.7 as out-of-range codes → corrupted frames. Fixed by
>    sizing each channel's processor to its own vocab + masking padding in the sampler
>    (`sglang_model.py`, `model_runner.py`). Referenced from
>    [Support MOSS-TTS (OpenMOSS) in SGLang Omni #608](https://github.com/sgl-project/sglang-omni/pull/608).
> 
> ## Results — Seed-TTS-Eval, H200, 4 GPUs, c=16/shard, max_new_tokens=4096
> Split	WER/CER	excl>50%	fails	rtf mean	rtf median	tput/shard	comment3
> EN	1.71%	1.37%	4/1088 (0.4%)	1.40	1.33	~2.6 req/s	13.70% / 15.4
> ZH	1.41%	1.36%	1/2020 (0.05%)	1.05	1.03	~2.65 req/s	5.42% / 8.08
> Refs: HF EN 1.79 / ZH 1.32; PR #608 EN 1.48 (rtf 9.79) / ZH 1.32 (rtf 7.42). We now match/beat HF on WER with ~8–11× lower rtf and ~7.5× throughput; failure tail gone. ZH rtf at target (~1.0); EN median 1.33 (mean inflated by a few slow outliers). `max_concurrency=16` pass underway to push EN rtf median under 1.

Good job. I’m glad to see this. Ref: https://github.com/sgl-project/sglang-omni/pull/608. I’ll close this PR.

sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -0,0 +1,496 @@
...
+            [float(d.text_temperature) for d in datas],
+            dtype=torch.float32,
+            device=device,
+        )
+        audio_temp = torch.tensor(
+            [float(d.audio_temperature) for d in datas],
+            dtype=torch.float32,
+            device=device,
+        )
+        text_top_p = float(datas[0].text_top_p)
```
zhaochenyang20: Only the first sample's parameter get used?

extend `_sample_tokens` so `top_p`, `top_k`, and repetition penalty are per-row. After the fix, add a two-request batch test where the requests use different `text_top_k` or `audio_top_k`, and assert that the mask and sampled output use each request's own parameters instead of `datas[0]`.

sglang_omni/models/moss_tts/request_builders.py
```diff
@@ -0,0 +1,566 @@
...
+        tts_params.get("language") or params.get("language")
+    )
+    instructions = _resolve_optional_text(
+        tts_params.get("instructions")
+        or tts_params.get("instruct")
+        or params.get("instructions")
+        or params.get("instruct")
+    )
+    return MossTTSState(
+        text=text,
```
zhaochenyang20, line 274: `_resolve_token_count` parses `${token:N}` from the beginning of the text, but `build_moss_tts_state` still stores the original `text` unchanged in `MossTTSState`. The upstream v1.5 processor duration API is `processor.build_user_message(text=..., tokens=...)`, which writes the count into the `- Tokens:` field.

When parsing `${token:N}`, return the stripped text together with the parsed count and store the cleaned content in `MossTTSState.text`; validate `token_count`, `duration_tokens`, and prefix-derived counts with the same `> 0` rule. The test should assert that the processor receives text without `${token:N}`, while the `tokens` field still equals the parsed value, and that `[pause Xs]`, pinyin, and IPA passthrough remain unchanged.

sglang_omni/models/moss_tts/request_builders.py
```diff
@@ -0,0 +1,566 @@
...
+        context = _PREPROCESSING_CONTEXT
+    if context is None:
+        raise RuntimeError(
+            "MOSS-TTS preprocessing context is not initialized; "
+            "create_preprocessing_executor must register it before requests run"
+        )
+
+    prepared = _prepare_moss_tts_request(payload, processor=context.processor)
+    with _PREPARED_REQUESTS_LOCK:
+        _PREPARED_REQUESTS[payload.request_id] = prepared
```
zhaochenyang20: Do not let the compute function permanently insert into the global handoff without knowing the abort state. A simple fix is to register the handoff only after preprocessing is ready to be emitted successfully by the scheduler, or to attach a request lifecycle token and delete late inserts after abort. At minimum, add a test that simulates abort during `_prepare_moss_tts_request` and asserts `_PREPARED_REQUESTS` is empty at the end.

sglang_omni/models/moss_tts/stages.py
```diff
@@ -0,0 +1,363 @@
...
+    if isinstance(auto_mapping, dict):
+        auto_mapping.setdefault("AutoModel", "PreTrainedModel")
+        if not hasattr(processing_utils, "MODALITY_TO_BASE_CLASS_MAPPING"):
+            processing_utils.MODALITY_TO_BASE_CLASS_MAPPING = auto_mapping
+
+    # MOSS-Audio-Tokenizer is loaded through AutoModel and is a PreTrainedModel.
+    # Transformers 4.57 otherwise rejects it in ProcessorMixin's optional
+    # audio_tokenizer branch before the model can be moved to the vocoder stage.
+    if hasattr(processing_utils, "PreTrainedAudioTokenizerBase"):
+        processing_utils.PreTrainedAudioTokenizerBase = PreTrainedModel
```
zhaochenyang20: Turn the compatibility patch into a local context: save the original `PreTrainedConfig`, `MODALITY_TO_BASE_CLASS_MAPPING`, and `PreTrainedAudioTokenizerBase`, temporarily replace them only around `_load_moss_processor_class` / `processor_cls.from_pretrained`, and restore them after loading. If the remote processor truly needs one replacement at runtime, keep only that minimal replacement and document the exact Transformers version and failure point.

.gitignore
```diff
@@ -45,3 +45,5 @@ results/
 benchmarks/eval/*.edit.lock
 .claude/*
 !.claude/skills/
+PR.md
+comment1.md
```
zhaochenyang20: delete these two

zhaochenyang20 (changes requested):

0xjeffro:
Hey Team, here are my reproduction results on 1×H200 with full-set, token-count=auto, and they're pretty consistent with the previous results. I've updated the benchmark and the corresponding launch commands in [a6b1dc9](https://github.com/sgl-project/sglang-omni/pull/609/commits/a6b1dc9cee292d779a5a7f4c1426245901606614). On my end, Qwen3-TTS has a high probability of runaway at c=16, will included the data in the next commit.

**Accuracy**

| | corpus | mean | median | std | excl>50% |
|---|---|---|---|---|---|
| EN | 1.68% | 1.74% | 0% | 8.6% | 1.32%（4 outlier） |
| ZH | 1.36% | 1.31% | 0% | 3.9% | 1.27%（2 outlier） |

**Generation speed**

| | latency | p95 | rtf | qps | tok/req-s |
|---|---|---|---|---|---|
| EN | 3.449 | 4.141 | 0.811 | 2.312 | 75.2 |
| ZH | 3.608 | 4.153 | 0.635 | 2.213 | 77.8 |

**ASR speed**

| | asr_lat | asr_rtf | thru |
|---|---|---|---|
| EN | 0.370 | 0.0854 | 2.70 |
| ZH | 0.109 | 0.0192 | 9.18 |

sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -0,0 +1,540 @@
...
+        pad_token_id = int(cfg.pad_token_id)
+        gen_slot = int(cfg.audio_assistant_gen_slot_token_id)
+        delay_slot = int(cfg.audio_assistant_delay_slot_token_id)
+        audio_start = int(cfg.audio_start_token_id)
+        audio_end = int(cfg.audio_end_token_id)
+        im_end = int(cfg.im_end_token_id)
+        audio_pad_code = int(cfg.audio_pad_code)
+
+        # --- gather per-request delay state ---
+        audio_lengths = torch.tensor(
```
yxs: `_sample_rows` rebuilds a dozen small tensors via `torch.tensor([... for d in datas], ...)` (audio_lengths/delayed/is_audio/gen_steps/temps/top_p/top_k...) every decode step, then ends with three `.tolist()` D2H calls (`audio_lengths`/`delayed`/`is_audio`) followed by a Python for-loop scattering state back into each `data`. Together with the per-row grouping `for row, (k, p) in enumerate(zip(top_k_row.tolist(), top_p_row.tolist()))` in `_sample_tokens`, this is fixed host overhead on every step.

Consider gather these per-request scalars into persistent device buffers updated in place, avoiding the per-step rebuild + `.tolist()`.

sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -0,0 +1,540 @@
...
+        self._pending_embeds = None
+        if rows is None or embeds is None:
+            return
+
+        eos_id = int(self.model.config.im_end_token_id)
+        for row_idx, sched_req in enumerate(scheduler_output.requests):
+            req_output = outputs[sched_req.request_id]
+            if req_output.data is None or int(req_output.data) == eos_id:
+                continue
+            sched_req.data.output_rows.append(rows[row_idx].detach().cpu().clone())
```
yxs: `sched_req.data.output_rows.append(rows[row_idx].detach().cpu().clone())` runs a `.cpu()` per request, per decode step. `.cpu()` is a device→host copy with an implicit sync, so this inserts up to `batch` sync points into the AR hot loop every step. 

Downstream `apply_sglang_moss_tts_result` already does `torch.stack(output_rows)`, which can run on GPU and be moved to host once at the end; keeping rows on GPU also removes the `.to(device=device)` re-upload inside `_apply_audio_repetition_penalty`.

sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -0,0 +1,540 @@
...
+                (0, self.model.hidden_size),
+                device=forward_batch.input_ids.device,
+                dtype=self.model.dtype,
+            )
+        return torch.cat(pieces, dim=0).to(
+            device=forward_batch.input_ids.device,
+            dtype=self.model.dtype,
+        )
+
+    def _write_decode_input_embedding(
```
yxs, line 101: `_decode_input_embedding` has only `max_running_requests`(=16) rows, while `_write_decode_input_embedding` does `weight[:batch_size].copy_(stacked)` and writes `input_ids[:batch_size] = arange(batch_size)` unconditionally. Today `max_running_requests` caps the batch, so `batch_size` cannot exceed 16 and there is no active bug; but the sibling `qwen3_tts/model_runner.py:186-189` guards the identical trick with `if batch_size > decode_feedback_embedding.num_embeddings: raise RuntimeError(...)`. This one omits it: if some future config lets the running batch exceed the embedding row count, `weight[:bs].copy_(stacked)` will shape-mismatch and `embedding(arange(bs))` will index out of range, with a hard-to-localize error.

sglang_omni/models/moss_tts/stages.py
```diff
@@ -0,0 +1,405 @@
...
+    checkpoint_dir = _resolve_checkpoint(model_path)
+    logger.info("Loading MOSS-TTS processor from %s on %s", checkpoint_dir, device)
+    try:
+        with _moss_transformers_processor_compat():
+            processor_cls = _load_moss_processor_class(checkpoint_dir)
+            processor = processor_cls.from_pretrained(
+                checkpoint_dir,
+                trust_remote_code=True,
+            )
+    except ImportError as exc:
```
yxs: The two except branches do exactly the same thing

sglang_omni/models/moss_tts/stages.py
```diff
@@ -0,0 +1,405 @@
...
+    ]:
+        processor_cls.attributes = ["tokenizer"]
+    return processor_cls
+
+
+def _normalize_moss_processor_config(processor: Any) -> None:
+    model_config = getattr(processor, "model_config", None)
+    if model_config is None:
+        return
+    for attr, default in (
```
yxs: why`_normalize_config` (model side) and `_normalize_moss_processor_config` (processor side) each maintain a nearly identical default table?

gaoyang07:
Could we consider an OpenMOSS-style multi-channel CUDA graph runner / buffer shape for MOSS-TTS, maybe as a guarded follow-up?

In our open-sourced SGLang version, we support rank-2 decode `input_ids` when `channels > 1` and allocates the logits buffer as `[max_num_token, sum(vocab_size_list)]`:
https://github.com/OpenMOSS/sglang/blob/6c3a99cf413f623541aae7ee14744f5a3c062a47/python/sglang/srt/model_executor/cuda_graph_runner.py#L136-L164

The captured `ForwardBatch` then carries these buffers:
https://github.com/OpenMOSS/sglang/blob/6c3a99cf413f623541aae7ee14744f5a3c062a47/python/sglang/srt/model_executor/cuda_graph_runner.py#L866-L874

And the MOSS-TTS model consumes rank-2 channel input ids in the embedding path:
https://github.com/OpenMOSS/sglang/blob/6c3a99cf413f623541aae7ee14744f5a3c062a47/python/sglang/srt/models/moss_tts_delay.py#L150-L198

This simple change, capturing multi-channel token rows directly gives a large speedup. This seems cleaner than the current rank-1 placeholder ids + `_decode_input_embedding` path, but it also changes shape contracts around `ForwardBatch.input_ids`, graph buffers, logits metadata, cache/chunk/retract, and TP, so it probably needs to be introduced carefully as a model-specific / explicit multi-channel graph path.

zhaochenyang20:
1. cookbook for moss；
2. add performance of moss, voxtral and qwen3 tts on H200 here: https://github.com/sgl-project/sglang-omni/blob/main/benchmarks/eval/benchmark_tts_seedtts.py

sglang_omni/models/moss_tts/request_builders.py
```diff
@@ -0,0 +1,601 @@
...
+    generation_kwargs: dict[str, Any] = {
+        "max_new_tokens": int(
+            params.get("max_new_tokens") or MOSS_TTS_DEFAULT_MAX_NEW_TOKENS
+        ),
+        # MOSS-TTS is a sampling model: the checkpoint's own generate() ships
+        # these defaults and the upstream reference scores were produced with
+        # them. Greedy (temperature=0) collapses a reference-conditioned codec
+        # LM into copying the reference audio, which destroys WER/CER. The
+        # "no sampling" eval requirement is met via reproducibility (fixed
+        # server random_seed + pytorch sampling backend), not temperature=0.
```
zhaochenyang20: 1. remove useless comment. If you think useful, make it with `note (chenyang):`
2.

zhaochenyang20: `/v1/audio/speech` now forwards `seed` into `tts_params` and `SamplingParams`, but the MOSS request builder only constructs `text_temperature`, `audio_temperature`, `top_p`, `top_k`, and `audio_repetition_penalty`; it never stores the seed in `MossTTSSGLangRequestData`, and `_sample_tokens` later calls global `torch.multinomial` directly. When a client retries the same prompt, reference audio, and `seed=123`, reproducibility depends on the process-wide PyTorch RNG state and batch interleaving rather than the user-provided seed. The same custom path also bypasses SGLang `SamplingParams.verify()` for public sampling fields, so inputs such as `repetition_penalty=0`, negative temperatures, or out-of-range `top_p` can silently become greedy behavior, produce infinite logits, or make the same API field mean different things for MOSS versus Qwen/Higgs.

zhaochenyang20: Add a validated seed and MOSS-specific sampling policy to `MossTTSState` / `MossTTSSGLangRequestData`; follow the Qwen3-TTS pattern by deriving text/audio sampling seeds from the public seed and using seeded categorical sampling inside `_sample_tokens`, or explicitly reject seed for MOSS if it is not supported. At the same time, validate `temperature`, `top_p`, `top_k`, and `audio_repetition_penalty` at the public boundary, and add tests for two cases: same-seed MOSS batch retries are reproducible, and invalid sampling parameters fail with a clear `ValueError`.

sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -0,0 +1,540 @@
...
+        n_vq: int,
+    ) -> torch.Tensor:
+        """Vectorized MOSS-TTS delay-pattern step over the whole batch.
+
+        Mirrors one step of the upstream ``MossTTSDelayModel.generate`` loop but
+        as batched tensor ops, so a decode step issues a couple of sampling
+        kernels instead of ``batch * (1 + n_vq)`` Python-level samples (the old
+        per-row/per-codebook loop was ~490 ms/step and dominated RTF). Per-request
+        delay state (audio_length / delayed_length / is_audio) is gathered from
+        the scheduler ``data`` objects and scattered back at the end.
```
zhaochenyang20: make comments short.

sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -0,0 +1,540 @@
...
+
+        pad_token_id = int(cfg.pad_token_id)
+        gen_slot = int(cfg.audio_assistant_gen_slot_token_id)
+        delay_slot = int(cfg.audio_assistant_delay_slot_token_id)
+        audio_start = int(cfg.audio_start_token_id)
+        audio_end = int(cfg.audio_end_token_id)
+        im_end = int(cfg.im_end_token_id)
+        audio_pad_code = int(cfg.audio_pad_code)
+
+        # --- gather per-request delay state ---
```
zhaochenyang20: remove useless comment. If you think useful, make it with `note (gaokai, jinjiang):`

sglang_omni/models/moss_tts/model_runner.py
zhaochenyang20: Keep `output_rows` as GPU tensors until `apply_sglang_moss_tts_result` finally needs to hand the rows to the vocoder or host relay; `_apply_audio_repetition_penalty` can already consume GPU history. For delay state and sampling policy, prefer Qwen3-TTS-style per-batch device buffers and tensor-mask top-k/top-p handling instead of rebuilding Python groups via `.tolist()` every step. As a minimal short-term fix, at least remove the per-request `.cpu()` and include profiler evidence comparing per-step synchronization count and RTF.

sglang_omni/models/moss_tts/request_builders.py
```diff
@@ -0,0 +1,601 @@
...
+    If the handoff has not been published yet (abort raced ahead of
+    ``preprocess_moss_tts_payload``), record the id so the pending insert is
+    discarded rather than leaking.
+    """
+
+    rid = str(request_id)
+    with _PREPARED_REQUESTS_LOCK:
+        if _PREPARED_REQUESTS.pop(rid, None) is None:
+            _ABORTED_REQUESTS.add(rid)
+
```
zhaochenyang20, line 163: Do not permanently tombstone a request unless the code knows it is in-flight preprocessing. One option is to register an in-flight token when `preprocess_moss_tts_payload()` begins; cleanup should record a late abort only when that token exists, and compute should remove the token/tombstone when it drops a late insert or successfully publishes the handoff. Add a test distinct from the existing abort-during-prepare case: call abort cleanup first, simulate the scheduler consuming the abort before compute, assert `_ABORTED_REQUESTS` is empty afterward, and assert the same request id can later publish a normal handoff.

sglang_omni/models/moss_tts/stages.py
```diff
@@ -0,0 +1,405 @@
...
+    All three drifts are load-time only (they are consulted by
+    ``ProcessorMixin.from_pretrained`` / the remote config code, not at runtime):
+    - the remote config references the renamed ``PreTrainedConfig``;
+    - ``MODALITY_TO_BASE_CLASS_MAPPING`` / the ``AutoModel`` entry must exist for
+      the optional audio-tokenizer branch;
+    - Transformers 4.57's ``ProcessorMixin`` rejects the AutoModel-loaded
+      MOSS-Audio-Tokenizer unless ``PreTrainedAudioTokenizerBase`` accepts a
+      plain ``PreTrainedModel``.
+    Leaking these into the global namespace can mask drift in unrelated models,
+    so they are scoped and reverted here.
```
zhaochenyang20: make comments short.

sglang_omni/models/moss_tts/stages.py
zhaochenyang20: Move the MOSS special-token defaults into a shared constant or small helper, for example `MOSS_TTS_SPECIAL_TOKEN_DEFAULTS`, and have both processor config and model config normalization use it. If `audio_pad_code` must depend on `audio_vocab_size`, let the shared helper accept `audio_vocab_size` and return one consistent set of defaults so one path does not hard-code `1024` while the other uses a variable.

sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -0,0 +1,540 @@
...
+                rows.append(queue.popleft())
+            else:
+                rows.append(queue.pop(0))
+        stacked = torch.stack(rows, dim=0).to(device=weight.device, dtype=weight.dtype)
+        with torch.no_grad():
+            weight[:batch_size].copy_(stacked)
+
+        row_ids = torch.arange(
+            batch_size,
+            dtype=torch.long,
```
zhaochenyang20, line 136: Add two fail-fast assertions at the start of `_write_decode_input_embedding`: `forward_batch.input_ids.numel() >= batch_size`, and `batch_size <= embedding.num_embeddings`. The error message should directly state that the MOSS decode batch exceeds the staged embedding rows.

zhaochenyang20: This is not an active bug today because `create_sglang_tts_engine_executor()` fixes both `max_running_requests` and graph batch sizes at 16, and `_decode_input_embedding` is allocated from the current global server args. However, line 123 directly runs `weight[:batch_size].copy_(stacked)`, and line 131 writes `arange(batch_size)` back into `input_ids`; if a later config allows the running batch to exceed the embedding rows, the failure becomes a shape mismatch or embedding index-out-of-range far away from the real cause. The Qwen3-TTS sibling runner already has an explicit `batch_size > decode_feedback_embedding.num_embeddings` guard for the same trick, and MOSS should keep the same contract.

sglang_omni/models/moss_tts/stages.py
zhaochenyang20: remove them

zhaochenyang20: Collapse this into a single `except Exception as exc:`, or if `ImportError` really needs distinct handling, make that branch report dependency-install guidance while the non-ImportError branch reports checkpoint/remote-code failures. With the current identical messages, one branch is clearer.

tests/unit_test/moss_tts/test_pipeline.py
zhaochenyang20, line 1: Add three focused tests in `tests/unit_test/moss_tts/test_pipeline.py`: same-seed `_sample_tokens` or `_sample_rows` reproduces under the same positions; `build_moss_tts_state` / generation kwargs raises `ValueError` for invalid sampling fields; and pre-start abort cleanup does not leave `_ABORTED_REQUESTS` behind or block a later normal preprocessing call with the same id.

zhaochenyang20 (changes requested):

GaokaiZhang:
## PR #608 review fixes
- Round 1: per-request sampling params (top_p/top_k/rep), `${token:N}` strip+validate,
  abort-safe handoff, scoped Transformers compat patch.
- Round 2: validate public sampling fields → `ValueError` (temp/top_p/top_k/rep/seed);
  per-request `seed` threaded into `_sample_tokens` (reproducible for bs==1); keep
  `output_rows` on GPU (drop per-step `.cpu()`); `_write_decode` fail-fast guard;
  abort lifecycle token (no permanent tombstone); shared `MOSS_TTS_SPECIAL_TOKEN_DEFAULTS`;
  collapse duplicate `except`; shorter comments. +tests (18 MOSS unit tests pass).

## Results — Seed-TTS-Eval, H200, c=16 client / preprocessing 8, compile + CUDA graph

| Split | WER/CER | excl >50% | fails | rtf mean | rtf median |
| EN | 1.77% | 1.27% | 5/1088 (0.5%) | 1.39 | 1.30 |
| ZH | 1.37% | 1.37% | 0/2020 (0.0%) | 1.02 | 0.99 |

sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -0,0 +1,561 @@
...
+            dtype=torch.float32,
+            device=device,
+        )
+        # Batched multinomial shares one RNG, so a per-request seed is only
+        # well-defined for a single stream; bs==1 + seed is reproducible per
+        # position, larger batches fall back to the global (server-seeded) RNG.
+        generator = None
+        seed = getattr(datas[0], "seed", None)
+        if batch_size == 1 and seed is not None:
+            generator = torch.Generator(device=device)
```
zhaochenyang20: Reuse the Qwen3-TTS pattern: derive per-row, per-position, per-channel seeds from the public seed, keep them in device-side batch metadata, and sample with `multinomial_with_seed` or an equivalent per-row sampler. The test should build a two-request batch, retry at the same positions, and prove that neighboring rows do not perturb same-seed output. If batched seed is out of scope for this PR, the MOSS path should explicitly reject `seed` instead of accepting a public field that only works at batch size 1.

zhaochenyang20: `/v1/audio/speech` now forwards `seed` into `tts_params` and `SamplingParams`, and `MossTTSSGLangRequestData` stores it, but the actual sampler only creates a `torch.Generator` when `batch_size == 1` and only reads `datas[0].seed`. Under normal server concurrency, batched requests fall back to the process-wide/global RNG.

sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -0,0 +1,561 @@
...
+            generator.manual_seed(
+                (int(seed) * 1_000_003 + int(datas[0].generation_steps))
+                & ((1 << 63) - 1)
+            )
+
+        # --- text channel ---
```
zhaochenyang20: remove this comment. And, consider this:

an explicit `max_new_tokens=0` is treated as missing and replaced with the default 4096.

zhaochenyang20: Distinguish `None` from explicit numeric values. Only `None` should use `MOSS_TTS_DEFAULT_MAX_NEW_TOKENS`; all explicit values should be converted to int and validated as `> 0`. Add MOSS custom-builder tests for `max_new_tokens=0` and negative values so this path no longer bypasses public sampling/generation validation.

sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -0,0 +1,561 @@
...
+            | (next_text == delay_slot)
+        )
+        audio_lengths = audio_lengths + increment.long()
+        audio_lengths[next_text == audio_end] = 0
+        delayed[(delayed == _INT64_MAX) & (next_text == delay_slot)] = 0
+        not_inf = delayed != _INT64_MAX
+        delayed[not_inf] = delayed[not_inf] + 1
+        delayed[delayed > n_vq] = _INT64_MAX
+
+        # --- scatter state back to per-request data ---
```
zhaochenyang20: delete these useless comments

zhaochenyang20: delete these useless comments

sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -0,0 +1,561 @@
...
+            data.audio_length = int(audio_lengths_l[i])
+            data.delayed_length = (
+                _INF_DELAY if delayed_l[i] == _INT64_MAX else int(delayed_l[i])
+            )
+            data.is_audio = bool(is_audio_l[i])
+
+        rows = torch.empty((batch_size, n_vq + 1), dtype=torch.long, device=device)
+        rows[:, 0] = next_text
+        rows[:, 1:] = next_audio
+        return rows
```
zhaochenyang20, line 387: Keep delay state, sampling policy, seed, and position in per-batch device buffers, and apply top-k/top-p through tensor masks instead of per-step `.tolist()` plus Python grouping. Move to host only once at the vocoder/host relay boundary.

zhaochenyang20 (changes requested):
Add `docs/cookbook/moss_tts.md` following the existing TTS cookbook structure. It should include serve commands, the MOSS config, OpenAI speech API examples, reference cloning, base64/path references, duration controls, control-token examples, seed limitations, benchmark commands, and links to the H200 reference numbers.

GaokaiZhang:
## Round-3 review fixes
- **Seed (a)** — replaced the bs==1-only `torch.Generator` hack with Qwen3-TTS's
  `multinomial_with_seed`: a blake2b per-request seed combined with a per-(step, channel)
  `position`. A token now depends only on its own seed/position, so `seed` is reproducible at
  **any** batch size and never perturbed by batch neighbours. Test proves reproducibility +
  neighbour-independence.
- **max_new_tokens (b)** — only `None` falls back to the 4096 default; explicit values are
  `int()`-converted and validated `> 0` (explicit `0` no longer silently becomes 4096). Builder
  tests for `0` / negatives.
- **Comments (c)** — removed the `# --- … ---` divider comments.
- **Sampling path (d)** — seed/positions are per-row device tensors; top-k/top-p are fully
  vectorized via tensor masks (no `.tolist()` + Python `(k,p)` grouping); 3 per-step state syncs
  collapsed to 1. *Deferred:* persisting delay-state in cross-step device buffers needs stable
  per-request slots in the dynamically-batched OmniScheduler — a scheduler-level refactor.

## Validation — Seed-TTS-Eval full set, H200, c=8, token-count=auto

| Split | WER/CER | excl >50% | fails | rtf mean | rtf median |
| EN (1088) | 1.67% | 1.19% | 5 (0.5%) | 0.93 | 0.88 |
| ZH (2020) | 1.37% | 1.37% | 0 (0.0%) | 0.77 | 0.76 |

No regression vs the committed baseline (EN 1.68% / ZH 1.36%) — the seed/sampler rewrite is
distribution-preserving. RTF stays < 1; the small rise vs the dedicated-reference run is uniform
across both languages (parallel EN+ZH host-CPU contention), not a code regression. 22/22 MOSS
unit tests pass.

## Also in this round
- Added `docs/cookbook/moss_tts.md` (serve/config, OpenAI API, reference cloning incl.
  path/URL/base64, `${token:N}` duration, markup, seed limitations, benchmarks + H200 numbers);
  registered in the toctree + TTS overview.

zhaochenyang20 (approved):

Issue #603, opened 2026-05-28
[Perf] Higgs TTS: batched audio encoder


xinlij:
## Motivation

The audio encoder stage (`create_audio_encoder_executor`) collects up to `max_batch_size=8` requests via `SimpleScheduler`, but no `batch_compute_fn` is provided — each request is still encoded serially with a single `codec.encode_reference` call on a `[1, 1, L]` waveform.

`HiggsAudioV2TokenizerModel.encode()` already accepts batched input `(batch_size, channels, num_samples)` and returns `(batch_size, num_quantizers, codes_length)` (see `_vendored/higgs_audio_v2_tokenizer_hf.py` line 789). Stacking same-length waveforms into `[B, 1, L]` and running a single forward pass would reduce per-request encoder latency at concurrency ≥ 2 — the same gap that #569 / PR #574 closed for the vocoder.

## Current path

```python
# HiggsAudioCodec.encode_reference — single waveform at a time
wav = _to_mono_3d(waveform)                          # [1, 1, L]
codes_BNT = self.model.encode(wav).audio_codes       # [1, N, T]
return codes_BNT.squeeze(0).transpose(0, 1).cpu()    # [T, N]

# stages.py — _encode()
ref_codes_TN = codec.encode_reference(waveform, sample_rate=24000)  # one at a time
```

`SimpleScheduler` is instantiated with `max_batch_size=8` but without `batch_compute_fn`, so batch collection is unused.

## Proposed change

1. Add `encode_batch` to `HiggsAudioCodec`, using **bucketing by waveform length** rather than zero-padding to `L_max`.

The acoustic encoder uses symmetric Conv1d padding (`padding = (kernel - 1) // 2`), making it non-causal — zero-padding shorter waveforms to a longer `L_max` corrupts codes near the boundary, the same issue PR #574 encountered on the decoder side. Bucketing avoids this entirely: only same-length waveforms are stacked into a single forward pass; items with different lengths are processed in separate per-bucket calls.

```python
def encode_batch(self, waveforms: list[torch.Tensor]) -> list[torch.Tensor]:
    """Encode a batch of [1, 1, L_i] waveforms → list of [T_i, N] code tensors."""
    from collections import defaultdict
    buckets: dict[int, list[int]] = defaultdict(list)
    for i, w in enumerate(waveforms):
        buckets[w.shape[-1]].append(i)

    results = [None] * len(waveforms)
    for length, indices in buckets.items():
        if len(indices) == 1:
[...]

PR #610, opened 2026-05-29
Add HiggsAudioCodec.encode_batch (batched audio encoder)

xinlij:
## Motivation

`HiggsAudioV2TokenizerModel.encode()` natively supports batched forward `[B, 1, L] -> [B, N, T]`, but `HiggsAudioCodec` only exposed single-clip `encode_reference`. For offline / bulk encoding use cases (dataset pre-encoding, batch reference-audio ingestion, eval-time preprocessing), users had to loop externally and could not exploit GPU parallelism.

This PR adds `HiggsAudioCodec.encode_batch` as the standard API for offline / bulk encoding. We **also** evaluated wiring it into the serving pipeline's `audio_encoder` stage — that turned out to regress throughput, so it is not wired in. See appendix.

### `HiggsAudioCodec.encode_batch(waveforms)` — new public API

Accepts `List[Tensor[1, 1, L_i]]` and returns the per-clip codes. Design notes:

- **Length bucketing.** The acoustic encoder uses symmetric (non-causal) Conv1d padding, so zero-padding shorter waveforms to a common `L_max` and running a single forward corrupts the codes. We bucket by exact length — **only equal-length waveforms are stacked into a single `[B, 1, L]` forward**; unequal lengths fall back to per-bucket `encode_reference` calls.
- **Short-clip alignment.** Clips shorter than `SAMPLE_RATE` samples (1 s @ 24 kHz) are zero-padded to 1 s before bucketing, matching `encode_reference` behaviour.
- **Order preservation.** Output order matches input order across buckets.
- **Empty input.** Returns an empty list.

### Unit tests — `tests/unit_test/higgs_tts/test_pipeline.py`

Five tests covering the above:

- `test_higgs_audio_codec_encode_batch_empty` — empty input
- `test_higgs_audio_codec_encode_batch_same_length_batched` — equal-length stacked into one forward
- `test_higgs_audio_codec_encode_batch_short_waveforms_padded_and_batched` — short clips padded to 1 s before bucketing
- `test_higgs_audio_codec_encode_batch_different_lengths_separate_calls` — unequal lengths use separate per-bucket calls
- `test_higgs_audio_codec_encode_batch_order_preserved` — output order preserved across buckets

## Use Cases

- Offline dataset pre-encoding — encode an entire reference-audio corpus in one pass
- Bulk reference-audio ingestion / migration tools
- Eval / benchmark preprocessing

## Appendix: Why this is NOT wired into the serving pipeline

We also evaluated wiring `batch_compute_fn=_encode_batch` into `create_audio_encoder_executor` (mirroring the batched vocoder in #574). Full SeedTTS EN evaluation (1088 samples) on **1× NVIDIA H200**, temperature=0.8, top-k=50. Base commit: `10d0402` (main). Model: `boson-sglang/higgs-audio-v3-tts-4b-base`.

### Speed

| Metric | c=1 main | c=1 batch | c=16 main | c=16 batch | c=32 main | c=32 batch |
|--------|----------|-----------|-----------|------------|-----------|------------|
| Latency mean (s) | 0.662 | 0.675 | 1.639 | 2.003 | 2.921 | 3.897 |
| Latency median (s) | 0.588 | 0.599 | 1.513 | 1.872 | 2.762 | 3.827 |
| Latency p95 (s) | 0.861 | 0.882 | 2.168 | 2.703 | 3.530 | 4.636 |
| Latency p99 (s) | 1.183 | 1.254 | 2.808 | 3.192 | 4.065 | 5.207 |
| RTF mean | 0.146 | 0.149 | 0.380 | 0.476 | 0.695 | 0.967 |
| **Throughput (req/s)** | **1.511** | **1.481** | **9.390** | **7.678** | **10.823** | **8.097** |
| Output tokens (mean) | 124 | 124 | 122 | 126 | 124 | 126 |

### Quality

| Metric | c=1 main | c=1 batch | c=16 main | c=16 batch | c=32 main | c=32 batch |
|--------|----------|-----------|-----------|------------|-----------|------------|
| WER corpus (excl >50%) | 1.35% | 1.35% | 1.40% | 1.32% | 1.47% | 1.30% |
| >50% WER samples | 1 | 1 | 3 | 4 | 0 | 2 |

### Throughput delta

| Concurrency | main (req/s) | batch (req/s) | Delta |
|-------------|--------------|---------------|-------|
| 1 | 1.511 | 1.481 | **-2.0%** |
| 16 | 9.390 | 7.678 | **-18.2%** |
| 32 | 10.823 | 8.097 | **-25.2%** |

### Root cause

The encoder and AR engine share a single GPU, and the AR engine is the actual bottleneck (~79% of per-request time). Batching the encoder makes its GPU usage burstier and steals cycles from the AR engine — the added latency scales with concurrency (+364 ms at c=16, +976 ms at c=32) and lands downstream on AR decode, not in the encoder itself. This contrasts with the vocoder (#574), which sits **after** the AR engine and therefore does not contend with the bottleneck.

The serving-side "avoid recomputing the same reference's codes" win is already handled by the reference-code cache in #605.

## Related Issues

#603

## Checklist

- [x] Add unit tests.

### DIFF
--- a/sglang_omni/models/higgs_tts/audio_codec.py
+++ b/sglang_omni/models/higgs_tts/audio_codec.py
@@ -186,6 +186,50 @@ class HiggsAudioCodec:
         return codes_BNT.squeeze(0).transpose(0, 1).to(torch.long).cpu()
 
     @torch.no_grad()
+    def encode_batch(self, waveforms: list[torch.Tensor]) -> list[torch.Tensor]:
+        """Batch-encode ``[1, 1, L_i]`` float32 waveforms → ``[T_i, N]`` int64 code tensors.
+
+        Only same-length items are batched into a single forward pass.
+        The non-causal acoustic encoder corrupts shorter items when padded to
+        a longer ``L_max``, so mixed-length inputs are encoded per-bucket.
+        Clips shorter than ``SAMPLE_RATE`` samples are zero-padded before
+        bucketing, matching the behaviour of :meth:`encode_reference`.
+        """
+        if not waveforms:
+            return []
+        if len(waveforms) == 1:
+            return [self.encode_reference(waveforms[0])]
+
+        padded: list[torch.Tensor] = []
+        for w in waveforms:
+            wav = w.to(torch.float32)
+            if wav.shape[-1] < self.SAMPLE_RATE:
+                wav = F.pad(wav, (0, self.SAMPLE_RATE - wav.shape[-1]))
+            padded.append(wav)
+
+        buckets: dict[int, list[int]] = {}
+        for i, w in enumerate(padded):
+            buckets.setdefault(w.shape[-1], []).append(i)
+
+        results: list[torch.Tensor | None] = [None] * len(waveforms)
+        for indices in buckets.values():
+            if len(indices) == 1:
+                results[indices[0]] = self.encode_reference(padded[indices[0]])
+            else:
+                batch = torch.cat([padded[i] for i in indices])  # [B, 1, L]
+                batch = batch.to(device=self.device, dtype=self._dtype)
+                codes_BNT = self.model.encode(batch).audio_codes  # [B, N, T]
+                for j, idx in enumerate(indices):
+                    results[idx] = codes_BNT[j].transpose(0, 1).to(torch.long).cpu()
+
+        out: list[torch.Tensor] = []
+        for i, result in enumerate(results):
+            if result is None:
+                raise RuntimeError(f"encode_batch did not produce codes for item {i}")
+            out.append(result)
+        return out
+
+    @torch.no_grad()
     def decode(self, codes_TN: torch.Tensor) -> torch.Tensor:
         """``[T, num_codebooks]`` → mono waveform ``[L]``."""
         if codes_TN.ndim != 2:
--- a/sglang_omni/models/higgs_tts/audio_codec.py
+++ b/sglang_omni/models/higgs_tts/audio_codec.py
@@ -260,6 +260,7 @@ class HiggsAudioCodec:
     @torch.no_grad()
     def decode_batch(self, codes_list: list[torch.Tensor]) -> list[torch.Tensor]:
         """Batch-decode variable-length ``[T_i, N]`` tensors into ``[L_i]`` waveforms."""
+
         def _batch_fn(batch_items: list[torch.Tensor]) -> list[torch.Tensor]:
             stacked = torch.stack(batch_items)
             codes_BNT = stacked.transpose(1, 2).to(device=self.device, dtype=torch.long)
--- a/sglang_omni/models/higgs_tts/audio_codec.py
+++ b/sglang_omni/models/higgs_tts/audio_codec.py
@@ -185,49 +185,63 @@ class HiggsAudioCodec:
         codes_BNT = self.model.encode(wav).audio_codes
         return codes_BNT.squeeze(0).transpose(0, 1).to(torch.long).cpu()
 
-    @torch.no_grad()
-    def encode_batch(self, waveforms: list[torch.Tensor]) -> list[torch.Tensor]:
-        """Batch-encode ``[1, 1, L_i]`` float32 waveforms → ``[T_i, N]`` int64 code tensors.
-
-        Only same-length items are batched into a single forward pass.
-        The non-causal acoustic encoder corrupts shorter items when padded to
-        a longer ``L_max``, so mixed-length inputs are encoded per-bucket.
-        Clips shorter than ``SAMPLE_RATE`` samples are zero-padded before
-        bucketing, matching the behaviour of :meth:`encode_reference`.
-        """
-        if not waveforms:
+    def _bucketed_batch(
+        self,
+        items: list[torch.Tensor],
+        *,
+        bucket_key_fn,
+        single_fn,
+        batch_fn,
+        error_label: str,
+    ) -> list[torch.Tensor]:
+        """Run single_fn on singleton buckets and batch_fn on multi-item buckets."""
+        if not items:
             return []
-        if len(waveforms) == 1:
-            return [self.encode_reference(waveforms[0])]
-
-        padded: list[torch.Tensor] = []
-        for w in waveforms:
-            wav = w.to(torch.float32)
-            if wav.shape[-1] < self.SAMPLE_RATE:
-                wav = F.pad(wav, (0, self.SAMPLE_RATE - wav.shape[-1]))
-            padded.append(wav)
+        if len(items) == 1:
+            return [single_fn(items[0])]
 
         buckets: dict[int, list[int]] = {}
-        for i, w in enumerate(padded):
-            buckets.setdefault(w.shape[-1], []).append(i)
+        for i, item in enumerate(items):
+            buckets.setdefault(bucket_key_fn(item), []).append(i)
 
-        results: list[torch.Tensor | None] = [None] * len(waveforms)
+        results: list[torch.Tensor | None] = [None] * len(items)
         for indices in buckets.values():
             if len(indices) == 1:
-                results[indices[0]] = self.encode_reference(padded[indices[0]])
+                results[indices[0]] = single_fn(items[indices[0]])
             else:
-                batch = torch.cat([padded[i] for i in indices])  # [B, 1, L]
-                batch = batch.to(device=self.device, dtype=self._dtype)
-                codes_BNT = self.model.encode(batch).audio_codes  # [B, N, T]
-                for j, idx in enumerate(indices):
-                    results[idx] = codes_BNT[j].transpose(0, 1).to(torch.long).cpu()
+                batch_results = batch_fn([items[i] for i in indices])
+                for idx, result in zip(indices, batch_results):
+                    results[idx] = result
 
         out: list[torch.Tensor] = []
         for i, result in enumerate(results):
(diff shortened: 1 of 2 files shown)


xinlij:
> we shouldn't wire `batch_compute_fn` into the audio encoder. The regression isn't the 2 ms wait: it scales with concurrency (+364 / +976 ms per request at c=16 / 32), so the cost lands downstream on the AR engine, which shares the single GPU with the encoder.
> 
> The encoder isn't the bottleneck, the AR engine is, so making the encoder's GPU work burstier (batching) just steals from it. `encode_batch` is fine to keep as a utility for offline / bulk encoding; just drop the wiring.

Thanks for the explanation!Good to know.  Dropped `_encode_batch` and the `batch_compute_fn` wiring; `HiggsAudioCodec.encode_batch` is kept as a utility.

Ccyest (dismissed):
LGTM — approving.

Ccyest:
**Design update + rebased onto current main.**

After benchmarking, we are **not** wiring `encode_batch` into the serving `audio_encoder` stage. Batching the encoder there *regresses* throughput at c=16/32: the encoder is not the bottleneck (AR decode is ~79% of per-request time), and it shares the GPU with the AR engine — so a batched encode is a GPU burst that stalls the AR bottleneck (the whole running batch), and the bursty feed drops AR occupancy. Measured ~−15–25% throughput, and the added cost scales with concurrency (it lands downstream on AR, not on the 2 ms batch wait). The serving-side "avoid recomputing reference codes" win is already handled by the merged #605 reference-code cache.

So this PR is now scoped to **`HiggsAudioCodec.encode_batch` as a standalone offline / bulk-encoding utility** — length-bucketed so the non-causal encoder doesn't corrupt padded short clips — plus unit tests. It is **not** part of the live pipeline.

Rebased onto current main (now includes #605): the only net change vs main is the `encode_batch` method + its tests; `stages.py` is untouched (keeps main's cache-aware `_encode`).

sglang_omni/models/higgs_tts/audio_codec.py
```diff
@@ -185,6 +185,50 @@ def encode_reference(
...
+            buckets.setdefault(w.shape[-1], []).append(i)
+
+        results: list[torch.Tensor | None] = [None] * len(waveforms)
+        for indices in buckets.values():
+            if len(indices) == 1:
+                results[indices[0]] = self.encode_reference(padded[indices[0]])
+            else:
+                batch = torch.cat([padded[i] for i in indices])  # [B, 1, L]
+                batch = batch.to(device=self.device, dtype=self._dtype)
+                codes_BNT = self.model.encode(batch).audio_codes  # [B, N, T]
```
yxs: The multi-item bucket copies back one item at a time: `for j, idx in enumerate(indices): results[idx] = codes_BNT[j].transpose(0, 1).to(torch.long).cpu()`. Each `.cpu()` is a separate device→host copy with an implicit sync, so a bucket of B items pays B syncs. The `decode_batch` sibling it was copied from (`:272` in the same file) does a **single** batched `.cpu()`: `audio = self.model.decode(codes_BNT).audio_values.cpu()`, then slices on CPU.

Do a single batched `.to(torch.long).cpu()`, then transpose+slice on CPU, fully mirroring `decode_batch`. The CPU-side `transpose` is a zero-cost view.

```python
batch = torch.cat([padded[i] for i in indices])  # [B, 1, L]
batch = batch.to(device=self.device, dtype=self._dtype)
codes_BNT = self.model.encode(batch).audio_codes.to(torch.long).cpu()  # [B, N, T]
for j, idx in enumerate(indices):
    results[idx] = codes_BNT[j].transpose(0, 1)
```

sglang_omni/models/higgs_tts/audio_codec.py
```diff
@@ -185,6 +185,50 @@ def encode_reference(
...
+        a longer ``L_max``, so mixed-length inputs are encoded per-bucket.
+        Clips shorter than ``SAMPLE_RATE`` samples are zero-padded before
+        bucketing, matching the behaviour of :meth:`encode_reference`.
+        """
+        if not waveforms:
+            return []
+        if len(waveforms) == 1:
+            return [self.encode_reference(waveforms[0])]
+
+        padded: list[torch.Tensor] = []
```
yxs: `encode_batch` assumes every input is `[1,1,L]` but never validates it. `encode_reference` starts by running inputs (1-D / 2-D / ndarray) through `_to_mono_3d` to normalise them to `[1,1,L]`; `encode_batch` skips that. So `torch.cat(..., dim=0)` at `:219` only yields `[B,1,L]` when every item happens to be `[1,1,L]` — pass a 1-D or 2-D item (both of which `encode_reference` accepts) and the batch crashes or is silently wrong.

```python
padded: list[torch.Tensor] = []
for w in waveforms:
    wav = _to_mono_3d(w).to(torch.float32)
    if wav.shape[-1] < self.SAMPLE_RATE:
        wav = F.pad(wav, (0, self.SAMPLE_RATE - wav.shape[-1]))
    padded.append(wav)
```

sglang_omni/models/higgs_tts/audio_codec.py
```diff
@@ -185,6 +185,50 @@ def encode_reference(
         codes_BNT = self.model.encode(wav).audio_codes
         return codes_BNT.squeeze(0).transpose(0, 1).to(torch.long).cpu()
 
+    @torch.no_grad()
+    def encode_batch(self, waveforms: list[torch.Tensor]) -> list[torch.Tensor]:
+        """Batch-encode ``[1, 1, L_i]`` float32 waveforms → ``[T_i, N]`` int64 code tensors.
```
yxs: remove this and add a note if needed

sglang_omni/models/higgs_tts/audio_codec.py
xinlij: Fixed — moved `.to(torch.long).cpu()` outside the loop so the whole batch is transferred in a single D2H copy, mirroring `decode_batch`. Thanks for pointing out the comparison.

sglang_omni/models/higgs_tts/audio_codec.py
xinlij: Fixed — added `_to_mono_3d` to normalize inputs before processing, matching the behavior of `encode_reference`.

sglang_omni/models/higgs_tts/audio_codec.py
xinlij: Removed the docstring and replaced with a brief inline comment.

yxs (approved):
looks good

zhaochenyang20 (changes requested):
Make this default on for higgs with the command of:

```bash
sgl-omni serve \
  --port 8000
```

tests/unit_test/higgs_tts/test_pipeline.py
```diff
@@ -868,3 +868,85 @@ def test_decode_batch_bit_exact_with_single_decode() -> None:
     assert call_log == [(2, 10)]
     assert torch.equal(single_a, batch_results[0])
     assert torch.equal(single_b, batch_results[1])
+
+
+# ---------------------------------------------------------------------------
+# HiggsAudioCodec.encode_batch
+# ---------------------------------------------------------------------------
```
zhaochenyang20: delete this

zhaochenyang20: Make the fake encoder produce distinguishable codes tied to each input row, for example by encoding each waveform amplitude or row index into `audio_codes`, then assert that every `encode_batch([a, b, c])` output corresponds to the original input rather than only checking shape. Add batch-input tests for 1-D tensors, 2-D tensors, and `np.ndarray`, ideally comparing `encode_batch` against per-item `encode_reference`, so the normalization regression called out in review stays protected.

zhaochenyang20 (changes requested):
Make this default on for higgs with the command of:

```bash
sgl-omni serve \
  --port 8000
```

tests/unit_test/higgs_tts/test_pipeline.py
xinlij: Done. The fake encoder now fills `codes[i]` with `int(round(batch[i,0,0] * 100))`, making each item's output tied to its waveform amplitude. Added two tests: `test_higgs_audio_codec_encode_batch_matches_encode_reference` (bit-exact comparison against per-item `encode_reference` for 1-D, 2-D, 3-D, and `np.ndarray` inputs) and `test_higgs_audio_codec_encode_batch_input_normalisation` (verifies all input formats normalize identically). Also updated `test_higgs_audio_codec_encode_batch_order_preserved` to use distinct amplitudes so a bucket-swap bug would be caught.

xinlij:
> Make this default on for higgs with the command of:
> 
> ```shell
> sgl-omni serve \
>   --model-path boson-sglang/higgs-audio-v3-TTS-4B-grpo05200410999 \
>   --port 8000
> ```

encode_batch is kept as an offline/bulk utility only. Per our benchmark results, the encoder isn't the bottleneck — the AR engine is (~79% of per-request time), and it shares the GPU with the encoder. Batching the encoder creates a GPU burst that stalls the AR engine, regressing throughput by ~15–25% at c=16/32. The reference-code caching from #605 already handles the repeated-encode cost in serving. We will also add recommended serving parameters for Higgs to the cookbook before finalizing.

zhaochenyang20 (approved):

PR #611, opened 2026-05-29
[Bugfix] ThreadedSimpleScheduler: cap _aborted set to prevent memory leak

vuuihc:
## Motivation

`ThreadedSimpleScheduler.abort()` previously retained every aborted
`request_id` for the lifetime of the stage process. Long-running Higgs TTS and
S2-Pro preprocessing stages could therefore grow `_aborted` without bound.

The abort marker is also load-bearing: it must suppress requests when abort and
work messages arrive out of order, and it must remain effective while a worker
is already running.

## Modifications

Align `ThreadedSimpleScheduler` cleanup with the lifecycle used by
`OmniScheduler`:

- Keep unresolved pre-arrival abort tombstones in deterministic FIFO order,
  bounded at 10000 entries and trimmed to 5000.
- Remove an already queued request directly from `inbox` while preserving other
  queued requests.
- Cancel pending futures and keep running-request aborts in a separate
  lifecycle set that is not subject to FIFO eviction.
- Consume pre-arrival tombstones in `start()` and running abort markers in the
  future completion callback.

This preserves `abort`-before-`new_request` behavior while ensuring queued and
running requests cannot be resurrected by a large storm of newer aborts. The
change is local to `ThreadedSimpleScheduler`; `Stage` and other scheduler types
are unchanged.

## Tests

CPU-only unit tests cover:

- abort arriving before enqueue
- queued request removal without dropping queue peers
- pending future cancellation
- queued and running requests under a 10000+ abort storm
- deterministic FIFO tombstone bounds

Local verification:

- `40 passed, 3 skipped` for the Threaded scheduler and Stage unit suites
- `4 passed, 11 deselected` for streaming abort tests
- pre-commit hooks passed

## Accuracy Test

N/A — scheduler bookkeeping only; no model-side code is changed.

## Benchmark & Profiling

N/A — normal request scheduling is unchanged. Inbox scanning occurs only on
abort.

## Checklist

- [x] Add unit tests.
- [x] Update the PR description.

## CI

CI runs on self-hosted GPU runners and requires the `run-ci` label.


vuuihc:
Rebased onto current `main`, added a couple of CPU unit tests, and cleaned up the commit history (no GPU work).

PTAL when you have a moment — @zhaochenyang20 @GaokaiZhang @Ratish1 @JiaxinD

Mirrors the `SimpleScheduler` `_aborted` cap from #451.

zhaochenyang20:
@vuuihc thanks for raising this bug. Are you in the wechat group? You can add me through 18015766633

tests/unit_test/scheduling/test_threaded_simple_scheduler.py
```diff
@@ -0,0 +1,35 @@
...
+    """``_aborted`` is trimmed once it exceeds 10000 entries (mirror SimpleScheduler)."""
+    scheduler = ThreadedSimpleScheduler(lambda payload: payload, max_concurrency=1)
+
+    for i in range(10000):
+        scheduler.abort(f"req-{i}")
+    assert len(scheduler._aborted) == 10000
+
+    scheduler.abort("req-overflow")
+
+    assert "req-overflow" in scheduler._aborted
```
AkazaAkane: This assertion is flaky because the trim uses list(self._aborted)[:excess], and set iteration order varies with Python hash randomization. The newly added "req-overflow" can itself be included in the evicted entries, so this test passes or fails depending on PYTHONHASHSEED.

I reproduced the exact scenario across 15 hash seeds, and "req-overflow" survived in only 6 runs. This therefore blocks the PR as written. Could we use deterministic eviction such as FIFO?

vuuihc: Verified. This PR originally predates the FIFO implementation in `OmniScheduler`, so it mirrored the existing `SimpleScheduler` behavior. Oldest-first eviction is safer here because a newly aborted request may still be running when `Future.cancel()` returns false, and its abort marker must survive until `_finish()`. I’ll update this scheduler to use the same set + deque approach and adjust the test accordingly.

vuuihc: Updated in `ab86fba9` using the same set + deque FIFO approach as `OmniScheduler`. The test now verifies oldest-first eviction, and it passes across 15 different `PYTHONHASHSEED` values. PTAL.

vuuihc:
@JiaxinD could you please take another look at #611 when you have a moment?

I updated it in `ab86fba9` after @AkazaAkane's comment, switching `_aborted` eviction to set + deque FIFO and updating the test to verify oldest-first eviction. I also checked it across multiple `PYTHONHASHSEED` values.

I see the `run-ci` label is present, but only the cancel workflow is visible on my side. Is there anything else needed for CI/review?

sglang_omni/scheduling/threaded_simple_scheduler.py
```diff
@@ -66,7 +71,12 @@ def stop(self) -> None:
 
     def abort(self, request_id: str) -> None:
         with self._lock:
-            self._aborted.add(request_id)
+            if request_id not in self._aborted:
+                if len(self._aborted) >= _ABORTED_REQUEST_ID_LIMIT:
+                    while len(self._aborted) >= _ABORTED_REQUEST_ID_RETAINED:
```
JiaxinD: Unlike OmniScheduler.abort(), which also synchronously removes the request (FINISH_ABORT on running reqs, pops pending builds, drains inbox), here this evictable marker is the only guard for a running or inbox-queued request, so 5001 newer distinct aborts let its result reach the outbox. The stage runtime's _active_requests filter drops that stale result today and a 5000-abort storm within one task is implausible, so not blocking, just noting the marker is load-bearing here in a way it is not in the scheduler this mirrors.

JiaxinD (approved):

sglang_omni/scheduling/threaded_simple_scheduler.py
```diff
@@ -66,11 +73,48 @@ def stop(self) -> None:
...
+        self._aborted.add(request_id)
+        self._aborted_order.append(request_id)
+
+    def _discard_aborted_request_id(self, request_id: str) -> None:
+        if request_id not in self._aborted:
+            return
+        self._aborted.discard(request_id)
+        self._aborted_order.remove(request_id)
+
+    def _drain_inbox_for_request(self, request_id: str) -> bool:
```
JiaxinD: This drains and rebuilds the entire inbox while holding `self._lock`, so abort cost grows with queue depth and dispatch, `_finish` and `_wait_for_capacity` all stall for the duration. Measured on this head (3 rounds, 6000 deep inbox for the dispatch rows):

| case | result |
| --- | ---: |
| abort, empty inbox | 1.0 us |
| abort, 5000 queued | 4149 us |
| dispatch, no abort storm | ~2550/s |
| dispatch, concurrent abort storm | ~176/s |

Doing the drain outside the lock, or falling back to the marker when the inbox is deep, would keep the fix without the cliff.

sglang_omni/scheduling/threaded_simple_scheduler.py
```diff
@@ -66,11 +73,48 @@ def stop(self) -> None:
...
+                oldest = self._aborted_order.popleft()
+                self._aborted.discard(oldest)
+        self._aborted.add(request_id)
+        self._aborted_order.append(request_id)
+
+    def _discard_aborted_request_id(self, request_id: str) -> None:
+        if request_id not in self._aborted:
+            return
+        self._aborted.discard(request_id)
+        self._aborted_order.remove(request_id)
```
JiaxinD: `deque.remove` is O(n) and runs under the lock, so retiring 10k markers costs 0.9 ms when requests arrive in abort order and 294 ms when they arrive reversed (measured). A single insertion ordered `dict[str, None]` gives O(1) delete plus O(1) FIFO pop and drops the set/deque sync question entirely.

sglang_omni/scheduling/threaded_simple_scheduler.py
```diff
@@ -52,6 +58,7 @@ def start(self) -> None:
                     continue
                 with self._lock:
                     if msg.request_id in self._aborted:
+                        self._discard_aborted_request_id(msg.request_id)
                         continue
                     future = self._executor.submit(self._run_one, msg.data)
                     self._pending[msg.request_id] = future
```
JiaxinD: If two futures ever share a request_id, `abort()` marks the id and the older future's `_finish` consumes that marker, so the newer future's result is still emitted after the abort (reproduced against this class). The coordinator pops `_requests[id]` at abort time, so a client that reuses an id right after aborting can overlap the still running future.

tests/unit_test/scheduling/test_threaded_simple_scheduler.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+    scheduler.inbox.put(
+        IncomingMessage(
+            request_id="req-queued",
+            type="new_request",
+            data="must-not-run",
+        )
+    )
+    scheduler.abort("req-queued")
+    for i in range(10001):
```
JiaxinD: The storm test removes its only queued message first, so all 10001 aborts here scan an empty inbox and never exercise the populated inbox path. Queueing a few thousand peers before the storm would cover the expensive case.

JiaxinD (commented):
The lifecycle fix checks out: `_running_aborted` is bounded on both the cancel and the complete path, the set and the deque stay in exact sync, and 5 of the 6 new tests fail against the merge base. Note that CI has never run on this head, all four workflows sit at `action_required` because fork PRs from first time contributors need a maintainer to approve the run.

sglang_omni/scheduling/threaded_simple_scheduler.py
vuuihc: Fixed in 935b86a5. The inbox is no longer drained or rebuilt on abort. An aborted message stays queued and the dispatch loop drops it when it surfaces, so abort() is O(1) at any depth — 0.24 us at 0 / 128 / 6000 deep here (was 4149 us at 5000 deep). Throttled abort rates (1k–10k/s) show no throughput loss vs. the no-abort baseline; the ~176/s figure came from an unthrottled abort loop (~3M calls/s) contending the lock, which does not reflect a real abort rate.

sglang_omni/scheduling/threaded_simple_scheduler.py
vuuihc: Done in 935b86a5. The set + deque pair is gone. Speculative aborts now use a single insertion-ordered dict[str, None] as you suggested — O(1) membership, O(1) delete, O(1) oldest-first pop — and the set/deque sync question no longer exists.

sglang_omni/scheduling/threaded_simple_scheduler.py
vuuihc: Fixed in 935b86a5. Running aborts are now keyed by Future identity, not by request_id, so an older future's _finish only clears its own marker and cannot suppress a newer future sharing the id. A per-id count of un-reaped aborted futures also keeps a repeat abort of a running request idempotent, so it won't leave a tombstone that suppresses a genuine later reuse. Covered by test_aborted_future_does_not_suppress_another_future and test_repeated_running_abort_allows_reuse_after_quiescence.

tests/unit_test/scheduling/test_threaded_simple_scheduler.py
vuuihc: Addressed in 935b86a5. test_queued_aborts_past_cap_are_not_evicted now queues _ABORTED_REQUEST_ID_LIMIT + 100 real messages, aborts each, then drives the dispatch loop and asserts none execute — exercising the populated-inbox path rather than an empty one.

sglang_omni/scheduling/threaded_simple_scheduler.py
vuuihc: Addressed in 935b86a5. Tombstones are now split by role: _queued_aborts covers a request that is still queued or claimed for dispatch and is never subject to capacity eviction (inbox depth already bounds it), while only _speculative_aborts (unseen requests, which may never arrive) is capped. So a storm of newer distinct aborts can no longer evict the marker guarding a still-queued request. The load-bearing marker you flagged is now the un-evictable bucket.

vuuihc:
@JiaxinD thanks for the detailed pass. Pushed `935b86a5` addressing all five threads (replied inline):

- **L102 inbox rebuild** — removed; aborted messages stay queued and are dropped on dispatch. `abort()` is now O(1) at any depth (0.24 us at 0/128/6000 deep). The ~176/s figure traced to an unthrottled abort loop (~3M/s) contending the lock; throttled 1k–10k/s shows no throughput loss.
- **L100 deque.remove O(n)** — replaced with a single insertion-ordered `dict[str, None]`.
- **L64 shared request_id** — running aborts now keyed by Future identity; a per-id count keeps repeat aborts idempotent without suppressing a genuine reuse.
- **L76 load-bearing marker** — tombstones split into `_queued_aborts` (never evicted) vs `_speculative_aborts` (the only capped bucket), so an abort storm can't evict a still-queued request's guard.
- **L113 test** — the storm test now queues 10k+ real messages first, exercising the populated-inbox path.

Also closed the `inbox.get()`→Future-registration window via `_CountingInbox` claim tracking under the queue mutex, so an abort in that gap is no longer lost. New tests cover the claim window, over-cap queued storm, id reuse after quiescence, and non-new_request ids leaving no marker.

One thing outside my control: CI has still never run on any head — the four workflows sit at `action_required`. Given how subtle this concurrency path is, a real CI run before merge would be worth it. Could you approve the workflow run when you have a moment? PTAL.

sglang_omni/scheduling/threaded_simple_scheduler.py
```diff
@@ -65,12 +132,52 @@ def stop(self) -> None:
...
-            self._aborted.add(request_id)
             future = self._pending.pop(request_id, None)
+            if future is not None:
+                self._aborted_futures.add(future)
+                self._running_aborted_ids[request_id] = (
+                    self._running_aborted_ids.get(request_id, 0) + 1
+                )
+            elif self.inbox.is_reachable(request_id):
+                self._queued_aborts.add(request_id)
+            elif request_id in self._running_aborted_ids:
```
JiaxinD: This branch silently drops the abort when the id is reused: with an older aborted future still running, `abort()` for a genuinely new request under the same id records no tombstone, so that request runs when it arrives (reproduced on this head; 91587ac4 suppressed it). The comment holds only when the id still refers to the same request, and the coordinator releases the id at abort time, so a client can resubmit immediately.

sglang_omni/scheduling/threaded_simple_scheduler.py
```diff
@@ -16,25 +16,88 @@
...
+        self._claimed_counts: dict[str, int] = {}
+
+    def _put(self, item: IncomingMessage) -> None:
+        super()._put(item)
+        if item.type != "new_request":
+            return
+        request_id = item.request_id
+        self._request_counts[request_id] = self._request_counts.get(request_id, 0) + 1
+
+    def _get(self) -> IncomingMessage:
```
JiaxinD, line 38: `_CountingInbox` now carries an implicit invariant: every `_get` must be paired with a `release_claim`. Anything that dequeues without dispatching leaves a permanent claim, which makes `is_reachable` always true and routes every later abort into the uncapped `_queued_aborts`, reintroducing the growth this PR fixes; I reproduced it with `Queue.shutdown(immediate=True)` on 3.13. Nothing calls that today, so an assertion or a note on the invariant would be enough.

tests/unit_test/scheduling/test_threaded_simple_scheduler.py
```diff
@@ -0,0 +1,245 @@
...
+
+            release.set()
+            _wait_until(lambda: not scheduler._running_aborted_ids)
+            scheduler.inbox.put(_request("reused", "new"))
+            assert scheduler.outbox.get(timeout=5.0).data == "new"
+        finally:
+            release.set()
+
+
+def test_aborted_future_does_not_suppress_another_future() -> None:
```
JiaxinD: This does not reproduce the reused-id ordering it is meant to guard: the older future here is cancellable, so its callback runs inline and clears the marker before the newer result lands, and the same sequence passes against 91587ac4. The case that failed before is a stale older future that is already running, with the newer future in `_pending`, aborting the newer one and completing the stale one first.

JiaxinD (commented):
All four items from the last round are fixed, and the measured A/B is clean. Abort cost is now flat at 0.37 us from inbox depth 0 to 5000 (was 1.0 us to 4149 us), and marker retirement is order independent at 0.62 ms forward vs 0.69 ms reversed for 10k (was 0.9 ms vs 294 ms). More usefully, at a bounded abort rate with a 6000 deep inbox:

| aborts/s | 91587ac4 dispatch | this head |
| ---: | ---: | ---: |
| 0 | 2608/s | 2616/s |
| 100 | 1913/s | n/a |
| 1000 | 331/s | 2664/s |
| 10000 | 282/s | 2732/s |

So the old version lost 27% of dispatch throughput at only 100 aborts/s, and this head shows no measurable impact at 10k/s; only an unthrottled tight loop (~390k/s) still degrades it, which is GIL contention rather than this code. Capping only the speculative markers while leaving queued and claimed ones uncapped is the right split, it targets exactly the unbounded vector. 11 tests pass locally. Three small things below.

vuuihc:
Thanks for the follow-up review. I addressed the latest comments and lint issues, then squashed the changes into one commit.

Main updates:
- Separated bounded speculative abort markers from non-evictable queued/claimed aborts.
- Made abort promotion and enqueue atomic under a consistent lock order.
- Added the missing abort check for stage dispatch, including TP followers.
- Added deterministic regression coverage for the enqueue, dispatch, and stale-callback races.

One explicit limitation: request IDs must not be reused while an earlier lifecycle may still exist in the pipeline. Ambiguous repeated aborts intentionally fail closed and may suppress a new request using the same ID. Fully supporting overlapping ID reuse would require a generation/lifecycle identity, which is outside this PR.

Validation:
- 13 focused tests passed, including 20 repeated runs.
- All pre-commit hooks and formatting checks passed.

Could you please take another look when you have time? Thanks!

sglang_omni/scheduling/threaded_simple_scheduler.py
JiaxinD: This invariant is still implicit on current head 6878097a. Since _get() increments _claimed_counts, please either document that every dequeued new_request must be paired with release_claim() and that immediate-drain paths must not use _get(), or make shutdown/drain preserve the counts. A short invariant note or assertion is enough; after that, this thread is clear.

JiaxinD:
Thanks for rebasing and for continuing the lifecycle and race fixes beyond the bounded cleanup that landed through #1462. I have updated #1462's description and discussion to credit #611.

I reviewed current head 6878097a, and the 13 focused scheduler tests pass locally. I rechecked the _CountingInbox concern against the supported runtime: the project pins Python below 3.13, and no production path calls Queue.shutdown() on this inbox. It is a future-maintenance note rather than a current production issue, so no code change is required. I have approved the four workflow runs on this exact head.

sglang_omni/scheduling/threaded_simple_scheduler.py
JiaxinD: Rechecked against the supported runtime and current call graph: pyproject.toml pins Python below 3.13, and no production path calls Queue.shutdown() on this inbox. The immediate-drain case is therefore a future-maintenance invariant, not a current production trigger. This is non-blocking; no code change is required for this PR, and I have approved CI on head 6878097a.

vuuihc: > *This was generated by AI during triage.*

Tracked the future-maintenance invariant in #1680. Keeping #611 unchanged so the current CI run is not restarted; this remains outside the current supported runtime/call graph as discussed.

Issue #600, opened 2026-05-28
[Bug] OpenAI /v1/chat/completions text-only streaming emits a single chunk (no per-token TPOT)


edwingao28:
Discussed with @yuan-luo  
For Ming-Omni served through `/v1/chat/completions` with `stream=true` and `modalities=["text"]`, the OpenAI stream does not emit incremental text deltas during thinker generation. The client receives one non-empty `delta.content` chunk only after generation finishes, followed by the finish chunk and `[DONE]`.

This makes stream TTFT effectively equal to end-to-end latency, and TPOT cannot be measured from the OpenAI text stream.

Code path on main:

- `ming_omni/config.py`: the text pipeline is `thinker -> decode`, with `decode` as the only terminal stage.
- `ming_omni/bootstrap.py`: `make_thinker_stream_output_builder()` is only installed when `enable_streaming_tts=True`; the normal text pipeline does not install it.
- Even when enabled for streaming TTS, the thinker stream callback sends deltas only to `segmenter`, not to the OpenAI text channel. The code comment says streaming text deltas require either a stream-aware decode stage or a dedicated text fan-out stage.
- `ming_omni/stages.py:create_decode_executor()` is a `SimpleScheduler` and decodes the final full `output_ids` list after thinker completion.
- `serve/openai_api.py:_chat_stream()` only forwards chunks produced by `client.completion_stream()`, so it cannot produce token deltas if the Ming pipeline only sends a final decode result.

Expected behavior: `stream=true` should emit incremental `delta.content` events during generation, ideally per-token or small batches, similar to Qwen3-Omni's stream-aware decode path.

Observed behavior: exactly one non-empty text chunk containing the full assistant answer.

In a corrected GSM8K sweep with the same prompts, Ming-Omni text generation has similar end-to-end latency/QPS to vLLM-Omni at comparable concurrency. The issue is not total latency; it is that the OpenAI text stream exposes no incremental token/chunk timing, so stream TTFT/TPOT are not meaningful for Ming-Omni.

```

  ┌─────────────────────────────┬─────────────┬────────────┬─────────────┬────────────┬────────────────┬─────────────────┐
  │           backend           │ TTFT (mean) │ TTFT (p95) │ TPOT (mean) │ TPOT (p95) │ latency (mean) │ gen_toks (mean) │
  ├─────────────────────────────┼─────────────┼────────────┼─────────────┼────────────┼────────────────┼─────────────────┤
[...]

PR #613, opened 2026-05-29
[WIP] Fix Ming text streaming TTFT

Hangzhi:
**For Reviewers** This PR is a first draft. Polish WIP. 

## Summary
Fixes Ming-Omni text streaming so OpenAI `stream=true` receives incremental `delta.content` chunks instead of one final full-text chunk.

Changes:
- route Ming thinker text deltas to `decode`
- add a stream-aware Ming decode scheduler that forwards text deltas to the coordinator
- make the final streamed decode result emit only any missing text suffix, avoiding duplication while preserving tail text
- preserve Ming/Bailing `head_dim` and `rotary_dim` config values when present
- align Ming attention tensor shape handling with the Bailing attention contract

## Measurement

Issue: https://github.com/sgl-project/sglang-omni/issues/600

See [Measurement](#measurement) below for the commands, methodology, and results.

OpenAI-visible TTFT benchmark, deterministic 8-token schedule:

| backend/source | TTFT mean | TTFT p95 | E2E mean | content chunks |
|---|---:|---:|---:|---:|
| sglang-omni main | 0.4519 s | 0.4521 s | 0.4520 s | 1 |
| sglang-omni current fix | 0.1006 s | 0.1007 s | 0.4540 s | 8 |

Source-backed synthetic/profile comparison against the local vLLM-Omni repo, using `prefill_s=0.0398`, `token_s=0.0102`:

| backend/source | TTFT mean | TTFT p95 | E2E mean | content chunks |
|---|---:|---:|---:|---:|
| sglang-omni current fix | 0.0515 s | 0.0516 s | 0.1310 s | 8 |
| sglang-omni main | 0.1296 s | 0.1298 s | 0.1297 s | 1 |

Interpretation: main waits until all generated text is available before emitting the first non-empty OpenAI stream chunk. The fix changes Ming text streaming to the same OpenAI-visible shape as vLLM-Omni source semantics: TTFT is prefill plus the first generated token, not full generation latency.

Note: `vllm` is not installed in this environment. The vLLM-Omni comparison above verifies the local vLLM-Omni source contracts (`RequestOutputKind.DELTA` for streaming and OpenAI `delta.content` SSE emission) and applies the reported vLLM-like timing profile. It is not a full vLLM-Omni server benchmark.

## Verification

```bash
timeout 90s /usr/bin/python3 -m pytest -q \
  tests/unit_test/ming_omni/test_pipeline.py \
  tests/unit_test/ming_omni/test_streaming_speech_config.py \
  tests/unit_test/ming_omni/test_streaming_e2e_glue.py \
  tests/unit_test/ming_omni/test_thinker.py

timeout 30s /usr/bin/python3 -m pytest -q \
  tests/unit_test/serve/test_openai_api.py::test_chat_stream_failure_closes_without_done_sentinel

timeout 30s /usr/bin/python3 -m pytest -q \
  tests/unit_test/qwen3_omni/test_streaming.py::test_client_completion_stream_does_not_duplicate_full_text \
  tests/unit_test/qwen3_omni/test_streaming.py::test_client_completion_stream_non_streaming_keeps_full_text

/usr/bin/python3 -m py_compile \
  sglang_omni/models/ming_omni/bootstrap.py \
  sglang_omni/models/ming_omni/config.py \
  sglang_omni/models/ming_omni/stages.py \
  sglang_omni/models/ming_omni/thinker.py \
  sglang_omni/models/ming_omni/configuration.py

git diff --check
```

Observed:
- `52 passed in 4.36s`
- `1 passed in 2.81s`
- `2 passed, 2 warnings in 10.37s`
- `py_compile` passed
- `git diff --check` passed

### DIFF
--- a/sglang_omni/models/ming_omni/bootstrap.py
+++ b/sglang_omni/models/ming_omni/bootstrap.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+from collections.abc import Sequence
 from typing import Any
 
 from sglang_omni.models.ming_omni.pipeline.sampling import build_ming_sampling_params
@@ -19,6 +20,8 @@ def create_thinker_scheduler(
     tp_size: int = 1,
     nccl_port: int | None = None,
     enable_streaming_tts: bool = False,
+    enable_streaming_text: bool = False,
+    stream_text_targets: Sequence[str] | None = None,
 ):
     if tp_size < 1:
         raise ValueError(f"tp_size must be >= 1, got {tp_size}")
@@ -83,11 +86,18 @@ def create_thinker_scheduler(
     )
 
     stream_output_builder = None
-    if enable_streaming_tts:
+    if enable_streaming_tts or enable_streaming_text:
         eos_token_id = getattr(tokenizer, "eos_token_id", None)
+        targets = list(stream_text_targets or ())
+        if enable_streaming_text and "decode" not in targets:
+            targets.append("decode")
+        if enable_streaming_tts and "segmenter" not in targets:
+            targets.append("segmenter")
         stream_output_builder = make_thinker_stream_output_builder(
             tokenizer=tokenizer,
             eos_token_id=eos_token_id,
+            target_stages=targets,
+            client_stream_target_stages=("decode",),
         )
 
     return OmniScheduler(
@@ -253,8 +263,10 @@ def make_thinker_stream_output_builder(
     tokenizer: Any,
     eos_token_id: int | None,
     target_stage: str = "segmenter",
+    target_stages: Sequence[str] | None = None,
+    client_stream_target_stages: Sequence[str] = (),
 ):
-    """Build a per-token stream callback that emits text deltas to the segmenter.
+    """Build a per-token stream callback that emits text deltas to target stages.
 
     OmniScheduler calls this on every model step with the freshly generated
     token id. We maintain per-request running output_ids on ``req`` so we can
@@ -266,6 +278,28 @@ def make_thinker_stream_output_builder(
     import torch
 
     from sglang_omni.scheduling.messages import OutgoingMessage
+
+    targets = list(target_stages or [target_stage])
+    client_stream_targets = set(client_stream_target_stages)
+
+    def _client_streaming_requested(req_data: Any) -> bool:
+        payload = getattr(req_data, "stage_payload", None)
+        request = getattr(payload, "request", None)
+        params = getattr(request, "params", None) or {}
+        return bool(params.get("stream", False))
+
+    def _text_output_requested(req_data: Any) -> bool:
+        payload = getattr(req_data, "stage_payload", None)
+        request = getattr(payload, "request", None)
+        metadata = getattr(request, "metadata", None) or {}
+        modalities = metadata.get("output_modalities")
+        if modalities is None:
+            return True
+        if isinstance(modalities, str):
+            return modalities == "text"
+        if isinstance(modalities, (list, tuple, set)):
+            return "text" in {str(modality) for modality in modalities}
+        return True
 
     def _build_stream_output(request_id, req_data, req_output):
         req = getattr(req_data, "req", None)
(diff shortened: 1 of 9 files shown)


edwingao28:
@Hangzhi please fix lint

sglang_omni/models/ming_omni/thinker.py
```diff
@@ -102,9 +102,8 @@ def __init__(
             self.q_norm = RMSNorm(self.head_dim, eps=config.rms_norm_eps)
             self.k_norm = RMSNorm(self.head_dim, eps=config.rms_norm_eps)
 
-        # RoPE - using partial rotary factor
         self.rotary_emb = get_rope(
-            self.rotary_dim,
+            self.head_dim,
```
edwingao28, line 106: **Attention/RoPE change is correct but unproven.** `get_rope`'s first positional arg is head_size, so `self.head_dim` is the right value here — `main` passed `self.rotary_dim`, which is wrong whenever `rotary_dim < head_dim` (partial rotary). Removing the manual reshape + split is also correct: `apply_qk_norm` reshapes by `head_dim` internally and `RotaryEmbedding` applies partial rotary via `rotary_dim`. So this is effectively a latent-bug fix — but it changes attention math for **every** Ming request (text + speech) and is guarded only by a source-string test (`test_ming_attention_keeps_qkv_flat_for_radix_attention` greps the source, it does not check numerics).

1. Please add a numerical forward-parity test (e.g. `cos ≥ 0.999` vs HF reference on a partial-rotary config), not a source-string assertion
2. Please either split the attention/RoPE/config-schema changes into a separate PR, or explain why they are required for the streaming fix in this PR.

sglang_omni/models/ming_omni/stages.py
```diff
@@ -101,6 +110,255 @@ def _attach_decode_final_metadata(
...
+    ) -> None:
+        from sglang_omni.scheduling.messages import IncomingMessage, OutgoingMessage
+
+        self.inbox: _queue_mod.Queue[IncomingMessage] = _queue_mod.Queue()
+        self.outbox: _queue_mod.Queue[OutgoingMessage] = _queue_mod.Queue()
+        self._tokenizer = tokenizer
+        self._eos_token_id = eos_token_id
+        self.stage_name = stage_name
+        self._running = False
+        self._state: dict[str, _StreamingDecodeState] = {}
```
edwingao28, line 215: `MingStreamingDecodeScheduler` robustness — two gaps:

1. The per-message `try`/`except` in `start()` correctly isolates per-request failures (matches the Qwen3-Omni scheduler contract), but there is no test for it. Add one that makes `_on_stream_chunk`/`_finalize` raise and asserts the scheduler thread keeps processing later requests.
2. `self._state` has no bounded eviction (only `_done_seen` does). A request that streams chunks then is aborted before `new_request`/`stream_done` reaches decode leaks its `_state` entry forever → unbounded memory under failure traffic. Add a TTL/bounded eviction, or guarantee `abort()` on every termination path.

sglang_omni/models/ming_omni/stages.py
```diff
@@ -101,6 +110,255 @@ def _attach_decode_final_metadata(
...
+                    return bytes(values).decode("utf-8", errors="ignore")
+                token_ids = [
+                    int(token_id) for token_id in data.detach().cpu().flatten()
+                ]
+                if token_ids:
+                    return self._tokenizer.decode(
+                        token_ids,
+                        skip_special_tokens=True,
+                    )
+        except Exception:
```
edwingao28, line 356: `except Exception: raise` is a no-op — it catches every exception only to immediately re-raise, identical to having no `try`/`except`. If the intent is to tolerate a missing torch, narrow to `except ImportError:` around just the `import torch` and `return str(data)`; otherwise drop the `try`/`except` entirely.

sglang_omni/models/ming_omni/bootstrap.py
```diff
@@ -253,8 +263,10 @@ def make_thinker_stream_output_builder(
     tokenizer: Any,
     eos_token_id: int | None,
     target_stage: str = "segmenter",
+    target_stages: Sequence[str] | None = None,
```
edwingao28, line 266: Dual public params `target_stage` (singular, default `"segmenter"`) + `target_stages` (plural) is a foot-gun: `target_stage` now only survives as a fallback when `target_stages is None`, and every internal call site passes `target_stages`. Drop `target_stage` (or wrap it into `target_stages` with a `DeprecationWarning`).

edwingao28 (changes requested):

edwingao28:
Please resolve file conflict and remove WIP in the title @Hangzhi, in general this fix PR looks good to me and I left some review comments.

yuan-luo:
Please remove the vllm-omni benchmark in the description.

edwingao28:
Thanks for the work here @Hangzhi . We decided to move forward with #656 as the merge path because it keeps the text streaming architecture a bit cleaner: thinker emits raw token ids and decode owns client-visible detokenization.

That said, #613 was very helpful in identifying the correct speech/streaming-speech topology. In particular, the idea that speech thinkers must stream completion to decode, and that decode must accept stream signals before payload, directly informed the final #656 changes, which I added you as co author.

I really appreciate the investigation and design work here.

Hangzhi:
@edwingao28 sounds good. Let me close this PR in favor of https://github.com/sgl-project/sglang-omni/pull/656

Sorry that I did get the comments resolved earlier.

PR #617, opened 2026-05-30
perf(qwen3-omni): enable talker partial-start by default + --talker-partial-start serve flag

Hayden727:
## Summary

Enable Qwen3-Omni talker **partial-start** by default on the 2-GPU speech pipeline, and expose it as a runtime toggle. Partial-start lets the talker begin audio generation after `partial_start_min_chunks` (5) thinker text chunks instead of waiting for the *full* thinker text.

This came out of a profiling pass on the default 2-GPU speech topology (thinker GPU0 / talker_ar + code2wav GPU1) on H200 with SeedTTS streaming. Per-stage event traces showed the talker's `input → first-codec` latency dominated TTFC (~749 ms at c=8) because the talker idled until the thinker finished its text (~400 ms residency). Partial-start removes that wait and, because the talker (GPU1) now overlaps the still-running thinker (GPU0), improves both latency and throughput.

## Changes

- `models/qwen3_omni/config.py`: `_talker_stage` `enable_partial_start` default `False → True`.
- `cli/serve.py`: add `--talker-partial-start {default,on,off}` (mirrors `--talker-cuda-graph`), plumbed to the `talker_ar` factory arg via `apply_partial_start_cli_overrides`; invalid modes fail fast with `typer.BadParameter`. This restores a per-run off-switch (`--talker-partial-start off`).
- `examples/run_qwen3_omni_speech_server.py`: `--enable-partial-start` becomes a `BooleanOptionalAction` (default on, `--no-enable-partial-start` to disable) and always applies the resolved value.
- Tests: serve-CLI on/off/default/invalid coverage, new config default, example-launcher default-on + disable.

## Performance (SeedTTS streaming, 50 samples, bf16, identical workload)

| metric | c=1 Δ | c=8 Δ |
|---|---|---|
| TTFC mean | **−20.1%** | **−34.5%** |
| TTFC p95 | −60.6% | −41.1% |
| RTF mean | −9.8% | −18.0% |
| RTF p95 | −12.3% | −28.2% (c=8 p95 1.02→0.73, now < realtime) |
| throughput | +15.7% | +20.6% |
| latency mean | −13.7% | −17.3% |
| ITL | −3.0% | +2.0% (flat, expected) |

## Firm-up of the #475 `audio_duration` concern

#475 shipped partial-start default-off with the note that `audio_duration` "must firm up before default flips." Verified with a paired per-sample experiment (same build, `--talker-partial-start off|on`, n=50; talker seed derived from request_id):

- **Noise floor (OFF vs OFF):** per-sample audio_duration varies **12.88%** run-to-run (temperature-0.7 sampling).
- **Effect (ON vs OFF):** **13.78%** — statistically identical to the noise floor; aggregate +3% is ~1.4 SE (not significant); 1/50 samples outside the noise envelope.
- **WER (Whisper, same 50):** OFF 1.95% → ON **1.24%**, 0 catastrophic (>50%) samples either way.

Conclusion: the audio_duration shift is sampling noise, not a systematic partial-start effect; intelligibility is intact. The default flip is safe, with `--talker-partial-start off` as an escape hatch.

## Test plan

- [x] `pytest tests/unit_test/qwen3_omni/test_cli.py test_example_launcher.py test_talker.py` — **104 passed** (in-container, full env).
- [x] GPU validation on ion8-h200-omni: c=1 & c=8 SeedTTS before/after; paired audio_duration + WER firm-up.
- [x] CI.

## Scope note

This PR is the validated, low-risk win. Two further profiler findings were investigated and intentionally deferred:
- **Talker launch-bound** (34% GPU occupancy, ~340K kernel launches): the CUDA-graph-replay / overlap-schedule levers require decoupling the talker feedback loop and carry real audio-correctness risk (~2–3 days, ~5–10% gain) — better as a separate research PR.
- **code2wav vocoder micro-launches**: low ROI (variable-length input, ~382 ms absolute), skipped.

### DIFF
--- a/sglang_omni/cli/serve.py
+++ b/sglang_omni/cli/serve.py
@@ -637,7 +637,10 @@ def apply_partial_start_cli_overrides(
         return pipeline_config
     _apply_stage_factory_args_override(
         pipeline_config,
-        stage_name="talker_ar",
+        stage_name=_resolve_talker_stage(
+            pipeline_config,
+            flag_name="--talker-partial-start",
+        ),
         updates={"enable_partial_start": mode == "on"},
         reason=f"talker partial-start mode to {mode!r}",
         flag_name="--talker-partial-start",
--- a/sglang_omni/cli/serve.py
+++ b/sglang_omni/cli/serve.py
@@ -17,6 +17,9 @@ _QWEN_COLOCATED_CONFIG_CLASS = "Qwen3OmniSpeechColocatedPipelineConfig"
 _HIGGS_ASYNC_DECODE_FACTORY = (
     "sglang_omni.models.higgs_tts.stages.create_sglang_tts_engine_executor"
 )
+_QWEN_PARTIAL_START_TALKER_FACTORY = (
+    "sglang_omni.models.qwen3_omni.stages.create_talker_ar_executor_from_config"
+)
 
 
 def launch_server(*args: object, **kwargs: object) -> object:
@@ -635,12 +638,24 @@ def apply_partial_start_cli_overrides(
     mode = _normalize_stage_toggle_mode("talker_partial_start", talker_partial_start)
     if mode == "default":
         return pipeline_config
+    stage_name = _resolve_talker_stage(
+        pipeline_config,
+        flag_name="--talker-partial-start",
+    )
+    matching_stages = _find_matching_stages(
+        pipeline_config,
+        stage_name=stage_name,
+        reason=f"talker partial-start mode to {mode!r}",
+    )
+    for stage in matching_stages:
+        if stage.factory != _QWEN_PARTIAL_START_TALKER_FACTORY:
+            raise typer.BadParameter(
+                "--talker-partial-start currently supports only Qwen3-Omni "
+                f"talker; stage {stage.name!r} uses factory {stage.factory!r}"
+            )
     _apply_stage_factory_args_override(
         pipeline_config,
-        stage_name=_resolve_talker_stage(
-            pipeline_config,
-            flag_name="--talker-partial-start",
-        ),
+        stage_name=stage_name,
         updates={"enable_partial_start": mode == "on"},
         reason=f"talker partial-start mode to {mode!r}",
         flag_name="--talker-partial-start",
--- a/sglang_omni/cli/serve.py
+++ b/sglang_omni/cli/serve.py
@@ -627,6 +627,24 @@ def apply_cuda_graph_cli_overrides(
     return pipeline_config
 
 
+def apply_partial_start_cli_overrides(
+    pipeline_config: PipelineConfig,
+    *,
+    talker_partial_start: str,
+) -> PipelineConfig:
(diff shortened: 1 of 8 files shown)


sglang_omni/cli/serve.py
```diff
@@ -627,6 +627,24 @@ def apply_cuda_graph_cli_overrides(
     return pipeline_config
 
 
+def apply_partial_start_cli_overrides(
```
yxs, line 633: the current hardcode makes `_find_matching_stages` raise "`Stage 'talker_ar' not found in pipeline; cannot set talker partial-start mode to 'on'`" — leaking the internal stage name to the user and diverging from the cuda-graph path's error style.
Consider reuse the existing `_resolve_talker_stage`

Hayden727:
@yxs good catch, fixed in 984dfe6.

`apply_partial_start_cli_overrides` now resolves the stage via the existing `_resolve_talker_stage` instead of hardcoding `"talker_ar"`. On a config without a talker (e.g. text-only `Qwen3OmniPipelineConfig`) it now raises the stable `--talker-partial-start is not supported by Qwen3OmniPipelineConfig` message, matching the `--talker-cuda-graph` style and no longer leaking the internal stage name.

Note I used `_resolve_talker_stage` (the talker *placement* role map) rather than `_resolve_talker_sglang_stage` (the ServerArgs variant the cuda-graph path uses), since `enable_partial_start` is a stage factory arg, not a SGLang ServerArgs override — both resolve to `talker_ar` for Qwen3-Omni today. Added `test_partial_start_cli_rejects_unsupported_config_with_stable_message`; in-container suite green (13 partial_start/talker/cuda_graph tests pass).

sglang_omni/cli/serve.py
```diff
@@ -627,6 +627,27 @@ def apply_cuda_graph_cli_overrides(
...
+    _apply_stage_factory_args_override(
+        pipeline_config,
+        stage_name=_resolve_talker_stage(
+            pipeline_config,
+            flag_name="--talker-partial-start",
+        ),
+        updates={"enable_partial_start": mode == "on"},
+        reason=f"talker partial-start mode to {mode!r}",
+        flag_name="--talker-partial-start",
+    )
```
JingwenGu0829, line 662: Should we make `--talker-partial-start` reject non-Qwen talker configs before mutating `factory_args`? Right now it uses the generic `talker_role_to_stage()` map, which Ming, and future potential talker models' speech configs also expose for placement. Passing `--talker-partial-start on` on those configs injects `enable_partial_start`, but their talker factories do not accept that kwarg, so the server can fail later in the worker instead of failing fast in CLI validation. I would suggest gating this to the Qwen talker factory, or adding a dedicated partial-start capability map.

JingwenGu0829 (commented):
Looks like a good direction! Noted some minor issues which may be worth fixing.

zhaochenyang20 (approved):

Issue #601, opened 2026-05-28
[Tracking] Production Level TTS Serving


zhaochenyang20:
## Summary

Our TTS serving currently covers the core synthesis path but is missing several APIs we need for production-grade deployments. This issue tracks four related workstreams:

1. Complete the production-level TTS API surface.
2. Run a detailed validation pass on Higgs against the completed API set.
3. Migrate the CI TTS baseline from S2 Pro to Higgs.
4. Raise our overall test intensity from "CI-sized sweep" to "production-grade stress."

## Motivation

Production users hit a TTS endpoint in ways our current CI fundamentally does not exercise. They upload arbitrary voices, hold WebSockets open with malformed text, batch hundreds of requests with mixed languages, send adversarial parameter combinations, and generally find every corner case that a fixed sweep cannot reach. Our existing TTS CI sweep (~1088 requests) is generous for a green-light signal but not representative of production traffic. We need both the API breadth and the test depth to be ready for that.

### 1. Production-level API surface

The following endpoints/parameters need to be implemented or completed. Items that are already partially supported should be audited for full parameter coverage.

**`POST /v1/audio/speech`** — single-shot synthesis
- OpenAI-standard params: `input`, `model`, `voice`, `response_format`, `speed` (0.25–4.0)
- `response_format` values: `wav`, `mp3`, `flac`, `pcm`, `aac`, `opus`
- Extensions: `task_type` (CustomVoice / VoiceDesign / Base), `language`, `instructions`, `max_new_tokens`, `initial_codec_chunk_frames`, `stream`
- **Constraint**: `stream=true` requires `response_format="pcm"` — must be validated server-side
- Voice cloning params (Base task): `ref_audio` (HTTP URL / base64 data URL / `file://` URI), `ref_text`, `x_vector_only_mode`
- `file://` ref_audio requires the server to be started with `--allowed-local-media-path`
- Supported `language` values (11 total): `Auto`, `Chinese`, `English`, `Japanese`, `Korean`, `German`, `French`, `Russian`, `Portuguese`, `Spanish`, `Italian`
- **OpenAI SDK compatibility**: must work end-to-end with the official `openai` Python client (`client.audio.speech.create(...)` + `response.stream_to_file(...)`), not just raw HTTP

**`GET /v1/audio/voices`** — list voices
- Returns built-in presets and uploaded voices
[...]

PR #618, opened 2026-05-30
[Benchmark]: Add TTS serving benchmark

Ratish1:
## Motivation

Part of #601.

This PR adds a standalone TTS serving benchmark harness for SGLang Omni. The goal is to test serving behavior under API-contract coverage, malformed inputs, stateful voice operations, streaming behavior, and high-concurrency load, not only whether SeedTTS generation can run.

The Higgs TTS CI migration is split out of this PR. This review stays focused on the benchmark harness, scenario matrix, result contract, Docker execution path, and benchmark documentation.

## Architecture

There are two layers in the benchmark flow:

- The benchmark framework owns orchestration. In platform mode it provides `/etc/benchmark/spec.json`, starts the standalone Docker image, mounts `/var/benchmark/out`, and interprets process exit code separately from `results.json` pass/fail.
- The TTS serving harness owns the benchmark mechanics. It reads `base_url`, `model_name`, `test_type`, `seed`, and optional `params` from the spec, expands them into deterministic serving-stress scenarios, sends HTTP/SSE/WebSocket traffic to the target server, classifies every response, and writes artifacts.

```text
/etc/benchmark/spec.json
  -> spec parser
  -> deterministic TTS serving scenario builder
  -> HTTP/SSE/WebSocket clients
  -> result classifier
  -> /var/benchmark/out/{results.json,manifest.json,raw/*.jsonl,logs/harness.log}
```

Exit-code semantics:

- Non-zero exit: the harness could not run, for example invalid spec, unwritable output directory, or a missing auth env required by the configured run.
- Exit 0 with `results.json.overall.passed=false`: the harness ran correctly and the target service failed benchmark/API contracts.

Before all serving APIs are integrated, enabled-but-missing contracts are expected to fail explicitly. After those APIs are integrated, the same serving-stress spec should pass.

## Modifications

- Added `benchmarks/eval/benchmark_tts_serving.py` as the benchmark entry point for direct Python runs and Docker runs.
- Added `benchmarks/tts_serving/` with spec parsing, deterministic scenario generation, HTTP/SSE/WebSocket clients, result classification, metrics, artifact writing, and Docker packaging.
- Added a single serving-stress profile covering normal speech, malformed requests, adversarial text, long-tail text, multilingual payloads, reference-audio requests, SSE streaming, batch API contracts, voice lifecycle/cache/speaker-cap contracts, and WebSocket speech-stream probing.
- Added staged load support for closed-loop, ramp, soak, WebSocket burst, voice cache pressure, speaker-cap pressure, and mixed 512-way burst traffic.
- Added explicit unsupported-contract reporting for enabled API gaps. Missing batch/voices/WebSocket contracts produce endpoint/operation/path/status samples in `results.json` and fail the benchmark as intended.
- Added OpenAI-style error-envelope validation for expected request failures, so malformed inputs must fail with a structured error instead of arbitrary server errors or unexpected success.
- Added `benchmarks/tts_serving/README.md` and `benchmarks/tts_serving/examples/stress.json` documenting the harness contract, endpoint contracts, serving-stress spec, Docker usage, and result artifacts.

## H200 Validation

Server:

```bash
  --port 8000
```

Observed model id:

```text
boson-sglang/higgs-audio-v3-TTS-4B-grpo05200410999
```

Staged serving-stress run:

```bash
python -m benchmarks.eval.benchmark_tts_serving \
  --spec tts-serving-bench/stress/spec.json \
  --out tts-serving-bench/stress/out
```

| Metric | Value |
|---|---:|
| Profile | stress |
| Passed | false |
| Total scenarios | 2050 |
| Succeeded | 608 |
| Failed | 1442 |
| Coverage contract valid | false |
| Load generation valid | true |
| Latency mean | 3.1453784852580555 s |
| Latency p95 | 24.693251197924837 s |
| Latency p99 | 34.081026152009144 s |
| Peak inflight | 512 |

Status counts:

```text
ok: 612
unsupported_contract: 359
failed: 764
unexpected_success: 187
invalid_error_response: 71
invalid_voice_response: 57
```

Endpoint mix:

```text
speech: 817
speech_sse: 111
batch: 143
voices: 317
websocket: 662
```

Stress interpretation:

- `harness_status=ok`, so the harness completed and wrote the required artifacts.
- `load_generation_valid=true`, so the 512-way stress run was not polluted by client-side saturation.
- `coverage_contract_valid=false` because voice cache pressure currently lacks observable cache counters: `eviction_count`, `hit_count`, `miss_count`, `cache_byte_usage`, and `delete_invalidation_counter`.
- `unsupported_contract_count=9`, covering missing enabled batch and voice operations such as `/v1/audio/speech/batch` and `/v1/audio/voices`.
- This is the intended benchmark-first signal: the benchmark exposes the serving/API gaps now, and the follow-up API work should move these failures to passing contracts.

## Checklist

- [x] Add standalone TTS serving benchmark harness.
- [x] Add serving-stress documentation and checked-in stress spec.
- [x] Validate the benchmark against a live H200 Higgs TTS server.
- [ ] Calibrate thresholds on the target CI/H20 environment after the benchmark/API contract is stable.


benchmarks/tts_serving/http_client.py
```diff
@@ -0,0 +1,2196 @@
...
+    try:
+        payload = json.loads(body)
+    except json.JSONDecodeError:
+        return False
+    if not isinstance(payload, dict):
+        return False
+    return any(
+        isinstance(payload.get(key), (dict, list, str)) and bool(payload[key])
+        for key in ("error", "message", "detail")
+    )
```
zhaochenyang20: malformed requests, invalid enums, `stream=true` with non-`pcm`, and bad `ref_audio` cases all depend on `_is_valid_error_response` to decide whether the server rejected the request correctly. The current function returns true for any non-empty `error`, `message`, or `detail`, so responses like `{ "detail": "Not Found" }` or `{ "message": "bad" }` are recorded as `expected_error` and capability pass even though they do not satisfy the OpenAI-compatible `{error: {message, type, param, code}}` shape required by #601. A server that only emits default FastAPI 422/404 detail can therefore get a green result on the most important public API validation contract.

zhaochenyang20: Make `_is_valid_error_response` a strict, status-aware validator. Except for the documented voice-delete exception, HTTP 400 should require an object `error` with non-empty `message`, `type == "BadRequestError"`, and `code == 400`; HTTP 404 should require `type == "NotFoundError"` and `code == 404`; `param` may be `None` or a string, but the field should exist. `_classify_http_failure` should pass the status into the validator, and unit tests should cover the error shape for each malformed scenario.

benchmarks/tts_serving/http_client.py
```diff
@@ -0,0 +1,2196 @@
...
+            "successful item media_type does not match requested format "
+            f"(format={expected_format!r}, media_type={media_type!r})"
+        )
+    try:
+        decoded = base64.b64decode(audio_data, validate=True)
+    except binascii.Error as exc:
+        return f"successful item audio_data is not valid base64: {exc}"
+    if not decoded:
+        return "successful item audio_data decoded to empty bytes"
+    return None
```
zhaochenyang20: `/v1/audio/speech/batch` is part of the production API surface, and the benchmark must prove that every successful item contains audio matching the requested format. `_validate_batch_item` currently only base64-decodes the field and checks that the decoded bytes are non-empty; it does not call `_is_audio_response` or check the WAV/FLAC/Ogg/ADTS/PCM contract. I verified locally that `audio_data=base64("not audio")` with `media_type="audio/wav"` returns `None`, which means it is treated as valid. A batch endpoint can therefore return arbitrary non-empty bytes and still pass as long as the media type string looks right.

zhaochenyang20: After decoding in `_validate_batch_item`, build headers from `media_type` and call `_is_audio_response(decoded, headers, expected_format)`. On failure, return a validation error that includes the expected format, media type, and decoded byte count. On success, add the decoded byte count and computable audio duration back into the batch-level `ScenarioResult` so batch metrics do not omit the actual audio volume.

benchmarks/tts_serving/report.py
```diff
@@ -0,0 +1,863 @@
...
+        failures.append(_coverage_gap("speech.sse", ["speech.sse"]))
+    if "batch" in enabled_endpoint_set:
+        failures.extend(_batch_coverage_failures(scenarios))
+    if "voices" in enabled_endpoint_set:
+        failures.extend(
+            _voice_coverage_failures(spec, scenarios, voice_upload_coverage)
+        )
+    if "websocket" in enabled_endpoint_set:
+        failures.extend(_websocket_coverage_failures(scenarios))
+    return failures
```
zhaochenyang20, line 470: I would block this because, after comparing against the TTS documentation, the benchmark only covers the core request shapes; it does not cover the full API surface or prove real serving scale. `_coverage_failures` checks only the built-in speech, SSE, batch, voices, and WebSocket scenario sets. With the default stress spec, 512 generated scenarios return `coverage_failures=[]`, while the benchmark still does not cover precomputed custom voices via `custom_voice_dir` and manifest loading, uploaded-voice restart persistence, cross-model shared-LRU invalidation, `SPEAKER_SAMPLES_DIR`, configurable `SPEAKER_MAX_UPLOADED`, `tts_batch_max_items`, `--stage-overrides` `max_num_seqs`/`gpu_memory_utilization`, the supported-model matrix, or WebSocket `split_granularity` values beyond `sentence`. `PARAM_KEYS` also accepts only profile/request/load/voice-count/provider fields, so these documented deployment parameters cannot be expressed in the spec. The benchmark can therefore report full coverage while not covering all documented APIs and production serving knobs.

zhaochenyang20: Turn the TTS documentation into an explicit coverage matrix for APIs, configuration, and state contracts, and make `coverage_failures` check it item by item. Each untested dimension should either have a scenario and validator or be marked `out_of_scope` in the spec/report with a reason; it should not disappear by default. Concretely, add a multi-target/model spec, restart/persistence stage, precomputed voice fixture, cross-model cache-invalidation observation point, server deployment-knob fields, `tts_batch_max_items` variants, stage-override recording/validation, and a real serving profile based on target duration/RPS/concurrency. The report should mark the coverage contract complete only after these dimensions are tested or explicitly excluded.

benchmarks/tts_serving/scenarios.py
```diff
@@ -0,0 +1,1597 @@
...
+    ("aac", "audio/aac"),
+    ("webm", "audio/webm"),
+    ("mp4", "audio/mp4"),
+)
+VOICE_UPLOAD_REJECT_FORMATS = VOICE_UPLOAD_SUCCESS_FORMATS[1:]
+VOICE_SMALL_UPLOAD_BYTES = VOICE_UPLOAD_WAV_FIXTURE_SIZE
+VOICE_MAX_UPLOAD_BYTES = 10 * 1024 * 1024
+VOICE_NEAR_LIMIT_BYTES = VOICE_MAX_UPLOAD_BYTES - 1
+VOICE_OVERSIZED_BYTES = VOICE_MAX_UPLOAD_BYTES + 1
+SPEAKER_MAX_UPLOADED = 1000
```
zhaochenyang20: #601 defines `SPEAKER_MAX_UPLOADED` as defaulting to 1000, not as a protocol constant that is always 1000. The scenario layer hard-codes `SPEAKER_MAX_UPLOADED = 1000`, and the HTTP sequence uses that constant to decide how many voices to upload before expecting the cap. If the target service uses a non-default cap, for example a release candidate lowers the cap for resource isolation, the benchmark keeps filling toward 1000 and treats the earlier valid 400 as a failure; if the cap is raised, it expects overflow too early. This freezes a deployment knob into the benchmark and can produce false failures or false passes.

zhaochenyang20: Add a `speaker_max_uploaded` or `voice_speaker_cap` field to the spec, default it to 1000, and store it in scenario metadata and report config. `_run_voice_speaker_cap_sequence` should use the target cap from the spec instead of importing a constant from `scenarios.py`. The README example should also state that the benchmark cap must match the server-side `SPEAKER_MAX_UPLOADED` configuration.

benchmarks/tts_serving/http_client.py
```diff
@@ -0,0 +1,2196 @@
...
+                session,
+                spec,
+                scenario,
+                result,
+                voice_name=voice_name,
+                prompt=f"Cache revisit synthesis request for {voice_name}.",
+            ):
+                return
+        _mark_success(result, capability="pass")
+    finally:
```
zhaochenyang20: both the PR and #601 treat cache pressure, LRU eviction, and repeated voice hit/miss behavior as production-grade coverage goals. `_run_voice_cache_pressure_sequence` currently uploads N voices, synthesizes with each voice, revisits a few old voices, and then calls `_mark_success` as long as those speech requests return PCM. It does not read server instrumentation, response headers, a debug endpoint, artifact counters, or any cache hit/miss/eviction signal. A `voices.cache_pressure` pass in the report therefore only proves that synthesis still works after voice uploads; it does not prove that the 512 MiB LRU was pressured, eviction happened, old voices were re-extracted, or delete invalidated cross-model caches. The risk is that a release gate treats a traffic-only test as a cache-contract pass.

zhaochenyang20: Split this into separate reportable results for traffic pressure and cache-contract validation, or add an optional instrumentation URL/header contract to the spec. Mark `voices.cache_pressure` as pass only when the benchmark can observe the expected eviction count, hit/miss count, cache byte usage, or delete invalidation counter; otherwise report it as a coverage gap or `instrumentation_missing`, not success.

zhaochenyang20 (changes requested):

benchmarks/tts_serving/voice_client.py
```diff
@@ -0,0 +1,1415 @@
...
+        )
+        return
+    if scenario.capability_key in {"voices.upload", "voices.lifecycle"}:
+        if _voice_upload_response_identifier(payload):
+            _mark_success(result, capability="pass")
+            return
+        _mark_protocol_error(
+            result,
+            status="invalid_voice_response",
+            error="voice upload response must include id, voice_id, or name",
```
zhaochenyang20, line 1879: regular `voices.upload` scenarios call `_mark_success` as soon as the upload response exposes an identifier, and they never delete the voice they just created. The default production spec generates 11 successful upload scenarios, including four near-limit uploads just under 10 MB; those voices remain in `SPEAKER_SAMPLES_DIR` and in the server's `uploaded_voices` list. Repeated runs, different stage ids, stress random scenarios, or reruns after failures can accumulate persistent state and change `SPEAKER_MAX_UPLOADED`, the speaker-cap scenario, disk usage, and later benchmark baselines. A production benchmark must be repeatable and self-cleaning; this one dirties the target service.

Convert standalone upload into a stateful flow: after upload, verify metadata through `GET /v1/audio/voices`, delete the created name in a `finally` block, and list again to prove it is gone. Every scenario that creates a voice should register the name with one cleanup list; cleanup failure should fail the scenario, or at least be written as an explicit cleanup failure in results, instead of silently leaving persistent objects behind. Near-limit formats need the same cleanup path.

benchmarks/tts_serving/audio_validation.py
```diff
@@ -0,0 +1,249 @@
...
+        return AudioValidation(
+            False,
+            f"{response_format} audio is too small to be a generated response",
+        )
+    if not has_prefix:
+        return AudioValidation(
+            False,
+            f"{response_format} audio is missing the expected container prefix",
+        )
+    return AudioValidation(True)
```
zhaochenyang20: `validate_audio_response` is the shared success contract for speech, SDK, and batch audio, but `_validate_compressed` checks only minimum length and a magic prefix for `mp3`, `flac`, `opus`, and `aac`. I verified locally that `b"ID3" + b"\0" * 40`, `b"fLaC" + b"\0" * 40`, `b"OggS" + b"\0" * 40`, and `b"\xff\xf1" + b"\0" * 40` all return `AudioValidation(ok=True)`. That lets a server return garbage bytes with the right prefix while the benchmark records all four compressed formats as passing. The risk is that a release gate can treat unplayable compressed output as real TTS audio; #601 requires real audio for every `response_format`, not just a container prefix.

zhaochenyang20: Use real decoding or a sufficiently strict container parser for compressed formats. The Docker image can include a stable lightweight decoding path, or an existing audio library can parse duration and sample/frame count; the validator should require positive duration above the minimum, successful decoding, and no pure-silence or placeholder bytes. At minimum, add unit tests for the four forged payloads so they can no longer pass.

benchmarks/tts_serving/scenarios.py
```diff
@@ -0,0 +1,1835 @@
...
+        endpoint="speech",
+        category="speech_reference",
+        stage_id=stage.id,
+        capability_key="speech.reference",
+        payload=payload,
+        expect_success=True,
+        expected_status_class="success",
+        description="valid base64 ref_audio voice cloning request",
+        planned_metadata={"reference_case": "valid_base64_ref_audio"},
+    )
```
zhaochenyang20, line 762: `_speech_reference_base64_success` marks the base64 `ref_audio` case as a success scenario, but `_tiny_wav_data_uri()` generates a mono WAV with only a 48-byte data chunk filled with zeros. That is roughly one millisecond of silence and is not a valid voice-cloning reference. A correct implementation can reasonably return 400 because the reference is too short, silent, or unusable for speaker-feature extraction; the benchmark would record that reasonable rejection as a speech.reference failure. This scenario tests whether the service accepts a bad reference, not whether it supports base64 data URLs.

zhaochenyang20: Replace `_tiny_wav_data_uri()` with a real, small, valid reference audio fixture, for example the 1.2-second non-zero PCM WAV from `voice_upload_fixtures.py` encoded as a data URL. If the tiny or silent reference is worth keeping, it should be an expected-error scenario that requires a 400 OpenAI error envelope.

benchmarks/tts_serving/voice_client.py
```diff
@@ -0,0 +1,1415 @@
...
+            )
+            return
+        if not _is_valid_voice_delete_success(delete_body):
+            _mark_protocol_error(
+                result,
+                status="invalid_voice_response",
+                error="voice lifecycle delete response must be success JSON",
+            )
+            return
+    _mark_success(result, capability="pass")
```
zhaochenyang20: `run_voice_lifecycle` uploads a voice, checks that DELETE returns 2xx and `_is_valid_voice_delete_success` is true, and then marks the lifecycle as passing. It does not call `GET /v1/audio/voices` again to confirm the name disappeared from `uploaded_voices`, and it does not try synthesis with the deleted voice and require a controlled 404. A server can return `{"success": true}` while leaving the disk object, the list entry, and later voice use intact, and this benchmark would still green `voices.lifecycle`. The #601 delete contract is to delete the uploaded voice and invalidate cached artifacts, not merely to return success JSON.

zhaochenyang20: Add hard checks after lifecycle delete: the uploaded voice must no longer appear in the list; if uploaded voices can be referenced by name, a speech request using the deleted voice must return a structured 404; and when cache-invalidation instrumentation exists, the delete should advance the invalidation counter. The same helper should be reused by cleanup paths so an implementation that reports delete success while keeping the object cannot pass.

zhaochenyang20 (changes requested):

benchmarks/tts_serving/voice_client.py
Ratish1: done

benchmarks/tts_serving/audio_validation.py
Ratish1: done

benchmarks/tts_serving/scenarios.py
Ratish1: done

benchmarks/tts_serving/voice_client.py
Ratish1: done

benchmarks/tts_serving/voice_client.py
```diff
@@ -0,0 +1,1684 @@
...
+    session: aiohttp.ClientSession,
+    spec: BenchmarkSpec,
+    voice_names: list[str],
+) -> str | None:
+    for voice_name in reversed(voice_names):
+        delete_url = api_url(spec.base_url, f"/v1/audio/voices/{voice_name}")
+        try:
+            status, body, _ = await _raw_delete(session, delete_url)
+        except (aiohttp.ClientError, asyncio.TimeoutError) as exc:
+            return f"voice cleanup failed for {voice_name!r}: {exc}"
```
JingwenGu0829: Could we make cleanup best-effort across all `voice_names` and aggregate failures instead of returning on the first failed delete/list verification? The README says repeated runs should not consume persistent speaker slots, and the cache-pressure/speaker-cap flows can create many uploaded voices. With the current early returns paths, one failed cleanup can prevent later voices from being cleaned up, leaking state into subsequent benchmark runs.

benchmarks/tts_serving/scenarios.py
```diff
@@ -0,0 +1,1804 @@
...
+        expect_success=False,
+        expected_status_class="client_error",
+        expected_http_status=400,
+        expected_error_type="BadRequestError",
+        description="intentionally bad reference audio",
+        planned_metadata={"reference_case": reference_case},
+    )
+
+
+def _speech_sse(index: int, spec: BenchmarkSpec, stage: LoadStage) -> Scenario:
```
JingwenGu0829, line 839: The README says `speech_sse` covers streaming error cases, but `_speech_sse` only creates the success payload with `stream=true` and `response_format=pcm`. The invalid streaming case, `stream_non_pcm`, is currently generated under the normal malformed speech payloads, so it does not exercise the SSE parser/error path. Should we add an explicit expected-error `speech_sse` scenario?

JingwenGu0829 (commented):
Browsed through readme and got some potential findings that might help! It's a huge PR so I think I can't read all of the code. But I'm willing to help, review and implement further with the help of agent.

JiaxinD (commented):
Re-reviewed against the current head. The earlier threads from @zhaochenyang20 and @JingwenGu0829 look genuinely addressed (strict status-aware error envelope, batch audio validation, real compressed-format decoding, valid ref fixtures, stateful lifecycle delete plus 404 checks, best-effort cleanup aggregation, `speech_sse` error scenario, configurable `speaker_max_uploaded`). The coverage-matrix completeness thread (`report.py`) is the one I would still call partially open, so I am leaving that to your existing thread. A few new lower-severity items below; none blocking.

### Low
* **PCM audio validation rejects a missing/empty `Content-Type` while WAV accepts it** (`audio_validation.py`): `EXPECTED_AUDIO_CONTENT_TYPES["wav"]` includes `""`, but `"pcm"` maps to `PCM_CONTENT_TYPES`, which does not. A conforming server returning valid raw PCM for `response_format=pcm` but omitting the header is flagged `invalid_audio_response`, while the identical missing-header case for WAV passes. The in-repo server always sets `audio/pcm` so CI is unaffected, but the harness is meant to validate arbitrary servers. PCM already accepts `application/octet-stream`, so rejecting `""` is inconsistent: add `""` to `PCM_CONTENT_TYPES` for parity (or treat missing/empty uniformly), plus a small unit test.
* **WebSocket non-streaming `expect: "audio"` consumes exactly one binary frame** (`ws_client.py`): the per-sentence scripts use `{audio.start, audio, audio.done}`, but `expect: "audio"` returns after one BINARY frame. If the server splits a sentence's PCM into more than one binary frame even with `stream_audio=False`, the next step `expect: "audio.done"` gets a second BINARY frame, hits the guard, and is marked a protocol error: a false-fail despite correct audio. The streaming path already drains via `_expect_audio_until_done`. Make the non-streaming step drain consecutive binary frames until the next text event. (Low, since it only bites if the server emits multiple frames per sentence in non-streaming mode.)
* **Dockerfile runs as root**: artifacts written to the bind-mounted `/var/benchmark/out` land on the host owned by root, so a non-root user re-running cannot clean prior output without sudo. Bake an unprivileged user (`useradd ... benchmark` plus `USER benchmark`), or document `--user "$(id -u):$(id -g)"` in the README.
* **Exit-code docs contradict the error path**: the catch-all `except Exception` in `benchmark_tts_serving.py` calls `write_artifacts(...)` (which writes `results.json` with `harness_status="error"`) then `return 2`, so `results.json` can coexist with a non-zero exit, contradicting "`0`: the harness ran and wrote results.json", and the listed non-zero causes omit the most common one (any unhandled runtime error). Align the docs and have consumers branch on `harness_status`, not the exit code alone.
* **Internal duplication / drift risk** (DRY, no behavior change): a few helpers are duplicated and can silently diverge. (1) coverage value-gaps are computed twice, in `_*_coverage_matrix` and again in `_*_coverage_failures`; (2) `_mark_unsupported_ws_contract` is a near-verbatim copy of the shared `_mark_unsupported_contract`; (3) `UNSUPPORTED_HTTP_STATUSES = {404, 405, 501}` is defined three times (sdk_client, http_contracts, and `UNSUPPORTED_WS_STATUSES` in ws_client); (4) the transport-error result-marking block is copy-pasted across http_client/sdk_client/ws_client; (5) `_scenario_response_format` is reimplemented in http_client and sdk_client and has already drifted (different precedence and return type); (6) the voice HTTP record-then-classify preamble repeats across about 6 functions in `voice_client.py` despite the file already centralizing it via `_raw_post`/`_merge_raw_voice_response`. Consolidating each into a single shared definition would remove the drift surface.
* **Dockerfile build hygiene**: no `.dockerignore`, so the build context is the whole repo (including `.git` and any local caches) since the documented build uses `.` as context; and the base image is pinned only by the mutable tag `python:3.11-slim`. A small repo-root `.dockerignore` is the higher-payoff one; pin by digest if image reproducibility matters.

### Nits
* **Soak `request_rate` is silently overwritten** (`spec.py`): a user-provided rate on a soak stage is discarded without warning. Reject it (or log a warning) instead.
* **README mislabels `manifest.json`**: it is listed as containing a "spec hash", but `write_artifacts()` only writes `scenario_set_hash` (a hash of the generated scenario set, not the spec). Correct the comment, or add a real `spec_hash` if drift detection is intended.

benchmarks/eval/benchmark_tts_serving.py
```diff
@@ -0,0 +1,462 @@
...
+    scheduled_task_count = 0
+    for scenario, offset in zip(scenarios, offsets, strict=True):
+        planned_start = stage_start + offset
+        delay_s = planned_start - time.perf_counter()
+        if delay_s > 0:
+            await asyncio.sleep(delay_s)
+        done = {task for task in pending if task.done()}
+        if done:
+            pending.difference_update(done)
+            results.extend(task.result() for task in done)
```
xinlij: If any task in `done` raised an exception, `task.result()` re-raises it inside the generator, which aborts `extend` mid-way. The remaining tasks in `done` are already removed from `pending` (line above), so their results are silently lost. Wrap with try/except per task or filter out failed tasks before extending.

benchmarks/eval/benchmark_tts_serving.py
```diff
@@ -0,0 +1,462 @@
...
+                    generator_lag=max(0.0, actual_start - planned_start),
+                )
+            )
+            continue
+        active_requests += 1
+        peak_inflight = max(peak_inflight, active_requests)
+        pending.add(asyncio.create_task(run_planned(scenario, offset)))
+        peak_pending_tasks = max(peak_pending_tasks, len(pending))
+    if pending:
+        results.extend(await asyncio.gather(*pending))
```
xinlij: `asyncio.gather` without `return_exceptions=True` re-raises the first task exception and discards all other results. Use `return_exceptions=True` so a single failing scenario doesn't abort the rest.

benchmarks/tts_serving/http_client.py
```diff
@@ -0,0 +1,446 @@
...
+            status="invalid_sse_response",
+            error=(
+                "SSE speech endpoint returned 2xx without text/event-stream "
+                f"content-type: {response.headers.get('Content-Type')!r}"
+            ),
+        )
+        return
+
+    buffer = bytearray()
+    audio_buffer = bytearray()
```
xinlij: `audio_buffer` accumulates all PCM chunks for the entire stream with no size cap. For a long TTS response, this can hold tens of MB per connection — at 512 concurrent SSE connections in the burst stage, total memory usage can reach GB scale. Consider validating and discarding each chunk incrementally instead of buffering the full stream.

xinlij (changes requested):

benchmarks/tts_serving/scenarios.py
```diff
@@ -0,0 +1,1836 @@
...
+    "wrong_input_type",
+    "bad_response_format",
+    "bad_language",
+    "bad_task_type",
+    "speed_below_min",
+    "speed_above_max",
+    "stream_non_pcm",
+    "negative_max_new_tokens",
+    "adversarial_text",
+)
```
zhaochenyang20, line 136: Reference coverage includes HTTP success, base64 success, x-vector-only, and a `file:///etc/passwd` disallowed failure. #601 explicitly requires `file:// ref_audio` gated by `--allowed-local-media-path`, and the Higgs validation matrix requires a `file:// with --allowed-local-media-path` happy path. This benchmark proves only that an illegal local path is rejected; it does not prove that a deployment-explicit allowed local media path works. If #659 allowlist plumbing is broken, the benchmark can still show `speech.reference_cases` as complete.

benchmarks/tts_serving/scenarios.py
```diff
@@ -0,0 +1,1836 @@
...
+    ("mp4", "audio/mp4"),
+)
+VOICE_UPLOAD_REJECT_FORMATS = VOICE_UPLOAD_SUCCESS_FORMATS[1:]
+VOICE_NEAR_LIMIT_FORMATS = VOICE_UPLOAD_SUCCESS_FORMATS
+VOICE_NEAR_LIMIT_GENERATED_FORMATS = (
+    ("wav", "audio/wav"),
+    ("mp3", "audio/mpeg"),
+    ("flac", "audio/flac"),
+    ("mp4", "audio/mp4"),
+)
```
zhaochenyang20: #601 requires audio just under the 10 MB limit in every supported format: `wav/mp3/flac/ogg/aac/webm/mp4`. The code defines `VOICE_NEAR_LIMIT_FORMATS` as all seven formats, but `VOICE_NEAR_LIMIT_GENERATED_FORMATS` contains only `wav/mp3/flac/mp4`; `ogg/aac/webm` are placed in `VOICE_NEAR_LIMIT_DEFERRED_FORMATS`. Worse, `report.py` marks `voices.near_limit_deferred_formats` as `out_of_scope` unconditionally, so the coverage contract does not fail for missing three required formats. The benchmark can pass with a matrix below #601's requirement.

benchmarks/tts_serving/audio_validation.py
```diff
@@ -0,0 +1,337 @@
...
+
+PCM_CONTENT_TYPES = frozenset({"application/octet-stream", "audio/pcm", "audio/raw"})
+EXPECTED_AUDIO_CONTENT_TYPES = {
+    "wav": {"audio/wav", "audio/x-wav", "application/octet-stream", ""},
+    "pcm": PCM_CONTENT_TYPES,
+    "mp3": {"audio/mpeg", "audio/mp3"},
+    "flac": {"audio/flac"},
+    "aac": {"audio/aac", "audio/aacp"},
+    "opus": {"audio/opus", "audio/ogg"},
+}
```
zhaochenyang20: `EXPECTED_AUDIO_CONTENT_TYPES["wav"]` accepts the empty string, but `pcm` accepts only `application/octet-stream`, `audio/pcm`, and `audio/raw`. I verified with a non-zero PCM buffer above the minimum duration that `validate_audio_response(..., response_format="pcm", content_type="")` returns false while `application/octet-stream` returns true. Raw PCM has no container header; if a server returns valid PCM but omits the header, the benchmark records `invalid_audio_response`. The same validator disables Content-Type requirements in the SDK path, so PCM classification differs by path; the risk is a valid service being incorrectly marked as failure by the benchmark.

benchmarks/tts_serving/report.py
```diff
@@ -0,0 +1,1557 @@
...
+VOICE_CACHE_OBSERVABILITY_COUNTERS = (
+    "eviction_count",
+    "hit_count",
+    "miss_count",
+    "cache_byte_usage",
+    "delete_invalidation_counter",
+)
+
+
+def build_results_report(
```
zhaochenyang20, line 35: `build_results_report` computes coverage from `scenarios`, not from `results`. If a scenario is generated but never reaches the corresponding response validator due to load-generator saturation, transport setup failure, or a client-side false failure, the coverage matrix still shows that value as observed. It is good that overall `load_generation_valid` becomes false on saturation, but the coverage matrix itself says "tested" without distinguishing "planned" from "executed and classified." That weakens release-report interpretability.

 Split the matrix into at least `planned_coverage` and `executed_coverage`. `coverage_contract_valid` should be based on results that reached the relevant validator, or be marked indeterminate when an entire class only saw transport/load-generator failures, rather than tested. Keep the raw scenario set hash for reproducibility.

benchmarks/tts_serving/Dockerfile
zhaochenyang20, line 1: The Dockerfile uses `python:3.11-slim` as root, and the README `docker run -v "$PWD/results...:/var/benchmark/out"` writes root-owned artifacts on the host. A non-root user may need sudo to clean or rerun. This benchmark is intended for repeated release/staging use and should not default to bad file ownership; the risk is automation or developers failing to rerun because output cleanup needs elevated permissions.

Create a non-root `benchmark` user in the image and set `USER benchmark`, or at least recommend `--user "$(id -u):$(id -g)"` in the README. Ensure `/var/benchmark/out` remains writable.

zhaochenyang20: The README uses the repo root as Docker build context, but there is no `.dockerignore`, so `.git`, local caches, temporary results, and unrelated files are sent to the Docker daemon. The base image is also the mutable tag `python:3.11-slim`. Benchmark artifacts should come from reproducible build inputs and runtime environments; the risk is slow builds, local context exposure, or different benchmark environments on different build dates.

zhaochenyang20: Add a repo-root `.dockerignore` excluding at least `.git`, `.venv`, `__pycache__`, benchmark output, and model/cache directories. If reproducibility matters, pin the base image by digest.

zhaochenyang20 (changes requested):

benchmarks/tts_serving/audio_validation.py
```diff
@@ -0,0 +1,337 @@
...
+    if ffmpeg is None:
+        return _DecodedPcm(
+            False,
+            error=(
+                "ffmpeg is required to validate compressed TTS audio responses "
+                f"({response_format})"
+            ),
+        )
+    try:
+        completed = subprocess.run(
```
edwingao28, line 228: `subprocess.run(ffmpeg, ...)` reached synchronously from the async path (`http_client.py` `_handle_binary_response` to `validate_audio_response`, and `ws_client.py`). On asyncio's single loop it freezes every in-flight request for the whole decode (up to the 10s timeout) whenever the server returns mp3/flac/aac/opus. At the 512-way burst this serializes all requests behind each decode, so the measured latency / `peak_inflight` / `generator_lag` no longer reflect the server — yet `load_generation_valid` still reports `true`.

Could we offload at the async call sites, e.g. `await asyncio.to_thread(validate_audio_response, ...)`? `sdk_client.py` already does exactly this; the HTTP/WS paths just don't.

benchmarks/tts_serving/scenarios.py
```diff
@@ -0,0 +1,1836 @@
...
+    return hashlib.sha256(encoded).hexdigest()
+
+
+def _build_stage_scenarios(spec: BenchmarkSpec, stage: LoadStage) -> list[Scenario]:
+    rng = random.Random(f"{spec.seed}:{stage.id}")
+    endpoint_set = set(stage.enabled_endpoints or spec.params.enabled_endpoints)
+    required_scenarios = _required_stage_scenarios(spec, stage, endpoint_set)
+    if _stage_voice_speaker_cap_count(spec, stage):
+        return required_scenarios
+    scenarios = required_scenarios[: stage.request_count]
```
edwingao28: This silently drops required coverage when `request_count` < the required-scenario count . The checked-in `stress.json` `closed-1` stage (`request_count: 32`) keeps 32 and discards the rest required scenarios including the malformed / expected-error contract checks. `scenario_set_hash` and the coverage matrix still report "complete", so the run goes green while never exercising most negative-path contracts, the opposite of this benchmark's benchmark-first goal.

Bound only the weighted *fill* by `request_count`; never truncate the required set. Raise or at warn when `request_count < len(required_scenarios)`.

benchmarks/tts_serving/http_client.py
```diff
@@ -0,0 +1,446 @@
...
+                if scenario.endpoint == "speech_sse":
+                    await _handle_sse_response(
+                        response,
+                        result,
+                        start,
+                        scenario,
+                    )
+                else:
+                    await _handle_binary_response(response, result, start, scenario)
+    except (aiohttp.ClientError, asyncio.TimeoutError) as exc:
```
edwingao28, line 125: This catches only `ClientError` / `TimeoutError`. Any other exception (a `KeyError` from a malformed `scenario.payload`, aiohttp's `RuntimeError: Session is closed`, `CancelledError`) escapes the task and lands on the unguarded harvest (`task.result()` / `asyncio.gather` in `benchmark_tts_serving.py`), which re-raises and aborts the whole stage — dropping sibling results and orphaning every still-pending task (leaked sockets).

This is the root trigger behind the `.result()` / `gather` comments: `return_exceptions=True` alone isn't enough. Wrap each scenario body in `except Exception` -> a failed `ScenarioResult`, so one bad scenario can't take down the stage. (Same narrow catch in `ws_client.py`; `sdk_client.py` has no guard around `to_thread`.)

benchmarks/tts_serving/report.py
```diff
@@ -0,0 +1,1557 @@
...
+    voice_upload_coverage = _voice_upload_coverage(scenarios or [])
+    coverage_matrix = _coverage_matrix(spec, scenarios or [], voice_upload_coverage)
+    coverage_failures = _coverage_failures(
+        spec,
+        scenarios or [],
+        voice_upload_coverage,
+        coverage_matrix,
+    )
+    coverage_contract_valid = not coverage_failures
+    passed = (
```
edwingao28, line 80: With an empty result set every conjunct here is vacuously true (`all([]) == True`, `not any([]) == True`), so `overall.passed` becomes `true` when **zero scenarios ran**. A misconfigured spec (typo in `enabled_endpoints`, `request_count: 0`) then reports green having validated nothing, and CI gating on `overall.passed` lets it through.

Fix: require real work, e.g. add `total > 0` to this condition.

benchmarks/tts_serving/artifacts.py
```diff
@@ -0,0 +1,155 @@
...
+        (out_dir / "logs" / "harness.log").write_text(
+            "\n".join(lines) + "\n", encoding="utf-8"
+        )
+    except OSError as exc:
+        raise ArtifactError(f"failed to write harness log: {exc}") from exc
+
+
+def _write_json(path: Path, payload: dict[str, Any]) -> None:
+    try:
+        path.write_text(
```
edwingao28: `results.json` / `manifest.json` are written with `write_text` (truncate-then-write). If the process dies mid-write (OOM / SIGTERM / pod eviction — all normal in the platform/Docker mode this targets), the primary machine-read artifact is left as truncated, invalid JSON — indistinguishable from "the run never happened", exactly in the failure cases where partial results matter most.

Could we write to a temp file then `os.replace(tmp, path)` (atomic rename on POSIX)> Consider `allow_nan=False` in `json.dumps` too, so a stray NaN fails loudly instead of emitting invalid JSON.

edwingao28 (commented):
Due to the size of this PR, I wasn’t able to review every line in detail, but I noticed several areas that could be improved.

benchmarks/tts_serving/voice_client.py
Ratish1: done

benchmarks/tts_serving/scenarios.py
Ratish1: done

benchmarks/eval/benchmark_tts_serving.py
Ratish1: done

benchmarks/eval/benchmark_tts_serving.py
Ratish1: done

benchmarks/tts_serving/http_client.py
Ratish1: done

benchmarks/tts_serving/scenarios.py
Ratish1: done

benchmarks/tts_serving/scenarios.py
Ratish1: done

benchmarks/tts_serving/audio_validation.py
Ratish1: done

benchmarks/tts_serving/report.py
Ratish1: done

benchmarks/tts_serving/Dockerfile
Ratish1: done

benchmarks/tts_serving/audio_validation.py
Ratish1: done

benchmarks/tts_serving/scenarios.py
Ratish1: done

benchmarks/tts_serving/http_client.py
Ratish1: done

benchmarks/tts_serving/report.py
Ratish1: done

benchmarks/tts_serving/artifacts.py
Ratish1: done

Ratish1:
Thanks for the reviews @JiaxinD @JingwenGu0829 @edwingao28 @xinlij @zhaochenyang20 . I have addressed all of them

benchmarks/tts_serving/audio_validation.py
```diff
@@ -0,0 +1,339 @@
...
+
+def _has_aac_adts_prefix(body: bytes) -> bool:
+    return (
+        len(body) >= MIN_AUDIO_FRAME_PREFIX_BYTES
+        and body[0] == 0xFF
+        and (body[1] & 0xF0) == 0xF0
+    )
+
+
+def _has_nonzero_int16_sample(body: bytes) -> bool:
```
yxs, line 339: `_has_nonzero_int16_sample` scans every 16-bit sample in Python via `struct.iter_unpack("<h", ...)`, returning on the first non-zero. For normally generated audio it usually hits quickly, but the `near_limit` scenarios upload/validate WAVs close to 10MB (`VOICE_NEAR_LIMIT_BYTES = 10MB-1`), so the worst case (long all-zero prefix, or all-zero) iterates ~5M samples in pure Python. It runs inside `asyncio.to_thread` so it does not block the event loop, but it noticeably slows the individual scenario.

benchmarks/tts_serving/ws_client.py
```diff
@@ -0,0 +1,551 @@
...
+def _is_valid_audio_start(event: dict) -> bool:
+    return (
+        isinstance(event.get("sentence_index"), int)
+        and event["sentence_index"] >= 0
+        and isinstance(event.get("sentence_text"), str)
+        and bool(event["sentence_text"])
+        and isinstance(event.get("format"), str)
+        and event["format"] == "pcm"
+        and isinstance(event.get("sample_rate"), int)
+        and event["sample_rate"] == PCM_SAMPLE_RATE
```
zijiexia: This hardcodes the `audio.start` sample rate to `== PCM_SAMPLE_RATE` (24000), but the sample rate is per-model: `fishaudio_s2_pro` and `ming_omni` both output 44100 (see models/fishaudio_s2_pro/payload_types.py:42), and the server forwards the model's real rate into the event (serve/openai_api.py:616). So running WS against these models, `audio.start` carries `sample_rate: 44100` and gets wrongly rejected as `invalid audio.start`.

The SSE path already does this right (http_client.py:415 reads the rate from the event; 24000 is only a fallback). Maybe align WS: don't require a fixed value, just validate it's a positive int. Downstream (:321/:455) already uses the per-event `ws_active_sample_rate`, so duration/RTF will compute correctly for 44.1 kHz too.

benchmarks/tts_serving/README.md
```diff
@@ -0,0 +1,248 @@
...
+benchmark container.
+
+## Scenario Matrix
+
+The matrix is deterministic for a given spec seed. It covers:
+
+- speech success paths across response formats, task types, speed boundaries,
+  reference-audio shapes, allowed `file://` references, SDK calls, and SSE
+  streaming
+- malformed speech requests that must return OpenAI-compatible error envelopes
```
zijiexia: > OpenAI-compatible error envelopes

seems not in  `benchmarks/tts_serving/error_contract.py:8-11, 32, 34`

benchmarks/tts_serving/README.md
```diff
@@ -0,0 +1,248 @@
...
+## Endpoint Contracts
+
+`enabled_endpoints` controls which API families are part of the run:
+
+| Endpoint family | Contract |
+|-----------------|----------|
+| `speech` | `POST /v1/audio/speech` with response formats, task types, language handling, speed bounds, reference audio, SDK compatibility, and malformed-request classification. |
+| `speech_sse` | `POST /v1/audio/speech` with server-sent event streaming and streaming error cases. |
+| `batch` | `POST /v1/audio/speech/batch` with 1-32 item batches, per-item overrides, item-level success/error records, and oversized batch rejection. |
+| `voices` | `GET`, `POST`, and `DELETE /v1/audio/voices` with upload formats, metadata, overwrite, delete, speaker-cap, cleanup, race, and cache-pressure behavior. |
```
zijiexia, line 133: Can we also document the body of the contract?

benchmarks/tts_serving/http_client.py
```diff
@@ -0,0 +1,469 @@
...
+async def _handle_binary_response(
+    response: aiohttp.ClientResponse,
+    result: ScenarioResult,
+    start: float,
+    scenario: Scenario,
+) -> None:
+    result.http_status = response.status
+    result.http_status_class = classify_http_status(response.status)
+    result.response_headers = dict(response.headers)
+    body = await response.read()
```
zijiexia: read the whole file into memory might not be a good idea. consider using `response.content.iter_chunked(...)` and a cap (e.g., 64 MiB), or set `aiohttp.ClientTimeout(sock_read=...)` plus `read_until_eof`

benchmarks/tts_serving/audio_validation.py
```diff
@@ -0,0 +1,336 @@
...
+                "s16le",
+                "-acodec",
+                "pcm_s16le",
+                "-ac",
+                str(FFMPEG_DECODE_CHANNELS),
+                "-ar",
+                str(PCM_SAMPLE_RATE),
+                "pipe:1",
+            ],
+            input=body,
```
zijiexia, line 247: same here: https://github.com/sgl-project/sglang-omni/pull/618/changes#r3386690861

benchmarks/tts_serving/http_client.py
```diff
@@ -0,0 +1,469 @@
...
+        body, body_text = await _response_body_and_text(response)
+        result.response_bytes = len(body)
+        if _is_unsupported_http_status(response.status, scenario):
+            _mark_unsupported_contract(result, scenario, body=body_text)
+            return
+        _classify_http_failure(response.status, body_text, result, scenario)
+        return
+    content_type = str(response.headers.get("Content-Type", "")).lower()
+    if "text/event-stream" not in content_type:
+        body = await response.read()
```
zijiexia: same here: https://github.com/sgl-project/sglang-omni/pull/618/changes#r3386690861

benchmarks/tts_serving/README.md
```diff
@@ -0,0 +1,248 @@
+# TTS Serving Benchmark
+
+This benchmark validates the serving behavior of SGLang Omni's
+OpenAI-compatible TTS service. It is separate from SeedTTS quality evaluation:
+SeedTTS measures model quality and speed on a fixed corpus, while this
+benchmark exercises API contracts, malformed requests, stateful voice
+operations, streaming behavior, and high-concurrency load.
+
+Note: before all serving APIs are integrated, enabled-but-missing contracts are
```
zijiexia, line 9: > before all serving APIs are integrated, enabled-but-missing contracts are expected to fail explicitly; after those APIs are integrated, the same benchmark spec should pass.

This sentence implies the example spec becomes green once the server-side APIs land. But the cache-observability gap doesn't go away when the server adds endpoints — it only goes away when the **harness itself** grows a client that reads observability counters. A first-time user runs `examples/stress.json`, sees `overall.passed=False` plus an `instrumentation_missing` failure, and reasonably concludes their server has a bug instead of recognizing it as a harness-side TODO.

Maybe keep `examples/stress.json` runnable to green by default (`voice_cache_pressure_voice_count: 0`) and add a separate `examples/stress_with_cache_pressure.json` for the intentionally-failing variant?

benchmarks/tts_serving/http_contracts.py
```diff
@@ -0,0 +1,227 @@
...
+        _mark_protocol_error(
+            result,
+            status=status,
+            error=f"{error_prefix}: {exc}",
+        )
+        result.error_type = exc.__class__.__name__
+        return None
+
+
+def _is_valid_missing_voice_delete_response(status: int, body: str) -> bool:
```
zijiexia, line 224: While running the harness end-to-end I noticed that returning a fully-compliant OpenAI error envelope for a missing-voice DELETE 404 is still rejected.  The harness requires `success: false` at the top level. Can we document it somewhere in the README?

benchmarks/tts_serving/ws_client.py
Ratish1: done

benchmarks/tts_serving/README.md
Ratish1: done

benchmarks/tts_serving/README.md
Ratish1: done

benchmarks/tts_serving/http_client.py
Ratish1: done

benchmarks/tts_serving/README.md
Ratish1: done

benchmarks/tts_serving/http_contracts.py
Ratish1: done

benchmarks/tts_serving/http_client.py
Ratish1: done

benchmarks/tts_serving/audio_validation.py
Ratish1: done

zhaochenyang20 (approved):

Issue #571, opened 2026-05-25
[Perf] Enable torch.compile for S2-Pro Slow AR backbone + Fast AR codebook decoder


Ccyest:
## Motivation

S2-Pro currently runs with CUDA graph capture enabled on the Slow AR path but with zero `torch.compile` anywhere. The Qwen3 backbone (24 layers) runs fully eager kernels inside the captured graph, and the Fast AR codebook decoder (`_decode_codebooks`: constrained semantic sampling + 9-step `forward_kvcached` loop) is also entirely eager.

Benchmarking from PR #153 (@Ratish1) already quantified the opportunity on this model:

| Configuration | Startup time | Steady-state throughput | Δ throughput |
|---|---|---|---|
| CUDA graph only (current) | ~33s | 88 tok/s | baseline |
| Partial compile (fast head only) | ~54s | 121 tok/s | **+37%** |
| Full-model compile | ~137s | 126 tok/s | +43% |

Partial compile captures ~85% of the total compile gain at ~20% of the startup cost. The Fast AR codebook decoder is the primary compile target — its tight loop of small matmuls, embedding lookups, and argmax steps benefits most from kernel fusion.

## Architecture context

S2-Pro uses a Dual-AR architecture:

```
Slow AR (Qwen3 24L backbone, RadixAttention, CUDA-graph-captured)
  ↓ hidden_states
_decode_codebooks:
  ├── Constrained semantic sampling (top-k/top-p with RAS + repetition penalty)
  ├── project_in → forward_kvcached(codebook_idx=0)
  └── for cb_idx in 1..9:
        forward_kvcached → argmax → embeddings → next step
  ↓ output_codes [bs, num_codebooks+1]
```

The Fast AR loop runs 9 sequential `audio_decoder.forward_kvcached()` calls per Slow AR step. Each call is a small 4-layer transformer forward — individually too small to saturate the GPU, but fuseable by torch.compile into a much more efficient kernel sequence.

## Scope

- Model: `sglang_omni/models/fishaudio_s2_pro/sglang_model.py` (`S2ProSGLangTextModel`)
- Model runner: `sglang_omni/models/fishaudio_s2_pro/model_runner.py` (`FishS2ProModelRunner`)
- Stage factory: `sglang_omni/models/fishaudio_s2_pro/stages.py` (`create_sglang_tts_engine_executor`)
- Fast AR head: `audio_decoder` (loaded in `bootstrap.py`, attached via `setup_vq_decode`)

[...]

PR #625, opened 2026-05-31
[Perf]  Enable S2-Pro Fast-AR torch-compile on decode path.

SandyLuXY:
## Motivation

Enable effective torch-compile acceleration for S2-Pro fast AR decode path and make its CLI compile controls apply to the pipeline stage.

## Modifications
1. `sglang_omni/models/fishaudio_s2_pro/config.py`
- Added `talker_sglang_role_to_stage()` override to map `talker` to `tts_engine`.

2. `sglang_omni/models/fishaudio_s2_pro/stages.py`
- Added `_compile_s2pro_codebook_decoder(model, max_batch_size=...)` helper.
- Enables S2-Pro talker compile by default with `enable_torch_compile=True` and `torch_compile_max_bs=16`.
- Compiles the audio decoder layer `forward_kvcached` bound methods (not module `forward`) during stage setup.
- Installs the compiled callable list with `audio_decoder.set_compiled_forward_kvcached_layers(..., max_batch_size=server_args.torch_compile_max_bs)` before CUDA graph capture.
- Uses `SGLANG_TORCH_COMPILE_MODE` (default: `max-autotune-no-cudagraphs`).
- Runs local S2-Pro compile when `server_args.enable_torch_compile` is set, then disables generic engine-level compile (`server_args.enable_torch_compile = False`) after local compile setup.
- Keeps the local compiled path governed by `torch_compile_max_bs`, even after generic engine-level compile is disabled for CUDA graph capture.

3. `sglang_omni/models/fishaudio_s2_pro/fish_speech/models/text2semantic/modeling.py`
- Initializes `_eager_forward_kvcached_layers` to eager `layer.forward_kvcached` callables.
- Added `FishQwen3AudioDecoder.set_compiled_forward_kvcached_layers(..., max_batch_size=...)` so stage setup can register compiled callables without putting `torch.compile(...)` in the model file.
- Keeps eager and compiled callable lists side by side.
- Dispatches once at `forward_kvcached` entry: batches `<= torch_compile_max_bs` use compiled callables, while larger batches use eager callables.
- Avoids dynamic attribute lookup/fallback in the repeated codebook loop and prevents capture batches above the cap from triggering extra local Inductor compilations.

4. `tests/unit_test/fishaudio_s2_pro/test_pipeline.py`
- Added CLI override coverage for S2-Pro talker compile flags targeting `tts_engine`.
- Added regression coverage for the stage compile helper.
- Added engine setup regression coverage that S2-Pro defaults pass `enable_torch_compile=True` and `torch_compile_max_bs=16`, the local helper receives the same cap, and generic compile is disabled before CUDA graph capture.
- Added decoder dispatch regression coverage that batches at or below the cap use compiled callables and batches above the cap use eager execution.

## Related Issues
https://github.com/sgl-project/sglang-omni/issues/571

## Accuracy Test
**Setup:**
- SeedTTS EN full set (`1088` samples), `concurrency=16`, `warmup=1`
- 3 runs each:
  - baseline: `main` @ `f1302c0`
  - compile: `s2pro-fast-dec-compile` @ `a647c78` (`--talker-torch-compile on`)

| Metric | Baseline | Compile |
| --- | ---: | ---: |
| `WER corpus` | `1.0131%` | `0.9964%` |
| `WER per-sample mean` | `0.9661%` | `0.9532%` |
| `Speaker similarity mean` | `64.9245` | `65.0996` |
| `Evaluated / skipped` | `1088 / 0` | `1088 / 0` |

## Benchmark & Profiling
Same setup as described in the _Accuracy Test_ section, 3 runs per variant. 
| Metric | Baseline (mean ± std) | Compile (mean ± std) | Delta (Compile vs Baseline) |
| --- | ---: | ---: | ---: |
| `latency_mean_s` | `16.196 ± 0.490` | `16.168 ± 0.227` | `-0.17%` |
| `latency_p95_s` | `22.058 ± 0.355` | `21.903 ± 0.183` | `-0.70%` |
| `throughput_qps` | `0.983 ± 0.031` | `0.984 ± 0.014` | `+0.10%` |
| `rtf_mean` | `4.404 ± 0.131` | `4.393 ± 0.060` | `-0.25%` |
| `audio_throughput_s_per_s` | `3.901 ± 0.119` | `3.908 ± 0.060` | `+0.19%` |
| `output_tok_per_req_s` | `76.767 ± 0.586` | `91.400 ± 0.100` | `+19.06%` |
| `output_throughput` (tok/s) | `84.000 ± 2.587` | `84.133 ± 1.274` | `+0.16%` |
| `startup_ready_s` | `55.737 ± 0.579` | `62.076 ± 9.547` | `+11.37%`* |
| `cold_first_request_s` | `8.256 ± 1.173` | `8.446 ± 0.903` | `+2.30%` |
| `completed / failed` | `1088 / 0` | `1088 / 0` | `no change` |

  *_The large std on compile `startup_ready_s` (`62.076 ± 9.547`) is dominated by one outlier run (`73.085s`): in compile run 3, the preprocessing-stage HuggingFace `Fetching 13 files` step took about `15s`, while it was near-instant in compile runs 1/2. If that outlier is excluded, compile `startup_ready_s` mean is `56.572s`._

### Summary:
- Decode-path efficiency improved strongly (`output_tok_per_req_s` +19.06%).
- End-to-end serving KPIs (`latency`, `throughput`, `rtf`) are effectively flat (within about ±1%).

## Checklist

- [x] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/models/fishaudio_s2_pro/fish_speech/models/text2semantic/modeling.py
+++ b/sglang_omni/models/fishaudio_s2_pro/fish_speech/models/text2semantic/modeling.py
@@ -948,9 +948,9 @@ class FishQwen3AudioDecoder(PreTrainedModel):
         self._eager_forward_kvcached_layers: list[
             Callable[[Tensor, Tensor, Tensor], Tensor]
         ] = [layer.forward_kvcached for layer in self.layers]
-        self._compiled_forward_kvcached_layers: list[
-            Callable[[Tensor, Tensor, Tensor], Tensor]
-        ] | None = None
+        self._compiled_forward_kvcached_layers: (
+            list[Callable[[Tensor, Tensor, Tensor], Tensor]] | None
+        ) = None
         self._compiled_forward_kvcached_max_bs = 0
         self.norm = RMSNorm(config.dim, eps=config.norm_eps)
         self.output = nn.Linear(config.dim, config.vocab_size, bias=False)
--- a/sglang_omni/models/fishaudio_s2_pro/fish_speech/models/text2semantic/modeling.py
+++ b/sglang_omni/models/fishaudio_s2_pro/fish_speech/models/text2semantic/modeling.py
@@ -1009,12 +1009,11 @@ class FishQwen3AudioDecoder(PreTrainedModel):
                 layer.attention.kv_cache.k_cache.zero_()
                 layer.attention.kv_cache.v_cache.zero_()
 
-    def compile_forward_kvcached_layers(self, *, mode: str) -> int:
-        """Compile Fast AR decoder layer calls used by the codebook loop."""
-        self._forward_kvcached_layers = [
-            torch.compile(layer.forward_kvcached, mode=mode) for layer in self.layers
-        ]
-        return len(self._forward_kvcached_layers)
+    def set_forward_kvcached_layers(
+        self,
+        forward_kvcached_layers: list[Callable[[Tensor, Tensor, Tensor], Tensor]],
+    ) -> None:
+        self._forward_kvcached_layers = forward_kvcached_layers
 
     def forward_kvcached(
         self,
@@ -1041,8 +1040,8 @@ class FishQwen3AudioDecoder(PreTrainedModel):
         # cache_seqlens: current position in cache for each batch item
         cache_seqlens = self.input_pos.expand(bsz).to(torch.int32)
 
-        for layer_call in self._forward_kvcached_layers:
-            x = layer_call(x, freqs_cis, cache_seqlens)
+        for layer in self._forward_kvcached_layers:
+            x = layer(x, freqs_cis, cache_seqlens)
 
         x = self.norm(x)
         return self.output(x)
--- a/sglang_omni/models/fishaudio_s2_pro/fish_speech/models/text2semantic/modeling.py
+++ b/sglang_omni/models/fishaudio_s2_pro/fish_speech/models/text2semantic/modeling.py
@@ -1031,9 +1031,13 @@ class FishQwen3AudioDecoder(PreTrainedModel):
         # cache_seqlens: current position in cache for each batch item
         cache_seqlens = self.input_pos.expand(bsz).to(torch.int32)
 
-        layers = getattr(self, "_compiled_layers", self.layers)
-        for layer in layers:
-            x = layer.forward_kvcached(x, freqs_cis, cache_seqlens)
+        compiled_layers = getattr(self, "_compiled_kvcached_layers", None)
+        if compiled_layers is not None:
+            for compiled_layer in compiled_layers:
+                x = compiled_layer(x, freqs_cis, cache_seqlens)
+        else:
+            for layer in self.layers:
+                x = layer.forward_kvcached(x, freqs_cis, cache_seqlens)
 
         x = self.norm(x)
         return self.output(x)
--- a/sglang_omni/models/fishaudio_s2_pro/fish_speech/models/text2semantic/modeling.py
+++ b/sglang_omni/models/fishaudio_s2_pro/fish_speech/models/text2semantic/modeling.py
@@ -1031,7 +1031,8 @@ class FishQwen3AudioDecoder(PreTrainedModel):
         # cache_seqlens: current position in cache for each batch item
         cache_seqlens = self.input_pos.expand(bsz).to(torch.int32)
 
-        for layer in self.layers:
+        layers = getattr(self, "_compiled_layers", self.layers)
+        for layer in layers:
             x = layer.forward_kvcached(x, freqs_cis, cache_seqlens)
 
         x = self.norm(x)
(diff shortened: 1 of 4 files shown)


tests/unit_test/fishaudio_s2_pro/test_pipeline.py
```diff
@@ -152,3 +158,155 @@ def test_fish_tts_accepts_default_top_k_sentinel() -> None:
...
+
+    def compiled_b(
+        x: torch.Tensor, freqs_cis: torch.Tensor, cache_seqlens: torch.Tensor
+    ) -> torch.Tensor:
+        del freqs_cis, cache_seqlens
+        seen_calls.append("b")
+        return x + 3
+
+    decoder._compiled_kvcached_layers = [compiled_a, compiled_b]
+    decoder._compiled_layers = [object()]
```
yxs: looks like a dead test

sglang_omni/models/fishaudio_s2_pro/fish_speech/models/text2semantic/modeling.py
```diff
@@ -1031,8 +1031,13 @@ def forward_kvcached(
         # cache_seqlens: current position in cache for each batch item
         cache_seqlens = self.input_pos.expand(bsz).to(torch.int32)
 
-        for layer in self.layers:
-            x = layer.forward_kvcached(x, freqs_cis, cache_seqlens)
+        compiled_layers = getattr(self, "_compiled_kvcached_layers", None)
```
yxs: This method runs ~`num_codebooks`(~10) times per semantic token in the codebook loop, and an utterance has hundreds of tokens, thousands of calls per request.
The getattr cost is negligible against the matmuls, so this is fine for perf, but "dynamic attribute + getattr fallback" is itself a smell.

yxs:
@SandyLuXY
Based on your benchmark data.
The +19% on `output_tok_per_req_s` is a real decode speedup, but the e2e metrics (latency / throughput / rtf) are flat within ±1%. That means the fast-AR decode is too small a slice of the end-to-end wall time, shaving 19% off it just gets lost in the noise, so this isn't really an e2e perf win.
That said, it's fine to land, but not a perf improvement, a real e2e speedup would need profiling more.

SandyLuXY:
@yxs Thanks for the thorough review! Addressed the comments and turned torch compile on by default for S2 pro.

sglang_omni/models/fishaudio_s2_pro/stages.py
```diff
@@ -21,6 +21,28 @@
...
+        for layer in audio_decoder.layers
+    ]
+    audio_decoder.set_forward_kvcached_layers(compiled_forward_kvcached_layers)
+    logger.info(
+        "Compiled %d Fast AR decoder layers (mode=%s)",
+        len(compiled_forward_kvcached_layers),
+        compile_mode,
+    )
+
+
```
zhaochenyang20: when a user starts S2-Pro with the default settings or explicitly passes a smaller cap such as `--talker-torch-compile-max-bs 2`, the helper still replaces every `audio_decoder.layers` `forward_kvcached` method with a compiled callable at `stages.py:34-38`. Then `stages.py:267-269` sets `server_args.enable_torch_compile` back to `False` before CUDA graph capture, so SGLang's own `compile_bs = [bs for bs in capture_bs if bs <= server_args.torch_compile_max_bs]` branch no longer applies. SGLang capture still iterates over all `capture_bs` values and pre-runs `run_once` for each batch size, while the S2-Pro inner compiled callables are already installed unconditionally on the hot path; with the default `max_running_requests=64`, `torch_compile_max_bs=16` no longer prevents capture batches such as 24, 32, 40, 48, 56, and 64 from invoking compiled Fast-AR layers. This directly breaks the PR's performance control plane: users think they limited compile batch size, but they may get extra Inductor compilations, longer startup/capture time, and higher compile-cache or memory pressure, making the benchmark behavior hard to reason about.

zhaochenyang20: Pass the batch-size cap into the local helper explicitly instead of only storing it in `ServerArgs`. For example, call `_compile_s2pro_codebook_decoder(model, max_batch_size=server_args.torch_compile_max_bs)`, keep both eager and compiled callables, and install decoder callables that use the compiled version only when `x.shape[0] <= max_batch_size`; otherwise they should call the original `layer.forward_kvcached`. If a per-layer wrapper is undesirable, store this state on `FishQwen3AudioDecoder` and choose the compiled or eager layer list at the `forward_kvcached` entry. The important invariant is that S2-Pro's local compiled path must still honor `torch_compile_max_bs` even after `server_args.enable_torch_compile` is set back to `False` to disable generic engine compile.

zhaochenyang20: Also, add a unit test.

Add two tests. First, fake `build_sglang_server_args`, `create_sglang_infrastructure`, `load_audio_decoder`, and `init_device_graphs`, then assert that S2-Pro defaults pass `enable_torch_compile=True` and `torch_compile_max_bs=16`, that the helper receives the same max-batch cap, and that the generic compile flag is disabled before graph capture. Second, add a focused unit test for `_compile_s2pro_codebook_decoder` or the decoder dispatch: call the installed callable with a batch size at or below the cap and with a batch size above the cap, then assert that the former uses the compiled callable and the latter uses eager execution.

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

sglang_omni/models/fishaudio_s2_pro/stages.py
SandyLuXY: Good catch! Fixed the batch size override issue and added test coverage. Updated PR description accordingly.

PR #638, opened 2026-06-01
feat: default Higgs TTS async-decode on with --async-decode toggle

Hayden727:
## Summary

Enable **one-step-lookahead async decode by default** for the Higgs TTS `tts_engine` stage, and replace the bool `--enable-async-decode` flag with a tri-state `--async-decode default|on|off` so users can opt out (the old flag stays as a hidden deprecated alias for `--async-decode on`). Async decode overlaps each AR step's host-side token collect with the next GPU forward; batches below `--async-decode-min-batch-size` (default 2) take the synchronous fast path, so concurrency 1 is unaffected.

The capability already existed (opt-in); this PR flips the Higgs default to on, backed by a full-set quality validation, and adds the off-switch + tests.

## Why it's safe to default on

**Quality — full SeedTTS set (EN=1088, ZH=2020), H200.** Async-on matches the synchronous path within run-to-run noise. Reporting outlier-excluded corpus WER (the raw corpus is dominated by 1–2 pathological >50%-WER samples per split; per-sample median is 0% everywhere):

| Lang | corpus WER (excl >50%) | SIM |
|------|------------------------|-----|
| EN async **on** | 1.37% | 66.1 |
| EN async **off** | 1.21% | 66.2 |
| ZH async **on** | 1.16% | 72.2 |
| ZH async **off** | 1.15% | 72.4 |

ON−OFF: ΔWER +0.16pp EN / +0.01pp ZH, ΔSIM −0.1 / −0.2 pts — all within noise, and both lanes match the in-repo baseline (PR #534: EN 1.36% excl, ZH 1.14%). Token-level parity is covered by `tests/unit_test/higgs_tts/test_async_decode_runner.py`.

**Throughput** — a controlled sweep (256 output tokens, 48 prompts, H200) measured async-on at roughly +10–15% qps / −11..−13% latency at concurrency 4/8/16, neutral at concurrency 1; indicative, not a full-set benchmark.

## Changes

- `models/higgs_tts/config.py`: `enable_async_decode: True` in the `tts_engine` `factory_args` (Higgs-scoped; global `OmniScheduler`/factory signature defaults stay `False`, so other models are untouched).
- `cli/serve.py`: bool `--enable-async-decode` → tri-state `--async-decode default|on|off`, mirroring `--talker-partial-start`. The old flag is kept as a hidden deprecated alias for `--async-decode on` (serve uses `ignore_unknown_options=True`, so removing it would break existing scripts via `parse_extra_args`); a conflicting `--async-decode off` is rejected.
- `tests/unit_test/higgs_tts/test_cli_async_decode.py`: config default-is-on (via real `resolve_stage_factory_args`), tri-state override, min-batch-size apply/`<1` reject, invalid-mode reject, unsupported-factory reject, no-op/fail-fast on configs without a `tts_engine` stage, and the deprecated-alias mapping + registration guard.

## Disable

```bash
sgl-omni serve --model-path <higgs-model> --port 8000 --async-decode off
```

## Validation

`test_cli_async_decode.py` 13 passed; `pipeline/test_async_decode.py` 16 passed; `qwen3_omni/test_cli.py` 25 passed (shared `serve.py` helper regression). pre-commit (autoflake/isort/black/ruff/ast) pass.

### DIFF
--- a/docs/cookbook/higgs_tts.md
+++ b/docs/cookbook/higgs_tts.md
@@ -46,6 +46,38 @@ sgl-omni serve \
   --port 8000
 ```
 
+### Async decode (default on)
+
+The `tts_engine` stage runs **one-step-lookahead async decode by default** for
+Higgs TTS: each AR step's host-side token collect overlaps the next GPU forward.
+Batches below `--async-decode-min-batch-size` (default 2) take the synchronous
+fast path, so concurrency 1 is unaffected.
+
+Quality is unchanged. On the full SeedTTS set (EN=1088, ZH=2020, H200) the
+async-on corpus WER and speaker similarity match the synchronous path within
+run-to-run noise: EN 1.37 vs 1.21% excl-outlier WER / SIM 66.1 vs 66.2; ZH 1.16
+vs 1.15% WER / SIM 72.2 vs 72.4. Token-level parity is covered by
+`tests/unit_test/higgs_tts/test_async_decode_runner.py`.
+
+A separate controlled throughput sweep (fixed 256 output tokens, 48 prompts,
+H200) measured async-on at roughly +10–15% qps / −11 to −13% latency at
+concurrency 4/8/16 and neutral at concurrency 1; treat those as indicative
+rather than a full-set benchmark.
+
+Toggle it with the tri-state `--async-decode` flag:
+
+```bash
+# Disable async decode (synchronous AR loop)
+sgl-omni serve --model-path <higgs-model> --port 8000 --async-decode off
+
+# Force on / use the config default (on for Higgs) — equivalent here
+sgl-omni serve --model-path <higgs-model> --port 8000 --async-decode on
+sgl-omni serve --model-path <higgs-model> --port 8000 --async-decode default
+
+# Raise the min batch size below which decode bypasses the lookahead (default 2)
+sgl-omni serve --model-path <higgs-model> --port 8000 --async-decode-min-batch-size 4
+```
+
 ## Synthesizing Speech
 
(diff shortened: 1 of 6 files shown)


Hayden727:
Addressed the CLI backward-compat finding in `f3c15fd`.

**Root cause confirmed:** `serve` is registered with `ignore_unknown_options=True` (`cli/__init__.py`), so dropping the first-class `--enable-async-decode` option made it fall through to `ctx.args` → `ConfigManager.parse_extra_args()`, which raises `Missing value for argument: --enable-async-decode` (or swallows the next arg as its value).

**Fix:** re-added `--enable-async-decode` / `--enable_async_decode` as a **hidden deprecated option** that maps to `--async-decode on`, and rejects a contradictory `--async-decode off` with a clear `BadParameter`. A deprecation warning is logged when it's used.

**Regression tests added** (`tests/unit_test/higgs_tts/test_cli_async_decode.py`):
- `test_enable_async_decode_alias_stays_registered_option` — introspects the Typer/Click command and asserts both `--async-decode` and `--enable-async-decode` are registered options, directly guarding the `parse_extra_args` fall-through.
- `_resolve_async_decode_flag` reconciliation: alias → `on`, passthrough when absent, and `--enable-async-decode` + `--async-decode off` → conflict raised.

**Validation:** `test_cli_async_decode.py` 13 passed, `qwen3_omni/test_cli.py` 25 passed (container @ f1302c0). pre-commit (autoflake/isort/black/ruff/ast) green.

zhaochenyang20 (approved):

Issue #646, opened 2026-06-01
Investigate Qwen3-ASR concurrency scaling on SeedTTS ASR CI


zhaochenyang20:
## Background

#643 

After replacing Whisper ASR with Qwen3-ASR, ASR transcription should in principle support higher concurrency. For the TTS/WER pipeline we increased ASR fanout to `32` for DP=2, but the lightweight stage-1 Qwen3-ASR correctness gate showed an unexpected behavior: on the same 20 SeedTTS EN samples, lower ASR concurrency was faster than higher concurrency.

This is interesting enough to track separately because stage 1 is currently intended to ensure Qwen3-ASR is both accurate and healthy before later TTS/WER stages rely on it.

## Current Observation

For the 20-sample SeedTTS EN ASR correctness gate:

| ASR concurrency | total time | throughput | latency mean | latency p95 | RTF mean | RTF p95 | corpus WER | max sample WER |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 32 | ~1.943s | 10.295 samples/s | 0.193s | 0.555s | 0.0409 | 0.1421 | 0.0070 | 0.0667 |
| 2 | 1.225s | 16.331 samples/s | 0.121s | 0.159s | 0.0240 | 0.0302 | 0.0069 | 0.0667 |

`ASR throughput` is computed as:

```python
throughput_samples_per_s = len(latencies_s) / wall_clock_s
```

So this is not caused by throughput being derived from latency sum. The wall-clock time itself is lower at concurrency 2 for this small test.

The WER difference is small, but the direction is also slightly worse at concurrency 32. This may just be noise, but it should be verified with repeated runs and a larger sample set.

## Questions To Answer

1. Why does `concurrency=32` produce lower throughput than `concurrency=2` on the 20-sample SeedTTS ASR gate?
2. Is Qwen3-ASR actually batching effectively under this endpoint/router path, or are requests mostly queued/serialized with additional overhead?
3. Does high concurrency affect ASR output determinism or WER, or is the observed WER delta just run-to-run noise?
4. For DP=2, what is the best ASR concurrency for:
   - small correctness gates,
   - full SeedTTS EN transcription,
   - downstream TTS WER computation?

## Possible Causes

- The 20-sample gate is too small for `concurrency=32`; it floods the router/workers and measures queueing overhead more than steady-state throughput.
[...]

PR #643, opened 2026-06-01
[Feature] Implement Qwen3-ASR and wire it up as a seed-tts WER backend (#65)

0xjeffro:
## Motivation

TTS benchmarks (and the TTS CI stages) score WER by transcribing the generated audio back to text with Whisper. This ASR pass is a major cost: Whisper's encoder effectively runs at batch size 1, so transcribing a full SeedTTS set is slow and is a meaningful chunk of TTS CI wall-clock time.

This PR (issue #65) integrates Qwen3-ASR into sglang-omni and wires it up as a WER backend for the seed-tts benchmark keeping accuracy on par with Whisper while cutting transcription time substantially, as the groundwork for switching the WER scorer (and speeding up the WER stages in CI).

## Modifications

**Qwen3-ASR model (served on sglang-omni)**
  - New `sglang_omni/models/qwen3_asr/`, ported from upstream sglang sgl-project/sglang#22073. Reuses sglang's existing `Qwen3ForCausalLM` decoder and `Qwen3OmniMoeAudioEncoder`, so it runs on our pinned sglang 0.5.8 without forking.
  - Registered in the omni model runner; serves the OpenAI-compatible `/v1/audio/transcriptions` endpoint. Verified on both 0.6B and 1.7B, EN + ZH.
  
**Benchmark integration**
  - `benchmark_tts_seedtts` gains `--qwen3-asr-port` to score WER via a running Qwen3-ASR server, alongside the existing local-Whisper / omni-Whisper-router backends. EN only for now.

## Related Issues

#65 
References upstream sgl-project/sglang#22073

## Accuracy Test

WER parity with the Whisper baseline (local, H200, EN, transcribe-only):
  - SeedTTS EN set, Higgs TTS audio (1088 samples): Qwen3-ASR-1.7B **~1.5%** vs
    Whisper **1.46%**.
  - The 20 SeedTTS reference clips used by CI stage-0: corpus WER **0.692%** (current Whisper stage-0 threshold 0.70%), passes stage-0's correctness bar locally.

## Benchmark & Profiling

Transcribe time on the SeedTTS EN set (1088 samples, local):
  - Whisper: 409s
  - Qwen3-ASR-1.7B (sequential) : 105s (~4x faster)
  - Qwen3-ASR-1.7B (client concurrency=4) : <60s 

## Checklist

- [x] Format your code according with pre-commit.
- [ ] Add unit tests.
- [ ] Update documentation / docstrings / example tutorials as needed.
[...]

PR #647, opened 2026-06-01
[CI] Align Qwen3-ASR stage-1 gate to concurrency=2 and add ASR concurrency-scaling benchmark

MelodyyyYin:
# PR for issue #646

### Root cause

The stage-1 gate already runs at `concurrency=2` (`test-tts-ci.yaml` exports
`QWEN3_ASR_CI_CONCURRENCY=2`), but its speed baselines in
`test_qwen3_asr_ci.py` were still the **concurrency=32** numbers from #646
(throughput `10.295`, latency mean `0.193`, p95 `0.555`, RTF mean `0.0409`,
RTF p95 `0.1421`). At concurrency=2 the real numbers are much better
(throughput `16.3`, latency `0.121`, …), so the speed thresholds were 2–5×
too loose to catch a regression on the workload CI actually runs.

The issue's observation that "concurrency=2 is faster than concurrency=32 on
the 20-sample gate" is expected: on such a small workload, high fan-out mostly
measures burst queueing / tail latency rather than steady-state throughput. ASR
fan-out only pays off on large workloads. WER is essentially concurrency
independent (the small delta is run-to-run noise).

### Changes

- **`tests/test_model/test_qwen3_asr_ci.py`**: recalibrate the stage-1 speed
  thresholds to the concurrency=2 measurements from #646, and make the in-code
  default concurrency `2` (was `32`) so local runs and `tune-ci-thresholds`
  calibrate against the same workload CI runs. WER bounds are unchanged
  (concurrency independent). Added comments explaining the choice.
- **`benchmarks/eval/benchmark_qwen3_asr_concurrency.py`** (new): a reproducible
  ASR fan-out sweep. Transcribes SeedTTS reference clips directly against a
  running Qwen3-ASR router across a configurable concurrency list with repeats,
  and reports corpus/per-sample WER, wall-clock, throughput, latency
  percentiles, RTF, and per-worker routing balance — the metric matrix #646
  asks for. Gives repeatable concurrency-scaling data to pick the right fan-out
  per workload.
- **`.claude/skills/tune-ci-thresholds/models/qwen3-asr-v1/config.yaml`**:
  correct the stale comment that claimed the gate runs at 32; it is a small
  20-sample gate at concurrency=2, while the full-set TTS WER stages still fan
  out at 32.
- **`benchmarks/README.md`**: document the new script and the concurrency-scaling
  finding / decision.


Billy1900:
The inversion (c=32 slower than c=2 on the 20-sample test) traces back to `process_input_requests` in `OmniScheduler` — it calls `_request_builder` in a serial for-loop, and for Qwen3-ASR each call does a full
  mel extraction (~8ms/sample on CPU). So when 10 requests land on a worker in one burst, you're paying ~80ms of CPU preprocessing before the GPU sees anything. With `disable_overlap_schedule=True` on top,
  there's no pipeline hiding that either. Details are https://github.com/sgl-project/sglang-omni/issues/646#issuecomment-4599579552

MelodyyyYin:
# PR #647 update — reuse the benchmark path in the Qwen3-ASR gate

## What the review asked for
- Our CI is essentially a reuse of the Benchmark, so the stage-1 Qwen3-ASR gate
  should not re-implement transcription/metrics — it should import them directly
  from the Benchmark.
- The ASR benchmark should be decomposed onto the existing benchmark
  abstractions (dataset → task → runner → metrics) and reuse them well.


## What changed
- `benchmarks/eval/benchmark_qwen3_asr_concurrency.py` now also defines the
  shared transcription + scoring path:
  - `make_asr_send_fn` — async `/v1/audio/transcriptions` send_fn → `RequestResult`
  - `run_asr_transcription` — runs it through the existing `BenchmarkRunner` /
    `RunConfig` (no hand-rolled thread pool)
  - `build_asr_eval_results` — WER via `calculate_wer_metrics`, speed via
    `calculate_asr_speed_metrics` + `compute_speed_metrics` (for `rtf_p95`); no
    hand-rolled percentile
  - keeps the concurrency-sweep driver, table, and JSON output
- `tests/test_model/test_qwen3_asr_ci.py` is now a thin gate that **imports**
  `run_asr_transcription` / `build_asr_eval_results` from the benchmark module,
  then launches the router, runs one `concurrency=2` pass, and applies the
  thresholds. It drops `_transcribe_with_omni`, `_percentile`,
  `_warm_up_asr_router`, and the hand-built `wer_summary` / `speed_metrics`
  dicts.


## Values are unchanged
- I did not touch any threshold. The speed reference bounds come from the earlier concurrency=2 calibration commit and the WER bounds from the original Qwen3-ASR PR (#643); both are carried over as-is. CI runs on H20 and that behavior can't be reproduced on the H200 dev box I have, so the existing values are kept.

## Verification
- Ran the refactored gate and the benchmark end-to-end on a dev GPU box to confirm they execute, share one code path, and still emit the same `qwen3_asr_results.json` schema the thresholds / tune config read.
- Dev-box speed numbers are a different hardware regime than the H20 thresholds, so they are not used to judge or change any value.
- `black` / `isort` clean; no unused imports.

PR #655, opened 2026-06-02
Add PCM audio streaming controls for Higgs TTS

### LINKED ISSUE #654: [Feature] Add raw PCM speech streaming and initial codec chunk tuning
## Summary

Add a general speech streaming capability for `/v1/audio/speech` that can return raw PCM byte streams while preserving the existing SSE/base64 streaming behavior by default. Higgs TTS should be the first model to use the new capability, including request-level `initial_codec_chunk_frames` tuning for lower TTFA.

This is a focused follow-up to #601.

## Motivation

Current speech streaming returns SSE events with base64-encoded audio chunks. That is useful for metadata and compatibility, but it is not ideal for low-latency audio playback: base64 increases payload size and clients must unwrap JSON before feeding audio to a player.

vLLM-Omni uses raw PCM chunks for TTS streaming when `stream=true` and `response_format="pcm"`. For sglang-omni, we should add that capability without breaking existing clients that already consume the SSE stream.

## Proposed API shape

Keep the existing default:

- `stream=true` continues to return `text/event-stream` SSE chunks.

Add an explicit raw audio mode:

- `stream=true`
- `response_format="pcm"`
- `stream_format="audio"`

In this mode, the response should be a continuous `audio/pcm` byte stream. Sample-rate metadata should be exposed through response headers such as `X-Sample-Rate`, `X-Channels`, and `X-Bit-Depth`.

## Initial codec chunk tuning

Add `initial_codec_chunk_frames` as a general speech request parameter. It should be carried through the request path in a model-neutral way, with model-specific schedulers deciding how to map codec-frame counts onto their decode windows.

Higgs should be the first adopter:

- use `initial_codec_chunk_frames` only for the first emitted audio chunk;
- after the first chunk, return to the existing steady-state chunk/follow-up behavior;
- keep the existing overlap/holdback behavior for follow-up chunks.

## Compatibility concerns for the PR body

Do not make `stream=true + response_format="pcm"` switch to raw PCM by default in the first PR. That would break current sglang-omni clients and docs that expect SSE events with `audio.speech.chunk` JSON and base64 payloads.

The PR should call out this tradeoff explicitly:

- SSE remains the default because it carries chunk index, finish reason, usage, and error metadata cleanly.
- Raw PCM is better for actual audio playback and avoids base64 overhead, but it has limited in-band metadata.
- `stream_format="audio"` gives us the low-latency path while preserving backwards compatibility.

## Acceptance criteria

- `/v1/audio/speech` supports `stream_format="audio"` for `stream=true + response_format="pcm"`.
- Existing SSE streaming remains unchanged when `stream_format` is omitted.
- Invalid raw streaming combinations, such as `stream_format="audio"` with non-PCM `response_format`, fail fast with a clear 4xx error.
- Higgs TTS reads `initial_codec_chunk_frames` and applies it to the first streaming vocoder chunk only.
- Unit tests cover the raw PCM serve path, compatibility fallback to SSE, invalid parameter handling, and Higgs first-chunk behavior.
- Docs mention the compatibility choice and how to request raw PCM streaming.

Hayden727:
## Summary

This PR adds raw PCM audio streaming support for `/v1/audio/speech` and wires the first-chunk tuning knob into Higgs TTS.

- Add `stream_format="audio"` for raw `audio/pcm` streaming when `stream=true` and `response_format="pcm"`.
- Derive raw PCM response metadata from the first audio chunk (`X-Sample-Rate`, `X-Channels`, `X-Bit-Depth`) instead of assuming 24 kHz, and reject mid-stream sample-rate changes.
- Add `initial_codec_chunk_frames` request plumbing and make Higgs apply it only to the first vocoder chunk.
- Keep existing SSE streaming as the default `stream=true` behavior for compatibility.
- Extend SeedTTS benchmark tooling so SSE can request `response_format=wav` or `response_format=pcm`, raw audio transport still forces PCM, and both transports can be compared with or without `initial_codec_chunk_frames`.
- Update TTS/Higgs/API docs and unit coverage.

## Compatibility Note

This intentionally follows the conservative API behavior we discussed. vLLM-Omni treats `stream=true + response_format=pcm` as raw PCM streaming by default. This PR instead requires `stream_format="audio"` so existing SGLang-Omni clients that rely on SSE under `stream=true` do not silently change behavior.

If we want exact vLLM-Omni request semantics later, the follow-up would be to make `stream=true + response_format=pcm` imply raw PCM, but that is a compatibility-sensitive API change.

## Experiment

Hardware platform: NVIDIA H200.

Model: Higgs TTS `boson-sglang/higgs-audio-v3-TTS-4B-grpo05200410999`.

Workload: full SeedTTS EN+ZH, generate-only, `concurrency=16`, `warmup=10`.

### Original full-set run

Software version: SGLang-Omni branch commit `7142f79` for the full SeedTTS measurements below; current PR head `17c5c57` adds the review fixes for dynamic raw PCM headers and the missing benchmark controls.

| Split | Mode | Success | TTFC mean | Latency mean | Throughput | First payload |
| EN | SSE default | 1087/1088 | 0.857s | 1.272s | 12.468 qps | 121668 bytes |
| EN | Raw PCM default | 1088/1088 | 0.826s | 1.232s | 12.923 qps | 116490 bytes |
| EN | Raw PCM initial=1 | 1088/1088 | 0.216s | 1.316s | 12.103 qps | 1920 bytes |
| ZH | SSE default | 2020/2020 | 0.729s | 1.398s | 11.409 qps | 122880 bytes |
| ZH | Raw PCM default | 2020/2020 | 0.730s | 1.423s | 10.831 qps | 121045 bytes |
| ZH | Raw PCM initial=1 | 2020/2020 | 0.171s | 1.512s | 10.560 qps | 1920 bytes |

The original table shows that raw PCM default is close to SSE default, while `initial_codec_chunk_frames=1` changes TTFC much more. It should not be read as evidence that PCM transport alone causes the large TTFC improvement.

### Follow-up ablation

Reviewer rerun environment: `ion9-h200-omni`, container `sglang-omni-pcm`, PR commit `7142f7991c44550ae3f74f21632b75948a5ca89b`. The same H200 server was kept up across all rows; generated audio was full SeedTTS EN/ZH with `concurrency=16`, `warmup=10`.

| Split | Mode | Success | TTFC mean | Latency mean | Throughput | Chunks mean | First payload |
| EN | SSE default | 1088/1088 | 0.8308s | 1.223s | 13.017 qps | 2.08 | 121757.6 bytes |
| EN | Raw PCM default | 1088/1088 | 0.8225s | 1.225s | 12.994 qps | 2.29 | 115758.6 bytes |
| EN | SSE initial=1 | 1088/1088 | 0.2193s | 1.324s | 12.035 qps | 2.97 | 1920.0 bytes |
| EN | Raw PCM initial=1 | 1088/1088 | 0.2094s | 1.303s | 12.230 qps | 3.67 | 1920.0 bytes |
| ZH | SSE default | 2020/2020 | 0.7368s | 1.437s | 11.098 qps | 2.54 | 122880.0 bytes |
| ZH | Raw PCM default | 2020/2020 | 0.7333s | 1.428s | 11.173 qps | 2.99 | 116303.4 bytes |
| ZH | SSE initial=1 | 2020/2020 | 0.1597s | 1.487s | 10.380 qps | 3.40 | 1920.0 bytes |
| ZH | Raw PCM initial=1 | 2020/2020 | 0.1643s | 1.483s | 10.765 qps | 4.21 | 1920.0 bytes |

Relative changes from the reviewer ablation:

| Comparison | TTFC change | Throughput change | Latency change |
|---|---:|---:|---:|
| EN raw PCM default vs SSE default | -1.00% | -0.18% | +0.16% |
| ZH raw PCM default vs SSE default | -0.48% | +0.68% | -0.63% |
| EN SSE initial=1 vs SSE default | -73.60% | -7.54% | +8.26% |
| ZH SSE initial=1 vs SSE default | -78.33% | -6.47% | +3.48% |
| EN raw PCM initial=1 vs SSE initial=1 | -4.51% | +1.62% | -1.59% |
| ZH raw PCM initial=1 vs SSE initial=1 | +2.88% | +3.71% | -0.27% |

Interpretation:

- Default raw PCM transport is functional, but it does not show a material TTFC improvement by itself in these runs.
- The large TTFC reduction is primarily attributable to `initial_codec_chunk_frames=1`, which shrinks the first Higgs vocoder emission to a 1920-byte first PCM payload.
- `initial_codec_chunk_frames=1` trades lower TTFC for more audio chunks and lower overall throughput / higher mean latency versus default SSE.
- With the same `initial_codec_chunk_frames=1` setting, raw PCM and SSE have similar TTFC; the remaining delta is small and not the dominant performance lever.
- Raw PCM remains useful as an explicit transport/format option for clients that want raw audio bytes, but the performance claim should not attribute the large TTFC gain to PCM transport alone.

## Validation

- `python -m compileall sglang_omni/serve/openai_api.py benchmarks/tasks/tts.py benchmarks/eval/benchmark_tts_seedtts.py tests/unit_test/serve/test_openai_api.py tests/unit_test/moss_tts/test_pipeline.py`
- `python -m black sglang_omni/serve/openai_api.py benchmarks/tasks/tts.py benchmarks/eval/benchmark_tts_seedtts.py tests/unit_test/serve/test_openai_api.py tests/unit_test/moss_tts/test_pipeline.py`
- `git diff --check`
- pre-commit hooks during commits
- Targeted unit tests for `/v1/audio/speech` passed in the H200 validation environment: `16 passed`
- Targeted benchmark-control tests passed in the H200 validation environment: `3 passed`
- A full run of `tests/unit_test/moss_tts/test_pipeline.py` also includes an unrelated MOSS prefill assertion failure in the validation environment, so that full-file run is not counted as validation evidence for this PR; the review-related benchmark tests pass.
- Full SeedTTS EN+ZH generate-only A/B benchmark above

## Metric Note

The earlier SSE-vs-raw-PCM ablation table should not use `Chunks mean` as an apples-to-apples cross-transport comparison when generated before `fff1ecf`: SSE counted decoded audio events, while raw PCM counted arbitrary `iter_any()` client read frames. `fff1ecf` updates the raw PCM benchmark parser to preserve HTTP chunk boundaries with `iter_chunks()`, so future raw PCM `audio_chunks_mean` and `first_audio_payload_bytes_mean` reflect HTTP body chunks instead of TCP/client read splits.

Fixes #654


Hayden727:
Update after the post-initial Higgs streaming scheduler optimization (`f493f09`).

Change: keep the 1920B first raw PCM chunk, then delay the next vocoder decode to `steady_boundary + followup_stride` so the post-first stream emits fewer small chunks.

Validation:

- `black --check sglang_omni/models/higgs_tts/vocoder_scheduler.py tests/unit_test/higgs_tts/test_pipeline.py`
- `python3 -m pytest tests/unit_test/higgs_tts/test_pipeline.py -q` -> 30 passed
- Full SeedTTS EN+ZH, H200, `concurrency=16`, `warmup=10`, raw PCM audio stream
- Artifact: `/data/perf-baselines/sglang-omni/sota/pr655-pcm-20260603T020510/results/after-0b10dbb-postinitial-followup-gpu6`

Compared with the previous PR-head full run (`after-53121b4-stage-sched`), this recovers both latency and throughput while preserving the 1920B first chunk:

| Split | Latency mean | Latency p95 | QPS | Audio throughput | TTFC mean | Chunks mean | First payload |
| EN previous | 1.286s | 1.831s | 12.379 | 54.154 s/s | 0.2483s | 3.14 | 1920B |
| EN latest | 1.239s | 1.729s | 12.861 | 56.931 s/s | 0.2342s | 2.71 | 1920B |
| EN delta | -3.65% | -5.57% | +3.89% | +5.13% | -5.68% | -13.69% | 0% |
| ZH previous | 1.466s | 1.932s | 10.886 | 64.039 s/s | 0.1683s | 3.72 | 1920B |
| ZH latest | 1.421s | 1.910s | 11.232 | 66.068 s/s | 0.1621s | 3.32 | 1920B |
| ZH delta | -3.07% | -1.14% | +3.18% | +3.17% | -3.68% | -10.75% | 0% |

Main comparison below reuses the already collected main baseline from this campaign (`before-0e54f35-clean`) as requested; no new main run was launched.

| Split | Mode | Latency mean | Latency p95 | QPS | Audio throughput | TTFC mean | Chunks mean |
| EN | main | 1.258s | 1.723s | 12.654 | 57.790 s/s | 0.8194s | 2.75 |
| EN | latest PR | 1.239s | 1.729s | 12.861 | 56.931 s/s | 0.2342s | 2.71 |
| EN | delta | -1.51% | +0.35% | +1.64% | -1.49% | -71.42% | -1.45% |
| ZH | main | 1.410s | 1.852s | 11.316 | 66.567 s/s | 0.7282s | 3.28 |
| ZH | latest PR | 1.421s | 1.910s | 11.232 | 66.068 s/s | 0.1621s | 3.32 |
| ZH | delta | +0.78% | +3.13% | -0.74% | -0.75% | -77.74% | +1.22% |

Takeaway: the latest PR head keeps the large TTFC win from the 1920B first chunk and largely removes the earlier latency/throughput regression. EN is slightly faster than main on latency/QPS; ZH is still a very small throughput/latency regression versus main, but much smaller than before, with ~78% lower TTFC.

yxs:
@Hayden727 over all looks good.
just one thing,
`audio_chunk_count` isn't comparable across SSE vs raw PCM

The metric counts different units per transport, so the "Chunks mean" column in the A/B table isn't apples-to-apples?

- SSE: `len(pcm_chunks)` where each element is one decoded SSE audio event — a server-defined logical chunk.
- Raw PCM: `pcm_chunks` comes from `response.content.iter_any()`, HTTP read framing. A single large server write (default first payload ~115KB) is split across multiple TCP reads, inflating the count.

That should be why raw PCM "Chunks mean" is systematically higher (EN initial=1: 3.67 vs 2.97).

Hayden727:
@yxs I checked this carefully. Your `e657dc1` cleanup is correct, but it was orthogonal to the chunk-count concern: it only removed the dead duplicate `metadata["tts_params"]["initial_codec_chunk_frames"]` write while keeping the real `extra_params` path.

Fixed the benchmark metric issue in `fff1ecf`:

- raw PCM streaming now uses `response.content.iter_chunks()` and accumulates until `end_of_http_chunk`, instead of counting arbitrary `iter_any()` read fragments;
- `audio_chunk_count` now counts raw PCM HTTP body chunks when boundaries are available;
- `first_audio_payload_bytes` now uses the full first HTTP chunk, not the first client read fragment;
- added `test_tts_benchmark_raw_pcm_uses_http_chunk_boundaries`, which simulates one HTTP chunk split across multiple reads and verifies it still counts as one chunk;
- updated the metric docs and PR body note to clarify that older SSE-vs-raw-PCM tables generated before this fix should not compare `Chunks mean` across transports.

Validation:

- pre-commit hooks passed on commit
- `black --check benchmarks/tasks/tts.py benchmarks/metrics/performance.py tests/unit_test/moss_tts/test_pipeline.py`
- `isort --check-only benchmarks/tasks/tts.py benchmarks/metrics/performance.py tests/unit_test/moss_tts/test_pipeline.py`
- targeted parser tests: 3 passed

Hayden727:
Follow-up after `fff1ecf` raw PCM HTTP chunk metric fix.

I reran the full SeedTTS raw PCM benchmark in the original validation environment as requested:

- GPU: H200 
- Workload: full SeedTTS EN+ZH, `generate-only`, `concurrency=16`, `warmup=10`, `--stream --stream-format audio --response-format pcm --ref-format references`
- Artifact: `/data/perf-baselines/sglang-omni/sota/pr655-pcm-20260603T020510/results/after-fff1ecf-http-chunk-metrics-gpu6`
- Validation before rerun: targeted chunk-boundary and Higgs follow-up scheduler tests passed (`2 passed`)

Current PR head is `7cd3e48`, which is only a merge from `main` containing docs/logo changes. The serving/benchmark code validated here is the `fff1ecf` metric fix plus the earlier performance changes.

Does the metric fix affect performance numbers? No material latency/throughput regression showed up. The expected change is `Chunks mean`, because raw PCM now counts preserved HTTP body chunks instead of arbitrary `iter_any()` client read fragments.

### Metric-fix rerun vs previous optimized run

| Split/row | Latency mean | Latency p95 | QPS | Audio throughput | TTFC mean | Chunks mean | First payload |
| EN previous optimized (pre-fix metrics) | 1.239s | 1.729s | 12.861 | 56.931 s/s | 0.2342s | 2.71 | 1920B |
| EN after `fff1ecf` metric fix | 1.215s | 1.760s | 13.105 | 56.195 s/s | 0.2413s | 2.13 | 1920B |
| EN delta | -1.94% | +1.79% | +1.90% | -1.29% | +3.03% | -21.40% | +0.00% |
| ZH previous optimized (pre-fix metrics) | 1.421s | 1.910s | 11.232 | 66.068 s/s | 0.1621s | 3.32 | 1920B |
| ZH after `fff1ecf` metric fix | 1.421s | 1.855s | 11.236 | 67.102 s/s | 0.1549s | 2.56 | 1920B |
| ZH delta | +0.00% | -2.88% | +0.04% | +1.57% | -4.44% | -22.89% | +0.00% |

### Current PR vs reused main baseline

Main baseline below reuses the already collected campaign baseline (`before-0e54f35-clean`); no new main run was launched.

| Split/row | Latency mean | Latency p95 | QPS | Audio throughput | TTFC mean | Chunks mean | First payload |
| EN main baseline | 1.258s | 1.723s | 12.654 | 57.790 s/s | 0.8194s | 2.75 | 91980B |
| EN current PR | 1.215s | 1.760s | 13.105 | 56.195 s/s | 0.2413s | 2.13 | 1920B |
| EN delta | -3.42% | +2.15% | +3.56% | -2.76% | -70.55% | -22.55% | -97.91% |
| ZH main baseline | 1.410s | 1.852s | 11.316 | 66.567 s/s | 0.7282s | 3.28 | 93090B |
| ZH current PR | 1.421s | 1.855s | 11.236 | 67.102 s/s | 0.1549s | 2.56 | 1920B |
| ZH delta | +0.78% | +0.16% | -0.71% | +0.80% | -78.73% | -21.95% | -97.94% |

Takeaway: `fff1ecf` changes the raw PCM benchmark accounting, not the serving hot path. Latency/QPS remain within normal run variance versus the previous optimized run, while the chunk count drops because it now reflects HTTP chunk boundaries. Versus the reused main baseline, the PR still keeps the ~70-79% TTFC reduction with no material throughput regression: EN QPS is higher, ZH QPS is -0.71%, and ZH audio throughput is slightly higher.

yxs (approved):

PR #659, opened 2026-06-02
[TTS] Add OpenAI-compatible speech API

Ratish1:
## Motivation

Part of #601.

This adds the OpenAI-compatible `/v1/audio/speech` API surface for TTS serving. The change is scoped to speech synthesis request parsing, HTTP/SSE response handling, request lowering, model-facing streaming controls, and documentation. Uploaded voice lifecycle, batch speech, and WebSocket speech APIs are left to the follow-up API branches.

## Architecture

```mermaid
flowchart LR
    C[Client] --> A[POST /v1/audio/speech]
    A --> V[CreateSpeechRequest validation]
    V --> S[SpeechService]
    S --> G[GenerateRequest]
    G --> P[Pipeline client]
    P --> M[TTS model stages]
    M --> R[Audio response]
    R --> C
```

`SpeechService` owns OpenAI-compatible request validation and converts speech requests into the existing `GenerateRequest` path. Non-streaming requests return encoded audio in the requested response format. Streaming speech is constrained to PCM and uses SSE chunks so the API can expose incremental audio without changing the pipeline client contract.

## Modifications

- Added `CreateSpeechRequest` fields for the speech API, including TTS text, voice alias, response format, reference audio/text, generation parameters, and streaming controls.
- Added `SpeechService` as the validation and lowering layer for `/v1/audio/speech`. It normalizes response formats, task type, language, local/data/HTTP media references, and OpenAI-style validation errors before building `GenerateRequest`.
- Added `/v1/audio/speech` handling in the OpenAI-compatible server for non-streaming audio responses and SSE streaming responses.
- Added audio encoding support for the speech response formats used by the API path.
- Wired `initial_codec_chunk_frames` from the speech API into `GenerateRequest.extra_params`, Higgs stream metadata, and the Higgs streaming vocoder so the first streamed audio chunk can be controlled by the request instead of being only documented metadata.
- Added `--allowed-local-media-path` plumbing so `file://` reference audio can be enabled explicitly by deployments instead of being accepted globally.
- Updated TTS user and developer docs for the speech endpoint, request parameters, local media allowlisting, and model integration expectations.
- Added targeted API/protocol tests for request validation, error responses, streaming format enforcement, reference validation, and request lowering.

## Accuracy Tests

- `.venv/bin/python -m pytest tests/unit_test/serve/test_speech_protocol.py tests/unit_test/serve/test_openai_api.py -q`
  - Result: `39 passed`.
- Higgs non-streaming speech curl:
  - `curl -X POST http://127.0.0.1:8000/v1/audio/speech -H "Content-Type: application/json" -d '{"input":"Hello from SGLang Omni.","response_format":"wav"}' --output speech.wav`
  - Result: request completed successfully, downloaded ~108 KiB, and the output audio was confirmed working.
- Higgs SSE streaming speech curl:
  - `curl -N -X POST http://127.0.0.1:8000/v1/audio/speech -H "Content-Type: application/json" -d '{"input":"Hello from SGLang Omni streaming.","stream":true,"response_format":"pcm"}'`
  - Result: streamed `audio.speech.chunk` events with `format="pcm"`, `mime_type="audio/pcm"`, `sample_rate=24000`, terminal usage metadata (`prompt_tokens=11`, `completion_tokens=73`, `total_tokens=84`, `engine_time_s=0.369473`), and `data: [DONE]`.
- Higgs local file reference curl with `--allowed-local-media-path`:
  - `POST /v1/audio/speech` using `file://$PWD/docs/_static/audio/gaokao-listening.wav` as `ref_audio`.
  - Result: returned `200 OK` and produced `speech_file_ref.wav` (~140 KiB).

## Checklist


sglang_omni/serve/openai_api.py
```diff
@@ -612,6 +666,82 @@ async def _speech_stream(
...
+    while True:
+        try:
+            disconnected = await asyncio.wait_for(
+                request.is_disconnected(),
+                timeout=HTTP_DISCONNECT_POLL_INTERVAL_S,
+            )
+        except asyncio.TimeoutError:
+            disconnected = False
+        if disconnected:
+            return
```
JingwenGu0829: Do you think we should add a real sleep/yield in this loop when `is_disconnected()` returns `False`? In the current code, an immediately-returning `False` can make this task spin continuously, and then `_await_speech_response()` never returns even after `client.speech()` has completed.  A small fix would be to sleep for `HTTP_DISCONNECT_POLL_INTERVAL_S` on the false path, or otherwise make the disconnect watcher block/yield between polls.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,430 @@
...
+                for reference in request.references
+            )
+        if request.ref_audio is not None:
+            ref: dict[str, Any] = {"audio_path": request.ref_audio}
+            if request.ref_text is not None:
+                ref["text"] = request.ref_text
+            references.append(ref)
+        if references:
+            prompt = {"text": request.input, "references": references}
+
```
JingwenGu0829: [Codex_detected] Could we keep `data:` reference audio in a representation the model loaders can decode? Right now the service accepts a base64 data URL, but then lowers top-level `ref_audio` into `{"audio_path": <data-url>}`. Higgs later treats that as a URL string and hands it to the HTTP client, which rejects `data:` URLs. I would suggest converting data URLs into the existing dict/base64 shape, or updating the Higgs loader to handle `data:` through the media connector, plus a small regression test for a data URL reference.

JingwenGu0829 (commented):
Two small things that might be worthy to fix.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,430 @@
...
+                if not isinstance(payload[field_name], bool):
+                    raise bad_request(
+                        f"{field_name} must be a boolean", param=field_name
+                    )
+
+    def _normalize_speech_reference(
+        self, reference: SpeechReference
+    ) -> SpeechReference:
+        updates: dict[str, Any] = {}
+        for field_name in ("audio_path", "ref_audio", "audio"):
```
JiaxinD: **`SpeechReference` alias fields `ref_audio`/`audio` are validated but silently dropped for S2-Pro/Higgs.**

`_normalize_speech_reference` normalizes whichever of `audio_path`/`ref_audio`/`audio` is set but writes the value back under the **same key**, and `build_generate_request` then emits each reference via `reference.model_dump(exclude_none=True)`. So a reference sent as `{"ref_audio": "..."}` or `{"audio": "..."}` reaches the pipeline under that key, but the backends diverge:

* `fishaudio_s2_pro/stages.py` reads only `ref_data.get("audio_path")`, producing an empty `Reference`.
* `higgs_tts/stages.py` sets `reference_audio` only from `bytes`/`base64`/`data` or `audio_path`/`path`, so it stays `None`.

Net effect: a request using the newly added per-reference aliases is accepted and validated, then silently drops the reference for S2-Pro/Higgs, and voice cloning falls back to zero-shot with **no error**. (The top-level `ref_audio` field is fine; it is canonicalized to `audio_path`. Only the `SpeechReference` aliases are affected.)

Suggest canonicalizing into `audio_path` during normalization so every backend sees one key:

```python
def _normalize_speech_reference(self, reference: SpeechReference) -> SpeechReference:
    canonical = reference.audio_path or reference.ref_audio or reference.audio
    updates: dict[str, Any] = {"ref_audio": None, "audio": None}
    if isinstance(canonical, str):
        updates["audio_path"] = self._normalize_media_reference(canonical, param="references.audio_path")
    return reference.model_copy(update=updates)
```

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,430 @@
...
+            value = getattr(reference, field_name)
+            if isinstance(value, str):
+                updates[field_name] = self._normalize_media_reference(
+                    value, param=f"references.{field_name}"
+                )
+        return reference.model_copy(update=updates)
+
+    def _normalize_media_reference(self, value: str, *, param: str) -> str:
+        url = urlparse(value)
+        if url.scheme in {"http", "https"}:
```
JiaxinD: **SSRF: http(s) `ref_audio` URLs are fetched server-side with no host/domain allowlist.** (This is pre-existing, since `references` already accepted URLs, but this PR is the one hardening media references, so I am flagging it here.)

`_normalize_media_reference` passes any http/https URL straight through with no validation. Downstream, `higgs_tts/utils.py::load_audio_to_24k` issues `global_http_connection.get_sync_client().get(src)`, and the shared sync client is built with `follow_redirects=True`. The existing `allowed_media_domains` guard (`MultiModalResourceConnector._assert_url_allowed`) is **never applied** on this path. (`fishaudio_s2_pro/stages.py` likewise feeds the URL into `torchaudio.load`, which can fetch via the ffmpeg backend.)

A client can thus make the server GET internal/metadata endpoints, e.g. `ref_audio="http://169.254.169.254/latest/meta-data/..."`, `http://localhost:8000/`, `http://10.0.0.5:6379/`. The decoder usually fails to parse the response, but the request is still sent: blind SSRF (internal port scan, service trigger, timing oracle), and `follow_redirects=True` lets a public URL 302-pivot into the internal network. The PR hardens `file://` but leaves http(s) open.

Suggested fix: an opt-in `--allowed-media-domains` mirroring `--allowed-local-media-path`, rejecting refs whose host is not allowlisted (and resolves to a non-private/non-link-local IP). At minimum route TTS reference fetching through `MultiModalResourceConnector` so the existing domain check applies, and disable redirects (or re-validate the host after each redirect).

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,430 @@
...
+        return self.prepare_request(request)
+
+    def prepare_request(self, request: CreateSpeechRequest) -> CreateSpeechRequest:
+        """Validate and normalize a request that was already parsed."""
+
+        updates: dict[str, Any] = {}
+
+        input_text = request.input
+        if not isinstance(input_text, str) or not input_text.strip():
+            raise bad_request("input must be a non-empty string", param="input")
```
JiaxinD, line 107: **No size/length limits on `input` or reference audio (DoS / memory exhaustion).**

`prepare_request` only rejects an empty `input` (`CreateSpeechRequest.input` is a bare `str` with no `max_length`), and reference-audio fetching buffers and decodes the entire payload before any size check:

* Sync fetch (`models/higgs_tts/utils.py`): `io.load_bytes(response.content)` buffers and decodes the whole body, with no byte cap and no per-call timeout.
* Async fetch (`preprocessing/resource_connector.py`): `data = response.content` is fully buffered; the `len(data) > 1MB` branches only `logger.debug(...)`.
* The `file://` and `data:` loaders validate path/scheme but never check `st_size` / payload size before decoding.

So a multi-GB remote URL, a giant allowed `file://`, or a huge `data:` URL is fully fetched and decoded into memory before the higgs-specific `_MAX_REF_AUDIO_SEC=100` cap (which only bounds the retained waveform) can fire. Combined with the SSRF surface, an attacker can also point the server at a slow/large endpoint to tie up the fetch thread pool.

Suggested fix: add a configurable `max_length` to `input` (OpenAI caps at 4096 chars with a 400 `string_above_max_length`); cap reference-audio bytes by streaming and aborting past a threshold (e.g. `httpx` `client.stream(...)` plus a running size check) instead of reading `response.content` unconditionally; apply the same cap to `data:` and to `file://` via `st_size`.

sglang_omni/serve/openai_api.py
```diff
@@ -612,6 +666,82 @@ async def _speech_stream(
...
+
+
+async def _prepend_speech_stream_event(
```
JiaxinD: **SSE prefetch wrapper leaks the coordinator stream queue/future on client disconnect.**

The streaming path prefetches `first_event = await anext(speech_events)` (which admits the request and registers `_stream_queues[request_id]` plus `_completion_futures[request_id]`), then hands `_prepend_speech_stream_event(first_event, speech_events)` to `StreamingResponse`. This wrapper drives the inner generator via a bare `async for` with no `try/finally`.

When Starlette calls `.aclose()` on the wrapper after a client disconnect, `GeneratorExit` unwinds the wrapper frame **without** closing the `speech_events` iterator the `async for` was consuming (an `async for` does not `aclose()` its iterator on `GeneratorExit`). So `Coordinator.stream`'s `finally`, the only place that pops `_stream_queues`/`_completion_futures`, never runs deterministically, and `_handle_stream` keeps decoding and pushing audio into the orphaned **unbounded** `asyncio.Queue` for a gone client until GC collects it. The chat path is unaffected because it passes `_chat_stream` directly to `StreamingResponse`.

Fix: make the wrapper deterministically close the inner generator.

```python
async def _prepend_speech_stream_event(first_event, stream):
    try:
        yield first_event
        async for event in stream:
            yield event
    finally:
        await stream.aclose()   # idempotent; safe on normal exhaustion too
```

Same idea for the `anext()` prefetch error branch in `create_speech`: call `await speech_events.aclose()` before returning the error response, so cleanup is deterministic if the first chunk raises.

JiaxinD (commented):
Reviewed the speech-service refactor, SSE streaming, and `file://` allowlisting. Overall solid: clean service/error split, thorough validation, good tests, and the `file://` allowlist (`resolve()` plus `relative_to()`, with a symlink-escape test) looks correct. Higher-impact items are inline; the rest are below.

### High, but the affected file is not in this PR's diff
**Doc-integration tests will 400 under the new `stream => pcm` rule.** `prepare_request` raises `bad_request("stream=true requires response_format='pcm'")` before any SSE is emitted, but the existing `tests/docs/s2pro/test_docs_tts_s2pro.py` is not updated:

* `test_voice_cloning_streaming` posts `{"stream": True}` (defaults to `wav`), so `response.raise_for_status()` fails on 400.
* `test_voice_cloning_streaming_wav_reassembly` posts `{"stream": True, "response_format": "wav"}`, so it fails on 400; it also still decodes each SSE chunk as WAV, but the emitted format is now PCM.

These run in the docs/nightly lane, so unit CI stays green, but the docs suite regresses. Please update both to send `"response_format": "pcm"` and reassemble raw `base64.b64decode(...)` chunks (`setnchannels(1)/setsampwidth(2)/setframerate(24000)`), mirroring the updated `tts.md`. A negative test asserting `stream=true` plus `wav` returns 400 would lock in the contract.

### Low
* **Encoder check ordering** (`prepare_request`): `_validate_encoder_dependency` runs before the stream/format check, so `stream=true` plus `mp3` on a host without pydub returns 500 instead of the more accurate 400. Move the stream guard above the encoder check.
* **Error-envelope `type`/`code` are non-conforming** (`speech_errors.py`): `type` is `"BadRequestError"`/`"InternalServerError"` (the SDK exception class names) rather than OpenAI's category strings `invalid_request_error`/`server_error`, and `code` is the HTTP status integer rather than a string slug or `null`. This will not break the OpenAI SDK (it keys on HTTP status), but a consumer branching on the documented taxonomy silently misses. Suggest `error_type="invalid_request_error"`/`"server_error"` and `code=None`.
* **SSE stream never aborts the pipeline on client disconnect** (`_speech_stream`): no `try/finally`, so a disconnect never calls `client.abort(request_id)`, and the AR scheduler plus vocoder keep generating to `max_new_tokens` for a dead connection. This is pre-existing (it also affects `_chat_stream`), so feel free to defer, but a `finally: if not completed: await client.abort(request_id)` would stop the waste. (The realtime WS path already does this.)
* **Missing optional encoder returns 500** (`_validate_encoder_dependency`): a client picking an advertised format (`flac`/`mp3`/`aac`/`opus`) on a server without the codec dep gets a 500 `InternalServerError`, i.e. a static capability limitation reported as a server bug. A 4xx (e.g. 422) or 503 is more actionable. This branch is also untested.
* **Non-streaming compressed encode blocks the event loop** (optional, pre-existing): `mp3`/`aac`/`opus` run pydub then ffmpeg synchronously inside `client.speech`. This PR makes them a no-fallback hard path, so ffmpeg is now guaranteed to run. Consider `await asyncio.to_thread(encode_audio, ...)`.
* **Docs streaming example hardcodes 24000 Hz** (`docs/basic_usage/tts.md`): the SSE chunk carries `sample_rate`, but the example hardcodes `setframerate(24000)`. This guide is built around S2-Pro, which emits 44100 Hz, so copying it verbatim yields audio at roughly 0.54x speed (nearly an octave down). Read `audio.get("sample_rate", ...)` from the chunk instead.

### Nits
* `openai_api.py`: `_build_speech_generate_request = build_speech_generate_request` is now dead code (its only consumer, the assert in `test_openai_api.py`, was removed). Delete it.
* `_normalize_media_reference`: for `references[].audio_path` violations the envelope reports `param=references.audio_path` but the message text hardcodes `ref_audio`. Interpolate `param` so the two agree.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,430 @@
...
+    return resolved
+
+
+def build_speech_generate_request(
+    req: CreateSpeechRequest,
+    default_model: str,
+) -> GenerateRequest:
+    """Compatibility wrapper for existing callers."""
+
+    return SpeechService(default_model=default_model).build_generate_request(req)
```
Ccyest: build_speech_generate_request() creates a new SpeechService per call. The only non-test import is in openai_api.py:71, but it's only used for the dead-code alias _build_speech_generate_request (line 765) — the endpoint uses app.state.speech_service. The only real callers are 3 tests.

Suggestion: Delete the build_speech_generate_request wrapper and its import + alias in openai_api.py. Update the 3 tests to construct SpeechService(default_model=...).build_generate_request(req) directly.

sglang_omni/serve/speech_service.py
Ccyest, line 1: Nit Open Question:
SpeechService sounds like the speech synthesis service itself, but it only validates requests and builds GenerateRequest —  70% of its code is validation logic. --> so personally would suggest to rename it to `SpeechRequestValidator` or something similar

sglang_omni/serve/speech_errors.py
```diff
@@ -0,0 +1,93 @@
...
+
+    return JSONResponse(
+        status_code=status_code,
+        content=openai_error_payload(
+            message,
+            error_type=error_type,
+            param=param,
+            code=status_code if code is None else code,
+        ),
+    )
```
Ccyest, line 63: Nit: `openai_error_payload` / `openai_error_response` are generic OpenAI error envelope builders — nothing speech-specific. When ASR or chat endpoints need the same structured error format, consider promoting this module to `serve/errors.py` (or similar) so all endpoints share one implementation. 

But it is okay to leave it now with some comment for the refactor work coming up.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,430 @@
...
+        """Validate and normalize a request that was already parsed."""
+
+        updates: dict[str, Any] = {}
+
+        input_text = request.input
+        if not isinstance(input_text, str) or not input_text.strip():
+            raise bad_request("input must be a non-empty string", param="input")
+
+        response_format = _normalize_response_format(request.response_format)
+        _validate_encoder_dependency(response_format)
```
xinlij: This check is static — both `importlib.util.find_spec` and `shutil.which` results do not change at runtime. Consider moving `_validate_encoder_dependency` to `SpeechService.__init__` so it runs once at startup rather than on every request.

sglang_omni/serve/openai_api.py
```diff
@@ -559,46 +601,58 @@ async def _speech_stream(
...
+                response_format=response_format,
+                sample_rate=sample_rate,
+                speed=speed,
+                allow_format_fallback=False,
+            )
+            if not audio_bytes:
+                continue
+            actual_format = MIME_TO_FORMAT.get(mime_type, response_format)
+            payload = {
+                "id": f"speech-{request_id}",
```
xinlij: Pre-existing bug (not introduced by this PR, but worth fixing): `request_id` is already `f"speech-{uuid.uuid4()}"`, so `f"speech-{request_id}"` on lines 636 and 658 produces `"speech-speech-<uuid>"` in every SSE event. Both lines should just use `request_id` directly.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,430 @@
...
+            )
+        if request.ref_audio is not None:
+            ref: dict[str, Any] = {"audio_path": request.ref_audio}
+            if request.ref_text is not None:
+                ref["text"] = request.ref_text
+            references.append(ref)
+        if references:
+            prompt = {"text": request.input, "references": references}
+
+        extra_params: dict[str, Any] = {}
```
xinlij: Since `request.initial_codec_chunk_frames` is already stored within `tts_params`, can it be used directly, thereby eliminating the need for a dedicated `extra_params` field solely to store `request.initial_codec_chunk_frames` again?

xinlij (changes requested):

sglang_omni/serve/openai_api.py
```diff
@@ -104,6 +119,12 @@ def create_app(
     app.state.client = client
     app.state.model_name = model_name or "sglang-omni"
     app.state.realtime_enabled = enable_realtime
+    app.state.speech_service = SpeechService(
+        default_model=app.state.model_name,
+        allowed_local_media_paths=(
+            [allowed_local_media_path] if allowed_local_media_path else None
+        ),
+    )
```
WilsonZheng0327, line 143: allowed_local_media_path seems to be singular by convention (`preprocessing/resource_connector.py` uses singular as well), so if the CLI only takes in one, is there a need to convert this to a list? SpeechService loops through this list to check validity, maybe can just check the singular one.

sglang_omni/serve/protocol.py
```diff
@@ -5,7 +5,7 @@
 
 from typing import Any
 
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import AliasChoices, BaseModel, ConfigDict, Field
 
 # ---------------------------------------------------------------------------
```
yxs: lets remove these later

WilsonZheng0327:
Routing TTS references through `MultiModalResourceConnector.load_resource` might be a good idea. It applies `_assert_url_allowed`, enforces containment at load time and not only at admission in `serve/speech_service.py::_normalize_media_reference`, and removes the duplicate containment logic.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,430 @@
...
+            if isinstance(value, str):
+                updates[field_name] = self._normalize_media_reference(
+                    value, param=f"references.{field_name}"
+                )
+        return reference.model_copy(update=updates)
+
+    def _normalize_media_reference(self, value: str, *, param: str) -> str:
+        url = urlparse(value)
+        if url.scheme in {"http", "https"}:
+            return value
```
zhaochenyang20: `SpeechService._normalize_media_reference` returns any `http`/`https` URL unchanged. The Higgs path later fetches that string in `models/higgs_tts/utils.py` through the shared httpx client, whose connector enables `follow_redirects=True` in `preprocessing/resource_connector.py`. This path does not use the existing `MultiModalResourceConnector` allowed-domain checks and does not reject private, link-local, or localhost destinations. A caller can make the server request `169.254.169.254`, localhost, internal services, or a public URL that redirects into the internal network. Even if decoding fails later, the request has already been sent, which is blind SSRF and an internal probing surface; the risk is internal service triggering, metadata probing, and continued backend buffering of large responses.

Route TTS reference audio fetching through the allowlisted media connector, or add an explicit `--allowed-media-domains` / `--allowed-media-hosts` knob. Resolve the hostname and reject private, link-local, loopback, and reserved IPs; either disable redirects or re-validate every hop; stream the HTTP body and abort once a byte limit is exceeded. Do not pass arbitrary user URLs directly to model loaders.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,430 @@
...
+
+        prompt: Any = request.input
+        references: list[dict[str, Any]] = []
+        if request.references:
+            references.extend(
+                reference.model_dump(exclude_none=True)
+                for reference in request.references
+            )
+        if request.ref_audio is not None:
+            ref: dict[str, Any] = {"audio_path": request.ref_audio}
```
zhaochenyang20: The top-level `ref_audio` accepts `data:audio/...;base64,...`, and `_validate_data_url` accepts valid base64. But `build_generate_request` then writes it as `{"audio_path": request.ref_audio}`. The Higgs loader treats the string as `_is_url(src)` and calls httpx `get(src)`; a `data:` URL is not an HTTP URL and fails in practice. The API docs and validation claim data URL support, but the real request is not lowered into a representation model loaders can decode. That breaks the #601 base64-reference happy path.

Convert `data:` URLs at the speech-service boundary into the dict/base64 shape that backends already support, for example `{media_type, data}`, instead of putting them in `audio_path`. Alternatively, make backend loaders parse `data:` through the media connector. Add a regression test asserting that `ref_audio=data:audio/wav;base64,...` produces a `GenerateRequest.prompt.references[0]` that is not an `audio_path` string and is decodable by Higgs/Qwen.

sglang_omni/serve/openai_api.py
```diff
@@ -612,6 +666,74 @@ async def _speech_stream(
     yield f"data: {STREAM_DONE_SENTINEL}\n\n"
 
 
+async def _await_speech_response(
```
yxs, line 907: We can consider collapse the three cleanup sites into a single `finally` that cancels+drains both tasks once via `asyncio.gather`; keep the dual abort path (abort the backend both on disconnect and on external cancel) behind one `aborted` flag.
This preserves the "external cancel must also abort" semantics while leaving exactly one cleanup site.

sglang_omni/serve/openai_api.py
```diff
@@ -612,6 +666,74 @@ async def _speech_stream(
...
+    finally:
+        if not disconnect_task.done():
+            disconnect_task.cancel()
+            with suppress(asyncio.CancelledError):
+                await disconnect_task
+
+
+async def _wait_for_request_disconnect(request: Request) -> None:
+    while not await request.is_disconnected():
+        await asyncio.sleep(HTTP_DISCONNECT_POLL_INTERVAL_S)
```
zhaochenyang20, line 970: `_wait_for_request_disconnect` loops around `await asyncio.wait_for(request.is_disconnected(), timeout=0.05)`, but when `is_disconnected()` immediately returns `False` there is no sleep or yield. Starlette/FastAPI `is_disconnected()` can return false immediately, which lets the watcher task monopolize the event loop. `_await_speech_response()` waits for `FIRST_COMPLETED` between the speech task and the watcher, so even a completed speech task may not resume. I tried running `tests/unit_test/serve/test_openai_api.py::test_speech_response_disconnect_aborts_active_request`, and it hung on this head, matching this path; the production risk is that normal non-streaming speech requests can fail to return because the watcher spins.

Explicitly `await asyncio.sleep(HTTP_DISCONNECT_POLL_INTERVAL_S)` on the false path, or replace this with a blocking request-receive/polling implementation. Keep the `client.abort(request_id)` behavior on disconnect, and add a test proving `_await_speech_response()` returns after the speech task completes when `is_disconnected()` returns false immediately.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,430 @@
...
+        if not isinstance(payload, dict):
+            raise bad_request("speech request body must be a JSON object")
+        self._validate_raw_payload(payload)
+        try:
+            request = CreateSpeechRequest.model_validate(payload)
+        except ValidationError as exc:
+            raise bad_request(_validation_error_message(exc)) from exc
+        return self.prepare_request(request)
+
+    def prepare_request(self, request: CreateSpeechRequest) -> CreateSpeechRequest:
```
yxs, line 92: `prepare_request` normalizes a `file://` reference into a bare absolute path (`_normalize_media_reference` returns `str(file_path)`, no scheme).
But that same `_normalize_media_reference`, given a bare path, falls into the `if url.scheme != "file"` branch and `raise bad_request("must be an http, https, data, or file:// URL")`.
So **calling `prepare_request` twice on the same file:// request fails on the second call**, it is not idempotent.

sglang_omni/serve/openai_api.py
```diff
@@ -612,6 +666,74 @@ async def _speech_stream(
...
+
+
+async def _prepend_speech_stream_event(
+    first_event: str,
+    stream: AsyncIterator[str],
```
zhaochenyang20: The streaming route prefetches the first event with `anext(speech_events)` and passes `_prepend_speech_stream_event(first_event, speech_events)` to `StreamingResponse`. That wrapper only yields the first event and then `async for event in stream`; it has no `finally: await stream.aclose()`. If the client disconnects at the wrapper layer, Starlette closes the wrapper, but the inner `speech_events` iterator is not deterministically closed. `_speech_stream` has the request-abort logic in its `finally`, but that finally will not reliably run. The coordinator queue/future and backend generation can remain alive until GC or natural completion; the risk is that a disconnected client keeps occupying worker capacity and stream queues.

 Add a `try/finally` around the wrapper and call `await stream.aclose()` in the finally block. If the route takes an error-response branch after prefetching the first event, close `speech_events` first. Add a test that prefetch succeeds, the StreamingResponse consumer closes the wrapper, and `client.abort(request_id)` plus stream cleanup are executed.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,430 @@
...
+            "ref_audio data URL must include base64 media data",
+            param=param,
+        )
+    try:
+        base64.b64decode(encoded, validate=True)
+    except (binascii.Error, ValueError) as exc:
+        raise bad_request(
+            "ref_audio data URL must include valid base64 media data",
+            param=param,
+        ) from exc
```
zhaochenyang20: `prepare_request` only checks that `input` is non-empty and has no max character count. `_validate_data_url` decodes the entire base64 payload at once. `file://` allowlisting validates the path but not `st_size`. HTTP reference fetching in the backend buffers the whole response content in memory. A caller can submit very long text, a huge data URL, a huge allowed local file, or a huge remote response and exhaust API-process or worker memory. Higgs `_MAX_REF_AUDIO_SEC` limits the retained waveform duration, not the bytes downloaded, read, or decoded before that point; the risk is a single request causing DoS or worker OOM. 

Add a max input length at the API boundary, for example an OpenAI-compatible 4096-character limit or a configurable cap. Add one byte cap for reference audio. Estimate `data:` decoded length before decoding and reject oversized payloads; check `st_size` for `file://`; stream HTTP reads and abort once the cap is exceeded. Oversized inputs should return a 400/413-style client error instead of letting a worker OOM.

sglang_omni/serve/speech_service.py
zhaochenyang20: `_normalize_speech_reference` accepts `audio_path`, `ref_audio`, and `audio`, validates the string scheme, but writes the value back under the original field. `build_generate_request` then emits `reference.model_dump(exclude_none=True)`. The S2-Pro stage reads only `audio_path`; Higgs also derives reference audio only from bytes/base64/data or `audio_path`/`path`. A user can send `{"ref_audio": "..."}` or `{"audio": "..."}` and pass validation, but the model will not see the reference audio, so voice cloning degrades to no-reference behavior or fails without a 400.

zhaochenyang20: Canonicalize during normalization: `canonical = reference.audio_path or reference.ref_audio or reference.audio`, validate it, write it to `audio_path`, and clear the alias fields. Add request-lowering tests for `references=[{"ref_audio": data_url}]` and `references=[{"audio": http_url}]`, asserting that the backend sees `audio_path` or a decodable data dict rather than an alias key.

sglang_omni/serve/speech_service.py
zhaochenyang20: `prepare_request` calls `_validate_encoder_dependency` for the `response_format` before checking whether `stream=true` requires `pcm`. On a host without pydub/ffmpeg, `stream=true,response_format=mp3` can hit the optional codec dependency branch first and return a 500 `InternalServerError` instead of the more accurate and stable 400 `BadRequestError`. This is an error-classification bug and affects both benchmark and SDK consumers of contract violations.

Move the `if request.stream and response_format != "pcm"` guard before encoder dependency checks. Add a test independent of pydub/ffmpeg availability that all non-PCM streaming requests return a stable 400.

zhaochenyang20 (changes requested):
delete this test from CI: tests/docs/s2pro/test_docs_tts_s2pro.py Remove the file

sglang_omni/serve/openai_api.py
Ratish1: done

sglang_omni/serve/speech_service.py

sglang_omni/serve/speech_service.py

sglang_omni/serve/speech_service.py

sglang_omni/serve/speech_service.py
Ratish1: done

sglang_omni/serve/openai_api.py
Ratish1: done

sglang_omni/serve/speech_service.py

sglang_omni/serve/openai_api.py
Ratish1: done

sglang_omni/serve/protocol.py
Ratish1: done

sglang_omni/serve/speech_service.py

sglang_omni/serve/speech_service.py

sglang_omni/serve/openai_api.py
Ratish1: done

sglang_omni/serve/openai_api.py
Ratish1: done

sglang_omni/serve/speech_service.py
Ratish1: done

sglang_omni/serve/openai_api.py
Ratish1: done

sglang_omni/serve/speech_service.py

sglang_omni/serve/speech_service.py
Ratish1: done

sglang_omni/serve/speech_errors.py
Ratish1: done

sglang_omni/serve/speech_service.py

sglang_omni/serve/speech_service.py

sglang_omni/serve/openai_api.py
Ratish1: done

Ratish1:
Thank you for the reviews! @JingwenGu0829 @JiaxinD @yxs @xinlij  @WilsonZheng0327 @Ccyest @zhaochenyang20 . I have addressed them all.

sglang_omni/cli/serve.py
```diff
@@ -207,6 +208,15 @@ def _validate_encoder_mem_reserve(value: float | None) -> float | None:
     return float(value)
 
 
+def _validate_allowed_local_media_path(value: str | None) -> str | None:
```
yxs, line 224: The same "expanduser → resolve → exists → is_dir" directory check is duplicated three times: the CLI's `_validate_allowed_local_media_path`, `speech_service`'s `_resolve_allowed_local_media_path`, and inside `MultiModalResourceConnector.__init__`. 

On the startup path the directory is resolved and checked three times, with three different error strings ("must be a directory" / "must be a directory: {path}" / "Invalid path: {path}").

Looks like it is a DRY violation.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,481 @@
...
+
+    def _normalize_media_reference(self, value: str, *, param: str) -> str:
+        descriptor = self._load_media_reference_descriptor(value, param=param)
+        return _media_reference_from_descriptor(descriptor)
+
+    def _load_media_reference_descriptor(
+        self, value: str, *, param: str
+    ) -> dict[str, str]:
+        url = urlparse(value)
+        if url.scheme in {"http", "https"}:
```
yxs: this PR makes `/v1/audio/speech` reject **all** http/https reference audio, but **every** voice-cloning example in `higgs_tts.md` uses `audio_path: "https://huggingface.co/..."`. So anyone who copies the Higgs doc's curl and sends it gets a 400 instead of audio.

I let claude verified this end-to-end: using a fastapi `TestClient`, I POSTed two **verbatim** request bodies from the docs (non-streaming L89-101, streaming L151-160) to `/v1/audio/speech` through the full HTTP route — both return HTTP 400 (output below). In the test the backend client's `speech` / `generate` were stubbed to assert-if-reached, and that assert never fired — proving the request is rejected at the API boundary during validation, before any model/GPU. As a control, the same body is accepted on `origin/main` (the old `build_speech_generate_request` passes it through with no validation, and Higgs's `load_audio_to_24k` actually fetches that URL).

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,481 @@
...
+                continue
+            updates.update(
+                self._load_media_reference_descriptor(
+                    value, param=f"references.{field_name}"
+                )
+            )
+            break
+        return reference.model_copy(update=updates)
+
+    def _normalize_media_reference(self, value: str, *, param: str) -> str:
```
yxs: `_normalize_media_reference` first calls `_load_media_reference_descriptor` to get a structured descriptor (`{audio_path}` or `{data, media_type}`), then `_media_reference_from_descriptor` collapses it back into a string (data URL or path) stored in `request.ref_audio`; later `build_generate_request` re-parses that string into a dict via `_reference_dict_from_media_reference`. For data URLs this means: base64-decode once during validation (discarded) → rebuild the data-URL string → re-parse downstream. Functionally correct, but an unnecessary descriptor→string→descriptor detour.

Also `build_generate_request`'s `validate` defaults to `True`, so any caller using the default re-runs the entire load (file re-stat / base64 re-decode).

sglang_omni/serve/openai_api.py
```diff
@@ -559,49 +600,61 @@ async def _speech_stream(
...
-                "data": base64.b64encode(audio_bytes).decode("ascii"),
-                "format": actual_format,
-                "mime_type": mime_type,
-                "sample_rate": sample_rate,
-            },
-            "finish_reason": None,
-        }
-        yield f"data: {json.dumps(payload)}\n\n"
-        chunk_index += 1
+            audio_bytes, mime_type = encode_audio(
```
yxs: In a prior round @JiaxinD  flagged "synchronous encoding blocking the event loop"; the non-streaming path now fixes it with `asyncio.to_thread(encode_audio, ...)`, but the streaming path `_speech_stream` still calls `encode_audio(...)` synchronously on the event loop for every SSE chunk. Impact is small today since streaming is PCM-only and `encode_pcm` is just `clip + *32767 + astype(int16) + tobytes` (microseconds). But the two paths handle the same concern inconsistently, and if streaming formats are ever broadened this becomes a real loop-blocking point.

sglang_omni/serve/openai_api.py
```diff
@@ -714,6 +797,84 @@ async def _body():
...
+    except asyncio.CancelledError:
+        if not aborted:
+            await client.abort(request_id)
+        raise
+    finally:
+        if not speech_task.done():
+            speech_task.cancel()
+        if not disconnect_task.done():
+            disconnect_task.cancel()
+        await asyncio.gather(speech_task, disconnect_task, return_exceptions=True)
```
Hayden727: [P1] This cleanup can hang the non-streaming speech failure path. On current head `929aa0a`, running `timeout 120s python -m pytest tests/unit_test/serve/test_speech_protocol.py tests/unit_test/serve/test_openai_api.py -vv -o faulthandler_timeout=60` in the `sglang-omni-hayden` container stalls in `test_non_streaming_http_faults_return_500[qwen3-omni]`; faulthandler shows `TestClient.post()` still waiting after entering this endpoint path. When `speech_task` fails quickly, the handler should return the OpenAI-style 500, but this `finally` waits on the disconnect watcher cleanup. Please make the disconnect watcher non-blocking/timeout-bounded, or otherwise avoid awaiting it in the backend-error return path, and add coverage for the backend-failure case.

sglang_omni/preprocessing/resource_connector.py
```diff
@@ -85,27 +158,34 @@ def __init__(
...
+            )
 
-        self.allowed_media_domains = allowed_media_domains or []
+        self.allowed_media_domains = [
+            domain.lower() for domain in allowed_media_domains or []
+        ]
 
     def _assert_url_allowed(self, url_spec: Any) -> None:
         """Check if URL hostname is in allowed domains."""
         if (
```
Hayden727, line 246: [P1] This still leaves remote speech references unrestricted by default: `allowed_media_domains` is empty unless the CLI flag is provided, so this guard skips all host validation. Because `SpeechRequestValidator` loads `ref_audio` / `references.*` server-side before model execution, a public `/v1/audio/speech` endpoint can fetch loopback, RFC1918, link-local, or metadata-service URLs. The 10 MB cap and content-type check reduce payload risk but do not prevent SSRF/probing. Please require explicit remote-media opt-in or an allowlist by default, and reject loopback/private/link-local/reserved resolved IPs for the initial URL and every redirect.

Hayden727 (commented):
I traced the speech request path against the PR goal and validated the current head `929aa0a7223f9179cb36476201abcbcd3facf088`. Two coverage gaps remain: non-streaming backend failures can hang instead of returning the OpenAI-style error, and remote reference audio is still unrestricted by default.

docs/cookbook/higgs_tts.md
```diff
@@ -448,7 +458,7 @@ resp = requests.post(
 | `voice` | string | `"default"` | Voice identifier (ignored when `references` is set) |
 | `response_format` | string | `"wav"` | Output audio format |
 | `stream` | bool | `false` | Enable streaming via SSE |
-| `references` | list | `null` | Reference audio for voice cloning; each item has `audio_path` (local path or HTTP URL) and `text` (transcript) |
+| `references` | list | `null` | Reference audio for voice cloning; each item has `audio_path` (allowed remote URL, data URL, or allowed `file://` URI) and `text` (transcript) |
```
Hayden727: [P2] The cookbook still does not show the allowlist needed for the remote-reference examples above. The launch command at the top starts the server without `--allowed-media-domain`, but the voice-cloning examples all use `https://huggingface.co/...` reference audio, so copying the cookbook now returns 400 before reaching the backend: `Remote media URLs require --allowed-media-domain to be configured.` Even `--allowed-media-domain huggingface.co` is not enough for the shown URL, because HuggingFace redirects it to `cas-bridge.xethub.hf.co`, and the connector revalidates every redirect hop. Please update the Higgs launch instructions/example to include the required remote domains, e.g. both `huggingface.co` and the current download host, or switch the examples to data/file references so the documented flow works as written.

Hayden727 (commented):
Re-reviewed latest head `6173aa1a57ce69c7db30fa9ebd3ddc4a8706550e`. The previous non-streaming failure hang and default-unrestricted remote media issue are fixed in code and validated, but one docs/API mismatch remains for the Higgs remote-reference examples.

tests/unit_test/serve/test_openai_api.py
zijiexia, line 1: maybe we could add a small `TestClient` test that POSTs an SDK-shaped body (`{model, voice, input, response_format}`) and asserts the binary response is what `stream_to_file` would consume?

zhaochenyang20:
good job so far. If zijie approved and passed the CI. we can merge @Ratish1 @zijiexia

tests/unit_test/serve/test_openai_api.py
```diff
@@ -107,6 +107,140 @@ async def generate(self, request: Any, request_id: str | None = None):
...
+        yield GenerateChunk(
+            request_id=request_id or "speech-1",
+            modality="audio",
+            audio_data=None,
+            sample_rate=24000,
+            finish_reason="stop",
+        )
+
+
+class BlockingStreamingSpeechClient:
```
zijiexia: Does this test still needed?

tests/unit_test/serve/test_openai_api.py
```diff
@@ -357,13 +570,16 @@ def test_speech_stream_defaults_to_sse_for_compatibility() -> None:
...
-    assert response.text.endswith("data: [DONE]\n\n")
+    assert response.headers["content-type"].startswith("audio/pcm")
+    assert response.headers["x-sample-rate"] == "24000"
+    assert response.headers["x-channels"] == "1"
+    assert response.headers["x-bit-depth"] == "16"
+    assert response.content == expected
 
 
-def test_speech_stream_audio_format_returns_raw_pcm_bytes() -> None:
+def test_speech_stream_returns_raw_pcm_bytes() -> None:
```
zijiexia: is this test duplicated with `test_speech_stream_defaults_to_raw_pcm ` as current default is raw pcm?

zijiexia (approved):
with a little cleanup maybe

tests/unit_test/serve/test_openai_api.py
Ratish1: yes, removed now

tests/unit_test/serve/test_openai_api.py
Ratish1: removed

zhaochenyang20 (approved):

PR #662, opened 2026-06-02
[TTS] Add uploaded voice APIs

Ratish1:
## Motivation

Stacked on #659. 

This PR adds persistent uploaded voice support for the OpenAI-compatible TTS serving path. Users can upload a reference clip once through `/v1/audio/voices`, list registered voices, reuse an uploaded voice by name in `/v1/audio/speech`, and delete it when it is no longer needed.

## Architecture

Uploaded voices are owned by the HTTP serving layer and lowered into the existing model-specific reference-audio paths before the request reaches the model pipeline.

```mermaid
flowchart LR
    Upload[POST /v1/audio/voices] --> Store[SpeakerSampleStore]
    Store --> Disk[.safetensors voice sample]
    Store --> Cache[shared speaker artifact cache]
    Speech[POST /v1/audio/speech voice=name] --> Resolve[resolve uploaded voice]
    Resolve --> Higgs[Higgs ref audio/ref code path]
    Resolve --> Qwen[Qwen3-TTS Base ref audio / x-vector path]
    Delete[DELETE /v1/audio/voices/name] --> Store
    Delete --> Cache
```

The API layer owns upload validation, persistence, list/delete behavior, and conversion to reference audio. Model configs expose whether named `voice` values must resolve to uploaded voices, so `SpeechService` does not parse model names or special-case individual TTS families.

## Modifications

- Added `/v1/audio/voices` GET/POST/DELETE routes with bounded upload reads, upload validation, persistent `.safetensors` storage, restore-on-startup behavior, overwrite handling, and delete cleanup.
- Added a shared speaker artifact cache with per-voice invalidation so uploaded voice overwrites/deletes clear Higgs and Qwen cached artifacts consistently.
- Lower uploaded voices into `/v1/audio/speech` as reference audio before model execution. Explicit `ref_audio` / `references` still take precedence over `voice` lookup.
- Wired Higgs TTS uploaded voice metadata into waveform/ref-code caching so repeated use of the same uploaded voice avoids recomputing reusable prompt artifacts.
- Wired Qwen3-TTS Base uploaded voices into the existing reference-audio path, including x-vector-only mode when no upload transcript is available. The Qwen cache path now handles x-vector prompts without `ref_code` tensors.
- Added model-config capability plumbing for named voice policy: Higgs and Qwen3-TTS Base require uploaded voice resolution for non-default voice names, while preset-voice models can keep their own built-in voice semantics.
- Updated TTS docs with uploaded voice usage, accepted formats, limits, persistence location, and cache invalidation behavior.

## Accuracy Tests

- `.venv/bin/python -m pytest tests/unit_test/serve/test_speech_voices.py tests/unit_test/serve/test_openai_api.py -q`
  - Result: `23 passed`.
- Full unit tests were also confirmed passing 
- Higgs uploaded voice live curl flow on `boson-sglang/higgs-audio-v3-TTS-4B-grpo05200410999`:
  - `POST /v1/audio/voices` with `docs/_static/audio/gaokao-listening.wav` returned uploaded metadata for `test_voice` (`file_size=505038`, `mime_type=audio/wav`).
  - `GET /v1/audio/voices` returned `default` and `test_voice` plus cache stats.
  - `POST /v1/audio/speech` with `voice="test_voice"` produced `uploaded_voice.wav` (~228 KiB).
  - `DELETE /v1/audio/voices/test_voice` returned `{"success": true, "message": "Voice 'test_voice' deleted successfully"}`.
- Qwen3-TTS Base uploaded voice live curl flow:
  - `POST /v1/audio/speech` with `voice="test_voice"`, `task_type="Base"`, `x_vector_only_mode=true`, and `response_format="wav"` returned `200 OK` and produced `qwen_uploaded_voice.wav` (~150 KiB).

## Speed Tests and Profiling

Not run. This PR changes API routing, upload persistence, and reusable speaker artifact caching; no latency/throughput benchmark was run for this PR.

## Checklist


sglang_omni/serve/openai_api.py
```diff
@@ -118,6 +153,66 @@ def create_app(
...
+        return JSONResponse(
+            content={
+                "success": True,
+                "message": f"Voice '{name}' deleted successfully",
+            }
+        )
+
+
+async def _read_voice_upload(audio_sample: UploadFile) -> bytes:
+    audio_bytes = await audio_sample.read(MAX_VOICE_UPLOAD_BYTES + 1)
```
JiaxinD, line 283: **The new `POST /v1/audio/voices` endpoint has no request-body size limit.**

`_read_voice_upload` correctly caps the in-memory bytes with `read(MAX_VOICE_UPLOAD_BYTES + 1)`, but that runs only **after** FastAPI fully parses the multipart form: with `audio_sample: UploadFile = File(...)`, Starlette streams the part into a `SpooledTemporaryFile` that rolls over to a temp file on disk before the handler body executes. `create_app` installs only `CORSMiddleware`, so a client can POST a multi-GB body and have it spooled to the server's temp dir before the bounded read runs, i.e. a transient disk/inode-exhaustion DoS under concurrency. (The pre-existing `/v1/audio/transcriptions` route shares this with an unbounded `file.read()`, but this PR adds a new write-capable public endpoint that inherits it.)

Cheap first line: a `Content-Length` check returning 413 before the body is buffered.

```python
cl = request.headers.get("content-length")
if cl is not None and cl.isdigit() and int(cl) > MAX_VOICE_UPLOAD_BYTES:
    raise HTTPException(status_code=413, detail="request body too large")
```

The header check is bypassable via chunked transfer-encoding, so for a robust guard add a small streaming body-size-limiting ASGI middleware, but the header check handles the common case and is worth adding regardless.

sglang_omni/scheduling/speaker_cache.py
```diff
@@ -0,0 +1,125 @@
...
+        return sys.getsizeof(value) + sum(
+            estimate_cache_bytes(key) + estimate_cache_bytes(item)
+            for key, item in value.items()
+        )
+    if isinstance(value, list | tuple | set | frozenset):
+        return sys.getsizeof(value) + sum(estimate_cache_bytes(item) for item in value)
+    return sys.getsizeof(value)
+
+
+_GLOBAL_SPEAKER_CACHE = SpeakerArtifactCache()
```
JiaxinD, line 125: **Per-voice cache invalidation never reaches the worker-process artifact caches.**

`_GLOBAL_SPEAKER_CACHE` is a per-process singleton, but the pipeline stages run in **separate spawned worker processes** (`mp_runner.py` uses `get_context("spawn")`; `stage_process.py` builds each stage via `import_string(spec.factory)` in the child). With spawn, every worker re-imports this module and gets its own `_GLOBAL_SPEAKER_CACHE`.

So the large model-side artifacts, namely Higgs `reference_waveform`/`reference_codes` (`stages.py`) and Qwen `voice_clone_prompt` (`request_builders.py`), are `put()` into the **worker** caches, but `clear_voice()` is only ever called from `SpeakerSampleStore.upload`/`.delete` in the API process and only clears **that** process's cache (which holds just the `wav_data_url` entry). There is no IPC to propagate invalidation. Consequences:

1. On overwrite/delete, worker-side artifacts for that voice are never freed; they linger until LRU eviction (bounded by the 512 MiB budget).
2. `cache_stats` in `GET /v1/audio/voices` (`list_response()` then `self.cache.stats()`) reflects only the near-empty API-process cache; `entries`/`memory_bytes`/`delete_invalidation_counter` never reflect the worker caches.
3. `tts.md` overstates this: "invalidates cached speaker artifacts for that voice across TTS model types" is true within one process, false across the process boundary.

Audio correctness is **not** affected: `_next_upload_timestamp` makes `created_at` strictly increasing, so overwrite produces a new `voice_version` (a new key, with the stale entry orphaned and never re-served), and a deleted voice can no longer be resolved. This is a worker-memory, observability, and docs issue.

Fix (pick one). Low-effort (recommended): reword `tts.md` to drop the cross-process invalidation promise and either drop `cache_stats` or label it API-process-only. Full fix: propagate `clear_voice(name)` to workers over the existing stage-control channel.

JiaxinD (commented):
Reviewed the incremental uploaded-voice changes on top of #659. Defensively written: atomic safetensors writes, bounded `read(MAX+1)`, `safe_open(framework="np")` load, a name pattern blocking separators, and genuinely thorough store/cache tests. The two Medium items are inline; the rest are below. (Noting it is a stacked draft, so happy to hold these until it is ready or #659 lands.)

### Low
* **Uploaded voice silently overrides an explicit `task_type`** (`build_generate_request`): when a voice resolves, `task_type` is unconditionally forced to `"Base"`, discarding an explicit `task_type` set a few lines above. `voice=<uploaded>` plus `task_type="CustomVoice"` is contradictory (Qwen3 CustomVoice/VoiceDesign reject ref audio), but it is swallowed rather than reported. Suggest raising `bad_request` on the conflict instead of overriding.
* **`SPEAKER_MAX_UPLOADED` not enforced on restore** (`_restore()`): it loads every `*.safetensors` with no cap check, while `upload()` enforces it, so after a restart the registry can exceed the limit (e.g. an operator lowers the cap). Either cap the restore (keep newest-first, log skipped) or clarify in docs that the cap governs only new uploads.
* **Hand-rolled WAV decoder duplicates `AudioMediaIO` with divergent sample-rate semantics** (`_decode_reference_audio`): `_decode_wav_upload` preserves the native rate while the `AudioMediaIO` fallback force-resamples to 24 kHz, so an int-PCM WAV and an MP3 or float-WAV of the same clip get different stored `sample_rate`/`num_samples` and a different cached data URL. `AudioMediaIO._parse_wav_bytes` is already a superset (it handles float WAV that the `wave`-based path rejects). Not a correctness bug (the model re-resamples), but gratuitous: suggest decoding everything through `AudioMediaIO` (about 30 fewer lines), or add a comment if the native-rate fast path is intentional.
* **Test gaps**: (a) no test that an explicit `ref_audio`/`references` overrides a same-named uploaded voice (the precedence is correct in code but unguarded); (b) the rejected-MIME, `file_size == 0`, and `VOICE_SILENCE_THRESHOLD` 400 branches in `upload()` are untested; (c) the DELETE path never asserts that `delete()` calls `clear_voice` (other tests exercise invalidation only via the overwrite path). A store-level delete-invalidation test (pre-populate cache, `store.delete`, assert entry gone and `delete_invalidation_counter == 1`) would close (c).

### Nits
* **`uploaded_voice_fingerprint` is plumbed end-to-end but never read**: set into `tts_params`, threaded onto both state dataclasses, serialized in every `to_dict`/`from_dict`, yet no cache key or validation consumes it (`SpeakerCacheKey` uses `created_at` as the version). Drop the plumbing, or fold it into `SpeakerCacheKey` if content-change detection independent of the timestamp is intended.
* **Decode-failure message leaks internal exception detail**: `_decode_reference_audio` returns the raw PyAV/libav exception string to the client via `bad_request`. Return a generic message and log the detail server-side.
* **`normalize_voice_name` accepts pure-punctuation names** (`.`, `..`, `-`): not path traversal (no separators, and the name always gets `.safetensors` appended), just a degenerate identifier plus a hidden dotfile. Requiring at least one alphanumeric (`^(?=.*[A-Za-z0-9])[A-Za-z0-9_.-]+$`) cleans this up.

sglang_omni/serve/speech_voices.py
```diff
@@ -0,0 +1,563 @@
...
+    except Exception as exc:
+        tmp_path.unlink(missing_ok=True)
+        raise bad_request(f"Failed to save uploaded voice: {exc}") from exc
+
+
+def _replace_voice_file(temp_path: Path, path: Path) -> None:
+    try:
+        os.replace(temp_path, path)
+    except OSError as exc:
+        raise bad_request(f"Failed to save uploaded voice: {exc}") from exc
```
xinlij: `OSError` from `unlink` / `os.replace` is a server-side IO failure (disk full, permissions), not a client error. This should be `internal_error` (500) rather than `bad_request` (400). As written, a disk full condition returns HTTP 400 with `param="name"`, which will mislead callers into thinking they passed a bad voice name.

sglang_omni/serve/speech_voices.py
```diff
@@ -0,0 +1,563 @@
...
+                return False
+            try:
+                voice.file_path.unlink(missing_ok=True)
+            except OSError as exc:
+                logger.warning(
+                    "Failed to delete voice file %s: %s", voice.file_path, exc
+                )
+                raise bad_request(
+                    f"Failed to delete voice '{voice.name}' from storage",
+                    param="name",
```
xinlij: `OSError` from `unlink` / `os.replace` is a server-side IO failure (disk full, permissions), not a client error. This should be `internal_error` (500) rather than `bad_request` (400). As written, a disk full condition returns HTTP 400 with `param="name"`, which will mislead callers into thinking they passed a bad voice name.

sglang_omni/serve/speech_voices.py
```diff
@@ -0,0 +1,563 @@
...
+        from sglang_omni.preprocessing.audio import AudioMediaIO
+
+        samples, sample_rate = AudioMediaIO(target_sr=DEFAULT_SAMPLE_RATE).load_bytes(
+            audio_bytes
+        )
+    except Exception as exc:
+        raise bad_request(
+            f"audio_sample could not be decoded: {exc}",
+            param="audio_sample",
+        ) from exc
```
xinlij, line 433: The full exception message `{exc}` is returned verbatim to the client. Since `AudioMediaIO` uses ffmpeg internally, this can leak server-side details like temp file paths (`/tmp/.voice.xxx.tmp`) and ffmpeg version strings — useful for targeting known CVEs. Consider logging the full error and returning a generic message instead.

xinlij (changes requested):

sglang_omni/serve/openai_api.py
```diff
@@ -118,6 +153,66 @@ def create_app(
...
+                consent=consent,
+                audio_bytes=await _read_voice_upload(audio_sample),
+                filename=audio_sample.filename,
+                content_type=audio_sample.content_type,
+                ref_text=ref_text,
+                speaker_description=speaker_description,
+            )
+        except SpeechAPIError as exc:
+            return speech_error_response(exc)
+        return JSONResponse(content=response)
```
zhaochenyang20, line 260: `_read_voice_upload` reads only `MAX_VOICE_UPLOAD_BYTES + 1` inside the handler, but FastAPI/Starlette parses the multipart body first and stores the `UploadFile` part in a `SpooledTemporaryFile`, rolling over to disk after the in-memory threshold. A caller can POST a multi-GB body, and the 10 MB check happens too late after the temp directory and inodes may already be exhausted. This PR adds a new public write endpoint, so handler-level reads are not enough; the risk is disk or inode DoS under concurrent uploads.

zhaochenyang20: Add ASGI body-size limiting middleware before the multipart parser, or configure a max request body for `/v1/audio/voices` in the server stack. A `Content-Length` fast reject is useful as a first layer but cannot be the only defense because chunked transfer can bypass it. The cap should include multipart overhead, for example 10 MB audio plus small form fields, and return 413 or the documented 400.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,494 @@
...
+        if request.instructions is not None:
+            tts_params["instructions"] = request.instructions
+        if request.ref_audio is not None:
+            tts_params["ref_audio"] = request.ref_audio
+        if request.ref_text is not None:
+            tts_params["ref_text"] = request.ref_text
+        if uploaded_voice is not None:
+            tts_params["task_type"] = "Base"
+            tts_params["ref_audio"] = uploaded_voice.ref_audio
+            if uploaded_voice.voice.ref_text is not None:
```
zhaochenyang20: `prepare_request` parses the user-provided `task_type`, but if `voice` resolves to an uploaded voice, `build_generate_request` unconditionally sets `tts_params["task_type"] = "Base"`. `voice=<uploaded>, task_type="CustomVoice"` or `"VoiceDesign"` is a contradictory request and should return 400. The current implementation swallows the user's intent and turns it into Base. For Qwen/Higgs this changes the model path, and for the benchmark it disguises an API contract violation as successful synthesis; the risk is that clients receive audio generated with semantics different from their request.

 If an uploaded voice resolves and the request explicitly sets a non-`Base` `task_type`, raise `bad_request` with `param` pointing to `task_type` or `voice`. Only inject Base when `task_type` is absent or explicitly `Base`. Add unit tests for uploaded voice plus `CustomVoice`, uploaded voice plus `VoiceDesign`, and uploaded voice plus Base.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,494 @@
...
+            if isinstance(value, bool) or not isinstance(value, int):
+                raise bad_request("top_k must be an integer", param="top_k")
+        for field_name in ("speed", "temperature", "top_p", "repetition_penalty"):
+            if field_name in payload and payload[field_name] is not None:
+                value = payload[field_name]
+                if isinstance(value, bool) or not isinstance(value, (int, float)):
+                    raise bad_request(
+                        f"{field_name} must be a number", param=field_name
+                    )
+        for field_name in ("stream", "x_vector_only_mode"):
```
zhaochenyang20, line 299: `upload()` rejects new names once `self.max_uploaded` is reached, but `_restore()` loads every `*.safetensors` file in the directory. If an operator lowers `SPEAKER_MAX_UPLOADED`, or if historical files already exceed the cap, a restart will load a registry larger than the configured limit. That makes speaker-cap benchmark initial state and production cap semantics unpredictable; the risk is that the cap contract no longer holds after restart.

During restore, sort by `created_at` or mtime, load at most `max_uploaded`, log warnings for skipped files, or explicitly document that the cap only limits new uploads. Since #601 says uploads past the cap return 400 and voices restore from disk, restore should preserve the cap invariant.

sglang_omni/serve/speech_voices.py
```diff
@@ -0,0 +1,563 @@
...
+DEFAULT_SPEAKER_SAMPLES_DIR = Path("~/.cache/sglang-omni/speakers").expanduser()
+DEFAULT_SPEAKER_MAX_UPLOADED = 1000
+MAX_VOICE_UPLOAD_BYTES = 10 * 1024 * 1024
+MIN_REFERENCE_AUDIO_SECONDS = 1.0
+MAX_REFERENCE_AUDIO_SECONDS = 30.0
+VOICE_SILENCE_THRESHOLD = 1e-5
+VOICE_METADATA_INT_FIELDS = frozenset(
+    {"created_at", "file_size", "sample_rate", "num_samples"}
+)
+VOICE_NAME_PATTERN = re.compile(r"^[A-Za-z0-9_.-]+$")
```
zhaochenyang20: `VOICE_NAME_PATTERN = ^[A-Za-z0-9_.-]+$` accepts names such as `"."`, `".."`, and `"-"`. The path appends `.safetensors`, so this is not path traversal, but it creates hidden or degenerate identifiers that are awkward for list/delete/CLI usage; the risk is that users and operators cannot reliably identify those uploaded voices.

Require at least one ASCII letter or digit, for example `^(?=.*[A-Za-z0-9])[A-Za-z0-9_.-]+$`, and add rejected-name tests.

sglang_omni/serve/speech_voices.py
zhaochenyang20, line 433: `_decode_reference_audio` catches a broad exception and returns the raw PyAV/libav exception string through `bad_request`. That exposes internal decoder, path, or codec-build details to callers and makes error messages unstable for benchmark matching.

Return a stable generic client message such as `"audio_sample could not be decoded as a supported audio format"` and log the detailed exception server-side only.

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,494 @@
...
+        if request.ref_text is not None:
+            tts_params["ref_text"] = request.ref_text
+        if uploaded_voice is not None:
+            tts_params["task_type"] = "Base"
+            tts_params["ref_audio"] = uploaded_voice.ref_audio
+            if uploaded_voice.voice.ref_text is not None:
+                tts_params["ref_text"] = uploaded_voice.voice.ref_text
+            tts_params["uploaded_voice_name"] = uploaded_voice.voice.normalized_name
+            tts_params["uploaded_voice_created_at"] = uploaded_voice.voice.created_at
+            tts_params["uploaded_voice_fingerprint"] = uploaded_voice.voice.fingerprint
```
zhaochenyang20: This field is written into `tts_params` and propagated through Higgs/Qwen payload state and serialization, but the cache key uses `uploaded_voice_name` plus `created_at`, not the fingerprint. The field is dead plumbing today; it expands state schemas and makes readers think content fingerprints participate in stale-artifact detection when they do not; the risk is future cache-key work relying on a field that appears meaningful but has no semantics.

 Either remove the fingerprint plumbing or include it in `SpeakerCacheKey` and use it to replace or strengthen the `created_at` version. Pick one to avoid state drift.

zhaochenyang20 (changes requested):

Ratish1:
I will address these reviews after #659 is merged. Thanks @JiaxinD @xinlij @zhaochenyang20

sglang_omni/scheduling/speaker_cache.py
```diff
@@ -0,0 +1,125 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Shared LRU cache for uploaded-speaker feature artifacts."""
```
yxs, line 2: Consider move `speaker_cache.py` to a neutral layer.

Since `SpeakerArtifactCache` is a pure LRU with zero HTTP/serving dependency, yet it lives under `sglang_omni/serve/` and is imported backwards by `models/higgs_tts/stages.py` and `models/qwen3_tts/request_builders.py`.

yxs, line 2: `SpeakerArtifactCache`'s mechanism,`OrderedDict` + `move_to_end` + byte-budget eviction via `popitem(last=False)`, is almost line-for-line isomorphic to the existing `StageOutputCache`, which is also an `OrderedDict` LRU + `max_bytes` + byte eviction.

zijiexia (approved):

sglang_omni/serve/openai_api.py
```diff
@@ -158,6 +229,110 @@ def create_app(
...
+            param="audio_sample",
+        )
+    return audio_bytes
+
+
+def _is_voice_upload_scope(scope: dict[str, Any]) -> bool:
+    return (
+        scope.get("type") == "http"
+        and scope.get("method") == "POST"
+        and scope.get("path") == "/v1/audio/voices"
```
zijiexia, line 296: follow-up (non-blocking, pre-existing — echoing @JiaxinD 's earlier note): the new
body-limit middleware covers `/v1/audio/voices`, but `create_transcription` is still
an unauthenticated `await file.read()` with no cap (~L1204), same multipart-spool
DoS shape. Out of scope for this PR (it predates it), but since the streaming
middleware now exists, a follow-up could extend `_is_voice_upload_scope` to
`/v1/audio/transcriptions` with its own (larger) `max_bytes`. Fine as a separate PR.

zhaochenyang20 (approved):

PR #665, opened 2026-06-02
[TTS] Add batch and WebSocket speech APIs

Ratish1:
## Motivation

Stacked on top of #662.

This adds the remaining TTS serving API surfaces for batch speech generation and stateful WebSocket speech streaming. The endpoints reuse the same speech request preparation path as `/v1/audio/speech`, so model-specific request handling stays behind the existing client/coordinator flow.

## Modifications

- Adds `POST /v1/audio/speech/batch` with request-level defaults, item-level overrides, ordered results, item-level error envelopes, max-item configuration, and request-disconnect cancellation for in-flight batch work.
- Adds `WS /v1/audio/speech/stream` with `session.config`, `input.text`, `input.done`, sentence/clause splitting, non-streaming binary audio frames, PCM streaming frames, and explicit session lifecycle events.
- Adds strict serving-boundary validation for batch and WebSocket duration/generation fields before Pydantic coercion, including structured errors for invalid `token_count`, `duration_tokens`, and unsupported batch streaming.
- Adds WebSocket cleanup for task cancellation, send failures, API-side stream failures, and passive peer disconnects while generation is active.
- Documents the batch and WebSocket APIs in `docs/basic_usage/tts.md` and adds focused serving unit coverage for protocol, batch, and WebSocket lifecycle behavior.

## Accuracy Tests

- `.venv/bin/python -m pytest -q tests/unit_test/serve/test_speech_protocol.py tests/unit_test/serve/test_speech_batch.py tests/unit_test/serve/test_speech_ws.py` (`73 passed`)
- `pre-commit run --files sglang_omni/serve/speech_ws.py tests/unit_test/serve/test_speech_ws.py`

Higgs TTS validation:

- Batch request with three items succeeded with ordered results, mixed item overrides, and a PCM item override (`total=3`, `succeeded=3`, `failed=0`).
- Batch default validation returned request-level `400` for `token_count=0` with an OpenAI-style error envelope.
- Batch item validation returned item-level errors for `duration_tokens=-1` and `stream=true` while preserving the batch envelope.
- WebSocket non-streaming mode returned `session.configured`, `audio.start`, binary audio, `audio.done`, and `session.done`.
- WebSocket PCM streaming mode returned per-sentence `audio.start` / binary audio / `audio.done` events and `session.done`.
- WebSocket validation checked invalid `stream_audio=true` with `response_format=wav` and invalid `duration_tokens=0`.
- Interrupting a long batch left `/health` healthy and a follow-up `/v1/audio/speech` request succeeded.


sglang_omni/serve/openai_api.py
```diff
@@ -522,12 +644,13 @@ async def create_speech(req: CreateSpeechRequest) -> Response:
...
                 speed=req.speed,
+                allow_format_fallback=False,
             )
         except ClientError as exc:
-            raise HTTPException(status_code=500, detail=str(exc)) from exc
+            return speech_error_response(internal_error(str(exc)))
         except Exception as exc:
             logger.exception("Error generating speech for request %s", request_id)
-            raise HTTPException(status_code=500, detail=str(exc)) from exc
+            return speech_error_response(internal_error(str(exc)))
```
zhaochenyang20: In #659 head, non-streaming `/v1/audio/speech` uses `_await_speech_response(...)`, which waits for both the speech task and request disconnect and calls `client.abort(request_id)` on disconnect. In the current #665 head, this has regressed back to direct `await client.speech(...)` with no disconnect watcher. SSE regresses similarly: #659 `_speech_stream` tracked `active_request` and had `finally: await client.abort(request_id)`, while #665 `_speech_stream` is just `async for chunk in client.generate(...)` with no abort finally. When clients disconnect, the AR scheduler/vocoder can keep generating to `max_new_tokens` for a dead connection, which is exactly the expensive leak production streaming has to prevent.

Restore #659 `_await_speech_response` semantics into the #665 final head, restore the SSE generator `try/finally` abort, and keep #665's sleep in `_wait_for_request_disconnect`. Add regression tests proving `client.abort` is called for single-shot disconnect and SSE consumer close, so stacked PRs cannot drop the earlier fix again.

sglang_omni/serve/openai_api.py
```diff
@@ -612,122 +741,96 @@ async def _speech_stream(
...
-        audio = audio.squeeze()
-    if audio.ndim > 1:
-        if audio.shape[0] < audio.shape[-1]:
-            audio = audio[0]
-        else:
-            audio = audio[:, 0]
-
-    total_samples = int(audio.shape[-1]) if audio.ndim else 0
-    if not is_terminal:
-        return audio, emitted_samples + total_samples
```
zhaochenyang20, line 1402: #665 still prefetches with `anext(speech_events)` and passes `_prepend_speech_stream_event(first_event, speech_events)` to `StreamingResponse`. The wrapper has no `finally: await stream.aclose()`. This is the same cleanup bug as #659, but worse in #665 because `_speech_stream` itself also lost its abort finally. If the client disconnects after prefetch, the inner generator and coordinator stream cleanup are nondeterministic and can leave queues/futures alive while audio chunks continue to be pushed; the risk is a dead stream leaking worker capacity and memory.

 Wrap the prepending iterator in `try/finally` and close the inner async iterator. If the route returns an error response after prefetch, close the iterator first. The test belongs in #665 because this final head owns the wrapper alongside batch/WS routes.

sglang_omni/serve/speech_ws.py
```diff
@@ -0,0 +1,501 @@
...
+            await generation_task
+        disconnect_task.result()
+        raise WebSocketDisconnect
+
+    async def _watch_client_disconnect(self) -> None:
+        while True:
+            message = await self.websocket.receive()
+            if message.get("type") == "websocket.disconnect":
+                raise WebSocketDisconnect
+            self.buffered_receive_messages.append(message)
```
zhaochenyang20, line 486: `_run_generation_until_disconnect` starts `_watch_client_disconnect()` for every sentence generation. That watcher calls `websocket.receive()` and appends every non-disconnect message to `self.buffered_receive_messages`. There is no frame-count or aggregate-byte limit. A client can keep sending 128 KB text frames while the server is generating a long sentence, and the watcher will buffer all of them. `MAX_BUFFERED_TEXT_CHARS` is checked only later when `_handle_input_text` processes text payloads, so it does not bound the watcher deque itself. One WebSocket session can therefore grow memory during generation; the risk is single-connection DoS.

Do not store non-disconnect messages unboundedly in the disconnect watcher. Options: allow only one buffered message during generation and close/abort on overflow, or track buffered ASGI message bytes/count and immediately send an error, abort the active request, and close the session when the cap is exceeded. A cleaner structure is to separate the receive loop from generation and use a bounded `asyncio.Queue` for client input.

sglang_omni/serve/openai_api.py
```diff
@@ -612,122 +741,96 @@ async def _speech_stream(
...
-    if references:
-        prompt = {"text": req.input, "references": references}
 
-    return GenerateRequest(
-        model=req.model or default_model,
-        prompt=prompt,
-        sampling=sampling,
-        stage_params=req.stage_params,
-        stream=req.stream,
-        output_modalities=["audio"],
```
zhaochenyang20: `_create_speech_batch_with_disconnect_watch` only calls `disconnect_task.cancel()` in `finally` and does not `await` it under `suppress(asyncio.CancelledError)`. This is usually a short-lived pending task, but under high batch-cancellation concurrency it creates noisy task lifecycle behavior and makes it harder for tests to prove the watcher exited. #659 `_await_speech_response` cancels and awaits both tasks; keep the same pattern here. The risk is pending-task warnings or unstable cleanup ordering on cancellation paths.

Cancel and await the watcher in the finally block, and add a test proving both batch completion and disconnect cancellation paths leave no pending-task warnings.

sglang_omni/serve/speech_ws.py
```diff
@@ -0,0 +1,501 @@
...
+            if active_request_id is not None:
+                await self._abort_request(active_request_id)
+            raise WebSocketDisconnect from exc
+
+    async def _abort_request(self, request_id: str) -> None:
+        if self.active_request_id != request_id:
+            return
+        self.active_request_id = None
+        if hasattr(self.client, "abort"):
+            await self.client.abort(request_id)
```
zhaochenyang20: `SpeechWebSocketSession.client` is typed as `Client`, and production cleanup relies on `client.abort`. Using `hasattr(self.client, "abort")` silently skips abort when a test fake lacks the method and makes the real client contract unclear. WebSocket cleanup is production-critical; missing abort should not be treated as optional capability. The risk is tests passing without actually verifying real cleanup.

 Call `await self.client.abort(request_id)` directly and make test fakes implement abort. If a protocol is needed, define a typed Protocol instead of dynamic `hasattr`.

tests/unit_test/serve/test_openai_api.py
```diff
@@ -204,6 +299,11 @@ def test_speech_stream_success_emits_done_sentinel() -> None:
     assert payload["finish_reason"] == "stop"
 
 
+def test_speech_stream_without_audio_fails_without_done_sentinel() -> None:
+    with pytest.raises(ClientError, match="No audio output generated"):
+        asyncio.run(_collect_speech_stream(EmptyStreamingSpeechClient()))
+
+
```
zhaochenyang20: #665 adds batch/WS tests, but the final head regresses #659 single-shot/SSE cancellation. The tests still cover `_speech_stream` success/failure, but not consumer close or request disconnect causing `client.abort`. A stacked PR should preserve and extend the earlier PR's regression tests; otherwise each layer can overwrite fixes from the previous layer.

Add three tests in the #665 final head: single-shot disconnect abort, SSE wrapper close abort plus aclose, and WebSocket buffered-receive overflow close/abort. That protects the base speech route from later batch/WS edits.

zhaochenyang20 (changes requested):

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,735 @@
...
+            extra_params=extra_params,
+            stream=request.stream,
+            output_modalities=["audio"],
+            metadata={
+                "task": "tts",
+                "tts_params": tts_params,
+            },
+        )
+
+    async def create_speech_batch(
```
yxs, line 308: `create_speech_batch` immediately does `asyncio.create_task(client.speech(...))` per item and `asyncio.gather`s them all. With the default `tts_batch_max_items=32`, a single batch request bursts up to 32 concurrent TTS submissions to the coordinator with no semaphore / chunked throttling.

Since `pipeline/coordinator.py` has a `request_admission` event and stream queues, so the engine side does schedule/queue, meaning this is not an unbounded resource leak.

sglang_omni/serve/speech_ws.py
```diff
@@ -0,0 +1,501 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Stateful WebSocket serving for text-to-speech streaming."""
```
yxs, line 2: The repo already has a mature WebSocket session, `sglang_omni/serve/realtime/session.py` (`RealtimeSession`).
Why `speech_ws.py` reimplements its core helpers nearly verbatim with identical semantics?

sglang_omni/serve/speech_service.py
```diff
@@ -0,0 +1,735 @@
...
+        try:
+            request = CreateSpeechRequest.model_validate(payload)
+        except ValidationError as exc:
+            raise bad_request(
+                _validation_error_message(exc),
+                param=_validation_error_param(exc, prefix=f"items.{index}"),
+            ) from exc
+        return self.prepare_request(request)
+
+    def _validate_batch_defaults(self, batch: CreateSpeechBatchRequest) -> None:
```
yxs, line 490: To reuse `CreateSpeechRequest`'s validation, `_validate_batch_defaults` injects a fake input `"__batch_defaults_validation__"` into the defaults and calls `prepare_request`. This is a smell, validation is driven by a magic string, and if `prepare_request` later adds input-content checks (length cap, language detection), the sentinel could spuriously trip or slip through.

sglang_omni/serve/openai_api.py
Ratish1: done

sglang_omni/serve/openai_api.py
Ratish1: done

sglang_omni/serve/speech_ws.py
Ratish1: done

sglang_omni/serve/openai_api.py
Ratish1: done

sglang_omni/serve/speech_ws.py
Ratish1: done

tests/unit_test/serve/test_openai_api.py
Ratish1: done

sglang_omni/serve/speech_service.py
Ratish1: No change here. Batch fanout remains bounded by tts_batch_max_items, and coordinator admission/scheduling owns engine-side queuing, so this is not an unbounded resource leak.

sglang_omni/serve/speech_ws.py
Ratish1: Kept separate intentionally. Realtime is audio-input/VAD/conversation oriented, while this endpoint is text-input/audio-output with sentence buffering, binary PCM frames, and prepared TTS reference reuse. Forcing reuse would couple different protocols.

sglang_omni/serve/speech_service.py
Ratish1: done

sglang_omni/serve/speech_ws.py
```diff
@@ -0,0 +1,660 @@
...
+        return exc
+    if isinstance(exc, ValidationError):
+        first_error = exc.errors()[0] if exc.errors() else {}
+        message = first_error.get("msg") or "invalid speech WebSocket config"
+        location = ".".join(str(item) for item in first_error.get("loc", ()))
+        return bad_request(f"{location}: {message}" if location else str(message))
+    return bad_request(str(exc))
+
+
+def _speech_audio_delta(
```
zijiexia: `_speech_audio_delta` is a verbatim copy of `_select_speech_audio_delta` in `openai_api.py:1280` and has already drifted — this copy lost the multi-channel-downmix comment. Please extract one shared helper (e.g. in `client/audio.py`) so the cumulative-vs-incremental terminal-chunk logic can't diverge between the two paths.

sglang_omni/serve/speech_ws.py
```diff
@@ -0,0 +1,660 @@
...
+            if isinstance(exc, SpeechAPIError):
+                error = exc
+            else:
+                error = internal_error(str(exc))
+                logger.exception("TTS WebSocket sentence failed: %s", request_id)
+            await self._send_error(error)
+        finally:
+            if self.active_request_id == request_id:
+                self.active_request_id = None
+            if self.websocket.application_state == WebSocketState.CONNECTED:
```
zijiexia: On a real peer disconnect Starlette only flips `client_state`, not application_state, so this guard (and `_send_json` at line 548) still treats the socket as live and tries to send audio.done, which can raise on uvicorn and mask the original `WebSocketDisconnect. teardown()` (line 612) already uses `client_state` — please make these consistent (or skip `audio.done` when the failure is a disconnect/cancel); the current tests pass only because the mock manually flips `application_state`.

sglang_omni/serve/speech_service.py
```diff
@@ -226,6 +305,243 @@ def build_generate_request(
...
+                )
+
+    def _build_batch_item_request(
+        self,
+        batch: CreateSpeechBatchRequest,
+        item: SpeechBatchItem,
+        *,
+        index: int,
+    ) -> CreateSpeechRequest:
+        extra_fields = sorted((item.model_extra or {}).keys())
```
zijiexia: Batch items reject unknown fields here, but `CreateSpeechRequest` (`protocol.py:183`, default `extra="ignore"`) silently ignores them on `/v1/audio/speech`, and the batch envelope ignores them too — so the same field works on the single endpoint yet fails as a batch item. Maybe align the behavior (reject at all levels, or document the intentional difference).

zijiexia (commented):
Sorry for my late response, some small comments

sglang_omni/serve/speech_ws.py
Ratish1: done

sglang_omni/serve/speech_ws.py
Ratish1: done

sglang_omni/serve/speech_service.py

sglang_omni/serve/speech_ws.py
```diff
@@ -0,0 +1,643 @@
...
+        except asyncio.CancelledError:
+            await _cancel_tasks(generation_task, disconnect_task)
+            raise
+        except Exception:
+            await _cancel_tasks(generation_task, disconnect_task)
+            raise
+
+    async def _watch_client_disconnect(self) -> None:
+        while True:
+            message = await self.websocket.receive()
```
zijiexia, line 473: Because receive() here and the append at line 487 aren't atomic with the watcher cancel at line 459, a client frame that resolves exactly as generation completes can be discarded instead of buffered — narrow/low-impact, but worth either a comment or buffering before the cancel point.

zijiexia (approved):

sglang_omni/serve/speech_ws.py
```diff
@@ -0,0 +1,643 @@
...
+                f"{message_kind} WebSocket message exceeds {max_bytes} bytes"
+            )
+
+    async def _send_json(self, payload: dict[str, Any]) -> None:
+        if not self._can_send():
+            return
+        await self.websocket.send_text(json.dumps(payload))
+
+    async def _send_error(self, error: SpeechAPIError) -> None:
+        await self._send_json(
```
zijiexia: Does this error format align with https://github.com/sgl-project/sglang-omni/issues/601?

sglang_omni/serve/protocol.py
```diff
@@ -314,6 +315,127 @@ class CreateSpeechRequest(BaseModel):
...
+    seed: int | None = None
+    stage_params: dict[str, dict[str, Any]] | None = None
+
+
+class SpeechBatchResult(BaseModel):
+    """One item result in a batch text-to-speech response."""
+
+    index: int
+    status: str
+    success: bool
```
zijiexia: are these two redundant?

Ratish1: Yes, they are. removed the redundant success field and derive batch counts from status now

sglang_omni/serve/speech_ws.py
Ratish1: Fixed it now and WebSocket errors now use top-level message/param/code fields instead of the nested REST error envelope

sglang_omni/serve/speech_ws.py
Ratish1: added the watcher completion check after the yield so disconnect/overflow is not swallowed

zhaochenyang20 (approved):

PR #624, opened 2026-05-31
[Docs] Add Ming-flash-omni-2.0 cookbook

edwingao28:
## Motivation

SGLang-Omni had no cookbook for Ming-flash-omni-2.0. This adds a customer-facing guide for launching Ming via the generic `sgl-omni serve --model-path` entry point (text, text+audio, and streaming-speech variants) and documents the supported `/v1/chat/completions` request shapes (text/image/audio/video in, text or text+audio out).

## Modifications

- `docs/cookbook/ming_omni.md` — new cookbook: prerequisites, architecture (3 pipeline variants), server configs, input/output examples, request parameters, SGLang-Omni benchmark tables, and known limitations.
- Benchmark section is SGLang-Omni-only: GSM8K thinker, MMMU image-text, non-streaming talker, streaming talker, and streaming-vs-non-streaming audio equivalence — all as concurrency-sweep tables on H100-class hardware.
- `docs/index.rst`, `README.md` — link the cookbook into the docs nav.
- `tests/docs/ming_omni/test_docs_ming_omni_static.py` — static checks: page is discoverable, documents supported launchers/request shapes, surfaces benchmark sections, preserves caveats.

## Checklist

- [x] Format your code according to the [Format code with pre-commit](https://docs.sglang.io/developer_guide/contribution_guide.html#format-code-with-pre-commit).
- [x] Follow the SGLang code style [guidance](https://docs.sglang.io/developer_guide/contribution_guide.html#code-style-guidance).

PR #670, opened 2026-06-03
Verify SGLang-Omni cookbook docs

MelodyyyYin:
### Summary

- Verified and updated SGLang-Omni model cookbooks for TTS, ASR, omni, and
  LLaDA2-Uni paths.
- Added missing cookbooks for Fish Audio S2-Pro, Qwen3-ASR, and Whisper ASR.
- Removed Ming-Omni docs/modeling changes from this PR; Ming-Omni is handled
  separately in #624.
- Kept a small LLaDA2-Uni scheduler fix because the cookbook path otherwise
  fails against the SGLang `PrefillAdder` API shipped in the current dev image.

### Notable Changes

- Documented current public request parameters for Higgs, Qwen3-TTS, Voxtral,
  Qwen3-Omni, Qwen3-ASR, Whisper ASR, MOSS-TTS, and Fish Audio S2-Pro.
- Removed unsupported Higgs public `reference_codes` / `reference_text`
  documentation; the public speech schema exposes `references`, `ref_audio`,
  and `ref_text`.
- Added a concise Whisper ASR cookbook and marked the path experimental
  (encoder batch size 1, prefer `response_format=json`).
- Re-verified Voxtral startup behavior: the `tts_generation` stage captures
  CUDA graphs in current code/runtime logs.
- Linked Qwen3-TTS CustomVoice and VoiceDesign to the Qwen3-TTS cookbook.
- Kept Qwen3-Omni launch commands under the selector and retained only small
  verified parameter fixes.

### LLaDA2-Uni runtime fix

- The dev image runs `sglang 0.0.0.dev1+gdce8b0606`, which is ahead of the
  `sglang==0.5.8` pin. In that build `PrefillAdder.__init__` no longer accepts
  `dllm_config` and `PrefillAdder.dllm_staging_reqs` is gone, so the old DLLM
  scheduler path raised before reaching the model.
- `dllm_scheduler.py` now targets the current `PrefillAdder` API
  (`prefill_max_requests=1`, staging tracked from `adder.can_run_list`).
- With this fix, **`/v1/chat/completions` works for LLaDA2-Uni**
  (text output). Verified live on the dev image (`sglang 0.0.0.dev1`): a
  text-only chat request returns a coherent reply with `finish_reason: stop`
  (HTTP 200). This replaces the earlier note that the call failed because the
  scheduler passed `dllm_config` to `PrefillAdder`.
- This change can be split into its own PR if maintainers want this one to be
  strictly docs-only.

### Validation

- Focused runtime/static checks for Whisper ASR and Voxtral startup behavior.
- Static schema/code checks for Higgs public speech fields and Qwen3-Omni
  parameters.
- Verified the LLaDA2-Uni `/v1/chat/completions` path live on the dev image
  (`sglang 0.0.0.dev1`): server starts, a text-only chat request returns a
  coherent reply with `finish_reason: stop` and HTTP 200.


MelodyyyYin:
# Cookbook Verification Update

- Fixed and validated the Ming-Omni cookbook path:
  - Added missing prerequisites.
  - Fixed the SGLang `LinearBase` import-order startup issue.
  - Added explicit speaker prompt support for checkpoints without `talker/data/voice_name.json`.
  - Forced Ming AudioVAE Qwen2 configs to eager attention so speech mode does not require `flash_attn`.
  - Validated H200 full multimodal text serving with `--tp-size 2`.
  - Validated H200 speech serving with thinker `--tp-size 2`, talker on a separate GPU, `--cpu-offload-gb 80`, and a prompt wav/text.
- Fixed and validated the LLaDA2-Uni cookbook path:
  - Updated the DLLM scheduler for the current SGLang `PrefillAdder` API.
  - Used DLLM block-size chunked prefill and per-request token result handling.
  - Disabled CUDA graph by default for the experimental DLLM thinker path.
  - Validated text and image+text chat-completion requests.
- Updated the Ming-Omni and LLaDA2-Uni cookbooks with the validated commands, request examples, and current limitations.
- Ran focused checks:
  - `py_compile` on changed Python files.
  - `black` on changed Python files.
  - `pytest tests/unit_test/pipeline/test_runtime_adapter.py` (`13 passed`).
  - `pytest tests/unit_test/ming_omni/test_omni_serve.py tests/unit_test/ming_omni/test_pipeline.py` (`59 passed`).
  - `sphinx-build -b html docs /data/tmp/cookbook_verify/docs_build` (succeeded with existing non-cookbook warnings).
  - `git diff --check` (clean).

docs/basic_usage/tts.md
zhaochenyang20, line 1: I think in some sense, we can delete this `tts.md`. What should we put for all the model's general part? @xinlij

docs/basic_usage/tts.md
```diff
@@ -22,7 +22,7 @@ uv pip install --no-deps qwen-tts==0.1.1
 
 | Model family | Example config | Request notes |
 |---|---|---|
-| Fish Speech S2-Pro | `examples/configs/s2pro_tts.yaml` | Supports plain TTS and voice cloning with `references` |
+| [Fish Speech S2-Pro](../cookbook/fishaudio_s2_pro.md) | `examples/configs/s2pro_tts.yaml` | Supports plain TTS and voice cloning with `references` |
 | [Voxtral TTS](../cookbook/voxtral_tts.md) | `examples/configs/voxtral_tts.yaml` | Uses `input`, `voice`, `response_format`, and `max_new_tokens`; use `--no-ref-audio` for SeedTTS benchmarking |
 | [Qwen3-TTS Base](../cookbook/qwen3_tts.md) | `examples/configs/qwen3_tts_0_6b.yaml`, `examples/configs/qwen3_tts_1_7b.yaml` | Requires reference audio through `ref_audio` or `references[0].audio_path`; `language` defaults to `auto` |
 | Qwen3-TTS CustomVoice | `examples/configs/qwen3_tts_0_6b_customvoice.yaml` | Text-only requests use the checkpoint speaker table; missing `voice` defaults to `Vivian` |
```
zhaochenyang20: Qwen3-TTS CustomVoice also need be linked to `qwen3_tts.md`

docs/basic_usage/tts.md
```diff
@@ -123,7 +123,7 @@ curl -X POST http://localhost:8000/v1/audio/speech \
 
 For natural-sounding Fish Speech S2-Pro results, use Voice Cloning with a reference audio clip.
 
-### Voice Cloning
```
zhaochenyang20, line 126: this is not `with python`. Curl is not python.

docs/cookbook/higgs_tts.md
```diff
@@ -16,7 +16,7 @@ reference clip, and fine-grained inline control over emotion, style, sound effec
 
 ![Higgs Audio v3 Generation Architecture](../_static/image/higgs-architecture.png)
 
-Higgs autoregressive decoder consumes interleaved text and audio tokens. Audio is encoded by the **Higgs Tokenizer** into 8 codebooks at 25 fps, staggered via a **delay pattern**, then mapped to backbone hidden states through a **multi-codebook fused embedding**. Output codes pass through a **multi-codebook fused head**, are de-delayed, and decoded back to waveform. Multi-turn generation interleaves `<|text|>…<|audio|>…` chunks so each new chunk is grounded on reference + prior chunks.
+Higgs consumes interleaved text and audio tokens. Reference audio is encoded into 8 codebooks, the decoder generates delayed codebook tokens, and the vocoder reconstructs 24 kHz speech.
```
zhaochenyang20: why change this?

docs/cookbook/higgs_tts.md
```diff
@@ -137,15 +137,7 @@ Reference output:
...
-
-Higgs TTS implements streaming via [Server-Sent Events (SSE)](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events). Each SSE event carries a base64-encoded WAV chunk. Your client can decode and play each chunk as it arrives, rather than buffering the entire response.
-
-Enable streaming by setting `"stream": true` in the request body. During generation, the vocoder emits incremental audio chunks; the terminal event is intentionally slim and carries metadata such as `sample_rate` and `usage` instead of repeating the full waveform. Inside the pipeline, audio chunks use the compact `audio_waveform` payload (`bytes` plus `audio_waveform_shape`, `audio_waveform_dtype`, and `sample_rate`), which the HTTP layer encodes into the SSE `audio.data` field.
-
-1. Use curl
-
-Set `"stream": true` in your request body:
+Set `"stream": true` to receive audio chunks over Server-Sent Events (SSE). Audio events carry base64-encoded audio in `audio.data`; the final metadata event has `audio: null`, then the stream ends with `data: [DONE]`.
```
zhaochenyang20: why change this?

docs/cookbook/higgs_tts.md
```diff
@@ -161,48 +153,6 @@ curl -N -X POST http://localhost:8000/v1/audio/speech \
 ```
 The `-N` flag disables curl's output buffering so SSE events are printed as they arrive.
 
-2. Use Python
```
zhaochenyang20, line 186: why delete this?

docs/cookbook/higgs_tts.md
```diff
@@ -404,21 +354,6 @@ Reference output:
 | `<\|prosody:expressive_high\|>` | More expressive delivery |
 | `<\|prosody:expressive_low\|>` | Flatter delivery |
 
-### Pre-encoded reference codes
```
zhaochenyang20, line 522: I think we can not delete this. But do we actually verified this? @xinlij

docs/cookbook/higgs_tts.md
```diff
@@ -428,8 +363,7 @@ resp = requests.post(
 | `response_format` | string | `"wav"` | Output audio format |
 | `stream` | bool | `false` | Enable streaming via SSE |
 | `references` | list | `null` | Reference audio for voice cloning; each item has `audio_path` (local path or HTTP URL) and `text` (transcript) |
-| `reference_codes` | list[list[int]] | `null` | Pre-encoded discrete codes, shape `[T, 8]` — alternative to `references[0].audio_path` |
-| `reference_text` | string | `null` | Transcript of reference audio when supplying `reference_codes` |
```
zhaochenyang20, line 547: good catch. Please verify this, what's the actually parameter name.

docs/cookbook/higgs_tts.md
```diff
@@ -439,8 +373,6 @@ resp = requests.post(
 
 ### Throughput
 
-[TODO (yichi, Huapeng): This should be updated in the last minute.]
```
zhaochenyang20, line 557: keep this todo.

docs/cookbook/ming_omni.md
zhaochenyang20, line 1: I think there is one PR from @edwingao28 should we merge that instead? @xinlij

docs/cookbook/qwen3_omni.md
```diff
@@ -21,7 +21,39 @@ uv pip install -v .
...
+python examples/run_qwen3_omni_speech_server.py \
+  --model-path Qwen/Qwen3-Omni-30B-A3B-Instruct \
+  --gpu-thinker 0 \
+  --gpu-talker 1 \
+  --gpu-code-predictor 1 \
+  --gpu-code2wav 1 \
+  --port 8000
+```
+
+Use the selector below for other hardware, TP, topology, and precision combinations.
```
zhaochenyang20: why do not we put these launching command under the selector?

docs/cookbook/qwen3_tts.md
```diff
@@ -53,10 +57,9 @@ sgl-omni serve \
 
 ## Synthesizing Speech
 
-### Zero-shot
-
+### Text-only Requests
 
-Qwen3-TTS does not support zero-shot synthesis.
+Qwen3-TTS Base checkpoints require a reference clip. Text-only requests are supported by CustomVoice and VoiceDesign checkpoints; see [TTS Model Usage](../basic_usage/tts.md) for those launch commands.
```
zhaochenyang20, line 62: good catch

docs/cookbook/voxtral_tts.md
```diff
@@ -23,6 +23,7 @@ The model repository is public, so no Hugging Face token is required.
 ## Server Configuration
 
 The pipeline is `preprocessing → tts_generation → vocoder`.
+First startup can take several minutes while the `tts_generation` stage captures CUDA graphs.
```
zhaochenyang20, line 26: this is not true. You should verify this. CUDA Graph is not captured here.

docs/cookbook/whisper_asr.md
```diff
@@ -0,0 +1,71 @@
...
+| `response_format` | string | `json` | Use `json`; `text` and `verbose_json` can trigger a 500 response |
+| `temperature` | float | unset | Leave unset in the current tree; explicit values can trigger a 500 response |
+
+The request builder also supports `task` (`transcribe` by default) and `max_new_tokens`, but the public transcription endpoint currently exposes only the fields above. The route uses the ASR stage default unless the pipeline is configured another way. For smoke tests, keep the request minimal and use `response_format=json`.
+
+## Known Limitations
+
+- This path is not yet correctness-validated. On the current test environment, `openai/whisper-large-v3` starts on a clean H200 and returns the OpenAI-compatible response schema, but short local smoke-test clips returned an empty `text` field.
+- Non-`json` response formats, explicit `language`, and explicit `temperature` can fail with `500` (`'NoneType' object has no attribute 'view'`). After that failure, the same server process can continue returning 500s for later requests; restart the server before retrying.
+- First startup can spend several minutes in torch compile / CUDA graph capture.
```
zhaochenyang20: This is not true. Verify it your own.

docs/cookbook/whisper_asr.md
```diff
@@ -0,0 +1,71 @@
...
+|---|---|---|---|
+| `file` | file | required | Audio file uploaded as multipart form data |
+| `model` | string | server default | Model identifier |
+| `language` | string | unset | Leave unset in the current tree; explicit values can trigger a 500 response |
+| `response_format` | string | `json` | Use `json`; `text` and `verbose_json` can trigger a 500 response |
+| `temperature` | float | unset | Leave unset in the current tree; explicit values can trigger a 500 response |
+
+The request builder also supports `task` (`transcribe` by default) and `max_new_tokens`, but the public transcription endpoint currently exposes only the fields above. The route uses the ASR stage default unless the pipeline is configured another way. For smoke tests, keep the request minimal and use `response_format=json`.
+
+## Known Limitations
```
zhaochenyang20, line 66: One thing important left:

whisper only support batch size = 1 on the encoder.

examples/run_ming_omni_speech_server.py
```diff
@@ -212,10 +227,20 @@ def _launch_speech_server(args: argparse.Namespace) -> None:
...
     _apply_stage_factory_updates(
         config,
         stage_name=talker_stage_name,
-        updates={"voice": args.voice},
+        updates={
+            "voice": args.voice,
+            "prompt_wav_path": prompt_wav_path,
+            "prompt_text": getattr(args, "prompt_text", None),
+        },
     )
```
zhaochenyang20: why do you change ming's modling?

zhaochenyang20 (changes requested):
If this is a pure docs PR, why you change the modeling files?

docs/cookbook/ming_omni.md
xinlij: Yes, there is no need to add or modify any Ming-omni related documentation in this PR; that will be handled by #624.

docs/cookbook/higgs_tts.md
xinlij: This was originally derived from Higgs's model card; however, it appears this section has been removed in the latest version. Could @SilinMeng0510  please confirm this?

docs/basic_usage/tts.md
xinlij: I don't have any specific ideas at the moment. We can place the general content here for now, and temporarily rename this section from "Basic Usage" to "General Usage". Do you have any thoughts on this?

xinlij (changes requested):

MelodyyyYin:
I re-checked the latest PR review comments and current diff against `upstream/main`.

- Ming-Omni: resolved. There are no Ming docs, server, or modeling changes left
  in the current PR diff; Ming-Omni is left for #624.
- Higgs TTS: resolved. The architecture text, streaming wording, Python example,
  and throughput TODO are back to the main-branch structure. The only remaining
  Higgs change is the verified public-API correction from unsupported top-level
  `reference_codes` / `reference_text` to `ref_audio` / `ref_text`.
- Qwen3-Omni: resolved. Launch commands are back under the selector; only the
  verified request-parameter fixes remain.
- TTS general page: resolved. Qwen3-TTS CustomVoice and VoiceDesign link to
  `qwen3_tts.md`, the curl heading no longer says Python, and the sidebar caption
  is now `General Usage`.
- Voxtral TTS: resolved. The CUDA graph startup note was re-verified in code and
  live H200 logs, so it remains.
- Whisper ASR: resolved. The cookbook follows the same concise structure as the
  ASR cookbook and no longer includes environment-specific smoke-test results.
  It documents only user-facing limitations, including encoder batch size 1.
- Modeling files: mostly resolved. Ming modeling changes are gone. The remaining
  non-doc change is the small LLaDA runtime blocker fix needed for the verified
  LLaDA cookbook path; split it out if the maintainers want this PR to be
  strictly docs-only.

docs/basic_usage/tts.md
MelodyyyYin: How about keep it as a concise "General TTS Usage" page with a model capability matrix and common endpoint examples, while moving model-specific launch details and benchmark details into each cookbook?

sglang_omni/scheduling/dllm_scheduler.py
```diff
@@ -149,8 +152,8 @@ def _schedule_next_batch(self) -> ScheduleBatch | None:
             None,  # running_batch
             0.5,  # new_token_ratio
             self.server_args.max_prefill_tokens,
-            self.server_args.chunked_prefill_size,
-            dllm_config=self.dllm_config,
+            self._chunked_prefill_size,
+            prefill_max_requests=1,
```
JiaxinD, line 156: This is a real behavior change riding in a docs PR, so a few questions. Which sglang build hit the original PrefillAdder error? The pinned sglang==0.5.8 still accepts dllm_config (and dllm_staging_reqs still exists there, schedule_policy.py L380/L432), and even current sglang main still takes dllm_config, only dllm_staging_reqs is gone. If the runtime image is ahead of the pin, bumping the pin seems cleaner than coding against an unpinned API.

Separately, prefill_max_requests=1 plus re-adding every staging req through add_chunked_req means the stage now runs one request at a time (add_one_req returns OTHER once can_run_list has an entry), whereas the old dllm path budgeted max_running_requests * block_size. Intended for the experimental path? If that cap is ever lifted, the flat-list branch in _apply_results below will zip ints and break, so worth a comment there at least.

Also the PR body still says /v1/chat/completions fails on exactly the call this diff fixes, worth refreshing it to say whether chat works now.

docs/cookbook/fishaudio_s2_pro.md
```diff
@@ -0,0 +1,92 @@
...
+| `top_p` | float | `0.8` | Top-p sampling |
+| `top_k` | int | `30` | Top-k sampling; must be `-1` or between `1` and `30` |
+| `repetition_penalty` | float | `1.1` | Repetition penalty |
+
+## Validation
+
+S2-Pro has docs tests that exercise the same `/v1/audio/speech` examples:
+
+```bash
+pytest tests/docs/s2pro/test_docs_tts_s2pro.py -s -x
```
JiaxinD: tests/docs/ does not exist anywhere in the repo, so this pytest command fails at collection. Point it at an existing test or drop the Validation section for now.

docs/cookbook/qwen3_tts.md
```diff
@@ -16,14 +16,17 @@ Qwen3-TTS Base uses the upstream `qwen-tts` package, which currently pins Transf
 Install it only in environments that serve Qwen3-TTS:
 
 ```bash
-uv pip install transformers==4.57.3 accelerate==1.12.0 sox einops
+sudo apt-get update && sudo apt-get install -y sox
```
JiaxinD: The recommended dev image runs as root without sudo installed, so this exact line fails with command not found inside the container. Plain apt-get update && apt-get install -y sox is safer.

sglang_omni/scheduling/dllm_scheduler.py
MelodyyyYin: On which build hit the error and the pin: confirmed by inspecting the running dev image (`docker exec sglang-omni-YueYin`) that it ships `sglang 0.0.0.dev1+gdce8b0606`, ahead of the `sglang==0.5.8` pin, and in that build `PrefillAdder.__init__` does not accept `dllm_config`, `PrefillAdder` no longer exposes `dllm_staging_reqs`, and it does expose `prefill_max_requests`, so against the build this branch actually runs on the old code (`dllm_config=...`, `adder.dllm_staging_reqs`) raises before any forward pass and your assumption that current main still takes `dllm_config` does not hold here; we can't cleanly bump the pin because the image ships an untagged dev build rather than a released tag, so coding against the current `PrefillAdder` API is what makes the path run, and I'm happy to split this into its own PR if you'd rather keep this one docs-only. On `prefill_max_requests=1` and `_apply_results`: that is intended for now since this experimental DLLM stage processes one request per batch, and I added a comment in `_apply_results` documenting the `prefill_max_requests=1` assumption and explicitly warning that the flat-list normalization must be revisited together with the scheduling cap if it is ever raised above 1, otherwise the `zip(batch.reqs, ...)` would pair each `Req` with a single int. On the PR body: good catch, I refreshed it — verified live on the dev image (`sglang 0.0.0.dev1`) that a text-only `/v1/chat/completions` request to LLaDA2-Uni returns a coherent reply with `finish_reason: stop` (HTTP 200) after this scheduler fix, so the body no longer claims the call fails.

zhaochenyang20 (approved):

Issue #637, opened 2026-06-01
MOSS TTS Performance Optimization Roadmap


gaoyang07:
> **NOTE: To track the latest MOSS-TTS changes after the [MOSS-TTS Local Transformer v1.5 launch post](https://www.lmsys.org/blog/2026-06-17-moss-tts-local-v15), follow #891 and later.**

## Goal

Low latency and high throughput with no model performance reduction

We maintain two MOSS-TTS architecture tracks under the same tokenizer and pretraining recipe to study the impact of token modeling patterns in discrete speech modeling:

- **Delay Pattern**: the primary scalable architecture, using a simpler single-backbone multi-head design. It is better suited for large model scaling, long-context generation, duration/pronunciation control, and optimized inference backends.
    - Model arch and weights: https://huggingface.co/OpenMOSS-Team/MOSS-TTS-v1.5
- **Local Transformer**: a quality-oriented variant with an extra frame-local AR module. It adds architectural complexity but improves modeling efficiency, showing lower token losses and stronger speaker similarity at smaller model scale.
    - Model arch and weights: https://huggingface.co/OpenMOSS-Team/MOSS-TTS-Local-Transformer

<img width="949" height="580" alt="Image" src="https://github.com/user-attachments/assets/4fba155f-3015-4842-94d2-983834fd1583" />

Help wanted: https://github.com/sgl-project/sglang-omni/issues/607

## Done

- [x] Delay-pattern model support with radix cache #609 @GaokaiZhang @0xjeffro 
    - Earlier attempts: #608 @sjmshsh
    - [x] AR CUDA Graph + torch.compile
    - [x] Doc implementation
    - [x] Cookbook
    - [x] benchmark performance in H200

### Optimization

Baseline: require RTF < 1 for concurrency 16 on H100

**Model support**
- [x] MOSS-TTS-Local-Transformer-v1.5 #728 @GaokaiZhang
    - legacy 1.0: Local-transformer arch model with radix cache #716 @draftbk
    - [ ] Playgroud
    - [x] Cookbook #792 @MelodyyyYin
    - [ ] Doc implementation
    - [ ] benchmark performance in H100/H200

**Pipeline**
- [x] decouple preprocessing and ref audio encoding #730 @sjmshsh V1 pr: #699 @AkazaAkane
- [x] preprocessing, AR, and vocoder in the default single-GPU pipeline #810 @Ratish1 

**Encoder**
[...]

PR #674, opened 2026-06-03
Feat/moss streaming

BruceLoveDecimal:
## Motivation

Implements the MOSS-TTS streaming AR chunk + streaming vocoder path for #637, building on the MOSS pipeline added in #609.

The goal is to support low-latency streaming speech output without relying on unverified remote-code codec streaming APIs. The implementation follows the existing Higgs/Fish pattern: AR emits delayed audio-code rows incrementally, and the vocoder performs stateless overlap-window re-decode using the existing `processor.decode_audio_codes` API.

## Modifications

- Added MOSS AR streaming output from `tts_engine` to `vocoder`.
- Added MOSS streaming vocoder scheduler with overlap-window incremental decode.
- Ensured streaming terminal payload is treated as an end signal, avoiding double decode / duplicate audio.
- Added TP/process-split safe MOSS prepared-request handoff via inline `StagePayload`.
- Added TP args support for the MOSS AR factory: `tp_rank`, `tp_size`, `nccl_port`, and `total_gpu_memory_fraction`.
- Added unit tests for streaming vocoder behavior, terminal guard, multi-segment parity, abort cleanup, and TP/process-split prepared handoff.

## Related Issues

Part of #637.  
Builds on #609.

## Accuracy Test

No model weights, kernels, or architecture math were changed.

E2E functional checks passed for streamed vs non-streamed MOSS-TTS outputs, including short text, long narration, pause-marker text, Chinese text, and terminal-payload guard cases. Streamed and non-streamed output durations matched exactly in the audited cases.

### TP=1 Audit

Remote E2E audit on single H20, TP=1: PASS.

Functional streaming latency:

| Case | Audio Duration | First Chunk Latency | Chunks |
| --- | ---: | ---: | ---: |
| short_plain_64 | 5.20s | 1.745s | 9 |
| long_narration_192 | 23.92s | 1.691s | 38 |
| pause_markers_96 | 7.76s | 1.666s | 13 |
| zh_text_128 | 10.32s | 1.572s | 17 |
| terminal_guard_128 | 10.32s | 1.665s | 17 |

Concurrency benchmark:

| Concurrency | QPS | TTFA p50 | TTFA p95 | RTF p50 |
| 1 | 0.388 | 1.694s | 1.694s | 0.4955 |
| 2 | 0.650 | 1.872s | 1.917s | 0.5867 |
| 4 | 1.011 | 1.918s | 2.046s | 0.7464 |
| 8 | 1.371 | 2.303s | 2.561s | 1.0886 |

### TP=2 Audit

Remote E2E audit on dual H20, TP=2: PASS.

Functional streaming latency:

| Case | Audio Duration | First Chunk Latency | Chunks |
| --- | ---: | ---: | ---: |
| short_plain_64 | 5.20s | 1.571s | 9 |
| long_narration_192 | 22.48s | 1.224s | 36 |
| pause_markers_96 | 7.76s | 1.189s | 13 |
| zh_text_128 | 10.32s | 1.245s | 17 |
| terminal_guard_128 | 10.32s | 1.226s | 17 |

Concurrency benchmark:

| Concurrency | QPS | TTFA p50 | TTFA p95 | RTF p50 |
| 1 | 0.432 | 1.199s | 1.199s | 0.4453 |
| 2 | 0.692 | 1.499s | 1.522s | 0.5514 |
| 4 | 1.049 | 1.880s | 1.954s | 0.7191 |
| 8 | 1.277 | 2.996s | 3.161s | 1.1726 |

## Checklist

- [x] Add unit tests.


sglang_omni/models/moss_tts/streaming_vocoder.py
JingwenGu0829, line 1: This scheduler ends up very close to the Higgs one — the `_*StreamState`, the metadata latching, the
overlap-window/holdback decode math, and `_build_usage` (which is identical across MOSS/Higgs/Fish). Since we already have `models/tts_streaming.py` for shared TTS streaming helpers,  it would be worth pulling the common pieces up there .

This could also potentially fix the correctness issue in the file that codex review found: 


### sglang_omni/models/moss_tts/streaming_vocoder.py (`on_streaming_new_request`, ~lines 857-862)
I think there's a correctness issue here. In the running pipeline the AR stage sends the vocoder its stream
chunks and the `stream_done` *before* the terminal payload (that's what `can_accept_stream_before_payload`
is for, and it's what `_route_result` does — `stream_done` then payload, both after the chunks). The base
scheduler holds that `stream_done` until the payload shows up, and then replays it right after
`on_streaming_new_request`. Because this method always creates a brand-new empty `_MossStreamState`, the rows
we accumulated from the chunks get thrown away just before the final flush — so we lose the held-back tail on
every stream, and for short inputs we can lose all the audio. Higgs and Fish both use
`self._stream_states.setdefault(...)` here for exactly this reason (and the base lifecycle already clears
state on abort/completion, so the "start clean" comment is covered). Could we switch to `setdefault` and only
update `sample_rate`, and add a test where the chunks + done arrive before `new_request`? That ordering is
the one the pipeline actually produces, and our current tests do it the other way around, which is why this
didn't show up.

sglang_omni/models/moss_tts/request_builders.py
```diff
@@ -483,12 +520,11 @@ def preprocess_moss_tts_payload(payload: StagePayload) -> StagePayload:
         _INFLIGHT_REQUESTS.discard(rid)
         aborted = rid in _ABORTED_REQUESTS
         _ABORTED_REQUESTS.discard(rid)
-        if not aborted:
-            # Aborted-while-preprocessing drops the handoff so it never lingers.
-            _PREPARED_REQUESTS[rid] = prepared
 
     data = prepared.state.to_dict()
     data[_MOSS_TTS_PREPARED_MARKER] = payload.request_id
```
JingwenGu0829: After the move to the inline handoff, it looks like nothing writes `_PREPARED_REQUESTS` anymore — so the
marker branch in `pop_prepared_moss_tts_request` always returns `None`, the marker is still stamped into
every payload but never read, and the "same-process fallback" comment doesn't quite match the code now.
Shall we drop the dict + marker + dead branch and keep just the inline path?

sglang_omni/models/moss_tts/request_builders.py
```diff
@@ -538,6 +580,122 @@ def _resolve_audio_payload_bounds(
...
+        params = getattr(getattr(stage_payload, "request", None), "params", None)
+        if not isinstance(params, dict) or not bool(params.get("stream", False)):
+            return []
+
+        req = getattr(req_data, "req", None)
+        if req is not None and int(getattr(req, "is_chunked", 0) or 0) > 0:
+            return []
+
+        cfg = getattr(req_data, "model_config", None)
+        output_rows = getattr(req_data, "output_rows", None)
```
JingwenGu0829, line 698: This runs every decode step and re-stacks the whole `output_rows` history (GPU tensors) and re-finds the
bounds with `.item()` each time. Since the audio start index doesn't move and `streamed_row_count` already
tracks progress, could we cache `start` and only scan the new rows? That drops a per-step GPU→host sync on
the decode loop and the growing re-stack.

JingwenGu0829 (commented):
Left some comments hoping it could help with streaming! 

Since there's actually correctness issue with the current implementation, it would be great if you could also provide the correctness benchmark with the PR after your fix (for exmaple, the WER) Thanks!

BruceLoveDecimal:
> Left some comments hoping it could help with streaming!
> 
> Since there's actually correctness issue with the current implementation, it would be great if you could also provide the correctness benchmark with the PR after your fix (for exmaple, the WER) Thanks!

Thanks for the review! Good catch on the stream lifecycle ordering. I mostly tested the payload-first path locally, but the real pipeline can deliver chunks + stream_done before the terminal payload, so the current `on_streaming_new_request` cleanup logic is too eager.

I’ll fix this by preserving any pre-payload `_MossStreamState` with `setdefault()` and only updating payload-derived metadata such as sample rate when the terminal payload arrives. And I’ll add a regression test for the bad ordering case:
1. stream chunks arrive first,
2. `stream_done` is held by the scheduler,
3. terminal payload arrives,
4. final flush still uses the accumulated rows and emits the held-back tail.

For correctness benchmarking, I’ll also add a supporting real-model content correctness guardrail, using the same MOSS/repo benchmark style: compare `stream=False` vs `stream=True` and report:

- EN WER / ZH CER
- evaluated / skipped / failed / outlier counts
- insertion / deletion / repetition-style failure summaries where available

This should make the correctness impact and the streaming-output behavior explicit. If there are any other problem, please feel free to ping me.

BruceLoveDecimal:
Hi @JingwenGu0829, I have updated the PR based on the review feedback:
1. Correctness fix: `streaming_vocoder.py` now preserves an existing per-request stream state when the terminal payload arrives after `stream_done`. Previously `on_streaming_new_request` always created a fresh `_MossStreamState`, which could drop rows accumulated from earlier chunks and lose the final flush tail.
2. Stream builder optimization: cached three fields:
3. Added unit test coverage:
4. And following the existing evaluation practice in sglang-omni, I ran seed-tts-eval as the correctness benchmark on H20. The results are as follows:

Lang | Mode | Success | WER corpus | WER mean | WER p95 | WER max | TTS latency | RTF | Streaming Metrics
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
EN | nonstream | 50/50 | 1.60% | 1.55% | 12.50% | 20.00% | 4.384s | 1.0904 | -
EN | stream | 50/50 | 1.95% | 2.00% | 12.96% | 25.00% | 6.015s | 1.4787 | TTFP 4.488s, chunks 7.18
ZH | nonstream | 50/50 | 0.86% | 0.81% | 4.55% | 12.50% | 4.634s | 0.9194 | -
ZH | stream | 50/50 | 1.40% | 1.36% | 6.48% | 17.65% | 6.996s | 1.3886 | TTFP 4.913s, chunks 8.40

JingwenGu0829:
@BruceLoveDecimal Thanks for the benchmark and fix! However I think the WER and latency under streaming mode are both facing a abnormal drop (emprically streaming mode won't cause such much downgrade in performance and precision. Empirically the TTFT should not be ~4s too. 

 Would you mind taking a further self-investigation? Or I can find some experienced developers to help with this branch, or I can help myself as I have free time next week.

BruceLoveDecimal:
Hi @JingwenGu0829, thanks for pointing this out. I also noticed this yesterday and did some initial investigation.

For WER/CER, the current benchmark may not be a fully controlled streaming-vs-non-streaming comparison. The run used seed=null, so streaming and non-streaming were generated as two independent sampling runs. Since MOSS-TTS uses sampling by default, the observed WER/CER delta may include generation randomness and ASR variance, not only streaming quality regression.

For TTFT, I found that MOSS-TTS has a structural first-audio delay because its delayed RVQ layout uses n_vq=32, so the streaming vocoder cannot emit the first complete audio frame until enough delayed rows are available. I also found one scheduling issue where the first chunk could be delayed by an extra follow-up stride after holdback; I’m working on fixing that, but the remaining TTFT still needs deeper investigation.

I’ll continue looking into both the benchmark methodology and the streaming path. Once I have a confirmed conclusion, I’ll update the PR with the details.

GaokaiZhang:
Here are some of my replication results.

DP=1, no router, `moss_tts.yaml`, `--ref-format references --token-count auto`, c=8, **`--seed 42` for both modes**, ASR=Qwen3-ASR-1.7B, full set (EN=1088/ZH=2020), 1×H200. Fixed seed removes the `seed=null` confound (that's the earlier "WER drop").

| Lang | Mode | Success | WER/CER corpus | mean | p95 | max | latency | RTF | Streaming |
|---|---|---|---|---|---|---|---|---|---|
| EN | nonstream | 1088/1088 | 1.53% | 1.49% | 9.09% | 118.18% | 4.12s | 0.97 | — |
| EN | stream | 1088/1088 | 1.78% | 1.73% | 10.00% | 118.18% | 6.67s | 1.56 | TTFP 5.07s, chunks 7.5 |
| ZH | nonstream | 2020/2020 | 1.14% | 1.12% | 5.57% | 105.56% | 4.31s | 0.76 | — |
| ZH | stream | 2020/2020 | 1.60% | 1.57% | 8.70% | 105.56% | 8.10s | 1.42 | TTFP 6.18s, chunks 9.5 |

(EN=WER, ZH=CER.) No abnormal regression — non-stream is in the normal MOSS band (cf. PR #609 EN 1.93%).

**Content: small consistent vocoder penalty.** Stream is ~0.3–0.5 pp worse in every pair. A **c=1 control** (batch=1 ⇒ identical AR tokens, isolating the vocoder) reproduces it: EN WER 0.85→1.28, ZH CER 0.25→0.75. Mechanism: each overlap window is an *independent* codec `decode()` with cold conv/transformer state and only `stream_overlap_tokens=2` frames of left-context (vs full context in the single decode) → small seam deviations. Caveat: at n=64 the control is only *suggestive*; the same-direction full set backs it. `test_..._matches_full_decode` can't catch this (its fake decoder has a 1-frame receptive field).

**TTFA — the "~4 s TTFT" is reference-audio encoding, not the streaming vocoder.** Single warm request, `token_count=16` (1.36 s audio):

| request | first audio / total |
|---|---|
| non-stream, **no ref** | 0.53 s |
| non-stream, local-file ref | 2.76 s |
| non-stream, URL ref | 3.4 s |
| stream, **no ref** (tc=64) | 0.50 s (total 1.25 s) |
| stream, with ref | ~3.0–3.2 s |

The ~2.2 s is synchronous per-request reference encoding (same ref 3× = 2.76 s each — not cached), +~0.6 s network for URL refs. **Non-stream pays it identically**, so the ~3.2 s floor is pre-existing (from #609), not the streaming path. seed-TTS-eval clones every prompt, so TTFA ≈ full latency and is length-independent (~3.2 s for 1.4 s *and* 16 s of audio). Drop the ref and streaming actually helps (0.5 s vs 1.25 s full). c=8 eval TTFP (EN 5.07/ZH 6.18 s) = this floor + queueing.

**Throughput regression** (c=8 RTF 0.97→1.56 EN / 0.76→1.42 ZH, qps −38/−47%): *not* whole-window re-decode — the decode window is incremental (`window_start=emitted_frames-overlap`, ~1.22× frames total). For long outputs the cost is the per-chunk **re-stack of the full delayed-row history + `apply_de_delay_pattern`/`split_moss_audio_segments`** over the growing tensor (O(N²)); for short outputs it's the ~N small `decode_audio_codes` calls + the float32 vocoder sharing the AR's GPU.

**Suggestions (ignore if I'm wrong)**
1. The dominant TTFA cost is the pre-existing ~2.2 s ref encode, re-paid per request — cache it per ref-hash (helps repeated-voice streaming) / overlap with prefill. This, not the vocoder, is what gates first-audio latency.
2. De-delay/segment only the new tail instead of re-stacking `delayed_rows[:rows_end]` each chunk — removes the O(N²) host work (dominant for long-form). For short clips, fewer/larger decode calls + not colocating the vocoder on the AR GPU is the bigger RTF lever.
3. First chunk fires at `n_vq+followup_stride=40`; set `next_decode_rows=n_vq+holdback`(=33) in the `emit_until_raw<=0` branch → ~7 rows sooner (minor; only visible once the ref floor is gone).
4. Content penalty: more left-context / persist codec state across windows to cut cold-context seam artifacts, and add a parity test against the real codec.

gaoyang07:
> Here are some of my replication results.
> 
> DP=1, no router, `moss_tts.yaml`, `--ref-format references --token-count auto`, c=8, **`--seed 42` for both modes**, ASR=Qwen3-ASR-1.7B, full set (EN=1088/ZH=2020), 1×H200. Fixed seed removes the `seed=null` confound (that's the earlier "WER drop").
> 
> Lang	Mode	Success	WER/CER corpus	mean	p95	max	latency	RTF	Streaming
> EN	nonstream	1088/1088	1.53%	1.49%	9.09%	118.18%	4.12s	0.97	—
> EN	stream	1088/1088	1.78%	1.73%	10.00%	118.18%	6.67s	1.56	TTFP 5.07s, chunks 7.5
> ZH	nonstream	2020/2020	1.14%	1.12%	5.57%	105.56%	4.31s	0.76	—
> ZH	stream	2020/2020	1.60%	1.57%	8.70%	105.56%	8.10s	1.42	TTFP 6.18s, chunks 9.5
> (EN=WER, ZH=CER.) No abnormal regression — non-stream is in the normal MOSS band (cf. PR #609 EN 1.93%).
> 
> **Content: small consistent vocoder penalty.** Stream is ~0.3–0.5 pp worse in every pair. A **c=1 control** (batch=1 ⇒ identical AR tokens, isolating the vocoder) reproduces it: EN WER 0.85→1.28, ZH CER 0.25→0.75. Mechanism: each overlap window is an _independent_ codec `decode()` with cold conv/transformer state and only `stream_overlap_tokens=2` frames of left-context (vs full context in the single decode) → small seam deviations. Caveat: at n=64 the control is only _suggestive_; the same-direction full set backs it. `test_..._matches_full_decode` can't catch this (its fake decoder has a 1-frame receptive field).
> 
> **TTFA — the "~4 s TTFT" is reference-audio encoding, not the streaming vocoder.** Single warm request, `token_count=16` (1.36 s audio):
> 
> request	first audio / total
> non-stream, **no ref**	0.53 s
> non-stream, local-file ref	2.76 s
> non-stream, URL ref	3.4 s
> stream, **no ref** (tc=64)	0.50 s (total 1.25 s)
> stream, with ref	~3.0–3.2 s
> The ~2.2 s is synchronous per-request reference encoding (same ref 3× = 2.76 s each — not cached), +~0.6 s network for URL refs. **Non-stream pays it identically**, so the ~3.2 s floor is pre-existing (from #609), not the streaming path. seed-TTS-eval clones every prompt, so TTFA ≈ full latency and is length-independent (~3.2 s for 1.4 s _and_ 16 s of audio). Drop the ref and streaming actually helps (0.5 s vs 1.25 s full). c=8 eval TTFP (EN 5.07/ZH 6.18 s) = this floor + queueing.
> 
> **Throughput regression** (c=8 RTF 0.97→1.56 EN / 0.76→1.42 ZH, qps −38/−47%): _not_ whole-window re-decode — the decode window is incremental (`window_start=emitted_frames-overlap`, ~1.22× frames total). For long outputs the cost is the per-chunk **re-stack of the full delayed-row history + `apply_de_delay_pattern`/`split_moss_audio_segments`** over the growing tensor (O(N²)); for short outputs it's the ~N small `decode_audio_codes` calls + the float32 vocoder sharing the AR's GPU.
> 
> **Suggestions (ignore if I'm wrong)**
> 
> 1. The dominant TTFA cost is the pre-existing ~2.2 s ref encode, re-paid per request — cache it per ref-hash (helps repeated-voice streaming) / overlap with prefill. This, not the vocoder, is what gates first-audio latency.
> 2. De-delay/segment only the new tail instead of re-stacking `delayed_rows[:rows_end]` each chunk — removes the O(N²) host work (dominant for long-form). For short clips, fewer/larger decode calls + not colocating the vocoder on the AR GPU is the bigger RTF lever.
> 3. First chunk fires at `n_vq+followup_stride=40`; set `next_decode_rows=n_vq+holdback`(=33) in the `emit_until_raw<=0` branch → ~7 rows sooner (minor; only visible once the ref floor is gone).
> 4. Content penalty: more left-context / persist codec state across windows to cut cold-context seam artifacts, and add a parity test against the real codec.

From the benchmark results, we can probably consider merging #648 and #699 first.

Both changes are relatively lightweight, and they should have a meaningful impact on the current streaming bottleneck, especially the reference-audio encoding latency that dominates TTFA/TTFP. After those land, we can rebase and rerun the streaming benchmarks with the updated baseline.

That should give us a cleaner measurement of the remaining streaming-specific overhead, separate from the pre-existing reference-encoding cost.

luojiaxuan:
Move to #1311

PR #676, opened 2026-06-03
feat: add global --tp flag with model-specific per-stage expansion

### LINKED ISSUE #672: [RFC] Encoder TP follow-up: rank-colocated placement, whole-GPU memory, and sharded handoff
## Scope - what this issue tracks

Follow-up to #375 and #615.

#615 gives us the first Path-B baseline for SGLang-backed Qwen3-Omni image/audio Encoder TP. It shows that separate encoder placement can reduce image encoder per-rank peak and allow some long video/audio requests that TP1 rejects by modeled encoder admission cost.

The remaining design problem is colocated placement. Encoder TP admission cost and encoder process peak are not the same thing as physical whole-GPU peak. In colocated layouts, thinker/talker/code2wav resident memory, AR KV/activation, encoder staging, tensor broadcast buffers, CUDA cache, and concurrent image/audio work all stack on the same physical GPU.

This issue separates work that can reasonably land in #615 from follow-up work that should happen after #615 merges.

## Key distinction

We should separate two concepts that are easy to mix up:

- **Rank-colocated placement** means `GPU i = encoder_rank_i + thinker_rank_i`. This is only a placement / co-residency claim: the two ranks live on the same physical GPU.
- **Rank-aligned sharded handoff** means `encoder_rank_i` produces an output shard that can directly feed `thinker_rank_i`. This is a producer-consumer partition contract, not something TP gives automatically.

Both stages using TP is not enough to imply sharded handoff. Standard VLM paths often materialize full multimodal embeddings before feeding the language model, and ordinary TP may all-reduce hidden states. If the encoder and thinker partition different axes, rank colocation only gives us placement locality, not gather/scatter elimination.

## Can land in #615

These items are directly tied to the current Encoder TP implementation and should stay small enough to review with #615.

- [ ] Make #615's documentation and logs consistently distinguish encoder process peak from physical whole-GPU peak. This belongs in #615 because the PR already adds encoder timing/memory instrumentation and should not overclaim that lower encoder admission cost means colocated whole-GPU safety.

- [ ] Keep #615's memory-budget semantics explicit in schema validation, error messages, and PR docs. `total_gpu_memory_fraction` should be described as resident placement budget, while `encoder_activation_budget_bytes` should be described as encoder-stage admission budget, not as a full runtime memory cap.

- [ ] Keep the current colocated headroom guard in #615 as a conservative final safety check, not as a projected memory model. If the guard rejects a request, the error/log should make clear that modeled encoder admission passed but whole-GPU free headroom was insufficient.

- [ ] Finish localized duplicate-buffer cleanup that is already inside #615's implementation boundary. Examples include keeping metadata broadcast tensor-free, releasing non-entry follower cache/output references after forward, and shortening rank0 staging lifetimes where this can be covered by existing encoder scheduler tests.

- [ ] Document that #615 still uses full input tensor broadcast to encoder TP followers. This should be a clear follow-up limitation in #615, because replacing it with chunk/scatter/SP is a larger design change.

- [ ] Keep #615's existing topology as a transition path: thinker TP1 plus image/audio encoder TP2 in colocated-2gpu layout. This benchmark is useful for debugging #615, but should not be documented as the final recommended colocated TP policy.

- [ ] Include the minimal #615 benchmark evidence needed to interpret the PR: separate-layout image TP1 vs TP2 for long video, separate-layout audio TP1 vs TP2 for long audio, and current colocated transition-path behavior. Each benchmark should report encoder process peak and physical whole-GPU peak separately.

- [ ] Avoid adding new policy surfaces to #615 unless they are required for the current implementation. #615 should close the Path-B Encoder TP baseline, not also introduce model-wide placement policy, sharded handoff contracts, or SP/CP.

## After #615 merges - separate PRs

These items change runtime policy, placement semantics, or memory modeling enough that they should be reviewed independently after #615.

- [ ] Add a model-specific Qwen3-Omni parallelism / placement policy that expands server-level `--tp` intent into explicit stage configs. For example, `--tp 2` could expand to `thinker.tp=2`, `image_encoder.tp=2`, and `audio_encoder.tp=1`, with stage config remaining the runtime ground truth.

- [ ] Implement and benchmark the rank-colocated Qwen3-Omni candidate: thinker TP2 plus vision encoder TP2 plus audio TP1. This should validate `GPU i = thinker_rank_i + vision_encoder_rank_i` as a colocated capacity/headroom policy, without claiming sharded handoff.

- [ ] Keep `thinker_tp=1, vision_encoder_tp=2` as an expert override after #615, but do not promote it as the recommended colocated TP layout. It is useful when the thinker and encoder TP worlds are intentionally uncoordinated, but it does not answer the model-wide colocated TP policy question.

- [ ] Add an explicit sharded-handoff contract before claiming gather/scatter savings. A future `handoff-policy=sharded_aligned` mode should only be allowed when `EncoderOutputPartitionSpec` and `ThinkerInputPartitionSpec` are compatible; TP alone should not imply that compatibility.

- [ ] Replace the final colocated headroom guard with a projected whole-GPU memory model. The model should account for resident thinker/talker/code2wav weights, AR KV/activation, encoder in-flight work, broadcast/staging buffers, CUDA cache, and concurrent image/audio encoder activity.

- [ ] Add a startup-time placement-aware sizing heuristic for Qwen3-Omni colocated TP. The heuristic should turn model defaults, GPU memory, and media limits into initial resident and encoder activation budgets, but this should not be an online auto-tuning project.

- [ ] Decide the default audio encoder TP policy with follow-up benchmarks. The expected default is audio TP1, with audio TP2 enabled only for long audio, `audio_truncation=false`, tight audio activation budgets, or benchmark-backed cases.

- [ ] Integrate chunk-aware encoder paths as a separate memory optimization. This should evaluate whether upstream chunked-prefill/per-image encoder paths can reduce staging pressure without breaking parity, and it should not be treated as sharded handoff by itself.

- [ ] Treat Ming-Omni as a validation target for generic orchestration plus a model-specific local TP backend. Ming currently has image encoder TP hooks/launcher surface, but it should not be assumed to be covered by #615's SGLang-backed Qwen3-Omni encoder path or by any generic sharded-handoff contract.

- [ ] Prototype SP/CP versus TP for long video. This should test whether sequence/context partitioning is a better axis than TP for long-video activation memory and for any future sharded encoder-to-thinker handoff.

## Follow-up benchmark matrix

- [ ] Separate-layout image encoder TP1 vs TP2 on long video. This benchmark should confirm the #615 result that TP2 lowers image encoder process peak and can admit requests that TP1 rejects under the same modeled encoder budget.

- [ ] Separate-layout audio encoder TP1 vs TP2 on long audio with `audio_truncation=false`. This benchmark should decide when audio TP2 is useful enough to override the default audio TP1 policy.

- [ ] Current colocated transition path: thinker TP1 plus image/audio encoder TP2. This benchmark should measure whole-GPU peak and guard behavior for the #615-style topology, without treating it as the final recommended colocated TP layout.

- [ ] Rank-colocated Qwen3-Omni candidate: thinker TP2 plus vision encoder TP2 plus audio TP1. This benchmark should test colocated capacity/headroom for `GPU i = thinker_rank_i + vision_encoder_rank_i`, but it should not claim sharded handoff unless a partition contract exists.

- [ ] Qwen3-Omni vs Ming-Omni comparison once Ming image encoder TP support and validation are internally consistent. The goal is to see whether a thinker-heavy model makes rank-colocated placement more important, not to assume Ming is already covered by the Qwen3-Omni SGLang-backed path.

- [ ] SP/CP prototype versus TP for long video. This benchmark should test whether sequence/context partitioning is a better axis than TP for long-video activation memory and for any future sharded encoder-to-thinker handoff.

## Non-goals

- Do not claim TP2 is faster than TP1.
- Do not claim encoder admission cost is whole-GPU safety.
- Do not claim rank-colocated TP automatically removes gather/scatter.
- Do not implement online auto-tuning in this RFC.

timeuser4:
## What

Add a `--tp` CLI option that expands a high-level tensor parallel intent into model-specific per-stage TP sizes, as requested in #672.

For Qwen3-Omni:
- `--tp 2` → thinker=2, image_encoder=2, audio_encoder=1
- `--tp 4` → thinker=4, image_encoder=4, audio_encoder=2
- `--tp 8` → thinker=8, image_encoder=8, audio_encoder=4

Per-stage overrides (`--thinker-tp-size`) applied afterward take precedence.

## How

- `PipelineConfig.global_tp_stage_config(tp)` classmethod — models override to define their mapping; default returns `None` (unsupported).
- `apply_global_tp_expansion()` in `serve.py` — applies the mapping before per-stage overrides.
- Qwen3Omni text and speech configs both define the mapping. Audio encoder scales at `max(1, tp // 2)` since it doesn't benefit from full TP.

## Tests

7 new unit tests covering: text config, speech config, identity at tp=1, scaling at tp=4, no-op at None, unsupported config error, per-stage override precedence.

Closes #672 (the `--tp` expansion subtask).

### DIFF
--- a/sglang_omni/cli/serve.py
+++ b/sglang_omni/cli/serve.py
@@ -485,6 +485,34 @@ def apply_thinker_server_args_cli_overrides(
     return pipeline_config
 
 
+def apply_global_tp_expansion(
+    pipeline_config: PipelineConfig,
+    *,
+    global_tp: int | None,
+) -> PipelineConfig:
+    """Expand a global ``--tp N`` into model-specific per-stage TP sizes.
+
+    The mapping is defined by each model config's
+    :meth:`PipelineConfig.global_tp_stage_config` classmethod.  Per-stage
+    CLI overrides (e.g. ``--thinker-tp-size``) applied later take precedence.
+    """
+    if global_tp is None:
+        return pipeline_config
+
+    stage_tp_map = type(pipeline_config).global_tp_stage_config(global_tp)
+    if stage_tp_map is None:
+        raise typer.BadParameter(
+            f"--tp is not supported by {type(pipeline_config).__name__}"
+        )
+
+    for stage in pipeline_config.stages:
+        if stage.name in stage_tp_map:
+            tp = stage_tp_map[stage.name]
+            stage.tp_size = tp
+            stage.parallelism.tp = tp
+    return pipeline_config
+
+
 def apply_parallelism_cli_overrides(
     pipeline_config: PipelineConfig,
     *,
@@ -867,6 +895,17 @@ def serve(
         Literal["debug", "info", "warning", "error", "critical"],
         typer.Option(help="Log level (default: info)."),
     ] = "info",
+    global_tp: Annotated[
+        int | None,
+        typer.Option(
+            "--tp",
+            help=(
+                "Global tensor parallel size. Expands to model-specific per-stage "
+                "TP sizes (e.g. for Qwen3-Omni: thinker=tp, image_encoder=tp, "
+                "audio_encoder=max(1,tp//2)). Per-stage overrides take precedence."
+            ),
+        ),
+    ] = None,
     thinker_tp_size: Annotated[
         int | None,
         typer.Option(
@@ -1055,6 +1094,10 @@ def serve(
         merged_config,
         cpu_offload_gb=cpu_offload_gb,
         quantization=quantization,
+    )
+    merged_config = apply_global_tp_expansion(
+        merged_config,
+        global_tp=global_tp,
     )
     merged_config = apply_parallelism_cli_overrides(
         merged_config,
--- a/sglang_omni/cli/serve.py
+++ b/sglang_omni/cli/serve.py
@@ -1,6 +1,9 @@
 from __future__ import annotations
 
+import importlib
+import inspect
 import logging
+import warnings
 from typing import Annotated, Literal, NoReturn
 
 import typer
@@ -485,6 +488,17 @@ def apply_thinker_server_args_cli_overrides(
     return pipeline_config
 
 
+def _factory_supports_tp(factory_path: str) -> bool:
+    """Check if a stage factory accepts ``tp_rank`` in its signature."""
+    try:
+        module_path, func_name = factory_path.rsplit(".", 1)
+        module = importlib.import_module(module_path)
+        func = getattr(module, func_name)
+        return "tp_rank" in inspect.signature(func).parameters
+    except Exception:
+        return False
+
+
 def apply_global_tp_expansion(
     pipeline_config: PipelineConfig,
     *,
@@ -493,8 +507,11 @@ def apply_global_tp_expansion(
     """Expand a global ``--tp N`` into model-specific per-stage TP sizes.
 
     The mapping is defined by each model config's
-    :meth:`PipelineConfig.global_tp_stage_config` classmethod.  Per-stage
-    CLI overrides (e.g. ``--thinker-tp-size``) applied later take precedence.
+    :meth:`PipelineConfig.global_tp_stage_config` classmethod.  For each
+    stage in the mapping, the factory signature is probed for ``tp_rank``
+    support — stages whose factories don't accept TP kwargs fall back to
+    ``tp_size=1`` with a warning.  Per-stage CLI overrides applied later
+    take precedence.
     """
     if global_tp is None:
         return pipeline_config
@@ -509,10 +526,19 @@ def apply_global_tp_expansion(
         )
 
     for stage in pipeline_config.stages:
-        if stage.name in stage_tp_map:
-            tp = stage_tp_map[stage.name]
-            stage.tp_size = tp
-            stage.parallelism.tp = tp
+        if stage.name not in stage_tp_map:
+            continue
+        tp = stage_tp_map[stage.name]
+        if tp > 1 and not _factory_supports_tp(stage.factory):
+            warnings.warn(
+                f"Stage '{stage.name}' factory does not support TP kwargs; "
+                f"keeping tp_size=1. Encoder TP support is in progress "
+                f"(see https://github.com/sgl-project/sglang-omni/pull/615).",
+                stacklevel=2,
+            )
+            continue
+        stage.tp_size = tp
+        stage.parallelism.tp = tp
     return pipeline_config
 
 
--- a/sglang_omni/cli/serve.py
+++ b/sglang_omni/cli/serve.py
@@ -499,7 +499,10 @@ def apply_global_tp_expansion(
     if global_tp is None:
         return pipeline_config
 
-    stage_tp_map = type(pipeline_config).global_tp_stage_config(global_tp)
+    try:
+        stage_tp_map = type(pipeline_config).global_tp_stage_config(global_tp)
+    except ValueError as exc:
+        raise typer.BadParameter(str(exc)) from exc
     if stage_tp_map is None:
         raise typer.BadParameter(
             f"--tp is not supported by {type(pipeline_config).__name__}"
--- a/sglang_omni/models/qwen3_omni/config.py
+++ b/sglang_omni/models/qwen3_omni/config.py
@@ -263,6 +263,18 @@ class Qwen3OmniPipelineConfig(PipelineConfig):
         return {"thinker": "thinker"}
 
     @classmethod
+    def global_tp_stage_config(cls, tp: int) -> dict[str, int] | None:
+        if tp < 1:
+            raise ValueError(f"global tp must be >= 1, got {tp}")
+        # image_encoder mirrors thinker TP; audio_encoder scales at half
+        # the rate (TP >= 2 uses half of thinker TP, minimum 1).
+        return {
+            "thinker": tp,
+            "image_encoder": tp,
+            "audio_encoder": max(1, tp // 2),
+        }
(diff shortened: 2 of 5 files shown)


sglang_omni/cli/serve.py
```diff
@@ -485,6 +485,37 @@ def apply_thinker_server_args_cli_overrides(
...
+        raise typer.BadParameter(
+            f"--tp is not supported by {type(pipeline_config).__name__}"
+        )
+
+    for stage in pipeline_config.stages:
+        if stage.name in stage_tp_map:
+            tp = stage_tp_map[stage.name]
+            stage.tp_size = tp
+            stage.parallelism.tp = tp
+    return pipeline_config
```
JingwenGu0829, line 542: When I run serve --tp 2 against the default Qwen3-Omni configs it raises at placement-plan build time: ValueError: Stage 'image_encoder': TP placement requires a list of 2 unique GPU ids, got scalar gpu=0. 

The expansion sets each stage's tp_size, but the default stages carry a scalar gpu, and _resolve_stage_gpu_ids in config/placement.py needs a list of tp_size unique GPU ids once tp_size > 1. 

Could we have the expansion also assign the per-rank GPU ids (and re-validate through the same _validate_stage_parallelism_config path the --thinker-gpus flow uses)? There's also no --image-encoder-gpus/talker GPU-list flag, so this can't be satisfied by hand today either — even --tp 2 --thinker-gpus 0,1 still fails on image_encoder.  Maybe we could at least fail fast here with a message that mentions --tp, so users aren't sent debugging into the placement layer.

tests/unit_test/qwen3_omni/test_cli.py
```diff
@@ -405,3 +407,102 @@ def test_partial_start_cli_rejects_unsupported_config_with_stable_message():
         match="--talker-partial-start is not supported by Qwen3OmniPipelineConfig",
     ):
         apply_partial_start_cli_overrides(config, talker_partial_start="on")
+
+
+# -- global TP expansion tests --
+
+
```
JingwenGu0829, line 414: The tests all stop at checking tp_size and none builds the placement plan or the launch specs that actually consume it — which is why both issues I mentioned above stay green.

sglang_omni/models/qwen3_omni/config.py
```diff
@@ -250,6 +250,21 @@ def _speech_stages(
...
+    image_encoder mirrors thinker TP; audio_encoder scales at half the rate
+    (TP >= 2 uses half of thinker TP, minimum 1).
+    """
+    if tp < 1:
+        raise ValueError(f"global tp must be >= 1, got {tp}")
+    return {
+        "thinker": tp,
+        "image_encoder": tp,
+        "audio_encoder": max(1, tp // 2),
+    }
```
JingwenGu0829, line 265: The mapping shards the image encoder at full TP (and the audio encoder at tp>=4), but create_image_encoder_executor / create_audio_encoder_executor in stages.py only accept model_path/device/dtype, while any tp_size>1 stage gets tp_rank/tp_size/nccl_port injected into its factory args and called as factory(**factory_args). 

So once the GPU-list issue is fixed, the encoder stages would then crash at construction with an unexpected-keyword TypeError. The thinker and talker_ar factories already accept those kwargs, so would it make sense to map only thinker (and talker_ar for speech) for now and keep the encoders at tp_size=1, unless we want to add real TP support to the encoder factories?

JingwenGu0829 (commented):
Maybe I'm wrong, but there are serveral severe issues that is blocking your modified pipeline to actually initialize. They are pretty much partially caused by a non-rigor unit test.

Additionally, per your next commit, can you provide an end-to-end test including all the crucial performance metrics with the global --tp flag on to self-check that the whole pipeline is working as-is?

sglang_omni/cli/serve.py
timeuser4: OK. Since encoder TP support is being added in #615, I'll keep encoders at tp_size=1 for now. I'll probe whether encoder TP can be applied by checking if the factory signature includes tp_rank. If the encoder factories don't support TP kwargs yet, the expansion will fall back to tp_size=1 for encoders and emit a warning pointing to #615.

sglang_omni/models/qwen3_omni/config.py
timeuser4: OK. I'll take your advice. #615 is adding encoder TP support. In the meantime, when `--tp N` is used and encoder TP isn't supported yet, thinker (and talker_ar for speech) will keep TP=N while encoders automatically fall back to tp_size=1 with a warning to the user.

tests/unit_test/qwen3_omni/test_cli.py
timeuser4: OK. I'll add tests that:
1. Call build_stage_placement_plan() on the expanded config to verify the full path doesn't crash
2. Verify the encoder TP fallback. When encoder TP isn't supported, encoders stay at tp_size=1 and a warning is emitted

timeuser4:
> Additionally, per your next commit, can you provide an end-to-end test including all the crucial performance metrics with the global --tp flag on to self-check that the whole pipeline is working as-is?

OK. Once #615 merges, I'll run e2e tests with key performance metrics. My hardware is 8x4090 (20GB each), which may not fit the full Qwen3-Omini BF16 model. I'll try with higher TP (4 or 8) first. If that doesn't work, would a quantized Qwen3-Omini  (AWQ 4-bit) comparison?

PR #677, opened 2026-06-03
[CI] Upgrade the pinned SGLang backend (0.5.8 → latest)

### LINKED ISSUE #658: Upgrade the pinned SGLang backend (0.5.8 → latest)
## Motivation

https://github.com/sgl-project/sglang-omni/issues/441

We pin `sglang==0.5.8` (Jan 2026); upstream is now 0.5.12. Two reasons to upgrade.

 Omni currently runs only on Hopper and Ampere. B-cards are becoming the common deployment target in the US and we don't support them yet. Newer SGLang ships the SM100 path we'd need: FA4 with Blackwell support, multimodal attention defaulting to FA4 on SM100, and TRT-LLM DSA kernels default on SM100.

Upstream has reworked the multimodal/encoder path (stronger transformers/VLM modeling backend). Much of our encoder-side glue is hand-rolled against 0.5.8 internals; the newer APIs should let us delete some of it and make future encoder work easier.

## Pinned / gated to 0.5.8 today (must move together)

- `pyproject.toml` L29 — `sglang==0.5.8`
- `model_runner/_sglang_qwen3_vl_patches.py` L29 — a hard allowlist `{"0.5.8","0.5.8.post1"}` that rejects any other version at runtime; it guards Qwen3-VL vision-layout patches that must be re-validated against the new layout.

## Version-tied debt to clear during the upgrade

- `models/qwen3_omni/components/talker.py` L59 — FIXME: dense code-predictor MLP quantization changed in newer SGLang.
- `models/qwen3_omni/config.py` L19 — FIXME: DeepGEMM compile workaround.
- `model_runner/model_worker.py` L378 — mirrors 0.5.8 FP8 CUTLASS MoE assertions; must be re-synced against upstream.
- `models/qwen3_asr/request_builders.py` L190/L210 — stashes fields in `MultimodalDataItem` assuming 0.5.8 has no `model_specific_data`; check whether newer SGLang formalizes it.
- `models/qwen3_omni/hf_config.py` L23 — relies on 0.5.8 MRotaryEmbedding selection.

## Coupled constraint

SGLang ≥0.5.10 moves to transformers 5.x, but we pin `transformers<5.0` (pyproject). The upgrade forces a transformers bump too; models need re-validation against it.

## Scope

- [ ] Bump the pin + allowlist; existing CI green on Hopper/Ampere (no regression).
- [ ] Bring up and add CI coverage on a Blackwell Machine (really?).
- [ ] Replace hand-rolled encoder glue with the new upstream encoder APIs where available.
- [ ] Resolve the FIXMEs above; drop workarounds the upgrade makes unnecessary.

Billy1900:
## Motivation

Upgrade the pinned SGLang backend (0.5.8 → latest)

## Modifications

- pyproject.toml: pin sglang==0.5.12; relax transformers<5.0 → >=5.0 (SGLang >=0.5.10 requires transformers 5.x)
_sglang_qwen3_vl_patches.py: expand _SUPPORTED_SGLANG_VERSIONS to include 0.5.9 through 0.5.12.post1
- talker.py: remove _quant_config_for_code_predictor_dense_mlp and its mlp.gate substring-matching workaround; SGLang >=0.5.9 uses dotted-boundary matching so "mlp.gate" no longer bleeds into "mlp.gate_up_proj"
- hf_config.py, model_worker.py: drop stale "0.5.8" from comments
- tests: remove the two unit tests that validated the removed workaround; update version references in docstrings/requirements

## Related Issues
Closes #658 

## Checklist
- test_fp8_backend_config.py: all test passed
- test_qwen3_asr/: all test passed
- All other importable unit tests: all passed except 2 (for test_qwen_thinker_cuda_graph_capture_lifecycle, sgl_kernel needs libnvrtc.so.13 (CUDA 13.0) — driver mismatch)
- No Blackwell hardware, No A100, not tested

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,8 +16,8 @@ dependencies = [
     "numpy>=1.24.0",          # Array operations
     "pydantic>=2.0.0",        # Data validation and serialization
     "PyYAML>=6.0",            # YAML config parsing
-    "torch==2.9.1",          # PyTorch for tensor operations
-    "torchvision==0.24.1",    # torchvision for vision processor
+    "torch==2.11.0",          # PyTorch for tensor operations; matches sglang==0.5.12
+    "torchvision==0.26.0",    # torchvision for vision processor
     "accelerate>=0.27.0",     # init_empty_weights for meta initialization
     "transformers>=5.0",  # HF models/tokenizers; SGLang >=0.5.10 requires transformers 5.x
     "safetensors>=0.4.3",     # Direct weight loading
@@ -63,7 +63,7 @@ dependencies = [
     "omegaconf",
     "descript-audiotools",
     "descript-audio-codec",
-    "torchaudio",
+    "torchaudio==2.11.0",
     # Gradio playground
     "gradio>=4.0.0",
     # Ming-Omni
@@ -79,6 +79,9 @@ sgl-omni-router = "sglang_omni_router.serve:main"
 [tool.uv]
 override-dependencies = [
     "protobuf>=6.31.1,<7.0.0",
+    # kernels>=0.15 breaks transformers 5.x hub_kernels (LayerRepository missing version).
+    # Pin until transformers ships a fix (tracked upstream: huggingface/transformers#46039).
+    "kernels<0.15",
 ]
 
 [tool.pytest.ini_options]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,14 +19,14 @@ dependencies = [
     "torch==2.9.1",          # PyTorch for tensor operations
     "torchvision==0.24.1",    # torchvision for vision processor
     "accelerate>=0.27.0",     # init_empty_weights for meta initialization
-    "transformers<5.0",   # HF models/tokenizers
+    "transformers>=5.0",  # HF models/tokenizers; SGLang >=0.5.10 requires transformers 5.x
     "safetensors>=0.4.3",     # Direct weight loading
     "pillow>=10.0.0",         # Required by AutoImageProcessor
     "huggingface-hub>=0.36.0",# HF model downloads
     "datasets>=2.14.0",       # Arrow-based SeedTTS dataset loading
     "fastapi>=0.110.0",       # OpenAI-compatible API adapter
     "uvicorn>=0.23.0",        # ASGI server for FastAPI
-    "sglang==0.5.8",
+    "sglang==0.5.12",
     "nixl",
     "mooncake-transfer-engine>=0.3.0",
     "logger",
--- a/tests/unit_test/qwen3_omni/test_fp8_backend_config.py
+++ b/tests/unit_test/qwen3_omni/test_fp8_backend_config.py
@@ -391,14 +391,14 @@ def test_model_config_has_moe_prefers_effective_text_config() -> None:
         pytest.param(False, True, False, False, id="cutlass_runtime_rejected"),
     ],
 )
-def test_fp8_cutlass_moe_support_matches_sglang_0_5_8_contract(
+def test_fp8_cutlass_moe_support_check(
     monkeypatch: pytest.MonkeyPatch,
     cutlass_supported: bool,
     sm90_supported: bool,
     sm100_supported: bool,
     expected_supported: bool,
 ) -> None:
-    """Mirrors the CUTLASS FP8 MoE assertions in pinned SGLang 0.5.8."""
+    """Validates the CUTLASS FP8 MoE support check."""
     _install_fake_cutlass_support_modules(
         monkeypatch,
         cutlass_supported=cutlass_supported,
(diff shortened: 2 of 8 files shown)


pyproject.toml
```diff
@@ -16,17 +16,17 @@ dependencies = [
...
-    "transformers<5.0",   # HF models/tokenizers
+    "transformers>=5.0",  # HF models/tokenizers; SGLang >=0.5.10 requires transformers 5.x
     "safetensors>=0.4.3",     # Direct weight loading
     "pillow>=10.0.0",         # Required by AutoImageProcessor
     "huggingface-hub>=0.36.0",# HF model downloads
     "datasets>=2.14.0",       # Arrow-based SeedTTS dataset loading
     "fastapi>=0.110.0",       # OpenAI-compatible API adapter
     "uvicorn>=0.23.0",        # ASGI server for FastAPI
-    "sglang==0.5.8",
+    "sglang==0.5.12",
```
JingwenGu0829, line 29: Since the 0.5.8-specific dense-MLP quant workaround is removed, `_SUPPORTED_SGLANG_VERSIONS` should probably stop accepting `0.5.8` and `0.5.8.post1`. Otherwise the repo appears to support an SGLang version that may now regress on this path.

tests/unit_test/qwen3_omni/test_fp8_backend_config.py
```diff
@@ -391,14 +391,14 @@ def test_model_config_has_moe_prefers_effective_text_config() -> None:
         pytest.param(False, True, False, False, id="cutlass_runtime_rejected"),
     ],
 )
-def test_fp8_cutlass_moe_support_matches_sglang_0_5_8_contract(
+def test_fp8_cutlass_moe_support_check(
```
JingwenGu0829, line 394: Other than the unit test, empirically upgrading pinned sglang version will introduce a series of unexpecting runtime issues. 

As a result, if you have access to developer machines,could you include the end-to-end GPU test, specifically Qwen3-Omni text audio/video, talker audio generation/WER, and the existing benchmark-style checks for MMMU/ MMSU/Video-MME/Video-AMME. FP8 should also be included if support is expected to remain unchanged.

JingwenGu0829 (commented):
Upgrading sglang version ping is an important PR. As a result, left some comments helping that it could help you make the pr more rigorous!

pyproject.toml
Billy1900: I am not sure about this. let the owner make such decision.

tests/unit_test/qwen3_omni/test_fp8_backend_config.py
Billy1900: sure. so I left some comments there, waiting for someone who have the developer machines could help me test it.

pyproject.toml
JingwenGu0829: I confused. You deleted the regression path for the old version so you should not allow the older sglang version as compatiable in pyproject.toml.

I'm also confused by 'let the owner make such decision', is there any specific developer that you mean by code owner in sglang-omni that you hope to get feedback from?

Lordworms:
I have an NVIDIA GB10 machine and tested this PR end-to-end with
Qwen3-ASR-1.7B on CUDA 13.0.

The server started successfully with:

- NVIDIA GB10 (compute capability 12.1)
- PyTorch 2.11.0+cu130
- SGLang 0.5.12
- sglang-kernel 0.4.2.post2+cu130

However, a real transcription request failed because Qwen3-ASR records
multimodal audio offsets as half-open ranges, while SGLang 0.5.12
interprets the end offset as inclusive.

For an audio encoder output containing 60 tokens, Omni recorded:

    (audio_start, audio_start + 60)

SGLang therefore calculated 61 tokens and failed with:

    split_with_sizes expects split_sizes to sum exactly to 60,
    but got split_sizes=[61]

Changing the end offset to:

    audio_start + num_audio_tokens - 1

fixes the mismatch. I also added a regression test and verified the full
HTTP transcription path on GB10. The request now succeeds and returns:

    {"text":"How many cars are there in the picture?"}

I can open a PR against this branch with the fix if that would be helpful.

pyproject.toml
Billy1900: you can fix that if you want, thanks.

Billy1900:
> I have an NVIDIA GB10 machine and tested this PR end-to-end with Qwen3-ASR-1.7B on CUDA 13.0.
> 
> The server started successfully with:
> 
> * NVIDIA GB10 (compute capability 12.1)
> * PyTorch 2.11.0+cu130
> * SGLang 0.5.12
> * sglang-kernel 0.4.2.post2+cu130
> 
> However, a real transcription request failed because Qwen3-ASR records multimodal audio offsets as half-open ranges, while SGLang 0.5.12 interprets the end offset as inclusive.
> 
> For an audio encoder output containing 60 tokens, Omni recorded:
> 
> ```
> (audio_start, audio_start + 60)
> ```
> 
> SGLang therefore calculated 61 tokens and failed with:
> 
> ```
> split_with_sizes expects split_sizes to sum exactly to 60,
> but got split_sizes=[61]
> ```
> 
> Changing the end offset to:
> 
> ```
> audio_start + num_audio_tokens - 1
> ```
> 
> fixes the mismatch. I also added a regression test and verified the full HTTP transcription path on GB10. The request now succeeds and returns:
> 
> ```
> {"text":"How many cars are there in the picture?"}
> ```
> 
> I can open a PR against this branch with the fix if that would be helpful.

sure, would like to see that.

PR #680, opened 2026-06-04
Tolerate missing per-model dependencies in _register_omni_model

draftbk:
## Motivation

`_register_omni_model` imports every model in one block, so one missing optional dependency aborts startup. Now each registers independently and failures are logged and skipped (like [`sglang.srt.models.registry`](https://github.com/sgl-project/sglang/blob/14ed9b448edb544b9b1246cd32191227c44edea7/python/sglang/srt/models/registry.py#L104-L109)), unchanged when all deps are present.

## Test

With every dependency installed, all ten architectures still register (no behavior change). With only one model's dependencies present (e.g. MOSS-TTS-v1.5, without `qwen_vl_utils`), the unavailable models are logged and skipped and the engine boots and serves, instead of crashing with `ModuleNotFoundError` at startup.


draftbk:
Hi @zhaochenyang20, CI is green now. Could you take a look when you have a chance? Happy to address any comments.

zhaochenyang20 (approved):

PR #695, opened 2026-06-05
[Feat] Support MOSS-TTS async decode opt-in

guozhihao-224:
## Summary

Implements opt-in async decode support for MOSS-TTS, following the Higgs launch/resolve split while keeping MOSS async disabled by default.

- Move MOSS row, feedback embedding, delay-state, and sampling-step commits into the collect path so one-step lookahead cannot overwrite runner-global pending state.
- Add MOSS async `post_decode_launch` / `post_decode_resolve` hooks, launch-side stop/length guards, and pending finish bridging back to SGLang's `finished_reason`.
- Wire `enable_async_decode` and `async_decode_min_batch_size` through the MOSS TTS engine factory and the shared `--async-decode` CLI override.
- Add unit coverage for async hook behavior, stop/length overrun protection, launch-side sampling-step semantics, CLI plumbing, and MOSS pipeline wiring.

## Why async stays opt-in for MOSS

MOSS and Higgs both have strict step-to-step decode dependencies. The important difference is the existing state ownership model. Higgs keeps decode feedback state in a per-request / row-indexed sampler pool, so async launch can update stable per-request rows without relying on a runner-global pending slot.

MOSS currently keeps next-step feedback in request-local Python state (`output_rows`, `pending_feedback_queue`, delay state, sampling counters), while the old synchronous runner staged rows/embeddings through runner-global `_pending_rows` / `_pending_embeds`. Those runner-global pending slots are unsafe under one-step lookahead because step N+1 can launch before step N is resolved.

This PR therefore uses a correctness-first design based on the current MOSS state model: launch commits the sampled row, feedback embedding, delay state, and pending stop/length state before the next decode step is scheduled. That makes async decode correct and opt-in usable, but it also leaves very little work in resolve to overlap. The A800 validation below confirms this: async event waits are hits, while most cost stays in launch.

To get a larger async speedup, MOSS likely needs a deeper refactor toward the Higgs-style model: a per-request / row-indexed MOSS feedback buffer or pool where launch writes next-step feedback into stable rows, `before_decode` reads those rows directly, and request-state/output commit can be decoupled from the heavy launch path.

The A800 validation below did not show a stable default-on win. After the stop fix, async-on token counts were comparable to async-off and did not run away, but latency/throughput were roughly flat to slightly slower depending on the run. This PR therefore only adds opt-in support via config/CLI.

## Correctness validation

Stop/length handling:

- MOSS launch now publishes `result.next_token_ids` immediately so SGLang remains the source of truth for final stop/length finish reasons.
- MOSS also marks launch-side `moss_stop_pending` / pending finish state so a one-step lookahead overrun cannot append extra MOSS rows or feedback before SGLang resolves the request.
- The launch-side `moss_sampling_steps` counter feeds `_sample_rows`, avoiding the async one-step lag from base `generation_steps` finalize.

Audio smoke:

- A800 GPU c4 smoke run generated 4 audio files.
- Manual listening check: audio quality was normal and generated content matched the corresponding text/file.

## Benchmark / profiling results

Environment:

- A800 GPU 80 GB
- Model: `OpenMOSS-Team/MOSS-TTS-v1.5`
- Existing remote HF/model cache, no model download during runs
- `--talker-cuda-graph off`
- `benchmarks.eval.benchmark_tts_seedtts`
- Generate-only, English references, `--token-count auto`

Normal c4 benchmark after the stop fix:

| Mode | Mean latency | QPS | Output throughput | Notes |
|---|---:|---:|---:|---|
| async off | 4.885 s | 0.804 | 69.5 tok/s | baseline |
| async on | 4.993 s | 0.788 | 68.1 tok/s | slightly slower |

Profiled c4/32 A-B run:

| Mode | Completed | Failed | Mean latency | QPS | Output tokens |
| async off | 32 | 0 | 4.957 s | 0.770 | 2706 |
| async on | 32 | 0 | 4.947 s | 0.775 | 2706 |

Representative async profile from the same run:

- `query_hit=820`, `query_miss=0`: lookahead did not block on the CUDA event.
- `resolve.post_decode_resolve ~= 0.003 ms`
- `resolve.finalize ~= 0.049 ms`
- `scheduler.process_batch_result ~= 0.078 ms`
- `launch.prepare_and_forward ~= 17.6 ms`
- `launch.post_decode_launch ~= 7.9 ms`
- MOSS collect breakdown:
  - `sample_rows ~= 4.4 ms`
  - `channel_logits ~= 2.2 ms`
  - `feedback_embeds ~= 1.1 ms`

Interpretation: async overlap is functioning, but the resolve-side work is already tiny. Most MOSS decode cost remains launch-side in this PR because the current MOSS request-state/feedback-queue implementation requires next-step feedback to be prepared before the next launch.

## Follow-up optimization observations

These were investigated separately and are not included in this PR:

- Selective audio-head logits reduced `channel_logits` from about 2.26 ms to about 1.58-1.63 ms, but c4/32 end-to-end improvement was small.
- Top-k subset sampling reduced `sample_rows` from about 4.42 ms to about 4.20 ms, but mean latency only moved from about 4.85 s to about 4.84 s.
- Any top-k subset implementation must preserve `multinomial_with_seed` semantics, because its deterministic Gumbel noise is derived from the original token id / column index.

Likely future launch-side targets:

- reduce GPU-host scalar syncs in `_sample_rows`;
- reduce full-vocab top-p/top-k work while preserving fixed-seed semantics;
- optimize feedback embedding construction when most audio channels are pad;
- refactor MOSS feedback ownership toward a Higgs-style per-request / row-indexed feedback buffer so async can decouple next-step feedback staging from request-state/output commit.

## reference
https://github.com/sgl-project/sglang-omni/issues/637

### DIFF
--- a/sglang_omni/cli/serve.py
+++ b/sglang_omni/cli/serve.py
@@ -14,9 +14,10 @@ logger = logging.getLogger(__name__)
 
 _STAGE_TOGGLE_MODE = Literal["default", "on", "off"]
 _QWEN_COLOCATED_CONFIG_CLASS = "Qwen3OmniSpeechColocatedPipelineConfig"
-_HIGGS_ASYNC_DECODE_FACTORY = (
-    "sglang_omni.models.higgs_tts.stages.create_sglang_tts_engine_executor"
-)
+_ASYNC_DECODE_FACTORIES = {
+    "sglang_omni.models.higgs_tts.stages.create_sglang_tts_engine_executor",
+    "sglang_omni.models.moss_tts.stages.create_sglang_tts_engine_executor",
+}
 _QWEN_PARTIAL_START_TALKER_FACTORY = (
     "sglang_omni.models.qwen3_omni.stages.create_talker_ar_executor_from_config"
 )
@@ -670,7 +671,9 @@ def _apply_stage_factory_args_override(
     updates: dict[str, object],
     reason: str,
     supported_factory: str | None = None,
+    supported_factories: set[str] | None = None,
     flag_name: str | None = None,
+    supported_display: str | None = None,
 ) -> None:
     matching_stages = _find_matching_stages(
         pipeline_config,
@@ -678,10 +681,14 @@ def _apply_stage_factory_args_override(
         reason=reason,
     )
     for stage in matching_stages:
-        if supported_factory is not None and stage.factory != supported_factory:
+        allowed = supported_factories
+        if allowed is None and supported_factory is not None:
+            allowed = {supported_factory}
+        if allowed is not None and stage.factory not in allowed:
             display_flag = flag_name or reason
+            supported = supported_display or "the supported stage factories"
             raise typer.BadParameter(
-                f"{display_flag} currently supports only Higgs TTS; "
+                f"{display_flag} currently supports only {supported}; "
                 f"stage {stage.name!r} uses factory {stage.factory!r}"
             )
         factory_args = dict(stage.factory_args or {})
@@ -728,8 +735,9 @@ def apply_async_decode_cli_overrides(
         stage_name="tts_engine",
         updates=updates,
         reason="async decode override",
-        supported_factory=_HIGGS_ASYNC_DECODE_FACTORY,
+        supported_factories=_ASYNC_DECODE_FACTORIES,
         flag_name="--async-decode/--async-decode-min-batch-size",
+        supported_display="Higgs TTS and MOSS-TTS",
     )
     return pipeline_config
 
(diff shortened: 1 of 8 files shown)


guozhihao-224:
I believe we need to refactor MOSS feedback ownership, similar to Higgs, into a Higgs-style feedback buffer indexed by request/line, Launch would write only next-step feedback state into stable pool rows, before_decode would read those rows directly, and request/output commits could move to resolve or journal apply, so that asynchronous operations can decouple the pausing of next feedback from request status/output submission. This reduces the workload of launch and increases the async overlap space.

Do you have any other ideas? cc @zhaochenyang20 @gaoyang07

PR #698, opened 2026-06-05
Upgrade pinned SGLang backend

### LINKED ISSUE #658: Upgrade the pinned SGLang backend (0.5.8 → latest)

Hayden727:
## Motivation

Closes #658.

## Notes

Opening this as a draft placeholder for the SGLang backend upgrade. Follow-up commits will add the actual dependency/runtime updates after guidance.

## Validation

Not run yet; no code changes are included in this placeholder PR.


Hayden727:
### 🏎️ Perf A/B — Qwen3-ASR (H100)

`main` vs `sglang==0.5.12.post1`, SeedTTS-ASR corpus (20 EN), concurrency 32, identical workload/host.

| metric | main (sglang 0.5.x / torch 2.9 / tf 4.57) | post1 (0.5.12.post1 / torch 2.11 / tf 5.6) | Δ |
|---|---|---|---|
| WER (corpus) | 0.69% | 0.69% | = (no quality regression) |
| throughput | 50.2 samp/s | **60.0 samp/s** | **+19.5%** |
| latency mean | 0.344 s | **0.275 s** | −20% |
| latency p95 | 0.385 s | **0.314 s** | −18% |
| RTF mean | 0.069 | **0.055** | −20% |

**No regression — post1 ~20% faster, WER identical.**

Hayden727:
### 🏎️ Perf A/B — Qwen3-Omni MMMU (H100)

`main` vs `sglang==0.5.12.post1`, MMMU-CI-50 (image→text), Qwen3-Omni-30B-A3B-Instruct (bf16), identical workload/host.

| metric | main | post1 | Δ |
|---|---|---|---|
| accuracy | 60.0% (30/50) | 60.0% (30/50) | = (no quality regression) |
| output throughput | 812 tok/s | **940 tok/s** | **+15.7%** |
| request throughput | 1.461 req/s | 1.454 req/s | ≈ |

**No regression — accuracy identical, decode throughput +16%.** (Includes the transformers-5.6 vision-encoder fix.)

Hayden727:
### 🏎️ Perf A/B — Qwen3-Omni MMSU (H100)

`main` vs `sglang==0.5.12.post1`, MMSU-CI-2000 (audio→text), Qwen3-Omni-30B-A3B-Instruct (bf16), identical workload/host.

| metric | main | post1 | Δ |
|---|---|---|---|
| accuracy | 70.75% (1415/2000) | 70.00% (1400/2000) | −0.75pp (within run-to-run noise) |
| output throughput | 125.0 tok/s | **128.1 tok/s** | +2.5% |

**No regression — accuracy within noise, throughput slightly higher.**

Hayden727:
### 🏎️ Perf A/B — Higgs TTS (H100)

`main` vs `sglang==0.5.12.post1`, SeedTTS voice-clone, non-streaming, concurrency 16, identical workload/host.

| metric | main | post1 | Δ |
|---|---|---|---|
| RTF mean | 0.212 | **0.204** | −3.7% |
| throughput | 17.67 req/s | **18.09 req/s** | +2.4% |
| audio duration mean | 4.36 s | 4.43 s | ≈ |

**No regression — post1 slightly faster (RTF lower, throughput higher).**

Hayden727:
## LLaDA2-Uni — post1 latency regression: root cause located

Tested on **H100**, LLaDA2.0-Uni, post1 stack vs main stack, **both with `disable_cuda_graph=True`** (graph usage kept identical to `main` for a fair apples-to-apples comparison).

### Symptom
Single-request latency regresses ~70% on post1 (identical greedy output, "Paris", 2 tokens):
- fixed 2-token latency: **main 0.058s → post1 0.100s**
- concurrent chat n=24 c=8: **main 1.25 → post1 1.07 req/s**

### What it is NOT (all measured and ruled out)
| hypothesis | result |
|---|---|
| GPU kernels slower | ❌ identical — `fused_moe_kernel` 21.6 vs 21.7ms; Self CUDA total 38.2 vs 38.0ms |
| torch 2.11 per-op CPU dispatch slower | ❌ microbench identical — launch 4.74/4.69µs, `.item()` sync 8.82/8.90µs, argmax 12.44/12.36µs |
| more denoising iterations | ❌ identical (5 = 5) |
| more model forwards per block | ❌ identical (per-step + one final) |
| more `.item()` syncs | ❌ reducing post1's count *below* main's recovered only 2.6ms — still slower |
| scheduler / batch-build overhead | ❌ `sched`/`build`/`post` phases all <1ms, identical |

### Root cause: lost CPU–GPU pipelining in upstream sglang's rewritten dllm forward
Forcing a per-phase `cuda.synchronize()` collapses **both** stacks to ~41ms. Without it, `main` pipelines the denoising loop down to ~24ms (CPU dispatch overlaps GPU compute), while post1 runs **fully serialized at ~41ms with no overlap** — its uninstrumented time already equals its force-serialized time.

GPU-busy time and per-op CPU costs are identical on both. The only difference is that post1's `sglang.srt.dllm.algorithm.low_confidence` **multi-batch rewrite serializes execution** and loses the CPU–GPU overlap the older single-batch implementation achieved. This is an execution-*pipelining* regression in upstream sglang, **not** a per-op CPU-cost or GPU-kernel regression.

### Mitigation (not applied here, to stay consistent with `main`)
post1's `CudaGraphRunner` supports the dllm path (`ForwardMode.DLLM_EXTEND`, `num_tokens_per_bs=block_size`). Enabling cuda graph collapses each forward's launches into one graph replay and recovers the regression: eager 0.100s / 1.07 req/s → **+graph 0.038s / 4.05 req/s (2.6× / 3.8×)**. We are intentionally **keeping graph off to match `main`** rather than masking the upstream issue with a config divergence.

### Follow-up
The upstream pipelining regression can be patched at the sglang-omni layer via a **monkeypatch of `low_confidence.py`** — same mechanism as the existing Qwen3-VL HF-parity patches, no sglang fork needed. Two viable directions:
1. restore the single-batch pipelined path for `batch_size == 1` (the serving-common case), or
2. enable the dllm cuda-graph specifically for this stage.

Pinning the exact serialization point (which kernel gap the GPU idles on) requires an nsys timeline; the torch profiler's own overhead exceeds the ~17ms signal.

Hayden727:
## MOSS-TTS — perf A/B (main vs post1)

Tested on **H100**, voice-clone via `/v1/audio/speech` (SeedTTS reference), concurrent n=16 c=4.

| metric | main | post1 |
|---|---|---|
| throughput | 1.727 req/s | **1.775 req/s** (+2.8%) |
| latency mean | 2.293s | **2.229s** |
| latency p95 | 2.417s | **2.313s** |
| RTF mean | 0.912 | **0.883** |
| errors | 0 | 0 |

**No regression** — post1 is marginally faster (RTF 0.883 vs 0.912). Audio output valid on both.

Hayden727:
## Fish-S2-Pro — perf A/B (main vs post1)

Tested on **H100**, voice-clone via `/v1/audio/speech` (SeedTTS reference), concurrent n=16 c=4.

| metric | main | post1 |
|---|---|---|
| throughput | 0.780 req/s | **0.826 req/s** (+5.9%) |
| latency mean | 4.956s | **4.716s** |
| latency p95 | 5.714s | **5.409s** |
| RTF mean | 1.207 | **1.165** |
| errors | 0 | 0 |

**No regression** — post1 is faster (RTF 1.165 vs 1.207). Audio output valid on both. (Note: the post1 `_rematerialize_audio_decoder_buffers` bootstrap fix is required for Fish to run on the post1 stack; main runs the pre-bump path.)

Hayden727:
## Voxtral TTS — perf (post1 vs main) — regression FIXED

Tested on **H100**, preset-voice (`cheerful_female`) via `/v1/audio/speech`, concurrent (n=64, c=8). Both servers capture cuda graph (`bs [1, 2, 4, 8, 12, 16]`) — like-for-like.

### Root cause of the initial ~8% RTF overhead
Both sides graph-on, so the gap was pure GPU-kernel time. Torch-profiler kernel attribution (10 reqs): post1 total GPU **1312ms vs main 1170ms (+12%)**, and **GEMM was 83% of it** (post1 945ms vs main 827ms). Attention was at parity. Cause: on torch 2.11 / CUDA 13, inductor's **default GEMM lowering** routes the compiled acoustic-transformer matmuls onto split-K cuBLAS (`nvjet_sm90`) kernels ~14% slower than the triton templates torch 2.9 selected.

### Fix (commit `9aab606b`)
Enable inductor **per-shape GEMM autotuning** (`max_autotune_gemm`, backends `TRITON,ATEN`) in the Voxtral generation stage when `torch_compile` is on — inductor benchmarks triton vs aten per shape and keeps the fastest. Scoped to the Voxtral generation process; only cost is a little one-time startup autotune.

### After the fix (in-code, no env vars)

| metric | post1 (fixed) | main |
|---|---|---|
| RTF mean | **~0.210** | ~0.208 |
| throughput | ~11.0 req/s | ~11.0 req/s |
| errors | 0 | 0 |

post1 went from RTF 0.227 (−8%) to **~0.210 — at parity with main**. Regression eliminated. (Unit test added; validated total GPU time 1312 → 1071ms.)

Hayden727:
## Whisper large-v3 ASR — post1 vs main

Tested on **H100**, `/v1/audio/transcriptions`, concurrent (n=48, c=8).

| metric | post1 | main |
|---|---|---|
| throughput | ~23.5 req/s | — (cannot serve) |
| latency mean | ~0.34s | — |
| errors | 0 / 48 | 48 / 48 (HTTP 500) |
| transcript | ✅ correct | ❌ empty (request errors) |

**post1 restores a path that is broken on main.** On the main stack every transcription request returns HTTP 500 with `RuntimeError: 'NoneType' object has no attribute 'view'` in the Whisper decode path — main cannot serve Whisper at all. post1 serves it cleanly:

> "We asked over twenty different people, and they all said it was his."

at ~23.5 req/s with zero errors. Since main produces no successful request, there is no throughput A/B to draw — this is a capability the bump fixes, not a regression.

Hayden727:
## Qwen3-TTS 0.6B — post1 vs main

Tested on **H100**, voice-clone via `/v1/audio/speech` (ref_audio + ref_text), concurrent (n=32, c=8).

| metric | post1 | main |
|---|---|---|
| throughput | ~4.4 req/s | — (cannot start) |
| latency mean | ~1.7s | — |
| RTF mean | ~0.69 | — |
| errors | 0 | startup crash |

**main cannot start Qwen3-TTS.** The pipeline dies during startup with:

```
TypeError: check_model_inputs() missing 1 required positional argument: 'func'
RuntimeError: Process pipeline died during startup (exit code 1)
```

The Qwen3-TTS model code requires the newer `transformers` `check_model_inputs` decorator signature that the bump (transformers 5.6.0) provides; main's older `transformers` ships the old signature. post1 serves it cleanly (HTTP 200, valid WAV, RTF ~0.69 — real-time) with zero errors. Bump-enabled capability — no regression to compare.

Hayden727:
## Qwen3-Omni speech-output (FP8 colocated) — regression found & fixed, post1 vs main

Tested on **H100**, single-GPU FP8 colocated (`qwen3_omni_fp8_colocated.yaml`, `marksverdhei/Qwen3-Omni-30B-A3B-FP8`), full speech pipeline (thinker → talker → code2wav) via `/v1/chat/completions` with `modalities: ["text", "audio"]`.

### Regression found (now fixed — commit `dc9839e4`)
post1 **crashed the FP8 `talker_ar` stage during CUDA graph capture**:

```
RuntimeError: sgl_per_token_group_quant_8bit_v2(...) failed to dispatch data type Float8_e4m3fn
Process talker_ar died during startup (exit code 1)
```

Root cause (our code, not upstream): the shared thinker attention `forward_core` cast `attn_output` to `o_proj.weight.dtype`. For native FP8 checkpoints that weight is `float8_e4m3fn`, so the activation got cast to fp8 and handed to `per_token_group_quant_fp8`. The pre-bump sgl_kernel tolerated an fp8 input there; post1's `sgl_per_token_group_quant_8bit_v2` has no fp8-input dispatch, so capture aborts and kills the talker stage at startup. Fixed by casting `attn_output` to the activation compute dtype (`v.dtype`, bf16) instead of the quantized weight storage dtype. **This is also the most likely cause of the qwen3-omni H20 CI startup crash** (same FP8 talker graph-capture path) — to be confirmed when an H20 box is available.

### A/B after the fix

| metric | post1 | main |
|---|---|---|
| RTF mean | ~0.29 | ~0.31 |
| throughput | ~3.8 req/s | ~3.8 req/s |
| errors | 0 | 0 |
| audio | ✅ 112KB wav, correct | ✅ correct |

Steady-state across 5 runs (n=16–32, c=4): post1 RTF is consistently ~6% **lower** (better) than main, throughput within noise. **No regression** — at parity / marginally ahead once the talker fix is applied.

Hayden727:
## Ming-flash-omni-2.0 (TP=8) — post1 vs main

Tested on **H100** (TP=8, all 8 GPUs), `inclusionAI/Ming-flash-omni-2.0` (~222GB bf16) via `/v1/chat/completions`.

| | post1 | main |
|---|---|---|
| load | ✅ | ❌ cannot import |
| E2E | ✅ "Paris." | — |
| throughput (chat n=16–24, c=8) | ~0.26 req/s | — |
| errors | 0 | startup crash |

**main cannot load Ming-flash-omni-2.0.** Constructing the `image_encoder` stage hits a circular import in the pre-bump sglang stack:

```
ImportError: cannot import name 'LinearBase' from partially initialized module
'sglang.srt.layers.linear' (most likely due to a circular import)
```

triggered when `ming_omni/components/projectors.py` imports `sglang.srt.layers.linear` (→ `quantization/__init__` → `auto_round` → back to `linear.LinearBase`). It is deterministic and reproducible (a bare `import VisionProjector` fails the same way). post1's sglang restructures these imports, so it loads and serves Ming cleanly at TP=8 with correct chat output.

Bump-enabled capability — no throughput A/B to compare. For the record, Ming is a 222GB omni MoE; ~0.26 req/s on a 64-token chat at TP=8 is the post1 baseline (heavy by nature, not a regression).

Hayden727:
## LLaDA2-Uni — fair-condition regression root cause (deep-dive)

Under a **strictly fair eager-vs-eager** comparison on H100 (both `disable_cuda_graph=True`, multi-block workload, KV-cache fix applied), post1 is **~6–7% slower** than main (~1.18 vs ~1.30 req/s). Here is what it actually is — and what it is **not**.

### It is NOT kernels, launches, or a single serialization point
Torch-profiler timeline (CUDA activity) + py-spy (launch-mode):

| | post1 | main |
|---|---|---|
| c=1 GPU busy | 1068ms | 1088ms (post1 *less*) |
| c=1 host CUDA API | 749ms | 735ms (within 2%) |
| c=8 GPU busy | 3626ms | 3603ms (**equal**) |
| c=8 WALL | +885ms | — |
| c=8 GPU idle | 78.7% | 77.7% |
| └ of the +885ms, CUDA API | only +80ms | — |

- **GPU work is identical** (post1 even slightly less); host CUDA API (launch/sync/memcpy) within 2–3%.
- The extra wall is ~780ms of **non-CUDA-API CPU "dead air"**; the serial `bs=1` diffusion decode is inherently launch/CPU-bound (~25K tiny kernel launches/request).
- py-spy shows that CPU time is **distributed** (top leaf only 5.7% = upstream bf16 linear `apply`; `_group_limited_topk` ~5%; op dispatch 4%; attention ~3%). **No dominant hot-spot; `low_confidence` is not even in the top frames.** The earlier "low_confidence serialization" hypothesis was wrong.

### Conclusion
The ~6–7% is the **eager floor**: the cumulative effect of thousands of tiny eager ops with marginally different host cost on torch 2.11 (per-op dispatch microbenches identical between 2.9 and 2.11). It is **not isolable to a fixable point** — there is no clean monkeypatch lever. The only structural elimination is **CUDA graph** (collapses the per-step launch storm → measured **3.6× faster, 1.18 → 4.26 req/s, output coherent**), but post1's dllm graph captures the full bs set `[1,2,4,8,12,16]` while main's is capped at `bs=[1]`, so that is a **capability** win, not a fair same-mode kernel comparison.

**Bump decision for LLaDA2 is binary:** accept the small fair-condition ~6–7%, or ship cuda graph for post1 (capability win that reverses it to 3.6×). There is no third "fair fix".

sglang_omni/models/higgs_tts/request_builders.py
```diff
@@ -40,6 +40,10 @@ def reset_request(self, req_id: str) -> None: ...
 _HiggsResultAdapter = Callable[[HiggsSGLangRequestData], StagePayload]
 
 
+def _perf_counter() -> float:
```
yxs, line 43: why `_perf_counter` seam only in higgs, will creates cross-model inconsistency

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -532,6 +533,12 @@ def _prepare_request_limits(self, req_data: Any) -> str | None:
             req_data.max_new_tokens = int(req.sampling_params.max_new_tokens)
         return None
 
+    def _ensure_request_limit_compat_defaults(self) -> None:
```
yxs: why this? looks like a dead on the normal path.
`OmniScheduler.__init__` already sets `self.page_size` at and `self.max_total_num_tokens`,  both `hasattr` guards in this helper are always true on the normal init path and the body never runs, right?

yxs (approved):

yxs:
The suite was all pass on H20, a couple of stages are pre-existing flake and may just need a CI re-run or two to go green.

###  MMSU
MMSU speed P95 recalibrated.

The new stack (sglang 0.5.8→0.5.12.post1 + transformers 4.57→5.6 + torch 2.9→2.11) is ~16% slower on MMSU: 60.8 → 50.6 qps (3 CI rounds, σ≈0.1).
This is diffuse per-request host overhead, NOT GPU compute — all major kernels (MoE / attention / audio encoder) are speed-identical cross-version (util-ratio == qps-ratio).
Not a single fixable kernel/config, so P95 is recalibrated to the 0.5.12 reality (62.5 → 50.6 qps).

This added host-side overhead is general to the bump, not MMSU-specific, the same GPU under-utilization shows up in other models' profiles too. Two follow-up directions (out of scope for this PR):
1. Upstream — trim the per-request host overhead in the bumped framework (sglang scheduler / IPC, transformers).
2. Optimization — the pipeline is host/IPC-bound (GPU idle most of the wall), so there's large headroom to optimize it directly (cut cross-stage IPC, better overlap / batching), independent of the bump.

###  LLaDA2-Uni
@yuanluo Leaving the LLaDA2-Uni post1 regression to you. One profiling note: this pipeline is host/IPC-bound, the GPU is busy only ~22% of the wall, so there's a lot of headroom to optimize.


cc @Hayden727 @zhaochenyang20

PR #700, opened 2026-06-05
[Higgs]: Refine Higgs decode mode CLI

Ratish1:
## Motivation

Higgs TTS now defaults to async decode, so the old `--async-decode default|on|off` flag exposed implementation wording and made the no-argument default harder to reason about. This PR replaces that user-facing surface with an explicit decode-mode selector: omit the flag to use the pipeline default, or pass `--decode-mode async|sync` when an override is needed.

## Modifications

- Replace the Higgs serve-time `--async-decode` / hidden `--enable-async-decode` flags with `--decode-mode async|sync`. Internally, the override still maps to the existing `enable_async_decode` scheduler factory argument.
- Rename the advanced min-batch tuning flag to `--async-lookahead-min-batch-size`, while continuing to map it to the existing internal `async_decode_min_batch_size` field. The CLI now rejects this lookahead threshold when combined with `--decode-mode sync`, since sync mode never uses async lookahead.
- Replace the Higgs CLI unit coverage with decode-mode tests covering the async default, explicit async/sync overrides, omitted-mode behavior, threshold validation, unsupported pipeline configs, and removal of the old async-decode flag spellings.
- Update the tests README entry to point at the new decode-mode test file and CLI contract.

## Related Issues

N/A

## Accuracy Test

- Unit tests: passed.

## Benchmark & Profiling

N/A. This is a CLI/config override change and does not alter the Higgs scheduler or model execution path.

## Checklist

- [x] Add unit tests.

## CI


sglang_omni/cli/serve.py
```diff
@@ -973,36 +973,26 @@ def serve(
...
         ),
     ] = False,
-    async_decode: Annotated[
-        str,
+    decode_mode: Annotated[
+        str | None,
         typer.Option(
-            "--async-decode",
-            "--async_decode",
+            "--decode-mode",
```
yxs, line 1024: The `serve` command uses `allow_extra_args=True` + `ignore_unknown_options=True`. This means once a flag is no longer a registered typer option, it does NOT raise "unknown option" — it falls through to `ctx.args` → `ConfigManager.parse_extra_args` → `merge_config` → `config_cls(**cfg_copy)`.

PopSoda2002 (commented):
LGTM! Can we make the default is async decode?

Ratish1:
Yes @PopSoda2002 , Higgs already defaults to async decode through the pipeline config (`enable_async_decode=True`). The new `--decode-mode sync` flag is only for explicitly opting out; omitting the flag keeps async decode as the default.

zhaochenyang20 (approved):

Ratish1:
TTS CI has passed
Screenshot 2026-06-13 at 3 35 07 PM

Issue #702, opened 2026-06-05
[Feat] Shared TTS pipeline-state base + stage helpers


Ccyest:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang-omni/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

The per-request state object that flows between pipeline stages is currently defined
per model, so its serialization and a few stage helpers are repeated across every TTS
model. `#512`'s `EncoderModelRunner` reads and writes this state, and would benefit from
a shared contract to type against. This issue extracts that small shared surface so
`#512` and future model integrations build on one base.
### Problem (evidence in current tree)

**No shared base for the per-request state.** Six standalone `@dataclass`es, zero
inheritance, each with its own `to_dict`/`from_dict`:

| Class | File | LOC |
|---|---|---|
| `HiggsTtsState` | `models/higgs_tts/payload_types.py` | 105 |
| `Qwen3TTSState` | `models/qwen3_tts/payload_types.py` | 106 |
| `MossTTSState` | `models/moss_tts/payload_types.py` | 110 |
| `S2ProState` | `models/fishaudio_s2_pro/payload_types.py` | 113 |
| `Qwen3OmniPipelineState` | `models/qwen3_omni/payload_types.py` | — |
| `LLaDA2UniPipelineState` | `models/llada2_uni/payload_types.py` | — |

The shared subset across them is small but identical: the token-usage triplet, the
decoded-audio tail, a repeated torch→list serialization helper, the
`load_state` / `store_state` wrappers in every `stages.py`, and a byte-for-byte
identical `_resolve_checkpoint` in four stage modules.

### Proposal

Add to two existing shared homes:

1. `scheduling/pipeline_state.py` — `PipelineStateBase` (common fields + generic
   `to_dict`/`from_dict` with an overridable `_serialize_value` tensor hook) + generic
   `load_state` / `store_state` / `build_usage`. Sits next to `SGLangARRequestData`.
2. `utils/checkpoint.py` — `resolve_checkpoint()`.

Model states subclass `PipelineStateBase` and declare only model-specific fields —
base owns the mechanism, subclass owns the semantics.

### Non-goals

- Don't unify input / generation params — they legitimately differ per model; the base
  covers only usage + audio tail.
- Keep each model's tensor-serialization policy via `_serialize_value`; behavior-check
[...]

Issue #661, opened 2026-06-02
[Epic] Large-scale refactor: extract reusable components, enforce code/comment norms, split CI


zhaochenyang20:
Two problems compound each other:

1. **Per-model code is bloating**, mostly from agent-written PRs that pass accuracy CI but are hard to extend. Three failure modes from real review:
   - **Filler comments.** A new TTS model lands ~3,000 LOC; ~500 of those are meaningless agent-emitted comments that carry no information and hurt readability.
   - **No reuse of existing abstractions.** Agents reimplement server start/stop for every benchmark instead of reusing `managed_omni_server` / `start_server_from_cmd` / `stop_server` / `wait_for_gpu_memory_release` in `benchmarks/benchmarker/utils.py`.
   - **Redundant unit tests.** `tests/unit_test/` is 29,291 LOC vs only 6,221 LOC of model/integration tests (`tests/test_model/`), a ~5× imbalance, for behavior already covered by per-model accuracy CI.

   Current per-model size (`sglang_omni/models/`): ming_omni 10.2k, qwen3_omni 9.7k, fishaudio_s2_pro 7.1k, higgs_tts 4.3k, voxtral_tts 3.1k, qwen3_tts 2.9k, moss_tts 2.4k. ming_omni and qwen3_omni are the worst.

2. **Every TTS model converges on the same pipeline shape and then re-implements the same optimizations locally.** Reading the Higgs and MOSS roadmaps, the identical patterns recur as model-local code: reference-audio preprocessing/encoding, stage output caching, batched encoder/vocoder, streaming vocoder state, async decode, CUDA graph capture, per-request pipeline-state serialization. Each is written 2–5× across `higgs_tts / moss_tts / moss_tts_local / fishaudio_s2_pro / qwen3_tts`.

We already have the proof the fix works: **#476 replaced `HiggsScheduler` with `OmniScheduler`, deleted ~440 LOC, and lost nothing.** That is the template for the entire optimization surface, not just the scheduler.

Adding the N-th TTS model means writing model math + declaring capabilities, not copy-pasting an optimization stack.

Push optimization *mechanics* into shared base classes parameterized by model-specific hooks; keep only model *math* (sampling, codec sessions, dense/MoE architecture, row layout) in the model directory.

Framework owns reusable mechanics. Model directories own model semantics.

- **Comments: no AI filler.** A comment is either owner-attributed (`# note (<your name>): ...`) or deleted — nothing in between.
[...]

PR #707, opened 2026-06-06
[WIP][Higgs]: Split TTS runtime ownership and centralize SGLang Req attrs

Ratish1:
## Motivation

Part of #661.

This PR keeps Higgs as the starter refactor target while avoiding fake shared abstractions. Higgs-only lifecycle and state-machine code now lives under `sglang_omni/models/higgs_tts/`, and the only new shared helper is the current SGLang `Req` attr adapter used by multiple model request builders.

The intended review boundary is: Higgs runtime decomposition first, one proven shared SGLang request helper second.

## Modifications

- Split Higgs-owned runtime responsibilities out of the large stage/model-runner files:
  - `higgs_tts/codebook_layout.py` owns Higgs delayed-codebook layout helpers and codec special ids.
  - `higgs_tts/preprocessing.py` owns Higgs input normalization, reference preparation, and reference-audio encode handoff.
  - `higgs_tts/reference_audio.py` owns Higgs reference-audio cache keys and reference-code fingerprints.
  - `higgs_tts/sampler_runtime.py` owns the sampler pool, CUDA-graph staging buffers, and per-request sampled-code storage.
  - `higgs_tts/streaming_vocoder.py` owns Higgs streaming window/delta planning while `vocoder_scheduler.py` keeps scheduler integration.

- Kept model-specific code local instead of over-generalizing it:
  - no `tts_common` package in the final PR surface;
  - codec waveform normalization/bucketed dispatch remains inside `higgs_tts/audio_codec.py` because it has no second caller with the same contract;
  - Higgs sampler EOC wind-down and delayed-row streaming remain Higgs-specific;
  - TTS usage/state helpers remain local because #702 proposes a broader `PipelineStateBase`/stage-helper design for that surface.

- Added `sglang_omni/scheduling/sglang_request.py` with `attach_sglang_req_attrs()`.
  - This centralizes repeated SGLang `Req` compatibility attrs used by Higgs, FishAudio S2-Pro, MOSS, Qwen3-TTS, and Voxtral request builders.
  - The helper uses an `_UNSET` sentinel so omitted attrs are left untouched while explicit `None` values still set attrs exactly like the previous per-model code.

- Removed section-divider comment noise in the touched Higgs test/refactor surface.

## Speed Tests and Profiling

Not rerun locally after this final ownership-boundary rewrite.

## Checklist

- [ ] Add unit tests.

## CI


Ratish1:
superseded by the broader TTS refactor

PR #715, opened 2026-06-07
fix(api): gate speech usage headers behind opt-in

### LINKED ISSUE #174: [Refactor] Decouple TTS usage metrics from OpenAI-compatible `/v1/audio/speech` response
## Background

In PR #173 we added `X-Prompt-Tokens`, `X-Completion-Tokens`, and `X-Engine-Time` as custom HTTP response headers on the `/v1/audio/speech` endpoint to expose per-request token usage and engine timing for benchmarking.

While this works, it **does not conform to the OpenAI `/v1/audio/speech` API specification**. The OpenAI speech endpoint returns a raw audio body with no usage metadata — there is no standard way to carry token counts or timing info in the response.

Embedding custom headers is a pragmatic short-term solution, but it diverges from the API contract we claim to be compatible with, and may confuse clients that expect strict OpenAI compatibility.

## Problem

- Custom `X-*` headers on a compatibility endpoint break the "drop-in replacement" promise.
- There is no OpenAI-standard mechanism to return usage info from the speech endpoint.
- As we add more models and metrics, stuffing everything into headers does not scale.

## Possible Directions (open for discussion)

1. **Separate `/v1/audio/speech/usage` or query param** — return usage in a sidecar endpoint or via `?include_usage=true` that wraps the response in JSON.
2. **Trailing headers (HTTP chunked)** — send audio as chunked body, append usage as trailing headers. Requires client support.
3. **Keep headers but behind an opt-in flag** — only emit `X-*` headers when the client sends a specific request header (e.g., `X-Include-Usage: true`), so default behavior stays OpenAI-compatible.
4. **SSE / streaming mode with structured events** — similar to chat completions streaming, emit audio chunks + a final `usage` event.

None of these is clearly superior. Community input is welcome.

## Current Behavior (status quo)

The `/v1/audio/speech` endpoint returns:
- **Body**: raw WAV/MP3 audio bytes
- **Headers**: `X-Prompt-Tokens`, `X-Completion-Tokens`, `X-Engine-Time` (non-standard)

## Relevant Code

- `sglang_omni/serve/openai_api.py` — header injection in `_register_speech`
- `sglang_omni/models/fishaudio_s2_pro/pipeline/stages.py` — usage dict construction in vocoder stage
- `sglang_omni/client/types.py` — `UsageInfo.engine_time_s` field
- `benchmarks/benchmark_tts_speed.py` — reads `X-*` headers for metrics

jinzijian:
## Motivation

Implements Option A from #174.

`/v1/audio/speech` should stay OpenAI-compatible by default and return raw audio without non-standard usage headers. Benchmark/debug callers can still request usage headers explicitly.

## Changes

- Add an explicit `X-Include-Usage: true` opt-in for non-streaming speech usage headers.
- Keep `X-Prompt-Tokens`, `X-Completion-Tokens`, and `X-Engine-Time` out of default speech responses.
- Update the TTS benchmark HTTP send path to request usage headers explicitly, preserving existing token/engine-time metrics.
- Add focused OpenAI API tests for default behavior, opt-in behavior, and opt-in with missing usage.

Fixes #174

## Testing

- `pytest tests/unit_test/serve/test_openai_api.py -q`
- `python -m ruff check sglang_omni/serve/openai_api.py benchmarks/tasks/tts.py tests/unit_test/serve/test_openai_api.py`

### DIFF
--- a/benchmarks/tasks/tts.py
+++ b/benchmarks/tasks/tts.py
@@ -65,6 +65,7 @@ MOSS_TTS_TOKEN_COUNT_AUTO = "auto"
 MOSS_TTS_ZH_TOKENS_PER_CHAR = 3.098411951313033
 MOSS_TTS_EN_TOKENS_PER_CHAR = 0.8673376262755219
 MOSS_TTS_MIN_AUTO_TOKEN_COUNT = 32
+TTS_USAGE_HEADERS = {"X-Include-Usage": "true"}
 UTMOS_BATCH_SIZE = 8
 
 
@@ -1859,7 +1860,9 @@ def make_tts_send_fn(
         )
         start_time = time.perf_counter()
         try:
-            async with session.post(api_url, json=payload) as response:
+            async with session.post(
+                api_url, json=payload, headers=TTS_USAGE_HEADERS
+            ) as response:
                 if response.status != 200:
                     result.error = f"HTTP {response.status}: {await response.text()}"
                 elif stream and stream_format == "audio":
--- a/sglang_omni/serve/openai_api.py
+++ b/sglang_omni/serve/openai_api.py
@@ -18,9 +18,10 @@ import json
 import logging
 import time
 import uuid
+from collections.abc import Mapping
 from typing import Any
 
-from fastapi import FastAPI, File, Form, HTTPException, UploadFile, WebSocket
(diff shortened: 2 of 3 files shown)


jinzijian:
Closing this draft for now while I self-review and run more local tests. I will reopen a PR when it is ready.

Issue #599, opened 2026-05-27
Qwen3-TTS and Voxtral-TTS Roadmap


GaokaiZhang:
## Goal
Low latency and high throughput on Qwen3-TTS and Voxtral-TTS with no quality regression vs official numbers.

## Done

- [x] Model support — #444 / #451 @GaokaiZhang
  - Earlier attempts: #76 @adarshxs, #248 @liusy58, #466 @draftbk
- [x] Serving contract hardening (request-owned seeds, audio_waveform bytes, explicit `gpu_id`) — #570 @Ratish1
- [x] Full SeedTTS EN/ZH H200 baseline (post-#570 tables)
- [x] Voxtral and Qwen3-TTS cookbooks — #585 @JiaxinD

### Evaluation Adding
- [ ] seed-tts speaker similarity benchmark for Qwen3-TTS and Voxtral (per #444 acceptance) — pattern: #469
- [ ] benchmark performance in different hardware nodes like H100, H200
- [ ] score-alignment writeup vs Mistral / Qwen official numbers

### Code refactor
- [ ] Factor `BaseTTSStageScheduler` — #450 @PopSoda2002
- [ ] Migrate Qwen3-TTS and Voxtral-TTS off `SimpleScheduler` after #450

### Bug
- [ ] Qwen3-TTS 0.6B EN max-length repetition outlier (`max_new_tokens=2048` hits)
  - Note: diagnosed in #570 as stochastic decode tail — EOS rank ~1500–1800, prob ~1e-16, outside `top_k=50`. Fixed-seed replay reproduces. #570 kept Qwen's default decoding policy. **TODO: evaluate alt `top_k` / temperature / repetition_penalty / retry-on-length** on the captured bad seed; verify no regression on 1.7B / ZH / Voxtral.

### Optimization
Reference: https://www.baseten.co/blog/cost-efficient-high-performance-qwen3-tts/#dynamic-frame-accumulation
- [ ] torch.compile + CUDA Graph for AR backbones — #520 @GaokaiZhang : https://github.com/sgl-project/sglang-omni/pull/527 (H200 acceptance rows: qps +11–21%, RTF -9 to -17%)
- [ ] Voxtral flow-matching head capture (deferred from #520; per-NFE shape semantics differ from AR)
- [ ] Qwen3-TTS continuous prompt embedding × torch.compile audit (#520 pitfall — radix-cache hash must survive re-trace)
[...]

PR #718, opened 2026-06-08
[Voxtral] Cross-request batched vocoder decode

0xtoward:
## What

Coalesce concurrent Voxtral vocoder (audio-tokenizer decode) requests into a single `decode_helper_batch_async` call via `SimpleScheduler`'s `batch_compute_fn`, instead of decoding one request at a time. Implements the **"batched vocoder decode for Voxtral"** item from the Voxtral roadmap (#599).

This mirrors the Higgs batched vocoder decode (#574, merged), applied to Voxtral.

## Why it's safe (causality)

Voxtral's decoder is fully causal (`CausalConv1d` / `CausalConvTranspose1d`). `decode_helper_batch_async` already pads each request's chunks to the max chunk length and crops every output back to its own length; because the convolutions are causal, this pad-to-max + crop **does not contaminate across requests of different lengths** (a non-causal codec — e.g. Higgs' DAC — cannot be padded this way and must bucket by exact length).

Each request keeps its own warmup-context frames (prepended) and trims them independently after decode, so batching is purely a throughput optimization with no cross-request coupling.

Batched output matches single-request decode up to bf16 batch-dependent conv-algorithm noise (rel ~1e-2, quality-neutral) — the same property leveraged by #574, gated on WER rather than bit-exactness.

## Correctness

- New unit test `tests/unit_test/voxtral_tts/test_vocoder_batch.py` (deterministic fake tokenizer): asserts the batch path produces, per request, output identical to the single path across mixed request lengths (warmup-trim / fade applied per request, results routed back to the correct payload), and that it issues exactly one decode call covering the whole batch.
- On-device WER validation (below) confirms the bf16 numeric noise is quality-neutral.

## Performance

A800, `Voxtral-4B-TTS-2603`, `benchmark_tts_seedtts` (concurrency 16, 128 samples, `cheerful_female`, en):

| metric | baseline (batch=1) | batched (batch=4) | delta |
|---|---|---|---|
| throughput (req/s) | 6.063 | 6.869 | **+13.3%** |
| audio throughput (s/s) | 31.51 | 34.545 | +9.6% |
| mean latency (s) | 2.35 | 2.21 | −6% |
| RTF mean | 0.457 | 0.441 | better |
| WER corpus (whisper large-v3) | 0.51% | 0.44% | no regression |
| WER >50% outliers | 0 / 128 | 0 / 128 | — |
| success | 128 / 128 | 128 / 128 | — |

## Config

Batch size defaults to 4 (`max_batch_wait_ms=2`), overridable via `SGLANG_OMNI_VOXTRAL_VOCODER_MAX_BATCH` (set to `1` to disable batching). No change to `audio_tokenizer.py` or `simple_scheduler.py` — both already support this path.

### DIFF
--- a/sglang_omni/models/voxtral_tts/pipeline/stages.py
+++ b/sglang_omni/models/voxtral_tts/pipeline/stages.py
@@ -350,16 +350,38 @@ def create_vocoder_executor(
     *,
     device: str = "cuda:0",
     gpu_id: int | None = None,
+    max_batch_size: int = 4,
+    max_batch_wait_ms: int = 2,
 ) -> SimpleScheduler:
-    """Factory for the vocoder (audio tokenizer decode) stage."""
+    """Factory for the vocoder (audio tokenizer decode) stage.
+
+    Concurrent requests are coalesced into a single ``decode_helper_batch_async``
+    call (``batch_compute_fn`` below). Voxtral's decoder is fully causal
+    (``CausalConv1d`` / ``CausalConvTranspose1d``), so the helper's pad-to-max +
+    crop does **not** contaminate across requests of different lengths (a
+    non-causal codec could not be padded this way). The batched output matches
+    single-request decode up to bf16 batch-dependent conv-algorithm noise
+    (rel ~1e-2, quality-neutral — WER-validated, same property as the Higgs
+    batched vocoder decode #574 / MOSS #651).
+    """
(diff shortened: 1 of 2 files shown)


0xtoward:
Closing for now — want to do a deeper perf pass (profiling + larger batch sizes) before this goes up. Will resubmit once the throughput story is more complete.

PR #719, opened 2026-06-08
Refactor MOSS decode state into row-indexed pool

guozhihao-224:
## Summary
- Add `MossDecodeStatePool` for stable per-request / row-indexed decode state.
- Store next-step feedback, delay state, sampling steps, generated history, and stop state in the pool.
- Make MOSS decode input read feedback from the pool when available.
- Write sampled feedback/history/step state back to the pool during collect.
- Add `MossDecodeJournal` so output-row append can be applied from journal state instead of runner-global `_pending_rows/_pending_embeds`.
- Release/reset pool rows from the result adapter.

### DIFF
--- a/sglang_omni/models/moss_tts/model_runner.py
+++ b/sglang_omni/models/moss_tts/model_runner.py
@@ -10,6 +10,7 @@ from sglang.srt.layers.sampler import multinomial_with_seed
 
 from sglang_omni.model_runner.base import ModelRunner
 from sglang_omni.models.moss_tts.request_builders import _INF_DELAY
+from sglang_omni.models.moss_tts.state_pool import MossDecodeJournal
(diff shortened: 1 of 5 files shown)


Ccyest:
Hi @guozhihao-224, thanks for contribution!
MOSS just had new modal v1.5 so we are closing all old structured based optimization, but we will co-author you for credit.

Also, what is your wechat id? I can add you to developer community.

guozhihao-224:
> Hi @guozhihao-224, thanks for contribution! MOSS just had new modal v1.5 so we are closing all old structured based optimization, but we will co-author you for credit.
> 
> Also, what is your wechat id? I can add you to developer community.

Okay, my WeChat ID is: 335247945 @Ccyest

PR #722, opened 2026-06-09
[Quant]: Add unified quantization support for AutoRound W4A16 and FP8

lvliang-intel:
## Motivation

This PR adds inference support for models quantized with [AutoRound](https://github.com/intel/auto-round) ([paper](https://arxiv.org/abs/2309.05516)), a post-training quantization method that delivers significantly higher accuracy at low bit-widths (e.g., 2-bit) and broad model compatibility across LLMs and VLMs. AutoRound is already integrated into vllm, vllm-omni, pytorch/ao, llm-compressor, and Hugging Face transformers.
The core goal is to avoid model-specific quantization branching by introducing a small, unified quantization abstraction, then building on it to support:

- AutoRound INT4 checkpoint detection and loading
- FP8 checkpoint detection and runtime configuration
- Qwen3-Omni stage-aware quantization config handling (thinker/talker colocated architecture)

**Why this matters for serving**: AutoRound INT4 cuts weight VRAM usage substantially vs. BF16, unlocking higher max concurrency under the same GPU memory budget. On RTX6000D, INT4 at concurrency=16 achieves **0.878 req/s** vs. BF16's peak of **0.352 req/s** at concurrency=4 — a **2.5×** throughput improvement — making larger Qwen3-Omni deployments practical on constrained hardware.

https://huggingface.co/Intel/Qwen3-Omni-30B-A3B-Instruct-int4-AutoRound

**Setup**
```bash
uv venv .venv -p 3.12
source .venv/bin/activate
uv pip install -v -e .
```

## Modifications

This PR keeps the scope focused on quantization-related code paths only.

1. Introduce a lightweight quantization abstraction layer
A new quantization package defines a minimal QuantizationMethod ABC with a lazy auto-registration registry. The base class exposes only the four contracts that are genuinely quantization-specific (name, detect, configure, preprocess_weights), while layer construction and weight-loading mechanics are left to SGLang's native paths. A single public entry point (resolve_weight_preprocessor) returns a uniform (name, tensor) → tensor callable so every weight-loading site gets the same preprocessing logic without knowing which method is active. Config extraction walks composite-model configs transparently, handling nested sub-configs and both quantization_config / compression_config keys.

2. Add AutoRound INT4 and FP8 as concrete method implementations
Two built-in methods ship under the new abstraction: AutoRound (INT4 packed-weight deshuffling, group-size reshaping, and MoE backend wiring) and FP8 (weight_scale_inv reciprocal conversion). Backend-policy decisions (MoE runner, GEMM runner) that previously lived inside the FP8 method's configure() are moved up into model_worker's existing policy function so the method itself stays thin and policy is centralized.

3. Wire the abstraction into existing serving paths
Model-worker backend policy resolution now goes through the unified detection/config path once, then delegates to the method's configure() before applying Omni-specific overrides. Stage weight loaders (thinker, talker) replace hard-coded per-quantization helper calls with a single resolve_weight_preprocessor() invocation, making them quantization-agnostic. Config-manager stage-parallelism alias resolution is extended to match by stage name in addition to numeric index.

## Related Issues

N/A

### MMMU accuracy results
| Hardware | Precision | Max Concurrency | Total Samples | Correct | Accuracy | Failed Requests | MC Parse Fallback |
| A100 | INT4 | 32 | 50 | 30 | 60.0% | 0 | 3 |
| A100 | INT4 | 16 | 50 | 32 | 64.0% | 0 | 1 |
| A100 | INT4 | 8 | 50 | 33 | 66.0% | 0 | 0 |
| A100 | BF16 | 4 | 50 | 28 | 56.0% | 0 | 2 |
| A100 | BF16 | 2 | 50 | 30 | 60.0% | 0 | 4 |
| RTX6000D | BF16 | 4 | 50 | 31 | 62.0% | 0 | 1 |
| RTX6000D | BF16 | 2 | 50 | 30 | 60.0% | 0 | 1 |
| RTX6000D | BF16 | 8 | 50 | 8 | 16.0% | 40 | 1 |
| RTX6000D | INT4 | 32 | 50 | 32 | 64.0% | 0 | 1 |
| RTX6000D | INT4 | 16 | 50 | 31 | 62.0% | 0 | 1 |
| RTX6000D | INT4 | 8 | 50 | 31 | 62.0% | 0 | 0 |
| RTX6000D | FP8 | 32 | 50 | 30 | 60.0% | 0 | 2 |
| RTX6000D | FP8 | 16 | 50 | 28 | 56.0% | 0 | 2 |
| RTX6000D | FP8 | 8 | 50 | 28 | 56.0% | 0 | 0 |

### Notes
- A100 BF16 currently only works in eager mode in this setup.
- `flashinfer_cutlass` reports `Invalid backend: 80` on A100 BF16 in this environment, known issue
- Under this limitation, A100 BF16 could only be validated up to `max-concurrency=4`.
- RTX6000D BF16 at `max-concurrency=8` showed instability with 40 failed requests, so the more stable BF16 operating points were `max-concurrency=2` and `4`.
- AutoRound INT4 accuracy is on par with or better than BF16 at all stable operating points.

## Benchmark & Profiling

AutoRound INT4 reduces weight memory enough to unlock meaningfully higher serving concurrency than the BF16 baseline on the same GPU.

### MMMU speed results
| Hardware | Precision | Max Concurrency | Completed | Failed | Latency Mean (s) | Latency Median (s) | Latency P95 (s) | Latency P99 (s) | Output Throughput (tok/s) | Output Tokens / Request-s | Output Tokens Mean | Prompt Tokens Mean | Throughput (req/s) |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| A100 | INT4 | 32 | 50 | 0 | 54.607 | 51.716 | 102.977 | 106.174 | 327.9 | 12.8 | 700 | 1307 | 0.469 |
| A100 | INT4 | 16 | 50 | 0 | 23.887 | 13.334 | 77.952 | 77.990 | 357.0 | 26.2 | 627 | 1307 | 0.570 |
| A100 | INT4 | 8 | 50 | 0 | 18.643 | 8.690 | 54.762 | 57.468 | 275.9 | 36.5 | 681 | 1307 | 0.405 |
| A100 | BF16 | 4 | 50 | 0 | 39.593 | 25.195 | 109.096 | 110.075 | 62.3 | 15.9 | 631 | 1307 | 0.099 |
| A100 | BF16 | 2 | 50 | 0 | 30.370 | 14.647 | 104.449 | 104.673 | 37.9 | 19.2 | 583 | 1307 | 0.065 |
| RTX6000D | BF16 | 4 | 50 | 0 | 10.848 | 6.712 | 30.912 | 32.812 | 221.8 | 58.0 | 629 | 1307 | 0.352 |
| RTX6000D | BF16 | 2 | 50 | 0 | 7.238 | 4.350 | 21.506 | 22.119 | 165.0 | 86.7 | 628 | 1307 | 0.263 |
| RTX6000D | BF16 | 8 | 10 | 40 | 15.786 | 9.072 | 45.613 | 47.117 | 91.9 | 42.7 | 674 | 330 | 0.136 |
| RTX6000D | INT4 | 32 | 50 | 0 | 29.883 | 26.911 | 62.296 | 63.780 | 449.0 | 19.3 | 576 | 1307 | 0.780 |
| RTX6000D | INT4 | 16 | 50 | 0 | 15.336 | 9.231 | 50.457 | 50.551 | 519.7 | 38.6 | 592 | 1307 | 0.878 |
| RTX6000D | INT4 | 8 | 50 | 0 | 11.002 | 6.091 | 35.843 | 37.823 | 395.9 | 53.6 | 590 | 1307 | 0.671 |
| RTX6000D | FP8 | 32 | 50 | 0 | 27.004 | 21.807 | 57.476 | 59.987 | 515.2 | 23.0 | 622 | 1307 | 0.829 |
| RTX6000D | FP8 | 16 | 50 | 0 | 14.189 | 7.337 | 45.266 | 45.534 | 612.7 | 44.3 | 629 | 1307 | 0.974 |
| RTX6000D | FP8 | 8 | 50 | 0 | 10.320 | 6.539 | 31.378 | 31.632 | 447.4 | 59.9 | 618 | 1307 | 0.724 |
### Benchmark commands
```bash
python -m benchmarks.eval.benchmark_omni_mmmu \
  --model qwen3-omni \
  --port <PORT> \
  --max-samples 50 \
  --max-concurrency <N> \
  --output-dir <OUTPUT_DIR>
Example serving commands used in validation:

sgl-omni serve \
  --config examples/configs/qwen3_omni_colocated_h20.yaml \
  --colocate \
  --model-name qwen3-omni \
  --model-path /mnt/ctrl/disk2/lvl/models/Qwen3-Omni-30B-A3B-Instruct \
  --port 8088
sgl-omni serve \
  --config examples/configs/qwen3_omni_colocated_h20.yaml \
  --colocate \
  --model-name qwen3-omni \
  --model-path /mnt/ctrl/disk2/lvl/models/Qwen3-Omni-30B-A3B-Instruct-int4-AutoRound \
  --port 8088

sgl-omni serve \
  --config examples/configs/qwen3_omni_fp8_colocated.yaml \
  --colocate \
  --model-name qwen3-omni \
  --model-path /mnt/ctrl/disk2/lvl/models/Qwen3-Omni-30B-A3B-FP8 \
  --port 8008
```

## Checklist

- [x] Add unit tests.

## CI


docs/cookbook/qwen3_omni.md
```diff
@@ -36,6 +36,7 @@ Colocated topology requires `--config examples/configs/qwen3_omni_colocated_h20.
 |---|---|---|---|---|
 | Thinker-only | — | — | BF16 | ✅ |
 | Thinker-only | — | — | FP8 | ✅ |
+| Thinker-only | — | — | AutoRound INT4 | ✅ |
```
xinlij, line 39: Please update the Server Configuration accordingly.

xinlij (changes requested):

docs/cookbook/qwen3_omni.md
lvliang-intel: docs/_static/js/qwen3_omni_server_generator.js updated

lvliang-intel:
@xinlij,
could you please help to review again? Please approve it if no further comments. Thanks.

sglang_omni/quantization/loader.py
```diff
@@ -0,0 +1,133 @@
...
+    config: dict[str, Any],
+) -> QuantizationConfig | None:
+    """Detect and parse quantization config from model config.
+
+    Args:
+        config: Model config dict (from config.json or hf_config.to_dict())
+
+    Returns:
+        QuantizationConfig or None if no quantization
+    """
```
Ratish1: Should we trim some of these docstring across the PR?

sglang_omni/quantization/methods/fp8.py
```diff
@@ -0,0 +1,202 @@
...
+        # Determine MoE backend
+        moe_runner_backend = getattr(server_args, "moe_runner_backend", "auto")
+        if moe_runner_backend == "auto":
+            if has_moe and has_native_fp8 and self._is_cutlass_supported():
+                server_args.moe_runner_backend = "cutlass"
+            elif not has_moe:
+                server_args.moe_runner_backend = "auto"
+
+        # Determine GEMM backend for dense layers
+        fp8_gemm_backend = getattr(server_args, "fp8_gemm_runner_backend", None)
```
Ratish1: This changes the existing backend contract more broadly than the Qwen3 policy does today. `_configure_backend_policy()` now runs the quantization method config first, so every detected FP8 checkpoint with `fp8_gemm_runner_backend` unset or `"auto"` gets forced to `"triton"` here.

That does not match the existing Qwen3-Omni policy below: talker FP8 gets dense GEMM Triton, but thinker FP8 and non-Qwen FP8 are expected to preserve `"auto"`. The existing backend tests still encode that expectation, but they only call `_apply_model_worker_backend_policy()` directly, so they miss this composed path.

should we remove the global dense GEMM override from `FP8Quantization.configure()`, or gate it with the same arch/stage conditions as `_apply_model_worker_backend_policy()`. Also add a unit test that exercises the full `_configure_backend_policy()` ordering.

sglang_omni/quantization/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+
+    if quant_config:
+        from sglang_omni.quantization import QuantizationRegistry
+        method = QuantizationRegistry.get(quant_config.method)()
+        method.configure(server_args, model_config)
+"""
+
+from sglang_omni.quantization.base import QuantizationMethod, QuantizedLinearSpec
+from sglang_omni.quantization.config import QuantizationConfig
+from sglang_omni.quantization.loader import (
```
Ratish1: question: should this package expose `QuantizedWeightLoader` as part of the unified quantization surface?

Qwen3-Omni loading already does the right thing by mapping checkpoint names and then delegating to the SGLang per-parameter `weight_loader` methods. That path needs shard ids, expert ids, fused QKV handling, fused MoE handling, and quantized parameter loader contracts. The generic loader here only exact-matches parameter names and cannot replace that production path.

sglang_omni/quantization/methods/autoround.py
```diff
@@ -0,0 +1,368 @@
...
+
+    # Map a sub-model architecture to the checkpoint weight-name prefix that the
+    # corresponding stage strips when loading weights.
+    _ARCH_CHECKPOINT_PREFIX = {
+        "Qwen3OmniThinkerForCausalLM": "thinker.",
+        "Qwen3ASRForConditionalGeneration": "thinker.",
+        "Qwen3OmniTalker": "talker.",
+    }
+
+    def _normalize_block_names_for_stage(self, model_config: Any) -> None:
```
Ratish1: can we add direct unit coverage for this method. This is the behavior that makes Qwen3-Omni AutoRound work with SGLang: the checkpoint can carry `thinker.` / `talker.` prefixes, while the stage-local SGLang modules are named without that owner prefix.

The current unit tests cover generic detection and remapping, but they do not cover `configure()` mutating `model_config.hf_config.quantization_config`. Since SGLang `AutoRoundConfig` matches runtime layer names with `startswith`, this prefix normalization is the main contract to protect.

tests/test_model/test_qwen3_omni_quantization_autoround.py
```diff
@@ -0,0 +1,252 @@
...
+
+    return QuantizationBenchmarkResult(
+        summary=results["summary"],
+        per_sample=results["per_sample"],
+        output_dir=output_dir,
+    )
+
+
+@pytest.mark.benchmark
+def test_autoround_model_loads_and_responds(
```
Ratish1: This test is named `loads_and_responds`, but it never sends a request. It only checks `/workers` and `/v1/models`, so a quantized worker can pass this and still fail on the first forward pass, sampler call, FP8 scale load, or AutoRound kernel path.

The benchmark fixture above does run real requests, but nothing consumes it. Please either use the fixture and call `_assert_benchmark_results()`, or add one tiny chat-completion request here. Keeping it benchmark-marked is fine; the issue is that the committed GPU test should prove generation, not only router registration.

tests/test_model/test_qwen3_omni_quantization_fp8.py
```diff
@@ -0,0 +1,241 @@
...
+
+    return FP8QuantizationBenchmarkResult(
+        summary=results["summary"],
+        per_sample=results["per_sample"],
+        output_dir=output_dir,
+    )
+
+
+@pytest.mark.benchmark
+def test_fp8_model_loads_and_responds(
```
Ratish1: Same concern as the AutoRound test: this is called `loads_and_responds`, but it only checks worker health and model registration. That does not exercise the quantized forward path.

Please either consume `fp8_benchmark_results` and assert the benchmark output, or send one minimal chat-completion request here. For FP8 especially, the first actual request is where scale loading, backend selection, and kernel compatibility problems tend to show up.

Ratish1 (commented):
Great PR @lvliang-intel . Left some comments. Lmk what do you think

sglang_omni/quantization/methods/fp8.py
lvliang-intel: Good catch. That‘s indeed a bug. I've already removed the global dense GEMM override from FP8Quantization.configure(), it now only touches moe_runner_backend.  The related unit test also has been added.

sglang_omni/quantization/__init__.py
lvliang-intel: The generic QuantizedWeightLoader and loader.py have been deleted. The package's public surface is now:
(1) QuantizationMethod, QuantizationConfig, detect_quantization_config, QuantizationRegistry
(2) WeightPreprocessor / resolve_weight_preprocessor
resolve_weight_preprocessor is a thin config-to-callable resolver, it does no weight loading itself. The Qwen3-Omni loading path calls it to get the preprocess_weights hook, then handles shard/expert mapping, fused QKV/MoE, and per-parameter weight_loader calls independently.

sglang_omni/quantization/methods/autoround.py
lvliang-intel: Added: TestAutoRoundConfigureStagePrefix

tests/test_model/test_qwen3_omni_quantization_autoround.py
lvliang-intel: Thanks for these suggestions. Already fixed.

tests/test_model/test_qwen3_omni_quantization_fp8.py
lvliang-intel: Done

Ratish1:
Could you fix the lint as well @lvliang-intel . Thanks

sglang_omni/models/qwen3_omni/components/talker.py
```diff
@@ -45,6 +47,14 @@
 from sglang_omni.vendor.sglang.server_args import get_global_server_args
 from sglang_omni.vendor.sglang.utils import make_layers
 
+# Stay compatible across releases that renamed/split optional fields such as
+# ``acc_linear_penalties`` on SGLang v0.5.8.
```
Ratish1: No need to carry this compatiability over. we already pin sglang==0.5.12.post1

We can remove `_SAMPLING_BATCH_INFO_FIELDS`, the `dataclasses` import, and the conditional `acc_linear_penalties` kwarg

lvliang-intel: done

lvliang-intel:
> Could you fix the lint as well @lvliang-intel . Thanks

lint issues fixed.

sglang_omni/quantization/registry.py
```diff
@@ -0,0 +1,183 @@
...
+
+        quant_method = quant_config.get("quant_method")
+        if quant_method is None:
+            return None
+
+        # Normalize method name for matching
+        normalized = str(quant_method).lower().replace("-", "_").replace(" ", "_")
+
+        # Try exact match first
+        for name, method_cls in cls._methods.items():
```
Ratish1: This exact-name fast path bypasses the method-specific detector. For FP8, that matters: FP8Quantization.detect() explicitly requires weight_block_size, but QuantizationRegistry.detect({"quantization_config": {"quant_method": "fp8"}}) returns FP8Quantization() here before that check ever runs.

can wemake exact-name matching validate through method_cls.detect(config), or split "method name is known" from "this checkpoint is supported by this method". Otherwise non-native/per-tensor FP8 gets classified as the native block-FP8 path.

sglang_omni/config/manager.py
Ratish1: _resolve_list_index() now supports paths like stages.thinker.tp_size, but _sync_stage_parallelism_aliases() still only looks for numeric keys like stages.4.tp_size. So manager.merge_config({"stages.thinker.tp_size": 2, "stages.thinker.gpu": [0, 1]}) updates tp_size but does not mirror it into parallelism.tp, and PipelineConfig then rejects the merged config as conflicting.

can we canonicalize named stage paths to numeric paths before alias sync, or make the sync helper recognize both stages.{index} and stages.{stage_name} forms. Also add the named-stage equivalents of the existing numeric tests.

sglang_omni/quantization/registry.py
```diff
@@ -0,0 +1,183 @@
...
+            name_normalized = name.lower().replace("-", "_").replace(" ", "_")
+            if name_normalized == normalized:
+                return method_cls()
+
+        # Fall back to detection by class
+        for method_cls in cls._methods.values():
+            try:
```
Ratish1: For built-in methods, broad exception swallowing here is too defensive. If FP8Quantization.detect() or AutoRoundQuantization.detect() raises, that is a bug in the method/config shape, not a reason to silently continue and return None. The same applies to auto_register_all(): it catches broad Exception, logs, then sets _initialized = True, which permanently hides a broken built-in import for the process.

fail fast for built-in detection/import errors. If future third-party methods have optional dependencies, catch those at the narrow optional import boundary instead of making the core registry silently degrade.

sglang_omni/quantization/weight_preprocess.py
```diff
@@ -0,0 +1,77 @@
...
+
+    return None
+
+
+def resolve_weight_preprocessor(config: Any = None) -> WeightPreprocessor:
+    """Return the active quantization method's ``preprocess_weights`` callable."""
+    from sglang_omni.quantization.methods.fp8 import FP8Quantization
+
+    method = _resolve_active_method(config)
+    if method is None:
```
Ratish1: A generic resolve_weight_preprocessor() should return an identity preprocessor when no active quantization method is detected. Defaulting to FP8Quantization() makes every unquantized or unknown checkpoint run through an FP8-specific transform, which is surprising and hides the invariant that FP8 scale conversion should only happen for detected FP8 checkpoints.

return a no-op identity function when method is None, and only return FP8Quantization.preprocess_weights after FP8 is actually detected.

sglang_omni/quantization/methods/fp8.py
```diff
@@ -0,0 +1,112 @@
...
+
+    @classmethod
+    def detect(cls, config: dict[str, Any]) -> bool:
+        """Detect FP8 quantization from config."""
+        quant_config = config.get("quantization_config", {})
+        quant_method = str(quant_config.get("quant_method", "")).lower()
+        weight_block_size = quant_config.get("weight_block_size")
+        return quant_method == "fp8" and weight_block_size is not None
+
+    def configure(self, server_args: Any, model_config: Any) -> None:
```
Ratish1: This method still makes backend-policy decisions even though model_worker._apply_model_worker_backend_policy() already owns the Qwen3-Omni FP8 backend policy. The two implementations are not equivalent: this one is not arch-scoped, and _model_has_native_fp8_block_quant() only uses getattr(), while the model-worker helper handles both dict and object quant configs.

I would remove backend selection from FP8Quantization.configure() and keep this class focused on FP8 detection plus weight preprocessing. If method-specific configuration really needs to stay here, please delegate to the same shared helper/policy path used by model_worker, with the same Qwen3 arch guards and dict/object config handling.

sglang_omni/quantization/methods/autoround.py
```diff
@@ -0,0 +1,138 @@
...
+    name = "auto-round"
+
+    @classmethod
+    def detect(cls, config: dict[str, Any]) -> bool:
+        """Detect AutoRound quantization from config."""
+        quant_config = config.get("quantization_config", {})
+        quant_method = str(quant_config.get("quant_method", "")).lower()
+
+        # Match various AutoRound naming conventions
+        return quant_method in (
```
Ratish1: This detector accepts autoround, auto_round, and inc, but pinned SGLang 0.5.12.post1 only registers auto-round in QUANTIZATION_METHODS and --quantization choices. Since SGLang ModelConfig._verify_quantization() runs before this Omni registry can configure anything, those aliases do not actually make such checkpoints load unless the HF config is normalized earlier.

Either remove the aliases and only advertise auto-round, or normalize the checkpoint quantization config before SGLang ModelConfig verification. Keeping tests for aliases that the runtime cannot accept gives a false sense of compatibility.

sglang_omni/quantization/methods/autoround.py
```diff
@@ -0,0 +1,138 @@
...
+        "Qwen3OmniTalker": "talker.",
+    }
+
+    def _normalize_block_names_for_stage(self, model_config: Any) -> None:
+        """Strip the active stage's checkpoint prefix from the quant config."""
+        hf_config = getattr(model_config, "hf_config", None)
+        if hf_config is None:
+            return
+        quant_config = getattr(hf_config, "quantization_config", None)
+        if not isinstance(quant_config, dict):
```
Ratish1: Detection normalizes object-shaped quantization configs, but configure() only mutates dict-shaped configs and silently returns for everything else. If AutoRound was detected, this stage-prefix normalization is not optional and skipping it means SGLang's AutoRoundConfig can target checkpoint names instead of stage-local module names.

we can normalize config access once and handle dict/object consistently, or assert/raise if the detected AutoRound config is not in the expected mutable shape.

sglang_omni/quantization/registry.py
lvliang-intel: Good catch. Indeed an issue here. I have followed your suggestion to refine the code.

sglang_omni/config/manager.py
lvliang-intel: Thanks for this suggestion. `_sync_stage_parallelism_aliases` now builds segments = [str(index), stage_name] and checks overrides against both stages.0.tp_size and stages.thinker.tp_size forms. The named-stage tests are also already included in `test_config_manager.py`.

sglang_omni/quantization/registry.py
lvliang-intel: The fallback loop no longer wraps detect() in a try/except, and `_auto_register_all` no longer catches broad Exception. Both now fail fast instead of silently degrading as your good suggestion.

sglang_omni/quantization/weight_preprocess.py
lvliang-intel: Good finding. When no method is detected, we now return `lambda name, weight: weight` instead of falling back to FP8Quantization().

sglang_omni/quantization/methods/fp8.py
lvliang-intel: FP8Quantization.configure() is now a no-op, and the three backend helper methods have been removed entirely. The class is focused solely on detection and weight preprocessing.

sglang_omni/quantization/methods/autoround.py
lvliang-intel: Narrow detect() to only match "auto-round" since the autoround omni models all have the latest config name.

sglang_omni/quantization/methods/autoround.py
lvliang-intel: Good suggestion. Add `_to_mutable_dict()` which handles dict, to_dict(), and __dict__ shapes consistently, and raises TypeError for anything unsupported instead of silently skipping.

sglang_omni/models/qwen3_omni/components/talker.py
```diff
@@ -1439,6 +1448,8 @@ def load_weights(self, weights: Iterable[Tuple[str, torch.Tensor]]) -> None:
             num_experts=self.config.text_config.num_experts,
         )
 
+        preprocess_weight = resolve_weight_preprocessor(self.config)
```
Ratish1: lets resolve the weight preprocessor from the root Qwen3-Omni config, not from the narrowed talker config.

The public FP8 checkpoint stores `quantization_config` at the root. This constructor rebuilds `Qwen3OmniMoeTalkerConfig` from `config.talker_config.to_dict()`, so `self.config` can no longer expose the root quantization metadata. That means `resolve_weight_preprocessor(self.config)` can return the identity function even though SGLang still constructs FP8 layers from root `hf_config.quantization_config`.

That is a regression from the old unconditional `convert_fp8_weight_scale_inv_for_sglang()` path: talker `weight_scale_inv` tensors may stop getting the Qwen3-Omni reciprocal conversion while the layer is still native FP8. lets keep `self.root_config` before narrowing `self.config` and use it here,

sglang_omni/quantization/config.py
```diff
@@ -0,0 +1,86 @@
...
+            "bits": self.bits,
+            "group_size": self.group_size,
+            "sym": self.sym,
+            "packing_format": self.packing_format,
+        }
+
+    @property
+    def is_block_quantization(self) -> bool:
+        """Check if this is a block quantization method."""
+        return self.group_size > 0
```
Ratish1: can we remove or narrow this public `QuantizationConfig` helper before it becomes a misleading API.

For the FP8 checkpoint supported by this PR, block quantization is identified by `weight_block_size`, not by `group_size`. Local SGLang's `Fp8Config` reads `weight_block_size` and all block-FP8 linear/MoE paths branch on that field. The current `is_block_quantization` implementation only checks `group_size > 0`, and the tests make that pass by fabricating an FP8 config with both `group_size` and `weight_block_size`.

 the simplest fix is to remove `is_block_quantization`, `is_per_channel`, and `to_backend_config()` unless a real caller needs them. If this dataclass stays exported, please add `weight_block_size` and make the block-quant predicate method-aware.

sglang_omni/model_runner/model_worker.py
```diff
@@ -556,6 +580,131 @@ def _is_fp8_cutlass_moe_supported() -> bool:
...
+        return None, None
+
+    # Normalize to the canonical registered name so downstream lookups
+    # succeed regardless of which alias the checkpoint used.
+    quant_config.method = method_instance.name
+
+    return method_instance.name, quant_config
+
+
+def _build_config_dict(model_config: ModelConfig) -> dict | None:
```
Ratish1: Lets collapse the quantization-config discovery logic into one helper.

The PR now has `_get_hf_quantization_config()`, `_build_config_dict()`, and `weight_preprocess._resolve_active_method()` all walking slightly different config shapes. That is exactly how backend policy can see root FP8 metadata while a stage loader sees identity preprocessing after being handed a stripped talker config.

SGLang's own loader has one root-oriented contract: read `model_config.hf_config.quantization_config`, then fall back to text config / compression config. SGLang-Omni should mirror that with one narrow helper and pass the same root config through the stage load paths instead of maintaining three variants.

sglang_omni/quantization/methods/autoround.py
```diff
@@ -0,0 +1,157 @@
...
+            block_list = []
+
+        if not block_list:
+            return
+
+        normalized_blocks = [
+            entry[len(stage_prefix) :] if entry.startswith(stage_prefix) else entry
+            for entry in block_list
+        ]
+        if normalized_blocks == block_list:
```
Ratish1: Lets normalize AutoRound `extra_config` independently from `block_name_to_quantize`.

This early return means a config with already-normalized block names but still-prefixed `extra_config` regex keys will skip regex normalization entirely. SGLang AutoRound uses `layer_name.startswith(...)` for `block_name_to_quantize` and `re.fullmatch(...)` for `extra_config`, so both fields have to match the split stage module names.

Also, the current regex rewrite uses a global `replace()` on the whole pattern. That can remove `thinker.` / `thinker\\.` from places that are not the stage prefix. normalize block names and extra regex keys as separate changes, then write back if either changed.

sglang_omni/quantization/registry.py
```diff
@@ -0,0 +1,187 @@
...
+        Safe to call repeatedly: a guard short-circuits after the first run.
+        """
+        if cls._initialized:
+            return
+
+        package = "sglang_omni.quantization.methods"
+        for module_name in _BUILTIN_METHOD_MODULES:
+            try:
+                importlib.import_module(f".{module_name}", package=package)
+            except ImportError as e:
```
Ratish1: should we fail fast for built-in quantization method import failures?

These modules are shipped built-ins, not optional third-party plugins. Catching `ImportError` here and then setting `_initialized = True` can permanently hide a broken `fp8` or `autoround` import and later degrade into "unknown quantization method" or identity preprocessing.

tests/test_model/test_qwen3_omni_quantization_fp8.py
Ratish1, line 1: we already have a Qwen3 FP8 GPU CI stage that runs the Video-AMME Talker path.  this is just redundant, if there is anything extra to cover, we can just add them as CPU unit tests.

Ratish1 (commented):
Thanks for addressing the previous comments @lvliang-intel . I have left some more comments. Also, lets update sglang-omni/tests/README.md as well.

sglang_omni/quantization/methods/autoround.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def _normalize_extra_config_keys(
+        self, quant_config: dict[str, Any], stage_prefix: str
+    ) -> bool:
+        """Strip ``stage_prefix`` from the leading edge of every regex key."""
+        extra_config = quant_config.get("extra_config")
+        if not (isinstance(extra_config, dict) and extra_config):
+            return False
+
+        escaped_prefix = stage_prefix.replace(".", r"\.")
```
Ratish1: The `extra_config` prefix normalization still misses the public AutoRound checkpoint shape.

The current Intel checkpoint has:

```json
".*thinker\\.model\\.layers\\.\\d+\\.mlp\\.gate.*"
```

the code only strips `thinker.` / `thinker\.` when the pattern starts directly with that prefix, so the public pattern stays unchanged after thinker stage normalization. SGLang's `AutoRoundConfig.get_layer_config()` runs regex matching against stage-local names such as `model.layers...`, so this normalized config no longer matches the same module names the checkpoint config was written for.

Lets handle the real exported regex form as part of the stage rewrite, and add a unit test using the current Intel config pattern. A narrow helper for leading regex wildcards, e.g. `.*thinker\.` -> `.*`, is enough; this does not need a general regex parser.

sglang_omni/model_runner/model_worker.py
```diff
@@ -556,6 +556,72 @@ def _is_fp8_cutlass_moe_supported() -> bool:
...
+    quant_dict = extract_quantization_config(hf_config)
+    if quant_dict is None:
+        return None, None
+
+    # ``QuantizationConfig.from_checkpoint_config`` expects the
+    # ``{"quantization_config": {...}}`` wrapper shape, so re-wrap the
+    # already-extracted payload.
+    config_dict = {"quantization_config": quant_dict}
+
+    quant_config = QuantizationConfig.from_checkpoint_config(config_dict)
```
Ratish1: This detection path creates a `QuantizationConfig`, mutates `quant_config.method`, returns it, and then the only production caller discards it and looks the method up again by name.

That is more state than the startup path needs. lets return the detected method instance from `_detect_quantization_method()` and call `method.configure(...)` directly, or return only the method name if the instance is intentionally not reused. Right now the parsed `QuantizationConfig` mostly exists so tests can assert it, not because production needs it.

The simpler shape is easier to audit: extract checkpoint quant dict -> registry detects method -> configure that method -> model_worker backend policy runs.

tests/unit_test/quantization/test_registry.py
```diff
@@ -0,0 +1,341 @@
...
+        pass
+
+
+class TestQuantizationRegistry:
+    """Tests for QuantizationRegistry."""
+
+    def test_register_method(self) -> None:
+        """Test registering a quantization method."""
+        # Clear any existing registration
+        QuantizationRegistry._methods.clear()
```
Ratish1: These registry tests mutate `QuantizationRegistry._methods` and `_initialized` directly but do not snapshot/restore them around each test.

That is fragile because built-in methods register via module import side effects. Once `sglang_omni.quantization.methods.fp8` or `autoround` is already imported, clearing `_methods` and then calling `auto_register_all()` can import cached modules without re-running the decorators. Some tests manually re-register the class they need, but the global registry state still leaks to later tests.

lets add an autouse fixture in this file that snapshots `_methods` and `_initialized` before each test and restores them in `finally`. Tests that need a blank registry can still clear it inside the fixture boundary.

tests/test_model/test_qwen3_omni_quantization_autoround.py
Ratish1, line 1: Should we remove this as it is covered by tests/unit_test/quantization/test_integration.py?

Ratish1 (commented):
Can we also update the PR description and test after addressing these comments below @lvliang-intel . Great job so far.

Ratish1:
Hey @lvliang-intel , further looking at this PR I think a refactor is needed. Lets bring it down to the actual SGLang-Omni compatibility layer instead of introducing a second quantization framework.

I checked against the pinned local SGLang v0.5.12.post1: SGLang already registers auto-round and fp8, parses checkpoint quantization_config, builds the real AutoRoundConfig / Fp8Config, validates dtype/ device support, constructs quantized layers, and runs post-load quantization hooks. So SGLang-Omni should not own a parallel QuantizationConfig, generic registry, or broad method abstraction here. What do you think about this?

The Omni-specific work appears to be only from what I see:

  1. Normalize Qwen3-Omni composite checkpoint names for stage-local SGLang modules, e.g. thinker.model.layers / .*thinker\\.model... -> model.layers / .*model... before SGLang builds AutoRoundConfig.
  2. Keep the FP8 weight_scale_inv reciprocal conversion for Qwen3-Omni custom weight loaders before calling each parameter’s SGLang weight_loader.
  3. Keep Qwen3-specific FP8 backend policy in model_worker, where it already has the architecture/stage context.

 So we should consider trimming the PR around those contracts. A small private helper module with fixed dispatch for AutoRound stage normalization and FP8 weight preprocessing would be much better imo. Feel free to let me know your thoughts. Also, I can also help with the testing of the PR after you refactor it. Thanks

lvliang-intel:
> Hey @lvliang-intel , further looking at this PR I think a refactor is needed. Lets bring it down to the actual SGLang-Omni compatibility layer instead of introducing a second quantization framework.
> 
> I checked against the pinned local SGLang v0.5.12.post1: SGLang already registers auto-round and fp8, parses checkpoint quantization_config, builds the real AutoRoundConfig / Fp8Config, validates dtype/ device support, constructs quantized layers, and runs post-load quantization hooks. So SGLang-Omni should not own a parallel QuantizationConfig, generic registry, or broad method abstraction here. What do you think about this?
> 
> The Omni-specific work appears to be only from what I see:
> 
> 1. Normalize Qwen3-Omni composite checkpoint names for stage-local SGLang modules, e.g. thinker.model.layers / .*thinker\.model... -> model.layers / .*model... before SGLang builds AutoRoundConfig.
> 2. Keep the FP8 weight_scale_inv reciprocal conversion for Qwen3-Omni custom weight loaders before calling each parameter’s SGLang weight_loader.
> 3. Keep Qwen3-specific FP8 backend policy in model_worker, where it already has the architecture/stage context.
> 
> So we should consider trimming the PR around those contracts. A small private helper module with fixed dispatch for AutoRound stage normalization and FP8 weight preprocessing would be much better imo. Feel free to let me know your thoughts. Also, I can also help with the testing of the PR after you refactor it. Thanks

Thanks for the thorough review. This was really helpful. I think I agree with the overall direction.
On the implementation points, I think you're right and I'll simplify accordingly:
1 Remove the QuantizationConfig dataclass since SGLang already owns config parsing and layer construction.
2 Replace the decorator-based registry with a simple dispatch helper.
3 Drop the ABC, since it's introducing unnecessary indirection (e.g. no-op methods like configure() for FP8).

Just to share a bit of context behind the original design: the motivation wasn't to introduce a second quantization framework, but to avoid duplicating the small amount of Omni-specific preprocessing in every model directory.
The existing FP8 support lives inside each model implementation, so adding another quantization method (like AutoRound) would require touching multiple model files and need to do it for every supported model. If we later add more methods (e.g. AWQ), we'd end up repeating the same integration pattern again. That's why I experimented with a thin compatibility layer. The idea was to follow a pattern similar to vLLM-Omni how vLLM-Omni sits on top of vLLM's native quantization support: SGLang would still own quantization discovery, config parsing, layer construction, backend selection, and execution, while Omni would only centralize the small amount of model-specific preprocessing (such as stage-name normalization and FP8 weight preprocessing).  If the goal had only been to support AutoRound, I would have just followed the existing FP8 pattern, since that would have been the smallest change. 

That said, I agree the current abstraction is heavier than necessary. Based on your feedback, I'll trim the PR down so the Omni side only contains the minimal compatibility logic you suggested:
1 AutoRound stage-local checkpoint name normalization before SGLang builds AutoRoundConfig.
2 FP8 weight_scale_inv reciprocal conversion before invoking each parameter's SGLang weight_loader.
3 Keep the existing Qwen3-specific FP8 backend policy in model_worker.

Everything else (quantization detection, config parsing, layer construction, backend selection, post-load hooks, etc.) will remain owned by SGLang.

I'll work on the refactor over the next day. Thanks again for the detailed feedback, and I'd really appreciate another look once I push the updated version.

lvliang-intel:
## New test data based on latest code


### MMMU Accuracy Results

| Hardware | Precision | Max Concurrency | Total Samples | Correct | Accuracy | Failed Requests | MC Parse Fallback |
|----------|-----------|------------------|----------------|---------|----------|------------------|--------------------|
| A100     | INT4      | 32               | 50             | 32      | 64.0%    | 0                | 0                  |
| A100     | INT4      | 16               | 50             | 34      | 68.0%    | 0                | 1                  |
| A100     | INT4      | 8                | 50             | 32      | 64.0%    | 0                | 0                  |
| A100     | BF16      | 4                | 50             | 28      | 56.0%    | 0                | 3                  |
| A100     | BF16      | 2                | 50             | 32      | 64.0%    | 0                | 1                  |

### MMMU Speed Results

| Hardware | Precision | Max Concurrency | Completed | Failed | Latency Mean (s) | Latency Median (s) | Latency P95 (s) | Latency P99 (s) | Output Throughput (tok/s) | Output Tokens/Request-s | Output Tokens Mean | Prompt Tokens Mean | Throughput (req/s) |
|----------|-----------|------------------|-----------|--------|-------------------|----------------------|-------------------|-------------------|------------------------------|----------------------------|----------------------|----------------------|------------------------|
| A100     | INT4      | 32               | 50        | 0      | 34.914            | 28.921               | 81.611            | 83.55             | 359.9                        | 17.3                       | 604                  | 1307                 | 0.595                  |
| A100     | INT4      | 16               | 50        | 0      | 22.851            | 12.339               | 74.861            | 74.962             | 367.7                        | 27.1                       | 620                  | 1307                 | 0.593                  |
| A100     | INT4      | 8                | 50        | 0      | 16.747            | 9.983                | 53.235            | 54.775             | 275.6                        | 37.2                       | 623                  | 1307                 | 0.443                  |
| A100     | BF16      | 4                | 50        | 0      | 40.66             | 24.473               | 106.341           | 107.99             | 66.5                         | 17.1                       | 695                  | 1307                 | 0.096                  |
| A100     | BF16      | 2                | 50        | 0      | 36.827            | 17.712               | 114.661           | 115.961           | 34.0                         | 17.2                       | 634                  | 1307                 | 0.054                  |

lvliang-intel:
@Ratish1, could you please help to review again?

sglang_omni/quantization/__init__.py
Ratish1, line 1: Can we move this out of a top-level `sglang_omni.quantization` package?. what are your thoughts on this, since we seem to support Qwen only.

sglang_omni/quantization/__init__.py
```diff
@@ -0,0 +1,313 @@
...
+    "resolve_quant_config",
+    "get_weight_preprocessor",
+    "needs_quant_config_normalization",
+    "normalize_quant_config",
+]
+
+
+# --------------------------------------------------------------------------- #
+# quantization_config discovery
+# --------------------------------------------------------------------------- #
```
Ratish1: remove these box comments only

docs/basic_usage/qwen3_omni.md
```diff
@@ -268,6 +268,29 @@ SGLANG_JIT_DEEPGEMM_PRECOMPILE=1 sgl-omni serve \
...
+  --model-name qwen3-omni \
+  --model-path Intel/Qwen3-Omni-30B-A3B-Instruct-int4-AutoRound \
+  --port 8008
+```
+
+AutoRound quantization provides:
+- **~50% memory reduction** compared to BF16 (from ~60GB to ~30GB)
+- **~25% memory reduction** compared to FP8 (from ~40GB to ~30GB)
+- **Automatic weight preprocessing** via the unified quantization abstraction
+
```
Ratish1, line 295: wording should be updated here

tests/README.md
Ratish1: remove the `test_quantization.py` since we seem to remove the file from the unit tests

sglang_omni/quantization/__init__.py
lvliang-intel: Thanks. I agree the single-file layout makes it read as Qwen-specific, fair point.
That said, I'd like to keep sglang_omni.quantization at the top level. Because it's shared infrastructure used by `model_runner/model_worker.py`, alongside `model_runner/`, `scheduling/`, and `sampling/`, and it exposes unified entry points any model loader can use. Moving it under a model-specific directory wouldn't remove the duplication concern.
To address the layout concern directly, I'd suggest splitting it into a thin `__init__.py` plus a sibling `quantization.py` holding the actual logic. What do you think?  If you'd prefer a different location given the constraint above, let me know which way you'd like and I'll go with that.

Ratish1: I slightly prefer `sglang_omni/quantization.py` over a `sglang_omni/quantization/` package.

lvliang-intel: OK. Code moved to sglang_omni/quantization.py as you suggested.

hshen14:
Hi @Ratish1 Thanks for your kind review. I saw you tried simplifying qwen3-omni quantization layer to focus on INT4 support. I think the original design that @lvliang-intel proposed was to try making the quantization layer flexible to support other quantization schemes e.g., MXFP4, NVFP4. May I know whether you expect this PR would focus on INT4? What's your suggestion on the support of other quantization schemes? Thanks.

Ratish1:
> Hi @Ratish1 Thanks for your kind review. I saw you tried simplifying qwen3-omni quantization layer to focus on INT4 support. I think the original design that @lvliang-intel proposed was to try making the quantization layer flexible to support other quantization schemes e.g., MXFP4, NVFP4. May I know whether you expect this PR would focus on INT4? What's your suggestion on the support of other quantization schemes? Thanks.

 Hi @hshen14, thanks for asking. My intent for this PR now is to focus on adding and validating AutoRound W4A16 support for the Qwen3-Omni checkpoint while preserving the existing FP8 behavior. The simplification was not meant to make the compatibility layer permanently INT4-only. It was because AutoRound seemed to be the only tested checkpoint originally in this PR. MXFP4 and NVFP4 were motivations for the abstraction, but their Qwen3-Omni checkpoints and integration requirements were not tested here so I thought we could focus on the AutoRound W4A16 for now and later add focused support for the other quant formats.

I’m fully open to supporting MXFP4, NVFP4, and other formats. My suggestion would be to handle them through focused follow-ups that identify the exact checkpoint, quantized stages, weight-layout differences, hardware requirements, and testing. SGLang already owns the general quantization machinery, including registration, config parsing, quantized layers, weight loading, and post-load processing, so SGLang-Omni should add only the model-specific compatibility needed for each format. If several validated formats end up requiring the same Omni-specific behavior, we can then introduce a shared abstraction around those concrete requirements. This is also the direction @lvliang-intel and I discussed above in several of the reviews I made. Could you let me know what you think about this, I’d be glad to align on the next formats or further discussion over SGLang Slack (Ratish P) (https://slack.sglang.io/), or feel free to share your WeChat ID if that is easier. Thanks.

Ratish1:
Hey @lvliang-intel , I also wanted to mention that when I was testing AutoRound W4A16 validation on all benchmarks, the video benchmarks also exposed an existing async decode bug unrelated to the quantization loader. Stale lookahead filtering called SGLang’s `ScheduleBatch.filter_batch()`, which cleared `out_cache_loc` even though the retained extend requests still owned those KV-cache slots. The next forward therefore passed None as the KV-store indices, eading to an illegal CUDA memory access and thinker process abort. I opened that fix separately here https://github.com/sgl-project/sglang-omni/pull/1010.

hshen14:
> > Hi @Ratish1 Thanks for your kind review. I saw you tried simplifying qwen3-omni quantization layer to focus on INT4 support. I think the original design that @lvliang-intel proposed was to try making the quantization layer flexible to support other quantization schemes e.g., MXFP4, NVFP4. May I know whether you expect this PR would focus on INT4? What's your suggestion on the support of other quantization schemes? Thanks.
> 
> Hi @hshen14, thanks for asking. My intent for this PR now is to focus on adding and validating AutoRound W4A16 support for the Qwen3-Omni checkpoint while preserving the existing FP8 behavior. The simplification was not meant to make the compatibility layer permanently INT4-only. It was because AutoRound seemed to be the only tested checkpoint originally in this PR. MXFP4 and NVFP4 were motivations for the abstraction, but their Qwen3-Omni checkpoints and integration requirements were not tested here so I thought we could focus on the AutoRound W4A16 for now and later add focused support for the other quant formats.
> 
> I’m fully open to supporting MXFP4, NVFP4, and other formats. My suggestion would be to handle them through focused follow-ups that identify the exact checkpoint, quantized stages, weight-layout differences, hardware requirements, and testing. SGLang already owns the general quantization machinery, including registration, config parsing, quantized layers, weight loading, and post-load processing, so SGLang-Omni should add only the model-specific compatibility needed for each format. If several validated formats end up requiring the same Omni-specific behavior, we can then introduce a shared abstraction around those concrete requirements. This is also the direction @lvliang-intel and I discussed above in several of the reviews I made. Could you let me know what you think about this, I’d be glad to align on the next formats or further discussion over SGLang Slack (Ratish P) (https://slack.sglang.io/), or feel free to share your WeChat ID if that is easier. Thanks.

Hi @Ratish1 Thanks for the explanation. Agreed to focus on W4A16 in this PR. Could you please help review this PR and merge if this look good? Will submit a RFC on the support of more quantization schemes.

Ratish1:
> > > Hi @Ratish1 Thanks for your kind review. I saw you tried simplifying qwen3-omni quantization layer to focus on INT4 support. I think the original design that @lvliang-intel proposed was to try making the quantization layer flexible to support other quantization schemes e.g., MXFP4, NVFP4. May I know whether you expect this PR would focus on INT4? What's your suggestion on the support of other quantization schemes? Thanks.
> > 
> > 
> > Hi @hshen14, thanks for asking. My intent for this PR now is to focus on adding and validating AutoRound W4A16 support for the Qwen3-Omni checkpoint while preserving the existing FP8 behavior. The simplification was not meant to make the compatibility layer permanently INT4-only. It was because AutoRound seemed to be the only tested checkpoint originally in this PR. MXFP4 and NVFP4 were motivations for the abstraction, but their Qwen3-Omni checkpoints and integration requirements were not tested here so I thought we could focus on the AutoRound W4A16 for now and later add focused support for the other quant formats.
> > I’m fully open to supporting MXFP4, NVFP4, and other formats. My suggestion would be to handle them through focused follow-ups that identify the exact checkpoint, quantized stages, weight-layout differences, hardware requirements, and testing. SGLang already owns the general quantization machinery, including registration, config parsing, quantized layers, weight loading, and post-load processing, so SGLang-Omni should add only the model-specific compatibility needed for each format. If several validated formats end up requiring the same Omni-specific behavior, we can then introduce a shared abstraction around those concrete requirements. This is also the direction @lvliang-intel and I discussed above in several of the reviews I made. Could you let me know what you think about this, I’d be glad to align on the next formats or further discussion over SGLang Slack (Ratish P) (https://slack.sglang.io/), or feel free to share your WeChat ID if that is easier. Thanks.
> 
> Hi @Ratish1 Thanks for the explanation. Agreed to focus on W4A16 in this PR. Could you please help review this PR and merge if this look good? Will submit a RFC on the support of more quantization schemes.

Thanks @hshen14  for the discussion. I will review it further and see if more changes are needed or else we can move forward with this PR asap.

Ratish1:
Hey @lvliang-intel, there seems to be a small conflict, could you fix it. Thanks

lvliang-intel:
> Hey @lvliang-intel, there seems to be a small conflict, could you fix it. Thanks

Conflict fixed.

Ratish1 (approved):

PR #716, opened 2026-06-07
[Model] Add MOSS-TTS Local support

draftbk:
Adds **MOSS-TTS Local** ([OpenMOSS-Team/MOSS-TTS-Local-Transformer](https://huggingface.co/OpenMOSS-Team/MOSS-TTS-Local-Transformer)) — 1.7B depth-transformer TTS as a self-contained pipeline.

## Validation (full Seed-TTS-eval, shipped config)
Measured here with Qwen3-ASR (WER/CER) and a WavLM speaker-verification model (SIM); single full-set run, stochastic sampling (±~0.1pp run-to-run). Upstream numbers are from the [model card](https://huggingface.co/OpenMOSS-Team/MOSS-TTS-Local-Transformer).

| Metric | Local (here) | Upstream (model card) |
|--------|------|------|
| EN WER | 1.96 | 1.93 |
| ZH CER | 1.25 | 1.44 |
| EN SIM | 70.3 | 73.28 |
| ZH SIM | 76.4 | 79.62 |

Comparability caveat: we don't know which ASR/SV models the card used, so the two columns may not be directly comparable (e.g. the lower ZH CER here could just be an ASR difference, not better synthesis). SIM sits ~3 pts under the card on both languages. We'd guess this is a speaker-verification-model offset rather than a regression — on this WavLM the merged MOSS-TTS Delay scores 67.1 EN (80-sample subset) vs the card's 70.86, a similar-sized gap, and Local (70.3, full set) lands above that Delay run — but we can't confirm that without the card's exact SV model, and Local/Delay weren't scored on identical samples.

Perf: single-stream (concurrency 1) RTF ~2.7.

## TODO
- Per-request sampling overrides (currently uses upstream `generate()` defaults)
- CUDA-graph the depth predictor (runs eager today)
- Streaming output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #728, opened 2026-06-10
[Model] Add MOSS-TTS-Local-Transformer-v1.5 support

GaokaiZhang:
Adds native serving for [OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5](https://huggingface.co/OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5) (#637, TODO item 1: model support with radix cache).

v1.5 is not a delay-pattern model: the 36-layer Qwen3 backbone consumes one summed embedding per 12.5 Hz frame (text channel + 12 RVQ channels), and a 1-layer frame-local transformer decodes each frame — a binary continue/stop head, then 12 sequentially sampled codebooks, each fed back as the next local position. The codec is MOSS-Audio-Tokenizer-v2 (~2B params, RVQ-12, native 48 kHz stereo).

## Implementation

- 3-stage pipeline (preprocessing → AR engine → vocoder) following the MOSS Delay layout from #609. Serving by model id resolves directly: the v1.5 checkpoint declares `architectures: ["MossTTSLocalModel"]`.
- Radix cache: multi-channel prompt rows hash to surrogate token ids (the MOSS Delay scheme), so repeated reference audio hits the prefix cache; prefill skips `prefix_indices`.
- Backbone decode runs under SGLang CUDA graphs (next-frame embeddings staged through a fixed table; decode input_ids are row indices).
- The per-frame local decode (1+12 micro-steps plus 13 seeded sampling passes) is also CUDA-graph captured per batch-size bucket. Eager it is launch-bound at ~22 ms/frame; replay takes ~4 ms. Falls back to eager when `audio_repetition_penalty != 1`.
- Reference encodes coalesce into batched codec calls (a single encode costs ~0.25 GPU-seconds through the ~1B-param encoder; concurrent requests share one `batch_encode`). Codec stages auto-place on the second visible GPU, falling back to GPU 0 on single-GPU hosts. The preprocessing and vocoder stages keep separate codec instances because the codec's chunked decode flips module-global streaming state.
- Output stays stereo end to end: `[2, N]` 48 kHz payloads and multi-channel WAV/PCM encoding in the serve path (mono models unaffected).
- Sampling is seeded per (request, step, channel) via `multinomial_with_seed`, so `seed` reproduces at any concurrency. Defaults match the model card: text 1.0/1.0/50, audio 1.7/0.8/25, repetition penalty 1.0.

## Validation (full Seed-TTS-eval, EN)

2×H100 80GB (GPU 0: AR engine, GPU 1: codec encode + vocoder), DP 1, no router, served by model id with stock defaults; ASR Qwen/Qwen3-ASR-1.7B.

|---|---|---|---|---|---|---|
| MOSS-TTS-Local-v1.5, c=16, token-count auto, 2×H100 | EN | WER 1.86% raw / 1.65% excl. >50% outliers | 1088/1088 | 0 / 0 | 7.78 req/s | 0.4878 |
| MOSS-TTS-Local-v1.5, c=16, token-count auto, 2×H100 | ZH | CER 1.30% raw / 1.17% excl. >50% outliers | 2020/2020 | 0 / 0 | 8.62 req/s | 0.3306 |

EN rtf p99 0.999, mean latency 2.05 s. For comparison, the MOSS Delay EN reference in the benchmark header is WER 1.93% / rtf 0.913 (H100, c=16). 2×H100 is one DP-1 replica with the AR engine on GPU 0 and the codec stages on GPU 1; single-GPU serving works but the codec encoder contends with the decode loop.

Run-to-run note: a repeat EN run measured rtf_mean 0.53 / WER 2.66% raw, 1.86% excl. — the raw WER swings with the count of >50%-WER runaway samples (3 vs 11 of 1088 across runs; same tail behavior is visible in the Delay reference and the v1.0 numbers in #716), while the excluded-outlier metric stays within ASR noise.

Repro:

```bash
CUDA_VISIBLE_DEVICES=0,1 sgl-omni serve --model-path OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5 --port 8021
python -m benchmarks.eval.benchmark_tts_seedtts --meta zhaochenyang20/seed-tts-eval-arrow --lang en \
  --model OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5 --port 8021 \
  --ref-format references --token-count auto --max-concurrency 16 --output-dir results/moss_local_en
```

20 CPU unit tests under `tests/unit_test/moss_tts_local/`: local-transformer incremental-vs-recompute equivalence (incl. interleaved RoPE), graph/eager sampler parity, registry resolution, prompt/state builders, repetition-penalty semantics, stereo payload + WAV encoding.

Relation to #716: that PR targets the v1.0 depth-transformer checkpoint, whose config declares the Delay architecture (explicit config file required). v1.5 ships its own `moss_tts_local` model type, so this path is independent; both use the `models/moss_tts_local` module name, so whichever lands second needs a rebase.

Follow-ups: streaming output, async decode (row-indexed state pool), ZH split numbers.

### DIFF
--- a/sglang_omni/client/client.py
+++ b/sglang_omni/client/client.py
@@ -195,7 +195,10 @@ class Client:
         if len(audio_chunks) == 1:
             audio_data = audio_chunks[0]
         else:
-            audio_data = np.concatenate([to_numpy(c) for c in audio_chunks])
+            arrays = [to_numpy(c) for c in audio_chunks]
+            # Rank-2 chunks are [channels, samples]: concatenate on time.
+            axis = -1 if arrays[0].ndim > 1 else 0
+            audio_data = np.concatenate(arrays, axis=axis)
 
         encode_kwargs: dict[str, Any] = {
             "response_format": response_format,
--- a/sglang_omni/client/client.py
+++ b/sglang_omni/client/client.py
@@ -208,7 +208,6 @@ class Client:
             audio_data = audio_chunks[0]
         else:
             arrays = [to_numpy(c) for c in audio_chunks]
-            # Rank-2 chunks are [channels, samples]: concatenate on time.
             axis = -1 if arrays[0].ndim > 1 else 0
             audio_data = np.concatenate(arrays, axis=axis)
 
--- a/sglang_omni/client/client.py
+++ b/sglang_omni/client/client.py
@@ -10,6 +10,7 @@ from typing import Any, AsyncIterator, Callable
 import numpy as np
 
 from sglang_omni.client.audio import (
+    DEFAULT_SAMPLE_RATE,
     FORMAT_MIME_TYPES,
     audio_to_base64,
     encode_audio,
@@ -87,6 +88,7 @@ class Client:
         """
         text_parts: list[str] = []
         audio_chunks: list[Any] = []
+        sample_rate: int | None = None
         last_chunk: GenerateChunk | None = None
         finish_reason: str | None = None
 
@@ -96,6 +98,8 @@ class Client:
                 text_parts.append(chunk.text)
             if chunk.audio_data is not None:
                 audio_chunks.append(chunk.audio_data)
+            if chunk.sample_rate is not None:
+                sample_rate = chunk.sample_rate
             if chunk.finish_reason is not None:
                 finish_reason = chunk.finish_reason
 
@@ -109,8 +113,14 @@ class Client:
             if len(audio_chunks) == 1:
                 combined = audio_chunks[0]
             else:
-                combined = np.concatenate([to_numpy(c) for c in audio_chunks])
-            audio_b64 = audio_to_base64(combined, output_format=audio_format)
+                arrays = [to_numpy(c) for c in audio_chunks]
+                axis = -1 if arrays[0].ndim > 1 else 0
+                combined = np.concatenate(arrays, axis=axis)
+            audio_b64 = audio_to_base64(
+                combined,
+                sample_rate=sample_rate or DEFAULT_SAMPLE_RATE,
+                output_format=audio_format,
+            )
             audio = CompletionAudio(
                 id=f"audio-{request_id}",
                 data=audio_b64,
@@ -145,7 +155,9 @@ class Client:
             audio_b64: str | None = None
             if chunk.modality == "audio" and chunk.audio_data is not None:
                 audio_b64 = audio_to_base64(
-                    chunk.audio_data, output_format=audio_format
+                    chunk.audio_data,
+                    sample_rate=chunk.sample_rate or DEFAULT_SAMPLE_RATE,
+                    output_format=audio_format,
                 )
 
             yield CompletionStreamChunk(
--- a/sglang_omni/models/moss_tts_local/model_runner.py
+++ b/sglang_omni/models/moss_tts_local/model_runner.py
@@ -154,7 +154,9 @@ class MossTTSLocalModelRunner(ModelRunner):
             return
         hidden_states = getattr(result.logits_output, "hidden_states", None)
         if not isinstance(hidden_states, torch.Tensor):
-            raise RuntimeError("MOSS-TTS Local model output did not include hidden states")
+            raise RuntimeError(
+                "MOSS-TTS Local model output did not include hidden states"
+            )
         if hidden_states.ndim == 3:
             hidden_states = hidden_states[:, -1, :]
 
@@ -219,9 +221,8 @@ class MossTTSLocalModelRunner(ModelRunner):
                 positions=gen_steps * num_channels + channel + 1,
             )
 
-        use_graph = (
-            rep_histories is None
-            and batch_size <= getattr(self.model, "frame_graph_max_bs", 0)
+        use_graph = rep_histories is None and batch_size <= getattr(
+            self.model, "frame_graph_max_bs", 0
         )
         if use_graph:
             stop_choice, codes = self.model.decode_frame_graphed(
@@ -250,9 +251,7 @@ class MossTTSLocalModelRunner(ModelRunner):
             torch.full((batch_size,), end_id, dtype=torch.long, device=device),
         )
 
-        rows = torch.empty(
-            (batch_size, num_channels), dtype=torch.long, device=device
-        )
+        rows = torch.empty((batch_size, num_channels), dtype=torch.long, device=device)
         rows[:, 0] = next_text
         rows[:, 1:] = codes
 
--- a/sglang_omni/models/moss_tts_local/model_runner.py
+++ b/sglang_omni/models/moss_tts_local/model_runner.py
@@ -1,0 +1,335 @@
+# SPDX-License-Identifier: Apache-2.0
+"""MOSS-TTS Local (v1.5) model runner for OmniScheduler."""
+
+from __future__ import annotations
+
+from typing import Any
+
+import torch
+
+from sglang_omni.model_runner.base import ModelRunner
+from sglang_omni.models.moss_tts.model_runner import MossTTSModelRunner
+from sglang_omni.scheduling.types import RequestOutput
+
+
+class MossTTSLocalModelRunner(ModelRunner):
+    """Drives the per-frame local-transformer decode and feedback embeddings.
+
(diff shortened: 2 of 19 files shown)


GaokaiZhang:
Follow-up commits since the initial revision:

- `71c7db6` graphs the frame-decode feedback path and hardens the codec stages. The next-step feedback embedding (assistant-slot + 12 code embeddings) is now computed inside the captured frame graph instead of a separate eager pass, per-batch sampling parameter tensors are cached across steps, and the per-request GPU clones in output collection became two per-step snapshots. Hardening from review: the local-transformer KV buffers are frozen after graph capture (a later reallocation would have left the graphs reading freed memory), reference encodes get a 120 s result timeout and a 100 s duration cap (mirroring the Higgs limit), an empty generation no longer fails its vocoder batch peers, and the streaming / raw-PCM / compressed paths downmix the 48 kHz stereo payload instead of dropping a channel (WAV keeps both channels).
- `cc681f0` adds tests: CUDA-graph vs eager frame-decode parity (GPU-gated), repetition-penalty history gathering, generation-kwargs precedence, and coalescer error isolation.

Performance effect: engine-side decode improved slightly (100.6 vs 98.2 frames/s per request in-engine on the full EN set), but end-to-end rtf_mean is unchanged within run-to-run noise (0.53 vs 0.49 mean, 0.48 vs 0.46 median on repeat full-set runs sharing the host with other workloads). The table numbers in the description stand; the run-to-run note covers the spread.

Remaining follow-ups for the rtf 0.25 @ c=16 target: the engine itself is at ~0.13 rtf (≈10 ms/frame at bs16), so the gap is prefill, reference encode, and vocoder latency plus queueing — async decode with a row-indexed state pool and the encoder/preprocessing fusion items on the #637 roadmap are the natural next steps.

0xjeffro:
Hey Team, here is reproduce result based on 4dc7ca3 (Handle retraction resume in prefill embeds), use the same command above, running on H100 SXM * 2, also attach the CN results in case it's useful:


|---|---|---|---|---|---|---|
| MOSS-TTS-Local-v1.5, c=16, non-streaming, 2×H100 | EN | WER 2.29% raw / 1.80% excl. >50% outliers | 1088/1088 | 0 / 0 | 6.67 req/s | 0.5703 |
| MOSS-TTS-Local-v1.5, c=16, non-streaming, 2×H100 | ZH | CER 1.35% raw / 1.28% excl. >50% outliers | 2020/2020 | 0 / 0 | 7.13 req/s | 0.3939 |

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -0,0 +1,384 @@
...
+            stop_choice == 0,
+            torch.full((batch_size,), slot_id, dtype=torch.long, device=device),
+            torch.full((batch_size,), end_id, dtype=torch.long, device=device),
+        )
+
+        rows = torch.empty((batch_size, num_channels), dtype=torch.long, device=device)
+        rows[:, 0] = next_text
+        rows[:, 1:] = codes
+
+        result.next_token_ids = next_text
```
gaoyang07: Should we use the full 13-channel row ids for radix key instead of text only to prevent prefix matching when there is a possible re-prefilling?

sglang_omni/client/client.py
zhaochenyang20: the MOSS Local vocoder now writes the real `sample_rate` into the payload at `sglang_omni/models/moss_tts_local/stages.py:409`, and `GenerateChunk` carries a `sample_rate` field, but non-streaming chat completion aggregates audio and calls `audio_to_base64(combined, output_format=audio_format)` without passing that field. `audio_to_base64()` defaults to `DEFAULT_SAMPLE_RATE=24000`, so a `/v1/chat/completions` request with `modalities: ["audio"]` and 48 kHz stereo output gets a WAV header that says 24 kHz; I verified the same helper encodes a `(2, 4800)` stereo array with a 24000 header.

The same issue exists in the streaming chat path at `sglang_omni/client/client.py:147`, where each audio chunk is passed directly to `audio_to_base64(chunk.audio_data, output_format=audio_format)`. The speech route now uses the chunk or payload sample rate in `Client.speech()` and `_speech_stream()`, so chat audio is the path that remains wrong. The user-visible consequence is not a minor metadata mismatch: playback speed and pitch are wrong, 48 kHz content is played as 24 kHz, duration doubles, pitch drops, and the API response no longer represents the model output.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -0,0 +1,452 @@
...
+    encode_batch_size: int = 8,
+    encode_batch_wait_ms: int = 4,
+) -> SimpleScheduler:
+    device = _resolve_codec_device(device)
+    processor = _load_moss_tts_local_processor(model_path, device=device)
+    reference_encoder = _BatchedReferenceEncoder(
+        processor,
+        max_batch_size=encode_batch_size,
+        max_batch_wait_ms=encode_batch_wait_ms,
+    )
```
zhaochenyang20, line 245: The default device selection splits the configured resource fact from the runtime fact. `MossTTSLocalPipelineConfig` declares preprocessing, `tts_engine`, and vocoder as the same `process="pipeline"` with `gpu=0`, so the existing placement planner records only GPU 0 for those stages. However, `create_preprocessing_executor()` and `create_vocoder_executor()` accept `device` rather than `gpu_id`, so runtime placement is not injected, and `_resolve_codec_device(None)` returns `cuda:1` on multi-GPU hosts. In practice default serving loads two codec instances on GPU 1 while topology, memory budgeting, and conflict validation still believe the pipeline only occupies GPU 0.

zhaochenyang20: Do not implicitly choose `cuda:1` on the default path. Make the codec device an explicit stage/runtime configuration that the placement planner can see; viable options include an auditable codec resource field, separate process/stage placement for codec work, or at minimum requiring users to set `factory_args.device` explicitly before using the second GPU and documenting that in the example config. Whichever approach is chosen, the resource plan, logs, and actual CUDA module placement need to agree.

zhaochenyang20 (changes requested):

sglang_omni/models/moss_tts_local/model_runner.py
GaokaiZhang: Checked — confirmed. The text channel emits the same assistant-slot id for every generated frame, so a re-prefill after KV-pressure retraction could falsely prefix-match another identical-prompt request's cached generated region. Fixed in 289170a: generated frames now use the same full 13-channel row hash as prompt rows for the radix key, folded below the special-token band (the scheduler finishes any request whose generated id crosses the vocab boundary via `Req._check_vocab_boundary_finish`); the stop decision keeps the raw `audio_end` id for eos detection.

Full EN set re-run at this commit — performance unchanged (best rtf of our three full-set runs):

|---|---|---|---|---|---|---|
| MOSS-TTS-Local-v1.5 @289170a, c=16, token-count auto, 2×H100 | EN | WER 2.21% raw / 1.79% excl. >50% outliers | 1088/1088 | 0 / 0 | 7.98 req/s | 0.4751 |

The merged Delay path keys generated frames on its text channel the same way — worth the same treatment in a follow-up.

zhaochenyang20:
# MOSS-TTS-Local-Transformer-v1.5 Results

Seed-TTS EN full set, non-streaming, `max-concurrency=16`, `token-count=auto`, `ref-format=references`.

## Performance

| Metric | Value |
| --- | ---: |
| Samples | 1088 / 1088 |
| Failed | 0 |
| Throughput | 10.355 req/s |
| RTF mean | 0.3682 |
| RTF median | 0.3437 |
| RTF p95 | 0.5636 |
| RTF p99 | 0.6520 |
| Latency mean | 1.538 s |
| Latency median | 1.506 s |
| Latency p95 | 1.989 s |
| Latency p99 | 2.560 s |
| Audio duration mean | 4.401 s |

## Correctness

| Metric | Value |
| --- | ---: |
| WER raw | 2.67% |
| WER excl. >50% outliers | 1.78% |
| >50% WER samples | 11 / 1088 |
| Skipped | 0 |

zhaochenyang20 (approved):

PR #729, opened 2026-06-10
[Higgs TTS] Warmup-captured CUDA graph for vocoder codec decode

### LINKED ISSUE #581: [Perf] Higgs TTS: CUDA graph capture for vocoder decode
## Motivation

The vocoder decode stage converts multi-codebook audio codes into a 24 kHz waveform through the DAC acoustic decoder — a chain of 5 RVQ embedding lookups, a linear projection, 5 ConvTranspose1d upsampling blocks (each with residual units), and a final Conv1d. Each `codec.decode` call launches ~30 small GPU kernels whose individual launch overhead dominates wall-clock time, especially for short-to-medium utterances where the kernels themselves are fast but memory-bound.

The entire decode path is a fixed, deterministic feedforward computation with no data-dependent control flow: embedding lookups, linear projections, transposed convolutions, ELU activations, and residual additions. This makes it an ideal target for CUDA graph capture, which records the full kernel sequence once and replays it in a single GPU dispatch, eliminating per-kernel launch overhead.

## Current Implementation

The vocoder stage in `stages.py` calls `codec.decode()` eagerly, launching the full kernel chain per request:

```python
def _vocode(payload: StagePayload) -> StagePayload:
    ...
    codes_TN = reverse_delay_pattern(delayed_LN)
    codes_TN = torch.where(codes_TN >= codec_vocab, torch.zeros_like(codes_TN), codes_TN)
    waveform = codec.decode(codes_TN)   # eager: ~30 kernel launches
    ...
```

`HiggsAudioCodec.decode()` reshapes and dispatches a single `B=1` call:

```python
def decode(self, codes_TN: torch.Tensor) -> torch.Tensor:
    codes_BNT = codes_TN.transpose(0, 1).unsqueeze(0).to(device=self.device, dtype=torch.long)
    return self.model.decode(codes_BNT).audio_values.squeeze(0).squeeze(0).cpu()
```

The underlying `HiggsAudioV2TokenizerModel.decode()` runs:

1. **RVQ decode** — 5 codebook embedding lookups + linear projections, residual-summed to `(B, 256, T)`
2. **fc2** — `Linear(256 → 256)` acoustic projection
3. **Acoustic decoder (DAC)** — 5 ConvTranspose1d blocks with strides `[8, 5, 4, 2, 3]` (product = 960), each followed by residual Conv1d units; final Conv1d `32 → 1` channel

All operations are CUDA-graph-safe: no CPU syncs, no `.item()` calls, no data-dependent branching, deterministic allocation pattern.

### 1. Add `CUDAGraphCodecDecoder` in `audio_codec.py`

A wrapper around `HiggsAudioCodec` that captures and replays CUDA graphs:

- **Static buffers**: pre-allocate a fixed input buffer `(B, N, T_bucket)` and hold a reference to the output tensor `(B, 1, T_bucket * 960)` produced during capture.
- **Capture**: for each bucket, run a warmup pass, then record via `torch.cuda.CUDAGraph`.
- **Replay**: pad input codes to the nearest bucket, `copy_()` into the static input buffer, `graph.replay()`, slice the output to original length.
- **Fallback**: if T exceeds the largest bucket, fall back to eager decode.

```python
class CUDAGraphCodecDecoder:
    def __init__(self, codec: HiggsAudioCodec, *, buckets: list[int], max_batch_size: int = 1):
        self._codec = codec
        self._graphs: dict[tuple[int, int], _CapturedGraph] = {}
        self._capture_all(buckets, max_batch_size)

    def decode(self, codes_TN: torch.Tensor) -> torch.Tensor:
        T = codes_TN.shape[0]
        bucket = self._nearest_bucket(T)
        if bucket is None:
            return self._codec.decode(codes_TN)     # fallback
        g = self._graphs[(1, bucket)]
        g.input_buf[:, :, :T] = codes_TN.T
        g.input_buf[:, :, T:] = 0                   # pad
        g.graph.replay()
        return g.output_ref[0, 0, : T * 960].clone()
```

### 2. Define sequence-length buckets

At the codec's 75 Hz frame rate, typical TTS outputs span 1–30 seconds:

| Bucket (frames) | Duration  | Output samples |
|:-----------------|:----------|:---------------|
| 128              | 1.7 s     | 122,880        |
| 256              | 3.4 s     | 245,760        |
| 512              | 6.8 s     | 491,520        |
| 1,024            | 13.7 s    | 983,040        |
| 2,048            | 27.3 s    | 1,966,080      |

With `max_batch_size` up to 4, total captured graphs = 5 buckets x 4 batch sizes = 20.

### 3. Wire into `create_vocoder_executor`

Add an `enable_cuda_graph` parameter (default `True`). When enabled, construct `CUDAGraphCodecDecoder` at stage init and use it in `_vocode`:

```python
def create_vocoder_executor(
    model_path: str,
    *,
    device: str = "cuda:0",
    dtype: str = "bfloat16",
    enable_cuda_graph: bool = True,
    max_batch_size: int = 4,
    max_batch_wait_ms: int = 2,
):
    decoder = (
        CUDAGraphCodecDecoder(codec, buckets=[128, 256, 512, 1024, 2048], max_batch_size=max_batch_size)
        if enable_cuda_graph
        else codec
    )
    ...
```

### 4. Compose with batched decode (#569)

This optimization is orthogonal to #569 (batched vocoder decode). The two compose naturally:

- **Without #569**: each request decodes individually; graphs are captured at `B=1` for each T bucket.
- **With #569**: `decode_batch` pads variable-length requests to a common T and decodes them together; graphs are captured at `(B, T_bucket)` pairs, and `CUDAGraphCodecDecoder.decode_batch()` selects the right captured graph.

Both can land independently; the bucket scheme extends to batched shapes without redesign.

## Scope

- `sglang_omni/models/higgs_tts/audio_codec.py` — add `CUDAGraphCodecDecoder` with capture, replay, and eager fallback
- `sglang_omni/models/higgs_tts/stages.py` — integrate CG decoder into `create_vocoder_executor`
- `sglang_omni/models/higgs_tts/config.py` — expose `enable_cuda_graph` in vocoder `factory_args`
- `tests/unit_test/higgs_tts/` — test CG decode correctness vs. eager decode

## Acceptance Criteria

- **Correctness:** CG decode output matches eager decode output (within float32 tolerance) for all bucket sizes; single-item and batched paths produce consistent results; existing pipeline tests pass unchanged.
- **Performance:** Lower per-request vocoder latency at all sequence lengths; no regression at `B=1`; warmup overhead under 10 s.

## Known Pitfalls

- `torch.tensor(0.0, device=...)` inside the vendored RVQ decode loop allocates a scalar each call; PyTorch's private memory pool handles this during capture, but an alternative is to pre-initialize the accumulator outside the graph boundary if issues arise.
- Padding codes with zeros (a valid codebook index) generates audio in the padded region; this audio is discarded via output slicing, but boundary artifacts from ConvTranspose receptive fields could bleed into the last few samples of real output — validate with A/B listening tests.
- Graph memory footprint scales with `max(T_bucket) * max_batch_size`; the largest graph (B=4, T=2048) allocates ~30 MB for the output buffer alone, and the 20-graph set should total under 500 MB, but warrants measurement on the target GPU.
- `ConvTranspose1d.output_padding` is baked into each module at model init and frozen into the captured graph — any post-init modification would silently produce wrong output on replay.
- If sequence lengths frequently exceed 2048 frames (~27 s), the eager fallback fires repeatedly; add a larger bucket or raise the cap to cover the workload.

cc @Ccyest @PopSoda2002

0xtoward:
## Motivation
Closes #581. The Higgs vocoder codec decode is launch-overhead bound (~420 `cudaLaunchKernel` of ~27 distinct kernels per B=1 streaming decode). A warmup-sealed CUDA graph collapses those launches: ~2.5x per-call, bit-exact (maxerr=0 on the real DAC). Opt-in (`SGLANG_OMNI_HIGGS_VOCODER_CUDA_GRAPH`, default off) → the eager path is unchanged.

## Modifications
- `vocoder_cuda_graph.py`: `VocoderCudaGraphRunner` — warmup-capture exact `(B, T)` shapes once at startup, seal, then replay-or-eager-fallback; never captures during serving (avoids racing the co-located AR stage's graph replays / shared mempool).
- `audio_codec.py` / `stages.py`: env gate + a capture range derived from the stream params.
- vendored RVQ decode: drop the zero-scalar seed (its H2D copy aborts capture) + fp32-pinned accumulation.
- Tests: runner replay/seal/fallback + codebook-mismatch; RVQ fp32-pin; warmup-range coverage; real-DAC bit-exact.

Draft to track #581.

## Performance (Higgs vocoder CUDA graph, default-off)

| Benchmark | Setup | Baseline: eager | This PR: CUDA graph | Speedup |
|---|---|---|---|---|
| Per-call decode latency (microbench) | standalone, B=1, T=64 frames | 5.40 ms | 2.17 ms | **2.5×** (bit-exact, maxerr=0) |
| E2E throughput — streaming | SeedTTS-EN, 1088 req, c=16 | 9.05 req/s | 9.74 req/s | **+7.7%** |
| E2E first-chunk latency (TTFC) — streaming | SeedTTS-EN, 1088 req, c=16 | baseline | −2% | lower is better |
| E2E throughput — non-streaming | SeedTTS-EN, 1088 req, c=16 | ~flat | ~flat | ~1.0× (B>1 batches miss the B=1 graph by design) |
| Quality (WER) | SeedTTS-EN | 0.44% | 0.29% | neutral |

Microbench: per-call decode is launch-bound at B=1 (~420 kernel launches collapse to 1 graph replay; same GPU work, hence bit-exact). E2E gain is bounded by the vocoder being <1% of the AR-dominated pipeline; 0 failed requests in either mode.


0xtoward:
Happy to notice that a following PR (#798) has adopted the same
warmup-sealed vocoder-CG pattern (for MOSS), and there is a RFC
(#802) consolidating these templates — cc @JiaxinD @Hayden727. @zhaochenyang20 
Glad to align on a shared base runner if useful.

zhaochenyang20:
@0xtoward Sorry for lagging so long. Would you like to have a further discuss with us? My wechat is 18015766633

PR #740, opened 2026-06-10
[Perf] MOSS-TTS Local: content-keyed LRU for reference codes

GaokaiZhang:
Adds a content-keyed LRU for encoded reference codes to the MOSS-TTS Local preprocessing stage (#637 roadmap: LRU cache for audio encoder, mirroring the Higgs design from #562).

Reference encoding costs ~0.25 GPU-seconds through the ~1B-param MOSS-Audio-Tokenizer-v2 encoder. The cache keys on a blake2b hash of the reference file contents (memoized per `(path, size, mtime_ns)` stat signature, so unchanged files hash once) and stores the CPU `[T, 12]` code tensors, capped at 256 items / 256 MB (`encode_cache_items` / `encode_cache_mb`, 0 disables). Repeated reference audio — one voice serving many requests — skips the codec entirely.

Hit-path effect (same reference, sequential requests against a live server, 2×H100): first request (miss) 2.44 s, subsequent requests (hits) 0.81 s mean.

Full Seed-TTS EN as the no-regression check — every reference in the set is unique, so this run exercises the pure miss path (stat + hash overhead, ~1–5 ms per request):

| Model / config | Split | WER / CER | Samples | Failed / skipped | Throughput | rtf_mean |
|---|---|---|---|---|---|---|
| MOSS-TTS-Local-v1.5 + ref LRU, c=16, token-count auto, 2×H100 | EN | WER 2.27% raw / 1.71% excl. >50% outliers | 1088/1088 | 0 / 0 | 7.91 req/s | 0.4811 |

Baseline on the same hardware (merged main, same day): rtf_mean 0.5053, WER 2.21% raw / 1.79% excl. — the deltas are within the run-to-run band documented in #728.

Unit tests cover hit/second-request reuse, stat-signature invalidation on content change, and item-cap eviction.

Issue #731, opened 2026-06-10
[Feat] MOSS TTS: LRU cache for reference audio encoder


Ccyest:
## Motivation

In typical TTS serving the set of speaker reference audios is small and reused heavily — same speaker prompt, many target texts. Once the preprocessing and encoding stages are decoupled (#730), the audio encoder stage will run `processor.encode_audios_from_wav()` on every request that supplies a raw waveform, even when the same reference audio has been encoded moments ago.

An LRU cache keyed by the reference-audio identity would let repeat requests skip the codec forward entirely and reuse cached `reference_codes`. The hit path becomes a dict lookup with no GPU work.

## Current encoder mechanism

Reference audio encoding currently runs inside the preprocessing stage (coupled — see #730). Once decoupled, the dedicated encoder stage will look like:

```python
def _encode(payload):
    state = load_state(payload)
    waveform = state.reference_waveform
    ref_codes = processor.encode_audios_from_wav(waveform, sample_rate=24000)
    state.reference_codes = ref_codes
    return payload
```

Every request with a raw waveform pays the full codec encode cost, regardless of whether an identical waveform was processed seconds ago.

## Proposed design

Reuse the `StageOutputCache` infrastructure (already used by Higgs) with content-based cache keys:

Cache key:
- `audio_hash`: BLAKE2 / SHA256 of the raw waveform bytes (or a content hash for file-path references)
- Reference text (if it affects encoding)

Cache value:
- Encoded reference codes (the heavy GPU output)

### Considerations

- **Eviction**: LRU with configurable `maxsize` (default 128). Each cached entry is `[T, 32]` long tensor (32 codebooks). Typical ref audio is ~3–10s → a few hundred frames → tens of KB per entry. Cache of 128 is well under 10 MB.
- **Hash cost**: BLAKE2 of a 10s waveform (~480 KB float32) is sub-millisecond on CPU, much cheaper than the codec forward.
- **No correctness risk**: identical waveform → identical codec output (deterministic encoder).
- **Dependency**: requires #730 (decouple preprocessing) to land first, so there is a clean encoder stage to cache.

## Acceptance Criteria

- [ ] Cache hit path skips codec forward entirely
- [ ] Cache miss path produces bit-identical output to uncached path
- [ ] Repeated requests with same reference audio show measurable latency reduction
- [ ] Cache eviction works correctly under memory pressure
- [ ] Existing unit tests still pass

## References

- Higgs LRU cache: #562, PRs #563 #605
[...]

Issue #730, opened 2026-06-10
[Perf] MOSS TTS: decouple preprocessing and reference audio encoding


Ccyest:
## Motivation

The MOSS TTS pipeline currently has 3 stages: `preprocessing → tts_engine → vocoder`. The preprocessing stage (`create_preprocessing_executor` in `sglang_omni/models/moss_tts/stages.py`) runs both CPU text tokenization **and** reference audio encoding (codec forward pass) in the same `SimpleScheduler` with `max_concurrency=8`.

This coupling causes two problems:

1. **No dedicated GPU encoder stage.** Reference audio encoding runs on CPU (`device="cpu"` in `_load_moss_processor`), missing the opportunity for GPU-accelerated batched encoding.
2. **Cannot cache or batch encode independently.** Since encoding is buried inside the processor's internal call path (`processor.encode_audios_from_wav()`), there is no clean boundary to add an LRU cache or batched GPU encode — the prerequisites for #562-style and #603-style optimizations.

The Higgs TTS pipeline already uses the 4-stage split: `preprocessing → audio_encoder → tts_engine → vocoder`, where the dedicated `audio_encoder` stage runs on GPU with its own `SimpleScheduler`, content-hash cache, and batched encode support.

## Current path

```
preprocessing (CPU, max_concurrency=8)
    ├── text tokenization
    ├── reference audio waveform loading (file / base64)
    └── reference audio codec encoding ← coupled here
→ tts_engine (GPU, OmniScheduler)
→ vocoder (GPU, SimpleScheduler)
```

Reference encoding is triggered inside `preprocess_moss_tts_payload()` → `_reference_for_processor()` → processor internal encode. The processor produces merged prompt rows containing both text and audio tokens, so there is no clean separation point today.

## Proposed change

Split into 4 stages matching Higgs:

```
preprocessing (CPU) — text tokenization + waveform loading only
→ audio_encoder (GPU) — reference audio codec encoding
→ tts_engine (GPU) — AR decode
→ vocoder (GPU) — waveform synthesis
```

Key steps:
1. Extract reference waveform loading from `_reference_for_processor()` into `MossTTSState` fields (raw waveform + sample rate)
2. Create `create_audio_encoder_executor()` that takes the raw waveform, runs `processor.encode_audios_from_wav()` on GPU, and stores `reference_codes` in state
3. Move prompt-row assembly (merging text tokens + encoded audio tokens) to either the encoder stage or a lightweight post-encode step
[...]

PR #743, opened 2026-06-10
[Perf] Reference audio cache for MOSS-TTS-Local-v1.5 (LRU + single-flight dedup)

JiaxinD:
# [Perf] Reference audio cache for MOSS-TTS-Local-v1.5 (LRU + single-flight dedup)

Addresses #731 (roadmap: #637) — the MOSS-TTS reference-audio encoder LRU cache.  
Builds on AkazaAkane's prior design in #648 and the Higgs reference cache in #605.

---

## Motivation

MOSS-TTS-Local-v1.5's `_BatchedReferenceEncoder` runs the ~1B-param causal codec encoder
for every reference audio, costing roughly **0.25 GPU-seconds per encode** on the codec
device. #728 already coalesces concurrent requests into batches (up to 8 per 4 ms window,
`stages.py:202`), but the same reference arriving at different times still re-encodes
independently. For the common "fixed speaker" workload this is redundant work on a GPU
already shared with AR prefill.

Radix-cache in the SGLang backend covers AR-side prefix reuse (#728 body); this PR targets
only the **encode side** (the incremental cost above and beyond what #728 provides).

---

## What this PR does (5 commits)

| Commit | Title | Key change |
|---|---|---|
| a5fa07c | `[Refactor]` Extract shared reference-audio cache key helpers | Higgs memo/sentinel/hash functions → `preprocessing/cache_key.py`; Higgs re-exports aliases (zero behaviour change) |
| 400e189 | `[Perf]` Add cached reference encoder for MOSS-TTS Local | `CachedReferenceEncoder` wrapping `_BatchedReferenceEncoder` |
| 6a7b7b1 | `[Config]` Wire ref-audio cache flags into preprocessing executor | Three kwargs + `config.py` factory_args; 60 s stats logging |
| b19baa7 | `[Perf]` Cache data-URI references (bytes: keyspace) + add duration gate | `encode_data_uri`; `bytes:` keyspace; data-URI duration check |
| fa4c22a | `[Perf]` Add opt-in stat-trusting fast path for reference cache keys | `trust_stat` flag (zero-I/O memo hit); co-authored with GaokaiZhang (#740) |
| 2855ac1 | `[Review]` Address review | validate capacity; DRY `_cached_encode`; **MOSS defaults `trust_stat=False`**; log cadence constant |

---

## Cross-model note (commit a5fa07c touches `higgs_tts/stages.py`)

Commit 1 is a pure **extraction refactor** of `higgs_tts/stages.py`: the memo dict,
sentinel reader, and `reference_path_cache_key` function were moved verbatim into
`sglang_omni/preprocessing/cache_key.py`, and `higgs_tts/stages.py` now re-imports them
at the same names (`stages.py:53–60`). No Higgs logic was changed; all existing Higgs
tests pass without modification (`tests/unit_test/higgs_tts/test_pipeline.py:117–194`).
The refactor is a prerequisite so MOSS-TTS-Local can share the same key scheme without
duplicating 80 lines.

---

## Architecture

```
request
  └─ CachedReferenceEncoder.encode(path)
        │
        ├─ 1. duration gate (>100 s → reject)           stages.py:266
        │     (file path: gated first via cheap torchaudio.info;
        │      data-URI: checked inside the leader so concurrent
        │      same-payload requests share one sf.read)
        ├─ 2. key = reference_path_cache_key(path)       stages.py:267
        │       └─ stat-memo (size+mtime+ctime) → sentinel (8 KB×3) → xxh3_64
        │          → "file:{hex}" or None (missing/URL)  cache_key.py:115–135
        │
        ├─ 3. cache hit → return stored.clone().to(long) stages.py:276–280
        ├─ 4. in-flight (single-flight) → follower waits stages.py:280–286
        └─ 5. miss → register leader Future
                └─ _BatchedReferenceEncoder.encode(path) (4 ms coalescer)
                        └─ processor.encode_audios_from_path(unique_paths)
                              ↓
               TOCTOU re-stat before put               stages.py:313–319
               cache.put(key, stored_int32_cpu)
               future.set_result(stored)
```

**Keyspaces** (two distinct prefixes, never cross-contaminate):

| Input | Prefix | Key content |
|---|---|---|
| File path | `file:` | xxh3\_64 of full file bytes, gated by stat-memo + sentinel |
| data-URI | `bytes:` | xxh3\_64 of raw decoded bytes |
| Missing / URL / tensor | `None` | cache and single-flight bypassed entirely |

**Storage**: `detach().to("cpu", torch.int32)` — matches Higgs code-cache discipline
(`higgs_tts/stages.py:365`). On hit: `.clone().to(torch.long)`. On miss: original codec
tensor returned unchanged (bit-identical to OFF=True path).

**Capacity defaults** (`config.py:23–25`):

| Flag | Default | Worst-case RAM |
|---|---|---|
| `ref_audio_cache_max_items` | 256 | ~15 MB (100 s clips × 12 VQ × 4 B × 256) |
| `ref_audio_cache_max_bytes` | 64 MiB | hard backstop |

**Failure semantics**: leader failure clears `_inflight` without writing to cache;
each follower receives an independent `RuntimeError(...) from cause` (not the shared
exception instance — mirrors the existing `stages.py:216–221` note in
`_BatchedReferenceEncoder`). Third request becomes a new leader and may succeed.

---

## Correctness gate

Unit test `test_cached_reference_encoder_on_off_hit_bit_identical` (`test_pipeline.py:~490`):
- Runs the full `_prepare_moss_tts_local_request` chain with a fake-but-correct processor
- Asserts OFF path == ON-miss path == ON-hit path (identical `prompt_rows` and
  `input_ids_list`)
- Verifies int32 round-trip is lossless at the boundary value 1023
- Confirms `encode_count == 1` on the second (hit) call

Smoke test on novita-h100 with real MOSS-TTS-Local-Transformer-v1.5 weights (GPU 3):
- hit-path CPU tensor accepted by `processor.build_user_message` ✓
- T=62 frames / 5.0 s = 12.40 Hz ≈ 12.5 Hz empirically confirmed ✓

---

### Corroborating numbers from #740 (same approach, GaokaiZhang's implementation)

> These measure #740's implementation of the same content-keyed LRU, not this
> branch's code. They corroborate the mechanism; our own Phase 3 numbers on this
> branch (table below) will be the primary evidence. Source: #740 PR body, 2×H100.

- Hit-path effect (same reference, sequential requests, live server): first request
  (miss) **2.44 s** → subsequent requests (hits) **0.81 s** mean.
- SeedTTS EN full (pure miss path, every reference unique), c=16, 2×H100:
  rtf_mean **0.4811**, WER 2.27% raw / 1.71% excl., 1088/1088, 0 failed, 7.91 req/s.
  Same-day baseline on merged main: rtf_mean 0.5053, WER 2.21% raw / 1.79% excl.
  (deltas within the run-to-run band documented in #728).

### Our Phase 3 (this branch) — measured

> **Environment (full disclosure):** novita-h100. A sole-occupant run was planned on NUMA1
> (GPU5+6); mid-run an unrelated co-tenant took GPU6+7 and saturated NUMA1 CPU, so the
> benchmark **migrated to GPU4+5 with CPU pinned to NUMA0** (`numactl --cpunodebind=0
> --membind=0`). Cross-NUMA GPU access is a **constant** overhead identical for the ON and
> OFF arms and every round (ABAB interleaved), so it does **not bias the A/B** — it only
> shifts the absolute rtf baseline. Each round ran a co-tenant-contamination gate; all 6
> Load-A rounds passed clean. Full methodology + per-round data in the run report.

**Load A — SeedTTS EN full set (1088), ABAB ON/OFF ×3 (1088/1088 each, 0 failed)**

| metric | Cache OFF (mean ± 95% CI) | Cache ON (mean ± 95% CI) |
|---|---|---|
| RTF mean | 0.5096 ± 0.0387 | 0.4973 ± 0.0257 |
| throughput req/s | 7.443 ± 0.633 | 7.637 ± 0.414 |
| RTF p99 | 1.0547 ± 0.1942 | 1.0029 ± 0.0450 |

The ON/OFF 95% CIs **overlap** → the cache is statistically indistinguishable from OFF =
**no regression** (zero measurable overhead). The nominal ~2.5% ON edge is within noise at 3
rounds; the saved encode work surfaces as returned GPU headroom (below), not an rtf gain at
c=16 where AR generation dominates.

**Native reuse + savings (hard counts).** SeedTTS EN recurs speakers across target texts, so
the standard benchmark itself exercises the cache. One full-set ON snapshot:

| lookups | encodes (misses) | hits | merged (single-flight) | reuse | encodes avoided | GPU-s saved |
|---|---|---|---|---|---|---|
| 915 | 566 | 71 | 278 | 38.1% | 349 | ~87 |

`merged` (single-flight dedup) is a capability **#740 lacks** — hundreds of concurrent
same-reference requests folded onto one in-flight encode instead of re-encoding.

**WER sanity (quality non-regression):** ON **2.39% raw / 1.82% excl-outliers** vs OFF
**2.22% raw / 1.68% excl** (Qwen3-ASR, 1088/1088 each, 6 outliers each). The ~0.17%
difference is AR sampling noise (temp=0.8 → non-identical audio run-to-run) — same band as
#740's 2.27%/2.21%. Codes-level ON==re-encode identity is separately proven by unit test
T5; WER_ON ≈ WER_OFF confirms no audible regression.

**Load B — fixed-speaker reuse sweep (N=512, cache ON), mechanism curve (hard counts):**

| K speakers | encodes performed (=K) | encodes saved (512−K) | GPU-s saved (×0.25) | hit-rate |
|---|---|---|---|---|
| 1 | 1 | 511 | ~128 | 0.998 |
| 4 | 4 | 508 | ~127 | 0.992 |
| 16 | 16 | 496 | ~124 | 0.969 |
| 64 | 64 | 448 | ~112 | 0.875 |
| 256 | 256 | 256 | ~64 | 0.500 |

Lower K → more reuse → more encodes skipped → more GPU-seconds returned to the AR engine.
(Load-B end-to-end throughput was AR-generation-bound with token-count uncontrolled in the
client, so it is **not** a cache comparison; the value is the exact encode/GPU-second
savings above, which are hard counts with zero statistical noise.)

---

## Relation to #740 and #730

**#740** ("[Perf] MOSS-TTS Local: content-keyed LRU for reference codes", GaokaiZhang)
independently implemented the same #637 roadmap item. We have **converged on this PR
as the base** (coordinated with GaokaiZhang). The stat-only memo fast path from #740 is
incorporated here as the opt-in `trust_stat` flag (**MOSS defaults it `False` per review;
opt-in for immutable-reference deployments**) (commit "Add opt-in stat-trusting fast
path", co-authored with GaokaiZhang), and #740's benchmark numbers are cited below with
attribution. #740 will close in favor of this PR.

**#730 / #744** (decouple): #744 ("Split preprocessing; add audio_encoder stage") is the
MOSS-TTS-Local half of the #730 decouple — it moves the heavy reference-encode work out of
`preprocessing` into a dedicated `audio_encoder` stage. This PR is **complementary**: the
cache is an outer wrapper around that encoder and migrates with it. When #744 merges after
#743, the `CachedReferenceEncoder` wrapping and its flags move from
`create_preprocessing_executor` to `create_audio_encoder_executor` (~13 lines, no logic
change); reference-encode semantics are byte-identical across the split. Porting the same
cache to the older `moss_tts` (Delay) path is a separate follow-up (cf. AkazaAkane's
#648/#699).

---

## Cap discussion / Future work

The reference encode cache and the AR-side prefix cache are **orthogonal**: a cache hit
eliminates the codec GPU-second but still consumes a preprocessing-worker slot from
`max_concurrency=16`. If the AR concurrency cap is raised to 32 in a future PR,
`preprocessing max_concurrency=16` may become the new bottleneck under high load —
the cache shifts where time is spent (codec → Python dispatch) but does not by itself
increase the worker pool size. Raising `max_concurrency` in tandem, or combining with
the async-decode lookahead work in the pipeline, would complete the optimisation.

---

## Testing

```
pytest tests/unit_test/moss_tts_local/test_pipeline.py  -x   # T1–T9 + existing tests
pytest tests/unit_test/higgs_tts/test_pipeline.py       -x   # Higgs regression (T1/T2)
pytest tests/unit_test/preprocessing/                   -x   # key helper tests
```

94 collected, 94 passed on novita-h100 (sglang-omni docker, full deps).

New tests added:

| Test | What it covers |
|---|---|
| `test_cached_reference_encoder_on_off_hit_bit_identical` | Bit-identical gate (T5) |
| `test_cached_reference_encoder_lru_eviction` | Item-count + byte-budget eviction (T3) |
| `test_cached_reference_encoder_true_concurrency_dedup` | Single-flight, 8 threads (T4) |
| `test_cached_reference_encoder_failure_does_not_poison` | Independent exceptions, no cache poison (T6) |
| `test_cached_reference_encoder_return_value_isolation` | Clone isolation (T7) |
| `test_cached_reference_encoder_duration_gate` | >100 s rejected pre-inflight (T8) |
| `test_cached_reference_encoder_data_uri_hit_miss` | bytes: keyspace hit/miss |
| `test_cached_reference_encoder_file_bytes_keyspaces_do_not_collide` | Keyspace isolation |
| `test_cached_reference_encoder_data_uri_duration_gate` | data-URI duration check |

### DIFF
--- a/sglang_omni/models/moss_tts_local/stages.py
+++ b/sglang_omni/models/moss_tts_local/stages.py
@@ -264,7 +264,10 @@ class CachedReferenceEncoder:
         path = str(path)
         # Duration gate first: >100 s must never enter the cache or inflight dict.
         _BatchedReferenceEncoder._check_reference_duration(path)
-        key = _reference_path_cache_key(path)
+        # trust_stat: skip the sentinel byte-read on memo hits — the codes cache
+        # lookup needs the key on every request, so the stat-only fast path saves
+        # an 8-24 KB read on the hot fixed-speaker path (idea from #740).
+        key = _reference_path_cache_key(path, trust_stat=True)
         if key is None:
             # Uncacheable (URL, missing file, etc.) — bypass entirely.
             return self._encoder.encode(path)
@@ -311,7 +314,7 @@ class CachedReferenceEncoder:
             raise
 
         # TOCTOU re-stat: skip cache if file changed between key computation and encode.
-        rekey = _reference_path_cache_key(path)
+        rekey = _reference_path_cache_key(path, trust_stat=True)
         stored = result.detach().to("cpu", dtype=torch.int32)
         with self._lock:
             if rekey == key:
--- a/sglang_omni/models/moss_tts_local/stages.py
+++ b/sglang_omni/models/moss_tts_local/stages.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 
 import concurrent.futures
 import logging
+import os
 import queue
 import threading
 import time
@@ -449,6 +450,17 @@ def create_preprocessing_executor(
     ref_audio_cache_max_items: int = 256,
     ref_audio_cache_max_bytes: int = 64 * 1024 * 1024,
 ) -> SimpleScheduler:
+    # Runtime kill switch / A-B toggle: MOSS_REF_AUDIO_CACHE=0 disables the cache
+    # without a config edit (the design's "线上回退" lever). Unset => kwarg default.
+    env_toggle = os.environ.get("MOSS_REF_AUDIO_CACHE")
+    if env_toggle is not None:
+        ref_audio_cache = env_toggle.strip().lower() not in (
+            "0",
+            "false",
+            "no",
+            "off",
+            "",
+        )
     device = _resolve_codec_device(device, gpu_id)
     processor = _load_moss_tts_local_processor(model_path, device=device)
     reference_encoder: Any = _BatchedReferenceEncoder(
--- a/sglang_omni/models/moss_tts_local/stages.py
+++ b/sglang_omni/models/moss_tts_local/stages.py
@@ -7,6 +7,7 @@ import concurrent.futures
 import logging
 import queue
 import threading
+import time
 from typing import Any
 
 import torch
@@ -257,6 +258,7 @@ class CachedReferenceEncoder:
         self._hits = 0
         self._misses = 0
         self._merged = 0
+        self._last_log_time: float = 0.0
 
     def encode(self, path: str) -> torch.Tensor:
         path = str(path)
@@ -317,7 +319,28 @@ class CachedReferenceEncoder:
             self._inflight.pop(key, None)
         leader_fut.set_result(stored)
         # Return the original tensor (miss path == cache-off path, bit-identical).
+        self._maybe_log()
         return result
+
+    def _maybe_log(self) -> None:
+        now = time.monotonic()
+        if now - self._last_log_time < 60.0:
+            return
+        with self._lock:
+            if now - self._last_log_time < 60.0:
+                return
+            self._last_log_time = now
+            snapshot = (
+                self._hits,
+                self._misses,
+                self._merged,
+                len(self._cache._cache),
+                self._cache.current_bytes,
+            )
+        logger.info(
+            "MOSS-TTS Local ref cache: hits=%d misses=%d merged=%d entries=%d bytes=%d",
+            *snapshot,
+        )
 
     def stats(self) -> dict:
         with self._lock:
@@ -338,14 +361,23 @@ def create_preprocessing_executor(
     max_concurrency: int = 16,
     encode_batch_size: int = 8,
     encode_batch_wait_ms: int = 4,
+    ref_audio_cache: bool = True,
+    ref_audio_cache_max_items: int = 256,
+    ref_audio_cache_max_bytes: int = 64 * 1024 * 1024,
 ) -> SimpleScheduler:
     device = _resolve_codec_device(device, gpu_id)
     processor = _load_moss_tts_local_processor(model_path, device=device)
-    reference_encoder = _BatchedReferenceEncoder(
+    reference_encoder: Any = _BatchedReferenceEncoder(
         processor,
         max_batch_size=encode_batch_size,
         max_batch_wait_ms=encode_batch_wait_ms,
     )
+    if ref_audio_cache:
+        reference_encoder = CachedReferenceEncoder(
+            reference_encoder,
+            max_items=ref_audio_cache_max_items,
+            max_bytes=ref_audio_cache_max_bytes,
+        )
     set_moss_tts_local_preprocessing_context(
         processor=processor, reference_encoder=reference_encoder
     )
--- a/sglang_omni/models/moss_tts_local/stages.py
+++ b/sglang_omni/models/moss_tts_local/stages.py
@@ -342,6 +342,87 @@ class CachedReferenceEncoder:
             *snapshot,
         )
 
+    def encode_data_uri(self, ref_audio: str, *, processor: Any) -> torch.Tensor:
+        """Cache-aware encode for data-URI references (bytes: keyspace).
+
+        Adds the duration check that _reference_for_processor lacks, and routes
+        through the same LRU + single-flight machinery as file-path references.
+        file: and bytes: keys never collide because the encode chains differ.
+        """
+        import base64
+        import io
+
+        from sglang_omni.models.moss_tts.request_builders import _DATA_URI_RE
+        from sglang_omni.preprocessing.cache_key import hash_bytes as _hash_bytes
+
+        match = _DATA_URI_RE.match(ref_audio)
+        if match is None:
+            raise ValueError(f"encode_data_uri: not a data URI ({ref_audio[:40]!r}...)")
+
+        raw = base64.b64decode(match.group("data"))
+        key = f"bytes:{_hash_bytes(raw)}"
+
+        leader_fut: concurrent.futures.Future | None = None
+        follower_fut: concurrent.futures.Future | None = None
+
+        with self._lock:
+            stored = self._cache.get(key)
+            if stored is not None:
+                self._hits += 1
+                out = stored.clone().to(torch.long)
+            elif key in self._inflight:
+                self._merged += 1
+                follower_fut = self._inflight[key]
+            else:
+                self._misses += 1
+                leader_fut = concurrent.futures.Future()
+                self._inflight[key] = leader_fut
+
+        if stored is not None:
+            self._maybe_log()
+            return out  # type: ignore[return-value]
+
+        if follower_fut is not None:
+            timeout = _BatchedReferenceEncoder.ENCODE_TIMEOUT_S + 10
+            try:
+                stored = follower_fut.result(timeout=timeout)
+            except Exception as cause:
+                raise RuntimeError(
+                    f"reference encode failed (data-URI, merged): {cause}"
+                ) from cause
+            self._maybe_log()
+            return stored.clone().to(torch.long)
+
+        # Leader: decode audio, check duration, encode.
+        assert leader_fut is not None
+        try:
+            import soundfile as sf
+
+            audio, sample_rate = sf.read(
+                io.BytesIO(raw), dtype="float32", always_2d=True
+            )
+            duration = audio.shape[0] / max(int(sample_rate), 1)
+            if duration > _BatchedReferenceEncoder.MAX_REFERENCE_SECONDS:
+                raise ValueError(
+                    f"reference audio is {duration:.1f}s long; the limit is "
+                    f"{_BatchedReferenceEncoder.MAX_REFERENCE_SECONDS:.0f}s"
+                )
+            wav = torch.from_numpy(audio.T)
+            result = processor.encode_audios_from_wav([wav], int(sample_rate))[0]
+        except BaseException as exc:
+            with self._lock:
+                self._inflight.pop(key, None)
+            leader_fut.set_exception(exc)
+            raise
+
+        stored = result.detach().to("cpu", dtype=torch.int32)
+        with self._lock:
+            self._cache.put(key, stored)
+            self._inflight.pop(key, None)
+        leader_fut.set_result(stored)
+        self._maybe_log()
+        return result
+
     def stats(self) -> dict:
         with self._lock:
             return {
--- a/sglang_omni/models/moss_tts_local/stages.py
+++ b/sglang_omni/models/moss_tts_local/stages.py
@@ -26,8 +26,12 @@ from sglang_omni.models.moss_tts_local.request_builders import (
     preprocess_moss_tts_local_payload,
     set_moss_tts_local_preprocessing_context,
 )
+from sglang_omni.preprocessing.cache_key import (
+    reference_path_cache_key as _reference_path_cache_key,
+)
 from sglang_omni.proto import StagePayload
 from sglang_omni.scheduling.simple_scheduler import SimpleScheduler
+from sglang_omni.scheduling.stage_cache import StageOutputCache
 from sglang_omni.utils.audio_payload import audio_waveform_payload
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -227,6 +233,206 @@ def _worker(self) -> None:
...
+                    f"reference encode failed (data-URI, merged): {cause}"
+                ) from cause
+            self._maybe_log()
+            return stored.clone().to(torch.long)
+
+        assert leader_fut is not None
+        try:
+            import soundfile as sf
+
+            audio, sample_rate = sf.read(
```
gaoyang07, line 396: We need to read raw first, and then decode the raw to avoid invalid _DATA_URI_RE refs?

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -227,6 +233,206 @@ def _worker(self) -> None:
...
+            *snapshot,
+        )
+
+    def encode_data_uri(self, ref_audio: str, *, processor: Any) -> torch.Tensor:
+        """Cache-aware encode for data-URI refs through the same LRU + single-flight
+        as file paths (adds the duration check _reference_for_processor lacks).
+
+        Note(Jiaxin): file: and bytes: keyspaces never collide — the two decode
+        chains differ, so codes aren't guaranteed identical for the "same" audio.
+        """
```
Ccyest, line 379: encode() and encode_data_uri() each implement a nearly identical single-flight flow.

Suggestion: Extract a private method _cached_encode(key, encode_fn) that encapsulates the single-flight skeleton. encode and encode_data_uri would only be responsible for key computation + constructing the encode_fn callable.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -227,6 +233,206 @@ def _worker(self) -> None:
...
+            if rekey == key:
+                self._cache.put(key, stored)
+            self._inflight.pop(key, None)
+        leader_fut.set_result(stored)
+        self._maybe_log()
+        return result  # original tensor: miss path stays bit-identical to cache-off
+
+    def _maybe_log(self) -> None:
+        now = time.monotonic()
+        if now - self._last_log_time < 60.0:
```
gaoyang07, line 355: Make 60.0 configurable maybe?

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -227,6 +233,206 @@ def _worker(self) -> None:
...
+        self._last_log_time: float = 0.0
+
+    def encode(self, path: str) -> torch.Tensor:
+        path = str(path)
+        # Note(Jiaxin): duration gate runs first — a >100 s ref must never reach
+        # the cache or the inflight dict.
+        _BatchedReferenceEncoder._check_reference_duration(path)
+        # trust_stat: stat-only fast path; the codes lookup needs the key on every
+        # request (see reference_path_cache_key).
+        key = _reference_path_cache_key(path, trust_stat=True)
```
gaoyang07: Is it ok to set `trust_stat=True` as default?

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -227,6 +233,206 @@ def _worker(self) -> None:
...
+    Hit path returns a fresh .clone().to(long) so callers cannot mutate cached state.
+    Stores codes as int32 on CPU (lossless for codebook values in [0, 1023]).
+    """
+
+    def __init__(
+        self,
+        encoder: _BatchedReferenceEncoder,
+        *,
+        max_items: int = 256,
+        max_bytes: int = 64 * 1024 * 1024,
```
Hayden727, line 252: [P3] These are new public `factory_args`, so it would be safer to validate them before building the `StageOutputCache`. Today a negative `ref_audio_cache_max_items` starts successfully, but the first cacheable `put()` can pop the only entry and then keep evicting from an empty `OrderedDict`, raising `KeyError` at request time. Can we fail fast here (or in `create_preprocessing_executor`) for negative capacities and add a small negative-config regression test?

sglang_omni/models/moss_tts_local/stages.py
AkazaAkane, line 396: One minor note: the PR mentions that the duration gate happens before in-flight registration, but the code actually registers _inflight[key] before running sf.read() and the duration check. I actually think the code's way is better, as it prevents concurrent sf.read() calls on the same massive payload (saving CPU/RAM)

sglang_omni/models/moss_tts_local/stages.py
AkazaAkane: trust_stat=True -> Once a file has been hashed once, future requests avoid even the small sentinel read. That matters because the cache key is computed on every request.

But if the path is reused as a mutable scratch file, （for example /tmp/ref.wav）and a caller overwrites it between requests, the cache may return stale codes if the stat tuple does not change enough. The worst outcome is not a crash; it is incorrect speaker/reference conditioning.

So probably need 1. a comment/config option saying MOSS assumes path references are immutable while serving; or 2. A safe-mode option using trust_stat=False for deployments with mutable reference paths.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -227,6 +233,206 @@ def _worker(self) -> None:
...
+
+        from sglang_omni.models.moss_tts.request_builders import _DATA_URI_RE
+        from sglang_omni.preprocessing.cache_key import hash_bytes as _hash_bytes
+
+        match = _DATA_URI_RE.match(ref_audio)
+        if match is None:
+            raise ValueError(f"encode_data_uri: not a data URI ({ref_audio[:40]!r}...)")
+
+        raw = base64.b64decode(match.group("data"))
+        key = f"bytes:{_hash_bytes(raw)}"
```
Ccyest, line 391: encode_data_uri decodes the entire data-URI payload (potentially hundreds of KB) before computing hash_bytes(raw) for the cache key. On a cache hit, these decoded raw bytes are immediately discarded, each time can cause CPU waste.

I feel like hashing the base64-encoded string directly as the cache key can be a solution, but I checked higgs we also did the entire data hashing. I remember we might discussed it previously but forgot why we made that decision. 

Theoretically it should be correct tho, any thoughts?

JiaxinD: Dug up the original decision, it's from #605 (commit 1bbac5a, by BBuf), recorded in the tests rather than a discussion thread: test_higgs_reference_cache_key_ignores_media_type pins down two invariants: (1) media_type is a decode hint the codec ignores, so it must not split keys; (2) raw bytes and their base64 encoding must resolve to the same content key. Hashing the base64 string would violate both: the same audio arriving as bytes vs base64, or wrapped with a different media_type, would miss the cache and re-encode. That's why the key is computed on decoded raw bytes (content-addressed): _reference_audio_cache_key, then b64decode, then hash_media_item(raw), giving bytes:{hash}.

On skipping the decode: I looked at the numbers and it's thinner than it sounds. The expensive part (0.25s codec encode) is already absorbed by the codes cache on a hit, so what's left on the hit path is just the b64decode + xxh3, both µs-level C code. A string-level memo in front (same structure as our _REF_PATH_HASH_MEMO on the file path) is ~1-2h to add and low-risk, but the memo key itself is the problem: keying on the full base64 string is memory-heavy (~5.8MB per 100s clip), and keying on a cheap hash of the base64 costs about the same as hashing the raw bytes (base64 is 1.33x larger). So unless profiling shows key-compute as a hotspot, which is unlikely given the codes cache eats the real cost, I'd leave the keyspace as is. Happy to add the memo as a follow-up if a workload surfaces where it matters.

sglang_omni/models/moss_tts_local/stages.py
JiaxinD: Good catch — fixed in `2855ac1`. `CachedReferenceEncoder.__init__` now raises `ValueError` for non-positive `max_items`/`max_bytes` (fail-fast at construction, before any request can hit the empty-`OrderedDict` eviction path you described), with a regression test `test_cached_reference_encoder_rejects_nonpositive_capacity`.

sglang_omni/models/moss_tts_local/stages.py
JiaxinD: Done in `2855ac1`. Extracted `_cached_encode(key, encode_fn, *, desc, revalidate)` holding the whole single-flight skeleton (lock -> hit / inflight-follower / leader, follower wait, leader encode + put). `encode()` and `encode_data_uri()` now only compute the key and pass an `encode_fn` callable; `encode()` additionally passes a `revalidate()` for its TOCTOU re-stat (data-URI needs none, so it omits it). All concurrency / failure-isolation / clone tests stay green.

sglang_omni/models/moss_tts_local/stages.py
JiaxinD: Done in `2855ac1` -- pulled the cadence out as a class attribute `CachedReferenceEncoder.LOG_INTERVAL_S` (= 60.0), so it is easy to tune or override.

sglang_omni/models/moss_tts_local/stages.py
JiaxinD: The flow is already validate-then-decode: `_DATA_URI_RE.match(ref_audio)` runs first and the `if match is None: raise ValueError(...)` guard rejects non-data-URI refs, so `base64.b64decode(...)` only ever sees a matched payload and `raw` is the decoded bytes we hash. Were you thinking of a different invalid case (e.g. a matched header but corrupt base64)? That currently raises from `b64decode` in the leader and fails just that request.

JiaxinD: You're right, and the asymmetry is intentional -- thanks for catching the doc mismatch. File-path refs gate on duration first (`torchaudio.info` is a cheap metadata read; no point registering inflight for an oversized file). Data-URI refs run the duration check inside the leader (after inflight registration) for exactly the reason you gave: it needs `sf.read()` of the decoded buffer, so doing it once in the leader lets concurrent same-payload requests merge instead of each re-reading a large payload. Added a `Note` comment at the data-URI leader in `2855ac1`, and I'll fix the PR-description wording.

sglang_omni/models/moss_tts_local/stages.py
JiaxinD: On the protection chain first: the memo key is `(size, mtime_ns, ctime_ns)`, and `ctime` updates on essentially any write -- so a normal overwrite of a scratch file like `/tmp/ref.wav` **does** invalidate (new ctime -> memo miss -> re-hash). The only residual stale window is a replacement that preserves size **and** mtime_ns **and** ctime_ns at once, reachable only by rolling the clock back during the write -- not a normal operation.

That said, you're both right to be cautious. The sentinel read this skips is microsecond-level, and the real cost (the ~0.25 s codec encode) is already absorbed by the codes cache -- so the optimization doesn't earn even a small stale-read doubt. I've defaulted the MOSS call sites back to `trust_stat=False` (`2855ac1`), keeping the sentinel-validated path; the `trust_stat` flag and fast path stay in `reference_path_cache_key` as an opt-in for deployments that can guarantee immutable references. Thanks both.

Ccyest (approved):

Issue #736, opened 2026-06-10
[Perf] MOSS-TTS Local v1.5: native pool-resident CUDA graph for the frame-decode sampling path


Ccyest:
> **Rescoped 2026-06-11.** This issue was originally written against the **Delay** model (`MossTTSModelRunner._collect_moss_step`, 33 delay-pattern heads) and assumed MOSS TTS has no CUDA graph. Both are stale: the optimization target is **MOSS-TTS Local v1.5** (`moss_tts_local`), and Local v1.5 **already** has a frame-decode CUDA graph from #728. So the task is not "add CUDA graph" — it is "make the existing capture *native* over the row-indexed state pool and remove the per-frame host syncs." Corrected scope below.

### What #728 already captures (the baseline)
`sglang_model.py::init_frame_decode_graphs` captures `_decode_frame_graphable` — the per-frame local-transformer decode (`1 + n_vq` micro-steps + the 13 seeded sampling passes) — into one CUDA graph per batch bucket `[1, 2, 4, 8, 16]`. Eager is ~500 kernels / ~22 ms per frame (launch-bound, batch-independent); replay collapses that to a few ms. It uses:
- **private static scratch buffers** for inputs with per-frame **H2D copy-in** (`decode_frame_graphed`),
- static output buffers with **copy-out**,
- the local-transformer KV cache pre-sized to max bs and **frozen** against reallocation.

So the capture today is **narrow — compute only**. The per-frame orchestration around it is still eager.

### Problem: the eager glue around the graph
Each frame, outside the captured region (`MossTTSLocalModelRunner._collect_frame`):
1. gather per-request sampling params into the scratch buffers — Python gather + **H2D copy-in**;
2. clone codes/feedback out of the static output buffers — **copy-out**;
3. `_row_radix_token_ids()` → `build_row_cache_key_ids(rows)` — a **GPU→CPU→blake2b→GPU** radix-key hash (`model_runner.py:273`);
4. write `pool.feedback_embeds` / journal.

(1) and (3) are per-frame **host syncs**: they serialize the GPU, keep small-batch host-bound, and block async-decode (#734) overlap.

### Scope: native, pool-resident capture
Built on the row-indexed state pool from #745 (implements #735):
- Capture reads sampling params/state **directly from `pool.{text_temp,audio_temp,top_p,top_k,seeds,...}[stable row slots]`** — no per-frame gather / copy-in.
[...]

Issue #735, opened 2026-06-10
[Perf] MOSS TTS: state-pool refactor for decode state management


Ccyest:
## Motivation

MOSS TTS decode state is currently managed through request-local Python objects (`MossTTSSGLangRequestData` in `sglang_omni/models/moss_tts/request_builders.py`). Next-step feedback — delay-pattern counters, last-sampled codes per codebook, EOC countdowns — is stored and consumed through per-request Python state. This design has two consequences:

1. **Limits async decode benefit.** The async launch phase must finish sampling, feedback embedding preparation, and next-step state commit before the next decode step can start (see #734). There is much less resolve-side work to overlap than in Higgs, which uses row-indexed GPU buffers.
2. **Blocks CUDA graph capture.** CUDA graph capture works best with fixed-shape tensor state, not request-local Python queues and control flow. The current `_sample_rows()` path in `MossTTSModelRunner` uses per-request Python state machines that cannot be captured in a graph.

## Proposed design

Refactor toward a Higgs-style per-request / row-indexed state pool:

1. **Stable row-indexed GPU buffers** for next-step-critical state: `last_codes_BN`, `delay_count_B`, `eoc_countdown_B`, `is_audio_B`, etc. Each active request owns a row index.
2. **State commit in launch phase**: async launch commits only the next-step-critical state to the GPU buffers.
3. **Output collection in resolve phase**: output row append, final finish handling, and host result collection move to the resolve or journal-apply phase.

This mirrors the `HiggsBatchedSamplerState` pattern where all per-step state is tensorized and indexed by batch row, enabling both async decode overlap and CUDA graph capture.

## Scope

- `sglang_omni/models/moss_tts/model_runner.py` — refactor `_sample_rows()` and `_collect_moss_step()` to use row-indexed buffers
- `sglang_omni/models/moss_tts/request_builders.py` — extract state-pool fields from `MossTTSSGLangRequestData`
- New file: state pool class (e.g., `MossTTSBatchedSamplerState`)

## Acceptance Criteria

- [ ] Decode output bit-identical to current request-local state path
- [ ] State pool uses fixed-shape GPU tensors indexed by batch row
- [ ] Async decode launch/resolve split produces measurable speedup
- [ ] Foundation ready for CUDA graph capture (no Python control flow in the hot loop)
- [ ] Existing unit tests and SeedTTS benchmarks show no regression

## Known Pitfalls

[...]

Issue #734, opened 2026-06-10
[Perf] MOSS TTS: validate and optimize async decode path


Ccyest:
## Motivation

The Higgs TTS pipeline uses `post_decode_launch` / `post_decode_resolve` async hooks to overlap CPU-side work (output collection, state updates) with the next GPU decode step. MOSS TTS does not currently wire into these hooks — all decode-step work runs synchronously.

PR #695 implemented a correctness-first, opt-in async decode path for MOSS. Profiling showed limited speedup because MOSS's current synchronous decode-state design stores next-step feedback in request-local Python objects. The async launch half still has to finish sampling, feedback embedding preparation, and next-step state commit before the following decode step can start — leaving much less resolve-side work to overlap than in Higgs.

## Current state

MOSS TTS `MossTTSModelRunner` in `sglang_omni/models/moss_tts/model_runner.py`:
- `post_process_outputs()` handles all output collection synchronously
- `_collect_moss_step()` extracts channel logits and samples next row in-line
- `_sample_rows()` does delay-pattern state management, multi-codebook sampling, and feedback embedding preparation — all in the synchronous path
- No `post_decode_launch` / `post_decode_resolve` split

## Goal

Validate the async decode path from #695 and identify which operations can be moved to the resolve phase. The key bottleneck is that `_sample_rows` combines next-step-critical state (feedback embeddings) with output-only state (appending to output_rows). Separating these would allow:
- **Launch**: sample tokens, prepare feedback embeddings, commit to decode state
- **Resolve**: append output rows, update finish conditions, host-side result collection

## Acceptance Criteria

- [ ] Async decode produces bit-identical output to synchronous path
- [ ] Measurable latency reduction at concurrency ≥ 2
- [ ] No regression at concurrency = 1
- [ ] Clear profiling data showing launch vs resolve time split

## References

- Higgs async decode: PR #590
- MOSS async decode PR: #695 @guozhihao-224
- State-pool refactor (prerequisite for deeper async optimization): see #637 roadmap
- Tracking issue: #637

PR #745, opened 2026-06-11
[Refactor] MOSS-TTS-Local v1.5 decode-state pool (#735)

JiaxinD:
> Addresses #735; unblocks #736 (interface contract in the "For #736" section below).

Row-indexed GPU decode-state pool for the Local Transformer (v1.5) path, moving
next-step-critical per-request decode state out of request-local Python objects
into stable row-indexed buffers, with output collection moving to a per-step
journal. **Strictly behavior-neutral** (bit-identical outputs, verified by a
fixed-seed gate) apart from one disclosed bugfix. Unblocks #736.

> Scope reality check: the #735 issue text describes the Delay path (32
> codebooks, `_sample_rows`, `MossTTSModelRunner`). The Local v1.5 path differs
> (12 codebooks, 1+12 micro-step frame, binary continue/stop head, frame decode
> already fully CUDA-graphed by #728). This PR targets the Local path only;
> changes are confined to `sglang_omni/models/moss_tts_local/` + tests + one
> standalone gate script. No other model path is touched.

## What moves into the pool

`MossTTSLocalDecodeStatePool` (`state_pool.py`), allocated in
`MossTTSLocalSGLangModel.__init__` before any graph capture, addresses fixed for
process lifetime. `P = max_running_requests + 1` rows (derived from
`_decode_input_embedding.weight.shape[0] + 1` — no literal row count); row `P-1`
reserved as the padding row for #736.

| State | Was | Now |
| --- | --- | --- |
| next-frame feedback embedding | `pending_feedback_queue` deque on each request | `feedback_embeds [P, hidden] bf16`, scattered at end of `_collect_frame`, gathered in `before_decode` |
| sampling params + seed (request-static) | `_param_cache` (composition-keyed single slot) | `[P]` fp32/int64 rows, written once at acquire, gathered per step |
| per-step rows for output collection | `_pending_rows/_pending_embeds` runner-global single slot | `MossTTSLocalDecodeJournal(rids, pool_rows, rows)` attached to `result` each step |

Row lifecycle: acquired on first collect (idempotent by rid), stable across
retraction/resume, released at finish (`result_adapter`) / abort
(`abort_callback`); `reset_row` zeros the row (a fresh/reset row gathers as
zeros — an exact replica of the old empty-queue fallback).

## Commits

1. `tests: fixed-seed sync parity harness` — gate scaffold + S0 CUDA-graph determinism baseline (do-not-modify sentinel).
2. `bugfix: skip chunked prefill rows in post_process_outputs` — **disclosed behavior change** (see below).
3. `state_pool: add MossTTSLocalDecodeStatePool + journal` — pure data structures.
4. `wire pool lifecycle into model + adapters` — acquire/release hinges.
5. `feedback embeddings: queue -> pool rows`.
6. `pending rows -> per-step journal`.
7. `absorb _param_cache into pool rows`.
8. `review: harden journal contract` — raise (not assert) on journal/batch
   misalignment, drop a dead test-only fallback so the journal is the single
   output-collection path, add the resume-overwrite regression test.

Review-round follow-ups (on top of the 8 above):

9. `7ee050b` refactor: pool-level `reset_for_refill` API (per @gaoyang07 / @Ccyest).
10. `2f33b06` docs: trim review-flagged comments, drop the internal codename (per @Ratish1).
11. `090c9cf` perf: **capture-safe GPU radix hash** for generated rows — disclosed behavior change (per @gaoyang07; see below + `docs/design/gpu_radix_hash.md`).
12. `ce41ea2` fix: **skip `generation_steps` for non-final chunked-prefill rows** — deepens the commit-2 chunked guard (per @Ratish1; see below).
13. `ab0dd3c` test: update the stale radix assertion to the post-090c9cf spec + chunked-skip test imports.

## Disclosed behavior changes

**1. Chunked-prefill handling (commit 2 + `ce41ea2`).** The output processor
unconditionally provides `data` for intermediate prefill chunks; pre-this-PR a
mid-prefill chunk's garbage feedback could enter the FIFO queue and surface
first. Commit 2 adds an `is_chunked > 0` guard in `post_process_outputs`
(mirroring Higgs `H/model_runner.py:285`), skipping frame emission for
mid-prefill chunks. `ce41ea2` deepens it (per @Ratish1): sampling here is
positional (`position = generation_steps * num_channels + channel`), so a
mid-prefill chunk could still perturb output by advancing `generation_steps` — a
K-chunk prefill advanced it K times vs once for the no-chunk path, shifting the
final chunk's first real frame off the bit-identical position. A model-agnostic
`ModelRunner.finalize_skip_rids()` hook (default empty), unioned into
`_finalize`'s existing `skip_rids`, suppresses the advance for non-final chunked
rows so only the final chunk advances once. (On v1.5's default config chunked
prefill is structurally unreachable on >=80GB cards — both guards are defensive
+ correct.)

**2. Generated-row radix key algorithm (`090c9cf`).** The per-frame radix key
moved from a host-side blake2b (`build_row_cache_key_ids`) to a capture-safe
tensor-native GPU polynomial hash (`radix_hash.gpu_radix_row_hash`), per
@gaoyang07's review — removing one GPU->CPU sync per decoded frame. Output-
neutral: the key only drives radix cache hit/miss, never the sampled tokens
(proof + the two-layer verification rubric in `docs/design/gpu_radix_hash.md`).
Prompt-row keying keeps blake2b. The pool refactor (commits 3-8) is
behavior-neutral relative to the post-commit-2 baseline.

## Verification

Because the radix key never feeds sampling, the refactor **and** the radix-hash
swap split into two independent layers (rubric: `docs/design/gpu_radix_hash.md`).
Run on H100 at `ab0dd3c`.

**Output layer — bit-identity** (sampling is key-value-independent):

| Check | Result |
| --- | --- |
| S0 frame-decode determinism gate (`test_s0_gate.py::test_s0_graph_replay_is_deterministic`, gpu-marked) | PASS |
| graph-vs-eager sentinel (`test_decode_frame_graphed_matches_branchless_eager`, do-not-modify) | PASS |
| full `moss_tts_local` unit suite on GPU | 72 passed |

**Key layer — GPU radix-hash properties** (`test_radix_hash.py`, on GPU, 7/7):

| Property | Result |
| --- | --- |
| determinism across repeats | PASS |
| collision (adjacent / permuted rows) | PASS |
| EOS preservation (raw `audio_end` kept) | PASS |
| output domain (fold < 151643) | PASS |
| reference correctness + dtype/device-follow | PASS |

**Untrusted re-review** of the pool scatter/gather timing, journal tripwire,
param-reset path, and test set: no behavior diff (hardening folded into the
journal-contract commit).

The hash swap removes one per-frame D2H; the perf delta is below the measurement
noise floor (±1.5% cross-session) and is covered by the existing neutrality table.

## Performance (ABAB, A=merge-base 76b258e vs B=this PR 2ca0069)

Neutral = 95% CI overlap. Paired same-box A/B, ABAB-interleaved, 5 rounds/arm,
generate-only, 50 SeedTTS-EN samples @ concurrency 16, on H100 (NUMA node0:
GPU 3 AR / GPU 4 codec). Hard-isolation harness (server cores 0-23,64-87; client
24-31,88-95; numactl --membind=0). No co-tenant benchmark was active during this
run, and the AR claim-holder was removed (its periodic matmul was a same-card
contention source), so the measured run was effectively isolated on GPU 3+4.

| Metric | A (merge-base) | B (PR-A) | Neutral? |
| --- | --- | --- | --- |
| engine output tok/req-s @ c=16 (frames/s per req) | 100.82 +/- 2.62 | 99.96 +/- 1.27 | YES (CIs overlap; delta 0.85%) |
| SeedTTS-EN rtf_mean @ c=16 | 0.6864 +/- 0.0231 | 0.7143 +/- 0.0443 | YES (CIs overlap) |

The engine-side throughput (the metric PR-A's pool gather/scatter could affect) is
within 0.85% with overlapping CIs (no regression). rtf additionally includes
prefill / queueing / codec, which PR-A does not touch.

## For #736 (@nagisa-kunhah)

The design doc carries a state-pool interface contract (buffer layout
invariants, writer rules per phase, capture-boundary rules). The pool's
`feedback_embeds` deliberately matches `_decode_input_embedding.weight`
dtype/device so the zero-copy aliasing (table weight := pool buffer, decode
input_ids := pool row ids) is a drop-in; the padding row `P-1` is reserved for
done/non-real row routing. New `[P]` fields are ABI-additive.

## Coordination

@guozhihao-224 — #719 is the Delay-path sibling (`MossDecodeStatePool`); naming
here is aligned (`MossTTSLocalDecodeStatePool` / `MossTTSLocalDecodeJournal`).


sglang_omni/models/moss_tts_local/state_pool.py
```diff
@@ -0,0 +1,156 @@
...
+        )
+        self.audio_top_k = torch.zeros(
+            self.num_rows, device=self.device, dtype=torch.int64
+        )
+        self.seeds = torch.zeros(self.num_rows, device=self.device, dtype=torch.int64)
+
+        self._rid_to_row: dict[str, int] = {}
+        self._params_written_rids: set[str] = set()
+        # Real rows 0..P-2 are assignable; the padding row stays out of the
+        # free list so it is never handed to a request.
```
Ccyest, line 72: _params_written_rids is declared with a leading underscore (private), yet model_runner.py accesses it directly.

Fix:
Add two methods on the pool to encapsulate the write-once contract so model_runner never touches _params_written_rids directly.

```
# state_pool.py
def ensure_params(self, row_idx: int, rid: str, data: Any) -> None:
    """Write params if not already written for this acquisition."""
    if rid not in self._params_written_rids:
        self.write_params(row_idx, data)
        self._params_written_rids.add(rid)

def invalidate_params(self, rid: str) -> None:
    """Force params re-write on next ensure_params (retraction path)."""
    self._params_written_rids.discard(rid)

# model_runner.py — _collect_frame
pool.ensure_params(row, rid, sched_req.data)

# model_runner.py — _build_prefill_input_embeds (retraction)
pool.invalidate_params(sched_req.request_id)
pool.reset_row(pool_row)
```

Ccyest:
also, could you please fix lint on this pr

sglang_omni/models/moss_tts_local/request_builders.py
```diff
@@ -436,6 +434,10 @@ def request_builder(payload: StagePayload) -> MossTTSLocalSGLangRequestData:
         return build_sglang_moss_tts_local_request(payload, model=model)
 
     def result_adapter(data: MossTTSLocalSGLangRequestData) -> StagePayload:
-        return apply_sglang_moss_tts_local_result(data.stage_payload, data)
+        result = apply_sglang_moss_tts_local_result(data.stage_payload, data)
+        # Release the finished request's decode-state pool row (mirrors Higgs
+        # request_builders.py:186); recycles the row for a waiting request.
+        model.reset_request(data.stage_payload.request_id)
+        return result
```
edwingao28: If `apply_sglang_moss_tts_local_result` raises, `reset_request` never runs — the scheduler catches the exception (`omni_scheduler.py:761`) and moves on, so the row leaks permanently. Enough failures exhaust the pool and every new request dies at `acquire_row`.

```suggestion
        try:
            return apply_sglang_moss_tts_local_result(data.stage_payload, data)
        finally:
            # Release the finished request's decode-state pool row (mirrors Higgs
            # request_builders.py:186); recycles the row for a waiting request.
            model.reset_request(data.stage_payload.request_id)
```

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -395,21 +373,29 @@ def post_process_outputs(
...
         end_id = int(self.model.config.audio_end_token_id)
-        # rows/embeds are step-private tensors (the graph outputs were
-        # snapshotted in _collect_frame), so per-request views are stable.
-        for row_idx, sched_req in enumerate(scheduler_output.requests):
+        for i, sched_req in enumerate(scheduler_output.requests):
+            # Alignment tripwire: a raise (not a bare assert) so it survives
+            # python -O and matches the model package's raise convention; the
+            # journal is built from this same requests list, so misalignment
+            # only ever surfaces a real bug, never silent frame corruption.
+            if journal.rids[i] != sched_req.request_id:
```
edwingao28: If `scheduler_output.requests` is longer than the journal, `journal.rids[i]` raises a bare `IndexError` before this tripwire fires.

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -311,8 +279,18 @@ def sample_audio(logits: torch.Tensor, channel: int) -> torch.Tensor:
             embeds = self.model._prepare_multi_modal_inputs(
                 rows.to(device=self.model.device)
             )
-        self._pending_rows = rows
-        self._pending_embeds = embeds.detach()
+        row_t = torch.tensor(
+            pool_rows, dtype=torch.long, device=pool.feedback_embeds.device
+        )
+        pool.feedback_embeds[row_t] = embeds.detach().to(
```
edwingao28: The c2 guard skips frame emission, but `post_prefill` calls `_collect_frame` unconditionally, a mid-prefill chunk's garbage embedding is still scattered here. 
It is safe today only because the final chunk overwrites before decode reads; that's an undocumented invariant that #736 and PR-B lookahead could break silently (and the S0 gate doesn't exercise chunked prefill). I suggest we skip the scatter/journal for `is_chunked > 0` rows.

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -311,8 +279,18 @@ def sample_audio(logits: torch.Tensor, channel: int) -> torch.Tensor:
             embeds = self.model._prepare_multi_modal_inputs(
                 rows.to(device=self.model.device)
             )
-        self._pending_rows = rows
-        self._pending_embeds = embeds.detach()
+        row_t = torch.tensor(
```
edwingao28: nit: second `torch.tensor(pool_rows, ...)` upload in this function — build once and reuse to save an H2D copy per step. Optional; ABAB numbers are neutral.

sglang_omni/models/moss_tts_local/state_pool.py
edwingao28: +1 on this

edwingao28:
Leaved some reviews and hope these helps!

sglang_omni/models/moss_tts_local/request_builders.py
Hayden727: Confirmed. This row leak is real because `OmniScheduler.stream_output()` catches result-adapter exceptions and continues after only clearing embed references; it does not call the model abort/reset callback for this finished request. I will patch this with a `finally` cleanup and add a regression so adapter failures cannot exhaust the row pool.

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -94,7 +93,10 @@ def _build_prefill_input_embeds(
                 # stranded by the retraction.
                 generated = torch.stack(data.output_rows, dim=0)
                 rows = torch.cat([rows.to(generated.device), generated], dim=0)
-                data.pending_feedback_queue.clear()
+                pool_row = self.model._state_pool.row_for(sched_req.request_id)
+                if pool_row is not None:
+                    self.model._state_pool.invalidate_params(sched_req.request_id)
+                    self.model._state_pool.reset_row(pool_row)
```
gaoyang07: Building on @Ccyest's suggestion in https://github.com/sgl-project/sglang-omni/pull/745#discussion_r3394136367, I think we can add a pool-level lifecycle API:

```python
def reset_for_refill(self, rid: str) -> bool:
    row_idx = self.row_for(rid)
    if row_idx is None:
        return False
    self.invalidate_params(rid)
    self.reset_row(row_idx)
    return True
```
Then the runner can simply call `self.model._state_pool.reset_for_refill(sched_req.request_id)` here.

Simpler and more intuitive.

sglang_omni/models/moss_tts_local/model_runner.py
gaoyang07, line 271: The full-row radix key is needed for correctness, but this implementation puts a CPU sync in the decode hot path. `_row_radix_token_ids()` calls `build_row_cache_key_ids(rows)`, and that helper always moves `rows` to CPU and hashes with numpy/blake2b. Here `rows` is produced by the GPU local-frame decode, so this forces GPU->CPU sync, Python hashing, then CPU->GPU upload for `next_token_ids` on every generated frame.

Since this PR is preparing the path for the after CUDA Graph / async decode, I think the generated-row radix key should be tensor-native on GPU. We can keep `build_row_cache_key_ids()` for prompt preprocessing, but use a separate GPU hash in `_row_radix_token_ids()` with fixed coefficients/buffers, fold the result below `151643`, and preserve raw `audio_end` for EOS. This keeps the full 13-channel key semantics without introducing a per-step host sync.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -344,6 +344,13 @@ def create_sglang_tts_engine_executor(
         model=model
     )
 
+    def abort_request(request_id: str) -> None:
+        # An aborted request may be mid-preprocessing (drop its prepared
+        # handoff) and/or hold a decode-state pool row (release it); both
+        # legs are idempotent no-ops when the request never reached them.
```
Ratish1: nit: we should trim or remove this comment

sglang_omni/models/moss_tts_local/state_pool.py
```diff
@@ -0,0 +1,166 @@
...
+        # P = max_running_requests + 1; the +1 is the reserved padding row.
+        self.num_rows = int(weight.shape[0]) + 1
+        self.padding_row = self.num_rows - 1
+        self.hidden_size = int(weight.shape[1])
+        self.device = weight.device
+        self.dtype = weight.dtype
+
+        # Feedback embedding for the next decode step (one row per request);
+        # bf16 to match the staging table dtype so before_decode's gather is a
+        # plain copy and the future zero-copy aliasing (#736) is a drop-in.
```
Ratish1: same

tests/unit_test/moss_tts_local/test_pipeline.py
```diff
@@ -643,3 +643,60 @@ def test_encode_audio_stereo_wav_and_mono_fallback():
         np.ones(64, dtype=np.float32) * 0.1, response_format="wav", sample_rate=48000
     )
     assert struct.unpack("<H", mono_blob[22:24])[0] == 1
+
+
+# ---------------------------------------------------------------------------
+# is_chunked guard in post_process_outputs (bugfix)
+# ---------------------------------------------------------------------------
```
Ratish1: lets remove these type of comments aswell

scripts/gate_s0_sync_parity.py
Ratish1, line 1: seems like a local script file which shouldn't be here

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -178,56 +171,29 @@ def _collect_frame(
 
         cfg = self.model.config
         device = hidden_states.device
+        pool = self.model._state_pool
+        pool_rows = []
+        for sched_req in requests:
```
AkazaAkane, line 174: This is still per-frame Python control flow in the decode hot path. The state-pool direction is good, but `_collect_frame()` still loops over `requests` every step to acquire rows, ensure params, build `datas`, create `gen_steps`, collect repetition penalties, and later filter chunked rows.

Suggested follow-up: keep active pool rows and next-step-critical fields like generation_steps as [P] pool tensors, then gather via row_t in the launch path, e.g. gen_steps = pool.generation_steps[row_t], and update those rows as part of state commit. That would make this path closer to CUDA Graph / async-decode readiness while leaving journal/output collection on the resolve side.

tests/unit_test/moss_tts_local/test_parity_harness.py
Ratish1, line 1: same, I think this should be removed aswell?

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -311,8 +274,27 @@ def sample_audio(logits: torch.Tensor, channel: int) -> torch.Tensor:
             embeds = self.model._prepare_multi_modal_inputs(
                 rows.to(device=self.model.device)
             )
-        self._pending_rows = rows
-        self._pending_embeds = embeds.detach()
+        emit_indices = [
```
Ratish1, line 278: This guard is incomplete for chunked prefill. By the time we filter `emit_indices`, `_collect_frame()` has already sampled a frame and published `result.next_token_ids` / `schedule_batch.output_ids` for
  the chunked row.

  Then the shared `ModelRunner._finalize()` increments `data.generation_steps` for every request in the batch, including `is_chunked > 0` rows. That means a middle prefill chunk consumes a generation  step / RNG position while intentionally not appending `output_rows` or writing feedback. The final chunk will then sample from a different base position than the equivalent no-chunk path.

 The skip needs to happen before Local frame production/finalize bookkeeping for chunked rows, or `_finalize()` needs a skip set so `generation_steps` is not advanced for those rows.

sglang_omni/models/moss_tts_local/state_pool.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+class MossTTSLocalDecodeJournal:
+    """Step-private record carrying the frame this step produced to collection.
+
+    Pool rows are overwritten by the same request every step, so they cannot
+    carry the "consume one step later" output data. The journal pins the
+    per-step ``rows`` tensor together with the request ids (for an alignment
+    assertion at apply time) and the pool rows the step touched; it is attached
+    to the step's ``batch_result`` so the async-decode lookahead window (PR-B)
```
Ratish1: here we refer "PR-B" ,  we should try to avoid these kind of wording in docstrings

sglang_omni/models/moss_tts_local/state_pool.py
```diff
@@ -0,0 +1,178 @@
...
+        self.invalidate_params(rid)
+        self.reset_row(row_idx)
+        return True
+
+    def row_for(self, rid: str) -> int | None:
+        """Return ``rid``'s row, or ``None`` if it holds no row."""
+        return self._rid_to_row.get(rid)
+
+
+class MossTTSLocalDecodeJournal:
```
Ratish1, line 156: Should the naming be changed here? . Maybe MossTTSLocalDecodeFrameResult or MossTTSLocalDecodeStepOutput . wdyt @JiaxinD

sglang_omni/models/moss_tts_local/state_pool.py
```diff
@@ -0,0 +1,178 @@
...
+embedding and the request-static sampling parameters/seed) lives in stable,
+process-lifetime GPU buffers indexed by a per-request row. Output-only frame
+collection moves to a per-step :class:`MossTTSLocalDecodeJournal`.
+
+The layout mirrors the Higgs ``HiggsBatchedSamplerState`` row pool and the
+Delay-path ``MossDecodeStatePool`` (#719): ``P = max_running_requests + 1``
+rows, with the last row (``padding_row = P - 1``) reserved — never acquired,
+never request-owned — so a future CUDA-graph that must route non-real/done
+rows somewhere has a stable target (#736). The buffer addresses are fixed for
+the process lifetime; adding a field is ABI-additive.
```
Ratish1: should change these docstring similarly aswell

sglang_omni/models/moss_tts_local/stages.py
JiaxinD: Fixed in 2f33b06: trimmed the `abort_request` rationale comment to two lines.

sglang_omni/models/moss_tts_local/state_pool.py
JiaxinD: Fixed in 2f33b06: condensed the feedback `bf16` comment.

tests/unit_test/moss_tts_local/test_pipeline.py
JiaxinD: Fixed in 2f33b06: removed the section banner.

sglang_omni/models/moss_tts_local/state_pool.py
JiaxinD: Fixed in 2f33b06: replaced the internal `PR-B` codename with `#734`.

sglang_omni/models/moss_tts_local/state_pool.py
JiaxinD: Fixed in 2f33b06: condensed the module and `padding_row` docstring.

sglang_omni/models/moss_tts_local/model_runner.py
JiaxinD: Implemented in 7ee050b: `MossTTSLocalDecodeStatePool.reset_for_refill(rid)` does exactly this (`row_for`, then `invalidate_params` + `reset_row`, returns `False` when the rid holds no row), and `_build_prefill_input_embeds` calls it on the retraction path. Covered by `test_reset_for_refill_clears_active_row` and `test_reset_for_refill_is_noop_for_unheld_rid`. Thanks!

sglang_omni/models/moss_tts_local/model_runner.py
JiaxinD: Fixed in ce41ea2. Root cause: sampling on this path is positional (`position = generation_steps * num_channels + channel`), so a mid-prefill chunk can only perturb output by advancing `generation_steps`. A K-chunk prefill advanced it K times versus once for the no-chunk path, shifting the final chunk's first real frame off the bit-identical position. Fix: a model-agnostic `ModelRunner.finalize_skip_rids()` hook (default empty), unioned into `_finalize`'s existing `skip_rids`; MOSS-TTS Local returns the non-final chunked rids (`is_chunked > 0`), so only the final chunk advances once, matching the single-shot path. Tests: base-hook wiring, chunked-rid selection, chunked-versus-single-shot generation_steps parity; output-layer S0 gate green.

sglang_omni/models/moss_tts_local/model_runner.py
JiaxinD: Done in 090c9cf: a GPU polynomial hash on the generated-row path (`radix_hash.gpu_radix_row_hash`), with blake2b kept for prompt preprocessing, removing the per-frame device-to-host sync. Verified with the two-layer rubric (`docs/design/gpu_radix_hash.md`): output layer bit-identical (S0 plus the graph-versus-eager sentinel) and key layer 7/7 (determinism, collision, EOS, domain). Thanks for pushing this direction.

scripts/gate_s0_sync_parity.py
JiaxinD: Good call. I've folded the S0 gate into a proper regression test under tests/ (gpu-marked, skips without CUDA) and dropped the standalone scripts from the diff. The bit-identity claim in the Verification section now points at a test anyone can run, and the upcoming #734 and #736 changes can reuse the same check instead of reinventing it.

tests/unit_test/moss_tts_local/test_parity_harness.py
JiaxinD: Same here: folded into the gpu-marked regression test (`tests/unit_test/moss_tts_local/test_s0_gate.py`) and dropped the standalone parity harness from the diff.

sglang_omni/models/moss_tts_local/model_runner.py
JiaxinD: Fully agree: that's the second half of the pooling work (orchestration into [P] tensors, not just state). Tracking it as a follow-up after this PR and #734 land; it pairs naturally with the CUDA Graph capture work in #736.

sglang_omni/models/moss_tts_local/state_pool.py
JiaxinD: The journal records which pool rows a step touched, used by the alignment assertion in post_process_outputs and by the #734 lookahead window, so it's bookkeeping rather than the step's output itself. Renaming it to DecodeStepOutput would risk confusion with the actual batch result. That said, naming in this area is your call. If you prefer a rename I'm not attached, just want to keep the bookkeeping versus output distinction visible in whatever name we pick.

Ccyest (approved):

sglang_omni/models/moss_tts_local/model_runner.py
JiaxinD: Opened #757 to track this.

Issue #732, opened 2026-06-10
[Perf] MOSS TTS: batched reference audio encoding


Ccyest:
## Motivation

Once the preprocessing and encoding stages are decoupled (#730), the audio encoder stage will process reference audio one request at a time. The MOSS audio tokenizer's `encode_audios_from_wav()` likely supports batched input (multiple waveforms in a single forward pass), but the current serving path doesn't exploit this.

At concurrency ≥ 2, stacking same-length waveforms into a batched forward pass reduces per-request encoder latency — the same pattern that Higgs #603 / PR #610 delivered for the Higgs DAC encoder.

## Current path

After decoupling (#730), the encoder stage will look like:

```python
# One waveform at a time
def _encode(payload):
    state = load_state(payload)
    ref_codes = processor.encode_audios_from_wav(state.reference_waveform, sample_rate=24000)
    state.reference_codes = ref_codes
    return payload
```

`SimpleScheduler` collects up to `max_batch_size` messages, but without `batch_compute_fn` each request is still encoded serially.

## Proposed change

1. Add a `_encode_batch` function that stacks waveforms and runs a single batched codec forward pass.
2. Use **bucketing by waveform length** rather than zero-padding to avoid boundary artifacts from the non-causal Conv1d encoder (same issue Higgs encountered in #603).
3. Wire `batch_compute_fn=_encode_batch` into the audio encoder's `SimpleScheduler`.

### Known Pitfalls

- **Non-causal encoder requires bucketing.** Zero-padding shorter waveforms corrupts codes near the boundary. Only same-length waveforms should be batched together.
- **Minimum length requirement.** Short clips may need padding to meet the codec's minimum input length.
- **Dependency**: requires #730 (decouple preprocessing) to land first.

## Acceptance Criteria

- [ ] `_encode_batch([single_waveform])` produces bit-identical output to the single-encode path
- [ ] Multi-item batches with same-length waveforms produce identical codes to individual encodes
- [ ] Encoder throughput strictly higher than serial baseline at concurrency ≥ 2
- [ ] No regression at concurrency = 1
- [ ] Existing unit tests still pass

## References

- Higgs batched encoder: #603, PR #610
- Prerequisite: #730 (decouple preprocessing)
- Previous attempt: #649 @AkazaAkane
- Tracking issue: #637

SandyLuXY:
Hi @Ccyest, I'm interested in working on the issue.

AkazaAkane:
[...]

PR #749, opened 2026-06-11
[Feat] Add batched reference audio encoding for MOSS-TTS Local

SandyLuXY:
## Motivation

MOSS-TTS Local voice-cloning requests repeatedly encode reference audio before generation. Under concurrent serving, those independent reference encodes can bottleneck preprocessing even though the processor can batch the path-based reference encode work. This PR batches compatible reference paths, preserves per-request fallback behavior, and keeps the existing cache protections intact.

## Modifications

1. `sglang_omni/models/moss_tts_local/stages.py`
   - Extends `_BatchedReferenceEncoder` so concurrent `encode()` calls are drained into a single reference encode batch up to `max_batch_size`.
   - Prefers the processor's `encode_audios_from_path` API, which already handles path loading, resampling, padding, and batched codec encode for mixed reference lengths.
   - Deduplicates repeated paths within a drained batch while preserving each caller's result ordering.
   - Validates batched result counts and falls back to per-path encode so malformed batch output or one bad reference does not fail unrelated requests.
   - Keeps waveform encode support as a fallback for processors without the path API, including target sample-rate discovery and per-item retry after a waveform batch failure.
   - Memoizes the reference-duration gate by stable content key inside `CachedReferenceEncoder`, while still checking duration before cache or inflight reuse for uncached content.

2. `tests/unit_test/moss_tts_local/test_pipeline.py`
   - Adds coverage for stage wiring, concurrent batch coalescing, duplicate-path deduplication, order preservation, malformed batch output fallback, no-path waveform fallback, and path-preferred behavior when both APIs are present.
   - Covers cache-duration memoization by content key so overwritten references are rechecked while repeated identical references avoid extra `torchaudio.info` calls.

## Related Issues
#732 

## Accuracy Test

**Setup:**

- Compared refs: baseline `main @ c8ea7ad`; candidate `feat/moss-local-batched-audio-encode` based on `0b551f8`
- Dataset/eval: `zhaochenyang20/seed-tts-eval-arrow`, `lang=en`, 1088 samples / 666 unique audio files, `--ref-format references`, `--token-count auto`.
- Hardware: 2x H100

All rows completed with `quality_status=ok`, `1088/1088` WER samples evaluated, `1088/1088` similarity samples evaluated, and zero failed requests.

| concurrency | main WER | current WER  | main speaker similarity | current speaker similarity | 
| ---: | ---: | ---: | ---: | --- |
| 1 | 2.19% | 2.75% | 65.575 | 65.369 | 
| 16 | 2.46% | 2.15% | 65.536 | 65.451 | 

## Benchmark & Profiling

| ---: | --- | ---: | ---: | --- |
| 1 | mean latency (s) | 0.628 | 0.634 | +1.0% |
| 1 | p95 latency (s) | 0.851 | 0.853 | +0.2% |
| 1 | mean RTF | 0.1453 | 0.1466 | +0.9% |
| 1 | audio throughput (s/s) | 7.002 | 6.943 | -0.8% |
| 1 | request throughput (QPS) | 1.591 | 1.578 | -0.8% |
| 16 | mean latency (s) | 1.499 | 1.567 | +4.5% |
| 16 | p95 latency (s) | 1.920 | 2.012 | +4.8% |
| 16 | mean RTF | 0.3577 | 0.3737 | +4.5% |
| 16 | audio throughput (s/s) | 46.767 | 44.704 | -4.4% |
| 16 | request throughput (QPS) | 10.629 | 10.158 | -4.4% |

### Summary:

- The latest batch-only c1/c16 comparison fails the throughput gate at both tested concurrencies.
- c1 regresses slightly at `-0.8%` audio throughput with zero failed requests.
- c16 regresses by `-4.4%` audio throughput with zero failed requests.
- The exploratory `ref_audio_cache_max_items=1024` comparison is retained below for context only; it is not the fair batch-only comparison because `main` keeps the `256` cache-item cap.

## Reference Audio Cache Capacity Variant

This earlier exploratory run used `ref_audio_cache_max_items=1024` on the candidate side while baseline `main` used `256`; `ref_audio_cache_max_bytes` stayed at `64 MiB`. Because this changes cache capacity and batched reference encoding together, the tables below are context only and are not the batch-only result used for the PR gate.

**Setup:**

- Compared refs: baseline `main @ c8ea7ad`; candidate artifact `0b551f8`.
- Hardware: `2x H100` 
- All rows completed `1088/1088` TTS requests with zero failed requests

| concurrency | main WER | current WER |  main speaker similarity | current speaker similarity |
| ---: | ---: | ---: | ---: | --- |
| 1 | 2.26% | 2.53% |  65.464 | 65.383 |
| 16 | 2.03% | 2.26% |  65.171 | 65.409 |

Higher audio s/s and QPS are better; lower latency and RTF are better.

| 1 | mean latency (s) | 0.627 | 0.618 | -1.4% |
| 1 | p95 latency (s) | 0.843 | 0.833 | -1.1% |
| 1 | mean RTF | 0.1451 | 0.1430 | -1.4% |
| 1 | audio throughput (s/s) | 7.026 | 7.121 | +1.4% |
| 1 | request throughput (QPS) | 1.597 | 1.618 | +1.3% |
| 16 | mean latency (s) | 1.742 | 1.424 | -18.3% |
| 16 | p95 latency (s) | 2.246 | 1.843 | -18.0% |
| 16 | mean RTF | 0.4149 | 0.3380 | -18.5% |
| 16 | audio throughput (s/s) | 40.326 | 49.669 | +23.2% |
| 16 | request throughput (QPS) | 9.163 | 11.285 | +23.2% |

## Checklist

- [x] Add unit tests.

## CI


zhaochenyang20:
Great job so far!

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -197,26 +199,168 @@ def _drain_batch(self) -> list[tuple[str, concurrent.futures.Future]]:
...
-                    try:
-                        results[path] = self._processor.encode_audios_from_path([path])[
-                            0
-                        ]
-                    except Exception as exc:
-                        results[path] = exc
+            if self._path_encoder() is not None:
+                results = self._encode_paths_with_fallback(unique_paths)
+            else:
+                results = self._encode_wavs_from_paths_with_fallback(unique_paths)
```
AkazaAkane: This seems to diverge from the core safety constraint in #732. The issue explicitly says the non-causal Conv1d encoder should batch only same-length waveforms because zero-padding shorter waveforms can corrupt boundary codes.

Here the worker always prefers `encode_audios_from_path()` when available, and the class docstring says the path API forwards variable-length refs through padded `batch_encode`. Can we either prove with a real MOSS tokenizer parity test that `encode_audios_from_path([a, b])` produces exactly the same codes as separate `encode_audios_from_path([a])` / `encode_audios_from_path([b])` for unequal lengths, or keep exact-length bucketing before invoking the batched encode path?

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -269,18 +413,21 @@ def __init__(
...
         # trust_stat left False (review feedback): keep the sentinel byte-read so a
         # same-size+mtime+ctime overwrite cannot stale-hit. The flag stays available
         # in reference_path_cache_key for deployments that guarantee immutable refs.
         key = _reference_path_cache_key(path)
         if key is None:
+            _BatchedReferenceEncoder._check_reference_duration(path)
             return self._encoder.encode(path)  # uncacheable (URL/missing) -> bypass
+        # The key is computed first only to memoize the duration gate by stable
+        # file content. The gate still runs before cache/inflight access.
+        self._check_reference_duration_once(path, key)
```
AkazaAkane: This weakens the previous “duration gate first” protection. `CachedReferenceEncoder.encode()` now computes `_reference_path_cache_key(path)` before checking duration, and `reference_path_cache_key()` can call `path.read_bytes()` to hash the full file.

That means a >100s local reference can still trigger a full file read/hash before being rejected. Can we preserve the early resource gate, either by running `_check_reference_duration()` before the full-content cache key or by using a cheaper duration-check memo key that does not require full-file hashing first?

AkazaAkane (commented):
I think the design might somehow contradict to #732 . Please correct me if i am wrong.

The PR adds _BatchedReferenceEncoder, not an explicit _encode_batch. It batches paths through encode_audios_from_path() when available, or batches loaded wavs in a fallback.

The PR explicitly says the path API forwards variable-length refs through padded batch_encode, and says exact-length splitting creates many tiny calls. That is the opposite of the issue’s bucketing requirement.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -197,26 +199,168 @@ def _drain_batch(self) -> list[tuple[str, concurrent.futures.Future]]:
                 break
         return batch
 
+    def _target_sample_rate(self) -> int:
```
gaoyang07, line 181: This feels a bit too defensive for just getting the sample rate...Can we make `processor.model_config.sampling_rate` an explicit contract here instead of probing multiple fallback owners?

The MOSS-TTS Local processor already treats `model_config.sampling_rate` as the authority in `encode_audios_from_wav()` / `encode_audios_from_path()`. If it is missing or invalid, this processor instance is not in a valid state for this path, so silently falling back to `audio_tokenizer`, `audio_tokenizer.config`, or `48000` makes the behavior harder to reason about and can hide configuration bugs.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -197,26 +202,173 @@ def _drain_batch(self) -> list[tuple[str, concurrent.futures.Future]]:
...
+        try:
+            value = getattr(model_config, "sampling_rate", None)
+        except Exception:
+            value = None
+        try:
+            sample_rate = int(value)
+        except (TypeError, ValueError):
+            sample_rate = 0
+        if sample_rate <= 0:
+            raise RuntimeError(
```
Hayden727, line 195: Made `processor.model_config.sampling_rate` an explicit contract for waveform reference encoding instead of silently falling back to tokenizer state or 48000. That makes invalid processor state fail fast rather than encoding with a hidden sample-rate assumption.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -197,26 +202,173 @@ def _drain_batch(self) -> list[tuple[str, concurrent.futures.Future]]:
...
+
+        if load_failed_paths and self._path_encoder() is not None:
+            results.update(self._encode_paths_with_fallback(load_failed_paths))
+
+        if not loaded:
+            return results
+
+        buckets: dict[int, list[tuple[str, torch.Tensor]]] = {}
+        for path, wav in loaded:
+            buckets.setdefault(int(wav.shape[-1]), []).append((path, wav))
```
Hayden727, line 317: Fixed the #732 safety issue by bucketing loaded/resampled refs on exact waveform length before calling the wav batch encoder. This avoids mixed-length padding reaching the non-causal codec encoder. Validated on novita-h100-omni with `CUDA_VISIBLE_DEVICES=0 pytest -q tests/unit_test/moss_tts_local/test_pipeline.py -q` (48 passed) at e9991151.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -269,26 +421,67 @@ def __init__(
...
+        self._duration_checked_keys: dict[str, None] = {}
+        self._duration_checked_max_items = max_items * 4
 
     def encode(self, path: str) -> torch.Tensor:
         path = str(path)
-        # Note(Jiaxin): duration gate runs first — a >100 s ref must never reach
-        # the cache or the inflight dict.
-        _BatchedReferenceEncoder._check_reference_duration(path)
+        duration_key = self._duration_memo_key(path)
+        self._check_reference_duration_once(path, duration_key)
```
Hayden727, line 441: Restored the duration gate ahead of `_reference_path_cache_key()`, so over-limit local refs are rejected before the full-file cache hash/read path. The regression test monkeypatches `_reference_path_cache_key` and asserts it is not called for a >100s ref. Same H100 validation: 48 passed at e9991151.

sglang_omni/models/moss_tts_local/config.py
```diff
@@ -21,7 +21,7 @@ def _stages(*, codec_device: str) -> list[StageConfig]:
             factory_args={
                 "device": codec_device,
                 "ref_audio_cache": True,
-                "ref_audio_cache_max_items": 256,
+                "ref_audio_cache_max_items": 1024,
```
Hayden727, line 24: Re-applied the cache item cap increase to 1024 because the author's SeedTTS run shows the batch-only variant regressed while the larger ref cache was the observed win for the 600+ unique refs workload. This still needs a fair c1/c16 SeedTTS rerun before making a final perf claim; unit validation passed on H100 at e9991151.

Hayden727 (commented):
Pushed fix commit e9991151 to the PR branch.

Validation:
- Local: py_compile, git diff --check, black --check, isort --check-only.
- Remote H100 (novita-h100-omni, container sglang-omni-hayden-moss-ref-encode): `CUDA_VISIBLE_DEVICES=0 pytest -q tests/unit_test/moss_tts_local/test_pipeline.py -q` -> 48 passed.

Note: I did not rerun the full SeedTTS c1/c16 benchmark. The cache-cap change is included because the reported data points to cache capacity as the win, but the PR still needs a fair perf rerun before claiming throughput improvement.

Hayden727:
Full SeedTTS EN generate-only validation on the pushed fix commit `e9991151`:

Environment:
- Host/container: `novita-h100-omni` / `sglang-omni-hayden-moss-ref-encode`
- Model/config: `OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5`, `examples/configs/moss_tts_local.yaml`
- Workload: `zhaochenyang20/seed-tts-eval-arrow`, `lang=en`, 1088 samples, 666 unique ref audio files
- Method: cold server restart per concurrency, warmup 10, generate-only, no ASR/WER phase
- GPUs: `CUDA_VISIBLE_DEVICES=2,3`

Results:

| concurrency | completed | failed | latency mean / p95 / p99 | RTF mean / p95 | throughput | audio throughput |
| --- | ---: | ---: | --- | --- | ---: | ---: |
| 1 | 1088 | 0 | 0.652s / 0.906s / 1.128s | 0.1509 / 0.1891 | 1.534 req/s | 6.750 audio-s/s |
| 16 | 1088 | 0 | 2.259s / 3.275s / 4.778s | 0.5379 / 0.8752 | 7.056 req/s | 31.056 audio-s/s |

Artifact dirs:
- `/data/perf-baselines/sglang-omni/20260612-153312-novita-h100-omni-moss-local-pr749-fix-c1-e9991151`
- `/data/perf-baselines/sglang-omni/20260612-155659-novita-h100-omni-moss-local-pr749-fix-c16-e9991151`

This validates the fixed branch on the full 666-ref workload with zero request failures. I am not treating this as a formal before/after claim against the previous PR head or base branch yet; that would still need a matching run on the old commit with the same host/GPU/concurrency/server settings.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -134,13 +135,14 @@ class _BatchedReferenceEncoder:
     Each request needs its reference run through the ~1B-param codec encoder
     (~0.25 GPU-seconds). The preprocessing workers call :meth:`encode`
     concurrently; a single daemon thread drains the queue and encodes up to
-    ``max_batch_size`` files in one ``batch_encode`` forward, which costs
-    barely more than a single encode. Failures fall back to per-item encodes
-    so one bad file only fails its own request.
+    ``max_batch_size`` files. When waveform encoding is available, it
```
yxs, line 114: FYI.
I ran a direct control on H100 with the real `OpenMOSS-Team/MOSS-Audio-Tokenizer-v2` (the same codec serving uses): take one 2.0s reference A and encode it (1) alone, (2) batched with an equal-length 2.0s neighbor D (no padding), and (3) batched with a 5.3s neighbor B (A padded to 5.3s). The result is decisive — (2) and (3) produce **bit-identical** codes (both differ from the alone-encoding in 41/300 entries, on the **exact same** frames `[5,11,18,21,22,23,24]`), and it reproduces (repeat run is bitwise identical). This proves three things: ① padding has **zero** effect on valid frames (otherwise the padded case (3) would differ more than the unpadded case (2)); ② there is **no cross-item leakage** (swapping neighbor D for B leaves A's codes unchanged); ③ what actually changes the codes is **batch size itself** — bs=1 canonical, bs=2 drifts in 41 entries, bs=3 in 47, a bf16 + SDPA-fallback numerical effect across batch shapes that has nothing to do with padding.

AkazaAkane (commented):
I conduct further test on the padding issue. FYI @yxs @Hayden727 

<html><head></head><body><h2>Padding / batching control result</h2><h3>Environment</h3>

-  GPU: NVIDIA H100 80GB HBM3, GPU0 only
-  Model: OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5
-  Sample rate: 48000
-  Torch: 2.9.1+cu128
-  Backend warning: flash-attn unavailable, fell back to SDPA, bf16
-  A length: 2.0s, 96000 samples
-  B/Lmax: 5.3s, 254400 samples
-  Valid frames for A: 25, entries compared: 25 * 12 = 300

<hr><h3>PR-path probe</h3><p>The PR-path probe passed.</p><p><code>_BatchedReferenceEncoder</code> bucketed exact waveform lengths as:</p><pre><code class="language-text">[11, 11] and [17]
</code></pre><p>No mixed-length wav batch was observed.</p><hr><h2>Interpretation</h2><p>This result separates two issues:</p><ol><li><p><strong>Batch-size drift exists</strong>, but it is relatively small in this test.</p><p>Same-length batching changes A by <code>18 / 300</code> entries compared with solo encoding:</p><pre><code class="language-text">bs=1 solo A vs bs=2 same-length [A, D]     18 / 300
bs=1 solo A vs bs=3 same-length [A,D,D2]   18 / 300
</code></pre><p>This is likely the bf16 + SDPA batch-shape numerical drift.</p></li><li><p><strong>Long zero-padding is not harmless in the direct tokenizer test.</strong></p><p>Padding A from 2.0s to 5.3s changes A’s valid-frame codes substantially:</p><pre><code class="language-text">zero-pad [A_zero, B] vs bs=2 same-length [A, D]   245 / 300
zero-pad [A_zero, B] vs bs=1 solo A               244 / 300
</code></pre><p>This is far larger than the same-length batch-size drift of <code>18 / 300</code>.</p></li><li><p><strong>The fixed-shape controls are the strongest evidence.</strong></p><p>In these tests, batch size, tensor shape, dtype, backend, and neighbor setup are fixed. Only the padded tail content changes.</p><pre><code class="language-text">fixed shape: zero vs replicate   158 / 300
fixed shape: zero vs reflect     245 / 300
fixed shape: zero vs random      245 / 300
</code></pre><p>Therefore, the result does not support the stronger claim that “padding has zero effect on valid frames.” In this setup, changing only the padded tail can substantially change A’s valid-frame codes.</p></li><li><p><strong>Padding length matters.</strong></p><p>Short padding did not affect this sample:</p><pre><code class="language-text">Lmax=2.01s zero-pad vs same-length bs=2   0 / 300
Lmax=2.5s zero-pad vs Lmax=2.01s          0 / 300
</code></pre><p>But longer padding did:</p><pre><code class="language-text">Lmax=5.3s zero-pad vs Lmax=2.01s          245 / 300
Lmax=10s zero-pad vs Lmax=2.01s           259 / 300
</code></pre><p>So the correct conclusion is not “zero padding always corrupts codes,” but rather: long padding can affect valid-frame codes for this tokenizer/backend setup.</p></li></ol><hr><h2>Conclusion</h2><p>The current PR routing looks safe because <code>_BatchedReferenceEncoder</code> buckets exact waveform lengths before wav batch encode, and the probe did not observe mixed-length wav batching.</p><p>However, the direct tokenizer test does <strong>not</strong> support the stronger claim that zero padding has zero effect on valid frames. In this synthetic case, long padding and fixed-shape tail changes affected A’s valid-frame codes substantially and reproducibly.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -197,26 +202,173 @@ def _drain_batch(self) -> list[tuple[str, concurrent.futures.Future]]:
...
+                        try:
+                            results[path] = self._encode_one_path(path)
+                        except Exception as path_exc:
+                            results[path] = path_exc
+        return results
+
+    def _encode_path_batch(self, paths: list[str]) -> dict[str, Any]:
+        if self._wav_encoder() is not None:
+            return self._encode_wavs_from_paths_with_fallback(paths)
+        return self._encode_paths_with_fallback(paths)
```
zhaochenyang20: When the wav API is unavailable, do not batch multiple paths unless the real tokenizer proves that `encode_audios_from_path([a, b])` is bit-identical to per-item encoding for unequal lengths.

We need an experiment on this.

sglang_omni/models/moss_tts_local/config.py
zhaochenyang20: Re-isolate the cache-cap change from the batching change: at minimum run `main@same-base + cap=256`, `candidate + cap=256`, and `candidate + cap=1024` on the same host/GPU with cold starts, the same concurrency, and the same workload. If the real win is cache capacity, retitle or split the PR as a cache-cap optimization; if batching still regresses, keep batching off by default or restrict it to proven-benefit scenarios.

In short, we need to do dedicated experiments to see whether the improvements came from, or does it has no improvement. If it's not, we can close it.

zhaochenyang20 (changes requested):

SandyLuXY:
Thanks @yxs @AkazaAkane for the investigation!

I double checked whether mixed-length batched reference encoding is bit-identical to per-item encoding on the seedtts data. **TL;DR is `encode_audios_from_path()` is **not bit-identical** to per-item encoding for either equal-length or unequal-length references.**

```python
processor.encode_audios_from_path([a, b])
```

vs.

```python
[
    processor.encode_audios_from_path([a])[0],
    processor.encode_audios_from_path([b])[0],
]
```

## Setup

- Dataset: SeedTTS EN reference WAVs
- GPU: H100
- Equality check: exact `torch.equal(...)` plus raw tensor SHA256
- Mismatch rate: element-level code mismatch over `[T, 12]` codec-token tensors

Backends tested:
- **SDPA**: `flash_attn` intentionally blocked
- **FA2**: isolated `flash-attn==2.8.3` overlay, with FA2 symbols available

## Results

**80-pair run across diversified frame-delta buckets:**

| Backend | Batch-vs-Single Code Mismatch | Single-Repeat Sanity Check |
| --- | ---: | ---: |
| SDPA | 5.800402% | 0.000000% |
| FA2 | 5.768386% | 0.000000% |

`Batch-vs-Single Code Mismatch` counts unequal code elements across the four comparisons per pair: `[a, b][0]` vs `[a]`, `[a, b][1]` vs `[b]`, `[b, a][0]` vs `[b]`, and `[b, a][1]` vs `[a]`. `Single-Repeat Sanity Check` compares repeated per-item encodes, e.g. `encode([a])` vs `encode([a])` again.

Length-difference correlation was effectively absent:

| Backend | Pearson(delta, mismatch) | Spearman(delta, mismatch) |
| --- | ---: | ---: |
| SDPA | 0.0217 | -0.0456 |
| FA2 | 0.0854 | 0.0785 |

So mismatch rate is not simply “larger duration gap => more mismatch.”

**Same-length check:**

- 12 same-length SeedTTS reference pairs, 3 repeats each
- each pair had identical frame count and sample rate across `a` and `b`
- checked both `encode_audios_from_path(...)` and `encode_audios_from_wav(...)`
- compared `[a, b]` and `[b, a]` outputs against per-item outputs

| Backend | Input API | Single-Repeat Check | Batch-vs-Single Comparisons | Batch Mismatch Rate |
| SDPA / no FA2 | path | 72/72 passed | 0/144 passed | 5.788982% |
| SDPA / no FA2 | wav | 72/72 passed | 0/144 passed | 5.788982% |
| FA2 | path | 72/72 passed | 0/144 passed | 7.148693% |
| FA2 | wav | 72/72 passed | 0/144 passed | 7.148693% |

Additional observations from the same-length runs:
- path and wav APIs produced identical tensor digests within each backend
- batched output for each item was stable across `[a, b]` vs `[b, a]`
- all single-item repeat checks were exactly stable

## Hypothesis
This is the hypothesis of the bit-level mismatch after digging into the tokenizer code:

Every `nn.Linear` / `WNConv1d(k=1)` in the encoder and quantizer is a GEMM whose M dimension is `B * sequence_length`. Different M causes cuBLAS to select a different kernel configuration with a different floating-point accumulation order, producing results that differ at the ULP level. All other operations (norms, RoPE, activations, patching) are strictly per-token and shape-invariant — the GEMMs are the sole source of divergence. The residual VQ's hard `argmin` then amplifies these sub-ULP encoder differences into entirely different discrete codes.

zhaochenyang20:
Great job so far. I think this turned out really well — nice work. And the more I sit with it, the more it makes sense: this encoder is fundamentally a neural encoder, not a symbolic one, so it's naturally going to exhibit this kind of compute drift from GEMM-shape differences. That makes being mindful of batch invariance a genuinely important thing here. That said, I don't think we should actually go build any batch-invariance machinery for this encoder — that would almost certainly be a net negative.
One small suggestion: you could write up a short piece — a Xiaohongshu post or a LinkedIn article — walking through the thinking behind this PR. Even though we ended up not adding batching to the encoder, the whole line of exploration was really worthwhile in its own right. It'd be great to tell the story of how this PR played out from start to finish. I'm pretty sure the RL crowd would enjoy reading it and playing around with it. haha

PR #753, opened 2026-06-11
[Feat] Moss tts local streaming

### LINKED ISSUE #739: [Feat] MOSS TTS: AR stream chunk + streaming vocoder decode
## Motivation

MOSS TTS currently generates all audio codes in the AR engine, then sends the complete `delayed_audio_codes` tensor to the vocoder for a single batch decode. This means the user hears nothing until the entire utterance is generated — first-chunk latency equals full generation latency.

Streaming TTS breaks the AR output into chunks and vocalizes them incrementally, so the user hears the first audio chunk while the AR engine is still generating subsequent frames. This dramatically reduces perceived latency for long utterances.

Higgs TTS already implements this pattern (PR #597, PR #614) with streaming TTS schedulers that emit partial audio chunks as they become available.

## Proposed design

Two components:

### 1. AR stream chunking

The AR decode loop accumulates output rows in `MossTTSModelRunner.post_process_outputs()`. Instead of waiting for EOS, emit a chunk every N frames (configurable, e.g., 10–20 frames at 12.5 fps = 0.8–1.6s of audio per chunk):

- After every N new rows, package the chunk's `delayed_audio_codes` and send to vocoder
- Handle delay-pattern correctly: the first chunk needs `n_vq - 1` extra frames for the delay pattern to stabilize
- Mark chunks as intermediate (not final) to distinguish from the last chunk

### 2. Streaming vocoder decode

The vocoder stage processes each chunk as it arrives rather than waiting for the complete sequence:

- Decode each chunk independently or maintain a rolling context window
- Stream the decoded PCM to the client via SSE or raw PCM HTTP streaming
- Handle chunk boundaries to avoid audible artifacts (overlap-add or crossfade at boundaries)

### API surface

The existing streaming response format (SSE with base64-encoded PCM chunks) is already supported at the API level. The implementation needs to wire the AR → vocoder → HTTP response pipeline to emit chunks incrementally.

## Acceptance Criteria

- [ ] First audio chunk delivered within N frames of AR generation start (not at EOS)
- [ ] Streaming output sounds identical to non-streaming (no audible chunk boundary artifacts)
- [ ] RTF at concurrency = 1 not degraded vs non-streaming path
- [ ] Time-to-first-audio strictly lower than non-streaming path for utterances > 2s
- [ ] Both SSE and raw PCM streaming formats supported
- [ ] Non-streaming path still works as before (streaming is opt-in)

## Known Pitfalls

- **Delay-pattern chunk boundaries.** MOSS uses a 32-codebook delay pattern. The first chunk needs enough frames for the delay to settle; cutting too early produces silent/garbage codebook channels. Need to compute the minimum chunk size as `chunk_frames + n_vq - 1`.
- **Vocoder boundary artifacts.** ConvTranspose decoders can produce discontinuities at chunk boundaries. May need overlap-add or a context window that extends into the previous chunk.
- **Per-chunk overhead.** Each chunk requires a separate vocoder forward pass. If chunks are too small, the per-call overhead dominates. Need to tune chunk size for the latency/overhead tradeoff.

## References

- Higgs streaming TTS: PR #597, PR #614
- Existing streaming API: SSE + raw PCM response format in `sglang_omni/serve/openai_api.py`
- Existing PR: #674 @BruceLoveDecimal
- Tracking issue: #637

edwingao28:
## Motivation
MOSS-TTS-Local (v1.5) only decodes offline: the client gets audio after the full generation finishes. Its codec (MOSS-Audio-Tokenizer-v2) is natively streamable — causal convs + bounded-context attention with per-slot state — so unlike sliding-window retrofits over offline codecs, chunked output is bit-identical to offline decode by construction: no chunk-boundary timbre/loudness seams and no overlap re-decode cost. 
This PR wires MOSS-TTS-Local into the streaming TTS API.

## Modifications

  - `streaming_vocoder.py` (new): `MossTTSLocalStreamingVocoderScheduler` + `_CodecStreamSession` — one long-lived batched `codec.streaming()` session. Streaming requests hold per-request slots (exec-masked uniform steps, reset-on-release for reuse); near-due streams coalesce into one decode step; one batched D2H per step over active slots only. Non-streaming decodes go through dedicated offline slots of the same session once it exists (the codec forbids nested streaming contexts); before that, the pre-existing `processor.decode_audio_codes` path is unchanged. `no_grad` around session step/reset; a failed step fails only that stream, not batch peers.
  - `model_runner.py`: `post_process_outputs` forwards each generated row of requests carrying `stream_metadata` to the vocoder stage via the scheduler outbox — one D2H per step regardless of how many requests stream.
  - `request_builders.py`: `build_moss_tts_local_stream_metadata` — per-request stream contract (`n_vq`, `initial_codec_chunk_frames`).
  - `config.py`: `stream_to=["vocoder"]` on the engine stage; `can_accept_stream_before_payload` on the vocoder stage.
  - `stages.py`: `create_vocoder_executor` returns the streaming scheduler; the old `SimpleScheduler` offline batch loop moved into its offline lane.
  - Tests: `test_streaming_vocoder.py` — stateful fake codec; streamed PCM must equal offline decode exactly; slot isolation/reuse/exhaustion fallback; abort; offline-lane equivalence with and without a live session.

## Related Issues

close #739

### 1. TTFA (time to first audio payload, with reference audio)

| concurrency | non-stream (= full latency) | stream TTFA | speedup |
|---|---|---|---|
| c=1 (EN, 100) | 0.70 s | **0.126–0.133 s** | 5.3× |
| c=4 (EN, 400) | 0.83 s | 0.56 s | 1.5× |
| c=8 (EN, 400) | 1.48 s | 1.63 s | — |
| c=16 (EN, 400) | 2.93 s | 3.58 s | — 

### 2. Quality (seed-tts-eval, corpus WER, seed 42 both modes)

| run | non-stream | stream | Δ (on−off) |
|---|---|---|---|
| EN c=1 (100) | 1.911 % | 1.456 % | −0.455 pp |
| ZH c=1 (100) | 0.633 % | 1.002 % | +0.369 pp |
| EN full (1088, c=16) | 2.738 % | 2.529 % | −0.209 pp |
| ZH full (2020, c=16) | 1.512 % | 1.313 % | −0.199 pp |
| EN c=8 (400) | 2.506 % | 2.624 % | +0.118 pp |
| EN c=16 (400) | 2.436 % | 2.296 % | −0.140 pp |

| | c=1 | c=4 | c=8 | c=16 |
|---|---|---|---|---|
| qps off | 1.42 | 4.79 | 5.35 | 5.37 |
| qps on | **1.72** | 3.68 | 3.68 | 3.81 |
| qps Δ | **+21 %** | −23.3 % | −31.1 % | −29.1 % |
| rtf on/off | 0.144/0.171 | 0.263/0.199 | 0.530/0.363 | 1.024/0.720 |
| latency on/off (s) | 0.58/0.70 | 1.08/0.83 | 2.15/1.48 | 4.13/2.93 |
| stream TTFA (s) | 0.13 | 0.56 | 1.63 | 3.58 |

## Checklist

- [x] Add unit tests.

## CI


Hayden727 (commented):
Leaving two inline comments on streaming chunk-contract and session lifecycle issues.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -0,0 +1,631 @@
...
+                for entry in slotted
+                if len(entry[1].pending) >= entry[1].threshold
+            ]
+            if not due:
+                return
+            floor = min(
+                min(len(state.pending) for _, state in due),
+                join_floor,
+            )
+            participants = [
```
Hayden727, line 489: [P2] This coalescing path can emit a stream before that request has reached its own threshold. For example, a request with `initial_codec_chunk_frames=0` is supposed to skip the small first chunk and wait for `_stream_chunk_frames`, but once another request is due, `floor` can be `join_floor` (default 5) and this request becomes a participant with only 5 pending frames. That breaks the explicit request contract and raw PCM first-chunk behavior under concurrency. Please keep participant eligibility tied to each stream's threshold, or add a separate opt-in for below-threshold riders plus a regression covering explicit zero.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -0,0 +1,631 @@
...
+    """
+
+    def __init__(self, codec: Any, *, stream_slots: int, offline_slots: int) -> None:
+        self._codec = codec
+        self._stream_slots = int(stream_slots)
+        self._offline_slots = int(offline_slots)
+        self._batch_size = self._stream_slots + self._offline_slots
+        self._device = next(codec.parameters()).device
+        self._free_stream_slots = list(range(self._stream_slots))
+        self._exit_stack = contextlib.ExitStack()
```
Hayden727, line 83: [P2] `_CodecStreamSession` enters `codec.streaming(...)`, but nothing closes the `ExitStack`; `MossTTSLocalStreamingVocoderScheduler` inherits `stop()` which only flips `_running`. If a stage is stopped/restarted in the same process, this codec instance can remain in streaming mode with stale per-slot state/resources, and later offline decode/session creation can hit the nested streaming guard this PR is working around. Please add a session `close()` and call it from scheduler `stop()`/teardown while clearing stream state.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
Hayden727: Fixed in `6af81d81` by keeping explicit first-chunk opt-out streams out of below-threshold coalesced steps; added `test_explicit_zero_initial_chunk_is_not_pulled_below_steady`. Verified in the `novita-h100-omni` container with `python -m py_compile ...` and `PYTHONPATH=. python -m pytest tests/unit_test/moss_tts_local/test_streaming_vocoder.py -q` (`17 passed`).

sglang_omni/models/moss_tts_local/streaming_vocoder.py
Hayden727: Fixed in `6af81d81` by adding idempotent streaming-session close and closing it from scheduler teardown, with direct `stop()` cleanup for non-running test usage; added `test_stop_closes_persistent_streaming_session`. Verified in the `novita-h100-omni` container with `python -m py_compile ...` and `PYTHONPATH=. python -m pytest tests/unit_test/moss_tts_local/test_streaming_vocoder.py -q` (`17 passed`).

Hayden727 (approved):
Re-reviewed latest head `6af81d813bfd592b7e1bef9d075047c2c7854279`. The previous streaming coalescing and codec session teardown findings are fixed and both threads are resolved. I traced the path from MOSS local AR journal rows through the runner stream outbox, runtime stream routing, StreamingSimpleScheduler payload/done ordering, and the vocoder streaming/offline lanes; I do not see remaining blocking issues. Validation: targeted vocoder test on the same head in the novita-h100-omni container passed (`17 passed`), local py_compile for changed files passed, and GitHub lint/docs/test-layout are green. Omni CI is still gated only because the PR is missing the `run-ci` label.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -0,0 +1,670 @@
...
+        same forward and the steady state converges to one batched step per
+        cadence instead of one per stream.
+
+        Each iteration's chunks go to the outbox immediately, so a decode
+        failure in a later iteration cannot drop audio that was already
+        produced. A failed step leaves its participants' slot state
+        indeterminate, so all of them are failed and aborted — requests that
+        did not participate are untouched.
+        """
+        join_floor = max(
```
gaoyang07, line 469: The first-chunk threshold is still not request-local under coalescing.

The new guard only handles the explicit `initial_codec_chunk_frames=0` case. For positive first-chunk thresholds, `_can_join_coalesced_step()` still falls through to `len(state.pending) >= floor` before the stream has emitted anything.

A concrete case is:
- request A has `initial_codec_chunk_frames=1` and becomes due with 1 pending frame;
- request B has `initial_codec_chunk_frames=5` (or the scheduler default) and currently has only 1 pending frame;
- `_pump_streams()` computes `floor=1`;
- B is not due against its own threshold, but `_can_join_coalesced_step(B, 1)` returns true because `initial_chunk_frames != 0` and `pending >= floor`.

That makes B emit a 1-frame first chunk, then marks `emitted_any=True` and switches B to the steady threshold. Its requested/default 5-frame first chunk is never honored.

This does not corrupt the waveform, but it violates the per-request first-chunk/TTFA tuning contract and makes chunking depend on unrelated concurrent streams. Please keep first emissions from joining a coalesced step below their own latched threshold, while preserving the existing near-due coalescing for streams that have already emitted. A regression test should cover a positive first threshold, e.g. B with `initial_codec_chunk_frames=5` pending below 5 and A with `initial_codec_chunk_frames=1` becoming due; only A should emit until B reaches 5.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
edwingao28: ty!

sglang_omni/models/moss_tts_local/streaming_vocoder.py
edwingao28: Fixed in 6a4bc7ce: a stream that hasn't emitted its first chunk now never joins a coalesced step below its own latched threshold,  due-against-own-threshold and post-first-emission coalescing unchanged). 

Added test_positive_initial_chunk_is_not_pulled_below_threshold with the exact shape you described (B initial=5 buffering 4 frames, A initial=1 due): only A emits, and B's first chunk is exactly 5 frames once it fills.

tests/unit_test/moss_tts_local/test_streaming_vocoder.py
zhaochenyang20, line 1: Put the test file under the test Readme.

zhaochenyang20 (approved):

zhaochenyang20:
evaluation results:

# PR 753 SeedTTS Benchmark Report

## Scope

- PR: https://github.com/sgl-project/sglang-omni/pull/753
- Local branch: `pr-753-moss-tts-local-streaming`
- Commit tested: `e389a89602f57f019f1f009817db03aa440cd737`
- Model: `OpenMOSS-Team/MOSS-TTS-v1.5`
- Dataset: `zhaochenyang20/seed-tts-eval-arrow`
- Language: `en`
- Samples: `1088`
- TTS concurrency: `16`
- ASR model: `Qwen/Qwen3-ASR-1.7B`
- ASR concurrency: `32`
- GPU: single `NVIDIA B200` on `CUDA_VISIBLE_DEVICES=4`

This report covers the completed EN SeedTTS full-set run for non-streaming and streaming SSE. Both modes completed the full pipeline: auto start TTS, generate audio, stop TTS, start ASR, compute WER, stop ASR.

## Commands

Non-streaming:

```bash
CUDA_VISIBLE_DEVICES=4 python -m benchmarks.eval.benchmark_tts_seedtts \
  --model OpenMOSS-Team/MOSS-TTS-v1.5 \
  --port 8000 \
  --ref-format references \
  --token-count auto \
  --output-dir results/pr753_moss_tts_en_nonstream \
  --lang en \
  --max-concurrency 16
```

Streaming SSE:

```bash
CUDA_VISIBLE_DEVICES=4 python -m benchmarks.eval.benchmark_tts_seedtts \
  --model OpenMOSS-Team/MOSS-TTS-v1.5 \
  --port 8000 \
  --ref-format references \
  --token-count auto \
  --output-dir results/pr753_moss_tts_en_stream \
  --lang en \
  --max-concurrency 16 \
  --stream
```

## TTS Speed Results

| Metric | Non-stream | Streaming SSE | Delta |
|---|---:|---:|---:|
| Total requests | 1088 | 1088 | 0 |
| Completed requests | 1088 | 1088 | 0 |
| Throughput | 1.957 req/s | 1.932 req/s | -1.28% |
| Latency mean | 8.155 s | 8.259 s | +1.28% |
| Latency median | 8.081 s | 8.285 s | +2.52% |
| Latency p95 | 10.126 s | 10.209 s | +0.82% |
| Latency p99 | 11.278 s | 11.665 s | +3.43% |
| RTF mean | 1.9098 | 1.9314 | +1.13% |
| RTF median | 1.8447 | 1.8574 | +0.69% |
| RTF p95 | 2.7951 | 2.7999 | +0.17% |
| RTF p99 | 3.1879 | 3.1965 | +0.27% |
| Audio duration mean | 4.479 s | 4.481 s | +0.04% |
| Audio throughput | 8.765 audio-s/s | 8.658 audio-s/s | -1.22% |
| Output throughput | 174.1 tok/s | 172.0 tok/s | -1.21% |
| Output tokens/request-s | 25.4 | 25.2 | -0.79% |
| Output tokens mean | 89 | 89 | 0 |
| Output tokens total | 96825 | 96842 | +17 |
| Prompt tokens mean | 164 | 164 | 0 |
| Prompt tokens total | 178874 | 178874 | 0 |

## Streaming-Specific Metrics

| Metric | Streaming SSE |
|---|---:|
| TTFC mean | 8.258 s |
| TTFC median | 8.2842 s |
| TTFC p95 | 10.2084 s |
| TTFC p99 | 11.6643 s |
| Audio chunks mean | 1.0 |
| Audio chunks p95 | 1.0 |
| First audio payload bytes mean | 215068.2 |
| First audio payload bytes p95 | 307200.0 |

Observation: streaming returns one audio chunk on average in this run, so TTFC is effectively the same as total latency.

## WER Results

| Metric | Non-stream | Streaming SSE | Delta |
|---|---:|---:|---:|
| Evaluated / total | 1088 / 1088 | 1088 / 1088 | 0 |
| Skipped | 0 | 0 | 0 |
| WER corpus | 1.9761% | 1.6746% | -0.3014 pct point |
| WER per-sample mean | 2.0746% | 1.6714% | -0.4031 pct point |
| WER per-sample median | 0.0000 | 0.0000 | 0 |
| WER per-sample std | 0.0934 | 0.0685 | -0.0249 |
| WER per-sample p95 | 0.1000 | 0.1000 | 0 |
| WER per-sample max | 128.57% | 114.29% | -14.28 pct points |
| WER corpus excluding >50% samples | 1.4228% | 1.4021% | -0.0207 pct point |
| Samples with >50% WER | 7 (0.64%) | 3 (0.28%) | -4 samples |

## ASR Speed Results

| Metric | Non-stream WER ASR | Streaming WER ASR |
|---|---:|---:|
| ASR model | `Qwen/Qwen3-ASR-1.7B` | `Qwen/Qwen3-ASR-1.7B` |
| Evaluated / total | 1088 / 1088 | 1088 / 1088 |
| ASR concurrency | 32 | 32 |
| ASR total time | 56.917 s | 56.505 s |
| ASR throughput | 19.115 samples/s | 19.255 samples/s |
| ASR RTF mean | 0.3825 | 0.3790 |

## Result Files

- Non-streaming speed: `results/pr753_moss_tts_en_nonstream/speed_results.json`
- Non-streaming WER: `results/pr753_moss_tts_en_nonstream/wer_results.json`
- Non-streaming ASR speed: `results/pr753_moss_tts_en_nonstream/asr_speed_results.json`
- Streaming speed: `results/pr753_moss_tts_en_stream/speed_results.json`
- Streaming WER: `results/pr753_moss_tts_en_stream/wer_results.json`
- Streaming ASR speed: `results/pr753_moss_tts_en_stream/asr_speed_results.json`
- Generated audio count:
  - Non-streaming: `1088` wav files
  - Streaming: `1088` wav files

## Conclusion

The streaming SSE path is performance-neutral for this single-GPU SeedTTS EN run. Compared with non-streaming, streaming throughput is lower by only `1.28%`, mean latency is higher by `1.28%`, and mean RTF is higher by `1.13%`. WER is not worse; streaming measured slightly better on this run (`1.67%` vs `1.98%` corpus WER), within normal sample-level variation.

The result looks reasonable for the PR: streaming has a small overhead, no failed requests, full-set WER completes, and there is no obvious performance regression.

tests/unit_test/moss_tts_local/test_streaming_vocoder.py
edwingao28: addressed

Issue #733, opened 2026-06-10
[Perf] MOSS TTS: torch.compile the audio encoder


Ccyest:
## Motivation

Once the audio encoder runs as a dedicated GPU stage (#730), applying `torch.compile` to the codec encoder forward pass can fuse kernels and reduce launch overhead — the same optimization Higgs applied to its DAC encoder in PR #612.

The MOSS audio tokenizer's encoder is a standard Conv1d stack that should compile cleanly without graph breaks.

## Scope

- Apply `torch.compile` to the MOSS audio tokenizer's encode path
- Measure compile-time overhead (expected: 1–2 minutes on first call)
- Verify no graph breaks in the encode forward path
- Benchmark latency reduction at various concurrencies

## Acceptance Criteria

**Correctness**
- [ ] Compiled encoder produces bit-identical codec codes vs eager mode
- [ ] Existing unit tests still pass

**Performance**
- [ ] Encoder latency strictly lower than eager baseline
- [ ] No regression at concurrency = 1
- [ ] Compile-time startup penalty documented

## Known Pitfalls

- **Graph breaks from dynamic shapes.** If waveform lengths vary, `torch.compile` may recompile for each unique shape. Consider using `torch.compile(dynamic=True)` or padding to fixed length buckets.
- **Dependency**: best applied after #730 (decouple preprocessing) so the encoder runs as a dedicated GPU stage.

## References

- Higgs encoder compile: PR #612
- Prerequisite: #730 (decouple preprocessing)
- Assigned: #653 @nagisa-kunhah
- Tracking issue: #637

SandyLuXY:
Hi @Ccyest, I'm interested in taking this issue.

PR #755, opened 2026-06-12
[Perf] Compile MOSS-TTS v1.5 audio encoder behind opt-in

SandyLuXY:
## Motivation

MOSS-TTS Local reference-audio preprocessing runs through the codec encoder and quantizer before the AR engine can generate audio. This PR adds an opt-in `torch.compile` path for the codec quantizer so we can reduce preprocessing overhead under concurrent serving, while keeping the default runtime behavior unchanged.

## Modifications

1. `sglang_omni/models/moss_tts_local/stages.py`
   - Adds `enable_quantizer_torch_compile`, `quantizer_torch_compile_mode`, and `quantizer_torch_compile_warmup_seconds` to `create_preprocessing_executor`.
   - Compiles the loaded processor's `audio_tokenizer.quantizer.forward` with `torch.compile(fullgraph=False, dynamic=True, mode=...)`, then wraps the compiled callable with `torch.no_grad()`.
   - Warms the compiled quantizer through `processor.encode_audios_from_wav()` using configurable positive warmup durations and the processor/tokenizer sample rate.
   - Restores the original instance `forward` on compile or warmup failure and raises a clear runtime error.
   - Adds `MOSS_LOCAL_QUANTIZER_TORCH_COMPILE=0|false|no|off` as an ops kill switch, and reuses the same env-disable parsing for `MOSS_REF_AUDIO_CACHE`.

2. `sglang_omni/models/moss_tts_local/config.py`
   - Wires the preprocessing stage with explicit default compile settings:
     `enable_quantizer_torch_compile=false`, `quantizer_torch_compile_mode=default`, and `quantizer_torch_compile_warmup_seconds=[1.0]`.

3. `tests/unit_test/moss_tts_local/test_pipeline.py`
   - Covers default config wiring, compile wrapper/warmup behavior, failure restoration, preprocessing executor wiring, and the env kill switch.

## Related Issues
#733 

## Accuracy Test

**Setup:**

- Baseline: `main @ c8ea7ad` using `configs/moss_tts_local_main.yaml`
- Candidate: `feat/moss-tts-quantizer-torch-compile @ 465f4c4`
- Dataset/eval: SeedTTS EN, `zhaochenyang20/seed-tts-eval-arrow`, 1088 samples per variant/concurrency
- Runtime: one run per variant at concurrency 1 and 16, logical `cuda:0` for the AR engine and `cuda:1` for preprocessing/vocoder codec processors
- Hardware/software: 2x NVIDIA H100

All WER and speaker-similarity checks evaluated 1088/1088 samples with 0 skipped.

| Concurrency | Metric | Baseline | Candidate | 
| ---: | --- | ---: | ---: | 
| 1 | Corpus WER | 0.020598 | 0.020682 | 
| 1 | Speaker similarity | 65.186785 | 65.198589 | 
| 16 | Corpus WER | 0.024115 | 0.022942 |
| 16 | Speaker similarity | 65.239819 | 65.477828 | 

## Benchmark & Profiling

| Concurrency | Metric | Baseline | Candidate | Delta |
| 1 | Failed requests | 0 | 0 | 0 |
| 1 | Mean latency | 0.621 s | 0.624 s | +0.48% |
| 1 | P95 latency | 0.839 s | 0.841 s | +0.24% |
| 1 | Mean RTF | 0.1437 | 0.1442 | +0.35% |
| 1 | Throughput | 1.609 QPS | 1.603 QPS | -0.37% |
| 16 | Failed requests | 0 | 0 | 0 |
| 16 | Mean latency | 1.587 s | 1.514 s | -4.60% |
| 16 | P95 latency | 2.024 s | 1.966 s | -2.87% |
| 16 | Mean RTF | 0.3782 | 0.3596 | -4.92% |
| 16 | Throughput | 10.042 QPS | 10.521 QPS | +4.77% |

### Summary:

- c=1 is effectively neutral in this single run, with all latency, throughput, and quality deltas below 0.5%.
- c=16 improves end-to-end serving metrics: mean latency -4.60%, p95 latency -2.87%, mean RTF -4.92%, and throughput +4.77%, with no quality regression.

## Checklist

- [x] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/models/moss_tts_local/stages.py
+++ b/sglang_omni/models/moss_tts_local/stages.py
@@ -9,6 +9,7 @@ import os
 import queue
 import threading
 import time
+from collections.abc import Sequence
 from typing import Any
 
 import torch
@@ -102,6 +103,89 @@ def _load_moss_tts_local_processor(model_path: str, *, device: str) -> Any:
             # would corrupt the quantizer codebooks.
             audio_tokenizer.to(device)
     return processor
+
+
+def _normalize_quantizer_compile_warmup_seconds(
+    warmup_seconds: Sequence[float] | None,
+) -> tuple[float, ...]:
+    source = (1.0,) if warmup_seconds is None else warmup_seconds
+    values = tuple(float(value) for value in source)
+    if not values:
+        raise RuntimeError("MOSS-TTS Local quantizer torch.compile warmup is empty")
+    for value in values:
+        if value <= 0:
+            raise RuntimeError(
+                "MOSS-TTS Local quantizer torch.compile warmup seconds "
+                "must be positive"
+            )
+    return values
+
+
+def _quantizer_compile_sample_rate(processor: Any) -> int:
+    sample_rate = int(processor.audio_tokenizer.sampling_rate)
+    if sample_rate <= 0:
+        raise RuntimeError(
+            "MOSS-TTS Local quantizer torch.compile warmup sample rate must be positive"
+        )
+    return sample_rate
+
+
+def _quantizer_compile_channels(processor: Any) -> int:
+    channels = int(processor.audio_tokenizer.number_channels)
+    if channels <= 0:
+        raise RuntimeError(
+            "MOSS-TTS Local quantizer torch.compile warmup channels must be positive"
+        )
+    return channels
+
+
+def _warmup_moss_tts_local_quantizer(
+    processor: Any,
+    warmup_seconds: Sequence[float],
+) -> None:
+    encode_audios = processor.encode_audios_from_wav
+    if not callable(encode_audios):
+        raise RuntimeError(
+            "MOSS-TTS Local processor.encode_audios_from_wav is unavailable"
+        )
+    sample_rate = _quantizer_compile_sample_rate(processor)
+    channels = _quantizer_compile_channels(processor)
+    for seconds in warmup_seconds:
+        num_samples = max(1, int(round(float(seconds) * sample_rate)))
+        wav = torch.zeros((channels, num_samples), dtype=torch.float32)
+        encode_audios([wav], sample_rate)
+
+
+def _compile_moss_tts_local_quantizer(
+    processor: Any,
+    *,
+    compile_mode: str | None = "default",
+    warmup_seconds: Sequence[float] | None = None,
+) -> None:
+    audio_tokenizer = processor.audio_tokenizer
+    quantizer = audio_tokenizer.quantizer
+    if not callable(quantizer):
+        raise RuntimeError("MOSS-TTS Local audio_tokenizer quantizer is not callable")
+
+    compile_kwargs: dict[str, Any] = {"fullgraph": False, "dynamic": True}
+    if compile_mode is not None:
+        compile_kwargs["mode"] = compile_mode
+    normalized_warmup = _normalize_quantizer_compile_warmup_seconds(warmup_seconds)
+
+    try:
+        audio_tokenizer.quantizer = torch.compile(quantizer, **compile_kwargs)
+        _warmup_moss_tts_local_quantizer(processor, normalized_warmup)
+    except Exception:
+        audio_tokenizer.quantizer = quantizer
+        logger.exception("MOSS-TTS Local quantizer torch.compile failed")
+        raise
+
+    logger.info(
+        "Enabled MOSS-TTS Local quantizer torch.compile "
+        "(mode=%s, warmup_seconds=%s)",
+        compile_mode,
+        list(normalized_warmup),
+    )
 
 
 class _BatchedReferenceEncoder:
@@ -409,6 +493,9 @@ def create_preprocessing_executor(
     ref_audio_cache: bool = True,
     ref_audio_cache_max_items: int = 256,
     ref_audio_cache_max_bytes: int = 64 * 1024 * 1024,
+    enable_quantizer_torch_compile: bool = False,
+    quantizer_torch_compile_mode: str | None = "default",
+    quantizer_torch_compile_warmup_seconds: Sequence[float] | None = None,
 ) -> SimpleScheduler:
     # MOSS_REF_AUDIO_CACHE=0 disables the cache at startup (ops kill switch / A-B
     # toggle) without a config edit; unset => kwarg default.
@@ -423,6 +510,12 @@ def create_preprocessing_executor(
         )
     device = _resolve_codec_device(device, gpu_id)
     processor = _load_moss_tts_local_processor(model_path, device=device)
+    if enable_quantizer_torch_compile:
+        _compile_moss_tts_local_quantizer(
+            processor,
+            compile_mode=quantizer_torch_compile_mode,
+            warmup_seconds=quantizer_torch_compile_warmup_seconds,
+        )
     reference_encoder: Any = _BatchedReferenceEncoder(
         processor,
         max_batch_size=encode_batch_size,
--- a/sglang_omni/models/moss_tts_local/config.py
+++ b/sglang_omni/models/moss_tts_local/config.py
@@ -20,6 +20,9 @@ def _stages(*, codec_device: str) -> list[StageConfig]:
             factory=f"{_PKG}.stages.create_preprocessing_executor",
             factory_args={
                 "device": codec_device,
+                "enable_quantizer_torch_compile": False,
+                "quantizer_torch_compile_mode": "default",
+                "quantizer_torch_compile_warmup_seconds": [1.0],
                 "ref_audio_cache": True,
                 "ref_audio_cache_max_items": 256,
                 "ref_audio_cache_max_bytes": 64 * 1024 * 1024,
--- a/tests/unit_test/moss_tts_local/test_pipeline.py
+++ b/tests/unit_test/moss_tts_local/test_pipeline.py
@@ -332,56 +332,82 @@ def test_create_preprocessing_executor_env_toggle(monkeypatch):
     )
 
 
-def test_quantizer_torch_compile_wraps_forward_and_warms_up(monkeypatch):
+class _RecordingQuantizer(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.calls = 0
+        self.compiled_calls = 0
+        self.last_n_quantizers = "unset"
+
+    @torch.no_grad()
+    def forward(self, z, input_length, n_quantizers=None):
+        del input_length
+        self.calls += 1
+        self.last_n_quantizers = n_quantizers
+        return z
+
+
+class _CompiledQuantizer(torch.nn.Module):
+    def __init__(self, eager_quantizer):
+        super().__init__()
+        self.eager_quantizer = eager_quantizer
+
+    def forward(self, *args, **kwargs):
+        self.eager_quantizer.compiled_calls += 1
+        return self.eager_quantizer(*args, **kwargs)
+
+
+class _WarmupFailingQuantizer(torch.nn.Module):
+    def forward(self, *args, **kwargs):
+        del args, kwargs
+        raise RuntimeError("warmup failed")
+
+
+class _FakeAudioTokenizer(torch.nn.Module):
+    def __init__(
+        self,
+        *,
+        number_channels: int = 1,
+        sampling_rate: int = 48000,
+    ):
+        super().__init__()
+        self.number_channels = number_channels
+        self.sampling_rate = sampling_rate
+        self.quantizer = _RecordingQuantizer()
+
+
+class _FakeQuantizerProcessor:
+    def __init__(
+        self,
+        *,
+        number_channels: int = 1,
+        sampling_rate: int = 48000,
+    ):
+        self.audio_tokenizer = _FakeAudioTokenizer(
+            number_channels=number_channels,
+            sampling_rate=sampling_rate,
+        )
+        self.warmup_shapes = []
+
+    def encode_audios_from_wav(self, wavs, sample_rate):
+        assert sample_rate == self.audio_tokenizer.sampling_rate
+        self.warmup_shapes.append(tuple(wavs[0].shape))
+        input_length = torch.tensor([wavs[0].shape[-1]])
+        return [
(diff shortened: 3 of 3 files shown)


zhaochenyang20:
Great job!!!

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -115,6 +121,125 @@ def _load_moss_tts_local_processor(model_path: str, *, device: str) -> Any:
...
+) -> None:
+    encode_audios = getattr(processor, "encode_audios_from_wav", None)
+    if not callable(encode_audios):
+        raise RuntimeError(
+            "MOSS-TTS Local processor.encode_audios_from_wav is unavailable"
+        )
+    sample_rate = _quantizer_compile_sample_rate(processor)
+    for seconds in warmup_seconds:
+        num_samples = max(1, int(round(float(seconds) * sample_rate)))
+        wav = torch.zeros((1, num_samples), dtype=torch.float32)
```
Hayden727: [P1] This warmup tensor is hard-coded to mono, but `OpenMOSS-Team/MOSS-Audio-Tokenizer-v2` is configured with `number_channels=2`, and upstream `_prepare_waveform_batch()` rejects 2-D wavs whose first dimension is not the tokenizer channel count. When `enable_quantizer_torch_compile=True`, this calls `encode_audios_from_wav([wav], sample_rate)` during startup and fails before serving. Please derive the channel count from `audio_tokenizer.number_channels`/config and warm with `(channels, num_samples)`; I verified a regression test fails on current head with `bad warmup shape: (1, 48000)`.

Hayden727 (commented):
Left one correctness note for the opt-in compile path.

sglang_omni/models/moss_tts_local/stages.py
Hayden727: Fixed in 19ac3eab6f25024b684cf468894b10f6c3aa8de3 by deriving warmup channels from the tokenizer/config and adding a stereo warmup regression test.

Validation on novita-h100-omni / sglang-omni-hayden-moss-ref-encode:
- Red test on ba11ab43 failed with `bad warmup shape: (1, 48000)`.
- Targeted compile/wiring subset: 5 passed.
- `tests/unit_test/moss_tts_local/test_pipeline.py`: 42 passed.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -115,6 +121,148 @@ def _load_moss_tts_local_processor(model_path: str, *, device: str) -> Any:
...
+
+    instance_forward = _get_instance_attr(quantizer, "forward")
+    compile_target = _bind_unwrapped_method(target_forward, quantizer)
+    compile_kwargs: dict[str, Any] = {"fullgraph": False, "dynamic": True}
+    if compile_mode is not None:
+        compile_kwargs["mode"] = compile_mode
+    normalized_warmup = _normalize_quantizer_compile_warmup_seconds(warmup_seconds)
+
+    try:
+        compiled_forward = torch.compile(compile_target, **compile_kwargs)
```
Hayden727: [P1] This compiled quantizer path still misses #733's acceptance criteria in full E2E validation. I re-ran current head `19ac3eab` on `novita-h100-omni` in `sglang-omni-hayden-moss-ref-encode`, model `OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5`, SeedTTS EN full set, `generate-only`, warmup=10, same server config except this flag. c=1 was neutral/slightly positive (`latency_mean_s` 0.648 -> 0.646, `p95` 0.908 -> 0.896, QPS 1.542 -> 1.547), but c=16 regressed and produced failures: off had 1088/1088 success, mean 2.023s, RTF 0.4814, QPS 7.861; compile-on had 1083/1088 success, mean 2.084s, RTF 0.5015, QPS 7.586. The 5 failed requests were HTTP 500s from `ForwardBatch.init_new -> clamp_position` with `RuntimeError: Detected that you are using FX to symbolically trace a dynamo-optimized function`. I also checked codec correctness directly: after equal eager warmup, eager repeat and compiled repeat were each stable, but eager-vs-compiled codec codes were not bit-identical for `common_voice_en_10119832-common_voice_en_10119840` (`shape=[48,12]`, max code delta 526). A multi-warmup diagnostic (`[1,2,4,8,16]`) avoided the c=16 failures once, but still regressed vs eager (mean +4.15%, RTF +4.05%, QPS -3.73%), so that does not look like a viable fix. Given #733 requires bit-identical codec codes, lower encoder latency, and no c=1 regression, I don't think this should be presented as satisfying the issue yet.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -115,6 +121,148 @@ def _load_moss_tts_local_processor(model_path: str, *, device: str) -> Any:
...
+    audio_tokenizer = getattr(processor, "audio_tokenizer", None)
+    return int(
+        getattr(audio_tokenizer, "sampling_rate", 0)
+        or getattr(getattr(audio_tokenizer, "config", None), "sampling_rate", 0)
+        or getattr(getattr(processor, "model_config", None), "sampling_rate", 0)
+        or 48000
+    )
+
+
+def _quantizer_compile_channels(processor: Any) -> int:
```
yxs, line 133: `_quantizer_compile_channels` falls back to `1` when none of `number_channels` / `config.number_channels` / `config.channels_numbers` are readable on the tokenizer, so the warmup builds a `[1, num_samples]` mono wav. But this repo explicitly states "The v2 codec is natively stereo" (`stages.py:771`, vocoder keeps the `[channels, samples]` 2-channel layout). I verified on torch 2.11 that **even with `dynamic=True`, a size-1 dimension is force-specialized to static by Dynamo** — after warming with channels=1, the first real channels=2 request always triggers a recompile (`graphs: 1→2`). So if at runtime those three attrs are all unreadable, the warmup compiles the wrong channel count and the entire "front-load compile cost at startup" selling point collapses — the compile cost lands on the first real stereo request (a first-request latency spike). It does not produce wrong output (functionally still correct), but it directly affects whether the claimed gains are stable.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -115,6 +121,148 @@ def _load_moss_tts_local_processor(model_path: str, *, device: str) -> Any:
...
+        )
+    sample_rate = _quantizer_compile_sample_rate(processor)
+    channels = _quantizer_compile_channels(processor)
+    for seconds in warmup_seconds:
+        num_samples = max(1, int(round(float(seconds) * sample_rate)))
+        wav = torch.zeros((channels, num_samples), dtype=torch.float32)
+        encode_audios([wav], sample_rate)
+
+
+def _compile_moss_tts_local_quantizer(
```
yxs, line 159: **File path** `reference_encoder.encode(path)` → serialized through `_BatchedReferenceEncoder`'s **single** daemon worker thread → the compiled quantizer is called from one thread only, safe. (2) **Data-URI path** `encode_data_uri(...)` → inside `_cached_encode`, the leader runs `processor.encode_audios_from_wav(...)` directly on a **worker thread** (one of 16) → leaders for distinct data-URIs (distinct cache keys) are **not serialized at all**, so the compiled `quantizer.forward` **can be invoked concurrently from multiple threads**.

zhaochenyang20 (changes requested):
If we accept the compile path as the current best-performance path, change the default pipeline config to `enable_quantizer_torch_compile=True` and change the `create_preprocessing_executor()` default argument to `True` as well, so custom configs that omit the field also take the best path. Keep the existing `MOSS_LOCAL_QUANTIZER_TORCH_COMPILE=0|false|no|off` behavior as the ops kill switch, and state in the PR description that compile is default-on while the environment variable is only for rollback.

So I do not think that we need to have a parameter, especially environment parameter, for most models quantization Torch compile because I think a lot of the models have the same feature. So this should be like a general server argument for all the models, not an environment parameter for moss itself.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -43,6 +46,9 @@
     "Launch with trust_remote_code=True and make sure the checkpoint can load "
     "OpenMOSS-Team/MOSS-Audio-Tokenizer-v2."
 )
+_MOSS_LOCAL_QUANTIZER_COMPILE_ENV = "MOSS_LOCAL_QUANTIZER_TORCH_COMPILE"
+_DISABLE_ENV_VALUES = {"0", "false", "no", "off", ""}
+_MISSING = object()
```
zhaochenyang20: tend not to use env args.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -115,6 +121,148 @@ def _load_moss_tts_local_processor(model_path: str, *, device: str) -> Any:
...
+    instance_dict = getattr(instance, "__dict__", None)
+    if isinstance(instance_dict, dict) and name in instance_dict:
+        return instance_dict[name]
+    return _MISSING
+
+
+def _restore_instance_attr(instance: Any, name: str, value: Any) -> None:
+    if value is _MISSING:
+        try:
+            delattr(instance, name)
```
zhaochenyang20: So we tend not to use dynamic attribute. `delattr` looks to weird.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -115,6 +121,148 @@ def _load_moss_tts_local_processor(model_path: str, *, device: str) -> Any:
...
+        except AttributeError:
+            pass
+    else:
+        setattr(instance, name, value)
```
zhaochenyang20: Not use dynamic attribute.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -115,6 +121,148 @@ def _load_moss_tts_local_processor(model_path: str, *, device: str) -> Any:
...
+
+
+def _quantizer_compile_sample_rate(processor: Any) -> int:
+    audio_tokenizer = getattr(processor, "audio_tokenizer", None)
+    return int(
+        getattr(audio_tokenizer, "sampling_rate", 0)
+        or getattr(getattr(audio_tokenizer, "config", None), "sampling_rate", 0)
+        or getattr(getattr(processor, "model_config", None), "sampling_rate", 0)
+        or 48000
+    )
```
zhaochenyang20: Not use dynamic attribute.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -115,6 +121,148 @@ def _load_moss_tts_local_processor(model_path: str, *, device: str) -> Any:
...
+
+
+def _quantizer_compile_channels(processor: Any) -> int:
+    audio_tokenizer = getattr(processor, "audio_tokenizer", None)
+    value = (
+        getattr(audio_tokenizer, "number_channels", 0)
```
zhaochenyang20: Not use dynamic attribute.

zhaochenyang20 (changes requested):

SandyLuXY:
Ran the full benchmark and confirmed regression for c=16 as @Hayden727 mentioned above.

**Setup:**

- Baseline: `main @ 671683e`
- Candidate: `feat/moss-tts-quantizer-torch-compile @ a217c76`
- Dataset/eval: SeedTTS EN, `zhaochenyang20/seed-tts-eval-arrow`, 1088 samples per variant/concurrency
- Hardware/software: 2x NVIDIA H100

All WER and speaker-similarity checks evaluated 1088/1088 samples with 0 skipped.

| Concurrency | Metric | Baseline | Candidate |
| ---: | --- | ---: | ---: |
| 1 | Corpus WER | 0.022691 | 0.021770 |
| 1 | Speaker similarity | 65.434543 | 65.442539 |
| 16 | Corpus WER | 0.027213 | 0.025036 |
| 16 | Speaker similarity | 65.633151 | 65.448595 |

| Concurrency | Metric | Baseline | Candidate | Delta |
| 1 | Failed requests | 0 | 0 | 0 |
| 1 | Mean latency | 0.746 s | 0.742 s | -0.54% |
| 1 | P95 latency | 1.025 s | 1.022 s | -0.29% |
| 1 | Mean RTF | 0.1724 | 0.1715 | -0.52% |
| 1 | Throughput | 1.340 QPS | 1.347 QPS | +0.52% |
| 16 | Failed requests | 0 | 0 | 0 |
| 16 | Mean latency | 2.076 s | 2.169 s | +4.48% |
| 16 | P95 latency | 2.900 s | 3.064 s | +5.66% |
| 16 | Mean RTF | 0.4944 | 0.5161 | +4.39% |
| 16 | Throughput | 7.679 QPS | 7.347 QPS | -4.32% |

### Summary:

- c=1 is effectively neutral in this single previous run, with candidate mean latency -0.54%, p95 latency -0.29%, mean RTF -0.52%, and throughput +0.52%.
- c=16 currently regresses end-to-end serving metrics in the latest reruns: mean latency +4.48%, p95 latency +5.66%, mean RTF +4.39%, and throughput -4.32%.
- Quality does not show regression.

---
### Additional profiler analysis

I also checked the request profiler for the c=16 runs. The regression does not appear to come from the AR `tts_engine`; it is concentrated on the codec side.

| Stage | Baseline avg | Candidate avg | Delta |
| --- | ---: | ---: | ---: |
| `preprocessing` | 312.361 ms | 331.434 ms | +19.073 ms |
| `tts_engine` | 690.361 ms | 693.852 ms | +3.491 ms |
| `vocoder` | 1057.415 ms | 1121.109 ms | +63.694 ms |

The `tts_engine` stage is basically flat, while both `preprocessing` and `vocoder` get slower. This matches the pipeline placement:

```text
tts_engine: logical cuda:0
preprocessing codec processor: logical cuda:1
vocoder codec processor: logical cuda:1
```

The `torch.compile` change only targets the preprocessing quantizer, but preprocessing and vocoder share the codec GPU. If the compiled quantizer consumes more `cuda:1` time, c=16 traffic amplifies that cost and also delays vocoder work on the same GPU. That explains why c=1 is effectively neutral while c=16 regresses: under low load the extra preprocessing cost is mostly hidden, but under c=16 the shared codec GPU becomes the bottleneck.

Issue #721, opened 2026-06-09
[Investigation] Profiling AR-stage concurrency knees 


estellaliu233:
## Summary

I did a lightweight profiling pass on several workloads, including Higgs TTS, Qwen3-Omni, and Qwen3-ASR. Across the initial runs, AR / generation-related stages appear to dominate the request path.

The clearest case is Higgs TTS. On the full SeedTTS EN set, increasing client concurrency from 16 to 32 gives almost no throughput gain, but significantly worsens latency and RTF.

Higgs full SeedTTS EN:

- c16 baseline:
  - Throughput: 14.656/s
  - Mean latency: 1.087s
  - P95 latency: 1.634s
  - RTF mean: 0.2632

- c32 baseline:
  - Throughput: 14.916/s
  - Mean latency: 2.122s
  - P95 latency: 2.807s
  - RTF mean: 0.5302

From c16 to c32, throughput improves by only ~1.8%, while mean latency increases by ~95%, p95 by ~72%, and RTF by ~101%. This looks like the workload is past its useful active-concurrency region.

I also tried a simple AR-stage admission cap of 16 at client concurrency 32.

- c32 with AR cap16:
  - Throughput: 14.498/s
  - Mean latency: 2.078s
  - P95 latency: 2.663s
  - RTF mean: 0.5120

Compared with c32 baseline:

- Throughput: -2.8%
- Mean latency: -2.1%
- P95 latency: -5.1%
- RTF mean: -3.4%

The c16 -> c32 result shows a clear concurrency knee. The AR cap16 experiment gives only small improvements, so over-admission is still a possible factor, but the current cap experiment does not strongly confirm it. I would conduct finer-grained queue/timing attribution to identify further before deciding whether admission control is the right fix.

## Additional quick sweep

I ran 50-sample generation-only sweeps for several other models:

- FishAudio S2-Pro, c16 -> c32:
  - Throughput: +1.7%
  - Mean latency: +67.1%
  - P95 latency: +68.2%
  - RTF: +74.9%

- Qwen3-TTS 0.6B, c16 -> c32:
  - Throughput: +5.5%
  - Mean latency: +63.1%
  - P95 latency: +73.7%
  - RTF: +75.0%

- MOSS-TTS, c16 -> c32:
  - Throughput: +9.9%
  - Mean latency: +57.7%
  - P95 latency: +43.4%
  - RTF: +58.2%

- Voxtral TTS, c16 -> c32:
  - Throughput: +19.5%
[...]

PR #756, opened 2026-06-12
[Higgs TTS] Raise AR server default to 64 and expose standard batch knobs

estellaliu233:
# Summary

This PR raises the Higgs TTS AR serving default and exposes the related SGLang generation-stage batch knobs through typed CLI / benchmark parameters.

Changes:

- Raise Higgs AR server defaults from `16/16` to `64/64`, where `A/B` means:
  - `max_running_requests=A`
  - `cuda_graph_max_bs=B`
- Add typed serve / benchmark overrides for the two standard SGLang engine parameters:
  - `--max-running-requests`
  - `--cuda-graph-max-bs`
- Apply those overrides through a generic generation-stage path, rather than a Higgs-only helper. This covers TTS `tts_engine` stages and SGLang talker stages such as Qwen3-Omni `talker_ar`.
- If only one of the two flags is set, fill the other with the same value. This keeps the common tuning path simple while preserving the underlying SGLang parameter semantics.
- Size Higgs model-side CUDA graph / sampler buffers from SGLang `max_running_requests`, which fixes larger capture sizes such as `128`.
- Remove redundant Higgs batch-size constants and let the stage constructor defaults be the single source of truth.

Related: #721

## Why

The previous Higgs AR default was conservative at `16/16`. In serving benchmarks, this limited throughput at higher client concurrency because the server could not admit/capture enough concurrent decode batches.

The latest sweep points to `64/64` as the practical Higgs default:

- `32/32` clearly improves c32 throughput/RTF over `16/16`.
- `64/64` extends the useful throughput plateau to c64.
- `128/128` is useful as a stress/tuning option, but it is not justified as the default from the current evidence.

The earlier c16 low-concurrency attribution should be treated as inconclusive rather than as a confirmed larger-graph regression. The measurements were single-run and not precise enough to separate small effects from variance. This PR therefore focuses on the default and tuning interface, not on claiming a root cause for c16 noise.

## Benchmark Setup

- Dataset: SeedTTS EN full set, `1088` samples
- Hardware: H200

### `16/16` vs `32/32` at concurrency 32

| config | completed | failed | median latency | p95 latency | p99 latency | throughput | RTF mean |
| `16/16` | 1088 | 0 | 2.020s | 2.584s | 2.957s | 15.377 req/s | 0.5107 |
| `32/32` | 1088 | 0 | 1.445s | 2.200s | 2.664s | 21.261 req/s | 0.3638 |

`32/32` vs `16/16` at c32:

- median latency: `-28.5%`
- p95 latency: `-14.9%`
- p99 latency: `-9.9%`
- throughput: `+38.3%`
- RTF mean: `-28.8%`

### Scalability sweep

| config | c16 RTF mean | c32 RTF mean | c64 RTF mean | c64 throughput |
|---|---:|---:|---:|---:|
| `32/32` | 0.2478 | 0.3687 | 0.7623 | 20.693 req/s |
| `64/64` | 0.2486 | 0.3807 | 0.6667 | 23.604 req/s |
| `128/128` | 0.2627 | 0.3733 | 0.6636 | 23.689 req/s |

Current read:

- `64/64` gives a clear c64 gain over `32/32`.
- `64/64` and `128/128` are very close at c64, so `128/128` is not necessary as the default.
- `64/64` keeps a conservative margin while raising the old `16/16` serving limit.

## Quality / WER

Full SeedTTS EN WER, `1088` samples:

| config | evaluated | skipped | corpus WER | mean WER | p95 WER | max WER | >50% WER samples |
| `16/16` | 1088 | 0 | 1.19% | 1.15% | 9.09% | 64.29% | 1 |
| `32/32` | 1088 | 0 | 1.04% | 1.01% | 8.33% | 42.86% | 0 |

No quality regression was observed in the checked runs.

## Startup / CUDA Graph Capture Cost

Startup sweep, 3 runs per setting, median ready time:

| max_running_requests | cuda_graph_max_bs | median ready time | peak GPU memory | status |
|---:|---:|---:|---:|---|
| 16 | 16 | 56.69s | 122062 MiB | ok |
| 32 | 32 | 55.46s | 122122 MiB | ok |
| 64 | 64 | 56.41s | 122298 MiB | ok |
| 128 | 128 | 60.83s | 122880 MiB | ok after capture fix |

For `64/64`:

- startup time is effectively unchanged vs `16/16`
- peak memory increases by about `236 MiB`

## 128 Capture Fix

The previous Higgs internal sampler / CUDA graph buffers were effectively sized around the old internal max-batch assumption. With larger CUDA graph capture sizes such as `128`, capture could fail in Higgs decode embedding with a batch shape mismatch.

This PR resolves the Higgs internal max batch size from SGLang `max_running_requests`, so the model-side sampler / CUDA graph buffers match the serving configuration.


PopSoda2002 (approved):
LGTM if test passed and no regression

benchmarks/eval/benchmark_tts_seedtts.py
```diff
@@ -634,6 +652,34 @@ def _build_arg_parser() -> argparse.ArgumentParser:
...
+    parser.add_argument(
+        "--server-extra-arg",
+        action="append",
+        default=[],
+        help=(
+            "Extra argument passed to `sglang_omni.cli serve` when this "
+            "benchmark starts the TTS server. May be repeated. Example: "
+            "stages.2.factory_args.server_args_overrides.max_running_requests=32"
+        ),
+    )
```
zhaochenyang20: This is concerning to me. If all the args can be passed in with `--server-extra-arg`, server will only have one arguemnt. We should not use these kinds of ambiguous names. We should keep all the arguments in the CI, one by one. Do not add arguments like this.

benchmarks/eval/benchmark_tts_seedtts.py
```diff
@@ -634,6 +652,34 @@ def _build_arg_parser() -> argparse.ArgumentParser:
...
+    parser.add_argument(
+        "--tts-cuda-graph-max-bs",
+        type=int,
+        default=None,
+        help=(
+            "Convenience override for Higgs/TTS AR server cuda_graph_max_bs. "
+            "Equivalent to a stages.2 server override."
+        ),
+    )
```
zhaochenyang20: Why should we keep them seperate? I think so far we should keep them the same. I mean, we should only have one argument: the maximum running request is the maximum CUDA Graph batch size. So we should reduce these parameters to one parameter.

zhaochenyang20 (changes requested):

benchmarks/eval/benchmark_tts_seedtts.py
estellaliu233: Thanks! I updated the argument path to avoid exposing a generic server-extra-arg interface. The benchmark now uses a Higgs-specific `--higgs-ar-max-batch-size`, and `sglang_omni.cli serve` also has a typed `--higgs-ar-max-batch-size` flag. Internally, this sets both `max_running_requests` and `cuda_graph_max_bs` on the Higgs `tts_engine` stage.

I also ran full scalability sweeps for `32/32`, `64/64`, and `128/128`, where `A/B` means `max_running_requests=A` and `cuda_graph_max_bs=B`.

Once CUDA Graph batch size is no longer capped at 16, Higgs continues scaling beyond c16. In this setup, RTF stays below 1 through c64 and crosses 1 at c128, so c64 is the last useful scalability point for this workload.

For the default-size decision, the readout is a bit more nuanced:

* `128/128` gives the best high-concurrency RTF, especially at c64/c128.
* However, at the useful c64 point, `128/128` is only marginally better than `64/64` in RTF: `0.6636` vs `0.6667`.
* At c128, all configs are already overloaded because RTF mean is above 1.
* At c16, `128/128` is worse than both `32/32` and `64/64`.
* `64/64` reaches essentially the same practical throughput plateau as `128/128`, while avoiding going all the way to 128.

So my current recommendation is:

* If we want the safest minimal-change default for this PR, `32/32` is still reasonable. If we want the default to reflect the scalability result, I think `64/64` is the better practical default.
* I would not recommend 128/128 as the default from this sweep. It is useful as a high-concurrency / stress-test tuning option, but it does not provide meaningful improvement over 64/64 around the practical plateau and shows worse low-concurrency behavior in this run.

I also did a lightweight single-run attribution for the 128 behavior. Output length variation explains about 39% of the raw latency delta. Increasing `max_running_requests` alone is nearly flat, around ~0.8% normalized overhead. The remaining ~5–6% normalized slowdown appears when CUDA Graph coverage is actually increased to 128. Capture time itself does not increase much, so I would not describe this as capture-time overhead directly. It is more like a runtime/memory residual from larger graph coverage, but I did not further decompose that in this PR.

I added the RTF curve and sweep tables with throughput, latency, RTF, audio throughput, output token/s, and failures. All runs used the same model/dataset setup and had 0 failures.

For follow-ups, I think there are two useful directions that are better kept separate from this PR:

1. add a lightweight server-side sweep utility so we can reuse the same methodology for MOSS 1.5 and other models;
2. separately profile the larger-graph-coverage residual to check whether it comes from memory footprint, graph/runtime behavior, or another runtime effect.

This PR can stay focused on the Higgs-specific argument cleanup, the scalability characterization, and the default-size recommendation.

estellaliu233: ## Scalability Results

### `32/32`

| concurrency | median lat (s) | p95 lat (s) | p99 lat (s) | throughput (req/s) | RTF mean | RTF p95 | audio throughput (s/s) | output tok/s |
| 1 | 0.583 | 0.860 | 0.982 | 1.656 | 0.1433 | 0.1661 | 7.045 | 187.7 |
| 2 | 0.674 | 0.998 | 1.246 | 2.868 | 0.1690 | 0.2239 | 12.042 | 321.1 |
| 4 | 0.673 | 0.977 | 1.163 | 5.818 | 0.1659 | 0.2016 | 24.363 | 649.8 |
| 8 | 0.796 | 1.188 | 1.392 | 9.526 | 0.1969 | 0.2450 | 41.309 | 1099.4 |
| 16 | 0.997 | 1.486 | 1.815 | 15.081 | 0.2478 | 0.3151 | 65.406 | 1740.7 |
| 32 | 1.449 | 2.241 | 2.794 | 21.037 | 0.3687 | 0.5357 | 88.223 | 2352.8 |
| 64 | 2.999 | 3.995 | 4.515 | 20.693 | 0.7623 | 1.1144 | 86.734 | 2313.2 |
| 128 | 6.108 | 7.228 | 7.883 | 20.570 | 1.5261 | 2.3123 | 86.227 | 2299.7 |

### `64/64`

| concurrency | median lat (s) | p95 lat (s) | p99 lat (s) | throughput (req/s) | RTF mean | RTF p95 | audio throughput (s/s) | output tok/s |
| 1 | 0.586 | 0.863 | 0.990 | 1.650 | 0.1439 | 0.1672 | 7.020 | 187.0 |
| 2 | 0.674 | 0.997 | 1.176 | 2.881 | 0.1683 | 0.2264 | 12.112 | 323.0 |
| 4 | 0.675 | 1.009 | 1.190 | 5.766 | 0.1669 | 0.2088 | 24.242 | 646.4 |
| 8 | 0.778 | 1.142 | 1.352 | 9.715 | 0.1933 | 0.2338 | 42.033 | 1118.8 |
| 16 | 1.002 | 1.495 | 1.769 | 15.241 | 0.2486 | 0.3170 | 64.992 | 1731.5 |
| 32 | 1.503 | 2.307 | 2.802 | 20.357 | 0.3807 | 0.5519 | 85.390 | 2277.2 |
| 64 | 2.591 | 3.929 | 4.608 | 23.604 | 0.6667 | 1.0719 | 99.043 | 2641.3 |
| 128 | 5.474 | 7.255 | 8.129 | 22.367 | 1.3988 | 2.2466 | 93.675 | 2498.4 |

### `128/128`

| concurrency | median lat (s) | p95 lat (s) | p99 lat (s) | throughput (req/s) | RTF mean | RTF p95 | audio throughput (s/s) | output tok/s |
| 1 | 0.585 | 0.863 | 0.991 | 1.650 | 0.1439 | 0.1676 | 7.020 | 187.1 |
| 2 | 0.675 | 1.003 | 1.208 | 2.880 | 0.1692 | 0.2350 | 12.060 | 321.7 |
| 4 | 0.670 | 0.989 | 1.139 | 5.753 | 0.1655 | 0.2056 | 24.487 | 652.4 |
| 8 | 0.780 | 1.149 | 1.294 | 9.838 | 0.1931 | 0.2313 | 41.984 | 1118.5 |
| 16 | 1.063 | 1.594 | 1.964 | 14.240 | 0.2627 | 0.3367 | 61.846 | 1645.8 |
| 32 | 1.467 | 2.241 | 2.727 | 20.773 | 0.3733 | 0.5398 | 86.945 | 2319.0 |
| 64 | 2.592 | 3.873 | 4.363 | 23.689 | 0.6636 | 1.0822 | 99.444 | 2651.9 |
| 128 | 5.148 | 7.083 | 7.820 | 23.577 | 1.3344 | 2.1731 | 98.728 | 2633.2 |

estellaliu233: higgs_rtf_vs_concurrency_20260613

estellaliu233: ## Low-Concurrency c16 Attribution

I did a lightweight single-run attribution for the c16 behavior, specifically comparing sending 16 requests when CUDA Graph batch size is 16 vs 128.

The direct comparison is:

| config | client concurrency | max_running_requests | cuda_graph_max_bs | throughput | RTF mean | median lat | p95 lat | p99 lat |
| `16/16` | 16 | 16 | 16 | 15.497 | 0.2486 | 1.003s | 1.492s | 1.744s |
| `128/128` | 16 | 128 | 128 | 14.240 | 0.2627 | 1.063s | 1.594s | 1.964s |

Relative to `16/16`, `128/128` changed by:

| metric | delta |
|---|---:|
| median latency | +6.0% |
| p95 latency | +6.8% |
| p99 latency | +12.6% |
| RTF mean | +5.7% |
| throughput | -8.1% |
| output tokens | +3.48% |

I also ran split configurations to isolate the source:

| config | client concurrency | max_running_requests | cuda_graph_max_bs | actual captured batches | throughput | RTF mean | median lat | p95 lat | p99 lat |
|---|---:|---:|---:|---|---:|---:|---:|---:|---:|
| `16/16` | 16 | 16 | 16 | `[1,2,4,8,12,16]` | 15.497 | 0.2486 | 1.003s | 1.492s | 1.744s |
| `16/128` | 16 | 16 | 128 | `[1,2,4,8,12,16]` | 13.962 | 0.2511 | 1.002s | 1.509s | 1.791s |
| `128/16` | 16 | 128 | 16 | `[1,2,4,8,12,16]` | 15.383 | 0.2506 | 1.017s | 1.487s | 1.744s |
| `128/128` | 16 | 128 | 128 | `[1,2,4,8,12,16,24,...,128]` | 14.240 | 0.2627 | 1.063s | 1.594s | 1.964s |

Attribution readout:

- `128/128` was slower than `16/16` by about 5-6% after normalizing by output length / audio duration.
- Output length variation explains about 39% of the raw latency delta.
- Increasing `max_running_requests` alone was nearly flat: `128/16` showed about ~0.8% normalized overhead vs `16/16`.
- One caveat is that `16/128` did not actually capture graph batches above 16. Effective graph coverage appears bounded by `max_running_requests`.
- Therefore, the remaining ~5-6% normalized slowdown appears when graph coverage is actually increased to 128 under `128/128`.
- Given the capture-time cost in the pr description(the median ready time only increased from 56.69s at `16/16` to 60.83s at `128/128`, while peak GPU memory
  increased by about 818 MiB), It is more likely runtime/memory residual from larger graph coverage, and I did not further decompose it in this PR.

benchmarks/benchmarker/utils.py
```diff
@@ -230,6 +230,7 @@ def managed_omni_server(
     model_path: str,
     port: int,
     host: str,
+    extra_args: list[str] | None = None,
```
yxs: breaks the "required-first, optional-last" readability convention

estellaliu233: Thanks. Done.

zhaochenyang20:
Thanks for the thorough sweep — the characterization work here is genuinely useful. A few things before this lands.

**1. Default size & the variance question**

The 16/16 → 32/32 jump at c32 is a real signal, but the rest of the sweep looks single-run, and at this precision I don't think it can support the finer conclusions:

- 64/64 vs 128/128 at c64 differ by 0.6667 vs 0.6636 RTF — under 0.5%, almost certainly noise.
- The c16 "5–6% regression from larger graph coverage" is undercut by your own control: 16/16 and 16/128 have *identical* effective graph coverage (both capped at 16) yet differ ~10% in throughput (15.497 vs 13.962). That 10% is the noise floor, and it's larger than the effect you're attributing. Both "slow" runs (16/128, 128/128) nominally set `cuda_graph_max_bs=128`, but one never captured above 16 — so that's not a self-consistent causal story.

Could you re-run with N≥5 per (config, concurrency) and report mean ± CI, with GPU clocks locked and output length controlled (fixed seed / greedy, or normalize per actual output tokens)? My guess is 64/64 and 128/128 become statistically indistinguishable. The c16 attribution should then be downgraded from a finding to "not separable at current precision" — let's not bake a regression into #781 before it's confirmed.

For the default itself: **64/64** is the right call — it saturates the plateau at c64 (throughput 23.6 vs 32/32's 20.7, a real gap) without low-concurrency downside. 128/128 isn't justified as a default. Note this changes the PR's framing from "default 32" to "default 64" — title, description, and the 32/32 benchmark narrative will need updating.

**2. Drop the single flag — go back to the two standard engine params**

Collapsing into `--higgs-ar-max-batch-size` works against SGLang's engine conventions and hard-codes "must be equal" as a *constraint* when it's only a *recommendation*. Please revert to the two standard params, each independently overridable on the `tts_engine` stage:

- `--max-running-requests`
- `--cuda-graph-max-bs`

Help text should *suggest* keeping them equal, and add fill-in logic: if only one is set, the other defaults to it.

```python
if max_running_requests is None and cuda_graph_max_bs is not None:
    max_running_requests = cuda_graph_max_bs
elif cuda_graph_max_bs is None and max_running_requests is not None:
    cuda_graph_max_bs = max_running_requests
```

This also removes the load-bearing assumption in `_resolve_max_batch_size`: with fill-in, the two are equal whenever either is set, so reading `max_running_requests` for the model-side buffer is always safe.

**3. Kill the constants — default lives in the stage constructor signature**

The four batch-ish constants are redundant and confusing: `DEFAULT_MAX_CONCURRENCY=16`, `DEFAULT_AR_SERVER_MAX_RUNNING_REQUESTS=32`, `DEFAULT_AR_SERVER_CUDA_GRAPH_MAX_BS=32`, `_DEFAULT_MAX_BATCH_SIZE=64`. The `DEFAULT_MAX_CONCURRENCY` comment ("Saturates near c=16... higher concurrency only queues") also directly contradicts this PR's own c64 scaling result.

Let's make the stage executor constructor the single source of truth — no module-level constants:

- CLI flags default to `None`; they only override when the user passes a value.
- The real default (64) lives in the stage constructor signature: `max_running_requests: int = 64`, `cuda_graph_max_bs: int = 64`.
- Delete all four `DEFAULT_*` / `_DEFAULT_*` constants and their `__all__` exports. In `config.py`, drop the now-unused imports and let the constructor defaults apply.

A function-parameter default is the standard way to express this and keeps every entry point — CLI, tests, programmatic — consistent, unlike a CLI-only `default=64` which would diverge for non-CLI callers.

Also: tighten the `_resolve_max_batch_size` fallback — `except Exception` is too broad and silently shrinks the buffer back to 64, which re-triggers the capture mismatch above 64 with no signal. Narrow it and log a warning.

**Tests**

- Rewrite the `test_higgs_ar_batch_size_*` cases for the two independent flags.
- Add a case for the fill-in logic: set only `cuda_graph_max_bs=128`, assert `max_running_requests` becomes 128, and the reverse.
- Update `test_pipeline.py` assertions from `DEFAULT_AR_SERVER_*` to the constructor default (64).

Net: the PR shifts from "add a Higgs-specific flag" to "bump Higgs default to 64/64 + make the two standard engine params independently overridable with fill-in." Happy to look again once it's reshaped — and let's get the two CI failures (unit-test, stage 10 Video-AMME Talker) green before re-review.

estellaliu233:
> Thanks for the thorough sweep — the characterization work here is genuinely useful. A few things before this lands.
> 
> **1. Default size & the variance question**
> 
> The 16/16 → 32/32 jump at c32 is a real signal, but the rest of the sweep looks single-run, and at this precision I don't think it can support the finer conclusions:
> 
> * 64/64 vs 128/128 at c64 differ by 0.6667 vs 0.6636 RTF — under 0.5%, almost certainly noise.
> * The c16 "5–6% regression from larger graph coverage" is undercut by your own control: 16/16 and 16/128 have _identical_ effective graph coverage (both capped at 16) yet differ ~10% in throughput (15.497 vs 13.962). That 10% is the noise floor, and it's larger than the effect you're attributing. Both "slow" runs (16/128, 128/128) nominally set `cuda_graph_max_bs=128`, but one never captured above 16 — so that's not a self-consistent causal story.
> 
> Could you re-run with N≥5 per (config, concurrency) and report mean ± CI, with GPU clocks locked and output length controlled (fixed seed / greedy, or normalize per actual output tokens)? My guess is 64/64 and 128/128 become statistically indistinguishable. The c16 attribution should then be downgraded from a finding to "not separable at current precision" — let's not bake a regression into #781 before it's confirmed.
> 
> For the default itself: **64/64** is the right call — it saturates the plateau at c64 (throughput 23.6 vs 32/32's 20.7, a real gap) without low-concurrency downside. 128/128 isn't justified as a default. Note this changes the PR's framing from "default 32" to "default 64" — title, description, and the 32/32 benchmark narrative will need updating.
> 
> **2. Drop the single flag — go back to the two standard engine params**
> 
> Collapsing into `--higgs-ar-max-batch-size` works against SGLang's engine conventions and hard-codes "must be equal" as a _constraint_ when it's only a _recommendation_. Please revert to the two standard params, each independently overridable on the `tts_engine` stage:
> 
> * `--max-running-requests`
> * `--cuda-graph-max-bs`
> 
> Help text should _suggest_ keeping them equal, and add fill-in logic: if only one is set, the other defaults to it.
> 
> ```python
> if max_running_requests is None and cuda_graph_max_bs is not None:
>     max_running_requests = cuda_graph_max_bs
> elif cuda_graph_max_bs is None and max_running_requests is not None:
>     cuda_graph_max_bs = max_running_requests
> ```
> 
> This also removes the load-bearing assumption in `_resolve_max_batch_size`: with fill-in, the two are equal whenever either is set, so reading `max_running_requests` for the model-side buffer is always safe.
> 
> **3. Kill the constants — default lives in the stage constructor signature**
> 
> The four batch-ish constants are redundant and confusing: `DEFAULT_MAX_CONCURRENCY=16`, `DEFAULT_AR_SERVER_MAX_RUNNING_REQUESTS=32`, `DEFAULT_AR_SERVER_CUDA_GRAPH_MAX_BS=32`, `_DEFAULT_MAX_BATCH_SIZE=64`. The `DEFAULT_MAX_CONCURRENCY` comment ("Saturates near c=16... higher concurrency only queues") also directly contradicts this PR's own c64 scaling result.
> 
> Let's make the stage executor constructor the single source of truth — no module-level constants:
> 
> * CLI flags default to `None`; they only override when the user passes a value.
> * The real default (64) lives in the stage constructor signature: `max_running_requests: int = 64`, `cuda_graph_max_bs: int = 64`.
> * Delete all four `DEFAULT_*` / `_DEFAULT_*` constants and their `__all__` exports. In `config.py`, drop the now-unused imports and let the constructor defaults apply.
> 
> A function-parameter default is the standard way to express this and keeps every entry point — CLI, tests, programmatic — consistent, unlike a CLI-only `default=64` which would diverge for non-CLI callers.
> 
> Also: tighten the `_resolve_max_batch_size` fallback — `except Exception` is too broad and silently shrinks the buffer back to 64, which re-triggers the capture mismatch above 64 with no signal. Narrow it and log a warning.
> 
> **Tests**
> 
> * Rewrite the `test_higgs_ar_batch_size_*` cases for the two independent flags.
> * Add a case for the fill-in logic: set only `cuda_graph_max_bs=128`, assert `max_running_requests` becomes 128, and the reverse.
> * Update `test_pipeline.py` assertions from `DEFAULT_AR_SERVER_*` to the constructor default (64).
> 
> Net: the PR shifts from "add a Higgs-specific flag" to "bump Higgs default to 64/64 + make the two standard engine params independently overridable with fill-in." Happy to look again once it's reshaped — and let's get the two CI failures (unit-test, stage 10 Video-AMME Talker) green before re-review.


  I ran the same N=5 check at c16 for `64/64` vs `128/128`.

  ## Setup

  - Model: `bosonai/higgs-audio-v3-tts-4b`
  - Dataset: EN, 1088 samples
  - Concurrency: 16
  - Seed: 42
  - Runs: 5 per config
  - Metric: mean ± 95% CI, computed as `mean ± 1.96 * std / sqrt(5)`
  
Caveat: I still could not explicitly lock GPU clocks because `nvidia-smi -lgc/-rgc` failed with insufficient permission. I recorded GPU telemetry before and after each run. The sampled SM/memory clocks stayed fixed at 1980/3201 MHz across the recorded points.

  ```bash

  ## Aggregate Results

   config     RTF mean ± 95% CI    throughput req/s ± 95% CI                                                                                                              
  ━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━
   64/64        0.2386 ± 0.0063               15.981 ± 0.343
  ─────────  ───────────────────  ───────────────────────────
   128/128      0.2440 ± 0.0038               15.647 ± 0.205

  ## Per-Run Values

   config     run    completed    failed    RTF mean    throughput req/s                                                                                                  
  ━━━━━━━━━  ━━━━━  ━━━━━━━━━━━  ━━━━━━━━  ━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━
   64/64        1         1088         0      0.2514              15.349
  ─────────  ─────  ───────────  ────────  ──────────  ──────────────────
   64/64        2         1088         0      0.2347              16.125
  ─────────  ─────  ───────────  ────────  ──────────  ──────────────────
   64/64        3         1088         0      0.2356              15.869
  ─────────  ─────  ───────────  ────────  ──────────  ──────────────────
   64/64        4         1088         0      0.2357              16.252
  ─────────  ─────  ───────────  ────────  ──────────  ──────────────────
   64/64        5         1088         0      0.2356              16.309
  ─────────  ─────  ───────────  ────────  ──────────  ──────────────────
   128/128      1         1088         0      0.2497              15.380
  ─────────  ─────  ───────────  ────────  ──────────  ──────────────────
   128/128      2         1088         0      0.2375              15.963
  ─────────  ─────  ───────────  ────────  ──────────  ──────────────────
   128/128      3         1088         0      0.2446              15.703
  ─────────  ─────  ───────────  ────────  ──────────  ──────────────────
   128/128      4         1088         0      0.2448              15.454
  ─────────  ─────  ───────────  ────────  ──────────  ──────────────────
   128/128      5         1088         0      0.2434              15.737

  ## GPU Telemetry Ranges

   config     samples    SM clock MHz    mem clock MHz          power W    temp C       memory MiB                                                                        
  ━━━━━━━━━  ━━━━━━━━━  ━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━  ━━━━━━━━  ━━━━━━━━━━━━━━━
   64/64           10       1980-1980        3201-3201    121.60-244.21     30-37    122298-124242
  ─────────  ─────────  ──────────────  ───────────────  ───────────────  ────────  ───────────────
   128/128         10       1980-1980        3201-3201    117.70-243.03     30-44    122880-124820

  Both the RTF and throughput confidence intervals overlap at c16. I think this supports treating the earlier c16 difference as not separable at the current
  precision, rather than as a confirmed larger-graph-coverage regression.

benchmarks/benchmarker/utils.py
```diff
@@ -231,6 +231,7 @@ def managed_omni_server(
     port: int,
     host: str,
     log_file: Path | None,
+    extra_args: list[str] | None = None,
```
zhaochenyang20: do not have any ambiguous argument. What's the `extra_args` for? Make it explicit.

benchmarks/eval/benchmark_tts_seedtts.py
```diff
@@ -429,6 +435,18 @@ def _config_from_args(args: argparse.Namespace) -> TtsSeedttsBenchmarkConfig:
...
+def _build_higgs_ar_server_args(
+    max_running_requests: int | None,
+    cuda_graph_max_bs: int | None,
+) -> list[str]:
+    args: list[str] = []
+    if max_running_requests is not None:
+        args.extend(["--max-running-requests", str(max_running_requests)])
+    if cuda_graph_max_bs is not None:
+        args.extend(["--cuda-graph-max-bs", str(cuda_graph_max_bs)])
+    return args
```
zhaochenyang20: I think this function should be more general, not only for `--cuda-graph-max-bs` and `--max-running-requests`.

zhaochenyang20: higgs server has rather more parameters

sglang_omni/cli/serve.py
```diff
@@ -1044,6 +1087,32 @@ def serve(
...
+    max_running_requests: Annotated[
+        int | None,
+        typer.Option(
+            "--max-running-requests",
+            "--max_running_requests",
+            help=(
+                "Higgs TTS AR server max_running_requests override for the "
+                "tts_engine stage. Recommended to keep equal to "
+                "--cuda-graph-max-bs; if only one is set, the other is filled "
+                "with the same value."
```
zhaochenyang20: in short, this parameter shall not only for higgs. So I think the naming and function logits should be rather general. To all the models, qwen3 omni and TTSs.

sglang_omni/models/higgs_tts/model.py
```diff
@@ -42,7 +45,20 @@ class HiggsGenParams:
...
+def _resolve_max_batch_size(configured_max_batch_size: int | None) -> int:
+    try:
+        from sglang.srt.server_args import get_global_server_args
+
+        return int(get_global_server_args().max_running_requests)
+    except (ImportError, AttributeError, TypeError, ValueError) as exc:
+        fallback = int(configured_max_batch_size or 64)
+        logger.warning(
+            "Falling back to configured Higgs max_batch_size=%s because SGLang "
+            "global server args are unavailable: %s",
```
zhaochenyang20: never use `%s`. Use f-string

zhaochenyang20 (changes requested):
In short, this PR should be rather general. Modify the base class, not for Higgs alone.

benchmarks/benchmarker/utils.py
estellaliu233: I updated it to max_running_requests: int | None = None and cuda_graph_max_bs: int | None = None

benchmarks/eval/benchmark_tts_seedtts.py
estellaliu233: Since extra_args is deleted, this helper function is useless.

sglang_omni/cli/serve.py
estellaliu233: done and added unit tests to qwen3 omni and other tts models.

sglang_omni/models/higgs_tts/model.py
estellaliu233: done

sglang_omni/models/higgs_tts/model.py
```diff
@@ -94,7 +108,7 @@ def __init__(
         config: HiggsMultimodalQwen3Config,
         quant_config=None,
         prefix: str = "",
-        max_batch_size: int = _DEFAULT_MAX_BATCH_SIZE,
+        max_batch_size: int = 64,
```
zhaochenyang20: could we set the default parameter in higgs' stage configs, rather than here?

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -255,7 +252,7 @@ def create_audio_encoder_executor(
     device: str = "cuda:0",
     dtype: str = "bfloat16",
     num_codebooks: int = 8,
-    max_batch_size: int = DEFAULT_MAX_CONCURRENCY,
+    max_batch_size: int = 16,
```
zhaochenyang20: put it 64

zhaochenyang20: align with other parameters

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -410,7 +409,7 @@ def create_vocoder_executor(
     *,
     device: str = "cuda:0",
     dtype: str = "bfloat16",
-    max_batch_size: int = DEFAULT_MAX_CONCURRENCY,
+    max_batch_size: int = 16,
```
zhaochenyang20: what's this parameter for?

zhaochenyang20 (changes requested):

sglang_omni/models/higgs_tts/model.py
estellaliu233: For the Higgs model-side default: I removed the model-level `max_batch_size` parameter. The serving default now lives in the Higgs `tts_engine` stage constructor as
  `max_running_requests=64` / `cuda_graph_max_bs=64`, and the model-side sampler buffer size is derived from SGLang `max_running_requests`.

sglang_omni/models/higgs_tts/stages.py
estellaliu233: I removed `audio_encoder` `max_batch_size` / `max_batch_wait_ms` because this stage uses `SimpleScheduler` without `batch_compute_fn`, so those batching parameters were
  dead code.

sglang_omni/models/higgs_tts/stages.py
estellaliu233: I renamed vocoder `max_batch_size` to `vocoder_decode_batch_size` to make the semantics explicit. This parameter controls the non-streaming vocoder `decode_batch` call,
  not the SGLang AR `max_running_requests` / `cuda_graph_max_bs`, so I kept the existing default for now.
  
  
  I think a broader repo-wide cleanup of stage-local batch naming across TTS / Qwen3-Omni would be useful, but I kept this PR scoped to the Higgs AR default and the
  directly related Higgs cleanup.

sglang_omni/cli/serve.py
```diff
@@ -136,19 +136,40 @@ def _resolve_talker_sglang_stage(
...
 
+def _resolve_generation_sglang_stage(
+    pipeline_config: PipelineConfig,
+    *,
+    flag_name: str,
+) -> str:
+    stage_name = type(pipeline_config).talker_sglang_role_to_stage().get("talker")
+    if stage_name is not None:
+        return stage_name
+    if any(stage.name == "tts_engine" for stage in pipeline_config.stages):
```
zhaochenyang20: the PR now exposes standard SGLang generation-stage batch knobs, but `_resolve_generation_sglang_stage()` still hard-codes the `tts_engine` stage name as a CLI-level fallback. That happens to cover Higgs, Qwen3-TTS, MOSS, and FishAudio/S2 pipelines whose stage is named `tts_engine`, and it covers Qwen3-Omni speech through `talker_sglang_role_to_stage()`, but it misses the existing Voxtral TTS pipeline: Voxtral is also an SGLang-backed TTS generation stage, its stage is named `generation`, and its factory still hard-codes `max_running_requests=16`, so passing `--max-running-requests` currently raises `BadParameter`. This means the generic generation-stage contract is still a CLI guess rather than an explicit `PipelineConfig` role map.

zhaochenyang20: Derived from the root-cause issue above: the new tests are no longer only Higgs decode-mode tests. The file now imports Qwen3-TTS and Qwen3-Omni speech and verifies generation server-args overrides, but the module docstring still says this is a Higgs TTS decode-mode test module. That placement makes future maintainers read the generic CLI override path as Higgs-private behavior, and it makes systematic coverage for sibling pipelines such as Voxtral, MOSS, and FishAudio harder to add.

zhaochenyang20: Move the `apply_generation_server_args_cli_override()` tests into a generic CLI/serve test file, for example `tests/unit_test/cli/test_generation_server_args.py` or the existing serve CLI test area, and cover the explicit role map, the migrated TTS configs, Qwen3-Omni speech, unsupported configs, and either Voxtral support or an intentionally unsupported Voxtral behavior in the same matrix. The Higgs decode-mode test file should keep only `decode_mode` and `async_lookahead_min_batch_size` behavior for Higgs/MOSS.

zhaochenyang20 (changes requested):

sglang_omni/cli/serve.py
estellaliu233: Thanks. I updated the generation server-args path to use an explicit `PipelineConfig` contract instead of a CLI-side stage-name fallback.

  Changes:
  - Added `PipelineConfig.generation_sglang_role_to_stage()`.
  - Removed the `tts_engine` fallback from `_resolve_generation_sglang_stage()`.
  - Migrated the SGLang-backed generation configs to declare their generation stage explicitly:
    - Higgs / Qwen3-TTS / MOSS / MOSS-Local / FishAudio S2-Pro -> `tts_engine`
    - Qwen3-Omni speech -> `talker_ar`
    - Voxtral TTS -> `tts_generation`
  - Added `server_args_overrides` support to the Voxtral generation factory so the CLI override is actually applied to `build_sglang_server_args()`.
  - Moved generation server-args tests out of the Higgs decode-mode test file into `tests/unit_test/serve/test_generation_server_args.py`.

  The new test matrix covers:
  - explicit role-map behavior;
  - migrated TTS configs;
  - Qwen3-Omni speech;
  - Voxtral TTS;
  - unsupported configs, including the fail-fast case when a config has a `tts_engine` stage but does not declare `generation_sglang_role_to_stage()`.

estellaliu233: @zhaochenyang20  Just a gentle follow-up. I believe this PR is mostly converged on my side. I’ll be traveling over the next two weeks soon but can still make small fixes if needed. Please let me know if there’s anything else you’d like me to address for this PR. Thanks!

estellaliu233:
## Follow-up TODOs

  This PR is now focused on the Higgs AR default bump and the generic generation-stage override path. There are a few related cleanup / generalization items that are worth
  tracking separately.

  1. **Audit CUDA graph batch sizing for other SGLang-backed generation stages.**

     This PR fixed the Higgs-side mismatch where model-side CUDA graph / sampler buffers were effectively sized around the old internal max-batch assumption. Given the observations from #721, other models
     may have similar assumptions around `max_running_requests`, `cuda_graph_max_bs`, or stage-local batch buffers.

     Follow-up work:
     - check other SGLang-backed generation stages for hidden fixed batch-size assumptions;
     - add a lightweight server utility / validation path that can verify the effective CUDA graph batch coverage and model-side buffer sizing for a given stage;
     - use the same check for related issues such as #721 and future CUDA graph tuning work.

  2. **Continue Higgs batch naming cleanup.**

     This PR already removes the stale Higgs constants and renames the vocoder batch knob. A remaining cleanup example is the internal Higgs model naming: `_max_batch_size`
     is now resolved from SGLang `max_running_requests` and is used to size the sampler pool, so a follow-up can rename it to reflect that semantics more directly.

     Follow-up work:
     - audit the remaining Higgs batch-related names after this PR;
     - remove or rename any stale Higgs parameters that no longer affect behavior;

  3. **Audit cross-model generation batch conventions.**

     The new `generation_sglang_role_to_stage()` path makes this PR more general, but there is still room to make model naming and parameter plumbing more consistent across
     Higgs, MOSS, Qwen3-TTS, Qwen3-Omni, Voxtral, FishAudio, etc.

     Examples to investigate:
     - Voxtral sets `max_running_requests=16` and `torch_compile_max_bs=16` but does not explicitly set `cuda_graph_max_bs`, while several TTS factories set all three
     related knobs.
     - FishAudio S2-Pro defaults `max_running_requests=64` but `torch_compile_max_bs=16`, and also sizes decoder/model-side buffers from `server_args.max_running_requests`.

     Follow-up work:
     - decide which generation-stage batch defaults should be standardized and which should remain model-specific;
     - reduce model-specific naming differences where they are not semantically meaningful;

  I suggest handling these as separate follow-up PRs. Combining all of them into this PR would make the scope too large and harder to review.

cc: @zhaochenyang20

zhaochenyang20 (approved):

Ccyest:
## Cross-validation on 1× H100 80GB

Cross-validating this PR's H200 sweep on **1× NVIDIA H100 80GB HBM3**.

- **Dataset:** SeedTTS-EN full set (N=1088), throughput-only
- **Method:** n=3 mean per point; one server per config reused across the concurrency sweep; matched config `max_running_requests == cuda_graph_max_bs`; client concurrency raised until RTF reaches 1
- Column headers `16/16 … 128/128` are `max_running_requests / cuda_graph_max_bs`; the first column is **client concurrency** (`--max-concurrency`)

image

### RTF mean (n=3) — client concurrency × config

| client concurrency | `16/16` | `32/32` | `64/64` | `128/128` |
|---:|---:|---:|---:|---:|
| 16 | 0.2489 | 0.2488 | 0.2472 | 0.2464 |
| 32 | 0.5167 | 0.3537 | 0.3530 | 0.3534 |
| 48 | 0.7848 | 0.5612 | 0.4648 | 0.4623 |
| 64 | 1.0427 | 0.7519 | 0.5850 | 0.5783 |
| 96 | – | 1.1371 | 0.8892 | 0.8628 |
| 128 | – | – | 1.1909 | 1.1457 |

### Throughput req/s (n=3) — client concurrency × config

| client concurrency | `16/16` | `32/32` | `64/64` | `128/128` |
|---:|---:|---:|---:|---:|
| 16 | 15.23 | 15.35 | 15.36 | 15.55 |
| 32 | 15.28 | 21.30 | 20.64 | 21.71 |
| 48 | 15.17 | 20.58 | 24.04 | 23.75 |
| 64 | 15.25 | 20.44 | 26.33 | 25.05 |
| 96 | – | 20.24 | 25.12 | 26.84 |
| 128 | – | – | 24.93 | 26.79 |

**Takeaways (consistent with the PR):**
- RTF=1 crossing moves right as the batch size grows: `16/16` c≈61 → `32/32` c≈85 → `64/64` c≈108 → `128/128` c≈112.
- `64/64` and `128/128` are nearly identical (peak throughput 26.3 vs 26.8 req/s) — supports keeping `64/64` as the default.
- `128/128` CUDA-graph capture **did not OOM on H100 80GB** (the ~122 GB peak was H200; on H100 the KV cache auto-sizes down), so the 128 capture fix holds here. All 24 points: 1088/1088 completed, 0 failures.

Ccyest:
### Why `128/128` adds no gain over `64/64`
Short answer: on a single GPU the Higgs pipeline can't fill a decode batch larger than ~78 at these concurrencies, so the extra graphs captured for `64 < bs ≤ 128` are essentially never replayed. Memory is nowhere near the limit.

**Capture cost is tiny and VRAM headroom is large (no OOM):**

| config | captured batch sizes (#graphs) | capture VRAM | capture time | free VRAM after capture |
|---|---|---:|---:|---:|
| `16/16` | [1,2,4,8,12,16] (6) | 0.11 GB | 2.28 s | 11.53 GB |
| `32/32` | …+24,32 (8) | 0.13 GB | 2.59 s | 11.48 GB |
| `64/64` | …→64 (12) | 0.29 GB | 2.65 s | 11.35 GB |
| `128/128` | …→128 (20) | 0.80 GB | 2.89 s | 10.81 GB |

All configs: weights 7.60 GB + KV cache auto-sized to ~58.6 GB, leaving ~11.6 GB before capture. On H100 80 GB the KV pool shrinks to fit, so the `128` capture does **not** OOM — even at `128/128` there is still **10.81 GB free** afterward. (The PR's ~122 GB peak was an H200 figure; the 128 fix on H100 is a shape-correctness fix, not a memory one.)

**No memory pressure at runtime, for any config:**
- Peak KV occupancy ~3% (`token usage ≈ 0.03`, ≈12.9k / 427k tokens).
- **Zero** eviction / preemption / OOM across all 4 servers' logs.

**The real ceiling is the decode batch, not memory** — peak concurrent decode `#running-req`:

| client concurrency | `64/64` | `128/128` |
|---:|---:|---:|
| 64 | 64 | 64 |
| 96 | 64 | **78** |
| 128 | 64 | **78** |

`128/128` never fills past **78** even at `c = 128`: requests spend much of their life in non-decode stages (ref-audio encode / vocode) or queued.
Actually the mean concurrency of decode is around 50.

zhaochenyang20:
## Q1. Beyond 64: compute saturation or KV cache eviction?
- 64/64 and 128/128 overlap at peak (~26.3 vs ~26.8 req/s); RTF=1 crossing only c≈108→c≈112.
- We assumed "memory bound." Unverified — could be KV cache eviction/preemption instead.
- Check at c96/c128, 64/64 vs 128/128: preemption/eviction counts, KV cache util, radix hit rate, waiting-queue depth. Saturation → eviction ≈ 0; cache-bound → preemption visible + util pinned.

## Q2. Past max concurrency: why does throughput decline, not plateau?
- Right panel: 16/16 flat; 32/32 and 64/64 roll over ~5% after peak; only 128/128 plateaus. H200 same direction.
- We keep `max_running_requests == cuda_graph_max_bs` → no eager fallback → excess should just queue → expected plateau.
- Contradiction: small cap (16/16) is the flat one, large caps decline. (Saw similar silent eager fallback on MOSS.)
- Check: when does replay→eager actually trigger; per-step overhead from padding/scheduling/queue at high concurrency; confirm decline is signal not n=3 variance.
- Hypothesis: same root as Q1 — large cap holds more concurrent reqs in KV → more eviction → decline; 16/16 holds 16 → flat.

## Q3. Capture cost + memory-leak check
- Capturing larger looks nearly free: startup 56.69s→60.83s, H200 peak mem +~236 MiB, H100 128/128 no OOM (KV auto-sizes down), 24/24 points 0 failures.
- Quantify: capture cost = memory (graph buffers) + capture time (startup), as scaling curves vs `cuda_graph_max_bs`.
- Leak check: repeated restart / config cycling in one run — does peak mem climb monotonically?

zhaochenyang20:
**1. Why doesn't the slope change?**

With `max_running_requests == cuda_graph_max_bs`, the actual running batch is capped by `max_running_requests` and never exceeds `cuda_graph_max_bs`. As client concurrency increases further, the excess requests only sit in the waiting queue; the running batch always stays within captured sizes, always hits graph replay, and never falls back to eager. So the slope of the RTF/throughput curves does not distort, and the linear RTF rise past high concurrency is purely queueing delay.

**2. Throughput regresses instead of plateauing**

I'm mainly thinking about this along the lines of prefill/decode interleaving. Once decode throughput saturates, the waiting queue keeps deepening; each time a decode slot frees up, the scheduler immediately prioritizes the prefill of a new request, so prefill and decode interleave heavily. This dilutes the effective decode duty cycle, which lowers token/s and instead causes throughput to regress after reaching its peak rather than holding at the peak.

Issue #757, opened 2026-06-12
[Perf] MOSS-TTS Local v1.5: vectorize per-frame launch orchestration into [P] pool tensors


JiaxinD:
**Motivation**

Following @AkazaAkane's review on #745 (https://github.com/sgl-project/sglang-omni/pull/745#discussion_r3398199811): the next-step-critical fields the launch path reads each frame, the active rows and `generation_steps`, should become `[P]` pool tensors so the launch path can gather them in one shot with `row_t` (for example `gen_steps = pool.generation_steps[row_t]`), with the state updates folded into the state commit. Today `_collect_frame` still runs a per-request Python loop every frame to assemble this state, which is the remaining scalar control flow on the decode hot path.

**Dependencies**

Builds on the #745 state pool (the row-indexed `[P]` buffers and the rid-to-row map). Depends on the launch and resolve split from #734, so the journal and output collection stay on the resolve side, per @AkazaAkane's original suggestion.

**Payoff**

Clears the remaining per-frame Python control flow out of the #736 CUDA Graph capture region, so the launch path becomes a pure tensor gather plus graph replay with no scalar host loop.

AkazaAkane:
Thanks for opening the issue. I will take this.

PR #758, opened 2026-06-12
[Perf] MOSS-TTS-Local v1.5 async decode (one-step lookahead) (#734)

JiaxinD:
> Addresses #734, follows #735 (PR-A: #745).

Enables the #590 one-step-lookahead async decode on the MOSS-TTS-Local v1.5 path,
built on the #735 state pool (PR-A: #745).

## Model-agnostic-layer changes (disclosed upfront)
- `omni_scheduler.py`: `forward_ct += 1` per forward (restores SGLANG_TEST_RETRACT;
  standalone omni fix); `_resolve_and_process` drops retracted reqs (double-free guard).
- `base.py`: `execute_resolve` skip_rids extended with `is_retracted`.
- `serve.py`: async-decode CLI gating (set-ized, per #695).
Higgs note: the retracted-drop is a shared crash-fix only; Higgs frame accounting
for a retracted in-flight frame is a separate issue.

## Correctness: the bs=1 stop-boundary path

The one-step lookahead defers a decode step's host-side collect by one iteration.
The frame's radix token ids (which carry the stop id `== end_id` for a stop frame)
are published at launch as `result.next_token_ids`; the base aliases that tensor
onto `schedule_batch.output_ids` so the next step can build its input. The next
decode step folds `output_ids` into `forward_batch.input_ids` and overwrites it in
place. Under lookahead the next launch runs before this step's lagged resolve, so
without a snapshot the published stop id is gone by the time resolve hands it to
`process_batch_result`: the eos finish is lost and a bs=1 request runs to the
4096-frame cap. `post_decode_launch` therefore takes a device snapshot of the
published ids (a GPU to GPU clone, no host sync) and `post_decode_resolve` restores
it. Only bs=1 eos-stop under lookahead is exposed (audio codes come from the
journal, and the multi-request scenarios finish on length), which is why the
sequential gate never caught it. CPU regression:
`test_async_resolve_preserves_stop_id_through_output_ids_clobber` (commit be17872).

## Verification

PR-B is default-OFF. Correctness is an ON/OFF gate that compares vocoder-entry
`state.audio_codes` per seed, with two tiers: S0/S1 are checked for exact
bit-identity (bs=1 is deterministic and lookahead-forced); S2-S5 are structural
(no crash, every seed dumped, lookahead engaged via the hit/miss counter, every
request finished). Exact bit-identity is incompatible with real-concurrency
lookahead for multi-request batches: an OFF-vs-OFF control (two pure-sync
concurrent runs, zero lookahead) of the same seeds already diverges in frame count
and values, because per-step batch composition is timing-dependent and the binary
stop head amplifies the low-bit hidden drift. The full methodology and the
OFF-vs-OFF control are recorded in `test_gate_scenarios.py`'s module docstring.

One harness fix en route: `max_new_tokens` and `repetition_penalty` are top-level
`CreateSpeechRequest` fields, not `extra_body`, so the old client dropped both,
which is why S5's rep-penalty had not been routing to sync.

**CPU unit tests: PASS** at `8128136` (the rebased head). 136 passed across
`tests/unit_test/moss_tts_local/`, `tests/unit_test/pipeline/test_async_decode.py`
and `tests/unit_test/pipeline/test_scheduler.py`, covering the launch/resolve
hooks, the C5 sampling-step counter, lookahead routing, the launch/resolve versus
sync parity test, and the gate scenario contract (`test_gate_scenarios.py`).

**GPU gates: PASS** (concurrent driver, counter-asserted). S1 stays exact; S2-S5
are structural; S5's binding check is `counter == 0` (whole-batch sync routing):

| scenario | criterion | counter (hit+miss) | result |
| --- | --- | --- | --- |
| S0 (sync self-compare) | exact | n/a | PASS |
| S1 (bs=1 lookahead) | exact | hit=35 | PASS (bit-identical) |
| S2 (bs=4 concurrent) | structural | hit=39 | PASS |
| S3a (concurrent transition) | structural | hit=43 | PASS |
| S3b (cross-bucket transition) | structural | hit=30 | PASS |
| S4 (retraction) | structural | hit=44 | PASS |
| S5 (rep-penalty sync routing) | structural | hit+miss=0 | PASS (sync routing) |

## Performance

This section is mechanism validation, not a gains table. None of the numbers below should be read as a speedup, and none as a regression.

**Conclusion: within noise, no resolvable speedup, no regression.** On a quiet H100 ABAB sweep (OFF vs ON, n=5 rounds, 95% CI), every ON-vs-OFF delta at every concurrency from c1 to c16 has overlapping confidence intervals. There is no resolvable speedup and no resolvable regression. Async decode here is a correctness contract plus a reserved overlap slot, not a present win.

**The mechanism executes as designed.** The lookahead query-hit rate is 100% (hit=400, miss=0): every launched step's GPU work is already done by the time its lagged resolve queries it. This proves the path runs exactly as specified, the one-step lookahead engages and resolves cleanly. It is a path-correctness result, not a host-overlap win. The 100% hit rate says the machinery works; it does not say the host had a tail to hide.

**Within-noise CI table (c1 to c16, H100, ABAB OFF/ON, n=5, 95% CI).** Labelled within noise, not gains. Build f09744b (sglang 0.5.8), cards 4=AR / 5=codec, node1, samples=50, first fully-quiet window. The signed deltas below are CI-overlapping differences, not directional results: a nonzero sign here is noise, not a slowdown or a speedup.

| c | rtf OFF | rtf ON | p99 OFF | p99 ON | status |
| 1 | 0.1872 +/-0.0030 | 0.1873 +/-0.0013 | 1.332 +/-0.020 | 1.332 +/-0.031 | within noise |
| 2 | 0.2086 +/-0.0021 | 0.2141 +/-0.0042 | 1.516 +/-0.061 | 1.526 +/-0.053 | within noise |
| 4 | 0.2954 +/-0.0064 | 0.3037 +/-0.0317 | 2.089 +/-0.161 | 2.068 +/-0.193 | within noise |

Every ON/OFF pair above has overlapping CIs. The c1 to c4 rows carry tight CIs (+/-1 to 2% on the stat metrics), consistent with the near-empty host tail measured below.

**c8 and c16 are HELD.** At high concurrency on this box the measurement noise exceeds any effect: CI half-widths are too wide to resolve (for example c8 OFF rtf +/-18%, p99 +/-32%; c16 ON rtf +/-13%), versus +/-1 to 2% at c1 to c4, and round-to-round spread reaches roughly 40% (c8 OFF rtf 0.4465 to 0.6299). The slow rounds line up with intermittent background-load spikes, and the p99 deltas swing in opposite directions across c8 and c16, which is noise rather than a trend. Per the no-suspect-tier rule, c8 and c16 are not placed in any table. Raw data is archived and available on request.

**Why neutral.** In-process profiling on the head stack (8128136, sglang 0.5.12.post1) shows the host work that the lookahead overlaps, the deferred resolve collect, is about 0.06 ms per step on MOSS (resolve 0.061 ms; the adjacent tolist 0.029 ms and journal 0.014 ms are similarly tiny), against a host step of about 10.4 ms. After the GPU radix hash (#745) there is almost no host tail left to hide, so async is approximately equal to sync on this path today. The launch/resolve machinery is therefore a correctness contract (the bs=1 stop-id device snapshot) plus an overlap slot reserved for later, not a current throughput lever.

**Two-stage plan.** Merge PR-B now for correctness, default-OFF. The default-flip is a separate, later decision driven by data, taken once #757 vectorization makes the resolve path heavy enough that the reserved overlap slot has substance to hide.

**Build and stack note.** The c1 to c16 sweep ran on f09744b (sglang 0.5.8, pre-#698 rebase). The profiling that gives the roughly 0.06 ms host tail ran on the head's 8128136 / sglang 0.5.12.post1 stack. These are different builds; any ship-grade perf number should be re-taken on the head stack, and none is claimed here.


sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -200,7 +244,9 @@ def _collect_frame(
         audio_top_k = params["audio_top_k"]
         sampling_seeds = params["seeds"]
         gen_steps = torch.tensor(
-            [int(d.generation_steps) for d in datas], dtype=torch.long, device=device
+            [self._advance_sampling_position(d) for d in datas],
```
Hayden727: [P1] This advances `sampling_steps` for every `_run_frame_decode()` row before chunked rows are filtered by `finalize_skip_rids()`. For non-final chunked prefill, `_finalize()` intentionally skips `generation_steps` so chunked prefill stays bit-identical to single-shot, but `sampling_steps` still moves. A three-chunk prefill with `is_chunked=(2, 1, 0)` therefore samples positions `[0, 1, 2]`, so the final emitted prefill frame uses position `2` instead of the single-shot position `0`. This affects sync/default-off behavior too. Please either avoid advancing `sampling_steps` for non-final chunked rows or roll it back, and add a test that checks the final chunk's sampling position matches single-shot.

scripts/gate_async_decode.py
```diff
@@ -0,0 +1,245 @@
...
+                "repetition_penalty": r.repetition_penalty,
+            },
+        }
+        req = urllib.request.Request(
+            url,
+            data=json.dumps(body).encode(),
+            headers={"Content-Type": "application/json", "x-request-id": rid},
+        )
+        try:
+            with urllib.request.urlopen(req, timeout=600) as resp:
```
Hayden727: [P2] The S2-S5 gate scenarios claim to validate batched async behavior (`bs=4`, `bs=7->6`, `bs=2->1`, mixed rep-penalty routing), but this loop sends one request and waits for its full response before sending the next. That likely exercises single-request runs instead of the intended scheduler batches, so a PASS would not prove lookahead batching, transition overrun, retraction under batch pressure, or S5 sync routing. Please send the scenario requests concurrently and assert the lookahead hit/miss counters for the intended path (S5 should enforce hits == 0, not just mention it in a comment).

Hayden727 (changes requested):
Requesting changes for two correctness/validation issues I found while tracing the MOSS async-decode path. I checked the latest head (`fa5e7181094dc54ba20b97ca3866cc726a79a83a`); local pytest could not run in my environment because `torch` is not installed, so the evidence below is from code-path tracing plus a minimal state-machine counterexample.

sglang_omni/models/moss_tts_local/model_runner.py
JiaxinD: Fixed in f09744b. `_run_frame_decode` now advances the launch-side counter only for emitted rows: a non-final chunked row takes a read-only position `max(sampling_steps, generation_steps)` instead of advancing it, mirroring how `finalize_skip_rids` holds `generation_steps` for those rows. So a three-chunk `is_chunked=(2, 1, 0)` prefill samples its final emitted frame at the single-shot position `0`, not `2`.

Since this is on the sync / default-off path too, I re-ran the full `tests/unit_test/moss_tts_local` suite plus the S0 sync gate after the change (green).

Added `test_chunked_rows_do_not_advance_sampling_steps`: it drives `_run_frame_decode` for a single-shot prefill versus a three-chunk prefill and asserts the final chunk's sampling position (`sampling_steps`) matches single-shot in both.

scripts/gate_async_decode.py
JiaxinD: You are right. The sequential send kept the server at `bs=1`, below `async_decode_min_batch_size` for every scenario but S1, so the ON arm silently took the synchronous fast path and S2-S5 proved bit-identity of sync against sync. Only S1 (forced to `min_batch_size=1`) actually engaged the lookahead, which is also why S1 alone surfaced a real `bs=1` stop-boundary bug: the stop frame publishes `end_id` as `next_token_ids`, the base aliases that tensor onto `schedule_batch.output_ids`, and the next decode step overwrites it in place; under lookahead the next launch runs before this step's lagged resolve, so resolve never saw `end_id` and the request ran to the 4096-frame cap. Fixed in be17872 by snapshotting the published ids on device at launch and restoring them at resolve (no D2H), with a CPU regression test.

Rewrote the gate (b88310a):

* requests now POST concurrently behind a `Barrier` so they form a real batch;
* every ON arm asserts the lookahead hit/miss counter (`MOSS_GATE_QUERY_DUMP` -> `--counter-file`). S1-S4 must show `hit+miss > 0`; S5 enforces `hit+miss == 0` (whole-batch sync routing) as a hard assertion, not a comment.

I also had to tier the criteria. An OFF-vs-OFF control (two pure-sync concurrent runs, no lookahead at all) of the same seeds already diverges in frame count and values (e.g. seed 1002: 49 vs 36 frames; seed 1000: same frame count, different values). Per-step batch composition is timing-dependent, so the CUDA-graph bucket trajectory varies run-to-run and the binary stop head amplifies the low-bit hidden drift into different frame counts, with zero lookahead involved. Exact bit-identity and real-concurrency lookahead coverage are therefore mutually exclusive for multi-request scenarios, and only `bs=1` (S1) gets both. So S1 stays exact (the anchor), and S2-S5 are structural: no crash, every seed produces a dump, lookahead engaged via the counter (S5 `== 0`), every request finishes. The verification methodology is recorded in `test_gate_scenarios.py`'s module docstring.

Results under the new methodology:

| scenario | criterion | counter | result |
| --- | --- | --- | --- |
| S1 (bs=1 lookahead) | exact | hit=35 | PASS (bit-identical) |
| S2 (bs=4 concurrent) | structural | hit=39 | PASS |
| S3a (concurrent transition) | structural | hit=43 | PASS |
| S3b (cross-bucket) | structural | hit=30 | PASS |
| S4 (retraction) | structural | hit=44 | PASS |
| S5 (rep-penalty sync routing) | structural | hit+miss=0 | PASS (sync routing) |

One harness fix en route: `max_new_tokens` and `repetition_penalty` are top-level `CreateSpeechRequest` fields, not `extra_body`, so the old client dropped both, which is why S5's penalty had not been routing to sync.

Hayden727: Rechecked current head `63c6e175de7ec6c8c5e292dc68580d765aad8ccf`. The gate rewrite described in the reply is not present in the current PR head: after fetching the PR, `be17872c627a292124888a6175b913a9366cb299` and `b88310ad099ff6d9eb4f2d040c289b1bc8e30d94` are not reachable/available, `docs/design/async_decode_gate.md` is absent, and `send_scenario()` still sends requests sequentially with blocking `urlopen(...).read()`. It also still places `max_new_tokens` and `repetition_penalty` under `extra_body`, which the reply says is the old broken shape. So this thread is not fixed in the code currently on the PR; please push/rebase the gate rewrite onto this branch before relying on the S1-S5 results.

Hayden727: Rechecked the latest PR head: the gate issue is addressed by the updated concurrent/structural gate design, with the GPU harness moved out of the in-tree diff and the CI-testable contract covered by `tests/unit_test/moss_tts_local/test_gate_scenarios.py`. I also pushed `326f8c3b` to tighten the counter parser so malformed counter files fail instead of being treated as `(0, 0)`. Local validation: `python -m pytest tests/unit_test/moss_tts_local/test_gate_scenarios.py -q` -> 12 passed; `py_compile` and formatting checks passed. The model pipeline test still cannot run in this local env because `torch` is not installed.

sglang_omni/cli/serve.py
```diff
@@ -14,8 +14,13 @@
 
 _STAGE_TOGGLE_MODE = Literal["default", "on", "off"]
 _QWEN_COLOCATED_CONFIG_CLASS = "Qwen3OmniSpeechColocatedPipelineConfig"
-_HIGGS_ASYNC_DECODE_FACTORY = (
-    "sglang_omni.models.higgs_tts.stages.create_sglang_tts_engine_executor"
+# Set-ized (was a single Higgs constant) so the --async-decode CLI gate can
+# admit more than one model runner (#695 form).
```
Ratish1: nit: we can remove this comment

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -159,6 +180,19 @@ def _collect_frame(
         del forward_batch
         if not requests:
             return
+        rows, end_id = self._run_frame_decode(result, requests)
+        # Radix key is a capture-safe GPU hash (#745): a device op, no host sync.
```
Ratish1: nit: no need to mention PR numbers in comments here

sglang_omni/models/moss_tts_local/model_runner.py
Ratish1, line 107: Reset `sampling_steps` on every re-prefill, not only when `output_rows` is non-empty. Under async lookahead, `_run_frame_decode()` advances `sampling_steps` at launch, while `generation_steps` only advances at resolve. If KV pressure retracts a request after its first launch but before that lagged step resolves, `_resolve_and_process()` drops the retracted row, `generation_steps` remains 0, and `output_rows` is still empty. This branch then skips the reset, so the resumed prefill samples its first emitted frame at RNG position 1 instead of the sync/single-shot position 0.

Move `data.sampling_steps = int(data.generation_steps)` and `reset_for_refill()` out of the `if data.output_rows:` block, or otherwise key the reset on refill/retraction state rather than generated-row count. Please add a CPU regression where `generation_steps=0`, `sampling_steps=1`, `output_rows=[]`, and a refill brings `sampling_steps` back to 0 before the next `_run_frame_decode()`.

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -268,34 +322,64 @@ def sample_audio(logits: torch.Tensor, channel: int) -> torch.Tensor:
...
+        the stop id and silently dropping a bs=1 eos finish (4096-frame runaway).
+        The clone preserves it; resolve swaps it back.
+        """
+        del forward_batch
+        if not requests:
+            return None
+        rows, end_id = self._run_frame_decode(result, requests)
+        next_token_ids = self._row_radix_token_ids(rows, rows[:, 0], end_id)
+        result.next_token_ids = next_token_ids
+        return next_token_ids.clone()
```
Ratish1, line 367: This hook is not following the async-decode launch/resolve contract used by the base runner and Higgs. `post_decode_launch()` returns `next_token_ids.clone()`, which is still a device tensor; no pinned host buffer is allocated, and no non-blocking D2H is enqueued before the base records the CUDA event. Resolve then restores that device tensor, so the first actual host materialization still happens later in `SGLangOutputProcessor.process(...).tolist()`.

That makes the lookahead counter misleading for MOSS: `event.query()` only says the device clone is ready, not that the host-side collect snapshot is ready. If this PR is meant to claim host-collect overlap, please stage the required resolve data into a pinned CPU buffer at launch, as Higgs does with `_next_host_staging(...).copy_(..., non_blocking=True)`, and have resolve consume that host snapshot. If the device clone is intentionally only a correctness workaround for the stop-id aliasing bug, then this should not present the MOSS path as implementing the same async D2H overlap mechanism?

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -1003,8 +1011,14 @@ def _resolve_and_process(self, batch, sched_output, pending_step) -> None:
         _mark_sampler_finished sets) must be KEPT so process_batch_result emits
         it — only reqs finished in a *prior* step are the overrun to drop.
         """
-        pre_finished = [r.finished() for r in batch.reqs]
-        # rids finished in a PRIOR step (overrun) — suppress their stream emit
+        # A request retracted at step S (KV freed) is still in step S+1's lagged
+        # batch; without dropping it, process_batch_result would re-free its KV
+        # (the upstream enable_overlap guard is off here). Treat it like a
+        # prior-step finish: drop it from this resolve batch.
```
Ratish1: nit: can trim this comment

sglang_omni/models/moss_tts_local/model_runner.py
JiaxinD: Fixed in 8128136. The sampling_steps reset and reset_for_refill now run on every retraction re-prefill, governed by whether the request still holds a pool row, not by output_rows being non-empty. So a request retracted before it emitted any frame (empty output_rows but a held row) still realigns the launch-side counter to generation_steps and clears its stranded row; both are no-ops on a fresh prefill. Regression: test_resume_with_empty_output_rows_still_resets_sampling_steps (generation_steps=0, sampling_steps=1, output_rows=[] resets to 0). Re-ran full CPU plus S0 plus S1 plus S4 on the sglang 0.5.12.post1 stack: all green.

sglang_omni/models/moss_tts_local/model_runner.py
JiaxinD: Agreed, reworded. The device clone is a correctness snapshot for the stop-id overwrite (the base aliases the published next_token_ids onto output_ids and the next step overwrites it in place before the lagged resolve, so without it a bs=1 eos finish is lost), not a host-collect D2H overlap. After the GPU radix hash, MOSS's resolve does no heavy host work, so pinned staging has no object in this form. I reframed the PR text to launch/resolve as a correctness contract plus an overlap slot reserved for when #757 vectorization makes resolve heavy again, and fixed the body/docstrings that implied MOSS already does D2H overlap. The 100% query-hit counter now reads as proof the lookahead path executes, not a host-overlap result.

sglang_omni/cli/serve.py
JiaxinD: Done in 8128136: dropped the serve.py factory comment, removed the #745 reference from the radix comment, and trimmed the retract comment. Also applied the rule across this PR: inline code comments carry no PR or issue numbers.

sglang_omni/scheduling/omni_scheduler.py
zhaochenyang20: `_event_loop_async_decode()` first calls `_resolve_pending_async()` before the synchronous fast path, and that drain can mark or process a request as retracted for a lagged pending step. The code then checks only `any(r.finished() for r in batch.reqs)` at line 1123 before calling `batch.filter_batch()`. If the stale `batch` contains a request whose only stale state is `is_retracted=True`, it is not filtered and line 1130 sends that already-retracted request into synchronous `run_batch`. The same file's `_resolve_and_process()` already treats `is_retracted` as a prior-step overrun at line 1017, so this is not an intentional semantic difference; the fast-path analogue simply missed retractions. The failure mode is that a retracted request can participate in forward/finalize again, re-trigger KV release, or carry a frame that should have been discarded.

sglang_omni/model_runner/base.py
zhaochenyang20, line 186: the substance of @Ratish1 's comment was reworded in the PR text, but the shared code comments and naming still say the old thing. `ModelRunner.execute_launch()` stores the return value from `post_decode_launch()` in `host_buf`, and lines 160-162 still say the event is recorded after async D2H so `event.query()==True` means the host buffer is ready. MOSS `post_decode_launch()` returns `next_token_ids.clone()`, which remains on device; it does not allocate pinned CPU staging and does not enqueue a non-blocking D2H. The current code can work as a stop-id correctness snapshot, but the shared contract still tells future maintainers and benchmarks to interpret query-hit as host-collect overlap evidence; the risk is that a later default flip or performance regression analysis is decided from the wrong metric.

zhaochenyang20 (changes requested):
If we accept this PR as delivering a real MOSS-TTS Local performance win, this should be treated as a pre-merge issue. The current default is still `enable_async_decode: bool = False`, and the default `MossTTSLocalPipelineConfig` only passes `gpu_id` and `dtype` in `tts_engine.factory_args`; it does not enable async decode. The CLI `--async-decode default` also resolves back to the pipeline default. As a result, users who start the server with the simplest command/config still run the synchronous decode path and must explicitly pass `--async-decode on` to get the performance path claimed by the PR. That conflicts with the performance-PR principle: if this is the best-performing path, the default user experience should not remain on the old path. The test suite also still documents `Default stays OFF ... pending the Phase-3 flag-flip PR`, which means the implementation treats the win as a later default flip rather than this PR's default behavior. The failure mode is that default-start users keep running the old synchronous path, so benchmark claims and default production behavior diverge. Later tuning can then assume “#758 already optimized the default” while the real default performance problem remains hidden behind a flag.

sglang_omni/scheduling/omni_scheduler.py
JiaxinD: Good catch, fixed in 7ddebda. The synchronous fast path now drops finished AND retracted reqs before run_batch (it checked only finished()), aligning with the _resolve_and_process prior-step overrun semantics. The drop is extracted into _drop_stale_overrun, which uses filter_batch(keep_indices=...) so a request the drain retracted does not re-enter forward/finalize and re-free its already-freed KV. Regression test test_omni_scheduler_fast_path_drops_retracted_req. Re-ran full CPU plus S0 plus S1 plus S4 on the sglang 0.5.12.post1 stack: all green.

sglang_omni/model_runner/base.py
JiaxinD: Agreed, fixed in 7ddebda. Renamed the shared _PendingStep slot host_buf to launch_buf and rewrote the base contract to be implementation-neutral: the event recorded after post_decode_launch means the launched step is published, and launch_buf is whatever the launch returns for resolve, either a device snapshot of the published ids (MOSS-TTS-Local, no host copy) or a pinned host staging buffer an async D2H wrote into (Higgs). Removed the comments that told maintainers and benchmarks to read query-hit as host-collect overlap evidence. Higgs keeps host_buf as its local name since it genuinely stages to host.

sglang_omni/cli/serve.py
```diff
@@ -14,8 +14,11 @@
 
 _STAGE_TOGGLE_MODE = Literal["default", "on", "off"]
 _QWEN_COLOCATED_CONFIG_CLASS = "Qwen3OmniSpeechColocatedPipelineConfig"
-_HIGGS_ASYNC_DECODE_FACTORY = (
-    "sglang_omni.models.higgs_tts.stages.create_sglang_tts_engine_executor"
+_ASYNC_DECODE_FACTORIES = frozenset(
+    {
+        "sglang_omni.models.higgs_tts.stages.create_sglang_tts_engine_executor",
+        "sglang_omni.models.moss_tts_local.stages.create_sglang_tts_engine_executor",
```
Hayden727, line 20: [P2] This adds MOSS to the async-decode allowlist, but the public `--async-decode` help below still says “Currently supported by Higgs TTS” and describes “per-step host collect overlaps the next GPU forward”. That is now misleading for MOSS: its `post_decode_launch()` returns a device `next_token_ids.clone()` and the PR now frames query-hit as path-execution/correctness evidence rather than host-collect overlap. Please update the CLI help, and the same stale host-buffer/async-D2H wording in `ModelRunner.execute_launch()`, `_run_batch_launch()`, and `_event_loop_async_decode()`, to the neutral launch-buffer contract so users/benchmarks do not interpret MOSS as proving the old host-overlap path.

sglang_omni/model_runner/base.py
zhaochenyang20, line 148: 7ddebda only neutralized `_PendingStep` and part of the hook comments, while the `execute_launch()` docstring still says “snapshot into a pinned host buffer” and “right after that async D2H.” `omni_scheduler.py` has the same residue: `_run_batch_launch()` says “async D2H of the collect snapshot,” and `_event_loop_async_decode()` says it “RESOLVES the previous step's host-side collect” and that “CPU work overlaps.” Those statements can be true for Higgs pinned host staging, but they are not true for the current MOSS implementation: MOSS returns a device clone used only as the stop-id correctness snapshot, with no host copy and no host-collect overlap proven by query-hit.

Rewrite the `execute_launch()`, `_run_batch_launch()`, and `_event_loop_async_decode()` docstrings as “launch publishes a model-specific resolve payload and records an event after publication.” Also state explicitly that `launch_buf` can be either a device-side correctness snapshot or pinned host staging; only the pinned-host-staging variant provides host-D2H overlap. This keeps the shared code consistent with the new `_PendingStep` comment, MOSS `post_decode_launch()`, and the PR body.

zhaochenyang20 (changes requested):

sglang_omni/model_runner/base.py
JiaxinD: Done in a9a63bd. Finished the sweep across every remaining spot, not just _PendingStep: the execute_launch docstring, _run_batch_launch, _event_loop_async_decode, the base post_decode_launch / post_decode_resolve defaults, _next_host_staging, the query-hit counter comment, and the serve.py help. The shared contract now states launch_buf is a device-side correctness snapshot (MOSS, no host copy) or pinned host staging (Higgs), and only the latter provides host-D2H overlap; the query-hit counter comment now says it measures path execution (the launched step's GPU work published in time), not host overlap. Backstop grep is zero across all six phrases (pinned host buffer / async D2H / host collect overlap / host buffer ready / CPU work overlaps / supported by Higgs). I also reworded the body Performance section so it reads as mechanism validation, not a gains table.

sglang_omni/cli/serve.py
JiaxinD: Done in a9a63bd. The --async-decode help no longer says "Currently supported by Higgs TTS"; it now reads "Available for Higgs TTS and MOSS-TTS-Local" and frames the overlap as a host-staging (Higgs) effect rather than a generic one. While here I swept the same neutralization across the shared launch/resolve docstrings. Backstop grep is zero across the six phrases (pinned host buffer / async D2H / host collect overlap / host buffer ready / CPU work overlaps / supported by Higgs).

Hayden727 (approved):

gaoyang07 (approved):
LGTM

zhaochenyang20 (approved):

PR #759, opened 2026-06-12
[Perf] Pool MOSS-TTS Local frame launch state

### LINKED ISSUE #757: [Perf] MOSS-TTS Local v1.5: vectorize per-frame launch orchestration into [P] pool tensors

Hayden727:
## Summary

Closes #757.

- add pool-resident `[P]` `generation_steps` for MOSS-TTS Local decode launch state
- prepare active pool rows once and reuse `row_t` so `_collect_frame` gathers launch inputs from pool tensors
- mirror committed `generation_steps` into the pool through a batch model-runner hook while preserving retraction/re-prefill position
- move MOSS Local audio repetition penalty state into the pool: `[P]` penalty tensor plus `[P, n_vq, audio_vocab]` generated-token presence mask
- keep the normal no-penalty frame path graphable and mask-free; penalty-enabled rows use the eager frame path with pool-resident history
- extend MOSS local state-pool tests, base runner hook tests, and the test catalog entry

## Why

The MOSS-TTS Local frame path already stores feedback embeddings and request-static sampling params in row-indexed pool tensors. `_collect_frame` still rebuilt active rows, `generation_steps`, and repetition-penalty history from Python request objects each frame. This moves the remaining next-step launch state into the pool so the launch path can gather with `row_t`, matching the #757 follow-up from the #745 review.

For repetition penalty specifically, `audio_repetition_penalty` is now request-static pool state and generated audio history is a pool-resident presence mask. The default no-penalty path does not build request-local `datas` and does not update the history mask. When a request enables audio repetition penalty, the runner intentionally bypasses frame-decode CUDA graph replay for that batch, applies the penalty from pool tensors, updates only active penalty rows after emitted frames, and rebuilds the mask from `output_rows` only on the retraction/re-prefill slow path.

## Performance / CUDA Graph Note

This does **not** claim a large end-to-end TTS latency win. The measured impact is local to the graph replay launch-prep path:

- CUDA graph behavior was tested on H100: both `origin/main` and this PR capture `bs=[1,2,4,8,16]`; the latest PR run logs both SGLang backbone graph capture and `MOSS-TTS Local frame-decode CUDA graphs captured for bs=[1, 2, 4, 8, 16]`.
- The no-penalty path remains the CUDA-graph path: it does not materialize request-local `datas`, does not walk `data.output_rows`, and does not update the repetition-history mask.
- The penalty-enabled path remains eager because frame-decode graphs do not include repetition-penalty logit mutation; the eager path now reads pool tensors/masks instead of reconstructing per-request history inside `_collect_frame`.
- Synthetic launch-only prep benchmark (`_collect_frame` active rows + params + `generation_steps`, no model/vocoder/HTTP): new path reduces representative per-step prep from `65.53us -> 43.74us` at bs=16, `69.75us -> 43.73us` at bs=32, and `79.77us -> 44.17us` at bs=64.
- Synthetic total prep benchmark including the generation-step pool commit after switching to batch commit: bs=1 is slightly slower (`84.87us -> 86.06us`), bs=4 is roughly flat (`87.46us -> 87.18us`), and larger batches improve modestly: bs=8 `91.31us -> 89.70us`, bs=16 `93.60us -> 91.52us`, bs=32 `101.41us -> 96.70us`.
- E2E single-request smoke did not show a meaningful request-level win: `x-engine-time` mean was effectively flat (`0.276331s` main vs `0.276070s` PR), while HTTP wall time was noisier and slightly slower (`0.369102s` main vs `0.374820s` PR). Treat this as a hot-path cleanup / graph-readiness change, not an E2E throughput claim.

## Validation

Local:
- `python -m compileall -q sglang_omni/models/moss_tts_local/model_runner.py sglang_omni/models/moss_tts_local/state_pool.py tests/unit_test/moss_tts_local/test_state_pool.py tests/unit_test/moss_tts_local/test_pipeline.py`
- `git diff --check`
- pre-commit hooks during commit: passed

Remote H100 unit tests:
- env: `FLASHINFER_DISABLE_VERSION_CHECK=1` to bypass the container existing flashinfer/flashinfer-jit-cache version mismatch
- `pytest -q tests/unit_test/moss_tts_local/test_state_pool.py tests/unit_test/moss_tts_local/test_pipeline.py -q`: completed 100%, 2 warnings
- `pytest -q tests/unit_test/moss_tts_local/test_state_pool.py::test_cached_pool_rows_drive_collect_and_batched_step_commit -q`: passed, 2 warnings
- `pytest -q tests/unit_test/moss_tts_local -q`: completed 100%, 2 warnings

Remote H100 E2E:
- host/container: `novita-h100-omni` / `sglang-omni-hayden`
- server snapshot: runtime code at `a47c6abf`; latest `0e3aaf8d` is test-only
- config: `examples/configs/moss_tts_local.yaml`, model `OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5`
- launch: `CUDA_VISIBLE_DEVICES=6,7 FLASHINFER_DISABLE_VERSION_CHECK=1 PYTHONPATH=. python -m sglang_omni.cli serve --config examples/configs/moss_tts_local.yaml --host 127.0.0.1 --port 28257 --log-level info`
- readiness: `GET /health` -> `200`; server selected an alternate free port when `28257` was still occupied by the previous validation process
- no-penalty request: `POST /v1/audio/speech` with `${token:8}`, `max_new_tokens=12`, `seed=11` -> `HTTP 200`, `content-type: audio/wav`, `184364` bytes, `x-engine-time: 0.595348`
- penalty request: same payload plus `repetition_penalty=1.2` -> `HTTP 200`, `content-type: audio/wav`, `184364` bytes, `x-engine-time: 0.514209`
- cleanup: temporary server processes and `/tmp/sglang_omni_issue757_pf` snapshot removed; remaining GPU process on host GPU 7 belonged to a different container, not this validation

### DIFF
--- a/sglang_omni/models/moss_tts_local/model_runner.py
+++ b/sglang_omni/models/moss_tts_local/model_runner.py
@@ -95,6 +95,7 @@ class MossTTSLocalModelRunner(ModelRunner):
         forward_batch: Any,
         requests: list,
     ) -> torch.Tensor:
+        self._stage_repetition_penalty_state(forward_batch, requests)
         pieces = []
         for sched_req in requests:
             data = sched_req.data
@@ -162,6 +163,7 @@ class MossTTSLocalModelRunner(ModelRunner):
                 "MOSS-TTS Local decode batch exceeds the staged decode-embedding "
                 f"rows ({batch_size} > {pool.padding_row})"
             )
+        self._stage_repetition_penalty_state(forward_batch, requests)
         row_tensor, pool_rows = pool.prepare_active_rows(requests)
         with torch.no_grad():
             weight[:batch_size].copy_(pool.feedback_embeds[row_tensor])
@@ -184,14 +186,14 @@ class MossTTSLocalModelRunner(ModelRunner):
     ) -> None:
         if not requests:
             return
-        rows, end_id = self._run_frame_decode(result, requests)
+        rows, end_id = self._run_frame_decode(result, forward_batch, requests)
         # Radix key is a capture-safe GPU hash: a device op, no host sync.
         next_text = rows[:, 0]
         next_token_ids = self._row_radix_token_ids(rows, next_text, end_id)
         result.next_token_ids = next_token_ids
         schedule_batch.output_ids = next_token_ids
 
-    def _run_frame_decode(self, result: Any, requests: list):
+    def _run_frame_decode(self, result: Any, forward_batch: Any, requests: list):
         """GPU half shared by sync ``_collect_frame`` and async
         ``post_decode_launch``. Returns ``(rows, end_id)`` and does NOT publish
         ``next_token_ids``; the caller does, because the async path keeps a
@@ -208,8 +210,7 @@ class MossTTSLocalModelRunner(ModelRunner):
         cfg = self.model.config
         device = hidden_states.device
         pool = self.model._state_pool
-        datas = [sched_req.data for sched_req in requests]
-        batch_size = len(datas)
+        batch_size = len(requests)
         num_channels = int(cfg.n_vq) + 1
 
         row_t = getattr(forward_batch, "moss_pool_row_t", None)
@@ -241,8 +242,13 @@ class MossTTSLocalModelRunner(ModelRunner):
             if not self._is_chunked_request(sched_req)
         }
         gen_steps = pool.generation_steps[row_t].to(device=device)
-        rep_penalties = [float(d.audio_repetition_penalty) for d in datas]
-        rep_histories = self._gather_rep_histories(datas, rep_penalties, device)
+        rep_penalties = getattr(forward_batch, "moss_rep_penalties", None)
+        rep_datas = getattr(forward_batch, "moss_rep_datas", None)
+        rep_histories = (
+            None
+            if rep_datas is None or rep_penalties is None
+            else self._gather_rep_histories(rep_datas, rep_penalties, device)
+        )
 
         def sample_text(logits: torch.Tensor) -> torch.Tensor:
             return MossTTSModelRunner._sample_tokens(
@@ -341,10 +347,9 @@ class MossTTSLocalModelRunner(ModelRunner):
         the stop id and silently dropping a bs=1 eos finish (4096-frame runaway).
         The clone preserves it; resolve swaps it back.
         """
-        del forward_batch
         if not requests:
             return None
-        rows, end_id = self._run_frame_decode(result, requests)
+        rows, end_id = self._run_frame_decode(result, forward_batch, requests)
         next_token_ids = self._row_radix_token_ids(rows, rows[:, 0], end_id)
         result.next_token_ids = next_token_ids
         return next_token_ids.clone()
@@ -408,6 +413,27 @@ class MossTTSLocalModelRunner(ModelRunner):
         )
         data.sampling_steps = s + 1
         return s
+
+    @staticmethod
+    def _stage_repetition_penalty_state(forward_batch: Any, requests: list) -> None:
+        """Attach slow-path repetition state only when a penalty is active."""
+        if not requests:
+            forward_batch.moss_rep_datas = None
+            forward_batch.moss_rep_penalties = None
+            return
+        active = any(
+            float(getattr(sched_req.data, "audio_repetition_penalty", 1.0)) != 1.0
+            for sched_req in requests
+        )
+        if not active:
+            forward_batch.moss_rep_datas = None
+            forward_batch.moss_rep_penalties = None
+            return
+        forward_batch.moss_rep_datas = [sched_req.data for sched_req in requests]
+        forward_batch.moss_rep_penalties = [
+            float(getattr(sched_req.data, "audio_repetition_penalty", 1.0))
+            for sched_req in requests
+        ]
 
     @staticmethod
     def _gather_rep_histories(
--- a/sglang_omni/models/moss_tts_local/model_runner.py
+++ b/sglang_omni/models/moss_tts_local/model_runner.py
@@ -481,6 +481,33 @@ class MossTTSLocalModelRunner(ModelRunner):
         if pool is not None:
             pool.commit_generation_step(sched_req.request_id, generation_steps)
 
+    def on_generation_steps_advanced(
+        self, advanced_steps: list[tuple[Any, int]], forward_batch: Any
+    ) -> None:
+        pool = getattr(self.model, "_state_pool", None)
+        if pool is None or not advanced_steps:
+            return
+        steps = [int(generation_steps) for _, generation_steps in advanced_steps]
+        row_t = getattr(forward_batch, "moss_pool_row_t", None)
+        if row_t is not None and int(row_t.numel()) == len(steps):
+            step_t = torch.tensor(steps, dtype=torch.long, device=row_t.device)
+            pool.commit_generation_steps(row_t, step_t)
+            return
+        rows = []
+        row_steps = []
+        for sched_req, generation_steps in advanced_steps:
+            row = pool.row_for(sched_req.request_id)
+            if row is None:
+                continue
+            rows.append(row)
+            row_steps.append(int(generation_steps))
+        if not rows:
+            return
+        device = pool.generation_steps.device
+        row_t = torch.tensor(rows, dtype=torch.long, device=device)
+        step_t = torch.tensor(row_steps, dtype=torch.long, device=device)
+        pool.commit_generation_steps(row_t, step_t)
+
     def post_process_outputs(
         self,
         result: Any,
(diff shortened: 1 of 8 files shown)


sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -170,19 +172,14 @@ def _collect_frame(
...
         device = hidden_states.device
         pool = self.model._state_pool
-        pool_rows = []
-        for sched_req in requests:
-            rid = sched_req.request_id
-            row = self.model.acquire_row(rid)
-            pool_rows.append(row)
-            pool.ensure_params(row, rid, sched_req.data)
         datas = [sched_req.data for sched_req in requests]
         batch_size = len(datas)
```
AkazaAkane: datas is a Python list of request-local Python objects. Building it inside _collect_frame() means the frame path still walks scheduler request objects each step. For the normal no-penalty path, we can avoid creating datas at all. use:
`batch_size = len(requests)`
It only build datas in a slow fallback path if repetition penalty is active.

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -199,9 +196,7 @@ def _collect_frame(
         audio_top_p = params["audio_top_p"]
         audio_top_k = params["audio_top_k"]
         sampling_seeds = params["seeds"]
-        gen_steps = torch.tensor(
-            [int(d.generation_steps) for d in datas], dtype=torch.long, device=device
-        )
+        gen_steps = pool.generation_steps[row_t].to(device=device)
         rep_penalties = [float(d.audio_repetition_penalty) for d in datas]
```
AkazaAkane: **audio_repetition_penalty** is still read from Python request data, even though it is also a request-static sampling parameter. That makes it inconsistent with the rest of the state-pool direction. Longer term, `audio_repetition_penalty` should be a `[P]` pool tensor, and the penalty-enabled path should move history/application out of `data.output_rows` into pool-resident history or token-presence masks.

AkazaAkane (commented):
I would not block this PR on full repetition-penalty tensorization, but I would avoid claiming all per-frame Python work is gone. This PR clears active-row / generation-step launch state; repetition penalty should be tracked as the next follow-up.

tests/unit_test/moss_tts_local/test_state_pool.py
```diff
@@ -327,6 +373,103 @@ def prepare_multi_modal_inputs(rows):
...
+    data = SimpleNamespace(
+        req=SimpleNamespace(is_chunked=0),
+        generation_steps=0,
+        extra_model_outputs={},
+        output_rows=[],
+    )
+    sched_req = SimpleNamespace(request_id="rid", data=data)
+    row = pool.acquire_row("rid")
+
+    runner._finalize(
```
JiaxinD, line 802: This test (and `test_collect_frame_reads_generation_steps_from_pool`) drives a single request with a `forward_batch` that has no `moss_pool_row_t`, so it only hits the per-rid `row_for` fallback. The vectorized fast path in `on_generation_steps_advanced` and the `_collect_frame` reuse of the cached `moss_pool_row_t` (the optimization this PR adds) aren't covered anywhere; `moss_pool_row_t` never appears in the tests. A regression that left the cached row tensor stale or misaligned with `requests` would still pass green. Worth one test that calls `_write_decode_input_embedding`, then feeds that same `forward_batch` into both `_collect_frame` and `_finalize` with two requests in non-identity row order.

.github/actions/omni-setup/action.yaml
```diff
@@ -28,7 +28,10 @@ runs:
         echo "HF_HUB_ENABLE_HF_TRANSFER=0" >> "${GITHUB_ENV}"
         echo "HF_ENDPOINT=${HF_ENDPOINT:-https://hf-mirror.com}" >> "${GITHUB_ENV}"
         # PyPI mirror + shared wheel cache (HF mirror does not accelerate pip/uv).
+        # Fall back to the canonical PyPI index for newly published CUDA wheels
+        # that may not have reached the mirror yet.
         echo "UV_INDEX_URL=${UV_INDEX_URL:-https://pypi.tuna.tsinghua.edu.cn/simple}" >> "${GITHUB_ENV}"
+        echo "UV_EXTRA_INDEX_URL=${UV_EXTRA_INDEX_URL:-https://pypi.org/simple}" >> "${GITHUB_ENV}"
```
JiaxinD: I think this CI change does not behave as the comment describes.

With uv, UV_INDEX_URL is treated as the default index, which has the lowest priority. UV_EXTRA_INDEX_URL is treated as an additional index, and additional indexes are searched before the default index. So setting UV_EXTRA_INDEX_URL to PyPI does not make PyPI a fallback after the TUNA mirror. It makes PyPI higher priority than the mirror for resolution.

That means this will mostly bypass the mirror instead of only using PyPI when a newly published CUDA wheel has not reached the mirror yet.

Could we either adjust the configuration to match the intended fallback behavior, or update the comment to reflect that PyPI is intentionally the higher priority index?

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -156,7 +162,6 @@ def _collect_frame(
         schedule_batch: Any,
         requests: list,
     ) -> None:
-        del forward_batch
         if not requests:
             return
         hidden_states = getattr(result.logits_output, "hidden_states", None)
```
zhaochenyang20: do not use `getattr` in any PRs.

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -170,19 +175,23 @@ def _collect_frame(
...
-        batch_size = len(datas)
+        batch_size = len(requests)
         num_channels = int(cfg.n_vq) + 1
 
-        row_t = torch.tensor(
-            pool_rows, dtype=torch.long, device=pool.feedback_embeds.device
-        )
+        row_t = getattr(forward_batch, "moss_pool_row_t", None)
+        pool_rows = getattr(forward_batch, "moss_pool_rows", None)
+        if row_t is None or pool_rows is None:
```
zhaochenyang20: do not use `getattr` in any PRs.

sglang_omni/models/moss_tts_local/state_pool.py
```diff
@@ -36,6 +38,9 @@ def __init__(self, model: Any) -> None:
         self.hidden_size = int(weight.shape[1])
         self.device = weight.device
         self.dtype = weight.dtype
+        config = getattr(model, "config", None)
+        self.n_vq = int(getattr(config, "n_vq", 12) or 12)
+        self.audio_vocab_size = int(getattr(config, "audio_vocab_size", 1024) or 1024)
```
zhaochenyang20: do not use `getattr` in any PRs.

tests/unit_test/moss_tts_local/test_state_pool.py
```diff
@@ -327,6 +441,281 @@ def prepare_multi_modal_inputs(rows):
...
+    def decode_frame(hidden_states, *, sample_text, sample_audio):
+        del hidden_states, sample_text, sample_audio
+        called["eager"] = True
+        return (
+            torch.zeros(1, dtype=torch.long),
+            torch.full((1, 12), 7, dtype=torch.long),
+        )
+
+    model.decode_frame_graphed = decode_frame_graphed
+    model.decode_frame = decode_frame
```
zhaochenyang20, line 626: The current `test_collect_frame_uses_eager_path_when_audio_repetition_penalty_active` fake `decode_frame()` discards both `sample_text` and `sample_audio` and returns fixed codes. That only proves that a penalty batch selects the eager path and that an emitted frame updates history; it does not prove that `_collect_frame()`'s `sample_audio()` reads the pool presence mask, applies the penalty for the channel, and then calls the sampler. If someone later removes `_apply_audio_repetition_penalty_mask()` or passes the wrong channel, this test would still pass.

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

PR #762, opened 2026-06-12
[Perf] MOSS-TTS Local v1.5: native pool-resident CUDA graph for the frame-decode sampling path

### LINKED ISSUE #736: [Perf] MOSS-TTS Local v1.5: native pool-resident CUDA graph for the frame-decode sampling path

nagisa-kunhah:
## Motivation

Make the MOSS-TTS Local decode path CUDA-graph-friendly by moving decode
inputs and sampled outputs into the row-indexed decode-state pool.

The previous implementation had two decode modes behind a forward-sampling
bool. The old fallback path copied the next-step feedback embedding into
`_decode_input_embedding.weight`, while the native path used a separate set of
private `_cg_*` staging buffers for sampling params, base positions, sampled
rows, and feedback embeddings. That left the decode state split across several
places and made CUDA graph capture harder to reason about.

This PR removes that split. Decode now has one state model:

- `before_decode` stages only stable pool row ids in `_cg_pool_rows`;
- model forward reads decode input embeddings from `pool.feedback_embeds`;
- model forward runs the local frame sampling step during decode;
- model forward writes sampled rows and candidate feedback back into pool
  outputs;
- runner collection snapshots those pool outputs and keeps the existing
  request journal / chunked-prefill behavior.

close #736

## What Changed

- Added fixed active-batch decode staging buffers on
  `MossTTSLocalSGLangModel` for CUDA-graph replay: pool-row ids, feedback
  embeddings, sampling params/seeds/steps, next feedback/steps, sampled rows,
  and graph-computed next token ids.
- Kept the decode-state pool focused on durable per-request state: feedback
  embedding, sampling params/seed, generation/sampling steps, and audio
  repetition-penalty history. Per-step sampled rows and token ids now live in
  model-owned active-slot buffers instead of pool-owned output buffers.
- Updated `before_decode` to stage pool state into the model active buffers and
  write decode `input_ids` as active slot ids. It also pads staging up to the
  CUDA graph bucket size when graph padding is enabled, routing extra slots to
  the reserved padding row.
- Kept the reserved padding row out of normal allocation and gave it safe
  sampling defaults (`temperature=1.0`, `top_p=1.0`, text top-k `50`, audio
  top-k `25`, seed `0`, repetition penalty `1.0`) so padded graph slots can be
  sampled without touching real request state.
- Moved the normal decode sampling path into
  `MossTTSLocalSGLangModel.forward()`: decode forward reads staged feedback,
  runs `_decode_frame_graphable`, writes sampled multi-channel rows, computes
  GPU radix token ids, and stages the next feedback embedding / sampling step.
- Updated decode collection and async launch to consume the forward-sampled
  active buffers, publish `next_token_ids`, journal emitted rows, and commit
  next feedback / sampling steps back into the request's pool row.
- Preserved the legacy hidden-state collection path for prefill and for decode
  requests with non-default `audio_repetition_penalty`; lookahead eligibility
  continues to reject batches that exceed the frame graph limit or require that
  repetition-penalty fallback.
- Stopped mutating `_decode_input_embedding.weight` during decode preparation.
  The table remains only a sizing/device anchor for the pool and active decode
  buffers.
- Expanded CPU unit coverage for padding defaults, bucket padding, active
  buffer staging, native forward-sample collection, repetition-penalty fallback,
  async decode launch, and the no-pool-write contract inside decode forward.

## Design Rationale

The key idea is to make the pool row the single durable identity for a running
request's decode state. `_cg_pool_rows` is intentionally small and
graph-friendly: it only maps the current batch slots to stable pool rows. The
larger state tensors live in process-lifetime pool buffers, so graph replay can
read/write fixed addresses while requests come and go.

This also makes decode retraction/resume easier to reason about. Feedback for
the next decode step, the sampling base position, and the sampled row produced
by the current step all live under the same request-owned row. Collection then
copies the step output into the journal before the next replay can overwrite
the pool output buffers.

The padding row remains reserved and is never acquired by a real request. CUDA
graph bucket padding can route non-real rows there without mutating real
request state.

## Correctness

Focused unit tests:

```bash
PYTHONPATH=$PWD .venv/bin/python -m pytest \
  tests/unit_test/moss_tts_local/test_state_pool.py \
  tests/unit_test/moss_tts_local/test_pipeline.py -q
```

Result:

```text
72 passed, 2 warnings
```

The tests cover pool buffer shape/defaults, row-id staging, decode embedding
immutability, missing internal state errors, pool-output collection,
chunked-prefill feedback suppression, and decode collection with repetition
penalty enabled.

## Main Comparison Benchmark

Benchmark source: `moss_tts_native_forward_sampling_benchmark.md`.

Setup:

- Model: `/root/sglang-omni/model_cache/moss_tts_local_v15`
- Endpoint: `/v1/audio/speech`
- Request args: `token_count=96`, `max_new_tokens=192`, `seed=1234`,
  `response_format=wav`
- GPU observed in service logs: NVIDIA H800 PCIe

Baseline note: plain `main` did not start in this environment because the
cached remote processor passed `codec_weight_dtype` into codec
`AutoModel.from_pretrained`, while the cached codec class rejected that
argument. To obtain a runnable main baseline without modifying the main
worktree source, the main service was launched with a startup-only
`sitecustomize.py` shim that drops `codec_weight_dtype` from
`AutoModel.from_pretrained`. The runner/model code under test still came from
main.

Delta columns are current branch relative to the main baseline. Negative wall
clock / latency is better; positive throughput is better.

| Case | Main ok | Current ok | Main audio total (s) | Current audio total (s) | Wall delta | Audio throughput delta | Mean latency delta | P95 latency delta |
| c4/n8 | 8/0 | 8/0 | 30.72 | 30.72 | -0.86% | +0.87% | -0.57% | -1.86% |
| c8/n32 | 32/0 | 32/0 | 122.88 | 122.88 | -1.19% | +1.20% | -1.12% | -2.84% |
| c16/n64 | 64/0 | 64/0 | 245.80 | 245.76 | -2.37% | +2.41% | -1.77% | -7.12% |

Observations:

- All compared cases completed successfully on both main baseline and current
  branch.
- Current branch is faster than main in all three cases. The clearest
  throughput gain is c16/n64: `14.9518x` audio throughput vs `14.5996x` on
  main (`+2.41%`).
- Tail latency is lower in all three cases: p95 improves by `1.86%`, `2.84%`,
  and `7.12%` for c4, c8, and c16 respectively.
- These are single-run results per case. Treat differences below about
  `1-2%` as noise unless repeated runs confirm the trend.

## Checklist

- [x] Add unit tests.

## CI

the label remains. Draft PRs are skipped even if labeled.

### DIFF
--- a/sglang_omni/models/moss_tts_local/model_runner.py
+++ b/sglang_omni/models/moss_tts_local/model_runner.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import bisect
 from typing import Any
 
 import torch
@@ -19,11 +20,9 @@ class MossTTSLocalModelRunner(ModelRunner):
     """Drives the per-frame local-transformer decode and feedback embeddings.
 
     Per step: the backbone (radix-cached, CUDA-graphed) produces one hidden
-    state per request; :meth:`_collect_frame` then runs the batched local
-    micro-decode — a binary continue/stop decision and 12 sequentially
-    sampled RVQ codes — and stages the next frame's summed embedding through
-    ``model._decode_input_embedding`` so the next decode step stays
-    CUDA-graph-replayable (decode input_ids are row indices).
+    state per request. Prefill collection runs the local micro-decode in the
+    runner; decode collection reads the frame sampled inside model forward and
+    staged in the row-indexed decode-state pool.
     """
 
     _outbox: Any | None = None
@@ -59,7 +58,7 @@ class MossTTSLocalModelRunner(ModelRunner):
     ) -> None:
         del is_lookahead
         del schedule_batch
-        self._write_decode_input_embedding(forward_batch, requests)
+        self._prepare_forward_sample_inputs(forward_batch, requests)
 
     def post_prefill(
         self,
@@ -74,7 +73,7 @@ class MossTTSLocalModelRunner(ModelRunner):
             is_prefill_only = False
         if bool(is_prefill_only):
             return
-        self._collect_frame(result, forward_batch, schedule_batch, requests)
+        self._collect_frame_legacy(result, forward_batch, schedule_batch, requests)
 
     def post_decode(
         self,
@@ -83,7 +82,10 @@ class MossTTSLocalModelRunner(ModelRunner):
         schedule_batch: Any,
         requests: list,
     ) -> None:
-        self._collect_frame(result, forward_batch, schedule_batch, requests)
+        if bool(getattr(forward_batch, "moss_has_audio_repetition_penalty", False)):
+            self._collect_frame_legacy(result, forward_batch, schedule_batch, requests)
+            return
+        self._collect_frame_from_forward_sample(result, schedule_batch, requests)
 
     def lookahead_eligible(self, batch: Any) -> bool:
         """Route to sync when the batch cannot take the graphed frame-decode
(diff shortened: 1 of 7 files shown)


sglang_omni/models/moss_tts_local/stages.py
zhaochenyang20, line 1: line 480 exposes the `forward_sample_in_forward` factory argument, and enabling it routes requests through a new pool-resident sampling data path without any real GPU/model parity run or performance benchmark in the PR. This parameter is not ordinary config plumbing; it changes how decode steps read feedback embeddings, compute sampling base positions, and hand sampled rows back to the scheduler. CPU unit tests can prove buffer shapes and fake-runner behavior, but they cannot prove equivalence for the real bf16 CUDA graph, real `MossTTSLocalTransformer`, real stop head, and seeded sampling when the opt-in is on versus off. It is fine to keep iterating as Draft; it is not yet a mergeable optimization.

zhaochenyang20 (changes requested):

sglang_omni/models/moss_tts_local/stages.py
nagisa-kunhah: ok, remove it

sglang_omni/models/moss_tts_local/sglang_model.py
```diff
@@ -276,6 +345,39 @@ def _prepare_multi_modal_inputs(self, input_ids: torch.LongTensor) -> torch.Tens
...
         return embeds
 
+    def _decode_pool_state(
+        self,
+    ) -> tuple[torch.Tensor, MossTTSLocalDecodeStatePool]:
+        if not hasattr(self, "_cg_pool_rows"):
+            raise RuntimeError("MOSS-TTS Local decode pool rows are not initialized")
+        if not hasattr(self, "_state_pool"):
+            raise RuntimeError("MOSS-TTS Local decode state pool is not initialized")
+        return self._cg_pool_rows, self._state_pool
```
Ccyest: delete dead code

sglang_omni/models/moss_tts_local/sglang_model.py
```diff
@@ -315,10 +419,46 @@ def forward(
             hidden_states,
             forward_batch,
         )
-        # The local-transformer frame decode (binary stop head + 12 sequential
-        # codebook samples) runs in the model runner after the graph-captured
-        # backbone returns; emitting hidden states with dummy logits keeps the
-        # backbone CUDA-graph replay free of model-specific outputs.
+        if is_decode:
+            batch_size = int(sample_hidden_states.shape[0])
+            self._check_active_decode_buffers()
```
Ccyest: forward() in sglang_model.py calls _check_active_decode_buffers() twice during decode mode. delete one

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -174,42 +170,241 @@ def _write_decode_input_embedding(
...
+        self._forward_sample_native_decode = not has_audio_repetition_penalty
 
     def _collect_frame(
         self,
         result: Any,
         forward_batch: Any,
         schedule_batch: Any,
         requests: list,
+    ) -> None:
+        self._collect_frame_legacy(result, forward_batch, schedule_batch, requests)
```
Ccyest: looks pretty useless, should delete it and update caller

sglang_omni/models/moss_tts_local/model_runner.py
Ccyest: this method is the only collection entry point for the prefill path, and the fallback for decode when audio_repetition_penalty != 1.0. It is not "legacy" — it is a core path that remains in active production use. The name legacy implies it will be deprecated, but as long as the repetition penalty feature exists, this method cannot be removed.

Rename _collect_frame_legacy to _collect_frame_eager — frame decode runs outside the CUDA graph, forming a clear contrast with _collect_frame_from_forward_sample (frame decode runs inside forward()'s graph).

sglang_omni/models/moss_tts_local/sglang_model.py
```diff
@@ -135,6 +137,73 @@ def __init__(
...
 
+    def init_forward_sample_buffers(self) -> None:
+        """Allocate fixed active-batch decode buffers for captured decode."""
+        weight = self._decode_input_embedding.weight
+        max_running_requests = int(weight.shape[0])
+        device = weight.device
+        dtype = weight.dtype
+        # The pool has max_running_requests + 1 rows; the last row is reserved
+        # for CUDA graph padding. The pool may not be constructed yet, but its
+        # padding row is therefore exactly max_running_requests.
```
Ccyest: feel like this explanation is not very informative, it will be better if we can briefly describe what is allocated here.

Also, when adding notes, remember to put your name in. For example: Notes (Yichi): balabalabala

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -174,42 +170,241 @@ def _write_decode_input_embedding(
...
+
+    def _collect_frame_from_forward_sample(
+        self,
+        result: Any,
+        schedule_batch: Any,
+        requests: list,
+    ) -> None:
+        # Sampling already ran inside MossTTSLocalSGLangModel.forward() via the
+        # native decode-forward path. Collection only snapshots the fixed
+        # active-slot step buffers and publishes the graph-computed token ids.
```
Ccyest: same here, please add your name in the note

sglang_omni/models/moss_tts_local/sglang_model.py
```diff
@@ -315,10 +419,46 @@ def forward(
...
+                dtype=self._cg_active_next_feedback_embeds.dtype
+            )
+            self._cg_active_next_sampling_steps[:batch_size] = (
+                active_sampling_steps + 1
+            )
+
+        # Prefill flow consumes hidden_states in the model runner and samples
+        # there. Decode flow above has already written the sampled frame to
+        # fixed staging buffers. Dummy logits keep SGLang's output contract
+        # satisfied.
```
Ccyest: also for comment here

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -174,42 +170,241 @@ def _write_decode_input_embedding(
         forward_batch: Any,
         requests: list,
     ) -> None:
-        batch_size = len(requests)
-        if batch_size == 0:
+        self._prepare_forward_sample_inputs(forward_batch, requests)
+
+    def _decode_staging_batch_size(self, raw_batch_size: int) -> int:
```
yxs: `_decode_staging_batch_size` gates the staging-to-bucket expansion on `_moss_local_decode_graph_padding` / `_moss_local_decode_cuda_graph_bs`. Grepping the whole repo (`grep -rn _moss_local_decode_graph_padding --include=*.py`): **these two attributes are assigned ONLY in `tests/unit_test/moss_tts_local/test_state_pool.py:472-473,600-601`; no production code ever sets them.** `_prepare_forward_sample_inputs` reads them via `getattr(..., False)`, so in production `_decode_staging_batch_size` always returns `raw_batch_size` and the `staging_batch_size > n_real` branch (extending `staged_pool_rows` with `pool.padding_row`) is dead.

Ccyest: +1 on this one

sglang_omni/models/moss_tts_local/model_runner.py
```diff
@@ -174,42 +170,241 @@ def _write_decode_input_embedding(
...
+                    device=model._cg_active_sampling_steps.device,
+                    dtype=model._cg_active_sampling_steps.dtype,
+                )
+            )
+            model._cg_active_audio_repetition_penalty[:staging_batch_size].copy_(
+                pool.audio_repetition_penalty[active_rows_t].to(
+                    device=model._cg_active_audio_repetition_penalty.device,
+                    dtype=model._cg_active_audio_repetition_penalty.dtype,
+                )
+            )
```
Ccyest: Open to discuss: I know we use the explicit pattern in higgs, but higgs only has 4 field while moss has 10.
This looks a bit redundant to me.
Fix suggestion:
Drive it with a table of (model-attr, pool-attr) name pairs in a loop;

```
_STAGE_FIELDS = (
    ("_cg_active_text_temp", "text_temp"),
    ("_cg_active_text_top_p", "text_top_p"),
    ("_cg_active_text_top_k", "text_top_k"),
    # ... audio_temp / audio_top_p / audio_top_k / seeds / sampling_steps / audio_repetition_penalty
)
for dst_name, src_name in self._STAGE_FIELDS:
    dst = getattr(model, dst_name)
    dst[:staging_batch_size].copy_(getattr(pool, src_name)[active_rows_t].to(dtype=dst.dtype))
```

sglang_omni/models/moss_tts_local/state_pool.py
```diff
@@ -134,18 +134,37 @@ def release_row(self, rid: str) -> None:
...
         self.seeds[row_idx] = 0
+
+    def _set_padding_defaults(self) -> None:
+        """Sampling defaults used when CUDA graph buckets include padding."""
+        row_idx = self.padding_row
+        self.text_temp[row_idx] = 1.0
+        self.text_top_p[row_idx] = 1.0
+        self.text_top_k[row_idx] = 50
+        self.audio_temp[row_idx] = 1.0
+        self.audio_top_p[row_idx] = 1.0
```
Ccyest: The defaults `text_top_k=50`, `audio_top_k=25`, temp/top_p=1.0 are hardcoded in three places: `init_forward_sample_buffers` (model active buffer init), `state_pool._set_padding_defaults` (padding row), and the `static_inputs` of `init_frame_decode_graphs` (graph capture inputs). All three must agree for "padding-row garbage gets discarded" and "capture inputs match runtime" to hold, yet nothing keeps them in sync — editing one and missing the others fails silently.

Extract named constants (e.g. module-level `DEFAULT_TEXT_TOP_K = 50`, `DEFAULT_AUDIO_TOP_K = 25`) and reference the same ones from all three sites.

Ccyest:
Also, please fix lint every time before pushing.

After all fixes, please run seedtts full set test on c=32 to see accuracy and throughput.

nagisa-kunhah:
## Benchmark Setup

| Item | Value |
|---|---|
| Benchmark | `benchmarks.eval.benchmark_tts_seedtts` |
| Mode | `--generate-only` |
| Dataset | `zhaochenyang20/seed-tts-eval-arrow` |
| Model | `OpenMOSS-Team/MOSS-TTS-v1.5` |
| Language | `en` |
| Max concurrency | `32` |
| Requests | `1088` |
| Config | `examples/configs/moss_tts_local_colocated.yaml` |
| Default native decode | `forward_native_decode_enabled=false` |
| Default graph padding | `decode_graph_padding=false` |

## Performance Comparison

| Metric | Current | Main rerun | Delta vs main |
|---|---:|---:|---:|
| Completed / total | 1088 / 1088 | 1088 / 1088 | 0 |
| Throughput, req/s | 5.032 | 5.005 | +0.54% |
| Audio throughput, s/s | 11.073 | 11.014 | +0.54% |
| Output throughput, tok/s | 276.8 | 275.3 | +0.54% |
| Latency mean, s | 6.321 | 6.355 | -0.54% |
| Latency p95, s | 7.600 | 7.650 | -0.65% |
| RTF mean | 3.0064 | 3.0213 | -0.49% |
| RTF p95 | 4.2352 | 4.2136 | +0.51% |
| Output tokens mean | 55.0 | 55.0 | 0.00% |
| Audio duration mean, s | 2.201 | 2.200 | +0.05% |

Main rerun commit: e16464272073df098bfea0db25a6924721e756ac (e164642 [Perf] Pool MOSS-TTS Local frame launch state (#759)). WER is comparable between main and current branch: wer_corpus=1.2637 on main vs 1.2377 on current over 1088/1088 samples, with current slightly lower.

Test method: run SeedTTS EN full c=32 generation for both main and the current branch, then run the ASR/WER evaluation on the generated audio and compare corpus-level WER over the same 1088 samples.

cc @Ccyest

Ccyest:
## Benchmark — MOSS-TTS-Local v1.5 (candidate vs upstream main)

**Setup:** baseline main @ `6ed012e` vs candidate (this PR). Model `OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5` (Local v1.5). 2× H100 80GB, not-colocate (backbone GPU0, codec GPU1). `benchmark_tts_seedtts`, SeedTTS EN full set (1088), `--token-count auto`, `--ref-format references`, 1 run per concurrency. Server: `FLASHINFER_USE_CUDA_NORM=1`, `max_running_requests=32`, `cuda_graph_bs=[1,2,4,8,16,32]`, `cuda_graph_max_bs=32` (both sides identical). All rows completed 1088/1088 with zero failed requests.

| concurrency | main WER | current WER |
|---|---|---|
| 4 | 2.59% | 2.80% |
| 16 | 2.41% | 2.24% |
| 32 | 2.75% | 2.31% |

Higher audio s/s and QPS are better; lower latency and RTF are better.

|---|---|---|---|---|
| 4 | mean latency (s) | 0.642 | 0.751 | +17.0% |
| 4 | p95 latency (s) | 0.880 | 1.086 | +23.4% |
| 4 | mean RTF | 0.149 | 0.177 | +18.6% |
| 4 | audio throughput (s/s) | 27.400 | 23.404 | -14.6% |
| 4 | request throughput (QPS) | 6.226 | 5.318 | -14.6% |
| 16 | mean latency (s) | 1.737 | 1.708 | -1.7% |
| 16 | p95 latency (s) | 2.018 | 1.937 | -4.0% |
| 16 | mean RTF | 0.416 | 0.410 | -1.6% |
| 16 | audio throughput (s/s) | 40.312 | 40.989 | +1.7% |
| 16 | request throughput (QPS) | 9.160 | 9.313 | +1.7% |
| 32 | mean latency (s) | 4.695 | 3.417 | -27.2% |
| 32 | p95 latency (s) | 5.917 | 3.868 | -34.6% |
| 32 | mean RTF | 1.138 | 0.825 | -27.5% |
| 32 | audio throughput (s/s) | 29.602 | 40.694 | +37.5% |
| 32 | request throughput (QPS) | 6.726 | 9.253 | +37.6% |

**Notes:**

1. We don't see a meaningful performance improvement from the native in-forward decode CUDA graph, so it is gated behind a feature flag (`forward_native_decode_enabled`) that **defaults to `False`**.
2. `bs=32` should in theory deliver higher throughput, but main currently runs in eager mode and effective max concurrency only reaches 16. Going beyond that is **out of scope** for this PR and will be addressed in a separate follow-up issue.

PR #763, opened 2026-06-12
[Hardware] Resolve GPU issues on Blackwell

zhaochenyang20:
As titled.


sglang_omni/models/qwen3_asr/stages.py
```diff
@@ -13,50 +27,20 @@ def create_sglang_qwen3_asr_executor(
...
-    # Leave unset by default so SGLang/Omni auto-tunes the static KV/radix
-    # memory budget from GPU size, chunked prefill, and CUDA graph settings.
     mem_fraction_static: float | None = None,
-    # SeedTTS WER inputs are unique audio clips, so caching audio embeddings on
-    # GPU has essentially no reuse and only retains runtime memory.
     mm_embedding_cache_size_bytes: int = 0,
-    # CUDA graph capture is fast for this ASR path; torch.compile dominates
-    # startup latency and gives little benefit for the benchmark workload.
     enable_torch_compile: bool = False,
+    mm_attention_backend: str | None = "triton_attn",
```
Ratish1: This should not default to `triton_attn` for every Qwen3-ASR launch. Current SGLang already has hardware-aware multimodal attention selection in `VisionAttention._determine_attention_backend()`: Hopper defaults to `fa3`, Blackwell SM100 defaults to `fa4`, and other CUDA paths fall back to `triton_attn`.

By setting `mm_attention_backend="triton_attn"` here, this PR bypasses that policy globally and can silently downgrade non-Blackwell runs, while also preventing fixed Blackwell paths from using the
backend SGLang would select.

we can keep this default as `None` and only override the backend when the current device actually requires the Blackwell fallback, or expose it as an explicit runtime/config override instead of changing
the Qwen3-ASR default for all hardware.

benchmarks/eval/benchmark_tts_seedtts.py
```diff
@@ -31,7 +31,7 @@
 
     python -m benchmarks.eval.benchmark_tts_seedtts \
         --meta zhaochenyang20/seed-tts-eval-arrow \
-        --model boson-sglang/higgs-audio-v3-tts-4b --port 8000 \
+        --model bosonai/higgs-audio-v3-tts-4b --port 8000 \
```
Ratish1, line 34: Why change this?

tests/unit_test/pipeline/test_gpu_compat.py
Ratish1, line 1: I dont think we need unit tests for this PR. wdyt?

benchmarks/eval/benchmark_tts_seedtts.py
zhaochenyang20: boson-sglang is not publiced.

tests/unit_test/pipeline/test_gpu_compat.py
zhaochenyang20: we can remove it.

Ratish1 (approved):

PR #767, opened 2026-06-12
fix: resolve tensor device mismatch and CI artifact glob in Video-AMM…

### LINKED ISSUE #765: CI stage 10 / stage 11 (Video-AMME Talker) fail frequently — need root-cause investigation
### Background

In the Qwen3-Omni CI pipeline, two **video-related** stages have been failing often:

| Stage | Name | Test file | Workflow job |
|-------|------|-----------|--------------|
| **Stage 10** | Video-AMME Talker (WER + speed) | `tests/test_model/test_qwen3_omni_videoamme_talker_ci.py` | `stage-10-videoamme-talker` |
| **Stage 11** | FP8 Thinker TP=2 (Video-AMME Talker) | `tests/test_model/test_qwen3_omni_videoamme_talker_tp2_ci.py` | `stage-11-thinker-tp2` |

Both are defined in `.github/workflows/test-qwen3-omni-ci.yaml`.

They run the **Video-AMME Talker** path (Video+Audio → Text+Audio) and check text accuracy, text-audio WER, and speed metrics. Stage 11 additionally covers the **FP8 Thinker TP=2** configuration.

### Problem

These two stages **fail frequently in CI**, which hurts PR merge confidence and overall CI reliability. The root cause is still unclear. Possibilities include:

1. **Correctness** — a bug in the video encoder / preprocessor path causing accuracy or WER threshold failures
2. **Resource pressure** — video inputs are memory-heavy, leading to **OOM** or GPU memory not being released in time
3. **Stability** — timeouts, flaky retries still failing, or environment/concurrency issues on the CI host

### Initial hypotheses

- **Encoder handling**: video encoding / preprocessing (e.g. `pixel_values_videos`, video token injection) may have edge cases that only show up in the full Talker pipeline.
- **OOM**: Stage 10 runs with concurrency 16 over 20 samples; Stage 11 adds TP=2 + FP8 on top of that, so memory pressure is higher. Timeouts are 30 / 35 minutes — failures may also be OOM or hangs rather than metric regressions.

### What we need help with

1. **Collect recent failure logs** from Actions for stage 10 / 11 and categorize failures (threshold assert, OOM / `CUDA out of memory`, timeout, process crash, etc.).
2. **Compare with stage 9** (`test_qwen3_omni_videoamme_ci.py`, Video-AMME accuracy + speed only). If stage 9 is stable but 10/11 are not, narrow the investigation to Talker / audio output / TP=2 paths.
3. **Reproduce locally or on CI** by running the two pytest files in isolation on the H20 CI host (or equivalent GPU), watching encoder-stage memory and stack traces.
4. **Propose a fix or mitigation** — e.g. encoder memory optimization, CI concurrency/sample tuning, threshold recalibration, or an actual bug fix.

### Why this matters

Video is a core capability of Qwen3-Omni. Getting stage 10 and 11 stable would significantly improve confidence in the **video+audio multimodal pipeline** and make CI much more trustworthy for merges.

### Relevant files

- Workflow: `.github/workflows/test-qwen3-omni-ci.yaml` (`stage-10-videoamme-talker`, `stage-11-thinker-tp2`)
- Tests: `tests/test_model/test_qwen3_omni_videoamme_talker_ci.py`, `tests/test_model/test_qwen3_omni_videoamme_talker_tp2_ci.py`
- Benchmark: `benchmarks/eval/benchmark_omni_videoamme.py`
- Flaky retry wrapper: `.github/scripts/run_flaky_pytest.sh`

### Suggested labels

`ci`, `bug`, `video`, `help wanted`

MelodyyyYin:
## Title

```
fix(ci): stabilize Qwen3-Omni Video-AMME Talker stages (#765)
```

## Body

Resolves #765 — the Qwen3-Omni Video-AMME Talker CI stages (stage 10 and the FP8 Thinker TP=2 stage 11) failed intermittently. This PR fixes the real bugs and, with maintainer sign-off, aligns the stage-11 speed gate with the config the fix actually requires.

### What we changed

1. **Talker prefill device mismatch.** `talker_prefill.py` ran `torch.stack` over mixed CPU/CUDA tensors. Move all tensors to a common device before the stack.

2. **Stage 11 GPU-1 OOM.** Under TP=2 the thinker rank-1 shares GPU 1 with `talker_ar` + `code2wav`; at `thinker-mem-fraction-static 0.55` the concurrency-16 video-prefill activation spike overran the card. Lower it to **0.40** and set `PYTORCH_ALLOC_CONF=expandable_segments:True` (PyTorch 2.9 renamed `PYTORCH_CUDA_ALLOC_CONF`; both names are set so the setting always applies). → no more OOM.

3. **`pidfd_getfd: Operation not permitted` on stage 11.** Once requests reached the disaggregated relay, it passes fds between stage processes via `pidfd_getfd`, which needs `CAP_SYS_PTRACE`. Add `--cap-add=SYS_PTRACE` to stage 11's container (only stage 11 uses the disaggregated server; the router/colocated stages do not). → requests complete.

4. **CI artifact globs.** The post-stage upload globs never matched pytest's `mktemp` output dirs (`mktemp("videoamme_audio")` → `videoamme_audio0`). Add the `*` wildcard for stages 4/6/8/10/11 (`mmmu_audio*` / `mmsu_audio*` / `videomme_audio*` / `videoamme_audio*`). Text-only stages 3/5/7/9 use `tmp_path / "x"` (literal subdir, no suffix) and are intentionally left unchanged.

5. **Orphan GPU cleanup + script consolidation.** The retry path missed orphan `multiprocessing.spawn` workers that hold `/dev/nvidia*` fds but are invisible to `nvidia-smi`. `delete_gpu_process.sh` gains a `--kill-orphans` flag (scan `/proc/*/fd` + pkill). The redundant `ensure_gpus_idle.sh` (which had become a thin shim) is removed, and its callers (`run_all_wer_ci_aligned.sh`, `benchmarks/benchmarker/utils.py`) plus the docs are repointed to `delete_gpu_process.sh --kill-orphans`. `omni-post-stage` now reaps orphans so a teardown that leaves VRAM-holding workers cannot time out the cleanup wait and turn a passing stage red; `omni-setup` stays without the flag (the container is fresh at stage start).

6. **Stage 11 latency gate 146.7 → 150 (maintainer-approved).** `latency_mean_s_max` was derived (`apply_slack`, 12.5% slack) from a P95 baseline of 130.381 s that was calibrated at `thinker-mem-fraction-static 0.55`. But 0.55 OOMs on H20 at concurrency 16 (requests fail and no latency is even produced) — which is exactly why fix (2) lowers it to 0.40. At 0.40 the stage runs at ~148 s, and **0 of 18 historical latency samples ever met 146.7** (min 146.764, mean 148.98) across multiple branches: the gate was unreachable for the only config that actually works. Every historical run did have at least one attempt ≤ 148.3 s, so a 150 s gate passes reliably under the existing flaky-retry wrapper. Pinned `latency_mean_s_max` to 150.0; the rtf and throughput gates are unchanged.

### Notes / non-changes

- **Stages stay parallel** (no `needs:` chain): the self-hosted runner is a single executor and already serializes jobs, so a serialization chain adds nothing.
- **Stage 10 accuracy** (Video-AMME Talker, `thinker-text` gate 0.65) is a **pre-existing flaky threshold**, not a regression from this PR: the model averages ~0.61 (floor 0.50) on the 20-sample eval, so it flips green/red run-to-run on unrelated branches too (it passed on one run of this branch and scored 0.60 on the very next run with no inference-path change). The test file and the 0.65 gate are not in this PR's diff and are left to maintainers.
- No other numeric/threshold values are changed.


zhaochenyang20 (approved):

sjmshsh:
https://github.com/sgl-project/sglang-omni/pull/766

I also made the related fix, but your workflow had already started, so I closed this PR for now.

keke0315:
Hi @MelodyyyYin , A simple mitigation to try is moving `code2wav` off GPU 1 in the TP=2 fixture.

In `tests/test_model/conftest.py` line 153:

--gpu-code2wav 1

could be changed to:

--gpu-code2wav 0

The failure logs show GPU 1 with very very little free memory, moving code2wav to GPU 0 might slightly reduce GPU 1 pressure

It may also be helpful:

In `tests/test_model/conftest.py` line 153:

 --thinker-mem-fraction-static, 

let's say, from 0.55 to 0.5, to leave more memory available under concurrency.

This should only affect the TP=2 fixture and should not change the other CI stages. I have also submitted jobs to reproduce this CI run on a similar setup and will share the results once it finishes.

Hope this would be helpful.

GaokaiZhang:
### Stage 11 (FP8 Thinker TP=2) — root cause and fix

The device-mismatch fix made **stage 10 green**, but **stage 11 kept failing for an unrelated reason**: a **GPU-1 memory OOM**, not a correctness bug. Under TP=2, GPU 1 co-hosts three stages — thinker **rank 1**, `talker_ar`, and `code2wav`. In the failing run all 10 requests returned HTTP 500; the worker traceback shows GPU 1 fully consumed by *legitimate* stages only — `thinker 84.69 + talker 9.90 + code2wav 0.74 GiB ≈ 95 GiB`, 7.44 MiB free — the thinker having grown ~32 GiB past its static reservation during the concurrency-16 video prefill. (All 3 flaky-retry attempts OOM'd.)

Two independent things made this fail consistently:

**1. Orphan TP workers poison retries** — already fixed here (the `delete_gpu_process.sh` → `ensure_gpus_idle.sh` commit).
The TP rank-1 `multiprocessing.spawn` children aren't listed by `nvidia-smi` as compute apps, so the old cleanup left them holding VRAM between attempts — once attempt 1 OOM'd, attempts 2–3 started dirty and were doomed. Confirmed locally: after a normal process kill, a spawn worker still held **~49 GiB**; only scanning `/proc/*/fd` for open `/dev/nvidia*` handles reclaims it.

**2. Too-tight per-attempt budget** — the remaining piece, **proposed in MelodyyyYin/sglang-omni#1 onto this branch** (`conftest.py`: thinker `mem_fraction_static 0.55 → 0.40`; `+ PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` on stage 11).
Even a *clean* attempt is borderline at 0.55. The thinker KV pool is ~94% unused at this size, so the smaller static reservation costs nothing and restores the headroom the co-located stages need for the prefill activation spike. `expandable_segments` cuts allocator fragmentation on the shared card.

**Validation** — reproduced in the CI image on 2× H100-80GB (*less* memory than the H20 runner, so a stricter test). Clean A/B on identical GPUs:

| Config | Requests | Peak GPU mem | OOM |
|---|---|---|---|
| `0.55` (stock) | **10/10 fail** | 80975 MiB (99.3%) | 2 — `thinker_tp1` killed |
| `0.40` + `expandable_segments` | **10/10 pass** | 77513 MiB (95%) | 0 |

The ~45 GiB concurrency-16 prefill activation spike overruns the card at `0.55`; the smaller static reservation makes room for it. On the H20's 95 GiB cards the same config leaves ~12 GiB more headroom than this 80 GiB stress test.

**Diff:** `tests/test_model/conftest.py` (thinker `0.55→0.40`) and `.github/workflows/test-qwen3-omni-ci.yaml` (stage-11 `PYTORCH_CUDA_ALLOC_CONF`), on top of the `ensure_gpus_idle.sh` retry-cleanup change.

keke0315:
> ### Stage 11 (FP8 Thinker TP=2) — root cause and fix
> The device-mismatch fix made **stage 10 green**, but **stage 11 kept failing for an unrelated reason**: a **GPU-1 memory OOM**, not a correctness bug. Under TP=2, GPU 1 co-hosts three stages — thinker **rank 1**, `talker_ar`, and `code2wav`. In the failing run all 10 requests returned HTTP 500; the worker traceback shows GPU 1 fully consumed by _legitimate_ stages only — `thinker 84.69 + talker 9.90 + code2wav 0.74 GiB ≈ 95 GiB`, 7.44 MiB free — the thinker having grown ~32 GiB past its static reservation during the concurrency-16 video prefill. (All 3 flaky-retry attempts OOM'd.)
> 
> Two independent things made this fail consistently:
> 
> **1. Orphan TP workers poison retries** — already fixed here (the `delete_gpu_process.sh` → `ensure_gpus_idle.sh` commit). The TP rank-1 `multiprocessing.spawn` children aren't listed by `nvidia-smi` as compute apps, so the old cleanup left them holding VRAM between attempts — once attempt 1 OOM'd, attempts 2–3 started dirty and were doomed. Confirmed locally: after a normal process kill, a spawn worker still held **~49 GiB**; only scanning `/proc/*/fd` for open `/dev/nvidia*` handles reclaims it.
> 
> **2. Too-tight per-attempt budget** — the remaining piece, **proposed in [MelodyyyYin#1](https://github.com/MelodyyyYin/sglang-omni/pull/1) onto this branch** (`conftest.py`: thinker `mem_fraction_static 0.55 → 0.40`; `+ PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` on stage 11). Even a _clean_ attempt is borderline at 0.55. The thinker KV pool is ~94% unused at this size, so the smaller static reservation costs nothing and restores the headroom the co-located stages need for the prefill activation spike. `expandable_segments` cuts allocator fragmentation on the shared card.
> 
> **Validation** — reproduced in the CI image on 2× H100-80GB (_less_ memory than the H20 runner, so a stricter test). Clean A/B on identical GPUs:
> 
> Config	Requests	Peak GPU mem	OOM
> `0.55` (stock)	**10/10 fail**	80975 MiB (99.3%)	2 — `thinker_tp1` killed
> `0.40` + `expandable_segments`	**10/10 pass**	77513 MiB (95%)	0
> The ~45 GiB concurrency-16 prefill activation spike overruns the card at `0.55`; the smaller static reservation makes room for it. On the H20's 95 GiB cards the same config leaves ~12 GiB more headroom than this 80 GiB stress test.
> 
> **Diff:** `tests/test_model/conftest.py` (thinker `0.55→0.40`) and `.github/workflows/test-qwen3-omni-ci.yaml` (stage-11 `PYTORCH_CUDA_ALLOC_CONF`), on top of the `ensure_gpus_idle.sh` retry-cleanup change.



> ### Stage 11 (FP8 Thinker TP=2) — root cause and fix
> The device-mismatch fix made **stage 10 green**, but **stage 11 kept failing for an unrelated reason**: a **GPU-1 memory OOM**, not a correctness bug. Under TP=2, GPU 1 co-hosts three stages — thinker **rank 1**, `talker_ar`, and `code2wav`. In the failing run all 10 requests returned HTTP 500; the worker traceback shows GPU 1 fully consumed by _legitimate_ stages only — `thinker 84.69 + talker 9.90 + code2wav 0.74 GiB ≈ 95 GiB`, 7.44 MiB free — the thinker having grown ~32 GiB past its static reservation during the concurrency-16 video prefill. (All 3 flaky-retry attempts OOM'd.)
> 
> Two independent things made this fail consistently:
> 
> **1. Orphan TP workers poison retries** — already fixed here (the `delete_gpu_process.sh` → `ensure_gpus_idle.sh` commit). The TP rank-1 `multiprocessing.spawn` children aren't listed by `nvidia-smi` as compute apps, so the old cleanup left them holding VRAM between attempts — once attempt 1 OOM'd, attempts 2–3 started dirty and were doomed. Confirmed locally: after a normal process kill, a spawn worker still held **~49 GiB**; only scanning `/proc/*/fd` for open `/dev/nvidia*` handles reclaims it.
> 
> **2. Too-tight per-attempt budget** — the remaining piece, **proposed in [MelodyyyYin#1](https://github.com/MelodyyyYin/sglang-omni/pull/1) onto this branch** (`conftest.py`: thinker `mem_fraction_static 0.55 → 0.40`; `+ PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True` on stage 11). Even a _clean_ attempt is borderline at 0.55. The thinker KV pool is ~94% unused at this size, so the smaller static reservation costs nothing and restores the headroom the co-located stages need for the prefill activation spike. `expandable_segments` cuts allocator fragmentation on the shared card.
> 
> **Validation** — reproduced in the CI image on 2× H100-80GB (_less_ memory than the H20 runner, so a stricter test). Clean A/B on identical GPUs:
> 
> Config	Requests	Peak GPU mem	OOM
> `0.55` (stock)	**10/10 fail**	80975 MiB (99.3%)	2 — `thinker_tp1` killed
> `0.40` + `expandable_segments`	**10/10 pass**	77513 MiB (95%)	0
> The ~45 GiB concurrency-16 prefill activation spike overruns the card at `0.55`; the smaller static reservation makes room for it. On the H20's 95 GiB cards the same config leaves ~12 GiB more headroom than this 80 GiB stress test.
> 
> **Diff:** `tests/test_model/conftest.py` (thinker `0.55→0.40`) and `.github/workflows/test-qwen3-omni-ci.yaml` (stage-11 `PYTORCH_CUDA_ALLOC_CONF`), on top of the `ensure_gpus_idle.sh` retry-cleanup change.

Please also check if the attempt failed due to Latency/RTF exceeding the threshold.

Reproduce occasional slight latency threshold violations (H100), and the CI environment (H20) consistently stays below the latency threshold.

.github/scripts/run_flaky_pytest.sh
```diff
@@ -42,7 +42,7 @@ should_retry() {
 
 cleanup_between_attempts() {
   echo "Cleaning GPU state before retry..."
-  if ! bash .github/scripts/delete_gpu_process.sh; then
+  if ! bash .github/scripts/ensure_gpus_idle.sh; then
```
zhaochenyang20: This is strange. `.github/scripts` we have [delete_gpu_process.sh](https://github.com/sgl-project/sglang-omni/blob/main/.github/scripts/delete_gpu_process.sh) and [ensure_gpus_idle.sh](https://github.com/sgl-project/sglang-omni/blob/main/.github/scripts/ensure_gpus_idle.sh). If we only need one, delete the redundant one.

.github/workflows/test-qwen3-omni-ci.yaml
```diff
@@ -424,18 +424,22 @@ jobs:
...
 
   stage-11-thinker-tp2:
     name: stage 11 - FP8 Thinker TP=2 (Video-AMME Talker)
     runs-on: [self-hosted]
     timeout-minutes: 35
     container:
       image: crpi-n6adu6llixz83q37.cn-hangzhou.personal.cr.aliyuncs.com/hongccc/sglang-omni:dev
+      # TP=2 disaggregated relay passes fds between stage processes via
+      # pidfd_getfd, which needs CAP_SYS_PTRACE (else every request 500s with
+      # "pidfd_getfd: Operation not permitted"). Only stage 11 uses this server.
```
zhaochenyang20: I do not understand this comment, explain it indetails. And, `Note (Yin):`

.github/workflows/test-qwen3-omni-ci.yaml
```diff
@@ -457,13 +461,16 @@ jobs:
         env:
           HF_ENDPOINT: https://hf-mirror.com
           TORCHINDUCTOR_CACHE_DIR: ${{ env.OMNI_CI_HOME }}/.torchinductor
+          # Reduce allocator fragmentation on GPU 1, where the TP=2 thinker
+          # rank-1 is co-located with talker_ar + code2wav (issue #765).
```
zhaochenyang20: does this means, Thinker is TP 2 but talker and code2wav are on CUDA:0. Thus CUDA:0 is easier to OOM than CUDA:1?

In this sense, we can use the minimal fix for stage 11 for now. But either way, we need to do multiple-talker/code2wav, or we should do TP to them both.

cc @JingwenGu0829 @ischencheng

tests/test_model/conftest.py
```diff
@@ -151,8 +151,14 @@ def qwen3_omni_fp8_talker_server_tp2(tmp_path_factory: pytest.TempPathFactory):
             "1",
             "--gpu-code2wav",
             "1",
+            # 0.40 (not 0.55): under TP=2 the thinker rank-1 shares GPU 1 with
+            # talker_ar + code2wav, so 0.55 left too little headroom for the
+            # thinker's concurrency-16 video-prefill activation spike -> OOM
+            # (issue #765, stage 11). The KV pool is ~94% unused at this size,
+            # so the smaller static reservation costs nothing and restores the
+            # margin the co-located stages need.
```
zhaochenyang20: add `note (yin):`

zhaochenyang20 (changes requested):

.github/scripts/run_flaky_pytest.sh
MelodyyyYin: The two scripts are not redundant — they have a parent-child relationship. `ensure_gpus_idle.sh` calls `delete_gpu_process.sh` internally (line 74). The distinction:

- `delete_gpu_process.sh`: kills processes that `nvidia-smi` lists as active compute apps
- `ensure_gpus_idle.sh`: first kills orphan processes invisible to `nvidia-smi` (detected by scanning `/proc/PID/fd` for open `/dev/nvidia*` file descriptors, which still hold VRAM), then delegates to `delete_gpu_process.sh`

`delete_gpu_process.sh` is kept because it is used standalone in other contexts (e.g. benchmarker calibration) where the orphan scan is not needed. `run_flaky_pytest.sh` specifically needs `ensure_gpus_idle.sh` because a TP=2 crash leaves orphan `multiprocessing.spawn` workers with live CUDA contexts that `delete_gpu_process.sh` alone would miss.

.github/workflows/test-qwen3-omni-ci.yaml
MelodyyyYin: **GPU layout (why GPU 1, not GPU 0):** The fixture configures `--gpu-talker 1 --gpu-code2wav 1`, so the TP=2 thinker shards rank-0 to GPU 0 and rank-1 to GPU 1, while talker and code2wav both land on GPU 1. GPU 1 therefore co-locates three residents (thinker rank-1, talker_ar, code2wav); GPU 0 only has thinker rank-0. That is why GPU 1 is the memory pressure point and the comment targets it.

.github/workflows/test-qwen3-omni-ci.yaml
MelodyyyYin: **What `pidfd_getfd` is:** Linux syscall that lets one process duplicate a file descriptor from another process's fd table. The TP=2 disaggregated server runs four stage processes (thinker rank-0, thinker rank-1, talker_ar, code2wav); the relay coordinator passes socket/shm fds between them at request time using `pidfd_getfd`. Docker containers do not have `CAP_SYS_PTRACE` by default, so the syscall fails with EPERM → every routed request returns a 500 with "pidfd_getfd: Operation not permitted". Adding `--cap-add=SYS_PTRACE` grants the capability and the relay works normally. Stage 11 is the only stage that runs this multi-process TP=2 server, so the cap-add is scoped to stage 11 only.

**Comment format:** Updated to `# note (Yue Yin):` format in commit `34d5462`. Sorry for missing this in the first pass — that comment came in via the PR #2 merge and wasn't caught.

**`PYTORCH_CUDA_ALLOC_CONF` deprecation:** The env var was renamed to `PYTORCH_ALLOC_CONF` in PyTorch 2.9.1. Fixed in commit `41d06c4`.

.github/scripts/run_flaky_pytest.sh
zhaochenyang20: No. This is not first principled. These two files have similar naming. At least you can merge it into one.

MelodyyyYin: Done, here's the updated version — merged the orphan scan into delete_gpu_process.sh so it now kills both
  nvidia-smi-visible processes and /proc/fd-visible orphans by default. ensure_gpus_idle.sh is now a one-line shim
  for backward compat, and run_flaky_pytest.sh calls delete_gpu_process.sh directly.

  On your two concerns:
  - Only target specified GPUs: already scoped via CUDA_VISIBLE_DEVICES, the orphan scan respects the same GPU
  filter.
  - Don't kill others: each CI stage runs in its own PID namespace, so the scan only sees processes within the
  current container — other users' processes are invisible by construction.
  
  Current memory threshold is 1024 MiB, stricter than 2 GB, so that's fine as-is.

.github/workflows/test-qwen3-omni-ci.yaml
```diff
@@ -350,6 +357,7 @@ jobs:
 
   stage-9-videoamme:
     name: stage 9 - Video-AMME accuracy + speed
+    needs: [stage-8-videomme-talker]
```
zhaochenyang20: why do you change this? Make the test into sequential? Revert this back. All the tests should run in parallel.

zhaochenyang20 (changes requested):

.github/scripts/ensure_gpus_idle.sh
zhaochenyang20, line 1: By merge, I mean delete one file. Only keep one

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

PR #461, opened 2026-05-17
[RFC] stage-gpu-process topology

JingwenGu0829:
Issue #459

PR #768, opened 2026-06-13
Fix: bound IPC namespace prefix to keep ipc:// paths under sun_path limit

yxs:
## Problem

Serving a model whose path is long (e.g. a full HuggingFace cache path) crashes at startup:

```
zmq.error.ZMQError: ipc path "/tmp/sglang_omni/home-sglang-omni-cache-huggingface-hub-models--.../stage_preprocessing.sock" is longer than 107 characters (sizeof(sockaddr_un.sun_path)).
```

`create_ipc_runtime_dir` (`sglang_omni/pipeline/runtime_config.py`) derives the per-run IPC directory from `config.name` — which defaults to the **model path** when `--model-name` is unset — with no length bound:

```python
namespace_prefix = re.sub(r"[^0-9a-z]+", "-", config.name.lower()).strip("-")
path = Path(tempfile.mkdtemp(prefix=f"{namespace_prefix}-", dir=base_root))
```

The longest endpoint built from this dir is `ipc://{base_dir}/stage_{name}.sock`. Unix domain sockets cap the path (`sockaddr_un.sun_path`) at ~107 chars, so a long model path overflows it and ZMQ refuses to bind.

## When it was introduced

- The unbounded `config.name`-derived prefix landed in **#461** (`[Refactor] stage-gpu-process topology`) — harmless then, because the control plane used **TCP** endpoints, which have no `sun_path` limit.
- It became a real crash in **#509** (`[Refactor] Remove TCP control-plane endpoints`), which switched the control plane to `ipc://` Unix domain sockets.

## Fix

Bound the model-name-derived prefix so the longest `ipc://` path always fits, computed from the fixed parts of the path:

```python
overhead = len(str(base_root)) + _ENDPOINT_PATH_OVERHEAD + _LONGEST_STAGE_NAME
namespace_prefix = namespace_prefix[: max(8, _IPC_SUN_PATH_BUDGET - overhead)]
```

`mkdtemp`'s random suffix still guarantees uniqueness, so only the human-readable prefix is truncated. Robust to a long `base_path` too, not just long model names.

## Verification

Calling the real `create_ipc_runtime_dir` with a 127-char model name + a real `zmq` ipc bind:

| | longest ipc path | `zmq` bind |
|---|---|---|
| before | 174 chars | **FAILED** (`sun_path > 107`) |
| after | 89 chars | **OK** |

### DIFF
--- a/sglang_omni/pipeline/runtime_config.py
+++ b/sglang_omni/pipeline/runtime_config.py
@@ -10,15 +10,17 @@ import tempfile
 from dataclasses import dataclass
 from pathlib import Path
 
+import zmq
+
 from sglang_omni.config.placement import StagePlacementPlan, build_stage_placement_plan
 from sglang_omni.config.schema import PipelineConfig, StageConfig
 from sglang_omni.config.topology import ProcessTopologyPlan, build_process_topology_plan
 
 logger = logging.getLogger(__name__)
 
-_IPC_SUN_PATH_BUDGET = 100
-_LONGEST_STAGE_NAME = 24
-_ENDPOINT_PATH_OVERHEAD = len("/-XXXXXXXX/stage_.sock")
+# PyZMQ checks the filesystem path after ``ipc://`` against this budget.
+_IPC_SUN_PATH_BUDGET = getattr(zmq, "IPC_PATH_MAX_LEN", 100)
+_TEMPFILE_RANDOM_SUFFIX_LEN = 8
 
 
 class IpcRuntimeDir:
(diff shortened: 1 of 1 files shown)


sglang_omni/pipeline/runtime_config.py
```diff
@@ -67,6 +71,8 @@ def create_ipc_runtime_dir(config: PipelineConfig) -> IpcRuntimeDir:
     namespace_prefix = re.sub(r"[^0-9a-z]+", "-", config.name.lower()).strip("-")
     if not namespace_prefix:
         namespace_prefix = "pipeline"
+    overhead = len(str(base_root)) + _ENDPOINT_PATH_OVERHEAD + _LONGEST_STAGE_NAME
+    namespace_prefix = namespace_prefix[: max(8, _IPC_SUN_PATH_BUDGET - overhead)]
```
Hayden727: This still does not enforce the IPC path budget in all valid configs. `max(8, _IPC_SUN_PATH_BUDGET - overhead)` keeps at least eight namespace chars even when a long `endpoints.base_path` leaves no room, and `_LONGEST_STAGE_NAME` can be lower than the real endpoint suffix because `StageConfig.name` is not length-constrained. I reproduced this on the PR head with a temp base path: the generated stage socket path was 109 chars while `zmq.IPC_PATH_MAX_LEN` was 103, and `zmq.Socket.bind()` failed. The budget should be derived from the actual endpoint suffixes and fail fast when the base path/stage names alone cannot fit.

Hayden727 (commented):
One additional runtime contract issue found while validating the bot comments.

sglang_omni/pipeline/runtime_config.py
Hayden727: Fixed in 1becf8b8. Re-validated on the updated head: compileall passes, and direct ZMQ bind checks for long model path + temp base path and long stage name + default base path both bind successfully with longest socket path 103/103.

Hayden727 (approved):

PR #769, opened 2026-06-13
[Perf][Qwen3-Omni] Prefill batching for concurrent streaming + rollout-safety foundation (#760)

luojiaxuan:
## Summary

Addresses #760 (Milestone 1). Two layered commits in one PR:

1. **`[Qwen3-Omni] scheduling: admission control + observability for rollout safety`** — a small diagnostics/safety foundation (env + metric helpers, queue-time admission control, scheduler/stage telemetry, `Admission rejected` → HTTP 429). The #760 work reuses these helpers, and this instrumentation is what surfaced the OOM noted below.
2. **`[Perf][Qwen3-Omni] prefill batching for concurrent streaming (#760)`** — the actual #760 fix: opt-in, TP-safe prefill-batching knobs.

All new knobs default to **current behavior** (off / no-op), so existing CI thresholds are unchanged.

## The problem (#760)

Under many concurrent streaming sessions the thinker fragments prefill into tiny (size 1–2) batches instead of coalescing toward the running/token budget the way vLLM continuous batching does. It is worst for a TP>1 thinker on prefill-heavy streaming (e.g. live speech translation).

## What's in commit 2

| Knob | Default | Effect |
|---|---|---|
| `SGLANG_OMNI_PREFILL_COALESCE_MS` | `0` (off) | Idle→prefill: briefly wait to group queued requests so the first prefill is not size-1. Never stalls active decode. |
| `SGLANG_OMNI_PREFILL_COALESCE_MIN` | `2` | Min queue depth before the window engages (avoids a low-concurrency latency tax). |
| `SGLANG_OMNI_PREFILL_COALESCE_MAX_TOKENS` | `chunked_prefill_size` | Token budget bounding the coalesced prefill + its transient broadcast/activation memory. |
| `SGLANG_OMNI_CHUNKED_PREFILL_SIZE` | unset | Enable chunked prefill for the thinker (talker projected-embeds path unaffected). |
| `SGLANG_OMNI_TP_IDLE_BROADCAST_SKIP` | `false` | Replace the per-loop `broadcast_pyobj` with a cheap int broadcast on idle TP loops. |
| `SGLANG_OMNI_LOG_PREFILL_STATS` | `false` | Periodic avg/max prefill-batch + queue-depth log to quantify fragmentation. |

Plus: auto-set `disable_custom_all_reduce` for a multi-process TP>1 thinker (the custom all-reduce kernel is not wired for the disaggregated TP launch), so a plain `serve` launch is correct without a manual override.

**TP-safety:** the coalesce window uses a **fixed iteration count** and a **`waiting_queue`-length break condition** (identical across ranks after the recv broadcast), so every rank issues the same number of collectives and stays in lockstep.

## Measurements (2× A6000, thinker TP=2, 30B base)

- **Text** sweep: coalescing lifts `avg_batch` ~N/2 → ~N; throughput/latency unchanged (text prefill is cheap / decode-bound).
- **Audio** sweep (~30s clips, ~400 prefill tok/req): coalescing lifts `avg_batch` ~N/3 → ~N, but throughput does **not** improve (baseline continuous batching already saturates thinker prefill, ~12k tok/s here).
- ⚠️ With a count-only target and **no** token budget, coalescing accumulated 32 audio requests (each carrying multimodal CUDA feature tensors through `broadcast_pyobj`) and **OOM-crashed `thinker_tp1` at N=32**. `..._MAX_TOKENS` (token budget) fixes it (0 failures, throughput == baseline) and `..._MIN` removes the low-N latency tax. The coalesce window is therefore documented as **experimental**; the always-safe wins are the TP `disable_custom_all_reduce` auto-fix, chunked prefill, the idle-broadcast skip, and the `[prefill-stats]` instrumentation.

## Test plan

- [x] Unit suite passes in the `lmsysorg/sglang-omni:dev` image (incl. new admission / telemetry / 429 and thinker-TP-override tests).
- [x] Text + audio TP=2 A/B on 2× A6000 / 30B (`[prefill-stats]` confirms the avg_batch lift; the OOM was reproduced and then fixed by the token budget).
- [x] Rebase onto latest `main` (this branch is based on an earlier `main`; the PR diff is clean, but I will rebase for merge).

## Notes for reviewers

- Opened as a **draft** for early feedback.
- An unrelated local change (inline-audio extraction in `preprocessor.py`) was intentionally **left out** to keep this PR focused; happy to send it as a separate PR.
- Planned follow-ups (separate PRs): **M2** = CUDA graph for multimodal prefill / re-enable custom all-reduce under multi-process TP; **M3** = a TP=2 concurrency-throughput CI stage.

### DIFF
--- a/benchmarks/eval/qwen3_omni_rollout_stress.py
+++ b/benchmarks/eval/qwen3_omni_rollout_stress.py
@@ -1,0 +1,316 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Rollout-style stress runner for Qwen3-Omni.
+
+This intentionally reuses the exact same image prompt N times so Qwen3-Omni
+operators can measure how same-prompt multi-rollout traffic scales before a
+dedicated rollout-aware fanout implementation exists.
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import base64
+import logging
+import time
+from dataclasses import asdict, replace
+from pathlib import Path
+from typing import Any
+
+import aiohttp
+
+from benchmarks.benchmarker.data import RequestResult
+from benchmarks.benchmarker.runner import BenchmarkRunner, RunConfig
+from benchmarks.benchmarker.utils import save_json_results, wait_for_service
+from benchmarks.dataset.mmmu import MMMUSample, image_to_data_uri, load_mmmu_samples
+from benchmarks.metrics.performance import compute_speed_metrics, print_speed_summary
+
+logger = logging.getLogger(__name__)
+
+
+def _build_base_url(args: argparse.Namespace) -> str:
+    return args.base_url or f"http://{args.host}:{args.port}"
+
+
+def _parse_counts(raw: str) -> list[int]:
+    counts = [int(value.strip()) for value in raw.split(",") if value.strip()]
+    if not counts or any(count <= 0 for count in counts):
+        raise argparse.ArgumentTypeError("rollout counts must be positive integers")
+    return counts
+
+
+def _duplicate_prompt(sample: MMMUSample, count: int) -> list[MMMUSample]:
+    return [
+        replace(
+            sample,
+            sample_id=f"{sample.sample_id}:rollout-n{count}-{index}",
+        )
+        for index in range(count)
+    ]
+
+
+def _make_rollout_send_fn(
+    *,
+    model_name: str,
+    api_url: str,
+    rollout_group_id: str,
+    max_tokens: int,
+    temperature: float,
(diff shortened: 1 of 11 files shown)


luojiaxuan:
Superseded by a single-responsibility split (addresses the fat-PR feedback):
- #770 - standalone concurrency/rollout stress tool
- #771 - **#760 prefill batching + TP fix** (the core change; default-off knobs, with an honest before/after A/B)
- #772 - admission control + HTTP 429 + telemetry (stacked on #771)

Closing this in favor of those.

PR #770, opened 2026-06-13
[Qwen3-Omni] benchmarks: concurrency / rollout stress runner

luojiaxuan:
## Summary
Standalone closed-loop concurrency / rollout pressure runner for Qwen3-Omni at
`benchmarks/stress_tests/qwen3_omni_rollout_stress.py`. **No runtime code
changes.**

It sends the same prompt at a sweep of concurrency levels and records per-level
**req/s, output tok/s, prompt-token counts, and p50/p95/p99 latency** (plus any
profiler events the server emits). This is the tool used to characterize
concurrent serving for issue #760 and to capture before/after numbers.

It lives under `benchmarks/stress_tests/` (not `benchmarks/eval/`) because it is
a serving pressure / diagnostics tool, not a model-capability eval — keeping it
out of the eval suite avoids implying it is a standard accuracy metric. Usage
lives in the module docstring:

```bash
# run from the repo root (imports are rooted at `benchmarks.`)
python benchmarks/stress_tests/qwen3_omni_rollout_stress.py \
  --base-url http://localhost:8000 --rollout-counts 1,2,4,8,16,32 \
  --max-tokens 256 --output-dir results/rollout_stress
```

Drives a multimodal MMMU sample by default (`--repo-id` / `--sample-index`);
pass `--text-only` for a text prompt or `--prompt-override` to set the text.
Results are written to `<output-dir>/rollout_stress_results.json`; `--no-profile`
skips the server-side event profiler.

## Test plan
- [x] AST / syntax check
- [x] Drives a TP=2 Qwen3-Omni-30B server and emits a per-level JSON summary

---
_3-PR split of #769 (single-responsibility): stress tool #770 | #760 prefill batching + TP fix #771 | admission + telemetry #772 (stacked on #771)._

### DIFF
--- a/benchmarks/eval/benchmark_omni_rollout_stress.py
+++ b/benchmarks/eval/benchmark_omni_rollout_stress.py
@@ -5,13 +5,11 @@ Reuses the exact same prompt N times across a sweep of concurrency levels so
 operators can measure how same-prompt multi-rollout (and, more generally,
 many-concurrent-session) traffic scales: per-level requests/s, output tokens/s,
 prompt-token counts, and p50/p95/p99 latency, plus any profiler events the
-server emits. This is a pressure / diagnostics tool for serving (e.g. issue
-#760), not a model-capability eval -- hence ``benchmarks/stress_tests/`` rather
-than ``benchmarks/eval/``.
+server emits. This is a serving pressure / diagnostics tool (e.g. issue #760).
 
-Run from the repo root (imports are rooted at ``benchmarks.``)::
+Run from the repo root as a module so the ``benchmarks.`` imports resolve::
 
-    python benchmarks/stress_tests/qwen3_omni_rollout_stress.py \\
+    python -m benchmarks.eval.benchmark_omni_rollout_stress \\
         --base-url http://localhost:8000 \\
         --rollout-counts 1,2,4,8,16,32 \\
         --max-tokens 256 \\
--- a/benchmarks/eval/benchmark_omni_rollout_stress.py
+++ b/benchmarks/eval/benchmark_omni_rollout_stress.py
@@ -1,4 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
+# note (luojiaxuan):
 """Closed-loop concurrency / rollout stress runner for Qwen3-Omni.
 
 Reuses the exact same prompt N times across a sweep of concurrency levels so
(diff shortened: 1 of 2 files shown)


benchmarks/stress_tests/qwen3_omni_rollout_stress.py
```diff
@@ -0,0 +1,334 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Closed-loop concurrency / rollout stress runner for Qwen3-Omni.
+
+Reuses the exact same prompt N times across a sweep of concurrency levels so
+operators can measure how same-prompt multi-rollout (and, more generally,
+many-concurrent-session) traffic scales: per-level requests/s, prefill
+tokens/s, and p50/p95/p99 latency, plus any profiler events the server emits.
```
yxs: The docstring and PR body both claim it records "per-level req/s, **prefill tok/s**, and p50/p95/p99 latency". But what `compute_speed_metrics` actually produces: only `throughput_qps` (req/s), `latency_median/p95/p99_s`, `output_throughput` (completion tokens / wall = **decode** tok/s), and `prompt_tokens_total` / `prompt_tokens_mean` (prompt-token **counts**, not a rate).

luojiaxuan: Fixed: docstring and PR body now describe output tok/s and prompt-token counts instead of prefill tok/s.

benchmarks/eval/benchmark_omni_rollout_stress.py
```diff
@@ -0,0 +1,334 @@
...
+        --rollout-counts 1,2,4,8,16,32 \\
+        --max-tokens 256 \\
+        --output-dir results/rollout_stress
+
+By default it drives a multimodal MMMU sample (``--repo-id`` / ``--sample-index``)
+with audio output enabled; pass ``--text-only`` for a text-only prompt or
+``--prompt-override`` to supply the text. Results are written to
+``<output-dir>/rollout_stress_results.json``; pass ``--no-profile`` to skip the
+server-side event profiler.
+"""
```
zhaochenyang20, line 24: Adds your author name to this file.

zhaochenyang20 (dismissed):

benchmarks/eval/benchmark_omni_rollout_stress.py
luojiaxuan: Done

yxs (approved):

yxs:
@luojiaxuan 
Some non-blocking follow-ups:
- `_make_rollout_send_fn` overlaps a lot with the existing `make_mmmu_send_fn`, could reuse it later instead of keeping a second copy.
- Each level sends `count` requests once, so p95/p99 at low N are over very few samples; a `--repeats` option (like `benchmark_qwen3_asr_concurrency.py` has) would steady them.
- the file is now under `benchmarks/eval/`, which against PR body

PR #771, opened 2026-06-13
[Qwen3-Omni] TP correctness

luojiaxuan:
## Summary

Trimmed this PR down to the always-on TP correctness fix for Qwen3-Omni/Ming thinker TP launches.

- Add a shared `PipelineConfig.tensor_parallel_disable_custom_all_reduce_stages` hook.
- Use that hook for Qwen3-Omni and Ming thinker stages so TP>1 `sglang_omni serve` launches inject `disable_custom_all_reduce=True` for the thinker path.
- Keep non-thinker stages and TP=1 unchanged.
- Keep the Qwen3 thinker topology-gated custom all-reduce declaration so the follow-up topology gate can relax this only on verified P2P/NVLink meshes.

## Scope

This PR no longer includes the default-off prefill/coalescing/stats/idle-broadcast tuning machinery. Those diagnostics can come back separately if we want to continue that line of investigation.

## Test Plan

- [x] `git diff --check`
- [x] `python -m py_compile sglang_omni/config/schema.py sglang_omni/models/ming_omni/config.py sglang_omni/models/qwen3_omni/config.py tests/unit_test/qwen3_omni/test_pipeline.py`
- [ ] Targeted pytest was attempted, but local collection is blocked by the installed `deep_gemm` package asserting `CUDA_HOME is not None` during import.


sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -337,6 +343,189 @@ def __init__(self, dllm_config=None):
...
+        self._prefill_stats_reqs = 0
+        self._prefill_stats_max = 0
+
+    def _tp_inbox_is_idle_collective(self, local_count: int) -> bool:
+        """Agree across all TP ranks whether the entry rank's inbox was empty,
+        via a single cheap int broadcast, so idle loops can skip the (pickling)
+        ``broadcast_pyobj`` path.
+
+        TP-safe: every rank issues exactly one int broadcast here in lockstep,
+        and the boolean result is derived from the broadcast value (identical on
```
yxs: This method's `try` wraps both `torch.distributed.broadcast` (the collective) and the subsequent `int(flag.item()) == 0` (a local comparison) under `except Exception: return False`. To be clear: **this is not a live bug today.** Real failures are essentially symmetric — a collective either succeeds on all ranks or fails on all, and on failure every rank `return False`s and falls back to `broadcast_pyobj` in lockstep, no desync; and `flag.item()` on a valid CPU int64 tensor is essentially incapable of raising. So a deadlock would require the near-impossible asymmetric case "broadcast succeeds but one rank raises at `.item()`". The concern is structural: **a broad `except` wrapping a collective is precisely the anti-pattern that breeds TP hangs in this codebase** (and it contradicts the lockstep care taken everywhere else in the comments). Keeping the local comparison inside the `try` buys nothing and only widens the asymmetric-failure surface. And this knob (`SGLANG_OMNI_TP_IDLE_BROADCAST_SKIP`) is exactly the "benchmark before flipping the default" path — tighten it before any default flip.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -337,6 +343,189 @@ def __init__(self, dllm_config=None):
...
+            / 1000.0
+        )
+        _coalesce_target = env_int("SGLANG_OMNI_PREFILL_COALESCE_TARGET")
+        self._prefill_coalesce_target = (
+            _coalesce_target
+            if _coalesce_target and _coalesce_target > 0
+            else getattr(self, "max_running_requests", None)
+        )
+        # Minimum waiting-queue depth before the window engages, so a lone
+        # request never pays the idle wait (avoids the low-concurrency latency
```
yxs: `_prefill_coalesce_target` defaults to `max_running_requests` (lines 357-361). But `max_running_requests` is a concrete, typically large int in this codebase (line 180 does `self.max_running_requests // self.pp_size` with no None-guard, so it is always a number, usually tens to thousands). The only early-breaks in coalesce are `len(waiting_queue) >= target` (line 457) or the token budget tripping (line 459). At moderate concurrency (e.g. N=2-7, exactly the #760 stress range), the queue depth never reaches `target` and the token budget may not trip, so the loop runs the full `iterations = window_s/poll_s` window every time. In other words: once `coalesce_min <= queue < target` holds, that batch waits the entire `window_s` for nothing (even if no further requests arrive). The author's own measurement says "prefill batching improves ... though throughput remains compute-bound" — i.e. batching barely helps throughput (GPU-bound). Combined, at moderate concurrency this window is likely "pure added latency, no throughput gain", and once #770's stress tool enables this knob, the measured latency will be inflated and *mislead* the #760 tuning conclusion.

sglang_omni/models/qwen3_omni/config.py
```diff
@@ -266,6 +266,21 @@ def mem_fraction_role_to_stage(cls) -> dict[str, str]:
     def encoder_mem_reserve_role_to_stage(cls) -> dict[str, str]:
         return {"thinker": "thinker"}
 
+    @classmethod
+    def tensor_parallel_server_args_overrides(
```
yxs: This PR adds two byte-identical copies of `if stage_name == "thinker" and tp_size > 1: return {"disable_custom_all_reduce": True}` to the two Qwen3 configs. Together with MingOmni's existing 3 copies, there are now **5 identical implementations**, with the base class returning `{}`. I acknowledge this PR is **following an established convention** (MingOmni did it first), not inventing the smell; and Qwen3 uses the literal `"thinker"` while MingOmni uses the `THINKER_STAGE` constant — even that is inconsistent.

sglang_omni/scheduling/omni_scheduler.py
luojiaxuan: Agreed. Narrowed the try to the broadcast call only and added a regression test.

sglang_omni/scheduling/omni_scheduler.py
luojiaxuan: Agreed. The coalesce loop now stops on the first quiet poll instead of waiting the full window when no new requests arrive; PR body updated too.

sglang_omni/models/qwen3_omni/config.py
luojiaxuan: Agreed. Moved this into the shared PipelineConfig hook; Qwen3 and MingOmni now only declare their thinker TP stage.

sglang_omni/scheduling/sglang_backend/server_args_builder.py
```diff
@@ -18,6 +20,15 @@ def build_sglang_server_args(
     **overrides: Any,
 ) -> ServerArgs:
     """Build ServerArgs with shared defaults for all SGLang AR engines."""
+    if chunked_prefill_size is None:
+        # Issue #760: opt-in chunked prefill so long (audio) prefills interleave
```
yxs: clean up all AI comment, use `# Note (Jiaxuan):` if needed

sglang_omni/models/qwen3_omni/config.py
```diff
@@ -250,21 +251,26 @@ def _speech_stages(
 }
 
 
-class Qwen3OmniPipelineConfig(PipelineConfig):
-    """6-stage text-only pipeline."""
-
+class _Qwen3OmniBasePipelineConfig(PipelineConfig):
```
yxs, line 254: This PR adds `tensor_parallel_disable_custom_all_reduce_stages = (THINKER_STAGE,)` to `_Qwen3OmniBasePipelineConfig`, so the Qwen3 `sglang_omni serve` path injects `disable_custom_all_reduce=True` for a thinker TP>1. That part is correct (it fills the qwen3 TP-correctness gap). But #783's topology-relaxation mechanism opts in through a **separate** method, `topology_gated_custom_all_reduce_stages()`, and #783 only wired it for Ming — **not Qwen3**. With both PRs merged: the Qwen3 serve path injects disable=True, and #783's serve-layer gate, finding Qwen3 absent from the opt-in set, **will not** relax it → Qwen3 silently falls back to NCCL on NVLink, losing #783's measured +6.5%, with no error or log. Note #783 already switched the Qwen3 **example script** to `should_disable_thinker_custom_all_reduce` (topology-relaxed)

sglang_omni/scheduling/sglang_backend/server_args_builder.py
luojiaxuan: Already cleaned in 8aa6315; the remaining kept comments use the required prefix. Resolving this outdated thread.

sglang_omni/models/qwen3_omni/config.py
luojiaxuan: Done in c87c242: Qwen3 thinker now opts into `topology_gated_custom_all_reduce_stages()` and the Qwen3 TP test covers it.

sglang_omni/scheduling/sglang_backend/server_args_builder.py
```diff
@@ -18,6 +20,18 @@ def build_sglang_server_args(
     **overrides: Any,
 ) -> ServerArgs:
     """Build ServerArgs with shared defaults for all SGLang AR engines."""
+    if chunked_prefill_size is None:
+        # note (luojiaxuan): Issue #760: opt-in chunked prefill so long
```
yxs: just keep one

yxs:
@luojiaxuan , Could we trim this PR down to just the always-on TP-correctness fix, drop the default-off prefill machinery?
Based on PR body, it is not "effectively flat".

luojiaxuan:
Done in 8e33fc7: trimmed this PR down to the TP correctness path only. The default-off prefill/coalescing/stats/idle-broadcast machinery is removed from this PR, and I updated the PR body to match the new scope.

yxs (approved):

PR #772, opened 2026-06-13
[Qwen3-Omni] admission control + telemetry for rollout safety

luojiaxuan:
Opt-in **overload protection + observability** so concurrent serving fails
controlled instead of OOM-ing a worker. All guards default off / current
behavior.

Code only — no `docs/basic_usage` change. The admission knobs are operational
overload controls, so their mechanics are inline comments in
`scheduling/omni_scheduler.py` and the operator-facing table lives here rather
than in the basic-usage guide.

- `scheduler`: admission control at queue time — reject when the waiting queue
  is full (`SGLANG_OMNI_ADMISSION_MAX_WAITING`) or free CUDA memory is below a
  guard (`SGLANG_OMNI_ADMISSION_MIN_FREE_GPU_MEMORY_MB`), plus a per-request
  KV-headroom factor (`SGLANG_OMNI_ADMISSION_TOKEN_HEADROOM`).
- `serve/openai_api`: map an "Admission rejected" error to **HTTP 429** (instead
  of a worker OOM or a router 500).
- `observability`: CUDA-memory, KV-allocator, and queue-size helpers (extends
  the small module added in #771).
- `scheduler` + stage telemetry (queue/inbox/outbox/KV/CUDA snapshots) emitted
  through the existing event profiler when it is active.

## Admission knobs (all default OFF / current behavior)

| Env var | Default | Effect |
|---|---|---|
| `SGLANG_OMNI_ADMISSION_MAX_WAITING` | `max_queued_requests` | Reject new requests when the scheduler waiting queue is at/above this depth. |
| `SGLANG_OMNI_ADMISSION_MIN_FREE_GPU_MEMORY_MB` | unset (off) | Reject when free CUDA memory is below this guard. |
| `SGLANG_OMNI_ADMISSION_TOKEN_HEADROOM` | `1.0` | Per-request KV-headroom factor applied to the capacity check. |

Example (controlled overload instead of a crash):

```bash
SGLANG_OMNI_ADMISSION_MAX_WAITING=32 \
SGLANG_OMNI_ADMISSION_MIN_FREE_GPU_MEMORY_MB=4096 \
sgl-omni serve --model-path Qwen/Qwen3-Omni-30B-A3B-Instruct \
  --config examples/configs/qwen3_omni_colocated_h20.yaml --colocate --port 8008
```

Files unique to this PR (on top of #771): `pipeline/stage/runtime.py`,
`serve/openai_api.py`, `scheduling/observability.py` (+3 helpers),
`scheduling/omni_scheduler.py` (admission/telemetry), and the two test files.

## Test plan
- [x] Rebased cleanly onto the updated #771 (current `main`); re-ran
  `tests/unit_test/pipeline/test_scheduler.py` and
  `tests/unit_test/serve/test_openai_api.py` — **passed** (admission rejection,
  telemetry snapshot, and the 429 path).

---
_3-PR split of #769 (single-responsibility): stress tool #770 | #760 prefill batching + TP fix #771 | admission + telemetry #772 (stacked on #771)._

### DIFF
--- a/sglang_omni/scheduling/observability.py
+++ b/sglang_omni/scheduling/observability.py
@@ -1,0 +1,83 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Best-effort scheduler and CUDA telemetry helpers."""
+
+from __future__ import annotations
+
+import os
+from typing import Any
+
+
+def env_float(name: str, *, default: float) -> float:
+    value = os.environ.get(name)
+    if value is None or value.strip() == "":
+        return default
+    try:
+        return float(value)
+    except ValueError:
+        return default
+
+
+def env_int(name: str) -> int | None:
+    value = os.environ.get(name)
+    if value is None or value.strip() == "":
+        return None
+    try:
+        return int(value)
+    except ValueError:
+        return None
+
+
+def safe_qsize(queue_obj: Any) -> int | None:
+    qsize = getattr(queue_obj, "qsize", None)
+    if not callable(qsize):
+        return None
+    try:
+        return int(qsize())
+    except Exception:
+        return None
+
+
+def batch_size(batch: Any) -> int:
+    if batch is None:
+        return 0
+    return len(getattr(batch, "reqs", ()) or ())
+
+
+def allocator_available_tokens(allocator: Any) -> int | None:
+    available_size = getattr(allocator, "available_size", None)
+    if not callable(available_size):
+        return None
+    try:
+        return int(available_size())
+    except Exception:
+        return None
+
+
+def cuda_memory_snapshot(gpu_id: int | None) -> dict[str, int | None]:
+    snapshot: dict[str, int | None] = {
+        "gpu_id": gpu_id,
+        "cuda_allocated_bytes": None,
+        "cuda_reserved_bytes": None,
+        "cuda_max_allocated_bytes": None,
+        "cuda_free_bytes": None,
+        "cuda_total_bytes": None,
+    }
+    if gpu_id is None:
+        return snapshot
+    try:
+        import torch
+
+        if not torch.cuda.is_available():
+            return snapshot
+        device = int(gpu_id)
+        snapshot["cuda_allocated_bytes"] = int(torch.cuda.memory_allocated(device))
+        snapshot["cuda_reserved_bytes"] = int(torch.cuda.memory_reserved(device))
+        snapshot["cuda_max_allocated_bytes"] = int(
+            torch.cuda.max_memory_allocated(device)
+        )
+        free_bytes, total_bytes = torch.cuda.mem_get_info(device)
+        snapshot["cuda_free_bytes"] = int(free_bytes)
+        snapshot["cuda_total_bytes"] = int(total_bytes)
+    except Exception:
+        return snapshot
+    return snapshot
--- a/sglang_omni/scheduling/observability.py
+++ b/sglang_omni/scheduling/observability.py
@@ -3,28 +3,7 @@
 
 from __future__ import annotations
 
-import os
 from typing import Any
-
-
-def env_float(name: str, *, default: float) -> float:
-    value = os.environ.get(name)
-    if value is None or value.strip() == "":
-        return default
-    try:
-        return float(value)
-    except ValueError:
-        return default
-
-
-def env_int(name: str) -> int | None:
-    value = os.environ.get(name)
-    if value is None or value.strip() == "":
-        return None
-    try:
-        return int(value)
-    except ValueError:
-        return None
 
 
 def safe_qsize(queue_obj: Any) -> int | None:
--- a/sglang_omni/scheduling/omni_scheduler.py
+++ b/sglang_omni/scheduling/omni_scheduler.py
@@ -903,28 +903,18 @@ class OmniScheduler:
             return local_waiting
 
         value = torch.tensor([local_waiting], dtype=torch.int64)
-        try:
-            torch.distributed.all_reduce(
-                value,
-                op=torch.distributed.ReduceOp.MAX,
-                group=cpu_group,
-            )
-        except Exception:
-            logger.debug(
-                "Failed to compute TP-consistent waiting queue depth for admission",
-                exc_info=True,
-            )
-            return local_waiting
+        torch.distributed.all_reduce(
+            value,
+            op=torch.distributed.ReduceOp.MAX,
+            group=cpu_group,
+        )
         return int(value.item())
-
-    def _cuda_free_bytes(self) -> int | None:
-        snapshot = cuda_memory_snapshot(getattr(self, "gpu_id", None))
-        free_bytes = snapshot.get("cuda_free_bytes")
-        return int(free_bytes) if free_bytes is not None else None
 
     def _admission_cuda_free_bytes(self) -> int | None:
         """Return a TP-consistent free-memory value for admission decisions."""
-        local_free = self._cuda_free_bytes()
+        snapshot = cuda_memory_snapshot(getattr(self, "gpu_id", None))
+        free_bytes = snapshot.get("cuda_free_bytes")
+        local_free = int(free_bytes) if free_bytes is not None else None
         if getattr(self, "tp_size", 1) <= 1:
             return local_free
         cpu_group = getattr(self, "tp_cpu_group", None)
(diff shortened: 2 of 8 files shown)


sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -602,6 +858,93 @@ def _request_kv_capacity_error(self, req: Any) -> str | None:
             f"{mem_hint}"
         )
 
+    def _admission_rejection_reason(self, request_id: str, req: Any) -> str | None:
```
yxs: `_admission_rejection_reason` is called in the per-payload loop of `process_input_requests`, which runs on **every TP rank** over the same broadcast-replicated payloads — every rank must derive the same `waiting_queue`. The `max_waiting` guard uses `len(self.waiting_queue)` (replicated state) and is safe. But the `min_free` guard calls `_cuda_free_bytes()` → `cuda_memory_snapshot(self.gpu_id)` → `torch.cuda.mem_get_info(device)`, and `self.gpu_id = tp_worker.gpu_id` (line 141) is **each rank's own GPU**. With TP=2, rank0 is on GPU0 and rank1 on GPU1, and their free memory differs at any moment (fragmentation, other processes). So for the **same request**, rank0 may accept (free=5GB) while rank1 rejects (free=3GB < 4GB threshold). The rejecting rank does `abort + continue` (no append); the accepting rank does `append` — `waiting_queue` immediately differs by one. Worse: once diverged, every subsequent admission check (including the otherwise-safe `max_waiting`), #771's coalesce break conditions, and `get_next_batch_to_run`'s batch composition all diverge → TP collective desync / hang. The trigger is exactly the PR body's example: `SGLANG_OMNI_ADMISSION_MIN_FREE_GPU_MEMORY_MB=4096` with tp_size>1. This is the very bug #771 worked so hard to avoid, reintroduced by #772 on the same path.

luojiaxuan: Good catch. Fixed by reducing free CUDA memory to the TP-group minimum before admission; added a regression test.

luojiaxuan:
@yxs Fixed and resolved the thread. Also unstacked to make PR clearer. PTAL when you have a chance.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -611,6 +675,128 @@ def _request_kv_capacity_error(self, req: Any) -> str | None:
...
```
yxs: The `max_waiting` guard uses a purely local `len(self.waiting_queue) >= max_waiting` with no collective. Traced the full call chain, it is currently TP-safe: the request stream into `process_input_requests` is broadcast via `broadcast_pyobj` in `_recv_scheduler_messages` (only the entry rank reads its inbox; the rest receive the broadcast), and every `waiting_queue` mutation (append / abort / retract) is rank-symmetric with no `is_entry_rank` branch. So `len(waiting_queue)` is identical across ranks, the `max_waiting` branch is taken in lockstep, and the subsequent `min_free` `all_reduce` is always entered symmetrically — no deadlock.

The moment the `min_free` collective is skipped on one rank because `max_waiting` returned early there, you get a classic mismatched-collective deadlock. Safe today, but the day someone adds a leader-only `waiting_queue` mutation (even popping one entry on the entry rank only, or doing a per-rank abort off local state), `max_waiting` desyncs and the deadlock becomes real.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -611,6 +675,154 @@ def _request_kv_capacity_error(self, req: Any) -> str | None:
...
```
yxs: The signature is `(self, request_id: str, req: Any)`, but neither parameter is ever read in the body; the trailing `del request_id, req` "consumes" them to silence the linter's unused-argument check. This is a smell

sglang_omni/scheduling/observability.py
```diff
@@ -0,0 +1,83 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Best-effort scheduler and CUDA telemetry helpers."""
```
yxs, line 2: The file docstring says "Best-effort scheduler and CUDA telemetry helpers", but it mixes two distinct responsibilities
1. genuine telemetry/probing
2. plain env parsing

sglang_omni/scheduling/omni_scheduler.py
luojiaxuan: This is now guarded by `_admission_waiting_queue_depth()`, which uses TP `all_reduce(MAX)` before the early return so ranks make the same admission decision.

sglang_omni/scheduling/omni_scheduler.py
luojiaxuan: Done in 48c0a9a: removed the unused parameters and updated the callers/tests.

sglang_omni/scheduling/observability.py
luojiaxuan: Done in 48c0a9a: moved env parsing helpers into `sglang_omni.scheduling.env`, leaving observability focused on telemetry/queue helpers.

luojiaxuan (commented):
Reviewed. Admission control and telemetry are cohesive and there are no unresolved review threads; no extra simplification needed.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -611,6 +676,155 @@ def _request_kv_capacity_error(self, req: Any) -> str | None:
...
+                return (
+                    "Admission rejected: free CUDA memory is below the configured "
+                    "guard threshold "
+                    f"(free_bytes={free_bytes}, threshold_bytes={min_free}). "
+                    "Retry later or lower SGLANG_OMNI_ADMISSION_MIN_FREE_GPU_MEMORY_MB."
+                )
+
+        return None
+
+    def _admission_waiting_queue_depth(self) -> int:
```
yxs, line 896: Both methods are `try: all_reduce(...) except Exception: fall back to local / None`.
`all_reduce` is a collective, it must be called in matched lockstep across every rank in the group. The moment one rank's `all_reduce` actually throws, it catches, `return local_waiting`, proceeds, handles the next request, and calls all_reduce *again*, while the other ranks are still blocked inside the collective for *this* request.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -639,6 +853,10 @@ def _run_batch(self, batch, pp_proxy_tensors=None):
         a ``GenerationBatchResult``.  We bridge the two formats here.
         """
         self._emit_prefill_start_for_batch(batch)
+        self._emit_scheduler_telemetry(
+            "scheduler_batch_start",
```
yxs: `run_batch` is the per-forward hot path (called once per decode step).

When the event recorder `is_active()`, that's one CUDA sync injected into the decode loop on every forward step; at a few dozen decode steps/sec, that's a few dozen syncs/sec, measurably perturbing decode latency and throughput. This is that when `is_active()` is false the whole thing is a cheap no-op (one attribute check + a call), so the cost is **only paid when profiling is on**; but profiling is exactly when you want to measure true latency, and this observer effect contaminates the numbers you're measuring.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -611,6 +676,155 @@ def _request_kv_capacity_error(self, req: Any) -> str | None:
...
+            )
+        except Exception:
+            logger.debug(
+                "Failed to compute TP-consistent waiting queue depth for admission",
+                exc_info=True,
+            )
+            return local_waiting
+        return int(value.item())
+
+    def _cuda_free_bytes(self) -> int | None:
```
yxs: `_cuda_free_bytes` discards a full snapshot for one field, duplicating the sync on the admission path

sglang_omni/scheduling/omni_scheduler.py
luojiaxuan: Fixed. Removed full telemetry snapshots from the per-forward run_batch hot path.

sglang_omni/scheduling/omni_scheduler.py
luojiaxuan: Fixed. Admission free-memory reads now use one snapshot directly instead of the _cuda_free_bytes wrapper.

sglang_omni/scheduling/omni_scheduler.py
luojiaxuan: Fixed. Admission TP collectives now propagate failures instead of falling back locally, so ranks cannot diverge after a collective error.

PR #773, opened 2026-06-13
[Perf] Compile MOSS-TTS Local frame sampler before CUDA graph capture

Ratish1:
## Motivation

MOSS-TTS Local already captures the per-frame local-transformer decode path with CUDA graphs, but the seeded frame sampler still runs as an eager PyTorch tensor program inside that capture. This PR compiles only that sampler helper before frame CUDA graph capture, so normal MOSS Local serving gets the faster sampler path without adding a new CLI flag, server arg, or pipeline config option.

This is intentionally narrower than compiling the whole frame decode. The local transformer and logit projections stay on the existing eager math path; only `sample_seeded_branchless` is compiled and then used by the existing frame decode graph capture.

## Modifications

- Compile `sample_seeded_branchless` during `MossTTSLocalSGLangModel.init_frame_decode_graphs(...)`, before the frame CUDA graphs are captured.
- Reuse SGLang's CUDA-graph compile configuration and the existing `SGLANG_TORCH_COMPILE_MODE` fallback pattern, defaulting to `max-autotune-no-cudagraphs`.
- Keep the public MOSS Local serving/config surface unchanged. The optimization is an internal part of frame graph initialization, not a new user-facing compile target.
- Add a unit test that verifies the sampler callable is replaced by the compiled callable once and that the compile configuration hook is invoked once.

## Accuracy Tests

Full SeedTTS EN generation + Qwen3-ASR transcription was run through the normal MOSS Local server path.

WER comparison:

| metric | main | sampler | delta |
|---|---:|---:|---:|
| total_samples | 1088 | 1088 | 0 |
| evaluated | 1088 | 1088 | 0 |
| skipped | 0 | 0 | 0 |
| wer_corpus | 0.0249519 | 0.0249519 | +0.0000000 |
| wer_per_sample_mean | 0.0262710 | 0.0249839 | -0.0012871 |
| wer_per_sample_median | 0.0 | 0.0 | +0.0 |
| wer_per_sample_std | 0.107845 | 0.100036 | -0.007809 |
| wer_per_sample_p95 | 0.133333 | 0.125 | -0.008333 |
| wer_per_sample_max | 1.75 | 1.2 | -0.55 |
| wer_below_50_corpus | 0.0171395 | 0.0163907 | -0.0007488 |
| n_above_50_pct_wer | 9 | 9 | 0 |
| pct_above_50_pct_wer | 0.8272% | 0.8272% | +0.0000 pp |

The sampler branch completed all samples with no generation failures and no ASR skips. Corpus WER was unchanged in this paired run, below-50 WER improved slightly, and the >50% WER outlier count stayed the same. The normal server log confirmed the intended path:

```text
Compiled MOSS-TTS Local frame sampler (mode=max-autotune-no-cudagraphs)
MOSS-TTS Local frame-decode CUDA graphs captured for bs=[1, 2, 4, 8, 16]
```

## Speed Tests and Profiling

Full SeedTTS EN generation speed comparison from the same normal serving setup:

| metric | main | sampler | delta |
|---|---:|---:|---:|
| total_requests | 1088 | 1088 | 0 |
| completed_requests | 1088 | 1088 | 0 |
| failed_requests | 0 | 0 | 0 |
| latency_mean_s | 1.544 | 1.373 | -0.171 (-11.08%) |
| latency_median_s | 1.506 | 1.305 | -0.201 (-13.35%) |
| latency_p95_s | 2.349 | 2.083 | -0.266 (-11.32%) |
| latency_p99_s | 2.726 | 2.528 | -0.198 (-7.26%) |
| audio_duration_mean_s | 4.401 | 4.401 | +0.000 |
| rtf_mean | 0.364 | 0.3258 | -0.0382 (-10.49%) |
| rtf_median | 0.3414 | 0.3056 | -0.0358 (-10.49%) |
| rtf_p95 | 0.592 | 0.5537 | -0.0383 (-6.47%) |
| rtf_p99 | 0.7495 | 0.7179 | -0.0316 (-4.22%) |
| throughput_qps | 5.172 | 5.808 | +0.636 (+12.30%) |
| output_tok_per_req_s | 73.6 | 105.1 | +31.5 (+42.80%) |
| output_throughput | 284.6 | 319.5 | +34.9 (+12.26%) |
| output_tokens_mean | 55.0 | 55.0 | +0.0 |
| output_tokens_total | 59857 | 59859 | +2 |
| prompt_tokens_mean | 132.0 | 132.0 | +0.0 |
| prompt_tokens_total | 144051 | 144051 | 0 |
| audio_throughput_s_per_s | 22.764 | 25.563 | +2.799 (+12.30%) |

## Notes on compile boundary

During the investigation for MOSS Local torch.compile support, broader boundaries were tested as well. Compiling the full frame decode and compiling logit/projection helpers changed generated audio codes in direct frame parity checks. That is risky for MOSS Local because each sampled RVQ code feeds the next local-transformer step and future frame feedback.

The sampler-only boundary was the stable one: it keeps model math and logit projections unchanged while compiling the tensor sampling program that runs inside the existing frame graph capture.

## Checklist

- [ ] Add or update tests as needed.
- [ ] Update documentation as needed.

### DIFF
--- a/sglang_omni/models/moss_tts_local/sglang_model.py
+++ b/sglang_omni/models/moss_tts_local/sglang_model.py
@@ -15,6 +15,7 @@ the model runner after each backbone step.
 from __future__ import annotations
 
 import logging
+import os
 from typing import Any, Callable, Iterable, Optional, Tuple
 
 import torch
@@ -134,6 +135,8 @@ class MossTTSLocalSGLangModel(torch.nn.Module):
         # above. Allocated here, before any frame/backbone graph capture, so
         # its addresses are fixed for the process lifetime.
         self._state_pool = MossTTSLocalDecodeStatePool(self)
+        self._compiled_frame_sampler: Callable[..., torch.Tensor] | None = None
+        self._frame_compile_configured = False
 
     def acquire_row(self, rid: str) -> int:
         """Assign (or return the existing) decode-state pool row for ``rid``."""
@@ -357,6 +360,27 @@ class MossTTSLocalSGLangModel(torch.nn.Module):
 
     _sample_seeded_branchless = staticmethod(sample_seeded_branchless)
 
+    def _ensure_frame_compile_config(self) -> None:
+        if self._frame_compile_configured:
+            return
+        from sglang.srt.model_executor.cuda_graph_runner import set_torch_compile_config
+
+        set_torch_compile_config()
+        self._frame_compile_configured = True
+
+    def _ensure_frame_sampler_compile(self) -> None:
+        if self._compiled_frame_sampler is None:
+            compile_mode = os.environ.get(
+                "SGLANG_TORCH_COMPILE_MODE", "max-autotune-no-cudagraphs"
+            )
+            self._ensure_frame_compile_config()
+            self._compiled_frame_sampler = torch.compile(
+                sample_seeded_branchless,
+                mode=compile_mode,
+            )
+            self._sample_seeded_branchless = self._compiled_frame_sampler
+            logger.info(f"Compiled MOSS-TTS Local frame sampler (mode={compile_mode})")
+
     @torch.no_grad()
     def _decode_frame_graphable(
         self,
@@ -444,6 +468,8 @@ class MossTTSLocalSGLangModel(torch.nn.Module):
             max(max(buckets), max_eager_bs), device, self.dtype
         )
         self.local_transformer.freeze_kv_cache()
+        self._ensure_frame_sampler_compile()
+        frame_decode = self._decode_frame_graphable
         self._frame_graphs: dict[
             int,
             tuple[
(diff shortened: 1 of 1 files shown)


sglang_omni/models/moss_tts_local/sglang_model.py
```diff
@@ -494,7 +518,7 @@ def init_frame_decode_graphs(self, batch_sizes: list[int]) -> None:
                 feedback,
             )
         logger.info(
-            "MOSS-TTS Local frame-decode CUDA graphs captured for bs=%s", buckets
+            f"MOSS-TTS Local frame-decode CUDA graphs captured for bs={buckets}"
```
zhaochenyang20, line 521: good

zhaochenyang20 (approved):
good to go after @gaoyang07 's verification.

PR #782, opened 2026-06-14
docs(moss-tts): document MOSS-TTS-Local variant in cookbook

xinlij:
## Summary

Adds the **MOSS-TTS-Local-Transformer-v1.5** variant to the MOSS-TTS cookbook (`docs/cookbook/moss_tts.md`), which previously only documented the delay model (`MOSS-TTS-v1.5`).

### Prerequisites
- Distinguishes the two checkpoints (single-GPU delay model vs. two-GPU Local variant).
- Adds the `hf download OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5` command.
- Notes the codec (`MOSS-Audio-Tokenizer-v2`, 48 kHz output) is loaded via the checkpoint's remote code, so `trust_remote_code` must be enabled.

### Server Configuration
- Splits into two subsections, one per checkpoint.
- Documents the Local launch with `examples/configs/moss_tts_local.yaml` and its **two-GPU** default (AR engine on `cuda:0`, codec on `cuda:1`).
- Notes the single-GPU option via `config_cls: MossTTSLocalColocatedPipelineConfig`.

The request shape and generation parameters are identical to the delay model, so the existing synthesis examples apply to both variants. Note the output sample rate differs (Local 48 kHz vs. delay 24 kHz).

### Verified against the codebase
- Architecture (36-layer Qwen3 backbone + 1-layer frame-local transformer): `sglang_omni/models/moss_tts_local/sglang_model.py`.
- Two-GPU default (`codec_device="cuda:1"`) vs. colocated (`codec_device="cuda:0"`): `sglang_omni/models/moss_tts_local/config.py`.
- `config_cls: MossTTSLocalColocatedPipelineConfig` resolves via the `Variants` lookup in `sglang_omni/models/registry.py` (`get_config_cls_by_name`).
- 48 kHz output: `sglang_omni/models/moss_tts_local/payload_types.py` (`sample_rate = 48000`).

### DIFF
--- a/docs/cookbook/moss_tts.md
+++ b/docs/cookbook/moss_tts.md
@@ -9,21 +9,40 @@ duration control, and the vocoder reconstructs 24 kHz speech. In SGLang-Omni it 
 `preprocessing → tts_engine → vocoder` pipeline and is served through the OpenAI-compatible
 `/v1/audio/speech` endpoint.
 
+SGLang-Omni serves two MOSS-TTS checkpoints:
+
+- **`MOSS-TTS-v1.5`** (default) — the delay-pattern model described above; a single-GPU
+  `preprocessing → tts_engine → vocoder` pipeline.
+- **`MOSS-TTS-Local-Transformer-v1.5`** — the "Local" variant, which pairs the Qwen3 backbone
+  with a local frame transformer and the `MOSS-Audio-Tokenizer-v2` codec (reconstructing **48 kHz**
+  speech). Its pipeline defaults to **two GPUs** (the ~1B-param codec runs on a second device so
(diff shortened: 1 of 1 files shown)


docs/cookbook/moss_tts.md
gaoyang07, line 2: For readability and discoverability, we might consider splitting this section into two dedicated subsections from the beginning: one for the Delay Pattern model and one for the Local Transformer model. Each subsection could describe the corresponding architecture, serving pipeline, hardware assumptions, and usage examples. This would make the tradeoff between the two token modeling patterns more explicit and avoid mixing model-specific details in a single flow.

PR #783, opened 2026-06-14
perf(serve): gate thinker custom all-reduce on P2P/NVLink topology

luojiaxuan:
## Summary
- The TP-thinker config path (`PipelineConfig.tensor_parallel_server_args_overrides`, e.g. Ming-Omni) and the Qwen3-Omni speech example launcher force `disable_custom_all_reduce=True` for **every** `tp_size>1` thinker. So tensor-parallel ranks always fall back to NCCL even on NVLink, where custom (P2P) all-reduce is faster.
- This gates the disable on the **actual GPU topology**: custom all-reduce stays disabled unless NVML confirms every TP GPU pair is P2P-capable (e.g. NVLink), in which case it is left enabled.
- Detection is **NVML-only (no CUDA context created in the coordinator)** and **conservative** -- any uncertainty (pynvml missing, query error, `<2` GPUs) preserves the safe disabled default, so CPU/CI behavior is unchanged.

## Why
Found during the #760 TP-concurrency investigation. A step-phase micro-profile showed the Qwen3-Omni thinker is GPU-forward-bound; the per-layer TP all-reduce runs every decode/prefill step, and forcing NCCL leaves NVLink performance on the table.

## Measurement
2x RTX A6000 NVLink (NV4), pure `Qwen3-Omni-30B-A3B-Instruct` TP=2 (no-RAG), 32 concurrent streaming sessions (ACL6060 en->zh), 3 runs/mode (server warm), identical harness:

| mode (N=32) | seg/s runs | mean | p50 RTT | BLEU |
|---|---|---|---|---|
| stock (NCCL) | 7.43 / 7.98 / 8.10 | 7.835 +/-0.293 | 867 ms | ~32.7 |
| custom-AR (NVLink) | 7.73 / 8.72 / 8.58 | **8.343 +/-0.440** | 830 ms | ~32.5 |

**+6.5% throughput, BLEU parity.**

## Changes
- `sglang_omni/utils/gpu_compat.py`: `gpu_ids_support_p2p_mesh()` (NVML pairwise P2P status, tri-state) + `should_disable_thinker_custom_all_reduce()`.
- `sglang_omni/cli/serve.py`: in `_apply_tensor_parallel_server_args_overrides`, relax a TP `disable_custom_all_reduce=True` to `False` when the stage GPUs form a P2P mesh.
- `examples/run_qwen3_omni_speech_server.py`: use the helper instead of hardcoding `True`.
- Tests: helper unit test (mock NVML), Ming serve-CLI gate (P2P on/off), Qwen3 example (P2P on/off).

## Scope / notes
- Default behavior is unchanged on non-P2P topologies and whenever NVML can't confirm a mesh.
- The standalone Ming example launchers (`examples/run_ming_omni_*.py`) still set the flag directly; they can adopt the same helper in a trivial follow-up.

## Test plan
- [x] `pytest tests/unit_test/pipeline/test_gpu_compat_custom_ar.py tests/unit_test/ming_omni/test_omni_serve.py tests/unit_test/qwen3_omni/test_example_launcher.py` -> 63 passed
- [ ] CI unit suite

### DIFF
--- a/sglang_omni/cli/serve.py
+++ b/sglang_omni/cli/serve.py
@@ -12,6 +12,7 @@ from sglang_omni.preprocessing.resource_connector import (
     resolve_allowed_local_media_path,
 )
 from sglang_omni.serve.protocol import DEFAULT_TTS_BATCH_MAX_ITEMS
+from sglang_omni.utils.gpu_compat import should_disable_thinker_custom_all_reduce
 
 logger = logging.getLogger(__name__)
 
@@ -479,6 +480,40 @@ def _validate_colocated_gpu_override(
         )
 
 
+def _stage_tp_gpu_ids(stage: object) -> list[int]:
+    gpu = getattr(stage, "gpu", None)
+    if gpu is None:
+        return []
+    if isinstance(gpu, int):
+        return [gpu]
+    return [int(g) for g in gpu]
+
+
+def _gate_custom_all_reduce_on_topology(
+    stage: object,
+    updates: dict[str, object],
+) -> dict[str, object]:
+    """Relax a TP ``disable_custom_all_reduce=True`` override on capable topologies.
+
+    The config-level overrides disable custom all-reduce for every TP thinker.
+    Custom (P2P/NVLink) all-reduce is faster than NCCL and safe when the TP GPUs
+    form a P2P mesh, so re-enable it in that case; otherwise keep it disabled.
+    """
+    if updates.get("disable_custom_all_reduce") is not True:
+        return updates
+    gpu_ids = _stage_tp_gpu_ids(stage)
+    if should_disable_thinker_custom_all_reduce(gpu_ids):
+        return updates
+    refined = dict(updates)
+    refined["disable_custom_all_reduce"] = False
+    logger.info(
+        "Enabling custom all-reduce for stage '%s': GPUs %s form a P2P mesh",
+        getattr(stage, "name", "?"),
+        gpu_ids,
+    )
+    return refined
+
+
 def _apply_tensor_parallel_server_args_overrides(
     pipeline_config: PipelineConfig,
 ) -> None:
@@ -490,6 +525,7 @@ def _apply_tensor_parallel_server_args_overrides(
         )
         if not updates:
             continue
+        updates = _gate_custom_all_reduce_on_topology(stage, updates)
         _apply_stage_server_args_override(
             pipeline_config,
             stage_name=stage.name,
--- a/sglang_omni/cli/serve.py
+++ b/sglang_omni/cli/serve.py
@@ -12,7 +12,7 @@ from sglang_omni.preprocessing.resource_connector import (
     resolve_allowed_local_media_path,
 )
 from sglang_omni.serve.protocol import DEFAULT_TTS_BATCH_MAX_ITEMS
-from sglang_omni.utils.gpu_compat import should_disable_thinker_custom_all_reduce
+from sglang_omni.utils.gpu_compat import should_disable_custom_all_reduce_for_gpus
 
 logger = logging.getLogger(__name__)
 
@@ -493,32 +493,26 @@ def _gate_custom_all_reduce_on_topology(
     stage: object,
     updates: dict[str, object],
     *,
-    disable_custom_all_reduce: bool | None = None,
+    gpu_ids: tuple[int, ...],
+    should_disable: bool,
 ) -> dict[str, object]:
-    """Relax a TP ``disable_custom_all_reduce=True`` override on capable topologies.
+    """Relax a TP ``disable_custom_all_reduce=True`` override on a P2P-capable topology.
 
     The config-level overrides disable custom all-reduce for every TP thinker.
     Custom (P2P/NVLink) all-reduce is faster than NCCL and safe when the TP GPUs
-    form a P2P mesh, so re-enable it in that case; otherwise keep it disabled.
-    SGLang still performs its own communicator support checks during worker
-    startup before using the custom all-reduce path.
+    form a P2P mesh, so re-enable it when the caller's topology probe confirms one
+    (``should_disable`` is False); otherwise keep it disabled. SGLang still performs
+    its own communicator support checks during worker startup before using the
+    custom all-reduce path.
     """
-    if updates.get("disable_custom_all_reduce") is not True:
-        return updates
-    gpu_ids = _stage_tp_gpu_ids(stage)
-    should_disable = (
-        should_disable_thinker_custom_all_reduce(gpu_ids)
-        if disable_custom_all_reduce is None
-        else disable_custom_all_reduce
-    )
-    if should_disable:
+    if updates.get("disable_custom_all_reduce") is not True or should_disable:
         return updates
     refined = dict(updates)
     refined["disable_custom_all_reduce"] = False
     logger.info(
         "Enabling custom all-reduce for stage '%s': GPUs %s form a P2P mesh",
         getattr(stage, "name", "?"),
-        gpu_ids,
+        list(gpu_ids),
     )
     return refined
 
@@ -542,12 +536,13 @@ def _apply_tensor_parallel_server_args_overrides(
             gpu_ids = tuple(_stage_tp_gpu_ids(stage))
             if gpu_ids not in topology_gated_custom_ar_cache:
                 topology_gated_custom_ar_cache[gpu_ids] = (
-                    should_disable_thinker_custom_all_reduce(gpu_ids)
+                    should_disable_custom_all_reduce_for_gpus(gpu_ids)
                 )
             updates = _gate_custom_all_reduce_on_topology(
                 stage,
                 updates,
-                disable_custom_all_reduce=topology_gated_custom_ar_cache[gpu_ids],
+                gpu_ids=gpu_ids,
+                should_disable=topology_gated_custom_ar_cache[gpu_ids],
             )
         _apply_stage_server_args_override(
             pipeline_config,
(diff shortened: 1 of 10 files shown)


luojiaxuan:
throughput free win PR PTAK: https://github.com/sgl-project/sglang-omni/pull/783/changes/7b6d3418c80a4d1b8d7fe09afece98c922c28710

sglang_omni/utils/gpu_compat.py
```diff
@@ -146,3 +146,72 @@ def apply_gpu_compat_env_defaults(
         target_env[key] = value
         logger.info("Applied GPU compatibility env override: %s=%s", key, value)
     return overrides
+
+
+def gpu_ids_support_p2p_mesh(
```
yxs, line 151: #771's comment justifying `disable_custom_all_reduce=True` for the TP thinker reads: "Multi-process thinker TP requires the NCCL all-reduce path; the custom all-reduce kernel **is not wired for the disaggregated TP launch**." That is a **mechanism-level** reason — it sounds like custom all-reduce simply isn't wired up under the disaggregated multi-process TP launch, independent of topology. But #783 enables custom all-reduce on exactly that disaggregated launch by flipping the flag alone (no kernel wiring change) and measures +6.5% with BLEU parity on 2x A6000 NVLink. Both cannot be true: if #771's "not wired for the disaggregated TP launch" is literally correct, flipping a flag in #783 could not make it work. The likeliest truth is that #771's comment is **inaccurate** — the real constraint is **topology** (custom AR needs P2P; without it, it fails/falls back), not the launch mechanism. But until this ambiguity is resolved, there's a latent correctness risk: if #771's launch-wiring constraint is in fact **real** (and the 2-GPU case merely happened to dodge it), then "P2P mesh confirmed" is only a **necessary, not sufficient** condition, and enabling on larger TP (4/8) or other NVLink topologies could still crash or corrupt. Current validation stops at TP=2.

Hayden727 (commented):
[P2] Make the custom-AR relaxation an explicit topology-gated contract

`sglang_omni/cli/serve.py:492`

Issue: `_apply_tensor_parallel_server_args_overrides()` now runs every config-level TP override through `_gate_custom_all_reduce_on_topology()`, so any future `tensor_parallel_server_args_overrides()` result containing `disable_custom_all_reduce=True` can be flipped back to `False` solely because the stage GPUs form a P2P mesh. That is correct for the current Ming thinker policy, where the hardcoded disable appears to be topology-only, but the hook itself is generic: a model/stage could set the same ServerArgs field for a non-topology reason, e.g. a runner/launch contract that genuinely requires NCCL. The helper name is thinker-specific, but the application is all stages and all future config classes.

Evidence: traced the path `apply_parallelism_cli_overrides()` -> `_apply_tensor_parallel_server_args_overrides()` -> `_gate_custom_all_reduce_on_topology()` -> SGLang `ServerArgs.disable_custom_all_reduce`. In upstream SGLang `v0.5.12.post1`, custom AR has its own final guards (world size, same node, NVLink/P2P) and falls back when the communicator is disabled, so the measured NVLink path looks plausible. The Omni-side contract concern is that the shared override hook no longer preserves a config class's intentional hard disable if that disable is not topology-relaxable.

Impact: this makes the shared config hook unsafe for future TP stages/models: a future config can correctly return `disable_custom_all_reduce=True` for a model-specific or launch-mechanism reason, then `serve` silently overrides it on P2P/NVLink hardware.

Fix direction: make the topology relaxation opt-in at the owning policy layer. For example, keep the gate in Ming/Qwen thinker-specific launcher/config code, or extend the config hook to distinguish topology-gated disables from hard disables. A regression test with a dummy config returning a hard `disable_custom_all_reduce=True` while the topology helper returns `False` would pin the contract.

Validation performed on PR head `7b6d3418c80a4d1b8d7fe09afece98c922c28710`:
- `pytest tests/unit_test/pipeline/test_gpu_compat_custom_ar.py -q` -> 6 passed
- `python3 -m compileall -q sglang_omni/utils/gpu_compat.py sglang_omni/cli/serve.py examples/run_qwen3_omni_speech_server.py tests/unit_test/pipeline/test_gpu_compat_custom_ar.py` -> passed
- `git diff --check origin/main...origin/pr/783` -> clean
- Full PR test command could not run in my local env because `transformers` is not installed during collection.

sglang_omni/utils/gpu_compat.py
Hayden727: Addressed in `cc3ff103`.

The code now makes the topology relaxation an explicit config contract instead of treating every `disable_custom_all_reduce=True` as topology-only:

- `PipelineConfig.topology_gated_custom_all_reduce_stages()` defaults to an empty set, so hard disables remain hard disables.
- Ming opts only its thinker stage into topology relaxation; the Qwen example path remains thinker-specific.
- `_gate_custom_all_reduce_on_topology()` now documents that this is only the coordinator-side early gate and SGLang still performs communicator support checks during worker startup before actually using custom AR.
- Added a regression test proving a fake config with a hard `disable_custom_all_reduce=True` is not relaxed even when the topology helper reports P2P-capable.

I also traced upstream SGLang `v0.5.12.post1`: `CustomAllReduceV2` / legacy custom AR check same-node, supported world size, NVLink/P2P capability, and fall back disabled before the all-reduce call path. So #771’s “not wired for the disaggregated TP launch” wording is too broad; the PR no longer encodes that ambiguity as a generic framework rule.

Validation on `cc3ff103`:
- `pytest tests/unit_test/pipeline/test_gpu_compat_custom_ar.py -q` -> 6 passed
- `python3 -m compileall -q ...` over the edited Python files -> passed
- `git diff --check` -> clean
- Full targeted pytest still cannot collect in my local env because `transformers` is not installed.

luojiaxuan:
H100 custom all-reduce follow-up for this PR:

- Hardware/topology: 2x H100 80GB on GPU4/5; `nvidia-smi topo` shows NV18 between the two ranks.
- Setup: Qwen3-Omni thinker, BF16 TP=2, c32 text-only decode, `thinker_memory_fraction=mem_fraction_static=0.55`, CUDA graph hit rate 1.000.
- Both arms had the MRoPE text-only fast path enabled, so this isolates NCCL/control vs custom-AR.
- Toggle: diagnostic run forced `disable_custom_all_reduce=True` for the control arm and `False` for the custom-AR arm, matching the runtime path this PR should enable on a valid P2P/NVLink mesh.
- custom-AR was active in the logs: `custom_all_reduce: Registering 776 cuda graph addresses`.
- Throughput:
  - NCCL before: 4023.5 / 4019.9 / 3380.1 tok/s. The 3380.1 run is treated as shared-machine noise.
  - custom-AR: 4384.6 / 4466.1 / 4471.0 tok/s.
  - NCCL after: 4002.1 / 4090.5 tok/s.
  - Stable/control mean excluding the outlier is ~4034 tok/s; custom-AR mean is 4440.6 tok/s, about +10.1%.
  - Median delta is +11.1% vs the before-control and +10.4% vs the after-control.
- Phase view:
  - NCCL after: step ~7.31 ms, forward ~6.45 ms, host_pre ~0.48 ms.
  - custom-AR: step ~6.68 ms, forward ~5.85 ms, host_pre ~0.47 ms.
- Conclusion: the topology-gated custom-AR path appears beneficial on H100 NVLink as well, with a clear c32 text-only decode gain in this diagnostic harness.

sglang_omni/utils/gpu_compat.py
luojiaxuan: This was addressed by the explicit topology-gated contract in cc3ff103 and follow-up comment cleanup in b6b7f72; resolving the stale thread.

luojiaxuan (commented):
Reviewed. The topology gate helper is small and conservative; no simplification needed from this pass.

sglang_omni/cli/serve.py
```diff
@@ -479,17 +480,58 @@ def _validate_colocated_gpu_override(
...
+    )
+    return refined
+
+
 def _apply_tensor_parallel_server_args_overrides(
     pipeline_config: PipelineConfig,
 ) -> None:
     config_cls = type(pipeline_config)
+    topology_gated_custom_ar_stages = (
+        config_cls.topology_gated_custom_all_reduce_stages()
```
yxs, line 525: The serve-CLI path (`_apply_tensor_parallel_server_args_overrides`) only calls `_gate_custom_all_reduce_on_topology` for stages in the set returned by `config_cls.topology_gated_custom_all_reduce_stages()`. That classmethod is overridden only in Ming (`_MingOmniBasePipelineConfig` → `{THINKER_STAGE}`); Qwen3 (`_Qwen3OmniBasePipelineConfig`) does **not** override it and inherits the empty base set.

a Qwen3 thinker launched via `sgl-omni serve` keeps custom all-reduce hard-disabled even on an NVLink box, while the same Qwen3 thinker launched via `examples/run_qwen3_omni_speech_server.py` gets the topology relaxation. opposite behavior across the two entry points.

sglang_omni/utils/gpu_compat.py
```diff
@@ -146,3 +146,82 @@ def apply_gpu_compat_env_defaults(
...
```
yxs, line 151: NVML init/shutdown is paired per call, so if multiple gated stages ever exist (e.g. thinker plus a TP encoder), NVML gets re-initialized N times.

luojiaxuan: Fixed. The serve CLI caches topology decisions by TP GPU tuple, avoiding repeated NVML queries for multiple gated stages.

yxs (approved):

yxs:
on 2×H200 (NVLink) a graph-mode microbench puts the thinker's per-all-reduce at ~2µs (custom AR) vs ~18µs (NCCL), roughly 5–9× across decode batch sizes (bf16, hidden 2048).
It's one all-reduce per decode step, so the e2e gain scales with how decode-bound the run is: +8% at 512-token decode, tapering as prefill dominates on short decodes. NVML P2P-gated, so no effect and no regression where custom AR wouldn't help.

Issue #780, opened 2026-06-14
RL support roadmap for SGLang-Omni


JingwenGu0829:
# SGLang-Omni × Miles RL Roadmap (Revised 2026.7.20)

## Goal
Build an end-to-end RL stack for multimodal-input and audio-output models, using SGLang-Omni for rollout inference and Miles for rollout orchestration, reward computation, and training.

The roadmap follows this dependency order:

```mermaid
%%{init: {'flowchart': {'nodeSpacing': 25, 'rankSpacing': 30}, 'themeVariables': {'fontSize': '13px'}}}%%
flowchart TB
    IC["Inference contract (sglang-omni)"]
    MI["Miles audio + video input support"]
    H["Higgs"]
    FS["Fish Audio S2-Pro"]
    QT["Qwen3-Omni Thinker<br/>(image / audio / video input)"]
    FULL["Full Qwen3-Omni RL"]

    IC --> MI
    MI -->|TTS: audio output| H
    H --> FS --> FULL
    MI -->|Thinker workstream| QT
    QT --> FULL
```

### Inference side

The following inference infrastructure is already available and is not part of the remaining roadmap:

- [x] RL administration and disk-based refit: https://github.com/sgl-project/sglang-omni/pull/678
- [x] Distributed weight synchronization: https://github.com/sgl-project/sglang-omni/pull/784
- [x] Miles-compatible text `/generate`, token logprobs, metadata, and `weight_version`: https://github.com/sgl-project/sglang-omni/pull/785
- [x] Higgs multi-codebook actions, rollout logprobs, and action masks: https://github.com/sgl-project/sglang-omni/pull/823

- [x] multimodal inputs + Higgs rollout semantics: https://github.com/sgl-project/sglang-omni/pull/1032 and
- [ ] the remaining inference-side prerequisite is https://github.com/sgl-project/sglang-omni/pull/992

### Training side

Training-side work currently lives on the https://github.com/yxs/miles/ fork, not yet upstreamed to `radixark/miles` main,
https://github.com/radixark/miles/issues/1700 , miles main won't accept modality-specific RL in core yet.

- [ ] https://github.com/yxs/miles/pull/1
- [x] https://github.com/yxs/miles/pull/5

## 1. Audio and Video Input in Miles

Add first-class audio and video input support across Miles dataset processing, rollout requests, retries, and prefill-logprob replay. Existing image support must remain intact.

Miles should keep:
- Raw media sources for SGLang-Omni rollout requests.
[...]

PR #784, opened 2026-06-14
[RL] distributed weight-sync

yxs:
## What
#780

Implements the **distributed weight-update (RLHF refit) path** for the SGLang-Omni admin control plane, so an external trainer can stream updated weights into a *running* omni server over NCCL — no restart, no checkpoint round-trip to disk. This is the inference-side half of the RL loop.

Before this PR the three distributed endpoints were stubbed (`501 Not Implemented`). This PR un-stubs them and plumbs them end-to-end through every layer of the multi-stage pipeline.

## Why

In an RLHF rollout the trainer holds rank-0 and periodically broadcasts the freshly-updated policy weights to every inference replica. The inference engine must accept those weights live. This mirrors sglang's `update_weights_from_distributed` contract, adapted to the omni multi-stage (coordinator → stage → scheduler) architecture.

## Design

The control plane carries **metadata only** (parameter names / dtypes / shapes / group name); the tensors themselves move over a dedicated NCCL process group, never through the HTTP/IPC control path.

Three endpoints, plumbed `openai_api (HTTP) → Client → Coordinator.admin → Stage runtime → OmniScheduler → ModelWorker → sglang ModelRunner`:

| Endpoint | Purpose |
| --- | --- |
| `POST /init_weights_update_group` | Rendezvous a NCCL group with the trainer. Inference rank = `rank_offset + tp_rank`. |
| `POST /update_weights_from_distributed` | Receive the broadcast tensors, keyed by `(names, dtypes, shapes, group_name)`. |
| `POST /destroy_weights_update_group` | Tear the group down. |

Key properties:

- **Admin requests drain inline on the scheduler serving thread** (`_process_admin_requests()` at the top of each `_event_loop_*`), so a refit is serialized against generation rather than racing it. By default a refit is **rejected while requests are in flight** (`engine_paused` reporting reflects the real post-call state).
- **HTTP boundary is schema-guarded** — `UpdateWeightsFromDistributedRequest` declares `names` / `dtypes` / `shapes` as required and equal-length; the worker re-checks length before calling into sglang (whose `zip()` silently truncates on a length mismatch).
- **Router single-replica guard** — `/init_weights_update_group` is rejected (`422`) when more than one live worker is targeted, because each replica needs a *distinct* `rank_offset`; fanning one init to N replicas would collide ranks. Multi-replica refit is explicitly out of scope here.
- **Scope: only non-tied `body.*` transformer params are broadcast.** Tied audio-codebook weights need trainer-side `convert_to_hf` mapping, which is the training side's concern, not the inference control plane under test here.

1. **Refit efficiency test** — the e2e prints end-to-end refit latency (`[refit] N params updated in Xs`).
2. **Refit parity test** — load an *instruct* model, refit it to the *base* checkpoint over the distributed path, and assert that **every inference parameter the refit changes bit-matches a server loaded directly from base** (per-tensor SHA256). Parity is checked inference-vs-inference, not against the raw checkpoint, because the model renames and fuses params during `load_weights` (the 397 broadcast `body.*` checkpoint tensors collapse to ~226 fused parameters in the served model) — checkpoint and served model don't share keys. This proves bit-parity, not merely "the checksum changed".
3. **Refit router API** — the router endpoint + the single-replica guard above.

## Testing

**End-to-end (2-GPU):** `tests/test_model/test_rl_distributed_weight_update.py` — boots Higgs TTS-4B *instruct*, refits it to *base* over NCCL, and asserts every refit-changed inference param bit-matches a server loaded directly from base (SHA256). Confirms the server still serves audio; prints latency.

**Validated** on 2×H100, `sglang==0.5.12.post1`, torch 2.11+cu130: refit of all **397** broadcast `body.*` params in **0.868s**; **every refit-changed inference parameter bit-matched the base-loaded server (SHA256 parity)**; server kept serving audio after the refit.


sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -1115,20 +1145,62 @@ def _admin_update_weights_from_distributed(
                 "message": "stage does not support update_weights_from_distributed",
                 "data": {"skipped": True, "unsupported": True},
             }
+        return self._run_weight_update_with_lifecycle(
```
Hayden727, line 1160: This path can resume the worker after `update_weights_from_distributed` returns failure, but upstream SGLang reports that a failed distributed update may leave `ModelRunner` weights partially updated and should be discarded. Please keep this scheduler/worker paused or otherwise out of routing on this failure path until an explicit reload/recovery succeeds, instead of reusing the disk-update lifecycle that restores `_engine_paused` in `finally`.

sglang_omni_router/app.py
```diff
@@ -37,6 +37,9 @@
 _ADMIN_UPDATE_PATHS = {
     "/pause_generation",
     "/update_weights_from_disk",
+    "/update_weights_from_distributed",
+    "/init_weights_update_group",
```
Hayden727, line 41: Putting `/init_weights_update_group` on the temporary-disable lifecycle is not sufficient: if the trainer never joins, the worker scheduler can remain blocked in SGLang’s NCCL rendezvous after the admin request times out, and the router will restore this worker to routing even though `/health` only proves the coordinator is running. Please keep the init target disabled or mark it unhealthy on timeout/failure until an explicit recovery or scheduler round-trip succeeds.

Hayden727 (approved):

PR #785, opened 2026-06-14
[RL] add Miles-compatible /generate rollout endpoint

yxs:
## What

#780

Adds a Miles-compatible `POST /generate` RL-rollout endpoint to the worker serve app and the router, and threads rollout artifacts (sampled-token logprobs, `weight_version`) end-to-end so an RL trainer can drive generation through one call and read back what it needs for the policy update.

## Request / response contract

**Request** — exactly one prompt form, plus sampling + rollout controls:

| field | notes |
| --- | --- |
| `input_ids` | pre-tokenized prompt — the primary form Miles sends |
| `messages` | OpenAI-style chat messages |
| `prompt` | raw text |
| `sampling_params` | `temperature`, `top_p`, `max_new_tokens`, `seed`, … |
| `return_logprob` | default `true` |
| `metadata` | opaque; echoed back in `meta_info.request_metadata` |

> On Qwen3-Omni the preprocessing consumes `input_ids` or `messages`; a raw `prompt` string is reserved for text-only LLM pipelines.

**Response** — `text` + `meta_info`:

```json
{
  "text": "...",
  "meta_info": {
    "finish_reason": {"type": "stop", "length": null},
    "prompt_tokens": 13, "completion_tokens": 8,
    "weight_version": "default",
    "request_metadata": {"rollout_id": 1},
    "output_token_logprobs": [[-3.8e-05, 785], [0.0, 6722], ...]
  }
}
```

`output_token_logprobs` is one `[logprob, token_id]` per generated token, mirroring SGLang's native `meta_info`.

## Design

- **Logprobs come from the sampler, not recomputed from raw logits.** `sampled_logprobs_to_list` records the sampler-produced `next_token_logprobs`, preserving temperature / original-logprob semantics so the stored logprob belongs to the policy that actually sampled the token.
- **Pre-tokenized `input_ids` reach the thinker directly.** Miles sends the exact tokens it trains on; re-tokenizing server-side would risk rollout/training token drift. The Qwen3-Omni preprocessing was messages-only (`input_ids` 500'd at `normalize_messages`), so this PR adds a pre-tokenized branch that builds the thinker state directly and skips the chat template + HF processor (text-only; multimodal `input_ids` still go through the `messages` path).
- **Empty `output_token_logprobs` `[]` is preserved** — a zero-generated-token rollout is a valid value, kept distinct from "no logprobs captured" (`null`).
- **`omni_rollout` is forward-declared only.** The field is plumbed end-to-end but not populated in this PR; the multimodal trainable-action container is the next step. Miles sets `return_omni_rollout=false`.
- **Count fields (`prompt_tokens` / `completion_tokens` / `cached_tokens`) always serialize as ints** (0 when unknown), never `null` — RL consumers accumulate them (e.g. `cached_tokens += meta_info.get("cached_tokens", 0)`), where a `null` would raise `int + None`.

## Validation

End-to-end on **H100 + Qwen3-Omni-30B-A3B-Instruct (sglang 0.5.12.post1)**, against a live `/generate` server:

- **`input_ids` ≡ `messages` equivalence** — the same chat-templated prompt sent as `input_ids` and as `messages`, greedy decode, produces **identical generated token-ids** (`[785, 6722, 315, 9625, 374, 12095, 13, 151645]`) and identical text. The pre-tokenized path feeds the thinker with no drift.
- `output_token_logprobs` shape `[logprob, token_id]`, all logprobs ≤ 0, and **`len == completion_tokens`**.
- **Concurrent batch alignment** — 8 simultaneous rollout requests with different prompts each get `output_token_logprobs` whose token-ids decode back to *their own* response text (no batch cross-contamination).
- **Count fields are ints** — `cached_tokens == 0`, `prompt_tokens` integer (not `null`).
- `finish_reason` `stop` / `length` (with `length` set on cap), `weight_version`, `request_metadata` echo, `return_logprob=false` → `null` logprobs, ambiguous prompt inputs → `400`.
- The exact Miles rollout payload (`input_ids` + `return_omni_rollout=false` + `return_routed_experts` / `return_indexer_topk` + `output_modalities=["text"]` + `metadata`) → `200`.

**Unit tests:** 123 passed — request conversion, client artifact aggregation (incl. empty-list and multi-terminal decode), sampler-logprob recording (non-1.0 temperature regression + batch>1 row alignment), router opaque forwarding, and pre-tokenized preprocessing.

## Companion

Miles side: `miles.rollout.generate_hub.omni_thinker` (wires `--custom-generate-function-path` to this endpoint, sends `input_ids` + reads `output_token_logprobs`).

Dev fork of miles for SGLang-Omni RL: https://github.com/yxs/miles/


Hayden727 (commented):
Review notes from E2E validation of the /generate rollout path.

sglang_omni/model_runner/base.py
```diff
@@ -542,7 +546,33 @@ def _sample_next_token_ids(
...
         self._apply_codec_suppress_tokens(logits_output, requests)
-        return self.tp_worker.model_runner.sample(logits_output, forward_batch)
+        # Snapshot the post-penalty logits before sampling, while they are fresh:
+        # the async lookahead reuses this buffer one step later, so RL-rollout
+        # logprobs must be read here, not in _finalize.
+        rollout_logits = None
+        if any(getattr(sr.data, "return_logprob", False) for sr in requests):
+            logits = getattr(logits_output, "next_token_logits", None)
+            if logits is not None and logits.ndim == 2:
+                rollout_logits = logits.detach().clone()
```
Hayden727: [P1] This snapshots the raw post-penalty logits before calling the sampler, then `gather_sampled_logprobs` records `log_softmax(raw_logits)`. Upstream SGLang's normal `return_logprob` path reports logprobs after temperature scaling unless `SGLANG_RETURN_ORIGINAL_LOGPROB` is set, so any rollout with `temperature != 1` stores old logprobs for a different policy than the one that sampled the token. Please reuse the sampler-produced `next_token_logprobs` or mirror the sampler's temperature/original-logprob semantics, and add a non-1.0 temperature regression.

sglang_omni/serve/openai_api.py
```diff
@@ -690,6 +697,151 @@ def _build_chat_generate_request(req: ChatCompletionRequest) -> GenerateRequest:
...
+    if result.audio is not None:
+        audio = GenerateAudio(data=result.audio.data, format=audio_format)
+
+    meta_info = GenerateMetaInfo(
+        finish_reason=finish_reason,
+        prompt_tokens=prompt_tokens,
+        completion_tokens=completion_tokens,
+        weight_version=result.weight_version or "",
+        request_metadata=req.metadata,
+        output_token_logprobs=result.output_token_logprobs,
```
Hayden727: [P2] The request defaults `return_omni_rollout` to true and this flag is forwarded to stages, but the response builder only copies `output_token_logprobs`/`weight_version`. Since `CompletionResult`/`GenerateChunk` have no `omni_rollout` field, `/generate` returns `meta_info.omni_rollout: null` even when requested. Either plumb the action-trace payload through to `GenerateMetaInfo.omni_rollout`, or make the PR/docs explicit that this is only the text-AR subset.

sglang_omni/model_runner/base.py
Hayden727: Fixed in 3c35a848. The rollout path now enables sampler logprob capture and records the sampler-produced `next_token_logprobs`, so temperature/original-logprob semantics stay aligned with the token that was sampled. Added a non-1.0 temperature regression in `tests/unit_test/model_runner/test_rollout_logprobs.py` plus helper coverage in `tests/unit_test/scheduling/test_request_data.py`.

sglang_omni/serve/openai_api.py
Hayden727: Fixed in 3c35a848. `omni_rollout` is now part of `GenerateChunk` and `CompletionResult`, copied through the client result/stream builders, and emitted in `GenerateMetaInfo` when `return_omni_rollout` is requested. Added `/generate` and client tests for both present and disabled cases.

sglang_omni/serve/openai_api.py
```diff
@@ -690,6 +697,154 @@ def _build_chat_generate_request(req: ChatCompletionRequest) -> GenerateRequest:
...
+        prompt=None,
+        prompt_token_ids=req.input_ids,
+        messages=messages,
+        sampling=sampling,
+        stage_sampling=stage_sampling,
+        stage_params=req.stage_params,
+        extra_params=extra_params,
+        stream=req.stream,
+        max_tokens=sampling.max_new_tokens,
+        output_modalities=req.output_modalities,
```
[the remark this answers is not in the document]
Hayden727: Fixed in f84c6d2f. `_build_rollout_generate_request` now defaults omitted rollout `output_modalities` to `["text"]` while preserving explicit caller modalities, including audio. Added converter tests for both default text and explicit `text+audio` cases.

sglang_omni/serve/openai_api.py
```diff
@@ -690,6 +697,154 @@ def _build_chat_generate_request(req: ChatCompletionRequest) -> GenerateRequest:
...
+        audio = GenerateAudio(data=result.audio.data, format=audio_format)
+
+    meta_info = GenerateMetaInfo(
+        finish_reason=finish_reason,
+        prompt_tokens=prompt_tokens,
+        completion_tokens=completion_tokens,
+        weight_version=result.weight_version or "",
+        request_metadata=req.metadata,
+        output_token_logprobs=result.output_token_logprobs,
+        omni_rollout=result.omni_rollout if req.return_omni_rollout else None,
```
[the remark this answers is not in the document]
Hayden727: Fixed in f84c6d2f. `_build_generate_response` now emits `output_token_logprobs` only when `return_logprob` is true, matching the request-level control. Added a `/generate` response test for `return_logprob=false`.

sglang_omni/models/qwen3_omni/request_builders.py
```diff
@@ -796,6 +797,14 @@ def apply_thinker_result(
     if finish_reason is not None:
         thinker_out["finish_reason"] = finish_reason
 
+    weight_version = getattr(result, "weight_version", None)
+    if weight_version is not None:
+        thinker_out["weight_version"] = weight_version
+
+    output_token_logprobs = getattr(result, "output_token_logprobs", None)
+    if output_token_logprobs:
+        thinker_out["output_token_logprobs"] = output_token_logprobs
```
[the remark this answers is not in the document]
Hayden727: Fixed in f84c6d2f. `apply_thinker_result` now checks `output_token_logprobs is not None`, so an empty list remains a valid artifact instead of being dropped. Added a Qwen3 pipeline regression for the empty-list case.

sglang_omni/serve/openai_api.py
```diff
@@ -704,6 +711,158 @@ def _build_chat_generate_request(req: ChatCompletionRequest) -> GenerateRequest:
...
+    """Convert a rollout GenerateRequest into a client GenerateRequest."""
+    sampling = _rollout_sampling_from_dict(req.sampling_params or {})
+
+    messages: list[Message] | None = None
+    if req.messages is not None:
+        messages = [
+            Message(role=m.get("role", "user"), content=m.get("content"))
+            for m in req.messages
+        ]
+    elif req.prompt is not None:
```
JingwenGu0829: Here prompt is stored as a part of message instead of GenerateRequest.prompt which got asserted in the return as prompt=None, this may cause trouble in tts rl.

sglang_omni/models/qwen3_omni/request_builders.py
```diff
@@ -796,6 +797,14 @@ def apply_thinker_result(
     if finish_reason is not None:
         thinker_out["finish_reason"] = finish_reason
 
+    weight_version = getattr(result, "weight_version", None)
+    if weight_version is not None:
+        thinker_out["weight_version"] = weight_version
+
+    output_token_logprobs = getattr(result, "output_token_logprobs", None)
```
JingwenGu0829, line 804: `Client.completion()` collapses an empty `output_token_logprobs` list to `None`. Empty list is a valid "zero generated actions" value and should not be messed with missing logprob capture. Maybe we should track presence separately and add coverage for `[]` reaching `meta_info.output_token_logprobs`.

Hayden727 (approved):

yxs:
Regarding 58bf913d1abe7c9be143d276f54267543d4a8f3a, e5385311ddcb6e3efe4b3c19739d12a7f2149d89, 8d21758f76cc207de91cb381e430207d68179c5a.

All three target the same goal, no silent fallbacks on the /generate rollout path. The internal-contract cleanup and the boundary fail-loud are both the right direction.

few amends

1. Type the contract instead of hand-validating it. Replace `sampling_params: dict[str, Any]` + the manual whitelist with a typed `RolloutSamplingParams` (`extra="forbid"`), and `messages` with `RolloutMessage` (role/content required). This gives unknown-key rejection and value-type validation (e.g. `temperature="high"` → 422 instead of a deep 500) for free, removes whitelist drift, and subsumes most of the imperative validation in 8d21758f76cc207de91cb381e430207d68179c5a
2. Keep fail-loud proportional to actual rollout corruption. - `weight_version`: echo it (null before the first sync is fine) instead of 500 — a missing version label doesn't corrupt tokens/logprobs, and a hard 500 conflicts with the "deliberate optional" in e5385311ddcb6e3efe4b3c19739d12a7f2149d89. leave version tracking to the trainer may better.
3. Router: keep the `output_modalities` capability inference; drop the `metadata`-scan — routing shouldn't depend on opaque trainer metadata, and `x-sglang-omni-route-capabilities` already covers explicit overrides.

cc @JingwenGu0829 @luojiaxuan @Hayden727

PR #789, opened 2026-06-14
feat(qwen3-omni): enable mixed-chunk by default on the thinker path (…

luojiaxuan:
### What

Default `enable_mixed_chunk=True` and `chunked_prefill_size=8192` for the
Qwen3-Omni thinker, set in `create_sglang_thinker_executor_from_config`
(`sglang_omni/models/qwen3_omni/stages.py`). Both stay overridable via
`server_args_overrides` (pass `enable_mixed_chunk=False` to opt out). One file,
+16/-1. Relates to #760.

### Why

The thinker scheduler advances **prefill XOR decode** per step. Under concurrent
streaming a large fraction of steps are prefill-only, and any in-flight decodes
just wait for that step to finish. This is a duty-cycle problem, not an
occupancy or arrival problem.

Measured on `Qwen3-Omni-30B-A3B-Instruct`, TP=2, 32 concurrent streaming-SST
sessions (env-gated scheduler probe):

- ~**98%** of pure-prefill steps had decode-ready requests in `running_batch`
  that were passed over (**~18 per step**) — i.e. the workload is *not*
  arrival-bound.
- decode advanced on only ~58% of scheduler iterations.

Mixed-chunk folds those running decodes into the chunk-prefill (extend) step,
restoring the decode duty cycle.

### Result (same-node A/B, en→zh ACL6060 streaming SimulEval)

| metric | baseline | mixed-chunk default | delta |
|---|---|---|---|
| throughput (seg/s) | ~8.1 | ~9.3 | **+14%** |
| StreamLAAL_CA (computation-aware) | ~2280 | ~2143 | **-6%** |
| BLEU | 32.1 | 32.1 | parity |
| pure-prefill steps stranding ready decodes | ~98% | **~0%** | — |

Hardened: holds under swapped A/B ordering (mixed-first), **no regression at
N=1 / N=8**, and robust across `chunked_prefill_size ∈ {4096, 8192, 16384}`
(flat ~9.2–9.3 seg/s; 8192 chosen as a safe default).

### Verifying the default actually engages (no CLI flag)

Launched the thinker with **no** mixed-chunk CLI flags so the *code default* is
the only source of the feature. Runtime confirms:

```
[thinker args] ... chunked_prefill_size=8192 ...
[prefill attribution] prefill_steps=161 with_ready_decodes=0 (0.0%) ...
```

vs. the prior baseline (no default, no flag): `chunked_prefill_size != 8192` and
~96–98% of prefill steps stranding ready decodes.

### Scope / risk

- Affects the **Qwen3-Omni** thinker only. ming-omni has its own
  `create_sglang_thinker_executor_from_config`; the talker is untouched.
- `cli/serve.py` and the qwen3-omni stage `factory_args` do not set these keys,
  so nothing clobbers the default; an explicit `server_args_overrides` (or a
  `--chunked-prefill-size` / `--enable-mixed-chunk` flag) still wins.
- Mixed-chunk only engages when `chunked_prefill_size > 0`. For short streaming
  prefills the chunk size is a no-op; for very long prefills, chunking at 8192
  is the standard recommended behavior.

### DIFF
--- a/sglang_omni/models/qwen3_omni/stages.py
+++ b/sglang_omni/models/qwen3_omni/stages.py
@@ -943,7 +943,22 @@ def create_sglang_thinker_executor_from_config(
 ):
     """Returns OmniScheduler for thinker."""
 
-    overrides: dict[str, Any] = {"disable_cuda_graph": False}
+    # The thinker runs prefill XOR decode per scheduler step, so under
+    # concurrent streaming a large fraction of steps are prefill-only while
+    # in-flight decodes stall (measured on Qwen3-Omni-30B TP=2, 32 streams:
+    # ~98% of prefill steps had decode-ready reqs in running_batch passed over,
+    # ~18/step). Mixed-chunk folds those running decodes into the chunk-prefill
+    # (extend) step, restoring the decode duty cycle: +14% throughput and ~6%
+    # lower computation-aware latency at quality parity, with no low-concurrency
+    # regression and robustness to chunked_prefill_size (#760). Defaults here so
+    # the streaming-SST thinker path benefits out of the box; either key can be
+    # overridden via server_args_overrides (set enable_mixed_chunk=False to opt
+    # out). Note: mixed-chunk only engages when chunked_prefill_size > 0.
+    overrides: dict[str, Any] = {
+        "disable_cuda_graph": False,
+        "enable_mixed_chunk": True,
+        "chunked_prefill_size": 8192,
+    }
     if server_args_overrides:
         overrides.update(server_args_overrides)
     overrides["tp_size"] = tp_size
--- a/sglang_omni/models/qwen3_omni/stages.py
+++ b/sglang_omni/models/qwen3_omni/stages.py
@@ -942,7 +942,7 @@ def create_sglang_thinker_executor_from_config(
     total_gpu_memory_fraction: float | None = None,
 ):
     """Returns OmniScheduler for thinker."""
-    # note (luojiaxuan)
+    # note: (luojiaxuan)
     # The thinker runs prefill XOR decode per scheduler step, so under
     # concurrent streaming a large fraction of steps are prefill-only while
     # in-flight decodes stall (measured on Qwen3-Omni-30B TP=2, 32 streams:
--- a/sglang_omni/models/qwen3_omni/stages.py
+++ b/sglang_omni/models/qwen3_omni/stages.py
@@ -942,7 +942,7 @@ def create_sglang_thinker_executor_from_config(
     total_gpu_memory_fraction: float | None = None,
 ):
     """Returns OmniScheduler for thinker."""
-    # note: (luojiaxuan)
+    # note (luojiaxuan):
     # The thinker runs prefill XOR decode per scheduler step, so under
     # concurrent streaming a large fraction of steps are prefill-only while
     # in-flight decodes stall (measured on Qwen3-Omni-30B TP=2, 32 streams:
--- a/sglang_omni/models/qwen3_omni/stages.py
+++ b/sglang_omni/models/qwen3_omni/stages.py
@@ -942,7 +942,7 @@ def create_sglang_thinker_executor_from_config(
     total_gpu_memory_fraction: float | None = None,
 ):
     """Returns OmniScheduler for thinker."""
-
+    # note (luojiaxuan)
     # The thinker runs prefill XOR decode per scheduler step, so under
     # concurrent streaming a large fraction of steps are prefill-only while
     # in-flight decodes stall (measured on Qwen3-Omni-30B TP=2, 32 streams:


luojiaxuan:
Follow-up to the attribution work #760. Short version: the streaming-SST thinker
runs **prefill XOR decode** per scheduler step, so it stalls in-flight decodes
whenever it picks a prefill. Folding those decodes into the extend step
(`--enable-mixed-chunk`, already supported, just never enabled) is a **+14–18%
seg/s** win at quality parity — a pure flag flip.

**Full evidence + one-command repro:**
[`benchmarks/diagnostics/qwen3_omni_tp_vllm_gap/`](https://github.com/luojiaxuan/sglang-omni/tree/diag/qwen3-omni-tp-vllm-gap/benchmarks/diagnostics/qwen3_omni_tp_vllm_gap)
(fork, FINDINGS §6).

### 1. Attribution: is the prefill stall a real lever, or arrival-bound?
Instrumented each pure-prefill step to count decode-ready reqs sitting in
`running_batch` that were passed over (`[prefill attribution]`, env-gated).
Steady-state N=32:

- **96.7–98.1% of prefill steps had ready decodes waiting** (mean **~18–20**
  stranded/step); only ~2–3% were genuinely decode-empty.
- `waiting_queue ≈ 0.3` during prefill steps → these are *small single-new-chunk*
  prefills, **not** an admission backlog.

So the workload is **not arrival-bound**: sglang spends ~39% of steps prefill-only
while ~18 sessions sit decode-ready. Fusion is a valid lever.

### 2. The fix: enable mixed-chunk (clean same-node A/B, N=32)
`--enable-mixed-chunk --chunked-prefill-size 8192`, no code change:

| metric (cold / warm) | baseline | mixed-chunk |
|----------------------|----------|-------------|
| seg/s                | 7.05 / 8.21 | **8.30 / 9.37** (+18% / +14%) |
| BLEU                 | 31.9 / 32.3 | 32.9 / 32.8 (parity) |
| StreamLAAL_CA (ms)   | 2468 / 2241 | **2285 / 2115** (−6%) |
| prefill steps stranding decodes | **98.1%** (~20/step) | **0.0%** |

The probe confirms the mechanism: under mixed-chunk `ready_decode_hist={0: …}`
(every prefill step now has an empty `running_batch`) and decode-only steps drop
**1324 → 816** as that work moves into fused steps.

### 3. Hardening (all controls pass)
- **Swapped order** (mixed brought up *first*): mixed 9.28 / 9.26 vs baseline
  7.83 / 8.41 = **+14%** → not an ordering/caching artifact.
- **No low-concurrency regression**: N=1 0.935 vs 0.891 (+5%); N=8 4.314 vs
  4.303 (parity). The win is concentrated where the decode batch is contended.
- **chunked-prefill-size robust**: warm N=32 at 4096 / 8192 / 16384 =
  9.27 / 9.26 / 9.23 seg/s (flat); 8192 is a safe default.

### 4. Two corrections to earlier comments in this thread
- **"mixed-chunk → null" was a measurement error** — it compared 9.26 against an
  inflated warm baseline and called +6% noise. The clean back-to-back A/B + the
  attribution probe (98% → 0% starvation) show it is the one knob that actually
  moves throughput, because it targets the prefill/decode **duty cycle**, not
  occupancy.
- The decode bottleneck is **GPU-forward-bound**, not host-CPU (a
  `cuda.synchronize()` step-phase profile attributed the thinker's "100% CPU" to
  a GPU-sync wait). vs Other Framework the differentiator is the **duty cycle**: Other Framework fuses
  prefill+decode (~95% of steps advance decode) while stock sglang advanced
  decode on ~58%. Mixed-chunk closes most of that.

**Suggested action:** enable `--enable-mixed-chunk --chunked-prefill-size 8192`
by default for concurrent streaming. Happy to share the raw StreamLAAL/BLEU JSON
and server logs.

sglang_omni/models/qwen3_omni/stages.py
```diff
@@ -943,7 +943,22 @@ def create_sglang_thinker_executor_from_config(
...
+    # concurrent streaming a large fraction of steps are prefill-only while
+    # in-flight decodes stall (measured on Qwen3-Omni-30B TP=2, 32 streams:
+    # ~98% of prefill steps had decode-ready reqs in running_batch passed over,
+    # ~18/step). Mixed-chunk folds those running decodes into the chunk-prefill
+    # (extend) step, restoring the decode duty cycle: +14% throughput and ~6%
+    # lower computation-aware latency at quality parity, with no low-concurrency
+    # regression and robustness to chunked_prefill_size (#760). Defaults here so
+    # the streaming-SST thinker path benefits out of the box; either key can be
+    # overridden via server_args_overrides (set enable_mixed_chunk=False to opt
+    # out). Note: mixed-chunk only engages when chunked_prefill_size > 0.
```
zhaochenyang20, line 956: use `#note (your name)` to add comment.

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

PR #792, opened 2026-06-14
docs: add MOSS-TTS-Local cookbook

MelodyyyYin:
### Title

docs: add MOSS-TTS-Local cookbook

### Body

**What**

Add a user-facing cookbook for `OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5` at
`docs/cookbook/moss_tts_local.md`, and register it in the `docs/index.rst` cookbook toctree.

**Contents**

- Architecture intro and how it differs from MOSS-TTS (delay): frame-local decoding (no delay
  pattern), RVQ-12 audio codebooks, 48 kHz stereo output, RTF < 1 at concurrency 16.
- Prerequisites (model download; `soundfile` for data-URI refs).
- Server Configuration: the checkpoint auto-resolves `MossTTSLocalModel`, so `serve` needs only
  `--model-path` (no `--config`, like Higgs TTS); the default layout runs the AR engine and
  codec/vocoder on separate GPUs, so `CUDA_VISIBLE_DEVICES=0,1`.
- Synthesizing Speech: basic (reference-less), voice cloning (`references` / `ref_audio` +
  `ref_text`), Python, reference sources (path / URL / base64 data URI), streaming
  (`stream: true` + `response_format: pcm` → raw PCM), duration control (`${token:N}` /
  `token_count`), and text markup / style (`instructions`) / `language`.
- Generation Parameters table (per-channel `text_*` / `audio_*` defaults), Seed Reproducibility,
  Benchmarking, Benchmark Results, and Known Limitations.

**Scope**

Docs only — one new file plus one toctree line, the same way every cookbook in main is
registered. No code changes and no edits to the Supported Models table.

**Benchmark numbers**

The Benchmark Results table reports the reference numbers from PR #728 (Seed-TTS-Eval full set,
2× H100, concurrency 16, `--token-count auto`) and is explicitly labelled as a reproducibility
reference, not a CI threshold.

**Verification**

Every factual claim in the cookbook was cross-checked against the source (serve/config
resolution, generation-parameter defaults, the streaming SSE envelope, RVQ-12 / 48 kHz output,
and reference caching). The "12.5 Hz" frame-rate figure was dropped because it is not
substantiated anywhere in the repository.


gaoyang07:
Updated the PR to also improve the existing MOSS-TTS cookbook intro, add delay/local architecture diagrams, and refine the MOSS-TTS-Local benchmark section.

The Local cookbook now separates inference speed from quality metrics: Seed-TTS-Eval reports latency / RTF / throughput only, while multilingual voice-clone results report WER and SIM. Also added the measured audio sampling params and a note on a potentially better tested setting.

xinlij:
1. Have all the commands or code snippets in the cookbook been verified by running them?
2. Could we include some examples in the cookbook, similar to how Higgs does it?

docs/cookbook/moss_tts_local.md
```diff
@@ -0,0 +1,284 @@
...
+
+```bash
+curl -N -X POST http://localhost:8000/v1/audio/speech \
+  -H "Content-Type: application/json" \
+  -d '{
+    "input": "Get the trust fund to the bank early.",
+    "ref_audio": "https://huggingface.co/datasets/zhaochenyang20/seed-tts-eval-mini/resolve/main/en/prompt-wavs/common_voice_en_10119832.wav",
+    "ref_text": "We asked over twenty different people, and they all said it was his.",
+    "stream": true,
+    "response_format": "pcm"
```
SandyLuXY, line 140: From my local testing, need to add `"stream_format": "audio"` otherwise the generated `output_stream.wav` will be pure noise. I listened to the generated .wav after including `"stream_format": "audio"`, it becomes as expected.

zhaochenyang20 (approved):

PR #798, opened 2026-06-15
[Perf] MOSS-TTS-Local: CUDA-graph the streaming vocoder codec decode (default-on, bit-identical)

JiaxinD:
## Background: the streaming vocoder is the serving bottleneck

A per-request per-stage profile (built-in `RequestEventRecorder`) of the MOSS-TTS-Local colocate
pipeline locates the c8 serving bottleneck in the streaming vocoder (the MOSS-Audio-Tokenizer-v2 codec
decode): at sustained inflight~8 the vocoder stage is ~58-63% of per-request wall-clock and the
worst-scaling stage, while the AR backbone is ~24-32%. Under streaming (SSE) the codec decodes in
~one tiny per-frame step each scheduler tick, so it pays kernel-launch overhead thousands of times
per request, a launch-bound regime that a CUDA graph targets.

## What this PR does

Captures the streaming codec decode step (`_CodecStreamSession.step` -> `codec._decode_frame`) into
per-T CUDA graphs and replays them, keeping the decode fully **stateful** (per-slot causal offset/KV)
and **bit-for-bit identical** to the eager decode.

- **B fixed at the full slot width (default 16); one graph per T, no padding.** The streaming step
  always builds `[n_vq, B_full, T]` and selects active slots via `exec_mask`.
- **Stateful, made graph-capturable.** The persistent `codec.streaming()` session is retained so
  per-slot causal state lives across steps (chunk-boundary continuity). The upstream codec stored its
  streaming attention KV cache by reassigning `state.cached_keys` to a new tensor every step (its
  `data_ptr` changes), which a CUDA graph cannot follow, so replay read the capture-time cache and the
  decode was not bit-identical. An in-place cache write (stable address, value-identical to eager)
  makes it graph-capturable. **This cache-storage fix is contributed by CloudRipple (codec team) in
  #811; it is the prerequisite that unblocks the stateful graph, and is ported into this PR.**
- **Warmup-captured, sealed, never captures during serving**; the captured-T set is a data-driven
  broad-exact set (see "Graph-T set selection"), overridable via the `cuda_graph_frames` pipeline-config field.
- **Default on, fail-safe to eager** on every failure path (capture OOM, low VRAM, replay error,
  uncaptured shape, default-on opt-out via the `cuda_graph: false` pipeline-config field).

## Quality: bit-identity (stateful, real)

The CUDA-graph replay is bit-for-bit identical to the **stateful** eager codec decode. There is no
audio-quality trade to measure: the output is the same bytes.

| check | result |
|---|---|
| Streaming PCM vs eager codec decode (stateful, multi-chunk) | byte-for-byte identical (`torch.equal`, maxdelta 0.0) |
| Bit-identity test `test_vocoder_cuda_graph.py` | 14/14 passed |
| Sliding-window bit-identity (1600 frames, far beyond the bounded streaming context) | byte-for-byte identical at chunk_t 5 and 25 (maxdelta 0.0) |
| coverage | T in {1,5,25,100} x active-slots {1,8} x multi-step x multi-seed, with the persistent streaming session live |
| CPU scheduler tests `test_streaming_vocoder.py` | 26/26 passed |
| WER / audio-quality delta | N/A: output is bit-for-bit identical to eager, so quality is provably unchanged |

This is a genuine stateful check: the test exercises the codec with its persistent streaming session
live, so the per-slot causal state must advance correctly across chunked steps and the graph replay
must reproduce it exactly. (An earlier iteration that discarded the streaming context ran the codec
stateless, which made the test tautological and degraded chunk-boundary audio; restoring the stateful
session plus the in-place cache fix is what makes the test real and the audio correct.)

## Graph-T set selection (measured)

The decode emits a spread of step lengths T. We evaluated candidate capture sets against the real T
distribution from per-T serving stats (T=5 is ~38% of steps from min-coalescing; the rest are a
small-T tail plus the first-chunk/steady/cap sizes).

| capture set | graphs | step coverage | padding | measured VRAM | bit-identity |
|---|---|---|---|---|---|
| focused {join, first, steady, cap} | 3 | ~45% | none | 10.68 GB | pass |
| bucket round-up (e.g. {1,2,4,8,12,16,24}) | 7 | ~93% | 75% of steps, incl the ~38% mode | n/a | **fails** |
| power-of-two ({1,2,4,8,16,32}) | 6 | 100% | 93% of steps, up to 15 padded frames | n/a | **fails** |
| **broad-exact (measured-frequent T + cap)** | 13 | ~87% | none | 10.99 GB | **pass (maxdelta 0)** |
| full-exact (every observed T) | ~23 | 100% | none | ~11 GB | pass |

Elimination, from measured data:

- **Padding breaks bit-identity.** Decoding a real T=5 step through a padded larger-T graph diverges
  from the correct T=5 eager decode by maxdelta 0.07 at the same step (the ConvTranspose receptive
  field pulls padded frames into the boundary output) and 0.64-0.67 on following steps (the padded K/V
  pollute the sliding-window cache). Both bucket schemes pad the dominant T (38% of steps), so they
  are eliminated on correctness, not performance.
- **Focused covers too few steps** (~45%), leaving more than half on the eager path.
- **VRAM is near-flat across the exact sets**: the graph pool is dominated by the single cap (T=100)
  graph; each additional small-T graph adds ~0.03 GB. Going from 3 to 13 graphs (coverage 45% -> 87%)
  costs +0.31 GB. So broad coverage is essentially free, and full-exact's extra graphs for the rare
  tail are diminishing returns.

Broad-exact (the measured-frequent T set plus the cap) is the only candidate that is simultaneously
high-coverage, zero-padding, bit-identical, and VRAM-feasible. It is the default; the set is
config-overridable (`cuda_graph_frames`) for other chunk configs.

Caveat: the T distribution is from one serving sample, though its concentration at small T is
structural (min-coalescing).

## Performance (measured on the bit-identical stateful build)

Because the graph replay is byte-for-byte identical to the eager decode (14/14, maxdelta 0; plus the
long-audio sliding-window check above), every number below is a pure latency/throughput effect with no
audio-quality trade.

### Per-step decode: the launch-overhead collapse (B=16)

The eager streaming decode is launch-bound: its per-step cost is ~flat ~66 ms across T=4..13 (it would
grow ~6x with T if it were compute-bound), so the cost is per-decode kernel-launch overhead, not
compute. The CUDA graph replaces that launch storm with one replay (eager = unpatched
eager (`cuda_graph: false`) path; median of 3 blocks x 200 iters):

| step length T | eager ms/step | graph ms/step | speedup |
|---|---|---|---|
| 4 | 66.3 | 30.1 | 2.20x |
| 5 (dominant, ~38% of steps) | 65.8 | 30.7 | 2.14x |
| 8 | 65.6 | 34.0 | 1.93x |
| 13 | 65.4 | 40.4 | 1.62x |
| 25 | 74.8 | 58.3 | 1.28x |
| 100 (offline cap) | 222.9 | 215.3 | 1.04x |

The win is largest exactly where streaming lives (small T); at the T=100 offline cap compute dominates
and the graph adds little (it never hurts).

### End-to-end streaming (seed-tts-eval en, SSE, paired eager vs graph)

The reliable, reproducible effect is the vocoder stage speedup and the per-request latency it buys.
The throughput gain follows from that latency cut (Little's law), so its magnitude depends on the host
sustaining concurrency and is host-load dependent; we report it as a range.

| concurrency | vocoder span (stable) | e2e p50 latency (stable) | throughput (host-dependent) | note |
|---|---|---|---|---|
| c1 | **-44%** | -6% | ~0 to +8% | AR-bound: the vocoder is only ~22% of c1 e2e |
| c8 | **-40%** | **-22%** | **+16% to +29%** | sweet spot (vocoder ~75% of e2e); +29% on a clean card, ~+16% on a busy one |
| c16 | **-22%** | -10% to -18% | +13% to +22% | offline lane brings larger-T steps |

The vocoder span (-40% at c8) and e2e latency (-22% at c8) reproduce across measurement windows and
across cards, and are the signal to trust. The throughput delta is more variable: it is the
Little's-law expression of the same latency cut, so a clean dedicated card realizes ~+29% while a busy
card realizes ~+16% (same latency cut, lower effective concurrency). At c1 the vocoder is not the
bottleneck, so the e2e gain is small.

Dependency on VRAM headroom (important): the graph captures a few GB of CUDA graphs at warmup and needs
that headroom to seal. On a VRAM-tight card (e.g. a busy shared GPU) the capture is skipped and the
decode safely degrades to eager: no speedup, but no harm. So a run that shows "no improvement" is
usually a card without the headroom to seal the graphs (the captured count is logged as "CUDA graphs
sealed: N T"), not a regression. The default capture set is the small-T streaming shapes only; the
T=max_step_frames cap is excluded to keep the warmup VRAM peak low.

Honest caveats:

- Rounds are display-grade (a few paired rounds), not a CI-bounded estimate; absolute req/s is
  box-sensitive, so the paired eager -> graph delta and the latency/span are the signals to trust.
- The AR (`tts_engine`) stage rises ~+9-16% under graph (pipeline rebalancing as throughput climbs);
  the e2e gain is net of it. Throughput and e2e latency are one fact (Little's law at fixed inflight).
- c24 is past the useful operating point (RTF > 1.7, overload); throughput is noise-dominated there,
  so we do not claim a c24 number.

## Credit and relation to existing work

- **In-place attention-cache fix: CloudRipple (codec team), #811.** This is the prerequisite that
  makes the stateful streaming decode CUDA-graph-capturable; it is ported here and credited in the
  commit and code. Tracking: #801.
- The per-shape warmup-captured runner is adapted from the Higgs vocoder graph (credited in the code),
  specialized to MOSS's fixed batch width + `exec_mask` gating.

## Review fixes addressed

Replay-failure isolation now wraps the whole graphed step through output materialization (a CUDA-graph
error can surface asynchronously on the D2H, not only inside `decode_step`); `decode_step` returns the
static output buffers directly (no per-replay clone) and the scheduler consumes them immediately;
`has_cuda_graph_runner()` is true only when graphs were captured and a zero-graph runner is dropped;
the scheduler owns the capture shape range; the non-streaming offline-lane routing is documented; plus
the import/comment/docstring nits.

## Configuration and lifecycle

- **Config-scoped, not env vars.** The CUDA graph is controlled by model-scoped pipeline-config
  fields, injected into the vocoder factory: `cuda_graph` (bool, default true), `cuda_graph_frames`
  (list, default the built-in broad-exact set), `cuda_graph_min_free_gb` (float, default 3). Default
  on, fail-safe to eager; no CLI-flag or env-var surface.
- **Captured in the factory, before the stage is ready.** Capture runs synchronously while the stage
  process is built, so the serving loop never races a half-captured graph. Same-process guarantee
  (colocate; a split-topology codec process captures before its own ready). Capture is bound to the
  codec's own device, so it is correct on a non-default GPU.
- **Recapture after non-streaming traffic.** When non-streaming traffic closes the idle graphed
  startup session, the next streaming session re-captures (attempted at most once per session, so a
  low-VRAM skip does not re-probe per step). That first post-non-streaming streaming request pays a
  one-time graph-warmup latency (the recapture runs synchronously on the scheduler loop), fail-safe to
  eager on low VRAM; streaming-only traffic uses the factory session and never incurs it.
- **`decode_step` cleanup.** The replay path dropped its unused `codes_lengths` argument; the captured
  graph decodes full-T for all slots and `exec_mask` gates valid outputs (the eager fallback still
  uses lengths). `codes_step` is unchanged.

## Safety / rollout

- **Default on**, opt out with the `cuda_graph: false` pipeline-config field (model-scoped, no env var or CLI flag).
- **Fail-safe to eager** on every path (flag off, low VRAM, capture OOM, replay error, uncaptured
  shape), each covered by a test.
- Touches only the MOSS-vocoder files; the codec-side cache fix lands upstream (#811) and is applied
  here as a runtime patch on the codec's decoder attention modules.


sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -73,17 +86,53 @@ class _CodecStreamSession:
...
         self._stream_slots = int(stream_slots)
         self._offline_slots = int(offline_slots)
         self._batch_size = self._stream_slots + self._offline_slots
+        self._n_vq = int(n_vq)
         self._device = next(codec.parameters()).device
         self._free_stream_slots = list(range(self._stream_slots))
-        self._exit_stack = contextlib.ExitStack()
         self._closed = False
+        self._cg_runner: Any | None = None
+        # CUDA-graph step accounting (only populated when a runner is attached): T -> step count
```
Hayden727: trim this comment.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -73,17 +86,53 @@ class _CodecStreamSession:
...
+        # served by a captured graph vs fallen back to eager. Lets the perf run report capture
+        # hit-rate; host-side ints only (no GPU sync).
+        self._cg_graph_t: Counter = Counter()
+        self._cg_eager_t: Counter = Counter()
+        self._cg_total_steps = 0
+        # Enter streaming WITHOUT retaining the returned ExitStack. Retaining it (the natural
```
Hayden727: trim.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
Hayden727: If comment, format as "Note:(Jiaxin Deng) xxx"

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -266,6 +349,20 @@ def __init__(
 
     def start(self) -> None:
         try:
+            # Capture the codec-decode CUDA graphs ONCE here, before the serving loop: the codec is
```
Hayden727: same.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -266,6 +349,20 @@ def __init__(
...
+                try:
+                    captured = self.warmup_cuda_graph()
+                    logger.info(
+                        "MOSS vocoder CUDA graphs captured at startup: T=%s", captured
+                    )
+                except Exception:
+                    logger.exception(
+                        "MOSS vocoder CUDA-graph warmup failed; serving with eager vocoder"
+                    )
             super().start()
```
AkazaAkane, line 375: I checked that StreamingSimpleScheduler.start() is currently blocking, so this should work today: the warmed _session survives while super().start() runs the message loop, and _close_streaming_session() only runs after stop. Still, this lifecycle is subtle because the CUDA graphs are stored on _CodecStreamSession, not the scheduler. Can we add a short comment or assertion near start() saying this warmup relies on super().start() being blocking, and maybe add a small smoke check/log that the serving session has _cg_runner after startup? That would prevent a future scheduler refactor from silently discarding the warmed graphs.

JiaxinD: Good catch 🙏 You're right, the lifecycle is sneaky: the graphs live on _CodecStreamSession so it all hinges on start() being blocking. Added an assertion + comment near start() spelling that out, plus a startup log that warns if _cg_runner is missing after warmup, so a future scheduler refactor can't silently drop the warmed graphs. Pushed!

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -145,12 +211,41 @@ def step(self, slot_codes: dict[int, torch.Tensor]) -> dict[int, torch.Tensor]:
...
+                try:
+                    graphed = self._cg_runner.decode_step(
+                        codes_step, codes_lengths, exec_mask
+                    )
+                except Exception:
+                    # Note: (Jiaxin Deng) a replay-path failure leaves THIS step's per-slot streaming
+                    # state indeterminate (replay advances it; CUDA-graph errors surface async), so it
+                    # can't be bit-safely eager-retried. Disable the runner so every FUTURE step degrades
+                    # to eager (bit-identical from live state) -- one transient failure must not cascade.
+                    # This step's participants abort upstream (bit-safe: never emit a non-identical output).
```
Ratish1: nit: trim this comment

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -266,6 +361,34 @@ def __init__(
...
+                    )
+                except Exception:
+                    logger.exception(
+                        "MOSS vocoder CUDA-graph warmup failed; serving with eager vocoder"
+                    )
+                # Note: (Jiaxin Deng) the warmed graphs live on self._session (its runner); the serving
+                # loop reuses that SAME session via _ensure_session. This hinges on super().start() below
+                # running the BLOCKING serving loop on this object -- if a refactor made it non-blocking, or
+                # swapped/recreated the session, the warmed graphs would be silently dropped. The assert
+                # pins the precondition; the warning guards against a serving session without the runner.
```
Ratish1: same

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
```diff
@@ -0,0 +1,223 @@
...
+     ``torch.where`` reads the live mask and advances ONLY active slots' state -- inactive (paused/free)
+     slots are preserved.
+   - Warmup + capture advance state, so the caller MUST reset all slots after ``warmup()`` (see
+     ``_CodecStreamSession``); serving then advances from clean state.
+
+Capture happens ONCE at warmup (GPU quiescent) then the runner is sealed -- never captures during
+serving (live capture corrupts the co-located AR stage's graph mempool). Capture is best-effort: any
+shape that fails to capture (e.g. an unexpected ``.item()`` host sync) is dropped and falls back to
+eager, so correctness never depends on capture succeeding. Bit-identity vs eager is gated by
+``tests/unit_test/moss_tts_local/test_vocoder_cuda_graph.py``.
```
Ratish1: same trim this and dont use PR or issue references here

tests/unit_test/moss_tts_local/test_vocoder_cuda_graph.py
```diff
@@ -0,0 +1,281 @@
+# SPDX-License-Identifier: Apache-2.0
+"""命门 (lifeline) bit-identity gate for the MOSS streaming-vocoder codec-decode CUDA graph.
```
Ratish1: avoid Chinese in the code docstrings

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -73,17 +88,51 @@ class _CodecStreamSession:
...
+        # Per-T graph-vs-eager step counts for capture-hit-rate reporting (host-side, no GPU sync).
+        self._cg_graph_t: Counter = Counter()
+        self._cg_eager_t: Counter = Counter()
+        self._cg_total_steps = 0
+        # Note: (Jiaxin Deng) enter streaming but DON'T retain the ExitStack -- holding it
+        # deterministically breaks multi-step CUDA-graph replay (step 1+ reads stale state, ~0.4 PCM
+        # error). cleanup is explicit in close() via _stop_streaming.
         with torch.no_grad():
-            self._exit_stack.enter_context(codec.streaming(self._batch_size))
+            codec.streaming(self._batch_size).__enter__()
```
Ratish1: This changes the streaming session ownership contract from a retained context manager to a discarded `__enter__()` call, and shutdown now depends on private `_stop_streaming()` at close.

That is not equivalent to the previous `ExitStack.enter_context(codec.streaming(...))` behavior. A normal context manager is allowed to clean itself up when the context-manager object is dropped; the existing CPU fake in `test_streaming_vocoder.py` has exactly that shape and has no `_stop_streaming()` method. More importantly, the constructor compatibility check still only requires `streaming`, `_set_streaming_exec_mask`, and `_decode_frame`, even though close now requires `_stop_streaming`.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -266,6 +361,34 @@ def __init__(
 
     def start(self) -> None:
         try:
+            # Note: (Jiaxin Deng) capture graphs ONCE here -- codec loaded, GPU quiescent, before the
+            # serving loop (never capture mid-serve). default on, opt out with MOSS_VOCODER_CUDA_GRAPH=0; best-effort -> eager.
+            if _vocoder_cuda_graph_enabled():
+                captured: list[int] = []
+                try:
+                    captured = self.warmup_cuda_graph()
```
Ratish1: This startup warmup creates `_session` before the serving loop receives any request, which changes the non-streaming path even before the first stream arrives.

The file-level contract still says non-streaming requests use `processor.decode_audio_codes` until a streaming request opens the persistent session. With this default-on warmup, `_decode_codes_rows()` now sees `_session is not None` from process start and routes non-streaming requests through the session offline lane instead. That is a serving semantic change, not just a streaming optimization.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -145,12 +211,41 @@ def step(self, slot_codes: dict[int, torch.Tensor]) -> dict[int, torch.Tensor]:
...
+            else:
+                self._cg_eager_t[step_t] += 1
+            self._cg_total_steps += 1
+            if self._cg_total_steps % 2000 == 0:
+                self._log_cg_stats()
         # One batched D2H per step, active slots only.
         slots = list(slot_codes)
-        audio_cpu = result.audio[slots].detach().to("cpu", torch.float32)
-        lengths_cpu = result.audio_lengths[slots].detach().to("cpu")
+        audio_cpu = audio[slots].detach().to("cpu", torch.float32)
```
Ratish1: Replay failure isolation needs to include post-replay output materialization, not only exceptions raised inside `decode_step()`.

CUDA graph errors can surface asynchronously. If `graph.replay()` queues work and the failure appears on this D2H copy, `_pump_streams()` will abort the participants, but `_cg_runner` will remain installed because the disable logic only wraps `decode_step()`. The next step can then try the same broken runner again, which violates the default-on fallback claim.

we can track whether this step used graph replay and wrap the whole graphed step through output materialization. If any exception occurs before returning outputs for a graphed step, set `_cg_runner = None`, log once, and re-raise so participants abort.

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
```diff
@@ -0,0 +1,223 @@
...
+        entry = self._graphs.get(int(t))
+        if entry is None:
+            return None
+        graph, static_codes, static_lengths, static_audio, static_audio_lengths = entry
+        # Replicate eager inputs exactly (codes, lengths, exec_mask) so replay is bit-for-bit identical.
+        self._codec._set_streaming_exec_mask(exec_mask)
+        static_codes.copy_(codes_step)
+        static_lengths.copy_(codes_lengths)
+        graph.replay()
+        return static_audio.clone(), static_audio_lengths.clone()
```
Ratish1: This clones the full static graph output buffer on every graph hit, even though the caller immediately slices active slots and copies only those slots to CPU.

For a launch-bound path, adding a full `B_full` device allocation and device-to-device copy per replay is the wrong hot-path tradeoff. It also adds allocator churn in a long-running streaming service. The existing MOSS AR graph code returns static output slices and makes the caller responsible for copying anything it keeps before the next replay.

Can we return the static buffers directly and consume active slots immediately in `_CodecStreamSession.step()`, or pass active slots into the runner and copy only the needed outputs.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -73,17 +88,51 @@ class _CodecStreamSession:
...
+        if self._cg_runner is None:
+            self._cg_runner = MossVocoderCudaGraphRunner(
+                self._codec, batch_size=self._batch_size, n_vq=self._n_vq
+            )
+        self._cg_runner.warmup(frames)
+        self._reset_slots(list(range(self._batch_size)))
+        return self._cg_runner.captured_frames()
+
+    def has_cuda_graph_runner(self) -> bool:
+        return self._cg_runner is not None
```
Ratish1: `has_cuda_graph_runner()` should not report success when the runner captured zero graphs.

Low VRAM, capture failure, or an empty capture set can leave `_cg_runner` installed with `captured_frames() == []`. In that state the serving path still pays an extra `decode_step()` call on every step only to fall back to eager, and lifecycle warnings cannot distinguish "runner object exists" from "graphs are usable".

Please clear `_cg_runner` after warmup when no frames were captured, or make this return `bool(self._cg_runner and self._cg_runner.captured_frames())`.

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
```diff
@@ -0,0 +1,223 @@
...
+class MossVocoderCudaGraphRunner:
+    """Warmup-captured, sealed replay of exact-T CUDA graphs for the MOSS codec decode (B fixed)."""
+
+    def __init__(
+        self,
+        codec,
+        *,
+        batch_size: int,
+        n_vq: int,
+        max_frames: int = 128,
```
Ratish1, line 94: The runner has an independent `max_frames=128` limit while the scheduler computes capture frames from `self._max_step_frames`.

Today the default `max_step_frames` is 100, so the mismatch is latent. But direct construction or future factory exposure can produce scheduler-approved capture sizes above 128 that the runner silently skips. Shape policy should have one owner.

we can maybe pass `max_frames=self._max_step_frames` when constructing `MossVocoderCudaGraphRunner`, or remove the runner's separate default limit and let the scheduler own the capture range.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -15,13 +15,28 @@
...
 import logging
+import os
+from collections import Counter
 from dataclasses import dataclass, field
 from typing import Any, Mapping
 
 import torch
 
+
+def _vocoder_cuda_graph_enabled() -> bool:
```
Ratish1: This helper is inserted between third-party imports and local imports.

Please keep imports grouped first, then module constants/helpers. Move `_vocoder_cuda_graph_enabled()` below the local imports and `logger`/constant setup.

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
```diff
@@ -0,0 +1,223 @@
...
+        self,
+        codes_step: torch.Tensor,
+        codes_lengths: torch.Tensor,
+        exec_mask: torch.Tensor,
+    ) -> tuple[torch.Tensor, torch.Tensor] | None:
+        """Replay the captured graph for ``[n_vq, B_full, T]`` codes, else return None (eager).
+
+        Sets the live exec_mask (host-side, into the codec's fixed mask buffers) so the captured
+        torch.where gating advances only active slots, copies codes into the static input, replays,
+        and returns cloned (audio, audio_lengths) static outputs. ``codes_lengths`` is intentionally
```
Ratish1: the docstring says `codes_lengths` is intentionally not copied, but the implementation copies it on line 221.

Lets update the docstring to match the implementation. If copying lengths is required for bit identity, say that directly. If it is not required, remove the copy and keep the invariant true.

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
```diff
@@ -0,0 +1,223 @@
...
+            # a tight box. Checked per-T because each capture allocates; free only drops through the loop.
+            if not self._enough_free_vram():
+                free, _ = torch.cuda.mem_get_info(self._device)
+                logger.warning(
+                    "MOSS vocoder CG: free VRAM %.1fGB < %.1fGB headroom; skipping T=%d+ (eager)",
+                    free / 1024**3,
+                    self._min_free_bytes / 1024**3,
+                    t,
+                )
+                break
```
Ccyest: nit: warmup() calls` self._enough_free_vram()` (which queries` torch.cuda.mem_get_info` once), and when it returns False, the warning block queries` mem_get_info `a second time for the log message. If memory changes between the two queries, the logged free might be ≥ the threshold, producing a confusing log.

Fix:
Have `_enough_free_vram() `return the free value alongside the boolean, or inline the single query in warmup().

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -15,13 +15,28 @@
...
+    """True unless the streaming-vocoder codec-decode CUDA graph is opted OUT
+    (``MOSS_VOCODER_CUDA_GRAPH=0``). Default ON; every failure path (capture OOM, low VRAM,
+    replay error, uncaptured shape) degrades to the eager decode, so default-on never crashes.
+    """
+    return os.environ.get("MOSS_VOCODER_CUDA_GRAPH", "1").strip().lower() not in (
+        "",
+        "0",
+        "false",
+        "no",
+    )
```
Ccyest: “off” is a common boolean used in  K8s/Docker env by dev ops. I would suggest add this too just in case user set` MOSS_VOCODER_CUDA_GRAPH=off` but it is not actually turned off.

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
```diff
@@ -0,0 +1,223 @@
...
+        return 1 <= t <= self._max_frames
+
+    def _enough_free_vram(self) -> bool:
+        free, _ = torch.cuda.mem_get_info(self._device)
+        return free >= self._min_free_bytes
+
+    @torch.no_grad()
+    def _reset_state(self) -> None:
+        """Reset every streaming module's per-slot state to offset 0 (in-place)."""
+        reset_mask = torch.ones(self._batch_size, dtype=torch.bool, device=self._device)
```
gaoyang07: Warmup reset currently creates a fresh all-true reset mask and then calls full `state.reset(...)` on every streaming module. After capture we only need to restore the logical streaming state before serving; stale KV values do not need to be zeroed if offsets/positions/end offsets are reset so they remain invisible.

Can this use the same reset fast path as slot release, e.g. `reset_exec_mask=False` and `reset_cache_values=False`, or a codec-level reset API that avoids clearing K/V rows?

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Added both: start() asserts the warmed _session survives into the blocking super().start(), and warns via has_cuda_graph_runner() if the serving session ends up without a runner.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Done, trimmed to 2 lines.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Done, trimmed to 2 lines.

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
JiaxinD: Done, trimmed and removed the issue refs from this docstring. The only remaining one is the CloudRipple #811 credit on the cache-patch helper, kept as contributor attribution.

tests/unit_test/moss_tts_local/test_vocoder_cuda_graph.py
JiaxinD: Done, all test docstrings are now English; the Chinese wording is removed.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Resolved by the merge: the retained ExitStack is restored (enter_context at init, exit_stack.close() at close), so the original context-manager ownership contract holds. Bit-identity now comes from CloudRipple's in-place cache patch (#811), not from dropping the streaming state.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Acknowledged, it is intentional: default-on warmup opens the session at startup, so non-streaming routes through the session offline lane, which is output-equivalent to decode_audio_codes (now documented in the module docstring and decode_offline). Opt out with MOSS_VOCODER_CUDA_GRAPH=0 to keep the processor path.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Done. The output D2H materialization is now inside the same replay guard, so an async graph error surfacing on the copy also disables the runner, not just exceptions raised inside decode_step().

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
JiaxinD: Done. decode_step returns the static buffers directly (no per-replay clone); the caller slices active slots to CPU immediately, before the next replay.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Done. has_cuda_graph_runner() now returns bool(runner and captured_frames()), and warmup drops a zero-graph runner so serving never probes decode_step only to fall back.

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
JiaxinD: Done. The scheduler now owns the shape range: it constructs the runner with max_frames = the largest T it requests, so there is a single owner.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Done. Moved _vocoder_cuda_graph_enabled() below the imports and the logger/constant setup.

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
JiaxinD: Done. Removed the per-replay codes_lengths copy (T is fixed per captured graph, so lengths are constant) and updated the docstring to match.

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
JiaxinD: Done. _enough_free_vram() now returns (enough, free) and the warning logs that same free value, so there is a single mem_get_info query.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Good call. Added 'off' to the opt-out set (now ''/0/false/no/off).

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
JiaxinD: Pre-allocated the reset mask (no per-call rebuild). Kept full state.reset since it is warmup-only, not the hot path, and a clean pre-capture state is safest for bit-identity; can switch to the reset_cache_values=False fast path if you prefer.

JiaxinD:
A same-machine paired repro script (eager vs graph A/B on one card). Read the **eager -> graph delta**, not absolute req/s (throughput is host-load sensitive). It prints whether the graph **actually sealed**: if the card lacks VRAM headroom at startup the capture is skipped and the run silently falls back to eager, so the graph columns would be invalid in that case (the script flags this as `GRAPH NOT SEALED` and shows free VRAM).

```bash
#!/usr/bin/env bash
# ============================================================================
# repro_vocoder_cg.sh -- neutral SAME-MACHINE PAIRED A/B for the MOSS streaming
#                        vocoder CUDA graph. Prints results as-is; draws no
#                        conclusion for you. Run it, read the diagnostics.
#
# WHAT IT DOES
#   On ONE card it starts a single colocated MOSS server twice on the SAME
#   config, flipping only MOSS_VOCODER_CUDA_GRAPH:
#       eager = MOSS_VOCODER_CUDA_GRAPH=0
#       graph = default (on)
#   drives seed-tts-eval streaming at c8 and c16, prints the PAIRED delta
#   (throughput / latency / rtf), plus per-arm VRAM diagnostics.
#
# HOW TO READ IT  (please read before concluding anything)
#   * Same-machine paired: compare the DELTA (graph vs eager), not absolute
#     req/s. On a busy host throughput under-reads for BOTH arms; latency_med
#     and rtf_med are per-request and far more robust -- trust those if
#     throughput looks noisy.
#   * The graph arm is ONLY valid if you see "GRAPH SEALED OK (13 T)". If it
#     says "GRAPH NOT SEALED" or "PARTIALLY SEALED", the captured-graph set was
#     skipped (almost always: free VRAM below the ~3 GB headroom our guard needs
#     at warmup, e.g. a neighbor task on that card), so the "graph" run silently
#     fell back to eager and the graph columns are MEANINGLESS. The script prints
#     free VRAM at startup so you can see this. This is a KNOWN behavior: the
#     graph needs VRAM headroom; under pressure it degrades to eager (safe, but
#     no speedup).
#   * Two harness details (main drifted): client uses --stream (NOT
#     --stream-format); server needs --allowed-local-media-path /tmp (else
#     voice-clone ref_audio is rejected HTTP 400 and every request fails).
#   * Single colocated server (NOT the 2-replica router CI uses).
#   * Expected on a DEDICATED card with the graph sealed: c8 ~+29% throughput /
#     ~-25% latency, c16 ~+20% / ~-18%. If the card is VRAM-tight and the graph
#     can't seal, expect ~0 (eager fallback) -- fix the VRAM, don't conclude
#     "no speedup".
#
# USAGE: edit CONFIG, then: chmod +x repro_vocoder_cg.sh && ./repro_vocoder_cg.sh
# ============================================================================
set -u

# ===== CONFIG (edit these) ==================================================
REPO=${REPO:-/path/to/sglang-omni}                # repo root of this branch (pr/moss-vocoder-cuda-graph)
YAML=${YAML:-/tmp/moss_colocate_repro.yaml}       # auto-written below (single-GPU colocate); override to use your own
CARD=${CARD:-0}                                   # GPU index (pick one with plenty of free VRAM)
PORT=${PORT:-8790}
CONCS=${CONCS:-"8 16"}
MODEL=${MODEL:-OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5}
META=${META:-zhaochenyang20/seed-tts-eval-arrow}
PYBIN=${PYBIN:-python}
# ===========================================================================

export PYTHONPATH="$REPO${PYTHONPATH:+:$PYTHONPATH}"   # test THIS working tree + resolve `benchmarks`

# Single-GPU colocated config (AR + vocoder on one card) -- this is the topology the numbers were
# measured on. Auto-written so you don't have to hunt for it; the repo example is NON-colocated.
if [ "$YAML" = /tmp/moss_colocate_repro.yaml ]; then
  cat > "$YAML" <<'YAMLEOF'
config_cls: MossTTSLocalColocatedPipelineConfig
model_path: OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5
relay_backend: shm
YAMLEOF
fi

declare -A THR LAT RTF PEAK
declare -A SEALMSG
stop() { pkill -f "sglang_omni.cli serve" 2>/dev/null; sleep 3; }
trap stop EXIT
gpu_free() { nvidia-smi --query-gpu=memory.free --format=csv,noheader,nounits -i "$CARD" | tr -d ' '; }
gpu_used() { nvidia-smi --query-gpu=memory.used --format=csv,noheader,nounits -i "$CARD" | tr -d ' '; }

run_flag() {  # $1=flag(0/1) $2=label
  local flag=$1 label=$2 log="/tmp/repro_srv_${label}.log"
  stop
  echo "==> [$label] free VRAM on card $CARD before start: $(gpu_free) MiB  (need >~3 GB headroom for graph)"
  MOSS_FUSED_SAMPLER=0 MOSS_VOCODER_CUDA_GRAPH="$flag" CUDA_VISIBLE_DEVICES="$CARD" \
    "$PYBIN" -m sglang_omni.cli serve --config "$YAML" --port "$PORT" \
    --decode-mode sync --allowed-local-media-path /tmp > "$log" 2>&1 &
  local up=0 i
  for i in $(seq 1 240); do
    [ "$(curl -s -o /dev/null -w '%{http_code}' "localhost:$PORT/health" 2>/dev/null)" = 200 ] && { up=1; break; }
    grep -qiE "Traceback|out of memory|Address already in use" "$log" 2>/dev/null && break
    sleep 2
  done
  if [ "$up" != 1 ]; then echo "    !! server failed to come up (OOM? port busy?); tail of $log:"; tail -15 "$log"; return 1; fi
  sleep 5
  if [ "$flag" = 1 ]; then
    local seal ncap guard
    seal=$(grep "graphs sealed" "$log" | tail -1)
    ncap=$(printf '%s' "$seal" | grep -oE "sealed: [0-9]+ T" | grep -oE "[0-9]+")
    guard=$(grep -iE "free VRAM .*headroom; skipping" "$log" | tail -1)
    if [ "${ncap:-0}" = 13 ]; then
      SEALMSG[graph]="GRAPH SEALED OK (13 T captured)"
      echo "    GRAPH SEALED ✓  (13 T captured) -- graph is genuinely active"
    elif [ "${ncap:-0}" -gt 0 ] 2>/dev/null; then
      SEALMSG[graph]="GRAPH PARTIALLY SEALED (${ncap}/13 T) -- some captures skipped"
      echo "    GRAPH PARTIALLY SEALED ⚠  (${ncap}/13 T) -- some shapes skipped (VRAM?), graph delta understated"
    else
      SEALMSG[graph]="GRAPH NOT SEALED -- EAGER FALLBACK, graph results INVALID"
      echo "    GRAPH NOT SEALED ✗  -- running EAGER FALLBACK; the graph columns below are MEANINGLESS"
    fi
    [ -n "$guard" ] && echo "    VRAM GUARD TRIGGERED -> ${guard#*] }"
  fi
  local C n thr lat rtf dlog
  for C in $CONCS; do
    n=$((C * 5)); dlog="/tmp/repro_drive_${label}_c${C}.log"
    "$PYBIN" -m benchmarks.eval.benchmark_tts_seedtts --use-existing-server --generate-only \
      --stream --lang en --model "$MODEL" --meta "$META" --host localhost --port "$PORT" \
      --concurrency "$C" --max-samples "$n" --warmup 0 > "$dlog" 2>&1
    thr=$(grep "Throughput (req/s)" "$dlog" | tail -1 | grep -oE '[0-9.]+$')
    lat=$(grep "Latency median (s)" "$dlog" | tail -1 | grep -oE '[0-9.]+$')
    rtf=$(grep "RTF median"          "$dlog" | tail -1 | grep -oE '[0-9.]+$')
    THR[${label}_c${C}]=${thr:-NA}; LAT[${label}_c${C}]=${lat:-NA}; RTF[${label}_c${C}]=${rtf:-NA}
    PEAK[${label}_c${C}]=$(gpu_used)
    echo "    [$label c$C] thr=${thr:-NA} req/s  lat_med=${lat:-NA}s  rtf_med=${rtf:-NA}  vram=${PEAK[${label}_c${C}]} MiB"
  done
  stop
}

echo "############################################################################"
echo "# MOSS vocoder CUDA-graph PAIRED repro (card $CARD). Compare DELTA, not absolute."
echo "# Graph valid ONLY if 'GRAPH SEALED ✓ (13 T)'. Expected (dedicated card):"
echo "#   c8 ~+29% thr / ~-25% lat ; c16 ~+20% / ~-18%.  VRAM-tight -> graph degrades to eager."
echo "############################################################################"
run_flag 0 eager || { echo "eager run failed"; exit 1; }
run_flag 1 graph || { echo "graph run failed"; exit 1; }

pct() { awk -v e="$1" -v g="$2" -v inv="$3" 'BEGIN{ if(e=="NA"||g=="NA"||e+0<=0){print"NA";exit} v=(inv?(1-g/e):(g/e-1))*100; printf "%+.1f%%", v }'; }
echo
echo "===================== PAIRED RESULT (same machine, card $CARD) ====================="
echo "graph capture: ${SEALMSG[graph]:-<none>}"
printf "%-4s | %-24s | %-24s | %-22s\n" "conc" "throughput e->g (delta)" "latency_med e->g (delta)" "rtf_med e->g (delta)"
for C in $CONCS; do
  printf "%-4s | %-24s | %-24s | %-22s\n" "c$C" \
    "${THR[eager_c$C]:-NA}->${THR[graph_c$C]:-NA} ($(pct "${THR[eager_c$C]:-NA}" "${THR[graph_c$C]:-NA}" 0))" \
    "${LAT[eager_c$C]:-NA}->${LAT[graph_c$C]:-NA} ($(pct "${LAT[eager_c$C]:-NA}" "${LAT[graph_c$C]:-NA}" 1))" \
    "${RTF[eager_c$C]:-NA}->${RTF[graph_c$C]:-NA} ($(pct "${RTF[eager_c$C]:-NA}" "${RTF[graph_c$C]:-NA}" 1))"
done
echo "VRAM per arm (incl graph pool): eager c8=${PEAK[eager_c8]:-NA} graph c8=${PEAK[graph_c8]:-NA} MiB  (graph holds ~+3-4 GB of captured graphs)"
echo "==================================================================================="
echo "NOTE: throughput is the most host-load-sensitive metric; on a busy card it under-reads"
echo "      for BOTH arms -> read the latency_med / rtf_med delta instead. And if 'graph capture'"
echo "      above is NOT 'SEALED ✓', the graph never engaged (VRAM headroom) -> graph columns invalid."
```

Edit `REPO` (your branch checkout) and `CARD` at the top; the colocate config is auto-written. Expected on a dedicated card with the graph sealed: **c8 ~+29% throughput, c16 ~+20%**. The two easy-to-miss bits are baked in: client uses `--stream` (not `--stream-format`), server needs `--allowed-local-media-path /tmp`.

luojiaxuan:
CI repro notes for the MOSS TTS streaming failure:

- Reproduced locally on the exact CI merge commit `42d1072cdae15e2f457f7c8166c93400ba8514e4` from run `27651319825`, using the same `hongccc/sglang-omni:dev` image digest and `TTS_CI_MODEL=moss`.
- Command: `pytest tests/test_model/test_tts_ci.py -v -s -x --concurrency 16 --tts-stage tts-stage-2-stream`.
- Local single-attempt result matches the CI failure mode: `Completed requests: 1086`, `Failed requests: 2`, router delta `successful=[549, 537]`, `routed=[551, 538]`.
- Local repro used two visible H100 80GB GPUs, so this is not only an H20 runner artifact.
- The failed requests are HTTP 500s from MOSS reference-audio preprocessing: `reference encode failed ... CUDA error: CUBLAS_STATUS_ALLOC_FAILED when calling cublasCreate(handle)`.
- CI attempts show the same pattern: a small number of failed routed requests, then the `All routed requests should succeed` assertion failure.
- This does not look like a one-off GitHub Actions cancellation; it looks like a reproducible VRAM/headroom or startup-time resource issue around reference encoder initialization.
- Suggested next check: rerun the same stage with `MOSS_VOCODER_CUDA_GRAPH=0` to isolate whether the new vocoder CUDA graph memory pressure is pushing reference encoding over the edge.

luojiaxuan:
Follow-up with the latest PR merge commit:

- Retested the current PR merge commit `639d3f36105191406b6cc775503c6b838cdba054` (`3c38b0d24f33b2b7aca58f9efb16a645c59e4d08` merged into `main`) with the same MOSS TTS streaming CI stage.
- Command: `pytest tests/test_model/test_tts_ci.py -v -s -x --concurrency 16 --tts-stage tts-stage-2-stream` with `TTS_CI_MODEL=moss`.
- Result: passed locally: `2 passed, 5 deselected in 405.74s`.
- TTS streaming result: `Completed requests: 1088`, `Failed requests: 0`; router delta `successful=routed` for both workers (`551/551`, `538/538`).
- WER path also passed: `1088/1088` evaluated, `0` skipped, corpus WER `2.04%`.
- So the earlier repro was for the old CI merge commit `42d1072cdae15e2f457f7c8166c93400ba8514e4`; the latest PR version appears to fix that failure mode.

Relative performance vs `main` / eager baseline:

- The linked tracking issue is #801. It has the early motivation and blocker notes, including the original per-step target (`T=5, B=8`: eager `53.3 ms` vs graph `18.5 ms`, about `2.87x`).
- The current PR body has the more relevant final baseline: paired eager/main-path (`MOSS_VOCODER_CUDA_GRAPH=0`) vs graph on the stateful bit-identical build.
- Per-step B=16 decode from the PR body: `T=5` (dominant shape) `65.8 ms -> 30.7 ms` (`2.14x`); small streaming shapes range from `1.62x` to `2.20x`; the offline cap `T=100` is compute-bound and only `1.04x`.
- End-to-end streaming vs main/eager from the PR body: at `c8`, vocoder span `-40%`, p50 latency `-22%`, throughput `+16% to +29%`; at `c16`, vocoder span `-22%`, p50 latency `-10% to -18%`, throughput `+13% to +22%`.
- My local run above is a CI-topology stability/absolute-perf check (`4.551 req/s`, mean latency `3.490s`, mean RTF `0.8384`), not a paired local `main` baseline run.

Minor note: I still saw one non-fatal startup warning around stream capture fallback (`operation not permitted when stream is capturing`), but there were no HTTP 500s or `CUBLAS_STATUS_ALLOC_FAILED` failures in this run.

zhaochenyang20:
https://github.com/CloudRipple will co-author

luojiaxuan:
Update on the non-streaming regression:

- Root cause: startup vocoder CUDA-graph warmup creates a persistent `_session`, and `_decode_codes_rows()` treated any live session as a reason to route non-streaming batches through `_session.decode_offline()`. That changes pure non-streaming traffic from the main baseline `processor.decode_audio_codes` path and matches the CI symptoms: many HTTP 500 / CUDA OOMs plus lower throughput.
- Fix: keep track of whether the session has actually served streaming work. If the session is only the idle startup/warmup session, close it before non-streaming decode and keep the original processor path. Once real streaming has used the session, the existing offline-lane behavior is preserved.
- Patch: I could not push directly to the PR fork, so I pushed a cherry-pickable branch: `luojiaxuan:codex/pr798-nonstream-fix`, commit `c7a77f1`.
- Unit check: `python -m pytest tests/unit_test/moss_tts_local/test_streaming_vocoder.py -q -p no:cacheprovider` -> 19 passed.
- Local non-streaming speed check (`test_voice_cloning_non_streaming`, c16, 2x H100): 1088/1088 completed, 0 failed, throughput 7.305 req/s, mean latency 2.181s, mean RTF 0.5244; no HTTP 500 / CUDA OOM observed.

zhaochenyang20 (approved):

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -38,6 +33,18 @@
 _SOURCE_HINT = "MOSS-TTS Local"
 
 
+def _vocoder_cuda_graph_enabled() -> bool:
+    """True unless ``MOSS_VOCODER_CUDA_GRAPH`` is 0/false/no/off (default ON; every failure path
+    degrades to eager, so default-on never crashes)."""
+    return os.environ.get("MOSS_VOCODER_CUDA_GRAPH", "1").strip().lower() not in (
```
gaoyang07: The environment variable change here feels a bit inelegant to me.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
zhaochenyang20, line 1: if the first request is non-streaming, `_decode_codes_rows()` calls `_close_idle_startup_session_locked()` at `streaming_vocoder.py:749`, sees `not self._session_used_by_streaming` and empty streaming state, then closes and clears the session. A later real streaming request creates a fresh `_CodecStreamSession` in `_ensure_slot()`, but warmup only ran once in `start()`, so the new session has no `_cg_runner` and streaming permanently falls back to eager for that process. This ordering is realistic in mixed traffic or non-streaming probes after health checks; the direct performance impact is that default-on remains enabled while streaming gets no graph benefit and falls back to eager.

zhaochenyang20: Do not let “preserve the pure non-streaming baseline” implicitly destroy “keep the captured streaming graphs.” One fix is to avoid closing a startup session that already has captured frames and make an explicit choice for whether non-streaming should use the offline lane; another is to recapture synchronously for the first later streaming request under the same readiness and locking guarantees; another is to explicitly disable vocoder graphs for non-streaming-only CI or deployments. Whichever path is chosen, add a test that warms a fake runner, runs a non-streaming batch first, then runs a streaming chunk, and asserts the later streaming session still has a usable graph runner, or that logs and metrics explicitly say graphs were permanently disabled by non-streaming-first traffic.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -266,6 +355,31 @@ def __init__(
 
     def start(self) -> None:
         try:
+            # Note: (Jiaxin Deng) capture graphs ONCE here -- codec loaded, GPU quiescent, before the
+            # serving loop (never capture mid-serve). default on, opt out with MOSS_VOCODER_CUDA_GRAPH=0; best-effort -> eager.
+            if _vocoder_cuda_graph_enabled():
+                captured: list[int] = []
+                try:
+                    captured = self.warmup_cuda_graph()
```
zhaochenyang20: the PR still runs warmup capture inside the scheduler thread's `start()`, while pipeline readiness only waits for `StageRuntime.start()` to launch that thread. `sglang_omni/pipeline/stage/runtime.py:190` calls `self._scheduler_thread.start()` and immediately records the stage as started; `sglang_omni/pipeline/stage_workers.py:418` then marks the process ready. At that point the vocoder thread may still be inside `warmup_cuda_graph()` capture. Requests can already enter other stages in the same process and the same GPU context, so first traffic can compete with capture on CUDA streams, allocator state, or graph mempools, causing partial capture failure or complete fallback. Even if eager fallback preserves audio correctness, default-on performance then depends on startup traffic timing.

zhaochenyang20 (changes requested):

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -144,13 +201,47 @@ def step(self, slot_codes: dict[int, torch.Tensor]) -> dict[int, torch.Tensor]:
...
-        audio_cpu = result.audio[slots].detach().to("cpu", torch.float32)
-        lengths_cpu = result.audio_lengths[slots].detach().to("cpu")
+        graphed = None
+        graph_failed = False
+        try:
+            with torch.no_grad():
+                if self._cg_runner is not None:
+                    try:
+                        graphed = self._cg_runner.decode_step(
+                            codes_step, codes_lengths, exec_mask
```
gaoyang07: `codes_step` seems not used in decode_step()

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Done. Env vars replaced by model-scoped pipeline-config fields (`cuda_graph` / `cuda_graph_frames` / `cuda_graph_min_free_gb`), injected into the vocoder factory by `model_post_init`. No env or CLI surface.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Fixed. Streaming session creation routes through `_ensure_session_graphed` (3 paths: `_ensure_slot`, slot-starved `on_stream_done`, `_decode_payload_codes`), recapturing after non-streaming closes the startup session, at-most-once via a `warmup_attempted` flag. Added the regression test you described, parametrized over all 3 paths.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Fixed. Capture moved from the scheduler thread's `start()` into the factory: `create_vocoder_executor` calls `warmup_now` synchronously before the stage process is marked ready, so first traffic can't race capture.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: The unused arg was `codes_lengths` (the graph replays with baked static lengths), now removed from `decode_step`. `codes_step` is still used (copied into the static buffer); the eager fallback still uses lengths.

tests/unit_test/moss_tts_local/test_streaming_vocoder.py
```diff
@@ -826,3 +846,172 @@ def test_is_streaming_payload(monkeypatch) -> None:
...
+    else:  # no_chunk_done: terminal payload replay with no chunks -> _decode_payload_codes
+        scheduler._on_done("nc")
+        scheduler._on_streaming_new_request(
+            "nc", _terminal_payload(_rows(5, seed=5), request_id="nc")
+        )
+
+    assert scheduler._session is not None
+    assert (
+        id(scheduler._session) != startup_id
+    ), "a fresh session must have been created"
```
zhaochenyang20: Keep a reference to the old object instead of only its integer id, for example `startup_session = scheduler._session`, assert `startup_session._closed` after non-streaming traffic, and later assert `scheduler._session is not startup_session`, `len(calls) == 2`, and `scheduler._session.has_cuda_graph_runner()`. Holding the old object prevents Python from freeing and reusing its address, and the test then states the real invariant: the old session was closed and a new session was recaptured.

zhaochenyang20 (changes requested):

tests/unit_test/moss_tts_local/test_streaming_vocoder.py
JiaxinD: Done, applied your approach: retained `startup_session` ref, assert `startup_session._closed` after non-streaming, then `scheduler._session is not startup_session`; plus `len(calls) == 2` (recapture happened) and `has_cuda_graph_runner()` on the new session. Pushed; all 3 params pass.

zhaochenyang20 (approved):

Hayden727 (approved):

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
```diff
@@ -0,0 +1,255 @@
...
+        self._batch_size = int(batch_size)
+        self._n_vq = int(n_vq)
+        self._device = next(codec.parameters()).device
+        self._max_frames = int(max_frames)
+        self._max_graphs = int(max_graphs)
+        self._warmup_iters = int(warmup_iters)
+        # Min free VRAM to attempt a capture (each graph is multi-GB); below it we skip -> eager,
+        # so a VRAM-tight box degrades gracefully instead of OOM-ing.
+        self._min_free_bytes = int(float(min_free_gb) * (1024**3))
+        self._graphs: dict[int, tuple] = {}
```
Ratish1: self._graphs stores five positional values and decode_step() unpacks them later. That is fragile for graph code because adding one static input or output buffer will silently shift positions unless every unpack is updated.

should we add a NamedTuple or frozen dataclass such as _CapturedVocoderGraph(graph, static_codes, static_lengths, static_audio, static_audio_lengths) and type _graphs as dict[int, _CapturedVocoderGraph].

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -64,27 +58,77 @@ def _build_usage(state: MossTTSLocalState) -> dict[str, Any] | None:
...
+        self._cg_graph_t: Counter = Counter()
+        self._cg_eager_t: Counter = Counter()
+        self._cg_total_steps = 0
+        # Retain the streaming ExitStack so per-slot causal state lives across steps (closed in close());
+        # graph replay is kept bit-identical to this stateful decode by the in-place cache patch.
+        self._exit_stack = contextlib.ExitStack()
         with torch.no_grad():
             self._exit_stack.enter_context(codec.streaming(self._batch_size))
 
+    def warmup_cuda_graph(self, frames, *, min_free_gb: float = 3.0) -> list[int]:
```
Ratish1: nit: add type hint for frames here?

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
```diff
@@ -0,0 +1,255 @@
...
+                    logger.warning(
+                        "MOSS vocoder CG: free VRAM %.1fGB < %.1fGB headroom; skipping T=%d+ (eager)",
+                        free / 1024**3,
+                        self._min_free_bytes / 1024**3,
+                        t,
+                    )
+                    break
+                try:
+                    self._capture_t(t)
+                except (
```
Ratish1: nit: except (Exception) as exc: should just be except Exception as exc:.

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
```diff
@@ -0,0 +1,255 @@
...
+        self._min_free_bytes = int(float(min_free_gb) * (1024**3))
+        self._graphs: dict[int, tuple] = {}
+        self._pool = None
+        self._sealed = False
+        # Reused all-active mask for the warmup-only state reset (avoid re-allocating it per captured T).
+        self._reset_mask = torch.ones(
+            self._batch_size, dtype=torch.bool, device=self._device
+        )
+
+    def _eligible(self, t: int) -> bool:
```
Ratish1: nit: here we do exact codec frame count, but _eligible(t) and _capture_t(t) hide that behind a one-letter variable. can we rename  these to something like _is_supported_frame_count(frame_count) and _capture_frame_count(frame_count) . wdyt

sglang_omni/models/moss_tts_local/config.py
```diff
@@ -97,6 +97,24 @@ def talker_sglang_role_to_stage(cls) -> dict[str, str]:
         default_factory=lambda: _stages(codec_device="cuda:0", colocated=True)
     )
 
+    # Streaming-vocoder CUDA-graph knobs, injected into the vocoder factory by model_post_init.
+    # Default on, fail-safe to eager; cuda_graph_frames=None uses the built-in broad-exact set.
+    cuda_graph: bool = True
+    cuda_graph_frames: list[int] | None = None
```
Ratish1, line 103: should we validate cuda_graph_frames and cuda_graph_min_free_gb at the config boundary.

Today invalid frame counts are silently filtered later, and [] is falsey so it means "use defaults" rather than "capture no frames". A negative cuda_graph_min_free_gb is also accepted. For a default-on optimization, bad config should fail fast instead of looking like a low-VRAM skip or a successful eager fallback.

Ratish1 (approved):
LGTM with some nits and small changes to do

JiaxinD:
Thanks for the review. All five done in aa98f7a: NamedTuple for the captured-graph buffers, `frames: list[int]` hint, `except Exception as exc`, renames to `_is_supported_frame_count` / `_capture_frame_count`, and config-boundary fail-fast validation.

sglang_omni/models/moss_tts_local/config.py
JiaxinD: Done, fail-fast at the boundaries: reject `[]` (use `cuda_graph: false` to disable), reject non-positive frames at config, and reject `> max_step_frames` at scheduler init, so user-supplied frames are never silently filtered. Only the auto-default set keeps an in-range clamp, since adapting the generated default to a small `max_step_frames` is not dropping user input.

PR #803, opened 2026-06-16
[CI Perf] Trace stage relay payload IO

luojiaxuan:
## Summary
- add explicit stage IO tracing via `StageConfig.trace_stage_io` / `trace_stage_io_sync`
- record relay payload tensor summaries and write/read timing for stage edges
- emit `stage_io_trace` profiler events and server log JSON lines while disabled by default

## Validation
- `python3 -m py_compile` on changed Python files
- focused pytest was not run locally because both available Python runtimes are missing `pytest`

## Notes
This is instrumentation only and does not change relay behavior unless tracing is enabled explicitly in pipeline config.

Tracked in #934.

### DIFF
--- a/sglang_omni/config/schema.py
+++ b/sglang_omni/config/schema.py
@@ -174,6 +174,8 @@ class StageConfig(BaseModel):
 
     # --- Relay (auto-inferred from gpu when None) ---
     relay: RelayConfig | None = None
+    trace_stage_io: bool = False
+    trace_stage_io_sync: bool = False
 
     def model_post_init(self, __context: Any = None) -> None:
         fields_set = self.__pydantic_fields_set__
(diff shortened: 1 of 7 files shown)


luojiaxuan (commented):
Reviewed. Stage IO tracing is env-gated and reasonably localized; no low-risk simplification needed.

luojiaxuan:
Rebased onto current main and removed the trace env gate. Stage IO tracing is now enabled explicitly through StageConfig.trace_stage_io / trace_stage_io_sync.

luojiaxuan:
Roadmap tracker: #934. This PR is the relay IO visibility piece for the broader Qwen3 Omni serving/CI optimization story.

PR #807, opened 2026-06-16
T1: shared pipeline-state base + all 6 TTS migrations (+ quick-wins)

Hayden727:
## Summary

Template **T1 — Pipeline state serialization** from the TTS optimization RFC (tracking issue #661): a shared `PipelineStateBase` that de-duplicates the per-model `to_dict` / `from_dict` / `build_usage` boilerplate while leaving each model's tensor strategy model-owned.

This revision aligns the PR with the RFC:

- **All six in-scope TTS models migrated.** higgs / moss / moss_local / qwen3_tts / fishaudio_s2_pro already subclassed `PipelineStateBase`; this adds the last one — **voxtral** (its state lives in `voxtral_tts/io.py`, not `payload_types.py`).
- **TypedTensor escape hatch lifted into the base.** Voxtral's exact bytes+dtype+shape codec moves to `scheduling/typed_tensor.py` (`encode_typed_tensor` / `decode_typed_tensor`, re-exported from `pipeline_state`) as the shared primary escape hatch for exact dtype round-trips. Tensor handling stays model-owned — `.tolist()`, keep-CPU-tensor (`serialize_value`), and TypedTensor coexist (RFC §5).
- **Opt-in `schema_version` guard** (`None` default = no guard, preserves today's behavior) via `append_schema_version` / `check_schema_version`.
- **Quick-wins folded in** (the RFC's three-item quick-win bundle; the first two were already here): `usage_payload` (= the shared `build_usage`), `resolve_checkpoint` (`utils/checkpoint.py`), and now **`reverse_delay_pattern`** — the de-delay transform shared in `utils/codec_delay.py` (`allow_short` toggles Higgs-raise vs MOSS-empty), deleting the higgs/moss copies.

### Split out of this PR
- **T3 (SimpleScheduler concurrent-batch unlock)** — the `max_concurrency>1 × batch_compute_fn` guard removal — is behavior-changing (RFC P3) and **had no consumer here** (moss_local coalesces via its own `queue.Queue` daemon thread, not `SimpleScheduler`). It is reverted from this PR (`simple_scheduler.py` + its concurrency test return to `main`) and parked on a separate branch for isolated review when a concurrent-batch encoder actually needs it.

## Behavior class

🟢 **Behavior-preserving** — structural serialization dedup only; `schema_version` is opt-in and unused, so there is no runtime change.

## Validation

- CPU unit tests (`tests/unit_test/scheduling/test_pipeline_state.py`): `schema_version` opt-in guard, TypedTensor round-trip / int32 promotion / legacy-list fallback / missing, the six-model base contract, and load/store round-trip. Voxtral state round-trip verified directly.
- `reverse_delay_pattern` validated against both call sites (Higgs `allow_short=False` → raise; MOSS `allow_short=True` → empty), output-identical to the deleted copies.
- ⚠️ Full unit suite + SeedTTS perf/accuracy gate (RFC §15) to be re-run on the H100/H200 container before merge. T1 is behavior-preserving, so the prior SeedTTS numbers below are expected to hold; they were measured **before** the voxtral migration + T3 split.

## SeedTTS Results (prior measurement, pre-voxtral / pre-T3-split)

Full SeedTTS vs main. Each row completed 1088/1088 requests with 0 failed and 0 skipped.

| Model | Mode | WER before -> after | QPS before -> after | Latency before -> after | RTF before -> after |
|---|---|---:|---:|---:|---:|
| FishAudio | non-stream | 0.012225 -> 0.010969 | 1.100 -> 1.165 | 7.253s -> 6.845s | 1.9537 -> 1.8428 |
| FishAudio | stream | 0.011555 -> 0.012308 | 1.080 -> 1.134 | 7.389s -> 7.040s | 1.9872 -> 1.8931 |
| MOSS | non-stream | 0.016746 -> 0.017500 | 2.606 -> 2.581 | 3.061s -> 3.090s | 0.7193 -> 0.7263 |
| MOSS | stream | 0.019677 -> 0.016746 | 2.571 -> 2.582 | 3.101s -> 3.090s | 0.7288 -> 0.7273 |
| Higgs | non-stream | 0.011136 -> 0.010215 | 15.325 -> 15.006 | 1.029s -> 1.060s | 0.2397 -> 0.2439 |
| Higgs | stream | 0.011722 -> 0.011220 | 13.463 -> 14.270 | 1.105s -> 1.116s | 0.2488 -> 0.2493 |

Regression gate passed at measurement time: no mean throughput, latency, or RTF regression beyond 5%; max WER absolute increase was +0.000754.


Hayden727:
## Performance validation — SeedTTS paired (main vs this PR)

Ran on H100. Each cell is a **paired, interleaved** main-vs-PR comparison (`benchmark_tts_seedtts --generate-only`, concurrency 8, SeedTTS EN). **0 failed requests** throughout.

| Model | Mode | QPS (main → PR) | RTF (main → PR) | samples |
|---|---|---|---|---|
| **moss_tts_local** | non-stream | 7.60 → 7.61 (+0.1%) | 0.2477 → 0.2475 (−0.1%) | 1088 |
| **moss_tts_local** | stream | 4.89 → 4.91 (+0.4%) | 0.3814 → 0.3801 (−0.3%) | 1088 |
| **higgs_tts** | non-stream | 10.20 → 10.26 (+0.6%) | 0.1882 → 0.1890 (+0.4%) | 1088 |
| **higgs_tts** | stream | 9.69 → 9.86 (+1.7%) | 0.1970 → 0.1967 (−0.2%) | 1088 |
| **fishaudio_s2_pro** | non-stream | 1.076 → 1.076 (~0%) | 2.161 → 2.164 (+0.1%) | 200 |
| **fishaudio_s2_pro** | stream | 1.058 → 1.064 (+0.5%) | 2.196 → 2.196 (−0.1%) | 200 |

**No regression.** All deltas are within noise; RTF (the compute-bound metric) is flat (±0.5%) everywhere; 0 failed requests. This is consistent with T1 being a behavior-preserving serialization refactor — the changed code (`to_dict`/`from_dict`/`build_usage`, the shared `reverse_delay_pattern`) runs **once per request at stage boundaries**, not in the per-token generation loop.

<details>
<summary>Note on higgs non-stream (why it first looked like a −3% dip)</summary>

A naive paired ordering (`main, pr, main, pr`) initially showed higgs non-stream QPS −3.2%. That was a **methodology artifact**: PR always sampled *after* its paired main, so a monotonic system drift in that run (10.26 → 10.14 → 10.01 → 9.48) biased PR downward. Re-measured with a **drift-balanced order** (`main, pr, pr, main`, so both average run-position 2.5):

```
raw qps:  main-a=10.125  pr-a=10.226  pr-b=10.299  main-b=10.270
balanced: main=10.197  pr=10.262  → PR +0.6% (faster)
GPU:      1980 MHz before every run, 34–38 °C  → no thermal throttling
```

PR comes out marginally faster and the per-run QPS is stable, confirming the original dip was the measurement, not the code.
</details>

Coverage: all 6 in-scope TTS models migrate to `PipelineStateBase` and are covered by the unit suite (439 passed, incl. `moss_tts_local`). Validation env: container from `hongccc/sglang-omni:dev`, paired against `origin/main`.

sglang_omni/scheduling/typed_tensor.py
```diff
@@ -0,0 +1,58 @@
...
+msgpack cannot pack a Tensor (it can pack bytes). Voxtral's audio_codes ends up
+there, so it must be bytes. A keep-CPU-tensor field (serialize_value) only works
+when it never crosses a control-plane message (relay side-channel hops only).
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+import numpy as np
```
JiaxinD: Top-level `import torch` / `import numpy` here, and pipeline_state imports this module unconditionally, so all 6 payload_types now hard-require torch+numpy at import time. main imported torch lazily (inside `_tensor_to_payload`); moving these into encode/decode keeps that lazy-import path.

tests/unit_test/scheduling/test_pipeline_state.py
```diff
@@ -0,0 +1,179 @@
...
+    tensor = torch.tensor([1, 2], requires_grad=False)
+
+    value = PipelineStateBase.serialize_value(tensor)
+
+    assert isinstance(value, torch.Tensor)
+    assert value.device.type == "cpu"
+    assert value.tolist() == [1, 2]
+
+
+def test_tts_pipeline_states_share_base_usage_contract() -> None:
```
JiaxinD, line 84: This contract test only checks subclassing + that to_dict/from_dict are callable; it never round-trips the 6 states, so a model's to_dict drifting (the migration's main risk) would still pass. Suggest asserting `from_dict(to_dict(s))` key parity per model.

sglang_omni/scheduling/pipeline_state.py
```diff
@@ -0,0 +1,99 @@
...
+
+@dataclass
+class PipelineStateBase:
+    """Shared usage/serialization mechanics; tensor strategy stays subclass-owned."""
+
+    sample_rate: int = 24000
+    prompt_tokens: int = 0
+    completion_tokens: int = 0
+    engine_time_s: float = 0.0
+    schema_version: int | None = None  # opt-in guard; None = no check
```
JiaxinD: `schema_version` plus `append_schema_version`/`check_schema_version` are added but no migrated model sets or checks them (only tests touch them). Suggest dropping until a model needs it.

sglang_omni/scheduling/pipeline_state.py
```diff
@@ -0,0 +1,99 @@
...
+    def check_schema_version(data: dict[str, Any], expected: int) -> None:
+        """Raise if the payload's schema_version differs from expected; untagged is ok."""
+        found = data.get("schema_version")
+        if found is not None and int(found) != int(expected):
+            raise ValueError(
+                "pipeline-state schema_version mismatch: "
+                f"payload={found}, expected={expected}"
+            )
+
+    def usage_dict(self) -> dict[str, Any] | None:
```
JiaxinD: `usage_dict()` has no callers; all models call `build_usage(state)` directly. Looks like dead surface, suggest removing.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -33,17 +34,16 @@
...
-        return None
-    usage = {
-        "prompt_tokens": int(state.prompt_tokens),
-        "completion_tokens": int(state.completion_tokens),
-        "total_tokens": int(state.prompt_tokens + state.completion_tokens),
-    }
-    if state.engine_time_s:
-        usage["engine_time_s"] = round(float(state.engine_time_s), 6)
-    return usage
+def _resolve_sample_rate(processor: Any) -> int:
```
JiaxinD: `_resolve_sample_rate` is never called; `self._sample_rate` is still set from the constructor arg (line 347). Wire it in or remove.

JiaxinD (commented):
Reviewed the T1 refactor. One import-time regression plus some unused surface from the migration.

yxs (approved):

Issue #797, opened 2026-06-15
Stage 11 (Video-AMME Talker, FP8 Thinker TP=2) — Performance Investigation


MelodyyyYin:
### Goal
Stage 11 (Video-AMME Talker, FP8 Thinker TP=2) latency was ~148–170 s; a GPU snapshot (GPU0 100% / GPU1 4%) suggested load-imbalance and TP-ing code2wav was considered. This issue tracked the full investigation. **Resolved** — see TL;DR.

---

## TL;DR (final, supersedes the original per-phase reading below)
- **Root cause was NOT the talker AR decode** — it was **delivery-bound**. The thinker sat idle ~130 s with a full queue while `mm_aggregate` did a full materialize + re-serialize of the ~55 MiB video embeds under concurrency-16 head-of-line pressure on the relay. Every compute stage is small (~30 s total); both GPUs were ~85 % idle. The original "talker_ar 164 s dominant" number was stage **wall-lifetime (idle-spinning)**, not busy time — corrected in Updates 7–8.
- **Fix: #808 lazy `TensorRef`** (publish embeds to the relay once, forward a ref, resolve once): **165 s → 33 s (~5×)**, accuracy preserved (reproduced in Update 9). Equivalent to the pre-#808 **payload reduction #804** — both attack the same stall.
- **After #808 the latency floor (~30 s) is the talker AR decode = audio length** (`corr(audio_duration, latency)=0.96` vs `corr(prompt_tokens, latency)=0.49`). The only same-architecture lever below ~30 s is shortening the generated audio/answer length (eval-config, bounded by the accuracy 0.40 / WER 0.0175 gates).
- **Low GPU util (70–90 % idle) is the expected signature of a latency-bound benchmark at low concurrency (10 reqs / 4 unique videos), not a balance bug.**

## Levers evaluated — all exhausted, redundant, or blocked
| Lever | Verdict |
|---|---|
| Vision-encoder data-parallel | bit-exact but encode is 0.19 s → wrong target; reverted |
| Parallelize CPU preprocessing | preprocessing 0.79 s/req → non-bottleneck |
| Talker TP=2 | **blocked**: FP8 MoE intermediate 384 → 192/rank, not divisible by quant `block_n=128` |
| Talker cuda-graph | no latency win + #408 RNG-under-graph correctness risk (~5% garbage audio) |
| Talker async one-step lookahead | **slower** (host-bound, `event.query hit=1600 miss=0`) |
[...]

PR #808, opened 2026-06-16
[Qwen3-Omni] Lazy TensorRef path for large inter-stage tensors (#797)

luojiaxuan:
## Motivation (#797)

Stage 11's latency is dominated by inter-stage **delivery** of the large video embeddings, not compute. Per the #797 investigation (Updates 7-8), the thinker idles ~140s with a full prefill queue while the video embeddings are **materialized at `mm_aggregate` and re-serialized to the thinker** over the relay — two full round-trips of the big tensor for a stage (`mm_aggregate`) that only forwards it.

## What this does

Adds an **env-gated lazy `TensorRef` path** for large allowlisted tensors:

- `image_encoder` publishes the tensor to the relay **once** and emits a small reference dict.
- `mm_aggregate` merges/forwards the ref **without materializing** it (the ref dict passes through the existing `write_payload`/`extract_tensors` untouched, since it has no tensor leaves).
- the `thinker` resolves the ref **once**, right before its scheduler.

Default behavior is **unchanged** unless `SGLANG_OMNI_ENABLE_TENSOR_REFS=1`.

```
SGLANG_OMNI_ENABLE_TENSOR_REFS=1
SGLANG_OMNI_TENSOR_REF_EDGES=image_encoder:mm_aggregate:thinker
SGLANG_OMNI_TENSOR_REF_THRESHOLD_MB=8
SGLANG_OMNI_TENSOR_REF_PATHS=video_embeds,deepstack_visual_embeds_image,deepstack_visual_embeds_video
```

## Implementation

- **`sglang_omni/pipeline/tensor_ref.py`** (new): `TensorRef` + env-driven `TensorRefPolicy` (`from_env` returns `None` — fully inert — unless enabled and the `(from,to)` edge is declared).
- **`relay_io.py`**: `write_payload` gains optional `from_stage/to_stage/tensor_ref_policy` (defaulted, so existing call sites are byte-for-byte unchanged); allowlisted large tensors are externalized via the existing `write_blob`, with the blob op tracked as a background task (the consumer stage, not this hop, reads it). `materialize_tensor_refs` resolves a ref only at its declared `consumer_stage` and is identity-preserving (ref-free payloads are returned as-is, no per-stage container churn).
- **`stage/runtime.py`**: `_execute` materializes refs **after** TP fanout (followers materialize on their own relay/device); `_send_to_stage` wires the per-edge policy and surfaces `tensor_ref_stats` into the `stage_hop_sent` trace event.
- **`qwen3_omni/merge.py`**: `_non_empty` recognizes `TensorRef` dicts as non-empty so `merge_for_thinker` keeps forwarding them.

## Results (Stage 11 Video-AMME Talker, FP8 thinker TP=2)

H100 2×GPU, SHM relay, 10 samples / concurrency 16, prompt-tokens-mean 14444 every run:

| TensorRef | latency_mean_s | accuracy | thinker stall |
|---|---|---|---|
| OFF (cold) | 160.7 | 0.50 | ~124 s |
| OFF (warm) | 179.3 | 0.50 | ~140 s |
| **ON** | **37.3** | 0.50 | none |
| **ON** | **36.2** | 0.50 | none |

**~4.5× lower `latency_mean_s`, accuracy unchanged.** Verified not a warm-cache artifact: both OFF runs (one cold, one warm) reproduce the #797 thinker stall (`#queue-req 9`, ~2 min of silence between the first request's prefill chunks); both ON runs eliminate it. Engagement confirmed in-log (`tensor_ref active: externalizing video_embeds (55.0 MiB) on edge image_encoder->mm_aggregate`).

## Tests

`tests/unit_test/pipeline/test_tensor_ref.py` (new) + additions to `test_stage.py` and `test_pipeline.py`: relay round-trip, consumer-stage gating, TP-fanout ordering (followers get the unresolved ref), deferred-blob-op is not awaited inline, and Qwen merge ref pass-through. Full `tests/unit_test` suite green (only pre-existing shared-host `/tmp` permission failures remain).

## Not in scope / known limits

- No default CUDA-IPC full-payload path (possible follow-up once lifecycle/release is in place).
- No active release-on-abort: an aborted in-flight ref leaks its SHM segment / self-heals its NIXL credit via timeout (accepted for v1).

## Acknowledgements

Thanks to @MelodyyyYin for the earlier exploration. Although that path was not merged, it helped rule out an incorrect optimization direction and shaped this final approach.

Co-authored-by: Melody Yin <melodyyin31@gmail.com>

### DIFF
--- a/sglang_omni/pipeline/relay_io.py
+++ b/sglang_omni/pipeline/relay_io.py
@@ -99,12 +99,6 @@ def restore_tensors(obj: Any, tensor_dict: dict[str, torch.Tensor]) -> Any:
         return obj
 
 
-# ---------------------------------------------------------------------------
-# Lazy TensorRef: externalize large allowlisted tensors instead of inlining
-# them into the payload buffer. Intermediate stages forward the small ref
-# dict untouched; only the declared consumer stage resolves it.
-# ---------------------------------------------------------------------------
-
 _BACKGROUND_REF_TASKS: set[asyncio.Task] = set()
 _LOGGED_REF_EDGES: set[tuple[str, str]] = set()
 
@@ -161,7 +155,6 @@ async def publish_tensor_ref(
         path=path,
         shape=tuple(tensor.shape),
         dtype=str(tensor.dtype),
-        device=str(tensor.device),
         nbytes=tensor.numel() * tensor.element_size(),
         blob_key=blob_key,
         blob_metadata=blob_metadata,
--- a/sglang_omni/pipeline/relay_io.py
+++ b/sglang_omni/pipeline/relay_io.py
@@ -8,16 +8,26 @@ Extracted from worker/data_plane.py and worker/runtime.py.
 """
 from __future__ import annotations
 
+import asyncio
 import base64
 import io
+import logging
 import pickle
 from multiprocessing.reduction import ForkingPickler
 from typing import Any
+from uuid import uuid4
 
 import torch
 
+from sglang_omni.pipeline.tensor_ref import (
+    TensorRef,
+    TensorRefPolicy,
+    is_tensor_ref_dict,
+)
 from sglang_omni.proto import DataReadyMessage, StagePayload
 from sglang_omni.relay.base import Relay
+
+logger = logging.getLogger(__name__)
 
 
 def _dtype_alignment(dtype: torch.dtype) -> int:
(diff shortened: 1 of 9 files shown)


sglang_omni/pipeline/tensor_ref.py
```diff
@@ -0,0 +1,145 @@
...
+@dataclass(frozen=True)
+class TensorRef:
+    ref_id: str
+    request_id: str
+    producer_stage: str
+    consumer_stage: str
+    path: str
+    shape: tuple[int, ...]
+    dtype: str
+    device: str
```
yxs: `TensorRef` stores `device`, and `publish_tensor_ref` fills it with the source tensor's device (`cuda:0` in tests). But `read_tensor_ref` only checks shape and dtype — it neither checks nor restores device — reading into a tensor on `relay.device` (ShmRelay defaults to `"cpu"`). So a CUDA tensor on the producer resolves to a CPU tensor on the consumer. This matches the non-ref path (which also stages through `relay.device`),  the `device` field stores a value that is never used and contradicts the actual result, may misleading anyone reading the code into thinking the resolve yields CUDA.

sglang_omni/pipeline/relay_io.py
```diff
@@ -85,6 +99,247 @@ def restore_tensors(obj: Any, tensor_dict: dict[str, torch.Tensor]) -> Any:
         return obj
 
 
+# ---------------------------------------------------------------------------
```
yxs: remove those

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -659,9 +660,28 @@ async def _execute(self, payload: Any) -> None:
...
+            #
+            # CONTRACT: a TensorRef must be resolved exactly once per blob. The
+            # relay get is destructive (the SHM backend unlinks on read), so two
+            # ranks resolving the same ref would double-unlink. This holds today
+            # because the only ref consumer (thinker) runs OmniScheduler with
+            # requires_tp_work_fanout=False -- it never reaches this fanout, and
+            # materialize runs once on the leader. Do NOT enable a ref edge whose
+            # consumer is a requires_tp_work_fanout=True stage without making the
+            # relay read non-destructive / reference-counted first.
             self._tp_fanout.fanout_work(payload)
```
yxs, line 682: The single-resolve invariant here is load-bearing: the relay `get` is destructive (the SHM backend unlinks on read), so a TensorRef must be resolved exactly once. It is safe today only because the sole ref consumer (thinker) runs `OmniScheduler` with `requires_tp_work_fanout=False`, so `materialize_payload_tensor_refs` runs once on the leader and followers receive the already-resolved stream via `broadcast_pyobj`. If a `requires_tp_work_fanout=True` stage ever becomes a ref consumer (e.g. a TP'd talker), each follower would independently resolve the same blob and double-unlink it.

Is the single-consumer (thinker-only) assumption permanent, or should the relay blob-read become reference-counted / non-destructive now so multi-consumer refs are safe?

sglang_omni/pipeline/tensor_ref.py
```diff
@@ -0,0 +1,145 @@
...
+class TensorRefPolicy:
+    """Per-edge policy controlling which tensor leaves get externalized."""
+
+    threshold_bytes: int
+    from_stage: str
+    to_stage: str
+    consumer_stage: str
+    path_allowlist: tuple[str, ...]
+
+    def should_externalize(self, path: str, tensor: torch.Tensor) -> bool:
```
yxs, line 100: The default `path_allowlist` includes `deepstack_visual_embeds_image` / `deepstack_visual_embeds_video`, but those are `list[torch.Tensor]`. `should_externalize` strips the list index (`...image[0]` → `image`) and matches the allowlist, so each list element is externalized as its own blob / relay segment (N segments = number of deepstack layers) instead of one blob for the whole list — each with its own SHM-create + credit overhead. (Only triggers when individual layer tensors each exceed the threshold.)
Is that intended, or should an allowlisted `list[Tensor]` publish as a single blob?

sglang_omni/pipeline/tensor_ref.py
luojiaxuan: Done in 8da893d: removed the unused `device` field from `TensorRef` so the ref no longer implies producer-device restore semantics.

sglang_omni/pipeline/relay_io.py
luojiaxuan: Done in 8da893d: removed the added TensorRef section banner comments.

sglang_omni/pipeline/stage/runtime.py
luojiaxuan: Not permanent. For this PR the TensorRef edge stays thinker-only and the single-resolve contract is now documented and covered by a destructive-read regression test. We should make blob reads ref-counted/non-destructive before enabling refs for TP-fanout consumers.

sglang_omni/pipeline/tensor_ref.py
luojiaxuan: Intended for now. Each deepstack layer is externalized element-wise, and the policy docstring now calls that out explicitly. We can revisit packing the full list if SHM/credit overhead shows up in measurements.

sglang_omni/pipeline/relay_io.py
```diff
@@ -85,6 +99,240 @@ def restore_tensors(obj: Any, tensor_dict: dict[str, torch.Tensor]) -> Any:
...
+    The blob write is awaited (the bytes must be copied before this hop's
+    small envelope is sent), but its relay op is tracked in the background —
+    the consumer stage, not this hop, is the one that actually reads it.
+    """
+    edge = (from_stage, to_stage)
+    if edge not in _LOGGED_REF_EDGES:
+        _LOGGED_REF_EDGES.add(edge)
+        logger.info(
+            "tensor_ref active: externalizing %s (%.1f MiB) on edge %s->%s, "
+            "consumer=%s",
```
Ratish1: can we use f formatting here to align with the codebase?

luojiaxuan: Done in e8bb102: switched the TensorRef log to f-string formatting.

luojiaxuan:
Yes, the env gate is intentional. TensorRefs are off by default and are also edge/path/size-gated. Smaller inter-stage tensors stay inline to avoid extra relay blob creation, metadata, and SHM/credit overhead.

I ran a two-hop SHM microbenchmark on H100 (`cuda:0 -> SHM cpu`) to choose the default threshold. Mean improvement vs all-inline:

```text
threshold  mean improvement vs all-inline
0.125MB    54.73%
1MB        54.64%
2MB        54.50%
8MB        52.80%
128MB      0.00%
```

The sweep suggests 8 MiB was too conservative. I lowered the default to 2 MiB in ed3a5d8: it keeps almost the same measured gain as the aggressive 0.125 MiB threshold, while leaving more margin to avoid very small tensor overhead. 44ed214 also keeps the regression test covering small allowlisted tensors staying inline.

luojiaxuan:
Updated the TensorRef threshold based on the sweep: default is now 2 MiB in ed3a5d8. The H100 two-hop SHM sweep showed 2 MiB at 54.50% mean improvement vs all-inline, very close to the best 0.125 MiB result at 54.73%, but with a safer margin for very small tensor overhead.

luojiaxuan (commented):
Reviewed. TensorRef is scoped to policy, relay IO, and runtime materialization; no unresolved threads and no extra simplification found.

tests/unit_test/qwen3_omni/test_pipeline.py
```diff
@@ -1122,6 +1123,38 @@ def test_qwen_mm_aggregate_keeps_lightweight_inputs_and_prunes_after_merge() ->
...
+    """A lazily-externalized video_embeds ref survives merge unresolved."""
+    ref = TensorRef(
+        ref_id="req-qwen:tensor_ref:image_encoder:mm_aggregate:abc:video_embeds",
+        request_id="req-qwen",
+        producer_stage="image_encoder",
+        consumer_stage="thinker",
+        path="encoder_outs.image_encoder.video_embeds",
+        shape=(4, 8),
+        dtype="torch.bfloat16",
+        device="cuda:0",
```
yxs: [P1] Remove stale device argument from TensorRef test fixture

This test now constructs `TensorRef` with `device="cuda:0"`, but `TensorRef` no longer defines a `device` field after this PR removed that metadata. When this unit test runs it fails at construction with `TypeError: TensorRef.__init__() got an unexpected keyword argument 'device'`, so the Qwen merge regression never exercises the intended unresolved-ref pass-through behavior. Please drop the stale argument, or restore the schema field if it is still intended.

sglang_omni/pipeline/relay_io.py
```diff
@@ -85,6 +99,236 @@ def restore_tensors(obj: Any, tensor_dict: dict[str, torch.Tensor]) -> Any:
...
+    edge = (from_stage, to_stage)
+    if edge not in _LOGGED_REF_EDGES:
+        _LOGGED_REF_EDGES.add(edge)
+        logger.info(
+            f"tensor_ref active: externalizing {path} "
+            f"({tensor.numel() * tensor.element_size() / 1024**2:.1f} MiB) "
+            f"on edge {from_stage}->{to_stage}, consumer={consumer_stage}"
+        )
+    blob_key = f"{request_id}:tensor_ref:{from_stage}:{to_stage}:{uuid4().hex}:{path}"
+    blob_metadata, op = await write_blob(relay, blob_key, tensor)
```
Hayden727, line 237: Please make TensorRef blob ownership explicit here: this creates a separate relay blob, but abort/failure cleanup only calls `relay.cleanup(request_id)`, which does not unlink the SHM blob if the consumer never materializes the ref. Track the created ref blob metadata per request and drain/release it from abort/discard/failure paths, or add an explicit relay blob release API before enabling this path.

Hayden727:
Workload: Video-AMME Talker, `max_samples=10`, `max_concurrency=16`, `max_tokens=256`, `video_fps=2`, `video_max_frames=128`, `video_max_pixels=401408`, text+audio output, WER disabled. Each mode was run twice.

| model | TensorRef | latency_mean_s | latency_p95_s | throughput_qps | rtf_mean | failed | tensor_ref_active |
| BF16/Instruct | OFF | 16.645 [11.781, 21.508] | 18.259 | 0.567 | 1.2555 | 0 | 0 |
| BF16/Instruct | ON | 11.597 [11.558, 11.636] | 13.288 | 0.732 | 0.8569 | 0 | 2 |
| FP8 TP=2 | OFF | 85.397 [79.250, 91.544] | 87.061 | 0.114 | 6.1268 | 0 | 0 |
| FP8 TP=2 | ON | 17.480 [17.348, 17.611] | 19.070 | 0.508 | 1.2754 | 0 | 2 |

Delta (ON vs OFF mean):

- BF16/Instruct: latency_mean -30.33%, latency_p95 -27.22%, throughput +29.10%, rtf_mean -31.75%.
- FP8 TP=2: latency_mean -79.53%, latency_p95 -78.10%, throughput +343.67%, rtf_mean -79.18%.

Interpretation note: the BF16 first OFF run included cold-cache/compile effects; the warmed BF16 OFF run was 11.781s vs warmed ON 11.558s, so BF16 steady-state benefit appears small. The FP8 TP=2 result is stable across both runs and shows a large win.

Hayden727 (approved):

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -1322,10 +1369,29 @@ def _record_aborted_request_id(self, request_id: str) -> None:
 
     def _on_abort(self, request_id: str) -> None:
         self._record_aborted_request_id(request_id)
+        self._release_tracked_tensor_refs(request_id)
         self.relay.cleanup(request_id)
         self._clear_request_state(request_id)
         self.scheduler.abort(request_id)
 
+    def _remember_payload_tensor_refs(self, request_id: str, payload: Any) -> None:
```
yxs, line 1377: `_remember_payload_tensor_refs()` currently replaces the per-request tracked refs on every payload, and clears the entry when the latest payload has no refs. That loses refs in a normal fan-in ordering:
1. `image_encoder` payload arrives with a TensorRef -> `_unresolved_tensor_refs[request_id]` tracks the blob.
2. `preprocessing` or `audio_encoder` payload arrives without refs -> `_remember_payload_tensor_refs()` pops the tracked entry.
3. The request aborts before the merged payload finishes materializing -> `_release_tracked_tensor_refs()` has nothing to release, so the earlier blob is leaked.

The repro pauses inside materialize_payload_tensor_refs() to model the abort interleaving; at the abort point the tracked ref map is already empty and the blob remains in relay storage.

PR #809, opened 2026-06-16
T2a: reusable reference-audio encode (LRU + single-flight) + moss_local migration

Hayden727:
## Summary

Template **T2a — Reference audio encode / LRU cache / single-flight dedup** from the TTS optimization RFC (tracking issue #661). The framework gains the reusable reference-encode *mechanics*; the model keeps its encode function and key policy.

`scheduling/reference_encoder.py`:
- **`ReferenceEncodeCache`** — content-addressed LRU (byte-bounded, on `StageOutputCache`) with **single-flight** miss coalescing (`threading.Lock` + `concurrent.futures.Future`): N concurrent requests for the same reference share one encode. Followers are woken on success *and* failure without poisoning the entry; the leader's result is only cached if `revalidate()` still holds.
- **`BatchedReferenceEncoder`** — coalesces concurrent encodes into one batched worker call (a `queue.Queue` daemon thread), de-duplicating within a batch and falling back to per-item encode if the batch call fails.
- **Reference migration:** `moss_tts_local` (the RFC's reference impl) is rewired onto both, deleting its bespoke cache/dedup/batcher.

## Reasonable deviations from the RFC (documented)

The RFC's T2a sketch is followed in spirit, with three deliberate deviations that *reduce* duplication and risk:

1. **No new `AudioRef` / `resolve_audio_ref`.** That normalizer already exists — more capable and multi-modal — as `preprocessing/cache_key.py` (`reference_path_cache_key` with the `trust_stat` large-file opt-in from #740; `compute_media_cache_key` for path/URL/data-URI/bytes). moss_local already keys through it. Adding `AudioRef` would duplicate it, the exact thing #661 targets, so we reuse it.
2. **No dependency on T3.** Batching runs on a standalone `queue.Queue` daemon thread, not `SimpleScheduler`'s concurrent-batch path — so this PR needs neither the behavior-changing `max_concurrency>1 × batch_compute_fn` unlock nor a separate scheduler change. (The RFC's T2b→T3 dependency is an artifact of routing batch through `SimpleScheduler`; the standalone batcher avoids it.)
3. **Scope = the reference model only.** higgs and qwen3_tts resolve reference artifacts through the process-wide **`SpeakerArtifactCache`** singleton (shared with `serve/speech_voices.py`); per RFC §6 their migration must be sequenced together and re-validated on GPU, so it is deferred to a follow-up. fishaudio (cache-less) and moss_tts (blocked on #730's 3→4-stage split) are also follow-ups. voxtral is N/A (rejects reference audio).

## Behavior class

🟢 **Output-equivalent** — adds cache + single-flight dedup; same encoded artifacts, fewer redundant GPU encodes (different timing).

## Validation

- Unit: `tests/unit_test/scheduling/test_reference_encoder.py` (4) — cache hit/miss, single-flight coalescing, failure propagation without poisoning; moss_tts_local pipeline (43).
- ⚠️ SeedTTS perf/accuracy gate (RFC §15) to re-run on the H100/H200 container before merge; the table below is the prior measurement of this PR's behavior.

## SeedTTS Results (prior measurement)

Full SeedTTS vs main. Each row completed 1088/1088 requests with 0 failed and 0 skipped. MOSS non-stream speed columns use a paired rerun mean (generate-only); its WER is from the full E2E run.

| Model | Mode | WER before -> after | QPS before -> after | Latency before -> after | RTF before -> after |
|---|---|---:|---:|---:|---:|
| FishAudio | non-stream | 0.012225 -> 0.011555 | 1.100 -> 1.134 | 7.253s -> 7.037s | 1.9537 -> 1.8981 |
| FishAudio | stream | 0.011555 -> 0.011052 | 1.080 -> 1.059 | 7.389s -> 7.532s | 1.9872 -> 2.0280 |
| MOSS | non-stream | 0.016746 -> 0.013983 | 2.558 -> 2.532 | 3.118s -> 3.151s | 0.7337 -> 0.7411 |
| MOSS | stream | 0.019677 -> 0.017081 | 2.571 -> 2.589 | 3.101s -> 3.082s | 0.7288 -> 0.7245 |
| Higgs | non-stream | 0.011136 -> 0.010801 | 15.325 -> 15.348 | 1.029s -> 1.037s | 0.2397 -> 0.2392 |
| Higgs | stream | 0.011722 -> 0.010383 | 13.463 -> 15.254 | 1.105s -> 1.042s | 0.2488 -> 0.2470 |

Regression gate passed at measurement time: no throughput, latency, or RTF regression beyond 5%; WER did not regress.

### DIFF
--- a/sglang_omni/models/moss_tts_local/stages.py
+++ b/sglang_omni/models/moss_tts_local/stages.py
@@ -3,12 +3,8 @@
 
 from __future__ import annotations
 
-import concurrent.futures
 import logging
 import os
-import queue
-import threading
-import time
 from typing import Any
 
 import torch
@@ -33,8 +29,11 @@ from sglang_omni.models.moss_tts_local.streaming_vocoder import (
 from sglang_omni.preprocessing.cache_key import (
     reference_path_cache_key as _reference_path_cache_key,
 )
+from sglang_omni.scheduling.reference_encoder import (
+    BatchedReferenceEncoder as _SharedBatchedReferenceEncoder,
+)
+from sglang_omni.scheduling.reference_encoder import ReferenceEncodeCache
 from sglang_omni.scheduling.simple_scheduler import SimpleScheduler
-from sglang_omni.scheduling.stage_cache import StageOutputCache
 
 logger = logging.getLogger(__name__)
 
@@ -130,13 +129,18 @@ class _BatchedReferenceEncoder:
         max_batch_wait_ms: int = 4,
     ) -> None:
         self._processor = processor
-        self._max_batch_size = max(int(max_batch_size), 1)
-        self._max_wait_s = max(float(max_batch_wait_ms), 0.0) / 1000.0
-        self._queue: queue.Queue[tuple[str, concurrent.futures.Future]] = queue.Queue()
-        self._thread = threading.Thread(
-            target=self._worker, name="moss-local-ref-encode", daemon=True
+        self._encoder = _SharedBatchedReferenceEncoder(
+            lambda paths: self._processor.encode_audios_from_path(paths),
+            single_encode_fn=lambda path: self._processor.encode_audios_from_path(
+                [path]
+            )[0],
+            validate_fn=self._check_reference_duration,
+            key_fn=str,
+            max_batch_size=max_batch_size,
+            max_batch_wait_ms=max_batch_wait_ms,
+            encode_timeout_s=self.ENCODE_TIMEOUT_S,
+            worker_name="moss-local-ref-encode",
         )
-        self._thread.start()
 
     @classmethod
     def _check_reference_duration(cls, path: str) -> None:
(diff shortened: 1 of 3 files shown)


Hayden727:
Pausing this PR (converting to draft). Since it was branched, `main` has independently rewritten moss_tts_local's reference encoder: it added `_PathReferenceJob`/`_WaveformReferenceJob` job types, `encode_wav` / `encode_data_uri`, and its **own** local cache + single-flight (`CachedReferenceEncoder` with `_inflight`, keyed via `preprocessing/cache_key.py:reference_path_cache_key`). That both supersedes this PR's extraction (which deleted the older local encoder) and adds capabilities the shared `scheduling/reference_encoder.py` classes don't yet express. Net: the T2a cache/single-flight mechanic now exists locally on main, so this extraction is largely redundant as-is. Parking until we decide between (a) re-doing the extraction on top of main's richer encoder, or (b) sharing only the cache. Focus moves to #807 (T1).

Hayden727:
Closing per the above: main has independently re-implemented and out-featured this extraction in moss_tts_local. If we revisit T2a, we'll either re-do the extraction on main's richer encoder or share only the cache — tracked under the RFC (#661).

Issue #801, opened 2026-06-16
[Tracking] MOSS-TTS-Local streaming vocoder: full stateful CUDA graph (blocked on codec KV-cache fix)


JiaxinD:
## Goal

Capture the full **stateful** MOSS-TTS-Local streaming vocoder decode (`_CodecStreamSession.step` to
`codec._decode_frame`) into CUDA graphs and replay it, instead of re-launching the per-frame kernel storm.
Under streaming the codec decodes one tiny chunk per scheduler tick, so it is launch-bound: the target is a
large per-step speedup. Measured ~2.87x on the per-step decode (T=5, B=8: eager 53.3 ms vs graph replay
18.5 ms) when run without the blocker below.

This issue tracks the PRs and discussion for the vocoder CUDA graph direction. Current status: **blocked on a
codec-side KV-cache storage fix** (see Blocker).

## State distribution (measured, confirmed with Gary)

Where the cross-step streaming state actually lives in the decode path. From running one `_decode_frame` step
and checking which `_streaming_state` modules mutate (98 of them, all under `decoder`):

- **Decoder attention x92** (`MossAudioTokenizerMultiheadAttention` / `MHAState`): each holds the streaming
  KV cache (`cached_keys` / `cached_values` / `cached_positions`) plus a RoPE `offset`.
- **Decoder transformer x6** (`MossAudioTokenizerTransformer` / `TransformerState`): each holds `offsets`
  (a position counter).
- **Quantizer** (`MossAudioTokenizerResidualLFQ`) is **stateless**: not a StreamingModule (no
  `_streaming_state`), `decode_codes` is a pure lookup.
- **conv / upsampling**: no streaming state.

So a full stateful vocoder graph has exactly **one** state-mechanism blocker (the KV cache), not many. The
`offset` / `offsets` (92 + 6) are already in-place tensor updates and are graph-safe.

## Blocker / prerequisite (codec team)

The streaming attention KV cache is stored by **reassigning the state attribute to a freshly allocated tensor
every step**, so its `data_ptr` changes each step. A CUDA graph records ops on fixed addresses and does not
follow a Python-level attribute rebind, so replay keeps reading the capture-time (empty) cache. This is the
one thing that blocks stateful graph capture (with state restored, graph replay diverges from eager by
maxdelta ~0.3 to 0.96 on real codes: step 0 exact, step 1 onward stale).

[...]

PR #811, opened 2026-06-16
[Perf] MOSS-TTS: Add CUDA graph codec decoder

CloudRipple:
## Motivation
This pull request introduces support for CUDA Graphs in the MOSS-TTS Local streaming vocoder, aiming to improve inference efficiency and performance, especially for batch audio decoding on compatible CUDA devices. The changes add new options for enabling CUDA Graphs, implement logic to manage CUDA Graph warmup and execution, and ensure compatibility and fallback when CUDA Graphs are not available.

## Modifications
**CUDA Graphs Integration and Configuration:**

* Added support for CUDA Graphs in the streaming vocoder by introducing new parameters (`use_cuda_graph`, `cuda_graph_step_frames`) to the `create_vocoder_executor` and vocoder scheduler classes, allowing users to enable and configure CUDA Graph execution. [[1]](diffhunk://#diff-3fa7e30e7aba8595743c0ca045b50e50cf8131c8bb2ccdd4868f11e0ae583922R591-R592) [[2]](diffhunk://#diff-3fa7e30e7aba8595743c0ca045b50e50cf8131c8bb2ccdd4868f11e0ae583922R603-R604) [[3]](diffhunk://#diff-862eccdc1e989e72b58b23348d7e36787c0cb2f130015fb50665c8762e598314R431-R432)
* Implemented CUDA Graph warmup routines and step frame bucketization, including helper functions (`_default_cuda_graph_step_frames`, `_normalize_cuda_graph_step_frames`) to determine and normalize step frame schedules for graph capture. [[1]](diffhunk://#diff-862eccdc1e989e72b58b23348d7e36787c0cb2f130015fb50665c8762e598314R48-R82) [[2]](diffhunk://#diff-862eccdc1e989e72b58b23348d7e36787c0cb2f130015fb50665c8762e598314L76-R126)

**Codec and Session Management Enhancements:**

* Updated `_CodecStreamSession` to manage CUDA Graph decoder instances, warm up graphs for various chunk and slot configurations, and select the appropriate graph bucket for each decoding step. [[1]](diffhunk://#diff-862eccdc1e989e72b58b23348d7e36787c0cb2f130015fb50665c8762e598314L76-R126) [[2]](diffhunk://#diff-862eccdc1e989e72b58b23348d7e36787c0cb2f130015fb50665c8762e598314R156-R243)
* Modified the `step` and `decode_offline` methods to leverage CUDA Graphs when enabled, falling back to standard decoding otherwise, and optimizing batch decoding with graph bucket selection. [[1]](diffhunk://#diff-862eccdc1e989e72b58b23348d7e36787c0cb2f130015fb50665c8762e598314L120-R265) [[2]](diffhunk://#diff-862eccdc1e989e72b58b23348d7e36787c0cb2f130015fb50665c8762e598314R340-R388) [[3]](diffhunk://#diff-862eccdc1e989e72b58b23348d7e36787c0cb2f130015fb50665c8762e598314L191-R398)

**Thread Safety and Codec Compatibility:**

* Added locking and surface preparation for CUDA Graphs in codec operations to ensure thread safety and compatibility, including the use of `codec_cuda_graph_capture_lock` and `ensure_codec_decoder_cuda_graph_surface` in various processor and worker routines. [[1]](diffhunk://#diff-3fa7e30e7aba8595743c0ca045b50e50cf8131c8bb2ccdd4868f11e0ae583922R21-R24) [[2]](diffhunk://#diff-3fa7e30e7aba8595743c0ca045b50e50cf8131c8bb2ccdd4868f11e0ae583922R105) [[3]](diffhunk://#diff-3fa7e30e7aba8595743c0ca045b50e50cf8131c8bb2ccdd4868f11e0ae583922R191) [[4]](diffhunk://#diff-3fa7e30e7aba8595743c0ca045b50e50cf8131c8bb2ccdd4868f11e0ae583922L191-R204) [[5]](diffhunk://#diff-3fa7e30e7aba8595743c0ca045b50e50cf8131c8bb2ccdd4868f11e0ae583922R397) [[6]](diffhunk://#diff-862eccdc1e989e72b58b23348d7e36787c0cb2f130015fb50665c8762e598314R18-R31) [[7]](diffhunk://#diff-862eccdc1e989e72b58b23348d7e36787c0cb2f130015fb50665c8762e598314R458-R490)

**Usability and Logging Improvements:**

* Enhanced logging to inform users when CUDA Graphs are enabled, disabled, or being warmed up, and ensured proper fallback and informative messages when CUDA Graphs are not supported on the device. [[1]](diffhunk://#diff-862eccdc1e989e72b58b23348d7e36787c0cb2f130015fb50665c8762e598314R458-R490) [[2]](diffhunk://#diff-862eccdc1e989e72b58b23348d7e36787c0cb2f130015fb50665c8762e598314R501-R507)

These changes collectively improve the efficiency and flexibility of the MOSS-TTS Local streaming vocoder, particularly for users with CUDA Graph-compatible hardware.

## Related Issues

#801 

## Accuracy Test

WIP

## Benchmark & Profiling

WIP

## Checklist

- [x] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/models/moss_tts_local/codec_cuda_graph.py
+++ b/sglang_omni/models/moss_tts_local/codec_cuda_graph.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 import logging
 import math
 import threading
-from contextlib import nullcontext
+from contextlib import contextmanager, nullcontext
 from types import MethodType, SimpleNamespace
 from typing import Any
 
@@ -16,7 +16,6 @@ logger = logging.getLogger(__name__)
 
 codec_cuda_graph_capture_lock = threading.RLock()
 
-_ROPE_CACHE_POSITIONS = 65536
 _ROPE_CACHE_KEY_ATTR = "_sglang_omni_cuda_graph_rope_cache_key"
 _ROPE_CACHE_COS_ATTR = "_sglang_omni_cuda_graph_rope_cos"
 _ROPE_CACHE_SIN_ATTR = "_sglang_omni_cuda_graph_rope_sin"
@@ -75,28 +74,60 @@ def _decoder_rope_head_dims(codec: Any) -> dict[int, int]:
     return head_dims
 
 
+def _decoder_rope_position_multiplier(codec: Any) -> int:
+    """Max decoder sequence upsampling factor seen by any RoPE module."""
+
+    multiplier = 1
+    max_multiplier = 1
+    decoder = getattr(codec, "decoder", ())
+    for decoder_module in decoder:
+        modules = decoder_module.modules() if hasattr(decoder_module, "modules") else ()
+        for module in modules:
+            rope = getattr(module, "rope", None)
+            if rope is not None and hasattr(rope, "max_period"):
+                max_multiplier = max(max_multiplier, multiplier)
+        multiplier *= max(int(getattr(decoder_module, "downsample_ratio", 1) or 1), 1)
+    return max_multiplier
+
+
 def _ensure_rope_cache(
     rope: Any,
     *,
     device: torch.device,
     head_dim: int,
-    positions: int = _ROPE_CACHE_POSITIONS,
+    positions: int | None = None,
 ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
-    cache_key = (str(device), int(head_dim), int(positions), float(rope.max_period))
-    if getattr(rope, _ROPE_CACHE_KEY_ATTR, None) != cache_key:
-        half_dim = int(head_dim) // 2
-        ds = torch.arange(half_dim, device=device, dtype=torch.float32)
-        freqs = torch.exp(ds * (-math.log(float(rope.max_period)) * 2 / int(head_dim)))
-        position_ids_f = torch.arange(positions, device=device, dtype=torch.float32)
-        phase = position_ids_f.view(-1, 1) * freqs.view(1, -1)
-        setattr(rope, _ROPE_CACHE_KEY_ATTR, cache_key)
-        setattr(rope, _ROPE_CACHE_COS_ATTR, torch.cos(phase))
-        setattr(rope, _ROPE_CACHE_SIN_ATTR, torch.sin(phase))
-        setattr(
-            rope,
-            _ROPE_CACHE_POS_ATTR,
-            torch.arange(positions, device=device, dtype=torch.long),
+    requested_positions = max(int(positions or 1), 1)
+    existing_key = getattr(rope, _ROPE_CACHE_KEY_ATTR, None)
+    if (
+        isinstance(existing_key, tuple)
+        and len(existing_key) == 4
+        and existing_key[0] == str(device)
+        and existing_key[1] == int(head_dim)
+        and int(existing_key[2]) >= requested_positions
+        and existing_key[3] == float(rope.max_period)
+    ):
+        return (
+            getattr(rope, _ROPE_CACHE_COS_ATTR),
+            getattr(rope, _ROPE_CACHE_SIN_ATTR),
+            getattr(rope, _ROPE_CACHE_POS_ATTR),
         )
+
+    positions = requested_positions
+    cache_key = (str(device), int(head_dim), int(positions), float(rope.max_period))
+    half_dim = int(head_dim) // 2
+    ds = torch.arange(half_dim, device=device, dtype=torch.float32)
+    freqs = torch.exp(ds * (-math.log(float(rope.max_period)) * 2 / int(head_dim)))
+    position_ids_f = torch.arange(positions, device=device, dtype=torch.float32)
+    phase = position_ids_f.view(-1, 1) * freqs.view(1, -1)
+    setattr(rope, _ROPE_CACHE_KEY_ATTR, cache_key)
+    setattr(rope, _ROPE_CACHE_COS_ATTR, torch.cos(phase))
+    setattr(rope, _ROPE_CACHE_SIN_ATTR, torch.sin(phase))
+    setattr(
+        rope,
+        _ROPE_CACHE_POS_ATTR,
+        torch.arange(positions, device=device, dtype=torch.long),
+    )
     return (
         getattr(rope, _ROPE_CACHE_COS_ATTR),
         getattr(rope, _ROPE_CACHE_SIN_ATTR),
@@ -126,6 +157,7 @@ def _cached_rope_forward(
         self,
         device=q.device,
         head_dim=head_dim,
+        positions=max(time_steps, 1),
     )
     positions = offset.to(device=q.device, dtype=torch.long).view(
         -1, 1
@@ -161,7 +193,9 @@ def _cached_rope_forward(
     return qo.view(*dims, head_dim), ko.view(*dims, head_dim)
 
 
-def patch_codec_rope_for_cuda_graph(codec: Any) -> None:
+def patch_codec_rope_for_cuda_graph(
+    codec: Any, *, cache_positions: int | None = None
+) -> None:
     """Make decoder RoPE graph-capturable by avoiding trig ops in capture."""
 
     try:
@@ -175,7 +209,27 @@ def patch_codec_rope_for_cuda_graph(codec: Any) -> None:
             rope.forward = MethodType(_cached_rope_forward, rope)
         head_dim = head_dims.get(id(rope))
         if head_dim is not None:
-            _ensure_rope_cache(rope, device=device, head_dim=head_dim)
+            _ensure_rope_cache(
+                rope,
+                device=device,
+                head_dim=head_dim,
+                positions=cache_positions,
+            )
+
+
+@contextmanager
+def _use_original_codec_rope(codec: Any):
+    restored: list[tuple[Any, Any]] = []
+    for rope in _decoder_rope_modules(codec):
+        original = getattr(rope, _ROPE_ORIGINAL_FORWARD_ATTR, None)
+        if callable(original):
+            restored.append((rope, rope.forward))
+            rope.forward = original
+    try:
+        yield
+    finally:
+        for rope, current_forward in restored:
+            rope.forward = current_forward
 
 
 def _cuda_graph_update_streaming_cache(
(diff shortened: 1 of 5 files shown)


sglang_omni/models/moss_tts_local/codec_cuda_graph.py
```diff
@@ -0,0 +1,577 @@
...
+    cos_cache, sin_cache, position_offsets = _ensure_rope_cache(
+        self,
+        device=q.device,
+        head_dim=head_dim,
+    )
+    positions = offset.to(device=q.device, dtype=torch.long).view(
+        -1, 1
+    ) + position_offsets[:time_steps].view(1, -1)
+    flat_positions = positions.reshape(-1)
+    rotr = cos_cache.index_select(0, flat_positions).view(
```
Hayden727, line 166: This fixed RoPE cache is too small for the default MOSS-TTS Local decode path. Default `max_new_tokens` is 4096, and the upstream decoder applies five `patch_size=2` upsampling stages before the final transformer, so RoPE positions can reach about 131k; this `index_select` will go out of range before a max-length request finishes. Please size/grow the cache to the required decoder-local position, or fall back before replay, and cover this default-scale case.

tests/unit_test/moss_tts_local/test_streaming_vocoder.py
```diff
@@ -271,6 +461,135 @@ def payload(params: dict[str, Any]) -> StagePayload:
...
+    rope = codec.decoder[0].rope
+    offset = torch.tensor([0, 3])
+    q = torch.randn(2, 2, 5, 4)
+    k = torch.randn(2, 2, 5, 4)
+    expected_q, expected_k = rope(q, k, offset, time_before_heads=False)
+
+    codec_cuda_graph.patch_codec_rope_for_cuda_graph(codec)
+
+    actual_q, actual_k = rope(q, k, offset, time_before_heads=False)
+    torch.testing.assert_close(actual_q, expected_q)
```
JiaxinD: This (and the other component tests) use `torch.testing.assert_close`, which allows a tolerance. For a bit-identity goal, a sub-ULP drift here would pass but break the end-to-end `torch.equal` property. Since the cached RoPE is intended to be bit-exact to `apply_rope`, I would assert `torch.equal` here so the test actually guards bit-identity rather than approximate equality.

sglang_omni/models/moss_tts_local/codec_cuda_graph.py
```diff
@@ -0,0 +1,577 @@
...
+                        setattr(state, attr, snapshot_value.clone())
+                elif snapshot_value is None and isinstance(current_value, torch.Tensor):
+                    if attr == "cached_positions":
+                        current_value.fill_(-1)
+                    else:
+                        current_value.zero_()
+                else:
+                    setattr(state, attr, snapshot_value)
+
+    def _correct_streaming_offsets(
```
JiaxinD, line 608: This correction is correct, but it is worth making explicit that the bucketed-padding path it supports is only safe on terminal steps. The padded (zero-code) frames get K/V written into the sliding window at positions [real_len, chunk_size); after this offset correction the next real step starts at real_len, exactly on top of those padded entries, and the SDPA bias (delta >= 0, pos_k >= 0, delta < context) does not mask them by position. So padding a non-terminal step would make the next step attend to padded K/V and silently produce non-bit-identical audio. Today padding is gated to `final_step=True` followed by a slot reset, so it is safe, but I would add an assert (pad only when the slot resets right after) and cover the padded-final-step in the bit-identity gate so a refactor cannot regress it silently.

sglang_omni/models/moss_tts_local/codec_cuda_graph.py
```diff
@@ -0,0 +1,577 @@
...
+    kor = kr * rotr - ki * roti
+    koi = kr * roti + ki * rotr
+
+    dtype = q.dtype
+    qo = torch.stack([qor.to(dtype), qoi.to(dtype)], dim=-1)
+    ko = torch.stack([kor.to(dtype), koi.to(dtype)], dim=-1)
+    return qo.view(*dims, head_dim), ko.view(*dims, head_dim)
+
+
+def patch_codec_rope_for_cuda_graph(codec: Any) -> None:
```
JiaxinD: This replaces `rope.forward` globally, so the eager fallback path (`_decode_frame` in the except handler) also runs the patched RoPE; it is not a clean fallback to the original. The math looks bit-exact so output parity is fine, but combined with the fixed cache size (see `_ROPE_CACHE_POSITIONS`) it means the eager fallback is not independent of the graph-mode patch. Worth either scoping the patch so the fallback uses the original `apply_rope`, or guarding the cache.

sglang_omni/models/moss_tts_local/codec_cuda_graph.py
```diff
@@ -0,0 +1,577 @@
...
+from types import MethodType, SimpleNamespace
+from typing import Any
+
+import torch
+
+logger = logging.getLogger(__name__)
+
+codec_cuda_graph_capture_lock = threading.RLock()
+
+_ROPE_CACHE_POSITIONS = 65536
```
JiaxinD: This is an unguarded ceiling. The streaming offset grows monotonically, so once `offset + T - 1 >= 65536`, `index_select` goes out of bounds and raises; and because `patch_codec_rope_for_cuda_graph` is global, the eager fallback raises too, so there is no graceful degradation. It is about 87 min of audio at this frame rate so normal TTS will not hit it, but a long or looped stream, or an un-reset slot, would. I would bounds-guard the cached positions (or fall back to the original `apply_rope` beyond the cache).

JiaxinD (commented):
Thanks for picking this up. The stateful direction here is right: the persistent streaming() session plus patching `_update_streaming_cache` to write the cache in place (stable address across replays) is exactly what unblocks graph capture, and capturing via snapshot/restore is a clean way to keep capture from corrupting live state. The cached RoPE math also looks bit-exact to the original `apply_rope`, and the SWA-boundary listening check is a good signal.

The main thing I would want before this lands is an end-to-end bit-identity gate. The current component tests use `torch.testing.assert_close` (tolerance) plus synthetic-tensor checks and address stability, which are necessary but do not prove that on the real codec, with streaming state restored and over multiple chunked steps, the graphed decode is bit-for-bit equal to eager. Listening and WER catch gross artifacts but not a small numeric drift that still sounds fine (I hit exactly that on this path: a degraded decode was only about +5.8% WER and barely audible, yet not bit-identical). Concretely, a gpu-marked test that decodes a multi-chunk sequence eager vs graphed on the real MOSS-Audio-Tokenizer-v2 codec and asserts `torch.equal` per slot at several chunkings and active-slot counts (including the padded-final-step path and the SWA boundary). Happy to share or pair on the harness shape.

Two more: Benchmark is WIP, a paired real-machine before/after (eager vs graph, same branch) would pin the win, and it should be large since this per-step decode is launch-bound. And please confirm the unhappy paths (capture failure, capture OOM, a replay error at serving, an uncaptured shape) each degrade to eager rather than crash, ideally each with a test.

A few specific hardening points are inline. Overall really nice to see this unblocked, thank you.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -179,7 +188,8 @@ def _worker(self) -> None:
             unique_paths = list(dict.fromkeys(path for path, _ in batch))
             results: dict[str, Any] = {}
             try:
-                encoded = self._processor.encode_audios_from_path(unique_paths)
+                with codec_cuda_graph_capture_lock:
```
zhaochenyang20, line 191: Narrow the capture protection to the actual capture window, and use a per-codec or per-device lock for normal reference encodes only when that specific codec instance cannot safely run concurrently with capture. If encode and capture on the same codec must be mutually exclusive, manage the lock through a `WeakKeyDictionary` or device key; do not let a `cuda:0` reference encode block `cuda:1` graph capture, and do not make ordinary data-URI misses share the graph-capture global lock indefinitely.

zhaochenyang20 (changes requested):

sglang_omni/models/moss_tts_local/codec_cuda_graph.py
```diff
@@ -0,0 +1,577 @@
...
+        if callable(original):
+            return original(state, cached_k, cached_v, cached_pos, k_all, v_all, pos_k)
+        raise RuntimeError("CUDA graph codec attention cache is not initialized")
+
+    exec_mask = state.exec_mask.view(-1, 1, 1, 1)
+    exec_mask_pos = state.exec_mask.view(-1, 1)
+    new_cached_k = k_all[:, :, -int(context) :, :].contiguous()
+    new_cached_v = v_all[:, :, -int(context) :, :].contiguous()
+    new_cached_pos = pos_k[:, -int(context) :].contiguous()
+    state_cached_keys.copy_(torch.where(exec_mask, new_cached_k, cached_k))
```
gaoyang07, line 269: One performance concern with this approach: the patched cache update keeps storage stable, but it still materializes and writes back the full sliding-window KV cache on every step/layer:

```python
new_cached_k = k_all[:, :, -context:, :].contiguous()
state_cached_keys.copy_(torch.where(exec_mask, new_cached_k, cached_k))
```

So even though this fixes CUDA graph address stability, the bandwidth cost remains proportional to the full attention context, not just the newly decoded `T` frames. With the MOSS decoder having many streaming attention layers, this could become a significant memory-bandwidth cost.

There is also an extra copy boundary because `quantizer.decode_codes()` stays eager: each replay copies `decoder_hidden_states [B, D, T]` into the graph static input. A fuller codec-native solution would likely use a fixed ring KV buffer and in-place scatter/index-copy only the new K/V frames, and potentially include the quantizer in the graph so the static input is `codes [Nq, B, T]` rather than decoded hidden states.

This implementation is a good correctness-first step, but I would not treat it as the final performance shape unless benchmarks show the full-window KV copy and eager-quantizer boundary are negligible.

sglang_omni/models/moss_tts_local/codec_cuda_graph.py
```diff
@@ -0,0 +1,577 @@
...
+    else:
+        setattr(module, "attention_implementation", attention_implementation)
+
+
+def set_codec_attention_backend_for_cuda_graph(codec: Any) -> None:
+    """CUDA graph mode always uses the SDPA decoder attention path."""
+
+    setter = getattr(codec, "set_attention_implementation", None)
+    if callable(setter):
+        setter("sdpa")
```
gaoyang07, line 302: The SDPA choice still needs further discussion.

gaoyang07:
Merged in #798

PR #813, opened 2026-06-16
Serialize reference encodes at B=1

### LINKED ISSUE #812: [Perf] Fix ref-audio cache first-fill instability from batch-shape-sensitive codec tokens
<html><body>
<html><head></head><body><h2>Description</h2><p>Follow-up to #774. Related Issue: #637 </p><p>The current MOSS-TTS Local reference-audio cache is content-addressed, but the cached value can be produced by an arbitrary batched reference-encoder shape. This is unsafe because MOSS-TTS Local reference encoding appears deterministic for a fixed execution shape, but not invariant across batch/GEMM shapes.</p><p>Current design, simplified:</p><pre><code class="language-text">CachedReferenceEncoder.encode(path)
  -&gt; key = reference_path_cache_key(path)  # content-only file:{hash}
  -&gt; cache hit: return cached tensor
  -&gt; cache miss leader: self._encoder.encode(path)
       where self._encoder is _BatchedReferenceEncoder
       and _BatchedReferenceEncoder may coalesce paths into encode_audios_from_path(unique_paths)
  -&gt; store result under content-only key
</code></pre><p>So the cache key implies:</p><pre><code class="language-text">same audio bytes -&gt; same cached reference codes
</code></pre><p>But #774 and the additional tracing below show that this assumption is not always true under BF16 MOSS-Audio-Tokenizer-v2 execution:</p><pre><code class="language-text">same audio bytes + different effective batch/GEMM shape -&gt; different valid codec tokens
</code></pre><p>This means the cache can become first-fill/order dependent:</p><pre><code class="language-text">first request for A is solo     -&gt; cache[A] = solo codes
first request for A is batched  -&gt; cache[A] = batched-shape codes
</code></pre><p>Both values are stored under the same content-only key, but they may differ in valid codec token IDs.</p><h2>Additional evidence</h2><p>I traced the solo-vs-duplicate-batch difference deeper and localized one concrete divergence to:</p><pre><code class="language-text">encoder.7.input_proj
</code></pre><p>The divergence is not inside:</p><pre><code class="language-text">encoder.7.transformer.layers[*]
</code></pre><p>The input to the diverging linear is identical, but the output differs:</p>

Audio | encoder.7.input_proj.input | encoder.7.input_proj.output
-- | -- | --
female-voice.wav | max diff 0.0 | max diff 0.015625
male-voice.wav | max diff 0.0 | max diff 0.015625

<p>An isolated <code>F.linear</code> comparison using the saved real input and real weights reproduced the same max diff:</p><ul><li><p><code>female-voice.wav</code>: max diff <code>0.015625</code></p></li><li><p><code>male-voice.wav</code>: max diff <code>0.015625</code></p></li></ul><p>So this does not appear to be caused by surrounding transformer logic. It is a shape-sensitive GEMM dispatch/reduction-order effect at <code>encoder.7.input_proj</code>, and the BF16 drift can propagate through quantization and flip final discrete audio codes.</p><p>I also ran a padding-content experiment: zero padding, random padding, and repeat-last padding produced identical valid-region tokens. That rules out padded-content leakage through attention for this setup. The causal mask appears structurally correct; the remaining issue is numeric shape sensitivity, not attention mask leakage.</p><h2>Why this matters for <code>ref-audio-cache</code></h2><p>The current cache design is useful and should be preserved. The codec encoder costs about <code>~0.25 GPU-seconds</code> per reference encode, so avoiding repeated reference encoding is valuable.</p><p>The problem is that the current cache contract is under-specified.</p><p>A content-addressed cache is safe only if the cached value is canonical for the audio content. Today, a cache miss can be filled through <code>_BatchedReferenceEncoder</code>, so the cached value can depend on the batch shape that happened to populate the cache first.</p><p>This affects:</p><ul><li><p>reproducibility across concurrency/request ordering</p></li><li><p>CI parity checks for cache hit vs miss</p></li><li><p>content-addressed cache correctness</p></li><li><p>debugging, because the same reference audio may not map to the same codec tokens across runs with different concurrent neighbors</p></li></ul><p>The issue is not that the cache exists. The issue is that either:</p><pre><code class="language-text">cache value must be canonical
</code></pre><p>or:</p><pre><code class="language-text">cache key must encode the execution shape
</code></pre><p>Today the key is content-only, but the value may be shape-dependent.</p><h2>Expected behavior</h2><p>For a content-addressed reference-audio cache, one of these contracts should hold.</p><h3>Preferred contract: canonical content cache</h3><p>For any cacheable reference audio, the cache value should be produced through one canonical encode path, independent of request concurrency and batching.</p><p>Example desired behavior:</p><pre><code class="language-python"># after clearing cache
canonical = encode_ref_audio_canonical(path_a)

# concurrent duplicate requests may arrive together
out0 = cached_encoder.encode(path_a)
out1 = cached_encoder.encode(path_a)

assert torch.equal(out0, canonical)
assert torch.equal(out1, canonical)
</code></pre><p>Then the content-only key is valid because the value is explicitly canonicalized.</p><p>The canonical path should be documented as:</p><pre><code class="language-text">content key -&gt; canonical B=1 reference-audio codes
</code></pre><p>not:</p><pre><code class="language-text">content key -&gt; whichever batched encoder shape populated the cache first
</code></pre><h3>Important implementation constraint</h3><p><code>encode_canonical(path)</code> must bypass <code>_BatchedReferenceEncoder</code> entirely.</p><p>It should not be implemented as:</p><pre><code class="language-python">_BatchedReferenceEncoder.encode(path)
</code></pre><p>or as a single-element batched call such as:</p><pre><code class="language-python">encode_audios_from_path([path])
</code></pre><p>or any equivalent path that still goes through batching, <code>unique_paths</code> coalescing, stacking, padding, batch flattening, or scheduler batch logic.</p><p>The canonical cache-fill path should use the true solo encode route: one reference audio, one canonical waveform/tensor, one stable B=1 execution path. Otherwise, an implementation could still accidentally produce a value through a different reshaping/layout path than the real solo encode, which would not fix the cache contract.</p><h2>Actual behavior</h2><p>A content-only key is used, but misses are filled by <code>_BatchedReferenceEncoder</code>, which may batch multiple requests before calling <code>encode_audios_from_path(unique_paths)</code>.</p><p>Therefore the first cache fill can be solo-shaped or batch-shaped depending on timing/concurrency. Since #774 and the additional trace show solo-vs-batch codec tokens can differ, the same audio content can populate the same cache key with different valid codec tokens.

<h2>Why more experiments are needed</h2><p>

The best cache design depends on the actual serving workload.

The important variables are:

ref audio encode cost:   ~0.25 GPU-seconds per audio
cache hit cost:          ~0, memory lookup
typical ref pool size:   likely small, reused heavily

If the workload uses a small pool of reference voices, then the dominant performance variable is cache hit rate, not miss-fill throughput.

For example:

10 reference voices
10,000 requests
encode cost ~= 0.25 GPU-seconds per unique voice

A canonical content cache pays roughly:

10 * 0.25 = 2.5 GPU-seconds total

Without an effective cache, the system may pay:

10,000 * 0.25 = 2,500 GPU-seconds total

In this regime, whether the 10 initial cache fills are solo or batched barely matters for steady-state throughput. What matters much more is:

same content -> one stable cache entry -> high hit rate

So we should not optimize the cache design only for miss-fill batching. We should first measure whether cold misses are actually a meaningful bottleneck in realistic workloads.<h2>Options considered</h2><h3>Option 1: Canonical content cache with true solo miss fill</h3><p>Keep the cache key content-addressed, but require cache population to use a canonical solo encode path.</p><pre><code class="language-text">file:{content_hash} -&gt; encode_canonical(path)
</code></pre><p>Pros:</p><ul><li><p>Preserves the current high cache hit rate.</p></li><li><p>Makes the content-only key semantically correct.</p></li><li><p>Removes first-fill/order dependence.</p></li><li><p>Makes cache hit/miss parity easier to test.</p></li><li><p>Keeps behavior reproducible across concurrency patterns.</p></li><li><p>Limits the fix to cache population; uncached/batched paths can still exist separately.</p></li><li><p>Avoids trying to encode every cuBLASLt/backend dispatch detail into the key.</p></li></ul><p>Cons:</p><ul><li><p>Cache misses are not batched.</p></li><li><p>First-time reference encoding may be slower under high concurrency.</p></li><li><p>Requires a clear separate canonical encode function and regression tests.</p></li><li><p>If many unique references arrive once and never repeat, this sacrifices some possible batched miss throughput.</p></li></ul><h3>Option 2: Shape-specific cache key</h3><p>Keep batched miss fill, but include execution-shape metadata in the key.</p><p>Example:</p><pre><code class="language-text">file:{content_hash}:dtype:{dtype}:device:{device}:shape:{B,T,B*T}:backend:{backend}
</code></pre><p>Pros:</p><ul><li><p>Preserves exact batched-encoder semantics.</p></li><li><p>Allows batched cache fills.</p></li><li><p>Avoids forcing solo execution on cache misses.</p></li></ul><p>Cons:</p><ul><li><p>Fragments the cache and reduces hit rate.</p></li><li><p>Makes keys much more complex.</p></li><li><p>Hard to define a complete key: cuBLASLt algorithm choice, CUDA/PyTorch version, hardware, dtype, layout, and backend may all matter.</p></li><li><p>Still may not fully capture all dispatch differences.</p></li><li><p>Makes reproducibility harder to reason about.</p></li><li><p>Same content can intentionally have multiple cached values, which weakens the value of a content-addressed cache.</p></li></ul><h3>Option 3: Disable caching for batched reference encoding</h3><p>Only use cache for known solo paths; bypass cache for batched paths.</p><p>Pros:</p><ul><li><p>Simple correctness rule.</p></li><li><p>Avoids storing arbitrary batched values under content-only keys.</p></li><li><p>Low implementation complexity.</p></li></ul><p>Cons:</p><ul><li><p>Loses many cache benefits under concurrent serving.</p></li><li><p>Duplicate reference audio in concurrent requests may still pay repeated encode cost.</p></li><li><p>Does not provide a single clean cache contract for all cacheable reference audio.</p></li><li><p>More conditional behavior to reason about.</p></li></ul>
<h2>Environment</h2><p>Observed while debugging MOSS-TTS Local reference encoding:</p><ul><li><p>Model: <code>OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5</code></p></li><li><p>Codec/tokenizer: <code>OpenMOSS-Team/MOSS-Audio-Tokenizer-v2</code></p></li><li><p>Hardware: NVIDIA H100</p></li><li><p>dtype: BF16 codec execution</p></li><li><p>Attention path during debugging: SDPA</p></li><li><p>Relevant related issue: #774</p></li><li><p>Relevant current cache implementation: #743 / <code>CachedReferenceEncoder</code> in <code>sglang_omni/models/moss_tts_local/stages.py</code></p></li></ul><p>I can add the exact PyTorch/CUDA build from the latest trace run if needed.</p></body></html>
</body>
</html>

AkazaAkane:
## Motivation

#812 reports that the MOSS-TTS Local reference-audio cache can become first-fill/order dependent: the cache key is content-addressed, but codec token output can vary when the same audio is encoded under different batch shapes. Cacheable reference fills therefore need a canonical single-reference encode path instead of inheriting batch/coalescing semantics from `_BatchedReferenceEncoder`.

## Modifications

- Keep `_BatchedReferenceEncoder` serialized per unique path for non-cacheable/unwrapped reference encodes, avoiding cross-file `encode_audios_from_path(...)` batches.
- Add `_CanonicalReferenceEncoder`, a separate serialized worker used for cache fills. It decodes file refs to a single waveform and encodes through `encode_audios_from_wav([wav], sample_rate)`, bypassing `_BatchedReferenceEncoder.encode()` for cacheable file misses.
- Wire `CachedReferenceEncoder` to use the canonical worker on cacheable file misses while preserving single-flight dedup, LRU storage, TOCTOU revalidation, and CPU `torch.long` return semantics.
- Route matching data-URI refs through the same canonical worker when it owns the processor, while keeping the existing `bytes:` keyspace separate from `file:` keys.
- Strengthen focused cache tests to use real temp files for cacheable paths, replace the barrier-based single-flight test with event/counter synchronization, and add a shape-sensitive fake processor regression that proves cache misses do not use the batched path.

## Related Issues

Closes #812

## Accuracy Test

Device: 1×H100 GPU, hongccc/sglang-omni:dev, c = 8, batch = 8
The SeedTTS table should be read as an end-to-end regression check, not a miss-path throughput benchmark. The main claim of this PR is correctness: cacheable cold fills now use a canonical B=1 path, and cache hits remain unchanged. The latency/QPS numbers are side observations from the same eval, not a claimed performance improvement.

For miss-path performance, the expected tradeoff is explicit: distinct cold misses are serialized instead of batched, while repeated refs still benefit from single-flight and cache hits. A targeted cold-miss microbench would be the right benchmark if we want to quantify that tradeoff separately.

Full SeedTTS eval completed for both branches and both languages:

| Branch | Lang | Requests | Failed | WER corpus | Lat mean | Lat p95 | RTF mean | QPS |
| PR | EN | 1088 | 0 | 1.20% | 12.127s | 16.514s | 3.2958 | 1.311 |
| main | EN | 1088 | 0 | 1.19% | 12.440s | 17.043s | 3.3728 | 1.279 |
| PR | ZH | 2020 | 0 | 0.83% | 12.742s | 16.091s | 2.4157 | 1.252 |
| main | ZH | 2020 | 0 | 0.90% | 11.709s | 14.979s | 2.2158 | 1.362 |

Summary:

- EN: PR is slightly faster, about +2.5% QPS and -2.5% mean latency, with effectively unchanged WER: 1.20% vs 1.19%.
- ZH: PR improves WER, 0.83% vs 0.90%, but is slower: -8.1% QPS and +8.8% mean latency.

Cache-correctness comparison:

| Branch | Concurrent cold miss A+B | Cached value for A |
|---|---|---|
| main | batched together | batch-shaped value `[2]` |
| PR | serialized B=1 fills | canonical solo value `[1]` |

Old-main reproducer:

```text
output A/B: [2] [2]
cache hit A: [2]
processor calls: [(2, ('B.wav', 'A.wav'))]
```

PR shape-sensitive test:

```text
out_a: [1]
out_b: [1]
cache hit A: [1]
processor call sizes: all 1
```

These cache-specific tests support the PR's claimed benefit: the PR fixes concurrent cold-fill cache pollution where the first cached reference encoding could depend on what other reference audio arrived in the same miss window.

## Benchmark & Profiling

See SeedTTS latency/RTF/QPS results in the accuracy section above. Cold cacheable reference fills now use the serialized canonical fill path; cache hits remain unchanged.

## Tests

- `tests/unit_test/moss_tts_local/test_ref_audio_cache.py`: 9 passed
- Full unit suite: 1329 passed
- Real MOSS tokenizer cache parity test: 1 passed

## Checklist

- [x] Add unit tests.

## CI


sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -270,8 +262,54 @@ def _worker(self) -> None:
...
+        ] = queue.Queue()
+        self._thread = threading.Thread(
+            target=self._worker, name="moss-local-ref-canonical-encode", daemon=True
+        )
+        self._thread.start()
+
+    def encode_file(self, path: str) -> torch.Tensor:
+        import soundfile as sf
+
+        audio, sample_rate = sf.read(path, dtype="float32", always_2d=True)
```
JiaxinD: This sf.read + duration check + audio.T block duplicates the decode in encode_data_uri._encode (481-495); a shared _decode_to_wav(src) helper would keep the two from drifting.

tests/unit_test/moss_tts_local/test_ref_audio_cache.py
```diff
@@ -0,0 +1,295 @@
...
+
+        def encode_audios_from_path(self, paths: list[str]) -> list[torch.Tensor]:
+            self.path_calls.append(len(paths))
+            return [torch.tensor([2], dtype=torch.long) for _ in paths]
+
+        def encode_audios_from_wav(
+            self, wavs: list[torch.Tensor], sample_rate: int
+        ) -> list[torch.Tensor]:
+            self.wav_calls.append(len(wavs))
+            return [torch.tensor([1], dtype=torch.long) for _ in wavs]
```
JiaxinD, line 302: This fake returns a constant [1] ignoring wavs/sample_rate and the refs are mono 16 kHz, so the test guards dispatch to the wav path but not the stereo or non-target sample-rate decode parity the canonical switch depends on. Is that parity covered by the real-tokenizer parity test?

Ccyest:
what parameters are used when testing the benchmark?

tests/unit_test/moss_tts_local/test_ref_audio_cache.py
Ccyest, line 1: feel like only test e2e encoder behavior catch should be enough. Testing helpers are bit redundant here.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -318,9 +367,14 @@ def encode(self, path: str) -> torch.Tensor:
         key = _reference_path_cache_key(path)
         if key is None:
             return self._encoder.encode(path)  # uncacheable (URL/missing) -> bypass
+        encode_fn = lambda: (
+            self._canonical_encoder.encode_file(path)
+            if self._canonical_encoder is not None
+            else self._encoder.encode(path)
+        )
         return self._cached_encode(
```
Ccyest, line 375: is canonical encoder a strictly better way to encode than the encoder we already have?
if so why writing another Canonical encoder class instead of re-writing CachedReferenceEncoder to use the canonical encoding?

sglang_omni/models/moss_tts_local/stages.py
Ccyest, line 237: if encode one by one, this no longer useful, and become dead code.

Ccyest:
the change has theoretical improvement on performance but still needs more solid benchmark data to support.

sglang_omni/models/moss_tts_local/stages.py
AkazaAkane: `_drain_batch` is still load-bearing for uncacheable paths (URL refs, stat-unreachable files), which bypass `_CanonicalReferenceEncoder` and still route through `_BatchedReferenceEncoder`. Within the batch window, lines 242–262 deduplicate concurrent requests for the same uncacheable path into a single encode call. The change was to encode each *unique* path individually (line 246: `encode_audios_from_path([path])[0]`) instead of stacking them into a B>1 call — the queue and dedup logic itself is still needed.

sglang_omni/models/moss_tts_local/stages.py
AkazaAkane: `_CanonicalReferenceEncoder` is not strictly better for all paths, so collapsing the two classes would lose precision. `_BatchedReferenceEncoder` is still used for uncacheable refs (URL refs, stat-unreachable files) where no stable content key exists and shape-stability of the result does not matter. Forcing canonical (wav-level decode + B=1 serialize) on those paths would add unnecessary overhead with no correctness benefit. `_CanonicalReferenceEncoder` only engages on cacheable file misses, where the canonical guarantee is required. Keeping them separate makes the two contracts explicit.

tests/unit_test/moss_tts_local/test_ref_audio_cache.py
AkazaAkane: The `_BatchedReferenceEncoder` unit tests are not redundant with the e2e `CachedReferenceEncoder` tests because the two paths cover different behaviors. The cached encoder tests use `_FakeCanonicalEncoder` for cache misses, so they never exercise the batched encoder path. The batched encoder is still the live path for uncacheable refs (URLs, missing files), and `test_batched_encoder_always_uses_b1_per_path` is the core regression test for the bug this PR fixes — it directly verifies that concurrent encodes never coalesce into a B>1 call. `test_decode_reference_to_wav_preserves_stereo_and_sample_rate` also covers a boundary in the shared decode helper that the fake-processor tests cannot reach.

tests/unit_test/moss_tts_local/test_ref_audio_cache.py
AkazaAkane: Stereo and non-target sample-rate parity is covered by two tests:

1. `test_decode_reference_to_wav_preserves_stereo_and_sample_rate` (line 67) directly tests `_decode_reference_to_wav` with a real 22050 Hz stereo file via both file path and `BytesIO`, asserting the shape is `(2, 160)` and sample rate is preserved.
2. `test_cache_miss_uses_canonical_wav_encoder_not_batched_path` (line 278) writes real soundfile-backed stereo 22050 Hz refs and asserts `wav_shapes == [(2, 160), (2, 160)]` and `sample_rates == [22050, 22050]` reach the processor.

`_FakeProcessor.encode_audios_from_wav` returning a constant is intentional for tests that only need to verify dispatch routing (canonical vs batched path), not decode fidelity.

AkazaAkane:
@Ccyest On the benchmark questions: this PR is a correctness fix for the cache first-fill instability described in #812, not a performance optimization. The SeedTTS eval in the description is a regression check — the primary metric is WER (1.20% vs 1.19% EN; 0.83% vs 0.90% ZH), confirming no quality degradation. The latency/RTF/QPS numbers are side observations from that same run, not a perf claim.

On parameters: the eval used the standard SeedTTS test set (1088 EN / 2020 ZH requests) with default server config (Device: 1×H100 GPU, hongccc/sglang-omni:dev, c = 8, batch = 8) and a warm cache (all ref audio pre-encoded). The ZH slowdown is an artifact of cold canonical fills now being serialized one-at-a-time rather than coalesced, which is expected and the intended behavior for a correctness fix.

For throughput-focused benchmarking of the miss path, the relevant context from #812 is that realistic workloads use a small reused ref pool, so cold miss cost (~0.25 GPU-s per unique voice) is paid once regardless of batching. Happy to run a targeted miss-path throughput benchmark if that would help.

AkazaAkane:
# PR Test Result: Reference Cache Correctness + SeedTTS Stability

Repos / artifacts:

```text
Main:        sglang-omni-main
PR #813:    sglang-omni-pr813
1*H100 GPU, hongccc/sglang-omni:dev
```

## Experiment 1: Token-Level Cache Correctness

### Purpose

This is the primary correctness test.

For each reference audio, I compare the cached reference tokens against canonical solo B=1 tokens.

The test checks both:

1. **Miss-fill value**
   - What value is written into cache on first fill?

2. **Cache-hit value**
   - What value is replayed after the cache is warm?


### Controls

| Control | Reason |
|---|---|
| Cache cleared before each config | Ensures each batch setting performs its own first fill |
| Canonical solo B=1 tokens computed separately | Provides stable reference baseline |
| First fills use unique refs | Avoids hiding the issue through duplicate single-flight merging |
| Same 16 refs used across branches/configs | Makes main vs PR comparison paired and directly comparable |
| Miss-fill and hit values both checked | Distinguishes transient encode drift from persisted cache corruption |

### Token Correctness Result

| Branch | Configured ref batch sizes | Refs checked | Refs mismatching canonical solo B=1 | Miss-fill mismatch | Cache-hit mismatch | Interpretation |
| main | B=1 | 16 | 0 / 16 | No | No | Solo path is canonical |
| main | B=2 / B=4 / B=8 | 16 | 15 / 16 | Yes | Yes | Batch-shaped values can be stored and replayed |
| PR #813 | B=1 / B=2 / B=4 / B=8 | 16 | 0 / 16 | No | No | Cacheable fills are batch-invariant |

### Cache Parity

| Branch | Configs | Cache hits | Cache misses | Merged events in token-check path | Result |
|---|---|---:|---:|---:|---|
| main | B=1 / B=2 / B=4 / B=8 | 16 | 16 | 0 | Clean cache accounting |
| PR #813 | B=1 / B=2 / B=4 / B=8 | 16 | 16 | 0 | Clean cache accounting |

---

## Experiment 2: End-to-End SeedTTS Stability

### Purpose

This test checks whether the token-level cache correctness issue is visible in end-to-end SeedTTS quality/stability.

The expected improvement is not necessarily lower average WER. The expected improvement is that same logical inputs should be less sensitive to ref-encoder batch size / first-fill schedule. So the key quality metric is **cross-batch variance**, not only mean WER.

### Dataset

The locally available SeedTTS EN split had at most 2 prompts per unique reference audio, so I used:

```text
16 unique refs × 2 prompts/ref = 32 requests
3 repeats
client concurrency = 16
configured ref batch sizes = B=1 / B=2 / B=4 / B=8
```

This is a useful correctness/stability benchmark, but it is not a perfect high-reuse serving simulation because each reference audio is only reused twice.

### Observed Ref-Encoder Batch Sizes

| Branch | Configured B=1 | Configured B=2 | Configured B=4 | Configured B=8 | Interpretation |
|---|---:|---:|---:|---:|---|
| main | `{1: 48}` | `{2: 24}` | `{4: 12}` | `{8: 6}` | Main exercised actual batched ref-encode paths |
| PR #813 | `{1: 48}` | `{1: 48}` | `{1: 48}` | `{1: 48}` | PR forces canonical B=1 cacheable fills |
| PR #813 mode | canonical_wav | canonical_wav | canonical_wav | canonical_wav | Expected behavior |

This confirms the benchmark actually isolated the PR behavior: main used batched reference encoding, while PR #813 used canonical B=1 encoding for cacheable fills.

---

## End-to-End Quality Results

| Metric | main | PR #813 | Relative change | Interpretation |
|---|---:|---:|---:|---|
| Cross-batch WER std | 0.00381 | 0.00237 | -37.8% | PR reduces WER variance across batch configs |
| Cross-batch WER std, percentage points | 0.381 pp | 0.237 pp | -37.8% | Same result in pp units |
| Mean WER across B=1/2/4/8 | ~1.15% | ~1.17% | roughly neutral | No meaningful average WER improvement |
| Speaker similarity cross-batch std | 0.398 | 0.395 | roughly neutral | Speaker SIM stability is essentially unchanged |

Result is that cross-batch WER variance decreased from `0.00381` to `0.00237`, about a **37.8% reduction**.

---

## Performance Sanity Check

### Performance Result

| Metric | main | PR #813 | Relative change | Interpretation |
|---|---:|---:|---:|---|
| Avg p95 latency | 7.607s | 9.740s | +28.0% | Visible cold-miss tail-latency cost |

Avg QPS and Avg latency looks ok. But we observe p95 latency goes up. This is the expected tradeoff. On `main`, cache miss fills can be processed through batched reference encoding. On PR #813, cacheable miss fills are forced through canonical B=1 encoding. Therefore, cold-miss-heavy workloads can show higher p95 latency.

---

## Summary Table

| Question | Result | Conclusion |
|---|---|---|
| Does main produce batch-dependent reference tokens? | Yes. B=2/4/8 had 15/16 refs mismatching canonical solo B=1 tokens. | The issue is reproducible. |
| Does the mismatch persist through cache hits? | Yes. Main mismatches appeared on both miss-fill and cache-hit reads. | Batch-shaped values can be stored and replayed. |
| Does PR #813 fix token correctness? | Yes. B=1/2/4/8 had 0/16 mismatches. | PR makes cacheable fills batch-invariant. |
| Does average WER improve? | No meaningful improvement. Main ~1.15%, PR ~1.17%. | Mean quality is neutral. |
| Does WER stability improve? | Yes. Cross-batch WER std decreased from 0.00381 to 0.00237. | About 37.8% lower WER variance across batch configs. |
| Does speaker similarity change? | No meaningful change. Std 0.398 vs 0.395. | Speaker SIM is neutral. |

PR #816, opened 2026-06-17
[Higgs] Fuse top-k/top-p sampling with sgl_kernel renorm

BBuf:
## Motivation

Profiling Higgs Audio v3 TTS decode on H100 showed the per-step multi-codebook sampler running a **full-vocab `torch.sort` + `topk(K_MAX)` + cumsum/scatter logit masking** every decode step (`radixSortKVInPlace`, `sbtopk::gatherTopK`, `softmax_warp_forward` kernels). For the codebook vocab (V=1026, 8 codebooks) this is a chain of small launch-bound kernels.

## Change

In `_sample_independent_batched`, replace the sort-based top-k/top-p logit masking with the fused `sgl_kernel.top_k_renorm_prob` / `top_p_renorm_prob` kernels, keeping the existing graph-safe `multinomial` and the greedy/argmax short-circuit unchanged.

## Correctness (numerically equivalent)

Verified against the original sort-based path over a sweep of `B ∈ {1,4,16} × temperature ∈ {0.7,1.0,1.3} × top_k ∈ {None,5,50,200,1026} × top_p ∈ {None,0.8,0.9,0.95,1.0}` (225 configs):

- **max abs probability difference = 4.8e-7** (fp32 epsilon)
- **identical support** (zero tokens differ in the kept set)

The only theoretical difference is at an exact `cumsum == top_p` boundary, where the fused path uses the standard nucleus (`cum >= p`) convention vs the prior `cum > p` + shift; this is measure-zero for continuous logits and never triggered in the sweep. Inputs are forced **contiguous fp32** as required by the flashinfer renorm kernels (a non-contiguous tensor silently corrupts the result).

## Performance

Higgs decode A/B on a single H100 (n=2 each, fixed request, GPU time per decode step from torch profiler, normalized by step count):

| | µs/step |
|---|---|
| baseline | 4152.1 |
| this PR | 4082.8 |

**−1.67% per-step GPU time.** The sampler kernel block (radixSort + gatherTopK + softmax) drops from ~64 to ~32 µs/step. CUDA graph capture and audio output are unaffected (verified end-to-end: valid 24 kHz speech, no graph break).

### DIFF
--- a/sglang_omni/models/higgs_tts/sampler.py
+++ b/sglang_omni/models/higgs_tts/sampler.py
@@ -13,6 +13,8 @@ from __future__ import annotations
 from dataclasses import dataclass
 
 import torch
+from sgl_kernel import top_k_renorm_prob as _fused_top_k_renorm
+from sgl_kernel import top_p_renorm_prob as _fused_top_p_renorm
 
 from sglang_omni.models.higgs_tts.utils import BOC_ID, EOC_ID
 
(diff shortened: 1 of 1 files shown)


BBuf:
### Seed-TTS A/B (generate + transcribe) — quality is unchanged

Ran the full Seed-TTS **English** eval set (1088 samples) before/after this PR, using the **same ASR for both** so the delta is attributable to the sampler change only.

| | samples | WER (corpus) | WER (per-sample mean) | >50% WER | failed |
| `main` (sort-based sampler) | 1088 | **0.879%** | 0.850% | 0 | 0 |
| this PR (fused renorm) | 1088 | **0.904%** | 0.887% | 0 | 0 |
| Δ | — | **+0.025 pp** | +0.038 pp | 0 | 0 |

**Conclusion: WER is statistically unchanged** (0.88% vs 0.90%, Δ 0.025 pp) — within run-to-run sampling noise (both are independent stochastic generations at temperature > 0). This is consistent with the numerical-equivalence check from the PR description: the renorm path reproduces the sort path's post-top-k/top-p distribution to ~5e-7 with identical support across a temperature × top_k × top_p sweep, so the fused kernels do not alter the sampling distribution.

> Note: I used `openai/whisper-large-v3` as the transcriber (with `--asr-concurrency 1`, since the whisper encoder path only supports batch size 1 here) instead of the default Qwen3-ASR, so the **absolute** WER differs from CI's official numbers — but baseline and PR use the same ASR, so the **delta** is valid. CI will produce the official Qwen3-ASR numbers.

#### Reproduction (single H100)

```bash
# 0. dataset
python -m benchmarks.dataset.prepare --dataset seedtts

# Do the following twice: once on `main`, once on this PR branch.

# 1. start a higgs server (local ref-audio needs --allowed-local-media-path)
sgl-omni serve --model-path bosonai/higgs-audio-v3-tts-4b \
  --allowed-local-media-path /tmp --port 8150

# 2. generate (TTS) against the running server
python -m benchmarks.eval.benchmark_tts_seedtts --generate-only --use-existing-server \
  --base-url http://127.0.0.1:8150 \
  --meta zhaochenyang20/seed-tts-eval-arrow --lang en --ref-format references \
  --model bosonai/higgs-audio-v3-tts-4b \
  --output-dir results/<variant> --max-concurrency 8 --skip-gpu-cleanup

# 3. transcribe + WER (auto-starts the ASR server)
python -m benchmarks.eval.benchmark_tts_seedtts --transcribe-only \
  --asr-model-path openai/whisper-large-v3 --asr-concurrency 1 \
  --meta zhaochenyang20/seed-tts-eval-arrow --lang en \
  --model bosonai/higgs-audio-v3-tts-4b \
  --output-dir results/<variant> --skip-gpu-cleanup
```

Ratish1 (approved):

PR #818, opened 2026-06-17
Improve Qwen3-Omni thinker text-only decode throughput on H100

luojiaxuan:
## Summary

- Add env-gated per-step/build profiling for the Qwen3-Omni thinker decode path.
- Add an env-gated text-only decode MRoPE fast path that reuses decode positions instead of rebuilding/copying tiny CPU MRoPE tensors each step.
- Treat `MultimodalInputs(mm_items=[])` with zero `mrope_position_delta` as text-equivalent; nonzero delta and real multimodal inputs still use the original path.

## H100 Findings

Fixed-concurrency diagnostic result, before the full SLO-constrained throughput sweep:

- Setup: Qwen3-Omni thinker, BF16, TP=2, H100 GPU4/5 NV18, c32 text-only decode, `thinker_memory_fraction=mem_fraction_static=0.55`, CUDA graph hit rate 1.000.
- MRoPE text-only decode fast path:
  - Throughput: 3858.4 -> 4085.7 tok/s (+5.9%).
  - Step: 7.68 -> 7.30 ms.
  - Build: 0.61 -> 0.26 ms.
  - MRoPE: 0.37 -> 0.03 ms.
  - Forward stayed flat: ~6.45 -> ~6.43 ms.
- Companion custom all-reduce result is covered by #783, not reimplemented in this PR:
  - NCCL stable control: ~4034 tok/s.
  - custom-AR: 4440.6 tok/s mean / 4466.1 tok/s median (+10-11%).
  - Step/forward: ~7.31/6.45 ms -> ~6.68/5.85 ms.

## Metric Framing

The target metric should be SLO-constrained throughput, not raw tok/s or standalone latency. This PR contributes a low-risk host/build reduction that improves fixed-c32 decode throughput while preserving CUDA graph hit rate. The next benchmark should add p95 ITL/TTFT and a c16/c24/c32/c40/c48 sweep to report max sustainable tok/s under a fixed streaming latency budget.

## Validation

- `PYTHONPYCACHEPREFIX=/tmp/sglang-omni-pycache python3 -m py_compile sglang_omni/model_runner/base.py tests/unit_test/model_runner/test_base_hooks.py`
- `git diff --check`
- Container smoke test for MRoPE zero-delta/nonzero-delta behavior.
- H100 c32 A/B logs:
  - `/data/luojiaxuan/h100-host-scheduler-runs/20260617-mrope-ab2/`
  - `/data/luojiaxuan/h100-host-scheduler-runs/20260617-custom-ar-ab/`

Note: host-side `python3 -m pytest tests/unit_test/model_runner/test_base_hooks.py -q` could not run because `pytest` is not installed in the host Python environment.

### DIFF
--- a/sglang_omni/model_runner/base.py
+++ b/sglang_omni/model_runner/base.py
@@ -7,6 +7,7 @@ pass, sampling, logit post-processing, and output extraction.
 from __future__ import annotations
 
 import logging
+import threading
 import time
 from dataclasses import dataclass
 from typing import Any
@@ -26,6 +27,7 @@ from sglang_omni.scheduling.types import (
 from sglang_omni.utils.env import env_flag
 
 logger = logging.getLogger(__name__)
+_FORWARD_BATCH_MROPE_PATCH_LOCK = threading.Lock()
 
 
 def _current_sglang_sampling_backend() -> str | None:
@@ -112,6 +114,8 @@ class ModelRunner:
         self._mrope_patch_enabled: bool = self._build_profile_on or env_flag(
             "SGLANG_OMNI_FAST_MROPE_DECODE"
         )
+        if self._mrope_patch_enabled:
+            self._install_forward_batch_mrope_patch()
 
     def _next_host_staging(self, device_staging: torch.Tensor) -> torch.Tensor:
         """Return a pinned host staging buffer mirroring ``device_staging``'s
@@ -325,9 +329,6 @@ class ModelRunner:
             ForwardBatch,
         )
 
-        if bool(getattr(self, "_mrope_patch_enabled", False)):
-            self._patch_forward_batch_mrope(ForwardBatch)
-
         if self.device.type == "cuda":
             torch.cuda.set_device(self.device)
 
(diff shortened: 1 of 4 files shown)


sglang_omni/model_runner/base.py
```diff
@@ -127,13 +148,24 @@ def execute(self, scheduler_output: Any) -> ModelRunnerOutput:
...
+        t3 = time.perf_counter() if prof else 0.0
+        out = self._finalize(
             batch_result,
             forward_batch,
             schedule_batch,
             model_worker_batch,
             scheduler_output,
         )
+        if prof:
+            self._last_phase = {
```
yxs, line 189: `_last_phase` is written only by the synchronous `execute()`; the async-decode `execute_launch` / `execute_resolve` path never sets it. Once `_event_loop_overlap` (which calls `_record_step_phase`) runs the async path, `_record_step_phase` always falls into the `last_phase is None` else-branch (after `omni_scheduler.py`) — build/forward/post/finalize all record 0, the whole step time lands in `stream`, yet `[step phases]` still prints `gpu=<forward>=0.00 ... host_post=...`. So anyone profiling the real overlap/async serving path gets wrong GPU/host attribution (forward=0 doesn't mean forward was free).

sglang_omni/model_runner/base.py
```diff
@@ -75,6 +77,16 @@ def __init__(self, tp_worker: Any, output_processor: Any):
...
         self._async_query_miss: int = 0
+        self._phase_on: bool = bool(os.environ.get("SGLANG_OMNI_PHASE_PROFILE"))
+        self._phase_sync: bool = bool(os.environ.get("SGLANG_OMNI_PHASE_SYNC"))
+        self._last_phase: dict[str, Any] | None = None
+        self._build_profile_on: bool = bool(
+            os.environ.get("SGLANG_OMNI_BUILD_PROFILE")
+        )
+        self._last_build_phase: dict[str, float] | None = None
+        self._mrope_patch_enabled: bool = self._build_profile_on or bool(
+            os.environ.get("SGLANG_OMNI_FAST_MROPE_DECODE")
```
yxs: `bool(os.environ.get("..."))` is True for any non-empty string, including `"0"` and `"false"`. So `SGLANG_OMNI_FAST_MROPE_DECODE=0` does **not** disable the fast path — and this is a user-facing toggle that changes the mrope path, exactly the case where someone A/B-testing or debugging would set `=0`. The idiom appears ~10× across the two files.

sglang_omni/model_runner/base.py
```diff
@@ -261,12 +299,100 @@ def _build_forward_batch(self, scheduler_output: Any):
...
+                "mrope": float(
+                    getattr(forward_batch, "_sglang_omni_mrope_seconds", 0.0)
+                ),
+                "mrope_fast": float(
+                    bool(getattr(forward_batch, "_sglang_omni_mrope_fast", False))
+                ),
+            }
         return forward_batch, schedule_batch, model_worker_batch, is_prefill
 
+    def _patch_forward_batch_mrope(self, forward_batch_cls: Any) -> None:
```
yxs, line 383: `_patch_forward_batch_mrope` replaces upstream `sglang.srt.model_executor.forward_batch_info.ForwardBatch._compute_mrope_positions` (a core class omni doesn't own). Both early-returns are **silent**: `if not hasattr(...): return` and the sentinel short-circuit. If upstream renames or re-signatures the method, `SGLANG_OMNI_FAST_MROPE_DECODE=1` becomes "enabled but inert" with no log — a hard-to-diagnose silent perf regression on the next sglang bump.

sglang_omni/model_runner/base.py
luojiaxuan: Done in c939967: async launch/resolve now carries phase timings through `pending.phase` and writes `_last_phase` after finalize.

sglang_omni/model_runner/base.py
luojiaxuan: Done in c939967: replaced boolean env checks with `env_flag`, so values like `0`/`false`/`off` are parsed as disabled.

sglang_omni/model_runner/base.py
luojiaxuan: Done in c939967: missing upstream MRoPE hooks now warn once, and the already-installed path logs at debug level.

luojiaxuan (commented):
Reviewed. The profiling/MRoPE fast path is env-gated; no safe simplification found without changing diagnostic behavior.

sglang_omni/model_runner/base.py
```diff
@@ -254,6 +309,12 @@ def execute_resolve(
...
             skip_rids=skip_rids,
         )
+        if pending.phase is not None:
+            phase = dict(pending.phase)
+            phase["post"] = phase["post"] + (t_finalize0 - t_post0)
+            phase["finalize"] = time.perf_counter() - t_finalize0
+            self._last_phase = phase
+        return out
 
     def _build_forward_batch(self, scheduler_output: Any):
```
yxs, line 323: `_build_forward_batch` calls `self._patch_forward_batch_mrope(ForwardBatch)` on every build, which rewrites `_compute_mrope_positions`, a **private method on the upstream class object** `sglang.srt.model_executor.forward_batch_info.ForwardBatch`.

sglang_omni/model_runner/base.py
```diff
@@ -287,12 +354,117 @@ def _build_forward_batch(self, scheduler_output: Any):
...
+                "mrope": float(
+                    getattr(forward_batch, "_sglang_omni_mrope_seconds", 0.0)
+                ),
+                "mrope_fast": float(
+                    bool(getattr(forward_batch, "_sglang_omni_mrope_fast", False))
+                ),
+            }
         return forward_batch, schedule_batch, model_worker_batch, is_prefill
 
+    def _patch_forward_batch_mrope(self, forward_batch_cls: Any) -> None:
```
yxs, line 383: `_patch_forward_batch_mrope` uses `getattr(cls, "_sglang_omni_mrope_patch", False)` for idempotency, then non-atomically reads `original = cls._compute_mrope_positions`, installs the new method, and sets `cls._sglang_omni_mrope_patch = True`. There is no lock across these three steps. sglang's model runner is multi-process under TP>1 and build is usually single-threaded per process, so it looks safe; but this PR also introduces the async overlap path (`execute_launch`/`execute_resolve`), and the patch happens inside `_build_forward_batch`. If build ever runs under multiple threads (or multiple runners in one process each do their first build), two threads can both observe `_sglang_omni_mrope_patch=False` and each capture "the current `_compute_mrope_positions`" as `original`. The second thread captures the **patched version the first thread already installed** as `original` into `_sglang_omni_orig_compute_mrope_positions`, producing a patched-on-patched double wrapper that runs two nested try/finally + env reads per step and can corrupt the fallback chain.

luojiaxuan: Fixed. Patch installation is guarded by a module-level lock and keeps the original method stable on re-entry.

Issue #445, opened 2026-05-15
[Feature] LLaDA2.0-Uni model support


btw616:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang-omni/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

[LLaDA2.0-Uni](https://github.com/inclusionAI/LLaDA2.0-Uni) is a unified diffusion Large Language Model (dLLM) based on Mixture-of-Experts (MoE) that seamlessly integrates multimodal understanding and generation within a single model. We'd like to have sglang-omni support it.

### Model Architecture

<img width="2326" height="1502" alt="Image" src="https://github.com/user-attachments/assets/80f70318-5b1c-4137-b0d2-ba1a5fcec2ab" />

- **Unified dLLM-MoE Backbone**: Built on LLaDA 2.0, it unifies multimodal understanding and generation into a simple Mask Token Prediction paradigm.
- **Discrete Semantic Tokenizer**: Utilizes SigLIP-VQ to convert visual inputs into discrete semantic tokens, significantly enhancing multimodal understanding.
- **Efficient Diffusion Decoder**: Pairs discrete tokens with a specialized diffusion decoder for high-fidelity generation, enabling rapid 8-step inference via distillation.

### Roadmap

- [x] Initial LLaDA2.0-Uni support (text/image input + text output)
  - https://github.com/sgl-project/sglang-omni/pull/446
- [ ] Image output support for LLaDA2.0-Uni
- [ ] Streaming support
  - https://github.com/sgl-project/sglang-omni/pull/1455
- [ ] TP/EP support
- [ ] FP8 support
- [ ] CI & unit tests
- [x] Cookbook
  - https://github.com/sgl-project/sglang-omni/pull/598

### CC list

@zhaochenyang20

vibecodingengine:
I added support for image generation in #820 and marked the PR ready for review. Would appreciate some review when you have a chance.

kunchengit:
We opened seven Draft PRs covering LLaDA2-Uni thinker execution, reusable pipeline runtime, and image output in bounded pieces.

### Thinker execution

- #1486 — thinker tensor-parallel correctness and model-level decode CUDA graph support. This is an umbrella Draft for scope feedback and will be split into independently reviewable Ready PRs; expert parallelism remains out of scope.

### Reusable pipeline runtime

[...]

PR #820, opened 2026-06-18
Add LLaDA2-Uni image output token path

vibecodingengine:
## Description

This PR adds the LLaDA2-Uni image-output request path while keeping text requests on the existing SGLang dLLM scheduler.

- Routes `modalities: ["image"]`, `output_modalities: ["image"]`, and `image_config` requests to image-capable workers.
- Uses the checkpoint's upstream-style image-token generation path for image requests, including matching prompt tokenization, seed propagation, optional `flash_attn` fallback behavior, and decoder global RNG seeding.
- Lazily loads the image-token generator and image decoder, validates VQ token bounds, and returns OpenAI-compatible base64 PNG/JPEG image payloads.
- Adds docs and unit coverage for routing, API payload assembly, lazy decoding, image-token generation, upstream-style prompt tokenization, flash-attn fallback import, and decoder seeding.

Review notes:
- The decoder transport files under `sglang_omni/models/llada2_uni/components/decoder_transport/` are formatter-only copies of `inclusionAI/LLaDA2.0-Uni` `origin/main` at `3457030a9c737f77f38ad5ff657e7659243d3444` after this repo's pre-commit formatting path (`autoflake`, `isort`, `black`), so functional review can focus elsewhere.
- `decoder_model.py` is mostly upstream code. Local deltas are the optional `flash_attn` import guard, PyTorch SDPA fallback path, and the inference-only cleanup that removes inherited gradient-checkpointing / `torch.is_grad_enabled()` branches.
- The SGLang integration files (`image_decoder.py`, `image_token_generator.py`, `preprocessor.py`, `stages.py`), API/router changes, tests, and docs are not vendored copies and should still be reviewed.
- Known limitation: the staged integration does not preserve upstream phased memory lifetime. The HF image generator remains resident while `image_decode` loads and caches the decoder stack. With the local FP8 snapshot, retained image-only weights are roughly `16.2 GiB + 5.7 GiB + 0.5 GiB = ~22.4 GiB`, about +6.3 GiB versus upstream phased loading before activations, KV/cache, and allocator fragmentation. Closer upstream memory parity can be tracked as a follow-up.

## Test plan

- `.venv/bin/pytest tests/unit_test/llada2_uni/test_image_request_path.py tests/unit_test/serve/test_openai_api.py::test_chat_image_output_config_is_forwarded_to_generate_request tests/unit_test/serve/test_openai_api.py::test_chat_image_output_returns_images_array tests/unit_test/serve/test_generate_rollout.py::test_generate_image_response_includes_images_without_logprobs tests/unit_test/router/test_app.py::test_chat_image_output_routes_to_image_output_worker -q`
- `.venv/bin/python -m compileall -q sglang_omni/models/llada2_uni tests/unit_test/llada2_uni/test_image_request_path.py`
- `git diff --check`
- Ran CPU and GPU parity checks against upstream LLaDA2.0-Uni and verified SGLang-Omni image-output parity. Prompt: `a centered red apple on a plain wooden table, simple product photo, clear object, soft daylight`. CPU smoke parity used `segmind/tiny-sd`, 256x256, steps=12, guidance_scale=7.5, seed=1234, threads=4, deterministic=true, PNG output through `image_config`. GPU parity used `inclusionAI/LLaDA2.0-Uni-FP8`, 64x64, steps=2, block_length=16, cfg_scale=2.0, gen_length=4, decoder_steps=2, resolution_multiplier=2, seed=1234.

Not run: `ruff check` locally because Ruff is not installed in the virtualenv. The temporary CPU/GPU comparison scripts were not committed.

image


kunchengit:
Hi @vibecodingengine, thanks for working on LLaDA2-Uni image generation support! We appreciate the effort — especially the thorough test coverage, the upstream parity checks, and the clear documentation of known limitations.

We've been working on the same feature (tracked in #445) and wanted to share some thoughts on the architecture.

### Architectural concern

The core design choice here is loading the full LLaDA2-Uni model a second time via `AutoModelForCausalLM` and routing image requests to `model.generate_image()`. While this produces correct results, we think there's a more natural path for a serving framework:

LLaDA2-Uni image token generation is the same dLLM denoising loop as text generation — with the addition of CFG (batch=2 forward) and text-vocab masking. These can be implemented as an extension of the existing `LowConfidence` algorithm rather than bypassing the scheduler entirely.

The main tradeoffs of the HF bypass approach:

| | HF `generate_image` | Native dLLM CFG |
|---|---|---|
| Memory | +6.3 GiB (duplicate weights) | Zero extra |
| Batching | Single-request, synchronous | dLLM scheduler batches |
| Concurrency | Blocks GPU during image gen | Text & image share scheduler |
| Future improvements | Disconnected from SGLang | Inherits TP, CUDA graph, etc |

The `HybridScheduler` pattern also diverges from how other models handle multi-modal generation in sglang-omni — compare with the Ming-Omni approach (#876, #878) which reuses the thinker infrastructure.

### Our approach

We have a working implementation that keeps everything inside the dLLM scheduler:

- **`LowConfidenceCFG` algorithm** — batch=2 CFG with variance rescale, transfer schedule, text-vocab forcing
- **CFG batch planner** in `DllmScheduler` — creates/manages companion uncond Req
- **Image decoder** as a terminal stage — SigVQ + DiT ODE (similar to your `image_decoder.py`)
- **Thinking-mode T2I** — two-phase pipeline with thinker re-entry

We'll be posting our implementation shortly on #445. Some areas where your PR has great work we'd like to learn from:

- The **test structure** under `tests/unit_test/llada2_uni/` — clean pattern
- The **decoder transport** code organization
- The **API validation** in `openai_api.py`

Happy to discuss the architecture further and find ways to collaborate.

PR #821, opened 2026-06-18
Add RL refit admin API alias

Hayden727:
## Summary

Adds a Miles/UniRL-friendly `POST /refit` admin alias for the existing from-disk weight update path.

- Worker serve API accepts `/refit` with the same request contract as `/update_weights_from_disk`
- External router broadcasts `/refit` through the admin update lock and worker isolation path
- Adds focused SHA256 refit coverage that loads fake instruct weights, refits to fake base weights, and verifies the strict checksum matches
- Updates RL admin docs and the test catalog with refit validation guidance

## Validation

- Remote container: `novita-h100-omni`, container `sglang-omni-hayden-rl`
- `python -m compileall -q sglang_omni/serve/openai_api.py sglang_omni_router/app.py tests/unit_test/serve/test_openai_api.py tests/unit_test/router/test_app.py tests/unit_test/model_runner/test_weight_checker.py`
- `CUDA_VISIBLE_DEVICES=0 pytest tests/unit_test/model_runner/test_weight_checker.py tests/unit_test/serve/test_openai_api.py tests/unit_test/router/test_app.py -q` -> `109 passed in 3.63s`
- `pytest tests/test_model/test_rl_distributed_weight_update.py -s -q -vv --tb=short` -> `1 skipped in 0.09s` because the container does not have the Higgs base/instruct checkpoints cached (`base None`, `instruct None`)
- Local pre-commit hooks run by `git commit` passed

## Notes

The new route is intentionally conservative: `/refit` is an alias for disk refit metadata over the admin control plane. Bulk tensors/checkpoints remain on disk or the distributed data plane.

### DIFF
--- a/docs/developer_reference/rl_admin_control.md
+++ b/docs/developer_reference/rl_admin_control.md
@@ -31,6 +31,7 @@ The worker server supports:
 - `GET|POST /model_info`
 - `POST /pause_generation`
 - `POST /continue_generation`
+- `POST /refit`
 - `POST /update_weights_from_disk`
 - `POST /update_weights_from_tensor`
 - `POST /init_weights_update_group`
@@ -38,6 +39,8 @@ The worker server supports:
 - `POST /update_weights_from_distributed`
 - `GET|POST /weights_checker`
 
+`/refit` is a Miles/UniRL-friendly alias for the from-disk refit path and has
+the same request and response contract as `/update_weights_from_disk`.
 `/update_weights_from_disk` is the primary implemented update path. It pauses
 the target scheduler, optionally aborts active requests, calls the underlying
 SGLang model runner update method, optionally flushes cache, and resumes unless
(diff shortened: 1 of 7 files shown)


docs/developer_reference/rl_admin_control.md
```diff
@@ -31,13 +31,16 @@ The worker server supports:
 - `GET|POST /model_info`
 - `POST /pause_generation`
 - `POST /continue_generation`
+- `POST /refit`
```
yxs, line 34: It looks good overall, but I'm not sure if we should do it this way. The core issue is that we don't have a clear enough understanding of the entire roadmap yet.

I'd like to align on what `/refit` is *for*, because as built it's a disk-only alias of `/update_weights_from_disk` and three things don't line up:

1. `/refit` is hard-wired to the disk path — the slowest of the three. The distributed (NCCL) fast path that #784 already implemented, which is what an RL hot loop actually wants, isn't reachable through `/refit`, and tensor/IPC is still 501. So the "refit" entry covers exactly the transport we'd least want for per-step weight sync.
2.  "refit" means "from-disk refit path" or "distributed refit path", the consumer may confuse.
3. For now, Miles and UniRL — neither posts `/refit`; both call the literal `/update_weights_from_*` names. neither framework this targets calls it yet.

maybe we can hold it now, until we are clear. make `/refit` the real RL entry? or drop it?

cc @JingwenGu0829

PR #822, opened 2026-06-18
[MOSS]: Add SGLang-backed MOSS vocoder decoder

Ratish1:
## Motivation

This PR adds a faster non-streaming MOSS-TTS Local vocoder decode path by routing
the projected transformer stages through packed SGLang varlen FlashAttention.
The vocoder transformer is a large part of end-to-end generation latency; this
keeps the upstream codec embeddings, pretransform stages, and waveform projection
while replacing the non-streaming projected transformer attention execution path.

The optimized path is not bit-identical to the upstream processor output. Current
evidence shows the numerical drift is not primarily from SGLang FlashAttention
itself: the owned decoder with SDPA and the packed SGLang path drift similarly
versus upstream. The validation therefore focuses on mathematical alignment,
absence of invalid values, corpus WER/audio quality, and serving throughput.

## Design

The non-streaming vocoder path uses a scoped decoder replacement during the
tokenizer full-sequence decode path (`chunk_duration=None`):

```text
non-streaming MOSS codes
  -> tokenizer full-sequence decode (`chunk_duration=None`)
  -> scoped MOSS vocoder decoder replacement
  -> upstream codec embeddings / pretransform / waveform projection
  -> projected transformer stages
       input projection
       pack valid frames
       MOSS positional embedding / RoPE
       SGLang flash_attn_varlen_func
       output projection
       unpack valid frames
  -> waveform
```

Streaming sessions continue to use the existing stateful decoder path. The
optimized decoder is only installed around non-streaming decode and restored with
`try/finally`.

## Modifications

- Add `MossTTSLocalVocoderDecoder`, a MOSS-specific non-streaming decoder wrapper
  for projected transformer stages.
- Pack valid frame rows before attention and call SGLang `flash_attn_varlen_func`
  with MOSS local-causal window semantics.
- Preserve upstream non-attention vocoder mechanics: embeddings, patched
  pretransform stages, layer weights, output projection, and waveform projection
  remain owned by the loaded MOSS tokenizer module.
- Route non-streaming vocoder decode through the tokenizer full-sequence decode
  path (`chunk_duration=None`) with the optimized decoder installed, while
  keeping live streaming sessions on the original stateful path.
- Add focused unit coverage for decoder construction, scoped replacement,
  packed/unpacked paths, and streaming-path preservation.

## Accuracy Tests

Direct vocoder probe/oracle:

- Standard probes: `1x25`, `1x100`, `1x300`, `8x100`, `8x300`.
- Shape parity passed for all probes.
- No NaNs were observed.
- Packed SGLang attention oracle captured all calls.
- Worst packed attention oracle delta stayed in BF16-scale range:
  `0.015625-0.03125`.
- Waveform output is not bit-identical versus upstream processor decode.

Quality validation on SeedTTS EN 1088 generate-only set:

| case | commit | completed | failed | WER corpus | WER <50 | >50% outliers | UTMOS | speaker sim |
| upstream/main processor | `8a41fc4d` | 1088 | 0 | 2.18% | 1.58% | 7 | 3.9653 | 65.2950 |
| this PR | `a50c68c` | 1088 | 0 | 2.34-2.55% | 1.65-1.83% | 7-9 | 3.9642 | 65.3057 |

Same generated audio retranscribed three times:

| ASR pass | WER corpus | WER <50 | >50% outliers |
|---|---:|---:|---:|
| original | 2.55% | 1.83% | 9 |
| repeat 1 | 2.34% | 1.68% | 7 |
| repeat 2 | 2.35% | 1.65% | 8 |

Fixed-head duration buckets for this PR:

| bucket | count | WER corpus | WER <50 | >50% outliers |
|---|---:|---:|---:|---:|
| `<8s` | 1086 | 2.56% | 1.84% | 9 |
| `>=8s` | 2 | 0.00% | 0.00% | 0 |

Additional quality checks on this PR output:

- Audio integrity pass found no missing, unreadable, silent, empty, or
  duration-invalid files across `1088/1088` generated files.
- Output audio was consistently 48 kHz stereo with duration range
  `2.560-9.360s` and mean duration `4.401s`.
- Conservative full-scale clipping detector found `34` PR files with
  `max_abs >= 0.999`. This is reported as an integrity statistic, not a
  zero-clipping claim.

Boundary and ragged stress validation:

| check | result |
|---|---|
| boundary probes | shape preserved, zero oracle errors, zero oracle NaN calls |
| packed oracle worst delta | `0.00195312-0.03125` across boundary probes |
| long probes beyond local windows | waveform drift increases, but oracle remains BF16-scale |
| ragged boundary batch `16x1000` | packed max_abs `1.97266`, SDPA control max_abs `1.96875` |
| ragged long batch `10x1000` | packed max_abs `1.44141`, SDPA control max_abs `1.44141` |

The ragged controls show the large synthetic long/context-boundary waveform drift
is not unique to packed SGLang attention. It matches the owned SDPA execution
model, which is consistent with the non-bit-identical execution-model change.

Debug isolation performed after the WER comparison:

| candidate path | result |
|---|---|
| packed SGLang attention | waveform drift versus upstream processor |
| owned decoder with SDPA attention | similar waveform drift versus upstream processor |
| owned decoder delegating transformer layer body back to upstream source layer | similar waveform drift versus upstream processor |

This points to the non-streaming execution-model change as the remaining source
of waveform non-bit-exactness, not to a localized SGLang FlashAttention
packing/window bug.

## Speed Tests and Profiling

H100 1088 generate-only comparison:

| metric | upstream/main | this PR | delta |
|---|---:|---:|---:|
| completed / failed | 1088 / 0 | 1088 / 0 | - |
| QPS | 4.965 | 7.380 | +48.64% |
| RTF mean | 0.3847 | 0.2546 | -33.82% |
| latency mean | 1.608s | 1.082s | -32.71% |
| latency p95 | 2.127s | 1.486s | -30.14% |
| latency p99 | 2.632s | 1.710s | -35.03% |
| output throughput | 273.2 tok/s | 406.0 tok/s | +48.61% |
| output tokens total | not recorded | 59859 | - |


sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -786,12 +793,19 @@ def _decode_codes_rows(self, codes_list: list[torch.Tensor]) -> list[torch.Tenso
...
+            # Keep decode_audio_codes as the public codec path, but swap only the
+            # decoder module so non-streaming transformer attention uses packed
+            # SGLang FlashAttention. A live streaming session owns codec state, so
+            # it stays on the stateful session path below.
+            original_decoder = self._codec.decoder
+            self._codec.decoder = self._nonstream_decoder
+            try:
+                return [
+                    torch.as_tensor(wav).detach().to("cpu")
+                    for wav in self._processor.decode_audio_codes(codes_list)
```
Hayden727: This still routes the replacement decoder through `MossTTSLocalProcessor.decode_audio_codes()`, which hardcodes `audio_tokenizer.decode(..., chunk_duration=8)`. That tokenizer path opens `codec.streaming()` and calls `_decode_frame()` per 8s chunk, expecting the decoder stages to carry streaming KV/cache and position offsets. `MossTTSLocalVocoderDecoder` is a plain non-streaming `nn.Module`, so each internal chunk will be decoded independently with positions starting at 0 and no previous context. For non-streaming inputs longer than that chunk, this changes vocoder output at chunk boundaries. Please either bypass the processor chunked path and call the tokenizer full-sequence decode with `chunk_duration=None` when this decoder is installed, or implement streaming state/offset support in the wrapper; a >8s regression test through the processor/tokenizer path would catch this.

sglang_omni/models/moss_tts_local/vocoder_decoder.py
```diff
@@ -0,0 +1,495 @@
...
+
+from __future__ import annotations
+
+import importlib
+import math
+from collections.abc import Iterator
+from typing import Any
+
+import torch
+from sglang.jit_kernel.flash_attention import flash_attn_varlen_func
```
edwingao28: Can we guard this import? `_can_run_packed_flash()` can fall back at runtime, but this import happens at module load, so a missing `sglang.jit_kernel.flash_attention` / symbol would prevent the vocoder from importing before SDPA fallback gets a chance.

Existing flash users seem to use `None` + `try/except ImportError`; doing the same here and checking for `None` in `_can_run_packed_flash()` would make the fallback robust.

tests/unit_test/moss_tts_local/test_vocoder_decoder.py
```diff
@@ -0,0 +1,396 @@
...
+    source = _FallbackProjectedStage()
+    source.transformer.layers[0].self_attn.attention_implementation = (
+        "flash_attention_2"
+    )
+    wrapper = MossTTSLocalProjectedTransformer(source)
+    attn = wrapper.transformer.layers[0].self_attn
+    attn._can_run_packed_flash = lambda _: True  # type: ignore[method-assign]
+    calls = []
+
+    def fake_flash_attn(
```
edwingao28, line 220: The current flash tests stub `flash_attn_varlen_func` to return `q`, so they mostly validate packing metadata rather than the real packed flash numerics. Could we add a small CUDA-only bf16 test that runs the real kernel and compares against the SDPA path with `torch.testing.assert_close`?

Since the PR already notes a small WER delta, having one tighter numerical integration test would make this easier to trust. Also worth covering the zero-frame branch.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -786,12 +793,19 @@ def _decode_codes_rows(self, codes_list: list[torch.Tensor]) -> list[torch.Tenso
...
-            # session is live.
-            return [
-                torch.as_tensor(wav).detach().to("cpu")
-                for wav in self._processor.decode_audio_codes(codes_list)
-            ]
+            # Keep decode_audio_codes as the public codec path, but swap only the
+            # decoder module so non-streaming transformer attention uses packed
+            # SGLang FlashAttention. A live streaming session owns codec state, so
+            # it stays on the stateful session path below.
+            original_decoder = self._codec.decoder
```
edwingao28, line 853: Given the PR reports a small WER delta, should we put this behind a server arg or env flag? Defaulting to the new path is fine, but a kill switch would let ops revert just this vocoder path without rolling back the whole change.

sglang_omni/models/moss_tts_local/vocoder_decoder.py
```diff
@@ -0,0 +1,495 @@
...
+            window_size=self._flash_window_size(),
+        )
+        return self.out_proj(out.reshape(x.shape[0], self.embed_dim))
+
+    def _flash_window_size(self) -> tuple[int, int]:
+        if self.context is None or not self.causal:
+            return (-1, -1)
+        # MOSS's SDPA local mask keeps `context` total tokens including the current
+        # query token. FlashAttention's left-window argument counts prior keys.
+        return (max(int(self.context) - 1, 0), 0)
```
edwingao28, line 333: Can we lock this down with a tiny test? I think `context - 1` is right because FA's left window counts prior keys, while the MOSS mask includes the current token, but this is exactly the kind of off-by-one a masked-SDPA vs windowed-flash unit test would catch.

edwingao28:
The gains look quite impressive!

sglang_omni/models/moss_tts_local/streaming_vocoder.py
Ratish1: done. The owned non-streaming path now bypasses `decode_audio_codes()` and calls tokenizer `decode(..., chunk_duration=None)` while the packed decoder is installed. Added >100-frame coverage for this path.

sglang_omni/models/moss_tts_local/vocoder_decoder.py
Ratish1: done. The SGLang flash import is guarded and `_can_run_packed_flash()` falls back when the symbol is unavailable.

tests/unit_test/moss_tts_local/test_vocoder_decoder.py
Ratish1: done. Added a CUDA-only bf16 packed-flash vs SDPA reference test, plus zero-valid-length coverage.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
Ratish1: I did not add this for now. This path is intended to be the default implementation, and an env/server switch would double the production states. The dependency fallback remains covered by the import guard.

sglang_omni/models/moss_tts_local/vocoder_decoder.py
Ratish1: done. Added tests for the window tuple and for equivalence with the MOSS local-causal mask.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -781,17 +792,69 @@ def _store_vocoder_result(
...
-            ]
+            # The processor helper forces chunk_duration=8 and enters the
+            # tokenizer streaming loop. This decoder is non-streaming, so it must
+            # run through the tokenizer's full-sequence decode path.
+            original_decoder = self._codec.decoder
+            self._codec.decoder = self._nonstream_decoder
+            try:
+                return self._decode_codes_rows_nonstream(codes_list)
+            finally:
+                self._codec.decoder = original_decoder
```
gaoyang07, line 858: The non-streaming decode path no longer routes through `processor.decode_audio_codes()` — nice cleanup.

But the codec instance itself is still pulled from `processor.audio_tokenizer`, and the decode still relies on monkey-patching `codec.decoder` at call time here, it is still coupling codec acquisition to the HF processor object --- inelegant for me.

Not a blocker for this PR since the scope is attention optimization, but worth a follow-up: load the codec independently and own the full decode path in `_decode_codes_rows_nonstream` rather than monkey-patching `codec.decoder`.

Ratish1: Agreed. I kept this PR scoped to the attention path and preserved the existing processor-owned codec lifecycle to avoid changing load/device/dtype ownership in the same diff. Added a TODO for the follow-up to load and own the non-streaming codec directly so this path no longer needs the scoped decoder swap.

JiaxinD:
The `chunk_duration=None` fix is the right call, and the new tests (window-vs-MOSS-mask equality, CUDA bf16 packed-vs-SDPA, zero-length, flash-unavailable fallback) close most of the gaps.

I verified the fix on the real v1.5 codec (on synthetic codes, as a mechanism check, not a natural-speech quality measurement; deterministic across 3 runs): the old chunked path (stateless decoder + `chunk_duration=8`) diverged from the stateful reference by ~7% of peak amplitude across chunks 2+, with ~12% spikes at each 8s seam, while `chunk_duration=None` matches the stateful reference to ~0.1% everywhere with no seam. So the boundary issue is resolved.

One non-blocker: the quality table (WER 2.18 -> 2.44, etc.) is from the pre-fix commit `d23cd9cb`. Worth re-running on the fixed head, ideally bucketed by duration (<8s / >=8s), since the >8s tail behaves differently now.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
JiaxinD: Confirmed your call was right and the fix lands it. On the real v1.5 codec (synthetic codes, as a mechanism check), the old chunked path (stateless decoder + `chunk_duration=8`) diverged from the stateful reference by ~7% of peak across chunks 2+, with ~12% spikes at each 8s seam (sustained, not just a transient), while `chunk_duration=None` matches the stateful reference to ~0.1% with no seam. Pure-rope so no absolute-phase issue; the divergence was exactly the lost cross-chunk left context you flagged.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -781,17 +792,71 @@ def _store_vocoder_result(
...
+        )
+        padding_mask = torch.zeros(
+            len(codes_channels_first), max_len, device=device, dtype=torch.bool
+        )
+        for index, codes in enumerate(codes_channels_first):
+            length = int(codes.shape[1])
+            audio_codes[:, index, :length] = codes
+            padding_mask[index, :length] = True
+
+        decoded = self._codec.decode(
```
JiaxinD, line 823: Non-blocker: full-sequence decode here scales VRAM with utterance length (measured ~0.8GB @ 60s, ~1.6GB @ 120s, vs the chunked path's flat ~0.3GB; crossover ~20s). Negligible on H100 and fine for normal TTS lengths, but a max-length guard or chunked fallback for pathologically long inputs would be safe.

Ratish1: Agreed. I’m leaving this as a follow-up for now since a chunked fallback would need to preserve the same cross-chunk context semantics we just fixed, and adding a hard max-length guard is a broader decision

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -781,17 +792,71 @@ def _store_vocoder_result(
...
+        if audio is None or audio_lengths is None:
+            raise RuntimeError(
+                "audio_tokenizer.decode did not return audio/audio_lengths."
+            )
+        wavs = []
+        for index in range(int(audio.shape[0])):
+            length = int(audio_lengths[index].item())
+            wavs.append(
+                audio[index, :, :length].detach().to("cpu", torch.float32).contiguous()
+            )
```
zhaochenyang20: when non-streaming requests are merged by `_vocode_batch()`, the preceding `codec.decode(..., chunk_duration=None)` has already completed one batched GPU decode, but this loop then calls `.item()` on each `audio_lengths[index]` and copies each `audio[index, :, :length]` to CPU separately. On CUDA, `.item()` synchronizes the scalar length to the host; the following per-waveform D2H copies split batch materialization into multiple synchronization points. This is the non-streaming vocoder serving hot path that this PR is trying to speed up, and `max_batch_size` can batch several requests by default; the practical impact is that fuller batches create more tail synchronizations, making the latency and QPS win from packed attention easier to lose during output materialization."

Move CPU materialization of lengths and waveforms outside the loop, matching the existing streaming `step()` path: perform batched D2H first, then slice by CPU lengths. If copying padded audio for ragged batches is a concern, at minimum copy `audio_lengths` once and then decide whether a single batched audio copy is acceptable; the current per-request `.item()` should not remain in the GPU hot path.

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

PR #823, opened 2026-06-18
feat(higgs/rl): add rollout logprob + delay-pattern action-mask kernels

JingwenGu0829:
Closes part of #780 (1.1.2, Higgs).

## What

`POST /generate` with `return_omni_rollout=true` now returns
`meta_info.omni_rollout` for Higgs: the delayed multi-codebook **actions**, the
per-`(step, codebook)` selected-action **logprobs** (`logp_old`), and the
delay-pattern **action mask** — so an RL trainer can consume Higgs rollouts
directly. Built on the merged seed support (#824): a seeded `/generate` is a
fully reproducible rollout.

Adds: `rollout_trace.build_omni_rollout_trace` (the `omni_rollout` schema),
`utils.delay_pattern_action_mask` (the `T×N` real-audio parallelogram),
`sampler.selected_token_logprobs` (full-vocab `log_softmax(logits/T)` at the
sampled code, captured on both the eager and CUDA-graph paths), plus runner /
state / serialization wiring gated on `return_omni_rollout` / `return_logprob`.

## Validation

On-GPU, `boson-sglang/higgs-audio-v3-tts-4b` (H100), branch @ `0451f34`
(merged with `main`).

**Seeded reproducibility** (temp=0.8): same seed → identical `actions`
(`284fd49b…`) and `logprobs` (`013c9197…`); different seed differs; unseeded
random.

**Logprob correctness** — over many seeds, empirical frequency of a sampled code
must equal `exp(returned logp)` (pure temperature; catches a probs-vs-logprobs
error that reproducibility alone would miss):

| path | setup | result |
|---|---|---|
| eager / prefill (step 0) | temp 0.7, N=2000 | max z **1.92**, Σexp(logp) **0.9952** |
| eager / prefill (step 0) | temp 1.0, N=2000 | max z **1.72**, Σexp(logp) **0.9693** |
| decode / CUDA-graph (step 1, conditioned) | temp 0.7, N=4000 | max z **2.46**, Σexp(logp) **0.9748** |

All within binomial noise (top codes z<1.5), Σexp(logp)→1 ⇒ `logp_old` is the
**true** sampling log-probability on both paths, and temperature scaling is correct.

**Structural** (seeded run, `[81,8]`, T=74): `action_mask` == delay parallelogram
== real-audio (non-BOC/EOC) positions; `Σmask == T·N` (592); action logprobs
finite & ≤0; greedy (temp=0) byte-identical to `main`.

**Unit tests**: 102 Higgs + 36 model_runner/sampling green; lint clean.


sglang_omni/models/higgs_tts/rollout_trace.py
```diff
@@ -0,0 +1,91 @@
...
+        "name": stream_name,
+        "stage": stage,
+        "modality": "audio",
+        "action_type": "discrete",
+        "layout": "codebook_2d",
+        "flatten_order": "time_major",
+        "shape": [int(L), int(N)],
+        "vocab_size": int(codebook_vocab_size),
+        "actions": delayed_codes.to(torch.long).tolist(),
+        "logprobs": (
```
yxs, line 74: how about consider documenting the logp contract here: full-vocab log_softmax(logits/T), fp32, greedy uses raw logits.
Higgs's logprob is the sole source of truth for RL (miles's prefill recompute is text-only), good for future integration needs it written down.

sglang_omni/serve/openai_api.py
```diff
@@ -1063,12 +1063,17 @@ def _build_generate_response(
...
+        req.return_logprob
+        and result.output_token_logprobs is None
+        and result.omni_rollout is None
+    ):
         raise HTTPException(
             status_code=500,
             detail=(
-                "backend did not return output_token_logprobs "
-                "for return_logprob=true"
+                "backend did not return logprobs (output_token_logprobs or "
```
yxs: return_logprob defaults to True, but Higgs audio has no text logprobs, so a default request without return_omni_rollout=true 500s here. consider the error hint that return_omni_rollout=true is required; also 500 here vs 501 below, same case, consider unifying.

sglang_omni/models/higgs_tts/model_runner.py
```diff
@@ -233,8 +257,10 @@ def _collect_step_outputs_cg(
             )
         staging = self._decode_pack_gpu(n_real)
         combined_cpu = staging[:n_real].cpu()  # one blocking D2H (sync path)
+        logprobs_cpu = self.model._cg_logprobs_BN[:n_real].cpu()
```
yxs: Logprob capture is fully unconditional, pure waste on the default path.

`selected_token_logprobs` is computed unconditionally in `batched_step_direct`, and the captured logprobs are then unconditionally moved to host and appended per-step to `data.output_logprobs`.
Per decode step this adds:
1. on the **sync path** (`_collect_step_outputs_cg`) an extra blocking `self.model._cg_logprobs_BN[:n_real].cpu()`, a genuine per-step host sync, exactly the "host-device sync in a hot path" anti-pattern;
2. on the **async path** an extra pinned ping-pong buffer + a non-blocking copy; 
3. on **both paths** an extra `logprobs_cpu[b].to(torch.float32).clone()` + list append inside the collect loop.

yxs (approved):

PR #824, opened 2026-06-18
feat(sampling): Framework-level refractor to unify the seed in sampling API->model runner

JingwenGu0829:
## Summary

Deterministic, seed-controlled sampling is a prerequisite for reproducible
rollouts (e.g. RL: replaying a trajectory, group sampling with one seed per
sample, train/rollout logprob equivalence). Today seed handling in SGLang-Omni
is fragmented — a few models hand-roll it, the standard text path silently drops
the seed, and the discrete-TTS custom samplers ignore it entirely.

This PR makes a request `seed` **honored uniformly across every AR sampling
model**:

1. A reusable seed hook in the **base `ModelRunner`** so the standard SGLang
   sampler path honors `seed` without each model re-implementing it.
2. Per-model wiring of the **custom discrete-TTS samplers** (Higgs, Fish S2-Pro)
   that previously sampled with bare `torch.multinomial`.
3. Unifying the **Qwen3-Omni talker** onto the same seed scheme so both stages of
   the omni pipeline seed consistently.

It is a strict no-op when no request in the batch supplies a seed (normal
serving unchanged). If a standard-SGLang batch mixes seeded and unseeded rows,
the unseeded rows receive a request-id-derived fallback seed so TP ranks stay in
sync instead of minting rank-local random fallback seeds. This adds no
measurable throughput cost.

## Scope — which models honor `seed` after this PR

| Path | Models | Status |
|---|---|---|
| **Standard SGLang sampler** (via new `base._install_sampling_seeds`) | Qwen3-Omni thinker, Ming-Omni thinker, LLaDA2-Uni, Voxtral text-LM | **Fixed** — seed reached `SamplingParams` but was dropped by SGLang's `enable_deterministic_inference` gate; the base hook installs it onto `sampling_info` directly |
| **Custom discrete sampler** (wired here) | **Higgs TTS**, **Fish S2-Pro** | **Fixed** — now use `multinomial_with_seed` |
| **Custom sampler unified onto the shared scheme** | **Qwen3-Omni talker** | **Fixed** — was seed-aware but used `int(seed) else 0` (unseeded → fixed 0); now uses the shared `resolve_row_seed` (masked seed, or random when unseeded) |
| Already honored (hand-rolled; untouched) | MOSS-TTS, MOSS-TTS-Local, Qwen3-TTS | unchanged (base hook skips rows a subclass already seeded) |
| Out of scope (continuous / flow-matching) | Voxtral acoustic, Ming-Omni talker | N/A — no discrete-token seed |

→ Net: a full **Qwen3-Omni** omni rollout (thinker text + talker audio) is now
end-to-end reproducible with a seed — it was not before (the thinker dropped it).

## Design

- **`sglang_omni/sampling/seed.py`** — shared helpers: derive stable
  positive-int32 seeds (namespaced), plus fresh random seeds for model-owned
  unseeded rows.
- **`ModelRunner._install_sampling_seeds`** — called once in
  `_sample_next_token_ids` before sampling; builds per-row seeds from each
  request's `sampling_params.sampling_seed` and installs them on
  `forward_batch.sampling_info.sampling_seed`, so SGLang routes to
  `multinomial_with_seed`. No-op when no row carries a seed; in a mixed
  seeded/unseeded batch, unseeded rows get a request-id-derived fallback seed so
  TP ranks agree on the fallback. Skips when a subclass already installed its
  own. Strict field access (no defensive `getattr`) — a missing contract field
  fails loud.
- **Higgs / Fish** — each seeds its draw with
  `multinomial_with_seed(probs, per_request_seed, position)` where `position`
  encodes the AR step (and codebook, for Higgs’ `step·N + codebook`), making
  draws **batch-composition independent**. A `-1` sentinel keeps the exact legacy
  `torch.multinomial` path for unseeded rows, so unseeded decode is byte-identical
  to before; greedy/argmax is untouched.
- **Qwen3-Omni talker** — keeps its own static `SamplingBatchInfo` (needed for
  CUDA-graph capture); only the seed *values* change, now from `resolve_row_seed`.

## Validation (on-GPU, end-to-end)

Baseline `main` @ `df62e91` vs branch `seed-framework-support` (HEAD `3ba583a`).
Each server launched from its own checkout; audio compared by sha256 of response
bytes, text by content hash.

### A. No-op / no regression when unseeded

**Higgs TTS** — greedy (`temperature=0`), `/v1/audio/speech`:

| Prompt | main | branch | match |
|---|---|---|---|
| Hello, how are you today? | ac8cbe76f87f093f | ac8cbe76f87f093f | ✅ |
| The quick brown fox… | 97e4aa746c93c785 | 97e4aa746c93c785 | ✅ |
| Reinforcement learning… | ec6868db8279e653 | ec6868db8279e653 | ✅ |
| She sells sea shells… | 8b60bb0b3c1397f3 | 8b60bb0b3c1397f3 | ✅ |
| Numbers like one two three… | feb72edc899a5a87 | feb72edc899a5a87 | ✅ |

→ byte-identical main vs branch (re-confirmed on full branch: `f92061d824eaad1f`
on both). **Qwen3-Omni thinker** greedy, no seed: main == branch == `ff8fa6bef2c3`.

**Throughput** (no regression):

| Model | main (req/s) | branch (req/s) | Δ |
|---|---:|---:|---|
| Higgs (N=32, conc=4, temp=0) | 4.07 | 4.25 | +4.4% (noise) |
| Qwen3-Omni thinker (warmup + N=48×2, conc=4) | 14.2 | 14.29 | ~0% |

### B. `seed` is honored (the fix), `temperature>0`

Same request twice with the same seed must be reproducible on the branch; a
different seed must give a different (but itself reproducible) result; unseeded
stays random.

**Qwen3-Omni-30B full pipeline (thinker text + talker audio)** — FP8,
disaggregated (thinker GPU0 / talker+code2wav GPU1), `/v1/chat/completions`
`modalities:["text","audio"]`, temp=0.8:

| | seed=1234 ×2 | seed=5678 | no seed ×2 |
|---|---|---|---|
| main (`df62e91`) | `a40e1ffb…` vs `ad6fcc02…` → non-reproducible (bug) | — | — |
| branch (`3ba583a`) | `3b9c09e04b30e859` ×2 → ✅ **reproducible** | differs ✅ | random ✅ |

**Qwen3-Omni thinker** (standard path, text), temp=0.8, seed=1234 ×2: main
`4b70d0ab4232`/`f81a5c6aaed5` (non-reproducible) vs branch `1fac69e63028` ×2 (✅).

**Higgs TTS** (custom sampler), temp=0.8: branch seed=1234 reproducible
(`3dbc11c98d8bb618`); main non-reproducible; seed=5678 differs; unseeded random.

**Fish S2-Pro** (custom sampler), temp=0.8: seed=1234 and seed=5678 each
reproducible (`3dbc11c98d8bb618` / `b863c8fdec6cc010`) and distinct; unseeded
random.

### DIFF
--- a/sglang_omni/model_runner/base.py
+++ b/sglang_omni/model_runner/base.py
@@ -12,7 +12,7 @@ from typing import Any
 
 import torch
 
-from sglang_omni.sampling.seed import resolve_row_seed
+from sglang_omni.sampling.seed import SAMPLING_SEED_MASK, resolve_row_seed
 from sglang_omni.scheduling.types import (
     ModelRunnerOutput,
     RequestOutput,
@@ -577,17 +577,27 @@ class ModelRunner:
         """Install per-row ``seed``s onto ``sampling_info`` so SGLang routes to
         ``multinomial_with_seed``. No-op when no request set a seed, or when a
         subclass already installed its own (e.g. Qwen3-TTS).
+
+        Runs once per decode step. The resolved seed is a per-request constant, so
+        it is resolved once and cached back onto ``sampling_params.sampling_seed``
+        (an unseeded row mints one random seed for its whole generation); later
+        steps reuse it instead of re-running ``os.urandom`` / re-masking per step.
         """
         sampling_info = forward_batch.sampling_info
         if sampling_info.sampling_seed is not None:
             return
-        public_seeds = [sr.data.req.sampling_params.sampling_seed for sr in requests]
-        if all(seed is None for seed in public_seeds):
+        sampling_params = [sr.data.req.sampling_params for sr in requests]
+        if all(sp.sampling_seed is None for sp in sampling_params):
             return
+        row_seeds: list[int] = []
+        for sp in sampling_params:
+            seed = sp.sampling_seed
+            if seed is None or not (0 <= seed <= SAMPLING_SEED_MASK):
+                seed = resolve_row_seed(seed)  # random for None, mask otherwise
+                sp.sampling_seed = seed  # cache: stable across this request's steps
+            row_seeds.append(seed)
         sampling_info.sampling_seed = torch.tensor(
-            [resolve_row_seed(seed) for seed in public_seeds],
-            dtype=torch.long,
-            device=sampling_info.device,
+            row_seeds, dtype=torch.long, device=sampling_info.device
         )
 
     @staticmethod
--- a/sglang_omni/model_runner/base.py
+++ b/sglang_omni/model_runner/base.py
@@ -12,6 +12,7 @@ from typing import Any
 
 import torch
 
+from sglang_omni.sampling.seed import resolve_row_seed
 from sglang_omni.scheduling.types import (
     ModelRunnerOutput,
     RequestOutput,
@@ -545,6 +546,7 @@ class ModelRunner:
     ) -> Any:
         self._apply_repetition_penalty(logits_output, requests)
         self._apply_codec_suppress_tokens(logits_output, requests)
+        self._install_sampling_seeds(forward_batch, requests)
         wants_rollout_logprob = any(sr.data.return_logprob for sr in requests)
         if wants_rollout_logprob:
             self._enable_sampler_logprobs(forward_batch, len(requests))
@@ -570,6 +572,23 @@ class ModelRunner:
                 requests,
             )
         return next_token_ids
+
+    def _install_sampling_seeds(self, forward_batch: Any, requests: list) -> None:
+        """Install per-row ``seed``s onto ``sampling_info`` so SGLang routes to
+        ``multinomial_with_seed``. No-op when no request set a seed, or when a
+        subclass already installed its own (e.g. Qwen3-TTS).
+        """
+        sampling_info = forward_batch.sampling_info
+        if sampling_info.sampling_seed is not None:
+            return
+        public_seeds = [sr.data.req.sampling_params.sampling_seed for sr in requests]
+        if all(seed is None for seed in public_seeds):
+            return
+        sampling_info.sampling_seed = torch.tensor(
+            [resolve_row_seed(seed) for seed in public_seeds],
+            dtype=torch.long,
+            device=sampling_info.device,
+        )
 
     @staticmethod
     def _enable_sampler_logprobs(forward_batch: Any, batch_size: int) -> None:
--- a/sglang_omni/model_runner/base.py
+++ b/sglang_omni/model_runner/base.py
@@ -632,12 +632,14 @@ class ModelRunner:
 
     @staticmethod
     def _validate_seeded_sampling_supported(sampling_info: Any) -> None:
-        if sampling_info.need_min_p_sampling:
+        if getattr(sampling_info, "need_min_p_sampling", False):
             raise ValueError(
                 "SGLang seeded sampling does not support min_p yet; set min_p=0 "
                 "or omit request seed"
             )
-        if not (sampling_info.need_top_p_sampling or sampling_info.need_top_k_sampling):
+        need_top_p_sampling = getattr(sampling_info, "need_top_p_sampling", False)
+        need_top_k_sampling = getattr(sampling_info, "need_top_k_sampling", False)
+        if not (need_top_p_sampling or need_top_k_sampling):
             return
         if _current_sglang_sampling_backend() == "flashinfer":
             raise ValueError(
--- a/sglang_omni/model_runner/base.py
+++ b/sglang_omni/model_runner/base.py
@@ -12,7 +12,11 @@ from typing import Any
 
 import torch
 
-from sglang_omni.sampling.seed import SAMPLING_SEED_MASK, resolve_row_seed
+from sglang_omni.sampling.seed import (
+    SAMPLING_SEED_MASK,
+    derive_sampling_seed,
+    resolve_row_seed,
+)
 from sglang_omni.scheduling.types import (
     ModelRunnerOutput,
     RequestOutput,
@@ -29,6 +33,18 @@ def _current_sglang_sampling_backend() -> str | None:
         return get_global_server_args().sampling_backend
     except ValueError:
         return None
+
+
+def _rank_shared_unseeded_sampling_seed(request: Any, row_idx: int) -> int:
+    request_id = getattr(request, "request_id", None)
+    if request_id is None:
+        request_id = getattr(getattr(request, "data", None), "request_id", None)
+    if request_id is None:
+        req = getattr(getattr(request, "data", None), "req", None)
+        request_id = getattr(req, "rid", None)
+    if request_id is None:
+        request_id = f"row-{row_idx}"
+    return derive_sampling_seed("sglang-omni-unseeded-row", request_id)
 
 
 @dataclass
@@ -587,10 +603,11 @@ class ModelRunner:
         ``multinomial_with_seed``. No-op when no request set a seed, or when a
         subclass already installed its own (e.g. Qwen3-TTS).
 
-        Runs once per decode step. The resolved seed is a per-request constant, so
-        it is resolved once and cached back onto ``sampling_params.sampling_seed``
-        (an unseeded row mints one random seed for its whole generation); later
-        steps reuse it instead of re-running ``os.urandom`` / re-masking per step.
+        Runs once per decode step. User-provided seeds are resolved once and
+        cached back onto ``sampling_params.sampling_seed``. In a mixed
+        seeded/unseeded batch the SGLang sampler is batch-wide, so unseeded rows
+        receive a request-id-derived fallback seed instead of a rank-local random
+        seed; this keeps TP ranks in sync without mutating the public request seed.
         """
         sampling_info = forward_batch.sampling_info
         if sampling_info.sampling_seed is not None:
@@ -601,11 +618,13 @@ class ModelRunner:
             return
         self._validate_seeded_sampling_supported(sampling_info)
         row_seeds: list[int] = []
-        for sp in sampling_params:
+        for row_idx, (sp, request) in enumerate(zip(sampling_params, requests)):
             seed = sp.sampling_seed
-            if seed is None or not (0 <= seed <= SAMPLING_SEED_MASK):
-                seed = resolve_row_seed(seed)  # random for None, mask otherwise
-                sp.sampling_seed = seed  # cache: stable across this request's steps
+            if seed is None:
+                seed = _rank_shared_unseeded_sampling_seed(request, row_idx)
+            elif not (0 <= seed <= SAMPLING_SEED_MASK):
+                seed = resolve_row_seed(seed)  # mask and cache user seed
+                sp.sampling_seed = seed
             row_seeds.append(seed)
         sampling_info.sampling_seed = torch.tensor(
             row_seeds, dtype=torch.long, device=sampling_info.device
--- a/sglang_omni/model_runner/base.py
+++ b/sglang_omni/model_runner/base.py
@@ -22,6 +22,15 @@ from sglang_omni.scheduling.types import (
 logger = logging.getLogger(__name__)
 
 
+def _current_sglang_sampling_backend() -> str | None:
+    try:
+        from sglang.srt.server_args import get_global_server_args
+
+        return get_global_server_args().sampling_backend
+    except ValueError:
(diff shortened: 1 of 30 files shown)


sglang_omni/sampling/seed.py
```diff
@@ -0,0 +1,48 @@
...
+def derive_sampling_seed(public_seed: int, namespace: str) -> int:
+    """Stable positive-int32 seed from a public ``seed``, namespaced per use site
+    so different stages of one request don't draw identically."""
+    digest = hashlib.blake2b(
+        f"{namespace}:{int(public_seed)}".encode("utf-8"), digest_size=8
+    ).digest()
+    return int.from_bytes(digest, "little") & SAMPLING_SEED_MASK
+
+
+def resolve_row_seed(public_seed: int | None, *, namespace: str | None = None) -> int:
```
JiaxinD: No caller passes `namespace` (base/talker/higgs/fish all call `resolve_row_seed(seed)`), so `derive_sampling_seed` is dead and thinker+talker of one request derive the same row seed. Either wire per-stage namespaces or drop the param plus the docstring's namespacing claim.

sglang_omni/model_runner/base.py
```diff
@@ -571,6 +573,23 @@ def _sample_next_token_ids(
...
+        subclass already installed its own (e.g. Qwen3-TTS).
+        """
+        sampling_info = forward_batch.sampling_info
+        if sampling_info.sampling_seed is not None:
+            return
+        public_seeds = [sr.data.req.sampling_params.sampling_seed for sr in requests]
+        if all(seed is None for seed in public_seeds):
+            return
+        sampling_info.sampling_seed = torch.tensor(
+            [resolve_row_seed(seed) for seed in public_seeds],
```
JiaxinD: In a mixed batch, unseeded rows also get a fresh seed here and route through `multinomial_with_seed` instead of legacy `torch.multinomial`, so the "unseeded byte-identical" claim only holds when no row in the batch carries a seed.

sglang_omni/model_runner/base.py
```diff
@@ -571,6 +573,23 @@ def _sample_next_token_ids(
...
 
+    def _install_sampling_seeds(self, forward_batch: Any, requests: list) -> None:
+        """Install per-row ``seed``s onto ``sampling_info`` so SGLang routes to
+        ``multinomial_with_seed``. No-op when no request set a seed, or when a
+        subclass already installed its own (e.g. Qwen3-TTS).
+        """
+        sampling_info = forward_batch.sampling_info
+        if sampling_info.sampling_seed is not None:
+            return
+        public_seeds = [sr.data.req.sampling_params.sampling_seed for sr in requests]
```
yxs: `_install_sampling_seeds` runs in `_sample_next_token_ids`, i.e. once per decode step. As soon as any row sets a seed, it rebuilds the seed for **every** row via `[resolve_row_seed(seed) for seed in public_seeds]`: seeded rows recompute the same constant each step (wasteful but cheap), while unseeded rows hit an `os.urandom(4)` syscall every step to mint a fresh seed, followed by a `torch.tensor(..., device=cuda)` H2D copy.

sglang_omni/sampling/seed.py
```diff
@@ -0,0 +1,48 @@
+# SPDX-License-Identifier: Apache-2.0
```
yxs, line 1: The three things in the new `seed.py` — `SAMPLING_SEED_MASK = 0x7FFFFFFF`, `new_random_sampling_seed` (`os.urandom(4) & mask`), and `derive_sampling_seed` (`blake2b(f"{ns}:{seed}")`) — already exist as two near-verbatim copies: `qwen3_tts`'s `_QWEN3_TTS_SAMPLING_SEED_MASK` / `_new_qwen3_tts_sampling_seed` / `_derive_qwen3_tts_child_seed`, and `moss_tts`'s `_MOSS_TTS_SAMPLING_SEED_MASK` / `_new_moss_tts_sampling_seed` / `derive_moss_tts_sampling_seed`. moss's docstring even says "Mirrors the Qwen3-TTS pattern".

3 times now, QAQ.

sglang_omni/models/qwen3_omni/components/talker.py
yxs, line 896: same here, 
talker regenerate unseeded-row seeds via `os.urandom` every decode step.
I can update this together.

sglang_omni/model_runner/base.py
```diff
@@ -571,6 +573,33 @@ def _sample_next_token_ids(
             )
         return next_token_ids
 
+    def _install_sampling_seeds(self, forward_batch: Any, requests: list) -> None:
```
yxs, line 601: In the base path, as soon as **one** row sets a seed, `sampling_info.sampling_seed` becomes a full tensor (unseeded rows filled with random seeds); being non-None, it makes the SGLang sampler route the **entire batch** (including unseeded rows) through `multinomial_with_seed` instead of `torch.multinomial`.

which against "byte-identical results on unseeded requests" in PR

sglang_omni/models/higgs_tts/sampler.py
```diff
@@ -262,7 +279,17 @@ def _sample_independent_batched(
...
         probs = _fused_top_p_renorm(probs, tp)
+
     codes_flat = probs.multinomial(num_samples=1).squeeze(-1)
+    if seeds_B is not None:
+        # Seeded rows draw deterministically from (seed, step*N + codebook);
+        # unseeded rows (seed == NO_SEED) keep the torch.multinomial draw above.
+        cb = torch.arange(N, device=logits_BNV.device).view(1, N).expand(B, N)
+        positions = (step_B.view(B, 1) * N + cb).reshape(B * N)
+        seeds_flat = seeds_B.clamp_min(0).view(B, 1).expand(B, N).reshape(B * N)
+        seeded_flat = multinomial_with_seed(probs, seeds_flat, positions).squeeze(-1)
```
edwingao28: I think this needs to pass log-probs, not probs. In the pinned SGLang sampler, `multinomial_with_seed` is doing Gumbel-max over what it receives (`argmax(logprobs + gumbel)`); the built-in probs path does `torch.log(probs)` before calling it.

With `probs` here, the seeded path samples from `argmax(p + gumbel)`, which is a different distribution from the old `torch.multinomial(probs)` path. Can we pass `probs.clamp_min(torch.finfo(probs.dtype).tiny).log()` here instead?

sglang_omni/models/fishaudio_s2_pro/sglang_model.py
```diff
@@ -428,9 +436,15 @@ def _decode_codebooks(self, logits: Tensor, hidden_states: Tensor) -> None:
...
             top_k_logits / temperature.clamp(min=1e-5), dim=-1
         )
-        semantic_token = top_k_indices.gather(
-            -1, torch.multinomial(probs, num_samples=1)
-        ).squeeze(-1)
+        # Seeded rows draw reproducibly from (seed, step); unseeded rows keep the
+        # legacy torch.multinomial draw, so unseeded decode is unchanged.
+        seeds = self._sampling_seeds[:bs]
+        unseeded_choice = torch.multinomial(probs, num_samples=1)
+        seeded_choice = multinomial_with_seed(
```
edwingao28, line 443: Same issue here: this is feeding post-softmax probs into a helper that expects log-probs on the current SGLang version. The unseeded branch samples from `probs`, but the seeded branch now samples from a different distribution.

Please convert to log-probs before `multinomial_with_seed`, e.g. `probs.clamp_min(torch.finfo(probs.dtype).tiny).log()`.

sglang_omni/model_runner/base.py
```diff
@@ -571,6 +573,33 @@ def _sample_next_token_ids(
...
+        if sampling_info.sampling_seed is not None:
+            return
+        sampling_params = [sr.data.req.sampling_params for sr in requests]
+        if all(sp.sampling_seed is None for sp in sampling_params):
+            return
+        row_seeds: list[int] = []
+        for sp in sampling_params:
+            seed = sp.sampling_seed
+            if seed is None or not (0 <= seed <= SAMPLING_SEED_MASK):
+                seed = resolve_row_seed(seed)  # random for None, mask otherwise
```
edwingao28: This still makes mixed seeded/unseeded batches a little scary under TP. Once any row has a seed, every row gets a `sampling_seed`; for unseeded rows we mint that fallback seed locally and then route them through `multinomial_with_seed`.

If sampling runs on each TP rank, those ranks can mint different fallback seeds for the same unseeded row. SGLang also does not sync final token ids across TP by default, so this can desync the ranks. I'd rather keep truly-unseeded rows on the legacy `torch.multinomial` path, or derive/broadcast the fallback seed from something rank-shared.

tests/unit_test/higgs_tts/test_seeded_sampling.py
```diff
@@ -0,0 +1,62 @@
...
+    return _sample_independent_batched(
+        logits,
+        temperature=torch.ones(B, device="cuda"),
+        top_p=None,
+        seeds_B=torch.tensor(seeds_B, device="cuda"),
+        step_B=torch.full((B,), step, dtype=torch.long, device="cuda"),
+    )
+
+
+def test_same_seed_same_step_is_reproducible():
```
edwingao28, line 36: These are good seed-plumbing tests, but they only prove determinism. A deterministic sampler with the wrong distribution still passes all of them.

Can we add one distribution sanity check over many seeds, comparing the seeded histogram against the old `torch.multinomial(probs)` / expected probs within tolerance? That would catch the probs-vs-logprobs bug above. Fish should also get at least one real seeded GPU test, since the current CPU test stubs `multinomial_with_seed`.

sglang_omni/models/fishaudio_s2_pro/sglang_model.py
```diff
@@ -428,9 +436,15 @@ def _decode_codebooks(self, logits: Tensor, hidden_states: Tensor) -> None:
...
+            probs, seeds.clamp_min(0), self._step_count[:bs]
+        )
+        choice = torch.where((seeds >= 0).unsqueeze(-1), seeded_choice, unseeded_choice)
+        semantic_token = top_k_indices.gather(-1, choice).squeeze(-1)
```
yijiangt, line 447: Was looking at this and I think `multinomial_with_seed` expects log-space input, but probs here is post-softmax [0,1]. The Gumbel noise inside is large enough that against probs in [0,1] it could dominate and push sampling toward uniform, so seed control may effectively be ignored. Results still stay consistent per seed though, which is probably why repro tests don't catch it.

For reference, both `moss_tts/model_runner.py:467` and `moss_tts_local/local_transformer.py:55` pass logits here. So this call, along with `fishaudio_s2_pro/sglang_model.py:444`, are the two new ones in this PR that pass probs instead.

sglang_omni/model_runner/base.py
```diff
@@ -571,6 +573,33 @@ def _sample_next_token_ids(
...
+                sp.sampling_seed = seed  # cache: stable across this request's steps
+            row_seeds.append(seed)
+        sampling_info.sampling_seed = torch.tensor(
```
Hayden727, line 629: Installing `sampling_seed` here still makes common seeded requests hit unsupported upstream sampler paths: the default flashinfer backend asserts for seeded top-p/top-k sampling, and the pytorch path asserts for seeded `min_p`. Please either route seeded standard-sampler stages to a supported backend/parameter set or reject unsupported seeded params before the request reaches the sampler.

sglang_omni/models/higgs_tts/sampler.py
Hayden727: This should pass logits/logprobs, not probabilities, into `multinomial_with_seed`. The upstream kernel is Gumbel-max over logprobs, so feeding `probs` changes the draw to `softmax(probs)` (on H100, `[0.9, 0.1]` sampled at ~0.69 instead of ~0.90); keep the unseeded `torch.multinomial(probs)` path, but feed masked scores/logprobs to the seeded path here and in Fish.

edwingao28: +1

sglang_omni/models/fishaudio_s2_pro/sglang_model.py
edwingao28: +1

Hayden727 (approved):

PR #826, opened 2026-06-19
[MOSS]: Use thread_local capture_error_mode for CUDA graph

### LINKED ISSUE #825: MOSS streaming vocoder lazy CUDA-graph recapture can race colocated AR .tolist() under global capture mode
## Summary

MOSS streaming vocoder captures CUDA graphs with PyTorch's default `capture_error_mode="global"`. In colocated serving, a lazy vocoder CUDA-graph recapture can overlap with AR output processing that calls `next_token_ids.tolist()`.

If `next_token_ids` is CUDA-backed, `.tolist()` first moves the tensor to CPU, which can involve a GPU-to-CPU transfer/synchronization. Under global CUDA stream-capture mode, CUDA may reject potentially unsafe CUDA operations from sibling threads while another thread is capturing. The result can surface as a spurious AR-side 500:

```text
operation not permitted when stream is capturing
```

## Evidence

Confirmed against latest `upstream/main` at `ab109259`.

MOSS vocoder CUDA graph capture still omits `capture_error_mode`:

`sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py:161`

```python
with torch.cuda.graph(graph, pool=self._pool):
```

Because no mode is passed, PyTorch uses the default `capture_error_mode="global"`.

The lazy recapture path still exists:

`sglang_omni/models/moss_tts_local/streaming_vocoder.py:557-564`

The relevant docstring says a session created after non-streaming traffic closed the graphed startup session is re-captured, and that the recapture runs synchronously there.

The non-streaming path can close the idle startup session before later streaming traffic:

`sglang_omni/models/moss_tts_local/streaming_vocoder.py:784-787`

```python
with self._state_lock:
    self._close_idle_startup_session_locked()
```

The AR output path still calls `.tolist()` on `next_token_ids`:

`sglang_omni/scheduling/sglang_backend/output_processor.py:35`

```python
model_output.next_token_ids.tolist()
```

## Expected Behavior

Vocoder CUDA-graph capture should not cause unrelated colocated AR requests to fail.

## Actual Behavior

Under colocated serving, a vocoder lazy recapture can globally restrict CUDA operations in sibling threads. A concurrent AR `.tolist()` can fail with `operation not permitted when stream is capturing`, even though the AR request is not itself capturing the vocoder graph.

## Why This Happens

`torch.cuda.graph(...)` defaults to `capture_error_mode="global"`. In global mode, potentially unsafe CUDA operations in other threads may be rejected while a capture is active.

In this case, the vocoder thread is capturing a CUDA graph, while the AR/talker output processor may simultaneously run `.tolist()` on a CUDA tensor. That `.tolist()` is unrelated to the vocoder graph, but global capture mode can still reject it.

## Proposed Fix

Use thread-local capture mode for the vocoder graph capture:

```python
with torch.cuda.graph(
    graph,
    pool=self._pool,
    capture_error_mode="thread_local",
):
    result = self._codec._decode_frame(static_codes, static_lengths)
```

This scopes capture-safety checks to the vocoder capture thread and avoids rejecting unrelated AR CUDA operations in sibling threads.

## Additional Hardening

- Add a regression test/assert that MOSS vocoder graph capture passes `capture_error_mode="thread_local"`.
- Consider ensuring streaming vocoder graphs are captured before serving traffic, or avoid closing the graphed startup session when possible.
- Keep graceful eager fallback on capture failure; on tight VRAM systems, consider clearing cache before continuing to reduce follow-on fragmentation/OOM symptoms.

AkazaAkane:
## Motivation

`MossVocoderCudaGraphRunner` captures CUDA graphs for vocoder execution, but graph capture can interact badly with other CUDA work in the same process if capture error handling is process-global. In colocated serving, this can make graph capture less isolated than intended: one thread/component entering CUDA graph capture can affect unrelated live serving work in the same CUDA process.

This PR makes the vocoder graph capture scope explicitly thread-local by passing `capture_error_mode="thread_local"` to `torch.cuda.graph(...)`. The goal is to keep CUDA graph capture error handling local to the runner thread and reduce cross-thread failure coupling during lazy or repeated graph capture.

## Modifications

* Update `MossVocoderCudaGraphRunner` to call `torch.cuda.graph(...)` with:

  ```python
  capture_error_mode="thread_local"
  ```

* Add a focused unit test that uses `inspect` and `ast` to parse the runner source and assert that the `torch.cuda.graph(...)` call includes the `capture_error_mode` keyword.

* Add the necessary test imports for source inspection and AST parsing.

## Related Issues

Fixes #825.

## Accuracy Test

pytest and pre-commit passed.

This PR does not change model weights, model architecture, kernels, vocoder numerics, or decoding logic. It only changes CUDA graph capture error-scope behavior and adds a source-level regression test.

## Benchmark & Profiling

Not applicable / not expected to affect steady-state performance.

The change only adds a keyword argument to CUDA graph capture setup. Runtime graph replay behavior should remain unchanged. The intended benefit is safer capture isolation in colocated serving, especially when graph capture can happen outside pure startup warmup.

## Checklist

* [x] Add unit tests.

## CI

long as the label remains. Use `/tag-and-rerun-ci moss` to select the MOSS TTS CI model. Draft PRs are skipped even if labeled.

### DIFF
--- a/sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
+++ b/sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
@@ -158,7 +158,9 @@ class MossVocoderCudaGraphRunner:
         if self._pool is None:
             self._pool = torch.cuda.graph_pool_handle()
         graph = torch.cuda.CUDAGraph()
-        with torch.cuda.graph(graph, pool=self._pool):
+        with torch.cuda.graph(
+            graph, pool=self._pool, capture_error_mode="thread_local"
+        ):
             result = self._codec._decode_frame(static_codes, static_lengths)
             static_audio = result.audio
             static_audio_lengths = result.audio_lengths
--- a/tests/unit_test/moss_tts_local/test_vocoder_cuda_graph.py
+++ b/tests/unit_test/moss_tts_local/test_vocoder_cuda_graph.py
@@ -3,6 +3,10 @@
 
 from __future__ import annotations
 
+import ast
+import inspect
+import textwrap
+
 import pytest
 import torch
 
@@ -87,6 +91,36 @@ def test_some_graphs_captured(session_bundle):
     assert (
         captured
     ), "no codec-decode CUDA graphs captured (all shapes fell back to eager)"
+
+
+def test_cuda_graph_capture_uses_thread_local_error_mode():
+    from sglang_omni.models.moss_tts_local.vocoder_cuda_graph import (
+        MossVocoderCudaGraphRunner,
+    )
+
+    source = textwrap.dedent(
+        inspect.getsource(MossVocoderCudaGraphRunner._capture_frame_count)
+    )
+    tree = ast.parse(source)
+    graph_calls = [
+        node
+        for node in ast.walk(tree)
+        if isinstance(node, ast.Call)
+        and isinstance(node.func, ast.Attribute)
+        and node.func.attr == "graph"
+        and isinstance(node.func.value, ast.Attribute)
+        and node.func.value.attr == "cuda"
+        and isinstance(node.func.value.value, ast.Name)
+        and node.func.value.value.id == "torch"
+    ]
+    assert graph_calls, "MOSS vocoder CUDA graph capture call not found"
+    assert any(
+        keyword.arg == "capture_error_mode"
+        and isinstance(keyword.value, ast.Constant)
+        and keyword.value.value == "thread_local"
+        for call in graph_calls
+        for keyword in call.keywords
+    ), "MOSS vocoder CUDA graph capture must use thread-local error mode"
 
 
 @pytest.mark.skipif(not _HAS_CUDA, reason="needs CUDA + real codec")


Ratish1 (approved):
LGTM

sglang_omni/models/moss_tts_local/vocoder_cuda_graph.py
```diff
@@ -158,7 +158,9 @@ def _capture_frame_count(self, frame_count: int) -> None:
         if self._pool is None:
             self._pool = torch.cuda.graph_pool_handle()
         graph = torch.cuda.CUDAGraph()
-        with torch.cuda.graph(graph, pool=self._pool):
+        with torch.cuda.graph(
+            graph, pool=self._pool, capture_error_mode="thread_local"
+        ):
```
Ccyest, line 163: one wondering: if vocoder needs `thread_local `to prevent capture errors from leaking to other threads, doesn't the AR decode graph capture in the same process faces the exact same problem?

If so, please add this to AR decode too

Ccyest:
overall good PR, clean scoped, well defined issue. The only question I have is commented above.

Ccyest (approved):

Ratish1:
CI Passed https://github.com/sgl-project/sglang-omni/actions/runs/27839788427

Screenshot 2026-06-20 at 12 52 47 AM

PR #830, opened 2026-06-19
[Ming]: Use thread_local capture_error_mode for CUDA graph

### LINKED ISSUE #829: Fix Ming Omni live/lazy CUDA graph capture to use thread-local capture error mode
## Background

This is related to https://github.com/sgl-project/sglang-omni/issues/825.

In #825 / the MOSS vocoder CUDA graph discussion, the root issue is not vocoder-specific. The broader invariant is:

> CUDA graph capture that may happen after serving starts should not use PyTorch’s default global capture error mode, because global mode can make the capture sensitive to CUDA activity from sibling serving threads in the same process.

The MOSS vocoder fix handles one instance of this class by passing:

```python
capture_error_mode="thread_local"
```

to `torch.cuda.graph(...)`.

@Ccyest points out that Ming Omni appears to have a similar potential issue because it has model-local CUDA graph capture sites that are reachable from generation-time lazy initialization paths.

## Problem

Ming Omni currently has at least two local CUDA graph capture paths that use plain `torch.cuda.graph(...)`:

1. `CFMGraphExecutor._initialize_graph()`

```python
with torch.cuda.graph(self.graph):
    ...
```

This is called lazily from `CFMGraphExecutor.execute()` when `self.initialized` is false.

2. `MingOmniTalker.generate()`

```python
if model_graph is None:
    model_graph = torch.cuda.CUDAGraph()
    ...
    with torch.cuda.graph(model_graph):
        outputs_placeholder = self.model(...)
```

This is inside the generation loop and captures the Qwen2 model graph when the pooled graph entry is still empty.

Although `MingTalkerExecutor._load_models()` calls `self._talker.initial_graph()` during startup, the capture sites themselves remain request-path reachable if warmup is skipped, partial, interrupted, or not exhaustive. Therefore Ming has the same potential live-capture failure class as the MOSS vocoder path: graph capture can happen while the colocated process is already doing unrelated CUDA work on sibling threads.

## Proposed fix

Update Ming’s model-local CUDA graph capture sites to use `capture_error_mode="thread_local"`.

Expected changes:

```python
# CFMGraphExecutor._initialize_graph
with torch.cuda.graph(self.graph, capture_error_mode="thread_local"):
    ...
```

and:

```python
# MingOmniTalker.generate
with torch.cuda.graph(model_graph, capture_error_mode="thread_local"):
    ...
```

This keeps the fix lifecycle-based:

* startup-only graph capture can remain unchanged;
* lazy/live/repeated graph capture during serving should use `thread_local`.

## Why this should apply to Ming

Ming is different from most other SGLang-Omni models because it has local `torch.cuda.graph(...)` calls guarded by lazy runtime state:

* `CFMGraphExecutor.execute()` initializes a graph if the executor has not yet been initialized;
* `MingOmniTalker.generate()` captures `model_graph` if the pooled graph slot is still `None`.

That means the correctness of the current behavior depends on startup warmup always fully initializing every relevant graph. If that assumption breaks, graph capture can occur during live serving.

Using `thread_local` makes the capture site robust even if it is reached during serving.

## Why other models probably do not need the same fix

This should not be applied mechanically to every CUDA graph in the repo.

The important distinction is not the model name, but the graph capture lifecycle.

### MOSS Local frame decode

MOSS Local has a model-local frame-decode CUDA graph, but its docstring explicitly states that it must run during engine init while the device is otherwise idle. Under that lifecycle, there should not be sibling serving threads issuing unrelated CUDA work during capture. Therefore it does not need the same fix unless we find a path where frame-decode graphs can be lazily captured or recaptured after serving starts.

### SGLang-backed AR models

Other models such as Qwen3-ASR, Qwen3-TTS, Qwen3-Omni, Whisper ASR, Voxtral TTS, Fish S2-Pro, and Higgs TTS generally use SGLang-managed CUDA graph initialization through `model_worker.model_runner.init_device_graphs()` during executor/scheduler construction.

Those are closer to startup-only graph captures rather than model-local lazy captures inside the request/generation path. Therefore they do not appear to need the same targeted fix in this issue.

If a future model adds local `torch.cuda.graph(...)` capture that can run after serving starts, it should follow the same rule as Ming and MOSS vocoder.

## Acceptance criteria

* Ming’s `CFMGraphExecutor._initialize_graph()` uses `capture_error_mode="thread_local"`.
* Ming’s `MingOmniTalker.generate()` model graph capture uses `capture_error_mode="thread_local"`.
* Add a regression test that inspects Ming’s source with `ast` / `inspect` and verifies local `torch.cuda.graph(...)` calls include `capture_error_mode="thread_local"`.
* The test should be scoped to Ming’s model-local lazy capture sites, not all SGLang-managed `init_device_graphs()` users.
* No behavioral change is expected for graph replay outputs.
* This issue should remain scoped to Ming. Broader SGLang core CUDA graph policy can be discussed separately if needed.

## Suggested test

Add a unit test similar to the MOSS vocoder source-level test:

## Non-goals

* Do not change every SGLang-backed AR graph capture site.
* Do not change startup-only graph capture sites unless they can be reached during live serving.

AkazaAkane:
## Motivation

This PR fixes the Ming instance of the CUDA graph capture issue described in #829. As @Ccyest point it out.

Ming Omni has model-local CUDA graph capture sites that can be reached from lazy/runtime paths, not only from guaranteed startup-only initialization. If those captures happen after serving has started, PyTorch’s default global CUDA graph capture error mode can make the capture sensitive to unrelated CUDA work from sibling serving threads in the same colocated process.

This is the same failure class discussed in #825 for the MOSS vocoder: a local CUDA graph capture should not be able to fail because another subsystem is doing unrelated CUDA work in the same process.

For Ming, the risky sites are:

- `CFMGraphExecutor._initialize_graph()`
- `MingOmniTalker.generate()`

Both can capture CUDA graphs when their local graph state has not yet been initialized. Using `capture_error_mode="thread_local"` makes these capture sites robust if they are reached during live serving.

## Modifications

This PR updates Ming’s local CUDA graph capture sites to explicitly use thread-local CUDA capture error handling:

- Pass `capture_error_mode="thread_local"` in `CFMGraphExecutor._initialize_graph()`.
- Pass `capture_error_mode="thread_local"` in `MingOmniTalker.generate()`.
- Add `tests/unit_test/ming_omni/test_talker_cuda_graph.py`.
- The new unit test parses the Ming talker source with `ast` / `inspect` and asserts that all local `torch.cuda.graph(...)` calls include `capture_error_mode="thread_local"`.

This keeps the fix scoped to Ming’s model-local lazy/live capture paths. It does not mechanically change startup-only graph initialization paths or SGLang-managed AR graph initialization.

## Related Issues

Closes #829.

Related to #825.

## Accuracy Test

This change only modifies CUDA graph capture error handling mode. It does not change model computation, graph replay logic, tensor values, sampling, or decoding behavior.

Expected accuracy impact: no output change.

Added regression coverage:

```bash
python -m pytest tests/unit_test/ming_omni/test_talker_cuda_graph.py
```
pytest and pre-commit passed.

## Benchmark & Profiling

Not applicable / not expected to affect steady-state performance.

The change only adds a keyword argument to CUDA graph capture setup. Runtime graph replay behavior should remain unchanged. The intended benefit is safer capture isolation in colocated serving, especially when graph capture can happen outside pure startup warmup.

## Checklist

* [x] Add unit tests.

## CI

long as the label remains. Use `/tag-and-rerun-ci moss` to select the MOSS TTS CI model. Draft PRs are skipped even if labeled.

### DIFF
--- a/sglang_omni/models/ming_omni/talker/modeling_ming_omni_talker.py
+++ b/sglang_omni/models/ming_omni/talker/modeling_ming_omni_talker.py
@@ -190,7 +190,7 @@ class CFMGraphExecutor:
         # (execute) checks abort before _initialize_graph and on every replay.
         self.graph = torch.cuda.CUDAGraph()
         try:
-            with torch.cuda.graph(self.graph):
+            with torch.cuda.graph(self.graph, capture_error_mode="thread_local"):
                 self.gen_lat_placeholder = self.cfm.sample(
                     self.last_hidden_state_placeholder,
                     self.his_lat_placeholder,
@@ -558,7 +558,9 @@ class MingOmniTalker(nn.Module):
                         inputs_embeds_placeholder.copy_(inputs_embeds)
                         cache_position_placeholder.copy_(cache_position)
 
-                        with torch.cuda.graph(model_graph):
+                        with torch.cuda.graph(
+                            model_graph, capture_error_mode="thread_local"
+                        ):
                             outputs_placeholder = self.model(
                                 position_ids=position_ids_placeholder,
                                 cache_position=cache_position_placeholder,
--- a/tests/unit_test/ming_omni/test_talker_cuda_graph.py
+++ b/tests/unit_test/ming_omni/test_talker_cuda_graph.py
@@ -1,0 +1,71 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Ming talker CUDA graph source-level regression tests."""
+
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+
+_REPO_ROOT = Path(__file__).resolve().parents[3]
+_TALKER_SOURCE = (
+    _REPO_ROOT
+    / "sglang_omni"
+    / "models"
+    / "ming_omni"
+    / "talker"
+    / "modeling_ming_omni_talker.py"
+)
+
+
+def _method_node(
+    tree: ast.Module, class_name: str, method_name: str
+) -> ast.FunctionDef:
+    for node in tree.body:
+        if isinstance(node, ast.ClassDef) and node.name == class_name:
+            for item in node.body:
+                if isinstance(item, ast.FunctionDef) and item.name == method_name:
+                    return item
+    raise AssertionError(f"{class_name}.{method_name} not found")
+
+
+def _torch_cuda_graph_calls(node: ast.AST) -> list[ast.Call]:
+    return [
+        item
+        for item in ast.walk(node)
+        if isinstance(item, ast.Call)
+        and isinstance(item.func, ast.Attribute)
+        and item.func.attr == "graph"
+        and isinstance(item.func.value, ast.Attribute)
+        and item.func.value.attr == "cuda"
+        and isinstance(item.func.value.value, ast.Name)
+        and item.func.value.value.id == "torch"
+    ]
+
+
+def _has_thread_local_capture_error_mode(call: ast.Call) -> bool:
+    return any(
+        keyword.arg == "capture_error_mode"
+        and isinstance(keyword.value, ast.Constant)
+        and keyword.value.value == "thread_local"
+        for keyword in call.keywords
+    )
+
+
+def test_ming_lazy_cuda_graph_captures_use_thread_local_error_mode() -> None:
+    tree = ast.parse(_TALKER_SOURCE.read_text())
+    methods = [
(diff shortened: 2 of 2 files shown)


Ratish1 (approved):
LGTM

Ccyest (approved):

Issue #831, opened 2026-06-20
[Good First Issue] Optimize Qwen3-ASR for faster TTS WER evaluation


zhaochenyang20:
There's a relatively self-contained piece of work that's a great way to get started — if you'd like to join the SGLang Omni dev team, this is a good first thing to try.

In short: SGLang Omni computes the **WER of our TTS models by transcribing their output with Qwen3-ASR**. So Qwen3-ASR's performance directly matters to us — it sits on the evaluation hot path, and CI itself has strong requirements on the Qwen3-ASR model. Making Qwen3-ASR generally faster speeds up every TTS correctness check that depends on it.

## What we're asking for

If you're interested, try optimizing Qwen3-ASR to make it **generally faster**, and **come with your results attached** — before/after numbers, the benchmark command you ran, and the hardware you ran it on.

Things to keep in mind:

- **Use the profiling tools already in the repo.** All the profilers you need are already here. Paired with an AI assistant, they should be easy to put to good use — you don't need to build measurement infra from scratch.
- **Abstract and reuse components.** Don't reimplement what already exists (server start/stop, benchmark harness, scheduler machinery). Reach for existing abstractions first; if you find yourself copy-pasting, that's a signal to extract a shared helper instead.
- **Show the numbers.** A perf claim without a benchmark behind it isn't actionable. For hot-path changes, run a real benchmark (e.g. H100/H200) and report it.

## A note on using AI

We genuinely welcome the **responsible** use of AI here. That's part of how you build trust with the SGLang Omni team: let AI help you navigate the repo, use the profilers, and draft the abstraction — but own the result. Understand the change you're proposing, validate it, and don't ship filler comments or unverified perf claims. Responsible AI use, with real results behind it, is exactly what we're looking for.

## How to get started

- Locate the Qwen3-ASR transcription path used in the TTS WER computation and the CI that exercises it.
- Profile it with the existing repo tooling to find where time actually goes.
- Propose an optimization, measure it, and open a PR with before/after numbers + the command + the hardware.

If you're interested, comment below and feel free to ask questions. Looking forward to working with you!

0xjeffro:
Hey team fyi, we implemented Qwen3 ASR when the pinned version of sglang was 0.5.8. Now we have pinned an updated version, and this model is supported in it. Using that directly is a possible solution. :)

evanderfff123-boop:
Hi! I'd like to give this issue a try.

[...]

PR #832, opened 2026-06-20
[Perf] Parallelize Qwen3-ASR mel extraction to fix concurrency scaling inversion

Billy1900:
# solving #646, supporting #831

---
  What was wrong
  
  `process_input_requests` in OmniScheduler `calls _request_builder` in a plain serial for-loop. For Qwen3-ASR, each call does a full audio decode + WhisperFeatureExtractor mel FFT — about 8ms of CPU work per sample. That's fine when requests trickle in one at a time, but at high concurrency a burst of N requests gets preprocessed sequentially before the GPU sees any of them. With `disable_overlap_schedule=True` on top, there's zero CPU-GPU overlap hiding this stall.

  This is why c=32 appeared slower than c=2 on the 20-sample CI gate: c=2 naturally pipelines (1-2 requests per round, GPU starts in ~16ms), while c=32 dumps 10 requests on a single worker and pays ~80ms upfront before the first token is prefilled.

  ---
  How it's fixed

  Refactored `process_input_requests` into three phases:

  1. Serial — stream-state setup and readiness gating. Touches shared dicts (_pending_stream_chunks, _deferred_request_payloads), must stay serial.
  2. Parallel — _request_builder calls dispatched via a persistent ThreadPoolExecutor. Mel/FFT work releases the GIL so threads run truly in parallel across CPU cores.
  3. Serial — post-build validation and waiting_queue.append. Touches shared state again, stays serial.

  Gated behind a parallel_request_build: bool = False flag on OmniScheduler so it's strictly opt-in — all existing models are unaffected by default.

  ---
  Why Qwen3-ASR specifically

  Most models in this repo take text as input, so their `_request_builder` is basically just tokenization (<0.5ms). Parallelizing that would add more thread overhead than it saves.

  Qwen3-Omni and Ming-Omni are multimodal but structured differently — they have a dedicated GPU encoder stage that runs before the AR scheduler. By the time a request reaches OmniScheduler, the heavy lifting is already done.

  Qwen3-ASR is the odd one out: it's a single-stage model that ingests raw audio bytes directly. The mel extraction happens inline inside the scheduler's request builder, on the event loop thread, with no
  separate encoder stage. That's the only reason it needs this treatment.

  stages.py enables it with parallel_request_build=True, parallel_request_build_workers=max_running_requests. Nothing else opts in.

  ---
  Results

  Benchmark on H100, 20-sample burst, single server (no router):

| concurrency | before (samples/s) | after (samples/s) | gain |
| :--- | :--- | :--- | :--- |
| 1 | 4.76 | 6.74 | +42% |
| 2 | 8.10 | 10.13 | +25% |
| 4 | 10.37 | 12.90 | +24% |
| 8 | 12.41 | 18.56 | +50% |
| 16 | 14.92 | 20.47 | +37% |
| 32 | 27.22 | 24.70 | monotonically faster than all lower levels |

  Throughput now scales cleanly at every concurrency level. The inversion is gone.


sglang_omni/models/qwen3_asr/request_builders.py
```diff
@@ -69,27 +69,53 @@ def _audio_source_from_payload(payload: StagePayload) -> Any:
...
     else:
         raise ValueError(f"Unsupported Qwen3-ASR audio input: {type(source).__name__}")
-
     if audio.ndim == 2 and audio.shape[0] > 1:
         audio = audio.mean(dim=0, keepdim=True)
-    audio = audio.squeeze(0).to(torch.float32)
+    return audio.squeeze(0).to(torch.float32).cpu().numpy(), int(sr)
+
+
+def _load_audio_soundfile(source: Any) -> tuple[np.ndarray, int]:
```
0xjeffro, line 86: Could we reuse the shared loader here? or maybe pull these audio helpers into `preprocessing/audio.py` so other models can share them too? Feels like there's a bit of overlap with what's already there.

sglang_omni/scheduling/omni_scheduler.py
0xjeffro: The profiler's request_build metric may change meaning here because they no longer bracket a single build, so might be worth keeping start/end emit tight around the actual _request_builder call.

sglang_omni/models/qwen3_asr/stages.py
```diff
@@ -119,6 +134,8 @@ def create_sglang_qwen3_asr_executor(
         model_runner=ModelRunner(model_worker, output_proc),
         request_builder=request_builder,
         result_adapter=result_adapter,
+        parallel_request_build=True,
+        parallel_request_build_workers=max_running_requests,
```
0xjeffro, line 123: A bit worried about this, each build already runs torch ops that use torch.get_num_threads() threads, so max_running_requests workers $\times$ that could oversubscribe the CPU?

zhaochenyang20:
https://github.com/sgl-project/sglang-omni/pull/833

@Billy1900 Just come across your PR. thanks for contirbution. Are there more need to resolve in #832

0xjeffro:
Hey team, I rebased this PR and benchmarked it. From my side this is a negative optimization cause by the `max workers` concern I mentioned above https://github.com/sgl-project/sglang-omni/pull/832#discussion_r3445680652.

To verify this, I ran the following experiments:
Setup: single H200, DP=1, full SeedTTS EN (1088 samples), `--concurrencies 32 --repeats 3 --warmup`. Only `parallel_request_build_workers` is changed between rows.

| config | throughput | wall mean | lat mean | rtf | request_build avg | request_build p95 |
| **PR as-is** (workers = `max_running_requests` = 32) | 20.90/s | 52.06s | 1.527s | 0.330 | **74.7 ms** | 212.8 ms |
| workers = 2 | 19.59/s | 55.55s | 1.628s | 0.352 | 37.4 ms | 50.5 ms |
| **parallel OFF** (with #833 baseline) | **25.88/s** | **42.05s** | **1.232s** | **0.267** | **28.2 ms** | 46.1 ms |

Key points: `scheduler_request_build` time (col 4) scales **monotonically with worker count**, This confirms the CPU oversubscription issue we suspected before.

So my conclusion is  although the idea looks sound, in our current setup it oversubscribes the CPU (workers $\times$ torch threads >> cores), which inflates request_build or even may starves the rest of the pipeline, so anyway it ends up slower instead of faster. I'd prefer to hold off on, happy to discuss more if anybody see something I missed.

PR #834, opened 2026-06-20
[Perf] Optimize Qwen3-ASR WER transcription

guozhihao-224:
## Summary
#831 
- Share the async Qwen3-ASR transcription client between the concurrency benchmark and TTS WER evaluation.
- Switch Qwen3-ASR TTS WER transcription fan-out from blocking `requests` + `ThreadPoolExecutor` to `aiohttp` + `BenchmarkRunner`.
- Keep Whisper / non-Qwen3 ASR paths unchanged.
- Add a bounded prompt token-id cache in the Qwen3-ASR request builder with unit coverage.

## Scope

This PR intentionally contains only Phase 1/2 code changes:

- Phase 1: shared async Qwen3-ASR client for TTS WER and isolated ASR benchmark.
- Phase 2: bounded prompt token-id cache keyed by `(num_audio_tokens, forced_language)`.

## Environment

- Machine: 1x NVIDIA A800-SXM4-80GB

## Phase 1: TTS WER transcribe-only before/after

Both runs reuse the same 100 generated WAV files.

Command shape:

```bash
$PY -m benchmarks.eval.benchmark_tts_seedtts \
  --transcribe-only \
  --use-existing-server \
  --port 8000 \
  --model fishaudio/s2-pro \
  --output-dir /tmp/qwen3-asr-wer-20260620-173136/s2pro_tts_outputs_100 \
  --lang en \
  --asr-concurrency 32
```

| Metric | Before: requests + ThreadPoolExecutor | After: aiohttp + BenchmarkRunner | Change |
|---|---:|---:|---:|
| Evaluated / Total | 100 / 100 | 100 / 100 | unchanged |
| Skipped | 0 | 0 | unchanged |
| ASR total time (s) | 5.771 | 3.800 | -34.2% |
| ASR throughput (samples/s) | 17.328 | 26.319 | +51.9% |
| ASR latency mean (s) | 1.806 | 1.088 | -39.7% |
| ASR latency p95 (s) | 2.826 | 1.861 | -34.2% |
| ASR RTF mean | 0.5332 | 0.3089 | -42.1% |
| Corpus WER | 0.00637 | 0.00728 | small run-to-run variation |
| Successful samples | 100 | 100 | unchanged |

Compatibility checks:

- `wer_results.json` top-level keys remain `config`, `per_sample`, `summary`.
- Per-sample order is unchanged between before and after.
- `asr_speed_results.json` keeps `asr_model`, `asr_concurrency`, latency, RTF, throughput, and sample-count fields.

## Phase 1 stability repeats

Same fixed 100 WAVs, `--asr-concurrency 32`, async client path.

| Repeat | ASR total time (s) | Throughput (samples/s) | Latency mean (s) | Latency p95 (s) | RTF mean | Corpus WER | Evaluated / skipped |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 3.728 | 26.822 | 1.068 | 1.833 | 0.3029 | 0.00728 | 100 / 0 |
| 2 | 3.695 | 27.061 | 1.061 | 1.777 | 0.3008 | 0.00728 | 100 / 0 |
| 3 | 3.710 | 26.957 | 1.063 | 1.769 | 0.3007 | 0.00364 | 100 / 0 |
| mean | 3.711 | 26.947 | 1.064 | 1.793 | 0.3014 | 0.00607 | 100 / 0 |

## Isolated Qwen3-ASR concurrency baseline

Command shape:

```bash
$PY -m benchmarks.eval.benchmark_qwen3_asr_concurrency \
  --port 8000 \
  --concurrencies 1,2,4,8,16,32,64 \
  --repeats 3 \
  --output /tmp/qwen3-asr-wer-20260620-173136/qwen3_asr_concurrency_baseline.json
```

| conc | reps | wall(s) mean | thrpt mean | thrpt best | lat mean(s) | lat p95(s) | rtf mean | rtf p95 | corpus WER | max WER |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 3 | 143.353 | 7.599 | 7.789 | 0.131 | 0.158 | 0.0285 | 0.0366 | 0.0143 | 0.2727 |
| 2 | 3 | 95.706 | 11.368 | 11.399 | 0.176 | 0.214 | 0.0381 | 0.0491 | 0.0134 | 0.2727 |
| 4 | 3 | 68.135 | 15.969 | 16.105 | 0.250 | 0.320 | 0.0542 | 0.0729 | 0.0147 | 0.1818 |
| 8 | 3 | 55.182 | 19.717 | 19.861 | 0.405 | 0.532 | 0.0876 | 0.1183 | 0.0138 | 0.2000 |
| 16 | 3 | 45.282 | 24.028 | 24.212 | 0.665 | 0.870 | 0.1438 | 0.1952 | 0.0139 | 0.2857 |
| 32 | 3 | 38.852 | 28.005 | 28.173 | 1.139 | 1.514 | 0.2463 | 0.3312 | 0.0138 | 0.2727 |
| 64 | 3 | 38.263 | 28.435 | 28.569 | 2.234 | 2.671 | 0.4869 | 0.6567 | 0.0129 | 0.2857 |

Profiler summary from the same baseline:

| stage | interval | count | avg ms | p95 ms |
|---|---|---:|---:|---:|
| asr | `stage_input_received->stage_complete` | 22848 | 702.896 | 2307.697 |
| asr | `scheduler_request_build_start->scheduler_request_build_end` | 22848 | 19.959 | 22.556 |

## Shared helper refactor validation

After moving the async Qwen3-ASR client into `benchmarks/tasks/qwen3_asr.py`, both TTS WER and the isolated concurrency benchmark use the same request path.

| Scope | Result |
|---|---:|
| ASR smoke, 20 samples, conc=8 | 18.449 samples/s, corpus WER 0.0104 |
| ASR smoke, 20 samples, conc=32 | 26.401 samples/s, corpus WER 0.0138 |
| TTS WER transcribe-only, 100 samples, conc=32 | 26.485 samples/s, ASR total time 3.776s |
| TTS WER ASR latency mean / p95 | 1.068s / 1.787s |
| TTS WER corpus WER | 0.0064 |
| Unit tests | 8 passed |

## Phase 2: prompt token cache validation

The request builder now caches tokenized prompts by `(num_audio_tokens, forced_language)`. The cached value is immutable (`tuple[int, ...]`) and each request gets a copied `list` before audio pad replacement, so request-specific pad values do not mutate the cache.

| Scope | Result |
|---|---:|
| Unit tests | 10 passed |
| Fixed 100 WAV unique audio-token lengths | 47 / 100 |
| Fixed 100 WAV repeated samples by token length | 82 / 100 |
| Warm TTS WER transcribe-only throughput | 26.413 samples/s |
| Warm TTS WER ASR total time | 3.786s |
| Warm TTS WER ASR latency mean / p95 | 1.089s / 1.836s |
| Warm TTS WER corpus WER | 0.0055 |
| Warm profiler request_build avg / p95 | 21.038ms / 23.638ms |

## Validation commands

```bash
python3 -m py_compile \
  benchmarks/tasks/qwen3_asr.py \
  benchmarks/tasks/tts.py \
  benchmarks/eval/benchmark_qwen3_asr_concurrency.py \
  sglang_omni/models/qwen3_asr/request_builders.py \
  tests/unit_test/qwen3_asr/test_request_builders.py
```

Remote:

```bash
$PY -m pytest tests/unit_test/qwen3_asr -q
# 10 passed
```


benchmarks/tasks/asr_transcription.py
zhaochenyang20, line 1: change the name of this file. Task is describing the workload.

tests/unit_test/qwen3_asr/test_request_builders.py
```diff
@@ -117,6 +119,87 @@ def test_qwen3_asr_request_builder_records_inclusive_audio_offsets(monkeypatch)
     )
 
 
+def test_qwen3_asr_prompt_token_ids_are_cached_by_audio_length_and_language(
+    monkeypatch,
+) -> None:
+    num_mel_frames = 101
```
zhaochenyang20: why there is the magic number

zhaochenyang20 (changes requested):

0xjeffro:
Hi thanks for the work! I have a question, just have some doubts about phase 2(the prompt token cache):
CleanShot 2026-06-20 at 22 19 07@2x

Does this mean the prompt cache optimization has negligible effect or even introduces extra overhead? My understanding is the cache key is based on num_audio_tokens + language, and since audio lengths are diverse in test data, the hit rate is limited. Is the added complexity worth such a marginal gain? I'd rather go after the big fish first.

benchmarks/tasks/asr_transcription.py
```diff
@@ -0,0 +1,112 @@
...
+            content_type="audio/wav",
+        )
+
+        start_time = time.perf_counter()
+        try:
+            async with session.post(api_url, data=form) as response:
+                if response.status != 200:
+                    result.error = f"HTTP {response.status}: {await response.text()}"
+                else:
+                    payload = await response.json()
```
zhaochenyang20, line 74: Also, the TTS WER path now uses this new shared async client, and line 75 reads a 200 response with `payload.get("text", "")` before line 76 marks the request as successful unconditionally. The old TTS Qwen3-ASR path in `benchmarks/tasks/tts.py` used `response.json()["text"]`, so a router body that violated the OpenAI transcription contract surfaced as a protocol error; this PR turns the same TTS WER case into an empty transcription. That means a response without `text` enters the `calculate_wer_metrics` successes set, keeps `evaluated` unchanged, computes WER against an empty hypothesis, and also counts ASR latency and throughput as a successful sample. For the correctness gate and performance regression analysis, that is not an acceptable downgrade because it disguises a server/schema bug as a model transcription.

zhaochenyang20: Keep validating the schema for 200 responses. The minimal fix is to read `payload["text"]` in the shared helper, or explicitly require `"text" in payload` with a string value; if it is missing, set `result.error` and leave `is_success=False` instead of synthesizing an empty transcript. That gives the standalone benchmark and TTS WER the same fail-fast endpoint-contract semantics while still allowing a real empty transcription to be represented by `"text": ""`.

benchmarks/tasks/tts.py
```diff
@@ -289,6 +288,67 @@ def _transcribe_qwen3_asr(asr: dict, wav_path: str, lang: str) -> str:
...
+    lang: str,
+    *,
+    concurrency: int,
+) -> list[tuple[int, dict, SampleOutput]]:
+    samples = [
+        SimpleNamespace(
+            sample_id=str(entry.get("sample_id", "")),
+            ref_audio=entry["wav_path"],
+        )
+        for _idx, entry, _output in pending
```
zhaochenyang20, line 331: In the new Qwen3 async path, line 1042 has already used `entry["sample_id"]` to build the `SampleOutput`, so by the time line 328 builds the `SimpleNamespace`, `sample_id` is a required field and should not fall back through `entry.get("sample_id", "")`. This fallback does not protect a real input error; it creates two different contracts instead: the output object requires `sample_id`, while the sample sent to `BenchmarkRunner` allows an empty request id.

zhaochenyang20: Change line 328 to `sample_id=entry["sample_id"]`, or define a small typed sample dataclass that carries `sample_id` and `ref_audio`. Do not introduce an empty-string default after the internal code has already proven the field is required; if an old format must be supported, do explicit migration or validation at the `generated.json` loading boundary.

zhaochenyang20 (changes requested):
Your code is quite clear. Good!

guozhihao-224:
> Hi thanks for the work! I have a question, just have some doubts about phase 2(the prompt token cache): CleanShot 2026-06-20 at 22 19 07@2x
> 
> Does this mean the prompt cache optimization has negligible effect or even introduces extra overhead? My understanding is the cache key is based on num_audio_tokens + language, and since audio lengths are diverse in test data, the hit rate is limited. Is the added complexity worth such a marginal gain? I'd rather go after the big fish first.

Yes, in the current scenario, this cache offers almost no performance improvement.

luojiaxuan:
Profiling update for `benchmark_tts_seedtts.py --transcribe-only`:

Setup: generated 1088 Higgs TTS WAVs first, then reused the same audio against the same `Qwen/Qwen3-ASR-1.7B` server. This compares the client path only.

No request profiler:

| client | ASR conc. | total time | throughput | mean latency | p95 latency |
| main | 16 | 23.898s | 45.53/s | 0.351s | 0.490s |
| #834 | 16 | 24.972s | 43.57/s | 0.361s | 0.499s |
| main | 32 | 19.713s | 55.19/s | 0.578s | 0.792s |
| #834 | 32 | 19.588s | 55.54/s | 0.563s | 0.773s |
| main | 64 | 19.724s | 55.16/s | 1.149s | 1.464s |
| #834 | 64 | 20.040s | 54.29/s | 1.145s | 1.414s |

Conclusion:

- At concurrency 32, #834 is essentially tied with main: 55.54/s vs 55.19/s (+0.6%).
- Concurrency 64 does not improve throughput, so this workload is already ASR-server saturated around ~55 samples/s.
- Request build is not the bottleneck here: request-level profiling shows `scheduler_request_build` is only ~9-10 ms/request.
- I saw one slow main run under request profiling, but it did not reproduce; a repeat profile run was 53.31/s. So I would not use the slow profiled run as evidence.

My read: the small gain is expected for this benchmark at c=32. The old `requests + ThreadPoolExecutor` path can already feed the ASR server fast enough, so #834 improves the client implementation but does not materially raise saturated ASR throughput for this workload.

keke0315:
Following precious reviews,supplementary profiling update for benchmark_tts_seedtts.py --transcribe-only at c=32 to further validate the performance gain.

Under same settings to pr:

- Both runs reused the same 100 WAV files generated by Fish Audio S2-pro.
- All experiments were repeated three times
- asr-concurrency 32

## Environment
- Machine: on 1x NVIDIA H100 80GB HBM3.

| Metric | Main | PR #834 |
|----------|----------|----------|
| Evaluated / Total | 100 / 100 | 100 / 100 |
| Skipped | 0 | 0 |
| ASR total time (s) | 4.62 / 4.71 / 4.67 | 5.12 / 5.26 / 5.35 |
| ASR throughput (samples/s) | 21.6 / 21.2 / 21.4 | 19.5 / 19.0 / 18.7 |
| ASR latency mean (s) | 1.44 / 1.48 / 1.45 | 1.51 / 1.50 / 1.54 |
| ASR latency p95 (s) | 2.36 / 2.42 / 2.46 | 3.11 / 3.10 / 3.16 |
| ASR RTF mean | 0.43 / 0.44 / 0.43 | 0.44 / 0.44 / 0.45 |
| Corpus WER | 0.0045 / 0.0064 / 0.0073 | 0.0082 / 0.0082 / 0.0091 |
| Successful samples | 100 / 100 | 100 / 100 |

In my environment, the PR branch shows a small increased in latency and longer total time, lower throughput.

PR #839, opened 2026-06-20
[RFC] Trimming for FishAudio S2-Pro

GaokaiZhang:
First per-model trim from #661 on S2-Pro. Scope narrowed to the uncontroversial part after
@Ratish1's review.

What's here — deletions only (−62):
- The decorative `# ---` section banners above the stage factories in `stages.py`, and the
  `# -- Helpers --` banner in `payload_types.py`.
- Two vocoder tests that only re-run `StreamingSimpleScheduler` lifecycle already covered by
  `tests/unit_test/pipeline/test_streaming_simple_scheduler.py`. The S2-Pro-specific tests
  (crossfade, VQ injection, abort isolation, top-k validation) stay.

Kept per @Ratish1: the inline comments in `sglang_model.py` and the `request_builders.py` /
`tokenizer.py` class docstrings.

No behaviour change — non-functional edits plus two redundant tests; S2-Pro unit suite is green and
a full SeedTTS-EN run on the branch showed no regression (WER/throughput within noise).

The bigger lever, for a separate RFC: S2-Pro is the last AR TTS model still on a bespoke scheduler —
`fish_scheduler.py` (~591 LOC) reimplements what `OmniScheduler` already gives the other models.
Worth migrating (like #476 did for Higgs), but on its own, with a SeedTTS before/after gate.


sglang_omni/models/fishaudio_s2_pro/request_builders.py
```diff
@@ -16,7 +16,6 @@
 
 @dataclass
 class S2ProSGLangRequestData(SGLangARRequestData):
-    """S2-Pro per-request state."""
```
Ratish1, line 19: I think we can keep this?

sglang_omni/models/fishaudio_s2_pro/tokenizer.py
```diff
@@ -34,9 +34,6 @@ class Reference:
 
 
 class S2ProTokenizerAdapter:
-    """
-    Builds Qwen3 chat-format prompts using the ``Conversation`` class
-    """
```
Ratish1, line 39: same

sglang_omni/models/fishaudio_s2_pro/sglang_model.py
```diff
@@ -254,7 +254,6 @@ def setup_vq_decode(
         """Attach audio decoder and allocate persistent GPU buffers."""
         device = self.embed_tokens.weight.device
 
-        # Audio decoder (fast head)
```
Ratish1, line 257: we can keep these inline comments as well I think for this file. wdyt

Ratish1 (approved):
Thanks for the cleanup @GaokaiZhang . Left some comments. Lmk what do you think. Rest look good to me.

Ratish1:
Lets change the title as well @GaokaiZhang . Thanks

GaokaiZhang:
@Ratish1 thanks for the prompt review; resolved.

GaokaiZhang:
@Ratish1 @zhaochenyang20 gentle ping — approved and deletions-only, just waiting on the `run-ci` label. The M9 fish → OmniScheduler migration rewrites the same `stages.py` region, so I'd like this to land first.

Issue #836, opened 2026-06-20
[Tracking] Standardize CUDA graph / batch-size handling across SGLang-backed generation stages


zhaochenyang20:
PR #756 ([Higgs TTS] Raise AR server default to 64 and expose standard batch knobs) bumped the Higgs AR serving default to `64/64` and introduced a **generic generation-stage override path** so that the two standard SGLang engine parameters — `max_running_requests` and `cuda_graph_max_bs` — can be set on any SGLang-backed generation stage via the CLI, instead of through a Higgs-only helper.

Two artifacts from that PR are relevant here:

1. **A concrete Higgs bug was fixed.** The model-side CUDA graph / sampler buffers were effectively sized around an old internal max-batch assumption. With larger capture sizes (e.g. `cuda_graph_max_bs=128`), capture failed in Higgs decode embedding with a batch-shape mismatch. #756 now resolves the Higgs internal max batch size from SGLang `max_running_requests`, so the model-side buffers match the serving configuration.
2. **A generic contract was added.** `PipelineConfig.generation_sglang_role_to_stage()` replaced the CLI-side `tts_engine` stage-name guess, and the SGLang-backed generation configs now declare their generation stage explicitly (Higgs / Qwen3-TTS / MOSS / MOSS-Local / FishAudio S2-Pro → `tts_engine`; Qwen3-Omni speech → `talker_ar`; Voxtral TTS → `tts_generation`).

#756 was deliberately scoped to the Higgs default bump + the generic override path. The items below were intentionally **left out of #756 to keep it reviewable**, and are tracked here so they don't get lost.

The unifying concern is the same one observed in [[#721 Profiling AR-stage concurrency knees](https://github.com/sgl-project/sglang-omni/issues/721)](https://github.com/sgl-project/sglang-omni/issues/721) and [[#781 CUDA Graph capture-size scaling regression at low concurrency](https://github.com/sgl-project/sglang-omni/issues/781)](https://github.com/sgl-project/sglang-omni/issues/781): **the Higgs mismatch was probably not Higgs-specific.** Other SGLang-backed generation stages may carry similar hidden, hard-coded batch assumptions, and the per-model batch conventions have drifted apart. We want each of the three workstreams below to land as its own PR rather than as one large change.

---

## Workstream 1 — Audit CUDA graph batch sizing across SGLang-backed generation stages

[...]

PR #843, opened 2026-06-21
[TTS] Validate generation batch policies (#836 W1)

luojiaxuan:
## Summary

Addresses Workstream 1 from #836.

- Add a reusable generation batch policy report/validator for SGLang-backed generation stages.
- Standardize explicit CUDA graph batch policy construction and override syncing for generation stages.
- Use explicit CUDA graph batch lists that match SGLang's normal default capture buckets, so validation does not narrow graph coverage.
- Remove the old partial/require validation knobs: CUDA graph now requires explicit `cuda_graph_bs` and full `max_running_requests` coverage when enabled.
- Enforce `torch_compile_max_bs` coverage only when `enable_torch_compile=True`; S2-Pro now uses 64/64/64 for `max_running_requests`, `cuda_graph_max_bs`, and `torch_compile_max_bs`.
- Validate model-side batch buffers for Higgs TTS and FishAudio S2-Pro using explicit model properties / allocated buffer capacity.

## TTS Coverage

Covered models/stages: Higgs TTS, MOSS-TTS, MOSS-TTS Local, Qwen3-TTS, Voxtral TTS, FishAudio S2-Pro, and Qwen3-Omni talker AR.

## Validation

- `python -m pytest tests/unit_test/serve/test_generation_batch_policy.py tests/unit_test/higgs_tts/test_pipeline.py tests/unit_test/fishaudio_s2_pro/test_pipeline.py tests/unit_test/moss_tts/test_pipeline.py tests/unit_test/moss_tts_local/test_pipeline.py tests/unit_test/qwen3_tts/test_pipeline.py tests/unit_test/voxtral_tts/test_pipeline.py tests/unit_test/qwen3_omni/test_sglang_ar_budget.py -q` -> 233 passed
- `git diff --check` -> passed

Generation-only benchmark comparison, same H100 GPU/container. Baseline is `origin/main` at `8a41fc4`; PR is `f2a886c`. Suspicious/noisy rows were repeated to 3 total runs. For repeated rows, values are median (min-max), and delta uses medians. Qwen3-TTS baseline uses a rerun to replace the original outlier run1.

| Model | Samples / concurrency | Runs | PR req/s | Baseline req/s | Req delta | PR audio s/s | Baseline audio s/s | Audio delta | PR latency mean (s) | Baseline latency mean (s) |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Higgs TTS | 1088 / 16 | 1 | 15.473 | 15.410 | +0.4% | 64.928 | 64.527 | +0.6% | 1.028 | 1.032 |
| FishAudio S2-Pro | 256 / 64 | 3 | 1.268 (1.259-1.338) | 1.320 (1.303-1.384) | -3.9% | 4.792 (4.761-5.069) | 4.990 (4.916-5.235) | -4.0% | 44.929 (42.551-45.151) | 43.178 (41.108-43.686) |
| MOSS-TTS | 256 / 16 | 3 | 3.933 (3.854-4.021) | 3.864 (3.521-4.051) | +1.8% | 17.044 (16.707-17.416) | 16.753 (15.260-17.522) | +1.7% | 3.996 (3.908-4.081) | 4.066 (3.885-4.479) |
| Voxtral TTS | 50 / 16 | 3 | 7.453 (7.325-7.637) | 7.321 (5.918-7.731) | +1.8% | 39.615 (39.354-41.008) | 39.169 (31.274-42.132) | +1.1% | 1.960 (1.877-1.999) | 2.005 (1.844-2.472) |
| Qwen3-TTS 0.6B | 50 / 16 | 3 | 2.537 (2.466-2.554) | 2.606 (2.562-2.628) | -2.6% | 10.091 (9.752-10.550) | 10.591 (10.461-10.858) | -4.7% | 5.398 (5.346-5.554) | 5.444 (5.181-5.462) |

Notes:
- S2-Pro startup confirmed PR Fast AR compile `max_batch_size=64` and CUDA graph batches `[1, 2, 4, 8, 12, 16, 24, 32, 40, 48, 56, 64]`. Baseline compiled Fast AR with `max_batch_size=16` and the same CUDA graph buckets.
- Qwen3-TTS baseline original run1 was retested and excluded from the table because it generated many more output tokens (`4542`) and had p99 latency `47.441s`. The rerun normalized to `2509` output tokens, `2.606` req/s, and p99 latency `10.527s`.
- Voxtral had a single-run baseline outlier. After repeats, the large single-run speedup is not stable, so this PR does not claim it as a real gain.
- Artifact roots: `/data/sglang-omni-profile-runs/pr843-bench-defaultcg-20260621`, `/data/sglang-omni-profile-runs/pr843-baseline-main-20260621`, and `/data/sglang-omni-profile-runs/pr843-repeat-20260621`.

### DIFF
--- a/sglang_omni/models/fishaudio_s2_pro/stages.py
+++ b/sglang_omni/models/fishaudio_s2_pro/stages.py
@@ -18,7 +18,7 @@ from sglang_omni.models.fishaudio_s2_pro.request_builders import (
 )
 from sglang_omni.proto import StagePayload
 from sglang_omni.scheduling.generation_batch_policy import (
-    build_power_of_two_cuda_graph_bs,
+    build_default_cuda_graph_bs,
     sync_cuda_graph_bs_with_max_bs,
     validate_generation_batch_policy,
 )
@@ -251,7 +251,7 @@ def create_sglang_tts_engine_executor(
     patch_fish_config_for_sglang()
 
     overrides: dict[str, Any] = {
-        "cuda_graph_bs": build_power_of_two_cuda_graph_bs(64),
+        "cuda_graph_bs": build_default_cuda_graph_bs(64),
         "cuda_graph_max_bs": 64,
         "disable_cuda_graph": False,
         "mem_fraction_static": 0.85,
--- a/sglang_omni/models/fishaudio_s2_pro/stages.py
+++ b/sglang_omni/models/fishaudio_s2_pro/stages.py
@@ -55,6 +55,13 @@ def _compile_s2pro_codebook_decoder(model: Any, *, max_batch_size: int) -> None:
     )
 
 
+def _resolve_s2pro_model_buffer_bs(model: Any) -> int:
+    return min(
+        int(model.vq_decode_max_batch_size),
+        int(model._audio_decoder.kv_cache_max_batch_size),
+    )
+
+
 def _resolve_checkpoint(checkpoint: str) -> str:
     if os.path.isdir(checkpoint):
         return checkpoint
@@ -252,7 +259,7 @@ def create_sglang_tts_engine_executor(
         "chunked_prefill_size": 8192,
         "dtype": "bfloat16",
         "enable_torch_compile": True,
-        "torch_compile_max_bs": 16,
+        "torch_compile_max_bs": 64,
         "random_seed": int.from_bytes(os.urandom(4), "little") & 0x7FFFFFFF,
     }
     if server_args_overrides:
@@ -306,8 +313,7 @@ def create_sglang_tts_engine_executor(
     validate_generation_batch_policy(
         model_name="FishAudio S2-Pro",
         server_args=server_args,
-        model_buffer_bs=server_args.max_running_requests,
-        allow_partial_torch_compile_coverage=True,
+        model_buffer_bs=_resolve_s2pro_model_buffer_bs(model_worker.model_runner.model),
     )
 
     if bool(getattr(server_args, "enable_torch_compile", False)):
--- a/sglang_omni/models/fishaudio_s2_pro/stages.py
+++ b/sglang_omni/models/fishaudio_s2_pro/stages.py
@@ -17,6 +17,10 @@ from sglang_omni.models.fishaudio_s2_pro.request_builders import (
     make_tts_scheduler_adapters,
 )
 from sglang_omni.proto import StagePayload
+from sglang_omni.scheduling.generation_batch_policy import (
+    build_power_of_two_cuda_graph_bs,
+    sync_cuda_graph_bs_with_max_bs,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -239,6 +243,8 @@ def create_sglang_tts_engine_executor(
     patch_fish_config_for_sglang()
 
     overrides: dict[str, Any] = {
+        "cuda_graph_bs": build_power_of_two_cuda_graph_bs(64),
+        "cuda_graph_max_bs": 64,
         "disable_cuda_graph": False,
         "mem_fraction_static": 0.85,
         "max_running_requests": 64,
@@ -250,6 +256,7 @@ def create_sglang_tts_engine_executor(
     }
     if server_args_overrides:
         overrides.update(server_args_overrides)
+        sync_cuda_graph_bs_with_max_bs(overrides, server_args_overrides)
 
     server_args = build_sglang_server_args(
         checkpoint_dir,
--- a/sglang_omni/models/fishaudio_s2_pro/stages.py
+++ b/sglang_omni/models/fishaudio_s2_pro/stages.py
@@ -20,6 +20,7 @@ from sglang_omni.proto import StagePayload
 from sglang_omni.scheduling.generation_batch_policy import (
     build_power_of_two_cuda_graph_bs,
     sync_cuda_graph_bs_with_max_bs,
+    validate_generation_batch_policy,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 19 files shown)


sglang_omni/models/fishaudio_s2_pro/stages.py
zhaochenyang20, line 1: What if we unify all the model's CUDA graph bs? Why should we set partial-coverage to S2 Pro and require_cuda_graph_bs=False to Higgs? As long as the CI is covered and the user can launch the model with the default parameter to get the best performance, it's good.

zhaochenyang20 (changes requested):

tests/unit_test/serve/test_generation_batch_policy.py
zhaochenyang20, line 1: Add the test to `/tests/readme.md`

zhaochenyang20 (changes requested):
Just a bit curious, why the previous validation codes for Higgs CUDA Graph is not removed? This PR seems to be purely adding lines with few deletions.

sglang_omni/scheduling/generation_batch_policy.py
```diff
@@ -0,0 +1,201 @@
...
+                f"({cuda_graph_max_bs} < {max_running_requests})"
+            )
+
+    torch_compile_max_bs = _read_positive_int(
+        server_args,
+        "torch_compile_max_bs",
+        errors,
+        required=False,
+    )
+    if (
```
zhaochenyang20, line 107: `validate_generation_batch_policy()` should read `enable_torch_compile` first and require `torch_compile_max_bs` to cover `max_running_requests` only when generic compile is actually enabled.

the Higgs generation stage defaults to `max_running_requests=64` without enabling generic `torch.compile`. A real `ServerArgs` still carries the default `torch_compile_max_bs=32`; the current validator ignores `enable_torch_compile` and raises as soon as it sees 32 < 64. That makes the default Higgs TTS launch fail during stage startup, which is not the same as reporting a genuine batch-policy disagreement.

sglang_omni/models/fishaudio_s2_pro/stages.py
zhaochenyang20: unify S2-Pro compile coverage by making `torch_compile_max_bs` follow `max_running_requests` and `cuda_graph_max_bs` to 64

zhaochenyang20: `model_buffer_bs=server_args.max_running_requests` does not read any actual model-side buffer. S2-Pro's real buffers are allocated in `bootstrap_text_model_for_decode()` through `audio_decoder.setup_caches(max_batch_size=...)` and `text_model.setup_vq_decode(... max_batch_size=...)`; the current validator only passes back the request limit that was just used for allocation. If the audio decoder cache later reuses a smaller existing cache, the text-model VQ buffer is renamed, or bootstrap starts clamping, this check still passes and misses exactly the model-buffer drift W1 is meant to catch. The runtime Impact is that a future buffer regression can pass startup validation and surface only later as a high-concurrency shape failure or cache bounds failure.

sglang_omni/models/higgs_tts/stages.py
zhaochenyang20, line 1: Higgs is the only stage in this PR that passes a model-side sampler-pool size into the validator, but it does so via `getattr(model, "_max_batch_size", None)`. If W2 or a later cleanup renames `_max_batch_size` to a more accurate name, this call does not fail; it passes `None` into the validator and removes the model-buffer coverage check entirely. The goal of this PR is to expose hidden batch assumptions at startup, and a defaulted dynamic attribute lookup hides the same class of problem again.

Give the Higgs model an explicit property, such as `sampler_pool_max_running_requests`, and read it directly from the stage; in the short term, at least use `model._max_batch_size` so a missing attribute fails fast. Do not treat `None` as a successful "cannot validate but continue" state.

tests/unit_test/higgs_tts/test_pipeline.py
zhaochenyang20, line 1: the current fake `server_args` contains only `disable_cuda_graph`, `max_running_requests`, and `cuda_graph_max_bs`; it omits `enable_torch_compile=False` and the real `ServerArgs` default `torch_compile_max_bs=32`. That lets `test_higgs_tts_engine_enables_cuda_graph_by_default` keep passing after the new validator, while it does not cover the default Higgs startup failure from the P0 issue. The test is validating a stub with fewer fields than the production object, not the default startup contract.

Change this test to use a lightweight real `build_sglang_server_args()` construction, or at least make the fake include `enable_torch_compile=False`, `torch_compile_max_bs=32`, and `cuda_graph_bs=None`, then assert after the validator fix that default Higgs 64/64 startup does not fail because of a disabled compile field. Add a validator unit test for the legal case where `enable_torch_compile=False` and `torch_compile_max_bs < max_running_requests`.

zhaochenyang20 (changes requested):

sglang_omni/models/fishaudio_s2_pro/stages.py
luojiaxuan: Done. S2-Pro now uses 64/64/64, removes partial compile coverage, and validates allocated text VQ + Fast AR KV-cache capacity.

tests/unit_test/serve/test_generation_batch_policy.py
luojiaxuan: Done

sglang_omni/scheduling/generation_batch_policy.py
luojiaxuan: Done. Compile coverage is enforced only when enable_torch_compile is true, with disabled-compile unit coverage.

sglang_omni/models/higgs_tts/stages.py
luojiaxuan: Done. Added sampler_pool_max_running_requests and read it directly; no None fallback.

tests/unit_test/higgs_tts/test_pipeline.py
luojiaxuan: Done. The Higgs fake now includes disabled compile, torch_compile_max_bs=32, and explicit cuda_graph_bs; also added the validator legal case.

luojiaxuan:
Updated in 0c37699: removed the old partial/require validation knobs, unified explicit CUDA graph policy, and validated Higgs/S2-Pro defaults. SeedTTS Higgs gen-only stayed flat: 1088/1088, 0 failed, 65.486 audio s/s.

zhaochenyang20 (approved):

PR #846, opened 2026-06-21
[TTS] Standardize generation batch conventions (#836 W3)

luojiaxuan:
## Summary

Addresses Workstream 3 from #836.

- Add shared generation batch helpers for SGLang default CUDA graph buckets and `cuda_graph_max_bs` override syncing.
- Make Qwen3-TTS, Voxtral TTS, MOSS-TTS, MOSS-TTS Local, and FishAudio S2-Pro use the shared CUDA graph bucket defaults.
- Explicitly set missing `cuda_graph_max_bs` defaults for Qwen3-TTS, Voxtral TTS, and FishAudio S2-Pro.
- Align FishAudio S2-Pro `torch_compile_max_bs` with its default `max_running_requests=64`.

## Benchmarks

Generation-only SeedTTS benchmark on the same H100 GPU. Baseline: `origin/main` (`eaedf63`). PR: `1adada0`.

| Model | n / conc | Base req/s | PR req/s | Δ req/s | Base tok/s | PR tok/s | Δ tok/s | Notes |
|---|---:|---:|---:|---:|---:|---:|---:|---|
| MOSS-TTS | 256 / 16 | 3.808 | 3.782 | -0.7% | 320.8 | 318.3 | -0.8% | seed=0, 2 runs |
| MOSS-Local | 128 / 16 | 3.675 | 3.687 | +0.3% | 202.3 | 234.0 | +15.6% | seed=0, output length varied |
| Voxtral | 128 / 16 | 9.955 | 10.598 | +6.5% | 628.6 | 672.5 | +7.0% | `--no-ref-audio --voice cheerful_female` |
| Qwen3-TTS-0.6B | 50 / 16 | 2.383 | 2.116 | -11.2% | 127.5 | 123.5 | -3.2% | `temperature=0`, `max_new_tokens=256`, 2 runs |
| FishAudio S2-Pro | 128 / 64 | 1.232 | 1.264 | +2.6% | 93.9 | 97.3 | +3.7% | manual server, seed=0, 2 runs |

Notes:
- Qwen3 default sampling with `seed=0` produced one PR request that ran to 2048 tokens, so I used the constrained Qwen3 run above for a less noisy throughput check.
- S2-Pro logs confirm the intended change: Fast AR compile max batch size is `16` on baseline and `64` on this PR; CUDA graph buckets are unchanged at `[1, 2, 4, 8, 12, 16, 24, 32, 40, 48, 56, 64]`.

## Validation

- `pre-commit run --all-files --show-diff-on-failure`
[...]

PR #850, opened 2026-06-22
[TTS] Standardize generation batch conventions (#836)

luojiaxuan:
## Summary

Continues #846 from the fork branch so yijiangt can collaborate there, and incorporates #844.

- Add the CUDA graph batch validator from #844 for generation-stage capture/buffer audits.
- Replace the legacy `sync_cuda_graph_bs_with_max_bs` call pattern with shared generation batch override construction.
- Standardize default `cuda_graph_bs` / `cuda_graph_max_bs` plumbing across the SGLang-backed generation stages.
- Remove stale per-stage fallback logic that recomputed CUDA graph buckets after `server_args` had already been populated.
- Finish the Higgs sampler-pool naming cleanup by replacing the stale `_max_batch_size` name with a value named for its actual role.

Part of #836.

## H100 benchmark

Same H100 host, `CUDA_VISIBLE_DEVICES=0,1`, Higgs TTS CI non-streaming c16 speed stage.

Baseline: `main` @ `1a34db7`  
PR: `086b2a8`

| Metric | Baseline | PR | Delta |
|---|---:|---:|---:|
| Completed requests | 1088 | 1088 | 0 |
| Throughput (req/s) | 18.049 | 17.628 | -2.3% |
| Latency mean (s) | 0.882 | 0.902 | +2.3% |
| Latency p95 (s) | 1.265 | 1.257 | -0.6% |
| RTF mean | 0.207 | 0.210 | +1.4% |
| RTF p95 | 0.2564 | 0.2674 | +4.3% |

No material performance regression observed; both runs completed the full speed stage with zero failed requests.

### Startup / CUDA graph capture split

Same H100 host, 2 managed Higgs workers. The container exposed host GPUs 6/7 as worker-visible `CUDA_VISIBLE_DEVICES=0,1`. This measures startup until both workers are healthy, then stops immediately; it does not run the full 1088-sample speed stage.

Baseline: `sgl-project/main` @ `86e73bd`  
PR: `602996d`

I discarded the first baseline split run because it included one-time graph/kernel warmup (`Capture cuda graph end` around 20.8-20.9s). The table uses the warm-cache `r2` run for both refs; per-worker rows are averaged across the two managed workers.

| Startup metric | Baseline | PR | Delta |
|---|---:|---:|---:|
| Router startup wall time to 2/2 healthy workers | 29.03s | 30.03s | +1.00s (+3.4%) |
| `tts_engine` build -> memory-pool ready | 2.26s | 2.26s | +0.00s (+0.2%) |
| SGLang backend config -> memory-pool ready | 2.00s | 2.00s | -0.00s (-0.1%) |
| CUDA graph capture per worker | 2.64s | 2.73s | +0.10s (+3.6%) |
| Post-capture Higgs model/stage setup | 0.18s | 0.20s | +0.02s |
| `tts_engine` build -> stage started | 5.16s | 5.28s | +0.12s (+2.2%) |
| Explicit `init_device_graphs()` call | n/a | n/a | Higgs captures during SGLang worker init |

Both refs captured the same graph buckets: `[1, 2, 4, 8, 12, 16, 24, 32, 40, 48, 56, 64]`. I do not see a material startup or graph-capture regression. Logs are under `/data/jiaxuanluo/pr850-review/runs` on `host-85-234-79-62`.

## Validation

- `python -m pytest tests/unit_test/serve/test_generation_batch_policy.py tests/unit_test/model_runner/test_cuda_graph_batch_validator.py tests/unit_test/qwen3_tts/test_pipeline.py tests/unit_test/voxtral_tts/test_pipeline.py tests/unit_test/fishaudio_s2_pro/test_pipeline.py tests/unit_test/moss_tts/test_pipeline.py tests/unit_test/moss_tts_local/test_pipeline.py tests/unit_test/higgs_tts/test_pipeline.py -q` -> 260 passed
- H100 focused review-fix tests at `62990f6`: `python3 -m pytest tests/unit_test/serve/test_generation_batch_policy.py tests/unit_test/model_runner/test_cuda_graph_batch_validator.py tests/unit_test/qwen3_asr/test_pipeline.py tests/unit_test/whisper_asr/test_pipeline.py -q` -> 48 passed
- H100 targeted pre-commit at `62990f6`: `pre-commit run --files sglang_omni/models/qwen3_asr/stages.py sglang_omni/models/whisper_asr/stages.py sglang_omni/scheduling/bootstrap.py sglang_omni/utils/cuda_graph_batch_validator.py tests/unit_test/model_runner/test_cuda_graph_batch_validator.py tests/unit_test/qwen3_asr/test_pipeline.py tests/unit_test/whisper_asr/test_pipeline.py`
- H100 speed stage: `python -m pytest tests/test_model/test_tts_ci.py -v -s -x --concurrency 16 --tts-stage tts-stage-1-nonstream -k test_voice_cloning_non_streaming` -> 1 passed, 6 deselected


sglang_omni/utils/cuda_graph_batch_validator.py
```diff
@@ -0,0 +1,278 @@
...
+    buffer_source: str | None = None,
+) -> CudaGraphBatchReport:
+    """Compare the three batch-size facts and produce a verdict."""
+    findings: list[str] = []
+    ok = True
+
+    captured = [b for b in (captured_bs or []) if b is not None]
+    max_captured = max(captured) if captured else None
+
+    if buffer_capacity is not None and max_captured is not None:
```
SandyLuXY, line 139: `evaluate_cuda_graph_batch_sizing()` only checks whether `max(captured_bs)` exceeds the model buffer. It never checks whether captured graphs actually cover the configured runnable batch size. A stage configured with `max_running_requests=64`, `cuda_graph_max_bs=64`, `request_slots=64`, and `buffer_capacity=64` is reported as OK even if only `[1,2,4,8,16]` were captured.

Suggested fix: compare `max_captured_bs` against the effective expected target, roughly `min(cuda_graph_max_bs, max_running_requests, request_slots)` when those are available, and flag undercoverage.

SandyLuXY:
@luojiaxuan @yijiangt Nice work! Could you add a startup / CUDA-graph capture-time comparison for this PR? The Higgs c16 benchmark is useful, and it would help to also show whether the new graph-capture sequencing and bucket plumbing changes server startup time.

sglang_omni/utils/cuda_graph_batch_validator.py
luojiaxuan: Good catch, fixed in 2360121.

`evaluate_cuda_graph_batch_sizing()` now computes the effective expected capture target from the available `cuda_graph_max_bs`, `max_running_requests`, and `request_slots` values, then flags a mismatch when `max(captured_bs)` is below that target. I also added a regression test for the exact undercoverage shape you called out: `64/64/64` config with only `[1, 2, 4, 8, 16]` captured.

Local focused validation:

```text
python -m pytest tests/unit_test/serve/test_generation_batch_policy.py tests/unit_test/model_runner/test_cuda_graph_batch_validator.py -q
# 38 passed
```

luojiaxuan:
Startup / CUDA-graph capture-time comparison for the Higgs managed-router startup path:

Setup:
- Host: `host-85-234-79-62`, 2x H100 workers via `CUDA_VISIBLE_DEVICES=6,7`.
- Path: same Higgs TTS managed-router startup path as CI, 2 workers, stopped immediately after both workers were healthy. This measures startup/capture, not the full 1088-sample speed stage.
- Baseline: `sgl-project/main` @ `86e73bd`.
- PR: `2360121`.
- I discarded the first cold baseline run because it included one-time CUDA/kernel cache warmup (`Capture cuda graph end` around 22.8s). The table below uses warm-cache reruns `r2` and `r3` for both refs.

| Metric | Baseline avg | PR avg | Delta |
|---|---:|---:|---:|
| Startup wall time to router healthy | 28.53s | 29.04s | +0.50s (+1.8%) |
| CUDA graph capture time per worker | 2.63s | 2.71s | +0.08s (+2.9%) |

Both refs captured the same Higgs TTS graph buckets: `[1, 2, 4, 8, 12, 16, 24, 32, 40, 48, 56, 64]`, and both warm reruns reached 2/2 healthy workers. I do not see a material startup or capture-time regression here.

Run logs are on the H100 host under `/data/jiaxuanluo/pr850-startup/runs`.

sglang_omni/scheduling/bootstrap.py
```diff
@@ -82,3 +82,20 @@ def create_sglang_infrastructure(
...
         model_worker.model_config,
     )
+
+
+def create_sglang_infrastructure_defer_cuda_graph(
+    server_args: Any,
+    gpu_id: int,
+    **kwargs: Any,
+):
+    """Create infrastructure without initial graph capture; return whether to capture later."""
```
zhaochenyang20: The comment for this function should be rather detailed. Explain to the user why we should have a defer strategy for the CUDA Graph in need, and what kinds of things should be made a priority for the CUDA Graph, and what kinds of things should be covered by the CUDA Graph.

Use `#note (Jiaxuan):` for your comment.

sglang_omni/utils/cuda_graph_batch_validator.py
```diff
@@ -0,0 +1,298 @@
...
+    *,
+    stage: str,
+    max_running_requests: int | None,
+    cuda_graph_max_bs: int | None,
+    captured_bs: list[int] | None,
+    request_slots: int | None,
+    buffer_capacity: int | None,
+    buffer_source: str | None = None,
+) -> CudaGraphBatchReport:
+    """Compare the three batch-size facts and produce a verdict."""
```
zhaochenyang20: this comment is too vague.

sglang_omni/utils/cuda_graph_batch_validator.py
```diff
@@ -0,0 +1,298 @@
...
+    findings: list[str] = []
+    ok = True
```
zhaochenyang20: naming a variant `ok` is quite improfessional.

sglang_omni/utils/cuda_graph_batch_validator.py
zhaochenyang20: the same:

`max_running_requests=64`, `cuda_graph_max_bs=64`, `request_slots=64`, `buffer_capacity=64`, and `captured_bs=None` returns `ok=True`.

sglang_omni/models/qwen3_omni/stages.py
zhaochenyang20, line 1: the PR replaces Qwen3-ASR and Whisper ASR's explicit batch fields with `build_generation_batch_defaults(max_running_requests)`, but does not follow the TTS stages by calling `build_generation_batch_overrides()`. `build_generation_batch_defaults()` only returns `max_running_requests`, `cuda_graph_max_bs`, and `torch_compile_max_bs`; it does not generate `cuda_graph_bs`. I confirmed with the current `build_sglang_server_args()` that passing only those defaults leaves `server_args.cuda_graph_bs` as `None`. These ASR stages default `disable_cuda_graph=False` and then call `model_worker.model_runner.init_device_graphs()`; upstream SGLang `get_batch_sizes_to_capture()` calls `max(capture_bs)`, so `capture_bs=None` crashes during graph runner initialization. The failure mode is that ASR executor startup cannot complete CUDA graph capture, and Whisper ASR has the same issue at `sglang_omni/models/whisper_asr/stages.py:47`.

zhaochenyang20 (changes requested):
Add baseline/current startup data from the same H100 host to the PR description, at least splitting infrastructure construction, model-side setup, `init_device_graphs()`, and any stage-specific graph capture time.

sglang_omni/utils/cuda_graph_batch_validator.py
luojiaxuan: Done

sglang_omni/scheduling/bootstrap.py
luojiaxuan: Done

sglang_omni/utils/cuda_graph_batch_validator.py
luojiaxuan: Fixed: missing captured_bs now makes the report invalid when an effective target is known, and undercoverage is checked against min(cuda_graph_max_bs, max_running_requests, request_slots). Added regression tests for both cases.

sglang_omni/scheduling/bootstrap.py
luojiaxuan: Fixed: expanded the `# note (luojiaxuan):` comment and the function docstring to explain why CUDA graph capture is deferred and which steady-state decode shapes it should cover.

sglang_omni/utils/cuda_graph_batch_validator.py
luojiaxuan: Fixed: the function docstring now states the validation contract and explicitly calls out missing capture data as a mismatch when the serving target is known.

sglang_omni/utils/cuda_graph_batch_validator.py
luojiaxuan: Fixed: renamed the report field and local variable from `ok` to `is_valid`.

sglang_omni/models/qwen3_omni/stages.py
luojiaxuan: Fixed: removed `build_generation_batch_defaults()` and changed both ASR stages to use `build_generation_batch_overrides()`, which emits `cuda_graph_bs` unless explicitly overridden. Added Qwen3-ASR and Whisper ASR coverage for this.

tests/unit_test/qwen3_tts/test_pipeline.py
SandyLuXY: The PR now rebinds default `torch_compile_max_bs` to an overridden max_running_requests, so this test gets 2, but it still asserts the old fixed 16. If the expected behavior is 16, shall we pass `torch_compile_max_bs=16` from the Qwen3-TTS stage?

sglang_omni/utils/cuda_graph_batch_validator.py
```diff
@@ -0,0 +1,308 @@
...
+    except AttributeError:
+        max_running_requests = None
+        cuda_graph_max_bs = None
+
+    try:
+        request_slots = model_runner.req_to_token_pool.size
+    except AttributeError:
+        request_slots = None
+
+    captured_bs = read_captured_bs(model_runner)
```
SandyLuXY, line 307: `validate_stage()` never checks `server_args.disable_cuda_graph`, so a graph-disabled stage with no graph_runner.capture_bs is reported as `MISMATCH`. Repro on PR head `06b800c` returned `False` for `disable_cuda_graph=True`. This contradicts the intended audit behavior for disabled stages.

Disabled graphs should short-circuit to a valid no-graph/no-op report, or the evaluator needs an explicit `cuda_graph_enabled` input.

tests/unit_test/qwen3_tts/test_pipeline.py
zhaochenyang20: well. Thanks. I think we should put it up to 32? we use concurrency 32 for ASR in CI.

luojiaxuan: Fixed: Qwen3-TTS now pins both cuda_graph_max_bs and torch_compile_max_bs to 32, and the test expectation was updated.

sglang_omni/utils/cuda_graph_batch_validator.py
luojiaxuan: Fixed: graph-disabled stages now return a valid no-op audit report. Added a regression test for disable_cuda_graph=True with no graph_runner.

SandyLuXY (approved):

zhaochenyang20 (approved):

PR #853, opened 2026-06-22
chore(moss-tts): delete filler/anonymous comments

MelodyyyYin:
Delete filler and anonymous comments per #661 policy.

No logic changes — pure comment deletion.


MelodyyyYin:
Closing for now — per #661 planning, filler/anonymous-comment cleanup will be **folded into each model's refactor migration PR** ("delete what the refactor obsoletes") rather than run as a standalone pass, to avoid comment churn right before the code moves. Still tracked under #661.

PR #779, opened 2026-06-13
[Model] Add Zyphra ZONOS2 (MoE TTS) support

GaokaiZhang:
Adds native serving for **Zyphra ZONOS2** (Apache-2.0 MoE TTS — 9-codebook DAC @ 44.1 kHz, zero-shot voice cloning), closing #775. It runs through the standard multi-stage TTS pipeline instead of Zyphra's forked `minisgl` stack.

### Pipeline (`sglang_omni/models/zonos2/`)
`preprocessing` (NeMo TN¹ + UTF-8 byte tok + conditioning) → `speaker_encode` (Qwen3 voice embedding, content-hash cached) → `tts_engine` (MoE AR decode → 9 DAC codebooks, delay pattern) → `vocoder` (DAC 44.1 kHz). Registered as `Zonos2ForCausalLM` (resolved from `params.json`; the checkpoint ships no `config.json`); served **DP 1, no router**.

### Model
28-layer GQA-16/4 backbone; sonic **EDA MoE** (16 experts, top-1 default + layer-26 top-2, router-dim 128, dense layers 0–2/27); attention with QK-norm + learnable per-head temp + interleaved RoPE + headwise sigmoid gate; single `[9234, 2048]` multi-codebook head with `15·tanh(x/15)` soft-cap; loads the single-file `model.pth`.

### Correctness
Greedy decode matches the Zyphra reference 12/12 frames on a fixed prompt; the weight loader consumes all 507 checkpoint tensors with no leftovers. 18 CPU streaming-vocoder tests (`tests/unit_test/zonos2/test_streaming_vocoder.py`) lock in length alignment across emit sizes, the `on_stream_done` tail top-up, and the EOS / short / empty edge cases.

### Performance
Correctness-first batch-1 engine, then optimized into a real-time single-GPU server. All perf toggles are typed **factory/server args** on the pipeline stages (no env vars). The headline numbers use the opt-in flags `fp8 + frame_graph + async_decode + compile_sampler`, which are **off by default** — out-of-the-box serving is correctness-safe but slower.

**AR decode (`tts_engine`)**
- Per-step host↔device sync removal (bulk `keys.tolist()`; dropped the sampler's redundant `any()` guard).
- `torch.compile` on the MoE backbone.
- Opt-in **FP8** MoE experts (`fp8`; Hopper/Ada) — roughly halves the MoE-expert weight bytes read per step (bf16 attention/dense projections untouched).
[...]

PR #858, opened 2026-06-22
[Model] Add Ming-Omni-TTS 16.8B standalone TTS support

### LINKED ISSUE #256: [Feature] Ming-Omni 16.8B MoE Talker Supporting
### Checklist

### Motivation

The current talker is a 0.5B dense Qwen2 backbone that serves as a text encoder for the CFM/DiT audio generation pipeline. A 16B MoE talker would provide higher quality TTS with better expressiveness.

We will release a 16B MoE talker weight in Ming-flash-omni-2.0 model in huggingface to support this feature in advance.

### Related resources

_No response_

YzXiao101:
## Motivation

This PR adds standalone serving support for [`inclusionAI/Ming-omni-tts-16.8B-A3B`](https://huggingface.co/inclusionAI/Ming-omni-tts-16.8B-A3B).

This checkpoint is different from the existing Ming Omni talker path (dense 0.5B). It is a standalone TTS stack with a BailingMoE AR backbone, FlowLoss/CFM acoustic latent generation, reference-audio conditioning, and AudioVAE waveform decode.

The public checkpoint is now available, so this PR adds a native SGLang-Omni serving path.

### Scope

This PR adds:

- A new `sglang_omni.models.ming_tts` package and repo-level `examples/run_ming_omni_tts_server.py` launcher.
- A four-stage serving pipeline: `preprocessing -> reference_encode -> tts_engine -> audio_decode`.
- `/v1/audio/speech` support for default text-to-speech and one local reference audio with reference text.
- Prompt construction, tokenizer handling, typed cross-stage payloads, and strict checkpoint coverage for the Ming TTS checkpoint layout.
- A SGLang-backed BailingMoE AR engine for hidden-state generation, TP-aware backbone execution, CUDA graph replay, and graph-aligned tail graph replay.
- A Ming-specific acoustic tail for FlowLoss/CFM latent sampling, stop handling, latent feedback into the next AR step, and AudioVAE waveform decode.
- Explicit fail-fast handling for unsupported logits sampling fields, seed input, streaming audio, and prefix/radix cache.
- Unit tests for seed rejection, non-finite sampling-param rejection, empty-latent finish/audio decode, request/result adaptation, and reference-cache keying. CUDA graph and TP communication paths are covered by the manual GPU verification runs below, not by unit tests.

This PR intentionally leaves the following product features for a follow-up PR because they add separate API and behavior scope:

- Instruction/style/emotion/dialect controls.
- Multi-reference podcast-style generation.
- True streaming audio responses.
- Prefix/radix cache for Ming continuous acoustic state.
- A replacement for the existing Ming Omni assistant talker path.

### References

image

- Official model card: [`inclusionAI/Ming-omni-tts-16.8B-A3B`](https://huggingface.co/inclusionAI/Ming-omni-tts-16.8B-A3B)
- Official implementation: [`inclusionAI/Ming-omni-tts`](https://github.com/inclusionAI/Ming-omni-tts)
- Official architecture figure: [`ming_omni_tts.png`](https://github.com/inclusionAI/Ming-omni-tts/blob/main/figures/ming_omni_tts.png)
- Latest SGLang-Omni TTS integration reference: [`[Model] Add Zyphra ZONOS2 (MoE TTS) support
`](https://github.com/sgl-project/sglang-omni/pull/779)

### High-Level Design

ming_tts_serving_architecture

Ming TTS is not a normal logits-sampling TTS model. The AR backbone produces hidden states, and each decode step runs a model-specific tail: FlowLoss/CFM samples a continuous acoustic latent patch, the latent patch is projected back into the AR input space as feedback, and AudioVAE later decodes the generated latent sequence into waveform chunks.

That shape drives the implementation boundaries:

- `preprocessing` builds the prompt payload and rejects unsupported API features early.
- `reference_encode` is a no-op for text-only requests and extracts prompt latents plus speaker embedding for reference-audio requests.
- `tts_engine` owns the SGLang BailingMoE AR backbone, TP execution, AR CUDA graph, tail CUDA graph, and request-local latent/feedback state.
- `audio_decode` turns generated latent chunks into the final waveform with the shared Ming-Omni talker AudioVAE.

The feedback loop is the key distinction: each AR step emits a hidden state, the tail converts it into a latent patch and feedback embedding, and the next AR decode step consumes that feedback row. The entry rank owns the acoustic tail and final pipeline payload; follower ranks mirror only the token and feedback metadata needed to keep TP decode aligned.

### Feature Design

This PR focuses on the basic serving pipeline plus the smallest validated performance path. More advanced cache, streaming, and reference reuse features are split into follow-up PRs to keep this review tractable.

| feature | status | note |
| --- | --- | --- |
| SGLang AR CUDA graph | ✅ Implemented | Uses the SGLang graph runner for the BailingMoE AR backbone |
| Acoustic tail CUDA graph | ✅ Implemented | Captured on the entry rank and aligned with AR graph batch buckets |
| Tensor parallel AR backbone | ✅ Implemented | Applies to `tts_engine`; side stages stay single-rank |
| Single-rank side stages | ✅ Implemented | Covers preprocessing, reference encoding, and audio decode |
| Prefix/radix cache | ❌ Follow-up | Token-only cache keys are not enough for Ming's continuous latent state |
| Reference-audio content cache | ✅ Implemented | Shared `ReferenceEncodeService` (M4a): speaker embedding + prompt latent cached per reference file with full-content hash keys, single-flight dedup, LRU; enabled by default, opt out with `ref_audio_cache=False` |
| Streaming audio response | ❌ Follow-up | Streaming changes API behavior and state lifetime semantics |

### Code Design

The code shape follows the SGLang-Omni model-structured serving direction in https://github.com/sgl-project/sglang-omni/issues/661. It also reuses Ming-Omni talker components where the serving math and checkpoint protocol match, while keeping TTS-specific AR/tail feedback state inside `ming_tts`.

| file | key class | boundary and purpose |
| --- | --- | --- |
| `config.py` | `MingTTSPipelineConfig` | Defines the four-stage topology and placement. |
| `request_builders.py` | `build_ming_tts_payload` | Converts `/v1/audio/speech` into a Ming request payload. |
| `payload_types.py` | `MingTTSState` | Carries typed cross-stage state. |
| `reference_encode.py` | `MingTTSReferenceEncoder` | Handles optional reference-audio conditioning. |
| `engine_io.py` | `MingTTSSGLangRequestData` | Adapts Ming state to the SGLang AR runner boundary. |
| `model_runner.py` | `MingTTSModelRunner` | Owns AR execution, TP sync, feedback staging, and result assembly. |
| `sglang_model.py` | `MingTTSSGLangModel` | Owns BailingMoE AR execution and the Ming acoustic tail. |
| `audio_decode.py` | `MingAudioDecoder` | Decodes latent chunks with the shared Ming-Omni talker AudioVAE. |
| `engine_builder.py` | `MingTtsEngineBuilder` | Shared `TtsEngineBuilder` (M2) hooks own graph, TP, and backend startup policy at the stage boundary. |

## Related Issues

close https://github.com/sgl-project/sglang-omni/issues/256.

### Environment

| item | value |
| --- | --- |
| Model | `inclusionAI/Ming-omni-tts-16.8B-A3B` |
| ASR scorer | `Qwen/Qwen3-ASR-1.7B` |
| Dataset | `zhaochenyang20/seed-tts-eval-arrow` |
| Runtime image | `hongccc/sglang-omni:dev` |
| SGLang | `0.5.12.post1` |
| Torch | `2.11.0+cu130` |
| Headline hardware | 2x NVIDIA H100 80GB HBM3 |

### Official case oracle

| check | result |
| --- | --- |
| Official implementation replay | PASS: 11 active cookbook audio examples generated with the official repo |
| In-scope API replay | PASS: the reference-audio TTS cookbook example is reproduced through `/v1/audio/speech`, ASR WER 0.0 |
| Explicitly deferred product scope | 8 instruction / multi-reference / voice-design / speech-with-effect examples return expected unsupported-feature errors |
| Custom prompt scope | 1 music/BGM example is deferred because `/v1/audio/speech` does not expose Ming's raw `prompt` field |
| TTA caveat | record-only: the official cookbook comments that the TTA task should use `inclusionAI/Ming-omni-tta-0.5B` |

### Initial Single-GPU Correctness Reference

Note: **The SeedTTS reference/reference-audio slices below use the first 100 cases per slice. A full text-only SeedTTS audit is included in the final update.**

This is the first single-H100 bring-up correctness run, before the TP2/CUDA graph serving path was enabled. It is a historical sanity reference, not a same-commit A/B against the final candidate.

| slice | lang | requests | WER | n_above_50 |
| text_only | en | 100 | 0.00546 | 0 |
| text_only | zh | 100 | 0.00422 | 0 |
| reference | en | 100 | 0.00455 | 0 |
| reference | zh | 100 | 0.00738 | 0 |

### Current TP2 + AR Graph + Tail Graph Correctness

The current headline path revalidates the four 100-sample SeedTTS slices with TP2, AR CUDA graph, and tail CUDA graph enabled.

| slice | lang | requests | failed | WER | n_above_50 |
| text_only | en | 100 | 0 | 0.006369 | 0 |
| text_only | zh | 100 | 0 | 0.003163 | 0 |
| reference | en | 100 | 0 | 0.006369 | 0 |
| reference | zh | 100 | 0 | 0.008434 | 0 |

The final text-only full SeedTTS audit uses the same serving path.

| slice | lang | requests | failed | WER | n_above_50 |
| text_only | en | 1088 | 0 | 0.010718 | 0 |
| text_only | zh | 2020 | 0 | 0.008565 | 1 |

The full-ZH audit has one `n_above_50` outlier out of 2020 requests. Manual audio review confirmed the generated speech matches the target text; the extra ASR tail is an ASR hallucination and is not audible in the generated audio.

### Current Headline Benchmark: TP2 + AR Graph + Tail Graph

Configuration:

- 2x H100 80GB HBM3
- TP2 on `tts_engine`
- `GEN_CONCURRENCY=8`
- `MAX_RUNNING_REQUESTS=8`
- CUDA graph buckets: `1,2,4,8`
- torch.compile disabled
- prefix/radix cache disabled

SeedTTS 100-sample check, `warmup=8`:

| slice | lang | concurrency | samples | failed | rtf_mean | latency_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| text_only | en | 8 | 100 | 0 | 0.4705 | 2.028 | 3.833 | 17.086 | 0.006369 | 0 |
| text_only | zh | 8 | 100 | 0 | 0.4565 | 2.008 | 3.848 | 17.167 | 0.003163 | 0 |
| reference | en | 8 | 100 | 0 | 0.5629 | 2.223 | 3.517 | 14.731 | 0.006369 | 0 |
| reference | zh | 8 | 100 | 0 | 0.5110 | 2.909 | 2.681 | 15.530 | 0.008434 | 0 |

Full SeedTTS text-only audit, `warmup=0`:

| slice | lang | concurrency | samples | failed | rtf_mean | latency_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| text_only | en | 8 | 1088 | 0 | 0.4902 | 2.228 | 3.581 | 16.943 | 0.010718 | 0 |
| text_only | zh | 8 | 2020 | 0 | 0.4658 | 2.285 | 3.495 | 17.520 | 0.008565 | 1 |

### Initial Single-GPU Baseline Reference

This is a historical single-H100 reference, not a clean same-commit A/B against the current TP2 run.

| slice | lang | requests | rtf_mean | latency_mean_s | throughput_qps | audio_s/s | WER |
| text_only | en | 100 | 4.2518 | 18.613 | 0.421 | 1.881 | 0.00637 |
| text_only | zh | 100 | 3.7950 | 17.062 | 0.461 | 2.075 | 0.00422 |
| reference | en | 100 | 3.8372 | 15.898 | 0.492 | 2.040 | 0.00455 |
| reference | zh | 100 | 3.7874 | 22.001 | 0.358 | 2.077 | 0.00527 |

## Follow-ups

- Add a dedicated Ming TTS CI cookbook target.
- Leave advanced product features and optimization work to follow-up PRs, starting with https://github.com/sgl-project/sglang-omni/pull/1005.

## Checklist

- [x] CUDA graph and TP contracts verified manually on GPU (see Modifications; no unit tests for these paths).
- [x] Update user-facing documentation.
- [x] Provide accuracy results.
- [x] Provide benchmark results.


sglang_omni/models/ming_tts/sglang_request_builders.py
```diff
@@ -0,0 +1,467 @@
...
+                )
+            if normalized is not None and not (
+                normalized in ("stop", "matched", "eos", "finish_matched_token")
+                or "matched" in normalized
+                or "eos" in normalized
+            ):
+                raise RuntimeError(
+                    "Ming-Omni-TTS stop-head recorded stop_step="
+                    f"{ar_state.stop_step}, but SGLang finished_reason was {raw!r}"
+                )
```
edwingao28: **[Correctness] Final-step stop can collide with SGLang's length finish**

I think this can reject a valid generation at the decode budget boundary. If the stop head fires on `step == max_decode_steps - 1`, we record `ar_state.stop_step`, but SGLang can still report `finished_reason="length"` because `max_new_tokens` is exhausted at the same time. This branch then raises even though the model produced a normal final chunk.

Could we treat `"length"` as acceptable when `stop_step` is present, serialize this as `finish_reason="stop"`, and add a regression test where both `ar_state.stop_step` and `finished_reason="length"` are set?

edwingao28:
Thanks for pushing this, this is a big serving integration. The SGLang specific pieces look pretty solid: the TP entry-rank/follower split, the scheduler hook integration, and the fixed feedback rows for CUDA graph replay are all nontrivial.

However, there are a lot of code could be reused from existing ming talker. I left one correctness comment that I think is worth fixing before merge. The rest is more of a maintainability heads-up while the PR is still WIP, not a request to block this on a large refactor.

One thing worth keeping in mind as the structure settles: a fair amount of the acoustic/backbone stack overlaps with code we already have under `sglang_omni/models/ming_omni/`. I don't think we should directly import most of it as-is - the upstream paths, signatures, checkpoint assumptions, and a few behavior details have diverged - but a few pieces look like good candidates to lift into shared modules either here or in a follow-up.

Suggest for sharing:

- `SinusPositionEmbedding` / `TimestepEmbedder` / `CondEmbedder`
- `RMSNorm` / `FeedForward` / `FinalLayer`
- the EPSS timestep table + sway / SDE-step sampling helpers
- parts of `audio_vae/*`, with the `_attn_implementation="eager"` fix folded into the shared copy

Pieces suggest avoid sharing directly:

- the CFM class itself, since the call contract differs
- `MingARTailExecutor`, since this path correctly lets the SGLang model runner own graph capture and keeps the tail eager
- the BailingMoe backbone for now, since the routing variants differ enough that it probably needs a larger hook-based refactor

A reasonable path might be: 
1. keep this PR focused on getting the WIP serving path correct and stable 
2. follow up with shared-module cleanup for the pieces that are clearly reusable. 
The exact ordering does not need to be strict, but I mainly want to avoid the two Ming stacks drifting apart over time.

CC: @yuan-luo

sglang_omni/models/ming_tts/sglang_request_builders.py
YzXiao101: 🙏 Thanks! Yes, it is a bug, and I have fixed it and also added regression coverage for both the valid boundary case and the invalid early `stop_step +  length` case on commit https://github.com/sgl-project/sglang-omni/pull/858/commits/de5d63f8b818bf12a61ad4e30eea8b3a71afe406.

YzXiao101:
> However, there are a lot of code could be reused from existing ming talker. I left one correctness comment that I think is worth fixing before merge. The rest is more of a maintainability heads-up while the PR is still WIP, not a request to block this on a large refactor.

Yes, I agree with this suggestion. The current approach of importing fm and audio_vae directly from the upstream repo is not sustainable. During development I also found that many components should be shared with the existing ming_omni talker stack. 

Before the next round of performance work, I will first draft a small reuse plan for the Ming-family modules so ming_tts and ming_omni do not drift apart — that won't take long～

YzXiao101:
## Update: TP2 AR Graph + Tail Graph

Updates:

- Reuse or align FlowLoss/CFM, DiT/aggregator, and AudioVAE pieces where the serving math matches the `ming_omni` talker path.
- Move the Ming-specific AR tail into a graph-compatible decode path, covering FlowLoss sampling, latent feedback, stop handling, and request-local latent state.
- Capture tail CUDA graphs for exact active batch sizes when AR CUDA graph is enabled.

Benchmark:

`TP2 + AR Graph` $\rightarrow$ `TP2 + AR Graph + Tail Graph`

**avg `rtf_mean`: `3.5353` -> `0.5390` (~6.6x).**  
**avg `throughput`: `0.478` -> `3.200` (~6.7x).**


| slice | lang | RTF lift | QPS lift |
| --- | --- | ---: | ---: |
| text_only | en | **6.3x** | **6.3x** |
| text_only | zh | **7.3x** | **7.4x** |
| reference | en | **6.1x** | **6.3x** |
| reference | zh | **6.8x** | **6.8x** |

Accuracy:

| slice | lang | WER | rtf_mean | latency_s | qps | audio_s/s |
| text_only | en | 0.00546 | 0.5798 | 2.526 | 3.083 | 14.195 |
| text_only | zh | 0.00474 | 0.4742 | 2.086 | 3.713 | 16.621 |
| reference | en | 0.00637 | 0.5864 | 2.324 | 3.362 | 13.954 |
| reference | zh | 0.00474 | 0.5154 | 2.948 | 2.643 | 15.353 |

WIP, subsequent efforts will focus on advancing performance optimizations and applying code style cleanup prior to final review.

YzXiao101:
## Update: Streamlined Ming-TTS Serving Refactor

Update:

- Refactored the Ming-TTS serving path around the current SGLang-Omni multi-stage model structure.
- Reworked the AR/tail/runtime boundaries so TP2, CUDA graph, tail graph, request state, and payload conversion are owned by clearer Ming-TTS-local modules.
- Reused and aligned more of the existing Ming-Omni talker implementation where the serving math matches, while keeping TTS-specific AR/tail protocol isolated.
- Rebased on latest `main` and aligned the serving structure as much as possible with the recommended pattern in https://github.com/sgl-project/sglang-omni/issues/661.
- Removed several patch-style compatibility paths and tightened the code shape for reviewability.

Benchmark:

Setup: `TP2 + AR graph + tail graph`, 2x H100 80GB HBM3, `c=8`, 100 SeedTTS samples per slice.

| slice | lang | requests | failed | rtf_mean | latency_mean_s | qps | audio_s/s |
| text_only | en | 100 | 0 | 0.7055 | 2.945 | 2.664 | 11.772 |
| text_only | zh | 100 | 0 | 0.4539 | 2.002 | 3.866 | 17.321 |
| reference | en | 100 | 0 | 0.5791 | 2.313 | 3.385 | 14.253 |
| reference | zh | 100 | 0 | 0.5065 | 2.846 | 2.737 | 15.662 |

Average `rtf_mean`: `0.5613`; average throughput: `3.163` req/s.

Accuracy:

| slice | lang | WER | n_above_50 |
| --- | --- | ---: | ---: |
| text_only | en | 0.00546 | 0 |
| text_only | zh | 0.00369 | 0 |
| reference | en | 0.00819 | 0 |
| reference | zh | 0.00211 | 0 |

Note:

- This run used `warmup=0`; the first `text_only/en` concurrency batch carried startup/lazy-init cost, so its p95 latency is conservative.

sglang_omni/models/ming_tts/engine_io.py
```diff
@@ -0,0 +1,370 @@
...
+                prompt_tokens = 0
+            else:
+                shape = getattr(prompt_input_ids, "shape", None)
+                prompt_tokens = (
+                    int(shape[0])
+                    if shape is not None and len(shape)
+                    else len(prompt_input_ids)
+                )
+
+            generated = torch.stack(latent_chunks, dim=0)
```
edwingao28: Can we handle the no-latent terminal path before this `stack`? 

If the request aborts/errors before the first AR chunk, `latent_chunks` is empty and `torch.stack([])` raises a generic internal error, which would hide the original finish reason.

Maybe return a terminal state without `generated_latents` when there are no chunks and the finish reason is `abort`/`error` (or otherwise make the expected invariant explicit).

sglang_omni/models/ming_tts/request_builders.py
```diff
@@ -0,0 +1,366 @@
...
+    if temperature_value is None and params.get("ming_temperature") is not None:
+        temperature_value = params["ming_temperature"]
+    if (
+        temperature_value is None
+        and "temperature" in explicit_fields
+        and params.get("temperature") is not None
+    ):
+        temperature_value = params["temperature"]
+
+    seed_value = tts_params.get("seed", params.get("seed"))
```
edwingao28: Could we reject `seed` for now? It gets parsed and stored, but I don't see it reaching the FlowLoss sampling path yet. Since most other unsupported knobs fail fast, accepting this one feels a bit misleading: callers may expect reproducible output from a seed that currently does not affect generation.

edwingao28:
@YzXiao101  Thanks for the cleanup here. The current diff doesn't include the Ming-TTS tests described in the PR body.

`tests/unit_test/ming_tts/` only adds `__init__.py`, but the PR body says CUDA graph / TP contract unit tests are included, and the previous stop/length fix mentioned regrewssion coverage. Also, the body still describes projected-prefill radix cache as implemented/validated, while the current code force-disables radix cache and raises if it is enabled.

Could you add the missing tests and verify those behavoir aligned with pr body description?

YzXiao101:
## Update: Final Candidate Cleanup and SeedTTS Audit

Update:

- Flattened thin Ming-TTS adapter packages and simplified the BailingMoE/TTS model code, weight coverage reporting, request state, payload transport, and pipeline construction while preserving the TP + CUDA graph serving path.
- Aligned tail CUDA graph buckets with SGLang AR CUDA graph buckets and kept tail graph ownership on the entry rank.
- Added `examples/run_ming_omni_tts_server.py` as the repo-level Ming-TTS server launcher.
- Removed Ming-specific request profiling hooks and kept profiling as follow-up work instead of submit-path code.
- Moved prefix/radix cache support to a separate follow-up branch and reject it in this PR to keep the first serving version readable.
- Added PR-review contract fixes for unsupported seed input, empty-latent finish handling, and the related unit tests.

Benchmark:

Setup: 2x H100 80GB HBM3, `c=8`, TP2, AR CUDA graph, tail CUDA graph, radix cache disabled.

SeedTTS 100-sample check, `warmup=8`:

| slice | lang | concurrency | samples | failed | rtf_mean | latency_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| text_only | en | 8 | 100 | 0 | 0.4705 | 2.028 | 3.833 | 17.086 | 0.006369 | 0 |
| text_only | zh | 8 | 100 | 0 | 0.4565 | 2.008 | 3.848 | 17.167 | 0.003163 | 0 |
| reference | en | 8 | 100 | 0 | 0.5629 | 2.223 | 3.517 | 14.731 | 0.006369 | 0 |
| reference | zh | 8 | 100 | 0 | 0.5110 | 2.909 | 2.681 | 15.530 | 0.008434 | 0 |

Full SeedTTS text-only audit, `warmup=0`:

| slice | lang | concurrency | samples | failed | rtf_mean | latency_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| text_only | en | 8 | 1088 | 0 | 0.4902 | 2.228 | 3.581 | 16.943 | 0.010718 | 0 |
| text_only | zh | 8 | 2020 | 0 | 0.4658 | 2.285 | 3.495 | 17.520 | 0.008565 | 1 |

Note:

- All 100-sample subset checks passed with zero failed requests and zero `n_above_50`.
- Full EN passed with zero failed requests and zero `n_above_50`.
- Full ZH had one `n_above_50` outlier out of 2020 requests. Manual audio review confirmed the generated speech matches the target text; the extra ASR tail is an ASR hallucination and is not audible in the generated audio.
- The full-ZH corpus WER remains below 1%.

sglang_omni/models/ming_tts/engine_io.py
YzXiao101: 🙏 Thanks! Yes, it is a bug, and I have fixed it and added a UT on commit 8fc444f1.

sglang_omni/models/ming_tts/request_builders.py
YzXiao101: 🙏 Thanks! I have fixed it and added a UT on commit 8fc444f1.

sglang_omni/models/ming_tts/sglang_model.py
```diff
@@ -0,0 +1,1145 @@
...
+    def _make_tail_sampling_inputs(
+        self,
+        *,
+        batch_size: int,
+        device: torch.device,
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        noise_rows: list[torch.Tensor] = []
+        sde_rows: list[torch.Tensor] = []
+        timesteps = None
+        for _ in range(int(batch_size)):
```
edwingao28: Hot-path perf: `_make_tail_sampling_inputs()` loops over `bs` and calls `torch.randn` + `build_cfm_sampling_schedule(..., batch_size=1)` per row. The schedule helper already accepts `batch_size`, so could we generate `noise` as `(bs, latent_dim, patch_size)` and call `build_cfm_sampling_schedule(..., batch_size=bs)` once? That keeps per-request randomness while avoiding `O(bs)` tiny RNG launches each tail step.

sglang_omni/models/ming_tts/model_runner.py
```diff
@@ -0,0 +1,450 @@
...
+                    temperature=temperature_tensor,
+                )
+            )
+            sampled = tail_outputs.sampled
+            stop_prob = tail_outputs.stop_prob
+            feedback_embeddings = tail_outputs.feedback_embeddings
+            stop_flags = (stop_prob > 0.5) & (steps_tensor > 3)
+            length_flags = steps_tensor + 1 >= max_steps_tensor
+
+            stop_list = [bool(value) for value in stop_flags.detach().cpu().tolist()]
```
edwingao28: Hot-path sync: this does two D2H syncs per decode step (`stop_flags` and `length_flags`), and the follower path below does per-row `.item()` syncs for `stop_flags` / `feedback_mask`. 

Please batch those host reads. The host loop still needs the booleans, but it shouldn't need separate syncs for each flag/read.

sglang_omni/models/ming_tts/prompt_builder.py
```diff
@@ -0,0 +1,125 @@
...
+        input_ids.extend([tokenizer.special.audio_patch] * prompt_latent_token_count)
+
+    if input_ids[0] != tokenizer.special.role_start:
+        raise ValueError("Ming-Omni-TTS MoE prompt must start with role tokens")
+
+    return MingTTSPromptPlan(
+        input_ids=input_ids,
+        effective_prompt=effective_prompt,
+        prompt_tokens=len(input_ids),
+        audio_token_position=audio_token_position,
```
edwingao28, line 120: Could we add a small test for this position contract? These indices are later used to overwrite speaker rows and prompt-latent rows in `_build_prefill_input_embeds`; an off-by-one here would likely condition the wrong token rather than fail loudly.

Add a fake-tokenizer test that asserts the audio anchor, speaker injection pos, and prompt-latent span.

sglang_omni/models/ming_tts/sglang_model.py
YzXiao101: Thank you for your continued careful review. I have fixed this issue in commit https://github.com/sgl-project/sglang-omni/pull/858/changes/2cdc759b6e29547c6c1e753007adce50c77dc644

sglang_omni/models/ming_tts/model_runner.py
YzXiao101: Thank you for your continued careful review. I have fixed this issue in commit https://github.com/sgl-project/sglang-omni/pull/858/changes/2cdc759b6e29547c6c1e753007adce50c77dc644

sglang_omni/models/ming_tts/prompt_builder.py
YzXiao101: Thank you for your continued careful review. I have fixed this issue in commit https://github.com/sgl-project/sglang-omni/pull/858/changes/2cdc759b6e29547c6c1e753007adce50c77dc644

YzXiao101:
## Update: Latest Full SeedTTS Validation

Update:

- Completed the seed, finish, and empty-output contracts: unsupported request-local seeds fail explicitly, zero-latent finishes serialize correctly, and AudioVAE skips empty decode input.
- Reduced TP step synchronization to one control tensor plus feedback embeddings and one host decision snapshot per rank.
- Generated FlowLoss initial noise and SDE noise batch-wise while preserving the same sampling shapes and CUDA graph inputs.
- Added focused tests for prompt embedding positions, finish-reason precedence, empty latent output, and empty AudioVAE decode, and documented the new test coverage.

Benchmark:

Setup: 2x H100 80GB HBM3, `c=8`, TP2, AR CUDA graph, tail CUDA graph, radix cache disabled, `warmup=0`.

| slice | lang | requests | failed | rtf_mean | latency_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| text_only | en | 1088 | 0 | 0.5052 | 2.287 | 3.489 | 16.480 | 0.010550 | 0 |
| text_only | zh | 2020 | 0 | 0.4802 | 2.347 | 3.404 | 17.038 | 0.007216 | 0 |
| reference | en | 1088 | 0 | 0.5707 | 2.443 | 3.268 | 14.728 | 0.013564 | 3 |
| reference | zh | 2020 | 0 | 0.5298 | 2.970 | 2.690 | 15.399 | 0.009204 | 2 |

Result:

- All 6216 full-set requests completed successfully.
- The latest changes show no material accuracy or performance regression.
- These results will be included in the Ming-TTS cookbook currently in progress.

luojiaxuan:
1. **Merge main** — resolves the README/pyproject drift (README moved to the cookbook-link format; Ming-Omni-TTS added to the Speech generation bullet; the `zhon` dep you added is kept).
2. **[TTS Refactor] port** (`654a6c2`) — reworks the serving stack onto the shared surfaces from #985, keeping all model math (prompt build, FlowLoss/CFM tail, AudioVAE decode, weight loading) untouched:
   - **M2**: `stages.py`'s hand-rolled server-args/infra/scheduler startup is now `MingTtsEngineBuilder` on the shared `TtsEngineBuilder` (`engine_builder.py`); the TP guard rails, radix/chunked-prefill/torch-compile fail-fasts, and tail-graph init live in builder hooks. −355 lines in stages.py.
   - **T5a**: `audio_decode` now provides `MingTTSBatchVocoder` on `BatchVocoderBase`; `decode_ming_tts_audio_payload` delegates to the same implementation, so behavior (timing field, latent cleanup, waveform payload, usage) is unchanged.
   - **M4a**: `reference_encode` routes through the shared `ReferenceEncodeService` with a Ming hook — speaker embedding + prompt latent cached per reference file (content-sampled hash key, single-flight, LRU). Prompt build stays per-request. Opt out with `ref_audio_cache=False` factory arg.
   - **M8**: `ming_tts` exports `CAPABILITIES`, the config declares `requires_model_capabilities`, and the capabilities contract test covers `BailingMMNativeForConditionalGeneration`.
3. Two staleness fixes needed after the merge: `INITIAL_CODEC_CHUNK_FRAMES_PARAM` moved to `scheduling.streaming_vocoder` on main, and `_decode_float_tensor` now guards empty latent buffers (`torch.frombuffer` rejects empty payloads on torch ≥ 2.12; finish-before-first-chunk is a supported path with a unit test).

Deliberately **not** folded in: the radix-cache prefill keying and TP step-update rework from your #1005 draft — that's your in-flight work and it layers cleanly on top of this structure (`make_ming_tts_scheduler_adapters` gains kwargs, `engine_io.py` changes are orthogonal to the builder).

Local verification: `tests/unit_test/ming_tts/` (audio decode + request builders) and the capabilities/scheduling contract suites pass; `test_engine_io.py` needs real sglang and is covered by PR CI.

sglang_omni/models/ming_tts/reference_encode.py
```diff
@@ -0,0 +1,307 @@
...
+    def __init__(self, encoder: "MingTTSReferenceEncoder", *, model_identity: str):
+        self._encoder = encoder
+        self._model_identity = str(model_identity)
+
+    def normalize_input(self, raw_input: Any) -> str:
+        return str(raw_input)
+
+    def cache_key(self, item: str) -> ReferenceEncodeKey | None:
+        try:
+            digest = hash_file_sampled(item)
```
SandyLuXY: `hash_file_sampled()` only covers the first/last 8 KiB and file size, so two same-size WAVs with different audio in the middle can share a cache entry and the second request reuses the first speaker embedding/prompt latent. Please use `reference_path_cache_key(..., trust_stat=False)`, revalidate after encoding, and add a same-size WAV regression test.

sglang_omni/models/ming_tts/request_builders.py
```diff
@@ -0,0 +1,365 @@
...
+        except (TypeError, ValueError) as exc:
+            raise ValueError(f"Ming-Omni-TTS {name} must be an integer") from exc
+
+    def resolve_float(name: str, value: Any, default: float) -> float:
+        if value is None:
+            return float(default)
+        if isinstance(value, bool):
+            raise ValueError(f"Ming-Omni-TTS {name} must be a number")
+        try:
+            return float(value)
```
SandyLuXY: `resolve_float()` accepts `NaN`/`Inf`, and the checks below do not reject `NaN`, so non-finite values reach CFM and produce invalid latents instead of a request error. Please require `math.isfinite()` for `cfg`, `sigma`, and `temperature`, with `NaN`/`Inf` tests.

SandyLuXY (dismissed):
Great PR! Left some comments.
Also please update the PR body: it claims CUDA graph and TP communication unit tests, but none of the Ming tests cover those paths. It also marks the reference cache as follow-up even though it is enabled by default, and still refers to `MingTTSEngineStartup`. Please add the claimed tests or label them as manual coverage, then update the feature and code-design sections.

sglang_omni/models/ming_tts/reference_encode.py
luojiaxuan: Fixed in 6294a05 — the hook now keys with `reference_path_cache_key(..., trust_stat=False)` (full-content memoized hash, matching the fish/moss-local hook convention) and implements `revalidate` by recomputing the key. Added `test_reference_encode.py` with the same-size-WAV regression (two files differing only in the middle bytes must key separately), a plain-hit test, and a rewritten-file staleness test.

sglang_omni/models/ming_tts/request_builders.py
luojiaxuan: Fixed in 6294a05 — `resolve_float` now rejects non-finite values after conversion (`math.isfinite`), which covers cfg, sigma, and temperature in one place. Added parametrized NaN/+Inf/-Inf tests for all three. You're right that NaN passed every range check below (each comparison evaluates False).

luojiaxuan:
PR body updated per the review: the unit-test list no longer claims CUDA graph / TP communication tests (those paths are labeled as manual GPU verification), the reference-audio cache row is now ✅ implemented-by-default with the opt-out documented, and the code-design table points at `MingTtsEngineBuilder` in `engine_builder.py` instead of the deleted `MingTTSEngineStartup`.

sglang_omni/models/ming_tts/reference_encode.py
```diff
@@ -0,0 +1,314 @@
...
+        input_name = self.session.get_inputs()[0].name
+        embedding = self.session.run(None, {input_name: feat.unsqueeze(0).numpy()})[0]
+        return torch.as_tensor(embedding.reshape(1, -1), dtype=torch.float32)
+
+
+class _MingTTSReferenceEncodeHook(ReferenceEncodeHook[str, dict, dict]):
+    """M4a hook: cache (speaker embedding, prompt latent) per reference file.
+
+    The artifact is the text-independent conditioning bundle; prompt build
+    stays per-request in encode_payload. Keys are content-sampled file hashes
```
SandyLuXY: nit: `content-sampled` is stale now that this uses `reference_path_cache_key()`. This should say `full-content`.

luojiaxuan: Fixed in a341c80 — docstring now says full-content.

sglang_omni/models/ming_tts/model_runner.py
```diff
@@ -0,0 +1,448 @@
...
+        del forward_batch
+        if not requests:
+            return
+
+        hidden = result.logits_output.hidden_states
+        if hidden.ndim == 2:
+            hidden = hidden.unsqueeze(1)
+        hidden_states = hidden
+
+        if self._is_entry_rank:
```
edwingao28, line 353: Rank 0 can fail in `_run_entry_tail_step()` before this broadcast, while follower ranks still enter the collective and wait until the process-group timeout. Could we propagate the failure across ranks so a tail error doesn't hang the TP worker?

sglang_omni/models/ming_tts/engine_io.py
```diff
@@ -0,0 +1,377 @@
...
+            state.prompt_tokens = prompt_tokens
+            state.completion_tokens = len(latent_chunks)
+            state.engine_time_s = time.perf_counter() - decode_state.engine_start_s
+
+            for field_name, value in encode_generated_latents(generated).items():
+                setattr(state, field_name, value)
+
+            return store_ming_tts_state(payload, state)
+        finally:
+            data.release_tensors()
```
edwingao28: release_tensors() only runs from result_adapter, so a request aborted before producing a result skips this cleanup. Since the Req and request data reference each other and hold CUDA tensors, those allocations can remain until cyclic GC. 

I suggest we could add an abort callback that releases the request state, as Higgs, Moss, and Qwen3 TTS do.

edwingao28 (commented):
I pushed a small follow-up commit that tightens reference-input validation, fixes CUDA graph batch-size validation and cache byte accounting, and reuses the CFM schedule.

Also left two correctness & lifecycle comments worth addressing before merge. One non-blocking follow-up: FlowLoss and stop_head are only used on rank 0 but are still loaded on follower ranks. Avoiding those copies may in theory recover some follower memory, but it likely requires rank-aware weight loading and coverage changes. Given the size of this PR, we could leave it as a follow, but I'll defer to you on scope @YzXiao101.

YzXiao101:
## Update: Request State Ownership and TP Failure Propagation

Update:

- Moved recurrent per-request GPU state from scheduler request data into the model runner and materialized it only after scheduler admission.
- Kept scheduler request data CPU-owned, with the entry rank publishing the terminal latent snapshot before result adaptation.
- Wired the same idempotent request cleanup into normal completion, result-adapter failure, and abort paths.
- Propagated entry-rank acoustic-tail failures through the existing TP step update before re-raising, so follower ranks terminate the same batch.
- Added focused contract tests for entry/follower tail failure handling and terminal/abort cleanup.

Benchmark:

Setup: 2x H100 80GB HBM3, `c=8`, TP2, AR CUDA graph, tail CUDA graph, radix cache disabled

| slice | lang | requests | failed | rtf_mean | latency_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| text_only | en | 100 | 0 | 0.5121 | 2.198 | 3.545 | 15.970 | 0.004550 | 0 |
| text_only | zh | 100 | 0 | 0.4775 | 2.112 | 3.666 | 16.540 | 0.004217 | 0 |
| reference | en | 100 | 0 | 0.5657 | 2.247 | 3.476 | 14.694 | 0.005460 | 0 |
| reference | zh | 100 | 0 | 0.5034 | 2.847 | 2.736 | 15.753 | 0.003690 | 0 |

sglang_omni/models/ming_tts/model_runner.py
YzXiao101: Thanks for catching this！Fixed in [098f9b12](https://github.com/sgl-project/sglang-omni/pull/858/commits/098f9b1201f1329e907e8785a2043b49bd4bb45e)

sglang_omni/models/ming_tts/engine_io.py
YzXiao101: Aggreed with this proposal and I have refactored and fixed this issue in [098f9b12](https://github.com/sgl-project/sglang-omni/pull/858/commits/098f9b1201f1329e907e8785a2043b49bd4bb45e)

sglang_omni/models/ming_tts/model_runner.py
YzXiao101: Yes, It is not necessary for followers to load tail weight, but given the scope of this PR, I prefer to implement this optimization in a follow-up PR https://github.com/sgl-project/sglang-omni/pull/1005

edwingao28:
Since 098f9b12 also moves the request local GPU state into the model runner, could you confirm the TP2 + AR/tail graph smoke tests were rerun on this head for both text only and reference audio requests? After that, LGTM from my side. @YzXiao101

sglang_omni/models/ming_tts/model_runner.py
edwingao28: sounds good！

YzXiao101:
@edwingao28 Yes, I have rerun the 100-sample seed tts tests on 2×H100 for both text-only and reference-audio requests. See: https://github.com/sgl-project/sglang-omni/pull/858#issuecomment-4971699429. Since splitting the request-local state does not change the serving flow or the model’s match computation, the results remained consistent as expected. Thanks for your reminder.

Hi @luojiaxuan , could you please take another look at this PR? Thanks!

edwingao28 (approved):

luojiaxuan:
I opened #1070 to consolidate the existing seven launchers into model-local presets, with 194 net-deleted non-test lines after preserving the compatibility CLI help and examples. Since Ming TTS is not on main yet, I kept this PR independent. After #1070 lands, please replace examples/run_ming_omni_tts_server.py with a ming-tts-server preset instead of merging the 393-line standalone launcher.

YzXiao101:
@luojiaxuan Hi jiaxuan, since Ming-TTS is a new integration with no backward-compatibility burden, I removed the standalone example launcher directly. It now follows the other TTS models and keeps sgl-omni serve --config as the single serving entry, avoiding multiple sources of truth and reducing code duplication.

Could you please take a look at this PR? Thanks!

YzXiao101:
## Update: Single Ming-TTS Serving Entry

Update:

- Removed the standalone Ming-TTS example launcher.
- Kept `sgl-omni serve --config examples/configs/ming_omni_tts.yaml` as the single serving entry, avoiding duplicate launcher and pipeline configuration logic.

Benchmark:

Commit: `c2c93118`. Setup: 2x H200, TP2, concurrency 8, AR CUDA graph and tail CUDA graph enabled, 100 SeedTTS samples per slice, `warmup=8`.

| slice | lang | failed | rtf_mean | throughput_qps | WER | n_above_50 |
| text_only | en | 0 | 0.4574 | 3.893 | 0.005460 | 0 |
| text_only | zh | 0 | 0.4497 | 3.905 | 0.005271 | 0 |
| reference | en | 0 | 0.5038 | 3.896 | 0.006369 | 0 |
| reference | zh | 0 | 0.4641 | 2.950 | 0.004744 | 0 |

Result:

- All 400 requests completed successfully with no accuracy outliers.

luojiaxuan (approved):

PR #859, opened 2026-06-22
[fix] MOSS-TTS Full-Row Radix Cache Keying for Generated Audio

sjmshsh:
## Problem Fixed

MOSS-TTS decode produces multi-channel rows, but SGLang's request state and
radix cache operate on a one-dimensional token-id sequence. Using only
`rows[:, 0]` made different generated audio frames look identical whenever they
shared the same text/control token.

That is risky for radix cache reuse because a later retraction/refill can match
into the generated region. If the generated region is keyed only by the control
channel, the cache can treat distinct audio rows as the same prefix.

The practical issue is not sampling itself. Sampling already produced the full
MOSS-TTS row. The issue is the scheduler-visible key used for:

- appending `Req.output_ids`,
- extending the request's KV chain,
- matching prefixes in the radix tree,
- preserving correct stop detection.

### Shared GPU Row Hash

The branch adds a shared hash utility:

```text
sglang_omni/utils/radix_hash.py
```

It exposes:

- `poly_row_hash(rows)`
- `gpu_radix_row_hash(rows, next_text, end_id)`

The hash is implemented with tensor-native int64 torch operations. It does not
move generated rows to CPU, so it avoids a per-frame GPU-to-CPU synchronization
on the decode hot path.

Continuing rows are folded below the special-token band:

```text
[0, 151643)
```

The model-specific end token is preserved as its raw id so existing SGLang stop
logic continues to work.

### MOSS-TTS Runner

MOSS-TTS now hashes the full generated row when publishing `next_token_ids`.
Instead of:

```python
next_token_ids = rows[:, 0].contiguous()
```

the runner now does:

```python
next_text = rows[:, 0]
next_token_ids = self._row_radix_token_ids(
    rows, next_text, int(self.model.config.im_end_token_id)
)
```

This means two generated rows with the same text/control token but different
audio codes produce different radix-cache keys.

### MOSS-TTS Local Alignment

MOSS-TTS Local now imports the same shared hash helper from
`sglang_omni.utils.radix_hash`.

This removes duplicated model-local hash code and keeps MOSS-TTS and
MOSS-TTS Local aligned on the same generated-row key contract.

## Behavioral Guarantees

The branch preserves these important properties:

- Full generated audio rows participate in radix keying.
- The generated-row hash stays on device and is decode-hot-path friendly.
- `im_end_token_id` for MOSS-TTS remains visible to SGLang stop detection.
- `audio_end_token_id` in ordinary generated rows is hashed as row content
  rather than treated as the final request stop token.
- MOSS-TTS Local continues to preserve its own raw end token for stop handling.


sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -100,6 +113,16 @@ def _build_prefill_input_embeds(
             dtype=self.model.dtype,
         )
 
+    @staticmethod
+    def _clear_pending_feedback_queue(data: Any) -> None:
+        queue = getattr(data, "pending_feedback_queue", None)
+        if queue is None:
+            return
+        if hasattr(queue, "clear"):
```
zhaochenyang20: never use `hasattr` and `getattr`

zhaochenyang20 (changes requested):

sglang_omni/models/moss_tts/model_runner.py
sjmshsh: done

Issue #254, opened 2026-04-05
[Feature] Ming-flash-omni-2.0 Enhancements Roadmap


yuan-luo:
## Motivation

  Following up the feature of https://github.com/sgl-project/sglang-omni/pull/236 for Ming-Omni integration in sglang-omni.

  ## Overview

  This issue tracks the development roadmap for Ming-Omni (`inclusionAI/Ming-flash-omni-2.0`) integration in
  sglang-omni. The current implementation supports a 6-stage speech pipeline (preprocessing → audio_encoder →
  mm_aggregate → thinker → decode + talker) with a 208GB MoE thinker and 0.5B dense talker.
 

  ## Current State

  | Component | Status | Details |
  |-----------|--------|---------|
  | Thinker | TP=1, CPU offload | 208GB MoE, 256 experts/layer, runs on single H200 with `cpu_offload_gb=110` |
  | Talker | 0.5B dense | Qwen2 (hidden=896) + CFM/DiT + AudioVAE, ~3.8GB VRAM, RTF=0.06 |
  | Audio Input | Functional | Whisper encoder + Conv1d projection, file-based only |
  | Image Input | Not implemented | Config exists (`VisionConfig`), no encoder stage |
  | Image Generation | Not implemented | No diffusion stage for image output |
  | Concurrency | Configurable | `max_conc` parameter controls CUDA graph pool size (default=1) |
  | Benchmark | WER 0.77% | seed-tts-eval EN 50 samples, with number normalization |

  ---

  ## Phase 1: Performance & Stability

  ### 1.1 Thinker Tensor Parallelism

  **Priority: High**

  The thinker model is architecturally TP-ready (uses `QKVParallelLinear`, `RowParallelLinear`,
  `VocabParallelEmbedding`, `FusedMoE`), but currently runs at TP=1 with heavy CPU offloading.

  **Current bottleneck:**
  - 208GB model on 140GB H200 requires `cpu_offload_gb=110`
  - Token latency ~2s/token due to CPU↔GPU weight transfer
  - Startup time ~7 min (FusedMoE weight sharding)

  **Target:**
  - [x] Enable `tensor_parallel_size=2` in pipeline config (`build_sglang_server_args()`)
  - [x] Test on 2-GPU and 4-GPU configurations
  - [x] Verify weight loading correctness with TP (attention heads, expert sharding)
  - [x] Benchmark: token latency, startup time, memory per GPU

[...]

PR #866, opened 2026-06-22
[bigfix] Ming-Omni Streaming TTS all-segment overlap bugfix 

AkazaAkane:
## Motivation

Ming-Omni streaming TTS currently only enables token-level streaming / vocoder overlap for the first internal TTS segment.

In `MingOmniTalker._process_segment`, the current logic is:

```python
use_stream = stream and (count == 0)
```

So when stream=True, only count == 0 goes through the streaming tts_job path. Later segments use stream=False, which waits for the full talker token generation to finish and then runs token2wav once. This means only the first segment overlaps talker token generation with vocoder decoding; subsequent segments regress to serial AR/talker → vocoder execution.

This is inconsistent with the intent of streaming TTS. The streaming talker scheduler forwards stream=True into omni_audio_generation, and its output path is already capable of emitting multiple audio chunks per request/segment.

## Modifications

Enable streaming for every segment (use_stream = stream) and refactor segment indexing logic. Compute segment_start_idx/segment_end_idx from prior integer keys in cache_position instead of relying on list(values()) ordering; introduce next_start_idx to track subsegment boundaries and clamp per-chunk start/end to the segment end. Update cache_position keys for subchunks to "{count}_{idx}" and slice text using indices relative to the segment start. Add a unit test (test_process_segment_streams_every_internal_segment) that verifies streaming flags and cache positions for consecutive segments.

## Related Issues

#254 

## Accuracy Test

<html><head></head><body><div _ngcontent-ng-c1753873157="" class="markdown markdown-main-panel enable-luminous-fast-follows stronger enable-updated-hr-color" style="--animation-duration: 400ms; --fade-animation-function: ease-out; font-family: Google Sans Text, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;" id="model-response-message-contentr_39045eccd90b86a7" aria-live="polite" aria-busy="false" dir="ltr"><p data-path-to-node="0" style="font-family: Google Sans Text, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">Ran the one-GPU PR-vs-main smoke on 1*H100 GPU.</p><p data-path-to-node="1" style="font-family: Google Sans Text, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">This was a targeted Ming streaming overlap smoke, not the standard SeedTTS benchmark. It imports the real <code data-path-to-node="1" data-index-in-node="106" style="font-family: Google Sans Text, sans-serif !important; line-height: 1.15 !important; margin-top: 0px !important;">MingOmniTalker._process_segment</code> from each branch, forces 3 internal segments, and simulates AR/vocoder timing so segment 1+ serial-vs-streaming behavior is measurable.</p>

Branch | stream flags | chunks by segment | total mean | segment gap mean
-- | -- | -- | -- | --
origin/main | [true, false, false] | {0: 4, 1: 1, 2: 1} | 922.1 ms | 435.9 ms
pr-866 | [true, true, true] | {0: 4, 1: 4, 2: 4} | 602.6 ms | 201.1 ms

## Checklist

- [x] Add unit tests.

## CI

if labeled.

### DIFF
--- a/sglang_omni/models/ming_omni/talker/modeling_ming_omni_talker.py
+++ b/sglang_omni/models/ming_omni/talker/modeling_ming_omni_talker.py
@@ -942,7 +942,17 @@ class MingOmniTalker(nn.Module):
                             stream=True,
                             last_chunk=last_chunk,
                         )
-                        yield {"tts_speech": this_tts_speech.cpu()}
+                        (
+                            this_tts_speech,
+                            self.sil_holder_cache[this_uuid],
+                        ) = self.silence_holder(
+                            this_tts_speech,
+                            audio_detokenizer.config.sample_rate,
+                            self.sil_holder_cache[this_uuid],
+                            last_chunk,
+                        )
+                        if this_tts_speech.numel() > 0:
+                            yield {"tts_speech": this_tts_speech.cpu()}
                 else:
                     future.result()
                     this_tts_speech_token = self.tts_speech_token_dict[this_uuid]
--- a/sglang_omni/models/ming_omni/talker/modeling_ming_omni_talker.py
+++ b/sglang_omni/models/ming_omni/talker/modeling_ming_omni_talker.py
@@ -1231,17 +1231,24 @@ class MingOmniTalker(nn.Module):
 
         for text_ori in text_list:
             length = len(text_ori)
-            previous_segment_positions = [
-                position
+            previous_segment_items = [
+                (key, position)
                 for key, position in cache_position.items()
                 if isinstance(key, int)
             ]
-            if not previous_segment_positions:
+            segment_key = (
+                max(key for key, _ in previous_segment_items) + 1
+                if previous_segment_items
+                else 0
+            )
+            if not previous_segment_items:
                 segment_start_idx = 0
             else:
-                segment_start_idx = previous_segment_positions[-1][1] + 1
+                segment_start_idx = max(
+                    position[1] for _, position in previous_segment_items
+                ) + 1
             segment_end_idx = segment_start_idx + length - 1
-            cache_position.update({count: (segment_start_idx, segment_end_idx)})
+            cache_position.update({segment_key: (segment_start_idx, segment_end_idx)})
 
             if not is_chinese(text_ori):
                 text = normalize_numbers(text_ori)
@@ -1306,7 +1313,7 @@ class MingOmniTalker(nn.Module):
                     )
                     next_start_idx = this_end_idx + 1
                     cache_position.update(
-                        {f"{count}_{idx}": (this_start_idx, this_end_idx)}
+                        {f"{segment_key}_{idx}": (this_start_idx, this_end_idx)}
                     )
                     rel_start_idx = this_start_idx - segment_start_idx
                     rel_end_idx = this_end_idx - segment_start_idx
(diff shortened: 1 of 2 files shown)


AkazaAkane:
## Streaming TTS validation

This PR fixes a streaming responsiveness bug in Ming-Omni TTS. The expected impact is limited to `stream=True` requests whose text is split into multiple internal TTS segments. Non-streaming SeedTTS and quality metrics such as WER / speaker similarity are not expected to improve.

### Unit / regression tests

```bash
pytest -q tests/unit_test/ming_omni/test_talker_voice_validation.py
# 20 passed

pytest -q tests/unit_test/ming_omni
# 184 passed

git diff --check origin/main...HEAD
# passed
```

Coverage added:

- Verifies consecutive internal segments call `tts_job(stream=True)`.
- Verifies cache positions remain contiguous across streamed segments.
- Verifies multiple fragments returned by `cut_text_by_semantic_length` get distinct integer cache keys and distinct subchunk keys.

AkazaAkane:
@edwingao28 Do you have any idea on what other tests we might need to prove this bug fix?

sglang_omni/models/ming_omni/talker/modeling_ming_omni_talker.py
```diff
@@ -1248,8 +1261,9 @@ def _process_segment(
             if text and text[0] == "\uff0c":
                 text = text[1:]
 
-            use_stream = stream and (count == 0)
+            use_stream = stream
```
edwingao28, line 1274: Switching every segment to the stream path also changes what `tts_job` does at the end. The non-stream branch runs `silence_holder` but the stream branch skips it, so on main, segments 1+ got inter-segment silence padding and now they don't. I suspect part of the gap drop in the table (435 down to 201 ms) is actually lost padding rather than pure overlap.

Is dropping it intended? If the padding was just smoothing the joins, removing it could make later segments abut or clip in a way WER and speaker-sim won't catch. Could we either run `silence_holder` on the stream branch too, or confirm with a quick PR vs main total-sample-count check that the joins stay clean?

tests/unit_test/ming_omni/test_talker_voice_validation.py
```diff
@@ -288,6 +288,121 @@ def _fake_detokenizer(sample_rate=44100, vae_patch_size=4, hop_size=480):
     return SimpleNamespace(encoder=encoder, config=config)
 
 
+def test_process_segment_streams_every_internal_segment():
```
edwingao28, line 291: These tests pin the `stream` flag and the cache positions, which is great for the indexing refactor. But since `tts_job` is faked to yield once, they don't actually exercise the thing the PR fixes, which is segments 1+ emitting multiple chunks. A change that re-serialized later segments while leaving `stream=True` plumbed would still pass.

I suggest we add one test with a fake `tts_job` that yields N chunks across 2+ segments and asserts per-segment chunk count > 1? Basically the {0:4, 1:4, 2:4} smoke from the PR body, committed as a unit test, so the every-segment-overlap claim is self-checking in CI?

edwingao28 (commented):
Thanks for the contribution! Leaved some comments for implementation and unittest

edwingao28 (approved):

yuan-luo (approved):

PR #867, opened 2026-06-22
[Refactor] T6: StageCapabilities vocabulary (#661)

MelodyyyYin:
Implements Template 6 of #661 — a static, per-model capability vocabulary so CI can route by capability instead of hardcoded `if model_name == "...":` branches.

## What this adds
- `sglang_omni/models/stage_capabilities.py` — frozen `StageCapabilities` dataclass (five `supports_*` fields, all default `False`) + `collect_capabilities()` / `format_capability_table()` / `log_capability_table()`.
- A `CAPABILITIES` constant in each of the 6 in-scope TTS model `__init__.py`. Values reflect only merged/verified optimizations (matches the coverage matrix); in-progress and N/A stay `False`.
- Startup log of the capability table in `_register_omni_model` (degrades to a warning on failure).
- CPU-only contract test `tests/unit_test/models/test_stage_capabilities.py`: presence check + defaults-all-False + frozen.

Behavior-preserving — declarations + a startup log only, no runtime path change. ~100 LOC.

## Not done (intentionally out of scope)
- **Runtime-consistency warning** (RFC mechanism 3: warn when the runner reports a capability the static declaration denies). It needs T4's `cuda_graph_eligible()` hook, which does not exist yet — deferred to T4 so this PR does not pull in T4 surface.

## Gotcha found
- The presence check is **not unconditionally CPU-only**: importing some model packages (e.g. `higgs_tts`) transitively pulls heavy deps (`vllm`) absent in a bare env. Following the codebase's existing tolerance for missing per-model deps (#680), the test **skips** models it cannot import; the presence guarantee is enforced in the full-dependency model CI.

## Test
`pytest tests/unit_test/models/test_stage_capabilities.py` → **8 passed, 1 skipped** (higgs skipped — `vllm` absent locally). CPU, no GPU. Registry unchanged (17 archs).

### DIFF
--- a/sglang_omni/model_runner/sglang_model_runner.py
+++ b/sglang_omni/model_runner/sglang_model_runner.py
@@ -114,6 +114,13 @@ class SGLModelRunner(ModelRunner):
         except Exception as exc:
             logger.warning("sglang-omni: skipping Ming-Omni registration (%s)", exc)
 
+        try:
+            from sglang_omni.models.stage_capabilities import log_capability_table
+
+            log_capability_table()
+        except Exception as exc:
+            logger.warning("sglang-omni: capability table log skipped (%s)", exc)
+
     def _profile_available_bytes(self, pre_model_load_memory: float) -> int:
         """Profile KV-cache headroom for colocated SGLang AR stages.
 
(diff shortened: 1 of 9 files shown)


MelodyyyYin:
Superseded by #957 (M8 ModelCapabilities, merged). The v2 RFC renamed StageCapabilities -> ModelCapabilities, and #957 landed the base + per-model metadata + contract test. Closing.

Issue #862, opened 2026-06-22
[Roadmap] Multi-node Inference Support and Communication Enhancement


JingwenGu0829:
## Goal
The goal of this issue contains three main things:
* Get the current intra-node communication infra corect
* Get the communication topology clear
* Support high-efficiency multi-node inference

## Summary

sglang-omni today is a **single-host, multi-GPU** serving system. The whole-pipeline
deployment unit is one machine: process launch, control-plane endpoints, and the
default data relay all assume every stage lives on the same host.

For the next-gen omni model we are about to support, this is a common pipeline:

```
VAD/ASR → audio/video encoder → LLM backbone (200B+) → talker (TTS) → code2wav
```

The full pipeline does not fit on a single 8×H100 node. We therefore need to be able
to place **different stages on different physical nodes** and stream tensors between
them across the network.

More importantly, a key feature of Sglang-Omni is to solve the layout between process, stage and GPU, to disaggregate them and provide a clean implementation surface.  **Now we need to solve the topology between process, stage and <node,GPU>.**

---

### 1. Topology model: node / gpu / process / stage

Introduce **node** as the top placement coordinate and make GPU indices explicit as
`(node, local_gpu)`:

```
node (physical host)
  └── gpu (device, node-local index)
        └── process (OS process; owns ≤1 gpu on its node)
              └── stage-shard (one rank of one stage)
```

The **unit of placement is the stage-shard (rank)**, and every
shard has an address `(node_id, local_gpu_id)`. A *stage* is then a set of shards:

```
stage = [ shard_0, shard_1, ... shard_{tp_size-1} ]   # one shard per rank
shard_r  -->  process  -->  (node, gpu)
```

- Today, all shards of a stage share one node (intra-node TP). → single-node config.
- **Room for one-stage-multi-node:** because a shard's address is `(node, gpu)`, a
  stage's shards *can* later be spread across nodes (multi-node TP/PP) with **no schema
  redesign** — only the launcher/collective-init need to learn it (Phase 5). We design
  the placement plan in terms of per-shard `(node, gpu)` now so this stays a pure
  extension.

---

### 2. Communication protocols

There are **two classes** of traffic,

- **Class I — inter-stage edges** (the pipeline graph: A→B handoff, fan-in, streaming).
  Carried by the **relay** (bulk tensor data) plus the **control plane** (small
  signals). This is the omni-specific layer.
- **Class II — intra-stage parallelism** (within one stage's ranks: TP/EP collectives,
[...]

PR #869, opened 2026-06-23
feat(relay): intra-node NVLink CUDA-IPC data plane + centralized transport

JingwenGu0829:
Important
WIP

Phase0 #862


sglang_omni/comm/stage_io.py
```diff
@@ -0,0 +1,389 @@
...
+        data_ref.buffer.length,
+        dtype=torch.uint8,
+        device=relay_device(relay),
+    )
+    op = await relay.get_async(
+        metadata=data_ref.buffer.info,
+        dest_tensor=buf,
+        request_id=request_id,
+    )
+    await op.wait_for_completion()
```
JiaxinD, line 582: Nice work, the NVLink data plane clearly does its job. The cross-GPU send drops from ~23% (shm round-trip on main) to ~1% here.
One thing worth a look though: the receive side stays on the stage's single asyncio loop. _read_transfer_buffer -> wait_for_completion() blocks per chunk, so as concurrency climbs the receives serialize and eat the send-side win (py-spy on talker at c32 sits right in this path). The credits=2 default in the relay would also throttle ~32 in-flight pretty hard. Getting the completion-wait off the loop's critical path (and/or a bigger credit budget) is probably the lever.
Lines up with #907: cheaper GPU movement doesn't help e2e if the receive stays serial on the host loop. Have the full A/B + profiles if useful.

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -975,20 +905,31 @@ async def _send_to_stage(
...
 
-        metadata, op = await relay_io.write_payload(
-            self.relay, request_id, projected_payload
+        if target in self._same_process_targets:
+            transport_kind, relay = self._comm.router.relay_for_materialized_payload(
+                target
+            )
+        else:
+            transport_kind, relay = self._comm.router.relay_for(target)
+        data_ref, op = await self._comm.write_payload(
```
luojiaxuan: This should be reconciled with current main before merge. `main` now has the lazy TensorRef path (`TensorRefPolicy.from_env`, materialization, cleanup, and `tensor_ref_stats`). This call path no longer passes a policy or releases externalized blobs on send failure, so a rebase would regress large inter-stage tensors unless the same semantics are ported into `comm.stage_io`.

sglang_omni/pipeline/mp_runner.py
```diff
@@ -105,7 +111,7 @@ def _build_stage_groups(
             stage_endpoints=stage_endpoints,
             stream_targets=list(stage_cfg.stream_to),
             stream_done_to_fn=stage_cfg.stream_done_to_fn,
-            same_gpu_targets=same_gpu_targets,
+            gpu_stage_names=gpu_stage_names,
```
luojiaxuan, line 125: Mooncake appears unreachable from normal launches: `StageLaunchConfig.remote_stage_names` is never populated here, so every real router gets an empty remote set and cross-node edges fall through to `cuda_ipc` or `shm`. Please wire a config-derived remote-edge source into the launch spec and cover it through `_build_stage_groups()`.

sglang_omni/comm/router.py
```diff
@@ -0,0 +1,184 @@
...
+        return TransportKind.SHM
+
+    def outbound_stream(self, target: str, data: torch.Tensor) -> TransportKind:
+        if not isinstance(data, torch.Tensor):
+            raise TypeError(
+                "relay-backed stream chunks must be torch.Tensor, got "
+                f"{type(data).__name__}"
+            )
+        kind = self.outbound(target)
+        if kind is TransportKind.CUDA_IPC and not data.is_cuda:
```
luojiaxuan: This makes stream transport depend only on stage placement. A GPU-stage edge that emits a CPU tensor now raises instead of falling back to SHM, and CPU tensors inside stream metadata are restored as CUDA tensors when sent through the CUDA-IPC relay. Can we select stream transport from the actual tensor/metadata device, or keep an SHM fallback for CPU chunks?

sglang_omni/relay/cuda_ipc.py
```diff
@@ -0,0 +1,567 @@
...
+    async def wait_for_completion(self, timeout: float = 30.0) -> None:
+        if self._completed:
+            return
+        wait_start = _comm_now_ns()
+        loop = asyncio.get_event_loop()
+        deadline = loop.time() + timeout
+        polls = 0
+        while not self._ack_map.is_done(self._ack_index):
+            if loop.time() > deadline:
+                raise TimeoutError("cuda_ipc receiver did not ack slot in time")
```
luojiaxuan: On timeout this raises before releasing the CUDA-IPC credit or dropping retained tensors. If a receiver crashes or aborts before acking, the sender can permanently lose a slot and eventually block future sends. Please release the slot safely or mark the relay failed so later puts do not hang behind leaked credits.

luojiaxuan (commented):
A few inline comments.

jzheng17:
Opened #941, stacked on this branch (base = `phase0-intra-node-nvlink-relay`), with three focused intra-node data-plane hardening fixes — each a separate, lint-clean commit:

1. `fix(relay/cuda_ipc)` — remove the completion busy-spin (spin-then-backoff instead of `await asyncio.sleep(0)`) and release the CUDA-IPC slot on timeout (fixes the credit leak that could permanently consume 1 of 2 credits).
2. `fix(comm/router)` — transport follows the actual tensor device: a GPU-placed edge carrying a CPU tensor falls back to SHM instead of raising (the `router.py:67` review).
3. `fix(relay/shm)` — real bounded-outstanding backpressure, replacing the fire-and-forget no-op semaphore (unlink-as-ack).

Merging #941 folds these into this PR. A test & measurement plan for the completion/backpressure constants is posted as a comment there.

Heads-up: #941's red `lint` is **inherited from this branch** — pre-existing isort/black debt in `comm/__init__.py`, `comm/engine.py`, `comm/data_ref.py` (this branch's own CI is red on the identical hooks). I deliberately did not reformat those files to avoid colliding with your polish; the three commits above are lint-clean.

sglang_omni/comm/stage_io.py
JingwenGu0829: Good call! Updated the code. Now we have a event-based fully async sender-receiver schema right now.

sglang_omni/pipeline/mp_runner.py
JingwenGu0829: Mooncake is out of scope for this pr. Will do in the next phase (pr).

sglang_omni/comm/engine.py
```diff
@@ -0,0 +1,367 @@
...
+        self.router.close()
+
+    def ack_transfer(self, ack: DataAckMessage) -> None:
+        if ack.to_stage != self.router.stage_name:
+            raise ValueError(
+                f"data_ack for {ack.to_stage!r} delivered to {self.router.stage_name!r}"
+            )
+        pending = self._pending.get(ack.object_id)
+        if pending is None:
+            raise RuntimeError(f"unexpected data ack for {ack.object_id!r}")
```
luojiaxuan: Late or duplicate acks can arrive after the pending entry was timed out, popped, or failed during shutdown. Raising here bubbles through Stage._handle_message and can take down the sender for a stale ack, so this should be idempotent or logged instead of fatal.

sglang_omni/comm/engine.py
```diff
@@ -0,0 +1,367 @@
...
+    def _arm_pending(self, object_id: str) -> None:
+        pending = self._pending[object_id]
+        pending.task = asyncio.create_task(self._watch_pending(object_id, pending))
+        self._track_task(pending.task, f"comm ack {object_id}")
+
+    async def _watch_pending(self, object_id: str, pending: _PendingTransfer) -> None:
+        try:
+            await asyncio.wait_for(pending.ack, timeout=self._ack_timeout_s)
+            for op in pending.ops:
+                op.mark_receiver_done()
```
luojiaxuan, line 419: Mooncake PutOperation does not implement these receiver-ack hooks, so a successful data_ack for a Mooncake transfer will raise NotImplementedError here and the comm ack task reports a stage background failure. If Mooncake stays selectable in CommRouter, this path needs either no-op guarding or Mooncake-specific hook implementations.

sglang_omni/comm/router.py
```diff
@@ -0,0 +1,235 @@
...
+        if not devices or devices == {"cpu"}:
+            return TransportKind.SHM
+        if devices == {"cuda"}:
+            if self.self_is_gpu and target in self.gpu_stage_names:
+                return TransportKind.CUDA_IPC
+            raise ValueError(
+                f"cuda payload cannot be sent from {self.stage_name!r} to "
+                f"non-GPU target {target!r}"
+            )
+        raise ValueError(f"mixed or unsupported tensor devices in payload: {devices}")
```
luojiaxuan, line 145: This rejects mixed CPU and CUDA payloads, but Qwen3-Omni thinker inputs can combine CUDA encoder embeddings with CPU grid or length tensors from preprocessing. The default speech topology sends mm_aggregate to thinker across processes, so this can raise before delivery unless mixed-device payloads are split or materialized by a supported transport.

sglang_omni/relay/shm.py
```diff
@@ -46,33 +44,60 @@ def __init__(self, metadata: Any):
...
-        # Sender simply closes the local handle; Receiver is responsible for unlinking.
-        if not self._completed:
-            self._shm_obj.close()
+        if self._completed:
+            return
+        try:
+            await asyncio.wait_for(self._receiver_done, timeout=timeout)
+        except TimeoutError:
+            self._unlink_if_present()
+            raise
```
luojiaxuan: The new test expects this timeout to mention was not consumed in time, but re-raising the bare wait_for TimeoutError has an empty message. Please raise a contextual TimeoutError after unlinking so the test and the failure log match the intended behavior.

luojiaxuan (commented):
Follow-up review comments.

yxs (approved):

sglang_omni/relay/cuda_ipc.py
```diff
@@ -0,0 +1,913 @@
...
+        return CudaIpcPutOperation(
+            metadata,
+            ready_event=ready_event,
+            source_tensor=flat,
+            slot_index=slot_index,
+            num_slots=num_slots,
+            request_id=request_id,
+            size=size,
+            release_cb=lambda: allocator.release(offset, num_slots),
+            fail_cb=self._mark_failed,
```
yxs, line 786: Not blocking.

On `wait_for_completion` timeout, `fail_cb=self._mark_failed` marks the whole relay failed, so one timed-out receiver takes down every subsequent transfer on that shared per-edge relay.

It can hit healthy traffic: the relay is shared by all concurrent A→B requests, and completion waits are offloaded to an 8-thread `_wait_executor`. Under load, one request's ack can be delayed enough to reach the 30s ack_timeout_s while other requests on the same relay complete fine, so this can fail concurrent healthy requests, not just a genuinely-dead receiver.

The this an alternative, quarantine the timed-out slot and keep the relay alive, but isn't free either: the slot can't be safely reused (the peer may still read it), so it leaks per timeout and sustained timeouts slowly drain the pool.

So,

1. keep this, fail loud, but ensure completions can't spuriously time out under load
2. fail only the one transfer + bounded quarantine reclaimed when the receiver stage is known dead

yxs: Deferring the ack-timeout failure-policy redesign to Phase 3, it needs
reliable peer failure-detection, which Phase 3 builds (where the intra-node data plane and cross-node relay converge).

Target behavior could be: a slow ack must not fail the shared edge.
- soft timeout -> backpressure: hold the slot, let a slow-but-alive receiver finish (both localities);
- only a hard timeout counts as real death; the action then depends on locality, intra-node fails the edge loud, cross-node fails just that transfer and quarantines its slot (reclaimed once the peer is known dead).

PR #873, opened 2026-06-23
Support Dots tts 

### LINKED ISSUE #717: [Feature] Interest in supporting dots.tts continuous-latent TTS?
### Checklist

### Motivation

[dots.tts](https://huggingface.co/collections/rednote-hilab/dotstts) was recently released under Apache-2.0, and I wanted to ask whether supporting it is aligned with sglang-omni's roadmap.

It seems different from the current TTS integrations. Existing models such as Higgs TTS and Qwen3-TTS mostly generate discrete audio codes before vocoder decoding. dots.tts appears to use a continuous path instead: a Qwen2.5-based AR backbone drives a flow-matching acoustic head to generate continuous AudioVAE latent patches, which are then decoded by AudioVAE.

Because of that, this does not look like a simple adapter over the existing audio-code TTS path. It may require support for continuous-latent autoregressive TTS generation if the goal is native scheduling, batching, or streaming.

### Related resources

_No response_

albaNnaksqr:
## Summary

Close #717

This is a draft PR for native dots TTS support in `sglang-omni`.

Pipeline:

```text
preprocessing -> sglang-backed latent_engine -> vocoder
```

## Main changes:

- Add `DotsTTSForConditionalGeneration` registration and `DotsTTSPipelineConfig`.
- Load dots TTS core, Qwen2 config, speaker encoder, AudioVAE vocoder, tokenizer/config, latent stats, and safetensors artifacts natively.
- Remove the serving-time dependency on the external `dots_tts` Python runtime package.
- Vendor the required dots inference components under `_vendored/`, with unused training/runtime-only code removed.
- Move generic text preprocessing out of the vendored dots tree.
- Run the Qwen2 backbone through the SGLang model path.
- Move latent decode orchestration into `DotsTTSSGLangModel` / `DotsTTSModelRunner`.
- Keep dots DiT/flow, patch encoder, speaker encoder, and AudioVAE as native PyTorch side modules.
- Stream latent chunks from the latent engine to the vocoder stage.
- Support concurrent native requests in the v1 scheduler path.
- Batch compatible DiT/flow side steps across requests with the same flow batch key.
- Explicitly restrict v1 to `tp_size == 1`.

## Scope

Implemented in this PR:

- Native dots TTS pipeline stages.
- Native request builder, model runner, latent stream output, and AudioVAE vocoder stage.
- Native checkpoint/config loading for dots model artifacts.
- Speech API propagation of `max_new_tokens` into dots TTS generation params.
- Cross-request DiT/flow side-step batching for compatible requests.
- Unit coverage for registry, request building, native adapter, model runner, SGLang model behavior, pipeline behavior, and weight loading.
- GPU smoke coverage for native latent generation, batched flow decode, and speech API serving.

Not implemented in this PR:

- Tensor parallel support for dots side modules.
- Generalized DiT/flow batching across incompatible request shapes/schedules.
- Async side-step decode / overlap schedule.
- CUDA graph coverage for dots side modules.
- `tts_interleave` / double-streaming mode.
- Large-scale official Seed-TTS-Eval or multilingual benchmark.

### Unit / Static Checks

```bash
python -m pytest -q tests/unit_test/dots_tts tests/unit_test/serve/test_speech_protocol.py
python -m compileall -q sglang_omni/models/dots_tts tests/unit_test/dots_tts
git diff --check
pre-commit run --all-files --show-diff-on-failure
```

Result:

```text
120 passed
compileall passed
git diff --check passed
pre-commit passed
```

### GPU Environment

Tested on a single-GPU Spark machine:

```text
GPU: NVIDIA GB10
CUDA: 13.0
Python: 3.10
PyTorch: 2.11.0+cu130
SGLang: 0.5.12.post1
Transformers: 5.6.0
```

### Native GPU Smoke

Native latent/vocoder smoke result:

```text
latent_shape: (1, 4, 128)
latent_finite: True
feedback_shape: (1, 1, 1536)
feedback_finite: True
waveform_shape: (1, 1, 7680)
waveform_finite: True
sample_rate: 48000
```

### Batched DiT/Flow GPU Smoke

Two compatible requests were decoded in one DiT/flow side-step batch:

```text
step_fm_call_count: 1
step_fm_shapes: [(2, 5, 1024)]
latent_shapes: [(1, 4, 128), (1, 4, 128)]
latent_finite: [True, True]
feedback_shapes: [(1, 1, 1536), (1, 1, 1536)]
feedback_finite: [True, True]
next_token_ids: [0, 0]
finished: [False, False]
```

### Native vs Original Runtime Parity Smoke

Small EN/ZH smoke set with the same prompt/audio settings. This is parity smoke evidence, not an official benchmark.

| Metric | dots runtime | sglang-omni native | Delta |
|---|---:|---:|---:|
| WER mean, EN small set | 0.375 | 0.375 | +0.000 |
| CER mean, EN+ZH small set | 0.4069 | 0.4069 | +0.0000 |
| Speaker SIM | 78.57 | 78.12 | -0.45 |
| Sample rate | 48000 | 48000 | match |
| NaN/Inf | none | none | pass |

Notes:

- WER/CER uses Whisper `base` over a 3-sample smoke set: two English samples and one Chinese sample.
- Speaker SIM uses the dots checkpoint's CAM++ x-vector speaker encoder, scored as cosine similarity * 100 between prompt audio and generated audio.
- These numbers are smoke evidence, not a replacement for the official dots benchmark.

## Limitations / Follow-ups

- Extend DiT/flow batching beyond requests with identical flow batch keys.
- Add request-scoped side workspaces before enabling async decode.
- Add TP support for dots side modules.
- Add CUDA graph coverage where useful for side modules.
- Add larger benchmark coverage if this integration becomes a maintained model path.


albaNnaksqr:
Hi @zhaochenyang20  — this draft for native dots.tts (rednote-hilab) support
(closes #717) is functionally complete and I'd like to provide a quick update.

The Qwen2 AR backbone runs through the SGLang model path; the
DiT/flow head, patch encoder, CAM++ speaker encoder and the 48 kHz AudioVAE stay as native side modules. Latent chunks stream from the latent engine into the vocoder stage. Scoped to tp_size == 1 for v1.

Beyond a basic port. On top of the single-request pipeline it adds batched DiT/flow decode across requests (grouped by flow batch key) and concurrent native requests in the v1 scheduler — i.e. it targets the serving path, not just offline generation.

Vendoring. Upstream dots.tts isn't on PyPI, so the inference modules (AudioVAE, DiT/flow, patch + speaker encoders) are vendored under _vendored/, with training/runtime-only code stripped, pinned to upstream rednote-hilab/dots.tts@b995bdb7 (2026-06-16). Provenance and the upstream/integration boundary are documented in _vendored/VENDORED.md. 

On the vendored footprint. The bulk (~7k LOC) is model-specific acoustic math — the DiT/flow head (torchdiffeq ODE + MeanFlow path), the BigVGAN-style 48 kHz AudioVAE, the PyTorch CAM++ encoder, and the streaming semantic/patch encoder.
sglang-omni doesn't yet expose a shared home for continuous-latent flow-matching TTS, so dots carries its own. 

I looked at whether it could reuse the existing ming_omni talker (CFM + AudioVAE), but the implementations don't actually share: different ODE solver, a different VAE (Qwen2+ISTFT vs BigVGAN), different latent dims (64 vs 128) and sample rate (16 k vs 48 k), and different checkpoint layouts. So the size is intrinsic rather than duplication of existing code.

Two things I'd value your read on before un-drafting:
1. Is the vendoring + native-integration approach the direction you want, or should dots stay an external dependency?
2. Any concerns with the scheduler integration or the tp_size == 1 v1 restriction?

Keeping it as draft until I hear back. Thanks!

Hayden727:
model support merged on #1349

PR #874, opened 2026-06-23
[MOSS] Decouple MOSS-TTS Local codec from processor

Ratish1:
## Motivation

MOSS-TTS Local preprocessing and vocoder stages previously reached through the Hugging Face processor-owned `audio_tokenizer`. This keeps the codec load and processor state coupled, which makes the local codec path harder to reason about and harder to optimize independently.

This PR loads the MOSS-Audio-Tokenizer-v2 codec as an owned component for MOSS-TTS Local while keeping the processor responsible for text/config handling.

## Modifications

- Add an owned `MossTTSLocalAudioTokenizer` wrapper that loads `OpenMOSS-Team/MOSS-Audio-Tokenizer-v2`, prepares reference waveforms with the upstream channel/resample/loudness semantics, and exposes a narrow encode API.
- Load the MOSS-TTS Local processor without constructing its codec, then load the codec separately for preprocessing and vocoder stages.
- Route reference file paths and data-URI references through the owned codec encoder with the existing batching/cache path.
- Pass the owned codec model directly into the streaming vocoder scheduler instead of deriving it from `processor.audio_tokenizer`.
- Add unit coverage for owned codec loading, waveform preparation, batched reference encoding, data-URI handling, and vocoder executor construction.

## Validation

Unit gate:

```text
115 passed, 20 warnings
```

Full SeedTTS EN, `--max-concurrency 8`:

| metric | value |
|---|---:|
| completed / failed | `1088 / 0` |
| WER evaluated / skipped | `1088 / 0` |
| throughput_qps | `7.138` |
| rtf_mean / p95 / p99 | `0.2637 / 0.3794 / 0.4389` |
| latency mean / p95 / p99 | `1.119s / 1.530s / 1.717s` |
| output throughput | `392.5 tok/s` |
| output tokens total | `59832` |
| WER corpus | `2.28%` |
| WER below 50 | `1.83%` |
| >50% WER outliers | `5` |
| UTMOS mean | `3.9542` |
| speaker similarity mean | `65.1585` |

Audio integrity:

| check | value |
|---|---:|
| files | `1088` |
| missing | `0` |
| read errors | `0` |
| silent / zero-peak | `0` |
| empty | `0` |
| sample rate | `48000` for all |
| channels | stereo for all |
| duration min / mean / max | `1.68s / 4.399s / 9.36s` |
| full-scale clipped sample count | `20` |

### DIFF
--- a/sglang_omni/models/moss_tts_local/audio_tokenizer.py
+++ b/sglang_omni/models/moss_tts_local/audio_tokenizer.py
@@ -131,6 +131,7 @@ class MossTTSLocalAudioTokenizer:
 
     @staticmethod
     def _loudness_normalize(wav: torch.Tensor) -> torch.Tensor:
+        wav = wav.to(torch.float32)
         if wav.numel() == 0:
             return wav
         current_dbfs = 10.0 * torch.log10(torch.mean(wav**2) + 1e-9)
--- a/sglang_omni/models/moss_tts_local/audio_tokenizer.py
+++ b/sglang_omni/models/moss_tts_local/audio_tokenizer.py
@@ -1,0 +1,168 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Owned MOSS-Audio-Tokenizer-v2 loader for MOSS-TTS Local."""
+
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+import torch
+
+from sglang_omni.models.moss_tts.stages import (
+    _moss_transformers_processor_compat,
+    _resolve_checkpoint,
+)
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_MOSS_TTS_LOCAL_AUDIO_TOKENIZER = "OpenMOSS-Team/MOSS-Audio-Tokenizer-v2"
+_LOUDNESS_TARGET_DBFS = -20.0
+_LOUDNESS_GAIN_MIN_DB = -3.0
+_LOUDNESS_GAIN_MAX_DB = 3.0
+
+
+class MossTTSLocalAudioTokenizer:
+    """Narrow encode API plus owned access to the upstream v2 codec model."""
+
+    def __init__(
+        self,
+        model: Any,
+        *,
+        device: str,
+    ) -> None:
+        self.model = model
+        self.device = str(device)
+        config = model.config
+        self.sample_rate = int(config.sampling_rate)
+        self.number_channels = int(config.number_channels)
+
+    def encode_paths(
+        self,
+        paths: list[str],
+        *,
+        num_quantizers: int,
+    ) -> list[torch.Tensor]:
+        if not paths:
+            raise ValueError("paths must contain at least one audio path")
+        return self.encode_waveforms(
+            self.load_paths(paths),
+            num_quantizers=num_quantizers,
+        )
+
+    def load_paths(self, paths: list[str]) -> list[tuple[torch.Tensor, int]]:
+        import torchaudio
+
+        waveforms = []
+        for path in paths:
+            wav, sample_rate = torchaudio.load(path)
+            if int(sample_rate) != self.sample_rate:
+                wav = torchaudio.functional.resample(
+                    waveform=wav,
+                    orig_freq=int(sample_rate),
+                    new_freq=self.sample_rate,
+                )
+            waveforms.append((wav, self.sample_rate))
+        return waveforms
+
+    def encode_wavs(
+        self,
+        wavs: list[torch.Tensor],
+        sample_rate: int,
+        *,
+        num_quantizers: int,
+    ) -> list[torch.Tensor]:
+        return self.encode_waveforms(
+            [(wav, int(sample_rate)) for wav in wavs],
+            num_quantizers=num_quantizers,
+        )
+
+    def encode_waveforms(
+        self,
+        waveforms: list[tuple[torch.Tensor, int]],
+        *,
+        num_quantizers: int,
+    ) -> list[torch.Tensor]:
+        if not waveforms:
+            raise ValueError("waveforms must contain at least one waveform")
+        prepared = [
+            self._prepare_waveform(wav, sample_rate) for wav, sample_rate in waveforms
+        ]
+
+        with torch.inference_mode():
+            encoded = self.model.batch_encode(
+                prepared,
+                num_quantizers=int(num_quantizers),
+            )
+        audio_codes = encoded.audio_codes
+        audio_codes_lengths = encoded.audio_codes_lengths
+        if audio_codes is None or audio_codes_lengths is None:
+            raise RuntimeError(
+                "MOSS-TTS Local audio tokenizer encode returned empty "
+                "audio_codes/audio_codes_lengths"
+            )
+        codes_cpu = audio_codes.detach().to("cpu", torch.long)
+        lengths_cpu = audio_codes_lengths.detach().to("cpu")
+        return [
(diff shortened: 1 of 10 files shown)


sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -848,8 +833,6 @@ def _decode_codes_rows(self, codes_list: list[torch.Tensor]) -> list[torch.Tenso
             # The processor helper forces chunk_duration=8 and enters the
             # tokenizer streaming loop. This decoder is non-streaming, so it must
             # run through the tokenizer's full-sequence decode path.
-            # TODO(Ratish): Load and own the non-streaming codec directly so this
-            # path does not need to swap the processor-owned decoder at call time.
             original_decoder = self._codec.decoder
```
JiaxinD, line 836: Confirmed: the #822 decoder swap and the #798 graph capture both operate on `self._codec` here, which is now the directly-owned codec, so the mechanism is unchanged and I verified there is no break. Removing the TODO is right since the ownership half landed; the swap staying is expected (one codec, two decode modes).

sglang_omni/models/moss_tts_local/audio_tokenizer.py
```diff
@@ -0,0 +1,169 @@
...
+                "audio_codes/audio_codes_lengths"
+            )
+        codes_cpu = audio_codes.detach().to("cpu", torch.long)
+        lengths_cpu = audio_codes_lengths.detach().to("cpu")
+        return [
+            codes_cpu[:, index, : int(lengths_cpu[index])].transpose(0, 1).contiguous()
+            for index in range(int(codes_cpu.shape[1]))
+        ]
+
+    def _prepare_waveform(self, wav: torch.Tensor, sample_rate: int) -> torch.Tensor:
```
JiaxinD, line 110: Small ordering asymmetry: path jobs resample in `load_paths` before this folds channels, while wav/data-URI jobs fold here then resample. It commutes for the repeat/truncate cases so output matches, but it is a latent footgun, and `test_audio_tokenizer_path_resamples_before_channel_fold` feeds stereo input so the fold is a no-op there (it does not actually exercise resample-before-fold).

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -488,9 +580,14 @@ def create_preprocessing_executor(
...
-    processor = _load_moss_tts_local_processor(model_path, device=device)
+    processor = _load_moss_tts_local_processor(model_path)
+    audio_tokenizer = load_moss_tts_local_audio_tokenizer(
+        _resolve_audio_tokenizer_model_path(processor, codec_model_path),
+        device=device,
+    )
     reference_encoder: Any = _BatchedReferenceEncoder(
-        processor,
+        audio_tokenizer,
+        n_vq=int(processor.model_config.n_vq),
```
JiaxinD, line 590: `num_quantizers` is now passed explicitly as `model_config.n_vq`, whereas the old `processor.encode_audios_from_path` took no such arg. Worth confirming `n_vq` matches the codec's default quantizer count so reference codes stay identical to the previous path.

sglang_omni/models/moss_tts_local/audio_tokenizer.py
```diff
@@ -0,0 +1,169 @@
...
+        f"Loading MOSS-TTS Local audio tokenizer from {checkpoint_dir} on {device}"
+    )
+    try:
+        from transformers import AutoModel
+
+        with moss_transformers_processor_compat():
+            model = AutoModel.from_pretrained(
+                checkpoint_dir,
+                trust_remote_code=True,
+            )
```
SandyLuXY, line 160: The owned MOSS-TTS Local codec loader calls `AutoModel.from_pretrained(...)` without passing `codec_weight_dtype="bf16"`. The old processor-owned loading path defaulted the codec weights to bf16 before loading `MOSS-Audio-Tokenizer-v2`; without that kwarg, the codec model falls back to fp32 when its config does not provide `codec_weight_dtype`.

SandyLuXY: Please check line 216 of https://huggingface.co/OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5/blob/main/processing_moss_tts.py

sglang_omni/models/moss_tts_local/stages.py
Ratish1: Verified this, HF _assert_fixed_nq(None) resolves to model_config.n_vq, and it shows same-mode upstream/candidate reference codes are exact.

sglang_omni/models/moss_tts_local/audio_tokenizer.py
Ratish1: Good catch. Fixed by passing codec_weight_dtype="bf16" to AutoModel.from_pretrained, matching HF processing_moss_tts.py.

sglang_omni/models/moss_tts_local/audio_tokenizer.py
Ratish1: Done. I checked the current HF processor: `encode_audios_from_path` resamples before delegating to `encode_audios_from_wav`, while direct wav/data-URI encoding folds before resample. Kept that behavior and changed the test to use mono input so it actually verifies path resample happens before channel fold.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -62,6 +70,40 @@ class _ArMemoryBudget:
...
+    audio, sample_rate = sf.read(io.BytesIO(raw), dtype="float32", always_2d=True)
+    duration = audio.shape[0] / max(int(sample_rate), 1)
+    if duration > _MAX_REFERENCE_SECONDS:
+        raise ValueError(
+            f"reference audio is {duration:.1f}s long, limit is "
+            f"{_MAX_REFERENCE_SECONDS:.0f}s"
+        )
+    return torch.from_numpy(audio.T), int(sample_rate)
+
+
```
Ccyest: Nit: Looks like these two functions are helpers extracted from `CachedReferenceEncoder.encode_data_uri`, now placed as module-level functions at the top of stages.py. 

Consider moving these two functions into `audio_tokenizer.py` or making them @staticmethod on _BatchedReferenceEncoder to reduce stages.py bulk

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -142,34 +184,50 @@ def _resolve_codec_device(device: str | None, gpu_id: int | None) -> str:
...
+) -> str:
+    if codec_model_path is not None:
+        return codec_model_path
+    return str(
+        getattr(
+            processor.model_config,
+            "audio_tokenizer_name_or_path",
+            DEFAULT_MOSS_TTS_LOCAL_AUDIO_TOKENIZER,
+        )
+        or DEFAULT_MOSS_TTS_LOCAL_AUDIO_TOKENIZER
```
Ccyest: Nit: getattr function should already have a default value, wondering why we have a "or" guarding here?

Ccyest (approved):
lgtm, just some nit change request

sglang_omni/models/moss_tts_local/stages.py
Ratish1: correct, getattr now uses only the explicit default

sglang_omni/models/moss_tts_local/stages.py
Ratish1: done. moved the data-URI parse/decode helpers into _BatchedReferenceEncoder as private static methods.

PR #875, opened 2026-06-23
Refactor-T5a pilot: BatchVocoderBase + ref (qwen3_tts) + Qwen3-TTS migration

keke0315:
## Summary

This PR implements the first `T5a` pilot from #661 

- build a shared `BatchVocoderBase` for **non-streaming** TTS vocoders
- migrate `qwen3_tts` to the new backbone as the reference model
- delete the old Qwen3-TTS vocoder path

The goal is to establish a single template-method backbone for batch vocoders, where the base owns the control flow and model-specific code is narrowed to hooks.

## Scope

Included:
- shared `BatchVocoderBase`
- reusable hook structure for:
  - `prepare_item(...)`
  - `decode_batch(...)`
  - `store_result(...)`
- Qwen3-TTS non-streaming vocoder migration

To-do next:
- 1 follow-up PR (moss, voxtral..) 
- T5b StreamingVocoderBase

## Design

Follows the #661  refactor requirement that the **base owns the skeleton** and the model overrides only narrow hooks.

For Qwen3-TTS:
- `prepare_item(...)` extracts and validates vocoder inputs from payload/state
- `decode_batch(...)` performs batch decode through the existing tokenizer path
- `store_result(...)` writes audio payload, sample rate, modality, and usage back to the payload/state

Single-item execution reuses the same backbone as batch execution to avoid drift between `_vocode` and `_vocode_batch`.

## Behavior Preservation

This PR is intended to be behavior-preserving for Qwen3-TTS:
- still errors when `audio_codes` are missing
- preserves `ref_code_len` trimming behavior
- preserves `usage` writeback
- preserves `sample_rate`, `modality`, and final audio payload behavior
- still errors when batch decode count mismatches request count

## One-Backbone Guarantee

- Qwen3-TTS now uses the new backbone as its only non-streaming vocoder path


keke0315:
According to the [RFC file](https://ylarkus3pw750a4b.usttp.larksuite.com/docx/XHA8dNxOhopLKdxqtDDuQTULtCd) the original proposal appears to place both T5a (`BatchVocoderBase`) and T5b (`StreamingVocoderBase`) in a shared `vocoder_base.py`.

```text
sglang_omni/scheduling/vocoder_base.py
├── BatchVocoderBase
└── StreamingVocoderBase (t5b)
```

This PR introduces a separate`batch_vocoder.py`, which suggests that T5b may later use another separate file `streaming_vocoder.py`. That separation could be reasonable given T5b’s higher complexity

```text
sglang_omni/scheduling/
├── batch_vocoder.py
└── streaming_vocoder.py (t5b)
```

Please let me know which one to take, will update accordingly

PR #336, opened 2026-04-23
[Ming-Omni] Support diffusion based image generation 

yuan-luo:
## Motivation

To close https://github.com/sgl-project/sglang-omni/issues/257
Design Spec: https://github.com/sgl-project/sglang-omni/issues/304

image

## Checklist

- [ ] Format your code according with pre-commit.
- [ ] Add unit tests.
- [ ] Update documentation / docstrings / example tutorials as needed.
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed.
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.

PR #878, opened 2026-06-24
[Feat] Ming image generation

### LINKED ISSUE #876: [RFC] Support Ming-Omni Image Generation
## Summary

This RFC proposes an PR design for Ming-Omni image generation. The goal is not only attach a diffusion pipeline to the existing chat API but also make image generation a first class pipeline terminal with explicit contracts between serving, preprocessing, thinker hidden-state capture, semantic conditioning, image backend invocation, response assembly, testing, and challenge documentation.

The initial target is intentionally scoped for a focused, reviewable implementation on the current branch: an `image` pipeline profile with non-streaming text-to-image responses for requests whose output modality is exactly `["image"]`. Media-conditioned image generation, text+image co-output, the `full` omni topology, streaming image deltas, ByT5 text-rendering enhancement, and alternate diffusion backends are deferred to a separate scope file.

The central design is:

```text
HTTP request
  -> normalized image-generation intent
  -> preprocessor appends image query tokens
  -> thinker runs prefill-only hidden-state capture
  -> semantic conditioner projects query hidden states
  -> image backend generates image
  -> API returns image payload with clear status metadata
```

The key improvement is that every high-risk boundary in the initial PR becomes explicit and testable: image-generation routing, query-token span ownership, radix-cache alignment, hidden-state shape contracts, fallback behavior, image terminal placement, and response schema.

---

## Problem Statement

This RFC is build and revised on top of V0 image gen design spec  https://github.com/sgl-project/sglang-omni/issues/304 and its relevant implementation https://github.com/sgl-project/sglang-omni/pull/336, special thanks to @yuan-luo on this!

Ming-Omni image generation is deceptively complex because it crosses four system boundaries:

1. The public serving contract decides whether a request asks for image output.
2. The thinker runtime must expose the correct hidden states for image query tokens.
3. A semantic conditioner must translate thinker hidden states into image-backend conditioning.
4. An image backend must generate and return image bytes through the image terminal.

A prototype can appear to work while still hiding serious correctness risks:

- `image_generation` parameters can be confused with the routing switch for image output.
- Hidden states may represent only the uncached suffix of a prompt when radix prefix cache or chunked prefill is active.
- A `gen_mask` built for the full sequence can be applied to a shorter captured hidden-state tensor.
- Fallback from semantic conditioning to text-only or random embeddings can produce an image-shaped result while silently losing the intended semantics.
- `bailing_moe_model.py`-style code can be misread as the diffusion model, even though the actual denoising DiT lives behind the diffusion backend.
- E2E GPU validation is expensive, so unit tests can accidentally overfit mocked behavior and miss real runtime/cache failures.

This RFC designs the PR around those failure modes instead of treating them as incidental bugs.

---

## Goals

- Provide a first-class image terminal stage for Ming-Omni.
- Use `modalities=["image"]` as the user-facing output intent, with `image_generation` as parameterization.
- Reuse the already-loaded Ming thinker hidden states rather than loading a second full LLM in the production path.
- Define a strict semantic-conditioning contract:

  ```text
  query_hidden:      [B, 256, 4096]
  condition_embeds:  [B, 256, 2560]
  ```

- Make radix-cache and chunked-prefill alignment explicit, validated, and testable.
- Avoid silent degradation by default.
- Keep image generation behind a small backend interface.
- Return images through a clear non-streaming chat-completion response schema.
- Keep media-conditioned image generation, text+image co-output, `full` omni, streaming image responses, ByT5, and alternate diffusion backends outside the initial PR scope.
- Provide a testing plan that separates CI-safe coverage from optional manual GPU smoke evidence.
- Maintain active challenge communication as part of the engineering process.

## Design Frame: What This PR Is Building

The core design choice is to treat Ming image generation as a reusable image-generation terminal, not a one-off request handler. The immediate implementation is for Ming plus one diffusion backend, but the interfaces should separate the pieces that can realistically generalize from the pieces that are model-family specific.

There are three model families worth separating before implementation starts:

| Family | Shape | Examples | Initial decision |
|---|---|---|---|
| A. Decoupled diffusion | Thinker or connector produces conditioning; an external diffusion backend denoises | Ming + Z-Image-style backends | Implement this path |
| B. In-context denoising | The thinker itself, or thinker KV/cache, participates in denoising | BAGEL-style and SenseNova-U1-style systems |  document the boundary |
| C. AR-VQ image decoding | The model emits image tokens autoregressively instead of running diffusion | Janus-Pro-style systems | Out of diffusion scope, but we wnat to reuse response/codec pieces |

The reusable foundation is intentionally small:

- request normalization and output-modality routing
- image terminal lifecycle
- image-generation parameter normalization
- hidden-state capture metadata
- semantic-conditioning bundle shape
- image backend boundary
- image result serialization and status metadata

The family-specific layer is where Ming's 256 query tokens, hidden-state projection, and diffusion backend defaults live. That keeps the first PR honest: it implements Family A well, while leaving enough structure for later Family B or AR-VQ work without pretending those systems can be plugged in by registering a backend class.

## Relationship To The Existing Design Spec And PR #336 
This RFC intentionally keeps the first implementation narrower than the earlier image-generation design in #304 and the merged diffusion PR #336.

The earlier design explored the full image-generation stack: standalone semantic encoding, ZImage, SD3 comparison, ByT5 text-rendering support, and full pipeline integration. This RFC keeps the parts that are required for a robust production path and defers the rest until the base image terminal is validated.

The main differences are:

- This RFC starts from the thinker-fused path instead of shipping a standalone 200GB semantic encoder as the normal serving path.
- `modalities=["image"]` is the only routing switch. `image_generation` configures generation but does not route by itself.
- Text+image co-output, streaming image deltas, media-conditioned image generation, ByT5, SD3, and full omni serving remain deferred until their contracts are validated separately.
- Semantic-conditioning failure is not silently downgraded to text-only or random embeddings. Any degraded fallback must be explicit and observable.

## Initial Implementation Boundary

1. Validate `modalities=["image"]` and reject unsupported combinations.
2. Normalize `image_generation` parameters.
3. Add typed image-generation contracts.
4. Add an `image` pipeline profile with an `img_gen` terminal.
5. Append 256 image query tokens in preprocessing and record their span.
6. Capture thinker hidden states with enough metadata to align full or suffix captures.
7. Protect image query tokens from radix cache aliasing with per-request identity.
8. Project query hidden states into image conditioning.
9. Call a small image backend interface and serialize one PNG image.
10. Add CI-safe unit and mocked integration tests.
11. Add an example request and challenge write-up material.

### System Overview

```mermaid
flowchart LR
    A["/v1/chat/completions"] --> B["Request Normalizer"]
    B --> C["Ming Preprocessor"]
    C --> D["Thinker Prefill-Only Runtime"]
    D --> E["HiddenStateCapture"]
    E --> F["SemanticConditioner"]
    F --> G["ImageGenExecutor"]
    G --> H["DiffusionBackend"]
    H --> I["Image Response Adapter"]
    I --> J["message.images"]

    B -. image params .-> G
    C -. query span + gen_mask .-> E
    C -. prompt text .-> G
```

### Layer Responsibilities

| Layer | Responsibility | Must not do |
|---|---|---|
| Serving | Parse user request, normalize image intent, validate parameters | Infer semantic-conditioning behavior |
| Preprocessor | Append image query token span, record query span and generation mask | Run diffusion or inspect hidden states |
| Thinker runtime | Run prefill-only forward and capture hidden states | Know image-backend specifics |
| Hidden-state adapter | Align captured hidden states to query span and validate shape | Silently drop semantic conditioning |
| Semantic conditioner | Project `[B, N, 4096]` to `[B, N, 2560]` | Load a full second LLM in production path |
| Image executor | Own terminal lifecycle, call backend, serialize image payload | Mutate request routing decisions |
| Image backend | Generate image from prompt embeddings and params | Know OpenAI response schema |
| API adapter | Convert terminal payload to public response schema | Reinterpret model/runtime failures |

---

## Public API Contract

Image output is requested by modality:

```json
{
  "model": "ming-omni",
  "messages": [
    {"role": "user", "content": "A cat sitting by a window at sunset"}
  ],
  "modalities": ["image"],
  "image_generation": {
    "size": "1024x1024",
    "num_inference_steps": 28,
    "guidance_scale": 2.0,
    "seed": 42,
    "negative_prompt": "blurry"
  }
}
```

`modalities` answers: "What outputs does the user want?"  
`image_generation` answers: "How should the image output be generated?"

If `modalities` does not include `image`, the request must not route to the image terminal even if `image_generation` is present. The server should either ignore image params with a warning or reject the request as invalid. The recommended default is validation failure, because it catches client mistakes early.

The first PR formally supports non-streaming image-only responses. Requests that combine `"text"` and `"image"` output should be rejected until multi-terminal response merge is validated. Streaming requests that include `modalities=["image"]` should be rejected until image terminal events have a reviewed streaming schema.

Non-streaming response:

```json
{
  "choices": [
    {
      "message": {
        "role": "assistant",
        "content": null,
        "images": [
          {
            "b64_json": "iVBORw0KGgo...",
            "format": "png",
            "width": 1024,
            "height": 1024,
            "status": "semantic_conditioned"
          }
        ]
      },
      "finish_reason": "stop"
    }
  ]
}
```

Future streaming support should use terminal or near-terminal image events rather than token-like incremental deltas. That schema is intentionally outside the first production scope.

---

### ImageGenerationParams

The serving layer normalizes raw request params into one internal object:

| Field | Type | Default | Notes |
|---|---:|---:|---|
| `width` | int | 1024 | Parsed from `size` or explicit width |
| `height` | int | 1024 | Parsed from `size` or explicit height |
| `num_inference_steps` | int | 28 | ZImage default should be documented |
| `guidance_scale` | float | 2.0 | Keep initial defaults explicit and backend-tunable |
| `seed` | int \| null | null | Used for deterministic debug runs |
| `negative_prompt` | string | `""` | Stored even if not fully supported by semantic negative embeddings |
| `allow_degraded_fallback` | bool | false | Explicit opt-in only |

### ImageGenerationContext

The preprocessor attaches image-generation context to pipeline state:

```python
ImageGenerationContext(
    request_id: str,
    params: ImageGenerationParams,
    query_span_start: int,
    query_span_end: int,
    gen_mask: list[bool],
    query_tokens: Tensor,      # [256, 4096]
    prompt_text: str,
    cache_namespace: str,
)
```

The explicit `query_span_start` / `query_span_end` removes ambiguity. `gen_mask` remains useful, but the span gives alignment code a simpler invariant:

```text
query_span_end - query_span_start == num_query_tokens
```

### CapturedHiddenStates

The thinker runtime should not pass raw tensors without metadata. It should return:

```python
CapturedHiddenStates(
    tensor: Tensor,            # [S, H] or [B, S, H]
    full_sequence_length: int,
    captured_sequence_length: int,
    captured_is_suffix: bool,
    cache_prefix_length: int,
    hidden_size: int,
)
```

This makes cache behavior part of the contract instead of an implicit runtime detail.

### SemanticConditioningBundle

After alignment:

```python
SemanticConditioningBundle(
    positive: list[Tensor],    # each [N, 2560]
    negative: list[Tensor],    # each [N, 2560]
    source: Literal["thinker_hidden_states"],
    prompt_text: str,
    render_text: str | None,
)
```

The image backend consumes this bundle. It should not reach back into pipeline state.

---

### Decision 1: `modalities=["image"]` is the routing switch

`image_generation` is a parameter object, not a routing signal. This mirrors the rest of multi-modal serving: output modalities decide terminal stages; modality-specific config controls generation.

Alternative: route to image terminal whenever `image_generation` exists.  
Rejected because it makes malformed requests easy to accept and creates surprising behavior for clients that pass shared config objects.

### Decision 2: Production path reuses thinker hidden states

The production path uses the already-loaded Ming thinker and 256 image query tokens. The standalone semantic encoder path is useful for debugging or offline experiments, but it should not be the normal serving path because it duplicates the largest model component.

Alternative: keep a separate `MingSemanticEncoder` in the image backend.  
Rejected as the default because it increases memory pressure and creates two independent LLM load paths.

### Decision 3: Query-token cache safety is mandatory

Image query token positions must not be accidentally served from a prior request's radix prefix cache. The design uses two protections:

1. The query region receives per-request cache identity, so the query suffix is recomputed.
2. The hidden-state adapter still tail-aligns and validates `gen_mask` for suffix-only captures.

If captured hidden states do not cover all query positions, semantic conditioning fails loudly.

### Decision 4: No silent degraded fallback by default

If semantic conditioning fails, the default behavior is an explicit error or an image response with `status="failed_semantic_conditioning"` and no image payload. Any text-only or backend fallback is allowed only when `allow_degraded_fallback=true`.

Alternative: always fall back so the API returns an image.  
Rejected because it makes correctness failures look like successful generation.

### Decision 5: Text-rendering enhancement is deferred

The first PR does not implement ByT5 render-text enhancement. The initial semantic path is thinker-hidden-state conditioning only. Text-rendering control is deferred because it adds another model load path and a second conditioning contract.

### Decision 6: Image backend remains small and replaceable

The initial implementation keeps image generation behind a small interface:

```python
class ImageGenerationBackend:
    def load(self, model_path: str, device: torch.device) -> None: ...
    def generate(
        self,
        conditioning: SemanticConditioningBundle,
        params: ImageGenerationParams,
    ) -> PIL.Image.Image: ...
```

### Decision 7: Executor Lifecycle

Diffusion model loading and denoising are blocking GPU-heavy operations, so the image executor must not run them directly on the pipeline event loop. The executor owns backend load/unload, async offloading, request aborts, bounded in-flight work, PNG serialization, and structured stage errors. The backend stays model-specific: it denoises and decodes images, but does not know the public API response schema.

The executor owns lifecycle and serialization. The backend owns model-specific generation. The first PR should wire one backend path well; SD3 and backend comparison work belong in deferred scope.

---

### Diffusion Quality And CFG Safety
The image backend contract also needs a small quality-safety rule: a valid PNG is not enough.

ZImage generation is sensitive to classifier-free guidance settings. Prior design notes call out that high guidance can saturate certain prompts, especially Chinese ink-painting prompts, into near-white images. For the initial PR, `guidance_scale` should default conservatively and remain explicitly configurable. If `cfg_normalization` is available in the backend, it can be exposed later as a backend-specific tuning parameter.

Smoke tests should record pixel mean/std and treat blank-looking images as failed smoke evidence, not successful generation.

---

### Image Pipeline

```text
preprocessing
  -> mm_aggregate
  -> thinker
  -> img_gen terminal
```

For the first text-to-image path, requests with input audio, image, or video should be rejected while `modalities=["image"]` is active. This avoids mixing ordinary media placeholder embeddings with image-generation query tokens before their ordering contract is validated. The image terminal is the only required terminal for image-only requests.

The existing text pipeline remains responsible for text-only requests. Text+image co-output is deferred because it requires multi-terminal response merge validation.

### Deferred Full Omni Pipeline

The full omni pipeline is not part of the initial PR. The first production path is the `image` profile with image-only output. Text+image co-output and the `full` profile should remain unavailable unless a later PR validates multiple terminal outputs under realistic GPU placement. See the deferred scope file for the future shape.

---

### Normal Case

```text
full sequence:
  [text tokens][image_start][256 query patch tokens][image_end]

gen_mask:
  false ... false true x256 false

captured hidden states:
  [full_seq_len, 4096]

query_hidden:
  hidden_states[gen_mask] -> [256, 4096]
```

### Radix Cache / Suffix Capture Case

When prefix cache is active, captured hidden states may contain only the newly computed suffix:

```text
full sequence length:      900
captured sequence length:  258
cache prefix length:       642
```

Alignment rule:

```text
if captured_is_suffix:
    visible_mask = gen_mask[-captured_sequence_length:]
    require visible_mask.sum() == gen_mask.sum()
```

If the visible suffix does not contain all query positions, semantic conditioning must fail loudly. This prevents partial query conditioning from producing plausible but wrong images.

### Chunked Prefill

The initial implementation does not implement cross-chunk accumulation. Chunked prefill is accepted only if the final captured tensor contains the complete query span. If a chunk boundary lands inside the query span and earlier query hidden states are unavailable, the executor must report a structured failure rather than falling back silently. Accumulating query hidden states across chunks is deferred.

---

## Error Handling and Fallback Policy

| Failure | Default behavior | Optional degraded behavior |
|---|---|---|
| Missing image params but image requested | Use defaults | Same |
| Invalid image size | 400 validation error | None |
| Missing hidden states | Structured image-generation error | Text/backend fallback only if explicitly enabled |
| Mask/hidden-state mismatch | Structured image-generation error | None unless all query positions can be tail-aligned |
| Conditioner load failure | Startup failure for image pipeline | Disable image endpoint if configured as optional |
| Image backend OOM | 500 with clear stage error | None |
| Image serialization failure | 500 with image stage error | None |

Every degraded response must include:

```json
{
  "status": "degraded",
  "reason": "semantic_conditioning_unavailable"
}
```

and emit a warning log with request id and stage name.

---

## Observability

Image generation should log structured events:

| Event | Fields |
|---|---|
| `image_request_normalized` | request id, modalities, size, steps, guidance |
| `image_query_span_created` | request id, span start/end, query count |
| `hidden_states_captured` | request id, full length, captured length, cache prefix length |
| `semantic_conditioning_projected` | request id, input shape, output shape, norm mean/std |
| `diffusion_started` | request id, backend, size, steps, guidance, seed |
| `diffusion_finished` | request id, elapsed seconds, width, height, pixel mean/std |
| `image_response_serialized` | request id, format, byte length |
| `image_generation_degraded` | request id, reason |

For the challenge, these logs double as active communication material. Debugging notes should quote the same stage names and event names used in code.

---

### CI-Safe Unit Tests

- `ImageGenerationParams` parses `size`, `width`, `height`, seed, guidance, negative prompt.
- Request validation requires `modalities` to include `image` before routing image terminal.
- Preprocessor appends exactly one image query span.
- `query_span_end - query_span_start == num_query_tokens`.
- Hidden-state adapter handles:
  - full-length hidden states
  - suffix hidden states
  - 2D and 3D tensors
  - mask shorter than hidden states
  - suffix missing query positions
- Semantic conditioner preserves expected shapes and normalization.
- Executor serializes PIL images into PNG base64 with width/height metadata.
- Degraded fallback requires explicit opt-in.

### Mocked Integration Tests

- Request -> preprocessor -> fake thinker hidden states -> conditioner stub -> executor -> image payload.
- Image-profile terminal resolver waits only for `img_gen` on image-only requests.
- Text+image requests are rejected with a validation error.
- Media-conditioned image requests are rejected with a validation error.
- Streaming image requests are rejected with a validation error.
- Abort propagates to image terminal while diffusion generation is pending.
- Error in image terminal does not corrupt text terminal response.

### Runtime / Cache-Focused Tests

- Simulate radix prefix-cache suffix capture and verify tail alignment.
- Simulate query-region cache aliasing and verify per-request cache identity prevents reuse.
- Simulate chunked prefill where the query span is split and verify structured failure.

These can start as controlled mocks, but at least one manual test should exercise the real runtime cache path.

### Manual GPU Smoke

Manual GPU smoke is valuable challenge evidence if hardware is available, but it is not a merge blocker:

```text
thinker + image backend on available GPU placement
prompts:
  - English semantic prompt
assertions:
  - image is 1024x1024
  - pixel std > blank threshold
  - condition embeds shape [1,256,2560]
  - no degraded fallback event
  - logs include image_request_normalized, image_query_span_created, hidden_states_captured, image_response_serialized
```

## Pre-Implementation Spike Plan

The spike happens after this RFC and before the main implementation. It is not meant to build the feature. It is meant to cheaply verify the assumptions that would be expensive to discover halfway through the PR. A cheap check can be source reading, `rg`, a small script, a mocked unit test, or a short local smoke run.

Each spike note should use this format:

```text
Question:
Method:
Result:
Decision:
```

Recommended checks:

| Question | Cheap method | Decision if confirmed | Decision if rejected |
|---|---|---|---|
| Does request routing have reliable access to `modalities` before pipeline selection? | Inspect serving request normalization and add a tiny validation test | Use `modalities=["image"]` as the only routing switch | Add a normalization step before pipeline selection |
| Does `extra_key` or equivalent metadata enter the radix cache key? | Read cache-key construction and run a minimal two-request cache test | Use existing cache identity for query-token safety | Add per-request query cache namespace |
| Can the runtime expose full or suffix hidden-state metadata? | Inspect capture mode support and run a small hidden-state capture probe if possible | Use `CapturedHiddenStates` metadata directly | Wrap the tensor with explicit inferred metadata and reject ambiguous captures |
| Can a real `diffusers` reference path run locally? | Check import, model path, seed control, and one minimal reference call | Add golden reference comparison to validation evidence | Use mocked integration plus manual GPU smoke, and state the limitation in the PDF |
| Can the pre-refactor V0 baseline still be measured? | Try the known V0 checkout or locate recorded baseline numbers | Report the 30%-50% speed target against measured V0 | Report known V0 numbers with caveat and use `8177fe9f` as the measured no-regression baseline |
| Is the backend available on the expected hardware budget? | Dry-load or import the backend and record memory placement expectations | Run manual GPU smoke if hardware is available | Keep CI on fake backend and document missing GPU evidence clearly |

The spike is done when every question above has a result and a decision. These notes should be kept because they become the "challenges and debugging" material for the final PDF.

## Validation and Performance Gates

The PR uses three gates. They answer different questions and should not be blurred together.

### Smoke Gate

Smoke answers: "Is the line connected?"

One fixed prompt should go through the intended image path:

```text
HTTP request
  -> image route selected
  -> image query span created
  -> hidden states captured
  -> conditioning bundle produced
  -> image backend called
  -> PNG response returned
```

Smoke passes when the request returns one valid PNG with expected dimensions, the pixel standard deviation is above a blank-image threshold, required image-generation logs are present, and no degraded fallback event occurs. Smoke does not prove visual quality or reference correctness.

### E2E Correctness Gate

E2E answers: "Does the implementation behave like the intended design, not just a happy-path mock?"

The strongest available E2E evidence should include:

- an HTTP request through the actual server, not only direct executor calls
- a response containing image bytes and image metadata
- logs showing the image terminal was reached
- no silent semantic-conditioning fallback
- regression coverage for cache aliasing, suffix hidden-state alignment, missing conditioning, unsupported streaming, unsupported text+image co-output, and unsupported media-conditioned image generation

If a real reference path is available, add reference comparison using the same weights, seed, scheduler settings, and image size where possible:

- `condition_embeds` cosine similarity above `0.999` after L2 normalization
- early latent slices within documented bf16 tolerance
- final image difference within a documented tolerance
- golden slices captured from the external reference path, not from this implementation comparing with itself

If the reference path is not available in the current environment, the PDF must say that directly and separate what was proven by mocked integration, manual smoke, and code review.

### Performance Gate

Performance answers: "Did the redesign preserve or improve the speed story, and are the numbers honest?"

There are two baselines:

- **V0 pre-refactor baseline:** target 30%-50% faster end-to-end image generation if the old path can still be measured. If it cannot be run on the current environment, use recorded V0 numbers only with a caveat.
- **Teaching baseline `8177fe9f`:** no-regression baseline for the redesigned PR. This is the measured fallback if V0 cannot be reproduced.

Measure the following when hardware is available:

- image-generation latency, broken down into prefill, conditioning, denoise, and VAE decode
- mixed-traffic thinker decode tok/s if the implementation changes hidden-state capture cost for non-image requests
- concurrent image wall-clock only if executor concurrency lands in the PR

Performance passes when the comparison is either measured or honestly bounded, with commit, hardware, seed, batch size, prompt, image size, and method recorded. Missing performance evidence should be called out directly in the PDF, with a clear explanation of what evidence remains.

---

### Section 1: RFC and Interfaces

- Add this RFC.
- Add typed request/conditioning/result interfaces.
- Add docstrings that define tensor shapes and ownership.

Exit gate: reviewers agree on routing, hidden-state contract, and fallback policy.

### Section 2: Serving and Client Contract

- Normalize image-generation params.
- Validate `modalities`.
- Add image result schema for non-streaming responses.
- Reject image streaming requests until a separate streaming schema is reviewed.
- Add client result types and base64 handling.

Exit gate: schema tests pass; invalid modality/param combinations fail predictably.

### Section 3: Image Pipeline Routing

- Add `img_gen` terminal stage.
- Add image-only pipeline variant as the primary supported topology.
- Route image-only requests only to the image terminal.
- Reject text+image co-output and media-conditioned image requests.
- Validate image terminal GPU placement does not collide with thinker placement in the default profile.

Exit gate: config/routing tests pass; text-only requests do not enter image stage.

### Section 4: Preprocessor and Hidden-State Contract

- Append image query tokens.
- Record query span and generation mask.
- Attach per-request cache identity for query region.
- Capture hidden states with metadata.

Exit gate: unit and mocked runtime tests cover full and suffix captures.

### Section 5: Semantic Conditioning and Image Executor

- Load lightweight conditioner.
- Project query hidden states.
- Enforce strict fallback policy.
- Call backend and serialize image payload.

Exit gate: mock executor tests pass; shape checks are enforced.

### Section 6: Backend Adapter and Example

- Add one backend adapter behind the `ImageGenerationBackend` interface.
- Keep a fake/smoke backend available for CI-safe tests.
- Add an example server or request script for the image profile.

Exit gate: backend import and mocked image serialization tests pass.

---

## Risks and Mitigations

| Severity | Risk | Mitigation |
|---|---|---|
| P0 | Image response succeeds while semantic conditioning failed | Strict fallback policy; `status` metadata; warning logs |
| P0 | Query hidden states are misaligned under radix cache | CapturedHiddenStates metadata; tail alignment; per-request query cache identity |
| P0 | Chunked prefill loses query hidden states | Fail loudly in the initial PR; accumulation is deferred |
| P1 | Duplicate LLM load causes OOM | Production path uses existing thinker + lightweight conditioner |
| P1 | Backend-specific defaults hurt output quality | Initial params are explicit; backend tuning is deferred |
| P1 | API users confuse params with modality routing | Validate `modalities`; document request examples |
| P1 | Tests only prove mocks | Manual GPU smoke is documented if available; PDF states any missing evidence plainly |
| P2 | ByT5 text rendering expectations exceed capability | Defer ByT5 out of initial scope |
| P2 | Full omni profile hides resource or merge bugs | Keep full omni out of initial scope |
| P2 | Future image streaming stresses client memory | Defer streaming schema |

---

## Acceptance Criteria

The PR is accepted only when:

- Text-only Ming requests still work.
- Requests without `modalities=["image"]` do not enter image generation.
- Image-generation requests with input audio, images, or video return explicit validation errors.
- Image-only requests generate image terminal results through the pipeline, not side effects.
- Hidden-state capture is validated for both full and suffix captures.
- Semantic conditioning produces `[B,256,2560]` embeddings.
- Degraded fallback cannot happen silently.
- Non-streaming API exposes image data consistently.
- Text+image co-output returns an explicit validation error.
- Streaming image requests return an explicit validation error.
- Full omni image generation is not exposed in the first PR.
- Pre-implementation spike decisions are recorded, including cache-key behavior, hidden-state capture metadata, reference availability, and baseline availability.
- Unit and mocked integration tests pass in ordinary CI.
- Manual GPU smoke evidence is attached if available; otherwise the PDF states it was not run.
- Performance is measured or honestly bounded against V0 if available, otherwise against the teaching baseline `8177fe9f`.
- The PDF write-up explains design, implementation, debugging, tests, and remaining limitations.
- The PR description or comments show active progress/debugging communication.

---

## Design Defaults

The initial PR uses these defaults unless implementation evidence forces a change:

- Requests that pass `image_generation` without `modalities=["image"]` fail validation instead of being silently ignored.
- Backend or text-only fallback is never automatic. It requires `allow_degraded_fallback=true`.
- A standalone semantic encoder does not ship in the initial PR; the production path reuses thinker hidden states.
- Image results include `status` metadata so semantic conditioning, degraded behavior, and failures are visible.
- Manual GPU E2E is useful evidence but not a merge blocker. If it is not available, the PDF states that clearly and separates CI-safe evidence from manual evidence.

edwingao28:
## Motivation

Closes #876. Ming-flash-omni-2.0 emits image-conditioning tokens, but SGLang-Omni had no stage to turn them into images: `/v1/chat/completions` requests with `modalities: ["image"]` were unsupported and there was no image DiT in the pipeline. This PR adds end-to-end image generation — the thinker's gen-mask hidden states condition a Z-Image-Turbo DiT — served through the existing OpenAI-compatible chat endpoint.

## Modifications

Disaggregated two-stage pipeline (`MingOmniImagePipelineConfig`): a tensor-parallel thinker feeds a dedicated DiT GPU. Launched via `examples/run_ming_omni_image_server.py`, separate from the generic `sgl-omni serve --model-path` path.

**Conditioning bridge — thinker-fused (default `semantic_source="thinker"`):**
- `diffusion/semantic_conditioner.py` — `SemanticConditioner.project()`: `proj_in` Linear(4096→1536) → Qwen2 connector (28 layers, attention forced non-causal: `is_causal=False`) → `proj_out` Linear(1536→2560) → L2-norm → `[B, 256, 2560]` condition embeds. Loads only the connector + proj + `query_tokens`; does **not** reload the ~97B BailingMoeV2 (the SGLang thinker engine already ran that forward).
- `components/preprocessor.py` — `enable_image_gen` injects `query_tokens` + image start/patch/end markers; `model_runner/ming_thinker_model_runner.py` relays the gen-mask-selected hidden states to the conditioner.

**Image-gen terminal stage:**
- `components/image_gen_executor.py` — `_extract_params` parses the `image_generation` block (`size`/`width`/`height`, `num_inference_steps`, `guidance_scale`, `seed`, `negative_prompt`, `semantic_source`, `enable_text_rendering`), routes the semantic source, and concatenates ByT5 embeds when text rendering is enabled.
- `diffusion/zimage_backend.py` (default DiT) and `diffusion/sd3_backend.py` (alternate), behind `diffusion/backend.py` (`ImageGenParams`; defaults tuned for the distilled low-CFG Z-Image-Turbo: `guidance_scale=2.0`, `num_inference_steps=28`).
- `diffusion/byt5_encoder.py` — optional ByT5 glyph rendering of the last quoted span, gated by `--enable-byt5-text-rendering`.

**Optional standalone semantic path (gated, not the default):**
- `diffusion/semantic_encoder.py` + vendored `diffusion/bailing_moe_model.py` / `bailing_moe_config.py` — the reference `semantic_source="standalone"` route, loaded only with `--enable-standalone-semantic-encoder`; the supported/default route is the thinker-fused conditioner above.

**Public API contract:**
- `serve/openai_api.py` — `_validate_chat_image_generation_request` returns HTTP 400 when an `image_generation` block is present without `"image"` in `modalities`; image output is serialized into `message.images[]`.
- `serve/protocol.py` — `ChatCompletionImage` (`{id, data, format="png", width, height}`); `client/client.py` / `client/types.py` — image-generation request extraction.

**Config / GPU placement:**
- `config.py` — `MingOmniImagePipelineConfig` + `_validate_image_gen_gpu_not_in_thinker_tp_range` rejects DiT/thinker GPU overlap at launch; `--tp-size > 1` sets `disable_custom_all_reduce=True` for the thinker stage.

**Tests:** 11 new unit files under `tests/unit_test/ming_omni/` (executor, config, preprocessor, semantic_conditioner, zimage backend, client API, bootstrap, merge, query_info, request, diffusion imports), plus `tests/unit_test/test_sglang_output_processor.py` and `tests/test_model/test_production_image_gen_e2e.py` (real two-stage e2e).

## Accuracy Test

Image output is qualitative; verified by deterministic e2e generation (`seed: 42`) on H100, thinker TP4 (GPUs 0-3) + DiT on GPU 4. Two prompts each produce a valid 1024×1024 PNG in `choices[0].message.images[0].data`:
- `"A cat sitting on a windowsill watching the sunset"`
- `"一幅水墨画，画中有竹子和远山"`

Sample outputs are embedded in the cookbook PR #882. Ming-Omni unit suite, run in the pinned pre-commit env on the GPU box:

```
103 passed, 6 warnings in 5.38s
```

## Benchmark & Profiling

End-to-end on H100, warm pipeline, 1024×1024 @ 28 inference steps, thinker TP4 + one dedicated DiT GPU:

| Stage | Mean latency |
|---|---:|
| Total e2e (warm) | `~14 s / image` |
| DiT denoise (Z-Image-Turbo, Stage 2) | `~9.5 s` |

DiT denoise dominates the warm cost. Directional warm-pipeline numbers, not a release-wide guarantee. `guidance_scale=2.0` is the tuned operating point for the distilled, low-CFG Z-Image-Turbo DiT; SD-style 7-9 washes the image out.

## Checklist

- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed. (cookbook: #882)


edwingao28:
## Production E2E reference run — current PR (`feat/ming-image-generation-rewrite`)

Same harness and two prompts as the #336 reference run, re-run on this PR's head.

| | |
|---|---|
| Commit | `6db711e2` |
| Hardware | H100 80GB — thinker TP=4 (GPUs 0-3) + Z-Image-Turbo DiT (GPU 4), `mem_fraction_static=0.8` |
| Params | 1024×1024, 28 steps, `guidance_scale=2.0`, `seed=42`, `semantic_source=thinker`, `enable_text_rendering=false` |
| Result | 2/2 prompts → 1024×1024 PNG, `ok=True reason=stop`, **no fallback** |

**Targeted image-gen unit suite** (`tests/unit_test/ming_omni/`):

```
103 passed, 6 warnings in 5.54s
```

**Production E2E** (`tests/test_model/test_production_image_gen_e2e.py`):

```
=== START commit=6db711e2dd2bb770949e56e8d168ce136806f0cc MEM_FRACTION_STATIC=0.8 ===
=== Phase A: MingPreprocessor create + run (enable_image_gen=True) ===
[E2E_TIMING] preprocessor_create 4.703s
[E2E_TIMING] preprocess[0] 0.050s  input_ids=[1, 294] gen_mask_sum=256 prefill_only=True
[E2E_TIMING] preprocess[1] 0.036s  input_ids=[1, 292] gen_mask_sum=256 prefill_only=True
=== Phase B: building image pipeline (thinker TP4 gpu0-3, image_gen gpu4) ===
Load weight begin. avail mem=74.03 GB
Load weight begin. avail mem=77.71 GB
Load weight begin. avail mem=77.71 GB
Load weight begin. avail mem=77.71 GB
[E2E_TIMING] pipeline_startup 51.8s (thinker TP4 load + image_gen stage load: SemanticConditioner + ZImage)
=== Phase C: submitting 2 image-gen requests ===
[SemanticConditioner] Scales: [16], total tokens: 256
[SemanticConditioner] Projections loaded: proj_in=[1536, 4096], proj_out=[2560, 1536], query_tokens=[256, 4096] (256 tokens)
[SemanticConditioner] All components loaded on cuda:4 (~2.9 GB)
[E2E_TIMING] request[0] cold e2e 24.54s  ok=True reason=stop -> prod_0.png
[E2E_TIMING] request[1] warm e2e 14.05s  ok=True reason=stop -> prod_1.png
=== DONE rc=0 ===
```

### Output images (seed 42)

**Prompt 1 — "A cat sitting on a windowsill watching the sunset"**

Ming-Omni image output: cat on a windowsill at sunset

**Prompt 2 — "一幅水墨画，画中有竹子和远山"**

Ming-Omni image output: ink-wash bamboo and distant mountains

Peter-Nie2003:
@edwingao28  Can i take the job wire **Ming image generation into `sgl-omni serve`**. Right now the image pipeline can only be launched via `examples/run_ming_omni_image_server.py` — the real CLI has zero image/diffusion flags. Posting the plan first to align before I start:

### 1. New flags (naming aligned with sglang-diffusion)

| New flag | Type | Default | Help text |
|---|---|---|---|
| `--image-gen` | bool | `False` | Use the image-generation pipeline (thinker + diffusion DiT). Mutually exclusive with `--text-only`/`--colocate`. |
| `--diffusion-model-path` | str | `None` (required with `--image-gen`) | Path or HF id of the diffusion model bundle (scheduler + VAE + transformer), e.g. a Z-Image directory. |
| `--diffusion-backend` | `zimage`\|`sd3` | `zimage` | Diffusion backend implementation for image generation. |
| `--image-gen-gpu` | int | `None` → config default (gpu 1) | GPU id for the diffusion (image-gen) stage; must not overlap the thinker TP range. |
| `--enable-standalone-semantic-encoder` | bool | `False` | Load the standalone Ming semantic encoder so requests can use `semantic_source=standalone`. |
| `--enable-byt5-text-rendering` | bool | `False` | Load the ByT5 text encoder for rendering quoted text into generated images. |

### 2. Behavior / validation
 - --image-gen selects variant="image" in the config-resolution block (next to the existing --text-only branch).
- --image-gen requires --diffusion-model-path → else typer.BadParameter.
- --image-gen is mutually exclusive with --text-only and --colocate (same style as the existing --colocate/--text-only guard).
- The five image-only flags raise _raise_unsupported_flag(...) if passed without --image-gen (so a user pointing them at a speech pipeline gets a clear error, matching how --code2wav-gpu etc. behave on wrong pipelines).
- --image-gen-gpu flows through the existing _apply_stage_gpu_override(stage_name="image_gen") path, then re-runs _validate_image_gen_gpu_not_in_thinker_tp_range() for a friendly overlap error.
- The image pipeline already forces hidden-state capture (enable_return_hidden_states=True, disable_cuda_graph=True). So --thinker-cuda-graph on conflicts → validate and reject(fail loud) rather than silently mis-serve.

### 3. Scope
Server flags cover **loading + placement only**. `num_inference_steps` / `guidance_scale` / `size` / `negative_prompt` stay in the request body, keeping `ImageGenParams` as the single source of defaults rather than duplicating them.

edwingao28:
> @edwingao28 Can i take the job wire **Ming image generation into `sgl-omni serve`**. Right now the image pipeline can only be launched via `examples/run_ming_omni_image_server.py` — the real CLI has zero image/diffusion flags. Posting the plan first to align before I start:
> 
> ### 1. New flags (naming aligned with sglang-diffusion)
> New flag	Type	Default	Help text
> `--image-gen`	bool	`False`	Use the image-generation pipeline (thinker + diffusion DiT). Mutually exclusive with `--text-only`/`--colocate`.
> `--diffusion-model-path`	str	`None` (required with `--image-gen`)	Path or HF id of the diffusion model bundle (scheduler + VAE + transformer), e.g. a Z-Image directory.
> `--diffusion-backend`	`zimage`|`sd3`	`zimage`	Diffusion backend implementation for image generation.
> `--image-gen-gpu`	int	`None` → config default (gpu 1)	GPU id for the diffusion (image-gen) stage; must not overlap the thinker TP range.
> `--enable-standalone-semantic-encoder`	bool	`False`	Load the standalone Ming semantic encoder so requests can use `semantic_source=standalone`.
> `--enable-byt5-text-rendering`	bool	`False`	Load the ByT5 text encoder for rendering quoted text into generated images.
> ### 2. Behavior / validation
> * --image-gen selects variant="image" in the config-resolution block (next to the existing --text-only branch).
> * --image-gen requires --diffusion-model-path → else typer.BadParameter.
> * --image-gen is mutually exclusive with --text-only and --colocate (same style as the existing --colocate/--text-only guard).
> * The five image-only flags raise _raise_unsupported_flag(...) if passed without --image-gen (so a user pointing them at a speech pipeline gets a clear error, matching how --code2wav-gpu etc. behave on wrong pipelines).
> * --image-gen-gpu flows through the existing _apply_stage_gpu_override(stage_name="image_gen") path, then re-runs _validate_image_gen_gpu_not_in_thinker_tp_range() for a friendly overlap error.
> * The image pipeline already forces hidden-state capture (enable_return_hidden_states=True, disable_cuda_graph=True). So --thinker-cuda-graph on conflicts → validate and reject(fail loud) rather than silently mis-serve.
> 
> ### 3. Scope
> Server flags cover **loading + placement only**. `num_inference_steps` / `guidance_scale` / `size` / `negative_prompt` stay in the request body, keeping `ImageGenParams` as the single source of defaults rather than duplicating them.

LGTM! Feel free to stack another PR on top of my current one.

edwingao28:
# Update from 6.27


1. [feat] add `sgl-omni serve` image-gen flags (--image-gen / --diffusion-model-path /
   --image-gen-gpu / --enable-standalone-semantic-encoder / --enable-byt5-text-rendering) + tests
2. [fix] correct CLI override ordering, config-clobber, and add mutual-exclusion guards
   (--image-gen vs --text-only/--colocate; requires --diffusion-model-path)
3. [refactor] dedup + readability: drop dead code, extract _load_linear, dedupe guard msg / imports
4. [fix] StaticCache import in bailing_moe_model (was referenced unimported → NameError)
5. [refactor] remove SD3 fallback backend (+ factory/CLI/test cleanup, zero dangling refs)
6. [test] refactor unit tests: delete duplicates, add zimage/conditioner/byt5/CLI coverage,
   relocate test_sglang_output_processor; make production E2E script rigorous (fail-fast, real PNG asserts)

## CLI e2e validation 
  Ming-flash-omni-2.0 (thinker TP4 on GPU 0–3) + Z-Image-Turbo DiT on GPU 4.

  ### Results
  
  | Check | Result |
  |---|---| 
  | `tests/unit_test/` (full suite) | **1675 passed** |
  | GPU e2e — `--image-gen` (CLI flags, TP4) | **PASS** — 2 images, HTTP 200 |
  | GPU e2e — `--config` (YAML, no image flags) | **PASS** — 1 image, HTTP 200 |
  | GPU-overlap collision (`image_gen` vs thinker TP) | none (ordering fix) |
  
  ### Timing breakdown

  **Startup** (cold, weights in HF cache; relative to coordinator start):
  
  | phase | +sec |
  |---|---| 
  | stage process spawn (preprocessing → thinker ×4) | ~7s |
  | torch.distributed init (thinker TP4) | ~29s |
  | thinker weight load (BailingMoeV2, ~48 GB/rank) | ~49s |
  | Z-Image DiT backend load (5.8s) + semantic conditioner | ~69s |

  Server answers `/v1/models` at ~56–60s; the DiT stage finishes loading ~T+69s (lazy on `image_gen`), so it overlaps into the first request.

  **Per-request** (1024×1024, 28 steps, guidance 2.0, `semantic_source=thinker`, non-stream):

  | request | latency | note |
  |---|---|---|
  | 0 (cold) | **24.57s** | includes one-time DiT lazy-load (~10s) |
  | 1 (warm) | **14.09s** | steady-state per image |
  
  ### Logs
  
  `--image-gen` happy path:
  ```
  [e2e] launching server (thinker TP4 gpu0-3, image_gen gpu4) ...
  [e2e] server READY after ~60s
  [e2e] request[0] prompt='A cat sitting on a windowsill watching the sunset'
  [e2e] request[0] http=200  e2e=24.57s
  [e2e] request[0] IMAGE ok bytes=963761 fmt=png 1024x1024
  [e2e] request[1] prompt='一幅水墨画，画中有竹子和远山'
  [e2e] request[1] http=200  e2e=14.09s
  [e2e] request[1] IMAGE ok bytes=849349 fmt=png 1024x1024
  [e2e] RESULT: PASS
  ```
  
  `--config` path (YAML carries `dit_model_path`, no image CLI flags — exercises the clobber fix):
  ```
  [cfg-e2e] launching server via --config (NO image flags) ...
  [cfg-e2e] READY after ~58s
  [cfg-e2e] http=200
  [cfg-e2e] IMAGE ok bytes=1035760 1024x1024
  [cfg-e2e] RESULT: PASS
  ```

  All server processes torn down, GPUs returned to baseline after each run.

yuan-luo:
This Chinese painting has some accuracy difference, please double check it.

yuan-luo:
It should be something like this:
image

PR #883, opened 2026-06-25
[CI] Migrate Qwen3-Omni CI from H20 to 2×H100

yxs:
H20 (96 GB) is being decommissioned → Omni CI moves to 2×H100 (80 GB; GPU 6,7). The green H20 CI ran a colocated full-BF16 stack (router + 2 colocated replicas, ~70–90 GB in 96 GB). That ~90 GB video stack does not fit an 80 GB H100, so the migration is forced off the single colocated-BF16 layout.

Rather than force one topology on all stages, each stage runs the topology that fits its workload, and the 10 stages collectively exercise all five viable 2×H100 serving paths.

| | config | what it is | GPU layout |
|---|---|---|---|
| ① | FP8 colocated-DP2 | router + 2 full FP8 replicas (~30 GB each) | 1 replica/GPU, c16 split DP |
| ② | BF16 colocated-DP2 | router + 2 full BF16 replicas | 1 replica/GPU, c16 split DP |
| ③ | BF16 thinker-TP=2 | thinker sharded across both GPUs | thinker GPU0+1 / talker GPU1 |
| ④ | BF16 disaggregated | thinker and talker on separate GPUs | thinker GPU0 / talker GPU1 |
| ⑤ | FP8 thinker-TP=2 | FP8 thinker sharded, talker stacked | thinker GPU0+1 / talker GPU1 |

## Test

| # | stage | in→out | H20 config | **H100 config** | GPU0 / GPU1 | measured (H100 / ref) |
|---|---|---|---|---|---|---|
| 1 | thinker_length | text→text | colocated BF16 | ③ BF16 TP=2 | thinker GPU0+1 / talker idle | 3/3 passed (correctness) |
| 2 | tts | text→audio | colocated BF16 | ② BF16 colocated-DP2 | 2 BF16 replicas | serves (WER needs whisper) |
| 3 | mmmu | image→text | colocated BF16 | ① FP8 colocated-DP2 | 2 FP8 replicas | passed (H20 lat 10.2) |
| 4 | mmmu-talker | image→audio | colocated BF16 | ④ BF16 disaggregated | thinker GPU0 / talker GPU1 | 29.6 s (H20 16.9) |
| 5 | mmsu | audio→text | colocated BF16 | ② BF16 colocated-DP2 (thinker-only 0.92) | 2 thinker replicas | passed (H20 0.35) |
| 6 | mmsu-talker | audio→audio | colocated BF16 | ⑤ FP8 TP=2 | thinker GPU0+1 / talker GPU1 | 40/40 served, rtf 0.74 (H20 8.2) |
| 7 | videomme | video→text | colocated BF16 | ④ BF16 disaggregated | thinker GPU0 / talker idle | 11.75 s (H20 13.5) |
| 8 | videomme-talker | video→audio | colocated BF16 | ④ BF16 disaggregated | thinker GPU0 68 / talker GPU1 42 | 12.0 s (H20 17.6) |
| 9 | videoamme | video→text | colocated BF16 | ① FP8 colocated-DP2 | 2 FP8 replicas | acc .68 (H20 12.2) |
| 10 | videoamme-talker | video→audio | FP8 TP=2 | ⑤ FP8 TP=2 | thinker GPU0+1 / talker GPU1 78 | 184 s (H20 162) |

Coverage: ①×2 (3,9), ②×2 (2,5), ③×1 (1), ④×3 (4,7,8), ⑤×2 (6,10).


tests/test_model/conftest.py
```diff
@@ -46,140 +46,156 @@
...
 @pytest.fixture(scope="module")
 def qwen3_omni_fp8_talker_server_tp2(tmp_path_factory: pytest.TempPathFactory):
-    """Start Qwen3-Omni FP8 with TP=2 thinker and TP=1 talker."""
-    yield from _start_qwen3_omni_speech_server(
-        tmp_path_factory,
+    """FP8 colocated 2-replica Qwen3-Omni router used by the Video-AMME talker
+    stage. (The ``_tp2`` suffix is historical: this stage shipped as an FP8 TP=2
+    replica, but a single TP=2 replica OOMs the talker once concurrency stacks it
+    with thinker-rank-1 on one card, so it now uses the same FP8 colocated router
+    as the other video stages.) See _start_qwen3_omni_fp8_colocated_router."""
```
Ratish1: nit: trim comment

tests/test_model/conftest.py
```diff
@@ -46,140 +46,156 @@
...
+
+def _start_qwen3_omni_fp8_colocated_router(tmp_path_factory: pytest.TempPathFactory):
+    """Start two FP8 colocated Qwen3-Omni replicas (one full ~30 GB FP8 model per
+    H100) behind the managed router -- the green H20 CI topology (router + 2
+    colocated replicas) with BF16 swapped for FP8 so each replica fits one 80 GB
+    card. The router spreads request concurrency across the two data-parallel
+    replicas, so each card carries only half the in-flight video-prefill
+    activation (the binding constraint). Used by every video stage (videomme /
+    videoamme, text + talker); validated no-OOM at the H20 concurrency (16) with
+    ~20 GB headroom per card (peak ~61/80 GB on both cards)."""
```
Ratish1: same

tests/test_model/conftest.py
```diff
@@ -46,140 +46,156 @@
...
+# Base thinker context for the Qwen3-Omni servers.
+QWEN3_OMNI_TP2_THINKER_MAX_SEQ_LEN = 32768
+# FP8 colocated 2-replica topology for the VIDEO stages. Mirrors the green H20 CI
+# (router + 2 colocated full replicas); the only change is BF16->FP8 so each full
+# ~30 GB FP8 replica fits one 80 GB H100. The router spreads concurrency across the
+# two data-parallel replicas, so each card carries only half the in-flight
+# video-prefill activation. A single TP=2 replica cannot do this: TP shards weights
+# (not activation), so the replicated video-prefill activation plus a talker +
+# thinker-rank-1 stacked on one card OOMs once concurrency climbs (empirically
+# GPU 1 reaches ~76 GB by concurrency 2, OOMs well before concurrency 10).
```
Ratish1: same

tests/test_model/conftest.py
```diff
@@ -46,140 +46,156 @@
...
 def qwen3_omni_fp8_talker_server_tp2(tmp_path_factory: pytest.TempPathFactory):
-    """Start Qwen3-Omni FP8 with TP=2 thinker and TP=1 talker."""
-    yield from _start_qwen3_omni_speech_server(
-        tmp_path_factory,
+    """FP8 colocated 2-replica Qwen3-Omni router used by the Video-AMME talker
+    stage. (The ``_tp2`` suffix is historical: this stage shipped as an FP8 TP=2
+    replica, but a single TP=2 replica OOMs the talker once concurrency stacks it
+    with thinker-rank-1 on one card, so it now uses the same FP8 colocated router
+    as the other video stages.) See _start_qwen3_omni_fp8_colocated_router."""
+    yield from _start_qwen3_omni_fp8_colocated_router(tmp_path_factory)
```
Ratish1: nit: This fixture now launches the FP8 colocated router topology, but the fixture name still says `_tp2` and Stage 11 still runs the TP=2 test file.

we can either keep this fixture as a real FP8 TP=2 server, or rename/update the fixture, Stage 11 job labels, and test assertions to validate FP8 colocated router DP instead of TP=2.

examples/configs/qwen3_omni_colocated_h100_fp8.yaml
```diff
@@ -0,0 +1,49 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# Colocated Qwen3-Omni launcher config for the 2x H100 (80 GB) CI video stages.
+#
+# Mirrors examples/configs/qwen3_omni_colocated_h20.yaml (the green single-H20 CI
+# profile), with two changes forced by the 80 GB card:
+#   1. model_path -> the FP8 thinker checkpoint. The BF16 30B thinker (~57 GB
+#      weights) does not fit one 80 GB card colocated; the FP8 thinker (~30 GB)
+#      does, with room for KV + per-video prefill activation.
+#   2. thinker fraction 0.75 -> 0.55. FP8 halves the thinker weights, so a smaller
```
Ratish1: should we clarify the memory semantics in this comment and distinguish this file from the existing generic FP8 colocated config.

There are now two FP8 colocated configs with different thinker budgets:

- `examples/configs/qwen3_omni_fp8_colocated.yaml`: thinker `total_gpu_memory_fraction: 0.75`
- `examples/configs/qwen3_omni_colocated_h100_fp8.yaml`: thinker `total_gpu_memory_fraction: 0.55`

This new file is being used as the H100 video-CI router profile, so should the comment should say that explicitly?. It should also mention that Qwen3-Omni subtracts the default `encoder_mem_reserve=0.05`, so the effective SGLang/KV thinker budget is `0.50`, not `0.55`.

yxs:
follow up

1. Add `run-ci` → run the 11-stage Omni CI on H100
2. Stand up an H100 `sglang-omni-ci-calibration` container + add `sglang-h100-ci.yaml` to the `tune-ci-thresholds` skill.
3. `[CI] Calibration` — worst-of-N on the H100 container → recalibrated perf/accuracy thresholds.

.github/workflows/test-qwen3-omni-ci.yaml
```diff
@@ -422,22 +443,23 @@ jobs:
...
         uses: ./.github/actions/omni-post-stage
         with:
-          stage-label: Video-AMME Talker (WER + speed)
+          stage-label: Video-AMME Talker (FP8 colocated DP)
           artifact-path-globs: |
-            */videoamme_audio*/videoamme_results.json
+            */videoamme/videoamme_results.json
           artifact-upload-name: qwen3-omni-videoamme-talker-results
-          artifact-upload-path: /tmp/**/videoamme_audio*/videoamme_results.json
+          artifact-upload-path: /tmp/**/videoamme/videoamme_results.json
```
Ratish1: this changes Stage 10 artifact collection from `videoamme_audio*` to `videoamme`, but `test_qwen3_omni_videoamme_talker_ci.py` still writes `videoamme_results.json` under a `videoamme_audio*` temp directory.

lets restore both the printed glob and upload path to `*/videoamme_audio*/videoamme_results.json`, otherwise the Stage 10 test can pass while its result artifact is not printed or uploaded for calibration.

zhaochenyang20:
Pretty reasonable to me. Change 7 to 7. Video MME：FP 8 TP 2；

examples/configs/qwen3_omni_colocated_h100_bf16.yaml
zhaochenyang20, line 1: looks reasonable

examples/configs/qwen3_omni_colocated_h100_fp8.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+  audio_encoder:
+    runtime:
+      resources:
+        total_gpu_memory_fraction: 0.025
+
+  thinker:
+    runtime:
+      resources:
+        total_gpu_memory_fraction: 0.55  # effective 0.50 after default encoder_mem_reserve 0.05
```
zhaochenyang20: use `#note (xuesong):` for comments

tests/test_model/conftest.py
```diff
@@ -46,140 +46,222 @@
...
+QWEN3_OMNI_BF16_COLOCATED_VIDEO_ARGS = (
+    f"--config {QWEN3_OMNI_BF16_COLOCATED_CONFIG} --colocate "
+    f"--stages.0.factory-args.thinker-max-seq-len {QWEN3_OMNI_TP2_THINKER_MAX_SEQ_LEN} "
+    f"--stages.4.factory-args.thinker-max-seq-len {QWEN3_OMNI_TP2_THINKER_MAX_SEQ_LEN}"
 )
+QWEN3_OMNI_BF16_THINKER_CONFIG = "examples/configs/qwen3_omni_mmmu_h100.yaml"
+QWEN3_OMNI_BF16_THINKER_ARGS = f"--config {QWEN3_OMNI_BF16_THINKER_CONFIG}"
+QWEN3_OMNI_DISAGG_THINKER_MEM_FRACTION = "0.82"
+QWEN3_OMNI_DISAGG_TALKER_MEM_FRACTION = "0.40"
+QWEN3_OMNI_FP8_TP2_THINKER_MEM_FRACTION = "0.40"
```
zhaochenyang20, line 69: shall we move all the config to the config file? No need to put it here.

tests/test_model/test_qwen3_omni_thinker_length.py
```diff
@@ -42,21 +38,11 @@ def _post_chat(
 
 @pytest.fixture(scope="module")
 def router_server(tmp_path_factory: pytest.TempPathFactory):
-    worker_extra_args = (
-        "--config examples/configs/qwen3_omni_colocated_h20.yaml "
-        "--colocate "
-        f"--stages.0.factory-args.thinker-max-seq-len {THINKER_MAX_SEQ_LEN} "
-        f"--stages.4.factory-args.thinker-max-seq-len {THINKER_MAX_SEQ_LEN}"
+    from tests.test_model.conftest import _start_qwen3_omni_tp2
```
zhaochenyang20: why a private function is imported? `_start_qwen3_omni_tp2`? Also, import functions at the beginning of file.

tests/test_model/test_qwen3_omni_videoamme_talker_ci.py
zhaochenyang20, line 1: This is the removed stage 11, right?

zhaochenyang20 (dismissed):

tests/test_model/test_qwen3_omni_videoamme_talker_ci.py
yxs: yes

tests/test_model/conftest.py
yxs: These are launch-time topology args — GPU placement (`--gpu-thinker-tp 0,1`) and TP size (`--thinker-tp-size 2`), not model config.

tests/test_model/conftest.py
```diff
@@ -45,141 +45,228 @@
...
-QWEN3_OMNI_COLOCATED_WORKER_ARGS = (
-    "--config examples/configs/qwen3_omni_colocated_h20.yaml --colocate"
+QWEN3_OMNI_TP2_THINKER_MEM_FRACTION = "0.55"
+QWEN3_OMNI_TP2_TALKER_MEM_FRACTION = "0.20"
+QWEN3_OMNI_TP2_THINKER_MAX_SEQ_LEN = 32768
+QWEN3_OMNI_FP8_COLOCATED_CONFIG = "examples/configs/qwen3_omni_colocated_h100_fp8.yaml"
+QWEN3_OMNI_FP8_COLOCATED_VIDEO_ARGS = (
+    f"--config {QWEN3_OMNI_FP8_COLOCATED_CONFIG} --colocate "
+    f"--stages.0.factory-args.thinker-max-seq-len {QWEN3_OMNI_TP2_THINKER_MAX_SEQ_LEN} "
+    f"--stages.4.factory-args.thinker-max-seq-len {QWEN3_OMNI_TP2_THINKER_MAX_SEQ_LEN}"
```
Hayden727, line 55: Correction: these indices are correct and intentional, not a wrong-stage bug — in the speech pipeline stage 4 is `thinker` and stage 0 is `preprocessing`, which legitimately needs `thinker-max-seq-len` for the overlong-prompt 400 check, so both are set on purpose. And moving it into the YAML `stage_overrides` (the thread below) isn't possible today: `_apply_stage_overrides` (`config/manager.py:170`) accepts `runtime` overrides only and rejects `factory_args`, so a name-keyed override would fail config validation at load. Keeping it as a CLI `--stages.N.factory-args` override is the only supported path unless the schema is extended. No change needed here.

examples/configs/qwen3_omni_colocated_h100_fp8.yaml
```diff
@@ -0,0 +1,34 @@
+# SPDX-License-Identifier: Apache-2.0
+# Note (Xuesong): FP8 colocated profile for the H100 video-CI router stages (2 DP replicas);
+#   distinct from examples/configs/qwen3_omni_fp8_colocated.yaml (thinker 0.75) — this profile uses 0.55.
+
+config_cls: Qwen3OmniSpeechColocatedPipelineConfig
+name: qwen3-omni-colocated-h100-fp8
+model_path: marksverdhei/Qwen3-Omni-30B-A3B-FP8
```
Hayden727, line 7: Retracting this — not an issue. `sglang_omni_router/launcher/local.py:58` always passes `--model-path <launcher model_path>`, and `cli/serve.py:1166` makes an explicit `--model-path` override the config's `model_path`. So `SGLANG_OMNI_TEST_QWEN3_FP8_MODEL` is honored end-to-end; the `model_path` here is just a standalone-use default that the managed router always overrides. No change needed.

tests/test_model/conftest.py
Hayden727: `thinker-max-seq-len` is a factory arg, and `_apply_stage_overrides` (config/manager.py:170) only accepts `runtime` overrides — it raises `ValueError` on a `factory_args` key. So it cannot be moved into the YAML `stage_overrides` without extending the schema (a framework change, better as a follow-up). The current `--stages.0/.4.factory-args` CLI override is the only supported path today, and the indices are correct (0 = preprocessing for the length-validation 400, 4 = thinker). Leaving this open for your call: keep the CLI args, or file a follow-up to support `factory_args` in `stage_overrides`.

yxs: we can keep it for now

PR #888, opened 2026-06-25
[MOSS] Decouple streaming vocoder first-chunk default from the coalesce join floor

JiaxinD:
## Motivation

The MOSS-TTS-Local streaming vocoder used a single value, `initial_chunk_frames`
(factory default `5`), for two distinct jobs:

- the **first-chunk fallback** when a streaming request reaches the vocoder without an
  `initial_codec_chunk_frames` param
  ([`streaming_vocoder.py#L623`](https://github.com/sgl-project/sglang-omni/blob/30cde5a8dafae1458a7a450c0b00fcbcd2e76f6f/sglang_omni/models/moss_tts_local/streaming_vocoder.py#L623)), and
- the **coalescing join floor**
  ([`streaming_vocoder.py#L631-L632`](https://github.com/sgl-project/sglang-omni/blob/30cde5a8dafae1458a7a450c0b00fcbcd2e76f6f/sglang_omni/models/moss_tts_local/streaming_vocoder.py#L631-L632)).

The serve layer defaults streaming requests to a 1-frame first chunk for low TTFC: when a
streaming request omits `initial_codec_chunk_frames`, `_build_extra_params` injects
`RAW_PCM_DEFAULT_INITIAL_CODEC_CHUNK_FRAMES = 1`
([`speech_service.py#L63`](https://github.com/sgl-project/sglang-omni/blob/30cde5a8dafae1458a7a450c0b00fcbcd2e76f6f/sglang_omni/serve/speech_service.py#L63),
[`#L759-L760`](https://github.com/sgl-project/sglang-omni/blob/30cde5a8dafae1458a7a450c0b00fcbcd2e76f6f/sglang_omni/serve/speech_service.py#L759-L760)).
So the OpenAI speech API path already runs at first-chunk `1`. But a streaming caller that
reaches the vocoder *without* that injection falls back to the vocoder's `5` instead of the
production default `1`. It is low severity (TTFC-only, not a correctness bug, and the speech
API path is unaffected), but the two hardcoded defaults disagreeing is a footgun worth removing.

## Fix

Split the one value into two:

- `create_vocoder_executor`'s first-chunk default becomes `1`, matching the serve-layer
  streaming default, so a bypass caller's first chunk aligns with production.
- a new `coalesce_floor_frames` (default `5`) drives the join floor, which is left **unchanged**.

`join_floor` now reads `coalesce_floor_frames` instead of `initial_chunk_frames`; the
first-chunk fallback is untouched.

## Scope / behaviour

- **No change to the normal API path.** Streaming via the speech API already gets first-chunk
  `1` from the serve layer, and the `join_floor` value stays `5`, so steady-state coalescing is
  unchanged. The only behaviour change is that a bypass streaming caller's first-chunk fallback
  moves `5 -> 1`.
- Because `join_floor` keeps the same value (`5`), no perf / coverage / bit-identity retest is
  needed; this is a default-reconciliation change, not a coalescing or codec-math change.
- Exposing these as pipeline-config fields (rather than factory defaults), and any further
  unification of the serve and vocoder streaming defaults into one constant, is left as a
  follow-up; this PR only reconciles the default values.

## Tests

- New `test_factory_default_decouples_first_chunk_from_join_floor`: the factory default streams a
  1-frame first chunk while the join floor stays `5`, and the streamed PCM still concatenates to
  the offline reference.
- Existing streaming-vocoder tests pass unchanged: they construct with explicit `initial_chunk_frames`,
  so their first-chunk behaviour is unaffected, and their coalescing assertions still hold under the
  decoupled `join_floor`. `tests/unit_test/moss_tts_local/test_streaming_vocoder.py`: 33 passed locally.


Ratish1 (approved):
LGTM

PR #889, opened 2026-06-25
[Benchmark] Fix TTS serving stress scenario contracts

Ratish1:
## Motivation

Refs #601.

The TTS serving stress benchmark should now be strict against the merged speech, voice, batch, and WebSocket APIs. This PR aligns the benchmark contract with the supported serving behavior while keeping expected failures explicit and coverage-gated.

## Modifications

- Updates the stress scenario matrix to distinguish max-valid input from over-limit input, adds a required `speech.long_prefill_decode` contract with `4096` input chars and `max_new_tokens=2048`, and records exact coverage for that shape in `results.json`.
- Adds file-reference coverage to the checked-in stress spec and keeps the service-side requirement explicit through the documented `--allowed-local-media-path docs/_static/audio` launch path.
- Makes voice lifecycle scenarios self-cleaning and observable by uploading, listing, synthesizing with the uploaded voice, deleting, checking deletion, and validating cache invalidation behavior instead of leaving persistent state behind.
- Fixes WebSocket benchmark validation to consume the full audio session until `session.done`, including multi-sentence streams that produce multiple `audio.start` / `audio.done` cycles.
- Tightens batch and named-voice contracts so missing `model` / `voice`, unknown named voices, partial batch item failures, and expected error envelopes are classified deterministically.
- Updates TTS docs and cookbook examples to send explicit `model` and `voice` fields where the public serving API requires them.

## Root Cause

- The benchmark still had a few stale assumptions after the API work landed: a named preset voice was treated as a success case, WebSocket streaming was validated as if a long input produced only one audio cycle, and voice deletion expected cache invalidation without first forcing the uploaded voice into the cache.
- The stress spec also lacked a concrete file-reference scenario and did not explicitly require a valid long-prefill plus high decode-budget success path.

## Accuracy Tests

- stress run:

```bash
sgl-omni serve \
  --allowed-local-media-path docs/_static/audio \
  --port 8000

python -m benchmarks.eval.benchmark_tts_serving \
  --spec benchmarks/tts_serving/examples/stress.json \
  --out stress/out
```

Observed result:

```text
harness_status: ok
passed: True
overall: {'passed': True, 'total': 2124, 'traffic_total': 2124, 'capability_probe_total': 0, 'succeeded': 2124, 'failed': 0, 'load_generation_valid': True, 'coverage_contract_valid': True}
status_counts: {'ok': 1428, 'expected_error': 696}
coverage_failures: []
unsupported_contracts: []
```

The `speech.long_prefill_decode` coverage row was tested in all five speech-enabled stages with `input_chars=4096` and `max_new_tokens=2048`.

## Checklist


sglang_omni/serve/speech_service.py
```diff
@@ -487,6 +506,20 @@ def _build_batch_item_request(
...
+        if not isinstance(items, list):
+            return
+        default_payload = {
+            key: value for key, value in payload.items() if key != "items"
+        }
+        for index, item in enumerate(items):
+            if not isinstance(item, dict):
+                continue
+            merged = dict(default_payload)
+            merged.update(item)
```
zijiexia: This preserves an explicit JSON null, then calls _validate_required_generation_fields → raises 400 "voice is required". But the actual _build_batch_item_request (lines 472-474) builds the payload with model_dump(exclude_none=True) on both batch defaults and item, dropping the null so the item inherits the batch-level voice. The pre-check is stricter than the code it's meant to guard.

sglang_omni/serve/speech_service.py
```diff
@@ -144,20 +146,36 @@ def parse_batch_request(self, payload: Any) -> CreateSpeechBatchRequest:
...
         self._validate_raw_payload(payload)
+        if require_explicit_model_and_voice:
+            _validate_required_generation_fields(payload)
 
     def parse_generation_request(self, payload: Any) -> PreparedSpeechRequest:
         """Parse and prepare a raw HTTP payload for GenerateRequest lowering."""
 
-        return self.prepare_generation_request(self._parse_raw_request(payload))
+        return self.prepare_generation_request(
+            self._parse_raw_request(payload, require_explicit_model_and_voice=True)
```
zijiexia: `require_explicit_model_and_voice=True` runs `_validate_required_generation_fields` before any reference logic, so a cloning request with `ref_audio/references` (where `voice` is semantically irrelevant) but no voice now fails 400. Is this intentional? Maybe worth an explicit release note, or consider exempting voice from the required check when a reference is supplied.

benchmarks/tts_serving/ws_client.py
```diff
@@ -267,6 +280,65 @@ async def _expect_audio_until_done(
...
+                expect_success=expect_success,
+            )
+            if result.status in {"failed", "expected_error"}:
+                return event_type == "error"
+            if event_type in WS_CONTROL_EVENT_TYPES:
+                continue
+            if event_type in {"audio.start", "audio.done"}:
+                continue
+            if event_type == "session.done":
+                if binary_frames < min_binary_frames:
```
zijiexia, line 322: consider asserting the minimum frame count per `audio.start`/ `audio.done` cycle to preserve the "requiring incremental binary audio" intent.

docs/cookbook/qwen3_tts.md
```diff
@@ -152,7 +160,9 @@ for a full Python raw PCM consumer.
 
 | Parameter | Default | Notes |
 |---|---|---|
+| `model` | (required) | Served model identifier |
 | `input` | (required) | Text to synthesize |
+| `voice` | (required) | Voice identifier. Use `default` for Base reference-cloning requests |
```
zijiexia: maybe noting that the field has no effect for Base and is required only to satisfy the unified required-field validation.

sglang_omni/serve/speech_service.py

sglang_omni/serve/speech_service.py
Ratish1: clarified in the docs that Base reference cloning should use default while the reference clip provides speaker conditioning

benchmarks/tts_serving/ws_client.py
Ratish1: done

docs/cookbook/qwen3_tts.md
Ratish1: done

benchmarks/tts_serving/ws_client.py
```diff
@@ -179,6 +193,7 @@ async def _expect_next_event(
                 result,
                 audio_state,
                 expect_success=expect_success,
+                min_binary_frames_per_sentence=min_binary_frames,
```
zijiexia: `min_binary_frames` is undefined here → NameError. `_expect_next_event`'s signature is only `*, expected_event, expect_success`, so the moment this function receives a TEXT frame it raises `NameError: name 'min_binary_frames' is not defined`, crashing every WS scenario that uses `{"action": "expect"}` (`_websocket_normal`, `_websocket_multi_sentence`, `_websocket_clause_split`, and the error scenarios — 11 sites). Maybe drop this?

benchmarks/tts_serving/ws_client.py
```diff
@@ -267,12 +283,73 @@ async def _expect_audio_until_done(
...
+                return False
+            binary_frames += 1
+            continue
+        if msg.type == aiohttp.WSMsgType.TEXT:
+            event_type = _merge_text_event(
+                msg.data,
+                result,
+                audio_state,
+                expect_success=expect_success,
+                min_binary_frames_per_sentence=min_binary_frames,
```
zijiexia: Is this too strict? `min_binary_frames=2` is reused as a per-sentence threshold, but the server emits one binary frame per audio chunk (`_stream_sentence_audio`), so a short sentence yielding a single chunk fails the whole scenario even when total frames ≥ 2. The intent is to verify incremental streaming, not that every sentence emits ≥2 frames. Maybe decouple the thresholds? keep the session-total binary_frames >= min_binary_frames check, but pass a per-sentence minimum of 1 (or require "at least one sentence with ≥2 frames") instead of forwarding min_binary_frames as min_binary_frames_per_sentence here.

zijiexia (changes requested):
Can we re-validate it on the real model? Thanks Ratish

benchmarks/tts_serving/ws_client.py
Ratish1: fixed

benchmarks/tts_serving/ws_client.py
Ratish1: done

Ratish1:
Hey @zijiexia , tested again and all pass.

Screenshot 2026-07-07 at 3 55 24 PM

zijiexia (approved):
might worth noting that `/v1/audio/speech`, `/v1/audio/speech/batch`, and the WebSocket `session.config` now require an explicit `model` and `voice` (`voice` no longer defaults to `"default"`);

PR #892, opened 2026-06-26
[Perf] Moss streaming chunk coalescing

edwingao28:
## Motivation

MOSS-TTS Local's streaming vocoder is the one **vocoder-bound** audio path in this repo: the codec's
streaming `step()` is fixed-cost per forward — it costs about the same whether it decodes one lane or
eight. `_pump_streams` already coalesces all **currently-due** lanes into one `step()`, but it runs
once per incoming `stream_chunk` message. Under streaming concurrency the scheduler often pumps right
after appending a *single* chunk while more chunks for other lanes are still unread in the inbox, so
those lanes aren't "due" yet and miss the join — the coalesced step is narrower than it could be and
first-audio latency (TTFA) inflates.

This PR widens the coalescing window with **zero added latency**: drain the `stream_chunk` messages
that are *already queued*, append them all, then pump **once**. More lanes are ready at pump time, so
each fixed-cost `step()` serves more of them.

## Modifications

**1. Default-OFF batching seam in `StreamingSimpleScheduler`** (`sglang_omni/scheduling/streaming_simple_scheduler.py`)
- New opt-in class flag `_batch_stream_chunks` (default `False`). When `True`, `_handle_message`
  drains already-queued `stream_chunk` messages (`_collect_stream_chunk_batch`) and dispatches them
  as one batch via a new `on_stream_chunk_batch` hook.
- The drain mirrors `_collect_new_request_batch`: bounded by the batch cap, drops aborted ids, stops
  at the first non-`stream_chunk` message and pushes it back to the front of `_pending_messages` to
  preserve arrival order. No blocking wait → only already-queued chunks coalesce (zero added latency).
- **Default OFF ⇒ byte-identical for every current subclass** (Qwen3-Omni talker, Fish, Higgs): the
  per-chunk path is untouched and the default hook loops the per-chunk handler with the same per-item
  error isolation as the serving loop.

**2. MOSS-TTS Local opts in** (`sglang_omni/models/moss_tts_local/streaming_vocoder.py`)
- Sets `_batch_stream_chunks = True` and overrides `on_stream_chunk_batch` to append every drained
  chunk to its lane, then call the existing `_pump_streams()` **once**. The single-chunk path is
  refactored into `_ingest_chunk` + pump and stays byte-identical.

Per-request **concatenated** PCM is bit-identical, and the codes fed to each `step()` are unchanged —
only *how many* lanes share a `step()` changes. Coalescing several **already-queued same-request**
chunks before one pump can *widen* an emitted chunk's boundary (e.g. a first chunk of 8 frames instead
of `initial_chunk_frames=5`); the concatenation is still identical. Each streaming `step()` is now
capped at `stream_chunk_frames`, so a coalesced backlog never overshoots the #886 CUDA-graph capture
range — the two compose.

## Related Issues

Relates to #661.

## Accuracy Test

N/A — this changes streaming **scheduling** (how many lanes share a vocoder step), not model-side
compute. The streamed PCM is identical to the non-streaming offline decode; the unit suite asserts
this bit-for-bit, including through the **real batched seam**:
`test_batched_coalescing_matches_offline_decode` (concat PCM == offline under same-request coalescing)
and `test_batched_step_capped_at_chunk_frames` (`stream_slots > max_batch_size`, drain coalesces to one
pump, step stays ≤ `stream_chunk_frames`) in
`tests/unit_test/moss_tts_local/test_streaming_vocoder.py`.

## Benchmark & Profiling

1×H100-80GB, MOSS-TTS Local, stream-only, default config (`examples/configs/moss_tts_local.yaml`),
**same-GPU back-to-back A/B with GPU-quiescence between arms**: baseline = `main`, treatment = this
PR. **5 pairs per concurrency; figures are the median per-pair delta.** Both arms: **0 errors / no
OOM through c=64**.

> **Caveats.** Shared multi-tenant H100, n=5 pairs, no variance/CI; the c=32 throughput median is over
> 3 pairs after dropping 2 foreign-tenant-contaminated ones. Treat the **direction** (throughput up at
> mid concurrency, tapering as the 8 lanes saturate) as the result and the magnitudes as indicative,
> not measured — a dedicated-GPU re-run with all pairs + variance is the follow-up. The table also
> predates the `step()` ≤ `stream_chunk_frames` cap commit; re-measure pending.

| concurrency | TTFA p95 Δ (median) | throughput Δ (median) |
|---|---|---|
| 8  | **−6%** | **+4%**  |
| 16 | **−4%** | **+9%**  |
| 32 | **−5%** | **+11%** (peak) |
| 64 | **−5%** | **+4%**  |

- Throughput rises with concurrency to a **+11% peak at c=32**, then tapers to +4% at c=64 as the
  default 8 vocoder lanes (`stream_slots=8`) saturate and extra concurrency becomes pure queueing —
  coalescing can't get wider than the lane count.
- TTFA p95 improves a steady **−4…−6%** across the board (c=64 negative in all 5 pairs).
- c=1 is the inert control (no coalescing pressure): within noise, confirming no regression on the
  no-coalescing path.

WER results:
```
  ┌────────────────────────┬────────────┬─────────────────┬───────────────────┬───────┐
  │          arm           │ wer_corpus │ per-sample mean │ per-sample median │   n   │
  ├────────────────────────┼────────────┼─────────────────┼───────────────────┼───────┤
  │ baseline (d40cee8)     │ 1.60%      │ 1.58%           │ 0.00%             │ 50/50 │
  ├────────────────────────┼────────────┼─────────────────┼───────────────────┼───────┤
  │ treatment (PR 459cc60) │ 1.42%      │ 1.36%           │ 0.00%             │ 50/50 │
  └────────────────────────┴────────────┴─────────────────┴───────────────────┴───────┘
```
  WER is flat within noise (Δ = −0.18pp, treatment marginally lower; both medians 0.00%). Exactly the expected result: coalescing vocoder lanes doesn't touch token generation, so intelligibility is unchanged. 

## Checklist

- [x] Format your code according with pre-commit. (black 24.10.0 + isort 5.13.2)
- [x] Add unit tests. Base-seam tests for coalesce / boundary-pushback / abort-skip / cap /
  default-hook and per-item error isolation; the MOSS streaming oracle now also runs under batched
  coalescing (streamed PCM == offline decode).
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with
  merging the main branch, please remove yourself as a co-author when merging the PR.


sglang_omni/scheduling/streaming_simple_scheduler.py
```diff
@@ -417,6 +458,16 @@ def _handle_stream_chunk(self, request_id: str, item: Any) -> None:
...
 
+    def _handle_stream_chunk_batch(self, batch: list[IncomingMessage]) -> None:
+        items = [
+            (msg.request_id, msg.data)
+            for msg in batch
+            if not self._is_aborted(msg.request_id)
+        ]
+        if items:
+            self._stream_chunk_batch_hist[len(items)] += 1
+            self.on_stream_chunk_batch(items)
```
AkazaAkane, line 487: Should this mirror `_handle_stream_chunk()` and enqueue the returned `OutgoingMessage`s?

Right now `on_stream_chunk_batch()` has the same return shape as `on_stream_chunk()` (`list[OutgoingMessage]`), but `_handle_stream_chunk_batch()` drops the return value. That makes the new base hook easy to misuse by future subclasses.

Either this should enqueue returned messages here, or the hook should return `None` and document that batch overrides must enqueue internally.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -407,6 +409,26 @@ def on_streaming_new_request(self, request_id: str, payload: StagePayload) -> No
     def on_stream_chunk(
         self, request_id: str, item: StreamItem
     ) -> list[OutgoingMessage]:
+        self._ingest_chunk(request_id, item)
+        self._pump_streams()
```
AkazaAkane, line 419: Can we add a same-request queued-chunk test for this path?

Because the batch hook ingests all drained chunks before calling `_pump_streams()` once, multiple queued chunks for the same request can change emitted stream boundaries. For example, with `initial_chunk_frames=5`, if 8 chunks for one request are already queued, this path may emit an 8-frame first stream chunk instead of the previous 5-frame first chunk.

The concatenated PCM may still be identical, but the streaming boundary / TTFA behavior is not strictly unchanged unless we test or explicitly accept that.

AkazaAkane (commented):
Leave some comment

sglang_omni/scheduling/streaming_simple_scheduler.py
```diff
@@ -275,6 +294,28 @@ def _collect_new_request_batch(
             batch.append(msg)
         return batch
 
+    def _collect_stream_chunk_batch(
+        self, first_msg: IncomingMessage
+    ) -> list[IncomingMessage]:
+        """(wenyao) Front-pushback of the first non-chunk message preserves arrival order; no
+        blocking wait, so only already-queued chunks coalesce."""
+        batch = [first_msg]
+        cap = self._stream_chunk_batch_max or max(self._max_batch_size, 1)
```
SandyLuXY, line 307: This cap probably should not fall back to `_max_batch_size` for opt-in streaming chunk batching. `max_batch_size` also controls the non-streaming/offline vocoder batch path, while `stream_slots` is the number of streaming slots a codec step can serve. With a valid config like `stream_slots=16, max_batch_size=4`, we allocate 16 streaming slots but only drain 4 queued `stream_chunk` messages before pumping, so already-queued requests can still miss the intended coalesced step and get handled in a narrower follow-up pump.
Can MOSS set `_stream_chunk_batch_max = stream_slots`, or can we otherwise make the stream-chunk drain cap explicit instead of inheriting `_max_batch_size`? A small test with `stream_slots > max_batch_size` would make this harder to regress.

sglang_omni/scheduling/streaming_simple_scheduler.py
```diff
@@ -72,6 +75,7 @@ def __init__(
         self._completed_non_streaming_request_ids: set[str] = set()
         self._state_lock = threading.RLock()
         self._abort_lock = threading.Lock()
+        self._stream_chunk_batch_hist: collections.Counter = collections.Counter()
```
Ccyest: nit: _stream_chunk_batch_hist is a Counter recording batch-size distribution, but has no logging, no metrics exposure, and no public accessor.
Fix:
Follow the _log_cg_stats pattern: log the histogram in the vocoder scheduler's stop() or via a periodic callback, or add a log_batch_stats() method to the base class:

```
def _log_stream_chunk_batch_stats(self) -> None:
    if not self._stream_chunk_batch_hist:
        return
    total = sum(self._stream_chunk_batch_hist.values())
    logger.info(
        "%s stream-chunk batch stats: %d dispatches, sizes=%s",
        self.__class__.__name__,
        total,
        dict(sorted(self._stream_chunk_batch_hist.items())),
    )
```

edwingao28:
> **Methodology & data hygiene.** 
> Same-GPU back-to-back A/B with a GPU-quiescence drain between arms, on a shared multi-tenant H100. Throughput is reported as the **median of the un-contaminated pairs**: a pair's throughput is dropped when either arm reads `tput_rps < 2.5` at `c ≥ 8`, which is a foreign-tenant GPU stall, not a property of this change (TTFA for that pair is still kept — the stall inflates wall-clock throughput but not per-request first-audio latency). 
This affects **c=32 only, where 2 of 5 pairs are dropped**, so the c=32 throughput median is over 3 clean pairs. All
> other concurrencies use all 5 pairs. Every pair, both arms, had **0 errors / no OOM**. TTFA medians
> are robust but not monotone per-pair (a few pairs flip positive under noise); only **c=64 is negative in all 5 pairs**. Full per-pair data below.

<details>
<summary><b>Raw per-pair A/B data</b> (5 pairs × {baseline=main, treatment=this PR}, 0 errors throughout)</summary>

**c=1** — inert control, no coalescing pressure (n=48/arm)

| pair | TTFA p95 base→treat (s) | TTFA Δ | tput base→treat (rps) | tput Δ |
|---|---|---|---|---|
| r1 | 0.055→0.055 | +0% | 1.408→1.386 | −2% |
| r2 | 0.058→0.055 | −5% | 1.384→1.419 | +3% |
| r3 | 0.055→0.054 | −2% | 1.412→1.405 | +0% |
| r4 | 0.055→0.054 | −2% | 1.406→1.392 | −1% |
| r5 | 0.055→0.054 | −2% | 1.414→1.413 | +0% |
| **median** | | **−2%** | | **+0%** |

**c=8** (n=48/arm)

| pair | TTFA p95 base→treat (s) | TTFA Δ | tput base→treat (rps) | tput Δ |
|---|---|---|---|---|
| r1 | 0.381→0.363 | −5% | 4.983→5.336 | +7% |
| r2 | 0.332→0.398 | +20% | 5.171→5.186 | +0% |
| r3 | 0.484→0.344 | −29% | 5.035→5.237 | +4% |
| r4 | 0.481→0.394 | −18% | 5.106→5.201 | +2% |
| r5 | 0.436→0.408 | −6% | 5.024→5.270 | +5% |
| **median** | | **−6%** | | **+4%** |

**c=16** (n=64/arm)

| pair | TTFA p95 base→treat (s) | TTFA Δ | tput base→treat (rps) | tput Δ |
|---|---|---|---|---|
| r1 | 3.710→2.853 | −23% | 3.989→5.180 | +30% |
| r2 | 3.211→3.087 | −4% | 4.494→4.908 | +9% |
| r3 | 3.596→3.192 | −11% | 4.199→4.695 | +12% |
| r4 | 3.327→3.562 | +7% | 4.620→4.003 | −13% |
| r5 | 3.414→3.633 | +6% | 4.512→4.568 | +1% |
| **median** | | **−4%** | | **+9%** |

**c=32** (n=96/arm) — r3, r5 throughput contaminated (foreign-tenant stall, `tput_rps < 2.5`), dropped

| pair | TTFA p95 base→treat (s) | TTFA Δ | tput base→treat (rps) | tput Δ | note |
|---|---|---|---|---|---|
| r1 | 6.665→6.353 | −5% | 5.098→5.115 | +0% | |
| r2 | 6.333→5.804 | −8% | 4.900→5.519 | +13% | |
| r3 | 6.584→6.479 | −2% | 5.171→1.863 | −64% | contaminated → tput dropped |
| r4 | 6.755→6.463 | −4% | 4.947→5.484 | +11% | |
| r5 | 7.253→5.975 | −18% | 1.754→5.563 | +217% | contaminated → tput dropped |
| **median** | | **−5%** | | **+11%** | (tput over 3 clean pairs: +0/+13/+11) |

**c=64** (n=96/arm)

| pair | TTFA p95 base→treat (s) | TTFA Δ | tput base→treat (rps) | tput Δ |
|---|---|---|---|---|
| r1 | 14.062→13.268 | −6% | 4.700→4.832 | +3% |
| r2 | 14.413→13.774 | −4% | 4.553→4.744 | +4% |
| r3 | 14.189→13.545 | −5% | 4.647→4.837 | +4% |
| r4 | 14.089→12.634 | −10% | 4.791→5.279 | +10% |
| r5 | 14.770→14.233 | −4% | 4.538→4.733 | +4% |
| **median** | | **−5%** | | **+4%** |

</details>

sglang_omni/scheduling/streaming_simple_scheduler.py
edwingao28: addressed, changed `on_stream_chunk_batch()` to return `None` and documented that batch overrides must emit internaly. also updated batch path tests to match the non hook contract

sglang_omni/models/moss_tts_local/streaming_vocoder.py
edwingao28: updated test drives the real batched seam via queued `stream_chunk` messages

sglang_omni/scheduling/streaming_simple_scheduler.py
edwingao28: yeah it should be tied to streaming lane capacity rather than the offline batch width. added coverage with `stream_slots=16, max_batch_size=4`

sglang_omni/scheduling/streaming_simple_scheduler.py
edwingao28: fixed

sglang_omni/scheduling/streaming_simple_scheduler.py
```diff
@@ -417,6 +456,27 @@ def _handle_stream_chunk(self, request_id: str, item: Any) -> None:
...
+        items = [
+            (msg.request_id, msg.data)
+            for msg in batch
+            if not self._is_aborted(msg.request_id)
+        ]
+        if items:
+            self._stream_chunk_batch_hist[len(items)] += 1
+            self.on_stream_chunk_batch(items)
+
+    def _log_stream_chunk_batch_stats(self) -> None:
```
Ratish1: I dont think we need this logging in this scheduler. can we remove Can we remove _log_stream_chunk_batch_stats() and _stream_chunk_batch_hist entirely?. update tests as well.

edwingao28:
```
  ┌────────────────────────┬────────────┬─────────────────┬───────────────────┬───────┐
  │          arm           │ wer_corpus │ per-sample mean │ per-sample median │   n   │
  ├────────────────────────┼────────────┼─────────────────┼───────────────────┼───────┤
  │ baseline (d40cee8)     │ 1.60%      │ 1.58%           │ 0.00%             │ 50/50 │
  ├────────────────────────┼────────────┼─────────────────┼───────────────────┼───────┤
  │ treatment (PR 459cc60) │ 1.42%      │ 1.36%           │ 0.00%             │ 50/50 │
  └────────────────────────┴────────────┴─────────────────┴───────────────────┴───────┘
```
  WER is flat within noise (Δ = −0.18pp, treatment marginally lower; both medians 0.00%). Exactly the expected result: coalescing vocoder lanes doesn't touch token generation, so intelligibility is unchanged.

sglang_omni/scheduling/streaming_simple_scheduler.py
```diff
@@ -417,6 +456,27 @@ def _handle_stream_chunk(self, request_id: str, item: Any) -> None:
                 if not self._is_aborted(request_id):
                     self.outbox.put(out)
 
+    def _handle_stream_chunk_batch(self, batch: list[IncomingMessage]) -> None:
```
Ratish1, line 469: can we keep the batch path's input contract identical to the single-chunk path?

_handle_stream_chunk() validates that every stream payload is a StreamItem before calling subclass code. _handle_stream_chunk_batch() currently skips that validation and passes raw msg.data into the batch hook. The default hook happens to re-enter _handle_stream_chunk(), but the MOSS override does not, so invalid batched input now fails in subclass internals instead of at the scheduler boundary.

sglang_omni/scheduling/streaming_simple_scheduler.py
```diff
@@ -39,6 +39,9 @@ class StreamingSimpleScheduler:
     streaming requests are kept out of the non-streaming batch path.
     """
 
+    _batch_stream_chunks: bool = False
```
Ratish1: rename this to  `_can_batch_stream_chunks`

tests/unit_test/pipeline/test_streaming_simple_scheduler.py
```diff
@@ -163,3 +163,115 @@ def test_streaming_simple_scheduler_keeps_queued_control_message_out_of_batch()
...
+                )
+            )
+        return []
+
+
+class _DefaultBatchScheduler(_TestStreamingScheduler):
+    _batch_stream_chunks = True
+
+
+def test_inc3_opted_out_dispatches_one_at_a_time() -> None:
```
Ratish1: should rename this along with the other tests below

sglang_omni/scheduling/streaming_simple_scheduler.py
edwingao28: removed _stream_chunk_batch_hist and _log_stream_chunk_batch_stats(), updated tests to assert batchingbehavior directly instead of inspecting scheduler telemetry internals

sglang_omni/scheduling/streaming_simple_scheduler.py
edwingao28: fixed

tests/unit_test/pipeline/test_streaming_simple_scheduler.py
edwingao28: fixed

sglang_omni/scheduling/streaming_simple_scheduler.py
```diff
@@ -407,16 +455,27 @@ def _handle_streaming_new_request(self, request_id: str, payload: Any) -> None:
...
+        for msg in batch:
+            if self._is_aborted(msg.request_id):
+                continue
+            try:
+                item = self._validate_stream_chunk_item(msg.request_id, msg.data)
+            except Exception as exc:
+                self._emit_error(msg.request_id, exc)
+                self.abort(msg.request_id)
+                continue
+            items.append((msg.request_id, item))
```
Ratish1, line 480: small batch-validation edge case: after validating the batch, can we filter out items whose request was aborted during the same validation pass?

If request r has an earlier valid chunk and a later invalid chunk in the same drained batch, the later invalid chunk calls abort(r), but the earlier (r, item) is still left in items. MOSS skips aborted ids in its override, so the MOSS path is fine, but the base default hook would still call _handle_stream_chunk() for r and can recreate subclass state after abort() cleared it.

Filtering items once more before on_stream_chunk_batch(items), or skipping aborted ids in the default hook can keep the base clean.

Ratish1 (approved):
Rest LGTM @edwingao28

sglang_omni/scheduling/streaming_simple_scheduler.py
```diff
@@ -275,6 +293,28 @@ def _collect_new_request_batch(
             batch.append(msg)
         return batch
 
+    def _collect_stream_chunk_batch(
+        self, first_msg: IncomingMessage
+    ) -> list[IncomingMessage]:
+        """(wenyao) Front-pushback of the first non-chunk message preserves arrival order; no
```
JiaxinD: Nit: drop the `(wenyao)` author tag from the docstring (repo convention: no author tags in shipped comments). The wording itself is good.

tests/unit_test/moss_tts_local/test_streaming_vocoder.py
```diff
@@ -408,6 +409,94 @@ def test_factory_default_decouples_first_chunk_from_join_floor(monkeypatch) -> N
...
+            break
+        scheduler._handle_message(msg, None)
+    scheduler._on_done(request_id)
+    scheduler._on_streaming_new_request(
+        request_id, _terminal_payload(rows, request_id=request_id, params=params)
+    )
+    return _drain(scheduler)
+
+
+def test_batched_coalescing_matches_offline_decode(monkeypatch) -> None:
```
JiaxinD, line 446: Both batched-seam tests use a single request id, so multi-lane coalescing and the failure branch (`on_stream_chunk_batch` emit_error + off-lock cleanup, L422-438) aren't exercised through the real seam. Could we add a 2-lane batched test plus one where `_ingest_chunk` raises?

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -411,6 +415,29 @@ def on_streaming_new_request(self, request_id: str, payload: StagePayload) -> No
...
+        return []
+
+    def on_stream_chunk_batch(self, items: list[tuple[str, StreamItem]]) -> None:
```
JiaxinD, line 422: `on_stream_chunk_batch` hand-rolls abort()'s body then defers cleanup, while the base default hook uses `self.abort()`. Could we extract a shared helper (e.g. `_abort_state()`) so the two can't drift?

sglang_omni/scheduling/streaming_simple_scheduler.py
edwingao28: good catcha and fixed

tests/unit_test/moss_tts_local/test_streaming_vocoder.py
edwingao28: fixed

sglang_omni/models/moss_tts_local/streaming_vocoder.py
edwingao28: Fixed by extracting `_abort_state()` and reusing it from the MOSS batch path while keeping cleanup off lock.

PR #900, opened 2026-06-28
[Perf] Defer Omni stage factory imports to worker startup

stephenkgli:
## Motivation

Stage startup currently imports stage factory modules in the parent process while building `StageLaunchConfig`s. For Omni models, these factory modules can pull in heavy model-side dependencies during import. This work is unnecessary in the parent process because the child process will import and construct the factory anyway.

This PR reduces cold-start latency by deferring signature-dependent factory argument injection until the child process starts the stage.

## Modifications

- Split stage factory argument resolution into two phases:
  - parent process: resolve static args from `stage.factory_args`, `runtime_overrides`, and typed runtime fields without importing the factory;
  - child process: import the factory, inspect its signature, and inject standard defaults such as `model_path`, `gpu_id`, and `total_gpu_memory_fraction` only when the factory declares them.

## Related Issues

N/A

## Accuracy Test

Not applicable. This PR does not change model architecture, kernels, request building, sampling, or inference outputs. It only changes when factory modules are imported and when signature-dependent default kwargs are injected during startup.

## Benchmark & Profiling

This benchmark compares the latest `main` baseline against the optimized branch for CI Stage 1 Qwen3-ASR.

To make the comparison fair, I applied only the new `router_ready_s` metric to the `main` checkout. No startup optimization logic was applied to `main`.

- GPU: 2x 4080 SUPER, 32GB each
- Model: `Qwen3-ASR-1.7B`

| Variant  | `router_ready_s` | `router_ready_s` improvement | 
|---|---:|---:|
| `main + metric only` | `28.033s` | baseline |
| `optimized`  | `23.035s` | `-4.998s` / `17.83%` faster |

## Checklist

- [x] Add unit tests.

## CI

Please add the `run-ci` label to trigger self-hosted GPU CI.


codestrongestx:
I ran a single-H100 Qwen3-ASR startup probe:

| Variant | SHA | Startup ready time |
|---|---|---:|
| Baseline | `c631dc9` | `122.204s`, `120.187s` |
| Lazy factory import | `23b9054` | `90.622s`, `89.143s` |

| Metric | Baseline | Lazy factory import | Delta |
|---|---:|---:|---:|
| Mean startup ready time | `121.196s` | `89.883s` | `-31.313s` / `25.84% faster` |

Hardware: `1x NVIDIA H100 80GB HBM3`  
Model: `Qwen/Qwen3-ASR-1.7B`  
Measured from `sglang_omni.cli serve` launch until `/health` was ready.

sglang_omni/pipeline/mp_runner.py
0xjeffro: do we still need this? Looks like we already move gpu_id into factory_arg_defaults so base_factory_args will never contain it now?

(see line ~40-45 in `/config/runtime.py` in code diff)

sglang_omni/pipeline/stage_workers.py
```diff
@@ -438,6 +440,8 @@ def _construct_stage(
     gpu_id = spec.relay_config.get("gpu_id")
     if gpu_id is None:
         gpu_id = spec.factory_args.get("gpu_id")
```
0xjeffro: This will always return None now right? the new fallback below is doing all the work.

sglang_omni/pipeline/stage_workers.py
```diff
@@ -438,6 +440,8 @@ def _construct_stage(
...
+    if gpu_id is None and spec.gpu_id is not None:
+        gpu_id = spec.gpu_id
     if gpu_id is None and _factory_args_use_cuda(spec.factory_args):
```
0xjeffro: this is dead code now.

0xjeffro: `_factory_args_use_cuda` is only called here, might be worth clean up it too if this is a dead code.

sglang_omni/pipeline/mp_runner.py
stephenkgli: In some cases, e.g.  https://github.com/sgl-project/sglang-omni/blob/32586a586a454f8096ff589820ae8b3db8daf78d/sglang_omni/models/moss_tts/config.py#L48, people may fill gpu_id into factory_args, and I don't split it into factory_arg_defaults explicitly now, so it will go into base_factory_args. But the logic of using gpu_id is complicated, trying to simply it.

sglang_omni/pipeline/stage_workers.py
stephenkgli: ditto

sglang_omni/pipeline/stage_workers.py
stephenkgli: yes!

sglang_omni/pipeline/mp_runner.py
0xjeffro: oh cool thanks

sglang_omni/config/runtime.py
```diff
@@ -72,6 +106,19 @@ def reject_untyped_total_gpu_memory_fraction(
     )
 
 
+def reject_gpu_id_in_factory_args(
```
edwingao28, line 109: This new `gpu_id` guard doesn't have a test. Since you're dropping `gpu_id` from all the configs, might be worth locking in that it actually rejects — otherwise someone could quietly break the check later and CI would still be green. Maybe just:

```python
with pytest.raises(ValueError, match="owned by placement"):
    resolve_stage_static_factory_args(stage_with_gpu_id, cfg)
```

for both the `factory_args` and `runtime_overrides` cases.

sglang_omni/config/runtime.py
```diff
@@ -25,32 +26,69 @@ def resolve_stage_factory_args(
...
+) -> dict[str, Any]:
+    """Return standard factory kwargs used only when the factory declares them."""
 
-    if "gpu_id" in sig.parameters and "gpu_id" not in args:
-        args["gpu_id"] = (
+    defaults: dict[str, Any] = {
+        "model_path": global_cfg.model_path,
+        "gpu_id": (
             gpu_id
             if gpu_id is not None
```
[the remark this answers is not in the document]
edwingao28: I don't think this is actually a regression — the old `resolve_stage_factory_args` already injected `gpu_id=None` for unplaced stages (it falls back to `_resolve_primary_gpu_id`, which returns `None`). So the behavior here is unchanged. I'd leave it as is. If we ever want to stop passing `gpu_id=None` at all, that's a separate cleanup and would change behavior for CPU factories that declare `gpu_id`, so better on its own PR.

sglang_omni/pipeline/stage_workers.py
```diff
@@ -635,12 +639,19 @@ def _construct_scheduler(
     """Build a scheduler, serializing GPU factory work per visible device."""
 
     factory = import_string(spec.factory)
+    factory_arg_defaults = dict(spec.factory_arg_defaults)
+    factory_arg_defaults["gpu_id"] = gpu_id
+    factory_args = resolve_factory_signature_args(
+        factory,
```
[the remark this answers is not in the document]
stephenkgli: Just keep the logic same as the main branch

yxs:
I’m concerned about the repeated stage 2 speed failures; this might be a regression.

For the Qwen3-ASR startup claim in the PR body, I could not reproduce a clear win on H200 with the managed-router startup path:

- `main`: `23.036s`, `26.059s`, `23.037s`, `23.035s`, `25.040s`
- PR #900: `23.035s`, `23.044s`, `25.045s`, `34.046s`, `26.042s`

The startup numbers look unstable, so please run a few more base-vs-PR comparisons on the same runner/GPU setup and make sure this is a clear win.

stephenkgli:
@yxs I reran the managed-router CI Stage 1 startup comparison on  2x NVIDIA GeForce RTX 4080 SUPER 32GB using `pytest tests/test_model/test_qwen3_asr_ci.py -v -s -x`. `router_ready_s` is added to measure the elapsed time from launching the managed router until all router workers are healthy and routable. The baseline is current `main` with only the `router_ready_s` metric patch applied (`70b0077`), compared against PR #900 head (`e17d1be`). Metric below is `router_ready_s`.

| Variant | SHA | Runs (`s`) | Mean | Median | Result |
|---|---|---:|---:|---:|---:|
| `main + metric only` | `70b0077` | `29.060`, `31.076`, `29.064`, `30.076`, `29.051` | `29.665s` | `29.064s` | baseline |
| PR #900 | `e17d1be` | `24.063`, `24.048`, `26.107`, `24.064`, `25.038` | `24.664s` | `24.064s` | `-5.002s` / `16.86%` faster |

yxs (dismissed):

yxs (approved):

PR #903, opened 2026-06-29
[Refactor] T7: extract shared PreparedRequestQueue (#661)

MelodyyyYin:
Implements Template 7 of the #661 refactor — extracts the tri-state preprocessing → AR-engine handoff queue (`prepared` / `inflight` / `aborted` + lock) that `moss_tts` and `moss_tts_local` re-implement verbatim into a shared base, and migrates both models onto it.

### Changes
- **New `sglang_omni/scheduling/prepared_request_queue.py`**: generic `PreparedRequestQueue[CtxT, PrepT]` owning the tri-state, its lock, and the lifecycle methods (`set_context` / `clear_context` / `begin` / `fail_inflight` / `publish` / `abort` / `pop`).
  - `begin()` reads the context and marks the request in-flight under a single lock, preserving the original atomicity (a concurrent `clear_context` cannot leave a stale in-flight id).
  - The context is held as an **opaque object** — `moss_tts` carries `processor`, `moss_tts_local` also carries `reference_encoder`; the base does not look inside.
  - `publish()` returns whether the handoff was stored (dropped if aborted mid-flight) — a non-observable API addition; callers ignore it, behavior is unchanged.
- `moss_tts` (reference) and `moss_tts_local` now delegate to the queue; the duplicated per-model queue code is **deleted in the same change** (migrate-or-delete). Model-specific bits stay local: the context dataclass, payload marker, `RuntimeError` messages, and `_prepare_*` compute.
- **No change to `stages.py`** — the public function names/signatures (`cleanup_prepared_*` as `abort_callback`, `set_*_preprocessing_context`) are preserved.
- New CPU-only contract test covering the seven prepared/inflight/aborted race scenarios.

**Behavior-preserving (🟢).** The queue logic is line-for-line equivalent to the previous per-model code; no runtime-path or numeric change.

### Also fixes two pre-existing test failures (test-only)
`test_post_process_outputs_skips_chunked_rows` and `test_async_launch_resolve_matches_sync_collect` build the runner via `__new__` (bypassing `__init__`) and never stub `_outbox`, which `post_process_outputs` reads (`model_runner.py:641`). Stubbing `runner._outbox = None` (the non-streaming production default) fixes both. Verified these fail identically on pristine `origin/main` (CPU and GPU); unrelated to this refactor. **Production code is unchanged.**

### Out of scope
`qwen3_tts` is not part of T7 (single `_PREPARED_REQUESTS` dict, no inflight/aborted tri-state). PR #707's `attach_sglang_req_attrs` is orthogonal and not required here.

### Testing
Contract test 7/7; full `moss_tts` + `moss_tts_local` pipeline suites pass **79 / 0 / 0** on a single GPU (`CUDA_VISIBLE_DEVICES=0`). 7 files changed, +218 / −107.


sglang_omni/scheduling/prepared_request_queue.py
```diff
@@ -0,0 +1,97 @@
...
+                self.inflight.add(request_id)
+            return context
+
+    def fail_inflight(self, request_id: str) -> None:
+        """Roll back an in-flight request whose preprocessing raised."""
+        with self.lock:
+            self.inflight.discard(request_id)
+            self.aborted.discard(request_id)
+
+    def publish(self, request_id: str, prepared: PrepT) -> bool:
```
SandyLuXY, line 85: Small contract tightening suggestion: `publish()` currently only checks aborted, not whether the id is still inflight. The current MOSS callers do call `begin()` first, so this is not a happy-path regression, but as a shared helper it would be safer for `publish()` to fail closed unless it consumes an inflight id. That also avoids stale publishes after a context reset/reload. Could we add tests for publish without begin and publish after reset?

MelodyyyYin: Agreed — the aborted-only check let a publish after a context reset (or one that never began) leave a stale handoff; failing closed on non-inflight ids closes that.

— Yue Yin

sglang_omni/models/moss_tts/request_builders.py
```diff
@@ -473,17 +449,9 @@ def preprocess_moss_tts_payload(payload: StagePayload) -> StagePayload:
...
+        _QUEUE.fail_inflight(rid)
         raise
-    with _PREPARED_REQUESTS_LOCK:
-        _INFLIGHT_REQUESTS.discard(rid)
-        aborted = rid in _ABORTED_REQUESTS
-        _ABORTED_REQUESTS.discard(rid)
-        if not aborted:
-            # Aborted-while-preprocessing drops the handoff so it never lingers.
-            _PREPARED_REQUESTS[rid] = prepared
+    _QUEUE.publish(rid, prepared)
```
luojiaxuan: publish can now return False after context reset, but both callers still return a marker payload. Please handle False here and in local so AR does not later fail on missing prepared state.

sglang_omni/scheduling/prepared_request_queue.py
```diff
@@ -0,0 +1,98 @@
...
+from typing import Generic, TypeVar
+
+CtxT = TypeVar("CtxT")
+PrepT = TypeVar("PrepT")
+
+
+class PreparedRequestQueue(Generic[CtxT, PrepT]):
+    """Thread-safe tri-state handoff registry for preprocessing -> AR scheduler.
+
+    The public attributes (context / prepared / inflight / aborted
```
luojiaxuan: Can we keep this state private? Tests can use a small snapshot helper; exposing mutable sets and lock makes it easy to bypass the transition API.

luojiaxuan (commented):
Left two small comments.

sglang_omni/models/moss_tts/request_builders.py
MelodyyyYin: Both callers now skip the marker when publish returns False, so a dropped handoff (aborted mid-flight or a context reset) never reaches the AR stage as a marker for prepared state that no longer exists; the abort test asserts the marker is absent.

— Yue Yin

sglang_omni/scheduling/prepared_request_queue.py
MelodyyyYin: State is now private (underscore-prefixed) behind a read-only snapshot(); tests and introspection go through it, so nothing outside the queue can mutate the sets or take the lock to bypass the transition methods.

— Yue Yin

luojiaxuan (approved):
LGTM

PR #908, opened 2026-06-30
[Qwen3-Omni] Async decode (one-step lookahead) for the thinker text decode

JiaxinD:
## Motivation

The Qwen3-Omni thinker decode loop is host-bound: a per-step `next_token_ids.tolist()` sync is roughly 90% of decode host time. This PR extends the omni async-decode one-step lookahead (#590, Higgs) to the thinker, overlapping that sync with the next step's GPU forward. Result: about +6-10% thinker text-decode throughput at concurrency 8-32, bit-identical to the synchronous path for greedy generation. Async is default-on for text decode; speech / hidden-capture, logprob, and history- or seed-dependent sampling automatically route to the synchronous path (see the gate below).

## Design

`post_decode_launch` samples `next_token_ids` on GPU and issues a non-blocking D2H copy into a pinned ping-pong buffer; `post_decode_resolve` reads it one step later, overlapping the next forward. Async is on by default (parity with Higgs); `--decode-mode sync` disables it, and batches below `async_decode_min_batch_size` (default 2) take the sync fast path.

The one-step lag is bit-identical only when nothing in the batch reads the lagged generated history, a fixed RNG seed, or the per-forward hidden-state side channel. `ThinkerModelRunner.lookahead_eligible` routes such batches to the synchronous path:

```text
route to sync if any request:
  captures hidden states (speech / audio output)   _captured_aux_hidden_states side channel is overwritten by launch(N) before resolve(N-1) reads it -> talker gets mismatched hidden
  sets repetition / presence / frequency_penalty    penalize the generated history
  sets min_new_tokens                               suppress EOS by generated-token count
  sets sampling_seed                                the overrun sample shifts the RNG draw order
  requests return_logprob                           the lookahead sampler skips the base logprob-recording path
  sets logit_bias / custom_params                   conservative; not proven lag-safe
```

Audited all 25 `SamplingParams` fields; these are the only history- or seed-dependent ones. Grammar (regex/ebnf/json_schema/structural_tag) is inert in the thinker (`grammar_backend=None`); shape params (temperature/top_k/top_p/min_p) apply identically to both paths via the shared `forward_batch`; the rest are finish conditions or detok formatting. Audio-output (hidden-capture) and return_logprob are detected per request via `should_generate_audio_output` on the request's `_omni_data`, so a text-output request on a speech-enabled server still takes the async path (only the audio-output requests route to sync).

## Modifications

- Add `post_decode_launch` / `post_decode_resolve` / `lookahead_eligible` to `ThinkerModelRunner`; share a `_text_only_capture_guard` between `execute` and the async launch for hidden-capture parity.
- Gate audio-output (hidden-capture) and `return_logprob` requests to sync per batch via `should_generate_audio_output` on `_omni_data`, alongside the history/seed-dependent sampling params; text-output requests on a speech-enabled server stay async.
- Thread `enable_async_decode` (default on) and `async_decode_min_batch_size` through `config` factory_args -> thinker stage executor -> `create_thinker_scheduler`.
- Resolve the `--decode-mode` override target by factory so it reaches the `thinker` stage (previously `tts_engine` only).
- Add unit coverage for the gate (each gated class -> sync, plain greedy -> async) and for default-on / `--decode-mode sync`.

## Validation

Rebased onto `origin/main` (`038612ac`); unit tests 6/6.

**Bit-identical (text decode): async signature == sync signature, verified on this commit (`12145110`, 2026-07-02), greedy temp=0, deterministic triplicate. The async side is confirmed to actually take the lookahead (launch count), and the penalty case is confirmed routed to sync (0 launches):**

| edge case | async == sync signature | async lookahead launches |
|---|---|---|
| plain greedy | `15800438b1352786` | 1248 |
| natural EOS + overrun + drain to bs<2 | `a23e8df1db732d50` | 384 |
| repetition_penalty 1.3 (routed to sync) | `a6084c4f13ea02e6` | 0 (sync) |

Absolute signatures differ from the pre-rebase reference due to setup; the async == sync invariant is the claim.

**Per-batch gate on a speech-enabled server (only audio-output routed to sync, text recovered to async):** instrumented the `_captured_aux_hidden_states` write and collect read and ran an interleaved text(async) + speech(sync) workload; across 39 speech forwards and 11 text-async forwards, all 21 fresh speech collects read their own forward's tensor and 0 fresh reads returned `None` (a sync batch's forward and collect are sequential in one scheduler iteration, so a text `None` write cannot land between them). Recovered text-decode throughput, colocated FP8 speech server, async vs sync:

| conc | async tok/s | sync tok/s | delta |
|---|---:|---:|---:|
| 8 | 710 | 704 | +0.9% |
| 16 | 1334 | 1177 | +13.3% |

Speech E2E on the same server stays at parity (async ≈ sync, 0 failed at c=8/16). The `out_cache_loc=None` prefill crash is specific to audio-output/hidden-capture batches, not multimodal input (an image-input / text-output request runs async without it), so multimodal-input text requests also take the async path.

**Thinker text-decode throughput (the win), FP8, H100 80GB SXM5, greedy, median of 3:**

| concurrency | sync tok/s | async tok/s | delta |
|---|---:|---:|---:|
| 1 | 182 | 183 | +0.4% |
| 8 | 1203 | 1282 | +6.5% |
| 16 | 2231 | 2464 | +10.4% |
| 32 | 2263 | 2467 | +9.0% |

**Speech E2E perf, `benchmark_omni_rollout_stress.py` (colocated Qwen3-Omni FP8, `--enable-audio`, full talker path), async vs sync, concurrency sweep:**

| conc | success (async / sync) | output tok/s (async / sync) | throughput req/s (async / sync) | p50 / p95 / p99 latency s (async \| sync) |
|---|---|---|---|---|
| 8 | 8/8, 8/8 / 8/8, 8/8 | 55.8-77.5 / 58.4-77.0 | 0.23-0.31 / 0.23-0.31 | 24.6-24.7/25.8-33.0/26.0-35.0 \| 24.1-24.9/26.0-31.7/26.1-34.2 |
| 16 | 16/16 / 16/16 | 83.7 / 91.5 | 0.328 / 0.359 | 39.5/47.4/48.5 \| 37.0/42.9/44.1 |
| 32 | 32/32 / 32/32 | 90.5 / 96.1 | 0.357 / 0.376 | 55.4/76.2/85.7 \| 62.3/81.6/83.9 |

0 failed requests across every condition. Run-to-run variance is large on this shared box (two c=8 reps of the same condition span tok/s 55.8-77.5, with neighbor-tenant contention and 25-85 s per-request latencies), so the async-vs-sync deltas at c=16/32 sit inside that noise band; async and sync are on par with no regression.

**Speech E2E quality (rollout_stress has no WER, so seedtts covers it), Qwen3-ASR-1.7B, 48 samples, 3 runs each:**

| | WER corpus (3 runs) | evaluated | >50% outliers |
|---|---|---|---|
| sync | 2.03% / 1.50% / 1.29% | 48/48 | 0 / 1 / 0 |
| async | 1.48% / 0.93% / 1.69% | 48/48 | 0 / 1 / 1 |

The WER ranges overlap and both sit at the reference level (~2.14% on H200 full-set); async is not degraded, so hidden-state capture is not misaligned.

Mechanism: on the async server, speech / audio-output batches take the sync path via the gate, so parity is expected and is exactly what confirms the gate routes them correctly. Without the gate, async speech crashes with `out_cache_loc=None` in the multimodal prefill (a pre-existing branch issue in how the async event loop builds the multimodal ForwardBatch, independent of this PR); the gate keeps speech off that path. Enabling async for speech in the future would first need that prefill fixed.

## Default / rollback

Async decode is default-on for text decode (the +6-10% win, bit-identical). Speech / hidden-capture, `return_logprob`, and history/seed-dependent sampling automatically route to sync via `lookahead_eligible`; single-request (bs < 2) decode also uses the sync fast path. `--decode-mode sync` disables async entirely. Correctness is covered by the bit-identical gate plus both speech E2E suites above.

### DIFF
--- a/sglang_omni/model_runner/thinker_model_runner.py
+++ b/sglang_omni/model_runner/thinker_model_runner.py
@@ -325,26 +325,30 @@ class ThinkerModelRunner(ModelRunner):
         )
 
     def lookahead_eligible(self, batch: Any) -> bool:
-        """Route to sync where the one-step lag would diverge from sync. Speech /
-        hidden-capture batches: the captured-hidden side channel is per-forward, so
-        launch(N) overwrites it before resolve(N-1) reads it and the talker gets
-        mismatched hidden states. Sampling that reads the lagged output history
-        (repetition/presence/frequency penalty, min_new_tokens) or a fixed seed
-        also diverges; logit_bias / custom_params are routed conservatively.
-        return_logprob is routed to sync because the lookahead sampler skips the
-        base logprob-recording path.
+        """Route to sync where the one-step lag would diverge from sync. A request
+        that emits audio captures hidden states for the talker; the per-forward
+        _captured_aux_hidden_states side channel would be overwritten by a lookahead
+        launch(N) before resolve(N-1) collects it, so those requests route to sync
+        per batch. Sampling that reads the lagged output history (repetition /
+        presence / frequency penalty, min_new_tokens), a fixed seed, or
+        return_logprob (the lookahead sampler skips the base logprob path) also
+        diverges; logit_bias / custom_params are routed conservatively.
         """
-        # note (jiaxin deng): hidden-capture (speech/audio-output) batches must not
-        # use the lookahead; the _captured_aux_hidden_states side channel would be
-        # overwritten by launch(N) before resolve(N-1) collects it.
-        if self._text_model.layers_to_capture:
-            return False
+        from sglang_omni.models.qwen3_omni.request_builders import (
+            should_generate_audio_output,
+        )
+
         for req in batch.reqs:
-            # _omni_data is attached per request by OmniScheduler; guard the
-            # attribute access so a request data variant without the flag is
-            # simply treated as not requesting logprobs.
+            # note (jiaxin deng): fail closed if the request data is missing or None
+            # so a hidden-capture batch can never slip onto the async path.
             try:
-                needs_logprob = req._omni_data.return_logprob
+                data = req._omni_data
+            except AttributeError:
+                data = None
+            if data is None or should_generate_audio_output(data.stage_payload):
+                return False
+            try:
+                needs_logprob = data.return_logprob
             except AttributeError:
                 needs_logprob = False
             if needs_logprob:
--- a/sglang_omni/model_runner/thinker_model_runner.py
+++ b/sglang_omni/model_runner/thinker_model_runner.py
@@ -6,6 +6,7 @@ visual embeddings for Qwen3-Omni's thinker stage.
 """
 from __future__ import annotations
 
+import contextlib
 import logging
 from collections.abc import Callable
 from typing import Any
@@ -35,24 +36,36 @@ class ThinkerModelRunner(ModelRunner):
         self._outer_model = model.thinker
         self._text_model = self._outer_model.model
         self._embed_tokens = self._text_model.embed_tokens
+        self._th_host_bufs = None
+        self._th_slot = 0
 
         thinker_cfg = tp_worker.model_runner.model_config.hf_config.thinker_config
         self._image_token_id = thinker_cfg.image_token_id
         self._video_token_id = thinker_cfg.video_token_id
         self._audio_token_id = thinker_cfg.audio_token_id
 
+    @contextlib.contextmanager
+    def _text_only_capture_guard(self, requests: list[Any]):
+        # note (jiaxin deng): drop hidden-capture for an all-text batch, shared by
+        # sync execute() and async execute_launch so both take the same path.
+        capture_layers = self._text_model.layers_to_capture
+        if not (capture_layers and not self._batch_should_capture_hidden(requests)):
+            yield
+            return
+        saved_capture_layers = list(capture_layers)
+        self._text_model.layers_to_capture = []
+        try:
+            yield
+        finally:
+            self._text_model.layers_to_capture = saved_capture_layers
+
     def execute(self, scheduler_output: Any):
-        capture_layers = getattr(self._text_model, "layers_to_capture", None)
-        if capture_layers and not self._batch_should_capture_hidden(
-            scheduler_output.requests
-        ):
-            saved_capture_layers = list(capture_layers)
-            self._text_model.layers_to_capture = []
-            try:
-                return super().execute(scheduler_output)
-            finally:
-                self._text_model.layers_to_capture = saved_capture_layers
-        return super().execute(scheduler_output)
+        with self._text_only_capture_guard(scheduler_output.requests):
+            return super().execute(scheduler_output)
+
+    def execute_launch(self, scheduler_output: Any):
+        with self._text_only_capture_guard(scheduler_output.requests):
+            return super().execute_launch(scheduler_output)
 
     def _batch_should_capture_hidden(self, requests: list[Any]) -> bool:
         if self._should_capture_hidden is None:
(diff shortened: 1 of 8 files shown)


luojiaxuan:
Thanks for the PR. The thinker-only tok/s numbers look promising, but I do not think this should be default-on for all Qwen3-Omni thinker paths before a couple of correctness and validation gaps are covered.

Main concerns:

1. The speech/audio path may misalign hidden-state capture under launch-first async decode. Qwen3 speech thinker captures hidden states for the talker via the model-side `_captured_aux_hidden_states` side channel. In the async loop, step N is launched before step N-1 is resolved, and each forward can overwrite that side channel. So resolving N-1 may read hidden states from N. Pure text token parity would not catch this. Please either route hidden-capture / audio-output batches to sync, or snapshot the captured hidden states at launch and resolve from the pending step.

2. `return_logprob` looks bypassed in the async launch path. `post_decode_launch()` calls `_sample_lookahead()` directly, so it skips the base `_sample_next_token_ids()` logic that enables sampler logprobs and records `output_token_logprobs`. Since `/generate` defaults `return_logprob=True`, those batches should either route to sync or the logprob path needs to be made launch-safe.

Validation request: please add same-commit A/B results with `--decode-mode sync` vs async using `benchmarks/eval/benchmark_omni_rollout_stress.py`, not only raw tok/s. Please include throughput/output tok/s, p50/p95/p99 latency, and success rate across the concurrency sweep. Also please run at least one E2E speech/audio-output test with the talker enabled, to catch hidden-capture or audio-path regressions that thinker-only text metrics would miss.

sglang_omni/model_runner/thinker_model_runner.py
```diff
@@ -310,3 +323,67 @@ def _forward_with_omni_embeds(
...
+            self._th_slot = 0
+        buf = self._th_host_bufs[self._th_slot]
+        self._th_slot ^= 1
+        return buf
+
+    def _sample_lookahead(self, logits_output, forward_batch, requests):
+        # note (jiaxin deng): penalties never reach here (lookahead_eligible routes
+        # those batches to sync); only static suppress tokens are lag-safe.
+        self._apply_codec_suppress_tokens(logits_output, requests)
+        return self.tp_worker.model_runner.sample(logits_output, forward_batch)
```
yxs, line 386: Adding reproduction evidence for the two concerns raised above.

1. The `return_logprob` path is broken in the async branch here: `_sample_lookahead()` calls the sampler directly instead of the shared `_sample_next_token_ids()` helper, so it never enables sampler logprobs or appends `data.output_token_logprobs`. With a real Qwen3-Omni FP8 text server on h200a, 4 concurrent `/generate` requests using the default `return_logprob=True` all returned 500 because `output_token_logprobs` had length 1-2 while `completion_tokens` was 10. The same server with `seed=123`, which routes the batch to sync via `lookahead_eligible()`, returned 200 for all 4 requests with `logprobs_len == completion_tokens == 10`.

2. For speech/audio-output batches, `post_decode_launch()` only snapshots `next_token_ids`. The hidden-capture path still reads `model._captured_aux_hidden_states` later during resolve, but that side channel can be overwritten by the next launched forward before the pending step is processed. A CUDA contract probe with the real `SGLangOutputProcessor` reproduced step N token ids being paired with aux hidden states from step N+1.

I think `return_logprob` batches and hidden-capture/audio-output batches should route to sync unless/until these outputs are made launch-safe.

JiaxinD:
Thanks, both concerns are addressed in `ThinkerModelRunner.lookahead_eligible` (routes a batch to the synchronous path so the one-step lag can't diverge):

1. **Hidden-capture / audio-output** batches now route to sync (detected from the text model's active `layers_to_capture`), so the speech path never takes the launch-first lookahead and `_captured_aux_hidden_states` can't be read cross-step. **`return_logprob`** batches route to sync too, so `post_decode_launch` never bypasses the base logprob recording. Both reuse the sync-routing that already handles seeded sampling — the exact path yxs's h200a repro validated: default `return_logprob` under async returned 500, while `seed=123` routed to sync and returned 200. The fix sends the `return_logprob` and hidden-capture batches down that same verified sync path, so the 500 case now succeeds. Unit tests cover each gated class -> sync and plain greedy -> async.

2. **A/B with `benchmark_omni_rollout_stress.py`** (colocated Qwen3-Omni FP8, `--enable-audio`, full talker path), same commit, async vs sync:

| conc | success async/sync | output tok/s async/sync | req/s async/sync | p50/p95/p99 s async \| sync |
|---|---|---|---|---|
| 8 | 8/8 / 8/8 | 55.8-77.5 / 58.4-77.0 | 0.23-0.31 / 0.23-0.31 | 24.7/25.8/26.0 \| 24.9/26.0/26.1 |
| 16 | 16/16 / 16/16 | 83.7 / 91.5 | 0.328 / 0.359 | 39.5/47.4/48.5 \| 37.0/42.9/44.1 |
| 32 | 32/32 / 32/32 | 90.5 / 96.1 | 0.357 / 0.376 | 55.4/76.2/85.7 \| 62.3/81.6/83.9 |

0 failed requests everywhere; async and sync are on par. Run-to-run variance is large on this shared box (two same-condition c=8 reps span 55.8-77.5 tok/s, neighbor-tenant contention, 25-85 s per request), so the c=16/32 deltas sit inside that noise band.

3. rollout_stress has no WER, so quality parity is from a **seedtts E2E** run (Qwen3-ASR-1.7B, 48 samples, 3 runs each): WER async 1.48 / 0.93 / 1.69% vs sync 2.03 / 1.50 / 1.29% (overlapping, ~2% reference-level, 48/48 evaluated, 0-1 >50% outliers). No degradation -> hidden capture is not misaligned.

One more data point: without the gate, async speech crashes with `out_cache_loc=None` in the multimodal prefill (a pre-existing branch issue in how the async event loop builds the multimodal ForwardBatch, unrelated to this diff), so the gate is load-bearing beyond the misalignment.

Async stays on only for text-only decode; speech / logprob / history- or seed-dependent sampling route to sync unless/until the launch path is made launch-safe. Rebased onto `038612ac`, unit tests green. Would appreciate another look 🙏

zhaochenyang20:
fix lint bro 🐶

luojiaxuan:
Could we make this gate per-batch instead of checking only `layers_to_capture`?

As written, any speech-enabled thinker server has `layers_to_capture` set, so even text-only / no-audio batches are forced onto the sync path. I think the intended behavior is to route only actual hidden-capture / audio-output batches to sync, e.g. by also checking `_batch_should_capture_hidden(batch.reqs)`.

JiaxinD:
Good catch. The server-level gate did zero the async gain on any speech-enabled server, since every batch there has `layers_to_capture` set so all decode fell back to sync. Fixed by gating per batch.

**Granularity:** `lookahead_eligible` now routes to sync only the requests that actually emit audio (`should_generate_audio_output` on the per-request `_omni_data`), alongside the return_logprob / history- or seed-dependent sampling classes. Text-output requests on a speech server take the async path again. Mixed batches (text + audio request in one batch) route whole-batch to sync; unit tests cover each gated class.

**Hidden-capture safety, verified at runtime:** I instrumented the `_captured_aux_hidden_states` write (`_hidden_capture`) and the collect read (output_processor) and ran an interleaved text(async) + speech(sync) workload. Across 39 speech forwards and 11 text-async forwards, all 21 fresh speech collects read their own forward's tensor and 0 fresh reads returned None. A sync batch's forward and collect are sequential within one scheduler iteration, so a text batch's `None` write cannot land between them. Re-ran the three signature edge cases on this commit: async == sync bit-identical for all three (greedy / natural-EOS / penalty), with the async side confirmed to actually take the lookahead (launch counts 1248 / 384; the penalty case correctly routes to sync at 0). Absolute signatures shift with setup, so the claim is the async == sync invariant.

**On the crash you'd expect:** the `out_cache_loc=None` in the multimodal prefill is specific to audio-output / hidden-capture batches, not multimodal input. An image- or audio-input request with text output runs fine async (probed: 0/4 fail with the gate bypassed). So the gate keys on audio output, and multimodal-input text requests also get the speedup.

**Recovered text-decode throughput** (text output on the colocated speech FP8 server, async vs sync):

| conc | async tok/s | sync tok/s | delta |
|---|---|---|---|
| 8 | 710 | 704 | +0.9% |
| 16 | 1334 | 1177 | +13.3% |

Speech E2E stays at parity (async ≈ sync, 0 failed at c=8/16). Updated the PR, would appreciate another look 🙏

edwingao28 (commented):
(withdrawn)
sorry was testing with fable 5 and it went out of control 😂

luojiaxuan (approved):
LGTM, please wait for the remaining ASR CI before merging.

Issue #890, opened 2026-06-25
[Perf] Investigate Qwen3-ASR throughput gap: stage-1 standalone (~71/s) vs stage-2/3 post-TTS WER eval (~63/s)


zhaochenyang20:
In the TTS CI (PR #887, run 28197611964), the same `Qwen/Qwen3-ASR-1.7B` model transcribes audio in all three stages at the same concurrency (32) on the same hardware (H100 80GB), but:

- **stage-1** (standalone ASR correctness): **70.95 samples/s**
- **stage-2/3** (ASR speed block run after the TTS WER eval): **~62.9 samples/s**

https://github.com/sgl-project/sglang-omni/actions/runs/28197611964/job/83529192493
https://github.com/sgl-project/sglang-omni/actions/runs/28197611964/job/83529192608
https://github.com/sgl-project/sglang-omni/actions/runs/28197611964/job/83529192508

That's a ~12% throughput gap. Since every stage is ultimately doing ASR over SeedTTS-text utterances, the gap is bigger than I'd naively expect.

**This issue is primarily a request to investigate whether the gap is reasonable.** There are two acceptable outcomes:

1. The gap is a real, fixable artifact (config not propagating, co-residency pressure, etc.) → fix it and bring stage-2/3 back to stage-1 levels.
2. The gap is expected given the workload (shorter clips, GPU co-residency, etc.) → that's fine, but please **explain why** and leave the reasoning somewhere durable (a comment in the CI config / a short note in the perf docs) so the next person doesn't re-chase these 8 samples/s.

Either way, the deliverable is an explanation, not necessarily a speedup.

## Observed numbers

| metric | stage-1 (standalone) | stage-3 (post-TTS) |
|---|---|---|
| throughput (samples/s) | 70.95 | 62.92 |
| latency mean / p95 / p99 (s) | 0.449 / 0.606 / 0.710 | 0.504 / 0.735 / **1.337** |
| RTF mean | 0.097 | 0.119 |
| audio processed (s) | 5152.9 | 4788.6 |
| mean clip length (s) | 4.74 | ~4.40 |
| total time (s) | 15.33 | 17.29 |

stage-1 explicitly runs 2 ASR workers (least_request, splitting 575 / 577). stage-2 (non-streaming) shows the same family of numbers as stage-3.

## Things worth checking (each could explain "real" or "expected")

1. **Is the #885 async request-build path actually active in stage-2/3?**
[...]

PR #910, opened 2026-06-30
[CI] Add per-sample ASR diagnostics and audio artifacts to TTS CI

db-ol:
## Motivation

The TTS CI runs Qwen3-ASR across three stages but reports only aggregate ASR speed, and it deletes the generated audio after each stage. Issue #890 asks whether the stage-1 (reference audio) vs stage-2/3 (generated audio) throughput gap comes from a small number of bad generated clips. That cannot be checked today without the per-sample ASR timings or the audio, neither of which leaves the runner.

This is a temporary diagnostic to gather that data for #890. It does not change any threshold or workload. Once the cause is confirmed we plan to revert it or fold it into the actual fix, so it is not meant to live in the tree long term.

## Modifications

- `benchmarks/tasks/tts.py`: stage-2/3 transcription writes `asr_speed_per_sample.json` (per utterance `asr_latency_s`, `rtf`, `wer`, `audio_duration_s`, `is_success`) next to the existing aggregate `asr_speed_results.json`.
- `tests/test_model/test_qwen3_asr_ci.py`: stage-1 includes the `per_sample` rows it already computes in its results file, and writes the file under `QWEN3_ASR_RESULTS_DIR` when that variable is set. Behavior is unchanged when it is not.
- `tests/test_model/test_tts_ci.py`: the audio cleanup fixture keeps the generated audio in CI staging mode so the upload step can read it.
- `.github/workflows/test-tts-ci.yaml`: upload the per-sample JSONs, `wer_results.json`, `generated.json` and the generated audio as artifacts for all three stages. New uploads use `if-no-files-found: warn` so they never fail the run.

## Related Issues

Refs #890

## Accuracy Test

Not applicable. No model or kernel code is changed.

## Benchmark & Profiling

Not applicable. This only adds diagnostic artifacts and does not change the benchmark workload or thresholds.

## Checklist

- [ ] Add unit tests.

### DIFF
--- a/.github/workflows/test-tts-ci.yaml
+++ b/.github/workflows/test-tts-ci.yaml
@@ -62,6 +62,17 @@ jobs:
           HF_TOKEN: ${{ secrets.BOSON_HF_TOKEN }}
           QWEN3_ASR_MODEL_PATH: ${{ env.QWEN3_ASR_MODEL_PATH }}
           TORCHINDUCTOR_CACHE_DIR: ${{ env.OMNI_CI_HOME }}/.torchinductor
+          QWEN3_ASR_RESULTS_DIR: ${{ env.OMNI_CI_HOME }}/qwen3-asr-results
+
+      - name: Upload Qwen3-ASR per-sample artifact
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: tts-stage1-asr-per-sample
+          path: ${{ env.OMNI_CI_HOME }}/qwen3-asr-results/qwen3_asr_results.json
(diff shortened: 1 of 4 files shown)


db-ol:
Closing this. It was a diagnostic to gather data on whether the stage-1 vs stage-2/3 ASR throughput gap (#890) came from bad generated audio. That investigation is done and the answer is no. The real cause is a warmup timing asymmetry in the CI measurement (stage-1 excludes warmup requests from its timer, stage-2/3 does not), which I am fixing in a separate focused PR. Refs #890.

PR #911, opened 2026-07-01
[CI] Warm the TTS-stage ASR the same way as the standalone gate

db-ol:
## Motivation

The Qwen3-ASR speed reported by the standalone gate (stage-1) and by the post-TTS WER path (stage-2/3) are not directly comparable today, which made the ~12% gap in #890 look larger than it is. The two paths time different windows:

- stage-1 (`BenchmarkRunner`) runs `concurrency * 2` warmup requests that are excluded from its wall clock.
- stage-2/3 (`run_seedtts_transcribe`) has no warmup and times every request from the first one, so the cold start (CUDA graph capture, cache warmup) is counted inside the measured window and depresses the reported throughput.

## Modifications

- `benchmarks/tasks/tts.py`: `run_seedtts_transcribe` now runs the same warmup as stage-1 (`concurrency * 2` successful entries transcribed once and discarded) before starting the ASR timer, so both paths measure steady state. No threshold or workload change.

## Related Issues

Refs #890

## Accuracy Test

Not applicable. No model or kernel code changes, and WER is unaffected.

## Benchmark & Profiling

Before/after on a single RTX 4090 (in-process Qwen3-ASR, n=1088 generated clips, single stream):

| | measured throughput |
|---|---|
| before (no warmup, cold start inside timer) | 6.14 clips/s |
| after (warmup, steady state) | 6.17 clips/s |

First-clip cold latency was 0.90s vs a 0.16s steady state. On the 4090 the effect is small (+0.5%) because in-process Qwen3-ASR has a light cold start. On the H100 sglang router the cold start (CUDA graph capture during the first batch) is heavier, so the real effect is expected to be larger. The definitive before/after should come from the TTS CI on H100 (run the benchmark on main, then on this branch, and compare stage-2/3 ASR throughput).

This corrects the measurement so the two stages are comparable. It is expected to shrink but not fully close the ~12% gap. Any remainder is not per-clip ASR cost, which is the same for reference and generated audio when measured single stream, but a concurrency-level effect: the generated clips have higher length variance (duration CV about 0.26 vs 0.21), which lowers batch-packing efficiency under concurrency 32 and leaves the ASR server slightly under-utilized. That is a separate, benign workload characteristic and is not addressed here.

## Checklist

- [ ] Add unit tests.


Ccyest (approved):
lgtm

benchmarks/tasks/tts.py
```diff
@@ -911,6 +911,29 @@ def _transcribe_one_entry(
...
+    entries = [e for e in generated if e.get("is_success", False)][:count]
+    if not entries:
+        return
+    with concurrent.futures.ThreadPoolExecutor(max_workers=concurrency) as executor:
+        list(
+            executor.map(
+                lambda entry: _transcribe_one_entry(entry, asr, lang, device),
+                entries,
+            )
+        )
```
zhaochenyang20: In this way, CI should import this function. Not let them run seperate functions.

zhaochenyang20: CI is just a "threshold version" of `benchmarks/eval`.

zhaochenyang20 (changes requested):

benchmarks/tasks/tts.py
db-ol: Agreed, that's the right call. I'll drop `_warmup_asr` and route the ASR speed measurement in `run_seedtts_transcribe` through the shared `BenchmarkRunner` / `run_asr_transcription`, so warmup, concurrency, and timing all come from one place (and WER moves out of the timed window too, another current difference).

One scope check: `run_seedtts_transcribe` also serves Omni-Whisper, which transcribes in 30s chunks, while the shared path sends the whole clip in one request (Qwen3-ASR). I'd route the Qwen3-ASR case (what the TTS CI uses) through the shared path in this PR and unify Omni-Whisper separately, to keep this one small, but happy to do both here if you prefer. Which works for you?

zhaochenyang20: I think you only need to consider Qwen3 ASR. remove whisper is okay.

db-ol: Done. Dropped Omni-Whisper from `run_seedtts_transcribe`, so it now always uses the shared `run_asr_transcription` / `BenchmarkRunner` (Qwen3-ASR only), and removed the local transcribe loop and its helpers. A whisper model or a missing router port now raises a clear error. The Omni-Whisper backend functions stay in place for the consistency check and the voice clone evaluation, which still use them.

benchmarks/tasks/tts.py
```diff
@@ -971,62 +975,15 @@ def run_seedtts_transcribe(
...
-    )
-
-    tqdm_desc = (
-        f"Transcribing ({config.lang})" if not generation_mode else "WER transcribe"
-    )
+    if asr_router_port is None or _is_whisper_asr_model(asr_model_path):
+        raise ValueError(
+            "run_seedtts_transcribe now supports only Qwen3-ASR over an ASR router. "
+            "Pass asr_router_port and a Qwen3-ASR model."
+        )
```
zhaochenyang20: we should change this. Right now WER is computed by qwen3 ASR. But we will soon support new models. So do not add validation for the Qwen3-ASR mode. And, do not use `getattr` in sglang omni. pass in default parameter as Qwen3 ASR. And later it will be changed.

zhaochenyang20 (changes requested):

0xjeffro:
Co-authored-by: Jeffro Qu <0xjeffro@gmail.com>

benchmarks/tasks/tts.py
db-ol: Fixed in the latest commit. Removed the validation and both getattr calls. The model path now comes from config.asr_model_path, and the benchmark configs already default it to Qwen3 ASR, so new ASR models can be passed through later.

benchmarks/tasks/tts.py
```diff
@@ -888,62 +889,58 @@ def build_base_url(config: ServerEndpointConfig) -> str:
...
-    output.asr_latency_s = time.perf_counter() - asr_t0
-    return output
-
-
-def _resolve_asr_concurrency(config: SeedttsTranscribeConfig) -> int:
-    value = getattr(config, "asr_concurrency", DEFAULT_ASR_TRANSCRIBE_CONCURRENCY)
-    if value is None:
-        value = 1
-    return max(1, int(value))
+ASR_WARMUP_MULTIPLIER = 2
```
zhaochenyang20: move the const values to the beginning of file.

zhaochenyang20 (changes requested):

benchmarks/tasks/tts.py
```diff
@@ -888,62 +889,58 @@ def build_base_url(config: ServerEndpointConfig) -> str:
...
-            f"[{idx + 1}/{total}] Transcription failed: "
-            f"{entry['sample_id']} -- {output.error}",
+def _transcribe_generated_via_runner(
+    generated: list[dict],
+    router_port: int,
+    model_path: str,
+    lang: str,
+    concurrency: int,
+) -> tuple[list[SampleOutput], float]:
+    from benchmarks.eval.benchmark_qwen3_asr_concurrency import run_asr_transcription
```
zhaochenyang20: 1. import most of the subfunc at the beginning of file.

zhaochenyang20: when the TTS WER stage reaches `run_seedtts_transcribe`, the current code imports `run_asr_transcription` from `benchmarks.eval.benchmark_qwen3_asr_concurrency`; at the same time, the stage-1 benchmark file imports `SampleOutput`, `normalize_text`, and Qwen3 ASR constants from `benchmarks.tasks.tts`. This PR shares the runner timing window, but it still does not extract the real ASR/WER shared layer, so the ASR benchmark, TTS task, and CI threshold gate depend on each other in the wrong direction. Every future ASR model or WER CI stage will keep importing ASR primitives from a TTS module, the import names will not describe ownership, and WER scoring will remain split between `build_asr_eval_results` and `_apply_wer`; that maintenance risk grows with each new model and gate.

zhaochenyang20: Move the ASR/WER primitives out of `benchmarks/tasks/tts.py` into a new `benchmarks/tasks/asr.py`. At minimum, that module should own `SampleOutput`, `normalize_text`, Qwen3/Whisper ASR constants, the ASR router transcription send function, `run_asr_transcription`, the common WER apply/scoring helper, and the shareable parts of ASR speed/WER result assembly. Then make `benchmark_qwen3_asr_concurrency.py`, `tests/test_model/test_qwen3_asr_ci.py`, `tests/utils.py`, `run_seedtts_transcribe`, and the talker WER paths import `benchmarks.tasks.asr` instead of taking ASR logic from the TTS module. `benchmarks/eval/benchmark_qwen3_asr_concurrency.py` can remain the CLI/sweep wrapper, but it should not be the runtime shared library imported by the TTS task.

benchmarks/tasks/tts.py
zhaochenyang20, line 1: `run_seedtts_transcribe` still exposes `log_per_sample: bool = False`, but after this PR removed `_log_transcribe_result`, no code reads the parameter anymore. `benchmark_omni_seedtts.evaluate_generated_audio` still passes `log_per_sample=True`, so the caller expects the WER transcribe stage to print per-sample WER, ASR latency, and ref/hyp previews, but those diagnostics are now silently gone. Observability gets worse when CI fails, especially when debugging high-WER samples or ASR latency regressions in TTS stage 2/3; a parameter that exists but has no effect misleads the investigation.

zhaochenyang20 (changes requested):

benchmarks/tasks/tts.py
db-ol: Done. Added benchmarks/tasks/asr.py owning SampleOutput, normalize_text, the Qwen3/Whisper constants and backends, make_asr_send_fn, run_asr_transcription, apply_wer, and build_asr_eval_results (now scored through apply_wer, so the per-sample scoring is no longer split). The concurrency benchmark is a CLI wrapper again, and the gate, tests/utils.py, run_seedtts_transcribe, and the talker WER paths all import benchmarks.tasks.asr. With the circular import gone, tts.py imports everything at the top of the file.

benchmarks/tasks/tts.py
db-ol: Done. ASR_WARMUP_MULTIPLIER now sits with the other constants at the top of asr.py.

benchmarks/tasks/tts.py
db-ol: Good catch. Restored the per-sample diagnostics in the runner path. With log_per_sample=True each sample logs WER, ASR latency, and ref/hyp previews, and transcription failures still warn.

benchmarks/tasks/asr.py
```diff
@@ -0,0 +1,610 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Shared ASR task layer: transcription, WER scoring, and ASR speed assembly.
+
+Owns the ASR/WER primitives shared by the standalone ASR benchmark
+(``benchmarks/eval/benchmark_qwen3_asr_concurrency.py``), the Qwen3-ASR CI gate
+(``tests/test_model/test_qwen3_asr_ci.py``), the TTS WER stage
+(``benchmarks.tasks.tts.run_seedtts_transcribe``), and the talker WER paths.
+"""
```
zhaochenyang20, line 8: again, never use "`" in the coments, even in the beginning of file. Just like:

```python
Owns the ASR/WER primitives shared by the standalone ASR benchmark
(benchmarks/eval/benchmark_qwen3_asr_concurrency.py), the Qwen3-ASR CI gate
(tests/test_model/test_qwen3_asr_ci.py), the TTS WER stage
(benchmarks.tasks.tts.run_seedtts_transcribe), and the talker WER paths.
```

benchmarks/tasks/asr.py
```diff
@@ -0,0 +1,610 @@
...
+from benchmarks.dataset.seedtts import SampleInput
+from benchmarks.metrics.performance import compute_speed_metrics
+from benchmarks.metrics.wer import calculate_asr_speed_metrics, calculate_wer_metrics
+
+logger = logging.getLogger(__name__)
+
+OMNI_WHISPER_MODEL_PATH = "openai/whisper-large-v3"
+OMNI_WHISPER_REQUEST_TIMEOUT_S = 300
+# Whisper encoder accepts at most ~30 s per request (nb_max_frames=3000).
+# transformers pipeline uses chunk_length_s=30; mirror that for long talker audio.
```
zhaochenyang20: any comment should begin with:

`# note (your-name):`

benchmarks/tasks/asr.py
```diff
@@ -0,0 +1,610 @@
...
+OMNI_WHISPER_CHUNK_LENGTH_S = 30
+OMNI_WHISPER_CHUNK_STRIDE_S = 25
+OMNI_WHISPER_SAMPLE_RATE = 16000
+
+QWEN3_ASR_MODEL_PATH = os.getenv("QWEN3_ASR_MODEL_PATH", "Qwen/Qwen3-ASR-1.7B")
+QWEN3_ASR_REQUEST_TIMEOUT_S = 300
+QWEN3_ASR_MAX_NEW_TOKENS = int(os.getenv("QWEN3_ASR_MAX_NEW_TOKENS", "128"))
+# ASR transcription fan-out for WER, not TTS generation concurrency.
```
zhaochenyang20: any comment should begin with:

`# note (your-name):`

benchmarks/tasks/asr.py
```diff
@@ -0,0 +1,610 @@
...
+OMNI_WHISPER_REQUEST_TIMEOUT_S = 300
+# Whisper encoder accepts at most ~30 s per request (nb_max_frames=3000).
+# transformers pipeline uses chunk_length_s=30; mirror that for long talker audio.
+OMNI_WHISPER_CHUNK_LENGTH_S = 30
+OMNI_WHISPER_CHUNK_STRIDE_S = 25
+OMNI_WHISPER_SAMPLE_RATE = 16000
+
+QWEN3_ASR_MODEL_PATH = os.getenv("QWEN3_ASR_MODEL_PATH", "Qwen/Qwen3-ASR-1.7B")
+QWEN3_ASR_REQUEST_TIMEOUT_S = 300
+QWEN3_ASR_MAX_NEW_TOKENS = int(os.getenv("QWEN3_ASR_MAX_NEW_TOKENS", "128"))
```
zhaochenyang20: this parameter should be put in the config of qwen3 asr model. not in this file.

benchmarks/tasks/asr.py
```diff
@@ -0,0 +1,610 @@
...
+# ASR transcription fan-out for WER, not TTS generation concurrency.
+DEFAULT_ASR_TRANSCRIBE_CONCURRENCY = int(
```
zhaochenyang20: I highly dislike env args. Directly put it as: `DEFAULT_ASR_TRANSCRIBE_CONCURRENCY=32`.

benchmarks/tasks/asr.py
```diff
@@ -0,0 +1,610 @@
...
+    os.getenv("QWEN3_ASR_CONCURRENCY", os.getenv("SEEDTTS_ASR_CONCURRENCY", "32"))
+)
+# Warmup requests sent before the timed window, per unit of concurrency.
```
zhaochenyang20: any comment should begin with:

`# note (your-name):`

benchmarks/tasks/asr.py
```diff
@@ -0,0 +1,610 @@
...
+            ).strip()
+        )
+        if start + chunk_samples >= audio.shape[0]:
+            break
+    return " ".join(text for text in chunk_texts if text)
+
+
+def load_omni_whisper_asr(
+    router_port: int,
+    *,
```
zhaochenyang20: why should we accept `*` here?

benchmarks/tasks/asr.py
```diff
@@ -0,0 +1,610 @@
...
+    if asr_router_port is not None:
+        return load_router_asr(asr_router_port, model_path=asr_model_path)
+    return load_asr_model(lang, asr_device, generation_mode)
+
+
+def load_asr_model(lang: str, device: str, generation_mode: str | None = None):
+    """Legacy local ASR entry point.
+
+    WER now runs through SGLang Omni's OpenAI-compatible transcription endpoint.
+    Start an ASR server with ``sglang_omni.cli serve`` and pass its port instead
```
zhaochenyang20: again, never use "`" in the coments, even in the beginning of file. Just like:

```python
Owns the ASR/WER primitives shared by the standalone ASR benchmark
(benchmarks/eval/benchmark_qwen3_asr_concurrency.py), the Qwen3-ASR CI gate
(tests/test_model/test_qwen3_asr_ci.py), the TTS WER stage
(benchmarks.tasks.tts.run_seedtts_transcribe), and the talker WER paths.
```

zhaochenyang20 (changes requested):

benchmarks/metrics/wer.py
```diff
@@ -14,7 +14,7 @@
 )
 
 if TYPE_CHECKING:
-    from benchmarks.tasks.tts import SampleOutput
+    from benchmarks.tasks.asr import SampleOutput
```
zhaochenyang20: remove this `if TYPE_CHECKING:`

zhaochenyang20 (changes requested):

benchmarks/metrics/wer.py
zhaochenyang20: line 16

benchmarks/tasks/asr.py
db-ol: Done. Removed every backtick from comments and docstrings in the files this PR touches.

benchmarks/tasks/asr.py
db-ol: Done. Removed every backtick from comments and docstrings in the files this PR touches.

benchmarks/tasks/asr.py
db-ol: Done. All bare comments now start with # note (aaron):.

benchmarks/tasks/asr.py
db-ol: Done. All bare comments now start with # note (aaron):.

benchmarks/tasks/asr.py
db-ol: Done. All bare comments now start with # note (aaron):.

benchmarks/tasks/asr.py
db-ol: Done. Moved it into Qwen3ASRPipelineConfig factory_args (the stage factory already accepts max_new_tokens) and the WER clients no longer send it per request. Kept the value at 128, matching what the clients previously sent, so CI behavior is unchanged.

benchmarks/tasks/asr.py
db-ol: Done. DEFAULT_ASR_TRANSCRIBE_CONCURRENCY = 32 and QWEN3_ASR_MODEL_PATH are plain literals now.

benchmarks/tasks/asr.py
db-ol: No good reason, removed. The loaders take model_path as a normal default parameter now.

benchmarks/metrics/wer.py
db-ol: Done. A plain runtime import here would be circular (tasks/asr.py imports the WER calculators from this module), so SampleOutput moved down into benchmarks/metrics/wer.py instead. The metrics layer no longer reaches up into tasks, and the TYPE_CHECKING block and quoted annotations are gone.

zhaochenyang20 (approved):

zhaochenyang20:
Still seems gap:

6ae8eff4b4073974ea4a13734c35d224

f4ac897157df016b1d23bd7a19e3e6b9

https://github.com/sgl-project/sglang-omni/actions/runs/28608687974/job/84836732941?pr=911

https://github.com/sgl-project/sglang-omni/actions/runs/28608687974/job/84836733539?pr=911

zhaochenyang20:
Observed a noticeable regression in stage 10 latency on H100 recently — see this run: https://github.com/sgl-project/sglang-omni/actions/runs/28608687974/job/84839219775?pr=911

The current threshold of 155 was originally a single-run value picked by @MelodyyyYin , but latency now degrades to 180+ in recent runs.

Two hypotheses:

1. **A recent commit introduced a real regression** — either raw performance dropped, or variance of stage 10 on H100 increased significantly. A quick sanity check: run stage 10 ~10 times and report mean/std, so we can tell whether the distribution actually shifted or this is just noise.
2. **The 155 threshold was calibrated on H20 and never re-tuned for H100.** If that's the case, we should trace back the original calibration and figure out what went wrong there.

@db-ol could you also take a try after this PR?

db-ol:
Experiment plan for the remaining stage 1 vs stage 2/3 ASR speed gap (92.3 vs 82.9 samples/s in the latest CI run).

**Hypothesis.** With the harness now shared (same runner, warmup 64, concurrency 32, DP2 router), the residual gap comes from the corpora themselves. Stage 1 transcribes SeedTTS reference clips (5152.9s total, mean 4.74s per clip) while stage 2/3 transcribes MOSS generated clips (4788.4s total, mean 4.40s per clip, higher duration variance). Shorter clips carry a larger fixed per-request cost and higher variance packs batches less evenly, so samples/s is naturally lower. Earlier native-inference tests on a 4090 showed per-clip ASR cost is independent of audio quality (RTF vs WER correlation 0.065).

**Setup.** novita-h100 (same GPU SKU as CI), one Qwen3-ASR-1.7B router with 2 workers matching the CI topology, the shared run_asr_transcription runner, concurrency 32, warmup 64, full SeedTTS EN (1088 samples), 2 reps per condition.

**Conditions.**
1. Reference corpus, as in stage 1. Expected near 92 samples/s.
2. Generated corpus, MOSS-TTS clips for the same 1088 texts produced with the stage 2 recipe. Expected near 83 samples/s, reproducing the gap under an identical harness.
3. Duration-matched subsets. Pair each reference clip with a generated clip whose duration differs by less than 0.1s, matching on duration only (text not required to match), then run both subsets. If the gap collapses here, the duration distribution explains it.
4. Same-text cross-check. Per-pair latency for the same text in both corpora, confirming content itself adds no cost.

**Reported per condition.** Throughput (samples/s), latency mean/median/p95/p99, RTF, and corpus duration stats (mean, CV).

**Readout.** If 1 vs 2 reproduces the gap and 3 closes it, the two stages measure the same model on two datasets and their numbers should not be expected to match. Follow-up would be per-stage thresholds, documented with these results.

jzheng17:
Co-authored-by: jzheng17 [jzheng17@jhu.edu](mailto:jzheng17@jhu.edu)

db-ol:
Ran the corpus-swap experiment from the plan above. Full writeup below.

**Setup.** novita-h100, Qwen3-ASR-1.7B behind a 2-worker router (CI topology), the shared run_asr_transcription runner, concurrency 32, warmup 64. Every measured run got a freshly restarted router so no cache state carries across runs. 23 valid runs total. The generated corpus was produced with the stage 2 recipe (MOSS-TTS-Local-Transformer-v1.5, references, token_count auto, 1088/1088 succeeded) and its duration stats match the CI run almost exactly (total 4785.8s vs 4788.4s in CI).

**Corpora.**

| corpus | n | dur mean | dur CV | wav format |
|---|---|---|---|---|
| reference (stage 1) | 1088 | 4.69s | 0.21 | 24 kHz mono PCM_16, ~47 KiB/s |
| MOSS generated (stage 2/3) | 1088 | 4.40s | 0.26 | 48 kHz stereo PCM_16, ~187 KiB/s |

**Results (quiet machine, fresh server per run, mean of 2 reps).**

| condition | samples/s | vs ref |
|---|---|---|
| reference clips | 93.9 | baseline |
| generated clips | 85.6 | -8.9% (CI shows -10.2%) |
| generated clips re-encoded to 24 kHz mono | 91.4 | -2.7% |

**Findings.**

1. The gap reproduces under a byte-identical harness, so the harness and the serving path are exonerated.
2. The dominant cause is the wav container format. MOSS output is saved as 48 kHz stereo, so every generated clip uploads about 4x the bytes per audio second and the server pays the downmix and resample. Re-encoding the same clips to the reference format recovers about 70 percent of the gap.
3. Per-clip cost tracks the clip's own duration, not its content. In same-text pairs bucketed by generated-clip duration, the shortest quintile of generated clips is actually faster than its reference counterparts (0.288s vs 0.323s) and the longest is much slower (0.461s vs 0.330s). Correlation between per-pair latency delta and duration delta is 0.59.
4. Checked and excluded: decode length (generated audio carries fewer characters per audio second, 14.6 vs 15.5, so it cannot explain the residual), and stage-1 cache reuse (39 percent of stage-1 requests repeat one of 666 unique reference files, but repeats show no latency discount, ratio 0.99 to 1.00). After format normalization and duration matching a residual of roughly 3 percent remains, small, reproducible, and unexplained so far.
5. Side observation for the threshold discussion: one run degraded 40 percent when another user's job landed on a worker GPU mid-run, and background load moved batch means by 5 to 8 percent. Worst-of-N calibration on a shared machine needs to account for this.

**Conclusion.** Stage 1 and stage 2/3 measure the same ASR on materially different inputs, so their numbers should not be expected to match. Per-stage thresholds are the right call. Optionally, saving the WER wavs as 24 kHz mono in the eval pipeline would make stage 2/3 ASR about 7 percent faster, shrink the artifacts, and close most of the remaining gap. Happy to follow up with a PR carrying the scripts, a doc, and the mono 24 kHz wav change if the team wants it.

db-ol:
> Observed a noticeable regression in stage 10 latency on H100 recently — see this run: https://github.com/sgl-project/sglang-omni/actions/runs/28608687974/job/84839219775?pr=911
> 
> The current threshold of 155 was originally a single-run value picked by @MelodyyyYin , but latency now degrades to 180+ in recent runs.
> 
> Two hypotheses:
> 
> 1. **A recent commit introduced a real regression** — either raw performance dropped, or variance of stage 10 on H100 increased significantly. A quick sanity check: run stage 10 ~10 times and report mean/std, so we can tell whether the distribution actually shifted or this is just noise.
> 2. **The 155 threshold was calibrated on H20 and never re-tuned for H100.** If that's the case, we should trace back the original calibration and figure out what went wrong there.
> 
> @db-ol could you also take a try after this PR?

Happy to take this. Plan: run stage 10 ten times back to back on the H100 repro host with the machine otherwise quiet, and report mean/std/min/max of latency_mean_s.

One data point from the code: 155 was not an H20 leftover. It was pinned after the #765 mem-fraction 0.40 OOM fix, when the stage measured about 148s (see the note above VIDEOAMME_TALKER_TP2_THRESHOLDS), so the headroom is under 5 percent.

If the quiet-machine distribution sits near 148, the CI failures are co-tenant noise and the gate needs a proper worst-of-N margin. The corpus experiment above hit this directly, one run lost 40 percent the moment another job landed on a worker GPU. If the distribution sits at 160 or above, something after #765 regressed and I will bisect. Will sync with @MelodyyyYin to avoid duplicate runs.

jzheng17:
**Actually nvm, I read @db-ol's newest comments and experiment findings, the format finding basically settles it, and it also closes the batching-bubble angle I was raising.** 


Following up on the batch-packing point @db-ol made 2 comments earlier: I think there are 2 effects lumped together in "shorter clips carry a larger fixed per-request cost and higher variance packs batches less evenly," and because of that the plan for testing can show that this combined corpus matters but not which mechanism drives the gap.

Two distinct things:

1. Fixed per-request cost ∝ mean duration. Generated clips are shorter (4.40s vs 4.74s), so fixed per-request overhead is amortized over less audio. Order-independent.

2. Uneven packing ∝ variance — the actual "bubble." We dispatch with a saturated Semaphore(32) into the scheduler's continuous batching, so in-flight stays ~32 for most of the run. A pure "under-fill" therefore only shows up as tail drain at the end; the real variance cost is intra-batch padding as a continuous batch pads to its longest member, so higher duration-CV (0.26 vs 0.21) wastes compute even at full in-flight. Order-dependent.

Condition 3 (per-pair duration match within 0.1s) equalizes mean and variance at once, so if the gap closes there we still can't tell (1) from (2). I would suggest two cheap add-ons — no new datasets, no server changes — to separate them:

Proposed additions

1. Report realized in-flight concurrency for conditions 1 & 2: mean_in_flight = Σ(per-request latency) / wall_clock (both numbers are already in the runner logs): if ≈ 32 for both but generated slower → it's per-request cost / padding, not under-fill; elif generated < 32 and below reference → real under-fill (tail drain).
2. Add dispatch order as a variable. Run reference and generated each in native order and sorted by duration (today load_seedtts_samples dispatches in dataset order with no sort): if sorting shrinks the generated gap → intra-batch padding / uneven packing confirmed → fix is length-bucketed dispatch (or accept it and set per-stage thresholds); if sorting doesn't move it → the gap is fixed per-request cost, and reporting audio-seconds/s (or RTF) instead of samples/s would already make the two stages comparable.

Net: one number already in the logs, plus one A/B on dispatch order, tells us whether the residual is a batching bubble or just a shorter-clip amortization artifact — and each outcome points to a different fix.

Issue #907, opened 2026-06-30
Serving workloads are host/CPU-dispatch-bound on H100: implications for optimization direction


JiaxinD:
> Sharing results from a GPU profiling pass I ran after the H20 to H100 CI migration. The earlier memory/compute-bound guesses were reasonable from the indirect signal; direct measurement just paints a more specific picture, and I think it has real implications for where we spend optimization effort.

## 1. Background

After CI moved from H20 to H100, TTS/ASR got faster and Qwen3-Omni regressed slightly. The first read in the channel was that this points to memory-bound (Qwen3) vs compute-bound (TTS). I ran direct GPU measurements to check, on `novita-h100` (a single `NVIDIA H100 80GB HBM3 SXM5`, the same SKU as CI, 3.35 TB/s vs H20 4.0 TB/s, about 6.7x more BF16/FP8 compute). Each stage was run as a real single-card sglang-omni server at concurrency 16, with three independent methods:

- SM clock frequency sweep (causal test for compute-boundedness)
- nsys `--gpu-metrics` device-level hardware counters (SOL throughput %)
- CPU contention sweep (causal test for host-boundedness) + nsys CUDA timeline

## 2. Core finding: these serving workloads are host/CPU-dispatch-bound

The strongest result is a two-sided causal closure on ASR (direct evidence, not process of elimination):

- Cutting GPU compute (SM clock x0.455) drops throughput only 10% (sensitivity 0.10).
- nsys timeline shows the GPU is idle 94.3% of the time, waiting between tiny 3.2us kernels (the inter-kernel gap is about 17x the kernel duration).
- Cutting the server's host CPU to about 1/4 drops throughput 69% (sensitivity 0.69).

So cutting the GPU does almost nothing, the GPU is visibly idle most of the time, and cutting the host CPU directly throttles throughput. The bottleneck is host CPU dispatch, not the GPU.

This generalizes across stages in the device-level SOL data: Tensor Active is 0.6 to 4.9% everywhere (vs a BF16 matmul reference at 59.3%), so the H100's 6.7x tensor compute is essentially unused, and no stage saturates DRAM bandwidth either (max 42.9%, all well under saturation).

### Evidence table 1: device-level SOL per stage (nsys hardware counters, c16)

| Stage (H100 80GB SXM5) | Tensor% | DRAM R+W% | SM% | classification |
|---|---:|---:|---:|---|
[...]

PR #912, opened 2026-07-01
[Bench] Add same-card DP / server-throughput + GPU-idle profiling tooling

JiaxinD:
Measurement-only tooling to profile serving **throughput** and **GPU idle** for an omni
TTS pipeline under same-card data parallelism (DP=N replicas on one GPU), with/without CUDA
MPS, driven either by a manual client split or through the real `sglang-omni` router.
Goal: one shared measurement methodology so results are comparable across people instead of
every ad-hoc harness measuring a different thing.

**Draft**: shared for reproduction and a unified measurement methodology, not targeting merge.

## Scope (measurement only)
Adds `benchmarks/dp_throughput/` (scripts + README). No change to the router, models, or
serving path. Fitting >1 replica on one card needs a one-line local patch that is
deliberately NOT included here (documented in the README as a manual repro step).

## Methodology
- **Primary metric: device-level SM Active %** via `nsys --gpu-metrics` (whole physical GPU,
  cross-process comparable; GPU idle ~= 100 - SM%). Throughput is corroboration only, it is
  easy to mis-measure (batch size, audio length, co-tenant CPU, client artifacts).
- **Client control**: concurrency plateau + saturated in-window `#running-req` (queue>0) +
  continuous closed-loop load with one client process per replica (independent GILs).

## Results this tooling produced (Higgs TTS 4B, 1x H100; SM% primary, qps corroboration)
Throughput (qps), same-card DP, vs DP1:

| DP | manual, MPS off | manual, MPS on | via router, MPS on |
|----|-----------------|----------------|--------------------|
| DP1 | 62.3 | 62.3 | 62.3 |
| DP2 | 76.2 (+22%) | 105.9 (+70%) | 87.7 (+41%) |
| DP3 | 78.9 (+27%) | 133.3 (+114%) | 84.9 (+36%) |

- MPS is the unlock: without it same-card DP plateaus by DP2; with it, GPU idle drops
  70% -> 27% and DP3 reaches +114%. DP4 does not fit on 80 GB.
- The single-process router caps throughput at ~85 qps (its async event loop maxes one core),
  so DP3-via-router is no better than DP2-via-router: most of the MPS+DP gain is lost.

## notes
- MPS is an unprivileged per-user daemon (Default compute mode). Do not `kill -9` MPS clients
  (corrupts the server). router+MPS is unstable across back-to-back reps: the 2nd sustained
  rep tends to crash a worker with `cudaErrorMpsRpcFailure`.
- Rep counts: manual columns are 2 tight reps each; router+MPS is DP2 = mean of 2 clean reps,
  DP3 = 1 clean rep (the paired 2nd rep crashed under MPS). Single worker per replica;
  relative scaling is the result, not absolute qps.

## Reproduce
See `benchmarks/dp_throughput/README.md`. Example (DP=1..3, MPS on, 2 reps):
`MEM_FRACS=0.40,0.40,0.27 bash benchmarks/dp_throughput/scan_dp.sh 3 1 2`


db-ol:
Used this tooling to reproduce the same-card DP results on novita-h100 (GPU 5, H100 80GB SXM5), at eaf36ef with the README mem-fraction patch applied. Same methodology: Higgs TTS 4B voice clone, conc 96 per replica, 60s nsys windows at 10 kHz, MEM_FRACS 0.40,0.40,0.27, fresh MPS daemon per rep, 2 reps per condition (router DP3 is 1 rep). Client control passed (plateau by conc 96, running-req 62 of 64 with queue 31 in all manual windows).

Throughput (qps), vs DP1 of the same column:

| DP | manual, MPS off | manual, MPS on | via router, MPS on |
|----|-----------------|----------------|--------------------|
| DP1 | 64.9 | 62.8 | 62.8 |
| DP2 | 81.6 (+26%) | 110.5 (+76%) | 95.0 (+51%) |
| DP3 | 86.3 (+33%) | 147.2 (+134%) | 89.2 (+42%) |

All four findings from #907 hold on my run: the GPU is ~72% idle at single-replica saturation (SM ~28%), DP without MPS plateaus, MPS unlocks +134% at DP3 (idle 72% to 30%), and the router loses most of the gain (router DP3 lands below router DP2, workers underfed with running-req med dropping 62 to 45 to 25 and queue pinned at 0, router process at 88 to 96% of one core). Absolute qps is 5 to 10% above the original numbers, likely newer main and different host load, but every relative effect matches.

Three small notes from setting the tooling up, in case they help others reproduce:

1. hongccc/sglang-omni:dev needs msgpack and jiwer for current main.
2. The router launcher spawns workers via the sgl-omni entrypoint, which is missing when running the repo via PYTHONPATH without installing. A shim on PATH that execs `python -m sglang_omni.cli` fixes worker startup.
3. In router_bench/run_router_condition.sh, graceful_teardown stops MPS before sweeping the workers' multiprocessing.spawn children, so the MPS quit hangs on still-attached clients. Reordering it like scan_dp.sh (kill spawn children and GPU processes first, stop MPS last) made back-to-back router reps stable for me. I did not hit cudaErrorMpsRpcFailure in any rep after that change, so the ordering might be the source of the instability mentioned in the PR description.

Thanks @JiaxinD for the tooling, it made the repro straightforward.

PR #913, opened 2026-07-01
[Qwen3]: Defer non-streaming Code2Wav audio materialization

Ratish1:
## Motivation

This PR reduces CPU/GPU transfer work in the Qwen3-Omni Code2Wav scheduler for non-streaming TTS responses.

Before this change, each decoded Code2Wav window was immediately materialized as CPU float32 NumPy, even when the request was non-streaming and the client would only receive the final audio. The scheduler then concatenated those NumPy chunks at request completion.

After this change, streaming behavior is unchanged, but non-streaming requests keep decoded audio windows as detached torch tensors and build the final audio payload once at request completion.

## Modifications

Streaming requests are intentionally unchanged because streaming clients need each audio chunk immediately:

```text
decoded window tensor -> CPU float32 NumPy -> emit stream audio chunk
```

Non-streaming requests before:

```text
decoded window tensor -> CPU float32 NumPy per window -> np.concatenate -> payload
```

Non-streaming requests after:

```text
decoded window tensor -> keep detached tensor per window -> torch.cat -> payload
```

The change is a data-materialization cleanup. It does not change prompt construction, thinker generation, talker codec-token generation, vocoder inputs, window boundaries, left-context trimming, sample rate, or payload schema.

The final implementation also keeps the memory behavior bounded:

- Trimmed non-streaming tensor views are compacted before retention so the scheduler does not pin pre-trim backing storage until request finalization.
- Final tensor serialization copies to CPU before float32 conversion, avoiding a full fp32 temporary on GPU for fp16/bf16 Code2Wav outputs.

## Accuracy Tests

For the exact changed operation, an external local materialization probe held decoded vocoder tensors fixed and compared the old per-window NumPy materialization against the new final tensor materialization. It passed with byte-identical `audio_waveform` payloads for float32, float16, bfloat16, single-chunk, multi-chunk, and non-contiguous tensor cases.

Observed CUDA proof output:

```text
PASS float32_many_chunks: dtype=torch.float32 chunks=5 samples=4745
PASS float16_many_chunks: dtype=torch.float16 chunks=4 samples=2339
PASS bfloat16_many_chunks: dtype=torch.bfloat16 chunks=4 samples=4426
PASS float32_single_chunk: dtype=torch.float32 chunks=1 samples=8192
PASS float32_noncontiguous: dtype=torch.float32 chunks=3 samples=4369
PASS: old and new non-streaming Code2Wav payload bytes are identical
```

Focused unit result:

```text
1 passed
```

WER was evaluated with `Qwen/Qwen3-ASR-1.7B` on both generated output dirs.

| Metric | PR | Main `d7e62b0e` |
| --- | ---: | ---: |
| generated completed / failed | 1088 / 0 | 1088 / 0 |
| WER corpus | 2.11% | 2.65% |
| WER excl. >50% | 1.94% | 1.91% |
| >50% WER samples | 2 | 3 |
| WER sample mean | 2.28% | 2.86% |
| WER sample p95 | 14.29% | 13.04% |
| WER sample max | 214.29% | 842.86% |

ASR speed:

| Metric | PR | Main |
| --- | ---: | ---: |
| ASR evaluated / total | 1088 / 1088 | 1088 / 1088 |
| ASR skipped | 0 | 0 |
| ASR latency mean | 0.604s | 0.582s |
| ASR throughput | 52.75 samples/s | 54.78 samples/s |

## Speed Tests and Profiling

Full SeedTTS EN voice-clone A/B compared.

| Metric | PR | Main `d7e62b0e` | Delta |
| --- | ---: | ---: | ---: |
| completed / failed | 1088 / 0 | 1088 / 0 | same |
| throughput qps | 5.376 | 5.237 | +2.7% |
| rtf_mean | 0.8275 | 0.8410 | -1.6% |
| latency mean | 2.968s | 3.048s | -2.6% |
| latency p95 | 4.376s | 4.448s | -1.6% |
| latency p99 | 5.230s | 5.472s | -4.4% |
| audio throughput | 19.722 s/s | 19.537 s/s | +0.9% |
| output throughput | 76.6 tok/s | 74.6 tok/s | +2.7% |
| output tokens total | 15503 | 15500 | +3 tokens |

## Checklist

- [x] Format code according to the contribution guide.
- [ ] Update documentation as needed.

### DIFF
--- a/sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
+++ b/sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
@@ -173,11 +173,14 @@ class Code2WavScheduler(StreamingSimpleScheduler):
         full_audio = None if stream_enabled else self._concat_audio_parts(audio_parts)
         payload = self._payloads[request_id]
         if logger.isEnabledFor(logging.DEBUG):
+            emitted_samples = sum(self._audio_numel(part) for part in audio_parts)
             logger.debug(
-                "Code2Wav finalize req=%s code_chunks=%s audio_parts=%s final_samples=%s",
+                "Code2Wav finalize req=%s code_chunks=%s audio_parts=%s "
+                "emitted_samples=%s final_payload_samples=%s",
                 request_id,
                 len(self._code_chunks[request_id]),
                 len(audio_parts),
+                emitted_samples,
                 0 if full_audio is None else self._audio_numel(full_audio),
             )
         # Streaming clients already received per-chunk audio; final result is
--- a/sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
+++ b/sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
@@ -73,7 +73,7 @@ class Code2WavScheduler(StreamingSimpleScheduler):
         # Per-request state
         self._code_chunks: dict[str, list[torch.Tensor]] = {}
         self._emitted: dict[str, int] = {}
-        self._audio_chunks: dict[str, list[np.ndarray]] = {}
+        self._audio_chunks: dict[str, list[np.ndarray | torch.Tensor]] = {}
         self._stream_enabled: dict[str, bool] = {}
         super().__init__(compute_fn=None)
         self._payloads = self._stream_payloads
@@ -169,7 +169,8 @@ class Code2WavScheduler(StreamingSimpleScheduler):
                 RuntimeError(f"code2wav produced no audio for {request_id!r}"),
             )
             return []
-        full_audio = np.concatenate(audio_parts).astype(np.float32, copy=False)
+        stream_enabled = self._stream_enabled.get(request_id, False)
+        full_audio = None if stream_enabled else self._concat_audio_parts(audio_parts)
         payload = self._payloads[request_id]
         if logger.isEnabledFor(logging.DEBUG):
             logger.debug(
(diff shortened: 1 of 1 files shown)


luojiaxuan:
Thanks for the cleanup. I reviewed the Code2Wav path and do not see a correctness blocker.

One small non-blocking nit: in the streaming finalize path, `full_audio` is intentionally `None`, so the debug log now reports `final_samples=0` even though audio chunks were already emitted to the stream. Since the final streaming result is metadata-only this does not affect behavior, but it may be clearer to either log the already-emitted sample count for streaming requests or rename that field to something like final payload samples.

I also ran a one-shot H200 A/B with the same host/container/cache, SeedTTS EN full set, voice clone, non-streaming, concurrency 16:

| Metric | PR `2b30243` | Main `70b0077` | Delta |
| --- | ---: | ---: | ---: |
| completed / failed | 1088 / 0 | 1088 / 0 | same |
| throughput qps | 7.066 | 7.180 | -1.6% |
| rtf_mean | 0.6525 | 0.6402 | +1.9% |
| latency mean | 2.256s | 2.221s | +1.6% |
| latency p95 | 3.578s | 3.336s | +7.3% |
| latency p99 | 4.685s | 4.562s | +2.7% |
| audio throughput | 25.561 s/s | 25.966 s/s | -1.6% |
| output throughput | 100.7 tok/s | 102.3 tok/s | -1.6% |
| output tokens total | 15502 | 15503 | -1 token |

I did not reproduce the reported +2.7% QPS in this run; the PR was slightly slower here. Since this is a single run and the deltas are small, I would treat this as perf noise rather than a clear regression, but it would be worth either repeating once more or checking whether the baseline SHA/config differs from the numbers in the PR description.

Ratish1:
Closing this because the proposed materialization strategy has the wrong tradeoff for Code2Wav. Deferring CPU materialization keeps decoded audio tensors alive on GPU until request finalization and making that safe requires compacting trimmed tensor views, which adds per-window GPU copies and allocator pressure. So, that cost outweighs the removed per-window CPU NumPy materialization.

 The issue is that this design shifts work from CPU materialization into GPU memory lifetime / copy pressure on the active serving path, which was not a good fit for this scheduler.

PR #920, opened 2026-07-02
[router] Stream proxied responses instead of buffering (Phase 0 of #907 router optimization)

JiaxinD:
## Motivation
Part of the #907 router optimization. The single-process router saturates one CPU core; profiling shows the cap is two parts: ~73% fixed per-request overhead (asyncio + httpx + FastAPI, which caps ~820 qps even with trivial responses) plus ~27% byte-proportional relay at 300 KB (rising with audio length). This PR removes the byte-growth half by not buffering the response body; the fixed-overhead half is addressed separately by the multi-process work (Phase 1). The two are decoupled, so this lands on its own.

## Change
Merge `_forward_non_streaming` and `_forward_streaming` into one streaming `_forward_relay`, so the router no longer materializes the full (base64 audio) response body on its event loop. `proxy.py` only, +13/-75 (net -62 lines; also drops the now-unused `_response_error` helper).

## Results [all measured]
* 117/117 router unit tests pass (offline httpx `MockTransport` suite: opaque forwarding, failure/retryable-status paths, header stripping, `active_requests` lifecycle, capability routing, admin, health, SSE streaming).
* A/B at router-core saturation (100% CPU, mock upstream), buffered baseline vs this streaming change:

  | response size | throughput | peak RSS |
  |---|---|---|
  | 300 KB | +2.4% | -35% (108 -> 70 MB) |
  | 1.5 MB | +22% | -78% (519 -> 112 MB) |

  The throughput win grows with audio size (the buffer + join cost is size-proportional); the memory win is large. The one-shot mock understates the real streaming-worker benefit (TTFB and memory under load).

## Semantic changes (please sign off)
* A mid-stream upstream failure **after** a 2xx status now truncates the streamed response instead of returning 502. This is inherent to streaming (status and headers commit before the body starts). Connect-time and pre-body failures still return a clean 502 JSON, unchanged.
* Non-streaming responses now use chunked transfer-encoding (no `content-length` header). Transparent to standard HTTP clients; inherent to relaying a body of unknown length without buffering it.
* For a worker-failure status code, the recorded `worker.last_error` is now `status=NNN` instead of a response-body snippet, since the streaming path cannot read the body without consuming it. (No behavior change for callers; observability only.)

### DIFF
--- a/tests/unit_test/router/test_app.py
+++ b/tests/unit_test/router/test_app.py
@@ -1334,6 +1334,48 @@ def test_non_sse_midstream_failure_is_not_injected_with_error_event() -> None:
     assert all(worker.active_requests == 0 for worker in app.state.workers)
 
 
+def test_active_requests_held_across_body_relay_then_released() -> None:
+    # Note (Jiaxin Deng): recording active_requests at each yield proves the
+    # count is held while the body is still relaying (0 if released early), 0 after.
+    app_holder: list[FastAPI] = []
+    observed_during_relay: list[int] = []
+
+    class ObservingStream(httpx.AsyncByteStream):
+        async def __aiter__(self):
+            worker = app_holder[0].state.workers[0]
+            observed_during_relay.append(worker.active_requests)
+            yield b"chunk-1"
+            observed_during_relay.append(worker.active_requests)
+            yield b"chunk-2"
+
+    def handler(request: httpx.Request) -> httpx.Response:
+        if request.url.path == "/health":
+            return httpx.Response(200, json={"status": "healthy"}, request=request)
+        if request.url.path == "/v1/audio/speech":
+            return httpx.Response(
+                200,
+                stream=ObservingStream(),
+                headers={"content-type": "audio/wav"},
+                request=request,
+            )
+        raise AssertionError(f"unexpected request path: {request.url.path}")
+
+    async_client = httpx.AsyncClient(transport=httpx.MockTransport(handler))
+    app = create_app(
+        _router_config(worker_configs=[WorkerConfig(url="http://worker-a:8101")]),
+        client=async_client,
+    )
+    app_holder.append(app)
+
+    with TestClient(app) as client:
+        response = client.post("/v1/audio/speech", json={"model": "qwen3-omni"})
+
+    assert response.status_code == 200
+    assert response.content == b"chunk-1chunk-2"
+    assert observed_during_relay == [1, 1]
+    assert all(worker.active_requests == 0 for worker in app.state.workers)
+
+
 def test_least_request_avoids_worker_with_active_stream_load() -> None:
     def handler(request: httpx.Request) -> httpx.Response:
         if request.url.path == "/health":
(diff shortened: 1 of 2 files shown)


zhaochenyang20:
This is excellent work.

On the two semantic changes: approved, let's follow standard nginx/Envoy behavior. WDYT? @Ratish1

Ratish1 (approved):
LGTM. Is there any unit tests we need to change/add @JiaxinD

JiaxinD:
Good call. Added tests for the new streaming semantics on this branch: a mid-stream upstream failure after a 2xx truncates (not a 502) with active_requests cleanup; connect/pre-response failures still return the router 502 (also pinned by the existing `test_upstream_request_failure_returns_502_and_cleans_active_count`); `worker.last_error` records `status=NNN`; and the relayed response has no `content-length` header.

One update to the boundary we pinned last round: for `text/event-stream` responses specifically, a mid-stream upstream failure now emits a terminal SSE error event (matching the worker's own streaming-error format) and ends cleanly, instead of a raw truncation. Non-SSE bodies (JSON/audio) still truncate as before: injecting SSE text into a binary stream would be worse than truncating. The two existing streaming tests were updated to the new SSE behavior (cleanup/failure-recording assertions kept); truncation tests for non-SSE are unchanged. Since this changes behavior after your approval, please re-review the appended commits.

Separately, testing surfaced a pre-existing in-flight counter leak on main (`aclose()` runs before `decrement_active()` in the streaming cleanup, so a broken mid-stream connection leaks the count and drifts `least_request`). Split to PR #931.

123 router tests pass (117 existing, 2 updated to the new SSE behavior, plus 6 new); the new-behavior tests fail on the pre-change code (failing-before / passing-after).

tests/unit_test/router/test_app.py
```diff
@@ -1125,6 +1125,215 @@ def handler(request: httpx.Request) -> httpx.Response:
...
+                "/v1/audio/speech",
+                json={"model": "qwen3-omni"},
+            ) as response:
+                assert response.status_code == 200
+                b"".join(response.iter_bytes())
+
+    assert all(worker.active_requests == 0 for worker in app.state.workers)
+
+
 def test_least_request_avoids_worker_with_active_stream_load() -> None:
```
Ratish1, line 1379: Should we add one direct lifecycle test that proves `active_requests` stays
incremented while the downstream response body is still open, then returns to
zero after the body is consumed or the stream context exits.

The existing least-request test manually seeds `workers[0].active_requests =
10`, and the error tests assert cleanup after a response is done. Those are
useful, but they do not catch a regression where `_forward_relay()` increments
and then decrements before the body has actually been relayed. Since
`least_request` uses `active_requests` directly, this is the key runtime
invariant for long audio/SSE relays.

JiaxinD: Added `test_active_requests_held_across_body_relay_then_released`: the upstream stream samples `active_requests` at each yield, so it fails unless the count stays 1 for the whole relay and returns to 0 after. Verified it catches a decrement moved before the body relay.

PR #923, opened 2026-07-02
[TTS Refactor][M2] TTS engine factory: Extract shared TTS engine builder

SandyLuXY:
## Motivation
This PR implements the [M2 TtsEngineBuilder](https://ljp2lkmtlez9.jp.larksuite.com/wiki/GjBpwdJE0ie0JMkMlmzjdceDphf?from=from_copylink) refactor as described in [SGLang-Omni-TTS-Refactor-RFC-v2](https://osgbw74w8zwb.sg.larksuite.com/docx/L3XId4GHYoJqSKxwXYqlXaBrgNd?from=from_copylink).

Six TTS AR engine factory paths repeated the same SGLang bootstrap sequence while mixing model-specific setup into each `stages.py` file. This PR centralizes the invariant engine construction order in a shared builder and keeps model policy inside each model directory.

- Preserve the public stage factory entry points used by pipeline configs.
- Make CUDA graph timing, server-arg construction, batch-policy validation, output processor creation, scheduler assembly, and post-scheduler wiring follow one reviewed order.
- Reduce future conflict risk in `stages.py` by moving AR-engine bootstrap mechanics into model-local `engine_builder.py` files.

## Modifications

1. `sglang_omni/scheduling/engine_factory.py`
   - Adds `TtsEngineBuilder`, a template builder for TTS SGLang AR engine stages.
   - Normalizes `device` / `gpu_id`, resolves checkpoints, builds generation-batch overrides, constructs `ServerArgs`, and creates SGLang infrastructure through the deferred CUDA graph path.
   - Runs model setup before generation-batch validation, optional model-local compile, CUDA graph capture, output processor creation, adapter creation, scheduler construction, and post-scheduler hooks.
   - Exposes hooks for model-specific checkpoint resolution, defaults, override adjustment, server-arg customization, model setup, buffer sizing, compile behavior, extra graph setup, adapters, abort callbacks, scheduler specialization, and stream-outbox binding.

2. Model-local engine builders
   - `qwen3_tts/engine_builder.py`: preserves Qwen compatibility patches, HF config registration, speech-tokenizer loading, `Qwen3TTSModel` wrapper setup, preprocessing context, local backbone compile, and request cleanup.
   - `moss_tts/engine_builder.py`: preserves MOSS-TTS delay defaults, `MossTTSDelaySGLangModel` override, scheduler adapters, and prepared-request cleanup.
   - `moss_tts_local/engine_builder.py`: preserves colocated AR memory budgeting, process-memory fallback, frame decode graph capture, async decode kwargs, composite abort cleanup, and stream outbox binding.
   - `higgs_tts/engine_builder.py`: preserves Higgs defaults, rope cache truncation, buffer-size validation, max-new-token adapter cap, async decode kwargs, reset callback, and stream outbox binding.
   - `fishaudio_s2_pro/engine_builder.py`: preserves Fish config patching, audio decoder loading, tokenizer adapter wiring, local codebook-decoder compile, buffer-size validation, and `FishScheduler` construction.
   - `voxtral_tts/pipeline/engine_builder.py`: preserves decrypted SGLang config generation, inductor GEMM autotune, voice embedding loading, Voxtral adapters, and native SGLang compile/graph behavior.

3. TTS stage factories
   - Replaces the inline AR-engine bootstrap bodies in FishAudio S2-Pro, Higgs TTS, MOSS-TTS, MOSS-TTS Local, Qwen3-TTS, and Voxtral TTS with thin delegation to model-local builders.
   - Keeps the existing factory function names and signatures stable for config dotted-path resolution.
   - Moves Higgs onto the shared deferred CUDA graph bootstrap sequence so graph capture happens after model setup, matching the builder-owned ordering used by the other migrated TTS engines.
   - Higgs graph-capture safety: `truncate_rope_to_bf16` still runs before CUDA graph capture and only bf16-rounds the existing fp32 RoPE cache in place via `copy_`, preserving tensor dtype, shape, and stride for the captured graph.

4. Tests and docs
   - Adds `tests/unit_test/scheduling/test_engine_factory.py` to verify CPU-only import, builder phase order, override propagation, deferred graph restoration, scheduler customization, and post-scheduler wiring.
   - Updates per-model factory tests to monkeypatch the new builder-path module seams while preserving existing assertions for graph buckets, compile flags, memory budget propagation, scheduler construction, and adapter wiring.
   - Updates `tests/README.md` to include the new `unit_test/scheduling/` directory.

## Related Issues
#661 

## Accuracy Test

**Setup:**

- Benchmark: `benchmarks.eval.benchmark_tts_seedtts`, SeedTTS EN full set, non-streaming, `warmup=1`, `concurrency=16`, `request_rate=inf`, ASR `Qwen/Qwen3-ASR-1.7B` with `asr_concurrency=32`.
- Hardware: H100 80GB
- `S2-Pro`, `Voxtral-4B-TTS`, `MOSS-TTS`, `MOSS-TTS Local`, `Qwen3-TTS`, and `Higgs TTS`: baseline `main @ 966ddcf6`, candidate `feat/tts-engine-factory-refactor @ a96d8ba`

| Model | Baseline WER | Candidate WER | Delta |
| --- | ---: | ---: | --- |
| `S2-Pro` | 1.14% | 1.12% | -0.017 pp |
| `Voxtral-4B-TTS` | 1.42% | 1.38% | -0.042 pp | 
| `Higgs TTS` | 1.62% | 1.44% | -0.18 pp | 
| `MOSS-TTS` | 1.76% | 1.47% | -0.293 pp | 
| `MOSS-TTS Local` | 2.23% | 2.65% | +0.427 pp | 
| `Qwen3-TTS` | 1.81% | 1.67% | -0.14 pp | 

## Benchmark & Profiling

**Setup:** Same benchmark setup and artifacts as described in the _Accuracy Test_ section. The table below reports generation-phase metrics. 

| Model | Throughput QPS | Audio throughput | Mean latency | Mean RTF |
| `S2-Pro` | 1.206 -> 1.222 (+1.33%) | 4.793 -> 4.859 (+1.38%) | 13.193 -> 13.004 (-1.43%) | 3.5855 -> 3.5256 (-1.67%) |
| `Voxtral-4B-TTS` | 11.566 -> 11.281 (-2.46%) | 66.299 -> 64.731 (-2.37%) | 1.371 -> 1.407 (+2.63%) | 0.2417 -> 0.2473 (+2.32%) |
| `Higgs TTS` | 13.675 -> 13.542 (-0.97%) | 61.468 -> 63.421 (+3.18%) | 1.083 -> 1.114 (+2.86%) | 0.2488 -> 0.2481 (-0.28%) |
| `MOSS-TTS` | 2.229 -> 2.273 (+1.97%) | 9.987 -> 10.180 (+1.93%) | 7.153 -> 6.986 (-2.33%) | 1.6667 -> 1.6242 (-2.55%) |
| `MOSS-TTS Local` | 7.849 -> 7.970 (+1.54%) | 34.548 -> 35.076 (+1.53%) | 2.033 -> 2.001 (-1.57%) | 0.4850 -> 0.4757 (-1.92%) |
| `Qwen3-TTS` | 3.623 -> 3.583 (-1.10%) | 17.931 -> 17.156 (-4.32%) | 3.526 -> 3.550 (+0.68%) | 0.7355 -> 0.7641 (+3.89%) |

## Checklist
- [x] Add unit tests.

## CI


sglang_omni/scheduling/engine_factory.py
```diff
@@ -0,0 +1,217 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def make_adapters(self, model: Any) -> tuple[Any, Any]:
+        raise NotImplementedError
+
+    def make_abort_callback(self) -> Any | None:
+        return None
+
+    def make_scheduler(
```
JiaxinD, line 180: Consider letting the base `make_scheduler` accept extra scheduler kwargs. Higgs and MOSS-Local duplicate the whole `OmniScheduler(...)` call just to pass async-decode options, and a subclass that overrides this and forgets `abort_callback=self.make_abort_callback()` would silently lose the abort path.

sglang_omni/models/qwen3_tts/engine_builder.py
```diff
@@ -0,0 +1,106 @@
...
+    model_arch_override = "Qwen3TTSTalker"
+
+    def __init__(self, *, attn_implementation: str | None = None) -> None:
+        self.attn_implementation = attn_implementation
+        self.wrapper: Any | None = None
+
+    def resolve_checkpoint(self, model_path: str) -> str:
+        return qwen3_stages._resolve_checkpoint(model_path)
+
+    def pre_infra_setup(self, checkpoint_dir: str) -> None:
```
JiaxinD, line 31: Missing `qwen-tts` is now detected here, after `resolve_checkpoint` has run in `build()`, so an HF model id downloads the full checkpoint before failing (and the old bare `ImportError` became `RuntimeError`). Could probe the import before resolving the checkpoint.

sglang_omni/models/moss_tts_local/engine_builder.py
```diff
@@ -0,0 +1,171 @@
...
+        checkpoint_dir: str,
+        device: str,
+        gpu_id: int,
+        server_args: Any,
+    ) -> None:
+        del checkpoint_dir, device, gpu_id, server_args
+        self.model = model_worker.model_runner.model
+
+    def post_cuda_graph_setup(self, model: Any, server_args: Any) -> None:
+        model.init_frame_decode_graphs(list(server_args.cuda_graph_bs))
```
JiaxinD, line 119: A couple of attributed gotcha comments got dropped in the move: the frame-decode-graph rationale here and luojiaxuan's radix-cache namespacing note in Higgs. Please carry them into the builders.

JiaxinD (commented):
M2 looks good overall, no blockers. Verified the six-model migration line-by-line against the old inline bootstrap (independent review passes): moss_tts, fish, voxtral, and moss_local come out behavior-equivalent, the public factory signatures are preserved, and the tests are only strengthened. A few inline items, plus two notes:

* WER table: MOSS-TTS Local is the only regression (2.23 to 2.65). This model's migration is line-for-line equivalent, so it is very likely single-round SeedTTS noise; a rerun should close it.
* Higgs deferred-graph move: the safety argument (`truncate_rope_to_bf16` is in-place, no dtype/shape/stride change) is worth stating in the PR body instead of leaning on the WER row, which is within noise.

sglang_omni/scheduling/engine_factory.py
SandyLuXY: Agreed. This should be a base-builder extension point instead of duplicating the whole scheduler constructor. I added an `extra_scheduler_kwargs()` hook, had the base `make_scheduler()` preserve `abort_callback=self.make_abort_callback()`, and moved the Higgs/MOSS-Local async-decode args into that hook with a unit test covering both the extra kwargs and abort callback.

sglang_omni/models/qwen3_tts/engine_builder.py
SandyLuXY: I added a Qwen preflight import before `_resolve_checkpoint()` so missing `qwen-tts` fails immediately for remote model ids, and added a unit test that proves `_resolve_checkpoint()` is not called on that failure path.

sglang_omni/models/moss_tts_local/engine_builder.py
SandyLuXY: Good catch! Restored the MOSS-Local frame-decode graph rationale next to `init_frame_decode_graphs()` and the Higgs radix-cache `Req.extra_key` namespacing note next to the builder defaults.

JiaxinD (approved):

sglang_omni/scheduling/engine_factory.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def build(
+        self,
+        model_path: str,
+        *,
+        device: str = "cuda:0",
+        gpu_id: int | None = None,
+        dtype: str = "bfloat16",
+        server_args_overrides: dict[str, Any] | None = None,
+        **model_kwargs: Any,
```
Ratish1: should we trim the builder hook contract here. **model_kwargs is not used by any .build() call, and server_args_overrides is forwarded into generation_defaults() even though override merging already happens in build_generation_batch_overrides().

the base builder owns override merging, while subclasses own model defaults/policy. The concrete builders already receive real model-specific knobs through their constructors, we could remove **model_kwargs, make generation_defaults() take only the values it needs, and keep adjust_overrides(overrides) narrow.

sglang_omni/models/higgs_tts/engine_builder.py
```diff
@@ -0,0 +1,99 @@
...
+        return model_runner_mod.HiggsTTSModelRunner(model_worker, output_proc)
+
+    def make_adapters(self, model: Any) -> tuple[Any, Any]:
+        return request_builders.make_higgs_scheduler_adapters(
+            model,
+            max_new_tokens_cap=self.max_new_tokens,
+        )
+
+    def make_abort_callback(self) -> Any | None:
+        if self.model is None:
```
Ratish1: we could make this fail fast instead of returning None.

In the shared builder order, setup_model() must have set self.model before scheduler construction. If it has not, the builder lifecycle is broken; silently creating a scheduler without the model abort callback would leak model-owned request state on abort. Use assert self.model is not None and return self.model.reset_request.

sglang_omni/models/moss_tts_local/engine_builder.py
```diff
@@ -0,0 +1,147 @@
...
+
+        return model_runner_mod.MossTTSLocalModelRunner(model_worker, output_proc)
+
+    def make_adapters(self, model: Any) -> tuple[Any, Any]:
+        return request_builders.make_moss_tts_local_scheduler_adapters(model=model)
+
+    def make_abort_callback(self) -> Any | None:
+        def abort_request(request_id: str) -> None:
+            request_builders.cleanup_prepared_moss_tts_local_request(request_id)
+            if self.model is not None:
```
Ratish1: same

Ratish1 (commented):
Rest looks good. Just few small comments

SandyLuXY:
@Ratish1 Thanks! Addressed the comments above: narrowed the builder hook contract, made the Higgs abort callback fail fast, and applied the same fail-fast/full-cleanup behavior to MOSS-TTS Local. Targeted tests are updated.

Issue #917, opened 2026-07-02
[CI] Stage 10 Seems to be corrupted


zhaochenyang20:
Observed a noticeable regression in stage 10 latency on H100 recently — see this run: https://github.com/sgl-project/sglang-omni/actions/runs/28608687974/job/84839219775?pr=911

The current threshold of 155 was originally a single-run value picked by @MelodyyyYin , but latency now degrades to 180+ in recent runs.

Two hypotheses:

1. **A recent commit introduced a real regression** — either raw performance dropped, or variance of stage 10 on H100 increased significantly. A quick sanity check: run stage 10 ~10 times and report mean/std, so we can tell whether the distribution actually shifted or this is just noise.
2. **The 155 threshold was calibrated on H20 and never re-tuned for H100.** If that's the case, we should trace back the original calibration and figure out what went wrong there.

@MelodyyyYin  @db-ol could you also take a try after this #911 PR?

_Originally posted by @zhaochenyang20 in https://github.com/sgl-project/sglang-omni/issues/911#issuecomment-4869169835_

MelodyyyYin:
Investigated + opened a fix: #929.

### Root cause — it's both hypotheses, and it's mostly systematic (not just noise)
The gate is **tighter than its own calibration base**. In the test:
```python
_VIDEOAMME_TALKER_TP2_AUDIO_P95[16]["latency_mean_s"] = 170.36   # P95 base
VIDEOAMME_TALKER_TP2_THRESHOLDS = apply_slack(...)                # x1.125 -> 191.7 gate
VIDEOAMME_TALKER_TP2_THRESHOLDS[16]["latency_mean_s_max"] = 155   # ...then capped BELOW the base
```
So a run at the calibrated ~170s slow path **deterministically** exceeds 155 — not flaky. And `git log -L` shows the P95 base rose from the old ~130s era (146.7 gate) to **170.36**, i.e. a real slowdown was calibrated *into the base* while the `155` override stayed tight. So **H1 (a real regression got baked into the base) and H2 (the 155 override was never re-tuned) are both true.**

Separately, **#808 (lazy TensorRef; merged, default-off) was not enabled in CI** — nothing sets `SGLANG_OMNI_ENABLE_TENSOR_REFS` in `tests/`/`.github/`/the image — so stage-10 ran the ~170s slow path.

### Fix (#929)
[...]

PR #804, opened 2026-06-16
[CI Perf] Reduce stage 11 video talker payload

luojiaxuan:
## Summary
- trim Stage 11 Video-AMME TP2 video payload with both `video_min_pixels=6272` and `video_max_pixels=6272`; max-only did not reduce prompt tokens because the upstream video min-pixel clamp still applied
- keep H100/H20-safe FP8 TP2 startup args: lower thinker/talker mem fractions, thinker CUDA graph off, partial-start min chunks 3
- add `extra_request_params` plumbing for video benchmarks so Stage 11 sends `talker_prefill_user_context=false` and PR3 can avoid replaying large multimodal prompt embeds in talker prefill
- keep the existing accuracy/WER/speed threshold semantics; no warmup or threshold tuning

## Validation
- `CUDA_HOME=/usr/local/cuda-13.0 /data/sglang-omni-codex/container-sweep-env/omni/bin/python -m pytest tests/unit_test/qwen3_omni/test_stage11_payload_gate.py tests/unit_test/benchmarks/test_video_understanding.py tests/unit_test/qwen3_omni/test_example_launcher.py -q`
  - `25 passed in 3.63s`
- Combined PR2+PR3 full Stage 11 on H100 2x80GB, GPUs 2,3, `video_min_pixels=6272`, `video_max_pixels=6272`, `talker_prefill_user_context=false`:
  - `3 passed in 155.34s (0:02:35)`
  - accuracy `0.6000`, failed requests `0`, latency mean `46.912s`, RTF mean `3.3461`, prompt tokens mean `1132`, throughput `0.209 req/s`
  - WER corpus `0.0098 (0.98%)`, >50% WER samples `0`

## Sweep Notes
| config | result | pytest wall | latency mean | prompt mean | accuracy | WER | note |
|---|---:|---:|---:|---:|---:|---:|---|
| max-only `150528` | pass | `324.56s` | `115.054s` | `9580` | `60%` | `0.26%` | max-only leaves min-pixel clamp active |
| `3136/3136` | fail | `144.27s` | `33.961s` | `748` | `50%` | `2.72%` | WER above threshold |
| `4704/4704` | fail | `97.35s` | `38.197s` | `876` | `30%` | n/a | accuracy below threshold |
| `5488/5488` | pass | `141.57s` | `39.152s` | `1004` | `50%` | `1.04%` | faster, but drops accuracy/WER vs `6272` |
[...]

PR #929, opened 2026-07-02
fix(ci): recalibrate stage-10 talker latency gate for the #932 TensorRef fast path (#917)

MelodyyyYin:
> **Stacked on #932** (rebased onto it). After #932 (*explicit TensorRef pipeline policies*), this PR reduces to a single change: recalibrating the stage-10 talker latency gate. The env-removal + fast-path enablement that earlier revisions of this PR carried are now handled by #932 (it declares the `image_encoder -> mm_aggregate` TensorRef policy in the Qwen pipeline config), so those were dropped to avoid conflicting with its explicit-declaration design.

## Problem (#917)
Stage-10 (`stage-10-videoamme-talker-tp2`) failed its talker latency gate (155s; actual 160-186s). Two causes: (1) the #808 lazy-TensorRef fast path was not enabled in CI, so the pipeline ran the ~170s slow path; (2) the 155 gate was set *below its own P95 calibration base* (`latency_mean_s` 170.36 x1.125 = 191.7), so a run at the calibrated slow path deterministically exceeds it.

## Fix
- **#932** enables the fast path (declares the `image_encoder -> mm_aggregate` TensorRef edge policy in the Qwen pipeline config) -> stage-10 now runs ~22-38s.
- **This PR** recalibrates the gate `155 -> 80`: headroom over the fast path + CI contention, still far below ~170s so a TensorRef regression is caught. (The stale 170.36 P95 base is left as-is; only the override is retuned.)

## Validation
- E2E (rebuilt CI-image container, GPUs 6,7): clean fast-path runs 38.06s / 22.11s (matches #808 A/B ~37s + earlier reproduction ~33s), comfortably under the new 80s gate. #932 declares the identical policy (consumer=thinker, 2MB, same visual paths), so this carries over.

Refs #917. Depends on #932.

### DIFF
--- a/tests/test_model/test_qwen3_omni_videoamme_talker_tp2_ci.py
+++ b/tests/test_model/test_qwen3_omni_videoamme_talker_tp2_ci.py
@@ -64,10 +64,10 @@ _VIDEOAMME_TALKER_TP2_AUDIO_P95 = {
 }
 VIDEOAMME_TALKER_TP2_THRESHOLDS = apply_slack(_VIDEOAMME_TALKER_TP2_AUDIO_P95)
 
-# note (Yue Yin, Chenyang): 0.55-calibrated latency baseline (146.7 gate) is
-# unreachable on the #765 0.40 OOM-fix config (~148s); pin the gate to 155
-
-VIDEOAMME_TALKER_TP2_THRESHOLDS[16]["latency_mean_s_max"] = 155
+# note (Yue Yin): #932 enables the image_encoder->mm_aggregate TensorRef fast path
+# (~22-38s) for this pipeline, so the 155 gate (calibrated for the ~170s slow path,
+# and below its own 170.36 P95 base) no longer fits. Recalibrate to the fast path. #917
+VIDEOAMME_TALKER_TP2_THRESHOLDS[16]["latency_mean_s_max"] = 80
 
 
 @dataclass
--- a/tests/test_model/test_qwen3_omni_videoamme_talker_tp2_ci.py
+++ b/tests/test_model/test_qwen3_omni_videoamme_talker_tp2_ci.py
@@ -64,11 +64,6 @@ _VIDEOAMME_TALKER_TP2_AUDIO_P95 = {
 }
 VIDEOAMME_TALKER_TP2_THRESHOLDS = apply_slack(_VIDEOAMME_TALKER_TP2_AUDIO_P95)
 
-# note (Yue Yin): #932 enables the image_encoder->mm_aggregate TensorRef fast path
-# (~22-38s) for this pipeline, so the 155 gate (calibrated for the ~170s slow path,
-# and below its own 170.36 P95 base) no longer fits. Recalibrate to the fast path. #917
-VIDEOAMME_TALKER_TP2_THRESHOLDS[16]["latency_mean_s_max"] = 80
-
 
 @dataclass
 class _TalkerEvalArtifacts:
--- a/sglang_omni/config/__init__.py
+++ b/sglang_omni/config/__init__.py
@@ -19,6 +19,7 @@ from sglang_omni.config.schema import (
     StageConfig,
     StageResourceConfig,
     StageRuntimeConfig,
+    TensorRefEdgeConfig,
 )
 from sglang_omni.config.topology import (
     ProcessGroupPlacement,
@@ -51,4 +52,5 @@ __all__ = [
     "PlacementConfig",
     "RelayConfig",
     "EndpointsConfig",
+    "TensorRefEdgeConfig",
 ]
--- a/sglang_omni/config/schema.py
+++ b/sglang_omni/config/schema.py
@@ -20,6 +20,28 @@ class RelayConfig(BaseModel):
     device: str = "cpu"
 
 
+class TensorRefEdgeConfig(BaseModel):
+    """Explicit lazy tensor handoff policy for one outgoing stage edge."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    consumer_stage: str
+    threshold_mb: float = 2.0
+    paths: tuple[str, ...]
+
+    def model_post_init(self, __context: Any = None) -> None:
+        if self.threshold_mb < 0:
+            raise ValueError("tensor_ref_edges threshold_mb must be non-negative")
+        if not self.consumer_stage.strip():
+            raise ValueError("tensor_ref_edges consumer_stage must not be empty")
+        if not self.paths:
+            raise ValueError("tensor_ref_edges paths must not be empty")
+        cleaned = tuple(path.strip() for path in self.paths)
+        if any(not path for path in cleaned):
+            raise ValueError("tensor_ref_edges paths must not contain empty values")
+        self.paths = cleaned
+
+
 class EndpointsConfig(BaseModel):
     """Endpoint allocation settings."""
 
@@ -174,6 +196,7 @@ class StageConfig(BaseModel):
 
     # --- Relay (auto-inferred from gpu when None) ---
     relay: RelayConfig | None = None
+    tensor_ref_edges: dict[str, TensorRefEdgeConfig] = Field(default_factory=dict)
 
     def model_post_init(self, __context: Any = None) -> None:
         fields_set = self.__pydantic_fields_set__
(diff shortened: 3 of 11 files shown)


tests/test_model/test_qwen3_omni_videoamme_talker_tp2_ci.py
```diff
@@ -42,6 +43,27 @@
     wait_for_gpu_memory_release,
 )
 
+# note (Yue Yin): stage-10 runs the #808 lazy-TensorRef fast path (~33s vs the
+# ~170s slow path); the latency gate below is calibrated for it. Enabled here (not
+# just in CI env) so local runs match. Export the same two vars to reproduce. #917.
+_TENSOR_REF_ENV = {
+    "SGLANG_OMNI_ENABLE_TENSOR_REFS": "1",
+    "SGLANG_OMNI_TENSOR_REF_EDGES": "image_encoder:mm_aggregate:thinker",
+}
```
zhaochenyang20: I think we should refactor this. I highly dislike environment arguments. these parameter should be removed, and possibly adds as a server args, and default turn on.

tests/test_model/test_qwen3_omni_videoamme_talker_tp2_ci.py
```diff
@@ -42,6 +43,27 @@
...
+@pytest.fixture(scope="module", autouse=True)
+def _enable_tensor_refs():
+    _prev = {k: os.environ.get(k) for k in _TENSOR_REF_ENV}
+    os.environ.update(_TENSOR_REF_ENV)
+    yield
+    for k, v in _prev.items():
+        if v is None:
+            os.environ.pop(k, None)
+        else:
+            os.environ[k] = v
```
zhaochenyang20: do not use env arg

zhaochenyang20 (changes requested):

tests/test_model/test_qwen3_omni_videoamme_talker_tp2_ci.py
MelodyyyYin: Done in 6044b71 — dropped all env args and made #808 default-on:
- `tensor_refs_enabled()` now returns `True` by default (no `SGLANG_OMNI_ENABLE_TENSOR_REFS`).
- `TensorRefPolicy.from_env` -> `for_edge`: edges/threshold/paths come from module defaults (`DEFAULT_TENSOR_REF_EDGES`, alongside the existing `DEFAULT_TENSOR_REF_PATHS`), not the environment; `_parse_edges` removed.
- This `_enable_tensor_refs` fixture is gone; the fast path is the default, so local == CI with nothing exported.

On "possibly server args": I went with default-on + code config rather than a server flag — it's the simplest reusable form and, unlike env, needs no propagation into the spawned stage workers. Happy to add a runtime disable switch if you'd like one. Unit tests updated to the new API.

tests/test_model/test_qwen3_omni_videoamme_talker_tp2_ci.py
MelodyyyYin: Removed in 6044b71 — no env args anywhere now. Enablement is default-on via `tensor_refs_enabled()`; the edge comes from `DEFAULT_TENSOR_REF_EDGES`.

MelodyyyYin:
Addressed the review (@zhaochenyang20) in 6044b71 - removed all TensorRef env args and made #808 default-on:

- `tensor_refs_enabled()` -> default `True`
- `TensorRefPolicy.from_env` -> `for_edge`; edges/threshold/paths from module defaults (`DEFAULT_TENSOR_REF_EDGES` next to the existing `DEFAULT_TENSOR_REF_PATHS`); `_parse_edges` gone
- dropped the stage-10 `_enable_tensor_refs` env fixture - the fast path is the default, so local == CI with nothing exported
- unit tests moved off `setenv` to the new API / monkeypatching module defaults (43/43 pass)

On "possibly server args": I chose default-on + code config over a server flag - simplest reusable form, and (unlike env) it needs no propagation into the spawned stage workers. Easy to add a runtime disable switch if you want one. The gate change (155->80) is unchanged; behavior matches the env-on fast path I validated (~22-38s) since `for_edge` yields the same policy.

zhaochenyang20:
Follow our first principle: make a realistic impact.

This should not be a pure CI optimization. CI is a projection of the real user path. Stage-10 runs the same `write_payload`/`read_payload` code every user hits on FP8 TP2 with video+audio input. So far, this may look weird for Qwen3 Omni, but for a larger model, the request is realistic.

If the gate drops from ~170s to ~22-38s, that is transport work eliminated from the serving path itself — users with large video embeds have been paying it in production this whole time. This is why default-off-with-a-CI-fixture was the wrong shape, and default-on with `local == CI == user` is the right one.

zhaochenyang20:
The scope of default-on must be derived, not guessed. The invariants for a safe edge are, concretely:

1. Every hop between producer and declared consumer treats the paths as opaque — including `merge_fn` / `project_payload` / `route_fn`, not just executors.
2. Exactly one resolver, exactly once (blobs are currently single-resolve; SHM unlinks on read).
3. The consumer's relay can reach the producer's blob directly (implicit today on single-host; a real constraint for disaggregated placement).
4. Abort/failure paths release blobs (already generic).

**Third, and most important: every case that fails these invariants today is an incomplete implementation, not a fundamental limit.** We'd better have follow-up issues filed for each, because the end state should be "on for every stage, by construction":

- Ref-counted / non-destructive blob reads → unlocks `requires_tp_work_fanout=True` consumers and multi-branch routing.
- "Writing intermediate stage" (the Ming-Omni case) is really an undeclared consumer → give it resolve-modify-republish (or COW) semantics instead of excluding the model.
- `project_mm_aggregate_to_talker_ar` currently copies all of `thinker_inputs`, so the talker branch hauls embeds it never uses (a full third hop on the slow path) and carries an unresolved ref it must never touch on the fast path. Strip those fields in the projection — this is a win independent of TensorRef.
- Move edge declarations from module-global defaults into per-pipeline config (aligns with #932), so enabling a new model is a declaration + audit of the three questions above (who touches it, who resolves it, can the consumer reach the producer), not a naming coincidence.

tests/test_model/test_qwen3_omni_videoamme_talker_tp2_ci.py
```diff
@@ -64,10 +64,10 @@
 }
 VIDEOAMME_TALKER_TP2_THRESHOLDS = apply_slack(_VIDEOAMME_TALKER_TP2_AUDIO_P95)
 
-# note (Yue Yin, Chenyang): 0.55-calibrated latency baseline (146.7 gate) is
-# unreachable on the #765 0.40 OOM-fix config (~148s); pin the gate to 155
-
-VIDEOAMME_TALKER_TP2_THRESHOLDS[16]["latency_mean_s_max"] = 155
+# note (Yue Yin): #808 is default-on now, so stage-10 runs its ~22-38s fast path, not the
+# ~170s slow path the old 155 gate targeted. 80 = fast-path headroom for CI contention,
+# still far below ~170s so a TensorRef regression is caught. #917
```
zhaochenyang20: given the new situation. we can remove this. Please screenshot this PR's stage 10 to provide the new speed.

And we will do a recalibration for this CI later, and I strongly think that not only this CI would be affected, indeed a lot of the stages for Qwen3 Omni should be better.

luojiaxuan:
Opened the Qwen3 Omni optimization roadmap here: #934. It tracks this Stage-10 TensorRef/default-on work together with #804, #803, and #932.

MelodyyyYin:
Rebased onto #932 (which declares the `image_encoder->mm_aggregate` TensorRef policy in pipeline config). This PR is now just the stage-10 gate override removal on top; will rebase to `main` once #932 merges.

tests/test_model/test_qwen3_omni_videoamme_talker_tp2_ci.py
MelodyyyYin: Removed in 7415ea8 -> gate falls back to `apply_slack`; recalibration deferred to #934. Stage-10 now ~21s (fast path).

MelodyyyYin:
Stage-10 @ 7415ea8 — latency **20.948s** (p95 22.5s), fast path active (`tensor_ref active: video_embeds 55MiB, image_encoder->mm_aggregate`), all gates green. ~8x vs the ~170s slow path. Gate override removed per review; recalibration tracked in #934. (Speed screenshot below.)

Screenshot 2026-07-03 at 10 53 44 PM

Issue #921, opened 2026-07-02
[Roadmap] Router & Same-GPU DP


JiaxinD:
## Goals & context

Serving on H100 is host / CPU-dispatch-bound, not GPU-bound (#907): the GPU sits ~72% idle at single-replica saturation (SM ~28%). Same-GPU DP + CUDA MPS recovers most of that idle (DP3 + MPS = +114% [measured], +134% in db-ol's independent reproduction on #912), but a single-process Python router then becomes the next single-core bottleneck (via-router DP3 lands below via-router DP2, router pinned at 88 to 96% of one core).

The goal is not MPS itself but maximizing single-GPU serving performance, across three layers:

1. **Single-worker efficiency** — cut kernel-launch, sync, Python/GIL, and scheduling overhead (Workstreams 1 & 5 below).
2. **Same-GPU DP × MPS** — when one replica still leaves measurable GPU headroom (recipes section, Workstreams 2 & 3).
3. **Router scaling** — so higher backend capacity doesn't hit a new single-core ceiling (Phases 0–3, Workstream 4).

The serving layer is model-agnostic (Higgs / MOSS / Qwen3-Omni go through the same router path), so the router and DP × MPS tracks cover the MOSS series too; MOSS's intra-worker vocoder bottleneck is orthogonal and stays a separate track.

## Router Phase 0: Streaming relay

Stream proxied responses instead of buffering, removing the byte-growth half of the router cost.
- https://github.com/sgl-project/sglang-omni/pull/920 — merged
- A fix for a pre-existing bug: https://github.com/sgl-project/sglang-omni/pull/931 — merged
- https://github.com/sgl-project/sglang-omni/pull/1014 — overload protection (admission control + unified `--max-connections` semantics), merged
- https://github.com/sgl-project/sglang-omni/pull/1071
- https://github.com/sgl-project/sglang-omni/pull/1015 — bench-side hardening (keep clients alive on per-request failures), PR open

## Router Phase 1: Python CP/DP router

- https://github.com/sgl-project/sglang-omni/pull/1049 — PR open in review
- https://github.com/sgl-project/sglang-omni/pull/1077 — process lifecycle: auto-kill orphaned stage workers on parent crash, PR open, approved

[...]

PR #931, opened 2026-07-02
[router] Fix in-flight counter leak on mid-stream upstream failure

JiaxinD:
## Summary
Pre-existing in-flight counter leak in the streaming relay, found while testing #920.

In `_forward_streaming`'s cleanup, `await upstream.aclose()` runs before `worker.decrement_active()`. If `aclose()` raises on a broken mid-stream connection (a real case when the upstream fails mid-body), `decrement_active()` and the completion recording are skipped, so the in-flight count is never decremented and `least_request` load balancing drifts permanently toward the affected worker.

## Fix
Wrap `aclose()` in a `try/finally` so `decrement_active()` and the completion recording always run, whether or not `aclose()` raises. `proxy.py` only.

## Test [measured]
`test_streaming_inflight_count_decrements_even_if_aclose_raises`: a mid-stream failure whose `aclose()` also raises. Fails on current main (the count leaks to 1), passes with the fix (0). Full router suite green.

## Note
Pre-existing on main, independent of #920. #920 unifies the two relay methods into `_forward_relay`, which inherits the same cleanup, so this fix and #920 overlap; whichever merges second rebases to carry the fix into `_forward_relay`.

### DIFF
--- a/tests/unit_test/router/test_app.py
+++ b/tests/unit_test/router/test_app.py
@@ -1125,6 +1125,43 @@ def test_streaming_failure_records_single_worker_failure() -> None:
     assert worker.state == "healthy"
 
 
+def test_streaming_inflight_count_decrements_even_if_aclose_raises() -> None:
+    class AcloseRaisingStream(httpx.AsyncByteStream):
+        async def __aiter__(self):
+            yield b"data: chunk\n\n"
+            raise httpx.ReadError("stream boom")
+
+        async def aclose(self) -> None:
+            raise RuntimeError("aclose boom")
+
+    def handler(request: httpx.Request) -> httpx.Response:
+        if request.url.path == "/health":
+            return httpx.Response(200, json={"status": "healthy"}, request=request)
+        if request.url.path == "/v1/chat/completions":
+            return httpx.Response(
+                200,
+                stream=AcloseRaisingStream(),
+                headers={"content-type": "text/event-stream"},
+                request=request,
+            )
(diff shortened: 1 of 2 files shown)


zhaochenyang20:
Please track this in https://github.com/sgl-project/sglang-omni/issues/921

Ratish1 (approved):
LGTM with one small comment

tests/unit_test/router/test_app.py
```diff
@@ -1125,6 +1125,43 @@ def handler(request: httpx.Request) -> httpx.Response:
...
+        with client.stream(
+            "POST",
+            "/v1/chat/completions",
+            json={"model": "qwen3-omni", "stream": True},
+        ) as response:
+            b"".join(response.iter_bytes())
+
+    # Note (Jiaxin Deng): the in-flight count must decrement even though aclose()
+    # raised, otherwise it leaks and least_request drifts permanently.
+    assert all(worker.active_requests == 0 for worker in app.state.workers)
```
Ratish1, line 1162: This test should also assert the paired request accounting, not only `active_requests == 0`.

The cleanup path now guarantees three lifecycle actions after a broken stream/`aclose()` failure: close attempt, `decrement_active()`, and `record_routed_request(...)`. The test currently only protects the in-flight counter, so a future change could still skip `record_routed_request()` and this test would pass.

JiaxinD: Added: now also asserts routed_requests==1 and failed_requests==1, so the completion accounting is protected alongside the in-flight decrement.

PR #932, opened 2026-07-02
[CI Perf] Use explicit TensorRef pipeline policies

luojiaxuan:
## Summary
- remove TensorRef env gates and the from_env parser
- add explicit TensorRef edge policy config on StageConfig
- wire the Qwen image_encoder -> mm_aggregate TensorRef policy directly in pipeline config

## Tests
- python3 -m py_compile on changed Python files
- bundled Python Qwen config smoke for image_encoder TensorRef policy
- pytest not run locally: both available Python runtimes are missing pytest

### DIFF
--- a/sglang_omni/config/__init__.py
+++ b/sglang_omni/config/__init__.py
@@ -19,6 +19,7 @@ from sglang_omni.config.schema import (
     StageConfig,
     StageResourceConfig,
     StageRuntimeConfig,
+    TensorRefEdgeConfig,
 )
 from sglang_omni.config.topology import (
     ProcessGroupPlacement,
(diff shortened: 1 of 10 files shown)


luojiaxuan:
Roadmap tracker: #934. This PR is the per-pipeline TensorRef policy declaration piece, so model enablement becomes an explicit audit instead of env/default naming logic.

MelodyyyYin:
Heads-up: `lint` is red here (black nits in `mp_runner.py` + `test_stage.py`); `pre-commit run --all-files` fixes it. #929 stacks on this and already carries the fix.

PR #933, opened 2026-07-02
[TTS Refactor][M8] Model capabilities: Add static TTS model capability metadata

SandyLuXY:
## Motivation
This PR implements the [M8 StageCapabilities](https://ljp2lkmtlez9.jp.larksuite.com/wiki/JkybwluLCiSRQXkb7Cije9tipQM?from=from_copylink) refactor as described in [SGLang-Omni-TTS-Refactor-RFC-v2](https://osgbw74w8zwb.sg.larksuite.com/docx/L3XId4GHYoJqSKxwXYqlXaBrgNd?from=from_copylink).

Introduce static, architecture-level TTS capability metadata so CI checks, developer documentation, and startup logs can query what each TTS model family supports without inspecting stage implementations or runtime-only pipeline state.

This is additive metadata. It does not route requests, change scheduler behavior, or replace instance-level `PipelineConfig` capability methods that can still depend on a specific checkpoint or deployment.

## Modifications

1. `sglang_omni/models/model_capabilities.py`
   - Adds the frozen `ModelCapabilities` dataclass with explicit boolean fields for reference audio, batch vocoder, streaming vocoder, CUDA graph support, and torch compile support.
   - Adds `get_model_capabilities(architecture)` to resolve a registered architecture through `PIPELINE_CONFIG_REGISTRY`, import its model package, and return its exported `CAPABILITIES`.
   - Returns `None` for unknown architectures or registered models without capability metadata, and raises a `TypeError` when a model exports a malformed `CAPABILITIES` value.

2. TTS model package `CAPABILITIES` exports
   - Adds `CAPABILITIES` constants to `qwen3_tts`, `higgs_tts`, `moss_tts`, `moss_tts_local`, `fishaudio_s2_pro`, and `voxtral_tts`.
   - Updates package `__all__` exports where needed so capability metadata is part of each model package's explicit public surface.
   - Covers the current TTS capability matrix for reference audio, batch vocoder, streaming vocoder, CUDA graph, and torch compile support.

3. `sglang_omni/serve/launcher.py`
   - Adds `_model_capabilities_log_summary(...)` to convert a loaded pipeline config into a compact capability summary.
   - Adds `_log_model_capabilities(...)` and emits a sorted JSON `Model capabilities:` startup log after placement/topology logging.
   - Wraps capability lookup failures in a warning so startup logging cannot break server launch.

4. Tests and documentation
   - Adds `tests/unit_test/models/test_model_capabilities.py` covering frozen/explicit dataclass fields, per-model exports, registry lookup, alias resolution, static architecture metadata semantics, non-TTS/unknown behavior, and launcher logging.
   - Adds `tests/unit_test/models/__init__.py` for the new unit test package.
   - Updates `tests/README.md` to document the new model registry/capability test area.

## Related Issues
N/A

## Accuracy Test

N/A

## Benchmark & Profiling

N/A

## Checklist

- [x] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/models/model_capabilities.py
+++ b/sglang_omni/models/model_capabilities.py
@@ -10,10 +10,24 @@ from types import ModuleType
 
 @dataclass(frozen=True)
 class ModelCapabilities:
-    """Static capability declaration for a TTS model architecture.
+    """Static capability declaration for a model architecture.
 
-    These flags describe what a model architecture can support, not what a
-    specific pipeline instance enables for a checkpoint or deployment.
+    These flags describe what a model architecture can support. Concrete
+    checkpoint and deployment policy stays with ``PipelineConfig`` methods. For
+    example, a Qwen3-TTS CustomVoice checkpoint can reject uploaded reference
+    audio even though the architecture declares reference-audio support.
+
+    Fields:
+    - supports_reference_audio: the architecture can condition on reference
+      audio when the selected checkpoint/deployment allows it.
+    - supports_batch_vocoder: the architecture can produce batched waveform
+      output.
+    - supports_streaming_vocoder: the architecture can stream vocoder output
+      before the full generation payload is complete.
+    - supports_cuda_graph: the architecture has a CUDA graph path.
+    - supports_torch_compile: the architecture has an owned ``torch.compile``
+      path, including codec, codebook, or frame-sampler compiles. This is not
+      limited to the generic SGLang ``enable_torch_compile`` server arg.
     """
 
     supports_reference_audio: bool
--- a/sglang_omni/models/model_capabilities.py
+++ b/sglang_omni/models/model_capabilities.py
@@ -1,0 +1,64 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Capability declarations for model architectures."""
+
+from __future__ import annotations
+
+import importlib
+from dataclasses import dataclass
+from types import ModuleType
+
+
+@dataclass(frozen=True)
+class ModelCapabilities:
+    """Static capability declaration for a TTS model architecture.
+
+    These flags describe what a model architecture can support, not what a
+    specific pipeline instance enables for a checkpoint or deployment.
+    """
+
+    supports_reference_audio: bool
+    supports_batch_vocoder: bool
+    supports_streaming_vocoder: bool
+    supports_cuda_graph: bool
+    supports_torch_compile: bool
+
+
+def get_model_capabilities(architecture: str) -> ModelCapabilities | None:
+    """Look up capabilities for a registered model architecture."""
+    module = _model_package_for_architecture(architecture)
+    if module is None:
+        return None
+    return _module_model_capabilities(module)
+
+
+def _model_package_for_architecture(architecture: str) -> ModuleType | None:
+    from sglang_omni.models.registry import PIPELINE_CONFIG_REGISTRY
+
+    config_cls = PIPELINE_CONFIG_REGISTRY.configs.get(architecture)
+    if config_cls is None:
+        return None
+    package = config_cls.__module__.rsplit(".", 1)[0]
+    return importlib.import_module(package)
+
+
+def _module_model_capabilities(module: ModuleType) -> ModelCapabilities | None:
+    capabilities = getattr(module, "CAPABILITIES", None)
+    if capabilities is None:
+        return None
+    return _ensure_model_capabilities(
+        capabilities, f"{module.__name__}.CAPABILITIES"
+    )
+
+
+def _ensure_model_capabilities(
+    capabilities: object, source: str
+) -> ModelCapabilities:
+    if not isinstance(capabilities, ModelCapabilities):
+        raise TypeError(f"{source} must be a ModelCapabilities instance")
+    return capabilities
+
+
+__all__ = [
+    "ModelCapabilities",
+    "get_model_capabilities",
+]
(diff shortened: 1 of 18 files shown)


sglang_omni/models/moss_tts_local/__init__.py
```diff
@@ -1,2 +1,14 @@
...
 """MOSS-TTS Local (v1.5) pipeline package."""
+
+from sglang_omni.models.model_capabilities import ModelCapabilities
+
+CAPABILITIES = ModelCapabilities(
+    supports_reference_audio=True,
+    supports_batch_vocoder=True,
+    supports_streaming_vocoder=True,
+    supports_cuda_graph=True,
+    supports_torch_compile=False,
```
JiaxinD: This looks off: MOSS-Local torch.compiles its frame sampler in the default path (`sglang_model.py` via `init_frame_decode_graphs`), yet Higgs, Fish, and Qwen3 all set `supports_torch_compile=True` for their codec/codebook compiles. Whichever semantic is intended, please define it in the field docstring and make this value consistent (flip to True under the any-torch.compile reading).

sglang_omni/models/model_capabilities.py
```diff
@@ -0,0 +1,60 @@
...
+
+@dataclass(frozen=True)
+class ModelCapabilities:
+    """Static capability declaration for a TTS model architecture.
+
+    These flags describe what a model architecture can support, not what a
+    specific pipeline instance enables for a checkpoint or deployment.
+    """
+
+    supports_reference_audio: bool
```
JiaxinD, line 33: Suggest the field docstrings note that per-deployment truth stays with `PipelineConfig`: e.g. a CustomVoice checkpoint rejects uploaded references even though the architecture sets `supports_reference_audio=True`.

sglang_omni/models/model_capabilities.py
```diff
@@ -0,0 +1,60 @@
...
+def _module_model_capabilities(module: ModuleType) -> ModelCapabilities | None:
+    capabilities = getattr(module, "CAPABILITIES", None)
+    if capabilities is None:
+        return None
+    return _ensure_model_capabilities(capabilities, f"{module.__name__}.CAPABILITIES")
+
+
+def _ensure_model_capabilities(capabilities: object, source: str) -> ModelCapabilities:
+    if not isinstance(capabilities, ModelCapabilities):
+        raise TypeError(f"{source} must be a ModelCapabilities instance")
```
JiaxinD, line 67: The `TypeError`-on-malformed-CAPABILITIES contract and the launcher warning branch are both untested. One negative test each would lock them in.

tests/unit_test/models/test_model_capabilities.py
```diff
@@ -0,0 +1,184 @@
...
+    "VoxtralTTSForConditionalGeneration": ModelCapabilities(
+        supports_reference_audio=False,
+        supports_batch_vocoder=False,
+        supports_streaming_vocoder=False,
+        supports_cuda_graph=True,
+        supports_torch_compile=True,
+    ),
+}
+
+TTS_PACKAGE_NAMES = {
```
JiaxinD: Is CAPABILITIES optional or required for TTS packages? This frozen name list means a 7th TTS model stays green whether or not it declares CAPABILITIES; if it should be required, discover TTS packages by a property instead.

tests/unit_test/models/test_model_capabilities.py
```diff
@@ -0,0 +1,184 @@
...
+from dataclasses import MISSING, FrozenInstanceError, fields
+
+import pytest
+
+from sglang_omni.models.model_capabilities import (
+    ModelCapabilities,
+    get_model_capabilities,
+)
+from sglang_omni.models.registry import PIPELINE_CONFIG_REGISTRY
+from sglang_omni.serve.launcher import (
```
JiaxinD: Module-level import of `serve.launcher` pulls fastapi/uvicorn at collection time, so the whole file (including the pure-dataclass tests) fails to collect without the serving stack. Could move this import into the three launcher tests.

JiaxinD (commented):
M8 mechanism and launcher wiring look solid: failure-isolated startup logging, no new import cycle, and the value table is pinned by tests. Cross-checked all 30 capability values against main and 29 hold; one blocker on the remaining value plus a few smaller items inline.

sglang_omni/models/moss_tts_local/__init__.py
SandyLuXY: Thanks for catching this! Fixed in 8fd743c4. MOSS-TTS Local now reports `supports_torch_compile=True`, and the docstring says this covers architecture-owned compile paths like the frame sampler, not just `enable_torch_compile`.

sglang_omni/models/model_capabilities.py
SandyLuXY: Agreed. `ModelCapabilities` now states that these are architecture-level flags and that checkpoint/deployment behavior stays with `PipelineConfig`; it also calls out the Qwen3-TTS CustomVoice reference-audio case.

sglang_omni/models/model_capabilities.py
SandyLuXY: Added coverage for malformed `CAPABILITIES` through lookup and for launcher warning isolation when capability summary resolution fails

tests/unit_test/models/test_model_capabilities.py
SandyLuXY: Good q, registered TTS configs now opt in with `requires_model_capabilities=True`, and the coverage test discovers those configs from the registry instead of a frozen package-name list.

JiaxinD (approved):

zhaochenyang20:
move to #957  and merge!!

Issue #934, opened 2026-07-03
[Roadmap] Qwen3 Omni serving/CI performance optimization


luojiaxuan:
## Why

Follow the first principle: optimize real serving needs, not a CI artifact. CI is a projection of the user path. The Stage-10 FP8 TP2 video+audio workload hits the same `write_payload` / `read_payload` transport code used by serving, so reducing CI latency here should translate to real Qwen3 Omni serving wins.

This roadmap tracks the connected work to make TensorRef and related Qwen3 Omni pipeline optimizations default-on by construction, with measured impact, bounded rollout risk, and clear safety boundaries instead of ad hoc CI-only switches.

Context: @MelodyyyYin @zhaochenyang20 @edwingao28 

## Active PRs

- [ ] #929 - Stage 10 TensorRef default-on + latency recalibration
- [ ] #804 - Stage 11 payload reduction + talker prefill projection slimming
- [ ] #803 - Stage relay payload IO tracing
- [ ] #932 - explicit TensorRef pipeline policy config

## Main optimization tracks

- [ ] **Transport default-on:** TensorRef per-pipeline policy plus safe-edge invariants, so large inter-stage tensors are refs by default where the pipeline proves the edge is safe.
- [ ] **TP fanout unlock:** ref-counted or non-destructive blob reads, so TP follower fanout and multi-branch routing stop being hard boundaries for TensorRef.
- [ ] **Pipeline payload discipline:** every branch projection carries only the fields that branch actually consumes; #804 is the first concrete cut in this direction.

## Quantitative targets

These are initial targets; update the table with measured baselines from CI artifacts / `stage_io_trace` once #803 and #929 provide fresh data.

| Area | Current baseline | Target | Source / owner |
| --- | ---: | ---: | --- |
| Stage 10 FP8 TP2 Video-AMME latency | recent failures around `160-185s`; TensorRef runs reported around `22-38s` | p95 <= `80s` over CI-like repeated runs; no retry above `100s` | #917, #929 |
| Stage 11 reduced video payload | PR #804 selected config: test wall `155.34s`, latency mean `46.912s`, prompt mean `1132`, WER `0.98%` | keep test wall <= `180s`, latency mean <= `55s`, prompt mean <= `1200`, corpus WER <= `2.5%`, zero >50% WER samples | #804 |
[...]

PR #935, opened 2026-07-03
[Ming-Omni] Make mm_aggregate a pure relay and enable the video_embeds TensorRef edge

edwingao28:
> [!NOTE]
> Stacked on #932 (branch rebased onto `luojiaxuan/tensor-ref-explicit-policy`), merge order 932 -> 935. Until #932 lands, the diff below includes its commits — review the top 3 commits (`[Ming-Omni] ...`).

## Motivation

`merge_for_thinker` in `sglang_omni/models/ming_omni/pipeline/merge.py` transforms embeds at `mm_aggregate` (`_as_tensor` + `dim()`/`squeeze(0)`), so Ming cannot join the TensorRef path (#808). Worse, an unresolved `TensorRef` reaching the merge is silently dropped — `_as_tensor` swallows the `torch.as_tensor(ref_dict)` exception and returns `None`, deleting `video_embeds` from thinker inputs with no error; masked today only because TensorRef is off by default.

## Modifications

- `pipeline/merge.py`: `_non_empty` recognizes TensorRef dicts via shape metadata (same pattern #808 added to the qwen3 merge); embeds forwarded untouched; `_as_tensor` removed.
- `stages.py`: audio/image encoder executors flatten `[B, T', H] -> [T', H]` at the producer (`flatten_embed_batch_dims`), before `write_payload` externalizes the tensor — `mm_aggregate` becomes a pure relay; no COW semantics needed for Ming.
- `config.py`: declare the Ming TensorRef edge via #932's `TensorRefEdgeConfig` — `image_encoder -> mm_aggregate`, consumer `thinker`, `paths=("video_embeds",)` (video only for now, matching the qwen3 scope; image/audio can follow once #803 trace data justifies them).
- `tests/unit_test/ming_omni/test_pipeline.py`: regression test that an unresolved `video_embeds` ref survives merge (fails on the pre-fix code with `KeyError: 'model_inputs'`), plus producer-flatten tests for the audio and image executors.

## Related Issues

Ming rollout step of #934. Builds on #808. Stacked on #932.

## Accuracy Test

Behavior-neutral with refs off: the squeeze moves from `mm_aggregate` to the producer, thinker inputs are identical. Unit tests (run on #932's branch + these commits):

```
$ pytest tests/unit_test/ming_omni/test_pipeline.py tests/unit_test/pipeline/test_tensor_ref.py -q
38 passed
```

Ming has no GPU CI stage, so before merge I will run a manual refs-on E2E on a GPU box and post output-parity + latency results here.

## Benchmark & Profiling

Numbers will come from the manual E2E above; #808 measured 160-179s -> 36-37s (~4.5x) on the equivalent qwen3 stage 11 video path.

## Checklist

- [x] Add unit tests.

## CI

if labeled.

### DIFF
--- a/benchmarks/eval/benchmark_omni_mmmu.py
+++ b/benchmarks/eval/benchmark_omni_mmmu.py
@@ -84,7 +84,7 @@ from benchmarks.dataset.mmmu import load_mmmu_samples
 from benchmarks.metrics.mmmu import compute_mmmu_metrics, print_mmmu_accuracy_summary
 from benchmarks.metrics.performance import compute_speed_metrics, print_speed_summary
 from benchmarks.metrics.wer import print_wer_summary
-from benchmarks.tasks.tts import (
+from benchmarks.tasks.asr import (
     DEFAULT_ASR_TRANSCRIBE_CONCURRENCY,
     compute_text_audio_consistency,
 )
--- a/benchmarks/eval/benchmark_omni_mmsu.py
+++ b/benchmarks/eval/benchmark_omni_mmsu.py
@@ -107,14 +107,14 @@ from benchmarks.dataset.mmsu import MmsuSample, load_mmsu_samples
 from benchmarks.metrics.mmsu import compute_mmsu_metrics, print_mmsu_summary
 from benchmarks.metrics.performance import compute_speed_metrics
 from benchmarks.metrics.wer import print_wer_summary
+from benchmarks.tasks.asr import (
+    DEFAULT_ASR_TRANSCRIBE_CONCURRENCY,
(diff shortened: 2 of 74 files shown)


luojiaxuan:
Currently, only audio_executor_flattens_embed_batch_dim is tested. It would be better to also add an image/video flatten test, since the main benefit of TensorRef will most likely come from video embeds.

Also, build_thinker_inputs now assumes that audio_out/image_out are dicts; it used to be more lenient with non-dict types. It's not a big deal, but we could keep the guard in place to prevent bad payloads from directly throwing an AttributeError

PR #936, opened 2026-07-03
[TTS Refactor][M6] StreamingVocoderBase: base + MOSS-TTS-Local migration

GaokaiZhang:
Extracts the request-local streaming-vocoder lifecycle into `sglang_omni/scheduling/streaming_vocoder.py` (`StreamingVocoderBase(StreamingSimpleScheduler)`) and migrates MOSS-TTS-Local onto it. First T5b/M6 migration from #661 (module doc: [06_streaming_vocoder](https://osgbw74w8zwb.sg.larksuite.com/docx/IvBVdwzk6o6OWhxI7FLlsHmugnc)); higgs / fish / qwen3-omni code2wav follow as separate PRs per the tracker's per-model split. Also settles the layout question raised on #875: separate `streaming_vocoder.py` next to `batch_vocoder.py`.

The base owns what the four streaming vocoders each re-implement today: per-request state registry, the `params["stream"]` gate, the chunk metadata/tensor scaffold with contract latching, threshold → decode-delta → emit ordering, `initial_codec_chunk_frames` resolution (absorbing `models/tts_streaming.py`), stream-done flush → final chunk → terminal result sequencing, nothing-emitted fallback, and abort semantics — late chunks for aborted or completed ids are dropped and never recreate state. Cross-request coalescing is expressed through `select_step_participants` / `build_step_plan` / `run_step` / `on_step_failure`; MOSS-local, the hardest case (persistent codec session, slot lifecycle, CUDA-graph capture/replay), validates that surface per the RFC's hardest-case-first rule. Cursor math, buffers, codec sessions, graph handling and result shapes stay model-owned.

Behavior class: Green (output-equivalent). The old MOSS-local lifecycle is deleted in this PR (no dual path); `tts_streaming.py` is deleted and its higgs importers get path-only changes — higgs itself is not migrated here. Client-visible error text is preserved via the base's `stream_source_hint`.

Tests: 18 CPU contract tests in `tests/unit_test/scheduling/test_streaming_vocoder.py` (state lifecycle, late-chunk drop, latch immutability, done-before-payload, nothing-emitted fallback, coalesced-step failure isolation, one-backbone assertions). All pre-existing MOSS-local behavioral tests pass unmodified; two that only re-covered base scaffold moved into the contract suite.

Full SeedTTS-EN gate (complete en split, c=16, DP1, no router, 1x H100, `--max-running-requests 64 --cuda-graph-max-bs 64`), MOSS-TTS-Local-Transformer-v1.5, main @ 1cb04ad vs this branch:

| main, non-stream | en | 2.76% (1.91% excl. 10 outliers >50%) | 65.35 | 1088/1088 | 0 / 0 | 8.67 req/s | 0.438 |
| **this PR, non-stream** | en | 2.22% (1.74% excl. 6) | 65.49 | 1088/1088 | 0 / 0 | 8.65 req/s | 0.439 |
| main, stream | en | 2.38% (1.76% excl. 8) | 65.62 | 1088/1088 | 0 / 0 | 5.91 req/s | 0.636 |
| **this PR, stream** | en | 2.27% (1.75% excl. 6) | 65.00 | 1088/1088 | 0 / 0 | 5.91 req/s | 0.633 |

Streaming TTFP mean/p95: 1.501/2.040 s (main) vs 1.490/2.072 s (this PR). WER deltas are sampling noise (server-default temperature); throughput/RTF/TTFP are flat.


sglang_omni/scheduling/streaming_vocoder.py
```diff
@@ -0,0 +1,496 @@
...
+    def _stream_state_items(self) -> list[tuple[str, StreamStateT]]:
+        return list(self._stream_states.items())
+
+    def _stream_has_emitted(self, request_id: str) -> bool:
+        return request_id in self._emitted_stream_ids
+
+    def _mark_stream_emitted(self, request_id: str) -> None:
+        self._emitted_stream_ids.add(request_id)
+
+    def _record_completed_stream_request_id(self, request_id: str) -> None:
```
SandyLuXY, line 284: `_record_completed_stream_request_id` records completed stream IDs in a `set`, then trims `list(self._completed_stream_request_ids)[:excess]` once the cap is exceeded. A `set` does not preserve completion order, so this can evict a recently completed request while keeping older entries.

luojiaxuan (commented):
Code-quality/accounting note: this base extraction is a net increase by itself (GitHub shows `+1260/-294`, net `+966`; production code is about net `+399`). That is acceptable for the shared base if the follow-up migrations are counted as the same refactor paydown, but please keep the aggregate LOC visible in the stack. With #939’s true incremental diff and #937, production code is net delete; without those follow-ups, this PR alone does not meet the final net-delete goal.

sglang_omni/scheduling/streaming_vocoder.py
```diff
@@ -0,0 +1,496 @@
...
+        finally:
+            self._shutdown_stream_states()
+
+    def stop(self) -> None:
+        was_running = self._running
+        super().stop()
+        if not was_running:
+            self._shutdown_stream_states()
+
+    def _shutdown_stream_states(self) -> None:
```
SandyLuXY, line 142: Shutdown currently clears `_stream_states` without going through `release_stream_resources()`. MOSS closes the whole session in `on_serving_stop()`, but this base contract will leak any subclass-owned per-request resource on stop/teardown. Please release each live state through the same hook used by completion/abort, and add a stop-path contract test.

tests/unit_test/scheduling/test_streaming_vocoder.py
SandyLuXY, line 1: Please add the new `tests/unit_test/scheduling/test_streaming_vocoder.py` entry to the scheduling section in `tests/README.md`; the test tree is stale now.

sglang_omni/scheduling/streaming_vocoder.py
```diff
@@ -0,0 +1,496 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Streaming vocoder scheduler base (TTS refactor M6, issue #661).
```
SandyLuXY: The module docstring should describe the durable scheduler contract, not migration history. Please move that provenance out of the production docstring or keep it as a short local note near the moved helper.

tests/unit_test/scheduling/test_streaming_vocoder.py
```diff
@@ -0,0 +1,589 @@
...
+    scheduler.on_stream_chunk_batch([("a", _item([1])), ("b", _item([2]))])
+    messages = _drain(scheduler)
+    errors = [m for m in messages if m.type == "error"]
+    assert {m.request_id for m in errors} == {"a", "b"}
+    assert all(m.type == "error" for m in messages)
+    assert scheduler._stream_states == {}
+    assert scheduler._is_aborted("a") and scheduler._is_aborted("b")
+    assert "release:a" in scheduler.calls and "release:b" in scheduler.calls
+    # note (Gaokai Zhang): late chunks after the failed step stay dropped and must
+    # never recreate state (T9's lazy abort relies on this).
```
SandyLuXY: nit: developers are very likely not having context about what `T9` means. Please phrase the comment to make it more clear

sglang_omni/scheduling/streaming_vocoder.py
GaokaiZhang: Fixed in 78c570c — insertion-ordered dict now, trimmed oldest-first, with a contract test.

sglang_omni/scheduling/streaming_vocoder.py
GaokaiZhang: Fixed in 6393a65: shutdown routes every live state through `clear_stream_state` (the completion/abort hook) before `on_serving_stop()`, and a failing release is logged so the remaining states and teardown still run. Stop-path contract tests added for both.

tests/unit_test/scheduling/test_streaming_vocoder.py
GaokaiZhang: Added in 6393a65.

sglang_omni/scheduling/streaming_vocoder.py
GaokaiZhang: Dropped in 6393a65 — the docstring now only describes the contract, including the stop-path release.

tests/unit_test/scheduling/test_streaming_vocoder.py
GaokaiZhang: Reworded in 6393a65 without the codename.

SandyLuXY:
@GaokaiZhang Please fix the merge conflicts of the PR.

sglang_omni/scheduling/streaming_vocoder.py
```diff
@@ -0,0 +1,505 @@
...
+            return
+        self.latch_stream_contract(request_id, state, payload, origin="payload")
+
+    def on_stream_chunk(
+        self, request_id: str, item: StreamItem
+    ) -> list[OutgoingMessage]:
+        state = self._ingest_stream_item(request_id, item)
+        if state is None:
+            return []
+        if self._can_batch_stream_chunks:
```
edwingao28: One small hardening thought while this path is being moved into the shared base. I think this shape came from mine earlier moss coalescing work, so this is not a new regression, but now that `StreamingVocoderBase` is becoming the public hook surface it may be worth closing the contract here.

The normal serving path routes through `_handle_stream_chunk()` -> `on_stream_chunk_batch()`, so ingest + pump run under `_state_lock` and abort cleanup happens after the lock is released. A direct call to `on_stream_chunk()` when `_can_batch_stream_chunks` is true still runs the same coalesced pump without that wrapper.

Could we either route this branch through the batch helper too, or explicitly make direct `on_stream_chunk()` unsupported for coalescing schedulers? That would keep future tests/subclasses from accidentally using the wrong entry point.

GaokaiZhang: Went with the explicit option in 0df3f94: on_stream_chunk() now raises for coalescing schedulers and points at on_stream_chunk_batch, so the unlocked pump path is gone. Tests moved to the batch entry, plus a contract test that the rejection precedes ingestion.

GaokaiZhang:
Resolved in ecfb847 (T7 overlap in moss_tts_local/request_builders.py).

SandyLuXY (approved):

PR #937, opened 2026-07-03
[TTS Refactor][M9] Fish -> OmniScheduler: migrate the AR stage, delete FishScheduler

GaokaiZhang:
Part of #661 (M9; audit doc: [09_fish_scheduler_audit](https://osgbw74w8zwb.sg.larksuite.com/docx/GNrsdU36Lo3KpExfZ7alWkhKgpe)). Same shape as #476: the S2-Pro AR stage moves onto the shared `OmniScheduler`, and the bespoke `FishScheduler` (591 LOC, the same four-class stack #476 removed for higgs) is deleted in the same PR. Net −816 lines. Stacked on #839 (first commit here); I'll rebase once it lands.

Behavior class: green with perf-check — same model math, sampling and stop conditions; what changes is batch composition and abort bookkeeping.

**Wiring.** Finish detection, batch planning and the KV/radix lifecycle come from upstream (fish's `Req` already carried stop/eos `im_end` and `max_new_tokens`). Stream chunks move to a `stream_output_builder` emitting the same `target="vocoder"` / `modality="audio_codes"` message; `engine_time_s` and the empty-output `ValueError` move into the adapters (higgs pattern). No `abort_callback`: fish row-syncs its model buffers from request data every step, so there is no per-request model slot to release.

**What the migration surfaced:**
- `Req.vocab_size` must be `len(tokenizer)`: the semantic tokens live in the added vocab (up to id 155773), and upstream `check_finished`'s vocab-boundary guard — which FishScheduler never executed — otherwise rewrites every request's first sampled code to `im_end` and finishes it.
- `Req.extra_key` now fingerprints the ref VQ codes: the prompt embeds only cb0 as token ids while cb1..9 arrive as embeddings, so two reference audios with an identical cb0 sequence must not share radix KV (the hole higgs closed in #476).
- the request budget is clamped to `context_length − 1 − len(prompt)`: OmniScheduler pre-rejects requests that cannot fit, but FishScheduler used to admit them and stop at `im_end`; clamping keeps that request class served.
- two shared `omni_scheduler.py` fixes that apply to higgs as well: `_emit_stream_output` now skips aborted rids (an abort landing mid-step could ship one more chunk into the vocoder's stream), and `_aborted_request_ids` gets FishScheduler's 10000→5000 cap, evicting oldest-first.
- restored the `fish_qwen3_omni` arch mapping dropped in #435 — the documented `--model-path fishaudio/s2-pro` launch has been failing at startup since then.

**Deliberate deltas, all inaudible:** an `im_end` landing exactly on the length cap reports `length` instead of `stop` (frame counts identical in every case); results for aborted requests are emitted and dropped by the stage runtime instead of suppressed in-scheduler (higgs parity); the audited "per-step radix write" turned out to fire once per request at the prefill boundary (`decode_batch_idx == 0` is only true there — `prepare_for_decode` increments it before every decode step), which is exactly where upstream caches the same prompt-only `fill_ids` — no radix delta.

**Tests.** Model-level tests kept; controller-flavored ones re-anchored on `collect_s2pro_step_outputs`; new contract tests for the request builder (stop/eos/vocab/cap/clamp/extra_key sensitivity), result adapter (finish_reason, engine_time_s, empty-output error) and stream gating; two new shared scheduler tests for the abort fixes. Deleted scheduler-internals tests map to the shared suite: queued/running abort → `test_omni_scheduler_abort_cleans_queued_request_immediately` / `…_abort_marks_running_request_for_finish`, batch-failure cleanup → `…_run_batch_failure_emits_error_and_aborts`, adapter-error emission → `…_result_adapter_failure_emits_error_without_raise`; planner/controller internals died with the file. Abort-in-flight exercised e2e: 16 concurrent streams, half aborted mid-generation, follow-up batch clean, no vocoder crash, no stuck KV.

**Gate** — full SeedTTS EN (1088), `--lang en --max-concurrency 16`, DP1, no router, `--max-running-requests 64 --cuda-graph-max-bs 64`, H100:

| s2-pro, main @1cb04ad, nonstream | en | 1.09% | 65.21 | 1088/1088 | 0 / 0 | 1.19 req/s | 3.63 |
| s2-pro, this PR, nonstream | en | 1.16% | 64.90 | 1088/1088 | 0 / 0 | 1.18 req/s | 3.67 |
| s2-pro, main @1cb04ad, stream | en | 1.16% | 65.34 | 1088/1088 | 0 / 0 | 1.22 req/s | 3.55 |
| s2-pro, this PR, stream | en | 1.32% | 65.27 | 1088/1088 | 0 / 0 | 1.17 req/s | 3.69 |
| s2-pro, this PR, stream (repeat) | en | 1.20% | 65.01 | 1088/1088 | 0 / 0 | 1.16 req/s | 3.72 |

Stream TTFP mean/p95 12.04/16.35 s → 12.50/16.99 s; inter-chunk mean/p95 0.46/1.36 s → 0.48/1.40 s. Sampling is unseeded, so per-run WER moves ~±0.1pp (the stream repeat shows the spread); speaker similarity is flat and the throughput/rtf deltas stay within the noise of two concurrent-GPU runs. Zero retracts in all server logs.


sglang_omni/models/fishaudio_s2_pro/request_builders.py
```diff
@@ -52,6 +57,20 @@ def validate_s2pro_top_k(top_k: int) -> None:
...
+    # note (Gaokai Zhang) only cb0 of the ref VQ codes becomes prompt token ids;
+    # cb1..N ride in as embeddings, so extra_key must hash all codebooks to keep
+    # same-cb0 prompts from sharing radix KV across different reference audio.
+    if not vq_parts:
+        return None
+    digest = hashlib.blake2b(digest_size=16)
+    for part in vq_parts:
+        codes = part.to(torch.int32).contiguous()
+        digest.update(str(tuple(codes.shape)).encode())
+        digest.update(codes.numpy().tobytes())
```
luojiaxuan, line 70: Could we make this helper device-agnostic before calling `.numpy()`? The current preprocessing path likely leaves `vq_parts` on CPU, but this helper accepts tensors directly and will fail if a future caller passes CUDA tensors. Hashing `part.detach().to(device="cpu", dtype=torch.int32).contiguous()` keeps the `extra_key` independent of where the request state was built.

luojiaxuan (commented):
Static review only: the migration direction looks right and this PR is net-delete overall. Please address the robustness nit inline.

tests/unit_test/fishaudio_s2_pro/test_tts.py
```diff
@@ -745,301 +541,120 @@ def test_fish_s2pro_mixed_batch_keeps_terminal_and_audio_state_separate() -> Non
...
-    scheduler._inflight_request_ids = set()
-    scheduler.resource_manager = SimpleNamespace(
-        free=lambda request: freed.append(request.request_id)
+def test_fish_tts_result_adapter_maps_finish_reason_and_engine_time() -> None:
+    request_builder, result_adapter, _ = make_tts_scheduler_adapters(
+        tokenizer=FakeFishTokenizer()
     )
+    data = request_builder(make_s2pro_payload(request_id="req-result"))
+    data.output_codes = [torch.tensor([[100], [1], [2]], dtype=torch.long)]
+    data.finish_reason = "length"
```
luojiaxuan, line 615: Can we add a scheduler-level Fish test that reaches max_new_tokens and asserts the final finish_reason is length? This adapter test sets the field manually, so it does not cover the deleted FishIterationController length path.

tests/unit_test/pipeline/test_scheduler.py
```diff
@@ -499,6 +504,64 @@ def test_omni_scheduler_abort_cleans_queued_request_immediately() -> None:
     assert cleaned == ["req-wait"]
 
 
+def test_omni_scheduler_emit_stream_output_skips_aborted_requests() -> None:
```
luojiaxuan, line 521: Can we keep a Fish-specific abort-during-step test here? This generic check only proves the stream message is skipped, while the deleted Fish tests also pinned output_codes mutation, result suppression, and deferred resource cleanup when abort lands mid-update.

luojiaxuan (commented):
Found two Fish equivalence test gaps. CI also still needs run-ci.

sglang_omni/models/fishaudio_s2_pro/request_builders.py
```diff
@@ -52,6 +57,20 @@ def validate_s2pro_top_k(top_k: int) -> None:
         )
 
 
+def _ref_vq_fingerprint(vq_parts: list | None) -> str | None:
```
SandyLuXY: `_ref_vq_fingerprint(vq_parts: list | None)` is private and currently called after `vq_parts` is normalized to tensors, but the signature does not document that contract and the body assumes every element has `.detach()`. Consider annotating the helper and dataclass field more narrowly, for example `list[torch.Tensor] | None`, so future callers do not treat it as accepting arbitrary list-like VQ data.

tests/unit_test/fishaudio_s2_pro/test_pipeline.py
SandyLuXY, line 444: Can we assert this fake OmniScheduler receives a callable `stream_output_builder`? The Fish migration depends on `make_adapters()` saving the third adapter and `extra_scheduler_kwargs()` passing it into OmniScheduler; without an engine-level assertion, the standalone adapter test would still pass if this wiring regressed.

sglang_omni/models/fishaudio_s2_pro/request_builders.py
GaokaiZhang: Done in 4a3771a, exactly as suggested.

tests/unit_test/pipeline/test_scheduler.py
GaokaiZhang: Added in dae920f (`test_omni_scheduler_fish_abort_during_step_suppresses_chunk_and_result`). Writing it caught a real gap: `stream_output` still shipped terminal results for aborted requests. It now drops them and runs the deferred abort callback there — higgs/moss release their pool rows through that callback, and `reset_request` is idempotent so the double-call race is safe. The test pins untouched `output_codes`, the suppressed result, and the deferred cleanup.

tests/unit_test/fishaudio_s2_pro/test_tts.py
GaokaiZhang: Added in dae920f — builds a real Req with `max_new_tokens=2`, drives `check_finished()` into the upstream length finish, and asserts the terminal payload reports `length` through `stream_output` + the result adapter.

sglang_omni/models/fishaudio_s2_pro/request_builders.py
GaokaiZhang: Narrowed both to `list[torch.Tensor] | None` in dae920f. Left `S2ProState.vq_parts` as-is — it's the un-normalized wire type.

tests/unit_test/fishaudio_s2_pro/test_pipeline.py
GaokaiZhang: Added in dae920f: the engine test asserts the fake OmniScheduler got a callable `stream_output_builder`.

SandyLuXY (approved):

PR #938, opened 2026-07-03
[router] Auto-size --max-connections to the worker pool

JiaxinD:
## Motivation

The router caps upstream connections pool-wide at `--max-connections=100` by default: one shared `httpx.AsyncClient` serves all workers, and the value mirrors httpx's own `DEFAULT_LIMITS.max_connections` rather than being sized to the pool. Any pool whose aggregate in-flight capacity exceeds 100 is silently under-fed. Measured on a saturated same-card DP3 Higgs pool (3 workers x `max_running_requests=64` = 192 in-flight): the default costs ~30% throughput and doubles p50, with workers idling at ~30/64 running and an empty queue while requests wait inside the router.

## Change

* `--max-connections` default: `100` to auto, `min(4096, 128 x num_workers)`. Explicit values still win.
* Any resolved value below `64 x num_workers` logs an under-feed warning at startup (covers both explicit values and auto hitting the 4096 cap on very large pools).
* Runtime worker registration logs the same warning when the grown pool outruns the startup-sized client (the shared client is sized once at startup; resizing it live is out of scope here).
* The resolved budget shows in the existing startup config log line.
* Docs row updated (`docs/basic_usage/omni_router.md`).

## Evidence

Same-window direct-to-workers = 100%; saturated 3-worker pool （DP3), continuous closed-loop seedtts voice-clone via `/v1/chat/completions`, 288 offered, 60s window.

| router config | agg qps | vs direct | router CPU (one core) | workers |
|---|---|---|---|---|
| default 100 (before) | 49.9 to 51.1 | 68 to 70% | ~65%, p50 doubled | under-fed ~30/64, queue 0 |
| manual `--max-connections 512` | 69.9 / 72.3 / 74.6 | 97 to 102% | ~12% | saturated 62/64 |
| auto-sized 384 (this PR, zero flags) | 71.5 / 73.1 | 96.8 / 97.5% | ~13% | saturated 62/64 |

## Tests

7 new unit tests: auto formula (3 workers resolves to 384), 4096 cap, explicit pass-through, under-budget warning, cap-outgrown warning, and the CLI default/explicit plumbing. The new-behavior tests fail on the pre-change code. Full router test directory: 120 passed (the 4 pre-existing failures reproduce identically on origin/main; environment-dependent, unrelated).

Root cause analysis and the full A/B are in the #907 diagnosis thread and the #921 roadmap (Phase 1 gate experiment). No overlap with #920 (`proxy.py` untouched).


sglang_omni_router/app.py
```diff
@@ -196,6 +200,15 @@ async def create_worker(request: Request) -> JSONResponse:
 
         worker = Worker(config=worker_config)
         workers.append(worker)
+        if config.max_connections < MIN_CONNECTIONS_PER_WORKER * len(workers):
+            logger.warning(
+                "max_connections=%d is below %d x %d workers after registration; "
+                "the upstream client is sized at startup and can under-feed the "
+                "grown pool",
```
Ratish1: lets use f formatting

sglang_omni_router/config.py
```diff
@@ -174,6 +189,23 @@ def _validate_workers(self) -> "RouterConfig":
...
+    def _resolve_max_connections(self) -> "RouterConfig":
+        if self.max_connections is None:
+            self.max_connections = min(
+                MAX_AUTO_CONNECTIONS, CONNECTIONS_PER_WORKER * len(self.workers)
+            )
+        if self.max_connections < MIN_CONNECTIONS_PER_WORKER * len(self.workers):
+            logger.warning(
+                "max_connections=%d is below %d x %d workers; the cap is pool-wide "
+                "(one shared upstream client) and can under-feed the pool at "
+                "saturation",
```
Ratish1: same

sglang_omni_router/app.py
JiaxinD: done

sglang_omni_router/config.py
JiaxinD: done

Ratish1 (approved):

PR #940, opened 2026-07-03
[TTS Refactor][T5a]Batch Vocoder Base (non-streaming)

AkazaAkane:
Part of #661 (T5a). Extract the non-streaming batch-vocoder scheduler skeleton into `BatchVocoderBase`: the shared base owns the single/batched `SimpleScheduler` wiring, while each model keeps only the three model-specific hooks (`prepare_item`, `decode_batch`, `store_result`). Same shape as the T5a template: Qwen3-TTS, MOSS-TTS and Voxtral-TTS all move onto the shared vocoder base; the per-model codec/vocoder math and ModelRunner stay model-specific. Net +98 lines.

Behavior class: preserving with perf-check — same model math and output post-processing; what changes is only scheduler plumbing and the decomposition of Voxtral's previous inline `_vocode` into the same three-hook shape.

**Wiring.** `BatchVocoderBase.build_scheduler()` provides both the single-item path and the batched path, awaits the model-specific `decode_batch`, and checks result-count consistency before calling `store_result`. Qwen3-TTS and MOSS-TTS delete their nested vocoder closures and now subclass the base directly. Voxtral-TTS is decomposed from its inline `_vocode`: warmup prepend moves into `prepare_item`, tokenizer decode moves into `decode_batch`, and warmup trim + fade-in stay in `store_result`.

**What the migration surfaced:**
- the vocoder abstraction boundary is scheduling-only: the shared base owns prepare → decode_batch → store_result orchestration, but model-specific codec/vocoder forward logic remains in each model file.
- Qwen3-TTS already had the target split, so the migration is mostly closure deletion plus hook wiring.
- MOSS-TTS also had the target split; its `decode_batch` remains item-wise because the existing MOSS decode path already decodes segmented delayed codes per request.
- Voxtral-TTS did not previously expose prepare/store hooks or batching. This PR lands the decomposition under `BatchVocoderBase` while preserving the warmup-prepend, trim and fade-in behavior.
- `decode_batch` is declared `async def` for a uniform scheduler await interface. The in-scope codec calls remain synchronous internally.

**Deliberate deltas, all intended to be behavior-preserving:** Qwen3-TTS and MOSS-TTS use `build_scheduler(max_batch_size=max_batch_size, max_batch_wait_ms=max_batch_wait_ms)` instead of locally constructing `SimpleScheduler`. Voxtral-TTS also uses `BatchVocoderBase`, but keeps `max_batch_size=1, max_batch_wait_ms=0` to preserve its previous non-batched runtime behavior while still sharing the scheduler skeleton. No codec math, model-runner layer, sampling behavior, or audio post-processing is intentionally changed.

**Tests.** Scheduler unit coverage updated to await the async batch function. Accuracy/perf coverage comes from the full SeedTTS EN parity gate below across all three migrated models.

**Dispatch microbenchmark** — local scheduler-only microbenchmark with 300 fake requests compared the old sync closure dispatch against the new `async def` + `run_until_complete` dispatch with identical fake work:

| Config | Pure dispatch delta | With 5ms/item work |
|---|---:|---:|
| `max_batch_size=8`, `max_batch_wait_ms=2` | +4 us/request | ratio 0.97 |
| `max_batch_size=1`, `max_batch_wait_ms=0` | +31 us/request | ratio 1.02 |

The async wrapper cost is microsecond-scale and cannot explain a 30-50% end-to-end regression. Qwen3-TTS uses the same new machinery and stayed neutral in the full benchmark.

**Gate** — interleaved SeedTTS EN parity run, H100 GPU0, DP1, no router, manual default server launch, SeedTTS EN 1088, `--lang en`, `--max-concurrency 16`, three interleaved runs per branch:

| Model | Branch | Completed | WER | Mean Lat | P95 Lat | P99 Lat | RTF Mean | QPS | Audio Throughput |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|
| Qwen3-TTS 1.7B | main | 1088/1088, 0 fail | 1.708% | 3.164s | 4.610s | 5.717s | 0.7606 | 4.865 | 19.946x |
| Qwen3-TTS 1.7B | this PR | 1088/1088, 0 fail | 1.859% | 3.216s | 4.534s | 6.314s | 0.7513 | 4.771 | 20.288x |
| MOSS-TTS | main | 1088/1088, 0 fail | 1.850% | 4.355s | 5.352s | 5.960s | 1.0294 | 3.657 | 16.369x |
| MOSS-TTS | this PR | 1088/1088, 0 fail | 1.675% | 4.286s | 5.250s | 5.826s | 1.0058 | 3.717 | 16.648x |
| Voxtral TTS | main | 1088/1088, 0 fail | 1.281% | 1.310s | 1.996s | 2.342s | 0.2309 | 12.120 | 69.339x |
| Voxtral TTS | this PR | 1088/1088, 0 fail | 1.264% | 1.327s | 2.017s | 2.464s | 0.2343 | 11.967 | 68.139x |

PR vs main median deltas:

| Model | Mean Lat | P95 Lat | P99 Lat | RTF Mean | QPS | Audio Throughput | WER |
| Qwen3-TTS 1.7B | +1.64% | -1.65% | +10.44% | -1.22% | -1.93% | +1.71% | +0.151 pp |
| MOSS-TTS | -1.58% | -1.91% | -2.25% | -2.29% | +1.64% | +1.70% | -0.176 pp |
| Voxtral TTS | +1.30% | +1.05% | +5.21% | +1.47% | -1.26% | -1.73% | -0.017 pp |

The original single-run MOSS and Voxtral performance regressions did not reproduce under interleaved repeated measurement. MOSS is slightly faster on this PR; Voxtral is within ~1-5% depending on metric. Qwen3 is neutral on mean/P95/RTF/throughput, with only a marginal P99-latency median at +10.44%; individual Qwen3 P99 runs were `main=[6.032, 5.717, 5.609]` and `PR=[6.535, 6.314, 5.529]`, so run03 was PR-faster and there is no clear stage-level profiling target. No code remediation was made because the repeated run does not show a reproducible MOSS/Voxtral regression.


AkazaAkane:
@JiaxinD @SandyLuXY  Could you please review and run ci?

sglang_omni/models/voxtral_tts/pipeline/stages.py
```diff
@@ -297,21 +298,16 @@ def _load_audio_tokenizer(checkpoint_dir: str, audio_config: dict, device: str):
...
+    _N_WARMUP = 2
+    _FADE_IN_MS = 10
 
-    logger.info("Loading Voxtral audio tokenizer for vocoding...")
-    audio_tokenizer = _load_audio_tokenizer(checkpoint_dir, {}, device)
+    def __init__(self, audio_tokenizer: Any) -> None:
+        self._audio_tokenizer = audio_tokenizer
 
-    def _vocode(payload: StagePayload) -> StagePayload:
+    def prepare_item(
```
SandyLuXY, line 310: nit: The old inline `_vocode()` had a useful comment explaining why the first audio-code frame is repeated as warmup context. The new class keeps `_N_WARMUP = 2` and `_FADE_IN_MS = 10`, but the strategy rationale is now less visible. A short class docstring or local comment would preserve that context.

AkazaAkane: I added back the original local comment and a class docstring. Should be CI ready.

sglang_omni/models/moss_tts/stages.py
```diff
@@ -196,7 +190,13 @@ def _decode_audio(
         )
         return waveform, sample_rate
 
-    def _store_vocoder_result(
+    async def decode_batch(
```
luojiaxuan, line 193: Can we add a small MOSS vocoder unit test too? This path now goes through BatchVocoderBase, while the only new scheduler test covers Qwen3.

sglang_omni/models/voxtral_tts/pipeline/stages.py
```diff
@@ -297,54 +298,75 @@ def _load_audio_tokenizer(checkpoint_dir: str, audio_config: dict, device: str):
...
             codes_with_warmup = torch.cat([warmup, audio_codes], dim=0)
-            warmup_samples = n_warmup * audio_tokenizer.downsample_factor
         else:
             codes_with_warmup = audio_codes
 
-        results = audio_tokenizer.decode_helper_batch_async([codes_with_warmup])
-        audio_np = results[0]
+        return state, codes_with_warmup
+
+    async def decode_batch(
```
luojiaxuan, line 334: Can we add a small unit test for this split? Voxtral carries warmup prepend, trim, and fade-in behavior, so this refactor should pin the sample-level output.

luojiaxuan (commented):
Found a couple of test gaps. CI also needs run-ci before merge.

sglang_omni/models/moss_tts/stages.py
AkazaAkane: added

sglang_omni/models/voxtral_tts/pipeline/stages.py
AkazaAkane: added

SandyLuXY (approved):

PR #941, opened 2026-07-04
[WIP on 869]Intra-node data-plane hardening: completion wait, router device-fallback, SHM backpressure

jzheng17:
Stacked on #869 — this PR targets the `phase0-intra-node-nvlink-relay` branch, so merging it folds these commits into #869; `main` is unaffected until #869 itself lands.

Three focused, independently reviewable fixes to the Phase-0 intra-node data plane (one commit each). Full rationale is in each commit message.

### 1. `fix(relay/cuda_ipc)` — completion busy-spin + credit leak on timeout
The CUDA-IPC completion wait polled with a loop calling `await asyncio.sleep(0)`, which never sleeps — pegging the event-loop thread at 100% CPU and hammering `cudaEventQuery` (driver-lock contention that steals CPU from kernel launches, the opposite of what a host/CPU-dispatch-bound server wants; cf. #907). Separately, on the 30s timeout the sender slot was never released, permanently leaking 1 of only 2 credits and eventually deadlocking puts. Fix: a shared spin-then-backoff helper (a short spin catches the common fast completion with no added latency; exponential backoff frees the CPU on slow/hung transfers) plus release-the-slot-on-timeout. The spin/backoff constants are provisional defaults — see the test & measurement plan comment. A `TODO(comm)` is left for a truly event-driven wait (eventfd + `add_reader`, or a named pipe).

### 2. `fix(comm/router)` — transport follows the actual tensor device
`CommRouter.outbound_stream` raised when a GPU-placed edge carried a CPU-resident chunk; it now falls back to SHM (the host transport) instead. Transport selection follows the actual tensor device, not placement alone. Also documents the same-node invariant in `_physical_outbound`. (Fixes the router half of the CPU-tensor review thread; the `stage_io` stream-metadata device round-trip is a separate follow-up.)

### 3. `fix(relay/shm)` — real bounded-outstanding backpressure
`ShmRelay` acquired its flow-control credit and released it immediately (fire-and-forget), so the semaphore provided no backpressure. The credit is now held until the receiver consumes the block — reusing the `unlink` the receiver already performs as the ack (poll `/dev/shm` with spin+backoff), and released on timeout too so an absent consumer cannot leak a slot. No new ack channel and no pre-allocated pool; deliberately scoped to the cold, CPU-only path (bulk tensors go via cuda_ipc/mooncake). Verified that every shm put path awaits completion, so the held credit is always released.

Not covered here (follow-ups): the `stage_io` stream-metadata device round-trip, the TensorRef port, and cross-node activation.

### DIFF
--- a/sglang_omni/comm/__init__.py
+++ b/sglang_omni/comm/__init__.py
@@ -1,17 +1,17 @@
 # SPDX-License-Identifier: Apache-2.0
 """Unified communication primitives for stage, rank, and node data movement."""
 
-from sglang_omni.comm.engine import CommEngine
-from sglang_omni.comm.router import CommRouter
 from sglang_omni.comm.data_ref import (
     BackendRef,
+    DataKind,
+    DataLayout,
+    DataRef,
     MetadataTensorRef,
     TensorMeta,
-    DataRef,
-    DataLayout,
-    DataKind,
     TransportKind,
 )
+from sglang_omni.comm.engine import CommEngine
+from sglang_omni.comm.router import CommRouter
 
 __all__ = [
     "BackendRef",
--- a/sglang_omni/comm/data_ref.py
+++ b/sglang_omni/comm/data_ref.py
@@ -70,7 +70,9 @@ class BackendRef:
         cls, *, transport: TransportKind, relay_info: dict[str, Any]
     ) -> "BackendRef":
         transfer_info = _dict(relay_info, "transfer_info")
-        return cls(transport=transport, info=relay_info, length=_int(transfer_info, "size"))
+        return cls(
+            transport=transport, info=relay_info, length=_int(transfer_info, "size")
+        )
 
     def to_dict(self) -> dict[str, Any]:
         return {
@@ -160,7 +162,9 @@ class DataRef:
             layout=DataLayout(_str(value, "layout")),
             buffer=BackendRef.from_dict(_dict(value, "buffer")),
             header=_optional_str(value, "header"),
-            tensors=tuple(TensorMeta.from_dict(item) for item in _list(value, "tensors")),
+            tensors=tuple(
+                TensorMeta.from_dict(item) for item in _list(value, "tensors")
+            ),
             shape=_ints(value, "shape") if "shape" in value else None,
             dtype=_optional_str(value, "dtype"),
             offset=_int(value, "offset") if "offset" in value else None,
--- a/sglang_omni/comm/engine.py
+++ b/sglang_omni/comm/engine.py
@@ -7,8 +7,8 @@ from typing import Any
 import torch
 
 from sglang_omni.comm import stage_io
-from sglang_omni.comm.router import CommRouter
 from sglang_omni.comm.data_ref import DataRef, TransportKind
+from sglang_omni.comm.router import CommRouter
 from sglang_omni.proto import StagePayload
 from sglang_omni.relay.base import Relay
 
--- a/sglang_omni/comm/router.py
+++ b/sglang_omni/comm/router.py
@@ -51,6 +51,11 @@ class CommRouter:
         return self._physical_outbound(target)
 
     def _physical_outbound(self, target: str) -> TransportKind:
+        # Invariant: anything not in remote_stage_names is assumed same-node, so a
+        # future placement pass MUST populate remote_stage_names for every
+        # cross-node edge -- otherwise a cross-node target silently falls through
+        # to cuda_ipc/shm here. A hard assertion needs the Phase-1 node config,
+        # which does not exist yet.
         if target in self.remote_stage_names:
             return TransportKind.MOONCAKE
         if self.self_is_gpu and target in self.gpu_stage_names:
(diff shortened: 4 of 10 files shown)


jzheng17:
### Test & measurement plan for completion/consume-wait constants - fix(relay/cuda_ipc) cont'd

**The completion/consume-wait constants are provisional defaults, not measured** — `_WAIT_SPIN_SECONDS`, `_WAIT_BACKOFF_{MIN,MAX}_SECONDS` (cuda_ipc) and `_CONSUME_*` (shm). The wait path already emits `elapsed_ms` and `polls` per transfer via `_comm_trace` (`cuda_ipc_put_wait_ack` / `cuda_ipc_get_wait_copy`). Procedure (H100, per #907 methodology; fengyuanyu1 offered runs):

1. Collect the wait-time distribution per edge on a realistic Qwen3-Omni speech workload; bucket `elapsed_ms` by edge (small latency-sensitive, e.g. `talker_ar -> code_predictor`, vs bulk, e.g. `encoder -> thinker`).
2. Set the spin window to about the p90-p95 of the small edges' completion time, so the latency-critical common case is caught by spin (no backoff) while we never spin on multi-ms bulk copies.
3. Set the backoff cap well below the bulk edges' copy time (<= ~10%), so tail detection latency is negligible next to the copy.
4. Consider exposing the constants as env/config overrides so profiling can sweep them without recompiling.

**Success criteria (before vs after, concurrency 1 -> 8 -> 32):** event-loop-thread CPU and `polls`/transfer drop by orders of magnitude; `cudaEventQuery` rate / driver-lock contention down; **no regression** in decode p50/p99 TPOT or throughput (guardrail); slot starvation is tracked separately (that is the `credits` knob).

**Per-patch functional tests:**
- **Completion fix** (CPU-only): drive `wait_for_completion` with an ack that never fires plus a short timeout -> assert `TimeoutError` AND the credit is released and subsequent puts do not deadlock (loop credits+1 times); assert `polls` stays small under a delayed ack (no busy-spin). Existing `test_cuda_ipc_relay.py` round-trips pass on H100.
- **CPU-tensor fallback** (CPU-only): `CommRouter` -- a GPU-placed edge with a CPU tensor selects SHM, not raise.
- **SHM backpressure** (CPU-only): with `credits=K`, `K+1` concurrent puts whose receivers do not consume -> the `K+1`-th put blocks until a slot frees; CPU-tensor round-trip for correctness.

**What runs where:** router / backpressure / credit-leak logic is CPU-only -> CI; cuda_ipc round-trips plus the spin/backoff profiling require CUDA -> H100 only. Treat the constants as unvalidated until an H100 run tunes them.

jzheng17:
Note on the failing `lint` check: it comes from pre-existing isort/black formatting debt in #869's own files (`comm/__init__.py`, `comm/engine.py`, `comm/data_ref.py`), surfaced because CI runs `pre-commit run --all-files` over the whole tree. #869's own CI is red on the identical hooks. The three commits in this PR (`relay/cuda_ipc`, `comm/router`, `relay/shm`) are lint-clean; I intentionally did not reformat #869's files to avoid colliding with the base branch's own cleanup.

jzheng17:
### Follow-up: making the CUDA-IPC completion wait truly event-driven (eventfd + `add_reader`)

The completion fix here is an interim `spin+backoff` that removes the busy-spin but still **polls** a shared-mmap ack byte on the sender side. The intended end-state is a genuinely event-driven wait. Preferred route: **eventfd + `loop.add_reader`** (fallback: a named pipe).

**Changes required (contained to `relay/cuda_ipc.py`):**
- `_AckMap` (today: a byte in a shared mmap the sender polls via `is_done`) becomes an **eventfd**: the receiver `write()`s it in `mark_done`; the sender registers it with `loop.add_reader(efd, cb)` and `await`s a `Future` the callback resolves.
- The ack identifier carried in `DataRef.buffer.info` (today `ack_path` + `ack_index`) becomes the eventfd. Because an eventfd is a **raw fd valid only in the creating process**, it must cross the process boundary — either **fork-inherited** or sent over a unix socket via **`SCM_RIGHTS`**. This fd-passing is the one extra piece of plumbing versus the path-based ack file.
- The sender-side `wait_for_completion` poll loop becomes an `add_reader`-driven `await` (no spin, no backoff, no timeout-by-polling).
- Cleanup: remove the reader + close the fd.
- **Not touched:** the router, the `DataRef` schema at large, transport selection, other relays, or the receiver-side CUDA-event wait (that half can independently move to `run_in_executor(event.synchronize)`).

**Benefits:**
- **Truly event-driven, zero-CPU:** the sender blocks until the receiver signals — no polling, no `cudaEventQuery` / `is_done` churn, no CPU burned while waiting (which matters in the host/CPU-dispatch-bound regime, cf. #907).
- **Native asyncio fit:** an eventfd is an fd, and asyncio watches fds directly via `loop.add_reader(fd, cb)` — the event loop itself wakes the waiting coroutine, with **no helper thread**. (A POSIX named semaphore, by contrast, has a blocking `sem_wait` that is not an fd, so it would need `run_in_executor` to borrow a thread.)
- **Prompt, clean failure detection:** a closed/errored fd surfaces peer death immediately, so we can fail fast with a real reason instead of the current blunt 30s spin-then-timeout.
- **No shared-file lifecycle race:** removes the ack mmap file whose owner/unlink handling can race on abort.

**Fallback — named pipe (FIFO):** if fd-passing proves impractical, a named pipe is **path-based** (opened by path like today's ack file, no fd-passing) and a FIFO read is likewise `add_reader`-drivable — so it keeps the event-driven property while avoiding `SCM_RIGHTS`. It is slightly heavier than eventfd and needs FIFO lifecycle management. A named semaphore is the least attractive of the three for asyncio (path-based but needs a thread).

JingwenGu0829:
Did a small fix on :

Added fail-loud handling for CUDA IPC ack timeout to avoid unsafe slot reuse, and unlink SHM blocks on timeout before releasing credit to avoid /dev/shm leaks.

To avoid bloating this pr-on-pr, I think it's better to merge this first, and then we do the further refinements such as https://github.com/sgl-project/sglang-omni/pull/941#issuecomment-4880699321 on the main pr (869) or new pr-on-pr. I will try to put the next improvement direction, or the final state of phase 0 in the RFC doc.

PR #953, opened 2026-07-04
perf(qwen3-omni): drop unused deepstack embeds from the talker projection

MelodyyyYin:
## What
`project_mm_aggregate_to_talker_ar` shipped the whole `thinker_inputs` to the talker. The talker prefill scatters only image/video/audio embeds (`talker_prefill.merge_prompt_modality`) and never reads the **deepstack** visual embeds -> drop them from the talker projection. On the #932 TensorRef fast path they'd otherwise arrive as unresolved refs the talker must never touch. Tests cover the tensor and ref cases.

Scoped to deepstack only. The `video/image/audio_embeds` fast/slow scatter divergence (the talker does use them when they are real tensors) is a separate question — see @luojiaxuan's thread — and is left unchanged here.

Refs #934.


tests/unit_test/qwen3_omni/test_tensor_ref_safe_edge.py
```diff
@@ -0,0 +1,105 @@
...
+        "image_deepstack_visual_embeds",
+        "video_deepstack_visual_embeds",
+        "deepstack_visual_embeds",
+    ):
+        assert dropped not in out
+    for kept in ("video_embeds", "image_embeds", "audio_embeds", "video_grid_thw"):
+        assert kept in out
+
+
+def test_talker_projection_keeps_used_ref_opaque() -> None:
```
luojiaxuan: I would avoid treating this as a safe-edge invariant. `video_embeds` is consumed by talker prefill when it is a real tensor, so keeping a thinker-only unresolved ref here codifies the #932 fast/slow path divergence. Could we drop or xfail the used-embed ref tests and keep this PR scoped to deepstack trimming?

MelodyyyYin: Done in ba7c979 — dropped the used-embed ref tests and scoped the PR to deepstack only. Renamed to `test_talker_projection.py`; it now just asserts the deepstack embeds are dropped (tensor + ref cases). The video/image/audio fast/slow scatter divergence stays an open question, not codified here.

luojiaxuan (approved):
LGTM

Issue #924, opened 2026-07-02
ASR Model Support & Optimization Roadmap


Ccyest:
## Goal

Support and optimize ASR models in SGLang-Omni. Currently tracking two model families:

1. **[MOSS-Transcribe-Diarize](https://huggingface.co/OpenMOSS-Team/MOSS-Transcribe-Diarize)** — end-to-end multi-speaker ASR + speaker diarization, supports up to ~90 minutes audio input
2. **[Fun-ASR-Nano](https://arxiv.org/abs/2509.12508)** — multilingual ASR based on SenseVoice SANM encoder architecture

### Active Investigation

- **#959** — MOSS-TD accuracy & throughput regression: SGLang Omni is ~2pp worse on CER and ~20% lower on req/s compared to vanilla SGLang serving the same checkpoint. Root cause investigation ongoing (audio front-end parity, adapter post-processing, sampling/stop-condition divergence, batch-dependent numerics).

---

### Architecture

```
Audio (16kHz) → WhisperFeatureExtractor → mel [B, T, 128]
  → WhisperEncoder (24L, d=1024)
  → VQAdaptor (Linear→SiLU→Linear→LayerNorm, project to LLM hidden_size)
  → time_merge (4× frame merge)
  → masked_scatter_ into text embeddings
  → Qwen3 LLM (28L, hidden=1024) → text with speaker IDs [S01]/[S02] + timestamps
```

| Component | Details |
|-----------|---------|
| Encoder | `WhisperEncoder` — 24-layer Transformer, d_model=1024, 16 attention heads |
| Adapter | `VQAdaptor` — 2-layer MLP (Linear→SiLU→Linear→LayerNorm), bias=True. Name is a historical artifact — no actual vector quantization; the adapter does continuous projection from encoder space (d×merge=4096) to LLM hidden_size (1024) via `masked_scatter_` embedding injection |
| LLM | Qwen3 (28L, hidden_size=1024), `tie_word_embeddings=True` |
| Special tokens | `audio_start`: 151669, `audio_end`: 151670, `placeholder`: 0 |
| Output | Speaker labels `[S01]`/`[S02]` + timestamps embedded in text |

### Done

- [x] **M-PR1** Model support + diarization output parsing (#914)
  - `MossTranscribeDiarizeForConditionalGeneration` SGLang model class + `VQAdaptor` adapter
  - HF config registration (`MossTranscribeDiarizeConfig`)
  - Pipeline config + stage factory + request builder
[...]

PR #954, opened 2026-07-04
refactor(moss-td): fold per-item encoder loop into a single forward in get_audio_feature

draftbk:
Addresses the **M-PR6 (batched inference)** item in #924.

## What

Replace the per-item encode loop in `get_audio_feature` with a single batched `whisper_encoder` forward (stack all chunks → one pass → reassemble per audio), matching the pattern in `whisper_asr` / `qwen3_asr`.

## Accuracy — no regression

movies800, A100-40GB bf16, 3 runs each @ concurrency 8:

| metric | baseline | this PR |
|---|---|---|
| CER (%) | 6.54 / 6.52 / 6.12 | 6.51 / 6.40 / 6.58 |
| cp_CER (%) | 12.34 / 13.03 / 11.70 | 12.68 / 13.02 / 12.97 |

This PR stays inside the baseline run-to-run band.

## The batching doesn't actually engage today

- **Observed** (profiled @ concurrency 16): `get_audio_feature` always receives a single item (`items=1`), even when the LLM prefill batches up to 14 sequences — so it degenerates to the per-item path (identical GPU time).
- **Root cause** (upstream): `mm_utils._get_chunked_prefill_embedding` loops per request (`embedding_items[items_size[i]:items_size[i+1]]`) and calls the embedder once per request. `qwen3_asr` shares this dispatch; `whisper_asr` uses its own `forward`.
- **Encoder cost** (single-request trace): ~13 ms encoder vs ~1–2 s of LLM decode, and the batch speedup is ~1.3× (measured, saturates by batch 16) — so even if engaged the end-to-end effect is minimal.

## Next steps (needs a call)

1. **Merge as cleanup** — aligns with `whisper_asr`/`qwen3_asr`, −14 lines; code comment notes `items` is always 1.
2. **Close** — runtime no-op.
3. **Enable in sglang core** — hoist encode out of the per-request loop in `_get_chunked_prefill_embedding`, flag-gated. Better as a general upstream option benchmarked on an encoder-bound model.

Leaning (1) or (2) — opening for maintainers to call.


draftbk:
Superseded by #971 — re-opened as an in-repo PR (same branch/commit) so CI can use the HF token to pull the gated MOSS-Transcribe-Diarize model. Fork PRs don't get repo secrets.

Issue #959, opened 2026-07-05
[Tracking] MOSS-Transcribe-Diarize: SGLang serving beats SGLang Omni on both accuracy and throughput (val_800)


zhaochenyang20:
<html>
<hr>
<h2>Background</h2>
<p>MOSS-Transcribe-Diarize is the MOSS team's multi-speaker ASR model (transcription + speaker diarization over a single audio input of up to ~90 minutes). It currently has two serving implementations:</p>
<ul>
<li><strong>SGLang</strong>: sgl-project/sglang#29850 — model config, <code>MossTranscribeDiarizeForConditionalGeneration</code>, and an OpenAI-compatible transcription adapter integrated into SRT</li>
<li><strong>SGLang Omni</strong>: #914 (model support), plus the movies800 evaluation and ASR CI stages from #942 → #955 (merged)</li>
</ul>
<p>Per our discussion with the MOSS team, the ASR model will be completed and maintained in <strong>SGLang Omni</strong>, since Omni provides the more complete ASR environment: dedicated eval entrypoints (<code>benchmarks/eval/eval_transcribe_diarize.py</code>), diarization-aware metrics (CER / cpCER / ΔCER), and the ASR CI stages that landed in #955. This also makes sgl-project/sglang#29850 a useful <strong>reference implementation</strong> for the investigation below.</p>
<h2>Problem</h2>
<p>Using the multi-speaker transcription benchmark infra now in Omni, we ran a head-to-head comparison on val_800 (movies800 validation split, 800 samples): the same model checkpoint served by SGLang vs. SGLang Omni ("Protocol" below denotes the serving stack). <strong>SGLang wins on both accuracy and throughput at every concurrency level we tested.</strong></p>
<h3>Accuracy — val_800</h3>

Protocol | Concurrency | cer_no_spk ↓ | cp_cer ↓ | delta_cer ↓
-- | -- | -- | -- | --
sglang | 8 | 6.464 | 13.692 | 7.228
sglang-omni | 8 | 8.405 | 17.130 | 8.725
sglang | 16 | 6.457 | 13.686 | 7.228
sglang-omni | 16 | 8.063 | 15.307 | 7.245
sglang | 32 | 6.462 | 13.708 | 7.246
sglang-omni | 32 | 8.198 | 15.789 | 7.591

<h3>Observations</h3>
<ol>
[...]

PR #969, opened 2026-07-05
[ASR] Default transcription requests to greedy decoding

Ratish1:
## Motivation

/v1/audio/transcriptions decodes at an effective temperature of 1.0 when the request does not set one. The endpoint only forwards temperature when the form field is present, so the SamplingParams dataclass default of 1.0 reaches the model request. Output is sampled instead of greedy, which leaves only 114 of 800 val_800 outputs reproducible across identical runs and costs about 1 pp cer_no_spk. This is the largest accuracy contributor confirmed in #959. The OpenAI transcription API defaults temperature to 0 and the sglang reference implementation decodes greedily, so 0.0 matches both.

## Modifications

build_transcription_generate_request now always sets temperature in the extra params, defaulting to 0.0 when the request does not provide one. Explicit client values pass through unchanged. Updated the builder unit test and added one for the explicit passthrough.

## Related Issues

Addresses the temperature part of the accuracy divergence in #959. The residual gap to the issue's original numbers did not reproduce on current main and stays open there. Part of #924.

## Accuracy Test

A/B on val_800 with MOSS-Transcribe-Diarize, one H100, concurrency 16, fresh managed server per run, two reps per arm. CER cells show rep 1 / rep 2. Full details in https://github.com/sgl-project/sglang-omni/issues/959#issuecomment-4884981708

| arm | cer_no_spk | cp_cer | identical outputs across reps (of 800) |
|---|---|---|---|
| no temperature sent (effective 1.0) | 6.529 / 6.535 | 11.859 / 13.147 | 114 |
| temperature=0 | 5.481 / 5.484 | 11.496 / 11.500 | 792 |

Note the default arm measures 6.53, close to the CI harness 6.61, not the 8.06 to 8.41 in the original issue table, so the larger reported gap likely involves a different commit or serving config. A separate dtype A/B ruled out bf16 vs fp16 (0.03 pp).

## Benchmark & Profiling

No throughput impact expected from a decoding default change. The qps difference between the two arms was within cross-GPU noise (details in the linked comment). The throughput gap reported in #959 is structural and not touched by this PR.

## Checklist

- [x] Add unit tests.

cc @Ccyest

### DIFF
--- a/sglang_omni/serve/openai_api.py
+++ b/sglang_omni/serve/openai_api.py
@@ -1608,8 +1608,9 @@ def build_transcription_generate_request(
         params["language"] = language
     if prompt is not None:
         params["prompt"] = prompt
-    if temperature is not None:
-        params["temperature"] = temperature
+    # note (aaron): the client layer fills in the SamplingParams default of 1.0
+    # when this key is absent. 0.0 (greedy) matches the OpenAI API, see issue #959.
+    params["temperature"] = temperature if temperature is not None else 0.0
 
     return GenerateRequest(
         model=model,
--- a/tests/unit_test/serve/test_openai_api.py
+++ b/tests/unit_test/serve/test_openai_api.py
@@ -836,6 +836,8 @@ def test_transcription_request_builds_asr_generate_request() -> None:
         "language": "en",
         "temperature": 0.0,
     }
+    omni_req = Client._build_omni_request(gen_req)
+    assert omni_req.params["temperature"] == 0.0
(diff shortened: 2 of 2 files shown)


Ccyest (approved):
we can merge once ci pass

zhaochenyang20:
Hey ratish, I think we should turn on the default best for MTD. So did you change that? Also, if it affect the accuarcy a lot, we should recalibrate the CI.

PR #930, opened 2026-07-02
[TTS Refactor][M4a] ReferenceEncodeService: MOSS-local cache cleanup

luojiaxuan:
## Motivation

MOSS-local already had model-local LRU + same-key single-flight in `CachedReferenceEncoder`. M4a should own that reusable cache/single-flight/failure-propagation skeleton in `ReferenceEncodeService`, while MOSS-local keeps the model-specific codec semantics and different-key batch coalescing owner.

## Module doc

- Lark design: https://osgbw74w8zwb.sg.larksuite.com/docx/PM8Ed39XAoHkLoxguvHlFiO5ghc
- Repo API doc: `docs/developer_reference/reference_encode_service.md`

## Issue

- #925

## Behavior class

Yellow / preserving intent. MOSS-local should keep the same path and data-URI artifact behavior, but the cache/single-flight mechanics now come from the shared M4a service.

## Old path deletion

Removes the local `CachedReferenceEncoder` cache/single-flight wrapper from `sglang_omni/models/moss_tts_local/stages.py` and replaces it with `_MossLocalReferenceEncoder`, a thin model-local adapter over `ReferenceEncodeService`.

`_BatchedReferenceEncoder` remains model-local and remains the only owner of different-key batch coalescing in this PR.

## Modifications

- Adds `_MossLocalReferenceEncodeHook` for MOSS-local path and data-URI reference inputs.
- Stores CPU int32 reference codes and returns caller-owned CPU long tensors, matching the old wrapper intent.
- Keeps file duration gates, `reference_path_cache_key(...)` revalidation, and file/data-URI keyspace separation.
- Keeps `MOSS_REF_AUDIO_CACHE=0` behavior as the startup kill switch.

## Contract tests

- [x] `tests/unit_test/moss_tts_local/test_pipeline.py::test_cached_reference_encoder_on_off_hit_bit_identical`
- [x] `tests/unit_test/moss_tts_local/test_pipeline.py::test_cached_reference_encoder_lru_eviction`
- [x] `tests/unit_test/moss_tts_local/test_pipeline.py::test_cached_reference_encoder_rejects_nonpositive_capacity`
- [x] `tests/unit_test/moss_tts_local/test_pipeline.py::test_cached_reference_encoder_true_concurrency_dedup`
- [x] `tests/unit_test/moss_tts_local/test_pipeline.py::test_cached_reference_encoder_failure_does_not_poison`
[...]

PR #926, opened 2026-07-02
[TTS Refactor][M4a] ReferenceEncodeService: base + FishAudio S2-Pro migration

luojiaxuan:
## Motivation

M4a extracts the reusable mechanics for ad-hoc TTS reference encoding: cache-keyed lookup, byte-bounded LRU storage, same-key single-flight, no-poison failure handling, and caller-owned artifact loading. FishAudio S2-Pro is the first production migration because its threaded preprocessing path can currently duplicate same-key reference encodes.

## Module doc

- Lark design: https://osgbw74w8zwb.sg.larksuite.com/docx/PM8Ed39XAoHkLoxguvHlFiO5ghc
- Repo API doc: `docs/developer_reference/reference_encode_service.md`

## Issue

- #925

## Behavior class

Yellow. Output artifacts should remain equivalent, but same-key concurrent ad-hoc references now share one in-flight encode and successful artifacts may be served from a bounded cache.

## Old path deletion

Replaces FishAudio S2-Pro's inline ad-hoc reference helpers in `sglang_omni/models/fishaudio_s2_pro/stages.py`:

- `_encode_reference_waveform(...)`
- `_encode_reference_audio(...)`
- `_encode_reference_data(...)`

The codec/waveform semantics moved into the Fish-specific hook; no M4b batch coalescing is introduced.

## Modifications

- Adds `sglang_omni/scheduling/reference_encoder.py` with `ReferenceEncodeKey`, `ReferenceEncodeHook`, and synchronous `ReferenceEncodeService`.
- Adds the developer API doc and links it from the docs index.
- Wires FishAudio S2-Pro supported ad-hoc reference payloads through the shared service while keeping model-specific normalization, keying, codec encode, and artifact policy local.

## Contract tests

- [x] `tests/unit_test/scheduling/test_reference_encoder.py::test_same_key_concurrent_single_flight`
- [x] `tests/unit_test/scheduling/test_reference_encoder.py::test_cache_hit_returns_loaded_artifact`
- [x] `tests/unit_test/scheduling/test_reference_encoder.py::test_key_none_bypasses_cache`
- [x] `tests/unit_test/scheduling/test_reference_encoder.py::test_exception_propagates_to_all_waiters_and_does_not_poison`
- [x] `tests/unit_test/scheduling/test_reference_encoder.py::test_artifact_larger_than_budget_is_returned_but_not_cached`
- [x] `tests/unit_test/scheduling/test_reference_encoder.py::test_lru_eviction_respects_max_bytes`
[...]

PR #970, opened 2026-07-05
[TTS Refactor][M4a] ReferenceEncodeService: base service (standalone)

luojiaxuan:
## Motivation

Land the shared ad-hoc reference-encode service on `main` on its own, so each model migration (FishAudio #926, Qwen3-TTS #927, MOSS-local #930) can retarget `main` **independently** instead of stacking. Nothing depends on model code here, so this is the natural first thing to merge.

## Behavior class

Green. The service is **not wired into any model** in this PR — pure addition of an unused module + doc + tests. No runtime path changes, so no accuracy/throughput/latency impact and nothing for CI to gate yet. Perf gating happens per-migration when a model adopts the service.

## Modifications

- `sglang_omni/scheduling/reference_encoder.py`: `ReferenceEncodeKey`, `ReferenceEncodeHook`, synchronous `ReferenceEncodeService`. Owns mechanics only — cache-keyed lookup, byte-bounded LRU via `StageOutputCache`, same-key single-flight, no-poison failure fanout, caller-owned artifact load, basic stats. Model semantics (normalize/key/encode/store/load/revalidate) live in the caller's hook.
- Depends only on the existing `StageOutputCache`; no new deps.
- Developer reference doc + `docs/index.rst` toctree entry.

### Hang-safe leader path

The leader runs `revalidate()` and `cache.put()` inside the same `try/except` as `encode_one()`: any exception there still drops the in-flight entry and fails the future. Otherwise a raising `revalidate()` (e.g. a reference file mutated during the encode window) would leave same-key followers blocked on a future that never resolves — permanent poison + timeout-length hangs under concurrency.

## Contract tests

`tests/unit_test/scheduling/test_reference_encoder.py`: single-flight merge, cache hit, key-None bypass, exception-does-not-poison, over-budget-not-cached, LRU/byte-budget eviction, revalidate-false-not-cached, follower-timeout isolation, stats, and the two new `revalidate`-raises regression tests.

## Follow-ups (retarget onto this once merged)

- #926 FishAudio S2-Pro migration — drop the base-service files, keep only the Fish hook.
- #927 Qwen3-TTS migration — retarget `main`.
- #930 MOSS-local migration — retarget `main`; converge to a single MOSS PR (the env-var-free version) and drop the duplicate.

## Validation

- [x] `python3 -m py_compile sglang_omni/scheduling/reference_encoder.py tests/unit_test/scheduling/test_reference_encoder.py`
- [ ] `pytest tests/unit_test/scheduling/test_reference_encoder.py` — run in CI (local env lacks torch/pytest).

### DIFF
--- a/docs/developer_reference/reference_encode_service.md
+++ b/docs/developer_reference/reference_encode_service.md
@@ -1,0 +1,125 @@
+# Reference Encode Service
+
+`ReferenceEncodeService` owns reusable mechanics for ad-hoc TTS reference
+encoding:
+
+- cache-keyed lookup;
+- byte-bounded LRU storage;
+- same-key single-flight while an encode is in progress;
+- failure propagation to waiters without caching failures;
+- artifact store/load conversion and caller-owned return values;
+- basic cache statistics.
+
+The service is for ad-hoc request references. Registered or uploaded voices
+continue to use `SpeakerArtifactCache`, because they have a different lifetime,
+key space, and invalidation path.
+
+## API Shape
+
+The implementation lives in `sglang_omni/scheduling/reference_encoder.py`.
+
+```python
+@dataclass(frozen=True)
+class ReferenceEncodeKey:
+    model_id: str
+    model_revision: str
+    encoder_id: str
+    encoder_config_hash: str
+    artifact_kind: str
+    input_key: str
+    options_key: str = ""
+
+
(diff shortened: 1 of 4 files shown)


luojiaxuan:
Superseded: reverting to the stacked approach (#926 base+Fish → #927 → #930 MOSS), merged in order. The base service stays in #926, so a standalone base PR isn't needed.

Issue #906, opened 2026-06-29
higgs-tts 在 H200上GPU利用率低


Jingsong-Yan:
model: bosonai/higgs-tts-3-4b
GPU：单卡 H200
1. 现象1：GPU 利用率低；
   1. 已经预计算好了codec，并构建了[读 codec] -> [TTS] -> [写 wav] 流水线，确保io不是瓶颈；
   2. 已按 stages.py 中 DEFAULT_MAX_CONCURRENCY = 64，并行请求64个请求；
   3. 生产1-2s的音频，大约在50qps；生产10s左右的音频，大约在15qps；
   4. GPU 功耗基本在150-300w（H700上限700w），nvidia-smi gpu利用率不超过50%。

<img width="1528" height="1106" alt="Image" src="https://github.com/user-attachments/assets/ae3e3747-82a8-42df-8102-b7ad3fb48ca6" />

2. 现象2：当客户端超过64次个请求，会请求失败。

zhaochenyang20:
We discussed yesterday. Jaxing and Ratish would take care of this PR after the refactor.

zhaochenyang20:
@Ratish1 But to me, Second issue that when the concurrency is over 64, the system would break down seems quite important to us. Ratish, could you take a look at this issue first? I think this is a bug and we should cover it.

zhaochenyang20:
我们用 MPS 来完成这个工作 https://sgl-project.github.io/sglang-omni/basic_usage/mps_dp.html

PR #977, opened 2026-07-06
[higgs] Enable --mem-fraction-static for the tts_engine stage

JiaxinD:
# [higgs] Enable --mem-fraction-static for the tts_engine stage

## Why

To run several Higgs replicas on one GPU (same-GPU data parallelism behind CUDA
MPS, part of #921 / #906), each replica needs a per-replica KV-pool memory knob so
the replicas fit side by side. Higgs had none: the `HiggsTtsEngineBuilder` pins
`mem_fraction_static=0.85` in `generation_defaults()`, and the global
`--mem-fraction-static` CLI flag was rejected at startup for Higgs because
`HiggsTtsPipelineConfig` declared no `mem_fraction_role_to_stage` (that flag only
applies to pipelines that map a role to a stage, e.g. Qwen3-Omni thinker/talker).
So one Higgs replica filled the card and a second could not be launched.

## What

Map the AR ("talker") role to the `tts_engine` stage:

```python
@classmethod
def mem_fraction_role_to_stage(cls) -> dict[str, str]:
    return {"talker": "tts_engine"}
```

That is the whole change. It mirrors how MOSS-TTS and the Qwen3-Omni pipelines
already declare the role map, so `--mem-fraction-static V` now flows through the
existing path with no new machinery:

* `apply_mem_fraction_cli_overrides` sets `tts_engine.runtime.sglang_server_args.mem_fraction_static = V`,
* the typed-runtime bridge injects it into the stage's `server_args_overrides`,
* `TtsEngineBuilder.build` merges `server_args_overrides` over `generation_defaults()`,
  so `V` wins over the `0.85` default.

When the flag is omitted, behavior is unchanged (the `0.85` default still applies).

## Tests

`tests/unit_test/higgs_tts/test_pipeline.py` adds four cases:

* `mem_fraction_role_to_stage()` returns `{"talker": "tts_engine"}`.
* `--mem-fraction-static 0.27` pins `tts_engine` to `0.27`.
* `--talker-mem-fraction-static 0.3` pins `tts_engine` to `0.3`.
* `--thinker-mem-fraction-static` is still rejected (Higgs has no thinker stage).

Failing-before / passing-after: before this change the role map is empty, so the
first three raise `typer.BadParameter` ("requires a pipeline with a supported
SGLang AR mem_fraction_static target"); after, they pass. Verified locally that
`HiggsTtsPipelineConfig.mem_fraction_role_to_stage()` flips from `{}` to
`{"talker": "tts_engine"}`. The CLI-apply cases exercise the same import path as
the existing Qwen3-Omni mem-fraction tests, which (like these) require the full
`sglang` + `fastapi` runtime to import; they run in CI. `ruff` and `black` are
clean on the changed files.

## Scope

One config method plus tests. No change to the `0.85` default, no other model, no
router. Follow-up doc (same-GPU DP guide) depends on this PR and references it.


Ratish1 (approved):
LGTM

tests/unit_test/higgs_tts/test_pipeline.py
```diff
@@ -1463,3 +1465,52 @@ def test_higgs_audio_codec_encode_batch_input_normalisation() -> None:
         assert torch.equal(
             r, ref
         ), f"input format {i} produced different codes than encode_reference"
+
+
+def test_higgs_mem_fraction_role_to_stage_targets_tts_engine() -> None:
+    # --mem-fraction-static is only accepted when a pipeline maps a role to a
+    # stage; higgs routes the AR ("talker") role to its sglang tts_engine stage.
```
zhaochenyang20: Follow our comments rules: https://github.com/zhaochenyang20/sglang-diffusion-routing/issues/32#issuecomment-4918483294

zhaochenyang20 (changes requested):

zhaochenyang20 (approved):

PR #973, opened 2026-07-06
encoder cuda graph for Moss TD

yijiangt:
## Motivation

Adds a per-chunk-count CUDA graph for the MOSS-Transcribe-Diarize Whisper encoder. The encoder is a fixed-shape, stateless pure function (`[num_chunks, num_mel_bins, input_feature_len]` → `[num_chunks, encoder_len, d_model]`). Only the batch dim (chunk count) varies, so we bucket over chunk count and pad up to the nearest captured bucket. 

## Modifications

 - `encoder_cuda_graph.py`: new `WhisperEncoderCudaGraphRunner` class. It captures one CUDA graph per chunk count, with warmup, a VRAM headroom check before each capture, and eager fallback if anything goes wrong.
  - `sglang_model.py`: `init_encoder_graphs()` builds the runner at startup, and the encoder now runs through it when a graph is available (otherwise fallback to eager).
  - `stages.py`: does the capture on startup; defaults to chunk counts `[1,2,4,8,16,32]`, which can be overridden with `encoder_graph_chunk_buckets`.

## Related Issues

Addresses the **M-PR4** (Encoder CUDA graph) item in #924. Continues #961

## Correctness

End-to-end (movies800, greedy, concurrency=1): 800/800 samples produce byte-identical `result_text` with vs without the encoder graph. The same eval run also shows all diarization metrics match exactly (speed metrics included below for reference):

  | metric | upstream | graph |
  |---|---|---|
  | cer | 6.55% | 6.55% |
  | cer_no_spk | 6.55% | 6.55% |
  | cp_cer | 12.96% | 12.96% |
  | cer_no_spk_cp_valid | 6.44% | 6.44% |
  | delta_cer | 6.52% | 6.52% |
  | cer_valid_samples | 784 | 784 |
  | cp_cer_valid_samples | 773 | 773 |
  | exact_matches | 23 | 23 |
  | failed_requests | 0 | 0 |
  | rtf_mean | 0.0202 | 0.0192 |
  | rtf_p95 | 0.0304 | 0.0290 |
  | throughput (req/s) | 4.653 | 4.948 |
  | latency_mean (s) | 0.215 | 0.202 |

## Benchmark & Profiling

Ran ASR CI stage 1 locally on **NVIDIA H100 80GB** (movies800, 2-worker router, c=16), 3 runs each, upstream vs this PR.

  **upstream (no encoder graph):**

  | run | CER | cpCER | CER-valid / cpCER-valid | RTF mean | RTF p95 | throughput |
[...]

PR #972, opened 2026-07-05
Encoder LRU cache for MOSS-Transcribe-Diarize: +17% throughput under load, +28.6% ceiling (M-PR3)

xinlij:
## What

Adds a **CPU-resident LRU cache for the audio encoder output** (WhisperEncoder + VQAdaptor) of MOSS-Transcribe-Diarize, wired into `get_audio_feature`. Identical audio (keyed on the `MultimodalDataItem.hash` fingerprint) reuses its encoded features instead of re-running the encoder.

Backed by `StageOutputCache` (LRU, 64-entry cap + byte budget, CPU-resident). Sized via `encoder_cache_size_bytes` (default 4 GiB in the pipeline config); `init_encoder_cache` builds it on the model in the stage factory.

Closes #924.

## Measured value

**Bottom line: on a loaded long-running server (KV pool pressured, radix evicting prefixes) — the realistic production regime — the encoder cache delivers +17.1% pass-2 throughput / −14.6% wall. That is ~62% of the +28.6% pure-encoder-skip ceiling, exactly what you expect when ~62% of prefixes get evicted.**

Benchmarked on SeedTTS (single worker, twice-run populate→timed so pass 2 is all repeated audio; 5-round alternating medians to control for launch/thermal drift). The value is conditional on whether SGLang's **radix/prefix KV cache** already covers the workload — in MOSS the audio is a prompt **prefix**, so identical audio produces an identical prefix radix can skip (encoder included). I isolated each factor with env kill-switches:

| Scenario | how it was produced | pass2 throughput vs cache-off | Verdict |
|---|---|---|---|
| **radix ON + KV pressure** (loaded server) | shrink KV pool to force radix eviction | **+17.1%** (71.2 / 60.9), wall −14.6% | **Realistic production case — encoder cache backs up radix-evicted prefixes** |
| radix OFF (ceiling) | `disable_radix_cache` | +28.6% (72.6 / 56.4), wall −22.2% | Pure encoder-skip upper bound |
| radix ON, KV roomy | default config, low load | ~0% (76.1 / 76.0) | No-op **by design** — radix has headroom and skips the prefix itself; confirms the cache is *complementary* to radix, not redundant |

Cross-checks behind the numbers:
- **Cache actually hits / radix actually covers**: instrumented encode-call counters showed that with radix on and no pressure, pass 2 barely re-enters the encoder (radix skips it) — which is why the cache looks neutral there, not because it's broken.
[...]

PR #979, opened 2026-07-06
feat(moss-td): opt-in torch.compile for the Whisper encoder

draftbk:
Opt-in `torch.compile(mode="reduce-overhead")` for the MOSS-Transcribe-Diarize Whisper encoder (#924), as an alternative encoder acceleration to the CUDA graph from #973 — the two are mutually exclusive, not stacked (reduce-overhead already fuses *and* CUDA-graphs the encoder). Enabling the new `encoder_torch_compile` stage arg (off by default) makes the encoder use compile **instead of** the CUDA graph — when the flag is on, the encoder CUDA graph is not built. Both mechanisms share one `encoder_chunk_buckets` arg (default `1..8`, ~4 min of audio); `dynamic=False` compiles per exact chunk count, so counts outside the warmed buckets fall back to eager.

When to use: encoder-bound, high-concurrency short-audio serving (see the numbers below). Leaving it off keeps the default #973 CUDA graph.

## Results

Test env: A100-40GB · bf16 · torch 2.11.0+cu130 · latest `main` (includes #972 CPU encoder cache and #973 encoder CUDA graph) · movies800 (96 samples) · greedy · fresh server per run · `mem_fraction_static=0.5`. 3 runs each; all 96/96 completed. LLM CUDA graph on for every row — only the encoder path differs.

### Isolated encoder forward (n=1, CUDA-event timed, ~19 warm samples)

| encoder path | ms | vs eager |
|---|---|---|
| eager | 10.32 (sd 0.12) | — |
| CUDA graph (#973) | 9.72 (sd 0.004) | −5.8% |
| compile (fusion only) | 8.59 (sd 0.13) | −16.7% |
| compile reduce-overhead | 8.20 (sd 0.13) | **−20.5%** |

### End-to-end throughput (qps), 3 runs each

| conc | eager | CUDA graph | compile | compile RO |
|---|---|---|---|---|
| 1 | 2.27 / 2.34 / 2.34 | 2.35 / 2.34 / 2.34 | 2.35 / 2.36 / 2.34 | 2.36 / 2.34 / 2.36 |
| 4 | 7.43 / 7.35 / 7.42 | 7.56 / 7.43 / 7.54 | 7.56 / 7.57 / 7.57 | 7.61 / 7.54 / 7.69 |
| 8 | 11.62 / 11.62 / 11.58 | 11.88 / 11.93 / 11.73 | 11.98 / 11.94 / 11.85 | 12.03 / 12.05 / 12.04 |
| 16 | 16.33 / 16.34 / 16.62 | 16.79 / 16.87 / 16.71 | 17.11 / 16.78 / 17.10 | 17.30 / 17.08 / 17.27 |

Means:

| conc | eager | CUDA graph | compile | compile RO | RO vs eager | RO vs CUDA graph |
|---|---|---|---|---|---|---|
| 1 | 2.315 | 2.345 | 2.350 | 2.353 | +1.6% | +0.3% |
| 4 | 7.401 | 7.510 | 7.565 | 7.614 | +2.9% | +1.4% |
| 8 | 11.606 | 11.847 | 11.921 | 12.040 | +3.7% | +1.6% |
| 16 | 16.426 | 16.788 | 16.999 | **17.218** | **+4.8%** | **+2.6%** |

CER: eager and CUDA graph are numerically exact; compile drifts ~0.05–0.2 pp from fusion numerics.


Ccyest:
please rebase and run a benchmark comparison against upstream main. Since CUDA Graph just merged.

sglang_omni/models/moss_transcribe_diarize/sglang_model.py
```diff
@@ -99,6 +100,22 @@ def init_encoder_cache(self, max_bytes: int) -> None:
...
+        p = next(self.whisper_encoder.parameters())
+        frames = int(cfg.max_source_positions) * 2
+        pos = torch.arange((frames - 1) // 2 + 1, device=p.device, dtype=torch.long)
+        with torch.no_grad():
+            for n in buckets:
+                feats = torch.zeros(
+                    n, int(cfg.num_mel_bins), frames, device=p.device, dtype=p.dtype
+                )
+                self.encoder_runner(feats, pos, None)
+
```
Ccyest: The default buckets (1,2,3,4) only cover roughly ≤2 minutes of audio; 
No eager fallback on long audio.

Suggested fix: similar to cudagraph, pre warm-up small bucket size like [1,9), which covers 4 minutes. And fallback to eager if exceeds it.

sglang_omni/models/moss_transcribe_diarize/sglang_model.py
```diff
@@ -75,6 +75,7 @@ def __init__(
         super().__init__()
         self.config = config
         self.whisper_encoder = WhisperEncoder(config.audio_config, quant_config)
+        self.encoder_runner = self.whisper_encoder
```
Ccyest: **rebase related from claude code**
Derived from the root-cause issue above (both mechanisms must coexist after the rebase). self.encoder_runner = self.whisper_encoder goes through nn.Module.__setattr__ and registers the same encoder under a second name in _modules; after compilation, the OptimizedModule hangs in as a third copy (encoder_runner._orig_mod.*). Consequence: state_dict() / named_parameters() now expose two to three sets of encoder parameter names. Today's load_weights happens to be unaffected (it only maps checkpoint names onto whisper_encoder.*, and loading happens before compilation — I traced the params_dict lookups), but anything that later relies on state_dict (checkpoint saving, weight refit, parameter accounting) will see duplicate entries. This is a landmine. Main's #973 already established the cleaner pattern: self._encoder_graph_runner = None is a plain attribute (None is not a Module, so nothing is registered), with an explicit if/else at the call site.

**Fix: **:
```
# __init__
self._compiled_encoder = None  # plain attr; not a submodule

# compile_encoder
self._compiled_encoder = torch.compile(self.whisper_encoder, dynamic=False)

# call site (post-rebase, precedence made explicit)
if self._encoder_graph_runner is not None:
    features = self._encoder_graph_runner.run(...)
elif self._compiled_encoder is not None:
    features = self._compiled_encoder(...)
else:
    features = self.whisper_encoder(...)
```

sglang_omni/models/moss_transcribe_diarize/stages.py
```diff
@@ -90,6 +90,8 @@ def create_sglang_moss_transcribe_diarize_executor(
     mem_fraction_static: float | None = 0.80,
     mm_embedding_cache_size_bytes: int = 0,
     encoder_cache_size_bytes: int = 0,
+    encoder_torch_compile: bool = False,
+    encoder_compile_buckets: Tuple[int, ...] = (1, 2, 3, 4),
```
Ccyest: **Rebase related from claude code:**
Derived from the root-cause issue. After the rebase, the same physical quantity — the per-forward chunk count of the encoder — ends up with two factory arguments: this PR's encoder_compile_buckets: Tuple[int, ...] = (1, 2, 3, 4) and #973's encoder_graph_chunk_buckets: list[int] | None = None (expanding to [1..8] by default). One is a tuple, the other a list; default coverage goes to 4 vs 8; and #973's capture() filters invalid values (int(x) >= 1), deduplicates, and skips buckets when VRAM is low, whereas compile_encoder takes the buckets verbatim — passing 0 warms an empty-batch graph, and duplicates re-run warmup. Two nearly-identical knobs in the same function signature is guaranteed user confusion.

Suggestion:
Unify into a single encoder_chunk_buckets argument shared by both mechanisms (their physical meaning of a bucket is identical), and consolidate the {int(b) for b in buckets if int(b) >= 1} normalization in one place.

draftbk: Done

sglang_omni/models/moss_transcribe_diarize/sglang_model.py
draftbk: Done

sglang_omni/models/moss_transcribe_diarize/sglang_model.py
draftbk: Fixed.

draftbk:
> please rebase and run a benchmark comparison against upstream main. Since CUDA Graph just merged.

Done, rebased onto latest main and re-ran the eval, 4-way comparison is in the description.

Ccyest (approved):

PR #984, opened 2026-07-07
fix(moss-td): mitigate greedy repetition loops on non-speech audio (#975)

### LINKED ISSUE #975: MOSS-TD: greedy transcription default can enter deterministic repetition loops on non-speech audio (−19% stage-1 throughput from one movies800 sample)
## Background

#969 defaulted `/v1/audio/transcriptions` to greedy decoding (temperature=0), fixing the ~1 pp `cer_no_spk` gap and the output reproducibility problem from #959 (114/800 → 792/800 identical outputs across identical runs). The accuracy side of that change is solid and reproduces.

While benchmarking #973 I could not reproduce pre-#969 throughput numbers on current main, and bisected the gap to #969 itself.

## Problem

Greedy decoding can enter a **deterministic repetition loop** on non-speech audio (laughter). One movies800 sample is enough to depress the ASR CI stage-1 throughput metric by ~19%.

### Bisect — same host (1×H200 pair, GPUs 5+6), ASR CI stage 1 methodology (2-worker router, c=16, movies800 × 800), runs minutes apart

| ref | contents | QPS |
|---|---|---|
| `2465af21` | main after #971, **before #969** | **36.64** |
| `478886ec` | `2465af21` + **#969 only** | **29.74 (−19%)** |
| `da258393` | #973 head (= `478886ec` + encoder graph) | 34.2–34.4 |

All runs 800/800 completed, 0 failed. Per-request latency mean is nearly unchanged (0.434s → 0.469s); the loss is a wall-clock tail effect (see mechanism).

### Root cause — live single-sample probe

Same server (`da258393`, single H200, mrr=16), same audio, only the `temperature` form field differs. `val_000280.flac` is a 9.7s clip whose reference is `[S1]Fools.[laugh][S1] Oops.` — it contains laughter:

| request | latency | raw server response (`text`) |
|---|---|---|
| `temperature=0`, val_000280 | **10.14s** | `[0.12][S01] Fools.[1.14][1.22][S01] Ha ha ha ha ha ha ha ha ha ha ha …` — repeats until the max_new_tokens cap |
| `temperature=1`, val_000280 | 0.38s | `[S01] Fools! [S01] Hahahaha! … [S01] Oops!` — terminates normally |
| control sample, either temperature | 0.08–0.09s | normal |

The model transcribes the laughter, the argmax chain enters a `ha → ha → ha` fixed point, and greedy decoding has no mechanism to escape it — it decodes ~2048 tokens (the adapter's max_new_tokens cap) instead of ~40. Sampling escapes the loop stochastically within a few steps. The behavior is fully deterministic: the same sample is the slowest request in every greedy run I have (9.5s at c=1 vs 0.23s median, ~20–21s at c=16, across 6+ independent runs and both #973 arms).

### Why no metric catches it

- **CER/cpCER**: the repeated `ha ha …` is stripped by text normalization — the final `result_text` is 11 characters and scores fine.
- **Failed requests**: none — the request completes.
- **p95 latency**: the straggler is beyond p95 (1/800).
- **Throughput**: the only visible symptom. In a c=16 closed-loop run of 800 samples (~22s wall), the straggler enters around t≈9s and runs ~20s, stretching the wall to ~27s — exactly the −19%.

## Why this is serious

1. **It pollutes every performance benchmark taken after #969.** Any A/B where one arm's base includes #969 and the other's doesn't is off by ~19% on this workload before the change under test does anything (this is how it surfaced in #973's review).
2. **Long-form audio makes it worse, not better.** movies800 is ~10s clips with one laugh sample in 800. Real long-form input (M-PR7 targets ~90 min) contains far more non-speech: laughter, music, applause — each segment is an independent chance to enter the loop, and each loop burns up to max_new_tokens of decode.
3. **Latency SLO**: a 40× per-request latency multiplier on affected inputs.

## Not a revert

Greedy's accuracy and reproducibility gains are real and match the OpenAI transcription API default. The fix should be loop mitigation on the ASR decode path, not going back to T=1.0.

## Candidate fixes

Roughly in order of implementation cost:

1. **Duration-scaled max_new_tokens budget** — a 9.7s clip should not be allowed 2048 output tokens. Budgeting ~20 tokens/s of audio caps the damage (10.1s → ~1s) without touching decoding semantics. Cheap stop-loss, does not fix the transcript.
2. **Repetition penalty / no-repeat-ngram on the transcription path** — breaks the loop and likely recovers the correct tail (`Oops.`).
3. **OpenAI-Whisper-style temperature fallback** — detect degenerate output (compression ratio is a ready-made detector; the looped output compresses extremely well) and retry the request at higher temperature. Most faithful to the reference behavior, most code.

## Action items

- [ ] Add a stop-loss (duration-scaled max_new_tokens) to the ASR request builders
- [ ] Decide on loop mitigation (repetition penalty vs temperature fallback) and implement
- [ ] Add val_000280-style non-speech samples to the ASR CI stage so the regression is gated
- [ ] Re-baseline stage-1 throughput expectations after the fix (current numbers are depressed by the straggler)

## Repro

```bash
# bisect (ASR CI stage 1 methodology)
git worktree add /tmp/wt-pre969 2465af21 && cd /tmp/wt-pre969
CUDA_VISIBLE_DEVICES=5,6 python3 -m pytest tests/test_model/test_asr_ci_multi_speaker.py -s -q
# → ~36.6 QPS; repeat on 478886ec → ~29.7 QPS

# single-sample probe
CUDA_VISIBLE_DEVICES=5 python3 -m sglang_omni.cli serve \
  --model-path OpenMOSS-Team/MOSS-Transcribe-Diarize --port 8033 \
  --max-running-requests 16 --cuda-graph-max-bs 16 --mem-fraction-static 0.80
AUD=$HF_HOME/hub/datasets--zhaochenyang20--movies800/snapshots/*/validation/audio/val_000280.flac
time curl -X POST http://127.0.0.1:8033/v1/audio/transcriptions \
  -F "file=@$AUD" -F "model=OpenMOSS-Team/MOSS-Transcribe-Diarize" \
  -F "response_format=verbose_json" -F "temperature=0"   # ~10s, "Ha ha ha ..."
# same with temperature=1 → ~0.4s, normal transcript
```

Part of #924. Related: #959 (root-cause thread), #969 (introduced the greedy default; its "no throughput impact" measurement read the −7% single-server delta as cross-GPU noise — it is this straggler), #973 (where the discrepancy surfaced).

cc @Ratish1 @db-ol

Ccyest:
Closes part of #924. Fixes #975.

## Problem

Greedy transcription (`temperature=0`, the `/v1/audio/transcriptions` default since #969) can enter a **deterministic repetition loop** on non-speech audio. On the movies800 laughter sample `val_000280` (`[S1]Fools.[laugh][S1] Oops.`) the argmax chain enters a `ha → ha → ha` fixed point and decodes to the `max_new_tokens` cap. That one request became a **15.5s straggler** that depressed ASR stage-1 throughput ~19% while being invisible to every gate: the `ha ha …` is stripped by text normalization (CER scores fine), the request does not fail, and it sits beyond p95.

## Fix — two independent guards

Both tuned on the full movies800 **800-sample × c=16 sweep** (single H200):

1. **Default `repetition_penalty = 1.05`, resolved at the transcription endpoint** (next to the `temperature=0` greedy default). It breaks the loop and recovers the correct tail. Callers can still override — `repetition_penalty=1.0` restores pure greedy. Resolved at the endpoint (not the client `SamplingParams`, which always injects `1.0`) so the default actually takes effect; models whose request builder ignores `repetition_penalty` (qwen3-asr) are unaffected.
2. **Duration-scaled `max_new_tokens` stop-loss** in the MOSS-TD request builder (`floor 256 + 40 tok/s of audio`). A clip's transcript cannot need more tokens than this generous budget, so any loop that survives the penalty stays latency-bounded instead of running to the cap. The rate is far above real transcript density (~15–20 tok/s incl. timestamps/speaker tags), so no legitimate transcript is truncated.

## Why 1.05 — the sweep

| rep | qps | cer_no_spk | cp_cer | clean-corpus cer | #>50%CER | lat_max | val_000280 |
|-----|-----|-----------|--------|------------------|----------|---------|-----------|
| 1.0 (+budget) | 27.8 | 7.45 | 14.98 | 4.66 | 29 | 3.02s | loops (136% CER) |
| 1.02 | 32.4 | 7.44 | 15.08 | 4.68 | 29 | 2.71s | **still loops** |
| **1.05** | 32.5 | **5.67** | **13.31** | 4.79 | 28 | 0.48s | Fools … **Oops.** |
| 1.1 | 33.4 | 5.98 | 13.77 | 5.17 | 28 | 0.30s | |
| 1.2 | 34.2 | 7.71 | 16.35 | 6.59 | **36** | 0.17s | |

- The loop breaks at **rep=1.03** (1.02 still loops). Aggregate CER is **U-shaped with a minimum at 1.03–1.05**; beyond that the penalty damages legitimate repeated tokens (Chinese reduplication like `这不…这不…`, repeated timestamps) and both `cer_no_spk` and the `>50%-CER` sample count climb.
- **1.05** is chosen: just past the loop-breaking threshold with a safety margin for the more varied non-speech in long-form audio (M-PR7), accuracy statistically tied with the 1.03 optimum, straggler eliminated (`lat_max` 15.5s → 0.5s).

Verified end-to-end: a plain request (no `repetition_penalty`) now returns `[S01] Fools. … Ah ha! [S01] Oops.` in 0.5s; `repetition_penalty=1.0` reproduces the loop (budget-bounded).

> Note: the 30 movies800 samples with CER > 50% are a **separate** issue — they are all short (1–4s), fast, hard-audio clips (Cantonese / Thai / singing / short laughs), not repetition loops. The repetition penalty does not (and is not expected to) move them; they are a model-quality matter, not a decoding one.

## CI

- Adds a **max-latency gate** (`MOSS_TD_MAX_LATENCY_S = 8.0`) to the movies800times assertions in `test_asr_ci_multi_speaker.py` — the loop straggler is a single-request tail that mean/p95 miss; the duration budget makes any per-request latency bounded, so an absolute (hardware-generous) max catches a reintroduced loop. Scoped to movies800times so it does not touch the long-form aishell4 stage.
- Unit tests for the duration budget, endpoint default resolution, and per-request override.

## Follow-up (not in this PR)

- Re-baseline `MOSS_TD_THROUGHPUT_QPS_REF` upward via the worst-of-N calibration on the CI host — the current `29.991` is depressed by the straggler this PR removes. The gate still passes (throughput only improves), but the reference is now conservative.


Ccyest:
## Benchmarks

1×H200, concurrency=16, single managed server, full sample sets. Candidate = this PR @5d81a952; baseline = `main` @144919ab. `repetition_penalty` is duration-gated (≤120s), so long-form is unaffected.

### movies800times (800 samples, `max_new_tokens=5120`)

| metric | baseline (`main`) | candidate (#984) |
|---|---:|---:|
| completed / failed | 800 / 0 | 800 / 0 |
| throughput (qps) | 25.35 | 26.58 |
| `cer_no_spk` (%) | 21.68 | **5.66** |
| `cp_cer` (%) | 29.21 | **13.31** |
| samples > 50% CER | 29 | 28 |
| latency mean (s) | 0.56 | 0.59 |
| **latency max (s)** | **20.18** | **3.54** |
| `val_000280` laughter clip (s) | **20.18** | **0.53** |

At the model's default `max_new_tokens` the loop garbage inflates aggregate `cer_no_spk` to 21.68% (matching the current CI ref 21.67 — i.e. the ref is dominated by loop output). Breaking the loop drops it to **5.66%**; the `ha ha …` laughter clip goes 20.18s → 0.53s.

### aishell4_long (20 samples, `max_new_tokens=65536`)

| metric | baseline (`main`) | candidate (#984) |
|---|---:|---:|
| completed / failed | 20 / 0 | 20 / 0 |
| `cer_no_spk` (%) | 13.87 | 13.41 |
| `cp_cer` (%) | 14.36 | 13.91 |
| speaker DER (%) | 10.39 | 10.15 |
| latency mean (s) | 265.1 | 266.7 |
| latency max (s) | 414.9 | 412.9 |

Long-form is at pure greedy (audio > 120s → penalty off), so it tracks the baseline within run-to-run noise — no diarization/latency regression. An un-gated global `rep=1.05` was measured to regress this stage badly (DER 10.4 → 16.8, missed-detection ~2×), which is why the penalty is duration-gated.

tests/test_model/test_asr_ci_multi_speaker.py
```diff
@@ -440,6 +442,15 @@ def _assert_movies800times_results(
         speed.get("rtf_p95"),
         MOSS_TD_RTF_P95_MAX,
     )
+    successful_latencies = [
+        o.latency_s for o in outputs if o.is_success and o.latency_s is not None
```
haojin2: Is `outputs` defined in this function?

haojin2: Is `outputs` defined in this function? Maybe should fetch latency from `results`?

haojin2 (changes requested):

tests/test_model/test_asr_ci_multi_speaker.py
Ccyest: `outputs` was indeed undefined in this scope. `compute_speed_metrics` now reports `latency_max_s`, and the check reads it from `results["speed"]` like the other latency metrics.

Ccyest:
## Benchmarks (re-run @1ff5b842, rebased comparison)

1×H200, concurrency=16, single managed server, full sample sets, all four runs sequential on the same otherwise-idle GPU. Candidate = `main` + this PR merged (PR head @1ff5b842); baseline = current `main` @b09dc995. `main` has moved past the PR base since the last run (now includes #972 encoder LRU cache), so the candidate was re-merged onto current `main` to keep the comparison apples-to-apples; no dependency changes between the refs. `repetition_penalty` is duration-gated (≤120s), so long-form is unaffected.

### movies800times (800 samples, `max_new_tokens=5120`)

| metric | baseline (`main`) | candidate (#984) |
|---|---:|---:|
| completed / failed | 800 / 0 | 800 / 0 |
| throughput (qps) | 24.59 | 25.45 |
| `cer_no_spk` (%) | 21.68 | **5.65** |
| `cp_cer` (%) | 29.21 | **13.30** |
| samples > 50% CER | 29 | 28 |
| latency mean (s) | 0.58 | 0.63 |
| **latency max (s)** | **20.58** | **4.21** |
| `sample-000280` laughter clip (s) | **20.58** | **0.54** |

Reproduces the earlier numbers: breaking the greedy loop drops aggregate `cer_no_spk` 21.68% → **5.65%**; the `ha ha …` laughter clip goes 20.58s → 0.54s. The new `latency_max_s` speed metric (added for the CI check in this PR) reads 4.21s on the candidate — under the 8s CI threshold.

### aishell4_long (20 samples, `max_new_tokens=65536`)

| metric | baseline (`main`) | candidate (#984) |
|---|---:|---:|
| completed / failed | 20 / 0 | 20 / 0 |
| `cer_no_spk` (%) | 13.42 | 13.30 |
| `cp_cer` (%) | 14.00 | 13.89 |
| speaker DER (%) | 10.15 | 10.12 |
| latency mean (s) | 239.5 | 239.2 |
| latency max (s) | 385.3 | 381.2 |

Long-form is at pure greedy (audio > 120s → penalty off), so it tracks the baseline within run-to-run noise — no diarization/latency regression. Both sides are faster than the previous run (latency mean ~265s → ~239s) because the baseline now includes the #972 encoder LRU cache.

Ccyest:
## Benchmarks — round 2 @5d02bb6c (after merging main / #989)

`main` now ships #989's `T=0.8/top_p=0.95/top_k=50` sampling defaults, so this PR was re-merged onto it: it keeps #989's `explicit_generation_params` plumbing but restores greedy decoding, handling the #975 loop with the duration-gated `repetition_penalty=1.05` instead. This round is therefore a head-to-head of the two loop-mitigation strategies. 1×H200, concurrency=16, single managed server, full sample sets, sequential runs on the same idle GPU. Candidate = this PR @5d02bb6c; baseline = `main` @c3b6bd13.

### movies800times (800 samples, `max_new_tokens=5120`)

| metric | baseline (`main`, T=0.8 sampling) | candidate (#984, greedy+rep) |
|---|---:|---:|
| completed / failed | 800 / 0 | 800 / 0 |
| throughput (qps) | 25.27 | 25.99 |
| `cer_no_spk` (%) | 5.89 | **5.66** |
| `cp_cer` (%) | 13.27 | 13.31 |
| speaker DER (%) | 22.13 | 21.79 |
| latency mean (s) | 0.63 | 0.61 |
| latency max (s) | 2.76 | 3.36 |

Both strategies break the repetition loop on short audio — within noise of each other here. The difference is determinism: greedy reproduces the same transcript every run, while sampling respreads (#989's own calibration: `cer_no_spk` 5.65–5.99 across 5 runs).

### aishell4_long (20 samples, `max_new_tokens=65536`)

| metric | baseline (`main`, T=0.8 sampling) | candidate (#984, greedy+rep) |
|---|---:|---:|
| completed / failed | 20 / 0 | 20 / 0 |
| `cer_no_spk` (%) | 13.64 | 13.83 |
| `cp_cer` (%) | 14.14 | 14.39 |
| **speaker DER (%)** | **15.60** | **10.46** |
| latency mean (s) | 250.9 | 240.2 |

The long-form diarization regression from T=0.8 sampling reproduces on H200: DER 15.60% under the current `main` defaults vs **10.46%** with greedy (audio > 120s → penalty gated off → pure greedy). #989's own H100 calibration showed the same spread (DER 11.9–22.7 across 5 runs). CER deltas are within greedy batching run-to-run noise.

CI threshold notes: accuracy refs from #989's calibration are kept (greedy+rep passes them with margin); movies800 speed refs carry headroom for the ~8% repetition-penalty latency overhead, and the aishell4_long speed maxes are restored to the pre-#989 greedy-calibrated values (the sampling speed-up tracked the truncated-output DER regression). Worst-of-N recalibration on the H100 CI host to follow.

tests/test_model/test_asr_ci_multi_speaker.py
Ccyest: Update: the PR has been slimmed down to just the sampling-parameter change (greedy defaults + duration-gated `repetition_penalty`, gate now inside the request builder) after #989 landed the explicit-params machinery. The max-latency check this thread was about is gone with the rest of the scaffolding — a recurrence of the #975 loop is already caught by the tightened `cer_no_spk` ref (loop output inflates it from ~6% to ~21.7%, far past the 1.1× slack).

Ccyest:
Closing in favor of #991 — same head commit (ce37a93f), re-opened from an in-repo branch. Root cause of the red ASR stage 1 here: fork PRs never receive `secrets.HF_TOKEN` (the job env shows it empty), and the MOSS-TD model repo is private, so the managed server 404s at startup before the benchmark even begins. Stage 2 passes because Qwen3-ASR is public. No token rotation can fix this for a fork PR; benchmark history and the repetition-penalty sweep remain here for reference.

PR #986, opened 2026-07-07
[Serving] Add same-GPU DP with CUDA MPS guide and KV token cap

JiaxinD:
## What

Adds `docs/basic_usage/same_gpu_dp.md` (linked from `docs/index.rst`) and
`examples/launch_same_gpu_dp.sh`: a conditional recipe for packing several serving
replicas on one GPU behind CUDA MPS, with a pinned single-GPU H100 Higgs case study.

## Why

A tuned single replica of a small TTS model inside the omni pipeline can leave much of
a high-compute GPU idle (host/pipeline-bound serving). CUDA MPS lets colocated
replicas share the GPU, which can convert that idle compute into aggregate throughput
when memory, CPU placement, and per-replica load allow it. The gate is measured
headroom on your own setup, not the GPU model; the guide includes a compute-bound
counterexample (MOSS-TTS-Local, ~13 qps ceiling regardless of replica count).

## Changes

* `docs/basic_usage/same_gpu_dp.md`: when it helps, tune-the-single-replica-first,
  private MPS daemon pattern, sequential launch with per-replica KV pool inspection,
  per-replica saturation, PID-level MPS attach verification, routing, ownership-based
  safe teardown, limits, and the case study.
* `docs/index.rst`: add the page to the General Usage toctree.
* `examples/launch_same_gpu_dp.sh`: `up`/`verify`/`down`/`list` launcher that records
  replica PIDs/process groups, ports, and logs in a per-run state directory, checks
  ports and refuses a second run on the same GPU, health-gates sequential startup,
  compares expected replica processes against the MPS client list (mapping saved per
  run), and tears down only the processes recorded for that run, keeping the state
  directory whenever cleanup cannot be confirmed. The launcher is a tested example,
  not a production process supervisor.
* Generation-stage `--max-total-tokens` is exposed so a run can explicitly cap each
  replica's KV token capacity, and `examples/launch_same_gpu_dp.sh` gains an optional
  `MAX_TOTAL_TOKENS` that applies the same cap to every replica (unset keeps the
  engine's auto-profiled behavior). This is for more controllable, comparable
  per-replica KV sizing; it is not a direct fix for `cudaErrorMpsRpcFailure`, and
  launch/runtime stability under a token cap has not been re-measured.

## Verification

Pinned setup: one H100 80 GB (driver 580.126.20 / CUDA 13), sglang-omni
`a78de4cb3edae4da1c5c49278d77aaa41d01e5b4`, sglang `0.5.12.post1`,
`bosonai/higgs-tts-3-4b` @ `7556c17e05201fccd9c8cc120bc216dcc7b5d561`,
`/v1/audio/speech` (seed-tts-eval EN, 300 samples per client), default
`max_running_requests=64` / `cuda_graph_max_bs=64`, interleaved rounds with fresh
servers on a shared host. All attempted runs are accounted for:

| Config | Completed / attempted | Aggregate qps | Relative to single |
|---|---:|---:|---:|
| Single c96 | 4/4 | 21.7–22.1 | baseline |
| DP2 + MPS, 2×c64 | 3/5 nominal | 31.5–37.7 | 1.4–1.7x |
| DP3 + MPS, 3×c64 | 2/4 nominal, +1 degraded completed | 39.9–46.9 nominal | 1.8–2.1x |

Nominal DP runs outperformed the default-cap single baseline, but DP2 and DP3 also
showed meaningful launch and runtime fragility on the shared host (one mid-run
`MpsRpcFailure`, two failed launches, one degraded-but-completed run — all reported in
the doc). DP3 was not clearly repeatably better than DP2, so the guide recommends
starting with DP2 and validating locally.

## Key findings and limitations

* `--mem-fraction-static` values are per-replica requests, not additive device shares;
  later replicas can receive much smaller KV pools, so the guide requires inspecting
  each replica's logged allocation.
* Whether the default admission cap binds the single baseline is runtime-dependent: it
  did not in our repeated runtime (queue stayed zero; raising the caps did not change
  throughput), but it did in the reviewer's one-shot runtime. The doc keeps the two
  runtimes separate.
* Scope: one H100, one model, one endpoint, one software revision, with shared-host
  variability. Not evidence for H200, multi-GPU scaling, or production stability;
  router overhead is not measured (direct-to-replica benchmarking by design).

## Validation

Sphinx docs build (no new warnings from this diff), pre-commit hooks, and `bash -n` on
the example script all pass; content checks confirm no leftover review notes or legacy
benchmark headlines. Raw benchmark artifacts are available on request.


zhaochenyang20:
Raised from fable when reproducing the performance gain:

# Same-GPU DP (MPS): please compare against a *tuned* single-replica baseline

## TL;DR
The `+114%` (DP3+MPS vs DP1) is baseline-dependent. On main+#977, H100 80GB, Higgs
(`bosonai/higgs-tts-3-4b`), when I give a **single replica the same cores + a full KV
pool + enough concurrency**, it beats DP3+MPS at equal cores and lower latency:

| config (30 cores total) | qps | p50 latency |
|---|---|---|
| **1 replica, mf=0.60, c=48** | **34.2** | 1.29 s |
| 1 replica, mf=0.60, c=32 | 29.2 | 1.03 s |
| DP3 + MPS, 3×(10 cores, mf=0.27), 3×c16 | 27.5 | 1.69 s |

DP1 in the guide (1 replica, few cores, low concurrency) is an *under-driven* baseline.
Higgs *is* host-bound (GPU util stuck ~62% while qps climbs 9→34 over c8→c48), but the
winning fix here is "one replica, more cores, more concurrency," not same-GPU DP.
Please re-run the DP-vs-baseline comparison against the recipe below before claiming a
net DP win.

## Strongest single-replica baseline recipe (what DP must beat)
Give the single replica everything a DP pool would collectively use: all the CPU cores
on the card's NUMA node, a large KV pool, and a concurrency high enough to saturate it.

```bash
export LD_LIBRARY_PATH=/lib/x86_64-linux-gnu:$LD_LIBRARY_PATH   # cu130 venv fix
CARD=0
NODE=$(cat /sys/class/drm/card$CARD/device/numa_node)          # card's NUMA node
# 1) Launch ONE replica, pinned to the node, large KV pool:
CUDA_VISIBLE_DEVICES=$CARD numactl --cpunodebind=$NODE --membind=$NODE \
  sgl-omni serve --model-path bosonai/higgs-tts-3-4b --model-name higgs \
  --mem-fraction-static 0.85 --host 127.0.0.1 --port 8000
# 2) Sweep client concurrency and take the PEAK qps (client on SEPARATE cores):
for C in 16 32 48 64 96; do
  numactl --membind=$NODE -C <cores-not-used-by-server> \
    --use-existing-server --generate-only \
    --host 127.0.0.1 --port 8000 \
    --model bosonai/higgs-tts-3-4b --meta zhaochenyang20/seed-tts-eval-arrow \
    --ref-format references --lang en --max-samples 500 --max-concurrency $C \
    --output-dir results/base_c$C
done
# Baseline qps = max over the sweep (subject to a latency SLO if you have one).
```
Report the baseline's peak qps, its latency at that point, and GPU util (nvidia-smi
`utilization.gpu` and, better, nsys SM-active). This is the number DP must beat.

## Fair head-to-head protocol (equal resources)
Hold the *total* GPU-node resources constant and only vary the split:

- Fixed budget: all CPU cores on the card's NUMA node, one 80 GB H100.
- Point A (baseline): 1 replica, mf≈0.85, all node cores, sweep concurrency → peak qps.
- Point B (DPk): k replicas, mf≈0.85/k... (or the largest that fits k), cores split
  evenly and *dedicated* per replica via `numactl -C`, MPS on, one client stream per
  replica; sum qps.
- Report for A and B: aggregate qps, p50/p95 latency, GPU util. DP "wins" only if
  B_qps > A_qps at comparable latency.

Two things that materially change the result and must be stated:
1. **Core pinning is mandatory.** Without `numactl` per-replica pinning, my DP2
   *regressed* below DP1 (8.5 vs 14.8 qps) — replicas starved each other on CPU
   dispatch. With dedicated cores, DP2 = 22.8.
2. **DP splits the KV pool too.** k replicas at mf=0.85/k each batch less than one
   replica at mf=0.85; that batch-efficiency loss (plus MPS overhead) is why DP3 (27.5)
   < 1 replica (34) here.

## Ask
Please state the DP1 baseline's exact cores + concurrency, and re-run Point A above.
If a tuned single replica still loses to DPk at equal cores/VRAM, the DP win is real and
I'll withdraw this. If not, the guide's headline numbers should be re-scoped to
"vs a single client stream," not "vs saturation."

(Env note: measured on main+#977, 8×H100 80GB, driver 580 / CUDA 13; full data and
scripts in /data/chenyang/mps_dp_study/FINDINGS.md.)

JiaxinD:
@zhaochenyang20  Thanks for the detailed write-up. I went back and re-audited my side and realized there may be a few additional confounding variables besides the benchmarking protocol itself, so before arguing about the numbers, I'd like to make sure we're comparing exactly the same setup.

Would you mind sharing a few details from the run that produced the results in this comment?

* Exact `sglang-omni` commit SHA, or the branch/head you used
* Whether `#977` was merged, cherry-picked, or applied as a local patch
* The corresponding `sglang` revision/version
* The exact server launch command and benchmark command, if they differ from the comment
* The effective runtime values for `max_running_requests` and `cuda_graph_max_bs`, or the startup log where these are printed
* Any local modifications or dirty working tree, if applicable

I'll pin my environment to the same revision and rerun on the same H100 setup. If I still see a discrepancy after everything is aligned, I can then narrow it down to the remaining variables instead of debating across different software revisions.

If it's easier, feel free to just point me to the branch/commit and logs you used, and I can reproduce from there.

zhaochenyang20:
# Same-GPU Data Parallelism (MPS) — verification report for PR #986 / #977

Verification of the "Same-GPU DP (MPS)" guide (#986) and the Higgs
`--mem-fraction-static` enabler (#977), run on our 8×H100 box.

---

## 1. TL;DR conclusions

1. **The doc's core claim is correct.** For host-bound Higgs, DP+MPS on one H100
   roughly doubles throughput: **DP2 = +100%, DP3 = +116%** over a *tuned* single
   replica, at equal/lower latency. DP3's +116% ≈ the roadmap's measured +114%.
2. **The gain is real only when two conditions hold**, and the guide should state both:
   - **Each replica must be driven to saturation.** Higgs' `max_running_requests=64`
     is a fixed default; a per-replica concurrency of ~16 leaves each replica at 25%
     batch and makes DP look like a *loss*. Drive per-replica concurrency near the cap.
   - **CPU core-pinning (`numactl`) is mandatory.** Without dedicated cores per replica,
     DP2 *regressed below* DP1 — replicas starve each other on host-side dispatch.
3. **MPS is the unlock.** Without MPS, 3 time-sliced replicas give ~+3% (they serialize
   on one GPU context); with MPS they co-reside and scale.
4. **It is model-dependent.** Higgs is host-bound (GPU never saturates → DP reclaims the
   idle). MOSS-TTS-Local becomes GPU-compute-bound (~13 qps ceiling) → same-GPU DP gives
   almost no peak gain there. "Measure the model's own idle first" (which the guide says)
   is the right gate.
5. **8×H100:** per-card DP2–DP3 + one MPS daemon *per GPU*; scales linearly across cards.
   DP3 adds only ~+8% over DP2 while being more memory-tight and launch-fragile.

> Honesty note: my first pass concluded "a single replica beats DP3." That was an
> **unfair test** — I under-drove each DP replica (c16). Corrected below. Credit to the
> reviewer for flagging the per-replica-concurrency / effective-batch confound.

---

## 2. Environment (pin to this to reproduce)

- sglang-omni commit: `a78de4cb3edae4da1c5c49278d77aaa41d01e5b4` (origin/main; #977 merged
  as `1046fbb4`); working tree clean.
- sglang: `0.5.12.post1`
- model: `bosonai/higgs-tts-3-4b` (`HiggsMultimodalQwen3ForConditionalGeneration`)
- dataset: `zhaochenyang20/seed-tts-eval-arrow`, EN split
- HW: H100 80 GB HBM3, driver 580.126.20 / CUDA 13; 2 NUMA nodes,
  node 0 = cpus `0-31` (+HT `64-95`), cards 0-3 on node 0.
- Runtime values (from startup log, identical across configs): `max_running_requests=64`,
  `cuda_graph_max_bs=64`, cuda-graph bs `[1,2,4,8,12,16,24,32,40,48,56,64]`.
  KV pool #tokens: 427355 @mf0.85 / 182700 @mf0.42 / 97503 @mf0.27 — all ≫ 64 requests,
  so batching is **not** KV-limited; the difference across configs is purely the DP split.
- **cu130 venv gotcha:** `export LD_LIBRARY_PATH=/lib/x86_64-linux-gnu:$LD_LIBRARY_PATH`
  before any launch, else CUDA Error 803 (host libcuda must precede the venv compat stub).

---

## 3. Results

### 3a. Higgs is host-bound (single replica, no MPS, all cores, isolated run)
| concurrency | qps | latency | rtf | GPU util (nvidia-smi) |
|---|---|---|---|---|
| 8  | 9.1  | 0.86s | 0.23 | 77% |
| 16 | 18.1 | 0.84s | 0.22 | 62% |
| 32 | 27.7 | 1.05s | 0.27 | 67% |
| 48 | 34.1 | 1.25s | 0.33 | **61%** |

qps keeps climbing, rtf stays ~0.2–0.3, and util is stuck ~62% → the GPU is never
saturated by one replica (low SM occupancy, consistent with roadmap #921's SM≈28%).
This idle is what DP reclaims.

### 3b. Fair comparison — each replica driven to saturation (one card, back-to-back)
Peak over a per-replica concurrency sweep {32,48,64,96}; one client per replica, pinned
to separate cores; identical workload/commit/sglang.

| config | cores | mem_fraction | peak agg qps | latency | vs single |
|---|---|---|---|---|---|
| single replica | 32 | 0.85 | 29.9 (c96) | 2.84s | base |
| DP2 + MPS | 16 × 2 | 0.42 | 59.7 (c64) | 1.93s | **+100%** |
| DP3 + MPS | 10 × 3 | 0.27 | 64.5 (c64) | 2.75s | **+116%** |

(Shared host; absolute qps drifts with box load — rely on back-to-back ratios.)

### 3c. MPS is the unlock (pinned, per-replica c16)
| config | agg qps | vs DP1 | latency |
|---|---|---|---|
| DP1 | 14.8 | base | 1.04s |
| DP3 + MPS | 27.5 | +86% | 1.69s |
| DP3, **no MPS** (time-slice) | 15.3 | **+3%** | 3.06s |

Time-slicing gives essentially nothing; MPS is what lets replicas co-reside.

### 3d. Core-pinning is load-bearing
Without `numactl` per-replica pinning, DP2 = **8.5 qps (BELOW DP1's 14.8)** — replicas
fought over CPU dispatch. With dedicated cores per replica, DP2 = 22.8 (and 59.7 when
also driven to saturation). For a host-bound model, pinning is not optional.

### 3e. Memory / packing (Higgs)
~23 GB per replica at mf=0.27 → **3 fit an 80 GB card**. DP4 at mf≈0.11 fails with
"Colocated GPU budget leaves no KV-cache headroom" → **4 do not fit 80 GB** (matches the
guide). Per-replica footprint ≈ `80·mf + ~10 GB` fixed (weights + MPS ctx + codec).

### 3f. Contrast: MOSS-TTS-Local is NOT host-bound
Single replica hits a **compute ceiling ~13 qps** (c16=6.9, c32=9.4, c48=12.6, c64=13.4,
util ~81%). Every arrangement converges there: DP2@c16=12.3, DP3@c16=13.0,
DP2@c32=13.0, 1rep@c64=13.4. So same-GPU DP gives ~0 peak gain for MOSS — it only shifts
the latency/throughput frontier. This is why the guide's "measure your own idle first"
gate matters.

### 3g. Multi-GPU scaling (MOSS DP2/card, per-GPU MPS daemon)
1 card = 12.5 qps, 2 cards = 27.2, 6 cards = 75.2 (≈12.5/card) — **linear**. A single
node-wide MPS daemon spanning multiple GPUs collapses (`cudaErrorMpsRpcFailure`); the fix
is **one MPS daemon per GPU** (isolated pipe dir; daemon `CUDA_VISIBLE_DEVICES=<card>`,
client `CUDA_VISIBLE_DEVICES=0` + that card's pipe).

---

## 4. Exact commands

Server (per replica; pin each to a DEDICATED core block on the card's NUMA node):
```bash
export LD_LIBRARY_PATH=/lib/x86_64-linux-gnu:$LD_LIBRARY_PATH     # cu130 fix
# MPS daemon (DP only), node-wide, NOT device-restricted:
env -u CUDA_VISIBLE_DEVICES nvidia-cuda-mps-control -d
CUDA_VISIBLE_DEVICES=$CARD numactl --cpunodebind=0 --membind=0 -C $LO-$HI \
  sgl-omni serve --model-path bosonai/higgs-tts-3-4b --model-name higgs \
  --mem-fraction-static $MF --host 127.0.0.1 --port $PORT
#   single: 1 rep, MF=0.85, cores 0-31
#   DP2   : 2 reps, MF=0.42, cores 0-15 / 16-31
#   DP3   : 3 reps, MF=0.27, cores 0-9 / 10-19 / 20-29
```
Benchmark (one client per replica, in parallel, on SEPARATE cores; sweep C, take peak):
```bash
numactl --membind=0 -C 64-95 \
  --use-existing-server --generate-only \
  --host 127.0.0.1 --port $PORT \
  --model bosonai/higgs-tts-3-4b --meta zhaochenyang20/seed-tts-eval-arrow \
  --ref-format references --lang en --max-samples 300 --max-concurrency $C \
  --output-dir results/$TAG
# agg qps = sum of per-replica summary.throughput_qps; sweep C over {32,48,64(,96)}
```
Teardown (order matters — never SIGKILL a live MPS client):
```bash
pkill -TERM -f "bin/sgl-omni"; sleep 8          # drain the card first
echo quit | nvidia-cuda-mps-control             # then stop the daemon
pkill -9 -f "bin/sgl-omni"; pkill -9 -f "multiprocessing-fork"   # then reap survivors
```

---

## 5. Operability findings (worth adding to the guide)
- **Drive each replica to saturation**; state the per-replica concurrency and note
  `max_running_requests=64`. An under-driven baseline mis-reports DP as a loss.
- **Core-pinning is mandatory** for host-bound models; put `numactl` in the main text,
  not just the example script, and explain why.
- **Teardown must also kill the stage-worker children** — they are `multiprocessing-fork`
  processes, *not* named `sgl-omni`; killing only the parent orphans them and they hold
  ~23 GB each → the next launch OOMs. Drain by GPU compute-apps, not by process name.
- **SIGKILL of a live MPS client wedges the server** (`cudaErrorMpsRpcFailure`) — the
  guide's warning is correct and I reproduced it.
- **Multi-GPU needs one MPS daemon per GPU** (shared daemon across GPUs collapses).
- **DP3 launch is racy under MPS** (`MpsRpcFailure` during concurrent cuda-graph capture);
  stagger launches / retry.
- The default launch example uses Higgs, which requires #977 — keep them ordered, or
  default the example to MOSS-TTS-Local (works today) with Higgs flagged as needing #977.

---

## 6. Recommendation for 8×H100
- Best robust config: **DP2 per card + one MPS daemon per GPU** (global DP16), linear
  across 8 cards; each replica driven near `max_running_requests`, cores pinned per replica
  on the card's NUMA node, clients on separate cores.
- **DP3/card (global DP24) is possible but marginal**: +~8% over DP2 per card, tighter on
  80 GB VRAM, and more launch-fragile. Prefer DP24 only if a run shows the extra replica
  still lifts aggregate qps on your checkpoint.
- Always gate on the model: profile single-replica GPU idle first (host-bound → DP helps;
  compute-bound like MOSS → it won't).

Full logs and scripts: `/data/chenyang/mps_dp_study/` (`fair_report.txt`, `fair_test.sh`,
`FINDINGS.md`).

zhaochenyang20:
I think there are two immediate things that we can do as following. The first thing is that try to implement the strictly equal KV cache strategy. That is adding a new parameter to set the KV cache with gigabytes, not with relative amount of the total left GPU. And after that, we should set up the MPS to coverage in our CI.

PR #987, opened 2026-07-07
[CI] Let Omni CI inherit runner GPU assignment

yxs:
This updates Omni CI jobs to inherit NVIDIA_VISIBLE_DEVICES from the assigned h100 self-hosted runner instead of hardcoding GPU 6,7 in the workflow container options.

The runner infra now assigns a fixed GPU pair to each h100 runner process. Keeping the workflow value inherited from the runner lets the existing baseline runner continue exposing GPU 6,7 while additional h100 runners can expose their own GPU pairs without changing the Omni CI workflow again.

This PR only changes workflow container options.


zhaochenyang20 (approved):
great job

PR #992, opened 2026-07-08
[RL] Add Higgs TTS rollout data support

Hayden727:
## What this PR does

- Implements the issue #780 Higgs `/generate` rollout contract with pre-tokenized prompts and sampled codebook actions, logprobs, and masks.
- Adds typed image, audio, and video inputs for message-tokenized or caller-tokenized `/generate` requests, plus reproducible processed-input metadata.
- Records exact Higgs behavior logprobs after temperature, top-k, and top-p filtering; forced cells remain masked and min-p/repetition penalty remain rejected.
- Keeps eager, batched, synchronous, async-decode, and CUDA-graph rollout paths aligned.
- Advertises the existing NCCL distributed-refit transport per stage and propagates served weight versions.


yxs:
Since @JingwenGu0829 has already separated out the multimodal input, let’s consider continue following the roadmap in smaller steps.

PR #994, opened 2026-07-08
[Intel][XPU] Enable Intel XPU's support for SGLang-Omni

siju-samuel:
## Motivation

Add Intel GPUs (XPU) support alongside the NVIDIA CUDA backend.
Qwen3-ASR, Qwen3-TTS, and Qwen3-Omni all serve end-to-end on Arc Pro B60
    ASR and TTS single-card, 
    Qwen3-Omni 30B-A3B thinker across 8 cards with tensor parallelism.

## Modifications
- Device abstraction (sglang_omni/utils/device.py): resolve the active accelerator (cuda/xpu/cpu), wrap the torch.cuda.* surface, and add a caps() capability registry so features gate on capability, not device identity. On CUDA it resolves to the exact prior calls (no regression).
- Serve / pipeline / runner: --device {cuda,xpu} flag; device-agnostic relay device string and set_device; multi-XPU TP device mapping (all cards visible + real gpu_id on XPU for XCCL); build ForwardBatch directly on sglang 0.5.13.
- SGLang 0.5.12 -> 0.5.13 compatibility: post-hoc memory-pool / attention / lora / cuda-graph init and renamed-attr backfill; 0.5.13 scheduler-state init and output-streamer routing; tolerant ServerArgs kwargs reconciler.
- XPU backend policy (central ModelWorker): SGLang selects the graph runner per device (xpu -> XPUGraphRunner, a torch.compile wrapper that stays eager and is None-safe on the forward path), so the graph path is left to SGLang rather than force-disabled. The CUDA-only kernel fallbacks — triton MoE runner (no FP8 CUTLASS) and fused_qk_norm_rope-off — are gated on their own caps (sgl_kernel/fp8), keeping them independent of the graph decision. No-op on CUDA.
- Seeded sampler (sglang_omni/sampling/xpu_sampler.py): pure-torch, bit-identical murmur_hash32 / multinomial_with_seed for XPU, rebinding SGLang's Triton path only on XPU (CUDA unchanged). Wired into the central ModelWorker XPU policy.
- Per-model device fixes for qwen3_asr / qwen3_omni / qwen3_tts: resolve device strings via the device layer, gate CUDA-only fast paths, run the Mimi codec on CPU on XPU (one-shot, off the hot path; the full KV pool leaves no device headroom for it), and other XPU-safe adjustments.
- Qwen3-Omni multi-XPU tensor parallelism (sglang_omni/utils/xpu_sglang_compat.py): correct SGLang's XPU free-memory accounting to real free memory (it over-reported total-minus-allocator, over-sizing the KV pool); raise the post-load TP barrier timeout for slow multi-rank shard loads; build the thinker request's origin_input_ids as array("q", ...) and wrap the custom omni forward in a 0.5.13 forward_context; tolerate the is_chunked -> inflight_middle_chunks rename.
- Packaging: pyproject_xpu.toml (+xpu torch pins, lean Qwen3 core + extras), scripts/xpu/install_xpu.sh, docker/xpu.Dockerfile.
- Docs / tests: README Hardware Support + Intel XPU install guide; device-layer, seeded-sampler, and XPU-feasibility unit tests under tests/unit_test/xpu/.

## Checklist

- [x] Add unit tests.

## CI

if labeled.

cc: @rbabukv


siju-samuel:
@FrankLeeeee @shuaills @FlamingoPg @ocss884 @zhaochenyang20 
 Could you please review and add the `run-ci` label?

sglang_omni/sampling/xpu_sampler.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    n, m = logprobs.shape
+    col_indices = torch.arange(m, device=logprobs.device)
+    hashed = murmur_hash32_torch(seed, positions, col_indices)
+
+    # Seeded Gumbel-max: u = hash / uint32_max, gumbel = -log(-log(u)). Clamp u
+    # to (tiny, 1 - eps) before the logs so both extremes stay finite.
+    _f64 = torch.finfo(torch.float64)
+    x = hashed.to(torch.float64) / float(_UINT32_MAX)
+    x = x.clamp_(min=_f64.tiny, max=1.0 - _f64.eps)
```
alexnails: This did not look right, checked. bit identical issue:

Upstream (`srt/layers/sampler.py:594-597`) clamps the **log result**:
```python
x = hashed.to(torch.float64) / torch.iinfo(torch.uint32).max
x.log_().clamp_(min=torch.finfo(x.dtype).min).neg_()  # -log(x)
x.log_().neg_()
```

sglang_omni/model_runner/model_worker.py
```diff
@@ -439,13 +538,70 @@ def _apply_model_worker_backend_policy(
...
+    if is_qwen3_omni_arch:
+        # Gate each CUDA-only-kernel fallback on its own capability.
+        if (
+            getattr(server_args, "enable_fused_qk_norm_rope", False)
+            and not _caps.sgl_kernel
+        ):
+            server_args.enable_fused_qk_norm_rope = False
+            logger.info("Disabling fused_qk_norm_rope (sgl_kernel unavailable)")
+        if (
+            moe_runner_backend in ("auto", "flashinfer_cutlass", "cutlass")
```
alexnails: Two bugs:

1. SM90: a bf16 Qwen3-Omni talker that previously resolved to `flashinfer_cutlass` now silently gets `triton`.

2. when `effective_quantization is None` -> FP8 CUTLASS MoE applicable is not the same predicate as "flashinfer_cutlass bf16 MoE applicable

sglang_omni/model_runner/model_worker.py
```diff
@@ -439,13 +538,70 @@ def _apply_model_worker_backend_policy(
     has_moe = _model_config_has_moe(model_config)
     has_native_fp8_block_quant = _model_config_has_native_fp8_block_quant(model_config)
 
+    # Non-CUDA: NVIDIA-only fast paths unavailable; gate each on its own
+    # capability (sgl_kernel/fp8) for the portable bf16 path. Graph capture is
+    # left to SGLang (xpu -> XPUGraphRunner).
+    from sglang_omni.utils import device as _omni_device
+
+    _caps = _omni_device.caps()
```
alexnails: Make `Capabilities` fields lazy (see the `device.py` comment) fixes both a bad reimport + some other runtime issues

sglang_omni/utils/device.py
```diff
@@ -0,0 +1,488 @@
...
+    try:
+        return bool(
+            cutlass_fp8_supported() and (is_sm90_supported() or is_sm100_supported())
+        )
+    except Exception:  # noqa: BLE001
+        return False
+
+
+@functools.lru_cache(maxsize=4)
+def caps(device_type: str | None = None) -> Capabilities:
```
alexnails: 1. L6 promises says no device init, but the CUDA branch calls `_cuda_fp8_cutlass_supported()` (device-property query → CUDA context init) and flash attn imports eagerly at construction. Since `caps()` is now called from `_apply_model_worker_backend_policy` before the model runner exists, that init happens early. We can make this lazy: `functools.cached_property` on a class, or storing callables so a caller that only needs `.sgl_kernel` never triggers the FP8 arch 

2.  _cuda_fp8_cutlass_supported() @ L398 duplicates _is_fp8_cutlass_moe_supported

sglang_omni/scheduling/sglang_backend/server_args_builder.py
```diff
@@ -2,10 +2,71 @@
...
 from typing import Any
 
 from sglang.srt.server_args import ServerArgs
 
+logger = logging.getLogger(__name__)
+
+
+# Map old ServerArgs field names to per-phase names used by newer builds.
+_SERVER_ARGS_RENAMES: dict[str, tuple[str, ...]] = {
+    "cuda_graph_max_bs": ("cuda_graph_max_bs_decode", "cuda_graph_max_bs_prefill"),
```
alexnails: This rename does not look correct?

sglang_omni/scheduling/generation_batch_policy.py
```diff
@@ -81,18 +106,32 @@ def validate_generation_batch_policy(
         server_args.max_running_requests,
         errors,
     )
-    cuda_graph_enabled = not bool(server_args.disable_cuda_graph)
+    # Default True (graph disabled) when absent, matching cuda_graph_batch_validator.
+    cuda_graph_enabled = not bool(getattr(server_args, "disable_cuda_graph", True))
```
alexnails: we should log this behavior?

[claude exp]
Defaulting the missing attribute to `True` means *graph disabled* → `cuda_graph_enabled = False` → **the entire batch-policy validation is silently skipped**. Same pattern in `cuda_graph_batch_validator.py:304`, and `_read_cuda_graph_field` returning `_NOT_EXPOSED` skips the sizing check with only an `info` log.

These validators exist to catch mis-sized CUDA-graph batches — on an upstream rename they now quietly stop protecting us instead of failing loudly. Suggest at minimum bumping these to `warning`, and ideally gating on the strict flag you already introduced in `server_args_builder.py` (`SGLANG_OMNI_STRICT_SERVER_ARGS`) so CI fails closed on API drift.

tests/unit_test/xpu/test_xpu_basics.py
```diff
@@ -0,0 +1,765 @@
+#!/usr/bin/env python
```
alexnails: this really is not a test suite but more of host check?

sglang_omni/model_runner/model_worker.py
```diff
@@ -439,13 +538,70 @@ def _apply_model_worker_backend_policy(
...
+            logger.info("Selecting 'triton' MoE runner (bf16, no FP8 CUTLASS)")
+
     if (
         model_arch_override == "Qwen3OmniTalker"
         and effective_quantization is None
         and moe_runner_backend == "auto"
     ):
-        # Note:(Chenchen Hong) flashinfer_cutlass MoE deadlocks CUDA-graph
-        # capture on H20 (no H20 kernel coverage); triton captures cleanly there.
+        # flashinfer_cutlass MoE deadlocks CUDA-graph capture on H20; use triton
```
alexnails: Some of these comment only issues are still prevalent, lets take a pass and not rm them ?

sglang_omni/sampling/xpu_sampler.py
```diff
@@ -0,0 +1,163 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Pure-torch replacement for SGLang's Triton seeded sampler.
+
+The Triton ``murmur_hash32`` kernel SIGSEGVs on XPU, taking down any seeded
```
alexnails: The profile did not have this issue? (edit: realize this is because of eager run)

sglang_omni/utils/device.py
```diff
@@ -0,0 +1,488 @@
...
+            fp8=_cuda_fp8_cutlass_supported(),
+        )
+
+    if dev_type == "xpu":
+        # See tests/unit_test/xpu/test_xpu_basics.py for the host probe.
+        return Capabilities(
+            device_type="xpu",
+            # SGLang picks the eager XPUGraphRunner per device; graph path is
+            # SGLang-managed, not force-disabled here.
+            cuda_graph=True,
```
alexnails: [asked claude to put eager comment given graph usage and that we run fully eager]

**Profiling data on what this `cuda_graph=True`-but-eager choice actually costs.**

From a Kineto trace of a Qwen3-TTS run on a single Arc Pro B60 (24.0 s span):

| | |
|---|---|
| GPU busy (union of all kernel + memcpy intervals) | **915.9 ms — 3.8% of span** |
| GPU idle | 23.10 s — 96.2% |
| Streams in use | **1** — no overlap |
| Kernel launches | 101,287 |
| Total kernel time | 866 ms → **8.6 µs avg** |
| `urEnqueueKernelLaunchWithArgsExp` | **3,981 ms → 39.3 µs avg** |
| `xpu_driver` events | 374,187 — ~3.7 per launch |

**The host spends 4.6× longer enqueuing each kernel than the GPU spends executing it.** And the idle isn't one big stall: 49,114 gaps >100 µs totaling 11.74 s, roughly 71 bubbles per decode step. Only 2.57 s is in the ten gaps >100 ms (largest 1,051 ms, presumably load/warmup).

That's the signature of a decode loop re-enqueuing ~150 kernels one at a time, every step, which is what "graph path is SGLang-managed" resolves to when the runner stays eager. Setting `cuda_graph=True` here is defensible as a dispatch signal, but the name now means the opposite of what a reader expects, and nothing in the pipeline reports that no capture is happening.

sglang_omni/utils/device.py
```diff
@@ -0,0 +1,488 @@
...
+            sgl_kernel=_module_importable("sgl_kernel"),
+            flash_attn=_module_importable("flash_attn")
+            or _module_importable("flashinfer"),
+            triton=_module_importable("triton"),
+            # FP8 CUTLASS needs a supported arch + SGLang, so probe.
+            fp8=_cuda_fp8_cutlass_supported(),
+        )
+
+    if dev_type == "xpu":
```
alexnails: We should probably be porting over SRTPlatform or making a new OmniPlatform that takes the latest version of SGL that ships with the XPU platform. This may cause a lot of work for this PR so we can talk about it over slack.

[Claude dive]

The Python hot path in a Qwen3-TTS XPU trace from this branch's validation run:

```
28,595  sglang/srt/layers/utils/multi_platform.py(81): forward
19,307  sglang/srt/layers/layernorm.py(573): forward_xpu
 4,644  sglang/srt/layers/activation.py(126): forward_xpu
 3,440  sglang/srt/layers/rotary_embedding/base.py(452): forward_xpu
```

Those `forward_xpu` methods are `MultiPlatformOp` dispatch, selected by `SRTPlatform.get_dispatch_key_name()`. So the SGLang build being profiled is **already resolving XPU kernels through the platform plugin system**, while this file maintains a second, parallel notion of device capability alongside it.

That's the concrete cost of the duplication: two device abstractions active in the same process, with `Capabilities` unable to see or influence what `MultiPlatformOp` picked. It's also why `caps().flash_attn = False` and the actual attention backend that got selected can disagree without anything noticing — see my comment on `model_worker.py`, where the run landed on `torch_native_backend` rather than the triton default this PR assumes.

The `torch.cuda.*` wrapper in this module genuinely has no upstream equivalent yet (most `DeviceMixin` methods are still `[Planned]`) and should stay. But this capability registry has a real counterpart that is demonstrably live in the profiled build. Deriving these fields from `current_platform` where the flags already exist — `supports_fp8()`, `support_cuda_graph()`, `is_pin_memory_available()` — would collapse the two views into one and make the eventual `platforms/xpu.py` adoption mechanical.

sglang_omni/model_runner/model_worker.py
```diff
@@ -439,13 +538,70 @@ def _apply_model_worker_backend_policy(
...
+        # XPU free-memory calc over-reports, so the KV pool over-allocates;
+        # correct it before the pool is sized.
+        patch_available_gpu_memory_for_xpu()
+        # Raise the post-load TP monitored-barrier so slow multi-rank shard I/O
+        # (large MoE across many cards) doesn't trip it.
+        patch_unbalanced_load_timeout()
+
+    # XPU attention backend defaults to SGLang's selection (triton). The
+    # 'intel_xpu' backend (+ paged KV) is opt-in via SGLANG_OMNI_XPU_ATTENTION.
+    _xpu_attn = os.environ.get("SGLANG_OMNI_XPU_ATTENTION", "").strip()
```
alexnails: The profile did not run on this backend?

pyproject_xpu.toml
```diff
@@ -0,0 +1,140 @@
...
+
+    # --- Model loading / HF -------------------------------------------------
+    "accelerate>=0.27.0",
+    "transformers>=5.6.0",         # aligned to the validated XPU SGLang line (5.8.x)
+    "safetensors>=0.4.3",
+    "pillow>=10.0.0",
+    "huggingface-hub>=0.36.0",
+    "datasets>=2.14.0",
+
+    # NOTE: sglang is intentionally NOT pinned here — install the XPU SGLang
```
alexnails: We need to pin an sgl version here.... there is function signature drift sadly cc: @zhaochenyang20

alexnails (commented):
first pass

sglang_omni/model_runner/model_worker.py
KMS07: answering to 1, it will happen on non CUDA right as _caps.device_type != "cuda"

Ratish1:
Hey @siju-samuel , could you rebase and fix the conflicts?. Thanks

sglang_omni/utils/device.py
```diff
@@ -0,0 +1,66 @@
...
+    build. Rewrite only that case -- an accelerator type this host does not have
+    -- so an explicitly requested accelerator, index included, is preserved.
+    ``index`` overrides the index when the caller supplies one (e.g. a stage's
+    ``gpu_id``). A ``"cpu"`` request stays on cpu and drops any index, since a
+    cpu stage is not bound to a card.
+
+    The spec is split by hand rather than through ``torch.device``, which rejects
+    a type this torch build does not know at all (e.g. ``"npu:0"``) -- the very
+    case this needs to retarget.
+    """
```
Ratish1, line 42: lets remove backticks from docstrings and similarly other places in the code

sglang_omni/platforms/__init__.py
```diff
@@ -19,6 +20,13 @@ def _is_npu_available() -> bool:
     return bool(npu.is_available())
 
 
+def _is_xpu_available() -> bool:
+    try:
+        return bool(torch.xpu.is_available())
+    except Exception:  # noqa: BLE001 - a probe must not break platform resolution
```
Ratish1: Could we follow the NPU probe above and only treat a missing torch.xpu attribute as unavailable?. Catching every exception also hides Level Zero or driver initialization failures, so platform resolution continues without preserving the real XPU error and startup fails later through the base path. Those runtime errors should be surfaced here.

sglang_omni/model_runner/model_worker.py
```diff
@@ -472,6 +473,34 @@ def _apply_model_worker_backend_policy(
...
+
+    if _device_type == "xpu":
+        from sglang_omni.utils.xpu_sglang_compat import (
+            patch_available_gpu_memory_for_xpu,
+        )
+
+        # Correct XPU free memory before the KV pool is sized against it.
+        # Graph capture is left to SGLang, which dispatches XPUGraphRunner by
+        # device (model_runner_components/cuda_graph_setup.py).
+        patch_available_gpu_memory_for_xpu()
```
Ratish1: Could we check the return value here? This patch is described as required to prevent Arc KV-pool overcommit, but if the SGLang import or required XPU memory API is unavailable it returns False and
startup continues with the accounting we already consider unsafe. should we fail with an actionable error when the correction cannot be installed.

sglang_omni/model_runner/model_worker.py
```diff
@@ -472,6 +473,34 @@ def _apply_model_worker_backend_policy(
...
+
+    if is_qwen3_omni_arch:
+        if (
+            moe_runner_backend in ("auto", "flashinfer_cutlass", "cutlass")
+            and _device_type == "xpu"
+        ):
+            # SGLang's XPU MoE path asserts the runner is 'triton' (see
+            # layers/quantization/unquant.py forward_xpu); 'auto' and the CUTLASS
+            # runners fail that assert, so pin triton here.
```
Ratish1: also, for all comments lets try to follow this rule https://github.com/zhaochenyang20/sglang-diffusion-routing/issues/32#issuecomment-4918483294

docs/get_started/installation_xpu.md
```diff
@@ -0,0 +1,161 @@
...
+sgl-omni serve --model-path /path/to/Qwen3-ASR-1.7B --host 0.0.0.0 --port 8000
+# transcribe:
+curl -s -X POST http://localhost:8000/v1/audio/transcriptions \
+  -F "file=@sample.wav" -F "model=/path/to/Qwen3-ASR-1.7B"
+```
+
+### Qwen3-TTS (text-to-speech, single XPU)
+
+```bash
+sgl-omni serve --model-path /path/to/Qwen3-TTS-12Hz-1.7B-Base --host 0.0.0.0 --port 8000
```
Ratish1, line 156: This Qwen3-TTS recipe is not runnable after either documented XPU installation. Neither pyproject_xpu.toml nor the Dockerfile installs the required qwen-tts package or the Python sox package, and the image also lacks the system sox binary documented by qwen3_tts/stages.py. Could we add the documented --no-deps Qwen3-TTS setup before this recipe (and to the image), so it does not
replace the pinned Transformers stack?

sglang_omni/models/qwen3_omni/components/thinker_model.py
```diff
@@ -4,12 +4,19 @@
 from typing import Any, Dict, Iterable, List, Optional, Tuple
 
 import torch
-from sgl_kernel import fused_qk_norm_rope
 from torch import nn
 from transformers import PretrainedConfig
 
+try:
+    from sgl_kernel import fused_qk_norm_rope
+except Exception:  # noqa: BLE001 - optional kernel; fall back to pure-PyTorch
```
Ratish1: XPU needs this CUDA kernel to remain optional, but pinned SGLang 0.5.16 already implements the Qwen3 path through sglang.jit_kernel.fused_qknorm_rope and can_use_fused_qk_norm_rope(). Could this thinker reuse that CUDA-only guarded path instead of catching every exception from sgl_kernel? That keeps the XPU fallback while preserving SGLang's compatibility checks and warning when the CUDA kernel cannot load.

sglang_omni/models/qwen3_omni/components/thinker_model.py
```diff
@@ -623,7 +634,10 @@ def __init__(
             prefix=add_prefix(prefix, "embed_tokens"),
         )
 
-        alt_stream = torch.cuda.Stream()
+        _dev_type = current_platform.device_type
+        alt_stream = (
+            torch.get_device_module(_dev_type).Stream() if _dev_type != "cpu" else None
```
Ratish1: Pinned SGLang only supplies Qwen3 MoE's alt_stream on CUDA. During graph capture,
the shared apply_qk_norm() implementation still enters torch.cuda.stream(alt_stream), so passing an XPU stream breaks the opt-in XPU full-graph path. Could we keep alt_stream=None off CUDA here, matching the pinned Qwen3 implementation? XPU eager execution still uses the normal Q/K path.

pyproject_xpu.toml
```diff
@@ -0,0 +1,101 @@
...
+build-backend = "setuptools.build_meta"
+
+# Intel XPU variant of pyproject.toml (see docs/get_started/installation_xpu.md).
+# Torch family pinned to +xpu; CUDA-only wheels dropped; deps scoped to the
+# XPU-supported models. sglang is verified against v0.5.16 but cannot be
+# pinned here -- every published wheel pulls CUDA torch.
+
+[project]
+name = "sglang-omni"
+version = "0.1.0"
```
Ratish1: should we use the same dynamic version source as the main pyproject here? currently the sglang omni version  is 0.1.1

docs/get_started/installation_xpu.md
```diff
@@ -0,0 +1,161 @@
+# 🚀 Installation — Intel XPU
+
+Installs `sglang-omni` for **Intel GPUs (XPU)**. The default
+[installation](./installation.md) pins CUDA-only wheels and would clobber a `torch+xpu` stack.
+Mirroring upstream SGLang ([Intel XPU docs](https://docs.sglang.io/platforms/intel_xpu.html),
```
Ratish1: this SGLang XPU documentation link returns 404. lets update it to
https://docs.sglang.io/docs/hardware-platforms/xpu

sglang_omni/scheduling/engine_factory.py
```diff
@@ -60,13 +60,16 @@ def build(
...
         from sglang_omni.scheduling import bootstrap as scheduling_bootstrap
         from sglang_omni.scheduling import sglang_backend
+        from sglang_omni.utils.device import remap_accelerator_spec
 
         checkpoint_dir = self.resolve_checkpoint(model_path)
-        if gpu_id is not None:
-            device = f"cuda:{gpu_id}"
-        gpu_id = int(device.split(":")[-1]) if ":" in device else 0
+        # Config.py passes a literal "cuda:0", which is not this host's backend.
+        device = remap_accelerator_spec(device, gpu_id)
```
Ratish1: The [RFC #1310](https://github.com/sgl-project/sglang-omni/issues/1310) requires the resolved device to be carried into `ServerArgs`. `device` becomes `xpu:N` here, but is not passed to `build_sglang_server_args()`, so SGLang detects again. Could we pass `xpu` into `ServerArgs` and retain `N` in Omni placement?

sglang_omni/pipeline/stage_workers.py
```diff
@@ -790,7 +791,32 @@ def _prepare_cuda_environment(
...
     log: logging.Logger,
 ) -> None:
-    """Map TP rank processes to one visible CUDA device before torch init."""
+    """Map TP rank processes to their accelerator before torch init.
+
+    CUDA isolates each rank via CUDA_VISIBLE_DEVICES + local gpu_id=0 (set_device
+    alone would not hide peer cards). On XPU that isolation hides peers and hangs
+    XCCL discovery, so keep all cards visible and address them by gpu_id instead.
+    """
+    if spec.tp_size > 1 and current_platform.device_type == "xpu":
```
Ratish1: the RFC #1310  places worker selector and environment preparation behind platform policy and before accelerator initialization. This `pop()` runs only after the spawned interpreter imports the process target, so it cannot guarantee that the mask is removed before imported code initializes XPU or Level Zero. `_patched_spawn_env()` already applies and restores temporary environment changes around `proc.start()`. Could `get_stage_process_env()` express removals as well, so `XPUOmniPlatform` owns how `ZE_AFFINITY_MASK` is handled while preserving the current XPU device-index behavior and keeping the shared launcher platform-neutral?

sglang_omni/comm/router.py
```diff
@@ -142,6 +143,9 @@ def outbound_payload(self, target: str, payload: Any) -> TransportKind:
             if self.self_is_gpu and target in self.gpu_stage_names:
                 return TransportKind.CUDA_IPC
             return TransportKind.SHM
+        # Cuda_ipc is CUDA-only, so XPU payloads relay over shm.
```
Ratish1: The [RFC #1310](https://github.com/sgl-project/sglang-omni/issues/1310) requires one platform transport decision. This branch makes only `outbound_payload()` select SHM for XPU, while `_physical_outbound()` and `inbound()` can still select CUDA IPC and `outbound_stream()` uses a separate `is_cuda` rule. SHM receive also leaves accelerator-origin tensors on CPU. Could one platform policy drive all these paths and either materialize the tensor on the receiver’s assigned device or reject the unsupported edge?. I think this way it keeps XPU transport contract consistent.

sglang_omni/model_runner/model_worker.py
```diff
@@ -472,6 +473,39 @@ def _apply_model_worker_backend_policy(
...
+                "Intel XPU: could not install the free-memory correction (see the "
+                "logged cause above). Without it the KV pool is sized against "
+                "SGLang's CUDA accounting, which over-reports free memory on XPU. "
+                "Requires torch.xpu.mem_get_info and an importable "
+                "sglang.srt.utils.common."
+            )
+
+    if is_qwen3_omni_arch:
+        if (
+            moe_runner_backend in ("auto", "flashinfer_cutlass", "cutlass")
```
Ratish1: the pinned SGLang v0.5.16 already resolves the unquantized XPU `auto` path to a Triton runner before `forward_xpu()`. Could we leave `auto` to SGLang and only reject or remap explicit CUDA-only backends here?

pyproject_xpu.toml
```diff
@@ -0,0 +1,105 @@
+[build-system]
+# >=64 for PEP 660: the editable install paths need build_editable.
+requires = ["setuptools>=64"]
```
Ratish1: This install uses `--no-build-isolation`, so it uses the environment’s setuptools. The PEP 639 license fields below fail with setuptools 76.1 and succeed with 77.0.1. Could we require `setuptools>=77.0.0` here and validate the version in installer.

sglang_omni/utils/xpu_sglang_compat.py
```diff
@@ -0,0 +1,109 @@
...
+            device, gpu_id, distributed=False, empty_cache=True, cpu_group=None
+        ):
+            dev_type = getattr(device, "type", None) or str(device).split(":", 1)[0]
+            if dev_type == "xpu":
+                if empty_cache:
+                    try:
+                        torch.xpu.empty_cache()
+                    except Exception:  # noqa: BLE001
+                        pass
+                free_gpu_memory, total = torch.xpu.mem_get_info(gpu_id)
```
Ratish1: Pinned SGLang v0.5.16 already owns the XPU memory query, allocator cap, query fallback, distributed minimum, and GB conversion. This replacement copies that logic, globally rebinds imports, and drops the query fallback. Could the Arc `memory_reserved` correction be upstreamed or implemented as a narrow version-gated adapter that delegates the rest to SGLang?

docker/xpu.Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+# purpose: the XPU channel ships no torchaudio 2.12+xpu.
+RUN pip install --no-cache-dir --extra-index-url ${TORCH_XPU_INDEX} \
+        torch==2.12.0+xpu \
+        torchvision==0.27.0+xpu \
+        torchaudio==2.11.0+xpu \
+        torchcodec==0.12.0
+
+# SGLang XPU build: install from source (its own XPU pyproject + sgl-kernel-xpu)
+# rather than the CUDA wheel.
+RUN git clone --branch ${SGLANG_XPU_BRANCH} --single-branch ${SGLANG_XPU_REPO} sglang \
```
Ratish1, line 34: should the platform profile here lock kernel artifacts?. Although SGLang is pinned to v0.5.16, its XPU manifest references `sgl-kernel-xpu` without a VCS revision, so rebuilds can resolve different commits. Could we pin the kernel revision here or consume a digest-pinned XPU base?

sglang_omni/platforms/interface.py
```diff
@@ -14,10 +14,19 @@
 class OmniPlatform(DeviceMixin):
     _omni_platform_qualname: str | None = None
 
+    # Single source for every CommRouter transport decision. True preserves the
+    # pre-existing default; a platform without CUDA IPC overrides it to relay over shm.
+    supports_cuda_ipc: bool = True
```
Ratish1: `_peer_uses_cuda_ipc()` treats an omitted platform override as CUDA-IPC support, but `_build_relay(TransportKind.CUDA_IPC)` always constructs a `cuda:N` relay. Any non-CUDA or dynamically wrapped platform that inherits this default can therefore enter a CUDA-only backend. lets default the capability to `False` and opt in explicitly in `CUDAOmniPlatform`.

sglang_omni/platforms/xpu.py
```diff
@@ -0,0 +1,50 @@
...
+        TP ranks keep every card visible and address theirs by gpu_id."""
+        del env
+        if spec.tp_size <= 1:
+            return {}
+        if spec.gpu_id is None:
+            raise ValueError(f"tp stage {spec.stage_name!r} requires a GPU id")
+        return {
+            # Cleared in the parent: an inherited mask hides peers, and Level Zero
+            # may already be initialized by the time the child could unset it.
+            "ZE_AFFINITY_MASK": None,
```
Ratish1: `ZE_AFFINITY_MASK` restricts and reindexes the Level Zero device handles seen by the process. With an external mask such as `4,5`, SGLang binds TP ranks to logical `xpu:0` and `xpu:1` inside that namespace; deleting the mask widens visibility while retaining those indices, so they can denote different devices. should we preserve and validate a peer-visible group mask, or reject an insufficient mask for TP, instead of unconditionally deleting it.

sglang_omni/utils/device.py
```diff
@@ -0,0 +1,54 @@
...
+
+    index (e.g. a stage's gpu_id) wins over the spec's own; cpu drops any index.
+    Split by hand: torch.device rejects "npu:0", the very case needing a retarget.
+    """
+    from sglang_omni.platforms import current_platform
+
+    dev_type, _, raw_index = str(spec).strip().partition(":")
+    dev_type = dev_type.lower()
+    if dev_type != "cpu" and (
+        dev_type not in _ACCEL_TYPES or not _accel_available(dev_type)
```
Ratish1: `remap_accelerator_spec()` re-runs device detection and cannot tell whether `cuda` came from an internal legacy default or an explicit operator value. Make automatic selection explicit at the call boundary, for example, `device=None` resolve only that case through `current_platform`, and validate any supplied device against CPU/the resolved platform while preserving its placement index. Then `_accel_available()` and its broad exception path can be removed.

docker/xpu.Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+#   docker run -it --device /dev/dri --shm-size 32g sglang-omni:xpu
+
+FROM intel/deep-learning-essentials:2025.3.2-0-devel-ubuntu24.04 AS base
+
+ARG SGLANG_XPU_REPO=https://github.com/sgl-project/sglang.git
+ARG SGLANG_XPU_BRANCH=v0.5.16
+# SGLang's XPU manifest requires sgl-kernel-xpu with no revision, so pinning SGLang
+# alone does not pin the SYCL kernels. Pass a commit SHA for a reproducible image;
+# that repo publishes no tags, so main is the only floating default available.
+ARG SGL_KERNEL_XPU_REF=main
```
Ratish1: SGLang v0.5.16 does not identify a kernel release: its XPU manifest installs `sgl-kernel-xpu` from unversioned Git `main`. Defaulting this argument to `main` therefore lets the same image source resolve different kernel code. let's pin the kernel commit used to validate this PR and keep the build argument only as an explicit override.

sglang_omni/models/higgs_tts/config.py
```diff
@@ -79,7 +79,7 @@ def process_safe_edges(cls) -> frozenset[tuple[str, str]]:
             process="pipeline",
             factory=f"{_PKG}.stages.create_sglang_tts_engine_executor",
             factory_args={
-                "device": "cuda",
+                "device": None,
```
Ratish1: Can we revert the unrelated device changes in ArkASR (`config.py:26`, `stages.py:12`), dots.tts (`config.py:38,46,55`), Fish S2 Pro (`config.py:55`, `stages.py:308`), FunASR (`config.py:36`, `stages.py:96`), Higgs (`config.py:82`, `stages.py:463`), Ming (`stages.py:106`), MOSS (`moss_transcribe_diarize/config.py:42`, `stages.py:74`, `moss_tts/stages.py:502`, `moss_tts_local/config.py:19,169,262,285`, `stages.py:605`), Voxtral (`stages.py:171`), and Whisper (`config.py:34`, `stages.py:12`), since this PR only qualifies Qwen3-ASR, Qwen3-TTS, and Qwen3-Omni on XPU?

sglang_omni/pipeline/stage_workers.py
```diff
@@ -794,7 +794,12 @@ def _prepare_cuda_environment(
...
     log: logging.Logger,
 ) -> None:
-    """Map TP rank processes to one visible CUDA device before torch init."""
+    """Map TP rank processes to their accelerator before torch init.
+
+    Which variables to set is platform policy; this only applies whatever the platform
+    returns, and normalizes gpu_id only when the platform narrowed the process to a
+    single visible device.
+    """
     if os.environ.get("SGLANG_ONE_VISIBLE_DEVICE_PER_PROCESS") == "true":
```
Ratish1: This CUDA placement marker is checked before XPU platform policy. If an XPU TP child inherits it as `true`, every rank is normalized to `gpu_id=0` and returns before XPU validation, so all ranks bind `xpu:0`. Pinned SGLang consumes this marker only under `is_cuda_alike()`. Could this fast path use the same platform guard and add a rank-1 regression test that keeps the XPU device, factory, and communication IDs at 1?

scripts/xpu/install_xpu.sh
```diff
@@ -0,0 +1,178 @@
...
+  if [[ -f "${BACKUP}" ]]; then
+    cp -f "${BACKUP}" "${PYPROJECT}"
+    rm -f "${BACKUP}"
+    echo "restored original pyproject.toml"
+  fi
+  rm -rf "${REPO_ROOT}/sglang_omni.egg-info" "${REPO_ROOT}/build" 2>/dev/null || true
+}
+trap restore EXIT INT TERM
+
+cp -f "${PYPROJECT}" "${BACKUP}"
```
Ratish1, line 155: If a previous run is killed after the swap, this fixed backup is the only copy of the original `pyproject.toml`. Rerunning now overwrites it with the already-swapped XPU manifest, so `restore()` cannot recover the user file. Could the script refuse to proceed when a stale backup exists and print recovery steps, or use a unique temporary backup instead of `cp -f` here?

sglang_omni/comm/router.py
```diff
@@ -92,13 +92,15 @@ def outbound_stream(self, target: str, data: torch.Tensor) -> TransportKind:
             )
         if target in self.remote_stage_names:
             return TransportKind.MOONCAKE
-        if not data.is_cuda:
+        # is_cuda would read False for an accelerator tensor off CUDA, sending every
+        # XPU/NPU stream chunk over host shm.
+        if data.device.type != current_platform.device_type:
```
Ratish1, line 169: On XPU, Qwen3-TTS emits device-resident codec chunks, but this SHM path copies them to CPU and `read_stream_chunk()` cannot restore their source residency because raw `DataRef` does not record it. The default same-process topology is unaffected, but the declared process-safe `--isolate-stage vocoder` path then passes CPU codes to an XPU decoder. Could stream reads restore accelerator-origin tensors on the receiver device like payload reads, or reject this XPU process split until that contract exists?

sglang_omni/pipeline/stage_workers.py
```diff
@@ -794,7 +794,12 @@ def _prepare_cuda_environment(
     spec: StageLaunchConfig,
     log: logging.Logger,
 ) -> None:
-    """Map TP rank processes to one visible CUDA device before torch init."""
+    """Map TP rank processes to their accelerator before torch init.
```
Ratish1, line 797: This function now prepares both CUDA and XPU workers, but its name still makes the shared launcher look CUDA-owned. Could we rename it to `_prepare_accelerator_environment()` and update its callers and tests?

sglang_omni/utils/device.py
```diff
@@ -0,0 +1,53 @@
...
+Automatic selection is explicit: ``device=None`` is the *only* input that resolves
+through ``current_platform``. A supplied device is honored and validated, never
+re-detected, so an operator's choice can never be silently retargeted.
+"""
+
+from __future__ import annotations
+
+import logging
+
+logger = logging.getLogger(__name__)
```
Ratish1: This module creates `logger`, but never uses it. Could we remove the `logging` import and logger assignment?

tests/unit_test/qwen3_omni/test_thinker_platform_gates.py
```diff
@@ -0,0 +1,39 @@
...
+
+from __future__ import annotations
+
+import inspect
+
+from sglang_omni.models.qwen3_omni.components import talker, thinker_model
+
+
+def test_thinker_gates_read_the_resolved_platform_not_torch_cuda() -> None:
+    source = inspect.getsource(thinker_model)
```
Ratish1, line 21: These tests inspect exact source strings and counts rather than the platform-policy behavior, so harmless refactors can fail while a semantic regression can preserve the spelling. Could the CUDA gate be a small helper tested with CUDA/XPU platform doubles and used by the thinker and both talker constructors?

tests/unit_test/test_stage_device_contract.py
```diff
@@ -0,0 +1,65 @@
...
+    assert parameter is not None, f"{factory_path} takes no device parameter"
+    return str(parameter.annotation)
+
+
+@pytest.mark.parametrize("model", _MODELS)
+def test_none_device_stages_target_a_none_aware_factory(model: str) -> None:
+    for label, stage in _stages_with_device(model):
+        if stage.factory_args["device"] is not None:
+            continue
+        annotation = _device_annotation(stage.factory)
```
Ratish1: Expanding the topology scan still accepts any `str | None` annotation without proving that the factory resolves `None`; a factory can be re-annotated and continue forwarding `None` unchanged. Could this test exercise an explicit resolver/factory contract instead of inferring behavior from the annotation text?

sglang_omni/models/higgs_tts/config.py
siju-samuel: I did try it — but the revert doesn't hold, because EngineFactory.build() is shared by every model and the device resolution there is load-bearing for two of the three qualified ones.

sglang_omni/utils/device.py doesn't exist on main at all; it's new in this PR. Upstream's build() hardcodes:


if gpu_id is not None:
    device = f"cuda:{gpu_id}"
gpu_id = int(device.split(":")[-1]) if ":" in device else 0

sglang_omni/scheduling/engine_factory.py
```diff
@@ -47,18 +47,23 @@ def build(
...
         from sglang_omni.scheduling import bootstrap as scheduling_bootstrap
         from sglang_omni.scheduling import sglang_backend
+        from sglang_omni.utils.device import resolve_device_spec
 
         checkpoint_dir = self.resolve_checkpoint(model_path)
-        if gpu_id is not None:
+        if device is None:
+            device = resolve_device_spec(None, gpu_id)
+        elif gpu_id is not None:
             device = f"cuda:{gpu_id}"
```
Ratish1: This rewrites an explicit `device="xpu"` (or `cpu`) to `cuda:N` whenever placement supplies `gpu_id`, so the stage can diverge from the resolved platform. Now that unsupported model defaults are CUDA-explicit again, could the shared builder keep `device = resolve_device_spec(device, gpu_id)` to preserve the selected platform and placement index?

sglang_omni/scheduling/engine_factory.py
```diff
@@ -74,6 +79,7 @@ def build(
             **self.generation_defaults(dtype=dtype),
         )
         self.adjust_overrides(overrides)
+        overrides.setdefault("device", torch.device(device).type)
```
Ratish1: `server_args_overrides` can already contain `device`, so `setdefault()` can leave SGLang on CUDA after this stage resolved XPU, or vice versa. Could we validate any existing value against `torch.device(device).type`, reject a mismatch, and then pass the resolved type so placement and SGLang keep one device authority?

sglang_omni/comm/stage_io.py
```diff
@@ -457,8 +461,11 @@ async def write_stream_chunk(
 async def read_stream_chunk(
     relay: Relay,
     data_ref: DataRef,
+    local_device: str | None = None,
 ) -> tuple[torch.Tensor, dict[str, Any] | None]:
     data = await read_tensor(relay, data_ref)
+    if data_ref.device is not None:
```
Ratish1, line 467: `write_tensor()` records source residency, but `_with_stream_metadata()` reconstructs this `DataRef` without `data_ref.device`. Qwen3-TTS attaches metadata to every codec chunk, so an isolated XPU vocoder still receives the SHM CPU tensor. Could we preserve the outer device, restore nested metadata refs from their recorded devices, and cover a metadata-bearing write/read round trip?

tests/unit_test/test_stage_device_contract.py
```diff
@@ -0,0 +1,97 @@
...
+    if _RESOLVER in names:
+        return True
+    if depth <= 0:
+        return False
+    module = sys.modules.get(getattr(func, "__module__", ""))
+    for name in names:
+        helper = getattr(module, name, None)
+        if callable(helper) and _resolves_device(helper, depth - 1):
+            return True
+    if "build" in names and _resolves_device(
```
Ratish1: This is stronger than checking annotations, but `_resolves_device()` is still a bytecode-name heuristic: any factory containing a `build` name is credited with `SGLangGenerationEngineBuilder.build`, even if it calls another builder. Could we test the qualified Qwen factories with lightweight patched builders/resolvers instead of recursively searching `co_names`?

scripts/xpu/install_xpu.sh
```diff
@@ -0,0 +1,201 @@
...
+    echo "ERROR: leftover backup found: ${BACKUP}"
+    echo
+    echo "A previous run was interrupted after swapping pyproject.toml, so that"
+    echo "backup is the only copy of your original manifest. Restore it first:"
+    echo
+    echo "  cp ${BACKUP} ${PYPROJECT} && rm ${BACKUP}"
+    echo
+    echo "In a git checkout you can equivalently run:"
+    echo
+    echo "  git checkout -- pyproject.toml && rm ${BACKUP}"
```
Ratish1: The `git checkout --` alternative is not equivalent when the original manifest had uncommitted changes: it discards them and then deletes the backup containing them. The `cp` command above restores the exact file; could we remove this alternative?

sglang_omni/comm/router.py
```diff
@@ -159,15 +158,15 @@ def outbound_stream(self, target: str, data: torch.Tensor) -> TransportKind:
             )
         if target in self.remote_stage_names:
             return TransportKind.MOONCAKE
-        if not data.is_cuda:
+        if data.device.type != current_platform.device_type:
```
Ratish1, line 169: XPU/NPU select SHM, but this branch calls `_cuda_ipc_peer_available()` before reading that platform decision. On a cross-device XPU edge it queries `torch.cuda.device_count()` and logs a CUDA fallback even though CUDA IPC was never eligible. Could we resolve `get_intra_node_transport()` first and only run the peer probe when it returns `CUDA_IPC`? This preserves CUDA peer fallback and keeps non-CUDA routing behind platform policy.

tests/unit_test/test_stage_device_contract.py
```diff
@@ -0,0 +1,141 @@
...
+    model.eval = lambda: model
+    monkeypatch.setattr(
+        code2wav_scheduler, "load_code2wav_model", lambda *a, **k: model
+    )
+
+    scheduler = code2wav_scheduler.create_code2wav_scheduler("unused", device=None)
+
+    assert scheduler._device.type == platforms.current_platform.device_type
+    # Placement was not requested, so the backend's current card is bound.
+    assert scheduler._device.index is not None
```
Ratish1: This test also runs on CPU, where `torch.device("cpu").index` is correctly `None`, so the unconditional assertion fails before exercising an accelerator contract. Could we condition this assertion on a non-CPU platform or pin a fake accelerator platform in the test?

pyproject_xpu.toml
```diff
@@ -0,0 +1,107 @@
...
+dynamic = ["version"]
+description = "Multi-stage pipeline framework for omni models (Intel XPU build)"
+readme = "README.md"
+license = "Apache-2.0"
+license-files = ["LICENSE"]
+requires-python = ">=3.10,<3.13"
+dependencies = [
+    # --- Core framework (device-agnostic) ----------------------------------
+    "pyzmq>=25.0.0",
+    "msgpack>=1.0.0",
```
Ratish1, line 24: `sglang_omni.comm` imports `msgspec` directly, and the main manifest declares it, but the XPU manifest only declares `msgpack`. Could we add `msgspec` here instead of relying on the separately installed SGLang package to provide it transitively?

sglang_omni/models/qwen3_omni/components/thinker_model.py
```diff
@@ -229,6 +230,7 @@ def __init__(
         self.compatible_with_fused_qk_norm_rope = (
             not isinstance(self.rotary_emb, MRotaryEmbedding)
         ) and self.head_dim in (64, 128, 256)
+        _yarn_factor, _, _, _ = compute_yarn_parameters(config)
```
Ratish1: In pinned SGLang this value is passed into `can_use_fused_qk_norm_rope()` to qualify YaRN support, but here all four results are discarded and the fused branch recomputes them when needed. If Omni is not adopting that capability check here, could we remove this dead initialization-time call?

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -530,15 +533,18 @@ def create_code2wav_scheduler(
...
-        concrete_device = torch.device(device)
+    concrete_device = torch.device(resolve_device_spec(device, gpu_id))
     if concrete_device.type != "cpu" and concrete_device.index is None:
         concrete_device = current_platform.get_device(
             torch.get_device_module().current_device()
         )
     device = str(concrete_device)
+    if enable_cuda_graph and concrete_device.type != "cuda":
+        logger.info(
+            "Code2Wav CUDA graph disabled on %s (CUDA-only runner); running eager",
```
Ratish1: can we use f formatting here to be consistent with the codebase

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -530,15 +533,18 @@ def create_code2wav_scheduler(
...
-        concrete_device = current_platform.get_device(gpu_id)
-    else:
-        concrete_device = torch.device(device)
+    concrete_device = torch.device(resolve_device_spec(device, gpu_id))
     if concrete_device.type != "cpu" and concrete_device.index is None:
         concrete_device = current_platform.get_device(
             torch.get_device_module().current_device()
         )
     device = str(concrete_device)
+    if enable_cuda_graph and concrete_device.type != "cuda":
```
Makcum888e: you can disable graph for code2wav by override `enable_code2wav_graph` in XPUOmniPlatform

sglang_omni/scheduling/engine_factory.py
```diff
@@ -47,18 +47,24 @@ def build(
         self,
         model_path: str,
         *,
-        device: str = "cuda:0",
+        device: str | None = None,
```
Ratish1, line 50: `build()` is shared, and Zonos2 still calls it without `device=`. This new `None` default therefore resolves that unqualified CUDA-only model onto the ambient XPU/NPU platform. Could Zonos2 retain an explicit `"cuda:0"` at its factory/build call, or otherwise preserve the old default for unqualified callers, while the three qualified Qwen factories pass `None`?

sglang_omni/models/qwen3_asr/stages.py
```diff
@@ -9,7 +9,7 @@
 def create_sglang_qwen3_asr_executor(
     model_path: str,
     *,
-    device: str = "cuda:0",
+    device: str | None = None,
```
Ratish1, line 12: Runtime placement injects `gpu_id` only when the factory signature declares it. This factory accepts `device=None` but not `gpu_id`, so placement on card 1 calls the shared builder with both values unset; it resolves a bare accelerator and then supplies `gpu_id=0` to SGLang. Could this factory accept `gpu_id: int | None = None` and forward it to `.build()`?

Ratish1 (approved):
LGTM

PR #1005, opened 2026-07-09
[1/2] [Perf] Improve Ming-Omni-TTS serving: streaming

YzXiao101:
## Motivation

[PR #858](https://github.com/sgl-project/sglang-omni/pull/858) added the standalone Ming-Omni-TTS pipeline.

A single-H200 profile showed throughput plateauing near concurrency 8 while compute and memory bandwidth remained underused. Non-streaming AudioVAE decode was the main high-concurrency bottleneck, dominated by many 1-5 us kernels and host launch and synchronization overhead.

This PR reduces that overhead by decoding each non-streaming request once, while adding request-visible streaming from committed acoustic latents.

AudioVAE CUDA graph work remains separate in [#1250](https://github.com/sgl-project/sglang-omni/pull/1250).

### Scope

This PR adds:

- PCM streaming from committed acoustic latent patches.
- One full-sequence AudioVAE decode per non-streaming request.
- Configurable M-patch initial and K-patch steady streaming AudioVAE cadence.

### High-Level Design

```text
text / reference audio
        |
        +--> reference encode ---- separate content cache
        |              |
        v              v
prompt tokens + projected conditioning
        |
        v
AR backbone <---------------- acoustic feedback row
        |                               ^
        v                               |
acoustic tail ---- committed latent patch
        |
        +-- stream=true  --> cadence --> stateful AudioVAE --> PCM chunks
        |
        `-- stream=false --> full latent sequence --> one AudioVAE decode
```

### Decision & Design

**Streaming audio.** The first non-terminal AudioVAE call consumes M latent patches but emits no PCM because it must retain right-hand context for linear upsampling. A later call supplies that context and emits the initial group; subsequent calls consume K-patch steady groups. The provided configuration uses `M=2,K=4`. The concurrency-8 sweep showed that M2/K4 removes the first-seam underrun seen with M1/K4 while leaving RTF and throughput effectively unchanged. K=1/2 invokes eager AudioVAE more often and is substantially slower; K=6/8 delays first audio and produces fewer, larger PCM payloads. Terminal residuals bypass the cadence and flush immediately.

**AudioVAE execution.** Non-streaming requests decode the complete latent sequence once to reduce the profiled small-kernel launch overhead. Streaming requests retain Qwen KV, upsample lookahead, and ISTFT overlap state per request, preserving waveform continuity when AR generation pauses and later resumes. Only committed acoustic-tail output from the TP entry rank is published, preventing follower duplication and uncommitted audio.

## Related PRs

- [#858](https://github.com/sgl-project/sglang-omni/pull/858) adds the standalone Ming-Omni-TTS serving pipeline used by this PR.
- [#1250](https://github.com/sgl-project/sglang-omni/pull/1250) tracks separate AudioVAE CUDA graph work and is not part of this PR.

## Validation Setup

| Item | Value |
|---|---|
| Benchmark commit | `1ed62262` |
| Container | `hongccc/sglang-omni@sha256:e28596c20e7eb187ee0f58625c0fd6c26d077da355d521c29c7003555670122b` |
| Model | `inclusionAI/Ming-omni-tts-16.8B-A3B` @ `2efcd28d04f83b1bd471af639b4210c9ecf2cd33` |
| Dataset | Full `zhaochenyang20/seed-tts-eval-arrow` @ `27f4c1adee83b5b29b7c4b375f6b976324bda308`, EN (1,088) and ZH (2,020), text-only and reference-audio slices |
| ASR scorer | `Qwen/Qwen3-ASR-1.7B` @ `7278e1e70fe206f11671096ffdd38061171dd6e5` |
| Deployment | 1x H200 141 GB, TP1, concurrency 8, eight warmup requests |
| Graphs | AR and acoustic-tail CUDA graphs enabled; AudioVAE eager |
| Streaming cadence | Two initial patches, then four-patch steady groups (`M=2,K=4`) |

#### Streaming

| Slice | Lang | Samples | Failed | Corpus WER | RTF Mean | Latency Mean (s) | First Audio Mean (s) | Throughput (qps) | Audio s/s |
|---|---|---:|---:|---:|---:|---:|---:|---:|---:|
| text-only | EN | 1,088 | 0 | 0.90% | 0.2686 | 1.236 | 0.6208 | 6.462 | 30.617 |
| text-only | ZH | 2,020 | 0 | 0.71% | 0.2592 | 1.278 | 0.6131 | 6.253 | 31.417 |
| reference | EN | 1,088 | 0 | 1.09% | 0.3107 | 1.358 | 0.7471 | 5.883 | 26.622 |
| reference | ZH | 2,020 | 0 | 0.75% | 0.2670 | 1.507 | 0.7326 | 5.303 | 30.377 |

#### Non-Streaming

| Slice | Lang | Samples | Failed | Corpus WER | RTF Mean | Latency Mean (s) | Throughput (qps) | Audio s/s |
| text-only | EN | 1,088 | 0 | 0.91% | 0.2035 | 0.959 | 8.328 | 39.369 |
| text-only | ZH | 2,020 | 0 | 0.69% | 0.1965 | 0.983 | 8.134 | 40.790 |
| reference | EN | 1,088 | 0 | 1.07% | 0.2308 | 1.031 | 7.742 | 35.021 |
| reference | ZH | 2,020 | 0 | 0.71% | 0.1966 | 1.125 | 7.103 | 40.737 |

#### Streaming Playback Continuity

`N/A` means a legal single-payload response with no seam to score.

| Slice | Lang | Scored / N/A | First-seam p95 / p99 (s) | C50 / C100 / C200 | Positive later underruns / later seams |
|---|---|---:|---:|---:|---:|
| text-only | EN | 1,088 / 0 | 0 / 0 | 100% / 100% / 100% | 0 / 1,713 |
| text-only | ZH | 2,020 / 0 | 0 / 0 | 100% / 100% / 100% | 0 / 3,645 |
| reference | EN | 1,071 / 17 | 0 / 0 | 100% / 100% / 100% | 0 / 1,550 |
| reference | ZH | 2,020 / 0 | 0 / 0 | 100% / 100% / 100% | 0 / 4,733 |

### Result

- All 12,432 generation requests completed without serving failures.
- Worst-slice corpus WER was 1.09%. Streaming produced first audio in 0.61-0.75 seconds; non-streaming retained higher complete-response throughput because it decodes each full latent sequence once.
- Every scoreable streaming request reached C50/C100/C200, and all 11,641 later seams had zero measured underrun. The 17 N/A reference-EN requests returned one PCM payload and therefore had no seam to score.

## Checklist

- [x] Update user-facing documentation.
- [x] Provide fresh full-corpus accuracy results.
- [x] Validate streaming and non-streaming output.
- [x] Validate the radix-disabled configuration and fail-fast rejection of radix enablement.

## Appendix: Streaming Cadence Sweep Validation

For a non-terminal stream, the first AudioVAE call buffers M patches and emits no PCM. The next call supplies lookahead and emits PCM for that initial group. Later calls consume K patches; terminal residuals flush immediately.

Setup: commit `1ed62262`, 1x H200, TP1, concurrency 8, `warmup=8`, AudioVAE eager. Each cell measured 100 requests in each of four SeedTTS slices. The tables retain `M={1,2,4}` and `K={1,2,4,6,8}`. All reported cells completed 400/400 requests.

Means are averaged across the four slices; p95 and WER use the worst slice.

| M | K | RTF mean | Latency p95 worst (s) | QPS mean | Audio s/s mean | First audio mean / p95 worst (s) | Chunks/request mean | Corpus WER worst |
| 1 | 1 | 0.5475 | 3.493 | 3.1917 | 14.838 | 1.162 / 1.869 | 13.70 | 1.05% |
| 1 | 2 | 0.3591 | 2.256 | 4.8230 | 22.601 | 0.579 / 0.909 | 7.15 | 1.16% |
| 1 | 4 | 0.2823 | 1.826 | 6.1045 | 28.610 | 0.628 / 0.936 | 3.84 | 0.64% |
| 1 | 6 | 0.2603 | 1.797 | 6.5350 | 30.844 | 0.701 / 0.985 | 2.71 | 1.16% |
| 1 | 8 | 0.2560 | 1.722 | 6.6440 | 31.217 | 0.802 / 1.083 | 2.19 | 0.55% |
| 2 | 1 | 0.5330 | 3.665 | 3.2640 | 15.241 | 1.138 / 1.840 | 12.79 | 0.55% |
| 2 | 2 | 0.3546 | 2.233 | 4.9278 | 22.930 | 0.639 / 0.937 | 6.60 | 1.16% |
| **2** | **4** | **0.2811** | **1.794** | **6.1215** | **28.639** | **0.673 / 1.024** | **3.60** | **1.05%** |
| 2 | 6 | 0.2613 | 1.716 | 6.5813 | 30.653 | 0.755 / 1.029 | 2.55 | 0.64% |
| 2 | 8 | 0.2514 | 1.727 | 6.7615 | 31.734 | 0.829 / 1.167 | 2.04 | 1.16% |
| 4 | 1 | 0.4688 | 3.361 | 3.6845 | 17.128 | 1.093 / 1.855 | 10.75 | 0.55% |
| 4 | 2 | 0.3321 | 2.217 | 5.1888 | 24.371 | 0.770 / 1.068 | 5.68 | 1.16% |
| 4 | 4 | 0.2787 | 1.848 | 6.1705 | 29.025 | 0.836 / 1.164 | 3.09 | 0.55% |
| 4 | 6 | 0.2568 | 1.745 | 6.6580 | 31.373 | 0.873 / 1.247 | 2.25 | 1.05% |
| 4 | 8 | 0.2512 | 1.740 | 6.7660 | 31.620 | 0.951 / 1.321 | 1.77 | 0.55% |

N/A means a legal single-payload response with no seam to score. Counts are summed; first-seam percentiles and C50/C100/C200 use the worst slice.

| M | K | Scored / N/A | First-seam underrun p95 / p99 worst (s) | C50 / C100 / C200 worst | Positive later underruns / later seams |
| 1 | 1 | 400 / 0 | 0.0302 / 0.1012 | 96.00% / 98.00% / 100.00% | 0 / 4,680 |
| 1 | 2 | 400 / 0 | 0.1076 / 0.1375 | 78.00% / 92.00% / 100.00% | 0 / 2,059 |
| 1 | 4 | 400 / 0 | 0.1605 / 0.2200 | 32.00% / 66.00% / 98.00% | 0 / 734 |
| 1 | 6 | 392 / 8 | 0.2160 / 0.3984 | 29.00% / 61.00% / 91.30% | 0 / 293 |
| 1 | 8 | 370 / 30 | 0.3569 / 0.4280 | 27.00% / 49.00% / 79.01% | 0 / 104 |
| 2 | 1 | 400 / 0 | 0 / 0 | 100.00% / 100.00% / 100.00% | 0 / 4,316 |
| 2 | 2 | 400 / 0 | 0 / 0 | 100.00% / 100.00% / 100.00% | 0 / 1,838 |
| **2** | **4** | **396 / 4** | **0 / 0** | **100.00% / 100.00% / 100.00%** | **0 / 642** |
| 2 | 6 | 382 / 18 | 0 / 0 | 100.00% / 100.00% / 100.00% | 0 / 236 |
| 2 | 8 | 345 / 55 | 0.0422 / 0.0803 | 95.71% / 98.99% / 100.00% | 0 / 70 |
| 4 | 1 | 399 / 1 | 0 / 0 | 100.00% / 100.00% / 100.00% | 0 / 3,502 |
| 4 | 2 | 396 / 4 | 0 / 0 | 100.00% / 100.00% / 100.00% | 0 / 1,477 |
| 4 | 4 | 383 / 17 | 0 / 0 | 100.00% / 100.00% / 100.00% | 0 / 453 |
| 4 | 6 | 341 / 59 | 0 / 0 | 100.00% / 100.00% / 100.00% | 0 / 158 |
| 4 | 8 | 274 / 126 | 0 / 0 | 100.00% / 100.00% / 100.00% | 0 / 34 |

- All reported cells passed serving and worst-slice corpus WER stayed at or below 1.16%; WER does not select the cadence.
- The issue is the first seam: M1/K4 reaches only 32%/66%/98% C50/C100/C200, while M2/K4 reaches 100%/100%/100%. All reported later seams have zero underrun.
- At K=4, moving from M=1 to M=2 leaves RTF and QPS effectively unchanged (0.2823 to 0.2811 RTF; 6.1045 to 6.1215 QPS) and adds 45 ms mean first-audio latency. M=4 raises first audio further to 0.836 s for little additional throughput.
- K=1/2 invokes AudioVAE more often and is substantially slower. K=6/8 improves throughput, but delays first audio and produces fewer, larger PCM payloads. K=4 is the balance point between throughput, first-audio latency, and delivery cadence.

The selected default is therefore `M=2,K=4`.


YzXiao101:
## Update: Streaming Audio Output

**Update:**

- Added request-local streaming AudioVAE decode state and incremental PCM delivery while preserving the existing non-streaming WAV path.
- Added terminal codec flush handling so the final buffered waveform is emitted before stream completion.

**Benchmark:**

Commit: `aab51edb`. Setup: 2x H200, TP2, concurrency 8, AR CUDA graph and tail CUDA graph enabled, 100 SeedTTS samples per slice, `warmup=8`.

| slice | lang | first audio mean_s (baseline -> stream) | first audio p95_s (baseline -> stream) | streaming ITL p95_s | RTF mean (baseline -> stream) | QPS (baseline -> stream) |
| text_only | en | 2.189 -> 1.349 (-38%) | 2.690 -> 1.825 (-32%) | 0.182 | 0.5064 -> 0.5825 | 3.545 -> 3.146 |
| text_only | zh | 2.182 -> 0.710 (-67%) | 2.524 -> 1.085 (-57%) | 0.235 | 0.4905 -> 0.5349 | 3.540 -> 3.351 |
| reference | en | 2.163 -> 1.252 (-42%) | 2.767 -> 1.760 (-36%) | 0.180 | 0.5632 -> 0.5898 | 3.606 -> 3.387 |
| reference | zh | 2.863 -> 1.137 (-60%) | 3.351 -> 1.945 (-42%) | 0.217 | 0.5056 -> 0.5392 | 2.715 -> 2.544 |

For the non-streaming baseline, first-audio wait is the full-response latency.

**Accuracy:**

| slice | lang | baseline WER | streaming WER | streaming n_above_50 |
| text_only | en | 0.005460 | 0.009099 | 0 |
| text_only | zh | 0.003690 | 0.004744 | 0 |
| reference | en | 0.005460 | 0.006369 | 0 |
| reference | zh | 0.003163 | 0.003163 | 0 |

Result:

- The streaming run completed all 400 requests, emitted multiple audio chunks per request, and passed all four WER gates with zero `n_above_50` samples.
- Streaming reduced mean first-audio wait by 38-67% and p95 by 32-57%; ITL p95 was 180–235 ms, below the 290 ms first PCM chunk and approximately 320 ms steady-state chunks.
- Incremental AudioVAE invocation increased mean RTF by 5-15% and reduced QPS by 5-11%.

YzXiao101:
## Update: Prompt Prefix Cache

Update:

- Added deterministic prompt cache identities for text-only and reference-conditioned requests, including speaker and prompt-latent metadata in the reference fingerprint.
- Limited radix insertion to fresh prompt prefill so generated acoustic feedback is never shared across requests.
- Preserved retraction correctness by replaying request-local feedback state while allowing the original prompt KV to be reused.

Benchmark:

Commit: `417096ee`. Setup: 2x H200, TP2, concurrency 8, AR CUDA graph and tail CUDA graph enabled, 100 SeedTTS samples per slice, `warmup=8`.

| slice | lang | RTF mean (baseline -> cache) | latency mean_s (baseline -> cache) | QPS (baseline -> cache) |
| text_only | en | 0.5240 -> 0.5562 | 2.281 -> 2.361 | 3.400 -> 3.288 |
| text_only | zh | 0.4741 -> 0.5121 | 2.103 -> 2.226 | 3.678 -> 3.468 |
| reference | en | 0.5469 -> 0.7047 | 2.155 -> 2.746 | 3.617 -> 2.837 |
| reference | zh | 0.4921 -> 0.5705 | 2.790 -> 3.248 | 2.776 -> 2.404 |

Accuracy:

| slice | lang | baseline WER | cache WER | cache n_above_50 |
| text_only | en | 0.005460 | 0.005460 | 0 |
| text_only | zh | 0.003690 | 0.004744 | 0 |
| reference | en | 0.005460 | 0.006369 | 0 |
| reference | zh | 0.002636 | 0.004217 | 0 |

Result:

- The cache-enabled run completed all 400 requests and passed all four WER gates with zero `n_above_50` samples.
- SGLang reused 6,214 of 23,806 prefill tokens (26.1%); the cache-disabled baseline reused none.
- This short, diverse SeedTTS workload did not convert the partial prompt hits into an end-to-end speedup: mean RTF increased by 6-29% and QPS decreased by 3-22%. Workloads with longer or more frequently repeated prompts are required to evaluate the intended latency and throughput benefit.

YzXiao101:
## Update: Single-H200 TP1 Validation

Update:

- Validated the standard non-streaming pipeline with reference encoding, AR generation, and AudioVAE decode colocated on one H200 using TP1.

Benchmark:

Commit: `417096ee`. Setup: 1x H200, TP1, concurrency 8, AR CUDA graph and tail CUDA graph enabled, radix cache disabled, 100 SeedTTS samples per slice, `warmup=8`.

| slice | lang | failed | rtf_mean | latency_mean_s | throughput_qps | WER | n_above_50 |
| text_only | en | 0 | 0.5092 | 2.197 | 3.537 | 0.005460 | 0 |
| text_only | zh | 0 | 0.5232 | 2.307 | 3.341 | 0.004217 | 0 |
| reference | en | 0 | 0.7507 | 2.837 | 2.749 | 0.004550 | 0 |
| reference | zh | 0 | 0.6242 | 3.569 | 2.193 | 0.011070 | 1 |

Result:

- All 400 requests completed successfully; startup, weight loading, CUDA graph capture, and all four serving slices completed on one H200.

YzXiao101:
## Update: Single-H200 System Profile

Update:

- Added opt-in NVTX ranges for the AR backbone, acoustic tail, reference encoder, and AudioVAE decode without changing the default serving path.
- Built a repeatable single-H200 profile workflow that separates clean performance runs, request-stage events, and Nsight Systems collection.
- Added preflight checks and local analysis for workload stability, process attribution, CUDA timelines, and H200 hardware counters.

Benchmark:

Commit: `bcd7bdf0`. Setup: 1x H200, TP1, AR CUDA graph and tail CUDA graph enabled, streaming and caches disabled. Each scenario used disjoint warmup requests and two clean measured repeats.

| slice | concurrency | requests/repeat | rtf_mean | latency_mean_s | throughput_qps | audio_s/s |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| text_only | 1 | 16 | 0.1493 | 0.679 | 1.473 | 6.717 |
| text_only | 8 | 32 | 0.4788 | 2.037 | 3.570 | 15.489 |
| text_only | 16 | 64 | 0.9185 | 3.996 | 3.591 | 16.121 |
| reference | 1 | 16 | 0.1658 | 0.891 | 1.123 | 6.118 |
| reference | 8 | 32 | 0.5365 | 2.750 | 2.663 | 14.339 |
| reference | 16 | 64 | 1.0054 | 4.955 | 2.907 | 15.209 |

Result:

- All six scenarios passed the workload and aggregate-throughput stability checks with less than 5% drift between clean repeats.
- Throughput reached its current plateau near concurrency 8. Raising concurrency to 16 increased mean latency by 80-96% while improving audio throughput by only 4-6%.
- H200 Tensor, SM, HBM, PCIe, power, and memory evidence showed that hardware capacity was not saturated. The dominant pattern was a large number of 1-5 us kernels and host-side launch/synchronization overhead.
- Stage events and process-level Nsight attribution identified non-streaming AudioVAE decode as the primary high-concurrency bottleneck and serialized reference encoding as a secondary limit.
- Nsight reduced measured audio throughput by approximately 14-25%, so clean repeats remain the performance source of truth and profiler runs are used only for causal attribution.

YzXiao101:
## Update: Non-Streaming AudioVAE Full-Sequence Decode

Update:

- Replaced per-latent cached reconstruction in the non-streaming path with one full-sequence AudioVAE decode.
- Kept the incremental stateful decoder unchanged for streaming responses.

Benchmark:

Baseline commit: `09b0d11a`; optimized commit: `65025859`. Setup: 1x H200, TP1, concurrency 8, AR CUDA graph and tail CUDA graph enabled, streaming and radix cache disabled, 100 SeedTTS samples per slice, `warmup=8`.

| slice | lang | RTF mean (baseline -> optimized) | latency mean_s (baseline -> optimized) | QPS (baseline -> optimized) | audio_s/s (baseline -> optimized) | WER (baseline -> optimized) |
| text_only | en | 0.4888 -> 0.2017 | 2.100 -> 0.900 | 3.689 -> 8.648 | 16.596 -> 38.743 | 0.005460 -> 0.004550 |
| text_only | zh | 0.4632 -> 0.2015 | 2.050 -> 0.899 | 3.759 -> 8.700 | 16.877 -> 38.893 | 0.003690 -> 0.004744 |
| reference | en | 0.5333 -> 0.2393 | 2.093 -> 0.989 | 3.724 -> 7.874 | 15.491 -> 32.731 | 0.004550 -> 0.005460 |
| reference | zh | 0.4806 -> 0.1891 | 2.704 -> 1.082 | 2.876 -> 7.244 | 16.463 -> 41.470 | 0.003163 -> 0.005271 |

Result:

- Both runs completed all 400 requests with zero `n_above_50` samples.
- The controlled A/B observed 53-60% lower mean latency and 2.1-2.5x higher request and audio throughput.

YzXiao101:
## Update: AudioVAE CUDA Graph

Update:

- Added configurable CUDA graph replay for the Qwen2 portion of non-streaming and streaming AudioVAE decode.
- Kept cache-boundary transitions eager and recorded capture and replay counts for runtime review.

Benchmark:

Commit: `e73bc5da`. Setup: 1x H200, TP1, concurrency 8, AR CUDA graph and tail CUDA graph enabled, radix cache disabled, 100 SeedTTS samples per slice, `warmup=8`. The A/B changed only the AudioVAE graph setting.

Non-streaming:

| slice | lang | RTF mean (eager -> graph) | latency mean_s (eager -> graph) | QPS (eager -> graph) | audio_s/s (eager -> graph) | WER (eager -> graph) | n_above_50 (eager -> graph) |
| text_only | en | 0.2085 -> 0.1934 | 0.929 -> 0.868 | 8.382 -> 8.960 | 37.390 -> 40.196 | 0.004550 -> 0.005460 | 0 -> 0 |
| text_only | zh | 0.2051 -> 0.1933 | 0.905 -> 0.878 | 8.587 -> 8.900 | 37.864 -> 40.641 | 0.004744 -> 0.005271 | 0 -> 0 |
| reference | en | 0.2477 -> 0.2367 | 1.027 -> 0.982 | 7.563 -> 7.962 | 31.729 -> 33.377 | 0.005460 -> 0.005460 | 0 -> 0 |
| reference | zh | 0.1955 -> 0.1849 | 1.115 -> 1.057 | 7.034 -> 7.318 | 40.176 -> 41.775 | 0.010543 -> 0.002636 | 1 -> 0 |

Streaming:

| slice | lang | RTF mean (eager -> graph) | latency mean_s (eager -> graph) | first audio mean_s (eager -> graph) | QPS (eager -> graph) | ITL p95_s (eager -> graph) | WER (eager -> graph) |
| text_only | en | 0.5960 -> 0.3498 | 2.563 -> 1.467 | 1.535 -> 0.504 | 3.053 -> 5.321 | 0.163 -> 0.293 | 0.005460 -> 0.004550 |
| text_only | zh | 0.5581 -> 0.3254 | 2.463 -> 1.435 | 1.128 -> 0.481 | 3.166 -> 5.443 | 0.207 -> 0.273 | 0.005271 -> 0.003690 |
| reference | en | 0.6264 -> 0.4147 | 2.455 -> 1.585 | 1.480 -> 0.636 | 3.187 -> 4.970 | 0.174 -> 0.272 | 0.005460 -> 0.005460 |
| reference | zh | 0.5673 -> 0.2969 | 3.240 -> 1.672 | 1.414 -> 0.615 | 2.424 -> 4.675 | 0.196 -> 0.255 | 0.006326 -> 0.003163 |

Result:

- AudioVAE graph replay executed 424 times in non-streaming decode and 1,084 times in streaming decode. Both graph-enabled runs completed all 800 requests with zero `n_above_50` samples.
- Averaged across the four slices, non-streaming mean RTF and latency decreased by 5.7% and 4.8%, while request and audio throughput increased by 5.0% and 6.0%.
- Streaming mean RTF and latency decreased by 40.9% and 42.6%, first-audio latency decreased by 59.8%, and request and audio throughput increased by 72.5% and 72.4%.

YzXiao101:
## Update: Streaming AudioVAE Cadence

Update:

- Added request-local streaming cadence with one initial latent patch, two steady-state patches per decode, and immediate terminal-residual flush.
- Coalesced ready steady-state requests into shared AudioVAE Qwen2 graph replays while preserving request-local KV, upsample, and ISTFT state.

Benchmark:

Commits: `e73bc5da` (steady `K=1`) -> `eb8f0501` (steady `K=2`). Setup: 1x H200, TP1, concurrency 8, AR, tail, and AudioVAE CUDA graphs enabled, radix cache disabled, streaming PCM responses, 100 SeedTTS samples per slice, `warmup=8`.

| slice | lang | RTF mean (K=1 -> K=2) | latency mean_s (K=1 -> K=2) | first audio mean_s (K=1 -> K=2) | ITL p95_s (K=1 -> K=2) | QPS (K=1 -> K=2) | K=2 chunks mean | WER (K=1 -> K=2) | n_above_50 (K=1 -> K=2) |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| text_only | en | 0.3498 -> 0.3035 (-13.2%) | 1.467 -> 1.279 (-12.8%) | 0.5040 -> 0.4952 (-1.7%) | 0.2930 -> 0.3441 (+17.4%) | 5.321 -> 6.112 (+14.9%) | 6.73 | 0.004550 -> 0.004550 | 0 -> 0 |
| text_only | zh | 0.3254 -> 0.2891 (-11.2%) | 1.435 -> 1.268 (-11.6%) | 0.4810 -> 0.5002 (+4.0%) | 0.2730 -> 0.3555 (+30.2%) | 5.443 -> 6.143 (+12.9%) | 6.77 | 0.003690 -> 0.003690 | 0 -> 0 |
| reference | en | 0.4147 -> 0.3682 (-11.2%) | 1.585 -> 1.438 (-9.3%) | 0.6360 -> 0.6688 (+5.2%) | 0.2720 -> 0.3586 (+31.8%) | 4.970 -> 5.456 (+9.8%) | 6.27 | 0.005460 -> 0.005460 | 0 -> 0 |
| reference | zh | 0.2969 -> 0.2716 (-8.5%) | 1.672 -> 1.521 (-9.0%) | 0.6150 -> 0.6237 (+1.4%) | 0.2550 -> 0.3237 (+26.9%) | 4.675 -> 5.116 (+9.4%) | 8.70 | 0.003163 -> 0.011070 | 0 -> 1 |

Result:

- All 400 requests completed without generation or server errors. The strict gate failed only because the known stochastic `reference/zh` sample `10002539-00000054` produced one WER outlier.

YzXiao101:
## Update: New Default Single-H200 TP1 Baseline

Update:

- Updated the recommended deployment to TP1 on one H200 with AR, acoustic-tail, and AudioVAE CUDA graphs enabled.
- Kept streaming request-controlled and prompt radix caching disabled by default.
- Validated streaming and non-streaming requests independently from the same pipeline configuration.

Benchmark:

Commit: `be1626bf`. Setup: 1x H200, TP1, concurrency 8, AR, tail, and AudioVAE CUDA graphs enabled, radix cache disabled, steady streaming cadence K=2, 100 SeedTTS samples per slice, `warmup=8`.

Streaming:

| slice | lang | rtf_mean | latency_mean_s | first_audio_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| text_only | en | 0.3197 | 1.361 | 0.5220 | 5.748 | 26.137 | 0.004550 | 0 |
| text_only | zh | 0.3053 | 1.338 | 0.5213 | 5.797 | 26.027 | 0.004217 | 0 |
| reference | en | 0.3835 | 1.503 | 0.6971 | 5.256 | 21.798 | 0.005460 | 0 |
| reference | zh | 0.2780 | 1.567 | 0.6560 | 4.955 | 28.541 | 0.003163 | 0 |

Non-streaming:

| slice | lang | rtf_mean | latency_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| text_only | en | 0.1919 | 0.871 | 8.950 | 40.585 | 0.004550 | 0 |
| text_only | zh | 0.1935 | 0.878 | 8.781 | 39.874 | 0.004744 | 0 |
| reference | en | 0.2436 | 1.008 | 7.761 | 32.408 | 0.004550 | 0 |
| reference | zh | 0.1924 | 1.098 | 7.118 | 40.590 | 0.003163 | 0 |

YzXiao101:
@edwingao28 Hi wenyao could you help look at this? thx!

YzXiao101:
## Update: Split AudioVAE CUDA Graph for Review

Update:

- Removed the AudioVAE decoder CUDA graph from this PR to keep the remaining streaming, prompt-cache, full-sequence decode, and cadence changes easier to review independently.
- Moved the decoder graph implementation and follow-up whole-decoder graph work to [#1250](https://github.com/sgl-project/sglang-omni/pull/1250).
- Preserved eager streaming and non-streaming behavior, including steady streaming cadence K=2.

Benchmark:

Commit: `2350eddf`. Setup: 1x H200, TP1, concurrency 8, AR and acoustic-tail CUDA graphs enabled, AudioVAE CUDA graph absent, radix cache disabled, 100 SeedTTS samples per slice, `warmup=8`.

Streaming:

| slice | lang | rtf_mean | latency_mean_s | first_audio_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| text_only | en | 0.3725 | 1.597 | 0.6252 | 4.903 | 21.933 | 0.005460 | 0 |
| text_only | zh | 0.3412 | 1.516 | 0.4737 | 5.140 | 23.289 | 0.005799 | 0 |
| reference | en | 0.3974 | 1.578 | 0.6551 | 5.017 | 20.805 | 0.005460 | 0 |
| reference | zh | 0.3342 | 1.871 | 0.5759 | 4.170 | 23.714 | 0.011597 | 1 |

Non-streaming:

| slice | lang | rtf_mean | latency_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| text_only | en | 0.2341 | 1.034 | 7.554 | 33.505 | 0.008189 | 0 |
| text_only | zh | 0.2093 | 0.932 | 8.383 | 37.448 | 0.003690 | 0 |
| reference | en | 0.2467 | 1.000 | 7.809 | 31.910 | 0.004550 | 0 |
| reference | zh | 0.1911 | 1.116 | 7.036 | 41.068 | 0.002109 | 0 |

Result:

- Both modes completed all 800 generation requests with no serving errors or leftover requests. Non-streaming passed all four WER gates; streaming recorded one known stochastic near-silent `reference/zh` outlier.
- PR #1005 now has one smaller source of truth for the non-graph optimizations, while graph-specific review and further optimization continue in #1250.

YzXiao101:
## Update: Restore Shared AudioVAE Boundary

Update:

- Reverted the helper-level changes to the shared Ming-Omni AudioVAE decoder. They only exposed internal decoder steps for Ming-TTS and unnecessarily expanded this PR's Ming-Omni regression surface.
- Ming-TTS now uses the existing `AudioVAE.decode` contract for both streaming and full-sequence decode, keeping its scheduling and request state model-local without changing the AudioVAE computation.

Benchmark:

Commit: `8a006b79`. Setup: 1x H200, TP1, concurrency 8, AR and acoustic-tail CUDA graphs enabled, AudioVAE CUDA graph absent, radix cache disabled, steady streaming cadence K=2, 100 SeedTTS samples per slice, `warmup=8`.

Streaming:

| slice | lang | rtf_mean | latency_mean_s | first_audio_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| text_only | en | 0.3564 | 1.535 | 0.5238 | 5.060 | 22.796 | 0.004550 | 0 |
| text_only | zh | 0.3493 | 1.547 | 0.5223 | 5.050 | 22.737 | 0.004217 | 0 |
| reference | en | 0.4023 | 1.621 | 0.6763 | 4.819 | 20.402 | 0.004550 | 0 |
| reference | zh | 0.3513 | 1.965 | 0.6056 | 4.009 | 22.706 | 0.010543 | 1 |

Non-streaming:

| slice | lang | rtf_mean | latency_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| text_only | en | 0.2028 | 0.911 | 8.539 | 38.364 | 0.004550 | 0 |
| text_only | zh | 0.1996 | 0.899 | 8.723 | 39.358 | 0.005271 | 0 |
| reference | en | 0.2401 | 0.977 | 7.978 | 32.831 | 0.008189 | 0 |
| reference | zh | 0.1892 | 1.099 | 7.124 | 41.376 | 0.004744 | 0 |

sglang_omni/models/ming_tts/engine_io.py
```diff
@@ -16,21 +18,57 @@
...
+        json.dumps(metadata, sort_keys=True, separators=(",", ":")).encode("utf-8")
+    )
+    for tag, tensor in ((b"speaker", speaker), (b"latent", prompt_latent)):
+        digest.update(tag)
+        if tensor is None:
+            digest.update(b":none")
+            continue
+        value = tensor.detach().contiguous()
+        digest.update(f":{value.dtype}:{tuple(value.shape)}:".encode("ascii"))
+        digest.update(value.numpy().tobytes())
```
edwingao28: With `disable_radix_cache=true`configurationevery reference request still materializes and hashes full byte copies of both conditioning tensors. `ChunkCache` does not consume `Req.extra_key`, so this adds request-sized CPU allocation and scanning without enabling reuse. 

ideally we should:
Cache disabled：extra_key=None。
Cache enabled：use reference encoder generated stable content key first
For inputs without a precomputed key will fall back to the current tensor fingerprint

lmk ur thoughts about this!

tests/unit_test/ming_tts/test_engine_io.py
```diff
@@ -68,16 +102,44 @@ def test_ming_tts_result_adapter_serializes_empty_latent_output() -> None:
...
+                request_id=request_id,
+                request=OmniRequest(inputs="hello"),
+                data=state.to_dict(),
+            )
+        )
+        return data.req.extra_key
+
+    key = extra_key("reference-a", 1.0, 2.0)
+
+    assert extra_key("reference-b", 1.0, 2.0) == key
```
edwingao28: This verifies only that the conditioning namespace changes andnever populates and reuses prompt KV through the enabled RadixCache or exercises retraction.

tests/unit_test/ming_tts/test_engine_io.py
```diff
@@ -124,6 +184,85 @@ def test_ming_tts_result_adapter_infers_length_at_max_steps() -> None:
...
+        "is_last": True,
+    }
+    assert messages[0].data.device.type == "cpu"
+    assert messages[0].data.dtype == torch.float32
+    assert data.pending_stream_patch is None
+    assert build_ming_tts_stream_output("req-ming-tts", data, None) == []
+
+
+def test_ming_tts_streaming_vocoder_initial_and_terminal_cadence() -> None:
+    decoder = _FakeStreamingDecoder()
```
edwingao28: `_FakeStreamingDecoder` verifies scheduler cadence but cannot catch regressions in the real AudioVAE KV/upsample/ISTFT state or terminal tail flush. Add a small CPU parity test that decodes the same latents eagerly and incrementally, then checks output length and numerical parity after the final flush.

YzXiao101:
## Update: Review Follow-ups and Final SeedTTS100 A/B

Update:

- Removed prompt-digest work from the radix-disabled path. The pipeline now resolves `disable_radix_cache` once; the reference stage computes and caches a stable conditioning digest only when radix caching is enabled, and the TTS stage consumes that digest without rescanning the conditioning tensors.
- Tightened prompt-cache identity. The cache namespace identifies text-only or reference conditioning, while position-local synthetic row IDs distinguish injected speaker/prompt-latent rows from literal tokenizer IDs. Generated acoustic feedback remains request-local and is never inserted into the shared radix tree.
- Replaced the cache-key-only tests with real SGLang `RadixCache` coverage for prompt insertion, reuse, reference isolation, shared-prefix matching, row-type isolation, and retraction without generated-suffix insertion.
- Added a small real AudioVAE CPU differential test. It verifies that incremental decode has an empty first non-terminal output, flushes the terminal tail, and matches full-sequence decode in output length and numerical values.
- Removed the unused coalesced-vocoder scheduling path because eager AudioVAE currently decodes one request at a time. The direct cadence path now has a retraction test covering exactly-once latent publication, preserved AR feedback and AudioVAE state, terminal flush, and final PCM output.
- Made the audio-decode contract explicitly single-request instead of advertising unsupported batch-vocoder capability.

Benchmark:

Commit: `5fe8c52d`. Setup: 1x H200, TP1, concurrency 8, eight warmup requests, AR and acoustic-tail CUDA graphs enabled, AudioVAE eager, steady streaming cadence K=2. Each child uses the first 100 SeedTTS cases from text-only/reference and EN/ZH; radix-off ran before radix-on on the same GPU. The four children completed 1,600 generation requests in total.

Streaming:

| slice | lang | RTF mean (off -> on) | latency mean_s (off -> on) | first audio mean_s (off -> on) | QPS (off -> on) | WER (off -> on) | n_above_50 (off -> on) |
| text_only | en | 0.3571 -> 0.3502 | 1.557 -> 1.514 | 0.6640 -> 0.5453 | 5.020 -> 5.147 | 0.006369 -> 0.005460 | 0 -> 0 |
| text_only | zh | 0.3567 -> 0.3555 | 1.561 -> 1.589 | 0.5069 -> 0.5280 | 5.012 -> 4.920 | 0.003690 -> 0.003690 | 0 -> 0 |
| reference | en | 0.4158 -> 0.3926 | 1.630 -> 1.555 | 0.6332 -> 0.6370 | 4.837 -> 5.058 | 0.004550 -> 0.004550 | 0 -> 0 |
| reference | zh | 0.3455 -> 0.3374 | 1.945 -> 1.923 | 0.5917 -> 0.5824 | 4.013 -> 4.072 | 0.011070 -> 0.004744 | 1 -> 0 |

Non-streaming:

| slice | lang | RTF mean (off -> on) | latency mean_s (off -> on) | QPS (off -> on) | WER (off -> on) | n_above_50 (off -> on) |
| text_only | en | 0.2053 -> 0.2043 | 0.928 -> 0.917 | 8.379 -> 8.509 | 0.007279 -> 0.006369 | 0 -> 0 |
| text_only | zh | 0.2024 -> 0.2017 | 0.927 -> 0.915 | 8.435 -> 8.552 | 0.005271 -> 0.003690 | 0 -> 0 |
| reference | en | 0.2427 -> 0.2394 | 1.015 -> 0.997 | 7.692 -> 7.797 | 0.005460 -> 0.005460 | 0 -> 0 |
| reference | zh | 0.1905 -> 0.1898 | 1.099 -> 1.096 | 7.138 -> 7.093 | 0.005271 -> 0.003163 | 0 -> 0 |

YzXiao101:
- Radix identity now keeps model input IDs separate from cache IDs: synthetic out-of-vocabulary tags prevent projected speaker and prompt-latent rows from aliasing ordinary token IDs, while the reference `extra_key` isolates different conditioning values. Requests with the same conditioning can still reuse their unchanged discrete prefix.
- The reference encoder computes one schema-versioned BLAKE2b digest over tensor dtype, shape, and bytes only when radix caching is enabled. It intentionally does not reuse the reference content key because encoding is not guaranteed to be deterministic. The digest is stored in the reference artifact and passed through request state, avoiding repeated tensor copies and hashing in the scheduler adapter. 
- Unit tests now cover digest policy and stability, concrete RadixCache population/reuse/isolation, target-layout divergence, retraction without generated-prefix insertion, and real CPU AudioVAE full-versus-incremental parity through terminal flush.

@edwingao28 Hi wenyao could you help look at this again, thx !

sglang_omni/models/ming_tts/config.py
```diff
@@ -82,22 +87,114 @@ def process_edge_resources(
...
+                "Ming-Omni-TTS audio_vae_steady_chunk_patches must be positive"
+            )
+        stages = {stage.name: stage for stage in self.stages}
+        required_stages = {
+            PREPROCESSING_STAGE,
+            REFERENCE_ENCODE_STAGE,
+            TTS_ENGINE_STAGE,
+            AUDIO_DECODE_STAGE,
+        }
+        missing_stages = required_stages - stages.keys()
```
edwingao28: This checks that the stages exist, but not that `tts_engine` actually routes to `audio_decode` via `stream_to` — and streaming depends on that edge, not just on the stage being present.

A config written before this PR has all four stages but no such edge. It loads fine, non streaming works fine, and then every `stream=true` request fails at runtime with "no endpoint configured for stream target
'audio_decode'". Schema validation won't catch it either: it only verifies that `stream_to` entries name known stages, not that the required one is there.

Can you asserting the edge here so stale configs fail at load instead of per request?

sglang_omni/models/ming_tts/streaming_vocoder.py
```diff
@@ -0,0 +1,205 @@
...
+            if not state.terminal_received:
+                raise RuntimeError(
+                    f"Ming-Omni-TTS stream {request_id!r} ended without a decoded "
+                    "terminal latent patch"
+                )
+            if state.emitted_samples <= 0:
+                raise RuntimeError(
+                    f"Ming-Omni-TTS stream {request_id!r} completed without audio"
+                )
+            return None
```
edwingao28, line 169: When a request fails before its first tail step it still reaches on_stream_done with zero patches, so the first raise masks the original engine error with "ended without a decoded terminal latent patch".

Returning None from both raises lets on_stream_done fall through to fallback_full_decode, which keeps the terminal result audio free

edwingao28: resolved as engine exception wont send stream_done

YzXiao101:
## Update: Validate the Required Streaming Edge

Update:

- Ming-TTS pipeline configs now require `tts_engine.stream_to` to include `audio_decode`, so an invalid custom config fails during config loading instead of on its first streaming request.
- Added focused config UT coverage for both the default edge and missing-edge rejection.

## Regress Test

Commit: `14b55a7a`. Setup: 1x H200, TP1, concurrency 8, AR and acoustic-tail CUDA graphs enabled, AudioVAE CUDA graph absent, radix cache disabled, streaming PCM responses with steady cadence K=2, 100 SeedTTS samples per slice, `warmup=8`.

| slice | lang | requests | failed | rtf_mean | latency_mean_s | first_audio_mean_s | throughput_qps | audio_s/s | WER | n_above_50 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| text_only | en | 100 | 0 | 0.3602 | 1.533 | 0.5360 | 5.107 | 22.763 | 0.004550 | 0 |
| text_only | zh | 100 | 0 | 0.3510 | 1.555 | 0.5074 | 5.019 | 22.629 | 0.003690 | 0 |
| reference | en | 100 | 0 | 0.4106 | 1.632 | 0.6936 | 4.796 | 20.028 | 0.004550 | 0 |
| reference | zh | 100 | 0 | 0.3408 | 1.926 | 0.5599 | 4.062 | 23.293 | 0.002636 | 0 |

Result:

- All 400 streaming requests completed successfully with no WER outliers after merging the latest main branch, exercising the required `tts_engine -> audio_decode` stream edge end to end.

docs/cookbook/ming_tts.md
```diff
@@ -129,7 +160,7 @@ python -m benchmarks.eval.benchmark_tts_seedtts \
   --meta zhaochenyang20/seed-tts-eval-arrow \
   --output-dir results/ming_tts_reference_en \
   --lang en --ref-format references \
-  --max-new-tokens 256 --max-concurrency 8 --warmup 0
+  --max-new-tokens 256 --max-concurrency 8 --warmup 8
```
edwingao28: The reproduction command only runs the reference, non-streaming case, while the tables below include text-only and streaming results. Those variants need `--no-ref-audio` and/or `--stream` on both generation and transcription, and each run needs a separate `--output-dir`; otherwise `generated.json` is overwritten and the saved result metadata can be mislabeled. Please document the full flag and output-directory matrix used for the reported modes.

docs/cookbook/ming_tts.md
```diff
@@ -89,16 +96,37 @@ curl -X POST http://localhost:8000/v1/audio/speech \
 
 `ref_audio` and `ref_text` are accepted as shorthand for the single `references` item.
 
+### Streaming
+
+Streaming responses use raw signed 16-bit 44.1 kHz PCM. The decoder first consumes one latent
```
edwingao28: `output.pcm` is headerless, but the guide only specifies the bit depth and sample rate; the channel count and byte order are missing, and the example does not show how to consume the response incrementally. Please document the guaranteed PCM wire format and use `curl -N` with a minimal ffplay/ffmpeg example so users can correctly interpret and observe the streamed chunks.

sglang_omni/models/ming_tts/config.py
```diff
@@ -82,22 +87,119 @@ def process_edge_resources(
...
+        if missing_stages:
+            raise ValueError(
+                "Ming-Omni-TTS pipeline is missing required stages: "
+                f"{sorted(missing_stages)}"
+            )
+        preprocessing = stages[PREPROCESSING_STAGE]
+        reference_encode = stages[REFERENCE_ENCODE_STAGE]
+        tts_engine = stages[TTS_ENGINE_STAGE]
+        audio_decode = stages[AUDIO_DECODE_STAGE]
+        if AUDIO_DECODE_STAGE not in tts_engine.stream_to:
```
edwingao28: The new `stream_to` check validates only the producer half of the pre-payload streaming contract. `audio_decode.can_accept_stream_before_payload` still defaults to `False`, so a custom Ming config can pass `model_post_init` and then abort every streaming request as soon as the first latent arrives before the terminal payload. Please validate that the target stage has `can_accept_stream_before_payload=True` here and add a config regression test.

sglang_omni/models/ming_tts/stages.py
```diff
@@ -199,19 +146,20 @@ def create_audio_decode_executor(
...
-    max_batch_size: int = 1,
-    max_batch_wait_ms: int = 0,
-) -> SimpleScheduler:
-    if decode_mode != "chunked":
-        raise ValueError("Ming-Omni-TTS currently supports only decode_mode='chunked'")
-
-    from sglang_omni.models.ming_tts.audio_decode import (
-        MingAudioDecoder,
-        MingTTSBatchVocoder,
+    audio_vae_steady_chunk_patches: int = 2,
```
edwingao28: Removing the exported `create_tts_engine_executor` alias and the previously accepted `decode_mode`, `max_batch_size`, and `max_batch_wait_ms` factory arguments makes imports and custom Ming configs based on the base version fail at import or startup. Please keep a deprecated forwarding alias and accept the legacy arguments when set to their former defaults, rejecting non-default values, or explicitly document this as a breaking config/API change.

docs/cookbook/ming_tts.md
```diff
@@ -153,45 +184,47 @@ python -m benchmarks.eval.benchmark_tts_seedtts \
...
 
-### Recommended TP2
+### Recommended Single-H200 TP1
 
-The recommended TP2 configuration was evaluated on the complete
-**Seed-TTS-Eval EN and ZH splits** with **2× H100 80 GB**, concurrency 8, AR and tail CUDA
-graphs enabled, and **Qwen3-ASR-1.7B** for transcription.
+The recommended TP1 configuration was evaluated on **1× H200 141 GB** with concurrency 8,
+eight warmup requests, and the full Seed-TTS-Eval EN and ZH splits. AR and acoustic-tail CUDA
+graphs were enabled, while AudioVAE decode remained eager and prompt radix caching was disabled.
```
edwingao28: Please update your streaming & non streaming results once we passed all ci and ready to merge this pr

docs/cookbook/ming_tts.md
```diff
@@ -42,6 +42,13 @@ sgl-omni serve \
   --port 8000
 ```
 
+The provided configuration enables the AR and acoustic-tail CUDA graphs. AudioVAE decode remains
+eager. Streaming and prompt radix caching remain opt-in: requests are non-streaming unless
+`stream` is set, and `disable_radix_cache` is `true` in the TTS engine configuration. To reuse
```
edwingao28: Prompt radix caching is disabled by default and this PR's measurements show neutral-to-slower end-to-end performance on the stated workload, yet the feature adds conditioning digests, synthetic cache IDs, cache-insertion policy, and retraction plumbing across the pipeline. This substantially enlarges the correctness and review surface of the streaming change without a demonstrated target-workload gain. If it stays here, please add a cache-friendly repeated-prompt benchmark that demonstrates an end-to-end benefit.

sglang_omni/models/ming_tts/engine_io.py
YzXiao101: Fixed in fc974581: removed the conditioning digest and all prompt-cache-only work from the supported radix-disabled path because prompt radix caching is no longer part of this PR.

tests/unit_test/ming_tts/test_engine_io.py
YzXiao101: Addressed in fc974581: the cache-key-only tests were removed together with the prompt-cache feature rather than retained as misleading lifecycle evidence for behavior that this PR no longer provides.

tests/unit_test/ming_tts/test_engine_io.py
YzXiao101: Fixed in fc974581: added a small real CPU AudioVAE differential test that decodes identical latents in full and incremental modes and verifies the empty lookahead output, terminal flush, output length, and numerical parity.

sglang_omni/models/ming_tts/config.py
YzXiao101: Fixed in fc974581 and tightened in 4a60f3dc: Ming config now requires the exact tts_engine.stream_to=[audio_decode] producer edge so stale or ambiguous topologies fail during config loading instead of on the first streaming request.

docs/cookbook/ming_tts.md
YzXiao101: Fixed in [08b71d22](https://github.com/sgl-project/sglang-omni/pull/1005/commits/08b71d225877b3fbf6492f72e9aa57cc45955b1e): the Cookbook now documents the complete streaming/non-streaming × text-only/reference × EN/ZH matrix, with matching flags in both generation and transcription and a separate output directory for every scenario.

docs/cookbook/ming_tts.md
YzXiao101: Fixed in [d59e95bc](https://github.com/sgl-project/sglang-omni/pull/1005/commits/d59e95bcfa764b71438ca1d31965e11d1fdafc4c): the Ming cookbook now specifies headerless mono signed 16-bit little-endian PCM at 44.1 kHz, its response headers and EOF contract, and incremental curl --no-buffer | ffplay plus saved-PCM ffmpeg examples.

sglang_omni/models/ming_tts/config.py
YzXiao101: Fixed in [74e07b51](https://github.com/sgl-project/sglang-omni/pull/1005/commits/74e07b516424f766a336509921d290e43bea6610) and incorporated into the exact topology contract in [4a60f3dc](https://github.com/sgl-project/sglang-omni/pull/1005/commits/4a60f3dc4170b6978bac651a192f2ebbf5405d42): Ming config now requires audio_decode.can_accept_stream_before_payload=true, with a regression test that rejects the invalid consumer before startup.

sglang_omni/models/ming_tts/stages.py
YzXiao101: Fixed in [411a1834](https://github.com/sgl-project/sglang-omni/pull/1005/commits/411a183464ad0df5fa84ed9e57f2d45fca54a9ce): we restored the create_tts_engine_executor forwarding alias and preserved the truthful max_batch_size=1/max_batch_wait_ms=0 shape with early rejection of other values, while documenting and rejecting legacy decode_mode because accepting chunked while executing full-sequence decode would provide false compatibility.

docs/cookbook/ming_tts.md
YzXiao101: Addressed in [fc974581](https://github.com/sgl-project/sglang-omni/pull/1005/commits/fc9745811503f373aa1fb3d26b2218f2f878230e): we removed the complete prompt-cache implementation and kept radix enablement fail-fast because, although an ideal 3,524-token exact-repeat C1 case achieved 99.97% reuse and 18.84% higher QPS, the standard fixed-reference C8 workload reused 59.72%/57.09% of prompt tokens while EN QPS improved only 3.34% and ZH QPS regressed 3.00% with a 16.16% TTFA-p95 regression, so it did not demonstrate stable target-workload ROI for this review surface.

sglang_omni/models/ming_tts/stages.py
YzXiao101: Re-audited the complete Ming-TTS unit-test suite, configuration ownership and topology, and user-facing documentation end to end, reorganizing coverage and fixing additional contract gaps rather than limiting the changes to the reported lines.

edwingao28 (approved):

YzXiao101:
## Update: Initial and Steady Streaming Cadence

Update:

- Split streaming AudioVAE cadence into pipeline-level `initial_chunk_patches=M` and `steady_chunk_patches=K`; both are configurable without changing decoder math.
- Set the default to `M=2,K=4`. Terminal residuals still flush immediately.

Benchmark:

commit `1ed62262`, 1x H200, TP1, concurrency 8, `warmup=8`, AudioVAE eager, 400 SeedTTS requests per cadence.

| Cadence | Failed | RTF mean | QPS mean | First audio mean / p95 worst (s) | C50 / C100 / C200 worst | Later underruns / seams | Corpus WER worst |
| M1/K4 | 0 | 0.2823 | 6.1045 | 0.628 / 0.936 | 32% / 66% / 98% | 0 / 734 | 0.64% |
| **M2/K4** | **0** | **0.2811** | **6.1215** | **0.673 / 1.024** | **100% / 100% / 100%** | **0 / 642** | **1.05%** |

Final full-corpus validation used the same selected cadence: all 12,432 requests completed, worst-slice WER was 1.09%, every scoreable request reached C50/C100/C200, and all 11,641 later seams had zero measured underrun.

**The complete streaming and non-streaming full-corpus results are included in the PR description above. This update highlights only the cadence comparison used to select M=2,K=4.**

Result:

- Moving from M1/K4 to M2/K4 adds 45 ms mean first-audio latency while leaving RTF and QPS effectively unchanged, and **removes the measured first-seam underrun**.
- K=1/2 pays substantially more AudioVAE invocation overhead. K=6/8 is faster but delays first audio and emits fewer, larger PCM payloads. K=4 is the balanced steady cadence for this workload.

@luojiaxuan, Hi jaxan could you please help review the first-chunk cadence design in this update? thx ~

sglang_omni/models/ming_tts/engine_io.py
```diff
@@ -173,7 +157,36 @@ def result_adapter(data: MingTTSSGLangRequestData) -> StagePayload:
...
+    request_id: str,
+    data: MingTTSSGLangRequestData,
+    req_output: Any,
+) -> list[OutgoingMessage]:
+    del req_output
+    patch = data.pending_stream_patch
+    if not data.is_streaming or patch is None:
+        return []
+
+    latent = patch.latent.detach().to(device="cpu", dtype=torch.float32).contiguous()
```
luojiaxuan, line 170: Can we keep the latent on GPU to avoid the GPU-to-CPU-to-GPU copy here?

luojiaxuan (approved):

sglang_omni/models/ming_tts/engine_io.py
YzXiao101: Thanks for the review. I had noticed this issue as well. The same GPU→CPU→GPU round-trip also exists in the non-streaming path, so I plan to optimize both paths together in a separate PR. The idea is to keep acoustic latents on GPU through audio_decode, let the pipeline runtime select local-object or CUDA IPC transport, and only move the final waveform to CPU.

PR #1007, opened 2026-07-09
[Qwen3-Omni] Capture talker predictor decode with CUDA graph

Hayden727:
## Summary

- Add a lazy per-`(batch_size, code_dtype)` CUDA graph cache for Qwen3-Omni talker predictor single-token decode.
- Keep eager fallback for prefill, CPU/non-CUDA inputs, unsupported token dtypes, active CUDA stream capture, and graph capture failures.
- Add unit coverage for `thread_local` capture mode and the serving-realistic `torch.int32` token id path.

## Motivation

Qwen3-Omni talker decode spends a meaningful fraction of time in the code predictor, which launches many small kernels for each generated audio token. Capturing the single-token predictor path amortizes Python/kernel launch overhead while preserving the existing eager implementation as the fallback.

The graph cache is dtype-aware because SGLang serving passes sampled token ids as `torch.int32`, while many local/unit paths use `torch.long`.

## Performance

Measured on `hyper00-omni`, H200 GPU 3, `hongccc/sglang-omni:dev`, SeedTTS-50 generate-only, `max_concurrency=4`, `warmup=10`, `max_samples=50`.

| Variant | QPS | Mean latency | Mean RTF |
| --- | ---: | ---: | ---: |
| main (`b09dc99`) | 2.605 | 1.514s | 0.4705 |
| predictor CUDA graph (`b09dc99` + patch) | 3.404 | 1.160s | 0.3466 |

This is +30.7% QPS and -23.4% mean latency in that run. The branch has since been rebased onto latest `origin/main` (`c00e6f9d`).

I also tested stacking `torch.compile` inside the manual CUDA graph. It worked, but only added ~1.6% QPS over the manual CUDA graph while adding roughly 2+ minutes of cold capture/compile startup across decode buckets, so this PR keeps the lower-risk manual graph path only.

## Validation

- Local: `py_compile`, `black --check`, `isort --check-only`, `git diff --check`
- Commit hooks: passed during commit
- H200 after rebase onto `origin/main`: `CUDA_VISIBLE_DEVICES=3 TMPDIR=/data/tmp pytest -q tests/unit_test/qwen3_omni/test_talker.py -k "code_predictor or cuda_graph_capture or predictor_decode_graph"`
  - `4 passed, 61 deselected`

### DIFF
--- a/sglang_omni/models/qwen3_omni/components/talker.py
+++ b/sglang_omni/models/qwen3_omni/components/talker.py
@@ -103,24 +103,32 @@ class _PredictorDecodeGraph:
     @torch.no_grad()
     def _capture(self) -> None:
         device = self.layer0_codes.device
-        warmup_stream = torch.cuda.Stream(device=device)
-        current_stream = torch.cuda.current_stream(device=device)
-        warmup_stream.wait_stream(current_stream)
-        with torch.cuda.stream(warmup_stream):
-            for _ in range(2):
-                self.model._code_predictor_forward_incremental_eager(
-                    self.layer0_codes,
-                    self.talker_hidden,
-                )
-        current_stream.wait_stream(warmup_stream)
+        with torch.cuda.device(device):
+            warmup_stream = torch.cuda.Stream(device=device)
+            current_stream = torch.cuda.current_stream(device=device)
+            warmup_stream.wait_stream(current_stream)
+            with torch.cuda.stream(warmup_stream):
+                for _ in range(2):
+                    self.model._code_predictor_forward_incremental_eager(
+                        self.layer0_codes,
+                        self.talker_hidden,
+                    )
+            current_stream.wait_stream(warmup_stream)
 
-        with torch.cuda.graph(self.graph, capture_error_mode="thread_local"):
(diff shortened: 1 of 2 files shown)


yxs:
@Hayden727 could you rerun a few rounds on the same setup and report mean/std? I reran it with same set up and got QPS +19.4%, mean latency -17.1%, mean RTF -17.8%.

Hayden727:
Thanks, good call. I reran 3 rounds on `hyper01` with the same SeedTTS-50 setup (c=4, warmup=10, voice-clone).

Mean/std:
- main: QPS 2.701 +/- 0.429, latency 1.486 +/- 0.246s, RTF 0.4679 +/- 0.0826
- this PR: QPS 3.070 +/- 0.524, latency 1.310 +/- 0.250s, RTF 0.4091 +/- 0.0845

So the mean delta is +13.7% QPS, -11.8% latency, -12.6% RTF. I think my earlier +30% number was from a single run that caught a cold/warm-state mismatch. The predictor graph is captured lazily, so the first pass is a bit noisy; hot PR rounds were still around 3.28-3.46 QPS.

yxs (approved):

PR #1010, opened 2026-07-09
[Async]: Fix async decode stale batch KV slot ownership

Ratish1:
## Summary

Fix Omni async decode stale-batch handling so an in-flight lookahead step is resolved before scheduling pending prefill/chunked work, and preserve decode `out_cache_loc` when stale decode rows are filtered after a pending-step drain.

This keeps the sync scheduler path unchanged and only touches the async decode event loop. The focused CPU regression coverage now checks that pending async work is drained before waiting prefill scheduling and that stale decode filtering keeps the prepared KV slot locations for surviving rows.

## Root Cause

The async decode loop could build the next batch with `get_next_batch_to_run()` while a previous lookahead decode step was still pending. If the pending step then finished or retracted a request, `_drop_stale_overrun()` filtered that already-prepared stale batch before forwarding it.

SGLang `ScheduleBatch.filter_batch()` intentionally clears `out_cache_loc`, because generic filtering invalidates the current forward allocation metadata. The async loop then forwarded the filtered batch without rebuilding it, so the backend could receive `out_cache_loc=None` as the KV-store indices. In the failing video workloads this surfaced downstream as `store_cache` receiving `None` for the DLTensor indices argument, followed by CUDA illegal memory access / thinker process abort.

The fix avoids prepared extend/mixed stale-batch surgery by draining pending async decode before prefill/chunked scheduling, and keeps the remaining stale-overrun filtering on the decode-shaped path where `out_cache_loc` is row-aligned and can be restored for kept rows.

### DIFF
--- a/tests/unit_test/pipeline/test_async_decode.py
+++ b/tests/unit_test/pipeline/test_async_decode.py
@@ -521,6 +521,8 @@ def _new_scheduler_for_async_loop():
     s._admin_lock = threading.Lock()
     s._admin_queue = queue.Queue()
     s.chunked_req = None
+    s.is_mixed_chunk = False
+    s.running_batch = types.SimpleNamespace(batch_is_full=False)
     s.waiting_queue = []
     return s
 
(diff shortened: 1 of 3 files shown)


Hayden727 (approved):

tests/unit_test/pipeline/test_async_decode.py
Ccyest, line 517: Nit: fake_batch doesn't have out_cache_loc, this doesn't trigger errors today (because _FakeBatch reqs never finish, so _drop_stale_overrun returns early), but if a future test case makes a _FakeBatch req finish and enters the drop path of _drop_stale_overrun, it will raise AttributeError.

Ccyest (approved):
lgtm, just some nit fix

tests/unit_test/pipeline/test_async_decode.py
Ratish1: Addressed in 543f0a64. `_FakeBatch` now carries `out_cache_loc` and mirrors `ScheduleBatch.filter_batch()` clearing it. I also moved the no-op allocator fields into the shared scheduler fixture so a future finished-row case can execute the complete stale-drop path instead of failing at the next missing fixture attribute.

Issue #999, opened 2026-07-09
[Ming] P0 Optimization Tracking


AkazaAkane:
# [Ming] P0 Optimization Tracking

Tracking of P0 Optimization of Ming-Omni, Part of #254.

### 1. Slim pipeline payloads #1000 

* Add `project_payload` for thinker → decode / talker / segmenter.
* Clear `encoder_outs` after aggregate builds `thinker_inputs`.
* Verify with relay payload-byte assertion and profiler hop latency.

### 2. Mel offload #1013 

* Offload mel computation with `asyncio.to_thread`.
* Current mel preprocessing can block the event loop under concurrent requests.
* Verify with concurrent audio request smoke test and latency comparison.

### 3. Talker duration guard O(n²) fix #1016

* Replace repeated `torch.cat(all_wavs).shape` length checks with a running sample counter.
* Keep final `torch.cat` only for final audio assembly.
* Verify same duration-limit behavior and reduced overhead on multi-chunk requests.

## Success Criteria

* Smaller stage relay payloads.
* No event-loop blocking from mel preprocessing.
* Talker duration checks become O(1) per chunk.
* No output correctness regression.

edwingao28:
Please fix the merge conflicts in all three PRs, @AkazaAkane

AkazaAkane:
> Please fix the merge conflicts in all three PRs, [@AkazaAkane](https://github.com/AkazaAkane)

All addressed.

edwingao28:
close this issue as all three prs have been merged

PR #1013, opened 2026-07-09
[Ming-Omni] Offload mel preprocessing from event loop

AkazaAkane:
### Motivation

#999 Ming-Omni preprocessing currently loads audio through `asyncio.to_thread`, but mel extraction still runs synchronously inside `MingPreprocessor.__call__`. Since preprocessing is an async stage, synchronous mel extraction can block the event loop and delay unrelated concurrent requests.

### Changes

- Added a small synchronous helper for per-waveform mel feature extraction.
- Offloaded mel computation with `asyncio.to_thread`.
- Preserved existing output structure:
  - `mel_features_list`
  - `mel_lengths_list`
  - `audio_token_counts`
  - placeholder count calculation before prompt construction

### Validation

- Unit test verifies Ming mel extraction is offloaded through `asyncio.to_thread`.
- Existing Ming pipeline tests pass.

### Expected impact

This does not reduce single-request mel compute cost. It improves concurrent request behavior by preventing CPU mel extraction from blocking the preprocessing event loop.


tests/unit_test/ming_omni/test_pipeline.py
```diff
@@ -223,6 +223,30 @@ def test_ming_audio_encoder_moves_inputs_to_component_device() -> None:
...
+def test_ming_preprocessor_offloads_mel_to_thread() -> None:
+    from sglang_omni.models.ming_omni.components.preprocessor import MingPreprocessor
+
+    source = Path("sglang_omni/models/ming_omni/components/preprocessor.py").read_text(
+        encoding="utf-8"
+    )
+    call_source = inspect.getsource(MingPreprocessor.__call__)
+
+    assert "_compute_mel_features_for_waveform" in source
+    assert (
```
edwingao28: Can we test _compute_mel_features_for_waveform() directly here? Source matching is brittle and doesn’t protect the returned tuple contract. We can monkeypatch compute_mel_spectrogram and assert (mel_tensor, mel_len, audio_token_count) instead.

sglang_omni/models/ming_omni/components/preprocessor.py
```diff
@@ -513,18 +528,27 @@ async def __call__(self, payload: StagePayload) -> StagePayload:
...
-                estimate_audio_feature_length(
-                    mel.shape[0],
-                    getattr(self._audio_config, "ds_kernel_size", 1),
-                    getattr(self._audio_config, "ds_stride", 1),
-                )
+        if waveforms:
+            ds_kernel_size = getattr(self._audio_config, "ds_kernel_size", 1)
+            ds_stride = getattr(self._audio_config, "ds_stride", 1)
+
+            mel_results = await asyncio.gather(
```
edwingao28, line 535: nit: I think the default fail fast behavior of `asyncio.gather` is the right choice here, but can u make that intent explicit with `return_exceptions=False`? The image/video gather path above uses `return_exceptions=True`, so spelling this out would keep future readers from accidentally making mel failures look recoverable.

tests/unit_test/ming_omni/test_pipeline.py
AkazaAkane: Replaced brittle source-matching tests with a direct monkeypatched _compute_mel_features_for_waveform() tuple-contract test

sglang_omni/models/ming_omni/components/preprocessor.py
AkazaAkane: Added return_exceptions=False to the Ming mel asyncio.gather call

edwingao28 (approved):

edwingao28:
lgtm as it is minor fix for ming only and passed all ci

PR #1014, opened 2026-07-09
[router] Overload protection: admission control + unified --max-connections semantics

JiaxinD:
Measured today: past ~0.75x of its capacity the relay does not degrade, it collapses (goodput drops to 17 to 31% of optimum, clients see an FD-exhaustion TCP error storm or mass timeouts), and a single bad-input client can evict every healthy worker. Five commits, one logical unit each:

1. **Eviction signal:** relayed 500s no longer feed the worker-eviction circuit (`WORKER_EVICTION_STATUS_CODES = {408, 429, 502, 503, 504}`); they are labeled `outcome=upstream_5xx` and still counted in worker stats. Measured incident: one misbehaving client produced 78x500 and turned a healthy Qwen3-Omni pool into 757x503 for everyone else. A genuinely sick worker is still caught by `/health` probes. (Whether 429 should also move out is left as a review question.)
2. **nofile check:** startup compares the `RLIMIT_NOFILE` soft limit against `2 x upstream pool size + headroom` and warns with the concrete `ulimit -n` to run; `--strict-limits` fails fast. The measured FD-storm collapse came from exactly this silent misconfiguration.
3. **Admission control (core):** a global in-flight bound over the model data plane; at the bound the router fast-rejects with 503 + `Retry-After: 1` + an OpenAI-style envelope, before reading the request body. Health/admin endpoints are never gated; `GET /health` reports `inflight` / `peak_inflight` / `rejected_total`. The slot is released in the streaming iterator `finally`, on every early-return and exception path, with a GC-finalizer backstop and a fail-closed ordering under a raising upstream close (both from adversarial cross-review, both with failing-before tests).
4. **`--max-connections` becomes the admission bound:** previously it capped the httpx pool, so excess requests queued inside the pool until `PoolTimeout` surfaced as a late 502. Now it is the admission bound and the pool is sized to `max(max_connections, max_inflight)`, eliminating that failure mode structurally; the 128-per-worker auto default (#938) carries over as the default bound. One measured design point: the bound is a latency knob, not just a divergence backstop; an oversized bound on a single core degrades service itself (see evidence), so docs carry sizing guidance (`capacity x acceptable latency`). Rejecting before reading the body costs the client its keep-alive connection, which is intentional: draining large bodies would make shedding expensive exactly when it must be cheap.
5. **Docs:** `omni_router.md` gets the new `--max-connections` meaning, the rejection contract, an Overload Behavior section, and the updated eviction semantics.

## When this pays off (real same-GPU DP pool, one H100)

The mock rig isolates router mechanics; this section is the same A/B against a real pool: 3 Higgs TTS workers on one H100 (the #921 same-GPU DP deployment shape; time-sliced here, the recipe adds MPS for higher capacity), real voice-clone payloads, router pinned to one core. [measured (real)]

At normal load the routers are invisible and identical (closed loop, c=288, 82 s):

| target | goodput qps |
|---|---|
| workers direct (3 clients, no router) | 80.1 |
| via base router | 80.6 |
| via hardened router | 81.2 |

Past capacity they diverge (open loop, offered 200 = 2.5x pool capacity, nofile=1024, 90 s):

| router | goodput qps | TCP-level client errors | timeouts | late 502s | 503s (rej p50 / p99 ms) | p50 of 200s |
|---|---|---|---|---|---|---|
| base | 51.2 (64% of capacity) | 9369 | 744 | 1380 | 1364 (transient evictions) | 9.4 s |
| hardened (auto bound 384) | **79.6 (~100% of capacity)** | **0** | **0** | 0 | 10392, 0.4 / 5.6 | 3.8 s |

`/health` after the run confirms the accounting: `peak_inflight=384` (the auto bound), `rejected_total=10392`, `inflight=0`. So the payoff scenario is exactly the DP-dense deployment #921 is building toward: DP (and DP+MPS) multiplies the worker capacity behind one relay, and the auto bound scales with worker registration (128 x N); when offered load spikes past pool capacity, the base router loses a third of the pool's throughput to connection churn while every client sees raw TCP errors and timeouts, and the hardened router keeps the pool saturated and converts all excess into sub-6 ms 503s that clients can retry on.

One honest note on this rig: with real workers (multi-second service times), the 128-per-worker auto bound is close to right-sized, because in-flight work is dominated by requests genuinely executing in workers. The sizing guidance in commit 4 matters most for thin-payload or many-worker shapes where the same auto bound would exceed `capacity x acceptable latency`.

## Omni compatibility (real Qwen3-Omni speech pipeline through the hardened router)

Full speech pipeline (thinker + talker + code2wav on 2 GPUs), 342 KB base64-audio voice-clone requests, direct vs via the hardened router. No fallback path exists or is needed: omni traffic goes through the same relay as everything else. [measured (real)]

| check | direct | via hardened |
|---|---|---|
| non-stream speech (2.4 MB audio JSON) | 0.24 qps, p50 4.53 s, body 2402029 B | 0.22 qps, p50 4.49 s, body 2388293 B, intact |
| SSE speech streaming | 46 audio chunks p50, first chunk 129 ms, `[DONE]` all | 47 chunks p50, first chunk 132 ms (+3 ms), `[DONE]` all, content-type preserved |
| image+audio multimodal (1.3 MB requests) | (same worker) | all 200, concurrent with normal speech traffic, zero errors |
| worker pool after the run | | healthy, `consecutive_failures=0`; admission `peak_inflight=4`, `rejected_total=0` |

Streaming semantics, big-body relay, and multimodal routing are unchanged through the hardened head, and admission is invisible at omni's request rates (the worker's own concurrency cap binds about 30x earlier than the router's auto bound). The bad-input eviction scenario itself is covered by the behavior-side replay below and by the original incident measurement in commit 1's message.

## Tests

19 new unit tests across the five commits, all failing-before where behavior changed (eviction, fast-reject, slot release on every path, pool sizing, nofile modes, exemptions and stats). Full router suite green on Linux.

### Protection side (open loop, 300KB response shape, nofile=1024, 60 s rounds; offered 680 = ~0.9x capacity, 900 past it)

| router | offered | goodput qps | p50 / p99 of 200s (ms) | TCP-level client errors | 503s (rej p50 / p99 ms) |
|---|---|---|---|---|---|
| base (origin/main), no admission | 680 | 237.6 | 2509 / 13668 | 22023 | 2140 (transient evictions) |
| hardened, auto bound (128 x 3 = 384) | 680 | 255.1 | 1068 / 6135 | 403 | 24268, 19.6 / 53.5 |
| hardened, bound 64 (x2) | 680 | 484.5 / 486.0 | ~89 / 570 to 580 | 0 | ~11650, 6.0 / 17 |
| hardened, bound 64 | 900 | 467.9 | 88 / 633 | 0 | 26105, 6.3 / 16 |

Right-sized bound: zero TCP errors, all excess as fast 503 (p50 6 ms), p99 0.6 s vs 13.7 s in the collapse, goodput 2x the collapsed baseline. The auto-bound row motivates the sizing guidance.

### PoolTimeout elimination (misconfigured `--max-connections 64 --max-inflight 800`, offered 900, 5 s timeout)

| head | `PoolTimeout` log lines | late 502s | fast 503s | client timeouts |
|---|---|---|---|---|
| before pool balancing | 3644 | 4030 | 12719 (rej p50 90 ms) | 758 |
| after pool balancing | 0 | 590 | 30041 (rej p50 24 ms) | 0 |

Residual 502s are this misconfiguration's FD exhaustion, which the pool-size-based nofile check now flags at startup (the old check stayed silent on exactly this case).

### Cost side (closed loop c=24, ABAB x3 per side, base = origin/main)

| shape | version | CPU ms/req mean (x3) | qps mean |
|---|---|---|---|
| 1 B response | base | 1.049 | 952.1 |
| 1 B response | hardened | 1.039 | 956.1 |
| 300 KB response | base | 1.509 | 660.6 |
| 300 KB response | hardened | 1.484 | 672.5 |

Fixed-layer delta -1.0% / -1.7%, inside round-to-round noise (bar was < +5%); throughput flat.

### Behavior side (bad-input 500 replay: 3 flaky mock workers, healthy client B c=8 + bad client A 20 poison/s, threshold 2)

| run | normal client B | pool at end |
|---|---|---|
| before, with bad client | 24.0 qps of 200s, drowned in 236583x503 | all 3 workers unhealthy |
| after, with bad client (x3) | 152.2 to 152.7 qps, 100% 200s | all 3 healthy |
| after, solo baseline (x3) | 150.0 to 150.8 qps | all 3 healthy |

B with the bad client present matches B alone (within noise): relaying the 500s costs healthy traffic nothing.

## Boundary

* The worker `active_requests` count keeps its existing exposure on the never-started-iterator path (only the admission slot got the GC backstop; a leaked slot locks the router out globally, a skewed worker count only biases least_request). Follow-up if wanted.
* Why not uvicorn `--limit-concurrency`: bare 503 without `Retry-After` or envelope, counts every connection including health/admin, no route scoping or logging.


sglang_omni_router/proxy.py
```diff
@@ -115,8 +172,57 @@ def __init__(
...
+            release()
+            raise
+        # The streaming relay hands its slot to the response iterator; every
+        # other path is done with it as soon as the response object exists.
+        if not isinstance(response, StreamingResponse):
+            release()
+        else:
+            # Note (Jiaxin Deng): if the ASGI stack drops the response without
+            # ever starting the iterator, GC frees the slot (release is idempotent).
+            weakref.finalize(response, release)
```
Ratish1, line 274: This fallback must close the upstream response and decrement the worker count, not only release admission. By this point `_forward_relay` has already opened the upstream stream and incremented `worker.active_requests`. Starlette sends `http.response.start` before it begins iterating the body, so a client disconnect during that send never enters the iterator's `finally`.

can use one idempotent async cleanup owner for upstream close, worker accounting, and admission release, and invoke it from both iterator completion and a response-level `__call__` `finally`. Add an ASGI regression whose `send` raises on `http.response.start` and assert all three resources are released exactly once.

sglang_omni_router/proxy.py
```diff
@@ -258,7 +365,7 @@ def record_worker_failure_once(
                 error=error,
             )
 
-        if upstream.status_code in WORKER_REQUEST_FAILURE_STATUS_CODES:
+        if upstream.status_code in WORKER_EVICTION_STATUS_CODES:
```
Ratish1, line 425: Do not feed worker-capacity responses into the liveness circuit. This set still includes `429`, and it includes `503` even though the existing router test uses a worker `503` body of `scheduler overloaded, retry later`. These responses can mean a reachable, healthy worker is applying backpressure.

With one healthy worker and `health_failure_threshold=2`, two relayed `429`s on this head produce statuses `[429, 429, 503]`: the first two responses mark the worker unhealthy, and the router generates the third response because no worker remains routable. In a pool this shifts more traffic to the survivors and can cascade through every worker during overload

we can remove `429` from eviction and either remove overload-shaped `503` responses as well or document and enforce a narrower status invariant. Keep them in request statistics, and let `/health` probes determine liveness.

sglang_omni_router/serve.py
```diff
@@ -32,6 +32,40 @@
...
+        return
+    required = 2 * config.upstream_pool_size + _NOFILE_HEADROOM
+    if soft_limit >= required:
+        return
+    message = (
+        f"nofile soft limit {soft_limit} is below {required} "
+        f"(2 x upstream_pool_size={config.upstream_pool_size} + "
+        f"{_NOFILE_HEADROOM} headroom); under load the relay exhausts file "
+        f"descriptors and clients see raw connection errors. Raise the limit "
+        f"(ulimit -n {required}) or lower --max-connections."
```
Ratish1: Make this remediation name the option that actually drives `upstream_pool_size`. For the supported configuration `--max-connections 64 --max-inflight 800`, the warning computes the limit from a pool size of `800` but tells the operator to lower `--max-connections`. Lowering that flag cannot change `max(64, 800)`, so the warning and `--strict-limits` failure remain

Include both resolved values and recommend lowering `--max-inflight` when it is the maximum. Update the `--strict-limits` help and router documentation to use the same resolved-pool terminology.

sglang_omni_router/proxy.py
```diff
@@ -115,8 +172,57 @@ def __init__(
...
+                headers={"Retry-After": _OVERLOAD_RETRY_AFTER_SECS},
+            )
+        release = _ReleaseOnce(self._admission)
+        try:
+            response = await self._forward_admitted(request, path, release)
+        except BaseException:
+            release()
+            raise
+        # The streaming relay hands its slot to the response iterator; every
+        # other path is done with it as soon as the response object exists.
```
Ratish1: can remove this comment  or add `note(Jiaxin):` here

sglang_omni_router/proxy.py
JiaxinD: Confirmed. Added an idempotent async cleanup owner `_RelayCleanup` (upstream close, worker decrement, admission release, exactly once), invoked from both the iterator `finally` and `_RelayResponse.__call__`, so the path where `send` raises on `http.response.start` releases all three (previously only admission had the GC backstop).

New failing-before ASGI regression `test_streaming_response_send_failure_releases_all_resources_once` raises on `http.response.start` and asserts each of the three releases exactly once. Pushed in `afa1696a`.

sglang_omni_router/proxy.py
JiaxinD: Confirmed. Narrowed `WORKER_EVICTION_STATUS_CODES` to gateway failures `{502, 504}`; 429/503 (and 408) no longer feed the liveness circuit, and liveness stays with `/health` probes. They are still counted in request statistics, and transport failures still evict. I also dropped 408 so the invariant is one line: a relayed status evicts only on gateway failure (502/504); backpressure and application statuses do not.

Test: `test_relayed_status_evicts_only_on_gateway_failure` (parametrized; 429/503/408 failing before, 500/502/504 as regression guards). `Failure Handling` docs updated. Pushed in `afa1696a`.

sglang_omni_router/serve.py
JiaxinD: Confirmed. The warning now prints both resolved values and recommends the flag that actually drives `max(max_connections, max_inflight)`: for `--max-connections 64 --max-inflight 800` it now recommends lowering `--max-inflight`. `--strict-limits` help and the router docs use the same resolved-pool terminology.

Tests updated to assert the recommended flag: `test_nofile_check_follows_the_upstream_pool_size` (recommends `--max-inflight`) and `test_nofile_check_warns_when_soft_limit_too_low` (recommends `--max-connections`), both failing before. Pushed in `afa1696a`.

sglang_omni_router/proxy.py
JiaxinD: Done, rewritten in `note (Jiaxin)` form (it now describes the slot-ownership handoff for the refactored cleanup owner). `afa1696a`

sglang_omni_router/serve.py
```diff
@@ -32,6 +32,50 @@
...
+    required = 2 * pool_size + _NOFILE_HEADROOM
+    if soft_limit >= required:
+        return
+    # Note (Jiaxin Deng): the pool is max(max_connections, max_inflight); name the
+    # flag that actually drives it, otherwise lowering the smaller one cannot
+    # change the resolved pool size and the warning never clears.
+    max_connections = config.max_connections
+    max_inflight = config.effective_max_inflight
+    remediation_flag = (
+        "--max-inflight" if max_inflight > max_connections else "--max-connections"
```
Ratish1: it seem we should also handle an explicitly configured tie here?. With `--max-connections 512 --max-inflight 512`, both values bind `max(...)`, but this warning says to lower only `--max-connections`. Doing that leaves `--max-inflight=512`, so the pool remains 512 and the warning does not clear.

The default tie is different: when `max_inflight is None`, lowering `--max-connections` also lowers the derived admission bound and is correct. For explicit equal values, recommend lowering both flags (or raising `nofile`), and possibly add a unit test for the `512/512` case.

Ratish1 (approved):
LGTM with one small comment

sglang_omni_router/serve.py
JiaxinD: Good catch, fixed. An explicit `--max-connections == --max-inflight` tie now recommends lowering both (lowering one alone leaves the other holding the pool); the derived tie (`--max-inflight` unset) still recommends `--max-connections`, since that also lowers the derived admission bound. Added regression tests for both tie cases (`test_nofile_check_explicit_tie_recommends_both_flags`, `test_nofile_check_derived_tie_recommends_max_connections`). `37e3fb0a`

PR #1016, opened 2026-07-09
[Ming-Omni] talker duration guard fix

AkazaAkane:
## Motivation

The Ming-Omni talker duration guard previously checked accumulated audio length by repeatedly calling `torch.cat(all_wavs, dim=-1).shape[...]` inside the streaming loop.

That is unnecessarily expensive: every duration check materializes a new concatenated waveform tensor and recopies all chunks generated so far. For multi-chunk generations, this makes the guard O(n²) in accumulated audio size even though the guard only needs the current sample count.

This PR replaces those repeated concat-based length checks with a lightweight running sample counter.

## Modifications

* Add `total_samples` inside `MingOmniTalker._process_segment`.
* Increment `total_samples` whenever a generated `tts_speech` chunk is appended/yielded.
* Replace per-chunk `torch.cat(all_wavs, dim=-1).shape[...]` duration checks with `total_samples / sample_rate`.
* Preserve the existing guard placement and duration-limit semantics.
* Keep chunk yielding behavior unchanged.
* Add `_run_process_segment_with_chunk_lengths` test helper for deterministic duration-guard tests.
* Add unit tests for:

  * single chunk below the guard
  * multi-chunk generation below the guard
  * stopping after the chunk that crosses the duration guard
  * strict two-second boundary behavior
  * empty generation
  * ensuring the duration guard does not call `torch.cat` per chunk

## Related Issues

Refs #999.

## Accuracy Test

all tests passed

## Benchmark & Profiling

No full model benchmark was run for this PR.

Expected performance impact is localized to long / multi-chunk Ming talker generations:

* before: duration guard repeatedly concatenated all prior waveform chunks
* after: duration guard uses an O(1) running sample counter per chunk
* final audio assembly behavior is unchanged

The regression test monkeypatches `torch.cat` to raise and confirms the duration guard no longer depends on per-chunk concatenation.

## Checklist

* [ ] Format your code according with pre-commit.
* [x] Add unit tests.
* [ ] Update documentation / docstrings / example tutorials as needed.

## CI


tests/unit_test/ming_omni/test_talker_voice_validation.py
```diff
@@ -288,6 +288,42 @@ def _fake_detokenizer(sample_rate=44100, vae_patch_size=4, hop_size=480):
...
+            talker,
+            text,
+            prompt=None,
+            instruction=None,
+            spk_emb=None,
+            audio_detokenizer=detok,
+            prompt_text=None,
+            prompt_wav_lat=None,
+            prompt_wav_emb=None,
+            stream=True,
```
edwingao28: Could we add a stream=False case here? The duration guard has separate streaming and aggregate paths, but these tests only cover the streaming one.

sglang_omni/models/ming_omni/talker/modeling_ming_omni_talker.py
```diff
@@ -1273,6 +1273,7 @@ def _process_segment(
 
             use_stream = stream
             all_wavs: list = []
+            total_samples = 0
```
edwingao28, line 1275: all_wavs is no longer read now that the guard uses total_samples. Could we remove it and the two appends? Keeping it retains every emitted chunk until _process_segment exits.

AkazaAkane: Removed unused all_wavs retention and appends

tests/unit_test/ming_omni/test_talker_voice_validation.py
AkazaAkane: Added a stream=False duration-guard test case by parameterizing the helper

edwingao28 (approved):

edwingao28:
lgtm and merge this pr as this is minor fix for ming only and passed all ci

Issue #985, opened 2026-07-07
[TTS Refactor] Roadmap and PR tracker


luojiaxuan:
This is the TTS-specific roadmap tracker for the refactor effort under #661.

New TTS refactor PRs should use the title prefix `[TTS Refactor]` and link this issue in the PR body, normally as `Part of #985` or under `Related Issues`. Keep model math and semantics in model directories; move reusable scheduling, caching, engine construction, capability metadata, and vocoder mechanics into shared framework surfaces.

## Source docs

- RFC / design doc: https://osgbw74w8zwb.sg.larksuite.com/docx/L3XId4GHYoJqSKxwXYqlXaBrgNd
- Parent epic: #661
- ReferenceEncodeService sub-tracker: #925

### Landed

- [x] T1 PipelineStateBase + all six TTS migrations and quick wins — #807
- [x] M2 TTS engine factory: shared TTS engine builder — #923
- [x] M4a ReferenceEncodeService: FishAudio S2-Pro migration — #926
- [x] M4a ReferenceEncodeService: MOSS-local cache cleanup — #930
- [x] M8 Model capabilities: static TTS model capability metadata — #957
- [x] M4a ReferenceEncodeService: Qwen3-TTS taxonomy migration — #927
- [x] T5a BatchVocoderBase: non-streaming vocoder base — #940
- [x] M6 StreamingVocoderBase: base + MOSS-TTS-Local migration — #936
- [x] M6 StreamingVocoderBase: Higgs migration — #939
- [x] M9 Fish -> OmniScheduler: migrate AR stage and delete FishScheduler — #937
- [x] M10 Unit-test pruning slice 1: migrated-surface duplicates (moss / moss-local / qwen3-tts) — #1019
- [x] Dead-code sweep: remove unreachable implementation code — #1069
- [x] Declarative TTS pipeline-state serialization — #1050
- [x] Ming-Omni-TTS integration: adopt M2/M4a/M5/M8 shared TTS surfaces — #858
- [x] Ming-Omni-TTS shared-surface follow-up: declarative state serialization, typed-tensor float transport, checkpoint/config dedup (plus Qwen3-TTS/Voxtral checkpoint-resolution cleanup) — #1103

### Active

- [x] ZONOS2 integration on shared TTS surfaces: keyed tensor reference-encode hook, shared store_state, shared state contract test (supersedes #779) — #1112

- [x] TTS engine-builder forwarding cleanup — #1068
[...]

PR #939, opened 2026-07-03
[TTS Refactor][T5b] StreamingVocoderBase: Higgs migration

GaokaiZhang:
Stacked on #936 (this diff includes its commits until it merges) — review scope here is the Higgs commit only.

Second T5b migration from #661, on top of #936's `StreamingVocoderBase`: Higgs moves onto the base with no base changes — the hook surface validated on MOSS-local absorbed the simpler request-local streamer as-is. `StreamingVocoderBase` is the streaming sibling of T5a's `BatchVocoderBase` (`scheduling/vocoder_base.py`, merged via #940); this PR touches only the streaming path.

One file of production code changes: `higgs_tts/vocoder_scheduler.py` (−241/+147). The duplicated lifecycle (state registry, stream gate, chunk scaffold, done sequencing, nothing-emitted fallback, abort/late-chunk handling) is deleted; delay-pattern cursor math (`raw = rows − num_codebooks + 1`), holdback/overlap re-decode with exact hop-length trimming, contract latching content, the metadata-only terminal result and the non-streaming batch-decode path stay model-owned and verbatim (the latter is T5a territory: it migrates onto `BatchVocoderBase` separately, not here). No coalescing. `stages.py` wiring unchanged (same constructor surface). Client-visible strings preserved via `stream_source_hint="Higgs"`, with one theoretical exception (None-metadata chunk after a payload-latched contract now errors; unreachable from the only producer — `_emit_code_chunk` always attaches the metadata dict).

Behavior class: Green (output-equivalent). Old path deleted in this PR; no dual path.

Tests: all 34 pre-existing higgs pipeline tests pass unmodified (none deleted — each covers model-owned math, not base scaffold); added latch-immutability cases that previously had no model-level coverage. Base contract suite unchanged.

Full SeedTTS-EN gate (complete en split, c=16, DP1, no router, 1x H100, `--max-running-requests 64 --cuda-graph-max-bs 64`), bosonai/higgs-audio-v3-tts-4b, main @ 1cb04ad vs this branch:

| Model / config | Split | WER / CER | Speaker SIM | Samples | Failed / skipped | Throughput | rtf_mean |
|---|---|---|---|---|---|---|---|
| main, non-stream | en | 0.95% | 67.15 | 1088/1088 | 0 / 0 | 14.93 req/s | 0.256 |
[...]

PR #1019, opened 2026-07-11
[TTS Refactor] Prune migrated-surface duplicate unit tests (moss / moss-local / qwen3-tts)

luojiaxuan:
## What

First slice of the **unit-test pruning** follow-up from #661 ("Comment cleanup and unit-test pruning follow-ups — TBD" in the #985 workstream checklist; the comment-cleanup half closed via #852–#856). Deletes 8 per-model tests in TTS dirs whose behavior is owned by shared bases that already landed with the refactor and is contract-tested there. **-358 / +9 test lines; no non-test lines touched.**

Behavior class: test-only, output-equivalent (no runtime code changed). Affected models: moss_tts, moss_tts_local, qwen3_tts (test dirs only).

## Deletions and the coverage that gates each one

**State round-trips → T1 contract** (`scheduling/test_pipeline_state.py::test_tts_pipeline_state_round_trips_preserve_payload_fields` exercises all six TTS states with strictly richer payloads, incl. tensor fields; its normalizer also catches tensor-vs-list type regressions):
- `moss_tts_local::test_state_round_trip`
- `qwen3_tts::test_qwen3_tts_state_round_trip_preserves_request_fields`
- `moss_tts::test_moss_tts_state_round_trip_keeps_tensors_native`

**Reference-cache mechanics → M4a contract** (post-#930, `_MossLocalReferenceEncoder` is a thin adapter over the shared `ReferenceEncodeService`; the mechanics live in the service and are contract-tested in `scheduling/test_reference_encoder.py`):
- `moss_tts_local::test_cached_reference_encoder_lru_eviction` → `test_lru_eviction_respects_max_bytes` + `test_artifact_larger_than_budget_is_returned_but_not_cached`
- `moss_tts_local::test_cached_reference_encoder_true_concurrency_dedup` → `test_same_key_concurrent_single_flight`; the per-caller clone-isolation assertion stays covered by the kept `test_cached_reference_encoder_return_value_isolation`
- `moss_tts_local::test_cached_reference_encoder_failure_does_not_poison` → `test_exception_propagates_to_all_waiters_and_does_not_poison` (+ revalidate variants)
- `qwen3_tts::test_qwen3_tts_adhoc_reference_failure_does_not_poison` → same no-poison service contract; qwen3's taxonomy wiring stays covered by the kept `test_qwen3_tts_adhoc_voice_clone_prompt_uses_reference_service`
- `moss_tts_local::test_cached_reference_encoder_rejects_nonpositive_capacity` → **folded into the contract**: capacity validation lives in `ReferenceEncodeService.__init__`, so this PR adds `test_constructor_rejects_nonpositive_capacity` to `scheduling/test_reference_encoder.py` (+9) and drops the per-model copy

## Deliberately kept

- moss-local hook-owned behavior: on/off bit-identity, return-value isolation, duration gates, data-URI keyspaces
- all `_BatchedReferenceEncoder` tests (M4b batching is still model-owned, deferred pending profiling)
- per-model vocoder tests (they are the only T5a coverage until #875 lands the shared `test_batch_vocoder.py`)
- config/stage-wiring contract tests (deployment contracts, not mock theater)

## Out of scope (avoid conflicts with in-flight PRs)

- `higgs_tts/test_pipeline.py` — touched by #939
- `fishaudio_s2_pro/*` tests — touched by #937
- `test_streaming_vocoder.py` / `test_state_pool.py` — fold into T5b/T4 contracts per #661
- `voxtral_tts` — T1 migration still TBD, nothing to prune against
- next slices: qwen3_omni (~7.0k) / ming_omni (~5.6k) mock tests and framework dirs, per #661

## Verification

- `pre-commit` on changed files: all hooks pass (autoflake finds no orphaned imports; deleted tests used function-local imports)
- `scheduling/test_reference_encoder.py` + `scheduling/test_pipeline_state.py`: 22 passed locally
- `moss_tts_local/test_pipeline.py`: identical pass/fail set to unmodified main in the same env (the 5 env failures are missing-sglang/torch-version artifacts, present on baseline too) — the deletions change nothing else
- Gate: `PR Test / unit-test` runs the full suite with the pinned env

Part of #985. Refs #661.


SandyLuXY:
@luojiaxuan The changes look good overall.

One thing to note is that the three state round-trip tests are not fully replaced by `test_tts_pipeline_state_round_trips_preserve_payload_fields`: its helper at `tests/unit_test/scheduling/test_pipeline_state.py:133` compares `state.to_dict()` with `restored.to_dict()`, so a field omitted by `to_dict()` is missing on both sides and the test still passes. I verified this by dropping `assistant_start_length`: a value of 2 restored as 0 without failing the shared test. Please make the shared test assert the original populated fields, with the expected tensor/list behavior for each model, before deleting these tests.

luojiaxuan:
Good catch, thanks — fixed in 790843b.

You're right that the dict-vs-dict comparison in `_assert_round_trip_preserves_payload` can't see a field that `to_dict()` silently drops: the missing field is absent from *both* the "before" and "after" snapshots, so nothing diffs. Added `_assert_restored_fields`, which asserts each populated field directly against the **restored object's attributes** (not a re-serialized dict), using explicit expected values that encode each model's actual tensor-vs-list round-trip behavior:

- Qwen3-TTS flattens `audio_codes`/`audio_samples` to lists in `to_dict()` and never rebuilds them in `from_dict()` — expected is a list
- Voxtral round-trips `audio_codes` as a tensor via `encode_typed_tensor`/`decode_typed_tensor` but flattens `audio_samples` to a list — mixed per-field
- MOSS / MOSS-Local / S2-Pro / Higgs round-trip their tensor fields natively via `serialize_value` or plain passthrough — expected stays a tensor

I verified the new assertion actually catches the bug class you found: temporarily dropped the `assistant_start_length` line from `MossTTSState.to_dict()` locally and reran — it fails with `AssertionError: MossTTSState.assistant_start_length: 0 != 2`, then reverted before pushing. All 22 shared-scheduling tests plus the 3 pruned model test files pass locally against this env.

tests/unit_test/scheduling/test_pipeline_state.py
```diff
@@ -147,102 +161,219 @@ def test_tts_pipeline_state_round_trips_preserve_payload_fields() -> None:
...
+                "ref_text": "ref",
+                "language": "en",
+                "instructions": "bright",
+                "token_count": 5,
+                "generation_kwargs": {"top_p": 0.8},
+                "audio_codes": torch.tensor([[1, 2, 3], [4, 5, 6]]),
+                "prompt_tokens": 5,
+                "completion_tokens": 7,
+                "engine_time_s": 0.5,
+            },
```
SandyLuXY: The restored-attribute check now catches the `assistant_start_length` omission, but it still does not replace MOSS-Local's deleted sample-rate assertion: this expected mapping omits `sample_rate`. I changed `MossTTSLocalState.to_dict()` to emit 24000 and the shared test still passed while the restored state changed from 48000 to 24000; the deleted model test would fail that regression. Please keep explicit restored sample-rate coverage here, or make the helper cover every serialized/restored field with overrides only for the intentional tensor-to-list conversions.

luojiaxuan: Fixed in 34d734e — took your second option: `_assert_restored_fields` now iterates every dataclass field of the state and asserts the restored attribute equals the originally constructed value, with `overrides` naming only the intentional tensor-to-list conversions (Qwen3-TTS `audio_codes`/`audio_samples`, Voxtral `audio_samples`). Omitting a field is structurally impossible now — there is no hand-written expected mapping left to forget. Verified with your exact mutation: hardcoding `MossTTSLocalState.to_dict()` sample_rate to 24000 fails with `MossTTSLocalState.sample_rate: 24000 != 48000`, then reverted. Also -67 net lines since the per-model expected dicts collapsed into small override maps.

SandyLuXY (approved):

PR #1020, opened 2026-07-11
Add Higgs-Audio-v3-STT (bosonai/higgs-audio-v3-stt) ASR pipeline

zhudianGG:
Adds `bosonai/higgs-audio-v3-stt` as a single-stage batched ASR pipeline following the `qwen3_asr` pattern: Whisper-large-v3-style audio tower (+AvgPool1d(2)) and depthwise-conv/MLP projector feeding sglang's `Qwen3ForCausalLM`; fixed 4s audio chunks (12.5 embeddings/s) spliced as one contiguous `<|AUDIO|>` span via `general_mm_embed_routine`.

Implementation notes:
- Native `HiggsAudio3Config` registered under `model_type=higgs_audio_3` — the checkpoint's remote code targets the transformers-4 layer API (`encoder_layer(...)[0]` strips the batch dim on transformers 5), so tower/projector are implemented natively and `trust_remote_code` is not needed.
- The checkpoint's `text_config` claims `tie_word_embeddings=true` but ships a distinct `audio_decoder_proj.text_lm_head` (weights differ from `embed_tokens`); the config forces untied and `load_weights` maps the explicit head.
- Mel features from the whisper-large-v3 processor (checkpoint ships no `preprocessor_config.json`), padded to the longest chunk.

Validation (RTX 5090, LibriSpeech test-clean, n=50, concurrency 4): **WER 1.64%, 0 errors** — identical to the reference implementation — at 25.8 req/s, E2E p50 0.135s. Unit tests included (`tests/unit_test/higgs_audio_asr/`).


tests/unit_test/higgs_audio_asr/test_pipeline.py
Ratish1, line 1: can we update `tests/README.md`

Ratish1:
Hey @zhudianGG, thanks for the contribution!. Could you add my wechat Id Ratish_P and we can further discuss about this PR in detail.

sglang_omni/models/higgs_audio_asr/audio_tower.py
```diff
@@ -0,0 +1,93 @@
...
+``WhisperEncoderLayer`` stack) followed by ``AvgPool1d(2)`` over time and
+a final LayerNorm — 25 embeddings/s. Projector = depthwise
+``Conv1d(stride 2)`` temporal downsample + 2-layer ReLU MLP into LLM
+space — 12.5 embeddings/s. Parameter paths mirror the checkpoint's
+``audio_tower.*`` / ``audio_encoder_proj.*`` so weights load by name.
+
+Reimplemented (rather than running the checkpoint's remote code) because
+the shipped modeling code targets the transformers-4 layer API:
+``encoder_layer(...)[0]`` strips the batch dim on transformers 5.
+"""
```
Ratish1, line 7: For comments and docstrings across the code, could you follow this rule https://github.com/zhaochenyang20/sglang-diffusion-routing/issues/32#issuecomment-4918483294

sglang_omni/models/higgs_audio_asr/audio_tower.py
```diff
@@ -0,0 +1,93 @@
...
+        return self.conv1.weight.dtype
+
+    def forward(self, input_features: torch.Tensor) -> torch.Tensor:
+        """(batch, num_mel_bins, T_mel) -> (batch, ~T_mel/4, d_model)."""
+        hidden = F.gelu(self.conv1(input_features))
+        hidden = F.gelu(self.conv2(hidden))
+        hidden = hidden.permute(0, 2, 1)  # (B, T, D)
+        hidden = hidden + self.embed_positions.weight[: hidden.shape[1]]
+        for layer in self.layers:
+            hidden = layer(hidden, None)
```
Ratish1: can we pass the valid-frame attention mask into every Whisper encoder layer. This currently sends `None`, so a partial final chunk attends to the right-padding added to match the preceding 4-second chunks. Because Whisper attention is bidirectional, slicing the projected output afterward does not recover the correct valid embeddings.

we could build the post-`conv2` 4-D mask from `feature_attention_mask` and  pass it to each layer

sglang_omni/models/higgs_audio_asr/sglang_model.py
```diff
@@ -0,0 +1,174 @@
...
+        return self.pattern.pad_input_tokens(input_ids, mm_inputs)
+
+    def get_audio_feature(self, items: List[MultimodalDataItem]) -> torch.Tensor:
+        """Batch-encode padded 4 s chunks, project, slice each chunk to its
+        valid (downsampled) length, concatenate in order."""
+        device = next(self.audio_tower.parameters()).device
+        dtype = self.audio_tower.dtype
+
+        embeds: list[torch.Tensor] = []
+        for item in items:
```
Ratish1, line 76: batch these items before invoking the audio tower. `items` contains the scheduled prefill batch, but this loop runs the complete 32-layer Whisper encoder once per request, serializing the most expensive encoder work and defeating continuous batching.

pad and concatenate features/masks across the scheduled batch, run one masked tower/projector call per batch or explicit length bucket, and split the projected embeddings by the retained valid lengths. Remove the GPU-to-CPU length round trip from this path as part of the same change.

sglang_omni/models/higgs_audio_asr/request_builders.py
```diff
@@ -0,0 +1,249 @@
...
+_AUDIO_EOS = "<|audio_eos|>"
+
+# The reference transcribe.py prompt (lowercase, no punctuation output).
+DEFAULT_TRANSCRIBE_PROMPT = (
+    "Transcribe the speech. Output only the spoken words in lowercase "
+    "with no punctuation."
+)
+# Empty think block: the transcript starts immediately (enable_thinking=False
+# in the reference pipeline).
+_THINK_SUFFIX = "<think>\n\n</think>\n\n"
```
Ratish1: should we not hardcode no-thinking mode as the reference behavior. The pinned checkpoint's `transcribe.py`, model-card example, default to `enable_thinking=True`, while this suffix permanently disables it and the request API offers no way to select the documented mode.

can we expose an explicit thinking policy, default to the documented behavior unless there is evidence that justifies a divergence, and handle complete and incomplete think blocks at token level in the result adapter.

sglang_omni/models/higgs_audio_asr/request_builders.py
```diff
@@ -0,0 +1,249 @@
...
+]:
+    if feature_extractor is None:
+        raise ValueError("Higgs-Audio-ASR is missing a feature_extractor")
+
+    audio_pad_token_id = int(tokenizer.convert_tokens_to_ids(_AUDIO_PAD))
+    stop_token_ids = [
+        int(tokenizer.convert_tokens_to_ids("<|im_end|>")),
+        int(tokenizer.convert_tokens_to_ids("<|endoftext|>")),
+    ]
+    vocab_size = int(tokenizer.vocab_size)
```
Ratish1: use the language model vocabulary size here, not `tokenizer.vocab_size`. For this checkpoint, the tokenizer reports a base vocabulary of `151643`, while the model output dimension is `151936`; generated added tokens such as `<think>` (`151667`) are therefore classified by SGLang as an invalid/NaN token.

we can pass `config.text_config.vocab_size` or `model_config.vocab_size` into the adapter

Ratish1 (commented):
Thanks for the PR @zhudianGG , left some review comments. Lets also fix the lint.

tests/unit_test/higgs_audio_asr/test_pipeline.py
zhudianGG: Ok, I will update that

sglang_omni/models/higgs_audio_asr/audio_tower.py
zhudianGG: Encoder attention mask (audio_tower.py): the tower now takes per-sample valid mel lengths and passes a key-padding bias to every Whisper encoder layer, built from the post-conv2 valid length. With bidirectional attention the previous None let a partial final chunk attend to the shared right-padding and slicing afterward couldn't recover it. Added test_tower_masking_isolates_partial_chunk: batches a padded partial chunk with a longer one and asserts its embeddings equal running it alone.

sglang_omni/models/higgs_audio_asr/sglang_model.py
zhudianGG: get_audio_feature now pads+concatenates all scheduled requests' chunks into a single masked tower/projector call (one 32-layer pass for the batch instead of one per request), then splits the projected output by retained valid lengths. Removed the per-item GPU→CPU length round trip — a single host transfer of the final small length vector remains (needed for the variable-length split).

sglang_omni/models/higgs_audio_asr/request_builders.py
zhudianGG: no longer hardcodes the empty-think suffix. enable_thinking is an explicit policy defaulting to the checkpoint's documented behavior (True, per transcribe.py / model card), overridable per request. The result adapter resolves both complete </think> and truncated-open <think> blocks at token level (matching the reference _parse_output).

sglang_omni/models/higgs_audio_asr/request_builders.py
zhudianGG: same as prior reply

sglang_omni/models/higgs_audio_asr/audio_tower.py
zhudianGG: reply as next comment

zhudianGG:
Thanks for the thorough review @Ratish1! Pushed e64e036 (rebased on latest main) addressing all points. Lint is green locally (isort/autoflake/black).

PR #1034, opened 2026-07-12
[ASR] Duration-aware output budget, context clamp, and 400 mapping for long audio transcription

db-ol:
## Motivation

The MOSS-TD checkpoint raised max_position_embeddings from 40960 to 131072 on Jul 6, which makes 90 minute audio representable in one context. The transcription request path still predates that change and fails badly on long audio:

1. The default max_new_tokens of 5120 silently truncates any transcript longer than about 20 minutes of dense speech. A 38.7 minute AISHELL4 meeting returns HTTP 200 with 6782 of 18400 characters and finish_reason length. A 90 minute request with default params would lose most of the transcript.
2. max_new_tokens is not clamped to the remaining context. A 38.7 minute input with max_new_tokens=999999 is rejected with HTTP 500 even though the model can serve it with a smaller output budget. The long audio CI already sends 65536, which a 90 minute input pushes past the context limit.
3. The transcription endpoint maps every engine error to HTTP 500. Chat and speech endpoints already return 400 for client-fixable errors, transcription does not. Audio past the context limit (about 164 minutes) is a client error and should say so.
4. On a 76.7 minute run, 1 of 738 output segments carried a corrupted end timestamp (4809.8s in a 4601.3s file), which leaks into verbose_json as is.

## Modifications

* request_builders.py: when the request does not set max_new_tokens and the operator did not configure a fixed default, the output budget now scales with audio duration (10 tokens per audio second, about twice the observed dense meeting rate). Explicit request values and operator-configured defaults still win, and an explicit value below 1 is rejected with a message that maps to 400. Requests are built with enforce_request_limits=True (as the TTS builders already do), so the scheduler clamps max_new_tokens to the remaining context and rejects prompts that do not fit at all, through the upstream validation path. The processor limit follows the served context length instead of a hardcoded 131072, and a request supplied max_length is capped at that limit, so prompts past the context are rejected early with a clear message rather than deferring to a later scheduler rejection. When decoded audio is empty the duration scaled budget falls back to the fixed default with a warning.
* openai_api.py: create_transcription maps client-fixable errors to HTTP 400 with the same marker mechanism the chat endpoint uses. Two new markers cover the upstream input length validation and the hub processor max_length rejection. For stream=true the route now pulls the first chunk before sending response headers, so admission rejections also map to 400 instead of degrading into an SSE error frame under HTTP 200. Errors after the first chunk still arrive as SSE payloads. Expected client errors no longer log an ERROR level stack trace.
* transcription_adapters/moss_transcribe_diarize.py: verbose_json segments with timestamps outside the audio duration (plus one 5s marker interval of tolerance) are clamped, with a warning log counting repairs.
* stages.py: passes the served context length and the operator intent for max_new_tokens into the adapters. A context_length entry in server_args_overrides now reaches ServerArgs instead of colliding with the explicit keyword, so the served context can be overridden per deployment.
* docs/cookbook/moss_transcribe_diarize.md: the request parameter table and the long audio example now describe the duration scaled default, the fixed operator value, explicit request precedence, and the scheduler clamp.

## Related Issues

Part of the long audio milestone (M-PR7) in #924. Builds on the context default from #982.

## Accuracy Test

Behavior A/B on one H100, current main (14d1d584) vs this branch, MOSS-Transcribe-Diarize at the current hub revision (d7231bb). Audio is AISHELL4 validation content.

| request | main | this branch |
|---|---|---|
| 38.7 min, default params | 200, transcript cut at 6782 chars | 200, full 18400 chars, cer_no_spk 17.351 |
| 38.7 min, max_new_tokens=999999 | 500 in 2s (KV capacity rejection) | 200, clamped, full transcript |
| 90.0 min, max_new_tokens=999999 | 500 (same rejection) | 200 in 163s, 792 segments covering the full 90 min |
| 192 min (over context) | 500, Prompt/audio sequence exceeds max_length | 400, same message correctly classified |

The 38.7 minute transcript on this branch is character-identical to a pre-change control run made with an explicit max_new_tokens=30000, so on this sample the clamp and budget changes did not alter decoded content.

The 90 minute request doubles as a long context probe: the same 38.7 minutes of content was scored alone and again while sitting at token positions 41k to 72k of the 90 minute context. cer_no_spk is 17.06 at depth vs 17.35 alone, so this probe shows no positional degradation on this sample up to the 90 minute mark. Corpus level validation across 45 to 90 minutes is separate follow-up work tracked in #924. Peak KV usage for the 90 minute request was 102367 tokens (10.9 GiB at bf16). Zero corrupted timestamps in this run, and the sanitizer covers the rare case with a unit test reproducing the observed 4809.8s outlier.

## Benchmark & Profiling

Request path only, no model or scheduler hot path changes. No throughput impact expected. Request timings are in the table above.

## Checklist

- [x] Add unit tests.


sglang_omni/serve/openai_api.py
```diff
@@ -127,6 +127,8 @@
     "Requested token count exceeds the model's maximum context length",
     "accepts audio up to",
     "max_new_tokens must be",
+    "exceeds the maximum allowed length",
```
SandyLuXY, line 132: These markers only affect the non-streaming `client.completion()` catch. With `stream=true`, the route returns `StreamingResponse` before `client.generate()` runs, so the identical over-context error is logged by `_transcription_stream` and returned as an SSE error under HTTP 200. Please validate admission before response headers are sent, map this case to 400.

sglang_omni/models/moss_transcribe_diarize/stages.py
```diff
@@ -195,6 +195,11 @@ def create_sglang_moss_transcribe_diarize_executor(
         processor=processor,
         tokenizer=tokenizer,
         max_new_tokens=resolved_max_new_tokens,
+        # note (db-ol): server_args wins over the resolved value when a
+        # server_args_overrides entry changes context_length, and the builder
+        # limit must match what the scheduler actually enforces.
+        context_length=int(server_args.context_length),
```
SandyLuXY, line 210: `build_generation_batch_overrides()` retains `context_length`, while the call above also supplies `context_length=resolved_context_length`, so `server_args_overrides={"context_length": 8192}` raises `TypeError: got multiple values for keyword argument 'context_length'` 🤔️

sglang_omni/models/moss_transcribe_diarize/request_builders.py
```diff
@@ -289,7 +304,19 @@ def request_builder(payload: StagePayload) -> MossTranscribeDiarizeRequestData:
...
         top_p = _sampling_param(params, explicit_fields, "top_p", DEFAULT_TOP_P, float)
         top_k = _sampling_param(params, explicit_fields, "top_k", DEFAULT_TOP_K, int)
-        request_max_new_tokens = int(params.get("max_new_tokens") or max_new_tokens)
+        # note (db-ol): the model default was sized for short clips and
+        # silently cuts transcripts past about 20 minutes. Scale the default
+        # budget with duration unless the operator configured a fixed one.
+        requested_max_new_tokens = params.get("max_new_tokens")
+        if requested_max_new_tokens:
+            request_max_new_tokens = int(requested_max_new_tokens)
+        elif duration_scaled_default:
```
SandyLuXY, line 321: This changes the public omitted-`max_new_tokens` default, but the mtd cookbook still documents a fixed `5120` default and instructs long-audio users to set `65536`. Please update the request-parameter table and long-audio example to explain the duration-scaled default, operator-fixed defaults, explicit request precedence, and scheduler clamping.

SandyLuXY (changes requested):

sglang_omni/serve/openai_api.py
db-ol: Good catch. Thanks! fixed. The stream branch now pulls the first chunk before returning the response, the same way the speech PCM route does, and reuses the blocking path's error mapping, so admission failures return 400 before headers go out. Errors after the first chunk still arrive as SSE payloads since the status is already sent by then. Added streaming tests for the 400, 500, and happy path cases. Chat streaming has the same gap, I kept it out of this PR.

sglang_omni/models/moss_transcribe_diarize/stages.py
db-ol: You're right. Fixed by popping context_length out of the merged overrides before the build_sglang_server_args call, so an override now wins over the model config default. The regression test keeps the real override merge and asserts an 8192 override reaches ServerArgs and the request builder. The other executors have the same latent collision, but their context lengths derive from encoder token counts, so I left the precedence decision there to a separate cleanup.

sglang_omni/models/moss_transcribe_diarize/request_builders.py
db-ol: Thanks. Updated. I kept the explicit 65536 in the examples and benchmark commands since explicit values keep runs reproducible across default changes.

db-ol:
Thanks very much for your review, @SandyLuXY ! I've addressed your comments. Could you please take another look when you have a moment?

sglang_omni/serve/openai_api.py
```diff
@@ -1629,10 +1632,34 @@ async def create_transcription(
...
             adapter = resolve_adapter(getattr(app.state, "architectures", None))
             duration_s = _probe_audio_duration(audio_bytes)
+            chunk_stream = client.generate(gen_req, request_id=request_id)
+            # note (db-ol): pull the first chunk before sending response
+            # headers. Admission rejections such as audio past the context
+            # limit arrive as the first stream event, and once headers go out
+            # the response is locked to 200 and errors can only degrade into
+            # SSE payloads.
+            try:
+                first_chunk = await anext(chunk_stream)
```
SandyLuXY: This first-chunk wait runs before `StreamingResponse` starts its disconnect watcher, so a client that disconnects during long-audio prefill leaves the backend running until the first chunk arrives. Please race `anext(chunk_stream)` against `request.is_disconnected()` as `_speech_audio_response()` does, abort and close the stream when disconnect wins.

db-ol: Thanks. You're right, the peek moved the first chunk wait ahead of the response's disconnect watcher. Fixed. The wait now races against the disconnect poller through the same helpers `_speech_audio_response` uses, aborting and closing the stream when disconnect wins. Added a unit test that a disconnected client triggers the abort and closes the stream.

luojiaxuan (approved):

SandyLuXY (approved):
LGTM

db-ol:
Writing up the CI findings here, as @SandyLuXY suggested. Short version: the only failure that tracks this branch is Qwen3 stage 2, and the data points at its speed thresholds rather than at this change, which touches only the MOSS transcription path.

**Qwen3 stage 2, run 30721116821, attempts 1 to 4.** WER and speaker similarity passed in every attempt. Only the concurrency 16 speed gates fail. Across 9 completed measurements, three gates were never met:

| metric | gate | best | worst |
|---|---|---|---|
| output_tok_per_req_s | >= 7.3 | 7.0 | 5.7 |
| latency_mean_s | <= 2.0 | 2.098 | 2.581 |
| rtf_mean | <= 0.65 | 0.6555 | 0.8376 |
| throughput_qps | >= 7.11 | 7.053 (cleared twice) | 5.873 |

The best values come from attempt 4, where stage 2 ran alone on the runner, so contention does not explain the last 1 to 4 percent. The worst ones line up with host load spikes above 100. My read is that the stage 2 speed thresholds from the #1260 recalibration sit slightly above what the runner sustains today, though I may be missing context from the calibration run.

The other reds along the way were environmental and all cleared on rerun: a container SIGKILL in unit test, an artifact cascade from a cancelled TTS stage 1, a talker_ar startup crash at mem_fraction_static 0.2 (the error reports minimum viable 0.2008), and a stage 8 miss by 2 to 3x during the load spike.

Attempt 5 is running now on a quiet host, everything else is green and stage 2 is the last job in flight. I will post the result here. Happy to help recalibrate stage 2 if that is the fix.

db-ol:
Attempt 5 result: the same speed gates failed in all three tries (throughput_qps 6.603 vs 7.11 in the last two, rtf_mean 0.667 to 0.766 vs 0.65). One caveat, the host load climbed above 140 while the job ran, so this is not a clean idle measurement and attempt 4 remains the cleanest data point. I will fire one more rerun when the runner is actually quiet.

db-ol:
Attempt 6 finished: stage 2 failed the same speed gates again (best of two tries qps 6.426 vs 7.11, rtf 0.7237 vs 0.65), and the host got busy again mid run, so a clean idle window never materialized tonight. The tally for stage 2 on this head is 6 attempts and 14 completed measurements with the three gates never met together, while WER and similarity passed every time. Stage 8 flips with host load (green in attempts 1 to 3 and 5, red in 4 and 6) and looks purely environmental. I will stop rerunning to save runner time and leave the call on the stage 2 thresholds to the CI owners.

PR #1036, opened 2026-07-12
[ASR] Add a 90 minute tier and output format gates to the long audio CI

db-ol:
## Motivation

The long audio CI from #982 validates MOSS-TD up to the 38 minute AISHELL4 clips, while the checkpoint advertises 90 minutes. The 45 to 90 minute range had no coverage, and validating it surfaced a failure mode the current CI cannot see: on one 90 minute concat arrangement the model dropped every timestamp marker from the first output token and stopped decoding at about 70 percent of the audio. A sample whose output loses timestamps is skipped by the DER metric rather than failed, and its cp_cer stays computable from the bare speaker tags, so today's aggregate gates stay green while the output is unusable.

## Modifications

* benchmarks/tasks/transcribe_diarize.py: merge_concat_references joins per clip references into one timeline (timestamps shifted per clip, speaker tags remapped to disjoint ranges, hard error on a reference without parseable segments), and build_long_audio_concat_sample concatenates already staged clips into one long sample, truncating the first clip to hit a target duration. Unit tests cover the merge and its guards.
* tests/test_model/test_asr_ci_multi_speaker.py: a new aishell4_long90 tier runs one 90.0 minute concat (built from the same staged clips the aishell4_long stage uses) through the existing eval path. Gates are request success, DER validity of the output, a missed detection ceiling of 0.5 so partially timestamped output also fails, and a catastrophic cer_no_spk bound of 30. These are deliberately loose bounds rather than calibrated thresholds, since the tier has one sample: the healthy run measures 11.63 cer_no_spk and 0.039 missed detection, the known failure measures 80.3 and loses DER validity entirely. Speaker attribution and speed are printed report-only. The existing 20 clip stage additionally gates speaker_timestamp_der_valid_samples == 20, which is the check that turns the timestamp dropout class from silently skipped into a failure.

## Related Issues

Part of the long audio milestone (M-PR7) in #924. Extends the tier added in #982. The request path fixes this builds on are in #1034. The timestamp dropout itself is a model side issue tracked in #1038, this PR makes it visible to CI.

## Accuracy Test

Offline validation grid on one H100 (current main plus #1034, hub revision d7231bb), six concat arrangements from AISHELL4 validation clips, each constituent clip scored in place against its own reference after removing the concat offset:

| clip | CER at 31 to 36k tokens deep | at 41 to 48k | at 61 to 72k |
|---|---|---|---|
| val_000000 | 8.51 | 7.92 | 8.32 |
| val_000001 | 17.37 | 17.89 | 17.06 |
| val_000002 | 4.97 | n/a | 4.96 |

Per clip CER moves less than 0.6 points across depth, so on this grid transcription quality tracks content, not context position. Five of six arrangements were clean end to end. The sixth (a different filler clip leading the same 90 minutes) reproduced the timestamp dropout deterministically and is the motivating case: replayed through this PR's exact metric path it yields speaker_timestamp_der_valid_samples 0 and trips the new gate, while the current gates pass it.

To confirm the tightened gate cannot break main, all 20 existing aishell4_long clips were rerun through the same metric path on an H100: speaker_timestamp_der_valid_samples 20 of 20, every skip counter 0, and the aggregates (cer_no_spk 13.83, cp_cer 14.97, der 10.42) match the calibrated stage baselines from #982.

## Benchmark & Profiling

Adds one 90 minute request to the ASR CI stage, about 3 minutes of decode on H100 (RTF near 0.03 observed). No serving code changes.

## Checklist

- [x] Add unit tests.


benchmarks/tasks/transcribe_diarize.py
```diff
@@ -143,6 +143,111 @@ def load_movies800_samples(
...
+    for reference, offset_s, limit_s in zip(
+        clip_references, clip_offsets_s, keep_until_s, strict=True
+    ):
+        max_speaker = 0
+        for match in CONCAT_SEGMENT_RE.finditer(reference):
+            start = float(match.group(1))
+            speaker = int(match.group(2))
+            end = float(match.group(4))
+            max_speaker = max(max_speaker, speaker)
+            if limit_s is not None and end > limit_s:
```
SandyLuXY, line 170: This drops a segment whenever its end crosses the cut, but the waveform still keeps the segment's prefix. For the actual 797.88 s filler cut, the `[785.28, 798.63]` segment is removed while 12.60 s remains audible; align the audio cut to a reference boundary and pad to the target duration.

SandyLuXY (changes requested):

benchmarks/tasks/transcribe_diarize.py
db-ol: Thanks. Fixed. The filler is now cut at the last reference boundary within the budget and silence padded back to the target duration, so audio and references stay consistent and the total length is unchanged. Added a unit test pinning the boundary alignment, the silence pad, and the dropped cross cut segment.

tests/test_model/test_asr_ci_multi_speaker.py
```diff
@@ -505,6 +529,31 @@ def test_moss_transcribe_diarize_multi_speaker_datasets(
...
+        samples=[aishell4_long90_sample],
+        outputs=long90_outputs,
+        wall_clock_s=long90_wall_clock_s,
+        repo_id=AISHELL4_REPO_ID,
+        dataset="aishell4_long90",
+    )
+    _print_and_save_results(
+        results=long90_results,
+        tmp_path=tmp_path,
+        filename="moss_transcribe_diarize_aishell4_long90_results.json",
```
SandyLuXY, line 552: This new result filename is absent from both artifact lists in `.github/workflows/test-asr-ci.yaml`, so the current ASR job neither prints nor uploads the 90-minute result JSON. Add the long90 glob to `artifact-path-globs` and `artifact-upload-path` so failures retain their diagnostic evidence.

SandyLuXY (changes requested):
Thanks for fixing the issues! Left one more comment ,and please update `tests/README.md` for the new `aishell4_long90` tier and result artifact.

tests/test_model/test_asr_ci_multi_speaker.py
db-ol: Fixed, added the long90 glob to both artifact lists and documented the tier in tests/README.md.

SandyLuXY (approved):
lgtm

PR #1037, opened 2026-07-12
[MPS-DP]: Enforce equal KV capacity for same-GPU DP

Ratish1:
## Motivation

Related to #921. Sequential replicas in the H100 Higgs same-GPU DP recipe profile KV capacity against progressively less free GPU memory. A shared `max_total_tokens` value is only an upper bound, so a replica that cannot satisfy it can still start with a smaller KV pool. This makes nominally identical replicas differ in serving capacity.

This change is scoped to the validated H100 Higgs recipe. It does not add an H200 configuration or claim a hardware-independent token cap.

## Modifications

- Require a common positive `MAX_TOTAL_TOKENS` value when the same-GPU launcher starts more than one replica, and pass that value to every server process.
- Verify each healthy replica's resolved KV capacity from its startup log and stop the launch if the value is missing or differs from the requested cap.
- Use the model's existing memory-fraction default unless `MF` is explicitly supplied, so the common token cap owns per-replica KV sizing without changing normal Higgs serving defaults.
- Organize the launcher under `examples/mps_dp/` and document the validated H100 Higgs DP3 workflow, including private MPS lifecycle, NUMA-local CPU placement, attachment verification, and run-scoped teardown.

## Speed Tests and Profiling

H100 Higgs DP3 with MPS, three equal 100,000-token KV pools, `max_running_requests=64`, `cuda_graph_max_bs=64`, and non-streaming SeedTTS generation:

| Concurrency per replica | Rep | QPS | Latency p50 | Latency p95 | Latency p99 | RTF p50 | RTF p95 | Audio-s/s | Output tok/s |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 48 | 1 | 39.690 | 2.860s | 4.330s | 5.832s | 0.682 | 0.946 | 171.822 | 4,573.4 |
| 48 | 2 | 39.579 | 2.855s | 4.397s | 5.732s | 0.682 | 0.940 | 171.656 | 4,568.5 |
| 48 | 3 | 39.212 | 2.847s | 4.323s | 5.722s | 0.682 | 0.938 | 170.231 | 4,530.2 |
| 64 | 1 | 44.045 | 3.503s | 5.480s | 6.809s | 0.829 | 1.260 | 187.475 | 4,995.2 |
| 64 | 2 | 44.134 | 3.500s | 5.426s | 6.654s | 0.828 | 1.253 | 187.546 | 4,997.6 |
| 64 | 3 | 44.210 | 3.488s | 5.398s | 6.812s | 0.830 | 1.236 | 188.245 | 5,015.6 |

### DIFF
--- a/docs/basic_usage/mps_dp.md
+++ b/docs/basic_usage/mps_dp.md
@@ -8,7 +8,7 @@ Same-GPU data parallelism runs several complete serving replicas on one GPU and 
 
 ## Deploy
 
-The steps below are one continuous flow. We provide a script `examples/launch_same_gpu_dp.sh` to wrap them behind a per-run state directory (`up/verify/down/list`). It records replica PIDs, process groups, ports, and logs, refuses to start when ports are taken or another run is recorded on the same GPU, health-gates sequential startup, surfaces each replica's logged KV allocation, compares expected replica processes against the MPS client list, and tears down only the processes it recorded. The manual steps below explain each mechanism, checkpoint, and failure mode; the launcher is a convenience wrapper around them, not a production supervisor, and it does not remove your responsibility to check actual KV capacity, MPS attachment, and saturation. Detailed instructions are as follows:
+The steps below are one continuous flow. We provide a script `examples/mps_dp/launch.sh` to wrap them behind a per-run state directory (`up/verify/down/list`). It records replica PIDs, process groups, ports, and logs, refuses to start when ports are taken or another run is recorded on the same GPU, health-gates sequential startup, surfaces each replica's logged KV allocation, compares expected replica processes against the MPS client list, and tears down only the processes it recorded. The manual steps below explain each mechanism, checkpoint, and failure mode; the launcher is a convenience wrapper around them, not a production supervisor, and it does not remove your responsibility to check actual KV capacity, MPS attachment, and saturation. Detailed instructions are as follows:
 
 1. **Choose the GPU and NUMA node.**
 
@@ -64,7 +64,7 @@ Memory profiling does not coordinate these independent processes. The launcher t
 
 ```bash
 MAX_TOTAL_TOKENS=<COMMON_FEASIBLE_TOKEN_CAP> \
-  bash examples/launch_same_gpu_dp.sh up
+  bash examples/mps_dp/launch.sh up
 ```
 
 The H100 Higgs runs used the following caps:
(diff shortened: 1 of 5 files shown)


docs/basic_usage/mps_dp.md
```diff
@@ -8,102 +8,58 @@ Same-GPU data parallelism runs several complete serving replicas on one GPU and
...
+The validated H100 Higgs DP3 cap is `100000` tokens per replica. This value is a starting point for the tested configuration, not a universal hardware default. Recalculate the cap after changing the model, GPU, runtime, replica count, memory settings, or CUDA-graph settings. If a replica cannot allocate the common cap, lower it or reduce the replica count.
 
-The launcher passes the same cap to every replica when `MAX_TOTAL_TOKENS` is set:
+3. **Drive every replica to saturation.**
 
-```bash
-MAX_TOTAL_TOKENS=<COMMON_FEASIBLE_TOKEN_CAP> \
-  bash examples/launch_same_gpu_dp.sh up
-```
+To reach maximum throughput, feed replicas one by one: keep sending to a replica until it is saturated (`#queue-req > 0`), then move to the next. Aggregate concurrency alone can leave some replicas under-driven. SGLang Omni's router does not yet support this fill-one-then-next behavior; that is planned for later work. Equal KV capacity makes the replicas comparable, but it does not by itself balance their queues. Validate the routing policy and per-replica saturation under your workload.
```
JiaxinD: Non-blocking, docs framing. "To reach maximum throughput, feed replicas one by one" (and line 52) reads as a measured routing recommendation, but the case study drove one dedicated client per replica (equal parallel load saturating every replica at once), which is not sequential filling, and no sequential vs round-robin comparison was run. This PR's diff also removes the prior sentence that random routing over equal workers is the most efficient way to keep the pool saturated, which was the more evidence-aligned claim (local runs indicate capacity-proportional splitting does not help when KV is not the bottleneck).

Suggestion: reframe around the measured goal (every replica reaches saturation) and mark fill-one-then-next as one untested single-source technique, not a measured optimum. Keep the existing "Validate the routing policy and per-replica saturation under your workload" hedge.

examples/mps_dp/launch.sh
```diff
@@ -181,8 +210,9 @@ teardown_state() {
   # Note (jiaxin): these GPUs are shared; teardown only signals processes recorded
   # in this run's state, never scans the whole GPU, and keeps the state directory
   # whenever cleanup cannot be confirmed, so nothing is hidden from inspection.
-  local state=$1 keep=${2:-} pgid t live raw
+  local state=$1 keep=${2:-} pgid t live raw control_pid=""
   [ -n "$state" ] && [ -f "$state/replicas.tsv" ] || die "invalid or missing run state '$state'"
+  control_pid=$(mps_control_pid "$state" || true)
   while IFS=$'\t' read -r _ _ pgid _ _; do
     kill -TERM -- "-$pgid" 2>/dev/null || true
```
JiaxinD, line 233: Non-blocking, teardown robustness. This TERM loop (and the SIGKILL pass near line 263) signals every recorded pgid with no liveness or identity check first. Because the keep-state branches and `up()`'s stale-dir refusal both defer cleanup to a later `down`, a run whose leader PID was recycled by an unrelated same-user process group would have that group signalled on `down <RUN_ID>`. Probability is low (needs PID wraparound onto a group leader plus retained stale state), but it is a wrong-kill vector on a shared host, and `tracked_pids`/`pid_is_live` only guard the drain and final checks, not this initial signal.

Suggestion: gate the signal on the recorded leader still being live (for example `pid_is_live "$leader_pid" || continue` before the kill), and for full identity record each leader's start time (`ps -o lstart= -p $pid`) at the launch write (line 387) and skip the kill unless the live leader's start time still matches.

JiaxinD (commented):
## Review: same-GPU DP MPS launcher (H100 Higgs recipe)

Overall this is a solid, carefully written recipe and I endorse the H100 direction. The launcher uses a private per-run MPS pipe, sequential health-gated replica starts, fail-closed equal-KV enforcement (a missing or mismatched resolved `#tokens:` value aborts the launch), explicit MPS attachment verification, and a run-scoped teardown that only signals processes it recorded and keeps the state directory whenever cleanup cannot be confirmed. An independent shell-correctness pass and a KV-enforcement pass over the launcher found no correctness or safety blockers.

**Action: COMMENT. Blocking findings: 0. Non-blocking findings: 4 (2 inline, 2 here).**

Inline:
1. `docs/basic_usage/mps_dp.md:44`: fill-one-then-next is framed as the way to reach maximum throughput without a measured routing comparison, and this PR also drops the prior "random routing over equal workers is most efficient" sentence.
2. `examples/mps_dp/launch.sh:217`: teardown signals every recorded pgid with no liveness or identity gate, a low-probability PID-reuse wrong-kill on a delayed `down`.

Non-blocking, here:

3. **Perf evidence (PR body table).** The table does not state whether the three `Rep` rows are three replicas or three repetitions, or whether `QPS` is per-replica or aggregate. Read as per-replica it implies about 6x; read as aggregate it is about 2x, which matches the docs case study (DP3 1.8 to 2.1x over a ~22 qps single). Please label the table (aggregate QPS, N repetitions) and cite the matched single-replica baseline next to it so the two series line up and the reader cannot misread a 2x gain as 6x. This is a labeling fix, no re-run needed.

4. **Doc rename coordination.** This renames `docs/basic_usage/same_gpu_dp.md` (merged via #986) to `mps_dp.md`. The pending figures draft #1042 targets the old filename and will conflict with the rename; worth coordinating so the figures land on the new name.

**Scope.** The H100-only scope is correct. H200 is honestly marked unvalidated and there is no YAML config in this PR, so keeping H200 plus any YAML as a separate PR is the right split. The equal-KV motivation is well supported: the docs give concrete measured per-replica token counts (97,503 / 53,149 / 20,961) for sequential `mf=0.27` replicas, and the enforcement is a correct upper-bound cap, not a throughput claim.

**CI.** `lint`, `build-docs`, and `test-layout` pass. `omni-ci-gate` is failing only because it is the `run-ci` label gate (the heavy ASR/TTS/Qwen3-Omni jobs are skipping pending a maintainer trigger); it is not a code failure.

docs/basic_usage/mps_dp.md
Ratish1: done

examples/mps_dp/launch.sh
Ratish1: done

JiaxinD (approved):

PR #1041, opened 2026-07-12
Add TP integration to Moss; Moss-TD TP parity test

haojin2:
## Modifications
- Add TP integration to Moss-TD (ASR) model
- Add TP Parity test for Moss

## Related Issues
TBA

## Accuracy Test
TBA

## Benchmark & Profiling
TBA

## Checklist

- [x] Add unit tests.

## CI

if labeled.

### DIFF
--- a/sglang_omni/models/moss_transcribe_diarize/stages.py
+++ b/sglang_omni/models/moss_transcribe_diarize/stages.py
@@ -112,6 +112,8 @@ def create_sglang_moss_transcribe_diarize_executor(
     request_build_max_pending: int | None = 16,
     stream_emit_interval_s: float = 0.05,
     server_args_overrides: dict[str, Any] | None = None,
+    tp_rank: int = 0,
+    nccl_port: int | None = None,
 ):
     gpu_id = int(device.split(":")[-1]) if ":" in device else 0
 
@@ -165,6 +167,8 @@ def create_sglang_moss_transcribe_diarize_executor(
         server_args,
         gpu_id,
         model_arch_override="MossTranscribeDiarizeForConditionalGeneration",
+        tp_rank=tp_rank,
+        nccl_port=nccl_port,
     )
 
     if want_cuda_graph:
--- a/tests/test_model/test_moss_asr_tp_parity_ci.py
+++ b/tests/test_model/test_moss_asr_tp_parity_ci.py
@@ -1,0 +1,148 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Opt-in TP parity test for MOSS-TD.
+
+This test launches real GPU servers, so it is skipped by default for local machines
+without the MOSS-TD checkpoint and CUDA resources.
+"""
+
+from __future__ import annotations
+
+import os
+import subprocess
+import sys
+import wave
+from pathlib import Path
+from typing import Any
+
+import pytest
+
+from tests.utils import (
+    TPParityServerProcess,
+    extract_text,
+    get_available_port,
+    post_json,
+    stop_server_process,
+    wait_health,
+)
+
+MODEL_NAME = os.environ.get("MOSS_TD_MODEL_NAME", "moss-transcribe-diarize")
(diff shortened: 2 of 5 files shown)


luojiaxuan (changes requested):
TP launch is currently broken because the factory neither accepts tp_size nor passes it into ServerArgs, so tp_size 2 fails before model startup. Please also make the parity test opt-in by default, run black, rebase onto current main, and rerun CI.

tests/test_model/test_moss_asr_tp_parity_ci.py
```diff
@@ -0,0 +1,148 @@
...
+    extract_text,
+    get_available_port,
+    post_json,
+    stop_server_process,
+    wait_health,
+)
+
+MODEL_NAME = os.environ.get("MOSS_TD_MODEL_NAME", "moss-transcribe-diarize")
+MODEL_PATH = os.environ.get("MOSS_TD_MODEL_PATH", "OpenMOSS-Team/MOSS-Transcribe-Diarize")
+RUN_FLAG = os.environ.get("RUN_MOSS_TD_TP_PARITY", "1") == "1"
```
luojiaxuan, line 30: This test is documented as opt-in, but the default value 1 enables a two-GPU run in ordinary collections. Default to 0 and let the dedicated CI job set RUN_MOSS_TD_TP_PARITY=1.

sglang_omni/models/moss_transcribe_diarize/stages.py
```diff
@@ -112,6 +112,8 @@ def create_sglang_moss_transcribe_diarize_executor(
     request_build_max_pending: int | None = 16,
     stream_emit_interval_s: float = 0.05,
     server_args_overrides: dict[str, Any] | None = None,
+    tp_rank: int = 0,
```
luojiaxuan, line 115: TP launch injects tp_size into this factory, but the signature does not accept it, so TP=2 fails before model startup. Add tp_size here and set overrides["tp_size"] before build_sglang_server_args.

haojin2:
> TP launch is currently broken because the factory neither accepts tp_size nor passes it into ServerArgs, so tp_size 2 fails before model startup. Please also make the parity test opt-in by default, run black, rebase onto current main, and rerun CI.

@luojiaxuan Okay will address accordingly, meanwhile will need to clean up some of my local changes first

ShaoKang22:
[pr1041-moss-td-fixes.patch](https://github.com/user-attachments/files/30747791/pr1041-moss-td-fixes.patch)
### Test update

I rebased this PR onto the latest main and tested it with sglang 0.5.16 on 8x H200.

I made the following fixes to get the TP parity test running:

- Rebased to latest main to resolve `ModelRunner.__init__()` compatibility issues.
- Added `tp_size` to the MOSS-TD stage factory and injected it into overrides when `tp_size > 1`.
- Removed overrides keys that are incompatible with sglang 0.5.16:
  - `cuda_graph_max_bs`
  - `cuda_graph_bs`
  - `cuda_graph_max_bs_decode`
  - `cuda_graph_bs_decode`
  - `sampling_backend`
  - `torch_compile_max_bs`
- Temporarily skipped `validate_generation_batch_policy` because it accesses `server_args.cuda_graph_config.decode.max_bs`, which is not available in this version.
- Updated the TP parity test to use `audios` + `<audio>` request format.
- Updated the test to launch the server with a YAML config instead of CLI `--stages.0.gpu`, since the CLI parser does not handle GPU lists correctly.
- Added `"terminal": True` to the `asr` stage config to satisfy `StageConfig` validation.
- Changed GPU IDs in the test config to logical IDs, e.g. `[0, 1]`, because `CUDA_VISIBLE_DEVICES` remaps physical GPUs.

### Result

Both TP=1 and TP=2 servers start successfully, and the generated text outputs are consistent.

```bash
export RUN_MOSS_TD_TP_PARITY=1
pytest tests/test_model/test_moss_asr_tp_parity_ci.py -v

PR #1045, opened 2026-07-13
[Perf] Overlap MOSS-TD audio encoding with LM compute (pre-LM encoder service)

Ccyest:
## Motivation

The whisper encoder ran inside the LM forward: every prefill stalls all running requests for a batch-1 encode (~5.4 ms), and sglang's per-request mm dispatch cannot batch or hide it. This PR encodes at request-build time on a dedicated thread + CUDA stream and hands results to the LM through sglang's native `precomputed_embeddings` path — the compute-bound encoder overlaps the memory-bound decode on the same GPU. Repeated (radix-warm) audio is served from a hash-keyed cache instead of re-encoding. No new config surface.

## Benchmarks

H200, movies800 full set (800 samples), base vs PR interleaved on the same GPU, paired per-sample analysis. Both arms run with `encoder_torch_compile=True` (with #1046's mode fix — reduce-overhead crashes independent of this PR) on top of main `73905ed4`:

| concurrency | 1 | 4 | 16 | 32 |
|---|---:|---:|---:|---:|
| median latency (paired delta) | +4.9ms | +4.1ms | **−33.8ms (−9%)** | **−98.9ms (−17%)** |
| latency p95 (s) | 0.51→0.55 | 0.56→0.56 | 0.98→0.93 (−5%) | **1.59→1.33 (−16%)** |
| steady-state throughput (qps) | 2.16→2.32 | 9.75→9.62 | **33.7→37.0 (+9.7%)** | **43.7→53.2 (+21.8%)** |
| samples faster | 9% | 35% | 81% | 91% |
| CER (base / PR) | 5.82 / 5.77 | 5.82 / 5.82 | 5.90 / 5.81 | 5.82 / 5.80 |

Gains grow monotonically with concurrency (crossover between c=8 and c=16); below that the thread-handoff overhead costs single-digit ms. The same sweep with torch.compile off shows the same magnitudes (−31ms/+10% at c=16, −96ms/+23% at c=32) — the win is orthogonal to how the encoder kernels are generated. Steady-state qps excludes the two greedy-repetition straggler samples (#975), which dominate makespan identically on both arms.

Router CI methodology (DP=2, default config): movies800 non-stream 37.8 → 38.8/39.5 qps (+2.7~4.5%), stream and AISHELL4-long at parity, all checks green including `stream text_ttft_p95` (0.068-0.070 vs 0.0774 threshold).

Correctness: c=1 outputs bit-identical to main (800/800, incl. 44k-char outputs); CER/cpCER/DER in-band everywhere; AISHELL4-long chunked-prefill × precomputed path exercised end-to-end (0 failures).


sglang_omni/models/moss_transcribe_diarize/encoder_service.py
```diff
@@ -0,0 +1,118 @@
...
+            self._item_count += len(items)
+            if self._batch_count % 50 == 1:
+                logger.info(
+                    "MOSS-TD pre-LM encoder stage: %d batches, %d items "
+                    "(avg %.2f items/batch, last batch: %d)",
+                    self._batch_count,
+                    self._item_count,
+                    self._item_count / self._batch_count,
+                    len(items),
+                )
```
zhaochenyang20, line 98: f-string

zhaochenyang20 (changes requested):

sglang_omni/models/moss_transcribe_diarize/encoder_service.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+class BatchedAudioEncoderService:
+    ENCODE_TIMEOUT_S = 300.0
+
+    def __init__(self, model: Any) -> None:
+        self._model = model
+        self._device = next(model.whisper_encoder.parameters()).device
+        self._stream = torch.cuda.Stream(device=self._device)
+        self._cache = StageOutputCache(
```
0xjeffro, line 33: Should we protect put/get method in this `StageOutputCache` with a lock to make our concurrent calls safe?

stephenkgli:
https://github.com/sgl-project/sglang-omni/blob/c33186ab946b485f2432cdd890ba662f97ec13b9/sglang_omni/models/moss_transcribe_diarize/sglang_model.py#L209

this comment is stale after this change

sglang_omni/models/moss_transcribe_diarize/encoder_service.py
```diff
@@ -0,0 +1,121 @@
...
+            max_size=_CACHE_MAX_ENTRIES,
+            max_bytes=_CACHE_MAX_BYTES,
+            cache_device="cpu",
+        )
+        self._cache_lock = threading.Lock()
```
stephenkgli: It would be better to make the cache thread-safe, so that every caller can use it safely.

Ccyest:
## Benchmark: PR #1045 vs upstream/main

**Setup** — `benchmark_asr_transcribe_diarize` · dataset `movies800times` · N=100 · concurrency 16 · throughput + accuracy.
baseline `e795c352` (upstream/main) vs candidate `51b02bb8` (this PR head, pre-`87f099b3`). No dependency changes — pure code comparison. main's commits since the PR merge-base touch only TTS/Fish/MPS files (none on the ASR path), so main HEAD ≡ merge-base for this benchmark.
Single H200 per server (MOSS-TD is single-stage ASR); candidate and baseline each pinned to their own idle card, 3 repeats/side, medians reported.

**Accuracy (CER — Chinese multi-speaker, not WER): no regression**

|---|---|---|---|---|
| 16 | cer_no_spk (%) | 5.63 | 5.63 | +0.0% |
| 16 | cp_cer (%) | 15.52 | 15.52 | +0.0% |

**Throughput / latency** (higher audio s/s & QPS better; lower latency & RTF better)

|---|---|---|---|---|
| 16 | mean latency (s) | 0.621 | 0.566 | −8.9% |
| 16 | p95 latency (s) | 1.370 | 1.228 | −10.4% |
| 16 | mean RTF | 0.0539 | 0.0492 | −8.7% |
| 16 | audio throughput (s/s) | 342.4 | 373.2 | **+9.0%** |
| 16 | request throughput (QPS) | 24.20 | 26.37 | **+9.0%** |

Per-repeat QPS (tight, non-overlapping): candidate 26.26 / 26.37 / 26.43 · baseline 24.29 / 23.81 / 24.20. A swap-card check (candidate→GPU1 = 26.10 QPS, baseline→GPU0 = 24.13 QPS) confirms the gap tracks the code, not the card.

All 8 runs completed 100/100 requests with **0 failures**, so the thread-safe pre-LM encoder cache is correct under concurrency=16. On the review discussion: this supports keeping the cache internally thread-safe (direction 1) — it delivers +9% throughput / −9–10% latency with zero accuracy change. The `remove_if` deadlock footgun (predicate re-entering the lock) is fixed in `87f099b3` by evaluating the predicate outside the lock, plus concurrency regression tests.

<sub>Note: the request-timing window is short (~3.8–4.1 s for N=100 at rtf≈0.05), so QPS is reported as the median of 3 repeats rather than a single run.</sub>

tests/unit_test/scheduling/test_stage_cache.py
zhaochenyang20, line 1: change `tests/readme.md`

zhaochenyang20 (approved):

tests/unit_test/scheduling/test_stage_cache.py
Ccyest: Done — added `tests/unit_test/scheduling/test_stage_cache.py` to `tests/README.md` (folder tree + the `unit_test/scheduling/` description covering `StageOutputCache` concurrency/eviction) in 58ee7f47.

PR #1046, opened 2026-07-13
fix(moss-td): drop reduce-overhead from encoder torch.compile (illegal memory access)

Ccyest:
## Motivation

`encoder_torch_compile=True` (#979) hard-crashes real serving: `mode="reduce-overhead"`'s cudagraph trees corrupt memory alongside the decode CUDA graphs that always run in this process. Full movies800 e2e at c=16 (H200, torch 2.11.0+cu130): 800/800 requests fail with `CUDA error: an illegal memory access` ~60-90s in, with async decode both on and off. CI never caught it because the flag defaults to off and #979's tests stub the infra.

## Fix

Compile with the default mode. Same run with this one-argument change: 800/800 success, 0 IMA, CER 5.83 (in-band), median latency 0.371s (parity with the encoder-CUDA-graph baseline 0.364-0.377s).


draftbk (approved):

sglang_omni/models/moss_transcribe_diarize/sglang_model.py
```diff
@@ -171,10 +171,7 @@ def compile_encoder(self, chunk_buckets, input_feature_len: int) -> None:
                     continue
                 warmed.append(n)
         self._compiled_chunk_buckets = frozenset(warmed)
-        logger.info(
-            "MOSS-TD encoder torch.compile(reduce-overhead) warmed buckets=%s",
-            warmed,
-        )
+        logger.info("MOSS-TD encoder torch.compile warmed buckets=%s", warmed)
```
Ratish1: nit: can we use f formatting?

Ratish1 (approved):
LGTM with one nit comment

PR #1047, opened 2026-07-13
[Perf] Isolate Fish S2-Pro reference encoding from GPU stages

Hayden727:
## Summary

- move Fish S2-Pro's CPU-only reference preprocessing into its own process
- bound the shared PyTorch intra-op pool from the process CPU affinity and the existing preprocessing concurrency
- preserve an explicit `OMP_NUM_THREADS` operator override
- keep `tts_engine` and `vocoder` colocated on the same GPU; no extra GPU or user-facing flag

## Motivation

The current topology runs CPU reference encoding in the same OS process as the GPU TTS engine and vocoder. With concurrent cold references, the CPU codec's native thread pool delays GPU scheduling instead of overlapping with it.

A stage trace on latest main showed:

| Stage | Mean | p95 |
| --- | ---: | ---: |
| preprocessing | 5.236 s | 7.460 s |
| tts_engine | 1.175 s | - |
| vocoder | 0.199 s | - |

Preprocessing accounted for 79.2% of measured stage residence, while scheduler queue p95 was only 31.6 ms. After isolation, preprocessing fell to 0.915 s mean / 1.248 s p95; the added preprocessing-to-TTS IPC hop was 12.9 ms mean / 29.1 ms p95.

## H200 benchmark

- base: `1618b7ce26f5ea464e849cfcd943c0e429a711a6`
- image: `hongccc/sglang-omni:dev`
- hardware: one H200, same GPU for main and this branch
- workload: 128 distinct SeedTTS voice-cloning references, streaming, `max_new_tokens=2048`, concurrency 16
- two 20-request warmups used a disjoint reference set, so the measured references were cold
- 128/128 requests completed with zero failures in both runs

| Metric | main | this PR | Delta |
| --- | ---: | ---: | ---: |
| QPS | 1.383 | 3.297 | **+138.4%** |
| Mean latency | 11.093 s | 4.724 s | **-57.4%** |
| Mean RTF | 2.7773 | 1.1624 | **-58.2%** |
| Mean TTFC | 9.8960 s | 2.9896 s | **-69.8%** |

At concurrency 8, a separate cold-reference run improved QPS from 1.363 to 2.883 (+111.5%).

### Warm-cache control

I repeated the same 128 references after populating the in-process cache. This checks that the extra process hop does not regress the common fixed-speaker path.

| Metric | main | this PR | Delta |
| --- | ---: | ---: | ---: |
| QPS | 4.674 | 4.669 | -0.1% |
| Mean latency | 3.313 s | 3.332 s | +0.6% |
| Mean RTF | 0.7980 | 0.8050 | +0.9% |
| Mean TTFC | 0.9350 s | 0.8978 s | -4.0% |

The warm-cache path is effectively unchanged.

## Resource tradeoffs

One-second `nvidia-smi dmon` samples during the cold c16 run showed GPU nonzero-SM duty cycle rising from 72.5% to 88.9%, and all-sample mean SM utilization from 50.9% to 55.1%. The optimization keeps the GPU fed rather than changing model compute.

The separate process costs about 2.0 GiB additional host RSS and increases total server threads from 700 to 808. Sampled peak GPU FB usage was about 2.1 GiB higher under the faster run, while model load, KV pool, and post-CUDA-graph available-memory logs were identical. A 4/8/28 intra-op thread sweep selected 8 as the best throughput/resource point; an explicit `OMP_NUM_THREADS` remains authoritative.

## Validation

- `pytest -q tests/unit_test/fishaudio_s2_pro` (63 passed)
- pre-commit on all three changed files
- cold-reference and warm-cache end-to-end SeedTTS A/B on the server described above


Hayden727:
Rebased onto latest main `1618b7ce` as `420c6e83`. Post-rebase validation in `hongccc/sglang-omni:dev`: Fish unit suite **21 passed**, all pre-commit hooks passed, and a latest-head H200 streaming smoke completed **16/16** requests with zero failures. The performance table in the PR body remains the controlled 128-cold-reference A/B on `73905ed4`; I am not mixing the smaller smoke run into that claim.

sglang_omni/models/fishaudio_s2_pro/stages.py
```diff
@@ -30,6 +30,27 @@
 
 logger = logging.getLogger(__name__)
 
+_MAX_PREPROCESSING_INTRAOP_THREADS = 8
+
+
+def _configure_preprocessing_threads(worker_count: int) -> int:
+    if "OMP_NUM_THREADS" in os.environ:
+        return torch.get_num_threads()
```
SandyLuXY: `OMP_NUM_THREADS` is treated as authoritative by presence only, but this returns whatever PyTorch selected: in clean subprocesses `OMP_NUM_THREADS=0` left the pool at 64, and `OMP_NUM_THREADS=3 MKL_NUM_THREADS=5` returned 5. Please validate the override, set the effective PyTorch pool explicitly (or fail startup), and cover this with a real subprocess instead of mocking `get_num_threads()`.

yxs (approved):

PR #1049, opened 2026-07-13
[router] Add multiprocess CP/DP router with shared admission and crash-safe control plane

JiaxinD:
## Summary
This PR addresses: the single-process Python relay is bound to one core (~65% of per-request CPU is fixed overhead independent of response size), so throughput does not scale on a multi-core host. The router becomes one control plane plus N stateless data-plane relay processes behind a shared listening socket. Default `--router-processes 1` is behaviour-identical to today's router; `N >= 2` scales forwarding across cores while preserving the admission bound, admin surface, metrics, and crash semantics.

## Architecture

- **Supervisor** binds the public socket once and passes the inherited fd to N **data-plane (DP)** children (shared accept queue), spawns one **control-plane (CP)** child, and owns the admission shared-memory segment and a parent-death pipe. It reaps a dead child before reclaiming its resources, bumps the generation, and respawns; it fails closed on a crash loop and tears the tree down in two phases.
- **CP** is the single owner of the worker registry, health checker, admin surface, and weight-update journal. It publishes a routable-worker **snapshot** (atomic file, monotonic per-epoch seq) on every state change and on a keepalive cadence.
- **DP** is a stateless relay fed by the snapshot: it relays the model routes through the shared admission, forwards the privileged admin surface and aggregate `/health` to the CP (which re-checks auth), serves `/v1/models` locally, and sheds with a fast 503 when the snapshot goes stale.
- **Shared admission**: one mmap slot per DP, per-slot seqlock; the global in-flight bound is soft (overshoot bounded by `N - 1`).
- **Restart/recovery**: a fenced child stops serving; a restarted CP gets a fresh epoch and rebuilds counters from a since-CP-start baseline; a crashed DP's slot is reclaimed after reap; a weight update that dies mid-transaction is recovered fail-closed from the journal.

## Correctness / failure semantics

Grouped; specific corner cases live in the tests, the review guide, and `CROSS-REVIEW-NOTES.md`.

1. **Endpoint ownership + internal auth**: data/public-metadata/health on the DP; worker CRUD (deliberately unauthenticated, frozen by test) and weight-update/broadcast on the CP, which re-checks the admin key the DP forwards. The internal channel is a `0700` UDS (or a token-guarded localhost fallback).
2. **Snapshot staleness + CP/DP restart**: a stale DP sheds with a fast 503 and flips `/ready`; one republish restores it. A restarted CP is a new epoch, so a bare seq from a previous CP cannot satisfy it; a heartbeat the CP has no record of gets 428 (re-register), reserving 409 for a genuinely stale generation, so a fast CP restart cannot false-fence the fleet.
3. **Epoch/incarnation + generation fencing**: workers carry a regenerated incarnation id through registry → snapshot → DP view → failure reports; a changed incarnation replaces the DP worker object so an in-flight reference keeps its identity. Seqlock readers retry across a mid-write marker; acquire/release verify slot ownership and no-op if reclaimed.
4. **ACK barrier + crash-safe journal**: the CP publishes the disabled snapshot, waits for every live/expected DP to acknowledge that epoch+seq, then broadcasts; on timeout it fails closed. The target set is fsynced to a stable host:port path under a durable state directory (`--router-state-dir`, default `$XDG_STATE_HOME/sglang-omni-router`) before disable/publish, so it survives a host reboot the remote workers outlive; a CP that restarts mid-transaction re-disables the journaled targets until an operator re-enables them (admin-authenticated); an unreadable journal fails closed.
5. **Shared admission + seqlock**: soft global bound; a two-slot retired fold stays mid-write across both writes so aggregates never transiently double-count; `/health` reports an admission read error rather than a 500 if the shm is momentarily unstable.
6. **Observability + cleanup/reclamation**: cumulative totals keyed by `(dp_index, generation, counter_seq)`, high-water-clamped so a regressed report cannot move totals backwards; failure events deduplicated by `(dp_index, generation, failure_seq)`; a dead DP's admission slot is reclaimed after reap so its in-flight budget is freed.

## Performance

Four distinct measurements, kept separate:

1. **Stage 1 scaffold scaling** — a throwaway scaffold used only to prove the multi-process premise; not final-implementation numbers.
2. **Final complete Router scaling** — the full CP/DP stack (the matrix below).
3. **Router-limited capacity** — mock upstream (>10k qps), so the relay's single core is the bottleneck (matrix and overload).
4. **Worker-limited direct/via parity** — real GPU workers (~80 qps/pool), so the GPU is the bottleneck; measures whether routing adds overhead.

**Setup**: novita-h100, 128-core dual-socket (NUMA node0 0-31 / node1 32-63). Router pinned to node0 cores 0-15; oha clients on node1 cores 48-55; mock upstreams on node1 cores 40-47. Mock upstream returns a fixed base64-audio JSON (no GPU). TTS shape = ~230 B request / 300 KB response; Omni shape = 342 KB request / 300 KB response. Matrix/cost = closed-loop (`oha -c <knee>`, each config at its throughput knee), 30 s rounds, formal triplicate; overload = open-loop `hammer_ol.py`, offered 4800; warm-up 8-22 s. Python 3.12 / uvicorn 0.47 / httpx 0.28. Artifacts on the box under `/data/dp2exp/p1/` (`p1_gate.sh`, `w5_*.sh`, `runs5final/`, `runs_confirm/`).

The baseline is the **original single-process Router**. `N` denotes the number of data-plane relay processes in the new Router; the supervisor and one control-plane process run in addition to these N relay processes. `N` is not the number of GPU workers or model replicas. The N=1 row measures compatibility overhead; N>=2 enables the new multiprocess architecture. All speedups below are relative to the original Router.

**Router throughput** [measured, formal triplicate, mock upstream so the router core is the bottleneck; zero errors, all relay processes 100% CPU] — QPS shown as `median [min, max]`:

| Router | Relay procs | TTS QPS | TTS speedup | Omni QPS | Omni speedup |
| ------ | ----------: | ------- | ----------: | -------- | -----------: |
| Original Router | 1 | 779 [778, 787] | 1.00x | 567 [567, 570] | 1.00x |
| New Router, compatibility mode | 1 | 781 [769, 789] | 1.00x | 549 [549, 568] | 0.97x |
| New multiprocess Router | 2 | 1544 [1533, 1548] | 1.98x | 1087 [1072, 1087] | 1.92x |
| New multiprocess Router | 4 | 3027 [3025, 3043] | 3.89x | 2082 [2037, 2117] | 3.67x |
| New multiprocess Router | 8 | 6021 [5999, 6043] | 7.73x | 4023 [3950, 4047] | 7.09x |

The new Router in N=1 compatibility mode is performance-neutral relative to the original Router within run-to-run noise (the Omni 0.97x reflects overlapping ranges, not a regression). Once the Router is CPU-bound, N=2/4/8 relay processes raise capacity to **7.73x (TTS) / 7.09x (Omni)** the original single-process Router at N=8.

Baseline provenance: the Original Router row is the matched single-process baseline from the same cost rounds as the matrix (same host, pinning, payloads, duration, and triplicate method); Original Router QPS is measured directly, not inferred from CPU cost. The full N=1/2/4/8 matrix was measured on the pre-rebase router terminal state (`f40ae05b`) and the Original-Router baseline alongside it; N=1/N=4, the Original-Router baseline, crash-under-load, and parity were re-confirmed on the rebased publication head after aligning to #1014's `afa1696a` (Original Router re-measured 767 TTS, within noise; New N=4/N=1 = 3.92x TTS / 3.86x Omni). The relay hot path (`selector.py`, `snapshot.py`, admission `try_acquire`/`release`) is byte-identical between the two heads.

- Overload (N=4, bound 64, offered 4800): goodput **2264 qps**, fast rejects **152,636**, peak inflight **65** vs `bound + N - 1 = 67`, timeouts **0**, rejected_total matches the client 503 count exactly.
- Keep-alive skew (2 long-lived connections, ~3% of the bound): zero spurious 503s.

**Worker-limited parity** [measured on `a5e0a70a`; re-confirmed on the rebased head]. GPU: 1x H100; workload: 3x Higgs TTS workers (time-sliced), real voice-clone payloads; Router N=2; concurrency 3x96.

| path | goodput QPS | p50 / p99 | runs |
| ---- | ----------: | --------- | ---: |
| direct | 79.3 | 2567 / 24510 ms | 2 (fwd/rev) |
| via Router N=2 | 82.6 | 2639 / 23933 ms | 2 (fwd/rev) |

Order-symmetric (via-first 94.8%, direct-first 104.1%); order-merged via/direct = **99.3% (>= 98%)**, worker distribution 2388/2388/2386, zero errors. Re-confirmed on the rebased head: via Router N=2 81.2 vs direct 79.2 qps (102.5%, via >= direct, no routing overhead).

## Crash under load [measured, rebased head, N=4, fixed non-overload ~1600 qps offered]

| failure | client-visible errors / timeouts | goodput | recovery | final accounting |
| ------- | -------------------------------- | ------- | -------- | ---------------- |
| SIGKILL one DP | 1 disconnect (its single in-flight) / 0 timeouts | 1597 qps (unaffected) | ~2.6 s (respawn + gen bump) | inflight=0, generation advanced, other 3 DPs kept serving |
| SIGKILL CP | 0 errors / 0 timeouts | 1597 qps (unaffected) | ~4.7 s (fresh epoch, snapshot republished) | new epoch, service restored, registry intact, inflight=0 |

Killing a DP costs only its own in-flight request; the shared socket keeps the other DPs serving and the supervisor respawns the slot. Killing the CP is client-invisible because the DPs keep serving from their last snapshot during the gap (the stale timeout is not reached), then re-register with the new epoch.

## Validation

- **Linux Router suite: 280 passed** (includes real-process integration: two DPs sharing the socket, SIGKILL respawn with a bumped generation, shared-admission accounting, clean shutdown).
- **Windows Router suite: 279 passed + 4 known environment failures** (`test_launcher_cleans_up_managed_workers_on_health_timeout`, `test_launcher_cleans_up_managed_workers_on_startup_interrupt`, `test_launcher_waits_for_managed_workers_in_parallel`, `test_router_console_script_entrypoint_resolves`) — all use `os.getpgid`/console-script resolution that Windows lacks; they fail identically on the #1014 base and are unrelated to this PR.
- Five adversarial review rounds found 2 critical, 30 major, and 13 minor issues; all were closed with regression tests/fixes or explicit scope dispositions (`CROSS-REVIEW-NOTES.md`).
- lint/format clean; default `--router-processes 1` behaviour-identical.

## Compatibility and limitations

- Default `--router-processes 1` keeps the current single-process router; all endpoints and the API are unchanged.
- Multi-process shared admission relies on x86-64 total-store-order plus CPython; other architectures log a warning and are unsupported for `N >= 2` (Linux-only for the shared socket / UDS / shm).
- Rollback: `--router-processes 1`, or revert this PR (it does not touch #1014).
- Follow-ups (out of scope): an operation-id / weight-update status API for lost-response idempotency (the journal prevents mixed-weight enable on crash, not lost-response dedup); an automatic router-process count.

## Review guide

1. lifecycle and shared socket (`supervisor.py`, `dp_runner.py`, `cp_runner.py`)
2. CP/DP ownership (`control_plane.py`, `data_plane.py`, endpoint split in `app.py`)
3. barrier and journal (`control_plane.py`, `update_journal.py`)
4. observability (`observability.py`)
5. admission and fencing (`admission_shm.py`)
6. failure tests (`test_supervisor_integration.py`, `test_control_plane.py`, `test_data_plane.py`)
7. docs and benchmarks (`docs/basic_usage/omni_router.md`)


sglang_omni_router/control_plane.py
```diff
@@ -0,0 +1,429 @@
...
+    dp_ack_timeout_secs: float = DEFAULT_DP_ACK_TIMEOUT_SECS,
+    dp_liveness_secs: float = DEFAULT_DP_LIVENESS_SECS,
+    snapshot_keepalive_secs: float = DEFAULT_SNAPSHOT_KEEPALIVE_SECS,
+    expected_data_planes: int | None = None,
+    admission_shm_path: str | None = None,
+    journal_path: str | None = None,
+) -> FastAPI:
+    if expected_data_planes is not None and expected_data_planes < 1:
+        raise ValueError("expected_data_planes must be >= 1 when set")
+    workers = build_workers(config.workers)
```
Hayden727: A restarted control plane rebuilds `workers` from the original static config, so `/workers` additions and deletions are lost: deleted workers return and dynamically added workers disappear when the startup snapshot is published. Please persist or recover the CP-owned registry across restarts and add a worker-CRUD restart regression test.

sglang_omni_router/update_journal.py
```diff
@@ -0,0 +1,136 @@
...
+        callers can fail closed instead of treating it as "no transaction".
+        """
+        try:
+            with open(self._path, encoding="utf-8") as f:
+                data = json.load(f)
+        except FileNotFoundError:
+            return []
+        except (OSError, ValueError) as exc:
+            raise JournalUnreadableError(str(exc)) from exc
+        workers = data.get("worker_ids", [])
```
Hayden727: Valid but malformed journal JSON bypasses fail-closed recovery: a top-level list raises `AttributeError`, while non-string worker IDs are silently discarded and can make `has_pending()` return false. Please validate that the document is an object and every `worker_id` is a string, raising `JournalUnreadableError` otherwise, and cover both shapes with regressions.

sglang_omni_router/control_plane.py
```diff
@@ -0,0 +1,425 @@
...
+        if report.failure_seq in state["seqs"]:
+            return True
+        state["seqs"].add(report.failure_seq)
+        if len(state["seqs"]) > 4096:
+            state["floor"] = max(state["seqs"]) - 1024
+            state["seqs"] = {seq for seq in state["seqs"] if seq > state["floor"]}
+        return False
+
+    @app.post("/internal/worker_failure", dependencies=[internal_auth])
+    async def worker_failure(report: WorkerFailureReport) -> JSONResponse:
```
Hayden727, line 444: This mutation path does not fence reports against the currently registered DP generation. After generation 2 registers, a delayed generation 1 failure report still returns 200 and can mark the worker unhealthy; the internal counters endpoint has the same gap. Validate the dp_index and generation against internal_state before applying either report and cover an old-generation arrival after replacement.

sglang_omni_router/app.py
```diff
@@ -526,7 +796,20 @@ async def _broadcast_admin_request_locked(
...
+            outcome_known = results is not None
+            _restore_admin_disabled_state(
+                path, workers, previous_disabled, results, outcome_known
+            )
+            _notify_registry_change(app)
+            if journal is not None:
+                if outcome_known:
+                    # ACK abort ([]) or a completed broadcast (list): the
+                    # outcome is known to the caller, so nothing is uncertain
+                    journal.clear()
```
Hayden727: A completed broadcast is not necessarily a safe outcome: with two workers, if one update succeeds and the other returns 500, this clears the journal and restores both workers as routable, leaving a mixed-weight pool serving traffic. Keep the targets disabled and journaled whenever any result fails or has an unknown transport outcome, and add a partial-success regression test.

sglang_omni_router/app.py
```diff
@@ -48,14 +53,63 @@
...
+    if not unresolved:
+        return
+    live = {worker.worker_id for worker in workers}
+    still_present = [worker_id for worker_id in unresolved if worker_id in live]
+    for worker_id in still_present:
+        worker = _find_worker(workers, worker_id)
+        if worker is not None:
+            worker.set_disabled(True)
+    if len(still_present) != len(unresolved):
+        journal.keep(still_present)
```
Ratish1: this seems to overwrite the durable journal with only still_present. A dynamically registered worker is absent after a full supervisor restart because the registry snapshot lives in the old per-run workdir, so this clears the only evidence that its update may have partially applied. Registering the same stable worker ID later makes it routable again.

Keep unresolved IDs as tombstones until an explicit authenticated resolution clears them. Reject or create-disabled any registration whose stable ID is unresolved, and add dynamic-worker/full-restart plus pending delete/re-add tests.

sglang_omni_router/app.py
```diff
@@ -197,30 +285,47 @@ async def create_worker(request: Request) -> JSONResponse:
...
+        lock, rejected = _registry_lock_or_reject(app)
+        if rejected is not None:
+            return rejected
+        if lock is not None:
+            await lock.acquire()
+        try:
+            if any(worker.url == worker_config.url for worker in workers):
+                return _error_response(409, "worker already registered")
+
+            worker = Worker(config=worker_config)
```
Ratish1: After a partial update leaves this stable worker ID journaled and disabled, the DELETE followed by POST of the same URL reaches this constructor and defaults to disabled=False. 

Reject the registration or create it disabled whenever its stable ID is unresolved. Worker deletion must not erase that tombstone.

sglang_omni_router/app.py
```diff
@@ -285,6 +433,10 @@ async def update_worker(worker_id: str, request: Request) -> JSONResponse:
 
         if requested_disabled is not None:
             worker.set_disabled(requested_disabled)
+            if requested_disabled is False:
+                journal = getattr(app.state, "update_journal", None)
+                if journal is not None:
+                    journal.discard(worker.worker_id)
```
Ratish1: an authenticated re-enable is a legitimate operator action, but discard() cannot modify an unreadable file. This handler still returns 200 and enables the worker while every later weight update remains blocked with 409.

Expose and document an explicit authenticated recovery operation that validates or replaces the journal and reports failure if durable resolution did not occur.

sglang_omni_router/data_plane.py
```diff
@@ -0,0 +1,614 @@
...
+        while True:
+            await asyncio.sleep(counter_flush_interval_secs)
+            counter_seq += 1
+            payload = {
+                "dp_index": dp_index,
+                "generation": generation,
+                "counter_seq": counter_seq,
+                "workers": [
+                    {
+                        "worker_id": worker.worker_id,
```
Ratish1, line 391: Worker replacement at the same documented URL creates a fresh DP object whose counters restart at zero, while the CP high-water remains keyed only by stable worker_id. 

Send incarnation/counter epoch and retire the previous segment when it changes. Add a DELETE -> POST same-URL metric regression

sglang_omni_router/supervisor.py
```diff
@@ -0,0 +1,571 @@
...
+    def _start_inner(self) -> None:
+        if self._workdir is None:
+            self._workdir = tempfile.mkdtemp(prefix="sglang-omni-router-")
+        os.chmod(self._workdir, 0o700)
+
+        config_path = os.path.join(self._workdir, "router_config.json")
+        with open(config_path, "w", encoding="utf-8") as f:
+            f.write(self._config.model_dump_json())
+
+        self._socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
Ratish1: --host ::/::1 works through the N=1 Uvicorn path, but this unconditional AF_INET socket fails with gaierror before child startup. This is a direct compatibility regression for Linux IPv6 deployments.

Resolve the configured address with getaddrinfo(..., AI_PASSIVE) or reuse Uvicorn's family-selection logic.

sglang_omni_router/supervisor.py
```diff
@@ -0,0 +1,571 @@
...
+    def shutdown(self) -> None:
+        # Note (Jiaxin Deng): signal every DP first so none keeps accepting
+        # from the shared socket while its siblings drain, then reap together.
+        for slot in self._dp_slots.values():
+            self._signal_child(slot.process)
+        for slot in self._dp_slots.values():
+            self._await_child(slot.process)
+        self._dp_slots.clear()
+        if self._cp_process is not None:
+            self._stop_child(self._cp_process)
```
Ratish1, line 626: During an ordinary graceful shutdown, this waits for the CP before closing the parent-held public socket. If the CP is draining, TCP connections still succeed after all acceptors are gone and clients hang in the backlog. 

we could lose the parent's listener at shutdown start or immediately after signaling the DPs; their inherited descriptors remain valid for their own drain.

sglang_omni_router/admission_shm.py
```diff
@@ -0,0 +1,379 @@
...
+        try:
+            if not self._still_owner():
+                self._fenced()
+                return False
+            total = self._total_inflight()
+        except SeqlockUnstableError:
+            # Note (Jiaxin Deng): a sibling slot is stuck mid-write (writer
+            # crashed); shed instead of blocking the event loop.
+            logger.warning("admission slot unstable; shedding until reclaim")
+            return False
```
Ratish1, line 326: this False path does not increment _rejected_total, so the client receives a rejection that /health never
 counts. If the dying slot remains unreadable during reclaim, its earlier cumulative count can also be lost.  we can make the counters crash-recoverable and count this path, or document rejected_total as best-effort across DP crashes rather than exact.

sglang_omni_router/app.py
```diff
@@ -197,30 +285,47 @@ async def create_worker(request: Request) -> JSONResponse:
...
 
     @app.get("/workers")
     async def list_workers() -> JSONResponse:
-        return JSONResponse(_pool_summary(workers, status="ok", include_workers=True))
+        return JSONResponse(
+            _pool_summary(
+                workers,
+                status="ok",
+                include_workers=True,
+                overlay=getattr(app.state, "worker_stats_overlay", None),
```
Ratish1, line 359: in N>=2 mode this overlay gives /workers the DP totals, while GET /workers/{id} returns the CP-local worker, which never handles data traffic. Use the same overlay for detail responses

sglang_omni_router/control_plane.py
```diff
@@ -0,0 +1,432 @@
...
+        if report.incarnation and worker.incarnation != report.incarnation:
+            return JSONResponse({"status": "ok", "stale_incarnation": True})
+        if _failure_already_applied(report):
+            return JSONResponse({"status": "ok", "deduplicated": True})
+        worker.record_request_failure(
+            failure_threshold=config.health_failure_threshold,
+            status_code=report.status_code,
+            error=report.error,
+        )
+        publish()
```
Ratish1: Every gateway failure currently serializes and replaces the full snapshot, including failures that only increment a counter after the worker is already unhealthy.

Have the state transition report whether routability changed and publish only then; the keepalive covers periodic liveness writes

sglang_omni_router/control_plane.py
```diff
@@ -0,0 +1,437 @@
...
+            return error
+        worker = _find_worker(workers, report.worker_id)
+        if worker is None:
+            return _error_response(404, "worker not found")
+        if report.incarnation and worker.incarnation != report.incarnation:
+            return JSONResponse({"status": "ok", "stale_incarnation": True})
+        if _failure_already_applied(report):
+            return JSONResponse({"status": "ok", "deduplicated": True})
+        before = (worker.is_routable, worker.state)
+        worker.record_request_failure(
```
Ratish1, line 460: A request can still be in flight when an operator sets is_dead=true. Its matching-incarnation failure report reaches this call and, at the failure threshold, changes the authoritative state from dead to unhealthy. The next successful health probe no longer skips the worker and changes it to healthy/routable.

Deduplicate the event first, then consume without applying it while worker.is_dead. Add the exact PUT-dead -> delayed failure -> successful health-probe regression; deduplicating before the dead check prevents a retried event from applying after a later explicit clear-dead.

sglang_omni_router/supervisor.py
```diff
@@ -0,0 +1,583 @@
...
+            self._start_inner()
+        except BaseException:
+            # Note (Jiaxin Deng): a half-started supervisor must not leak
+            # spawned children, the bound socket, or workdir artifacts.
+            self._cleanup_partial_start()
+            raise
+
+    def _cleanup_partial_start(self) -> None:
+        for slot in self._dp_slots.values():
+            self._stop_child(slot.process)
```
Ratish1: If spawning a later DP fails, this cleanup stops and waits for each earlier DP sequentially and closes the parent listener only after the CP drain. During a blocked first wait, sibling DPs remain completely unsignaled; after the last acceptor exits, the open parent listener again accepts connections into a backlog with no acceptor. This violates  transactional-startup and reproduces the black hole fixed in normal shutdown().

Close the parent listener first, signal every started DP before awaiting any of them, reap all DPs, then stop the CP. Extend the existing spawn-failure test to assert ordering and real-port refusal during rollback.

sglang_omni_router/app.py
```diff
@@ -284,6 +457,18 @@ async def update_worker(worker_id: str, request: Request) -> JSONResponse:
         worker.replace_config(next_config)
 
         if requested_disabled is not None:
+            if requested_disabled is False:
+                journal = getattr(app.state, "update_journal", None)
+                if journal is not None and not journal.discard(worker.worker_id):
```
Ratish1: _apply_worker_update() calls replace_config() before this new failure point. With a corrupt journal, PUT {"disabled": false, "model": "new-model"} returns 503 but changes the model anyway; in CP mode the keepalive then publishes that rejected change to every DP. 

Validate authentication, journal resolution, and every other fallible precondition before committing the staged config/state.

sglang_omni_router/internal_channel.py
```diff
@@ -0,0 +1,195 @@
...
+        preserved_seq = 0
+        if (
+            current is not None
+            and hello.generation == current.generation
+            and hello.pid == current.pid
+        ):
+            # Note (Jiaxin Deng): idempotent re-register; the DP's applied
+            # snapshot did not go backwards.
+            preserved_seq = current.last_applied_seq
+        record = DataPlaneRecord(
```
Ratish1, line 104: If a heartbeat is processed but its response is lost, the DP sets registered=False and re-registers with the same generation/PID. This path preserves only sequence 7 but resets its epoch to "" and serving to true. The CP then reports false degradation and can delay or abort an ACK barrier until another heartbeat repairs it.

Preserve sequence, epoch, and serving for the exact same generation/PID.

sglang_omni_router/update_journal.py
```diff
@@ -0,0 +1,148 @@
...
+    def _write(self, data: dict) -> None:
+        directory = os.path.dirname(self._path) or "."
+        os.makedirs(directory, exist_ok=True)
+        tmp_path = f"{self._path}.tmp"
+        with open(tmp_path, "w", encoding="utf-8") as f:
+            f.write(json.dumps(data))
+            f.flush()
+            os.fsync(f.fileno())
+        os.replace(tmp_path, self._path)
+        self._fsync_dir()
```
stephenkgli, line 229: `_fsync_dir()` swallows OSError, so `journal.begin()` can succeed and the weight update can proceed without a durable journal. After a partial update and host crash, recovery may see no journal and re-enable workers with mixed weight versions.

sglang_omni_router/data_plane.py
```diff
@@ -0,0 +1,615 @@
...
+            else:
+                if (
+                    worker.model != config.model
+                    or worker.capabilities != config.capabilities
+                ):
+                    worker.replace_config(config)
+                if entry.incarnation:
+                    worker.incarnation = entry.incarnation
+            worker.state = entry.state
+            worker.disabled = entry.disabled
```
stephenkgli, line 135: The DP and CP both call `record_request_failure()`for the same event. A successful CP health check resets the CP counter, but the snapshot only overwrites the DP’s `state` and `disabled` fields, leaving its `consecutive_failures` stale. With threshold 3, a DP can retain 2 failures, receive a healthy snapshot, then locally evict the worker on the next failure while the CP has counted only 1. Please let the DP only report failures and consume CP health state

stephenkgli (commented):
Thanks for the great work on this. Left a couple of comments for your consideration.

sglang_omni_router/app.py
JiaxinD: Fixed the false success: discard() now reports whether the entry was durably resolved, and the re-enable returns 503 with the journal path instead of 200. Left open since I did not add a separate recovery endpoint; say the word if you want one.

sglang_omni_router/selector.py
```diff
@@ -14,9 +14,17 @@ class NoEligibleWorkerError(RuntimeError):
...
+        policy: RoutingPolicy,
+        *,
+        seed: int | None = None,
+        rr_offset: int = 0,
+    ) -> None:
         self.policy = policy
-        self._rr_index = 0
+        # Note (Jiaxin Deng): rr_offset staggers round-robin starts so N
+        # fresh DPs do not all herd onto worker 0 at once.
+        self._rr_index = rr_offset
```
Ratish1, line 27: _select_round_robin() stores (index + 1) % len(candidates). If worker B is temporarily unhealthy—or a model/capability filter leaves only A—every DP resets its cursor to zero. When B returns, all DPs select A and then B in lockstep, permanently defeating this stagger and recreating cross-process herding. I

Keep a monotonic cursor and apply modulo only while indexing, or preserve an explicit stable per-DP phase.

sglang_omni_router/supervisor.py
```diff
@@ -0,0 +1,583 @@
...
+    if not fd_value:
+        return
+
+    def _watch() -> None:
+        try:
+            while os.read(int(fd_value), 1):
+                pass
+        except OSError:
+            pass
+        os.kill(os.getpid(), signal.SIGTERM)
```
Ratish1, line 175: On supervisor EOF this watcher sends one SIGTERM. Both child runners use Uvicorn without a graceful-shutdown timeout, so an active request that never completes can keep the child—and its inherited public listener—alive indefinitely after a supervisor SIGKILL. The supervisor's normal escalation timer no longer exists to send the later SIGKILL; I reproduced Uvicorn remaining alive in this state.

Add a child-owned watchdog that escalates after a fixed deadline, and/or configure a bounded Uvicorn graceful timeout for both CP and DP. The integration should kill the supervisor while a DP request is deliberately stuck and assert that every descendant and listener disappears within the bound.

sglang_omni_router/app.py
```diff
@@ -197,30 +292,56 @@ async def create_worker(request: Request) -> JSONResponse:
...
+                )
+            workers.append(worker)
+            if config.max_connections < MIN_CONNECTIONS_PER_WORKER * len(workers):
+                logger.warning(
+                    f"max_connections={config.max_connections} is below "
+                    f"{MIN_CONNECTIONS_PER_WORKER} x {len(workers)} workers after "
+                    "registration; the upstream client is sized at startup and can "
+                    "under-feed the grown pool"
+                )
+            await app.state.health_checker.check_worker_health(worker)
```
Ratish1: Registration holds the same lock used to exclude RL weight updates while awaiting this arbitrary worker /health request; clear-dead does the same. A slow or blackholed candidate makes all other CRUD return 409 and prevents an update from starting for the full health timeout. This is reachable during ordinary scale-up or recovery.

Probe staged state outside the short transaction, then acquire the lock, revalidate the worker/version, commit, and publish. Keep serialization around authoritative mutation, not around network latency.

sglang_omni_router/admission_shm.py
```diff
@@ -0,0 +1,392 @@
...
+            # now would race the fold, so this shed cannot be counted.
+            logger.warning("own admission slot unstable; shedding until reclaim")
+            return False
+        try:
+            total = self._total_inflight()
+        except SeqlockUnstableError:
+            # Note (Jiaxin Deng): a sibling slot is stuck mid-write (writer
+            # crashed); shed instead of blocking the event loop. The own slot
+            # is intact, so the rejection is counted like any other shed.
+            logger.warning("sibling admission slot unstable; shedding until reclaim")
```
Ratish1: A DP killed with an odd sequence word makes every request retry that slot 1,000 times, increment/rewrite its own slot, and synchronously emit this warning until supervisor reclaim. A crash at multi-thousand-QPS rate can therefore amplify itself on every surviving event loop.

Cache a fail-closed unstable-slot state until generation/reclaim/version changes, keep counting rejected requests, and log only the state transition or at a bounded rate.

sglang_omni_router/health.py
```diff
@@ -82,6 +87,8 @@ async def _check_worker_health(self, worker: Worker) -> None:
                 timeout=self._config.health_check_timeout_secs,
             )
         except httpx.HTTPError as exc:
+            if worker.is_dead:
```
Ratish1: The new dead check closes the direct late-result race, but a reachable ABA remains: probe A starts, the operator marks the worker dead, then clears dead and probe B succeeds; if old probe A fails afterward, is_dead is false again and A demotes the newly healthy worker. 

Capture a health/state epoch when the probe starts, increment it on mark-dead/clear-dead, and apply the result only if the epoch still matches.

sglang_omni_router/data_plane.py
```diff
@@ -0,0 +1,611 @@
...
+            if entry.capabilities:
+                config_kwargs["capabilities"] = set(entry.capabilities)
+            config = WorkerConfig(**config_kwargs)
+            worker = self._workers.get(entry.url)
+            incarnation_changed = (
+                worker is not None
+                and entry.incarnation
+                and worker.incarnation != entry.incarnation
+            )
+            if worker is None or incarnation_changed:
```
Ratish1, line 116: Replacing the Worker object correctly prevents a late incarnation-A failure from mutating B, but an in-flight A request still records completion on the retired object. Counter flushes enumerate only view.workers(), so once B replaces A that completion is never sent to the CP ledger. 

Keep retired incarnations reportable until active_requests == 0 and a final acknowledged flush, or decouple counter ownership from the routable view. Add DELETE/re-add of the same URL while an A request is in flight, with A completing after B is reported.

sglang_omni_router/data_plane.py
```diff
@@ -0,0 +1,611 @@
...
+            if key.lower() not in _FORWARD_RESPONSE_STRIP
+        }
+        return Response(
+            content=upstream.content,
+            status_code=upstream.status_code,
+            headers=response_headers,
+        )
+
+    for path, methods in _FORWARDED_CP_ROUTES:
+        app.api_route(path, methods=list(methods))(_forward)
```
Ratish1: Registering one untyped _forward function for this table changes N>=2 OpenAPI: /workers/{worker_path} has no declared path parameter, GET/PUT/DELETE share an operation ID, the public parameter name differs from N=1's worker_id, and auth/response schemas and summaries disappear. Generated clients and schema-based tooling therefore break when only --router-processes changes.

Share typed route definitions/wrappers with the CP or explicitly publish the canonical schema.

docs/basic_usage/omni_router.md
```diff
@@ -456,3 +456,55 @@ point with:
 ```bash
 python -m sglang_omni_router.serve --help
 ```
+
+## Multi-Process Router (Control/Data-Plane Split)
+
+> Linux only. Enabled with `--router-processes N`; the default `1` keeps the
```
estellaliu233: Nit, non-blocking: Router-processes is the entry point for this feature. It's missing from the Router Arguments table above. Should we also add it to the table?

estellaliu233 (commented):
> Left one small documentation suggestion.

sglang_omni_router/data_plane.py
JiaxinD: Forwarded routes now reuse the single-process route names and the worker_id path parameter, and a test pins that both schemas agree on shared operations. Left alone: /weights_checker already shares one operation id for GET and POST in the single-process app, so N>=2 inherits that rather than diverging.

sglang_omni_router/app.py
JiaxinD: The journal is out of tmp: it now resolves under a durable state directory (`--router-state-dir`, else `SGLANG_OMNI_ROUTER_STATE_DIR`/`$XDG_STATE_HOME/sglang-omni-router`), so it survives a host reboot the remote workers outlive; startup fails closed if that directory is unusable, with no temp fallback.
Per-run runtime files (config, internal socket, snapshot, admission shm) stay in the temporary workdir on purpose. Still no separate recovery endpoint, so leaving this open for your call.

sglang_omni_router/update_journal.py
```diff
@@ -0,0 +1,267 @@
...
+            return
+        except OSError as exc:
+            raise JournalUnwritableError(f"{self._path}: {exc}") from exc
+        self._fsync_dir()
+
+    def _write(self, data: dict) -> None:
+        directory = os.path.dirname(self._path) or "."
+        tmp_path = f"{self._path}.tmp"
+        try:
+            _make_private_dir(directory)
```
Ratish1: On the first transaction, this call creates state_dir/<endpoint-key>/, but _write() later fsyncs only the journal file and <endpoint-key>/. It never fsyncs state_dir, whose directory entry is what makes the new endpoint directory survive a host crash. The sequence journal.begin() succeeds → workers receive different weights → router host crashes can therefore restart with the entire endpoint directory and journal missing, re-enabling an uncertain pool.

When _make_private_dir(directory) creates the endpoint directory, fsync state_dir before journal.begin() can report success. Do this only when the directory entry is newly created, not on every journal rewrite.

sglang_omni_router/data_plane.py
```diff
@@ -0,0 +1,649 @@
...
+        INTERNAL_TCP_URL_ENV,
+        INTERNAL_UDS_ENV,
+        SNAPSHOT_PATH_ENV,
+    )
+
+    config = load_config_from_env()
+    token = os.environ.get(INTERNAL_TOKEN_ENV)
+    timeout = httpx.Timeout(config.request_timeout_secs)
+    uds = os.environ.get(INTERNAL_UDS_ENV)
+    internal_limits = httpx.Limits(max_connections=8, max_keepalive_connections=4)
```
Ratish1: This eight-connection client is used both by public CP forwarding and by registration, heartbeat, snapshot ACK, counters, and worker-failure reports. Eight concurrent slow forwarded admin operations landing on one DP can occupy every connection. In particular, concurrent updates can remain queued behind
the CP's 300-second admin lock while leaving no connection for that DP's heartbeat and snapshot ACK. With the 6-second liveness and 10-second ACK windows, this can falsely degrade the DP and abort an otherwise safe update.

Give registration/heartbeat/ACK/counter/failure traffic one small bounded, short-timeout client and use a separate bounded client for public forwarding. Reuse those clients for the DP lifetime, account for both in the supervisor's FD estimate, and add a saturation test proving control traffic completes while every public-forward slot is occupied.

sglang_omni_router/observability.py
```diff
@@ -0,0 +1,182 @@
...
+                and report.counter_seq <= entry.counter_seq
+            ):
+                return False
+            if report.generation > entry.generation:
+                self._retire(entry)
+                entry = None
+
+        per_worker = entry.per_worker if entry is not None else {}
+        for item in report.workers:
+            ledger = per_worker.get(item.worker_id)
```
Ratish1: The DP payload lists the current worker first and any draining retired object afterward. During replacement, one report therefore contains both (worker_id, incarnation B) and (worker_id, incarnation A). This loop stores only one ledger per worker ID, so each row treats the preceding row as a replacement and retires it again. I reproduced totals 0 -> 3 -> 5 for only three actual requests when the reports were A, then [B, A], then [B].

Keep active segments by (worker_id, incarnation) and aggregate them by stable worker ID. Because each report is a full snapshot for that DP, fold a segment absent from the next report into the stable worker's retired accumulator and delete it; retaining every historical incarnation would make memory and /workers aggregation grow without bound.

sglang_omni_router/control_plane.py
```diff
@@ -0,0 +1,447 @@
...
+                # Note (Jiaxin Deng): momentarily unstable shm (fold in
+                # flight); report a read error, do not gate readiness on it.
+                admission_error = str(exc)
+        ready_indices = {
+            record.dp_index
+            for record in live_dps
+            if record.last_applied_epoch == cp_epoch
+            and record.serving
+            and (
+                admission_view is None
```
Ratish1, line 350: A DP killed during its slot write leaves an odd slot until supervisor reclaim. Every surviving DP's try_acquire() then fails closed because it cannot sum that sibling, so the data plane returns 503 for every model request. This branch nevertheless ignores the admission error and counts all live heartbeat-serving records as ready. 

Keep the existing read fail-fast, then perform only a brief bounded asynchronous recheck/debounce so a normally live writer preempted mid-write cannot flap readiness. If the slot remains unreadable, report zero
serving-ready capacity and 503 for that health evaluation; the next probe can recover immediately after reclaim. Do not add a shared lock, blocking sleep, or longer synchronous spin.

docs/basic_usage/omni_router.md
```diff
@@ -456,3 +458,96 @@ point with:
...
+- The supervisor restarts crashed children (with a fail-closed budget for
+  rapid crash loops), fences replaced processes by generation, and tears the
+  tree down in order on SIGTERM.
+
+Behavior differences to know about:
+
+- **The admission bound is shared and soft.** The in-flight bound spans all
+  DPs through a shared-memory counter array. Concurrent admission checks can
+  overshoot the bound by at most `N - 1` requests. In `/health`, `inflight` is
+  an instantaneous sum (not a linearizable value), `rejected_total` is exact,
```
Ratish1: This public guide still promises rejected_total is exact. State that it is exact while live slots remain readable and best-effort across DP crash/reclaim windows.

tests/unit_test/router/test_control_plane.py
```diff
@@ -0,0 +1,1021 @@
...
+            "/internal/register", json={"dp_index": 0, "generation": 1, "pid": 1}
+        )
+        # Note (Jiaxin Deng): ack everything the CP will ever publish in this test
+        client.post(
+            "/internal/heartbeat",
+            json={
+                "dp_index": 0,
+                "generation": 1,
+                "pid": 1,
+                "last_applied_seq": 10_000,
```
Ratish1: A DP cannot legitimately report sequence 10,000 before the CP publishes the disabled-worker snapshot, but this test does exactly that. The CP trusts the future value, so the update never proves that the publication was observed and acknowledged; the same shortcut is reused by _ack_everything().

Start the update, wait until the disabled snapshot's actual sequence is visible, heartbeat that exact epoch/sequence from each DP, and then assert the broadcast begins.

sglang_omni_router/update_journal.py
JiaxinD: Fixed: `_make_private_dir` now reports what it created and `_write` fsyncs each new directory's parent before the write can report success, only on creation.
A test pins that the state dir's inode is synced on the first `begin()` and not on a rewrite.

sglang_omni_router/data_plane.py
JiaxinD: Split: control traffic (register, heartbeat, ACK, counters, failure) now owns a 4 connection pool with a 5s timeout, forwarding keeps a separate 8 connection pool, both reused for the DP lifetime and both counted in the supervisor fd log.
Added a saturation test that pins the heartbeat landing while every forwarding slot is occupied.

sglang_omni_router/observability.py
JiaxinD: Fixed: segments are keyed by (worker_id, incarnation) and aggregated by stable id, and a segment absent from the next full snapshot is folded into the retired accumulator and dropped, so the map stays bounded.
Two tests, one replaying your A, [B, A], [B] sequence.

sglang_omni_router/control_plane.py
JiaxinD: Fixed: the read stays fail fast with a bounded async debounce (3 reads, 10ms apart); a slot still unreadable after it reports zero serving ready and 503, and the next probe recovers after reclaim. No lock, no blocking sleep, no longer spin.
Note: the retired fold slot is not read by `try_acquire`, so a stalled fold still only costs the aggregate numbers; both cases now have their own test.

docs/basic_usage/omni_router.md
JiaxinD: Done: documented as exact while live slots stay readable, best-effort across a DP crash and reclaim window.

tests/unit_test/router/test_control_plane.py
JiaxinD: Fixed: the test now waits for the disabled snapshot's real seq, asserts the broadcast has not started, then acks that exact seq.
The pre-ack helper is renamed `_preack_future_seq` and kept only where the assertion is a different failure reason.

sglang_omni_router/app.py
JiaxinD: Added it: `POST /weight_update_journal/resolve`, admin authenticated, requires `{"acknowledge": true}`, 409 while an update holds or is queued for the lock, reports the ids it dropped, and 503 rather than success if the file survives.
Documented next to `--router-state-dir`; the affected workers stay disabled until re-enabled one by one.

sglang_omni_router/data_plane.py
```diff
@@ -0,0 +1,697 @@
...
+                        "request body exceeds max_payload_size "
+                        f"({config.max_payload_size} bytes)"
+                    ),
+                    "type": "payload_too_large",
+                    "code": 413,
+                }
+            },
+        )
+
+    async def _forward(request: Request) -> Response:
```
AkazaAkane, line 591: This buffers the complete public admin request before the CP authenticates it. With the default 512 MiB payload limit and no concurrency/admission bound on this path, concurrent unauthorized requests can exhaust DP memory without reaching an authenticated handler. bytearray followed by bytes() can also temporarily retain two copies.

Please either validate the admin bearer token in the DP before consuming the body—while retaining the CP re-check—or give forwarded admin traffic a small dedicated body limit plus a hard concurrency bound or streaming relay.

sglang_omni_router/admission_shm.py
```diff
@@ -0,0 +1,419 @@
...
+import time
+from dataclasses import dataclass
+from typing import Callable
+
+logger = logging.getLogger("sglang_omni_router.admission_shm")
+
+_PLATFORM_WARNED = False
+
+
+def _warn_if_unverified_platform() -> None:
```
AkazaAkane, line 58: The shared-memory protocol states that correctness relies on x86-64 TSO and that the sequence word is not a native atomic, but non-x86 platforms only receive a warning and continue running an unsupported synchronization algorithm.

Please reject --router-processes >= 2 at CLI/supervisor startup unless the platform is x86-64, or replace this with actual cross-process atomics. The CLI and documentation should say “x86-64 Linux only,” rather than only “Linux only.” Please add a startup-validation test with a mocked ARM machine value.

sglang_omni_router/serve.py
```diff
@@ -274,6 +307,34 @@ def main(argv: Sequence[str] | None = None) -> None:
             config = build_config_from_args(args)
 
         check_file_descriptor_limit(config, strict=args.strict_limits)
+        # Note (Jiaxin Deng): fail closed here rather than on the first weight
+        # update, which would discover the bad path with the pool disabled.
+        state_dir = ensure_state_dir(resolve_state_dir(config.router_state_dir))
```
AkazaAkane: This makes the default --router-processes 1 path depend on a new writable durable-state volume because state-directory validation happens before the process-count branch. Read-only or home-less containers that could run the base router now exit during startup, even if no weight-update endpoint is ever used.

Please either initialize the journal lazily for N=1 and fail the weight-update operation rather than router startup, or explicitly treat this as a breaking deployment requirement and update the compatibility claim and tests accordingly.

sglang_omni_router/proxy.py
```diff
@@ -384,6 +407,30 @@ async def _forward_relay(
         worker.increment_active()
         try:
             upstream = await self._client.send(upstream_request, stream=True)
```
AkazaAkane, line 409: active_requests is not balanced when the task is cancelled while awaiting upstream response headers. increment_active() runs before client.send(), but CancelledError bypasses both HTTPError handlers; the outer layer releases admission only.

In multiprocess mode, this can also keep a retired worker incarnation reportable forever because retirement waits for active_requests == 0. Please catch cancellation/BaseException around send(), decrement and close any partially created response as appropriate, then re-raise.

sglang_omni_router/data_plane.py
JiaxinD: Bounded on its own terms rather than by `--max-payload-size`: admin bodies now cap at 1MiB and concurrent forwards are bounded, with the excess rejected 503, so an unauthenticated caller cannot hold N x 512MiB. Auth stays on the CP.
Two tests: the oversize body is refused before buffering (and a body above `--max-payload-size` but under the admin cap still forwards), and the concurrency bound sheds.

sglang_omni_router/admission_shm.py
JiaxinD: Now refused, not warned: the supervisor rejects `--router-processes >= 2` unless the machine is x86-64, and the CLI help and docs say x86-64 Linux.
Tests cover aarch64, arm64, riscv64 and an empty machine string, plus N=1 still allowed there.

sglang_omni_router/serve.py
JiaxinD: You are right, that was a regression: N=1 predates the journal and must start read-only or home-less. Startup now only fails for `N >= 2`, where serving weight updates is the point; N=1 logs a warning and the update itself refuses with 503.
Also found while testing this: POSIX reports ENOTDIR when an ancestor of the journal path is a file, and `pending()` read that as unreadable, so such a router disabled its whole pool at startup. Fixed, both paths have tests.

sglang_omni_router/proxy.py
JiaxinD: Fixed: the send is wrapped so cancellation decrements and re-raises. A test drives a cancelling client and pins `active_requests` back to 0.
Nothing partial to close there, `send` either returns the response or raises; the streaming half already cleans up in its `finally`.

sglang_omni_router/data_plane.py
```diff
@@ -0,0 +1,729 @@
...
+                        logger.critical(
+                            "registration persistently rejected; failing closed"
+                        )
+                        on_fenced()
+                        return
+            except httpx.HTTPError:
+                # Note (Jiaxin Deng): CP unreachable; the stale-snapshot gate
+                # governs serving, keep retrying until the CP is back.
+                registered = False
+            await asyncio.sleep(heartbeat_interval_secs)
```
Ratish1: `refresh_loop()` can apply the disabled-worker snapshot within 200 ms, but the CP learns last_applied_seq only through this periodic heartbeat.  published sequence 2 immediately after a heartbeat while polling the snapshot every 10 ms; the DP applied it promptly, but the CP-visible ACK arrived exactly 2.000 seconds later. With several independently phased DPs, the barrier waits for the slowest beat, so a frequent RL weight update can spend nearly two seconds with the pool disabled before the worker broadcast even starts.

Separate update acknowledgement from the liveness cadence: have a successful snapshot apply wake the heartbeat loop (or send a small explicit ACK) immediately, while retaining the periodic heartbeat as the idle backstop. This removes synchronization latency without adding work to model requests. Add a interval regression that publishes just after a heartbeat and requires the new sequence to be reported within roughly one refresh interval, not one heartbeat interval.

sglang_omni_router/serve.py
```diff
@@ -274,6 +307,45 @@ def main(argv: Sequence[str] | None = None) -> None:
...
+            # file, so the admin key and log level travel via the environment.
+            if args.admin_api_key:
+                os.environ["SGLANG_OMNI_ADMIN_KEY"] = args.admin_api_key
+            os.environ["SGLANG_OMNI_ROUTER_LOG_LEVEL"] = log_level
+            logger.info(
+                f"Starting SGLang-Omni Router (multi-process) on "
+                f"{config.host}:{config.port} with "
+                f"{args.router_processes} data planes"
+            )
+            supervisor = RouterSupervisor(
```
Ratish1, line 364: RouterSupervisor(...) is the first place that rejects max_inflight < router_processes, unsupported architectures, and N>=2 with least_request. With --launcher-config, however, launch_and_wait() has
already started and waited for the real model workers. A deterministic invalid invocation can therefore allocate GPUs and finish model startup only to be rejected and shut down afterward.

Run these deterministic checks before launch_and_wait(), while retaining the supervisor-side validation for direct callers. Add a launcher test asserting that an invalid multiprocess configuration starts no worker
process.

sglang_omni_router/data_plane.py
```diff
@@ -0,0 +1,729 @@
...
+        )
+
+    async def _forward_worker(request: Request, worker_id: str) -> Response:
+        # Note (Jiaxin Deng): worker_id is unused; it exists so the published
+        # schema names the path parameter as the single-process routes do.
+        return await _forward(request)
+
+    for path, methods, name in _FORWARDED_CP_ROUTES:
+        handler = _forward_worker if "{worker_id" in path else _forward
+        app.add_api_route(path, handler, methods=list(methods), name=name)
```
Ratish1: Reusing route names fixed part of the schema drift, but the generic DP handler still has no auth dependency or declared security metadata.  /update_weights_from_disk: N=1 publishes an optional Authorization header, while N>=2 publishes neither that parameter nor asecurity requirement, although the CP requires the same bearer key at runtime. Generated clients and API tooling therefore change when only --router-processes changes.

Define one HTTP bearer security scheme and apply the same operation security requirement to authenticated N=1 and N>=2 routes. Extend schema parity to compare security and parameters, not only route names and path-parameter identity. .

sglang_omni_router/update_journal.py
```diff
@@ -0,0 +1,361 @@
...
+            # progress (fail closed).
+            return True
+
+    def begin(self, path: str, worker_ids: list[str]) -> None:
+        self._write({"path": path, "worker_ids": sorted(worker_ids)})
+
+    def keep(self, worker_ids: list[str]) -> None:
+        """Persist the still-unresolved target set (empty clears)."""
+        if worker_ids:
+            self._write({"worker_ids": sorted(worker_ids)})
```
Ratish1: begin() records the admin path, but the first partial/unknown outcome rewrites the document with only worker_ids. Recovery remains fail-closed, so this is not a weight-safety defect, but it discards the only durable clue distinguishing a disk reload, distributed update, or pause operation during incident response.

Read and retain the existing path when narrowing the unresolved target set, and cover begin -> keep -> pending document in the journal tests.

docs/basic_usage/omni_router.md
```diff
@@ -456,3 +458,115 @@ point with:
...
+Resolution order:
+
+1. `--router-state-dir`
+2. `SGLANG_OMNI_ROUTER_STATE_DIR`
+3. `$XDG_STATE_HOME/sglang-omni-router`
+4. `~/.local/state/sglang-omni-router`
+
+The directory is created owner-only (`0700`, journal file `0600`) and is keyed
+by the router's `host:port`, so several routers on one machine keep separate
+journals. Startup fails if the directory cannot be created or written: there is
```
Ratish1: This now says startup always fails, but the latest code deliberately lets N=1 start with UnavailableJournal and rejects only a later weight update; N>=2 still fails startup. The nearby statement that N=1 is “unchanged” is also too strong because N=1 now performs journal recovery and can keep uncertain workers disabled.

State the two modes explicitly so operators know whether an unusable durable state path prevents startup or defers the failure until an update request.

sglang_omni_router/data_plane.py
JiaxinD: Fixed: a snapshot apply that advances the seq now wakes the heartbeat loop immediately, with the periodic beat kept as the idle backstop. Test pins the ACK arriving promptly with the heartbeat period set to 600s.

sglang_omni_router/serve.py
JiaxinD: Fixed: the deterministic checks now run from the CLI args before launch_and_wait, with a test asserting an invalid multiprocess invocation starts no worker. The inflight bound pre-checks only when set explicitly (the derived default needs the worker count), so the supervisor still re-validates it.

sglang_omni_router/data_plane.py
JiaxinD: Fixed: forwarded admin routes now declare the same optional Authorization parameter as N=1 when an admin key is configured, and the parity test compares parameters and security too. Kept the parameter form instead of a new bearer scheme so the N=1 published schema (shared admin_auth helper, also used by the engine server) stays unchanged.

sglang_omni_router/update_journal.py
JiaxinD: Fixed: keep() now reads and retains the recorded admin path when narrowing the target set; the begin, keep, pending document shape is pinned in the journal tests.

docs/basic_usage/omni_router.md
JiaxinD: Fixed: the section now states both modes explicitly (N>=2 fails startup, N=1 starts with the journal disabled and refuses updates with 503), and the multiprocess note no longer claims N=1 is unchanged.

sglang_omni_router/supervisor.py
```diff
@@ -0,0 +1,635 @@
...
+LOG_LEVEL_ENV = "SGLANG_OMNI_ROUTER_LOG_LEVEL"
+
+# Note (Jiaxin Deng): the shared admission seqlock relies on x86-64 store
+# ordering, so anywhere else the protocol is simply unvalidated; refuse rather
+# than run it and hope.
+SUPPORTED_MACHINES = ("x86_64", "amd64")
+
+_LISTEN_BACKLOG = 2048
+_SHUTDOWN_GRACE_SECS = 10.0
+CHILD_GRACEFUL_SHUTDOWN_SECS = 5
```
Ratish1, line 79: This constant is passed to every DP as Uvicorn's timeout_graceful_shutdown, so an ordinary supervisor SIGTERM or rolling deployment waits only five seconds and then cancels all remaining serving
tasks. TTS/Omni generations last longer than five seconds, and the router itself permits requests up to 1800 seconds by default. Those clients receive a truncated stream or disconnect during a routine shutdown, whereas N=1 leaves Uvicorn's graceful timeout unset.

Separate normal draining from fencing: make the normal DP drain timeout deployment-configurable, use a supervisor wait budget longer than that timeout, and preserve the current short hard-exit deadline only for the orphan watcher.

JiaxinD: Fixed: the DP drain is now --shutdown-drain-secs (default --request-timeout-secs, matching N=1 where uvicorn has no graceful deadline), the supervisor waits out that drain before SIGKILL, and the orphan watcher keeps its short hard exit. Tests pin the uvicorn deadline, the supervisor wait budget, and the CLI default.

Ratish1 (approved):
LGTM

PR #1050, opened 2026-07-14
[TTS Refactor] Declarative wire serialization for TTS pipeline states

luojiaxuan:
## What

Slimming slice: replace the six hand-written `to_dict`/`from_dict` pairs (313 lines across higgs / qwen3-tts / fish / moss / moss-local / voxtral) with `DeclarativeStateBase`, which derives both methods from field metadata. Models now declare wire behavior inline:

- `wire(...)` fields carry an **emit rule** — `always` / `not_none` / `truthy` — inferred by default (None-defaulted fields emit only when set). Higgs keeps its `sample_rate` / `audio_samples` condition model-local.
- and a **codec** — `int`/`int_or`/`opt_int`/`float`/`str`/`str_or`/`bool`/`dict`/`list` coercions, `tensor_cpu` (native tensor, detached to CPU), `tensor_list` (flatten, stays list), `tensor_restore` (flatten, restore to tensor), `tensor_items` (element-wise), `typed_tensor` (the `{name}_bytes/_shape/_dtype` triple).

`PipelineStateBase` keeps its `NotImplementedError` contract for non-declarative subclasses (qwen3-omni / llada2 keep their custom impls; they can migrate later).

Behavior class: refactor, output-equivalent. Net **−116 non-test** lines (+282/−398).

## Wire-format proof

1. **Byte-identical dump**: constructed all six states with rich payloads *and* default-only payloads, dumped normalized `to_dict()` on main vs this branch — `diff` clean (12/12 cases).
2. **Contract test**: `test_tts_pipeline_state_round_trips_preserve_payload_fields` (hardened in #1019) pins the wire layout via dict comparison *and* every restored attribute field-completely per model.
3. Local runs: scheduling + capabilities suites green (33 passed); moss-local pipeline suite matches the main baseline exactly. Full suite gate: PR unit-test CI.

## Notes for review

- The `*_or` codec variants preserve each model's existing `or default` decode semantics (e.g. qwen3 `task_type`/`language`, moss `sample_rate`); plain `int`/`str` preserve the strict variants. These map 1:1 from the deleted hand-written code.
- Emission order inside the dict changed (usage fields now always last); payload dicts are order-insensitive (msgpack maps), and the contract test compares by key set + values.
- Ming-TTS (#858) can adopt the same base in a follow-up once it lands.

Part of #985. Refs #661 (per-request pipeline-state serialization listed as a 2–5× duplicated pattern).


sglang_omni/scheduling/pipeline_state.py
```diff
@@ -55,6 +60,196 @@ def append_usage_fields(self, data: dict[str, Any]) -> None:
...
+    def from_dict(cls: type[StateT], data: Any) -> StateT:
+        if not isinstance(data, dict):
+            data = {}
+        kwargs: dict[str, Any] = {}
+        for f in dataclasses.fields(cls):
+            spec = _spec_of(f)
+            if spec.codec == "typed_tensor":
+                from sglang_omni.scheduling.typed_tensor import decode_typed_tensor
+
+                kwargs[f.name] = decode_typed_tensor(
```
SandyLuXY: `typed_tensor` takes a different missing-key path from every other codec: `decode_typed_tensor({})` returns `None`, and that gets passed to the constructor, replacing any dataclass default. Skip this kwarg when no legacy or encoded key is present, and cover a non-`None` default in the test.

sglang_omni/scheduling/pipeline_state.py
```diff
@@ -55,6 +60,196 @@ def append_usage_fields(self, data: dict[str, Any]) -> None:
...
+    emit: str | None = None,
+    codec: str = "raw",
+) -> Any:
+    """dataclasses.field carrying wire metadata for DeclarativeStateBase.
+
+    `emit` defaults by inference: fields whose default is None emit only when
+    not None; everything else always emits. `codec="typed_tensor"` expands to
+    the {name}_bytes/_shape/_dtype key triple via scheduling.typed_tensor.
+    """
+    if codec != "typed_tensor" and codec not in _CODECS:
```
SandyLuXY, line 153: `emit` accepts arbitrary strings, so a typo falls through as `always` and silently changes the wire format. Validate the mode in `wire()`, validate `with:` anchors once the dataclass fields are available, and add typo and empty-anchor tests.

sglang_omni/scheduling/pipeline_state.py
```diff
@@ -55,6 +60,196 @@ def append_usage_fields(self, data: dict[str, Any]) -> None:
...
+    *,
+    default_factory: Any = MISSING,
+    emit: str | None = None,
+    codec: str = "raw",
+) -> Any:
+    """dataclasses.field carrying wire metadata for DeclarativeStateBase.
+
+    `emit` defaults by inference: fields whose default is None emit only when
+    not None; everything else always emits. `codec="typed_tensor"` expands to
+    the {name}_bytes/_shape/_dtype key triple via scheduling.typed_tensor.
```
SandyLuXY: nit: avoid using `` in doc string.

sglang_omni/scheduling/pipeline_state.py
```diff
@@ -55,6 +60,196 @@ def append_usage_fields(self, data: dict[str, Any]) -> None:
...
+    if f.default is not MISSING and f.default is None:
+        return "not_none"
+    return "always"
+
+
+@dataclass
+class DeclarativeStateBase(PipelineStateBase):
+    """PipelineStateBase with to_dict/from_dict derived from field metadata.
+
+    Subclasses declare wire behavior inline with `wire(...)` fields instead of
```
SandyLuXY: Same as above, avoid using `` in comment

sglang_omni/scheduling/pipeline_state.py
```diff
@@ -55,6 +60,196 @@ def append_usage_fields(self, data: dict[str, Any]) -> None:
...
+    hand-writing the serialization pair; plain fields default to
+    always-emitted raw passthrough (None-defaulted fields emit only when set).
+    Usage fields keep the append_usage_fields contract. The field-complete
+    round-trip contract test in tests/unit_test/scheduling/test_pipeline_state.py
+    pins both the wire layout and the restored attributes per model.
+    """
+
+    def to_dict(self) -> dict[str, Any]:
```
SandyLuXY, line 215: Every `to_dict()` and `from_dict()` call walks `dataclasses.fields()`, re-reads metadata, infers emission, parses `with:` strings, looks up codecs, and rebuilds coupled-field state. These methods run at every model pipeline boundary.

Consider compiling and caching an immutable per-class field plan once, including logical name, physical codec, resolved emission mode, anchor, default behavior, and usage handling. Validate the plan once at class initialization or first use.

SandyLuXY (changes requested):

sglang_omni/scheduling/pipeline_state.py
```diff
@@ -55,6 +61,297 @@ def append_usage_fields(self, data: dict[str, Any]) -> None:
...
+        coupled: list[_FieldPlan] = []
+        for plan in _plan_for(type(self)):
+            if plan.is_usage:
+                continue
+            if plan.anchor is not None:
+                coupled.append(plan)
+                continue
+            self._encode_field(data, plan, plan.emit)
+        for plan in coupled:
+            if plan.anchor in data:
```
SandyLuXY: `to_dict()` still checks `plan.anchor in data`, so `emit="with:codes"` drops the dependent field when `codes` uses `typed_tensor` because only `codes_bytes/_shape/_dtype` are emitted; chained anchors are also declaration-order dependent. Track logical field emission separately from physical wire keys and either resolve dependencies independently of field order or reject chained/cyclic anchors, with typed-anchor and reverse-order-chain regression tests.

sglang_omni/scheduling/pipeline_state.py
```diff
@@ -55,6 +61,297 @@ def append_usage_fields(self, data: dict[str, Any]) -> None:
...
+    return "always"
+
+
+def _plan_default_factory(f: dataclasses.Field) -> Callable[[], Any] | None:
+    if f.default_factory is MISSING:  # type: ignore[misc]
+        return None
+    return f.default_factory  # type: ignore[return-value,misc]
+
+
+def _has_encoded_typed_tensor_payload(data: dict[str, Any], name: str) -> bool:
```
SandyLuXY: `_has_encoded_typed_tensor_payload()` treats any one of the bytes/shape/dtype keys as a payload, but `decode_typed_tensor()` returns `None` when bytes or shape is missing, so a partial payload such as `{"codes_shape": [1]}` silently replaces a non-`None` dataclass default.
Distinguish absent, complete, and malformed representations; fail fast on partial key sets (with dtype optionality made explicit).

sglang_omni/scheduling/pipeline_state.py
```diff
@@ -55,6 +61,297 @@ def append_usage_fields(self, data: dict[str, Any]) -> None:
...
+            encode, decode = _CODECS[spec.codec]
+        plans.append(
+            _FieldPlan(
+                name=f.name,
+                emit=emit,
+                codec=spec.codec,
+                anchor=anchor,
+                encode=encode,
+                decode=decode,
+                default_value=_default_of(f),
```
SandyLuXY: `_build_plan()` executes `default_factory` through `_default_of()` on the first `to_dict()`, and `from_dict()` calls `plan.default()` even for codecs such as `list` that ignore the value; a call-count repro advances from 1 after construction to 2 after serialization and 3 after restore.
Cache the factory callable without evaluating it and invoke it only for decoders that consume the default.

sglang_omni/scheduling/pipeline_state.py
luojiaxuan: Done.

sglang_omni/scheduling/pipeline_state.py
luojiaxuan: Done.

sglang_omni/scheduling/pipeline_state.py
luojiaxuan: Done.

sglang_omni/scheduling/pipeline_state.py
luojiaxuan: Done.

sglang_omni/scheduling/pipeline_state.py
luojiaxuan: Done.

sglang_omni/scheduling/pipeline_state.py
luojiaxuan: Done.

sglang_omni/scheduling/pipeline_state.py
luojiaxuan: Done.

sglang_omni/scheduling/pipeline_state.py
luojiaxuan: Done.

tests/unit_test/scheduling/test_pipeline_state.py
```diff
@@ -349,3 +354,134 @@ def test_typed_tensor_legacy_list_fallback_and_missing() -> None:
...
+
+    @dataclass
+    class _FactoryState(DeclarativeStateBase):
+        items: list[int] = wire(default_factory=make_items, codec="list")
+
+    state = _FactoryState()
+    assert calls == 1
+
+    payload = state.to_dict()
+    assert calls == 1
```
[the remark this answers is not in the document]
luojiaxuan: Done.

tests/unit_test/scheduling/test_pipeline_state.py
```diff
@@ -349,3 +354,134 @@ def test_typed_tensor_legacy_list_fallback_and_missing() -> None:
...
+
+    restored = _FactoryState.from_dict(payload)
+    assert calls == 1
```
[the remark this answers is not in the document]
luojiaxuan: Done.

sglang_omni/scheduling/pipeline_state.py
```diff
@@ -55,6 +62,336 @@ def append_usage_fields(self, data: dict[str, Any]) -> None:
...
+@dataclass(frozen=True)
+class _WireSpec:
+    emit: str | None = None  # always | not_none | truthy | with:<field>
+    codec: str = "raw"
+
+
+_DEFAULT_SPEC = _WireSpec()
+
+
+@dataclass(frozen=True)
```
Hayden727: The compiled `_FieldPlan` cache and anchor dependency graph are much broader than the current contract: production has a single Higgs `sample_rate` → `audio_samples` coupling. Please iterate the dataclass fields directly and keep that one conditional emission model-local, removing the plan cache, recursive graph traversal, and cycle machinery.

luojiaxuan: One model-specific condition does not justify a shared dependency graph.

SandyLuXY (approved):

Issue #925, opened 2026-07-02
[TTS Refactor][M4] ReferenceEncodeService: shared ad-hoc reference cache and single-flight


luojiaxuan:
## Design status

M4a is the production refactor for ad-hoc reference-audio cache + same-key single-flight. M4b is a profile-gated follow-up for different-key batch coalescing and should not land unless M4a profiling shows reference encode remains a meaningful bottleneck.

Design links:
- Lark design: https://osgbw74w8zwb.sg.larksuite.com/docx/PM8Ed39XAoHkLoxguvHlFiO5ghc
- Repo API doc: `docs/developer_reference/reference_encode_service.md`
- Historical context: #661 and closed #809

## Open decisions

- [ ] Confirm final model/checkpoint revision key source per migrated model.
- [x] Confirm PR3 MOSS-local M4a cleanup stays below the ~1000 LOC review budget. PR3 is net-reducing and changes 340 lines across 2 files.
- [ ] Decide M4b after PR3 profiling: implement shared batch coalescer, defer, or keep MOSS-local model-owned batching. Current status: deferred pending profiling data.

## Task checklist

- [ ] PR 1: `[TTS Refactor][M4a] ReferenceEncodeService: base + FishAudio S2-Pro migration` — #926
- [ ] PR 2: `[TTS Refactor][M4a] ReferenceEncodeService: Qwen3-TTS taxonomy migration` — #927
- [ ] PR 3: `[TTS Refactor][M4a] ReferenceEncodeService: MOSS-local cache cleanup` — #930
- [ ] PR 4: `[TTS Refactor][M4b] ReferenceEncodeService: MOSS-local batch coalescing` or explicit profile-based deferral — deferred pending PR3 profiling

## PR links

- PR 1: #926
- PR 2: #927
- PR 3: #930
- PR 4 / M4b decision: Deferred until PR3 profiling proves M4a still leaves reference encode as a bottleneck and shared batch coalescing has measured speedup.

db-ol:
Sharing two measurements from probing reference-audio reuse on the serving paths, in case they are useful for the ReferenceEncodeService design.

Setup: novita-h100, fresh server per run. TTS probe: MOSS-TTS-Local, 200 voice-clone requests at concurrency 16, one group sharing a single reference clip vs one group with per-sample references. ASR probe: Qwen3-ASR behind the CI router topology, 1088 requests over 666 unique clips (422 repeats), comparing per-clip latency between first occurrences and repeats.

[...]

PR #1051, opened 2026-07-14
[TTS Refactor][M4a] Higgs: route ad-hoc reference codes through ReferenceEncodeService

luojiaxuan:
## What

Migrates the last model off hand-rolled reference-code caching: the Higgs `audio_encoder` stage's `StageOutputCache` choreography is replaced by the shared `ReferenceEncodeService` with a Higgs hook. This closes the M4a migration checklist in #925 — fish (#926), qwen3-tts (#927), moss-local (#930), and now higgs all ride the shared service.

- Keys are the **waveform-content hashes already computed at preprocessing time** (`state.reference_code_cache_key`), so identical reference audio hits across request ids and source forms (path / bytes / base64). Content keys cannot go stale, so the default revalidate applies.
- **Uploaded voices stay on the versioned speaker cache** (invalidation by voice re-upload is its contract, not the content-cache's); on a speaker-cache miss the encode now flows through the service, so the content-keyed entry also warms for name-independent reuse.
- Missing content keys pass through uncached — matching the old `StageOutputCache` None-key behavior.
- Ad-hoc references gain single-flight dedup (concurrent same-content requests encode once), byte-bounded LRU, and service stats/log lines.

Behavior class: refactor; encode math untouched (`codec.encode_reference` → shape check → `apply_delay_pattern` moved verbatim into the hook). Affected model: higgs_tts only.

## Tests

- New `tests/unit_test/higgs_tts/test_reference_encode_hook.py`: same-content-key cache hit with int32-store/long-load round-trip against `apply_delay_pattern` ground truth; None-key uncacheable passthrough; codec shape-mismatch fail-loud.
- Shared service mechanics (single-flight, LRU, no-poison) are covered by `scheduling/test_reference_encoder.py`.
- Gate: TTS CI runs higgs end-to-end when selected; PR unit-test CI runs the new hook tests.

## Notes

This is a consistency/mechanics slice, not a big deletion (+161/−21): most of the higgs cache region is uploaded-voice business logic that stays. The deletion-oriented slices are #1050 (serialization base) and the pending MOSS-local batcher decision.

Part of #985. Refs #925, #661.


sglang_omni/models/higgs_tts/stages.py
```diff
@@ -341,32 +407,29 @@ def _encode(payload: StagePayload) -> StagePayload:
...
+            speaker_cache.get(speaker_code_cache_key)
+            if speaker_code_cache_key is not None
+            else None
+        )
         if cached_delayed is not None:
             delayed_rows = cached_delayed.tolist()
         else:
-            ref_codes_TN = codec.encode_reference(waveform, sample_rate=24000).to(
-                torch.long
+            delayed = reference_service.get_or_encode(
```
SandyLuXY, line 424: tests/unit_test/higgs_tts/test_pipeline.py:406 still asserts fake_codec.calls == 2, but after this fallback the third same-content request hits ReferenceEncodeService and the actual call count is 1. Please update that test for same-content reuse and add a changed content key/version case to cover re-upload invalidation.

SandyLuXY (changes requested):

SandyLuXY (approved):

PR #1053, opened 2026-07-14
[MOSS-TTS Local] Fix split topology and batch stream transport

Hayden727:
## Motivation

The existing `split` variant assigned codec work to `cuda:1`, but all three stages still shared the `pipeline` process. In practice, one process initialized the AR engine on GPU 0 and then the codec/vocoder on GPU 1; vocoder CUDA graph capture failed. Simply separating the stages also did not work because the prepared-request handoff used a process-local queue.

The streaming path had a second bottleneck: the AR runner copied and relayed one code row per decode step. Consumer-side coalescing from #892 can combine chunks that are already queued, but it cannot remove those producer-side D2H copies and relay messages.

## What this PR changes

- Make `MossTTSLocalSplitPipelineConfig` a real two-process topology:
  - `tts_engine` on GPU 0
  - `preprocessing` and `vocoder` in a codec process on GPU 1
- Carry the small CPU prepared-request state through `StagePayload` only for the cross-process variant; the default colocated path keeps the existing process-local handoff.
- Buffer generated code rows on the AR GPU and flush at the vocoder's decode boundaries:
  - first flush follows `initial_codec_chunk_frames`
  - steady-state flush follows `stream_chunk_frames`
  - EOS force-flushes the tail
- Flatten all requests due in the same scheduler step into one `torch.stack` and one D2H copy, then relay 2-D `[frames, channels]` chunks.
- Accept both legacy 1-D rows and new 2-D row batches in the MOSS-TTS Local streaming vocoder.

The default single-GPU topology and public API are unchanged; this adds no new flag.

## Why this is different from #892

#892 drains already-arrived messages in the vocoder inbox before pumping the codec. This PR reduces how many messages and D2H copies the producer creates in the first place. The two optimizations compose.

For 16 requests at concurrency 8, AR-to-vocoder stream messages dropped from **698 to 53 (-92.4%)**.

## Performance

Environment: 2x NVIDIA H200 for the split variant, streaming voice cloning, MOSS-TTS-Local-Transformer-v1.5, `token_count=auto`.

### Isolated batching delta on the same two-GPU split topology

| Workload | Split before | This PR | QPS | Mean latency | TTFC | RTF |
| c8, 64 requests | 7.771 | 8.282 | **+6.6%** | **-6.7%** | **-10.7%** | **-6.6%** |
| c16, 64 requests | 7.047 | 8.594 | **+22.0%** | **-18.8%** | **-18.1%** | **-19.2%** |

At c16, mean codec-GPU utilization rose from **55.0% to 69.5%** (p95 89.5%, max 93%). Three additional treatment repeats measured **8.692 / 9.019 / 8.818 QPS**, all with zero failures.

### Full SeedTTS EN scale-up comparison

This comparison intentionally measures the user-visible scale-up path: main's default 1x H200 topology versus this PR's 2x H200 split topology. It is not presented as a per-GPU efficiency comparison.

| Metric | main, 1x H200 | This PR, 2x H200 | Delta |
|---|---:|---:|---:|
| Requests | 1088/1088 | 1088/1088 | no failures |
| QPS | 5.889 | 8.436 | **+43.2%** |
| Mean latency | 2.705 s | 1.885 s | **-30.3%** |
| RTF | 0.6380 | 0.4514 | **-29.2%** |
| TTFC | 1.5576 s | 1.4094 s | **-9.5%** |

Both full runs used the same SeedTTS EN 1088 set, concurrency 16, streaming references, automatic duration tokens, and seed 42. The full A/B was collected on main `73905ed4`; the branch was subsequently rebased onto `1618b7ce`, whose intervening commits do not change the MOSS runtime. An exact latest-head smoke run completed 64/64 at c16 and measured **8.954 QPS**.

## Quality

One shared Qwen3-ASR-1.7B service transcribed both 1088-output sets:

| WER | main | This PR |
|---|---:|---:|
| Corpus WER | 1.968% | 2.110% |
| Corpus WER excluding >50% sample outliers | 1.622% | 1.641% |

The raw delta is +0.142 percentage points and is concentrated in two additional high-WER stochastic outliers; excluding the documented >50% outlier class, the delta is +0.020 points. All 1088 WAV formats matched; 1083/1088 shapes matched, with five stop-frame length differences expected from changed process/batch shapes under seeded sampling.

## Validation

- Rebased on `origin/main@1618b7ce`
- `pre-commit run --files ...`: all hooks passed
- `pytest tests/unit_test/moss_tts_local/test_pipeline.py tests/unit_test/moss_tts_local/test_streaming_vocoder.py tests/unit_test/moss_tts_local/test_state_pool.py -q`: **137 passed**
- Latest-head H200 E2E smoke: **64/64**, zero failures, **8.954 QPS** at c16

Related to #661. Conceptually orthogonal to the frame-decode kernel work in #762.

### DIFF
--- a/sglang_omni/models/moss_tts_local/config.py
+++ b/sglang_omni/models/moss_tts_local/config.py
@@ -24,9 +24,14 @@ _COLOCATED_CODEC_MEM_RESERVE = 0.15
 _AR_MEM_FRACTION_STATIC = 0.85
 _REF_AUDIO_CACHE_MAX_ITEMS = 8192
 _REF_AUDIO_CACHE_MAX_BYTES = 64 * 1024 * 1024
+_DEFAULT_STREAM_CHUNK_FRAMES = 25
+_DEFAULT_INITIAL_CHUNK_FRAMES = 1
 
 
 def _stages(*, codec_device: str, colocated: bool) -> list[StageConfig]:
+    codec_process = "pipeline" if colocated else "codec"
+    tts_engine_process = "pipeline" if colocated else "tts_engine"
+    codec_gpu = 0 if colocated else 1
     tts_engine_runtime = StageRuntimeConfig(
         resources=StageResourceConfig(
             total_gpu_memory_fraction=(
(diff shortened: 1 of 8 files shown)


Hayden727:
Closing after equal-resource and restart-paired validation showed that this is not a resource-efficient serving topology. On the same 2x H200 budget, the split pipeline reached 8.803 QPS median while two independent main replicas reached 12.232 QPS aggregate (-28.0% for split), with worse latency/TTFC. I also isolated the producer batching idea on the default 1x H200 topology: c16 A/B/A three-run means were 7.217 / 7.224 / 7.108 QPS (candidate / main / candidate), so the earlier apparent gain did not reproduce across server restarts. The implementation is functionally valid, but the performance case is not strong enough to merge.

PR #1055, opened 2026-07-14
[Perf] Reuse idle vocoder lanes for MOSS-TTS Local streaming

Hayden727:
## Motivation

The default MOSS-TTS Local persistent vocoder session has a fixed CUDA-graph batch width of 16, but partitions it as 8 streaming slots plus 8 offline slots. During normal streaming-only serving, half of those persistent codec-state lanes cannot be owned by live requests.

Above concurrency 8, requests that cannot acquire a streaming slot buffer all codec rows and decode the complete utterance on finalization. Those large final flushes frequently exceed the captured vocoder frame range (`T=1..25`) and fall back to eager execution. This creates a throughput plateau while leaving already-allocated graph lanes unused.

## Modifications

- Repartition the existing B16 persistent session as 15 streaming slots plus 1 reserved offline progress slot. This does not increase CUDA-graph width, GPU count, or VRAM allocation.
- Allow mixed offline work to lease currently idle streaming slots, capped by the existing offline `max_batch_size`. Leased lanes leave the free pool and return only after a successful state reset; reset failures quarantine the lane.
- Let MOSS-TTS Local collect at most one queued codec row per request in each streaming pump. This preserves the configured initial chunk boundary under backlog while still batching rows from distinct requests.
- Keep the distinct-request collection policy disabled by default for every other scheduler.

## Related Issues

No linked issue.

## Accuracy Test

No model weights, sampling math, or codec math are changed.

- A single-request main-vs-PR streaming output is bit-identical:
  `sha256=a65fa4f37e58d9376b6f23b66432b9e0b61819c61ce4fab246af6ff74797e136`.
- Every benchmark run completed 64/64 requests successfully.
- The mean generated audio duration is identical between variants (`3.628 s`).
- Unit tests cover default session width, mixed offline borrowing, borrowed-state reset, distinct-request ordering/pushback, and first-chunk preservation.

```text
194 MOSS-TTS Local tests + 24 shared streaming-vocoder tests = 218 passed
```

## Benchmark & Profiling

Setup:

- Base commit: `1618b7ce26f5ea464e849cfcd943c0e429a711a6`
- Workload: 64 distinct voice-cloning references, streaming PCM, `seed=42`, `max_new_tokens=2048`
- Three independent runs per concurrency for both main and this PR; values below are arithmetic means

Negative deltas are better for latency, RTF, and TTFC.

| Concurrency | Main QPS | PR QPS | QPS delta | Main latency | PR latency | Latency delta | Main RTF | PR RTF | RTF delta | Main TTFC | PR TTFC | TTFC delta |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 8 | 6.359 | 6.227 | -2.1% | 1.210 s | 1.234 s | +2.0% | 0.3368 | 0.3432 | +1.9% | 0.193 s | 0.187 s | -3.4% |
| 16 | 6.769 | 7.173 | +6.0% | 2.186 s | 2.047 s | -6.4% | 0.6232 | 0.5790 | -7.1% | 1.220 s | 0.851 s | -30.3% |
| 32 | 7.089 | 8.658 | +22.1% | 3.767 s | 3.075 s | -18.4% | 1.0820 | 0.8762 | -19.0% | 2.816 s | 1.640 s | -41.8% |

The c8 result shows the low-concurrency tradeoff explicitly; the benefit starts once the original 8 streaming slots become saturated.

A paired c32 run with 100 ms `nvidia-smi` sampling showed that the gain does not come from using more hardware:

| Metric | Main | PR | Delta |
|---|---:|---:|---:|
| QPS | 7.132 | 7.988 | +12.0% |
| Mean GPU utilization | 70.4% | 66.7% | -3.7 pp |
| Mean power | 267.9 W | 257.1 W | -10.8 W |
| Mean memory | 118075 MiB | 117762 MiB | -313 MiB |

The persistent-session logs also improved from `700/705` graphed vocoder steps (99.3%) on main to `1420/1423` (99.8%) on this branch. Main recorded eager final-flush steps at `T={32,33,34,38}`; the branch only observed `T={32,33}` despite serving more measured runs in that process.

## Borrowed-Lane Safety Validation

Follow-up commit `a4906029` makes lane ownership fail-closed:

- Streaming leases are tracked explicitly, so invalid or duplicate releases are rejected.
- Offline-borrowed lanes are removed from the free pool for the full decode lifetime.
- A lane returns only after reset succeeds; reset failure quarantines it.
- If decode and cleanup both fail, the original decode error is preserved.
- A codec session with no resettable streaming state fails instead of silently reusing state.

Failure-injection tests cover decode-after-state-advance, successful cleanup and reuse, cleanup failure, quarantine, and original-error preservation. On `hyper00-omni`, the final code passed `194` MOSS-TTS Local tests plus `24` shared streaming-vocoder lifecycle tests.

A paired performance regression check compared the pre-hardening PR head `bb2d501b` with final code hash `482220b9...`: same H200 GPU 6, image, 64-request SeedTTS subset, streaming PCM, seed 42, concurrency 32, one prewarm plus three measured runs per side.

| Metric | Pre-hardening | Final | Delta |
|---|---:|---:|---:|
| QPS | 8.400 | 8.415 | +0.18% |
| Mean latency | 3.167 s | 3.159 s | -0.23% |
| Mean RTF | 0.9018 | 0.8991 | -0.30% |
| Mean TTFC | 1.680 s | 1.663 s | -1.02% |

The ranges overlap, so this is evidence of no measurable regression, not an additional performance claim. Final-code serving completed 256/256 requests with zero server errors; all 192 measured outputs matched the baseline channel count, sample width, sample rate, and frame count.

## Checklist

- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed (internal behavior only; docstrings updated).


sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -281,14 +332,15 @@ class MossTTSLocalStreamingVocoderScheduler(
     """Decode MOSS-TTS Local codec rows incrementally on the v2 codec."""
 
     _can_batch_stream_chunks = True
+    _stream_chunk_batch_distinct_requests = True
```
SandyLuXY, line 337: `_stream_chunk_batch_distinct_requests` only limits messages collected in one pump; a slot-starved request can already have several rows in `state.pending`. After it reacquires a lane, `build_step_plan()` consumes that backlog (initial=1 reproduced a 4-frame first chunk), so cap the first plan at the un-emitted request's threshold and add a slot-starvation/reacquisition test.

sglang_omni/models/moss_tts_local/streaming_vocoder.py
```diff
@@ -223,40 +240,74 @@ def step(self, slot_codes: dict[int, torch.Tensor]) -> dict[int, torch.Tensor]:
...
+        borrow_count = min(
+            max(requested_wave_size - len(reserved), 0),
+            len(self._free_stream_slots),
+        )
+        borrowed = self._free_stream_slots[-borrow_count:] if borrow_count else []
+        if borrowed:
+            del self._free_stream_slots[-borrow_count:]
+        available = reserved + borrowed
+        wave_size = min(requested_wave_size, len(available))
+        decode_succeeded = False
```
Ccyest, line 267: maybe we should leave a comment: known tradeoff in the comment ("under stream saturation, offline batches degrade to serial waves and block the pump")

Ccyest (approved):

Issue #1022, opened 2026-07-11
[Tracker][Qwen3-Omni Perf] Serving Optimization


edwingao28:
This is the concise roadmap and PR tracker for Qwen3-Omni serving performance under #1018.

Keep detailed investigation, benchmarks, gates, and negative results in #1018 or the corresponding sub-trackers; use this issue for status and ownership.

New model-local PRs use the `[Qwen3-Omni Perf]` title prefix and state `Part of #1022` in the body.

## Source issues

- RFC, design, benchmark contract, and evidence: #1018
- Transport / data-plane subtracker: #934
- Multi-node communication: #862
- Process-level stage replicas: #1307
- Shared TTS interfaces: #985 (closed)
- Profiling context: #797, #907

## Current PR map

Bold numbers are #1018 item numbers.

### Landed

- [x] **1.1** Async extend / mixed-batch cache correctness — #1027
- [x] **1.2** Thinker admission cap, configuration half — #1135
- [x] **1.3** Static hidden capture and breakable prefill CUDA graph — #1364, #1380, #1381, #1532, #1537
- [x] **1.3 follow up** Enable breakable prefill graphs for speech — #1519

- [x] **1.5** Batched multimodal embed merge — #1161
- [x] **1.6** Vectorized multimodal rotary positions — #1160
- [x] **1.7** Mixed prefill and decode chunks on the thinker — #789
- [x] **2.2** Shrink the thinker to talker stream payload — #1574 (issue https://github.com/sgl-project/sglang-omni/issues/1573)

- [x] **2.6** Per-chunk event-loop overhead — #1384
- [x] **3.1** Predictor attention with SDPA `enable_gqa` — #1164
- [x] **3.2** Talker sampling-state reuse — #1043
- [x] **3.4** Device-resident row state and feedback ownership — #1167
- [x] **4.1** Exact-shape CUDA graph replay for Code2Wav — #1101
- [x] **4.2** Bounded wait/floor batching and the batching sweep — #1126, #1217 (closed unmerged; issue #1026 resolved)
- [x] **4.4** Chunk-aligned Code2Wav dispatch with batched CUDA graph capture — #1237
- [x] **4.5** Code2Wav output-path overlap — #1567
- [x] Cache talker thinker-embed shard resolution — #1187
- [x] Default code2wav onto the thinker GPU — #1235
- [x] Process-level stage replicas — #1175, tracked in #1307 (issue #1131)

### Active

[...]

Issue #1018, opened 2026-07-10
[RFC][Qwen3-Omni Perf] Serving Optimization Roadmap


edwingao28:
## Motivation

The current evidence does not provide a trustworthy high-concurrency asynchronous text baseline. Profiling observations remain measurement candidates unless causal evidence confirms them.

The asynchronous text path had a reproduced c32 failure, which #1027 has since fixed. Separate synchronous diagnostics suggested thinker admission headroom, but they did not validate normal asynchronous serving behavior, and the asynchronous baseline they were meant to anchor has still not been rebuilt.

For `text -> text+audio`, xRT rises through c16 while time to first audio also rises. Talker and code2wav residency identifies where to profile next, not a confirmed compute hotspot.

One fixed audio fixture succeeds through c4 for `audio -> text` and `audio -> text+audio`. Higher-concurrency behavior and output quality remain unmeasured.

## Current State

**Status header, updated 2026-08-15.** The tables in this section are the founding audit at `45582ff5`. They are kept as the historical baseline and are no longer the current state of the code. Two rows are now superseded. The asynchronous c32 cache failure was fixed by #1027 and its issue is closed, so that failure no longer reproduces on main; the c32 and c48 asynchronous baseline has still not been rebuilt, which is 1.1's open half. The thinker admission cap shipped at 64 through #1135. Since this audit on 2026-07-08, eighteen further PRs have landed across all four tracks; #1022 holds the live map, and each item below carries its own status. Track 1 in particular now has a merged prefill CUDA graph, which the founding audit could not assume.

All results below are audited at main revision `45582ff5` unless marked coverage-only. All audits use `Qwen/Qwen3-Omni-30B-A3B-Instruct` on H100 with thinker TP2.

TTFT and TTFA measure first-text and first-audio latency. xRT is emitted audio duration per wall second; these runs estimate it from encoded payload size.

| Path or test | Concurrency | Result | What this does not establish |
|---|---:|---|---|
| Asynchronous `text -> text` | c32 | `out_cache_loc=None` reached `store_cache`, followed by failed requests and stage exits. | c32 is not a clean throughput baseline. The subsequent asynchronous c48 result is invalid. |
| Synchronous `text -> text` | c32, c48 | Request cap: `1321.3 -> 1710.3 tok/s` at c32; `1379.2 -> 1954.1 tok/s` at c48. | Diagnostic headroom only; the normal asynchronous path remains unvalidated. |
[...]

PR #1063, opened 2026-07-15
[Qwen3-Omni Perf] Add non-greedy sampling flags to test harness

### LINKED ISSUE #1040: [Bench][Qwen3-Omni Perf] Add non-greedy (temp>0) sampling mode to the benchmark harness
## Problem

`bench_client.py` runs every request at `temperature=0`. At temperature 0 the sampler goes down the greedy path — `top_k` becomes 1, `is_all_greedy` is true, and it just calls `torch.argmax`. The `sampling_backend` setting is never reached. So every text-throughput number we have, the concurrency-32 baseline included, was measured with argmax, and the benchmark simply can't see a change to the sampling backend.

That's why item 2.1 (thinker FlashInfer sampling) came back demoted: swapping the backend gave bit-identical greedy output and no throughput change, because the harness never runs it. It also explains why the "+5–6%" we saw in an earlier environment never showed up again on clean main.

## What to add

A way to run the benchmark with real sampling:

- [ ] A flag to set `temperature`, `top_p`, `top_k`, `min_p` per run. Default stays greedy so existing numbers don't move.
- [ ] One agreed non-greedy profile for sampling A/Bs (for example `temperature 0.7`, `top_p 0.8`, `top_k 20`).
- [ ] Record the sampling settings with the other per-run metadata that #1018 already requires.

## Priority

Low. This is a tooling gap, not on any current critical path. It's a prerequisite for ever revisiting 2.1, which is itself waiting on 1.1/A12. No model changes.

Part of #1022. Technical basis: #1018 (Benchmark Contract).

WilsonZheng0327:
## Motivation

The rollout stress harness only exposes `--temperature`; `top_p`/`top_k`/`min_p` are never sent. At `temperature=0` the sampler takes the greedy `torch.argmax` path and never invokes the configured `sampling_backend`, so the harness cannot measure sampling-backend changes (the negative evidence recorded for item 2.1 in #1022). This adds a supported non-greedy sampling mode for A/B runs.

## Modifications

- Add `--top-p` / `--top-k` / `--min-p` flags to `benchmark_omni_rollout_stress.py`. Unset flags are omitted from the request payload, leaving server defaults in effect.
- Non-greedy runs (`temperature != 0`) with none of the three flags set fall back to the agreed non-greedy A/B profile: `top_p 0.8, top_k 20, min_p 0`. Setting any flag sends exactly what was set.
- Greedy runs (`--temperature 0`) never fill in the profile — the payload carries only `temperature`, byte-identical to prior greedy baselines.
- Record `temperature` / `top_p` / `top_k` / `min_p` in the results-JSON `config` block, per the #1018 metadata contract.

## Related Issues

Closes #1040. Part of #1022.

## Checklist

- [x] A flag to set `temperature`, `top_p`, `top_k`, `min_p` per run. Default stays greedy so existing numbers don't move.
- [x] One agreed non-greedy profile for sampling A/Bs (for example `temperature 0.7`, `top_p 0.8`, `top_k 20`).
- [x] Record the sampling settings with the other per-run metadata that #1018 already requires.

### DIFF
--- a/benchmarks/eval/benchmark_omni_rollout_stress.py
+++ b/benchmarks/eval/benchmark_omni_rollout_stress.py
@@ -21,6 +21,15 @@ with audio output enabled; pass ``--text-only`` for a text-only prompt or
 ``--prompt-override`` to supply the text. Results are written to
 ``<output-dir>/rollout_stress_results.json``; pass ``--no-profile`` to skip the
 server-side event profiler.
+
+Sampling: ``--temperature`` is always sent (default 0.8). For non-greedy
+runs (temperature != 0), when none of ``--top-p`` / ``--top-k`` / ``--min-p``
+is given, requests use the agreed non-greedy A/B profile (top_p 0.8,
+top_k 20, min_p 0); setting any of them sends exactly the flags you set and
+omits the rest (server defaults apply). At ``--temperature 0`` the profile
+is never filled in: the sampler takes the greedy ``torch.argmax`` path and
+never invokes the configured ``sampling_backend``, so greedy runs cannot
+observe sampling-backend changes (issue #1040).
 """
 
 from __future__ import annotations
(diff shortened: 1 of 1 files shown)


WilsonZheng0327:
## Notes @edwingao28 

- `bench_client.py` doesn't seem to exist, possibly a local uncommitted test file?
- #1040 mentioned default "stays" greedy but default was `temperature=0.8`, kept it this way but will change if needed.

PR #1065, opened 2026-07-15
feat(comm): port TensorRef support to the Phase-0 data plane

fengyuanyu1:
Stacked on #869. This PR targets `phase0-intra-node-nvlink-relay`; `main` remains unchanged until #869 lands.

## Context

TensorRef forwarding landed on `main` in #808, while #869 replaced the old `relay_io` path with the new `comm` / `stage_io` data plane. The need to reconcile them was identified in [this #869 review](https://github.com/sgl-project/sglang-omni/pull/869#discussion_r3521966803), and #941 retained “the TensorRef port” as a follow-up.

## Changes

- Add a versioned `TensorRef` backed by `DataRef(kind=tensor_ref)`.
- Add `StageConfig.tensor_ref_edges` configuration for paths, thresholds, and final consumers.
- Externalize only configured tensors at or above the threshold.
- Keep producer resources alive until the final consumer sends a success or failure ACK.
- Let intermediate stages forward references without reading or acknowledging them.
- Handle materialization, duplicate references, read failures, request failures, and aborts.
- Report TensorRef count and physical bytes in hop events.
- Fall back to the Relay selected by `CommRouter` when direct CUDA-IPC serialization is unavailable. This fixes the observed Qwen payload header of 165,440 bytes exceeding the 64 KiB direct-path limit.
- Enable Qwen3-Omni TensorRefs for visual deepstack tensors at or above 2 MiB on `image_encoder -> mm_aggregate -> thinker`.

In speech configurations, the main image/video embeddings stay on the ordinary payload path because the talker also consumes them. `merge.py` is unchanged because its existing merge logic already forwards nested TensorRef values without interpreting them.

## Relationship to #932

#932 implements per-edge TensorRef policies on its `main`-based `relay_io` path. This PR follows the same configuration and Qwen3-Omni policy design.

The data-plane implementations differ:

- #932 reuses #808's Relay blob representation; this PR uses #869's `DataRef`.
- #932 matches leaf field names; this PR matches complete payload paths.
- #932 uses the old Relay release lifecycle; this PR retains producer operations until final-consumer ACKs.
- Failure and abort cleanup are reported to the original producer through failed ACKs.

The #932 policy model is the design reference; its patch is not applied directly because it targets a different communication stack.

### Unit tests

The tests added by this PR do not require a GPU.

```bash
python -m pytest -q \
  tests/unit_test/pipeline/test_tensor_ref.py \
  tests/unit_test/pipeline/test_comm_engine_ack.py \
  tests/unit_test/pipeline/test_runtime_schema.py \
  tests/unit_test/pipeline/test_stage.py \
  tests/unit_test/qwen3_omni/test_tensor_ref.py
```

Result: `76 passed`. Pre-commit passed on all changed files.

### Performance tests

Not included in this PR. Baseline and TensorRef performance comparisons will be added separately.

## Functional validation

Two one-request Qwen3-Omni smoke runs passed on commit `fbe129ef88c37554d325ceb8e9fa360a3ffb9084`:

- Cross-GPU TensorRef: `image_encoder` ran on logical GPU 1 and the final consumer `thinker` ran on logical GPU 0. All 3 TensorRefs, totaling 65,863,680 bytes, recorded `cuda_ipc_get_async` with `src_device=1` and `dst_device=0`.
- Same-GPU fallback: the same workload that produced a 165,440-byte direct header completed through the configured CUDA-IPC Relay fallback.

For the cross-GPU run, the published object-ID set matched both the consumer-read set and the producer-ACKed set. These are correctness checks, not performance results.

Related: #862, #808, #869, #932, #941

### DIFF
--- a/sglang_omni/comm/data_ref.py
+++ b/sglang_omni/comm/data_ref.py
@@ -17,6 +17,7 @@ class TransportKind(str, Enum):
 
 class DataKind(str, Enum):
     STAGE_PAYLOAD = "stage_payload"
+    TENSOR_REF = "tensor_ref"
     STREAM_CHUNK = "stream_chunk"
     STREAM_METADATA_TENSOR = "stream_metadata_tensor"
     KV_PAGES = "kv_pages"
--- a/sglang_omni/comm/engine.py
+++ b/sglang_omni/comm/engine.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import asyncio
 import logging
+import uuid
 from contextlib import suppress
 from typing import Any, Callable
 
@@ -11,8 +12,9 @@ import msgspec
 import torch
 
 from sglang_omni.comm import stage_io
-from sglang_omni.comm.data_ref import DataRef, TransportKind
+from sglang_omni.comm.data_ref import DataKind, DataRef, TransportKind
 from sglang_omni.comm.router import CommRouter
+from sglang_omni.comm.tensor_ref import TensorRef
 from sglang_omni.profiler.comm_trace import elapsed_ms as _comm_elapsed_ms
 from sglang_omni.profiler.comm_trace import emit as _comm_trace
 from sglang_omni.profiler.comm_trace import now_ns as _comm_now_ns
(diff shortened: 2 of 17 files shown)


fengyuanyu1:
Hi @JingwenGu0829 @Ratish1 @luojiaxuan @JiaxinD , could you please review this follow-up to #869? Thanks!

JingwenGu0829:
Hi, per discussion before with @luojiaxuan @Ratish1 @zhaochenyang20 and in the WeChat group last week, we are not going to support tensorref after the 869 lands. 

Since tensor ref, under correct communication scheme, is only useful for communiation scenarios such as A->B->C where B is a pure aggregation stage without any actual calculation. In the future we are not going to support any of these aggregate stages, so tensorref will not be meaningful in the future.

Closing this PR. Thanks for the effort and sorry for the waste of time.

PR #1068, opened 2026-07-15
[TTS Refactor] Remove TTS engine builder forwarding

luojiaxuan:
Part of #985.

TtsEngineBuilder now uses the shared checkpoint resolver by default. Fish, Higgs, MOSS-TTS, MOSS-TTS Local, and Voxtral remove pure forwarding overrides; Qwen3-TTS keeps its dependency probe override.

The public create_tts_engine_executor aliases in MOSS-TTS, MOSS-TTS Local, and Qwen3-TTS remain as direct aliases for dotted-path compatibility. The unused Qwen3-TTS audio conversion helper is removed. Runtime behavior and pipeline factory paths stay unchanged.

Non-test diff: +5/-39, net 34 lines deleted.

Validation: 6 engine-factory compatibility tests passed in the SGLang Omni development container. isort, Ruff, Python compilation, and git diff checks passed.


sglang_omni/models/moss_tts/stages.py
```diff
@@ -131,10 +131,6 @@ def create_sglang_tts_engine_executor(
     )
 
 
-def create_tts_engine_executor(*args, **kwargs) -> Any:
```
SandyLuXY, line 134: These non-underscored aliases have been importable since each model integration, so downstream imports or custom factory paths that use them now fail during dotted-path resolution. I reproduced all three paths resolving on `88561c7` and raising `ImportError` on `461863b`.

sglang_omni/models/moss_tts_local/stages.py
```diff
@@ -580,10 +580,6 @@ def create_sglang_tts_engine_executor(
     )
 
 
-def create_tts_engine_executor(*args, **kwargs) -> Any:
```
SandyLuXY, line 583: Same here

sglang_omni/models/qwen3_tts/stages.py
```diff
@@ -167,10 +153,6 @@ def create_sglang_tts_engine_executor(
     )
 
 
-def create_tts_engine_executor(*args, **kwargs) -> Any:
```
SandyLuXY, line 170: same here

SandyLuXY (changes requested):

sglang_omni/models/moss_tts/stages.py
luojiaxuan: Done.

sglang_omni/models/moss_tts_local/stages.py
luojiaxuan: Done.

sglang_omni/models/qwen3_tts/stages.py
luojiaxuan: Done.

SandyLuXY (approved):
LGTM! Please update the PR body to say that the aliases were retained for dotted-path compatibility and refresh the diff statistics.

luojiaxuan:
Done, updated the PR body and diff stats.

PR #1070, opened 2026-07-15
[TTS Refactor] Unify Omni example launchers

luojiaxuan:
What:
Consolidate seven Qwen3-Omni and Ming-Omni example launchers behind one run_omni.py preset registry. Keep the old script paths as thin command-entry compatibility wrappers, with model-specific parsers and runners under examples/launchers.

Why:
The launchers duplicated argument parsing, stage mutation, server startup, offline request execution, and WAV output. New model examples now add a model-local preset module instead of another standalone launcher.

Impact:
Non-test diff is +1548/-1742, or 194 net-deleted lines. The refactor replaces stale offline gpu_placement and apply_server_args_overrides calls with the current stage configuration API while preserving the compatibility launchers' usage examples and contract-defining help.

Checks:
54 launcher contract tests passed on Taurus after the help migration. Ruff check, ruff format --check, and isort --check-only passed for the changed Python files. All unified and compatibility entry points complete --help successfully.

Follow-up:
#858 remains independent because its Ming TTS model is not on main yet. Once this lands, its 393-line launcher can be represented as another preset instead.

Part of #985.


examples/run_qwen3_omni_speech_server.py
```diff
@@ -1,444 +1,21 @@
...
 
-Each stage runs in its own process with dedicated GPU placement.
-Supports text + audio responses via the OpenAI chat completions API.
+try:
+    from examples import _omni_launcher as _launcher
+except ModuleNotFoundError:
+    import _omni_launcher as _launcher
 
-Usage::
+_launch_speech_server = _launcher.launch_qwen_speech_server
```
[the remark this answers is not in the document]
luojiaxuan: Replaced both compatibility aliases with delegating functions. Existing imports remain valid without unused global variables.

examples/run_qwen3_omni_server.py
```diff
@@ -1,215 +1,20 @@
...
-Usage::
+try:
+    from examples import _omni_launcher as _launcher
+except ModuleNotFoundError:
+    import _omni_launcher as _launcher
 
-    python examples/run_qwen3_omni_server.py \
-        --model-path Qwen/Qwen3-Omni-30B-A3B-Instruct \
-        --port 8000
+_launch_text_server = _launcher.launch_qwen_text_server
```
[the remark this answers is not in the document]
luojiaxuan: Kept the legacy import contract by replacing the variable alias with a small delegating function. This removes the CodeQL global-variable finding without breaking the existing launcher tests.

examples/run_ming_omni_server.py
```diff
@@ -1,272 +1,21 @@
...
-Usage::
+try:
+    from examples import _omni_launcher as _launcher
+except ModuleNotFoundError:
+    import _omni_launcher as _launcher
 
-    python examples/run_ming_omni_server.py \
-        --model-path inclusionAI/Ming-flash-omni-2.0 \
-        --port 8000
+_launch_text_server = _launcher.launch_ming_text_server
```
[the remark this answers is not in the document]
luojiaxuan: Kept the legacy import contract by replacing the variable alias with a small delegating function. This removes the CodeQL global-variable finding without breaking the existing launcher tests.

examples/run_qwen3_omni_speech_server.py
```diff
@@ -1,444 +1,21 @@
...
+_parse_thinker_tp_gpu_list = _launcher._parse_thinker_tp_gpu_list
```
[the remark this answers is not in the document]
luojiaxuan: Follow-up: moved the tests to the shared launcher module and removed the compatibility function entirely. The legacy wrapper now contains only command entry points.

examples/run_qwen3_omni_speech_server.py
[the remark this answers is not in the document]
luojiaxuan: Follow-up: moved the tests to the shared launcher module and removed the compatibility function entirely. The legacy wrapper now contains only command entry points.

examples/run_ming_omni_server.py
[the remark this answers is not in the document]
luojiaxuan: Follow-up: moved the tests to the shared launcher module and removed the compatibility function entirely. The legacy wrapper now contains only command entry points.

examples/_omni_launcher.py
```diff
@@ -0,0 +1,1010 @@
...
+    _add_model_path(parser, "inclusionAI/Ming-flash-omni-2.0")
+    _add_offline_args(
+        parser,
+        prompt="你好！给我讲一个有趣的事情。",
+        system="你是一个友好的AI助手。请用自然、温暖的语气说话。",
+        max_new_tokens=256,
+        output="./output_audio.wav",
+    )
+    parser.add_argument("--audio-path", type=str, default=None)
+    parser.add_argument("--voice", type=str, default="DB30")
```
SandyLuXY: `--voice` is parsed here but `run_ming_speech()` never writes it to the `talker` stage: a `--voice CUSTOM_VOICE` config probe still leaves `talker.factory_args["voice"] == "DB30"`, which is what `create_talker_executor()` consumes. Please apply the requested voice to the talker stage.

examples/_omni_launcher.py
```diff
@@ -0,0 +1,1010 @@
...
+        output=args.output,
+        label="Ming-Omni",
+    )
+
+
+def _build_ming_text_parser() -> argparse.ArgumentParser:
+    parser = _parser("Run one Ming-Omni request with text output.")
+    _add_model_path(parser, "inclusionAI/Ming-flash-omni-2.0")
+    parser.add_argument("--prompt", type=str, default="你好，请介绍一下你自己。")
+    parser.add_argument("--thinker-max-seq-len", type=int, default=8192)
```
SandyLuXY: This flag never reaches `run_ming_text()`'s thinker stage: `--thinker-max-seq-len 1234` still leaves `factory_args["thinker_max_seq_len"] == 8192`, and the factory uses that value as SGLang's `context_length`. Please propagate the parsed value to the thinker factory args and cover the offline config path with a non-default value.

examples/_omni_launcher.py
```diff
@@ -0,0 +1,1010 @@
...
+    if args.thinker_only:
+        if args.gpu_audio_encoder is not None or args.gpu_image_encoder is not None:
+            raise ValueError(
+                "--gpu-audio-encoder/--gpu-image-encoder cannot be used "
+                "with --thinker-only"
+            )
+        _configure_ming_thinker_only(config)
+
+    server_updates: dict[str, object] = {}
+    if args.tp_size and args.tp_size > 1:
```
SandyLuXY: Consider rejecting values below one before building the config and add zero/negative regression cases, matching the other TP-enabled presets.

examples/_omni_launcher.py
```diff
@@ -0,0 +1,1010 @@
...
+        _build_ming_speech_parser,
+        lambda args: _run_async(run_ming_speech, args),
+        spawn=True,
+    ),
+    "ming-text": LauncherPreset(
+        "One Ming-Omni text request",
+        _build_ming_text_parser,
+        lambda args: _run_async(run_ming_text, args),
+    ),
+}
```
SandyLuXY, line 23: `_omni_launcher.py` is now advertised as the extension point, but 927 of its 1,010 lines own Qwen/Ming checkpoint defaults, stage names, placement/TP policy, voices, request schemas, and config mutations; the README asks every future model to keep extending that shared file. Please leave only registry/CLI plumbing in the shared module and move each model's preset builders/runners into model-local modules (or per-model `examples/launchers/` modules), then register their preset maps centrally.

examples/_omni_launcher.py
```diff
@@ -0,0 +1,1010 @@
...
+    preset_name: str,
+    argv: Sequence[str] | None = None,
+) -> None:
+    preset = PRESETS[preset_name]
+    if preset.spawn:
+        mp.set_start_method("spawn", force=True)
+    logging.basicConfig(
+        level=logging.INFO,
+        format="%(asctime)s [%(levelname)s] %(name)s: %(message)s",
+    )
```
SandyLuXY, line 43: This hard-coded level drops the existing `LOGLEVEL` contract from all seven entry points: in a fresh process, `LOGLEVEL=DEBUG` selects `DEBUG` on the base launchers but remains `INFO` here. Please read the environment override and preserve the per-preset fallback (Ming text previously defaulted to `DEBUG`), with a fresh-process regression test.

examples/_omni_launcher.py
```diff
@@ -0,0 +1,1010 @@
...
+
+
+def _add_qwen_speech_mem_args(parser: argparse.ArgumentParser) -> None:
+    _add_mem_fraction(parser, "Set mem_fraction_static for both Qwen AR stages.")
+    parser.add_argument("--thinker-mem-fraction-static", type=float, default=None)
+    parser.add_argument("--talker-mem-fraction-static", type=float, default=None)
+
+
+def _build_qwen_speech_server_parser() -> argparse.ArgumentParser:
+    parser = _parser("Launch Qwen3-Omni with text and audio OpenAI responses.")
```
SandyLuXY: This replaces the previous detailed CLI usage text, while several arguments below also lose their contract-defining help: thinker TP/GPU-list coupling, partial-start defaults and minimum, and colocation constraints are no longer visible in `--help` or documented in the updated README. Please restore the relevant parser help/usage text and add assertions for a few key descriptions instead of checking only that `--help` exits successfully.

SandyLuXY (changes requested):

examples/_omni_launcher.py
luojiaxuan: Done.

examples/_omni_launcher.py
luojiaxuan: Done.

examples/_omni_launcher.py
luojiaxuan: Done.

examples/_omni_launcher.py
luojiaxuan: Done.

examples/_omni_launcher.py
luojiaxuan: Done.

examples/_omni_launcher.py
luojiaxuan: Done.

examples/launchers/ming_omni.py
```diff
@@ -0,0 +1,423 @@
...
+        default=1,
+        help="Tensor parallel size for thinker (must be >= 1)",
+    )
+    add_relay_backend(target)
+    target.add_argument("--voice", type=str, default="DB30")
+    add_mem_fraction(target, "Set mem_fraction_static for the thinker stage.")
+    target.add_argument("--cpu-offload-gb", type=int, default=None)
+    target.add_argument("--enable-streaming-tts", action="store_true")
+    add_server_args(target, model_name="ming-omni")
+    return target
```
SandyLuXY, line 279: The Qwen speech help is restored, but the other compatibility entry points still lose contract-defining guidance. For example, `ming-speech-server --help` now lists `--enable-streaming-tts` and `--cpu-offload-gb` without explaining the 8-stage streaming path, the default 7-stage path, the checkpoint offload requirement, or TP start-GPU semantics, and the README does not replace that text. Please migrate the remaining Ming/offline help and examples.

luojiaxuan: Done, restored the remaining Ming and offline help contracts and examples.

SandyLuXY:
One note: `qwen3-speech-server --colocated` fails placement because the launcher selects `Qwen3OmniSpeechColocatedPipelineConfig` without populating typed `total_gpu_memory_fraction` budgets. I checked the merge base `88561c7`, and the legacy `examples/run_qwen3_omni_speech_server.py --colocated` path already fails with the same error, including with `--mem-fraction-static 0.5`, so I’m treating this as a pre-existing Qwen launcher/config issue rather than a regression from this PR.

SandyLuXY (approved):

PR #1057, opened 2026-07-14
[TTS Refactor] Use upstream Transformers Higgs tokenizer

SandyLuXY:
## Motivation

- SGLang-Omni currently vendors the Higgs Audio V2 tokenizer because older Transformers releases did not provide it. The pinned Transformers 5.6 release now includes the same config and model classes.
- This PR removes the vendored model and uses the upstream Transformers classes instead. It retains the bundled codec config and existing checkpoint weight loading path, with no changes to the checkpoint format or encode/decode interface.

## Modifications

1. `sglang_omni/models/higgs_tts/audio_codec.py`
   - Import the Higgs Audio V2 tokenizer config and model directly from
     Transformers.
   - Load the bundled architecture config with `from_json_file()` while keeping
     the existing checkpoint-prefix extraction, frozen parameters, and
     encode/decode facade unchanged.
   - Update codec documentation and compatibility errors to describe the
     upstream implementation.
2. Higgs tokenizer assets and packaging
   - Remove `_vendored/__init__.py` and the 940-line vendored tokenizer module.
   - Move the byte-identical codec JSON to
     `configs/higgs_audio_v2_tokenizer.json` and update setuptools package-data
     so the file remains in wheels.
3. Tests and documentation
   - Add a meta-device regression test that pins the upstream class binding,
     527-key state-dict structure, representative tensor shapes, frame rate,
     and codebook count.
   - Update `tests/README.md` with the new Higgs codec coverage.

## Related Issues
#985 

## Accuracy Test

**Setup:**

- Model/checkpoint: `bosonai/higgs-tts-3-4b`
- Baseline: `main @ 1618b7ce`; candidate: `xinyu/refactor-higgs-upstream-tokenizer @ d24c96fc`.
- Dataset: full EN split of `zhaochenyang20/seed-tts-eval-arrow`, 1088 samples, non-streaming generation with reference cloning.
- Hardware: 1 x NVIDIA H100 80GB HBM3, physical GPU 0; generation concurrency 8 with one warmup request.

| Metric | Baseline | Candidate |
| --- | ---: | ---: |
| Generation completed / failed | 1088 / 0 | 1088 / 0 |
| WER evaluated / skipped | 1088 / 0 | 1088 / 0 |
| Corpus WER | 0.93% | 0.86% |
| Per-sample mean WER | 0.91% | 0.83% |
| Per-sample max WER | 42.86% | 42.86% |

## Benchmark & Profiling

Same setup as described in the _Accuracy Test_ section. 
[...]

PR #1071, opened 2026-07-16
[Perf][Higgs TTS] Prefill admission coalescing + out-of-process vocoder + compiled codec decode

MayDomine:
## Motivation

Single-replica Higgs TTS serving is host-bound: the scheduler event loop saturates while the GPU idles at ~20% SM. Per-segment timing and NVTX profiling (full analysis on #921) attribute the loop time to three fixable costs:

1. **Synchronous prefill steps** ran at arrival granularity — 23 steps/s carrying only ~2 requests each, paying a ~25 ms mostly batch-size-independent host cost per step (60% of the loop).
2. **The DAC vocoder thread shared the scheduler process**: it contends for the GIL (prefill steps measured +38% slower while it is active) and its decode thread is the first component to saturate.
3. Vocoder per-chunk decode cost itself (17 ms in-process, 11.6 ms once isolated).

## Modifications

One commit, three stacked changes:

1. **Prefill admission coalescing** (`OmniScheduler.get_new_batch_prefill` gate): hold prefill until `prefill_coalesce_requests` are waiting or the oldest has waited `prefill_coalesce_wait_ms` (chunked prefill bypasses the gate; the deadline bounds added TTFB). Plumbed as Higgs stage factory args, default off.
2. **Vocoder in its own process** (`process="vocoder"` in the Higgs pipeline config): removes the GIL contention and gives the vocoder its own core. Sharing GPU 0 across process groups requires the explicit per-stage `total_gpu_memory_fraction` budgets added alongside (0.03 encoder / 0.85 engine / 0.10 vocoder).
3. **`torch.compile` on the codec decode** (`compile_decode` factory arg on the vocoder stage, following the audio-encoder precedent from #612; enabled by default in the Higgs pipeline config): 11.6 → 5.8 ms per chunk.

## Accuracy Test

Changes 1–2 alter scheduling and process placement only, not numerics; generated token streams are unaffected by construction. For the compiled decode path (change 3), the seed-tts-eval WER A/B is posted below: 1.005% -> 0.996% corpus WER (within noise, parity confirmed).

## Benchmark & Profiling

1× H100 80 GB, seed-tts-eval voice-clone (~157 prompt / ~115 output tokens), non-streaming, conc = cap, 90 s windows. Ladder measured on the rebased head (base includes #939/#1057); the cap sweep and DP tables below are from the pre-rebase head and will be refreshed with the #1037 recipe follow-up.

**Single replica ladder** (cap 96; re-measured on the rebased head, 2×90 s means; `QPS × lat ≈ 96` sanity-checked per row):

| config | QPS | lat mean / p95 | QPS×lat |
| --- | ---: | --- | ---: |
| main | 49.5 | 1.92 s / 6.0 s | 95.3 |
| + coalescing (K=32, T=300 ms) | 64.9 (+31%) | 1.50 s / 3.4 s | 97.7 |
| + vocoder own process | 85.1 (+31%) | 1.13 s / 3.1 s | 95.8 |
| + compiled decode | **99.7 (+17%)** | 0.96 s / 2.9 s | 95.6 |

**cap = conc sweep** (full stack; cap is now a real throughput/latency knob):

| cap | 96 | 128 | 160 | 192 | 224 | 256 | 288 | 320 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| QPS | 99.3 | 114.7 | **124.9** | 131.8 | 136.8 | 142.6 | 145.6 | 152.8 |
| p95 (s) | 3.0 | 3.5 | 4.0 | 4.4 | 5.5 | 5.8 | 7.2 | 6.2 |

Pareto knee at cap=160.

**Same-GPU DP × MPS** (3 replicas × cap 160, per-replica core slices):

| deployment | aggregate QPS | per-replica | lat mean |
| --- | ---: | ---: | --- |
| DP3 × MPS, old recipe (#921) | 124.4 | ~41 | — |
| **DP3 × MPS, this stack** | **323.5 / 330.9 → ~327** | ~109 (87% scaling) | 1.47 s |

56.3 GB used. One measurement caveat that becomes a follow-up: the KV pool is sized from free VRAM at boot, so same-GPU replicas end up with unequal pools (12.5 / 6.3 / 2.0 GB measured) and the last replica throttles to ~40% of its peers. The 327 number pins per-replica pools equal (54k tokens) via a measurement-time override; exposing an explicit KV budget knob lands with the #1037 recipe follow-up.

**Vocoder placement side-metrics** (review follow-up): idle boot VRAM +130 MiB for the separate codec process (69,883 vs 69,753 MiB); streaming TTFP (seed-tts-eval-50, conc 16, `--stream`) improves slightly out-of-process — mean 1.006 s / p95 2.00 s vs 1.089 s / 2.08 s in-process, i.e. the shm+ZMQ hop is repaid by the GIL relief.

Numbers re-validated on the review-fixed head (`0b6cdaa2`): 100.3 / 99.6 QPS at cap=conc=96.

Profiling method and full bottleneck decomposition (per-segment loop timers + NVTX under nsys, production rates) are documented on #921.

## Checklist

- [x] Add unit tests (`tests/unit_test/scheduling/test_prefill_coalesce_gate.py`, incl. partial-admission and abort paths).
- [ ] Update documentation / docstrings / example tutorials as needed. *(recipe update for #1037 planned as follow-up)*

### DIFF
--- a/sglang_omni/models/higgs_tts/config.py
+++ b/sglang_omni/models/higgs_tts/config.py
@@ -46,6 +46,9 @@ class HiggsTtsPipelineConfig(PipelineConfig):
             factory=f"{_PKG}.stages.create_audio_encoder_executor",
             factory_args={"device": "cuda"},
             gpu=0,
+            runtime=StageRuntimeConfig(
+                resources=StageResourceConfig(total_gpu_memory_fraction=0.03)
+            ),
             next="tts_engine",
         ),
         StageConfig(
@@ -58,15 +61,21 @@ class HiggsTtsPipelineConfig(PipelineConfig):
                 "enable_async_decode": True,
             },
             gpu=0,
+            runtime=StageRuntimeConfig(
+                resources=StageResourceConfig(total_gpu_memory_fraction=0.85)
+            ),
             next="vocoder",
             stream_to=["vocoder"],
         ),
         StageConfig(
             name="vocoder",
-            process="pipeline",
+            process="vocoder",
             factory=f"{_PKG}.stages.create_vocoder_executor",
-            factory_args={"device": "cuda"},
+            factory_args={"device": "cuda",  "compile_decode": True},
             gpu=0,
+            runtime=StageRuntimeConfig(
+                resources=StageResourceConfig(total_gpu_memory_fraction=0.10)
+            ),
             terminal=True,
             can_accept_stream_before_payload=True,
         ),
--- a/sglang_omni/models/higgs_tts/config.py
+++ b/sglang_omni/models/higgs_tts/config.py
@@ -5,7 +5,12 @@ from __future__ import annotations
 
 from typing import Any, ClassVar
 
-from sglang_omni.config import PipelineConfig, StageConfig
+from sglang_omni.config import (
+    PipelineConfig,
+    StageConfig,
+    StageResourceConfig,
+    StageRuntimeConfig,
+)
 
 _PKG = "sglang_omni.models.higgs_tts"
 
--- a/sglang_omni/models/higgs_tts/config.py
+++ b/sglang_omni/models/higgs_tts/config.py
@@ -71,7 +71,7 @@ class HiggsTtsPipelineConfig(PipelineConfig):
             name="vocoder",
             process="vocoder",
             factory=f"{_PKG}.stages.create_vocoder_executor",
-            factory_args={"device": "cuda",  "compile_decode": True},
+            factory_args={"device": "cuda", "compile_decode": True},
             gpu=0,
             runtime=StageRuntimeConfig(
                 resources=StageResourceConfig(total_gpu_memory_fraction=0.10)
--- a/sglang_omni/models/higgs_tts/engine_builder.py
+++ b/sglang_omni/models/higgs_tts/engine_builder.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import importlib
+import os
 from typing import Any
 
 from sglang_omni.models.higgs_tts import request_builders
@@ -29,6 +30,8 @@ class HiggsTtsEngineBuilder(TtsEngineBuilder):
         async_decode_min_batch_size: int,
         stream_stride: int = DEFAULT_HIGGS_STREAM_STRIDE,
         stream_followup_stride: int = DEFAULT_HIGGS_STREAM_FOLLOWUP_STRIDE,
+        prefill_coalesce_requests: int = 0,
+        prefill_coalesce_wait_ms: float = 60.0,
     ) -> None:
         self.max_new_tokens = max_new_tokens
         self.max_running_requests = max_running_requests
@@ -37,6 +40,8 @@ class HiggsTtsEngineBuilder(TtsEngineBuilder):
         self.async_decode_min_batch_size = async_decode_min_batch_size
         self.stream_stride = stream_stride
         self.stream_followup_stride = stream_followup_stride
+        self.prefill_coalesce_requests = prefill_coalesce_requests
+        self.prefill_coalesce_wait_ms = prefill_coalesce_wait_ms
         self.model: Any | None = None
 
     def generation_defaults(
@@ -100,6 +105,8 @@ class HiggsTtsEngineBuilder(TtsEngineBuilder):
         return {
             "enable_async_decode": self.enable_async_decode,
             "async_decode_min_batch_size": self.async_decode_min_batch_size,
+            "prefill_coalesce_requests": self.prefill_coalesce_requests,
+            "prefill_coalesce_wait_ms": self.prefill_coalesce_wait_ms,
         }
 
     def post_scheduler_setup(self, scheduler: Any, model_runner: Any) -> None:
--- a/sglang_omni/models/higgs_tts/engine_builder.py
+++ b/sglang_omni/models/higgs_tts/engine_builder.py
@@ -31,6 +31,7 @@ class HiggsTtsEngineBuilder(TtsEngineBuilder):
         stream_followup_stride: int = DEFAULT_HIGGS_STREAM_FOLLOWUP_STRIDE,
         prefill_coalesce_requests: int = 0,
         prefill_coalesce_wait_ms: float = 60.0,
+        total_gpu_memory_fraction: float | None = None,
     ) -> None:
         self.max_new_tokens = max_new_tokens
         self.max_running_requests = max_running_requests
@@ -41,6 +42,7 @@ class HiggsTtsEngineBuilder(TtsEngineBuilder):
         self.stream_followup_stride = stream_followup_stride
         self.prefill_coalesce_requests = prefill_coalesce_requests
         self.prefill_coalesce_wait_ms = prefill_coalesce_wait_ms
+        self.total_gpu_memory_fraction = total_gpu_memory_fraction
         self.model: Any | None = None
 
     def generation_defaults(
@@ -57,7 +59,13 @@ class HiggsTtsEngineBuilder(TtsEngineBuilder):
             "max_running_requests": self.max_running_requests,
             "cuda_graph_max_bs": self.cuda_graph_max_bs,
             "disable_cuda_graph": False,
-            "mem_fraction_static": 0.85,
+            # The stage's runtime.resources.total_gpu_memory_fraction is the
+            # single source of truth when set; 0.85 only applies without one.
+            "mem_fraction_static": (
+                self.total_gpu_memory_fraction
+                if self.total_gpu_memory_fraction is not None
+                else 0.85
+            ),
             "chunked_prefill_size": 8192,
             "dtype": "bfloat16",
         }
--- a/sglang_omni/models/higgs_tts/engine_builder.py
+++ b/sglang_omni/models/higgs_tts/engine_builder.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import importlib
+import logging
 from typing import Any
 
 from sglang_omni.models.higgs_tts import request_builders
@@ -13,6 +14,8 @@ from sglang_omni.models.higgs_tts.vocoder_scheduler import (
     DEFAULT_HIGGS_STREAM_STRIDE,
 )
 from sglang_omni.scheduling.engine_factory import TtsEngineBuilder
+
+logger = logging.getLogger(__name__)
 
 
 class HiggsTtsEngineBuilder(TtsEngineBuilder):
@@ -33,6 +36,13 @@ class HiggsTtsEngineBuilder(TtsEngineBuilder):
         prefill_coalesce_wait_ms: float = 60.0,
         total_gpu_memory_fraction: float | None = None,
     ) -> None:
+        if total_gpu_memory_fraction is not None and not (
+            0.0 < total_gpu_memory_fraction < 1.0
+        ):
+            raise ValueError(
+                "Higgs tts_engine total_gpu_memory_fraction must be in (0, 1): "
+                "it drives sglang mem_fraction_static, which requires < 1"
+            )
         self.max_new_tokens = max_new_tokens
         self.max_running_requests = max_running_requests
         self.cuda_graph_max_bs = cuda_graph_max_bs
@@ -67,6 +77,22 @@ class HiggsTtsEngineBuilder(TtsEngineBuilder):
             "chunked_prefill_size": 8192,
             "dtype": "bfloat16",
         }
+
+    def adjust_overrides(self, overrides: dict[str, Any]) -> None:
+        # Note: (Jiaxin Deng) an explicit mem_fraction_static override (e.g.
+        # --talker-mem-fraction-static) wins, but never silently.
+        expected = self.total_gpu_memory_fraction
+        if expected is None:
+            return
+        actual = overrides.get("mem_fraction_static")
+        if actual is not None and abs(actual - expected) <= 1e-9:
+            return
+        logger.warning(
+            "Higgs tts_engine mem_fraction_static=%s overrides the "
+            "placement-validated total_gpu_memory_fraction=%s",
+            actual,
+            expected,
+        )
 
     def customize_server_args(self, server_args: Any) -> None:
         server_args.disable_overlap_schedule = True
--- a/sglang_omni/models/higgs_tts/engine_builder.py
+++ b/sglang_omni/models/higgs_tts/engine_builder.py
@@ -4,7 +4,6 @@
 from __future__ import annotations
 
 import importlib
-import os
 from typing import Any
 
 from sglang_omni.models.higgs_tts import request_builders
--- a/sglang_omni/models/higgs_tts/engine_builder.py
+++ b/sglang_omni/models/higgs_tts/engine_builder.py
@@ -59,8 +59,6 @@ class HiggsTtsEngineBuilder(TtsEngineBuilder):
             "max_running_requests": self.max_running_requests,
             "cuda_graph_max_bs": self.cuda_graph_max_bs,
             "disable_cuda_graph": False,
-            # The stage's runtime.resources.total_gpu_memory_fraction is the
-            # single source of truth when set; 0.85 only applies without one.
             "mem_fraction_static": (
                 self.total_gpu_memory_fraction
                 if self.total_gpu_memory_fraction is not None
--- a/sglang_omni/models/higgs_tts/stages.py
+++ b/sglang_omni/models/higgs_tts/stages.py
@@ -517,7 +517,7 @@ def create_vocoder_executor(
                 dtype=torch.long,
                 device="cpu",
             )
-            # note (stephenkgli): Match serving's contiguous [T, N] layout and
+            # Note: (stephenkgli) match serving's contiguous [T, N] layout and
             # warm the zero-one-specialized batch and frame-count classes.
             for frame_count in _VOCODER_COMPILE_WARMUP_FRAME_COUNTS:
                 frame_codes_TN = warm_codes_TN[:frame_count]
--- a/sglang_omni/models/higgs_tts/stages.py
+++ b/sglang_omni/models/higgs_tts/stages.py
@@ -79,6 +79,7 @@ _REF_CODE_CACHE_MAX_ITEMS = 256
 _REF_CODE_CACHE_MAX_BYTES = 256 * 1024 * 1024
 _REF_WAVEFORM_CACHE_MAX_ITEMS = 256
 _REF_WAVEFORM_CACHE_MAX_BYTES = 512 * 1024 * 1024
+_VOCODER_COMPILE_WARMUP_FRAME_COUNTS = (1, 8)
 
 
 def _reference_audio_cache_key(reference_audio: Any) -> str | None:
@@ -506,11 +507,20 @@ def create_vocoder_executor(
         eager_decode = codec.model.decode
         try:
             codec.model.decode = torch.compile(eager_decode, dynamic=True)
-            warm_codes = codec.encode_reference(
-                torch.zeros(codec.SAMPLE_RATE // 2), sample_rate=codec.SAMPLE_RATE
+            warm_codes_TN = torch.zeros(
+                (
+                    max(_VOCODER_COMPILE_WARMUP_FRAME_COUNTS),
+                    int(codec.model.config.num_quantizers),
+                ),
+                dtype=torch.long,
+                device="cpu",
             )
-            codec.decode(warm_codes)
-            codec.decode_batch([warm_codes[:8], warm_codes[:8]])
+            # note (stephenkgli): Match serving's contiguous [T, N] layout and
+            # warm the zero-one-specialized batch and frame-count classes.
+            for frame_count in _VOCODER_COMPILE_WARMUP_FRAME_COUNTS:
+                frame_codes_TN = warm_codes_TN[:frame_count]
+                codec.decode(frame_codes_TN)
+                codec.decode_batch([frame_codes_TN, frame_codes_TN])
         except Exception:
             logger.warning(
                 "torch.compile of the codec decode failed; falling back to the "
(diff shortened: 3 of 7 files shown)


MayDomine:
## Accuracy: WER A/B for the compiled codec decode

seed-tts-eval (en, 1088 samples), fixed sampler seed 42, identical server config, transcribed with Qwen3-ASR-1.7B. Only the vocoder decode path differs (eager vs `torch.compile(..., dynamic=True)`).

| metric | eager decode (baseline) | compiled decode | Δ |
| --- | ---: | ---: | ---: |
| WER (corpus) | 1.005% | 0.996% | −0.008 pp |
| WER (per-sample mean) | 0.981% | 0.962% | −0.019 pp |
| samples > 50% WER | 0 | 0 | — |
| evaluated | 1088/1088 | 1088/1088 | — |

Within evaluation noise — parity confirmed for `compile_decode=True` as the default.

sglang_omni/models/higgs_tts/config.py
```diff
@@ -42,6 +42,9 @@ def generation_sglang_role_to_stage(cls) -> dict[str, str]:
             factory=f"{_PKG}.stages.create_audio_encoder_executor",
             factory_args={"device": "cuda"},
             gpu=0,
+            runtime=StageRuntimeConfig(
```
JiaxinD, line 54: `StageRuntimeConfig` and `StageResourceConfig` are never imported (line 8 only brings in `PipelineConfig, StageConfig`), so importing this module raises `NameError`: higgs serving and the higgs unit tests die at collection on this head. Fix is the moss-style import list, and the benchmark numbers need a re-run on the fixed head since this exact tree cannot start.

sglang_omni/models/higgs_tts/config.py
```diff
@@ -54,15 +57,21 @@ def generation_sglang_role_to_stage(cls) -> dict[str, str]:
                 "enable_async_decode": True,
             },
             gpu=0,
+            runtime=StageRuntimeConfig(
+                resources=StageResourceConfig(total_gpu_memory_fraction=0.85)
```
JiaxinD, line 70: Nothing enforces these fractions for higgs: no stage factory declares `total_gpu_memory_fraction` and `infra_kwargs()` is empty, so they only satisfy the colocation check while the engine still sizes KV from free VRAM at profile time. Process groups also start concurrently under the per-GPU startup lock, so KV capacity now depends on whether the vocoder loads first; plumbing the budget like qwen3's `_apply_colocated_ar_memory_contract` would make these numbers real.

sglang_omni/models/higgs_tts/config.py
```diff
@@ -54,15 +57,21 @@ def generation_sglang_role_to_stage(cls) -> dict[str, str]:
...
+            ),
             next="vocoder",
             stream_to=["vocoder"],
         ),
         StageConfig(
             name="vocoder",
-            process="pipeline",
+            process="vocoder",
```
JiaxinD, line 77: Out of process, the codec loads once per process, so its weights are resident twice on gpu0 (the sharing note at `utils.py:31` is now stale) and stream chunks cross shm+ZMQ instead of an in-process queue. Worth adding the VRAM delta and a streaming TTFB number to the body, the benchmarks only cover non-streaming.

sglang_omni/models/higgs_tts/config.py
```diff
@@ -54,15 +57,21 @@ def generation_sglang_role_to_stage(cls) -> dict[str, str]:
...
             factory=f"{_PKG}.stages.create_vocoder_executor",
-            factory_args={"device": "cuda"},
+            factory_args={"device": "cuda", "compile_decode": True},
```
JiaxinD, line 79: If inductor fails at first decode there is no eager fallback and no off switch short of a custom pipeline config, so the terminal stage errors every request until restart. The moss vocoder logs and degrades to eager when capture fails; the same pattern or an env kill switch would make default-on safer.

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -421,13 +425,16 @@ def create_vocoder_executor(
...
+    compile_decode: bool = False,
 ):
     """Decode Higgs delayed codes to a mono 24 kHz waveform.
 
     Codec weights are extracted from the TTS checkpoint itself.
     """
     checkpoint_dir = resolve_checkpoint(model_path)
     codec = get_or_load_codec(checkpoint_dir, device, dtype)
+    if compile_decode:
+        codec.model.decode = torch.compile(codec.model.decode, dynamic=True)
```
JiaxinD: `torch.compile` is lazy and nothing calls decode during init, so the first request eats the compile stall on the vocoder's single scheduler thread after the stage is already marked ready (and more than once: B=1, B>=2, and the small first streaming chunk compile separately). The encoder above and moss's `warmup_now()` both warm inside the factory; a dummy decode here keeps the stall out of serving.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -810,6 +819,27 @@ def _emit_request_error(self, request_id: str, error: Exception) -> None:
...
+            return _Upstream.get_new_batch_prefill(self)
+        if len(self.waiting_queue) >= self.prefill_coalesce_requests:
+            self._prefill_coalesce_t0 = None
+            return _Upstream.get_new_batch_prefill(self)
+        now = time.perf_counter()
+        if self._prefill_coalesce_t0 is None:
+            self._prefill_coalesce_t0 = now
+            return None
+        if now - self._prefill_coalesce_t0 >= self.prefill_coalesce_wait_s:
+            self._prefill_coalesce_t0 = None
```
JiaxinD: `_prefill_coalesce_t0` tracks a queue episode, not the oldest request: it resets on every pass-through, so when upstream admits partially or returns None under KV pressure the leftovers re-wait a full window each cycle, and aborting the arming request hands a newcomer the nearly expired deadline. Keying on the oldest queued request's entry time (upstream already records it) would make the deadline actually bound added TTFB; either way the partial-admission path deserves a test.

sglang_omni/models/higgs_tts/config.py
MayDomine: Fixed in 0b6cdaa2 — the lint squash dropped the imports while keeping the usages.

sglang_omni/models/higgs_tts/config.py
MayDomine: Correct — today they are declarative only: they satisfy the colocation check and document intent, while KV is still sized from free VRAM at profile time (we hit exactly the boot-order inhomogeneity you describe at DP3: 12.5/6.3/2.0 GB pools). Plumbing the budget for real via the qwen3 `_apply_colocated_ar_memory_contract` pattern is queued as the follow-up alongside the #1037 recipe work (same fix also needs an explicit KV pin for multi-replica). Kept out of this PR to limit scope; happy to link the follow-up here when it's up.

sglang_omni/models/higgs_tts/config.py
MayDomine: Measured both. Idle boot VRAM: 69,883 MiB (own process) vs 69,753 MiB (in-process) → +130 MiB for the duplicate codec + context. Streaming (seed-tts-eval-50, conc 16, --stream): TTFP mean 1.006 s / p95 2.00 s own-process vs 1.089 s / 2.08 s in-process — the shm+ZMQ hop is more than paid back by the GIL relief, so streaming improves slightly too. Added both numbers to the body and refreshed the stale sharing note at utils.py (0b6cdaa2).

sglang_omni/models/higgs_tts/config.py
MayDomine: Done in 0b6cdaa2: the factory now compiles inside a try, warms the compiled path immediately, and on any failure logs a warning and restores the eager decode — so an inductor failure degrades at startup instead of erroring every request.

sglang_omni/models/higgs_tts/stages.py
MayDomine: Done in 0b6cdaa2: the factory warms the compiled decode with an encode-derived dummy (B=1 full clip and a B=2 batched slice) before the stage reports ready, so the compile stall stays out of serving; with dynamic=True the small first streaming chunk reuses the dynamic artifact. The warmup doubles as the failure probe for the eager fallback above.

sglang_omni/scheduling/omni_scheduler.py
MayDomine: Good catch — reworked in 0b6cdaa2: the deadline is now keyed on the oldest queued request's enqueue time (stamped once at queue entry; retracted requests keep their original stamp), and the episode timer is gone entirely. Partial admission leftovers therefore release immediately, and an abort can't hand a newcomer an expired window. Both paths plus the unstamped-request case are covered in tests/unit_test/scheduling/test_prefill_coalesce_gate.py.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -810,6 +820,22 @@ def _emit_request_error(self, request_id: str, error: Exception) -> None:
...
+        # loop burns its time on tiny prefill batches. Hold admission until the
+        # batch is worth its fixed step cost; the deadline is measured from the
+        # oldest queued request, so partial admission or aborts never restart it.
+        if self.prefill_coalesce_requests <= 1 or self.chunked_req is not None:
+            return _Upstream.get_new_batch_prefill(self)
+        waiting = self.waiting_queue
+        if not waiting or len(waiting) >= self.prefill_coalesce_requests:
+            return _Upstream.get_new_batch_prefill(self)
+        now = time.perf_counter()
+        oldest = min(getattr(req, "_coalesce_enqueue_t", now) for req in waiting)
```
db-ol: One latent issue that survives the rework: the deadline still comes from each rank's own clock. Each rank stamps `_coalesce_enqueue_t` with its local perf_counter and compares it against its local now, so at tp_size > 1 two ranks can disagree on expiry in the same iteration. One rank then builds a prefill batch and enters the TP collective while the other returns None, and there is no watchdog to recover. Higgs runs tp=1 today, but OmniScheduler is shared across models. Request build workers are already force disabled at tp_size > 1 in `__init__` to protect the same lockstep invariant. Could the gate get the same guard, or the decision be made on the entry rank and broadcast?

db-ol:
Thanks for rerunning the numbers on the fixed head. One gap remains: the branch still bases on main@c00e6f9, now about 27 commits behind. Two merged PRs rewrote the components measured here. #939 migrated the Higgs streaming vocoder onto StreamingVocoderBase, and #1057 replaced the vendored codec with the upstream Transformers implementation, which is the module the compile and the new warmup target. A trial merge shows no textual conflicts, so the rebase looks mechanical. Would you rerun the vocoder isolation and compiled decode rows on a rebased head, so the +31% and +18% steps reflect what this merges into?

One more small thing, a sanity check on the coalescing row. With conc = cap = 96, QPS times mean latency should be about 96, and the other rows land within 2% (48.5 x 1.97, 83.8 x 1.13, 99.3 x 0.95). The coalescing row gives 63.8 x 1.65 = 105. Is 1.65 s a typo, or did that run use a different concurrency?

sglang_omni/scheduling/omni_scheduler.py
MayDomine: Guarded in 0027ed22: coalescing is force-disabled at tp_size > 1 with a warning, mirroring the request-build workers. Rank-0-decides + broadcast would add a per-iteration collective to the host path this PR is trimming, so deferring that until a tp>1 model actually wants the gate.

MayDomine:
Rebased onto current main (#939/#1057 included, no conflicts; the compile still binds cleanly to the upstream Transformers codec). Full ladder re-measured on the rebased head — steps hold: 49.5 → 64.9 (+31%) → 85.1 (+31%) → 99.7 (+17%), and the PR table now prints QPS×lat per row (95.3–97.7, all ≈96).

Good catch on the coalescing row: 1.65 s was a transcription slip (run 2's latency instead of the 3-run mean, 1.52 s — which satisfies the invariant: 63.8 × 1.52 = 96.9). Table replaced with the rebased-head means.

sglang_omni/models/higgs_tts/engine_builder.py
AkazaAkane: Related to the thread on the declarative total_gpu_memory_fraction budgets: I traced the plumbing — resolve_stage_factory_arg_defaults only injects the fraction when the factory signature declares it, and none of the three Higgs factories do, so the engine's real budget comes from this hardcoded 0.85 while the config's 0.85 is documentation that happens to match. Two copies of the same number will drift: someone edits the config fraction, placement validation still passes, and this line silently wins.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -814,6 +832,22 @@ def _emit_request_error(self, request_id: str, error: Exception) -> None:
             )
         )
 
+    def get_new_batch_prefill(self):
+        # NOTE(maydomine): requests often arrive faster than they batch, so the
+        # loop burns its time on tiny prefill batches. Hold admission until the
+        # batch is worth its fixed step cost; the deadline is measured from the
+        # oldest queued request, so partial admission or aborts never restart it.
+        if self.prefill_coalesce_requests <= 1 or self.chunked_req is not None:
```
AkazaAkane, line 837: The gate doesn't consider whether any decode work exists. When running_batch is empty (cold start, traffic trough), holding prefill amortizes nothing — the loop is idle anyway — and costs pure TTFB: up to T (300 ms in the benchmark config) added to a lone request's first token for zero throughput gain.
Suggest an additional bypass, which makes the gate strictly non-regressive at low load, which also strengthens the case for eventually enabling it by default.
```
if self.running_batch is None or self.running_batch.is_empty():
    return _Upstream.get_new_batch_prefill(self)
```

tests/unit_test/scheduling/test_prefill_coalesce_gate.py
```diff
@@ -0,0 +1,131 @@
...
+def test_abort_does_not_hand_newcomers_an_expired_deadline(upstream, clock):
+    # The request that opened the window was aborted; a fresh arrival must
+    # wait its own window rather than inherit the nearly expired one.
+    sched = _StubScheduler(coalesce_requests=8, wait_ms=60.0)
+    clock.return_value = 100.1
+    sched.waiting_queue = [_req(100.09)]  # newcomer, 10ms old
+    assert sched.get_new_batch_prefill() is None
+
+
+def test_unstamped_request_is_treated_as_fresh(upstream, clock):
```
AkazaAkane: This test currently pins the dangerous semantics from the comment above: it asserts the unstamped request is held, but nothing asserts it is ever released. If the gate self-heals by stamping on miss, please flip this to assert the request is held at first observation and released once the mocked clock advances past wait_ms. As-is, the test would keep passing even if unstamped requests were held forever.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -814,6 +832,22 @@ def _emit_request_error(self, request_id: str, error: Exception) -> None:
...
+            return _Upstream.get_new_batch_prefill(self)
+        waiting = self.waiting_queue
+        if not waiting or len(waiting) >= self.prefill_coalesce_requests:
+            return _Upstream.get_new_batch_prefill(self)
+        now = time.perf_counter()
+        oldest = min(getattr(req, "_coalesce_enqueue_t", now) for req in waiting)
```
AkazaAkane: getattr(req, "_coalesce_enqueue_t", now) re-evaluates to the current time on every scheduler iteration, so a request that never got stamped yields now - oldest == 0 forever — it can only be released by the K threshold, never by the deadline. At low traffic that's an unbounded hold.

It's unreachable today because enqueue_if_live() (L744-746) is the only entry point into waiting_queue for new requests and retracted requests reuse already-stamped objects. But the failure mode is silent: any future enqueue path that bypasses enqueue_if_live, or a rename of the attribute, degrades the gate to K-only release with all unit tests still green (they stub the queue directly).

Suggested Fix. With this, an unstamped request starts aging from first observation and is guaranteed release within T.
```
now = time.perf_counter()
oldest = now
for req in waiting:
    t = getattr(req, "_coalesce_enqueue_t", None)
    if t is None:
        t = req._coalesce_enqueue_t = now  # treat as enqueued now; starts aging next tick
    oldest = min(oldest, t)
```

Also extent test_unstamped_request_is_treated_as_fresh to advance the mocked clock past wait_ms and verify that the request is eventually released.

sglang_omni/models/higgs_tts/engine_builder.py
MayDomine: Wired in 07d0149e: the tts_engine factory now declares total_gpu_memory_fraction, so the config value is injected and drives mem_fraction_static — single source; the hardcoded 0.85 only applies when the config omits a budget.

sglang_omni/scheduling/omni_scheduler.py
MayDomine: Added in a2cb0c0b — agreed that it makes the gate strictly non-regressive at low load.

tests/unit_test/scheduling/test_prefill_coalesce_gate.py
MayDomine: Done in a2cb0c0b — the test now asserts held at first observation, then released once the mocked clock passes wait_ms.

sglang_omni/scheduling/omni_scheduler.py
MayDomine: Fixed with your stamp-on-miss suggestion in a2cb0c0b; unstamped requests start aging at first observation and release within T.

sglang_omni/models/higgs_tts/engine_builder.py
```diff
@@ -49,7 +55,11 @@ def generation_defaults(
             "max_running_requests": self.max_running_requests,
             "cuda_graph_max_bs": self.cuda_graph_max_bs,
             "disable_cuda_graph": False,
-            "mem_fraction_static": 0.85,
+            "mem_fraction_static": (
+                self.total_gpu_memory_fraction
```
JiaxinD, line 73: The single-source fix routes `total_gpu_memory_fraction` (a device-total colocation budget, summed to 0.98 by placement) into sglang's `mem_fraction_static`, which is a fraction of FREE memory at boot, not of total; the framework already exposes `runtime.sglang_server_args.mem_fraction_static` for this knob. Default 0.85 is safe, but a non-default fraction (or a boot where the vocoder process already allocated) diverges from what placement validated, and that sglang field silently overrides this one with no cross-guard.

sglang_omni/models/higgs_tts/config.py
```diff
@@ -46,6 +51,9 @@ def mem_fraction_role_to_stage(cls) -> dict[str, str]:
             factory=f"{_PKG}.stages.create_audio_encoder_executor",
             factory_args={"device": "cuda"},
             gpu=0,
+            runtime=StageRuntimeConfig(
+                resources=StageResourceConfig(total_gpu_memory_fraction=0.03)
```
JiaxinD, line 55: Only the tts_engine factory declares `total_gpu_memory_fraction`, so these encoder 0.03 and vocoder 0.10 values stay declarative: placement validates a 0.98 budget that nothing enforces on those two processes. Separately this field accepts 1.0 while sglang's `mem_fraction_static` requires < 1, so a 1.0 here would reach the engine as a boot-time OOM.

tests/unit_test/scheduling/test_prefill_coalesce_gate.py
```diff
@@ -0,0 +1,153 @@
...
+def test_reaching_target_releases_before_deadline(upstream, clock):
+    sched = _StubScheduler(coalesce_requests=3, wait_ms=60.0)
+    sched.waiting_queue = [_req(100.0)]
+    assert sched.get_new_batch_prefill() is None
+
+    sched.waiting_queue = [_req(100.0)] * 3  # target reached, clock unchanged
+    assert sched.get_new_batch_prefill() is _UPSTREAM_BATCH
+
+
+def test_partial_admission_leftovers_keep_their_deadline(upstream, clock):
```
JiaxinD, line 110: This and `test_abort_does_not_hand_newcomers_an_expired_deadline` assert the end state (old stamps pass through, a lone newcomer is held) but never drive a real partial admission or abort, since upstream is mocked to a non-mutating sentinel. They would stay green if the queue-removal or enqueue-stamp path regressed; one real admit-then-re-enter cycle would pin the transition they name.

JiaxinD (commented):
Re-reviewed the updated head with a second model. Verified resolved: the config imports; the gate deadline is now keyed on the oldest queued request (stamp-on-miss mutates the request, and partial admission or an abort no longer restart the window); the tp_size>1 guard closes db-ol's per-rank clock-divergence hang; the compile warmup actually runs and exercises the compiled decode (CPU warmup codes are moved to the device inside codec.decode, so it is not a silent eager fallback); and the eager fallback restores a working decode. The O(N) gate scan is bounded to O(K) by the len>=K passthrough, so it is not a hot-loop cost.

What remains is the memory-fraction wiring (inline) plus a couple of test-coverage gaps. None of it changes the default higgs config, so it ships safe today; the mem-fraction mapping is a latent hazard only for non-default fractions. The boot-race between the engine and vocoder processes is unchanged by this PR (pre-existing).

sglang_omni/models/higgs_tts/engine_builder.py
JiaxinD: Fixed in e1b25695: a diverging mem_fraction_static override still wins but now logs the divergence from the placement-validated budget instead of silently shadowing it.

sglang_omni/models/higgs_tts/config.py
JiaxinD: The 1.0 case now fails fast in e1b25695 (builder rejects fractions outside (0, 1)); encoder/vocoder enforcement stays with the queued #1037 follow-up.

tests/unit_test/scheduling/test_prefill_coalesce_gate.py
JiaxinD: Added in e1b25695: a queue-mutating upstream drives a real admit-then-re-enter cycle, plus a newcomer-after-drain stamping case.

Ratish1 (approved):

PR #1073, opened 2026-07-16
[Perf] Generalize prefill admission coalescing across supported AR stages

MayDomine:
## Motivation

The prefill admission-coalescing gate added in #1071 lives in `OmniScheduler` and is model-agnostic, but only the Higgs TTS stage can enable it. Every OmniScheduler AR stage pays the same per-step prefill cost structure, and a second model already shows the gate generalizes: MOSS-TTS-Local gains +12.5% QPS with the identical mechanism.

## Modifications

Follow-up to #1071, rebased onto `main`:

1. Plumb `prefill_coalesce_requests` / `prefill_coalesce_wait_ms` through the remaining supported AR stages: MOSS-TTS-Local, MOSS-Transcribe-Diarize, Fun-ASR, and the Qwen3-Omni thinker. Defaults stay off.
2. Add typed per-stage `runtime.scheduling` config as the canonical YAML/CLI source. `SchedulingConfig` owns type and range validation; model configs explicitly opt supported stages in, while factory signatures remain the final capability guard. Existing #1071 `factory_args` / `runtime_overrides` remain compatible with a deprecation warning.
3. Add `--prefill-coalesce-requests` / `--prefill-coalesce-wait-ms`. CLI values apply to every opted-in stage and can be differentiated per stage in YAML; unsupported pipelines fail with `BadParameter` without importing optional model dependencies.
4. Cover CLI overrides, typed YAML transport, legacy compatibility, unsupported factories, no-op behavior, and lossy/non-finite value rejection.

## Accuracy Test

Scheduling-only change; numerics untouched (same argument as #1071 changes 1–2).

## Benchmark

Second-model validation: MOSS-TTS-Local on **seed-tts-eval-50 (en) voice-clone** — same closed-loop workload as the #1071 Higgs numbers (non-streaming `/v1/chat/completions`, ref audio + ref text per request, `max_new_tokens=512`). 1× H100, cap=conc=96, K=32/T=300 ms:

| config | QPS | lat mean | QPS×lat |
| --- | ---: | --- | ---: |
| baseline (3×90 s) | 9.00 | 10.66 s | 95.9 |
| + coalescing via the new CLI flags (2×90 s) | **9.98 (+10.9%)** | 9.75 s | 97.3 |

Measured end-to-end through the new `--prefill-coalesce-*` flags on this branch. Smaller than Higgs' +31% as expected — MOSS requests are ~10× heavier, so the amortizable per-request fixed cost is a smaller share. Higgs numbers are in #1071. Qwen3-Omni thinker and MOSS-TD are plumbed but unmeasured; the flag defaults off, so enabling remains a per-deployment decision.

## Checklist

- [x] Add unit tests.


sglang_omni/cli/serve.py
```diff
@@ -1164,6 +1196,30 @@ def serve(
...
+    prefill_coalesce_requests: Annotated[
+        int | None,
+        typer.Option(
+            "--prefill-coalesce-requests",
+            "--prefill_coalesce_requests",
+            help=(
+                "Hold prefill admission until this many requests are waiting "
+                "(or the oldest has waited --prefill-coalesce-wait-ms), "
+                "amortizing the per-step host cost. 0 disables (default). "
+                f"Available for {_ASYNC_DECODE_SUPPORTED_MODELS}."
```
JiaxinD: `_ASYNC_DECODE_SUPPORTED_MODELS` names four models but `_ASYNC_DECODE_FACTORIES` also contains the Qwen3-Omni thinker, which this PR plumbs. This help and the BadParameter above tell Qwen3-Omni users the flag is unsupported when it actually works; the constant needs updating.

sglang_omni/cli/serve.py
```diff
@@ -853,6 +853,38 @@ def apply_decode_mode_cli_overrides(
...
+    prefill_coalesce_requests: int | None,
+    prefill_coalesce_wait_ms: float | None,
+) -> PipelineConfig:
+    updates: dict[str, object] = {}
+    if prefill_coalesce_requests is not None:
+        if prefill_coalesce_requests < 0:
+            raise typer.BadParameter("--prefill-coalesce-requests must be >= 0")
+        updates["prefill_coalesce_requests"] = int(prefill_coalesce_requests)
+    if prefill_coalesce_wait_ms is not None:
+        if prefill_coalesce_wait_ms <= 0:
```
JiaxinD: `--prefill-coalesce-wait-ms nan` passes this check (`nan <= 0` is False) and the scheduler deadline then never fires, so admission holds until the running batch drains. `if not (prefill_coalesce_wait_ms > 0)` also rejects NaN.

sglang_omni/cli/serve.py
```diff
@@ -853,6 +853,38 @@ def apply_decode_mode_cli_overrides(
...
 
+def apply_prefill_coalesce_cli_overrides(
+    pipeline_config: PipelineConfig,
+    *,
+    prefill_coalesce_requests: int | None,
+    prefill_coalesce_wait_ms: float | None,
+) -> PipelineConfig:
+    updates: dict[str, object] = {}
+    if prefill_coalesce_requests is not None:
+        if prefill_coalesce_requests < 0:
```
JiaxinD: These ranges are only enforced on the CLI path; YAML factory_args reach OmniScheduler unvalidated, and negative values silently disable the gate. Validating in OmniScheduler.__init__ instead would make both paths fail fast.

sglang_omni/cli/serve.py
```diff
@@ -1164,6 +1196,30 @@ def serve(
...
+        ),
+    ] = None,
+    prefill_coalesce_wait_ms: Annotated[
+        float | None,
+        typer.Option(
+            "--prefill-coalesce-wait-ms",
+            "--prefill_coalesce_wait_ms",
+            help=(
+                "Upper bound on the extra time-to-first-token a queued request "
+                "pays for prefill coalescing. Default 60."
```
JiaxinD, line 1316: Setting only this flag is a silent no-op: the gate engages only when prefill_coalesce_requests >= 2 (1 also disables it, not just 0). Worth a note here or a warning in the override helper.

tests/unit_test/scheduling/test_cli_prefill_coalesce.py
```diff
@@ -0,0 +1,74 @@
...
+    return resolve_stage_factory_args(stage, config)
+
+
+@pytest.mark.parametrize(
+    ("config_cls", "stage_name"),
+    [
+        (HiggsTtsPipelineConfig, "tts_engine"),
+        (MossTTSLocalPipelineConfig, "tts_engine"),
+        (MossTranscribeDiarizePipelineConfig, "asr"),
+        (FunASRPipelineConfig, "asr"),
```
JiaxinD, line 52: Qwen3-Omni thinker is in the supported factory set but missing here; test_thinker_async_decode_flag.py shows the same pattern works with `Qwen3OmniPipelineConfig(model_path="dummy")`.

tests/unit_test/scheduling/test_cli_prefill_coalesce.py
```diff
@@ -0,0 +1,74 @@
+# SPDX-License-Identifier: Apache-2.0
+"""CLI override tests for prefill admission coalescing.
+
+``--prefill-coalesce-requests/--prefill-coalesce-wait-ms`` apply to every
+pipeline whose AR stage runs on OmniScheduler (the async-decode factory set);
```
JiaxinD: Several other AR stages also run on OmniScheduler without this plumbing (qwen3_asr, whisper_asr, ming thinker, qwen3 talker); the accurate scope is the async-decode factory set. Same applies to the PR title.

tests/unit_test/scheduling/test_cli_prefill_coalesce.py
```diff
@@ -0,0 +1,74 @@
...
+    assert args["prefill_coalesce_requests"] == 32
+    assert args["prefill_coalesce_wait_ms"] == 300.0
+
+
+def test_omitting_both_flags_is_a_noop():
+    config = HiggsTtsPipelineConfig(model_path="dummy")
+    apply_prefill_coalesce_cli_overrides(
+        config, prefill_coalesce_requests=None, prefill_coalesce_wait_ms=None
+    )
+    assert "prefill_coalesce_requests" not in _ar_stage_args(config, "tts_engine")
```
JiaxinD: This still passes if the None/None path wrote only wait_ms; snapshotting the full args dict before the call and asserting equality would pin the no-op.

sglang_omni/cli/serve.py
stephenkgli: maybe `if not math.isfinite(value) or value <= 0` is better, because `if not (prefill_coalesce_wait_ms > 0)`  can't reject inf

sglang_omni/cli/serve.py
MayDomine: Fixed—CLI and OmniScheduler now share `validate_prefill_coalesce_args`, so CLI and YAML paths enforce the same constraints from one validation implementation.

sglang_omni/scheduling/prefill_coalesce.py
```diff
@@ -0,0 +1,26 @@
...
+
+def validate_prefill_coalesce_args(
+    prefill_coalesce_requests: int | None,
+    prefill_coalesce_wait_ms: float | None,
+) -> tuple[int | None, float | None]:
+    """Validate and normalize coalescing arguments from any config entrypoint."""
+    requests = (
+        None if prefill_coalesce_requests is None else int(prefill_coalesce_requests)
+    )
+    if requests is not None and requests < 0:
```
AkazaAkane: prefill_coalesce_requests=1 passes validation, but get_new_batch_prefill() treats every value <= 1 as disabled. The CLI help currently says only 0 disables coalescing, so 1 becomes a silent no-op.

Could we either reject 1 here or explicitly document that both 0 and 1 disable the gate? I slightly prefer rejecting 1, since it is probably an accidental configuration.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -200,16 +201,21 @@ def __init__(
 
         # Note: (maydomine) coalescing gate: hold prefill until K requests wait
         # or the oldest has waited T ms; 0 disables.
-        if int(prefill_coalesce_requests) > 1 and int(server_args.tp_size) > 1:
+        requests, wait_ms = validate_prefill_coalesce_args(
+            prefill_coalesce_requests,
+            prefill_coalesce_wait_ms,
+        )
+        assert requests is not None and wait_ms is not None
```
AkazaAkane: Could we avoid using assert for validating these configuration values?

validate_prefill_coalesce_args() accepts None, so an explicit YAML value such as prefill_coalesce_requests: null would produce an AssertionError. Assertions also disappear under python -O, after which the failure would occur later during comparison or division.

It would be clearer to either make the scheduler validator require concrete values or raise a descriptive ValueError here.

sglang_omni/cli/serve.py
```diff
@@ -853,6 +855,47 @@ def apply_decode_mode_cli_overrides(
...
+        return pipeline_config
+    if prefill_coalesce_requests is None:
+        logger.warning(
+            "--prefill-coalesce-wait-ms alone does not enable coalescing; the "
+            "gate engages only when --prefill-coalesce-requests is >= 2"
+        )
+    matching_stages = [
+        stage
+        for stage in pipeline_config.stages
+        if stage.factory in _ASYNC_DECODE_FACTORIES
```
AkazaAkane: Could we define a separate _PREFILL_COALESCE_FACTORIES set instead of reusing _ASYNC_DECODE_FACTORIES here?

Async decode and prefill admission coalescing are separate capabilities, even though their supported factory sets currently happen to be identical. Reusing this set couples the two features: adding a new async-decode factory later would automatically make the CLI inject prefill_coalesce_* arguments, even if that factory has not added support for them.

A separate capability set would make the supported scope explicit and prevent future unsupported-keyword regressions.

MayDomine: Done in 7914c2d4: prefill coalescing now has its own `_PREFILL_COALESCE_FACTORIES` / `_PREFILL_COALESCE_SUPPORTED_MODELS`, defined as literal copies with no reference to the async-decode set, so adding a factory to one has no effect on the other.

sglang_omni/scheduling/omni_scheduler.py
MayDomine: Done in 7914c2d4: the assert is now a descriptive `ValueError` ("...must be concrete values at the scheduler...use 0 to disable coalescing instead of null"), so an explicit YAML `null` fails loudly at construction and the check survives `python -O`.

sglang_omni/scheduling/prefill_coalesce.py
MayDomine: Done in 7914c2d4, with the documented-no-op option rather than rejection: `1` is still accepted but logs a warning ("prefill_coalesce_requests=1 disables coalescing: the admission gate only engages at >= 2... Use 0 to disable explicitly, or >= 2 to enable"), and the 0/1-both-disable semantics are now documented in the validator docstring and the CLI help. Went this way instead of rejecting because 0 and 1 genuinely produce identical behavior in the gate, so hard-failing on 1 would break any existing config that uses it as an off switch — the warning surfaces the likely misconfiguration without the compat break. Also added a test covering warn-on-1 / silent-on-0.

sglang_omni/scheduling/prefill_coalesce.py
```diff
@@ -0,0 +1,40 @@
...
+    prefill_coalesce_wait_ms: float | None,
+) -> tuple[int | None, float | None]:
+    """Validate and normalize coalescing arguments from any config entrypoint.
+
+    Both 0 and 1 leave the admission gate disabled: the gate only engages at
+    >= 2, since a batch of one has nothing to coalesce with. 0 is the explicit
+    off switch; 1 is accepted but warns, as it is usually a misconfiguration.
+    """
+    requests = (
+        None if prefill_coalesce_requests is None else int(prefill_coalesce_requests)
```
JiaxinD: `int()` runs before the range check, so YAML `prefill_coalesce_requests: -0.5` truncates to 0 and silently disables the gate instead of tripping the `>= 0` guard (`2.9` becomes 2, `true` becomes 1, and `wait_ms: true` lands as 1.0). Worth rejecting bools and non integral counts before converting; the CLI is typed, so this only bites the YAML and direct construction paths.

sglang_omni/cli/serve.py
```diff
@@ -853,6 +870,47 @@ def apply_decode_mode_cli_overrides(
...
+        raise typer.BadParameter(str(exc)) from exc
+
+    updates: dict[str, object] = {}
+    if requests is not None:
+        updates["prefill_coalesce_requests"] = requests
+    if wait_ms is not None:
+        updates["prefill_coalesce_wait_ms"] = wait_ms
+    if not updates:
+        return pipeline_config
+    if prefill_coalesce_requests is None:
```
JiaxinD: This warns whenever the CLI flag is absent, including when the stage's YAML `factory_args` already sets `prefill_coalesce_requests: 32` and the operator is only tuning the wait from the command line. Checking the resolved stage args instead would stop it crying wolf on the YAML plus CLI mix.

JiaxinD (commented):
Two blockers before this can land: the head conflicts with `main` in `sglang_omni/cli/serve.py`, and `omni-ci-gate` fails in 4s (missing `run-ci`), so ASR / TTS / Qwen3-Omni / PR Test are all skipped and the two new test files have never executed in CI. The three fixes claimed for 7914c2d4 all check out.

sglang_omni/scheduling/prefill_coalesce.py
MayDomine: Done in 167d185a: bools are rejected for both fields, and non-integral request counts raise before `int()` can truncate — your exact cases (`-0.5`→0, `2.9`→2, `true`→1, `wait_ms: true`→1.0) are now ValueErrors and pinned in the test matrix. Integral floats like `32.0` still normalize.

sglang_omni/cli/serve.py
MayDomine: Done in 167d185a: the warning now inspects the targeted stages' resolved `factory_args` first — tuning only `--prefill-coalesce-wait-ms` while the YAML already sets `prefill_coalesce_requests >= 2` no longer warns; it only fires when no targeted stage has the gate enabled. Branch is also rebased onto current main (3f97d73b).

MayDomine:
@JiaxinD I reworked the configuration path based on the review discussion:

- `runtime.scheduling` is now the typed, per-stage source for both YAML and CLI.
- `SchedulingConfig` owns type/range validation; model configs opt supported stages in without importing optional model dependencies.
- Factory signatures remain the final capability check, and unsupported typed config fails explicitly.
- Existing #1071 `factory_args` / `runtime_overrides` remain compatible through the same validator, with a deprecation warning.
- Added coverage for legacy compatibility, mixed-source conflicts, unsupported factories, and lossy/non-finite values.

Remote config+scheduling tests pass (124 passed), and pre-commit is clean. Could you please re-review the PR?

sglang_omni/config/runtime.py
```diff
@@ -26,6 +34,15 @@ def resolve_stage_factory_args(
 
     args = resolve_stage_static_factory_args(stage_cfg, global_cfg)
     factory = import_string(stage_cfg.factory)
+    scheduling = stage_cfg.runtime.scheduling
+    if scheduling is not None and scheduling.model_dump(exclude_none=True):
+        signature = inspect.signature(factory)
+        missing = set(_PREFILL_COALESCE_KNOBS) - signature.parameters.keys()
```
JiaxinD: This raise only fires inside `resolve_stage_factory_args`, which nothing on the launch path calls: `mp_runner` uses `resolve_stage_static_factory_args` plus `resolve_stage_factory_arg_defaults`, and the child filters by signature in `stage_workers`. Running that exact parent plus child sequence on this head, an unsupported factory is called without the kwargs and no error is raised, so typed config on an unsupported stage is silently dropped in production.

sglang_omni/cli/serve.py
```diff
@@ -858,6 +859,81 @@ def apply_decode_mode_cli_overrides(
...
+            prefill_coalesce_wait_ms=prefill_coalesce_wait_ms,
+        )
+    except ValueError as exc:
+        raise typer.BadParameter(str(exc)) from exc
+    # Note (maydomine): Model configs opt in with a scheduling block so CLI
+    # validation does not import optional model dependencies.
+    matching_stages = [
+        stage
+        for stage in pipeline_config.stages
+        if stage.runtime.scheduling is not None
```
JiaxinD: Matching on the marker rather than the factory makes discovery depend on where the config came from: `ConfigManager.from_file` rebuilds supplied stages directly, so a full stage YAML written before this PR has no `runtime.scheduling` and the CLI rejects it with `no stage ... does`. Reproduced on this head by dropping the field from a dumped Higgs config; 7914c2d4 selected the same stage by factory identity regardless of source.

sglang_omni/cli/serve.py
```diff
@@ -858,6 +859,81 @@ def apply_decode_mode_cli_overrides(
...
+            "--prefill-coalesce-wait-ms alone does not enable coalescing; the "
+            "gate engages only when prefill_coalesce_requests is >= 2 (via "
+            "--prefill-coalesce-requests or runtime.scheduling in the config)"
+        )
+    cli_updates = cli_values.model_dump(exclude_none=True)
+    for stage in matching_stages:
+        scheduling = stage.runtime.scheduling
+        if scheduling is None:
+            continue
+        stage.runtime.scheduling = scheduling.model_copy(update=cli_updates)
```
JiaxinD: Writing the typed value onto a stage that still carries a legacy `factory_args.prefill_coalesce_requests` makes `_validate_runtime_sources` raise `sets ... through both` at startup, so a #1071 YAML plus `--prefill-coalesce-requests 64` fails and blames the user for a conflict the CLI created (executed on this head). Either clear the legacy key when the CLI writes the typed one, or let the CLI value win over the legacy source.

tests/unit_test/scheduling/test_cli_prefill_coalesce.py
```diff
@@ -0,0 +1,159 @@
...
+from sglang_omni.models.moss_transcribe_diarize.config import (
+    MossTranscribeDiarizePipelineConfig,
+)
+from sglang_omni.models.moss_tts_local.config import MossTTSLocalPipelineConfig
+from sglang_omni.models.qwen3_omni.config import Qwen3OmniPipelineConfig
+
+
+def _ar_stage_args(config: PipelineConfig, stage_name: str) -> dict[str, object]:
+    stage = next(s for s in config.stages if s.name == stage_name)
+    return resolve_stage_factory_args(stage, config)
```
JiaxinD: Every resolver level test goes through `resolve_stage_factory_args`, which the runtime never calls, which is why the unsupported-factory test passes while the launch path drops the value. Asserting through `resolve_stage_static_factory_args` plus `resolve_stage_factory_arg_defaults` plus `resolve_factory_signature_args` would pin the sequence `mp_runner` and `stage_workers` actually use.

JiaxinD (commented):
The validation rewrite is a real improvement: every lossy value from the last round is now rejected, the false wait-ms warning is gone, and the 16 schema tests pass locally. Two things still make typed config unreliable though, the capability check never runs on the launch path and CLI discovery now depends on a marker that older configs do not carry.

sglang_omni/config/runtime.py
MayDomine: Fixed in 577aa37b: the unsupported typed-field check now runs in `resolve_factory_signature_args`, the child-side resolver used by `stage_workers`, so production raises before calling a factory that would drop the configured value.

sglang_omni/cli/serve.py
MayDomine: Fixed in 577aa37b: CLI discovery again uses an independent `_PREFILL_COALESCE_FACTORIES` set, so full-stage YAML created before `runtime.scheduling` is still recognized. A regression test removes the field from a dumped Higgs config and verifies the CLI override.

sglang_omni/cli/serve.py
MayDomine: Fixed in 577aa37b: each explicit CLI field removes the matching legacy key from both `factory_args` and `runtime_overrides` before writing the typed value, so CLI precedence is preserved. Both legacy sources are covered.

tests/unit_test/scheduling/test_cli_prefill_coalesce.py
MayDomine: Fixed in 577aa37b: the test helper now composes `resolve_stage_static_factory_args` + `resolve_stage_factory_arg_defaults` + `resolve_factory_signature_args`, matching the actual parent/child launch sequence. Remote results: 44 targeted and 128 config+scheduling tests passed.

sglang_omni/cli/serve.py
```diff
@@ -32,7 +32,19 @@
     }
 )
 _ASYNC_DECODE_SUPPORTED_MODELS = (
-    "Higgs TTS, MOSS-TTS-Local, MOSS-Transcribe-Diarize, and Fun-ASR"
+    "Higgs TTS, MOSS-TTS-Local, MOSS-Transcribe-Diarize, Fun-ASR, "
+    "and the Qwen3-Omni thinker"
```
ChangyiYang: This string change breaks `tests/unit_test/higgs_tts/test_cli_decode_mode.py:94`, which still matches the old `MOSS-Transcribe-Diarize, and Fun-ASR` — `PR Test / unit-test` fails on all 3 attempts. Because the run is fail-fast and this trips at 10%, the new `tests/unit_test/scheduling/test_cli_prefill_coalesce.py` never executes in CI either.

MayDomine: Fixed in 89c98582 by updating the stale supported-model assertion to include the Qwen3-Omni thinker. The previously failing test now passes locally.

ChangyiYang:
Question on the `scheduling=SchedulingConfig()` markers in the five model configs — I want to check I'm reading their role correctly.

Running this branch's resolvers, an all-`None` `SchedulingConfig()` and the default `None` look indistinguishable at every read site: `resolve_stage_factory_arg_defaults` returns the same defaults dict, `_validate_runtime_sources` takes no action either way, and CLI discovery matches on factory identity (`_PREFILL_COALESCE_FACTORIES`) rather than on the marker. The guard in `resolve_factory_signature_args` also doesn't see an empty marker, so one placed on a stage whose factory can't accept the knobs is accepted silently.

Given the PR description frames these as "model configs explicitly opt supported stages in" — is the intent that they surface in `model_dump()` as self-documentation, or is something meant to gate on them? Asking mainly because otherwise the frozenset and the markers end up as two independent records of the same scope, free to drift.

Minor related note: `test_typed_scheduling_field_reaches_factory_args` builds its value via `type(stage.runtime.scheduling)(...)`, which quietly depends on the marker being present. The file already imports `SchedulingConfig`, so naming it directly there would decouple the test from this question either way.

sglang_omni/config/runtime.py
```diff
@@ -189,6 +227,27 @@ def _merge_factory_arg_overrides(
         args[key] = value
 
 
+def _normalize_legacy_scheduling_args(
+    stage_name: str,
+    args: dict[str, Any],
+) -> None:
+    legacy_values = {
+        knob: args[knob] for knob in _PREFILL_COALESCE_KNOBS if knob in args
```
ChangyiYang: `prefill_coalesce_requests: null` behaves differently depending on where it's written:

- in `factory_args` → `_normalize_legacy_scheduling_args` keys off `if knob in args`, so `None` reaches the factory and `OmniScheduler.__init__` raises.
- in `runtime.scheduling` → `resolve_stage_factory_arg_defaults` skips it on `is not None`, so the factory default (`0`) applies silently.

So "explicit null fails loudly" holds on the deprecated path but not the canonical one. Intended? `model_fields_set` would let the schema tell an omitted field from an explicit null, if they should agree.

MayDomine: Fixed now. `prefill_coalesce_requests: null` will throw a explicit error now.

JiaxinD (commented):
Verified on this head: `sglang_omni/config/` matches main, and the CLI, YAML and unsupported factory paths all behave through the real resolver sequence.

JiaxinD (approved):

stephenkgli:
Please rebase on the latest main branch.

Issue #1066, opened 2026-07-15
[DP+MPS consequent work] Case Study on H200 with Higgs TTS Model


jambow0320:
# [DP+MPS consequent work] Case Study on H200 with Higgs TTS Model

> Tracking: sgl-project/sglang-omni#1052 · Task 2 (additional recipes — other compute cards)
> Measurement: #912 tooling verbatim 

## Setup

| Item | Value |
| --- | --- |
| GPU | 1× H200 141GB SXM, driver 590.48.01 / CUDA 13 |
| Model / software | bosonai/higgs-tts-3-4b (`7556c17e`), sglang-omni `1618b7c`, sglang `0.5.12.post1` |
| Launch | equal-KV discipline of `examples/mps_dp/launch.sh`: `--max-total-tokens 100000` for every condition. |
| CPU | 32 NUMA-local physical cores split per replica; clients on disjoint physical cores, one per replica; no SMT siblings |
| Load / measurement | #912 `load_client.py` (closed-loop seed-tts voice clone, conc 96 per replica) + `run_condition.sh` (60 s nsys device window @10 kHz, SM Active % primary, qps corroboration); saturation held in every window (`running-req` med 62/64, `queue` med ≥30) |
| Reps | 2 per condition, fresh replicas + fresh MPS daemon per rep |

## 1. Throughput and SM Active % vs DP

| Config | qps (r1 / r2) | SM Active % | vs DP1, same column |
| --- | --- | --- | --- |
| DP1, MPS off | 50.2 / 50.8 | 20.0 / 20.2 | 1.00× |
| DP2, MPS off | 73.8 / 74.3 | 30.0 / 29.4 | 1.47× |
| DP3, MPS off | 82.8 / 83.7 | 33.3 / 33.8 | 1.65× |
| DP4, MPS off | 87.3 / 87.3 | 35.7 / 35.8 | 1.73× |
| DP5, MPS off | 89.6 / 89.4 | 36.4 / 36.4 | 1.77× |
| DP1, MPS on | 49.5 / 50.1 | 20.6 / 20.7 | 1.00× |
| DP2, MPS on | 93.9 / 93.2 | 39.1 / 40.1 | **1.88×** |
| DP3, MPS on | 132.1 / 129.6 | 55.6 / 55.9 | **2.63×** |
| DP4, MPS on | 159.7 / 157.6 | 69.2 / 69.8 | **3.19×** |
[...]

PR #1076, opened 2026-07-16
[MPS]: Add H100 and H200 MPS DP hardware profiles

Ratish1:
## Motivation

Add reusable hardware profiles for the same-GPU Higgs DP + CUDA MPS launcher. The profiles keep equal KV sizing explicit while separating portable model/runtime settings from host-specific GPU and CPU placement.

## Modifications

- Allow `examples/mps_dp/launch.sh` to start replicas from an SGLang Omni pipeline config while preserving the existing environment-based interface.
- Resolve the generation stage's configured `max_total_tokens` through `ConfigManager` so the launcher can continue verifying the exact KV capacity of every replica.
- Add validated Higgs profiles for H100 DP3 (`100000` KV tokens per replica) and H200 DP8 (`30000` KV tokens per replica), both with `max_running_requests=64` and `cuda_graph_max_bs=64`.
- Document how pipeline configs, launcher placement settings, and per-replica KV caps interact.

## Related Issues

Related to #1066  #921

## Benchmark & Profiling

H200 DP8 + MPS, full SeedTTS English dataset, concurrency 64 per replica:

| Runs | Requests | Failed | Mean throughput | Mean p50 / p95 / p99 |
|---:|---:|---:|---:|---:|
| 3 | 26,112 / 26,112 | 0 | 68.415 req/s | 6.356 / 10.144 / 12.519 s |

All replicas resolved exactly `30000` KV tokens, all eight MPS clients attached, and the GPU returned to idle after every run.

### DIFF
--- a/docs/basic_usage/mps_dp.md
+++ b/docs/basic_usage/mps_dp.md
@@ -28,20 +28,28 @@ Pick a GPU that is idle, then find its NUMA node from the PCI bus id (drm card o
 2. **Launch the replicas.**
 
 ```bash
-CORE_BLOCKS="0-9 10-19 20-29" MAX_TOTAL_TOKENS=100000 bash examples/mps_dp/launch.sh up
+CONFIG=examples/mps_dp/configs/higgs_h100_dp3.yaml N=3 CORE_BLOCKS="0-9 10-19 20-29" bash examples/mps_dp/launch.sh up
 ```
 
-The command above is the validated H100 Higgs DP3 recipe. The launcher uses three replicas by default, resolves the GPU's NUMA node, assigns one local port per replica, starts a private MPS daemon, waits for each replica's health check before starting the next, and verifies MPS attachment. The CPU blocks are specific to the tested host; derive the correct blocks for your own CPU topology.
+The command above is the validated H100 Higgs DP3 recipe. For the validated H200 Higgs DP8 recipe, use:
 
-The example leaves `--mem-fraction-static` unset and uses the model's existing default. Launching replicas sequentially avoids overlapping memory profiling and CUDA-graph capture during startup.
+```bash
+CONFIG=examples/mps_dp/configs/higgs_h200_dp8.yaml N=8 CORE_BLOCKS="0-3 4-7 8-11 12-15 16-19 20-23 24-27 28-31" bash examples/mps_dp/launch.sh up
+```
+
+The pipeline config supplies the model and per-replica runtime settings. The launcher environment supplies host-local placement, including the GPU, replica count, and CPU blocks. The launcher resolves the GPU's NUMA node, assigns local ports automatically, starts a private MPS daemon, waits for each replica's health check before starting the next, and verifies MPS attachment. The CPU blocks above are specific to the tested hosts; derive the correct non-overlapping blocks for your own CPU topology.
+
+Both profiles set `mem_fraction_static` to `0.85`; `MF` can override it. When `CONFIG` is unset, the existing `MODEL` and `MAX_TOTAL_TOKENS` interface remains available. Launching replicas sequentially avoids overlapping memory profiling and CUDA-graph capture during startup.
 
 Identical `--mem-fraction-static` flags do **not** mean identical KV capacity. `--mem-fraction-static` budgets model weights and the KV pool against the GPU memory available when each replica starts. Roughly, the profiled KV memory is the requested fraction of free memory measured before model loading, minus model and fixed runtime allocations. It is a per-replica budget, not an additive share of the card. Because replicas start sequentially, earlier ones have already reserved memory, so later ones see a smaller free pool and allocate fewer KV tokens even when every flag is the same (in one run, three sequential `mf=0.27` replicas received 97,503 / 53,149 / 20,961 KV tokens).
 
 ![What same-GPU DP spends in VRAM and what it reclaims](../_static/image/same-gpu-dp-vram.svg)
 
-Memory profiling does not coordinate KV allocation across independent replica processes. For `N > 1`, the launcher therefore requires one common `MAX_TOTAL_TOKENS` value and passes it to every replica as `--max-total-tokens`. SGLang treats this value as an upper bound; the launcher rejects startup unless every replica resolves exactly that capacity. The cap is independent of the request-level `max_new_tokens` limit and does not distribute requests between replicas.
+Memory profiling does not coordinate KV allocation across independent replica processes. For `N > 1`, the launcher therefore requires one common `max_total_tokens` value, either from the pipeline config or from `MAX_TOTAL_TOKENS`. SGLang treats this value as an upper bound; the launcher rejects startup unless every replica resolves exactly that capacity. The cap applies independently to each replica; it is not divided across the pool. It is also independent of the request-level `max_new_tokens` limit and does not distribute requests between replicas.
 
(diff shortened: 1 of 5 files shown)


examples/mps_dp/launch.sh
```diff
@@ -12,15 +15,20 @@
 #   bash examples/mps_dp/launch.sh down [RUN_ID]
 #
 # Environment for `up` (defaults in parentheses):
-#   MODEL (bosonai/higgs-tts-3-4b), MODEL_NAME (higgs), GPU_ID (0), N (3),
-#   BASE_PORT (8801),
+#   CONFIG: optional SGLang Omni pipeline config. When unset, MODEL is used.
+#   MODEL (bosonai/higgs-tts-3-4b; unavailable with CONFIG),
+#   MODEL_NAME (higgs without CONFIG; pipeline name with CONFIG), GPU_ID (0), N (3),
+#   BASE_PORT (8801), PYTHON_BIN (python),
```
JiaxinD: Documented default is `python` but the code defaults to `python3` (line 31). While touching this block, worth also stating that env `MAX_TOTAL_TOKENS` takes precedence over the config's value when both are set.

examples/mps_dp/launch.sh
```diff
@@ -358,10 +391,11 @@ up() {
...
   {
     echo "run_id=$run"; echo "gpu_id=$gpu"; echo "gpu_uuid=$uuid"; echo "numa_node=$node"
-    echo "model=$model"; echo "model_name=$model_name"; echo "n=$n"
-    echo "mem_fraction_static=${mf:-default}"
+    echo "config=${config:-none}"; echo "model_path=$model_path_manifest"
+    echo "model_name=${model_name:-from_config}"; echo "n=$n"
+    echo "mem_fraction_static_cli_override=${mf:-none}"
     echo "base_port=$base_port"; echo "core_blocks=$CORE_BLOCKS"
-    echo "max_total_tokens=${MAX_TOTAL_TOKENS:-auto/profiled}"
+    echo "max_total_tokens=${expected_max_total_tokens:-auto/profiled}"
```
JiaxinD, line 407: With CONFIG and N=1 the resolution at line 339 is skipped, so this records `auto/profiled` while the config's `max_total_tokens` actually caps the server. Resolving the config value regardless of N (and only gating the equality check on N>1) would keep the manifest truthful.

examples/mps_dp/launch.sh
```diff
@@ -421,15 +455,15 @@ up() {
...
     resolved_tokens=""
     resolved_tokens=$(grep -m1 -oE '#tokens:[[:space:]]*[0-9]+' "$log" \
       | grep -oE '[0-9]+$' || true)
     if [ "$n" -gt 1 ]; then
       [ -n "$resolved_tokens" ] \
         || die "replica $i is healthy but its resolved KV capacity is missing from $log"
-      [ "$resolved_tokens" = "$MAX_TOTAL_TOKENS" ] \
-        || die "replica $i resolved $resolved_tokens KV tokens; expected $MAX_TOTAL_TOKENS"
+      [ "$resolved_tokens" = "$expected_max_total_tokens" ] \
+        || die "replica $i resolved $resolved_tokens KV tokens; expected $expected_max_total_tokens"
```
JiaxinD, line 475: The `#tokens:` grep takes the first match, so CONFIG mode implicitly assumes a single-SGLang-engine pipeline: a qwen3-omni style config boots thinker and talker engines and the first line can belong to the wrong stage, failing (or coincidentally passing) the check against the generation-stage cap. The two higgs profiles are safe; worth a comment or guard documenting the single-engine assumption.

examples/mps_dp/config.py
```diff
@@ -0,0 +1,54 @@
...
+    return value
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("config", help="SGLang Omni pipeline config")
+    args = parser.parse_args()
+    try:
+        print(resolve_max_total_tokens(args.config))
+    except (OSError, ValueError) as exc:
```
JiaxinD: `ConfigManager.from_file` can also raise `yaml.YAMLError` on malformed YAML and `KeyError` when `config_cls` is missing (manager.py:131, 139), and those escape this handler as raw tracebacks. Separately, a small unit test loading both profiles and asserting 100000/30000 would pin the resolver contract these YAMLs depend on.

docs/basic_usage/mps_dp.md
```diff
@@ -28,20 +28,26 @@ Pick a GPU that is idle, then find its NUMA node from the PCI bus id (drm card o
...
 
-The validated H100 Higgs DP3 cap is `100000` tokens per replica. This value is a starting point for the tested configuration, not a universal hardware default. Recalculate the cap after changing the model, GPU, runtime, replica count, memory settings, or CUDA-graph settings. If a replica cannot allocate the common cap, lower it or reduce the replica count.
+The H100 Higgs DP3 profile uses `100000` tokens per replica. The H200 Higgs DP8 profile uses `30000` tokens per replica, preserving GPU memory headroom for the colocated audio encoder and vocoder. These values are specific to their configurations, not universal hardware defaults. Recalculate the cap after changing the model, GPU, runtime, replica count, memory settings, or CUDA-graph settings. If a replica cannot allocate the common cap, lower it or reduce the replica count.
```
JiaxinD: Worth one more sentence scoping the H200 value: 30000 tokens with `max_running_requests=64` is far below worst-case demand (64 x up to 2048 new tokens), so on workloads with longer outputs than SeedTTS the scheduler will retract and requeue rather than hold 64 truly concurrent requests.

examples/mps_dp/launch.sh
Ratish1: Fixed both: `PYTHON_BIN` now defaults to `python`, matching the existing SGLang Omni shell launchers, and the launcher states and implements `MAX_TOTAL_TOKENS` precedence over the config value.

examples/mps_dp/launch.sh
Ratish1: Fixed. CONFIG resolution now runs for `N=1` as well, so the manifest records the configured cap. A config with no explicit cap still preserves the existing `auto/profiled` behavior for `N=1`; the equal-cap requirement remains gated on `N>1`.

examples/mps_dp/launch.sh
Ratish1: Added a pre-launch guard for CONFIG with `N>1`. It uses the pipeline public SGLang role maps and rejects configs with multiple logical SGLang engine stages before the first-log-line equality check can become ambiguous. The shipped Higgs profiles resolve to one stage.

examples/mps_dp/config.py
Ratish1: Handled `yaml.YAMLError` and `KeyError` through the CLI error path. I kept this example-only change without a new unit-test file; I manually verified both shipped profile caps (100000/30000), environment override precedence, multi-engine rejection, and malformed YAML returning a clean argparse error without a traceback.

docs/basic_usage/mps_dp.md
Ratish1: Added the scope note: 30,000 is below the 64 × 2,048 worst-case output demand before input tokens, so 64 is an admission ceiling; if KV fills, SGLang retracts requests back to the waiting queue.

JiaxinD (approved):

Issue #1075, opened 2026-07-16
[DP+MPS consequent work] Multi-GPU H100 with Higgs TTS Model


lvy010:
ref #1052  

Multi-GPU DP3+MPS scales near-linearly with Higgs TTS. 

- [x] **Baseline**: Compare QPS and SM activity of DP1/DP2/DP3 schemes with MPS enabled/disabled on a single H100

- [x] **Intra-NUMA Scaling**: Test the Node0 dual-card DP3+MPS scheme to verify the linear scalability of single-card gains when sharing CPU cores.
- [x] **Cross-NUMA Scaling**: Test the Node0+Node1 cross-node configuration.

Record the QPS and SM activity for all configurations (referencing degraded performance data, output logs, failure cases, etc.) and provide startup scripts to support result reproduction.

lvy010:
### Baseline — H100 DP × MPS Reproduction

**Setup:**
- 1× H100 80GB, driver 580.126.20 / CUDA 13.0, sglang-omni `847b5019`, sglang `0.5.12.post1`
- Launcher `examples/mps_dp/launch.sh`
- Workload `/v1/audio/speech`. DP2/DP3 replicas each capped at `--max-total-tokens 100000`.

**Result:**

| Configuration | Aggregate QPS | Per-replica | SM Active % | MPS on÷off |
|---|---|---|---|---|
| DP1 | 21.1 / 19.6 | — | **~29%** (max 78) → GPU ~71% idle | — |
| DP2, MPS **off** | 12.58 | 6.4 / 6.2 (even) | 96% | — |
| DP2, MPS **on** | 19.16 | 9.7 / 9.5 (even) | 91% | **1.52×** |
| DP3, MPS **off** | 12.77 / 12.59 | ~4.2 each | ~99% | — |
| DP3, MPS **on** | 25.13 / 24.21 | ~8.0 each | 94% | **≈1.94×** |

**tips:**
- Manual MPS-off replicas must be launched with `setsid … </dev/null &` or they're SIGHUP-killed on terminal cleanup
- Orphaned `multiprocessing.spawn` workers hold VRAM after `pkill` and must be killed by venv-python path.

Next:  Intra-NUMA Scaling

lvy010:
### Intra-NUMA & Cross-NUMA Scaling — H100 Multi-GPU DP3+MPS

**Setup:**

[...]

PR #1077, opened 2026-07-17
fix(pipeline):auto kill orphaned workers on parent crash

lvy010:
- Problem (ref #1075): when the server is killed by SIGKILL/OOM/crash, `daemon=True` stage worker processes become orphaned and continue to occupy GPU memory, because the normal `mp_runner.stop()` path never executes.
- Solution: install `PR_SET_PDEATHSIG(SIGKILL)` at the top of `stage_process_main`, so the kernel reaps each worker the moment its parent dies.
- Local test: unit tests in `tests/unit_test/pipeline/test_stage_parent_death_signal.py` pass, and after `kill -9` on the parent process, GPU0 is freed from ~23 GB to 0 MiB with no leftover workers.


sglang_omni/pipeline/stage_workers.py
```diff
@@ -356,12 +357,36 @@ async def shutdown(self, join_timeout: float = 30.0) -> None:
             self._startup_error_channels.clear()
 
 
+def _install_parent_death_signal() -> None:
+    if sys.platform != "linux":
+        return
+    try:
+        import ctypes
+
+        parent_pid = os.getppid()
```
AkazaAkane: parent_pid is sampled inside the child after the spawn bootstrap. If the server dies before this target begins, the worker is already reparented, so both getppid() calls return the new parent and the worker survives. It can then continue into _run_process() and allocate GPU memory as an orphan. Please pass the expected parent PID from StageGroup.spawn()—for example, os.getpid() as an additional process argument—and compare against that after prctl. This covers parent death both before and during signal installation.

AkazaAkane (commented):
The PR description says tests/unit_test/pipeline/test_stage_parent_death_signal.py passes, but the branch changes only one production file—25 additions in stage_workers.py; the referenced test file is not committed.

sglang_omni/pipeline/stage_workers.py
lvy010: done — now pass the launcher PID into the worker and check it before and after `prctl`, exiting on mismatch.

lvy010:
> The PR description says tests/unit_test/pipeline/test_stage_parent_death_signal.py passes, but the branch changes only one production file—25 additions in stage_workers.py; the referenced test file is not committed.

thank u. test added

tests/unit_test/pipeline/test_stage_parent_death_signal.py
```diff
@@ -0,0 +1,114 @@
...
+    def _fail(*_a, **_k):  # pragma: no cover - must never run off Linux
+        raise AssertionError("ctypes must not be touched off Linux")
+
+    import ctypes
+
+    monkeypatch.setattr(ctypes, "CDLL", _fail)
+    stage_workers._install_parent_death_signal(EXPECTED_PARENT)
+
+
+def test_parent_death_signal_requests_sigkill_on_parent_exit(monkeypatch) -> None:
```
JiaxinD, line 36: These thoroughly test `_install_parent_death_signal` in isolation but not the plumbing this PR adds: nothing asserts `spawn()` threads `expected_parent_pid` through and `stage_process_main` installs it, so a silent revert to the racy state the earlier review fixed would stay green. A spy on `_install_parent_death_signal` driven through `spawn()` would lock it in.

lvy010: thank u.  added `test_spawn_threads_expected_parent_pid_into_install`

AkazaAkane (commented):
lgtm.

tests/unit_test/pipeline/test_stage_parent_death_signal.py
```diff
@@ -0,0 +1,175 @@
...
+    monkeypatch.setattr(stage_workers.sys, "platform", "linux")
+    monkeypatch.setattr(ctypes, "CDLL", lambda *a, **k: fake)
+    monkeypatch.setattr(stage_workers.os, "getppid", lambda: EXPECTED_PARENT)
+
+    stage_workers._install_parent_death_signal(EXPECTED_PARENT)
+
+    assert fake.calls, "prctl was not called"
+    option, sig = fake.calls[0][0], fake.calls[0][1]
+    assert option == 1
+    assert sig == signal.SIGKILL
```
JiaxinD: Forcing platform to linux makes the helper evaluate `signal.SIGKILL`, which Windows Python lacks, so this test and the died-during-install one fail on a Windows checkout (the AttributeError is swallowed and `fake.calls` stays empty). `getattr(signal, "SIGKILL", 9)` in the helper or a win32 skipif keeps the suite portable; the rest of tests/unit_test is currently platform-clean.

JiaxinD (approved):

AkazaAkane:
please fix lint with pre-commit

PR #1080, opened 2026-07-17
[Bugfix] Abort backend inference when a streaming response ends early

liuhuijiayou:
## Motivation

Streaming Chat Completions and Audio Transcriptions did not abort backend inference when their SSE response ended early, so the backend kept generating for a response that could no longer be delivered, holding GPU compute, KV cache, and scheduler capacity until the request finished on its own.

Affected endpoints:

* `POST /v1/chat/completions`
* `POST /v1/audio/transcriptions`

Starlette pulls a chunk from the response body iterator and then calls `send()` outside that generator, so a failed or cancelled `send()` leaves the SSE formatter suspended at its `yield` and never runs cleanup written inside it. Closing the outer generator is not sufficient either, because it does not deterministically close the nested generators underneath:

```text
response body iterator
  -> Chat / ASR formatter
    -> Client.completion_stream()
      -> Client.generate()
        -> Coordinator.stream()
```

A failing `send()` does not prove the client disconnected deliberately; Starlette maps any `OSError` to `ClientDisconnect`, and the same failure can come from a connection reset, a proxy timeout, or server shutdown. Cleanup therefore triggers on "this response can no longer be delivered" rather than on a specific disconnect cause.

## Modifications

* Add `_ClosableStreamingResponse`, which closes the response body iterator at the ASGI ownership boundary, covering `send()` failures, cancellation, and normal exit. Cleanup failures are logged and never mask the original exception.
* Use `contextlib.aclosing()` in `Client.generate()`, `Client.completion_stream()`, and both SSE formatters so closure cascades deterministically down to the coordinator.
* Abort in `Coordinator.stream()` when it exits while the request is still non-terminal. The coordinator submits and tracks the request, so it owns abort policy; Chat and ASR only format events and close their client stream.
* Skip the abort when the request is no longer in the active table, so normal completion sends none and `_submit_request()` failures before registration send none.
* Log abort failures with `logger.warning(..., exc_info=True)` instead of suppressing them, and always release the stream queue and completion future.

`Coordinator.abort()` is unchanged, and the TTS streaming path is untouched because it already managed this lifecycle correctly. Existing explicit `client.abort()` callers still broadcast exactly once, since `abort()` returns early once the request is no longer active.

## Related Issues

N/A

## Accuracy Test

N/A. This change does not modify model computation, sampling behavior, or generated output.

## Benchmark & Profiling

N/A. This is a request-lifecycle and resource-cleanup fix.

## Tests

Added deterministic CPU-only regression tests, with no ZMQ or CUDA:

* `test_chat_asgi_send_failure_aborts_backend_and_cleans_state` drives `StreamingResponse.__call__` with a failing body send, so the failure happens outside the formatter generator.
* `test_client_completion_stream_close_reaches_coordinator_owner` and `test_transcription_stream_close_reaches_coordinator_owner` close a real client stream and assert the abort reaches the coordinator.
* `test_coordinator_stream_early_close_aborts_and_cleans_state` asserts the abort broadcast and cleanup of `_requests`, `_stream_queues`, and `_completion_futures`.
* `test_coordinator_stream_natural_completion_does_not_abort` asserts no abort is sent.
* `test_coordinator_stream_abort_failure_is_logged` asserts a failing abort is logged and still releases coordinator state.

Verified separately that the original `send()` failure is preserved and that cancellation semantics are unchanged, including the case where the abort must await while the request task is already cancelled.

Validation commands:

```bash
pytest -q tests/unit_test/client/test_completion_rollout.py tests/unit_test/pipeline/test_coordinator.py tests/unit_test/serve/test_openai_api.py
pre-commit run --files sglang_omni/client/client.py sglang_omni/pipeline/coordinator.py sglang_omni/serve/openai_api.py tests/unit_test/pipeline/test_coordinator.py tests/unit_test/serve/test_openai_api.py
```

Results:

```text
pytest: 76 passed
pre-commit: all hooks passed
```

## Checklist

* [x] Format the code according to pre-commit.
* [x] Add unit tests.
* [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed.

### DIFF
--- a/sglang_omni/serve/openai_api.py
+++ b/sglang_omni/serve/openai_api.py
@@ -26,7 +26,7 @@ import math
 import time
 import uuid
 from collections.abc import Awaitable, Callable
-from contextlib import suppress
+from contextlib import aclosing, suppress
 from typing import Any, AsyncIterator
 
 from fastapi import (
@@ -46,6 +46,7 @@ from fastapi.responses import (
     Response,
     StreamingResponse,
 )
+from starlette.types import Receive, Scope, Send
 
 from sglang_omni.client import (
     Client,
@@ -138,6 +139,29 @@ def _is_bad_request_error(exc: Exception) -> bool:
 
 class _RequestBodyTooLarge(Exception):
     pass
+
+
+class _ClosableStreamingResponse(StreamingResponse):
+    """Close the response body iterator at the ASGI ownership boundary."""
+
+    async def __call__(
+        self,
+        scope: Scope,
+        receive: Receive,
+        send: Send,
+    ) -> None:
+        try:
+            await super().__call__(scope, receive, send)
+        finally:
+            await self._close_body_iterator()
+
+    async def _close_body_iterator(self) -> None:
+        try:
+            await _close_async_iterator_if_supported(self.body_iterator)
+        except asyncio.CancelledError:
+            logger.warning("Cancelled while closing streaming response body")
+        except Exception:
+            logger.warning("Failed to close streaming response body", exc_info=True)
 
 
 class VoiceUploadBodyLimitMiddleware:
@@ -650,7 +674,7 @@ def _register_chat_completions(app: FastAPI) -> None:
             audio_format = req.audio.get("format", "wav")
 
         if req.stream:
-            return StreamingResponse(
+            return _ClosableStreamingResponse(
                 _chat_stream(
                     client,
                     gen_req,
@@ -756,87 +780,93 @@ async def _chat_stream(
     model: str,
     req: ChatCompletionRequest,
     audio_format: str,
-):
+) -> AsyncIterator[str]:
     """Streaming chat completion generator (yields SSE events)."""
     role_sent = False
     requested_modalities = req.modalities or ["text"]
     finish_reason: str | None = None
     final_usage: UsageResponse | None = None
 
-    async for chunk in client.completion_stream(
+    chunk_stream = client.completion_stream(
         gen_req,
         request_id=request_id,
         audio_format=audio_format,
-    ):
-        # Capture finish info for the dedicated finish chunk after the loop.
-        # Some pipelines only emit a final aggregate chunk; do not drop its
-        # text/audio just because it already carries a finish reason.
-        if chunk.finish_reason is not None:
-            finish_reason = chunk.finish_reason
-            if chunk.usage is not None:
-                final_usage = UsageResponse(
-                    prompt_tokens=chunk.usage.prompt_tokens or 0,
-                    completion_tokens=chunk.usage.completion_tokens or 0,
-                    total_tokens=chunk.usage.total_tokens or 0,
+    )
+    async with aclosing(chunk_stream):
+        async for chunk in chunk_stream:
+            # Capture finish info for the dedicated finish chunk after the loop.
+            # Some pipelines only emit a final aggregate chunk; do not drop its
+            # text/audio just because it already carries a finish reason.
+            if chunk.finish_reason is not None:
+                finish_reason = chunk.finish_reason
+                if chunk.usage is not None:
+                    final_usage = UsageResponse(
+                        prompt_tokens=chunk.usage.prompt_tokens or 0,
+                        completion_tokens=chunk.usage.completion_tokens or 0,
+                        total_tokens=chunk.usage.total_tokens or 0,
+                    )
+                has_payload = (
+                    chunk.modality == "text"
+                    and bool(chunk.text)
+                    and "text" in requested_modalities
+                ) or (
+                    chunk.modality == "audio"
+                    and chunk.audio_b64 is not None
+                    and "audio" in requested_modalities
                 )
-            has_payload = (
+                if not has_payload:
+                    continue
+
+            delta = ChatCompletionStreamDelta()
+            emit = False
+
+            # Send role on first chunk
+            if not role_sent:
+                delta.role = "assistant"
+                role_sent = True
+                emit = True
+
+            # Text chunk
+            if (
                 chunk.modality == "text"
-                and bool(chunk.text)
+                and chunk.text
                 and "text" in requested_modalities
-            ) or (
+            ):
+                delta.content = chunk.text
+                emit = True
+
+            # Audio chunk
+            if (
                 chunk.modality == "audio"
                 and chunk.audio_b64 is not None
                 and "audio" in requested_modalities
-            )
-            if not has_payload:
+            ):
+                delta.audio = ChatCompletionAudio(
+                    id=f"audio-{request_id}",
+                    data=chunk.audio_b64,
+                )
+                emit = True
+
+            if not emit:
                 continue
 
-        delta = ChatCompletionStreamDelta()
-        emit = False
-
-        # Send role on first chunk
-        if not role_sent:
-            delta.role = "assistant"
-            role_sent = True
-            emit = True
-
-        # Text chunk
-        if chunk.modality == "text" and chunk.text and "text" in requested_modalities:
-            delta.content = chunk.text
-            emit = True
-
-        # Audio chunk
-        if (
-            chunk.modality == "audio"
-            and chunk.audio_b64 is not None
-            and "audio" in requested_modalities
-        ):
-            delta.audio = ChatCompletionAudio(
-                id=f"audio-{request_id}",
-                data=chunk.audio_b64,
+            stream_resp = ChatCompletionStreamResponse(
+                id=response_id,
+                created=created,
+                model=model,
+                choices=[
+                    ChatCompletionStreamChoice(
+                        index=0,
+                        delta=delta,
+                        finish_reason=None,
+                    )
+                ],
             )
-            emit = True
 
-        if not emit:
-            continue
-
-        stream_resp = ChatCompletionStreamResponse(
-            id=response_id,
-            created=created,
-            model=model,
-            choices=[
-                ChatCompletionStreamChoice(
-                    index=0,
-                    delta=delta,
-                    finish_reason=None,
-                )
-            ],
-        )
-
-        data = stream_resp.model_dump(exclude_none=True)
-        for choice in data.get("choices", []):
-            choice.setdefault("finish_reason", None)
-        yield f"data: {json.dumps(data)}\n\n"
+            data = stream_resp.model_dump(exclude_none=True)
+            for choice in data.get("choices", []):
+                choice.setdefault("finish_reason", None)
+            yield f"data: {json.dumps(data)}\n\n"
 
     # Finish chunk: empty delta + finish_reason.
     finish_resp = ChatCompletionStreamResponse(
@@ -1599,7 +1629,7 @@ def _register_transcriptions(app: FastAPI) -> None:
             )
             adapter = resolve_adapter(getattr(app.state, "architectures", None))
             duration_s = _probe_audio_duration(audio_bytes)
-            return StreamingResponse(
+            return _ClosableStreamingResponse(
                 _transcription_stream(
                     client,
                     gen_req,
@@ -1684,15 +1714,17 @@ async def _transcription_stream(
     post-processed transcript.
     """
     final_text: str | None = None
+    chunk_stream = client.generate(gen_req, request_id=request_id)
     try:
-        async for chunk in client.generate(gen_req, request_id=request_id):
-            if chunk.finish_reason is not None:
-                if isinstance(chunk.text, str) and chunk.text:
-                    final_text = chunk.text
-                continue
-            if chunk.modality == "text" and chunk.text:
-                event = TranscriptionTextDeltaEvent(delta=chunk.text)
-                yield f"data: {event.model_dump_json(exclude_none=True)}\n\n"
+        async with aclosing(chunk_stream):
+            async for chunk in chunk_stream:
+                if chunk.finish_reason is not None:
+                    if isinstance(chunk.text, str) and chunk.text:
+                        final_text = chunk.text
+                    continue
+                if chunk.modality == "text" and chunk.text:
+                    event = TranscriptionTextDeltaEvent(delta=chunk.text)
+                    yield f"data: {event.model_dump_json(exclude_none=True)}\n\n"
     except Exception as exc:
         logger.exception("Error streaming transcription for request %s", request_id)
         payload = {"type": "error", "error": {"message": str(exc)}}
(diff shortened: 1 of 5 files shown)


sglang_omni/serve/openai_api.py
```diff
@@ -743,107 +743,145 @@ async def _chat_stream(
...
+        # client disconnect (CancelledError / GeneratorExit) or an ordinary
+        # serving-layer processing error — so abort the in-flight request to
+        # stop the backend and release GPU / KV-cache / scheduler resources,
+        # then close the inner stream. The pending CancelledError,
+        # GeneratorExit, or ordinary exception continues to propagate; a
+        # failing abort must not mask it, so cleanup errors are swallowed.
+        if backend_completed:
+            await _close_async_iterator_if_supported(chunk_stream)
+        else:
+            with suppress(Exception):
```
AkazaAkane: Abort failures are fully silent. with suppress(Exception) swallows a failed abort with zero logging — in the exact scenario the AbortFailingChatStreamClient test simulates (abort backend unavailable), you'd leak the request with no trace. I'd replace the suppress with a try/except that does logger.warning("Failed to abort request %s after stream interruption", request_id, exc_info=True).

sglang_omni/serve/openai_api.py
```diff
@@ -743,107 +743,145 @@ async def _chat_stream(
     req: ChatCompletionRequest,
     audio_format: str,
 ):
-    """Streaming chat completion generator (yields SSE events)."""
+    """Streaming chat completion generator (yields SSE events).
```
AkazaAkane: ALL comment should follow style guide: [zhaochenyang20/sglang-diffusion-routing#32 (comment).](https://github.com/zhaochenyang20/sglang-diffusion-routing/issues/32#issuecomment-4918483294)

sglang_omni/serve/openai_api.py
```diff
@@ -743,107 +743,145 @@ async def _chat_stream(
...
-        usage=final_usage,
-    )
-    data = finish_resp.model_dump(exclude_none=True)
-    for choice in data.get("choices", []):
-        choice.setdefault("finish_reason", None)
-    yield f"data: {json.dumps(data)}\n\n"
-
-    yield f"data: {STREAM_DONE_SENTINEL}\n\n"
+        yield f"data: {STREAM_DONE_SENTINEL}\n\n"
+    finally:
```
Ratish1: Explicitly close the response body iterator at the ASGI boundary. This `finally` runs only if this generator is resumed, cancelled while awaiting the backend, or explicitly closed. If the client drops after Starlette has pulled a chunk and `send()` fails or is cancelled, `StreamingResponse` exits outside this generator without deterministically calling `aclose()`.

I have tested this and confirmed that direct generator close aborts, but ASGI send failure still leaks the request. Add a shared streaming response cleanup boundary that closes its body iterator on send failure, cancellation, and normal exit. Test it through `StreamingResponse.__call__`, not by calling this generator's `aclose()` directly.

sglang_omni/serve/openai_api.py
```diff
@@ -743,107 +743,145 @@ async def _chat_stream(
...
+        # ``backend_completed`` is only True once the inner iterator has
+        # finished on its own. In that case simply close the (already
+        # exhausted) generator. Otherwise the stream was interrupted early —
+        # client disconnect (CancelledError / GeneratorExit) or an ordinary
+        # serving-layer processing error — so abort the in-flight request to
+        # stop the backend and release GPU / KV-cache / scheduler resources,
+        # then close the inner stream. The pending CancelledError,
+        # GeneratorExit, or ordinary exception continues to propagate; a
+        # failing abort must not mask it, so cleanup errors are swallowed.
```
Ratish1: trim this comment

sglang_omni/serve/openai_api.py
```diff
@@ -743,107 +743,145 @@ async def _chat_stream(
...
+    (cancellation or otherwise) is always preserved; a failing abort during
+    cleanup never masks it.
+    """
     role_sent = False
     requested_modalities = req.modalities or ["text"]
     finish_reason: str | None = None
     final_usage: UsageResponse | None = None
 
-    async for chunk in client.completion_stream(
+    chunk_stream = client.completion_stream(
```
Ratish1, line 790: should we move backend lifetime ownership to the stream that admits the request. `Coordinator.stream()` submits and tracks the request, but its current `finally` only drops the local queue/future; wrapping abort policy around this Chat formatter leaves `Client.generate()`, `Client.completion_stream()`, and other consumers unsafe.

we can make`Coordinator.stream()` abort when it exits while the request is still non-terminal, and use `contextlib.aclosing()` in the Client wrapper generators so closure cascades deterministically. Chat and ASR should format events and close their client stream, not own backend abort policy.

tests/unit_test/serve/test_openai_api.py
```diff
@@ -644,6 +822,187 @@ async def _drive() -> None:
...
+            "chatcmpl-req-1",
+            0,
+            "qwen3-omni",
+            req,
+            "wav",
+        )
+        # Consume the first emitted SSE event, then disconnect (aclose).
+        first = await anext(stream)
+        assert first.startswith("data: ")
+        await stream.aclose()
```
Ratish1: this does not exercise a client disconnect; it directly invokes the favorable generator-close path implemented by this PR. The exact PR head passes this behavior while still leaking on ASGI send failure and coordinator early close.

add at least one ASGI-level disconnect/send-failure case that exits outside the formatter generator, plus a real `Client`/`Coordinator` early-close case. Assert the abort broadcast and cleanup of `_requests`, `_stream_queues`, and `_completion_futures`, then retain a natural-completion case that asserts no abort.

Ratish1 (commented):
Thanks for the contribution @liuhuijiayou . Lets also update the title of the PR

sglang_omni/serve/openai_api.py
liuhuijiayou: Fixed. The silent `suppress(Exception)` is gone. Abort failures are now logged with a stack trace, and the abort itself lives in `Coordinator.stream()`, which owns the request:

```python
finally:
    try:
        if request_id in self._requests:
            try:
                await self.abort(request_id)
            except Exception:
                logger.warning(
                    "Failed to abort request %s after stream interruption",
                    request_id,
                    exc_info=True,
                )
    finally:
        self._stream_queues.pop(request_id, None)
        self._completion_futures.pop(request_id, None)
```

Two properties this preserves:

* A failing abort no longer masks the original send failure or cancellation.
* The stream queue and completion future are released in the outer `finally`, so a broken control plane cannot leak coordinator state.

`test_coordinator_stream_abort_failure_is_logged` covers this by making `broadcast_abort()` raise and asserting both the warning and the state cleanup.

sglang_omni/serve/openai_api.py
liuhuijiayou: Thanks for the pointer. I went through the guide and applied it to this diff:

* Removed the long explanatory docstring block on `_chat_stream` and kept the original one-line docstring, so the comments no longer restate what the code already shows.
* The new class carries a single-line docstring only: `"""Close the response body iterator at the ASGI ownership boundary."""`
* Narrowed the `try/except` regions instead of wrapping large blocks, per P0 "do NOT over-catch".
* Added the missing return annotation on `_chat_stream` (`-> AsyncIterator[str]`), per P3 "all public APIs and function signatures must include type hints".
* No non-ASCII or debug comments in the diff; imports follow stdlib -> third-party -> local.

Most of the remaining line count in `_chat_stream` is the original body shifted one level deeper by the `async with aclosing(...)` scope, not new logic. `git diff -w` shows the actual change is small.

sglang_omni/serve/openai_api.py
liuhuijiayou: Good catch, and your diagnosis matches what I reproduced. Added `_ClosableStreamingResponse` as the shared cleanup boundary:

```python
class _ClosableStreamingResponse(StreamingResponse):
    """Close the response body iterator at the ASGI ownership boundary."""

    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
        try:
            await super().__call__(scope, receive, send)
        finally:
            await self._close_body_iterator()
```

Both streaming Chat and ASR responses now use it. Cleanup errors are logged inside `_close_body_iterator()`, so the `finally` cannot replace the original send failure or cancellation.

I used `finally` rather than catching `ClientDisconnect`, because on Starlette 0.48 `StreamingResponse.__call__` has two shapes and only one of them surfaces an exception:

```python
if spec_version >= (2, 4):
    try:
        await self.stream_response(send)
    except OSError:
        raise ClientDisconnect()
else:
    # task group + listen_for_disconnect, ends by cancellation
```

I verified three paths end-to-end through `response(scope, receive, send)` with a real `Client` and `Coordinator`:

| path | outcome | abort broadcast | leaked state |
| --- | --- | --- | --- |
| spec 2.4, `send()` raises `OSError` | `ClientDisconnect` | yes | none |
| spec 2.0, `http.disconnect` | returns | yes | none |
| outer request task cancelled | `CancelledError`, `task.cancelled()` is `True` | yes | none |

I also checked the case where the abort has to actually await (simulating a real ZMQ send) while the task is already cancelled; the abort still completes, because `aclose()` gets its own cleanup window.

`test_chat_asgi_send_failure_aborts_backend_and_cleans_state` drives this through `StreamingResponse.__call__` with a failing body send, not by calling the formatter's `aclose()`.

sglang_omni/serve/openai_api.py
liuhuijiayou: Done. That docstring block is removed; `_chat_stream` keeps its original one-line docstring.

sglang_omni/serve/openai_api.py
liuhuijiayou: Agreed, and I took this approach. Backend lifetime now lives in `Coordinator.stream()`, which is the code that submits and tracks the request:

```python
finally:
    try:
        if request_id in self._requests:
            try:
                await self.abort(request_id)
            except Exception:
                logger.warning(...)
    finally:
        self._stream_queues.pop(request_id, None)
        self._completion_futures.pop(request_id, None)
```

The `request_id in self._requests` check is what distinguishes early exit from normal completion: completion handling already pops the request from the active table, so a stream that finishes naturally sends no abort. It also avoids a pointless abort when `_submit_request()` fails before registering the request. `test_coordinator_stream_natural_completion_does_not_abort` pins that.

Closure cascades with `contextlib.aclosing()` in `Client.generate()` and `Client.completion_stream()`, so Chat, ASR, Realtime, and any other consumer get the same behavior just by closing their stream. Chat and ASR only format events and close the client stream now; they no longer call `client.abort()`.

I checked that both wrapper layers are actually required rather than assuming it. Reverting `client.py` to its previous state and re-running the tests:

| `client.py` state | Chat | ASR |
| --- | --- | --- |
| no `aclosing()` | leaks, `aborts == []` | leaks, `aborts == []` |
| `aclosing()` in `generate()` only | still leaks | fixed |
| `aclosing()` in both | fixed | fixed |

Chat leaks in the middle row because it goes through the extra `completion_stream()` hop, so closure stops there.

On interaction with the existing explicit `client.abort()` calls in TTS, `speech_ws`, and `realtime`: `abort()` returns early once the request is no longer active, so an explicit abort followed by closing the stream broadcasts exactly once. Same for the backend-failure path, where `_handle_completion()` already broadcast. `Coordinator.abort()` itself is unchanged.

tests/unit_test/serve/test_openai_api.py
liuhuijiayou: You were right that the old test only exercised the favorable generator-close path. Replaced it with tests that fail against the previous implementation.

ASGI-level, exits outside the formatter (`test_chat_asgi_send_failure_aborts_backend_and_cleans_state`):

```python
async def send(message):
    if message["type"] != "http.response.body":
        return
    body_ready.set()
    raise RuntimeError("client vanished during body send")

response_task = asyncio.create_task(response(scope, receive, send))
...
with pytest.raises(RuntimeError, match="client vanished during body send"):
    await response_task
assert [msg.request_id for msg in control_plane.aborts] == [request_id]
assert request_id not in coordinator._requests
assert request_id not in coordinator._stream_queues
assert request_id not in coordinator._completion_futures
```

Real `Client` / `Coordinator` early close:

* `test_client_completion_stream_close_reaches_coordinator_owner`
* `test_transcription_stream_close_reaches_coordinator_owner`
* `test_coordinator_stream_early_close_aborts_and_cleans_state`

Natural completion still asserts no abort:

* `test_coordinator_stream_natural_completion_does_not_abort` asserts `control_plane.aborts == []`

Plus `test_coordinator_stream_abort_failure_is_logged` for the failing-abort case.

All of these run on CPU with no ZMQ or CUDA, using the existing `RecordingCoordinatorControlPlane` fixture. They assert coordinator internals deliberately, since the leak being fixed is exactly those tables retaining entries.

```text
pytest: 76 passed
pre-commit: all hooks passed
```

liuhuijiayou:
Thanks @AkazaAkane @Ratish1 for the detailed review. Pushed a reworked implementation and replied inline; summary of what changed.

The abort policy moved out of the Chat formatter and into `Coordinator.stream()`, which is the code that submits and tracks the request. Chat and ASR now only format events and close their client stream. Closure cascades from the ASGI boundary down to the coordinator through `contextlib.aclosing()`, so the leak is covered even when `send()` fails outside the formatter generator.

| Feedback | Resolution |
| --- | --- |
| Silent abort failures | Logged with `logger.warning(..., exc_info=True)`; no `suppress` |
| Close body iterator at the ASGI boundary | `_ClosableStreamingResponse`, used by Chat and ASR |
| Move backend lifetime to the admitting stream | Abort lives in `Coordinator.stream()`; `aclosing()` in the client wrappers |
| Trim the comment | Removed; original one-line docstring kept |
| Comment style guide | Applied, plus the missing `_chat_stream` return annotation |
| Test through `StreamingResponse.__call__` | Added, with real `Client` / `Coordinator` early-close and natural-completion cases |

Two notes worth flagging:

A failing `send()` does not prove a deliberate client disconnect; Starlette maps any `OSError` to `ClientDisconnect`, and the same failure can come from a connection reset, proxy timeout, or shutdown. Cleanup is therefore keyed on the response no longer being deliverable. I renamed the PR title accordingly.

Both `aclosing()` call sites are load-bearing. With only `Client.generate()` wrapped, ASR is fixed but Chat still leaks, because Chat goes through the extra `completion_stream()` hop.

`Coordinator.abort()` is unchanged and the TTS path is untouched, since it already handled this lifecycle.

```text
pytest: 76 passed
pre-commit: all hooks passed
```

Ratish1 (approved):
Looks good @liuhuijiayou

Dayuxiaoshui:
@liuhuijiayou Looks good

PR #1082, opened 2026-07-17
[Bugfix] Avoid unretrieved completion-future errors for streaming requests

AkazaAkane:
## Motivation

The coordinator creates a completion future for both non-streaming and streaming
requests.

Non-streaming callers await this future through `submit()`, so coordinator error
paths must propagate exceptions through it. Streaming callers instead receive
errors through the stream queue and never await the completion future.

Calling `future.set_exception(...)` for a streaming request therefore leaves an
exception that is never retrieved. When the future is released, asyncio may log:

```text
Future exception was never retrieved
````

This can occur when a streaming request is:

* explicitly aborted;
* failed because a pipeline stage returned an error;
* failed by `fail_pending_requests()` after a coordinator-level fatal error.

This issue becomes more visible when streaming API disconnects actively abort
backend requests, as implemented in #1080.

## Modifications

* Added `_reject_completion_future()` to resolve coordinator futures according
  to the request type:

  * non-streaming requests retain `set_exception(...)`, preserving `submit()`
    error propagation;
  * streaming requests cancel the unused completion future because their error
    is delivered through the stream queue.
* Applied the shared helper to:

  * `Coordinator.abort()`;
  * `Coordinator.fail_pending_requests()`;
  * the failure path in `Coordinator._handle_completion()`.
* Preserved the existing stream error messages and coordinator cleanup behavior.

## Related Issues

Related to #1080.

## Accuracy Test

N/A. This change only affects coordinator error handling and does not modify
model computation or generated output.

## Benchmark & Profiling

N/A. This change does not affect the inference hot path.

## Tests

Added coordinator-level regression coverage for:

* aborting a streaming request cancels its unused completion future;
* dropping the aborted future does not trigger the asyncio exception handler;
* a mid-stream stage failure cancels the unused completion future;
* streaming consumers still receive the original error through the stream queue.

Existing tests continue to verify that non-streaming callers receive the original
exception through the completion future, including stage failures,
`fail_pending_requests()`, and explicit aborts.

Validation:

```bash
pytest -q tests/unit_test/pipeline/test_coordinator.py
pre-commit run --all-files --show-diff-on-failure
```

## Checklist

* [x] Add unit tests.
* [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed.
* [x] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.

## CI

`run-ci` label. This is a CPU-only coordinator lifecycle change; the targeted
unit test is listed above.

```
```

### DIFF
--- a/sglang_omni/pipeline/coordinator.py
+++ b/sglang_omni/pipeline/coordinator.py
@@ -127,9 +127,7 @@ class Coordinator:
         for request_id, info in list(self._requests.items()):
             info.state = RequestState.FAILED
             info.error = message
-            future = self._completion_futures.get(request_id)
-            if future is not None and not future.done():
-                future.set_exception(RuntimeError(message))
+            self._reject_completion_future(request_id, RuntimeError(message))
             queue = self._stream_queues.get(request_id)
             if queue is not None:
                 await queue.put(
@@ -418,6 +416,19 @@ class Coordinator:
             entry_info.control_endpoint,
         )
 
+    def _reject_completion_future(self, request_id: str, exc: BaseException) -> None:
+        # Note: (Akazaakane) Non-streaming callers await the completion future,
+        # so errors must be propagated with set_exception(). Streaming callers
+        # receive errors through the stream queue and never await that future;
+        # cancel it instead to avoid "Future exception was never retrieved".
+        future = self._completion_futures.get(request_id)
+        if future is None or future.done():
+            return
+        if request_id in self._stream_queues:
+            future.cancel()
+        else:
+            future.set_exception(exc)
+
     async def abort(self, request_id: str) -> bool:
         """Abort a request.
 
@@ -445,10 +456,9 @@ class Coordinator:
         info.state = RequestState.ABORTED
 
         # Resolve future with error
-        if request_id in self._completion_futures:
-            self._completion_futures[request_id].set_exception(
-                asyncio.CancelledError(f"Request {request_id} aborted")
-            )
+        self._reject_completion_future(
+            request_id, asyncio.CancelledError(f"Request {request_id} aborted")
+        )
         if request_id in self._stream_queues:
             await self._stream_queues[request_id].put(
                 CompleteMessage(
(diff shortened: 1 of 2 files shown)


sglang_omni/pipeline/coordinator.py
```diff
@@ -127,9 +127,7 @@ async def fail_pending_requests(self, error: BaseException | str) -> None:
         for request_id, info in list(self._requests.items()):
             info.state = RequestState.FAILED
             info.error = message
-            future = self._completion_futures.get(request_id)
-            if future is not None and not future.done():
-                future.set_exception(RuntimeError(message))
+            self._reject_completion_future(request_id, RuntimeError(message))
```
Ratish1, line 130: lets add a streaming regression case for this path. `fail_pending_requests()` is one of the failure sources identified in the PR, but its existing test is non-streaming and therefore exercises `set_exception()`, not the new cancellation branch. Start an active stream, fail pending requests, and assert that the stream receives the fatal error, the unused future is cancelled, and no unretrieved-future warning is emitted.

tests/unit_test/pipeline/test_coordinator.py
```diff
@@ -306,3 +307,101 @@ async def _run() -> None:
...
+        loop.set_exception_handler(
+            lambda _loop, context: handler_contexts.append(context)
+        )
+
+        task, error_sink, future = await _drive_stream_until_registered(
+            coordinator, "req-1"
+        )
+
+        assert await coordinator.abort("req-1") is True
+        await task
```
Ratish1: bound this and the stage-failure `await task` with `asyncio.wait_for(..., timeout=1)`, matching the existing coordinator stream test. If terminal queue delivery regresses, an unbounded await hangs the test process instead of failing at the lifecycle boundary.

Ratish1 (approved):
LGTM, just small comments to address

sglang_omni/pipeline/coordinator.py
AkazaAkane: Added.

tests/unit_test/pipeline/test_coordinator.py
AkazaAkane: Changed to asyncio.wait_for(..., timeout=1).

AkazaAkane:
@Ratish1 merge ready.

PR #1089, opened 2026-07-18
[Qwen3-Omni] Fix colocated launcher memory budgets

luojiaxuan:
Related to #985. Follow-up to #1070.

The qwen3-speech-server --colocated preset now applies the existing calibrated typed budgets before launch. Explicit AR memory overrides update the same typed budgets, and overcommitted values are rejected by placement.

Tests: 70 Qwen launcher and colocation tests passed. Black, isort, ruff, and diff checks passed.

### DIFF
--- a/examples/launchers/qwen3_omni.py
+++ b/examples/launchers/qwen3_omni.py
@@ -58,6 +58,14 @@ Examples:
       --prompt "Read me a bedtime story." --output audio.wav
 """
 
+_DEFAULT_COLOCATED_MEMORY_BUDGETS = {
+    "image_encoder": 0.025,
+    "audio_encoder": 0.025,
+    "thinker": 0.75,
+    "talker_ar": 0.12,
+    "code2wav": 0.02,
+}
+
 
 def _parse_thinker_tp_gpu_list(spec: str, tp_size: int) -> list[int]:
     try:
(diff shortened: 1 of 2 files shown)


examples/launchers/qwen3_omni.py
```diff
@@ -420,12 +449,19 @@ def launch_qwen_speech_server(args: argparse.Namespace) -> None:
...
     set_stage_gpu(config, "code2wav", gpu_code2wav)
-    _resolve_speech_mem_fractions(
+    thinker_mem_fraction, talker_mem_fraction = _resolve_speech_mem_fractions(
         config,
         global_mem_fraction_static=args.mem_fraction_static,
         thinker_mem_fraction_static=args.thinker_mem_fraction_static,
         talker_mem_fraction_static=args.talker_mem_fraction_static,
+        typed_runtime=args.colocated,
     )
+    if args.colocated:
```
SandyLuXY, line 459: This new colocated branch changes the memory contract, but the `--mem-fraction-static` help still says omitted flags let SGLang choose automatically. In the colocated path `_set_colocated_memory_budgets()` now supplies typed budgets, and the Qwen AR builders derive explicit SGLang `mem_fraction_static` values from them, so users will see calibrated budgets rather than automatic pool sizing. Please update the speech-server memory flag help to call out the `--colocated` exception.

SandyLuXY (commented):
Thanks for the fix! Left a minor comment.

PR #1090, opened 2026-07-18
[TTS] Add Audar-TTS V1 Turbo

luojiaxuan:
## Summary

- Add Audar-TTS-V1-Turbo with GGUF text generation and NeuCodec audio codecs.
- Reuse DeclarativeStateBase, ReferenceEncodeService, model capabilities, and the shared pipeline runtime.
- Add OpenAI speech request lowering, reference-audio transport coverage, and production contract tests.

Closes the Audar validation part of #985.

## Refactor result

Non-test, non-documentation LOC for the same model capability:

| Scope | Pre-T1 | Latest | Delta |
| --- | ---: | ---: | ---: |
| Minimal integration | 575 | 542 | -33 (-5.7%) |
| Production-equivalent | 797 | 619 | -178 (-22.3%) |
| Production capability premium | 222 | 77 | -145 (-65.3%) |

The control implementation is at ec25335e on luojiaxuan/audar-tts-pre-t1-production. The LOC result stacks on #1096 at 7aa65583; correctness and performance remain tied to the locked 98b63433 snapshot.

## Correctness

- All 28 paired requests produced identical 285-code sequences and identical 24 kHz waveforms.
- A separate 50-sentence FLEURS Arabic run produced 50/50 identical acoustic-code, float-waveform, and PCM-WAV hashes.

The target Arabic text is the reference and Qwen3-ASR-1.7B supplies the Arabic
hypothesis; no translation is used. Every paired waveform hash matches, and the
ASR artifact verifies each transcribed WAV hash. Latest was transcribed once;
pre-T1 inherits this result from exact WAV identity. The same ASR path was run
on the selected human FLEURS recordings for recognizer context:

| Audar TTS metric | Pre-T1 | Latest | Delta |
| --- | ---: | ---: | ---: |
| Arabic corpus WER | 5.43% | 5.43% | 0.00 pp |
| Arabic corpus CER | 1.46% | 1.46% | 0.00 pp |
| Arabic corpus BLEU | 88.75 | 88.75 | 0.00 |
| Arabic chrF++ | 95.57 | 95.57 | 0.00 |

Latest is the direct ASR evaluation. Pre-T1 uses the same metrics because all
50 PCM-WAV inputs are byte-identical to latest; it is not an unmeasured quality
claim.

The FLEURS source-audio WER is 8.91%, but it is not used as a quality delta:
the human audio can diverge from its written transcript and has a different
sample rate. Hash identity establishes refactor equivalence; ASR is a separate
absolute-intelligibility check.

## Performance

H100, one serial GGUF engine, four alternating runs, seven requests per run with the first request excluded:

| Metric | Pre-T1 | Latest | Delta |
| --- | ---: | ---: | ---: |
| Stage-sum latency | 1.0779 s | 1.0764 s | -0.13% |
| RTF | 0.18910 | 0.18885 | -0.13% |
| Engine throughput | 293.95 codes/s | 294.43 codes/s | +0.16% |

The pre-T1 inter-run latency spread is about 0.55%, so this result is performance parity, not a speedup.
The quality corpus is not used for performance: its fixed pre-T1-first order
includes startup ramp; after samples 0-19, throughput is 274.78 vs 274.02 codes/s.

## Verification

- Audar, shared tensor-hook, capability, and pipeline-state tests: 79 passed.
- Pre-T1 production tests: 14 passed.
- Ruff format and lint passed.
- Raw machine JSON, pytest logs, and hashes are under benchmarks/audar_tts/artifacts.


benchmarks/audar_tts/artifacts/performance/latest-order1.json
zhaochenyang20, line 1: remove this? why should we keep `benchmarks/audar_tts`?

examples/configs/audar_tts_turbo.yaml
zhaochenyang20, line 1: if a config is such simple, could we remove this?

zhaochenyang20 (changes requested):

benchmarks/audar_tts/artifacts/performance/latest-order1.json
luojiaxuan: Done, removed the Audar benchmark harness and generated artifacts from this PR.

examples/configs/audar_tts_turbo.yaml
luojiaxuan: Kept the two required fields because the model repo has no resolvable architecture metadata. Removed the redundant relay_backend default and added a config load test.

sglang_omni/models/audar_tts/request_builders.py
```diff
@@ -0,0 +1,141 @@
...
+
+    reference = references[0] if references else None
+    if reference is None and tts_params.get("ref_audio") is not None:
+        reference = _reference_from_value(tts_params["ref_audio"])
+    if reference is None:
+        raise ValueError("Audar-TTS requires reference audio")
+
+    reference_text = reference.get("text") or tts_params.get("ref_text")
+    if not isinstance(reference_text, str) or not reference_text.strip():
+        raise ValueError("Audar-TTS requires the reference transcript")
```
SandyLuXY, line 48: This rejects an Audar public request only after it has entered the backend. Consider moving Audar's public request checks into the pre-backend speech validation path.

sglang_omni/serve/protocol.py
```diff
@@ -239,6 +239,7 @@ class GenerateResponse(BaseModel):
 SUPPORTED_TTS_LANGUAGES = frozenset(
     {
         "Auto",
+        "Arabic",
```
SandyLuXY, line 242: This widens the global `/v1/audio/speech` language enum for every TTS model, but Audar's README says it ignores `language`, while Qwen3-TTS resolves non-auto languages through `self.config.codec_language_id[language.lower()]`. A Qwen3-TTS request with `language=Arabic` can now pass public validation and fail inside the model path instead of failing fast.

zhaochenyang20 (approved):

sglang_omni/models/audar_tts/request_builders.py
luojiaxuan: Fixed in #1104. Audar now requires exactly one transcribed reference in SpeechRequestValidator before backend dispatch.

sglang_omni/serve/protocol.py
luojiaxuan: Fixed in #1104. Arabic is now scoped to Audar, so Qwen3-TTS rejects it during public request validation.

PR #1095, opened 2026-07-19
[Fun-ASR] Add pre-LM audio encoder service with caching (FPR-3)

AkazaAkane:
## Motivation

Fun-ASR performs full-context audio encoding before language-model decoding. Previously, this work occurred after LM admission, limiting batching efficiency under concurrent short-utterance workloads and repeating the same encoding work for identical audio requests.

This PR adds a pre-LM encoder service that:

- Moves complete audio encoding ahead of LM admission.
- Batches concurrently arriving encoder requests.
- Deduplicates concurrent requests for identical audio through single-flight execution.
- Caches complete LM-ready embeddings in a bounded CPU LRU cache.
- Prevents incomplete or failed encodings from being admitted to the LM stage.

The cache stores only complete encoder/adaptor outputs. It does not reuse a prefix encoding for a longer partial audio window.

## Modifications

- Add `FunAsrPreLMEncoderService` for batched audio encoding before LM admission.
- Add single-flight deduplication so concurrent requests for identical audio share one encoder execution.
- Add a bounded CPU LRU cache with configurable entry and byte limits.
- Build cache namespaces from the model, frontend configuration, dtype, device type, and multimodal attention backend.
- Key individual cache entries using the normalized audio fingerprint.
- Validate cached embedding shape, token count, and dtype before reuse.
- Attach precomputed embeddings directly to the multimodal request and release the original CPU feature tensor after encoding.
- Use a dedicated CUDA stream for pre-LM encoder execution.
- Fall back to per-item encoding when a batched encode fails, while propagating unrecoverable failures to the affected requests.
- Add encoder queue, batching, latency, failure, and cache statistics.
- Make `StageOutputCache` safe for concurrent access.
- Extract `_get_audio_feature_uncached()` so the regular and pre-LM paths share the same underlying encoder implementation.
- Add `audio_fingerprint` and `num_audio_tokens` to Fun-ASR request metadata.
- Add unit coverage for encoder batching, cache behavior, concurrent deduplication, request construction, model integration, and failure handling.

## Related Issues

Related to #924 — F-PR3 pre-LM encoder service.

## Accuracy Test

Evaluated on the complete **1,088-sample SeedTTS English dataset**, containing **666 unique audio files**, using `FunAudioLLM/Fun-ASR-Nano-2512-hf` on an NVIDIA H100 80GB at concurrency 32.

Current main, the PR cold-start pass, and the PR warm-cache pass all completed the full dataset with:

- **0 skipped requests**
- **0.01710 corpus WER (1.710%)**

The pre-LM encoder and cache paths therefore preserve the current-main accuracy result.

### H100 concurrency-32 results

| Metric | Current main | PR cold-start | PR warm |
|---|---:|---:|---:|
| Throughput | 25.962 samples/s | 34.738 samples/s | 86.789 samples/s |
| Improvement | — | **+33.80% / 1.338×** | **+234.29% / 3.343×** |
| Wall time | 41.907 s | 31.320 s | 12.536 s |
| Mean latency | 1.229 s | 0.914 s | 0.362 s |
| p95 latency | 1.802 s | 1.114 s | 0.504 s |
| p99 latency | 3.015 s | 1.896 s | 0.800 s |
| Mean RTF | 0.26585 | 0.19965 | 0.07828 |
| Corpus WER | 0.01710 | 0.01710 | 0.01710 |
| Skipped | 0 | 0 | 0 |

Relative to current main:

- The PR cold-start pass improved throughput by **33.80%**, reduced mean latency by **25.6%**, and reduced wall time by **25.3%**.
- The PR warm-cache pass improved throughput by **234.29%**, or **3.343×**, reduced mean latency by **70.5%**, and reduced wall time by **70.1%**.
- All runs produced identical corpus WER and completed every request.

The two PR passes used one fresh server process. The first pass started with an empty encoder cache, and the second pass reused the cache populated by the first pass.

### 20260720 Update
Benchmark complete on one H100 80GB at concurrency 32, comparing the PR’s pinned main base (f916b86c) with PR head (e6f49902).

| Cold-cache metric | Main | PR | Change |
| --- | --- | --- | --- |
| Throughput | 32.05 samples/s | 41.06 samples/s | +28.1% |
| Wall time | 33.95 s | 26.50 s | −21.9% |
| Mean latency | 0.996 s | 0.771 s | −22.7% |
| p95 latency | 1.489 s | 0.892 s | −40.1% |
| p99 latency | 2.047 s | 0.972 s | −52.5% |
| Mean SM busy | 25.3% | 43.3% | +18.0 pp |
| Time below 10% SM | ~1.10 s | ~0.70 s | −0.40 s |
| Peak GPU memory | 73,193 MiB | 71,909 MiB | −1,284 MiB |
| Corpus WER | 1.710% | 1.710% | identical |

Both runs completed all 1,088 requests with no skips.

I also ran a strict no-hit control containing 666 SHA-256-verified unique audio files. PR telemetry showed zero hits and zero single-flight merges. In that case:

- Throughput was effectively neutral: +0.17%
- Mean latency: −1.7%
- p95 latency: −31.2%
- p99 latency: −49.2%
- Mean SM busy: 28.1% → 43.5%
- GPU idle time: ~0.70 s → ~0.41 s

So the defensible conclusion is: the PR clearly improves standard cold-cache throughput through pre-LM batching plus same-pass single-flight reuse. With every request strictly unique, it improves GPU utilization and tail latency substantially, but does not materially improve aggregate throughput.

## Strict no-hit control — complete results

Setup: 666 SHA-256-verified unique SeedTTS EN audio files, concurrency 32, fresh server/cache, one H100 80GB. PR telemetry confirmed hits=0, merged=0, and failed=0.

### Performance

| Metric | Main (f916b86c) | PR (e6f49902) | Change |
| --- | --- | --- | --- |
| Throughput | 32.1777 samples/s | 32.2310 samples/s | +0.17% |
| Wall time | 20.6975 s | 20.6634 s | −0.17% |
| Mean latency | 0.9910 s | 0.9739 s | −1.72% |
| Median latency | 0.9592 s | 0.9838 s | +2.56% |
| p95 latency | 1.5430 s | 1.0619 s | −31.18% |
| p99 latency | 2.1765 s | 1.1053 s | −49.22% |
| Total request latency | 659.977 s | 648.605 s | −1.72% |
| Mean RTF | 0.21514 | 0.21562 | +0.22% |
| Median RTF | 0.20944 | 0.21454 | +2.43% |
| p95 RTF | 0.3326 | 0.2998 | −9.86% |
| Audio processed | 3,139.096 s | 3,139.096 s | identical |

### GPU utilization

| Metric | Main | PR | Change |
| --- | --- | --- | --- |
| Mean SM busy | 28.10% | 43.53% | +15.43 pp / +54.91% |
| Median SM busy | 30.0% | 43.0% | +13.0 pp |
| p95 SM busy | 33.0% | 58.0% | +25.0 pp |
| Active-period mean SM busy | 29.07% | 44.41% | +15.33 pp |
| Exactly-idle samples | 3.35% | 1.98% | −1.37 pp |
| Exactly-idle equivalent time | ~0.70 s | ~0.41 s | −0.29 s |
| Samples below 10% SM | 3.35% | 1.98% | −1.37 pp |
| Mean GPU memory | 73,002.3 MiB | 71,901.6 MiB | −1,100.7 MiB |
| Peak GPU memory | 73,062.9 MiB | 71,963.5 MiB | −1,099.4 MiB |
| Mean power | 144.03 W | 161.64 W | +17.61 W |
| Peak power | 148.11 W | 167.67 W | +19.56 W |
| Telemetry window | 20.945 s | 20.923 s | −0.10% |
| NVML samples | 209 | 202 | — |
| Sampling interval | 100 ms | 100 ms | identical |

### Correctness

| Metric | Main | PR | Change |
| --- | --- | --- | --- |
| Evaluated | 666 | 666 | identical |
| Skipped | 0 | 0 | identical |
| Corpus WER | 0.017715 | 0.017715 | identical |
| Mean per-sample WER | 0.018156 | 0.018156 | identical |
| Median per-sample WER | 0 | 0 | identical |
| p95 per-sample WER | 0.1000 | 0.1000 | identical |
| Maximum per-sample WER | 0.2857 | 0.2857 | identical |
| Mean audio duration | 4.7134 s | 4.7134 s | identical |

### Cache telemetry

Cold-start pass:

- 666 cache misses
- 421 concurrent requests merged through single-flight
- 1 ordinary cache hit
- 422 of 1,088 requests avoided an encoder execution, for **38.79% effective reuse**
- The service's narrow `hits / (hits + misses)` metric was 0.15% because it excludes single-flight merges

The 666 misses match the dataset's 666 unique audio files. Single-flight deduplication merged repeated audio requests that arrived before the first encoding of the corresponding audio had completed.

Warm-cache pass:

- 1,088 ordinary cache hits
- 0 misses
- 0 single-flight merges
- **1,088 / 1,088 cache hits**

Once the cache was fully populated, every request reused a complete cached encoder output without launching another encoder execution.

## Checklist

- [ ] Format the code according to pre-commit.
- [x] Add unit tests.

## CI

The complete unit suite was verified on this PR:
- Effective result: **2,125 / 2,125 unit tests passed**

if labeled.


sglang_omni/models/fun_asr/encoder_service.py
WilsonZheng0327, line 1: `sglang_omni/scheduling/reference_encoder.py` already exists, wondering if there's a specific reason why this is a standalone class.

sglang_omni/models/fun_asr/encoder_service.py
```diff
@@ -0,0 +1,358 @@
...
+            batch = self._drain_batch()
+            dequeue_time = time.perf_counter()
+            queue_waits = [dequeue_time - enqueued_at for _, _, enqueued_at in batch]
+            with self._lock:
+                self._queue_wait_count += len(queue_waits)
+                self._queue_wait_total_s += sum(queue_waits)
+                self._queue_wait_max_s = max(
+                    self._queue_wait_max_s, max(queue_waits, default=0.0)
+                )
+            items = [item for item, _, _ in batch]
```
Ccyest, line 319: The worker is a daemon thread with `while True` blocking on `queue.get()` and no `close()`/sentinel exit path. this PR's `test_encoder_service.py` constructs a fresh service per test case, so one test process accumulates a dozen-plus threads that never exit.

**Fix:**

```python
_SHUTDOWN = object()

def close(self) -> None:
    self._queue.put(_SHUTDOWN)
    self._thread.join(timeout=5)
```

sglang_omni/models/fun_asr/encoder_service.py
```diff
@@ -0,0 +1,358 @@
...
+    def _drain_batch(
+        self,
+    ) -> list[tuple[Any, concurrent.futures.Future[torch.Tensor], float]]:
+        batch = [self._queue.get()]
+        while True:
+            try:
+                batch.append(self._queue.get_nowait())
+            except queue.Empty:
+                break
+        return batch
```
WilsonZheng0327: unbounded drain, following moss something like capping the drain at max_batch_size=8 with a 4 ms wait

sglang_omni/models/fun_asr/encoder_service.py
Ccyest: This is shared question on MTD too. Please fix it here and we will extract this module out during refactoring.

sglang_omni/models/fun_asr/sglang_model.py
```diff
@@ -376,6 +376,13 @@ def pad_input_ids(self, input_ids: List[int], mm_inputs: MultimodalInputs):
         return self.pattern.pad_input_tokens(input_ids, mm_inputs)
 
     def get_audio_feature(self, items: List[MultimodalDataItem]) -> torch.Tensor:
+        return self._get_audio_feature_uncached(items, None)
+
+    def _get_audio_feature_uncached(
+        self,
+        items: List[MultimodalDataItem],
+        forward_batch: Optional[ForwardBatch],
```
Ccyest: nit: fun_asr has no encoder graph/compile machinery, so here the `forward_batch` parameter is not used.

sglang_omni/models/fun_asr/request_builders.py
```diff
@@ -261,6 +264,9 @@ def request_builder(payload: StagePayload) -> FunASRRequestData:
         ]
         audio_item.offsets = [(audio_start, audio_start + num_audio_tokens - 1)]
 
+        if audio_encoder_service is not None:
+            audio_encoder_service.encode_item(audio_item)
+
```
Ccyest: `encode_item` runs before the `_request_token_budget` validation and the context-length check: a request doomed to rejection (illegal `max_new_tokens` or over context) still pays a full GPU encode and occupies cache space. Move `encode_item` after all validations, just before `Req` construction;

sglang_omni/models/fun_asr/encoder_service.py
```diff
@@ -0,0 +1,358 @@
...
+    mask = getattr(item, "feature_attention_mask", None)
+    if mask is not None:
+        return max(int(fun_asr_low_frame_rate_length(int(mask.sum()))), 1)
+    feature = getattr(item, "feature", None)
+    if isinstance(feature, torch.Tensor) and feature.dim() >= 1:
+        return max(int(fun_asr_low_frame_rate_length(int(feature.shape[-1]))), 1)
+    return None
+
+
+class FunAsrPreLMEncoderService:
```
WilsonZheng0327: name change => FunASRPreLMEncoderService, capitalization consistency

sglang_omni/models/fun_asr/encoder_service.py
```diff
@@ -0,0 +1,358 @@
...
+    num_tokens = getattr(item, "num_audio_tokens", None)
+    if num_tokens is not None:
+        return int(num_tokens)
+    mask = getattr(item, "feature_attention_mask", None)
+    if mask is not None:
+        return max(int(fun_asr_low_frame_rate_length(int(mask.sum()))), 1)
+    feature = getattr(item, "feature", None)
+    if isinstance(feature, torch.Tensor) and feature.dim() >= 1:
+        return max(int(fun_asr_low_frame_rate_length(int(feature.shape[-1]))), 1)
+    return None
```
Ccyest: nit: In production the request builder unconditionally writes `num_audio_tokens` and `audio_fingerprint` (`request_builders.py:247-251`), so the mask/feature fallbacks in `_expected_audio_tokens` are reachable only by unit-test

sglang_omni/models/fun_asr/stages.py
```diff
@@ -43,7 +47,10 @@ def create_sglang_fun_asr_executor(
     mm_embedding_cache_size_bytes: int = 0,
     enable_torch_compile: bool = False,
     mm_attention_backend: str | None = None,
-    request_build_max_workers: int = 2,
+    enable_pre_lm_encoder: bool = True,
+    pre_lm_cache_max_entries: int = 4096,
+    pre_lm_cache_max_bytes: int = 2 * 1024**3,
```
WilsonZheng0327: name change => pre_lm_cache_size_bytes following conventions

sglang_omni/models/fun_asr/sglang_model.py
```diff
@@ -376,6 +376,13 @@ def pad_input_ids(self, input_ids: List[int], mm_inputs: MultimodalInputs):
...
+    ) -> torch.Tensor:
```
WilsonZheng0327: wrapper doesn't seem to serve much, and introduces an unused parameter.
I believe the "uncached" is suggesting that this function is taken when it can't find it in cache, but _get_audio_feature_uncached naming suggests a cached counter part which doesn't exist. 
reverting back to get_audio_feature should work just fine

WilsonZheng0327 (commented):
Working on reproducing results (H200)

sglang_omni/models/fun_asr/encoder_service.py
AkazaAkane: I checked both `ReferenceEncodeService` and the merged MOSS-TD pre-LM implementation in #1045. `ReferenceEncodeService` currently performs `encode_one` inline, and its `encode_batch` hook is reserved but unused, so reusing it would not provide the bounded cross-request batching, dedicated CUDA stream, or multimodal-item attachment needed here without a broader refactor. MOSS-TD also uses a standalone pre-LM service. I kept the Fun-ASR service standalone for now and agree that the common worker/cache mechanics can be extracted during the planned refactor.

sglang_omni/models/fun_asr/encoder_service.py
AkazaAkane: Fixed in `f683d88d`. I added an idempotent `close()` path with a shutdown sentinel and lifecycle lock, reject new enqueues after shutdown, and join the worker with a five-second timeout. The encoder-service tests now close every created service, with a regression check confirming that the worker exits.

sglang_omni/models/fun_asr/encoder_service.py
AkazaAkane: Fixed in `f683d88d`. `_drain_batch()` now caps each batch at 8 items and uses a single 4 ms collection deadline for the whole batch.

sglang_omni/models/fun_asr/sglang_model.py
AkazaAkane: Fixed in `f683d88d`. Fun-ASR has no encoder graph/compile path requiring `forward_batch`, so I removed the parameter and folded the encoder implementation directly into `get_audio_feature()`.

sglang_omni/models/fun_asr/request_builders.py
AkazaAkane: Fixed in `f683d88d`. `encode_item()` now runs after token-budget, context-length, and sampling-parameter validation, immediately before `Req` construction. I also added checks ensuring invalid token budgets and context overruns never invoke the encoder.

sglang_omni/models/fun_asr/encoder_service.py
AkazaAkane: Fixed in `f683d88d` — renamed the class and all references to `FunASRPreLMEncoderService`.

sglang_omni/models/fun_asr/encoder_service.py
AkazaAkane: Fixed in `f683d88d`. `_expected_audio_tokens()` now requires the request builder's explicit `num_audio_tokens` metadata. I also tightened cache keying to require the full normalized `audio_fingerprint` instead of falling back to the shorter generic item hash.

sglang_omni/models/fun_asr/stages.py
AkazaAkane: Fixed in `f683d88d` — renamed the stage/config parameter and corresponding tests to `pre_lm_cache_size_bytes`.

sglang_omni/models/fun_asr/sglang_model.py
AkazaAkane: Fixed in `f683d88d`. I removed `_get_audio_feature_uncached()` and moved its implementation back into `get_audio_feature()`. The pre-LM service now calls `get_audio_feature(items)` directly.

WilsonZheng0327:
Reproduction attempt 7/19 on H200

| Metric | Current main | PR cold-start | PR warm |
|--------|--------------|---------------|---------|
| Throughput | 17.253 samples/s | 17.654 samples/s | 55.345 samples/s |
| Improvement | — | **+2.32% / 1.023×** | **+220.79% / 3.208×** |
| Wall time | 63.062 s | 61.629 s | 19.658 s |
| Mean latency | 1.847 s | 1.791 s | 0.576 s |
| p95 latency | 2.758 s | 2.086 s | 0.785 s |
| p99 latency | 3.366 s | 2.162 s | 0.939 s |
| Mean RTF | 0.39891 | 0.39375 | 0.12474 |
| Corpus WER | 0.01710 | 0.01710 | 0.01710 |
| Skipped | 0 | 0 | 0 |

for warm run, all hits cache, verified by the lack of statistical log since the worker is never woken up because everything hits cache

cold-start results are not ideal, likely because no room for thread parallelism in CPU, busy night.

tests/unit_test/fun_asr/test_encoder_service.py
```diff
@@ -0,0 +1,454 @@
...
+
+    assert item.precomputed_embeddings is None
+    assert len(service._cache) == 0
+
+
+def test_missing_token_count_raises() -> None:
+    service = _make_service()
+    item = SimpleNamespace(hash=1, feature=None, precomputed_embeddings=None)
+
+    with pytest.raises(RuntimeError, match="audio token count"):
```
0xjeffro: this test failed locally

0xjeffro: err msg should match with:
```python
def encode_item(self, item: Any) -> None:
        """Block until ``item.precomputed_embeddings`` holds the LM embedding.

        On success ``item.feature`` is cleared to release the CPU fbank/LFR
        tensor. Raises on encode failure; the request must not be admitted
        without the complete embedding.
        """
        expected_tokens = _expected_audio_tokens(item)
        if expected_tokens is None:
            raise RuntimeError(
                "Fun-ASR pre-LM encode requires the item's num_audio_tokens"
            )
        key = self._cache_key(item)
```

AkazaAkane: Updated the failing assertion in test_encoder_service.py (line 393)to match num_audio_tokens

sglang_omni/models/fun_asr/encoder_service.py
```diff
@@ -0,0 +1,401 @@
...
+            logger.warning(
+                f"Fun-ASR pre-LM cache entry {key} failed validation "
+                f"(shape={tuple(cached.shape)}, dtype={cached.dtype}); "
+                f"evicting and re-encoding"
+            )
+            self._cache.remove_if(lambda candidate: candidate == key)
+
+        leader = False
+        with self._lock:
+            future = self._inflight.get(key)
```
SandyLuXY, line 183: The cache miss and in-flight registration are not atomic: a follower can read a miss just before the leader caches its result, then enter this lock after the leader removes the in-flight future and launch a duplicate encode despite the populated cache. Re-check the cache while making this leader decision, with consistent lock ordering, and add an orchestrated stale-miss race test.

sglang_omni/models/fun_asr/stages.py
```diff
@@ -122,11 +129,28 @@ def create_sglang_fun_asr_executor(
...
+            model,
+            cache_namespace=build_cache_namespace(
+                model,
+                model_path=model_path,
+                feature_extractor=feature_extractor,
+                mm_attention_backend=getattr(server_args, "mm_attention_backend", None),
+            ),
+            cache_max_entries=pre_lm_cache_max_entries,
+            cache_max_bytes=pre_lm_cache_size_bytes,
+        )
```
SandyLuXY, line 148: Constructing this service starts a daemon encoder worker, but the returned scheduler never owns or closes it, so scheduler stop and factory failures can leave the worker, model reference, and CPU cache alive. Please wire explicit shutdown ownership into the scheduler lifecycle.

sglang_omni/models/fun_asr/encoder_service.py
```diff
@@ -0,0 +1,401 @@
...
+        self._stream = (
+            torch.cuda.Stream(device=self._device)
+            if self._device.type == "cuda"
+            else None
+        )
+        self._cache = StageOutputCache(
+            max_size=cache_max_entries,
+            max_bytes=cache_max_bytes,
+            cache_device="cpu",
+        )
```
SandyLuXY, line 101: nit: These public cache limits are not validated: max_entries=-1 crashes the first encode with an empty-pop KeyError, while max_bytes=-1 silently disables caching. Reject negative capacities at startup, preferably in StageOutputCache, while preserving zero as the disabled value.

SandyLuXY (changes requested):

sglang_omni/models/fun_asr/encoder_service.py
AkazaAkane: Made cache lookup and single-flight registration race-safe, with an orchestrated regression test.

sglang_omni/models/fun_asr/encoder_service.py
AkazaAkane: Rejected negative StageOutputCache capacities while preserving zero as disabled.

sglang_omni/models/fun_asr/stages.py
AkazaAkane: Added scheduler-owned encoder shutdown and factory-failure cleanup.

sglang_omni/models/fun_asr/encoder_service.py
```diff
@@ -0,0 +1,412 @@
...
+                with self._lock:
+                    self._hits += 1
+                self._attach(item, cached)
+                return
+            logger.warning(
+                f"Fun-ASR pre-LM cache entry {key} failed validation "
+                f"(shape={tuple(cached.shape)}, dtype={cached.dtype}); "
+                f"evicting and re-encoding"
+            )
+            self._cache.remove_if(lambda candidate: candidate == key)
```
SandyLuXY: After reading an invalid cached embedding, `encode_item()` removes the key with a key-only `remove_if()`. If another encoder completion replaces that key with a valid embedding between the stale read and the removal, the stale reader deletes the valid replacement and launches a redundant encode.
Fix: Add an atomic compare-and-remove cache operation and remove the entry only if the cached object is still the exact invalid object that was observed.

tests/unit_test/fun_asr/test_encoder_service.py
SandyLuXY, line 1: Add to tests/README.md

sglang_omni/models/fun_asr/encoder_service.py
SandyLuXY: Fixed in `85c587e`

tests/unit_test/fun_asr/test_encoder_service.py
SandyLuXY: Fixed in `85c587e`

SandyLuXY (approved):

PR #1096, opened 2026-07-19
[TTS Refactor] Reduce reference encode hook boilerplate

luojiaxuan:
## Summary

- Add keyed and tensor hook defaults for shared key construction, revalidation, and tensor ownership/dtype policy.
- Migrate FishAudio S2-Pro, Higgs, MOSS-TTS Local, and Qwen3-TTS.
- Preserve MOSS one-time duration validation and Fish immutable-input revalidation fast paths.

## Result

Non-test, non-documentation Python: +111/-120, net -9.

The four model hooks shrink by 65 lines while the shared layer adds 56. Qwen structured prompt serialization remains model-specific.

## Verification

- 27 focused H100 tests passed.
- 13 shared scheduling tests passed locally.
- Pre-commit passed.


SandyLuXY (approved):
lgtm

PR #1101, opened 2026-07-19
[Qwen3-Omni Perf] exact-shape CUDA Graph replay for Code2Wav

edwingao28:
## Motivation

In Qwen3-Omni speech serving the pipeline is bottlenecked by the Talker and Code2Wav stages, not the Thinker. Each Code2Wav window pays per-call kernel-launch and Python overhead that CUDA Graph replay can remove. This PR enables that optimization by default for serving reachable c2w windows, while retaining eager execution for unsupported shapes and capture time incompatibilities.

## Modifications

- New `sglang_omni/models/qwen3_omni/components/code2wav_cuda_graph.py`: exact-shape graph runner. Capture is atomic (all graphs publish or none), memory-budgeted, and key-validated; replay failures abort the request instead of silently re-running eagerly; a PID guard fails closed across process forks.
- `code2wav_scheduler.py`: threshold-triggered windows may replay a graph; final stream tails and unsupported shapes run eagerly. Graph output is copied out while the scheduler state lock is held. Capture is restricted to the four serving-reachable shapes `B1/T{10,20,30,35}`; anything else is a key miss and falls back to eager.
- Enabled by default. The default c2w stage supplies a typed `runtime.resources.total_gpu_memory_fraction` of `0.02`, custom configs with CUDA Graph enabled must provide this budget or startup fails before model loading. Disable it with `runtime_overrides.code2wav.enable_cuda_graph: false`
- One change not gated by the flag: the vocoder module is now constructed in eval mode (`model.eval()`). It previously served in train mode; for this vocoder the two are functionally identical (no dropout / batch-norm), and eval mode is required for deterministic capture.

## Accuracy Tests

Graph replay output is bit-exact against eager: on H100, candidate (graph) vs control (eager) with fixed seed at concurrency 1 produced 16/16 fully bit-identical results (text, PCM bytes, chunk layout). Profiler evidence from the same server: 430 windows replayed via CUDA graph across all four keys, 8 final tails correctly fell back to eager as ineligible.

## Benchmarking and Profiling

The H100 A/B below was measured before the default-on follow-up. Both arms used the existing colocated H100 BF16 config; the candidate added only runtime_overrides.code2wav.enable_cuda_graph: true, while the control used the then-default eager path. Each arm used a fresh server, alternating arm order, 8 warmup + 64 measured requests per arm at concurrency 8, temperature 0, seed 777.

| pair | request/s | audio bytes/s | latency p50 | TTFA p50 |
|------|-----------|---------------|-------------|----------|
| 2    | +7.57%    | +7.07%        | −6.03%      | −0.03%   |
| 4    | +7.95%    | +7.48%        | −7.66%      | +1.72%   |
| 5    | +7.80%    | +6.58%        | −6.01%      | −1.62%   |
| mean | **+7.77%** | **+7.04%**   | **−6.57%**  | +0.02%   |

All three clean exclusive-window pairs are positive with a tight spread. Two additional pairs were attempted repeatedly but invalidated by shared-host co-tenancy during the runs (GPU seizure by other jobs, and batch-tie text flips at c8 — never a feature failure); raw artifacts of every attempt are retained. TTFA is flat because the win lands in decode latency rather than first audio.

audio bytes/s is reported alongside request/s because per-request audio length varies slightly under concurrent batching; it normalizes throughput by work actually produced.

**Updated benchmark result**
Code2Wav CUDA Graph is enabled by default with a typed 2% stage budget, missing budgets now fail before model load, and the CUDA Graph test coverage/invocation is documented. The related non-GPU suite passed 69/69 tests in the H200 container.

The follow-up does not change graph capture, replay, or the scheduler hot path, so the controlled H200 sweep below remains the result measured on ancestor `b4159d2a` (no rerun):

| concurrency | request/s control → graph | paired Δ | audio xRT control → graph | paired Δ | latency p50 Δ | TTFA p50 Δ |
| 8 | 0.460 → 0.503 | **+9.35%** | 28.57 → 31.24 | **+9.35%** | **−9.03%** | −2.53% |
| 16 | 0.575 → 0.635 | **+10.55%** | 35.72 → 39.48 | **+10.55%** | **−10.25%** | −6.27% |
| 32 | 0.669 → 0.752 | **+12.37%** | 41.57 → 46.71 | **+12.41%** | **−10.74%** | −2.07% |

1536/1536 measured requests and 224/224 warmups succeeded; all six graph arms published all four graphs with no replay failure or disable. Matched total audio work stayed within 0.076%.

## Checklist

- [x] Add unit tests.
- [x] Update documentation.


edwingao28:
H200 saturation sweep on `b4159d2a` — 1×H200, colocated BF16, fresh server per arm, temperature 0, seed 777, Talker capped at 800 tokens, two alternating A/B pairs per level, 128 measured requests per arm, and warmup equal to concurrency.

| concurrency | request/s control → graph | paired Δ | audio xRT control → graph | paired Δ | latency p50 Δ | TTFA p50 Δ |
| 8 | 0.460 → 0.503 | **+9.35%** | 28.57 → 31.24 | **+9.35%** | **−9.03%** | −2.53% |
| 16 | 0.575 → 0.635 | **+10.55%** | 35.72 → 39.48 | **+10.55%** | **−10.25%** | −6.27% |
| 32 | 0.669 → 0.752 | **+12.37%** | 41.57 → 46.71 | **+12.41%** | **−10.74%** | −2.07% |

- 1536/1536 measured requests and 224/224 warmups succeeded; all six candidate arms published 4 graphs with no replay failure or disable.
- All six matched pairs kept total audio work within 0.076%; deltas are means of the two paired percent changes.
- One uncapped c16 attempt was excluded before aggregation because Talker output work differed by 57.5% between arms.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -276,14 +379,38 @@ def create_code2wav_scheduler(
...
     """Factory: returns Code2WavScheduler."""
     if gpu_id is not None:
         device = f"cuda:{gpu_id}"
+    concrete_device = torch.device(device)
+    if concrete_device.type == "cuda" and concrete_device.index is None:
+        concrete_device = torch.device("cuda", torch.cuda.current_device())
+    device = str(concrete_device)
     model = load_code2wav_model(model_path, device=device, dtype=dtype)
+    cuda_graph_runner = None
+    if enable_cuda_graph:
```
SandyLuXY, line 421: An explicit enable with no typed stage budget silently starts a disabled runner and serves eagerly: the default speech config resolves `enable_cuda_graph=True` with `total_gpu_memory_fraction=None`, and `build()` only records `invalid_total_gpu_memory_fraction`. Please reject this missing required field at config/factory startup; capture-time incompatibilities can still use the intended eager fallback policy.

tests/unit_test/qwen3_omni/test_code2wav_cuda_graph.py
SandyLuXY, line 1: Please update `tests/README.md` for this test addition: the Qwen3-Omni tree does not list `test_code2wav_cuda_graph.py`, and the coverage summary still mentions only Code2Wav streaming/cleanup. Document the graph lifecycle, memory/failure coverage, and the real-GPU marker/invocation so the suite remains discoverable.

SandyLuXY (changes requested):
Nice work! Please turn on the cuda graph by default as it improves the performance.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
edwingao28: Fixed in `cbda9110`. The default Code2Wav stage now enables CUDA Graph with a typed `runtime.resources.total_gpu_memory_fraction=0.02` budget. A custom config that enables the graph without a typed budget now raises before device resolution or model loading; capture-time incompatibilities still use the existing eager fallback policy.

tests/unit_test/qwen3_omni/test_code2wav_cuda_graph.py
edwingao28: Fixed in `cbda9110`. `tests/README.md` now lists `test_code2wav_cuda_graph.py`, documents graph lifecycle, atomic rollback, memory/failure and replay coverage, and includes the real-GPU invocation: `pytest tests/unit_test/qwen3_omni/test_code2wav_cuda_graph.py -m gpu -q`.

edwingao28:
Review follow-up is in `cbda9110`: Code2Wav CUDA Graph is enabled by default with a typed 2% stage budget, missing budgets now fail before model load, and the CUDA Graph test coverage/invocation is documented. The related non-GPU suite passed 69/69 tests in the H200 container.

The follow-up does not change graph capture, replay, or the scheduler hot path, so the controlled H200 sweep below remains the result measured on ancestor `b4159d2a` (no rerun):

| concurrency | request/s control → graph | paired Δ | audio xRT control → graph | paired Δ | latency p50 Δ | TTFA p50 Δ |
| 8 | 0.460 → 0.503 | **+9.35%** | 28.57 → 31.24 | **+9.35%** | **−9.03%** | −2.53% |
| 16 | 0.575 → 0.635 | **+10.55%** | 35.72 → 39.48 | **+10.55%** | **−10.25%** | −6.27% |
| 32 | 0.669 → 0.752 | **+12.37%** | 41.57 → 46.71 | **+12.41%** | **−10.74%** | −2.07% |

1536/1536 measured requests and 224/224 warmups succeeded; all six graph arms published all four graphs with no replay failure or disable. Matched total audio work stayed within 0.076%.

sglang_omni/models/qwen3_omni/components/code2wav_cuda_graph.py
```diff
@@ -0,0 +1,536 @@
...
+
+@dataclass(frozen=True, slots=True)
+class GraphKey:
+    """One exact Code2Wav input shape, excluding fixed quantizer count."""
+
+    batch_size: int
+    frames: int
+
+
+CODE2WAV_GRAPH_KEYS = tuple(
```
yxs: `CODE2WAV_GRAPH_KEYS` is hardcoded to B=1, T is {10,20,30,35}, but those shapes are purely a function of `stream_chunk_size` (10) and `left_context_size` (25).
Both are factory kwargs reachable through the same runtime_overrides.code2wav path as enable_cuda_graph, and the runner never receives them.
So if someone enables graphs and also sets e.g. stream_chunk_size: 20, all 4 graphs are still captured (wasting VRAM + startup time), but the real serving windows become {20, 40, 45}

edwingao28: Thanks for the catcha and fixed. Graph shapes now derived from stream_chunk_size and left_context_size with test covering overridden configs

SandyLuXY (approved):

PR #1104, opened 2026-07-20
[TTS] Land Audar-TTS V1 Turbo on main

luojiaxuan:
## Summary

- Land the Audar-TTS integration from #1090 on main.
- #1090 was merged into its stacked base branch after #1096 had already merged into main.
- Address the remaining #1090 review feedback in the public speech validation path.
- Require exactly one transcribed Audar reference before backend dispatch.
- Scope Arabic language acceptance to the Audar pipeline so Qwen3-TTS still fails fast.

## Validation

- 138 focused unit tests passed on H100
- pre-commit passed on all changed files
- Diff against main contains the reviewed Audar integration and these validation fixes only

### DIFF
--- a/examples/configs/audar_tts_turbo.yaml
+++ b/examples/configs/audar_tts_turbo.yaml
@@ -1,0 +1,2 @@
+config_cls: AudarTTSPipelineConfig
+model_path: audarai/Audar-TTS-V1-Turbo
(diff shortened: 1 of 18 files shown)


zhaochenyang20 (approved):

SandyLuXY (approved):
LGTM

PR #1109, opened 2026-07-20
[Fun-ASR] Add pad+mask batched audio encoding in get_audio_feature

guozhihao-224:
#924 ：F-PR5 encode kernel / prerequisite for F-PR5; throughput gated on F-PR3 (#1095).
## Summary
On top of F-PR3 (pre-LM encoder service + cache + single-flight), this adds a pad+mask batched encode path for Fun-ASR-Nano (F-PR5 kernel), plus SDPA / fused QKV / FSMN reuse of v.

With the default max_batch_wait_ms=4 and request_build_max_workers=8, observed avg B ≈ 2.7, so end-to-end gains vs PR3 alone are negligible.
After tuning batching/workers to sustain avg B ≥ 8, cold throughput vs PR3 alone improves by +56% at c=16 and +84% at c=32, with unchanged WER.

Also wires pre_lm_max_batch_size / pre_lm_max_batch_wait_ms through Fun-ASR stages/config (defaults remain 8 / 4) so high-concurrency cold deployments can increase coalescing explicitly.

## Motivation
- F-PR3 already moves encode before LM admission and can coalesce cross-request work, but get_audio_feature in PR3 still runs per-item forwards.

- True batched GEMM is F-PR5; this change adds a SenseVoice-style pad+mask kernel so the pre-LM service can do one encoder/adaptor forward when B>1.

- PR3’s default 8/4ms comes from review alignment with MOSS conventions (latency-first, small capped batches), not from pad+mask throughput tuning. request_build_max_workers=8 also caps concurrent encode_item enqueue, so avg B stays low.

## Modifications
- sglang_omni/models/fun_asr/sglang_model.py
   - Pad+mask: SenseVoice-style [B,1,T] through SANM / FSMN / Adaptor; get_audio_feature pads to T_max, runs one forward, then slices by fun_asr_low_frame_rate_length
   - Skip mask on the common B=1 unpadded path for parity with the old forward
   - SDPA + runtime fused QKV (keep separate q/k/v Linears for HF checkpoint names)
   - SANM attention returns (attn_out, v) so FSMN reuses v (no duplicate v_proj)
- sglang_omni/models/fun_asr/stages.py / config.py
   - New knobs: pre_lm_max_batch_size (default 8), pre_lm_max_batch_wait_ms (default 4)
- tests/unit_test/fun_asr/test_model.py: mask / batched≡serial / fused QKV / invalid-input coverage

## Benchmark
Setup: 1× H100 80GB, bf16, SeedTTS EN 1088 clips, --repeats 1, no warmup unless marked warm
### A. Default PR3 knobs (batch=8, wait=4ms, workers=8)
Observed avg B ≈ 2.6–2.8

Scenario | PR3 alone | PR3 + pad/SDPA | Δ thrpt
-- | -- | -- | --
Cold c=16 | 47.56/s · lat 0.335s · WER 0.0171 | 48.12/s · lat 0.330s · WER 0.0172 | +1.2%
Cold c=32 | 48.86/s · lat 0.647s · WER 0.0171 | 49.79/s · lat 0.636s · WER 0.0171 | +1.9%
Warm c=16 | 137.65/s · lat 0.116s · WER 0.0171 | 135.84/s · lat 0.117s · WER 0.0171 | ≈0

Takeaway: Batches are too small for pad+mask to matter; warm is cache-dominated.

### B. High-concurrency cold knobs (batch=16, wait=50ms, workers=32, pending=64)
Observed avg B ≈ 8–10

Scenario | avg B | PR3 alone | PR3 + pad/SDPA | Δ thrpt
-- | -- | -- | -- | --
Cold c=16 | ~7.5–8.9 | 47.88/s · lat 0.332s · WER 0.0171 | 74.94/s · lat 0.213s · WER 0.0171 | +56%
Cold c=32 | ~9.8 | 51.71/s · lat 0.612s · WER 0.0171 | 95.30/s · lat 0.332s · WER 0.0172 | +84%

Encoder wall time (c=32, ~500-item log checkpoint):

type | PR3 alone | PR3 + pad/SDPA
-- | -- | --
encoder_time_s | encoder_time_s | ~13.4s | ~5.8s (~2.3×)

### C. Ablation: larger wait without larger workers
batch=16, wait=50ms, workers=8 → avg B only ~4, with heavy wait tax:

Cold c=16 | thrpt | lat
-- | -- | --
PR3 alone | 21.86/s | 0.729s
PR3 + pad/SDPA | 24.98/s | 0.637s

Takeaway: Throughput drops vs defaults; request_build_max_workers must rise for avg B to grow.

### D. Encoder microbench (same SDPA kernel; pad vs serial)

Case | serial | batched | speedup
-- | -- | -- | --
B=1, T=80 | 20.4ms | 25.8ms | 0.79×
B=4 equal T=80 | 80.2ms | 25.9ms | 3.1×
B=8 var (Seed-like) | 163.0ms | 27.1ms | 6.0×

## Conclusions
- F-PR5 (pad+mask) belongs on top of F-PR3; gains show up on cold / cache-miss paths.
- Under default 4ms/8 + workers=8, e2e looks flat because coalescing is insufficient, not because the kernel is useless.
- Suggested high-concurrency cold config: pre_lm_max_batch_size=16, pre_lm_max_batch_wait_ms=50, request_build_max_workers=32 (lower wait if latency budget is tight).
- Do not make 50ms the global default: c=1 / light load pays the full collection window and hurts PR3’s low-latency profile; keep 8/4 as the latency-first default.
- WER stays aligned with PR3 alone / cookbook (~0.0171).

## Related
- Roadmap: [#924](https://github.com/sgl-project/sglang-omni/issues/924) (F-PR3 / F-PR5)
- Base: [#1095](https://github.com/sgl-project/sglang-omni/pull/1095) (8/4ms suggested by [WilsonZheng0327](https://github.com/sgl-project/sglang-omni/pull/1095#discussion_r3611779719) to follow MOSS)

### DIFF
--- a/sglang_omni/models/fun_asr/sglang_model.py
+++ b/sglang_omni/models/fun_asr/sglang_model.py
@@ -8,6 +8,7 @@ from typing import Any, Iterable, List, Optional, Tuple
 
 import torch
 import torch.nn as nn
+import torch.nn.functional as F
 from sglang.srt.layers.quantization.base_config import QuantizationConfig
 from sglang.srt.managers.mm_utils import (
     MultiModalityDataPaddingPatternMultimodalTokens,
@@ -28,6 +29,42 @@ from .configuration_fun_asr import FunAsrNanoConfig
 from .tool_funcs.audio_lengths import fun_asr_low_frame_rate_length
 
 logger = logging.getLogger(__name__)
+
+
+def _sanm_mask_from_lengths(
+    lengths: torch.Tensor, max_len: int, *, dtype: torch.dtype, device: torch.device
+) -> torch.Tensor:
+    # note (guozhihao): SenseVoice pad mask [B, 1, T], 1=valid.
+    idx = torch.arange(max_len, device=device).unsqueeze(0)
+    return (idx < lengths.unsqueeze(1)).to(dtype=dtype).unsqueeze(1)
+
+
+def _apply_time_mask(x: torch.Tensor, mask: Optional[torch.Tensor]) -> torch.Tensor:
+    if mask is None:
+        return x
+    return x * mask.transpose(1, 2)
+
+
+def _additive_key_pad_mask(mask: torch.Tensor, dtype: torch.dtype) -> torch.Tensor:
+    # note (guozhihao): SenseVoice [B, 1, T] (1=valid) -> SDPA additive [B, 1, 1, T].
+    return torch.zeros(
+        mask.shape[0], 1, 1, mask.shape[-1], device=mask.device, dtype=dtype
+    ).masked_fill(mask.unsqueeze(1).eq(0), torch.finfo(dtype).min)
+
+
+def _fused_qkv_project(
+    x: torch.Tensor,
+    q_proj: nn.Linear,
+    k_proj: nn.Linear,
+    v_proj: nn.Linear,
+) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    # note (guozhihao): keep separate q/k/v Linears for HF checkpoint names;
+    # fuse into one GEMM at runtime.
+    weight = torch.cat([q_proj.weight, k_proj.weight, v_proj.weight], dim=0)
+    bias = None
+    if q_proj.bias is not None:
+        bias = torch.cat([q_proj.bias, k_proj.bias, v_proj.bias], dim=0)
+    return F.linear(x, weight, bias).chunk(3, dim=-1)
 
 
 class SinusoidalPositionEncoder(nn.Module):
@@ -62,24 +99,30 @@ class MultiHeadedAttentionSANM(nn.Module):
         self.k_proj = nn.Linear(in_feat, n_feat)
         self.v_proj = nn.Linear(in_feat, n_feat)
         self.out_proj = nn.Linear(n_feat, n_feat)
-        self.dropout = nn.Dropout(p=dropout_rate)
+        self.attn_dropout_p = float(dropout_rate)
 
-    def forward(self, x: torch.Tensor) -> torch.Tensor:
+    def forward(
+        self, x: torch.Tensor, mask: Optional[torch.Tensor] = None
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        # Returns (attn_out, v) so FSMN can reuse the same value projection.
         b, t, _ = x.size()
-        q = self.q_proj(x)
-        k = self.k_proj(x)
-        v = self.v_proj(x)
-        q_h = q.view(b, t, self.h, self.d_k).transpose(1, 2)  # (b, h, t, dk)
+        q, k, v = _fused_qkv_project(x, self.q_proj, self.k_proj, self.v_proj)
+        q_h = q.view(b, t, self.h, self.d_k).transpose(1, 2)
         k_h = k.view(b, t, self.h, self.d_k).transpose(1, 2)
         v_h = v.view(b, t, self.h, self.d_k).transpose(1, 2)
 
-        q_h = q_h * (self.d_k**-0.5)
-        scores = torch.matmul(q_h, k_h.transpose(-2, -1))  # (b, h, t, t)
-        attn = torch.softmax(scores, dim=-1)
-        p_attn = self.dropout(attn)
-        x = torch.matmul(p_attn, v_h)  # (b, h, t, dk)
-        x = x.transpose(1, 2).contiguous().view(b, -1, self.h * self.d_k)
-        return self.out_proj(x)
+        attn_mask = None if mask is None else _additive_key_pad_mask(mask, q.dtype)
+        dropout_p = self.attn_dropout_p if self.training else 0.0
+        out = F.scaled_dot_product_attention(
+            q_h,
+            k_h,
+            v_h,
+            attn_mask=attn_mask,
+            dropout_p=dropout_p,
+            is_causal=False,
+        )
+        out = out.transpose(1, 2).contiguous().view(b, t, self.h * self.d_k)
+        return self.out_proj(out), v
 
 
 class FunAsrNanoFSMN(nn.Module):
@@ -100,10 +143,16 @@ class FunAsrNanoFSMN(nn.Module):
         self.pad = nn.ConstantPad1d((left_padding, right_padding), 0.0)
         self.dropout = nn.Dropout(dropout_rate)
 
-    def forward(self, value_states: torch.Tensor) -> torch.Tensor:
+    def forward(
+        self, value_states: torch.Tensor, mask: Optional[torch.Tensor] = None
+    ) -> torch.Tensor:
+        # note (guozhihao): zero pad frames before/after the depthwise conv so
+        # kernel_size windows cannot leak padded values into valid frames.
+        value_states = _apply_time_mask(value_states, mask)
         hidden_states = self.conv(self.pad(value_states.transpose(1, 2)))
         hidden_states = hidden_states.transpose(1, 2) + value_states
-        return self.dropout(hidden_states)
+        hidden_states = self.dropout(hidden_states)
+        return _apply_time_mask(hidden_states, mask)
 
 
 class EncoderLayerSANM(nn.Module):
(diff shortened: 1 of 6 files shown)


sglang_omni/models/fun_asr/sglang_model.py
```diff
@@ -376,28 +449,62 @@ def pad_input_ids(self, input_ids: List[int], mm_inputs: MultimodalInputs):
...
-                feature = feature[:, :, :valid]
-            # [1, 560, T] → [1, T, 560] (encoder expects [B, T, D]).
-            xs = feature.permute(0, 2, 1).contiguous()
-            enc_out = self.audio_tower(xs)  # [1, T, 512]
-            adp_out = self.multi_modal_projector(enc_out)  # [1, T, 1024]
-            t_lfr = adp_out.shape[1]
-            num_tokens = int(fun_asr_low_frame_rate_length(t_lfr))
-            num_tokens = max(num_tokens, 1)
-            embeddings.append(adp_out[0, :num_tokens, :])  # [num_tokens, 1024]
+                valid = int(mask.to(device=device).sum().item())
```
SandyLuXY: This moves each CPU mask to CUDA only to immediately read its sum back with `.item()`, introducing one H2D transfer and a device sync per item; the loop also transfers each feature before copying it into a second GPU batch. Compute lengths and padding on the host and transfer the completed batch once so the new batched encoder path stays asynchronous.

sglang_omni/models/fun_asr/stages.py
```diff
@@ -113,6 +113,8 @@ def create_sglang_fun_asr_executor(
     enable_pre_lm_encoder: bool = True,
     pre_lm_cache_max_entries: int = 4096,
     pre_lm_cache_size_bytes: int = 2 * 1024**3,
+    pre_lm_max_batch_size: int = 8,
```
SandyLuXY, line 137: These are now public stage-config knobs, but the service silently clamps `max_batch_size <= 0` to 1 and negative wait values to 0, so invalid deployment config starts with a different policy instead of failing fast. Please validate both values at the factory boundary.

SandyLuXY (changes requested):

SandyLuXY:
@guozhihao-224 Could you share more context why we chose setup A (`batch=8, wait=4ms`)? Seems its throughput improvement is limited compared with setup B. Thanks!

guozhihao-224:
> @guozhihao-224 Could you share more context why we chose setup A (`batch=8, wait=4ms`)? Seems its throughput improvement is limited compared with setup B. Thanks!

@SandyLuXY Thanks for the question — here’s the updated data and rationale for keeping Setup A as the default.

### Why default stays Setup A (batch=8, wait=4ms, workers=8)
Setup A is the latency-first default inherited from F-PR3 / MOSS-style coalescing, not a throughput-tuned setting for pad+mask.

We re-ran ours-only A vs B on SeedTTS EN (1088 clips), cold, 3 rounds, H100:


conc | A 8/4/w8 (thrpt / lat) | B 16/50/w32 (thrpt / lat) | Δ thrpt | Δ lat
-- | -- | -- | -- | --
1 | 13.0/s · 0.080s | 9.8/s · 0.105s | −25% | +26 ms
16 | 50.5/s · 0.316s | 72.9/s · 0.218s | +44% | −98 ms
32 | 50.0/s · 0.633s | 98.5/s · 0.321s | +97% | −312 ms

WER ≈ 0.017 in all runs.

So under light load (c=1), Setup B is strictly worse because the 50 ms collection window is mostly wait tax (avg B stays ~1). That’s why we do not make B the global default.

Under high-concurrency cold (c=16/32), B wins clearly — which is exactly when pad+mask helps, once coalescing can sustain larger batches.

### Setup B internal sweep (c=32, 3-round mean)

config (batch / wait_ms / workers) | thrpt | lat | vs B 16/50/32
-- | -- | -- | --
16 / 50 / 32 (B) | 98.5/s | 0.321s | baseline
16 / 20 / 32 | 98.5/s | 0.321s | ≈0%
16 / 10 / 32 | 98.1/s | 0.323s | −0.5%
32 / 50 / 32 | 96.2/s | 0.329s | −2.4%
16 / 50 / 16 | 93.4/s | 0.338s | −5.2%

Takeaways:

- wait=10–20ms is already close to 50ms at c=32
- batch=32 does not help further
- dropping workers to 16 costs ~5% (need enough enqueue parallelism to grow avg B)

Suggested high-concurrency cold recipe:
pre_lm_max_batch_size=16, pre_lm_max_batch_wait_ms=20–50, request_build_max_workers≥32
(use the lower wait if latency budget is tight).

### vs current main (pad+mask effectiveness)
Same Suite B knobs, cold c=32, single aligned run on free GPU0/1:


type |thrpt | lat | WER
-- | -- | -- | --
origin/main | 53.1/s | 0.596s | 0.0171
this PR | 95.5/s | 0.332s | 0.0171
Δ | +80% |  - | -   

### Summary
- Default A: preserve PR3 low-latency behavior (c=1 must not regress)
- Knobs for B: opt into throughput-first cold path when concurrency is high
- Pad+mask is effective once avg B is large enough; flat Setup-A e2e is a coalescing issue, not a kernel issue

sglang_omni/models/fun_asr/sglang_model.py
```diff
@@ -376,28 +449,62 @@ def pad_input_ids(self, input_ids: List[int], mm_inputs: MultimodalInputs):
...
+        # stays bit-identical to the pre-batching encoder forward.
+        if batch_size == 1 and lengths[0] == t_max:
+            sanm_mask: Optional[torch.Tensor] = None
+        else:
+            ilens = torch.tensor(lengths, device=device, dtype=torch.long)
+            sanm_mask = _sanm_mask_from_lengths(
+                ilens, t_max, dtype=xs.dtype, device=device
+            )
 
+        enc_out = self.audio_tower(xs, sanm_mask)
```
SandyLuXY, line 505: When encoder compile is enabled, startup warms both forwards with `mask=None`, but every multi-item batch reaches this line with a tensor mask. Dynamo specializes that optional-argument branch, so the first real batch recompiles both modules; please warm the reachable mask signature and run benchmark.

sglang_omni/models/fun_asr/sglang_model.py
```diff
@@ -376,28 +449,62 @@ def pad_input_ids(self, input_ids: List[int], mm_inputs: MultimodalInputs):
...
+        batch_size = len(feats)
+        feat_dim = feats[0].shape[1]
+        t_max = max(lengths)
+        batched = feats[0].new_zeros(batch_size, feat_dim, t_max)
+        for i, (feat, length) in enumerate(zip(feats, lengths)):
+            batched[i, :, :length] = feat[0, :, :length]
+
+        xs = batched.permute(0, 2, 1).to(device=device, dtype=dtype, non_blocking=True)
+        # note (guozhihao): skip masking for the common B=1 unpadded path so it
+        # stays bit-identical to the pre-batching encoder forward.
```
luojiaxuan: Skipping the mask avoids the masked SDPA path, but this is not bit-identical to the old forward because this PR also switches attention to SDPA and fused QKV; please change the comment to say numerically equivalent.

luojiaxuan (commented):
One wording fix.

sglang_omni/models/fun_asr/sglang_model.py
guozhihao-224: @SandyLuXY Fix: compile warmup now runs three serving signatures: B1/None, B1/mask, B2/mask, with the same contiguous [B, T, D] layout as get_audio_feature (permute alone left stride[1]=1 and also forced a recompile).


case | old warmup | new warmup
-- | -- | --
B1 unpadded (mask=None) | ~0.2s | ~0.01s
B1 pad + mask | ~0.2s | ~0.01s
B2 pad + mask (first multi-item) | ~137s | ~0.01s
B4 pad + mask | ~0.1s (after B2 recompile) | ~0.01s (no recompile tax)

Startup warmup cost is similar (~67–69s). With enable_encoder_torch_compile=true, the first live encode is now ~0.09s instead of ~56s.

sglang_omni/models/fun_asr/sglang_model.py
guozhihao-224: Updated the comment to numerically equivalent to the unmasked encoder path.

LauraGPT (approved):
Reviewed the latest commit. The CPU-side batch assembly now avoids the per-item H2D/sync path, invalid batching knobs fail at the executor boundary with focused tests, and compile warmup covers the serving signatures (B1/None, B1/mask, B2/mask). Fun-ASR SeedTTS, unit tests, lint, and static analysis are green. The remaining failed jobs exercise untouched Qwen3-Omni/TTS paths; the latest main run also fails the same MMSU Talker stage.

luojiaxuan (approved):
Reviewed 709c706a. Compile warmup now covers the serving signatures, the contiguous input layout matches get_audio_feature, and relevant CI is green.

SandyLuXY (approved):

PR #1124, opened 2026-07-21
[MPS-DP] Same-GPU DP weight sharing over CUDA IPC + autodp auto-sizing launcher

yl3469:
> ## Current support scope (default OFF)
>
> Weight sharing is opt-in (`WEIGHT_SHARE=1`, default off; `autodp.sh` defaults it on because its sizing assumes sharing) and gated to **validated architectures with tp=pp=1**. Support is binary: a model is supported only if, at the current PR revision, its documented `examples/mps_dp` command passed a one-H100 end-to-end smoke — shared N=2 boot under MPS, health, attach verification, concurrent-request correctness, clean teardown and restart. **Supported (7): Higgs, MOSS local, MOSS delay, MOSS Transcribe-Diarize, Qwen3-ASR, Whisper, FunASR Nano**, each with a shipped config under `examples/mps_dp/configs/` (commands, evidence, and the per-model table live in `docs/basic_usage/mps_dp.md`). Everything else is **not supported** and is rejected in preflight, before the MPS daemon, state dir, or any replica exists; completed architecture audits for Ming TTS, Voxtral, Fish S2-Pro, LLaDA2, Qwen3-TTS, and both Qwen3-Omni engines are recorded in `AUDIT_ONLY_WEIGHT_SHARE_POLICIES` with support still in progress (LLaDA2 and Qwen3-Omni additionally cannot be driven by this launcher: no single generation SGLang stage). Per-step decode staging tensors are classified replica-private (MOSS local/delay) and the classification joins the manifest hash, fail-closed both ways. The **Qwen3-TTS and Ming-Omni-TTS results below are historical / pre-gate**; the Ming **~1.5-2.0x** figure is a projection, not a measurement. `autodp.sh` reports a **maximum *estimated* DP** (boot-validated, sustained-load-tested for Higgs only). Leader/follower is a whole-group lifecycle: the leader must outlive followers, no online weight updates while sharing, and followers require an explicit `--max-total-tokens`.

## Motivation

Same-GPU data parallelism under CUDA MPS ([sgl-project#1037](https://github.com/sgl-project/sglang-omni/pull/1037), [sgl-project#1042](https://github.com/sgl-project/sglang-omni/pull/1042), [sgl-project#1076](https://github.com/sgl-project/sglang-omni/pull/1076)) multiplies throughput on one GPU, but today each replica loads its **own copy** of the model weights — at DP4 with Higgs TTS 3-4B that is 4 × 7.55 GiB of identical AR-backbone weights resident on one card. This PR lets co-located DP replicas share **a single copy of the weights over CUDA IPC**: replica 0 (the leader) loads the checkpoint and exports IPC handles; replicas 1..N-1 attach zero-copy instead of loading.

Measured on one H200 (details below): **7.55 GiB × (N−1) VRAM saved** — 22.8 GiB at DP4 — at **statistical parity** in throughput, TTFC, and latency at every N. The reclaimed VRAM funds more KV cache or more replicas.

The PR also adds `autodp.sh`, a launcher that probes the model's real memory footprint and sizes a **maximum estimated DP** (boot-validated, not workload-proven) for the GPU + model + KV-cap combination, deriving `--mem-fraction-static` from a two-constraint model instead of trial-and-error.

## Performance
Off-vs-on measurements for the supported models (80 GB H100; support standard and evidence in `docs/basic_usage/mps_dp.md`):

| Model | Validated DP, off | Validated DP, on | VRAM, off | VRAM, on | Saved per follower | Throughput (aggregate), off vs on |
|---|---|---|---:|---:|---:|---|
| MOSS TTS delay | **DP1** (unshared DP2: replica 1 cannot fit its own weight copy) | **DP2** | n/a | 42.4 GB | 17.05 GiB | single 5.7 to shared DP2 agg 7.5 qps (+32%, per replica 3.8, preliminary); aligned-history byte-identical |
| Higgs TTS 3-4B | DP3 (unshared DP4 needs 98 GB) | **DP4** (74.9 GB idle, 78.4 load) | 73.7 GB at DP3 | 58.1 GB at DP3 | 7.55 GiB | parity at every N; shared DP4 +10% to +40% over shared DP3, round-matched |
| MOSS TTS local | DP3 at the card edge, graphs partly eager | DP3, 13.3 GB headroom, full graphs | 78.0 GB | 61.8 GB | 8.44 GiB | parity: DP2 17.9 vs 18.2, DP3 23.4 vs 23.4 (per replica 7.8) |
| Whisper large-v3-turbo | DP3 (40k) | **DP6** (19.7 GB) | 14.1 GB at DP3 | 10.7 GB at DP3 | 1.51 GiB | parity at DP3 (67.8 vs 68.0); shared DP6 agg 95.3 cold qps, +40% over DP3 |
| MOSS Transcribe-Diarize | DP3 (40k) | DP3 | 23.9 GB | 20.2 GB | 1.75 GiB | parity: agg 75.7 vs 70.9 (per replica 25.2 vs 23.6) |
| Qwen3-ASR 1.7B | DP3 (40k) | DP3 | 28.2 GB | 20.2 GB | 3.83 GiB | parity: agg 67.5 vs 64.5 (per replica 22.5 vs 21.5) |
| FunASR Nano | DP2 (30k) | DP2 | 11.8 GB | 10.2 GB | 1.57 GiB | parity: agg 40.6 vs 42.5 (per replica 20.3 vs 21.3) |

Rows are ordered by what sharing unlocked. Fixed-N parity is the mechanism-level expectation; single-pass cells are observations pending repetition. Sharing lifts the DP ceiling where weights bind: MOSS delay DP1 to DP2 (+32%), MOSS local edge to operable DP3 (+29% over DP2), Higgs DP3 to DP4 on H100 (+10% to +40%); Whisper kept scaling to DP6 (+40%), knee unfound. Not-supported models were removed from this table: their earlier boot/VRAM observations are exploratory data, not support evidence; audits are recorded in the audit-only registry.

## Modifications

**Core IPC library** — `sglang_omni/utils/ipc_weights.py` (new, ~510 LOC)
- Leader: after model load, exports every eligible weight tensor's CUDA IPC handle + metadata to a handle file under the run's state dir.
- Follower: imports the handles and aliases its module parameters onto the leader's storage (zero-copy attach; logs e.g. `attached 292 shared tensors (7.55 GiB aliased, zero-copy)`).
- Safety: device/UUID validation, dtype/shape checks, and clear failure modes documented (leader owns the storage — a run must always be restarted whole).

**Runner / config integration**
- `model_runner/sglang_model_runner.py`: `load_model` override — leader exports after load; follower attaches instead of loading (driven by `SGLANG_OMNI_WEIGHT_SHARE=leader:<dir>|follower:<dir>`).
- `config/schema.py`, `models/higgs_tts/engine_builder.py`, `pipeline/mp_runner.py`: plumbing.
- `models/higgs_tts/model_runner.py`: sync-free decode-launch cache — skips the composition-invariant sampling-param extraction/upload when batch composition is unchanged between decode steps (`SGLANG_OMNI_SYNCFREE_LAUNCH`, default ON).
- `models/qwen3_tts/config.py`: `mem_fraction_role_to_stage` so `--mem-fraction-static` CLI overrides pin the TTS engine stage (used by autodp and the Qwen3-TTS admission-depth experiments below).

**Launchers**
- `examples/mps_dp/launch.sh`: `WEIGHT_SHARE=1` toggle on the MPS-DP launcher (leader/follower roles, manifest field, teardown guidance) + `SERVE_EXTRA_ARGS` passthrough. Rebased onto the unified `CONFIG` launcher machinery from [sgl-project#1070](https://github.com/sgl-project/sglang-omni/pull/1070)/[sgl-project#1076](https://github.com/sgl-project/sglang-omni/pull/1076).
- `examples/mps_dp/autodp.sh` (new): `plan` probes one replica at the KV cap to measure the static footprint `s` and weight bytes `W`, then sizes max safe DP from
  `W + D·(s − W + h) ≤ M` (shared) / `D·(s + h) ≤ M` (unshared), and derives the last replica's required mem fraction; `up` launches at that size.

**Docs / benchmark**
- `docs/basic_usage/mps_dp.md`: weight-sharing usage + operational caveats.
- `benchmarks/eval/benchmark_tts_seedtts.py`: `--sample-offset` for disjoint per-client shards.
- `benchmarks/eval/bench_sweep.py` (new): open-loop request-rate sweep driver — per offered rate, one Poisson client per replica (`--concurrency 0`), harvesting per-request TTFC/RTF/latency into `sweep.json` with full sorted arrays for CDFs. This is what produced the open-loop knee data below; it fills the request-rate-sweep gap noted against `bench_serving`.

## Related Issues

Builds on [sgl-project#1037](https://github.com/sgl-project/sglang-omni/pull/1037) (equal-KV enforcement for same-GPU DP), [sgl-project#1042](https://github.com/sgl-project/sglang-omni/pull/1042) (same-GPU DP concept docs), [sgl-project#1070](https://github.com/sgl-project/sglang-omni/pull/1070) (unified Omni example launchers), and [sgl-project#1076](https://github.com/sgl-project/sglang-omni/pull/1076) (H100/H200 MPS DP hardware profiles).

## Accuracy Test

No model-math changes: followers alias the identical weight storage the leader loaded, and the sync-free launch cache is covered by CPU-only equivalence tests asserting cached decode steps produce buffers identical to the rebuild-every-step baseline (`tests/unit_test/higgs_tts/test_syncfree_launch_cache.py`, 20+ cases including admission/finish/reorder/pool-reallocation invalidation).

## Benchmark & Profiling

All numbers measured 2026-07-21 on this PR's merged tree, 1× H200 (141 GB), seed-tts-eval EN, streaming, open-loop Poisson clients (`benchmarks/eval/bench_sweep.py`), with **max-DP discovery, sizing, and launch done end-to-end by `examples/mps_dp/autodp.sh`** (probe → size → derive MF → launch, both sizing branches; the probe reads `W` from the leader's CUDA-IPC export log).

These series are historical evidence for the mechanism and sizing, recorded at the PR revision current on their date; they are not the support gate. Support is defined solely by the current-revision validation in `docs/basic_usage/mps_dp.md`; the Qwen3-TTS and Ming rows below predate the validated-architecture gate (both are not supported today).

### Largest DP without weight sharing vs largest DP with it

Weight sharing's (ws) headline is not parity at equal N — the refund **moves the max-DP frontier**. At a common 100k-token KV cap per replica:

| model | s (GiB/replica) | W (GiB) | max DP ws0 | max DP ws1 |
|---|---:|---:|---:|---:|
| Higgs TTS 3-4B | 24.1 | 7.55 | **5** (DP6 fails: KV under-resolution / runtime OOM) | **7** |
| Qwen3-TTS 0.6B | 14.1 | 1.75 | **8** | **9** |
| Ming-Omni-TTS 16.8B-A3B (MoE, TP1) | 41.7 | 31.42 | **3** | **8** (all 8 boot at the full cap; 7 followers attach the leader's 31.4 GiB zero-copy) |

Head-to-head at max DP, run at two CPU budgets (24 vs 48 server cores) to separate CPU starvation from real limits — Higgs (0 failures in all 24 stages):

| budget | config | peak achieved qps | TTFC p50 floor | TTFC p50/p99 at peak | RTF p50 at peak | VRAM |
|---|---|---|---|---|---|---|
| 24 srv cores | DP5 ws0 | 68.7  | 0.22 s | 0.27 / 1.68 s | 0.68 | 122.8 GiB |
| 24 srv cores | DP7 ws1 | 68.9 | 0.29 s | 0.56 / 2.07 s | 1.13 | 125.2 GiB |
| 48 srv cores | DP5 ws0 | 72.2 | 0.23 s | 0.66 / 2.50 s | 0.97 | — |
| 48 srv cores | DP7 ws1 | **73.1**  | 0.26 s | 1.24 / 3.22 s | 1.29 | — |

Findings:

- **Throughput**: both Higgs configs converge (~69 qps at 24 cores, ~72–73 at 48) — by DP5 the SM array is effectively saturated for this model, so the IPC weight-sharing (WS) frontier replicas add ≤1.3% qps even generously CPU-provisioned. **The frontier's value is VRAM and existence, not qps**: DP7 runs 2 extra replicas for +2.4 GiB total where unshared DP7 (~168 GiB) cannot boot, and at matched N the refund is 7.55 GiB × (N−1) — 22.6 GiB at DP5-equivalent load. Sub-knee TTFC and RTF are unchanged by sharing at every measured rate.
- **Qwen3-TTS**: DP9-ws1 matches DP8-ws0 on the runaway-robust metric (RTF p50 1.15–1.26 at matched load, both budgets) with a 14 GiB refund (8 × 1.75 GiB); aggregate qps at this model is noise-dominated by a documented 2048-token runaway tail (p99 pinned ~77–85 s in affected stages — the production fix is a `max_new_tokens` cap, not more replicas). Extra CPU moves neither model's ceiling: Higgs saturates at the GPU, Qwen at per-step GPU work under MPS contention plus admission depth.
- **Mechanism reach (historical, pre-gate)**: in these exploratory runs the leader exported its full weight set on a config-file pipeline (Qwen3-TTS) and 8 followers attached zero-copy with KV profiling, CUDA-graph capture, and health passing on IPC-aliased storage. This demonstrated the protocol beyond Higgs; it is not a support claim, and Qwen3-TTS remains not supported pending its own end-to-end validation.
- **Practical guidance**: spend the WS refund on KV capacity or admission depth at these model sizes; frontier *replicas* convert to throughput only while the SM array still has idle to reclaim.

### When does the frontier convert to throughput?

The three models span the whole answer space, and the discriminator is one comparison: **max-DP(ws0) vs the DP at which the GPU saturates for that model.**

- **Higgs 3-4B**: saturation ≈ DP5 = its ws0 ceiling → the WS replicas (DP6–7) find no idle to reclaim; gain ≈ 1.0× (measured, both CPU budgets). The refund's value is VRAM and the existence of configurations that cannot otherwise boot.
- **Qwen3-TTS 0.6B**: ws0 ceiling (DP8) already sits at its contention/admission-bound plateau → +1 replica is throughput-neutral (measured, RTF-parity).
- **Ming-Omni-TTS 16.8B**: the interesting regime — weights are so large that the ws0 ceiling (**DP3**) sits well *below* saturation, so the five WS-enabled replicas have real idle SM/HBM to reclaim. Anchoring on the model's reference numbers (TP1 on H100: ~1.2–1.4 qps/pipeline at RTF ≈ 1.45) scaled to H200 bandwidth, we gain **~1.5–2.0× aggregate throughput from DP8-ws1 over DP3-ws0** at an adequate CPU budget (~1.1–1.5× when CPU-starved at ~3 cores/replica) — i.e., for large-weight models the WS refund converts directly into throughput, not just VRAM. 

Rule: the larger the weights relative to the card, the lower the ws0 frontier falls below the saturation point — and the more of the sharing refund comes back as *throughput* rather than capacity. Weight sharing is most valuable on exactly the models where replication is least affordable.

## Checklist

- [x] Add unit tests. (`test_ipc_weights.py` incl. exact-set locks for both registries; `test_mps_dp_support.py` claim-to-execution suite: registry/docs/shipped-config consistency and fail-closed launcher paths; `test_weight_share_load_path.py`; `test_syncfree_launch_cache.py`. 76 tests pass on the validation box at the current head.)


sglang_omni/model_runner/sglang_model_runner.py
```diff
@@ -75,6 +75,112 @@ def __init__(
...
+
+        # TP/PP ranks are separate processes inheriting the env var and their
+        # shards share names/shapes/dtypes across ranks, so the handle file
+        # would collide and followers would silently attach another rank's
+        # shard. Refuse until the handle path is rank-qualified.
+        if getattr(self.server_args, "tp_size", 1) > 1 or (
+            getattr(self.server_args, "pp_size", 1) > 1
+        ):
+            raise ipc_weights.WeightShareError(
+                "SGLANG_OMNI_WEIGHT_SHARE requires tp_size == pp_size == 1"
```
JiaxinD: MOSS, Qwen3-TTS and Ming overwrite a registered embedding weight with per-request scratch every decode step, so sharing that parameter silently corrupts concurrent replicas. Restrict sharing to a validated-architecture allowlist, or exclude mutable parameters from the shared set.

sglang_omni/utils/ipc_weights.py
```diff
@@ -0,0 +1,510 @@
...
+                f"[weight-share] timed out after {timeout_s:.0f}s waiting for "
+                f"a leader handle file under {dir_path}; is the leader "
+                "replica up?"
+            )
+        time.sleep(poll_interval_s)
+
+
+def _load_payload(file_path: str, model: torch.nn.Module) -> dict[str, Any]:
+    with open(file_path, "rb") as fh:
+        payload = pickle.load(fh)
```
JiaxinD: The handle file is pickle-loaded from a world-accessible /tmp path with no owner or mode check, so a local user can read the IPC handle or plant a file. Validate dir and file uid, reject group/world bits and symlinks before load, and write the dir 0700 / file 0600.

sglang_omni/utils/ipc_weights.py
```diff
@@ -0,0 +1,510 @@
...
+    while not os.path.exists(file_path):
+        if time.monotonic() >= deadline:
+            raise TimeoutError(
+                f"[weight-share] timed out after {timeout_s:.0f}s waiting for "
+                f"leader handle file {file_path}; is the leader replica up?"
+            )
+        time.sleep(poll_interval_s)
+
+
+def wait_for_any_export(
```
JiaxinD, line 636: The leader pid is recorded in the payload but never checked, so if the leader dies followers keep serving on freed IPC memory. Add a liveness check before attach and a monitor that exits the follower when the leader is gone.

sglang_omni/model_runner/sglang_model_runner.py
```diff
@@ -75,6 +75,112 @@ def __init__(
...
+        finally:
+            self.server_args.load_format = original_load_format
+        self._weight_share_record = ipc_weights.follower_attach(
+            self.model,
+            ws.dir_path,
+            timeout_s=ws.attach_timeout_s,
+        )
+        # Return the dropped dummy-weight blocks to the driver so KV-pool
+        # profiling (and later replicas) see the freed memory.
+        torch.cuda.empty_cache()
```
JiaxinD, line 166: empty_cache here hands the freed weight memory to KV profiling, so a follower launched without an explicit max-total-tokens over-budgets KV by the shared-weight size. Require an explicit cap for followers; the launcher sets one but bare CLI does not.

sglang_omni/utils/ipc_weights.py
```diff
@@ -0,0 +1,510 @@
...
+    ForkingPickler's CPU shared-memory reduction does not survive a
+    file-based handoff between unrelated processes; followers copy those.
+
+    Returns a record ``{name: (data_ptr, shape, dtype)}`` of the IPC-shared
+    (leader-side) tensors for later :func:`verify_attachment` — the leader's
+    own storages must stay put too, since followers alias them. In-place
+    mutation (``copy_``) is fine; rebinding ``.data`` after export is not.
+    """
+    abs_path = os.path.abspath(file_path)
+    if abs_path in _EXPORTED_FILES:
```
JiaxinD, line 564: _EXPORTED_FILES is process-local, so it does not catch a stale file from a prior run or a second same-class engine writing the same path. Clear any existing file before export and bind the payload to a run id.

tests/unit_test/utils/test_ipc_weights.py
```diff
@@ -0,0 +1,299 @@
...
+    WeightShareError,
+    attach_weights,
+    export_weights,
+    get_weight_share_config,
+    handle_file_for_model,
+    verify_attachment,
+)
+
+
+class IdentitySerializer:
```
JiaxinD, line 34: These tests use a same-process identity serializer, so they prove the assignment path but not real cross-process CUDA aliasing. A two-process spawn test that mutates on the leader and reads on the follower would lock in the zero-copy contract.

sglang_omni/model_runner/sglang_model_runner.py
```diff
@@ -75,6 +75,112 @@ def __init__(
...
+        """
+        record = getattr(self, "_weight_share_record", None)
+        if record is not None:
+            from sglang_omni.utils import ipc_weights
+
+            ipc_weights.verify_attachment(self.model, record)
+        return super().init_device_graphs()
+
+    def _weight_update_blocked_reason(self) -> str | None:
+        ws = getattr(self, "_weight_share_config", None)
```
JiaxinD: getattr(self, "_weight_share_config", None) fails open: a path reaching update_weights without going through load_model would skip the guard. Set the attribute in __init__ so the guard is always live.

sglang_omni/model_runner/sglang_model_runner.py
JiaxinD: Done: Higgs-only allowlist checked in load_model before the model build (validate_weight_share_architecture); extend after a per-model audit.

sglang_omni/utils/ipc_weights.py
JiaxinD: Done: store dir 0700 / file 0600, uid+mode+O_NOFOLLOW validation before unpickle, mkstemp write.

sglang_omni/utils/ipc_weights.py
JiaxinD: Done: pid liveness (zombie + start-time) before and after attach, plus a monitor that exits the follower on leader death.

sglang_omni/model_runner/sglang_model_runner.py
JiaxinD: Done: a follower now requires an explicit --max-total-tokens.

sglang_omni/utils/ipc_weights.py
JiaxinD: Done: clear the stale file before export, bind model_path/revision and leader start-time; cross-run dir reuse stays a documented per-run-dir contract.

tests/unit_test/utils/test_ipc_weights.py
JiaxinD: Done: added test_ipc_weights_cuda.py, a two-process test asserting the follower sees the leader's in-place mutation through shared storage.

sglang_omni/model_runner/sglang_model_runner.py
JiaxinD: Done: the attributes are set in __init__ so the guard reads them directly, no getattr fallback.

sglang_omni/model_runner/sglang_model_runner.py
```diff
@@ -75,6 +79,135 @@ def __init__(
             server_args=server_args,
         )
 
+    # ------------------------------------------------------------------
+    # Same-GPU weight sharing (SGLANG_OMNI_WEIGHT_SHARE=leader|follower:<dir>)
+    # ------------------------------------------------------------------
```
Ratish1: remove these type of comments

sglang_omni/models/higgs_tts/engine_builder.py
```diff
@@ -108,6 +108,10 @@ def setup_model(
     ) -> None:
         del checkpoint_dir, device, gpu_id, server_args
         self.model = model_worker.model_runner.model
+        # Runs AFTER the weight-share leader export: under
+        # SGLANG_OMNI_WEIGHT_SHARE this mutates storage shared with follower
+        # replicas, which is safe only while it stays in-place, idempotent,
+        # and deterministic (see sglang_omni/utils/ipc_weights.py docstring).
```
Ratish1: no need for this comment I think

examples/mps_dp/autodp.sh
```diff
@@ -0,0 +1,193 @@
+#!/bin/bash
+# autodp.sh — size and launch same-GPU DP automatically.
+#
+# Computes the maximum SAFE DP for this GPU + model + KV cap from the
+# two-constraint sizing model (see note/2026-07-18-h200-same-gpu-dp-ipc-profiling.md):
```
Ratish1: seems like a local file comment that shouldnt be here?

sglang_omni/utils/ipc_weights.py
```diff
@@ -0,0 +1,812 @@
...
+Role selection is environment-driven so process supervisors (e.g.
+``examples/mps_dp/launch.sh``) can stay declarative::
+
+    SGLANG_OMNI_WEIGHT_SHARE=leader:/path/to/dir     # replica 0
+    SGLANG_OMNI_WEIGHT_SHARE=follower:/path/to/dir   # replicas 1..N-1
+
+The path is a *directory*; each SGLang engine writes/reads
+``<dir>/<ModelClass>.weights-ipc``, so multi-engine stages cannot clobber
+each other. Unset (or empty) means weight sharing is off.
+"""
```
zhaochenyang20, line 37: comments should follow our requirments:

https://github.com/zhaochenyang20/sglang-diffusion-routing/issues/32#issuecomment-4931931943

sglang_omni/utils/ipc_weights.py
```diff
@@ -0,0 +1,812 @@
...
+logger = logging.getLogger(__name__)
+
+ENV_WEIGHT_SHARE = "SGLANG_OMNI_WEIGHT_SHARE"
+ENV_WEIGHT_SHARE_TIMEOUT_S = "SGLANG_OMNI_WEIGHT_SHARE_TIMEOUT_S"
+DEFAULT_ATTACH_TIMEOUT_S = 1800.0
+_FORMAT_VERSION = 1
+_FILE_SUFFIX = ".weights-ipc"
+
+# Paths this process has already exported to; a second export to the same
+# file would silently invalidate handles followers may already hold.
```
zhaochenyang20: https://github.com/zhaochenyang20/sglang-diffusion-routing/issues/32#issuecomment-4918483294

sglang_omni/utils/ipc_weights.py
```diff
@@ -0,0 +1,812 @@
...
+
+    def _run(self) -> None:
+        while not self._stop.wait(self.poll_interval_s):
+            if pid_is_alive(self.leader_pid):
+                continue
+            logger.critical(
+                "[weight-share] leader pid=%s exited; terminating follower",
+                self.leader_pid,
+            )
+            os._exit(self.exit_code)
```
zhaochenyang20, line 483: `_check_leader_alive` compares `leader_start_time` before and after attachment, but `LeaderLivenessMonitor._run` only calls `pid_is_alive(self.leader_pid)` after serving begins. If the leader exits and its PID is reused before the next poll, the monitor treats an unrelated process as the original leader and the follower keeps running on invalid CUDA IPC mappings, which can crash or return corrupt output. This also means commit `1aaca0c6` protects only the attachment instant, not the runtime lifetime.

examples/mps_dp/autodp.sh
zhaochenyang20, line 1: After a successful probe, the code derives both `run_id` and `state_dir` from separate `launch.sh list | head -1` calls. `list` lexically enumerates every `STATE_ROOT/gpu-*` directory; it is neither filtered by `GPU_ID` nor bound to the just-completed `launch.sh up` output. If another GPU has the lexically first run, this reads the wrong log and then executes `launch.sh down "$run_id"` against that unrelated run while leaving the actual probe alive.

zhaochenyang20: When a pipeline rejects `--mem-fraction-static`, this retry scans every state directory for the target GPU, checks only the top-level PIDs from `replicas.tsv` with `kill -0`, and directly `rm -rf`s directories it considers inactive. That discards the start-time identity checks, PGID child checks, MPS-client ownership checks, and fail-closed state retention already implemented by `launch.sh teardown_state`. If the first cleanup failed precisely because a child or MPS daemon survived, this deletes the only tracking evidence, leaves unmanaged GPU/MPS resources behind, and proceeds to launch another run.

benchmarks/eval/bench_sweep.py
zhaochenyang20, line 1: This modulo scheme does not guarantee that intervals of length `SAMPLES` are disjoint within a stage. With `EN_TOTAL=1088, SAMPLES=200, N=5`, the second stage produces offsets `112, 312, 512, 712, 24`, so `24..223` overlaps `112..311`; `N=9` overlaps in the first stage. The PR body relies on DP5, DP7, and DP9 sweeps, so this is not a fringe input: shared-cache hits can change achieved QPS and TTFC and invalidate the benchmark evidence. `SAMPLES == EN_TOTAL` also makes the modulus divisor zero.

examples/mps_dp/autodp.sh
zhaochenyang20: `autodp.sh` explicitly presents Qwen3-TTS as the `CONFIG` example while defaulting `WEIGHT_SHARE=1`, but `SUPPORTED_WEIGHT_SHARE_ARCHITECTURES` currently allows only `HiggsMultimodalQwen3ForConditionalGeneration`. Following the documented Qwen3-TTS config path therefore fails before the probe model loads. More importantly, the PR body still presents Qwen3-TTS DP9 and Ming DP8 zero-copy attachment as evidence for the current head, even though the final hardening commits explicitly reject those paths. Reviewers cannot reproduce the headline cross-model claims from the code being reviewed.

benchmarks/eval/bench_sweep.py
zhaochenyang20: Each client's log file and `Popen` are created incrementally, but the overall spawn/wait lifecycle has no `try/finally`. If a later `open`, `Popen`, or `preexec_fn` fails, previously started Poisson clients continue sending traffic and their log handles remain open. That contaminates the next benchmark stage and makes failure diagnosis unreliable. This is an external-process resource boundary that requires explicit cleanup.

sglang_omni/utils/ipc_weights.py
zhaochenyang20, line 1: After confirming that the payload is a dictionary with the expected version, `_load_payload` immediately indexes `payload["model_class"]`; later code directly indexes `manifest_hash`, `ipc_blob`, `ipc_names`, and `value_blobs`. A truncated, manually damaged, or incomplete old private file leaks `KeyError` instead of the protocol's documented `WeightShareError`. This does not silently use wrong weights, but it removes file/field context from startup failures and leaves the actual payload schema scattered across several functions.

examples/mps_dp/autodp.sh
zhaochenyang20: When the formula produces `d <= 0`, the subsequent `d = max(d, 1)` turns "not even one replica plus headroom fits" into "max safe DP = 1." The script also does not validate the ranges of `s`, `w`, `h`, `MAX_DP`, or `MIN_CORES_PER_REPLICA`; `s - w + h <= 0` can divide by zero or invert the result. A public launcher claiming to calculate the "maximum SAFE DP" should reject these inputs before creating resources rather than emit a contradictory plan and wait for startup OOM.

sglang_omni/model_runner/sglang_model_runner.py
```diff
@@ -75,6 +79,132 @@ def __init__(
...
+        ):
+            raise ipc_weights.WeightShareError(
+                "SGLANG_OMNI_WEIGHT_SHARE requires tp_size == pp_size == 1"
+            )
+
+        architectures = (
+            [self._model_arch_override]
+            if self._model_arch_override is not None
+            else getattr(self.model_config.hf_config, "architectures", None)
+        )
```
zhaochenyang20, line 115: `tp_size` and `pp_size` are required `ServerArgs` fields and form the central contract deciding whether sharing is safe. Using `getattr(..., 1)` silently treats an upstream rename, incomplete mock, or broken initialization path as TP1/PP1, which is exactly the most permissive safety decision. At this model-dispatch boundary, dynamic attributes and defensive defaults prevent type checking and refactors from protecting the invariant.

benchmarks/eval/benchmark_tts_seedtts.py
```diff
@@ -535,6 +548,7 @@ def _build_arg_parser() -> argparse.ArgumentParser:
     )
     parser.add_argument("--output-dir", type=str, default="results/tts_seedtts")
     parser.add_argument("--max-samples", type=int, default=None)
+    parser.add_argument("--sample-offset", type=int, default=0)
```
zhaochenyang20, line 558: The new option uses plain `type=int`, so a negative value invokes Python's slice-from-the-end semantics instead of meaning "skip the first N samples." `_build_results_config` also omits `sample_offset`, so each client's `speed_results.json` cannot prove which dataset segment it measured. For an option introduced specifically to prevent cache contamination, these gaps break both misconfiguration handling and result provenance.

zhaochenyang20 (changes requested):

sglang_omni/model_runner/sglang_model_runner.py
JiaxinD: done

sglang_omni/models/higgs_tts/engine_builder.py
JiaxinD: done

examples/mps_dp/autodp.sh
JiaxinD: done

sglang_omni/utils/ipc_weights.py
JiaxinD: Done: removed backticks, kept why-only, signed Note (author) across the module.

sglang_omni/utils/ipc_weights.py
JiaxinD: Done in the same comment-style pass.

sglang_omni/utils/ipc_weights.py
JiaxinD: Done: the liveness monitor now compares the leader start-time too, so a mid-serving pid reuse is caught, not only the attach instant.

examples/mps_dp/autodp.sh
JiaxinD: Done: probe discovery and teardown scope to this GPU's newest run (not list | head -1); the mem-fraction retry tears down via launch.sh down, d<=0 is rejected, and sizing inputs are validated.

benchmarks/eval/bench_sweep.py
JiaxinD: Done: each client takes a disjoint slot ((stage+i)%NUM_SLOTS) with a SAMPLES>=EN_TOTAL guard; spawn/wait is wrapped in try/finally and the results file is closed.

sglang_omni/utils/ipc_weights.py
JiaxinD: Done: the payload is validated as a closed schema and unpickle/deserialize are wrapped, so a truncated or forged handle raises WeightShareError not KeyError.

sglang_omni/model_runner/sglang_model_runner.py
JiaxinD: Done: read tp_size/pp_size directly, no getattr default.

benchmarks/eval/benchmark_tts_seedtts.py
JiaxinD: Done: reject a negative offset and record it in the results config for provenance.

JiaxinD:
**Why sharing is gated per architecture, and how MOSS is shared safely**

Sharing aliases every registered tensor across replicas, which is only safe if they are all read-only at serving time. MOSS local, Qwen3-TTS, and Ming all stage per-request decode feedback into a registered embedding every decode step, e.g. `moss_tts_local/model_runner.py`:

```python
weight[:batch_size].copy_(pool.feedback_embeds[row_tensor])
```

(Qwen3-TTS: `decode_feedback_embedding.weight[:batch_size].copy_(stacked)`; Ming: `weight[:batch_size].copy_(feedback_embeddings.to(...))`.) With one shared storage, replica B can overwrite these rows between replica A's staging write and graph replay, so A decodes B's request. Nothing errors; the audio is silently wrong.

Each audited architecture now carries a share policy naming its replica-private tensors:

```python
"MossTTSLocalSGLangModel": WeightSharePolicy(
    private_tensor_names=frozenset({"_decode_input_embedding.weight"})
)
```

Immutable weights still alias one storage. Policy-private tensors ship by value: the follower copies into storage its own build created and never rebinds it, so graph-captured addresses stay replica-local. The classification is fail-closed protocol, not convention: it joins the manifest hash and payload (handle format v2), a policy naming a missing tensor refuses to start, a blob that IPC-shares a private tensor is refused, and shared rebinds are pointer-checked against private tensors before any private copy runs.

Current scope: Higgs (full sharing, no serving-time writes found in audit) and MOSS local (selective) pass the gate; all other architectures are rejected at startup.

Validated on one H100: 57 targeted tests green on GPU (POSIX fs-trust, two-process CUDA IPC isolation with forced write ordering plus a shared-scratch hazard arm, runner load path). MOSS A/B/C e2e at a common 60000-token cap: the follower attaches 8.44 GiB zero-copy and its process memory drops from 30.5 to 21.8 GiB; all 16 seeded reference outputs are byte-identical to the single-replica baseline on both replicas, including while the neighbor replica serves 190+ concurrent foreign requests (both directions, 0 errors); per-round DP2 throughput matches unshared DP2 (15.6/18.2 vs 15.5/17.9 qps aggregate).

Qwen3-TTS and Ming look policy-coverable (one private tensor each) but stay out for now: Qwen3-TTS's runtime module tree includes external `qwen_tts` components the in-repo audit cannot close, and Ming has TP-specific deployment and no DP e2e yet. Adding a model later = post-load mutation audit, a policy entry, isolation tests, DP e2e validation.

JiaxinD:
Scope update (supersedes the earlier tier table): support is now binary. **Supported (7): Higgs, MOSS local, MOSS delay, MOSS-TD, Qwen3-ASR, Whisper, FunASR** — each passed a one-H100 end-to-end smoke at the current head (shared N=2 boot, attach verification, concurrent correctness, clean teardown and restart) via its shipped config. Everything else is **not supported** and rejected in preflight before any resource is created; completed audits (Ming TTS, Voxtral, Fish, LLaDA2, Qwen3-TTS, both Qwen3-Omni engines) are recorded as audit-only in the gate, support in progress. Tests pin gate = launcher registry = shipped configs = docs table.

Also in this push: ASR launcher role maps, SGLANG_OMNI_STRICT_PORT=1, Fish rope truncation made in-place.

JiaxinD:
Off-vs-on measurements for the supported models (80 GB H100; support standard and evidence in `docs/basic_usage/mps_dp.md`):

| Model | Validated DP, off | Validated DP, on | VRAM, off | VRAM, on | Saved per follower | Throughput (aggregate), off vs on |
|---|---|---|---:|---:|---:|---|
| MOSS TTS delay | **DP1** (unshared DP2: replica 1 cannot fit its own weight copy) | **DP2** | n/a | 42.4 GB | 17.05 GiB | single 5.7 to shared DP2 agg 7.5 qps (+32%, per replica 3.8, preliminary); aligned-history byte-identical |
| Higgs TTS 3-4B | DP3 (unshared DP4 needs 98 GB) | **DP4** (74.9 GB idle, 78.4 load) | 73.7 GB at DP3 | 58.1 GB at DP3 | 7.55 GiB | parity at every N; shared DP4 +10% to +40% over shared DP3, round-matched |
| MOSS TTS local | DP3 at the card edge, graphs partly eager | DP3, 13.3 GB headroom, full graphs | 78.0 GB | 61.8 GB | 8.44 GiB | parity: DP2 17.9 vs 18.2, DP3 23.4 vs 23.4 (per replica 7.8) |
| Whisper large-v3-turbo | DP3 (40k) | **DP6** (19.7 GB) | 14.1 GB at DP3 | 10.7 GB at DP3 | 1.51 GiB | parity at DP3 (67.8 vs 68.0); shared DP6 agg 95.3 cold qps, +40% over DP3 |
| MOSS Transcribe-Diarize | DP3 (40k) | DP3 | 23.9 GB | 20.2 GB | 1.75 GiB | parity: agg 75.7 vs 70.9 (per replica 25.2 vs 23.6) |
| Qwen3-ASR 1.7B | DP3 (40k) | DP3 | 28.2 GB | 20.2 GB | 3.83 GiB | parity: agg 67.5 vs 64.5 (per replica 22.5 vs 21.5) |
| FunASR Nano | DP2 (30k) | DP2 | 11.8 GB | 10.2 GB | 1.57 GiB | parity: agg 40.6 vs 42.5 (per replica 20.3 vs 21.3) |

Rows are ordered by what sharing unlocked. Fixed-N parity is the mechanism-level expectation; single-pass cells are observations pending repetition. Sharing lifts the DP ceiling where weights bind: MOSS delay DP1 to DP2 (+32%), MOSS local edge to operable DP3 (+29% over DP2), Higgs DP3 to DP4 on H100 (+10% to +40%); Whisper kept scaling to DP6 (+40%), knee unfound. Not-supported models were removed from this table: their earlier boot/VRAM observations are exploratory data, not support evidence; audits are recorded in the audit-only registry.

zhaochenyang20 (changes requested):
https://github.com/yl3469/sglang-omni/blob/auto-dp-cuda-ipc/examples/mps_dp/config.py

here:

the latest commits advertise LLaDA2 and Qwen3-Omni as supported

I do not believe so

I would block this because the latest commits advertise LLaDA2 and Qwen3-Omni as supported, while the documented `launch.sh up` path for `N>1` still assumes the whole pipeline has one `generation` SGLang stage. Both `LLaDA2UniPipelineConfig` and the text-only `Qwen3OmniPipelineConfig` lack a `generation_sglang_role_to_stage()` mapping and fail at line 27 with “does not declare a generation stage.” The speech Qwen3-Omni config maps generation to `talker_ar`, but its `mem_fraction_role_to_stage()` contains both `thinker` and `talker_ar`, so the singleton check at line 34 rejects it instead. `launch.sh:353-358` enables this check unconditionally for every `N>1` config, leaving the documented two handle files and per-stage follower KV caps unreachable and causing startup to fail.

examples/mps_dp/launch.sh
```diff
@@ -387,7 +401,9 @@ up() {
   local uuid node run state
   uuid=$(nvidia-smi --query-gpu=uuid --format=csv,noheader -i "$gpu")
   node=$(resolve_numa "$gpu")
-  run="run-$(date +%Y%m%d-%H%M%S)-$$"
+  # Note (Jiaxin Deng): a caller (autodp) may pin RUN_ID so it can tear down exactly
+  # the run it started, instead of rediscovering the newest dir.
+  run="${RUN_ID:-run-$(date +%Y%m%d-%H%M%S)-$$}"
```
zhaochenyang20, line 420: The newly added `RUN_ID` is interpolated directly into `state=$STATE_ROOT/gpu-$gpu/$run` without requiring a single `run-*` basename or rejecting `/`, `..`, and control characters. With `RUN_ID=../gpu-1/run-x`, a GPU-0 launch lands in GPU 1's namespace and bypasses the active/stale-run guard that scans only `gpu-$gpu/run-*`; deeper traversal can escape `STATE_ROOT` entirely. Startup truncates `replicas.tsv` and overwrites `manifest` there, while a successful later `down` reaches `rm -rf -- "$state"` at line 309. A public identifier has therefore become an overwrite and recursive-deletion path.

benchmarks/eval/bench_sweep.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python3
```
zhaochenyang20: remove this shaba

benchmarks/eval/bench_sweep.py
```diff
@@ -0,0 +1,213 @@
...
+                  latency/RTF carry the comparison
+    BENCH_REF     1 (default) sends seed-tts-eval reference audio
+                  (--ref-format references); 0 benchmarks text-only synthesis
+                  (--no-ref-audio) for pipelines whose reference decode is
+                  unavailable in the serving environment
+    CLIENT_CORES  optional comma-separated CPU list to pin client processes
+                  to (keeps load generation off the server cores); unpinned
+                  when unset
+
+Per-stage results (offered vs achieved qps, TTFC/RTF/latency percentiles,
```
zhaochenyang20: add author name if have long comments

benchmarks/eval/bench_sweep.py
```diff
@@ -0,0 +1,213 @@
...
+    achieved_runner = audio_sps = 0.0  # from each client's runner wall clock
+    for i in range(N):
+        f = os.path.join(stage_dir, f"client{i}", "speed_results.json")
+        if not os.path.exists(f):
+            continue
+        with open(f) as fh:
+            d = json.load(fh)
+        for r in d["per_request"]:
+            if r.get("is_success"):
+                lats.append(r["latency_s"])
```
zhaochenyang20, line 144: Each stage/client uses a deterministic directory derived from the rate and client index, but the script neither rejects an existing stage directory nor removes an old `speed_results.json` before launching. If `OUT` is reused and the current client fails before writing results, line 144 still finds the previous run's file and line 147 loads it; `completed`, TTFC, RTF, and `achieved_qps_runner` then incorporate stale data. Although `client_fails` records the process failure, it does not prevent the stage from being written to `sweep.json` or consumed as a valid measurement. This can fabricate the benchmark evidence used for capacity-knee and parity claims.

examples/mps_dp/autodp.sh
```diff
@@ -0,0 +1,206 @@
...
+m = $M_MIB / 1024
+s = float("$STATIC_GIB")
+w = float("$WEIGHTS_GIB")
+h = float("$H_GIB")
+ws = $WS
+if not (m > 0 and s > 0 and h >= 0 and w >= 0 and (w < s or not ws)):
+    sys.exit("autodp: invalid sizing inputs m=%s s=%s w=%s h=%s" % (m, s, w, h))
+denom = (s - w + h) if ws else (s + h)
+if denom <= 0:
+    sys.exit("autodp: non-positive per-replica cost %.3f GiB; cannot size" % denom)
```
zhaochenyang20, line 172: This file is too dense. Adds more blank line for people to read. Also,

The new code now rejects the main invalid `s/w/h` ranges and `d < 1`, but the remaining public values still enter Python or shell arithmetic without validation. `WEIGHT_SHARE=2` is treated as enabled, `MIN_CORES_PER_REPLICA=0` divides by zero at line 142, and a nonnumeric `MAX_DP` becomes Python syntax; `N=0` passes the upper-bound check at line 161, lets `plan` report “launching N=0,” and only fails later in `core_blocks_for` during `up`. These inputs eventually error in some paths, but not at the promised sizing boundary, and `plan` and `up` disagree on validity.

benchmarks/eval/bench_sweep.py
```diff
@@ -0,0 +1,213 @@
...
+        "--disable-tqdm",
+    ]
+    if os.environ.get("BENCH_STREAM", "1") != "0":
+        cmd.append("--stream")
+    if os.environ.get("BENCH_REF", "1") != "0":
+        cmd.extend(["--ref-format", "references"])
+    else:
+        cmd.append("--no-ref-audio")
+    logf = open(os.path.join(OUT, f"rate{rate:g}_client{i}.log"), "w")
+    preexec = None
```
zhaochenyang20, line 89: The outer `finally` covers clients that were successfully appended to `procs`, but `_spawn_client` opens `logf` before calling `Popen`; if `Popen` raises, the tuple never returns and the outer scope cannot close that newly opened file. For earlier clients, `finally` only sends `terminate()`, immediately closes the log, and propagates without `wait()`, timeout-to-`kill()`, or child-tree handling, so the script cannot establish that traffic stopped before exit. This external-process boundary genuinely requires cleanup; this is not a request for generic defensive exception handling in internal code.

benchmarks/eval/bench_sweep.py
```diff
@@ -0,0 +1,213 @@
...
+        completed += s["completed_requests"]
+        failed += s["failed_requests"]
+        achieved_runner += s.get("throughput_qps") or 0
+        audio_sps += s.get("audio_throughput_s_per_s") or 0
+    ttfcs.sort()
+    rtfs.sort()
+    lats_sorted = sorted(lats)
+
+    def pct(arr, p):
+        return arr[min(int(p * len(arr)), len(arr) - 1)] if arr else None
```
zhaochenyang20: `pct` selects `arr[int(p * len(arr))]`. For four samples, p50 becomes index 2, the third sample; nearest-rank would be index 1, while NumPy's linear percentile interpolates between the middle samples. For 100 samples, p99 becomes index 99, while nearest-rank is index 98. These values are emitted as formal TTFC/RTF/latency p50/p90/p95/p99 metrics, and the repository already uses `numpy.percentile` consistently in `benchmarks.metrics.performance`. The resulting failure mode is that identical requests produce different percentiles across repository benchmarks, preventing capacity-knee and parity tables from sharing one metric definition.

zhaochenyang20 (changes requested):

JiaxinD:
Right, the matrix conflated engine audits with launcher support: LLaDA2 and text Qwen3-Omni fail the resolver at any N (no generation stage) and speech Qwen3-Omni fails the N>1 single-engine check, so none of them can reach the documented launch path. Support is now binary: the gate keeps only the seven models with a passing shared N=2 end-to-end smoke at the current head (Higgs, MOSS local, MOSS delay, MOSS-TD, Qwen3-ASR, Whisper, FunASR), each with a shipped config and recorded boot/attach/concurrent-correctness/teardown evidence; completed audits moved to an audit-only registry the gate rejects, unsupported configs fail in preflight before the MPS daemon or state dir exists, and new tests pin gate = launcher registry = shipped configs = docs table.

examples/mps_dp/launch.sh
JiaxinD: Fixed at head: RUN_ID is validated to a single run-* path component before any path use, with a fail-closed test.

examples/mps_dp/autodp.sh
JiaxinD: Fixed at head: every public knob is validated up front; plan and up now agree.

benchmarks/eval/bench_sweep.py
JiaxinD: Fixed at head: each client's previous output is removed before launch.

benchmarks/eval/bench_sweep.py
JiaxinD: Fixed at head: log handle closed if Popen raises; teardown signals each client's session with TERM, bounded wait, KILL, then reaps.

benchmarks/eval/bench_sweep.py
JiaxinD: Fixed at head: numpy.percentile, same definition as benchmarks.metrics.performance.

benchmarks/eval/bench_sweep.py
JiaxinD: Trimmed and signed.

benchmarks/eval/bench_sweep.py
JiaxinD: Trimmed and signed.

benchmarks/eval/bench_sweep.py
```diff
@@ -0,0 +1,233 @@
...
+    # Note (Yueying Li): remove any previous run's output for this client so a
+    # client that fails before writing results cannot resurrect a stale
+    # speed_results.json into this stage's aggregation.
+    shutil.rmtree(cdir, ignore_errors=True)
+    cmd = [
+        sys.executable,
+        "-m",
+        "benchmarks.eval.benchmark_tts_seedtts",
+        "--generate-only",
+        "--use-existing-server",
```
zhaochenyang20: The new code calls `shutil.rmtree(cdir, ignore_errors=True)` before each client starts, but silently swallows permission, filesystem, and symlink failures; I reproduced a `client0` symlink surviving while the call returned normally. Line 143 only accumulates a nonzero exit into `fails`, lines 159-179 still read `speed_results.json` from that deterministic path, and line 217 still appends the stage. When `OUT` is reused, an undeleted old directory plus a current client failure before overwrite can therefore inject stale TTFC/RTF/QPS into the new `sweep.json`; the prior blocker is only partially fixed.


Do not ignore deletion errors: use a unique run/stage directory, or fail immediately when the exact target cannot be removed, and assert before spawn that it is absent and not a symlink. Require every client result to carry this run's nonce; any nonzero exit, missing result, or nonce mismatch must invalidate the whole stage and prevent formal sweep output. Add symlink, read-only stale-directory, nonzero-exit, and partial-result tests.

zhaochenyang20 (approved):

benchmarks/eval/bench_sweep.py
JiaxinD: Fixed: stage dirs are per-run nonce-named and must not pre-exist (nothing is deleted); any client failure invalidates the whole stage and the sweep exits nonzero. Added symlink, read-only, nonzero-exit, and partial-result tests.

Issue #1121, opened 2026-07-21
[Roadmap] Extend reusable optimizations for host-bound speech serving


AkazaAkane:
## Motivation

#1071 by @MayDomine validated three reusable optimizations for host-bound speech serving:

1. Prefill admission coalescing.
2. Process isolation for CPU/GIL-heavy stages.
3. `torch.compile` for codec, vocoder, or encoder execution.

This issue tracks extending each optimization across all compatible TTS, ASR, and speech-output omni models.

## Tracking

| Component | Scope | Status | PR |
| --- | --- | --- | --- |
| Prefill admission coalescing | Generalize configuration, CLI plumbing, validation, and integration across compatible AR stages | Done| #1073 |
| Process isolation | Generalize placement and resource contracts for CPU/GIL-heavy stages |  Done  | #1125  #1159 |
| Compiled stage execution | Generalize compile, warmup, logging, and eager fallback across compatible stages | Suspended | #1165  |

## 1. Prefill admission coalescing

- [ ] Complete shared configuration and CLI plumbing.
- [ ] Integrate all compatible `OmniScheduler` AR stages.
- [ ] Avoid model-specific factory allowlists where possible.
- [ ] Test timeout, partial admission, abort, KV-pressure, and disabled paths.
- [ ] Benchmark and tune request thresholds and wait times per model.
- [ ] Keep disabled by default unless benchmarks justify enabling it.

## 2. Process isolation

- [ ] Identify stages that cause scheduler CPU or GIL contention.
- [ ] Add reusable process-placement configuration.
- [ ] Enforce actual per-process GPU-memory budgets.
- [ ] Measure duplicated weights and shared-memory/ZMQ overhead.
- [ ] Compare in-process and out-of-process throughput and latency.
- [ ] Validate streaming TTFA/TTFB separately from non-streaming throughput.

## 3. Compiled stage execution

- [ ] Extract shared compile, warmup, logging, and eager-fallback logic.
- [ ] Support model-defined warmup inputs and shape buckets.
- [ ] Add an explicit off switch.
- [ ] Track compile time and recompilation count.
- [ ] Verify eager fallback when compilation or warmup fails.
- [ ] Run correctness and quality parity checks.
- [ ] Enable by default only when cold-start and steady-state results justify it.

## Benchmark requirements

Each optimization PR should report:

- Throughput.
- Mean and p95 latency.
- Streaming TTFA/TTFB where supported.
- CPU utilization.
- GPU active/idle time.
[...]

PR #1125, opened 2026-07-21
[Perf] Add process isolation override for pipeline stages

AkazaAkane:
This PR adds opt-in, model-declared process placement overrides for pipeline stages, including singleton isolation and named process groups. It does not change default model topology. Benchmark results provide guidance for future model-specific default decisions.

## Motivation

Pipeline stages such as a TTS vocoder can benefit from running in a separate process, while frontend stages may need to stay grouped to preserve efficient handoffs. The best topology depends on the model, hardware, concurrency, request shape, and streaming mode. Users need a reusable way to experiment with model-declared process-safe boundaries without changing the topology declared by a model or YAML config.

`StageConfig.process` remains the source of truth. Omitting the CLI option preserves the declared topology; benchmarks in this PR are evidence for future model-specific default decisions rather than a universal default change.

## Modifications

- Add repeatable `--isolate-stage STAGE` singleton placement and `--stage-process STAGE=PROCESS` named/group placement, backed by `apply_stage_process_overrides()`.
- Deep-copy the selected config, support stable model role aliases, reject TP-stage moves, and validate conflicting or malformed assignments.
- Declare model-specific cross-process capability with `process_safe_edges()`, validating only the pipeline edges newly crossed by an override.
- Declare recommended same-GPU fractions with `process_edge_resources()` and apply them from those same newly crossed edges, so equivalent placements receive identical resource defaults while explicit fractions remain authoritative.
- Preserve existing model-default topology and reject model-unsafe boundaries before same-GPU placement validation.
- Support both separated and grouped Higgs-TTS frontend topologies: `preprocessing | audio_encoder | tts_engine | vocoder` and `(preprocessing + audio_encoder) | tts_engine | vocoder`.
- Remove consumed Higgs `reference_audio`, `references`, and `reference_codes` media from `request.inputs` after folding them into `HiggsTtsState`, avoiding repeated cross-process serialization while preserving text and unrelated request fields.
- Add coverage for singleton/grouped placement, edge-based safety and resource defaults, repeatable CLI options, option conflicts, assignment parsing, and Higgs raw/pre-encoded reference pruning.
- Document declarative and CLI placement, grouped topology, model-declared process-safe boundaries, edge resource recommendations, CUDA-context/VRAM/IPC trade-offs, and model applicability.

### Scope

Included in #1125:

- Generic process placement override
- Singleton and grouped CLI options
- Stable role aliases and placement validation
- Model-declared process-safe edges and edge resource recommendations
- Higgs frontend cross-process support and request-input pruning
- Unit tests
- Documentation
- Benchmarks

Not included in #1125:

- Changing model defaults
- Enabling process boundaries universally
- `torch.compile` parameter changes

## Related Issues

ref #1121 

### MOSS TTS LOCAL

| Mode | QPS | Δ vs baseline | Mean latency | p95 |
| :--- | :--- | :--- | :--- | :--- |
| Baseline | 7.711 | — | 2.050 s | 2.527 s |
| Vocoder isolated | 10.700 | +38.8% | 1.484 s | 1.982 s |

### FishAudio S2-Pro

Concurrency/cap: 64.

| Mode | Topology | QPS | Mean / p95 latency | Audio throughput | Mean / p95 / peak SM |
| :--- | :--- | :--- | :--- | :--- | :--- |
| Non-stream | pre \| (engine+voc) | 1.922 | 22.866 / 51.788 s | 6.834 s/s | 42.1 / 91.5 / 93% |
| Non-stream | pre \| engine \| voc | 2.311 | 21.919 / 39.233 s | 8.518 s/s | 53.2 / 89.0 / 93% |
| Stream | pre \| (engine+voc) | 3.811 | 13.464 / 33.668 s | 14.748 s/s | 57.3 / 83.1 / 86% |
| Stream | pre \| engine \| voc | 4.200 | 12.392 / 33.716 s | 16.288 s/s | 66.3 / 89.5 / 97% |

Isolating the vocoder produced:

- Non-streaming: +20.2% QPS, -24.2% p95 latency, and +11.1 percentage points mean SM.
- Streaming: +10.2% QPS, -8.0% mean latency, and +9.0 points mean SM.
- Mean first-audio latency: 8.445 → 7.485 s (-11.4%).
- Median first-audio latency: 2.309 → 1.265 s (-45.2%).
- First-audio p95 was effectively unchanged: 32.531 → 32.554 s.
- Mean inter-chunk interval: 2.230 → 2.182 s.
- Client-visible delivery frequency: approximately 12.39 → 13.65 audio chunks/s.
- Mean chunks/request remained 3.25; first payload remained 39,936 bytes.

The engine emits codec output toward the vocoder at decode-step granularity, while HTTP combines those into much larger audio chunks. Based on the codec’s 21.53 frames/audio-second and measured audio throughput, estimated engine→vocoder codec-frame traffic increased from roughly 318 to 351 messages/s. This is an estimate; the relay itself does not currently expose a direct message counter.

Conclusion: separating FishAudio’s engine and vocoder is clearly beneficial, including substantially better median first-audio latency. Tail first-audio latency remains generation-bound.

### Voxtral TTS

Concurrency/cap: 16.

| Topology | QPS | Mean / p95 latency | Audio throughput | Mean / p95 / peak SM |
| :--- | :--- | :--- | :--- | :--- |
| (pre+gen) \| voc | 13.200 | 1.205 / 2.003 s | 70.942 s/s | 60.6 / 64.0 / 67% |
| pre \| (gen+voc) | 12.567 | 1.262 / 2.096 s | 67.075 s/s | 56.8 / 62.5 / 64% |
| (pre+voc) \| gen | 12.678 | 1.251 / 2.078 s | 68.125 s/s | 60.1 / 67.0 / 74% |
| pre \| gen \| voc | 12.711 | 1.250 / 2.057 s | 67.794 s/s | 59.0 / 64.5 / 69% |

Vocoder-only isolation wins:

- 5.0% faster than preprocessing-only isolation.
- 4.1% faster than preprocessing+vocoder grouping.
- 3.8% faster than all-separate.
- It also has the lowest mean and p95 latency.

### Qwen3 TTS and Ming TTS

| Model | Concurrency | Isolation effect on QPS | Mean latency | p95 latency |
| :--- | :---: | :---: | :---: | :---: |
| Qwen3-TTS 0.6B | 32 | +7.8% | −9.4% | −17.1% |
| Qwen3-TTS 0.6B | 96 | +2.2% | −2.3% | −1.0% |
| Ming TTS | 32 | +54.2% | −35.9% | −35.2% |
| Ming TTS | 96 | +36.8% | −30.7% | −29.1% |

Conclusion: vocoder isolation is highly beneficial for Ming, moderately beneficial for Qwen at concurrency 32, and only marginally beneficial for Qwen at 96.

| Model | Conc. | Mode | QPS | Δ QPS | Mean lat. | Δ mean | Median lat. | p95 lat. | Δ p95 | p99 lat. | RTF mean | Audio s/s | Output tok/s | Completed |
| :--- | :---: | :--- | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
| Qwen3-TTS 0.6B | 32 | Baseline | 4.716 | — | 5.679 s | — | 4.781 s | 12.455 s | — | 16.615 s | 1.4142 | 19.751 | 246.9 | 256/256 |
| Qwen3-TTS 0.6B | 32 | Vocoder isolated | 5.085 | +7.8% | 5.144 s | −9.4% | 4.460 s | 10.323 s | −17.1% | 15.136 s | 1.2575 | 21.405 | 267.6 | 256/256 |
| Qwen3-TTS 0.6B | 96 | Baseline | 4.988 | — | 13.353 s | — | 13.166 s | 20.870 s | — | 31.471 s | 3.2855 | 21.031 | 262.9 | 256/256 |
| Qwen3-TTS 0.6B | 96 | Vocoder isolated | 5.099 | +2.2% | 13.044 s | −2.3% | 13.076 s | 20.666 s | −1.0% | 36.353 s | 3.1358 | 22.217 | 277.7 | 256/256 |
| Ming TTS | 32 | Baseline | 3.072 | — | 9.953 s | — | 9.837 s | 14.822 s | — | 16.705 s | 2.3506 | 13.241 | 41.4 | 256/256 |
| Ming TTS | 32 | audio_decode isolated | 4.737 | +54.2% | 6.380 s | −35.9% | 6.308 s | 9.605 s | −35.2% | 11.086 s | 1.4982 | 20.441 | 63.9 | 256/256 |
| Ming TTS | 96 | Baseline | 3.519 | — | 23.647 s | — | 25.256 s | 34.409 s | — | 40.806 s | 5.5849 | 15.268 | 47.7 | 256/256 |
| Ming TTS | 96 | audio_decode isolated | 4.815 | +36.8% | 16.382 s | −30.7% | 16.934 s | 24.407 s | −29.1% | 27.373 s | 3.8212 | 20.889 | 65.3 | 256/256 |

Additional workload details:

| Model | Conc. | Mode | Audio duration mean | Output tokens mean | Output tokens total | Prompt tokens mean | Prompt tokens total | Output tok/request-s |
| :--- | :---: | :--- | :---: | :---: | :---: | :---: | :---: | :---: |
| Qwen3-TTS 0.6B | 32 | Baseline | 4.188 s | 52 | 13,402 | 58 | 14,757 | 10.3 |
| Qwen3-TTS 0.6B | 32 | Vocoder isolated | 4.209 s | 53 | 13,470 | 58 | 14,757 | 11.1 |
| Qwen3-TTS 0.6B | 96 | Baseline | 4.216 s | 53 | 13,492 | 58 | 14,757 | 4.5 |
| Qwen3-TTS 0.6B | 96 | Vocoder isolated | 4.357 s | 54 | 13,942 | 58 | 14,757 | 4.8 |
| Ming TTS | 32 | Baseline | 4.310 s | 13 | 3,448 | 75 | 19,256 | 1.8 |
| Ming TTS | 32 | audio_decode isolated | 4.315 s | 13 | 3,452 | 75 | 19,256 | 2.6 |
| Ming TTS | 96 | Baseline | 4.339 s | 14 | 3,471 | 75 | 19,256 | 1.4 |
| Ming TTS | 96 | audio_decode isolated | 4.339 s | 14 | 3,471 | 75 | 19,256 | 1.5 |

## Checklist

- [x] Add unit tests.

## CI

if labeled.


sglang_omni/models/ming_tts/config.py
```diff
@@ -61,6 +73,9 @@ def generation_sglang_role_to_stage(cls) -> dict[str, str]:
             factory=f"{_PKG}.stages.create_sglang_tts_engine_executor",
             factory_args={"dtype": "bfloat16"},
             gpu=0,
+            runtime=StageRuntimeConfig(
+                resources=StageResourceConfig(total_gpu_memory_fraction=0.72)
```
JiaxinD: Unlike the other three models, Ming's AR factory does declare `total_gpu_memory_fraction`, so this 0.72 is live: it reaches `SGLangModelRunner._profile_available_bytes` and flips it off upstream free-memory profiling onto the explicit-budget path, changing default KV capacity on every Ming run with no `--isolate-stage`. Separately all four configs move their per-GPU placement total from 0 to 0.92/0.95/1.00, so a config carrying `max_total_gpu_memory_fraction_per_gpu: 0.90` that used to start now fails validation.

sglang_omni/config/process_overrides.py
```diff
@@ -0,0 +1,35 @@
...
+    role_map = type(config).isolation_role_to_stage()
+
+    for requested_name in isolate_stages:
+        stage = stages.get(requested_name)
+        if stage is None:
+            resolved_name = role_map.get(requested_name)
+            stage = stages.get(resolved_name) if resolved_name is not None else None
+        if stage is None:
+            raise ValueError(f"Unknown stage or isolation role: {requested_name}")
+        if stage.tp_size > 1:
```
JiaxinD: Non-TP is not the same as splittable: MOSS-TTS Local hands preprocessing state to the AR stage through a module level `PreparedRequestQueue` (`_QUEUE`), and Qwen3-TTS reads `_PREPROCESSING_CONTEXT`, which its own AR engine builder registers at `qwen3_tts/engine_builder.py:87`. `--isolate-stage preprocessing` passes every check here and the server starts, then every request fails with `preprocessing state is missing` or `context is not initialized`, so the admission test wants a per-model splittable set rather than a TP check.

sglang_omni/config/process_overrides.py
```diff
@@ -0,0 +1,35 @@
...
+            raise ValueError(
+                f"Stage {stage.name!r} already uses one process per TP rank"
+            )
+        stage.process = stage.name
```
JiaxinD: Naming the group after the stage can merge instead of isolate: if another stage already declares `process: <this stage name>` the two land in one component, and `fused_stages` re-merges afterwards, both silently. Asserting that the resolved component contains exactly the requested stage would turn either case into a loud failure.

sglang_omni/models/fishaudio_s2_pro/config.py
```diff
@@ -38,6 +43,9 @@ def generation_sglang_role_to_stage(cls) -> dict[str, str]:
             factory=f"{_PKG}.stages.create_sglang_tts_engine_executor",
             factory_args={"device": "cuda:0", "max_new_tokens": 2048},
             gpu=0,
+            runtime=StageRuntimeConfig(
+                resources=StageResourceConfig(total_gpu_memory_fraction=0.85)
```
JiaxinD: FishAudio's and Voxtral's AR factories never receive `total_gpu_memory_fraction`, so this 0.85 is accounting only: a `stage_overrides` entry setting `runtime.sglang_server_args.mem_fraction_static: 0.99` still passes the 0.95 budget while the engine actually takes 0.99. Worth saying in the doc that the guard validates a declaration, not the runtime value.

JiaxinD (commented):
The override itself is clean: the deep copy holds, alias resolution is deterministic with Ming's `vocoder -> audio_decode` pinned against the real config, TP stages are rejected with a useful message, and with the flag omitted the process groups really are byte-identical to main. Two structural things though: `non-TP` is the wrong admission test for `splittable`, and the resource declarations that let same-GPU isolation validate are not themselves behind the flag. CI has not run here either, `omni-ci-gate` fails on the missing `run-ci` label.

sglang_omni/config/process_overrides.py
```diff
@@ -0,0 +1,69 @@
...
+                resources.total_gpu_memory_fraction = memory_fraction
+        stage.process = stage.name
+        isolated_stage_names.append(stage.name)
+
+    from sglang_omni.config.placement import build_stage_placement_plan
+    from sglang_omni.config.topology import build_process_topology_plan
+
+    topology = build_process_topology_plan(
+        config,
+        build_stage_placement_plan(config),
```
JiaxinD: `build_stage_placement_plan` defaults to `apply_policy=True`, so this imports and runs the user's `placement_policy` in the middle of the CLI chain, before cuda-graph / decode-mode / torch-compile / partial-start are applied, and again later in `prepare_pipeline_runtime`. Passing `apply_policy=False` keeps the singleton check without moving policy execution.

sglang_omni/config/process_overrides.py
```diff
@@ -0,0 +1,69 @@
...
+        ].items():
+            resource_stage = stages.get(resource_stage_name)
+            if resource_stage is None:
+                raise ValueError(
+                    f"Isolation resources for stage {stage.name!r} reference "
+                    f"unknown stage {resource_stage_name!r}"
+                )
+            resources = resource_stage.runtime.resources
+            if resources.total_gpu_memory_fraction is None:
+                resources.total_gpu_memory_fraction = memory_fraction
```
JiaxinD: Attribute assignment skips `StageResourceConfig`'s validator: constructing it with `-0.5` raises, but a contract value of `-0.5` lands here and yields an accepted per-GPU total of `-0.4`, since the guard only checks the upper bound. The `is None` guard also lets a YAML `tts_engine: 0.01` plus `--isolate-stage vocoder` validate at `0.11` while the FishAudio builder still runs `mem_fraction_static: 0.85`.

sglang_omni/models/moss_tts_local/config.py
```diff
@@ -119,6 +120,16 @@ def talker_sglang_role_to_stage(cls) -> dict[str, str]:
     def generation_sglang_role_to_stage(cls) -> dict[str, str]:
         return {"generation": "tts_engine"}
 
+    @classmethod
+    def isolation_stage_resources(cls) -> dict[str, dict[str, float]]:
```
JiaxinD: The allow-list here is `vocoder` only, so the body's `Preprocessing only +19.3%` and `Both +33.6%` rows describe configurations this head rejects (I ran both, they raise `does not support process isolation`), and they could not have completed before either, since isolated preprocessing publishes into a process-local `_QUEUE` that the AR process cannot pop. Worth replacing them with supported-path numbers.

tests/unit_test/pipeline/test_topology.py
```diff
@@ -61,6 +85,531 @@ def test_stage_process_parses_from_schema_and_dotted_overrides() -> None:
...
+    config = MossTTSLocalSplitPipelineConfig(model_path="dummy")
+
+    with pytest.raises(
+        ValueError,
+        match="Stage 'vocoder' does not support process isolation",
+    ):
+        apply_stage_process_overrides(config, isolate_stages=["vocoder"])
+
+
+def test_qwen3_tts_rejects_vocoder_isolation_without_memory_contract() -> None:
```
JiaxinD: qwen3_tts is pinned as rejected here while the PR body carries its vocoder-isolated benchmark (+7.8% at c32, +2.2% at c96), so the evidence and the capability disagree. Same gap for moss_tts delay, zonos2, audar_tts, and higgs's `audio_encoder`, all of which still share `pipeline` with the AR stage.

JiaxinD (commented):
Re-reviewed 094c8698 against the three earlier findings and all three are genuinely fixed: the four model `stages` lists are now byte-identical to main so no default path receives a fraction and Ming's default KV sizing is back on the upstream profiling branch, the allow-list closes the process-local-handoff hole (I ran `--isolate-stage preprocessing` on MOSS and `tts_engine` on Ming, both rejected), and the singleton-group assertion covers both the process-name collision and the fused-stage case with asserted messages. 37 topology tests pass locally.

Four comments below: three small implementation items (the placement-policy re-entry, the unvalidated contract assignment, and the MOSS benchmark rows) plus one coverage gap. On coverage, five models still keep a vocoder-class stage in `pipeline` with the AR stage and are absent from the allow-list: qwen3_tts, higgs (`audio_encoder`), moss_tts delay, zonos2, and audar_tts. higgs is the cheapest to add since its three GPU stages already declare fractions (0.03 + 0.85 + 0.10). I could not construct ming_omni locally (missing msgspec), so please confirm that one yourself.

CI still has not run here, `omni-ci-gate` fails on the missing `run-ci` label.

tests/unit_test/pipeline/test_topology.py
AkazaAkane: changed, default to not isolate

sglang_omni/models/moss_tts_local/config.py
AkazaAkane: changed, default to not isolate

sglang_omni/config/process_overrides.py
```diff
@@ -0,0 +1,253 @@
...
+    role_map = type(config).isolation_role_to_stage()
+    resource_contracts = type(config).process_edge_resources()
+    baseline_processes = _declared_process_map(config)
+
+    requested: dict[str, str] = {}
+    for assignment in stage_processes or []:
+        requested_name, process_name = parse_stage_process_assignment(assignment)
+        stage = _resolve_stage(stages, role_map, requested_name)
+        _reject_tp_stage(stage)
+        requested[stage.name] = process_name
```
JiaxinD, line 59: A repeated `--stage-process` for the same resolved stage silently takes the last value: `preprocessing=tts_frontend audio_encoder=tts_frontend preprocessing=other` puts preprocessing in `other` and quietly breaks the grouping (executed on this head). Ming's `vocoder=g1` plus `audio_decode=g2` hits the same thing through the alias, so a duplicate-resolved-stage check would match the cross-flag rejection just below.

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -80,6 +80,14 @@
 _REF_WAVEFORM_CACHE_MAX_BYTES = 512 * 1024 * 1024
 _VOCODER_COMPILE_WARMUP_FRAME_COUNTS = (1, 8)
 
+# note (kaige li): preprocessing folds these into HiggsTtsState and nothing
```
JiaxinD, line 83: This block is #1159's `a5531e2d` almost verbatim, `note (kaige li)` included, and the two tests added below cover the same behavior its three tests do. The two heads still auto-merge, so whoever lands second just leaves duplicate coverage; worth agreeing with @stephenkgli on who carries it and dropping it from one side.

sglang_omni/config/process_overrides.py
```diff
@@ -0,0 +1,253 @@
...
+            edge, {}
+        ).items():
+            resource_stage = stages.get(resource_stage_name)
+            if resource_stage is None:
+                raise ValueError(
+                    f"Process-edge resources for {edge!r} reference unknown stage "
+                    f"{resource_stage_name!r}"
+                )
+            resources = resource_stage.runtime.resources
+            if resources.total_gpu_memory_fraction is None:
```
JiaxinD: The out-of-range hole is closed, but an understated explicit value still defeats the contract: FishAudio with `tts_engine.runtime.resources.total_gpu_memory_fraction: 0.01` plus `--isolate-stage vocoder` validates at a GPU total of 0.11 while the builder still runs `mem_fraction_static: 0.85`. Preserving an operator value is reasonable, but warning when it sits below the contract the model declares for that edge would keep the accounting honest.

sglang_omni/config/process_overrides.py
```diff
@@ -0,0 +1,253 @@
...
+        return False
+    return sum(1 for name in process_map.values() if name == process_name) == 1
+
+
+def _apply_process_edge_resources(
+    stages: dict[str, StageConfig],
+    new_cross_process_edges: set[tuple[str, str]],
+    resource_contracts: dict[tuple[str, str], dict[str, float]],
+) -> None:
+    for edge in sorted(new_cross_process_edges):
```
JiaxinD, line 146: When one split crosses two contracted edges, a stage named by both takes the first value in sorted edge order with no conflict check. No shipped model has overlapping contracts today, so this is about the mechanism the next model will extend rather than a live bug.

JiaxinD (commented):
Moving the contract from stages to edges is the right call, the hazard really is per handoff. I audited all 15 edges the nine models now declare safe and every one carries its state in the payload, and the three process-local-global handoffs (MOSS Local and MOSS delay `_QUEUE`, Qwen3-TTS `_PREPROCESSING_CONTEXT`) are correctly excluded; I ran `--isolate-stage preprocessing` on MOSS and on Qwen3-TTS and both are rejected with the edge-specific message. `apply_policy=False` and the `model_validate` fix are both in with a regression test, all nine default stage lists are still byte-identical to main, and 59 topology tests pass locally. `--stage-process a=G b=G` also reproduces #1159's exact Higgs `tts_frontend` topology, which closes the multi-stage-group gap from the last round.

Four comments below. Two more status notes: MOSS vocoder isolation lands at a per-GPU total of exactly 1.00, leaving nothing for the second CUDA context (measured at +1.5 to 1.9 GB elsewhere in this repo), so a peak-VRAM number from a real isolated run would be worth having. And nothing here exercises a request across a newly enabled edge in two real processes, so the whole safety argument is still code reading; CI has also never run (`omni-ci-gate` fails on the missing `run-ci` label). I can run that smoke myself if it helps.

JiaxinD:
Tried the runtime smoke myself and my box blocked it: MOSS-TTS-Local dies at startup with SIGKILL in my dev container both with and without `--isolate-stage` (Higgs the same), so nothing there is attributable to this PR. The Higgs attempt did reach real process spawning first, `tts_frontend` / `pipeline` / `vocoder` as three groups, so `--stage-process` gets further than the config-level tests show.

Since you have a working bench, could you run one request end to end across a newly enabled edge, say `--isolate-stage vocoder` on MOSS-TTS-Local, and post its peak VRAM? That is the one thing the 59 topology tests cannot show, and MOSS lands at a declared per-GPU total of exactly 1.00, so the peak number would also settle whether the extra CUDA context fits.

sglang_omni/models/higgs_tts/stages.py
stephenkgli: It seems that this PR covers what #1159 did, so I will close #1159, @AkazaAkane could you add performance improvement of higgs-tts into this PR?

sglang_omni/config/process_overrides.py
```diff
@@ -0,0 +1,253 @@
...
+        if _runs_alone(baseline_processes, stage.name):
+            continue
+        requested[stage.name] = stage.name
+        singleton_stage_names.append(stage.name)
+
+    if not requested:
+        return config
+
+    for stage_name, process_name in requested.items():
+        stages[stage_name].process = process_name
```
stephenkgli, line 82: `--stage-process audio_encoder=vocoder` will make the `audio_encoder` and `vocoder` of higgs-tts in the same process, but comments below:

https://github.com/sgl-project/sglang-omni/blob/765b4944b092290f74506152129a638250c85654/sglang_omni/models/higgs_tts/utils.py#L32-L33

says that these two stages should in different process. So you may remove the comments or add more check here.

docs/basic_usage/tts_process_topology.md
```diff
@@ -0,0 +1,131 @@
...
+  --model-path bosonai/higgs-tts-3-4b \
+  --stage-process preprocessing=tts_frontend \
+  --stage-process audio_encoder=tts_frontend
+```
+
+```text
+tts_frontend : preprocessing, audio_encoder
+pipeline     : tts_engine
+vocoder      : vocoder
+```
```
stephenkgli, line 51: `audio_encoder=tts_frontend` (STAGE=PROCESS) really confuse me, but the text below: `tts_frontend:preprocessing,audio_encoder` (PROCESS:STAGE_1,STAGE_2) seems much clearer to me. I’m not very familiar with command-line argument naming conventions, so my intuition may be wrong. Perhaps you could come up with a better design. cc @JiaxinD

stephenkgli (commented):
Great job! Left a few comments

sglang_omni/config/process_overrides.py
AkazaAkane: Addressed in `53c680fa`. Repeated `--stage-process` assignments are now rejected after role resolution, so both repeated literal names and the Ming `vocoder` / `audio_decode` alias pair fail loudly instead of silently taking the last value. Regression coverage was added for both cases. Pre-commit passes; unit, runtime, and VM tests have not been run yet.

sglang_omni/config/process_overrides.py
AkazaAkane: Addressed in `53c680fa`. Edge recommendations are collected before applying them, and different recommendations for the same stage now raise an error that identifies both edges and values. Identical recommendations remain valid. Regression coverage was added. Pre-commit passes; unit, runtime, and VM tests have not been run yet.

sglang_omni/config/process_overrides.py
AkazaAkane: Addressed in `53c680fa`. The old comment described the default topology as if it were an invariant. The cache is process-scoped: colocated stages share the cached codec instance, while stages assigned to separate processes load one copy each. I updated the comment rather than prohibiting this valid grouping. Pre-commit passes; runtime and VM tests have not been run yet.

docs/basic_usage/tts_process_topology.md
AkazaAkane: I kept `STAGE=PROCESS` because each repeated option is one assignment: place this stage in this process. It composes directly with stage aliases without adding comma-list parsing. In `53c680fa` I clarified both the CLI help and documentation to read it explicitly left to right: place `STAGE` in `PROCESS`. Pre-commit passes; unit, runtime, and VM tests have not been run yet.

sglang_omni/models/higgs_tts/stages.py
AkazaAkane: Thanks. Since #1159 is being closed in favor of this PR, this PR retains the Higgs implementation and coverage. The Higgs performance and end-to-end VM results are still pending; I will post them after those runs rather than claim unverified numbers here.

sglang_omni/config/process_overrides.py
AkazaAkane: Addressed in `53c680fa`. Explicit operator fractions remain authoritative, but the override now warns when a value is below the model-declared contract for the crossed edge, including the stage, edge, explicit value, and recommended value. The FishAudio `0.01` case has regression coverage and the explicit value is preserved.

AkazaAkane:
## Validation report

Tested PR #1125 at commit `5be6df4d` on an NVIDIA H100 80GB HBM3.

### Environment and methodology

- GPU device: physical GPU 3
- Reported capacity: **81,559 MiB**
- All server processes were pinned with `CUDA_VISIBLE_DEVICES=3`.
- The GPU was empty before each run.
- Activity on GPUs 0–2 from other workloads was excluded.
- VRAM usage was sampled once per second during server startup and request execution.

### Results summary

| Test | Topology | Result | Artifact |
|---|---|---|---|
| Focused topology suite | N/A | **62 passed in 10.41s** | N/A |
| MOSS-TTS baseline | 3 stages, 1 process | HTTP 200; **69,829 MiB** peak | 48 kHz WAV |
| MOSS-TTS isolated vocoder | 3 stages, 2 processes | HTTP 200; **70,711 MiB** peak | 48 kHz WAV |
| Higgs grouped frontend | 4 stages, 3 processes | HTTP 200 | 24 kHz WAV |
| Higgs separated frontend | 4 stages, 4 processes | HTTP 200 | 24 kHz WAV |
| Qwen isolated vocoder | 3 stages, 2 processes | HTTP 200 | 24 kHz WAV |

## 1. Focused regression suite

Command:

```bash
export PYTHONPATH="$PWD"
python -m pytest tests/unit_test/pipeline/test_topology.py -v -s -x
```

Result:

```text
62 passed in 10.41s
```

This includes the newer duplicate-stage-assignment and conflicting-resource-contract regressions.

## 2. MOSS-TTS Local baseline

Configuration:

```yaml
process_groups:
  pipeline:
    stages: [preprocessing, tts_engine, vocoder]
    gpu: 0
```

- Processes: **1**
- Declared GPU fraction: **0.90**

Runtime evidence:

```text
StageGroup pipeline: spawned 1 process
Process pipeline ready with stages=['preprocessing', 'tts_engine', 'vocoder']
MultiProcessPipelineRunner started: 3 stage(s), 1 process(es)
POST /v1/audio/speech HTTP/1.1 200 OK
```

Results:

- Peak VRAM: **69,829 MiB**
- Remaining memory at peak: **11,730 MiB**
- WAV sample rate: **48,000 Hz**
- Frames: **188,160**
- Duration: **3.92 seconds**

## 3. MOSS-TTS Local with isolated vocoder

Configuration:

```yaml
process_groups:
  pipeline:
    stages: [preprocessing, tts_engine]
    gpu: 0
  vocoder:
    stages: [vocoder]
    gpu: 0
```

- Processes: **2**
- Stage GPU fractions:
  - `preprocessing`: **0.05**
  - `tts_engine`: **0.90**
  - `vocoder`: **0.05**
  - Total: **1.00**

Runtime evidence:

```text
StageGroup pipeline: spawned 1 process
StageGroup vocoder: spawned 1 process
Process pipeline ready with stages=['preprocessing', 'tts_engine']
Process vocoder ready with stages=['vocoder']
MultiProcessPipelineRunner started: 3 stage(s), 2 process(es)
POST /v1/audio/speech HTTP/1.1 200 OK
```

Results:

- Peak VRAM: **70,711 MiB**
- Remaining memory at peak: **10,848 MiB**
- Increase from baseline: **882 MiB**, approximately **1.26%**
- WAV sample rate: **48,000 Hz**
- Frames: **172,800**
- Duration: **3.60 seconds**

This verifies that:

- The newly separated `tts_engine → vocoder` edge works over IPC.
- The vocoder runs in its own process and CUDA context.
- A declared GPU-fraction total of exactly **1.00** completes without OOM.
- The additional process used **882 MiB** more VRAM in this run.
- The isolated run retained more than **10 GiB** of physical headroom.

The WAV lengths differ because generation is stochastic. Both artifacts decoded successfully.

## 4. Higgs grouped frontend

Topology:

```text
tts_frontend: [preprocessing, audio_encoder]
pipeline:     [tts_engine]
vocoder:      [vocoder]
```

- Stages: **4**
- Processes: **3**
- Declared GPU-fraction total: **0.98**

A real HTTP reference clip was downloaded and encoded.

Runtime evidence:

```text
Higgs ref cache reference encode stats:
hits=0, misses=1, entries=1, bytes=3360, failed=0
POST /v1/audio/speech HTTP/1.1 200 OK
```

Artifact:

- WAV sample rate: **24,000 Hz**
- Frames: **88,320**
- Duration: **3.68 seconds**
- WAV: `sglang-pr-1125/higgs-grouped.wav`
- Log: `sglang-pr-1125/higgs-grouped.log:65`

This verifies that `--stage-process` works beyond topology construction, including reference-audio handling and frontend-to-engine IPC.

## 5. Higgs fully separated frontend

Topology:

```text
preprocessing: [preprocessing]  # CPU
audio_encoder: [audio_encoder]
pipeline:      [tts_engine]
vocoder:       [vocoder]
```

- Stages: **4**
- Processes: **4**
- Declared GPU-fraction total: **0.98**

Runtime evidence:

```text
Process preprocessing ready
Process audio_encoder ready
Process pipeline ready
Process vocoder ready
MultiProcessPipelineRunner started: 4 stage(s), 4 process(es)
Higgs ref cache: misses=1, entries=1, failed=0
POST /v1/audio/speech HTTP/1.1 200 OK
```

Artifact:

- WAV sample rate: **24,000 Hz**
- Frames: **88,320**
- Duration: **3.68 seconds**
- WAV: `sglang-pr-1125/higgs-separated.wav`
- Log: `sglang-pr-1125/higgs-separated.log:68`

This closes the functional Higgs VM-testing gap. It does not address the separate [performance-testing request](https://github.com/sgl-project/sglang-omni/pull/1125#discussion_r3666199924); a performance conclusion would require repeated latency and throughput measurements against the default topology.

## 6. Qwen vocoder-isolation smoke test

Topology:

```text
pipeline: [preprocessing, tts_engine]
vocoder:  [vocoder]
```

- Stages: **3**
- Processes: **2**
- Stage GPU fractions:
  - `tts_engine`: **0.85**
  - `vocoder`: **0.10**
  - Total: **0.95**

Runtime evidence:

```text
Process vocoder ready with stages=['vocoder']
Process pipeline ready with stages=['preprocessing', 'tts_engine']
MultiProcessPipelineRunner started: 3 stage(s), 2 process(es)
Qwen3-TTS reference encode stats:
  misses=1, entries=1, bytes=8320, failed=0
POST /v1/audio/speech HTTP/1.1 200 OK
```

Artifact:

- WAV sample rate: **24,000 Hz**
- Frames: **49,920**
- Duration: **2.08 seconds**
- WAV: `sglang-pr-1125/qwen-isolated.wav`
- Log: `sglang-pr-1125/qwen-isolated.log:210`

This directly verifies the Qwen `tts_engine → vocoder` resource-contract edge using a real reference-cloning request.

The VM initially lacked the Python `sox` package and had Transformers 5.6.0, while `qwen-tts==0.1.1` requires Transformers 4.57.3. The successful run used the documented Qwen dependency versions in the local virtual environment.

## Conclusion

The focused regression suite passed, and the requested isolation paths completed end to end for MOSS-TTS Local, Higgs, and Qwen.

The strongest resource-accounting result is the MOSS comparison:

- Baseline peak: **69,829 MiB**
- Isolated-vocoder peak: **70,711 MiB**
- Observed isolation overhead: **882 MiB**
- H100 capacity: **81,559 MiB**
- Remaining headroom at isolated peak: **10,848 MiB**

The Higgs tests establish functional correctness only; no Higgs performance conclusion is claimed.

AkazaAkane:
FYI @stephenkgli 
| Isolation strategy | QPS | Mean / P95 latency | Output tok/s | Peak VRAM |
|---|---|---|---|---|
| Default frontend in AR process | 28.934 | 3.372 / 4.475 s | 3,280 | 72,057 MiB |
| Grouped frontend process | 33.334 | 2.926 / 4.093 s | 3,760 | 73,373 MiB |
| Fully separated frontends | 31.495 | 3.105 / 4.268 s | 3,540 | 73,537 MiB |

Each result is the mean of two 90-second windows at concurrency/cap 96. All six windows had zero failures and passed the QPS × mean latency ≈ 96 check.

Key conclusions:

- Grouped preprocessing+audio_encoder improved QPS by 15.21% and reduced mean latency by 13.23% versus default.

stephenkgli (approved):
LGTM

Issue #1026, opened 2026-07-11
[Qwen3-Omni Perf] Code2wav ragged audio batching


edwingao28:
## Problem

Code2wav is the downstream bottleneck of the speech pipeline. Discovery instrumentation at the vocoder dispatch point shows the inbox backlog growing monotonically (max `484` at c8, `918` at c16) with ready-queue depth p50 = `8` — far above the batching threshold of 2. Meanwhile each emit decodes one item; xRT degrades `8.14 -> 45.86` from c1 to c16, and stage residency at c8 puts code2wav at `1.27 s` per request.

## Design status

Discovery run complete (2026-07-10) with an early promote signal (ready-depth far above threshold). The write-up with the ready-depth/backlog traces is pending and lands here first.

#940 (BatchVocoderBase, non-streaming) already landed under the #985 TTS refactor, and #936/#939 define the streaming surfaces — this issue adopts those surfaces rather than forking a Qwen3-Omni-local batching path, unless the write-up shows they cannot fit.

## Open decisions

- [ ] Build on the #940 base vs. a code2wav-local batcher (default: adopt #985 surfaces).
- [ ] Batch policy: fixed size vs. adaptive to ready-depth.
- [ ] TTFA guard: batching must not delay the first audio chunk beyond the contract's regression bound.

## Task checklist

- [x] Post the discovery write-up: dispatch-point queue traces at c `{4,8,16}` (posted below)
- [ ] Batch-size sweep at the dispatch point; assess fit with the #985 streaming surface (#936/#939)
- [ ] E2E paired A/B: xRT and TTFA at c `{8,16}` per #1018's contract; audio-touching change, so waveform comparison + WER-flat evidence required

Part of #1022. Design and benchmark contract: #1018 (item 4.2).

edwingao28:
Discovery write-up (instrumented run at `45582ff5`, vocoder dispatch point, c{4,8,16}, all requests succeeded; raw data in `discovery_runs/B4/` on the bench host).

Queue state sampled at each dispatch (1012 records):

| c | dispatches | inbox chunks p50/p95/max | distinct reqs with backlog p50 | ready at dispatch p50 |
|---:|---:|---|---:|---:|
| 4 | 132 | 39.5 / 202 / 280 | 4 | 1 |
| 8 | 264 | 118 / 401 / 484 | 8 | 1 |
| 16 | 528 | 321 / 663 / 918 | 16 | 1 |

[...]

PR #1126, opened 2026-07-21
[Qwen3-Omni Perf] Bounded wait/floor batching for Code2Wav vocoder

edwingao28:
## Motivation

Code2Wav is the downstream bottleneck of the Qwen3-Omni speech pipeline: at concurrency 8 its stage residency is 1.27 s per request versus 0.48 s for the Thinker (#1018). Each request's code stream is vocoded on its own, one window at a time, and the vocoder is under-utilized at batch size 1, so under concurrent load the stage leaves throughput unclaimed. This PR adds a bounded wait/floor cross-request coalescer for Code2Wav (roadmap #1018 Track 4.2, issue #1026). It is opt-in and off by default; with batching disabled the path is the current per-request decode.

## Modifications

- `code2wav_scheduler.py`: migrate `Code2WavScheduler` from `StreamingSimpleScheduler` onto `StreamingVocoderBase`, and implement the bounded batch-former in the base coalescing hooks (`select_step_participants` / `build_step_plan` / `run_step`). The max-wait timer is a deadline-bounded inbox `get`, not a sleep. Odd participant counts decompose into `{8, 4, 2, 1}` sub-batches. Four opt-in flags, all defaulting to current behavior: `enable_batching`, `max_batch_wait_ms`, `batch_floor`, `batch_ceiling`.
- `streaming_vocoder.py`: the base rejected a chunk unless `metadata["stream"] is True`, but the Talker latches the client streaming flag, which is legitimately `False` for a non-streaming client. Relaxed to accept any `bool`.
- `config.py`: thread the four flags through the `code2wav` stage `factory_args`; a top-level `runtime_overrides.code2wav` block overrides them per profile.
- Tests: new `test_code2wav_batching.py` (fire-rule state machine, sub-batch decomposition, batched-vs-per-request equivalence, lifecycle), plus parity adaptations to the migrated scheduler.

Scope: this PR is eager-only. Exact-shape CUDA Graph replay for Code2Wav is #1101 (Track 4.1), and the CUDA-Graph-with-batching composition is Track 4.4 — both are out of scope here. #1101 and this PR both modify `code2wav_scheduler.py`, so whichever lands second will rebase.

## Benchmarking and Profiling

Real Qwen3-Omni Code2Wav model on one H100, single-process scheduler A/B. Control is true per-request decode (batch size 1); candidate is the coalescing pump. Streams are fed with aligned windows so coalescing engages, 20 windows each.

| Concurrent streams | Control | Batched | Speedup | Attained batch |
|---|---|---|---|---|
| 4  | 551 ms  | 228 ms  | 2.42× | 4 |
| 8  | 1100 ms | 286 ms  | 3.84× | 8 |
| 16 | 2188 ms | 539 ms  | 4.06× | 8 |
| 32 | 4532 ms | 1128 ms | 4.02× | 8 |

The raw vocoder forward alone (eager, 30 frames, 16 codebooks) batches 1 to 8 at 2.49×; the scheduler A/B beats that because the per-request control path also pays dispatch overhead per stream (160 forward calls versus 20).

Where it does not help: a single-stream end-to-end speech benchmark at concurrency 1 through 8 shows no change within run-to-run noise. At single-stream serving the windows rarely align inside the wait, so coalescing seldom fires, and Code2Wav is a small fraction of the Thinker- and Talker-dominated end-to-end latency. The gain is a throughput lever under aligned high-concurrency Code2Wav load, not a latency win, which is why it stays off by default.

## Open items

- [ ] The batched output is bit-identical to per-request decode only on the deterministic test model. On the real bf16 vocoder it diverges up to ~9e-2 peak on ~0.2% of samples; in fp32 that collapses to ~2e-3 with no sample over 1e-2, confirming benign cuDNN batch-kernel numerics rather than a logic bug. For this audio-touching change the acceptance gate should be a waveform tolerance (SNR or max-diff bound), not bitwise equality.
- [ ] Sweep `max_batch_wait_ms` and `batch_floor` on a Code2Wav-bound high-concurrency workload per #1026, then gate the default on the result.

## Checklist

- [x] Follow the SGLang code style guidance.
- [x] Unit tests added for the batch-former, sub-batch decomposition, equivalence, and lifecycle.


sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -61,212 +69,343 @@ def __init__(
...
+            for _, state in self._stream_state_items()
+            if state.due_since is not None
+        ]
+        if not due:
+            return None
+        return min(due) + self._max_batch_wait_s
+
+    def _collect_stream_chunk_batch(self, first_msg):
+        batch = super()._collect_stream_chunk_batch(first_msg)
+        deadline = self._batch_deadline()
```
SandyLuXY, line 313: This evaluates the old deadline before the collected chunks are ingested, and the idle `start()` loop never re-pumps when that deadline expires. On this head a ready window stayed buffered for 205 ms with `max_batch_wait_ms=50`, while an older 300 ms deadline delayed another request's first window by 280 ms; please wake/re-pump at the earliest deadline, re-check first/floor after ingestion, and add threaded loop regressions.

tests/unit_test/qwen3_omni/test_code2wav_batching.py
SandyLuXY, line 1: This new suite is missing from `tests/README.md`. Please register and doc `test_code2wav_batching.py`.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -61,212 +69,343 @@ def __init__(
...
-                int(audio.shape[0]),
+        state.emitted = end
+        state.due_since = None
+        if audio.size == 0:
+            return None
+        if not state.audio_parts:
+            _emit_event(
+                request_id=request_id,
+                stage=None,
+                event_name="code2wav_first_audio",
```
SandyLuXY, line 239: This migration removes `_decode_and_emit`, but `docs/developer_reference/profiler.md` still names it as the `code2wav_first_audio` source and does not document the new batch start/end events. Please update that table to point to `decode_delta()` / `run_step()` and describe the batch-size, fire-reason, wait, shape, and execution metadata. Also update the related comment in the file.

sglang_omni/models/qwen3_omni/config.py
```diff
@@ -202,7 +202,13 @@ def _code2wav_stage(*, gpu: int, process: str) -> StageConfig:
         name="code2wav",
         process=process,
         factory=f"{_PKG}.components.code2wav_scheduler.create_code2wav_scheduler",
-        factory_args={"device": "cuda"},
+        factory_args={
+            "device": "cuda",
+            "enable_batching": False,
```
SandyLuXY: Please report xRT/throughput and TTFA p50/p95, per-forward batch sizes etc from the benchmarking runs in the PR body. I think we need to validate the setting passes the roadmap gate before exposing the feature in deployment config.

SandyLuXY (dismissed):

edwingao28:
Addressed review in `51d65f4d`.

### H100 E2E — BF16 colocated, zero-wait batching (floor 2)

| Concurrency | Req/s vs control | xRT vs control | TTFA p50 | TTFA p95 |
|---:|---:|---:|---:|---:|
| 1 | 1.115 (-0.4%) | 7.68 (-0.4%) | 266 ms (-1.3%) | 279 ms (+1.5%) |
| 8 | 3.035 (-2.3%) | 21.19 (-2.1%) | 1015 ms (+11.2%) | 1893 ms (+8.5%) |
| 16 | 3.626 (+16.9%) | 26.04 (+17.0%) | 2253 ms (-22.9%) | 4009 ms (-10.6%) |

All 96/96 measured requests succeeded. At c16, 339 Code2Wav forwards had mean batch 1.19 and distribution `{1:291, 2:31, 3:16, 4:1}`.

A real 10 ms wait regressed throughput 21–33%, so I rejected that setting. Recommendation: keep batching opt-in/default-off; zero-wait helps saturated c16 but is workload-dependent.

Validation: 2,348 unit tests passed.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
edwingao28: Fixed in 51d65f4d: ingest is now nonblocking before deadline/floor evaluation, the inbox wait tracks the earliest batch deadline, and idle schedulers pump due streams. Added threaded regressions for both deadline wakeup and cross-request delay. 2,348 tests pass.

tests/unit_test/qwen3_omni/test_code2wav_batching.py
edwingao28: Fixed in 51d65f4d: registered the Code2Wav scheduler suite in tests/README.md.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
edwingao28: Fixed in 51d65f4d: updated the profiler source path and documented code2wav_batch_start/code2wav_batch_end metadata.

sglang_omni/models/qwen3_omni/config.py
edwingao28: Posted corrected H100 E2E above. Zero-wait batching is +16.9% req/s at c16 but neutral/slightly negative at c1/c8; a 10 ms wait regresses throughput 21–33%. Per-forward c16 distribution is {1:291, 2:31, 3:16, 4:1}. Recommendation: keep default off.

yxs (approved):

PR #1128, opened 2026-07-21
[Feature] Enable calibrated FP8 KV cache for MOSS-Transcribe-Diarize

JiaxinD:
## Motivation

MOSS-Transcribe-Diarize decode is KV-read-bound: contexts run ~33k tokens per request, raising the admission cap 16 -> 64 leaves throughput flat, and the per-step KV traffic scales with batch size (profiling in #921). The direct lever is cutting KV bytes, but `kv_cache_dtype: fp8_e4m3` is unusable for this model today, twice over:

1. sglang's model runner calls `load_kv_cache_scales` on the entry class when `quantization_param_path` is set; the MTD entry class lacks the method, so the server fails at load.
2. Without calibrated scales the e4m3 cast produces NaN, not saturation (measured K amax reaches 584 vs the e4m3 max of 448), and output degenerates to runaway garbage (CER 5.3 -> 1800+).

## Modifications

Add `load_kv_cache_scales` to `MossTranscribeDiarizeForConditionalGeneration`: delegate to the Qwen3 decoder (same pattern as the moss_tts entry classes), then convert the scales the upstream JSON loader assigns as Python floats into 0-d float32 tensors, because the FA3 decode path calls `.expand()` on them and crashes on floats.

Unit tests cover the delegation, the float-to-tensor conversion, and tensor pass-through.

## Usage

Record per-layer K/V amax over a short bf16 calibration run, write the vLLM-legacy scales JSON (`model_type: qwen3`, 28 layers, `scale = amax / 448 * margin`), then launch with:

```yaml
runtime_overrides:
  asr:
    server_args_overrides:
      kv_cache_dtype: fp8_e4m3
      quantization_param_path: /path/kv_scales.json
```

## Evidence (1x H100, movies800times)

* Startup gates [measured]: FA3 backend retained (`fp8_e5m2` silently demotes to triton, do not use it), `Loaded KV cache scaling factors from ...` logged.
* Quality preserved [measured]: cp_cer 13.42-13.66 across four 256-sample fp8 runs (c8 and c32) vs 13.51-13.77 for bf16 on the same sets; a same-100-sample paired smoke gives 15.44 (fp8) vs 16.15 (bf16). Uncalibrated fp8_e4m3 for contrast: cp_cer 1800+.
* Throughput: no claim in this PR. The KV pool doubles in capacity and per-step KV reads halve in bytes, but our A/B window was contention-noisy (within-arm spread ~2x from co-located jobs); a quiet-window ABAB will follow as a comment.

## Checklist

- [x] Format your code with `pre-commit run --all-files`
- [x] Add unit tests

### DIFF
--- a/sglang_omni/models/moss_transcribe_diarize/sglang_model.py
+++ b/sglang_omni/models/moss_transcribe_diarize/sglang_model.py
@@ -417,5 +417,27 @@ class MossTranscribeDiarizeForConditionalGeneration(nn.Module):
                     ),
                 )
 
+    def load_kv_cache_scales(self, quantization_param_path: str) -> None:
+        """Delegate FP8 KV scale loading to the Qwen3 decoder.
+
+        Note: (Jiaxin Deng) the upstream JSON loader assigns Python floats,
+        but the FA3 decode path calls .expand() on the scales; convert to
+        0-d tensors so fp8_e4m3 KV works instead of crashing at decode.
+        """
+        self.language_model.load_kv_cache_scales(quantization_param_path)
+        device = next(self.language_model.parameters()).device
+        for layer in self.language_model.model.layers:
+            attn = layer.self_attn.attn
+            if isinstance(attn.k_scale, float):
+                attn.k_scale_float = attn.k_scale
+                attn.k_scale = torch.tensor(
+                    attn.k_scale, dtype=torch.float32, device=device
+                )
+            if isinstance(attn.v_scale, float):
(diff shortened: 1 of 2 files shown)


JiaxinD:
Throughput A/B attempted twice on the shared box; within-arm spread (2x+ on identical configs) exceeds any effect, so still no throughput claim either way. Quality gate is now 8 for 8 runs: cp_cer 13.42-13.75 (fp8) vs 13.51-13.77 (bf16) across c8 and c32.

Issue #1114, opened 2026-07-21
[Ming-Omni] Speech path fails on main: missing deps, hardcoded prompt-wav path in DB30 preset, talker arange() error


JiaxinD:
Attempted a Ming-Omni speech-path benchmark on current main (a24654fa era) with a full 5-GPU window (thinker TP4 + dedicated talker). Text-side health comes up; the speech path fails through three independent layers, each reproducible:

**1. Missing runtime deps.** First speech request 500s with `No module named 'x_transformers'` (pip also flags `websockets>=12.0` absent for the sglang-omni extra). Installing both moves past it.

**2. Default voice preset hardcodes an author-env path.** The checkpoint's `talker/data/voice_name.json` for the default voice `DB30` contains `"prompt_wav_path": "/workspace/sglang-omni-pr1000-head/tests/data/query_to_cars.wav"`. On any machine without that exact path, every speech request 500s with a missing-prompt-wav error. The wav itself ships in this repo (`tests/data/query_to_cars.wav`), so the preset should resolve it relative to the repo/checkout, not an absolute authoring path (a symlink works around it).

**3. Talker runtime error.** With 1 and 2 fixed, speech requests 500 with `arange() received an invalid combination of arguments` from the talker path.

Repro: `python -m sglang_omni.cli serve --model-path inclusionAI/Ming-flash-omni-2.0 --model-name ming-omni --thinker-tp-size 4 --thinker-gpus 0,1,2,3 --talker-gpu 4 --mem-fraction-static 0.80`, then any `/v1/audio/speech`-style request via `benchmarks.eval.benchmark_omni_seedtts --generate-only`. Happy to provide full server logs.

JiaxinD:
Reproduced the layer-3 talker error on 5x H100 (thinker TP4 + talker) at current main (24d7fb57, ming files byte-identical): text path healthy, the first speech request 500s with

```
TypeError: arange() received an invalid combination of arguments - got (Tensor, Tensor, device=torch.device)
  sglang_omni/models/ming_omni/talker/modeling_ming_omni_talker.py:545, in generate()
```

reached via the lazy talker warmup (`talker_executor._load_models -> initial_graph -> llm_job -> omni_audio_generation_func -> generate`).

[...]

PR #1132, opened 2026-07-21
[Fix] Ming speech path on transformers 5.x: cache_position ints, portable voice-preset paths

JiaxinD:
## Motivation

Addresses #1114 layers 2 and 3 (layer 1, the missing `x-transformers` dep, is already on main). The Ming speech path fails on current main:

1. **Layer 3**: the vendored talker feeds `StaticCache.get_seq_length()` straight into `torch.arange`. transformers 5.x returns a 1-element tensor there once the cache is non-empty (4.x returned int), so the first decode step of talker warmup raises `TypeError: arange() received an invalid combination of arguments - got (Tensor, Tensor, device=...)`. Root cause pinned by a CPU-only micro-repro in the #1114 thread.
2. **Layer 2**: `Ming-flash-omni-2.0`'s `talker/data/voice_name.json` ships absolute prompt-wav paths from the authoring machine (`/workspace/sglang-omni-pr1000-head/tests/data/query_to_cars.wav` for the default DB30 voice), so preset validation fails everywhere except the authoring rig.
3. **A further layer, surfaced once the two above were fixed**: the audio VAE builds plain `Qwen2Model` backbones that resolve to flash-attention, and transformers 5.6's FA integration crashes on them (`s_aux.to(...)` with `s_aux=None` in `flash_attention_forward`), killing prompt-wav registration.

## Modifications

* `talker/modeling_ming_omni_talker.py`: extract `_decode_cache_position` and coerce the cache length through `int()`.
* New `components/voice_presets.py`: `resolve_prompt_wav_path` tries the checkpoint-relative join first, then re-resolves path suffixes (longest first) against the talker dir and the fallback roots: `SGLANG_OMNI_VOICE_ASSET_ROOT` (the wheel-install escape hatch) and the checkout root when running from a checkout. Parent traversal fails closed, bare filenames only match inside the talker dir, and every fallback match must stay inside its root. Both `_validate_voice_presets` copies share it, log a warning when the fallback engages, and still fail closed when no candidate exists.
* `talker/audio_vae/vae_modules.py`: pin the VAE Qwen2 backbones to sdpa (the VAE path is not launch-bound; sdpa costs nothing measurable).

## Tests

`tests/unit_test/ming_omni/`: a contract test proving tensor bounds break raw `arange` (the #1114 failure mode) plus coercion cases for int / 1-element / 0-d cache lengths; suffix resolution under the talker dir, the env asset root, and the checkout root; traversal, containment, and same-name-shadowing fail closed; the no-candidate failure; VAE backbones asserted sdpa. Full ming_omni suite: 201 passed.

## Validation status

Unit-level complete. A 5-GPU end-to-end run (thinker TP4 + talker, no symlink workaround) at the two-fix commit verified layers 1-2 in vivo: health OK, text path OK, zero arange errors, the DB30 preset resolved through the suffix fallback (logged), and the remaining 500 was the VAE flash-attention crash this PR's third fix addresses. A full speech-success e2e at the final head follows as a comment.

### DIFF
--- a/sglang_omni/models/ming_omni/components/streaming_talker.py
+++ b/sglang_omni/models/ming_omni/components/streaming_talker.py
@@ -23,14 +23,13 @@ import numpy as np
 import torch
 
 from sglang_omni.models.ming_omni.components.streaming_text import uint8_tensor_to_text
+from sglang_omni.models.ming_omni.components.voice_presets import (
+    resolve_prompt_wav_path,
+)
 from sglang_omni.models.ming_omni.pipeline.next_stage import TALKER_STREAM_STAGE
 from sglang_omni.pipeline.stage.stream_queue import StreamItem
 from sglang_omni.proto import StagePayload
 from sglang_omni.scheduling.messages import IncomingMessage, OutgoingMessage
-
-from sglang_omni.models.ming_omni.components.voice_presets import (
-    resolve_prompt_wav_path,
-)
 
 logger = logging.getLogger(__name__)
 
--- a/sglang_omni/models/ming_omni/components/streaming_talker.py
+++ b/sglang_omni/models/ming_omni/components/streaming_talker.py
@@ -28,6 +28,10 @@ from sglang_omni.pipeline.stage.stream_queue import StreamItem
 from sglang_omni.proto import StagePayload
 from sglang_omni.scheduling.messages import IncomingMessage, OutgoingMessage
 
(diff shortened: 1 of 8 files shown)


JiaxinD:
End-to-end validation at head 271eb012 (5x H100, thinker TP4 + talker, no symlink workaround): health OK, text OK, speech request returns 200 with 1.2 MB of audio; zero arange errors, the DB30 preset resolved via the suffix fallback as designed.

PR #1134, opened 2026-07-22
[Perf] CUDA-graph the Qwen3-TTS code-predictor chain: ~1.9x at c8, ~3x at c32

### LINKED ISSUE #1119: [Perf] Qwen3-TTS negative scaling: eager per-token code predictor is 68% of the serving loop
### Symptom

Qwen3-TTS-12Hz-0.6B throughput DROPS when concurrency rises: 3.43 qps at c8 vs 2.76 qps at c32 (-20%) on one H100 (seed-tts-eval EN, 256 samples, fresh server per point). The server does batch under load (`#running-req` sits at 15-16 with CUDA graphs captured for the talker backbone at bs [1, 2, 4, 8, 12, 16]), so this is not an admission or graph-capture problem.

### Root cause [measured]

py-spy on the live serving process under steady c32 load (13,097 samples, 60 s at 200 Hz):

| frame (inclusive share of process) | % |
| --- | ---: |
| `run_batch` | 99.1 |
| `post_decode` -> `_collect_codes` | 76-81 |
| **`_predictor_forward_one_token` (sglang_model.py)** | **67.7** |
| `_predictor_cached_self_attention` | 39.5 |
| eager `UnquantizedLinearMethod.apply` | 13.8 |
| `_sample_subtalker_token_seeded` | 8.7 |

Two thirds of the serving loop is the code-predictor sub-transformer running **eagerly in Python inside `_collect_codes`**. *(edited: a code read corrected the mechanism below; the original text claimed per-request serial execution.)* The predictor IS batched across requests (its KV cache carries a batch dim), but per semantic token it runs `num_code_groups - 1` sub-iterations serially, each one an eager chain of lm_head, seeded sampling (full-vocab sort), codec embedding, and the full predictor layer stack, roughly a dozen kernel launches per layer. The AR talker backbone is CUDA-graphed; this chain is not, so every decode step pays thousands of Python-level launches. The residual question is which per-sub-iteration term grows with batch size to produce the negative c8 -> c32 slope (candidates: the [batch, vocab] sort in seeded sampling, SDPA over the growing predictor cache); the host-side launch share (68%) dominates either way.

### Fix directions

1. **CUDA-graph the whole per-token sub-iteration chain** (request batching already exists; shapes are fixed given the batch bucket, the predictor cache is only `num_code_groups + 1` tokens, and the chain is deterministic, so one capture per batch bucket collapses thousands of launches into one replay). This is the same pattern as the MOSS vocoder and Higgs decode-launch work.
2. Short of graphing, fusing/compiling the eager chain only shaves part of the launch tax (~14% sits in `UnquantizedLinearMethod.apply` alone).

Given the 68% inclusive share, graphing the predictor chain is the single biggest optimization available for this family; it should flip the c8 -> c32 slope positive.

### Setup

1x H100 80GB, `Qwen3TTSPipelineConfig` at current main, seed-tts-eval EN via `benchmark_tts_seedtts` (`--ref-format references`), fresh server per point, server NUMA-pinned. Profile: `py-spy record --pid <server> --subprocesses --rate 200` during steady c32.

JiaxinD:
## Motivation

Fixes #1119. Qwen3-TTS decode spends 67.7% of the serving loop inside `_collect_codes`: per semantic token the code predictor runs `num_code_groups - 1` serial sub-iterations of eager lm_head, seeded sampling, codec embedding, and the predictor layer stack, thousands of Python-level kernel launches per step. The talker backbone is CUDA-graphed; this chain was not, and its per-step cost produced the measured negative scaling (3.4 qps at c8 -> 2.8 at c32).

## Modifications

Capture the whole per-token predictor chain as one CUDA graph per batch bucket (reusing the backbone's bucket sizes). Graph keys pin every host branch of the eager sampling path: `(bucket, mode, quantized_top_k, has_top_p, has_unbounded_top_k)`. Per-step inputs (hidden state, semantic positions, per-row sampling params and seeds) flow through persistent device buffers via `copy_` before replay; mixed sampled/argmax batches fall back to eager. Kill switch: `SGLANG_OMNI_QTTS_PREDICTOR_GRAPH=0` (default on); the graph path also stays off when `server_args.disable_cuda_graph` is set or `tp_size > 1`, resolved at decode time because bootstrap forces `disable_cuda_graph` on during init for deferred capture.

Memory policy (from adversarial review): the sampled max top-k is ladder-quantized (4/8/16/32/50/64/.../1024, else full-sort) so one-off request values share graph keys instead of minting per-request graphs; total keys are capped at 32; after 8 distinct capture failures the graph path flips globally off with one log line, and a failed capture releases its pool via `CUDAGraph.reset()`. Disclosed behavior change: for top_k values not on the ladder (e.g. 40, which sorts at width 50), seeded sampling draws different, equally valid bits than the pre-PR eager code; the family default (50) is on the ladder, so the dominant signature is bit-unchanged. Ranks past a row's true top_k and nucleus-removed ranks are masked to `-inf` before the seeded sampler (its prob + gumbel scoring can otherwise select probability-0 ranks), so widening cannot sample outside the requested top_k and active `top_p` can no longer draw removed ranks; heterogeneous-top_k batches likewise change bits vs the pre-PR eager path, which could sample past a row's own k. The default path (top_k 50, top_p 1.0) has no masked ranks and keeps its bits. Graph-vs-eager bit-identity within this PR holds for all values.

## Evidence (1x H100, seed-tts-eval EN, 256 samples per leg, fresh server per leg)

Bit-identity gate: graphed vs eager codes `torch.equal` and summed embeddings 0-ulp at batch sizes 1/2/4/8/16 across four sampling variants (top-k, top-k + top-p, full-sort, argmax), padded replay and multi-step replay included; 44 dedicated tests (incl. memory-pool sharing across graph keys, key-sharing across request top_k values, per-step sampling-param refresh across replays, and masked-rank impossibility under ladder widening and nucleus removal, all proven to fail under injected bugs), full qwen3_tts suite 117 passed.

End-to-end serving throughput (full pipeline through the HTTP endpoint, client-measured completed requests per second), ABAB order, fresh server per leg, host loadavg logged:

| arm | c8 throughput (req/s) | c32 throughput (req/s) |
| --- | --- | --- |
| graph on | 4.86 / 4.26 | 5.94 |
| graph off | 2.29 / 2.48 | 2.00 |
| **delta** | **+91% (arm means, 4.56 vs 2.39)** | **+197% (single pair)** |

c8 ~1.9x (the on-off delta is 3.6x the worst within-arm spread, and the on legs landed on the worse load windows); c32 ~3x on a single pair. The #1119 negative c8 -> c32 slope flips positive with the graph and reproduces without it. Live serving captured 4 graph keys at c8 and 6 at c32 with zero auto-disables.

Real-checkpoint quality: corpus WER at c8, 256 samples, graph on vs off: **1.57% vs 1.38%** (within sampling noise), zero runaway (>50% WER) samples on either side, 256/256 generations succeeded with zero server errors on both arms.

## Checklist

- [x] Format your code with `pre-commit run --all-files`
- [x] Add unit tests


sglang_omni/models/qwen3_tts/sglang_model.py
```diff
@@ -358,6 +505,14 @@ def __init__(self, config: Any, quant_config: Any = None, prefix: str = "") -> N
         self._sub_sampled_has_top_p = False
         self._sub_sampled_max_top_k = 0
         self._sub_sampled_has_unbounded_top_k = False
+        self._predictor_graph_batch_sizes = self._normalize_predictor_graph_batch_sizes(
+            server_args,
+            max_batch_size=max_batch_size,
+        )
+        self._predictor_graphs: dict[tuple, _PredictorDecodeGraph] = {}
+        self._predictor_graph_disabled: set[tuple] = set()
+        self._predictor_graph_enabled = _predictor_graph_env_enabled()
```
SandyLuXY: `_predictor_graph_enabled` only honors `SGLANG_OMNI_QTTS_PREDICTOR_GRAPH`, so `server_args_overrides={"disable_cuda_graph": True}` and `--talker-cuda-graph off` still leave the predictor graph enabled and able to capture lazily. Please combine the predictor switch with the restored `server_args.disable_cuda_graph` value after `create_sglang_infrastructure_defer_cuda_graph()`.

sglang_omni/models/qwen3_tts/sglang_model.py
```diff
@@ -858,10 +1013,21 @@ def prepare_decode_buffers(self, requests: list[Any]) -> None:
...
-        self._sub_sampled_has_unbounded_top_k = len(bounded_top_ks) != len(
-            sampled_top_ks
-        )
+        has_unbounded_top_k = len(bounded_top_ks) != len(sampled_top_ks)
+        max_top_k = 0
+        max_bounded_top_k = max(bounded_top_ks, default=0)
+        if max_bounded_top_k > 0 and not has_unbounded_top_k:
+            # Note: (Jiaxin Deng) ladder-quantized so predictor-graph keys are
+            # shared across request top_k values; per-row masks keep true k.
+            quantized = _quantize_predictor_top_k(
```
SandyLuXY, line 1026: This widening does not preserve the request's true `top_k`: `sample_from_sorted_probs_with_seed_small_k()` adds Gumbel noise directly to probabilities, so masked ranks with probability 0 can still win. Please keep masked ranks impossible (or preserve the exact width); graph-vs-eager identity cannot catch this because both PR paths use the widened width.

SandyLuXY (changes requested):
Great work!! Left a few minor comments.

tests/unit_test/qwen3_tts/test_predictor_cuda_graph.py
```diff
@@ -0,0 +1,688 @@
...
+    talker._sub_sample_count = 0
+    talker._sub_has_sampled_rows = False
+    talker._sub_sampled_has_top_p = False
+    talker._sub_sampled_max_top_k = 0
+    talker._sub_sampled_has_unbounded_top_k = False
+
+    layer = SimpleNamespace(
+        input_layernorm=RMSNorm(HIDDEN, eps=1e-6).to(device, DTYPE),
+        post_attention_layernorm=RMSNorm(HIDDEN, eps=1e-6).to(device, DTYPE),
+        mlp=nn.Linear(HIDDEN, HIDDEN, bias=False).to(device, DTYPE),
```
edwingao28, line 125: This fixture replaces the production tensor-parallel MLP and projections with `nn.Linear`, so the CUDA graph tests never capture the TP collective path used by `Qwen3TTSTalkerDecoderLayer`. 

Please add a GPU test with TP=2 that captures and replays the predictor; the current single GPU coverage cannot detect collective capture or replay failures.

edwingao28 (commented):
left a test comment @JiaxinD, also should we add this new test file into tests/README.md？

sglang_omni/models/qwen3_tts/sglang_model.py
JiaxinD: Fixed in 5f6844ae: the switch now resolves at decode time as env AND not disable_cuda_graph AND tp_size == 1 (an init-time read would see the value bootstrap forces on for deferred capture).

sglang_omni/models/qwen3_tts/sglang_model.py
JiaxinD: Confirmed, the seeded sampler scores prob + gumbel so 0.0 ranks could win. Fixed in 5f6844ae: ranks past a row's true k are -inf before sampling, with a 100-seed test that fails under the old 0.0 mask.

tests/unit_test/qwen3_tts/test_predictor_cuda_graph.py
JiaxinD: 5f6844ae gates the predictor graph to tp_size == 1 (unit-tested), so TP serving keeps the eager path and single-GPU coverage matches the only graphed configuration. Happy to add a TP=2 capture test in a follow-up if we want graphs under TP.

JiaxinD:
Added test_predictor_cuda_graph.py to tests/README.md in 5f6844ae.

sglang_omni/models/qwen3_tts/sglang_model.py
JiaxinD: dd62e816 extends the same -inf treatment to nucleus-removed ranks (top_p had the identical 0.0 leak), with its own 100-seed test; default top_k 50 / top_p 1.0 bits unchanged.

edwingao28 (approved):

sglang_omni/models/qwen3_tts/sglang_model.py
```diff
@@ -39,6 +64,127 @@ def _sample_seeded_categorical(
...
+                    model._code_predictor_forward_incremental(
+                        self.layer0_codes,
+                        self.talker_hidden,
+                        semantic_positions=self.semantic_positions,
+                    )
+            current_stream.wait_stream(warmup_stream)
+
+            capture_stream = torch.cuda.Stream(device=device)
+            capture_stream.wait_stream(current_stream)
+            with torch.cuda.graph(
```
Hayden727, line 134: Each of the up to 32 graph keys captures with its own private allocator pool here, so request diversity can retain many copies of the predictor intermediates and exhaust GPU memory. Please create one model-owned `torch.cuda.graph_pool_handle()` and pass it as `pool=` to these sequential captures.

JiaxinD: Fixed in 5e83c6f8: all captures now go through one model-owned graph_pool_handle(), with a test asserting distinct keys record the same pool.

Hayden727 (approved):

estellaliu233:
Hi — I've been putting together a write-up on diagnosing host-bound serving: how to tell a GPU bubble apart from a slow kernel, which workloads are most exposed to it, and what engines and operators can each do about it. I used this PR as the **launch-bound** case:

The numbers are quoted from the PR description — 67.7% of the serving loop in `_collect_codes`, the negative scaling (3.4 QPS at c8 → 2.8 at c32), and ~1.9× at c8 / ~3× at c32 after capturing the predictor chain. The lesson I drew from it is that a captured backbone does not mean a captured step.

If I've misread the context anywhere, or overstated a conclusion the PR didn't actually measure, I'd be glad to correct it — thanks for the work.

PR #1027, opened 2026-07-11
[Qwen3-Omni Perf] Fix async drop-stale KV slot reslice for extend/mixed batches

edwingao28:
## Problem

Under asynchronous decode at c>=32, an extend/mixed prefill batch can reach the KV write with `out_cache_loc=None`, aborting the request server-side (~1/64 at c32 on the audited main revision `45582ff5`; CUDA illegal access and scheduler death on older deps).

<details>
<summary>Real traceback from the audited c32 run (revision 45582ff5)</summary>

```
2026-07-08 06:57:17,269 [ERROR] sglang_omni.scheduling.omni_scheduler: OmniScheduler batch failed for requests=['8689e71c-6028-4a89-8a2c-51a303551e84']
Traceback (most recent call last):
  File "/data/main-clean/sglang_omni/scheduling/omni_scheduler.py", line 815, in run_batch
    return self._run_batch(batch, pp_proxy_tensors)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/main-clean/sglang_omni/scheduling/omni_scheduler.py", line 837, in _run_batch
    mr_output = self._model_runner.execute(sched_output)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/main-clean/sglang_omni/model_runner/thinker_model_runner.py", line 64, in execute
    return super().execute(scheduler_output)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/main-clean/sglang_omni/model_runner/base.py", line 145, in execute
    batch_result = self._prepare_and_forward(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/main-clean/sglang_omni/model_runner/base.py", line 308, in _prepare_and_forward
    batch_result = self.custom_prefill_forward(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/main-clean/sglang_omni/model_runner/thinker_model_runner.py", line 86, in custom_prefill_forward
    return self._forward_with_omni_embeds(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/main-clean/sglang_omni/model_runner/thinker_model_runner.py", line 308, in _forward_with_omni_embeds
    hidden_states = outer.model(
                    ^^^^^^^^^^^^
[...]

Issue #1024, opened 2026-07-11
[Qwen3-Omni Perf] Thinker admission: raise max_running_requests default and expose CLI override


edwingao28:
## Problem

`scheduling/sglang_backend/server_args_builder.py:16` hardcodes `max_running_requests=16` for all sglang-omni AR engines, and the `--max-running-requests` CLI override is rejected for `Qwen3OmniPipelineConfig`. The decode batch (and the tied `cuda_graph_max_bs`) is capped at 16, so half of a c32 workload queues.

## Measured signal

- Synchronous diagnostic on clean main `45582ff5`: c32 `1321 -> 1710` tok/s (+29.4%), c48 `1379 -> 1954` tok/s (+41.7%). Opportunity signal only — it does not validate the normal async path.
- Earlier-environment patch 16 -> 64: TTFT p95 `3.36 -> 1.31 s` (-61%) with only +13% throughput — the cap is primarily a tail-latency bug at high concurrency.

## Design status

Blocked on #1023: the decisive A/B needs a trustworthy async baseline. #847 already raised the talker cap to 32; this issue covers the thinker.

## Open decisions

- [x] New default (32 vs 64) — pick from the post-#1023 A/B, keep 16 as documented rollback.
- [x] Thinker-local override vs. exposing the cap for all omni AR engines.

## Task checklist

- [x] Make the cap thinker-local, raise the default, expose a CLI override; keep the old value as rollback
- [x] Post-#1023 paired A/B at c `{1,16,32,48}` per #1018's contract (A/A noise band, K=3 repeats): no c1/c16 regression, output checks pass, zero failures
- [x] Record the native `audio -> text` baseline in the same sweep (currently missing)

Part of #1022. Design and benchmark contract: #1018 (item 1.2).

hiSandog:
The planned concurrency sweep could record memory headroom and queue or rejection behavior alongside throughput and TTFT. A value that improves high-concurrency latency may still be unsuitable as a default on smaller configurations. It would also help to warn when an override makes the admission cap inconsistent with the associated CUDA graph batch limit.

edwingao28:
cloased as finished in #1135

PR #1135, opened 2026-07-22
[Qwen3-Omni Perf] Raise thinker max_running_requests default to 64

edwingao28:
## Motivation

The thinker inherits `max_running_requests=16` from `build_sglang_server_args` (`server_args_builder.py:16`). Under concurrent load this admission cap binds before GPU capacity: decode-ready requests queue outside the scheduler and TTFT grows with the queue. There is no CLI escape hatch — `--max-running-requests` maps to the generation role (talker_ar on the speech config, rejected on the text-only config) — so the thinker cap is effectively hardcoded. The talker already defaults to 32; the thinker is the only stage still at 16.

Part of the thinker serving work in #1022. Stacked on #1027 (first commit belongs to that PR).

## Modifications

One line: `max_running_requests=64` in the qwen3_omni thinker overrides (`create_sglang_thinker_executor_from_config`). Config-level `runtime.sglang_server_args.max_running_requests` still wins. The KV pool is sized by `mem_fraction_static`, so the cap adds no KV memory; the extra footprint is CUDA-graph capture at larger batch sizes.

## Benchmarking and Profiling

Baseline: main at bfcfabec. Text-only thinker, TP2, Qwen3-Omni-30B-A3B-Instruct, 96 prompts, 200 words in / 128 tokens out, seed 1234, streaming. H100 2x dedicated GPUs, mem-fraction 0.85. Output tok/s and TTFT p95 (s):

| conc | 16 (base) | 32 | 64 | TTFT p95 base | TTFT p95 @64 |
|------|-----------|------|------|------|------|
| 1  | 204  | 201  | 194  | 0.05 | 0.07 |
| 16 | 1449 | 1949 | 1906 | 0.22 | 0.21 |
| 32 | 2061 | 2514 | 2218 | 1.15 | 0.26 |
| 48 | 2052 | 2364 | 2303 | 2.18 | 0.35 |
| 64 | 2056 | 2333 | 2228 | 3.15 | 0.45 |

Throughput +8–12% at concurrency 32–64; TTFT p95 drops from seconds to under 0.5s (−78% to −86% at concurrency 48/64). Cap 32 peaks slightly higher on raw throughput but keeps 2s+ TTFT p95 once concurrency passes the cap. The thinker's first token gates every downstream stage in the streaming pipeline, so the default goes to 64.

OOM checks, all clean:

- H100 text-only TP2, mem-fraction 0.85: peak 73 GiB / 80 GiB through concurrency 64.
- H200 text-only TP2, mem-fraction 0.50 with ~60 GiB of unrelated resident memory per GPU: peak 117 GiB / 143 GiB. (Shared box, so throughput numbers there were noisy; the table uses the dedicated H100 runs.)
- Speech pipeline at the tightest supported layout (thinker TP2 + talker + code2wav, mem-fraction 0.55): 16/16 requests OK at cap 64, RTF p50 0.245.

## Checklist

### DIFF
--- a/sglang_omni/models/qwen3_omni/stages.py
+++ b/sglang_omni/models/qwen3_omni/stages.py
@@ -960,15 +960,14 @@ def create_sglang_thinker_executor_from_config(
     # the streaming-SST thinker path benefits out of the box; either key can be
     # overridden via server_args_overrides (set enable_mixed_chunk=False to opt
     # out). Note: mixed-chunk only engages when chunked_prefill_size > 0.
-    overrides: dict[str, Any] = {
-        "disable_cuda_graph": False,
-        "enable_mixed_chunk": True,
-        "chunked_prefill_size": 8192,
-        "max_running_requests": 64,
-        "sampling_backend": "pytorch",
-    }
-    if server_args_overrides:
-        overrides.update(server_args_overrides)
+    overrides = build_generation_batch_overrides(
+        max_running_requests=64,
+        server_args_overrides=server_args_overrides,
+        disable_cuda_graph=False,
+        enable_mixed_chunk=True,
+        chunked_prefill_size=8192,
+        sampling_backend="pytorch",
+    )
     overrides["tp_size"] = tp_size
     has_explicit_colocated_mem_fraction = (
         total_gpu_memory_fraction is not None
@@ -990,6 +989,10 @@ def create_sglang_thinker_executor_from_config(
         model_path,
         context_length=thinker_max_seq_len,
         **overrides,
+    )
+    validate_generation_batch_policy(
+        model_name="Qwen3-Omni thinker",
+        server_args=server_args,
     )
     if total_gpu_memory_fraction is None:
         encoder_reserve_applied = _apply_qwen_thinker_encoder_reserve(
--- a/sglang_omni/models/qwen3_omni/stages.py
+++ b/sglang_omni/models/qwen3_omni/stages.py
@@ -964,6 +964,7 @@ def create_sglang_thinker_executor_from_config(
         "disable_cuda_graph": False,
         "enable_mixed_chunk": True,
         "chunked_prefill_size": 8192,
+        "max_running_requests": 64,
         "sampling_backend": "pytorch",
     }
     if server_args_overrides:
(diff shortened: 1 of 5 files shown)


sglang_omni/models/qwen3_omni/stages.py
```diff
@@ -964,6 +964,7 @@ def create_sglang_thinker_executor_from_config(
         "disable_cuda_graph": False,
         "enable_mixed_chunk": True,
         "chunked_prefill_size": 8192,
+        "max_running_requests": 64,
```
SandyLuXY: The Thinker sets only `max_running_requests=64`. It does not use `build_generation_batch_overrides()` or call `validate_generation_batch_policy()`, so CUDA Graph and compile coverage remain implicit and hardware-dependent. This differs from the Qwen3-Omni Talker in the same file, which builds and validates the complete policy.
Consider building the Thinker defaults through `build_generation_batch_overrides()`, validate the final `ServerArgs`, and add a factory test covering the default 64, rollback overrides, CUDA Graph coverage, and compile coverage. If deliberate eager fallback above the graph cap is required, make that an explicit,
validated Thinker policy and benchmark it.

SandyLuXY: Please add the Thinker rollback surface required by #1024: `--max-running-requests` is rejected for Qwen3-Omni text mode and targets `talker_ar` in speech mode, so it cannot override this value. Wire a Thinker-specific CLI override, test both variants, and document the actual `runtime_overrides.thinker.server_args_overrides.max_running_requests` config path.

SandyLuXY (changes requested):

sglang_omni/models/qwen3_omni/stages.py
yxs: looks like just one line diff after PR 1027 merge

edwingao28: Yeah just a one line diff to raise the thinker admission cap better utilize GPU capacity. Will rebase once 1027 lands

edwingao28: fixed in https://github.com/sgl-project/sglang-omni/pull/1135/commits/82aadfe3d1948de8da60349336c1d2d6c38389ce

edwingao28: also fixed in https://github.com/sgl-project/sglang-omni/pull/1135/commits/82aadfe3d1948de8da60349336c1d2d6c38389ce

SandyLuXY (approved):

Issue #1120, opened 2026-07-21
[Roadmap] Consumer GPU support for ASR and TTS serving


Hayden727:
## Summary

SGLang-Omni is primarily developed and validated on datacenter GPUs, while many supported ASR and TTS models are small enough to run on consumer hardware. This roadmap tracks support for single-GPU colocated ASR/TTS serving on Linux.

The first target is one RTX 4090 (Ada, SM89, 24 GB). RTX 5090 (consumer Blackwell, SM120, 32 GB) remains a later architecture-compatibility target, but the initial implementation and validation should start from a single RTX 4090.
All stages of a supported pipeline should run colocated on the same GPU. Tensor parallelism, multi-GPU stage placement, consumer-specific router work, RTX 5090 NVFP4 optimization, Windows/WSL2, and dedicated consumer-GPU CI are out of scope for this roadmap.
## Why this is needed

Current defaults contain assumptions inherited from H100/H200/B200-class systems:
- `flash-attn-4` is a required dependency even though it targets Hopper and Blackwell rather than Ada.
- Some model paths default to FA3 or use legacy FlashAttention APIs.
- Some FP8/CUTLASS paths only recognize SM90/SM100.
- Default concurrency, CUDA Graph batch sizes, KV-cache allocation, encoder caches, and static memory fractions are sized for high-memory GPUs.
- Model compatibility is often inferred from weight size even though KV cache, CUDA Graph pools, compile workspaces, codecs/vocoders, request state, and fragmentation determine the real peak.
RTX 4090 is SM89 and RTX 5090 is SM120; SM120 is not equivalent to B200/SM100. They need architecture-appropriate kernels and separate memory validation.

Hardware references: [RTX 4090](https://www.nvidia.com/en-us/geforce/graphics-cards/40-series/rtx-4090/), [RTX 5090](https://www.nvidia.com/en-us/geforce/graphics-cards/50-series/rtx-5090/), and [CUDA compute capability](https://developer.nvidia.com/cuda-gpus).
## User and platform assumptions
- Linux is the supported host platform for this roadmap.
- The primary user has one 24 GB RTX 4090.
- The target workload is local development, evaluation, and low-to-moderate-concurrency serving.
- Every supported model must run as a complete colocated pipeline on one GPU.
[...]

PR #1141, opened 2026-07-22
feat: add model-free GPU diagnostics

Dayuxiaoshui:
## Motivation

Issue #1120 needs a model-free hardware discovery path before startup-time backend selection and memory planning can be added for consumer GPU ASR and TTS serving.

This PR adds a small GPU preflight command that reports environment, device mapping, memory, CUDA/PyTorch versions, and backend importability without loading model configuration or weights and without starting the serving entrypoints. It also exposes reusable raw GPU capability data for downstream model code while keeping serving-time backend selection outside the diagnostics layer.

## Modifications

- Add `sgl-omni check-gpu` with compact terminal output and machine-readable `--json` output.
- Report `CUDA_VISIBLE_DEVICES` logical-to-physical mapping, GPU identity, compute capability, total/free memory, driver version, CUDA driver/runtime versions, and the PyTorch CUDA build.
- Expose `collect_gpu_capabilities()` and `get_gpu_compute_capability()` so downstream model PRs can consume model-free hardware facts without adding another GPU detector.
- Keep serving-time backend selection in model-owned code; diagnostics reports capabilities and backend importability but does not choose a serving backend.
- Isolate NVML failures per physical device and per queried field so one failed query does not hide later GPUs or discard otherwise valid metadata.
- Warn explicitly for unsupported MIG physical mapping/free-memory reporting.
- Probe backend imports in short-lived subprocesses so imports cannot initialize CUDA in the diagnostics process and a crashing extension cannot terminate the report.
- Resolve backend distribution names and versions through `importlib.metadata.packages_distributions()`, avoiding CUDA-variant-specific wheel names.
- Add opt-in `--strict` exit semantics for CI/preflight use. Strict mode exits nonzero when warnings are present, CUDA is unavailable, or no CUDA GPU is visible.
- Keep serving entrypoints lazily imported so `check-gpu` does not load a model or start the server.
- Add unit coverage for reordered visibility, NVML device/field failures, MIG handling, backend import failures and hard crashes, CUDA-variant distribution resolution, JSON output, strict exit codes, serving-entrypoint isolation, and the shared capability API boundary.

## Related Issues

Related to #1120.

## Accuracy Test

Not applicable. This PR does not change model architecture, kernels, serving-time backend selection, or inference numerics.

Validation:

```text
python -m pytest -q tests/unit_test/diagnostics tests/unit_test/pipeline/test_gpu_memory.py
.........................................                                [100%]
41 passed
```

Validated on an NVIDIA GeForce RTX 5090:

```bash
CUDA_VISIBLE_DEVICES=0 sgl-omni check-gpu --strict
```

```text
SGLang-Omni GPU diagnostics (no model loaded)
CUDA_VISIBLE_DEVICES: 0
Driver: 595.71.05
CUDA driver/runtime: 13.2 / unavailable
PyTorch/CUDA build: 2.8.0+cu128 / 12.8
GPUs:
  logical 0 -> physical 0 visible=0 name=NVIDIA GeForce RTX 5090 cc=12.0 memory=31.36GiB/31.84GiB uuid=GPU-830bed08-697e-e6a0-0e33-7937244c0830 pci=00000000:63:00.0
Backends:
  attention/flash-attn-4: not installed; version=-
  attention/flashinfer: not installed; version=-
  attention/triton: available; version=3.4.0
  attention/torch-sdpa: available; version=2.8.0+cu128
  gemm/sgl-deep-gemm: not installed; version=-
  gemm/sglang-kernel: not installed; version=-
  moe/quack-kernels: not installed; version=-
  quantization/torchao: not installed; version=-
  quantization/compressed-tensors: not installed; version=-
  communication/nixl: not installed; version=-
  communication/mooncake: not installed; version=-
```

The JSON output was parsed successfully, and `CUDA_VISIBLE_DEVICES=1,0` mapped logical GPUs to physical GPUs as `[(0, 1), (1, 0)]`. The shared capability accessor returned `(12, 0)` on the RTX 5090 without running optional backend probes.

## Benchmark & Profiling

No serving hot path is changed. The isolated backend probes completed the full command in approximately 6.1 seconds on the validation host. The shared capability path does not run those optional backend probes.

## Checklist

- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed. No user documentation is added in this scoped increment; CLI help and this PR description cover the new command.
- [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed. Not applicable to model accuracy or serving throughput.

## CI

if labeled.


sglang_omni/diagnostics/gpu.py
```diff
@@ -0,0 +1,444 @@
...
+def _package_version(distribution: str) -> str | None:
+    try:
+        return importlib.metadata.version(distribution)
+    except importlib.metadata.PackageNotFoundError:
+        return None
+
+
+def _module_available(module: str) -> bool:
+    try:
+        return importlib.util.find_spec(module) is not None
```
Hayden727: `find_spec()` only proves that the module is discoverable; a CUDA extension with missing shared libraries or incompatible symbols will still be reported as `importable`/`available`. Please either attempt the import and retain the exception as the diagnostic reason, or rename this result to `discoverable` so the report does not overstate backend availability.

sglang_omni/diagnostics/gpu.py
```diff
@@ -0,0 +1,444 @@
...
+) -> dict[str, Any]:
+    torch_uuid = _normalize_uuid(getattr(properties, "uuid", None))
+    if torch_uuid in by_uuid:
+        return by_uuid[torch_uuid]
+
+    if logical_index < len(visible_devices):
+        visible = visible_devices[logical_index]
+        if isinstance(visible, int):
+            return by_index.get(visible, {})
+        return by_uuid.get(_normalize_uuid(visible), {})
```
Hayden727, line 257: This exact lookup cannot map `CUDA_VISIBLE_DEVICES=MIG-...`: the NVML inventory is keyed by physical GPU UUIDs, so MIG-visible devices silently lose their physical index, free memory, and topology. Please enumerate/map MIG devices, or emit an explicit unsupported-mapping warning instead of returning an empty physical record.

tests/unit_test/diagnostics/test_gpu.py
```diff
@@ -0,0 +1,179 @@
...
+    check_gpu_module = importlib.import_module("sglang_omni.cli.check_gpu")
+    report = {
+        "schema_version": 1,
+        "environment": {"logical_device_count": 0},
+        "gpus": [],
+    }
+    monkeypatch.setattr(
+        check_gpu_module,
+        "collect_gpu_diagnostics",
+        lambda: report,
```
Hayden727, line 316: Mocking the collector means this test only checks Typer JSON output; it does not verify the stated lazy-import boundary. Please also assert that invoking `check-gpu` does not load modules such as `sglang_omni.serve.openai_api`/`launcher` (preferably in an isolated subprocess so prior imports cannot hide a regression).

Hayden727 (changes requested):

sglang_omni/diagnostics/gpu.py
Dayuxiaoshui: Updated backend checks to perform a real import_module() and retain the exception type and message as the diagnostic reason on failure.

sglang_omni/diagnostics/gpu.py
Dayuxiaoshui: Added an explicit warning for unresolved MIG-... mappings, marking physical memory and topology information as unsupported.

tests/unit_test/diagnostics/test_gpu.py
Dayuxiaoshui: Added an isolated subprocess test that invokes check-gpu and verifies serve.launcher and serve.openai_api are not loaded.

sglang_omni/diagnostics/gpu.py
```diff
@@ -0,0 +1,457 @@
...
+        system["cuda_driver_api_version"] = _cuda_version(
+            int(pynvml.nvmlSystemGetCudaDriverVersion_v2())
+        )
+    except Exception as exc:
+        warnings.append(f"NVML CUDA driver query failed: {exc}")
+
+    try:
+        count = int(pynvml.nvmlDeviceGetCount())
+        for physical_index in range(count):
+            handle = pynvml.nvmlDeviceGetHandleByIndex(physical_index)
```
Hayden727, line 185: One failed NVML query currently exits this entire loop, so every healthy GPU after the failing index disappears from the inventory. Please catch errors per `physical_index`, record an index-specific warning, and continue enumerating the remaining devices.

sglang_omni/diagnostics/gpu.py
```diff
@@ -0,0 +1,457 @@
...
+        matrix[source][target]
+        for source in range(device_count)
+        for target in range(device_count)
+        if source != target
+    ]
+    if all(pairs):
+        return matrix, "full", "All visible GPU pairs support direct peer access."
+    return (
+        matrix,
+        "unavailable",
```
Hayden727: This reports `unavailable` when P2P is only partial; for example, `0↔1=True` with GPU 2 isolated produces a matrix containing supported pairs but the summary says none are available. Please distinguish `partial` from `unavailable` while still disabling custom all-reduce unless the matrix is full.

Hayden727:
This is a worthwhile foundation for the hardware-discovery milestone in #1120, but I do not think the current head is ready to merge yet.

The independently useful core is the model-free `check-gpu` path: environment and device mapping, memory and CUDA/PyTorch versions, backend importability, machine-readable JSON, and avoiding the serving entrypoints. That gives later startup-time backend selection and memory planning a concrete source of facts without requiring model weights.

Before merging, I recommend:

- Fix the two remaining inline findings: isolate NVML failures per physical device, and distinguish partial P2P support from fully unavailable P2P.
- Remove the all-`None` selected-backend/CUDA Graph/`torch.compile` placeholders until those checks are implemented and their schema is known.
- Consider dropping topology/P2P from this first increment: #1120 currently targets single-GPU colocated serving and lists TP/multi-GPU placement as non-goals. They can return when a concrete multi-GPU consumer needs them.
- Split the `relay/nixl.py` import compatibility change into a separate PR. It changes the runtime relay path and expands the validation surface, while the diagnostic command does not require that behavior change.
- Add a short user-facing command example plus one real target-GPU output. The PR currently has no substantive description and no GPU CI evidence.

With that narrower scope and the remaining correctness fixes, I would support merging this as the first small deliverable under #1120 rather than waiting for selected-backend and startup integration work.

Dayuxiaoshui:
@Hayden727 Thanks for the detailed review. I have narrowed the PR to the single-GPU, model-free diagnostics scope in #1120: NVML failures are now isolated per `physical_index` so later healthy devices are still enumerated, topology/P2P reporting and the unimplemented backend-selection, CUDA Graph, and `torch.compile` placeholders have been removed, and the NIXL/relay compatibility changes have been reverted. The command continues to report the visible GPU mapping, hardware and CUDA/PyTorch versions, installed backend inventory, and machine-readable JSON without loading a model or starting a server. The diagnostics tests pass, and I also validated the command on an NVIDIA GeForce RTX 5090.

```bash
CUDA_VISIBLE_DEVICES=0 sgl-omni check-gpu
```

```text
SGLang-Omni GPU diagnostics (no model loaded)
CUDA_VISIBLE_DEVICES: 0
Driver: 595.71.05
CUDA driver/runtime: 13.2 / unavailable
PyTorch/CUDA build: 2.8.0+cu128 / 12.8
GPUs:
  logical 0 -> physical 0 visible=0 name=NVIDIA GeForce RTX 5090 cc=12.0 memory=31.36GiB/31.84GiB uuid=GPU-6453caef-41c8-9092-c647-651e6e26bded pci=00000000:50:00.0
Backends:
  attention/triton: available; version=3.4.0
  attention/torch-sdpa: available; version=2.8.0+cu128
```

sglang_omni/diagnostics/gpu.py
```diff
@@ -0,0 +1,354 @@
...
+def _package_version(distribution: str) -> str | None:
+    try:
+        return importlib.metadata.version(distribution)
+    except importlib.metadata.PackageNotFoundError:
+        return None
+
+
+def _module_import_error(module: str) -> str | None:
+    try:
+        importlib.import_module(module)
```
MayDomine: Real imports are the right call for accuracy, but doing them **in-process** makes `check-gpu` heavier and riskier than a diagnostics tool should be:

1. Several of these modules (`flashinfer`, `triton`, `deep_gemm`, `flash_attn.cute`) run JIT/CUDA setup at import time, so the command can initialize a CUDA context — and in some cases allocate device memory — on a GPU that is mid-serving, which is precisely when someone reaches for diagnostics.
2. A broken extension that hard-crashes on import (e.g. the `libcudart.so` case from your own test, which in practice can be a segfault rather than an `OSError`) takes the whole report down with it.

Consider probing each module in a short-lived subprocess (`[sys.executable, "-c", f"import {module}"]`, keep stderr as the `reason`) — the same isolation idea as your lazy-import subprocess test. That keeps the report alive through hard crashes and keeps the main process off the GPU.

sglang_omni/diagnostics/gpu.py
```diff
@@ -0,0 +1,354 @@
...
+    ("gemm", "sglang-kernel", "sglang-kernel", "sgl_kernel"),
+    ("moe", "quack-kernels", "quack-kernels", "quack"),
+    ("quantization", "torchao", "torchao", "torchao"),
+    (
+        "quantization",
+        "compressed-tensors",
+        "compressed-tensors",
+        "compressed_tensors",
+    ),
+    ("communication", "nixl", "nixl-cu13", "nixl_cu13"),
```
MayDomine: These two entries hardcode one CUDA-variant distribution name (`nixl-cu13`, `mooncake-transfer-engine-cuda13`). On a CUDA-12 stack the same backends ship as `nixl` / `mooncake-transfer-engine`: the module imports fine, but `_package_version()` returns `None`, and since the import probe is gated on `version is not None`, a fully usable backend is reported as `not installed`.

Suggest inverting the source of truth: always attempt the import (that is what actually matters for availability), and resolve the version via `importlib.metadata.packages_distributions()` (module → owning distributions) instead of a single pinned dist name. That keeps the report correct across cu12/cu13 wheels and future package renames.

sglang_omni/diagnostics/gpu.py
```diff
@@ -0,0 +1,354 @@
...
+
+    inventory, system, warnings = _nvml_inventory(pynvml)
+    try:
+        devices = _logical_devices(torch, visible_devices, inventory, warnings)
+    finally:
+        if pynvml is not None:
+            _shutdown_nvml(pynvml)
+
+    for device in devices:
+        device.pop("_handle", None)
```
MayDomine: `_handle` is stored in the NVML inventory, threaded through `_logical_devices`, and then unconditionally popped here without ever being read — and it would be unusable anyway, since `_shutdown_nvml` has already run in the `finally` above. Looks like a leftover from the removed topology/P2P pass; dropping the field removes code at all three sites.

sglang_omni/cli/check_gpu.py
```diff
@@ -0,0 +1,30 @@
...
+        ),
+    ] = False,
+) -> None:
+    """Report GPU mapping, runtime versions, and installed backends."""
+
+    report = collect_gpu_diagnostics()
+    if json_output:
+        typer.echo(json.dumps(report, indent=2, sort_keys=True))
+        return
+    typer.echo(render_gpu_diagnostics(report))
```
MayDomine: `check-gpu` always exits 0 — including when `cuda_available` is false, NVML init failed, or an installed backend fails to import. For the preflight/CI use-case in #1120 it would be useful to gate on this: exit nonzero (or add `--strict`) when there are warnings or no visible CUDA device. As-is, a launcher script cannot distinguish "healthy" from "completely blind" without parsing the JSON.

MayDomine (commented):
Went through the current head after the first review round — the scope narrowing and the follow-up fixes (real import probes, per-index NVML isolation, MIG warning, the subprocess lazy-import test) all look good.

A few more points inline. The two that matter: in-process backend imports (can touch a live GPU and dies with a broken extension — the exact failure this tool should survive), and the hardcoded cu13 distribution names (false `not installed` on cu12 stacks). Two smaller ones: `_handle` dead code and exit-code semantics.

Housekeeping:
- The PR description is still the empty template — please fill in Motivation/Modifications and put #1120 under Related Issues.
- This PR is opened from your fork's `main` branch; a feature branch per PR makes follow-ups and rebases much easier.
- Nit: `test_collect_gpu_diagnostics_preserves_reordered_visible_mapping` asserts the absence of strings from features that no longer exist (`Selection:`, `P2P:`, `Topology:`, `CUDA Graph/torch.compile:`). Those assertions won't age well — suggest dropping them.

Dayuxiaoshui:
cc @Hayden727 @MayDomine

Hayden727:
Linked to #1120. I checked the actual patch: this PR owns model-free device/backend diagnostics (`diagnostics/gpu.py`) and does not perform serving-time backend selection. Please keep that boundary; downstream model PRs can consume its capability data instead of adding another GPU detector.

Hayden727 (approved):

ChaoyuWang04:
SM120 datapoint from the MOSS-TD RTX 5090 validation (#1189) that may be worth a diagnostics check: FlashInfer can require a JIT toolchain at server startup on SM120.

Observed on main @ dc64a6b2, torch 2.11.0+cu130, sglang 0.5.12.post1, FlashInfer default-selected:

- A benchmark-harness-launched MOSS-TD server failed during CUDA-graph capture with a bare `FileNotFoundError: [Errno 2] No such file or directory: 'ninja'` from `flashinfer/jit/core.py` (`run_ninja`, building a batch-decode module).
- `pip install ninja` into the venv (and having it on PATH) resolved it.
- Note: an earlier manual `sgl-omni serve` launch of the same model on the same machine completed capture without triggering the JIT path; the trigger difference between the two launch configurations was not isolated.

Suggestion: `check-gpu` (or startup diagnostics) could verify ninja/toolchain presence when the selected backend may JIT on the running arch, and the failure message could name the missing-toolchain cause instead of the bare FileNotFoundError.

Also from the same validation, in case useful for the capability matrix: flash-attn-4 4.0.0b15 installs (py3-none-any) but its CuTe kernel dispatch fails on capability 12.0 with `AttributeError: 'NoneType' object has no attribute '_trait'` — FA4 is install-weight only on SM120 while serving runs FlashInfer (same class as the #1151 SM89 note).

Dayuxiaoshui:
@ChaoyuWang04  Thanks, this is a useful SM120 datapoint. Since #1183 owns the dependency baseline and may change the SGLang/FlashInfer startup behavior, I’ll reproduce both launch paths after it merges as part of the RTX 5090 installation compatibility matrix, checking ninja availability, when FlashInfer JIT is triggered, and the distinction between FA4 being installed/importable and actually executable on SM120; any follow-up should provide an actionable startup preflight error without expanding model-free diagnostics into serving-time backend selection.

Issue #1151, opened 2026-07-23
[Benchmark] Qwen3-ASR 1.7B on a single RTX 4090


wirybeaver:
Qwen3-ASR 1.7B RTX 4090 baseline and validation is complete.

Parent roadmap: #1120

Implementation and review:
- Umbrella Draft: #1152
- Serve lifecycle: #1153
- Qwen3-ASR RTX 4090 profile: #1154
- Reproducible benchmark tooling and results: #1155
- Full-length preprocessing correctness fix: #1176
- Context-capacity validation follow-up: #1193

Environment:
- Linux 6.8 / Ubuntu 22.04.5
- 1x RTX 4090 24,564 MiB (SM89), driver 580.159.04, 450 W limit
- CUDA 13.0, PyTorch 2.11.0, SGLang 0.5.12.post1, Transformers 5.6.0
- Base commit `aa6f92c30e5279ed831776d0d6649344df3a94cb` plus the proposed change set
- Model revision `7278e1e70fe206f11671096ffdd38061171dd6e5`
- SeedTTS revision `27f4c1adee83b5b29b7c4b375f6b976324bda308`

Validated profile:
- BF16
- FlashInfer LM attention; Triton multimodal attention selected explicitly for SM89
- CUDA Graph batches `[1, 2, 4, 8, 12, 16]`
- `max_running_requests=16`, `mem_fraction_static=0.65`
- `torch.compile` disabled

Quality/performance:
- SeedTTS EN, 1088/1088 clips at every level, three measured repeats after warmup:
  - c=1: 10.527 samples/s, 0.095 s mean latency, worst WER 0.0123
  - c=8: 45.102 samples/s, 0.177 s mean latency, worst WER 0.0123
  - c=16: 65.428 samples/s, 0.243 s mean latency, worst WER 0.0123
  - c=32 queueing stress: 65.747 samples/s, 0 skips
- EN WER range was 0.0120-0.0123 versus the documented H100 BF16 reference 0.0122.
- SeedTTS ZH, 2020/2020 clips at every level:
  - c=1: 11.671 samples/s, 0.086 s mean latency, CER 0.0062
  - c=8: 55.214 samples/s, 0.145 s mean latency, CER 0.0062
  - c=16: 79.532 samples/s, 0.201 s mean latency, CER 0.0062
  - c=32 queueing stress: 75.583 samples/s, 0 skips

Memory/stability:
[...]

PR #1152, opened 2026-07-23
[Benchmark] Validate Qwen3-ASR 1.7B on RTX 4090

wirybeaver:
## Summary
- add the measured BF16/SM89 profile and runtime diagnostics used for the single-RTX-4090 validation
- add pinned, provenance-rich EN/ZH benchmarks and a 30-minute mixed stability harness
- preserve full audio beyond Whisper's former 30-second preprocessing boundary and report genuine context limits separately
- include lifecycle/input fixes, tests, documentation, and measured results

## Validation
- Full SeedTTS EN: 1088/1088 samples at each concurrency, WER 0.0120-0.0123
- Full SeedTTS ZH: 2020/2020 samples at each concurrency, CER 0.0062
- 30-minute c=1/4/8/16 soak: 78,418 successful requests, zero unexpected errors
- Original soak peak sampled GPU memory: 17,990 MiB on one RTX 4090
- Targeted post-#1176/#1193 RTX 4090 validation:
  - 31-second input: 3/3 HTTP 200, identical 80-word transcript
  - 60-second input: 3/3 HTTP 200, identical 146-word transcript
  - 124.91-second over-context input: 3/3 HTTP 400 before mel extraction (`1637 + 128 > 1636` tokens)
  - targeted-run peak GPU memory: 17,464 MiB; minimum free memory: 6,618 MiB

The full SeedTTS sweeps and 30-minute short-clip soak were not repeated. All corpus samples completed while the 30-second guard was active, and `truncation=False` is behaviorally unchanged for those inputs.

## Incremental review PRs
- #1153 — generic stream cancellation and SIGTERM lifecycle fixes
- #1154 — Qwen3-ASR SM89 profile, diagnostics, and malformed-input handling; blocked by #1176
- #1155 — pinned benchmark provenance, resource monitoring, stability harness, and results
- #1176 — correctness fix preserving audio beyond 30 seconds with `truncation=False`
- #1193 — proactive context-capacity validation and HTTP 400 error mapping

The incremental PRs are the intended review and merge path. This Draft preserves the complete validation diff in one place and should not be merged independently. It should be synchronized after #1176 and #1193 land.

## Test plan
- [x] `python -m pytest -q tests/unit_test/qwen3_asr tests/unit_test/benchmarks` — 50 passed
- [x] full SeedTTS EN and ZH, three measured repeats after warmup
- [x] 30-minute mixed-concurrency stability run
- [x] targeted 31-second, 60-second, and over-context RTX 4090 validation
[...]

PR #1153, opened 2026-07-23
fix(serve): stop pipeline workers after handled signals

wirybeaver:
## Summary

- consume Uvicorn's already-handled SIGINT/SIGTERM after HTTP shutdown instead of re-raising them before launcher cleanup
- preserve Uvicorn's main-thread signal registration, graceful/forced-exit state, and original-handler restoration
- exercise the production `_run_server()` composition and assert the fake pipeline runner is stopped after an actual handled SIGTERM

## Current-main rebase

The original transcription-stream cleanup hunks were dropped while rebasing. Merged PR #1080 now owns this lifecycle at the correct boundaries: ASGI response closure cascades through `Client` and active `Coordinator` streams abort and clean their request state. Retaining the older endpoint-local abort logic would duplicate that behavior.

This PR now owns only the remaining launcher/process lifecycle gap.

## Validation

```text
python -m pytest \
  tests/unit_test/serve/test_openai_api.py \
  tests/unit_test/pipeline/test_ipc.py::test_pipeline_uvicorn_server_consumes_handled_sigterm \
  tests/unit_test/pipeline/test_ipc.py::test_launcher_uses_runner_and_mounts_profiler_routes \
  tests/unit_test/pipeline/test_ipc.py::test_launcher_stops_runner_when_server_raises -q
58 passed

black --check sglang_omni/serve/launcher.py tests/unit_test/pipeline/test_ipc.py
isort --check-only sglang_omni/serve/launcher.py tests/unit_test/pipeline/test_ipc.py
git diff --check origin/main...HEAD
Passed
```

The SIGTERM regression was also run against `origin/main` with the production file restored and failed, then passed with this branch restored.

Original RTX 4090 validation: manually terminating the launcher released the ASR worker and GPU memory.

## Tracking

- Benchmark report: #1151
- Consumer GPU roadmap: #1120


Hayden727:
Linked to #1120. Patch comparison confirms `serve/launcher.py` and `tests/unit_test/pipeline/test_ipc.py` are copied exactly into both #1152 and #1171. This PR should remain the owner of lifecycle cleanup; those PRs should rebase/drop the copied hunks after this lands.

ChaoyuWang04:
Adding corroborating data from the MOSS-TD path (found during the RTX 5090 validation, #1189):

On main @ dc64a6b2, terminating `sgl-omni serve` (MOSS-Transcribe-Diarize, colocated single GPU) leaves the stage worker process alive holding ~29 GB of GPU memory. Reproduced 3×:

- SIGTERM to the launcher: uvicorn exits cleanly ("Application shutdown complete", "Finished server process") but the worker survives;
- Killing the launcher's process group (`setsid` launch + `kill -TERM -- -PGID`) also does not reap it — the worker appears to detach into its own process group;
- Only killing the worker PID directly (identified via `nvidia-smi --query-compute-apps`) releases the memory.

Since this PR fixes the shared `serve/launcher.py` lifecycle, it should cover this path as well — flagging the own-process-group detail in case it matters for the cleanup mechanism. Happy to re-verify on the 5090/MOSS-TD setup once this lands.

Hayden727 (approved):
Re-reviewed after the narrowed rebase: signal handling now exits the coordinator-owned process tree without duplicating stream lifecycle policy, focused lifecycle tests pass, and the branch is mergeable. The remaining gate is the queued CI run.

wirybeaver:
The Omni - CI Stage2 failure is a flaky performance-threshold miss, not a functional regression.
  - Final retry completed all 50 requests successfully.
  - Mean latency was 2.024 s, exceeding the 2.0 s threshold by 24 ms (1.2%).
  - Other final metrics passed:
      - Output token rate: 7.3
      - Mean RTF: 0.6497
      - Throughput: 7.461 req/s

PR #833, opened 2026-06-20
Extract Qwen3-ASR mel at true audio length instead of padding to 3000 mel frames

0xjeffro:
## Motivation

Qwen3-ASR reuses Whisper's feature extractor, which pads every audio clip to 30s (3000 mel frames) before computing the mel spectrogram. SeedTTS clips are only ~4–7s, so most of that CPU mel/FFT work runs on silence.

Profiling the c=32 path shows per-request preprocessing (request_build, ~28ms) is a big chunk of wall-clock and runs serially before the GPU (https://github.com/sgl-project/sglang-omni/issues/831#issuecomment-4756457401). But Qwen3-ASR's audio encoder is variable-length and already drops the padded frames via feature_attention_mask, so the 30s padding is pure waste.

## Modifications

Extract the mel at the clip's true length instead of padding to 30s

## Related Issues

#831

#### End-to-end benchmark

| metric | before | after | Δ |
|---|---|---|---|
| throughput (samples/s) | 22.43 | 27.88 | **+24.3%** |
| wall mean (s) | 48.52 | 39.04 | −19.5% |
| latency mean (s) | 1.416 | 1.144 | −19.2% |
| latency p95 (s) | 1.925 | 1.546 | −19.7% |
| rtf mean | 0.3064 | 0.2474 | −19.3% |
| corpus WER (mean of 3) | 0.01297 | 0.01310 | unchanged (noise) |
| skipped | 0 | 0 | — |

WER per-repeat:
before: 0.0129/0.0128/0.0132; after: 0.0138/0.0127/0.0128.

#### Stage breakdown (mechanism)

| interval | before avg | after avg | Δ |
|---|---|---|---|
| `scheduler_request_build_start→end` | 27.9 ms | 18.5 ms | **−34%** |
| `scheduler_request_build_start→end` p95 | 37.6 ms | 27.0 ms | −28% |
| `stage_input_received→stage_complete` (end-to-end, incl. queue) | 1561 ms | 1285 ms | −18% |

---

## Checklist

- [x] Format your code according with pre-commit.
- [ ] Add unit tests.
- [ ] Update documentation / docstrings / example tutorials as needed.
- [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed.
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.

## CI

CI runs on self-hosted GPU runners and requires a maintainer to add the
[...]

PR #1154, opened 2026-07-23
feat(qwen3-asr): add RTX 4090 serving profile

wirybeaver:
## Summary
- add a conservative BF16 single-RTX-4090 profile with bounded admission, static memory, and CUDA Graph batches
- keep SM89 multimodal attention on upstream auto-selection while preserving current-main's SM100+ Triton fallback, and log backend/memory checkpoints at startup
- return actionable HTTP 400 responses for undecodable audio without imposing an artificial 30-second duration limit

## Dependency

#1176 and #1193 have merged. This branch now includes their full-length feature extraction and proactive context-capacity behavior through current `main`; it has no remaining prerequisite PR.

## Test plan
- [x] `python -m pytest -q tests/unit_test/qwen3_asr`
- [x] startup on one RTX 4090 with BF16, FlashInfer LM attention, and upstream-selected Triton multimodal attention
- [x] SM89/SM100 multimodal-attention fallback and resolved decode CUDA Graph bucket regressions
- [x] typed invalid-audio versus operational-loader HTTP 400/500 regressions
- [x] targeted full-length and context-capacity validation on the RTX 4090 profile:
  - 31-second LibriSpeech input: 3/3 HTTP 200, identical 80-word transcript
  - 60-second LibriSpeech input: 3/3 HTTP 200, identical 146-word transcript
  - 124.91-second over-context input: 3/3 HTTP 400 before mel extraction (`1637 + 128 > 1636` tokens)
  - peak sampled GPU memory: 17,464 MiB; minimum free memory: 6,618 MiB
- [x] existing 30-minute validation soak retained at least 6.4 GiB free

The full SeedTTS sweeps and 30-minute short-clip soak were not repeated: all samples completed under the previous 30-second guard, and `truncation=False` does not change preprocessing for inputs within that window.

Resource measurements use `nvidia-smi`. Nsight Compute is installed on the Runpod image but kernel profiling is unavailable because the driver requires admin-only profiling and the container lacks `CAP_SYS_ADMIN`.

## Tracking
- Benchmark report: #1151
- Umbrella Draft PR: #1152
- Full-length preprocessing fix: #1176 (merged)
- Context-capacity follow-up: #1193 (merged)


sglang_omni/models/qwen3_asr/request_builders.py
```diff
@@ -142,8 +143,19 @@ def _build_prompt_ids(num_audio_tokens: int, language: str) -> list[int]:
...
+            raise ValueError(
+                "Qwen3-ASR could not decode the uploaded audio; provide a valid "
+                "audio file."
+            ) from exc
         audio_duration_s = float(len(audio) / _SAMPLE_RATE)
+        if audio_duration_s > _MAX_AUDIO_DURATION_S:
+            raise ValueError(
+                "Qwen3-ASR accepts audio up to 30.0 seconds; split longer audio "
+                "before inference."
+            )
```
lijrjyan: This encodes the wrapper's 30-second window as an API contract. The underlying problem on current main is that audio beyond 30 s is silently truncated — HTTP 200 with full-duration usage (#1173), and that isn't fixed yet. A hard reject here would freeze the wrapper artifact into the API instead of fixing the silent truncation: the encoder accepts variable-length mel input, and the effective bound is the request-token/KV budget, not 30 seconds.

sglang_omni/models/qwen3_asr/request_builders.py
```diff
@@ -142,8 +143,19 @@ def _build_prompt_ids(num_audio_tokens: int, language: str) -> list[int]:
 
     def request_builder(payload: StagePayload) -> Qwen3ASRRequestData:
         params = payload.request.params or {}
-        audio = _load_audio(_audio_source_from_payload(payload))
+        try:
+            audio = _load_audio(_audio_source_from_payload(payload))
+        except Exception as exc:
```
lijrjyan: The decode-error hardening is independently useful regardless of the 30-s question — a clean, actionable error for undecodable input.

tests/unit_test/qwen3_asr/test_request_builders.py
```diff
@@ -117,6 +118,50 @@ def test_qwen3_asr_request_builder_records_inclusive_audio_offsets(monkeypatch)
     )
 
 
+def test_qwen3_asr_request_builder_rejects_undecodable_audio(monkeypatch) -> None:
```
lijrjyan, line 358: If the 30-s reject is dropped per the discussion above, this boundary test should move to the actual capacity limit (token/KV budget) instead of the wrapper window.

lijrjyan (changes requested):
The 30-second boundary here presents a wrapper artifact as a model limitation. Qwen3-ASR's encoder natively accepts variable-length mel sequences, and the official inference stack supports much longer audio via chunking; the 30-s behavior on main comes from truncation=True in the feature extractor, which I've reported in #1173 (silent truncation returning HTTP 200 with full-duration usage). I have a fix enabling the native variable-length path instead. The decode-error hardening in this file is independently valuable either way — suggestion below: keep the profile and the decode guard, drop the hard 30-s reject.

sglang_omni/models/qwen3_asr/request_builders.py
wirybeaver: You're right. I traced this back to #833: `truncation=True` was added together with `padding=\"longest\"` to avoid padding short SeedTTS clips to Whisper's 3,000-frame window. The 30-second ceiling was inherited from `WhisperFeatureExtractor`; it was not established as a Qwen3-ASR model limit.

During the 4090 validation I observed deterministic truncation beyond 30 seconds and added this guard to turn silent data loss into an explicit 400, but I incorrectly promoted that preprocessing artifact into an API/model constraint. The Qwen3-ASR encoder already consumes the attention mask and supports variable-length mel input; the actual single-request bound is the configured context/token budget (with chunking as a separate option beyond that).

I agree that this hard reject should be dropped. The decode-error guard remains independently useful, while #1176 addresses preservation of audio beyond 30 seconds.

sglang_omni/models/qwen3_asr/request_builders.py
wirybeaver: Agreed. This decode-error handling is independent of the long-audio issue and should remain even after removing the 30-second guard.

tests/unit_test/qwen3_asr/test_request_builders.py
wirybeaver: Agreed. The 30-second boundary test should be removed with the hard guard. Long-audio coverage should instead verify that preprocessing preserves frames beyond Whisper's window, as in #1176; validation and client-facing handling of the actual context/token capacity can be covered separately.

Hayden727:
Linked to #1120. I checked the actual `qwen3_asr/stages.py` patch: this PR owns Qwen3-ASR model/profile work, while the `sm == 89 or sm >= 100` policy duplicates the shared architecture policy in #1199. Please rebase on #1199 after it merges, remove the local SM predicate, and retain only Qwen3-ASR-specific changes.

wirybeaver:
Rebase and resolve conflicts. Need to rerun the benchmark on sglang's RTX4090 this weekend.

sglang_omni/models/qwen3_asr/engine_builder.py
```diff
@@ -94,12 +97,37 @@ def generation_defaults(self, *, dtype: str) -> dict[str, Any]:
         }
         if self.mm_attention_backend is not None:
             defaults["mm_attention_backend"] = self.mm_attention_backend
-        else:
-            sm_version = get_visible_gpu_sm_version(self.gpu_id)
-            if sm_version is not None and sm_version >= 100:
-                defaults["mm_attention_backend"] = "triton_attn"
         return defaults
```
Hayden727, line 116: Removing this `else` also drops current main's SM>=100 contract: without an operator override, upstream auto-selects FA4 on SM100 while main deliberately forces `triton_attn`, and this PR has no B200 validation. Preserve the SM>=100 fallback (explicit overrides still win) and cover SM89 auto versus SM100+ Triton in the production builder test.

sglang_omni/models/qwen3_asr/engine_builder.py
```diff
@@ -94,12 +97,37 @@ def generation_defaults(self, *, dtype: str) -> dict[str, Any]:
...
+        super().validate_before_infrastructure(server_args)
+        logger.info(
+            "Qwen3-ASR runtime profile: dtype=%s attention_backend=%s "
+            "mm_attention_backend=%s cuda_graph=%s cuda_graph_bs=%s "
+            "torch_compile=%s max_running_requests=%s mem_fraction_static=%s",
+            getattr(server_args, "dtype", None),
+            getattr(server_args, "attention_backend", None),
+            getattr(server_args, "mm_attention_backend", None),
+            not getattr(server_args, "disable_cuda_graph", False),
+            getattr(server_args, "cuda_graph_bs", None),
```
Hayden727: `build_sglang_server_args()` removes the legacy `cuda_graph_bs` key and stores the resolved buckets at `server_args.cuda_graph_config.decode.bs`, so this logs `None` on real 0.5.16 startups (reproduced on `922ad181`) while the test fake masks it. Use `get_decode_cuda_graph_bs(server_args)` and make the regression expose only the nested field.

sglang_omni/models/qwen3_asr/request_builders.py
```diff
@@ -142,9 +142,15 @@ def request_builder(payload: StagePayload) -> Qwen3ASRRequestData:
...
             None if requested_language is None else resolve_language(requested_language)
         )
-        prepared = prepare_audio(
-            payload, source_name="Qwen3-ASR", target_sample_rate=_SAMPLE_RATE
-        )
+        try:
+            prepared = prepare_audio(
+                payload, source_name="Qwen3-ASR", target_sample_rate=_SAMPLE_RATE
+            )
+        except Exception as exc:
```
Hayden727: `prepare_audio()` also raises transport, resource, and internal failures; for example, a remote URL HTTP 503 is wrapped here and then classified as HTTP 400 by `transcriptions.py` (reproduced on `922ad181`). Translate only a typed invalid-audio/decode exception, preserve operational failures as 5xx, and add route-level 400/500 regressions.

Hayden727 (changes requested):
Current main is merged and Qwen3-ASR automatic-language behavior from #1403 is preserved. Three production-path issues remain inline. Local review validation: 13 pipeline tests and 79 request-builder tests passed in a CPU stub harness; changed-file hooks, py_compile, LSP diagnostics, lint, and docs are clean. The Qwen3-ASR GPU CI is queued. Please also refresh the stale Dependency section (#1176/#1193 are merged) and index the new profile in the Qwen3-ASR cookbook/tests catalog before merge.

sglang_omni/models/qwen3_asr/request_builders.py
Hayden727: Fixed in `33b3ae3e`: Qwen3-ASR now translates only shared `AudioDecodeError`. The audio boundary checks decoder readiness, preserves operational exception causes (including wrapped OOM), and uses PyAV validation only after a decoder `RuntimeError` to positively classify invalid bytes or paths. Regressions cover backend failure, wrapped OOM, corrupt uploads/paths, and route-level HTTP 400/500 mapping.

sglang_omni/models/qwen3_asr/engine_builder.py
Hayden727: Fixed in `33b3ae3e`: the startup profile now reads `get_decode_cuda_graph_bs(server_args)`. The production-path fake exposes only `cuda_graph_config.decode.bs`, and `caplog` asserts the resolved bucket list.

sglang_omni/models/qwen3_asr/engine_builder.py
Hayden727: Fixed in `33b3ae3e`: restored the current-main SM>=100 `triton_attn` fallback while preserving explicit overrides, with dedicated SM89/SM100/SM120 and override regressions. The CUDA-graph-focused integration test pins SM89 so it remains hardware-independent.

Hayden727 (approved):
All three production-path findings are addressed in `33b3ae3e`: current-main SM100+ fallback and explicit override precedence are covered, startup logs use resolved decode CUDA Graph buckets, and invalid media is separated from decoder/backend/resource failures for both byte and path inputs. Local focused validation: 17 pipeline tests, 81 request-builder tests, 4 audio-taxonomy tests, and 2 transcription-route tests passed; changed-file hooks and Python compilation are clean. GPU CI remains the merge gate.

PR #1158, opened 2026-07-23
[Perf][MOSS-TTS] Restrict MOSS-TTS Delay audio-state text sampling

CloudRipple:
## Motivation

During MOSS-TTS Delay audio generation, the text channel can only choose between
the generation-slot and delay-slot control tokens. The existing path still
carried the full text vocabulary through masking, top-k/top-p processing, and
seeded sampling on every audio decode step.

This PR reduces that audio-state text-sampling work while preserving:

- the existing full `LogitsProcessor` projection;
- the 32 audio-codebook projection and batched sampling path;
- full-vocabulary sampling rules outside audio generation;
- per-request temperature, top-k, top-p, and seed behavior;
- seeded sampling results based on the original global vocabulary token IDs.

## Modifications

- Track whether scheduled MOSS-TTS requests are currently inside an audio span
  and enable the compact path only when the whole active batch is in the audio
  state. Mixed or non-audio batches retain the original full-vocabulary path.
- Register the generation-slot and delay-slot token IDs on the model. After the
  normal text `LogitsProcessor` runs, select only those two columns for
  audio-state text sampling.
- Add a two-candidate sampling path that:
  - preserves greedy, temperature, top-k, and top-p behavior;
  - hashes the original vocabulary IDs when performing CUDA seeded Gumbel-max;
  - maps the selected local column back to the original global token ID;
  - avoids the general top-k host synchronization for the two-token case.
- Update audio start/end state transitions so the runner returns to the full
  text path after the audio span ends.
- Add tests covering compact audio-state sampling, non-audio fallback, state
  transitions, control-logit selection, greedy token-ID mapping, and CUDA seeded
  equivalence with the original full-vocabulary sampler.
- Rename local generation-parameter loop variables in `request_builders.py`;
  this is a behavior-neutral cleanup.

## Related Issues

N/A.

## Accuracy Test

No model weights, audio-codebook candidate set, or non-audio text candidate set
is changed. The CUDA equivalence test verifies that compact control-token
sampling produces the same seeded result as full-vocabulary sampling with all
other text tokens masked.

```text
.venv/bin/pytest -q tests/unit_test/moss_tts/test_pipeline.py
37 passed, 20 warnings in 12.14s
```

The saved WAVs from the final performance repetitions 2, 3, 5, 6, and 7 were
then evaluated directly, without regenerating audio:

- ASR: `Qwen/Qwen3-ASR-1.7B`, concurrency 32;
- speaker similarity: WavLM large with the SeedTTS fine-tuned checkpoint;
- perceptual quality: `balacoon/utmos`, scores in `[1, 5]`;
- five full 1088-sample evaluations per side;
- every WER, similarity, and UTMOS run evaluated `1088/1088` samples with zero
  skips.

| Quality metric | Base mean +/- sample SD | Current mean +/- sample SD | Current - Base |
| --- | ---: | ---: | ---: |
| Qwen3-ASR corpus WER | 1.3514% +/- 0.0437 pp | 1.3782% +/- 0.0528 pp | +0.0268 pp |
| WER excluding >50% samples | 1.2852% +/- 0.0438 pp | 1.3120% +/- 0.0529 pp | +0.0268 pp |
| WavLM speaker similarity | 68.8941 +/- 0.0521 | 68.9143 +/- 0.0427 | +0.0202 |
| UTMOS | 3.9474 +/- 0.0031 | 3.9436 +/- 0.0060 | -0.0038 |

No material WER or speaker-similarity regression was observed. The WER shift
is smaller than the per-side run-to-run standard deviation, while mean speaker
similarity is slightly higher. The UTMOS shift is also smaller than the observed
run-to-run variation. No subjective listening evaluation was run.

## Benchmark & Profiling

End-to-end SeedTTS EN generation benchmark:

- model: `OpenMOSS-Team/MOSS-TTS-v1.5`;
- dataset: `zhaochenyang20/seed-tts-eval-arrow`, all 1088 EN samples;
- endpoint: non-streaming `/v1/audio/speech`;
- GPU: NVIDIA A800-SXM4-80GB;
- concurrency / max running requests / CUDA graph max batch size: `16 / 16 / 16`;
- one warm-up request, unlimited request rate, `max_new_tokens=2048`;
- fresh service and independent TorchInductor cache for every run;
- comparison: arithmetic mean of Current divided by arithmetic mean of Base,
  minus one. Per-pair relative deltas are not averaged into the primary result.

| Metric | Base mean | Current mean | Current vs. Base | Ratio-of-means 95% CI |
| QPS | 2.9382 | 2.9858 | +1.620% | [+1.125%, +2.117%] |
| Mean latency | 5.4190 s | 5.3330 s | -1.587% | [-2.070%, -1.102%] |
| P95 latency | 6.5612 s | 6.4688 s | -1.408% | [-2.112%, -0.700%] |
| Mean RTF | 1.26898 | 1.24954 | -1.532% | [-2.042%, -1.020%] |
| Audio throughput | 13.1644 | 13.3784 | +1.626% | [+1.129%, +2.124%] |
| Output throughput | 261.52 tok/s | 265.76 tok/s | +1.621% | [+1.123%, +2.122%] |

No kernel-level profiler trace was collected; the table reports end-to-end
serving performance.

## Checklist

- [x] Add unit tests.

## CI

long as the label remains. Use `/tag-and-rerun-ci moss` to select the MOSS-TTS
CI model. Draft PRs are skipped even if labeled.

GitHub CI has not been run yet.

### DIFF
--- a/sglang_omni/models/moss_tts/model_runner.py
+++ b/sglang_omni/models/moss_tts/model_runner.py
@@ -7,6 +7,7 @@ from typing import Any
 
 import torch
 from sglang.srt.layers.sampler import multinomial_with_seed
+from sglang.srt.layers.utils.hash import murmur_hash32
 
 from sglang_omni.model_runner.base import ModelRunner
 from sglang_omni.models.moss_tts.request_builders import _INF_DELAY
@@ -16,6 +17,24 @@ _NEG_INF = float("-inf")
 _INT64_MAX = torch.iinfo(torch.int64).max
 _UINT64_MASK = (1 << 64) - 1
 _INT64_SEED_MASK = (1 << 63) - 1
+
+
+@torch.compile(dynamic=True)
+def _multinomial_with_seed_and_token_ids(
+    logprobs: torch.Tensor,
+    seed: torch.Tensor,
+    positions: torch.Tensor,
+    token_ids: torch.Tensor,
+) -> torch.Tensor:
+    """Seeded Gumbel-max using original vocabulary ids as RNG columns."""
+
+    seed = seed.to(torch.uint64)
+    hashed = murmur_hash32(seed, positions, token_ids)
+    noise = hashed.to(torch.float64) / torch.iinfo(torch.uint32).max
+    noise.log_().clamp_(min=torch.finfo(noise.dtype).min).neg_()
+    noise.log_().neg_()
+    noise.add_(logprobs.to(torch.float64))
+    return torch.argmax(noise, dim=1)
 
 
 class MossTTSModelRunner(ModelRunner):
@@ -147,15 +166,25 @@ class MossTTSModelRunner(ModelRunner):
         schedule_batch: Any,
         requests: list,
     ) -> None:
-        channel_logits = self._channel_logits_from_result(result, forward_batch)
+        datas = [sched_req.data for sched_req in requests]
+        is_audio = bool(datas) and all(data.is_audio for data in datas)
+        channel_logits = self._channel_logits_from_result(
+            result,
+            forward_batch,
+            is_audio=is_audio,
+        )
         n_vq = len(channel_logits) - 1
         if n_vq <= 0:
             raise RuntimeError("MOSS-TTS requires at least one audio codebook head")
         if not requests:
             return
 
-        datas = [sched_req.data for sched_req in requests]
-        rows = self._sample_rows(channel_logits, datas, n_vq=n_vq)
+        rows = self._sample_rows(
+            channel_logits,
+            datas,
+            n_vq=n_vq,
+            is_audio=is_audio,
+        )
 
         next_token_ids = rows[:, 0].contiguous()
         result.next_token_ids = next_token_ids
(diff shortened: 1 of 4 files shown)


sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -147,15 +166,25 @@ def _collect_moss_step(
         schedule_batch: Any,
         requests: list,
     ) -> None:
-        channel_logits = self._channel_logits_from_result(result, forward_batch)
+        datas = [sched_req.data for sched_req in requests]
+        is_audio = bool(datas) and all(data.is_audio for data in datas)
```
luojiaxuan, line 170: Please add a batch-size-2 regression test with one audio row and one text row. It should confirm this gate selects the full-vocab path while the audio row remains limited to gen-slot and delay-slot.

sglang_omni/models/moss_tts/model_runner.py
```diff
@@ -591,9 +701,18 @@ def post_process_outputs(
...
         for row_idx, sched_req in enumerate(scheduler_output.requests):
             req_output = outputs[sched_req.request_id]
-            if req_output.data is None or int(req_output.data) == eos_id:
+            if req_output.data is None:
+                continue
+            text_token_id = int(req_output.data)
+            if text_token_id == audio_start_id:
+                sched_req.data.is_audio = True
+            elif text_token_id == audio_end_id:
+                sched_req.data.is_audio = False
```
luojiaxuan, line 714: Please add a two-step batch test where one of two audio requests emits audio_end, then confirm the next step calls the head and sampler with is_audio=False.

luojiaxuan (commented):
Please cover these two batch-state transitions before merge.

sglang_omni/models/moss_tts/model_runner.py
CloudRipple: The new `test_moss_collect_step_mixed_batch_uses_full_text_path` covers a batch-size-2 mixed audio/text batch. It verifies that both the head and sampler receive `is_audio=False`, while the real `_sample_rows` still restricts the audio row to gen-slot/delay-slot and allows the text row to select a normal vocabulary token.

sglang_omni/models/moss_tts/model_runner.py
CloudRipple: The new `test_moss_audio_end_in_batch_uses_full_text_path_on_next_step` starts with two audio requests, makes one emit `audio_end`, verifies their states become `False` and `True`, and confirms that the next decode step calls both the head and sampler with `is_audio=False`.

Ratish1 (approved):

PR #1159, opened 2026-07-23
[Higgs-TTS] Isolate preprocessing and audio encoder from the AR engine

stephenkgli:
## Motivation

The Higgs TTS pipeline currently places `preprocessing`, `audio_encoder`, and `tts_engine` in the same `pipeline` process. This couples reference-audio preparation to the autoregressive engine's Python process and makes the two paths share the same process-level runtime resources.

## Modifications

This change moves `preprocessing` and `audio_encoder` into one dedicated process. Keeping these two stages co-located preserves their local handoff, while isolating both stages from the autoregressive engine.

## Accuracy Test

Not related

## Benchmark & Profiling

- Dataset: SeedTTS
- 256 warmup requests per case
- Concurrency: 96
- `max_running_requests=96`
- `cuda_graph_max_bs=96`
- `prefill_coalesce_requests=32`
- `prefill_coalesce_wait_ms=300`
- Compiled vocoder decode enabled
- `max_new_tokens=2048`

| Metric | main | this PR | Change |
|---|---:|---:|---:|
| Throughput | 19.328 QPS | 28.894 QPS | **+49.50%** |
| Mean latency | 4.816 s | 3.254 s | **-32.44%** |
| P50 latency | 4.954 s | 3.214 s | **-35.12%** |
| P95 latency | 6.183 s | 4.531 s | **-26.71%** |
| P99 latency | 6.863 s | 5.196 s | **-24.28%** |

## Checklist

- [x] Add unit tests.

## CI

if labeled.

### DIFF
--- a/sglang_omni/models/higgs_tts/config.py
+++ b/sglang_omni/models/higgs_tts/config.py
@@ -24,6 +24,10 @@ class HiggsTtsPipelineConfig(PipelineConfig):
     drives the AR loop on the sglang backbone with the precomputed embed
     pasted at ``-100`` placeholder positions; vocoder reverses the delay
     pattern and decodes to waveform via the higgs-audio-v2-tokenizer codec.
+
+    Preprocessing and audio_encoder share a dedicated process so reference
+    waveforms stay on the local handoff while their host work is isolated from
+    the autoregressive engine.
     """
 
     architecture: ClassVar[str] = "HiggsMultimodalQwen3ForConditionalGeneration"
@@ -41,13 +45,13 @@ class HiggsTtsPipelineConfig(PipelineConfig):
     stages: list[StageConfig] = [
         StageConfig(
             name="preprocessing",
-            process="pipeline",
+            process="tts_frontend",
             factory=f"{_PKG}.stages.create_preprocessing_executor",
             next="audio_encoder",
         ),
         StageConfig(
             name="audio_encoder",
-            process="pipeline",
+            process="tts_frontend",
             factory=f"{_PKG}.stages.create_audio_encoder_executor",
             factory_args={"device": "cuda"},
             gpu=0,
--- a/tests/unit_test/higgs_tts/test_pipeline.py
+++ b/tests/unit_test/higgs_tts/test_pipeline.py
@@ -2,6 +2,7 @@
 
 import base64
 import logging
+import pickle
 import queue
 from types import SimpleNamespace
 from typing import Any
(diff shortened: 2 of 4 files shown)


sglang_omni/models/higgs_tts/config.py
```diff
@@ -41,13 +45,13 @@ def mem_fraction_role_to_stage(cls) -> dict[str, str]:
...
             name="preprocessing",
-            process="pipeline",
+            process="tts_frontend",
             factory=f"{_PKG}.stages.create_preprocessing_executor",
             next="audio_encoder",
         ),
         StageConfig(
             name="audio_encoder",
-            process="pipeline",
+            process="tts_frontend",
```
JiaxinD, line 54: This adds a third CUDA context on GPU 0 (tts_frontend's compiled codec joins pipeline and vocoder) while declared fractions already sum to 0.98, and the audio_encoder's 0.03 is accounting-only since create_audio_encoder_executor never receives total_gpu_memory_fraction. Could you add before/after peak VRAM from the c96 run to confirm the extra context still fits?

sglang_omni/models/higgs_tts/config.py
```diff
@@ -41,13 +45,13 @@ def mem_fraction_role_to_stage(cls) -> dict[str, str]:
     stages: list[StageConfig] = [
         StageConfig(
             name="preprocessing",
-            process="pipeline",
+            process="tts_frontend",
```
JiaxinD, line 48: With audio_encoder to tts_engine now cross-process, write_payload re-pickles payload.request including the raw reference_audio on every request even though the encoder already consumed it into payload.data. Consider nulling request.inputs reference_audio after preprocessing (or a project_payload on this edge); not a blocker given the +49.5% result.

JiaxinD (changes requested):

sglang_omni/models/higgs_tts/config.py
stephenkgli: On the 97,887 MiB GPU, peak memory was:

- main-r1: 85,724 MiB
- main-r2: 86,126 MiB
- this PR-r1: 87,635 MiB
- this PR-r2: 87,643 MiB

The paired peak deltas are +1,911 MiB and +1,517 MiB. The worst observed remaining headroom is 10,244 MiB (~10.0 GiB).

sglang_omni/models/higgs_tts/config.py
stephenkgli: done

tests/unit_test/higgs_tts/test_pipeline.py
```diff
@@ -652,6 +653,93 @@ def make_payload(request_id: str, url: str) -> StagePayload:
...
+
+    encoded = base64.b64encode(b"\x00\x11" * 4096).decode("ascii")
+    scheduler = stages.create_preprocessing_executor("ckpt", num_codebooks=2)
+    payload = scheduler._fn(
+        StagePayload(
+            request_id="raw-audio",
+            request=OmniRequest(
+                inputs={
+                    "text": "hello",
+                    "references": [{"media_type": "audio/wav", "data": encoded}],
```
JiaxinD, line 689: This covers the nested `references` form only, so top-level `reference_audio` is untested even though `_preprocess` reads it directly. Dropping `"reference_audio"` from `_CONSUMED_REFERENCE_INPUT_KEYS` leaves every new test green while `{"text": ..., "reference_audio": {...}}` keeps re-pickling the media on both hops.

JiaxinD (commented):
Traced every consumer of `request.inputs` reachable after preprocessing (audio_encoder, tts_engine, vocoder, request builders, result and stream adapters, coordinator completion/error/abort, serving layer, and the `return_omni_rollout` path, which builds from `HiggsTtsState`) and none of them read the stripped keys, so the drop is safe; the `raw_inputs` written at `coordinator.py:391` is pre-preprocessing and `payload.data` is replaced wholesale anyway. The peak VRAM numbers answer my earlier thread, and the topology test passes locally. Two status notes: CI has never run here (`omni-ci-gate` fails on the missing `run-ci` label, so TTS CI is skipped), and the +49.5% table was measured at 36812f1d, one commit behind this head.

stephenkgli:
Closed as #1125 already covered the isolation

Issue #1150, opened 2026-07-23
[Qwen3-Omni Perf] Thinker prefill: batch the multimodal embedding merge, remove per-request host syncs


edwingao28:
## Why

`_inject_multimodal_embeds` (`sglang_omni/model_runner/thinker_model_runner.py:115-272`, called from the custom prefill forward at `:83`) merges encoder embeddings into the text sequence with a requests × modalities Python loop. Each iteration does:

- `mask.any()` — device→host sync
- `int(mask.sum().item())` — device→host sync
- `input_embeds[torch.where(mask)[0] + start] = ...` — per-request scatter

plus more `.item()` calls on the deepstack branch (per deepstack layer for merged image+video, `:197`). That is ≈5–10 forced GPU syncs per multimodal request on the prefill hot path — more than 10 in the merged image+video deepstack case — scaling linearly with batch size.

Target design: the scheduler provides placement metadata up front (moved to device `non_blocking`), per-modality masks are derived with tensor ops, and one batch-wide scatter merges everything — O(1) host syncs per batch instead of O(requests × modalities), writing into a persistent preallocated buffer.

## Scope

Rewrite the merge as one batch-wide operation: precompute placement metadata where it is already known (request build time), derive masks with tensor ops, single scatter per modality (or one combined), no per-request `.item()`/`.any()` on the hot path. Keep deepstack handling equivalent.

Out of scope: prefill CUDA-graph/compile coverage (this is its prerequisite, tracked separately on the roadmap).

## Approach

1. At request-build time, record per-request per-modality token spans (already derivable from `pad_values` + input ids on CPU).
2. In the runner, build batch-level index tensors once, `non_blocking` H2D, scatter all embeds in one call per modality.
3. Keep the `_omni_consumed` chunked-prefill bookkeeping on CPU metadata, not derived from GPU masks.

## Quant target + command

Host-side prefill time per mm request should drop measurably in a trace (nsys/torch profiler); E2E: TTFT p50 at c8/c16 with mm inputs, no regression elsewhere. Text bench per measurement rules.

## Verification

- Behavior class: behavior-preserving — logits identical on a fixed mm fixture before/after (same scatter positions, same values).
- Regression test: multi-request batch with mixed modalities (image+audio in one batch, audio-only in another request) produces identical merged embeddings vs the old path.

## Deps

None. Unblocks (does not include) thinker prefill graph coverage.

Part of #1022. Design and benchmark contract: #1018 (item 1.5).

[...]

PR #1160, opened 2026-07-23
[Qwen3-Omni Perf] Vectorize thinker MRoPE; document talker linearization gate (#1149)

### LINKED ISSUE #1149: [Qwen3-Omni Perf] Thinker: vectorize MRoPE position computation; skip multimodal positions on speech stages
## Why

Two related costs, both per-request at build time:

1. **Thinker prefill positions.** `_compute_mrope_positions` (`sglang_omni/models/qwen3_omni/request_builders.py:442-499`) delegates to the sglang-dep HF port `get_rope_index_qwen3_omni` (`sglang/srt/layers/rotary_embedding/mrope_rope_index.py:214`, pinned `sglang==0.5.12.post1`), which is loop-heavy: a per-mm-segment loop plus a per-token while-loop for interleaved audio-in-video, with several host syncs. Cost grows with prompt length; hits TTFT on long multimodal prompts.
2. **Speech-stage positions.** The talker prompt runs the same full mm-aware `get_rope_index` at request build (`request_builders.py:731-740`), then every talker attention layer collapses the `[3, seq]` positions to row 0 at runtime (`components/thinker_model.py:265-270`, cheap slice but repeated per layer). Whether the talker actually needs mm-shaped positions is an open question: its prompt carries the thinker's mm placeholder tokens and grids, so today `mrope_position_delta` ≠ 0 and is consumed by the decode-position path — replacing with `arange` + delta 0 changes positions unless equivalence is proven first.

Target design: build thinker position blocks with vectorized numpy (`np.indices` / `np.broadcast_to` / one `concatenate`), leaving only the interleaved audio-in-video merge as a scan. For speech stages, first determine whether linear positions are equivalent for the talker's rope config; only then short-circuit to `torch.arange(seq_len).expand(3, -1)`.

## Scope

- Vectorize the thinker MRoPE index computation for the qwen3-omni path (either a local fast path in request_builders or an upstream sglang contribution — the port lives in the pinned sglang dep, so decide in the PR).
- Speech stages: investigate whether talker positions can be linear (delta semantics vs the decode-position path); if proven equivalent, build them directly and stop collapsing per layer. If not, this half closes with the analysis.

## Approach

1. Port the vectorized block-building pattern (per-segment `np.indices`-style expansion, single concatenate); keep the HF port as a reference oracle.
2. Talker positions: prove or refute `arange`+delta-0 equivalence against the current path (positions AND `mrope_position_delta` consumption at decode); apply only if proven.
3. Differential test: for a corpus of mixed image/video/audio prompts (incl. interleaved audio-in-video), new positions == HF-port positions exactly.

## Quant target + command

Request-build CPU time for long mm prompts (measure directly); TTFT p50 at c8 with mm inputs. Behavior-preserving, so a perf-neutral result with exact parity is still landable as code health — but the target is measurable TTFT improvement on long prompts.

## Verification

- Behavior class: behavior-preserving — thinker position tensors bit-identical to the current path on the differential corpus; logits identical on an E2E fixture. The talker-linearization half must pass the same bit-identical bar (positions and deltas) or it does not land.

## Deps

None. Independent of the batched embed merge (different hot spot, same file neighborhood — coordinate to avoid conflicts).

Part of #1022. Design and benchmark contract: #1018 (item 1.6).

guozhihao-224:
## Summary
Closes / close https://github.com/sgl-project/sglang-omni/issues/1149 (RFC https://github.com/sgl-project/sglang-omni/issues/1018 item 1.6; tracker https://github.com/sgl-project/sglang-omni/issues/1022).
### A — Thinker: vectorize MRoPE index (lands)
- Add local fast path `get_rope_index_qwen3_omni_vectorized` in `sglang_omni/models/qwen3_omni/mrope_positions.py`.
- Build text/audio/bos/eos and vision `[t,h,w]` blocks with NumPy (`meshgrid` / `stack`); one `concatenate` at the end.
- Audio-in-video merge uses `lexsort` (bit-identical to the HF/sglang per-token while-loop).
- Wire `_compute_mrope_positions` to the vectorized path (no longer calls the loop-heavy pinned sglang port on the hot path).
- Differential tests vs `sglang.srt.layers.rotary_embedding.mrope_rope_index.get_rope_index_qwen3_omni` (image / video / audio / interleaved / mixed).
### B — Talker: linearization (analysis + safe gate only)
- **Refuted** for multimodal talker prompts: talker HF config has `mrope_section`; backbone uses `MRotaryEmbedding`. Prefill uses full `[3, seq]` mm positions; decode uses `position = seq_len + mrope_position_delta - 1` (`ForwardBatch._expand_mrope_from_input`).
- Counterexample (user image + assistant tts-pad): positions ≠ `arange.expand(3,-1)`, `delta ≠ 0` (e.g. `-12`), first decode position diverges → fails the issue’s bit-identical bar → **must not** short-circuit those prompts.
- Land only a **safe** short-circuit via `talker_can_use_linear_mrope`: no grids, or grids but no `vision_start` / `audio_start` in `input_ids` (then `multimodal_nums == 0` → already linear). Otherwise keep full mm index in `build_sglang_talker_request`.

## Verification
Environment: 1×H100，qwen3-text-server（thinker-only），MMMU-ci-50，modalities=["text"]，temperature=0

### Correctness

| Check | Result |
|-------|--------|
| Differential unit (`test_mrope_positions.py`) | **17 passed** (image/video/audio/AIV/mixed + talker gate) |
| E2E greedy text (4 MMMU samples) | **identical** vs local HF-port A/B; not a full logits dump. Bit-identical positions are the primary behavior bar. |

A/B used a **local-only** temporary switch back to sglang `get_rope_index_qwen3_omni`; that switch is **not** in this PR.

### Quant — request-build CPU

Workload | seq | oracle | vectorized | speedup
-- | -- | -- | -- | --
long_mixed_mm | 2707 | 0.442 ms | 0.242 ms | 1.83×
audio_in_video | 1782 | 14.82 ms | 0.200 ms | 74×

### Quant — TTFT p50 @ c8 with mm
n=32, concurrency=8 (local A/B as above):

| Arm | TTFT p50 | TTFT mean |
|-----|----------|-----------|
| oracle | 0.643 s | 0.745 s |
| vectorized | 0.610 s | 0.725 s |
| Δ | **−0.033 s (~−5%)** | |

Another run on the same setup saw ~−25%; treat TTFT as noisy. **Land bar remains parity.**

## Notes / non-goals
- Independent of batched embed merge (#1150); same file neighborhood — rebase carefully if both land.
- Does not remove per-layer positions[0] collapse: talker already uses MRotaryEmbedding, so that path is N/A.
- Talker mm linearization does not land; only the proven-safe text-like gate does.


sglang_omni/models/qwen3_omni/mrope_positions.py
```diff
@@ -0,0 +1,323 @@
...
+
+            elif (
+                min_ed == ed_vision_start
+                and current_input_ids[ed_vision_start + 1] == video_token_id
+            ):
+                assert video_grid_np is not None and second_per_grids_np is not None
+                grid_t = int(video_grid_np[video_idx, 0])
+                grid_h = int(video_grid_np[video_idx, 1]) // merge
+                grid_w = int(video_grid_np[video_idx, 2]) // merge
+                t_scale = float(second_per_grids_np[video_idx]) * float(
```
edwingao28: This changes the FP32 operation order and can change the final integer RoPE index. For a 25 FPS input, I get `10` from the original path and `11` here after the int64 cast. Please keep the original multiplication order and add a non-integer-time-scale test.

sglang_omni/models/qwen3_omni/mrope_positions.py
```diff
@@ -0,0 +1,323 @@
...
+                video_pos = _vision_pos_ids(st_idx, grid_t, grid_h, grid_w, t_scale)
+                blocks.append(_merge_audio_in_video(video_pos, audio_pos))
+                video_len = int(video_grid_np[video_idx].prod()) // (merge * merge)
+                st += text_len + bos_len + audio_len + video_len + eos_len
+                audio_idx += 1
+                video_idx += 1
+                remain_videos -= 1
+                remain_audios -= 1
+
+            st_idx = float(blocks[-1].max() + 1) if blocks else 0.0
```
edwingao28: This is not equivalent to the original AIV merge. The original advances from the last emitted chunk, while this uses the max of the entire merged block.

For grid [1, 56, 56] and audio length 100, the delta changes from -771 to -769. Please preserve the original st_idx semantics and add a regression test.

edwingao28:
@guozhihao-224 could you update your benchmark result? Upstream had changed a lot. Please also add ur test file into test/README.md

guozhihao-224:
@edwingao28 Updated after rebasing onto current main, with refreshed benchmarks and README.

### Takeaways
- CPU path clearly wins, especially synthetic AIV (~38×); real image-text ~1.56×, real long video+audio ~1.25×.
- TTFT: small stable win on MMMU (~5%); Video-AMME ~flat.
- Land bar remains parity (bit-identical), matching #1149.
- Talker full linearization stays refuted; only the safe gate lands.

### Environment
```
Hardware: 1×H200
Server: qwen3-text-server (thinker-only), Qwen/Qwen3-Omni-30B-A3B-Instruct
Stack: current main + this branch
```
### measured
#### Request-build CPU
Time only get_rope_index_qwen3_omni (oracle / pinned sglang HF port) vs get_rope_index_qwen3_omni_vectorized.
Warmup + median of repeated iters; assert bit-identical positions/deltas first.
- Synthetic: hand-built long_mixed / AIV token+grid tensors (stress the AIV while-loop).
- Real: preprocess mmmu-ci-50 / Video_AMME_ci with the production preprocessor → reuse input_ids + grids / audio lengths (no image/video encode in the timed region).

#### TTFT A/B
  Launch thinker-only server; alternate arms V→O→… on a clean GPU (oracle via local-only env switch, not landed).
Stream /v1/chat/completions, measure time to first content token.
  - MMMU: n=32, c=8, max_tokens=64, 3 rounds each.
  - Video-AMME: n=32, c=8, max_tokens=32, thinker-max-seq-len=32768, video_fps=2 / max_frames=128 / max_pixels=401408, 2 rounds each.

### Quant — request-build CPU (synthetic, 5 rounds; bit-identical)

Workload | seq | oracle | vectorized | speedup
-- | -- | -- | -- | --
long_mixed_mm | 2757 | 1.080 ms | 0.486 ms | 2.22×
audio_in_video (AIV) | 1780 | 9.665 ms | 0.254 ms | 38.1×

### Quant — request-build CPU (real MMMU image-text, n=32)
Dataset: zhaochenyang20/mmmu-ci-50 (first 32; image+text only, no AIV):

Metric | oracle | vectorized | speedup
-- | -- | -- | --
per-sample median | 0.253 ms | 0.163 ms | 1.56×
sum over 32 | 8.29 ms | 5.36 ms | 1.55×

Seq: min 154 / median 279 / max 5445.

### Quant — TTFT @ c8 (stream, temperature=0)
MMMU-ci-50 (n=32), clean GPU, 3×3 alternating:

Arm | p50 per round (s) | median p50
-- | -- | --
vectorized | 0.455 / 0.483 / 0.463 | 0.463
oracle | 0.486 / 0.516 / 0.462 | 0.486

Δ ≈ −0.023 s (~−4.8%, ~1.05×)

### Video-AMME (n=32), clean GPU0, 2×2 alternating:

Arm | p50 per round (s) | median p50
-- | -- | --
vectorized | 25.16 / 25.30 | 25.23
oracle | 25.27 / 25.29 | 25.28

Δ ≈ −0.057 s (~1.002×) — essentially parity (encode + long prefill dominate).

sglang_omni/models/qwen3_omni/mrope_positions.py
```diff
@@ -0,0 +1,329 @@
...
+                blocks.append(merged)
+                video_len = int(video_grid_np[video_idx].prod()) // (merge * merge)
+                st += text_len + bos_len + audio_len + video_len + eos_len
+                audio_idx += 1
+                video_idx += 1
+                remain_videos -= 1
+                remain_audios -= 1
+
+            # Note (guozhihao): AIV uses last-column max; oracle appends per column.
+            next_st = float(blocks[-1].max() + 1) if blocks else 0.0
```
edwingao28: For the audio-in-video branch, this full-block `max()` is immediately overwritten by the last-column `max()` below, so each request still scans the entire merged audio/video block for no result. Compute `next_st` inside the two branches, using only `[:, -1]` for AIV.

tests/unit_test/qwen3_omni/test_mrope_positions.py
```diff
@@ -0,0 +1,535 @@
...
+    thinker_config = SimpleNamespace(
+        vision_config=SimpleNamespace(spatial_merge_size=2, tokens_per_second=None),
+        image_token_id=IMAGE_TOKEN_ID,
+        video_token_id=VIDEO_TOKEN_ID,
+        vision_start_token_id=VISION_START_TOKEN_ID,
+        audio_token_id=AUDIO_TOKEN_ID,
+        audio_start_token_id=AUDIO_START_TOKEN_ID,
+        position_id_per_seconds=POSITION_ID_PER_SECONDS,
+    )
+    result = _compute_mrope_positions(input_ids, model_inputs, thinker_config)
```
edwingao28, line 315: This test only checks parity with the old oracle, so it would still pass if `_compute_mrope_positions` were switched back to the loop-heavy implementation this PR is meant to remove. Spy or monkeypatch `get_rope_index_qwen3_omni_vectorized` so the hot-path wiring is actually protected.

tests/unit_test/qwen3_omni/test_mrope_positions.py
```diff
@@ -0,0 +1,535 @@
...
+        get_rope_index_qwen3_omni_vectorized,
+        linear_mrope_positions,
+        talker_can_use_linear_mrope,
+    )
+
+    tokens = [1, 2, 3, 4, 5]
+    ids = torch.tensor([tokens], dtype=torch.long)
+    model_inputs = {"image_grid_thw": torch.tensor([[1, 4, 4]], dtype=torch.long)}
+    cfg = _thinker_config_ns()
+    assert talker_can_use_linear_mrope(ids.view(-1), model_inputs, cfg) is True
```
edwingao28, line 458: The positive gate is only exercised as a helper here, while the builder integration test covers only the false branch. Add a `build_sglang_talker_request` case with grids but no multimodal markers and assert that it installs linear positions without calling `_compute_mrope_positions`.

edwingao28 (commented):
left some comments @guozhihao-224

sglang_omni/models/qwen3_omni/mrope_positions.py
edwingao28: Fixed in 0f814efc — `next_st` now computed per branch; AIV uses only `[:, -1]`, no full-block scan.

tests/unit_test/qwen3_omni/test_mrope_positions.py
edwingao28: Fixed in 0f814efc — test now monkeypatches a spy on `get_rope_index_qwen3_omni_vectorized` and asserts it is called by `_compute_mrope_positions`.

tests/unit_test/qwen3_omni/test_mrope_positions.py
edwingao28: Fixed in 0f814efc — added `test_build_talker_request_uses_linear_mrope_without_mm_markers`: grids w/o markers → linear positions + delta 0, and `_compute_mrope_positions` is patched to fail if called.

yxs (approved):

PR #1161, opened 2026-07-23
[Qwen3-Omni Perf] Thinker prefill: batch multimodal embedding merge, remove per-request host syncs

### LINKED ISSUE #1150: [Qwen3-Omni Perf] Thinker prefill: batch the multimodal embedding merge, remove per-request host syncs

cr-gao:
## What

Rewrites the Thinker prefill multimodal embedding merge
(`_inject_multimodal_embeds`) so placeholder placement comes from CPU-side
metadata and the merge is a single batch-wide scatter, removing the
per-request device→host syncs on the prefill hot path. Fixes #1150.

## Why

The old merge is a `requests × modalities` Python loop where each iteration
forces device→host syncs:

- `mask.any()` — sync
- `int(mask.sum().item())` — sync
- `input_embeds[torch.where(mask)[0] + start] = ...` — per-request scatter

plus more `.item()` on the deepstack branch (one per deepstack layer in the
merged image+video case). That is 6 forced syncs per multimodal request, 18
with interleaved image+video deepstack, scaling with batch size.

## Approach

- `build_sglang_thinker_request` records per-modality placeholder positions
  (`_omni_mm_positions`) from the final, pad-substituted prompt, as CPU int64
  tensors. The runner derives and caches them from `origin_input_ids` when
  absent, so placement is never read from GPU masks.
- The merge accumulates every request/modality into one `index_copy_` scatter
  per batch: the index is built and uploaded once, instead of the old
  per-request scatters. Chunk filtering and the prompt→batch row shift are
  vectorized on the CPU-side position tensors, so host work does not grow with
  the multimodal-token count.
- Deepstack handling is equivalent: for interleaved image+video, the slot
  assignment comes from a single `torch.sort` permutation, and its index is
  built once and reused across layers. The visual-position mask is built from
  the same metadata.
- Chunked-prefill bookkeeping (`_omni_consumed`) stays on CPU and follows the
  same offsets as before.

Out of scope: prefill CUDA-graph / compile coverage (its prerequisite, tracked
separately on the roadmap).

## Verification

**Correctness**

- **Golden parity tests** (`tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py`)
  pin the previous behavior — mixed batches, `pad_values`, all deepstack
  branches, chunked prefill — building expected embeddings from an oracle
  independent of the implementation. `test_no_host_syncs_on_hot_path` asserts
  the merge never calls `Tensor.item` / `Tensor.any` / `torch.where`.
- **Builder test** (`test_pipeline.py`) checks `_omni_mm_positions` is recorded
  from the padded prompt.
- **Differential parity on a real H800**: a 6000-seed randomized old-vs-new
  harness (batch + chunked, all deepstack branches incl. interleaved
  image+video) confirms element-wise-identical `input_embeds` / deepstack /
  masks on device.
- **End-to-end logits A/B** on the thinker-only server (Qwen3-Omni-30B-A3B,
  H800): a fixed multimodal fixture (text, image+text, audio query, image+audio),
  greedy, produces byte-identical output between the old commit and this branch.

**Merge cost vs multimodal-token count** (H800 PCIe, 8 requests, hidden=2048,
bf16, 3 deepstack layers). The thinker runs with `chunked_prefill_size: 8192`,
so 8192 is the ceiling for a single prefill forward. At 32×32 px per thinker
token that is roughly 8 concurrent full-resolution images.

| mm tokens | deepstack | main | this PR | main syncs | PR syncs |
|---|---|---|---|---|---|
| 256 | single-modality | 2.265 ms | **0.808 ms** | 48 | **0** |
| 2912 | single-modality | 2.108 ms | **0.807 ms** | 48 | **0** |
| 8192 | single-modality | 2.108 ms | **1.005 ms** | 48 | **0** |
| 256 | interleaved img+vid | 7.834 ms | **1.925 ms** | 144 | **0** |
| 2912 | interleaved img+vid | 7.664 ms | **2.024 ms** | 144 | **0** |
| 8192 | interleaved img+vid | 7.825 ms | **2.311 ms** | 144 | **0** |

main is flat in the token count — its cost is the fixed sync count, not the
data. This PR is 2.1–4.1× faster across the whole range and stays nearly flat
(32× more tokens costs it ~25%). Run-to-run spread on this bench is a few
percent, so read the ratios rather than the third decimal.

Worth reading the commits in order here, because the first version of this
change did not hold up at the top of the range. Removing the syncs by
accumulating placement into Python lists made host work grow with the
multimodal-token count; at the 8192 ceiling that reached 2.513 ms, i.e. *slower
than main*. Keeping the positions as CPU int64 tensors and filtering them with
vectorized ops is what makes the merge flat. The sweep above is the check that
caught it.

**Memory**: the batched scatter holds one transient for the concatenated
sources, measured at exactly `n_mm_tokens × hidden × 2 B` — 32 MiB at the 8192
ceiling, from the caching allocator's pool. This is deliberate: one
`index_copy_` gives a kernel sequence independent of how many requests and
modalities a batch happens to carry, which per-segment writes would not, and
that property matters for the graph-capture goal below.

**End-to-end** (thinker-only text server, Qwen3-Omni-30B-A3B, H800,
`benchmark_omni_rollout_stress` driving an MMMU image sample with text output):

- **No regression** on total latency (`--max-tokens 256`): c8 ~3.99 s, c16
  ~4.68 s, old ≈ new.
- **TTFT** (`--max-tokens 1`, so total latency ≈ TTFT), 5 repeats per side:
  old and new are indistinguishable — steady-state median ~0.216 s (c8) and
  ~0.29 s (c16) for both, within run-to-run noise (±~5 ms; one cold-start
  outlier in the first new run is warmup, not the merge).
- **Server stage timing** (thinker stage, from the run's profiler events, 120
  requests/side): prefill stage ~58 ms median (forward-dominated), old ≈ new;
  request build unchanged within noise for the added position recording.

Why no visible E2E delta: the removed syncs are a fixed count per
(request × modality × deepstack layer), so on this fixture the saving is a small
constant prologue to a prefill whose ~58 ms is dominated by the 30B forward — it
stays below the E2E noise floor. The load-bearing perf evidence is the
deterministic sweep above.

Where this pays off: those per-request device→host syncs are exactly what
prevents capturing the thinker prefill in a CUDA graph (a graph cannot contain a
sync). Removing them is the prerequisite for prefill graph coverage — the
separate roadmap item where a larger E2E win is expected (see #1022). This PR is
behavior-preserving and enables that, rather than delivering the graph win
itself.

Open to review.

Part of #1022. Design and benchmark contract: #1018 (item 1.5).


cr-gao:
Thanks — all three addressed in 4b23489:

1. **`int(prefix_lens[i])` → implicit `Tensor.item()` when lens are CPU tensors.** Good catch, and it was worse than an `.item()`: `extend_seq_lens_cpu` / `extend_prefix_lens_cpu` can arrive as CPU tensors, and `pos += length` then made the offsets tensor-valued and corrupted the scatter indices (out-of-bounds). Both are now normalized to Python lists once up front. Added a contract test that drives the merge with CPU-tensor lens and asserts zero `item`/`any`/`where`.

2. **Merged image+video deepstack `joint` device.** `joint` is now allocated on the target `device`, so the scatter never mixes CPU/CUDA if deepstack embeds arrive on host. (The repeated per-layer `.to()` of the indices the comment also mentions was already hoisted out of the layer loop in e06b08be.)

3. **`non_blocking=True` on the list-derived index H2D.** Correct — pageable, so it was a no-op there; dropped it on the index tensors and kept it on the embed sources (which may be pinned).

Behavior parity re-verified: unit golden tests + a 6000-seed randomized old-vs-new differential harness still element-wise-identical.

tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py
```diff
@@ -0,0 +1,423 @@
...
+    out, _, _ = runner._inject_multimodal_embeds(fb, sb)
+
+    expected = _base_embeds(runner, fb)
+    expected[0:1] = image_embeds[1:2]
+    torch.testing.assert_close(out, expected)
+
+
+# ---------------------------------------------------------------------------
+# Host-sync contract (RED until the batch-scatter rewrite lands)
+# ---------------------------------------------------------------------------
```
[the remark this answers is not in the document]
cr-gao: Fixed in 87cfaeb2 — the section header now states the contract instead of the TDD state it was written in, and the module docstring drops the same forward-looking wording.

tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py
```diff
@@ -0,0 +1,420 @@
...
+regardless of how the scatter is built.
+
+test_no_host_syncs_on_hot_path additionally asserts the merge's contract:
+no ``Tensor.item`` / ``Tensor.any`` / ``torch.where`` on the hot path.
+"""
+from __future__ import annotations
+
+import types
+
+import pytest
```
edwingao28: seems pytest isn't used here, any reason for this import?

tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py
```diff
@@ -0,0 +1,420 @@
...
+    r._video_token_id = VIDEO_ID
+    r._audio_token_id = AUDIO_ID
+    return r
+
+
+def _req(input_ids, omni_model_inputs, *, is_chunked=0, consumed=None):
+    return types.SimpleNamespace(
+        origin_input_ids=list(input_ids),
+        omni_model_inputs=omni_model_inputs,
+        _omni_consumed=consumed,
```
edwingao28, line 47: **nit**: `_req` never sets `_omni_mm_positions`, so the fixtures fall through to the `origin_input_ids` scan in `_req_mm_token_positions`. Production never takes that path — `build_sglang_thinker_request` always sets it. Instrumenting `Tensor.nonzero` over `_mixed_mm_batch()` gives 12 calls; setting the positions first gives 0. `test_build_time_positions_take_precedence_over_prompt_scan` is the only test on the shipped path, and it is single-request, image-only, unchunked, no deepstack. The chunked, deepstack and mixed-batch parity tests all pin the fallback. Please set positions in `_req` and keep one explicit fixture for the scan.

tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py
```diff
@@ -0,0 +1,420 @@
...
+        calls["any"] += 1
+        return orig_any(self, *a, **k)
+
+    def counting_where(*a, **k):
+        calls["where"] += 1
+        return orig_where(*a, **k)
+
+    monkeypatch.setattr(torch.Tensor, "item", counting_item)
+    monkeypatch.setattr(torch.Tensor, "any", counting_any)
+    monkeypatch.setattr(torch, "where", counting_where)
```
edwingao28: **nit**: The counter patches `torch.where`, which the merge never calls. The merge calls `.nonzero(as_tuple=True)`. A GPU-resident bool mask reintroduced through `.nonzero()` would sync and still report `{"item": 0, "any": 0, "where": 0}`. Please patch `Tensor.nonzero` as well.

tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py
```diff
@@ -0,0 +1,420 @@
...
+    out2, _, _ = runner._inject_multimodal_embeds(fb2, sb2)
+
+    expected2 = _base_embeds(runner, fb2)
+    expected2[1:2] = image_embeds[3:4]
+    torch.testing.assert_close(out2, expected2)
+    assert req.omni_model_inputs is None
+    assert req._omni_consumed is None
+
+
+# ---------------------------------------------------------------------------
```
edwingao28: nit: comment style doesn't follow repo rules

tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py
```diff
@@ -0,0 +1,420 @@
...
+        for _ in range(4)
+    ]
+    return _batches(reqs)
+
+
+def test_no_host_syncs_on_hot_path(monkeypatch):
+    # Placement comes from CPU-side metadata, so none of the ops that force a
+    # device->host sync should appear.
+    runner = _runner()
+    fb, sb = _mixed_mm_batch()
```
edwingao28, line 432: **nit**: Both sync tests build from `_mixed_mm_batch()`, which sets only `image_embeds` and `audio_embeds`, so the deepstack block at `thinker_model_runner.py:209-264` never executes under the counter. Two of the three `.nonzero()` sites are in that block. Please run the counter over a deepstack fixture too.

sglang_omni/model_runner/thinker_model_runner.py
```diff
@@ -109,99 +109,127 @@ def requested_capture_hidden_mode_decode(self, schedule_batch: Any, requests: li
...
 
         embed_input_ids = forward_batch.input_ids.clamp(
             0, self._embed_tokens.num_embeddings - 1
         )
         input_embeds = self._embed_tokens(embed_input_ids)
 
+        # CPU tensors on some sglang paths; int(tensor[i]) per request would
+        # put a .item() back on the hot path.
         extend_lens = forward_batch.extend_seq_lens_cpu
+        prefix_lens = forward_batch.extend_prefix_lens_cpu
```
edwingao28, line 155: This unconditional access breaks the existing test_qwen_model_runner_and_code_predictor_tensor_contracts stand-in, which has no extend_prefix_lens_cpu (and its request also has neither origin_input_ids nor _omni_mm_positions), so the unit suite raises AttributeError before it can validate the merge. Please update that fixture with the new prefix/position metadata, or preserve compatibility in the helper.

edwingao28 (commented):
@cr-gao Left some review and please follow repo comment style, all comments should be self-contained, don’t explain how you did something or explain what this code block did, just explain why we did this. ALso please add your name for all of the comments lile `# note (your first name): `

tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py
cr-gao: Removed.

tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py
cr-gao: Fixed.

tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py
cr-gao: Fixed. `positions=False` opts out, and `test_prompt_scan_fallback_when_positions_missing` is the one fixture that still reaches the scan.

tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py
cr-gao: Added. It also showed the two `.nonzero()` calls in the interleave branch were unnecessary, inverting the sort permutation (`slots[visual_order] = arange`) yields the same slots, so the counter can assert `nonzero: 0` on the shipped path instead of a permitted count. A 6000-seed differential fuzz against b64c2f05 on an H800 is still element-wise identical.

tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py
cr-gao: Added `_deepstack_mm_batch()` and `test_no_host_syncs_on_deepstack_path`. Checked the counter is not vacuous while I was there; the same fixture without positions reports 12 `nonzero`.

sglang_omni/model_runner/thinker_model_runner.py
cr-gao: Fixed in the fixture rather than with a `getattr` fallback in the merge. A real `ForwardBatch` always carries the field and `build_sglang_thinker_request` always records positions, so a fallback would let the hot path tolerate a shape that does not occur. Say the word if you would rather keep the stand-in minimal and make the helper tolerant instead.

cr-gao:
Done, in 52b1203 on top of ee9ef1c. 
Re-validated on an H800 PCIe at the new head: 3000 + 3000 seed differential fuzz against b64c2f05 element-wise identical, `test_talker.py` + `test_thinker_mm_embed_merge.py` 87 passed / 1 skipped, and the merge-cost sweep re-run, table refreshed.

cr-gao:
@edwingao28 Friendly ping. All review comments resolved; could you take a look?

edwingao28:
@cr-gao Hey could you resolve file conflicts?

tests/unit_test/qwen3_omni/test_thinker_mm_embed_merge.py
```diff
@@ -0,0 +1,438 @@
...
+    r._audio_token_id = AUDIO_ID
+    return r
+
+
+def _req(input_ids, omni_model_inputs, *, is_chunked=0, consumed=None, positions=True):
+    req = types.SimpleNamespace(
+        origin_input_ids=list(input_ids),
+        omni_model_inputs=omni_model_inputs,
+        _omni_consumed=consumed,
+        is_chunked=is_chunked,
```
edwingao28: While resolving the current conflicts, please update these fixtures to match SGLang 0.5.16's request state: the merged runner now reads `req.inflight_middle_chunks`, but this fixture still creates and mutates `is_chunked`. After the rebase, the multimodal parity tests will raise `AttributeError` before reaching their assertions; `test_talker.py` has the same stale stand-in. Please replace both fixtures and their chunk-state mutations with `inflight_middle_chunks`.

edwingao28: resolved in https://github.com/sgl-project/sglang-omni/pull/1161/commits/f27123371fa1beddcb40f960c8024bda97ecc2c4

cr-gao:
Hi @edwingao28 , thx for the review and seems that you have helped merge the main. Anything I should do at this time?

yxs (approved):

edwingao28 (approved):

edwingao28:
> Hi @edwingao28 , thx for the review and seems that you have helped merge the main. Anything I should do at this time?

We have handled the fix, thanks for your contribution!

charliechenye:
@edwingao28    behavior change: 

test_text_only_prefill_attaches_live_embeddings_without_official_batch_mutation in test_thinker_prefill_contract.py.

assert request._omni_mm_positions is official_positions no longer passes.

Instead:

    # note(chenye): post #1161 releases per-request multimodal
    # placement/consumption state after the final prefill chunk.
    assert request.omni_model_inputs is None
    assert request._omni_consumed is None
    assert request._omni_mm_positions is None

PR #1164, opened 2026-07-23
[Qwen3-Omni Perf] Predictor attention: SDPA enable_gqa instead of materialized KV expansion

### LINKED ISSUE #1145: [Qwen3-Omni Perf] Predictor attention: use SDPA `enable_gqa` instead of materialized KV expansion
## Why

The talker code predictor runs a ~16-step AR loop per talker token. Its attention expands K/V to match query heads by materializing copies before SDPA:

- `sglang_omni/models/qwen3_omni/components/talker.py:55-63` — `_repeat_kv` does `expand` + `reshape`, which physically copies K and V (2×) each call.
- Call sites: `talker.py:709-710` (`_direct_self_attention`, the eager multi-token path used by the streaming predictor executor) and `talker.py:1392-1393` (`_predictor_cached_self_attention`, cached decode).

Per predictor step that is 5 layers × (K+V) = 10 copies; 16 steps ≈ 160 small copy kernels per talker token (predictor: `num_hidden_layers=5`, `num_code_groups=16` in `hf_config.py:323,338`; heads 16/8 so n_rep=2). The decode path is captured inside the talker decode CUDA graph, so launch overhead is hidden — but the copies still execute on every replay.

PyTorch SDPA supports GQA natively: `scaled_dot_product_attention(..., enable_gqa=True)` broadcasts KV inside the kernel with no materialization. The repo already uses it elsewhere:

- `sglang_omni/models/voxtral_tts/voxtral_tts_audio_generation.py:254`
- `sglang_omni/models/qwen3_tts/sglang_model.py:1255`

## Scope

Remove `_repeat_kv` from the qwen3_omni predictor attention paths and pass `enable_gqa=True` to SDPA. Both call sites (multi-token and cached decode). Nothing else.

## Approach

1. In both call sites, drop the `_repeat_kv` calls and add `enable_gqa=(num_heads != num_kv_heads)` to the SDPA call.
2. Delete `_repeat_kv` if no other caller remains.
3. Re-capture path needs no change — the graph captures whatever the eager path does.

## Quant target + command

Predictor replay wall time should drop by a small single-digit percent; E2E speech A/B at c8 per the measurement rules in #1018. Not gated on an E2E win — this is also a code-health fix aligning with the other TTS models.

## Verification

- Behavior class: behavior-preserving. Codec tokens must be bit-identical before/after on a fixed-seed run (SDPA math is unchanged; only KV layout handling differs).
- Existing predictor unit tests pass; add one asserting identical output with/without the flag on a GQA-shaped toy config if none exists.

## Deps

None. Independent of #1039 / #1043.

Part of #1022. Design and benchmark contract: #1018 (item 3.1).

EdwardZhang1108:
## Motivation

The talker code predictor runs a ~16-step AR loop per talker token. Its attention expanded K/V by materializing copies before SDPA (`_repeat_kv`, `expand` + `reshape`): 5 layers × (K+V) × 16 steps ≈ 160 small copy kernels per talker token, executed on every decode-graph replay.

PyTorch SDPA supports GQA natively (`enable_gqa=True` broadcasts KV in-kernel, no materialization); `voxtral_tts` and `qwen3_tts` already use it. This PR aligns the qwen3_omni predictor with them.

## Modifications

- `_direct_self_attention` (multi-token eager) and `_predictor_cached_self_attention` (cached decode): drop `_repeat_kv`, pass `enable_gqa=attn.num_heads != attn.num_kv_heads` (conditional: head counts are per-TP-rank and equal at attention-TP=16).
- Delete `_repeat_kv` (no remaining callers).
- Add two CPU unit tests asserting bitwise equality vs a materialized-KV reference on a non-degenerate 4q/2kv toy config (with 1 KV head a wrong group order or dropped flag is undetectable).

## Related Issues

Closes #1145. Part of #1022. Measurement contract: #1018 (item 3.1).

## Correctness validation

**Expected:** behavior-preserving — bit-identical codec tokens and waveforms (SDPA math unchanged, only KV layout handling differs).

### Function-level equivalence (CUDA)

*What / why:* the changed SDPA call itself, in isolation — verifies `enable_gqa` produces identical numbers to the materialized-KV form on both changed paths before any serving stack is involved.

*Result:* bf16 (the production dtype) bitwise identical under default dispatch. fp32 shows 1-ULP deltas under default dispatch but is bitwise identical with the SDPA backend pinned to MATH — backend routing (fp32 cannot take the flash kernel), not math. Recorded, not gating.

### Codec-token comparison

*What / why:* the predictor's actual output contract (16 RVQ codes per frame) through the real server, fixed-seed fully-greedy A/B (thinker and talker both greedy), codec frames dumped per request/frame. Both execution modes are covered because the changed code runs inside a captured graph in production and eagerly in fallback paths.

| Mode | Gated requests | Frames compared | Result |
| --- | --- | --- | --- |
| Talker decode CUDA graph ON | 8 | 419 (×16 codes) | bit-identical |
| Graph OFF + predictor decode graph disabled (true per-step eager) | 8 | 419 (×16 codes) | bit-identical |

A 4-request concurrent burst per arm added bucket>1 coverage (202 frames per side, frame counts match; coverage only — batch composition is scheduler-dependent).

*Result:* every frame identical in both modes. Note for reviewers: `--talker-cuda-graph off` does not disable the predictor's own decode graph; the eager arm used an instrumentation-side kill switch applied identically to both trees.

### Waveform comparison

*What / why:* #1018 requires waveform evidence in addition to codec tokens; byte-level WAV equality also closes out the code2wav stage downstream of the change.

| Waveform metric | Result |
| --- | --- |
| Gated WAV pairs per mode | 8/8 |
| Identical WAV SHA-256 hashes | 8/8 (both modes) |
| Identical across execution modes (graph on vs off) | yes — same SHA set |
| Maximum absolute PCM error | 0 (byte-identical files) |

### WER

*What / why:* #1018 requires WER/MOS evidence appropriate to the risk; run through the CI pipeline (same Qwen3-ASR router).

| WER metric | Baseline arm | PR arm |
| --- | --- | --- |
| Evaluated | 50/50 | 50/50 |
| Corpus WER | 1.77% | 1.60% |
| Median per-sample WER | 0% | 0% |
| Samples above 50% WER | 0 | 0 |
| CI threshold (≤ 2.13%) | pass | pass |

*Result:* both arms clear the CI gate. These are two independent c8 arms (not sample-paired); the matched-condition differential WER is zero by construction — the c1 comparison above produced byte-identical waveforms.

### Unit tests

Full `pytest tests/ -m "not benchmark"` green on the patched tree (2195+ passed) via the CI flaky wrapper. Three environment-dependent tests deselected with baseline-also-fails evidence (host-topology probes, unrelated to this diff).

## Benchmark & Profiling — predictor replay microbenchmark

**Expected:** the ~160 materialization kernels per talker token disappear; small single-digit % replay improvement (issue estimate).

*What / why:* a predictor-only harness isolates the changed loop so a few-percent delta is not diluted by the talker MoE backbone or sampling; the A/A preflight establishes the noise floor before any A/B claim; the profiler pass attributes the win to the exact kernels the issue predicted.

*Setup:* predictor-only harness at real shapes (5 layers, 16q/8kv, head_dim 128, bf16), real `code_predictor_forward` graph replay timed with CUDA events on 1× H100; A/A preflight, then 4 counterbalanced pairs (AB, BA, BA, AB); profiler in a separate pass.

### Results

A/A noise per bucket: 0.007% / 0.002% / 0.005% (gate ≤ 2%).

Replay speedup (PR vs baseline, per pair):

| Bucket | AB | BA | BA | AB | Mean |
| bs=1 | +4.01% | +3.90% | +3.90% | +3.92% | **+3.93%** |
| bs=2 | +3.87% | +3.80% | +3.78% | +3.79% | **+3.81%** |
| bs=4 | +3.71% | +3.73% | +3.73% | +3.75% | **+3.73%** |

### Kernel attribution

| Metric | Baseline | PR | Delta |
| --- | --- | --- | --- |
| Copy kernels per frame (eager) | 1857 | 1697 | **−160** |

*Result:* the delta is exactly the statically predicted 5 × (K+V) × 16 = 160, and SDPA dispatches to identical cudnn/flash bf16 kernels on both sides — no backend routing change, no new materialization. The speedup lands in the predicted single-digit range and is consistent across all pairs (spread < 0.2%).

## End-to-end paired comparison (c8, per #1018)

**Expected:** neutral — the issue does not gate this change on an E2E win; this run is the no-regression guardrail on the canonical serving topology.

*What / why:* full text→text+audio serving on the #1018 TP2 disaggregated profile at c8, measuring the four contract metrics (latency, RTF, throughput, TTFA). A/A preflight first: deltas are only meaningful against the measured noise floor.

*Setup:* TP2 disaggregated (`--thinker-tp-size 2 --gpu-thinker-tp 0,1 --gpu-talker 1 --gpu-code2wav 1`) on 2× H100, thinker `mem_fraction_static=0.55` / talker `0.32`, model pinned @ 26291f79, SeedTTS-50 EN voice-clone streaming (thinker greedy at temperature 0, talker at default sampling), c8; per arm 4 rounds × 50 requests (round 0 discarded as in-server warmup, rounds 1–3 averaged); A/A preflight, then 4 counterbalanced pairs (AB, BA, BA, AB). All arms completed 150/150 measured requests with zero failures. Full field manifest (seed, cache mode, prompt IDs, per-request output shape) in the artifacts gist below.

A/A noise: latency_mean 0.9%, rtf_median 1.8%, throughput_qps 1.1%, audio_ttfp_mean 0.7% — all four metrics promotion-capable (≤ 5%).

### Results

| Pair | Order | Baseline req/s | PR req/s | Baseline TTFA mean (s) | PR TTFA mean (s) |
| 1 | AB | 4.035 | 4.016 | 0.900 | 0.835 |
| 2 | BA | 4.031 | 4.019 | 0.840 | 0.815 |
| 3 | BA | 4.036 | 3.957 | 0.847 | 0.825 |
| 4 | AB | 4.093 | 4.111 | 0.810 | 0.846 |

Mean deltas (PR vs baseline): latency_mean +0.09%, rtf_median +0.09%, throughput_qps −0.57%, audio_ttfp_mean −2.15% (negative = faster first audio).

*Result:* every mean delta is inside the A/A noise band — E2E-neutral as expected. The TTFA trend is mildly favorable but within noise and not claimed. The defensible wins remain the −160 kernels/token and ~+3.8% isolated replay.

## Environment

| Item | Value |
| --- | --- |
| Baseline | `aa6f92c3` (origin/main) |
| Hardware | H100 NVL 94GB (strict pin; all comparisons paired on identical hardware) |
| Driver / stack | 580.95.05, torch 2.11.0+cu130, sglang 0.5.12.post1, flashinfer 0.6.11.post1 |
| Raw artifacts | https://gist.github.com/EdwardZhang1108/c48e28cee5ad762fc6ce34791be5dda7 |


edwingao28:
@EdwardZhang1108 The c8 A/B is presented in good shape , but its run record omits several fields that contract requires: memory fraction, concrete fixture/prompt IDs, seed, output length setting, cache mode/reset/prefix-hit state, and output shape. Please add or link a run manifest containing these fields so the no regression result is reproducible and contract complete.

EdwardZhang1108:
@edwingao28  Thanks — added a full run manifest to the artifacts gist (`run_manifest.json`, plus the per-request `e2e_aa0_generated.json`): https://gist.github.com/EdwardZhang1108/c48e28cee5ad762fc6ce34791be5dda7

Field-by-field for the c8 A/B:

| Field | Value |
| --- | --- |
| Memory fraction | thinker `mem_fraction_static=0.55`, talker `0.32`, code2wav default. (Host note: the CI `0.20` talker default is 80GB-SXM tuning; on this 94GB NVL host it gives a 2947-token talker KV capacity, below the 4096 `talker_max_new_tokens` contract, so it rejects every full-length request at admission — 0.32 restores headroom.) |
| Fixture / prompt IDs | `zhaochenyang20/seed-tts-eval-50-arrow`, split `en`, 50 samples, speaker `Ethan`, voice-clone (distinct ref audio per sample). All 50 `sample_id`s listed in the manifest. |
| Seed | Thinker `temperature=0` → greedy argmax, seed-independent. Talker: the benchmark sends only top-level `temperature`, so the talker runs at server defaults (`talker_temperature=0.9`, `talker_top_k=50`, `talker_top_p=1.0`) with the per-request seed derived deterministically from `request_id` via xxhash. |
| Output length | thinker `max_tokens=256`, `talker_max_new_tokens=4096` (default); observed 15 codec frames/request mean (733 total over 50). |
| Cache mode / reset / prefix-hit | Mode: sglang default radix cache on thinker (not disabled), `chunked_prefill_size=8192` + mixed-chunk. Reset: fresh server process per arm (cold start); the 4 rounds within an arm share one server (warm). Prefix-hit: voice-clone with distinct refs and no shared system prompt limits cross-request content-prefix reuse to the fixed chat-template prefix; within an arm rounds 1–3 replay the same 50 prompts (warm), and the discarded round 0 covers the cold-cache pass. |
| Output shape | 24 kHz mono WAV per request; per-request `sample_id / audio_duration_s / latency_s / is_success` for all 50 in `e2e_aa0_generated.json`. |
| Requests | 150 measured/arm (rounds 1–3 × 50), 150/150 succeeded, 0 failed, across all 10 arms. |

tests/unit_test/qwen3_omni/test_talker.py
```diff
@@ -722,6 +725,139 @@ def test_qwen_predictor_decode_graph_uses_tensor_device_when_current_device_diff
...
+        q, cached_k, cached_v, is_causal=False
+    )
+    attn_output = attn_output.transpose(1, 2).reshape(
+        batch_size, attn.num_heads * attn.head_dim
+    )
+    attn_output, _ = attn.o_proj(attn_output)
+    return attn_output.reshape(batch_size, 1, hidden_size)
+
+
+def test_qwen_predictor_direct_attention_gqa_matches_materialized_kv(
```
edwingao28: This file is over 2000 lines, we should probably split some of the test into somewhere else The repo guidance asks for a split past 2000 lines:

> Keep files concise. If a file exceeds 2,000 lines of code, split it into multiple smaller files (e.g., `scheduler.py`, `scheduler_output_processor_mixin.py`).

It was already over before you touched it, so this is not yours to fix on its own. But the two attention tests plus their reference helpers are a self-contained group and would move cleanly into something like `test_talker_attention.py`.

tests/unit_test/qwen3_omni/test_talker.py
```diff
@@ -722,6 +725,139 @@ def test_qwen_predictor_decode_graph_uses_tensor_device_when_current_device_diff
...
+    )
+
+    device = torch.device("cpu")
+    talker = _build_real_step_predictor_graph_talker(
+        device, num_heads=4, num_kv_heads=2
+    )
+    attn = talker.code_predictor.model.layers[0].self_attn
+    # note (EdwardZhang1108): kv heads > 1, else wrong GQA group order passes by broadcast
+    assert attn.num_heads != attn.num_kv_heads and attn.num_kv_heads > 1
+    predictor = object.__new__(Qwen3OmniMoeTalkerCodePredictor)
```
edwingao28: _direct_self_attention never touches self, which is why this shell works. Can we just make it a `@staticmethod`? Then `object.__new__` isn't needed here, and the test won't break with a confusing AttributeError if someone adds a self. reference later.

edwingao28 (commented):
@EdwardZhang1108  The change looks right to me. I checked the grouping semantics and it is great!

btw `qwen3_tts` has the structurally identical predictor cached-attention path and already migrated to `enable_gqa`, but as a duplicated if/else:

https://github.com/sgl-project/sglang-omni/blob/3f97d73bc10ad4048b5a202cf586457350cee7d5/sglang_omni/models/qwen3_tts/sglang_model.py#L1240-L1257

Your one liner is the better form of the same fix. Once this merges the two near-identical predictors will differ in style for no reason. Welcome to create a follow up pr for that!

nits:
1. The two reference helpers repeat about eight lines (qkv split, reshape, `repeat_interleave`, `o_proj`). Could share a helper.
2. Both new tests run on CPU with `_IdentityRotary`, so they cover the math backend only. That is the right scope for a KV-broadcast regression test, but it does mean the bf16 and CUDA-graph evidence lives in the PR description rather than in CI. A `skipif(not torch.cuda.is_available())` variant would lock it in.

tests/unit_test/qwen3_omni/test_talker.py
EdwardZhang1108: Done — moved both attention tests and their reference helpers to `test_talker_attention.py`. The shared toy-talker builder went to `tests/unit_test/fixtures/qwen_predictor.py` rather than into the new test file, since the CUDA-graph tests in `test_talker.py` also use it and that avoids one test file importing another.

tests/unit_test/qwen3_omni/test_talker.py
EdwardZhang1108: Done — it's a `@staticmethod` now and the test calls it through the class, so the `object.__new__` shell is gone.

EdwardZhang1108:
@edwingao28  Thanks for the review! Both nits addressed:
- The reference helpers now share `_project_q_kv` / `_materialized_sdpa`; the duplicated projection/SDPA lines are gone.
- Both tests are parametrized over `cpu-fp32` plus a `skipif` `cuda-bf16` variant.

I ran the suite on an H100 (80GB HBM3) in the meantime — the four attention cases pass with the `cuda-bf16` variants actually executing, bitwise (`rtol=0, atol=0`):
```
test_..._direct_attention_gqa_matches_materialized_kv[cpu-fp32]   PASSED
test_..._direct_attention_gqa_matches_materialized_kv[cuda-bf16]  PASSED
test_..._cached_attention_gqa_matches_materialized_kv[cpu-fp32]   PASSED
test_..._cached_attention_gqa_matches_materialized_kv[cuda-bf16]  PASSED
```
`test_talker.py` is 71 passed / 1 skipped after the split, and the full `pytest tests/ -m "not benchmark"` run is green (2239 passed, 30 skipped; 3 host-topology tests deselected — they fail on the baseline on that host too). Full log in the artifacts gist (`unit_tests_h100.txt`): https://gist.github.com/EdwardZhang1108/c48e28cee5ad762fc6ce34791be5dda7

I'll open a follow-up PR to migrate the qwen3_tts predictor's if/else to the same one-liner form.

edwingao28 (approved):

edwingao28:
failed ci doesn't seem relevant to this pr

PR #1167, opened 2026-07-23
[Qwen3-Omni Perf] Reduce talker per-frame host overhead

edwingao28:
## Motivation

The talker decode loop pays four per-frame host costs that stall the scheduler thread (overlap schedule is force-disabled when feedback is enabled, so every sync lands on the hot path):

- `next_token_ids.tolist()` is called twice per frame, each a blocking pageable D2H.
- `_emit_code_chunks_and_feedback` clones ~16 per-request output rows per step out of the CUDA-graph fixed-address buffers.
- `_write_feedback_buffers` uploads a row-index tensor via pageable H2D every frame, even in the dense steady state where rows == range(batch_size).
- Token readback has no pinned staging at all.

## Modifications

- `model_runner/base.py`: `_stage_token_ids` copies sampled token ids once per frame into a pinned ping-pong host buffer (`_pinned_pingpong` helper, async copy + event); `_finalize` resolves it into `ModelRunnerOutput.host_token_ids`.
- `talker_model_runner.py`: post-prefill and post-decode stage token ids through the new path; `_emit_code_chunks_and_feedback` takes two batched clones (`_output_codes[:bs]`, `_output_embeds[:bs]`) and hands out row views, instead of 2 clones per request.
- `_write_feedback_buffers`: dense batches slice-assign `feedback_buffer[:batch_size]` and skip the index-tensor H2D; sparse batches keep the existing indexed path.
- `omni_scheduler.py` / `output_processor.py`: `process()` accepts an explicit `host_token_ids` so the mixin's `.tolist()` is host-only; the device tensor still drives the input chain.
- Tests: row-ownership invariant suite (prep-to-emit routing, sparse rows, stale mask, 4 overrun-drop alignment cases through the real `_resolve_and_process`), plus unit gates for staging, snapshot survival, feedback fast path; registered in `tests/README.md`.

## Related Issues

Part of #1018 (item 3.4).

## Accuracy Test

Audio and text outputs are byte-identical to base (144919ab) on H100: SHA-matched on the full response payloads across steady c4 and mixed workloads. 16/16 gate tests pass on CUDA; full unit suites 0 failures.

## Benchmark & Profiling

**End-to-end A/B** on H100 with the default stage split (thinker GPU0, talker+code2wav GPU1). Arms alternated with a full server relaunch on every switch, 5 pairs, warmup discarded, identical 8-prompt set, text+audio at concurrency 8, 32 requests per trial, 0 errors in 20 trials.

Per-pair results (higher is better for xRT and req/s, lower for TTFA):

| pair | xRT this PR | xRT base | ΔxRT | req/s this PR | req/s base | Δreq/s | TTFA this PR | TTFA base |
| 1 | 26.81 | 24.45 | +2.37 | 0.616 | 0.548 | +0.068 | 0.726 | 0.882 |
| 2 | 17.99 | 22.21 | -4.22 | 0.404 | 0.495 | -0.091 | 1.471 | 1.004 |
| 3 | 26.01 | 23.91 | +2.10 | 0.616 | 0.545 | +0.071 | 0.631 | 0.723 |
| 4 | 26.23 | 24.02 | +2.21 | 0.606 | 0.542 | +0.064 | 0.674 | 0.759 |
| 5 | 21.00 | 12.80 | +8.20 | 0.497 | 0.289 | +0.207 | 0.732 | 6.147 |

Summary over the three clean mid-run pairs (1, 3, 4; pair 2 is a lone dissenter, pair 5 ran last with both arms degraded):

| metric (c8, text+audio) | base | this PR | delta |
|---|---|---|---|
| xRT | 24.13 | 26.35 | **+2.22 (~+9%)** |
| requests/s | 0.545 | 0.613 | **+12.4%** |
| TTFA p50 (s) | 0.788 | 0.677 | **-0.111** |

4/5 pairs favor this PR. Concurrency 1 shows no delta, as expected: these are host-side costs that only bind under batched load.

**Per-frame attribution** at bs=8 (torch profiler, H100) — the mechanism behind the E2E delta:

| metric | base | this PR |
|---|---|---|
| pageable token D2H / frame | 1.99 | 0.01 (1 pinned + event) |
| aten::clone / step | 17.99 | 5.96 |
| feedback H2D / frame | 15.0 | 13.1 |

Remaining target: 13 feedback-path H2D per frame survive; tracked as follow-up under #1018 item 3.4.

## Checklist

- [x] Add unit tests.

## CI

if labeled.


sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -932,6 +932,11 @@ def _make_batch_result(batch, mr_output):
         next_token_ids = batch.output_ids
         if isinstance(next_token_ids, torch.Tensor):
             batch.input_ids = next_token_ids.to(torch.int64)
+        # Note (wenyao): reuse the runner-staged pinned host copy so the mixin's
+        # .tolist() is host-only; the device tensor still drives the input chain above
+        host_token_ids = getattr(mr_output, "host_token_ids", None)
```
SandyLuXY: nit: `host_token_ids` is now a declared `ModelRunnerOutput` field, so treating an absent field as `None` hides a broken runner contract and silently restores the device-side `.tolist()` path this change is removing. Please read `mr_output.host_token_ids` directly. The value may be `None`, but the field should not be missing.

tests/unit_test/qwen3_omni/test_talker_row_ownership.py
```diff
@@ -0,0 +1,215 @@
...
+    def __init__(self, rid: str, finished: bool, retracted: bool = False) -> None:
+        self.rid = rid
+        self._finished = finished
+        self.is_retracted = retracted
+
+    def finished(self) -> bool:
+        return self._finished
+
+
+def _resolve_scheduler(result: SimpleNamespace) -> tuple[OmniScheduler, list]:
```
SandyLuXY, line 274: These cases stub `_run_batch_resolve` and only test the pre-existing scheduler filter; they never execute the changed token staging, batched snapshot, or feedback write while rows reorder, retract, or finish. Add a runner/scheduler regression that changes the active order/set across steps (for example `[r0, r1] -> [r1, r0] -> [r0]`) and proves each request keeps its own token, code chunk, feedback row, and mask state.

sglang_omni/scheduling/omni_scheduler.py
edwingao28: fixed, now it reads mr_output.host_token_ids directly

tests/unit_test/qwen3_omni/test_talker_row_ownership.py
edwingao28: added test_row_ownership_tracks_current_batch_order_across_steps driving [r0, r1] -> [r1, r0] -> [r0] through the real paths

yxs (approved):

PR #1172, opened 2026-07-24
ci: preserve benchmark result on cleanup timeout

LauraGPT:
## Summary
- print scoped `nvidia-smi` and `fuser` diagnostics when GPU memory cleanup times out
- keep a completed benchmark result from turning red solely because post-stage cleanup reports residual memory with no killable process
- downgrade only the timeout exit code (`1`); configuration and scope errors remain strict failures

This is the standalone `main`-targeted follow-up requested in the discussion on #1093 and supersedes the dependent, now-closed #1099.

## Verification
- `python3 -m pytest -q tests/unit_test/test_tune_ci_thresholds.py` (10 passed)
- dynamic post-stage policy coverage for cleanup exit codes 0, 1, and 2
- `pre-commit run --files .github/actions/omni-post-stage/action.yaml .github/scripts/delete_gpu_process.sh tests/unit_test/test_tune_ci_thresholds.py`
- `bash -n .github/scripts/delete_gpu_process.sh`
- `git diff --check`

### DIFF
--- a/.github/actions/omni-post-stage/action.yaml
+++ b/.github/actions/omni-post-stage/action.yaml
@@ -121,4 +121,16 @@ runs:
         # note (Yue Yin): post-stage runs after the test in an isolated --rm
         # container, so reap nvidia-invisible orphans (spawn workers) that would
         # otherwise hold VRAM and time out the cleanup wait. setup must NOT.
+        # Teardown must not override a completed stage's result when the runner
+        # reports residual memory but no killable process. Setup and retry
+        # cleanup stay strict; this post-stage hook records diagnostics and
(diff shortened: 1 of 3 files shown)


LauraGPT:
The current red `omni-ci-gate` is only the repository label gate: its failed step says `Missing required label run-ci`; every downstream Omni job was skipped before executing. The author-side `/tag-and-rerun-ci` command did not add the label. When convenient, could a maintainer add `run-ci` or run the command with maintainer permissions? Docs, lint, test-layout, all local pre-commit hooks, and the 10 focused tests are green.

PR #1174, opened 2026-07-24
[Fix][TTS][MOSS] Fix request memory leak in MOSS-TTS Local Transformer v1.5

Endebert:
### Observed problem

`OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5` voice-cloning requests caused request-correlated host-memory growth in a long-running SGLang-Omni server.

In a controlled workload at concurrency 16, worker RSS increased from about 6.41 GiB to 8.86 GiB over 320 successful requests. That is roughly 7.8 MiB of net RSS growth per request and makes the model unsafe to serve continuously under load.

During broader testing on the original 16 GiB host with no swap, the kernel OOM killer later terminated the main worker at about 13.1 GiB anonymous RSS. Retained completed requests were a confirmed contributor, although the available logs do not establish that they were the only contributor to that OOM.

Memray showed about 3.9 GiB of outstanding allocations originating in msgpack request deserialization. With the full 89.091-second voice-cloning reference, each retained request held roughly 12 MiB of scheduler data.

### Root cause

`OmniScheduler` creates this bidirectional reference for an active request:

```text
Req._omni_data -> request_data.req -> Req
```

Several completion, rejection, and abort paths removed the request from scheduler queues without breaking the cycle. The complete deserialized multimodal payload therefore remained retained until cyclic GC collected it.

## Modifications

Release terminal request state explicitly when the scheduler no longer owns the request:

- Attach `_omni_data` only after the request passes admission and is still live under the admission lock.
- Clear the `Req._omni_data` side when the request finishes, aborts, leaves the waiting queue, or is removed from a batch.
- Preserve `request_data.req`, because already-launched async lookahead snapshots hold `request_data` directly and still require that link.
- Extend scheduler regression coverage for rejected admission, queued and running aborts, adapter errors, and batch failures.

The leak was reproduced with MOSS-TTS Local Transformer v1.5 because its voice-cloning payload makes the retained memory especially visible. The cleanup belongs in `OmniScheduler`, however, because terminal request ownership is scheduler-wide rather than model-specific.

## Related Issues

None found.

## Accuracy Test

This does not modify model execution, sampling, or request data before terminal cleanup.

```text
29 passed, 21 warnings in 15.58s
```

Command:

```bash
pytest -q tests/unit_test/pipeline/test_scheduler.py
```

## Benchmark & Profiling

Measured on an RTX 6000 Ada 48 GB test server with 32 GB host RAM, `OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5`, and the full 89.091-second voice-cloning reference:

- Memray showed about 3.9 GiB of outstanding allocations originating in msgpack request deserialization. Object inspection traced the retained payloads to the `Req._omni_data` / `request_data.req` cycle.
- Before the cleanup, worker RSS grew from about 6.41 GiB to 8.86 GiB over 320 requests at concurrency 16.
- With this cleanup, all 320 requests succeeded. After the initial CUDA/allocator warm-up, anonymous memory stayed within about 12 MiB for the remainder of the run, and VRAM also stabilized.

This change addresses host memory retained by terminal requests. It does not bound active generation length or change disconnected-client cancellation, and it is not expected to change model latency or throughput.

## Checklist

- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed. (No user-facing change; the async-snapshot invariant is documented on the helper.)

## CI

### DIFF
--- a/sglang_omni/scheduling/omni_scheduler.py
+++ b/sglang_omni/scheduling/omni_scheduler.py
@@ -54,6 +54,13 @@ _ABORTED_REQUEST_ID_LIMIT = 10000
 _ABORTED_REQUEST_ID_RETAINED = 5000
 
 
+def _break_request_data_cycle(req: Any) -> None:
+    """Detach completed Reqs while preserving data.req for async snapshots."""
+    data = getattr(req, "_omni_data", None)
+    if data is not None and getattr(data, "req", None) is req:
+        req._omni_data = None
+
+
 class _NoOpSender:
     """Stub for send_to_detokenizer — stream_output handles emission."""
 
@@ -714,7 +721,6 @@ class OmniScheduler:
             self._pending_stream_done.discard(req_id)
         self._deferred_request_payloads.pop(req_id, None)
         req = req_data.req
-        req._omni_data = req_data
         req_id = req.rid
         if bool(getattr(req_data, "enforce_request_limits", False)):
             error_msg = self._prepare_request_limits(req_data)
@@ -738,6 +744,7 @@ class OmniScheduler:
         def enqueue_if_live() -> None:
             if req_id in self._aborted_request_ids:
                 return
+            req._omni_data = req_data
             _emit_event(
                 request_id=req_id,
                 stage=None,
@@ -1026,6 +1033,7 @@ class OmniScheduler:
                         )
                 self._first_emit_done.discard(rid)
                 self._prefill_start_done.discard(rid)
+                _break_request_data_cycle(req)
                 continue
 
             # Build result payload from the Req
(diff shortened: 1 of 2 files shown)


Endebert:
Please note that investigation and implementation was done by GPT5.6 Sol, therefor I'd recommend a careful look over the changes, as I lack the expertise to confirm its correctness with confidence.

Endebert:
Closing as superseded by #1183. That PR carried the equivalent scheduler lifecycle fix via `fa2fd585` and merged it in `a8d3dd1`, with additional atomic terminal-claim, error-path, and cyclic-GC collectability coverage. Current `main` retains `_detach_request_data` across completion, abort, and batch-removal paths, so rebasing this conflicting duplicate would not add anything.

PR #1175, opened 2026-07-24
[Feat] Process replica

### LINKED ISSUE #1131: [Qwen3-Omni Perf] Runtime-level stage replicas: logical stages with N transparent replicas
## Summary

Add first-class replica support to the pipeline runtime: a stage declares `num_replicas: N` and the runtime spawns N replica processes, routes each request to one replica (sticky for the request's lifetime), and keeps everything transparent to model code — model configs, route hooks, and schedulers keep referring to the logical stage name. Replica counts are fully independent per stage; no pairing or ratio constraints. 

First user: the Qwen3-Omni speech pipeline, with profile and benchmark speech throughput stops scaling at concurrency = 8, at that point a request spends 1.54ss in talker_ar and 1.27s in code2wav but only 0.48s in the thinker, which still runs cleanly at concurrency = 48. The speech stages are the bottleneck; replicating them is the direct answer.
Target: **≥1.5× speech throughput at concurrency ≥ 16** with 2 replicas of each speech stage.

## Motivation

Benchmark facts (Qwen3-Omni speech, default placement — thinker on GPU 0, `talker_ar` + `code2wav` on GPU 1):

- Speech throughput stops scaling at concurrency = 8; the speech stages are the bottleneck.
- The talker runs with `disable_overlap_schedule=True` and no async decode (`talker_scheduler.py`), so host-side work serializes into every decode step; one instance caps at `max_running_requests=32` with degrading per-step latency.
- `Code2WavScheduler` is fully serialized across requests: one inbox thread runs the vocoder synchronously per 10-frame window with a blocking D2H copy per call (`code2wav_scheduler.py:235-260`); no cross-request batching.

Why solve this in the runtime rather than in the model layer (per-model named replica stages + a shared hash function):

- **Transparency**: model code keeps writing `"talker_ar"`. No literal-stage-name call sites to hunt down per model — the class of bug where one missed send site silently splits a request across replicas cannot exist, because resolution happens at a single point in the runtime send path.
- **Reusability**: any pipeline (higgs_tts, moss, fishaudio, future models) gets replicas by setting one config field, with any per-stage count.
- **Model validation untouched**: `Qwen3OmniPlacementPolicy` validates the 8 logical stage names (`placement.py:76-84`); with runtime-level replicas the logical view is preserved and that policy needs no change.

## Architecture today (what the design must respect)

- One stage = one OS process, spawned generically from `config.stages` (`pipeline/mp_runner.py`, `pipeline/stage_workers.py`); per-process `CUDA_VISIBLE_DEVICES` remap, per-GPU startup locks, per-stage relay already exist.
- **One name = one endpoint**: routing resolves a stage name to exactly one ZMQ endpoint (`mp_runner.py:62`; `stage/runtime.py:986,1173`). This is the assumption to generalize.
- Stages route to each other **directly** (decentralized); the `Coordinator` (`pipeline/coordinator.py`) only submits at the entry stage and tracks terminal completions — but it is the **single admission point** every request passes through, which the design exploits.
- Sticky state per request lives inside stage processes (talker: feedback/text queues, decode-embed history, KV; code2wav: `_code_chunks`/`_emitted`/`_audio_chunks` keyed by `request_id`) — all messages of a request must reach the same replica.
- The talker has **two independent upstreams** (mm_aggregate prefill latch via `route_fn`/`next`; thinker hidden-state stream via `stream_to`/`stream_done_to_fn`) — replica choice must be consistent across senders without inter-sender coordination.
- Cross-GPU per-chunk streaming is already exercised today: the thinker (GPU 0) streams hidden-state chunks to the talker (GPU 1) through the relay every chunk. A cross-GPU talker→code2wav codec stream (far smaller payloads) is the same pattern, not a new one.

### Layer 1 — Schema + process expansion

- `StageConfig` gains `num_replicas: int = 1` and `replica_devices`. Device semantics compose with TP from day one (needed later for thinker-class stages): `replica_devices` either enumerates the full pool (`num_replicas × tp_size` ids, sliced per replica) or a single per-replica template (`tp_size` ids, offset per replica) 
- `mp_runner._build_stage_groups` expands each logical stage into N replica processes at spawn: each replica gets its own endpoint (`stage_<name>_r<i>.sock`), its own `CUDA_VISIBLE_DEVICES` remap, its own relay — all existing per-stage mechanics, instantiated per replica.
- The flat `stage_endpoints = {name: endpoint}` map becomes a registry: `logical_name → [ReplicaInfo(replica_id, endpoint, gpu)]`. `num_replicas = 1` (the default) degenerates to exactly today's behavior.

### Layer 2 — Routing: bind at admission, propagate on the envelope

- **Coordinator binds once per request at submit time.** For every replicated logical stage it independently selects a replica (default policy: round-robin per stage; the policy is a pluggable interface — see Phase C) and produces `replica_bindings: {"talker_ar": 1, "code2wav": 0}`.
- **Bindings ride the message envelope**, not the payload: a new optional field on the proto envelope (`SubmitMessage` / `DataReadyMessage` / `StreamMessage`), forwarded automatically by the runtime on every outgoing message for that request. It survives `project_payload` untouched and reaches every sender with the request's first message — no runtime RPC, no shared binding table, no coordination between senders.
- **Send-path resolution**: `stage/runtime.py` `_send_to_stage` / `_target_result` resolve `(logical target, request's binding) → replica endpoint`. `route_fn` / `stream_to` return values keep being validated against **logical** names.
- **No pairing/ratio constraints.** Each stage's count and binding are independent. Note that with equal counts and per-stage round-robin, bindings align in practice anyway (every audio request advances both counters in lockstep), so equal-count deployments get same-index "pairing" for free without the runtime knowing the concept.

This keeps selection stateful (real load balancing is possible) while propagation stays coordination-free — a request's replica choice is computed exactly once, at the one place every request already passes through.

### Layer 3 — Coordinator completion + operations

- **Terminal resolution by logical name**: `_handle_completion` normalizes the completing replica's name back to its logical stage **before** the inactive-terminal filter (`coordinator.py:533-542`) and the multi-terminal collection (`coordinator.py:565-577`). Without this, a completion from `code2wav_r1` is silently dropped (debug log only) and the request hangs — this is fixed once in the runtime for all models.
- Abort is already a broadcast — covers all replicas unchanged.
- Admin ops (`_resolve_admin_stages`) fan out to all replicas of a logical target; metrics/log lines carry `(logical_name, replica_id)`.

## Qwen3-Omni enablement

```yaml
stage_overrides:
  talker_ar:
    num_replicas: 2
    replica_devices: "1,2"
  code2wav:
    num_replicas: 2
    replica_devices: "1,2"
```

Plain per-stage counts — no pairing declaration. Placement: thinker on GPU 0; talker_ar_r0 + code2wav_r0 on GPU 1, talker_ar_r1 + code2wav_r1 on GPU 2 (with equal counts and round-robin, a request's talker and code2wav land on the same index — and hence the same GPU — in practice). The ratio is a pure deployment knob: if measurements later show code2wav lagging (its serialized loop makes this plausible at high per-replica batch), bumping only `code2wav.num_replicas` is a config change. Memory reference: the colocated H200 profile budgets talker_ar at 0.123 and code2wav at 0.014 of one card (`examples/configs/qwen3_omni_colocated_h200.yaml:32-40`) — one replica of each per dedicated GPU leaves ample room to raise the talker's KV budget. The colocated (single-GPU) variant rejects `num_replicas > 1` explicitly.

## Work items (phased)

**Phase A — mechanism (runtime):**
1. Schema: `num_replicas` / `replica_devices` (pool + template modes, TP-composable) on `StageConfig` (`config/schema.py`).
2. Expansion: replica-aware `_build_stage_groups` + `StageGroup.spawn` (`mp_runner.py`, `stage_workers.py`); endpoint registry replacing the flat name→endpoint map; `StagePlacementPlanner` / `config/topology.py` expand replicas while exposing the logical view to model placement policies.
3. Envelope: optional `replica_bindings` field on the message proto; automatic propagation in the stage IO shell.
4. Binding: coordinator assigns per-stage bindings at submit through a pluggable policy interface (ship round-robin); completion normalization to logical names; admin fan-out.
5. Send path: logical-name + binding resolution in `_send_to_stage` / `_target_result` (`stage/runtime.py:986,1173`).

**Phase B — Qwen3-Omni enablement + benchmark:**
6. Example YAML/CLI for per-stage `num_replicas` on `talker_ar` / `code2wav`, with per-replica memory fractions.
7. Integration test: audio requests covering **every replica of both stages** end-to-end (round-robin makes coverage deterministic); unit tests for envelope propagation and binding stability.
8. A/B benchmark (below).

**Phase C (follow-up, not required to land A/B):** additional binding policies behind the Phase A interface — least-inflight; cache/session-affinity (prerequisite for ever replicating the thinker, whose prefix cache makes per-request round-robin wasteful); replica health tracking (stop binding to a dead replica).

## Validation

- **Default-path regression**: with `num_replicas = 1` everywhere, behavior and performance must be identical for all pipelines (the envelope field is absent, the registry degenerates to one endpoint per name). This gates Phase A.
- A/B benchmark: baseline (1× each) vs 2× each on separate GPUs, concurrency 8 / 16 / 32.
- Metrics: req/s, audio TTFP, audio RTF.
- Acceptance: **≥1.5× throughput at c ≥ 16**; no TTFP/RTF regression at c = 8; audio output bit-identical (or within tolerance) to single-instance for the same request.

## Risks

- **Blast radius**: this touches runtime code shared by every pipeline (mp_runner, stage IO shell, proto, coordinator). Mitigations: `num_replicas = 1` default must take exactly the existing code path; the envelope field is optional; full regression across the other model pipelines gates the merge.
- **Envelope/proto compatibility**: adding a field to the message envelope must not break relay backends (`nixl`/`nccl`/`shm`) or same-process local dispatch; audit `_send_to_stage` serialization and `LocalStageDispatcher`.
- **Cross-GPU talker→code2wav edge**: with unequal counts (or non-round-robin policies) a request's talker and code2wav can land on different GPUs, adding per-chunk relay latency on the hottest edge. Watch TTFP/RTF in the A/B; if it ever becomes material, same-index/same-GPU affinity can be added later as an optional binding-policy optimization — it is deliberately not a Phase A concept.
- **Coordinator as binding authority**: binding is O(#replicated stages) work at submit — negligible, and the coordinator is already on the submit path; it does not join the per-step data path.
- **Requires additional GPUs** for the Qwen3-Omni 2×2 layout (one extra card per additional replica set).

## Out of scope

- Cross-replica work stealing; migrating an in-flight request between replicas
- Same-GPU / same-index affinity constraints (revisit as an optional binding policy only if A/B data shows the cross-GPU edge matters)
- Dynamic scaling (adding/removing replicas at runtime); replica restart/failover beyond "stop binding new requests" (Phase C)
- Replicating the thinker (the mechanism supports it — schema composes replicas with TP, and the binding-policy interface is where its cache-affinity requirement will plug in — but it is not enabled or validated in this issue)
- Replicas in the colocated single-GPU variant
- In-process code2wav parallelism (cross-request chunk batching, multi-stream, async D2H) — replicas route around its single-thread serialization but don't fix it; separate issue

### LINKED ISSUE #1307: [Tracking][Pipeline] Process-level replicas
## Summary

This issue tracks the implementation and productionization of process-level replicas in the SGLang-Omni pipeline runtime.

## Motivation

Replicating the entire pipeline applies the same data-parallel configuration to every process, even when only one part of the pipeline is saturated. This can duplicate unnecessary model state and leave CPU cores or GPUs underutilized. Process-level replicas provide finer-grained data parallelism by allowing each process group to scale independently according to its workload and available resources.

## Existing work

- #1131 defines the original runtime-level stage-replica design and the Qwen3-Omni performance target.
- #1175 implements stage-level replica expansion, request-sticky round-robin binding, binding propagation, logical target resolution, and the initial Qwen3-Omni deployment.
- #1262 is a draft follow-up for replicated entry stages and logical source normalization at fan-in and streaming boundaries.
- #1276 record Higgs TTS replica benchmark results and documentation.

## TODO

1. **Replace stage-level replicas with process-level replicas.**  
   Replicate a complete process group instead of expanding each stage independently. Stages and TP ranks within the same process should share one replica identity and request binding while preserving local state and process topology.

2. **Support replicated entry stages.**  
   Allow the entry stage to belong to a replicated process group. The Coordinator should select a concrete entry replica during admission and verify that all configured entry replicas are ready before serving requests.

3. **Support fan-in from replicated stages.**  
   Model-facing fan-in APIs should continue using logical stage names even when data comes from a physical replica such as `encoder@r1`. Transport routing, ACKs, ordering, and observability should retain the physical replica identity.

4. **Validate compatibility across model pipelines.**  
   Test process-level replicas across representative TTS, multimodal, and TP-enabled pipelines, covering startup, inference, streaming, abort, output parity, and performance. Track each discovered incompatibility in a child issue and fix runtime-generic problems in the framework.

5. **Add request-hash cache-affinity routing.**  
   Add a deterministic routing policy that sends requests sharing reusable context to the same replica, using a stable session or prefix/cache key rather than only a unique request ID. Compare its cache-hit rate, latency, throughput, and load balance against round-robin routing.

6. **Add metrics and tooling for automated performance tuning.**  
   Expose machine-readable per-process and per-replica metrics such as queue depth, in-flight requests, service time, batch size, throughput, tail latency, cache-hit rate, CPU/GPU utilization, and memory pressure. Provide profiling, benchmarking, and configuration-search tools so future autotuners can optimize replica counts, placement, routing policies, and runtime parameters for the best workload-specific performance.

Peter-Nie2003:
## Motivation

The Qwen3-Omni speech tail (`talker_ar` → `code2wav`) becomes the throughput
bottleneck at higher concurrency while the thinker still has capacity. This PR
adds process-level replicas so independent worker instances can serve different
requests concurrently without model-specific routing code.

## Modifications

- Add `PipelineConfig.processes`, where each `ProcessConfig` declares
  `num_replicas` and explicit `replica_devices` for every replica and TP rank.
- Replicate a complete logical Process: all member stages are copied together.
- Bind each request to one replica per replicated Process at admission and carry
  the sticky bindings through payload, stream, completion, abort, and admin paths.
- Keep logical stage names in model code while runtime instances use the reserved
  `@rN` suffix.
- Validate resolved GPU placement, TP overlap, process-local edges, entry-stage
  resolution, memory budgets, factory `gpu_id` support, and conflicting CLI GPU
  overrides.
- Add documentation, migration guidance, unit coverage, and Qwen3-Omni examples.

`replica_devices` includes replica 0 and overrides member-stage GPU placement,
including with `num_replicas: 1`. GPU sharing involving `replica_devices` always
requires explicit `runtime.resources.total_gpu_memory_fraction` values.

Example 3-GPU deployment:

```yaml
stage_overrides:
  talker_ar:
    runtime:
      resources:
        total_gpu_memory_fraction: 0.123
  code2wav:
    runtime:
      resources:
        total_gpu_memory_fraction: 0.014

processes:
  talker_ar:
    num_replicas: 2
    replica_devices: [1, 2]
  code2wav:
    num_replicas: 2
    replica_devices: [1, 2]
```

With equal replica counts, the default round-robin policy keeps talker and
code2wav indices aligned. Custom policies are not guaranteed to preserve this
same-slot placement.

## Related Issues

Closes #1131.
Closes #1307.

## Accuracy Test

Validated at `151843fa`.

### CI

- All required GitHub checks pass.
- `PR Test / unit-test`: 5201 passed, 4 skipped, 46 deselected.
- All Qwen3-Omni CI stages pass.
- The existing 2-GPU `code2wav` replica test passes with identical PCM length
  and at most 2 LSB absolute PCM difference.

### Full process-replica correctness

Manually validated on 2× H100 80GB using the complete logical replica topology:

- GPU 0: thinker;
- GPU 1: `talker_ar@r0`, `talker_ar@r1`, `code2wav@r0`, and
  `code2wav@r1`.

Results:

- Across independent cold baseline/replica boots, 8/8 serial prompts produced
  identical text and WAV metadata, with PCM matching within 2 LSB.
- At c=8, 8/8 text outputs matched and 7/8 baseline/replica PCM outputs matched
  within 2 LSB. The single-instance baseline itself preserved serial PCM for
  only 1/8 prompts at c=8, while replica2 preserved 2/8. Concurrent PCM
  variation is therefore treated as existing talker batching behavior rather
  than a replica-specific regression.
- All 28 observed admission bindings were stable and same-slot:
  14× `talker_ar@r0 → code2wav@r0` and
  14× `talker_ar@r1 → code2wav@r1`.
- Four streaming requests were disconnected after their first audio chunk while
  four concurrent streams completed normally. All four follow-up requests
  completed successfully.
- All server processes and ports were released, and both GPUs returned to
  0 MiB after teardown.
- A Python `resource_tracker` semaphore warning appeared on both baseline and
  replica shutdown. No process, port, or GPU allocation remained after exit;
  semaphore ownership cleanup is tracked as a non-blocking follow-up.

The current-head 3-GPU benchmark separately exercised the physical
`replica_devices: [1, 2]` placement for 11,370 measured requests with zero
request failures.
## Benchmark & Profiling

Current-head H200 results at `151843fa`.

### Methodology

- Topologies:
  - **2-GPU baseline:** thinker + code2wav on GPU 0; talker on GPU 1;
  - **3-GPU isolated control:** thinker on GPU 0; talker on GPU 1; code2wav on
    GPU 2;
  - **3-GPU replica2:** thinker on GPU 0; one talker + code2wav pair on each of
    GPUs 1 and 2.
- Concurrency: 1, 8, 16, 32, 64.
- Three interleaved cold-start repeats per topology and concurrency.
- Two additional cold-start repeats at c=8.
- `n = 10 × concurrency`, plus 5 warmup requests per point.
- Fixed SeedTTS EN cohorts, `temperature=0`, natural EOS.
- 51 measured points and 11,370 requests; zero failures.

### Throughput across all three topologies

QPS, mean ± sample standard deviation:

| Config | c=1 | c=8* | c=16 | c=32 | c=64 |
| 2-GPU baseline | 2.10 ± 0.12 | 7.93 ± 0.30 | 10.88 ± 0.64 | 12.51 ± 0.25 | 13.49 ± 0.17 |
| 3-GPU isolated control | 2.04 ± 0.11 | 8.04 ± 0.32 | 11.38 ± 0.20 | 14.03 ± 0.19 | 13.47 ± 0.30 |
| 3-GPU replica2 | 1.96 ± 0.06 | 7.70 ± 1.17 | 12.97 ± 0.41 | 16.75 ± 0.31 | 19.09 ± 0.79 |
| Replica2 / 2-GPU baseline paired delta | −6.5% | −2.7% | +19.3% | +33.9% | +41.6% |
| Replica2 / 3-GPU control paired delta | −3.9% | −3.8% | +14.0% | +19.4% | +41.7% |

\* c=8 uses five cold-start repeats; all other points use three.

### Replica2 versus the 2-GPU baseline

Cells are mean paired deltas across independent cold-start pairs. Positive QPS
is better; negative latency, TTFA, RTF, and gap values are better.

| Metric | c=1 | c=8* | c=16 | c=32 | c=64 |
| Throughput (QPS) | −6.5% | −2.7% | +19.3% | +33.9% | +41.6% |
| Mean audio TTFA | +4.5% | +5.6% | −13.2% | −25.3% | −51.4% |
| P95 audio TTFA | +17.7% | +57.3% | −11.4% | −16.5% | −23.3% |
| Mean audio RTF | +7.2% | +5.3% | −17.3% | −26.7% | −30.0% |
| Mean E2E latency | +7.0% | +5.2% | −16.8% | −25.9% | −29.4% |
| P95 E2E latency | +8.3% | +32.5% | −14.3% | −20.4% | −14.4% |
| P95 inter-chunk gap | +33.4% | +4.3% | −30.8% | −12.0% | +30.4% |
| Mean text TTFT | −0.6% | −4.2% | −6.7% | +1.5% | +32.5% |

### Interpretation

- Replica2 improves throughput by 19.3%, 33.9%, and 41.6% at c=16, c=32,
  and c=64. No improvement or regression is claimed at c=1 or c=8.
- At c=64, replica2 reduces mean audio TTFA by 51.4%, mean audio RTF by 30.0%,
  and mean E2E latency by 29.4% while increasing text TTFT by 32.5% and
  inter-chunk p95 by 30.4%.
- The smaller relative gain than the historical run is primarily due to a much
  faster current baseline. Replica2 absolute throughput is maintained or
  improved at c=16 and c=32 and approximately flat at c=64.
- At c=64 the 3-GPU isolated control remains flat against the 2-GPU baseline,
  while replica2 adds 41.7% versus that control. At c=32, isolation also helps,
  but replica2 adds a further 19.4% versus the isolated control.

<details>
<summary>Full current-head metrics for all three topologies</summary>

Mean audio TTFA, ms, mean ± sample standard deviation:

| Config | c=1 | c=8* | c=16 | c=32 | c=64 |
| 2-GPU baseline | 241.0 ± 7.9 | 566.5 ± 32.2 | 825.6 ± 8.1 | 1273.8 ± 5.9 | 2866.5 ± 51.3 |
| 3-GPU isolated control | 245.0 ± 11.4 | 550.1 ± 31.7 | 788.1 ± 37.3 | 1084.5 ± 77.5 | 2868.0 ± 58.6 |
| 3-GPU replica2 | 251.8 ± 12.3 | 594.2 ± 171.4 | 717.0 ± 45.7 | 951.9 ± 44.5 | 1392.2 ± 50.9 |

P95 audio TTFA, ms, mean ± sample standard deviation:

| Config | c=1 | c=8* | c=16 | c=32 | c=64 |
| 2-GPU baseline | 267.8 ± 19.2 | 866.6 ± 130.9 | 1209.7 ± 107.8 | 1796.9 ± 289.7 | 3594.4 ± 109.7 |
| 3-GPU isolated control | 272.8 ± 33.8 | 906.1 ± 144.3 | 1281.6 ± 35.6 | 1645.9 ± 31.1 | 3589.1 ± 325.3 |
| 3-GPU replica2 | 316.6 ± 63.6 | 1296.9 ± 961.0 | 1074.4 ± 150.3 | 1475.4 ± 83.0 | 2768.7 ± 750.5 |

Mean audio RTF, mean ± sample standard deviation:

| Config | c=1 | c=8* | c=16 | c=32 | c=64 |
| 2-GPU baseline | 0.156 ± 0.009 | 0.299 ± 0.015 | 0.476 ± 0.028 | 0.748 ± 0.011 | 1.318 ± 0.023 |
| 3-GPU isolated control | 0.158 ± 0.012 | 0.292 ± 0.018 | 0.450 ± 0.008 | 0.657 ± 0.012 | 1.327 ± 0.025 |
| 3-GPU replica2 | 0.167 ± 0.011 | 0.314 ± 0.072 | 0.392 ± 0.008 | 0.548 ± 0.005 | 0.922 ± 0.042 |

Mean E2E latency, seconds, mean ± sample standard deviation:

| Config | c=1 | c=8* | c=16 | c=32 | c=64 |
| 2-GPU baseline | 0.477 ± 0.027 | 0.969 ± 0.040 | 1.440 ± 0.086 | 2.507 ± 0.047 | 4.647 ± 0.067 |
| 3-GPU isolated control | 0.491 ± 0.026 | 0.953 ± 0.035 | 1.375 ± 0.019 | 2.236 ± 0.031 | 4.628 ± 0.082 |
| 3-GPU replica2 | 0.510 ± 0.015 | 1.016 ± 0.209 | 1.196 ± 0.039 | 1.856 ± 0.039 | 3.278 ± 0.134 |

P95 E2E latency, seconds, mean ± sample standard deviation:

| Config | c=1 | c=8* | c=16 | c=32 | c=64 |
| 2-GPU baseline | 0.719 ± 0.044 | 1.379 ± 0.072 | 2.041 ± 0.106 | 3.421 ± 0.146 | 6.131 ± 0.169 |
| 3-GPU isolated control | 0.742 ± 0.049 | 1.360 ± 0.142 | 2.207 ± 0.106 | 3.294 ± 0.095 | 6.042 ± 0.182 |
| 3-GPU replica2 | 0.776 ± 0.027 | 1.806 ± 1.012 | 1.754 ± 0.213 | 2.725 ± 0.179 | 5.244 ± 0.535 |

P95 inter-chunk gap, ms, mean ± sample standard deviation:

| Config | c=1 | c=8* | c=16 | c=32 | c=64 |
| 2-GPU baseline | 92.8 ± 34.5 | 224.1 ± 31.3 | 385.7 ± 102.2 | 632.0 ± 67.9 | 638.7 ± 17.1 |
| 3-GPU isolated control | 94.5 ± 35.0 | 241.0 ± 28.8 | 430.9 ± 46.5 | 671.5 ± 18.0 | 634.2 ± 18.3 |
| 3-GPU replica2 | 119.3 ± 28.0 | 230.0 ± 45.0 | 257.7 ± 27.5 | 553.1 ± 33.8 | 831.9 ± 44.1 |

Mean text TTFT, ms, mean ± sample standard deviation:

| Config | c=1 | c=8* | c=16 | c=32 | c=64 |
| 2-GPU baseline | 108.8 ± 6.8 | 172.6 ± 28.0 | 211.8 ± 25.4 | 238.5 ± 9.6 | 252.0 ± 11.6 |
| 3-GPU isolated control | 113.4 ± 10.9 | 162.9 ± 20.2 | 187.6 ± 21.2 | 214.4 ± 14.2 | 255.2 ± 25.6 |
| 3-GPU replica2 | 108.0 ± 3.2 | 162.5 ± 18.1 | 195.4 ± 19.5 | 241.3 ± 15.8 | 332.0 ± 44.6 |

\* c=8 uses five cold-start repeats; all other points use three.

</details>

<details>
<summary>c=8 per-repeat throughput</summary>

| Cold pair | Baseline | Replica2 | Delta |
| --- | ---: | ---: | ---: |
| 1 | 7.605 | 8.183 | +7.6% |
| 2 | 8.345 | 8.465 | +1.4% |
| 3 | 8.086 | 5.622 | −30.5% |
| 4 | 7.898 | 8.068 | +2.2% |
| 5 | 7.710 | 8.143 | +5.6% |

Four of five pairs improved. Pair 3 contained isolated 3–3.8 s admission
stalls that did not reproduce in the other four runs. The outlier is retained:
the mean paired delta is −2.7%, the median is +2.2%, and no c=8 improvement or
regression is claimed.

</details>

<details>
<summary>Historical single-repeat comparison</summary>

Historical H200 results at `91ea41ab` used one cold run on a shared host and did
not include c=8:

| Config | c=1 | c=16 | c=32 | c=64 |
| 2-GPU baseline | 1.89 | 7.99 | 10.34 | 11.16 |
| 3-GPU isolated control | 2.05 | 9.16 | 10.42 | 10.66 |
| 3-GPU replica2 | 1.95 | 12.29 | 15.60 | 19.43 |

Absolute throughput change from the historical run to the current-head run:

| Config | c=1 | c=16 | c=32 | c=64 |
| 2-GPU baseline | +11.1% | +36.2% | +21.0% | +20.9% |
| 3-GPU isolated control | −0.5% | +24.2% | +34.6% | +26.4% |
| 3-GPU replica2 | +0.5% | +5.5% | +7.4% | −1.8% |

The current relative gain is smaller mainly because the baseline became much
faster. Since `91ea41ab`, upstream moved the default code2wav placement off the
talker GPU (`4ef32adf`) and added multiple thinker, talker, and code2wav
optimizations. Replica2 absolute throughput was maintained or improved at c=16
and c=32 and remained approximately flat at c=64.

The historical run already showed an inter-chunk p95 trade-off at c=64
(686 ms → 763 ms, +11.2%). The current +30.4% result is therefore not treated
as pure noise, but the two runs are not directly comparable because the code,
placement, and repeat methodology changed.

</details>

### Limitations

- The selected GPUs were checked empty before every boot and no external GPU
  activity was observed, but measurements were collected on a shared host.
- The independent sample count is the number of cold starts: three pairs per
  point and five at c=8. Requests within one boot share scheduler and GPU state.
- c=8 remains noisy; its non-reproducing admission-stall run is retained.
- Replica2 trades better high-concurrency throughput, audio TTFA, RTF, and E2E
  latency for worse c=64 text TTFT and inter-chunk p95.
- The benchmark itself validates serving stability and performance; semantic
  equivalence is covered separately by the 2-GPU full logical-topology test.

## Checklist

- [x] Pre-commit, formatting, and documentation checks.
- [x] Full current-head unit and Qwen3-Omni CI.
- [x] 2-GPU deterministic `code2wav` replica equivalence test.
- [x] Full 2-GPU logical-topology correctness with both `talker_ar` and
  `code2wav` replicated.
- [x] Concurrent routing, abort, recovery, and teardown validation.
- [x] Current-head 3-GPU benchmark.


sglang_omni/pipeline/stage/runtime.py
```diff
@@ -311,6 +333,7 @@ def _schedule_receive_task(
         self,
         msg: DataReadyMessage,
     ) -> None:
+        self._record_replica_bindings(msg.request_id, msg.replica_bindings)
```
edwingao28, line 389: Replica bindings only enter a stage via SubmitMessage or DataReadyMessage. However, same process payload, chunk, and signal dispatch call receive_local_ directly without propagating them.

A colocated replicated intermediary therefore has an empty binding table and fails when it sends to another replicated stage with `no replica binding`. Thread bindings through local dispatch and record them on every local receive path.

sglang_omni/pipeline/coordinator.py
```diff
@@ -74,6 +82,8 @@ def __init__(
         )
         self._terminal_stages_resolver = terminal_stages_resolver
         self._partial_results: dict[str, dict[str, Any]] = {}
+        self._replica_topology = replica_topology or ReplicaTopology()
```
edwingao28, line 90: This topology is never used by `_resolve_admin_stages()`. After expansion only instance names such as `talker_ar@r0` are registered, so `admin(stages=["talker_ar"])` raises `Unknown admin target` instead of performing the logical-name fan-out promised by this PR. Expand logical admin targets to all instances before validation and sending.

tests/test_model/test_qwen3_omni_process_replicas.py
```diff
@@ -0,0 +1,130 @@
...
+    log_file: Path = replica_server.log_file
+
+    for index in range(NUM_REQUESTS):
+        body = _post_audio_request(port, PROMPTS[index % len(PROMPTS)])
+        choice = body["choices"][0]
+        audio = choice["message"].get("audio") or {}
+        audio_b64 = audio.get("data")
+        assert audio_b64, f"request {index}: no audio in response: {body}"
+        audio_bytes = base64.b64decode(audio_b64)
+        assert len(audio_bytes) > 1000, (
```
edwingao28, line 125: The linked issue requires replica audio to be bit-identical, or within tolerance, to single-instance output for the same request, but this assertion only proves the WAV is nonempty. Add a baseline-versus-replica waveform comparison with a defined tolerance so routing errors that still produce audio cannot pass.

sglang_omni/pipeline/coordinator.py
```diff
@@ -539,8 +559,9 @@ async def _handle_completion(self, msg: CompleteMessage) -> None:
             self._requests.pop(request_id, None)
             return
 
+        from_stage = self._replica_topology.logical_name(msg.from_stage)
```
edwingao28: previous review noted streamed completions retain instance names; stream chunks have the same issue cuz _handle_stream() queues the original StreamMessage, so client GenerateChunk.stage_name becomes code2wav@r1 instead of logical code2wav. Normalize both from_stage and stage_name on outward stream messages while retaining the instance only in observability metadata.

sglang_omni/pipeline/replicas.py
```diff
@@ -0,0 +1,217 @@
...
+    """
+    replicas: dict[str, tuple[str, ...]] = {}
+    expanded: list[StageConfig] = []
+
+    for stage_cfg in stages_cfg:
+        if stage_cfg.num_replicas <= 1:
+            expanded.append(stage_cfg)
+            continue
+
+        per_replica_gpus = split_replica_devices(
```
edwingao28: fix lint here

sglang_omni/pipeline/replicas.py
```diff
@@ -0,0 +1,217 @@
...
+            if stage_cfg.replica_devices is not None
+            else stage_cfg.gpu,
+            stage_name=stage_cfg.name,
+            num_replicas=stage_cfg.num_replicas,
+            tp_size=stage_cfg.tp_size,
+        )
+
+        instance_names = []
+        for replica_id in range(stage_cfg.num_replicas):
+            instance = replica_instance_name(stage_cfg.name, replica_id)
```
edwingao28: This generated instance name can collide with an accepted logical stage name. ie. replicated worker alongside logical worker@r0. Expansion then produces duplicate names and endpoint/registry dictionaries overwrite one stage. Reserve the @r<digits> suffix or validate that expanded names are unique before runtime preparation.

sglang_omni/pipeline/replicas.py
```diff
@@ -0,0 +1,217 @@
...
+                gpu = gpu_ids[0]
+            else:
+                gpu = list(gpu_ids)
+            process = stage_cfg.process
+            if process is not None:
+                process = f"{process}{REPLICA_SEPARATOR}{replica_id}"
+            expanded.append(
+                stage_cfg.model_copy(
+                    update={
+                        "name": instance,
```
edwingao28: Renaming each replica here breaks the existing runtime_overrides lookup, which is keyed by logical stage name. resolve_stage_static_factory_args() indexes global_cfg.runtime_overrides by stage_cfg.name, so talker_ar@r0 and talker_ar@r1 fall back to defaults instead of inheriting the override configured for talker_ar. 

Either keep the lookup on the logical name or copy the overrides onto each replica before renaming.

tests/test_model/test_qwen3_omni_stage_replicas.py
```diff
@@ -0,0 +1,130 @@
...
+        audio = choice["message"].get("audio") or {}
+        audio_b64 = audio.get("data")
+        assert audio_b64, f"request {index}: no audio in response: {body}"
+        audio_bytes = base64.b64decode(audio_b64)
+        assert len(audio_bytes) > 1000, (
+            f"request {index}: audio payload suspiciously small "
+            f"({len(audio_bytes)} bytes)"
+        )
+
+    # Round-robin admission alternates replicas, so all requests succeeding
```
edwingao28: do we need these comments/trim them shorter?

edwingao28: I don't think this proves every replica actually served traffic. Four successful requests + the replica names in startup logs would still pass if everything went through `@r0`. Could we record which replica handled each request and assert that both talker and code2wav replicas show up?

sglang_omni/pipeline/stage/runtime.py
edwingao28, line 389: can a late message after abort leave this entry around? We record bindings before checking `_aborted`, so a late `DataReadyMessage` can recreate state that cleanup already removed, and the discard paths return without clearing it again. Maybe check abort before recording or pop it after discard?

sglang_omni/pipeline/mp_runner.py
```diff
@@ -50,6 +51,7 @@ def _build_stage_groups(
     endpoints: dict[str, str],
     placement_plan: StagePlacementPlan,
     process_plan: ProcessTopologyPlan,
+    replica_topology: ReplicaTopology | None = None,
```
edwingao28, line 54: I think colocated replicas miss the local fast path here. _resolve_same_process_targets() looks up the logical name, but these maps use expanded names like stage@r0, so the target gets dropped and falls back to SHM or CUDA IPC. Can we resolve the logical target through replica_topology here? lmk if I'm missing another normalization step.

tests/unit_test/pipeline/test_replicas.py
```diff
@@ -0,0 +1,339 @@
+# SPDX-License-Identifier: Apache-2.0
```
edwingao28, line 1: nit: add to test/README.md

edwingao28:
## Iso-GPU control benchmark

The benchmark in the description compares a 2-GPU baseline against a 3-GPU treatment. I wanted to know how much of the gain is replication and how much is just the extra card, so I added a third arm that spreads the same single-replica pipeline across 3 GPUs and re-ran all three back to back.

**Setup.** H200, `Qwen/Qwen3-Omni-30B-A3B-Instruct`, branch at `91ea41ab`. Seed-TTS eval (EN), streaming, `--generate-only`. 10/160/320/640 prompts at concurrency 1/16/32/64, 5 warmups per point, server restarted between arms. Thinker `max_running_requests` pinned to 64 on every arm, since the branch otherwise defaults to 16. All three arms ran in one session on the same three GPUs. 0 failed requests across all 12 points.

Arms:

1. `base2gpu` — current default. Thinker on GPU 0, talker_ar and code2wav both on GPU 1.
2. `iso3gpu` — the control. Thinker on GPU 0, talker_ar on GPU 1, code2wav on GPU 2. One stage per GPU, no replication.
3. `replica2` — this PR. Thinker on GPU 0, 2x talker_ar + 2x code2wav on GPUs 1 and 2.

### Throughput (req/s)

| Config | c=1 | c=16 | c=32 | c=64 |
|---|---|---|---|---|
| base2gpu | 1.89 | 7.99 | 10.34 | 11.16 |
| iso3gpu | 2.05 | 9.16 | 10.42 | 10.66 |
| replica2 | 1.95 | 12.29 | 15.60 | 19.43 |

At concurrency 64 that is 1.74x over `base2gpu` and 1.82x over `iso3gpu`.

### Audio TTFA, mean / p95 / p99 (ms)

| Config | c=1 | c=16 | c=32 | c=64 |
|---|---|---|---|---|
| base2gpu | 208/227/232 | 968/1854/1935 | 1416/1951/2340 | 3360/3977/4782 |
| iso3gpu | 208/226/231 | 821/1339/1473 | 1400/1823/1917 | 4106/5260/5505 |
| replica2 | 222/242/243 | 602/1083/1155 | 932/1343/1452 | 1364/1855/2047 |

### Audio RTF, mean / p95 / p99

| Config | c=1 | c=16 | c=32 | c=64 |
|---|---|---|---|---|
| base2gpu | 0.18/0.22/0.24 | 0.66/0.97/1.63 | 0.90/1.26/1.61 | 1.61/2.22/2.87 |
| iso3gpu | 0.16/0.19/0.19 | 0.56/0.87/1.10 | 0.89/1.24/1.69 | 1.73/2.57/3.80 |
| replica2 | 0.17/0.19/0.20 | 0.42/0.64/0.76 | 0.61/0.88/1.12 | 0.93/1.29/1.54 |

### Other metrics at concurrency 64

| Metric | base2gpu | iso3gpu | replica2 |
|---|---|---|---|
| End-to-end latency, mean / p95 (s) | 5.55 / 7.30 | 5.83 / 7.68 | 3.20 / 4.40 |
| Inter-chunk gap, mean / p95 (ms) | 529 / 686 | 415 / 700 | 445 / 763 |
| Text TTFT, mean / p95 (ms) | 117 / 430 | 132 / 587 | 190 / 592 |
| Audio seconds produced per second | 40.4 | 38.5 | 69.9 |

### What this shows

1. **The extra GPU on its own buys nothing.** `iso3gpu` is flat to slightly worse than the 2-GPU baseline at concurrency 64, on throughput (10.66 vs 11.16) and on TTFA (4106 vs 3360 ms). The speedup in this PR comes from replication, not from the added card.
2. **Splitting the pair across GPUs actually hurts.** Putting talker_ar and code2wav on separate GPUs adds a cross-GPU transfer to every chunk and gains no parallelism to pay for it. The sticky round-robin binding avoids that: at equal replica counts each request's talker_ar and code2wav land on the same GPU. The index-alignment note in the description is load-bearing, and this is evidence for it.
3. **RTF only gets back under realtime with replicas.** Both baselines sit above 1.0 at concurrency 64 (1.61 and 1.73); replicas reach 0.93. p95 is still 1.29, so the slowest requests stay above realtime even here.
4. **The text TTFT increase is a load effect, not a regression.** At matched throughput replicas are ahead: at 12.29 req/s (replicas, concurrency 16) TTFT is 101 ms mean and 198 ms p95, versus 117 ms and 430 ms for the baseline at 11.16 req/s. The 190 ms at concurrency 64 is the price of running at 19.43 req/s, which the baseline never reaches. Might be worth wording it that way in the description, since right now it reads like a cost.

### Caveats

1. Single repeat. Worth a second one before these numbers get quoted anywhere.
2. Inter-chunk p95 is slightly worse with replicas (763 vs 686 ms).
3. This ran on a shared host. An earlier attempt was discarded after other jobs took two of the three cards seconds after launch; the tell was `iso3gpu` scoring below the 2-GPU baseline. The harness now checks the target GPUs are free when each arm starts and samples occupancy every 20s. The numbers above are from a run where all three arms started on idle cards.

sglang_omni/pipeline/coordinator.py
edwingao28: Fixed in eae1f42c. `_handle_stream()` normalizes both `from_stage` and `stage_name` before the chunk is queued. The emit above it keeps the instance name.

sglang_omni/pipeline/replicas.py
```diff
@@ -0,0 +1,217 @@
...
+        )
+
+
+def split_replica_devices(
+    replica_devices: str | list[int] | None,
+    *,
+    stage_name: str,
+    num_replicas: int,
+    tp_size: int,
+) -> list[list[int] | None]:
```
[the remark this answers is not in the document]
edwingao28: Fixed in 70b6fc47.

tests/test_model/test_qwen3_omni_stage_replicas.py
```diff
@@ -0,0 +1,130 @@
...
+
+from __future__ import annotations
+
+import base64
+import sys
+from pathlib import Path
+
+import pytest
+import requests
+
```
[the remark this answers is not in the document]
edwingao28: Fixed in 70b6fc47.

sglang_omni/pipeline/coordinator.py
```diff
@@ -539,8 +559,9 @@ async def _handle_completion(self, msg: CompleteMessage) -> None:
...
         expected_terminal_stages = self._expected_terminal_stages(request_id)
```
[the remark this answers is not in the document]
edwingao28: Fixed in 70b6fc47.

sglang_omni/pipeline/stage/runtime.py
edwingao28: Fixed in 70b6fc47.

sglang_omni/pipeline/coordinator.py
edwingao28: Fixed in 70b6fc47.

sglang_omni/pipeline/coordinator.py
edwingao28: Fixed in 70b6fc47.

sglang_omni/pipeline/replicas.py
edwingao28: Fixed in 70b6fc47.

sglang_omni/pipeline/replicas.py
edwingao28: Fixed in 70b6fc47.

tests/test_model/test_qwen3_omni_stage_replicas.py
edwingao28: Trimmed in 70b6fc47.

sglang_omni/pipeline/stage/runtime.py
edwingao28: Yes it can; fixed in 70b6fc47.

sglang_omni/pipeline/mp_runner.py
edwingao28: Fixed in 70b6fc47.

tests/unit_test/pipeline/test_replicas.py
edwingao28: Added in 70b6fc47.

sglang_omni/pipeline/mp_runner.py
```diff
@@ -114,8 +119,11 @@ def _build_stage_groups(
             wait_for_fn=stage_cfg.wait_for_fn,
             merge_fn=stage_cfg.merge_fn,
```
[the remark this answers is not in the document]
edwingao28: Real but not fixed here; no config replicates a fan-in upstream today, leaving open so the constraint stays tracked.

sglang_omni/proto/messages.py
```diff
@@ -55,6 +56,8 @@ def to_dict(self) -> dict[str, Any]:
         if self.error is not None:
             _require_str(self.error, "error")
             d["error"] = self.error
+        if self.replica_bindings:
+            d["replica_bindings"] = dict(self.replica_bindings)
```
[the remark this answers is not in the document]
edwingao28: Not fixed; `replica_bindings` comes off the internal control plane, so leaving open as a hardening item.

tests/test_model/test_qwen3_omni_process_replicas.py
edwingao28: Not fixed; needs a baseline single-instance run to diff against, leaving open.

tests/test_model/test_qwen3_omni_stage_replicas.py
edwingao28: Fixed in 70b6fc47 — admission now logs the bindings and the test asserts both replicas were used.

sglang_omni/pipeline/replicas.py
```diff
@@ -0,0 +1,235 @@
...
+        ids = [int(part) for part in replica_devices.split(",") if part.strip()]
+    elif isinstance(replica_devices, int):
+        ids = [replica_devices]
+    else:
+        ids = [int(part) for part in replica_devices]
+
+    if len(ids) == num_replicas * tp_size:
+        return [ids[r * tp_size : (r + 1) * tp_size] for r in range(num_replicas)]
+    if len(ids) == tp_size:
+        return [[gpu_id + r * tp_size for gpu_id in ids] for r in range(num_replicas)]
```
stephenkgli: This may produce unexisted gpu_id? e.g.:

Available GPU：0,1,2,3

N=2, T=1, devices="3"
→ [[3], [4]]

The gpu_id 4 is unavailable.

edwingao28: solved in https://github.com/sgl-project/sglang-omni/pull/1175/commits/4e6812ab42e602cd622a8048c1dbadcca80be2f7

edwingao28 (approved):

stephenkgli:
Validated using two H200 GPUs.

| Test Category | Validated Scenarios | Result |
|---|---|---|
| Unit and Configuration | Validated replica topology, coordinator/runtime behavior, device placement, and rejection of invalid or conflicting GPU configurations across multiple model families. | PASS — 165 relevant tests |
| Replica Runtime and Data Transport | Tested a four-stage CPU pipeline with one entry process, two replicas of the first intermediate stage, three replicas of the second intermediate stage, and one terminal process. Also validated same-GPU and cross-GPU CUDA data transfer, streaming, completion, abort, recovery, shutdown, and resource cleanup. | PASS |
| Qwen3-TTS | Tested single-replica and two-replica deployments with serial, concurrent, streaming, abort/recovery, request-sticky routing, balanced replica usage, valid audio output, and a 12-sample quality-equivalence check. | PASS |
| Qwen3-Omni | Tested replicated audio decoding, two-GPU tensor parallelism, text/speech/multimodal requests, conditional routing, streaming, and valid text/audio output. | PASS with caveats — custom all-reduce fell back to NCCL, and conditional traffic showed replica load skew |
| Qwen3-ASR and Device Remapping | Tested balanced routing across two ASR replicas and verified that a single replica could be explicitly moved to another logical GPU with correct physical GPU isolation. | PASS |
| Cleanup and Stability | Verified abort recovery, clean worker termination, no residual requests, and GPU memory release after model shutdown. | PASS |

sglang_omni/pipeline/replicas.py
```diff
@@ -0,0 +1,278 @@
...
+                )
+
+
+class BindingPolicy(Protocol):
+    """Selects a replica for one (logical Process, request) at admission."""
+
+    def bind(self, process_name: str, num_replicas: int, request_id: str) -> int: ...
+
+
+class RoundRobinBindingPolicy:
```
JiaxinD, line 238: Note: the talker_ar/code2wav same-GPU pairing relies on these counters advancing in lockstep, which nothing tests or validates. Unequal replica counts pass validation but route most codec streams cross-GPU via relay; consider enforcing equal counts for stream-connected replicated processes.

sglang_omni/config/topology.py
```diff
@@ -236,39 +400,38 @@ def _validate_gpu_process_colocation(
...
+            record(gpu_id, topology_plan.tp_stage_to_processes[stage.name][rank], stage)
 
     require = config.placement.require_memory_fraction_for_colocation
     limit = config.placement.max_total_gpu_memory_fraction_per_gpu
     for gpu_id, process_names in gpu_processes.items():
         if len(process_names) <= 1:
             continue
-        missing = sorted(missing_fraction.get(gpu_id, set()))
+        missing = missing_fraction.get(gpu_id, set())
         if require and missing:
```
JiaxinD, line 430: With require_memory_fraction_for_colocation=False (qwen3 speech, ming_omni), stacking two default-fraction replicas on one card passes all validation and only fails at startup OOM. Should replica-induced GPU sharing require declared fractions regardless of the model opt-out?

sglang_omni/models/qwen3_omni/placement.py
```diff
@@ -46,24 +46,36 @@ def validate(self, config: PipelineConfig, plan: StagePlacementPlan) -> None:
...
-        talker = plan.stages.get("talker_ar")
-        if thinker is None or talker is None:
-            return
-        if thinker.tp_size != 1 or talker.tp_size != 1:
-            return
-        if not set(thinker.gpu_ids).intersection(talker.gpu_ids):
-            return
+        for thinker in plan.instances_of("thinker"):
+            for talker in plan.instances_of("talker_ar"):
+                if thinker.tp_size != 1 or talker.tp_size != 1:
```
JiaxinD, line 57: This skips the check whenever either side uses TP, so thinker tp=2 on [0,1] with replica_devices [1,2] lands talker_ar@r0 beside thinker rank 1 unvalidated. The skip predates this PR, but replicas make it one config away.

JiaxinD (commented):
Reviewed the replica mechanism end to end (send paths, fan-in normalization, binding atomicity, shutdown): the core routing looks solid and the default path checks out as equivalent. A few points on the PR itself:

* The body still describes the stage_overrides form, but the implementation is process-level `processes:`; the benchmark also ran at 91ea41ab before that switch. Worth updating and confirming the numbers still hold.
* The retirements riding along (`--isolate-stage`/`--stage-process`, `fused_stages`) are not mentioned in the body; existing configs break on upgrade, a migration note would help.
* Is entry-stage replication an intended public capability? It works and is tested but is undocumented.

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -159,6 +164,39 @@ def __init__(
...
+        return self._replica_topology.logical_name(from_stage)
+
+    def _resolve_target_instance(self, request_id: str, target: str) -> str:
+        if not self._replica_topology.is_replicated(target):
+            return target
+        bindings = self._replica_bindings.get(request_id)
+        replica_id = None if bindings is None else bindings.get(target)
+        if replica_id is None:
+            raise RuntimeError(
+                f"Stage {self.name}: no replica binding for target {target!r} "
```
JiaxinD, line 195: A missing binding here escalates to a stage-process crash via the outbox drain task, failing all in-flight requests. If that is an intentional invariant assertion a comment would help; otherwise consider a per-request failure instead.

sglang_omni/pipeline/coordinator.py
```diff
@@ -380,12 +400,20 @@ async def _submit_request(
...
+        replica_bindings = assign_replica_bindings(
+            self._logical_process_plan, self._binding_policy, request_id
+        )
+        bindings = replica_bindings or {}
+        entry_instance = (
+            self._replica_topology.resolve(self.entry_stage, bindings[self.entry_stage])
+            if self._replica_topology.is_replicated(self.entry_stage)
+            else self.entry_stage
+        )
+        entry_info = self._stages[entry_instance]
```
JiaxinD, line 415: An unregistered entry now surfaces as a bare KeyError; main had an explicit ValueError for this.

tests/test_model/test_qwen3_omni_process_replicas.py
```diff
@@ -0,0 +1,152 @@
...
+from tests.utils import (
+    disable_proxy,
+    server_log_file,
+    start_server_from_cmd,
+    stop_server,
+)
+
+REQUIRED_GPUS = 3
+
+pytestmark = pytest.mark.skipif(
```
JiaxinD, line 38: This needs 3 GPUs but the CI lanes run 2, so the only GPU test for replicas never runs in CI. Is a 2-GPU variant (thinker plus one replicated process) feasible so the mechanism gets real coverage?

sglang_omni/pipeline/replicas.py
stephenkgli: With the default RoundRobinBindingPolicy and equal replica counts, these counters cannot drift in production: every admission calls bind exactly once for every replicated Process before any conditional routing, and aborts or skipped stages do not affect the counters. Unequal counts or a custom policy may intentionally produce cross-GPU routing, but that is a configured topology choice rather than a missing runtime binding.

sglang_omni/pipeline/stage/runtime.py
stephenkgli: A missing binding is not reachable from a normal production request. The coordinator creates a binding for every replicated Process at admission, and SubmitMessage plus every downstream payload and stream path propagate the same binding map before routing. Reaching this branch therefore means an internally corrupted or hand-crafted envelope, so failing the stage process is intentional rather than treating it as a recoverable per-request error.

sglang_omni/pipeline/coordinator.py
stephenkgli: This state is not reachable through the production launcher. MultiProcessPipelineRunner waits for every stage process to become ready, checks that none died, registers every expanded stage, and only then creates the HTTP server; the entry instance and registrations also come from the same prepared topology. A missing entry can therefore occur only if the internal Coordinator is called before runner startup or its registry is manually mutated.

sglang_omni/config/topology.py
stephenkgli: This is intentional. `require_memory_fraction_for_colocation=False` is the model-level opt-out from requiring explicit memory fractions for colocated process groups, including sharing introduced by replicas. Declared fractions are still summed and checked against the placement limit; only missing-fraction enforcement is disabled. Requiring fractions specifically for replicas would change the existing opt-out contract rather than fix an unintended validation gap.

tests/test_model/test_qwen3_omni_process_replicas.py
stephenkgli: cc @edwingao28

stephenkgli:
> Reviewed the replica mechanism end to end (send paths, fan-in normalization, binding atomicity, shutdown): the core routing looks solid and the default path checks out as equivalent. A few points on the PR itself:
> 
> * The body still describes the stage_overrides form, but the implementation is process-level `processes:`; the benchmark also ran at 91ea41ab before that switch. Worth updating and confirming the numbers still hold.
> * The retirements riding along (`--isolate-stage`/`--stage-process`, `fused_stages`) are not mentioned in the body; existing configs break on upgrade, a migration note would help.
> * Is entry-stage replication an intended public capability? It works and is tested but is undocumented.

Missing docs will be added, thanks for reminding me of that. Entry stage is intended to be replicable, this PR should make every qualified stage (currently, the stage has gpu_id as arg) replicable.

tests/test_model/test_qwen3_omni_process_replicas.py
edwingao28: thx for pointing that out. let's change it to 2, should reach same coverage purpose

sglang_omni/pipeline/replicas.py
JiaxinD: Agreed the counters hold for the default policy. Since unequal counts are legal, a doc line on the cross-GPU consequence plus a lockstep regression test would pin the contract down.

sglang_omni/config/topology.py
JiaxinD: The opt-out predates replicas: it was scoped to hand-tuned colocated configs the model config itself curates, not user-composed sharing a replica parameter creates. Since no existing config uses replicas, requiring declared budgets for replica-induced sharing breaks nothing; suggest folding this into the ongoing gpu memory scheme discussion.

sglang_omni/models/qwen3_omni/placement.py
stephenkgli: fixed in d8a64d04

sglang_omni/pipeline/replicas.py
stephenkgli: fixed in f08ee777

tests/test_model/test_qwen3_omni_process_replicas.py
edwingao28: fixed in 81091c12

sglang_omni/config/topology.py
edwingao28: fixed in e3a54bd7

sglang_omni/models/qwen3_omni/placement.py
edwingao28: fixed in 645b17dd

tests/test_model/test_qwen3_omni_process_replicas.py
edwingao28: fixed in cd0ed203

PR #1176, opened 2026-07-24
fix(qwen3-asr): avoid silently truncating audio beyond 30 seconds

### LINKED ISSUE #1173: [Bug] Qwen3-ASR silently truncates audio longer than 30 seconds
### Checklist

- [x] 3. I have included the environment information and a minimal reproducible example.
- [x] 4. This is a bug report rather than a usage question.

### Describe the bug

Qwen3-ASR on current `main` silently discards audio after the first 30 seconds. The request still returns HTTP 200 and reports the original full duration in `usage`, so callers cannot tell that the transcription is incomplete.

The request builder calls `WhisperFeatureExtractor` with `padding="longest"` and `truncation=True`. On the 62.455-second reproduction file, this produces 3,000 valid mel frames. Setting only `truncation=False` produces 6,245 valid mel frames.

The official `qwen-asr==0.0.6` Transformers backend uses the same model weights and the same `WhisperFeatureExtractor`, but explicitly sets `truncation=False`. It transcribed the complete 62.455-second file in three out of three runs. This indicates that 30 seconds is a limit of the current SGLang-Omni preprocessing path rather than a Qwen3-ASR model capability limit.

Potentially related paths, identified by source inspection but not yet validated end to end:

- Whisper ASR, including Whisper Large v3, calls its feature extractor without overriding the default truncation behavior. It may therefore exhibit the same silent truncation for uploads longer than 30 seconds. This has not yet been reproduced with a running Whisper server.
- TTS and Omni quality evaluations that use Qwen3-ASR as the WER transcriber may indirectly ignore speech after 30 seconds when a generated sample exceeds that duration. This may affect long-form or max-token outliers from Qwen3-TTS, Higgs Audio, MOSS-TTS, Voxtral TTS, FishAudio, and Qwen3-Omni talker evaluations. The potential downstream impact has not yet been quantified.
- Fun-ASR Nano has a related 30-second implementation boundary, but it explicitly rejects longer audio instead of silently returning a partial transcript. It is therefore not confirmed to have this bug.
- MOSS-Transcribe-Diarize uses chunk-aware processor outputs, Qwen3-Omni's Hugging Face processor defaults to `truncation=False`, and Ming-Omni computes mel features from the complete waveform. Source inspection does not currently indicate the same truncation behavior in those paths.

### Reproduction

Start Qwen3-ASR from source:

```bash
sgl-omni serve \
  --config examples/configs/qwen3_asr_rtx4090.yaml \
  --model-path /path/to/Qwen3-ASR-1.7B \
  --host 127.0.0.1 \
  --port 8000 \
  --log-level debug
```

Download the single validation sample from `distil-whisper/librispeech_long` and submit its unmodified 62.455-second WAV:

```bash
curl -F file=@librispeech_long.wav \
  -F model=Qwen/Qwen3-ASR-1.7B \
  -F language=en \
  http://127.0.0.1:8000/v1/audio/transcriptions
```

Input provenance:

```text
Dataset: distil-whisper/librispeech_long
Dataset file: clean/validation-00000-of-00001-913508124a40cb97.parquet
Audio duration: 62.455 seconds
Audio sample rate: 16,000 Hz
Audio channels: 1
Audio SHA256: 8f418821d760804084ce9e054474098768b90c3e873489891f61162ba25ddfe3
```

Observed on SGLang-Omni `main`, repeated three times:

```text
HTTP status: 200
usage.seconds: 63
input feature shape: [1, 128, 3000]
feature attention-mask sum: 3000
transcript ending: "...and can discover."
```

All three responses were identical and stopped at the same incomplete sentence near the 30-second boundary.

Feature-extractor comparison on the same waveform:

```text
padding="longest", truncation=True:
  shape=[1, 128, 3000], attention_mask.sum()=3000

padding="longest", truncation=False:
  shape=[1, 128, 6245], attention_mask.sum()=6245
```

The official `qwen-asr==0.0.6` Transformers backend returned the complete transcript in three out of three runs. Its transcript continued after the SGLang stopping point and ended with:

```text
...before he says, like a shampooer in a Turkish bath, "Next man."
```

### Environment

```text
Operating system: Ubuntu 24.04.3 LTS
Kernel: 6.8.0-65-generic
Python: 3.12.3
SGLang-Omni commit: cf61f234e5a2cc7f63a24bdef84b446bd9a42f74
Qwen3-ASR model revision: 7278e1e70fe206f11671096ffdd38061171dd6e5
PyTorch: 2.11.0+cu130
PyTorch CUDA build: 13.0
Transformers: 5.6.0
NVIDIA driver: 580.95.05
GPU: NVIDIA GeForce RTX 5090
GPU memory: 32,607 MiB
CPU: AMD Ryzen 9 9950X 16-Core Processor
```

The official-engine comparison used:

```text
qwen-asr: 0.0.6
Transformers: 4.57.6
PyTorch: 2.11.0+cu130
Model revision: 7278e1e70fe206f11671096ffdd38061171dd6e5
```

Related pull request: [sgl-project/sglang-omni#1154 — feat(qwen3-asr): add RTX 4090 serving profile](https://github.com/sgl-project/sglang-omni/pull/1154) converts the silent truncation into an HTTP 400 response at 30 seconds, but does not enable the model's existing variable-length path.

lijrjyan:
## Motivation

Qwen3-ASR's encoder natively accepts variable-length mel sequences beyond Whisper's 30-second window — the 30-second limit is not a model capability boundary, it is an artifact of passing `truncation=True` to `WhisperFeatureExtractor`. With that flag, audio after 30 seconds is silently discarded: the request still returns HTTP 200 and reports the full original duration in `usage`, so callers receive a partial transcript with no signal that anything was dropped.

The correct behavior is to avoid silent truncation: feed the full mel sequence to the encoder when it fits the configured context, and let genuine capacity limits surface as explicit errors instead of silently wrong transcripts.

## Modifications

- Disable feature-extractor truncation while retaining variable-length padding.
- Add a 31-second request-builder contract test that verifies all 3,100 mel frames and attention-mask entries reach the request payload.

## Related Issues

Closes #1173.

PR #1154 proposes rejecting inputs longer than 30 seconds. This change instead uses the model's native variable-length input path; if the changes are consolidated, #1154's consumer profile can remain without its hard 30-second guard.

## Accuracy Test

Tested from commit `f1ae0ca4030c1542222ebb460aa098a34b8945cc` on one RTX 5090 with the unmodified 62.455-second `distil-whisper/librispeech_long` validation sample:

- mel features: `[1, 128, 6245]`, attention-mask sum: 6,245;
- long audio with `max_new_tokens=512`: 3/3 HTTP 200 responses produced the same complete 150-word transcript;
- existing short English fixture: 3/3 HTTP 200 responses matched the expected transcript.

The 512-token override isolates input preservation from the pipeline's separate 128-token default output budget.

## Benchmark & Profiling

N/A: this is a correctness fix and makes no throughput or latency claim. Longer inputs intentionally require more preprocessing and encoder memory than the previously truncated input.

With truncation disabled, long inputs are now bounded by the request-token/KV budget instead of being silently cut: at the current 1,636-token context and 128-token default output budget the limit is reached at approximately 114.96 seconds, and a 124.91-second negative control was rejected before scheduling (currently surfaced as HTTP 500). Two follow-ups are intentionally out of scope here: turning that capacity rejection into an actionable client error, and chunk-and-merge support (official-wrapper long-audio parity) so inputs beyond the context budget can be processed rather than rejected.

## Checklist

- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed. (N/A: no public option or documented limit changes.)
- [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed. (N/A for performance; the applicable accuracy test is reported above.)

## CI

Pre-submission checks:

```text
# original head f1ae0ca4, frozen test environment
python -m pytest -q tests/unit_test/qwen3_asr
10 passed, 20 warnings

# final head b7adaacd1 (comment and commit-title amend only), replayed on the frozen environment
python -m pytest -q tests/unit_test/qwen3_asr
11 passed, 20 warnings

# final head formatting: black / isort / ruff clean on both changed files;
git diff --check
Passed
```


wirybeaver (approved):
The change correctly removes Whisper's artificial 30-second preprocessing truncation while retaining true-length padding. The regression test verifies that a 31-second input preserves all 3,100 mel frames and attention-mask entries.

The genuine context/token-capacity validation and error mapping are handled in follow-up [#1193](https://github.com/sgl-project/sglang-omni/pull/1193). Any upload/preprocessing resource limits can remain a separate non-blocking follow-up; neither should block this correctness fix.

0xjeffro (approved):

PR #1181, opened 2026-07-25
[Fix][TTS][MOSS] Support all MOSS-TTS Local v1.5 languages

CloudRipple:
## Motivation

MOSS-TTS-Local-Transformer-v1.5 supports 31 languages, but the shared Speech API language set only contains 10 of them. As a result, requests using the other 21 languages documented in the model card are rejected during Speech API validation before reaching the MOSS-TTS-Local processor.

The shared API also accepts `Auto`, while the Hugging Face processor does not special-case that value. It treats the supplied language as literal prompt text, so passing `Auto` through would render `Auto` in the `- Language:` field instead of using the processor's unspecified-language behavior.

## Modifications

- Add the 21 missing model-card languages to `MossTTSLocalPipelineConfig.additional_speech_languages`, completing the model's 31-language set without changing the shared Speech API language list.
- Normalize case-insensitive `Auto` to `None` in the MOSS-TTS-Local request builder before calling `processor.build_user_message(...)`.
- Add regression coverage verifying that `Auto` is converted to `None` while explicit language names continue to pass through unchanged.

The newly accepted languages are Arabic, Cantonese, Czech, Danish, Dutch, Finnish, Greek, Hebrew, Hindi, Hungarian, Macedonian, Malay, Persian (Farsi), Polish, Romanian, Swahili, Swedish, Tagalog, Thai, Turkish, and Vietnamese.

## Related Issues

N/A.

## Accuracy Test

N/A.

## Benchmark & Profiling

N/A. The change is limited to request validation metadata and one case-insensitive string check during request construction; it does not affect the model execution or serving hot path.

## Checklist

- [x] Add unit tests.

## CI

if labeled.

### DIFF
--- a/sglang_omni/models/moss_tts_local/config.py
+++ b/sglang_omni/models/moss_tts_local/config.py
@@ -81,6 +81,31 @@ class MossTTSLocalPipelineConfig(PipelineConfig):
         "MossTTSLocal",
         "MossTTSLocalForConditionalGeneration",
     )
+    additional_speech_languages: ClassVar[frozenset[str]] = frozenset(
+        {
+            "Cantonese",
+            "Arabic",
+            "Czech",
+            "Danish",
+            "Dutch",
+            "Finnish",
+            "Greek",
+            "Hebrew",
+            "Hindi",
+            "Hungarian",
+            "Macedonian",
+            "Malay",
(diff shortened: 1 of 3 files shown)


Ratish1 (approved):
Looks good

Issue #1148, opened 2026-07-23
[Qwen3-Omni Perf] Code2Wav: wire configurable initial codec chunk size into the Qwen3-Omni vocoder scheduler


edwingao28:
## Why

The Qwen3-Omni vocoder scheduler uses one hard-coded chunk size for every chunk, including the first:

- `sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py:60` — `stream_chunk_size: int = 10`, applied uniformly (`:152` accumulate gate feeding `_decode_and_emit`).
- `sglang_omni/models/qwen3_omni/config.py:189` — `factory_args={"device": "cuda"}` only; the factory (`code2wav_scheduler.py:271-289`) already accepts `stream_chunk_size` / `left_context_size` kwargs, but config never passes them.

First-audio latency is bounded below by "wait for a full steady chunk". The shared primitive to fix this already exists — `resolve_initial_codec_chunk_frames` (`sglang_omni/models/tts_streaming.py:11-37`) — and is wired for higgs_tts and moss_tts_local, but `Code2WavScheduler` (a `StreamingSimpleScheduler`) never uses it.

A small first chunk (e.g. 4 frames) with a larger steady chunk (e.g. 25) and 25-frame left context is a sensible starting point, with one seam rule to verify: chunk 1 must use the whole initial chunk as its left context so the overlap-trim stays seamless across the size change. The current clamp `context = min(left_context_size, start)` (`:240`) already gives chunk 1 the full available history, so this is a property to verify with a test, not new logic.

## Scope

1. Make initial and steady chunk sizes configurable for the Qwen3-Omni code2wav stage (config + per-request override via the existing resolve helper).
2. Verify (unit test) the chunk-1 left-context handoff stays seamless across the size change.
3. Fixed-value sweep (initial ∈ {2, 4, 10}, steady ∈ {10, 25}) to pick defaults.

Out of scope: load-adaptive sizing (gated on this sweep, see roadmap), batching (#1026), CUDA graph (#1101). Note steady=25 interacts with #1101's capture shapes — coordinate defaults.

## Approach

1. Thread initial/steady chunk sizes through `config.py` factory_args into `Code2WavScheduler` (the factory kwargs already exist; add the initial size via the shared resolve helper).
2. In the accumulate gate, flush chunk 0 at the initial size; afterwards count from the initial offset.
[...]

PR #1182, opened 2026-07-25
optimizate  code2wav  default initial_codec_chunk_frames & stream_chunk_size & left_context_size

LukeWang166:
## Motivation

The Qwen3-Omni code2wav scheduler previously used a single hard-coded chunk size (`stream_chunk_size=10`) for every decode window, including the first, and never exposed this parameter through config. This means Time-To-First-Audio (TTFA) is bounded below by "wait for a full steady chunk." The balance between steady chunk size, left context size, and first-chunk size governs the trade-off between TTFA vs. steady-state throughput (xRT) vs. audio quality — there is no single "correct" value that fits all deployments.

- `initial_codec_chunk_frames` controls TTFA: a smaller first chunk means earlier first audio, but too small degrades audio quality due to the vocoder's receptive field requirements.
- `stream_chunk_size` controls steady xRT: larger chunks amortize vocoder calls but lengthen the gap between streamed audio segments.
- `left_context_size` controls seam quality: larger overlap = cleaner splice but more compute per window.

This PR makes all three configurable at the config level and overridable per request, so deployments can tune optimal values via TTFA/xRT sweeps (at c1/c8) for their hardware and latency budget without code changes.

## Modifications

**`sglang_omni/models/qwen3_omni/config.py`**
- `_code2wav_stage`: Added `stream_chunk_size=25`, `left_context_size=25`, `initial_codec_chunk_frames=0` to `factory_args`. The default `0` preserves legacy behavior (every chunk uses steady size); flipping to e.g. `4` enables a smaller first chunk for all requests.

**`sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py`**
- `Code2WavScheduler.__init__`: Added `initial_codec_chunk_frames` parameter and per-request `_req_initial_chunk_frames` dict.
- `on_streaming_new_request`: Calls new `_latch_initial_chunk_frames` to resolve the per-request first-chunk size from `payload.request.params`, routed through the shared `resolve_initial_codec_chunk_frames` helper (the same primitive higgs_tts and moss_tts_local use), so clamp/validation semantics stay identical.
- `on_stream_chunk` accumulate gate: First emit (`emitted == 0`) uses the initial chunk size as threshold; subsequent emits use steady size, counted from the initial offset. The existing cursor `emitted` already computes `ready = len(chunks) - emitted` from the right position, so no new offset logic is needed.
- `clear_stream_state`: Cleans up the new `_req_initial_chunk_frames` entry.
- Added `_latch_initial_chunk_frames` / `_initial_chunk_frames_for` helpers. An explicit `0` is an opt-out ("use steady for every chunk"), distinguished from "unset" via `is not None` so a request's intent is never silently overridden.
- `create_code2wav_scheduler` factory: Added `initial_codec_chunk_frames` pass-through.
- `_decode_incremental` unchanged: the existing `context = min(left_context_size, start)` already lets chunk 1 use the full initial chunk as left context, keeping the seam seamless across the size change via overlap-trim.

## Related Issues

https://github.com/sgl-project/sglang-omni/issues/1148

## Accuracy Test

full unit suites 0 failures.

## Benchmark & Profiling

Expected to impact performance. Plan: sweep `initial_codec_chunk_frames ∈ {0, 2, 4, 6}` × `stream_chunk_size ∈ {10, 25}`, recording TTFA + steady xRT at concurrency 1 and 8. Target: lower TTFA than steady-only config with acceptable steady xRT degradation (≥1.0 xRT) at c8. Results to be appended once GPU access is available.

## Checklist

- [x] Add unit tests.
- [x] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as co-author when merging the PR.

### DIFF
--- a/sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
+++ b/sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
@@ -120,16 +120,6 @@ class Code2WavScheduler(StreamingSimpleScheduler):
     def _latch_initial_chunk_frames(
         self, request_id: str, payload: StagePayload | None
     ) -> None:
-        """Resolve and cache the per-request first-chunk size.
-
-        A request may override the scheduler default via
-        ``params['initial_codec_chunk_frames']`` (0 opts out of a smaller
-        first chunk; values are clamped to ``stream_chunk_size``). Absent the
-        param, the scheduler-level ``initial_codec_chunk_frames`` default
-        applies. Called from ``on_streaming_new_request``; safe before or
-        after chunks arrive because the gate only consumes the value while
-        nothing has been emitted yet.
-        """
         if request_id in self._req_initial_chunk_frames:
             return
         params = None
@@ -148,10 +138,6 @@ class Code2WavScheduler(StreamingSimpleScheduler):
         self._req_initial_chunk_frames[request_id] = resolved
 
     def _initial_chunk_frames_for(self, request_id: str) -> int:
-        """Per-request first-chunk size, falling back to the scheduler default
-        when the payload has not arrived yet (chunks may precede the payload
-        because ``can_accept_stream_before_payload`` is set). 0 means 'use the
-        steady chunk size for every chunk'."""
         resolved = self._req_initial_chunk_frames.get(request_id)
         if resolved is not None:
             return resolved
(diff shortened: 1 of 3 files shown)


edwingao28:
@LukeWang166 please rebase ur pr, thanks!

edwingao28:
Could you also provide benchmark result after rebase so we could see the performance gain?

Issue #1048, opened 2026-07-13
Bump SGLang to v0.5.15 - cuda graph part


stephenkgli:
### Motivation

Upgrade SGLang from `0.5.12.post1` to `0.5.15` and prepare sglang-omni for future prefill CUDA Graph support.

SGLang `0.5.15` introduces a phase-aware CUDA Graph runtime with separate prefill and decode execution paths. Omni needs to align with this runtime before individual models can safely adopt prefill CUDA Graph.

### Problems to solve

- Omni still relies on the legacy CUDA Graph initialization lifecycle, which is incompatible with SGLang `0.5.15`.
- The upgrade must preserve existing decode CUDA Graph behavior and Omni-owned model graphs.
- Prefill CUDA Graph must not be enabled accidentally for existing models that have not been validated for it.
- Future model integrations should not need to redesign the Omni scheduler, phase lifecycle, or graph ownership.

### Proposed approach
The upgrade keeps the existing Omni scheduling flow and changes the boundary at which the SGLang runtime is initialized.

    [Bootstrap]

    Model-specific scheduler/engine builder
                       |
                       v
               Omni ModelWorker
                       |
                       v
    SGLModelRunner (SGLang ModelRunner)
                       |
                       +---- model and stage-specific setup
                       |
                       +---- init_cuda_graphs()
                                |
                                +---- EagerRunner
                                |       always available
                                |
                                +---- PrefillCudaGraphRunner
                                |       disabled for existing Omni models
                                |
                                +---- DecodeCudaGraphRunner
                                        preserves existing decode graphs

    Independently owned Omni graphs
    (encoder / frame decoder / vocoder)
                       |
                       +---- initialized separately from init_cuda_graphs()

    [Runtime]

    OmniScheduler
         |
         v
    Omni ModelRunner
         |
         v
    ModelWorker.forward_batch_generation(ForwardBatch)
         |
         v
    SGLModelRunner.forward(ForwardBatch)
         |
         +---- ForwardMode.EXTEND
         |         |
         |         +---- PrefillCudaGraphRunner, when explicitly enabled
         |         +---- EagerRunner, otherwise
         |
         +---- ForwardMode.DECODE
                   |
                   +---- DecodeCudaGraphRunner, when eligible
                   +---- EagerRunner, otherwise

The generic path delegates graph eligibility, stable-buffer management, capture,replay, and eager fallback to SGLang. It therefore does not depend on a specific prefill CUDA Graph backend.

[...]

PR #1183, opened 2026-07-25
Bump SGLang to 0.5.16

junliu-mde:
## Summary

Bump the SGLang backbone from `0.5.12.post1` to `0.5.16` and adapt sglang-omni's custom schedulers, model runners, CUDA Graph lifecycle, and model-specific execution paths to the 0.5.16 contracts.

This revision includes `main` through `a689c2c` and retains the performance and lifecycle changes merged through that revision. Performance is recorded below as a same-host comparison rather than used as a hard blocker for the version bump; this PR does not lower or retune any checked-in performance gate.

## Upstream SGLang breaking changes

The 0.5.12 to 0.5.16 jump crosses several upstream runtime-contract changes. The items below are SGLang compatibility breaks handled by this PR:

- `ForwardBatch.init_new()` now consumes `ScheduleBatch` directly, and scheduler selection returns a `NextBatchPlan`; Omni's custom schedulers and model-runner bridge previously assumed the 0.5.12 batch contract.
- Next-token handoff is split between the GPU `FutureMap` relay used by the next forward pass and a pinned-host reporting path used by CPU output processing; treating either rail as the old single token result can introduce stale tokens or synchronization.
- dLLM `run()` changed from a three-value to a five-value result and added per-request FDFO accept lengths and algorithm state; the old unpacking raises immediately and the new state must survive incomplete denoising blocks.
- Request KV accounting moved from `req.kv_allocated_len` to `req.kv.kv_allocated_len`; Qwen talker rollback must validate the new request shape before mutating any allocation or sequence counters.
- The Ming TP>1 MoE helper removed the `use_reduce_scatter` and `should_allreduce_fusion` parameters; those communication flags now travel through the scoped forward context.
- CUDA Graph setup is phase-specific in the newer runtime, with capture-time limits finalized after graph initialization; Omni-owned graphs and SGLang prefill/decode graphs therefore need an explicit ownership and post-capture sizing boundary.
- Strict `ServerArgs` mutation requires audited post-resolution changes to use `ServerArgs.override()` instead of assigning resolved fields directly.

## Compatibility work

- Adapt `ForwardBatch.init_new()`, `NextBatchPlan`, `FutureMap`, and the split GPU next-token / pinned-host reporting rails used by Omni's custom execution loops.
- Preserve the phase-specific CUDA Graph lifecycle, finalize post-capture KV sizing, and propagate the effective hybrid-SWA token capacity and any post-capture `max_running_requests` cap.
- Port dLLM to the five-value 0.5.16 result contract, including FDFO accept lengths, algorithm state, and retained KV/request-slot lifecycle.
- Port Qwen talker rollback to `req.kv.kv_allocated_len` and make rollback atomic when the upstream request shape is incompatible.
- Port Ming TP>1 MoE communication flags to the 0.5.16 scoped forward context.
- Normalize mixed Qwen3-ASR feature masks across CPU/GPU devices under the 0.5.16 batched multimodal embedding path.
- Route audited post-resolution `ServerArgs` changes through `ServerArgs.override()` so strict configuration mutation mode remains usable.
- Preserve the existing Qwen, Higgs, and MOSS model execution work already carried by this PR, including their CUDA Graph and streaming transport paths.

## Validation

- Full unit suite on the final merged tree: **2670 passed, 29 skipped**.
- Full pre-commit suite and `git diff --check`: passed.
- Independent audits covered dLLM/FDFO state, Qwen FutureMap plus pinned-host token flow, Qwen rollback, Ming TP collectives, post-capture KV sizing, strict `ServerArgs`, Qwen3-ASR masks, and the `main` merge through `a689c2c`.
- GitHub Docs Check, Test Layout, and Lint: passed. The Omni CI GPU jobs were not triggered because the PR does not currently have the maintainer-controlled `run-ci` label.

## Same-host performance

Both arms use the same H100 host, PyTorch 2.11, CI benchmark entry points, datasets, concurrency, CPU/NUMA isolation, and fresh-server lifecycle. TTS rows use the full 1088-sample SeedTTS EN set at concurrency 16 with two single-GPU workers; every arm completed 1088/1088 requests. Qwen Stage 5 uses the checked-in 2000-sample MMSU workload and completed 2000/2000 parseable requests on both arms.

| Workload | `main` (`a689c2c`, SGLang 0.5.12) | This PR (`aa1994b`, SGLang 0.5.16) | Delta |
|---|---:|---:|---:|
| Qwen Stage 5 | 18.597 req/s | 19.214 req/s | +3.3% |
| Higgs non-stream | 12.525 req/s | 13.877 req/s | +10.8% |
| Higgs stream | 12.062 req/s | 15.457 req/s | +28.1% |
| MOSS-Local non-stream | 4.669 req/s | 10.744 req/s | +130.1% |
| MOSS-Local stream | 4.774 req/s | 7.728 req/s | +61.9% |

These are single fresh-server runs and should be read as a regression check, not as stable claims for small deltas. Every row uses the same physical GPU pair and NUMA-isolated CPU sets for both revisions. Both revisions completed every request with no request failures. MOSS-Local stream improved mean TTFC from 1.854 s to 0.875 s; the final run still narrowly missed two existing absolute gates (2.059 s mean latency versus 2.0 s, and 0.482 RTF versus 0.47). The checked-in gates remain unchanged.

## Scope

The version bump does not change benchmark thresholds, seeds, sample counts, or benchmark concurrency. This PR retains the independently reviewed performance work already carried by the branch, including model-specific CUDA Graph and transport paths, bounded CPU thread and preprocessing concurrency, and the Higgs/MOSS process and memory topology changes. The table above measures the resulting branch as a whole; it is not an attribution claim for any individual optimization.

## Commit attribution

The SunskyXH commits and the yxs commits (`da337fd` Higgs shared mempool and `d26412d` MOSS non-EOS tail flush) remain independent commits with their original authors, author dates, messages, and patch contents.

---

<details>
<summary>Previous PR description (archived; no longer describes the current PR scope)</summary>

> The content below is preserved only as historical context. It describes performance experiments and benchmark/deployment changes that were removed from this PR and moved to separate follow-ups. Its measurements use older commits and mixed methodologies; they are not validation results for the current PR. Only the unfurled validation section above is current.

> **WIP.** Probable root causes for the MOSS-TD streaming and Higgs first-packet-tail regressions look identified — see the comments below. Remaining: the Higgs transient startup race.

## Background

sglang-omni runs its own multi-stage pipeline — custom scheduler event loops, model-runner wrappers, multi-process stage topology — on top of the SGLang backbone, pinned at 0.5.12.post1. SGLang 0.5.15 reworked the scheduler/worker boundary into an explicit protocol (FutureMap token relay, `ScheduleBatch` isolation, ambient `ForwardContext`, `inference_mode` execution), which breaks every place where Omni bypasses `Scheduler.run_batch`. This PR adapts those integration points so we can move the pin, stay reachable for upstream fixes, and pick up the backbone's performance work — most workloads see large same-box wins (table below).

Bump SGLang 0.5.12.post1 → 0.5.15.post1 and adapt to the new scheduler/worker execution model.

## Main changes

- New `model_runner/sglang_execution.py`: single compatibility boundary for the 0.5.15 scheduler/worker contract (FutureMap token relay, `ScheduleBatch` isolation, ambient `ForwardContext`). Single-stream by design.
- CUDA graph lifecycle aligned with 0.5.15's phase-aware runtime (separate prefill/decode paths, `init_device_graphs` → `init_cuda_graphs`): decode graph behavior and Omni-owned model graphs preserved; prefill CUDA graph stays opt-in per model — refs #1048.
- API renames tracked 1:1; `Req.inflight_middle_chunks` republished as legacy `is_chunked` for Omni model runners.
- Higgs: prefill CUDA graphs sized by `max_running_requests`, capture-safe codec decode, sampler seed-pool hardening, strict request-identity checks.
- Qwen3-Omni: talker bound to the SGLang execution context; prefill delegation + coalescing; MMSU topology split into pipeline/preprocessing/audio_encoder processes.
- MOSS-Local: process topology, cgroup-aware CPU budgeting (`utils/cpu.py`), stream transport batching.
- Dependency pins match sglang 0.5.15.post1.

## Behavior changes

1. Higgs unseeded requests now sample via the seeded hash sampler with a per-row random seed. The `NO_SEED = -1` sentinel is removed (`-1` cast to `uint64` in the hash sampler was a latent hazard).
2. `OmniScheduler._event_loop_overlap` now raises `NotImplementedError`; the 0.5.15 two-stream overlap contract is not portable to Omni's loop. Normal and async-decode loops are unchanged.
3. Qwen talker predictor: `torch.no_grad` → `torch.inference_mode`.

## Same-box A/B (old = `cf61f23` + 0.5.12.post1, same GPUs/dataset/concurrency)

| Workload | Old | Branch | Δ |
|---|---|---|---|
| Fun-ASR EN | 23.1–27.7 qps | 36.3 | **+31~57%** (WER equal) |
| Higgs TTS streaming (after change) | 10.4–10.8 qps | 13.5–16.1 | **+24~55%**, latency −19% |
| MOSS TTS non-stream / stream | 4.6 / 4.4 qps | 8.2 / 7.2 | **+76% / +66%** (TTFP −49%) |
| MOSS-TD ASR non-stream | ~~28.2 qps~~ → 28.9 | ~~27.9~~ → 31.7 | ~~−1%~~ → **+10%** (row predates `c90cc69`) |
| Qwen S4 / S5 / S6 | 0.36 / 10.9 / 0.40 qps | 0.35 / 27.9–29.6 / 0.66 | −2.5% (noise) / **+156%** / **+66%** |
| MOSS-TD ASR streaming (after change) | 27.3–38.4 qps (bimodal) | 35.6–37.0 stable | at old's best mode, no bimodality |
| Higgs streaming TTFP (after change) | p95 0.77–0.79s | p95 0.57–0.72s | tail recovered (pre-change +26% ❌) |

## Known regressions (probable root causes; details in the comments below)

- **MOSS-TD text streaming −28% — probable root cause identified** (~~`c90cc69`~~ → `c90cc69` + `da90394`; see comment below). Root cause: eight request-builder threads run torch/numpy fbank extraction inside the asr stage process; their GIL share stretches the scheduler's prefill host work to 50–325ms (measured pre-sync 0 / post-sync 0, no token-count correlation), each stall gates streaming decode, and stall-driven completion bursts re-cluster arrivals into a self-sustaining convoy. 0.5.12 sat at the edge of the same regime (clean runs bimodal 38.4/27.3 qps); the 0.5.15 adaptation's slightly higher per-iteration cost made it deterministic. Change: `request_build_max_workers` 8→2 plus `OMP_NUM_THREADS` bounding (the 5c982bb treatment). Post-change same-box A/B: stream 35.6–37.0 qps stable, non-stream 31.7 vs 28.9.
- **Higgs streaming TTFP tail (p95 +26%, p99 +76%) — probable root cause identified** (`39bded1`; see comment below). Same mechanism as MOSS-TD: all four Higgs stages share one pipeline process, and the 16-thread preprocessing pool's GIL share stalls the tts_engine scheduler on arrival bursts. Change: preprocessing pool 16→2 workers plus the same OMP bound. Post-change same-box A/B (two pairs): TTFC p95 0.57–0.72s vs old 0.77–0.79s, p99 0.99–1.58 vs 1.41, qps 13.5–16.1 vs 10.4–10.8; the second pair passes every CI threshold on this box.
- **Higgs startup**: transient KV-memory race — identical launch failed once (`mem-fraction 0.85 < required 0.883`), passed on retry with idle GPUs.

## Tests

Unit suite ~~2455~~ → 2513 passed / 29 skipped; pre-commit clean over the full diff. New tests pin the sampler seed invariants, prefill request-identity, vocoder compile/graph exclusivity, and the `is_chunked` republish; removed tests that asserted unreachable guards.

</details>


junliu-mde:
There might still be some weird benchmark parameter changes that shouldn't make it into the PR, but we can clean them up later.

junliu-mde:
About MOSS-TD regression:

The asr stage runs 8 request-builder threads doing torch/numpy fbank extraction in the same process as the scheduler. Their GIL share stretches the scheduler's prefill host work ~9x — probes showed 50-325ms of pure host time inside `ModelRunner.forward` on EXTEND batches. 0.5.12 build had the same problem.

Repeated runs on 0.5.12 showed a bimodal distribution (38.4 vs 27.3 qps, no middle ground).

  - **High peak (~38.4 qps)**: request arrivals stay spread out, so the fbank builder threads and the scheduler's prefill work interleave on the GIL without piling up.
  - **Low peak (~27.3 qps)**: one early prefill stall causes a batch of requests to complete together, so the next batch arrives together.

junliu-mde:
About the Higgs first-packet tail:

All four Higgs stages share one `pipeline` process, so the 16-thread preprocessing pool (tokenize / resample / delay-pattern encode — all GIL-held CPU torch work) competes with the tts_engine scheduler thread for the GIL, and each CPU torch op otherwise spawns a machine-sized OMP team against the cgroup CPU quota. Same mechanism as the MOSS-TD regression above.

Why it shows up as a tail rather than a throughput drop: TTFC is gated by prefill completion, so a scheduler stalled mid-prefill delays the first chunk directly. The pool only saturates on arrival bursts, so the worst stalls land on the bursts and hit p95/p99 (+27-36% vs 0.5.12) while the median barely moves. Unlike MOSS-TD it never locks into a self-sustaining convoy — it just takes a hit per burst.

Related PR: https://github.com/sgl-project/sglang-omni/pull/1180 https://github.com/sgl-project/sglang-omni/pull/1159

yxs:
Great work.
Each launch lands wherever the OS scheduler puts it, and landing on the far socket costs a lot. The bimodal numbers in this PR are likely what an unpinned run looks like.
Consider re-running with the server pinned to the GPU's socket and the client on the other, before the topology and worker-count changes land.

LauraGPT (commented):
Fun-ASR scoped review on exact head `da337fdf78b2311262ce4605426a54849cb2f34c`:

- Installed the exact PR dependency line: `sglang==0.5.15.post1`, `transformers==5.12.1`.
- `pytest -q tests/unit_test/fun_asr`: **45 passed**.
- Shared execution/scheduler/bootstrap tests touched by the SGLang upgrade: **58 passed**.
- Production-shape H100/bfloat16 validation with the real 50 encoder + 20 timestamp blocks, hidden size 512 and projector hidden size 1024: mixed-length batched outputs matched individual inference in both batch orders. The short item max/mean absolute differences were `0.039062/0.004765`, the long item was exact, and padded outputs were strictly zero even when the invalid input region was filled with `1024`.
- `pre-commit run --from-ref cf61f234e5a2cc7f63a24bdef84b446bd9a42f74 --to-ref HEAD`: all applicable hooks passed.

I found no Fun-ASR-side blocker in this head. The visible `omni-ci-gate` failure is the expected draft guard, with ASR CI skipped while the PR remains WIP, so I am leaving a scoped comment rather than approving the entire release bump.

LauraGPT:
Revalidated the updated exact head `44597a8ab7dae901b357a21f8ef5a299256ab1be` after it advanced from the previously reviewed `da337fdf78b2311262ce4605426a54849cb2f34c`.

The exact delta is limited to five MOSS-TD/Qwen3-Omni files and does not change `sglang_omni/models/fun_asr`, its tests, or the shared model-runner/scheduler paths covered by the prior review. Fresh validation on the new head:

- `pytest -q tests/unit_test/fun_asr`: **45 passed**
- `pytest -q tests/unit_test/qwen3_omni/test_thinker_async_decode_flag.py`: **3 passed**
- `pre-commit run --from-ref da337fdf... --to-ref 44597a8a...`: all applicable hooks passed

The current `omni-ci-gate` failure was also checked from job `89826043596`; it exits only because the PR is still draft (`PR is draft. Blocking Omni CI.`). I still find no Fun-ASR-side blocker in this exact head.

junliu-mde:
> Revalidated the updated exact head `44597a8ab7dae901b357a21f8ef5a299256ab1be` after it advanced from the previously reviewed `da337fdf78b2311262ce4605426a54849cb2f34c`.
> 
> The exact delta is limited to five MOSS-TD/Qwen3-Omni files and does not change `sglang_omni/models/fun_asr`, its tests, or the shared model-runner/scheduler paths covered by the prior review. Fresh validation on the new head:
> 
> * `pytest -q tests/unit_test/fun_asr`: **45 passed**
> * `pytest -q tests/unit_test/qwen3_omni/test_thinker_async_decode_flag.py`: **3 passed**
> * `pre-commit run --from-ref da337fdf... --to-ref 44597a8a...`: all applicable hooks passed
> 
> The current `omni-ci-gate` failure was also checked from job `89826043596`; it exits only because the PR is still draft (`PR is draft. Blocking Omni CI.`). I still find no Fun-ASR-side blocker in this exact head.

Yep I realized that, trying to split into multiple PRs, keep bump version branch clean.

AkazaAkane:
sglang main updates to v0.5.16 two days ago, maybe consider to bump to 5.16 directly given that most thing remain unchanged. https://github.com/sgl-project/sglang/releases/tag/v0.5.16

junliu-mde:
> sglang main updates to v0.5.16 two days ago, maybe consider to bump to 5.16 directly given that most thing remain unchanged. https://github.com/sgl-project/sglang/releases/tag/v0.5.16

In fact I checked diff b/w 0.5.15 and 0.5.16 of backbon sglang, major diff in LM adaption (GLM5.2 and Inkling), but sgl-kernel also has major refactor, I'm not sure about the impact. 

Since adaption on gaming GPU (4090/5090) is blocked by this PR, I prefer to bump sgl-kernel next time. cc @zhaochenyang20 How do you think.

Hayden727:
Linked to #1120. I checked the actual `pyproject.toml` patch: this PR owns the SGLang/Transformers/Kernels version update, while #1199 owns the higher-priority consumer-GPU dependency baseline (including optional FA4) and should merge first. Please rebase #1183 on #1199 afterward, preserve the optional-FA4 behavior, and retain only the remaining version updates.

stephenkgli:
https://github.com/sgl-project/sglang-omni/blob/7e837c22420c32e8894756feabb1dbfb224ab8fc/sglang_omni/models/arkasr/stages.py#L96-L97

arkasr should also use`init_cuda_graphs()`

pyproject.toml
```diff
@@ -21,24 +21,24 @@ dependencies = [
     "torch==2.11.0",          # PyTorch for tensor operations
     "torchvision==0.26.0",    # torchvision for vision processor, pinned version based on torch 2.11.0
     "accelerate>=0.27.0",     # init_empty_weights for meta initialization
-    "transformers==5.6.0",   # HF models/tokenizers
+    "transformers==5.12.1",  # Match SGLang 0.5.16
```
junliu-mde, line 24: ~~hf transformers 5.13.0 fixed this issue but sglang pinned 5.12.1, how should be do here?~~

Ratish replied me, ignore.

sglang_omni/models/higgs_tts/config.py
```diff
@@ -74,12 +79,23 @@ def mem_fraction_role_to_stage(cls) -> dict[str, str]:
         ),
         StageConfig(
             name="vocoder",
-            process="vocoder",
+            # Keep the LM and vocoder in one CUDA context by default.  Splitting
+            # them into same-GPU processes time-slices the H100 at ordinary
+            # serving concurrency and prevents decode/vocoder overlap.
+            process="pipeline",
```
junliu-mde, line 94: Yes, re-measured

tests/unit_test/higgs_tts/test_pipeline.py
```diff
@@ -1646,6 +1971,10 @@ def decode(self, codes_BNT):
     codec.model = FakeModel()
     codec.device = torch.device("cpu")
     codec._dtype = torch.float32
+    codec._decode_cuda_graphs = {}
+    codec._decode_cuda_graph_hits = 0
+    codec._decode_cuda_graph_misses = 0
+    codec._decode_cuda_graph_missed_shapes = set()
```
AkazaAkane, line 2058: The new single-flight guard looks correct, but this fake constructs `HiggsAudioCodec` with `object.__new__()` and therefore bypasses the lock initialization in `__init__`. The current unit job fails 3/3 here with `_decode_single_flight_lock` missing.

Maybe add:

```python
codec._decode_single_flight_lock = threading.Lock()
```

to this helper, matching the fake codecs in `test_audio_codec.py`, or construct the codec through its initializer?

pyproject.toml
AkazaAkane: test_vocoder_cuda_graph.py still passes the Hugging Face cache snapshot directory directly:

AutoModel.from_pretrained(snaps[0], trust_remote_code=True)

Under Transformers 5.12.1, that local snapshot path can resolve trusted-code symlinks into blobs/, where relative imports such as configuration_moss_audio_tokenizer.py are no longer found. The latest unit run stopped earlier on the new Higgs fake-codec failure, so it did not reach this test again.

sglang_omni/models/higgs_tts/request_builders.py
```diff
@@ -232,16 +230,18 @@ def request_builder(payload: StagePayload) -> HiggsSGLangRequestData:
...
+            apply_higgs_result(state, data)
+            if data.engine_start_s:
+                state.engine_time_s = _perf_counter() - data.engine_start_s
+            return StagePayload(
+                request_id=payload.request_id,
+                request=payload.request,
+                data=state.to_dict(),
+            )
+        finally:
+            model.reset_request(payload.request_id)
```
JiaxinD: This finally is the only place the row is released, but in stream_output the flush runs before the adapter is entered, so a raising flush skips it: the except only emits the error and continues, and on_request_finished sits outside the try entirely. After max_running_requests such failures _free_rows is empty and the engine wedges, so the reset wants to be at the stream_output level.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -1032,7 +1270,18 @@ def stream_output(self, reqs, return_logprob=False, skip_req=None):
...
+            data = None
+            with self._request_admission_lock:
+                is_aborted = rid in self._aborted_request_ids
+                if not is_aborted:
+                    data = req._omni_data
+                    if data is None:
+                        raise RuntimeError(
+                            f"Terminal request {rid!r} has no scheduler request data"
+                        )
+                    _detach_request_data(req)
```
JiaxinD: Detaching here, before on_request_finished and the flush, makes _find_request_data return nothing for a request still in running_batch, so trailing chunks and stream_done land in _pending_stream_chunks / _pending_stream_done, which are only cleared in _admit_request or abort. A normally finished request never clears them, and if the client reuses the rid those parked chunks replay into the new request.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -1032,7 +1270,18 @@ def stream_output(self, reqs, return_logprob=False, skip_req=None):
...
 
             rid = req.rid
-            if rid in self._aborted_request_ids:
+            data = None
+            with self._request_admission_lock:
+                is_aborted = rid in self._aborted_request_ids
+                if not is_aborted:
+                    data = req._omni_data
+                    if data is None:
+                        raise RuntimeError(
```
JiaxinD: This raise leaves stream_output, which upstream process_batch_result does not wrap, so it kills the scheduler thread while the stage keeps accepting requests. Its discriminator is the LRU-bounded _aborted_request_ids, so an evicted rid trips it on a long-lived server; a logger.error plus continue would degrade instead.

sglang_omni/models/higgs_tts/audio_codec.py
```diff
@@ -108,6 +138,14 @@ def __init__(
...
         self.device = device
         self._dtype = next(model.parameters()).dtype
+        self._decode_cuda_graphs: dict[int, _DecodeCudaGraph] = {}
+        self._decode_cuda_graph_hits = 0
+        self._decode_cuda_graph_misses = 0
+        self._decode_cuda_graph_missed_shapes: set[int] = set()
+        # Graphs for different frame counts share one private CUDA memory pool.
+        # Their temporary/output storage may therefore alias and decode must
+        # remain single-flight until the selected output has reached the host.
+        self._decode_single_flight_lock = threading.Lock()
```
JiaxinD, line 148: unit-test is red on this head: test_decode_batch_buckets_by_length builds the codec with object.__new__ and never sets this attribute, so decode_batch raises AttributeError (348 passed, 1 failed). Creating the lock lazily, or in the fake, fixes it.

sglang_omni/models/higgs_tts/audio_codec.py
```diff
@@ -239,34 +361,87 @@ def _batch_fn(batch_items: list[torch.Tensor]) -> list[torch.Tensor]:
...
-        return self.model.decode(codes_BNT).audio_values.squeeze(0).squeeze(0).cpu()
+        if decode_graphs and total_graph_lookups % 1024 == 0:
+            logger.info(
+                "Higgs codec decode CUDA graph stats: hits=%d misses=%d",
+                self._decode_cuda_graph_hits,
+                self._decode_cuda_graph_misses,
+            )
+        return audio.squeeze(0).squeeze(0).cpu()
+
+    def _acquire_decode_single_flight(self) -> None:
```
JiaxinD: acquire(blocking=False) plus raise turns any real overlap into a failed request rather than serialization, and in _vocode_payloads it fails every payload of the coalesced batch, not just the contended one. Capture is also unguarded: it swaps _decode_cuda_graphs and the pool wholesale while get_or_load_codec hands the same instance to any other stage in the process.

sglang_omni/models/moss_tts_local/config.py
```diff
@@ -62,9 +85,10 @@ def _stages(*, codec_device: str, colocated: bool) -> list[StageConfig]:
         ),
         StageConfig(
             name="vocoder",
-            process="pipeline",
+            process="vocoder" if colocated else "pipeline",
```
JiaxinD, line 88: With the vocoder in its own process its codec copy leaves the AR stage's process accounting, so nothing enforces the 0.18 while 0.15 + 0.67 + 0.18 already sits exactly on the default 1.0 placement ceiling. Preprocessing's 0.15 is also double counted inside the AR process, so the validator reports a full card at about 0.85 real commitment: it rejects a legitimate 0.19 vocoder while not guarding the vocoder-process OOM that actually can happen.

sglang_omni/models/higgs_tts/model_runner.py
```diff
@@ -439,12 +472,23 @@ def _build_prefill_input_embeds(
...
             codes = torch.tensor(codes_rows, dtype=torch.long, device=device)
-            consumed = data.num_ref_codes_consumed
+            # Radix reuse can begin this request's live extension after a
+            # cached reference-audio prefix. Derive the code row from the
+            # absolute prompt position instead of request-local progress.
+            consumed = sum(
+                token == AUDIO_PLACEHOLDER_ID
+                for token in data.req.origin_input_ids[: data.req.extend_range.start]
+            )
+            if consumed + n_placeholders > len(codes_rows):
```
JiaxinD, line 444: This coverage check cannot see cross-request reuse: reference audio is only AUDIO_PLACEHOLDER_ID repeated, so two requests with the same text and the same reference length have identical origin_input_ids, and with radix matching on the second one's extension starts after the cached placeholder block. consumed + n_placeholders still fits, nothing raises, and it generates on the first request's reference KV, i.e. the wrong voice, silently.

tests/unit_test/higgs_tts/test_pipeline.py
Ratish1: done

sglang_omni/models/higgs_tts/request_builders.py
Ratish1: done

sglang_omni/scheduling/omni_scheduler.py
Ratish1: done

sglang_omni/models/higgs_tts/audio_codec.py
Ratish1: done

sglang_omni/models/higgs_tts/audio_codec.py
Ratish1: done

sglang_omni/scheduling/omni_scheduler.py
Ratish1: done

sglang_omni/models/higgs_tts/model_runner.py
Ratish1: fixed

sglang_omni/models/moss_tts_local/config.py
Ratish1: fixed

yxs (approved):

Ratish1 (approved):

Hayden727:
Current-head revalidation after the fixes (`base=abbb7e48`, `head=4d5c19d7`):

### Real dLLM checkpoint E2E

I downloaded and served `inclusionAI/LLaDA2.0-Uni` on one H200 with the PR's SGLang 0.5.16 stack. The first real text request exposed a remaining API-contract break:

```text
Req._init_fill_ids_for_dllm(): list + array.array
TypeError: can only concatenate list (not "array.array") to list
```

SGLang 0.5.16 declares `Req.origin_input_ids` as `array[int]` and constructs `output_ids` and dLLM masks as `array("q")`; the Omni LLaDA builder was still passing a Python list. Commit `4d5c19d7` now converts prompt IDs to `array("q")` at that adapter boundary and adds a regression test through the real upstream dLLM fill path.

After the fix:

- real text request: HTTP 200, response `Blue`, 29 prompt + 2 completion tokens;
- two concurrent real-checkpoint requests: both HTTP 200 (about 7.3 s each after warmup);
- focused dLLM tests: **9 passed**;
- full final-head unit suite: **2830 passed, 1 skipped**;
- final-diff pre-commit: passed.

### Same-host performance rerun

Environment: 8× H200 host; server workers on physical GPUs 0/1; `taskset -c 0-55,112-167`; base uses SGLang `0.5.12.post1`, PR uses `0.5.16`; fresh managed-router lifecycle; Qwen uses the checked-in 2000-sample MMSU job, and each TTS arm completed 1088/1088 SeedTTS requests at concurrency 16 with zero request failures.

| Workload | Current base | PR | Delta |
|---|---:|---:|---:|
| Qwen Stage 5 | 39.225 req/s | 43.276 req/s | +10.3% |
| Higgs non-stream | 17.735 req/s | 18.041 req/s | +1.7% |
| Higgs stream | 17.572 req/s | 17.966 req/s | +2.2% |
| MOSS-Local non-stream | 10.339 req/s | 12.122 req/s | +17.2% |
| MOSS-Local stream, pair 1 | 8.185 req/s | 8.596 req/s | +5.0% |
| MOSS-Local stream, pair 2 | 8.131 req/s | 8.275 req/s | +1.8% |

Additional observations:

- Qwen accuracy was 70.1% → 71.1%; both arms missed this H200 host's checked-in absolute speed gates, while all 2000 responses were parseable and successful.
- Higgs stream mean TTFC was essentially flat (0.3325 s → 0.3335 s).
- MOSS non-stream mean latency improved 1.540 s → 1.314 s and p95 improved 2.073 s → 1.843 s.
- MOSS stream mean TTFC improved in both pairs (1.006–1.032 s → 0.883–0.963 s), but the tails regressed in both: latency p95 2.601–2.679 s → 2.912–3.281 s, and TTFC p95 1.530–1.560 s → 1.859–2.259 s. Both PR runs still passed the checked-in TTS test.

The Qwen/Higgs/MOSS non-stream arms were collected on predecessor `b04c2a68`; there are no changed Qwen/Higgs/MOSS/benchmark/test paths between `b04c2a68` and final head `4d5c19d7`. The second MOSS stream PR arm is on exact final head.

The PR-body H100 table compares an older `a689c2c` base. Current `main` already contains later TTS work, so its large deltas should not be presented as the incremental current-base result. The H200 rerun finds no throughput regression, but it does **not** reproduce a MOSS-stream tail improvement; the two paired p95 results above should remain visible as the performance caveat.

Hayden727 (approved):
Re-reviewed exact head 4d5c19d7 against base abbb7e48. The MOSS placement contract is now fixed in base, and a real inclusionAI/LLaDA2.0-Uni E2E exposed and then verified the array('q') Req adapter fix in 4d5c19d7. Final validation: 2830 unit tests passed (1 skipped), focused dLLM 9/9 passed, final-diff pre-commit passed, and real single/concurrent dLLM requests returned HTTP 200. Performance rerun details are in the linked PR comment; MOSS streaming tail latency remains a documented caveat, but checked-in gates pass and I do not find a remaining correctness blocker.

PR #1184, opened 2026-07-25
fix(qwen3-tts): expose finish_reason on /v1/audio/speech

### LINKED ISSUE #1178: [Bug] Qwen3-TTS drops length finish reason before /v1/audio/speech
## Checklist

- [x] 3. I have provided the environment information and a minimal reproducible demo below.
- [x] 4. This is a bug report rather than a question.

## Describe the bug

`/v1/audio/speech` never reports how generation terminated. A response cut short at the generation budget looks exactly like one that ended on codec EOS: HTTP 200 with decodable audio in both cases.

The scheduler does record it, as `{"type":"stop","matched":2150}` for codec EOS or `{"type":"length", ...}` for the budget. But `apply_sglang_qwen3_tts_result()` rebuilds the `StagePayload` without `data.finish_reason`, so the reason never reaches the vocoder result, `SpeechResult`, or the HTTP response. The `/v1/audio/speech/stream` WebSocket path already exposes a per-chunk finish reason; only the HTTP paths lose it.

Generation itself is bounded and correct: `init_req_max_new_tokens()` clamps the budget before the KV-capacity check, and no serving-side OOM occurs. Only the reporting is missing.

The cost shows up in diagnosis. In one benchmark run, 174 of 55,944 requests stopped at exactly `max_new_tokens=2048`, or 163.84 seconds of audio. The benchmark recorded them as successful, which was right for what it measured: the responses were available and decodable. But because no response carried a termination state, the first visible symptom was an out-of-memory failure downstream while scoring the abnormally long audio, and identifying the affected requests afterwards meant reconstructing the termination state offline from completion-token counts and audio durations across the whole run.

A request that stops at its budget should stay a successful response but expose `finish_reason=length`, as elsewhere in this repository. `/v1/rollout/generate` returns a typed `GenerateFinishReason(type, length)`, and chat completions carry `finish_reason`. Ming-TTS normalizes the reason in its `result_adapter`, at the same position where Qwen3-TTS drops it, and even infers `length` when the scheduler reason is absent, though its reason does not reach the response either because the client's single-terminal result branch never reads one. The same gap covers the server-side clamp from a requested budget down to the effective context/KV budget: that clamp is applied correctly, but the response does not show it either.

## Reproduction

```bash
curl -sS -D - -o /tmp/speech.wav http://127.0.0.1:8000/v1/audio/speech \
  -d '{"model":"Qwen/Qwen3-TTS-12Hz-0.6B-Base","input":"hello","voice":"default","max_new_tokens":1}'
```

Response: HTTP 200, decodable WAV, headers `X-Prompt-Tokens` / `X-Completion-Tokens` / `X-Engine-Time` — and nothing indicating that generation stopped at the budget after one token. The scheduler recorded `{"type":"length","length":1}` for the same request.

The same loss appears across the whole budget range. With `max_req_len=8191`, a 65-token prompt, and an available generation budget of 8125:

| Requested `max_new_tokens` | Effective | Scheduler result | HTTP response |
|---:|---:|---|---|
| 1 | 1 | `{"type":"length","length":1}` | 200 with audio, no finish reason |
| 8125 | 8125 | `{"type":"stop","matched":2150}` | 200 with audio, no finish reason |
| 10000 | 8125 | `{"type":"stop","matched":2150}` | 200 with audio, no finish reason |

The 10000-token request was silently clamped to 8125 and produced a token sequence identical to the at-capacity request. Neither the clamp nor the scheduler's finish reason is visible to the caller.

## Environment

- Ubuntu 24.04.3 LTS
- 1× NVIDIA GeForce RTX 5090, 32,607 MiB, SM120
- NVIDIA driver 595.71.05
- CUDA runtime 13.2
- PyTorch 2.11.0+cu130
- SGLang-Omni commit `043f9ae6bd10525eeec8c6d8bc114dbba3f7b731`
- Source tree `de0737f11c37272e12998744fcd6d027cb7de000`
- Qwen3-TTS model revision `5d83992436eae1d760afd27aff78a71d676296fc`
- SeedTTS revision `27f4c1adee83b5b29b7c4b375f6b976324bda308`

Peterren:
## Motivation

A Qwen3-TTS response from `/v1/audio/speech` never reports how generation
terminated. A response cut short at the generation budget looks exactly like
one that ended on codec EOS: HTTP 200 with decodable audio in both cases, and
no signal to tell them apart.

The scheduler already records it (`{"type":"length"}` vs `{"type":"stop"}`) and
`OmniScheduler` normalizes it onto `data.finish_reason` before the result
adapter runs, but `apply_sglang_qwen3_tts_result` rebuilds the `StagePayload`
without it, so the reason never reaches the vocoder result, `SpeechResult`, or
the HTTP response. The `/v1/audio/speech/stream` WebSocket path already exposes
a per-chunk finish reason; only the HTTP path loses it.

The practical cost: in one benchmark run, 174 of 55,944 requests stopped at
exactly `max_new_tokens` but were indistinguishable from clean completions, and
only surfaced downstream as an out-of-memory failure while scoring the
abnormally long audio.

## Modifications

- `qwen3_tts/request_builders.py`: `apply_sglang_qwen3_tts_result` now threads
  the finish reason through the result. It reads the scheduler-normalized
  `data.finish_reason`, falls back to the SGLang `Req.finished_reason`, and
  infers `length` when the reason is absent but generation hit the token budget
  (mirrors `ming_tts/engine_io.py`).
- `qwen3_tts/payload_types.py`: add `finish_reason` to `Qwen3TTSState` so it
  survives the vocoder stage's `load_state`/`store_state` round-trip (this is
  how `MingTTSState` and `S2ProState` already carry it).
- `client/types.py` + `client/client.py`: carry `finish_reason` on
  `SpeechResult` (`Client.generate` already reads it from the terminal payload).
- `serve/openai_api.py`: emit `X-Finish-Reason` from `/v1/audio/speech`,
  alongside the existing `X-Prompt-Tokens` / `X-Completion-Tokens` headers.

A successful budget-truncated request stays HTTP 200 and now also reports
`finish_reason=length`, matching the typed reasons elsewhere in the repo
(`/v1/rollout/generate`, chat completions).

## Related Issues

Closes #1178

## Accuracy Test

No model-side / numerical changes; this only propagates an existing
scheduler-computed field to the HTTP response.

## Benchmark & Profiling

No performance impact (metadata plumbing only).

Manual end-to-end verification on a single A100 with
`Qwen/Qwen3-TTS-12Hz-0.6B-Base` (voice-cloning Base path):

| Input | `max_new_tokens` | `X-Completion-Tokens` | `X-Finish-Reason` |
|---|---:|---:|---|
| long sentence | 8 | 8 | `length` |
| short input | 2048 | 11 | `stop` |
| same long sentence | 2048 | 49 | `stop` |

Same input with a tiny budget reports `length` and with an ample budget stops
naturally at `stop`, so truncation is now distinguishable from completion.

## Checklist

- [x] Format your code according with pre-commit (ruff / black / isort clean).
- [x] Add unit tests (result-adapter reason resolution + state round-trip).

### DIFF
--- a/sglang_omni/client/client.py
+++ b/sglang_omni/client/client.py
@@ -255,6 +255,7 @@ class Client:
             format=actual_format,
             sample_rate=sample_rate,
             usage=last_chunk.usage if last_chunk else None,
+            finish_reason=last_chunk.finish_reason if last_chunk else None,
         )
 
     # ------------------------------------------------------------------
--- a/sglang_omni/client/types.py
+++ b/sglang_omni/client/types.py
@@ -229,6 +229,7 @@ class SpeechResult:
     format: str
     sample_rate: int | None = None
     usage: UsageInfo | None = None
+    finish_reason: str | None = None
 
 
 class ClientError(Exception):
--- a/sglang_omni/models/qwen3_tts/payload_types.py
+++ b/sglang_omni/models/qwen3_tts/payload_types.py
@@ -32,3 +32,4 @@ class Qwen3TTSState(DeclarativeStateBase):
     audio_codes: Any | None = wire(None, codec="tensor_list")
     ref_code_len: int = wire(0, emit="truthy", codec="int")
     audio_samples: Any | None = wire(None, codec="tensor_list")
+    finish_reason: str | None = None
--- a/sglang_omni/models/qwen3_tts/request_builders.py
+++ b/sglang_omni/models/qwen3_tts/request_builders.py
@@ -1059,18 +1059,54 @@ def apply_sglang_qwen3_tts_result(
     else:
         codes = torch.empty((0, 0), dtype=torch.long)
 
+    finish_reason = _resolve_qwen3_tts_finish_reason(data)
+
+    result_data: dict[str, Any] = {
+        "audio_codes": codes,
+        "ref_code_len": data.ref_code_len,
+        "prompt_tokens": data.ref_code_len,
+        "completion_tokens": len(data.output_codes),
+        "engine_time_s": time.perf_counter() - data.engine_start_s,
+        "sample_rate": 24000,
+    }
+    if finish_reason is not None:
+        result_data["finish_reason"] = finish_reason
+
     return StagePayload(
         request_id=payload.request_id,
         request=payload.request,
-        data={
-            "audio_codes": codes,
-            "ref_code_len": data.ref_code_len,
-            "prompt_tokens": data.ref_code_len,
-            "completion_tokens": len(data.output_codes),
-            "engine_time_s": time.perf_counter() - data.engine_start_s,
-            "sample_rate": 24000,
-        },
+        data=result_data,
     )
+
+
+def _resolve_qwen3_tts_finish_reason(
+    data: Qwen3TTSSGLangRequestData,
+) -> str | None:
+    """Normalize the generation termination reason ("stop"/"length"/"abort")."""
+    raw = getattr(data, "finish_reason", None)
+    if raw is None and getattr(data, "req", None) is not None:
+        req_reason = getattr(data.req, "finished_reason", None)
+        if req_reason is not None and hasattr(req_reason, "to_json"):
+            raw = req_reason.to_json().get("type")
+        elif req_reason is not None:
+            raw = str(req_reason)
+
+    normalized = str(raw).lower() if raw is not None else None
+    if normalized is not None:
+        if "length" in normalized:
+            return "length"
+        if "abort" in normalized:
+            return "abort"
+        if "error" in normalized:
+            return "error"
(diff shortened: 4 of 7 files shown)


Hayden727:
Maintainer triage: #1185 is the selected landing path for #1178. It now covers the full scheduler → Qwen3-TTS state → vocoder → Client.speech path, the non-streaming X-Finish-Reason header, per-item batch JSON, focused tests, and public docs. This PR overlaps only the HTTP exposure layer and is currently conflicted, so there is no value in rebasing both implementations. Please treat #1184 as superseded by #1185; we will close it after #1185 lands. Thanks for the contribution and the API-level direction.

PR #1185, opened 2026-07-25
[Qwen3-TTS] Preserve termination reason on speech responses

### LINKED ISSUE #1178: [Bug] Qwen3-TTS drops length finish reason before /v1/audio/speech

lijrjyan:
## Motivation

Qwen3-TTS speech responses carry no termination state, so a length-capped partial completion is indistinguishable from natural codec EOS: both return HTTP 200 with decodable audio. The scheduler already produces the reason; the model result adapter drops it before the client can carry it into the speech response. Details and reproduction: #1178.

## Modifications

**Qwen3-TTS drops a reason the scheduler provides.** `apply_sglang_qwen3_tts_result()` rebuilds the `StagePayload` without `data.finish_reason`. The new `_qwen3_tts_finish_reason()` follows the Ming-TTS `result_adapter` precedent: take the scheduler reason, fall back to `req.finished_reason`, normalize `length` / `abort` / `error`, and infer `length` when generation reached its budget without a reason. `Qwen3TTSState` carries the field, so it rides the existing state round trip through the vocoder stage and no stage code changes.

Exposure follows existing conventions rather than inventing one. `SpeechResult` gains the field; `/v1/audio/speech` sets an `X-Finish-Reason` header next to the existing `X-Prompt-Tokens` / `X-Completion-Tokens` / `X-Engine-Time`, because the OpenAI specification defines no termination state for an endpoint whose body is raw audio; `SpeechBatchResult` gains a `finish_reason` field for the batch JSON path. The values (`stop`, `length`) match OpenAI chat completions and the WebSocket path.

Streaming `/v1/audio/speech` is intentionally not covered: it returns raw PCM and its headers are sent before generation completes, which is a property of the endpoint rather than of any pipeline.

## Related Issues

Closes #1178.

The EOS-starvation investigation that exposed this loss is tracked separately and is not a dependency: upstream context is #599 and #570.

## Accuracy Test

No generation semantics change. Requests that previously succeeded still succeed with the same audio.

N/A for model accuracy.

Verified at the HTTP boundary with paired base-versus-this-PR server runs on one RTX 5090 (SM120), Qwen3-TTS-12Hz-0.6B-Base, identical requests, reference audio and seed:

| Case | base | this PR |
|---|---|---|
| natural EOS | no header, `X-Completion-Tokens: 32` | `X-Finish-Reason: stop` |
| `max_new_tokens=1` | no header, `X-Completion-Tokens: 1` | `X-Finish-Reason: length` |
| `max_new_tokens=10000`, clamped to the effective budget | no header, `X-Completion-Tokens: 32` | `X-Finish-Reason: stop` |
| batch JSON | `finish_reason: null` on both items | `stop` and `length` per item |
| streaming | no header | no header |

## Benchmark & Profiling

N/A: no throughput or latency claim; the change adds one derived string per request.

## Checklist

- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed. (`docs/cookbook/qwen3_tts.md` documents the non-streaming header and batch field.)
- [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed. (N/A: reporting-only change.)

## CI

Original functional validation on the pre-rebase base (`cf61f234e`), using the repository's pinned hooks:

```text
suite            base                      feature head
qwen3_tts        73 passed                 78 passed        (+5)
serve            2 failed, 253 passed      2 failed, 253 passed
pipeline         9 failed, 277 passed      9 failed, 277 passed
```

Maintainer rebase and final refinements:

- rebased onto `fc08b9a9a3135e6fee477b208427bc73bab85228`
- reviewed head: `6c64acf916ac76828b93353ff868bdab6f9a1594`
- client plus speech API/batch regression suite: **91 passed**
- explicit `black-jupyter`, `autoflake`, `isort`, and `check-ast` pre-commit hooks: passed
- changed Python modules: `py_compile` passed
- independent code review: no Critical or Important findings; mergeable

GPU-backed behavior is covered by the paired RTX 5090 HTTP runs above. GitHub CI was retriggered on the rebased head.


Hayden727 (approved):
Re-reviewed the rebased finish-reason propagation: terminal status is preserved through the real speech response path, invalid/empty termination cases are covered, and the scoped test suite passed. No remaining code finding; CI/repository merge policy is the remaining gate.

PR #1186, opened 2026-07-26
[CI] Add Stage 5 TTS MPS validation

JiaxinD:
# [CI] Add Stage 5 TTS MPS validation

## Motivation

The canonical TTS pipeline validates Higgs or MOSS-TTS Local with ordinary DP2 across two H100s. Replacing it with MPS would mix resource-sharing behavior into the correctness reference and make regressions harder to attribute.

This PR keeps the existing four-stage TTS DAG unchanged and adds **Stage 5: MPS**, a reusable validation boundary after Stage 4. The new job proves that the selected model can run two production Router workers concurrently on one H100, retain the existing quality gates, and leave the GPU clean before Qwen3-ASR evaluation.

## Before and after

```mermaid
flowchart TB
  subgraph Before["Before"]
    direction LR
    B1["Stage 1<br/>ordinary DP2 / 2 H100s"] --> B2["Stage 2"] --> B3["Stage 3"] --> B4["Stage 4"]
  end
  subgraph After["After"]
    direction LR
    A1["Stage 1<br/>unchanged"] --> A2["Stage 2<br/>unchanged"] --> A3["Stage 3<br/>unchanged"] --> A4["Stage 4<br/>unchanged"] --> A5["Stage 5: MPS<br/>TTS non-streaming / DP2 / 1 H100"]
  end
  Before ~~~ After
```

Stage 5 is a CI boundary, not a fifth TTS transformation. Its outputs are evidence-only and cannot become Stage 3 inputs.

| Concern | Canonical stages | Stage 5: MPS |
|---|---|---|
| Placement | Original DP2 on two H100s | Two replicas on one H100 through a private MPS namespace |
| Weights | One copy per worker | Independent copies (`WEIGHT_SHARE=0`) |
| Routing | Production Router | Same Router with two MPS workers |
| Quality | Existing WER, SIM, UTMOS, audio integrity | Same WER and audio-integrity gates; speaker similarity carries an MPS-calibrated baseline (below) |
| Ordering | Stage 1 -> 2 -> 3 -> 4 | Serialized after Stage 4 |

## What changed

* Register TTS non-streaming under `stage-5-mps` and reuse CPU preflight's `run-higgs` / `run-moss` decision. Labels are mutually exclusive; unlabeled runs use a rerun-stable SHA-256 digest of `github.run_id`.
* Reuse `examples/mps_dp/launch.sh` with fixed `N=2`, one private MPS namespace, equal resolved KV capacity, verified attachment, and model-specific H100 configs.
* Require both Router workers to serve real canonical requests, then run a bounded eight-request overlap canary using server-side `CLOCK_MONOTONIC` model-path events.
* Gate performance on worst-of-five references in `tests/test_ci/tts_mps_ci_config.py`, calibrated through `.claude/skills/tune-ci-thresholds` and registered there so the next person can recalibrate them.
* Stop the Router and complete authoritative MPS teardown before Qwen3-ASR starts. Dirty process, port, state, or GPU-client evidence hard-fails, while diagnostics always upload and the existing post-stage cleanup remains the fallback.

## Layout

Helpers moved out of `.github/scripts` into the `tests.utils` package, so the stage imports them instead of patching `sys.path`, and the workflow invokes the runtime as `python -m tests.utils.tts_mps_runtime`. `tests/utils.py` became `tests/utils/__init__.py`, leaving every existing `from tests.utils import ...` caller unchanged. The stage itself moved to `tests/test_ci/test_tts_mps_dp2.py`, since `tests/test_model` holds model quality suites and this validates placement.

## Performance references

References are raw worst-of-five observations held pre-slack; CI slack is derived at assert time, so a recalibration only rewrites the `_REF` literals. An uncalibrated reference is `None`, and the stage then reports the observation and fails, so the gate cannot pass ungated.

Calibrated with `.claude/skills/tune-ci-thresholds` on one H100 of the CI host
against **SGLang 0.5.16** (values measured on 0.5.12.post1 before the #1183
bump were discarded, not carried over). Five strict-valid repeats per stage at
concurrency 16 over the full SeedTTS EN set (1088 requests, `ok == total`
every repeat; similarity scores its fixed 50-sample subset). Run directory
`20260802T033000Z_tts-mps_all_r8`, calibration commit `e38b8de1`.
`strict-audit` reports `STRICT + GIT: READY` for all four stages, and
re-running `apply-plan` after the write reports `direction=equal` for all ten
metrics, so no rounding tightened a reference past its observed worst.

| Preset | Metric | Per-run observations | Worst-of-5 written | Spread |
|---|---|---|---:|---:|
| Higgs | throughput_qps | 14.864 / 14.730 / 14.615 / 14.468 / 14.789 | 14.468 | 2.7% |
| Higgs | output_tok_per_req_s | 117.3 / 116.6 / 116.5 / 115.4 / 117.0 | 115.4 | 1.6% |
| Higgs | latency_mean_s | 1.071 / 1.079 / 1.087 / 1.099 / 1.074 | 1.099 | 2.6% |
| Higgs | rtf_mean | 0.2549 / 0.2576 / 0.2591 / 0.2625 / 0.2569 | 0.2625 | 3.0% |
| MOSS | throughput_qps | 11.732 / 11.506 / 11.380 / 11.583 / 11.486 | 11.380 | 3.1% |
| MOSS | output_tok_per_req_s | 61.9 / 61.8 / 61.6 / 63.1 / 61.7 | 61.6 | 2.4% |
| MOSS | latency_mean_s | 1.358 / 1.384 / 1.398 / 1.375 / 1.385 | 1.398 | 2.9% |
| MOSS | rtf_mean | 0.3182 / 0.3238 / 0.3269 / 0.3225 / 0.3247 | 0.3269 | 2.7% |

The MOSS bimodality observed on 0.5.12 (runs clustering at 9.5 and 7.1, 28 to
38% spread) is gone on 0.5.16: all five MOSS runs sit within 3.1% and absolute
throughput rose from 7.1-9.8 to 11.4-11.7 req/s, so the MOSS floor tightens
substantially instead of inheriting a contaminated spread.

**Speaker similarity gets an MPS-calibrated baseline instead of reusing the
canonical constant.** The canonical similarity reference is calibrated under
ordinary DP2 on two cards; under a shared card the observed Higgs distribution
straddles it (five earlier runs spanned 65.08 to 67.37 against a 66.06 line),
which makes the gate a coin flip without any measured MPS quality penalty. The
new baselines are worst-of-5 from the same run: Higgs 65.42 (per-run 67.33 /
68.03 / 65.42 / 66.21 / 68.11), MOSS 62.59 (per-run 64.24 / 65.10 / 63.42 /
62.59 / 64.94). WER and audio-integrity gates are reused unchanged from the
canonical stages.

Assertion slack stays in the policy module (`x0.75` for lower-bound metrics,
`x1.25` for upper-bound), matching the existing TTS stages, so the stored
values remain raw pre-slack observations.

## Why MPS without CUDA IPC?

MPS provides concurrent scheduling; CUDA IPC is a separate weight-sharing optimization. Both DP2 configurations fit on one 80 GB H100 with independent weights, so IPC is not required for this validation stage. Keeping `WEIGHT_SHARE=0` avoids leader/follower lifetime coupling, shared-payload eligibility rules, and IPC recovery state. The trade-off is two resident weight copies and less memory headroom. IPC should be reconsidered only if a future model/config no longer fits or larger KV becomes a requirement.

## Bugs this surfaced

Calibrating the stage required running it outside its dedicated CI job, which exposed three defects that CI had been masking:

* The MPS control socket path was never bounded. Beyond the 107-byte AF_UNIX `sun_path` limit the daemon starts, fails to bind, and exits, leaving only `Cannot find MPS control daemon process`. Any state root deeper than roughly forty characters reproduces it; CI worked because the job supplies a short root. `MpsLaunchSpec` now validates the resolved path and names the cause.
* The stage resolved its GPU through `nvidia-smi`, which enumerates physically and ignores `CUDA_VISIBLE_DEVICES`, so with `TTS_MPS_GPU_ID` unset it bound physical GPU 0 regardless of how the surrounding harness had scoped it.
* `calibration_presets[*].constant_filter` in `tune.py` was applied only on the `variants` path, so every preset of a variant-less test claimed the first preset's symbols. Left unfixed, one preset's worst-of-five would have been written over another's.

## Cost, scope, and follow-ups

The canonical DAG still uses two H100s exactly as before. Stage 5 exercises one H100 for about five to six minutes, but the current self-hosted runner still reserves its normal GPU pair. This adds coverage; it does not yet create schedulable fleet capacity.

This PR changes only TTS validation. It does not migrate ASR, enable CUDA IPC, replace canonical stages, add a routed-c96 capacity gate, or implement runner-side quarantine.


JiaxinD:
## Superseded-SHA parallel adversarial review and validation history

This comment is retained because it records the review findings that drove the lifecycle fixes now present in the current head. Its SHA, metrics, and readiness disposition are historical; the PR description contains the authoritative exact-head results.

<details>
<summary>Review findings, fixes, and retained failures</summary>

### Review scope and disposition

Four independent, read-only Claude Code reviewers inspected `origin/main...d6431df3` before the last H100 matrix. They did not read one another's reports. Their raw verdicts were:

| Area | Raw verdict | Codex disposition |
|---|---|---|
| Architecture and scope | PASS WITH FINDINGS | Accepted the shared-clock corruption and stale-state wedge; both were fixed |
| Workflow and lifecycle | BLOCK | Accepted the activity-reader/signature and profiler flush race, stale-state recovery, and clock-domain findings; fixed with regression tests |
| MPS and model correctness | BLOCK | Accepted the same lifecycle/clock findings; rejected requests to treat model-path overlap as an SM-capacity benchmark |
| Tests and failure boundaries | PASS WITH FINDINGS | Confirmed fail-closed evidence; retained concrete compatibility observations as non-blocking follow-up |

Why these fixes were required: a monotonic value in the existing wall-clock field corrupted unrelated profiler views; an interrupted persistent launcher state could block every later PR run; and fire-and-forget profiler control could make a healthy overlap canary read an empty or truncated stream. The final implementation keeps wall-clock events intact, stores monotonic time in dedicated metadata, reconciles stale launcher state while preserving diagnostics, waits for recorder creation and terminal events, and tolerates only a partial final JSONL record.

Codex independently reproduced each accepted trigger before accepting it. Findings about full runner quarantine, fleet scheduling, capacity benchmarks, or restoring CUDA IPC were deferred because this PR is a standalone TTS validation stage and none is required for its one-H100 independent-weight contract. The timeout concern was closed by measured Stage 5 totals of 316.6 s for Higgs and 367.6/334.8 s for MOSS against a 25-minute job budget. The final fix commit is the single PR commit at `aa524ce70b0ab7c642197f510f8b767188e84e04`.

### Exact-SHA H100 evidence

| Model | Job / artifact | Result and reason |
|---|---|---|
| Higgs | [job 90268553964](https://github.com/sgl-project/sglang-omni/actions/runs/30349138618/job/90268553964), [artifact 8687543823](https://github.com/sgl-project/sglang-omni/actions/runs/30349138618/artifacts/8687543823) | PASS: 1088/1088, Router 558/540, SIM 67.2291, WER 0.00888, UTMOS 4.1732, four overlap matches, all floors, evaluator, and final cleanup pass |
| MOSS attempt 1 | [job 90268716312](https://github.com/sgl-project/sglang-omni/actions/runs/30349136458/job/90268716312), [artifact 8687575346](https://github.com/sgl-project/sglang-omni/actions/runs/30349136458/artifacts/8687575346) | FAIL: SIM 63.4985 < 64.1168; dual traffic 548/550, overlap, all performance floors, evaluator, and cleanup pass |
| MOSS diagnostic | [job 90271420922](https://github.com/sgl-project/sglang-omni/actions/runs/30349136458/job/90271420922), [artifact 8687879592](https://github.com/sgl-project/sglang-omni/actions/runs/30349136458/artifacts/8687879592) | FAIL: SIM 62.9261 < 64.1168; dual traffic 552/546, overlap, all performance floors, evaluator, and cleanup pass |

Why MOSS remains blocked: the same exact-SHA canonical Stage 1 scored 64.4527 and older paired evidence showed floor-adjacent unseeded variance, but two final-SHA MPS failures cannot be converted into a pass. The original failure is retained, only one diagnostic Stage 5 rerun was performed, and no similarity threshold was changed.

Both exact-SHA workflows passed canonical TTS Stage 1, Stage 2, and Stage 3. Both fixed-Higgs Stage 4 jobs failed all three internal attempts on the unchanged serving-performance contract, dominated by batch32 p95 around 9.5-10.1 s versus 1.9825 s. Every retry and job-final cleanup returned both GPUs to 0 MiB. This is attributed to the existing serving floor/load path rather than Stage 5, but it remains red evidence rather than being hidden.

### Local and repository checks

- Relevant local tests: `46 passed, 1 skipped`.
- `pre-commit run --all-files`: pass.
- `git diff --check`: pass.
- Exact-SHA unit jobs: pass in both workflows.
- The final diff is one commit, 12 files, `+2590/-52`; the worktree was clean when the matrix was dispatched.

The parent workflows were cancelled only after all required TTS jobs and artifacts completed, because unrelated downstream Qwen3 jobs are outside this TTS-only review.

</details>

JiaxinD:
<details>
<summary><strong>Paired MOSS similarity attribution, Stage 4 merge-base baseline, and focused review disposition</strong></summary>

### Why this diagnostic was run

The exact PR SHA retained two MOSS Stage 5 similarity failures even though dual traffic, overlap, performance, evaluator, and cleanup passed. A code change is justified only if the failure is specific to MPS, so the diagnostic compared ordinary DP2 and MPS DP2 at one temporary SHA with the same public generation seed (`1186`). The temporary workflow changed only experiment control and is not part of this PR.

### Three fixed-seed pairs

The ordinary value below is the first attempt, selected before observing any retry. All failures are retained.

| Pair | Workflow | Ordinary DP2 SIM | MPS DP2 SIM | MPS - ordinary | MPS artifact |
|---|---|---:|---:|---:|---|
| A | [30411122770](https://github.com/sgl-project/sglang-omni/actions/runs/30411122770) | 63.9816 | 63.8095 | -0.1721 | [`8709446913`](https://github.com/sgl-project/sglang-omni/actions/runs/30411122770/artifacts/8709446913) |
| B | [30411124558](https://github.com/sgl-project/sglang-omni/actions/runs/30411124558) | 63.3479 | 63.7993 | +0.4514 | [`8709018286`](https://github.com/sgl-project/sglang-omni/actions/runs/30411124558/artifacts/8709018286) |
| C | [30411126033](https://github.com/sgl-project/sglang-omni/actions/runs/30411126033) | 63.8283 | 63.4694 | -0.3589 | [`8709156660`](https://github.com/sgl-project/sglang-omni/actions/runs/30411126033/artifacts/8709156660) |

Against the unchanged floor of 64.1168:

- ordinary first attempts: mean 63.7193, sample standard deviation 0.3306, pass rate 0/3;
- MPS attempts: mean 63.6927, sample standard deviation 0.1935, pass rate 0/3;
- paired delta: mean -0.0266, sample standard deviation 0.4243, range -0.3589 to +0.4514.

The fixed seed did not make generation deterministic. Ordinary retries on the same runner and topology still changed materially: pair B moved from 63.3479 to 64.1190, pair C from 63.8283 to 64.6146, and pair A produced 63.9816, 63.7995, then 64.2711.

GitHub scheduled ordinary and MPS jobs independently, so they were not pinned to the same physical runner; this remains a hardware confound and prevents claiming exact equivalence. The supported conclusion is narrower: ordinary DP2 fails the same floor, paired differences reverse sign, and their mean is near zero, so the evidence does not show a stable MPS-only quality regression.

Every MPS diagnostic completed 1088 canonical requests, two-worker traffic, four overlap matches, WER, UTMOS, all MPS performance floors, Qwen3-ASR evaluation, and clean final teardown. Only similarity failed.

### Stage 4 exact merge-base baseline

The exact PR-head Stage 4 jobs failed the unchanged batch32 p95 floor on all attempts. To test whether the PR introduced the red stage, one direct sample was run from exact PR merge base `a689c2c82a49a55207e8bb37919c7c55260ef0dd`, with only workflow isolation:

- [workflow 30413591393](https://github.com/sgl-project/sglang-omni/actions/runs/30413591393)
- [Stage 4 job 90456089822](https://github.com/sgl-project/sglang-omni/actions/runs/30413591393/job/90456089822)
- Result: failure before serving assertions because CUDA-graph capture exceeded the existing 180-second service-readiness window.
- Cleanup: success; final GPU memory was 0 MiB.

This is not an identical reproduction of the PR-head p95 failure, so it is not presented as the same root cause. It does establish that the exact merge-base Stage 4 is independently unstable. In the PR DAG Stage 5 starts only after Stage 4 completes, so Stage 5 cannot causally affect that earlier job.

### Focused adversarial review and Codex disposition

Four independent read-only Claude Code reviews examined seed control, Stage 4 causality, diagnostic validity, and potential remedies. Codex checked each concrete claim against code, logs, and artifacts.

Accepted observations:

- the diagnostic seed reaches real generation, but fixed seeding does not remove scheduling/numerical nondeterminism;
- independent runner placement is a real limitation;
- ordinary Stage 1 has a three-attempt wrapper while Stage 5 has one attempt, so only the preselected first ordinary attempt is used in paired statistics;
- the MOSS similarity threshold has effectively no observed slack and was not introduced by this PR;
- completed Stage 5 artifacts show clean teardown, weakening the theoretical possibility of cross-run MPS residue.

Rejected as PR fixes:

- adding Stage 5 retries, because it would select a lucky sample and hide the observed failure;
- changing the seed in production CI, because the same fixed seed still varied and would change the canonical quality contract without resolving the cause;
- lowering similarity, because threshold calibration is separate work and there is no reviewed evidence for a new floor;
- changing Stage 4 from this PR, because the stage is unchanged, runs before Stage 5, and is already unstable at merge base;
- requiring full runner quarantine or same-runner pinning, because those are runner-infrastructure controls outside this TTS validation change.

Disposition: no PR code change is supported by the evidence. The implementation is appropriate for human review as a Draft; whether existing MOSS similarity and Stage 4 CI instability are acceptable is an explicit maintainer/release-policy decision, not something to conceal with retries.

</details>

.github/scripts/tts_ci_selection.py
zhaochenyang20, line 1: could this reuse exisiting CI logits of `run-moss` and `run-higgs`? Why add a new file for selection?

zhaochenyang20 (commented):
Also, I do think that we should have a `benchmark/eval/tts_mps_perf.py`, and then the test file import utils from that `eval/`. Follow our existing principles.

.github/scripts/tts_ci_selection.py
JiaxinD: Fixed in `861ed8a8`. CPU preflight now consumes the existing `run-higgs` / `run-moss` labels and workflow override directly; only the no-label case adds the rerun-stable SHA-256 digest of `github.run_id`. Keeping the decision at this existing gate preserves conflict and config validation before any H100 job without a standalone selection module.

Hayden727:
Review result for head `07ee84d41ba504ebcdfc07a2ab2ab9fb678e86ed`: changes requested.

### 1. Preserve the PCI domain when resolving the GPU NUMA node

`derive_core_blocks()` drops the PCI domain and globs only by `bus:device.function`. On `hyper00-omni`, GPU 0 reports `00000000:08:00.0`, while sysfs contains both `0000:08:00.0` and `0001:08:00.0`. The unmodified PR therefore fails before launching either replica with:

```text
RuntimeError: cannot resolve one PCI device for GPU 0
```

Please normalize the full domain/bus/device address and resolve the exact sysfs path. A regression test should cover identical bus suffixes in two PCI domains.

### 2. Require all eight measured canary requests to succeed

`canary_results` is never checked. `read_model_path_activity()` counts all terminal events, including error/aborted requests, while the overlap oracle only requires two successes per replica. A synthetic trace with four successes and four errors currently passes. In addition, the benchmark's default `warmup=1` is inside the profiling window, so an “8 request” canary actually records nine model-path intervals; the H200 diagnostic run observed 5 + 4 intervals.

Please run this canary with warmup disabled, assert `total_requests == completed_requests == 8` and `failed_requests == 0`, and preserve those exact counts in the evidence.

### 3. Preserve the private MPS daemon log on failure

`_copy_raw_state()` copies `mps_ctl.err`, launcher logs, and activity files, but not `state/mps/log`, even though the launcher configures `CUDA_MPS_LOG_DIRECTORY` there. During a transient MPS daemon startup failure, the retained artifact only contained `Cannot find MPS control daemon process`; the daemon log that could explain the root cause was removed with the state directory.

Please copy `state/mps/log` before teardown using the same symlink protections, and add a failure-path test.

### Validation notes

With a diagnostic-only PCI resolution patch, the full flow ran on the shared H200 host: 2 replicas, 1088/1088 canonical requests, MPS attachment and overlap pass, QPS 16.804, mean latency 0.946 s, RTF 0.2299, WER 0.00963, UTMOS 4.1777, and clean teardown with no new GPU clients. Speaker similarity was 66.3103 versus the existing 66.3962 threshold; because this was a shared H200 rather than the target H100 runner, that result should be rechecked in the official H100 job.

The current full unit-test failure is reproducible on both the PR base and head (`MossTTSPipelineConfig` topology test), so I did not attribute that baseline failure to this PR.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -1011,6 +1017,7 @@ def _emit_prefill_start_for_batch(self, batch: ScheduleBatch) -> None:
             if rid in self._prefill_start_done:
                 continue
             self._prefill_start_done.add(rid)
+            _emit_mps_model_path_start(rid)
```
Hayden727: `OmniScheduler` also serves ordinary non-MPS requests, so this unconditional call labels any profiled request as an MPS model-path execution. Please keep the scheduler event execution-mode-neutral (for example, `model_path_start/end`) and emit or derive the MPS-specific evidence in the MPS runtime after daemon/client attachment is verified.

Hayden727: Fixed in e7d63580. Shared recorder/scheduler telemetry now uses execution-mode-neutral `model_path_start` / `model_path_end`, while the TTS MPS runtime derives its evidence only after launcher attachment validation confirms private MPS workers. The scheduler now pairs terminal events across success, error, abort, and shutdown, and Stage waits for scheduler-thread terminalization before teardown. Verified with the affected Stage, scheduler, recorder, and MPS-runtime suites: 100 passed, 3 skipped.

tests/utils/tts_mps_overlap.py
zhaochenyang20, line 1: why not move it under `tests/utils`?

tests/utils/tts_mps_runtime.py
zhaochenyang20, line 1: why not move it under tests/utils?

tests/test_model/test_tts_mps_ci.py
zhaochenyang20, line 1: I think we do not need to put this under `tests/test_model`. a better naming needed

zhaochenyang20: Also, this file is heavy. Add file comments for your idea of testing and add yourself as the author.

zhaochenyang20 (changes requested):

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -221,6 +223,21 @@ def _record_cleanup_error(component: str, exc: Exception) -> None:
...
+                        _SCHEDULER_THREAD_JOIN_TIMEOUT_S,
+                    )
+                    if scheduler_thread.is_alive():
+                        raise TimeoutError(
+                            "scheduler thread did not stop within "
+                            f"{_SCHEDULER_THREAD_JOIN_TIMEOUT_S:g}s"
+                        )
+                    self._scheduler_thread = None
+                except Exception as exc:
+                    _record_cleanup_error("scheduler thread", exc)
```
zhaochenyang20, line 240: `_scheduler_thread` is created with `daemon=True` at `runtime.py:190-195`, meaning that before this change `Stage.stop()` never joined it — the interpreter reclaimed it at process exit and the shutdown path could not fail because of this thread. The new block joins for 5 seconds and raises `TimeoutError` on expiry; the first error recorded by `_record_cleanup_error` is then re-raised at `runtime.py:255-256` as `RuntimeError(f"Stage {self.name} cleanup failed")`. So a production shutdown path that previously always succeeded now has a failing branch, and both `stage_workers.py:451` and `:487` will see it. The constant 5.0 carries no comment explaining where it came from; `_event_loop_normal` does check `while self._running` every iteration, but if the scheduler thread is stuck inside a large-batch forward or a blocking CUDA call, 5 seconds may not be enough. More importantly, the real motivation for this block — letting `_emit_remaining_model_path_ends` in `OmniScheduler.start()`'s `finally` complete before the profiler files are read — is nowhere in the code, so nobody reading it later can tell whether the 5 seconds is removable. This is a production behavior change shipped inside a PR whose description says "This PR changes only TTS validation".

zhaochenyang20: Three things: comment where the constant comes from and why the join exists; downgrade the timeout from `raise` to `logger.warning` plus an evidence entry in the summary (if CI wants strictness, assert it on the CI side rather than failing the production shutdown path); and state explicitly in the PR description that this PR touches three production files. If you insist on raising, justify the invariant "the scheduler thread always exits within 5s" and write that invariant into the comment.

```python
# Note (Jiaxin Deng): the scheduler thread emits terminal model-path events
# in its own `finally`; the standalone MPS stage reads those files right
# after Stage.stop() returns, so we wait for the thread instead of letting
# the daemon thread be reclaimed at process exit. A slow thread is logged,
# not fatal — shutdown correctness must not depend on this timeout.
scheduler_thread = self._scheduler_thread
if scheduler_thread is not None:
    await asyncio.to_thread(
        scheduler_thread.join, _SCHEDULER_THREAD_JOIN_TIMEOUT_S
    )
    if scheduler_thread.is_alive():
        logger.warning(
            "Stage %s scheduler thread did not stop within %gs",
            self.name,
            _SCHEDULER_THREAD_JOIN_TIMEOUT_S,
        )
    else:
        self._scheduler_thread = None
```

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -1192,8 +1213,8 @@ def abort(self, request_id: str, *, defer_running_cleanup: bool = True) -> None:
         self._deferred_request_payloads.pop(request_id, None)
         self._dirty_deferred_request_ids.discard(request_id)
         self._first_emit_done.discard(request_id)
-        self._prefill_start_done.discard(request_id)
         if not running_abort:
+            self._emit_model_path_end_once(request_id, status="aborted")
             self._release_immediate_request_resources(request_id)
```
zhaochenyang20, line 1578: Before this change, `self._prefill_start_done.discard(request_id)` ran unconditionally outside `if not running_abort:`; now it is replaced by `_emit_model_path_end_once(...)` and moved inside that branch. When `_mark_running_request_aborted()` returns True (the request is still in a running batch), the rid stays in `_prefill_start_done` and is never removed there. On the normal path the request gets `FINISH_ABORT`, reaches `stream_output`'s abort branch at `:1066` on the next step, and is cleared — so most of the time this is fine. But any path where a request disappears without passing through `stream_output` — a retraction when the KV pool fills, or a later `defer_running_cleanup=False` abort that drops it via `_remove_from_batch` — leaves that rid in the set forever. `_prefill_start_done` has no capacity bound, whereas `_aborted_request_ids` in the same file does carry `_ABORTED_REQUEST_ID_LIMIT` / `_ABORTED_REQUEST_ID_RETAINED` eviction, which shows the maintainers already consider this class of set to grow on a long-running server. Beyond memory there is a semantic side effect: `_prefill_start_done` also serves as the dedup key for `prefill_start`, so a leftover rid silently swallows a later `prefill_start` for the same id.

zhaochenyang20: Split model-path lifetime tracking out of `_prefill_start_done` into its own set, so "have we emitted prefill_start" and "is the model-path interval still open" stop constraining each other; and restore the unconditional cleanup of `_prefill_start_done` in the running-abort branch. Add a unit test covering the path where a running-aborted request never reaches `stream_output`.

```python
self._first_emit_done.discard(request_id)
self._prefill_start_done.discard(request_id)
self._emit_model_path_end_once(request_id, status="aborted")
if not running_abort:
    self._release_immediate_request_resources(request_id)
    ...
```

.github/workflows/test-tts-ci.yaml
```diff
@@ -249,3 +258,106 @@ jobs:
...
+
+      - name: Post-stage cleanup fallback
+        if: always()
+        uses: ./.github/actions/omni-post-stage
+        with:
+          stage-label: TTS MPS non-streaming validation
+          artifact-search-root: ${{ env.TTS_MPS_OUTPUT_ROOT }}
+          summary-only: "true"
+
+      - name: Upload TTS MPS evidence
```
zhaochenyang20, line 356: Are you sure this upload artifacts is the same as other stage? And these artifacts should be removed after the CI running.

sglang_omni/profiler/event_recorder.py
```diff
@@ -274,3 +274,41 @@ def emit(
...
         timestamp_ns=timestamp_ns,
     )
+
+
+def _read_host_boot_id() -> str | None:
+    try:
+        return Path("/proc/sys/kernel/random/boot_id").read_text().strip() or None
+    except OSError:
+        return None
+
```
zhaochenyang20, line 289: The boot id is constant for the process lifetime, yet `emit_model_path_start` and `emit_model_path_end` each call `_read_host_boot_id()`, i.e. two `open` + `read` + `close` syscalls per request, on the scheduler event-loop thread. The comment on that loop (`omni_scheduler.py:1699-1704`) already spells out that Python-side work on this thread lands directly on the co-located encoder/preprocessor, with a historical 600x lesson. The `is_active()` guard means this only happens during profiling — but profiling is exactly when we are measuring time, and adding avoidable syscalls to the hot path contaminates the measured object, since the overlap oracle measures precisely the intervals between `monotonic_ns` stamps.

**Fix:**

```python
@functools.cache
def _read_host_boot_id() -> str | None:
    try:
        return Path("/proc/sys/kernel/random/boot_id").read_text().strip() or None
    except OSError:
        return None
```

benchmarks/eval/tts_mps_perf.py
```diff
@@ -0,0 +1,96 @@
...
+            "output_tok_per_req_s": 40.0,
+        },
+        "maximum": {"latency_mean_s": 2.4, "rtf_mean": 0.60},
+        "provenance": {
+            "temporary": True,
+            "method": "wide_floor_with_over_30_percent_slack_from_h100_sample",
+            "accepted_run_ids": [30202304743],
+        },
+    },
+}
```
zhaochenyang20: The repo already has an explicit methodology for TTS thresholds: `tests/test_model/tts_ci_config.py:56-78` stores P95 observations, `THRESHOLD_SLACK_HIGHER = 0.75` / `THRESHOLD_SLACK_LOWER = 1.25` define the widening factors, `apply_slack` at `tests/utils.py:189` derives the thresholds, and `tune-ci-thresholds` can recalibrate the whole set. The new `MPS_PERFORMANCE_FLOORS` uses none of it: hardcoded round numbers like `10.0 / 43.0 / 1100.0 / 80.0 / 1.8 / 0.45`, with the observations surviving only as free text in `provenance.accepted_run_ids`, and `temporary: True` as a comment rather than an expiry mechanism. The consequence is twofold — the floors cannot be recalibrated by the existing tooling, and they are too wide to catch anything: Hayden727's measured Higgs QPS was 16.804 against a floor of 10.0, so a 40% drop still passes, versus canonical's `18.189 * 0.75 = 13.6`. The MPS line is nearly twice as loose.

Express the MPS floors through the same derivation so `tune-ci-thresholds` can manage them alongside everything else, and let `check_mps_performance` keep only the adjudication logic instead of carrying its own constant table.

```python
from tests.utils import apply_slack
from tests.test_model.tts_ci_config import (
    THRESHOLD_SLACK_HIGHER,
    THRESHOLD_SLACK_LOWER,
)

# Observed on one H100 under private-MPS DP2, runs 30196509700 / 30199556014.
_HIGGS_MPS_P95 = {
    16: {
        "throughput_qps": 16.80,
        "audio_throughput_s_per_s": 57.4,
        "output_throughput": 1_465.0,
        "output_tok_per_req_s": 106.0,
        "latency_mean_s": 0.946,
        "rtf_mean": 0.2299,
    }
}
HIGGS_MPS_THRESHOLDS = apply_slack(
    _HIGGS_MPS_P95, THRESHOLD_SLACK_HIGHER, THRESHOLD_SLACK_LOWER
)
```

.github/scripts/tts_mps_overlap.py
```diff
@@ -0,0 +1,134 @@
...
+    events: Iterable[dict[str, Any]],
+    *,
+    expected_run_id: str,
+    min_successes_per_replica: int,
+    min_matched_overlap_count: int,
+    measurement_uncertainty_ns: int,
+) -> dict[str, Any]:
+    if min_successes_per_replica < 2:
+        raise ValueError("overlap proof requires at least two repeated intervals")
+    if min_matched_overlap_count != 2:
```
zhaochenyang20: `min_matched_overlap_count` accepts only 2 (`:76-77` raises for anything else) and the pairing logic `combinations(candidates, 2)` at `:102` hardcodes 2 as well, so `matched_count = len(pairs)` is always 2 and the returned `"matched_overlap_count"` is a fixed constant. A parameter with exactly one legal value is not a parameter but a constant disguised as configuration; a reader will assume raising it demands stronger concurrency evidence, and it does not. The same problem shows in the error strings: `"required at least two"` at `:90` and `"required two one-to-one matches"` at `:112` both bake the number into the message and ignore the `min_successes_per_replica` that was actually passed (that one is genuinely used at `:87`). Incidentally, the "4 overlap matches" in the PR description is not producible by this code, which confirms the evidence table has drifted from the head.

Either genuinely support arbitrary N, or drop the parameter and promote 2 to a named module constant. I lean toward the latter — this is a bounded canary and 2 is sufficient; it does not need to be configurable.

```python
REQUIRED_ONE_TO_ONE_MATCHES = 2

def build_overlap_verdict(
    events, *, expected_run_id, min_successes_per_replica, measurement_uncertainty_ns
):
    ...
    if min(counts.values()) < min_successes_per_replica:
        raise ValueError(
            f"insufficient successful requests per replica: {counts}; "
            f"required at least {min_successes_per_replica}"
        )
    pairs = next(
        (
            pair
            for pair in combinations(candidates, REQUIRED_ONE_TO_ONE_MATCHES)
            if pair[0][0] != pair[1][0] and pair[0][1] != pair[1][1]
        ),
        None,
    )
```

tests/test_model/test_tts_mps_ci.py
```diff
@@ -0,0 +1,423 @@
...
+    with path.open("w", encoding="utf-8", newline="\n") as stream:
+        for event in events:
+            stream.write(json.dumps(event, sort_keys=True))
+            stream.write("\n")
+
+
+def test_tts_mps_non_streaming(
+    tmp_path_factory: pytest.TempPathFactory,
+    selected_tts_concurrencies: tuple[int, ...],
+) -> None:
```
zhaochenyang20: A single test function runs from `:145` to `:423`, far past the repo's 50-line function ceiling. Inside it there is a hand-rolled `cleanup_recorded` boolean state machine (`:163`, `:281`, `:314`, `:321`, `:325`), a nested `except BaseException` with an inner `try` and an inner `if not cleanup_recorded`, and six variables initialised to `None` whose assignment is guaranteed only by control flow (`before_workers`, `after_workers`, `speed_results`, `overlap`, plus `visible_mps_clients` and `performance`, which are assigned only inside the `try` yet read at `:291` and `:307`). The same teardown logic appears three times: the happy path (`:276`), the exception path (`:316`), and `tts_mps_runtime.finalize_from_cli` (`:707-734`) — and with the workflow's `if: always()` finalize step and the `omni-post-stage` fallback, the same concern is handled at five layers. Under that structure the next person to touch it will miss one. Separately, `timing={**json.loads(summary_path.read_text(...))["timing"], ...}` at `:413-416` reads the file inside a kwargs expression and then `update_summary` reads the same file again, which shows that `update_summary`'s whole-section-replacement semantics are mismatched with the caller's incremental-update need.

Wrap replica lifetime in a contextmanager so teardown is written once; change `update_summary` to merge semantics (`payload[key] = {**payload.get(key) or {}, **value}`) to eliminate both the four full rewrites of the `runtime` section and the read-inside-kwargs pattern; then split canonical generation, the canary overlap, and quality evaluation into three ~30-line helpers.

```python
@contextmanager
def _mps_replicas(spec: MpsLaunchSpec, summary_path: Path, baseline: list[dict]):
    snapshot = launch_replicas(spec)
    try:
        yield snapshot
    finally:
        cleanup = teardown_replicas(spec, snapshot, baseline_gpu_clients=baseline)
        update_summary(summary_path, cleanup=cleanup)
        require_clean_cleanup(cleanup)
```

tests/test_model/test_tts_mps_ci.py
```diff
@@ -0,0 +1,423 @@
...
+
+PROJECT_ROOT = Path(__file__).resolve().parents[2]
+sys.path.insert(0, str(PROJECT_ROOT / ".github" / "scripts"))
+
+from tts_mps_overlap import build_overlap_verdict  # noqa: E402
+from tts_mps_runtime import (  # noqa: E402
+    MpsLaunchSpec,
+    atomic_write_json,
+    capture_gpu_clients,
+    derive_core_blocks,
```
zhaochenyang20: regarding this function:

`usable = allowed[: max(REPLICA_COUNT, len(allowed) * 3 // 4)]` reserves 25% of cores for the host *before* checking `len(usable) < 4`. So when `allowed` is exactly 4 or 5, `4*3//4 = 3` and `5*3//4 = 3`, leaving `usable` at 3 and raising "MPS DP2 requires four allowed CPU cores on the GPU NUMA node" — even though the machine does have 4-5 usable cores and the stated condition holds. The message reports a self-inflicted truncation as an insufficient-host condition, sending whoever debugs it off to inspect NUMA and affinity for a problem that is not there. Also, `REPLICA_COUNT` (= 2) as the lower bound inside `max(...)` is meaningless, since the real lower bound is 4.

**Fix:**

```python
MINIMUM_CORES = 2 * REPLICA_COUNT
HOST_RESERVE_RATIO = 0.25

allowed = sorted(node_cpus & set(os.sched_getaffinity(0)))
if len(allowed) < MINIMUM_CORES:
    raise RuntimeError(
        f"MPS DP2 requires {MINIMUM_CORES} allowed CPU cores on the GPU "
        f"NUMA node, found {len(allowed)}"
    )
# Reserve a slice for host-side work, but never below the hard minimum.
usable = allowed[: max(MINIMUM_CORES, int(len(allowed) * (1 - HOST_RESERVE_RATIO)))]
```

sglang_omni/profiler/event_recorder.py
```diff
@@ -274,3 +274,41 @@ def emit(
...
+        request_id=request_id,
+        stage=None,
+        event_name="model_path_end",
+        metadata={
+            "clock": "CLOCK_MONOTONIC",
+            "host_boot_id": _read_host_boot_id(),
+            "monotonic_ns": time.monotonic_ns(),
+            "status": status,
+        },
+    )
```
zhaochenyang20: these two functions could be merged

zhaochenyang20 (changes requested):

tests/utils/tts_mps_runtime.py
JiaxinD: Done, now `tests/utils/tts_mps_runtime.py`; the workflow calls it as `python -m tests.utils.tts_mps_runtime`.

tests/test_model/test_tts_mps_ci.py
JiaxinD: Done, moved to `tests/test_ci/test_tts_mps_dp2.py`.

JiaxinD: Done, file comment covers the three oracles and why each exists, author added.

.github/workflows/test-tts-ci.yaml
JiaxinD: Done. Fixed name plus `overwrite: true` and `if-no-files-found: error` like the other stages, evidence paths only (it was uploading all 1088 wavs), and a step that removes the run outputs.

sglang_omni/profiler/event_recorder.py
JiaxinD: Done, `functools.cache`.

sglang_omni/profiler/event_recorder.py
JiaxinD: Done, both now call one `_emit_model_path`.

.github/scripts/tts_mps_overlap.py
JiaxinD: Done, parameter dropped for module constant `REQUIRED_ONE_TO_ONE_MATCHES`; messages now use the arguments actually passed.

tests/test_model/test_tts_mps_ci.py
JiaxinD: Done, your version. Added a parameterised 4/5-core test that failed before.

sglang_omni/pipeline/stage/runtime.py
JiaxinD: Done, downgraded to `logger.warning` with the reason written next to the constant. Existing test now asserts the warning instead of `pytest.raises`.

sglang_omni/scheduling/omni_scheduler.py
JiaxinD: Fixed the leak, but kept one set: emit `model_path_end` first, then discard unconditionally. Splitting adds an attribute that 25 `object.__new__` test fixtures do not set (18 failures); say the word and I will split plus update them.

tests/test_model/test_tts_mps_ci.py
JiaxinD: Done. Replica lifetime is a contextmanager so teardown is written once, `update_summary` merges, and the body split into generation / canary / quality. Test function 280 to 63 lines.

benchmarks/eval/tts_mps_perf.py
JiaxinD: Done, worst-of-5 via `tune-ci-thresholds` and registered under `models/tts`. Kept slack constants in the eval module rather than importing `apply_slack`, so `benchmarks/` does not depend on `tests/`; numbers and evidence are in the PR body.

tests/utils/tts_mps_overlap.py
JiaxinD: Done, now `tests/utils/tts_mps_overlap.py`.

benchmarks/eval/tts_mps_perf.py
JiaxinD: Recalibrated on 0.5.16 after the #1183 bump (run 20260802T033000Z_tts-mps_all_r8, 5/5 strict, direction=equal on re-check); worst-of-5 table in the PR body.

tests/test_ci/tts_mps_ci_config.py
```diff
@@ -0,0 +1,162 @@
...
+
+MPS_HIGGS_THROUGHPUT_QPS_REF: float | None = 14.468
+MPS_HIGGS_OUTPUT_TOK_PER_REQ_S_REF: float | None = 115.4
+MPS_HIGGS_LATENCY_MEAN_S_REF: float | None = 1.099
+MPS_HIGGS_RTF_MEAN_REF: float | None = 0.2625
+
+MPS_MOSS_THROUGHPUT_QPS_REF: float | None = 11.38
+MPS_MOSS_OUTPUT_TOK_PER_REQ_S_REF: float | None = 61.6
+MPS_MOSS_LATENCY_MEAN_S_REF: float | None = 1.398
+MPS_MOSS_RTF_MEAN_REF: float | None = 0.3269
```
zhaochenyang20, line 37: I am not so sure how other tests do for the reference performance. But generally, we put all the thresholds self-contained in `tests/`, and `benchmarks/eval/` purely put the eval script. Move all the values under `tests/` like other tests did.

zhaochenyang20 (approved):

tests/test_ci/tts_mps_ci_config.py
JiaxinD: Done, moved whole to tests/test_ci/tts_mps_ci_config.py; benchmarks/eval no longer carries thresholds.

Hayden727 (approved):

PR #1187, opened 2026-07-26
[Qwen3-Omni Perf] Cache talker thinker-embed shard resolution and rows

edwingao28:
## Motivation

`load_thinker_embedding_rows` in `sglang_omni/models/qwen3_omni/components/talker_prefill.py` runs once per streamed text chunk, on the talker scheduler thread, inside the receive path. Before this PR it re-resolved the embedding shard from scratch every time: glob the model directory, read the whole `*.safetensors.index.json`, `json.loads` it, then reopen the shard with `safe_open` to read a few rows. For a 30B checkpoint that is ~11.4 ms per chunk, and ~60% of it is parsing the same index file again and again.

At concurrency 8 the thinker delivers a wave's text in about 1.3 s (~1200 chunks), so the talker scheduler thread spends seconds parsing instead of stepping, and frame emission stalls. Profiling on H200 attributed 83–88% of all talker stall time (gaps longer than 200 ms) to this path, and 79–86% of the time-to-first-audio regression at concurrency 8 to those stalls.

## Modifications

All in `talker_prefill.py`, keyed by `model_path`:

- Cache the shard resolution (index parse → shard file + tensor name) in `_EMBED_SOURCE_CACHE`, so the index JSON is parsed once.
- Keep the `safe_open` handle in `_EMBED_HANDLE_CACHE` instead of reopening the shard per call.
- Removed `_read_rows_from_safetensor`, now unused.

An earlier revision also cached individual rows in a process-global `_EMBED_ROW_CACHE`. Review pointed out that repeated token ids never reach this function in production — `TalkerPrefillBuilder` already filters them through its own `_thinker_embed_cache` — so that cache was unreachable and has been dropped (1b6622bd). Row reads now always go through the cached handle.

The rows returned are byte-identical to the old path: same dtype, same values, the caller's `.to(device, dtype)` cast is unchanged. The function is only called from the talker scheduler thread.

## Related Issues

Found while working on the stall attribution for #1018 / #1022.

## Accuracy Test

Byte-identical at two levels:

- Unit parity against the real Qwen3-Omni-30B-A3B-Instruct snapshot: `bitwise_equal=True` for 6 batch shapes (n=1, 3, 5, 17, plus a batch with repeated ids).
- Server-level identity gate: deterministic recipe (`temperature 0`, `top_k 1`, `talker_temperature 0.0`, `talker_top_k 1`, `talker_top_p 1.0`), 4 prompts, run in all 6 benchmark server launches below. Audio sha256 identical in every arm:

```
p0 bytes=2885172 audio_sha=70b4964e76adde0b
p1 bytes=1014632 audio_sha=01199aefe739e26b
p2 bytes=1761312 audio_sha=d790863c5c9c9f19
p3 bytes=204804  audio_sha=06a0d32bb25f1908
```

Unit tests: `tests/unit_test/qwen3_omni/test_talker_prefill_embed_cache.py` (6 tests: row correctness, index parsed once, shard opened once, repeated ids return identical rows, no-index fallback, missing-weights error).

## Benchmark & Profiling

H200, Qwen3-Omni-30B-A3B-Instruct, thinker TP2 + talker + code2wav colocated on 2 GPUs. Six alternating server launches (base / cached ×3), each: warmup, deterministic identity check, then 16 requests at concurrency 8 with `max_tokens=200`. The cache was applied as an env-gated switch so both arms ran the same build.

These runs used the earlier revision that still carried the row cache. The numbers stand for the current code: the row cache was unreachable on the serving path (repeated ids are filtered one level up), so the measured path — resolve once, keep the handle open, read rows through it — is exactly what the PR now ships.

| arm | wall (s) | TTFA p50 (s) | TTFA p90 (s) | E2E p50 (s) | stalls >200ms | stall total (s) |
|---|---|---|---|---|---|---|
| base 1 | 33.9 | 5.09 | 7.33 | 14.29 | 12 | 9.33 |
| cached 1 | 27.5 | 1.81 | 2.72 | 12.19 | 2 | 1.06 |
| base 2 | 35.1 | 4.98 | 7.94 | 14.98 | 15 | 10.28 |
| cached 2 | 27.2 | 1.65 | 3.30 | 11.59 | 2 | 0.48 |
| base 3 | 36.1 | 5.73 | 8.33 | 15.25 | 12 | 11.25 |
| cached 3 | 25.6 | 1.78 | 2.77 | 11.03 | 2 | 0.92 |

Per-pair deltas:

| metric | pair 1 | pair 2 | pair 3 |
|---|---|---|---|
| time to first audio, p50 | −64.6% | −66.9% | −68.9% |
| stall time >200 ms | −88.6% | −95.3% | −91.8% |
| bench wall | −19.0% | −22.5% | −29.1% |
| end-to-end p50 | −14.7% | −22.6% | −27.6% |

py-spy on the talker scheduler thread confirms the mechanism: in the base arms, stall windows are 78–99% inside `append_text_chunk` with `json.loads` / `safe_open` leaves; in the cached arms those frames are gone and the long stall class disappears. Isolated per-call cost of `load_thinker_embedding_rows`: 11.4 ms → 0.02 ms.

## Checklist

- [x] Add unit tests.

## CI

if labeled.


sglang_omni/models/qwen3_omni/components/talker_prefill.py
```diff
@@ -28,41 +28,58 @@
...
-                    return _read_rows_from_safetensor(shard_path, tensor_name, row_ids)
+                    source = (shard_path, tensor_name)
+                    _EMBED_SOURCE_CACHE[model_path] = source
+                    return source
 
     raise KeyError(f"Unable to locate thinker embedding weights in {model_path}")
 
 
+def load_thinker_embedding_rows(model_path: str, row_ids: list[int]) -> torch.Tensor:
+    row_cache = _EMBED_ROW_CACHE.setdefault(model_path, {})
```
SandyLuXY: Repeat streamed tokens never reach this cache because the single production `TalkerPrefillBuilder` already filters them through `_thinker_embed_cache`. Consider removing the process-global row cache and keep only source/handle caching, or show a real cross-builder caller and give the additional cache a bounded invalidation lifecycle.

edwingao28: fixed in https://github.com/sgl-project/sglang-omni/pull/1187/commits/1b6622bd17f42eb14b0f54c7ed461f773e11cfde

SandyLuXY (approved):
LGTM. Non-blocking: `_EMBED_ROW_CACHE` was removed, while the PR title/body still claim rows are cached, please update the PR metadata.

edwingao28:
> LGTM. Non-blocking: `_EMBED_ROW_CACHE` was removed, while the PR title/body still claim rows are cached, please update the PR metadata.

Ty and fixed！

PR #1190, opened 2026-07-26
[ZONOS2] Release decode state at request termination

Ratish1:
## Motivation

ZONOS2 kept one decode-state row owned after the final request was terminated by the scheduler's length limit. The next decode batch happened to reclaim the stale row through `release_inactive()`, masking the leak during sustained traffic, but an idle server retained the last request indefinitely. Abort paths had the same missing model-state cleanup boundary.

## Modifications

- Move decode-state release out of the model runner and into the scheduler's terminal result adapter, using `finally` so result-conversion failures also release the row.
- Register `model.reset_request` as the scheduler abort callback. `release_row` remains idempotent, making abort-before-allocation and duplicate cleanup safe.
- Invalidate the cached active-row tensor whenever ownership changes, so a released request ID cannot reuse a stale row mapping.
- Keep sampler completion semantics on-device: `finished` still selects `EOS_SENTINEL` exactly as before. Remove only its host copy from async resolve metadata (`[B, n+3]` to `[B, n+2]`), which was used solely for model-runner-owned row reclamation; the scheduler's terminal and abort boundaries now own request lifetime.
- Add focused lifecycle tests covering a real SGLang `FINISH_LENGTH` result path, adapter failure, abort before/after allocation, complete row reset and mixed-batch reuse, and async resolve ownership.

## Root Cause

ZONOS2 treated the model sampler's internal EOS signal as the request-lifetime completion signal. The model runner released rows only when that signal became true. Scheduler termination is broader: a request can finish because of maximum length, abort, failure, or model EOS. In particular, `FINISH_LENGTH` completed downstream without releasing the ZONOS2 row.

`release_inactive()` ran before a later decode and therefore reclaimed request N when request N+1 arrived. It could not reclaim the final request after the server became idle.

## Accuracy Test

Tested on one NVIDIA H100 80GB with `Zyphra/ZONOS2` revision `65f1e80f94b599d474bb6af9094a803dc52f60bd`, PyTorch `2.11.0+cu130`, and async and sync decode.

| Case | Result |
|---|---|
| Exact parent `dc64a6b2`, four sequential `FINISH_LENGTH` requests | Reproduced: 4 terminals, 3 releases, final `held=1`, `free=63/64` |
| Candidate, same four-request reproduction | 4 releases, final `held=0`, `free=64/64` |
| Async/sync × streaming/non-streaming, `max_new_tokens=17/64/297` | 48/48 requests succeeded; valid audio; final `held=0`, `free=64/64` in every mode |
| Abort before state allocation | Cleanup was a no-op; free count remained 64 |
| Streaming abort after the first 4,096 PCM bytes | Owned row released immediately; no later request required; canary succeeded |
| Async mixed batch, lengths 17 and 297 | Short request released while the survivor completed 297 tokens; final `held=0`, `free=64/64` |
| 1,000 sequential requests alternating lengths 17/64 | 1,000/1,000 succeeded; final idle state `held=0`, `free=64/64` |

Across the instrumented runs, all 1,055 allocated rows were released, with one expected no-op for abort-before-allocation. Every successful release zeroed feedback, EOS/countdown, generation, and repetition state; ownership/free-list invariants had zero failures. The terminal set included both scheduler length completion and model EOS completion.

The 1,088 saved Fix EN outputs were scored without regeneration using `Qwen/Qwen3-ASR-1.7B` on physical GPU 0:

| Metric | Result |
|---|---:|
| Evaluated / skipped | 1,088 / 0 |
| Corpus WER | **1.34%** |
| Per-sample mean / median WER | 1.34% / 0.00% |
| Per-sample p95 / maximum WER | 9.09% / 66.67% |
| Corpus WER excluding samples above 50% | 1.29% |
| Samples above 50% WER | 1/1,088 (0.092%) |

The production lifecycle audit ran against `e1c3eb4d`; `f99e9ccb` changes tests only. The original test run at `e1c3eb4d` reported 44 passed and one test-design failure because it called `make_abort_callback()` before the builder's production setup sequence. The follow-up removes that unsupported pre-setup call and replaces the mock-only lifecycle checks with the focused production-boundary tests described above.

## Benchmark & Profiling

Full 1,088-sample SeedTTS EN comparison on the same H100, async decode, concurrency 16:

| Metric | Main `dc64a6b2` | Fix `e1c3eb4d` |
|---|---:|---:|
| Completed / failed | 1,088 / 0 | 1,088 / 0 |
| Throughput | 6.287 req/s | 6.345 req/s |
| Mean latency | 2.526 s | 2.508 s |
| Mean RTF | 0.6808 | 0.6724 |
| Corpus WER | 1.38% | 1.34% |
| WER excluding >50% outliers | 1.29% | 1.29% |
| >50% WER samples | 2 | 1 |

The lifecycle fix showed no throughput, latency, RTF, or aggregate WER regression in this matched full-corpus run.

## Checklist

- [x] Format the code according to pre-commit.
- [x] Add focused unit tests.
- [x] Update documentation and docstrings as needed; there is no public API or configuration change.
- [x] Provide H100 lifecycle, output, and performance results.

### DIFF
--- a/sglang_omni/models/zonos2/engine_builder.py
+++ b/sglang_omni/models/zonos2/engine_builder.py
@@ -220,6 +220,10 @@ class Zonos2EngineBuilder(TtsEngineBuilder):
 
         return make_zonos2_scheduler_adapters(model=model)
 
+    def make_abort_callback(self) -> Any | None:
+        assert self.model is not None
+        return self.model.reset_request
+
     def extra_scheduler_kwargs(self) -> dict[str, Any]:
         return {"enable_async_decode": self.async_decode}
 
--- a/sglang_omni/models/zonos2/model_runner.py
+++ b/sglang_omni/models/zonos2/model_runner.py
@@ -273,30 +273,30 @@ class Zonos2ModelRunner(ModelRunner):
             schedule_batch.output_ids = next_ids
 
         # launch_buf snapshots what the (lagged) resolve reads. Pack codes +
-        # the per-row EOS/finished state into one fresh int64 tensor (advanced
-        # indexing already copies, and cat makes a fresh tensor, so it is a
-        # snapshot even though the next launch mutates these pool rows). Record a
-        # CUDA event right after, so resolve's side-stream D2H waits only for this
-        # (codes ready), not the next forward. next_ids is cloned: the base
-        # aliases it onto output_ids, overwritten in place before this resolve.
+        # per-row EOS metadata into one fresh int64 tensor (advanced indexing
+        # already copies, and cat makes a fresh tensor, so it remains stable
+        # while the next launch mutates these pool rows). Record a CUDA event
+        # right after, so resolve's side-stream D2H waits only for this (codes
+        # ready), not the next forward. next_ids is cloned: the base aliases it
+        # onto output_ids, overwritten in place before this resolve.
         meta = torch.stack(
             (
                 pool.eos_frame_set[row_t].to(torch.int64),
                 pool.eos_frame_val[row_t],
-                finished.to(torch.int64),
             ),
             dim=1,
         )
-        packed = torch.cat((codes, meta), dim=1)  # [B, n+3] int64
+        packed = torch.cat((codes, meta), dim=1)  # [B, n+2] int64
         ev = torch.cuda.Event()
         ev.record()
         return (requests, packed, n, next_ids.clone(), ev)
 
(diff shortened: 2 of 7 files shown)


luojiaxuan (approved):
The cleanup boundary now matches scheduler terminal and abort semantics. The length-finish, adapter-failure, abort, mixed-batch, and async ownership paths are covered.

PR #1192, opened 2026-07-26
[TTS] Add MOSS-TTS-Realtime serving with kernel acceleration

luojiaxuan:
## What

Serving support for [MOSS-TTS-Realtime](https://huggingface.co/OpenMOSS-Team/MOSS-TTS-Realtime), a 24 kHz streaming TTS model for low-latency voice agents: Qwen3-1.7B global backbone, a 4-layer frame-local Transformer, and 16 RVQ codebooks from [MOSS-Audio-Tokenizer](https://huggingface.co/OpenMOSS-Team/MOSS-Audio-Tokenizer). 20 languages, optional zero-shot voice cloning from a reference clip.

## How

- **Global backbone** runs on the SGLang engine: paged attention, RadixAttention prefix caching, and batch-1 CUDA graphs for both the backbone and the full 16-codebook local decode.
- **Frame-local Transformer** uses a fused FlashAttention-3 kernel on Hopper (RoPE + KV update + GQA in a single launch), with an SDPA fallback elsewhere.
- **Audio path**: causal MOSS-Audio-Tokenizer encodes reference clips and vocodes generated frames to 24 kHz mono PCM, streamed as it is produced.
- **API**: standard `/v1/audio/speech` (WAV or streaming PCM). `ref_audio` accepts a local path (behind `--allowed-local-media-path`), an HTTP URL, or a base64 data URI. The pipeline serves one active request (request-1 latency design).

## Performance (1x H200, request-1)

| | native HF | this PR | delta |
|---|---:|---:|---:|
| TTFA, median | 299 ms | 134 ms | −55% |
| End-to-end latency, mean | 3.72 s | 1.49 s | −60% |
| RTF, mean | 0.66 | 0.26 | 2.5× faster |

Methodology, pinned revisions, and H100 local-attention numbers are under `benchmarks/results/`.

## Example

```bash
sgl-omni serve \
  --model-path OpenMOSS-Team/MOSS-TTS-Realtime \
  --config examples/configs/moss_tts_realtime.yaml \
  --allowed-local-media-path /data/voices \
  --port 8000

curl -X POST http://localhost:8000/v1/audio/speech \
  -d '{
    "model": "OpenMOSS-Team/MOSS-TTS-Realtime",
    "input": "你好，欢迎使用实时语音合成。",
    "ref_audio": "/data/voices/reference.wav",
    "response_format": "wav"
  }' -o out.wav
```

The synthesized speech follows the voice of `reference.wav`; omit `ref_audio` to use the model's default voice.

## Tests and docs

- Unit tests for the prompt/processor, local Transformer, model runner, and decode state pool: `tests/unit_test/moss_tts_realtime/`
- Cookbook page: `docs/cookbook/moss_tts_realtime.md`, registered in `docs/basic_usage/tts.md`


Hayden727:
Linked to #1120. I checked the actual local-attention implementation: `_resolve_attention_backend()` calls `sgl_kernel.flash_attn.is_fa3_supported()` and falls back to SDPA, rather than hard-coding Hopper/Ada/Blackwell checks. So there is no architecture-policy duplication to remove here; shared #1141 capability data is only useful later for consistent diagnostics/reporting.

PR #1195, opened 2026-07-27
Performance related follow-up after 0.5.15 bump

junliu-mde:
## Performance differences from #1183

This PR is stacked on #1183 and contains only performance/deployment changes that need independent ablation.

- **Fun-ASR:** batch audio-encoder inputs on GPU, increase request-build capacity for that batching path, and use a shorter encoder batching window.
- **Qwen3-Omni:** isolate text preprocessing and audio encoding from the AR process, bound OMP/tokenizer parallelism, and restore prefill-coalescing support.
- **Higgs TTS:** isolate preprocessing/audio encoding from the AR engine (#1159), reduce request and inter-stage transport overhead (#1180), bound preprocessing/OMP concurrency, and use codec decode CUDA Graphs for the default streaming frame domain.
- **MOSS-TD:** bound request-building and OMP contention and cap audio-encoder microbatches.
- **MOSS-TTS-Local:** isolate vocoder scheduling, assign explicit colocated GPU-memory budgets, and bound OMP contention.

## What remains

- Run each model group as an independent A/B against #1183 on the same H100 host; do not compare against historical CI numbers or a different server/client generation.
- Use the eight GPUs to run disjoint model groups in parallel, while keeping each A/B arm CPU/NUMA-isolated and free from cross-run cache warming.
- Repeat every arm and report generated-token or generated-audio throughput together with latency and TTFP tails; do not claim a gain from raw req/s when output lengths differ.
- Remove parameter-only changes that do not produce a repeatable benefit, then update this description with the final ablation table.
- Keep benchmark gates, fixed seeds, CI concurrency, and production NUMA policy unchanged.


junliu-mde:
Closing this draft instead of rebasing it. It still combines 38 commits across 94 files, conflicts with current main, and its required same-host A/B work is incomplete. The branch is based on the obsolete SGLang 0.5.15 transition, while current main pins 0.5.18 and later focused PRs have replaced several affected paths. Any remaining optimization should return as a small current-main PR with its own correctness tests and same-host A/B evidence.

PR #1198, opened 2026-07-27
[Consumer GPU] Add 24 GB TTS memory fallbacks

Hayden727:
## Summary

- add conservative low-memory fallbacks for MOSS-TTS and MOSS-TTS Local
- add memory-role mappings for FishAudio S2-Pro and Voxtral TTS
- preserve explicit environment overrides for users who want GPU preprocessing or float32 vocoding

## Dependency

Stacked on #1199. Merge #1199 first, then rebase this PR onto `main`. This PR contains only the low-memory model changes.

## Why

On 24 GB RTX 4090/4090D cards, colocated reference codecs and float32 vocoders consume enough memory to prevent otherwise viable pipelines from starting. The defaults use BF16 for the MOSS-TTS vocoder and CPU reference preprocessing for MOSS-TTS Local only on GPUs below 32 GiB.

## Validation

- RTX 4090D using `hongccc/sglang-omni:4090D`
- newly added behavior: 5 passed
- affected model suites: 140 passed; three unrelated Triton compilation tests were separately verified after correcting the server container runtime mount

Related to #1120.

### DIFF
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -1,0 +1,35 @@
+FROM hongccc/sglang-omni@sha256:6a8f60af7ca868dc266c118249d12fc73ba85e2e8075e5e31473bd25d349acfa
+
+ENV TRITON_LIBCUDA_PATH="/usr/lib/x86_64-linux-gnu"
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends sox \
+    && ln -sfn libcuda.so.1 /usr/lib/x86_64-linux-gnu/libcuda.so \
+    && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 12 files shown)


Hayden727:
Closing this mixed-model stack in favor of model-specific work, per the #1120 split-PR plan. MOSS-TTS Local is isolated in #1287, Voxtral TTS in #1288, Fish S2-Pro remains with its model PR #1243, and the 8B MOSS-TTS memory floor plus required 24/32 GB validation is now tracked explicitly in #1291. The old branch had unit-level fallback coverage but no end-to-end 8B consumer-GPU qualification, so it should not remain as an implied 24 GB support claim.

PR #1171, opened 2026-07-24
[Benchmark] Validate Fun-ASR Nano and Whisper Large v3 on RTX 4090

wirybeaver:
> [!NOTE]
> This draft stacked on the umbrella Draft of Qwen3-ASR 1.7B RTX4090 #1152, so GitHub will temporarily show the parent commits too. This Draft preserves the complete validation diff in one place and should not be merged independently. Check the Incremental review path.

## Summary
- add conservative BF16/SM89 profiles and startup diagnostics for Fun-ASR Nano and Whisper Large v3
- expose Whisper speech-to-English translation through `/v1/audio/translations` with bounded request validation
- add pinned, provenance-rich SeedTTS/CoVoST2 benchmarks and 30-minute mixed stability coverage
- publish complete RTX 4090 quality, performance, memory, stability, and known-limitation results

## Validation
- Fun-ASR SeedTTS EN/ZH: 1088/1088 and 2020/2020 at every level, three measured repeats, zero skips
- Whisper SeedTTS EN/ZH: 1088/1088 and 2020/2020 at every level, three measured repeats, zero skips
- Whisper CoVoST2 zh-to-English: 4898/4898, zero skips
- Fun-ASR 30-minute soak: 105,067 successful requests, zero unexpected errors
- Whisper 30-minute transcription/translation soak: 18,390 successful requests, zero unexpected errors
- Both soaks passed 28 malformed-input and 28 stream-cancel/reconnect events
- 156 focused unit tests and all changed-file pre-commit hooks passed

## Incremental review path
- #1153 — generic stream cancellation and SIGTERM lifecycle fixes
- #1155 — pinned benchmark provenance, resource monitoring, and stability foundation
- Personal-fork review PRs (no matching upstream incremental PRs opened):
  - https://github.com/wirybeaver/sglang-omni/pull/1 — Fun-ASR profile
  - https://github.com/wirybeaver/sglang-omni/pull/2 — Whisper profile and hardening
  - https://github.com/wirybeaver/sglang-omni/pull/3 — Whisper translations API
  - https://github.com/wirybeaver/sglang-omni/pull/4 — multi-model ASR validation
  - https://github.com/wirybeaver/sglang-omni/pull/5 — CoVoST2 translation benchmark
  - https://github.com/wirybeaver/sglang-omni/pull/6 — documentation capstone

The incremental PRs are the intended review and merge path.

## Known limitations
[...]

PR #1199, opened 2026-07-27
[Consumer GPU] Add architecture-aware Fish backend and runtime diagnostics

Hayden727:
## Summary

- report the GPU architecture family (Ada, Hopper, or Blackwell) alongside the exact SM version, including SM103/B300 as Blackwell
- choose the FishAudio S2-Pro Slow-AR default only after `server_args_overrides` merge: FA3 on SM90 and FlashInfer on validated SM89/SM100/SM120 targets
- preserve an explicit operator `attention_backend` without running automatic capability/backend selection
- fail before worker construction when compute capability is unknown/unsupported or when an auto-selected FlashInfer backend is unavailable
- log the resolved architecture plus effective prefill/decode attention backends and sampling backend from the SGLang-owned bootstrap layer

Architecture-family reporting and model policy remain separate. SM100, SM103, and SM120 are all reported as Blackwell because the diagnostic already includes the exact `sm=<version>`. Fish S2-Pro automatic backend selection continues to use a model-local explicit SM allowlist; SM103 remains unvalidated and therefore requires an explicit operator override.

## Scope

This PR intentionally preserves the repository's existing installation contract:

```bash
uv pip install -v -e .
```

The earlier uv project-level exclusion for `sglang==0.5.16 -> flash-attn-4`, the `fa4` extra, the `uv sync` documentation, and the CI uv pin were removed. Making upstream SGLang's external FA4 dependency optional is a packaging concern and should be handled separately rather than changing Omni's source-install workflow in this architecture PR.

## Priority and coordination

This is the shared GPU-architecture/runtime-diagnostic baseline for #1120. Dependent consumer-GPU PRs should rebase after it lands and remove duplicated policy hunks. In particular, #1243 should retain its Fast-AR fallback and validation while dropping overlapping Slow-AR selection code after rebase.

## Validation

- 96 focused architecture, GPU compatibility, Fish S2-Pro pipeline, and SGLang bootstrap tests passed locally
- tests cover SM89/SM90/SM100/SM120 selection; SM100/SM103/SM120 Blackwell-family identity while retaining Fish auto-selection rejection for SM103; unknown and unsupported SM failures; `SGLANG_IS_FLASHINFER_AVAILABLE=false`; explicit override precedence; inherited global phase backends; and explicit split-phase backends
- targeted pre-commit passed on all changed files
- diagnostic smoke for SM103 produced `sm=103, architecture=blackwell`
- source-level smoke confirmed an SM89 automatic default fails with an actionable Fish-specific error when upstream `is_flashinfer_available()` observes `SGLANG_IS_FLASHINFER_AVAILABLE=false`
- prior branch validation on RTX 4090D (SM89): `SM89 -> ada -> flashinfer`; GPU smoke must be rerun by CI after this revision

## Required CI follow-up

- Fish S2-Pro worker startup plus one request on SM90/FA3 and available SM89/SM100/SM120 FlashInfer hardware
- confirm the startup diagnostic reports the exact SM, architecture family, and effective prefill/decode backends selected by `ServerArgs.get_attention_backends()`

Closes part of #1120.

### DIFF
--- a/sglang_omni/models/fishaudio_s2_pro/engine_builder.py
+++ b/sglang_omni/models/fishaudio_s2_pro/engine_builder.py
@@ -10,6 +10,7 @@ from typing import Any
 from sglang_omni.models.fishaudio_s2_pro import request_builders
 from sglang_omni.models.fishaudio_s2_pro import stages as fish_stages
 from sglang_omni.scheduling.engine_factory import TtsEngineBuilder
+from sglang_omni.utils.gpu_compat import get_visible_gpu_sm_version
 from sglang_omni.vendor.sglang.server_args import override_server_args
 
 
@@ -40,6 +41,7 @@ class FishS2ProEngineBuilder(TtsEngineBuilder):
         dtype: str,
     ) -> dict[str, Any]:
         del dtype
+        sm_version = get_visible_gpu_sm_version(self.gpu_id)
         return {
             "max_running_requests": 64,
             "disable_cuda_graph": False,
@@ -47,13 +49,12 @@ class FishS2ProEngineBuilder(TtsEngineBuilder):
             "chunked_prefill_size": 8192,
             "dtype": "bfloat16",
             "enable_torch_compile": True,
+            "attention_backend": "fa3" if sm_version == 90 else "flashinfer",
             "random_seed": int.from_bytes(os.urandom(4), "little") & 0x7FFFFFFF,
         }
 
     def customize_server_args(self, server_args: Any) -> None:
         updates: dict[str, Any] = {"disable_overlap_schedule": True}
-        if server_args.attention_backend is None:
-            updates["attention_backend"] = "fa3"
         override_server_args(
             server_args,
             "sglang_omni.fishaudio_s2_pro.runtime_defaults",
--- a/sglang_omni/models/fishaudio_s2_pro/engine_builder.py
+++ b/sglang_omni/models/fishaudio_s2_pro/engine_builder.py
@@ -12,6 +12,44 @@ from sglang_omni.models.fishaudio_s2_pro import stages as fish_stages
 from sglang_omni.scheduling.engine_factory import TtsEngineBuilder
 from sglang_omni.utils.gpu_compat import get_visible_gpu_sm_version
 from sglang_omni.vendor.sglang.server_args import override_server_args
+from sglang_omni.vendor.sglang.utils import is_flashinfer_available
+
+_VALIDATED_AUTO_ATTENTION_BACKENDS = {
+    89: "flashinfer",
+    90: "fa3",
+    100: "flashinfer",
+    120: "flashinfer",
+}
+
+
+def _select_default_attention_backend(*, gpu_id: int) -> str:
+    sm_version = get_visible_gpu_sm_version(gpu_id)
+    if sm_version is None:
+        raise RuntimeError(
+            "FishAudio S2-Pro cannot select a default attention backend because "
+            f"CUDA compute capability for gpu_id={gpu_id} could not be detected. "
+            "Set server_args_overrides.attention_backend to an explicitly "
+            "validated backend."
+        )
+
+    backend = _VALIDATED_AUTO_ATTENTION_BACKENDS.get(sm_version)
+    if backend is None:
+        raise RuntimeError(
+            "FishAudio S2-Pro cannot select a default attention backend because "
+            f"SM{sm_version} is not validated. Supported automatic selection "
+            "targets are SM89, SM90, SM100, and SM120; otherwise set "
+            "server_args_overrides.attention_backend explicitly."
+        )
+
+    if backend == "flashinfer" and not is_flashinfer_available():
+        raise RuntimeError(
+            f"FishAudio S2-Pro selects FlashInfer on SM{sm_version}, but "
+            "FlashInfer is unavailable. Install and enable FlashInfer, ensure "
+            "SGLANG_IS_FLASHINFER_AVAILABLE is not false, or set "
+            "server_args_overrides.attention_backend to an explicitly "
+            "validated backend."
+        )
+    return backend
 
 
 class FishS2ProEngineBuilder(TtsEngineBuilder):
@@ -41,7 +79,6 @@ class FishS2ProEngineBuilder(TtsEngineBuilder):
         dtype: str,
     ) -> dict[str, Any]:
         del dtype
-        sm_version = get_visible_gpu_sm_version(self.gpu_id)
         return {
             "max_running_requests": 64,
             "disable_cuda_graph": False,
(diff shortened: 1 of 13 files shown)


LauraGPT:
@Hayden727 The current red `omni-ci-gate` is a policy gate, not a code-test result. In [run 30252573829](https://github.com/sgl-project/sglang-omni/actions/runs/30252573829), job `89933729586` fetched an empty label set and stopped at `Require run-ci label`; the ASR, TTS, and test jobs were skipped.

Because #1199 is the shared prerequisite for the Fun-ASR consumer-GPU work in #1154 and #1171, could a maintainer with CI permission add the `run-ci` label or run `/tag-and-rerun-ci`? I also verified that `LauraGPT` cannot add the label directly (GitHub returns 403), so I am leaving the permissioned action to a maintainer rather than issuing a command the handler cannot honor.

sglang_omni/models/fishaudio_s2_pro/engine_builder.py
```diff
@@ -40,20 +41,20 @@ def generation_defaults(
...
         del dtype
+        sm_version = get_visible_gpu_sm_version(self.gpu_id)
         return {
             "max_running_requests": 64,
             "disable_cuda_graph": False,
             "mem_fraction_static": 0.85,
             "chunked_prefill_size": 8192,
             "dtype": "bfloat16",
             "enable_torch_compile": True,
+            "attention_backend": "fa3" if sm_version == 90 else "flashinfer",
```
JiaxinD: When `get_visible_gpu_sm_version` returns `None`, which happens when both the NVML and torch queries fail and is only logged at debug, this now selects flashinfer, where the old code gave fa3 whenever the backend was unset. The other sm conditional sites here (arkasr, fun_asr, qwen3_asr) all use `sm_version is not None and sm_version >= 100` so a failed probe keeps the default; matching that shape would avoid a silent flip away from fa3 on an H100.

JiaxinD (commented):
Splitting the packaging change out was the right call, and the remaining diff checks out: `server_args_overrides` still beats the model defaults, since `build_generation_batch_overrides` merges `**incoming` after `**stage_defaults`, so an explicit operator `attention_backend` survives and the new test exercises that real path. `get_attention_backends()` does exist on upstream ServerArgs at 0.5.16, and `create_sglang_infrastructure_defer_cuda_graph` delegates to the logging path, so the diagnostic fires for the Fish and ASR stages too. One thing below.

sglang_omni/models/fishaudio_s2_pro/engine_builder.py
Ratish1: letes validate backend availability before hard-selecting flashinfer. SGLang’s normal resolver honors SGLANG_IS_FLASHINFER_AVAILABLE=false and falls back when FlashInfer is unavailable, but supplying
attention_backend="flashinfer" here bypasses that resolution.

Resolve this after user overrides are merged, preserve explicit overrides, and either select a validated fallback or fail early with an actionable Fish-specific error. Otherwise worker initialization can fail later while constructing the FlashInfer backend.

tests/unit_test/fishaudio_s2_pro/test_pipeline.py
```diff
@@ -821,6 +835,41 @@ def test_s2pro_engine_disables_generic_compile_after_local_compile(
...
 
 
+@pytest.mark.parametrize(
+    ("sm_version", "expected_backend"),
+    [
+        (89, "flashinfer"),
+        (90, "fa3"),
+        (100, "flashinfer"),
+        (120, "flashinfer"),
+        (None, "flashinfer"),
```
Ratish1: A failed capability query should not be treated as evidence that FlashInfer is compatible. None can represent broken device mapping, unavailable NVML/PyTorch discovery, or another startup configuration
problem.

Please replace this expectation with an actionable startup failure, or explicitly validate FlashInfer availability before using it as the fallback. Add coverage for SGLANG_IS_FLASHINFER_AVAILABLE=false
as well.

sglang_omni/utils/gpu_compat.py
```diff
@@ -116,6 +131,22 @@ def get_visible_gpu_sm_version(
     return major * 10 + minor
 
 
+def describe_sglang_runtime_configuration(server_args: Any, gpu_id: int) -> str:
```
Ratish1: This helper introduces an SGLang ServerArgs contract into the otherwise model-independent GPU compatibility layer, including an untyped Any and a call to get_attention_backends().

Please move the SGLang-specific renderer under scheduling/sglang_backend, or keep it local to bootstrap, and leave gpu_compat.py responsible for GPU facts such as SM discovery and architecture naming.

sglang_omni/models/fishaudio_s2_pro/engine_builder.py
Hayden727: Fixed in 3f2672b2. A failed capability probe no longer silently flips the model default to FlashInfer. Automatic selection now runs after user overrides merge; without an explicit backend, unknown and unsupported SM values fail before worker construction with an actionable override path.

Hayden727: Fixed in 3f2672b2. `adjust_overrides()` now resolves the model default after `server_args_overrides` merge and returns immediately for an explicit `attention_backend`. Auto-selected FlashInfer calls the vendored upstream `is_flashinfer_available()`, which honors the environment flag, import availability, and CUDA; failure produces a Fish-specific startup error before `ModelWorker` construction.

tests/unit_test/fishaudio_s2_pro/test_pipeline.py
Hayden727: Fixed in 3f2672b2. The `None -> flashinfer` expectation was removed. New production-path coverage rejects unknown capability and unsupported SM80, rejects FlashInfer when `SGLANG_IS_FLASHINFER_AVAILABLE=false`, and proves an explicit backend remains usable when automatic discovery is unavailable.

sglang_omni/utils/gpu_compat.py
Hayden727: Fixed in 3f2672b2. The SGLang-specific renderer and its typed `Protocol` now live privately in `scheduling/bootstrap.py`; `utils/gpu_compat.py` retains only GPU SM discovery and architecture naming. The inherited-global and split-phase diagnostic tests moved to the bootstrap test module.

tests/unit_test/pipeline/test_gpu_architecture.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@pytest.mark.parametrize(
+    ("sm_version", "architecture"),
+    [
+        (89, "ada"),
+        (90, "hopper"),
+        (100, "blackwell-datacenter"),
+        (120, "blackwell-consumer"),
+        (103, "sm103"),
```
HowardZorn: SM103 is also a known Blackwell variant (B300) for data centers. (https://github.com/QwenLM/FlashQLA/issues/27)

Hayden727: Updated in a6b775d4. Shared GPU facts now classify SM100, SM103/B300, and SM120 under the `blackwell` architecture family. The startup diagnostic already logs the exact `sm=<version>`, so separate datacenter/consumer architecture labels duplicated information. Fish S2-Pro policy remains independent and keyed by its explicit validated SM allowlist; SM103 still requires an explicit operator backend override.

Ratish1 (approved):

PR #1203, opened 2026-07-27
[Benchmark][ASR] Report RTFx and repeat coverage

Ratish1:
## Motivation

The SeedTTS ASR sweep already records successful input-audio duration and benchmark wall time, but its result assembly did not expose the resulting audio throughput. Its repeat summary also assumed every repeat produced WER, latency, and RTF values, which could fail formatting or obscure coverage when requests were skipped.

## Modifications

- Expose successful input-audio seconds per wall-clock second as `rtfx` in the shared ASR speed result while preserving the existing per-request RTF metrics.
- Represent unavailable WER, latency, and RTF values explicitly, aggregate only valid repeat values, and report each metric's valid-repeat count.
- Sum evaluated, total, and skipped requests across repeats and show `evaluated/requests` in the concurrency table so partial work remains visible.
- Print RTFx in per-repeat and aggregate benchmark output and document its ASR semantics for the SeedTTS, Fun-ASR, and Qwen3-ASR benchmark paths.

## H100 Validation

Full SeedTTS EN sweep with Fun-ASR-Nano, using 1,088 samples and three measured repeats per concurrency:

| Concurrency | RTFx | Requests/s | Evaluated | Corpus WER |
|---:|---:|---:|---:|---:|
| 1 | 117.552 | 24.820 | 3264/3264 | 1.71% |
| 2 | 202.273 | 42.709 | 3264/3264 | 1.71% |
| 4 | 333.948 | 70.511 | 3264/3264 | 1.71% |
| 8 | 496.598 | 104.854 | 3264/3264 | 1.71% |
| 16 | 663.429 | 140.079 | 3264/3264 | 1.71% |
| 32 | 775.889 | 163.824 | 3264/3264 | 1.71% |

All 19,584 measured requests completed, and corpus WER remained stable across the sweep.

### DIFF
--- a/benchmarks/eval/benchmark_asr_seedtts.py
+++ b/benchmarks/eval/benchmark_asr_seedtts.py
@@ -250,8 +250,8 @@ def _aggregate(repeats: list[dict]) -> dict:
     }
 
 
-def _format_metric(value: float | None, digits: int) -> str:
-    return "n/a" if value is None else f"{value:.{digits}f}"
+def _format_metric(value: float | None, digits: int, suffix: str = "") -> str:
+    return "n/a" if value is None else f"{value:.{digits}f}{suffix}"
 
 
 def _print_table(aggregates: list[dict]) -> None:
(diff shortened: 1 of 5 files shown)


benchmarks/eval/benchmark_asr_seedtts.py
```diff
@@ -234,26 +250,36 @@ def _stat(key: str) -> dict:
     }
 
 
+def _format_metric(value: float | None, digits: int) -> str:
+    return "n/a" if value is None else f"{value:.{digits}f}"
```
AkazaAkane: _format_metric(None, ...) returns n/a, so appending s here produces n/as. Could we conditionally include the unit, or let the formatting helper accept a suffix, so unavailable latency prints as n/a?

AkazaAkane (commented):
The full benchmark validates the normal all-success path, but the primary behavior added here is for partial or fully skipped repeats. Could we add a small unit test for _aggregate and formatting with:

one valid repeat plus one all-skipped repeat;
all repeats skipped;
missing RTF while latency/WER remain valid?

It should verify summed evaluated/total/skipped, metric n counts, None aggregation, and that throughput/RTFx still include zero-throughput failed repeats.

benchmarks/eval/benchmark_asr_seedtts.py
Ratish1: `_format_metric()` now appends the suffix only for numeric values, so unavailable latency renders as `n/a` rather than `n/as`.

SandyLuXY (approved):
lgtm

PR #1206, opened 2026-07-28
[Feat] Add SGLang-compatible abort request API

JingwenGu0829:


sglang_omni/pipeline/coordinator.py
```diff
@@ -543,6 +548,24 @@ def _on_abort_task_done(
...
+        elif rid:
+            request_ids = [
+                request_id
+                for request_id in self._requests
+                if request_id.startswith(rid)
+            ]
+        else:
+            return 0
+
+        results = await asyncio.gather(
```
yxs, line 564: `abort_requests` gathers without `return_exceptions`, so one failure loses the whole batch

PR #1217, opened 2026-07-28
[Qwen3-Omni Perf] Code2Wav batching sweep harness and waveform tolerance gate

dyyfk:
## Motivation

Measurement infrastructure for the two open items left by #1126 (roadmap #1018 item 4.2, issue #1026): sweep `max_batch_wait_ms` x `batch_floor` on a Code2Wav-bound high-concurrency workload to decide the batching defaults, and replace bitwise equality with a waveform tolerance gate for this audio-touching change.

The serving-level benchmark cannot pose the question: #1126 showed no E2E change at c1-8 because Talker windows rarely align, so coalescing seldom fires. This PR adds a component-level harness that drives `Code2WavScheduler` directly through its inbox/outbox contract, following the aligned-windows-fed-straight-to-the-scheduler approach of #1126's component A/B, plus a staggered-arrivals mode that models the misaligned case the default has to survive.

This PR is code-only (no GPU numbers yet). The sweep itself lands as follow-up evidence on #1026 per #1018's contract: exploratory sweep first, then K=3 paired repeats with an A/A noise band for the final candidates. Per item 4.2, keeping the default off is an acceptable outcome if no combination passes.

## Modifications

- `benchmarks/eval/benchmark_code2wav_batching.py`: parameterized Code2Wav-bound load benchmark.
  - Three arms: `serial-eager` (#1126 control), `serial-graph` (production default since #1101), `batched`. Batching and CUDA graph are mutually exclusive today, so the default-on decision must beat the graph-serial baseline, not just eager; the three arms separate the two effects.
  - Arrival modes: `aligned` (coalescing upper bound) and `staggered` (per-stream Poisson frame arrivals with random phase). Codes and arrival timings are seeded by (seed, repeat, stream) only, so runs pair across arms and wait/floor settings.
  - Per run: TTFA p50/p95, xRT, per-request RTF, chunks/request, forward batch-size histogram, single-request forward share (the item 4.2 gate metric), fire-reason counts, per-forward execution mode, and forward busy fraction. Failed requests invalidate the run rather than being dropped.
  - `--compare-waveforms` runs a lockstep batched-vs-serial equivalence pass on identical codes (free-running window boundaries are timing-dependent, so waveforms are only comparable under lockstep feeding).
  - `--fake` runs a deterministic CPU vocoder for smoke testing without a checkpoint.
- `benchmarks/metrics/waveform_tolerance.py`: SNR + peak absolute diff + exceeding-sample fraction, with `tolerance_failures` returning explicit failure reasons. Initial thresholds proposed from #1126's measurements (bf16 peak ~9e-2 on ~0.2% of samples; fp32 ~2e-3): SNR >= 40 dB, peak <= 0.2, <= 1% of samples over 1e-2. These are a starting proposal for the exploratory run, not a decided gate.
- Tests: `tests/unit_test/benchmarks/test_waveform_tolerance.py` (metric math, threshold reasons, length-mismatch and silent-reference edges) and `test_code2wav_batching_benchmark.py` (serve-loop smoke for serial and batched arms on the CPU fake, lockstep equivalence attains batch >= 2 with SNR = inf on the deterministic model).

No product code is modified; the existing `code2wav_batch_start/end` profiler events already expose batch size and fire reason for serving-side validation, so no new instrumentation was needed for the gate metric.

## Related Issues

Part of #1022. Follow-up to #1126; work items tracked in #1026. Design and benchmark contract: #1018 (item 4.2).

## Accuracy Test

N/A (benchmark tooling only; no model-side code touched). The waveform tolerance checker itself is unit-tested.

## Benchmark & Profiling

CPU smoke (`--fake`, both modes, serial-eager + batched arms, `--compare-waveforms`): all runs valid, aligned mode attains full batch, staggered shows partial coalescing, equivalence pass exact (SNR = inf) on the deterministic model. H100 sweep results will follow on #1026.

## Checklist

- [x] Add unit tests.
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed. (Phase 2: exploratory sweep, then K=3 paired repeats per #1018.)


dyyfk:
H100 results are in: exploratory sweep (112 runs) + K=3 confirmation (72 runs) posted on https://github.com/sgl-project/sglang-omni/issues/1026#issuecomment-5113455550. Summary: no wait/floor combination passes the item 4.2 gate on misaligned traffic (batching regresses ~4x xRT and 6-8x TTFA vs the serial CUDA-graph default; forward-share gate fails everywhere); the aligned upper bound does beat graph serial by ~17% at c16 with floor=8. Recommendation: keep enable_batching default off; mechanism analysis and a shape-quantized-fire direction for Track 4.4 are in the issue comment. bf16 waveform tolerance measured at SNR 35.4-39.1 dB / peak <= 0.048 / <= 0.3% samples over 1e-2 -> proposing SNR >= 35 dB, peak <= 0.1, <= 0.5% as the gate defaults.

edwingao28:
Close this PR as it is for benchmark purpose only

PR #1222, opened 2026-07-29
[TTS][MOSS] Decouple MOSS-TTS Delay codec and run reference encoding on GPU

CloudRipple:
## Motivation

The MOSS-TTS Delay processor currently owns and loads the MOSS audio tokenizer. This couples prompt construction to codec execution and leaves reference-audio encoding on CPU in the preprocessing stage.

Under concurrent Seed-TTS workloads, CPU reference encoding cannot feed prepared requests to the AR stage quickly enough. This PR separates the codec from the processor and places the preprocessing codec on the stage-assigned GPU, while keeping an explicit CPU override for deployments with limited GPU memory.

## Modifications

- Add a standalone `MossTTSAudioTokenizer` loader and compatibility wrapper for reference encoding and generated-code decoding.
- Construct the upstream MOSS processor with `audio_tokenizer=None`, then load dedicated codec instances for preprocessing and vocoder execution.
- Preserve codec checkpoint metadata from `processor_config.json`, honor an explicit `codec_model_path`, and fall back to `OpenMOSS-Team/MOSS-Audio-Tokenizer` only when checkpoint metadata is absent or falsey.
- Route string and `os.PathLike` reference inputs through the standalone reference encoder and the shared `load_audio` entry point, covering local files, file URLs, HTTP(S), and base64 data URIs consistently.
- Place the preprocessing stage on GPU by default and derive the codec device from the placement-provided `gpu_id`. An explicit `device` argument still takes precedence, so `device="cpu"` remains available.
- Keep preprocessing codec execution in FP32; this PR does not change model weights, sampling parameters, or codec algorithms.
- Rename `codec.py` to `delay_pattern.py` because the module contains delay-pattern segmentation rather than codec model execution.
- Add regression coverage for codec metadata resolution, explicit codec overrides, stage GPU placement, explicit CPU placement, `PathLike` references, shared audio loading, processor-compatible code layout, and standalone vocoder decoding.

## Related Issues

#1233

## Accuracy Test

| Metric | Latest main | Current | Current - main | Paired bootstrap 95% CI |
|---|---:|---:|---:|---:|
| Raw corpus WER | 1.2978% | 1.1471% | -0.1507 pp | [-0.3551, +0.0414] pp |
| WavLM speaker similarity | 68.8304 | 69.0354 | +0.2050 | [-0.1322, +0.5432] |
| UTMOS | 3.9507 | 3.9455 | -0.0052 | [-0.0169, +0.0066] |

## Benchmark & Profiling

Benchmark configuration:

- Model and dataset: MOSS-TTS-v1.5, full Seed-TTS English set
- 1088 measured requests per run, three independent runs per group
- Fresh service and 32-request warmup for every run
- NVIDIA A800-SXM4-80GB, exclusive physical GPU
- Request concurrency 16; preprocessing concurrency 8
- `mem_fraction_static=0.65`
- Fixed per-request seed base `20260720`
- `max_new_tokens=2048`
- Group means are compared directly; results are not averaged from per-run ratios

| Metric | Latest main mean | Current mean | Relative change |
|---|---:|---:|---:|
| Request throughput | 2.9958 QPS | 4.4395 QPS | **+48.19%** |
| Audio throughput | 13.4210 audio-s/s | 19.8879 audio-s/s | **+48.19%** |
| Output-token throughput | 266.6246 tok/s | 395.1052 tok/s | **+48.19%** |
| Benchmark wall time | 363.1991 s | 245.0723 s | **-32.52%** |
| Mean request latency | 5.3156 s | 3.5904 s | **-32.46%** |
| P95 request latency | 6.4677 s | 4.5429 s | **-29.76%** |

Per-run QPS:

- Latest main: `2.9877`, `2.9691`, `3.0306` (CV `1.05%`)
- Current: `4.4414`, `4.4244`, `4.4529` (CV `0.32%`)

All 6528 measured requests succeeded. No run was flagged by Modified-Z (`|z| > 3.5`) or Tukey 3x-IQR checks. No HTTP failure, CUDA OOM, traceback, or foreign GPU process was observed. Even the conservative slowest-current versus fastest-main comparison is `+45.99%` QPS.

Observed service GPU memory increased from approximately 60.9 GiB to 64.0 GiB (`+3.1 GiB`) because preprocessing now owns a GPU codec instance. Deployments that prefer lower GPU memory usage can explicitly configure the preprocessing codec with `device="cpu"`.

## Checklist

- [x] Add unit tests.

## CI

if labeled.


sglang_omni/models/moss_tts/config.py
```diff
@@ -40,6 +40,8 @@ def generation_sglang_role_to_stage(cls) -> dict[str, str]:
             name="preprocessing",
             process="pipeline",
             factory=f"{_PKG}.stages.create_preprocessing_executor",
+            factory_args={"dtype": "float32"},
+            gpu=0,
```
SandyLuXY, line 44: Please update the cookbook with the new GPU footprint, the CPU configuration, and the current reference-decoding dependencies.

CloudRipple: Will update the cookbook after this series of optimizations is complete.

SandyLuXY (approved):

estellaliu233:
Hi — I've been putting together a write-up on diagnosing host-bound serving: how to tell a GPU bubble apart from a slow kernel, which workloads are most exposed to it, and what engines and operators can each do about it. I used this PR as the **host-overhead** case — CPU work at a stage boundary issues no CUDA calls, so on an Nsight timeline it shows up as a bubble with an empty CUDA API row:

The numbers are quoted from the PR — A800 c16 2.9958 → 4.4395 QPS (+48.19%), three runs per group at 1088 requests each with preprocessing concurrency 8, and roughly +3.1 GiB of GPU memory for the dedicated preprocessing codec.

If I've misread the context anywhere, or overstated a conclusion the PR didn't actually measure, I'd be glad to correct it — thanks for the work.

CloudRipple:
> Hi — I've been putting together a write-up on diagnosing host-bound serving: how to tell a GPU bubble apart from a slow kernel, which workloads are most exposed to it, and what engines and operators can each do about it. I used this PR as the **host-overhead** case — CPU work at a stage boundary issues no CUDA calls, so on an Nsight timeline it shows up as a bubble with an empty CUDA API row:
> 
> https://estellaliu233.github.io/posts/gpu-bubbles-host-bound/
> 
> The numbers are quoted from the PR — A800 c16 2.9958 → 4.4395 QPS (+48.19%), three runs per group at 1088 requests each with preprocessing concurrency 8, and roughly +3.1 GiB of GPU memory for the dedicated preprocessing codec.
> 
> If I've misread the context anywhere, or overstated a conclusion the PR didn't actually measure, I'd be glad to correct it — thanks for the work.

@estellaliu233  Thanks for the write-up, and for taking the time to read through the PR. Happy to have it used as a case study, but I'd like to correct two points so the article doesn't carry a conclusion the PR doesn't actually support.

1. The ~+3.1 GiB extra GPU memory was an artifact of the weight-loading implementation at the time, not a property of the design. That figure came from how the codec weights were read in that revision. In the current implementation the encoder and decoder loading for the codec has been fixed, and the dedicated preprocessing codec no longer requires that extra GPU memory. So please don't present +3.1 GiB as an inherent overhead introduced by decoupling the codec.

2. The main work in this PR is not a host-bound story. The PR is about decoupling the MOSS-TTS delay codec and moving work that belongs on the GPU to where it belongs. The CPU-side gap you observe at the stage boundary is a symptom of that work having been in the wrong place — it is precisely what the change removes — rather than the optimization target or an accepted cost of the architecture.

estellaliu233:
> > Hi — I've been putting together a write-up on diagnosing host-bound serving: how to tell a GPU bubble apart from a slow kernel, which workloads are most exposed to it, and what engines and operators can each do about it. I used this PR as the **host-overhead** case — CPU work at a stage boundary issues no CUDA calls, so on an Nsight timeline it shows up as a bubble with an empty CUDA API row:
> > https://estellaliu233.github.io/posts/gpu-bubbles-host-bound/
> > The numbers are quoted from the PR — A800 c16 2.9958 → 4.4395 QPS (+48.19%), three runs per group at 1088 requests each with preprocessing concurrency 8, and roughly +3.1 GiB of GPU memory for the dedicated preprocessing codec.
> > If I've misread the context anywhere, or overstated a conclusion the PR didn't actually measure, I'd be glad to correct it — thanks for the work.
> 
> @estellaliu233 Thanks for the write-up, and for taking the time to read through the PR. Happy to have it used as a case study, but I'd like to correct two points so the article doesn't carry a conclusion the PR doesn't actually support.
> 
> 1. The ~+3.1 GiB extra GPU memory was an artifact of the weight-loading implementation at the time, not a property of the design. That figure came from how the codec weights were read in that revision. In the current implementation the encoder and decoder loading for the codec has been fixed, and the dedicated preprocessing codec no longer requires that extra GPU memory. So please don't present +3.1 GiB as an inherent overhead introduced by decoupling the codec.
> 2. The main work in this PR is not a host-bound story. The PR is about decoupling the MOSS-TTS delay codec and moving work that belongs on the GPU to where it belongs. The CPU-side gap you observe at the stage boundary is a symptom of that work having been in the wrong place — it is precisely what the change removes — rather than the optimization target or an accepted cost of the architecture.

Thanks for the careful correction — both points are fair. I'll revise the post to drop the +3.1 GiB , and to reframe #1222 as moving GPU-appropriate work out of the wrong place rather than as a host-bound optimization target. I'll follow up here once it's updated.

PR #1225, opened 2026-07-29
[TTS] Use model-owned streaming vocoder defaults

### LINKED ISSUE #1224: [TTS] Use model-owned streaming vocoder defaults
The speech API injects initial_codec_chunk_frames=1 for every streaming TTS model. Codec frame rates and steady vocoder cadence differ, so this lowers first-audio latency by starving playback.

H100 calibration:
- Higgs: 100ms continuity 0% -> 100%, underrun p95 0.893s -> 0.012s at 20 frames.
- MOSS-TTS Local: 9.4% -> 100%, 0.373s -> 0.002s at 5 frames.
- ZONOS2: 0% -> 100%, 3.151s -> 0s at 40 frames.

Let the API forward only explicit overrides and let each model own its calibrated default.

luojiaxuan:
Fixes #1224.

The speech API forced a 1-frame first vocoder chunk across models. This improved first-audio latency by starving playback.

Changes:
- Forward only explicit initial_codec_chunk_frames overrides.
- Let the shared resolver apply model defaults while preserving explicit 0.
- Use 20 frames for Higgs, 5 for MOSS-TTS Local, and 40 for ZONOS2.
- Propagate Higgs defaults to the producer; default first flush is row 27.
- Derive the ZONOS2 producer first flush from the shared 18-row lookahead; default is row 58.

H100 calibration:

| Model | Initial frames | Continuity 100ms | Underrun p95 |
| --- | ---: | ---: | ---: |
| Higgs | 1 -> 20 | 0% -> 100% | 0.893s -> 0.012s |
| MOSS-TTS Local | 1 -> 5 | 9.4% -> 100% | 0.373s -> 0.002s |
| ZONOS2 | 1 -> 40 | 0% -> 100% | 3.151s -> 0s |

Higgs and MOSS use C16 with 32 EN samples and two repeats. ZONOS2 uses a C1 directional sweep because its C16 same-reference warmup exposed a separate hang.

Default-path H100 check without request overrides: Higgs 8/8 and ZONOS2 8/8 succeeded. Higgs first audio was 0.8s with ITL p99 0.589s; ZONOS2 first audio was 0.441s with ITL p99 0.301s.

Tests: 214 passed. Black, isort, and F401 checks passed.


sglang_omni/models/higgs_tts/stages.py
```diff
@@ -476,6 +477,7 @@ def create_vocoder_executor(
     max_batch_wait_ms: int = 2,
     stream_stride: int = DEFAULT_HIGGS_STREAM_STRIDE,
     stream_followup_stride: int = DEFAULT_HIGGS_STREAM_FOLLOWUP_STRIDE,
+    initial_chunk_frames: int = DEFAULT_HIGGS_INITIAL_CHUNK_FRAMES,
```
SandyLuXY, line 482: This installs the 20-frame default only on the vocoder, but the Higgs AR producer treats missing stream metadata as `0` and still waits 75 rows before its first message; with eight codebooks, a 20-frame first chunk should be forwarded at row 27. Propagate the same resolved default to the TTS-engine/request-builder path, preserve explicit `0`.

sglang_omni/models/zonos2/components/streaming_vocoder.py
```diff
@@ -41,7 +41,7 @@
 _STREAM_STEADY_CHUNK_FRAMES = (
     40  # ~0.46 s steady chunk (fewer vocoder forwards/chunk -> lower streaming RTF)
 )
-_STREAM_INITIAL_CHUNK_FRAMES = 5  # ~0.06 s first chunk for low TTFB
+_STREAM_INITIAL_CHUNK_FRAMES = 40
```
SandyLuXY: This 40-frame consumer threshold is incompatible with the producer's 24-then-32 row cadence and the OLA's 18-row withhold: 40 decoded frames need 58 rows, but the first two messages total 56, so first audio waits for a third message at row 88. Derive the producer's first flush from the same resolved threshold/lookahead (or configure it to at least 58 rows) and cover the producer-to-vocoder TTFA boundary.

sglang_omni/models/higgs_tts/stages.py
luojiaxuan: Fixed. The producer now receives the resolved 20-frame default, flushes at row 27, and preserves explicit 0. Added boundary tests.

sglang_omni/models/zonos2/components/streaming_vocoder.py
luojiaxuan: Fixed. The producer first flush now uses the resolved chunk and shared 18-row lookahead; the default is 58 rows. Added a producer-to-vocoder boundary test.

sglang_omni/models/zonos2/callbacks.py
```diff
@@ -102,7 +106,12 @@ def extract_zonos2_output(runner, result, scheduler_output, outputs) -> None:
         # length). Adaptive: the FIRST chunk uses a smaller threshold so the
         # first audio is produced sooner (lower TTFC); steady chunks use `chunk`.
         first = runner._stream_emit_first_chunk_frames
-        threshold = first if (first > 0 and start == 0) else chunk
+        if start == 0 and INITIAL_CODEC_CHUNK_FRAMES_PARAM in stream_metadata:
```
SandyLuXY, line 109: `build_zonos2_stream_metadata()` now always adds `initial_codec_chunk_frames`, so every production streaming request takes this branch and `runner._stream_emit_first_chunk_frames` is never used; even configured values `0` and `100` both flush at the derived default row 58. Preserve the configured producer boundary when the request omits an override, use this branch only for explicit request values (including `0`)

docs/basic_usage/tts.md
SandyLuXY: The docs still describe the removed global `initial_codec_chunk_frames=1` behavior in the TTS parameter table, API-server design, and Higgs cookbook, while the ZONOS2 cookbook still documents the old `24 -> 32` producer default and says the now-bypassed `stream_emit_first_chunk_frames=0` disables it. Please update all four surfaces to match the final model-owned defaults and request/config precedence.

sglang_omni/models/zonos2/callbacks.py
SandyLuXY: fixed in [8ea2b6d](https://github.com/sgl-project/sglang-omni/pull/1225/commits/8ea2b6d4208c3a389efc47c6947b18f483c0d629)

docs/basic_usage/tts.md
SandyLuXY: fixed in [8ea2b6d](https://github.com/sgl-project/sglang-omni/pull/1225/commits/8ea2b6d4208c3a389efc47c6947b18f483c0d629)

SandyLuXY (approved):

Issue #1202, opened 2026-07-27
TTS serving CI: batch32 latency gate is flaky by construction (p95 over 9 samples)


JiaxinD:
`TTS CI / stage 4 - serving stress` fails on unrelated PRs. The `batch32.latency_p95_s_max` gate takes p95 over only **9 samples**, so on that sample size p95 is effectively the max.

The `soak-300s` stage has a fat tail in passing runs too, so the gate passes or fails depending on which workload the tail happens to land on.

| run | soak median | soak p95 | soak p99 | share > 1.9825s | batch32 p95 |
|---|---|---|---|---|---|
| pass | 0.432 | 1.033 | 7.520 | 2.3% (12/512) | 0.952 |
| fail | 0.429 | 1.066 | 9.836 | 4.1% (21/512) | 10.192 |

The 9 batch32 latencies, same workload hash and same model in both runs:

* pass: 0.898 0.896 0.859 0.814 0.911 0.928 0.871 0.850 0.968
* fail: 0.914 0.822 2.186 0.864 0.902 **10.239 10.122 9.658** 1.017

The server is not slower: medians match, and 2126/2126 requests succeeded in the failing run with queue wait at 0.001s. Three of the nine batches simply landed in the tail.

Using the passing run's own 2.3% tail, P(at least one of 9 samples in the tail) is 19%; at 4.1% it is 31%.

Seen on #1049 and on `perf/code2wav-bounded-batching` (9.5 to 9.7s), both 3/3 attempts, so it is not PR specific.

Options: raise the sample count for `batch_32_all_valid`, gate on a median instead of a 9-sample p95, or set `SERVING_BATCH32_LATENCY_P95_S_REF` above the observed tail.

Sheehan20:
I'd like to take the median-gate approach here

PR #1230, opened 2026-07-30
[Bug-Fix] Fix flaky batch32 CI gate: hard-gate on p50, advisory-only on p95 (#1202)

PoTaTo-Mika:
## Motivation

The TTS serving CI `batch32.latency_p95_s_max` gate is flaky by construction. The `soak-300s` stage with `batch_size=32` produces only ~9 completed requests per run. With 9 samples, p95 is effectively the maximum, so a single tail-latency straggler (~8s when the median is ~0.9s) pushes p95 far past the 1.586s threshold and fails CI on otherwise healthy code.

Issue #1202 documents this with probability analysis: at a ~2.3% per-request tail rate, P(≥1 tail in 9 samples) ≈ 19%, meaning roughly one in five CI runs flakes for reasons unrelated to the PR under test.

## Modifications

1. **`tests/utils.py`** — Add `warnings` list and `warn()` method to `MetricCheckCollector` so gates can emit advisory diagnostics without failing CI.

2. **`tests/test_model/test_tts_serving_ci.py`**:
   - Add `"p50"` to `MetricGate.statistic` (the underlying benchmark report already computes `p50`).
   - Add `advisory: bool = False` to `MetricGate`. When `True`, threshold violations call `warn()` instead of `fail()`.
   - Replace the single `batch32.latency_p95_s_max` gate with two gates:
     - **`batch32.latency_p50_s_max`** (hard gate, `statistic="p50"`) — blocks CI only when the *median* degrades.
     - **`batch32.latency_p95_s_advisory`** (advisory gate, `statistic="p95"`, `advisory=True`) — warns when the tail spikes but does not block CI.
   - `_check_performance`: exclude advisory gates from the "needs calibration" check; route advisory violations through `warn()`.
   - `_write_benchmark_validation`: emit warnings via `logging.warning()` and include them in the output JSON.

3. **`.claude/skills/tune-ci-thresholds/models/tts/stages.yaml`** — Add `latency_p50_s` (PENDING) and `latency_p95_advisory_s` (OK) metrics for the batch32 gate group.

4. **`.claude/skills/tune-ci-thresholds/models/tts/config.yaml`** — Add `latency_p50_s` JSON path for the batch32 workload.

## Related Issues

#1202

## Verification

Tested against a real calibration-run benchmark result (9 batch32 samples, p50=0.92s, p95=8.68s):

| Gate | Statistic | Threshold | Value | Result |
|------|-----------|-----------|-------|--------|
| OLD `batch32.latency_p95_s_max` | p95 | 1.586s | 8.68s | ❌ FAIL (false positive) |
| NEW `batch32.latency_p50_s_max` | p50 | 1.0s | 0.92s | ✅ PASS |
| NEW `batch32.latency_p95_s_advisory` | p95 | 2.0s | 8.68s | ⚠️ WARN only |

The median gate correctly passes when overall latency is healthy, and the advisory gate surfaces the tail straggler without blocking CI.

Mock-gate unit tests covering five scenarios (normal, median-regression, tail-outlier, catastrophic, borderline) all produce the expected pass/fail/warn outcomes.

## Checklist

- [ ] Add unit tests.


.claude/skills/tune-ci-thresholds/models/tts/stages.yaml
```diff
@@ -779,13 +779,23 @@ tts_serving_batch32_speed:
       json_file: "tts-serving-ci0/benchmark/results.json"
       json_path: "metrics.by_stage_and_workload.soak-300s.batch_32_all_valid.metric_sample_counts.latency_s"
   metrics:
-    latency_p95_s:
-      source: "SERVING_BATCH32_LATENCY_P95_S_REF"
+    latency_p50_s:
```
Ratish1, line 782: lets make these metrics discoverable instead of manually adding them to the generated stage registry. Running tune.py discover --model tts currently removes both latency_p50_s and  latency_p95_advisory_s: the metric registry does not recognize LATENCY_P50, both p95 constants collapse to latency_p95_s, and the existing SERVING_BATCH32_LATENCY_P95_S_REF is selected instead.

Ratish1 (commented):
Thanks for the fix @PoTaTo-Mika , could you fix the lint and update the title

PoTaTo-Mika:
I fixed the keys and regenerated metrics with `tune.py`.
If there's anything else I need to do, please feel free to tell me.

tests/utils.py
```diff
@@ -11,7 +11,6 @@
 from typing import TYPE_CHECKING
 
 import pytest
-
 from benchmarks.benchmarker import utils as benchmark_utils
```
JiaxinD, line 15: isort wants the blank line kept after `import pytest`, here and in test_tts_serving_ci.py, and black reflows the `pending` genexp; `lint` is red on all three. `pre-commit run --all-files` fixes it.

tests/test_model/test_tts_serving_ci.py
```diff
@@ -209,14 +214,25 @@ class MetricGate:
...
+        "batch32.latency_p95_s_advisory",
         "stage_workload",
         "latency_s",
         "p95",
         "max",
-        _maximum(SERVING_BATCH32_LATENCY_P95_S_REF),
+        _maximum(SERVING_BATCH32_LATENCY_P95_ADVISORY_S_REF),
         stage="soak-300s",
         workload="batch_32_all_valid",
+        advisory=True,
```
JiaxinD, line 240: The same construction is still hard gated at `long.latency_p95_s_max`: `long_prefill_decode` builds exactly 5 scenarios (one per speech enabled stage, and the 2126 total matches #1202's 2126/2126), so its p95 is also just the max. Bigger headroom (27.46s limit against ~22s) is why it has not fired yet, but it is worth the same treatment or a follow up.

tests/test_model/test_tts_serving_ci.py
```diff
@@ -567,12 +587,20 @@ def _write_benchmark_validation(
     threshold_checks: MetricCheckCollector,
 ) -> None:
     failures = benchmark_checks.failures + measurement_checks.failures
+    all_warnings = (
+        benchmark_checks.warnings
+        + measurement_checks.warnings
+        + threshold_checks.warnings
+    )
+    for w in all_warnings:
+        logger.warning("CI advisory: %s", w)
```
JiaxinD, line 612: This never reaches the CI log: there is no `log_cli`, so pytest only replays log records for failing tests, which I confirmed with a passing `-v -s` run. `benchmark_validation.json` is also absent from `artifact-path-globs` in test-tts-ci.yaml, so adding `*/benchmark_validation.json` there would make the advisory visible.

JiaxinD (commented):
Replaying #1202's own failing run, the new p50 (1.017 against 1.146) passes where the old p95 (10.239 against 1.9825) fails, and `discover` on this head keeps both new metrics, so @Ratish1's thread is closed. One body fix: after the 1.25 slack the effective limits are 1.14625 and 1.9825, not the 1.0 / 2.0 / 1.586 in the table, and `latency_p50_s` ships as OK rather than PENDING.

.claude/skills/tune-ci-thresholds/models/tts/stages.yaml
```diff
@@ -1,952 +1,13 @@
...
-      source: "SERVING_MIXED_BURST_LATENCY_P95_S_REF"
-      json_file: "tts-serving-ci0/benchmark/results.json"
-      json_path: "metrics.by_stage.mixed-burst-512.latency_s.p95"
-      worst: max
-      display:
-        label: "Latency p95 (s)"
-        scale: 1
-        digits: 3
-      status: OK
 tts_serving_long_speed:
```
JiaxinD, line 951: `fa60501c` drops 22 of the 23 stanzas here (+15/-987), including `tts_serving_batch32_speed`, so the two metrics the earlier commit registered are gone again along with every Higgs and MOSS quality stanza. Looks like `discover` ran scoped to one stage; a full `tune.py --model tts discover` on a Linux checkout should restore it.

tests/test_model/test_tts_serving_ci.py
```diff
@@ -62,7 +66,8 @@
 SERVING_MIXED_BURST_LATENCY_P95_S_REF: float | None = 17.466
 SERVING_LONG_PROMPT_TOKENS_MIN_REF: float | None = 684
 SERVING_LONG_COMPLETION_TOKENS_MIN_REF: float | None = 321
-SERVING_LONG_LATENCY_P95_S_REF: float | None = 21.968
+SERVING_LONG_LATENCY_P50_S_REF: float | None = None
```
JiaxinD: A `None` ref on a non advisory gate puts `long.latency_p50_s_max` straight into `pending`, so `threshold_checks.assert_all()` raises before any measurement is read; I ran it on this head and stage 4 fails unconditionally. Either calibrate it or leave the gate advisory until it is.

tests/test_model/test_tts_serving_ci.py
```diff
@@ -51,7 +54,8 @@
 SERVING_REST_STREAM_INTER_CHUNK_P95_S_REF: float | None = 0.6098
 SERVING_REST_STREAM_LATENCY_P95_S_REF: float | None = 0.685
 SERVING_REST_STREAM_RTF_P95_REF: float | None = 0.1507
-SERVING_BATCH32_LATENCY_P95_S_REF: float | None = 1.586
+SERVING_BATCH32_LATENCY_P50_S_REF: float | None = 0.917
```
JiaxinD: This branch predates #1260, so all 31 `SERVING_*` constants here are the pre recalibration values and both this file and stages.yaml are in conflict; merging as is would revert that recalibration (`SERVING_WS_STREAM_LATENCY_P95_S_REF` 77.850 back to 37.662, soak 7.185 back to 4.441). Needs a rebase, with the p50 rederived on the new baseline.

JiaxinD (commented):
All three earlier comments are addressed: lint is green, `long` got the same p50 plus advisory split, and `*/benchmark_validation.json` is in the artifact globs. The new head has blockers below though, and the stages.yaml one undoes part of this PR's own earlier commit.

tests/test_model/test_tts_serving_ci.py
```diff
@@ -51,7 +54,8 @@
...
+SERVING_BATCH32_LATENCY_P95_ADVISORY_S_REF: float | None = 1.586
```
JiaxinD: Worth knowing the flake case moved: #1260 raised this ref to 10.222, so the resulting 12.78s ceiling already passes #1202's failing run at 10.239. The p50 split is still the better shape, but per the tune-ci-thresholds contract the new ref needs worst-of-5 with the run directory and calibration commit in an Evidence section.

Ratish1:
Hey @PoTaTo-Mika , could you fix the conflicts. Also, is there a need to change the long prefill decode scenario, the flaky CI was only for the batch32 correct?. Thanks for the fixes.

PoTaTo-Mika:
> Hey @PoTaTo-Mika , could you fix the conflicts. Also, is there a need to change the long prefill decode scenario, the flaky CI was only for the batch32 correct?. Thanks for the fixes.

I'm just about to resolve the conflicts :)

.claude/skills/tune-ci-thresholds/models/tts/stages.yaml
JiaxinD, line 782: The 11 serving stanzas still record `7518e47e...` while the head file hashes to `fa2f5a55...`, so `tune.py` stops at the sha check with `test sha mismatch` and serving cannot be calibrated at all. Rerunning `discover` fixes that but then reports `51 metric(s) OK, 2 need config`, the two being the now unreferenced `SERVING_BATCH32_LATENCY_P95_S_REF` and `SERVING_LONG_LATENCY_P95_S_REF`, so those want deleting first.

tests/test_model/test_tts_serving_ci.py
```diff
@@ -52,6 +55,10 @@
 SERVING_REST_STREAM_LATENCY_P95_S_REF: float | None = 0.9647539779543877
 SERVING_REST_STREAM_RTF_P95_REF: float | None = 0.22127384815467604
 SERVING_BATCH32_LATENCY_P95_S_REF: float | None = 10.222142587881535
+SERVING_BATCH32_LATENCY_P50_S_REF: float | None = (
```
JiaxinD, line 57: `1.2` is a hand typed round placeholder, which CONTRACT.md rules out in favour of N=5 worst of five; it permits 1.5s against the p50 actually observed in #1202 (0.896 passing, 1.017 failing), so roughly 50 to 67 percent of median drift. It also cannot be rewritten later: `_read_bare_value` returns `None` for this constant and for the long one on this head, because the regex wants the number at end of line and black's wrap breaks that (executed).

tests/test_model/test_tts_serving_ci.py
```diff
@@ -63,6 +70,10 @@
 SERVING_LONG_PROMPT_TOKENS_MIN_REF: float | None = 684.0
 SERVING_LONG_COMPLETION_TOKENS_MIN_REF: float | None = 94.0
 SERVING_LONG_LATENCY_P95_S_REF: float | None = 28.912055992987007
+SERVING_LONG_LATENCY_P50_S_REF: float | None = (
```
JiaxinD, line 72: With this advisory and its threshold `None`, and the p95 advisory too, `long_prefill_decode` now has no hard latency gate at all; before this PR it was hard at 28.912 x 1.25 = 36.14s. Keeping the old p95 hard until the p50 is calibrated would avoid dropping that coverage in the meantime.

JiaxinD (commented):
All three blockers from the last round are cleared: the 23 stanzas are back, `long.latency_p50_s_max` no longer forces an unconditional `assert_all()` failure (executed against a synthetic report, both collectors pass), and the rebase restored main's post #1260 values with all 27 inherited constants matching `upstream/main` exactly. What is left is the calibration workflow rather than the gate logic.

Issue #1236, opened 2026-07-31
[Feature] Traffic-aware adaptive dispatch policy for Code2Wav (Qwen3-Omni first)


dyyfk:
## Motivation

The Qwen3-Omni Code2Wav stage currently has two **static, mutually exclusive** execution modes (`code2wav_scheduler.py` raises if both are enabled):

- **serial + CUDA graph** (production default): one request per forward, fixed window shapes, ~100% graph replay.
- **wait/floor batching + eager** (#1126, default off per the evidence in #1026): accumulate compatible stream chunks, batch them, but no graph.

The H100 sweep for #1026 (112 exploratory + 72 confirm runs, posted in https://github.com/sgl-project/sglang-omni/issues/1026#issuecomment-5113455550 / #1217) showed that **neither static mode dominates — the winner depends on traffic**:

- Under **staggered (realistic) arrivals**, fixed batching is ~4× worse than serial-graph in xRT (24–34 vs 104–114) and 6–8× worse in TTFA. Mechanism: request windows grow to different shapes, so shape buckets almost never match — batches simply don't form, and the shape churn keeps eager kernels cold (2–5 ms/frame vs ~0.5 ms/frame for hot graph replay).
- Under **aligned arrivals**, batching wins: at concurrency 16 with floor 8 it beats serial-graph by ~17% (xRT 136 vs 116).

A dispatcher that observes traffic could get the best of both — batch when arrivals align, stay on the serial-graph path when they don't — instead of committing to one mode at config time. Discussed with @edwingao28: not currently on the roadmap, starting with Qwen3-Omni and generalizing later.

## Proposed design

**Prerequisite (evidence-driven): shape-quantized firing.** The sweep showed batches fail to form because window shapes diverge. Capping each dispatch at exactly one stream chunk of new frames collapses all steady-state windows to a tiny shape set `{context, context + chunk}` — making requests batchable *and* graph-replayable, so the `enable_batching × enable_cuda_graph` exclusivity can be lifted. Backlog drains over repeated uniform steps instead of one oversized forward.

**Signals the dispatcher can observe** (per shape queue):

- queue depth; oldest-request wait time; recent arrival rate
- predicted arrival time of the next compatible chunk
- available CUDA-graph shapes; historical execution time per batch size
- whether the GPU stage is currently busy
- whether a request is still waiting for its **first audio** (TTFA guard)
- request priority / SLO

**Actions per decision point:**

- fire a request solo now / fire the current batch now
[...]

PR #1237, opened 2026-07-31
[Qwen3-Omni Perf] Chunk-aligned Code2Wav dispatch with batched CUDA graph capture

dyyfk:
## Motivation

Mechanism PR for #1236 (traffic-aware Code2Wav dispatch), part of the #1022 Qwen3-Omni performance effort.

Today `Code2WavScheduler` forces a config-time choice: serial + CUDA graph (production default) **or** wait/floor batching + eager (#1126) — enabling both raises a `ValueError`. The H100 sweep for #1026 (write-up: https://github.com/sgl-project/sglang-omni/issues/1026#issuecomment-5113455550) showed why the batching path loses under realistic staggered arrivals: per-request windows grow to different shapes, so shape buckets almost never match — batches don't form, and the shape churn keeps eager kernels cold while the serial path replays its CUDA graph at full speed.

This PR adds the prerequisite mechanism from #1236: **chunk-aligned dispatch with batched CUDA graph capture**. Each step consumes exactly `stream_chunk_size` new frames, so every window stays inside the serial CUDA graph key set `{chunk, 2·chunk, …, chunk + context}` and per-request backlogs drain over repeated uniform steps instead of one variable-shape forward. With uniform windows, batched graphs become practical: requested `{2, 4, 8}` batch sizes share the serial frame set, and the runner publishes the batched keys that fit the configured graph-memory budget alongside the required batch-1 keys.

No new config knob is introduced — setting `enable_batching=True` together with `enable_cuda_graph=True` (previously rejected) activates chunk-aligned dispatch. Legacy batching (graph off) and the serial path are byte-for-byte unchanged, and production defaults are untouched, so this PR is inert until the #1236 policy work opts in.

## Modifications

`code2wav_scheduler.py`, `code2wav_cuda_graph.py`, and unit tests:

- Remove the `enable_batching × enable_cuda_graph` `ValueError` in `Code2WavScheduler.__init__` and `create_code2wav_scheduler`.
- Add `_step_frames()`: under chunk-aligned dispatch, cap each step at one stream chunk; otherwise preserve the previous full-ready-depth behavior. The scheduler uses this consistently for bucketing, window slicing, and profiler metadata.
- Generate requested graph keys from the serial frame thresholds and the scheduler's `{2, 4, 8}` decomposition sizes, capped by `batch_ceiling`.
- Capture all requested keys in one shared CUDA graph mempool. Batch sizes greater than 1 are a best-effort tier captured largest-first; memory-pressure retries drop the largest remaining batch-size class, while non-capacity tier-1 failures preserve the atomic batch-1 tier. A batch-1 capture failure still disables the runner.
- Expose `available_batch_sizes(frames)` from the graphs that were actually published. Chunk-aligned dispatch and step-plan decomposition follow that live per-frame availability, so partially published matrices use only replayable sub-batches; a runtime disable immediately stops chunk-aligned dispatch.
- Keep eager fallback for ineligible inputs and key misses, and disable batching only when the runner cannot publish even the serial tier.
- Record every sub-batch execution in `code2wav_batch_end.sub_batch_execution` so mixed graph/eager steps remain observable.
- Add coverage for scheduler coexistence, key generation and factory assembly, chunk drain and bucket behavior, dynamic decomposition, graph hits and fallbacks, tiered capture degradation, telemetry, waveform parity, and real-CUDA batch-1/batch-2 alternating replay from one shared pool.

Focused local validation: **97 passed, 2 skipped**. The skipped cases are the real-CUDA tests on a host without CUDA.

## Related Issues

Mechanism part of #1236. Builds on #1126, motivated by the evidence in #1026. Ref #1022, #1018.

## Accuracy Test

- Unit test asserts **exact waveform equality** between chunk-aligned dispatch and the serial reference on the CPU test model (same forwards, same windows, different grouping only).
- The bf16 waveform-tolerance gate from #1217 ran on H100 in both A/B rounds: lockstep equivalence vs the serial reference at c16 with max attained batch 8 — worst per-request SNR 35.35 dB (≥ the 35 dB bf16 floor measured in #1026), 0/16 failures.
- No kernels or model architecture touched.

## Benchmark & Profiling

Done — two H100 A/B rounds per the #1018 measurement contract, on the integration branch combining this PR with the #1217 component benchmark (`quantized` arm).

**Setup**: H100 via Modal (torch 2.11), staggered Poisson arrivals (frame-interval 5 ms, the talker-bound backlog regime from #1026), 20 windows/request, K=3 repeats per cell (means reported), arms `serial-graph` (production default) vs `quantized` (chunk-aligned dispatch; historical arm label). Tags: `ab3-confirm-*-20260811-100408`, `ab3-scale-*-20260811-100921`, `ab3-c1-*-20260811-105909`. Cross-fraction absolutes are not comparable (different H100 instances); all deltas are within-fraction. Measured on the integration head prior to the `main` sync; the only subsequent change on the Code2Wav path is a formatting-only edit, so the benchmarked code path is behaviorally identical to the current head.

### Production configuration (`total_gpu_memory_fraction=0.02` — runner publishes 12/16 graphs, B8 class over budget, shrink logged and served)

| concurrency | audio xRT serial -> chunk-aligned (zero-wait) | delta | xRT chunk-aligned (wait 8 ms, floor 8) | delta | TTFA p50 zero-wait | TTFA p50 heavy-wait |
| 1 | 14.5 -> 14.6 | ~0% | 13.2 | -9% | 0.060 -> 0.060 | 0.060 |
| 8 | 90.5 -> 100.2 | **+11%** | 96.1 | +6% | 0.061 -> 0.062 | 0.061 |
| 16 | 93.9 -> 107.6 | **+15%** | 141.7 | **+51%** | 0.064 -> 0.061 | 0.063 |
| 32 | 119.1 -> 134.5 | **+13%** | 168.7 | **+42%** | 0.079 -> 0.074 | 0.071 |
| 64 | 118.5 -> 132.5 | **+12%** | 167.1 | **+41%** | 0.136 -> 0.134 | 0.162 |

### Full graph set (fraction 0.05 — controlled comparison, 16/16 graphs; isolates what the missing B8 class costs and provides continuity with all previously posted numbers)

| concurrency | xRT serial -> chunk-aligned (zero-wait) | delta | xRT chunk-aligned (wait 8 ms, floor 8) | delta | TTFA p50 zero-wait | TTFA p50 heavy-wait |
| 1 | 14.5 -> 14.6 | ~0% | 13.5 | -7% | 0.060 -> 0.060 | 0.060 |
| 8 | 106.5 -> 106.8 | ~0% | 102.2 | -4% | 0.056 -> 0.056 | 0.056 |
| 16 | 116.9 -> 133.9 | **+15%** | 179.4 | **+53%** | 0.056 -> 0.054 | 0.054 |
| 32 | 113.7 -> 129.1 | **+14%** | 175.4 | **+54%** | 0.085 -> 0.073 | 0.074 |
| 64 | 112.5 -> 127.4 | **+13%** | 174.4 | **+55%** | 0.192 -> 0.141 | 0.141 |

### Findings

1. **Zero-wait at the production budget: +11-15% at every concurrency >=8, TTFA parity or better.** The missing B8 class is immaterial on this path — the 0.02 deltas match the full-graph-set deltas within noise.
2. **The B8 class only matters under heavy waiting at high concurrency.** There 8-wide groups actually form (single-forward share 0.33-0.47) and replaying them as 2xB4 costs ~12 points of the +54% (still +41-42%), with c64 TTFA slipping slightly below baseline (0.162 vs 0.136). We do **not** propose raising the default fraction to recover this: the vocoder's demand for batched graphs and the colocated talker's demand for KV cache both rise with concurrency, so the ~0.4 GB the B8 pool needs would be taken from talker KV in exactly the regime where KV is scarcest. When-not-to-use: leave the fraction at 0.02 unless a deployment is demonstrably vocoder-bound with measured slack on the talker GPU.
3. **No regression from the merge**: full-graph-set c32/64 matches the pre-merge scale sweep within noise.
4. **Correctness**: waveform equivalence passes at both fractions (worst SNR 35.26 dB at 0.02, where `max_batch=4` shows the B4 decompose path is genuinely exercised; 35.37 dB at 0.05; zero tolerance failures). The 0.02 build logs `published 12/16` and serves — the graceful-degradation path measured here is exactly what production runs.
5. Fixed batching without graphs stays far below baseline (xRT 32-53): batching only pays with batched graph replay, unchanged from #1026.
6. Defaults are unchanged by this PR (batching off). The heavy-wait penalty at low concurrency (c1 -7-9%, c8 -4%) against its gain at c16+ (+41-55%) is the zero-wait/heavy-wait crossover — load-scaled waiting remains #1236's policy question.

Graph-capture provenance (tag `graph-probe-20260811-092029`): at prod 0.02 the merged runner publishes 12/16 keys — all of {B1,B2,B4} x {t10,t20,t30,t35}, retained pool 634 MB of the 1207.5 MB budget; the B8 class is dropped legitimately (b8t35 retains 1266 MB > budget), the shrink retries once and publishes the rest (`attempts=2`, no disable). At 0.05: 16/16, single attempt, zero skipped keys. Policy table per 8 due windows (T=35): B8 graph 23.9 ms < B8 eager 28.0 < 2xB4 graph 29.6 < 4xB2 33.7 < 8xB1 replays 47.4.

Results JSONs with full provenance are retained; happy to attach any of them.

## Checklist

- [x] Add unit tests.


dyyfk:
> Naming note: the mechanism was initially called *shape-quantized dispatch*; it is now **chunk-aligned dispatch** (nothing is rounded away — leftover frames defer to the next step, like a scheduler quantum). The eval harness arm and the recorded JSONs keep the historical `quantized` label.

## H100 A/B results (two rounds): up to +32% staggered / +92% aligned over the production default, TTFA parity

**Setup** (per #1018): H100 80GB (Modal), torch 2.11.0+cu130; `Qwen/Qwen3-Omni-30B-A3B-Instruct` code2wav weights; #1217 component harness (`benchmark_code2wav_batching`) with synthetic codec streams; c∈{8,16}; aligned + staggered (Poisson) arrivals; `--frame-interval-ms 5` (talker-bound backlog regime, per #1026); K=3 repeats. Result JSONs carry both commit SHAs (eval branch = #1217 harness ⊕ this PR).

- Round 1: eval `bc72a957` (PR head `418ed0d6`, before batched graph keys) — 84/84 valid runs
- Round 2: eval `ba3d431d` (PR head `38f6d20c`) — 72/72 valid runs, `--graph-memory-fraction 0.05`

### Round 1 — mechanism only: found the bottleneck

Chunk alignment worked as designed: ~25% of staggered forwards batched (the fixed #1126 batcher forms almost none — shape mismatch), 3–4× faster than the fixed batcher, and free when batches don't form (aligned ≡ serial-graph, 115 vs 115 xrt). But it **lost to the production default under staggered load** (78 vs 103 xrt at c8, 88 vs 112 at c16; TTFA 2.5×): multi-request groups ran eager, and one batch-2 eager forward is slower than two size-1 graph replays. That result motivated the batched-key commits now in this PR (`91c9d1fd`, `47533b76`, `38f6d20c`).

### Round 2 — with batched graph keys (bs {2,4,8} × serial frames, 16 graphs)

xrt (audio-seconds per wall-second, higher is better; K=3 means; baseline = `serial-graph` production default):

| regime | serial-graph | chunk-aligned zero-wait (0,1) | chunk-aligned w2/f8 | chunk-aligned w8/f8 |
|---|---|---|---|---|
| aligned c8 | 114.3 | 114.2 | **219.6 (+92%)** | 219.2 |
| aligned c16 | 114.8 | 111.7 | **219.9 (+92%)** | 217.4 |
| staggered c8 | 102.4 | 103.3 (parity) | 87.3 (**−15%**) | 91.9 (−10%) |
| staggered c16 | 110.8 | **118.9 (+7%)** | 138.0 (+25%) | **146.4 (+32%)** |

**TTFA p50: parity at every operating point** — 0.022–0.023 s (aligned c8), 0.042–0.045 (aligned c16), 0.051–0.067 (staggered, vs 0.053–0.061 serial-graph). Round 1's 2.5× TTFA regression is gone: first-ready requests still fire immediately, and a batched first-chunk forward is now one graph replay. (Two isolated outlier repeats are visible in the JSONs — aligned-c16 (0,1) r1 and staggered-c16 w8/f8 r2 — one-off TTFA spikes that don't move the xrt conclusions.)

**Mechanism evidence:** 16/16 graphs published within the 0.05 memory fraction; **11,981 graph replays, zero `key_miss`** across the whole run — the chunk-aligned windows are closed over the captured key set under real load, as designed. The only eager fallbacks are 333 `ineligible` final-drain decodes (expected by design). At staggered c16 the forward count per run drops from 320 (serial) to 82–92.

**Waveform gate:** lockstep equivalence vs the serial reference, c16, max attained batch 8: worst per-request SNR **35.35 dB** (≥ the 35 dB bf16 floor measured in #1026), 0/16 failing requests — both rounds.

### When to use it / when not (the data #1236 asked for)

- **High concurrency, realistic arrivals (staggered c16): win.** +7% with zero-wait, +25…32% with a modest wait/floor policy.
- **Bursty/synchronized arrivals (aligned): big win.** Floor-8 batching nearly doubles throughput at zero TTFA cost.
- **Low concurrency (staggered c8): do NOT wait.** Waiting is −10…−15% — arrivals are too sparse to fill batches, so waiting just idles the GPU (busy 0.80 → 0.63–0.72). Zero-wait stays at parity, so the *mechanism* is safe there; only the *wait policy* must scale with load. This is exactly the decision #1236's adaptive policy will own.
- **Memory:** 16 graph keys instead of 4 (~fraction 0.02 → 0.05 in the harness). If the batched capture doesn't fit, the factory falls back to serial-only keys and disables batching — tight-memory deployments degrade to today's exact behavior.

**Scope note:** component-level benchmark — real code2wav weights, synthetic codec streams at controlled arrival rates (same methodology as #1126/#1217). End-to-end validation behind a real talker is the natural follow-up alongside the #1236 policy work.

Defaults are unchanged by this PR (both flags must be opted into), and the zero-wait configuration measured ≥ parity at every operating point. Raw JSONs with per-run batch histograms, runner stats and full provenance: `ab-confirm-20260804-221732.json`, `ab2-smoke-20260804-233429.json`, `ab2-confirm-20260804-233514.json` — happy to attach on request.

Marking ready for review.

dyyfk:
## Concurrency scale sweep c ∈ {1…128}: zero-wait is regression-free everywhere; waiting crosses over between c8 and c16 and reaches +58%

Follow-up to the A/B above, mapping the full concurrency range (same setup; eval `ba3d431d`, `ab2-scale-20260805-000731.json`, 126/126 valid runs, K=3). Arms: `serial-graph` production default, chunk-aligned zero-wait (0,1), chunk-aligned heavy-wait (8 ms / floor 8).

**Staggered (realistic Poisson arrivals), xrt vs baseline:**

| c | serial-graph | zero-wait (0,1) | heavy-wait (8,8) |
|---|---|---|---|
| 1 | 14.1 | 14.1 (±0%) | 13.1 (−8%) |
| 2 | 27.4 | 27.3 (±0%) | 24.8 (−9%) |
| 4 | 54.0 | 53.9 (±0%) | 48.3 (−11%) |
| 8* | 102.4 | 103.3 (±0%) | 91.9 (−10%) |
| 16 | 112.0 | 125.2 (+12%) | 160.0 (+43%) |
| 32 | 111.7 | 127.1 (+14%) | 169.4 (+52%) |
| 64 | 111.9 | 127.2 (+14%) | **176.6 (+58%)** |
| 128 | 109.8 | 123.9 (+13%) | 169.2 (+54%) |

\* c8 row from the ab2-confirm run above (same commit/config family).

Aligned holds ~+90% for heavy-wait at every c ≥ 16 (215–221 vs 113–115) with zero-wait at parity. (Aligned c ≤ 4 heavy-wait rows in the JSON are dominated by end-of-stream drain decodes — few forwards — and shouldn't be read as steady-state.)

**Takeaways:**

1. **Zero-wait is regression-free across the entire range** — bitwise-identical behavior to serial at c1, ±0% through c8, +12–14% from c16 up. This is the configuration that can safely be always-on.
2. **The wait policy's regression zone is c ≤ 8** (−8…−11%): arrivals are too sparse to fill batches, so the wait mostly idles the GPU (and in a finite run, stretches the tail). The crossover sits between c8 and c16; beyond it the gain grows to +52…58% (c32–64) and plateaus ≈ +54% at c128 — no cliff at high concurrency, batch-8 decomposition and bucket handling hold up at 128 streams (82,707 graph replays this run, still **zero `key_miss`**).
3. **One prediction falsified, and it matters for #1236:** we expected zero-wait to converge toward heavy-wait at high concurrency ("enough streams are due at once — no need to wait"). It doesn't: zero-wait plateaus at +13% with `single_share` ≈ 0.72 even at c128. The reason is structural — the dispatch loop fires a due stream before its peers' messages are even ingested, so simultaneous readiness is never *observed* without a small accumulation window. A ready-depth signal alone won't fix this; the #1236 policy needs a load-scaled wait (even a tiny one) to let the ready set materialize.
4. **TTFA:** parity through c32; at c64 heavy-wait is actually *better* than serial (0.133 vs 0.188 s p50 — higher throughput drains the queue faster); at c128 every config is queueing-dominated (~0.4–0.5 s p50) since offered load is ~20× capacity — that regime needs admission control, not dispatch tuning.
5. The serial baseline is flat (~110–115) across the whole range — it's compute-bound, so the new sustained capacity ceiling is what moves: **~177 xrt staggered / ~220 aligned, i.e. +58…92% more concurrent streams per GPU at the vocoder stage.**

Together with the tables above, this completes the operating map the #1236 policy needs: wait/floor should scale from (0,1) at low load to (τ≈2–8 ms, floor 4–8) once the due-stream rate clears the crossover.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -467,8 +517,12 @@ def run_step(
             codes = torch.stack(rows, dim=0)
             wav, execution_metadata = self._forward_codes(
                 codes,
-                graph_eligible=False,
+                graph_eligible=self._chunk_aligned_dispatch,
```
edwingao28: If a sub-batch fails after earlier sub-batches in this loop succeeded, the exception propagates before `decoded` is returned, but the successful sub-batches already advanced `state.emitted` / `state.audio_parts` — and `on_step_failure` then aborts every participant, so healthy requests silently lose an already-decoded chunk from the live stream.

Suggest catching the failure inside the loop, aborting only the participants from the failing sub-batch onward, and still returning `decoded` from the completed ones. Please also add a test for the fail-after-success case too, current`test_step_failure_isolates_participants` only covers a first-sub-batch failure.

sglang_omni/models/qwen3_omni/components/code2wav_cuda_graph.py
```diff
@@ -496,6 +496,15 @@ def _disable_runtime(self, reason: str) -> None:
...
         logger.exception("Code2Wav CUDA graph replay disabled the runner")
 
+    @property
+    def enabled(self) -> bool:
+        return self._enabled
+
+    @property
+    def max_captured_batch_size(self) -> int:
+        """Largest published batch dimension; 0 once the runner is disabled."""
+        return max((key.batch_size for key in self._graphs), default=0)
```
edwingao28: nit: this rescans the graph keys on every access and `_chunk_aligned_dispatch` reads it several times per step — worth caching and invalidating on build/disable.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -31,6 +32,11 @@
 logger = logging.getLogger(__name__)
 
 
+# Sub-batch sizes emitted by ``Code2WavScheduler._decompose_batch``; the
+# batched graph key set and the batch-ceiling clamp must stay in sync with it.
```
edwingao28: please follow repo comment rules

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -50,6 +56,22 @@ def _serial_threshold_graph_keys(
...
 
 
+def _batched_graph_keys(
+    stream_chunk_size: int,
+    left_context_size: int,
+    batch_ceiling: int,
+) -> tuple[GraphKey, ...]:
+    # Chunk-aligned dispatch decomposes every group into ``_DECOMPOSE_SIZES``
+    # sub-batches over the same window shapes as serial decode, so batched
+    # graphs reuse the serial frame set with larger batch sizes.
```
edwingao28: please follow repo comment rules

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -88,7 +110,15 @@ class Code2WavStreamState:
...
-    """Streaming vocoder scheduler. Same inbox/outbox interface as OmniScheduler."""
+    """Streaming vocoder scheduler. Same inbox/outbox interface as OmniScheduler.
+
+    With batching and the CUDA graph runner both enabled, dispatch is
+    chunk-aligned: each step consumes exactly ``stream_chunk_size`` new
+    frames (a backlog drains over repeated uniform steps), so every window
+    stays inside the serial graph key set and every sub-batch shape stays
+    inside the captured batched key set; groups replay CUDA graphs and fall
+    back to one eager forward on a key miss.
+    """
```
edwingao28: Don't think we need that much comments to explain this class

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -330,14 +369,23 @@ def _pump_due_streams(self) -> None:
     def _ready(self, state: Code2WavStreamState) -> int:
         return len(state.chunks) - state.emitted
 
+    def _step_frames(self, state: Code2WavStreamState) -> int:
+        """New frames the next step consumes. Chunk-aligned dispatch caps each
+        step at one stream chunk so windows stay inside the serial graph key
+        set and buckets stay comparable regardless of per-request backlog."""
```
edwingao28: please trim this

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -137,6 +163,19 @@ def __init__(
         if self._enable_batching:
             self._stream_chunk_batch_max = self._batch_ceiling
 
+    @property
+    def _chunk_aligned_dispatch(self) -> bool:
+        """True while batched graph replay is actually available.
+
+        Reads the runner's live key set, so a build-time or runtime disable
+        stops the one-chunk-per-step cap and the 8/4/2/1
+        decomposition with it; fail-closed for runners without the property.
```
edwingao28: please trim this

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -505,7 +559,10 @@ def run_step(
                 metadata={
                     **profile_metadata,
                     "audio_samples": audio_samples,
+                    # Last sub-batch, kept for event-shape compatibility; the
+                    # per-sub-batch list is the authoritative record.
```
edwingao28: comment style issue

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -545,16 +600,44 @@ def create_code2wav_scheduler(
...
-                stream_chunk_size,
-                left_context_size,
-            ),
+            graph_keys=graph_keys,
         )
+        if len(graph_keys) > len(serial_keys) and not cuda_graph_runner.enabled:
+            # The runner build is all-or-nothing; don't let a failed batched
+            # capture take the serial graphs down with it. Measured on H100:
+            # every batching mode without batched graphs loses to plain
+            # serial graph replay, so drop batching with the failed capture.
```
edwingao28: comment style issue

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
dyyfk: Fixed in c1a8cfcc. The failure is now caught inside the sub-batch loop: only the participants from the failing sub-batch onward are aborted, and `decoded` from the completed sub-batches is still returned so their audio reaches the client. One wrinkle: the base contract runs abort callbacks off `_state_lock` and relied on `run_step` raising to surface the failed ids — since we no longer raise, the ids are staged in `_pending_step_failures` and appended to the `_pump_streams` return value, so the callbacks still run off the lock. Base class untouched.

Added `test_step_failure_after_success_keeps_decoded_sub_batches`: plan [2, 1] with the second forward failing — asserts the first sub-batch keeps its audio and stays alive, the failed participant gets an error + abort cleanup, and the survivor decodes normally afterwards.

sglang_omni/models/qwen3_omni/components/code2wav_cuda_graph.py
dyyfk: Done in c1a8cfcc — cached in a field updated at the three points the key set changes (publish, build rollback, runtime disable); the property now just returns it. Added `test_max_captured_batch_size_follows_publish_and_disable` pinning the three states (4 / 0 / 0).

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
dyyfk: Done in c1a8cfcc.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
dyyfk: Done in c1a8cfcc.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
dyyfk: Done in c1a8cfcc.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
dyyfk: Done in c1a8cfcc.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
dyyfk: Done in c1a8cfcc.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
dyyfk: Done in c1a8cfcc.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
dyyfk: Trimmed back to the original one-liner in c1a8cfcc; the dispatch mechanics now live in the (signed, why-only) comments at the sites they explain.

dyyfk:
**Integration update: merged the 1-pool tiered graph capture (@Peter-Nie2003) into this PR**

Head is now `db393398` = merge of `Peter-Nie2003:code2wav-batched-graph-capture` @ `319ff28b` into the previous head `78b6c270`, plus a small caching commit. Peter's commits are preserved as merge parents.

The two branches modified the same two runtime files; resolutions:

- **Graph runner**: Peter's 1-pool rework taken wholesale — tier-1 keys captured largest-first into one shared mempool, per-key budget checks measure the retained footprint after `empty_cache()` (warmup transients excluded), shrink drops the largest remaining batch class and retries the whole attempt, B1 safety preserved by capture ordering instead of a separate pool.
- **Interface**: the scheduler now asks `available_batch_sizes(frames)` (Peter's interface); `max_captured_batch_size` is gone. The caching request from the earlier review round carries over — published sizes are cached per frame count and refreshed at publish/rollback/runtime-disable.
- **Factory**: no more serial-only rebuild on batched-capture failure; the runner degrades internally, and the factory drops batching only if the runner comes up fully disabled (batching without graph replay measured worse than serial on H100).
- **Dispatch policy (changed vs BOTH parents, per the H100 policy table below)**: a window with no batched graph published runs as **one whole-batch eager forward** instead of being shattered into per-request B1 replays (B8 eager 28.0 ms vs 8×B1 replays 47.4 ms per 8 due windows); a lone request still replays its B1 graph. @Peter-Nie2003 — this rewrites the decompose-to-published-sizes semantics from your branch for the no-batched-size case; flag if you disagree.
- **Events**: `code2wav_batch_end` keeps the flat last-sub-batch fields and adds an aggregated `execution_mode: mixed` for heterogeneous plans. The `_run_sub_batch` failure isolation from the review round is unchanged.

**H100 validation (tag `graph-probe-20260811-092029`, H100 via Modal, torch 2.11; probe ran at merge commit `2d01cb57` — the caching commit on top is behavior-neutral for capture):**

- Merged runner built at prod `total_gpu_memory_fraction=0.02`: **12/16 published** — all of {B1,B2,B4} x {t10,t20,t30,t35}, retained pool 634 MB of the 1207.5 MB budget. The warmup-cache budget fix works: the B4 family is no longer falsely skipped (the 2-pool build published only {B1,B2} here).
- The B8 class is dropped **legitimately**: the runner measures b8t35's retained footprint at 1266 MB (> 1207.5 MB budget; same 1266 MB at 0.05, so not a transient), the shrink retries once and publishes the rest — graceful degradation working as designed (`attempts=2`, no disable).
- At 0.05 (eval): 16/16 published, single attempt, zero skipped keys.
- Policy table per 8 due windows (T=35): B8 graph 23.9 ms < B8 eager 28.0 < 2xB4 graph 29.6 < 4xB2 33.7 < 8xB1 replays 47.4. With B8 missing at 0.02 the dispatcher decomposes 8-groups to [4,4] (29.6 ms) — near-parity with whole-batch eager (28.0 ms).
- Fraction discussion: b8t35 needs ~1266 MB; at 0.025 the graph budget would be ~1691 MB, so the full 16/16 (and the 23.9 ms best case, +17% over eager) becomes available. Worth deciding whether prod moves to 0.025 or stays at 0.02 with the B4 ceiling.

Unit suite: 88 pass locally (8 failures are local-env only: venv torch 2.3.1 lacks `torch.OutOfMemoryError`; repo pins torch 2.11).

**Not yet claimed**: end-to-end numbers for this merged head. All prior ab2 results were measured at fraction 0.05 on the pre-merge head; a full ab2 A/B (serial-graph / fixed-batching / chunk-aligned, at both 0.02 and 0.05) on `db393398` is the next step, and the PR stays as-is until those numbers land.

dyyfk:
**E2E A/B on the merged head (`de091f47`): production config validated across c1-c64**

Per @edwingao28's request: re-benchmarked on the merged commit, table format follows #1101. Setup: H100 via Modal (torch 2.11), eval branch `240901fd` (= merged head + harness-only adaptation), staggered Poisson arrivals (frame-interval 5 ms, the talker-bound backlog regime from #1026), 20 windows/request, K=3 repeats per cell (means reported), arms `serial-graph` (production default) vs `quantized` (chunk-aligned dispatch; historical arm label). Tags: `ab3-confirm-*-20260811-100408`, `ab3-scale-*-20260811-100921`, `ab3-c1-*-20260811-105909`. Cross-fraction absolutes are not comparable (different H100 instances); all deltas are within-fraction.

## Production configuration (`total_gpu_memory_fraction=0.02` — runner publishes 12/16 graphs, B8 class over budget, shrink logged and served)

| concurrency | audio xRT serial -> chunk-aligned (zero-wait) | delta | xRT chunk-aligned (wait 8 ms, floor 8) | delta | TTFA p50 zero-wait | TTFA p50 heavy-wait |
| 1 | 14.5 -> 14.6 | ~0% | 13.2 | -9% | 0.060 -> 0.060 | 0.060 |
| 8 | 90.5 -> 100.2 | **+11%** | 96.1 | +6% | 0.061 -> 0.062 | 0.061 |
| 16 | 93.9 -> 107.6 | **+15%** | 141.7 | **+51%** | 0.064 -> 0.061 | 0.063 |
| 32 | 119.1 -> 134.5 | **+13%** | 168.7 | **+42%** | 0.079 -> 0.074 | 0.071 |
| 64 | 118.5 -> 132.5 | **+12%** | 167.1 | **+41%** | 0.136 -> 0.134 | 0.162 |

## Full graph set (fraction 0.05 — controlled comparison, 16/16 graphs; isolates what the missing B8 class costs and provides continuity with all previously posted numbers)

| concurrency | xRT serial -> chunk-aligned (zero-wait) | delta | xRT chunk-aligned (wait 8 ms, floor 8) | delta | TTFA p50 zero-wait | TTFA p50 heavy-wait |
| 1 | 14.5 -> 14.6 | ~0% | 13.5 | -7% | 0.060 -> 0.060 | 0.060 |
| 8 | 106.5 -> 106.8 | ~0% | 102.2 | -4% | 0.056 -> 0.056 | 0.056 |
| 16 | 116.9 -> 133.9 | **+15%** | 179.4 | **+53%** | 0.056 -> 0.054 | 0.054 |
| 32 | 113.7 -> 129.1 | **+14%** | 175.4 | **+54%** | 0.085 -> 0.073 | 0.074 |
| 64 | 112.5 -> 127.4 | **+13%** | 174.4 | **+55%** | 0.192 -> 0.141 | 0.141 |

## Findings

1. **Zero-wait at the production budget: +11-15% at every concurrency >=8, TTFA parity or better.** The missing B8 class is immaterial on this path — the 0.02 deltas match the full-graph-set deltas within noise.
2. **The B8 class only matters under heavy waiting at high concurrency.** There 8-wide groups actually form (single-forward share 0.33-0.47) and replaying them as 2xB4 costs ~12 points of the +54% (still +41-42%), with c64 TTFA slipping slightly below baseline (0.162 vs 0.136). We do **not** propose raising the default fraction to recover this: the vocoder's demand for batched graphs and the colocated talker's demand for KV cache both rise with concurrency, so the ~0.4 GB the B8 pool needs would be taken from talker KV in exactly the regime where KV is scarcest. When-not-to-use: leave the fraction at 0.02 unless a deployment is demonstrably vocoder-bound with measured slack on the talker GPU.
3. **No regression from the merge**: full-graph-set c32/64 matches the pre-merge scale sweep (comment 5188830616) within noise.
4. **Correctness**: waveform equivalence passes at both fractions (worst SNR 35.26 dB at 0.02, where `max_batch=4` shows the B4 decompose path is genuinely exercised; 35.37 dB at 0.05; zero tolerance failures). The 0.02 build logs `published 12/16` and serves — the graceful-degradation path measured here is exactly what production runs.
5. Fixed batching without graphs stays far below baseline (xRT 32-53): batching only pays with batched graph replay, unchanged from #1026.
6. Defaults are unchanged by this PR (batching off). The heavy-wait penalty at low concurrency (c1 -7-9%, c8 -4%) against its gain at c16+ (+41-55%) is the zero-wait/heavy-wait crossover — load-scaled waiting remains #1236's policy question.

Results JSONs with full provenance are retained; happy to attach any of them.

edwingao28:
@dyyfk The Modifications section still describes the earlier implementation: max_captured_batch_size, an all-or-nothing build, a serial-only factory retry, and only four CUDA-runner lines. The current head instead uses available_batch_sizes, tiered partial publication in one shared pool, and no factory rebuild. Please update the description and test summary so they match the behavior being merged.

dyyfk:
@edwingao28 Good catch — the description was stale. Updated: the Modifications section now describes the behavior at the current head — `available_batch_sizes(frames)` reporting the graphs actually published, tiered best-effort publication of the batch>1 keys in the one shared mempool (memory-pressure retries drop the largest remaining batch-size class; the batch-1 tier stays atomic; no factory rebuild), and chunk-aligned dispatch following that live per-frame availability. The test summary is refreshed to the current suite as well (97 passed, 2 skipped — including tiered capture degradation and shared-pool batch-1/batch-2 alternating replay coverage).

edwingao28:
# Benchmark evidence for the two commits

**Setup.** All numbers measured on this PR's branch (`e7e2ce2`, merged with latest main) on 2×H200 (thinker+code2wav on GPU0, talker on GPU1), TP1 per stage, Qwen3-Omni-30B-A3B-Instruct, SeedTTS EN voice-clone, streaming, temperature 0, `mem-fraction-static 0.70`. Long-utterance workload (concatenated SeedTTS texts, natural audio 20.9–22.4 s per request — code2wav under real load). Batching knobs pinned identically on both arms: `enable_batching=true, max_batch_wait_ms=8, batch_floor=4, batch_ceiling=8`. Warm round discarded after every boot; 2 runs per cell, best-of-2; zero failed requests in all cells; per-level audio duration matched across arms within 1.6%.

**Noise control.** The baseline arm was re-run at the end of the session (drift control). Per-metric noise band = max(|baseline self-drift|, 2%). Verdicts below are only claimed outside that band.

## Commit `[Perf] let first code2wav chunks jump the inbox queue`

Baseline (this PR) vs baseline + queue-jump, identical config:

| level | metric | base | +queue-jump | Δ | noise band | verdict |
|---|---|---|---|---|---|---|
| c32 | req/s | 2.905 | 2.957 | +1.8% | 4.0% | neutral |
| c32 | e2e mean (s) | 10.816 | 10.598 | −2.0% | 4.2% | neutral |
| c32 | e2e p95 (s) | 12.531 | 11.764 | −6.1% | 7.4% | neutral |
| c32 | TTFC mean (s) | 2.047 | 2.064 | +0.8% | 2.0% | neutral |
| c32 | TTFC p95 (s) | 3.411 | 3.268 | −4.2% | 13.1% | neutral |
| c32 | TTFT (s) | 0.802 | 0.826 | +3.0% | 9.8% | neutral |
| c32 | ITL (s) | 0.329 | 0.321 | −2.3% | 3.4% | neutral |
| c64 | req/s | 3.079 | 3.157 | **+2.5%** | 2.0% | **positive** |
| c64 | e2e mean (s) | 20.340 | 19.626 | **−3.5%** | 2.5% | **positive** |
| c64 | e2e p95 (s) | 25.506 | 23.255 | **−8.8%** | 7.7% | **positive** |
| c64 | TTFC mean (s) | 4.091 | 4.043 | −1.2% | 2.9% | neutral |
| c64 | TTFC p95 (s) | 7.576 | 7.196 | −5.0% | 5.2% | neutral |
| c64 | TTFT (s) | 1.688 | 1.634 | **−3.2%** | 3.0% | **positive** |
| c64 | ITL (s) | 0.634 | 0.608 | **−4.2%** | 3.5% | **positive** |

Short-utterance regression check (natural ~3.4 s audio, 3 replicates per arm, batching on and off): all metrics within the replicate noise band at c1–c64 — no regression on light code2wav load either.

## Commit `[Bugfix] honor initial_codec_chunk_frames in batched code2wav dispatch`

Functional fix: the chunk-aligned batched dispatch waits for a full steady chunk before the first decode, which silently ignores a configured `initial_codec_chunk_frames` (the smaller first chunk that upstream added for TTFA). This restores the knob under batched dispatch (clamped so the window never exceeds one steady chunk). Inert when the knob is unset — default behavior is byte-identical. Covered by 2 new unit tests; full suite failure set identical to the clean branch.

edwingao28 (approved):

Issue #1234, opened 2026-07-30
[TTS] Track streaming continuity under concurrency


luojiaxuan:
TTFA alone does not capture playback starvation after the first audio chunk. Track streaming continuity under load and keep TTFA, latency, and throughput regressions visible.

Metric

A true multi-chunk request passes when its maximum playback underrun is at most 100 ms. Single-chunk output is N/A, not a failure.

C16 baseline

Snapshot: b12005d7, full Seed-TTS EN/ZH/zh-hardcase, one repeat.

| Model | EN | ZH | Hard |
| --- | ---: | ---: | ---: |
| Fish S2-Pro | 94.57% | 80.79% | 21.65% |
| Higgs Audio v3 | 55.33% | 19.06% | 56.25% |
| MOSS-TTS Local v1.5 | 6.16% | 26.73% | 68.42% |

Qwen3-TTS Base and MOSS-TTS Delay currently return one PCM chunk, so continuity is N/A for those serving paths.

Work

- Tune each streaming model without hiding TTFA or throughput regressions.
- Investigate follow-up vocoder scheduling and chunk cadence, not only initial buffering.
- Validate EN, ZH, and hardcase at concurrency 8, 16, and 32.
- Report continuity plus TTFA/latency p50, p95, and p99.

Related: #1225

luojiaxuan:
Qwen3-TTS is fixed in #1240: true stream and C100 are 100% at C8/C16/C32, with a 19.2 KB first payload. Streaming QPS is 30-48% lower, so the tradeoff stays visible in the PR.

luojiaxuan:
Fish S2-Pro is fixed in #1241. Full 400-sample hardcase runs now complete 400/400 at C16 and C32 with C100 100%; max underrun p95/p99 are 0.

PR #1240, opened 2026-07-31
[TTS][Qwen3] Add true streaming vocoder

luojiaxuan:
Part of #1234. Addresses #1248.

Adds true incremental codec and vocoder streaming for Base and reference cloning. Initial audio uses pre-captured decoder graphs. Reference codec extraction is batched and overlapped with speaker embedding; follow-ups use a lower-priority stream with an 8-frame stride.

H100 Seed-TTS EN full set, matched config and hardware, 1088 requests per setup.

| Concurrency | Current TTFA p95 | Baseline | Current C200 | Baseline C200 | Current/Baseline QPS | Current/Baseline EN WER |
| 8 | 0.266s | 0.330s | 84.74% | 0% | 6.114 / 2.921 | 1.06% / 1.07% |
| 16 | 0.320s | 0.497s | 12.41% | 0% | 7.746 / 3.107 | 1.06% / 1.06% |
| 32 | 0.541s | 0.895s | 3.03% | 0% | 9.612 / 3.155 | 1.05% / 1.06% |

All performance runs completed 1088/1088. First payload p95 is 3.75 KiB. C8 WER is from the new full run; C16/C32 use the existing full correctness matrix. This latency patch does not change talker sampling.

Tests: 112 passed. Ruff and Black passed.


sglang_omni/models/qwen3_tts/request_builders.py
```diff
@@ -582,9 +590,21 @@ class _Qwen3TTSAdhocReferenceHook(
...
+    def _encode_context(self):
+        device = torch.device(self._model.device)
+        if device.type != "cuda":
+            return nullcontext()
+        stream = getattr(self._stream_local, "stream", None)
+        if stream is None:
+            least_priority, _ = torch.cuda.Stream.priority_range()
+            stream = torch.cuda.Stream(device=device, priority=least_priority)
+            self._stream_local.stream = stream
+        return torch.cuda.stream(stream)
```
SandyLuXY, line 715: Exiting this context does not synchronize the producer stream, so `store_artifact()` can copy incomplete reference tensors to CPU; a delayed-fill repro caches `0.0` before the producer finishes with `7.0`. Add an event/`wait_stream` handoff, or synchronize before returning.

sglang_omni/models/qwen3_tts/streaming_vocoder.py
```diff
@@ -0,0 +1,891 @@
...
+                self._stream_followup_stride,
+            )
+            if stream_initial_followup_stride is None
+            else stream_initial_followup_stride
+        )
+        self._initial_max_batch_size = int(initial_max_batch_size)
+        self._initial_batch_wait_s = float(initial_batch_wait_ms) / 1000.0
+        self._followup_max_batch_size = int(followup_max_batch_size)
+        self._followup_batch_wait_s = float(followup_batch_wait_ms) / 1000.0
+        self._default_initial_chunk_frames = int(initial_chunk_frames)
```
SandyLuXY, line 211: Now `initial_chunk_frames=32, stream_stride=16` decodes at 16 frames when the payload arrives first but waits for 32 when the chunk arrives first, so TTFA depends on message ordering.

Prasannajaga:
hi @luojiaxuan I tested this PR locally with `Qwen3-TTS-12Hz-0.6B-Base` on an RTX 4050 (SM89) , using CUDA 13.0, PyTorch 2.11, and SGLang 0.5.12.post1.  With `SGLANG_OMNI_QTTS_NATIVE_ACTIVATION=0`, the server fails while JIT-compiling SGLang’s `activation.cuh`: 
 ```text
no matching function for call to select_kernel<true>
expected a type, got ActivationKind::kSiLU
```
With `SGLANG_OMNI_QTTS_NATIVE_ACTIVATION=1`, the server works and returns streaming PCM audio, but it uses the native PyTorch activation fallback. Did your H100 benchmark actually compile and use the fused `silu_and_mul` kernel? Could you share the exact CUDA, PyTorch, SGLang, and `sgl-kernel` versions, and confirm whether the JIT cache was clean? I would like to understand whether this is an SM89 compatibility issue or an environment/kernel-version difference.

NOTE: I tested this with my changes!

**Comment by Hayden727** [comment]
I narrowed and rebased the true-streaming work into #1286. The replacement keeps the codec-to-vocoder streaming path and its H100 benchmark provenance, while removing the unrelated 1.7B low-VRAM profile, preprocessing concurrency, max-running-request change, and reference-stream experiment. It also adds message-order, CUDA stream handoff/device-retention, model-mode gating, fallback-parity, and lifecycle regression coverage. The 1.7B consumer-GPU profile will be validated separately under #1120.

File: sglang_omni/models/qwen3_tts/request_builders.py (hunk 1 above)
Comment (luojiaxuan) [reply sglang_omni/models/qwen3_tts/request_builders.py:715]: Superseded by #1286, which removes this reference stream. The TTFA follow-up synchronizes batched CUDA results before caching.

File: sglang_omni/models/qwen3_tts/streaming_vocoder.py (hunk 2 above)
Comment (luojiaxuan) [reply sglang_omni/models/qwen3_tts/streaming_vocoder.py:211]: Superseded by #1286, which makes initial-chunk resolution order-independent and adds regression coverage.

File: sglang_omni/models/qwen3_tts/streaming_vocoder.py [hunk 3]
```diff
@@ -0,0 +1,990 @@
...
+        for batch_size in self._batch_sizes:
+            try:
+                static_input = torch.zeros(
+                    (batch_size, self._num_quantizers, self._input_frames),
+                    dtype=torch.long,
+                    device=self._device,
+                )
+                with torch.inference_mode(), torch.cuda.stream(capture_stream):
+                    for _ in range(2):
+                        self._decoder(static_input)
```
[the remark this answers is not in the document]
luojiaxuan: Moved to the TTFA follow-up, where the fake decoder implements the production callable contract.

sglang_omni/models/qwen3_tts/streaming_vocoder.py
```diff
@@ -0,0 +1,990 @@
...
+                graph = torch.cuda.CUDAGraph()
+                with (
+                    torch.inference_mode(),
+                    torch.cuda.graph(
+                        graph,
+                        pool=graph_pool,
+                        stream=capture_stream,
+                    ),
+                ):
+                    static_output = self._decoder(static_input)
```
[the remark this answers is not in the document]
luojiaxuan: Same fix in the TTFA follow-up.

luojiaxuan:
Superseded by #1286. TTFA-only changes are preserved on https://github.com/sgl-project/sglang-omni/tree/luojiaxuan/qwen3-tts-ttfa-optimization and will be opened after #1286 lands.

PR #1242, opened 2026-07-31
[Fix] Add MOSS preprocessing GPU budget

MayDomine:
## Summary
- add the missing `0.05` preprocessing GPU budget to the MOSS-TTS process-edge resource contract
- keep generic topology validation strict
- update the isolation regression test for MOSS's complete `1.0` GPU budget

Fixes the integration regression between #1125 and #1222.

## Test plan
- [x] `pre-commit run --files sglang_omni/config/topology.py sglang_omni/models/moss_tts/config.py tests/unit_test/pipeline/test_topology.py`
- [ ] PR unit-test CI

### DIFF
--- a/sglang_omni/config/topology.py
+++ b/sglang_omni/config/topology.py
@@ -234,6 +234,8 @@ def _validate_gpu_process_colocation(
 ) -> None:
     stage_by_name = {stage.name: stage for stage in stages}
     gpu_processes: dict[int, set[str]] = defaultdict(set)
+    process_stages: dict[tuple[int, str], set[str]] = defaultdict(set)
+    processes_with_fraction: set[tuple[int, str]] = set()
     missing_fraction: dict[int, set[str]] = defaultdict(set)
 
     for group in topology_plan.groups:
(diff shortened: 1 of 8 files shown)


MayDomine:
This fixes the integration issue between #1125 and #1222. @AkazaAkane @CloudRipple, could you please take a look at the process-level GPU memory budget semantics? GitHub would not allow formally requesting you as reviewers because the repository reports read-only collaborator permissions.

CloudRipple:
> This fixes the integration issue between #1125 and #1222. @AkazaAkane @CloudRipple, could you please take a look at the process-level GPU memory budget semantics? GitHub would not allow formally requesting you as reviewers because the repository reports read-only collaborator permissions.

@MayDomine  Hi, thanks for catching this issue.

The implementation looks reasonable under process-level budget semantics. However, since this regression was caused by #1222 making the MOSS-TTS preprocessing stage GPU-resident without updating its resource contract, would it be better for me to add the missing model-specific GPU budget to MOSS-TTS instead, following MOSS-TTS Local, rather than changing the generic validation semantics?

MayDomine:
Thanks, agreed. I narrowed the fix to the MOSS-TTS resource contract: preprocessing now gets `0.05`, while the generic topology validation remains unchanged. The isolated MOSS topology now declares `0.05 + 0.85 + 0.10 = 1.0`.

Ratish1 (approved):
LGTM

Issue #1196, opened 2026-07-27
[Feature] Enable FishAudio S2-Pro on RTX 5090 (SM120)


Dayuxiaoshui:
Related to [#1120](https://github.com/sgl-project/sglang-omni/issues/1120).

## Checklist

- [x] This is a feature request rather than a usage question.
- [x] This report is written in English.

## Motivation

FishAudio S2-Pro previously used FA3 for both of its attention paths. Slow-AR defaulted to FA3, while Fast-AR unconditionally called `sgl_kernel.flash_attn.flash_attn_with_kvcache`. The Fast-AR call enters an SM90-oriented KV-cache kernel path and prevents the model from running on consumer GPUs such as the RTX 4090 (SM89) and RTX 5090 (SM120).

S2-Pro needs an explicit architecture policy for both paths so that it can use a validated attention implementation on each supported compute capability.An explicit Slow-AR `attention_backend` override must continue to take priority.

The expected policy is:

| Compute capability | Slow-AR | Fast-AR KV cache |
|---|---|---|
| SM89 | FlashInfer | FlashInfer |
| SM90 | FA3 | FA3 |
| SM100 | FlashInfer | FlashInfer |
| SM120 | FlashInfer | FlashInfer |

Architectures without a defined policy should fail with a clear, actionable error instead of silently selecting an unvalidated kernel.

## Reproduced SM120 Failures

The following failures were reproduced on an RTX 5090 before applying the compatibility change:

1. Default startup failed during CUDA Graph capture when Fast-AR reached the Hopper-oriented kernel on SM120, with `no kernel image is available for execution on the device`.
2. Explicit FA4 selection failed with `Paged KV not supported on SM 12.0 in this PR`.
3. Slow-AR could start with FlashInfer or Triton in eager mode, but the first request still failed because Fast-AR unconditionally called the FA3 KV-cache path.

These failures block the S2-Pro SM120 roadmap item in #1120 before quality or performance evaluation can begin.

## Proposed Scope

- Select the Slow-AR backend before constructing `ServerArgs`: FA3 on SM90 and FlashInfer on SM89, SM100, and SM120.
- Preserve an explicit Slow-AR `attention_backend` override.
- Keep the existing FA3 Fast-AR KV-cache path on SM90 and use FlashInfer on SM89, SM100, and SM120.
- Preserve the dense NHD KV-cache layout and grouped-query attention shape.
[...]

PR #1243, opened 2026-07-31
[Feature] ：support FishAudio S2-Pro on consumer GPUs

Dayuxiaoshui:
## Motivation

FishAudio S2-Pro used FA3 for both of its attention paths. Its Fast-AR path called an SM90-oriented KV-cache kernel unconditionally, which prevents the model from running on consumer GPUs such as RTX 4090 (SM89) and RTX 5090
(SM120).

This PR makes both attention paths architecture-aware while keeping the change inside the existing S2-Pro builder and decoder. Explicit Slow-AR backend overrides continue to take priority.

## Modifications

- Select the Slow-AR backend before constructing `ServerArgs`: FA3 on SM90 and FlashInfer on SM89, SM100, and SM120.
- Keep an explicit `attention_backend` override unchanged.
- Use the existing FA3 Fast-AR KV-cache path on SM90 and a FlashInfer fallback on SM89, SM100, and SM120.
- Preserve the dense NHD KV-cache layout and grouped-query attention shape, and pass the batch-uniform cache position through the decoder layers so CUDA
  Graph capture does not require a device-to-host conversion.
- Reject unsupported compute capabilities with an actionable error instead of silently selecting an unvalidated kernel.
- Add focused coverage for backend selection, explicit overrides, Fast-AR dispatch, unsupported architectures, and complete KV-cache update semantics.

## Related Issues

Related to #1196.

Related to #1120.

## Accuracy Test

The final patch was tested with the real `fishaudio/s2-pro` model at revision `1de9996b6be38b745688de084d87a5633f714e4e` in BF16 on both consumer GPU architectures.

### RTX 4090 D (SM89)

| Test | Result |
|---|---|
| Non-streaming smoke | HTTP 200; valid 44.1 kHz, mono, 16-bit WAV; 4.226 s audio |
| Streaming smoke | HTTP 200; valid 44.1 kHz, mono, signed 16-bit PCM; 3.390 s audio |
| Concurrency 4, two rounds | 8/8 HTTP 200; every response was a valid WAV |
| Post-test health check | Healthy; all three pipeline stages running |

The complete S2-Pro unit-test directory passed:

```text
92 passed, 1 skipped
```

### RTX 5090 (SM120)

The SM120 rerun used SGLang 0.5.16, PyTorch 2.11.0+cu130, and one RTX 5090 with driver 610.43.02.

| Test | Result |
|---|---|
| Non-streaming smoke | HTTP 200; valid 44.1 kHz, mono, 16-bit WAV; 1.672 s audio |
| Streaming smoke | HTTP 200; valid 44.1 kHz, mono, signed 16-bit PCM; 1.300 s audio |
| Concurrency 4, two rounds | 8/8 HTTP 200; every response was a valid WAV |
| Post-test health check | HTTP 200; all three pipeline stages running |

The complete S2-Pro unit-test directory passed on the current branch:

```text
94 passed
```

These reruns are functional model tests, not formal SeedTTS WER/CER or speaker similarity evaluations. The change selects a compatible attention implementation and does not intentionally change model weights, sampling parameters, or decoding policy. The requests were not seeded, so generated audio duration is not expected to match across runs or architectures.

### RTX 4090 D (SM89)

Rerun configuration:

```text
GPU: RTX 4090 D (SM89, 24,564 MiB)
Slow-AR: FlashInfer (automatically selected)
Fast-AR: FlashInfer (automatically selected)
max_total_tokens: 20,000
max_running_requests: 4
CUDA Graph batch sizes: [1, 2, 4]
torch.compile: disabled
```

CUDA Graph capture completed for batch sizes 1, 2, and 4 in 4.08 seconds.Server logs reported `cuda graph: True` during decode, including batch size 4.

| Workload | Result | Latency |
|---|---:|---:|
| Non-streaming smoke | 1/1 | 2.149 s |
| Streaming smoke | 1/1 | 1.744 s first byte, 1.764 s total |
| Concurrency 4, round 1 | 4/4 | 2.745 s wall time |
| Concurrency 4, round 2 | 4/4 | 2.705 s wall time |

GPU memory was 17,678 MiB after startup and 18,096 MiB after the test matrix.The service remained healthy with no worker failure or CUDA kernel error.

### RTX 5090 (SM120)

Rerun configuration:

```text
GPU: RTX 5090 (SM120, 32,607 MiB)
Slow-AR: FlashInfer (automatically selected)
Fast-AR: FlashInfer (automatically selected)
max_total_tokens: 20,000
max_running_requests: 4
CUDA Graph batch sizes: [1, 2, 4]
torch.compile: disabled
```

CUDA Graph capture completed for batch sizes 1, 2, and 4 in 18.69 seconds.Server logs reported `cuda graph: True` during decode, including batch size 4.

| Workload | Result | Latency |
|---|---:|---:|
| Non-streaming smoke | 1/1 | 6.071 s |
| Streaming smoke | 1/1 | 0.571 s first byte, 0.572 s total |
| Concurrency 4, round 1 | 4/4 | 1.002 s wall time |
| Concurrency 4, round 2 | 4/4 | 1.056 s wall time |

GPU memory was 17,932 MiB after the test matrix. The service remained healthy with no worker failure or CUDA kernel error. These are single-host compatibility measurements, not a controlled comparison against SM90 or between the two consumer GPUs. The RTX 5090 non-streaming measurement was the first request after startup and includes first-request warmup.

## Checklist

- [x] Add unit tests.

## CI


Hayden727 (changes requested):
The changed dispatch/KV-cache path looks correct, and the SM89/SM120 evidence in #1196 covers the risky GPU branches. One important docs issue remains before merge:

- `sglang_omni/models/fishaudio_s2_pro/README.md:48,54,78-80` still says S2-Pro forces FA3, that Slow-AR CUDA Graph capture is not implemented, and frames FlashInfer only as a divergence hazard. That now contradicts this PR's policy and validation. On SM89/SM120, following the README and forcing FA3 reintroduces the unsupported-kernel failure this PR fixes.

Please update the model README in this PR: document SM90 → FA3 and SM89/SM100/SM120 → FlashInfer, clarify that explicit `attention_backend` overrides only Slow-AR while Fast-AR remains architecture-selected, describe the validated decode CUDA Graph scope (batches 1/2/4; `torch.compile` still unvalidated), and keep the early-EOS note explicitly historical/quality-unqualified.

Local review verification: `tests/unit_test/fishaudio_s2_pro/test_pipeline.py` passes 54/54 against the PR head. The complete directory cannot collect in this macOS Python 3.13 venv because current `torch_profiler.py` resolves `profile` as a function in a union annotation; that is unrelated to this diff.

sglang_omni/models/fishaudio_s2_pro/fish_speech/models/text2semantic/audio_decoder.py
```diff
@@ -29,6 +29,58 @@
...
+        raise ValueError("Fast-AR q, k, and v batch sizes must match")
+
+    if cache_position < 0:
+        raise ValueError("FlashInfer Fast-AR attention requires cache_position")
+
+    cache_end = cache_position + int(k.shape[1])
+    k_cache[:, cache_position:cache_end].copy_(k)
+    v_cache[:, cache_position:cache_end].copy_(v)
+
+    import flashinfer
```
Hayden727, line 68: Fast-AR's architecture and dependency requirements need to be validated during startup rather than on this first-request path. An explicit Slow-AR `attention_backend` makes `adjust_overrides()` return before the SM allowlist and `is_flashinfer_available()` checks, so SM80 or SM89 without FlashInfer currently boots and fails only here during live decode (I reproduced both cases). Please run an unconditional Fast-AR preflight before honoring the Slow-AR override and add production-builder regressions for unsupported SM and missing/disabled FlashInfer.

sglang_omni/models/fishaudio_s2_pro/README.md
```diff
@@ -44,14 +44,27 @@ By integrating S2's Dual-AR backbone into SGLang's paged-attention engine, we in
...
-- **CUDA Graph dual-cover of Slow AR and Fast AR** — The 9-step codebook loop is covered with CUDA graphs. Details at [Revisiting CUDA Graph: Core Mechanisms, Multi-Graph Memory Sharing, and Unified Coverage for Dual AR Models](https://github.com/zhaochenyang20/Awesome-ML-SYS-Tutorial/blob/main/torch/cuda-graph/readme-2-en.md).
-- **FlashAttention 3** — Forced FA3 backend to match training-time attention numerics, avoiding early-EOS divergence from flashinfer.
+- **Decode CUDA Graph coverage of Slow AR and Fast AR** — Bounded decode graph capture and replay have been validated on SM89 and SM120 for batch sizes 1, 2, and 4 with `torch.compile` disabled. Prefill graphs, larger batches, and CUDA Graph execution with `torch.compile` remain unvalidated. Details on the dual-AR graph design are available in [Revisiting CUDA Graph: Core Mechanisms, Multi-Graph Memory Sharing, and Unified Coverage for Dual AR Models](https://github.com/zhaochenyang20/Awesome-ML-SYS-Tutorial/blob/main/torch/cuda-graph/readme-2-en.md).
```
Hayden727: This documented unvalidated combination is still the production default: `generation_defaults()` sets `enable_torch_compile=True`, `compile_model()` stores compiled Fast-AR layer callables, and clearing the server flag only disables SGLang's generic compile path; CUDA Graph capture/replay still calls the compiled layers. The hardware evidence disables `torch.compile`, and the unit test only verifies flag plumbing, not compiled decode. Please either validate compile+graph replay on the SM89/SM120 FlashInfer path or default the consumer Fast-AR path to compile off until that validation exists.

Hayden727 (changes requested):
The previous README contradiction is fixed on `eae964c7`, but two production-path gaps remain in Fast-AR preflight and the default compile/graph configuration; details inline. Local validation: 96 passed/1 skipped across the pipeline, TTS, and streaming-vocoder tests, plus the vocoder test passed with a Python 3.13 profiler-import stub; 15 focused Fast-AR/backend tests passed. The full directory otherwise cannot collect in this venv because `torch_profiler.py` resolves `profile` as a function in a union annotation, unrelated to this diff. I also reproduced explicit Slow-AR overrides allowing unsupported SM80 and missing-FlashInfer SM89 startup before first-request failure.

sglang_omni/models/fishaudio_s2_pro/fish_speech/models/text2semantic/audio_decoder.py
Hayden727: Validation on PR head eae964c7: moss-omni (RTX 4090 D, SM89) exercised the real FlashInfer custom-op path, and hyper01-omni (H200, SM90) exercised the FA3 path. For batch sizes 1/2/4 across 10 cache positions, both matched the reference attention implementation with max absolute error 0.007812. This confirms the kernel paths work when their prerequisites are present, but it does not address the startup gap: an explicit Slow-AR attention_backend still bypasses the Fast-AR architecture/dependency preflight. Keeping this unresolved until the preflight is unconditional and the production-builder regressions reject unsupported SM and unavailable FlashInfer.

sglang_omni/models/fishaudio_s2_pro/README.md
Hayden727: Additional validation on PR head eae964c7: the full FishAudio S2-Pro unit suite passed on both hosts (98 passed each). Compile plus CUDA Graph capture/replay also passed on SM89 and SM90; the real configured Fast-AR shape passed batch sizes 1/2/4 on SM89. However, with real-shape random weights, small eager/compiled BF16 differences at batch 4 flipped an intermediate codebook argmax and autoregressively amplified the final logits difference to 3.03. This is not evidence that trained S2-Pro weights fail, but it shows that structural capture/replay checks alone are insufficient. A trained-checkpoint run could not be completed because the hosts lacked the 4.14 GB decoder shard and codec weights, and no SM120 host was available. Please validate the default compiled+graphed path with trained weights and output-quality checks on the claimed consumer targets, or default consumer Fast-AR compile off until that evidence exists.

Hayden727 (approved):

PR #1263, opened 2026-08-02
[Cosmos01] text ar UND

JingwenGu0829:
## Scope

- Add the text-only `preprocessing -> thinker -> decode` vertical slice.
- Reuse SGLang's Qwen3 AR model, attention kernels, KV cache, sampling, and TP.
- Keep media encoders, UND decoder/encoder outputs, GEN, and cross-stage KV transfer
  out of this PR.

## Implementation

- Load the official `Cosmos3OmniConfig`, tokenizer, chat template, and Nano
  generation defaults.
- Convert chat or pretokenized input into a process-safe pipeline payload.
- Map unified Cosmos3 UND/text weights into SGLang's packed Qwen3 parameters and
  filter GEN/vision tensors.
- Resolve the checkpoint's nested `transformer/` component separately from the
  root config/tokenizer directory.
- Enable SGLang decode CUDA graph by default with explicit batch buckets tied
  to `max_running_requests`; the default BS=16 policy captures
  `[1, 2, 4, 8, 12, 16]`.
- Support streaming and non-streaming detokenization with usage and finish
  metadata.

## Correctness fixes found while bringing up the model

- Text mRoPE is explicitly `[3, sequence]`: temporal, height, and width all use
  the same monotonically increasing text positions. Do not route it as a plain
  one-dimensional RoPE input.
- OpenAI transport sampling placeholders must not override Cosmos3-Nano's model
  defaults unless the caller explicitly supplied them.
- The unified checkpoint stores model shards under `transformer/`; SGLang's
  default root-level weight glob cannot discover them without a component path.
- Keep the root config/tokenizer path separate from the nested transformer
  weight path; pointing tokenizer loading at `transformer/` breaks startup.
- The Docker image may load CUDA 13's `compat/libcuda.so.580` ahead of the host
  `libcuda.so.595`, causing CUDA error 803. Put the host driver library first.

## Validation

- The real Nano checkpoint loads all seven AR shards and serves the three-stage
  pipeline on one H200 (15.41 GB model weights).
- Decode CUDA graph captured BS `[1, 2, 4, 8, 12, 16]` in 3.16 seconds with
  about 0.02 GB additional memory; a forced 128-token decode replayed with
  `cuda graph: True`.
- Four greedy prompts matched Hugging Face eager generation exactly across all
  28 generated token IDs. Selected-token logprob difference was at most 0.0473
  and averaged 0.00545 on the 22-token case (SGLang FA3 vs HF eager).
- Upstream SGLang Cosmos3 and SRT produced identical text mRoPE caches across
  257 positions (`max_abs = 0.0`).
- Non-streaming `/generate` logprobs and OpenAI SSE streaming both completed
  end to end with correct token usage and stop metadata.
- `315` unit and related pipeline/OpenAI regressions cover component-path
  propagation, root tokenizer loading, preprocessing, weight mapping, and
  detokenization.

### DIFF
--- a/sglang_omni/models/cosmos3/bootstrap.py
+++ b/sglang_omni/models/cosmos3/bootstrap.py
@@ -3,7 +3,50 @@
 
 from __future__ import annotations
 
+from pathlib import Path
 from typing import Any
+
+from huggingface_hub import snapshot_download
+
+
+def _has_transformer_weights(path: Path | None) -> bool:
+    return path is not None and path.is_dir() and any(path.glob("*.safetensors"))
+
+
+def resolve_transformer_weights_path(model_path: str) -> str:
+    """Resolve the unified checkpoint's nested transformer component."""
+
+    local_path = Path(model_path)
+    if local_path.exists():
+        transformer_path = local_path / "transformer"
+    else:
+        allow_patterns = [
+            "transformer/*.json",
+            "transformer/*.safetensors",
+        ]
+        try:
+            snapshot_path = snapshot_download(
+                model_path,
+                allow_patterns=allow_patterns,
+                local_files_only=True,
+            )
+        except FileNotFoundError:
+            snapshot_path = None
+        transformer_path = (
+            Path(snapshot_path) / "transformer" if snapshot_path is not None else None
+        )
+        if not _has_transformer_weights(transformer_path):
+            snapshot_path = snapshot_download(
+                model_path,
+                allow_patterns=allow_patterns,
+            )
+            transformer_path = Path(snapshot_path) / "transformer"
+
+    if transformer_path is None or not _has_transformer_weights(transformer_path):
+        raise FileNotFoundError(
+            f"Cosmos3 transformer weights not found under {model_path!r}"
+        )
+    return str(transformer_path)
 
 
 def create_thinker_scheduler(
@@ -43,13 +86,14 @@ def create_thinker_scheduler(
         tp_rank=tp_rank,
         nccl_port=nccl_port,
         model_arch_override="Cosmos3TextForCausalLM",
+        model_weights_path=resolve_transformer_weights_path(server_args.model_path),
         total_gpu_memory_fraction=total_gpu_memory_fraction,
     )
 
     output_processor = SGLangOutputProcessor(capture_hidden=False)
     model_runner = ModelRunner(model_worker, output_processor)
     tokenizer = get_tokenizer(
-        model_config.model_path,
+        server_args.model_path,
         trust_remote_code=True,
     )
     request_builder, result_adapter = make_text_scheduler_adapters(
@@ -76,4 +120,4 @@ def create_thinker_scheduler(
     )
 
 
-__all__ = ["create_thinker_scheduler"]
+__all__ = ["create_thinker_scheduler", "resolve_transformer_weights_path"]
--- a/sglang_omni/models/cosmos3/bootstrap.py
+++ b/sglang_omni/models/cosmos3/bootstrap.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 from typing import Any
 
 
-def create_text_scheduler(
+def create_thinker_scheduler(
     server_args: Any,
     gpu_id: int = 0,
     *,
@@ -16,7 +16,7 @@ def create_text_scheduler(
     enable_async_decode: bool = False,
     async_decode_min_batch_size: int = 2,
 ):
-    """Create the SGLang-backed Cosmos3 text AR scheduler."""
+    """Create the SGLang-backed Cosmos3 thinker OmniScheduler."""
 
     from sglang.srt.utils.hf_transformers_utils import get_tokenizer
 
@@ -76,4 +76,4 @@ def create_text_scheduler(
     )
 
 
-__all__ = ["create_text_scheduler"]
+__all__ = ["create_thinker_scheduler"]
--- a/sglang_omni/models/cosmos3/bootstrap.py
+++ b/sglang_omni/models/cosmos3/bootstrap.py
@@ -1,0 +1,79 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Cosmos3 text scheduler construction."""
+
+from __future__ import annotations
+
+from typing import Any
+
+
+def create_text_scheduler(
+    server_args: Any,
+    gpu_id: int = 0,
+    *,
+    tp_rank: int = 0,
+    nccl_port: int | None = None,
+    total_gpu_memory_fraction: float | None = None,
+    enable_async_decode: bool = False,
+    async_decode_min_batch_size: int = 2,
+):
+    """Create the SGLang-backed Cosmos3 text AR scheduler."""
+
+    from sglang.srt.utils.hf_transformers_utils import get_tokenizer
+
+    from sglang_omni.model_runner.base import ModelRunner
+    from sglang_omni.models.cosmos3.request_builders import (
+        make_text_scheduler_adapters,
+        make_text_stream_output_builder,
+    )
+    from sglang_omni.scheduling.bootstrap import create_sglang_infrastructure
+    from sglang_omni.scheduling.omni_scheduler import OmniScheduler
+    from sglang_omni.scheduling.sglang_backend import SGLangOutputProcessor
+
+    (
+        model_worker,
+        tree_cache,
+        req_to_token_pool,
+        token_to_kv_pool_allocator,
+        prefill_manager,
+        decode_manager,
+        model_config,
+    ) = create_sglang_infrastructure(
+        server_args,
+        gpu_id,
+        tp_rank=tp_rank,
+        nccl_port=nccl_port,
+        model_arch_override="Cosmos3TextForCausalLM",
+        total_gpu_memory_fraction=total_gpu_memory_fraction,
+    )
+
+    output_processor = SGLangOutputProcessor(capture_hidden=False)
+    model_runner = ModelRunner(model_worker, output_processor)
+    tokenizer = get_tokenizer(
+        model_config.model_path,
+        trust_remote_code=True,
+    )
+    request_builder, result_adapter = make_text_scheduler_adapters(
+        tokenizer=tokenizer,
+        vocab_size=model_config.vocab_size,
+        generation_config=model_config.hf_generation_config,
+    )
+
+    return OmniScheduler(
+        tp_worker=model_worker,
+        tree_cache=tree_cache,
+        req_to_token_pool=req_to_token_pool,
+        token_to_kv_pool_allocator=token_to_kv_pool_allocator,
+        server_args=server_args,
+        model_config=model_config,
+        prefill_manager=prefill_manager,
+        decode_manager=decode_manager,
+        model_runner=model_runner,
+        request_builder=request_builder,
+        result_adapter=result_adapter,
+        stream_output_builder=make_text_stream_output_builder(),
+        enable_async_decode=enable_async_decode,
+        async_decode_min_batch_size=async_decode_min_batch_size,
+    )
+
+
+__all__ = ["create_text_scheduler"]
--- a/sglang_omni/models/cosmos3/bootstrap.py
+++ b/sglang_omni/models/cosmos3/bootstrap.py
@@ -13,7 +13,10 @@ def _has_transformer_weights(path: Path | None) -> bool:
     return path is not None and path.is_dir() and any(path.glob("*.safetensors"))
 
 
-def resolve_transformer_weights_path(model_path: str) -> str:
+def resolve_transformer_weights_path(
+    model_path: str,
+    revision: str | None = None,
+) -> str:
     """Resolve the unified checkpoint's nested transformer component."""
 
     local_path = Path(model_path)
@@ -27,6 +30,7 @@ def resolve_transformer_weights_path(model_path: str) -> str:
         try:
             snapshot_path = snapshot_download(
                 model_path,
+                revision=revision,
                 allow_patterns=allow_patterns,
                 local_files_only=True,
             )
@@ -38,6 +42,7 @@ def resolve_transformer_weights_path(model_path: str) -> str:
         if not _has_transformer_weights(transformer_path):
             snapshot_path = snapshot_download(
                 model_path,
+                revision=revision,
                 allow_patterns=allow_patterns,
             )
             transformer_path = Path(snapshot_path) / "transformer"
@@ -86,7 +91,10 @@ def create_thinker_scheduler(
         tp_rank=tp_rank,
         nccl_port=nccl_port,
         model_arch_override="Cosmos3TextForCausalLM",
-        model_weights_path=resolve_transformer_weights_path(server_args.model_path),
+        model_weights_path=resolve_transformer_weights_path(
+            server_args.model_path,
+            revision=server_args.revision,
+        ),
         total_gpu_memory_fraction=total_gpu_memory_fraction,
     )
 
@@ -95,6 +103,7 @@ def create_thinker_scheduler(
     tokenizer = get_tokenizer(
         server_args.model_path,
         trust_remote_code=True,
+        tokenizer_revision=server_args.revision,
     )
     request_builder, result_adapter = make_text_scheduler_adapters(
         tokenizer=tokenizer,
--- a/sglang_omni/models/cosmos3/bootstrap.py
+++ b/sglang_omni/models/cosmos3/bootstrap.py
@@ -54,6 +54,14 @@ def resolve_transformer_weights_path(
     return str(transformer_path)
 
 
+def _refuse_weight_update(payload: dict[str, Any]) -> tuple[bool, str]:
+    del payload
+    return False, (
+        "Cosmos3 does not support weight updates; weights load only at "
+        "startup from the nested transformer/ checkpoint component"
+    )
+
+
 def create_thinker_scheduler(
     server_args: Any,
     gpu_id: int = 0,
@@ -97,6 +105,17 @@ def create_thinker_scheduler(
         ),
         total_gpu_memory_fraction=total_gpu_memory_fraction,
     )
+
+    # A weight update would resolve the advertised model root instead of the
+    # nested transformer/ component, so refuse every update path (PR #1263).
+    for admin_method in (
+        "update_weights_from_disk",
+        "update_weights_from_tensor",
+        "update_weights_from_distributed",
+        "init_weights_update_group",
+        "destroy_weights_update_group",
+    ):
+        setattr(model_worker, admin_method, _refuse_weight_update)
 
     output_processor = SGLangOutputProcessor(capture_hidden=False)
     model_runner = ModelRunner(model_worker, output_processor)
--- a/sglang_omni/models/cosmos3/bootstrap.py
+++ b/sglang_omni/models/cosmos3/bootstrap.py
@@ -90,8 +90,6 @@ def create_thinker_scheduler(
         tree_cache,
         req_to_token_pool,
         token_to_kv_pool_allocator,
-        prefill_manager,
-        decode_manager,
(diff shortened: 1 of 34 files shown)


sglang_omni/models/cosmos3/components/streaming_detokenizer.py
```diff
@@ -0,0 +1,220 @@
...
+        token_id = (
+            int(raw_token.item()) if hasattr(raw_token, "item") else int(raw_token)
+        )
+        state = self._ensure_state(request_id)
+        state.pending_tokens.append(token_id)
+        candidate = self._tokenizer.decode(
+            state.pending_tokens,
+            skip_special_tokens=True,
+        )
+        if "\ufffd" in candidate:
```
JackZeng0208: Could we use `candidate.endswith("\ufffd")` instead of `in`?

Just like what Ming-Omni does: https://github.com/sgl-project/sglang-omni/blob/main/sglang_omni/models/ming_omni/components/streaming_detokenizer.py#L165

If the replacement character appears in the middle of the text, it would block all subsequent streaming output.

sglang_omni/models/cosmos3/components/streaming_detokenizer.py
```diff
@@ -0,0 +1,220 @@
...
+                        data=exc,
+                    )
+                )
+
+    def stop(self) -> None:
+        self._running = False
+
+    def abort(self, request_id: str) -> None:
+        self._state.pop(request_id, None)
+        self._done_seen.pop(request_id, None)
```
JackZeng0208: Could we lock `_state` and `_done_seen`?  The`start()` and `abort()` run on different threads, and they can mutate the shared states at the same time.

Reference (from Ming-Omni): 
- https://github.com/sgl-project/sglang-omni/blob/main/sglang_omni/models/ming_omni/components/streaming_detokenizer.py#L86
- https://github.com/sgl-project/sglang-omni/blob/main/sglang_omni/models/ming_omni/components/streaming_detokenizer.py#L120

sglang_omni/models/cosmos3/components/streaming_detokenizer.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _ensure_state(self, request_id: str) -> _RequestState:
+        state = self._state.get(request_id)
+        if state is None:
+            state = _RequestState()
+            self._state[request_id] = state
```
JackZeng0208: Maybe we should evict stale orphan entries from `_state`. It we don't clean it, it will grow without any boundary. For instance, similar to `_done_seen`, set something like `_STATE_MAX`?

sglang_omni/models/cosmos3/components/streaming_detokenizer.py
```diff
@@ -0,0 +1,249 @@
...
+            return state
+
+    def _evict_idle_orphans(self, now: float) -> None:
+        cutoff = now - _STATE_ORPHAN_IDLE_S
+        with self._state_lock:
+            stale = [
+                request_id
+                for request_id, state in self._state.items()
+                if state.payload is None and not state.done and state.last_seen < cutoff
+            ]
```
WilsonZheng0327, line 117: Two race scenarios that might mess things up
1. abort -> stream_chunk -> stream_done
in this case, the row is resurrected, but payload=None, done=True, so no finalize

2. abort -> new_request
row is resurrected, payload yes, done=False, streaming, no finalize

Solution might be something like _done_seen but for aborted, then a one-line guard at the top of each of the three handlers:
```
if request_id in self._aborted:
    return
```

possibly the stale list should just be:
```
stale = [
    request_id
    for request_id, state in self._state.items()
    if state.last_seen < cutoff
]
```

sglang_omni/models/cosmos3/components/streaming_detokenizer.py
```diff
@@ -0,0 +1,249 @@
...
+    def _ensure_state(self, request_id: str) -> _RequestState:
+        with self._state_lock:
+            now = time.monotonic()
+            state = self._state.get(request_id)
+            if state is None:
+                state = _RequestState(last_seen=now)
+                self._state[request_id] = state
+            state.last_seen = now
+            if len(self._state) > _STATE_MAX:
+                self._evict_idle_orphans(now)
```
WilsonZheng0327: O(n) rescan per token per request if it doesn't end up evicting any orphans

cheap fix: `_last_evict_s` timestamp so the scan runs at most ~1/s
better fix: based on https://github.com/sgl-project/sglang-omni/pull/1263/changes#r3779780680
make `_state` OrderedDict, if the predicate is purely `last_seen`, pop the oldest in the OrderedDict

sglang_omni/models/cosmos3/components/text_preprocessor.py
```diff
@@ -0,0 +1,223 @@
...
+            request_id,
+            prompt_len,
+            int(max_new_tokens),
+            max_seq_len,
+        )
+        raise ValueError(
+            "Requested token count exceeds the model's maximum context length "
+            f"of {max_seq_len} tokens. You requested {prompt_len} prompt tokens "
+            f"and {int(max_new_tokens)} completion tokens."
+        )
```
WilsonZheng0327, line 127: identical in qwen3 and llada2, worth putting in sglang_omni/preprocessing/text.py

sglang_omni/models/cosmos3/components/text_preprocessor.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+def _normalize_text_messages(inputs: Any) -> list[dict[str, str]]:
+    if isinstance(inputs, dict):
+        populated_media = [key for key in _MEDIA_INPUT_KEYS if inputs.get(key)]
+        if populated_media:
+            raise ValueError(
+                "Cosmos3 text preprocessing does not support media inputs yet: "
+                + ", ".join(populated_media)
+            )
```
WilsonZheng0327: reimplements `_reject_media_inputs`

sglang_omni/models/cosmos3/request_builders.py
```diff
@@ -0,0 +1,292 @@
...
+}
+
+
+def _value_or_default(value: Any, fallback: Any) -> Any:
+    return fallback if value is None else value
+
+
+def _request_value(params: dict[str, Any], name: str, default: Any) -> Any:
+    value = params.get(name)
+    return default if value is None else value
```
WilsonZheng0327: these two functions are the same

sglang_omni/models/cosmos3/request_builders.py
```diff
@@ -0,0 +1,292 @@
...
+    PromptInputs,
+    TextOutput,
+)
+from sglang_omni.proto import EXPLICIT_GENERATION_PARAMS_KEY, StagePayload
+from sglang_omni.scheduling.messages import OutgoingMessage
+from sglang_omni.scheduling.sglang_backend import SGLangARRequestData
+from sglang_omni.scheduling.types import RequestOutput
+
+THINKER_STAGE = "thinker"
+DECODE_STAGE = "decode"
```
WilsonZheng0327: already defined in config.py

WilsonZheng0327 (commented):
Mostly comments on concurrency issues for streaming detokenizer. Potentially add some tests for the concurrency/bounding code. Thanks!

sglang_omni/models/cosmos3/components/streaming_detokenizer.py
JingwenGu0829: Implemented in `afeabd51`. `abort()` now records a bounded `_aborted` tombstone, and all three handlers check it inside `_state_lock`, so late chunks, terminal payloads, and done signals cannot resurrect an aborted request. I added regressions for both reported interleavings plus a deterministic test showing abort waits for an in-flight handler and blocks later messages. I kept the orphan-only eviction predicate: evicting a live or `done=True` row solely by `last_seen` could drop tokens or strand completion. The 10k/5k FIFO tombstone bound is intentionally a working-set assumption, like `_done_seen`; an extremely late message after tombstone eviction remains a theoretical residual case.

sglang_omni/models/cosmos3/components/streaming_detokenizer.py
JingwenGu0829: Implemented the cheap throttle in `afeabd51`: once `_state` is over the cap, the O(n) orphan scan now runs at most once per second. The first over-cap scan still runs immediately, and a regression test covers the throttling behavior.

sglang_omni/models/cosmos3/request_builders.py
JingwenGu0829: Done in `62711d06`: removed `_request_value` and now use `_value_or_default(params.get(name), default)` at the sampling call sites.

sglang_omni/models/cosmos3/request_builders.py
JingwenGu0829: Done in `62711d06`: `request_builders.py` now imports `THINKER_STAGE` and `DECODE_STAGE` from `cosmos3.config`. The existing `request_builders.__all__` re-exports are preserved for compatibility.

sglang_omni/models/cosmos3/components/text_preprocessor.py
JingwenGu0829: These helpers differ at the exact context limit: Qwen3/Cosmos3 reject `prompt + max_new_tokens == max_seq_len` with `>=`, while LLaDA2 accepts it with `>`. I avoided changing another model's policy in this Cosmos3-only PR and left the helper local. Would you be okay with handling unification separately using an explicit policy argument?

sglang_omni/models/cosmos3/components/text_preprocessor.py
JingwenGu0829: Done in `62711d06`: `_normalize_text_messages` now calls `_reject_media_inputs` instead of duplicating the media-key check. Existing media rejection tests pass unchanged.

tests/unit_test/cosmos3/test_bootstrap.py
```diff
@@ -0,0 +1,104 @@
...
+    monkeypatch.setattr(model_runner_base, "ModelRunner", lambda *args: object())
+    monkeypatch.setattr(
+        request_builders,
+        "make_text_scheduler_adapters",
+        lambda **kwargs: (object(), object()),
+    )
+    monkeypatch.setattr(
+        request_builders,
+        "make_text_stream_output_builder",
+        lambda: object(),
```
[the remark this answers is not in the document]
JingwenGu0829: Done in `62711d06`: replaced the unnecessary `lambda: object()` wrapper with `object`.

sglang_omni/models/cosmos3/components/text_preprocessor.py
JingwenGu0829: Good point. I compared the implementations and found one behavioral difference: Qwen3 and Cosmos3 reject prompt_len + max_new_tokens == max_seq_len, while LLaDA2 allows equality (>= versus >). A direct extraction would silently change one model’s behavior。

sglang_omni/models/cosmos3/bootstrap.py
```diff
@@ -0,0 +1,123 @@
...
+    local_path = Path(model_path)
+    if local_path.exists():
+        transformer_path = local_path / "transformer"
+    else:
+        allow_patterns = [
+            "transformer/*.json",
+            "transformer/*.safetensors",
+        ]
+        try:
+            snapshot_path = snapshot_download(
```
JackZeng0208, line 31: It looks like HF model revision only gets passed to `ModelConfig` now. `snapshot_download`, `get_tokenizer,` and `AutoTokenizer.from_pretrained` still use the default branch. So we could end up mixing a pinned config with unpinned weights or tokenizer. Could we pass revision through those paths too?

sglang_omni/models/cosmos3/request_builders.py
```diff
@@ -0,0 +1,284 @@
...
+    sampling_params.normalize(tokenizer)
+    sampling_params.verify(vocab_size)
+
+    req = Req(
+        rid=request_id or "cosmos3-request",
+        origin_input_text=prompt["prompt_text"],
+        origin_input_ids=input_ids.tolist(),
+        sampling_params=sampling_params,
+        eos_token_ids=_eos_token_ids(tokenizer, generation_config),
+        vocab_size=vocab_size,
```
rwang5203, line 165: Pretokenized `/generate` requests embed unchecked `input_ids`, so one out-of-range ID crashes default TP1 serving while TP2 substitutes a zero embedding.

Reject IDs outside `[0, vocab_size)` before `Req` construction and add TP1 and TP2 regressions.

sglang_omni/models/cosmos3/components/text_preprocessor.py
```diff
@@ -0,0 +1,230 @@
...
+            raise TypeError(f"Message {index} must be a dict with role/content")
+        role = message.get("role")
+        content = message.get("content")
+        if not isinstance(role, str) or not role:
+            raise ValueError(f"Message {index} must have a non-empty string role")
+        if not isinstance(content, str):
+            raise TypeError(
+                f"Message {index} content must be a string; multimodal content "
+                "will be added in a later Cosmos3 stage"
+            )
```
rwang5203, line 91: Text-only OpenAI content parts fail the string check, so valid structured chat requests return HTTP 500 before inference.

Normalize text parts to a string while still rejecting media parts, then add an API regression.

sglang_omni/models/cosmos3/components/streaming_detokenizer.py
```diff
@@ -0,0 +1,263 @@
...
+            candidate = self._tokenizer.decode(
+                state.pending_tokens,
+                skip_special_tokens=True,
+            )
+            if candidate.endswith("\ufffd"):
+                return
+            state.pending_tokens.clear()
+            if not candidate:
+                return
+            self._emit_text(request_id, candidate)
```
rwang5203, line 167: Raw token decoding never trims matched stops, so `stop` and `stop_token_ids` delimiters leak into streamed and final text.

Preserve matched-stop metadata, trim terminal text, and buffer possible stop prefixes while streaming.

sglang_omni/models/cosmos3/bootstrap.py
```diff
@@ -0,0 +1,132 @@
...
+    ) = create_sglang_infrastructure(
+        server_args,
+        gpu_id,
+        tp_rank=tp_rank,
+        nccl_port=nccl_port,
+        model_arch_override="Cosmos3TextForCausalLM",
+        model_weights_path=resolve_transformer_weights_path(
+            server_args.model_path,
+            revision=server_args.revision,
+        ),
```
rwang5203, line 103: Only startup resolves the nested `transformer/` directory, so disk updates with the advertised model root return HTTP 400 before loading.

Apply the Cosmos3 resolver during disk updates and preserve the logical root in model metadata.

sglang_omni/models/cosmos3/request_builders.py
```diff
@@ -0,0 +1,284 @@
...
+    raw_explicit_fields = metadata.get(EXPLICIT_GENERATION_PARAMS_KEY)
+    has_explicit_marker = raw_explicit_fields is not None
+    explicit_fields = set(raw_explicit_fields or ())
+
+    for field, transport_default in _TRANSPORT_SAMPLING_DEFAULTS.items():
+        if field in explicit_fields:
+            continue
+        value = effective.get(field)
+        if has_explicit_marker or value is None or value == transport_default:
+            effective.pop(field, None)
```
rwang5203, line 114: `SamplingParams()` and explicit `temperature=1.0` serialize identically, so the ordinary direct-client constructor silently applies Nano's 0.7 default.

Track explicitly supplied sampling fields and forward their marker from `Client`.

sglang_omni/models/cosmos3/payload_types.py
```diff
@@ -0,0 +1,56 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Cosmos3 payload schemas."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any, NotRequired, TypedDict
```
rwang5203: `typing.NotRequired` is unavailable on the declared Python 3.10 minimum, so constructing the Cosmos3 pipeline fails during module import.

Import `NotRequired` from `typing_extensions` and add a Python 3.10 construction regression.

sglang_omni/models/cosmos3/request_builders.py
JingwenGu0829: I think no production path in normal OpenAI format or \generate (which corresponds to the miles RL contracts) would produce mis-behaving tokenID.

Another thing is that it seems that from the start of sglang https://github.com/sgl-project/sglang/pull/363  /generate(input_ids=...) has been the trusted token-in/token-out rollout interface. Additionally sglang-omni models are not checking this too. 

So if we can't reproduce the production sad-path, maybe a double check here is a little over-protection?

sglang_omni/models/cosmos3/bootstrap.py
JingwenGu0829: I feel like resolving this during disk update is going to be super-invasive to the disk update, modelworker and scheduling logic. 

And since miles are not yet convering Cosmos for RL yet, do you think it's better we just explicitly disable disk update?

If in the future we actually have to encounter Cosmos with disk weight update (likely in RL, but RL uses nccl for weight update path more right now too), we do a followup pr at that time thinking how we re-enable disk weight update for cosmos.

JingwenGu0829: It's actually a workaround as proposed above. But as long as cosmos does not support RL, or we are not massively supporting models with diffusers it should be fine.

sglang_omni/models/cosmos3/request_builders.py
rwang5203: Makes sense. Closing.

sglang_omni/models/cosmos3/components/streaming_detokenizer.py
rwang5203: Thanks for addressing non-streaming output and `stop_token_ids`. One streaming case still remains. A live `stop=" gamma"` request emitted `" gamma"` before terminal metadata arrived. Could you buffer possible stop prefixes and add a normal-token regression?

sglang_omni/models/cosmos3/bootstrap.py
rwang5203: A quick follow-up here. The proposed disable has not landed, and the resolver path is unchanged. To close this before merge, please resolve the advertised root, add a clear and discoverable disable, or link the concrete follow-up in this thread.

JingwenGu0829: Yeah. Just landed. Did a simple model-specific strong rejection.

sglang_omni/models/cosmos3/components/text_preprocessor.py
```diff
@@ -0,0 +1,249 @@
...
+            text_parts: list[str] = []
+            for part in content:
+                if isinstance(part, str):
+                    text_parts.append(part)
+                elif isinstance(part, dict) and part.get("type", "text") == "text":
+                    text = part.get("text")
+                    if not isinstance(text, str):
+                        raise ValueError(
+                            f"Message {index} text part must have a string text field"
+                        )
```
rwang5203, line 78: A live Cosmos3 server returns HTTP 500 for a nonstring `text` field even though this is malformed client input.

Could we map this validation error to HTTP 400 and add a nonstring `text` API regression?

rwang5203 (changes requested):
Thank you for the cleanup and malformed-input fix, which now returns HTTP 400 in both modes while preserving the stop and weight-update protections. Cosmos3 still can't start because its bootstrap expects seven helper outputs and passes two manager arguments that the current scheduler no longer accepts. Could you align those contracts and confirm an exact pipeline launch before approval?

PR #1265, opened 2026-08-02
[Cosmos02] Vision encoder for UND (AR backbone)

JingwenGu0829:
Scope

- Extend the Nano thinker path from text-only generation to image/video input.
- Keep the vision tower in its own `vision_encoder` stage; diffusion
  infrastructure, GEN, and generation-side codecs/adapters remain out of scope.
- Preserve the Stack 1 text-only path and native thinker tensor parallelism.

## Architecture

- The checkpoint embeds `Qwen3VLVisionConfig` (`model_type=qwen3_vl_vision`).
- The corresponding HF class is `Qwen3VLVisionModel`; the standalone weights
  live in `vision_encoder/model.safetensors`.
- Nano returns one 4096-wide visual embedding plus three 4096-wide deepstack
  features captured at vision layers 8, 16, and 24.

## Pipeline

```text
preprocessing ─┬─> vision_encoder ─┐
               └───────────────────┴─> thinker -> decode
```

- Text-only requests route directly from preprocessing to `thinker`.
- Image/video requests send only processor tensors to `vision_encoder`; the
  thinker receives a lightweight prompt latch and encoder outputs, then merges
  them at its input boundary.
- There is no communication-only aggregation stage. Encoder embeddings cross
  stage transport once and use its direct CUDA IPC path into the thinker.

## Qwen-Omni patterns retained

- Non-overlapping Conv3d patch projection is replaced with an equivalent Linear.
- Vision requests use cost-bounded micro-batching (32 requests / 10 GiB budget).
- A 64-entry, 4 GiB CPU LRU caches encoder outputs and deduplicates identical
  media within one batch.
- Media-derived placeholder ids make SGLang's radix cache reusable for repeated
  media without allowing two different images/videos to share visual KV.
- Main visual embeddings and deepstack residual embeddings are injected only
  during prefill; decode continues through the ordinary SGLang KV-cache path.

## Correctness details

- The Nano export records Transformers 4.57.6. The vision wrapper retains its
  positional-interpolation arithmetic instead of silently taking the changed
  Transformers 5.12 path in the current container.
- Multimodal positions follow Cosmos3's own `mm_token_type_ids` algorithm. Text
  fills all three T/H/W axes; image/video spans use their merged visual grids.
- Video cache keys include FPS, frame cap, and pixel budgets because these alter
  the encoder input and output.
- Audio input is rejected explicitly. Nano Reasoner support in this stack is
  text plus image/video; sound latents belong to the later generation pipeline.

## Validation

- `44` Cosmos3 unit tests cover conditional routing, preprocessing, weight
  loading, batching/cache deduplication, thinker-side merging, mRoPE, and
  deepstack injection. The combined Cosmos3, registry, CLI, and pipeline
  selection is `152` tests; the complete unit collection needs optional packages
  absent from this container (`jiwer`, `x_transformers`, and `onnxruntime`).
- The real Nano vision checkpoint loaded on one H200. A 256x256 image produced
  64 main embeddings of shape `[64, 4096]` and three deepstack tensors with the
  same shape.
- The four-stage server completed image-to-text generation: a brown test image
  returned `Brown`, a blue image returned `Blue`, and the text-only bypass still
  returned `OK`.
- Repeating the same image reused 84 of 85 prompt KV tokens. Changing the image
  reused only the four-token text prefix, confirming media cache isolation.
- Decode CUDA graph captured batches `[1, 2, 4, 8, 12, 16]`; a 96-token
  multimodal generation reported `cuda graph: True` during decode.

### Full MMMU image-text benchmark

A single full-set run was recorded on 2026-08-01 against commit `e9f098e0`.
The server used one H200, TP=1, the four colocated stages above, and the Nano
snapshot `411f42a8fdfb8c5b2583cb8786e0938f49796eaa`. The benchmark used all 900
image-bearing samples from the 30-subject `MMMU/MMMU` validation split at
dataset revision `98e6ac0cb9b7b2cd2c991b85a50762edc4aedc68`, five warmups,
concurrency 8, greedy decoding, and `max_tokens=2048`.

| Metric | Result |
|---|---:|
| Accuracy | 57.44% (517/900) |
| Failed requests | 0/900 |
| Multiple-choice parse fallbacks | 23 |
| Request throughput | 2.265 req/s |
| Mean / median latency | 3.520 / 2.481 s |
| P95 / P99 latency | 10.413 / 14.759 s |
| Output throughput | 977.6 tok/s |
| Mean output tokens | 432 |

This is one directional accuracy-and-serving run, not a repeated performance
study. The environment used SGLang 0.5.16, Transformers 5.12.1, PyTorch
2.11.0+cu130, CUDA 13.0, and NVIDIA driver 595.71.05. All requests completed,
and server logs continued to report decode CUDA-graph execution. The raw JSON
was retained at `/tmp/cosmos3-mmmu-results.7NaOoe/mmmu_results.json` with SHA256
`e101159dc11cba45d10132f7b46e06906072f60f704855e093aa0230ea1719ac`.

### Official SGLang comparison

The same 900-sample run was repeated against the official SGLang Cosmos3
Reasoner implementation in
[sglang#30536](https://github.com/sgl-project/sglang/pull/30536), pinned to PR
head `80b2296eabd55fad910cf0efc5732156b8192e84`. The dataset revision, sample
order, prompts, answer parser, five warmups, concurrency 8, greedy decoding, and
`max_tokens=2048` were unchanged. Only the wire-format adapter changed: Omni's
top-level `images` field became standard OpenAI `image_url` content parts for
the upstream server.

| Backend | Accuracy | Failed | MC parse fallback |
|---|---:|---:|---:|
| SGLang-Omni, four stages | 57.44% (517/900) | 0 | 23 |
| Official SGLang PR #30536 | 57.22% (515/900) | 0 | 28 |

### DIFF
--- a/sglang_omni/models/cosmos3/components/text_preprocessor.py
+++ b/sglang_omni/models/cosmos3/components/text_preprocessor.py
@@ -7,6 +7,7 @@ import asyncio
 import logging
 from pathlib import Path
 from typing import Any
+from uuid import uuid4
 
 import torch
 from transformers import AutoProcessor, AutoTokenizer
@@ -257,7 +258,7 @@ class Cosmos3TextPreprocessor:
     async def _tokenize_multimodal(
         self,
         inputs: dict[str, Any],
-    ) -> tuple[dict[str, torch.Tensor], str, str | None]:
+    ) -> tuple[dict[str, torch.Tensor], str, str | None, str | None]:
         if self.processor is None:
             raise ValueError("Cosmos3 media inputs require the checkpoint processor")
 
@@ -344,8 +345,13 @@ class Cosmos3TextPreprocessor:
                     f"total_pixels={video_total_pixels}",
                 )
             )
-        cache_parts = [part for part in (image_cache_key, video_cache_key) if part]
-        return dict(encoded), prompt_text, "|".join(cache_parts) or None
+
+        request_cache_id = uuid4().hex
+        if images and image_cache_key is None:
+            image_cache_key = f"image:request:{request_cache_id}"
+        if videos and video_cache_key is None:
+            video_cache_key = f"video:request:{request_cache_id}"
+        return dict(encoded), prompt_text, image_cache_key, video_cache_key
 
     @staticmethod
     def _use_pretokenized_inputs(inputs: Any) -> bool:
@@ -392,9 +398,12 @@ class Cosmos3TextPreprocessor:
                     raise ValueError(
                         "Cosmos3 text-only pipeline does not accept image/video input"
                     )
-                encoded, prompt_text, cache_key = await self._tokenize_multimodal(
-                    inputs
-                )
+                (
+                    encoded,
+                    prompt_text,
+                    image_cache_key,
+                    video_cache_key,
+                ) = await self._tokenize_multimodal(inputs)
                 input_ids = _flatten_single_batch(encoded["input_ids"])
                 raw_attention_mask = encoded.get("attention_mask")
                 attention_mask = (
@@ -423,8 +432,10 @@ class Cosmos3TextPreprocessor:
                     )
                     if isinstance(encoded.get(key), torch.Tensor)
                 }
-                if cache_key is not None:
-                    vision_inputs["cache_key"] = cache_key
+                if image_cache_key is not None:
+                    vision_inputs["image_cache_key"] = image_cache_key
+                if video_cache_key is not None:
+                    vision_inputs["video_cache_key"] = video_cache_key
             else:
                 input_ids, attention_mask, prompt_text = self._tokenize_messages(inputs)
                 mm_token_type_ids = torch.zeros_like(input_ids)
--- a/sglang_omni/models/cosmos3/components/text_preprocessor.py
+++ b/sglang_omni/models/cosmos3/components/text_preprocessor.py
@@ -1,23 +1,30 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Text-only preprocessing for Cosmos3-Nano."""
+"""Text and vision preprocessing for Cosmos3-Nano."""
 
 from __future__ import annotations
 
+import asyncio
 import logging
 from pathlib import Path
 from typing import Any
 
 import torch
-from transformers import AutoTokenizer
+from transformers import AutoProcessor, AutoTokenizer
 
 from sglang_omni.models.cosmos3.payload_types import Cosmos3PipelineState
+from sglang_omni.preprocessing import (
+    compute_image_cache_key,
+    compute_video_cache_key,
+    ensure_image_list_async,
+    ensure_video_list_async,
+)
 from sglang_omni.preprocessing.text import ensure_chat_template
 from sglang_omni.proto import StagePayload
 
 logger = logging.getLogger(__name__)
 
 DEFAULT_MAX_NEW_TOKENS = 2048
-_MEDIA_INPUT_KEYS = ("images", "videos", "audios")
+VISION_STAGE = "vision_encoder"
 
 
 def load_cosmos3_tokenizer(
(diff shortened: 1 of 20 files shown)


Zhylkaaa:
Hi @JingwenGu0829, commenting to let you know I cleaned out the previous PR that was rebased onto broken main. Now the reasoner implementation lives in: https://github.com/sgl-project/sglang/pull/33572
Sorry for the confusion

JingwenGu0829:
@Zhylkaaa  Hi Im a little bit confused. The prev pr of this is https://github.com/sgl-project/sglang-omni/pull/1263

Zhylkaaa:
Hi @JingwenGu0829, I just saw you mentioned my previous PR in the comment: https://github.com/sgl-project/sglang/pull/30536 , which I closed and reopened with clean commit history. They should be functionally the same.

sglang_omni/models/cosmos3/components/text_preprocessor.py
```diff
@@ -169,6 +189,129 @@ def _tokenize_messages(self, inputs: Any) -> tuple[torch.Tensor, torch.Tensor, s
...
+                for part in (
+                    video_cache_key,
+                    f"fps={effective_fps}",
+                    f"max_frames={video_max_frames}",
+                    f"min_pixels={video_min_pixels}",
+                    f"max_pixels={video_max_pixels}",
+                    f"total_pixels={video_total_pixels}",
+                )
+            )
+        cache_parts = [part for part in (image_cache_key, video_cache_key) if part]
```
JackZeng0208: If one modality hash and the other returns None, `cache_parts` drops the None. So "image A + video V" and "image B + video V" share a cache key, and the stage cache/radix pad values serve A's vision output to B. When `media_cache_key` is None, `build_sglang_text_request` also skip placeholder substitution, so different images share a radix prefix. 

Could the combined key be all or nothing? Maybe a per-request unique fallback is better than raw placeholder ids?

rwang5203:
Head `c78fd235` conflicts with base `53ce88c3` across 15 files. [Omni CI](https://github.com/sgl-project/sglang-omni/blob/c78fd2353abc2b8687b1061b93b2b127191c0b4d/.github/workflows/omni-ci.yaml) skips non-main branches, leaving only [cancel](https://github.com/sgl-project/sglang-omni/actions/runs/32087163487/job/95561977158), [JavaScript](https://github.com/sgl-project/sglang-omni/actions/runs/32087162644/job/95561977041), and [Python](https://github.com/sgl-project/sglang-omni/actions/runs/32087162644/job/95561976913) checks.

Please rebase with the base fixes and revision pinning intact, then run labeled CI.

sglang_omni/models/cosmos3/components/text_preprocessor.py
```diff
@@ -195,6 +190,134 @@ def _tokenize_messages(self, inputs: Any) -> tuple[torch.Tensor, torch.Tensor, s
...
+
+        raw_images = inputs.get("images")
+        raw_videos = inputs.get("videos") or inputs.get("video")
+        video_fps = inputs.get("video_fps")
+        video_max_frames = inputs.get("video_max_frames")
+        video_min_pixels = inputs.get("video_min_pixels")
+        video_max_pixels = inputs.get("video_max_pixels")
+        video_total_pixels = inputs.get("video_total_pixels")
+        image_cache_key = compute_image_cache_key(raw_images)
+        video_cache_key = compute_video_cache_key(raw_videos)
```
rwang5203, line 273: URL keys are computed before fetch, so changed content at the same address can reuse stale encoder and radix-KV results.

You could hash fetched bytes and test mutable URLs plus byte-identical aliases.

sglang_omni/models/cosmos3/components/text_preprocessor.py
```diff
@@ -195,6 +190,134 @@ def _tokenize_messages(self, inputs: Any) -> tuple[torch.Tensor, torch.Tensor, s
...
+                    f"max_pixels={video_max_pixels}",
+                    f"total_pixels={video_total_pixels}",
+                )
+            )
+
+        request_cache_id = uuid4().hex
+        if images and image_cache_key is None:
+            image_cache_key = f"image:request:{request_cache_id}"
+        if videos and video_cache_key is None:
+            video_cache_key = f"video:request:{request_cache_id}"
```
rwang5203, line 353: Frame-list videos reach the processor but produce no content hash, so each request gets a UUID key and evicts reusable encoder entries.

Consider skipping request-scoped caching or hashing frame lists by content.

rwang5203 (changes requested):
Great job for rebuilding the branch on the current parent and adapting the topology to the canonical config APIs. The child topology reaches serving once the parent startup contract is corrected, but the mutable URL cache remains a correctness blocker because keys computed before fetch can reuse stale encoder and radix state. One approach is to derive cache identity from fetched content, add a mutable URL regression, and land the parent fix before approval. Can you take a look into it?

Issue #1248, opened 2026-07-31
[Feature] Qwen3-TTS: true streaming output and streaming input


zhaochenyang20:
On current `main`, Qwen3-TTS does **not** provide real low-latency streaming, even though the cookbook documents `"stream": true`.

**What works today**
- Non-streaming `/v1/audio/speech` (full WAV/PCM response).
- HTTP `"stream": true` + `"response_format": "pcm"` is accepted by the API/docs, but the pipeline still uses `BatchVocoderBase` with `supports_streaming_vocoder=False`. Codec generation finishes first; the vocoder then decodes the full utterance and the HTTP layer chunks the already-complete audio. That is buffered/"fake" streaming, not incremental decode.

Evidence on `main`:
- `sglang_omni/models/qwen3_tts/__init__.py`: `supports_streaming_vocoder=False`
- `sglang_omni/models/qwen3_tts/stages.py`: vocoder inherits `BatchVocoderBase`
- #1098 measured TTFA with `stream=true` as ~6.8s vs ~6.0s non-streaming on Qwen3-TTS-12Hz-1.7B-Base (first audio arrives only after full generation)

**What is missing**

1. **True streaming output**  
   Incremental codec → chunked Code2Wav decode → early PCM frames over HTTP (`/v1/audio/speech` with `stream=true`) and over WebSocket (`stream_audio=true`), with competitive TTFA (target: first audio well before utterance completion; vLLM-Omni-style windowed Code2Wav is a useful reference).

2. **Streaming input**  
   Reliable `/v1/audio/speech/stream` support for Qwen3-TTS when text arrives incrementally (e.g. LLM token/sentence stream):
   - accept `session.config` + repeated `input.text` + `input.done`
   - synthesize as text segments become available (`split_granularity`: sentence/clause)
   - with `stream_audio=true`, emit incremental PCM rather than one blob per segment  
   Today the WebSocket protocol exists, but Qwen3-TTS cannot deliver true incremental audio because the streaming vocoder path is not on `main`. CustomVoice / VoiceDesign also force `non_streaming_mode=True`, so scope for Base/reference-cloning vs those variants should be explicit.

**Related work (please avoid duplication)**
- Tracking / prior request: #1098
- In-progress PR: #1240 (`[TTS][Qwen3] Add true streaming vocoder`)
- Earlier attempt: #704
- Roadmap note: #599 (streaming path still listed as TODO historically)
- Continuity under concurrency: #1234

[...]

PR #1286, opened 2026-08-02
feat(qwen3-tts): add true incremental streaming

Hayden727:
## Summary

Narrow replacement for #1240, preserving only the Qwen3-TTS true-streaming path:

- emit generated codec rows from the AR stage as device-resident tensor stream chunks
- prepend reference codes once, then incrementally decode Base/reference-cloning traffic in the model-local vocoder
- use a high-priority initial CUDA stream and lower-priority, deadline-ordered follow-up batches
- preserve stream-before-payload handling, final-stream-before-result ordering, non-streaming batch decode, and abort/failure/stop cleanup
- advertise the streaming-vocoder capability and document the supported Base/reference-cloning scope

CustomVoice, VoiceDesign, and explicit `non_streaming_mode` generation keep the full-utterance fallback rather than publishing incremental codec rows.

## Split from #1240

Deliberately removed unrelated policy/experiments from the original PR:

- 1.7B low-VRAM profile
- preprocessing concurrency changes
- `max_running_requests` increase
- separate reference-encoding CUDA stream
- sampling-loop rename/comment-only churn

The 1.7B consumer-GPU profile will be validated independently under #1120.

## Correctness hardening

- clamp `initial_codec_chunk_frames` identically whether payload or codec chunk arrives first
- keep codec tensors on their source device; avoid a GPU→CPU→GPU round trip and CUDA value-reduction sync on every row
- make the vocoder CUDA stream wait for codec staging before decode
- keep full-decode fallback reference trimming identical to the existing non-streaming vocoder
- gate incremental codec emission on the model's `non_streaming_mode` contract

## Validation

- `87 passed, 11 deselected` for the Qwen3-TTS pipeline suite on macOS; the 11 deselections require unavailable local `sgl_kernel`
- `22 passed` for focused streaming, async lifecycle, vocoder batching, and production request-builder coverage
- `2 passed` for Qwen3-TTS capability export/lookup
- changed-file pre-commit hooks passed
- architecture drift audit: changes are model-local and use the existing `StageConfig.stream_to`, `can_accept_stream_before_payload`, `OmniScheduler.stream_output_builder`, and `StreamingVocoderBase` contracts

## Benchmark provenance

The H100 measurements below are from the original #1240 author and were not rerun in this narrowing pass (Seed-TTS EN, default config, 64 SGLang requests; baseline 1088 requests):

- C8: TTFA p95 0.421s vs 0.330s; QPS 5.825 vs 2.921; C200 90.6% vs 0%
- C16: TTFA p95 0.243s vs 0.497s; QPS 8.868 vs 3.107; C200 39.1% vs 0%
- C32: TTFA p95 0.422s vs 0.895s; QPS 11.348 vs 3.155; C200 21.9% vs 0%

Addresses #1248. Related to #1120.

### DIFF
--- a/docs/cookbook/qwen3_tts.md
+++ b/docs/cookbook/qwen3_tts.md
@@ -156,6 +156,11 @@ Streaming returns `audio/pcm` 16-bit mono PCM bytes with sample-rate metadata in
 the response headers. See the [Higgs TTS cookbook](../cookbook/higgs_tts.md#streaming)
 for a full Python raw PCM consumer.
 
+Base/reference-cloning checkpoints use true incremental codec and vocoder
+streaming for both this HTTP endpoint and `/v1/audio/speech/stream` WebSocket
+sessions with `stream_audio=true`. CustomVoice and VoiceDesign remain
+non-streaming.
+
 ## Generation Parameters
 
 | Parameter | Default | Notes |
--- a/sglang_omni/models/qwen3_tts/__init__.py
+++ b/sglang_omni/models/qwen3_tts/__init__.py
@@ -8,7 +8,7 @@ from . import config
 CAPABILITIES = ModelCapabilities(
     supports_reference_audio=True,
     supports_batch_vocoder=True,
-    supports_streaming_vocoder=False,
+    supports_streaming_vocoder=True,
     supports_cuda_graph=True,
     supports_torch_compile=True,
 )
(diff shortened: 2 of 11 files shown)


luojiaxuan:
Latest full H100 result from the split TTFA follow-up branch, not this PR head. Each setup used 1,088 Seed-TTS EN requests.

C8: Current/Baseline TTFA p95 0.266s/0.330s; C200 84.74%/0%; QPS 6.114/2.921; EN WER 1.06%/1.07%.
C16: Current/Baseline TTFA p95 0.320s/0.497s; C200 12.41%/0%; QPS 7.746/3.107; EN WER 1.06%/1.06%.
C32: Current/Baseline TTFA p95 0.541s/0.895s; C200 3.03%/0%; QPS 9.612/3.155; EN WER 1.05%/1.06%.

All runs completed 1,088/1,088. First payload p95 was 3.75 KiB.

luojiaxuan (approved):

PR #1290, opened 2026-08-02
[Fix] Keep MOSS-TD encoder OOM request-scoped

### LINKED ISSUE #1201: [Bug] ASR stage crashes instead of degrading after encoder OOM (MOSS-TD)
### Checklist

### Describe the bug

On a memory-constrained GPU (RTX 5090, 32 GB), long-audio requests at high concurrency under the default profile (`mem_fraction_static=0.80, max_running_requests=16`) first fail with encoder OOM — expected — but the stage then crashes and all subsequent requests fail:

```
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 3.10 GiB. GPU 0 has a total capacity of 31.36 GiB of which 858.75 MiB is free. ...
RuntimeError: scheduler crashed: Tensors must have same number of dimensions: got 1 and 3
RuntimeError: Dead stage process(es) detected: asr (pid=..., exit=1)
```

## Expectation

Encoder OOM should degrade gracefully (reject the offending request, keep the stage alive). The `Tensors must have same number of dimensions: got 1 and 3` crash suggests the post-OOM recovery path builds a malformed batch. H100-class CI never hits this because the pool leaves ample encoder headroom — consumer-GPU deployments (#1120) will.

Found during the RTX 5090 validation (#1189); memory-boundary data and raw logs there.

### Reproduction

1. `sgl-omni serve --model-path OpenMOSS-Team/MOSS-Transcribe-Diarize --max-running-requests 16 --cuda-graph-max-bs 16 --mem-fraction-static 0.80` on a 32 GB GPU (main @ dc64a6b2)
2. Send ~38-min audio (e.g. AISHELL4 `val_000013.flac`) at concurrency 16 to `/v1/audio/transcriptions`
3. Some requests OOM in the encoder (expected on 32 GB); shortly after, the scheduler crashes with the dimension error above and the ASR stage dies; every later request returns 500 instantly

Contrast: at concurrency 1 the same over-budget request fails as a clean request-level 500 with the OOM detail and the server stays healthy — that is the behavior high-concurrency OOM should follow too.

### Environment

- GPU: 1x NVIDIA GeForce RTX 5090 32,607 MiB (SM120), driver 595.71.05, power limit 575 W
- OS: Ubuntu 24.04, Linux 6.17.0-14-generic
- CUDA toolkit: 13.2
- Python 3.12.13; torch 2.11.0+cu130; sglang 0.5.12.post1; transformers 5.6.0; flashinfer (default-selected attention backend)
- sglang-omni: main @ dc64a6b2, editable install, no source changes
- Model: OpenMOSS-Team/MOSS-Transcribe-Diarize @ e5118b411bf5a77d7a90c4941066bec93c967312
- Serve args: --max-running-requests 16 --cuda-graph-max-bs 16 --mem-fraction-static 0.80
- Full run manifest: https://gist.github.com/ChaoyuWang04/d3105c53889740f038d6477dbf112b94 (run_manifest.json)

Hayden727:
## Summary

- keep MOSS-Transcribe-Diarize encoder item state transactional until the dedicated CUDA stream synchronizes successfully
- make a singleton encoder OOM request-scoped without retrying the same impossible allocation; retain per-item fallback for failed aggregate batches
- recover the allocator on the encoder stage's selected CUDA device
- format unexpected failure tracebacks for diagnostics, then detach live traceback/cause/context references before logging or publishing the request error
- cover the real queue/worker/`encode_item` future path, allocator cleanup ordering, buffered-log retention, failed-intermediate release, non-OOM traceback diagnostics, and follow-on request health

Closes #1201.

## Why

An asynchronous encoder OOM could surface at stream synchronization after `precomputed_embeddings` had already been published and `feature` cleared. The fallback then retried poisoned items, while raw exception tracebacks could retain failed encoder tensors. Under pressure this could cascade from one request failure into an unhealthy ASR stage.

Unexpected non-OOM encoder failures still need a complete traceback for diagnosis. The traceback is now formatted to text before live exception state is detached, preserving diagnostics without retaining failed tensor frames in futures or buffered log records.

The equivalent Fun-ASR pre-LM encoder lifecycle issue is tracked separately in #1301 so this PR remains model-local.

## Validation

- `python -m pytest tests/unit_test/moss_transcribe_diarize/test_encoder_service.py -q` — 6 passed
- full `tests/unit_test/moss_transcribe_diarize` suite — 60 passed, 14 skipped in the local CPU environment
- `pre-commit run --files sglang_omni/models/moss_transcribe_diarize/encoder_service.py tests/unit_test/moss_transcribe_diarize/test_encoder_service.py tests/README.md`

Real asynchronous CUDA OOM reclamation still requires the repository GPU lane; the unit tests model the poisoned-allocator, stream-publication, traceback-retention, and follow-on health contracts deterministically.

### DIFF
--- a/sglang_omni/models/moss_transcribe_diarize/encoder_service.py
+++ b/sglang_omni/models/moss_transcribe_diarize/encoder_service.py
@@ -12,6 +12,7 @@ import logging
 import queue
 import threading
 import traceback
+from dataclasses import dataclass
 from typing import Any
 
 import torch
@@ -22,6 +23,12 @@ logger = logging.getLogger(__name__)
 
 _CACHE_MAX_ENTRIES = 4096
 _CACHE_MAX_BYTES = 2 * 1024**3
+
+
+@dataclass(frozen=True)
+class _DetachedFailure:
+    exception: Exception
+    formatted_traceback: str
 
 
 class BatchedAudioEncoderService:
@@ -84,42 +91,33 @@ class BatchedAudioEncoderService:
         except Exception as batch_exc:
             if len(batch) == 1:
                 failure = self._detach_failure(batch_exc)
-                failure_type = type(failure).__name__
-                failure_message = str(failure)
                 logger.error(
-                    "MOSS-TD audio encode failed: %s: %s",
-                    failure_type,
-                    failure_message,
+                    "MOSS-TD audio encode failed:\n%s",
+                    failure.formatted_traceback,
                 )
-                self._recover_after_failure(failure)
-                batch[0][1].set_exception(failure)
+                self._recover_after_failure(failure.exception)
+                batch[0][1].set_exception(failure.exception)
                 return
 
             failure = self._detach_failure(batch_exc)
-            failure_type = type(failure).__name__
-            failure_message = str(failure)
             logger.error(
                 "MOSS-TD batched audio encode failed for %d items; "
-                "retrying per item: %s: %s",
+                "retrying per item:\n%s",
                 len(items),
-                failure_type,
-                failure_message,
+                failure.formatted_traceback,
             )
-            self._recover_after_failure(failure)
+            self._recover_after_failure(failure.exception)
             for item, future in batch:
                 try:
                     self._encode_batch([item])
                 except Exception as item_exc:
                     failure = self._detach_failure(item_exc)
-                    failure_type = type(failure).__name__
-                    failure_message = str(failure)
                     logger.error(
-                        "MOSS-TD per-item audio encode retry failed: %s: %s",
-                        failure_type,
-                        failure_message,
+                        "MOSS-TD per-item audio encode retry failed:\n%s",
+                        failure.formatted_traceback,
                     )
-                    self._recover_after_failure(failure)
-                    future.set_exception(failure)
+                    self._recover_after_failure(failure.exception)
+                    future.set_exception(failure.exception)
                 else:
                     self._record_success(1)
                     future.set_result(None)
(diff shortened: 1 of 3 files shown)


sglang_omni/models/moss_transcribe_diarize/encoder_service.py
```diff
@@ -72,33 +73,100 @@ def _drain_batch(self) -> list[tuple[Any, concurrent.futures.Future]]:
...
+        exc.__traceback__ = None
+        exc.__cause__ = None
+        exc.__context__ = None
+        if isinstance(exc, torch.OutOfMemoryError):
+            return torch.OutOfMemoryError(message)
+        if isinstance(exc, ValueError):
+            return ValueError(message)
+        return RuntimeError(f"{type(exc).__name__}: {message}")
+
+    def _recover_after_failure(self, exc: Exception) -> None:
```
JiaxinD, line 149: `FunASRPreLMEncoderService` has the same four problems this fixes: the raw exception goes into the future with its live traceback (fun_asr/encoder_service.py:325), `logger.exception` at :331, `_attach()` at :399 runs before `synchronize()` at :401, and there is no allocator recovery at all. Worth sharing this handler, or at least filing the sibling so it does not sit unfixed.

sglang_omni/models/moss_transcribe_diarize/encoder_service.py
```diff
@@ -72,33 +73,100 @@ def _drain_batch(self) -> list[tuple[Any, concurrent.futures.Future]]:
...
+                    failure_message,
                 )
+                self._recover_after_failure(failure)
+                batch[0][1].set_exception(failure)
+                return
+
+            failure = self._detach_failure(batch_exc)
+            failure_type = type(failure).__name__
+            failure_message = str(failure)
+            logger.error(
```
JiaxinD, line 103: Non OOM failures now lose their traceback entirely, since `_detach_failure` clears it and the log keeps only type and message, which makes an unexpected encoder error much harder to chase. Formatting the traceback to a string before detaching would keep both the diagnosis and the tensor release.

JiaxinD (commented):
This looks right to me. I reverted the production file to base and reran the new tests: all three fail, and two of them for the behavioural reason rather than a missing symbol, `assert None is tensor([1.])` showing base published embeddings after the sync raised, and `allocator remained poisoned after OOM` showing base never cleared the allocator. The failure also reaches the client properly, through the scheduler abort and on to a 500 or an SSE error event, so it is not turning into a silent empty transcription. Two notes below, neither blocking.

sglang_omni/models/moss_transcribe_diarize/encoder_service.py
Hayden727: Tracked the Fun-ASR sibling separately in #1301 rather than expanding this model-local PR. The issue records the raw Future exception, logger.exception retention, pre-synchronize item mutation, missing allocator recovery, and the required production-path/GPU validation.

sglang_omni/models/moss_transcribe_diarize/encoder_service.py
Hayden727: Fixed in 97592cdd. `_detach_failure()` now formats the complete traceback to text before clearing frames/cause/context, logs only that string, and publishes the detached exception to the Future. The new non-OOM regression proves the traceback remains diagnostic while failed intermediates and live exception state are released.

JiaxinD (approved):

SandyLuXY (approved):

PR #1294, opened 2026-08-02
[Fix] Restore accelerator tensors to receiver device after SHM transport

ppppqp:
## Motivation

Fix tensor placement when an accelerator-origin payload is transferred through the host shared-memory relay.

For CUDA IPC, it's possible to transfer tensors directly between GPU, i.e. sendor GPU -> receiving GPU. 

However, for other host-staged accelerator transports, we will fallback to sender GPU -> host SHM -> receiving GPU. The current `_restore_tensor_device` does not account for the host SHM -> receiving GPU part. As a result, GPU tensors reconstructed by the SHM relay remained on CPU.

## Modifications

  - Add CommRouter.receive_device, which exposes the receiving stage’s process-local device based on its configured gpu_id.
  - Add _resolve_tensor_destination to separate placement policy from tensor copying:
      - CPU-origin tensors remain on CPU.
      - Accelerator-origin tensors are placed on the receiving stage’s accelerator.
      - If no receiver device is provided, the tensor remains on the transport device.
 - Add a regression test covering a mixed CPU/GPU payload through the real SHM relay.
This does not change CUDA IPC behavior. Direct CUDA IPC payloads bypass the SHM restoration path, while relay-backed tensors that are already on the correct GPU make the final device operation a no-op.

## Related Issues

https://github.com/sgl-project/sglang-omni/issues/862

## Checklist

- [x] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/comm/engine.py
+++ b/sglang_omni/comm/engine.py
@@ -168,7 +168,12 @@ class CommEngine:
         request_id: str,
         data_ref: DataRef,
     ) -> StagePayload:
-        return await stage_io.read_payload(relay, request_id, data_ref)
+        return await stage_io.read_payload(
+            relay,
+            request_id,
+            data_ref,
+            destination_device=self.router.receive_device,
+        )
 
     async def send_stream_chunk(
         self,
--- a/sglang_omni/comm/router.py
+++ b/sglang_omni/comm/router.py
@@ -57,6 +57,14 @@ class CommRouter:
     def self_is_gpu(self) -> bool:
         return self.gpu_id is not None
 
+    @property
+    def receive_device(self) -> torch.device:
+        """Process-local device on which this stage consumes GPU tensors."""
+
+        if self.gpu_id is None:
+            return torch.device("cpu")
+        return torch.device("cuda", self.gpu_id)
+
     def is_local_object(self, target: str) -> bool:
         return target in self.same_process_targets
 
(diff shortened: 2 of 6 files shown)


ppppqp:
@JingwenGu0829 Could you take a look? Thanks!!

JingwenGu0829:
Hi, This PR appears to add support for a transport combination that the communication routing does not produce. 

CommRouter routes GPU-origin payloads between GPU stages through CUDA IPC, including mixed CPU/CUDA payloads. 

The regression test bypasses CommRouter and manually forces TransportKind.SHM instead of producing it in a real path.

Do we actually have a transportation path of GPU-SHM-CPU-SHM-GPU implemented in communication. If no such path exists, this is testing and fixing an unsupported state rather than a real communication path, which I guess is over protection?

JingwenGu0829:
Hi additionally I saw you upstreamed a PR for supporting AMD rocm. Is this pr also related to AMD support? If so we can talk further/maybe you can clarify further in your pr since I know a few about AMD communcation.

Since in designing this communication engine we didn't expect to give gpu-gpu payload a fallback on cpu shm

ppppqp:
> Hi additionally I saw you upstreamed a PR for supporting AMD rocm. Is this pr also related to AMD support? If so we can talk further/maybe you can clarify further in your pr since I know a few about AMD communcation.
> 
> Since in designing this communication engine we didn't expect to give gpu-gpu payload a fallback on cpu shm

Yep it was discovered when experimenting AMD support with shm relay as fallback. Agreed that it might be overprotection if it's a misues of shm relay, since in the end for ROCm we should rely on HIP IPC. We can close this if gpu-gpu payload should never fallback on cpu shm

Thanks for reviewing! Will also appreciate your review on the HIP IPC relay implementation (in future)

Issue #1296, opened 2026-08-03
[ASR] Fun-ASR serving speed on the CI host collapsed: the 128 to 139 req/s calibration band is no longer reachable


JiaxinD:
# [ASR] Fun-ASR serving speed on the CI host collapsed: the 128 to 139 req/s calibration band is no longer reachable

Since #1260 merged, every PR whose ASR stage 2 draws Fun-ASR fails the speed gate (floor 115.343). A strict recalibration today at the same commit never got close to the old band. All three data sources side by side:

| Session (all at `fc08b9a9`, sglang 0.5.16) | Throughput per round (req/s) | Latency mean (s) |
|---|---|---|
| #1260 calibration session, 2026-07-31 | 136.5 / 139.2 / 128.2 / 100.3 / 133.2 | 0.232 / 0.227 / 0.247 / 0.315 / 0.237 |
| CI failures, 2026-08-02 (runs 30736628336, 30726960557) | 92.0 and 91.6 | 0.345 and 0.344 |
| Recalibration, 2026-08-02 (12 rounds, full table in #1295) | 82.5 / 74.8 / 86.2 / 81.8 / 36.4 / 87.2 / 79.7 / 39.7 / 78.2 / 70.4 / 41.8 / 71.7 | 0.36 to 0.45 clean, 0.75 to 0.87 under load |
| Dose-response probes, 2026-08-03 (1-min load recorded per round) | 91.6 at load 35-57 / 81.1 at 32-44 / 53.4 at 103-174 / 26.9 at 152-163 | 0.344 / 0.388 / 0.589 / 1.177 |

Best of 12 rounds today is 87.2, which is 37 percent below the #1260 session median. Today's failing CI 92s are the fastest observations of the day, so the gate did not get unlucky; the band moved.

## What the data settles

* **Not a repo code delta.** Same commit, same pins, same venv path, same cards as standard calibration.
* **Not just instantaneous contention.** Rounds at 1-min load 20 to 26 still cap at 87, roughly half the old band.
* **A second, recurring slow mode is real.** Every sub-45 round coincided with load above 120 from CI-lane bursts. Fun-ASR is host-bound and stage 2 shares this box, so this mode is an operating condition, not an outlier. #1260 sampled it once (the 100.3 round) but shipped 128.158 instead of that session's worst-of-5 (100.346).

## Hypotheses, each with a cheap test

[...]

PR #1297, opened 2026-08-03
[CI] Persist the FlashInfer JIT cache in the per-PR CI home

JiaxinD:
# [CI] Persist the FlashInfer JIT cache in the per-PR CI home

## Motivation

Since the SGLang 0.5.16 bump, GPU jobs on the self-hosted runner keep recompiling the FlashInfer cutlass kernel set. Observed live while investigating #1296: host 1-min load 209 with about 85 parallel `cc1plus` / `cicc` / `ptxas` processes from a single job compiling `fused_moe_90` kernels into `/root/.cache/flashinfer/0.6.14/...`, even though the CI image ships 184 prewarmed objects for that same op family. The prewarm misses, and because `FLASHINFER_WORKSPACE_BASE=/root` points inside the job container, the compiled result is discarded when the container exits, so the next job pays the same multi-minute full-width storm again.

These storms are the load spikes behind the slow mode documented in #1296, and they perturb every host-bound speed gate on the runner (Fun-ASR stage 2, the TTS MPS stage, and the calibration sessions that set their references).

## Change

One step in `omni-setup`: seed `${OMNI_CI_HOME}/.cache/flashinfer` from the image's prewarmed cache on first use, then set `FLASHINFER_WORKSPACE_BASE=${OMNI_CI_HOME}`. A prewarm miss now compiles at most once per CI home instead of once per job, and the cache is reaped together with the home by the existing cleanup.

The seed uses copy-then-atomic-rename so a concurrent sibling job never sees a half-seeded cache; the loser of the race cleans up its temp copy and uses the winner's cache.

## Verification

* FlashInfer resolves the workspace as `$FLASHINFER_WORKSPACE_BASE/.cache/flashinfer/<version>/<arch>` (checked inside the CI image: base `/tmp/fitest` resolves to `/tmp/fitest/.cache/flashinfer/0.6.14/90a`), so the seeded path is exactly what the loader reads.
* Seed logic race-tested with two concurrent seeders: exactly one wins, no temp leftovers, a second pass is a no-op.
* End-to-end cache behavior across containers, measured with the smallest JIT module (`gen_act_and_mul_module("silu")`, `skip_prebuilt=False`) on the CI image: import-only baseline 7.9s; first container with a persistent workspace 10.5 to 14.3s (baseline plus one compile); second container on the same workspace 6.8 to 8.2s (compile fully skipped); two consecutive containers with the current `/root` workspace 14.2s and 14.7s (compile paid every time). The storming `fused_moe` cutlass set uses the same cache keying, so its 30 to 60 minute per-job compile collapses to once per CI home the same way.

## Out of scope

The prewarm gap itself: the image's prewarmed cache does not cover the cutlass MOE kernel set at all, which is why every venv misses (details in #1296). Extending the image prewarm removes even the once-per-home compile; this PR bounds the damage either way.


.github/actions/omni-setup/action.yaml
```diff
@@ -23,9 +23,23 @@ runs:
...
+        # FlashInfer does not use XDG_CACHE_HOME. Its JIT cache must survive
+        # the job container: on a miss it recompiles cutlass kernels for
+        # minutes at full host width, and under /root that work dies with the
+        # container, so every job on the runner pays it again (#1296). Seed
+        # the persistent CI home from the image's prewarmed cache once, then
+        # keep the workspace there. The copy and rename keep a concurrent
+        # sibling job from seeing a half-seeded cache.
+        if [ ! -d "${OMNI_CI_HOME}/.cache/flashinfer" ] && [ -d /root/.cache/flashinfer ]; then
+          mkdir -p "${OMNI_CI_HOME}/.cache"
+          seed="${OMNI_CI_HOME}/.cache/.flashinfer-seed-$$"
```
stephenkgli: `seed=...-$$` is unique only within the current PID namespace. Docker job containers have separate PID namespaces while sharing /data/omni-ci, so different containers can produce the same $$.

.github/actions/omni-setup/action.yaml
```diff
@@ -23,9 +23,23 @@ runs:
...
+          if cp -a /root/.cache/flashinfer "${seed}" \
+              && mv -T "${seed}" "${OMNI_CI_HOME}/.cache/flashinfer" 2>/dev/null; then
```
stephenkgli, line 40: If `cancel-in-progress: true` interrupts cp, it may leave a partial staging directory. If a later container reuses the same PID:
1. `cp -a source existing-seed` creates `seed/flashinfer/....`
2. The staging root retains the previous partial `0.6.14/...` contents.
3. `mv -T `publishes this malformed directory as the final cache.

.github/actions/omni-setup/action.yaml
AkazaAkane: Agree, need a unique cross container path like seed="$(mktemp -d "${OMNI_CI_HOME}/.cache/.flashinfer-seed.XXXXXXXX")"

JiaxinD: Done, switched to mktemp -d staging in a211c077.

.github/actions/omni-setup/action.yaml
JiaxinD: Done in a211c077: mktemp staging is never reused, so an interrupted copy just leaves an inert dir that is reaped with the CI home.

JiaxinD:
ASR stage 2 red is the known Fun-ASR floor issue (#1296): measured 90.8 vs the unreachable 115.3 floor, WER clean, unrelated to this change.

.github/actions/omni-setup/action.yaml
```diff
@@ -23,9 +23,26 @@ runs:
...
-        # architecture-specific prewarmed cache under /root/.cache/flashinfer.
-        echo "FLASHINFER_WORKSPACE_BASE=/root" >> "${GITHUB_ENV}"
+        # FlashInfer does not use XDG_CACHE_HOME. Its JIT cache must survive
+        # the job container: on a miss it recompiles cutlass kernels for
+        # minutes at full host width, and under /root that work dies with the
+        # container, so every job on the runner pays it again (#1296). Seed
+        # the persistent CI home from the image's prewarmed cache once, then
+        # keep the workspace there. The copy and rename keep a concurrent
+        # sibling job from seeing a half-seeded cache.
+        if [ ! -d "${OMNI_CI_HOME}/.cache/flashinfer" ] && [ -d /root/.cache/flashinfer ]; then
```
AkazaAkane, line 33: The cold-home ordering currently discards the seed immediately after copying it.

omni-setup seeds ${OMNI_CI_HOME}/.cache/flashinfer before dependency reconciliation. In the setup job, install-deps: true then invokes reconcile_omni_ci_env.sh; for a new or invalid home, that calls prepare_omni_venv.sh, which performs:

rm -rf "${OMNI_CI_HOME}"
mkdir -p "${OMNI_CI_HOME}"

The current CI run demonstrates this sequence: it logs Seeded FlashInfer JIT cache, then enters the full-rebuild path immediately afterward. The first downstream job seeds it again, so the cross-job JIT reuse still works, but every cold PR performs one complete cp -a that is immediately discarded.

Could we move the FlashInfer seeding after the reconcile/recovery step? This is a small fix and ensures that the persistent cache is initialized once on the cold path instead of copied twice.

PR #1260, opened 2026-08-01
Calibration 07 31

zhaochenyang20:
# Threshold changes — full CI recalibration after SGLang 0.5.16 bump

Calibration commit `c49126610845e814659622cd30c094d841c8256a` (branch `calibration_07_31`).
Policy: strict worst-of-5, 5/5 complete observations per stage, decision = `full`.
Values are **pre-slack raw references**; CI assertion slack stays in the tests.

`Δ%` is written-vs-previous. `spread` is (max-min)/|median| over the 5 observations.

### Correctness / quality (26)

| stage | symbol | dir | before | after | Δ% | spread | runs |
|---|---|---|---|---:|---:|---:|---|
| aishell4_long_diarization | `AISHELL4_LONG_DELTA_CER_PERCENT_REF` | loosens | 0.260558 | **0.362564** | +39.1% | 6% | 0.3409 0.3615 0.3626 0.3576 0.3404 |
| seedtts_fun_en_wer | `FUN_ASR_EN_SAMPLE_WER_MAX` | tightens | 0.3077 | **0.2858** | -7.1% | 0% | 0.2857 0.2857 0.2857 0.2857 0.2857 |
| multi_speaker_stream_diarization | `MOSS_TD_STREAM_DELTA_CER_PERCENT_REF` | loosens | 7.25181 | **7.42183** | +2.3% | 2% | 7.314 7.245 7.422 7.249 7.309 |
| aishell4_long_diarization | `AISHELL4_LONG_SPEAKER_TIMESTAMP_DER_PERCENT_REF` | loosens | 9.55717 | **9.70072** | +1.5% | 1% | 9.685 9.655 9.666 9.701 9.651 |
| multi_speaker_stream_diarization | `MOSS_TD_STREAM_CP_CER_PERCENT_REF` | loosens | 13.0498 | **13.223** | +1.3% | 1% | 13.13 13.06 13.22 13.07 13.12 |
| aishell4_long_diarization | `AISHELL4_LONG_CP_CER_PERCENT_REF` | loosens | 14.0463 | **14.2065** | +1.1% | 0% | 14.2 14.2 14.21 14.21 14.2 |
| seedtts_fun_zh_wer | `FUN_ASR_ZH_CORPUS_WER_MAX` | loosens | 0.0139 | **0.014** | +0.7% | 2% | 0.01378 0.01361 0.01364 0.0139 0.01361 |
[...]

PR #1298, opened 2026-08-03
[CI] Raise talker_ar mem fraction

db-ol:
## Motivation

Qwen3-Omni CI stage 2 (TTS speed + WER + SIM) failed on speed in 6 consecutive attempts and 14 completed measurements on #1034, while WER and speaker similarity passed every time. The cleanest attempt ran with no sibling jobs on the runner and still missed the gates by 1 to 4 percent (qps 7.053 vs 7.11, output tok per req 7.0 vs 7.3, latency 2.098 vs 2.0, rtf 0.6555 vs 0.65). The P95 references from #1260 appear to reflect a host state faster than what the runner sustains in practice. Full data: https://github.com/sgl-project/sglang-omni/pull/1034#issuecomment-5159677395

Separately, the TP2 talker_ar server crashes at startup intermittently since the SGLang 0.5.16 bump (#1183). Draft weights are now counted in the KV budget and the error reports minimum viable mem fraction 0.2008 against the configured 0.20, so startup survives or dies depending on GPU state.

## Modifications

- Widen the stage 2 voice cloning slack from the 0.875/1.125 defaults to 0.8/1.25, following the same per call override precedent as AISHELL4 long in the ASR CI. New gates: throughput_qps 6.5, output_tok_per_req_s 6.6, latency_mean_s 2.2, rtf_mean 0.73. Every idle and light load measurement from the #1034 data passes these with 5 to 10 percent margin, heavy load runs still fail and stay covered by the retry wrapper. Scope is limited to `VC_NON_STREAM_THRESHOLDS`, no other stage or metric changes.
- Raise `QWEN3_OMNI_TP2_TALKER_MEM_FRACTION` from 0.20 to 0.21, above the reported minimum viable 0.2008.

A proper recalibration of the stage 2 references can supersede the slack widening whenever the CI owners want to run one, happy to help with that.

## Related Issues

Follows up on the CI findings discussed in #1034. Related: #1183, #1260.

## Accuracy Test

Not applicable, threshold and test config constants only.

## Benchmark & Profiling

Threshold derivation verified against the 14 field measurements recorded in the #1034 comment thread.

## Checklist

- [ ] Add unit tests.


edwingao28:
Hi @db-ol thanks for separating the two issues here. 

For the stage 2 performance gates, widening the slack seems reasonable as an immediate unblock, but I am not sure it should be the final fix.

 #1260 did more than update the calibrated thresholds it also introduced destructive-round rejection. Slow or contended
rounds are excluded and replenished before the worst-of-N values are written.
This can make the resulting references represent a cleaner/faster host state than the shared runners used by normal PR CI. The repeated measurements in #1034 appear consistent with this distribution mismatch.

I will replace this PR with a proper recalibration that:

1. runs under conditions representative of the actual PR CI runners;
2. compares the default destructive-round rejection results with a run using `--no-destructive-rejection`;
3. records which rounds were rejected and why
4. replaces the stage-specific slack override with newly calibrated references if the data confirms the mismatch

The recalibration should address the underlying threshold methodology and solve this issue

db-ol:
Thanks @edwingao28! The destructive rejection explanation fits the data better than my host state guess, and today's run supports it: even with the widened 0.8/1.25 slack, stage 2 missed again under load (best try qps 6.321 vs 6.5), so widening slack just chases the tail of the same shifted distribution. Your recalibration plan sounds right to me, and the 14 measurements in the #1034 thread are yours to use as a field distribution reference for step 2.

Two questions:
1. Should this PR stay open as a temporary unblock while the recalibration is in flight, or close now and let affected PRs rerun against current gates?
2. The talker_ar mem fraction bump to 0.21 is independent of the threshold question. Should I split it into its own PR so it does not get lost when this one is replaced?

edwingao28:
> Thanks @edwingao28! The destructive rejection explanation fits the data better than my host state guess, and today's run supports it: even with the widened 0.8/1.25 slack, stage 2 missed again under load (best try qps 6.321 vs 6.5), so widening slack just chases the tail of the same shifted distribution. Your recalibration plan sounds right to me, and the 14 measurements in the #1034 thread are yours to use as a field distribution reference for step 2.
> 
> Two questions:
> 
> 1. Should this PR stay open as a temporary unblock while the recalibration is in flight, or close now and let affected PRs rerun against current gates?
> 2. The talker_ar mem fraction bump to 0.21 is independent of the threshold question. Should I split it into its own PR so it does not get lost when this one is replaced?

@db-ol 
1. yeah let's delete this part in this PR
2. Actually, I think the cleanest option is to repurpose #1298 as the standalone talker_ar memory-fraction fix rather than opening another PR

edwingao28 (approved):

PR #1304, opened 2026-08-03
Fix missing CPU operator events in scheduler-thread profiling      

estellaliu233:
## Summary

Route the opt-in Torch profiler lifecycle through the model scheduler thread.

Kineto CPU operator callbacks are thread-local. The profiler was started from
the stage asyncio/control-plane thread, while AR model execution runs on a
pre-existing scheduler thread. CUDA activities were recorded, but CPU-side Aten
operators from the AR execution path were absent.

Behavior is unchanged unless `SGLANG_TORCH_PROFILER_SCHEDULER_THREAD=1` is set.

A validation trace captured after routing the lifecycle through the scheduler
thread recorded roughly 494k Aten operator events, against none before.

## Design

`TorchProfiler` is a process-wide singleton, but one OS process hosts several
stages on a single asyncio loop (see `_run_process`). Ownership is therefore a
property of the **process**, not of a stage. `_resolve_torch_profiler_ownership`
computes it per `StageWorkerProcessSpec` and stamps each stage with one of three
modes:

| Mode | When | Behavior |
| --- | --- | --- |
| `scheduler_owner` | stage has `runtime.torch_profiler_owner=true` | drives the singleton via the scheduler admin queue |
| `scheduler_sibling` | colocated with an owner | does not touch the singleton at all |
| `direct` | flag off, or no owner in this process | pre-existing direct start/stop |

The sibling rule matters in practice: MOSS-TTS-Local colocates `preprocessing`
with the owning `tts_engine` in `process="pipeline"`. If a sibling handled the
broadcast first it would create the singleton on the control thread, and the
owner's start would then no-op on it — producing exactly the CPU-operator-less
trace this change exists to fix.

The `direct` fallback keeps a partial rollout safe: a process with no configured
owner still emits its own trace rather than silently emitting none.

## Guarantees

**Never runs on the wrong thread.** Two layers, because one is not enough:

- `Stage.start` spawns the scheduler thread and returns without waiting, so a
  control message can arrive before the thread publishes its id.
  `wait_until_scheduler_thread_ready` absorbs that startup race.
- `OmniScheduler.start` clears `_scheduler_thread_id` in its `finally` but leaves
  `_running` set until `stop` runs, so an exited scheduler loop leaves a durable
  window in which `_should_enqueue_admin` reports "run inline". A readiness check
  on the caller side cannot close this — the loop can exit right after it. The
  profiler therefore bypasses the generic `admin()` inline fallback and uses a
  strict path that **fails** rather than running on the caller's thread.

**Off the message path.** The readiness wait and the admin handoff each block
for up to 30s. Awaiting them from a message handler would keep `Stage.run` from
reaching its next `control_plane.recv()`, parking this stage's submits and acks
behind a profiler request. Handlers return immediately and the work runs on a
serialized worker, so a stop can never overtake the start it belongs to.

**Bounded, and no late execution.** Profiler admin requests use a 30s timeout
instead of the 300s generic default, so a dead scheduler cannot pin a stage's
control thread. The request also carries an absolute deadline: `_enqueue_admin`
reports a timeout but leaves the item queued, so a scheduler that drains it late
drops it instead of leaving an orphan profiler the caller was already told had
failed. Full CPU profiling is expensive enough that an orphan degrades serving.

**Failures do not kill the serving process.** `ProfilerControlClient` broadcasts
start/stop as fire-and-forget PUSH messages with no response channel, and an
exception escaping a stage message handler tears down every stage in that
process. Control failures are logged, not raised — a missing trace must not cost
the serving process. Note that this means a failed start is visible in stage
logs rather than at the HTTP call site; the broadcast API has no way to return
it.

## Changes

- Add `runtime.torch_profiler_owner` to identify the stage owning the
  process-wide `TorchProfiler` singleton; set it on the `tts_engine` stage of
  MOSS-TTS and MOSS-TTS-Local.
- Resolve ownership per process and propagate `torch_profiler_process_has_owner`
  through `StageLaunchConfig` to `Stage`.
- Add scheduler-thread profiler control to `OmniScheduler`, routed through the
  existing admin queue, behind a `SchedulerThreadProfilerControl` capability
  rather than a generic `admin` attribute check.
- Reject multiple profiler owners in one process; reject
  `SGLANG_TORCH_PROFILER_SCHEDULER_THREAD=1` with no owner configured anywhere
  (the flag would be a no-op).
- Fix `TorchProfiler.start` referencing `rank` before `rank = cls._get_rank()`.
  The old `not TorchProfiler.is_active()` guard made that branch unreachable;
  routing through the admin queue made it live, so any second start while a
  profiler was running raised `UnboundLocalError` — both the idempotent return
  and the restart warning. Same-`run_id` start is now genuinely idempotent.

The profiler admin action names remain internal to `OmniScheduler`; they are not
exposed as public administrative protocol operations.

## Testing

Added unit tests covering:

- same-`run_id` start is idempotent; different `run_id` restarts; `run_id=None`
  restarts (all three previously raised `UnboundLocalError`)
- owner delegates; colocated sibling never touches the singleton, including the
  worst-case ordering where the sibling handles the broadcast first
- ownerless process keeps the direct path
- ownership stamping is per process
- `TorchProfiler.start`/`stop` execute on the scheduler thread and not on the
  caller's, asserted on real thread ids through the real admin queue
- scheduler not yet ready, and scheduler exited after the readiness check —
  neither falls back to an inline run
- expired request is dropped when drained late; in-time request still runs
- admin failure surfaces to the caller and does not tear down the stage
- flag off keeps pre-existing behavior; `enable_torch=False` starts nothing
- the flag declared in pipeline/stage `env` config still requires an owner, and
  an explicit off in the launcher env is not overridden by config
- profiler delegation does not stall this stage's submits and acks, and queued
  operations run in arrival order

Also validated on a real MOSS-TTS-Local deployment using an NVIDIA H200.

  - Verified `TorchProfiler.start/stop` ran on the AR scheduler thread, not the stage control thread.
  - Three concurrent real TTS requests completed successfully with non-empty audio.
  - Same-run duplicate start was idempotent.
  - Both the owner process and ownerless vocoder switched from run-1 to run-2 and exported traces under run-2.
  - Owner AR traces contained 194k–333k CPU Aten events from real model execution.
  - Profiler control requests returned in milliseconds without blocking request processing.
  - Immediate stop followed by shutdown exported valid traces from both processes; the worker exited in about 7 seconds.
  - Flag-off direct profiling and `enable_torch: false` event-only recording remain compatible.

## Notes

This is an observability fix only. Full CPU profiling can introduce substantial
overhead and is not intended to be enabled during normal serving.

`_enqueue_admin` has no cancellation for non-profiler admin actions either — the
same late-execution shape applies to `update_weights_from_disk` and friends at
the 300s default. That is pre-existing and out of scope here; the deadline added
in this PR is scoped to the profiler path.

### DIFF
--- a/sglang_omni/pipeline/mp_runner.py
+++ b/sglang_omni/pipeline/mp_runner.py
@@ -32,6 +32,7 @@ from sglang_omni.pipeline.runtime_config import (
     build_comm_config,
     prepare_pipeline_runtime,
 )
+from sglang_omni.pipeline.stage.runtime import SCHEDULER_THREAD_PROFILER_ENV
 from sglang_omni.pipeline.stage_workers import (
     StageGroup,
     StageLaunchConfig,
@@ -200,9 +201,8 @@ def _resolve_torch_profiler_ownership(groups: list[StageGroup]) -> None:
     - a process without an owner: every stage keeps the pre-existing direct
       path, so a partial rollout never silently drops that process's trace.
 
-    A globally ownerless pipeline under
-    ``SGLANG_TORCH_PROFILER_SCHEDULER_THREAD=1`` is rejected, since the flag
-    would then be a no-op.
+    A globally ownerless pipeline with scheduler-thread profiling requested is
+    rejected, since the flag would then be a no-op.
     """
 
     owners: list[str] = []
@@ -222,11 +222,34 @@ def _resolve_torch_profiler_ownership(groups: list[StageGroup]) -> None:
                 stage_spec.torch_profiler_process_has_owner = bool(process_owners)
             owners.extend(process_owners)
 
-    if os.environ.get("SGLANG_TORCH_PROFILER_SCHEDULER_THREAD") == "1" and not owners:
+    if _scheduler_thread_profiling_requested(groups) and not owners:
         raise ValueError(
-            "SGLANG_TORCH_PROFILER_SCHEDULER_THREAD=1 requires at least one "
+            f"{SCHEDULER_THREAD_PROFILER_ENV}=1 requires at least one "
             "stage with runtime.torch_profiler_owner=true"
         )
+
+
+def _scheduler_thread_profiling_requested(groups: list[StageGroup]) -> bool:
+    """Whether any worker will see the scheduler-thread flag set.
+
+    Checking ``os.environ`` alone only covers the launcher's own environment.
+    The flag can also be declared in pipeline or stage ``env`` config, which
+    ``_patched_spawn_env`` injects at spawn time -- after this validation runs.
+    Reading only the launcher env would let an ownerless pipeline start with the
+    flag on, where it silently degrades to the pre-existing direct path.
+    """
+
+    launcher_value = os.environ.get(SCHEDULER_THREAD_PROFILER_ENV)
+    if launcher_value is not None:
+        # Spawn defaults only apply to keys absent from the launcher env, so
+        # this value is what every worker ends up seeing.
+        return launcher_value == "1"
+    return any(
+        stage_spec.env_defaults.get(SCHEDULER_THREAD_PROFILER_ENV) == "1"
+        for group in groups
+        for process_spec in group.process_specs
+        for stage_spec in process_spec.stage_specs
+    )
 
 
 def _attach_process_memory_fraction_defaults(groups: list[StageGroup]) -> None:
--- a/sglang_omni/pipeline/mp_runner.py
+++ b/sglang_omni/pipeline/mp_runner.py
@@ -204,10 +204,7 @@ def _validate_torch_profiler_owners(groups: list[StageGroup]) -> None:
                 )
             owners.extend(process_owners)
 
-    if (
-        os.environ.get("SGLANG_TORCH_PROFILER_SCHEDULER_THREAD") == "1"
-        and not owners
-    ):
+    if os.environ.get("SGLANG_TORCH_PROFILER_SCHEDULER_THREAD") == "1" and not owners:
         raise ValueError(
             "SGLANG_TORCH_PROFILER_SCHEDULER_THREAD=1 requires at least one "
             "stage with runtime.torch_profiler_owner=true"
--- a/sglang_omni/pipeline/mp_runner.py
+++ b/sglang_omni/pipeline/mp_runner.py
@@ -180,14 +180,30 @@ def _build_stage_groups(
             )
         )
     groups.extend(tp_groups)
-    _validate_torch_profiler_owners(groups)
+    _resolve_torch_profiler_ownership(groups)
     _attach_process_memory_fraction_defaults(groups)
 
     return groups
 
 
-def _validate_torch_profiler_owners(groups: list[StageGroup]) -> None:
-    """Require one process-local owner at most for the profiler singleton."""
+def _resolve_torch_profiler_ownership(groups: list[StageGroup]) -> None:
+    """Resolve profiler ownership per OS process and stamp it on each stage.
+
+    ``TorchProfiler`` is a process-wide singleton, but one OS process can host
+    several stages sharing a single asyncio loop (see ``_run_process``). So
+    ownership is a property of the process, not of a stage:
+
+    - a process with an owner: only that owner drives the singleton, through
+      its scheduler thread. Its colocated siblings must keep their hands off,
+      otherwise whoever handles the broadcast first wins the race and the
+      profiler ends up started (or stopped) on the wrong thread.
+    - a process without an owner: every stage keeps the pre-existing direct
+      path, so a partial rollout never silently drops that process's trace.
+
+    A globally ownerless pipeline under
+    ``SGLANG_TORCH_PROFILER_SCHEDULER_THREAD=1`` is rejected, since the flag
+    would then be a no-op.
+    """
 
     owners: list[str] = []
     for group in groups:
@@ -202,6 +218,8 @@ def _validate_torch_profiler_owners(groups: list[StageGroup]) -> None:
                     f"Process {process_spec.process_name!r} has multiple Torch "
                     f"profiler owners: {process_owners}"
                 )
+            for stage_spec in process_spec.stage_specs:
+                stage_spec.torch_profiler_process_has_owner = bool(process_owners)
             owners.extend(process_owners)
 
     if os.environ.get("SGLANG_TORCH_PROFILER_SCHEDULER_THREAD") == "1" and not owners:
--- a/sglang_omni/pipeline/mp_runner.py
+++ b/sglang_omni/pipeline/mp_runner.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import asyncio
 import logging
 import multiprocessing
+import os
 import socket
 from typing import Any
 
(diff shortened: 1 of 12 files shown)


sglang_omni/pipeline/stage/runtime.py
```diff
@@ -1585,7 +1582,26 @@ def _on_abort(self, request_id: str) -> None:
...
+            # Kineto CPU operator callbacks are thread-local. The AR scheduler
+            # runs in a dedicated thread that predates this control message, so
+            # starting here on the asyncio thread records CUDA activities but
+            # no CPU-side Aten operators. Route the opt-in profiler lifecycle
+            # through OmniScheduler's admin queue so start/stop execute on the
+            # actual compute thread. Ownership is explicit runtime intent,
+            # independent of a model-specific stage name. Sibling stages still
+            # participate in the event recorder below.
+            if self._torch_profiler_owner:
+                response = self.scheduler.start_torch_profiler(template, run_id)
```
AkazaAkane: This newly makes duplicate starts reach TorchProfiler.start() even when the same run is already active. Its idempotent branch references rank before rank = cls._get_rank(), so retrying /start_profile with the same run_id returns an admin failure with UnboundLocalError. Please move rank resolution before the active-profiler branch and add a test that starts the same run twice. Ideally preserve idempotent start semantics at this boundary as well.

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -1585,7 +1582,26 @@ def _on_abort(self, request_id: str) -> None:
 
     def _on_profiler_start(self, msg: ProfilerStartMessage) -> None:
         run_id = msg.run_id
-        if msg.enable_torch and not TorchProfiler.is_active():
+        scheduler_thread_torch = os.environ.get(_SCHEDULER_THREAD_PROFILER_ENV) == "1"
+        if msg.enable_torch and scheduler_thread_torch:
```
AkazaAkane: This branch consumes enable_torch for every stage when scheduler-thread mode is enabled, but non-owner stages do nothing. After process isolation, only the process containing torch_profiler_owner produces a Torch trace; other processes silently regress from the existing behavior. _validate_torch_profiler_owners() only requires one owner globally, so it does not catch this. Please model ownership per process and preserve direct profiling in processes without a scheduler-thread owner, or explicitly make owner-only profiling the supported behavior and update validation/API documentation accordingly.

AkazaAkane (commented):
The new Stage unit test proves only that a fake scheduler method was called; it does not prove that TorchProfiler.start() and stop() execute on the scheduler thread.

sglang_omni/pipeline/stage/runtime.py
estellaliu233: Thanks for catching this. I moved rank resolution before the active-profiler branch and added a regression test that starts the same run twice.

A repeated start with the same `run_id` now returns the existing trace path without stopping or rebuilding the profiler. I also verified this in the real MOSS-TTS-Local GPU E2E: the profiler object ID remained unchanged after the duplicate start, and serving continued successfully.

sglang_omni/pipeline/stage/runtime.py
estellaliu233: Thanks, I updated the implementation to resolve profiler ownership per OS process.

  The resulting behavior is:

  - The owner stage controls the process-wide singleton through its scheduler thread.
  - A colocated non-owner sibling never touches the singleton.
  - A process without an owner preserves the existing direct profiling path.

  I added ownership propagation through the process and stage launch specs, validation for multiple owners in one process, and regression coverage for owner,
  sibling, and ownerless processes.

  I also validated this with the real MOSS-TTS-Local topology: `preprocessing` and the owning `tts_engine` were colocated in one process, while `vocoder` ran in a
  separate ownerless process. Both processes exported traces. The ownerless process also switched from `fix-run-1` to `fix-run-2` after the follow-up fix.

estellaliu233:
> The new Stage unit test proves only that a fake scheduler method was called; it does not prove that TorchProfiler.start() and stop() execute on the scheduler thread.

  `test_scheduler_admin_runs_torch_profiler_on_the_scheduler_thread` now tests execution thread. 
  
  I also verified the complete Stage-to-scheduler path with the real
  MOSS-TTS-Local GPU E2E topology:

  - control thread ID: `132119562298496`
  - scheduler thread ID: `131970185033280`
  - `TorchProfiler.start()` thread ID: `131970185033280`
  - `TorchProfiler.stop()` thread ID: `131970185033280`

sglang_omni/pipeline/mp_runner.py
```diff
@@ -178,11 +180,55 @@ def _build_stage_groups(
...
+            if len(process_owners) > 1:
+                raise ValueError(
+                    f"Process {process_spec.process_name!r} has multiple Torch "
+                    f"profiler owners: {process_owners}"
+                )
+            for stage_spec in process_spec.stage_specs:
+                stage_spec.torch_profiler_process_has_owner = bool(process_owners)
+            owners.extend(process_owners)
+
+    if os.environ.get("SGLANG_TORCH_PROFILER_SCHEDULER_THREAD") == "1" and not owners:
```
JiaxinD: This only sees the launcher env: the flag can also arrive via pipeline/stage `env_defaults` injected at spawn, which passes this check and then silently no-ops in ownerless workers. Consider validating the merged env too.

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -1578,9 +1662,69 @@ def _on_profiler_start(self, msg: ProfilerStartMessage) -> None:
...
+        return False
+
+    def _start_torch_profiler_on_scheduler(
+        self, template: str, run_id: str | None
+    ) -> None:
+        """Delegate the start to the scheduler thread, or give up loudly."""
+
+        # Stage.start does not wait for the scheduler thread, so an early start
+        # is a legitimate startup race worth waiting out.
+        if not self._scheduler_thread_ready_for_profiler(
```
JiaxinD, line 1798: This ready wait plus the 30s enqueue wait run synchronously on the stage control loop, stalling submits/acks while they block. `_run_admin_operation` offloads via `run_in_executor` for the same reason, worth doing here too?

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -52,6 +52,28 @@
 
 GetNextFn = Callable[[str, Any], str | list[str] | None]
 GetStreamDoneTargetsFn = Callable[[str, Any], str | list[str] | None]
+_SCHEDULER_THREAD_PROFILER_ENV = "SGLANG_TORCH_PROFILER_SCHEDULER_THREAD"
```
JiaxinD: Neither this env var nor `runtime.torch_profiler_owner` is in the env table of docs/developer_reference/profiler.md yet, worth a docs entry since this is a user-facing opt-in.

sglang_omni/pipeline/mp_runner.py
estellaliu233: Validation now considers the effective launcher, pipeline, and stage environment, including `env_defaults`, so scheduler-thread mode cannot silently start in an ownerless configuration.  I also added regression coverage for config-provided flags and launcher-env precedence.

sglang_omni/pipeline/stage/runtime.py
estellaliu233: Profiler operations now run through a per-stage serialized worker queue, so `_handle_message` returns immediately while start/stop ordering is preserved. A real H200 E2E confirmed profiler control returned in milliseconds while concurrent TTS requests continued successfully without ACK timeouts or scheduler failures.

sglang_omni/pipeline/stage/runtime.py
estellaliu233: Added in the profiler.md.

PR #1306, opened 2026-08-03
[NPU] Add hardware abstraction layer and Ascend NPU support for Qwen3-Omni text only mode

Makcum888e:
## Motivation

Initial support for Ascend NPU

Only Qwen3-Omni with `--text-only` supported at the moment

## Modifications

added hardware abstraction layer (same way with SGL-Diffusion and SGLang) (current_platform)
replaced hardcoded `torch.cuda` with general `torch.get_device_module()`

## Accuracy Test

```
sgl-omni serve --model-path /home/weights/Qwen/Qwen3-Omni-30B-A3B-Instruct/ --thinker-cuda-graph off --thinker-gpus 2,3 --thinker-tp-size 2 --text-only
```

```
curl -X POST http://localhost:8000/v1/chat/completions -H "Content-Type: application/json" -d '{"model": "qwen3-omni", "messages": [{"role": "user", "content": "Say the word \"one\" exactly ten times in a row. Do not add any other words. Example: one, one, one... Repeat it exactly 10 times."}], "temperature": 0.01, "top_p": 0.1, "presence_penalty": 0.0, "frequency_penalty": 0.0, "modalities": ["text", "audio"]}' -o output.json
```

## Checklist

- [ ] Add unit tests.

## CI


ping1jing2:
LGTM

sglang_omni/platforms/interface.py
```diff
@@ -0,0 +1,14 @@
+"""
+SGLang Omni Hardware Platform Abstraction.
+
+Defines Omnilatform — the base class for Omni platform backends.  OmniPlatform inherits DeviceMixin for shared device operations and adds Omni-specific subsystem factory methods, capability flags, and configuration lifecycle hooks.
```
Ratish1:
```suggestion
Defines OmniPlatform — the base class for Omni platform backends.  OmniPlatform inherits DeviceMixin for shared device operations and adds Omni-specific subsystem factory methods, capability flags, and configuration lifecycle hooks.
```

sglang_omni/comm/router.py
```diff
@@ -142,6 +142,8 @@ def outbound_payload(self, target: str, payload: Any) -> TransportKind:
             if self.self_is_gpu and target in self.gpu_stage_names:
                 return TransportKind.CUDA_IPC
             return TransportKind.SHM
+        if "npu" in devices and devices <= {"cpu", "npu"}:
```
Ratish1: SHM can remain the selected fallback for NPU, but the RFC places transport selection behind a typed Omni policy based on placement, payload and platform capabilities. can we route this through that policy rather than adding an NPU-specific branch to the common router?

sglang_omni/platforms/__init__.py
```diff
@@ -0,0 +1,150 @@
...
+
+current_platform: OmniPlatform
+
+
+def __getattr__(name: str):
+    """Lazy initialization of current_platform on first access."""
+    if name == "current_platform":
+        global _current_platform
+        if _current_platform is None:
+            _current_platform = _resolve_platform()
```
Ratish1: The RFC defines the parent-resolved platform spec as Omni’s deployment-level authority and carries that identity and worker placement through the stage launch configuration. here we instead resolve  a new platform singleton independently in every process. Reusing SGLang’s device mixins is appropriate, but Omni should retain ownership of its process lifecycle. should we resolve the deployment platform in the parent and construct or seed the worker’s Omni platform from the serialized launch specification rather than repeating detection here?

sglang_omni/pipeline/stage_workers.py
```diff
@@ -795,7 +794,7 @@ def get_stage_process_env(
     env: Mapping[str, str] | None = None,
 ) -> dict[str, str]:
     """Return per-process env overrides needed before TP child startup."""
-    if spec.tp_size <= 1:
+    if spec.tp_size <= 1 or current_platform.is_npu():
```
Ratish1: in RFC, it keeps worker environment setup behind the Omni platform boundary. Could we move this
decision into a platform hook? The NPU implementation can still return {}, while CUDA keeps its existing environment mapping.

Ratish1 (commented):
Overall, looks good @Makcum888e . left some few comments, lmk what do you think

sglang_omni/pipeline/stage_workers.py
Makcum888e: moved to `current_platform`

sglang_omni/comm/router.py
Makcum888e: removed for now, as it no need for text-only mode, will be solved with talker support

sglang_omni/platforms/__init__.py
Makcum888e: https://github.com/sgl-project/sglang-omni/pull/1306/changes/9a16d508b5e3edb59907e458b26b62d061a8eb45 please, take a look at this commit, if it's ok for you

Ratish1 (approved):

sglang_omni/platforms/__init__.py
```diff
@@ -0,0 +1,157 @@
...
+
+_current_platform: OmniPlatform | None = None
+
+
+def _is_cuda_available() -> bool:
+    return bool(torch.cuda.is_available() and torch.version.hip is None)
+
+
+def _is_npu_available() -> bool:
+    return bool(torch.npu.is_available())
```
Hayden727: Stock PyTorch builds do not define `torch.npu`; when CUDA is unavailable this fallback raises `AttributeError` while importing `sglang_omni.platforms`. Read the optional attribute with a local `try`/`except AttributeError` and add a regression test for a missing `torch.npu`.

sglang_omni/platforms/cpu.py
```diff
@@ -0,0 +1,9 @@
+from sglang.srt.platforms.cpu import CpuDeviceMixin
+
+from sglang_omni.platforms.interface import OmniPlatform
+
+
+class CPUOmniPlatform(CpuDeviceMixin, OmniPlatform):
```
Hayden727, line 6: `_get_worker_process_env()` calls this hook for every TP stage, but `CPUOmniPlatform` inherits `OmniPlatform.get_stage_process_env()`, which raises `NotImplementedError`. Make non-CUDA platforms return `{}` (in the base or here) so CPU TP workers can spawn.

sglang_omni/platforms/__init__.py
```diff
@@ -0,0 +1,157 @@
...
+from sglang_omni.platforms.npu import NPUOmniPlatform
+from sglang_omni.platforms.spec import ResolvedPlatformSpec
+
+logger = logging.getLogger(__name__)
+
+_current_platform: OmniPlatform | None = None
+
+
+def _is_cuda_available() -> bool:
+    return bool(torch.cuda.is_available() and torch.version.hip is None)
```
Hayden727: This explicitly excludes HIP (`torch.version.hip is None`), but there is no ROCm fallback, so an AMD host resolves to the base `OmniPlatform` and later device methods fail. Preserve SGLang's ROCm platform/mixin path—or delegate discovery to SGLang—and cover a HIP-enabled probe.

sglang_omni/platforms/__init__.py
```diff
@@ -0,0 +1,157 @@
...
+    raise RuntimeError(
+        f"Multiple platform plugins activated: {names_str}. "
+        f"Set SGLANG_PLATFORM to select one."
+    )
+
+
+def _load_platform_class(qualname: str) -> type:
+    """Load an SRTPlatform subclass from its fully-qualified class name."""
+    cls = pkgutil.resolve_name(qualname)
+    if not isinstance(cls, type) or not issubclass(cls, OmniPlatform):
```
Hayden727: SGLang platform entry points are specified to return `SRTPlatform` subclasses, but this loader requires `OmniPlatform`; a conforming OOT plugin is rejected, and the child spec below would also collapse its identity to a built-in/base platform. Accept or wrap `SRTPlatform`, serialize the exact resolvable class/plugin identity, and add a parent-to-child round-trip test.

cocoa-liu:
@Makcum888e Hi, are you currently a Huawei employee? Our team is also working on NPU support for Qwen3-Omni. We would appreciate the opportunity to discuss and exchange ideas with you.
You can reach me at: 834237195@qq.com

PR #1315, opened 2026-08-03
[Comm] Add paged KV cache transfer infrastructure

JingwenGu0829:
## Summary

This PR adds a model-neutral, page-oriented KV cache transfer path to the Omni communication layer.

- Define KV pool/layout, source lease, receiver allocation, and paged-relay contracts.
- Add the control-plane `prepare -> ready -> data ready -> ack` protocol.
- Add the receiver `reserve -> direct copy -> commit/abort` lifecycle to `CommEngine` and the stage runtime.
- Add a same-node CUDA IPC backend that copies source pages directly into receiver-owned destination pages with SGLang's `transfer_kv_per_layer_mla` kernel.
- Add protocol and end-to-end lifecycle tests.

## Lifecycle

```mermaid
sequenceDiagram
    participant U as "UND / Prefill"
    participant C as "CommEngine"
    participant G as "GEN / Decode"

    U->>U: UND prefill finishes
    U->>U: Resolve source page indices
    U->>U: Pin source pages with a lease

    U->>C: send_kv_pages(request, source pages, target pool)
    C->>G: KVTransferPrepare
    G->>G: Reserve destination pages
    G-->>C: KVTransferReady(destination indices/ref)

    C->>G: Direct source pages → destination pages
    Note over C,G: One data movement, no pack or staging pool
    C->>G: DataReady(KV_PAGES)
    G->>G: Wait for transfer completion
    G->>G: Commit destination allocation
    G-->>C: DataAck

    C-->>U: send_kv_pages completes
    U->>U: Release source-page lease
    U->>G: Send ordinary StagePayload
    G->>G: Claim committed KV by request_id
    G->>G: Start decode or denoise
```

1. Source and destination register their long-lived KV pools.
2. The source pins the request's source pages and sends `KVTransferPrepareMessage`.
3. The receiver reserves destination pages and returns their indices plus a backend capability.
4. The source publishes its registered storage capability and sends a `DataRef(kind=KV_PAGES, layout=PAGED)`.
5. The receiver copies directly from source pages into the reserved destination pages, then commits the allocation.
6. The receiver sends the existing `DataAckMessage`; only then does the sender release its source-page lease.

The prepare state is authoritative. The later `DataRef.object_id` is the transfer ID, so the data-ready message does not duplicate pool layouts, page indices, or request metadata.

## Data movement

For CUDA IPC, this is a receiver-pull operation:

- KV storage handles are exported once per receiver registration and cached.
- A per-transfer interprocess CUDA event orders producer work before the copy.
- The SGLang page-copy kernel reads the selected source pages and writes the reserved destination pages directly.
- The ordinary payload staging pool and generic `put_async` path are not used, avoiding a pack/copy/unpack double copy.

The protocol is transport-neutral; this PR provides the same-node CUDA IPC implementation. Cross-node Mooncake/NIXL support and model-specific SGLang KV-pool adapters are follow-up work.

## Developer interface

A model or future PD-disaggregation integration provides cache ownership and page mappings. It does **not** select a relay, construct `KVTransferPrepareMessage` / `KVTransferReadyMessage`, build a `DataRef`, or manage the final ACK. `CommRouter` selects the physical transport from the compiled sender/receiver topology.

### Register long-lived KV pools

Both stages describe their existing page-addressable cache storage once during initialization:

```python
from sglang_omni.comm import KVBufferRegion, KVPool

comm.register_kv_pool(
    KVPool(
        pool_id="decode_kv",
        layout_id="model-layout-v1",
        page_size=page_size,
        buffers=tuple(
            KVBufferRegion(
                name=name,
                tensor=tensor,
                bytes_per_page=bytes_per_page,
            )
            for name, tensor, bytes_per_page in model_kv_buffers
        ),
    )
)
```

The pool wraps model-owned tensors; the communication layer does not allocate or replace the model's KV cache.

### Register receiver-owned allocation

The destination scheduler already owns page allocation and request lifetime. It exposes those operations through its existing `KVReceiver` interface, and the communication layer registers that interface directly:

```python
comm.register_kv_receiver(
    "decode_kv",
    gen_scheduler.kv_transfer_receiver,
)
```

This is symmetric with the source exposing its model-owned KV pool to the communication layer. `kv_transfer_receiver` is not a new stage or a communication-owned allocator. `CommEngine` calls its `reserve(request)` before the copy, `commit(request, destination)` after the copy completes, and `abort(request, destination, error)` if the transfer fails or is cancelled. The scheduler remains the owner of allocation, page-table publication, and page reclamation.

### Send request pages

The source resolves and pins its model-specific page indices, then makes one call:

```python
await comm.send_kv_pages(
    control_plane=stage.control_plane,
    request_id=request_id,
    source_pool_id="prefill_kv",
    source_page_indices=tuple(source_page_indices),
    target_pool_id="decode_kv",
    from_stage="prefill",
    to_stage="decode",
    target_endpoint=stage.endpoints["decode"],
    lease=source_page_lease,
    metadata={"sequence_length": sequence_length},
)
```

The optional `KVPageLease` pins source pages against eviction or reuse. `CommEngine` releases it only after receiver ACK, or when prepare/copy fails or is cancelled. Completion of `send_kv_pages()` therefore means the receiver has committed the destination pages and the sender may safely reuse its pages.

### Adding a PD transport

The model-facing API above should remain unchanged. Today paged KV transfer accepts only topology-selected `CUDA_IPC`; any other selected transport fails explicitly. A future cross-node PD implementation adds the paged pool registration, destination preparation, put, and get operations to the topology-selected Mooncake/NIXL relay, then enables that transport in `CommEngine`. PD model code should never pass a relay or backend handle explicitly.


sglang_omni/comm/engine.py
```diff
@@ -170,6 +204,37 @@ async def read_payload(
...
         return await stage_io.read_payload(relay, request_id, data_ref)
 
+    async def read_data(
+        self,
+        *,
+        relay: Relay,
+        request_id: str,
+        data_ref: DataRef,
+    ) -> StagePayload | None:
+        """Read one non-stream ``DataReady`` object.
```
Ratish1: nit: remove backticks from docstrings

sglang_omni/comm/engine.py
```diff
@@ -225,7 +290,289 @@ async def read_stream_chunk(
...
+                target_pool_id=target_pool_id,
+                from_stage=from_stage,
+                to_stage=to_stage,
+                target_endpoint=target_endpoint,
+                metadata=metadata,
+                transfer_id=transfer_id,
+            )
+        finally:
+            if lease is not None:
+                lease.release()
```
Ratish1: This `finally` runs when `_watch_pending()` times out after DataReady has already been published. At that point the receiver's asynchronous copy may be active; a local timeout is not a receiver terminal event. Releasing the lease then allows the scheduler to reuse and overwrite pages while the receiver is still reading them.

Transfer lease ownership to the pending operation once DataReady is published. Early failures that cannot have exposed the source may release it immediately, but an ambiguous ACK timeout or post-publication cancellation must keep the pages and IPC references pinned until a late terminal ACK arrives or teardown has first stopped the receiver.

sglang_omni/comm/kv_transfer.py
```diff
@@ -0,0 +1,110 @@
...
+    bytes_per_page: int
+
+    @property
+    def page_count(self) -> int:
+        return (
+            int(self.tensor.numel()) * int(self.tensor.element_size())
+        ) // self.bytes_per_page
+
+    def byte_view(self) -> torch.Tensor:
+        return self.tensor.view(torch.uint8).reshape(-1)
```
Ratish1: For a non-contiguous tensor, view(torch.uint8) may reject the layout or the following reshape(-1) may allocate a contiguous temporary. Exporting that temporary reads a snapshot rather than the live
KV allocation; using it as a destination writes data that is then discarded.

Require and retain an aliasing byte view when the pool is registered. KVBufferRegion should also reject non-positive bytes_per_page and byte lengths that are not exactly divisible by it. Validate the CUDA copy kernel’s bytes_per_page % 8 == 0 requirement separately in CudaIpcRelay before launching any copy.

sglang_omni/relay/cuda_ipc.py
```diff
@@ -902,11 +951,196 @@ async def get_async(
...
+            destination_page_indices,
+            dtype=torch.int64,
+            device=destination_device,
+        )
+        stream = torch.cuda.current_stream(destination_device)
+        with torch.cuda.device(destination_device), torch.cuda.stream(stream):
+            stream.wait_event(ready_event)
+            from sgl_kernel import transfer_kv_per_layer_mla
+
+            for source, target in zip(source_buffers, destination.buffers, strict=True):
```
Ratish1: A failure while building a later target byte view or validating/launching a later kernel occurs after earlier buffer kernels may already be queued. `get_kv_pages()` then raises without returning a `CudaIpcGetOperation`; the Stage sends a negative ACK immediately, and the sender can release its lease while those earlier kernels still read the source.

Precompute every source/target view and validate every kernel argument before this loop. If an exception can still occur after the first launch, wait for the stream tail before allowing the negative ACK to become terminal.

sglang_omni/comm/engine.py
```diff
@@ -225,7 +290,289 @@ async def read_stream_chunk(
     ) -> tuple[torch.Tensor, dict[str, Any] | None]:
         return await stage_io.read_stream_chunk(relay, data_ref)
 
+    def register_kv_pool(self, pool: KVPool) -> None:
+        self._kv_pools[pool.pool_id] = pool
```
Ratish1, line 299: The API says a long-lived pool is registered once, but this assignment accepts different storage under an existing `pool_id`. The CUDA relay then keeps the original registration ID and cached storage handles, so the two layers disagree about which allocation that ID names.

Repeated registration of the same storage must remain harmless because the transfer path re-registers with the relay, but a different storage identity under the same ID should fail immediately. Storage replacement does not need to be supported.

sglang_omni/comm/kv_transfer.py
```diff
@@ -0,0 +1,110 @@
...
+    """One page-addressable tensor in a registered KV pool."""
+
+    name: str
+    tensor: torch.Tensor
+    bytes_per_page: int
+
+    @property
+    def page_count(self) -> int:
+        return (
+            int(self.tensor.numel()) * int(self.tensor.element_size())
```
Ratish1: nit: are these `int()` conversions redundant?, `Tensor.numel()` and `Tensor.element_size()` already return python integers

sglang_omni/pipeline/stage/runtime.py
```diff
@@ -324,6 +326,10 @@ async def _handle_message(self, msg: Any) -> None:
             await self._on_submit(msg)
         elif isinstance(msg, DataAckMessage):
             self._comm.ack_transfer(msg)
+        elif isinstance(msg, KVTransferReadyMessage):
+            self._comm.kv_transfer_ready(msg)
+        elif isinstance(msg, KVTransferPrepareMessage):
```
Ratish1, line 331: Only a TP stage's leader is connected to the external control plane, and `KVTransferPrepareMessage` / `DataReadyMessage` are not fanned out to followers. Each TP rank nevertheless owns a different sharded KV pool. A transfer involving `tp_size > 1` would therefore prepare and copy rank 0 only.

can add rank-to-rank prepare/copy/ACK handling for every shard, or reject paged transfer when either endpoint is TP.

sglang_omni/comm/engine.py
JingwenGu0829: CommEngine.register_kv_pool() is initialization-only and is not called by the transfer path. Per-transfer re-registration happens only on KVTransferRelay, using the same long-lived KVPool retained by CommEngine. Replacing KV storage under an existing pool ID is outside the supported lifecycle; model/cache replacement tears down and recreates the stage, engine, and relay together. Therefore the engine and relay cannot diverge through the production path.

sglang_omni/pipeline/stage/runtime.py
JingwenGu0829: Yeah this is a critical point. Created #1383 to work on this (kv cache transfer under TP). Still WIP.

tests/unit_test/pipeline/test_kv_transfer.py
```diff
@@ -0,0 +1,395 @@
...
+                "transfer_info": {"size": len(source_page_indices)},
+                "fake_kv": True,
+                "key": "kv-put",
+            },
+            self.log,
+        )
+        self.put_ops.append(op)
+        return op
+
+    async def get_kv_pages(
```
Ratish1, line 59: The paged transfer test never executes the backend. _PagedRelay.get_kv_pages() only records the mapping and returns, while the existing CudaIpcRelay tests cover put_async() / get_async(), not register_kv_pool() / put_kv_pages() / get_kv_pages(). Storage-handle import, storage offsets, the multi-buffer kernel loop, page mapping, and untouched destination pages can therefore regress without
failing CI.

Add a two-process CudaIpcRelay paged round-trip using a non-identity mapping. Assert both copied pages and untouched destination pages, and include multiple buffers and a nonzero storage-
offset view.

Ratish1 (approved):
LGTM

PR #1321, opened 2026-08-04
[CI] Pin perf-gate test sessions to OMNI_CI_CPUSET when set

JiaxinD:
## Motivation

Perf gates on the shared runner measure host-bound serving stacks, and their throughput tracks host load (#1296). Measured on the CI host with the Fun-ASR SeedTTS EN speed stage (fresh server per round, CI semantics): with 64 synthetic busy loops as background load, an unpinned run keeps 55 to 62 percent of its quiet-host throughput, while a run pinned to 24 reserved cores with the load placed off those cores keeps about 92 percent. Per-request server CPU time itself inflates 1.5x when load spreads across the machine (SMT sibling occupancy plus all-core frequency drop), which is why gate floors calibrated in quiet windows fail under storms.

## Change

An opt-in autouse session fixture in `tests/test_model/conftest.py`: when `OMNI_CI_CPUSET` is set (Linux cpulist, e.g. `0-23,64-87`), the pytest process pins itself with `sched_setaffinity`, and every child it spawns, the managed router, its workers, and the bench client, inherits the affinity. Unset means no behavior change anywhere. Invalid specs fail fast with `ValueError`.

Assigning the per-lane cpulists stays runner-side configuration (each GPU lane exports its own `OMNI_CI_CPUSET`); this PR ships only the mechanism. Calibration runs should export the same variable so floors are measured under the same policy.

## Verification

* `tests/unit_test/ci/test_cpuset_pinning.py`: cpulist parsing (ranges, singles, rejects empty/inverted/garbage), no-op without the env, and a real `sched_setaffinity` round-trip (skipped off-Linux).
* Throughput retention numbers above from the pinning 2x2 on the CI host, replicated twice per cell (A/B rounds), loadavg and PSI recorded per round.

## Evidence (from the pinning experiment, Fun-ASR seedtts en, c32, fresh server per round)

| condition | qps (A / B) | server CPU ms/req |
|---|---|---|
| quiet baseline | 93.6 / 90.9 | 46 to 47 |
| 64 busy loops, unpinned | 48.2 / 57.5 | 72 to 83 |
| 64 busy loops, pinned 0-23, load on other socket | 83.5 / 81.9 | 51 to 52 |

### DIFF
--- a/tests/test_model/conftest.py
+++ b/tests/test_model/conftest.py
@@ -16,6 +16,48 @@ if TYPE_CHECKING:
 
     from tests.utils import ServerHandle
 
+
+def _parse_cpuset(spec: str) -> set[int]:
+    """Parse a Linux cpulist such as "0-23,64-87" into a CPU id set."""
+    cpus: set[int] = set()
+    for part in spec.split(","):
+        part = part.strip()
+        if not part:
+            continue
+        if "-" in part:
+            lo_text, hi_text = part.split("-", 1)
+            lo, hi = int(lo_text), int(hi_text)
+            if lo > hi:
+                raise ValueError(f"Invalid OMNI_CI_CPUSET range {part!r}")
+            cpus.update(range(lo, hi + 1))
+        else:
+            cpus.add(int(part))
+    if not cpus:
+        raise ValueError(f"OMNI_CI_CPUSET {spec!r} selects no CPUs")
+    return cpus
+
+
+def _apply_omni_ci_cpuset() -> set[int] | None:
+    spec = os.environ.get("OMNI_CI_CPUSET", "").strip()
+    if not spec or not hasattr(os, "sched_setaffinity"):
+        return None
+    cpus = _parse_cpuset(spec)
+    os.sched_setaffinity(0, cpus)
+    return cpus
+
+
+@pytest.fixture(autouse=True, scope="session")
+def pin_omni_ci_cpuset() -> None:
+    """Pin the test session, and every server it spawns, to OMNI_CI_CPUSET.
+
+    The gates in this directory measure host-bound serving stacks, so on a
+    shared runner concurrent jobs inflate per-request CPU cost and shift
+    calibrated floors. Child processes inherit the affinity, which keeps the
+    managed router, its workers, and the bench client on the reserved cores.
+    """
+    _apply_omni_ci_cpuset()
+
+
 TTS_ALLOWED_CONCURRENCIES = (1, 2, 4, 8, 16)
 TTS_STAGE_NONSTREAM = "tts-stage-1-nonstream"
 TTS_STAGE_STREAM = "tts-stage-2-stream"
--- a/tests/test_model/conftest.py
+++ b/tests/test_model/conftest.py
@@ -23,7 +23,7 @@ def _parse_cpuset(spec: str) -> set[int]:
     for part in spec.split(","):
         part = part.strip()
         if not part:
-            continue
+            raise ValueError(f"Empty component in OMNI_CI_CPUSET {spec!r}")
         if "-" in part:
             lo_text, hi_text = part.split("-", 1)
             lo, hi = int(lo_text), int(hi_text)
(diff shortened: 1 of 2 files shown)


tests/test_model/conftest.py
```diff
@@ -16,6 +16,48 @@
...
+        raise ValueError(f"OMNI_CI_CPUSET {spec!r} selects no CPUs")
+    return cpus
+
+
+def _apply_omni_ci_cpuset() -> set[int] | None:
+    spec = os.environ.get("OMNI_CI_CPUSET", "").strip()
+    if not spec or not hasattr(os, "sched_setaffinity"):
+        return None
+    cpus = _parse_cpuset(spec)
+    os.sched_setaffinity(0, cpus)
```
stephenkgli: On Linux, `sched_setaffinity()` may silently intersect the requested CPU set with the CPUs allowed by the current cgroup or cpuset. This implementation returns the requested set without reading back the effective affinity, so a job requesting 24 CPUs could run on only 8 while appearing successfully pinned. That would invalidate performance-gate calibration. After setting affinity, call `os.sched_getaffinity(0)`, require an exact match, and fail fast with both the requested and effective masks when they differ.

tests/test_model/conftest.py
```diff
@@ -16,6 +16,48 @@
 
     from tests.utils import ServerHandle
 
+
+def _parse_cpuset(spec: str) -> set[int]:
+    """Parse a Linux cpulist such as "0-23,64-87" into a CPU id set."""
+    cpus: set[int] = set()
+    for part in spec.split(","):
+        part = part.strip()
+        if not part:
```
AkazaAkane, line 25: Silently skipping empty components means malformed specifications such as 0,,1, ,0-3, and 0-3, are accepted.

Since the fixture is meant to fail fast on invalid configuration, I think an empty component should raise ValueError instead. Please also add parameterized tests for leading, trailing, and repeated commas.

tests/test_model/conftest.py
AkazaAkane: Agreed. The important check is the effective mask returned by `os.sched_getaffinity(0)` after `sched_setaffinity`, not the requested set.

Linux may intersect the request with the process's existing affinity or cpuset-cgroup restrictions without raising an error. The fixture should fail if `effective != requested` and log the requested, previous allowed, and effective masks for CI diagnosis.

JiaxinD: Done, the fixture now reads back os.sched_getaffinity(0) and raises with the requested, previously allowed, and effective masks on mismatch.

tests/test_model/conftest.py
JiaxinD: Done, empty components now raise ValueError, with parameterized tests for leading, trailing, and repeated commas.

JiaxinD:
Runner setup guide for OMNI_CI_CPUSET on the h100 box (2 sockets, 128 logical CPUs, NUMA node0 = 0-31,64-95, node1 = 32-63,96-127):

| Runner | GPU lane | NUMA | Suggested OMNI_CI_CPUSET |
|---|---|---|---|
| actions-runner-h100 | 6,7 | node1 | `32-47,96-111` |
| actions-runner-h100-2 | 4,5 | node1 | `48-63,112-127` |
| actions-runner-h100-3 | 2,3 | node0 | `16-31,80-95` |
| actions-runner-h100-4 | spare, no fixed lane | | leave unset |

Each lane gets 16 physical cores plus their HT siblings on the same NUMA node as its GPUs. CPUs 0-15,64-79 stay free for the host, the autoscaler, and GPU 0/1 users.

To enable, per runner:
1. Append `OMNI_CI_CPUSET=<value>` to `/home/sglang-omni/<runner-dir>/.env` (owned by the runner user, no sudo needed).
2. No service restart needed: runners run as `Runner.Listener run --once`, so the next job picks up the new env.

Note: the extra runners get their GPU pair from the autoscaler lane lease. Pairs are stable in practice (h100-2 = 4,5, h100-3 = 2,3), but if a lease ever moves across NUMA nodes the cpuset must move with it. Folding the cpuset into the supervisor lease assignment is the clean long-term follow-up.

PR #1327, opened 2026-08-04
[Fix] realtime VAD dropping audio at adjacent turn boundaries(#1322)

### LINKED ISSUE #1322: [Bug] Realtime VAD can drop audio at adjacent turn boundaries
### Checklist

- [x] This is a bug report rather than a general usage question.
- [x] This issue is written in English.

### Description

The realtime server-side VAD path can lose audio at a turn boundary when one input audio chunk contains the end of one utterance, enough silence to close that turn, and the beginning of the next utterance.

In that case, turn segmentation correctly recognizes that the first utterance has ended, but committing that turn also discards the uncommitted audio suffix. A subsequent speech-start transition may still be emitted even though the corresponding audio has already been dropped, and its timestamp may be calculated against the wrong stream position.

This is a pre-existing realtime input/segmentation issue and is not introduced by PR #1314.

### Reproduction

Use server-side VAD with the default 500 ms silence window and submit a single audio chunk containing:

```text
speech → 500+ ms silence → speech
```

A deterministic synthetic sequence of one 32 ms speech frame, sixteen 32 ms silent frames, and another 32 ms speech frame reproduces the problem.

Observed event sequence:

```text
speech_started → speech_stopped → committed → speech_started
```

The second speech region is no longer present in the retained input audio, while its start event is still delivered with an incorrect timestamp.

This boundary can also occur with smaller client chunks when silence accumulated from previous chunks crosses the VAD threshold partway through the current chunk.

### Expected behavior

- Committing the completed utterance consumes only the finalized audio range.
- Any uncommitted suffix belonging to the next utterance remains available for segmentation.
- VAD transitions are processed exactly once and against consistent stream state.
- Reported timestamps remain monotonic and correspond to the original audio timeline.

### Impact

This can drop the beginning or entirety of the next user turn in half-duplex conversations. The same segmentation behavior also needs to be reliable before server-owned barge-in or full-duplex interaction is added.

### Suggested validation

Add regression coverage for a single input chunk producing:

```text
speech_started → speech_stopped → speech_started
```

The test should verify that the second utterance's audio is preserved, each turn is committed once, and timestamps remain correct across the boundary.

guozhihao-224:
## Motivation
* Realtime server-side VAD could lose audio at a turn boundary when one input chunk contained the end of one utterance, enough silence to close that turn, and the start of the next utterance.
* Committing the first turn cleared the entire input buffer and reset VAD, so the uncommitted suffix was dropped while a second `speech_started` could still be emitted with an incorrect timestamp.
* This breaks half-duplex multi-turn conversations and needs to be solid before barge-in / full-duplex work.
## Modifications
1. `sglang_omni/serve/realtime/session.py`
   * On auto-commit, consume only the finalized audio prefix via `consume_committed_prefix` instead of clearing the whole buffer and resetting VAD.
   * Keep the VAD timeline continuous across turns.
   * Track `vad_origin_samples` and `buffer_origin_samples` separately so byte offsets and event timestamps stay aligned after a partial commit.
   * `input_audio_buffer.clear` still clears the buffer and resets VAD.
2. `sglang_omni/serve/realtime/audio_buffer.py`
   * Add `drop_prefix` so commit can retain any uncommitted suffix.
3. `tests/unit_test/serve/test_realtime_vad_boundary.py`
   * Cover single-chunk `speech_started → speech_stopped → committed → speech_started`.
   * Verify the second utterance's PCM is retained, the second turn can still be committed, and timestamps remain monotonic.
   * Add a real Silero path for adjacent-turn preservation, plus clear-path reset coverage.
## Related Issues
Fixes #1322
## Accuracy Test
N/A
## Benchmark & Profiling
N/A
## Test plan
- [x] `pytest tests/unit_test/serve/test_realtime_vad_boundary.py -v` on this branch (3 passed)
- [x] Same test file against `main`: fails (commit still clears/resets; second-turn audio dropped)
- [ ] Optional manual realtime check: server VAD, speech → 500ms+ silence → speech in one stream
## Checklist
- [x] Add unit tests.


charliechenye:
Was looking at the issue too :)

Nice fix: the prefix only consumption and separate VAD/buffer origins appear to address the reported audio loss and timestamp corruption.

One additional case from #1322 still looks uncovered: the silence threshold being accumulated across two client appends.

The current fake VAD test injects `speech_started → speech_stopped → speech_started` from one `process()` call, and the Silero test also submits the complete `speech → 16 silent frames → speech` sequence in one `_append()`. Neither test exercises VAD state carried across calls, where the second call crosses the silence threshold and then begins the next turn.

Could we add a `StreamingVAD` case shaped roughly like this?

```text
append 1: speech_1 + 15 silent frames
          → speech_started only; no commit (below threshold)

append 2: 1 silent frame + speech_2
          → speech_stopped
          → committed 
          → speech_started
```

The test should verify that:

* no commit occurs after append 1;
* append 2 crosses the 500 ms threshold and produces the expected event ordering;
* only the finalized prefix is consumed;
* `speech_2` remains in the retained buffer;
* the second `speech_started` timestamp remains on the original monotonic session timeline;
* automatic commit does not reset VAD.

It would be even stronger to subsequently stop and commit turn 2, then verify that its committed WAV contains the retained `speech_2` PCM, rather than checking only its duration.

This matters because the cross append shape is the normal network streaming form of the bug, and exercises state that the same append tests do not.

Thanks a lot. 
And great job 💯

guozhihao-224:
> Was looking at the issue too :)
> 
> Nice fix: the prefix only consumption and separate VAD/buffer origins appear to address the reported audio loss and timestamp corruption.
> 
> One additional case from #1322 still looks uncovered: the silence threshold being accumulated across two client appends.
> 
> The current fake VAD test injects `speech_started → speech_stopped → speech_started` from one `process()` call, and the Silero test also submits the complete `speech → 16 silent frames → speech` sequence in one `_append()`. Neither test exercises VAD state carried across calls, where the second call crosses the silence threshold and then begins the next turn.
> 
> Could we add a `StreamingVAD` case shaped roughly like this?
> 
> ```
> append 1: speech_1 + 15 silent frames
>           → speech_started only; no commit (below threshold)
> 
> append 2: 1 silent frame + speech_2
>           → speech_stopped
>           → committed 
>           → speech_started
> ```
> 
> The test should verify that:
> 
> * no commit occurs after append 1;
> * append 2 crosses the 500 ms threshold and produces the expected event ordering;
> * only the finalized prefix is consumed;
> * `speech_2` remains in the retained buffer;
> * the second `speech_started` timestamp remains on the original monotonic session timeline;
> * automatic commit does not reset VAD.
> 
> It would be even stronger to subsequently stop and commit turn 2, then verify that its committed WAV contains the retained `speech_2` PCM, rather than checking only its duration.
> 
> This matters because the cross append shape is the normal network streaming form of the bug, and exercises state that the same append tests do not.
> 
> Thanks a lot. And great job 💯

@charliechenye  Thanks for the careful review — and for catching the cross-append gap!

Agreed: the single-process() / single-_append() cases don’t exercise silence accumulating across client chunks, which is the normal network streaming shape of #1322.

I’ve added test_silero_silence_threshold_across_appends covering exactly the shape you suggested:

append 1: speech_1 + 15 silent frames → speech_started only; no commit
append 2: 1 silent frame + speech_2 → speech_stopped → committed → speech_started
It also checks that only the finalized prefix is consumed, speech_2 remains in the buffer, the second speech_started stays on the monotonic session timeline, and auto-commit does not reset VAD. After stopping turn 2, the second committed WAV is asserted to contain the retained speech_2 PCM (not just duration).

Happy to adjust further if anything still looks thin.

tests/unit_test/serve/test_realtime_vad_boundary.py
```diff
@@ -0,0 +1,251 @@
...
+    assert len(commits) == 2 and reset_calls == 0
+    assert second in _wav_pcm(commits[1][1])
+    await _stop(session)
+
+
+@pytest.mark.asyncio
+async def test_clear_resets_vad_timeline():
+    vad = FakeVAD([])
+    session, sent, _ = _session(vad)
+    session.audio_buffer.append_b64(_b64(_pcm(2, 1000)))
```
edwingao28, line 271: Nothing here catches an empty append quietly advancing `samples_consumed` can you add a case asserting it and `buffer_origin_samples` stay put?

sglang_omni/serve/realtime/session.py
```diff
@@ -187,24 +190,35 @@ async def handle_vad_emit(self, emit: Any) -> None:
...
+        discarded = self.audio_buffer.num_samples
+        self.buffer_origin_samples += discarded
+        self.vad_origin_samples = self.buffer_origin_samples
         self.audio_buffer.clear()
         self.utterance_start_byte = None
         self.utterance_item_id = None
         self.vad.reset()
 
+    def consume_committed_prefix(self, end_byte: int) -> None:
+        self.audio_buffer.drop_prefix(end_byte)
```
edwingao28, line 230: decoded_len can be 0 when the client sends an empty audio payload. In that case, tail(0) evaluates to buf[-0:], which returns the entire buffer instead of empty bytes.
As a result, the VAD processes previously consumed audio again. This advances its sample position and can eventually cause the commit boundary to be clamped incorrectly and silently dropping a turn.

charliechenye:
Thanks for adding the cross append regression. This now covers the case I had in mind, including preserving `speech_2`, keeping timestamps monotonic, avoiding a VAD reset, and verifying the actual PCM in the second committed WAV. Looks good to me.

One small observation on the code quality bot comments: I believe the repeated `reset_calls == 0` assertions are intentional rather than redundant. Each check occurs after another `_append(...)`, which may trigger a new `speech_stopped` and auto-commit path (and the nonlocal _counting_reset()), so confirming that the reset count remains unchanged after each state transition still provides useful coverage.

I also agree with @edwingao28 ’s empty append finding. The `tail(0)` behavior appears to be pre-existing, since both `tail()` and the `tail(decoded_len)` call were already present. However, this PR makes the problem more prominent because an auto-commit can now retain a nonempty suffix while keeping the VAD timeline alive. An empty append could therefore feed that retained suffix through VAD again and advance its sample position without receiving new audio.

Would it make sense to fix the invariant directly in `RealtimeAudioBuffer.tail()` so that `tail(0)` returns `b""`?

```python
def tail(self, num_bytes: int) -> bytes:
    assert 0 <= num_bytes <= len(self.buf), "Invalid tail length"
    if num_bytes == 0:
        return b""
    return bytes(self.buf[-num_bytes:])
```

An early return in `handle_audio_append()` could still be added defensively, but handling zero correctly inside `tail()` seems useful because it makes the helper’s behavior correct for every caller.

Curious what you and @edwingao28 think.

sglang_omni/serve/realtime/session.py
charliechenye: Good catch. An early return in handle_audio_append() would address this call site, but making tail(0) return b"" seems cleaner because it preserves the helper’s expected contract for every caller.

That said, since the underlying tail(0) behavior predates this PR, would you prefer to keep the fix narrowly scoped to handle_audio_append()?

edwingao28:
@guozhihao-224 please fix unit tests failure

edwingao28:
> Thanks for adding the cross append regression. This now covers the case I had in mind, including preserving `speech_2`, keeping timestamps monotonic, avoiding a VAD reset, and verifying the actual PCM in the second committed WAV. Looks good to me.
> 
> One small observation on the code quality bot comments: I believe the repeated `reset_calls == 0` assertions are intentional rather than redundant. Each check occurs after another `_append(...)`, which may trigger a new `speech_stopped` and auto-commit path (and the nonlocal _counting_reset()), so confirming that the reset count remains unchanged after each state transition still provides useful coverage.
> 
> I also agree with @edwingao28 ’s empty append finding. The `tail(0)` behavior appears to be pre-existing, since both `tail()` and the `tail(decoded_len)` call were already present. However, this PR makes the problem more prominent because an auto-commit can now retain a nonempty suffix while keeping the VAD timeline alive. An empty append could therefore feed that retained suffix through VAD again and advance its sample position without receiving new audio.
> 
> Would it make sense to fix the invariant directly in `RealtimeAudioBuffer.tail()` so that `tail(0)` returns `b""`?
> 
> ```python
> def tail(self, num_bytes: int) -> bytes:
>     assert 0 <= num_bytes <= len(self.buf), "Invalid tail length"
>     if num_bytes == 0:
>         return b""
>     return bytes(self.buf[-num_bytes:])
> ```
> 
> An early return in `handle_audio_append()` could still be added defensively, but handling zero correctly inside `tail()` seems useful because it makes the helper’s behavior correct for every caller.
> 
> Curious what you and @edwingao28 think.

Yeah I will go with fixing tail(). Returning b"" for tail(0) matches the helper’s expected behavior, and this PR makes the existing issue reachable through the retained suffix, so fixing it here feels in scope. With that in place, I don’t think we need an additional early return in handle_audio_append().

edwingao28 (approved):

JingwenGu0829:
Rebased the branch to mitigate the CPU contention in CI test.

Issue #1302, opened 2026-08-03
[Feature] Add Fun-CosyVoice3 to SGLang-Omni.


PoTaTo-Mika:
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang-omni/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

Fun-CosyVoice3 is a strong text-to-speech model with fine-grained control, multilingual synthesis, and few-shot voice cloning. It has been adopted by academic and industrial applications.

This issue proposes native support for `FunAudioLLM/Fun-CosyVoice3-0.5B-2512` in SGLang-Omni, including continuous batching for speech-token generation, streaming Flow + HiFT decoding, and the OpenAI-compatible `/v1/audio/speech` endpoint.

Tracing #1331 

### Related resources

- Hugging Face repository: https://huggingface.co/FunAudioLLM/Fun-CosyVoice3-0.5B-2512
- Official GitHub repository: https://github.com/QwenAudio/CosyVoice
- [SGLang-Omni TTS model integration guide](../developer_reference/tts_model_integration.md)
- [SGLang-Omni reference encode service](../developer_reference/reference_encode_service.md)

### Proposed pipeline

```text
preprocessing -> reference_encode -> tts_engine -> audio_decode
```

Implementation is split into two parallel workstreams:

```text
LLM:    text + LLM prompt tokens -> generated speech-token stream
Decode: reference audio + speech-token stream -> 24 kHz PCM stream
```

The initial API scope is:

| Inputs | Mode |
|---|---|
| `ref_audio` and `ref_text` | Zero-shot voice cloning |
| `ref_audio` without `ref_text` | Cross-lingual voice cloning |
| `ref_audio` and instructions | `instruct2` fine-grained control |

The checkpoint does not include `spk2info.pt`, so preset named speakers are not part of the initial scope. Voice conversion and the alternative `llm.rl.pt` checkpoint can be handled separately after the main TTS path is stable.

### Shared contract

The two workstreams should agree on this contract before implementation:

```python
@dataclass
class CosyVoice3ReferenceArtifact:
    llm_prompt_tokens: Tensor
    flow_prompt_tokens: Tensor
    prompt_mel: Tensor
    speaker_embedding: Tensor
    reference_text: str | None


@dataclass
class CosyVoice3TokenChunk:
[...]

PR #1331, opened 2026-08-04
[Feat]Add Fun-CosyVoice3 to SGLang-Omni.

PoTaTo-Mika:
## Motivation

One of the WeChat group members required support for CosyVoice.

CosyVoice is also one of the most well known Text-to-speech models, which uses a two-stage generation method like index-tts.

## Modifications

Add a new model, modified documentations.

## Related Issues

#1302 

## Accuracy Test

Under construction.

## Benchmark & Profiling

Under construction.

## Checklist

- [ ] Add unit tests.

## CI


luobudakk:
One observation from my RTX 3060 6GB colocated smoke test:

The vocoder currently constructs a full `CosyVoice3(checkpoint)`, which reloads the whole stack on the same GPU as `tts_engine` and then fails during startup.

Would it make sense to load only the sliced flow + HiFT path for the vocoder stage, instead of instantiating full `CosyVoice3()`?

docs/cookbook/fun_cosyvoice3.md
```diff
@@ -0,0 +1,216 @@
...
+uv pip install sox onnxruntime openai-whisper
+```
+
+Clone the CosyVoice repository with its Matcha-TTS submodule and add both
+packages to `PYTHONPATH`:
+
+```bash
+git clone --recursive https://github.com/FunAudioLLM/CosyVoice.git /opt/CosyVoice
+export PYTHONPATH="/opt/CosyVoice:/opt/CosyVoice/third_party/Matcha-TTS:$PYTHONPATH"
+```
```
ping1jing2, line 35: suggest to use variable here, it's convient for user to custom path.

```bash
COSYVOICE_PATH=/opt/CosyVoice
git clone --recursive https://github.com/FunAudioLLM/CosyVoice.git ${COSYVOICE_PATH}
export PYTHONPATH="${COSYVOICE_PATH}:${COSYVOICE_PATH}/third_party/Matcha-TTS:$PYTHONPATH"
```

sglang_omni/models/fun_cosyvoice3/stages.py
```diff
@@ -0,0 +1,234 @@
...
+        # note: the HiFT vocoder requires a minimum mel length for its
+        # convolutional upsampling. Pad short token sequences with a
+        # zero token so the vocoder has enough frames to process.
+        min_tokens = 10
+        if token.shape[1] < min_tokens:
+            pad_len = min_tokens - token.shape[1]
+            pad = torch.zeros(1, pad_len, dtype=token.dtype, device=token.device)
+            token = torch.cat([token, pad], dim=1)
+
+        with torch.autocast(device_type="cuda", enabled=self._fp16):
```
Makcum888e: please, use `current_platform.device_type` instead of `"cuda"`

sglang_omni/models/fun_cosyvoice3/stages.py
```diff
@@ -0,0 +1,234 @@
...
+    model_path: str,
+    *,
+    device: str = "cuda:0",
+    gpu_id: int | None = None,
+    dtype: str = "bfloat16",
+    max_batch_size: int = 8,
+    max_batch_wait_ms: int = 2,
+) -> SimpleScheduler:
+    if gpu_id is not None:
+        device = f"cuda:{gpu_id}"
```
Makcum888e: could we add new `current_platform.get_device_str(gpu_id)`?

ping1jing2 (approved):

Makcum888e (approved):

sglang_omni/models/fun_cosyvoice3/stages.py
```diff
@@ -0,0 +1,237 @@
...
+                torch.as_tensor(state.flow_embedding)
+                if state.flow_embedding is not None
+                else torch.zeros(1, 192)
+            )
+            wav = self._token2wav(
+                token=codes.unsqueeze(0),
+                prompt_token=prompt_token,
+                prompt_feat=prompt_feat,
+                embedding=embedding,
+                speed=state.speed,
```
Ratish1: speed is applied twice. this stage changes the mel length, then `/v1/audio/speech` passes the same value to the shared `encode_audio` path, which changes the waveform speed again. for example, `speed=2` produces roughly one quarter of the native duration instead of one half. remove the model-local speed change.

sglang_omni/models/fun_cosyvoice3/request_builders.py
```diff
@@ -0,0 +1,782 @@
...
+    temperature = float(gen_kwargs.get("temperature", 0.7)) if do_sample else 0.0
+    max_new_tokens = int(gen_kwargs.get("max_new_tokens", _DEFAULT_MAX_NEW_TOKENS))
+
+    sampling_params = SamplingParams(
+        max_new_tokens=max_new_tokens,
+        temperature=temperature,
+        top_p=float(gen_kwargs.get("top_p", 0.8)),
+        top_k=int(gen_kwargs.get("top_k", 20)),
+        repetition_penalty=float(gen_kwargs.get("repetition_penalty", 1.1)),
+        stop_token_ids=[EOS_ID],
```
Ratish1: cosyvoice3 treats all 200 ids in `[6561, 6761)` as terminal non-speech ids, not only `EOS_ID`. `_collect_tokens` drops these ids from the audio output, but the other 199 ids do not stop scheduling here, so generation can continue after the model has emitted a terminal token. use the full terminal range.

sglang_omni/models/fun_cosyvoice3/request_builders.py
```diff
@@ -0,0 +1,782 @@
...
+        raise RuntimeError(
+            "CosyVoice3 AR request builder requires a payload prepared by "
+            "preprocess_cosyvoice3_payload"
+        )
+
+    gen_kwargs = prepared.gen_kwargs
+    state = prepared.state
+    do_sample = bool(gen_kwargs.get("do_sample", True))
+    temperature = float(gen_kwargs.get("temperature", 0.7)) if do_sample else 0.0
+    max_new_tokens = int(gen_kwargs.get("max_new_tokens", _DEFAULT_MAX_NEW_TOKENS))
```
Ratish1: the fixed 2048 default drops cosyvoice3's generation-length contract. upstream suppresses stop tokens until `2 * target_text_tokens` and stops by `20 * target_text_tokens`; the target count excludes the reference prompt. preserve the target text token count before prompt concatenation, derive both bounds from it, cap the maximum at 2048, and still respect an explicit `max_new_tokens`. make sure the minimum suppresses all 200 stop ids rather than only a tokenizer eos id.

sglang_omni/models/fun_cosyvoice3/stages.py
```diff
@@ -0,0 +1,237 @@
...
+        prompt_token: torch.Tensor,
+        prompt_feat: torch.Tensor,
+        embedding: torch.Tensor,
+        speed: float = 1.0,
+    ) -> torch.Tensor:
+        device = next(self._flow.parameters()).device
+        # note: the HiFT vocoder requires a minimum mel length for its
+        # convolutional upsampling. Pad short token sequences with a
+        # zero token so the vocoder has enough frames to process.
+        min_tokens = 10
```
Ratish1: token 0 is a valid fsq speech token, not padding. adding it here synthesizes content that the ar stage did not generate. remove this padding and fail clearly when generation returns no usable speech tokens. if hift cannot accept the returned sequence, report a generation error instead of adding speech codes.

sglang_omni/models/fun_cosyvoice3/utils.py
```diff
@@ -0,0 +1,498 @@
...
+
+        if sample_rate != 16000:
+            raise ValueError(
+                f"Speech tokenizer expects 16kHz audio, got {sample_rate}Hz"
+            )
+
+        if audio.ndim == 1:
+            audio = audio.reshape(1, -1)
+
+        assert (
```
Ratish1: the 30-second limit applies to user input, so this should raise a clear `ValueError` instead of using `assert`. assertions can be disabled and currently turn an invalid reference into an internal assertion failure.

pyproject.toml
```diff
@@ -95,6 +95,11 @@ audar-tts = [
     "llama-cpp-python==0.3.34",
     "neucodec==0.0.6",
 ]
+fun-cosyvoice3 = [
```
Ratish1, line 98: `openai-whisper` and `onnxruntime` are already project dependencies, while the linux branch installs `onnxruntime-gpu` beside `onnxruntime`; both distributions export the same module and should not be installed together. remove these duplicate entries, or use this extra only for the actual compatible cosyvoice and matcha-specific dependencies and install it from the cookbook.

sglang_omni/models/fun_cosyvoice3/utils.py
```diff
@@ -0,0 +1,498 @@
...
+    def __init__(self, model_path: str, device: str = "cpu"):
+        import onnxruntime
+
+        option = onnxruntime.SessionOptions()
+        option.graph_optimization_level = (
+            onnxruntime.GraphOptimizationLevel.ORT_ENABLE_ALL
+        )
+        option.intra_op_num_threads = 1
+
+        providers = (
```
Ratish1: the base installation provides cpu `onnxruntime`, but this requests the cuda provider whenever the sglang device is cuda. do not install a second onnx runtime distribution just for reference preprocessing. use `CPUExecutionProvider`, as the speaker encoder below already does, or filter this list through `onnxruntime.get_available_providers()` before creating the session.

docs/cookbook/fun_cosyvoice3.md
```diff
@@ -0,0 +1,217 @@
...
+## Prerequisites
+
+Install `sglang-omni` by following [Installation](../get_started/installation.md).
+
+Fun-CosyVoice3 depends on the `cosyvoice` package (GitHub-only, no PyPI release) and several
+audio-processing libraries:
+
+```bash
+apt-get update && apt-get install -y sox
+uv pip install sox onnxruntime openai-whisper
```
Ratish1: `sox`, `onnxruntime`, and `openai-whisper` are already project dependencies. keep the system `sox` install, install the model's actual compatible cosyvoice and matcha-specific dependencies through the optional extra, and pin the cosyvoice commit and matcha submodule used for validation. do not install the upstream `requirements.txt` because its torch, torchaudio, transformers, and diffusers pins conflict with sglang-omni.

sglang_omni/models/fun_cosyvoice3/stages.py
```diff
@@ -0,0 +1,237 @@
...
+    model_path: str,
+    *,
+    device: str | None = None,
+    gpu_id: int | None = None,
+    dtype: str = "bfloat16",
+    max_batch_size: int = 8,
+    max_batch_wait_ms: int = 2,
+) -> SimpleScheduler:
+    device = resolve_device_spec(device, gpu_id)
+    checkpoint_dir = model_path
```
Ratish1: `model_path` can be a hugging face repo id. the tts stage resolves that id with the project's hugging face checkpoint resolver, but passing the raw id to `CosyVoice3` makes the vocoder resolve it through modelscope instead. resolve it with `sglang_omni.utils.checkpoint.resolve_checkpoint` here so both stages load the same local snapshot.

sglang_omni/models/fun_cosyvoice3/sglang_model.py
```diff
@@ -0,0 +1,126 @@
...
+    def forward(
+        self,
+        input_ids: torch.Tensor,
+        positions: torch.Tensor,
+        forward_batch: Any,
+        input_embeds: torch.Tensor = None,
+        get_embedding: bool = False,
+        pp_proxy_tensors: Optional[Any] = None,
+    ) -> Any:
+        fwd_mode = getattr(forward_batch, "forward_mode", None)
```
Ratish1: don't use `getattr` here. `forward_mode` is required on `ForwardBatch`.

sglang_omni/models/fun_cosyvoice3/model_runner.py
```diff
@@ -0,0 +1,149 @@
...
+        schedule_batch: Any,
+        requests: list,
+    ) -> None:
+        if result.next_token_ids is None:
+            return
+        token_ids = result.next_token_ids
+        if token_ids.ndim != 1:
+            token_ids = token_ids.reshape(-1)
+        for idx, sched_req in enumerate(requests):
+            token_id = int(token_ids[idx].item())
```
Ratish1: `.item()` inside this loop synchronizes the host with the gpu once per active request on every decode step. copy the token-id batch to cpu once, then append the cpu ids. the info log in the same token loop should also be removed.

sglang_omni/models/fun_cosyvoice3/stages.py
```diff
@@ -0,0 +1,237 @@
...
+        # The AR runner stores one-element tensors per step, which serialize as
+        # ``[num_tokens, 1]``. Flow consumes one unbatched token sequence here.
+        codes = torch.as_tensor(state.audio_codes, dtype=torch.long).reshape(-1)
+        return state, codes
+
+    async def decode_batch(
+        self, items: list[tuple[FunCosyVoice3State, torch.Tensor]]
+    ) -> list[tuple[Any, int]]:
+        results = []
+        for state, codes in items:
```
Ratish1: this loop runs one separate flow and hift call per request, so waiting to collect eight items adds queue delay without batching model execution. keep this pr small: set the vocoder batch size to 1, remove the batch wait, and set `supports_batch_vocoder=False` until one model call handles multiple requests.

benchmarks/eval/benchmark_tts_seedtts.py
```diff
@@ -111,6 +111,8 @@
 | Higgs TTS | EN, stream=False | 4.68%   | 4.16%               | 0.00%                 | 91.2%              | 1088/1088 | 0       | PR #534 [H200, full-set, c=16, CUDA Graph on, torch.compile off] |
 | Higgs TTS | ZH, stream=False | 1.14%   | 1.08%               | 0.00%                 | 2.7%               | 2020/2020 | 0       | PR #534 [H200, full-set, c=16, CUDA Graph on, torch.compile off] |
 | MOSS-TTS | EN, stream=False | 1.93%   | 1.98%               | 0.00%                 | 8.1%               | 1088/1088 | 0       | PR #609 [H100, full-set, c=16, token-count=auto] |
+| Fun-CosyVoice3-0.5B | EN, stream=False | 1.37% | 1.27%       | 0.00%                 | 5.1%               | 1088/1088 | 0       | PR #1331 [H200, full-set, c=16, voice-clone (ref_text)] |
+| Fun-CosyVoice3-0.5B | ZH, stream=False | 0.74% | 0.72%       | 0.00%                 | 2.6%               | 2020/2020 | 0       | PR #1331 [H200, full-set, c=16, cross-lingual (no ref_text)] |
```
Ratish1, line 115: the checked-in benchmark cli has no mode that keeps `ref_audio` while omitting `ref_text`: the default sends both, while `--no-ref-audio` removes both. what is the command or helper used for this benchmark or is there a missing cli option that we need to add here?

pyproject.toml
```diff
@@ -95,6 +95,15 @@ audar-tts = [
...
+    "conformer==0.3.2",
+    "HyperPyYAML==1.2.3",
+    "onnx==1.16.0",
```
Ratish1: `onnx` is not used by this serving path; it imports `onnxruntime` instead. remove `onnx==1.16.0`. a clean extra install is also missing `lightning` and `matplotlib`, which the pinned matcha modules import during model construction. add `lightning==2.6.5` and `matplotlib==3.11.1`.

sglang_omni/models/fun_cosyvoice3/sglang_model.py
```diff
@@ -0,0 +1,128 @@
...
+            hidden_states = hidden_states[last_indices]
+
+        logits = self.llm_decoder(hidden_states)
+        return LogitsProcessorOutput(
+            next_token_logits=logits, hidden_states=hidden_states
+        )
+
+    def load_weights(self, weights: Iterable[Tuple[str, torch.Tensor]]):
+        backbone_weights = []
+        custom_weights = {}
```
Ratish1: `custom_weights` is initialized but never read or updated in `load_weights`, so it has no effect on checkpoint loading. can we remove it?

docs/basic_usage/tts.md
```diff
@@ -41,7 +41,8 @@ for details.
 | [Qwen3-TTS Base](../cookbook/qwen3_tts.md) | `examples/configs/qwen3_tts_0_6b.yaml`, `examples/configs/qwen3_tts_1_7b.yaml` | Requires reference audio through `ref_audio` or `references[0].audio_path`. `language` defaults to `auto` |
 | [Qwen3-TTS CustomVoice](../cookbook/qwen3_tts.md) | `examples/configs/qwen3_tts_0_6b_customvoice.yaml` | Text-only requests use the checkpoint speaker table. Set `voice` to the desired checkpoint speaker |
 | [Qwen3-TTS VoiceDesign](../cookbook/qwen3_tts.md) | `examples/configs/qwen3_tts_1_7b_voicedesign.yaml` | Requires `task_type="VoiceDesign"` and non-empty `instructions`. No reference audio is required |
-| [Ming-Omni-TTS](../cookbook/ming_tts.md) | `examples/configs/ming_omni_tts.yaml` | Text-only synthesis or one local reference clip with its transcript; streaming; the provided config uses TP1 |
+| [Ming-Omni-TTS](../cookbook/ming_tts.md) | `examples/configs/ming_omni_tts.yaml` | Text-only synthesis or one local reference clip with its transcript; TP1 is supported and the provided config uses TP2 |
```
Ratish1: `examples/configs/ming_omni_tts.yaml` still sets `tp_size: 1`, so the provided config uses tp1. keep the original wording and remove the tp2 claim.

docs/basic_usage/tts.md
```diff
@@ -175,6 +176,17 @@ sgl-omni serve \
...
 ```
 
+For Fun-CosyVoice3:
+
+```bash
+sgl-omni serve \
+  --model-path FunAudioLLM/Fun-CosyVoice3-0.5B-2512 \
+  --config examples/configs/fun_cosyvoice3_0_5b.yaml \
+  --allowed-media-domain huggingface.co \
+  --allowed-media-domain cas-bridge.xethub.hf.co \
```
Ratish1, line 186: the reference url below redirects to `us.aws.cdn.hf.co`, and redirect targets are revalidated against this allowlist. add `--allowed-media-domain us.aws.cdn.hf.co` here so the documented request works.

Ratish1 (approved):

PR #1332, opened 2026-08-04
[Router] Add complete TTS routing support to Omni Router

Ratish1:
## Motivation

TTS serving exposes HTTP speech, batch synthesis, voice management, and WebSocket streaming APIs. Omni Router needs to handle this complete surface so data-parallel TTS replicas can share one ingress without losing capability-aware selection, health tracking, admission control, or uploaded-voice ownership.

## Modifications

- Add routing metadata for speech batches, voice-management operations, and TTS service classes, including per-worker request counters.
- Add a bidirectional proxy for `/v1/audio/speech/stream` with capability and model selection, shared admission limits, session accounting, and transport-aware health classification.
- Introduce an explicit voice owner for mutable uploaded-voice state. Voice operations and synthesis using uploaded voices remain on the owner, while built-in voices continue through normal load balancing.
- Hydrate the uploaded-voice registry from the owner and preserve concurrent uploads and deletes while hydration is in progress. If the owner is unavailable, state-dependent requests fail with `503` instead of being sent to an incompatible worker.
- Document the new routes and ownership behavior and cover the HTTP, WebSocket, routing, failure, and lifecycle contracts in the router test suite.

## Related Issues

Related to #601.


sglang_omni_router/route_metadata.py
```diff
@@ -144,6 +173,9 @@ def extract_route_metadata(request: Request, path: str, body: bytes) -> RouteMet
         required_capabilities=required_capabilities,
         body_exceeds_metadata_limit=body_exceeds_metadata_limit,
         route_capabilities_header_present=route_capabilities_header_present,
+        route_kind=route_kind,
+        service_class=_service_class_for_route(route_kind),
+        voice_names=_voice_names(route_kind, payload),
```
JiaxinD: Voice-owner pinning keys off `_is_json_request`, but the worker's `create_speech` calls `request.json()` regardless of Content-Type: a JSON speech body sent without a json Content-Type bypasses both the owner pin and the >1MiB conservative pin and gets load-balanced. Consider parsing SPEECH/SPEECH_BATCH bodies as JSON regardless of Content-Type, or rejecting non-JSON Content-Type at the router.

sglang_omni_router/proxy.py
```diff
@@ -664,11 +772,66 @@ def _request_id_from_headers(request: Request) -> str | None:
...
+            continue
+        payload = part.get_payload(decode=True)
+        if payload is None:
+            return None
+        charset = part.get_content_charset() or "utf-8"
+        try:
+            value = payload.decode(charset).strip()
+        except (LookupError, UnicodeDecodeError):
+            return None
+        return value or None
```
JiaxinD: `_multipart_text_field` returns the first `name` part, but the worker's `Form(...)` binds the last duplicate, so duplicate parts record the wrong name in the registry and the real voice then load-balances off-owner. Recording the stored name echoed in the worker's 2xx response body would close this.

sglang_omni_router/voice_routing.py
```diff
@@ -0,0 +1,205 @@
...
+        if not names:
+            return False
+        if not self._hydrated:
+            # The owner can serve both presets and uploaded voices. Falling back
+            # to it preserves correctness when registry discovery is unavailable.
+            return True
+        return bool(names & self._uploaded_names)
+
+    async def _run_hydration(self) -> None:
+        while not self._hydrated:
```
JiaxinD: If the 2xx is lost after the owner commits an upload, the registry misses the voice and hydration never re-runs, so requests for it load-balance off-owner until a retried upload repairs it. Is one-shot hydration intentional, or is periodic reconciliation worth adding?

sglang_omni_router/voice_routing.py
```diff
@@ -0,0 +1,205 @@
...
+            return True
+        names = {
+            normalized
+            for name in voice_names
+            if (normalized := _normalize_voice_name(name)) is not None
+        }
+        names.discard(DEFAULT_VOICE_NAME)
+        if not names:
+            return False
+        if not self._hydrated:
```
JiaxinD: Before hydration every non-"default" name is owner-required, so built-in named speakers (qwen3-tts "Vivian", qwen3-omni "Ethan" via voice/speaker) pin to the owner, or 503 pool-wide if it is unroutable; and a ValueError/404 from `/v1/audio/voices` retries forever, making that window permanent under version skew. Consider a bounded fallback after N failed hydrations, or gating only on known-uploaded names.

sglang_omni_router/voice_routing.py
```diff
@@ -0,0 +1,205 @@
...
+            owner = next(
+                (
+                    worker
+                    for worker in self._workers
+                    if can_own_uploaded_voices(worker.capabilities)
+                ),
+                None,
+            )
+            if owner is not None:
+                self._owner_url = owner.url
```
JiaxinD: `resolve_owner` picks the first capability-matching worker without checking `is_routable` and never re-resolves, so a worker disabled by journal recovery at startup becomes permanent owner: hydration stalls and named-voice traffic 503s despite a healthy capable peer. Preferring routable candidates on first resolution and surfacing owner/hydration state in stats would help.

sglang_omni_router/proxy.py
```diff
@@ -664,11 +772,66 @@ def _request_id_from_headers(request: Request) -> str | None:
...
+    )
+
+
 def _large_request_extra_capabilities_or_error(
     workers: list[Worker],
     metadata: RouteMetadata,
 ) -> tuple[set[Capability], str | None]:
-    if not metadata.body_exceeds_metadata_limit:
+    if (
+        metadata.route_kind is not RouteKind.GENERATION
```
JiaxinD: Early-returning non-GENERATION drops main's large-body capability widening for speech routes, and the multi-process data plane builds ProxyHandler without voice_routing, so a >1MiB speech body with reference audio can land on a worker lacking `audio_input` where main widened or failed closed. Keep the widening for speech kinds, or wire voice routing into the DP.

sglang_omni_router/proxy.py
```diff
@@ -363,12 +381,41 @@ async def _forward_admitted(
...
+        )
+        if voice_mutation is None:
+            return await self._select_and_forward(
+                request,
+                path,
+                body,
+                metadata,
+                release,
+            )
+        async with self._voice_mutation_lock:
```
JiaxinD: This lock is held across the whole upstream request while each queued mutation keeps an admission slot and a buffered ~10MiB body, and HTTP+WS share the admission pool. Could a burst of uploads starve all routes, or should the lock scope end before the upstream await?

docs/basic_usage/omni_router.md
```diff
@@ -385,6 +411,10 @@ when the router cannot infer a single safe worker set:
   `image_input`, `audio_input`, `video_input`, `audio_output`, or `streaming`
 - `X-SGLang-Omni-Route-Stream`: `true` or `false` for large streaming requests
 
+Speech and speech-batch JSON bodies larger than 1 MiB are conservatively pinned
+to the voice owner because the router cannot fully inspect them to rule out an
+uploaded voice reference. Route-hint headers do not override voice ownership.
+
 These headers are router-only hints and are not forwarded to workers.
```
JiaxinD, line 439: This sentence is now inaccurate for the WS path: `websocket_proxy._REQUEST_HEADERS_TO_STRIP` doesn't include the `x-sglang-omni-route-*` headers, so they are forwarded on the upstream handshake (and ignored for WS routing). Either strip them in `_forward_headers` or scope this to HTTP routes.

sglang_omni_router/route_metadata.py
Ratish1: done

sglang_omni_router/proxy.py
Ratish1: done

sglang_omni_router/voice_routing.py
Ratish1: done

sglang_omni_router/voice_routing.py
Ratish1: done

sglang_omni_router/voice_routing.py
Ratish1: done

sglang_omni_router/proxy.py
Ratish1: done

sglang_omni_router/proxy.py
Ratish1: done

docs/basic_usage/omni_router.md
Ratish1: done

JiaxinD (approved):

PR #1336, opened 2026-08-04
[TTS][Qwen3] Reduce first audio latency

luojiaxuan:
Follow-up to #1286. Reduces Qwen3-TTS first-audio latency by:

- batching reference-code extraction and overlapping it with speaker embedding
- using CUDA graphs for fixed-size initial vocoder decode
- prioritizing follow-up decode without delaying first audio

H100 Seed-TTS EN full set, matched config and hardware, 1088 requests per setup.

| Concurrency | Current TTFA p95 | Baseline | Current C200 | Baseline C200 | Current/Baseline QPS | Current/Baseline EN WER |
| 8 | 0.266s | 0.330s | 84.74% | 0% | 6.114 / 2.921 | 1.06% / 1.07% |
| 16 | 0.320s | 0.497s | 12.41% | 0% | 7.746 / 3.107 | 1.06% / 1.06% |
| 32 | 0.541s | 0.895s | 3.03% | 0% | 9.612 / 3.155 | 1.05% / 1.06% |

All runs completed 1088/1088. First payload p95 is 3.75 KiB. This change does not modify talker sampling.

Validation:
- repository pre-commit passed
- reference encoder unit tests: 14 passed
- Python compileall passed
- full Qwen3-TTS coverage runs in CI

### DIFF
--- a/sglang_omni/models/qwen3_tts/request_builders.py
+++ b/sglang_omni/models/qwen3_tts/request_builders.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import concurrent.futures
+import contextlib
 import hashlib
 import json
 import queue
@@ -578,17 +579,40 @@ class _Qwen3TTSAdhocReferenceInput:
     x_vector_only_mode: bool
 
 
+def _new_cuda_encode_stream(device: Any) -> torch.cuda.Stream | None:
+    if device is None or not torch.cuda.is_available():
+        return None
+    try:
+        resolved = torch.device(device)
+    except (TypeError, ValueError, RuntimeError):
+        return None
+    if resolved.type != "cuda":
+        return None
+    return torch.cuda.Stream(device=resolved)
+
+
+def _record_ref_code_consumer_stream(ref_code: Any) -> Any:
+    # note (luojiaxuan): reference codes may be allocated on the batcher's
+    # private stream; register the consumer stream with the caching allocator
+    # so a later batch cannot recycle the block while reads are still queued.
+    if isinstance(ref_code, torch.Tensor) and ref_code.is_cuda:
+        ref_code.record_stream(torch.cuda.current_stream(ref_code.device))
+    return ref_code
+
+
 class _Qwen3TTSRefCodeBatcher:
     def __init__(
         self,
         speech_tokenizer: Any,
         *,
+        device: Any | None = None,
         max_batch_size: int = 8,
         max_batch_wait_ms: float = 2.0,
     ) -> None:
         self._speech_tokenizer = speech_tokenizer
         self._max_batch_size = max(int(max_batch_size), 1)
         self._max_batch_wait_s = max(float(max_batch_wait_ms), 0.0) / 1000.0
+        self._encode_stream = _new_cuda_encode_stream(device)
         self._queue: queue.Queue[object] = queue.Queue()
         self._thread = threading.Thread(
             target=self._run,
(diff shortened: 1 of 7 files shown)


luojiaxuan:
@SandyLuXY #1286 has landed. This is the isolated TTFA follow-up; please approve for merge.

sglang_omni/models/qwen3_tts/request_builders.py
```diff
@@ -617,18 +737,38 @@ def options_key(self, item: _Qwen3TTSAdhocReferenceInput) -> str:
...
         with torch.no_grad():
-            prompt_items = self._wrapper.create_voice_clone_prompt(
-                ref_audio=item.ref_audio,
-                ref_text=item.ref_text,
-                x_vector_only_mode=item.x_vector_only_mode,
+            normalized = self._wrapper._normalize_audio_inputs([item.ref_audio])
+            if len(normalized) != 1:
+                raise ValueError("Qwen3-TTS expects exactly one reference audio")
+            waveform, sample_rate = normalized[0]
+            ref_code_future = self._ref_code_batcher.submit(waveform, sample_rate)
```
SandyLuXY, line 801: The production preprocessor is serial (`max_concurrency=1`, no batch function), and this hook waits on the future before it can dequeue another request, so the queue never forms a cross-request batch; a two-request production repro recorded tokenizer batches `[1, 1]`. Please correct me if I'm wrong 🤔️

sglang_omni/models/qwen3_tts/request_builders.py
```diff
@@ -617,18 +737,38 @@ def options_key(self, item: _Qwen3TTSAdhocReferenceInput) -> str:
...
+            speaker_sample_rate = self._model.speaker_encoder_sample_rate
+            if sample_rate != speaker_sample_rate:
+                import librosa
+
+                speaker_waveform = librosa.resample(
+                    y=speaker_waveform.astype("float32"),
+                    orig_sr=int(sample_rate),
+                    target_sr=speaker_sample_rate,
+                )
+            speaker_embedding = self._model.extract_speaker_embedding(
```
SandyLuXY, line 812: This worker and `extract_speaker_embedding()` both use CUDA stream 0 because neither path selects a stream, so their GPU kernels serialize. Shall we run reference-code encoding on a dedicated stream, hand the result off with an event/`wait_stream` before resolving the future, and cover the real CUDA path?

sglang_omni/models/qwen3_tts/request_builders.py
luojiaxuan: Good catch, you are right — the preprocessor was serial so the batch never really formed. Fixed in 83da919: preprocessing now runs with max_concurrency=8 (same as batcher max_batch_size), so cross-request batch can actually form. Also added a two-request test through the production preprocess path, tokenizer batch is [2] now.

sglang_omni/models/qwen3_tts/request_builders.py
luojiaxuan: Done in 83da919. Ref-code encoding now runs on a dedicated CUDA stream (picked from model.device), and the worker records an event on it and waits before resolving the future, so the default stream is not touched anymore — speaker embedding can really overlap. Consumer side does record_stream so the allocator will not recycle the side-stream block too early. Also added a real CUDA unit test to cover this path (asserts encode runs on the non-default stream).

SandyLuXY (approved):
lgtm

PR #1326, opened 2026-08-04
[ASR][Perf] Enable async decode for Qwen3-ASR

SandyLuXY:
## Motivation

Qwen3-ASR already uses the shared plain `ModelRunner`, but its stage did not enable the shared one-step-lookahead async decode path and the CLI rejected decode-mode overrides for its factory. As a result, host-side output processing
could not overlap the next decode forward through this mechanism under concurrent serving.

This implements Q-PR3 from the Qwen3-ASR high-concurrency roadmap while keeping the decode loop model-neutral.

## Modifications

1. `sglang_omni/models/qwen3_asr/stages.py`
   - Adds `enable_async_decode=True` and `async_decode_min_batch_size=2` to
     `create_sglang_qwen3_asr_executor`.
   - Passes both settings to `OmniScheduler` while continuing to use the shared
     `ModelRunner`; no Qwen3-ASR-specific decode loop is introduced.
2. `sglang_omni/cli/serve.py`
   - Adds `create_sglang_qwen3_asr_executor` to the async-decode factory
     allowlist.
   - Enables `--decode-mode async|sync` and
     `--async-lookahead-min-batch-size` for Qwen3-ASR and updates CLI help and
     validation text.
3. `tests/unit_test/qwen3_asr/test_pipeline.py` and
   `tests/unit_test/higgs_tts/test_cli_decode_mode.py`
   - Covers the default async mode and minimum batch threshold.
   - Verifies explicit factory propagation and CLI overrides for forced sync,
     async mode, and a custom threshold.
   - Reuses the shared async-decode lifecycle suites for finish, abort,
     stale/retracted rows, exact KV-slot cleanup, failure handling, and stream
     ordering.
4. `docs/cookbook/qwen3_asr.md` and `tests/README.md`
   - Documents the default lookahead behavior, threshold, forced-sync command,
     and corresponding test coverage.

## Related Issues

Related to #1324 (Q-PR3).

## Accuracy Test

**Setup:**
- **Refs:** baseline `main` @ `4ce2a6ca`; candidate `feat/qwen3-asr-async-decode` @ `98ec3e2a`.
- **Model:** `Qwen/Qwen3-ASR-1.7B`, cached snapshot `7278e1e70fe206f11671096ffdd38061171dd6e5`.
- **Dataset:** full `zhaochenyang20/seed-tts-eval-arrow` English split, 1,088 samples per repeat.
- **Workload:** non-streaming transcription, concurrency 16, three repeats per ref
[...]

Issue #1324, opened 2026-08-04
[ASR][Perf] Qwen3-ASR high-concurrency serving optimization roadmap


luojiaxuan:
## Goal

Close the high-concurrency serving gap between SGLang-Omni Qwen3-ASR and a fixed serving baseline, while completing the language-routing correctness needed for representative multilingual performance evaluation.

This is a focused follow-up to #924 and #831. It tracks five independently reviewable PRs so correctness, measurement, and optimization changes do not get mixed into one large patch.

## Current observation

An external report compared SGLang-Omni at commit `9924255` with a fixed baseline on the same pre-segmented audio workload. Both paths received the same VAD/chunked inputs.

| concurrency | Omni RTFx / rps / p50 / p95 | baseline RTFx / rps / p50 / p95 |
|---:|---:|---:|
| 1 | 19.5 / 2.2 / 0.49s / 0.80s | 19.8 / 2.2 / 0.47s / 0.70s |
| 8 | 98 / 11.1 / 0.55s / 1.16s | 118 / 13.3 / 0.46s / 0.96s |
| 16 | 187 / 20.3 / 0.57s / 1.06s | 196 / 21.3 / 0.50s / 0.86s |
| 32 | 300 / 30.4 / 0.75s / 1.41s | 402 / 40.7 / 0.54s / 1.15s |
| 64 | 375 / 38.7 / 1.25s / 2.30s | 466 / 48.1 / 0.77s / 1.70s |

At concurrency 1, throughput is at parity. At concurrency 32 and 64, Omni is approximately 25.3% and 19.5% lower in requests/s, with higher median and tail latency. The shape points to a concurrency-scaling problem rather than a single-request model-forward problem.

The reported commit predates the latest main and #1176, so the first step must reproduce the comparison on current main before attributing the gap to a specific component.

# PR Plan

### Problem

The current Qwen3-ASR request builder effectively supports only two prompt modes:

- `zh`, `cn`, and `zh*` force `Chinese`;
- every other value, including an omitted language, forces `English`.

This disables automatic language identification and silently maps valid non-Chinese language codes to English. It also prevents a representative multilingual performance and correctness gate.

### Scope

- Use the checkpoint's supported-language metadata as the source of truth.
- Normalize supported language names and common language-code aliases.
- When language is omitted or explicitly requests automatic detection, do not force a language prefix.
- Parse the generated language tag and return the detected language with the transcript.
[...]

PR #1340, opened 2026-08-05
[ASR][Perf] Current-main concurrency benchmark and bottleneck profile for Qwen3-ASR

luojiaxuan:
## Motivation

Q-PR2 of #1324. Reproduces the reported high-concurrency serving gap on current main with a pinned environment and enough internal telemetry to attribute request time to request build, admission queueing, encoder+prefill, and decode. These results are the before-baseline for Q-PR4 and Q-PR5 (Q-PR3 landed as #1326).

## Modifications

- `benchmarks/eval/benchmark_asr_seedtts.py`: adds `--save-raw-dir` (raw per-request JSONL), `--profile-events` (one extra profiled pass per concurrency with the stage/hop breakdown attached), `--sample-util` (host CPU + GPU utilization), `--fingerprint` (git/dependency/driver/model-revision identity), and surfaces p50 latency.
- New `benchmarks/eval/asr_profiling.py`: profile-control client, breakdown builder, utilization sampler, environment fingerprint (all dependency-light, unit-tested).
- Profiler: pairs queue-wait, build-to-queued, first-forward, and decode-tail intervals; emits `scheduler_prefill_end` once per request after its first executed batch (realized batch size in metadata) so non-streaming models get a prefill/decode split.
- Docs: `docs/developer_reference/qwen3_asr_concurrency_profile.md` report + cookbook pointer.

## Results

Single H100 80 GB per server, DP=1, full SeedTTS sets, one discarded warmup + three measured repeats, identical pre-segmented audio through the same client. Full tables and methodology in the doc; highlights:

Current, quiet host (EN 1088):

| conc | req/s | lat mean | lat p95 | WER | completed |
| 32 | 108.0 | 0.295s | 0.399s | 0.0122 | 3264/3264 |
| 64 | 103.0 | 0.610s | 0.741s | 0.0125 | 3165/3264 |

Current vs baseline, back-to-back same-host windows (EN):

| conc | current req/s | baseline req/s | ratio |
|---:|---:|---:|---:|
| 8 | 41.9 | 70.3 | 1.68x |
| 16 | 62.6 | 120.6 | 1.93x |
| 32 | 79.5 | 183.8 | 2.31x |
| 64 | 80.9 | 264.6 | 3.27x |

ZH tracks the same shape (2.7-3.2x); WER/CER stay in band everywhere; baseline completes 100% at c=64 while current sheds 1-2%.

Bottleneck decomposition (profiled passes): at c=64 admission queue wait explodes from ~6 ms to ~413 ms per request (the knee; `max_running_requests=32` + `request_build_max_pending=16`); the decode tail grows 55→~360 ms per request while GPU utilization stays at 33-54% (host-dispatch-bound, matching #907); encoder+prefill costs 28-46 ms per admission on the scheduler thread/default stream and stalls the running decode batch (Q-PR4 target); the build-to-queued FIFO gap reaches 16-30 ms (Q-PR5 scope).

## Validation

- 169 unit tests pass (profiler, benchmarks, qwen3_asr, scheduling) on the pinned H100 environment; repository pre-commit passes.
- The new telemetry is emit-gated: with no active profile run the added event path is a no-op, and measured repeats never run with profiling enabled.
- Raw per-request records, profiler event JSONL, and environment fingerprints for every table are preserved on the measurement host.

### DIFF
--- a/benchmarks/eval/benchmark_asr_seedtts.py
+++ b/benchmarks/eval/benchmark_asr_seedtts.py
@@ -99,6 +99,7 @@ from benchmarks.eval.asr_profiling import (
     UtilizationSampler,
     build_stage_breakdown,
     collect_environment_fingerprint,
+    collect_server_identity,
     start_request_profile,
     stop_request_profile,
 )
@@ -505,7 +506,13 @@ async def _run_profiled_pass(args, samples, concurrency: int) -> dict | None:
             started.append(url)
     except requests.RequestException as exc:
         for url in started:
-            stop_request_profile(url)
+            try:
+                stop_request_profile(url, run_id)
+            except requests.RequestException as stop_exc:
+                print(
+                    f"[conc={concurrency}] failed to stop profiling on "
+                    f"{url}: {stop_exc}"
+                )
         print(f"[conc={concurrency}] profiling unavailable, skipping: {exc}")
         return None
     try:
@@ -513,9 +520,14 @@ async def _run_profiled_pass(args, samples, concurrency: int) -> dict | None:
     finally:
         for url in started:
             try:
-                stop_request_profile(url)
-            except requests.RequestException:
-                pass
+                stop_request_profile(url, run_id)
+            except requests.RequestException as stop_exc:
+                # The pass metrics remain valid; profiling just keeps
+                # recording on that worker until the next explicit stop.
+                print(
+                    f"[conc={concurrency}] failed to stop profiling on "
+                    f"{url}: {stop_exc}"
+                )
     report = build_stage_breakdown(event_dir)
     return {
         "run_id": run_id,
@@ -607,9 +619,12 @@ def main() -> None:
         "results": aggregates,
     }
     if args.fingerprint:
-        payload["environment_fingerprint"] = collect_environment_fingerprint(
-            args.model_path
-        )
+        # The client fingerprint equals the server's only when both share one
+        # host and checkout; the server block records what the server reports.
+        payload["environment_fingerprint"] = {
+            "client": collect_environment_fingerprint(args.model_path),
+            "server": collect_server_identity(f"http://{args.host}:{args.port}"),
+        }
     output_path = os.path.abspath(args.output)
     with open(output_path, "w") as handle:
         json.dump(payload, handle, indent=2)
--- a/benchmarks/eval/benchmark_asr_seedtts.py
+++ b/benchmarks/eval/benchmark_asr_seedtts.py
@@ -89,11 +89,19 @@ import asyncio
 import json
 import os
 import statistics
+import time
 
 import requests
 
 from benchmarks.dataset.prepare import DATASETS
 from benchmarks.dataset.seedtts import SampleInput, load_seedtts_samples
+from benchmarks.eval.asr_profiling import (
+    UtilizationSampler,
+    build_stage_breakdown,
+    collect_environment_fingerprint,
+    start_request_profile,
+    stop_request_profile,
+)
 from benchmarks.tasks.asr import (
     FUN_ASR_MODEL_PATH,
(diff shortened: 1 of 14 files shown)


benchmarks/eval/benchmark_asr_seedtts.py
```diff
@@ -455,9 +601,15 @@ def main() -> None:
...
             "stream": args.stream,
+            "profile_events": args.profile_events,
+            "sample_util": args.sample_util,
         },
         "results": aggregates,
     }
+    if args.fingerprint:
+        payload["environment_fingerprint"] = collect_environment_fingerprint(
+            args.model_path
+        )
```
lijrjyan: A bit confused: `collect_environment_fingerprint()` appears to run in the benchmark client process, so its Git revision, packages, and model-cache metadata describe the client rather than the server at `args.host:args.port` (which host current commit and baseline implementation).

It would be better to store this as `client_fingerprint` and record the current/baseline server revision separately(maybe retrieving manifest from server side)?

benchmarks/eval/benchmark_asr_seedtts.py
```diff
@@ -382,7 +422,109 @@ def parse_args() -> argparse.Namespace:
...
+    event_dir = os.path.join(args.profile_event_dir, run_id)
+    profile_urls = [u.strip() for u in args.profile_urls.split(",") if u.strip()]
+    started: list[str] = []
+    try:
+        for url in profile_urls:
+            start_request_profile(url, run_id, event_dir)
+            started.append(url)
+    except requests.RequestException as exc:
+        for url in started:
+            stop_request_profile(url)
```
lijrjyan: add `run_id` param to func `stop_request_profile(url)` to avoid wildcard stopping other profilling session? Not sure dropping the `run_id` here is a good idea or not.

benchmarks/eval/benchmark_asr_seedtts.py
luojiaxuan: Good point — the fingerprint was really describing the client process. Split it now: the JSON gets a client block (git/deps/GPU of the benchmark process) plus a server block that records what the server itself reports via /v1/models. When client and server share one host and checkout they coincide, and the doc says so explicitly. A richer server-side manifest endpoint would be nice but is beyond this PR.

benchmarks/eval/benchmark_asr_seedtts.py
luojiaxuan: Agree, wildcard stop is risky — now both stop paths pass the run_id, so the benchmark only stops its own profiling session, and a failed stop prints a warning instead of being swallowed.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -1194,6 +1195,7 @@ def _run_batch(self, batch, pp_proxy_tensors=None):
         batch.forward_iter = self.forward_ct
         sched_output = self._build_sched_output(batch)
         mr_output = self._model_runner.execute(sched_output)
+        self._emit_prefill_end_for_batch(batch)
```
SandyLuXY, line 1198: This runs after every synchronous decode even when profiling is inactive and every request already emitted its one prefill-end event. I checked that a completed B=32 batch still cost about 3.23 microseconds per step locally. Please add an O(1) no-pending fast path before metadata construction and the request scan.

luojiaxuan: Done — added the O(1) fast path: _prefill_end_done only ever holds rids that are in _prefill_start_done and both sets discard together, so equal sizes prove nothing is pending and we return before building metadata or scanning reqs. Also added a unit test where the batch object raises on any attribute access, so the fast path provably touches nothing.

SandyLuXY (approved):
lgtm

PR #1341, opened 2026-08-05
[ASR][Perf] Pre-LM audio encoder service for Qwen3-ASR

luojiaxuan:
## Motivation

Q-PR4 of #1324. The Qwen3-ASR audio tower used to run inside multimodal embedding construction on the LM forward path: every admission stalled the whole running decode batch for the encoder forward (28-46 ms measured in #1340), on the scheduler thread and the default stream, and encoder work could not be queued, overlapped, or batched independently.

## Modifications

- New `sglang_omni/models/qwen3_asr/encoder_service.py`: mirrors the shared `PreLMEncoderService` pattern (Fun-ASR precedent). Complete audio-tower execution now happens at request-build time on a dedicated worker thread and CUDA stream, batched across concurrent requests through the model's existing pad-and-mask `get_audio_feature`, with single-flight deduplication and a bounded CPU LRU keyed on audio fingerprint + encoder-pipeline namespace.
- Requests are admitted only with the complete precomputed embedding attached (`MultimodalDataItem.precomputed_embeddings`); padded frames never become LM prompt tokens because splitting validates exact per-item token counts against the packed encoder output.
- Failure handling: multi-item batch failures retry per item; OOM failures detach tracebacks (no tensor pinning) and release cached blocks; the consumer stream is registered with the caching allocator so side-stream embeddings cannot be recycled while LM reads are queued.
- `request_build_max_workers` default 2 → 8: `encode_item` blocks its build worker, so batch formation needs concurrent submitters (same as Fun-ASR's config; #1340 measured the 2-worker FIFO drain as a 16-30 ms head-of-line cost).
- Queue/batch/latency/cache telemetry via `stats()` and periodic logs; `enable_pre_lm_encoder` and `pre_lm_*` knobs in the pipeline config.

## Isolated result on the #1340 benchmark (same host, single 80 GB GPU, DP=1, SeedTTS EN 1088, 3 repeats + warmup)

Cold = unique inputs (embedding cache disabled), warm = repeated inputs (cache on). Before-window concurrency 8/16 repeats were partially co-tenant-contended (min-max spans below).

| conc | before req/s | after cold req/s | after warm req/s | before lat mean | after cold lat mean |
| 1 | 13.2 | 10.4 | 12.0 | 0.075 s | 0.096 s |
| 8 | 26.6 (21.7-34.4) | **61.8** | 34.2 | 0.312 s | **0.129 s** |
| 16 | 46.7 (30.2-61.1) | **85.4** | 78.6 | 0.371 s | **0.186 s** |
| 32 | 99.1 | 102.3 | 102.1 | 0.321 s | 0.310 s |
| 64 | 105.1 | 80.9 | 93.6 | 0.598 s | 0.777 s |

- Corpus WER is 0.0122 at every level in every run — encoder and end-to-end outputs are quality-equivalent.
- Mid-concurrency throughput roughly doubles cold and mean latency more than halves: encoding now overlaps decode instead of stalling it.
- Two honest costs at the defaults: concurrency-1 latency pays ~20 ms for the service round-trip, and concurrency 64 regresses (more build-backlog shedding: encode now sits inside request build behind `request_build_max_pending=16`). Both interact with the admission policy that Q-PR5 sweeps (`pending`/`max_running`); its measurements will pick the final defaults and are running now.

## Validation

- 179 unit tests pass (qwen3_asr including 22 new encoder-service tests mirroring the Fun-ASR suite plus OOM/packed-output coverage, fun_asr encoder service, scheduling) on the pinned H100 environment; pre-commit clean.
- Cover: finish/abort via scheduler shutdown_callback, batch-failure per-item retry, OOM recovery, single-flight dedup under 8-thread concurrency, stale-cache recheck, eviction, token-count mismatch, encode-after-validation ordering.


SandyLuXY:
PR LGTM. However, the PR's cold table regresses concurrency 1 by 21% throughput / 28% mean latency and concurrency 64 by 23% / 30% which is a bit concerning, was it caused by per-run variance?

luojiaxuan:
Re the cold-table regressions (@SandyLuXY): checked carefully, neither is per-run variance, but they are different in nature.

**Concurrency 1** — real, fixed indirection cost. Repeats are tight on both sides (13.1/13.3/13.3 vs 10.4/10.4/10.4 req/s), and on a second host all 8 admission-knob configs land at 0.099–0.101 s regardless of settings, so it is not noise: moving encode behind the service costs thread handoff + a 4 ms batch-formation window + cross-stream sync + H2D attach (~20 ms total). Pushed a fix in this PR: the batch window now defaults to 0 (greedy drain — items queued while the previous batch encoded are taken instantly, so batches still form under load; same reasoning as the MOSS-TD encoder service). Re-measured cold: c=1 mean 0.100 → 0.093 s and c=8 batching throughput unchanged (57.0 vs 57.7 req/s). The remaining ~13 ms is the price of taking encode off the LM path, which buys the 1.8–2.3× at concurrency 8–16.

**Concurrency 64** — real but config-coupled, resolved by #1345. The tell is shedding doubling (2.9% → 5.9%): encode now sits inside request build behind `request_build_max_pending=16`, so the 64-deep loop overflows the build backlog. With #1345's pending=32 the combined state re-measures at zero shedding at c=64, and the documented `--max-running-requests 64` override reaches ~97–100 req/s (vs 51 with the old defaults, same host and regime).

Corpus WER stayed 0.0122 in every one of these runs.

SandyLuXY (approved):
lgtm

PR #1343, opened 2026-08-05
Align Omni CI with image Python 3.12

Hayden727:
## Summary

- align Omni CI with the image's Python 3.12 ABI
- reuse image-provided Torch and FlashInfer instead of installing ABI-specific copies into the CI venv
- keep editable project installs dependency-free so uv does not replace CUDA wheels
- accept CUDA local-version suffixes such as `2.11.0+cu130` when validating public-version pins

## Root cause

The image uses Python 3.12 while CI created a Python 3.11 venv. FlashInfer JIT therefore rewrote `build.ninja` with Python 3.11 include and package paths, invalidating objects prebuilt with Python 3.12.

## Impact

This is the prerequisite/high-priority CI change for #1120. Overlapping compatibility PRs should rebase after this PR merges.

Validated against:

`hongccc/sglang-omni:dev@sha256:374d0b1c30b2bff685b1716fc64a02ad3b3d0a90fe2ce73ce9861a6992c28101`

## Validation

- Python 3.12 fresh CI environment build on H100
- reusable-environment validation, editable refresh, and second validation passed
- all 14 exact dependency pins passed
- Torch and FlashInfer resolve from `/usr/local/lib/python3.12/dist-packages`
- a shared FlashInfer cache stayed unchanged and served the first call in about 20 ms
- pre-commit hooks and shell syntax checks passed


.github/scripts/prepare_omni_venv.sh
```diff
@@ -35,12 +35,12 @@ echo "Preparing fresh ${HOST} (full rebuild)"
...
 rm -rf "${OMNI_CI_HOME}"
 mkdir -p "${OMNI_CI_HOME}"
-uv venv "${HOST}" -p 3.11
+uv venv --system-site-packages "${HOST}" -p /usr/bin/python3.12
 
 rm -rf "./${VENV_NAME}"
 ln -sfn "${HOST}" "./${VENV_NAME}"
 source "${VENV_NAME}/bin/activate"
-uv pip install -e .
+uv pip install --no-deps -e .
```
stephenkgli: The CI setup creates and populates a Python 3.12 virtual environment, but the workflows invoke `pytest` directly. Because the environment is installed with `--no-deps`, it does not contain its own `pytest` executable, so the command resolves to `/usr/bin/pytest` and runs under `/usr/bin/python3` instead of the prepared virtual environment. Consequently, packages installed only in the virtual environment—such as `openai-whisper`—are unavailable during testing, which caused the ASR job to fail with `ModuleNotFoundError: No module named 'whisper'`. Run tests via `$OMNI_CI_HOME/bin/python -m pytest` and ensure retry wrappers preserve that interpreter.

.github/scripts/verify_omni_installed_pins.py
stephenkgli, line 61: The rebuild path installs the project with `uv pip install --no-deps -e .`, so dependencies declared in `pyproject.toml` are neither resolved nor installed. The resulting environment depends on packages already present in the CI image or handled by a few explicit fallback installations. Because the validator checks only a fixed import list and `exact == pins`, missing range-constrained or transitive dependencies can go undetected until tests fail at runtime.

.github/scripts/prepare_omni_venv.sh
```diff
@@ -35,27 +35,21 @@ echo "Preparing fresh ${HOST} (full rebuild)"
 rm -f "${OMNI_CI_HOME}/.omni-env-complete"
 rm -rf "${OMNI_CI_HOME}"
 mkdir -p "${OMNI_CI_HOME}"
-uv venv "${HOST}" -p 3.11
+uv venv --system-site-packages "${HOST}" -p /usr/bin/python3.12
```
AkazaAkane, line 38: This venv now inherits the container's system packages, so the container image digest is part of the environment identity.

In the current run, setup prepared the shared PR venv against one :dev digest, while a later TTS job pulled a different digest. The persisted venv then saw a different set of system packages, failed validation, and the downstream job attempted to rebuild the shared environment.

.github/scripts/prepare_omni_venv.sh
```diff
@@ -35,27 +35,21 @@ echo "Preparing fresh ${HOST} (full rebuild)"
...
-if ! python -c "import zhon.hanzi" 2>/dev/null; then
-  echo "zhon missing from prepared venv, installing pinned dependency..."
-  uv pip install --force-reinstall --no-deps --no-cache zhon==2.1.1
+mapfile -t MISSING_REQUIREMENTS < <(
+  python "${SCRIPT_DIR}/omni_missing_dependencies.py" pyproject.toml
+)
+if [ "${#MISSING_REQUIREMENTS[@]}" -gt 0 ]; then
+  echo "Installing dependencies missing from the image:"
+  printf '  %s\n' "${MISSING_REQUIREMENTS[@]}"
+  python -m pip install "${MISSING_REQUIREMENTS[@]}"
```
AkazaAkane: This installation path bypasses [tool.uv].override-dependencies.

That matters here because descript-audiotools declares protobuf<3.20, while the SGLang/CUTLASS dependency stack requires protobuf 6.x. The project's uv override exists to force the compatible protobuf version, but python -m pip install "${MISSING_REQUIREMENTS[@]}" runs a normal pip resolution and currently fails with ResolutionImpossible.

AkazaAkane:
LGTM. No remaining code-review blocker.

.github/actions/omni-setup/action.yaml
```diff
@@ -23,19 +23,17 @@ runs:
         echo "HOME=/github/home" >> "${GITHUB_ENV}"
         echo "HF_HOME=/github/home/.cache/huggingface" >> "${GITHUB_ENV}"
         echo "XDG_CACHE_HOME=${OMNI_CI_HOME}/.cache" >> "${GITHUB_ENV}"
-        # FlashInfer does not use XDG_CACHE_HOME. Keep it on the image's
-        # architecture-specific prewarmed cache under /root/.cache/flashinfer.
+        # The CI venv inherits the image's Python/Torch/FlashInfer environment,
+        # so it can safely reuse the architecture-specific prewarmed cache.
         echo "FLASHINFER_WORKSPACE_BASE=/root" >> "${GITHUB_ENV}"
+        echo "FLASHINFER_JIT_DEBUG=0" >> "${GITHUB_ENV}"
```
JiaxinD, line 29: verified on an H100 CI-repro box with the pinned image: `FLASHINFER_JIT_VERBOSE=1` alone flips FlashInfer JIT into debug builds (`-O0 -g`) and rebuilds 10 prewarmed fused_moe objects in ~6 min, so cache-change reports from VERBOSE validation runs are an artifact, not a regression. With this env block as is, the qwen3-omni chain reaches ready in 96s with zero `.o` rebuilds, and the old py3.11 venv flow recompiles fused_moe in 481s, so this guard plus the 3.12 alignment checks out.

SandyLuXY (approved):

sglang_omni/models/qwen3_asr/request_builders.py
```diff
@@ -146,10 +146,6 @@ def request_builder(payload: StagePayload) -> Qwen3ASRRequestData:
         audio = prepared.waveform
         audio_duration_s = prepared.duration_s
         fingerprint = prepared.fingerprint
-        lang_raw = str(params.get("language") or "en").strip().lower()
```
wirybeaver, line 149: thanks for the fix.

wirybeaver (commented):
LGTM. Thanks for fixing the ASR UT

PR #1347, opened 2026-08-05
[ASR] Long-audio chunking for Qwen3-ASR

0xjeffro:
## Motivation

Qwen3-ASR transcription is capped by the per-request context (~115s of audio; anything longer 400s, and before #1176 it silently truncated at 30s). This PR adds serving-layer chunking: audio past the clip limit is split at low-energy points into non-overlapping chunks, each transcribed as an ordinary short request, and the transcripts joined in order.

## Modifications

To make this PR easier to review, it answers the following questions:

1. How audio is chunked, what the split is based on, and boundary handling such as the shortest clip the model accepts (`serve/transcription_chunking.py`: `RMSSplitter` cuts at the quietest energy window within `search_window_s` before each nominal boundary; non-overlapping spans; `min_tail_s=0.5` avoids sub-minimum tail chunks, matching the official `MIN_ASR_INPUT_SECONDS`)

2. For fairness, how one long upload is kept from queueing out short requests (`AudioChunkingConfig.max_concurrent_chunks`, this is a per-request cap: the scheduler is FCFS, so without it a single upload could take all batch slot)

3. Parallelism base on 2. Chunks of one request run concurrently up to that cap (`asyncio.Semaphore` + `gather` in `_transcribe_audio_chunks`, `serve/openai_api.py`; measured on a 243s clip: 7.3s serial → 2.2s)

4.How chunk outputs merge. when a space is inserted at a seam (`join_transcript_parts`: only when both sides of the seam are space-delimited scripts, judged via `unicodedata.east_asian_width`, so CJK seams never gain a spurious space that would inflate CER)

5. Failure handling, client disconnect or any chunk failing fails the whole upload immediately, no waiting
   (`in_flight` ledger + `client.abort` per chunk id; `_await_transcription_with_disconnect_abort`; errors name the chunk and its time range)

6. Verbose response,`verbose_json` returns one segment per chunk with real timestamps instead of a single segment spanning the whole file (`build_verbose_response_from_chunks`, `transcription_adapters/base.py`)

7. `context_length` decoupled from `max_audio_clip_s` (chunk size),
   capacity is sized to the model's native 1,200s limit (official `MAX_ASR_INPUT_SECONDS`, `qwen3_asr/stages.py`), while the 60s default chunk size is a scheduling choice: smaller requests batch better (`qwen3_asr/config.py`)

TODO: docs (cookbook)

## Checklist

- [x] Add unit tests.

## CI


0xjeffro:
Smoke-tested this branch at commit [afee8dc](https://github.com/sgl-project/sglang-omni/pull/1347/commits/afee8dc5f57f1f695b1ab3c461767279100bc9c7) on a single RTX A6000.
Concatenated 50 clips from `zhaochenyang20/seed-tts-eval-50-arrow` (en split) into one 243s WAV, Corpus WER is 0.74%.

0xjeffro:
should co-author: @wirybeaver and @lijrjyan

sglang_omni/serve/transcription_chunking.py
```diff
@@ -0,0 +1,274 @@
...
+
+    ``duration_s <= 0`` means the duration probe failed (``soundfile`` could
+    not read the container). Such uploads keep today's behaviour rather than
+    paying a full decode just to find out how long they are.
+    """
+    if not config.allow_audio_chunking:
+        return False
+    if duration_s <= 0:
+        logger.debug("[transcription] audio duration unknown; skipping chunking")
+        return False
```
SandyLuXY, line 39: Returning `False` when the probe is unknown bypasses both chunking and `max_total_audio_s` for valid formats such as M4A that `load_audio` can decode but libsndfile cannot inspect.

sglang_omni/serve/transcription_chunking.py
```diff
@@ -0,0 +1,274 @@
...
+    rather than by whatever whitespace the model decoded; empty parts (an
+    all-silence chunk) are skipped.
+    """
+    joined = ""
+
+    for part in parts:
+        stripped = part.strip()
+        if not stripped:
+            continue
+        if joined and _is_spaced_script(joined[-1]) and _is_spaced_script(stripped[0]):
```
SandyLuXY, line 284: East Asian width is not a word-spacing predicate: this removes Korean spaces and inserts Thai spaces. Consider making seam joining language/script aware.

sglang_omni/serve/openai_api.py
```diff
@@ -223,6 +233,7 @@ def create_app(
     admin_api_key: str | None = None,
     tts_batch_max_items: int = DEFAULT_TTS_BATCH_MAX_ITEMS,
     architectures: list[str] | None = None,
+    audio_chunking: AudioChunkingConfig | None = None,
```
SandyLuXY, line 188: This introduces public endpoint limits and response behavior but leaves the cookbook TODO unresolved. Document the default chunk/total limits, streaming behavior, concurrency policy, supported-format fallback, and chunk-level `verbose_json` semantics.

SandyLuXY (changes requested):

wirybeaver:
Thank you for building the chunking foundation here. I opened #1390 as a focused correction stacked directly on this branch. It aligns F1 with the clarified Qwen3-ASR behavior: native processing through 1,200 seconds, a 4,096-token output budget, decode fallback for unknown metadata, exact transcript concatenation, incremental long-audio streaming, and no chunk-derived timestamps.

It intentionally keeps the existing scheduling policies unchanged. We can either review it as a stacked PR or merge/cherry-pick it into `feat/qwen3-asr-chunking` before this PR lands—whichever is easier for the maintainers.

sglang_omni/config/schema.py
```diff
@@ -212,6 +212,52 @@ def model_post_init(self, __context: Any = None) -> None:
...
+    # recording), so we leave the default value of `allow_audio_chunking` = False.
+    allow_audio_chunking: bool = False
+    # Note (Jeffro): Longest clip (chunk length) sent to the engine in one request.
+    # Must stay within what the model's context can hold (Qwen3-ASR sizes its
+    # context for the official 1,200s native limit); below that ceiling it
+    # is a scheduling trade-off: shorter chunks batch better and keep a
+    # long upload from monopolizing the engine, at the cost of more seams.
+    max_audio_clip_s: float = Field(default=60.0, gt=0)
+
+    max_total_audio_s: float | None = Field(default=3600.0, gt=0)
```
wirybeaver, line 237: Could we move `max_total_audio_s` and the per-request concurrency/fairness policy into a focused follow-up PR? Additional scheduling policies exceed F1's scope, while #1390 keeps the correction limited to the model-native limit and transcription correctness. Happy to help with the follow-up.

0xjeffro:
> Thank you for building the chunking foundation here. I opened #1390 as a focused correction stacked directly on this branch. It aligns F1 with the clarified Qwen3-ASR behavior: native processing through 1,200 seconds, a 4,096-token output budget, decode fallback for unknown metadata, exact transcript concatenation, incremental long-audio streaming, and no chunk-derived timestamps.
> 
> It intentionally keeps the existing scheduling policies unchanged. We can either review it as a stacked PR or merge/cherry-pick it into `feat/qwen3-asr-chunking` before this PR lands—whichever is easier for the maintainers.

Hi thanks! we synced offline before, streaming input is out of scope for this pr. but personally I think we could support it :)

sglang_omni/serve/transcription_chunking.py
0xjeffro: Thanks, I think the right way is align the probe's coverage with what `load_audio` can decode.

sglang_omni/config/schema.py
0xjeffro: Hi thanks for contribute but looks like I can't click the "create stack" button cuz "Pull requests from forks cannot be added to stacks."? could we open a follow up new pr for this?

sglang_omni/serve/transcriptions.py
```diff
@@ -297,20 +365,158 @@ def _event_for(chunk: GenerateChunk) -> str | None:
...
+            return_when=asyncio.FIRST_COMPLETED,
+        )
+        if work_task in done:  # transcription completed first
+            return work_task.result()
+        # Bounded: the work task's cleanup (child cancels + engine aborts)
+        # keeps running inside the task even after we stop waiting; the bound
+        # only protects this handler from a cleanup that itself hangs.
+        await _cancel_task_bounded(work_task)
+        raise asyncio.CancelledError
+    finally:
```
SandyLuXY, line 369: If this wrapper is itself cancelled, this `finally` block stops only the disconnect watcher, leaving `_transcribe_audio_chunks` and its engine requests alive. Cancel and await `work_task` whenever it is incomplete, and cover parent-task cancellation in addition to `request.is_disconnected()`.

sglang_omni/models/qwen3_asr/config.py
```diff
@@ -15,6 +15,14 @@ class Qwen3ASRPipelineConfig(PipelineConfig):
 
     architecture: ClassVar[str] = "Qwen3ASRForConditionalGeneration"
 
+    # The model reads the whole clip as prompt tokens (~13/audio second), so a
+    # request stops fitting the engine context past ~115s. 60s keeps ~2x
+    # margin; longer uploads are chunked at the serving layer.
```
SandyLuXY: Please follow comment style: `note (xxx):`

sglang_omni/models/qwen3_asr/config.py
```diff
@@ -33,7 +41,15 @@ def generation_sglang_role_to_stage(cls) -> dict[str, str]:
...
                 "max_running_requests": 32,
-                "max_new_tokens": 128,
+                # Note: Cap on generated tokens per request; generation just stops there, so
+                # a too-small cap silently drops the transcript tail. Speech produces
+                # ~5 output tokens per audio second: 128 only covered ~25s, and a 60s
+                # chunk needs ~300 -- 640 doubles that. Safe to raise: context_length
+                # adds max_new_tokens on top of the audio budget (engine_builder.py),
+                # so it grows by the same amount and the audio that fits is untouched.
+                # If max_audio_clip_s is raised, raise this along with it
+                # (~5 tokens per audio second).
```
SandyLuXY: Please follow comment style: `note (xxx):`

sglang_omni/models/qwen3_asr/config.py
```diff
@@ -15,6 +15,14 @@ class Qwen3ASRPipelineConfig(PipelineConfig):
 
     architecture: ClassVar[str] = "Qwen3ASRForConditionalGeneration"
 
+    # The model reads the whole clip as prompt tokens (~13/audio second), so a
+    # request stops fitting the engine context past ~115s. 60s keeps ~2x
+    # margin; longer uploads are chunked at the serving layer.
+    audio_chunking: ClassVar[AudioChunkingConfig] = AudioChunkingConfig(
+        allow_audio_chunking=True,
+        max_audio_clip_s=60.0,
```
SandyLuXY, line 20: This 60-second scheduling choice is also used as the hard streaming limit, so it rejects previously valid native requests even though this PR sizes the model context for 1200 seconds.

sglang_omni/serve/transcription_chunking.py
```diff
@@ -0,0 +1,287 @@
...
+        window = self.energy_window_samples
+        window_count = int(region.shape[-1]) // window
+        if window_count < 1:
+            # The region cannot hold one window; fall back to the boundary.
+            return search_end
+        frames = region[: window_count * window].reshape(window_count, window)
+        # Mean square orders identically to RMS, so the square root is skipped.
+        energies = np.mean(np.square(frames.astype(np.float64)), axis=1)
+        quietest = int(np.argmin(energies))
+        return search_start + quietest * window + window // 2
```
SandyLuXY, line 133: This still chooses a cut when the whole search window is speech; with non-overlapping child requests, a word straddling the cut has no complete context on either side. Add a silence/VAD acceptance threshold and use overlap plus de-duplication for fallback cuts through speech.

sglang_omni/serve/transcription_chunking.py
```diff
@@ -0,0 +1,287 @@
...
+    """
+    joined = ""
+
+    for part in parts:
+        stripped = part.strip()
+        if not stripped:
+            continue
+        if joined and _is_spaced_script(joined[-1]) and _is_spaced_script(stripped[0]):
+            joined += " "
+        joined += stripped
```
SandyLuXY, line 286: Script ranges cannot decide whether a punctuation-leading fragment continues the previous token: for example this could produce `don 't`, `3 .14`. Preserve the model's seam text exactly or use overlap-aware reconciliation instead of inserting whitespace here.

sglang_omni/models/qwen3_asr/config.py
```diff
@@ -33,7 +41,15 @@ def generation_sglang_role_to_stage(cls) -> dict[str, str]:
...
-                "max_new_tokens": 128,
+                # Note: Cap on generated tokens per request; generation just stops there, so
+                # a too-small cap silently drops the transcript tail. Speech produces
+                # ~5 output tokens per audio second: 128 only covered ~25s, and a 60s
+                # chunk needs ~300 -- 640 doubles that. Safe to raise: context_length
+                # adds max_new_tokens on top of the audio budget (engine_builder.py),
+                # so it grows by the same amount and the audio that fits is untouched.
+                # If max_audio_clip_s is raised, raise this along with it
+                # (~5 tokens per audio second).
+                "max_new_tokens": 640,
```
SandyLuXY: SGLang uses the requested maximum in prefill/admission token budgeting even when generation stops early, so a fivefold fixed reservation can reduce batching efficiency. Consider deriving the default output allowance from decoded audio length (while honoring explicit user overrides), so 60-second chunks receive enough budget without charging short requests for 640 tokens. And run baseline/current performance validation with repeated runs after the policy changes.

SandyLuXY (changes requested):

sglang_omni/models/qwen3_asr/config.py
wirybeaver: @SandyLuXY Feel free to review #1390, which is a stacking branch base on this PR and @luojiaxuan left some comments. I think we can have a quick review of #1390 and then merge it onto #1347.

I have corrected the 60s behavior in #1390 companioned with other fixes that picked from original #1350, where I completed e2e validation with a 1201s sample on the RTX6000 a few days ago. I will move out the streaming input feature from #1390 soon as @0xjeffro mentioned before.

wirybeaver:
Agreed—chunked transcription streaming has been removed from #1390. I moved it to the separate `wirybeaver:qwen3-asr-chunked-streaming` branch, based on the latest `feat/qwen3-asr-chunking` because it directly depends on this PR's `AudioChunkingConfig`, `ChunkPlan`, and chunk request lifecycle. I will keep it as a follow-up rather than expanding #1347/#1390's scope.

sglang_omni/serve/transcriptions.py
0xjeffro: thanks! fixed this in [4fddea1](https://github.com/sgl-project/sglang-omni/pull/1347/commits/4fddea1d6944fd5de9a0e383d6902b131c21293e)

sglang_omni/models/qwen3_asr/config.py
0xjeffro: thx, implement this in [b64aacf](https://github.com/sgl-project/sglang-omni/pull/1347/commits/b64aacf792c66a82f407344cebf752b3b4358f98)

sglang_omni/serve/transcription_chunking.py
0xjeffro: thanks xinyu, but since this pr is already large, can we keep its scope to current chunking pipeline and impl VAD / overlap-aware reconciliation in a followup pr? because current method already aligns with the existing baseline.

SandyLuXY (approved):

sglang_omni/serve/transcriptions.py
```diff
@@ -297,20 +365,158 @@ def _event_for(chunk: GenerateChunk) -> str | None:
...
+        # in_flight holds every chunk whose engine request has not finished,
+        # including chunks whose local task got cancelled above us.
+        pending_engine_requests = sorted(in_flight)
+        for task in tasks:
+            task.cancel()
+        await asyncio.gather(*tasks, return_exceptions=True)
+        # Cancelling the local task does not stop the engine; abort by id.
+        for chunk_request_id in pending_engine_requests:
+            try:
+                await client.abort(chunk_request_id)
```
lijrjyan, line 341: These abort calls are awaited sequentially without a deadline and may leave cleanup incomplete. However, I agree this PR should stay focused on the chunking pipeline since now it is already large. I also notice that similar patterns exist elsewhere in the repository, so this might be addressed consistently in a follow-up or a broader cleanup PR?

lijrjyan (approved):

sglang_omni/serve/transcription_chunking.py
lijrjyan: #1352 already has implementations and tests for silence acceptance and hard-cut fallback. I’ll extract the reusable pieces into a follow-up PR.

sglang_omni/serve/transcription_chunking.py
```diff
@@ -0,0 +1,287 @@
...
+# Half-open ``[start, end)`` sample range of one chunk.
+Span = tuple[int, int]
+
+
+def needs_chunking(duration_s: float, config: AudioChunkingConfig) -> bool:
+    """Whether a probed clip duration calls for splitting.
+
+    ``duration_s <= 0`` means the duration probe failed (``soundfile`` could
+    not read the container). Such uploads keep today's behaviour rather than
+    paying a full decode just to find out how long they are.
```
Ratish1, line 33: lets avoid using backticks in docstrings and similarly in other places

sglang_omni/serve/transcriptions.py
```diff
@@ -297,20 +365,158 @@ def _event_for(chunk: GenerateChunk) -> str | None:
...
+        if not disconnect_task.done():
+            await _cancel_task_bounded(disconnect_task)
+
+
+def _probe_audio_duration(audio_bytes: bytes) -> float:
+    """Best-effort audio duration (seconds) from raw upload bytes."""
+    try:
+        import av
+
+        with av.open(io.BytesIO(audio_bytes), metadata_errors="ignore") as container:
```
Ratish1: This calls PyAV’s container and stream metadata probe synchronously inside the async request handler. Both transcription paths invoke it before engine admission, so the event loop cannot process other requests until the probe returns.

should we restore soundfile.info(BytesIO(audio_bytes)) as the synchronous fast path. If SoundFile raises or produces a non-positive duration, run the current PyAV metadata probe through await asyncio.to_thread. This preserves the broader container coverage without putting PyAV probing on the event loop.

sglang_omni/serve/transcriptions.py
```diff
@@ -185,19 +208,53 @@ async def create_transcription(
...
+        if needs_chunking(duration_s, chunking):
+            try:
+                # Check duration before decoding: a small compressed file can
+                # hold hours of audio, and decoding that eats gigabytes.
+                check_total_duration(duration_s, chunking)
+            except ValueError as exc:
+                raise HTTPException(status_code=400, detail=str(exc)) from exc
+            # Decode + split in a worker thread: decoding a long file is pure
+            # CPU and would stall the event loop.
+            plan = await asyncio.to_thread(plan_audio_chunks, audio_bytes, chunking)
```
Ratish1, line 122: This full-upload decode happens before the per-request chunk semaphore and outside any shared planning admission limit. max_concurrent_chunks bounds child engine submissions, but it does not bound how many HTTP requests may simultaneously decode and retain complete waveforms.

Concurrent long uploads can therefore accumulate decoded waveforms in the HTTP process even though every individual upload satisfies max_total_audio_s. Long-audio planning needs shared admission, such
as a dedicated bounded executor, semaphore, or decoded-sample budget. Otherwise max_total_audio_s should not be described as a complete memory guard because it limits individual uploads rather than
aggregate resident planning state.

sglang_omni/models/qwen3_asr/config.py
```diff
@@ -14,6 +15,11 @@ class Qwen3ASRPipelineConfig(PipelineConfig):
     """Single-stage batched ASR pipeline for Qwen3-ASR checkpoints."""
 
     architecture: ClassVar[str] = "Qwen3ASRForConditionalGeneration"
+    audio_chunking: ClassVar[AudioChunkingConfig] = AudioChunkingConfig(
```
Ratish1, line 18: audio_chunking is a ClassVar, so Pydantic does not expose it as a model field. It is absent from model_dump(), which prevents dotted CLI overrides from resolving audio_chunking.*, while adding it to a
YAML configuration is rejected by extra="forbid". The runtime therefore always uses the class-level Qwen policy even though the cookbook presents these values as settings that can be changed.

If these limits are intended to be operator-configurable, this should be an instance field or have an explicit override mechanism. If they are intentionally fixed model metadata, the cookbook should describe them as fixed model defaults rather than configurable settings.

0xjeffro: Thanks Ratish good point! for this pr I prefer just change the cookbook to describe them as fixed default value since we only support Qwen3ASR at this time. But the problem is real, and I will note this down in the future we need to think carefully about these fields, which one is inherent in the model and which are configurable. :)

sglang_omni/serve/transcriptions.py
0xjeffro: Hi Ratish thanks for your review. I think soundfile would only be a bit faster, and also, we use PyAV now is aim to directly stay aligned with load_audio (PyAV can read M4A/WebM headers, see https://github.com/sgl-project/sglang-omni/pull/1347#discussion_r3733516936). So I'd preder to keep PyAV and just put it behind asyncio.to_thread.

0xjeffro: Besides, I noticed in CI stage 2 looks have a concurrency regression caused by this: we need the probe to decide whether chunking is needed, every request pays for it. We can pass the probe result down, so assemble_speech_to_text_response doesn't have to recompute it.

Issue #1267, opened 2026-08-02
[ASR Refactor] Roadmap and PR Tracker


SandyLuXY:
This is the ASR-refactor roadmap tracker.

New ASR refactor PRs should use the title prefix `[ASR Refactor]` and link this issue in the PR body.

## Source docs

- RFC / design doc: [SGLang-Omni ASR Refactor RFC](https://ljp2lkmtlez9.jp.larksuite.com/wiki/ISNgw31DjixPzNkEFcwjF2Ohpyh?from=from_copylink)

### Active

- [x] F1 Auto chunking for long audio (Qwen3-ASR) — Depends on M1
   - #1347 
   - #1350 
   - #1352 
- [x] F2  /v1/audio/translations — Depends on M5 
   - #1394 
   - #1453  
   - #1631 
   - #1645 
- [x] F3 Whisper prompt → prev-context tokens — Depends on M1, M3
   - #1339 
   - #1654 
- [x] F4 — Register Whisper into CI
- [x] F5 — Integrate and optimize pipecat-ai/stt-benchmark-data
   - #1667 
- [ ] F6 — Streaming support for Qwen3-ASR
   - #1671 

### Landed

- [x] M1 Shared audio preparation — Issue #1268 — PR #1274
- [x] M2 SGLang generation engine builder — Issue #1269 — PR #1277 
- [x] M3 Pre-LM encoder service — Issue  #1270 — PR #1273 
- [x] M4 Token text streaming — Issue #1271 — PR #1275 
- [x] M5 Transcription serve extraction — Issue #1272 — PR #1285 

### Closed / superseded context

None yet.

## Scope constraints

- Shared framework code owns reusable mechanics; model directories own model semantics and policy.
- Preserve existing input compatibility, outputs, errors, streaming behavior, initialization order, resource lifetime, and performance.
- Do not branch shared code on model names.
- Do not mix bug fixes, behavior changes, or new features into these migrations; track them in separate issues and PRs.

AkazaAkane:
Hi, I am claiming M3.

Dayuxiaoshui:
@SandyLuXY Hi, I am claiming M2

guozhihao-224:
@SandyLuXY  Hi, I am claiming M4

LG-0927:
> [@SandyLuXY](https://github.com/SandyLuXY) Hi, I am claiming M4

Hi @guozhihao-224, thanks for your interest in M4.

[...]

PR #1350, opened 2026-08-05
[ASR Feature] Auto-chunk long Qwen3-ASR audio

wirybeaver:
## Summary

- size Qwen3-ASR context for the official 1,200-second native input limit plus the upstream 4,096-token generation default
- keep audio at or below 1,200 seconds as one native model request
- auto-chunk longer uploads at low-energy, non-overlapping boundaries with hard 1,200-second maximums and 0.5-second minimum tails
- stitch blocking and streaming `/v1/audio/transcriptions` results in order, fail the whole upload on a chunk error, and abort active child streams on disconnect
- document the model-owned long-audio behavior

## Why 1,200 seconds, not 30/60 seconds

The original F1 note in #1267 described the old silent 30-second truncation. After #1176, preprocessing no longer truncates at Whisper's 30-second window. The follow-up issue comment identifies the remaining problem as Qwen3-ASR context sizing and points to the official wrapper's `MAX_ASR_INPUT_SECONDS = 1200` behavior.

This therefore deliberately differs from draft #1347's 60-second threshold: Qwen3-ASR processes up to 1,200 seconds natively and chunks only above the model-owned limit.

## Behavior

- no overlap or de-duplication: the official wrapper uses non-overlapping chunks and direct concatenation
- no timestamp remapping: Qwen3-ASR does not currently expose timestamped segments through this serving path
- model policy is default-disabled globally and enabled only by `Qwen3ASRPipelineConfig`

## Validation

- [x] `uv run pytest -q tests/unit_test/qwen3_asr tests/unit_test/preprocessing/test_transcription.py tests/unit_test/serve/test_transcription_chunking.py tests/unit_test/serve/test_openai_api.py` — 131 passed, 1 skipped
- [x] `uvx pre-commit run --from-ref upstream/main --to-ref HEAD`
- [x] Vast.ai GPU VM Docker end-to-end validation (no NCU)
  - public 48GB validation artifacts: https://gist.github.com/wirybeaver/05206d141f06fdf8bdced5739399feba
  - GPU-validated pre-rebase source: `e6fb693fa49a04e7554cf09303cd47e042a1be88`
  - image: `sha256:9e4294c08e6a9a06f6400aacaad161cac8d49dc655d91fb37da7ccf2bc99006a`
  - GPU: NVIDIA RTX 6000 Ada Generation, 49,140 MiB
  - server: concurrency 1, CUDA graph max batch 1, `--mem-fraction-static 0.5`
  - 125-second native request: HTTP 200, non-empty transcript, 101.80 s, usage 125 s
  - 1,201-second request: HTTP 200, non-empty transcript, 50.27 s, usage 1,201 s; server logs show ordered `chunk-0` and `chunk-1`, both successful
  - peak sampled GPU memory: 39,403 MiB for the 1,201-second request

SGLang's automatic KV-cache sizing consumed the additional VRAM and left less than 1 GiB free during the long prefill on 48 GiB GPUs. Pinning `mem_fraction_static=0.5` reserves workspace for the audio encoder; this is required deployment tuning for this maximum-length validation rather than a chunking failure.

## Related

- Tracks F1 in #1267
- Related draft: #1347

### DIFF
--- a/sglang_omni/models/qwen3_asr/config.py
+++ b/sglang_omni/models/qwen3_asr/config.py
@@ -5,7 +5,8 @@ from __future__ import annotations
 
 from typing import ClassVar
 
-from sglang_omni.config import PipelineConfig, StageConfig
+from sglang_omni.config import AudioChunkingConfig, PipelineConfig, StageConfig
+from sglang_omni.models.qwen3_asr.audio_lengths import QWEN3_ASR_MAX_INPUT_SECONDS
 
 _PKG = "sglang_omni.models.qwen3_asr"
 
@@ -14,6 +15,10 @@ class Qwen3ASRPipelineConfig(PipelineConfig):
     """Single-stage batched ASR pipeline for Qwen3-ASR checkpoints."""
 
     architecture: ClassVar[str] = "Qwen3ASRForConditionalGeneration"
+    audio_chunking: ClassVar[AudioChunkingConfig] = AudioChunkingConfig(
+        allow_audio_chunking=True,
+        max_audio_clip_s=float(QWEN3_ASR_MAX_INPUT_SECONDS),
+    )
 
     @classmethod
     def mem_fraction_role_to_stage(cls) -> dict[str, str]:
@@ -33,7 +38,7 @@ class Qwen3ASRPipelineConfig(PipelineConfig):
             factory_args={
                 "device": "cuda:0",
                 "max_running_requests": 32,
-                "max_new_tokens": 128,
+                "max_new_tokens": 4096,
                 "request_build_max_workers": 8,
                 "request_build_max_pending": 16,
                 "enable_pre_lm_encoder": True,
--- a/docs/cookbook/qwen3_asr.md
+++ b/docs/cookbook/qwen3_asr.md
@@ -75,7 +75,24 @@ print(resp.json()["text"])
 `verbose_json` is accepted, but currently returns the same minimal JSON shape as `json`:
 `{"text": "..."}`.
 
-`max_new_tokens` is supported inside the model request builder, but the public transcription endpoint does not currently expose it as a form field. The route uses the ASR stage default unless the pipeline is configured another way.
+`max_new_tokens` can be supplied as a multipart form field. The route uses the
+Qwen3-ASR stage default of 4,096 when it is omitted.
+
+## Long Audio
+
+Qwen3-ASR accepts up to 1,200 seconds of audio in one native model request.
+This is a model-owned limit rather than Whisper's 30-second feature-extractor
+window. Audio at or below the limit is therefore processed without serving-layer
+chunking.
+
+Longer uploads are decoded once, split at low-energy boundaries into
+non-overlapping chunks of at most 1,200 seconds, transcribed in order, and
+stitched into one response. This applies to blocking and streaming
+transcriptions. Chunk failures fail the whole upload instead of returning a
+transcript with a silent gap.
+
+Qwen3-ASR does not currently produce timestamped segments, so chunk offsets are
+not exposed in `verbose_json`.
 
 ## Benchmarking
 
--- a/sglang_omni/config/__init__.py
+++ b/sglang_omni/config/__init__.py
@@ -14,6 +14,7 @@ from sglang_omni.config.process_overrides import (
 )
 from sglang_omni.config.runtime import resolve_stage_factory_args
 from sglang_omni.config.schema import (
+    AudioChunkingConfig,
     CommConfig,
     EndpointsConfig,
     ParallelismConfig,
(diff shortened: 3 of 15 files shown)


wirybeaver:
Hi @SandyLuXY, when you have a chance, could you please review the assessment below? Feel free to triage to other maintainers that have bandwidth to take a look.

## PR #1350 and PR #1352

| Area | PR #1350 | PR #1352 |
|---|---|---|
| Native audio limit | Uses **1,200s**, following the latest clarification | Uses **30s**, following the earlier understanding of the requirement |
| Context sizing | Accounts for a 1,200s prompt plus 4,096 output tokens | Does not currently include the corresponding context-size adjustment |
| Boundary selection | Selects a low-energy 100ms window near the limit | Provides explicit sustained-silence detection with hard-cap fallback |
| Transcript joining | Uses direct concatenation, consistent with the official wrapper | Trims chunks and applies language-aware spacing |
| Streaming | Forwards child deltas incrementally and aborts the active child on disconnect | Completes the child requests before emitting the combined SSE response |
| Configuration | Uses a typed, centralized `AudioChunkingConfig`; no CLI options | Provides CLI enable/disable and maximum-duration options |
| Disabled oversized input | Does not currently return an explicit 400 | Provides an explicit 400 and worker-side enforcement |
| Validation | GPU-tested with 125s native and 1,201s chunked inputs | GPU-tested with 36–60s inputs |

### Engineering assessment

I believe **#1350 currently provides the stronger foundation** because:

- It addresses the latest understanding of the root cause: context sizing and the 1,200-second native limit.
- It centralizes the policy in a typed configuration rather than passing individual values through several layers.
- It preserves incremental streaming behavior.
- It follows the official wrapper's direct-concatenation behavior.

There are still areas in #1350 that may be worth refining:

- Placing `AudioChunkingConfig` at the generic pipeline level may be broader than necessary for the current use case.
- `sample_rate` is duplicated between `ChunkPlan` and each `ChunkSpan`.

### Specification assessment

Based on the latest clarification, **#1350 appears to align more closely with the intended 1,200-second behavior**.

#1352 also contains several valuable ideas that would strengthen this work:

1. Explicit sustained-silence detection with hard-boundary fallback.
2. A clear error when chunking is disabled for an oversized input.
3. Optional CLI overrides, if maintainers would like that configurability.

For consistency with the latest clarification and official wrapper, I suggest retaining the 1,200-second default, incremental streaming, and direct transcript concatenation unless maintainers prefer otherwise.

**Suggested path forward:** use #1350 as the base and, with agreement from the contributors and maintainers, incorporate the applicable strengths of #1352 and #1347.

@lijrjyan, author of #1352, and @0xjeffro, author of #1347: thank you for your work on this feature. Your contributions and experience would be very welcome here. If you are interested, please feel free to contribute directly to this branch or suggest changes. I would be glad to preserve attribution and include co-author credit for any incorporated work. A joint review would also be greatly appreciated.

wirybeaver:
Since maintainers have started reviewing #1347, I opened #1390 as a focused correction stacked directly on that branch. It carries forward the F1 semantics and validation from this implementation, including native processing through 1,200 seconds, chunking above that limit, and the 1,201-second Vast.ai RTX 6000 48 GB GPU validation.

I will likely close this PR to avoid duplicate effort and continue the work through the stacked path. The public GPU validation artifacts here will remain available for reference. Thank you to everyone who reviewed and contributed to this implementation.

sglang_omni/models/qwen3_asr/config.py
```diff
@@ -33,7 +38,7 @@ def generation_sglang_role_to_stage(cls) -> dict[str, str]:
             factory_args={
                 "device": "cuda:0",
                 "max_running_requests": 32,
-                "max_new_tokens": 128,
+                "max_new_tokens": 4096,
```
luojiaxuan, line 41: A fixed 4096-token default is charged to every short request during admission and will reduce batching throughput; keep the short-request default and derive a larger budget from audio duration or require an explicit override.

sglang_omni/models/qwen3_asr/config.py
```diff
@@ -14,6 +15,10 @@ class Qwen3ASRPipelineConfig(PipelineConfig):
     """Single-stage batched ASR pipeline for Qwen3-ASR checkpoints."""
 
     architecture: ClassVar[str] = "Qwen3ASRForConditionalGeneration"
+    audio_chunking: ClassVar[AudioChunkingConfig] = AudioChunkingConfig(
+        allow_audio_chunking=True,
+        max_audio_clip_s=float(QWEN3_ASR_MAX_INPUT_SECONDS),
```
luojiaxuan, line 20: The validation needed a reduced static memory fraction on a 48 GB GPU for this path, so enabling the 1200-second limit by default needs a safe capacity policy for smaller supported GPUs.

luojiaxuan (changes requested):
The default token and memory budgets need to remain safe for short requests and supported GPUs.

wirybeaver:
Check #1390 that is based on top of #1347

PR #1352, opened 2026-08-05
[ASR Feature] Auto-chunk long Qwen3-ASR audio at silence boundaries

lijrjyan:
## Motivation

F1 in #1267: long Qwen3-ASR uploads should be auto-chunked instead of failing at the deployed per-request context, and disabling chunking must never silently truncate.

This implements the F1 mechanics as **non-overlapping, silence-boundary chunking with direct concatenation** — the same shape as the official Qwen wrapper. An earlier revision of this draft carried a chunk-overlap + de-duplication layer; it was removed after validation showed why it does not belong in the default path:

- at the native 1,200 s window (where this policy is headed once the capacity line lands), a 5 s boundary-search window of real speech essentially always contains a pause, so the "no silence candidate" case the overlap was designed for does not occur in practice;
- the one scenario that defeats energy-based silence detection — noisy audio — is also where exact-match de-duplication is least reliable.

The overlap machinery is archived on a separate branch and can be proposed later as an opt-in feature; it is intentionally not part of this PR.

## Modifications

- `sglang_omni/preprocessing/transcription.py` — shared mechanics only: silence-aware chunk planning (hard cap, exact sample boundaries, boundaries never extend past the hard end, fixed-window fallback, strict non-overlap invariant `next.start == previous.end`) and language-aware transcript joining (no separator inserted between CJK segments, single space between Latin words).
- `sglang_omni/models/qwen3_asr/chunking.py` — model-owned policy: default-on, 30 s window matching what the current deployed context sizing admits; deliberately a single constant so it lifts to the native 1,200 s envelope once capacity lands (#1353). No model-name branching in shared code.
- `sglang_omni/serve/openai_api.py` + `launcher.py` — fan-out coordinator: one upload → N chunk completions → ordered join; JSON/text and SSE; any chunk failure aborts the remaining chunks and fails the whole request (a partial transcript is never returned as success). Policy reaches the endpoint through the standard `runtime_overrides` resolver, so the endpoint and the worker factory always agree.
- `sglang_omni/cli/serve.py` — `--asr-auto-chunk/--no-asr-auto-chunk`, `--asr-chunk-max-seconds`; validation has a single source (the model-owned validator), translated to CLI errors.
- Explicit client error (with actual duration and window seconds) when chunking is off and the input exceeds the window — enforced at the endpoint and again in the request builder.
- Requests at or below the window keep the exact pre-change single-request path, locked by tests.

Scope reduction, stated upfront: timestamp remapping (`verbose_json`/`srt`/`vtt`) is not implemented — the current serving path returns minimal JSON only; left for a follow-up.

## Validation

CPU unit suites: `tests/unit_test/preprocessing` + `qwen3_asr` + `serve` (chunk/endpoint suites) — 122 passed, 1 skipped; `ruff check` / `ruff format --check` clean on changed files.

GPU (this head, RTX A6000 48 GB, `Qwen/Qwen3-ASR-1.7B` rev `7278e1e`): server healthy; a 60 s English clip is chunked and transcribed coherently; a 36 s Chinese clip returns exact `usage.seconds` with no characters lost at the chunk boundary (the boundary shows only a sentence-split seam, the expected direct-concatenation behavior).

Boundary behavior (A100-SXM-64GB, 3 repeats per cell, single-forward no-chunk output as reference): silence-boundary non-overlapping chunking produced **zero boundary word errors in both English and Chinese** — the direct evidence behind removing the overlap layer from the default path.

## Related work / overlap disclosure

Canonical thread: #1267 (F1). Two other drafts are in flight; this is submitted as a comparison/consolidation candidate, not a competing claim:

- #1347 (@0xjeffro) — serving-layer chunking at the deployed clip limit, non-overlapping. Same design family; this draft adds the model-owned policy split, explicit 400, all-or-nothing failure semantics, and resolver-consistent configuration.
- #1350 (@wirybeaver) — resizes context to the native 1,200 s envelope and chunks only above it. Same concatenation semantics; the capacity/context work is deliberately **not** part of this PR — that is the separate long-input line (#1173, #1176 merged, #1193, #1353), which this policy constant is designed to follow.

Happy to fold any of these pieces into whichever PR the maintainers pick as canonical.

### DIFF
--- a/sglang_omni/cli/serve.py
+++ b/sglang_omni/cli/serve.py
@@ -5,7 +5,6 @@ from typing import Annotated, Literal, NoReturn
 
 import typer
 from sglang_omni.models.qwen3_asr.chunking import (
-    QWEN3_ASR_CHUNK_OVERLAP_SECONDS,
     QWEN3_ASR_CHUNK_WINDOW_SECONDS,
     validate_qwen3_asr_chunking,
 )
@@ -876,15 +875,12 @@ def apply_asr_chunking_cli_overrides(
     *,
     asr_auto_chunk: bool | None,
     asr_chunk_max_seconds: float | None,
-    asr_chunk_overlap_seconds: float | None,
 ) -> PipelineConfig:
     updates: dict[str, object] = {}
     if asr_auto_chunk is not None:
         updates["asr_auto_chunk"] = bool(asr_auto_chunk)
     if asr_chunk_max_seconds is not None:
         updates["asr_chunk_max_seconds"] = float(asr_chunk_max_seconds)
-    if asr_chunk_overlap_seconds is not None:
-        updates["asr_chunk_overlap_seconds"] = float(asr_chunk_overlap_seconds)
     if not updates:
         return pipeline_config
 
@@ -894,7 +890,7 @@ def apply_asr_chunking_cli_overrides(
     if not matching_stages:
         _raise_unsupported_flag(
             pipeline_config,
-            "--asr-auto-chunk/--asr-chunk-max-seconds/--asr-chunk-overlap-seconds",
+            "--asr-auto-chunk/--asr-chunk-max-seconds",
         )
     for stage in matching_stages:
         effective = dict(stage.factory_args or {})
@@ -907,12 +903,7 @@ def apply_asr_chunking_cli_overrides(
                     effective.get(
                         "asr_chunk_max_seconds", QWEN3_ASR_CHUNK_WINDOW_SECONDS
                     )
-                ),
-                float(
-                    effective.get(
-                        "asr_chunk_overlap_seconds", QWEN3_ASR_CHUNK_OVERLAP_SECONDS
-                    )
-                ),
+                )
             )
         except ValueError as exc:
             raise typer.BadParameter(str(exc)) from exc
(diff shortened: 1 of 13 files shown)


wirybeaver:
Thank you for the work on this implementation, especially the sustained-silence boundary detection, explicit oversized-input handling, and configuration options.

Based on the latest clarification in #1267, I believe #1350 is the better base for consolidation because it includes the corrected 1,200-second native limit and context sizing, chunks only above that limit, and preserves incremental streaming and direct transcript concatenation.

I would be very happy to collaborate on the same branch and preserve co-author attribution for incorporated work. Follow-up PRs that strengthen the shared implementation—particularly around VAD boundaries, disabled-path errors, or configuration—would also be very welcome.

luojiaxuan (commented):
This still uses the superseded 30-second policy and conflicts with the current serving path; please consolidate reusable tests or chunk planning into the selected F1 PR instead of rebasing a third endpoint implementation.

lijrjyan:
> This still uses the superseded 30-second policy and conflicts with the current serving path; please consolidate reusable tests or chunk planning into the selected F1 PR instead of rebasing a third endpoint implementation.

Got it. I noticed Sandy raised the silence-acceptance and hard-cut fallback concerns in #1347, and this implementation already has related logic and tests.

Since #1347 is already large, I’ll follow @0xjeffro and @SandyLuXY in keeping it focused on the chunking pipeline. I’ll move these reusable parts into a follow-up PR instead. I will close it once the follow-up PR is up.

lijrjyan:
Closing: superseded by #1347, which landed the chunking path.

PR #1353, opened 2026-08-05
fix(qwen3-asr): size capacity for the native 1200-second envelope

lijrjyan:
## Motivation

#1176 removed the silent 30-second feature truncation, but the per-request capacity is still sized from Whisper's extraction-window constant: `create_sglang_qwen3_asr_executor` derives `context_length` from `feature_extractor.nb_max_frames // 2`, which is not a Qwen3-ASR model limit. The official Qwen wrapper accepts up to `MAX_ASR_INPUT_SECONDS = 1200` in one forward and uses a 4,096-token generation default. On current main, a long clip that survives preprocessing is rejected (or its output budget silently exhausted at the flat stage default) by context arithmetic unrelated to the model.

This PR makes the capacity contract follow the model:

- one model-local duration contract (`max_audio_s`, default 1,200 s) sizes context, prefill, and the automatic output budget;
- no silent truncation on either side: input that cannot fit is explicitly rejected, and the automatic output budget scales with duration instead of stopping dense long speech at a flat ceiling;
- no chunking/stitching is added here — this is the capacity layer underneath the in-flight F1 chunking drafts (#1347, #1350, #1352).

## Modifications

- `request_builders.py`: automatic output budget `max(4096, ceil(10 × duration_s))` — keeps the upstream 4,096 floor, avoids a flat long-audio ceiling; precedence stays request > operator default > automatic, with explicit validation of `max_new_tokens`. `enforce_request_limits=True` so oversized explicit budgets are clamped to real remaining capacity and an input that cannot fit is rejected explicitly (never truncated).
- `stages.py`: `context_length = max_audio_tokens + tokenized prompt overhead + max output budget + 2` (the two scheduler reserve slots are measured, see below); `max_prefill_tokens = chunked_prefill_size = full max input` so the multimodal encoder is entered once per request while the mm embedding cache is disabled; `context_length` stays operator-overridable.
- `config.py`: stage default switches from the flat `max_new_tokens=128` (a silent output-truncation source) to `max_audio_s=1200.0`.
- Cookbook: documents the long-audio contract and that the automatic budget is a bounded guard, not an EOS guarantee.

## Validation

CPU unit suites on the rebased head: `tests/unit_test/qwen3_asr` + `preprocessing` + `scheduling` — 163 passed, 1 skipped. Tests pin the exact sizing arithmetic (15,600 audio tokens at 1,200 s; context = input + budget + 2) and the budget precedence/validation matrix.

Runtime evidence (RTX A6000 48 GB, `Qwen/Qwen3-ASR-1.7B` rev `7278e1e`, SGLang 0.5.16):

- a generated 1,200 s clip is admitted as **one prefill batch of 15,613 tokens** (15,600 audio + 13 prompt overhead) and completes with HTTP 200 — the "+2" reserve is exact at the boundary;
- `max_new_tokens=100000` on a short clip is clamped by the scheduler and completes instead of erroring;
- with a flat 4,096 budget, a dense repeated-speech 1,200 s stress clip terminates at the ceiling (`finish_reason=length` path), while the duration-scaled budget does not — the reason the automatic budget scales;
- **negative result, relevant to #1350**: with `chunked_prefill_size` left at 4,096/8,192 while context is enlarged, the 1,200 s request re-enters the audio encoder per prefill chunk (mm embedding cache is disabled) and **OOMed on the 48 GB A6000**; full-request prefill sizing is what made the envelope work. Configs retained as labelled negative experiments.

On an RTX 4090D 48 GB (driver 570.133.20, torch 2.11.0+cu128), a 20-fixture 30/60/120/300 s matrix (5 deterministic concatenated fixtures per bucket, sequential, `max_total_tokens=8192` profile) passed 20/20 with `usage.seconds` exact and near-linear latency; a 650 s probe (8,463 input tokens) was explicitly rejected with the capacity message rather than truncated. The rejection currently surfaces as HTTP 500 — converging with #1193 on 400 classification.

Not validated here: 24/32 GB devices, high-concurrency throughput vs the old 128-token default (larger per-request admission reservations), WER parity on natural long-form corpora. Stated as open items, not claims.

**A100-SXM-64GB concurrency validation** (torch 2.11.0+cu130, same venv both arms, heads `51e62f6` (main) vs `31dd4f1` (this branch), fresh server per arm, 3 repeats each): 128 distinct 5 s clips at concurrency 32 — main 48.0–50.2 req/s vs branch 42.1–51.0 req/s; **ranges overlap, no separable throughput regression** from the larger admission reservation. Peak device memory 57.0 GB (main) vs 43.0 GB (branch). Mixed traffic (1×1,200 s + 16×5 s, 3 repeats): short-request p50 latency delta +0.10 / +0.18 / +0.36 s vs a no-long-request control — head-of-line impact present but sub-second on this hardware class.

## Related work / overlap disclosure

- #1173 — behavior thread; #1176 — merged base (this PR is its capacity follow-up).
- #1193 (@wirybeaver) — proactive context validation + HTTP 400 mapping; complementary, and the 650 s probe above is direct evidence for it.
- #1350 (@wirybeaver) — includes its own context resizing (`prompt + 4096 + 1`). Differences here: duration-scaled output budget (vs flat 4,096), full-request prefill sizing (the A6000 OOM negative result above bites otherwise), and the measured `+2` reserve (vs `+1`). Happy to consolidate either direction — the chunking layer in #1350/#1347/#1352 sits on top of whichever capacity contract lands.
- #1267 — F1 depends on this line for a >30 s native window.

### DIFF
--- a/docs/cookbook/qwen3_asr.md
+++ b/docs/cookbook/qwen3_asr.md
@@ -78,13 +78,11 @@ print(resp.json()["text"])
 
 ## Long Audio
 
-The default stage is sized for Qwen3-ASR's native 1200-second single-forward
-envelope. It preserves the complete mel sequence (`truncation=False`) and does
-not split or stitch transcripts. When `max_new_tokens` is omitted, the output
-budget keeps the upstream Qwen vLLM 4096-token floor and grows to 12000 tokens
-at 1200 seconds so that the upstream default does not become a flat long-audio
-ceiling. The context and prefill limits include the corresponding 15600 audio
-tokens plus the actual tokenized prompt overhead.
+The default stage supports Qwen3-ASR's native 1200-second single-forward
+envelope without truncation or transcript stitching. Its automatic output
+budget keeps the upstream 4096-token floor and grows to 12000 tokens at that
+boundary. Context and prefill capacity include the corresponding 15600 audio
+tokens and tokenized prompt overhead.
 
 Operators can pin `max_new_tokens` or change the guaranteed native envelope
 with the ASR stage's `max_audio_s` factory argument. A request-level
--- a/docs/cookbook/qwen3_asr.md
+++ b/docs/cookbook/qwen3_asr.md
@@ -71,11 +71,26 @@ print(resp.json()["text"])
 | `language` | string | `en` | Language hint; `zh`/`cn` select Chinese, other values use English prompting |
 | `response_format` | string | `json` | `json`, `verbose_json`, or `text` |
 | `temperature` | float | `0.01` effective | Sampling temperature; `0` is converted to near-greedy `0.01` |
+| `max_new_tokens` | integer | automatic | Explicit positive transcription-token budget |
 
 `verbose_json` is accepted, but currently returns the same minimal JSON shape as `json`:
 `{"text": "..."}`.
 
-`max_new_tokens` is supported inside the model request builder, but the public transcription endpoint does not currently expose it as a form field. The route uses the ASR stage default unless the pipeline is configured another way.
+## Long Audio
+
+The default stage is sized for Qwen3-ASR's native 1200-second single-forward
+envelope. It preserves the complete mel sequence (`truncation=False`) and does
+not split or stitch transcripts. When `max_new_tokens` is omitted, the output
+budget keeps the upstream Qwen vLLM 4096-token floor and grows to 12000 tokens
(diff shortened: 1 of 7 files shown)


lijrjyan:
Closing: superseded by #1347/#1390, which landed the native 1,200 s capacity sizing.

PR #1355, opened 2026-08-06
[Qwen3-Omni] Fix 555-sample audio gap at every streamed code2wav chunk boundary

### LINKED ISSUE #1354: [Bug][Qwen3-Omni] Streamed code2wav drops 555 samples at every chunk boundary, causing audible clicks
## Symptom

Streamed Qwen3-Omni audio has an audible click at chunk boundaries. @RajPatel17  reproduced this and saved the raw PCM before it reached the browser. The waveform shows abrupt jumps: one chunk ends at one audio level and the next chunk begins at a different level.

## Root cause

Each streaming decode in `Code2WavScheduler` runs the vocoder on `[left_context | new_frames]` and trims `context * total_upsample` samples from the left:

https://github.com/sgl-project/sglang-omni/blob/main/sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py#L235-L237

The problem: the vocoder's output is always 555 samples shorter than `window_frames * total_upsample`. The `Qwen3OmniMoeCausalTransConvNet` layers in the HF model trim `kernel - stride` samples on the right at each decoder block. With upsample rates (8, 5, 4, 3) the deficit compounds to 8 → 45 → 184 → 555. This is structural and independent of weights or input length.

In a single full decode this only truncates the very end of the audio, which is harmless. In chunked decode it means **every chunk loses its last 555 samples (~23 ms at 24 kHz)**. The next chunk's emitted audio starts 555 samples later in time, so the stream has a ~23 ms hole at every boundary. With `stream_chunk_size=10` that is a click every 0.8 s of audio.

The upstream `Qwen3OmniMoeCode2Wav.chunked_decode` in transformers has the same flaw, but its default chunk size is 300 frames, so the gap only appears every 6 s and is much less noticeable.

## Evidence

Small random-init model on CPU, comparing full decode against the scheduler's chunking scheme (60 frames, chunk 10, left context 25, `total_upsample=1920`):

```
full decode:    114645 samples
streamed concat: 111870 samples   # 6 chunks x 555 missing = 3330
```

The missing samples are not recoverable from the emitted stream. They fall in the time range `[boundary - 555, boundary)`, which no chunk ever emits.

## Suggested fix

The missing samples do exist in the next chunk's decode window, at offset `context * total_upsample - 555`. So the trim is off by the deficit. Emitting the last `new_frames * total_upsample` samples instead of trimming a fixed left offset fixes both the gap and the length:

```python
wav = wav[..., -(end - start) * self._total_upsample :]
```

Verified on the same setup: streamed output becomes sample-exact in length (114645 == full), and the max error near every boundary drops to ~1e-8 (float noise from truncated attention/conv context). The 555-sample time shift is gone.

Notes for the fix:

- The deficit can be derived from `config.upsample_rates` instead of hardcoding 555, or avoided entirely by the keep-last-N form above.
- The batched path (`run_step`) has the same trim and needs the same change.
- The first chunk (context 0) still ends 555 samples short until the next chunk covers it; the final chunk's tail truncation matches full-decode behavior.

## Repro script

```python
import torch, numpy as np
from transformers.models.qwen3_omni_moe.configuration_qwen3_omni_moe import Qwen3OmniMoeCode2WavConfig
from transformers.models.qwen3_omni_moe.modeling_qwen3_omni_moe import Qwen3OmniMoeCode2Wav

torch.manual_seed(0)
cfg = Qwen3OmniMoeCode2WavConfig(hidden_size=64, num_hidden_layers=2, num_attention_heads=4,
                                 num_key_value_heads=2, intermediate_size=128, decoder_dim=64)
model = Qwen3OmniMoeCode2Wav(cfg).eval()
U = int(model.total_upsample)  # 1920, same as the real checkpoint
T, CHUNK, CTX = 60, 10, 25
codes = torch.randint(0, cfg.codebook_size, (1, cfg.num_quantizers, T))
with torch.no_grad():
    full = model(codes).reshape(-1)
    parts, start = [], 0
    while start < T:
        end = min(start + CHUNK, T)
        ctx = min(CTX, start)
        wav = model(codes[..., start - ctx:end]).reshape(-1)
        parts.append(wav[ctx * U:])          # current scheduler trim
        start = end
    stream = torch.cat(parts)
print(len(full), len(stream))  # 114645 vs 111870
```

edwingao28:
## Motivation

Fixes #1354.

Streamed Qwen3-Omni audio clicks at every chunk boundary. The vocoder's output is always 555 samples shorter than `window_frames * total_upsample` (the `CausalTransConvNet` right-trims compound to 8 → 45 → 184 → 555 across the decoder blocks). The scheduler trims a fixed `context * total_upsample` from the left, so every streamed chunk silently loses its last 555 samples (~23 ms at 24 kHz). The next chunk resumes 555 samples later in time, leaving a hole at each boundary — an audible click every 0.8 s with `stream_chunk_size=10`.

## Modifications

Keep the last `new_frames * total_upsample` samples of each decode window instead of trimming a fixed left offset, in both the serial path (`decode_delta`) and the batched path (`run_step`). The samples previously dropped from chunk N sit inside chunk N+1's decode window at offset `context * total_upsample - 555`, so the keep-last form recovers them exactly.

- `sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py`: trim fix in `decode_delta` and `run_step`.
- `tests/unit_test/fixtures/qwen_fakes.py`: `FakeCode2WavModel` gains an `output_deficit` knob to model the real vocoder's short output.
- Regression tests in `test_code2wav.py` and `test_code2wav_batching.py`: with a deficit model, each emitted chunk now has exactly `new_frames * total_upsample` samples.

## Validation

before & after fix audio output:
[after_fix.wav](https://github.com/user-attachments/files/30766283/after_fix.wav)
[before_fix.wav](https://github.com/user-attachments/files/30766286/before_fix.wav)

Small random-init model, 60 frames, chunk 10, context 25 (`total_upsample=1920`):

- Before: streamed concat 111870 samples vs 114645 for full decode (6 chunks × 555 missing).
- After: streamed concat 114645 samples, sample-exact length; max error near every boundary ~1e-8 (float noise).

Real checkpoint (Qwen3-Omni-30B-A3B-Instruct, H200): generated one sentence, decoded the same talker codes with the old and new trim. The old stream drops 555 samples per boundary; the new stream is sample-aligned with the full decode.

Unit tests: `test_code2wav.py`, `test_code2wav_batching.py`, `test_code2wav_cuda_graph.py` all pass.

### DIFF
--- a/sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
+++ b/sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
@@ -213,9 +213,7 @@ class Code2WavScheduler(StreamingVocoderBase[Code2WavStreamState, "list[int]"]):
             codes,
             graph_eligible=not is_final,
         )
-        trim = context * self._total_upsample
-        if trim:
-            wav = wav[..., trim:]
+        wav = wav[..., -(end - start) * self._total_upsample :]
         audio = wav.reshape(-1).detach().cpu().float().numpy().copy()
         if profile_metadata is not None:
             _emit_event(
@@ -475,9 +473,7 @@ class Code2WavScheduler(StreamingVocoderBase[Code2WavStreamState, "list[int]"]):
                     f"{len(group)} requests"
                 )
             context = min(self._left_context_size, group[0][1].emitted)
-            trim = context * self._total_upsample
-            if trim:
-                wav = wav[..., trim:]
+            wav = wav[..., -(window_frames - context) * self._total_upsample :]
             host = wav.detach().cpu().float()
             for i, (rid, state) in enumerate(group):
                 audio = host[i].reshape(-1).numpy().copy()
--- a/tests/unit_test/fixtures/qwen_fakes.py
+++ b/tests/unit_test/fixtures/qwen_fakes.py
@@ -154,12 +154,13 @@ class FakeAudioEncoderModel:
 
 
 class FakeCode2WavModel:
-    def __init__(self, *, total_upsample: int = 2) -> None:
+    def __init__(self, *, total_upsample: int = 2, output_deficit: int = 0) -> None:
         self.total_upsample = total_upsample
+        self.output_deficit = output_deficit
         self.calls: list[tuple[int, ...]] = []
 
     def __call__(self, codes: torch.Tensor) -> torch.Tensor:
         self.calls.append(tuple(codes.shape))
-        samples = int(codes.shape[-1]) * self.total_upsample
+        samples = int(codes.shape[-1]) * self.total_upsample - self.output_deficit
         base = codes.to(dtype=torch.float32).flatten(1).sum(dim=1).view(-1, 1, 1)
         return torch.arange(samples, dtype=torch.float32).view(1, 1, samples) + base
--- a/tests/unit_test/qwen3_omni/test_code2wav.py
+++ b/tests/unit_test/qwen3_omni/test_code2wav.py
@@ -759,3 +759,21 @@ def test_eos_chunk_is_skipped_and_never_decoded() -> None:
     audio = np.frombuffer(message.data.data["audio_waveform"], dtype=np.float32)
     assert model.calls == [(1, 2, 2)]
     assert audio.shape == (4,)
+
+
+def test_qwen_code2wav_emits_full_chunk_despite_model_output_deficit() -> None:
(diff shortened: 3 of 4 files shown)


SandyLuXY (approved):
lgtm

PR #1364, opened 2026-08-06
[higgs-tts] add breakable prefill CUDA graph (BCG) support

Ratish1:
## Summary

Adds breakable prefill CUDA graph (BCG) support against pinned sglang 0.5.16. The enablement layer is model-agnostic; models must adopt the payload contract before a deployment can turn it on, and non-adopted models refuse it at startup. Higgs TTS is the pilot adopter and ships with it **enabled by default** (generated bucket ladder at the qualified 512-token capture budget; longer prefills run eager). Other TTS/ASR models adopt in follow-ups by implementing the same contract and flipping their builder and capability flags.

A deployment disables it with one override:

```python
server_args_overrides = {"cuda_graph_backend_prefill": "disabled"}
```

Changes:

- **Payload channel**: runner-composed prefill embeddings and request ids cross the graph boundary as a frozen `OmniPrefillInputs` carried on `ForwardBatch.mm_inputs`, which sglang's prefill graph runner copies verbatim into its static batch (the static batch does not preserve `rids`). `forward_batch.input_embeds` stays `None` so the graph gate admits the batch. Attaching replaces the per-request `[None, ...]` placeholder that `prepare_for_extend` puts on every real batch, and refuses genuine multimodal inputs or a double attach.
- **Static slot registration**: a new worker config flag marks the model config multimodal after ServerArgs has already run its multimodal compatibility gates, which is the condition under which sglang registers the static `input_embeds` buffer slot the payload lands in at replay. The ordering invariant is documented at the flip site.
- **Bucket policy**: breakable prefill requires explicitly declared `cuda_graph_bs_prefill` buckets, matching the existing decode policy (`cuda_graph_bs` must be explicit). `build_default_prefill_cuda_graph_bs()` generates sglang's own ladder formula for a declared token budget, so deployments opt in with one number instead of a hand-written list. Buckets are validated against `chunked_prefill_size` and `max_prefill_tokens`, and gaps beyond the 2x padding factor warn about eager valleys.
- **Safety rails**: enabling breakable requires the engine builder to declare adoption of the contract, so a deployment override cannot turn it on for a model that never implemented it; a post-capture attestation verifies the constructed runner type, backend, captured bucket shapes, and the embeds slot, failing startup loudly on any of sglang's silent-downgrade paths.
- **Higgs adoption**: the decoder is exposed for sglang's prefill-graph discovery via a property (no duplicate module alias in the parameter tree), prefill conditioning is attached as the payload in the prefill hook, and the model forward consumes the payload as its single prefill input source. Dead request-id writes from the previous bridge are removed, and the old hard-disabled guard is replaced by the opt-in contract. The remaining eight model packages explicitly declare non-adoption, and the capability is included in the startup capabilities log.

Numerics: bucket-padded replay is not bit-identical to eager prefill. The divergence is fully localized to the dense GEMMs, whose cuBLAS kernel selection varies with the padded token count; every other op in the layer (norms, rope, activations, attention, the eager head) is invariant to padding. It is certified benign two ways: with batch-invariant kernels the identical padded replays become bit-identical to eager (30/30 WAV SHA match), and quality is statistically unchanged on the full Seed-TTS eval (below). Sampled outputs differ the same way they already do when batch composition changes.

## Testing

Unit: full suite 2,893 passed, 1 skipped, 0 failed (fleet is a no-op with BCG disabled, the default).

Correctness (H100, seed-paired 30-utterance oracle at c1 unless noted):

| Check | Result |
|---|---|
| Dense exact-shape BCG vs eager | 30/30 WAV + 30/30 action SHA match |
| Same request x20 (one bucket) | 1 unique WAV / action stream |
| Batch-invariant kernels, padded sparse replay vs eager | 30/30 WAV SHA match |
| Rollback (BCG disabled) vs original manifest | 30/30 match |
| Abort lifecycle (100 randomized disconnects, 58 pre-header / 42 mid-stream) | 100/100 confirmed aborts, liveness 10/10, post-abort oracle 30/30 |
| Shape-edge sweep (92 exact-token cases incl. 3-token minimum, 513/514/600 over cap) | 92/92 expected disposition; over-cap fell back to eager correctly |

Quality (full Seed-TTS-EN, 1,088 utterances, seed-paired, c16, radix on):

| Metric | Eager | BCG (ladder-512) | Delta |
|---|---|---|---|
| Corpus WER | 0.913% | 0.904% | ΔWER 95% CI [−0.084, +0.047] pp |
| Speaker SIM | 67.20 | 67.34 | ΔSIM 95% CI [−0.028, +0.311] |

Performance, non-streaming radix-on c16 (same 1,088-utterance run):

| Metric | Eager | BCG | Delta |
|---|---|---|---|
| Throughput | 16.030 req/s | 16.672 req/s | +4.00% |
| Mean latency | 0.995 s | 0.940 s | −5.53% |
| p99 latency | 1.651 s | 1.578 s | −4.42% |

Performance, streaming radix-off (48 utterances per arm):

| Concurrency | TTFA median | TTFA p99 | Throughput |
|---|---|---|---|
| c8 | −5.53% | −21.75% | +1.67% |
| c16 | −11.79% | −2.34% | +6.96% |

c1 streaming (3 interleaved warm A/B pairs, fresh server per arm): BCG better in all pairs — TTFA median −4.76% paired, TTFA p99 −3.92% / −7.26% in the stable pairs; throughput flat. Cold-server c1 tails affect eager and BCG alike and are removed by a workload warmup pass at boot (documented as deployment bootstrap practice; a single default warmup request is not sufficient at c1).

Capture budget (30-shape generated ladder, cap 512): 1.8–7 s, 0.30 GB.


sglang_omni/models/model_capabilities.py
```diff
@@ -28,13 +28,18 @@ class ModelCapabilities:
...
+      payload, eager-tail-safe head); enabling it per deployment stays with
+      the engine builder policy.
     """
 
     supports_reference_audio: bool
     supports_batch_vocoder: bool
     supports_streaming_vocoder: bool
     supports_cuda_graph: bool
     supports_torch_compile: bool
+    supports_breakable_prefill_cuda_graph: bool
```
stephenkgli, line 42: Maybe we can use a str field like `prefill_cuda_graph_backend`, `disabled` means PCG isn't supported, `breakable` means breakable PCG is supported, so that different models can use different kinds of backend.

stephenkgli (commented):
Have you ever tried different backends for higgs-tts, and compare their performance and correctness?

You can also benchmark using high concurrency like 48 or 96, we may get higher throughput.

sglang_omni/models/higgs_tts/engine_builder.py
```diff
@@ -79,6 +84,9 @@ def generation_defaults(
                 else 0.85
             ),
             "chunked_prefill_size": 8192,
+            # Qualified capture budget; longer prefills run eager.
+            "cuda_graph_backend_prefill": CudaGraphBackend.BREAKABLE,
```
AkazaAkane, line 88: This default is passed to ServerArgs through the same explicit cuda_graph_backend_prefill field used for deployment overrides. In upstream SGLang, that locks ("prefill", "backend"), which means the <4 GiB auto-prefill-CG memory safety fallback no longer applies.

So enabling BCG by default can turn a deployment that previously fell back to eager prefill into a startup/capture failure, even though the operator never explicitly requested BCG.

sglang_omni/scheduling/bootstrap.py
```diff
@@ -91,6 +93,12 @@ def create_sglang_infrastructure(
     model_runner.alloc_memory_pool()
     model_runner.init_attention_backends()
 
+    if model_worker.enable_prefill_input_embeds:
+        # The attention backends above cache is_multimodal at construction
+        # and must keep the model's real semantics; only the prefill graph
+        # runner reads the flag from here, to register its input_embeds slot.
+        model_worker.model_config.is_multimodal = True
```
AkazaAkane: The intent is only to make PrefillCudaGraphRunner.__init__ register the static input_embeds slot, but this mutates a semantic model property globally. The comment says only the prefill graph runner reads it from here, which happens to be true on the current Omni path, but upstream has other live readers of model_config.is_multimodal, so this creates a fragile dependency on initialization order and on which upstream methods Omni reuses.

Since PrefillCudaGraphRunner snapshots is_multimodal and builds its registry during init_cuda_graphs(), could we restore the real value immediately after capture, or otherwise localize this override to graph construction?

sglang_omni/models/model_capabilities.py
edwingao28, line 42: This new field is required, but DotsTTS was missed. Both `sglang_omni/models/dots_tts/__init__.py` and the Dots entry in `test_model_capabilities.py` still create `ModelCapabilities` without it. That makes the capability test fail during collection, and importing the Dots package raises a `TypeError`.

Could you add `supports_breakable_prefill_cuda_graph=False` in both places and rerun the tests?

sglang_omni/scheduling/generation_batch_policy.py
```diff
@@ -163,6 +217,84 @@ def validate_generation_batch_policy(
...
+    if ("prefill", "bs") not in server_args._cuda_graph_config_locked:
+        errors.append(
+            "breakable prefill CUDA graphs require explicit "
+            "cuda_graph_bs_prefill buckets (sglang's generated ladder is "
+            "not an accepted shape policy)"
+        )
+        return
+
+    prefill_cfg = server_args.cuda_graph_config.prefill
+    buckets = _normalize_cuda_graph_bs(
```
edwingao28, line 254: We convert the bucket values to `int` for validation, but never store the converted values back. So inputs like `[4.5, 8.5]` or `["4", "8"]` pass this check, while SGLang later receives the original values and tries to use them as tensor sizes, which fails at startup.

Could we either reject anything that is not an actual integer (including `bool`) or replace the config values with the normalized integers?

tests/unit_test/higgs_tts/test_pipeline.py
```diff
@@ -396,6 +441,35 @@ def fake_make_adapters(**kwargs):
...
+    builder = HiggsTtsEngineBuilder(
+        max_new_tokens=2048,
+        max_running_requests=64,
+        cuda_graph_max_bs=64,
+        enable_async_decode=False,
+        async_decode_min_batch_size=2,
+    )
+    builder.build(
+        "bosonai/higgs-tts-3-4b",
+        server_args_overrides={"disable_cuda_graph": True},
```
edwingao28: This test turns off all CUDA graphs, including decode graphs. It therefore does not test the rollback described in the PR, which only disables prefill graphs with `cuda_graph_backend_prefill="disabled"`.

Could we use that override here and check that decode graph initialization still runs, while the prefill embedding wiring and attestation stay off?

edwingao28 (commented):
I found three concrete issues that are worth addressing before this lands.

sglang_omni/models/higgs_tts/engine_builder.py
Ratish1: done

sglang_omni/scheduling/bootstrap.py
Ratish1: done

sglang_omni/models/model_capabilities.py
Ratish1: done

tests/unit_test/higgs_tts/test_pipeline.py
Ratish1: done

sglang_omni/scheduling/generation_batch_policy.py
Ratish1: done

AkazaAkane:
please address the failed unit tests

PR #1366, opened 2026-08-06
fix: remove duplicate repetition penalty application

### LINKED ISSUE #1360: [Bug] repetition_penalty is applied twice (effective penalty is squared)
### Checklist

### Describe the bug

`ModelRunner._sample_next_token_ids` applies `repetition_penalty` manually (`sglang_omni/model_runner/base.py`):

```python
def _sample_next_token_ids(self, logits_output, forward_batch, schedule_batch, requests):
    self._apply_repetition_penalty(logits_output, requests)      # <-- also done by SGLang
    self._apply_codec_suppress_tokens(logits_output, requests)
    self._install_sampling_seeds(forward_batch, requests)
    ...
    next_token_ids = self.tp_worker.model_runner.sample(logits_output, forward_batch)
```

SGLang already applies the same rule through `BatchedRepetitionPenalizer`, reached from `SamplingBatchInfo.apply_logits_bias` inside `tp_worker.model_runner.sample()`. Both run on the same logits, so the configured penalty is applied twice: **`repetition_penalty=2.0` behaves as 4.0.**

Measured on a single logit of an already-generated token:

```
with    the base.py call:  +6.5237 -> +1.6309   (4.00x)
without the base.py call:  +6.5237 -> +3.2619   (2.00x)
```

Both stages read the same `req.sampling_params.repetition_penalty` and act on the same token set, and `_sample_next_token_ids` always routes through `apply_logits_bias`, so there is no configuration in which only one of them runs. Confirmed at runtime: `penalizer_orchestrator.is_required = True`, and the number of entries SGLang modifies grows by one per decode step, matching `len(req.output_ids)`.

Separately, `_apply_repetition_penalty` rebuilds the full `(row, token, penalty)` list every decode step (`O(batch × output_len)`) and materialises it with `torch.tensor(python_list, device="cuda")` three times — a blocking pageable H2D copy on the scheduler thread. In our setup it accounted for ~59% of scheduler-thread CPU time and starved the GPU (SM utilisation averaged 73%, never reaching 90%); removing the call raised end-to-end throughput 8–13% across concurrency 1–64. That is a side effect, not the reason to fix it.

### Reproduction

Snapshot the logits on entry to `_sample_next_token_ids` (before either stage) and compare against the value after SGLang's `apply_logits_bias`. The snapshot must be taken outside `_apply_repetition_penalty` so the script still reports when that call is removed.

```python
import torch
from sglang.srt.sampling.sampling_batch_info import SamplingBatchInfo
from sglang_omni.model_runner.base import ModelRunner

snap, n = {}, [0]

_sample = ModelRunner._sample_next_token_ids
def traced_sample(self, logits_output, forward_batch, schedule_batch, requests):
    snap["raw"] = logits_output.next_token_logits.clone()      # before any shaping
    return _sample(self, logits_output, forward_batch, schedule_batch, requests)
ModelRunner._sample_next_token_ids = traced_sample

_bias = SamplingBatchInfo.apply_logits_bias
def traced_bias(self, logits):
    _bias(self, logits)                                        # after SGLang's pass
    if "raw" in snap and n[0] < 4:
        changed = (snap["raw"] != logits)[0].nonzero().flatten()
        if len(changed):
            t = int(changed[0])
            raw, final = snap["raw"][0, t].item(), logits[0, t].item()
            print(f"token {t}: raw {raw:+.4f} -> final {final:+.4f} = {raw / final:.2f}x")
            n[0] += 1
SamplingBatchInfo.apply_logits_bias = traced_bias
```

Import this before the engine starts, then run any decode with `repetition_penalty=2.0`. Greedy sampling (`temperature=0`) keeps the trace stable, and the ratio is what matters — the specific token ids and logit values vary with the model and sampling settings.

Observed:

```
token 33: raw +6.5237 -> final +1.6309 = 4.00x     # expected 2.00x
token 10: raw +5.8992 -> final +1.4748 = 4.00x
```

After deleting the `self._apply_repetition_penalty(...)` line, the same script prints:

```
token 33: raw +6.5237 -> final +3.2619 = 2.00x
token 10: raw +5.8992 -> final +2.9496 = 2.00x
```

### Environment

- SGLang 0.5.12.post1
- PyTorch 2.11.0+cu130, CUDA 13.0, Python 3.12
- NVIDIA L40S (46 GB), single GPU, TP 1
- overlap schedule disabled, FlashInfer attention backend, async decode enabled

albaNnaksqr:
ModelRunner._sample_next_token_ids applies repetition penalty before calling `tp_worker.model_runner.sample()`. The upstream SGLang sampler already applies repetition penalty through `BatchedRepetitionPenalizer`, so the same penalty
was applied twice to the same logits.

This change removes the redundant Omni-side repetition penalty and relies on the upstream implementation as the single source of truth for the standard sampling path.

Not affected:

- `models/zonos2/sampler.py` uses its own sampling path.
- `models/qwen3_tts` delegates to the base implementation and inherits this fix.

Closes https://github.com/sgl-project/sglang-omni/issues/1360

## Checklist

- [ ] Add unit tests.

## CI


Ratish1:
Thanks @albaNnaksqr for identifying and addressing the duplicate repetition penalty. I added you as a co-author on #1750, which supersedes this PR and includes the Qwen3-TTS-specific codec suppression and retraction handling. Thanks again for your contribution.

Issue #1367, opened 2026-08-06
[Roadmap] dots.tts Model Support and Performance Optimization


Hayden727:
## Goal

Continue improving dots.tts model coverage, single-request latency, concurrent throughput, time to first chunk, and memory efficiency without reducing speech quality.

This roadmap builds on the framework-native implementation merged in #1349. SGLang-Omni remains the serving runtime; this roadmap does not introduce a second runtime or vendor the upstream model implementation.

Tensor parallelism is intentionally out of scope.

## Current baseline

The following support has landed on current `main` (`2e607bc0`, 2026-08-10):

- [x] Framework-native `preprocessing -> reference_encode -> latent_engine -> vocoder` pipeline
- [x] Shared `TtsEngineBuilder` integration
- [x] Qwen2 backbone on the SGLang request lifecycle and paged KV cache
- [x] Streaming and non-streaming 48 kHz AudioVAE output
- [x] Stable streaming with backbone decode CUDA Graph enabled (#1395)
- [x] Real length-bucketed non-streaming AudioVAE batching (#1374)
- [x] Content-keyed reference-audio cache and same-key in-flight request merging
- [x] Deterministic leading-window long-reference speaker crops and parity-safe exact-length batched reference encode (#1420)
- [x] MeanFlow + Euler continuous batching with `max_running_requests=16` by default
- [x] SGLang backbone decode CUDA Graph with a model-owned feedback buffer (#1377)
- [x] Batched MeanFlow and semantic-encoder CUDA Graphs with eager fallback (#1385)
- [x] Full-slot DiT KV direct-view path with partial/fragmented gather fallback (#1389)
- [x] Batched EOS flags staged asynchronously and finish handling deferred to resolve (#1402)
- [x] Validated latent-engine and vocoder optimizations enabled by default (#1376)
- [x] Slot-indexed DiT and semantic-encoder KV/conv state pools
- [x] Acoustic-tail pool memory accounting, startup VRAM validation, and explicit slot-admission errors (#1408)
- [x] Fused QKV/AdaLN inference projections and per-slot RNG
- [x] Cleanup across finish, abort, retraction, failed prefill, and re-prefill
- [x] Compiled single-request semantic encoder, recurrent DiT step, and vocoder path
- [x] Full Seed-TTS-Eval EN end-to-end validation
- [x] SOAR and base checkpoint serving on the single-request CFG solver (#1391, #1393)
[...]

PR #1370, opened 2026-08-06
[dots.tts Perf] Skip the full vocabulary LM head on AR steps

db-ol:
## Motivation

First item of the #1367 roadmap. dots.tts consumes Qwen2 hidden states only. The next step input comes from the MeanFlow latent patch and semantic encoder, EOS is decided by the flow head eos projection, and DotsTTSModelRunner overwrites next_token_ids with the control token before the sampler fallback can fire. The wrapper still delegates to Qwen2ForCausalLM.forward, so every prefill and every AR decode step runs the full vocabulary logits projection (151672 x hidden per the shipped llm_config) plus the logits processor, and the result is discarded.

## Modifications

- DotsTTSSGLangModel.forward now calls the bare Qwen2 transformer (self.qwen2.model) and returns a LogitsProcessorOutput carrying the hidden states directly, skipping lm_head and the logits processor. Prefill returns all extend rows and decode returns one row per request, matching what CaptureHiddenMode FULL and LAST delivered before, so post_prefill row accounting and initialize_history see identical tensors. A dummy next_token_logits of shape (requests, 1) satisfies the result contract, following the same pattern already used by MingTTS and MOSS TTS, and HiggsTTS bypasses its backbone head the same way.
- Weight loading is unchanged. The lm_head weights still load but are never executed. The released checkpoint ships tie_word_embeddings true, so the head shares storage with the embedding and there is no separate weight memory to reclaim, the win is per step compute and kernel count.
- Two new unit tests: the fake backbone raises if anything enters the LM head path, and the tests pin full extend rows at prefill, per request rows at decode, and the dummy logits shape.

## Related Issues

First optimization item of #1367.

## Accuracy Test

Bit exact parity: all 1088 Seed-TTS-Eval EN wav files generated at seed 42 are byte identical between main and this PR (md5 per file), so WER and speaker similarity are unchanged by construction. tests/unit_test/dots_tts passes (60 tests) in a local container, the 5 failures present there fail identically on unmodified main in the same container (stale local sglang install), CPU CI carries the authoritative run.

## Benchmark & Profiling

Same card A/B per the #1367 contract posted in the comment below: speed neutral within host noise, confirmed by an A-B-A drift control where a main replay measures behind this PR, plus the bit exact audio parity above.

## Checklist

- [x] Add unit tests.

### DIFF
--- a/sglang_omni/models/dots_tts/sglang_model.py
+++ b/sglang_omni/models/dots_tts/sglang_model.py
@@ -8,6 +8,7 @@ from pathlib import Path
 from typing import Any
 
 import torch
+from sglang.srt.layers.logits_processor import LogitsProcessorOutput
 from sglang.srt.models.qwen2 import Qwen2ForCausalLM
 from torch import nn
 
@@ -82,15 +83,28 @@ class DotsTTSSGLangModel(nn.Module):
         forward_batch: Any,
         input_embeds: torch.Tensor | None = None,
         **kwargs: Any,
-    ):
+    ) -> LogitsProcessorOutput:
+        del kwargs
         if input_embeds is None:
             input_embeds = forward_batch.input_embeds
-        return self.qwen2(
+        hidden_states = self.qwen2.model(
             input_ids=input_ids,
             positions=positions,
             forward_batch=forward_batch,
             input_embeds=input_embeds,
-            **kwargs,
+        )
+        # note (db-ol): dots consumes hidden states only and the runner
+        # overwrites next_token_ids, dummy logits just satisfy the contract.
+        if forward_batch.forward_mode.is_extend():
+            extend_seq_lens = getattr(forward_batch, "extend_seq_lens", None)
+            request_count = (
+                int(extend_seq_lens.numel()) if extend_seq_lens is not None else 1
+            )
+        else:
+            request_count = int(hidden_states.shape[0])
+        return LogitsProcessorOutput(
+            next_token_logits=hidden_states.new_empty((request_count, 1)),
+            hidden_states=hidden_states,
         )
 
 
--- a/tests/unit_test/dots_tts/test_sglang_model.py
+++ b/tests/unit_test/dots_tts/test_sglang_model.py
@@ -1,5 +1,8 @@
+from types import SimpleNamespace
+
 import pytest
 import torch
+from sglang.srt.layers.logits_processor import LogitsProcessorOutput
 from torch import nn
 
 from sglang_omni.models.dots_tts.sglang_model import DotsTTSSGLangModel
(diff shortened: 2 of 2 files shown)


db-ol:
Same card A/B results per the #1367 contract. 1x H100, exclusive GPU (verified idle across a 150 s window before each sweep), Seed-TTS-Eval EN full 1088 samples, seed 42, generate only mode, concurrency 1/4/8/16. main is 0514ef6, PR is this branch on the same base, main replay is main run a third time after the PR sweep as a drift control.

| c | metric | main | PR | main replay |
|---|--------|------|----|-------------|
| 1 | qps | 0.833 | 0.797 | 0.770 |
| 1 | latency mean s | 1.200 | 1.254 | 1.298 |
| 1 | latency p95 s | 1.751 | 1.836 | 1.893 |
| 1 | rtf mean | 0.2908 | 0.3038 | 0.3141 |
| 4 | qps | 2.204 | 2.148 | |
| 4 | latency mean s | 1.813 | 1.861 | |
| 4 | rtf mean | 0.4388 | 0.4512 | |
| 8 | qps | 2.242 | 3.221 | |
| 8 | latency mean s | 3.565 | 2.480 | |
| 8 | rtf mean | 0.8530 | 0.6013 | |
| 16 | qps | 4.331 | 4.287 | 4.202 |
| 16 | latency mean s | 3.673 | 3.711 | 3.784 |
| 16 | latency p95 s | 5.419 | 5.450 | 5.694 |
| 16 | rtf mean | 0.8876 | 0.8977 | 0.9151 |

Two conclusions.

1. Bit exact parity. All 1088 wav files generated at c=1 are byte identical between main and this PR (md5 compared per file). WER and speaker similarity are therefore unchanged by construction, the transcription phase would score identical audio.

2. Speed is neutral within host noise. The PR measures 1 to 4.5 percent behind the first main sweep, but the main replay measures 3 to 8 percent behind its own first run, with a monotonic slowdown across the three sweeps in time order (main, then PR, then main replay). Same code drift exceeds the main to PR delta at both replayed levels, so the difference is host drift, not this change. That matches expectation, the removed projection costs well under a millisecond per AR step while the step time is dominated by the MeanFlow DiT and semantic encoder.

Two notes. The main c=8 run hit a stall (its p95 of 9.8 s exceeds its own c=16 p95 of 5.4 s), so c=8 is not a usable comparison point in either direction. And the released checkpoint ships tie_word_embeddings true, the lm_head shares storage with the embedding, so there is no separate weight memory to reclaim and the win of this PR is per step compute and kernel count only. I have removed the memory follow-up note from the description accordingly.

luojiaxuan (approved):

PR #1349, opened 2026-08-05
[Model] Add framework-native dots.tts support

Hayden727:
## Motivation

Add [dots.tts](https://github.com/studio-dots-ai/dots.tts) after the TTS pipeline refactor, with the model adapted to SGLang-Omni instead of porting the upstream serving runtime into this repository.

This PR provides a correct, deployable TP1 MeanFlow implementation through the OpenAI-compatible `/v1/audio/speech` API, including reference-audio voice cloning, streaming output, and continuous batching.

## Architecture

SGLang-Omni owns the serving runtime:

- `preprocessing → reference_encode → latent_engine → vocoder` uses the framework's pipeline stages and schedulers.
- The latent engine is built through the shared `TtsEngineBuilder` path, including generation-policy validation and abort/finish lifecycle hooks.
- The checkpoint is exposed through an in-memory Hugging Face composite config with a nested `llm_config`; no temporary checkpoint view or rewritten config is created on disk.
- The Qwen2 backbone runs through the standard SGLang path, including scheduling, prefill/decode, paged KV cache, hidden-state capture, abort, and finish handling.
- A model-local acoustic-tail adapter batches the dots MeanFlow DiT and semantic-encoder recurrence with slot-indexed KV/conv pools. It reuses the upstream model operators already provided by `dots.tts==0.2.1`; no backbone, vocoder, speaker encoder, or serving runtime is vendored.
- Reference encoding reuses `ReferenceEncodeService`; waveform output reuses `BatchVocoderBase` and `StreamingVocoderBase`.

PR size after the framework-native refactor and continuous-batching support: **36 files, +3,764 / -9** relative to current `main`.

## Modifications

- Add framework-native preprocessing and reference-conditioning stages.
- Register an in-memory dots Hugging Face config and an SGLang Qwen2 model with dots semantic/flow heads and checkpoint mapping.
- Feed generated latent embeddings back through SGLang decode and use the resulting hidden state for MeanFlow conditioning and EOS.
- Support streaming and non-streaming AudioVAE decode through shared vocoder schedulers.
- Reuse shared audio loading while matching upstream prompt trim and high-quality resampling.
- Add continuous batching for the MeanFlow + Euler path with fixed engine-wide NFE, batched DiT/semantic-encoder KV caches, fused QKV/AdaLN inference projections, and per-slot RNG.
- Reclaim acoustic slots on normal finish, abort/reset, prefill rollback, and re-prefill. Release is idempotent and clears the request's slot handle before the slot can be reused.
[...]

PR #1372, opened 2026-08-06
[Perf] dots.tts single-request path: async EOS readback, effective-backend logging, benchmark --server-config

luojiaxuan:
## Motivation

Low-hanging items from the dots.tts roadmap #1367 focused on the `max_running_requests=1` (single-request) path.

Two gaps made c=1 numbers look much worse than the merged implementation actually is:

1. `benchmarks.eval.benchmark_tts_seedtts` starts its managed server with `--model-path` only, so the canonical optimized deployment (`examples/configs/dots_tts.yaml`, `optimize: true`) is unreachable from the benchmark: c=1 runs measured the eager DiT/semantic-encoder path with per-patch eager AudioVAE decode (`merge_steps` is forced to 1 when the vocoder is not optimized).
2. The single-request `decode_next` reads the EOS flag with a blocking `.item()` **before** any DiT/semantic-encoder work is queued, so every step pays a host-sync bubble between the backbone forward and the acoustic tail. The batched path already avoids this ("Single readback after the tail work is queued so the sync overlaps it").

## Modifications

- `flow_head.decode_next`: keep the EOS flag on the GPU, stage it through a cached pinned buffer + CUDA event, and read it back only after the DiT + patch-encoder launches are queued. The copy overlaps the tail kernels; numerics are unchanged (same tensor, same threshold, later sync). CPU fallback keeps the direct `.item()` for the unit-test path.
- `engine_builder.setup_model` / `stages.create_vocoder_executor`: log the effective latent-engine and vocoder backend (roadmap: "Log the effective backend selected for every component").
- `benchmark_tts_seedtts` + `managed_omni_server`: new `--server-config` flag forwarded to `sglang_omni.cli serve` as `--config`, so benchmark runs can start the canonical optimized deployment directly. The value is recorded in `speed_results.json`.

## Benchmark

Single H100 (hyper01), Seed-TTS-Eval EN, `--ref-format references`, c=1, `max_running_requests=1`, seed 42, warmup 8, first 64 samples, same card class A/B (eager on one idle GPU, compiled on another; no co-tenant processes on the measured GPUs).

| Server | Latency mean (s) | RTF mean | RTF p50 | RTF p95 |
|---|---:|---:|---:|---:|
| eager (current benchmark default, no config) | 1.882 | 0.4811 | 0.4646 | 0.5711 |
| compiled (`--server-config examples/configs/dots_tts.yaml`, unpatched main) | 1.082 | 0.2796 | 0.2623 | 0.3592 |
| compiled + this PR | **1.051** | **0.2717** | **0.2548** | **0.3509** |

A full-set (1,088-sample) c=1 run of the eager default and of the compiled+patched path is in flight on the same card; final full-set speed + WER numbers will follow as a PR comment.

Repro command after this PR:

```bash
  --model dots-studio/dots.tts-mf \
  --server-config examples/configs/dots_tts.yaml \
  --ref-format references \
  --host 127.0.0.1 --port 30190 --lang en \
  --max-concurrency 1 --max-running-requests 1 \
  --warmup 8 --seed 42 \
  --output-dir results/dots-seedtts-en-c1-optimized
```

## Accuracy

Same 64-sample slice, same seed, Qwen3-ASR-1.7B (`--transcribe-only`):

| Server | Corpus WER | >50% WER samples |
|---|---:|---:|
| eager | 1.28% | 0 |
| compiled + this PR | 1.14% | 0 |

The EOS change is numerics-preserving (same tensor, same threshold, later readback); the compiled path itself is the already-merged single-request optimization from #1349, previously unreachable from the managed benchmark server.

## Validation

- `python -m pytest -q tests/unit_test/dots_tts` — 39 passed
- `black --check` on changed files — clean

## Related

- Roadmap #1367 (single-request items); model support #1349.

## Checklist

### DIFF
--- a/benchmarks/eval/benchmark_tts_seedtts.py
+++ b/benchmarks/eval/benchmark_tts_seedtts.py
@@ -253,6 +253,10 @@ class TtsSeedttsBenchmarkConfig:
     disable_tqdm: bool = False
     max_running_requests: int = 64
     cuda_graph_max_bs: int = 64
+    # Optional sglang-omni pipeline config yaml forwarded to the managed TTS
+    # server as ``--config`` (e.g. examples/configs/dots_tts.yaml to run the
+    # canonical optimized deployment).
+    server_config: str | None = None
     # Transcribe phase
     lang: str = "en"
     device: str = "cuda:0"
@@ -307,6 +311,7 @@ def _build_results_config(
         "initial_codec_chunk_frames": config.initial_codec_chunk_frames,
         "max_running_requests": config.max_running_requests,
         "cuda_graph_max_bs": config.cuda_graph_max_bs,
+        "server_config": config.server_config,
     }
 
 
@@ -449,6 +454,7 @@ def _config_from_args(args: argparse.Namespace) -> TtsSeedttsBenchmarkConfig:
         disable_tqdm=args.disable_tqdm,
         max_running_requests=args.max_running_requests,
         cuda_graph_max_bs=args.cuda_graph_max_bs,
+        server_config=args.server_config,
         lang=args.lang,
         device=args.device,
         similarity_checkpoint=args.similarity_checkpoint,
@@ -675,6 +681,17 @@ def _build_arg_parser() -> argparse.ArgumentParser:
         ),
     )
     parser.add_argument(
+        "--server-config",
+        type=str,
+        default=None,
+        help=(
+            "Optional sglang-omni pipeline config yaml passed to the managed "
+            "TTS server as --config (e.g. examples/configs/dots_tts.yaml for "
+            "the canonical optimized dots.tts deployment). Ignored with "
+            "--use-existing-server."
+        ),
+    )
+    parser.add_argument(
         "--skip-gpu-cleanup",
         action="store_true",
         help=(
@@ -769,6 +786,7 @@ def main() -> None:
             model_path=config.model,
             port=config.port,
             host=config.host,
+            server_config=config.server_config,
             max_running_requests=config.max_running_requests,
             cuda_graph_max_bs=config.cuda_graph_max_bs,
             log_file=Path(config.output_dir) / "server_logs" / "tts_server.log",
--- a/benchmarks/eval/benchmark_tts_seedtts.py
+++ b/benchmarks/eval/benchmark_tts_seedtts.py
@@ -253,9 +253,10 @@ class TtsSeedttsBenchmarkConfig:
     disable_tqdm: bool = False
     max_running_requests: int = 64
     cuda_graph_max_bs: int = 64
-    # Optional sglang-omni pipeline config yaml forwarded to the managed TTS
-    # server as ``--config`` (e.g. examples/configs/dots_tts.yaml to run the
-    # canonical optimized deployment).
+    # note (luojiaxuan): optional sglang-omni pipeline config yaml forwarded
+    # to the managed TTS server as ``--config`` (e.g.
+    # examples/configs/dots_tts.yaml to run the canonical optimized
+    # deployment).
     server_config: str | None = None
     # Transcribe phase
     lang: str = "en"
--- a/benchmarks/benchmarker/utils.py
+++ b/benchmarks/benchmarker/utils.py
@@ -249,6 +249,7 @@ def managed_omni_server(
     port: int,
     host: str,
     log_file: Path | None,
+    server_config: str | None = None,
     max_running_requests: int | None = None,
     cuda_graph_max_bs: int | None = None,
     mem_fraction_static: float | None = None,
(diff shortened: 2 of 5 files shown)


luojiaxuan:
Full-set results are in. Single H100 (same card class as the PR body's n=64 A/B, same host), Seed-TTS-Eval EN **1,088/1,088 samples, 0 failed**, c=1, `max_running_requests=1`, seed 42, non-streaming, Qwen3-ASR-1.7B for WER:

| Full-set c=1 | Latency mean / p50 / p95 / p99 (s) | RTF mean / p50 / p95 / p99 | Corpus WER |
|---|---|---|---:|
| eager (current benchmark default) | 1.865 / 1.818 / 2.720 / 3.225 | 0.4508 / 0.4462 / 0.4932 / 0.5385 | 1.25% |
| compiled + this PR (`--server-config examples/configs/dots_tts.yaml`) | **1.010 / 0.990 / 1.453 / 1.707** | **0.2459 / 0.2420 / 0.2785 / 0.3080** | 1.31% |

- **RTF mean 0.451 → 0.246 (−45%)**; even p95 stays below 0.28.
- WER 1.31% equals the #1349 c=16 reference (1.3146%); Δ vs the same-machine eager full run is +0.06 pp, inside the roadmap's 0.10 pp budget. 0 samples above 50% WER in both runs, max per-sample 33.3% in both.
- Caveat on the eager row: its middle third overlapped unrelated CPU load on the host (compile jobs of the A/B servers), which inflated it somewhat — the uncontended eager window ran at ~1.5–1.6 s/request (RTF ≈ 0.36–0.39). The compiled row ran on an otherwise idle host.
- Fixed-vs-marginal split on the compiled path: latency ≈ 0.44 s per request + ≈ 0.155 s per audio-second (≈ 25 ms per latent patch). The fixed part is dominated by cold reference encoding + prefill/first patch; the marginal part by the eager, non-CUDA-graph backbone decode step — consistent with the two remaining roadmap levers (#1367).

Raw outputs (host `hyper01`): `/data04/jaxan/dots-seedtts-en-full-c1` (eager) and `/data04/jaxan/dots-bench-exp/full-opt-eos-c1` (compiled + patch), each with `speed_results.json`, `wer_results.json`, and generated audio.

benchmarks/eval/benchmark_tts_seedtts.py
```diff
@@ -253,6 +253,10 @@ class TtsSeedttsBenchmarkConfig:
     disable_tqdm: bool = False
     max_running_requests: int = 64
     cuda_graph_max_bs: int = 64
+    # Optional sglang-omni pipeline config yaml forwarded to the managed TTS
```
zhaochenyang20: add `note (jiaxuan)`

zhaochenyang20 (approved):

benchmarks/eval/benchmark_tts_seedtts.py
luojiaxuan: done

luojiaxuan:
Pushed one more commit after the approval, per offline discussion on defaults (@zhaochenyang20 PTAL at the last commit): **`optimize` now defaults to `true`** for the dots.tts latent engine and vocoder, so the bare `--model-path` deployment and the managed benchmark server run the optimized path without needing `--server-config`.

Rationale: the flag also gates the vocoder — when it is off, `merge_steps` collapses to 1 and every latent patch takes its own eager AudioVAE decode, which taxes **every** concurrency level, not just c=1 (the #1349 c=16 baseline paid this too). The optimized path is the measured, recommended deployment; `optimize: false` in the pipeline config opts out. At `max_running_requests=1` the single-request DiT compile still warms up lazily on the first requests (~2–4 min on H100); at the default `max_running_requests=16` only the vocoder chunks compile, so startup impact is small.

`--server-config` remains useful for every other knob; the defaults just stop being the slow path.

luojiaxuan:
Correction on the hardware label in my benchmark numbers here: they were measured on a **single H200**, not H100 as the text says. I mislabelled the host; the measurements themselves are unchanged and every A/B pair was run on the same card with the same container, so the relative deltas (and the WER numbers) stand as reported.

This matters for the benchmark contract in #1367, which asks for H100 and H200 results to be reported separately — please read all of my dots.tts numbers in this thread as the H200 column. H100 numbers for these configurations have not been collected.

One consequence worth flagging, since it is now a cross-hardware comparison rather than a same-hardware one: the #1349 c=16 baseline (4.329 req/s, RTF mean 0.8884) was reported on H100, while current main at c=16 on this H200 measures 4.09 req/s / RTF 0.941 over the same full EN set. Different silicon, so this is not a regression claim — but it does mean nobody has yet shown a same-card c=16 improvement from the perf series, and that gap is worth closing with a proper H100 run before we quote c=16 speedups.

PR #1374, opened 2026-08-07
[Perf] dots.tts: batch AudioVAE decoding by length bucket

AkazaAkane:
## Motivation

The dots.tts AudioVAE decoded non-streaming terminal payloads one request at a time. The initial length-bucket implementation was also blocked in production because `stream_done` messages separated terminal payloads before the shared scheduler could form a batch.

## Modifications

- batch AudioVAE decode requests by a 32-frame length bucket
- pad and crop variable-length latent sequences while preserving request order
- defer done-before-payload control messages while collecting non-streaming terminal payloads; existing stream state and stream chunks keep their ordering
- expose the existing vocoder batch size/wait knobs with defaults of 4 requests and 2 ms
- log the first real AudioVAE batch for production-path verification

## Benchmark

H200, canonical optimized `examples/configs/dots_tts.yaml`, concurrency 16, 128 requests, 10 warmups, same GPU and seed, sequential runs.

Equal-length burst (AudioVAE batching upper bound):

| Metric | main (`f4a8cd93`) | this PR | Delta |
|---|---:|---:|---:|
| Throughput (req/s) | 7.028 | 7.948 | +13.09% |
| Mean latency (s) | 2.242 | 1.994 | -11.06% |
| P95 latency (s) | 2.599 | 2.482 | -4.50% |
| Mean RTF | 1.2739 | 1.1330 | -11.06% |
| Audio throughput (s/s) | 12.370 | 13.988 | +13.08% |

Mixed-length SeedTTS English (batches form less often): throughput 3.694 -> 3.696 req/s, mean latency 4.204 -> 4.184 s, and mean RTF 1.1286 -> 1.1269. This is neutral within run-to-run noise and shows no aggregate regression.

The production server log confirms the default configuration forms real batches: `requests=4 bucket_sizes=[4]` for the equal-length burst and `requests=2 bucket_sizes=[2]` for mixed-length SeedTTS.

## Accuracy

- Qwen3-ASR-1.7B WER on 128 mixed-length SeedTTS samples: main 0.95%, PR 0.95%
- fixed-latent AudioVAE comparison, equal lengths: mean absolute PCM error <= 7e-6, max <= 6.5e-4
- fixed-latent comparison with padding/cropping: mean absolute PCM error <= 9.5e-5; the largest difference (<= 0.0386) is confined to the crop boundary

## Validation

- H200: `pytest -q tests/unit_test/dots_tts tests/unit_test/pipeline/test_streaming_simple_scheduler.py` — 71 passed
- covers equal/mixed lengths, multiple buckets, order restoration, invalid shapes, single-item fallback, and done-before-payload batching
- `pre-commit` and `git diff --check` pass

## Related

- Roadmap #1367

## Checklist


Hayden727 (approved):

sglang_omni/scheduling/streaming_simple_scheduler.py
```diff
@@ -274,6 +274,14 @@ def _collect_new_request_batch(
             if self._is_aborted(msg.request_id):
                 continue
             if msg.type != "new_request":
+                if (
+                    msg.type == "stream_done"
+                    and msg.request_id not in self._stream_payloads
+                ):
+                    # Done-before-payload only latches state, so defer it and
```
luojiaxuan: add notes xxx

luojiaxuan (approved):

PR #1377, opened 2026-08-07
[Perf] dots.tts: SGLang backbone decode CUDA graph via a model-owned feedback buffer

luojiaxuan:
## Motivation

Roadmap #1367 item "Enable SGLang backbone decode CUDA Graph". After #1370 and #1372, the c=1 profile is ≈0.44 s fixed per request plus ≈25 ms per latent patch, and the per-patch marginal cost is dominated by the eager backbone decode step (launch-bound; ~28% GPU utilization at c=1).

The blocker: SGLang 0.5.16's decode CUDA graph only stages `forward_batch.input_embeds` into a static buffer for dFlash draft workers (`spec_algorithm.is_dflash_family() and is_draft_worker`), so a dots decode replay would silently ignore the per-step latent feedback and decode from control-token embeddings.

## Modifications

- `DotsTTSSGLangModel.enable_graph_feedback(max_batch_size)`: a model-owned persistent `[max_bs, hidden]` buffer. Decode forwards always read input embeddings from this buffer, so capture bakes the buffer address into the graph and eager decode uses the identical read path. Rows beyond the live batch are padding whose outputs are discarded.
- `DotsTTSModelRunner.before_decode`: with the buffer installed, copy the per-step feedback rows into it (batch order, same stream) instead of setting `forward_batch.input_embeds`.
- Hidden-state capture: the decode graph's `can_run` gate requires an exact hidden-mode match, so the builder sets `enable_return_hidden_states` when graphs are enabled (captures FULL) and the decode hook requests FULL (decode is one token per request, so FULL ≡ LAST rows).
- `DotsTTSEngineBuilder.setup_model` installs the buffer before `init_cuda_graphs` and logs the effective backbone backend. Bare `--model-path` keeps `disable_cuda_graph: true`; the graph path is enabled from the pipeline config (`disable_cuda_graph: false` under `latent_engine.server_args_overrides`).
- Eager fallback preserved: any decode batch the graph cannot serve runs the same eager path as today.

## Benchmark

Single H100 (hyper01), Seed-TTS-Eval EN, full 1,088 samples, `--ref-format references`, seed 42, non-streaming, post-#1370 + post-#1372 main, `optimize: true`, same container and host conditions for both rows, Qwen3-ASR-1.7B WER:

| Full-set c=1 (`max_running_requests=1`) | Latency mean / p50 / p95 / p99 (s) | RTF mean / p50 / p95 / p99 | Corpus WER |
|---|---|---|---:|
| compiled, eager backbone | 1.072 / 1.041 / 1.540 / 1.839 | 0.2611 / 0.2553 / 0.3086 / 0.3590 | 1.29% |
| compiled + backbone CUDA graph (this PR) | **0.769 / 0.749 / 1.093 / 1.283** | **0.1881 / 0.1842 / 0.2232 / 0.2539** | **1.24%** |

**RTF mean −28%** at c=1 with no WER regression (1088/1088 completed, 0 failed, 0 samples above 50% WER in both rows).

Concurrency-16 (128-sample smoke, `max_running_requests=16`, `cuda_graph_max_bs=16`, seed 42): graph **3.87 req/s, RTF mean 1.0765** vs eager backbone 3.668 req/s, 1.1295 on the paired same-subset smoke (~+5% throughput), corpus WER 0.95%, 0 failed in both. (Small-n c=16 numbers are queueing-dominated; they are only comparable within the pair.) The canonical `examples/configs/dots_tts.yaml` now enables the graph; `disable_cuda_graph: true` opts back into eager backbone decode.

For context, the same benchmark on the pre-#1370 eager default measured 0.4508 mean RTF (#1372 thread), so main-to-now c=1 is 0.4508 → 0.1881 (−58%) across #1370, #1372, and this PR.

## Numerics

Graph-captured attention kernels reduce in a different order than eager launches, so outputs are not bitwise identical: on an identical 64-sample slice, 50/64 requests decode the exact same patch count and the rest shift EOS by ≤3 patches (≤0.48 s). Full-set WER above is the quality gate (graph row is 0.05 pp better).

## Validation

- `python -m pytest -q tests/unit_test/dots_tts` — 44 passed (includes new buffer-routing and before_decode tests)
- `black --check` on changed files — clean
- c=16 smoke with graphs on: 128/128 completed, 0 failed, corpus WER 0.95%

## Related

- Roadmap #1367 (backbone CUDA graph item); builds on #1349, #1370, #1372; composes with #1376 (default optimize).

## Checklist

- [x] Add unit tests.
- [x] Full-set same-card A/B with fixed seed.

### DIFF
--- a/examples/configs/dots_tts.yaml
+++ b/examples/configs/dots_tts.yaml
@@ -8,5 +8,10 @@ runtime_overrides:
     server_args_overrides:
       mem_fraction_static: 0.20
       max_running_requests: 16
+      # note (luojiaxuan): backbone decode runs through the SGLang CUDA graph
+      # with the model-owned feedback buffer; delete these two lines (or set
+      # disable_cuda_graph: true) to fall back to eager backbone decode.
+      disable_cuda_graph: false
+      cuda_graph_max_bs: 16
   vocoder:
     optimize: true
--- a/sglang_omni/models/dots_tts/engine_builder.py
+++ b/sglang_omni/models/dots_tts/engine_builder.py
@@ -78,6 +78,11 @@ class DotsTTSEngineBuilder(TtsEngineBuilder):
             raise ValueError(
                 "dots.tts uses its DiT compile path; SGLang backbone compile is disabled"
             )
+        if not bool(overrides.get("disable_cuda_graph", True)):
+            # note (luojiaxuan): the decode graph must be captured with hidden states (FULL);
+            # its can_run gate requires an exact hidden-mode match with the
+            # acoustic tail's per-step request.
+            overrides["enable_return_hidden_states"] = True
 
     def setup_model(
         self,
@@ -91,6 +96,22 @@ class DotsTTSEngineBuilder(TtsEngineBuilder):
         del checkpoint_dir, device, gpu_id
         model = model_worker.model_runner.model
         max_running_requests = int(server_args.max_running_requests)
+        if not bool(server_args.disable_cuda_graph):
+            from sglang_omni.scheduling.generation_batch_policy import (
+                get_decode_cuda_graph_max_bs,
+            )
+
+            # note (luojiaxuan): installed before init_cuda_graphs so capture bakes the buffer
+            # address into the decode graph. Generously sized: rows are tiny
+            # (hidden_size elements) and capture may pad above
+            # max_running_requests.
+            model.enable_graph_feedback(
+                max(
+                    max_running_requests,
+                    int(get_decode_cuda_graph_max_bs(server_args) or 0),
+                    256,
+                )
+            )
         model.flow.optimize = self.optimize and max_running_requests == 1
         if self.optimize and max_running_requests > 1:
             logger.info(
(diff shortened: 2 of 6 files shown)


Hayden727 (approved):

luojiaxuan:
Correction on the hardware label in my benchmark numbers here: they were measured on a **single H200**, not H100 as the text says. I mislabelled the host; the measurements themselves are unchanged and every A/B pair was run on the same card with the same container, so the relative deltas (and the WER numbers) stand as reported.

This matters for the benchmark contract in #1367, which asks for H100 and H200 results to be reported separately — please read all of my dots.tts numbers in this thread as the H200 column. H100 numbers for these configurations have not been collected.

One consequence worth flagging, since it is now a cross-hardware comparison rather than a same-hardware one: the #1349 c=16 baseline (4.329 req/s, RTF mean 0.8884) was reported on H100, while current main at c=16 on this H200 measures 4.09 req/s / RTF 0.941 over the same full EN set. Different silicon, so this is not a regression claim — but it does mean nobody has yet shown a same-card c=16 improvement from the perf series, and that gap is worth closing with a proper H100 run before we quote c=16 speedups.

Issue #1357, opened 2026-08-06
[Feature] Add reusable Prefill CUDA Graph support for SGLang-Omni models


stephenkgli:
# 1. Model Family Matrix & Tracker

**Upstream Blog:** [https://www.lmsys.org/blog/2026-08-17-advanced-cuda-graph](https://www.lmsys.org/blog/2026-08-17-advanced-cuda-graph)

---

## Merged — Default ON

* **Higgs-TTS** (`PR #1364`) — `cuda_graph_backend_prefill=BREAKABLE` default ON.
* **MOSS-Transcribe-Diarize** (`PR #1379`) — Default ON; validated with the shared prefill graph policy across H100/H200.
* **Qwen3-ASR** (`PR #1458`) — Default ON; shared capture ladder up to the effective prefill cap.
* **Fun-ASR** (`PR #1398`) — Default ON; validated with graph replay, eager fallback, and significant prefill host-wall reduction.

---

## Merged — Opt-in Only

* **Qwen3-Omni** (`PR #1381`, `PR #1519`)
  * `#1381`: Thinker text/audio → text prefill.
  * `#1519`: Speech path support.
  * *Status:* Keep opt-in: performance is workload-dependent at high concurrency and capture adds material per-rank memory/startup overhead.
* **MOSS-TTS** (`PR #1575`) — Shared `OmniPrefillInputs` adopter merged; validated for replay/fallback, quality, and multi-concurrency performance. Remains opt-in.

---

## Open — Default-ON Candidates / Policy Pending

* **ARK-ASR** (`PR #1528`) — PR proposes default ON. Exact WER parity; +10.7% / +17.6% throughput at c8/c32, with a small -1.3% c1 regression.
* **Whisper-ASR** (`PR #1530`, stacked on `PR #1634`) — Default policy pending final `whisper-large-v3` A/B/C qualification under both pre-LM modes. Historical `whisper-base` results are positive but are not sufficient for the final default-on decision.
  * `PR #1634`: Isolate encoder/decoder cache execution-path change and establish the graph-disabled adopter baseline.
  * `PR #1530`: Evaluate Prefill CUDA Graph on top of that baseline and determine the final default policy.

---

## Open — Opt-in Adoption

* **Voxtral-TTS** (`PR #1580`) — Prefill graph path validated, but service-level benefit is workload-dependent; remains opt-in.
[...]

PR #1379, opened 2026-08-07
[Perf] Enable breakable prefill CUDA graph for MOSS-TD

PrinsYin:
## Summary

Enables breakable prefill CUDA graph for MOSS-Transcribe-Diarize through the shared contract in #1364.

#1364 #1357

## Changes

- declare `supports_breakable_prefill_cuda_graph=True` for MOSS-TD;
- enable `breakable` by default with the shared `build_default_prefill_cuda_graph_bs(4096)` policy (50 buckets up to 4096 tokens);
- reuse #1364.

## Compatibility notes

- MOSS-TD resolves as a multimodal model, but this PR intentionally opts it into breakable prefill CUDA graphs based on the validation below. In SGLang 0.5.16, setting `cuda_graph_backend_prefill=breakable` explicitly skips the automatic compatibility cascade. The validated topology does not use attention/decode context parallelism, LoRA, a MoE A2A backend, or DP attention. Deployments enabling these unvalidated combinations should override `cuda_graph_backend_prefill=disabled`.
- The default ladder is aligned with `max_prefill_tokens=4096` and `chunked_prefill_size=4096`. If either limit is lowered, `cuda_graph_bs_prefill` must also be overridden.

## Validation

Primary environment: H200, SGLang 0.5.16, PyTorch 2.11.0 / CUDA 13.0, `OpenMOSS-Team/MOSS-Transcribe-Diarize` snapshot `e8681d68e7042738ffca8ac8212bc8fcb1131ab8`. The paired quality matrix was collected at `f714e845`; final H200 performance and replay validation was rerun at current head `db254ee0`.

Correctness and replay:

- Movies800Time validation, concurrency 1: shared-50 completed 800/800 requests on both H100 and H200 with real graph replay and zero fallback/OOM.
- Exact transcripts differed from eager on 4/800 requests on both GPUs. On H100, 7 and 50 buckets produced identical graph outputs. On H200, 7 buckets matched eager while shared-50 differed on 4/800, showing an additional capture-shape-sensitive effect.
- In a radix-cache-disabled H200 follow-up, the four differing samples were repeated 20 times per arm. Outputs were stable within each arm; exact-shape replay matched eager, while the shared 144-token bucket reproduced all four differences.
- AISHELL4 long audio: 2/2 completed in each mode, with 0/2 transcript mismatches and identical aggregate metrics.
- Long-audio tracing replayed 16/16 chunked prefills.
- On H100, a current-7 -> shared-50 -> shared-50 -> current-7 c16 comparison completed 3200/3200 requests with zero prefill fallback.

Paired aggregate quality on Movies800Time validation:

| GPU | Comparison | Exact transcript mismatch | CER | cpCER | DER |
|---|---|---:|---:|---:|---:|
| H100 | eager -> shared-50 | 4/800 | 5.8107% -> 5.8107% | 13.0943% -> 13.0943% | 21.0366% -> 21.0366% |
| H200 | eager-equivalent 7 -> shared-50 | 4/800 | 5.8043% -> 5.7995% | 13.0847% -> 13.0863% | 21.0300% -> 21.0236% |

Per-sample paired bootstrap, 10,000 resamples:

| GPU | ΔCER 95% CI | ΔcpCER 95% CI | ΔDER 95% CI |
|---|---:|---:|---:|
| H100 | [0, 0] pp | [0, 0] pp | [-0.00088, +0.00088] pp |
| H200 | [-0.01518, 0] pp | [0, +0.00506] pp | [-0.01891, 0] pp |

All intervals include zero. No aggregate CER/cpCER/DER regression was observed in this paired validation scope; exact output identity does not hold.

Final current-head H200 run: Movies800Time validation 800 requests/run, concurrency 16, fresh-server order eager -> shared-50 -> shared-50 -> eager:

| Mode | QPS runs | Mean QPS | Mean latency | p95 | p99 |
| eager | 32.858 / 32.632 | 32.745 | 0.430 s | 0.885 s | 1.1945 s |
| breakable shared-50 | 37.841 / 38.029 | 37.935 | 0.363 s | 0.7305 s | 0.9940 s |
| delta | | **+15.85%** | **-15.58%** | **-17.46%** | **-16.79%** |

All 3200/3200 requests completed. The two shared-50 runs attested the exact default 50-bucket ladder and recorded 924 prefill graph batches with zero eager fallback/OOM. Prefill capture averaged 4.86 s and 0.41 GB; this one-time startup cost is excluded from request wall-clock QPS.

Final bucket-policy A/B/B/A on H100:

| Policy | QPS runs | Mean QPS | Mean latency | p95 | Mean capture time | Capture memory |
| model-specific 7 | 37.562 / 36.233 | 36.898 | 0.3640 s | 0.7675 s | 1.70 s | 0.26 GB |
| shared generic 50 | 36.335 / 37.691 | 37.013 | 0.3635 s | 0.7640 s | 3.46 s | 0.42 GB |

The generic helper's +0.31% mean-QPS difference was below the roughly 3.6% run spread, with effectively unchanged mean/p95 latency. It increases one-time capture cost; this PR keeps the shared policy for consistency with the common prefill CUDA graph configuration.

Concurrency-1 performance remained within run-to-run spread and is treated as inconclusive. `full` was slightly faster in a follow-up sweep but changed 170/800 transcripts; `tc_piecewise` was slower.

These checks do not establish exact public-transcript or hidden-token parity. They show no observed aggregate CER/cpCER/DER regression on the paired Movies800Time validation set.

## Tests

- `pytest -q tests/unit_test/moss_transcribe_diarize`: 88 passed
- MOSS-TD pipeline + shared prefill policy: 28 passed
- review-fix suite (`tests/unit_test/moss_transcribe_diarize` + `test_model_capabilities.py`): 121 passed
- pre-commit hooks on changed Python files: passed
- `git diff --check`: passed


PrinsYin:
## Follow-up: backend and token-bucket sweep on H200

### Backend matrix

All runs completed 800/800 requests. Every graph backend genuinely replayed; the scheduler logs contained 1,332/1,332 graph prefills for breakable and 1,361 graph prefills plus 2 eager fallbacks for full.

| Prefill backend | Mean QPS | Run spread (range/mean) | Mean latency | p95 | QPS vs eager | Mean capture time | Capture memory |
| disabled/eager | 30.705 | 0.84% | 0.4500 s | 0.9575 s | baseline | — | — |
| breakable | 36.502 | 2.27% | 0.3685 s | 0.7615 s | +18.88% | 2.085 s | 0.26 GB |
| full | **37.649** | 1.53% | **0.3515 s** | **0.7340 s** | **+22.62%** | 2.160 s | 0.24 GB |
| tc_piecewise/eager | 34.591 | 0.09% | 0.3890 s | 0.8025 s | +12.65% | 10.525 s | 0.23 GB |
| tc_piecewise/inductor | 35.496 | 4.31% | 0.3795 s | 0.7880 s | +15.60% | 22.465 s | 0.27 GB |

Full was 3.14% higher-QPS than breakable, but it did not pass the strict correctness gate below. Both TC variants were slower than breakable and had substantially higher startup cost.

### Strict serial correctness gate

Because concurrency changes continuous-batching composition, I reran eager, breakable, and full at concurrency 1 and compared all 800 public transcripts by sample ID.

| Backend | QPS | Mean latency | p95 | Transcript mismatch vs eager |
|---|---:|---:|---:|---:|
| eager | 3.586 | 0.278 s | 0.602 s | — |
| breakable | 3.727 | 0.268 s | 0.591 s | **0/800** |
| full | 3.775 | 0.264 s | 0.587 s | **170/800** |

Eager and breakable also had the same combined transcript SHA-256. Full's aggregate accuracy did not collapse (eager CER/cpCER/DER 5.8043/13.0847/21.0300; full 5.7821/13.0005/21.0522), but aggregate metrics do not replace the per-sample output-stability gate. The eager-vs-breakable control isolates this as full-specific on this setup. I therefore did not add `full` to the validated capability declaration.

### Bucket sweep for breakable

From two c16 runs, the current buckets replayed 1,332 prefills with token counts 92-3,989. I included the previously observed long-audio chunks (`4096 x 14`, `1506`, and `2067`) when deriving aligned candidates.

| Policy | Buckets | Estimated padding/raw, short+long | Mean QPS | QPS spread | QPS vs current | Graph fallbacks | Mean capture |
| current | 128,192,384,768,1536,2048,4096 | 24.48% | 35.5115 | **0.29%** | baseline | 0 / 0 | 2.17 s |
| tuned-7 | 128,192,384,512,832,1536,4096 | 15.96% | 36.0055 | 3.90% | +1.39% | 2 / 0 | 2.13 s |
| tuned-10 | 128,192,256,320,384,448,576,832,1536,4096 | 8.88% | 35.5330 | 5.25% | +0.06% | 0 / 1 | 3.98 s |

Lower theoretical padding did not produce a repeatable wall-clock improvement. Tuned-7's mean gain was below its own run spread and introduced occasional fallback; tuned-10 was effectively equal to current, noisier, and more expensive to capture.

### Resulting PR decision

- Keep `breakable` as the preferred/default backend.
- Keep `[128, 192, 384, 768, 1536, 2048, 4096]` as the default buckets.
- Do not declare `full` or either `tc_piecewise` variant validated in this PR.
- Treat the full-backend serial transcript divergence as a separate investigation rather than trading correctness confidence for the extra ~3% QPS.

sglang_omni/models/moss_transcribe_diarize/engine_builder.py
```diff
@@ -93,6 +98,8 @@ def generation_defaults(self, *, dtype: str) -> dict[str, Any]:
             "max_prefill_tokens": 4096,
             "chunked_prefill_size": 4096,
             "sampling_backend": "pytorch",
+            "cuda_graph_backend_prefill": "breakable",
+            "cuda_graph_bs_prefill": DEFAULT_MOSS_TD_PREFILL_CUDA_GRAPH_BS,
```
stephenkgli: Consider using `build_default_prefill_cuda_graph_bs`?

sglang_omni/models/moss_transcribe_diarize/engine_builder.py
```diff
@@ -93,6 +98,8 @@ def generation_defaults(self, *, dtype: str) -> dict[str, Any]:
             "max_prefill_tokens": 4096,
             "chunked_prefill_size": 4096,
             "sampling_backend": "pytorch",
+            "cuda_graph_backend_prefill": "breakable",
```
stephenkgli: Using constant `CudaGraphBackend.BREAKABLE`

stephenkgli (commented):
Great job! Left a few comments.

AkazaAkane (commented):
Since this enables prefill CG by default for every MOSS-TD deployment, can we also smoke-test the existing moss_td_h100_dp2.yaml configuration? That config is explicitly validated as two replicas on one 80GB H100 with mem_fraction_static=0.20, while the current validation is single-replica H200. I don't think we need another full benchmark—just confirm both replicas start, run a small workload without OOM, and report whether prefill CG actually captures/replays rather than being disabled by the memory gate.

PrinsYin:
ty great points! i'll go and test what u guys mentioned

sglang_omni/models/moss_transcribe_diarize/engine_builder.py
PrinsYin: Updated to use `build_default_prefill_cuda_graph_bs`. note that performance was roughly the same, while the generic set reserved slightly more graph memory (0.42 GB vs. 0.26 GB)

PrinsYin:
Tested on moss_td_h100_dp2.yaml. Mean QPS improved from 29.946 to 32.590 (+8.83%), while mean/p95 latency decreased from 0.507/1.033s to 0.464/0.955s (-8.39%/-7.60%).

stephenkgli (approved):
LGTM

sglang_omni/models/moss_transcribe_diarize/engine_builder.py
```diff
@@ -93,6 +100,8 @@ def generation_defaults(self, *, dtype: str) -> dict[str, Any]:
             "max_prefill_tokens": 4096,
             "chunked_prefill_size": 4096,
             "sampling_backend": "pytorch",
+            "cuda_graph_backend_prefill": CudaGraphBackend.BREAKABLE,
+            "cuda_graph_bs_prefill": build_default_prefill_cuda_graph_bs(4096),
```
JiaxinD, line 114: Body and the sweep comment still document `[128, 192, 384, 768, 1536, 2048, 4096]` and "all seven buckets were exercised", but since 25d8c083 this ships `build_default_prefill_cuda_graph_bs(4096)`, which is 50 buckets. Worth updating the body so the replay/perf evidence describes what actually merges.

sglang_omni/models/moss_transcribe_diarize/engine_builder.py
```diff
@@ -93,6 +100,8 @@ def generation_defaults(self, *, dtype: str) -> dict[str, Any]:
             "max_prefill_tokens": 4096,
             "chunked_prefill_size": 4096,
             "sampling_backend": "pytorch",
+            "cuda_graph_backend_prefill": CudaGraphBackend.BREAKABLE,
```
JiaxinD, line 113: Explicit `cuda_graph_backend_prefill` locks `(prefill, backend)`, so sglang skips its whole `_apply_cuda_graph_compatibility` cascade, including the rule that disables BCG for non-allowlisted multimodal archs (MOSS-TD resolves to `is_multimodal=True` via its `text_config` sub-config). Worth a line in the body confirming the other breakable rules (CP/DCP, LoRA, MoE A2A, DP attn) do not apply here.

sglang_omni/models/moss_transcribe_diarize/engine_builder.py
```diff
@@ -93,6 +100,8 @@ def generation_defaults(self, *, dtype: str) -> dict[str, Any]:
             "max_prefill_tokens": 4096,
```
JiaxinD, line 110: These caps and the hardcoded 4096 ladder are now coupled: a deployment that lowers `max_prefill_tokens` or `chunked_prefill_size` without also overriding `cuda_graph_bs_prefill` now fails startup on unreachable buckets. Worth calling out in the body.

sglang_omni/models/moss_transcribe_diarize/__init__.py
```diff
@@ -1,2 +1,15 @@
 # SPDX-License-Identifier: Apache-2.0
 """MOSS-Transcribe-Diarize model support."""
+
+from sglang_omni.models.model_capabilities import ModelCapabilities
+
+CAPABILITIES = ModelCapabilities(
```
JiaxinD, line 6: `MossTranscribeDiarizePipelineConfig` does not set `requires_model_capabilities = True`, so these six values sit outside the parametrization in `test_model_capabilities.py` and can drift silently.

tests/unit_test/moss_transcribe_diarize/test_pipeline.py
```diff
@@ -51,6 +79,20 @@ def test_moss_transcribe_diarize_config_uses_single_batched_stage() -> None:
     }
 
 
+def test_moss_transcribe_diarize_prefill_backend_policy() -> None:
+    builder = _make_moss_engine_builder()
+
+    assert (
+        type(builder).supports_breakable_prefill_cuda_graph
+        is CAPABILITIES.supports_breakable_prefill_cuda_graph
```
JiaxinD: This assertion cannot fail, the class attribute is assigned from that same object at `engine_builder.py:24`. The bucket assert below re-evaluates the production expression too; pinning the literal list, or `max(bs) == defaults["max_prefill_tokens"]`, would catch an actual drift.

tests/unit_test/moss_transcribe_diarize/test_pipeline.py
```diff
@@ -247,6 +289,7 @@ def _bump_init_cuda_graphs() -> None:
     monkeypatch.setattr(
         engine_factory, "build_generation_batch_overrides", lambda **k: {}
     )
+
```
JiaxinD: Stray blank line, unrelated to this change.

JiaxinD (commented):
One body-level note: the dependency note about #1364 leaving a stale `test_model_capabilities.py` fixture no longer applies, every fixture supplies the new field on both this head and main.

AkazaAkane:
Please address all Jiaxin's comment and I can help running CI.

PrinsYin:
@JiaxinD Thanks, I’ve addressed the comments
I also explored a code alternative that distinguishes a model default from an operator override and reruns SGLang’s compatibility gates:
https://github.com/PrinsYin/sglang-omni/commit/fcd4d787fab36df89d1e5a064b0e7fbc97f0492d
I left it out of this PR but would love your thoughts on whether this direction makes sense

tests/unit_test/moss_transcribe_diarize/test_pipeline.py
```diff
@@ -51,6 +76,20 @@ def test_moss_transcribe_diarize_config_uses_single_batched_stage() -> None:
...
 
 
+def test_moss_transcribe_diarize_prefill_backend_policy() -> None:
+    builder = _make_moss_engine_builder()
+
+    assert type(builder).supports_breakable_prefill_cuda_graph is True
+    defaults = builder.generation_defaults(dtype="bfloat16")
+    assert defaults["cuda_graph_backend_prefill"] == "breakable"
+    assert (
+        max(defaults["cuda_graph_bs_prefill"])
```
JiaxinD, line 106: `max(bs)` alone also passes for a degenerate `[4096]` ladder, where every short prefill falls back to eager and the whole win disappears. Keeping this coupling assert and adding `== build_default_prefill_cuda_graph_bs(4096)` would pin the shape too.

JiaxinD (commented):
The H100 A/B/B/A covers throughput and 0 fallback on the 50-bucket ladder, but the 0/800 transcript parity gate still reads as the 7-bucket config. Was the concurrency-1 parity run redone on the shipped ladder?

JiaxinD:
Direction makes sense. Today's lock also skips `_apply_cuda_graph_disaggregation_roles` and the CP/DCP, LoRA, MoE A2A and DP attn rules, not just the multimodal one, and both methods are idempotent on 0.5.16, so replaying the cascade closes a real gap.

Two things I would change. Setting `model_config.is_multimodal_breakable_cuda_graph_supported = True` self grants upstream's arch allowlist from an omni capability bit that makes a different claim (Higgs keeps embeds off `ForwardBatch` via the sidecar, MOSS-TD runs real `mm_inputs` through `general_mm_embed_routine`); an upstream PR adding the arch to `multimodal_breakable_cuda_graph_supported_model_archs`, backed by the replay evidence you already have, gets the same result without writing to upstream derived state. And the tests stub out `_apply_cuda_graph_compatibility` and `get_model_config`, so they only prove two private methods were called; a real `ServerArgs` with `enable_dp_attention=True` asserting the resolved prefill backend lands on `disabled` would prove the gate actually runs.

One latent trap if you keep it: the cascade rewrites breakable to `tc_piecewise` for archs on `multimodal_piecewise_cuda_graph_supported_model_archs`, which `_validate_prefill_graph_policy` then rejects outright. Nothing in omni is on that list today.

Agree it belongs in its own PR, it changes shared `engine_factory.py` and affects Higgs too.

PrinsYin:
@JiaxinD 
Good catch! the serial parity test had not been rerun on the shipped 50-bucket ladder when I wrote that result. I reran it, and the H200 result changed from 0/800 with 7 buckets to 4/800 with 50 buckets.

I then ran an old/new commit × 7/50-bucket matrix to rule out commit drift:

| GPU | old 7 vs new 7 | old 50 vs new 50 | 7 vs 50 | eager vs 7 | eager vs 50 |
| H100 | 0/800 | 0/800 | 0/800 | 4/800 | 4/800 |
| H200 | 0/800 | 0/800 | 4/800 | 0/800 | 4/800 |

All graph arms replayed 800/800 prefills with 0 fallback/OOM. CER/cpCER/DER was unchanged on H100 and nearly unchanged on H200.

Curious what you think about his and how do you think we should handle this?

PrinsYin:
quality evaluation for the shared 50-bucket policy instead of treating exact transcript parity as the only gate.
On H100, eager vs 50 buckets differed on 4/800 transcripts, but corpus CER/cpCER/DER were identical: 5.8107% / 13.0943% / 21.0366%. The 10k paired-bootstrap 95% CIs for the deltas were [0, 0] / [0, 0] / [-0.00088, +0.00088] pp.

On H200, 50 buckets also differed on 4/800 transcripts. The CER/cpCER/DER deltas were -0.00477 / +0.00159 / -0.00634 pp, with all 95% CIs including zero.

PrinsYin:
Re-ran the performance comparison on the final PR head `db254ee0` with the shipped shared 50-bucket default: one H200, DP1/TP1, Movies800Time validation 800/run, c16, fresh-server order eager -> PCG -> PCG -> eager.

| Mode | QPS runs | Mean QPS | Mean latency | p95 | p99 |
| eager | 32.858 / 32.632 | 32.745 | 0.430 s | 0.885 s | 1.1945 s |
| breakable shared-50 | 37.841 / 38.029 | 37.935 | 0.363 s | 0.7305 s | 0.9940 s |
| delta | | **+15.85%** | **-15.58%** | **-17.46%** | **-16.79%** |

All 3200/3200 requests completed. Both PCG runs attested the exact default 50-bucket ladder and recorded 924 prefill graph batches with 0 eager fallback/OOM. Prefill capture averaged 4.86 s / 0.41 GB; startup capture is excluded from request wall-clock QPS. No external compute process was observed on the GPU during the run.

PR #1382, opened 2026-08-07
[Comm] Add rank-to-rank TP KV cache transfers

JingwenGu0829:
## Summary

- allocate one rank-local communication endpoint per stage rank (`comm_<stage>_rank<N>`)
- keep ordinary pipeline traffic on the existing leader-owned stage endpoint
- send KV prepare, ready, data-ready, and ACK messages directly between matching rank endpoints
- keep KV pool registration, destination reservation, CUDA IPC copy, leases, and ACK state local to each rank
- reject heterogeneous TP sizes before sending

## Design

For a TP=2 transfer from stage A to stage B:

```text
ordinary stage traffic -> stage_B (leader)

KV shard 0: comm_A_rank0 <-> comm_B_rank0
KV shard 1: comm_A_rank1 <-> comm_B_rank1
```

Messages retain logical `from_stage` and `to_stage` names. The receiving process already knows its local rank and derives the reply endpoint from that rank, so the wire protocol does not need `tp_rank`, `tp_size`, a registration envelope, or leader-side TP proxy queues.

`CommRouter` continues to select the physical data transport by logical stage. Rank endpoints only select the process that handles the KV handshake.

## Scope

This change supports corresponding-rank KV transfer between stages with equal TP sizes over CUDA IPC. Heterogeneous TP re-sharding remains a follow-up.

Stacked on #1315. The base branch is `kv-transfer/infrastructure`.

## Validation

- `uv run pytest -q tests/unit_test/pipeline` — 415 passed
- `uv run pre-commit run --all-files` — passed

### DIFF
--- a/sglang_omni/comm/engine.py
+++ b/sglang_omni/comm/engine.py
@@ -340,7 +340,6 @@ class CommEngine:
         source_page_indices: tuple[int, ...],
         target_pool_id: str,
         to_stage: str,
-        target_endpoint: str,
         metadata: dict[str, Any] | None = None,
         transfer_id: str | None = None,
         lease: KVPageLease | None = None,
@@ -379,7 +378,7 @@ class CommEngine:
             self._outbound_kv_requests[transfer_id] = request_id
             await send_to_endpoint(
                 self._rank_send_sockets,
-                target_endpoint,
+                self.rank_endpoints[to_stage][self.tp_rank],
                 KVTransferPrepareMessage(
                     request_id=request_id,
                     transfer_id=transfer_id,
@@ -426,7 +425,7 @@ class CommEngine:
             try:
                 await send_to_endpoint(
                     self._rank_send_sockets,
-                    target_endpoint,
+                    self.rank_endpoints[to_stage][self.tp_rank],
                     DataReadyMessage(
                         request_id=request_id,
                         from_stage=from_stage,
--- a/sglang_omni/comm/engine.py
+++ b/sglang_omni/comm/engine.py
@@ -27,6 +27,7 @@ from sglang_omni.comm.kv_transfer import (
     KVReceiver,
 )
 from sglang_omni.comm.router import CommRouter
+from sglang_omni.pipeline.control_plane import PullSocket, PushSocket, send_to_endpoint
 from sglang_omni.profiler.comm_trace import elapsed_ms as _comm_elapsed_ms
 from sglang_omni.profiler.comm_trace import emit as _comm_trace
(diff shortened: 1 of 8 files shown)


sglang_omni/comm/engine.py
```diff
@@ -364,14 +364,21 @@ async def _send_kv_pages_impl(
...
             relay.register_kv_pool(pool)
 
             ready_future = asyncio.get_running_loop().create_future()
             self._kv_ready[transfer_id] = ready_future
             self._outbound_kv_requests[transfer_id] = request_id
-            await control_plane.send_to_stage(
-                to_stage,
+            await send_to_endpoint(
+                self._rank_send_sockets,
                 target_endpoint,
```
Ratish1: target_endpoint is not an independent routing choice: the receiver sends Ready and ACK messages to the source endpoint at its own tp_rank. Accepting an arbitrary endpoint can therefore mispair rank-local transfer state when endpoints are crossed.

should we remove target_endpoint from this API and derive self.rank_endpoints[to_stage] [self.tp_rank] internally. Reject rank-channel messages whose to_stage does not match the receiving engine before mutating transfer state?

tests/unit_test/pipeline/test_kv_transfer.py
```diff
@@ -261,112 +291,289 @@ async def _run() -> None:
...
+            assert relay.put_ops[0].waited
+            assert ("op_ack", "kv-put") in relay.log.events
+        finally:
+            await source.close()
+            await destination.close()
+
+    asyncio.run(_run())
+
+
+def test_equal_tp_transfer_copies_each_shard_over_its_peer_endpoint() -> None:
```
Ratish1, line 361: This test supplies the expected endpoint itself, and the fake records only  destination page mappings. It therefore cannot distinguish which source rank produced the transferred data; crossed endpoints with the same transfer ID can still make both receivers commit.

Tag each source operation with its TP rank and assert destination rank i   receives only source rank i. After endpoint selection moves into CommEngine, call send_kv_pages() without supplying an endpoint.

sglang_omni/pipeline/control_plane.py
Ratish1: KVTransferPrepareMessage and KVTransferReadyMessage are now handled by the rank-local CommEngine sockets, while Stage._handle_message() no longer handles them. Is anything still routing these messages through StageControlPlane, or can they be removed from this?

Ratish1 (approved):
left a few comments. rest looks good to me

PR #1386, opened 2026-08-07
fix: TP support for MOSS-Transcribe-Diarize

ShaoKang22:
## Motivation

This PR fixes Tensor Parallelism (TP) support for the MOSS-Transcribe-Diarize (MOSS-TD) model. The original PR #1041 introduced TP support but was based on an older version of sglang-omni and required rebase + compatibility fixes for the current main branch (sglang 0.5.16+).

Without this fix, TP=2 configuration fails because:
1. `MossTranscribeDiarizeEngineBuilder` does not pass `tp_rank`/`nccl_port` to `ModelWorker`, causing both ranks to bind the same NCCL port (`EADDRINUSE`).
2. The factory function signature lacks `tp_size`/`tp_rank` parameters, so TP config is ignored.
3. `gpu=0` (int) in default config is incompatible with TP stage validation which expects `gpu: [0]` (list).

## Modifications

- **`sglang_omni/models/moss_transcribe_diarize/stages.py`**:
  - Add `tp_size`, `tp_rank`, `nccl_port` to `create_sglang_moss_transcribe_diarize_executor()` signature.
  - Inject `tp_size` into `server_args_overrides` and strip incompatible keys (`cuda_graph_max_bs`, `sampling_backend`, etc.) for sglang 0.5.16 compatibility.
  - Set `builder._tp_rank` and `builder._nccl_port` before calling `.build()`.

- **`sglang_omni/models/moss_transcribe_diarize/engine_builder.py`**:
  - Add `infra_kwargs()` override to forward `tp_rank`/`nccl_port` to `create_sglang_infrastructure_defer_cuda_graph()`, ensuring each rank uses distinct NCCL ports.

- **`sglang_omni/models/moss_transcribe_diarize/config.py`**:
  - Change `gpu=0` → `gpu=[0]` and add `tp_size=1` to satisfy stage validation (`gpu` must be a list with length matching `tp_size`).

## Testing

Verified on NVIDIA H200 (8× 140GB):

| Config | GPUs | Result |
|--------|------|--------|
| TP=1 | `CUDA_VISIBLE_DEVICES=6` | ✅ Server starts, inference returns correct JSON |
| TP=2 | `CUDA_VISIBLE_DEVICES=2,7` | ✅ Server starts, both ranks initialize NCCL successfully, inference returns correct JSON |

**Request format used for validation:**
```json
{
  "model": "moss-transcribe-diarize",
  "messages": [{"role": "user", "content": "&lt;|audio_pad|&gt; transcribe"}],
  "audios": ["/tmp/dummy.wav"]
}
```

**Key environment for TP=2:**
```bash
env_defaults.MASTER_ADDR=127.0.0.1
env_defaults.MASTER_PORT=&lt;unique_port&gt;
factory_args.mem_fraction_static=0.65  # leaves room for NCCL buffers
```


ShaoKang22:
@luojiaxuan take over from @haojin2 .Please review this PR at your convenience.

PR #1390, opened 2026-08-07
[ASR] Align long-audio chunking with Qwen3-ASR native limits

wirybeaver:
Qwen3-ASR serving distinguishes the operational chunk size from the model's native single-request limit and the endpoint's total-upload limit:

```text
max_audio_clip_s = 60.0
max_native_clip_s = 1200.0
max_total_audio_s = 3600.0  # inherited default
```

- `stream=false`: audio longer than 60 seconds is automatically chunked; the complete upload may be up to 3,600 seconds by default.
- `stream=true`: uses `max_native_clip_s`, accepts one native request up to 1,200 seconds, and does not chunk; longer input is rejected.
- If `max_native_clip_s` is unset for another model, the streaming limit falls back to `max_audio_clip_s`.

Therefore, 1,200–3,600-second uploads are allowed only for `stream=false`, where they are split into operational chunks. They are not accepted as one native request.

#1350 was my original implementation and validated a 1,201-second upload on a Vast.ai RTX 6000 Ada 48GB. That run used the earlier 1,200-second chunk policy. After reconciling the review feedback on #1347 and #1350, this stacked PR retains #1347's conservative 60-second non-streaming operational chunk size while preserving the separate 1,200-second native limit.

## Modifications

- retain the 60-second operational chunk size for non-streaming requests while preserving the model-owned 1,200-second native limit
- retain the 4,096-token engine capacity from Qwen3-ASR's [official vLLM wrapper](https://github.com/QwenLM/Qwen3-ASR/blob/main/README.md), while deriving each unspecified request's admission budget from decoded audio duration:
  ```text
  Engine capacity:
  C = 4096

  Per-request default:
  B = min(C, max(128, ceil(audio_duration_s × 10)))
  ```
- decode uploads when metadata duration is unavailable, while preserving the original bytes for short audio and enforcing the decoded total-duration limit
- concatenate chunk transcripts exactly as returned by the model
- avoid presenting service chunk boundaries as model-derived timestamps in `verbose_json`

The existing blocking concurrency and total-duration policies from #1347 are intentionally left unchanged so this remains a focused F1 correctness patch. Chunk-aware transcription streaming and GPU-capacity-aware scheduling policies are deferred to follow-ups based on #1347.

## Test Plan

- `685 passed` across the Qwen3-ASR unit suite and affected transcription serving tests
- all pre-commit hooks passed on the changed files

### DIFF
--- a/sglang_omni/models/qwen3_asr/config.py
+++ b/sglang_omni/models/qwen3_asr/config.py
@@ -17,7 +17,7 @@ class Qwen3ASRPipelineConfig(PipelineConfig):
     architecture: ClassVar[str] = "Qwen3ASRForConditionalGeneration"
     audio_chunking: ClassVar[AudioChunkingConfig] = AudioChunkingConfig(
         allow_audio_chunking=True,
-        max_audio_clip_s=60.0,
+        max_audio_clip_s=float(QWEN3_ASR_MAX_INPUT_SECONDS),
         max_native_clip_s=float(QWEN3_ASR_MAX_INPUT_SECONDS),
     )
 
@@ -39,9 +39,7 @@ class Qwen3ASRPipelineConfig(PipelineConfig):
             factory_args={
                 "device": "cuda:0",
                 "max_running_requests": 64,
-                # Note (Jeffro): This is the floor for the per-request output budget.
-                # The request builder will scale the actual budget with audio duration.
-                "max_new_tokens": 128,
+                "max_new_tokens": 4096,
                 "enable_torch_compile": True,
                 "torch_compile_max_bs": 2,
                 "request_build_max_workers": 8,
--- a/sglang_omni/models/qwen3_asr/config.py
+++ b/sglang_omni/models/qwen3_asr/config.py
@@ -17,7 +17,7 @@ class Qwen3ASRPipelineConfig(PipelineConfig):
     architecture: ClassVar[str] = "Qwen3ASRForConditionalGeneration"
     audio_chunking: ClassVar[AudioChunkingConfig] = AudioChunkingConfig(
         allow_audio_chunking=True,
-        max_audio_clip_s=float(QWEN3_ASR_MAX_INPUT_SECONDS),
+        max_audio_clip_s=60.0,
         max_native_clip_s=float(QWEN3_ASR_MAX_INPUT_SECONDS),
     )
 
--- a/sglang_omni/serve/transcriptions.py
+++ b/sglang_omni/serve/transcriptions.py
@@ -5,25 +5,22 @@ from __future__ import annotations
 
 import asyncio
 import logging
-import math
 import uuid
 from collections.abc import Awaitable
 
 from fastapi import Depends, FastAPI, HTTPException, Request
-from fastapi.responses import JSONResponse, PlainTextResponse, Response
+from fastapi.responses import Response
 
 from sglang_omni.client import Client, ClientError, GenerateRequest
 from sglang_omni.config import AudioChunkingConfig
 from sglang_omni.serve import speech_to_text
 from sglang_omni.serve.openai_errors import is_bad_request_error
-from sglang_omni.serve.protocol import TranscriptionResponse, TranscriptionUsage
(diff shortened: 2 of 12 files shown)


sglang_omni/models/qwen3_asr/config.py
```diff
@@ -41,15 +39,7 @@ def generation_sglang_role_to_stage(cls) -> dict[str, str]:
...
-                # Note: Cap on generated tokens per request; generation just stops there, so
-                # a too-small cap silently drops the transcript tail. Speech produces
-                # ~5 output tokens per audio second: 128 only covered ~25s, and a 60s
-                # chunk needs ~300 -- 640 doubles that. Safe to raise: context_length
-                # adds max_new_tokens on top of the audio budget (engine_builder.py),
-                # so it grows by the same amount and the audio that fits is untouched.
-                # If max_audio_clip_s is raised, raise this along with it
-                # (~5 tokens per audio second).
-                "max_new_tokens": 640,
+                "max_new_tokens": 4096,
```
luojiaxuan, line 42: SGLang budgets the requested output maximum during admission, so this charges every short request for 4096 tokens; please derive the default from decoded audio duration while preserving explicit overrides.

luojiaxuan (changes requested):
The fixed 4096-token default would regress admission efficiency for ordinary short audio.

sglang_omni/models/qwen3_asr/config.py
wirybeaver: Addressed in 6cf6ffa. The engine/context capacity remains 4,096 tokens, while an unspecified request now derives its admission budget from decoded duration (`max(128, ceil(duration_s * 10))`, capped at 4,096). Explicit `max_new_tokens` values are preserved and validated against the configured capacity. Tests cover the duration-derived default, decoded-duration wiring, explicit overrides, and invalid values.

sglang_omni/serve/transcriptions.py
```diff
@@ -521,6 +495,44 @@ def _probe_audio_duration(audio_bytes: bytes) -> float:
...
+        except ValueError as exc:
+            raise HTTPException(status_code=400, detail=str(exc)) from exc
+
+    try:
+        decoded_duration_s, plan = await asyncio.to_thread(
+            decode_audio_chunks,
+            audio_bytes,
+            chunking,
+        )
+    except (RuntimeError, ValueError) as exc:
```
Ratish1: This catch is broader than the invalid-audio contract. load_audio raises AudioDecodeError for malformed user input while deliberately preserving operational decoder failures as RuntimeError. torch.OutOfMemoryError is also a RuntimeError subclass, so this maps decoder failures and out-of-memory conditions to HTTP 400 and presents server incidents as client errors.

The client-error path should catch AudioDecodeError specifically. Operational exceptions should continue to the endpoint’s logged HTTP 500 handling. Any additional client-error types should be enumerated explicitly rather than treating every decoder RuntimeError as malformed input.

Ratish1 (approved):

sglang_omni/serve/transcriptions.py
wirybeaver: Addressed in 56b878e. `_prepare_transcription_audio` now maps only `AudioDecodeError` to HTTP 400. Operational decoder failures, including `RuntimeError` and `torch.OutOfMemoryError`, now flow through the endpoint’s logged HTTP 500 handling. Added regression coverage for both operational cases while preserving the malformed-audio 400 behavior. Verification: 163 affected serve tests passed and pre-commit passed.

wirybeaver (commented):
Self-review note on `max_audio_clip_s=60.0`: after reconciling the feedback on #1347 and #1350, I restored 60 seconds as the conservative operational chunk size for non-streaming requests. The model's separate native limit remains `max_native_clip_s=1200.0`, so valid native streaming requests are not restricted to 60 seconds. This keeps the two concerns distinct: operational chunk sizing versus the model's hard input ceiling. A GPU-capacity-aware scheduling policy can be considered separately if needed.

PR #1391, opened 2026-08-07
[Model] dots.tts: serve the SOAR checkpoint on the single-request solver

luojiaxuan:
## Motivation

dots.tts ships three checkpoints. Only `dots.tts-mf` is usable today: serving `dots.tts-soar` fails at startup because the pipeline always builds the batched acoustic tail, which raises `dots.tts continuous batching currently requires MeanFlow`. SOAR is the higher-quality checkpoint upstream, so it is worth serving even before the batched tail learns flow matching.

This is the "dots.tts SOAR checkpoint" item in #1367.

## Modifications

The single-request solver already handles flow matching: `DiTSolver(meanflow=False)` runs the CFG double branch, and `DotsTTSFlowHead` already threads `fm_cfg_sequence`, `fm_null_g_cond`, and `guidance_scale` through it. What was missing was a way to reach that path.

- Add `examples/configs/dots_tts_soar.yaml`: `max_running_requests=1` (single-request solver), `num_steps=10` (upstream SOAR default; MeanFlow uses 4), compiled tail and vocoder, backbone decode CUDA graph.
- Make the batched-tail rejection actionable: it now names the checkpoint mode and points at the SOAR config instead of only stating the MeanFlow requirement.
- Add a unit test covering a flow-matching (`meanflow: null`) checkpoint: mode detection, `init_batched_tail` rejection, per-request `num_steps`, and a finite single-request decode step.
- Document both checkpoints in `docs/basic_usage/tts.md`.

Batched (concurrency > 1) SOAR serving still needs the CFG double branch in the batched tail and stays open in #1367.

Note on repo IDs: `rednote-hilab/dots.tts-*` is the old org name and HTTP-redirects to `dots-studio/dots.tts-*`. Both resolve through `huggingface_hub`, so either works as `--model-path`; the configs use the canonical `dots-studio` IDs.

## Accuracy

Seed-TTS-Eval EN, first 100 samples, seed 42, `dots-studio/dots.tts-soar` served with the new config (`num_steps=10`, `max_running_requests=1`, CFG `guidance_scale=1.2`), 1× H200. WER from `Qwen/Qwen3-ASR-1.7B`:

| Metric | SOAR |
|---|---:|
| Completed / failed | 100 / 0 |
| Corpus WER | **1.09%** |
| Per-sample WER mean | 1.12% |
| Samples above 50% WER | 0 |
| Latency mean | 1.451 s |
| RTF mean | 0.3895 |
| Audio duration mean | 3.819 s |

For reference, `dots.tts-mf` on the same host at `max_running_requests=1` measures RTF 0.1881 / corpus WER 1.24% over the full 1,088-sample set. SOAR is roughly 2× the per-request cost, which is expected: `num_steps=10` instead of 4, and flow matching runs the CFG double branch while MeanFlow does not.

## Validation

- `python -m pytest -q tests/unit_test/dots_tts` — 45 passed, 1 skipped
- Live H200 serve + generate on `dots-studio/dots.tts-soar` with the new config, WER measured with `Qwen/Qwen3-ASR-1.7B`.

## Related

- Roadmap #1367; model support #1349.

## Checklist

- [x] Add unit tests.
- [x] Update documentation and example configuration.

### DIFF
--- a/tests/unit_test/dots_tts/test_flow_head.py
+++ b/tests/unit_test/dots_tts/test_flow_head.py
@@ -368,7 +368,7 @@ def test_flow_matching_checkpoint_runs_the_single_request_solver(tmp_path) -> No
             "rotary_bias": True,
         },
         "vocoder": {"sample_rate": 48000},
-        # SOAR and base ship without a meanflow block.
+        # note (luojiaxuan): SOAR and base ship without a meanflow block.
         "meanflow": None,
     }
     flow = DotsTTSFlowHead(
@@ -384,7 +384,7 @@ def test_flow_matching_checkpoint_runs_the_single_request_solver(tmp_path) -> No
 
     assert flow.mode == "flow_matching"
     assert not flow.is_batched
-    # Single-request serving keeps per-request num_steps and CFG.
+    # note (luojiaxuan): single-request serving keeps per-request num_steps and CFG.
     flow.validate_request(num_steps=10, ode_method="euler")
 
     with pytest.raises(ValueError, match="max_running_requests=1"):
--- a/tests/unit_test/dots_tts/test_flow_head.py
+++ b/tests/unit_test/dots_tts/test_flow_head.py
@@ -343,8 +343,6 @@ def test_validate_request_batched_gates_prompt_and_span_budget() -> None:
     )
 
 
-
-
 def test_flow_matching_checkpoint_runs_the_single_request_solver(tmp_path) -> None:
     torch.save(
         {"mean": torch.zeros(LATENT_DIM), "var": torch.ones(LATENT_DIM)},
--- a/tests/unit_test/dots_tts/test_flow_head.py
+++ b/tests/unit_test/dots_tts/test_flow_head.py
@@ -341,3 +341,73 @@ def test_validate_request_batched_gates_prompt_and_span_budget() -> None:
         prompt_patch_count=0,
         total_span_count=10**6,
     )
+
+
+
+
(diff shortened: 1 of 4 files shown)


luojiaxuan:
This one is standalone and ready — 15/16 checks green, the last TTS stage is still running. It depends on nothing else.

#1393 (the dots.tts cookbook) is the one waiting on it: the cookbook's SOAR row and its SOAR performance table only make sense once `examples/configs/dots_tts_soar.yaml` exists. Merging this first removes every forward reference from the cookbook, so please take this ahead of #1393 rather than reviewing them as a pair.

Diff is small: one config, one error message, one unit test, plus the `basic_usage/tts.md` rows.

tests/unit_test/dots_tts/test_flow_head.py
```diff
@@ -341,3 +341,71 @@ def test_validate_request_batched_gates_prompt_and_span_budget() -> None:
...
+        optimize=False,
+    )
+    with torch.no_grad():
+        for parameter in flow.parameters():
+            parameter.normal_(0.0, 0.2)
+    flow = flow.eval()
+
+    assert flow.mode == "flow_matching"
+    assert not flow.is_batched
+    # Single-request serving keeps per-request num_steps and CFG.
```
zhaochenyang20: add your name

zhaochenyang20 (changes requested):

tests/unit_test/dots_tts/test_flow_head.py
luojiaxuan: done

zhaochenyang20 (approved):

Issue #1392, opened 2026-08-07
[Bug] dots.tts: streaming kills the pipeline stage process when the vocoder compiles


luojiaxuan:
## Summary

Streaming is broken for dots.tts on the canonical deployment. With `examples/configs/dots_tts.yaml`, the first `stream: true` request returns a 0.08 s fragment and kills the pipeline stage process; every later request then fails with `Dead stage process(es) detected: pipeline (pid=..., exit=1)`. The server never recovers.

Non-streaming is unaffected: a full Seed-TTS-Eval EN sweep (1,088 requests × 6 concurrency levels) completed with zero failures on the same build.

## Reproduce

Current `main` (b7e7762), `dots-studio/dots.tts-mf`, 1× H200:

```bash
sgl-omni serve --model-path dots-studio/dots.tts-mf \
  --config examples/configs/dots_tts.yaml \
  --allowed-local-media-path docs/_static/audio --port 8000

curl -sS -N -X POST http://127.0.0.1:8000/v1/audio/speech \
  -H "Content-Type: application/json" \
  -d '{"model":"dots-studio/dots.tts-mf",
       "input":"Get the trust fund to the bank early.",
       "references":[{"audio_path":"docs/_static/audio/female-voice.wav",
                      "text":"By repeating what students say, teachers can demonstrate that they are listening. By extending what students say."}],
       "stream":true,"response_format":"pcm","seed":42}' --output out.pcm
```

Observed, three configurations, three streaming requests each:

| Config | Result |
|---|---|
| `dots_tts.yaml` (vocoder `optimize: true`, `max_running_requests=16`) | req 1 returns 7,680 bytes (0.08 s) and the stage process exits 1; req 2, 3 fail with `Dead stage process(es) detected` |
| same, `max_running_requests=1` | req 1 fails with `Detected that you are using FX to symbolically trace a dynamo-optimized function`; process survives, req 2, 3 succeed |
| same, vocoder `optimize: false` | 3 / 3 succeed, 2.72 s of audio each |

## Cause

The upstream vocoder compiles each streaming chunk shape the first time it sees it (`VocoderInference._get_compiled_stream_step`, `dots_tts/modules/vocoder/vocoder_inference.py`). That FX trace therefore runs *during serving*, in the same process as the latent engine, and the tracer's `nn.Module.__call__` patch picks up the engine's module calls. Server-side traceback from the `max_running_requests=16` run:

```
[...]

PR #1393, opened 2026-08-07
[Docs] Add the dots.tts cookbook

luojiaxuan:
## Motivation

dots.tts is the only TTS family without a cookbook page. It appears in `docs/basic_usage/tts.md` as one table row and nowhere else, so there is no documented request surface, no statement of which checkpoints work, and no performance reference.

## Modifications

- Add `docs/cookbook/dots_tts.md` and register it in the TTS section of `docs/index.rst`.
- Add `dots.tts-base` to the model table and launch section in `docs/basic_usage/tts.md`.

Page contents:

- **Architecture**: the continuous-latent pipeline (Qwen2 backbone → MeanFlow DiT → semantic encoder → AudioVAE). This is the part users get wrong: there is no codebook and no token sampler, so `temperature` / `top_k` do nothing and the solver knobs (`num_steps`, `guidance_scale`) are what matter.
- **Checkpoints**: `dots.tts-mf` (MeanFlow, continuous batching), `dots.tts-soar` and `dots.tts-base` (flow matching, single-request solver with CFG, both on `examples/configs/dots_tts_soar.yaml`). base is verified here: 20/20 requests, corpus WER 0.88%.
- **Voice cloning** and **streaming** examples in curl and Python, using reference clips already in the repo.
- **Request parameters**: top-level fields and, separately, the solver knobs under `stage_params.latent_engine`. Sending those at the top level is silently dropped, which is easy to hit and impossible to notice; the page says so and shows the working form. Engine rejections surface as HTTP 500 with the engine message.
- **Performance**: measured ladder plus the commands to reproduce each row.

## Benchmark

Seed-TTS-Eval EN against a single server from `examples/configs/dots_tts.yaml` (`max_running_requests=16`, bf16, `num_steps=4`, backbone decode CUDA graph and graph-captured acoustic tail), client `--max-concurrency` swept, seed 42, **1× H100**. Two runs per row, WER from the first run. c=1 uses the first 50 samples; the rest use the full 1,088-sample set.

| Concurrency | Throughput (req/s) | Mean latency | RTF (per-req) | audio_s/s | WER |
| 1 | 0.90 | 1.11 s | 0.284 | 3.58 | 1.06% |
| 2 | 1.48 | 1.35 s | 0.329 | 6.16 | 1.25% |
| 4 | 2.43 | 1.64 s | 0.399 | 10.14 | 1.36% |
| 8 | 3.99 | 2.00 s | 0.486 | 16.65 | 1.31% |
| 16 | 4.64 | 3.43 s | 0.830 | 19.36 | 1.31% |
| 32 | 4.43 | 7.15 s | 1.797 | 18.47 | 1.27% |

Zero failed requests in every run, no sample above 50% WER. Throughput peaks at c=16.

The page also records the latency-oriented deployment: at `max_running_requests=1` the same 50-sample slice runs at 0.64 s mean latency (RTF 0.165) against 1.11 s (RTF 0.284) at the default 16, because the compiled single-request solver replaces the batched tail padding one request up to its smallest batch bucket.

Run-to-run spread is about 2% at c=16 (three separate runs: 4.453 / 4.510 / 4.541 req/s), so treat differences below that as noise.

## Validation

- Every command on the page was executed against a live server on the documented config.
- Streaming verified on current `main` after #1395: three streaming requests complete on the canonical config, 2.72 s of audio each.
- `dots.tts-base` verified end to end: 20/20 requests, corpus WER 0.88%, RTF 0.354.

## Related

- Model support #1349; perf #1370, #1372, #1376, #1377, #1385; streaming fix #1395; roadmap #1367.

## Checklist

- [x] Update documentation.
- [x] Provide H100 performance results with reproduction commands.


zhaochenyang20:
[dots.tts](https://huggingface.co/dots-studio/dots.tts-mf) is a text-to-speech model from rednote-hilab. The Hugging Face ids are dots-studio/dots.tts-*; the older rednote-hilab/dots.tts-* ids redirect there and work too. It outputs 48 kHz speech and clones a speaker from a short reference clip plus its transcript.

we only need to link to:

https://huggingface.co/dots-studio/dots.tts-soar
https://huggingface.co/dots-studio/dots.tts-mf

zhaochenyang20:
dots.tts is a continuous-latent model, not a codec model. The backbone emits no audio tokens. Each AR step gives a hidden state; a MeanFlow DiT uses it to sample one latent patch (4 frames × 128 dims); a semantic encoder turns the patch back into the next backbone input; an AudioVAE decodes latents to waveform. No codebook, no token sampler. So temperature and top_k do nothing here — use the solver knobs (num_steps, guidance_scale) instead.

Not sure whether this is applied to all the TTS models or only to the mean flow models.

zhaochenyang20:
Single-request serving added in https://github.com/sgl-project/sglang-omni/pull/1391 (examples/configs/dots_tts_soar.yaml, max_running_requests=1). Continuous batching stays MeanFlow-only

Just let it be a todo. No need to say support it after #1391

zhaochenyang20:
dots-studio/dots.tts-base | Not supported yet
-- | --


[dots-studio/dots.tts-base](https://huggingface.co/dots-studio/dots.tts-base)	Not supported yet

this should be supported after soar?

zhaochenyang20:
curl -X POST http://localhost:8000/v1/audio/speech \
  -d '{
    "model": "dots-studio/dots.tts-mf",
    "input": "Have a nice day and enjoy south california sunshine.",
    "references": [{
      "audio_path": "docs/_static/audio/male-voice.wav",
      "text": "Hey, Adam here. Let'\''s create something that feels real, sounds human, and connects every time."
    }],
    "seed": 42
  }' \
  --output output.wav
  
  Please use existing audios. No need to add any more audios into our repo. This just makes our repo expanded.

zhaochenyang20:
Reference input:

remove this

zhaochenyang20:
Reference output:

remove this

luojiaxuan:
Review order, so this does not sit waiting: merge #1391 (SOAR, standalone, 15/16 checks green) first, then this. #1391 is the only thing this page references that is not yet on main — once it lands I will drop the "added in #1391" wording and the page stands on its own.

Everything else here is independent of open work: the MF numbers, the request-parameter table, and the demo clips are all measured against main.

Two notes for the reviewer:

- The Fun-ASR speed check failed on this PR (throughput 91.4 vs threshold 115.3). This PR is one markdown file, two wav files and one line of `index.rst`, so it cannot move inference speed; the same check passes on #1391 and #1388. I have a rerun queued for when the current run finishes.
- I am adding a second performance table for SOAR (same protocol: c=1 on the first 50 samples, c>1 on the full set). That will be one more commit here, after #1391.

luojiaxuan:
All four addressed.

**"this should be supported after soar?"** — Yes. `dots.tts-base` has a byte-identical config to SOAR (both flow matching, no `meanflow` block), so it serves on `examples/configs/dots_tts_soar.yaml` with `--model-path dots-studio/dots.tts-base`. Verified on H200: 20/20 requests, 0 failed, corpus WER 0.88%, RTF mean 0.354. The table row now says so instead of "not supported yet".

**"Please use existing audios"** and the two **"remove this"** on Reference input / Reference output — removed both added clips (`dots-1.wav`, `dots-2.wav`) and the `<audio>` players. The page now only points at reference clips that were already in the repo, through the request examples.

Also in this push: the feature sections cover all three checkpoints; the performance table stays MeanFlow-only.

luojiaxuan:
All three fixed, thanks — the parameter table was the real defect.

**P1 — solver knobs silently dropped at top level.** Confirmed: `CreateSpeechRequest` has no `speaker_scale` / `guidance_scale` / `eos_threshold` / `num_steps` / `ode_method` / `max_generate_length` / `normalize_text` field, and the model does not set `extra="allow"`, so those keys never reach the engine. The section is now split into top-level fields and a `stage_params.latent_engine` table, with a curl example that actually passes them. The "a different `num_steps` is rejected" wording is corrected to say it fails only when it reaches the engine through `stage_params`.

**P2 — 500 on engine rejection.** Documented: a rejected solver value returns HTTP 500 carrying the engine message, and that is validation, not a server fault.

**P3 — streaming example not reproducible.** The streaming request now sets `"seed": 42`. The demo clips are gone in any case, per the earlier review.

Everything else you verified matches what I measured, with one difference worth recording: on the canonical config I saw the streaming crash as an FX tracing failure inside the batched tail (`path_of_module` → `NameError`) that kills the stage process, while you saw a CUDA illegal memory access at the EOS check. Same trigger (the vocoder compiling during serving) and same workaround, different downstream symptom — I have added your repro detail to #1392.

luojiaxuan:
Re-measured c=16 on an idle H100 (lane-free GPU, three full-set runs) since the earlier numbers in this PR came from an H200:

| run | req/s | mean latency | RTF | audio_s/s |
|---|---:|---:|---:|---:|
| 1 | 4.541 | 3.504 s | 0.8456 | 18.97 |
| 2 | 4.453 | 3.573 s | 0.8627 | 18.59 |
| 3 | 4.510 | 3.526 s | 0.8525 | 18.83 |
| mean | **4.501** | **3.534 s** | **0.8536** | **18.80** |

Run-to-run spread is 2.0% on all four metrics, 0 failed requests in each run. H200 measured 4.09 req/s / RTF 0.941 for the same command, so the H100-H200 gap is real rather than noise; the workload sits at ~35% GPU utilisation, so it is bound by launch and scheduling rather than by the silicon, and host conditions dominate.

The table in the page is still the H200 sweep. Say the word and I will either relabel it as H200 or re-run the whole 1-32 ladder on the H100 so the page carries one consistent set.

Also dropped the streaming workaround paragraph now that #1395 has landed — verified on current main that three streaming requests complete on the canonical config.

zhaochenyang20 (approved):

Issue #1396, opened 2026-08-08
[Tracking] ASR Optimization Roadmap


SandyLuXY:
## Goal

This issue tracks the ASR optimization roadmap. The goal is to optimize all dedicated ASR models without regressing accuracy, streaming, diarization, or request lifecycle correctness.

## Scope
- Qwen3-ASR
- MOSS-Transcribe-Diarize
- Whisper ASR
- Fun-ASR-Nano
- ARK-ASR-3B

### Qwen3-ASR

Related issue  #1324 
#### Active
- [ ] Q-PR13

#### Landed
- [x] Q-PR1 Raise concurrency ceiling, evaluate `max_running_requests=64`. PR #1421 
- [x] Q-PR2 Tune `async_decode_min_batch_size` to `1` to improve c=1 throughput. PR #1431 
- [x] Q-PR5 Support auto-language detection PR #1403 
- [x] Q-PR6 Prefill coalescing to reduce small-batch prefill PR #1404 
- [x] Q-PR7 Support streaming transcript output PR #1359 
- [x] Q-PR8 Query Qwen3-ASR embedding cache before mel extraction  #1422 
- [x] Q-PR9 #1425 
- [x] Q-PR10 Overlap pre-LM encode with the request-build backlog  @guozhihao-224  https://github.com/sgl-project/sglang-omni/pull/1569
- [x] Q-PR11 Qwen3-ASR encoder CUDA graphs  @guozhihao-224  https://github.com/sgl-project/sglang-omni/pull/1624, #1633 

#### Closed
- [x] Q-PR4 (explore) Split preprocessing into a separate stage. @guozhihao-224
- [x] Q-PR3 Add a Qwen3-ASR precomputed-embedding fast path. @guozhihao-224 PR
- [x] Q-PR12 Move Qwen3-ASR log-mel onto the encoder stream @guozhihao-224  https://github.com/sgl-project/sglang-omni/pull/1681

#### Landed

- [x] M-PR1 Add profiling to MOSS-Transcribe-Diarize #1430 
- [x] M-PR2 Tune pending-build-aware prefill coalescing for MOSS-TD, including active-decode coalescing. Sweep a MOSS-specific request target and wait window. Show the admitted-request count per prefill and end-to-end c1/c8/c16 results. - PR #1491 
   - Refer to #1404 and #1432.
[...]

PR #1397, opened 2026-08-08
[ASR Refactor][A-PR1] ARK-ASR: move engine setup to AsrEngineBuilder

jiannan-17:
## Motivation

ARK-ASR still builds its SGLang engine directly in `stages.py`. This PR moves that setup into an `AsrEngineBuilder`, following the other SGLang-backed ASR pipelines.

## Modifications

- Added `sglang_omni/models/arkasr/engine_builder.py` with `ArkasrEngineBuilder(AsrEngineBuilder)`:
  - frontend loading and context sizing in `pre_infra_setup`
  - existing server-args defaults in `generation_defaults`
  - `init_mm_embedding_cache` in `setup_model_resources`
  - adapter construction and `request_build_*` passthrough in `make_adapters` / `extra_scheduler_kwargs`
- Rewrote `stages.py` as a thin wrapper around the builder.
  - The factory signature, defaults, and dotted-path entry `create_sglang_arkasr_executor` are unchanged, so `ArkasrPipelineConfig` needs no changes.
- Updated the factory unit test with sentinel checkpoint-config values so it catches a builder that silently falls back to defaults.
- (Removed in 9ba1e5e to address Yuhao's review comment) ~~Behavior change: context_length can now be set through server_args_overrides. Previously, this raised a duplicate-keyword TypeError; ARK-ASR now handles it the same way as Qwen3-ASR.~~

## Related Issues

- Tracker: #1396 (A-PR1)

### Setup

- Baseline: `main` at `c5c5095e`
- Model: `AutoArk-AI/ARK-ASR-3B` at snapshot `1e28271b79edc97635783bea65abc89195a09ed3`, using `bfloat16`, CUDA graphs enabled, and `torch.compile` disabled
- Dataset: English split of `zhaochenyang20/seed-tts-eval-arrow`, 100 samples per repeat
- Workload: non-streaming transcription with `benchmarks/eval/benchmark_asr_seedtts.py` at concurrency 1 and 8
- Runs: two ABBA rounds (`baseline → candidate → candidate → baseline`), with a fresh server for each block. Each block included one warmup followed by three timed repeats, giving 12 timed repeats per side at each concurrency.
- Host: 1× NVIDIA H100 80GB SXM, 16 vCPU, 128 GB RAM, Ubuntu 24.04, Python 3.12.3, `sglang 0.5.16`, and `torch 2.11.0+cu130`

Corpus WER was `0.006701` in every timed repeat for both baseline and candidate. All eight smoke checks produced the same transcript for `tests/data/query_to_cars.wav`.

CPU unit tests `tests/unit_test/arkasr/` all 11 passed.

In each run, decode CUDA graph capture for `bs=[1, 2, 4, 8, 12, 16, 24, 32]` completed before the stage became ready. All servers shut down without errors.

### Benchmark

Each value is the mean ± SD from 12 timed runs for each version.

| Concurrency | Baseline samples/s | Candidate samples/s | Delta |
| ---: | ---: | ---: | ---: |
| 1 | 12.629 ± 0.106 | 12.605 ± 0.079 | -0.19% |
| 8 | 48.803 ± 1.245 | 48.470 ± 0.830 | -0.68% |

| Concurrency | Baseline P95 latency (s) | Candidate P95 latency (s) |
| ---: | ---: | ---: |
| 1 | 0.1032 ± 0.0011 | 0.1040 ± 0.0025 |
| 8 | 0.2268 ± 0.0088 | 0.2265 ± 0.0055 |

Candidate throughput was 0.19% lower at concurrency 1 and 0.68% lower at concurrency 8. P95 latency was close to baseline, and WER was unchanged.

## Checklist

- [x] Add unit tests.

## CI

CI runs on self-hosted GPU runners and requires a maintainer to add the `run-ci` label. Once labeled, every subsequent push re-triggers CI as long as the label remains.


sglang_omni/models/arkasr/engine_builder.py
```diff
@@ -0,0 +1,113 @@
...
+            sm_version = get_visible_gpu_sm_version(self.gpu_id)
+            if sm_version is not None and sm_version >= 100:
+                defaults["mm_attention_backend"] = "triton_attn"
+        return defaults
+
+    def adjust_overrides(self, overrides: dict[str, Any]) -> None:
+        # note (jiannan-17): context_length is an explicit server-args
+        # parameter, so the override must be popped here or
+        # build_sglang_server_args would receive it twice.
+        if "context_length" in overrides:
```
AkazaAkane: Qwen3-ASR propagates the finalized context_length into its request adapter and validates prompt + max_new_tokens against that budget before admission. ARK's adapter does not consume context_length, so this currently turns an unsupported override into a configuration the request path is unaware of. Since this PR is otherwise a builder migration, can we keep it behavior-preserving and drop the new context_length override support here? If we want to support it, I think it should be wired through the ARK request builder with the corresponding budget check and test.

sglang_omni/models/arkasr/engine_builder.py
```diff
@@ -0,0 +1,113 @@
...
+            self.context_length = int(overrides.pop("context_length"))
+
+    def customize_server_args(self, server_args: Any) -> None:
```
AkazaAkane: Do we need this hook for ARK? Unlike Qwen3-ASR/MOSS-TD, self.context_length is not consumed after this point or passed into the request adapter. If the override support above is removed, this looks unnecessary as well.

sglang_omni/models/arkasr/engine_builder.py
jiannan-17: i agree, for context_length support, it is better to just keep the original scope. 
i have cleaned it up and removed the override method(adjust_overrides) in 9ba1e5ee

sglang_omni/models/arkasr/engine_builder.py
jiannan-17: customize_server_args override also removed in 9ba1e5ee. thank you

tests/unit_test/arkasr/test_pipeline.py
```diff
@@ -74,52 +79,80 @@ def test_arkasr_factory_triggers_deferred_cuda_graph_capture(
     def _bump_init_cuda_graphs() -> None:
         calls["init_cuda_graphs"] += 1
 
+    adapter_kwargs: dict[str, object] = {}
+
     model_runner = SimpleNamespace(
         model=object(), init_cuda_graphs=_bump_init_cuda_graphs
     )
-    model_worker = SimpleNamespace(model_runner=model_runner)
+    model_worker = SimpleNamespace(gpu_id=0, model_runner=model_runner)
```
AkazaAkane: This fixture looks stale against current main. The shared builder now initializes graphs through `bootstrap.init_sglang_cuda_graphs()`, which reads `model_worker.enable_prefill_input_embeds`, and the generation policy also validates the resolved prefill CUDA-graph config.

Could we rebase and update this test to the current shared-builder contract? Since the shared helper is already tested centrally, I think this ARK test can stay narrower by mocking/asserting the graph-init helper rather than duplicating its worker internals.

tests/unit_test/arkasr/test_pipeline.py
```diff
@@ -74,52 +79,80 @@ def test_arkasr_factory_triggers_deferred_cuda_graph_capture(
...
     )
 
     assert calls["init_cuda_graphs"] == (1 if want_cuda_graph else 0)
+    # note (jiannan-17): the scheduler fake records its kwargs, proving the
+    # builder forwards the stage knobs instead of dropping them.
+    assert scheduler.request_build_max_workers == 2
+    assert scheduler.request_build_max_pending == 16
+    assert adapter_kwargs["merge_factor"] == 7
+    assert adapter_kwargs["audio_token_id"] == 4242
+    assert adapter_kwargs["max_new_tokens"] == 256
```
AkazaAkane, line 166: Could we also pin the computed `context_length` and `model_arch_override` here? Those are two important ARK-specific pieces moved from `stages.py` into the builder, and both could regress while the current adapter/scheduler assertions still pass.

tests/unit_test/arkasr/test_pipeline.py
jiannan-17: thank you for pointing this out. i have rebased onto the latest main and updated the test accordingly for the new BCG changes 
done in 1b57419b. full unit suite is green on an H100 against latest main.

tests/unit_test/arkasr/test_pipeline.py
jiannan-17: makes sense, both pinned in 1b57419b, and tested the full unit suite is green. thank you

luojiaxuan (approved):
No blocking issue in the engine-builder refactor.

SandyLuXY (approved):
lgtm

PR #1402, opened 2026-08-08
[Perf] dots.tts: defer batched EOS readback to ModelRunner resolve

guozhihao-224:
## Motivation
The batched acoustic-tail path still did `eos_hits.tolist()` inside `decode_batch` and applied `finished_reason` in the same step. Single-request already stages EOS through a pinned buffer (#1372); this lands the batched half of the #1367 item: keep EOS on device through the DiT / semantic-encoder launches, and apply finish during resolve.
## Modifications
- Stage batched EOS after tail launch (CUDA: pinned buffer + event; CPU: host snapshot)
- Add `resolve_batched_eos()`; refuse to overwrite an unresolved staging buffer
- Split ModelRunner flow into launch / resolve; sync `post_decode` runs both; finish is applied only in resolve
- Wire `post_decode_launch` / `post_decode_resolve` for a later async path without enabling overlap yet
- Unit tests for staging/resolve, overwrite guard, first-patch suppress mask, and ModelRunner finish handling
## Benchmark
H200 (same card, sequential A/B), `examples/configs/dots_tts.yaml`, Seed-TTS-Eval EN **1088/1088**, seed 42, warmup 10, `--ref-format references`, generate-only then Qwen3-ASR-1.7B WER.
| Config | Version | Latency mean | RTF mean | RTF p95 | req/s | audio s/s | Corpus WER | >50% WER |
| c=16 | main | 3.818 | 0.9237 | 1.1299 | 4.167 | 17.392 | 1.34% | 0 |
| c=16 | this PR | 3.786 | 0.9171 | 1.1277 | 4.202 | 17.531 | 1.28% | 0 |
| c=8 | main | 2.227 | 0.5427 | 0.6803 | 3.586 | 14.951 | 1.26% | 0 |
| c=8 | this PR | 2.276 | 0.5546 | 0.6939 | 3.508 | 14.625 | 1.32% | 0 |

Speed delta is within same-card noise (~1–2%). Sync path still resolves immediately after launch; the win is the deferred-EOS contract for a follow-up async overlap PR. WER has no regression (Δ within the 0.10 pp budget; 0 samples above 50%). All runs: 1088/1088 completed, 0 failed.
### Repro
Generate (full EN set; set `CONC` to `16` or `8`):
```bash
  --model dots-studio/dots.tts-mf \
  --server-config examples/configs/dots_tts.yaml \
  --ref-format references \
  --lang en \
  --seed 42 \
  --warmup 10 \
  --max-concurrency "$CONC" \
  --port 8211 \
  --output-dir results/dots-eos-c${CONC} \
  --generate-only \
  --skip-gpu-cleanup
```

A/B protocol: same GPU, sequential main vs this PR; baseline checked out at merge-base 0ac9f2b7, candidate at this branch tip.

## Validation
- pytest -q tests/unit_test/dots_tts — passed
- pre-commit run --all-files — clean
- CUDA smoke: batched path uses pinned+event staging (not the host-list fallback)
- H200 full-set A/B above: 1088/1088 completed, 0 failed on every run
## Related
- Roadmap #1367
- Single-request async EOS readback: #1372
## Checklist
- [x] Add unit tests.


sglang_omni/models/dots_tts/flow_head.py
```diff
@@ -636,16 +645,7 @@ def decode_batch(
...
-        eos_flags = eos_hits.tolist()
-        finished = [
-            (
-                False
-                if state.suppress_first_eos_check and state.decoded_patches == 0
-                else bool(eos_flags[row])
-            )
-            for row, state in enumerate(states)
-        ]
+        self._stage_batched_eos(eos_hits)
```
Hayden727, line 652: Move the _batched_eos_pending check before sample_patches() and encode_feedback(). A second decode with unresolved EOS currently advances the tail KV/sequence/RNG state before _stage_batched_eos() raises while decoded_patches remains unchanged; please add a regression asserting the rejected call leaves tail state unchanged.

Hayden727 (approved):

PR #1406, opened 2026-08-08
patch create_causal_mask signature of transformers 5.12 used by qwen3 tts 0.1.1 

Safiullah136:
Qwen-tts 0.1.1 calls `create_causal_mask(..., input_embeds=..., cache_position=...)`. Under transformers==5.12.1 the parameter is spelled `inputs_embeds` and `cache_position` has been dropped from the signature entirely, so every request 500s. This normalises the call: `input_embeds` is renamed to `inputs_embeds`, and `cache_position` is absorbed (dropped) unless the installed transformers version still accepts it.

This PR also adds a smoke test to test the above change. Specifically, the test serves one audio request in a clean uv environment.

Reference Issue: #1399 

## Checklist

- [ ] Format your code according with pre-commit.
- [ ] Add unit tests.
- [ ] Update documentation / docstrings / example tutorials as needed.
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed.
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.

## CI

CI runs on self-hosted GPU runners and requires a maintainer to add the
`run-ci` label. Once labeled, every subsequent push re-triggers CI as
long as the label remains. Use `/tag-and-rerun-ci higgs` or
`/tag-and-rerun-ci moss` to select a TTS CI model, and
`/tag-and-rerun-ci fun-asr` or `/tag-and-rerun-ci qwen3-asr` to select an ASR
CI model. One selector from each family can be combined, for example
`/tag-and-rerun-ci moss fun-asr`. Draft PRs are skipped even if labeled.

Issue #1399, opened 2026-08-08
SGLang Omni Production Serving V0.1.2


zhaochenyang20:
## Context
 
We have received detailed deployment feedback from the community on Qwen3-TTS and Qwen3-ASR. The feedback is substantive, reproducible, and in every case we were able to trace it to a concrete root cause in the tree. Several of the findings are things we had not measured ourselves, in particular the behaviour of TTS past the admission ceiling.
 
The positive signal is clear: the ASR throughput gap at high concurrency is closed, the ASR language hint is no longer hard-coded, and on `main` the Qwen3-TTS streaming path holds TTFA flat at ~180ms out to 75s of generated audio with a ~45% throughput gain at c8 (#1286, #1336). What is blocking adoption is not model quality — it is that the project cannot currently be installed cleanly, pinned to a version, or driven past c32.
 
This issue tracks that gap. It has four parts.
 
**Release plan.**
 
- **V0.1.1 — ship immediately.** Packaging and install-correctness only. We are not blocked on anything here; we simply have not cut a release because no major model launch has forced one. Section 1 is the entire scope.
- **V0.1.2 — ship once Sections 2, 3, and 4 are closed.** This is the first release we would describe as production-ready for Qwen3-TTS at concurrency.
**Suggested owners** are listed per section based on recent contribution history. Please confirm or reassign in the thread — I will finalise assignment once owners ack.
 
---
 
## 1. Release and packaging — V0.1.1
 
**Owner: TBD** · Target: this week
 
Today `pyproject.toml` declares `version = "0.1.0"`, there is no PyPI distribution, the only tag is `0.1.0` with no GitHub Release attached, and the documented container is an unversioned `:dev` tag. Downstream deployment configs therefore have nothing stable to pin against, and the `:dev` image has drifted behind the `transformers 5.12.1` / `sglang 0.5.16` stack that our own docs describe. This is cheap for us and expensive for every downstream user, so it goes first.
 
### 1.1 `qwen-tts` 0.1.1 breaks against `transformers` 5.12 — every request 500s
 
This is the highest-priority item in the issue: **Qwen3-TTS cannot be served from a clean install of `main`.**
 
[...]

PR #1407, opened 2026-08-08
[Model, Qwen3-TTS] Shim the Transformers mask factories for qwen-tts 0.1.1

AkazaAkane:
## Motivation

Qwen3-TTS cannot be served from a clean install of `main`. Both docs pages instruct `uv pip install --no-deps qwen-tts==0.1.1`, and that package calls the Transformers mask factories with `input_embeds=` and `cache_position=`. Under the pinned `transformers==5.12.1` the parameter is spelled `inputs_embeds` and `cache_position` has been dropped from the signature — its docstring still lists it as "Deprecated and unused", which is why the removal is easy to miss.

Every call raises `TypeError`. This lands on the **per-request** path: Qwen3-TTS Base requires reference audio on every request, and `request_builders.py:706`/`:718` call `speech_tokenizer.encode(...)` — the upstream `qwen_tts` tokenizer — for each one. Hence every request 500s. Users are patching this locally to run at all.

This is issue §1.1, scoped to that item only.

> **Consolidates #1406.** @Safiullah136 and I converged on this independently, minutes apart. This PR takes the better half of each and credits them as co-author; #1406 can be closed in favour of it.

## Modifications

**The shim** — `sglang_omni/models/qwen3_tts/compat.py`

`_patch_mask_factories()` normalises `input_embeds` → `inputs_embeds` and absorbs `cache_position`, following the file's existing idempotent `_PATCHED_FLAG` / `inspect.signature` pattern, gated per function so it stays a no-op on the 4.57.x releases `qwen-tts` itself pins.

**It has to cover the whole mask-factory family, not just `create_causal_mask`.** The issue's prescription is incomplete: `Qwen3TTSTokenizerV2`'s decoder builds one `mask_kwargs` dict and splats it into several factories on consecutive lines. Verified on an H100 — shimming `create_causal_mask` alone moved the failure from `modeling_qwen3_tts_tokenizer_v2.py:547` to `:551` (`create_sliding_window_causal_mask`), and every request still 500'd. Note also that the three factories carrying `**kwargs` are *not* tolerant of the rename — they still require `inputs_embeds` — so the gate is signature-based rather than keyed on `**kwargs`.

One structural detail worth a reviewer's eye: the patch is applied **before** the `check_model_inputs` block, which `return`s early once that patch is in place. Appending after it would silently no-op.

**A second, independent break in the documented install line**, found by the smoke test on its first run: `numba==0.65.1` (`pyproject.toml:41`) caps numpy at 2.4, but `numpy>=1.24.0` (`:17`) has no upper bound and there is no numpy entry in `[tool.uv] override-dependencies`. Nothing constrains numpy, so `uv pip install sox einops onnxruntime` resolved it to 2.5.1, shadowed the image's 2.3.5, and broke `librosa` → `qwen_tts`. `onnxruntime>=1.17` is already a pyproject dependency, so it is dropped from the line and the two genuine extras install with `--no-deps`.

**Docs** — the two pages carried diverging install lines. Reconciled so the documented environment and the test cannot drift apart, plus a note on why `--no-deps` is the supported path rather than a workaround, and the command to run the clean-install test. The `qwen-tts==0.1.1` pin itself is unchanged.

## Related Issues

Addresses §1.1 of #1399. Consolidates #1406.

## Accuracy Test

Not applicable — no model architecture, kernel, or numerics change. The shim renames a keyword argument and drops one Transformers already documents as unused, then delegates to the original function. `test_shim_accepts_input_embeds_and_absorbs_cache_position` asserts each shimmed call returns a tensor identical to the correctly-spelled call.

## Benchmark & Profiling

No performance regression — the wrapper is two dict operations per call.

One incidental improvement: the same `TypeError` was failing the streaming vocoder's initial decoder CUDA graph capture at **every** batch size, logged as a `WARNING` and silently degrading the vocoder to eager. Capture now succeeds at batch=1 and batch=2.

## Verification

Verified end to end on an H100, in a venv built from `pyproject.toml` plus the documented install line:

| `compat.py` | result |
|---|---|
| `main` | **500** — `TypeError: create_causal_mask() got an unexpected keyword argument 'input_embeds'`, raised from the reference-audio encode |
| this branch | **200**, WAV body, `1 passed in 55.21s` |

Also run:

- `pytest tests/unit_test/qwen3_tts/test_compat.py` — 11 passed against the real `transformers==5.12.1`
- `pre-commit` on all changed files — clean

## Checklist

- [x] Add unit tests.
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed. — N/A, see above


sglang_omni/models/qwen3_tts/compat.py
```diff
@@ -11,6 +11,16 @@
...
+# Note (Akazaakane): qwen-tts splats one mask_kwargs dict across the whole
+# family, so shimming create_causal_mask alone just moves the failure.
+_MASK_FACTORY_NAMES = (
+    "create_bidirectional_mask",
+    "create_bidirectional_sliding_window_mask",
+    "create_causal_mask",
+    "create_chunked_causal_mask",
+    "create_masks_for_generate",
+    "create_sliding_window_causal_mask",
+)
```
Ratish1, line 20: can we limit limit _MASK_FACTORY_NAMES to the factories imported by the supported qwen-tts artifact. qwen-tts==0.1.1 imports only create_causal_mask and create_sliding_window_causal_mask in its model and 12 Hz tokenizer modules; it does not import the bidirectional, chunked, or generate factories.

tests/unit_test/qwen3_tts/test_compat.py
```diff
@@ -0,0 +1,129 @@
...
+# Note (Akazaakane): the factories qwen-tts splats mask_kwargs into, per the
+# Qwen3TTSTokenizerV2 decoder forward.
+_SHIMMED_FACTORIES = (
+    "create_causal_mask",
+    "create_sliding_window_causal_mask",
+    "create_chunked_causal_mask",
+)
+
+
+@pytest.fixture
```
Ratish1: This test leaves shared Transformers functions patched for later tests in the same pytest process. lets  add an autouse fixture that restores every global changed by apply_qwen_tts_transformers_compatibility_patches(), including the mask factories, check_model_inputs, and the relevant ROPE_INIT_FUNCTIONS entry.

docs/cookbook/qwen3_tts.md
```diff
@@ -18,13 +18,38 @@ in place:
 
 ```bash
 apt-get update && apt-get install -y sox
-uv pip install sox einops onnxruntime
+uv pip install --no-deps sox einops
```
Ratish1, line 21: I think we can simplify this to:

uv pip install sox einops
uv pip install --no-deps qwen-tts==0.1.1

sox and einops can be installed with their normal dependency handling without conflicts. Only qwen-tts requires --no-deps because its Transformers 4.57.3 pin would replace the project’s Transformers 5.12.1 installation.

docs/basic_usage/tts.md
```diff
@@ -18,10 +18,20 @@ Qwen3-TTS uses the upstream `qwen-tts` package. Install it without dependencies
 so the SGLang-Omni Transformers 5.12 / SGLang 0.5.16 stack remains in place:
 
 ```bash
-uv pip install --upgrade sox einops
+apt-get update && apt-get install -y sox
+uv pip install --no-deps sox einops
```
Ratish1, line 22: same

Ratish1 (commented):
we have to update this file as well sglang_omni/models/qwen3_tts/stages.py , _QWEN_TTS_INSTALL_HINT tells users to install qwen-tts “and its Transformers 4.57.3 requirement.” Following that instruction replaces SGLang-Omni’s pinned Transformers stack and defeats the compatibility shim being added here.

lets update the runtime hint to give the same supported installation instructions as the documentation, including installing qwen-tts with --no-deps. keep the message concise and clear

sglang_omni/models/qwen3_tts/compat.py
AkazaAkane: Agreed, narrowed to create_causal_mask and create_sliding_window_causal_mask.

tests/unit_test/qwen3_tts/test_compat.py
AkazaAkane: Added an autouse fixture restoring all three globals apply_...() mutates — the mask factories, generic.check_model_inputs, and the ROPE_INIT_FUNCTIONS["default"] entry — plus a canary test asserting they're pristine at test start. Verified by running the suite in-process and inspecting afterwards; all four come back clean.

docs/cookbook/qwen3_tts.md
AkazaAkane: Adopted, in both docs pages, the runtime hint and the clean-install test fixture together.

One thing worth flagging: uv pip install sox einops onnxruntime did break this environment during testing — it resolved numpy to 2.5.1, past the ceiling numba==0.65.1 imposes, which broke librosa and with it import qwen_tts. onnxruntime is already a project dependency, so it's dropped from the line and I've documented why. Whether sox einops alone stays clean unconstrained I haven't confirmed on a GPU yet — but the clean-install test now runs exactly that command, so the next run settles it either way.

docs/basic_usage/tts.md
AkazaAkane: Done — same change applied there.

docs/cookbook/qwen3_tts.md
AkazaAkane: Tried it, and it doesn't hold — reverted in a2c2c525.

On an H100, uv pip install sox einops without --no-deps resolves numpy to 2.5.1, past the ceiling numba==0.65.1 imposes (numpy<=2.4). librosa then fails to import and import qwen_tts dies before the server starts:

File ".../librosa/filters.py", line 52, in <module>
    from numba import jit
ImportError: Numba needs NumPy 2.4 or less. Got NumPy 2.5.
No onnxruntime in that command — sox alone is enough, since pysox declares an unpinned numpy. The root cause is that pyproject.toml has numpy>=1.24.0 with no upper bound against a hard numba==0.65.1 pin, so any unconstrained install of a numpy-dependent package can break the environment. Bounding numpy in pyproject.toml would fix the class of problem and let this line be simplified as you suggest — happy to open a separate issue for that if you'd like, since it's a dependency change beyond this PR.

AkazaAkane:
Removed tests/test_model/test_qwen3_tts_clean_install.py and the cookbook section referencing it, and strengthened tests/unit_test/qwen3_tts/test_compat.py instead — now 17 tests, no GPU needed.

Beyond the shim behaviour, it now guards the install line itself: the commands in both docs pages must agree with each other and with the runtime _QWEN_TTS_INSTALL_HINT, and must keep --no-deps with no onnxruntime. Verified those bite by reinjecting the simplified line — three tests fail.

tests/unit_test/qwen3_tts/test_compat.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def test_transformers_globals_are_pristine_at_test_start() -> None:
+    """Canary for patch leakage between tests in one pytest process."""
+    for name, original in _PRISTINE.items():
+        assert getattr(masking_utils, name) is original
+    assert not getattr(generic.check_model_inputs, compat._PATCHED_FLAG, False)
+
+
+@pytest.mark.parametrize("name", compat._MASK_FACTORY_NAMES)
```
Ratish1: _PRISTINE already calls getattr(masking_utils, name) for every configured  factory during module collection, so a missing or misspelled name fails before this test can run. can we remove it.

Ratish1 (approved):
LGTM

tests/unit_test/qwen3_tts/test_compat.py
AkazaAkane: Pushed the redundant-test removal

PR #1408, opened 2026-08-08
[Enhancement] dots.tts: document and gate acoustic-tail pool memory

guozhihao-224:
## Motivation
Continuous batching eagerly allocates acoustic-tail state as`max_running_requests × (max_generate_length + 1)` full-length DiT / encoder pools. Startup only logged `slots` / `nfe` / `patch_capacity`, with no byte formula and no check against free CUDA memory. On tighter GPUs this either OOMs late or leaves operators guessing which knobs to lower.
`mem_fraction_static` only budgets the SGLang backbone KV cache and does not cover these pools.

This lands the Memory/admission bookkeeping items from #1367: document and log the pool formula, validate against free VRAM before allocate, and make slot-exhaustion a clear admission failure. Parameters are never silently resized. Incremental / bucketed tail-KV allocation is out of scope.

## Modifications
- Add `estimate_acoustic_pool_bytes()` matching `_allocate_pools` shapes (DiT KV, encoder KV, scratch, aux) and `AcousticPoolMemoryEstimate`
- Log estimated GiB breakdown plus `cuda_free` / `cuda_total` before allocate; verify allocated `nbytes` afterward
-  `validate_acoustic_pool_memory()`: fail fast when free VRAM < pools + 15% headroom for graphs/workspace; suggest how many full-length slots would fit
- Clarify live `acquire_slot` exhaustion as an admission error (no under-request resize)
- Cookbook: Memory and capacity section (formula, `mem_fraction_static` scope, which knobs shrink the pool)
- Unit tests for estimate↔allocated nbytes, linear slot scaling, mocked VRAMreject/pass, and admission error wording

## Benchmark
Not a throughput PR. Same-card speed A/B is not the acceptance bar here. H200 startup smoke (`examples/configs/dots_tts.yaml`, GPU1):
| Config | slots × patch_capacity | Estimated pools | Allocated | cuda_free | Result |
| --- | --- | ---: | ---: | ---: | --- |
| default | 16 × 501 | 16.75 GiB (dit_kv 11.03 / encoder_kv 1.47 / scratch 4.23 / aux 0.01) | 16.75 GiB | 110.07 GiB | gate pass |

Estimate matched allocated nbytes. The fail-fast path is covered by a mocked`mem_get_info` unit test (this H200 has enough free VRAM to reject naturally).

## Validation
- `pytest -q tests/unit_test/dots_tts/test_tail.py -k 'not cuda_graph'` — 9 passed
- `pre-commit run --all-files` — clean (if already run on the branch)
- H200 real startup log above: estimate == allocated, gate passes

## Related
- Roadmap #1367

## Checklist
- [x] Add unit tests.
- [ ] Benchmark A/B on the same card class with fixed seed and dataset slice.
  (N/A for this memory/admission PR; startup smoke + unit tests instead.)

### DIFF
--- a/sglang_omni/models/dots_tts/tail.py
+++ b/sglang_omni/models/dots_tts/tail.py
@@ -186,6 +186,158 @@ class DotsTtsTailSpec:
         return self.patch_capacity * self.unit_len
 
 
+def _dtype_nbytes(dtype: torch.dtype) -> int:
+    return torch.empty((), dtype=dtype).element_size()
+
+
+def _gib(num_bytes: int) -> float:
+    return float(num_bytes) / float(1 << 30)
+
+
+@dataclass(frozen=True)
+class AcousticPoolMemoryEstimate:
+    """Eager acoustic-tail pool byte budget matching _allocate_pools shapes."""
+
+    dit_kv_bytes: int
+    encoder_kv_bytes: int
+    scratch_bytes: int
+    aux_bytes: int
+    total_bytes: int
+    num_slots: int
+    patch_capacity: int
+    nfe: int
+    dtype: torch.dtype
+
+    @property
+    def bytes_per_slot(self) -> int:
+        return self.total_bytes // max(int(self.num_slots), 1)
+
+
+def estimate_acoustic_pool_bytes(
+    *,
+    spec: DotsTtsTailSpec,
+    dit_layers: int,
+    dit_heads: int,
+    dit_head_dim: int,
+    encoder_layers: int,
+    encoder_heads: int,
+    encoder_head_dim: int,
+    encoder_block: int,
+    encoder_conv_channels: int,
+    encoder_conv_padding: int,
+    mods_width: int,
+    dtype: torch.dtype,
+) -> AcousticPoolMemoryEstimate:
+    """Sum pool tensor bytes; excludes weights, backbone KV, and graph workspace."""
+    elem = _dtype_nbytes(dtype)
+    bool_elem = _dtype_nbytes(torch.bool)
+    slots = int(spec.num_slots)
+    nfe = int(spec.nfe)
+    dit_tokens = int(spec.dit_cache_tokens) + int(spec.unit_len)
+    dit_query = 2 * int(spec.unit_len)
+    dit_scratch_tokens = int(spec.dit_cache_tokens) + dit_query
+    encoder_tokens = int(spec.patch_capacity) * int(encoder_block)
+
+    dit_kv = (
+        2
+        * nfe
+        * int(dit_layers)
+        * slots
+        * int(dit_heads)
+        * dit_tokens
+        * int(dit_head_dim)
+        * elem
+    )
+    encoder_kv = (
+        2
+        * int(encoder_layers)
+        * slots
+        * int(encoder_heads)
+        * encoder_tokens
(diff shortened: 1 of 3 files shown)


sglang_omni/models/dots_tts/tail.py
```diff
@@ -186,6 +186,158 @@ def dit_cache_tokens(self) -> int:
...
+    headroom_ratio: float = 0.15,
+) -> None:
+    """Raise if free CUDA memory cannot hold the pools plus headroom."""
+    # note (guozhihao-224): CPU paths skip this gate so unit tests can allocate
+    # tiny pools; never silently lower max_running_requests or patch capacity.
+    if device.type != "cuda":
+        return
+    if headroom_ratio < 0.0:
+        raise ValueError("dots.tts acoustic pool headroom_ratio must be non-negative")
+    free_bytes, total_bytes = torch.cuda.mem_get_info(device)
```
Hayden727, line 321: This checks only driver-reported free VRAM, which excludes reusable blocks held by PyTorch's caching allocator. After caching 256 MiB on both H200 containers, I reproduced `free < required < free + (memory_reserved - memory_allocated)` while this function still raised, so a valid startup after model loading can be rejected; empty the unused cache before sampling (or otherwise include allocator-reusable bytes) and cover this case.

sglang_omni/models/dots_tts/tail.py
```diff
@@ -186,6 +186,158 @@ def dit_cache_tokens(self) -> int:
...
+    # note (guozhihao-224): 15% headroom covers CUDA-graph capture and scratch
+    # beyond the eager pool tensors themselves.
+    required = int(estimate.total_bytes * (1.0 + float(headroom_ratio)))
+    if free_bytes >= required:
+        return
+    fit_slots = max(int(free_bytes // max(estimate.bytes_per_slot, 1)), 0)
```
Hayden727: `fit_slots` omits the same headroom enforced above, so the suggested setting can immediately fail again: 12 GiB across 16 slots with 4 GiB free reports 5 slots, but those five require 4.31 GiB at 15% headroom. Include `1 + headroom_ratio` in the per-slot divisor and assert the recommendation passes the gate.

sglang_omni/models/dots_tts/tail.py
```diff
@@ -344,17 +563,32 @@ def _allocate_pools(self, mods_width: int) -> None:
...
     def acquire_slot(self) -> int:
         if not self._free_slots:
+            # note (guozhihao-224): slot exhaustion is an admission failure, not a
+            # cue to resize the pool under a live request.
+            in_use = int(self.spec.num_slots)
             raise RuntimeError(
-                "dots.tts acoustic tail ran out of slots; lower concurrency"
+                "dots.tts acoustic tail admission failed: ran out of slots "
+                f"({in_use}/{in_use} in use). Lower client concurrency or "
+                "max_running_requests; the engine does not silently shrink capacity."
```
Hayden727: Lowering `max_running_requests` after slot exhaustion reduces the pool and makes the same concurrency fail sooner; the cookbook correctly says to raise it when memory allows. Change this to ‘lower client concurrency or raise `max_running_requests`’ and make the test assert the direction.

Hayden727 (approved):

PR #1411, opened 2026-08-08
[Perf][A-PR4] Batched encoder inference for ARK-ASR

nagisa-kunhah:
## Motivation

Related: #1396 

## Modifications

- Batch all cache-miss ARK-ASR audio features in
  `ArkasrForConditionalGeneration.get_audio_feature`:
  - read each item's valid mel-frame length from
    `feature_attention_mask`;
  - trim pre-existing right padding, including arbitrary values in the padded
    region;
  - pad only the time dimension to the longest item in the batch;
  - make one `audio_encoder` call and split/crop embeddings back to the
    original item order and token counts.
- Add a raw mel-frame attention mask to `ArkAudioTower` and
  `ArkAudioMLPAdapter`.
  - clear invalid input frames before `conv1`;
  - clear invalid `conv1` outputs before the stride-2 `conv2` operation;
  - downsample the mask for SDPA, excluding padded keys/values;
  - clear padded hidden states after every encoder layer and after adapter
    LayerNorm.
- Add CPU regression tests for:
  - one encoder call for a variable-length batch;
  - numerical equality between batched and serial embeddings;
  - isolation from non-zero garbage in pre-padded audio regions.
- Add an opt-in real-model E2E test. It starts `sgl-omni serve`, sends eight
  distinct variable-length WAV payloads concurrently to
  `/v1/audio/transcriptions`, validates their transcripts, and verifies that
  the server produced a multi-sequence prefill batch.

## Related Issues

Related #1396

### Encoder numerical regression

The CPU regression test compares batched embeddings with serial embeddings for
mixed mel lengths `[9, 18, 25]`:

```python
torch.allclose(batched, serial, atol=1e-5, rtol=1e-5)
```

It also fills a short sample's pre-padded tail with `50.0` and verifies the
batched result remains numerically equal to serial inference.

### End-to-end accuracy versus `main`

The PR checkout and the unmodified `main` commit were evaluated with the same
ARK-ASR-3B checkpoint, a single NVIDIA H20, SeedTTS-50 English reference
audio, and concurrency 8. Each side used a fresh serving process.

| Metric | `main` | This PR | Delta |
|---|---:|---:|---:|
| Successful requests | 50 / 50 | 50 / 50 | 0 |
| Corpus WER | 0.0074 | 0.0074 | 0.0000 |
| Maximum sample WER | 0.0909 | 0.0909 | 0.0000 |
| Per-sample transcript differences | - | 0 / 50 | no change |

The raw per-sample benchmark records were compared by sample ID: all 50
predicted transcripts are identical between `main` and this PR.

Additional SeedTTS-mini smoke coverage on this PR:

| Split | Samples | Successful | Corpus WER |
|---|---:|---:|---:|
| English | 5 | 5 / 5 | 0.0000 |
| Chinese | 5 | 5 / 5 | 0.0000 |

### Reproduce accuracy evaluation

```bash
export MODEL_PATH=Audio8/ARK-ASR-3B
export HF_ENDPOINT=https://hf-mirror.com

python -m benchmarks.dataset.prepare --dataset seedtts-50

# Start the server in the checkout under test. Confirm the custom-code prompt.
sgl-omni serve --model-path "$MODEL_PATH" --host 127.0.0.1 --port 18000

  --host 127.0.0.1 \
  --port 18000 \
  --meta zhaochenyang20/seed-tts-eval-50-arrow \
  --lang en \
  --max-samples 50 \
  --concurrencies 8 \
  --repeats 1 \
  --model-path "$MODEL_PATH" \
  --disable-resource-monitor \
  --output asr_seedtts50.json \
  --save-raw-dir asr_seedtts50_raw
```

Run the same command once from an unmodified `main` checkout and once from the
PR checkout. Compare the resulting JSONL records by `sample_id` and predicted
text.

### `main` comparison: SeedTTS-50, concurrency 8

Both runs used the same checkpoint, hardware, server options, benchmark input,
and one measured repeat. Resource monitoring was disabled to avoid adding
sampling overhead. Results include the full serving path: HTTP request build,
audio preprocessing, encoder, LM prefill/decode, and scheduling.

| Metric | `main` | This PR | Improvement |
|---|---:|---:|---:|
| Wall-clock time | 2.823 s | 2.454 s | 13.1% lower |
| Throughput | 17.712 req/s | 20.372 req/s | **15.0% higher** |
| RTFx | 81.375 | 93.593 | **15.0% higher** |
| Mean latency | 0.440 s | 0.386 s | 12.3% lower |
| P50 latency | 0.342 s | 0.350 s | 2.3% higher |
| P95 latency | 0.954 s | 0.704 s | **26.2% lower** |
| Mean RTF | 0.0964 | 0.0845 | 12.3% lower |
| P95 RTF | 0.2148 | 0.1506 | **29.8% lower** |

The scheduler logs in the PR run show real multi-request prefill batches (for
example, `#new-seq: 2`, `#new-seq: 6`, `#new-seq: 5`, and `#new-seq: 4`).

### Isolated encoder-path burst comparison

For a second, encoder-focused measurement, the same real serving stack was
run with eight concurrent, cache-miss, variable-length WAV requests. A
temporary diagnostic-only baseline restored the previous per-item encoder loop;
the temporary switch was removed before this PR was staged.

| Encoder path | 8-request E2E wall time | Relative aggregate throughput |
|---|---:|---:|
| Previous per-item encoder loop | 0.749 s | 1.00x |
| This PR batched encoder | 0.214 s | **3.50x** |

This is intentionally reported separately from the SeedTTS-50 result: the
3.50x figure is a short burst that emphasizes encoder utilization, whereas the
15.0% figure is the more representative full-dataset E2E result.

### Run the opt-in E2E test

```bash
export MODEL_PATH=Audio8/ARK-ASR-3B

RUN_ARKASR_BATCHED_ENCODER_E2E=1 \
ARKASR_MODEL_PATH="$MODEL_PATH" \
pytest -q tests/test_model/test_arkasr_batched_encoder_e2e.py -s
```

The test is skipped by default because it requires CUDA and a local model
checkpoint.

## Checklist

- [x] Code was checked with `ruff` and `git diff --check`.
- [x] Add unit tests.
- [x] Add an opt-in real-model E2E test.
- [x] Update docstrings for the new masking contract.
- [x] Provide `main` throughput / latency benchmark and accuracy evaluation.

## CI

the label remains. Use `/tag-and-rerun-ci higgs` or


luojiaxuan (dismissed):
Please replace Fixes #1396 with Related #1396 so merging this ARK optimization does not close the ASR roadmap.

nagisa-kunhah:
> Please replace Fixes #1396 with Related #1396 so merging this ARK optimization does not close the ASR roadmap.

@luojiaxuan Thanks! Done.

sglang_omni/models/arkasr/audio_tower.py
```diff
@@ -181,10 +181,55 @@ def __init__(self, config):
...
+                dtype=inputs_embeds.dtype
+            )
         inputs_embeds = F.gelu(self.conv2(inputs_embeds))
         inputs_embeds = inputs_embeds.permute(0, 2, 1)  # [B, T_down, D]
+        frame_mask = None
+        sdpa_mask = None
+        if input_frame_mask is not None:
+            # conv2 has stride 2, kernel 3, and padding 1.  Its output index i
+            # is valid exactly when input frame 2*i is valid.
+            frame_mask = input_frame_mask[:, ::2]
```
WilsonZheng0327: this should be contiguous I believe.
SDPA will dispatch to basic MATH rather than optimized fused kernels if any input's last stride dim is not 1.
[:, ::2] creates a view of input_frame_mask such that the stride = 2, which means SDPA doesn't get any of the supposed optimization.

please test and compare results.

WilsonZheng0327 (dismissed):
For comments, make sure to remove any AI fluff (if exists), and add 
`note (xxx):` for comment ownership.

Thanks!

sglang_omni/models/arkasr/audio_tower.py
nagisa-kunhah: Thanks for the reminder. I also found a 7% performance improvement after making it contiguous, and I've updated the PR now.

AkazaAkane:
Could you check whether this PR depends on A-PR1. If so I would recommend wait until A-PR1 merged.

nagisa-kunhah:
> Could you check whether this PR depends on A-PR1. If so I would recommend wait until A-PR1 merged.

@AkazaAkane  No. To my understanding, the only overlap between A-PR4 and A-PR1 is `tests/unit_test/arkasr/test_pipeline.py`. A-PR4 only adds batched audio encoding in `audio_tower.py` and `sglang_model.py`, and leaves the ARK-ASR stage factory interface unchanged.

sglang_omni/models/arkasr/sglang_model.py
```diff
@@ -55,22 +56,99 @@ def pad_input_ids(self, input_ids: List[int], mm_inputs: MultimodalInputs):
...
+        # encode.  The tower derives its own post-conv2 SDPA mask.
+        if batch_size == 1:
+            encoder_mask = None
+        else:
+            frame_index = torch.arange(max_frames, device=device).unsqueeze(0)
+            encoder_mask = frame_index < torch.tensor(
+                mel_lengths, device=device, dtype=torch.long
+            ).unsqueeze(1)
+
+        encoded = self.audio_encoder(batched, attention_mask=encoder_mask)
```
AkazaAkane, line 168: Could we validate the peak-memory behavior at the largest encoder batch this path can actually receive?

SGLang batches all multimodal cache misses in one embedding call, while ARK allows up to 32 running requests. This changes the encoder from repeated B=1 forwards to one padded B=N forward, so peak activation memory can increase substantially when several long clips arrive together.

Please include at least c=16/32 with peak GPU memory, ideally with long/mixed-length inputs. If B=32 is not intended to be supported safely, we probably need an explicit encoder batch cap/fallback rather than relying on the scheduler batch shape.

nagisa-kunhah: @AkazaAkane Thanks for your review! I ran experiments on an H20. The peak memory scaled as follows:
| encoder batch size | peak allocated | peak reserved |
|-------------------:|---------------:|--------------:|
| B=1                | ~1.7 GiB       | ~1.7 GiB      |
| B=8                | ~4.7 GiB       | ~4.8 GiB      |
| B=16               | ~8.1 GiB       | ~8.4 GiB      |
| B=32               | ~15.0 GiB      | ~15.4 GiB     |

However, when I tested the full SGLang‑Omni serving path with 30‑second WAV inputs, SGLang allocated roughly 77.4 GiB to the KV cache, leaving only about 9.85 GiB of available memory after model loading. Under this configuration, an end‑to‑end test with c=16 triggered an OOM inside the ARK audio encoder. Given this constraint, should we provide a .yaml config file (or add a note in the documentation) to remind H20 users about this memory limitation in advance?

AkazaAkane: I think this is not ARK-specific in principle: other ASR models also have encoder memory that grows with batched audio length. The difference is that Qwen3-ASR and Fun-ASR already bound the pre-LM encoder batch size to 8, while MOSS-TD uses 2, so encoder batching is decoupled from request concurrency.

For ARK, I would prefer following the same pattern rather than relying only on an H20-specific YAML/doc warning. Since c=16 with 30s inputs can OOM in the normal serving configuration, could we add an explicit encoder batch cap (e.g. 8) and chunk larger batches internally? A documentation note about mem_fraction_static would still be useful as additional guidance.

nagisa-kunhah:
Update:
After fixing the potential OOM issue by setting an explicit encoder batch cap of 8, the performance of this PR is as shown below.

| Concurrency `c` | Successful requests | Wall-clock time | Throughput | Mean latency | P95 latency | Peak GPU memory |
| 1  | 50/50 | 5.389 s | 9.277 req/s | 0.108 s | 0.146 s | 86.64 GiB |
| 2  | 50/50 | 2.394 s | 20.885 req/s | 0.095 s | 0.140 s | 86.64 GiB |
| 4  | 50/50 | 1.538 s | 32.508 req/s | 0.122 s | 0.172 s | 86.64 GiB |
| 8  | 50/50 | 1.001 s | 49.956 req/s | 0.152 s | 0.238 s | 86.64 GiB |
| 16 | 50/50 | 0.720 s | 69.425 req/s | 0.217 s | 0.305 s | 86.64 GiB |
| 32 | 50/50 | 0.463 s | 108.031 req/s | 0.251 s | 0.374 s | 86.64 GiB |

The corresponding raw GPU measurements were:

| Concurrency `c` | Peak device memory | Peak ARK stage process memory | Minimum free memory |
|---:|---:|---:|---:|
| 1  | 88719 MiB | 88196 MiB | 9152 MiB |
| 2  | 88719 MiB | 88196 MiB | 9152 MiB |
| 4  | 88719 MiB | 88196 MiB | 9152 MiB |
| 8  | 88719 MiB | 88196 MiB | 9152 MiB |
| 16 | 88719 MiB | 88196 MiB | 9152 MiB |
| 32 | 88719 MiB | 88196 MiB | 9152 MiB |

The same 50 English SeedTTS samples were evaluated at every concurrency level:

| Concurrency `c` | Evaluated requests | Corpus WER | Maximum per-sample WER |
|---:|---:|---:|---:|
| 1  | 50/50 | 0.0074 | 0.0909 |
| 2  | 50/50 | 0.0074 | 0.0909 |
| 4  | 50/50 | 0.0074 | 0.0909 |
| 8  | 50/50 | 0.0074 | 0.0909 |
| 16 | 50/50 | 0.0074 | 0.0909 |
| 32 | 50/50 | 0.0074 | 0.0909 |

sglang_omni/models/arkasr/sglang_model.py
nagisa-kunhah: @AkazaAkane  Yes, I'm aggree with your suggestion. I have fix follower the approach mentioned above and the performance can be show below. Thanks!

AkazaAkane (approved):

SandyLuXY:
@nagisa-kunhah Could we run the benchmark on the full seedTTS en split (1088 samples) instead of the 50 samples for reliable speed metrics?

AkazaAkane:
## Focused 3-repeat experiment

One discarded warm-up followed by three measured repeats. Wall time is the per-repeat mean.

| C | Revision | Success | Wall (s) | Req/s | PR Δ | RTFx |
|---|---|---|---|---|---|---|
| 8 | main | 3264/3264 | 16.892 | 64.409 | — | 305.049 |
| 8 | PR | 3264/3264 | 16.283 | 66.838 | +3.77% | 316.551 |
| 16 | main | 3264/3264 | 10.813 | 100.623 | — | 476.559 |
| 16 | PR | 3264/3264 | 10.732 | 101.401 | +0.77% | 480.245 |
| 32 | main | 3264/3264 | 7.880 | 138.094 | — | 654.026 |
| 32 | PR | 3264/3264 | 7.636 | 142.490 | +3.18% | 674.847 |

| C | Revision | Mean latency (s) | P50 (s) | P95 (s) | Mean RTF | P95 RTF | Corpus WER | Max WER |
| 8 | main | 0.1238 | 0.1199 | 0.1656 | 0.02694 | 0.0381 | 0.0112 | 0.2857 |
| 8 | PR | 0.1193 | 0.1160 | 0.1609 | 0.02596 | 0.0366 | 0.0112 | 0.2857 |
| 16 | main | 0.1582 | 0.1609 | 0.2005 | 0.03435 | 0.0473 | 0.0112 | 0.2857 |
| 16 | PR | 0.1569 | 0.1597 | 0.1987 | 0.03404 | 0.0467 | 0.0112 | 0.2857 |
| 32 | main | 0.2297 | 0.2214 | 0.3329 | 0.04987 | 0.0725 | 0.0112 | 0.2857 |
| 32 | PR | 0.2223 | 0.2165 | 0.3203 | 0.04828 | 0.0702 | 0.0112 | 0.2857 |

Aggregate throughput delta: +2.58% arithmetic mean, +2.51% pooled. Verdict: E2E performance neutral.

## Original 1-repeat experiment

| C | Revision | Success | Wall (s) | Req/s | PR Δ | RTFx |
|---|---|---|---|---|---|---|
| 1 | main | 1088/1088 | 83.298 | 13.062 | — | 61.861 |
| 1 | PR | 1088/1088 | 84.377 | 12.895 | -1.28% | 61.070 |
| 2 | main | 1088/1088 | 47.829 | 22.748 | — | 107.735 |
| 2 | PR | 1088/1088 | 46.007 | 23.648 | +3.96% | 112.002 |
| 4 | main | 1088/1088 | 27.643 | 39.359 | — | 186.408 |
| 4 | PR | 1088/1088 | 27.065 | 40.200 | +2.14% | 190.391 |
| 8 | main | 1088/1088 | 16.511 | 65.895 | — | 312.085 |
| 8 | PR | 1088/1088 | 16.267 | 66.883 | +1.50% | 316.766 |
| 16 | main | 1088/1088 | 10.691 | 101.765 | — | 481.968 |
| 16 | PR | 1088/1088 | 11.061 | 98.360 | -3.35% | 465.844 |
| 32 | main | 1088/1088 | 7.858 | 138.458 | — | 655.753 |
| 32 | PR | 1088/1088 | 7.714 | 141.036 | +1.86% | 667.961 |
| 64 | main | 1088/1088 | 9.925 | 109.620 | — | 519.171 |
| 64 | PR | 1088/1088 | 9.733 | 111.780 | +1.97% | 529.403 |

| C | Revision | Mean latency (s) | P50 (s) | P95 (s) | Mean RTF | P95 RTF | Corpus WER | Max WER |
| 1 | main | 0.0764 | 0.0760 | 0.0919 | 0.01658 | 0.0212 | 0.0112 | 0.2857 |
| 1 | PR | 0.0774 | 0.0767 | 0.0928 | 0.01679 | 0.0213 | 0.0112 | 0.2857 |
| 2 | main | 0.0877 | 0.0875 | 0.1090 | 0.01908 | 0.0251 | 0.0112 | 0.2857 |
| 2 | PR | 0.0844 | 0.0845 | 0.1033 | 0.01833 | 0.0236 | 0.0112 | 0.2857 |
| 4 | main | 0.01013 | 0.0995 | 0.1288 | 0.02206 | 0.0298 | 0.0112 | 0.2857 |
| 4 | PR | 0.0992 | 0.0981 | 0.1233 | 0.02160 | 0.0290 | 0.0112 | 0.2857 |
| 8 | main | 0.1211 | 0.1176 | 0.1585 | 0.02637 | 0.0366 | 0.0112 | 0.2857 |
| 8 | PR | 0.1192 | 0.1166 | 0.1567 | 0.02594 | 0.0361 | 0.0112 | 0.2857 |
| 16 | main | 0.1565 | 0.1560 | 0.2032 | 0.03402 | 0.0472 | 0.0112 | 0.2857 |
| 16 | PR | 0.1617 | 0.1653 | 0.2005 | 0.03509 | 0.0476 | 0.0112 | 0.2857 |
| 32 | main | 0.2292 | 0.2207 | 0.3369 | 0.04984 | 0.0717 | 0.0112 | 0.2857 |
| 32 | PR | 0.2250 | 0.2175 | 0.3293 | 0.04892 | 0.0727 | 0.0112 | 0.2857 |
| 64 | main | 0.5751 | 0.5802 | 0.6874 | 0.12529 | 0.1658 | 0.0112 | 0.2857 |
| 64 | PR | 0.5626 | 0.5706 | 0.6687 | 0.12262 | 0.1632 | 0.0112 | 0.2857 |

SandyLuXY (approved):
lgtm

PR #1412, opened 2026-08-08
[Whisper ASR] Add Bucketed Encoder CUDA Graphs and Atomic Prefill

Dayuxiaoshui:
Related: [#1396](https://github.com/sgl-project/sglang-omni/issues/1396) 

## Motivation

Whisper's decoder already uses SGLang CUDA Graph and `torch.compile`, while its fixed-shape encoder still runs eagerly. Encoder kernel-launch overhead is visible on short audio and becomes more important as request concurrency increases.

Concurrent Whisper requests also exposed a request-lifecycle failure. Each request contains an indivisible 1500-token encoder prefix, but `chunked_prefill_size=4096` could admit two complete requests and only 1088 tokens from a third request. SGLang's encoder-decoder preparation then removed three complete encoder prefixes from the 4096-token batch, producing a negative `extend_num_tokens` value (`-404`) and crashing the scheduler.

This PR adds an opt-in bucketed encoder CUDA Graph path and makes Whisper prefill admission preserve the encoder-prefix boundary.

## Modifications

- Added `WhisperEncoderCudaGraphRunner` with configurable request-batch buckets, fixed input buffers, zero padding to the smallest fitting bucket, independent graph-owned buffers, and cloned outputs.
- Added eager fallback for unsupported feature shapes, batches larger than every captured bucket, insufficient VRAM, and per-bucket capture failures.
- Added one-time startup and first-replay logs so benchmark and serving logs prove which encoder graph buckets were captured and executed.
- Added `enable_encoder_cuda_graph` and `encoder_graph_batch_buckets` stage configuration. Encoder graphs are captured after SGLang's generation graphs.
- Disabled chunked prefill for Whisper and rejected positive `chunked_prefill_size` overrides because the encoder prefix must be admitted atomically. Requests that exceed the current prefill budget wait for a later batch.
- Added a real-checkpoint component benchmark that reports capture coverage, tensor shape, dtype, device, exact equality, CUDA-event latency, and speedup.
- Added unit coverage for configuration defaults, lifecycle ordering, bucket normalization, padding, eager fallback, output ownership, replay logging, and atomic-prefill enforcement.
- Updated the Whisper cookbook with configuration, reproduction commands, benchmark results, and the atomic-prefill constraint.

## Related Issues

- ASR optimization roadmap: [#1396](https://github.com/sgl-project/sglang-omni/issues/1396)

## Accuracy Test

The component comparison loaded the real `openai/whisper-base` FP16 encoder weights into SGLang-Omni's Whisper encoder. Eager and CUDA Graph outputs were exactly equal at batch sizes 1, 2, 4, and 8; maximum and mean absolute error were both `0.0`. Input shape was `[B, 80, 3000]`, output shape was `[B, 1500, 512]`, and output dtype/device remained `torch.float16`/CUDA.

The end-to-end comparison used the same 20 SeedTTS EN samples for eager and CUDA Graph modes. Every concurrency level evaluated 60 measured requests, and corpus WER remained `0.0415` in both modes:

| Concurrency | Eager evaluated | CUDA Graph evaluated | Eager corpus WER | CUDA Graph corpus WER |
|---:|---:|---:|---:|---:|
| 1 | 60/60 | 60/60 | 0.0415 | 0.0415 |
| 2 | 60/60 | 60/60 | 0.0415 | 0.0415 |
| 4 | 60/60 | 60/60 | 0.0415 | 0.0415 |
| 8 | 60/60 | 60/60 | 0.0415 | 0.0415 |

A synchronized local-fixture burst also verified identical eager/graph transcripts for `query_to_cars.wav`, `query_to_draw.wav`, and `cough.wav` through concurrency 8. The previously crashing eager and graph paths both completed c=1/2/4/8 without scheduler failure.

## Benchmark & Profiling

Hardware and model: one NVIDIA H200, `openai/whisper-base`, FP16. The component benchmark used 10 warmup iterations and 50 measured CUDA-event iterations per batch size.

```bash
CUDA_VISIBLE_DEVICES=0 python -m benchmarks.eval.benchmark_whisper_encoder \
  --model-path openai/whisper-base \
  --batch-sizes 1,2,4,8 \
  --warmup 10 --iterations 50 \
  --output whisper_encoder_graph.json
```

| Batch size | Eager mean (ms) | CUDA Graph mean (ms) | Speedup | Exact equality |
|---:|---:|---:|---:|---:|
| 1 | 2.108 | 0.706 | 2.98x | yes |
| 2 | 2.119 | 1.085 | 1.95x | yes |
| 4 | 2.210 | 1.774 | 1.25x | yes |
| 8 | 3.431 | 3.230 | 1.06x | yes |

The end-to-end benchmark used 20 SeedTTS EN samples, one discarded warmup, and three measured repeats per concurrency and mode:

```bash
  --port 8000 --model-path openai/whisper-base \
  --max-samples 20 --concurrencies 1,2,4,8 \
  --repeats 3 --warmup --output whisper_concurrency.json
```

| Concurrency | Eager req/s | CUDA Graph req/s | Throughput gain | Eager mean latency (s) | CUDA Graph mean latency (s) | Latency reduction |
| 1 | 19.57 | 20.29 | 3.7% | 0.051 | 0.049 | 3.9% |
| 2 | 28.41 | 30.87 | 8.7% | 0.070 | 0.065 | 7.1% |
| 4 | 37.90 | 41.70 | 10.0% | 0.104 | 0.094 | 9.6% |
| 8 | 42.10 | 49.00 | 16.4% | 0.185 | 0.158 | 14.6% |

All 480 measured end-to-end requests completed successfully. Serving logs showed encoder graph captures for buckets 1, 2, 4, and 8 and first replays for the batch-1 and batch-2 paths used under the configured 4096-token prefill budget.

## Checklist

- [x] Add unit tests.

## CI

Whisper currently has no dedicated CI selector, so this PR relies on the focused Whisper unit tests plus the H200 component and end-to-end validation above.


luojiaxuan (approved):
No blocking issue in the encoder graph or atomic-prefill changes.

sglang_omni/models/whisper_asr/engine_builder.py
```diff
@@ -7,6 +7,14 @@
 
 from sglang_omni.scheduling.engine_factory import AsrEngineBuilder
 
+_DEFAULT_ENCODER_GRAPH_BATCH_BUCKETS = (1, 2, 4, 8, 12, 16)
```
SandyLuXY, line 13: With `max_prefill_tokens=4096` and a 1,500-token encoder prefix, non-chunked admission can encode at most two uncached Whisper requests, so buckets 4/8/12/16 cannot replay and only consume capture time/VRAM (noticed that bucket 16 cost 1,350 MiB on large-v3-turbo). Derive or filter the default buckets from the final prefill budget and encoder prefix length, and test the resolved serving-reachable set.

SandyLuXY (commented):
Nice work! Please turn on the encoder cuda graph by default as it improves perf.

SandyLuXY (approved):
lgtm

benchmarks/eval/benchmark_whisper_encoder.py
SandyLuXY, line 1: Why do we need a separate whisper encoder under benchmarks?

SandyLuXY (changes requested):

SandyLuXY (approved):

PR #1415, opened 2026-08-08
[CI] Cpuset contention: report it, fail the stage, guard calibration rounds

JiaxinD:
Implements the mechanism agreed in the group discussion (report alone is not enough; intrusion should fail CI, and calibration should wait and re-run the affected round). Built on one sampler, applied in three layers:

1. **Report (all pinned sessions)**: a daemon sampler (30s interval) walks /proc and charges CPU time consumed on the pinned cores by processes outside the session tree. Session end prints one greppable line:
```
[cpuset-contention] cpuset=48-63,112-127 windows=42 foreign-cores mean=0.03 max=0.85 errors=0
```
2. **CI fails the stage**: on GitHub Actions, a foreign peak above 2 cores fails the session in fixture teardown, after test results are written. The stage's existing 3-attempt retry then re-measures on a clean window instead of gating on polluted numbers.
3. **Calibration waits and re-runs the round**: tune.py now waits before each launch while the cpuset carries foreign load (30s recheck, 15 min cap), and a round whose in-session peak crossed the limit is discarded and re-run after the cpuset clears, without touching completed rounds. A new Tab C watcher (`watch_calibration_cpuset.sh`) gives the operator a live view of the reserved cores.

No behavior change when `OMNI_CI_CPUSET` is unset. Sampling is read-only /proc traversal every 30s, well under 0.1 percent of one core, from the pytest process (the measured servers live in separate processes).

Unit tests cover cpu-list parsing, the process-tree walk, foreign-tick accounting, and a Linux sampler smoke; the contention-peak log parser is exercised against real summary lines.

### DIFF
--- a/tests/utils/ci_cpu_contention.py
+++ b/tests/utils/ci_cpu_contention.py
@@ -1,0 +1,147 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Report foreign CPU load on the pinned OMNI_CI_CPUSET during a session.
+
+Affinity is self-restraint, not a reservation: an unpinned process can still
+be scheduled onto the reserved cores and inflate what a perf gate measures.
+This sampler makes that attributable: each interval it charges CPU time
+consumed on the pinned cores by processes outside the session tree, and the
+fixture prints a summary so a failed speed gate can be triaged from the log.
+"""
+
+from __future__ import annotations
+
+import os
+import threading
+import time
+from dataclasses import dataclass
+
+_SAMPLE_INTERVAL_S = 30.0
+_WARN_FOREIGN_CORES = 1.0
+# note (Jiaxin Deng): above this, the session's speed numbers describe the
+# intruder, not the model; CI fails the stage so its retry re-measures.
+FAIL_FOREIGN_CORES = 2.0
+
+
+@dataclass
+class _Proc:
+    ppid: int
+    ticks: int
+    overlaps: bool
+
+
+def _parse_cpu_list(spec: str) -> set[int]:
+    cpus: set[int] = set()
+    for part in spec.split(","):
+        part = part.strip()
+        if not part:
+            continue
+        lo_text, _, hi_text = part.partition("-")
+        cpus.update(range(int(lo_text), int(hi_text or lo_text) + 1))
+    return cpus
+
+
+def _snapshot(cpuset: set[int]) -> dict[int, _Proc]:
+    procs: dict[int, _Proc] = {}
+    for entry in os.listdir("/proc"):
+        if not entry.isdigit():
+            continue
+        pid = int(entry)
+        try:
+            with open(f"/proc/{pid}/stat", "rb") as f:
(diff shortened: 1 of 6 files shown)


tests/utils/ci_cpu_contention.py
```diff
@@ -0,0 +1,147 @@
...
+            ticks = int(rest[11]) + int(rest[12])
+            allowed = ""
+            with open(f"/proc/{pid}/status", encoding="ascii", errors="replace") as f:
+                for line in f:
+                    if line.startswith("Cpus_allowed_list"):
+                        allowed = line.split(":", 1)[1].strip()
+                        break
+            # note (Jiaxin Deng): an unreadable mask counts as overlapping;
+            # a false positive beats an invisible intruder in this report.
+            overlaps = bool(cpuset & _parse_cpu_list(allowed)) if allowed else True
```
AkazaAkane, line 63: Cpus_allowed_list only tells us where a process may run, but ticks is the process's total CPU time across all CPUs. For an unpinned process with Cpus_allowed_list=0-127, we count its entire tick delta whenever its allowed mask overlaps our lane, even if it actually spent the whole interval on CPUs outside OMNI_CI_CPUSET.

For example, a compile job using 8 cores on another lane could be reported as ~8 foreign cores here simply because its affinity allows all CPUs. That could invalidate an otherwise clean perf run.

Can we either measure actual execution on the target CPUs, or add a controlled test showing that load pinned outside the target cpuset does not contribute while the same load pinned inside it does?

tests/utils/ci_cpu_contention.py
```diff
@@ -0,0 +1,147 @@
...
+                tree.add(child)
+                queue.append(child)
+    return tree
+
+
+def foreign_ticks(prev: dict[int, _Proc], cur: dict[int, _Proc], tree: set[int]) -> int:
+    return sum(
+        cur[pid].ticks - prev[pid].ticks
+        for pid in cur
+        if pid in prev and cur[pid].overlaps and pid not in tree
```
AkazaAkane, line 87: We only charge a PID when it exists in both prev and cur, with snapshots 30s apart.

A foreign process that starts and finishes between two snapshots is invisible. A process that starts during the interval and is still alive at cur also contributes nothing until the next window. So a 5–20s CPU burst can interfere with latency/throughput without ever appearing in foreign_ticks().

zhaochenyang20 (approved):

tests/utils/ci_cpu_contention.py
JiaxinD: Both real, thanks. Fixed in #1423 by measuring from the per-CPU busy counters minus the pinned tree's ticks, with the controlled inside-vs-outside test you asked for.

tests/utils/ci_cpu_contention.py
JiaxinD: Fixed in #1423: the per-CPU counters survive process exit, so short-lived bursts stay charged.

PR #1420, opened 2026-08-08
[dots.tts] Fix non-deterministic reference encoding; add parity-safe batched reference encode

buffett0323:
## Motivation
`dots.tts` reference encoding is non-deterministic. `SpeakerXVectorFeatures._crop_audio` (upstream `dots_tts` v0.2.1, `modules/speaker/encoder.py:82`) picks its window with the global `random` module and has no `self.training` guard, so it runs at inference:

``` start = random.randint(0, total_length - cropped_length)```

Any reference longer than `max_audio_seconds` (10 s on `dots-studio/dots.tts-mf`) gets a random crop, so voice cloning is not reproducible.

Two things hide this: the reference cache pins whichever crop was drawn first, and the server seeds the global RNG at startup, so a restart redraws the same window. What varies in production is the RNG position, each long reference consumes one `randint`, so the crop you get depends on how many other long references were encoded before yours, i.e. on request ordering and load.

This also blocks batching parity: the crop loops per item from the shared RNG stream, and grouping reorders items, so batched ≠ unbatched for any reference over 10 s.

## Modifications
- codec.py : `_speaker_input()` / `_speaker_sample_limit()` truncate to the leading `max_audio_seconds` before the speaker encoder, exactly what upstream emits when its draw returns 0. AudioVAE still receives full-length audio.
- codec.py : `encode_reference_batch` groups waveforms by exact sample length. Padding was tried and rejected on measurement: CAM++ masks its pooling, but the TDNN stack between fbank and pooling is unmasked and pulls zeros into the valid region, padded batching moved the speaker embedding by 1.386.
- stages.py  /  reference_encoder.py : plumb batch settings.
- dots_tts.yaml : ships disabled (`max_batch_size: 1`)

## Related Issues

Part of #1367 (reference encoder work items).

### Encode-Order Sensitivity
*Identical reference, process, and cache; only the number of other long references encoded first varies.*

| Preceding Long Encodes | Max Abs Diff (Before) | Cosine (Before) | After |
| :--- | :--- | :--- | :--- |
| 1 | 0.570 | 0.9833 | 0.0 |
| 2 | 0.529 | 0.9832 | 0.0 |
| 3 | 0.185 | 0.9976 | 0.0 |

### Repeated Encodes of the Same Input

| Case | Before | After |
| :--- | :--- | :--- |
| reference > 10 s, max abs diff | 0.738 | 0.0 |
| reference > 10 s, min cosine | 0.9739 | 1.0 |
| reference < 10 s (control) | 0.0 | 0.0 |
| global RNG seeded (control) | 0.0 | 0.0 |

* **Impact:** `0.738` represents **19.8%** of the distance between two completely different speakers (`3.72`). 
* **Root Cause:** The two control cases pin the issue directly on unseeded `random.randint` usage rather than standard numerical/floating-point noise.

### Batching Parity (H200)
* Speaker embedding is **bit-exact** with TF32 disabled (latents: `1.7e-5`).
* The `~6e-4` residual observed with TF32 enabled is due to **cuDNN tiling** effects (identical rows within the exact same batch also differ). 
* Therefore, parity is now successfully gated both ways.
## Checklist

- [x] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/models/dots_tts/codec.py
+++ b/sglang_omni/models/dots_tts/codec.py
@@ -235,7 +235,11 @@ class _DotsReferenceHook(KeyedReferenceEncodeHook[str, dict, dict]):
         return self.codec.encode_reference(item)
 
     def can_encode_batch(self) -> bool:
-        return True
+        if self.codec.device.type != "cuda":
+            return True
+        return not (
+            torch.backends.cuda.matmul.allow_tf32 or torch.backends.cudnn.allow_tf32
+        )
 
     def encode_batch(self, items: list[str]) -> list[dict]:
         return self.codec.encode_reference_batch(list(items))
--- a/sglang_omni/models/dots_tts/codec.py
+++ b/sglang_omni/models/dots_tts/codec.py
@@ -4,8 +4,10 @@
 from __future__ import annotations
 
 import json
+import logging
 import math
 import threading
+from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 
 import torch
@@ -27,6 +29,8 @@ from sglang_omni.scheduling.reference_encoder import (
 )
 from sglang_omni.utils.audio import load_audio
 from sglang_omni.utils.checkpoint import resolve_checkpoint
+
+logger = logging.getLogger(__name__)
 
 
 def _load_module(module: torch.nn.Module, path: Path) -> None:
@@ -67,8 +71,11 @@ class DotsAudioCodec:
         self.device = torch.device(device)
         self.lock = threading.RLock()
 
-    @torch.inference_mode()
-    def encode_reference(self, path: str) -> dict[str, torch.Tensor]:
+    @staticmethod
+    def _reference_load_workers(count: int) -> int:
+        return max(1, min(int(count), 8))
+
+    def _load_reference_waveform(self, path: str) -> torch.Tensor:
         waveform = load_audio(
             path,
             source_name="dots.tts reference",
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by buffett0323** [comment]
Ready for review and Lint and Test Layout pass locally.
FYI, @Hayden727 , thanks for pointing out the issue and would also appreciate your help for reviewing!

File: sglang_omni/models/dots_tts/codec.py [hunk 1]
```diff
@@ -84,14 +91,103 @@ def encode_reference(self, path: str) -> dict[str, torch.Tensor]:
...
+
         with self.lock:
-            speaker = self.speaker(audio.unsqueeze(0))
-            latent_distribution = self.inference.extract_latents(audio.unsqueeze(0))
-        return {
-            "speaker_embedding": speaker.detach().cpu().float(),
-            "latent_distribution": latent_distribution.detach().cpu().float(),
-        }
+            speaker = self.speaker(speaker_batch, audio_lengths=speaker_lengths)
+            latent_distribution = self.inference.extract_latents(batch)
```
Hayden727, line 118: This batched forward is not parity-safe with the runtime defaults (`cuda.matmul.allow_tf32=False`, `cudnn.allow_tf32=True`): on both H200s, two equal-length shipped speech references produced single-vs-batch max diffs of 0.0079 in the speaker embedding and 0.0152 in latents. Disable the non-parity backend here or keep batching unavailable until a real-model regression covers the accepted tolerance.

sglang_omni/models/dots_tts/stages.py
```diff
@@ -368,11 +368,18 @@ def create_reference_encode_executor(
     *,
     device: str | None = "cuda",
     gpu_id: int | None = None,
-    max_concurrency: int = 1,
+    max_concurrency: int = 8,
+    max_batch_size: int = 8,
```
Hayden727: Keep this default at 1. `ConfigManager.from_model_path()` creates `DotsTTSPipelineConfig` without the example YAML overrides, so standard model-path launches currently enable batching at 8 even though this PR says it ships disabled.

sglang_omni/scheduling/reference_encoder.py
```diff
@@ -151,21 +161,150 @@ def __init__(
...
+        ) = None
+        self._batch_thread: threading.Thread | None = None
+        if self._batching:
+            self._batch_queue = _queue_mod.Queue()
+            self._batch_thread = threading.Thread(
+                target=self._batch_worker,
+                name=batch_worker_name,
+                daemon=True,
+            )
+            self._batch_thread.start()
```
Hayden727, line 185: Wire this worker into the stage scheduler shutdown lifecycle. `create_reference_encode_executor()` returns `SimpleScheduler(encoder.encode_payload, ...)`, but `SimpleScheduler.stop()` never calls `encoder.close()`, so this thread remains alive after stage stop; add an owned shutdown path and a production-level regression.

Hayden727 (approved):

PR #1432, opened 2026-08-09
[Perf] Keep Qwen3-ASR prefill coalescing during active decode

luojiaxuan:
## Motivation

Qwen3-ASR uses pending request-build work to decide whether a small prefill should wait for the existing 16-request / 24 ms coalescing window. Once request construction drains, small prefills are released immediately even while a decode batch is active. H200 profiling showed that these scattered admissions repeatedly interrupt decode and leave large gaps between CUDA graph launches at high concurrency.

## Modifications

1. Add an opt-in scheduler policy that keeps coalescing after request-build work drains only while decode is active.
2. Keep the existing pending-build-aware early release while decode is idle, preserving light-load behavior.
3. Enable the policy only for Qwen3-ASR and thread it through the pipeline config, stage factory, and engine builder.
4. Add behavior tests for both active-decode coalescing and idle release.

Other model defaults are unchanged.

## Related Issues

Related to #1324 and #1396.

## Accuracy Test

- Model: `Qwen/Qwen3-ASR-1.7B`, revision `7278e1e70fe206f11671096ffdd38061171dd6e5`.
- Dataset: full SeedTTS EN, 1,088 clips, revision `27f4c1adee83b5b29b7c4b375f6b976324bda308`.
- Every measured configuration completed 3,264 of 3,264 requests with zero skips.
- Corpus WER was identical at `0.0121822034` for main and candidate.

## Benchmark & Profiling

Setup:

- Hardware: one NVIDIA H200 on hyper00, physical GPU 0.
- Runtime: BF16, FA3 attention, decode CUDA graphs enabled, torch.compile disabled.
- Baseline: current main at `efef31d612c4d9c101bbd62e4dadae960df1b05e`.
- Candidate: `6a4df39a245576f4a29aa00828511003d0db589c`.
- Protocol: same physical GPU, sequential fresh servers, one discarded full-corpus warmup followed by three measured full-corpus repeats per concurrency.

| Concurrency | Metric | Main | Candidate | Delta |
| 8 | Throughput | 96.994 req/s | 100.503 req/s | +3.62% |
| 8 | Mean latency | 82.128 ms | 79.194 ms | -3.57% |
| 8 | P95 latency | 121.166 ms | 117.236 ms | -3.24% |
| 64 | Throughput | 211.353 req/s | 254.768 req/s | +20.54% |
| 64 | Mean latency | 297.918 ms | 247.116 ms | -17.05% |
| 64 | P95 latency | 443.610 ms | 350.229 ms | -21.05% |

The candidate reaches 254.768 req/s at c64, 7.66% above the fixed 236.639 req/s serving baseline. A separate c2 validation completed all 3,264 requests at 21.636 req/s with the same WER.

## Validation

- `pytest -q tests/unit_test/qwen3_asr tests/unit_test/scheduling/test_prefill_coalesce_gate.py`: 550 passed.
- Pre-commit passed for all six changed files.
- Git diff check passed.

## Checklist

- [x] Add unit tests.
- [x] Provide reproducible throughput, latency, and accuracy measurements.

### DIFF
--- a/sglang_omni/scheduling/omni_scheduler.py
+++ b/sglang_omni/scheduling/omni_scheduler.py
@@ -196,6 +196,7 @@ class OmniScheduler:
         prefill_coalesce_wait_ms: float = 60.0,
         prefill_coalesce_when_idle: bool = False,
         prefill_coalesce_requires_pending_builds: bool = False,
+        prefill_coalesce_after_builds_during_decode: bool = False,
         request_build_max_workers: int = 1,
         request_build_max_pending: int | None = None,
         shutdown_callback: Callable[[], None] | None = None,
@@ -301,6 +302,9 @@ class OmniScheduler:
         self.prefill_coalesce_when_idle = bool(prefill_coalesce_when_idle)
         self.prefill_coalesce_requires_pending_builds = bool(
             prefill_coalesce_requires_pending_builds
+        )
+        self.prefill_coalesce_after_builds_during_decode = bool(
+            prefill_coalesce_after_builds_during_decode
         )
 
         # Token / memory info (upstream reads from tp_worker.get_worker_info)
@@ -1162,9 +1166,8 @@ class OmniScheduler:
         # so the coalesce hold-off returns an empty plan rather than None.
         if self.prefill_coalesce_requests <= 1 or self.chunked_req is not None:
             return _Upstream.get_new_batch_prefill(self, running_batch)
-        if not self.prefill_coalesce_when_idle and (
-            running_batch is None or running_batch.is_empty()
-        ):
+        decode_is_idle = running_batch is None or running_batch.is_empty()
+        if not self.prefill_coalesce_when_idle and decode_is_idle:
             return _Upstream.get_new_batch_prefill(self, running_batch)
         if self.prefill_coalesce_requires_pending_builds:
             with self._request_admission_lock:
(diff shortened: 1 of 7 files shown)


sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -1162,17 +1166,18 @@ def get_new_batch_prefill(self, running_batch):
...
             return _Upstream.get_new_batch_prefill(self, running_batch)
         if self.prefill_coalesce_requires_pending_builds:
             with self._request_admission_lock:
                 build_work_pending = bool(
                     self._pending_request_builds
                     or self._backlogged_request_build_payloads
                 )
-            if not build_work_pending:
+            if not build_work_pending and not (
+                self.prefill_coalesce_after_builds_during_decode and not decode_is_idle
```
SandyLuXY, line 1179: The Qwen3-ASR cookbook still says prefill releases “immediately when build work drains,” but this branch now keeps a small prefill waiting whenever the running decode batch is non-empty. Please update `docs/cookbook/qwen3_asr.md` to document that the immediate build-drain release applies only while decode is idle, and that active decode still waits for the request target or oldest-request deadline.

SandyLuXY (changes requested):

SandyLuXY (approved):
lgtm

sglang_omni/scheduling/omni_scheduler.py
luojiaxuan: Done, the cookbook now limits immediate build-drain release to idle decode and documents target or deadline coalescing during active decode.

PR #1433, opened 2026-08-09
[Perf][A-PR7]  Add Bucketed Encoder CUDA Graphs

wirybeaver:
## Summary
- Add opt-in bucketed CUDA graphs for the ARK-ASR audio encoder path.
- Capture/replay the full `audio_encoder` (Whisper/RoPE tower + MLP adapter) using model-internal `(batch, mel_frames)` buckets.
- Keep eager fallback for unsupported shapes, low free VRAM, capture/replay failures, and previously failed buckets.
- Expose only `enable_encoder_cuda_graph`; bucket sizing and memory heuristics remain private implementation details.

## Configuration

- `enable_encoder_cuda_graph` (default: `False`)

## Verification

- `uv run pytest -q tests/unit_test/arkasr`
- `uv run python -m compileall -q sglang_omni/models/arkasr tests/unit_test/arkasr`
- `uvx black --check sglang_omni/models/arkasr/config.py sglang_omni/models/arkasr/encoder_cuda_graph.py sglang_omni/models/arkasr/encoder_service.py sglang_omni/models/arkasr/engine_builder.py sglang_omni/models/arkasr/sglang_model.py sglang_omni/models/arkasr/stages.py tests/unit_test/arkasr/test_encoder_cuda_graph.py tests/unit_test/arkasr/test_pipeline.py`
- CI Docker image mounted repo: `hongccc/sglang-omni@sha256:374d0b1c30b2bff685b1716fc64a02ad3b3d0a90fe2ce73ce9861a6992c28101`, `python -m pytest -q tests/unit_test/arkasr` → `64 passed, 1 skipped`
- Remote H100 direct-container smoke with pushed image → `python -m pytest -q tests/unit_test/arkasr` → `65 passed`

## Docker / benchmark images

Latest production-default cache-enabled benchmark:

- Base: `hongccc/sglang-omni@sha256:374d0b1c30b2bff685b1716fc64a02ad3b3d0a90fe2ce73ce9861a6992c28101`
- Benchmark image: `rfgleamchaser/sglang-omni:arkasr-cache-default-5e51a053-v2`
- Benchmark image digest: `sha256:9948bb753d82d6d26af3dd40e28bcaf9abdc8dafb9a36a034f4d962db57f8821`
- Source commit in image: `5e51a0535e2c3a646abc16b3a6c25739a0a2e5dc`

Earlier encoder-focused no-cache benchmark:

- Benchmark image: `rfgleamchaser/sglang-omni:arkasr-cuda-graph-d18aa4936845-whisper`
- Benchmark image digest: `sha256:7602287e36243626de40002eae2ba1a410cf8ecf5d6aaacf6a4d51225d886a4c`
- Source commit in image: `d18aa4936845809bce405c9328914382113a4a9d`

## Latest production-default cache-enabled Full SeedTTS EN benchmark

Run on August 23, 2026 using one H100 80GB HBM3 (`driver 535.309.01`) with the direct-container workflow. Dataset: Full SeedTTS EN (`1088` clips), revision `27f4c1adee83b5b29b7c4b375f6b976324bda308`.

All serving settings remained at ARK-ASR defaults, including the enabled pre-LM cache (`4096` entries / `2 GiB`), `pre_lm_max_batch_size=8`, `pre_lm_max_batch_wait_ms=0`, `pre_lm_max_pending=32`, `encoder_max_batch_size=8`, and automatic `mem_fraction_static`. The only serving setting changed was `enable_encoder_cuda_graph=false/true`.

For each concurrency, both execution orders were measured with a fresh server for every switch: `false -> true` and `true -> false` (`16` server sessions total). Each session ran one initially empty-cache pass followed by three warm-cache passes on the same process. The cold table therefore aggregates two observations per mode; the warm table aggregates six observations per mode. Negative p95/p99 deltas mean lower latency.

### Initially empty cache

The cache begins empty, but SeedTTS contains repeated reference audio, so intra-pass cache hits accumulate during this first pass.

| concurrency | eager req/s | graph req/s | throughput delta | eager mean latency | graph mean latency | mean latency delta | eager P95 latency | graph P95 latency | P95 latency delta | eager P99 latency | graph P99 latency | P99 latency delta | eager WER | graph WER | WER abs delta |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 10.766 | 12.289 | **+14.14%** | 0.093s | 0.081s | **-12.43%** | 0.136s | 0.099s | **-27.19%** | 0.151s | 0.115s | **-23.59%** | 0.011199 | 0.011123 | -0.000076 |
| 8 | 40.284 | 54.245 | **+34.66%** | 0.198s | 0.147s | **-25.91%** | 0.271s | 0.217s | **-19.93%** | 0.580s | 0.524s | **-9.67%** | 0.011274 | 0.011350 | +0.000076 |
| 16 | 68.840 | 83.235 | **+20.91%** | 0.231s | 0.191s | **-17.33%** | 0.304s | 0.304s | -0.18% | 0.596s | 0.724s | **+21.52%** | 0.011161 | 0.011388 | +0.000227 |
| 32 | 113.865 | 113.347 | -0.45% | 0.278s | 0.280s | +0.75% | 0.380s | 0.510s | **+34.10%** | 0.623s | 0.787s | **+26.34%** | 0.011350 | 0.011388 | +0.000038 |

The order-balanced result is consistent in both directions: CUDA Graph improves initially-empty-cache throughput at concurrency `1/8/16`, peaks at `+34.66%` for concurrency `8`, and no longer improves throughput at concurrency `32`. At `32`, p95/p99 regress because the serving pipeline is already saturated.

### Warm cache

| concurrency | eager req/s | graph req/s | throughput delta | eager mean latency | graph mean latency | mean latency delta | eager P95 latency | graph P95 latency | P95 latency delta | eager P99 latency | graph P99 latency | P99 latency delta | eager WER | graph WER | WER abs delta |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 13.563 | 13.510 | -0.39% | 0.073s | 0.074s | +0.39% | 0.089s | 0.090s | +0.27% | 0.102s | 0.103s | +0.64% | 0.011199 | 0.011123 | -0.000076 |
| 8 | 59.433 | 60.149 | +1.20% | 0.134s | 0.132s | -1.22% | 0.175s | 0.174s | -0.55% | 0.199s | 0.203s | +2.13% | 0.011274 | 0.011350 | +0.000076 |
| 16 | 92.748 | 92.915 | +0.18% | 0.172s | 0.171s | -0.21% | 0.230s | 0.229s | -0.42% | 0.264s | 0.267s | +1.35% | 0.011085 | 0.011350 | +0.000265 |
| 32 | 124.315 | 126.172 | +1.49% | 0.256s | 0.252s | -1.56% | 0.348s | 0.340s | -2.21% | 0.439s | 0.427s | -2.75% | 0.011274 | 0.011350 | +0.000076 |

Once the embedding cache is populated, encoder CUDA Graph is mostly bypassed and throughput changes stay between `-0.39%` and `+1.49%`, as expected.

All `1088` requests completed in every repeat with zero skips. All eight graph-enabled sessions captured encoder graph buckets. There were no OOMs, HTTP 5xx responses, tracebacks, graph capture/replay failures, or leaked GPU processes. The initially-empty-cache order-paired throughput deltas were `+14.56%/+13.73%` at concurrency 1, `+38.17%/+31.35%` at 8, `+19.76%/+22.03%` at 16, and `-0.64%/-0.26%` at 32 for order 1/order 2 respectively.

WER stayed within `0.011085-0.011388`, but was not byte-identical. A few normalized transcripts also varied between repeated runs of the same mode under the default concurrent/async serving path, so the small WER differences cannot be attributed solely to encoder CUDA Graph.

## Earlier encoder-focused no-cache Full SeedTTS EN benchmark

H100 80GB direct-container, driver `595.71.05`, SeedTTS EN full (`1088` clips, dataset revision `27f4c1adee83b5b29b7c4b375f6b976324bda308`), repeats=`3`, warmup=`true`, `mem_fraction_static=0.75`.

This is the encoder-focused no-cache run: `pre_lm_cache_max_entries=0`, `pre_lm_cache_size_bytes=0`, `pre_lm_max_batch_wait_ms=1`. The cache-disabled setup is intentional because a normal cache-enabled serving run mostly bypasses encoder work after warmup and hides the graph signal.

The CUDA Graph run captured 12 encoder buckets (`B=1/2/4/8`, frames `512/768/1024`) with no low-VRAM fallback, no capture/replay failure, and no `ERROR`/`Traceback` in the server log. WER is comparable but not byte-identical after rounding (`0.0112` eager vs `0.0113` graph).

| concurrency | eager req/s | graph req/s | delta | eager mean latency | graph mean latency | eager P95 latency | graph P95 latency | eager WER | graph WER |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 8.548 | 9.864 | +15.4% | 0.117s | 0.101s | 0.136s | 0.120s | 0.0112 | 0.0113 |
| 2 | 12.927 | 18.928 | +46.4% | 0.155s | 0.105s | 0.210s | 0.132s | 0.0112 | 0.0113 |
| 4 | 19.061 | 30.464 | +59.8% | 0.209s | 0.131s | 0.305s | 0.173s | 0.0112 | 0.0113 |
| 8 | 31.142 | 46.779 | +50.2% | 0.256s | 0.170s | 0.351s | 0.233s | 0.0112 | 0.0113 |
| 16 | 37.684 | 59.417 | +57.7% | 0.422s | 0.268s | 0.556s | 0.363s | 0.0112 | 0.0113 |
| 32 | 37.365 | 75.389 | +101.8% | 0.847s | 0.425s | 1.054s | 0.582s | 0.0112 | 0.0113 |

## SeedTTS-50 benchmark (WER-identical smoke)

H100 80GB direct-container, SeedTTS-50 EN (`50` clips), repeats=`1`, warmup=`true`, `mem_fraction_static=0.75`, same no-cache encoder-focused config as above (`pre_lm_cache_max_entries=0`, `pre_lm_cache_size_bytes=0`, `pre_lm_max_batch_wait_ms=1`). This run captured 8 encoder buckets (`B=1/2/4/8`, frames `512/768`) with no capture/replay failures. WER is exactly unchanged after rounding (`0.0074` eager and graph at every concurrency).

| concurrency | eager req/s | graph req/s | delta | eager mean latency | graph mean latency | eager P95 latency | graph P95 latency | eager WER | graph WER |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 8.964 | 10.157 | +13.3% | 0.111s | 0.098s | 0.147s | 0.133s | 0.0074 | 0.0074 |
| 2 | 9.592 | 15.046 | +56.9% | 0.207s | 0.132s | 0.272s | 0.174s | 0.0074 | 0.0074 |
| 4 | 15.629 | 24.844 | +59.0% | 0.248s | 0.159s | 0.350s | 0.212s | 0.0074 | 0.0074 |
| 8 | 28.491 | 33.291 | +16.8% | 0.258s | 0.231s | 0.353s | 0.430s | 0.0074 | 0.0074 |
| 16 | 33.867 | 60.293 | +78.0% | 0.448s | 0.241s | 0.564s | 0.365s | 0.0074 | 0.0074 |
| 32 | 33.594 | 69.764 | +107.7% | 0.766s | 0.387s | 1.036s | 0.544s | 0.0074 | 0.0074 |

## Limitations / interpretation

- The latest table is the production-default, cache-enabled end-to-end result. The initially-empty-cache pass still gains cache hits within the pass because SeedTTS reuses reference audio.
- Encoder CUDA Graph provides a clear cold-path benefit at concurrency `1/8/16`, but not at `32`; the tail-latency regression at `32` should be considered before enabling it for saturated traffic.
- Warm-cache traffic mostly bypasses the encoder, so the measured effect is approximately neutral.
- The earlier no-cache tables remain useful as an encoder-path upper-bound signal, not as the expected gain for warmed production traffic.
- Full EN WER is close but not exactly identical. Default concurrent/async execution also produced a few transcript differences within the same mode, so a deterministic accuracy-only follow-up would need a separately controlled decode configuration.


sglang_omni/models/arkasr/encoder_cuda_graph.py
```diff
@@ -0,0 +1,286 @@
...
+            frame_bucket_step=self._frame_bucket_step,
+            max_frames=self._max_frames,
+        )
+        if batch_bucket is None or frame_bucket is None:
+            return None
+        key = (batch_bucket, frame_bucket)
+        if key in self._failed:
+            return None
+
+        with self._lock:
```
SandyLuXY, line 212: A waiter can pass the pre-lock `_failed` check, then retry the same capture/replay after the first caller marks the bucket failed. Recheck `_failed` under this lock before accessing the graph so queued callers honor the failed-bucket fallback.

sglang_omni/models/arkasr/engine_builder.py
```diff
@@ -71,6 +80,14 @@ def __init__(
         self.pre_lm_max_batch_size = pre_lm_max_batch_size
         self.pre_lm_max_batch_wait_ms = pre_lm_max_batch_wait_ms
         self.pre_lm_max_pending = pre_lm_max_pending
+        self.enable_encoder_torch_compile = enable_encoder_torch_compile
```
SandyLuXY: Consider removing the parameter before encoder torch.compile is implemented for ark-asr.

sglang_omni/models/arkasr/config.py
```diff
@@ -37,6 +37,11 @@ class ArkasrPipelineConfig(PipelineConfig):
                 "pre_lm_max_batch_size": 8,
                 "pre_lm_max_batch_wait_ms": 0,
                 "pre_lm_max_pending": 32,
+                "enable_encoder_cuda_graph": False,
```
SandyLuXY, line 44: The new public CUDA-graph control is not documented in `docs/cookbook/arkasr.md`. Please add the doc and supported/fallback behavior etc

sglang_omni/models/arkasr/config.py
```diff
@@ -37,6 +37,11 @@ class ArkasrPipelineConfig(PipelineConfig):
...
+                "encoder_graph_batch_buckets": [1, 2, 4, 8],
+                "encoder_graph_frame_bucket_step": 256,
+                "encoder_graph_max_frames": 3000,
+                "encoder_graph_min_free_gb": 3.0,
```
SandyLuXY: This exposes model-internal bucketing and memory heuristics that existing ASR integrations keep private (check Qwen3-ASR / Fun-ASR). Consider exposing only enable_encoder_cuda_graph and deriving / keeping others private

tests/unit_test/arkasr/test_encoder_cuda_graph.py
SandyLuXY, line 1: update tests/README.md

SandyLuXY (changes requested):

sglang_omni/models/arkasr/encoder_cuda_graph.py
wirybeaver: Good catch. Rechecked `_failed` under the runner lock so queued callers honor the sticky eager fallback, and added regression coverage for a bucket becoming failed while a caller waits. Fixed in 081e25ed.

sglang_omni/models/arkasr/config.py
wirybeaver: Added ARK-ASR cookbook documentation covering the opt-in configuration, lazy full-encoder capture, internal bucketing, and eager fallback behavior. Fixed in 081e25ed.

sglang_omni/models/arkasr/config.py
wirybeaver: Agreed. Only `enable_encoder_cuda_graph` remains public; batch/frame bucketing and the memory heuristic now remain private runner defaults. Fixed in 081e25ed.

tests/unit_test/arkasr/test_encoder_cuda_graph.py
wirybeaver: Updated `tests/README.md` with the new ARK-ASR CUDA graph test file and its coverage. Fixed in 081e25ed.

sglang_omni/models/arkasr/engine_builder.py
wirybeaver: Removed the ARK-ASR `enable_encoder_torch_compile` option from the stage and builder APIs since that path is not implemented. Fixed in 081e25ed.

sglang_omni/models/arkasr/encoder_cuda_graph.py
```diff
@@ -0,0 +1,325 @@
...
+        torch.cuda.current_stream(self._device).wait_stream(stream)
+        torch.cuda.synchronize(self._device)
+
+        if self._pool is None:
+            self._pool = torch.cuda.graph_pool_handle()
+        graph = torch.cuda.CUDAGraph()
+        with torch.cuda.graph(
+            graph,
+            pool=self._pool,
+            capture_error_mode="thread_local",
```
SandyLuXY: `thread_local` does not make runtime capture safe while the LM scheduler or an eager fallback is issuing CUDA work on another thread; PyTorch requires capture-time process-wide CUDA quiescence. Please capture the selected buckets before serving

SandyLuXY (changes requested):

sglang_omni/models/arkasr/encoder_cuda_graph.py
wirybeaver: ACK. I will remove all lazy capture related codes.

sglang_omni/models/arkasr/encoder_cuda_graph.py
```diff
@@ -0,0 +1,248 @@
...
+import threading
+from collections.abc import Sequence
+from dataclasses import dataclass
+
+import torch
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_BATCH_BUCKETS = (1, 2, 4, 8)
+_DEFAULT_FRAME_BUCKETS = (512, 768, 1024)
```
WilsonZheng0327, line 23: max frame = 3000 for 100 frames/sec and 30sec max. 
need more buckets to not only cover 1024-3000, but also 0-512

sglang_omni/models/arkasr/encoder_cuda_graph.py
```diff
@@ -0,0 +1,248 @@
...
+import logging
+import threading
+from collections.abc import Sequence
+from dataclasses import dataclass
+
+import torch
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_BATCH_BUCKETS = (1, 2, 4, 8)
```
WilsonZheng0327, line 22: this should be coupled with `encoder_max_batch_size`, else >8 batch size will just be eager

wirybeaver:
The new round of benchmarking is in progress

sglang_omni/models/arkasr/encoder_cuda_graph.py
```diff
@@ -0,0 +1,248 @@
...
+            except Exception as exc:
+                logger.warning(
+                    "ARK-ASR encoder CUDA graph replay failed for "
+                    "batch=%d frames=%d: %s; using eager for this bucket",
+                    batch_bucket,
+                    frame_bucket,
+                    exc,
+                )
+                self._graphs.pop(key, None)
+                return None
```
WilsonZheng0327, line 233: this try block is too wide, should just include `.replay()`

sglang_omni/models/arkasr/encoder_cuda_graph.py
```diff
@@ -0,0 +1,248 @@
...
+        self._done_event = torch.cuda.Event()
+        self._event_recorded = False
+
+    def _enough_free_vram(self) -> tuple[bool, int]:
+        if self._min_free_bytes <= 0:
+            return True, 0
+        free, _ = torch.cuda.mem_get_info(self._device)
+        return free >= self._min_free_bytes, free
+
+    def _capture(
```
WilsonZheng0327, line 84: missing _device.type != "cuda" guard

sglang_omni/models/arkasr/sglang_model.py
```diff
@@ -165,7 +165,16 @@ def _encode_audio_batch(self, items: List[MultimodalDataItem]) -> torch.Tensor:
                 mel_lengths, device=device, dtype=torch.long
             ).unsqueeze(1)
 
-        encoded = self.audio_encoder(batched, attention_mask=encoder_mask)
+        encoded: torch.Tensor | None = None
+        graph_runner = getattr(self, "encoder_cuda_graph_runner", None)
+        if graph_runner is not None:
+            # Bucketed graph replay always uses an explicit mask, including
+            # B=1.  ``None`` means the shape/bucket failed and the eager path
```
0xjeffro: lets avoid using backtick in comments and similarly in other places

tests/unit_test/arkasr/test_encoder_cuda_graph.py
```diff
@@ -0,0 +1,355 @@
...
+        num_attention_heads=4,
+        num_key_value_heads=2,
+        intermediate_size=64,
+        vocab_size=256,
+        audio_token_id=151663,
+    )
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="requires CUDA")
+def test_cuda_graph_matches_eager_embeddings_on_mixed_mel_lengths() -> None:
```
0xjeffro: this test failed from my side

sglang_omni/models/arkasr/encoder_cuda_graph.py
```diff
@@ -0,0 +1,248 @@
...
+_DEFAULT_MIN_FREE_GB = 3.0
+_DEFAULT_WARMUP_ITERS = 3
+
+
+@dataclass
+class _CapturedGraph:
+    graph: torch.cuda.CUDAGraph
+    features: torch.Tensor
+    mask: torch.Tensor
+    frame_index: torch.Tensor
```
0xjeffro: why we put this frame_index in CapturedGraph

sglang_omni/models/arkasr/encoder_cuda_graph.py
0xjeffro, line 23: We shouldn't hardcoded this, I prefer calculate it based on the relevant parameters.

docs/cookbook/arkasr.md
```diff
@@ -87,6 +87,36 @@ reached; it can release earlier when pending request-build work drains. Set
...
+    enable_encoder_cuda_graph: true
+```
+
+The feature captures the complete audio encoder, including the Whisper/RoPE
+tower and MLP adapter. Before the server becomes ready, it captures the common
+batch buckets `1, 2, 4, 8` at mel-frame buckets `512, 768, 1024`, for 12
+graphs total. Other shapes use the eager encoder path. Bucket policy and the
+free-memory safety margin are model-internal settings.
+
+A request runs through the existing eager encoder path when its shape is not
```
0xjeffro, line 110: we have deleted the logic of lazy capture

sglang_omni/models/arkasr/encoder_cuda_graph.py
```diff
@@ -0,0 +1,248 @@
...
+    callers get a cloned tensor for the real batch rows before a later replay
+    can reuse the bucket. Shapes not captured before serving use eager fallback.
+    """
+
+    def __init__(
+        self,
+        audio_encoder: torch.nn.Module,
+        *,
+        min_free_gb: float = _DEFAULT_MIN_FREE_GB,
+        warmup_iters: int = _DEFAULT_WARMUP_ITERS,
```
0xjeffro: I don't think this `warmup_iters` is necessary just use 3 is enough

sglang_omni/models/arkasr/encoder_cuda_graph.py
```diff
@@ -0,0 +1,248 @@
...
+                    exc,
+                )
+                self._graphs.pop(key, None)
+                return None
+
+
+__all__ = [
+    "ArkasrEncoderCudaGraphRunner",
+    "_bucket_batch",
+    "_bucket_frames",
```
0xjeffro: why export them

sglang_omni/models/arkasr/encoder_service.py
```diff
@@ -26,6 +26,11 @@
 encoder width, whereas for ARK it only sets how many requests reach the call.
 (MOSS-TD spells its drain size ``encoder_max_batch_size``; that name is
 already ARK's model-internal bound, so it is deliberately not reused here.)
+
+#1411 made ARK's ``get_audio_feature`` pad+mask batches across mixed mel
+lengths, so groups drained here now run as one fused tower+adapter forward.
```
0xjeffro: pls follow the format Note (xxx): if this is necessary

wirybeaver:
## H100 Full SeedTTS-EN benchmark — default limits, cache enabled

Benchmark completed on **August 24, 2026**. This rerun keeps all serving parameters at their defaults and compares the PR base against the current head with the ARK encoder CUDA Graph explicitly enabled only on Head.

### Configuration

- **Base:** `ca6aa45513ff0c9728ff122629fde96acc26054d`
- **Head:** `83fefa9ab867dc8fad0dbac1ae4f0a6044781732` with `--stages.asr.factory-args.enable-encoder-cuda-graph true`
- **CI parent:** `hongccc/sglang-omni@sha256:374d0b1c30b2bff685b1716fc64a02ad3b3d0a90fe2ce73ce9861a6992c28101`
- **Base image:** `rfgleamchaser/sglang-omni@sha256:c03e1c5614ac191d5324c0451548b6ac7cb8b31e4a1137b490822c3a8db64926`
- **Head image:** `rfgleamchaser/sglang-omni@sha256:a9f0a780ea606d0598d4cbfed31d7b1be24ef48a3c796ccb64f8fdd8b1c4e7a7`
- **GPU:** H100 SXM 80 GB, UUID `GPU-90751ec7-2cfa-4991-5522-0e165c668c22`, driver `580.173.02`, CUDA 13.0
- **Dataset:** full SeedTTS EN, 1,088 samples, revision `27f4c1adee83b5b29b7c4b375f6b976324bda308`
- **Outer order:** Base → Head → Head → Base
- **Per cell:** 6 measured observations per revision (3 in each order)
- **Empty cache:** fresh server per observation, one discarded 32-clip disjoint warmup, then one measured full-EN pass
- **Warm cache:** fresh server per order/revision/concurrency cell, one full-EN prime, then 3 measured full-EN repeats on that server

Latency deltas are **Head vs Base**; negative is better. Throughput deltas are positive when Head is faster.

### Initially empty embedding cache

| Concurrency | WER Base | WER Head | ΔWER | Throughput (samples/s) | Δ throughput | Mean latency | Δ mean latency | P95 latency | Δ P95 latency | P99 latency | Δ P99 latency |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 1.1199% | 1.1274% | +0.0076 pp | 9.908 → 11.865 | +19.75% | 100.8 → 84.2 ms | -16.51% | 150.1 → 103.2 ms | -31.28% | 166.2 → 118.5 ms | -28.67% |
| 8 | 1.1388% | 1.1123% | -0.0265 pp | 34.890 → 63.834 | +82.96% | 228.7 → 124.9 ms | -45.39% | 342.2 → 179.7 ms | -47.47% | 505.0 → 207.3 ms | -58.94% |
| 16 | 1.1274% | 1.1123% | -0.0151 pp | 56.552 → 99.647 | +76.21% | 281.9 → 159.5 ms | -43.44% | 366.8 → 216.7 ms | -40.91% | 406.0 → 260.0 ms | -35.96% |
| 32 | 1.1274% | 1.1123% | -0.0151 pp | 93.310 → 141.097 | +51.21% | 339.2 → 224.0 ms | -33.97% | 436.2 → 296.5 ms | -32.04% | 507.8 → 346.5 ms | -31.75% |

### Warm embedding cache

| Concurrency | WER Base | WER Head | ΔWER | Throughput (samples/s) | Δ throughput | Mean latency | Δ mean latency | P95 latency | Δ P95 latency | P99 latency | Δ P99 latency |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 1.1199% | 1.1274% | +0.0076 pp | 12.546 → 12.510 | -0.28% | 79.6 → 79.8 ms | +0.28% | 96.8 → 96.6 ms | -0.12% | 111.4 → 111.3 ms | -0.12% |
| 8 | 1.1388% | 1.1123% | -0.0265 pp | 64.870 → 65.015 | +0.22% | 122.7 → 122.4 ms | -0.28% | 168.8 → 170.1 ms | +0.78% | 194.1 → 197.7 ms | +1.89% |
| 16 | 1.1274% | 1.1199% | -0.0076 pp | 110.662 → 110.667 | +0.00% | 143.5 → 143.5 ms | +0.03% | 194.7 → 195.1 ms | +0.23% | 225.5 → 222.5 ms | -1.31% |
| 32 | 1.1236% | 1.1199% | -0.0038 pp | 165.385 → 168.360 | +1.80% | 191.0 → 187.5 ms | -1.82% | 258.4 → 250.3 ms | -3.11% | 321.4 → 334.0 ms | +3.92% |

### Validation and interpretation

- **96/96 measured full-EN passes completed:** 104,448 evaluated requests, 0 skipped.
- **Graph startup gate passed:** 32 Head server starts × 12 buckets = 384 captures; no logged capture or replay failures.
- All runs used the same evaluation-input SHA-256: `59f507226ca9697a70fe97a2e0cdde376bcb96f0ce5e649b4ed0bab022388eb9`.
- Empty-cache throughput gains were consistent across both execution orders: c1 `+20.2%/+19.3%`, c8 `+83.6%/+82.3%`, c16 `+74.6%/+77.8%`, c32 `+50.3%/+52.1%` (Order A/Order B).
- Warm-cache performance is essentially neutral, as expected when cached embeddings bypass encoder work. The largest warm-cache tail movement is c32 P99 `+3.92%`, while throughput is `+1.80%` and mean latency is `-1.82%`.
- WER across all cells stayed in the narrow `1.1123%–1.1388%` range. The small signed differences vary by concurrency/cache state and do not show a systematic CUDA-Graph regression.
- Raw JSONL, result JSON, server logs, resource/utilization metrics, per-repeat CSV, aggregate CSV, and SHA-256 manifests were retained locally before the Vast.ai instance was destroyed.

sglang_omni/models/arkasr/encoder_cuda_graph.py
```diff
@@ -0,0 +1,248 @@
...
+                for frame_bucket in _DEFAULT_FRAME_BUCKETS:
+                    enough, free = self._enough_free_vram()
+                    if not enough:
+                        logger.warning(
+                            "ARK-ASR encoder CUDA graph: free VRAM %.1fGB < %.1fGB "
+                            "headroom; remaining shapes use eager",
+                            free / 1024**3,
+                            self._min_free_bytes / 1024**3,
+                        )
+                        return
```
WilsonZheng0327: right now it iterates from smallest batch and smallest frame and returns on the first low-headroom check, so under memory pressure you keep (1, 512) and lose (8, 1024), which is where launch overhead amortizes best.

reference how whisper and MTD does it

WilsonZheng0327 (changes requested):
Thanks for your work! Promising results, potentially can be even better.

sglang_omni/models/arkasr/encoder_cuda_graph.py
wirybeaver: I tried to add 16 / 32 into the default bucket list and rolled back after benchmarking. Specifically, when the encoder_max_batch and pre_lm_encoder_batch (I don't recall the exact name) is set to 16 / 32, the shape bucket = 16 / 32 is never been replayed even with concurrency 32. I am afraid it will result in increase of server start up and VRAM usage while the Cuda Graph won't be replayed. The _DEFAULT_FRAME_BUCKETS (512, 768, 1024) also comes from the benchmarking result. Only these 3 shapes can be replayed. The total 12 Cuda Graph occupies 600MB VRAM. The rest available VRAM is 8GB-ish on H100 SXM 80GB.

If we still want to avoid hard coding, the max bucket size for capture can be set to encoder_max_batch. But I have no clear idea how to select the upbound of frame using relevant parameters. One simple approach on top of my head is to measure the available VRAM and then iteratively increase by 256 per step till (hitting OOM OR breach a _default_max_mel_frame OR 0.8 * available VRAM)?

wirybeaver: Check my comment https://github.com/sgl-project/sglang-omni/pull/1433#discussion_r3844539285 we can continue the discussion there.

sglang_omni/models/arkasr/encoder_cuda_graph.py
wirybeaver: Done in `1c278403`. The exception scope now covers only `entry.graph.replay()`; input-copy and output-copy errors propagate instead of being mislabeled as replay failures.

sglang_omni/models/arkasr/encoder_cuda_graph.py
wirybeaver: Done in `1c278403`. Startup capture now returns immediately when the encoder device is not CUDA, and CUDA event creation is guarded as well.

sglang_omni/models/arkasr/sglang_model.py
wirybeaver: Done in `1c278403`. Removed the backtick markup from the source comment.

tests/unit_test/arkasr/test_encoder_cuda_graph.py
wirybeaver: Updated in `1c278403`. The CUDA parity test now captures only the exercised `(4, 512)` bucket and asserts that capture succeeded before comparing outputs. The full local ARK-ASR suite passes with 69 passed and 1 CUDA-only test skipped on this non-GPU host.

sglang_omni/models/arkasr/encoder_cuda_graph.py
wirybeaver: Done in `1c278403`. Removed `frame_index` from `_CapturedGraph`; the runner now updates static per-row lengths and constructs the attention mask inside the captured graph.

docs/cookbook/arkasr.md
wirybeaver: Done in `1c278403`. The cookbook now explicitly states that all capture happens during startup and requests never trigger lazy capture.

sglang_omni/models/arkasr/encoder_cuda_graph.py
wirybeaver: Done in `1c278403`. Removed the configurable warmup count and fixed startup warmup at three iterations.

sglang_omni/models/arkasr/encoder_cuda_graph.py
wirybeaver: Done in `1c278403`. Only `ArkasrEncoderCudaGraphRunner` remains in `__all__`; the private bucket helpers are no longer exported.

sglang_omni/models/arkasr/encoder_service.py
wirybeaver: Done in `1c278403`. Rewrote this as a `Note (author):` entry.

sglang_omni/models/arkasr/encoder_cuda_graph.py
wirybeaver: Done in `1c278403`. Startup capture now visits batch and frame buckets largest-first, and insufficient-headroom buckets are skipped instead of terminating the entire capture pass.

sglang_omni/models/arkasr/encoder_cuda_graph.py
WilsonZheng0327: regarding the "16 / 32 never been replayed", i think the problem here is `pre_lm_max_batch_wait_ms`. draining time here isn't long enough to allow more requests to be batched together. matter of fact, do you have the numbers for how many times 1 2 4 8 are called? if you haven't checked these numbers, it's possible that even the higher ones of these four aren't getting used much. in that case, `pre_lm_max_batch_wait_ms` default tuning needs to be done.

regarding parameterizing the precapture, I believe you can derive `max_frames` from `feature_extractor.nb_max_frames` and the batch ceiling from `encoder_max_batch_size`.

wirybeaver:
I meant to modify this PR by Wilson's suggestion on inducting the bucket list from parameters and prioritize the large bucketing capture instead of hard coding the bucket list. But found out that https://github.com/sgl-project/sglang-omni/pull/1728 implemented those comments on top of current PR. @guozhihao-224 Appreciate your eagerness and passion on the improvement but I think it's better to create a incremental stacking PR next time if you want to improve other contributor's or asking for co-authoring to avoid duplicate work from scratch.

PR #1440, opened 2026-08-10
[Perf] dots.tts: bypass redundant vocoder staging

luojiaxuan:
## Motivation

The dots.tts vocoder always allocated a zero-padded batch and copied every latent input into it, even when a bucket contained one request or every request had the same frame length. Streaming also called `torch.cat` for the first two one-patch chunks, where the input list has exactly one tensor.

Those cases do not need padding.

## Modifications

- Pass a singleton non-streaming latent tensor directly to AudioVAE.
- Assemble equal-length multi-request buckets with one `torch.cat`.
- Keep the existing zero-padding and crop path for genuinely mixed-length buckets.
- Reuse the single pending streaming patch directly; retain `torch.cat` for merged multi-patch chunks.
- Add storage-identity regressions for singleton non-streaming and streaming inputs.

## Impact

This removes redundant zero initialization, per-row copies, and singleton concatenation while preserving request order, mixed-length behavior, waveform cropping, and AudioVAE inputs.

RTX A6000 / PyTorch 2.11.0+cu130 assembly microbenchmark (bf16, 128 frames x 128 dims):

| Equal-length batch | Before (us/batch) | After (us/batch) | Reduction |
| ---: | ---: | ---: | ---: |
| 1 | 39.951 | 1.082 | 97.29% |
| 4 | 85.686 | 14.653 | 82.90% |
| 8 | 187.155 | 17.626 | 90.58% |
| 16 | 357.004 | 18.554 | 94.80% |

The one-patch streaming assembly path drops from 14.424 us/chunk to a direct tensor reuse (0.089 us in the same microbenchmark).

These are isolated input-assembly numbers; AudioVAE forward remains the dominant vocoder cost, so this is not an end-to-end throughput claim.

## Validation

- `python -m pytest -q tests/unit_test/dots_tts/test_vocoder.py tests/unit_test/dots_tts/test_vocoder_streaming.py` — 14 passed
- `ruff check sglang_omni/models/dots_tts/vocoder.py tests/unit_test/dots_tts/test_vocoder.py tests/unit_test/dots_tts/test_vocoder_streaming.py`
- `git diff --check`

Tested locally and in the canonical Aries A6000 container with PyTorch 2.11.0+cu130, SGLang 0.5.16, and `dots.tts==0.2.1`.

Related: #1367

### DIFF
--- a/sglang_omni/models/dots_tts/vocoder.py
+++ b/sglang_omni/models/dots_tts/vocoder.py
@@ -55,9 +55,7 @@ class DotsTTSBatchVocoder(BatchVocoderBase):
 
         with self.codec.lock:
             for bucket_items in buckets.values():
-                frame_counts = [
-                    int(latents.shape[1]) for _, latents in bucket_items
-                ]
+                frame_counts = [int(latents.shape[1]) for _, latents in bucket_items]
                 max_frames = max(frame_counts)
                 if len(bucket_items) == 1:
                     padded = bucket_items[0][1].to(self.codec.device)
--- a/sglang_omni/models/dots_tts/vocoder.py
+++ b/sglang_omni/models/dots_tts/vocoder.py
@@ -55,16 +55,27 @@ class DotsTTSBatchVocoder(BatchVocoderBase):
 
         with self.codec.lock:
             for bucket_items in buckets.values():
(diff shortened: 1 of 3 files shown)


SandyLuXY (approved):
lgtm, please fix lint issue

PR #1434, opened 2026-08-09
[dots.tts] Add per-call-site contention profiling for the shared codec lock

buffett0323:
## Motivation
The reference encoder and the vocoder share one `DotsAudioCodec` and one `self.lock`, but run on different stage threads. Streaming decode must emit one chunk (`patch_size * merge_steps` = 8 frames, ~80 ms of real-time audio) per 80 ms; reference encode measured ~51 ms in #1420. So one cache miss is a large fraction of a chunk budget if it lands on the streaming thread. Also, there's no lock instrumentation anywhere in the repo, so neither "this is fine" nor "split the lock" could be defended. The roadmap gates the fix on data, so this PR ships the measurement.

## Related Issues
Part of #1367 (Measure codec lock contention).

## Modifications
- **`sglang_omni/utils/lock_profile.py`** — `ProfiledRLock`, a drop-in `RLock` recording per-site `acquisitions`, `contended`, `wait_s`, `max_wait_s`, `hold_s`, `max_hold_s`.
- **Six sites labeled**, all still sharing the one lock: `reference_encode` (`codec.py`); `vocoder_batch_decode`, `stream_state_init`, `stream_step`, `stream_flush`, `stream_fallback_decode` (`vocoder.py`).
- **Codec surface** — `lock_stats()` and `maybe_log_lock_stats()` (30 s rate limit). The rate-limit guard uses a **separate** lock: reporting through the codec lock would contend with and skew the measurement.
- **No behavior change.** `with self.codec.lock:` became `with labeled(self.codec.lock, "name"):`. No lock is split, no CUDA stream added, mutual exclusion is identical, the string is an attribution tag.

| site | acq | contended | wait_s | max_wait | hold_s | max_hold |
| :--- | :---: | :---: | :---: | :---: | :---: | :---: |
| `stream_step` | 80 | 14 | 0.4193 | 0.0755 | 0.2007 | 0.0025 |
| `reference_encode` | 5 | 5 | 0.0605 | 0.0252 | 0.1212 | 0.0251 |

## Overhead

Median of 15 runs x 100k acquisitions:

| | ns/acq | vs raw |
| --- | --- | --- |
| `threading.RLock` | 62.0 | — |
| profiler **disabled** (default) | 116.9 | +55 |
| profiler **enabled** | 1014.7 | +953 |

[...]

PR #1445, opened 2026-08-10
[Perf] dots.tts: separate reference and vocoder codec locks

Hayden727:
## Motivation

The dots.tts reference encoder and vocoder use disjoint AudioVAE submodules but currently share one process-wide lock. #1434 measured material cross-stage contention: 14/80 `stream_step` calls contended for 419 ms in total, while all 5 reference encodes also contended.

## Changes

- use an independent lock for speaker/AudioVAE reference encoding
- keep all batch and streaming vocoder calls serialized behind one vocoder lock because `VocoderInference` owns mutable streaming caches/state
- add a focused test that reference encoding never acquires the vocoder lock

No model weights, math, request scheduling, or output layout change.

## Performance

1x H200, SeedTTS EN, concurrency 8, 10 warmups, seed 42.

### Streaming, 128 samples (mean of 2 fresh-server runs)

| Metric | main (`2e607bc`) | This PR | Delta |
|---|---:|---:|---:|
| Request throughput | 1.427 req/s | 1.616 req/s | **+13.25%** |
| Mean latency | 5.544 s | 4.902 s | **-11.58%** |
| Mean RTF | 1.561 | 1.363 | **-12.66%** |
| Audio throughput | 5.377 s/s | 6.086 s/s | **+13.20%** |
| TTFC | 2.972 s | 1.817 s | **-38.87%** |
| ITL | 0.3669 s | 0.4394 s | **+19.78%** |

The ITL regression is disclosed: removing head-of-line blocking lets cold reference work progress sooner, which improves TTFC, total latency, and throughput but increases GPU competition between later audio chunks. Both paths remain faster than real time on this workload, and all runs completed 128/128 requests without failures.

### Non-streaming, 512 samples

| Metric | main (`2e607bc`) | This PR | Delta |
|---|---:|---:|---:|
| Request throughput | 3.072 req/s | 3.192 req/s | **+3.91%** |
| Mean latency | 2.587 s | 2.493 s | **-3.63%** |
| Mean RTF | 0.6372 | 0.6138 | **-3.67%** |
| Audio throughput | 12.742 s/s | 13.210 s/s | **+3.67%** |

Both runs completed 512/512 requests without failures.

## Correctness

With the real dots codec, three concurrent reference-encode + AudioVAE-decode trials were compared with serialized execution using fixed inputs. Speaker embeddings, latent distributions, and PCM were exactly equal (max absolute difference 0).

## Validation

- `pytest -q tests/unit_test/dots_tts/test_reference_encode_batching.py tests/unit_test/dots_tts/test_vocoder.py tests/unit_test/dots_tts/test_vocoder_streaming.py tests/unit_test/dots_tts/test_result_contracts.py` (42 passed)
- pre-commit hooks passed

Related: #1367
Follow-up to the contention evidence in #1434.

### DIFF
--- a/sglang_omni/models/dots_tts/codec.py
+++ b/sglang_omni/models/dots_tts/codec.py
@@ -69,7 +69,10 @@ class DotsAudioCodec:
         self.sample_rate = int(vocoder.sample_rate)
         self.hop_size = int(vocoder.hop_size)
         self.device = torch.device(device)
-        self.lock = threading.RLock()
+        # Encoder and decoder modules are disjoint; vocoder calls still share one
+        # lock because VocoderInference owns mutable streaming state caches.
+        self.reference_lock = threading.RLock()
+        self.vocoder_lock = threading.RLock()
 
     @staticmethod
     def _reference_load_workers(count: int) -> int:
@@ -113,7 +116,7 @@ class DotsAudioCodec:
         )
         speaker_batch, speaker_lengths = self._speaker_input(batch, audio_lengths)
 
-        with self.lock:
+        with self.reference_lock:
             speaker = self.speaker(speaker_batch, audio_lengths=speaker_lengths)
             latent_distribution = self.inference.extract_latents(batch)
 
--- a/sglang_omni/models/dots_tts/vocoder.py
+++ b/sglang_omni/models/dots_tts/vocoder.py
@@ -53,7 +53,7 @@ class DotsTTSBatchVocoder(BatchVocoderBase):
             )
             self._logged_batch = True
 
-        with self.codec.lock:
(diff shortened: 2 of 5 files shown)


luojiaxuan:
H100 full-set GPU-crossover A/B on current main: c16 base 4.760 vs split 4.783 req/s (+0.49%); c32 base/split medians 4.973/4.973 and means 4.988/4.947 (-0.82%, one slow split run), with all 12 runs completing 1088/1088 requests and zero failures. This supports positioning the change as a streaming optimization because the non-streaming +3.91% does not generalize to c16/c32; the current-main port is 6d54e21a.

PR #1431, opened 2026-08-09
[ASR][Perf] Enable async lookahead for single-request Qwen3-ASR

SandyLuXY:
## Motivation

Qwen3-ASR enables shared one-step-lookahead async decode by default, but its
minimum async batch size is currently two. Batch size one therefore falls back
to synchronous decode even though the measured async path improves the
single-request workload without changing WER or completion coverage.

This implements Q-PR2 from the ASR optimization roadmap by selecting the
measured Qwen3-ASR crossover while keeping the shared scheduler and other model
defaults unchanged.

## Modifications

1. `sglang_omni/models/qwen3_asr/stages.py`
   - Changes the Qwen3-ASR stage default `async_decode_min_batch_size` from `2`
     to `1`, enabling lookahead for every non-empty decode batch.
   - Leaves the shared `OmniScheduler` default and all other model policies
     unchanged.
2. `sglang_omni/cli/serve.py`
   - Updates `--async-lookahead-min-batch-size` help to describe the
     Qwen3-ASR-specific default and the default used by other supported models.
3. `tests/unit_test/qwen3_asr/test_pipeline.py`
   - Updates the production stage-factory default assertion to lock in the new
     threshold.
4. `docs/cookbook/qwen3_asr.md`
   - Documents that Qwen3-ASR now uses async lookahead at every decode batch
     size, including batch size one.

## Related Issues

Related to #1396 (Q-PR2).

## Accuracy Test

**Setup:**

- **Refs:** baseline `main` @ `efef31d6`; candidate
  `bench/async-lookahead-min-batch-1-20260807` @ `efef31d6` with
  `--async-lookahead-min-batch-size 1` (the behavior made default by this PR).
- **Model:** `Qwen/Qwen3-ASR-1.7B`, revision
  `7278e1e70fe206f11671096ffdd38061171dd6e5`.
- **Dataset:** full SeedTTS EN, revision
  `27f4c1adee83b5b29b7c4b375f6b976324bda308`, 1,088 clips.
- **Workload:** non-streaming transcription at concurrency `1,8,32`, three
  measured repeats per level, no discarded warmup.
- **Hardware / runtime:** physical GPU 1, NVIDIA H100 80GB HBM3.

| Metric | Main | Min batch size 1 | Delta |
| --- | ---: | ---: | ---: |
| Corpus WER, c=1 | 0.0122 | 0.0122 | 0.0000 |
| Corpus WER, c=8 | 0.0122 | 0.0122 | 0.0000 |
[...]

PR #1447, opened 2026-08-10
[Perf] Speed up Qwen3-ASR low and mid concurrency

luojiaxuan:
## Summary

- reduce Qwen3-ASR request-construction and scheduler overhead, including the batch-size-1 async lookahead from #1431
- use text RoPE positions and the fused QK-norm + RoPE kernel on the BF16 ASR path
- enable partial torch compile coverage through batch size 2 and tune prefill coalescing for low/mid concurrency

Related to #1396 and #1324. The #1431 work is included here, with SandyLuXY retained as co-author.

## H100 results

Single H100 80 GB GPU 0, Qwen/Qwen3-ASR-1.7B at revision 7278e1e70fe206f11671096ffdd38061171dd6e5, full SeedTTS EN 1088 at revision 27f4c1adee83b5b29b7c4b375f6b976324bda308. Each level used one discarded warmup plus three full repeats; concurrency 2 was additionally confirmed with five full repeats.

| Concurrency | Baseline req/s | This PR req/s | Delta | Mean / p95 latency |
|---:|---:|---:|---:|---:|
| 1 | 21.711 | 24.551 | +13.08% | 41 / 50 ms |
| 2 | 40.515 | 41.260 | +1.84% | 48 / 75 ms |
| 4 | 69.998 | 72.998 | +4.29% | 54 / 89 ms |
| 8 | 115.611 | 126.364 | +9.30% | 63 / 102 ms |
| 16 | 162.290 | 165.969 | +2.27% | 96 / 130 ms |
| 32 | 207.719 | 288.132 | +38.71% | 109 / 148 ms |
| 64 | 236.639 | 318.023 | +34.39% | 198 / 340 ms |

All measured requests completed. Corpus WER remained 0.0122.

## Validation

- focused unit tests: 115 passed
- full pre-commit suite: passed


sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -908,6 +908,11 @@ def _run_request_builder(self, payload: Any, active_stage: str | None) -> Any:
         )
         return req_data
 
+    def _sleep_during_idle(self) -> None:
+        with self._request_admission_lock:
```
SandyLuXY, line 912: This new required-field read breaks the existing async event-loop fixtures, and the GitHub unit job fails with `_request_admission_lock` missing. Please initialize the admission lock and pending-build map in every real-loop fixture and rerun the pipeline subtree so both changed idle call sites are covered.

SandyLuXY (changes requested):

sglang_omni/scheduling/omni_scheduler.py
luojiaxuan: Done, initialized the admission lock and pending-build map in the async real-loop fixture, and added normal-loop idle coverage. The focused idle-path tests pass.

SandyLuXY (approved):
lgtm

PR #1446, opened 2026-08-10
[Perf] dots.tts: slot-pooled batched streaming AudioVAE

guozhihao-224:
## Motivation
Roadmap item from #1367: move the streaming vocoder to slot-based batched state.

Upstream `VocoderStreamState` keeps scalar `total_frames` / `emitted_frames`, so `batch_size=N` is lockstep-only. Serving needs independent causal ages per request, while still decoding equal-T chunks for multiple live streams in one eager AudioVAE forward. The compiled / CUDA-Graph `stream_step` path remains disabled because of the allocator conflict with backbone decode graphs (#1392 / #1395).

**This PR lands the eager slot-pool + coalescing path only. Streaming AudioVAE CUDA Graph and first/steady/merge-step tuning stay follow-ups.**

## Modifications
- Add `DotsVocoderSlotPool`: per-slot LSTM / decoder window / frame counters; equal-T gather → one eager step → scatter; `@torch.no_grad()` on `step`/`flush`.
- Rework `DotsTTSStreamingVocoder` onto `StreamingVocoderBase` coalescing (`_can_batch_stream_chunks`): exact-T grouping via `select_step_participants` / `build_step_plan` / `run_step`; stream-done drains with flush + release under the codec lock.
- Wire `stream_slots` (default 16) through `stages.py` and `examples/configs/dots_tts.yaml`.
- Unit tests for equal-T batching, independent counters, admission failure, coalescing, flush/release.
Implementation note: the pool calls `VocoderInference` private eager helpers (`_decode_stream_latents` / `_decode_stream_window` / …) so rows can age independently. Prefer that over public `stream_step` until upstream exposes a per-slot API.
## Related
- Roadmap #1367 (streaming vocoder slot-based batched state)
- Soft file conflicts expected with #1440 (vocoder staging) and #1434 (codec lock labels); will rebase when they land
- Not claiming: streaming AudioVAE CUDA Graph; chunk / merge-step tuning; backbone `torch.compile`

## Benchmark
**Smoke (earlier, dual-GPU, n=64, generate-only, H200):** streaming c=1 flat; c=8 ~−9% RTF / +9% qps; c=16 ~−35% RTF / +43% qps / large TTFC drop. Treat as directional only (not same-card / not full EN / no WER).
**Canonical same-card A/B (in progress on this host):**
- H200 GPU2 serial baseline → slot
- Seed-TTS-Eval EN full 1088, seed 42, warmup 8, `--stream`
[...]

PR #1448, opened 2026-08-10
[Perf] dots.tts: graph streaming AudioVAE decode

Hayden727:
## Motivation

Capture and replay the AudioVAE forward pass used by the streaming slot pool introduced in #1446. This reduces the CPU work needed to launch GPU kernels for each audio chunk.

On H200 at concurrency 16, the paired benchmark shows 18% higher throughput, 20% lower mean latency, and 56% lower median time to first chunk. The tested outputs are bit-identical to eager execution.

This builds on @Hayden727's original capture implementation, adapted to the slot-pool path by @Lenoplus42. Related: #1650 and #1882 (Q4).

## Changes

- Extract the AudioVAE forward pass from `DotsVocoderSlotPool.step()` so it can be captured. Input gathering, state updates, and slot management remain outside the graph.
- Capture one native `torch.cuda.CUDAGraph` for each supported `(batch size, frame count)` at startup. The canonical config uses 16 graphs: batch sizes 1–4 and frame counts 4, 8, 12, and 16. Capture runs largest first on a dedicated stream, with a shared memory pool and two warmup passes per shape.
- Before replay, copy the live inputs and decoder state into the graph's buffers. Clone the outputs after replay so later calls cannot overwrite returned results. Shapes without a captured graph use eager execution, including shapes whose capture failed.
- Add `enable_streaming_audio_vae_cuda_graph`. It is enabled in `examples/configs/dots_tts.yaml`; the factory default is `false`. Log captured shapes at startup and replay/miss counts periodically during serving.

Capture uses native CUDA graphs without `torch.compile` or Inductor.

## Correctness and stability

- **Unit tests:** 122 passed on H200 with PyTorch 2.13.0+cu130, as reported in the test run. Graph tests cover batches of 1–4, both frame counts used by the test decoder, and requests at different decoding steps within the same batch.
- **Real AudioVAE:** nine steps with `dots.tts-mf@c28105ad`, covering batch sizes 1–4 and frame counts 4, 8, 12, and 16. Waveforms and pool state matched eager execution exactly (`torch.equal`), with nine replays and no misses.
- **End to end:** one streaming request with seed 42 produced byte-identical PCM with the flag off and on

## H200 serving benchmark

Measured on September 14, 2026, on one H200 in a shared host, using Seed-TTS-Eval EN streaming requests at concurrency 1 and 16 (`c1` and `c16`).

Both servers ran on the same GPU. Each comparison used alternating bursts of 48 requests, reversing which server received the first burst in each pair. We ran 11 pairs, discarded the first, then repeated with the server startup order reversed: 20 scored pairs per concurrency level.

| Metric | c1 main | c1 graph | Delta (20 pairs) | c16 main | c16 graph | Delta (20 pairs) |
| Completed / failed | 960 / 0 | 960 / 0 | - | 960 / 0 | 960 / 0 | - |
| Throughput (req/s) | 0.4920 | 0.4995 | +1.6% (14/20) | 1.9670 | 2.3365 | **+17.8%** (20/20) |
| Audio throughput (s/s) | 1.9860 | 2.0310 | +2.7% (14/20) | 7.9595 | 9.3790 | **+18.0%** (20/20) |
| Mean latency (s) | 2.0335 | 2.0010 | -1.7% (14/20) | 7.3465 | 5.9985 | **-19.6%** (20/20) |
| Mean RTF | 0.5106 | 0.5004 | -2.5% (14/20) | 1.9022 | 1.5197 | **-20.8%** (20/20) |
| Median TTFC (s) | 0.2712 | 0.2220 | **-18.3%** (20/20) | 1.5444 | 0.6660 | **-56.3%** (20/20) |
| TTFC p95 (s) | 0.2999 | 0.2919 | -6.7% (14/20) | 3.6147 | 1.4335 | **-59.7%** (20/20) |
| Mean ITL (s) | 0.2364 | 0.2340 | -1.2% (13/20) | 0.7238 | 0.6986 | -3.0% (14/20) |
| ITL p95 (s) | 0.3067 | 0.3313 | +5.1% (8/20) | 1.3198 | 1.2959 | -0.8% (12/20) |
| Playback underrun p95 (s) | 0.1623 | 0.1631 | +2.6% (8/20) | 1.1772 | 1.1525 | -4.2% (12/20) |

At **c16**, throughput, mean latency, RTF, and both TTFC metrics improved in all 20 pairs. At **c1**, median TTFC improved in all 20 pairs.

### Cost and profiling

- Server startup increased from 55 to 65 seconds with 16 graphs enabled.
- An isolated capture measurement reserved an additional 976 MiB of GPU memory.
- In separate c16 profiles, the vocoder thread's `cudaLaunchKernel` count fell from about 454,000 to 84,000, and time inside those calls fell from 5.1 to 1.4 seconds. Thread activity fell from 92% to 65%. These profiles help explain the launch overhead; the paired benchmark above measures serving performance.

The graph covers the streaming step. Final `flush` decoding still runs eagerly, and each step still decodes the full 45-frame window. Those are possible follow-ups.

## Checklist

- [x] Add unit tests.
- [x] Compare graph and eager outputs with the real codec and an end-to-end request.
- [x] Run paired benchmarks with both server startup orders.

### DIFF
--- a/sglang_omni/models/dots_tts/vocoder.py
+++ b/sglang_omni/models/dots_tts/vocoder.py
@@ -263,6 +263,7 @@ class DotsTTSStreamingVocoder(StreamingVocoderBase[_DotsStreamState, None]):
         codec: DotsAudioCodec,
         *,
         optimize: bool,
+        enable_streaming_audio_vae_cuda_graph: bool = False,
         merge_steps: int = 4,
         max_batch_size: int = 4,
         max_batch_wait_ms: int = 2,
@@ -281,7 +282,7 @@ class DotsTTSStreamingVocoder(StreamingVocoderBase[_DotsStreamState, None]):
             _DotsVocoderCudaGraphs(
                 codec, chunk_size=codec.patch_size * self.merge_steps
             )
-            if self.optimize
+            if self.optimize and enable_streaming_audio_vae_cuda_graph
             else None
         )
         super().__init__(
--- a/sglang_omni/models/dots_tts/vocoder.py
+++ b/sglang_omni/models/dots_tts/vocoder.py
@@ -128,6 +128,135 @@ class _DotsStreamState:
     received_patches: int = 0
 
 
+@dataclass
+class _CapturedVocoderGraph:
+    graph: torch.cuda.CUDAGraph
+    latents: torch.Tensor
+    hidden_h: torch.Tensor
+    hidden_c: torch.Tensor
+    window: torch.Tensor
+    valid_frames: torch.Tensor
+    output: tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]
+
+
+class _DotsVocoderCudaGraphs:
+    """Native exact-shape graphs that coexist with the backbone CUDA Graph."""
+
+    def __init__(self, codec: DotsAudioCodec, *, chunk_size: int) -> None:
+        self._codec = codec
+        self._chunk_size = int(chunk_size)
+        self._graphs: dict[int, _CapturedVocoderGraph] = {}
+        self._replays = 0
+        self._fallbacks = 0
+        try:
+            self._capture(
+                tuple(range(codec.patch_size, chunk_size + 1, codec.patch_size))
+            )
+        except Exception:
+            logger.warning(
+                "dots.tts vocoder CUDA graph setup failed; using eager fallback",
+                exc_info=True,
+            )
+
+    @torch.no_grad()
+    def _capture(self, frame_counts: tuple[int, ...]) -> None:
+        device = torch.device(self._codec.device)
+        if device.type != "cuda":
+            return
+        inference = self._codec.inference
+        with torch.cuda.device(device):
+            current_stream = torch.cuda.current_stream(device)
+            capture_stream = torch.cuda.Stream(device=device)
+            capture_stream.wait_stream(current_stream)
+            graph_pool = torch.cuda.graph_pool_handle()
+            for frames in sorted(frame_counts, reverse=True):
+                state = inference.init_stream_state(
+                    batch_size=1, chunk_size=self._chunk_size
+                )
+                hidden_h, hidden_c = state.lstm_hidden
+                latents = hidden_h.new_zeros(1, self._codec.latent_dim, frames)
+                window = state.decoder.window
+                valid_frames = window.new_zeros((), dtype=torch.int64)
+                run = lambda: inference._compiled_stream_step(
+                    latents, hidden_h, hidden_c, window, valid_frames
+                )
+                graph = torch.cuda.CUDAGraph()
+                try:
+                    with torch.cuda.stream(capture_stream):
+                        for _ in range(2):
+                            run()
+                    capture_stream.synchronize()
+                    with torch.cuda.graph(
+                        graph,
+                        pool=graph_pool,
+                        stream=capture_stream,
+                        capture_error_mode="thread_local",
+                    ):
+                        output = run()
+                    self._graphs[frames] = _CapturedVocoderGraph(
+                        graph,
+                        latents,
+                        hidden_h,
+                        hidden_c,
+                        window,
+                        valid_frames,
+                        output,
+                    )
+                except Exception as exc:
+                    graph.reset()
+                    logger.warning(
+                        "dots.tts vocoder CUDA graph capture failed for frames=%d: "
+                        "%s; using eager fallback",
+                        frames,
+                        exc,
+                    )
+            current_stream.wait_stream(capture_stream)
+            torch.cuda.synchronize(device)
+        logger.info(
+            "dots.tts streaming vocoder CUDA graphs: frames=%s",
+            sorted(self._graphs),
+        )
+
+    @torch.no_grad()
+    def decode(self, latents: torch.Tensor, state: Any) -> torch.Tensor | None:
+        captured = self._graphs.get(int(latents.shape[1]))
+        if captured is None or captured.window.shape != state.decoder.window.shape:
+            self._fallbacks += 1
+            if self._fallbacks == 1 or (self._replays + self._fallbacks) % 1024 == 0:
+                self._log_hit_rate()
+            return None
+        self._replays += 1
+        if self._replays == 1 or (self._replays + self._fallbacks) % 1024 == 0:
+            self._log_hit_rate()
+        captured.latents.copy_(latents.transpose(1, 2))
+        captured.hidden_h.copy_(state.lstm_hidden[0])
+        captured.hidden_c.copy_(state.lstm_hidden[1])
+        captured.window.copy_(state.decoder.window)
+        captured.valid_frames.fill_(
+            min(state.decoder.total_frames, state.decoder.window.size(-1))
+        )
+        captured.graph.replay()
+        audio_window, hidden_h, hidden_c, window = captured.output
+        state.lstm_hidden = (hidden_h.clone(), hidden_c.clone())
+        state.decoder.window = window.clone()
+        state.decoder.total_frames += int(latents.shape[1])
+        return self._codec.inference._slice_stream_audio_window(
+            audio_window, state, final=False
+        ).clone()
+
+    def __len__(self) -> int:
+        return len(self._graphs)
+
+    def _log_hit_rate(self) -> None:
+        calls = self._replays + self._fallbacks
+        logger.info(
+            "dots.tts vocoder CUDA graph hit rate: %.1f%% (%d/%d)",
+            100.0 * self._replays / calls,
+            self._replays,
+            calls,
+        )
+
+
 class DotsTTSStreamingVocoder(StreamingVocoderBase[_DotsStreamState, None]):
     def __init__(
         self,
@@ -148,6 +277,13 @@ class DotsTTSStreamingVocoder(StreamingVocoderBase[_DotsStreamState, None]):
         self.optimize = bool(optimize)
         self.merge_steps = int(merge_steps) if optimize else 1
         self._batch_vocoder = DotsTTSBatchVocoder(codec)
+        self._cuda_graphs = (
+            _DotsVocoderCudaGraphs(
+                codec, chunk_size=codec.patch_size * self.merge_steps
+            )
+            if self.optimize
+            else None
+        )
         super().__init__(
             self._batch_vocoder.decode_payload,
             batch_compute_fn=self._batch_vocoder.decode_payloads,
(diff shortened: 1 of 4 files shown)


Lenoplus42:
Hi @Hayden727 — checking in on this PR before picking up related work.

Since this was opened, #1446 (slot-pooled batched streaming AudioVAE) landed and reworked the streaming vocoder path: per-request stream state moved into `DotsVocoderSlotPool`, and same-T requests now run as one batched eager step
(dynamic B, gather/scatter per step). That introduces architectural conflicts here, as the captured target (single-request B=1 `_compiled_stream_step` with static per-request state buffers) no longer exists on main.

The benchmark here were also measured against the pre-#1446 baseline. 1446's batching will remove part of the per-request overhead, so the performance would be different here.

Parts of #1448 that still stand: native capture avoiding the Inductor allocator conflict (#1392/#1395), the bit-identical eager/graph validation methodology, and the explicit opt-in gating.

Update: will build on this as got confirmation.  My plan would be: rework capture onto the slot-pool step, re-run the canonical same-card A/B on current main, and revisit ITL tradeoff under batching.

Lenoplus42:
Status update, branch replaced with rework on top of #1446 & Hayden's work.

1446 moved streaming decode to a slot pool. The graphs now capture the slot-pool step's pure forward (LSTM + decoder window), keyed by `(batch, frames)`: batch 1..`max_batch_size`, frames `patch_size × 1..merge_steps`, 16 graphs on the canonical config.

- Unit tests `tests/unit_test/dots_tts` 122/122, including graph-vs-eager bit-identical cases on a fake decoder.
- Real codec: 9 coalesced steps covering batch 1..4, all four frame counts, and rows of unequal age — every waveform and the pool state `torch.equal` to eager; `replays=9 misses=0`.
- Capture cost: 16 graphs in 15.2 s, reserved memory +976 MiB.
- Per-step micro numbers (isolated, idle card): eager issues ~2800 kernel launches per step; graph replaces them with one `cudaGraphLaunch`. Step time (1,4): 35.1 → 26.8 ms; (4,16): 103.9 → 91.0 ms; (4,4): 97.3 → 89.7 ms. The remainder is GPU compute, so on an idle card this step is mostly GPU-bound.

**In progress**
Streaming profile per #1882 Q4 (c1 / c16 TTFA + ITL with a chunk-timestamp client, `/start_profile` trace and py-spy on the vocoder thread), with the flag off vs on. The isolated saving is 8–13 ms per step; whether that clears the A/A band under real serving load is what the profile has to show before any A/B claim. Will test with paired gate once that is in.

Lenoplus42:
Paired serving gate results (thanks @db-ol for the paired_gate methods support). 
One H200 on a shared host (driver 595.71.05, torch 2.13.0+cu130), branch head `ab9b710d` (rebased on main), canonical `examples/configs/dots_tts.yaml` with the flag off (base) / on (graph). Checked to make sure CUDA graphs captured appears.

| Metric | c1 main | c1 graph | Delta (20 pairs) | c16 main | c16 graph | Delta (20 pairs) |
| Completed / failed | 960 / 0 | 960 / 0 | - | 960 / 0 | 960 / 0 | - |
| Throughput (req/s) | 0.4920 | 0.4995 | +1.6% (14/20) | 1.9670 | 2.3365 | **+17.8%** (20/20) |
| Audio throughput (s/s) | 1.9860 | 2.0310 | +2.7% (14/20) | 7.9595 | 9.3790 | **+18.0%** (20/20) |
| Mean latency (s) | 2.0335 | 2.0010 | -1.7% (14/20) | 7.3465 | 5.9985 | **-19.6%** (20/20) |
| Mean RTF | 0.5106 | 0.5004 | -2.5% (14/20) | 1.9022 | 1.5197 | **-20.8%** (20/20) |
| Median TTFC (s) | 0.2712 | 0.2220 | **-18.3%** (20/20) | 1.5444 | 0.6660 | **-56.3%** (20/20) |
| TTFC p95 (s) | 0.2999 | 0.2919 | -6.7% (14/20) | 3.6147 | 1.4335 | **-59.7%** (20/20) |
| Mean ITL (s) | 0.2364 | 0.2340 | -1.2% (13/20) | 0.7238 | 0.6986 | -3.0% (14/20) |
| ITL p95 (s) | 0.3067 | 0.3313 | +5.1% (8/20) | 1.3198 | 1.2959 | -0.8% (12/20) |
| Playback underrun p95 (s) | 0.1623 | 0.1631 | +2.6% (8/20) | 1.1772 | 1.1525 | -4.2% (12/20) |

Values are medians over the 20 pairs; Delta is the median per-pair delta with the number of pairs graph wins. The A/A floor on this host (same config on both servers, 10 pairs) is within 1.2 percent at c1 and 1.4 percent at c16 on every metric except c16 TTFC p95 and underrun p95, which swing +-16 / +-3 percent between two identical servers and are not decisive here.

At c16 the first chunk arrives in less than half the time (1.54 -> 0.67 s median, 20 of 20 pairs), throughput is +18 percent and mean latency -20 percent, 20 of 20 on each; ITL mean and p95 are unchanged within the floor, so the earlier first chunk does not come at the cost of chunk cadence, and playback underrun does not move. At c1 the first chunk arrives 18 percent earlier (20 of 20) and everything else sits inside the floor.

Cost: startup 55 -> 65 s for the 16 graphs, +976 MiB reserved in the isolated capture measurement.

Where the gain comes from, per the profile posted above: with the flag off the vocoder thread is busy 92 percent of a c16 window, mostly issuing ~2800 kernel launches per step at 11-20 us each under CPU contention; with graphs it drops to 65 percent and the freed CPU goes to the latent-engine thread in the same process, so the first chunk lands sooner and the pipeline as a whole runs faster.

Lenoplus42:
Here's some follow-ups from the profiling.

1. Next step should capture `flush` too. Basically at the end of each request the slot pool re-decodes the whole 45-frame window eagerly to release the lookahead frames, ~2800 launches per request. With the step graphs on it is the dominant remaining launch source on the vocoder thread.

2. The vocoder thread now waits on the latent engine a third of the time. Will do a py-spy sample of the latent-engine process to see what we can do.

db-ol:
Can you help take a look, @Dayuxiaoshui ? Thanks

luojiaxuan (commented):
Solid validation and a well-designed paired benchmark. Before merge:

1. Rebase: this conflicts with main after #2267, and the #2338 / #2339 lint rules will apply.
2. The graph tests don't run in CI (inline).
3. A capture failure should fail startup, not fall back silently (inline).

Two questions:
- At c1, ITL p95 is +5.1% and underrun p95 +2.6%, both above your 1.2% A/A floor. 8/20 wins suggests noise, but could you post the per-pair spread?
- Any reason to keep the factory default off, rather than on with an opt-out?

tests/unit_test/dots_tts/test_vocoder_cuda_graph.py
```diff
@@ -0,0 +1,198 @@
...
+    for step_index, (expected, actual) in enumerate(zip(reference, candidate)):
+        assert expected.keys() == actual.keys()
+        for slot in expected:
+            assert torch.equal(expected[slot], actual[slot]), (
+                f"step {step_index} slot {slot} differs: "
+                f"max|delta|={(expected[slot] - actual[slot]).abs().max().item():.3e}"
+            )
+
+
+requires_cuda = pytest.mark.skipif(
```
luojiaxuan: These never run in CI. The CPU job skips them, and the accelerator job only selects `-m accelerator` (checked the 09-22 logs). Please add `@pytest.mark.accelerator`, like `test_tail.py`.

sglang_omni/models/dots_tts/vocoder_cuda_graph.py
```diff
@@ -0,0 +1,140 @@
...
+                            self.forward(*inputs)
+                    capture_stream.synchronize()
+                    with torch.cuda.graph(
+                        graph,
+                        pool=graph_pool,
+                        stream=capture_stream,
+                        capture_error_mode="thread_local",
+                    ):
+                        outputs = self.forward(*inputs)
+                except Exception as exc:
```
luojiaxuan: The flag is explicit, so a capture failure is a bug. Please let it raise at startup instead of warning and staying eager. This path is also untested.

sglang_omni/models/dots_tts/vocoder.py
```diff
@@ -377,8 +383,44 @@ def get_or_create_pool_locked(self) -> DotsVocoderSlotPool:
...
+                "got %s. Staying eager.",
+                self.codec.device,
+            )
+            runner = None
+        return runner
+
+    def cuda_graph_capture_keys(self) -> list[tuple[int, int]]:
+        return [
+            (batch, self.codec.patch_size * patches)
+            for batch in range(1, self.stream_chunk_batch_max + 1)
```
luojiaxuan: Coalesced steps only take 1 or `merge_steps` patches. The in-between frame counts only occur in the final drain, where B=1. So 6 of the 16 graphs (B>1 at 8/12 frames) never replay, about 6 s of startup. B should also be capped at `stream_slots`: with `max_running_requests=1`, only B=1 is reachable.

tests/unit_test/dots_tts/test_vocoder_cuda_graph.py
```diff
@@ -0,0 +1,198 @@
...
+        assert vocoder.ensure_slot_pool().graph_runner is None
+    assert "Staying eager" in caplog.text
+
+
+@requires_cuda
+def test_enabled_vocoder_captures_every_reachable_step_shape() -> None:
+    vocoder = make_streaming_vocoder(torch.device("cuda"), enable=True)
+    pool = vocoder.ensure_slot_pool()
+    assert pool.graph_runner is not None
+    assert pool.graph_runner.captured_keys == sorted(vocoder.cuda_graph_capture_keys())
```
luojiaxuan: Nit: this compares against the same function's output, so it doesn't check reachability. Assert the expected key list, or rename the test.

sglang_omni/models/dots_tts/vocoder_cuda_graph.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @property
+    def captured_keys(self) -> list[tuple[int, int]]:
+        return sorted(self.graphs)
+
+    @torch.no_grad()
+    def capture(self, keys: Iterable[tuple[int, int]]) -> None:
+        if self.device.type != "cuda":
+            raise RuntimeError("dots.tts vocoder CUDA graphs require a CUDA device")
+        # note (lennox): graphs share one mempool, so capture the largest key
```
luojiaxuan: Nit: growing the pool doesn't invalidate existing addresses. Capturing largest first just lets smaller graphs reuse the freed blocks.

sglang_omni/models/dots_tts/vocoder_cuda_graph.py
```diff
@@ -0,0 +1,140 @@
...
+    def run(
+        self,
+        packed: torch.Tensor,
+        hidden_h: torch.Tensor,
+        hidden_c: torch.Tensor,
+        window: torch.Tensor,
+        valid: torch.Tensor,
+    ) -> StepOutputs | None:
+        captured = self.graphs.get((int(packed.shape[0]), int(packed.shape[2])))
+        if captured is None or packed.device != self.device:
```
luojiaxuan: Nit: unreachable, since `packed` is always on the pool device.

sglang_omni/models/dots_tts/vocoder_slot_pool.py
```diff
@@ -78,12 +78,15 @@ def __init__(
         *,
         num_slots: int,
         chunk_size: int,
+        latent_dim: int,
+        graph_runner: Any | None = None,
```
luojiaxuan: Nit: no caller passes this. Could it be typed `DotsVocoderGraphRunner | None`?

Issue #1436, opened 2026-08-10
[Production Serving] V0.1.2 PR tracker


AkazaAkane:
This is the concise roadmap and PR tracker for the production-serving work under #1399.

Keep detailed investigation, benchmarks, and design decisions in #1399 or the corresponding sub-trackers; use this issue for status and ownership.

## Source issues

- Production serving / release gate: #1399
- Qwen3-TTS same-GPU MPS-DP: #1400
- `WEIGHT_SHARE=1` correctness: #1401

### Landed

- [x] **1.1** Qwen3-TTS clean-install / Transformers compatibility — #1407
- [x] **1.2** Release V0.1.1 to PyPI — #1414, #1416
- [x] **1.3** Cut V0.1.2 after production-serving gates close
- [x] **2.1–2.3** Qwen3-ASR automatic language detection — #1403
- [x] **3.1** Fix Qwen3-TTS admission / queueing cliff past c32 — #1449 
- [x] **3.2** Allow Qwen3-TTS / Higgs talker configuration above 32 requests — #1413
- [x] **3.3** Fix Qwen3-TTS playback continuity under load — #1529
- [x] **3.4** Restore Qwen3-TTS batch invariance #1475 
- [x] **4.1–4.3, 4.5** Qwen3-TTS MPS-DP config and docs — #1400
- [x] **4.5** `WEIGHT_SHARE=1` vocoder-stage fix and Qwen3-TTS validation — #1401 #1557

### Active

None

### Closed / superseded

- **4.4** Absolute KV-memory budget / fraction semantics — #1616, blocked by #1466 
- **1.1** #1406 — superseded by #1407
- **3.3** #1450— superseded by #1529
- **4.4** #1452 — superseded by #1616 

## Workstream checklist

- [x] Finish V0.1.1 release / packaging work
- [x] Qwen3-ASR automatic language detection
- [x] Fix and benchmark TTS overload behavior past the admission ceiling
- [x] Define and enforce TTS playback-continuity target
- [x] Restore Qwen3-TTS batch invariance
- [x] Finish Qwen3-TTS same-GPU MPS-DP documentation/configuration
- [x] Revalidate `WEIGHT_SHARE=1` correctness
- [ ] Resolve absolute KV-memory budgeting
- [x] Cut V0.1.2 once the #1399 release gates are closed

## Tracking rules

- Update this issue when a related PR opens, merges, closes, or changes scope.
[...]

PR #1449, opened 2026-08-10
[Perf] Qwen3-TTS: Bound waiting queue and fast-reject past admission ceiling (#1399 §3.1)

guozhihao-224:
## Summary
Fixes the Qwen3-TTS admission / queueing cliff past `max_running_requests` ([#1399](https://github.com/sgl-project/sglang-omni/issues/1399) §3.1 / [#1436](https://github.com/sgl-project/sglang-omni/issues/1436)).
- Normal Omni admission appended into an unbounded `waiting_queue`, bypassing upstream `_add_request_to_queue` / `_abort_on_queued_limit`. Past-ceiling load piled up as multi-second TTFA instead of failing fast.
- Enforce `max_queued_requests` in `OmniScheduler._enqueue_built_request`.
- Default Qwen3-TTS `max_queued_requests=16` (must be **≥ 1**: every request enters waiting before running).
- Add CLI `--max-queued-requests`.
- Map speech `"queue is full"` failures to HTTP **503**.
- Document the overload policy in `docs/cookbook/qwen3_tts.md`.
- Extend the standard SeedTTS bench with `--concurrencies` (generate-only) so sweeps can go past the admission ceiling.
## Root cause
Client TTFA at high concurrency was dominated by **scheduler queue wait**, not AR/vocoder compute.
Omni’s `_enqueue_built_request` path did not call `_abort_on_queued_limit` before `waiting_queue.append`, so raising offered load past `max_running_requests` kept admitting waiters into an unbounded FIFO.
## TTFA attribution method
Used the existing request-level profiler (same control plane as ASR profiling):
1. Serve 0.6B with `--max-running-requests 32` (pre-fix: unbounded waiting).
2. Around each concurrency pass: `POST /start_request_profile` → streaming SeedTTS generate-only → `POST /stop_request_profile`.
3. Summarize JSONL via `sglang_omni.profiler.views.build_report` / `benchmarks.eval.asr_profiling.build_stage_breakdown` on stage `tts_engine`.
4. Read these intervals (`docs/developer_reference/profiler.md`):

| Split | Interval | Meaning |
|---|---|---|
| **Queue wait** | `scheduler_queue_enter` → `scheduler_prefill_start` | Time in waiting queue after request build |
| **Engine → first emit** | `scheduler_prefill_start` → `scheduler_first_emit` | Work until first AR stream emission |

Helpers: `benchmarks.eval.asr_profiling.start_request_profile` / `stop_request_profile`. Attribution was investigation-only; durable follow-through is the fix + cookbook + `--concurrencies`.

### Pre-fix (H200, ceiling=32, unbounded waiting)

| conc | success | TTFA p95 | queue wait p95 |
|---:|---:|---:|---:|
| 32 | 64/64 | 0.53 s | 17 ms |
| 64 | 64/64 | 3.11 s | **2.61 s** |

Queue wait dominated; engine→first-emit stayed ~40–100 ms.
### Post-fix (H200, ceiling=32, `max_queued_requests=16`)

| conc | success | TTFA p95 | queue wait p95 |
|---:|---:|---:|---:|
| 32 | 49/64 | 0.45 s | 45 ms |
| 64 | 33/64 | 0.58 s | 75 ms |

Overload returns `503` `"The request queue is full."` instead of multi-second silent queueing. Some rejects at concurrency=32 under closed-loop load are expected once waiting depth fills. TTFA/latency stats are on successes only.
## Overload policy
Documented in `docs/cookbook/qwen3_tts.md`:
- Capacity ≈ `max_running_requests + max_queued_requests`
- Beyond that: fast-reject with HTTP **503** (do not admit and starve)
- Raising only `--max-running-requests` does **not** raise the waiting bound; set `--max-queued-requests` explicitly when needed
- For realtime serving, keep `max_queued_requests ≤ max_running_requests` unless a longer buffered spike is intentional
Past-ceiling sweep example:
```bash
  --generate-only --use-existing-server --stream \
  --port 8000 \
  --max-running-requests 32 \
  --max-queued-requests 16 \
  --concurrencies 16,32,48,64 \
  --max-samples 64
```

## Test plan

- [x] pytest tests/unit_test/pipeline/test_scheduler.py::test_enqueue_built_request_honors_max_queued_requests
- [x] pytest tests/unit_test/pipeline/test_scheduler.py::test_upstream_queue_limit_abort_is_translated_to_omni_output
- [x] pytest tests/unit_test/serve/test_speech_errors.py
- [x] pytest tests/unit_test/benchmarks/test_tts_seedtts_benchmark_config.py
- [x] H200 SeedTTS streaming sweep 16,32,48,64 with --max-running-requests 32 --max-queued-requests 16

## Notes / follow-ups
- max_queued_requests=0 rejects all admissions under upstream semantics; keep ≥ 1.
- This PR addresses §3.1 core: fix + explicit fast-reject policy + docs + standard bench past ceiling.
- Not in scope: §3.3 playback continuity, §3.4 batch invariance.
- Optional follow-up: a sustained-overshoot load generator (issue checklist item 1 used closed-loop stepped concurrency here; attribution already showed queue dominance).

## Related
- Closes / addresses: #1399 §3.1
- Tracker: #1436


sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -1035,6 +1035,17 @@ def _enqueue_built_request(
         def enqueue_if_live() -> None:
             if req_id in self._aborted_request_ids:
                 return
+            # note (guozhihao): normal Omni admission bypasses
+            # _add_request_to_queue; enforce max_queued_requests here.
+            if self._abort_on_queued_limit(req):
```
AkazaAkane, line 1202: Could we preserve the upstream priority prerequisites before calling _abort_on_queued_limit? Upstream _add_request_to_queue first runs _set_or_validate_priority(req), while this path calls the queue-limit helper directly. With priority scheduling enabled, a normal request with priority=None can therefore reach direction * recv_req.priority once the queue is full and raise instead of being admitted/rejected normally. A priority-enabled queue-full regression test would catch this as well.

benchmarks/eval/benchmark_tts_seedtts.py
```diff
@@ -479,6 +484,67 @@ def _parse_token_count(value: str) -> int | str:
...
+
+
+async def run_tts_concurrency_sweep(
+    config: TtsSeedttsBenchmarkConfig,
+    concurrencies: list[int],
+) -> dict[str, Any]:
+    """Run generate-only once per concurrency and write a summary JSON."""
+    rows: list[dict[str, Any]] = []
+    for concurrency in concurrencies:
+        point = replace(config, concurrency=concurrency)
```
AkazaAkane: Each sweep point keeps the same output_dir, but run_tts_seedtts_benchmark writes fixed speed_results.json, results.csv, generated.json, and sample audio paths. That means later concurrency points overwrite the detailed artifacts from earlier ones, leaving only their aggregate summaries in concurrency_sweep.json. Could each point use a per-concurrency output directory so the overload sweep remains inspectable/reproducible?

AkazaAkane (commented):
Can you add a validation of sustains overshoot rather than stepping concurrency after address comments? Thanks.

benchmarks/eval/benchmark_tts_seedtts.py
guozhihao-224: fixed in 74f14369cca976b86156855e9d0bff1270e8e67f, Each sweep point now writes to `<output-dir>/c<N>/` (`speed_results.json`, csv, audio). The summary JSON still lands at `<output-dir>/concurrency_sweep.json`. Sustained overshoot uses `<output-dir>/overshoot/.

sglang_omni/scheduling/omni_scheduler.py
guozhihao-224: fixed in 74f14369cca976b86156855e9d0bff1270e8e67f, `_enqueue_built_request` now matches upstream `_add_request_to_queue`: `_set_or_validate_priority(req)` runs first, then `_abort_on_queued_limit`.
`test_enqueue_built_request_honors_max_queued_requests` enables priority scheduling, admits a `priority=None` request, and asserts both requests get a non-None priority before the second is rejected with 503.

guozhihao-224:
> Can you add a validation of sustains overshoot rather than stepping concurrency after address comments? Thanks.

@AkazaAkane  Done. `--sustained-overshoot` is now the primary overload check: open-loop at 2× (max_running_requests + max_queued_requests) for `--overshoot-duration-s` (default 10s). It is mutually exclusive with `--concurrencies`.
On H200 / 0.6B / running=32 queued=16 (cap=48), 96 req/s × 10s, n=960:
- 97 success / 863 × 503 / 0 other
- 503 p50 1.2ms (793/863 < 10ms); p95 839ms from a small tail
- admitted TTFA p95 4.43s (the in-flight set under overload, not a silent FIFO)
`--concurrencies` remains as a closed-loop contrast; each point writes `<output-dir>/c<N>/`, soak writes `<output-dir>/overshoot/`.

guozhihao-224:
Design and numbers below (H200, current code).

**Policy.** Capacity = `max_running_requests + max_queued_requests`. Extra arrivals get HTTP 503 `The request queue is full.` — not an unbounded FIFO.
Enforced at three points, same public message:

1. Coordinator `max_in_flight = running + queued` — reject before preprocessing.
2. Scheduler skip-build when occupancy (waiting + pending builds + backlog + deferred) is full.
3. Enqueue: `_set_or_validate_priority` then `_abort_on_queued_limit`.

Cap is resolved from `generation_admission_defaults()` + stage/CLI overlays, only when both bounds are set. Qwen3-TTS defaults 16+16 (CLI 32+16 → cap 48). Fun-ASR / Higgs stay uncapped unless both are set.

**Setup.** `Qwen/Qwen3-TTS-12Hz-0.6B-Base`, `--max-running-requests 32 --max-queued-requests 16 --cuda-graph-max-bs 32`, SeedTTS generate-only stream.

**Pre-fix** (unbounded waiting, closed-loop): c32 64/64 TTFA p95 0.53s (queue wait 17ms) → c64 64/64 TTFA p95 3.11s (queue wait **2.61s**).

**Post-fix closed-loop** `--concurrencies` (artifacts `c<N>/`):

| conc | success | failed | TTFA p95 |
  |------|---------|--------|----------|
  | 16   | 64      | 0      | 0.45s    |
  | 32   | 61      | 3      | 0.39s    |
  | 48   | 26      | 38     | 0.49s    |
  | 64   | 18      | 46     | 0.38s    |

No c32→c64 TTFA cliff. Past ceiling, extras 503; admitted requests stay ~0.4s.

**Post-fix soak** `--sustained-overshoot` (96 req/s × 10s, n=960, artifacts `overshoot/`):
- 97 success / 863 × 503 / 0 other
- 503 p50 **1.2ms** (793/863 < 10ms); p95 839ms is a small tail
- success TTFA p95 4.43s = the in-flight set under 2× load, not silent queueing of all 960

`--sustained-overshoot` cannot be combined with `--concurrencies`.

benchmarks/benchmarker/runner.py
```diff
@@ -47,7 +47,15 @@ def __init__(self, config: RunConfig) -> None:
 
     async def run(self, samples: list, send_fn: SendFn) -> list[RequestResult]:
         timeout = aiohttp.ClientTimeout(total=self.config.timeout_s)
-        async with aiohttp.ClientSession(timeout=timeout) as session:
+        # Closed-loop runs are bounded by max_concurrency. Open-loop
```
AkazaAkane: should be note(yourname):

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -925,6 +950,19 @@ def _stage_request_build_payloads(
             backlog = self._backlogged_request_build_payloads
             pending_builds = self._pending_request_builds
             rejected: list[Any] = []
+            if self._waiting_queue_is_full():
```
edwingao28: _queued_admission_count() already includes backlog, so equality here means those requests are inside the configured bound.

benchmarks/benchmarker/runner.py
guozhihao-224: fixed in https://github.com/sgl-project/sglang-omni/pull/1449/commits/5362680589c215cbb392662a398c9d6b1daa4dd3

AkazaAkane (approved):
Capacity is max_running_requests + max_queued_requests under sustained load. During bursts from idle, in-flight admission state counts against the queued bound, so the effective ceiling can be as low as max_queued_requests until running slots fill. Size --max-queued-requests accordingly if your traffic is bursty.

PR #1450, opened 2026-08-10
[Perf] Qwen3-TTS: Raise default initial codec chunk for playback continuity (#1399 §3.3)

guozhihao-224:
## Summary
- Fixes Qwen3-TTS streaming playback underrun under concurrency ([#1399](https://github.com/sgl-project/sglang-omni/issues/1399) §3.3 / [#1436](https://github.com/sgl-project/sglang-omni/issues/1436)).
- Default first vocoder chunk `initial_codec_chunk_frames` raised from `1` (~80 ms) to `8` (~667 ms @ 12 Hz), so the buffer can cover the gap to the first follow-up chunk.
- Adds client-side playback continuity metrics: per-chunk audio duration, max underrun, and C50 / C100 / C200 pass rates in the SeedTTS stream speed path.
- Documents the Qwen3-TTS default and updates unit coverage for the new default and underrun helpers.
## Root cause
With the previous default of 1 codec frame, the first PCM payload is only ~0.08 s. Under load the first follow-up often arrives hundreds of milliseconds later, so an immediate-play client underruns on the first seam. Mid-stream vocoder / AR / handoff were secondary once the first chunk was large enough.
## Validation (H200, 0.6B Base, SeedTTS EN, 64 samples, stream)
Request override sweep (existing server):
| init | c | C50 | C100 | C200 | TTFA p95 | QPS |
| 1 | 32 | 0% | 0% | 3% | 0.65s | 9.6 |
| 4 | 32 | 94% | 97% | 97% | 0.83s | 10.3 |
| 8 | 32 | 100% | 100% | 100% | 0.85s | 11.1 |

Default path after this change (no request override):

| c | C50 | C100 | C200 | first payload |
|---:|---:|---:|---:|---:|
| 16 | 96.9% | 100% | 100% | 30720 B (~0.64 s) |
| 32 | 98.4% | 98.4% | 98.4% | 30720 B (~0.64 s) |

Meets the proposed V0.1.2 gate: C50 ≥ 90% @ c32.

## Test plan
- [x] `pytest tests/unit_test/benchmarks/test_playback_continuity.py`
- [x] 'pytest tests/unit_test/qwen3_tts/test_pipeline.py::test_qwen3_tts_streaming_vocoder_default_initial_chunk_is_continuity_safe`
- [x] H200 SeedTTS stream c16/c32 with default (no `--initial-codec-chunk-frames`)

## Notes / follow-ups
- End-to-end C50@c32 CI gate for Qwen3-TTS is left as a follow-up (current TTS CI matrix is mainly Higgs/MOSS).
- Full per-stage profiler attribution for mid-stream stalls is deferred; the controlled init sweep showed the first-seam underrun dominates under the old default.

## Related
- Addresses: #1399 §3.3
- Tracker: #1436
- Continuity metric background: #1234


benchmarks/metrics/playback_continuity.py
```diff
@@ -0,0 +1,82 @@
...
+    if threshold_s < 0:
+        raise ValueError("threshold_s must be non-negative")
+    scored = [value for value in max_underruns_s if value is not None]
+    if not scored:
+        return None
+    passed = sum(1 for value in scored if value <= threshold_s)
+    return passed / len(scored)
+
+
+def summarize_playback_continuity(
```
AkazaAkane, line 53: Could we avoid adding playback-continuity fields for workloads that do not produce streaming audio? compute_speed_metrics is shared by ASR/Omni benchmarks, so this currently makes every non-TTS successful request show up as playback_continuity_na_requests and adds c50/c100/c200=None to unrelated benchmark results.

sglang_omni/models/qwen3_tts/streaming_vocoder.py
```diff
@@ -29,7 +29,7 @@
 DEFAULT_QWEN3_TTS_STREAM_STRIDE = 16
 DEFAULT_QWEN3_TTS_STREAM_FOLLOWUP_STRIDE = 8
 DEFAULT_QWEN3_TTS_STREAM_INITIAL_FOLLOWUP_STRIDE = 8
-DEFAULT_QWEN3_TTS_INITIAL_CHUNK_FRAMES = 1
+DEFAULT_QWEN3_TTS_INITIAL_CHUNK_FRAMES = 8
```
AkazaAkane, line 32: Does this regress short utterances?

With the initial threshold raised to 8, a request that reaches EOS before 8 generated codec frames will not emit any streaming audio before the final flush. That also makes the request single-chunk, so the new continuity metric marks it N/A and excludes it from C50/C100/C200.

Could we add a short-utterance validation/test here? Otherwise the population most affected by this default change is invisible to the proposed release gate.

AkazaAkane:
closes as #1529 landed

PR #1452, opened 2026-08-10
[Feature] Replace Fractional KV Memory Settings with Absolute Per-Stage Budgets

adityavaid:
## Motivation
This change replaces ambiguous fractional GPU-memory configuration for SGLang-backed stages with explicit byte budgets.  Pointed out in section 4.4 of https://github.com/sgl-project/sglang-omni/issues/1399
It makes same-GPU MPS data parallelism predictable: each replica receives a declared KV-cache allocation and the launcher rejects an over-subscribed layout before it creates MPS state or starts a process.

The primary configuration is:

```yaml
runtime:
  memory:
    kv_cache_bytes: 2GiB
    total_reserve_bytes: 9GiB
```

`total_reserve_bytes` is the complete per-replica physical-VRAM reservation, including `kv_cache_bytes`.

## Modifications

- Adds `StageMemoryConfig.kv_cache_bytes`, accepting integer bytes or exact
  binary-size strings (`KiB`, `MiB`, `GiB`, and `TiB`).
- Adds `StageMemoryConfig.total_reserve_bytes`, using the same size syntax.
- Validates `kv_cache_bytes <= total_reserve_bytes` when both are specified.
- Retains `runtime.resources.total_gpu_memory_fraction` and
  `runtime.sglang_server_args.mem_fraction_static` for compatibility, but
  emits deprecation warnings.
- Rejects a stage that mixes `kv_cache_bytes` with either legacy fraction
  input.
  
  
### MPS Changes
 `examples/mps_dp/config.py` now resolves the generation-stage memory budget,
discovers the selected GPU's physical VRAM, and validates:

```text
replica_count * total_reserve_bytes <= physical_vram
```

Additional behavior:

- Requires a positive `kv_cache_bytes` for byte-budget preflight.
- Defaults an omitted `total_reserve_bytes` to the selected card's full
  physical VRAM at preflight time. Consequently, DP2+ configurations must
  declare a smaller explicit total reservation.
- Produces an arithmetic error containing replica count, KV allocation,
  per-replica reservation, requested total, and available VRAM.
- Exposes `--print-mps-memory-budget` to provide key/value budget metadata.

`examples/mps_dp/launch.sh` now:

- Resolves the target physical GPU UUID before preflight.
- Runs the config helper with `CUDA_VISIBLE_DEVICES` restricted to that UUID
  and queries logical GPU `0`, preventing inherited visibility masks from
  validating a different card.
- Performs memory preflight before creating run state, launching the MPS
  daemon, or starting replicas.
- Persists resolved budget values in the run manifest.

Requires a positive kv_cache_bytes for byte-budget preflight.
Defaults an omitted total_reserve_bytes to the selected card's full physical VRAM at preflight time. Consequently, DP2+ configurations must declare a smaller explicit total reservation.
Produces an arithmetic error containing replica count, KV allocation, per-replica reservation, requested total, and available VRAM.
Exposes --print-mps-memory-budget to provide key/value budget metadata.
examples/mps_dp/launch.sh now:

Resolves the target physical GPU UUID before preflight.
Runs the config helper with CUDA_VISIBLE_DEVICES restricted to that UUID and queries logical GPU 0, preventing inherited visibility masks from validating a different card.
Performs memory preflight before creating run state, launching the MPS daemon, or starting replicas.
Persists resolved budget values in the run manifest.

## Related Issues
N/A

## Accuracy Test
- Adds schema tests for parsing, invalid values, legacy compatibility, and the
  KV-less-than-or-equal-to-total-reservation invariant.
- Adds MPS preflight tests for exact arithmetic, oversubscription, missing GPU
  metadata, physical-VRAM defaulting, and invalid KV/total combinations.
- Updates launcher-support stubs and tests to use `total_reserve_bytes`.
- Adds runner and engine-builder coverage for KV-budget propagation.

## Benchmark & Profiling
N/A

## Checklist

- [ ] Add unit tests.

## CI


adityavaid:
Current Status :  
- Updating support for rest of the models
- testing for backward compatibility

adityavaid:
Update : Ready for review

examples/mps_dp/config.py
```diff
@@ -77,6 +76,143 @@ def resolve_max_total_tokens(
...
+    total_memory_bytes = device_info.total_memory_bytes
+    if total_reserve_bytes is None:
+        total_reserve_bytes = total_memory_bytes
+    if kv_cache_bytes > total_reserve_bytes:
+        raise ValueError(
+            f"{config_type.__name__} generation stage {stage.name!r} declares "
+            "runtime.memory.kv_cache_bytes greater than "
+            "runtime.memory.total_reserve_bytes"
+        )
+    requested_total_bytes = replicas * total_reserve_bytes
```
AkazaAkane: total_reserve_bytes currently appears to be a declared estimate rather than an enforced reservation. The preflight only checks the declared totals against physical VRAM, while the runtime receives only the KV budget. A replica whose weights, graphs, activations, and KV exceed its declared reserve can therefore pass this check and still oversubscribe the card. Can we either validate the actual per-replica footprint against this budget or make the weaker semantics explicit?

sglang_omni/model_runner/sglang_model_runner.py
```diff
@@ -49,6 +49,7 @@ class _OmniKVCacheConfigurator(KVCacheConfigurator):
     """
 
     total_gpu_memory_fraction: float | None = None
+    kv_cache_bytes: int | None = None
 
     def _profile_available_bytes(self, pre_model_load_memory: float) -> int:
```
AkazaAkane, line 55: Can we verify the byte budget remains authoritative through post-capture KV sizing? SGLang 0.5.16 recomputes the post-capture pool from free memory and mem_fraction_static, while several builders still set their existing fraction defaults. Otherwise the initial pool is byte-budgeted but the final pool can still depend on the legacy fraction semantics.

sglang_omni/models/arkasr/stages.py
```diff
@@ -36,6 +37,7 @@ def create_sglang_arkasr_executor(
         mm_attention_backend=mm_attention_backend,
         request_build_max_workers=request_build_max_workers,
         request_build_max_pending=request_build_max_pending,
+        kv_cache_bytes=kv_cache_bytes,
```
JiaxinD: ArkasrEngineBuilder.__init__ does not accept kv_cache_bytes and never calls super().__init__, so every ARK ASR launch raises TypeError here.

sglang_omni/models/moss_tts/stages.py
```diff
@@ -502,13 +502,15 @@ def create_sglang_tts_engine_executor(
...
     dtype: str = "bfloat16",
+    kv_cache_bytes: int | None = None,
     server_args_overrides: dict[str, Any] | None = None,
 ) -> Any:
     return MossTtsEngineBuilder().build(
         model_path,
         device=device,
         gpu_id=gpu_id,
         dtype=dtype,
+        kv_cache_bytes=kv_cache_bytes,
```
JiaxinD: Base build() has no kv_cache_bytes parameter, so every moss_tts launch raises TypeError, including the shipped moss_delay_h100_dp2.yaml. Other builders take it in the constructor.

sglang_omni/models/fishaudio_s2_pro/stages.py
```diff
@@ -320,6 +320,7 @@ def create_sglang_tts_engine_executor(
     return FishS2ProEngineBuilder(
         max_new_tokens=max_new_tokens,
         ras_window=ras_window,
+        kv_cache_bytes=kv_cache_bytes,
```
JiaxinD, line 324: kv_cache_bytes is not in this factory's signature, so every Fish launch raises NameError.

tests/unit_test/serve/test_sglang_bootstrap.py
```diff
@@ -150,6 +190,91 @@ def __init__(self, **kwargs) -> None:
     assert infrastructure[0].model_runner.model is FakeRunner.model
 
 
+def test_resolved_factory_kv_budget_reaches_runner_initialization(
```
JiaxinD: StageConfig, StageRuntimeConfig, StageMemoryConfig, PipelineConfig and resolve_stage_factory_args are never imported in this file, so this test dies with NameError.

sglang_omni/model_runner/sglang_model_runner.py
```diff
@@ -67,6 +68,11 @@ def _profile_available_bytes(self, pre_model_load_memory: float) -> int:
         upstream SGLang profiling semantics for ordinary non-colocated AR
         serving.
         """
+        if self.kv_cache_bytes is not None:
```
JiaxinD, line 72: The new slot breaks tests/unit_test/qwen3_omni/test_sglang_ar_budget.py: its fixture builds the configurator via __new__ and never sets kv_cache_bytes, so this read raises AttributeError.

examples/mps_dp/config.py
```diff
@@ -77,6 +76,143 @@ def resolve_max_total_tokens(
...
+    if device_info.total_memory_bytes is None:
+        name = device_info.name or "unknown GPU"
+        raise ValueError(
+            f"GPU {gpu_id} ({name}) total VRAM metadata is unavailable; cannot "
+            "preflight runtime.memory.kv_cache_bytes"
+        )
+
+    total_memory_bytes = device_info.total_memory_bytes
+    if total_reserve_bytes is None:
+        total_reserve_bytes = total_memory_bytes
```
JiaxinD: Defaulting to full VRAM makes any kv-only config fail preflight for replicas >= 2, and the error then cites a total_reserve_bytes the user never set. Either require it explicitly for N >= 2 or pick a saner default.

examples/mps_dp/config.py
JiaxinD: This arithmetic ignores WEIGHT_SHARE=1: weights are resident once under CUDA IPC, so valid layouts get charged (N-1) extra weight copies and are rejected.

sglang_omni/scheduling/engine_factory.py
```diff
@@ -82,8 +85,10 @@ def build(
         )
         self.customize_server_args(server_args)
         self.validate_before_infrastructure(server_args)
-
         infra_kwargs = dict(self.infra_kwargs())
+        infra_kwargs = dict(self.infra_kwargs())
```
JiaxinD, line 109: Duplicated line.

sglang_omni/config/schema.py
```diff
@@ -103,6 +180,18 @@ class StageRuntimeConfig(BaseModel):
     )
 
     def model_post_init(self, __context: Any = None) -> None:
+        kv_cache_bytes = self.memory.kv_cache_bytes
+        if kv_cache_bytes is not None:
+            if self.resources.total_gpu_memory_fraction is not None:
+                raise ValueError(
+                    "runtime.memory.kv_cache_bytes cannot be set together with "
```
JiaxinD, line 197: The MF override in cli/serve.py mutates mem_fraction_static after construction, so it bypasses this check and is then silently ignored at runtime (kv_cache_bytes wins in the configurator).

JiaxinD (changes requested):
Direction is right (#1399 section 4.4) and injecting the byte budget at the KV configurator level is the correct mechanism. But this cannot merge as is:

* arkasr, fishaudio and moss_tts engine builds all crash at startup on the new kwarg (see inline), and the shipped moss_delay_h100_dp2.yaml breaks with them
* two test files fail before testing anything (missing imports here, stale slots fixture in test_sglang_ar_budget.py), please run the full unit suite and paste the result
* colocated layouts still require total_gpu_memory_fraction in topology.py, which the schema forbids combining with kv_cache_bytes, so multi-stage-per-GPU configs cannot migrate to the new field
* branch currently conflicts with main, please rebase

examples/mps_dp/config.py
adityavaid: Made the weaker semantics explicit rather than enforcing a footprint for now , not sure how we can measure weights/graphs/activations before launch. open to suggestions here

sglang_omni/model_runner/sglang_model_runner.py
adityavaid: Verified against SGLang 0.5.16 source.  no logic change needed.

Got this from Claude : 
```
 `compute_post_capture_kv_resize` does bypass,  `_profile_available_bytes`, but it calls config_from_budget `(...,  cap_tokens=max_total_num_tokens)`, which clamps via min(max_tokens, cap_tokens). So  kv_cache_bytes is a ceiling that always holds — the pool can shrink, never exceed
 ```
 
 also verified this , plus also gated on SGLANG_ENABLE_POST_CAPTURE_KV_SIZING, which is  False by default.

File: sglang_omni/models/arkasr/stages.py (hunk 3 above)
Comment (adityavaid) [reply sglang_omni/models/arkasr/stages.py]: Fixed. Tried to do a rebase on GitHub web editor and thus missed it, my mistake. Ran all pipeline tests to ensure nothing is failing now.

File: sglang_omni/models/moss_tts/stages.py (hunk 4 above)
Comment (adityavaid) [reply sglang_omni/models/moss_tts/stages.py]: Fixed. same reason as above

File: sglang_omni/models/fishaudio_s2_pro/stages.py (hunk 5 above)
Comment (adityavaid) [reply sglang_omni/models/fishaudio_s2_pro/stages.py:324]: fixed.

File: tests/unit_test/serve/test_sglang_bootstrap.py (hunk 6 above)
Comment (adityavaid) [reply tests/unit_test/serve/test_sglang_bootstrap.py]: verified and Fixed, plus one more we didn't list, sglang_backend was also undefined.

File: sglang_omni/scheduling/engine_factory.py (hunk 9 above)
Comment (adityavaid) [reply sglang_omni/scheduling/engine_factory.py:109]: rebase issue , fixed now

File: sglang_omni/model_runner/sglang_model_runner.py (hunk 7 above)
Comment (adityavaid) [reply sglang_omni/model_runner/sglang_model_runner.py:72]: Fixed, class-level default and fixture sets it .

File: examples/mps_dp/config.py (hunk 1 above)
Comment (adityavaid) [reply examples/mps_dp/config.py]: Fixed,  per-replica check & explicit warning now

File: sglang_omni/config/schema.py (hunk 10 above)
Comment (adityavaid) [reply sglang_omni/config/schema.py:197]: Fixed.  apply_mem_fraction_cli_overrides  now validates every resolved target against  runtime.memory.kv_cache_bytes  before mutating any stage, so override raises  `typer.BadParameter`  instead of being silently dropped. 
Also added regression coverage in  tests/unit_test/qwen3_omni/test_pipeline.py  for the global flag, the per-role flag, and cross-stage atomicity.

**Comment by adityavaid** [comment]
Update : Some Unit tests were failing after rebase with latest commits. Currently working on fixing those without mocks

**Comment by JiaxinD** [comment]
Needs a rebase, the branch is currently conflicting with main. Also GPU CI has never run on this PR: the run-ci label is missing, add it or use /tag-and-rerun-ci after the rebase.

**Comment by adityavaid** [comment]
> Needs a rebase, the branch is currently conflicting with main. Also GPU CI has never run on this PR: the run-ci label is missing, add it or use /tag-and-rerun-ci after the rebase.

@JiaxinD done. I don't have permission to add labels though.

File: sglang_omni/model_runner/sglang_model_runner.py (hunk 7 above)
Comment (AkazaAkane) [reply sglang_omni/model_runner/sglang_model_runner.py:72]: This returns the declared value unchanged — pre_model_load_memory is ignored and device capacity is never consulted. kv_cache_bytes: 200GiB on an H100 is accepted at config time and fails at allocation. The only physical-capacity validation in the PR lives in examples/mps_dp/config.py, so sgl-omni serve outside the MPS launcher gets none of it, which is the third checkbox in §4.4 ("validate declared budgets against physical card capacity in preflight, so an over-subscribed config fails immediately with an arithmetic explanation").

A clamp against measured free memory here — with the same arithmetic error format used in the MPS helper — would cover both the MPS and the plain-serve paths.

File: examples/mps_dp/config.py [hunk 11]
```diff
@@ -77,6 +77,180 @@ def resolve_max_total_tokens(
...
+            f"greater than the {replicas}-way even split of GPU {gpu_id} "
+            f"({format_bytes_gib(total_reserve_bytes)}). Lower kv_cache_bytes, "
+            "reduce the replica count, or set runtime.memory.total_reserve_bytes "
+            "explicitly."
+        )
+
+    if weight_share:
+        requested_total_bytes = total_reserve_bytes
+    else:
+        requested_total_bytes = replicas * total_reserve_bytes
```
AkazaAkane, line 147: Shared weights don't imply a shared KV pool, but this branch drops the replica multiplier for the whole reservation. On a 16 GiB card with replicas=2, kv_cache_bytes=10GiB, total_reserve_bytes=12GiB, WEIGHT_SHARE=1: 10 ≤ 12 clears the kv-vs-reserve gate, 12 ≤ 16 clears capacity, and preflight approves a layout whose two private KV pools need 20 GiB.

The warning below is honest that the N-way total isn't validated, but the unvalidated component should be the shared weights, not KV. Minimum fix: always enforce replicas * kv_cache_bytes <= physical_vram regardless of weight-share, then keep a weaker warning for the unknown shared-weight/private-overhead remainder.

sglang_omni/config/topology.py
```diff
@@ -226,6 +226,20 @@ def _stage_gpu_ids(
     return resolve_stage_gpu_ids(gpu_placement, stage)
 
 
+def _stage_lacks_declared_memory_budget(stage: StageConfig) -> bool:
```
AkazaAkane, line 229: This helper's docstring says kv_cache_bytes is "the byte-budget replacement for total_gpu_memory_fraction," but the two quantities aren't substitutable in this gate. The fraction path resolves to total × fraction − stage_used — a whole-stage envelope, weights inclusive. kv_cache_bytes is returned verbatim as KV headroom in _OmniKVCacheConfigurator._profile_available_bytes (sglang_model_runner.py:72-75) and says nothing about weights, CUDA graphs, activations, or encoder memory.

Accepting it here relaxes the colocation gate to a component of the footprint, and the check that follows at line 287 still sums only fractions. So a two-process layout with fraction total == 0, two small KV declarations, and arbitrarily large model footprints now passes validation that previously required a total-memory envelope.

Ask: either require total_reserve_bytes for byte-budgeted stages participating in cross-process colocation and sum those per physical GPU, or keep requiring total_gpu_memory_fraction here until an enforceable byte replacement lands.

sglang_omni/config/placement.py
```diff
@@ -28,6 +29,8 @@ class GpuPlacement:
     total_gpu_memory_fraction: float
     has_memory_fraction: bool
     missing_fraction_stage_names: tuple[str, ...]
+    total_kv_cache_bytes: int
```
AkazaAkane, line 32: This sum is computed and surfaced in launcher._placement_log_summary, but nothing reads it as a constraint — _validate_gpu_process_colocation still compares only total_gpu_memory_fraction against the limit. If the intent is that byte budgets participate in colocation safety, this is where the enforcement is missing.

AkazaAkane (changes requested):

sglang_omni/model_runner/sglang_model_runner.py
adityavaid: Hi @AkazaAkane , 
I checked the SGLang path again.  mem_fraction_static  is documented against GPU capacity, but KV allocation still depends on available memory after model load.

I’d prefer to keep  kv_cache_bytes  exact and fail rather than clamp it to current free VRAM.

 For a temporary solution, would this be acceptable: 
preflight physical/available VRAM for diagnostics, then apply  total_reserve_bytes / physical_vram  as a PyTorch per-process allocator cap before model load?
 It is deterministic for PyTorch-managed allocations, while native SGLang absolute-byte accounting remains the long-term fix.

Does that match what you had in mind, or is there a better SGLang-native mechanism we should use?

PR #1394, opened 2026-08-07
[ASR Refactor] Extract shared speech-to-text serving helpers

lijrjyan:
## Motivation

M5 (#1285) consolidated transcription serving into `serve/transcriptions.py`, but the reusable mechanics are still fused with the endpoint's identity, down to a hard-coded `task="transcribe"` in request construction. `/v1/audio/translations` (#1267 F2) needs the same flow with a different task, so this behavior-preserving refactor extracts the mechanics first; the translations endpoint follows as a separate PR built on top.

## Modifications

- `serve/speech_to_text.py` (new): endpoint-neutral form parsing/validation, `build_speech_to_text_generate_request(..., task="transcribe")`, dispatch + error mapping, json/verbose_json/text serializers, SSE lifecycle, adapter resolution, duration probe. Old builder name kept as a compatibility alias.
- `serve/transcriptions.py`: thin endpoint layer; existing import paths re-exported.
- New characterization tests; `tests/README.md` registered.

## Validation

- Existing transcription tests (incl. all TestClient HTTP/SSE paths) pass with zero assertion changes.
- Full unit tree (2,200 tests): failure surface identical to clean main.
- Five-model A6000 acceptance (frozen `b7e77625f` vs `520486a2d`): 35 sequential response bodies byte-identical; warm-latency deltas <10%; concurrency-4 green on 4/5 models. Whisper concurrency-4 fails identically on base and head — pre-existing scheduler defect, tracked separately, deliberately not fixed here.
- ruff / format / isort / `git diff --check` clean.

## Related

#1267 (F2). The `/v1/audio/translations` endpoint follows as a separate PR on top of this seam.

PR #1453, opened 2026-08-10
[ASR Feature] Add /v1/audio/translations with per-model capability gating

lijrjyan:
## Motivation

Adds the OpenAI-compatible `/v1/audio/translations` endpoint (F2 in the ASR roadmap #1267), stacked on the shared speech-to-text serving helpers (#1394), with docs and tests updated.

## Modifications

- **Capability config** (`config/schema.py`, `models/whisper_asr/config.py`): new `PipelineConfig.supports_audio_translation()`, default **False**; only `whisper_asr` overrides True. Threaded launcher -> create_app -> app.state; runtime overrides cannot force it on.
- **Route** (`serve/translations.py`, registered in `serve/openai_api.py`): reuses the shared speech-to-text helpers with `task="translate"`; unsupported models get an explicit OpenAI-shaped 400 instead of a silent transcription.
- **response_format**: `json`/`verbose_json`/`text` implemented (`text` returns `text/plain`); `srt`/`vtt` -> 400 expressed as missing segment-timestamp capability; out-of-vocabulary values -> same OpenAI-shaped 400. SSE streaming shares the transcriptions lifecycle.
- **Tests**: two new suites — route behavior (`test_translations.py`) and capability chain (`test_translation_capability.py`).
- **Docs**: per-model support matrix and cookbook updates; the `language` field is documented as a SGLang-Omni extension.

## Related Issues

#1267 (F2). Stacked on #1394; the diff includes its commits until it merges, after which this PR rebases to the endpoint-only diff.

## Accuracy Test

- Focused unit suites: **91 passed**; full `tests/unit_test/serve/`: only the pre-existing `test_mps_dp_support` failures.
- A6000 GPU smoke (whisper-large-v3): **11/11 pass** — the same zh clip returns English via translations and Chinese via transcriptions; response formats, SSE, and all rejection paths behave as documented.
- GPU CI does not include Whisper; the smoke above is the runtime evidence.

## Benchmark & Profiling

Not applicable — serving-layer endpoint addition; no performance-sensitive path is modified.

## Checklist

- [x] Add unit tests.

## CI


SandyLuXY:
Hi @lijrjyan, could you rebase on main and resolve conflicts? Thanks!

lijrjyan:
Done! I can see no conflicts right now.

sglang_omni/serve/translations.py
```diff
@@ -0,0 +1,149 @@
...
+                request_id=request_id,
+                error_log_message="Error translating audio for request %s",
+            )
+        except HTTPException as exc:
+            return _http_exception_response(exc, param=None)
+
+        # note (Junnan Li): verbose_json keeps transcription parity: with no
+        # segment-timestamp channel, the shared adapter emits one placeholder
+        # segment.
+        return speech_to_text.assemble_speech_to_text_response(
```
0xjeffro, line 157: here we call `assemble_speech_to_text_response` without pass `duration_s` so will compute this in speech_to_text.py and this may somewhat blocking I/O on the event loop.

I prefer following our impl in transcriptions.py like, probes once up front with asyncio.to_thread and passes it through.

lijrjyan: Good catch — my branch predated the chunking changes and I should have caught this pattern when rebasing. Fixed in ed75d4b: duration is probed once with asyncio.to_thread and passed through, matching transcriptions.

0xjeffro (approved):

sglang_omni/models/whisper_asr/config.py
```diff
@@ -23,6 +23,12 @@ def mem_fraction_role_to_stage(cls) -> dict[str, str]:
     def generation_sglang_role_to_stage(cls) -> dict[str, str]:
         return {"generation": "asr"}
 
+    def supports_audio_translation(self) -> bool:
+        """Multilingual non-turbo Whisper checkpoints translate speech to
+        English; `*.en` and turbo checkpoints do not, which is documented
+        rather than gated, matching how sibling engines handle it."""
```
SandyLuXY: Please remove this

sglang_omni/serve/translations.py
```diff
@@ -0,0 +1,158 @@
...
+        duration_s = await asyncio.to_thread(
+            speech_to_text.probe_audio_duration, audio_bytes
+        )
+
+        gen_req = speech_to_text.build_speech_to_text_generate_request(
+            audio_bytes=audio_bytes,
+            filename=form.file.filename,
+            content_type=form.file.content_type,
+            model=model,
+            language=form.language,
```
SandyLuXY: Omitting `language` is the standard OpenAI translations request, but this reaches `_resolve_language(None)`, which forces Whisper's source-language prefix to English. Seems non-English audio sent through an official client can be decoded under the wrong language?

sglang_omni/serve/translations.py
```diff
@@ -0,0 +1,158 @@
...
+
+        if not app.state.supports_audio_translation:
+            return _invalid_request(
+                f"Model {model!r} does not support {TRANSLATIONS_ENDPOINT}; "
+                "use /v1/audio/transcriptions instead.",
+                param="model",
+            )
+
+        normalized_response_format = form.response_format.strip().lower()
+        if normalized_response_format in _SEGMENT_FORMATS:
```
SandyLuXY, line 77: Shall we consider supporting srt/vtt? Could do that in a follow-up PR as it requires segment timestamp support. @0xjeffro WDYT?

lijrjyan: I think we should and I actually already  implemented it. this feature  will be two stack prs and stacked on this pr. I will post them today.

sglang_omni/serve/translations.py
lijrjyan: Fixed! An omitted language is now auto-detected in one greedy step rather than defaulting to English. If `language` is selected, we could still skip the detect step and save about 0.2s in each chunk.

sglang_omni/models/whisper_asr/config.py
lijrjyan: Done. Already removed.

sglang_omni/serve/translations.py
lijrjyan: PR posted. #1526 adds `srt/vtt` on `/v1/audio/transcriptions`

sglang_omni/models/whisper_asr/request_builders.py
```diff
@@ -128,7 +133,17 @@ def make_whisper_scheduler_adapters(
...
+    # sit above the base vocabulary, so the sampling vocab must cover them.
+    vocab_size = len(tokenizer)
+    sot_token_id = int(tokenizer.convert_tokens_to_ids("<|startoftranscript|>"))
+    # note (Junnan Li): a uniform positive bias keeps the relative order of the
+    # language tokens while guaranteeing one of them wins the detection step.
+    language_token_bias = {
+        str(int(token_id)): 100.0
+        for token, token_id in tokenizer.get_added_vocab().items()
+        if _LANGUAGE_TOKEN_RE.match(token)
+    }
```
0xjeffro, line 144: can we directly use this instead of regular expression?

```py
language_token_bias = {
    str(int(token_id)): 100.0
    for token_id in generation_config.lang_to_id.values()
}
```

lijrjyan: Done. use direct `lang_to_id` here and also in the result side. Should be no regex right now.

wirybeaver:
When I benchmarked the Fun-ASR on the Consumer GPU 3 weeks ago, incidentally created a similar PR on my personal repo https://github.com/wirybeaver/sglang-omni/pull/3, overlapping with this formal PR. Will rebase when I resume the Fun-ASR RTX 4090 benchmark

sglang_omni/models/whisper_asr/request_builders.py
0xjeffro: lgtm

Issue #1437, opened 2026-08-10
[CI] Avoid retry exhaustion on transient cpuset contention (WIP)


AkazaAkane:
## Problem

CPU isolation is useful for keeping CI performance stable, but transient cpuset contention is currently treated like a stage failure.

If a requested cpuset is still occupied, retrying the stage immediately is likely to hit the same contention again. An unrelated benchmark can therefore consume its retry budget and eventually fail without actually running.

## Example

From PR #1397 (ARK-ASR optimization):

- Run: https://github.com/sgl-project/sglang-omni/actions/runs/31332470821
- Job: https://github.com/sgl-project/sglang-omni/actions/runs/31332470821/job/93309331036?pr=1397
- Job: `ci-test-2-gpu`
- Failed step: `Run TTS streaming benchmark stage`
- Job window: `2026-08-09 03:26:03 UTC` → `03:50:02 UTC`

Failure sequence:

1. The job starts and setup completes successfully.
2. `Run TTS streaming benchmark stage` starts.
3. The stage cannot acquire its requested CPUs because of cpuset contention.
4. This is counted through the stage failure/retry path.
5. Retrying soon afterward is likely to see the same CPUs still occupied, so additional retries can fail for the same reason.
6. The TTS stage is eventually marked failed; cleanup still succeeds.
7. The job therefore fails because of temporary runner resource contention 

## Proposal

When cpuset reservation fails because the CPUs are temporarily occupied:

```text
request cpuset
  |
  +-- available --> run stage
  |
  +-- contended --> wait / retry reservation
                       |
                       +-- available within budget --> run stage
                       |
                       +-- budget exhausted --> fail as cpuset contention
```

The cpuset wait should not consume the benchmark/test retry budget.

Use a separate time-based contention budget:

stage retry budget  -> actual benchmark/test failures
cpuset wait budget  -> time spent waiting for CPUs

The contention budget should be bounded by wall-clock time rather than another retry count. Otherwise the effective timeout depends on the polling interval.

For example:

cpuset_wait_timeout_s = <bounded timeout>

This should be one bounded wait budget for the stage, rather than resetting the budget on every reservation attempt.

If the budget expires, report a clear resource error including the requested cpuset and time spent waiting.

## PR sequence
Contended-pass → warning. ~10 lines in conftest.py. Independently shippable, immediate relief.
[...]

PR #1456, opened 2026-08-10
[CI] Contention report is advisory, never fatal

JiaxinD:
Fixes the retry-exhaustion failure mode from PR #1379: the MTD stage passed its gates on all three attempts and was killed each time by the contention guard in fixture teardown (foreign peaks 3.46 / 5.73 / 6.42 cores), spending 53 minutes to report that someone else was using the CPU.

The fail branch was unnecessary on physical grounds: contention can only depress perf numbers, never inflate them, so a gate that passed under intrusion passed for real. A gate that fails already retries through the normal failure path, with the `[cpuset-contention]` line in the log for one-grep triage. This removes the `pytest.fail` and keeps the report.

Calibration is unchanged and stays stricter (it rejects contaminated rounds and re-runs them, since worst-of-N references must not be depressed by intruders).

Complementary work tracked in #1437: wait for a clean cpuset before launching the stage instead of burning retry budget on a window that is known to be contended.

### DIFF
--- a/tests/test_model/conftest.py
+++ b/tests/test_model/conftest.py
@@ -64,16 +64,17 @@ def pin_omni_ci_cpuset() -> "Generator[None, None, None]":
     calibrated floors. Child processes inherit the affinity, which keeps the
     managed router, its workers, and the bench client on the reserved cores.
     Pinning restrains only this session, so a contention sampler reports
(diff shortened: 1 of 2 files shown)


stephenkgli (approved):
LGTM

PR #1458, opened 2026-08-10
[qwen3-asr] add breakable prefill CUDA graph (BCG) support

Ratish1:
Enables SGLang's breakable prefill CUDA graph for Qwen3-ASR, following the Higgs TTS
enablement (#1364). The captured region is the LM transformer body; the audio tower,
multimodal embedding composition, and logits processing stay eager, and composed
embeddings reach the captured body through SGLang's static `input_embeds` slot, so no
omni-side transport layer is needed.

Two seams had to be closed for this model. First, SGLang auto-disables the breakable
backend for multimodal models unless the backend is named explicitly, and an explicitly
named backend also skips SGLang's remaining breakable incompatibility rules, so the
policy layer has to enforce those itself. Second, the prefill graph bypasses
`forward()`, which is where positions get cast to the int32 the per-layer fused
qk-norm-rope kernel (#1447) requires, so the fused hook normalizes the dtype itself.
The position values agree between eager and capture by construction: Qwen3-ASR's mrope
rows are degenerate, so the runner's plain positions equal `forward()`'s row-0
substitution.

## Changes

- The fused qk-norm-rope hook normalizes positions to int32 before the kernel; the
  captured body skips `forward()`'s cast. Tests assert eager/capture position
  agreement and the kernel's dtype in both shapes.
- The engine builder names `cuda_graph_backend_prefill=breakable` and declares
  `supports_breakable_prefill_cuda_graph = True` (the factory refuses the breakable
  backend for models without the flag).
- The token-bucket ladder is derived after deployment overrides merge, from the
  effective caps: `chunked_prefill_size` (`-1` disables chunking and does not bound),
  `max_prefill_tokens`, `cuda_graph_max_bs_prefill`, `max_total_tokens`, and the
  resolved `context_length`. Lowering any cap shortens the ladder with it; an explicit
  bucket list is left untouched.
- The shared validator enforces the breakable incompatibility rules the explicit
  backend naming makes SGLang skip: attention CP, decode CP, LoRA, and MoE A2A.
- Partial nested `cuda_graph_config` prefill overrides (SGLang's highest-precedence
  form) are mirrored into the flat fields the composition reads, so nested `bs`
  without `max_bs` or the inverse can no longer resolve into a mismatch; conflicting
  flat and nested values are rejected.
- Disable per deployment with
  `server_args_overrides.cuda_graph_backend_prefill: disabled`.

## Benchmark

H100, full pinned SeedTTS EN (1088 clips), measured repeats after a discarded warmup,
arms back-to-back on the same GPU, baseline is main at `c6980be8c`. torch.compile is
enabled in both arms; main auto-resolves prefill graphs to disabled, the candidate
attests breakable graphs.

| Concurrency | Main req/s | BCG req/s | Throughput |
|---:|---:|---:|---:|
| 8 | 83.034 | 93.503 | +12.61% |
| 16 | 144.221 | 163.904 | +13.65% |
| 32 | 239.07 | 279.47 | +16.90% |
| 64 | 313.95 | 358.31 | +14.13% |

- At c8/c16: p95 latency 116.5 ms -> 106.2 ms and 133.4 ms -> 117.2 ms, mean RTF
  0.02086 -> 0.01852 and 0.02403 -> 0.02114, RTFx 393.26x -> 442.84x and
  683.04x -> 776.27x, per-request prefill profile 9.795 ms -> 5.051 ms and
  11.109 ms -> 5.989 ms.
- Corpus WER 0.0123335 and normalized-EN CER 0.0036427 in both arms, equal in every
  repeat; 6,528/6,528 requests succeeded per arm.
- Exact returned-transcript comparison at c1: 200/200 identical between eager and BCG.
- All 50 buckets from 4 through 4096 tokens capture and attest; prefill capture costs
  3.31 s and 0.59 GB. MPS DP2 attests all 50 buckets on both replicas.

### DIFF
--- a/sglang_omni/models/qwen3_asr/engine_builder.py
+++ b/sglang_omni/models/qwen3_asr/engine_builder.py
@@ -174,15 +174,15 @@ class Qwen3ASREngineBuilder(AsrEngineBuilder):
             return
         if "cuda_graph_bs_prefill" in overrides:
             return
-        # Derived from the merged caps: a ladder past any of them fails
-        # startup or captures buckets that can never replay.
+        # note(ratish): derived from the merged caps: a ladder past any of
+        # them fails startup or captures buckets that can never replay.
         caps = [
             overrides["max_prefill_tokens"],
             overrides.get("cuda_graph_max_bs_prefill"),
             overrides.get("max_total_tokens"),
             self.context_length,
         ]
-        if overrides["chunked_prefill_size"] != -1:
+        if overrides["chunked_prefill_size"] > 0:
             caps.append(overrides["chunked_prefill_size"])
         ladder = build_default_prefill_cuda_graph_bs(
             min(int(cap) for cap in caps if cap is not None)
--- a/sglang_omni/models/qwen3_asr/engine_builder.py
+++ b/sglang_omni/models/qwen3_asr/engine_builder.py
@@ -19,7 +19,11 @@ from sglang_omni.models.qwen3_asr.encoder_service import (
     build_cache_namespace,
 )
 from sglang_omni.scheduling.engine_factory import AsrEngineBuilder
-from sglang_omni.scheduling.generation_batch_policy import get_decode_cuda_graph_bs
+from sglang_omni.scheduling.generation_batch_policy import (
+    CudaGraphBackend,
+    build_default_prefill_cuda_graph_bs,
+    get_decode_cuda_graph_bs,
+)
 from sglang_omni.utils.gpu_compat import get_visible_gpu_sm_version
 from sglang_omni.utils.gpu_memory import format_bytes_gib, get_process_gpu_memory_bytes
 
@@ -29,6 +33,7 @@ logger = logging.getLogger(__name__)
 class Qwen3ASREngineBuilder(AsrEngineBuilder):
     model_name = "Qwen3-ASR"
     model_arch_override = "Qwen3ASRForConditionalGeneration"
+    supports_breakable_prefill_cuda_graph = True
 
     def __init__(
         self,
@@ -123,6 +128,7 @@ class Qwen3ASREngineBuilder(AsrEngineBuilder):
             "chunked_prefill_size": 4096,
             "sampling_backend": "pytorch",
             "dtype": dtype,
+            "cuda_graph_backend_prefill": CudaGraphBackend.BREAKABLE,
         }
         if self.mm_attention_backend is not None:
             defaults["mm_attention_backend"] = self.mm_attention_backend
@@ -164,6 +170,25 @@ class Qwen3ASREngineBuilder(AsrEngineBuilder):
     def adjust_overrides(self, overrides: dict[str, Any]) -> None:
         if "context_length" in overrides:
             self.context_length = int(overrides.pop("context_length"))
+        if overrides.get("cuda_graph_backend_prefill") == CudaGraphBackend.DISABLED:
+            return
+        if "cuda_graph_bs_prefill" in overrides:
+            return
+        # Derived from the merged caps: a ladder past any of them fails
+        # startup or captures buckets that can never replay.
+        caps = [
+            overrides["max_prefill_tokens"],
+            overrides.get("cuda_graph_max_bs_prefill"),
+            overrides.get("max_total_tokens"),
+            self.context_length,
+        ]
+        if overrides["chunked_prefill_size"] != -1:
+            caps.append(overrides["chunked_prefill_size"])
+        ladder = build_default_prefill_cuda_graph_bs(
+            min(int(cap) for cap in caps if cap is not None)
+        )
+        overrides["cuda_graph_bs_prefill"] = ladder
+        overrides["cuda_graph_max_bs_prefill"] = max(ladder)
 
     def customize_server_args(self, server_args: Any) -> None:
         self.context_length = int(server_args.context_length)
--- a/sglang_omni/models/qwen3_asr/sglang_model.py
+++ b/sglang_omni/models/qwen3_asr/sglang_model.py
@@ -61,8 +61,8 @@ def _fused_asr_forward_prepare_native(
             hidden_states,
         )
     if positions.dtype != torch.int32:
-        # The prefill CUDA graph bypasses forward()'s int32 cast; the kernel
-        # rejects int64 positions.
+        # note(ratish): the prefill CUDA graph bypasses forward()'s int32
+        # cast; the kernel rejects int64 positions.
         positions = positions.to(torch.int32)
     qkv, _ = attention.qkv_proj(hidden_states)
     fused_qk_norm_rope(
--- a/sglang_omni/models/qwen3_asr/sglang_model.py
+++ b/sglang_omni/models/qwen3_asr/sglang_model.py
@@ -60,6 +60,10 @@ def _fused_asr_forward_prepare_native(
             positions,
             hidden_states,
         )
+    if positions.dtype != torch.int32:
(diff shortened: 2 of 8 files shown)


PoTaTo-Mika:
I've reviewed and tested this pull request, all the unit tests have passed on my H200 device. 6258/6258 requests all passed.

PoTaTo-Mika (approved):
I've tested these changes on H200, and all the requests are passed. I think this pr can be merged.

sglang_omni/models/qwen3_asr/engine_builder.py
```diff
@@ -164,6 +170,25 @@ def validate_after_model_setup(self, model: Any, server_args: Any) -> None:
...
+            return
+        # Derived from the merged caps: a ladder past any of them fails
+        # startup or captures buckets that can never replay.
+        caps = [
+            overrides["max_prefill_tokens"],
+            overrides.get("cuda_graph_max_bs_prefill"),
+            overrides.get("max_total_tokens"),
+            self.context_length,
+        ]
+        if overrides["chunked_prefill_size"] != -1:
```
SandyLuXY: `chunked_prefill_size=0` is also normalized to disabled by SGLang's scheduler, but it is treated as a cap here and fails startup in `build_default_prefill_cuda_graph_bs(0)`. Consider treating only positive chunk sizes as ladder caps, matching the downstream scheduler and shared validator.

sglang_omni/models/qwen3_asr/engine_builder.py
```diff
@@ -164,6 +170,25 @@ def validate_after_model_setup(self, model: Any, server_args: Any) -> None:
     def adjust_overrides(self, overrides: dict[str, Any]) -> None:
         if "context_length" in overrides:
             self.context_length = int(overrides.pop("context_length"))
+        if overrides.get("cuda_graph_backend_prefill") == CudaGraphBackend.DISABLED:
+            return
+        if "cuda_graph_bs_prefill" in overrides:
+            return
+        # Derived from the merged caps: a ladder past any of them fails
```
SandyLuXY: Please follow the comment style: `# Note (xxx): `

sglang_omni/models/qwen3_asr/sglang_model.py
```diff
@@ -60,6 +60,10 @@ def _fused_asr_forward_prepare_native(
             positions,
             hidden_states,
         )
+    if positions.dtype != torch.int32:
+        # The prefill CUDA graph bypasses forward()'s int32 cast; the kernel
```
SandyLuXY: Same here

sglang_omni/scheduling/generation_batch_policy.py
```diff
@@ -78,6 +90,23 @@ def build_generation_batch_overrides(
     **stage_defaults: Any,
 ) -> dict[str, Any]:
     incoming = dict(server_args_overrides or {})
+    # The nested form wins in sglang; mirror its prefill fields into the flat keys.
```
SandyLuXY: Same here

sglang_omni/scheduling/generation_batch_policy.py
```diff
@@ -78,6 +90,23 @@ def build_generation_batch_overrides(
...
+        if nested_key not in nested_prefill:
+            continue
+        nested_value = nested_prefill[nested_key]
+        if flat_key in incoming and incoming[flat_key] != nested_value:
+            raise ValueError(
+                f"Conflicting {flat_key} and cuda_graph_config prefill "
+                f"{nested_key} values: "
+                f"{incoming[flat_key]!r} != {nested_value!r}"
+            )
+        incoming[flat_key] = nested_value
```
SandyLuXY, line 110: This mirrors nested `prefill.max_bs` but does not reconcile a lower-precedence stage-default bucket list. With Higgs defaults and nested `max_bs=128`, the resolved policy keeps `max(bs)=512` and fails validation.
Consider reconciling the buckets after merging and add a Higgs-default regression test.

SandyLuXY (commented):
Nice work! Left a few comments

sglang_omni/models/qwen3_asr/engine_builder.py
Ratish1: done

sglang_omni/scheduling/generation_batch_policy.py
Ratish1: done

tests/unit_test/qwen3_asr/test_pipeline.py
```diff
@@ -355,13 +414,130 @@ def _fake_create_infrastructure(server_args, gpu_id, **kwargs):
...
+    assert len(recorded.graph_init_calls) == 1
+    assert len(recorded.attest_calls) == 1
+
+
+@pytest.mark.parametrize(
+    ("cap_override", "expected_cap"),
+    [
+        ({"context_length": 1000}, 1000),
+        ({"chunked_prefill_size": 512}, 512),
+        ({"chunked_prefill_size": 0}, 1636),
```
SandyLuXY: This production path resets the Qwen3-ASR context length to 27,672 in pre_infra_setup(). With chunking disabled, max_prefill_tokens=4096 is therefore the limiting cap, so this expectation should be 4096 rather than 1636.

SandyLuXY (approved):

sglang_omni/models/qwen3_asr/engine_builder.py
```diff
@@ -164,6 +170,25 @@ def validate_after_model_setup(self, model: Any, server_args: Any) -> None:
...
+            return
+        if "cuda_graph_bs_prefill" in overrides:
+            return
+        # note(ratish): derived from the merged caps: a ladder past any of
+        # them fails startup or captures buckets that can never replay.
+        caps = [
+            overrides["max_prefill_tokens"],
+            overrides.get("cuda_graph_max_bs_prefill"),
+            overrides.get("max_total_tokens"),
+            self.context_length,
```
AkazaAkane, line 183: Is context_length actually a valid cap for this ladder? BCG buckets are keyed by the total prefill token count across the batch, while context_length is a per-request limit. A multi-request prefill can therefore have more total tokens than context_length while every request is still valid.

For example, with context_length=512, two 300-token extends give a 600-token BCG shape, but this would cap capture at 512 and force that batch to eager. max_prefill_tokens/chunked_prefill_size seem like the relevant per-forward caps here.

tests/unit_test/qwen3_asr/test_pipeline.py
Ratish1: fixed

sglang_omni/models/qwen3_asr/engine_builder.py
Ratish1: You're right about the shape mechanics: buckets are keyed by the total prefill tokens of a coalesced batch, and context_length only limits a single request, so a 600-token batch of two 300-token extends is legal past a 512 context. Those batches fall back to eager.

The reason context_length still sits in the caps is one layer down: whenever context_length is set (omni always sets it for Qwen3-ASR;the builder computes it), SGLang clamps the resolved prefill max_bs to it during ServerArgs resolution, capture warmup can't materialize a bucket larger than the model buffers hold. If we built buckets above that, the clamp lowers max_bs under max(bs) at construction and the policy validation fails at boot, so the ladder has to respect it.

In practice it never binds here: the shipped resolution gives context 27,672, so max_prefill_tokens/chunked_prefill_size (4096) are what actually shape the ladder, exactly as you say. If upstream ever relaxes the warmup clamp, dropping context_length from the caps could be  one-line change.

PR #1459, opened 2026-08-10
[dots.tts] Measure codec lock contention: GPU span timing + results

buffett0323:
## Motivation
Based on @Hayden727's guidance on #1434, end-to-end streaming jitter (inter-chunk p95/p99 + TTFC) is the primary signal and the per-call-site lock counters are attribution. Lock counters alone could not answer this: both paths release the codec lock before their D2H sync, so the lock covers kernel launches, not GPU execution. Reference encode can enqueue work, release the lock, and still delay `stream_step` on the shared default stream, low lock contention would not have proven absence of interference. 

So split ownership, not separate streams, and the blocking runs opposite to the roadmap's hypothesis. This PR is measurement only; the fix is deliberately left out because the roadmap gates it on this data.

## Modifications
- **`sglang_omni/utils/gpu_timing.py` (new)** — `CudaSpanTimer` records three timestamps per call site (CPU enqueue, GPU start, GPU end), giving `gpu_ms` **and `queue_ms`**: delay *before* the start event, which `elapsed_time(start, end)` cannot see. No extra hot-path synchronization — `drain()` is non-blocking via `event.query()`, since a `torch.cuda.synchronize()` there would flush the pipeline and mask the very queueing being measured. The ring buffer drops stale records and counts the drops rather than reading overwritten events.
- **`codec.py` / `vocoder.py`** — wrap the same 6 labeled sites from #1434 with spans; `maybe_log_lock_stats` → `maybe_log_contention`. Gated by `SGLANG_OMNI_PROFILE_GPU_SPANS`, independently of the lock profiler so the `wait≈0` + queueing case stays visible. Both off by default.
- **`benchmarks/eval/benchmark_tts_seedtts.py`** — `--reference-mode {per-sample,shared}` provides the cold/warm control pair (default preserves existing behavior).
- **Tests** — 33 new/extended, all GPU-free.

## Related Issues
Follow-up to #1434 (per-call-site lock profiling). Together these close the roadmap item in #1367:
> Measure codec lock contention: Split ownership or use separate streams only if profiling shows reference encoding blocks streaming decode.

## Accuracy Test

N/A

### Methodology & Control
We contrast active encoder execution with cached evaluation to isolate system contention:
- **`per-sample` (Cold):** Distinct prompts invoke active encoding routines.
- **`shared` (Warm):** A pinned prompt idles the encoder after request 1 (Negative Control).
- Verification: Unique file duplication hashes ensure clean execution; active arms processed 33–40 cold encodes vs. 1 warm encode.

### Results Matrix
*Environment: H200, streaming inference, 64 samples, 3 repeats per cell. Audio chunk budget = 80 ms.*

| Concurrency | ITL p95 Δ | Noise Band | TTFC p95 Δ | `stream_step` Max Wait | `reference_encode` Max Wait | Max GPU Queue |
|:---:|:---:|:---:|:---:|:---:|:---:|:---:|
| **1**  | +4.0%  | ±11.2% | **+51.9%** | 0 ms | 0 ms | 0.8 ms |
| **8**  | +1.4%  | ±7.6%  | **+50.2%** | 359 ms | 539 ms | 2.1 ms |
| **16** | −9.4%  | ±15.3% | **+31.9%** | 346 ms | **1077 ms** | 13.4 ms |

### Insights
1. **Streaming Smoothness Unaffected:** All Inter-Token Latency (ITL) p95 deltas reside within the noise floor.
2. **GPU Queueing Ruled Out:** Worst-case GPU queueing is `13.4 ms` (well below the `80 ms` budget), yielding a `0.3 ms` mean. This makes dedicated CUDA streams unnecessary.
3. **TTFC Regression via Inverted Contention:** At $c \ge 8$, lock contention is real but opposite to our roadmap hypothesis: `stream_step` holds the lock for `~157 ms` per call, blocking `reference_encode` for up to 1.08 second while its own wait time is only `12–14 ms`. Streaming decode blocks reference encoding.

### Nsight ($c=16$ Cold)
- `cudaLaunchKernel` accounts for 79% of CUDA API time (Max: `229 ms`).
- `stream_step` holding time strictly mirrors `gpu_ms` (`53.875s` vs `53.868s`), meaning kernel launches block on a full GPU ring buffer. 
- **Mechanism:** The CPU lock inherits raw GPU execution time during blocking launch calls. Moving to separate streams will not mitigate this since the bottleneck resides entirely within the CPU launch thread while the lock is active.

## Checklist

- [x] Add unit tests.

## CI

### DIFF
--- a/benchmarks/eval/benchmark_tts_seedtts.py
+++ b/benchmarks/eval/benchmark_tts_seedtts.py
@@ -175,12 +175,12 @@ import argparse
 import asyncio
 import logging
 import os
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from pathlib import Path
 
 from benchmarks.benchmarker.runner import BenchmarkRunner, RunConfig
 from benchmarks.benchmarker.utils import managed_omni_server
-from benchmarks.dataset.seedtts import load_seedtts_samples
+from benchmarks.dataset.seedtts import SampleInput, load_seedtts_samples
 from benchmarks.metrics.performance import (
     build_speed_results,
     compute_speed_metrics,
@@ -210,6 +210,28 @@ logger = logging.getLogger(__name__)
 DEFAULT_TTS_BENCHMARK_CONCURRENCY = int(os.getenv("TTS_BENCHMARK_CONCURRENCY", "16"))
 
 
+REFERENCE_MODES = ("per-sample", "shared")
+
+
+def _apply_reference_mode(samples: list[SampleInput], mode: str) -> list[SampleInput]:
+    """Select whether the reference-encode cache stays cold or warm.
+
+    Cache keys use file content, so "per-sample" forces encodes and "shared"
+    warms after the first request. Returns new objects because sample loading is memoized.
+    """
+    if mode not in REFERENCE_MODES:
+        raise ValueError(
+            f"reference_mode must be one of {REFERENCE_MODES}, got {mode!r}"
+        )
+    if mode == "per-sample" or not samples:
+        return samples
+    anchor = samples[0]
+    return [
+        replace(sample, ref_audio=anchor.ref_audio, ref_text=anchor.ref_text)
+        for sample in samples
+    ]
+
+
 @dataclass
 class TtsSeedttsBenchmarkConfig:
     model: str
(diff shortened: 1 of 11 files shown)


buffett0323:
Hi @Hayden727 ,

I feel like the issue is lock ownership, not CUDA streams. Because that lock stretches across the entire decode step, any `reference_encode` that hits mid-step just gets stuck waiting in line. So I'm thinking about splitting `reference_encode` out into its own lock so it stops queueing behind the streaming decode. Then re-run this exact cold/warm matrix to make sure TTFC bounces back while ITL stays flat. 

Looking at the queue numbers, adding separate CUDA streams feels like over-engineering right now, and I'd rather not add that complexity without a good reason.

If that makes sense, happy to grab it in a follow-up PR, thanks!

PR #1461, opened 2026-08-10
[CI] Confine unit-test and setup work to the runner lane cpuset

JiaxinD:
## Motivation

OMNI_CI_CPUSET pinning covers only the perf test processes (conftest fixture). The unit-test pytest, dependency install, env validation, model download, and cleanup steps still run with full machine affinity. Measured on the CI host today: the unit-test pytest of a concurrent omni-ci run averaged 7.2 cores and floated onto the lane freed for calibration, which kept tripping the calibration contention gate. The same mechanism pollutes neighboring perf lanes (the foreign peaks seen in #1379).

## Modifications

* New `.github/scripts/pin_to_ci_cpuset.sh`: sourced at the entry of the CPU-heavy CI scripts, pins the script shell to `OMNI_CI_CPUSET` via taskset, children inherit. Validates the effective mask by readback, fails the step on a bad or narrowed value, no-op when unset.
* Self-pin `run_flaky_pytest.sh`, `reconcile_omni_ci_env.sh`, `validate_omni_env_reusable.sh`, `ensure_hf_models.sh`, `cleanup_ci_pr_home.sh`, and the omni-setup restore step. Covers every call site, including the perf workflows' setup phases which already pass the variable.
* Pass `-e OMNI_CI_CPUSET` into the four job containers that were missing it (unit-test, setup, cleanup, close-cleanup), same passthrough as the perf workflows.
* Calibration skill lane table: GPU `0,1` now maps to `2-15,66-79`. Cores `0,1,64,65` become system cores where host daemons (categraf, dockerd) are confined, so neither calibration nor CI measures daemon noise.

## Runner-side changes (deployed on the CI host, disclosed here)

* All four `run-ci.sh` now self-pin the runner tree (listener, worker, docker CLI) to the lane cpuset; live trees pinned in place.
* `actions-runner-h100-4/.env` updated to `OMNI_CI_CPUSET=2-15,66-79`; autoscaler lane table matched.
* systemd drop-ins confine categraf and dockerd to the system cores (containerd deliberately untouched: container processes inherit its mask via shim spawn).

## Scope

TTS stage 5 (`tests/test_ci`, MPS DP2) stays unpinned as before: its thresholds were calibrated without a cpuset, so confining it must land together with a recalibration. Follow-up. `actions/checkout` inside the job container cannot be pinned from the repo side; the runner-tree pin bounds the rest.

## Evidence

* Simulated the pinned step shape in the CI image: worker shell, python child, and grandchild all confined to the requested cpuset; unset means no-op with full affinity; a bad value and a kernel-narrowed mask fail loudly (exit 2).
* Live affinity readout of this PR's own CI on the runner to follow in a comment.

### DIFF
--- a/.claude/skills/tune-ci-thresholds/hosts/sglang-h100-ci.yaml
+++ b/.claude/skills/tune-ci-thresholds/hosts/sglang-h100-ci.yaml
@@ -20,14 +20,12 @@ venv_python: /github/home/calibration/omni/bin/python
 # On 8× H100 shared hosts, GPU 6–7 run CI continuously — never pick or kill them.
 gpu_exclude: [6, 7]
 
-# note (Jiaxin Deng): each CI runner lane owns two GPUs and exactly 32
-# logical cores (16 physical + HT siblings). Keep in sync with the
-# production runner .env partition. Calibration must pin the same bundle:
-# TUNE_GPU_INCLUDE=0,1 → OMNI_CI_CPUSET=0-15,64-79, etc. PR CI does not
-# need to know its lane; calibration does, because it chooses the GPUs.
-# precheck refuses a busy cpuset rather than measuring under intrusion.
+# note (Jiaxin Deng): lane bundles must match the runner-side .env; precheck
+# refuses a busy cpuset rather than measuring under intrusion. Lane 0,1
+# excludes system cores 0,1,64,65 (categraf/dockerd confined there), so its
+# 28-core thresholds are conservative for the 32-core lanes.
 gpu_group_cpusets:
-  "0,1": "0-15,64-79"
+  "0,1": "2-15,66-79"
   "2,3": "16-31,80-95"
   "4,5": "48-63,112-127"
   "6,7": "32-47,96-111"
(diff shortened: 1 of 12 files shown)


SandyLuXY (approved):

.claude/skills/tune-ci-thresholds/hosts/sglang-h100-ci.yaml
```diff
@@ -20,14 +20,12 @@ venv_python: /github/home/calibration/omni/bin/python
...
-# note (Jiaxin Deng): each CI runner lane owns two GPUs and exactly 32
-# logical cores (16 physical + HT siblings). Keep in sync with the
-# production runner .env partition. Calibration must pin the same bundle:
-# TUNE_GPU_INCLUDE=0,1 → OMNI_CI_CPUSET=0-15,64-79, etc. PR CI does not
-# need to know its lane; calibration does, because it chooses the GPUs.
-# precheck refuses a busy cpuset rather than measuring under intrusion.
+# note (Jiaxin Deng): lane bundles must match the runner-side .env; precheck
+# refuses a busy cpuset rather than measuring under intrusion. Lane 0,1
+# excludes system cores 0,1,64,65 (categraf/dockerd confined there), so its
+# 28-core thresholds are conservative for the 32-core lanes.
```
AkazaAkane, line 26: Do we need to enforce that performance thresholds are calibrated from the 0,1 lane now? This lane has 28 logical cores while the other CI lanes have 32, but Mode C allows an independent full calibration on any lane. A threshold applied from a 32-core calibration may therefore be too aggressive when CI lands on 0,1.

JiaxinD: Right, the safe direction is one-way. Made it a rule in AGENT-PRECHECK (a212263e): threshold-writing calibrations anchor on the 0,1 lane, 32-core lanes are measurement-only.

Issue #1418, opened 2026-08-08
[TTS][Qwen3] 1.7B variant: dedicated single-instance serving optimization


JiaxinD:
## Goal

Community deployments primarily use `Qwen/Qwen3-TTS-12Hz-1.7B-Base`, but the original tuning work focused on 0.6B. On one H100, the 1.7B single instance peaked at 10.26 QPS while same-card DP2+MPS reached 15.26 QPS (+49%), pointing to host-side serving overhead rather than a model-compute limit.

Exit criterion: bring single-instance peak within 95% of same-card DP2+MPS on the same silicon, with C16 EN WER within 0.2 points of baseline and TTFA p95 at c8/c16 degraded by no more than 10%, or document a profile-backed structural reason for the remaining gap.

## Validated outcome

| Silicon | Baseline single instance | Optimized single instance | Same-card DP2+MPS | Optimized gain |
|---|---:|---:|---:|---:|
| H200 | 8.74 QPS | 12.40 QPS | 13.76 QPS | **+41.9%** |
| H100 | 10.59 QPS | 14.86 QPS | 15.61 QPS | **+40.3%**; 95.2% of pool peak |

Quality remained within the gate (H100 C16 EN WER: default 1.06%, optimized 1.11%, DP2 1.01%). The admission/queueing TTFA cliff was removed; H100 c64 TTFA p95 improved from 4.76s to 0.54s, and H200 c64 measured 1.08s.

## Completed work

- [x] **Admission, queueing, and host path:** #1413, #1449, #1462, #1649
- [x] **Benchmark and regression coverage:** #1505
- [x] **Streaming and vocoder foundation:** #1286, #1336, #1523, #1675, #1756, #1786
- [x] **Predictor and sampling optimizations:** #1641, #1726
- [x] **Evaluated NO-GO or superseded paths:** #704, #1643, #1722, #1733, #1790, #1847

## Remaining work

- [ ] **P0 — Eliminate the overload device-side assert.** Under load, an invalid codec value can reach the vocoder lookup and abort the pipeline (`vectorized_gather_kernel` index out of bounds; one observed run failed 230 requests). Use the reproducible `feat/qtts17-mask-shaping` fork branch to identify the first producer-side invalid tuple, enforce the producer invariant, and prove the signature is gone on both the accelerated repro and baseline paths.
[...]

PR #1413, opened 2026-08-08
[TTS] Fix talker SGLang role mappings for Qwen3-TTS and Higgs

charliechenye:
## Motivation

Addresses #1399 Section 3.2.

Qwen3-TTS rejects generation policies where `torch_compile_max_bs < max_running_requests`. Operators therefore need to raise the talker compile cap above the existing 32 request ceiling.

The CLI exposes `--talker-torch-compile-max-bs`, but Qwen3-TTS did not declare a talker SGLang stage, so the flag was rejected as unsupported. A cross model audit found the same missing capability declaration in Higgs TTS.

## Modifications

- Add `talker_sglang_role_to_stage() -> {"talker": "tts_engine"}` to Qwen3-TTS.
- Add the same capability mapping to Higgs TTS.
- Make compile cap validation errors mention the public fixing flag.
- Add CLI routing and fail closed capability tests.
- Add a real Qwen3-TTS engine builder regression for the 64/64/64 policy.
- Preserve rejection of undercovered 64/64/32 configurations.

## Related Issues

Addresses #1399 Section 3.2

## Accuracy Test

Not applicable: no model math, weights, sampling, or default execution behavior changes.

## Benchmark & Profiling

Not applicable: this fixes configuration reachability and does not change runtime defaults or performance behavior.

## Validation

- Generation policy tests: 16 passed.
- CLI routing tests: 18 passed.
- Qwen3-TTS engine policy regression: passed.
- Ruff and pre commit checks: passed.

## Checklist

- [x] Format code with pre commit.
- [x] Add unit tests.
- [x] No documentation changes required.
- [x] No benchmark or accuracy evaluation required.

## CI

CI runs on self-hosted GPU runners and requires a maintainer to add the
`run-ci` label. Once labeled, every subsequent push re-triggers CI as
long as the label remains. Use `/tag-and-rerun-ci higgs` or
`/tag-and-rerun-ci moss` to select a TTS CI model, and
`/tag-and-rerun-ci fun-asr` or `/tag-and-rerun-ci qwen3-asr` to select an ASR
CI model. One selector from each family can be combined, for example
`/tag-and-rerun-ci moss fun-asr`. Draft PRs are skipped even if labeled.

Issue #1400, opened 2026-08-08
[TTS][Qwen3] Same-GPU MPS-DP: validated DP2 recipe (+55%) and docs restructuring


JiaxinD:
## Context

Splits out Section 4 of #1399 (MPS same-GPU DP for Qwen3-TTS) into its own tracking issue, per the owner assignment there.

Community feedback (Baseten) read the ~76GB single-instance peak as "DP2 cannot fit on an 80GB H100 without IPC weight sharing", then read the `Qwen3TTSTalker` "not supported" entry in `docs/basic_usage/mps_dp.md` as MPS-DP being unavailable for this model entirely. Both conclusions are wrong, and #1399 section 4.2 establishes the docs are what produce them: the support matrix gates only `WEIGHT_SHARE=1` (CUDA-IPC weight sharing), while plain MPS-DP with a lowered per-replica `mem_fraction_static` is a separate, available path.

This issue carries (a) the validated recipe that resolves the two `<TBD>` values in #1399 section 4.3, and (b) the docs and config work from section 4.5 that follows from it.

## 1. Validated DP2 recipe (one 80GB H100, main `51e62f6e`)

The two TBDs are `mem_fraction_static: 0.30` and `max_total_tokens: 30000`. Exact launch behind the +55% number:

```bash
CONFIG=examples/configs/qwen3_tts_0_6b.yaml N=2 MF=0.30 MAX_TOTAL_TOKENS=30000 \
  GPU_ID=<gpu> CORE_BLOCKS="0-11 12-23" bash examples/mps_dp/launch.sh up
```

`WEIGHT_SHARE` at its default `0`; the `Qwen3TTSTalker` preflight gate was not bypassed. Two replicas fronted by the omni router (round_robin). Seed-TTS EN-1088, same harness as the published follow-up rows. Measured with the #1399 section 1.1 shim applied, which is unavoidable on the transformers 5.12 stack; the shim only renames kwargs.

| arm | c8 | c16 | c32 | peak qps |
|---|---|---|---|---|
| DP2 + MPS | 7.24 | 11.38 | 14.85 | **15.76 @c64** |
| single instance, same card | 6.53 | 9.99 | 8.88 | 10.15 @c64 |

* +55% at peak; at c32 the pool also wins latency p95 (3.04s vs 3.98s) and RTF p95 (0.61 vs 1.03), since per-replica load halves
[...]

PR #1462, opened 2026-08-11
[Perf] Qwen3-TTS: cut per-step host work in the AR decode loop

JiaxinD:
## Motivation

Qwen3-TTS 1.7B single-instance serving plateaus far below what the card sustains: the same H200 runs a same-card DP2+MPS pool ~1.5x faster (#1418). py-spy at c32 shows the AR scheduler thread burning 72% of one core in per-step Python work (repetition-penalty set rebuilds, suppress-token tensor rebuilds, sampling-buffer restaging, per-row output clones), which is why the ceiling is size independent (0.6B and 1.7B peak identical).

## Changes

All host-side, no numerics change:

* `model_runner/base.py`: repetition-penalty unique-token tracking is incremental per request (was a full `unique(output_ids)` rebuild per step, quadratic over generation; shrunk `output_ids` resets state). Codec suppress-token device tensors cached by list identity (was per-step Python loop + tensor build).
* `models/qwen3_tts/sglang_model.py`: `prepare_decode_buffers` reuses staged sampling state via a request-id fingerprint when batch composition is unchanged (pattern from qwen3_omni talker #1043); predictor-graph capture invalidates the fingerprint.
* `models/qwen3_tts/model_runner.py`: one batched output snapshot clone per step instead of two clones per row (pattern from #1167); feedback staging stacks straight into the graph buffer (pattern from #1439); decode row ids cached.
* `models/qwen3_tts/stages.py` + `scheduling/threaded_simple_scheduler.py`: preprocessing runs on `ThreadedSimpleScheduler` (pattern from higgs/fish/moss_tts_local); the threaded scheduler gains the `abort_callback` contract the qwen3 prepared-request registry cleanup requires (invoked on abort and on late finish after abort).
* `models/qwen3_tts/streaming_vocoder.py`: the per-decode `torch.cat` rebuilt the full stream history although the decode window only moves forward; dead frames behind `window_start` are pruned and slices translate by the offset, making each decode O(window) instead of O(stream). Windows verified byte-identical to full-history slicing across a simulated long stream.

## Evidence (one H200, EN-1088, Seed-TTS harness, main `25ae3d95` base)

Serving defaults for the measured rows: `--max-running-requests 64 --cuda-graph-max-bs 64 --talker-torch-compile-max-bs 64 --isolate-stage vocoder` (defaults retune tracked in #1418; measured numbers below marked measured, everything else inferred).

Single-pass grids (measured). Attribution: the serving flags alone (no code change, unlocked by #1413) already move the peak; this branch's own delta on top of those flags is the row-over-row difference below, concentrated at c32.

| qps | c8 | c16 | c32 | c64 | c96 | peak | delta vs baseline | C16 EN WER |
| baseline defaults | 6.08 | 8.40 | 8.74 | 7.44 | 8.29 | 8.74 | — | 0.99% |
| + admission 64 + vocoder isolate, main code | 6.26 | 8.30 | 8.85 | 11.96 | 12.05 | 12.05 | +37.9% | 0.99% |
| + this branch (same flags, pre-hardening head) | 6.44 | 8.83 | 10.98 | 10.28 | 12.62 | 12.62 | +44.4% | pending |
| DP2+MPS same silicon (#1400 recipe, MF=0.40) | 4.57 | 7.05 | 9.43 | 10.29 | 10.02 | 10.29 | +17.7% | n/a |

This branch over the flags-only row: c32 +24.1% (8.85 -> 10.98), peak +4.7% (12.05 -> 12.62); the flags rows carry the rest of the cumulative gain. Defaults re-tune to make the flags row the out-of-the-box behavior is tracked in #1418 section 2.

H100 calibration of the final head, one card back to back, quiet box (load 3.8-8), all cells 1088/1088:

| qps | c8 | c16 | c32 | c64 | c96 | peak | TTFA p95 @c64 |
| baseline defaults | 7.23 | 10.33 | 10.64 | 10.71 | 10.40 | 10.71 | 4.76 s |
| this branch + flags | 8.16 | 11.31 | 13.81 | 15.80 | 16.15 | **16.15 (+50.8%)** | **0.54 s** |
| DP2+MPS re-measured (#1400 recipe) | 7.14 | 11.13 | 14.81 | 13.81 | 15.66 | 15.66 | 2.80 s |

The optimized single instance now **exceeds** the same-card DP2+MPS pool peak (16.15 vs 15.66, +3.2%) while holding a 5x better TTFA p95 at c64, so the #1418 exit criterion is met outright rather than approached: colocation is no longer the recommended deployment for this variant at peak throughput, and it was never competitive on first-audio latency. DP2 still leads at c32 alone (14.81 vs 13.81). C16 EN WER on the optimized head is 1.12% against the 1.07% published baseline, inside the 0.2pt gate.

TTFA p95 stays flat at c8/c16 in every row (0.20-0.34s); with admission 64 the c64 cliff is gone (7.77s -> 0.69-0.97s).

Same-card three-arm ABAB, 5 rounds, H200, peak qps medians over c32-96: baseline 8.74, this branch + flags 12.40 (+42%, four clean rounds; one round lost to the pre-existing intermittent c64 device assert tracked in #1418 section 4), DP2+MPS 13.76. TTFA p95 at c64 medians: 0.95s vs 2.91s vs 5.93s. On H200 the branch lands at 90% of the pool peak with a 3x TTFA advantage; on H100 (table below) it lands at 95.2%.

Rebased head (adds the vocoder window pruning, current main base), single pass on H200, all cells 1088/1088: 6.34 / 8.90 / 10.95 / 11.50 / 10.96 qps at c8-c96, within the ABAB round spread above.

## Follow-up work in this branch

Three further changes landed after the first review round, each measured separately:

* **Streaming vocoder window pruning.** The per-decode `torch.cat` rebuilt the whole stream history although the window only moves forward; dead chunks are dropped and slices translate by the pruned offset. Windows verified byte-identical to full-history slicing.
* **Codec-id screen.** An id outside the codebook reached the embedding lookup and became a device-side assert that poisoned the CUDA context and killed every in-flight stream in the vocoder process (`validate_chunk` skips device tensors to avoid a per-chunk sync). The decoder input is now clamped into range and the per-row verdict is read back after the decode's existing `stream.synchronize()`, so no decode gains a synchronization, and only the offending rows fail while the rest of the batch still decodes.
* **Device-resident shaping masks + pinned id staging.** Repetition penalty and codec suppression move from per-step host index building to persistent device masks updated by a single scatter of the last sampled token (rebuilt whenever a history does not grow by exactly one token, so retract and restart stay correct); sampled ids are staged into pinned host memory so the output processor's `.tolist()` waits on an event instead of issuing a blocking pageable copy.

Paired A/B on one H200, three alternating rounds, masks on vs off (same head otherwise):

| c16 qps | r1 | r2 | r3 | mean |
|---|---|---|---|---|
| masks off | 8.61 | 8.51 | 8.63 | 8.58 |
| masks on | 9.44 | 9.26 | 9.49 | **9.40 (+9.5%)** |

At c32 the same comparison is neutral (11.11 vs 11.09), so the win is concentrated where the AR scheduler thread is the binding constraint. The pinned staging was A/B'd the same way and is **not** credited with a gain: three rounds gave -0.9%, +5.6%, +4.7% at c16, so any effect is below this setup's ~5% noise floor. It is kept because it is strictly less host work on the same path qwen3_omni already uses, not because it measured faster. The tradeoff is measurable and consistent: c16 TTFA p95 rises from 0.36s to 0.42s (+17%) as batches run deeper. Full-grid single pass with everything on: 6.95 / 8.97 / 11.42 / 13.04 / 13.53 qps at c8-c96, zero failures, which is 98% of the same-card DP2+MPS pool peak on that box. Overload behaviour improves too: c64 +22% qps over the masks-off head (13.04 vs 10.65) with TTFA p95 1.08s, and c96 TTFA p95 is better rather than worse (2.87s vs 3.74s) with lower p99 latency throughout. C16 EN WER on that head is 1.04% against 0.99% for the same-box baseline arms, inside the 0.2pt quality gate.

## Tests

* `tests/unit_test/qwen3_omni/test_logit_shaping.py`: incremental repetition penalty equivalence across growth and retract-shrink steps; suppress-token grouped/cached path equivalence vs scalar reference (shared and per-row lists, out-of-range ids, repeated calls through the cache).
* `tests/unit_test/qwen3_tts/test_pipeline.py`: abort-race cleanup test adapted to the threaded executor, same contract (late finish after abort still cleans the prepared registry); stream-prune window test asserting pruned decode windows stay byte-identical to full-history slicing while chunks actually get dropped.
* `tests/unit_test/qwen3_tts/test_mask_logit_shaping.py`: mask shaping equals the scalar reference across steady steps, request-id reuse, batch shrink, history shrink after a retract, and an emptied history.
* `tests/unit_test/qwen3_tts/test_pipeline.py`: a decode batch with one out-of-range row fails exactly that row, keeps the decoder from ever seeing the bad id, and still decodes the remaining rows.
* Full `qwen3_tts` + `scheduling` + `model_runner` + `qwen3_omni` unit suites green (191 in the TTS-facing subset after the follow-up work; 463 across the wider set earlier in the branch).

## Cross-model disclosure

`model_runner/base.py` and `threaded_simple_scheduler.py` are shared surfaces: the penalty/suppress helpers also serve the thinker path (qwen3_omni), and the threaded scheduler is used by higgs/fish/moss_tts_local preprocessing. The suppress/penalty changes are output-equivalent by unit test; the scheduler change is additive (new optional parameter, default None keeps prior behavior).

## Acknowledgements

Thanks to @vuuihc for the original bounded ThreadedSimpleScheduler abort-marker cleanup in #611. This PR incorporated that earlier work while adding the abort_callback contract; I missed the cross-reference when #1462 was merged.

### DIFF
--- a/sglang_omni/model_runner/base.py
+++ b/sglang_omni/model_runner/base.py
@@ -958,14 +958,15 @@ class ModelRunner:
             return
         vocab = logits.shape[1]
         device = logits.device
-        # Note: (Jiaxin Deng) suppress lists are static per request; caching the
-        # filtered device tensor by list identity makes every decode step after
-        # a request's first a cache hit instead of a rebuild.
+        # Note: (Jiaxin Deng) the suppress set comes from model config, so keying
+        # by content collapses the whole fleet onto one device tensor; the key
+        # itself is derived once per request because the builder hands out a
+        # fresh list object each time.
         cache = getattr(self, "_suppress_tensor_cache", None)
         if cache is None:
             cache = {}
             self._suppress_tensor_cache = cache
-        row_groups: dict[int, tuple[torch.Tensor, list[int]]] = {}
+        row_groups: dict[Any, tuple[torch.Tensor, list[int]]] = {}
         for row_idx, sched_req in enumerate(requests):
             data = sched_req.data
             suppress_tokens = data.suppress_tokens
@@ -977,25 +978,25 @@ class ModelRunner:
                     suppress_tokens = None
             if not suppress_tokens:
                 continue
-            key = (id(suppress_tokens), vocab, str(device))
-            hit = cache.get(key)
-            if hit is not None and hit[0] is suppress_tokens:
-                toks_t = hit[1]
-            else:
-                toks = [int(t) for t in suppress_tokens if 0 <= int(t) < vocab]
+            content = getattr(data, "_suppress_content", None)
+            if content is None:
+                content = tuple(int(t) for t in suppress_tokens)
+                data._suppress_content = content
+            key = (content, vocab, str(device))
+            toks_t = cache.get(key)
+            if key not in cache:
+                toks = [t for t in content if 0 <= t < vocab]
                 toks_t = (
                     torch.tensor(toks, dtype=torch.long, device=device)
                     if toks
                     else None
                 )
-                if len(cache) >= 64:
-                    cache.clear()
-                cache[key] = (suppress_tokens, toks_t)
+                cache[key] = toks_t
             if toks_t is None:
                 continue
-            group = row_groups.get(id(suppress_tokens))
+            group = row_groups.get(key)
             if group is None:
-                row_groups[id(suppress_tokens)] = (toks_t, [row_idx])
+                row_groups[key] = (toks_t, [row_idx])
             else:
                 group[1].append(row_idx)
         for toks_t, rows in row_groups.values():
--- a/sglang_omni/model_runner/base.py
+++ b/sglang_omni/model_runner/base.py
@@ -930,6 +930,12 @@ class ModelRunner:
                 continue
             output_ids = req.output_ids
             if not output_ids:
+                # Note: (Jiaxin Deng) a retract can replace output_ids with an
+                # empty list; drop the incremental state so a restart does not
+                # inherit stale tokens.
+                if getattr(data, "_rep_seen_len", 0):
+                    data._rep_seen_tokens = set()
+                    data._rep_seen_len = 0
                 continue
             unique = ModelRunner._rep_penalty_unique_tokens(data, output_ids, vocab)
             if not unique:
@@ -952,9 +958,9 @@ class ModelRunner:
             return
         vocab = logits.shape[1]
         device = logits.device
-        # Note: (Jiaxin Deng) suppress lists are static per request and shared
-        # across a deployment in practice; cache the filtered device tensor per
-        # (list identity, vocab, device) so the decode loop stops rebuilding it.
+        # Note: (Jiaxin Deng) suppress lists are static per request; caching the
+        # filtered device tensor by list identity makes every decode step after
+        # a request's first a cache hit instead of a rebuild.
         cache = getattr(self, "_suppress_tensor_cache", None)
         if cache is None:
             cache = {}
--- a/sglang_omni/model_runner/base.py
+++ b/sglang_omni/model_runner/base.py
@@ -119,6 +119,7 @@ class ModelRunner:
         self._async_query_miss: int = 0
         self._token_id_host_bufs: list[torch.Tensor] | None = None
         self._token_id_host_slot: int = 0
+        self._suppress_tensor_cache: dict[tuple, tuple[Any, torch.Tensor | None]] = {}
 
     def _stage_token_ids(self, result: Any, ids: torch.Tensor) -> None:
         # Note (wenyao): pinned host copy staged once at sample time so downstream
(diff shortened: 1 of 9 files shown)


sglang_omni/model_runner/base.py
```diff
@@ -946,13 +977,30 @@ def _apply_codec_suppress_tokens(self, logits_output: Any, requests: list) -> No
...
+            else:
+                toks = [int(t) for t in suppress_tokens if 0 <= int(t) < vocab]
+                toks_t = (
+                    torch.tensor(toks, dtype=torch.long, device=device)
+                    if toks
+                    else None
+                )
+                if len(cache) >= 64:
+                    cache.clear()
+                cache[key] = (suppress_tokens, toks_t)
```
0xjeffro: To my understand, no matter how the concurrency, it looks like this cache should only hold one element?
so what is this cap `64` guarding for?

sglang_omni/model_runner/base.py
```diff
@@ -933,8 +958,14 @@ def _apply_codec_suppress_tokens(self, logits_output: Any, requests: list) -> No
...
+        # filtered device tensor by list identity makes every decode step after
+        # a request's first a cache hit instead of a rebuild.
+        cache = getattr(self, "_suppress_tensor_cache", None)
+        if cache is None:
+            cache = {}
+            self._suppress_tensor_cache = cache
+        row_groups: dict[int, tuple[torch.Tensor, list[int]]] = {}
         for row_idx, sched_req in enumerate(requests):
             data = sched_req.data
             suppress_tokens = data.suppress_tokens
```
0xjeffro, line 972: nit: `data.suppress_token` is passed in from request builder, but logically this is a model config, so this should identical for every request. maybe we can just built once when model runner init, rather than cache them?

sglang_omni/model_runner/base.py
JiaxinD: Right, the cap was covering for identity keying: the builder hands out a fresh list per request, so it churned one entry per request. Now keyed by content, so it holds one entry and the cap is gone.

sglang_omni/model_runner/base.py
JiaxinD: Agreed it is model config; base cannot assume that for every model, so the content key is derived once per request and the tensor is shared fleet-wide. Same effect as building at init, without the assumption.

0xjeffro (approved):

JiaxinD:
Thanks @vuuihc for catching this and for the original work in #611. I have updated the PR description to credit that earlier patch; the missing cross-reference was my oversight.

PR #1463, opened 2026-08-11
[Perf] Topology-aware CPU allocator for pipeline stage processes

JiaxinD:
## Motivation

Host-bound speech serving collapses under CPU colocation: past a contention threshold a server whose colocated work is placed anywhere it likes loses 60 to 72% of its throughput, while the same server whose colocated work is confined to a bounded set of cores retains 92 to 98% (table below). Today that placement is hand-written per deployment (mps_dp CORE_BLOCKS, calibration cpusets, ad hoc taskset); nothing is NUMA/SMT-aware, per stage, or reusable, and nothing covers the serving parent. This PR derives it instead.

To be precise about the mechanism, since it decides how to read the table: what moves throughput is that colocated work is confined, not that the serial loop holds cores exclusively. Against hand-written CORE_BLOCKS the plan measures a wash (0.92 to 1.02x over five models under DP), so this PR buys a correct, automatic, whole-tree placement and a way to see contention, not a throughput multiplier on top of masks somebody already tuned by hand. The decomposition is below.

## Modifications

* New `sglang_omni/cpu_alloc/`: sysfs CPU/NUMA/SMT discovery (universe = `sched_getaffinity`, so container cpusets bound the plan), GPU-to-NUMA via PCI, a two-pool allocator that grants whole physical cores (both SMT siblings) to declared serial dispatch loops with an explicit degradation ladder, an auto-mode supervisor that applies and releases the plan by per-thread `/proc/<pid>/task` replay, and a host contention monitor.
* `--cpu-allocator off|static|auto` (default off, zero behavior change). `static` applies the plan for the process lifetime; `auto` applies it only while the tree is starved next to foreign load and releases it when that load goes away, which is what removes the small cost of pinning a quiet box. Plans are keyed by `ProcessTopologyPlan` (colocated groups, CPU-only stages, per-rank TP). Affinity is seeded on the spawning thread so import-time torch threads inherit it; `OMP_NUM_THREADS` env defaults are capped to the plan width.
* `stage_cpu_costs()` declarations for 8 pipelines (declaration-only edits to 8 model configs): `serial-loop` for dispatch loops that need exclusive cores, `gpu-bound` for everything else, and undeclared stages default to `gpu-bound`, so a model that declares nothing is a no-op even when enabled. Inside a container the plan uses the container's own CPU set while reading core and SMT relationships from sysfs, which stays truthful under `--cpuset-cpus`; a half-visible physical core is planned as the visible half only.
* `GET /host_contention`: foreign CPU load on the server's allowed cpuset.
* `python -m sglang_omni.cpu_alloc plan` emits NUMA/SMT-correct `CORE_BLOCKS`; it exits 3 when it refuses for capacity, so autodp.sh can tell a refusal from an unavailable planner and only falls back to its NUMA-blind split for the latter, with the blocks checked for count, emptiness and overlap before any launch. Usage and capacity-planning docs in `docs/basic_usage/cpu_allocator.md`.

## Where it sits

The allocator slots into the serve startup chain at exactly one point: after the process topology is resolved and before stage processes are spawned. Everything runs once in the serve parent; the request path is unchanged.

```mermaid
flowchart LR
    A["CLI parse"] --> B["load pipeline<br/>config"]
    B --> C["resolve process topology<br/>stages to OS processes,<br/>one per TP rank"]
    C --> P["CPU allocator (this PR)<br/>sysfs topology + stage_cpu_costs()<br/>-> per-process cpu_ids"]
    P --> D["spawn stage processes<br/>affinity seeded before proc.start(),<br/>inherited from birth incl. torch threads"]
    D --> E["HTTP serving<br/>request path unchanged"]

    P --> X["declared serial loops:<br/>exclusive whole physical cores<br/>(both SMT siblings)"]
    P --> Y["everything else:<br/>NUMA-node shared pool"]
    S["auto mode:<br/>apply the plan while starved,<br/>release it when the box goes quiet"] -.-> P
    E --- M["GET /host_contention:<br/>foreign load on the cpuset"]

    style P fill:#e8f4d9,stroke:#4a7c1f
```

## Accuracy Test

No numerics touched; default off is a zero-diff no-op. Every non-excluded campaign round served real inference (ASR transcript verified, TTS returns WAV) with zero request errors.

## Benchmark

1x H200 plus one 16-physical-core lane per model; colocated load = spin processes roaming the lane (`off`, today's behavior) vs confined to the shared pool by the allocator (`static`). Closed loop, A-B-A-B per load level, 90 s measured rounds, per-round `/proc` affinity evidence. QPS, 2-round means, ratio = static/off:

| model | endpoint, conc | no load | moderate load | heavy load | heavy-load retention (static vs own no-load) |
| Fun-ASR | transcriptions, c32 | 147.0 / 147.6 = **1.00x** | 141.6 / 147.1 = **1.04x** | 40.9 / 144.6 = **3.54x** | 98% |
| Qwen3-ASR | transcriptions, c32 | 146.4 / 144.6 = **0.99x** | 144.7 / 144.7 = **1.00x** | 56.7 / 141.7 = **2.50x** | 98% |
| Higgs TTS | speech, c32 | 21.7 / 20.5 = **0.95x** (1.01x isolated) | 17.8 / 21.8 = **1.22x** | 12.2 / 20.0 = **1.63x** | 92 to 97% |
| MOSS-TTS-Local | speech+ref, c32 | 10.5 / 10.7 = **1.01x** | 9.5 / 10.4 = **1.10x** | 7.4 / 10.6 = **1.44x** | ~100% |
| Fish S2-Pro | speech+ref, c8 | 2.41 / 2.41 = **1.00x** | 2.10 / 2.42 = **1.15x** | 1.09 / 2.37 = **2.18x** | 98% |
| dots.tts | speech+ref, c32 | 1.94 / 1.92 = **0.99x** | 1.72 / 1.91 = **1.11x** | 0.98 / 1.84 = **1.88x** | 96% |

Below the model's contention threshold the allocator costs nothing; past it the unprotected arm collapses to 28 to 60% of clean throughput while the protected arm retains 92 to 100%, with p95 2 to 3x lower (e.g. Qwen3-ASR 0.66 s vs 0.25 s).

Effect decomposition (Fun-ASR, heavy load, 2 rounds per arm, per review feedback): with the load roaming the lane, `off` measures 32.3/60.1 QPS (placement is chaotic) and `static` measures 61.7/58.3, so server pinning alone stabilizes the off arm at its best case; confining the load to the shared pool lifts `static` to 149.2/150.2. The headline ratios combine both effects: the static arm models the posture where colocated load is placed by the same plan. Higgs read 0.95x at no load during the campaign, when four models ran in parallel and its static rounds happened to sit under heavier background load (43 to 95 vs 29 to 69); an isolated re-run on a quiet box reads 1.01x (22.27 vs 22.58 QPS). Its pipeline process measures 1.17 cores of host work against the 2 physical cores it is granted, so the declaration is not tight.

Cost when the lane has spare capacity and the load generator shares it: with the closed-loop client inside the same lane as the server and no other tenants (the CI gate shape), `static` reads 150.5 vs 154.7 QPS over three rounds each, so the carve costs about 3% while tightening the spread only slightly (1.1 vs 1.7 QPS range). With the client on another node the same configuration reads 1.00x, which points at the client being squeezed rather than the server. The allocator is aimed at deployments whose lane co-tenants are pipeline processes and neighbors, not at a lane shared with the measurement harness.

NUMA ablation (per review discussion): with identical manual exclusive carving under heavy load, Fun-ASR measures 110.5/112.9 QPS on the GPU-local socket vs 113.0/113.8 on the far socket, boot-time first-touch included. GPU-node anchoring has no measured latency benefit on this H200, so the serving-side GPU-to-NUMA resolution chain was removed: exclusive grants now spread across nodes by capacity, and the node dimension earns its keep as capacity budgeting plus whole-core SMT semantics. The DP planning CLI keeps its node selection because that is what makes the blocks node-local in the first place. Given those blocks, launch.sh's own `--cpunodebind`/`--membind` is measurably free to drop: Qwen3-ASR DP2 at c32, three rounds per arm A-B-A-B-A-B with core pinning kept in both, reads 138.1 QPS with the binding and 139.2 without, zero request errors, so the binding earns nothing once the cores are already on the GPU's node.

Not in the table: Whisper (crashes at c8+ with the allocator off, upstream bug, filed as #1470) and Qwen3-TTS (known transformers incompatibility). Excluded rounds are recorded with reasons. A memory-streaming neighbor (64 MB per process) at the moderate level reproduces the spin result on Fun-ASR (147.4 vs 150.3 QPS, 1.02x), so the effect is not an artifact of cache-idle spin load. The autodp integration is verified end to end on hardware: the shell path emits byte-identical blocks to the planner CLI.

`/host_contention` drill: 3.6 foreign cores at idle (real neighbor leak), 26.2 within one interval under a 24-proc load, back to 3.5 after it stops.

## Review follow-ups

Six findings from @stephenkgli's pass, each with a failing-before test:

| finding | what was wrong | fix |
| --- | --- | --- |
| demoted process gets an empty mask | a process pushed out of the exclusive pool could end up with no mask, which the caller skips, so it kept the full cpuset and ran on top of every "exclusive" core | the shared pool is reserved whenever anyone will land in it, and the demoted process always gets a real, disjoint mask |
| projection never rolled back | a demand was charged to a node before the fit test, so an oversized demand stranded capacity there and pushed later demands off the box | placement is fit-then-charge, and a demand that does not fit costs nothing |
| serving parent unplanned | the process running uvicorn, the Coordinator and the relay kept the whole cpuset and competed with the cores its own children were promised | the parent is planned and seated on the granted cores; the shared pool was the wrong place for it, measured 33.2 QPS against 129.6 because every request goes through it |
| reaped children lost from accounting | a child's cpu time leaves `/proc` when it is reaped, so the tree's total walked backwards and the drop was reported as neighbour load | the last reading of a departed process is carried forward, and own cpu is floored at zero |
| monitor thread per `create_app()` | embedders and tests that build several apps leaked a sampling thread each, and `start()` after `stop()` never sampled again | one sampler per process, and `stop()` no longer wedges a restart |
| autodp fallback too permissive | every planner failure, including a real capacity refusal, silently became the NUMA-blind 3/4 split, which can break launch.sh's node-local contract | the planner exits 3 on a refusal, the fallback is reserved for an unavailable planner and says so, and blocks are validated for count, emptiness and overlap |

## What the plan is worth, decomposed

The headline table above pairs two things: the server is planned, and the colocated load is placed by that same plan. Four arms on one lane separate them (Fun-ASR, 48-process neighbour, 2 rounds each, QPS):

| lane | neighbour roams, unplanned | neighbour roams, planned | neighbour in the pool, unplanned | neighbour in the pool, planned |
| 8 physical cores (grant 5) | 37.6 | 37.3 | 128.2 | 129.1 |
| 16 physical cores (grant 5) | 46.5 | 47.4 | 125.9 | 125.5 |

Confining the neighbour is worth 2.7 to 3.4x. Planning the server on top of that is within noise in both directions and at both lane widths, so the allocator earns its place by placing every process this server owns, correctly and automatically, not by the exclusive grant itself. On a lane with no colocated load at all, five models under DP measure 0.92 to 1.02x against hand-written CORE_BLOCKS, so the plan is a wash there too and its value is that the blocks are derived rather than written by hand.

`auto` is optional and off by default. It was built to remove a standing cost that turned out to be the parent placement bug fixed here, so `static` no longer carries that cost; it is kept because it does what it says. Validated on hardware across the three phases (Fun-ASR, 8-core lane): quiet 146.9 QPS with the plan never applied, contended 36.0 after it applies within 45 s, and 146.8 once the neighbour leaves and it releases, against 147.8 / 42.5 / 145.8 for `off`. The machinery does what it says; applying the plan is throughput neutral either way.

Router (a separate process outside the plan, raised in review): it draws about 0.85 of a core at c128. Twelve rounds per scenario, four arms, QPS:

| scenario | nothing pinned | workers pinned | workers and router pinned | plus the allocator |
| quiet lane | 77.7 | 74.6 | 76.0 | 74.2 |
| 48-process neighbour | 40.5 | 40.0 | 39.8 | 39.5 |

Giving the router its own cores is within noise in both scenarios, so nothing is added for it here.

Two corrections to earlier numbers in this PR. The `WORKER_FAILED` rounds reported in an earlier router campaign were a harness fault, not a startup regression: the launcher moves to a free port when the one it is given is still held by the previous round, so the health check polled a port nothing was listening on. Rounds are now given fresh ports and the banner is checked. And the serving parent, added to the plan in this round, must sit with the stages rather than in the shared pool: seating it in the pool measured 33.2 QPS against 129.6, four rounds each, because every request goes through it.

## Cross-review

A second-model adversarial review of the design and diff produced 28 plus findings, all adjudicated; thread-level affinity replay, spawn-time seeding, per-rank TP NUMA anchoring, and the OMP cap came out of it.

## Checklist

- [x] Add unit tests.


sglang_omni/cpu_alloc/allocator.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+def _anchor_node(
+    demand: ProcessCpuDemand,
+    node_states: dict[int, _NodeState],
+    events: list[str],
+) -> int:
+    if demand.numa_node is not None and demand.numa_node in node_states:
+        return demand.numa_node
+    fallback = max(node_states, key=lambda n: len(node_states[n].free_cores))
```
AkazaAkane: The caller uses None when GPU-to-NUMA resolution fails, including UUID-based CUDA_VISIBLE_DEVICES. On a dual-NUMA host this can hard-pin a GPU process to the wrong socket, which may be worse than leaving it unpinned and contradicts the “planning without NUMA anchoring” behavior logged by the caller.

I think unresolved GPU NUMA should not result in a guessed hard affinity.

examples/mps_dp/autodp.sh
```diff
@@ -103,6 +103,16 @@ USED_MIB=$(nvidia-smi --query-gpu=memory.used --format=csv,noheader,nounits -i "
 
 # ---- derive CORE_BLOCKS lazily (server share = 3/4 of the cpuset) ------------
 core_blocks_for() {
```
AkazaAkane, line 129: Can we verify this path actually executes the new planner?

The command in the current diff appears to contain malformed line continuations and the sed replacement does not look like a normal capture-group reference. Since failures here are silently converted into the old fallback path, this can make the integration appear to work while never exercising the NUMA/SMT-aware planner.

A small test around the emitted CORE_BLOCKS would also protect this path from silently regressing.

AkazaAkane (commented):
I think this A/B changes two variables.

The allocator only changes affinity for this server's processes; it does not move foreign spin processes into the shared pool. If the load generator roams in the off arm but is externally confined in the static arm, the measured delta includes both server pinning and different interference placement.

Since the 1.4–3.5x numbers are the main evidence for this PR, can we keep the foreign-load affinity identical between the two arms, or clearly separate the benefit from external cpuset isolation versus the allocator itself?

sglang_omni/cpu_alloc/allocator.py
JiaxinD: Fixed in f8aee6d8: a demand without a resolvable NUMA anchor now stays in the shared pool with a logged event, no guessed pinning.

examples/mps_dp/autodp.sh
JiaxinD: Confirmed, this path always took the silent fallback (stray CR plus a corrupted sed replacement). Fixed in f8aee6d8: the planner prints the bare value via --format blocks, consumed verbatim and unit-tested.

AkazaAkane (commented):
I think degradation needs a group-level invariant before we do per-process fallback here.

For example, if N equivalent replicas each request 2 exclusive cores but the lane can only satisfy N-1 of them, the current policy can leave N-1 replicas exclusive and move one replica to the shared pool. Which replica loses exclusivity is then effectively determined by process naming.

That creates heterogeneous replicas inside the same pool, which can produce persistent throughput/tail-latency imbalance and is harder to reason about than failing open.

Could we define the degradation unit explicitly? My preference for the first version would be: equivalent replicas should receive equivalent CPU allocations. Parallel widths may shrink uniformly across the group; if the minimum symmetric allocation cannot fit, degrade the whole group to shared rather than individual replicas.

JiaxinD:
Fair point, ran the mssing(Fun-ASR, heavy load, 2 rounds per arm):

| arm | QPS |
| --- | ---: |
| off, load roams the lane | 32.3 / 60.1 |
| static, load roams identically | 61.7 / 58.3 |
| static, load in the shared pool | 149.2 / 150.2 |

Server pinning alone stabilizes the chaotic off arm at its best case; confining the managed load adds the rest. Body updated to attribute the two effects.

sglang_omni/cpu_alloc/allocator.py
```diff
@@ -0,0 +1,221 @@
...
+            granted[demand.process_name] = 0
+            events.append(
+                f"node {node}: {demand.process_name} wants "
+                f"{demand.exclusive_cores} core(s) but only {max(budget, 0)} "
+                f"remain; moved to the shared pool"
+            )
+
+    # Note (Jiaxin Deng): with no shared tenant the leftover would idle, so
+    # the exclusive holders take the whole node between them.
+    if not has_shared_tenant:
```
stephenkgli, line 187: A process demoted from exclusive to shared can receive an empty CPU mask. `has_shared_tenant` is computed before demotion, so this block may give every remaining core to successful exclusive holders; `_attach_cpu_plan` then skips the empty mask and the demoted process inherits the full cpuset, overlapping all "exclusive" cores.

examples/mps_dp/autodp.sh
```diff
@@ -103,6 +103,14 @@ USED_MIB=$(nvidia-smi --query-gpu=memory.used --format=csv,noheader,nounits -i "
...
+  local planned
+  # Note (Jiaxin Deng): NUMA/SMT-aware planner, whole physical cores per
+  # replica on the GPU's node; falls back to the plain 3/4 split on failure.
+  planned=$(PYTHONPATH="$HERE/../..${PYTHONPATH:+:$PYTHONPATH}" python3 -m sglang_omni.cpu_alloc plan --replicas "$1" --gpu-id "$GPU_ID" --format blocks 2>/dev/null) || planned=""
```
stephenkgli: [P0-BLOCKER] autodp sizes the replica count from all visible logical CPUs, but this planner only accepts physical cores on the GPU NUMA node. When local capacity is smaller, the planner error is suppressed and the fallback can cross NUMA nodes or split SMT siblings, producing `CORE_BLOCKS` that conflict with the `--cpunodebind` and `--membind` settings in `launch.sh`.

sglang_omni/cpu_alloc/host_metrics.py
```diff
@@ -0,0 +1,188 @@
...
+    try:
+        data = (
+            (proc_root / str(pid) / "stat")
+            .read_bytes()
+            .decode("ascii", errors="replace")
+        )
+    except OSError:
+        return 0
+    fields = data.rsplit(")", 1)[1].split()
+    return int(fields[11]) + int(fields[12])
```
stephenkgli, line 75: This reads only `utime` and `stime` from currently live processes. After an owned child such as ffmpeg or gzip is waited for, it disappears from the process tree and its CPU time moves to `cutime` and `cstime`, so the monitor can report negative own CPU and foreign CPU above the cpuset capacity.

sglang_omni/cpu_alloc/pipeline_plan.py
```diff
@@ -0,0 +1,115 @@
...
+            if s in costs and costs[s].host_class == "serial-loop"
+        )
+        if process_name in replicated and exclusive:
+            logger.warning(
+                "cpu_alloc: process %s is replicated; per-replica planning is "
+                "not supported yet, keeping it in the shared pool",
+                process_name,
+            )
+            exclusive = 0
+        demands.append(
```
stephenkgli, line 142: The plan creates CPU demands only for stage processes. The serving parent that runs Coordinator, Client, request handling, the monitor, and Uvicorn keeps the full inherited cpuset, so it can execute on every core advertised as exclusive to a stage.

sglang_omni/cpu_alloc/allocator.py
```diff
@@ -0,0 +1,221 @@
...
+        (d for d in demands if d.exclusive_cores),
+        key=lambda d: d.process_name,
+    )
+    anchored: dict[str, int] = {}
+    projected = dict.fromkeys(node_states, 0)
+    for demand in exclusive_demands:
+        node = _anchor_node(demand, node_states, projected, events)
+        if node is not None:
+            anchored[demand.process_name] = node
+            projected[node] += demand.exclusive_cores
```
stephenkgli: The placement pass adds every unanchored demand to projected before checking whether it can fit on the selected node. If an oversized demand is later demoted to shared, its projection is never rolled back and later demands are not retried. This can overflow another node while usable exclusive capacity remains stranded on the original node.

sglang_omni/serve/openai_api.py
```diff
@@ -401,6 +402,21 @@ async def health() -> JSONResponse:
...
 
 
+def _register_host_contention(app: FastAPI) -> None:
+    from sglang_omni.cpu_alloc.host_metrics import HostCpuContentionMonitor
+
+    # Note (Jiaxin Deng): daemon thread for the process lifetime; this
+    # FastAPI build has no shutdown hooks, stop() is for tests and embedders.
+    monitor = HostCpuContentionMonitor()
+    app.state.host_contention_monitor = monitor
+    monitor.start()
```
stephenkgli, line 419: Every `create_app()` call starts a daemon monitor immediately, but no app shutdown path stops it. Tests or embedders that construct multiple FastAPI apps leak one thread per app, and the monitor cannot be cleanly restarted because `start()` does not clear the stop event.

sglang_omni/cpu_alloc/pipeline_plan.py
stephenkgli: Also we should take router into consideration

examples/mps_dp/autodp.sh
stephenkgli: `D_MAX` and `CORE_BLOCKS` are computed from different CPU capacities. [`SRV_CORE_COUNT`](https://github.com/sgl-project/sglang-omni/blob/8e6f0cfe9b61865710959fd1eafd6d314502d69b/examples/mps_dp/autodp.sh#L165-L183) is based on three quarters of all visible logical CPUs, while [`plan_replica_blocks()`](https://github.com/sgl-project/sglang-omni/blob/8e6f0cfe9b61865710959fd1eafd6d314502d69b/sglang_omni/cpu_alloc/__main__.py#L41-L54) counts whole physical cores only on the GPU NUMA node. For example, on a dual-socket SMT2 host with 8 physical cores per node (32 logical CPUs total), the default `MIN_CORES_PER_REPLICA=2` lets sizing choose `N=12` from `24 / 2`, but the planner sees only 8 physical cores on the GPU node and correctly rejects 12 replicas.

[That capacity error is discarded and the logical-CPU fallback is used](https://github.com/sgl-project/sglang-omni/blob/8e6f0cfe9b61865710959fd1eafd6d314502d69b/examples/mps_dp/autodp.sh#L109-L126). The fallback splits the first 24 logical CPU IDs directly, so it does not guarantee that SMT siblings stay together or that blocks remain on the GPU node. This violates [`launch.sh`'s node-local `CORE_BLOCKS` contract](https://github.com/sgl-project/sglang-omni/blob/8e6f0cfe9b61865710959fd1eafd6d314502d69b/examples/mps_dp/launch.sh#L327-L334), and each replica is later started with [`--cpunodebind`, `--membind`, and `-C`](https://github.com/sgl-project/sglang-omni/blob/8e6f0cfe9b61865710959fd1eafd6d314502d69b/examples/mps_dp/launch.sh#L489), which may fail or apply a placement different from the advertised plan. [`autodp.sh plan` also exits before calling `core_blocks_for`](https://github.com/sgl-project/sglang-omni/blob/8e6f0cfe9b61865710959fd1eafd6d314502d69b/examples/mps_dp/autodp.sh#L210-L213), so it can report this invalid `D_MAX` without validating any blocks.

Please derive the CPU limit from the same GPU-node physical-core topology, treat a capacity rejection as fatal, and validate the generated blocks in `plan` mode. The fallback should only cover unavailable topology discovery.

examples/mps_dp/autodp.sh
```diff
@@ -103,6 +103,14 @@ USED_MIB=$(nvidia-smi --query-gpu=memory.used --format=csv,noheader,nounits -i "
 
 # ---- derive CORE_BLOCKS lazily (server share = 3/4 of the cpuset) ------------
 core_blocks_for() {
+  local planned
+  # Note (Jiaxin Deng): NUMA/SMT-aware planner, whole physical cores per
+  # replica on the GPU's node; falls back to the plain 3/4 split on failure.
```
stephenkgli: This fallback treats every planner failure as permission to use the topology-naive split. That is unsafe because the planner can reject a plan for a real capacity mismatch, not only because topology discovery is unavailable. [`SRV_CORE_COUNT`](https://github.com/sgl-project/sglang-omni/blob/8e6f0cfe9b61865710959fd1eafd6d314502d69b/examples/mps_dp/autodp.sh#L165-L183) is based on three quarters of all visible logical CPUs, while [`plan_replica_blocks()`](https://github.com/sgl-project/sglang-omni/blob/8e6f0cfe9b61865710959fd1eafd6d314502d69b/sglang_omni/cpu_alloc/__main__.py#L41-L54) counts whole physical cores only on the GPU NUMA node. For example, on a dual-socket SMT2 host with 8 physical cores per node (32 logical CPUs total), the default `MIN_CORES_PER_REPLICA=2` lets sizing choose `N=12` from `24 / 2`, but the planner sees only 8 physical cores on the GPU node and correctly rejects 12 replicas.

The command on the next line discards that error, then the fallback splits the first 24 logical CPU IDs directly. It therefore does not guarantee that SMT siblings stay together or that blocks remain on the GPU node. This violates `launch.sh`'s node-local `CORE_BLOCKS` contract, and each replica is later started with `--cpunodebind`, `--membind`, and `-C`, which may fail or apply a placement different from the advertised plan. `autodp.sh plan` also exits before calling `core_blocks_for`, so it can report this invalid `D_MAX` without validating any blocks.

sglang_omni/cpu_alloc/allocator.py
JiaxinD: Fixed in 579eb608: the shared pool is reserved whenever anyone will land in it, so a demoted process always gets a real, disjoint mask.
Test: `test_demoted_process_gets_a_mask_when_every_node_is_taken` (fails before).

sglang_omni/cpu_alloc/allocator.py
JiaxinD: Fixed in 579eb608: placement is fit-then-charge, so a demand that does not fit costs nothing and later demands still place.
Test: `test_a_demotion_frees_the_capacity_it_asked_for`.

sglang_omni/cpu_alloc/host_metrics.py
JiaxinD: Fixed in f19d3c90: the last reading of a departed process is carried forward and own cpu is floored at zero.
Test: `test_a_child_exiting_is_not_read_as_foreign_load`.

sglang_omni/cpu_alloc/pipeline_plan.py
JiaxinD: Fixed in f4e9fa65 and 13988945: the parent is planned, seated with the stages rather than in the pool.
Note: the pool placement I tried first measured 33.2 QPS against 129.6 (four rounds each), since every request goes through it.

sglang_omni/serve/openai_api.py
JiaxinD: Fixed in f19d3c90 and 0e9b1e66: one sampler per process, shared with the auto supervisor, and `stop()` no longer wedges a restart.

examples/mps_dp/autodp.sh
JiaxinD: Fixed in b49bf7f0: the planner exits 3 on a capacity refusal, the fallback is reserved for an unavailable planner and says so, and blocks are validated for count, emptiness and overlap before launch.
`autodp.sh plan` now prints and validates them too.

sglang_omni/cpu_alloc/pipeline_plan.py
JiaxinD: Measured it: the router draws about 0.85 of a core at c128, and giving it its own cores reads within noise on a quiet lane (76.0 vs 77.7) and under a 48-process neighbour (39.8 vs 40.5), twelve rounds per scenario.
Table is in the PR body; happy to add a router block to the planner CLI if you have a shape where it does pay.

PR #1472, opened 2026-08-11
perf(qwen3-asr): vectorize decode-step mrope positions

0xjeffro:
## Motivation

Qwen3-ASR inherits mrope_section from the Omni thinker config, so sglang rebuilds mrope_positions with a per-request loop on every decode step. ASR has no spatial axes, so every request's position is seq_len - 1 and the loop is pure overhead.

## Modifications

In this pr, we replace it with a single broadcast for decode batches stamped degenerate by the ASR request builder; everything else falls back to upstream.

Concretely, `sglang_omni/models/qwen3_asr/mrope_fast_path.py` monkey-patches `ForwardBatch._compute_mrope_positions` when engine build (from `Qwen3ASREngineBuilder.validate_before_infrastructure`, which runs in the process
that owns the decode loop), keeping the original bound as the fallback. The computation lives in sglang, so patching is how we override it without vendoring or forking that file.

## Related Issues

#1396 

## Accuracy Test

I test locally and measured on A6000, PE values are unchanged, and WER is also unaffected. 400 SeedTTS-EN clips at concurrency 32: decode launch.build 4.28ms -> 0.82ms per step, throughput +5.4% cold /+36.8% warm, corpus WER 0.0140 on both.

## Checklist

- [x] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/models/qwen3_asr/engine_builder.py
+++ b/sglang_omni/models/qwen3_asr/engine_builder.py
@@ -9,7 +9,7 @@ from typing import Any
 from sglang.srt.managers.mm_utils import init_mm_embedding_cache
 from transformers import AutoFeatureExtractor, AutoTokenizer
 
-from sglang_omni.models.qwen3_asr import request_builders
+from sglang_omni.models.qwen3_asr import mrope_fast_path, request_builders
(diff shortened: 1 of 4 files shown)


SandyLuXY (approved):
lgtm

PR #1475, opened 2026-08-11
[TTS][Qwen3] Support deterministic inference

0xtoward:
## Motivation
Section 3.4 #1399
Qwen3-TTS can produce different codec trajectories and PCM for the same prompt,
reference audio, and seed when the physical runtime batch changes. 

The investigation found three batch-dependent boundaries:

- The first observed B1/B6 Talker difference was at decoder layer 0, position 58, in the QKV projection path. The difference was fed back into later codec steps.
- With identical codec inputs, the tokenizer decoder still changed with its physical batch shape.
- After fixing the Talker and vocoder paths, fresh runs at the same concurrency still diverged. The remaining difference came from concurrent reference-audio preprocessing.

SGLang deterministic inference covers the Talker reduction path, but Qwen3-TTS still applied its private per-layer compile path, batched vocoder work, and concurrent preprocessing.

## Modifications

When `enable_deterministic_inference` is enabled:

- Serialize reference-audio preprocessing.
- Enable SGLang deterministic inference for the Talker.
- Skip Qwen3-TTS's private per-layer `torch.compile` path.
- Decode streaming plans and non-streaming vocoder payloads at physical batch size one.
- Capture and replay the initial vocoder CUDA Graph at batch size one.

The existing preprocessing concurrency, compiled Talker, batched vocoder, and vocoder graph buckets remain unchanged when the flag is disabled.

## Accuracy Test

B1:
[A]
homogeneous B8:
[A, A, A, A, A, A, A, A]
Mixed B8:
[A, B, C, D, E, F, G, H]

| Scope | Result |
| --- | --- |
| 0.6B Base, 3 Seed-TTS prompts | B1 repeats and homogeneous B8 produced byte-identical PCM |
| 0.6B Base, mixed B8 | Each of the eight requests matched its own B1 PCM oracle; observed Talker batch size was 8 |
| 1.7B Base, 2 Seed-TTS prompts | B1, homogeneous B6, and homogeneous B8 produced byte-identical PCM |
| B1/B16/B32 replay | 185 requests produced identical codec, waveform, and PCM hashes per sample |
| Full Seed-TTS EN, C16 vs C32 | 1088 / 1088 PCM byte-identical; no failed requests |
| Qwen3-ASR 1.7B evaluation | 1088 / 1088 evaluated; corpus WER 0.01155; no skips or failures |
| Unit tests | `tests/unit_test/qwen3_tts/test_pipeline.py`: 115 passed |
| GPU model tests | `tests/test_model/test_qwen3_tts_batch_invariance.py`: 2 passed |

## Benchmark & Profiling

| Concurrency | Baseline | Deterministic Proto | Throughput|
|---|---:|---:|---:|
| C1 | 1.996 req/s | 0.639 req/s | **-68.0%** |
| C8 | 7.128 req/s | 2.992 req/s | **-58.0%** |
| C32 | 12.153 req/s | 5.794 req/s | **-52.3%** |”

Serializing preprocessing changed full-corpus throughput by less than 1% relative to the preceding deterministic implementation with concurrent preprocessing.

## Checklist

- [x] Compare identical-prompt logits between B1 and B6, and identify the first divergence.
- [x] Identify batch-size-dependent execution paths and pin a deterministic opt-in path.
- [x] Add same-seed, same-prompt B1/B8 PCM byte-identity coverage with actual scheduler batches.
- [x] Pass the full TTS correctness evaluation at C16 and C32 and report the updated numbers.


tests/test_model/test_qwen3_tts_batch_invariance.py
```diff
@@ -0,0 +1,189 @@
...
+    stop = asyncio.Event()
+    poller = asyncio.create_task(_model_info(base_url, stop))
+    timeout = aiohttp.ClientTimeout(total=300)
+    try:
+        async with aiohttp.ClientSession(timeout=timeout) as session:
+            b1 = [
+                await _generate(session, base_url, payload)
+                for payload in serial_payloads
+            ]
+            b8 = await asyncio.gather(
```
AkazaAkane: B1 runs before B8 with the same references, but Qwen3-TTS caches reference-encode artifacts by reference/options. The B1 oracle can therefore populate the cache and let B8 skip fresh reference preprocessing entirely. Since preprocessing was one of the three nondeterministic boundaries found here, can we make this test exercise uncached references as well—for example by clearing/restarting the reference service between oracle and batch—and verify that path actually ran?

tests/test_model/test_qwen3_tts_batch_invariance.py
```diff
@@ -0,0 +1,189 @@
...
+        port=port,
+        host="127.0.0.1",
+        log_file=log_file,
+        server_config=str(config_path),
+        timeout=STARTUP_TIMEOUT,
+    ):
+        yield f"http://127.0.0.1:{port}"
+
+
+@pytest.mark.benchmark
```
AkazaAkane: These are the regression tests for the behavior this PR fixes, but marking them benchmark means PR CI does not execute them: the generic suite uses -m "not benchmark", and TTS CI only invokes its existing fixed test files. Can we wire this invariance coverage into TTS CI, or add a CI-selected deterministic smoke variant? Otherwise this regression can come back while CI stays green.

AkazaAkane (commented):
Is opt-in determinism the intended acceptance contract for #1399 §3.4? With the default False, same-seed requests can still change with physical batching unless the operator enables this flag. Given the reported throughput cost, opt-in seems reasonable, but then we should document explicitly that batch-invariant/byte-identical output is guaranteed only with deterministic inference enabled and state the performance tradeoff.

0xtoward:
> Is opt-in determinism the intended acceptance contract for #1399 §3.4? With the default False, same-seed requests can still change with physical batching unless the operator enables this flag. Given the reported throughput cost, opt-in seems reasonable, but then we should document explicitly that batch-invariant/byte-identical output is guaranteed only with deterministic inference enabled and state the performance tradeoff.

I'll set this PR intentionally opt-in because of the current throughput. BTW, In a separate prototype, a small-M batch-invariant GEMM scratch kernel recovered 92.8%, 47.5%, and 28.9% throughput over the current deterministic path at C1, C8, and C32. I am happy to follow up with a separate performance PR.

I also expanded the invariance validation beyond the targeted B1/B8 regression. For each case, I compared every batched row with the same request’s fresh-process B1 PCM oracle. All 120 comparisons matched byte-for-byte, despite changing the physical batch size (B2/B4/B8), neighboring requests, and row order.

tests/test_model/test_qwen3_tts_batch_invariance.py
0xtoward: The test in new commit uses two fresh servers: the B1 oracle runs on the first, while eight distinct references are the first requests to the B8 server; it also verifies the reference encoder starts with zero hits and a cache miss.

tests/test_model/test_qwen3_tts_batch_invariance.py
0xtoward: I added a standalone H100 job to the TTS CI workflow, but I’d appreciate guidance if there is a lighter preferred place for this model-level smoke.

AkazaAkane:
LGTM, can you resolve the conflict so I can help run CI?

AkazaAkane:
Thanks for contributing. Could we remove the CI and update the related documents for the qwen3-tts on enable_deterministic_inference? Some note on why we set default as false will be helpful (please follow the sglang-omni comment guideline).

0xtoward:
> Thanks for contributing. Could we remove the CI and update the related documents for the qwen3-tts on enable_deterministic_inference? Some note on why we set default as false will be helpful (please follow the sglang-omni comment guideline).

fixed :)

AkazaAkane (approved):

zhaochenyang20:
@0xtoward hello, Recently saw a lot of your amazing PRs, and I'm pondering: do you use WeChat or Slack? I want to have a discussion with you offline. Here is my wechat ID LoveDeathAndLLM

And, slack of sglang : slack.sglang.ai

Please DM Chayenne Zhao on slack. thanks so much~

PR #1476, opened 2026-08-11
[NPU] Qwen3-Omni talker support

Makcum888e:
## Motivation

Support Qwen3-Omni talker on Ascend NPU to enable audio output

## Modifications

replaced `torch.cuda` with general `torch.get_device_module()`
added new functions to `current_platform`
- `get_intra_node_transport` to allow every platform choose their backend for move between devices on the same node
- `get_fused_qk_norm_rope` to import sgl-kernel only for CUDA
- `apply_model_worker_backend_policy` moved from `model_worker` as it has many CUDA related adjustment

`normalize_quantization` and `model_config_has_moe` moved from `model_worker` to `utils/misc.py`

## Accuracy Test

```
sgl-omni serve --model-path /home/weights/Qwen/Qwen3-Omni-30B-A3B-Instruct/ --thinker-cuda-graph off --thinker-gpus 2,3 --thinker-tp-size 2 --talker-gpu 1 --code2wav-gpu 0 --talker-cuda-graph off --image-encoder-gpus 0
```
```
curl -X POST http://localhost:8000/v1/chat/completions -H "Content-Type: application/json" -d '{"model": "qwen3-omni", "messages": [{"role": "user", "content": "Say the word \"generate\" exactly ten times in a row. Do not add any other words. Example: one, one, one... Repeat it exactly 10 times."}], "temperature": 0.01, "top_p": 0.1, "presence_penalty": 0.0, "frequency_penalty": 0.0, "modalities": ["text", "audio"]}' -o output.json
```

## Checklist

- [ ] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/comm/router.py
+++ b/sglang_omni/comm/router.py
@@ -8,6 +8,7 @@ from typing import Any
 import torch
 
 from sglang_omni.comm.data_ref import TransportKind
+from sglang_omni.platforms import current_platform
 from sglang_omni.relay.base import Relay, create_relay
 
 
@@ -77,7 +78,7 @@ class CommRouter:
         if target in self.remote_stage_names:
             return TransportKind.MOONCAKE
         if self.self_is_gpu and target in self.gpu_stage_names:
-            return TransportKind.CUDA_IPC
+            return current_platform.get_intra_node_transport()
         return TransportKind.SHM
 
     def outbound_stream(self, target: str, data: torch.Tensor) -> TransportKind:
@@ -91,7 +92,7 @@ class CommRouter:
         if not data.is_cuda:
             return TransportKind.SHM
         if self.self_is_gpu and target in self.gpu_stage_names:
-            return TransportKind.CUDA_IPC
+            return current_platform.get_intra_node_transport()
         raise ValueError(
             f"cuda stream chunk cannot be sent from {self.stage_name!r} to "
             f"non-GPU target {target!r}"
@@ -101,7 +102,7 @@ class CommRouter:
         if from_stage in self.remote_stage_names:
             return TransportKind.MOONCAKE
         if self.self_is_gpu and from_stage in self.gpu_stage_names:
-            return TransportKind.CUDA_IPC
+            return current_platform.get_intra_node_transport()
         return TransportKind.SHM
 
     def relay(self, kind: TransportKind) -> Relay:
@@ -138,9 +139,12 @@ class CommRouter:
         devices = _tensor_devices(getattr(payload, "data", payload))
         if not devices or devices == {"cpu"}:
             return TransportKind.SHM
-        if "cuda" in devices and devices <= {"cpu", "cuda"}:
+        if current_platform.device_type in devices and devices <= {
+            "cpu",
+            current_platform.device_type,
+        }:
             if self.self_is_gpu and target in self.gpu_stage_names:
-                return TransportKind.CUDA_IPC
+                return current_platform.get_intra_node_transport()
             return TransportKind.SHM
         raise ValueError(f"mixed or unsupported tensor devices in payload: {devices}")
 
--- a/sglang_omni/comm/router.py
+++ b/sglang_omni/comm/router.py
@@ -62,7 +62,10 @@ class CommRouter:
         return target in self.same_process_targets
 
     def can_use_direct_cuda_ipc(self, target: str) -> bool:
-        return target in self._direct_cuda_ipc_targets
+        return (
+            target in self._direct_cuda_ipc_targets
+            and current_platform.get_intra_node_transport() == TransportKind.CUDA_IPC
+        )
 
     def outbound(self, target: str) -> TransportKind:
         if target in self.same_process_targets:
--- a/sglang_omni/models/qwen3_omni/components/code2wav_cuda_graph.py
+++ b/sglang_omni/models/qwen3_omni/components/code2wav_cuda_graph.py
@@ -58,28 +58,30 @@ class _TorchCudaApi:
     """Small injectable boundary around CUDA-only operations."""
 
     def device_context(self, device: torch.device) -> AbstractContextManager[Any]:
-        return torch.cuda.device(device)
+        return torch.get_device_module().device(device)
 
     def memory_stats(self, device: torch.device) -> dict[str, int]:
-        free_bytes, total_bytes = torch.cuda.mem_get_info(device)
+        free_bytes, total_bytes = torch.get_device_module().mem_get_info(device)
         return {
-            "allocated_bytes": int(torch.cuda.memory_allocated(device)),
-            "reserved_bytes": int(torch.cuda.memory_reserved(device)),
-            "max_reserved_bytes": int(torch.cuda.max_memory_reserved(device)),
+            "allocated_bytes": int(torch.get_device_module().memory_allocated(device)),
+            "reserved_bytes": int(torch.get_device_module().memory_reserved(device)),
+            "max_reserved_bytes": int(
+                torch.get_device_module().max_memory_reserved(device)
+            ),
             "free_bytes": int(free_bytes),
             "total_bytes": int(total_bytes),
         }
 
     def empty_cache(self) -> None:
-        torch.cuda.empty_cache()
+        torch.get_device_module().empty_cache()
 
     def new_static_input(
         self, shape: tuple[int, int, int], *, device: torch.device
     ) -> torch.Tensor:
         return torch.zeros(shape, dtype=torch.long, device=device)
 
-    def new_stream(self, device: torch.device) -> torch.cuda.Stream:
-        return torch.cuda.Stream(device=device)
+    def new_stream(self, device: torch.device) -> torch.Stream:
+        return torch.get_device_module().Stream(device=device)
 
     def warmup(
         self,
@@ -88,17 +90,17 @@ class _TorchCudaApi:
         *,
         iterations: int,
         device: torch.device,
-        stream: torch.cuda.Stream,
+        stream: torch.Stream,
     ) -> None:
-        current_stream = torch.cuda.current_stream(device)
+        current_stream = torch.get_device_module().current_stream(device)
         stream.wait_stream(current_stream)
-        with torch.cuda.stream(stream), torch.inference_mode():
+        with torch.get_device_module().stream(stream), torch.inference_mode():
             for _ in range(iterations):
                 model(static_input)
         current_stream.wait_stream(stream)
 
     def graph_pool_handle(self) -> Any:
-        return torch.cuda.graph_pool_handle()
+        return torch.get_device_module().graph_pool_handle()
 
     def capture(
         self,
@@ -106,14 +108,14 @@ class _TorchCudaApi:
         static_input: torch.Tensor,
         *,
         pool: Any,
-        stream: torch.cuda.Stream,
-    ) -> tuple[torch.cuda.CUDAGraph, torch.Tensor]:
-        current_stream = torch.cuda.current_stream(static_input.device)
+        stream: torch.Stream,
+    ) -> tuple[Any, torch.Tensor]:
+        current_stream = torch.get_device_module().current_stream(static_input.device)
         stream.wait_stream(current_stream)
-        graph = torch.cuda.CUDAGraph()
+        graph = torch.get_device_module().CUDAGraph()
         try:
             with torch.inference_mode():
-                with torch.cuda.graph(
+                with torch.get_device_module().graph(
                     graph,
                     pool=pool,
                     stream=stream,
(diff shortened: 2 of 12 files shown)


sglang_omni/comm/router.py
```diff
@@ -77,7 +78,7 @@ def _physical_outbound(self, target: str) -> TransportKind:
         if target in self.remote_stage_names:
             return TransportKind.MOONCAKE
         if self.self_is_gpu and target in self.gpu_stage_names:
-            return TransportKind.CUDA_IPC
+            return current_platform.get_intra_node_transport()
```
Ratish1, line 84: The default talker_ar and code2wav stages run in separate processes on the same device, so code2wav is added to _direct_cuda_ipc_targets. SGLang’s NPU bootstrap also makes an NPU tensor report is_cuda=True, which causes the direct CUDA-IPC branch in Stage._send_stream_chunk() to return before this platform transport policy is consulted.

Could we also require current_platform.get_intra_node_transport() == TransportKind.CUDA_IPC inside can_use_direct_cuda_ipc()? That keeps the existing CUDA shortcut but ensures NPU follows the SHM policy selected here.

sglang_omni/platforms/cuda.py
```diff
@@ -43,6 +63,126 @@ def get_stage_process_env(
...
+            )
+            moe_runner_backend = server_args.moe_runner_backend
+
+        if (
+            is_qwen3_omni_arch
+            and effective_quantization == "fp8"
+            and has_moe
+            and moe_runner_backend == "auto"
+            and has_native_fp8_block_quant
+            and _is_fp8_cutlass_moe_supported()
```
Ratish1, line 137: _is_fp8_cutlass_moe_supported() is not defined or imported in this module; its definition remains in model_worker.py. This branch therefore raises NameError for native block-FP8 Qwen3-Omni with  automatic MoE backend selection.

can we move the helper into this CUDA policy module, or into a dependency-neutral module, and update the tests to patch it at its new lookup location.

sglang_omni/platforms/cuda.py
```diff
@@ -43,6 +63,126 @@ def get_stage_process_env(
...
+        server_quantization = normalize_quantization(server_args.quantization)
+        if server_quantization is not None:
+            effective_quantization = server_quantization
+
+        moe_runner_backend = server_args.moe_runner_backend
+        is_qwen3_omni_arch = model_arch_override in (
+            "Qwen3OmniTalker",
+            "Qwen3OmniThinkerForCausalLM",
+        )
+        if is_qwen3_omni_arch and server_args.ep_size != 1:
```
Ratish1: This validation is an Omni ModelWorker invariant rather than CUDA backend policy. NPU now inherits the base policy and accepts ep_size > 1, while ModelWorker still constructs SGLModelRunner with moe_ep_size=1.

can we keep the Qwen3-Omni ep_size == 1 validation in common policy and leave only CUDA-specific backend selection in this override.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -265,8 +266,7 @@ def _forward_codes(
         graph_eligible: bool = False,
     ) -> tuple[torch.Tensor, dict[str, Any]]:
         with torch.no_grad():
-            if self._device.type == "cuda":
-                torch.cuda.set_device(self._device)
+            torch.get_device_module().set_device(self._device)
```
Ratish1: get_device_module() without an argument selects the process’s ambient accelerator rather than the scheduler’s configured device. On an accelerator host, a CPU Code2Wav scheduler therefore passes a CPU  device to CUDA/NPU set_device.

should we select the module from the device:

```python
torch.get_device_module(self._device).set_device(self._device)
```

sglang_omni/models/qwen3_omni/components/code2wav_cuda_graph.py
```diff
@@ -88,48 +90,48 @@ def warmup(
...
         pool: Any,
-        stream: torch.cuda.Stream,
-    ) -> tuple[torch.cuda.CUDAGraph, torch.Tensor]:
-        current_stream = torch.cuda.current_stream(static_input.device)
+        stream: torch.Stream,
+    ) -> tuple[Any, torch.Tensor]:
+        current_stream = torch.get_device_module().current_stream(static_input.device)
         stream.wait_stream(current_stream)
-        graph = torch.cuda.CUDAGraph()
+        graph = torch.get_device_module().CUDAGraph()
```
Ratish1: does this construct an NPU graph?. the standalone Code2Wav process resolves get_device_module() to torch.npu, which exposes NPUGraph, not CUDAGraph. The resulting AttributeError is caught during build, so the default graph-enabled Code2Wav configuration performs warmup and then silently disables every graph. The unchanged tensor.is_cuda validation would also reject native NPU inputs if construction were fixed.

should we implement NPU graphs in this PR or we keep this runner explicitly CUDA-only and disable Code2Wav graphs on NPU so it starts directly in eager mode?

tests/unit_test/qwen3_omni/test_fp8_backend_config.py
```diff
@@ -372,14 +374,14 @@ def test_model_worker_backend_policy_precedence(
 
     if case.error_match:
         with pytest.raises(ValueError, match=case.error_match):
-            model_worker._apply_model_worker_backend_policy(
+            current_platform.apply_model_worker_backend_policy(
```
Ratish1: current_platform is not imported, so this and the calls at lines 384, 417, and 761 raise NameError. The tests also still patch model_worker._is_fp8_cutlass_moe_supported, while production now looks up
that helper from the CUDA policy module.

lets exercise an explicit cuda.CUDAOmniPlatform() here and patch the helper in its module.

sglang_omni/platforms/cuda.py
Makcum888e: `_is_fp8_cutlass_moe_supported` moved to `cuda.py`

sglang_omni/platforms/cuda.py
Makcum888e: now I moved this logic to base class and call it with `super().` if this way is not suitable, I can create new function in `model_worker` for `ep_size` logic

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
Makcum888e: done

sglang_omni/models/qwen3_omni/components/code2wav_cuda_graph.py
Makcum888e: removed this changes, NPUGraph will be supported in future PR

tests/unit_test/qwen3_omni/test_fp8_backend_config.py
Makcum888e: done

sglang_omni/comm/router.py
Makcum888e: added this check `current_platform.get_intra_node_transport() == TransportKind.CUDA_IPC`, this seems like a reasonable check, however, this works before change and here's why:
In SGLang `transfer_to_npu` is used for replace some torch methods - it's why `is_cuda=True` in SGLang. But these workaround is not a universal solution and often creates more problems. So we not follow this way in SGL-Diffusion. I hope we can avoid using `transfer_to_npu` in SGL-Omni. So `is_cuda` will be `False` for NPU

sglang_omni/platforms/cuda.py
Ratish1: calling super() is correct. but since this restriction comes from ModelWorker always constructing SGLModelRunner with moe_ep_size=1, do you think ModelWorker would be the better owner for this check. CUDA can still call super() for the shared quantization logic.

Ratish1 (approved):
LGTM

sglang_omni/platforms/cuda.py
Makcum888e: moved this logic to model_worker

Makcum888e:
Benchmark results A3 2 cards (4 devices) - 2 for thinker, 1 for talker, 1 for others:
Qwen3-Omni server command:
```
sgl-omni serve --model-path /path/to/Qwen/Qwen3-Omni-30B-A3B-Instruct/ --thinker-gpus 2,3 --thinker-tp-size 2 --talker-gpu 1 --code2wav-gpu 0 --port 8764 --image-encoder-gpus 0 --talker-cuda-graph off
```
Qwen3-ASR server command (need to disable cuda graph for now):
```
sgl-omni serve --model-path /path/to/Qwen/Qwen3-ASR-1.7B/ --port 8765 --model-name Qwen/Qwen3-ASR-1.7B
```
generate command:
```
python -m benchmarks.eval.benchmark_omni_seedtts --meta zhaochenyang20/seed-tts-eval-arrow --port 8764 --output-dir results/qwen3_omni_tts_en --lang en --max-concurrency 8 --generate-only
```
transcribe command:
```
python -m benchmarks.eval.benchmark_omni_seedtts --meta zhaochenyang20/seed-tts-eval-arrow --port 8765 --output-dir results/qwen3_omni_tts_en --lang en --max-concurrency 128 --transcribe-only --asr-model-path /path/to/Qwen/Qwen3-ASR-1.7B/
```
generate `--max-concurrency 8`:
```
============================================================
                 TTS Speed Benchmark Result                 
============================================================
  Model:                         qwen3-omni
  Dataset:                       zhaochenyang20/seed-tts-eval-arrow
  Completed requests:            1088
  Failed requests:               0
------------------------------------------------------------
  Latency mean (s):              7.079
  Latency median (s):            6.855
  Latency p95 (s):               10.463
  Latency p99 (s):               12.875
  RTF mean:                      1.9171
  RTF median:                    1.9031
  RTF p95:                       2.0946
  RTF p99:                       2.251
  Audio duration mean (s):       3.72
  Audio throughput (s/s):        4.193
  Output throughput (tok/s):     16.1
  Output tokens/request-s:       2.0
  Output tokens (mean):          14
  Output tokens (total):         15506
  Prompt tokens (mean):          31
  Prompt tokens (total):         33889
  Throughput (req/s):            1.127
============================================================

============================================================
                  TTS WER Benchmark Result                  
============================================================
  TTS model:                     qwen3-omni
  Language:                      en
  Evaluated / Total:             1088/1088
  Skipped:                       0
------------------------------------------------------------
  WER (corpus, micro-avg):       0.0224 (2.24%)
------------------------------------------------------------
  WER per-sample mean:           0.0219 (2.19%)
  WER per-sample median:         0.0000
  WER per-sample std:            0.1068
  WER per-sample p95:            0.1111
  WER per-sample max:            2.8571 (285.71%)
  WER corpus (excl >50%):        0.0181 (1.81%)
  >50% WER samples:              2 (0.2%)
------------------------------------------------------------
  Latency mean (s):              7.078811764705883
  Latency p95 (s):               10.463155
  RTF mean:                      1.9171511604860578
  Throughput (req/s):            1.127
  Audio duration mean (s):       3.7205029411764703
============================================================


============================================================
                 ASR Speed Benchmark Result                 
============================================================
  ASR model:                     /home/weights/Qwen/Qwen3-ASR-1.7B/
  Evaluated / Total:             1088/1088
  ASR concurrency:               32
  Skipped:                       0
------------------------------------------------------------
  ASR latency mean (s):          2.068171922326602
  ASR latency median (s):        2.123109660024056
  ASR latency p95 (s):           2.6636261259933236
  ASR latency p99 (s):           2.9315928031684573
  ASR RTF mean:                  0.5932765885553406
  ASR RTF median:                0.5876068490241582
  ASR total time (s):            70.76445655996213
  ASR latency sum (s):           2250.171051491343
  ASR throughput (samples/s):    15.374950263033325
  Audio processed (s):           4047.9072000000024
============================================================
```
generate `--max-concurrency 128`:
```
============================================================
                 TTS Speed Benchmark Result                 
============================================================
  Model:                         qwen3-omni
  Dataset:                       zhaochenyang20/seed-tts-eval-arrow
  Completed requests:            1088
  Failed requests:               0
------------------------------------------------------------
  Latency mean (s):              44.053
  Latency median (s):            43.854
  Latency p95 (s):               54.079
  Latency p99 (s):               59.652
  RTF mean:                      12.9288
  RTF median:                    12.0078
  RTF p95:                       20.6837
  RTF p99:                       30.5686
  Audio duration mean (s):       3.727
  Audio throughput (s/s):        10.209
  Output throughput (tok/s):     39.0
  Output tokens/request-s:       0.3
  Output tokens (mean):          14
  Output tokens (total):         15500
  Prompt tokens (mean):          31
  Prompt tokens (total):         33889
  Throughput (req/s):            2.739
============================================================

============================================================
                  TTS WER Benchmark Result                  
============================================================
  TTS model:                     qwen3-omni
  Language:                      en
  Evaluated / Total:             1088/1088
  Skipped:                       0
------------------------------------------------------------
  WER (corpus, micro-avg):       0.1581 (15.81%)
------------------------------------------------------------
  WER per-sample mean:           0.1563 (15.63%)
  WER per-sample median:         0.0000
  WER per-sample std:            0.3678
  WER per-sample p95:            1.0000
  WER per-sample max:            3.3333 (333.33%)
  WER corpus (excl >50%):        0.0252 (2.52%)
  >50% WER samples:              144 (13.2%)
------------------------------------------------------------
  Latency mean (s):              44.053469301470585
  Latency p95 (s):               54.07926499999998
  RTF mean:                      12.92879441904357
  Throughput (req/s):            2.739
  Audio duration mean (s):       3.7269735294117647
============================================================


============================================================
                 ASR Speed Benchmark Result                 
============================================================
  ASR model:                     /home/weights/Qwen/Qwen3-ASR-1.7B/
  Evaluated / Total:             1088/1088
  ASR concurrency:               32
  Skipped:                       0
------------------------------------------------------------
  ASR latency mean (s):          2.1284455432448395
  ASR latency median (s):        2.1084511050139554
  ASR latency p95 (s):           2.95846289798792
  ASR latency p99 (s):           3.4191481200890825
  ASR RTF mean:                  0.614519622788703
  ASR RTF median:                0.5830751223766153
  ASR total time (s):            72.7999532300164
  ASR latency sum (s):           2315.748751050385
  ASR throughput (samples/s):    14.945064546434391
  Audio processed (s):           4054.9471999999996
============================================================
```

Issue #1466, opened 2026-08-11
[RFC] Unify configuration interfaces and resolution


MayDomine:
## Motivation

SGLang-Omni currently accepts configuration through model Python defaults, Pipeline YAML, dynamic dotted CLI options, dedicated typed CLI options, environment variables, and Router launcher YAML.

Multiple inputs are useful, but the same setting can follow different code paths for path resolution, type conversion, precedence, validation, and mutation. This creates three usability and maintenance problems:

1. Adding one setting can require duplicate YAML and CLI validation/mutation logic.
2. Developers do not have a single rule for deciding whether a setting belongs in the typed schema, CLI, `factory_args`, `runtime_overrides`, or a model-specific config.
3. Users do not have a clear way to discover where to change a setting or explain which source produced the final value.

## Proposal documents

- [Configuration Interface Refactoring Design Proposal (English)](https://github.com/sgl-project/sglang-omni/blob/sunao/config-interface-refactor-proposal/docs/design/configuration_interface_refactor_proposal_en.md)
- [配置接口重构设计 Proposal（中文）](https://github.com/sgl-project/sglang-omni/blob/sunao/config-interface-refactor-proposal/docs/design/configuration_interface_refactor_proposal.md)

Notion mirror: https://www.notion.so/maydomain/Configuration-Refactor-Proposal-for-sglang-omni-3b84e1dff05a808a8076ea49cf9a9a7e

## Proposed direction

Introduce one canonical typed configuration schema and one resolver:

- YAML remains a model-template overlay for persistent, reviewable Pipeline configuration.
- CLI keeps service-process options and uses `--set PATH=VALUE` for temporary Pipeline overrides.
- YAML, CLI, compatibility aliases, and Router worker settings all produce the same source-aware `ConfigPatch` representation.
- A single `ConfigResolver` owns type conversion, precedence, conflict detection, validation, and provenance.
- Runtime planners consume an immutable resolved config; derived runtime values remain outside the user configuration.
- Router launcher YAML replaces `worker_extra_args` shell fragments with structured worker config and `set` values.

Related historical issue: #107. That issue addressed earlier field-placement problems; this RFC covers the broader interface and resolution model.

## Tracking milestones

- [ ] Agree on ownership boundaries and the canonical path model
- [ ] Add `ConfigPatch`, source provenance, and a unified resolver without changing current resolved behavior
- [ ] Add `config schema`, `validate`, `resolve`, `explain`, and `migrate` commands
[...]

PR #1477, opened 2026-08-11
Bump SGLang to 0.5.17

SunskyXH:
## Motivation

Bump the pinned SGLang dependency from 0.5.16 to 0.5.17 and adapt the scheduler, sampling, kernel import and runtime context APIs that changed.

SGLang 0.5.17 also adds an overridable decode CUDA graph runner class (`srt/model_executor/model_runner.py`, invoked from `model_runner_components/cuda_graph_setup.py`), where 0.5.16 hardcoded `DecodeCudaGraphRunner`.

Dependency metadata delta in the SGLang wheel:

| requirement | 0.5.16 | 0.5.17 |
|---|---|---|
| `helion` | `==0.2.6` | `==1.4` |
| `flashinfer_python[cu13]` | `==0.6.14` | `==0.6.15.post1` |
| `sgl-deep-gemm` | `==0.1.4.post1` | `==0.1.5.post1` |
| `av` on Linux ARM | unpinned | `==16.1.0` |
| `xxhash` | absent | added, unpinned |

torch, transformers and sglang-kernel are unchanged. The only pin this repository changes directly is `sglang` in `pyproject.toml`.

## Modifications

This PR addresses eight SGLang 0.5.17 compatibility changes.

* Remove the obsolete `vocab_mask` and `apply_mask_func` arguments from `SamplingBatchInfo` construction. 0.5.17 folds them into `grammar_mask`.
* Import `flash_attn_varlen_func` from `sglang.kernels.ops.attention.flash_attention`. 0.5.17 retires `sglang.jit_kernel`. The import sits behind `except ImportError`, so on 0.5.17 the MOSS-TTS-Local vocoder would have dropped to SDPA without raising.
* Drop `server_args` from `SchedulerLogprobResultProcessor` construction.
* Pass `model_runner` to `SchedulerDPAttnAdapter`, now required.
* Pass the three throughput telemetry accessors to `SchedulerLoadInquirer`, now required.
* Initialize the four new telemetry state fields and stamp `ScheduleBatch.launch_ts` in both Omni execution paths. `OmniScheduler` calls upstream `process_batch_result`, which reads them.
* Add `dcp_size=1` to the scheduler test doubles required by the 0.5.17 token clamp.
* Write `pp_max_micro_batch_size` through `get_context().override()`.

The last item needs context. 0.5.17 completes the config split: `ServerArgs` is the read-only startup record, resolved config lives on namespace bags, and `get_context().override(...)` is the mutation entry point. `ServerArgs.override()` keeps its signature and still succeeds, but writes only that object, with no write through to the bags.

This repository uses `override_server_args` for these mutations, 23 call sites across 13 fields. The practical rule is narrower than "these overrides stop working":

* Writing a field through `ServerArgs.override` and later reading it back off `ServerArgs` still works, because the helper does set the attribute on that object.
* Writing a field and expecting a config bag reader to observe it does not work.

`pp_max_micro_batch_size` is the case that breaks. 0.5.17 moved `get_num_allocatable_reqs` from `get_server_args().pp_max_micro_batch_size` to `get_parallel().pp_max_micro_batch_size`, and nothing in Omni reads that value back off `ServerArgs`, so the scheduler saw `None` and died on `None - int` at the first batch. This PR routes it through `get_context().override`, matching upstream's own `Scheduler.__init__`.

Instrumenting the helper on a live Qwen3-TTS server on 0.5.17 shows the other sites on that path are not affected:

| site | published at call time | outcome |
|---|---|---|
| `scheduling/bootstrap.py:188` sets `disable_cuda_graph=True` | no | applies normally |
| `scheduling/bootstrap.py:202` restores `disable_cuda_graph=False` | yes | no effect on behavior; the caller carries the decision in the returned `want_cuda_graph`, and decode graph capture runs |
| `models/qwen3_tts/engine_builder.py:97` sets `enable_torch_compile=False` | yes | `sglang_model_runner.py` re-seeds `get_flags().capture.enable_torch_compile` from `ServerArgs`, so the value reaches the capture path |

The remaining call sites belong to model paths this run does not exercise and are not audited here.

Two known differences from upstream in the telemetry mirroring:

* The counters add per step host work. Omni does not call `load_inquirer.get_loads` and sets `load_snapshot_writer = None`, though upstream still exposes snapshots through `/v1/loads`.
* On the launch and resolve split path `launch_ts` is stamped at launch while the result is processed after a later resolve, so `total_prefill_busy_us` covers a different span than upstream's definition of elapsed time since launch.

## Related Issues

Related: #1466.

## Accuracy Test

ASR CI, `FunAudioLLM/Fun-ASR-Nano-2512-hf`, SeedTTS EN, 1088/1088 evaluated, 0 skipped, concurrency 32, two router workers on one H200, one run per side.

| metric | 0.5.16 | 0.5.17 |
|---|---:|---:|
| WER corpus, micro avg | 1.72% | 1.71% |
| WER per sample mean | 1.78% | 1.77% |
| WER per sample median | 0.0000 | 0.0000 |
| WER per sample std | 0.0457 | 0.0453 |
| WER per sample p95 | 10.00% | 10.00% |
| WER per sample max | 30.77% | 28.57% |
| samples over 50% WER | 0 | 0 |

Repeating the 0.5.16 run moves corpus WER by 0.01 percentage points, so the difference between the two versions is within the observed run to run variation.

`tests/test_model/test_asr_ci_seedtts.py` passes on both, 2 passed each.

TTS CI stage 1, `bosonai/higgs-tts-3-4b`, non-streaming, concurrency 16, two router workers on two H200, one run per side.

| metric | 0.5.16 | 0.5.17 |
|---|---:|---:|
| WER corpus, micro avg | 0.97% | 0.93% |
| WER per sample p95 | 0.0833 | 0.0833 |
| WER per sample max | 42.86% | 42.86% |

`tests/test_model/test_tts_ci.py --tts-stage tts-stage-1-nonstream` passes on 0.5.17 with 4 passed.

TTS CI stage 2, streaming, same model and topology. Both sides pass with 2 passed.

| metric | 0.5.16 | 0.5.17 |
|---|---:|---:|
| WER corpus, micro avg | 1.00% | 0.99% |

Qwen3-Omni CI, on four H200:

| test | 0.5.16 | 0.5.17 |
|---|---|---|
| `test_qwen3_omni_tts_ci.py` | passes | passes |
| `test_qwen3_omni_thinker_length.py` | passes | passes |
| `test_qwen3_omni_mmmu_talker_ci.py` | passes | passes |
| `test_qwen3_omni_mmsu_ci.py` | fails 3 speed checks: qps 41.7 under 48.04, tok/req/s 5.4 under 6.2, latency 0.382 over 0.3 | fails the same 3 checks: qps 41.1, tok/req/s 5.3, latency 0.389 |
| `test_qwen3_omni_mmmu_ci.py` | fails 3 speed checks: qps 1.512 under 1.62, tok/req/s 63.1 under 78.0, latency 9.208 over 8.4 | passes, three consecutive runs |
| `test_qwen3_omni_mmsu_talker_ci.py` | fails, MMSU audio accuracy 52.5 percent | fails, accuracy 57.5 percent |

`test_qwen3_omni_mmsu_talker_ci.py` scores accuracy over `MAX_SAMPLES = 40`, so one sample is 2.5 percentage points. Observed values across runs span 52.5 to above the 60 percent threshold. At that sample size the metric cannot separate the two versions and no conclusion is drawn from it here.

`test_asr_ci_multi_speaker.py` and `test_tts_serving_ci.py` fail on both sides on this host. Their failing checks are p95 latency thresholds calibrated on the project CI host, they differ from run to run on both versions, and the largest single overshoot in the set belongs to 0.5.16, not 0.5.17.

Coverage: 12 of the 17 GPU tests the CI workflows run. TTS CI stages 3 to 5 and the four Qwen3-Omni video jobs have not been run.

Environment for these runs. 0.5.16 was measured with flashinfer 0.6.14, helion 0.2.6 and sgl-deep-gemm 0.1.4.post1. 0.5.17 was measured with flashinfer 0.6.15.post1, helion 1.4.0 and sgl-deep-gemm 0.1.5.post1, which is the dependency set 0.5.17 requires. Each side therefore runs the stack its own pin implies.

Four apparent slowdowns were investigated and none reproduced. Each came from the first execution of a path on a given side, or from GPU memory still held by a previous run, and each returned to the baseline range on repeats: Qwen3-TTS throughput needed four runs to reach steady state, TTS stage 2 TTFC p95 fell from 0.5838 back into the 0.506 to 0.529 baseline range, `ws_stream` latency p95 in `test_tts_serving_ci.py` exceeded its threshold once at 13.82s and stayed under it on both repeats, and `test_qwen3_omni_mmmu_ci.py` measured qps 0.959 and latency 16.053s once on a cold inductor cache and then passed on three consecutive clean runs. Running `.github/scripts/delete_gpu_process.sh` between runs, as the CI workflows do, is necessary to get stable numbers.

One note on method for anyone repeating this. Shadowing the pinned sglang through `PYTHONPATH` works for most of these tests, but `test_tts_serving_ci.py` and `test_qwen3_omni_tts_ci.py` replace the worker environment with a fixed `PYTHONPATH`, which silently drops the shadow and runs the workers on the installed version. The numbers above for those two come from a real install of each version.

## Benchmark & Profiling

From the Accuracy Test runs above.

| ASR CI metric | 0.5.16 | 0.5.17 |
|---|---:|---:|
| RTF mean | 0.0498 | 0.0464 |
| RTF median | 0.0408 | 0.0386 |
| RTF p95 | 0.1183 | 0.1023 |
| throughput, samples/s | 139.68 | 149.53 |

| TTS CI stage 1 metric | 0.5.16 | 0.5.17 |
|---|---:|---:|
| latency mean, s | 0.8304 | 0.8430 |
| latency p95, s | 1.2507 | 1.2094 |
| RTF mean | 0.2022 | 0.2033 |
| throughput, req/s | 19.176 | 18.858 |

| TTS CI stage 2 metric | 0.5.16 | 0.5.17 |
|---|---:|---:|
| TTFC mean, s | 0.3282 | 0.3369 |
| TTFC p95, s | 0.5208 | 0.5838 |
| latency mean, s | 0.8473 | 0.8646 |
| latency p95, s | 1.2107 | 1.2304 |
| RTF mean | 0.2072 | 0.2092 |
| throughput, req/s | 18.761 | 18.396 |

Stage 2 TTFC p95 is the largest single gap measured so far at 12 percent. It comes from one run per side and has not been repeated.

Unit suite on 0.5.17 is 4107 passed, 1 failed. The failure, `test_mp_runner_startup_failure_includes_child_factory_traceback`, reproduces identically on 0.5.16: the test allows 10s for startup and a cold `import sglang_omni.pipeline.stage_workers` takes 18.6s on this host.

## Checklist

- [x] Add unit tests.

## CI


SunskyXH:
@FrankLeeeee — flagging this here since CODEOWNERS lists you for `/docker` and `.github`.

Omni CI fails at `setup - omni venv`:

```
flashinfer must come from the image, not /data/omni-ci/pr-1477/omni
Torch and FlashInfer must use the image installation for JIT cache reuse
```

`sglang==0.5.17` requires `flashinfer_python[cu13]==0.6.15.post1`, where 0.5.16 pinned `==0.6.14`. Installing the project therefore puts a venv-local flashinfer in place and `validate_omni_env_reusable.sh` rejects it. That version is a transitive requirement of sglang itself, so it happens regardless of what `pyproject.toml` says — this branch has failed the same way since its first CI run, well before the recent main merges.

Since the workflows run on a pinned digest and nothing in CI builds the image, I don't think there is a change inside this PR that unblocks it. It looks like it needs a CI image rebuilt against the 0.5.17 dependency set (flashinfer 0.6.15.post1, helion 1.4, sgl-deep-gemm 0.1.5.post1), plus a digest bump in the workflows.

I'm glad to take whichever part is useful — bumping the `flashinfer-python` pin and the `/root/.cache/flashinfer/<version>` COPY in `docker/Dockerfile`, or updating the digests once a new image is published. Which would you prefer?

For context, the adaptation itself is validated locally on H200: unit suite 4688 passed, and ASR CI plus TTS CI stages 1–2 match or beat the numbers in the PR description.

SunskyXH:
close this PR since we already bump to 0.5.18 in https://github.com/sgl-project/sglang-omni/pull/1719

PR #1468, opened 2026-08-11
[Docs] Document configuration flow and refactor proposal

MayDomine:
## Summary

- Document the current CLI, Pipeline YAML, Router YAML, and runtime configuration flow.
- Propose a canonical schema, source-aware `ConfigPatch`, and unified `ConfigResolver` with a phased compatibility plan.
- Provide the configuration interface refactor proposal in both English and Chinese.
- Add the configuration survey to the developer-reference navigation.

## Test plan

- [x] Run pre-commit checks on all changed documentation.
- [x] Verify that the English and Chinese proposals have matching heading and code-block structure.
- [ ] Run the full Sphinx documentation build; the local environment is currently missing `sphinx_tabs`.

## Tracking

Tracks #1466.

PR #1494, opened 2026-08-12
[Config Refactor] Add a canonical config core and make it authoritative

MayDomine:
## Motivation

Adding one configuration knob to sglang-omni meant touching six places.
`mem_fraction_static` was mentioned in 33 files (55 times in `cli/serve.py`
alone); `torch_compile` in 39 files. The same job — *find this stage's field
and write to it* — was implemented four separate times with subtly different
semantics (the dotted-path dict walker, 16 `apply_*_cli_overrides` helpers,
the five-source factory-kwarg merge, and the `stage_overrides` deep merge).
None of them could answer "where did this value come from", none validated a
path before the stage process started, and a value could be written from up to
three spellings whose conflicts were settled by helper execution order.

This PR replaces all of it with one write path and one path language, with
**no V1 compatibility**: the old spellings are removed, not translated.
See #1466 for the direction and #1468 for the proposal.

## The surface

Stage settings are grouped by the module that consumes them:

```yaml
config_cls: MossTTSPipelineConfig
model_path: OpenMOSS-Team/MOSS-TTS

stages:                      # name-keyed; merge semantics per field
  tts_engine:
    tp_size: 2               # stage top level -> placement/process planning
    engine:                  # -> SGLang ServerArgs (engine stages only)
      mem_fraction_static: 0.7
    scheduler:               # -> the stage's executor
      max_concurrency: 8
    model:                   # -> the stage factory
      dtype: bfloat16

shared:                      # one value into several stages
  - select: {stages: [preprocessing, latent_engine]}
    model: {num_steps: 4}
```

The command line is the same path language with the `stages.` prefix implied:

```bash
sgl-omni serve --config omni.yaml \
    --tts_engine.engine.mem_fraction_static 0.6 \
    --vocoder.process vocoder
```

The groups declare their commonly tuned fields for eager validation and pass
every other key through untouched — the vocabularies belong to the consumers
(SGLang ServerArgs, the factory signature), so the entry side never
format-checks unknown keys. **What used to be a `factory_args` entry is now
written under `model.*`** and reaches the factory under its own name; a key
the factory does not accept is an error at stage construction, never a silent
drop.

Stage topology — which stages exist, how they route, where requests enter —
lives in the model's `config.py` only. Config files and flags override
settings on those stages; they cannot add, remove or rename stages or set
`entry_stage`.

## The mechanism

Every source reduces to `ConfigPatch(path, value, source, layer, specificity)`,
compiled against the typed schema at construction (illegal paths fail with
`did you mean:`; CLI text is coerced by the declared type). The resolver is the
only code that writes into a configuration: dump, apply in precedence order,
rebuild, revalidate.

```
Layer:        MODEL_DEFAULT < USER_FILE < CLI
Specificity:  BROADCAST (--mem-fraction-static fan-out)
            < ROLE      (shared: selector expansion)
            < EXPLICIT  (a written-out path)
```

`ProvenanceMap` records each path's full history. `sgl-omni config resolve`
previews the configuration a launch with the same arguments would build
(`--show config|diff|provenance`) and `config explain PATH` names the source
that set a value and what it superseded — both assemble the identical patch
set `serve` does, including `--model-path`-beats-file and the broadcast flag.

## One rule for duplicates

Every spelling canonicalizes to one path at patch creation, so a single rule
covers what used to resolve four different ways:

* the same flag repeated with different values → refused (`is set twice at
  the same precedence … Pick one.`); repeated with the same value → tolerated.
  `parse_extra_args` returns ordered pairs, not a dict, so repeats survive to
  the resolver instead of collapsing last-wins;
* a higher layer or specificity wins silently (CLI over file, explicit over
  `shared:` over broadcast) — that is declared precedence, not a conflict;
* alias pairs cannot disagree because no aliases remain: `parallelism.tp`,
  the typed per-role flags, positional indices and `stage_overrides`/`set:`/
  `runtime_overrides` are deleted, and writing them fails at path compilation
  with directions to the current spelling;
* the one container case (`env`, a free mapping) composes: a dict written at
  the container deep-merges under any leaf patch instead of the deeper write
  silently winning.

## Intentional differences from V1

There is no compatibility layer and no deprecation window; V1 spellings fail
fast with guidance. Differences that change meaning rather than spelling:

* **positional indices are rejected everywhere** — stages are addressed by
  name (`stages.1.gpu` → "addressed by name"), and scalar lists such as
  `stream_to` are leaves replaced as a whole (`stream_to.0` → "replaced as a
  whole"). V1 accepted digit indices into both; index-based writes silently
  retargeted when the stage list changed, which is the failure mode the
  by-name surface exists to remove;
* **`--mem-fraction-static` precedence flips**: the broadcast flag now loses
  to an explicit dotted path for the same field. V1's typed flag beat the
  dotted spelling because its helper ran later. Command lines that pass both
  spellings for one stage get the dotted value now — with `config explain`
  showing the flag on the superseded line rather than resolving it by
  accident of ordering;
* **no other typed helpers remain**: the other 15 `apply_*` flags are deleted
  in this PR, not converted, so there is no half-converted precedence to
  explain. `--mem-fraction-static` survives as the only broadcast flag
  because one value fanning out to every engine stage is not expressible as
  a single dotted path.

## Numbers

109 files, +5,923 / −6,103. `cli/serve.py` 1512 → 425 lines;
`config/process_overrides.py` (289 lines) deleted; the V1 merge chain
(dict walker, `_convert_scalar` guessing, alias sync, five-source kwarg merge)
deleted. Adding a stage knob now means one schema field (or nothing, for a
pass-through key) — the path compiler, both surfaces, validation, conflict
detection and `config explain` pick it up from the declaration.


sglang_omni/config/resolver.py
```diff
@@ -0,0 +1,231 @@
...
+    for stage in stages:
+        if not isinstance(stage, dict) or "name" not in stage:
+            continue
+        name = stage["name"]
+        for left, right in _ALIAS_PAIRS:
+            left_path = left.replace("*", name)
+            right_path = right.replace("*", name)
+            left_written = left_path in written
+            right_written = right_path in written
+            if left_written == right_written:
```
JiaxinD: If a file sets `tp_size` and the CLI sets `parallelism.tp`, both sides are in `written`, the sync is skipped, and StageConfig validation rejects the pair. V1 let the CLI win here: repro `set: {stages.X.tp_size: 2}` plus `--stages.X.parallelism.tp 4`.

sglang_omni/config/path.py
```diff
@@ -0,0 +1,649 @@
...
+        if value.lower() == _NONE_SENTINEL:
+            return None
+        if annotation is Any or annotation is None:
+            return coerce_scalar_text(value)
+        try:
+            return TypeAdapter(annotation).validate_python(value)
+        except Exception:
+            pass
+        try:
+            return TypeAdapter(annotation).validate_json(value)
```
JiaxinD, line 319: A CLI `--set stages.X.runtime='{"max_seq_len": 100}'` returns a model instance from this line, so `_apply` skips its deep merge branch and replaces the whole container, resetting siblings like `mem_fraction_static` to defaults. The same value in a YAML `set:` block arrives as a dict and merges.

sglang_omni/cli/config.py
```diff
@@ -62,12 +76,228 @@ def export(
...
+    set_values: list[str] | None = None,
+) -> Resolution:
+    """Build the configuration ``sgl-omni serve`` would build from this input.
+
+    Not a re-implementation of the launch path: ``ConfigManager.merge_config``
+    resolves the very same patch set through the very same resolver. The only
+    difference is that a launch throws the provenance away and this keeps it,
+    which is what lets these commands answer *which source set this value*.
+    """
+    if config_file is None and model_path is None:
```
JiaxinD, line 129: When both `--config` and `--model-path` are given, the model path is silently ignored, while `serve` overwrites the file's model_path. Same arguments, different pipeline than launch.

sglang_omni/config/provenance.py
```diff
@@ -0,0 +1,124 @@
...
+    def winner(self, path: str) -> ProvenanceEntry | None:
+        for entry in reversed(self.entries.get(path, [])):
+            if entry.winning:
+                return entry
+        return None
+
+    def touched(self, path: str) -> bool:
+        return path in self.entries
+
+    def explain(self, path: str) -> str:
```
JiaxinD, line 124: This prints the winning patch's raw input, not the resolved value: `--set config_cls=Foo` resolves to `PipelineConfig` but explain reports `Foo`, and writing `tp_size` leaves the mirrored `parallelism.tp` reported as untouched.

sglang_omni/config/manager.py
```diff
@@ -69,39 +73,35 @@ def parse_extra_args(self, args: list[str]) -> dict[str, Any]:
...
+        into canonical patches and applied by
+        :class:`~sglang_omni.config.resolver.ConfigResolver`, which is the only
+        code that writes into a configuration. Legacy spellings -- positional
+        stage indices, paths the schema would rather people stopped using --
+        are accepted and reported, never silently dropped.
+
+        ``extra_patches`` carries patches a caller has already translated --
+        the typed CLI flags in ``sgl-omni serve``, for instance. Everything is
+        resolved together, in one patch set, so that writing the same path
+        two ways is refused (or settled by declared specificity) rather than
```
JiaxinD: Only cross spelling duplicates get refused: the same dotted flag repeated collapses last wins in `parse_extra_args` before patches exist, so `--stages.thinker.tp_size 2 --stages.thinker.tp_size 4` resolves silently to 4.

sglang_omni/config/patch.py
```diff
@@ -0,0 +1,298 @@
...
+        out: dict[str, ConfigPatch] = {}
+        for patch in self.ordered():
+            out[patch.key] = patch
+        return out
+
+    # ------------------------------------------------------------------
+    # checks
+    # ------------------------------------------------------------------
+
+    def conflicts(self) -> list[tuple[ConfigPatch, ConfigPatch]]:
```
JiaxinD, line 231: Keys on the exact path string, so a container patch `stages.X.runtime: {...}` and a leaf `stages.X.runtime.max_seq_len` at equal rank are never flagged and depth order settles it silently. That defeats the same path refusal `sources_from_config_file` promises for `set:` vs `stage_overrides`.

sglang_omni/cli/serve.py
```diff
@@ -1383,17 +1411,30 @@ def serve(
     # we use ctx to capture the arguments that are used to modify the configuration on the fly
     # we do expect the extra arguments to be pairs of names and values
     extra_args = config_manager.parse_extra_args(ctx.args)
```
JiaxinD: This raises plain ValueError for a trailing dotted key, above the try, and the catch at the merge below is `ConfigPathError` only, so `serve` tracebacks where `config resolve` prints a BadParameter. The alias conflict ValidationError from resolution escapes the same way.

sglang_omni/config/path.py
```diff
@@ -0,0 +1,649 @@
...
+        )
+
+    if core is Any:
+        return Segment(
+            raw=part, kind=SegmentKind.FREEFORM, annotation=Any, container=Any
+        )
+
+    if _is_plain_sequence(core):
+        raise ConfigPathError(
+            f"{_join_prefix(prefix)} is a list value and is replaced as a whole; "
```
JiaxinD, line 446: V1's `_resolve_list_index` accepted digit indices into plain lists, e.g. `stages.X.stream_to.0=y`; this now hard errors and the parity sweep never generates digit probes (the fixture's lists are empty), so it is an undocumented V2_REJECTS divergence.

sglang_omni/cli/serve.py
```diff
@@ -289,24 +282,28 @@ def _validate_tts_batch_max_items(value: int) -> int:
...
+    Successor to the post-merge ``apply_mem_fraction_cli_overrides`` mutation
+    helper. As patches the flags resolve together with dotted overrides and
+    ``--set``, which buys two things the helper could not offer:
+
+    * precedence is declared, not positional. ``--mem-fraction-static``
+      broadcasts at ``Specificity.BROADCAST``, the per-role flags land at
+      ``Specificity.ROLE``, and an explicit dotted or ``--set`` path outranks
+      both -- instead of the typed flag silently overwriting whatever the
+      merge produced because the helper happened to run afterwards;
+    * ``sgl-omni config explain`` can name the flag that set the value.
```
JiaxinD: Not true of the shipped command: `config resolve --mem-fraction-static 0.5` exits with `has no field 'mem_fraction_static'`; only `serve` builds these patches. Worth noting in the body that this also flips precedence: the typed flag used to beat a dotted override.

JiaxinD:
Beyond the inline comments, three things worth settling before fixing spot by spot:

* One rule for duplicates: the same field written twice resolves differently by spelling (literal repeat silently last-wins, index vs name errors, container vs leaf silently deeper-wins, `tp_size` vs `parallelism.tp` keeps both then fails validation). Canonicalizing every spelling to one path at patch creation would let same-rank refusal and layer precedence cover all cases uniformly.
* Digit indices into plain lists (`stages.X.stream_to.0`) were accepted by V1 and now hard error, with no `INTENTIONAL_DIFFERENCES` entry and no deprecation window. Intentional or missed?
* The mem-fraction precedence flip silently changes existing command lines, worth stating in the PR body. Also, what is the plan for the remaining typed helpers? Half-converted precedence on main is hard to explain.

sglang_omni/config/path.py
stephenkgli: Maybe we should only support setting "leaf" config e.g. stages.X.runtime.max_seq_len instead of stages.X.runtime ?

sglang_omni/cli/config.py
```diff
@@ -62,12 +76,228 @@ def export(
...
+            f"deprecated: {patch.key} <- {patch.source.describe()}: {patch.deprecated}",
+            err=True,
+            fg=typer.colors.YELLOW,
+        )
+
+
+@config_app.command(
+    context_settings={"allow_extra_args": True, "ignore_unknown_options": True}
+)
+def resolve(
```
stephenkgli, line 169: `config resolve` and `config explain `do not share the configuration assembly path used by `serve`: for example, they drop `--model-path `when `--config`is present and skip serve-specific typed-option patches. As a result, they can report a configuration different from the one that will actually be launched, making them unreliable for previewing or diagnosing server configuration.

sglang_omni/config/resolver.py
MayDomine: Fixed by removing the mirror entirely. `parallelism` no longer exists in the schema — `tp_size` is the only spelling (`--thinker.tp_size 2`), so there is no alias pair for two sources to disagree on and no sync step to skip. Writing `parallelism.tp` now fails at path compilation with a did-you-mean pointing at `tp_size`.

sglang_omni/config/path.py
MayDomine: Resolved by removing the surface and unifying the semantics. `--set` is gone; the dotted CLI and the YAML `stages:` mapping are the only write surfaces, and both are flattened to per-leaf patches before resolution (`_flatten` in `sources.py`), so the "same value merges from YAML but replaces from the CLI" split cannot occur. For the one true container case left (`env`, a free mapping), `_apply` deep-merges a dict written at a container path with any stronger-layer leaf patches instead of replacing wholesale.

sglang_omni/cli/config.py
MayDomine: Fixed. `config resolve`/`explain` now route `--model-path` through the same `patches_from_model_path_flag` helper `serve` uses: given both `--config` and `--model-path`, the flag becomes a CLI-layer patch that outranks the file's `model_path`, and `config explain model_path` names the flag as the winner with the file's value on the superseded line. Covered by `TestModelPathWithConfig` in `tests/unit_test/config/test_cli_config.py`.

sglang_omni/config/provenance.py
MayDomine: Fixed on both counts. The resolver records the value the rebuilt config actually holds per touched path (`record_resolved`), and `explain`'s headline reports that resolved value, noting a rewrite when it differs from the winning patch's input. The two examples no longer exist as inputs: `config_cls` is INTERNAL (not writable from any user-facing source), and the `parallelism.tp` mirror is deleted, so there is no shadow path to report as untouched.

sglang_omni/config/manager.py
MayDomine: Fixed. `parse_extra_args` now returns ordered `(key, value)` pairs instead of a dict, so a repeated flag survives parsing and the resolver rules on it: `--thinker.tp_size 2 --thinker.tp_size 4` is refused as `is set twice at the same precedence … Pick one.`, while repeating the same value is tolerated. Covered by `TestManager` in `tests/unit_test/config/test_config_set.py`.

sglang_omni/config/patch.py
MayDomine: The overlap this defeated is gone at the source: `set:` and `stage_overrides` are both removed, and the YAML `stages:` mapping plus `shared:` are flattened to per-leaf patches before resolution, so two same-rank writes to one leaf key on the exact same path string and are flagged. The remaining container/leaf pairing (an `env` dict plus a nested `env.KEY`) is deliberate compose-not-conflict: `_apply` deep-merges the container under the leaf rather than letting depth order clobber it.

sglang_omni/cli/serve.py
MayDomine: Fixed. `serve` now wraps `parse_extra_args` in its own try and converts the trailing-key `ValueError` to `typer.BadParameter`, and the merge-time catch widened to `(ConfigPathError, ValueError)` so resolution-time validation errors surface the same way. `config resolve`/`explain` share the identical catch, so both commands fail readably for the same inputs; covered by `TestServeErrors` in `tests/unit_test/config/test_cli_config.py`.

sglang_omni/config/path.py
MayDomine: Now documented behavior rather than an accidental divergence: positional indices are rejected everywhere by design — stages are addressed by name, and scalar lists (`stream_to`) are leaves replaced as a whole, with the refusal saying so (`is addressed by name, not by index` / `replaced as a whole`). The parity sweep and its frozen V1 oracle were removed together with the compatibility surface, so there is no undocumented V2_REJECTS bucket left; the rejection is pinned directly in `tests/unit_test/config/test_config_path.py`.

sglang_omni/cli/serve.py
MayDomine: Fixed. `config resolve`/`explain` now accept `--mem-fraction-static` and build its fan-out through the same `patches_from_broadcast_flags` helper `serve` uses, so the shipped command previews exactly what launches. On precedence: the flag fans out at BROADCAST specificity below an explicit dotted path by design — the specific spelling wins over the convenience fan-out, and `config explain` shows the flag on the superseded line. The remaining typed per-stage flags were deleted outright rather than re-ranked.

sglang_omni/cli/config.py
MayDomine: Fixed. `_resolve_sources` now assembles the identical patch set a launch builds: `--model-path` alongside `--config` becomes the same CLI-layer patch (flag wins over the file), and the one remaining serve typed option (`--mem-fraction-static`) is parsed and fanned out via the same serve helper. Every other typed option was deleted from `serve` itself, so there is no serve-only patch source left to diverge on; the tests assert `resolve` output equals `ConfigManager.merge_config` — the entry point `serve` calls.

MayDomine:
All three settled by the rework that landed since this review, and the PR body now states each explicitly (see "One rule for duplicates" and "Intentional differences from V1"):

1. **One rule for duplicates** — every spelling now canonicalizes to one path at patch creation, and there are no alias pairs left to disagree: `parallelism.tp`, the typed per-role flags, positional indices and `stage_overrides`/`set:` are deleted outright. A literal repeat is refused at the same precedence (`parse_extra_args` returns ordered pairs, so repeats survive to the resolver instead of collapsing last-wins); container-vs-leaf composes by deep merge for the one real container (`env`); everything else is settled by the declared layer/specificity ranks.

2. **Digit indices** — intentional, and now stated in the body rather than living as an unlabeled divergence. Stages are addressed by name and scalar lists are leaves replaced as a whole; both rejections carry a message saying so. There is no deprecation window because the PR ships zero V1 compatibility across the board — the parity sweep and its oracle were removed together with the surface they froze.

3. **Precedence flip and the remaining helpers** — the flip is documented in the body ("Intentional differences from V1"), and the half-converted state no longer exists: the other 15 typed helpers were deleted in this PR, not converted. `--mem-fraction-static` is the only flag that survives, as a BROADCAST-specificity fan-out that an explicit dotted path outranks — `config explain` shows the flag on the superseded line, so the outcome is inspectable rather than an accident of helper ordering. `config resolve`/`explain` also accept the flag now, so the shipped preview command matches a launch argument-for-argument.

JiaxinD (commented):
Re-reviewed at d532d63a. Six of the nine earlier items are fully fixed with regression tests, and the patch/resolver core plus the error engineering are in good shape; the factory merge kept all typed fields. What blocks this now is completion, not design:

* The repo has not finished its own no-V1 migration: qwen3_omni/placement.py (not in the diff) still reads the removed runtime group, so --colocate and the four migrated colocated yamls fail at placement; roughly 18 unit test files (~110 sites) still use factory_args or the old factory path string.
* Some legacy write paths survive outside the canonical resolver: the --model_path underscore spelling, the examples/launchers helpers with their own mem-fraction flags, and the MOSS_REF_AUDIO_CACHE env override that resolve/explain cannot see.
* Namespace reservation: #1175's top level processes: and #1452's KV byte budget are both hard-rejected by the current extra=forbid schema; worth reserving those two slots here so neither PR needs a schema fork.
* Nits: path.py repeats "engine.*, engine.*" in the runtime_overrides hint; a few docstrings still show model./scheduler. or removed flags; config export prints all-null factory/engine blocks.

sglang_omni/models/ming_omni/config.py
```diff
@@ -64,7 +64,7 @@ def _audio_encoder_stage(*, gpu: int, process: str) -> StageConfig:
     return StageConfig(
         name=AUDIO_STAGE,
         process=process,
-        factory=f"{_PKG}.stages.create_audio_encoder_executor",
+        factory_path=f"{_PKG}.stages.create_audio_encoder_executor",
         factory_args={"device": "cuda", "dtype": None},
```
JiaxinD: factory_args was removed but this file still passes it in six stage factories, so every MingOmni pipeline config raises ValidationError at construction. The thinker stage is also not classified as an engine stage, so engine.* writes to Ming are rejected.

sglang_omni/cli/serve.py
```diff
@@ -1465,97 +371,41 @@ def serve(
...
+        )
+    except (ConfigPathError, ValueError) as exc:
+        # Unknown path, unwritable path, two sources disagreeing at one
+        # precedence, or a value the schema refuses. Every one of these
+        # carries a message written to be read; a traceback would bury it
+        # under the merge internals.
+        raise typer.BadParameter(str(exc)) from exc
+    if colocate:
+        _validate_colocate_config(merged_config)
+    merged_config = apply_tensor_parallel_engine_overrides(merged_config)
```
JiaxinD, line 417: This runs after merge_config and writes engine.disable_custom_all_reduce without provenance, overriding an explicit user value; config resolve never runs it, so the preview diverges from launch for TP stages.

sglang_omni/config/sources.py
```diff
@@ -0,0 +1,401 @@
...
+}
+
+_STAGES_LIST_GUIDANCE = (
+    "stages must be a mapping keyed by stage name; the list form is no "
+    "longer accepted. Move each entry under its name and drop the name: "
+    "key:\n\n    stages:\n      thinker:\n        tp_size: 2"
+)
+
+
+def patches_from_dotted_cli(
```
JiaxinD, line 113: Dotted CLI JSON container values are coerced into typed group instances and applied as whole replacements, resetting sibling fields set by the file (measured: a factory JSON wiped max_concurrency to None). Either reject non-leaf CLI paths or flatten them like YAML.

sglang_omni/config/sources.py
```diff
@@ -0,0 +1,401 @@
...
+        prefix = f"stages.{stage_name}"
+        for path, value in _flatten(prefix, dict(body), config_cls):
+            patches.add(
+                ConfigPatch.create(path, value, source, root=config_cls, coerce=False)
+            )
+
+    return patches
+
+
+def sources_from_config_file(
```
JiaxinD, line 359: Top-level file values (model_path, name, placement) get no YAML provenance here, so explain attributes them to the model default.

sglang_omni/config/runtime.py
```diff
@@ -10,45 +28,107 @@
...
-    global_cfg: PipelineConfig,
+def typed_stage_kwarg_path(name: str) -> str:
+    """The stage-config path that produced a typed factory kwarg."""
+
+    if name == "server_args_overrides":
+        return "engine"
+    return f"factory.{name}"
+
+
+def apply_typed_stage_kwargs(
```
JiaxinD, line 97: A factory with **kwargs defeats this check: the dots vocoder declares **_: Any and never reads it, so both a typo and the correctly spelled factory.max_concurrency are silently swallowed there.

sglang_omni/models/ming_omni/config.py
MayDomine: Fixed in 808385eb. This was the rebase silently reintroducing the old spelling: upstream commits landed the stage factories back with factory_args dicts without a conflict, and imports stayed green because the stage list is a lazy default_factory. All six stages now build FactoryArgs, and the thinker is declared an engine stage (stage_config_types maps it to EngineStageConfig), so thinker.engine.* writes are accepted. All three MingOmni pipeline classes construct; the sweep that guards this now instantiates every registered pipeline class and variant rather than just importing the modules.

sglang_omni/cli/serve.py
MayDomine: Fixed in 250a69af. The derivation only fills gaps now: each derived key is checked against the stage's engine block first, and any key the merged configuration already sets -- model default, config file, or dotted flag -- is left exactly as written, so an explicit disable_custom_all_reduce survives. config resolve/explain run the same apply_tensor_parallel_engine_overrides after resolution, so the preview matches the launch for TP stages. Both directions are pinned by TestTensorParallelDerivation in tests/unit_test/config/test_cli_config.py. One remaining gap, flagged rather than hidden: explain shows the derived value but still attributes the path to the model default -- a dedicated provenance label for derived writes would extend the provenance model, so it is left for a follow-up if wanted.

sglang_omni/config/sources.py
MayDomine: Fixed in 250a69af, taking the reject option: a dotted flag addressing a settings group is refused at parse time with leaf guidance (--thinker.factory '{...}' -> "addresses a settings group, not a single value; write one field below it"), regardless of what the value looks like -- same shape as the existing whole-stage refusal. The CLI writes one leaf at a time; blocks belong to the YAML stages: mapping. Free-form keys (engine passthrough, env.VAR) are exempt from the container check because their open type makes the key itself the leaf. Pinned in TestDottedCli: a JSON mapping and a scalar at a group path are both refused, and the passthrough tests still pass.

sglang_omni/config/sources.py
MayDomine: Fixed in 250a69af. sources_from_config_file now emits a user-file patch for every top-level key besides config_cls (nested values such as placement are flattened per leaf), alongside the stages:/shared: patches, so config explain names the file as the winner for model_path and friends instead of attributing them to the model default. Covered by TestExplain::test_a_top_level_file_value_names_the_file.

sglang_omni/config/runtime.py
MayDomine: Fixed in 250a69af. The catch-all is removed from every stage factory that declared one and never read it -- the dots vocoder plus the zonos2 and minimax-music3 factories -- after checking that each config's stage_factory_kwargs hook only passes names those signatures declare, so nothing legitimate relied on it. With it gone, apply_typed_stage_kwargs refuses both a typo and a correctly spelled factory.* key the factory does not accept. The same commit fixes a latent crash next door: the dots stream_slots guard read the vocoder's factory group through the group's pre-rename attribute and would have raised at launch. Regression tests in tests/unit_test/dots_tts/test_vocoder.py cover the refusal and the declared-kwarg path.

sglang_omni/config/schema.py
```diff
@@ -145,27 +160,45 @@ class StageConfig(BaseModel):
...
     route_fn: str | None = None
 
-    # --- GPU / parallelism ---
+    # --- GPU / parallelism / process placement (parent-process consumers) ---
     gpu: int | list[int] | None = None
-    tp_size: int = 1
-    parallelism: ParallelismConfig = Field(default_factory=ParallelismConfig)
+    tp_size: int = Field(default=1, ge=1)
     process: str | None = None
+    gpu_memory_fraction: float | None = Field(
```
stephenkgli, line 265: The colocated Qwen placement policy still reads the removed `stage.runtime` fields. `StagePlacementPlanner.build()` invokes it for `Qwen3OmniSpeechColocatedPipelineConfig`, so colocated configs fail before worker startup with `AttributeError`.

sglang_omni/config/schema.py
```diff
@@ -145,27 +160,45 @@ class StageConfig(BaseModel):
...
+            "Per-stage-rank budget as a fraction of total physical GPU memory. "
+            "After TP expansion, each rank contributes this budget to its "
+            "assigned GPU; stages sharing an OS process contribute jointly to "
+            "that process's budget."
+        ),
+    )
+
+    # --- Consumer groups ---
+    engine: EngineArgs | None = None
+    factory: FactoryArgs = Field(default_factory=FactoryArgs)
```
stephenkgli, line 279: The V1 field removal is not reflected in the test suite: at least 16 unit-test modules still access `.factory_args` or `.runtime`. For example, `FunASRPipelineConfig(model_path="dummy").stages[0].factory_args` raises `AttributeError`, so those tests cannot pass.

sglang_omni/cli/serve.py
```diff
@@ -580,449 +225,18 @@ def _apply_tensor_parallel_server_args_overrides(
...
-    if not matching_stages:
-        raise typer.BadParameter(
-            "--decode-mode/--async-lookahead-min-batch-size currently supports "
-            f"only {_ASYNC_DECODE_SUPPORTED_MODELS}; no stage in this pipeline "
-            "uses a supported factory"
-        )
-    _apply_factory_args_updates(pipeline_config, matching_stages, updates)
-    return pipeline_config
-
+        already_set = stage.engine.overrides() if stage.engine is not None else {}
```
stephenkgli, line 222: An explicit `none` is overwritten by TP derivation. `EngineArgs.overrides()` drops `None`, so `--thinker.engine.disable_custom_all_reduce none` is treated as unset; `config explain` reports `None` as the CLI winner while launch uses `True`.

sglang_omni/config/sources.py
```diff
@@ -0,0 +1,427 @@
...
+        data = yaml.safe_load(f)
+    if not isinstance(data, dict):
+        raise ValueError(f"Config file {file_path!r} must contain a mapping")
+
+    data = dict(data)
+    for block, guidance in _REMOVED_TOP_LEVEL_BLOCKS.items():
+        if block in data:
+            raise ValueError(f"Config file {file_path!r}: {guidance}")
+
+    config_cls = PIPELINE_CONFIG_REGISTRY.get_config_cls_by_name(data["config_cls"])
```
stephenkgli, line 396: Missing or malformed config files bypass the CLI error handling. A missing `config_cls` raises `KeyError` and malformed YAML raises `ParserError`; both escape as tracebacks because the callers only catch `ConfigPathError` and `ValueError`.

sglang_omni/config/sources.py
```diff
@@ -0,0 +1,427 @@
...
+    stages_block = data.pop("stages", None)
+    shared_block = data.pop("shared", None)
+
+    config = config_cls(**data)
```
stephenkgli: Top-level YAML values are baked into the baseline and then applied again as patches. `--show diff` therefore omits file-set `model_path`/`name`, while provenance reports the same file value as both the model default and YAML winner.

sglang_omni/config/schema.py
MayDomine: Fixed in 5cae1c7d. The policy reads the grouped surface now -- stage.gpu_memory_fraction for the per-stage budget and stage.engine.mem_fraction_static for the engine override -- and its error messages name those spellings. The root cause of it surviving the earlier sweeps is worth naming: the policy class is loaded lazily from the placement_policy dotted string at plan time, so instantiating every pipeline class never touches it; the residue sweep now greps the retired spellings across the whole tree instead of relying on construction.

sglang_omni/config/schema.py
MayDomine: Fixed in 002db752. Every remaining module is migrated -- the sweep covered five spellings, not just the two named: .factory_args reads, the runtime groups (reads and writes), the string factory field, runtime_overrides= constructor kwargs, and .factory.endswith on the old string field. Assertions moved to where each value lives now (typed factory fields, engine.overrides(), gpu_memory_fraction, stage_factory_kwargs(), dotted merge_config writes). Three tests changed meaning deliberately: the device contract narrows to "qualified stages leave factory.device unset", the MOSS-TTS-Local None-concurrency rejection becomes clearing-falls-back-to-default (the old guard is gone with runtime_overrides), and the batch-invariance CI config writes the stages: mapping. A tree-wide grep for all five spellings is clean except the migration-guidance tables themselves.

sglang_omni/cli/serve.py
MayDomine: Fixed in 5cae1c7d, on the explain side. The behavior itself is coherent -- none clears the key back to unset, overrides() drops it, and an unset key is exactly what the derivation exists to fill -- but explain recorded its resolved values before the derivation ran, so it reported None while launch used True. config resolve/explain now re-record every patched path against the derived config: the headline reports the launched value, and the winner's line keeps the user's input with the rewrite noted (None <- cli dotted [winner] (resolved to True)). Treating an explicit none as a pin instead was considered and rejected: after the dump/rebuild roundtrip a cleared key is indistinguishable from one never set, so a pinned-None contract cannot survive a re-merge.

sglang_omni/config/sources.py
MayDomine: Fixed in 5cae1c7d. sources_from_config_file converts both cases into named refusals -- malformed YAML becomes "Config file '...' is not valid YAML: <parser detail>" and a missing config_cls becomes "Config file '...' must name its pipeline class in config_cls" -- as ValueError, which every CLI caller already converts to typer.BadParameter. Both surface as the readable error box, no traceback; verified through the CliRunner for resolve, and serve shares the same catch.

sglang_omni/config/sources.py
MayDomine: Fixed in 5cae1c7d. The baseline is no longer built from the full file: it is the class's own defaults plus only the keys construction requires (model_path, typically -- discovered by retrying on pydantic's missing-field errors so subclasses with extra required fields keep working), and every other top-level value is applied exactly once, as a patch. --show diff now lists a file-set name/placement change, and explain's model-default line shows the class default rather than echoing the file's value. One boundary remains by construction: model_path itself is a required field with no before-the-file value, so it still seeds the baseline and does not appear in diff -- there is no default to diff against.

sglang_omni/config/sources.py
```diff
@@ -0,0 +1,453 @@
...
+        if stage_name not in known:
+            raise ValueError(
+                f"stages.{stage_name}: no stage named {stage_name!r} in "
+                f"{config_cls.__name__}; stage topology lives in the model's "
+                f"config class. This pipeline has: {', '.join(known)}"
+            )
+        prefix = f"stages.{stage_name}"
+        for path, value in _flatten(prefix, dict(body), config_cls):
+            patches.add(
+                ConfigPatch.create(path, value, source, root=config_cls, coerce=False)
```
JiaxinD: Stage YAML goes through coerce=False, so YAML tp_size: 32.0 lands via pydantic lax conversion while the same dotted CLI value is rejected; the two entry surfaces should share the lossless contract.

sglang_omni/config/sources.py
```diff
@@ -0,0 +1,453 @@
...
+    which is what launching needs and the opposite of what explaining needs:
+    by the time the caller holds the config, the stage entry that set a value
+    has already been absorbed into it. Keeping the two apart is what lets
+    ``sgl-omni config explain`` name the file as a source.
+    """
+    # Local import: the registry pulls in model packages, several of which
+    # import this module's callers.
+    from sglang_omni.models.registry import PIPELINE_CONFIG_REGISTRY
+
+    with open(file_path, "r") as f:
```
JiaxinD: open() without encoding="utf-8" fails on zh-CN Windows for configs with non-ASCII comments (the fp8 example reproduces), and a missing file escapes as a bare FileNotFoundError; both deserve named refusals.

sglang_omni/cli/config.py
```diff
@@ -62,12 +82,233 @@ def export(
...
+        resolved = ConfigResolver(baseline).resolve(patches)
+        # The same post-merge derivation `serve` runs; it only fills engine
+        # keys no source set, so the previewed config is the launched one.
+        derived = apply_tensor_parallel_engine_overrides(resolved.config)
+        # The derivation may fill a key a source explicitly cleared (an
+        # engine key set to none is unset again, and unset keys are the
+        # derivation's to fill). Re-record what the launched config holds so
+        # explain's headline reports it, with the winner's own input kept on
+        # its line and the rewrite noted.
+        for patch in resolved.patches.ordered():
```
JiaxinD, line 177: This only re-records keys that already have patches, so a purely derived key gets no entry and explain falls through to "model default" for a value TP derivation set.

sglang_omni/config/resolver.py
```diff
@@ -0,0 +1,185 @@
...
+        self._base = base
+
+    @property
+    def config_cls(self) -> type[PipelineConfig]:
+        return type(self._base)
+
+    def resolve(self, patchset: ConfigPatchSet) -> ResolvedConfig:
+        patchset.require_no_conflicts()
+
+        data = self._base.model_dump()
```
JiaxinD, line 56: The baseline dump carries the already-derived name, so --config plus --model-path serves under the stale name; patching model_path should rederive the default name.

sglang_omni/config/sources.py
```diff
@@ -0,0 +1,453 @@
...
+        try:
+            data = yaml.safe_load(f)
```
stephenkgli: The YAML parser collapses duplicate mapping keys before patches are created. Writing `model_path` or `stages.thinker.gpu` twice silently keeps the second value, so same-precedence duplicate detection never sees the conflict.

sglang_omni/models/llada2_uni/config.py
```diff
@@ -22,40 +27,39 @@ class LLaDA2UniPipelineConfig(PipelineConfig):
...
     model_path: str
     stages: list[StageConfig] = [
         StageConfig(
             name=PREPROCESSING_STAGE,
             process="pipeline",
-            factory=f"{_PKG}.stages.create_preprocessing_executor",
-            factory_args={"thinker_max_seq_len": 8192},
-            runtime_arg_map={"max_seq_len": "thinker_max_seq_len"},
+            factory_path=f"{_PKG}.stages.create_preprocessing_executor",
+            factory=FactoryArgs(max_seq_len=8192),
```
stephenkgli, line 40: Both defaults set `factory.max_seq_len`, but the preprocessing and thinker factories accept `thinker_max_seq_len`. Worker startup rejects the default kwargs for both stages.

sglang_omni/config/path.py
```diff
@@ -0,0 +1,787 @@
...
+        "the entry stage is part of the stage topology and is declared by "
+        "the model's config class, not set from a user-facing source",
+    ),
+    (
+        "stages.*.name",
+        PathVisibility.INTERNAL,
+        "a stage name is its address; the config class declares it and every "
+        "user-facing source addresses the stage by that name",
+    ),
+)
```
stephenkgli, line 166: Only the stage collection/name and `entry_stage` are internal here. Routing and construction fields such as `next`, `terminal`, `stream_to`, and `factory_path` remain writable, so a config can replace the model-owned topology.

sglang_omni/config/placement.py
```diff
@@ -133,28 +127,15 @@ def resolve_gpu_stage_names(plan: StagePlacementPlan) -> set[str]:
...
     gpu = stage.gpu
     if gpu is None:
         return ()
     if isinstance(gpu, int):
-        if stage.tp_size > 1:
-            raise ValueError(
-                f"Stage {stage.name!r}: TP placement requires a list of "
-                f"{stage.tp_size} unique GPU ids, got scalar gpu={gpu}"
-            )
         return tuple(gpu for _ in range(stage.tp_size))
```
stephenkgli: These shape checks only run during model construction, while the launcher helpers mutate `tp_size` and `gpu` afterward. A mutated `tp_size=2, gpu=0` reaches this branch as two ranks on `(0, 0)`.

sglang_omni/models/dots_tts/config.py
```diff
@@ -63,71 +82,42 @@ def model_post_init(self, __context: Any = None) -> None:
...
-        ):
-            engine_value = latent_overrides.get(
-                engine_key, latent_engine.factory_args.get(engine_key, default)
+        # note (guozhihao-224): stream_slots must match backbone concurrency
+        # so a max_running_requests override cannot outrun vocoder admission
+        # after readiness. A pinned value that disagrees is refused here, on
+        # every rebuild; an unset value stays unset and is derived at launch
+        # (writing the derivation into the config would make it look pinned
+        # on the next merge).
+        stream_slots = self.stage_named("vocoder").factory.stream_slots
```
stephenkgli, line 111: The migration removed the binding between preprocessing and latent-engine `num_steps`/`max_generate_length`. A config can resolve with different values for the two stages, and the batched flow later rejects a request whose step count differs from the engine NFE.

sglang_omni/models/higgs_tts/config.py
```diff
@@ -96,24 +69,26 @@ def process_safe_edges(cls) -> frozenset[tuple[str, str]]:
...
-            },
-            gpu=0,
-            runtime=StageRuntimeConfig(
-                resources=StageResourceConfig(total_gpu_memory_fraction=0.10),
-            ),
-            terminal=True,
-            can_accept_stream_before_payload=True,
-        ),
-    ]
+            }
```
stephenkgli, line 106: The migration removed the consistency check for `stream_stride`, `stream_followup_stride`, and `initial_chunk_frames` between the TTS engine and vocoder. The two stages now accept different cadence values for the same stream.

sglang_omni/cli/config.py
```diff
@@ -62,12 +82,233 @@ def export(
...
+        canonical = canonicalize_dotted_key(path, resolution.baseline)
+        compiled = ConfigPath.parse(canonical, type(resolution.baseline))
+    except ConfigPathError as exc:
+        # Carries `did you mean:` suggestions, which a traceback would bury.
+        raise typer.BadParameter(str(exc)) from exc
+
+    if provenance.touched(compiled.raw):
+        print(provenance.explain(compiled.raw))
+        return
+    value = compiled.read(resolution.resolved.config)
```
stephenkgli: `ConfigPath.parse` accepts absent mapping/free-form keys, but this unguarded read raises `ConfigPathError`. `config explain env_defaults.OMP_NUM_THREADS` and absent `factory.*` keys terminate with a traceback.

sglang_omni/config/sources.py
```diff
@@ -0,0 +1,453 @@
...
+        ):
+            raise ValueError(f"{label}.select.stages must be a list of names")
+        missing = [name for name in wanted if name not in stage_names]
+        if missing:
+            raise ValueError(
+                f"{label}.select.stages names unknown stage(s) {missing}; "
+                f"this pipeline has: {', '.join(stage_names)}"
+            )
+        matched = [name for name in matched if name in wanted]
+    if select.get("engine"):
```
stephenkgli: `select.engine` is evaluated by truthiness and excluded stage names are not validated. `"false"` selects engines, `false` matches every stage, and a misspelled exclusion is silently ignored.

JiaxinD (approved):

PR #1516, opened 2026-08-13
[Perf][ASR] Cache the audio front-end constant tables across requests

JiaxinD:
## Motivation

A py-spy profile of the Fun-ASR serving process under load attributes about a quarter of its non-idle samples to rebuilding constant tables: `get_mel_banks` 10.8%, `_get_window` 7.2%, `_get_sinc_resample_kernel` 7.3%, `inverse_mel_scale` 2.2%. torchaudio 2.11 does not memoise any of them, so every request recomputes the mel filterbank, the window and the sinc resampling kernel from scratch even though they depend only on configuration.

## Modifications

* `sglang_omni/utils/audio.py`: memoise the sinc resample kernel by rate pair, options, device and dtype, then apply it directly. Falls back to `torchaudio.functional.resample` if those internals move. This path is shared, so every model that decodes audio at a rate other than the target benefits.
* `sglang_omni/utils/audio_features.py`: a `cached_fbank` helper that keeps the mel filterbank across requests. Used by Fun-ASR's `_extract_fbank` and by the two Ming speaker encoders (`ming_omni` talker, `ming_tts` reference encode), which rebuilt the same table per request. Zonos2 already builds its table once through torchaudio's `MelSpectrogram` module and is untouched.

## Accuracy Test

Bit-identity against the stock implementations is asserted, not approximated, all with `torch.equal`: 5 rate pairs for the resample path; 3 durations x 2 sample rates, the short-clip frame-length cap, and both window types (hamming for Fun-ASR, the kaldi default for Ming) for fbank. Each cache also asserts that it is actually hit, which caught a first version that silently fell back on every call.

## Benchmark

Fun-ASR on 1x H200, one 16-physical-core lane, c32 closed loop on `/v1/audio/transcriptions` with 48 kHz input, 80 s measured per round, arms interleaved, 2 rounds each:

| build | QPS | mean | p95 |
| --- | ---: | ---: | ---: |
| base | 154.5 | 0.207 s | 0.227 s |
| resample kernel cached | 164.2 (+6.3%) | 0.195 s | 0.216 s |
| both tables cached | **181.0 (+17.2%)** | 0.176 s | 0.196 s |

## Checklist

- [x] Add unit tests.

### DIFF
--- a/sglang_omni/models/fun_asr/configuration_fun_asr.py
+++ b/sglang_omni/models/fun_asr/configuration_fun_asr.py
@@ -254,9 +254,8 @@ def _fbank_with_cached_banks(
 ) -> torch.Tensor:
     """``kaldi.fbank`` for Fun-ASR's fixed options, with the mel table cached.
 
-    Note (Jiaxin Deng): bit-identity with ``kaldi.fbank`` under these options
-    is asserted in tests; options Fun-ASR never varies are inlined so the
-    cached table can be keyed by the few that do.
+    Note (Jiaxin Deng): the options Fun-ASR never varies are inlined so the
+    table can be keyed by the few that do; bit-identity is asserted in tests.
     """
     import torchaudio.compliance.kaldi as kaldi
 
--- a/sglang_omni/models/fun_asr/configuration_fun_asr.py
+++ b/sglang_omni/models/fun_asr/configuration_fun_asr.py
@@ -4,6 +4,8 @@ from __future__ import annotations
 
 from typing import Any, ClassVar
 
+import functools
+
 import numpy as np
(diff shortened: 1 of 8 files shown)


sglang_omni/models/fun_asr/configuration_fun_asr.py
```diff
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import functools
```
0xjeffro: why need this

0xjeffro (approved):

SandyLuXY (approved):
lgtm

PR #1517, opened 2026-08-13
feat(pipeline): add generic PD stage compilation

MelodyyyYin:
## Summary

PR1 introduces the topology, compile-time rewrite, placement, and launch identity needed for generic Prefill–Decode disaggregation. It expands one logical autoregressive stage into independently placed Prefill and Decode workers while keeping pipeline routing, output identity, and terminal behavior unambiguous.

This layer does not transfer KV pages or admit Decode work into a scheduler.

## Route and output identity

A PD-enabled logical stage `<name>` expands into:

- `<name>_prefill`, the destination for logical inbound routes;
- `<name>_decode`, the source/output identity observed by downstream waits, streaming, completion, and abort accounting.

The compiler therefore maintains separate maps:

- destination identity rewrites `next`, `stream_to`, dynamic route targets, and projected-payload destinations to Prefill;
- source/output identity resolves downstream `wait_for` and terminal completion through Decode.

This avoids treating the logical stage name as both a destination and a completed-output source.

## Terminal Prefill streaming

Prefill can produce the first sampled token and terminate the request, notably for `max_new_tokens=1`. That result remains on the normal tensor-bearing result path rather than becoming a continuation.

When this happens, the rewritten `stream_to` and `stream_done_to_fn` close the downstream stream as well as forwarding the terminal result. This preserves streaming completion and prevents the downstream result stage from waiting indefinitely.

## Configuration and launch plumbing

- Adds typed `PDConfig` and per-role placement.
- Adds frozen `PDExecution` metadata carrying physical role and partner.
- Keeps compiler-owned PD metadata outside strict user factory arguments.
- Requires explicit factory capability opt-in.
- Rejects invalid placement and generated-name collisions before launch.
- Runs expansion before process topology, endpoint allocation, and placement so Prefill and Decode receive independent runtime resources.
- Propagates typed PD metadata through runtime preparation, worker specifications, process launch, and `Stage`.

Pipelines without PD configuration retain their existing behavior.

## Scope

Included:

- logical PD expansion;
- distinct route-destination and output-source identities;
- typed role/partner metadata;
- independent placement and process topology;
- capability validation;
- coordinator completion and abort identity;
- terminal Prefill result and stream closure.

Not included:

- continuation schema or wire transport;
- KV transfer;
- Decode request reconstruction;
- scheduler admission or backpressure;
- model-specific resume state.

Those responsibilities are isolated in PR2–PR4.

## Stack

- **PR1 — #1517 (this PR):** base `main`
- **PR2 — #1538:** continuation and handoff contract, base `pd_kvcache`
- **PR3 — #1596:** generic runtime integration, base `pd_handoff`
- **PR4 — #1599:** Qwen adapter/result integration, base `pd_runtime`

## Validation

- `pytest -q tests/unit_test/pipeline` — **491 passed**
- Final stacked `pytest -q tests/unit_test/pipeline tests/unit_test/scheduling` — **769 passed**
- `git diff --check` — passed


jieyaokuz:
I noticed there is a recent failing test after 'run-ci' label attached. See [stage 8 test log file](https://github.com/sgl-project/sglang-omni/actions/runs/33043997695/job/98439119000?pr=1517)

This job failed after exhausting 3 retries, but the failure pattern looks unrelated to this PR's PD-disaggregation logic. Details are:

**Summary of the 3 retry attempts (Just search `short test summary` in the log file)**

| Attempt | Failed check(s) | Duration |
|---|---|---|
| 1/3 | `throughput_qps` 0.281 < 0.86; `output_tok_per_req_s` 1.0 < 3.4; `latency_mean_s` 42.114 > 12.2; `rtf_mean` 3.9282 > 1.12 (**4 checks failed**) | 147.41s |
| 2/3 | `rtf_mean` 1.3191 > 1.12 (**1 check failed**) | 68.05s |
| 3/3 | `rtf_mean` 1.1523 > 1.12(**1 check failed**) | 62.61s |

All three assertions came from `assert_speed_thresholds` at concurrency=16; no `Traceback`/exception appears anywhere in the log.

The reasons why I think this looks like CI-environment noise rather than a real regression:

- No exception, only threshold misses. Every failure is a plain `AssertionError` on a speed number, not a crash. ( searched the full raw log for `Traceback`, zero matches.)
- Severity drops monotonically across retries: 4 metrics badly missed → only `rtf_mean` missed → `rtf_mean` missed by ~3%. A deterministic logic bug would fail the same way every time; this pattern (converging toward the threshold) is the signature of shared-GPU/runner contention instead.
- Also, this pipeline never enables PD disaggregation. On this PR branch, `grep -rn "pd_disaggregation" --include="*.py" --include="*.yaml" . | grep -v "^./tests/"` only matches the compiler/schema definition files themselves. So no model factory or config anywhere sets `pd_disaggregation` on a real stage. Since `expand_pd_stages()` is a no-op whenever every stage's `pd_disaggregation` is `None`, the failure of the tests seems less likely to be caused by introducing PD.

I think we can escalate this to someone who has rights to rerun CI test. If a clean re-run passes, that would confirm this is a noise issue rather than something introduced by this PR, and it shouldn't block the merge. After that we are able to work on [roadmap 1760](https://github.com/sgl-project/sglang-omni/issues/1760) since it's blocked by this PR.

PR #1381, opened 2026-08-07
[Qwen3-Omni Perf] Add Breakable Prefill CUDA Graph support for Thinker text/audio prefill

charliechenye:
## Motivation

Qwen3-Omni Thinker prefill is an important contributor to time-to-first-token, especially under concurrent serving. The Qwen3-Omni performance roadmap tracks Prefill CUDA Graph as one opportunity to reduce repeated launch/dispatch overhead in this path.

Two prerequisite efforts establish the pieces needed to support this cleanly:

- #1161 optimizes the Qwen3-Omni Thinker multimodal embedding merge and removes the per request device-to-host synchronization points that made prefill graph capture impractical.
- #1364 introduces the model agnostic Breakable Prefill CUDA Graph (BCG) integration in SGLang Omni, including the `OmniPrefillInputs` payload contract, graph static `input_embeds` storage, bucket policy, startup attestation, and shared eager fallback.

This PR connects those pieces for Qwen3-Omni without moving graph policy into model specific code.

The model-local adopter only decides whether a Qwen Thinker prefill can be represented safely by the shared payload contract. SGLang remains the sole owner of graph admission, bucket selection, padding, replay, and shape-based eager fallback.

## Scope

The initial BCG adopter is intentionally limited to **text output** and to prefills representable through `OmniPrefillInputs`, including:

- text → text;
- audio input → text output;
- mixed text + audio input → text output.

Visual/deepstack and speech output paths remain outside the BCG sidecar in this PR.

The default Qwen3 Omni prefill graph backend is **not** changed; this remains an explicit opt-in optimization.

### Qwen3-Omni Thinker adopter

Add a model local `Qwen3OmniThinkerModelRunner` derived from the existing `ThinkerModelRunner`.

For a payload compatible prefill batch, it:

1. validates that the **whole batch** is representable by the current Qwen sidecar contract;
2. composes the normal text/audio prefill embeddings using the existing optimized Thinker merge;
3. attaches those embeddings through `OmniPrefillInputs`;
4. deliberately leaves `ForwardBatch.input_embeds` unset during graph admission;
5. delegates execution back to the normal SGLang model runner.

The Qwen runner does not implement a graph runner, graph bucket policy, or model specific eager fallback policy.

### Model side payload consumption

Update the Qwen3 Omni text output Thinker wrapper to consume `OmniPrefillInputs` as the prefill embedding source.

For sidecar prefill execution:

- the inner text model receives `input_embeds`;
- the original token ids remain available to the logits processor;
[...]

PR #1519, opened 2026-08-13
[Qwen3-Omni Perf] Enable breakable prefill graphs for speech

edwingao28:
## Motivation

PR #1380 added graph-stable layer-0/layer-24 capture for the speech handoff, and
PR #1381 added the Qwen thinker sidecar that lets SGLang own stable multimodal
embedding buffers, bucket padding, replay, and fallback. Speech still keeps the
prefill backend disabled in bootstrap even though those contracts make an opt-in
breakable path possible.

This PR builds on the now-merged #1381. The speech default remains unchanged.

## Modifications

- Allow explicitly configured breakable prefill graphs for the speech thinker,
  using the shared Qwen runner and graph initializer. Disabled-backend speech
  retains the existing generic runner.
- Keep the talker handoff on stable, cloned logical rows from layers 0 and 24;
  add bounded, environment-gated snapshots and positive replay/standard-eager/
  custom-eager counters for qualification only.
- Add a qualification-only eager replay mode that preserves graph admission,
  bucket padding, static buffers, and live serving metadata while bypassing only
  the final captured-body launch. Runtime attestation prevents it from being
  confused with production replay or SGLang's upstream debug mode.
- Add an opt-in H100 BF16 profile with the canonical 42-bucket ladder through
  2048 tokens. Larger prefills remain valid and run eagerly.
- Add a fail-closed H100 CI stage that checks text/audio replay, same-bucket
  refresh, exact embedding transport, finite hidden/logit tensors, matched-shape
  eager-vs-graph parity, repeated-audio determinism, valid WAV output, visual
  custom eager fallback, and over-cap standard eager fallback.
- Use pinned vLLM Omni commit
  `d77a35a32b25d1460ac5c72f66f4d9fd554399a2` as an architectural reference for
  stable replay storage and the explicit thinker/talker hidden-state boundary;
  SGLang remains the graph-runtime authority.

## Related Issues

- Builds on merged #1381
- Builds on #1380

## Accuracy Test

- Focused bootstrap, sidecar, hidden-capture, graph-policy, and shared-helper
  suite: `157 passed, 1 skipped`. The skip is the real-CUDA capture test on a
  non-CUDA development host.
- Streaming/output suite: `38 passed`.
- Qualification assertion/helper suite: `13 passed` locally; the exact-head
  Linux PR unit job also passed.
- Exact-head H100 run
  [31735907211](https://github.com/sgl-project/sglang-omni/actions/runs/31735907211)
  passed all Omni CI jobs, including Stage 11, on
  `4a188b407eb3fcaa605e61b91b329663a2c75c02`. The
  [Stage 11 artifact](https://github.com/sgl-project/sglang-omni/actions/runs/31735907211/artifacts/9199341326)
  records `qualification_status=passed`.
- Shape-matched text, same-bucket refresh, and audio comparisons had exact
  layer-0 embeddings, layer-24 hidden states, and next-token logits:
  `max_abs_error=0.0`, zero mismatches, and zero non-finite values. Repeating the
  captured audio request was also bit-exact, with the same token and valid
  non-empty WAV output.
- Production graph replay counters were `4` (`20`-token bucket twice and
  `64`-token bucket twice). The visual request took one custom eager fallback,
  and the 2419-token over-cap request took one standard eager fallback.
- MMSU keeps its original disabled-prefill-graph path; this PR does not alter
  sampling or MMSU retry policy. Stages 5 and 6 passed on the exact head.

## Compatibility

- The repository currently pins SGLang `0.5.16`; the H100 result above qualifies
  that pinned stack.
- SGLang `0.5.17` removes the private
  `PrefillCudaGraphRunner._static_num_tokens` field used only by the
  qualification counter. Production replay does not depend on that field, but
  the observer must be adapted and Stage 11 rerun as part of the 0.5.17 upgrade.

## Benchmark & Profiling
**Setup**: 
Setup: single H200, colocated Qwen3-Omni speech pipeline, SeedTTS-en voice clone, greedy decoding, byte-identical client in both arms, server boots alternated between arms. Branch head bbc2bbc (already merged with b3bbff6e). 8,448 speech requests, 0 failures.

| Conc | TTFT p50 (base→PR) | Δ TTFT | TTFT p95 (base→PR) | TTFA p50 (base→PR) | Δ TTFA | E2E p50 (base→PR) | Δ E2E | RTF p50 (base→PR) | req/s (base→PR) | Δ rps | audio_s parity | TTFT round scatter (base/PR) |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| c1 | 170 → 139 ms | **−18.0%** | 205 → 166 ms | 294 → 254 ms | −13.5% | 0.57 → 0.55 s | −3.7% | 0.185 → 0.164 | 1.62 → 1.75 | +7.5% | −0.59% | 1.2% / 2.0% |
| c8 | 291 → 265 ms | **−8.9%** | 458 → 416 ms | 589 → 559 ms | −5.1% | 1.35 → 1.30 s | −3.2% | 0.415 → 0.397 | 5.54 → 5.95 | **+7.3%** | −1.01% | 2.4% / 3.2% |
| c32 | 393 → 403 ms | +2.3% | 1024 → 915 ms | 1358 → 1471 ms | +8.4% | 3.44 → 3.58 s | +3.9% | 0.968 → 1.012 | 9.08 → 8.80 | −3.1% | −0.04% | 7.1% / 5.7% |
| c64 | 477 → 461 ms | −3.3% | 1975 → 1847 ms | 2445 → 2454 ms | +0.3% | 6.45 → 6.50 s | +0.8% | 1.758 → 1.784 | 10.05 → 10.01 | −0.4% | −0.45% | 9.2% / 43.0%* |

## Checklist

- [x] Format code with pre-commit, Black, isort, Ruff, and actionlint.
- [x] Add unit and integration tests.
- [x] Add the opt-in example config and targeted runtime docstrings.
- [x] Attach passing H100 replay/parity evidence from the qualified exact head.
- [ ] For reviewers: If you haven't made any contributions to this PR and are
  only assisting with merging the main branch, please remove yourself as a
  co-author when merging the PR.

## CI

This PR is ready-for-review rather than draft because Omni CI skips drafts.
`/tag-and-rerun-ci moss fun-asr` selects the same auxiliary lanes used by #1381;
Qwen3-Omni Stage 11 runs the speech prefill qualification gate.


charliechenye:
Really like the direction of this PR. Architecturally this feels like the right composition of #1380 and #1381: keep SGLang as the owner of graph admission, bucket selection, padding, replay, and fallback, while the Qwen side keeps ownership of the stable layer 0 / layer 24 Talker handoff. I also like that visual / DeepStack and over cap cases remain explicitly eager rather than broadening the graph contract in this PR.

I looked through the current H100 Stage 11 evidence, and I think the failure is useful, but I am not yet convinced it proves an audio BCG correctness bug.

The text cases look very strong. `text -> text + audio` actually replays BCG, layer 0 and layer 24 match eager exactly, logits match exactly, and the second request reusing the same bucket also refreshes correctly. That is pretty good evidence that the #1380 capture buffers themselves survive replay correctly.

For audio, the failure signature is much more specific:

`audio BCG input` -> layer 0 is exact -> layer 24 diverges -> logits diverge, while the first sampled token still matches.

The same process cached source control is a good idea because it removes audio encoder variation. One concern though: the eager oracle and BCG arm are not execution shape matched. The BCG audio request runs as the single 61 row request in a 64 token graph bucket, while the custom eager oracle is forced by coalescing the same audio request with a visual sibling, so that Thinker executes a substantially different mixed batch. That means the comparison currently changes both graph vs eager and transformer batch shape / composition at the same time.

Given the MoE body, I would be careful about interpreting the layer 24 delta as graph corruption until those variables are separated. The exact layer 0 match tells us the encoder output and sidecar transport are probably not where I would investigate first.

Would it be possible to add a qualification only matched shape oracle where the exact same audio request, embeddings, M-RoPE state, and batch composition run once through standard eager and once through BCG in the same process? A small diagnostic forcing one otherwise eligible prefill to eager, or a lower level Thinker test using a fixed captured audio embedding, would both seem cleaner than using a visual sibling to induce eager fallback.

If that matched comparison still diverges, I think the next experiment could be very narrow: compare an exact 61 token capture against the padded 64 token capture, then snapshot M-RoPE and one or two intermediate layer boundaries to locate the first divergence. If 61 is clean and 64 diverges, that points toward padding / graph metadata / shape-sensitive execution. If both diverge, then I would look much harder at replay state itself.

I would definitely resist loosening the 2e-2 tolerance based on the current result. The existing failure is doing exactly what a good qualification gate should do: giving us enough evidence to isolate the mechanism before calling the path safe.

Happy to help dig into the matched eager/BCG experiment if useful. This looks like a really promising continuation of the Thinker graph work.

charliechenye:
One 0.5.17 compatibility point I think is worth resolving before we treat the next H100 Stage 11 run as final qualification.

I compared the prefill runner contracts in 0.5.16 and 0.5.17. The production speech BCG approach still looks compatible to me: `_run_forward`, `_prefill_forward_context`, and the simple `BreakableCudaGraphBackend.replay()` boundary are still present, and #1381's late bound `input_embeds` contract still makes sense because 0.5.17 continues to reject a populated `ForwardBatch.input_embeds` during graph admission.

There is one concrete issue in the qualification instrumentation though. `_record_prefill_cuda_graph_usage()` currently reads `PrefillCudaGraphRunner._static_num_tokens` to positively identify the executed bucket. 0.5.17 no longer persists that field. `execute()` computes `shape_key = self._shape_key(static_num_tokens, forward_batch)` locally and passes it into `_execute_body_capture()` instead.

So I think the production path may survive #1477 mostly unchanged, but the current Stage 11 observer will need a small adaptation and the H100 evidence will need to be rerun on 0.5.17.

Given #1477 is already in flight, do we know the intended merge order? If #1477 is still some distance away, qualifying the current 0.5.16 path first still seems reasonable.

edwingao28:
@charliechenye Thanks — the matched-shape oracle is now in place and the exact-head H100 run passed: https://github.com/sgl-project/sglang-omni/actions/runs/31735907211

For the same 61-row audio request in the same 64-token bucket, qualification eager-replay versus captured BCG was exact at the input embeddings, layer 24, and next-token logits (`max_abs_error=0.0`, zero mismatches/non-finite values, same next token). Repeating the captured audio request was also bit-exact and produced the same valid non-empty WAV. Text and same-bucket refresh were exact as well. No tolerance was loosened. Artifact: https://github.com/sgl-project/sglang-omni/actions/runs/31735907211/artifacts/9199341326

I also confirmed the 0.5.17 point: `_static_num_tokens` was intentionally removed when `ShapeKey` became execute-local. That affects only our qualification counter, not production replay, but would make Stage 11 fail closed because replays are not counted. The current repository is pinned to 0.5.16, so this run qualifies that stack. Since #1477 is still open, the clean order is #1519 first, then adapt the observer and rerun Stage 11 as part of the 0.5.17 upgrade.

edwingao28:
## H200 Speech Pipeline A/B: breakable prefill CUDA graphs vs disabled

**Setup**: 
- Qwen3-Omni-30B-A3B-Instruct speech pipeline (thinker+code2wav / talker on 2×H200), SGLang 0.5.16, 
- identical tree `4a188b4` both arms — config-only diff: `cuda_graph_backend_prefill=breakable` + buckets `[256,512,768,1024,1536,2048]` vs disabled. Radix cache off, `chunked_prefill_size=2048`, deterministic ~643-token text prompts, temp 0, `max_tokens=96`
- speech (text+audio) output. 3 fresh rounds × c{1,2,4,8,16}, warmup excluded, **0 errors both arms** (276 requests/arm). TTFT = first text SSE delta, TTFA = first audio delta.

| Conc | TTFT p50 (base→PR) | Δ TTFT | TTFT p95 (base→PR) | TTFA p50 (base→PR) | Δ TTFA | E2E p50 (base→PR) | Δ E2E | RTF p50 (base→PR) | req/s (base→PR) | Δ rps | audio_s parity | TTFT round scatter (base/PR) |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| c1 | 284 → 253 ms | **−10.7%** | 295 → 263 ms | 284 → 253 ms | **−10.7%** | 2.27 → 2.22 s | −2.3% | 0.094 → 0.093 | 0.44 → 0.45 | +0.8% | ±1.1% | 1.7% / 2.4% |
| c2 | 313 → 284 ms | **−9.3%** | 444 → 362 ms | 313 → 284 ms | **−9.3%** | 2.47 → 2.41 s | −2.5% | 0.102 → 0.101 | 0.78 → 0.80 | +2.6% | ±1.7% | 1.7% / 2.0% |
| c4 | 423 → 373 ms | −11.8%* | 661 → 595 ms | 423 → 373 ms | −11.8%* | 2.85 → 2.83 s | −0.7% | 0.117 → 0.114 | 1.35 → 1.35 | 0% | ±2.5% | 14.9% / 5.6% |
| c8 | 747 → 725 ms | −3.0%* | 910 → 882 ms | 747 → 725 ms | −3.0%* | 3.73 → 3.66 s | −1.9% | 0.150 → 0.149 | 1.88 → 1.94 | +3.6% | ±2.8% | 13.3% / 22.9% |
| c16 | 1531 → 1425 ms | **−6.9%** | 1619 → 1524 ms | 1531 → 1427 ms | **−6.8%** | 6.29 → 6.17 s | −1.8% | 0.247 → 0.253 | 2.48 → 2.50 | +0.7% | ±0.3% | 2.6% / 5.6% |

examples/configs/qwen3_omni_colocated_h100_bf16.yaml
```diff
@@ -4,6 +4,56 @@ config_cls: Qwen3OmniSpeechColocatedPipelineConfig
 name: qwen3-omni-colocated-h100-bf16
 model_path: Qwen/Qwen3-Omni-30B-A3B-Instruct
 
+runtime_overrides:
+  thinker:
+    server_args_overrides:
+      disable_radix_cache: true
+      cuda_graph_backend_prefill: breakable
+      cuda_graph_bs_prefill:
```
AkazaAkane: Could we add the workload evidence for this 2048-token / 42-bucket capture budget?

The perf A/B in the PR used only [256, 512, 768, 1024, 1536, 2048], while the config being merged captures the full 42-bucket ladder. I don't see the actual scheduled prefill token distribution (min/median/p95/max) that justifies 2048 either.

Before merging, I think we should either benchmark the exact bucket configuration being shipped, or derive the qualified cap from measured scheduled-prefill shapes and use the shared build_default_prefill_cuda_graph_bs(cap) policy rather than duplicating the ladder here.

examples/configs/qwen3_omni_colocated_h100_bf16.yaml
```diff
@@ -4,6 +4,56 @@ config_cls: Qwen3OmniSpeechColocatedPipelineConfig
 name: qwen3-omni-colocated-h100-bf16
 model_path: Qwen/Qwen3-Omni-30B-A3B-Instruct
 
+runtime_overrides:
+  thinker:
+    server_args_overrides:
+      disable_radix_cache: true
```
AkazaAkane: Is disabling radix cache intended to become part of the normal BF16 colocated profile, or was this only for qualification?

The reported A/B also disables radix cache, so it establishes the graph result for cold/full prefills, but not for the normal cache-hit path. If radix caching is a supported deployment mode here, I think we should either keep the normal cache behavior and qualify representative cache-hit shapes, or make this explicitly a qualification/no-radix profile instead of changing the existing profile's semantics.

AkazaAkane (changes requested):
The replay/correctness story looks good to me now. I think there are some remaining qualification gaps before merge:

1. Could we determine the prefill capture budget from the actual scheduled prefill token distribution rather than hard-coding 2048?

I think we should first measure the graph-relevant token counts that actually reach the prefill runner (min/p50/p90/p95/p99/max), then compare replay coverage for a few candidate caps such as 1024/1536/2048. The selected cap should cover most real traffic without letting rare long prefills dictate capture time and GPU memory.

Once the cap is chosen, the bucket list should come from the shared build_default_prefill_cuda_graph_bs(cap) policy rather than being manually selected. Then benchmark the exact final configuration and report replay/fallback coverage plus capture/startup/memory cost.

2. Please measure startup/memory cost for the actual configuration being merged: PCG capture time, cold startup delta, allocated/reserved GPU memory delta, and remaining GPU headroom. The current profile captures 42 shapes, so successful boot alone isn't enough to tell us whether that capture budget is reasonable.

3. Please rerun current-head CI after fixing lint.

sglang_omni/model_runner/model_worker.py
```diff
@@ -68,6 +69,10 @@ def __init__(
         self._configure_backend_policy()
         self._init_model_runner()
         self._init_dllm_algorithm()
+        self._prefill_cuda_graph_replay_count = 0
+        self._prefill_cuda_graph_standard_eager_count = 0
+        self._prefill_cuda_graph_custom_eager_count = 0
+        self._prefill_cuda_graph_replay_buckets: Counter[int] = Counter()
```
stephenkgli: would be better to use a data class for these PCG fields.

sglang_omni/model_runner/base.py
```diff
@@ -440,6 +440,14 @@ def _prepare_and_forward(
...
                     forward_batch, schedule_batch, requests
                 )
+                if batch_result is not None:
+                    record_custom_eager = getattr(
+                        self.tp_worker,
+                        "record_custom_prefill_eager",
+                        None,
+                    )
+                    if callable(record_custom_eager):
+                        record_custom_eager()
```
stephenkgli: `_prepare_and_forward` is a high-level orchestration method with several extension points and hooks. PCG-specific logic should live behind one of those extension points rather than being embedded here. Please keep this method focused and move the instrumentation to a dedicated hook.

sglang_omni/scheduling/sglang_backend/output_processor.py
```diff
@@ -209,23 +210,47 @@ def _slice_per_request_tensor(
         *,
         request_index: int,
         scheduler_output: SchedulerOutput,
+        prefer_token_axis: bool = False,
```
stephenkgli: This function already infers the tensor layout from its shape, but that approach is fragile. Please determine whether the tensor is request-major or token-major based on the execution phase—prefill or decode—instead.

sglang_omni/model_runner/model_worker.py
edwingao28: Fixed in [7ca4f374](https://github.com/sgl-project/sglang-omni/commit/7ca4f374309eb5b1ad722328f3672456ea473ab0).

The PCG counters and replay buckets now live in a `_PrefillCudaGraphUsage` dataclass with an isolated `Counter` default, while the `/model_info` schema remains unchanged. Added regression coverage for state isolation and replay/eager serialization.

sglang_omni/model_runner/base.py
edwingao28: Fixed in [00a54380](https://github.com/sgl-project/sglang-omni/commit/00a5438002005a0f1d58b944c7d96b33d65015ce).

`_prepare_and_forward` now calls a generic `on_custom_prefill_forward` hook, and only Qwen3-Omni overrides it to record custom eager prefill usage. Added base ordering/fallback coverage and Qwen delegation coverage.

sglang_omni/scheduling/sglang_backend/output_processor.py
edwingao28: Fixed in [a65db0d8](https://github.com/sgl-project/sglang-omni/commit/a65db0d83344e52d271d30aed51e20e84eed7477).

Static auxiliary hidden capture now uses a dedicated phase-defined slicer: token-major for prefill and request-major for decode, with strict row-count validation. LAST/stream hidden tensors retain their request-major contract. Added ambiguous-shape and malformed-row regression coverage.

edwingao28:
## Benchmark: this branch vs upstream main (`b3bbff6e`)

**TL;DR** — This PR makes prefill CUDA graphs usable on the speech path. Measured against latest main: time-to-first-token drops up to **39%**, every metric improves at moderate concurrency (C8), and the text-only path is verified unaffected.

**Why the whole speech delta is attributable to this PR:** upstream main refuses to start when a prefill CUDA graph backend is enabled together with speech output (`bootstrap.py` raises at boot — we verified this arm cannot boot). This PR removes that guard, so these gains are unreachable on main today.

**Setup:** single H200, colocated Qwen3-Omni speech pipeline, SeedTTS-en voice clone, greedy decoding, byte-identical client in both arms, server boots alternated between arms. Branch head `bbc2bbc` (already merged with `b3bbff6e`). 8,448 speech requests, 0 failures.

### Speech pipeline (median across boots)

| C | req/s | TTFT mean | TTFT p95 | TTFA mean | TTFA p95 |
|---|---|---|---|---|---|
| first request after idle (C1) | — | **−38.8%** | — | −18.5% | — |
| 1 (3 boots) | +7.5% | **−24.8%** | −20.5% | −15.4% | −14.0% |
| 8 (2 boots) | **+7.3%** | **−11.7%** | **−14.0%** | −6.5% | **−18.6%** |
| 32 (3 boots) | −3.1% | −2.1% | −18.8% | +9.8% | −1.2% |
| 64 (4 boots) | −0.4% | −5.0% | −3.7% | +2.7% | +1.9% |


### Text-only path: no impact

For text, the prefill graph feature already exists on main (#1381); this PR only adds usage counters there. Measured on a separate 3-arm sweep: main-with-knob vs this-branch-with-knob differ by less than run-to-run noise (no regression), and the new counters report every thinker prefill replayed through the graph (replay count 371/189 across boots, eager count 0). For reference, that pre-existing knob cuts text TTFT by 62–82% — this PR extends the same class of benefit to speech.

Raw runs, configs, and per-boot server logs are archived and available on request.

---

### Same data, p50 view (per-request medians, pooled per boot then median across boots)

| Conc | TTFT p50 (base→PR) | Δ TTFT | TTFT p95 (base→PR) | TTFA p50 (base→PR) | Δ TTFA | E2E p50 (base→PR) | Δ E2E | RTF p50 (base→PR) | req/s (base→PR) | Δ rps | audio_s parity | TTFT round scatter (base/PR) |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| c1 | 170 → 139 ms | **−18.0%** | 205 → 166 ms | 294 → 254 ms | −13.5% | 0.57 → 0.55 s | −3.7% | 0.185 → 0.164 | 1.62 → 1.75 | +7.5% | −0.59% | 1.2% / 2.0% |
| c8 | 291 → 265 ms | **−8.9%** | 458 → 416 ms | 589 → 559 ms | −5.1% | 1.35 → 1.30 s | −3.2% | 0.415 → 0.397 | 5.54 → 5.95 | **+7.3%** | −1.01% | 2.4% / 3.2% |
| c32 | 393 → 403 ms | +2.3% | 1024 → 915 ms | 1358 → 1471 ms | +8.4% | 3.44 → 3.58 s | +3.9% | 0.968 → 1.012 | 9.08 → 8.80 | −3.1% | −0.04% | 7.1% / 5.7% |
| c64 | 477 → 461 ms | −3.3% | 1975 → 1847 ms | 2445 → 2454 ms | +0.3% | 6.45 → 6.50 s | +0.8% | 1.758 → 1.784 | 10.05 → 10.01 | −0.4% | −0.45% | 9.2% / 43.0%* |

audio_s parity = total generated audio seconds, PR vs base (within ±1% everywhere → both arms did the same work). \*The c64 PR-side 43% scatter is the first-boot warmup anomaly described above; boots 2–4 sit within 9% of each other.

examples/configs/qwen3_omni_colocated_h100_bf16.yaml
edwingao28: Mined the scheduled-prefill `#new-tokens` distribution from the posted 24-cell A/B (both arms):

| Level | batches >2048 | prefill tokens in them | max batch |
|---|---|---|---|
| C1 | 0% | 0% | ≤2048 |
| C8 | 0% | 0% | 1115 |
| C32 | 0.3% | ~2.4% | — |
| C64 | 3.5% | ~26–32% | 6797 |

Why 2048:
- Covers 100% of batches at C1/C8 — where this PR's wins live
- 1024→2048 converts the (1024, 2048] band (~26% of C64 base-arm prefill tokens) from eager to graphed
- Tail is unbounded (6797 observed), so no finite cap closes C64; further doublings buy less per capture time/memory cost

Will switch the example config to derive the ladder via the shared `build_default_prefill_cuda_graph_bs(cap)` policy instead of an inline bucket list. Full distribution write-up archived, can attach.

examples/configs/qwen3_omni_colocated_h100_bf16.yaml
edwingao28: - `disable_radix_cache: true` was qualification leftover — already removed from the branch (thread shows outdated)
- The posted 24-cell benchmark ran with radix cache **enabled** (default, both arms), on a voice-clone workload sharing the reference-audio prefix — most numbers are the cache-hit path
- Cold/warm split covers both regimes:

| Regime (C1) | TTFT | TTFA |
|---|---|---|
| cold (uncached prefix) | −38.8% | −18.5% |
| warm (cache hit) | −19.1% | −11.1% |

edwingao28:
@AkazaAkane @stephenkgli All comment addressed, can you PTAL?

sglang_omni/model_runner/model_worker.py
```diff
@@ -277,13 +287,92 @@ def forward_batch_generation(
...
+            return
+
+        if not can_run_graph:
+            # Note (wenyao): custom eager forwards (visual/deepstack) return
+            # before ModelWorker is called; intentionally absent here.
+            self._prefill_cuda_graph_usage.standard_eager_count += 1
+            return
+
+        runner = getattr(self.model_runner, "prefill_cuda_graph_runner", None)
+        actual_bucket = getattr(runner, "_static_num_tokens", None)
```
Ratish1: `_static_num_tokens` is a private runner field. The executed bucket is public state: the runner pads to the smallest captured bucket >= num_tokens and `capture_num_tokens` is sorted, so

```python
buckets = runner.capture_num_tokens
bucket = buckets[bisect.bisect_left(buckets, len(forward_batch.input_ids))]
```

gives the same value without the private read or the warning branch. Update the usage-test fake to set `capture_num_tokens` only.

sglang_omni/model_runner/base.py
```diff
@@ -638,6 +645,15 @@ def custom_prefill_forward(
         """
         return None
 
+    def on_custom_prefill_forward(
```
Ratish1: This hook exists only so the Qwen thinker runner can call `record_custom_prefill_eager`. `custom_prefill_forward` is already the extension point that produces the result, and the Qwen runner already overrides it. Increment there and drop this hook and its call in `_prepare_and_forward`:

```python
result = super().custom_prefill_forward(forward_batch, schedule_batch, requests)
if result is not None:
    self.tp_worker.record_custom_prefill_eager()
return result
```

The base runner then has no change in this PR.

sglang_omni/model_runner/model_worker.py
```diff
@@ -277,13 +287,92 @@ def forward_batch_generation(
...
         )
         return batch_result
 
+    def _record_prefill_cuda_graph_usage(
+        self,
+        forward_batch: Any,
+        *,
+        can_run_graph: bool,
+    ) -> None:
+        forward_mode = getattr(forward_batch, "forward_mode", None)
```
Ratish1: dont use getattr

```python
mode = forward_batch.forward_mode
if not mode.is_extend() or mode.is_cuda_graph():
    return
```

sglang_omni/model_runner/model_worker.py
```diff
@@ -277,13 +287,92 @@ def forward_batch_generation(
...
+            return
+        self._prefill_cuda_graph_usage.replay_count += 1
+        self._prefill_cuda_graph_usage.replay_buckets[int(actual_bucket)] += 1
+
+    def record_custom_prefill_eager(self) -> None:
+        """Record a custom prefill forward that bypasses SGLang graph dispatch."""
+        self._prefill_cuda_graph_usage.custom_eager_count += 1
+
+    def _prefill_cuda_graph_info(self) -> dict[str, Any]:
+        runner = getattr(self.model_runner, "prefill_cuda_graph_runner", None)
```
Ratish1: dont use getattr

```python
runner = self.model_runner.prefill_cuda_graph_runner
if isinstance(runner, PrefillCudaGraphRunner):
    capture_num_tokens = [int(v) for v in runner.capture_num_tokens]
    backend_runner = type(runner.backend).__name__
    input_embeds_slot = runner.buffer_registry.has_slot("input_embeds")
else:
    capture_num_tokens, backend_runner, input_embeds_slot = None, None, False
backend = self.server_args.cuda_graph_config.prefill.backend
```

`prefill_cuda_graph_runner` is None, an EagerRunner, or a PrefillCudaGraphRunner; upstream reads it unguarded and `attest_prefill_cuda_graphs` already uses the same isinstance.

sglang_omni/scheduling/sglang_backend/output_processor.py
```diff
@@ -214,18 +248,23 @@ def _slice_per_request_tensor(
...
+
+        if len(requests) == 1:
+            return tensor[0] if tensor.ndim >= 2 else tensor
         if tensor.shape[0] == num_requests:
             return tensor[request_index]
 
-        lengths = [req.extend_range.length for req in reqs]
-        total_tokens = sum(lengths)
-        if tensor.shape[0] == total_tokens:
+        is_extend_fn = getattr(
```
Ratish1: dont use getattr here also

sglang_omni/models/qwen3_omni/talker_model_runner.py
```diff
@@ -32,12 +37,39 @@ def __init__(
...
+        input_embeds, input_embeds_are_projected = composed
+        attach_omni_prefill_inputs(
+            forward_batch,
+            OmniPrefillInputs(
+                input_embeds=input_embeds,
+                input_embeds_are_projected=input_embeds_are_projected,
+            ),
+        )
+
     def custom_prefill_forward(
```
Ratish1: With projected rows attached in `before_prefill`, `_run_projected_prefill_forward` and the talker's `_forward_with_input_embeds` are reachable only from tests: the single talker request builder sets `input_embeds_are_projected=True`, so `Req.input_embeds` is None and `forward_batch.input_embeds` is never set. Keep one composition route: delete both, drop the `forward_batch.input_embeds` early return in `before_prefill`, reduce `_compose_prefill_embeds` to `any(projected_flags)`, and point the projected-prefill tests at `before_prefill` + `get_omni_prefill_inputs`. The `forward_batch.input_embeds` case needs no custom forward: upstream already forwards it as the `input_embeds` kwarg and `Qwen3OmniTalker.forward` projects it.

sglang_omni/scheduling/generation_batch_policy.py
```diff
@@ -155,6 +155,13 @@ def build_generation_batch_overrides(
     # operator-stated list is never trimmed; contradictions fail validation.
     prefill_bs = overrides.get("cuda_graph_bs_prefill")
     prefill_max_bs = overrides.get("cuda_graph_max_bs_prefill")
+    if (
```
Ratish1, line 183: This derives the ladder from `cuda_graph_max_bs_prefill` alone. qwen3_asr's `adjust_overrides` derives the same ladder from the min of `max_prefill_tokens`, `cuda_graph_max_bs_prefill`, `max_total_tokens`, `context_length` and `chunked_prefill_size`, and returns early once `cuda_graph_bs_prefill` is set. Since this runs first, a Qwen3-ASR override with `cuda_graph_max_bs_prefill` above `max_prefill_tokens` now fails startup in `_validate_prefill_graph_policy` where it previously clamped, and there are two policies for one knob.

Clamp here over the caps present in `overrides` (`chunked_prefill_size` when > 0, `max_prefill_tokens`, `max_total_tokens`, `context_length`), set `cuda_graph_max_bs_prefill` to the clamped cap, share the helper with qwen3_asr, and add the over-cap case to the policy tests.

tests/unit_test/qwen3_omni/test_thinker_prefill_graph_gate.py
```diff
@@ -0,0 +1,53 @@
...
+"""
+
+from __future__ import annotations
+
+import inspect
+
+from sglang_omni.models.qwen3_omni import bootstrap as qwen3_omni_bootstrap
+
+
+def _source() -> str:
```
Ratish1: These assert substrings of `inspect.getsource(bootstrap)`; they pass for any file containing the tokens and fail on a rename without exercising behavior. The wiring is already covered by the existing bootstrap tests (graph init worker, attestation call, `enable_prefill_input_embeds`). Delete the file.

tests/unit_test/qwen3_omni/test_talker_prefill_graph.py
```diff
@@ -0,0 +1,66 @@
...
+from types import SimpleNamespace
+
+
+def test_talker_advertises_mrope_from_its_backbone() -> None:
+    """Derived from the real rotary_emb instances, never hardcoded."""
+    import inspect
+
+    from sglang_omni.models.qwen3_omni.components import talker as talker_mod
+
+    src = inspect.getsource(talker_mod)
```
Ratish1: Source-text assertion; delete it. The two tests below exercise sglang's `_get_layer_model_positions`, not omni code; drop them too.

sglang_omni/model_runner/prefill_inputs.py
```diff
@@ -28,9 +28,14 @@ class OmniPrefillInputs:
     input_embeds covers exactly the extend-window tokens of the batch, in
     model dtype. Request identity remains owned by ``ForwardBatch.rids`` and
     is forwarded separately by ``SGLModelRunner``.
+
+    input_embeds_are_projected stays ``None`` for models whose forward does
+    not declare the parameter; only a runner composing embeddings already in
+    model space sets it, and only then is it forwarded.
```
Ratish1, line 34: dont use backticks in docstrings

sglang_omni/model_runner/model_worker.py
edwingao28: Fixed in [55003611](https://github.com/sgl-project/sglang-omni/commit/550036117edf5228283acc408c5a484e2d4cdb59).

Replay buckets now come from the sorted public `capture_num_tokens` list and the live input length. Removed the private `_static_num_tokens` read and updated the usage test to use public runner state only.

sglang_omni/scheduling/sglang_backend/output_processor.py
edwingao28: Fixed in [6e9dfa5a](https://github.com/sgl-project/sglang-omni/commit/6e9dfa5a76c6b94dd724d499d910d297db76a083).

Per-request hidden slicing now reads `batch_data.forward_mode.is_extend()` directly. Existing token-major prefill and request-major decode behavior is unchanged.

sglang_omni/models/qwen3_omni/talker_model_runner.py
edwingao28: Fixed in [969b08ab](https://github.com/sgl-project/sglang-omni/commit/969b08abd79ef546cf4660435ed6821beaea49a1).

Removed the dead Talker custom forward, projected-prefill wrapper, and manual input-embeds forward. Projected rows now use only `before_prefill` plus the Omni sidecar; ordinary `forward_batch.input_embeds` remains on the upstream forward path. Updated the projected-prefill tests around that single route.

sglang_omni/scheduling/generation_batch_policy.py
edwingao28: Fixed in [5137d7c4](https://github.com/sgl-project/sglang-omni/commit/5137d7c415f4d410572ebc357bb2155491d9d9c0).

The derived prefill cap is now clamped across the reachable token limits, written back to `cuda_graph_max_bs_prefill`, and shared with Qwen3-ASR. Added regression coverage for an explicit 2048 cap constrained by `max_prefill_tokens=768`.

tests/unit_test/qwen3_omni/test_talker_prefill_graph.py
edwingao28: Fixed in [9ad4a4cf](https://github.com/sgl-project/sglang-omni/commit/9ad4a4cfbb51210e0d391290574c67c589dccfb8).

Deleted the source-text assertion and the two tests that exercised only the upstream prefill runner helper.

sglang_omni/model_runner/model_worker.py
edwingao28: Fixed in [55003611](https://github.com/sgl-project/sglang-omni/commit/550036117edf5228283acc408c5a484e2d4cdb59).

`_record_prefill_cuda_graph_usage` now reads `forward_batch.forward_mode` directly and filters non-extend and CUDA-graph modes through the declared ForwardMode contract.

sglang_omni/model_runner/model_worker.py
edwingao28: Fixed in [55003611](https://github.com/sgl-project/sglang-omni/commit/550036117edf5228283acc408c5a484e2d4cdb59).

Prefill graph info now reads the runner and server config directly and uses `isinstance(PrefillCudaGraphRunner)` before reporting capture buckets, backend type, and the input-embeds slot.

sglang_omni/model_runner/prefill_inputs.py
edwingao28: Fixed in [6e9dfa5a](https://github.com/sgl-project/sglang-omni/commit/6e9dfa5a76c6b94dd724d499d910d297db76a083).

Removed inline-code backticks from the prefill sidecar docstrings.

sglang_omni/model_runner/base.py
edwingao28: Fixed in [2572aa57](https://github.com/sgl-project/sglang-omni/commit/2572aa577ca0761bdc09493ee2c78dcece4db101).

Removed the base observer hook and its orchestration call. The Qwen thinker override now records custom eager usage only when the inherited custom prefill path returns a result, with focused base and Qwen coverage.

tests/unit_test/qwen3_omni/test_thinker_prefill_graph_gate.py
edwingao28: Fixed in [9ad4a4cf](https://github.com/sgl-project/sglang-omni/commit/9ad4a4cfbb51210e0d391290574c67c589dccfb8).

Deleted the source-inspection gate file. The existing behavioral pipeline tests continue to cover deferred graph initialization, attestation, and `enable_prefill_input_embeds` wiring.

Ratish1 (approved):

AkazaAkane (approved):

PR #1523, opened 2026-08-13
[Qwen3-TTS] Add event-based vocoder resolve and reusable pinned staging

jiannan-17:
Part of #1418 (second-wave backlog: *“Streaming vocoder blocks on `stream.synchronize()` per decode batch and re-uploads `decoder_input` from pageable memory; use event-based resolve and pinned staging.”*).

## Summary

This PR establishes the synchronization foundation for asynchronous Qwen3-TTS streaming decode.

It moves CUDA completion waits and device-to-host materialization outside `_state_lock`, replaces per-launch Event allocation with a worker-owned reusable transfer slot, and makes the lifetime of asynchronous CUDA resources explicit.

After this PR, there are two follow-ups below
#### Follow-up 1: Migrate Code2Wav to the common staging utils

Code2Wav is intentionally unchanged in this PR. This PR extract sharing utils and Code2Wav can migrate to `GrowablePinnedBuffer` and `PinnedTransferSlot` in a separate PR
#### Follow-up 2: Decoder Graph, then launch-ahead

```text
This PR: set up framework for reusable staging, and synchronization
    ↓
Follow-up vocoder CUDA Graph work (tracked in #1675)
    ↓
Re-profile follow-up scheduling
    ↓
If justified: implement launch ahead(launch batch N+1 while resolving batch N)
```

After the graph changes land, the follow-up worker will be profiled again. Launch-ahead will only be implemented if the new trace shows real improvement.

## Flow

Previously, each CUDA decode followed:

```text
build plan                         locked
→ run CUDA decode and wait         unlocked
→ acquire _state_lock
→ copy audio from GPU to CPU       locked
→ update state and emit audio      locked
```

The D2H materialization could therefore hold `_state_lock` and block code ingestion or state updates for other requests.

This PR changes the path to:

```text
build plan                                      locked
→ launch H2D + decode + D2H + completion Event unlocked
→ wait for Event and materialize CPU tensors    unlocked
→ commit state and emit audio                   locked
```

The lock is reacquired only after the CPU audio deltas are ready.

### 1. Split build, launch, resolve, and commit

The asynchronous CUDA path is now divided into four stages:

```text
build plan → launch → resolve → commit and output
locked       unlocked  unlocked       locked
```
| Phase | Entry point | Responsibilities |
| --- | --- | --- |
| **Build** | `_build_decode_plan()` | Select the codec frames for the next decode. Build `decoder_input` and record the positions needed to extract only newly generated audio. |
| **Launch** | `_launch_decode_plans()` | Combine compatible plans into a batch. When available, stage CPU codec inputs through pinned host memory. Enqueue H2D, decoder execution, delta extraction, and D2H on the worker’s CUDA stream. Copy one audio delta per request into the worker-owned pinned output slot, then record the slot’s reusable completion Event. The returned handle owns the slot until resolution and keeps the exact CUDA inputs and outputs alive. |
| **Resolve** | `_Qwen3TTSDecodeHandle.resolve()` | Wait for the slot’s completion Event. Clone the staged deltas into independent CPU tensors, clear CUDA and pinned-view keepalives, and release the slot. |
| **Commit** | `_commit_initial()` / `_commit_followup()` | Verify that the request and decode plan are still current, update decode progress, queue the audio output, and schedule the next decode when more codec frames are available. Commit and output occur only after the CPU audio is ready. |

### 2. Reusable CUDA staging

| Component | Responsibility |
| --- | --- |
| `GrowablePinnedBuffer` | Provides grow-only pinned host storage. |
| `PinnedTransferSlot` | Combines pinned output storage with a reusable CUDA completion Event and exposes `record()` / `synchronize()`. |
| TTS `_DecodeSlot` | Combines a pinned input buffer, a `PinnedTransferSlot`, and TTS-specific `busy` / `broken` ownership state. |

Each participating thread owns one thread-local `_DecodeSlot`, including the initial worker, follow-up worker, and synchronous decode callers.

- Slots are never shared across threads.
- Scheduler policy, request ordering, retries, and failure recovery remain model-specific.
- Other TTS implementations and Code2Wav may adopt the common utilities separately; Code2Wav is unchanged in this PR.

### 3. Safety and ownership
Added TTS specific failure paths handles.
## Accuracy Test

Production-code validation used `8323d3f` on one H100 80GB with PyTorch 2.11.0+cu130. Final head `2f21adf3` only adds four `accelerator` test markers.

- Changed-file pre-commit and `py_compile` checks passed.
- Local isolated tests: 32 selected Qwen3-TTS cases and 4 CUDA-staging utility tests passed.
- H100 at `8323d3f`: `test_pipeline.py` and `tests/unit_test/utils/` passed with 229 tests and no failures.
- The four CUDA lifetime, synchronization, fallback, and Event-reuse tests also passed when run directly.
- On `2f21adf3`, the affected suites passed under both CI marker expressions: 220 passed and 9 deselected in the non-accelerator selection; 9 passed in the accelerator selection.

The targeted tests cover completion waits, independent CPU outputs, reusable Events, decoder-input lifetime, pageable fallback, invalid-row isolation, failure cleanup, and the synchronous short-final path.

For output parity, base `507cfaa0` and head `8323d3f` served the same fixed-seed 48-clip streaming workload. All 48 final WAV files were byte-identical by SHA-256, and every request produced the same number of audio chunks.

## Benchmark & Profiling

This PR is performance-neutral(makes no throughput or latency regression or improvement).

Base `507cfaa0` and head `8323d3f` were compared in four sequential server sessions ordered A/B/B/A. Each session ran two 256-clip trials at C8, C16, and C32, giving four trials per revision at each concurrency.

| Concurrency | Throughput | Request latency p95 | ITL p95 |
| ---: | ---: | ---: | ---: |
| C8 | +0.3% | 1.693 → 1.694 s | 0.227 → 0.227 s |
| C16 | +0.3% | 2.419 → 2.415 s | 0.332 → 0.332 s |
| C32 | -0.1% | 3.942 → 3.957 s | 0.338 → 0.339 s |

Launch-ahead is deferred until after the follow-up vocoder CUDA Graph work in #1675, followed by another profile.
Reason: Profiling found an independent next batch ready in approximately 50–60% of launches, but the post-completion gaps attributable to `resolve()` were predominantly microsecond-scale. The trace did not show enough removable GPU idle time to justify depth-2 launch-ahead under this workload.

## Scope

This PR intentionally does not:

- Launch batch N+1 while batch N is resolving.
- Modify Code2Wav.
- Claim an end-to-end throughput or latency improvement.


sglang_omni/models/qwen3_tts/streaming_vocoder.py
```diff
@@ -66,6 +66,71 @@ class _Qwen3TTSDecodePlan:
     absolute_emitted_frames: int
     generated_frames: int
     window_start: int
+    emitted_generated_frames: int
+
+
+class _PinnedBuffer:
```
JiaxinD: qwen3_omni's code2wav_scheduler (#1567) already has a near identical pinned staging + event mechanism (`_PinnedSlot` pool, reused events). Worth a note on why this doesn't reuse or share a primitive with it; reusing events would also avoid a per-batch `torch.cuda.Event` allocation.

sglang_omni/models/qwen3_tts/streaming_vocoder.py
```diff
@@ -452,8 +519,10 @@ def decode_delta(
         plan = self._build_decode_plan(state, is_final=is_final)
         if plan is None:
             return None
-        waveform = self._run_decode_plan(plan, stream=self._decode_stream)
-        return self._commit_decode_plan(state, plan, waveform)
+        handle = self._launch_decode_plans([plan], stream=self._decode_stream)
+        deltas = handle.resolve()
+        self._raise_for_bad_rows(handle.bad_rows, 1)
```
JiaxinD: The CPU and deterministic paths raise bad rows inside `_launch_decode_plans`, but the async path defers it to each caller; a future call site that forgets this would silently emit clamped audio. Consider folding the raise into `resolve()` (store the count on the handle).

sglang_omni/models/qwen3_tts/streaming_vocoder.py
```diff
@@ -496,24 +565,19 @@ def _build_decode_plan(
...
-
     def _screen_out_of_range_codes(self, decoder_input: torch.Tensor) -> Any:
-        # Note: (Jiaxin Deng) an out-of-range id makes the codec embedding lookup
-        # raise a device-side assert, which poisons the CUDA context and kills
-        # every in-flight stream in this process; validate_chunk cannot catch it
-        # because it skips device tensors. Clamp into range so the lookup is
-        # always safe, and return the per-row verdict to read back after the
-        # decode's existing sync: rows that needed clamping are failed, never
-        # emitted, so no added synchronization buys the same protection.
+        # Note: (jiannan-17) out-of-range codes can cause a CUDA device-side
```
JiaxinD: This drops the original rationale: a device-side assert poisons the CUDA context and kills every in-flight stream, and validate_chunk can't catch it since it skips device tensors, which is why this guard exists. Please restore that part and the original attribution (only the stale "read back after the decode's existing sync" half needed updating).

tests/unit_test/qwen3_tts/test_pipeline.py
```diff
@@ -1643,11 +1644,379 @@ def __exit__(self, exc_type, exc, traceback):
...
+
+    scheduler._handle_stream_done(payload.request_id)
+
+    chunk = scheduler.outbox.get_nowait()
+    assert chunk.type == "stream"
+    assert len(chunk.data["audio_waveform"]) == 2 * 4 * 4
+    assert scheduler.outbox.get_nowait().type == "result"
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="requires CUDA")
```
JiaxinD, line 1928: These three CUDA tests carry the lifetime/sync guarantees, but if the unit test lane is CPU only they never run in CI. Can you confirm a GPU lane covers them?

JiaxinD (commented):
Solid correctness work: launch/resolve adjacency, the empty-delta fallback sync, and the keepalive all check out, and the tests are real behavior tests. Two asks besides the inline ones:
* Perf: C8 +6.4% is a single run; either add rounds, or reposition the PR as a perf-neutral structural step toward N+1 overlap (which the C16/C32 numbers already support).
* nit: comment convention in this repo is `# note (name):` with the colon after the parens.

sglang_omni/models/qwen3_tts/streaming_vocoder.py
jiannan-17: thank you @JiaxinD for pointing out this. this makes sense.
this PR was developed in parallel with the Code2Wav scheduler changes, so I wasn’t aware of that implementation. 

below are the gaps between TTS and Code2Wav and how can we reuse them
| Mechanism | Qwen3-TTS before this PR | Code2Wav | Reuse decision |
| --- | --- | --- | --- |
| Growable pinned output buffer | Not available | `_PinnedSlot.buffer` | Reusable |
| Reusable CUDA Event | Not available; TTS calls `stream.synchronize()` | Each `_PinnedSlot` owns and reuses one Event | Reusable |
| Event `record()` / `synchronize()` | No explicit completion Event | Used to fence and materialize async D2H | Reusable |
| Event `query()` | Not needed because TTS resolves immediately | Required to check pending windows without blocking | Code2Wav-only |
| Pinned H2D input | Needed by TTS because `decoder_input` may come from pageable CPU memory | Not part of the Code2Wav slot | TTS needs an additional input buffer |
| Pinned D2H output | Not available | Already implemented in `_PinnedSlot` | Reusable |
| Slot ownership | A decode group would temporarily own a worker-local slot | A request stores a slot in `state.pending` across scheduler iterations | Model-specific; cannot share directly |
| Concurrent slots | Current TTS execution is depth-1 and needs one slot per thread | Multiple requests may hold pending slots; the scheduler maintains up to 32 | Code2Wav pool cannot be directly reused by TTS |
| Launch-ahead | Not available | Launches the next window before flushing the previous pending window | Scheduler-specific |
| Pool capacity | Not needed | `_MAX_PINNED_SLOTS`, free list, and created count | Code2Wav-specific |
| Retire and reap | Not needed | Required because abort must transfer a pending slot without blocking | Code2Wav-specific |
| Quarantine | TTS needs a worker-fatal/broken-slot policy when completion cannot be proven | Code2Wav quarantines individual pool slots and may continue serving | Different failure policies; cannot share directly |
| Decode handle | Needed to retain batched GPU inputs, waveform views, pinned views, and row verdicts until resolution | Uses `_PendingWindow(slot, samples)` instead | TTS-specific |
| Invalid-row isolation | Batched TTS rows must be rejected and valid rows retried | Not part of Code2Wav staging | TTS-specific |
| Commit and request ordering | Managed by the initial/follow-up worker and decode-plan checks | Managed by per-request `state.pending` and window emission | Model-specific |


and fyi due to the gap between TTS and Code2Wav, in the next revision, i will also explore if i need to add launch-ahead batch pipelining to TTS.

to be more specific: the current TTS path does not launch batch N+1 while batch N is being resolved, whereas Code2Wav already supports this overlap. i plan to add the same batch pipelining to TTS, then i can reuse the shared pinned buffer mechanism from Code2Wav

junliu-mde (approved):
Reviewed current head `d4ef7648` against main.

I do not see a correctness blocker in the event/slot ownership changes. The handle keeps the decoder input and waveform views alive until the completion event resolves; event failure drains the exact stream before releasing memory, and an unproven stream failure retains the resources and disables further CUDA decode. Invalid codec rows are raised from `resolve()` before any clamped audio can be committed. The focused staging utility tests pass locally (`4 passed`).

This does not conflict with #1643. #1523 changes the streaming chunk path (`_launch_decode_plans()` / `resolve()` / commit), plus a new reusable staging helper. Our fixed-vocoder-layout result uses the non-streaming `_vocode_payloads()` collector, while #1643 changes the AR model runner and scheduler. These are separate paths.

One metric correction for the linked discussion: the earlier `1.52 s` number was from a complete WAV response, so it was total latency, not streaming TTFB. Streaming TTFB needs `stream: true` and first-chunk timing.

The performance claim should remain perf-neutral infrastructure unless crossed multi-round data establishes a gain; the correctness and ownership structure is useful independently.

tests/unit_test/qwen3_tts/test_pipeline.py
jiannan-17: Good catch. After the CI split, these tests fell between the two jobs: the default job hides CUDA, while the accelerator job only selects marked tests. Commit `2f21adf3` adds the `accelerator` marker to all four real-CUDA vocoder tests.

jiannan-17:
> Reviewed current head `d4ef7648` against main.
> 
> I do not see a correctness blocker in the event/slot ownership changes. The handle keeps the decoder input and waveform views alive until the completion event resolves; event failure drains the exact stream before releasing memory, and an unproven stream failure retains the resources and disables further CUDA decode. Invalid codec rows are raised from `resolve()` before any clamped audio can be committed. The focused staging utility tests pass locally (`4 passed`).
> 
> This does not conflict with #1643. #1523 changes the streaming chunk path (`_launch_decode_plans()` / `resolve()` / commit), plus a new reusable staging helper. Our fixed-vocoder-layout result uses the non-streaming `_vocode_payloads()` collector, while #1643 changes the AR model runner and scheduler. These are separate paths.
> 
> One metric correction for the linked discussion: the earlier `1.52 s` number was from a complete WAV response, so it was total latency, not streaming TTFB. Streaming TTFB needs `stream: true` and first-chunk timing.
> 
> The performance claim should remain perf-neutral infrastructure unless crossed multi-round data establishes a gain; the correctness and ownership structure is useful independently.

Thanks for the correction. I removed the complete-WAV latency reference and now report streaming TTFC separately from end-to-end request latency.

sglang_omni/models/qwen3_tts/streaming_vocoder.py
jiannan-17: **UPDATE**
I did not add N+1 launch-ahead in this revision. Profiling did not show enough removable GPU idle time to justify the additional scheduling state. 
I marked this as the follow up item in PR description. The action is re-profile after the vocoder CUDA Graph work(#1675) complete.

sglang_omni/models/qwen3_tts/streaming_vocoder.py
jiannan-17: Addressed. Invalid-row validation is now part of resolve() for both pending and pageable handles.

sglang_omni/models/qwen3_tts/streaming_vocoder.py
jiannan-17: Addressed. I restored the original rationale and updated the comment

JiaxinD (approved):

jiannan-17 (commented):
The follow up actions all have been completed.
Follow up 1: https://github.com/sgl-project/sglang-omni/pull/1759
Follow up 2: #1930 and #1997 

All related task can be closed.

PR #1525, opened 2026-08-13
[Feature] Add semantic VAD turn detection for /v1/realtime

### LINKED ISSUE #968: [Feature] Implement semantic_vad turn detection (model-based end-of-turn) for /v1/realtime
### Checklist

- [x] 1. If the issue you raised is not a feature but a question, please raise a discussion at https://github.com/sgl-project/sglang-omni/discussions/new/choose Otherwise, it will be closed.
- [x] 2. Please use English, otherwise it will be closed.

### Motivation

**What:** Implement the reserved-but-unimplemented `semantic_vad` turn-detection mode for the Realtime endpoint, so the server can decide *when the user has finished their turn* from the content of their speech, not just from a fixed silence timeout.

**Current state:** [`TurnDetectionType`](https://github.com/sgl-project/sglang-omni/blob/main/sglang_omni/serve/realtime/events.py#L26) already defines `SEMANTIC_VAD = "semantic_vad"`, but it is a no-op stub: [`TurnDetection`](https://github.com/sgl-project/sglang-omni/blob/main/sglang_omni/serve/realtime/events.py#L29) has no `eagerness` field, and [`session.py`](https://github.com/sgl-project/sglang-omni/blob/main/sglang_omni/serve/realtime/session.py#L108) builds `StreamingVAD(VADConfig())` unconditionally with no dispatch on `turn_detection.type`. A client that requests `semantic_vad` today silently gets `server_vad`.

**Why it matters:** `server_vad` ends a turn purely on `silence_duration_ms`. This is the core turn-taking UX problem in real-time voice: a short timeout cuts users off when they pause mid-thought ("I'd like to transfer… uh… $500"), while a long timeout makes the assistant feel laggy after the user is clearly done. A semantic end-of-turn (EOU) signal fixes both, and completes the `server_vad`/`semantic_vad` parity the enum already implies.

**Prior art:** OpenAI Realtime's [`semantic_vad`](https://developers.openai.com/api/docs/guides/realtime-vad) uses a semantic classifier over the user's words plus an `eagerness` knob. LiveKit Agents and Pipecat both implement it the same way in open source — **a lightweight VAD (Silero) for voice activity + a separate EOU model on top**. This proposal follows that pattern.

**Proposed design (opt-in, fail-open, builds on the existing VAD, minimal blast radius):**
1. Keep Silero (the current `server_vad` detector) **unchanged** for frame-level speech/silence detection and segmentation.
2. On a short *candidate* pause (shorter than the `server_vad` commit window), run a small **audio-native EOU model** over the current utterance to score P(user finished). Below threshold → keep listening; above → emit `speech_stopped` + commit. A hard `max_silence` fallback guarantees the turn always ends (so a mis-firing model can never hang the session).
3. Add an OpenAI-compatible `eagerness` (`low` / `medium` / `high` / `auto`) field to `TurnDetection`, mapping to the EOU threshold + pause windows.
4. Introduce a small `TurnDetector` strategy + factory in `serve/realtime/`: `server_vad` → the current [`StreamingVAD`](https://github.com/sgl-project/sglang-omni/blob/main/sglang_omni/serve/realtime/vad.py) (unchanged, still the default); `semantic_vad` → a detector that *composes* Silero + the EOU model. `session.py` selects via `turn_detection.type`.
5. The emitted events (`input_audio_buffer.speech_started` / `speech_stopped`) are **unchanged**, so nothing downstream in the commit path changes. `semantic_vad` is **opt-in** and **fails open to `server_vad`** if the model/dependency is unavailable, and its **weights load from a local path** (works in air-gapped/offline deployments).

**Recommended EOU model:** [Pipecat Smart Turn v3](https://github.com/pipecat-ai/smart-turn) — audio-native, ~8M params (8 MB int8 ONNX, CPU-only, ~10–65 ms; [weights](https://huggingface.co/pipecat-ai/smart-turn-v3)), 23 languages, BSD-2, explicitly designed to run on a VAD's silence, zero GPU cost. 

I have a working prototype of this design, validated live against Qwen3-Omni-30B on sglang-omni, Silero + Smart Turn v3 (ONNX, CPU, fail-open), wired through `session.py`'s `handle_session_update` with no change to the emitted event contract. Happy to open the PR once the approach is confirmed. Open questions for maintainers: (a) is Smart Turn v3 the preferred EOU model, or another permissively-licensed audio-native one? 

### Related resources

- Reserved stub [`TurnDetectionType.SEMANTIC_VAD`](https://github.com/sgl-project/sglang-omni/blob/main/sglang_omni/serve/realtime/events.py#L26); Realtime endpoint added in #385.
- OpenAI Realtime `semantic_vad` + `eagerness`: https://developers.openai.com/api/docs/guides/realtime-vad
- Smart Turn v3 (recommended EOU model, BSD-2): https://github.com/pipecat-ai/smart-turn — weights: https://huggingface.co/pipecat-ai/smart-turn-v3

RajPatel17:
## Motivation

Realtime voice systems need to decide when a user has finished speaking before committing the audio and starting a response. The current `/v1/realtime` endpoint uses `server_vad`, which makes that decision from a fixed silence duration. Although `semantic_vad` is already reserved in `TurnDetectionType`, it does not currently select a different detector.

Fixed-silence endpointing creates a tradeoff: a short timeout responds quickly but can cut users off during a natural thinking pause, while a long timeout avoids premature turns but adds noticeable latency after a clearly completed sentence.

This PR implements semantic VAD by building on top of the existing Silero VAD rather than replacing it. Silero continues to detect speech, silence, and utterance boundaries. After a short candidate pause, Smart Turn v3.2 scores whether the utterance sounds semantically complete. Complete thoughts can be committed quickly, while incomplete thoughts remain open so resumed speech stays in the same turn. A hard maximum pause always guarantees progress.

The feature is opt-in and preserves the existing realtime contract. `server_vad` remains the default, and both detectors emit the same speech-started and speech-stopped events, so auto-commit, generation, transcription, barge-in, cancellation, and playback truncation continue to work unchanged.

For self-hosted and offline deployments, Smart Turn is loaded from a local path, verified against a pinned SHA-256, shared across sessions, and never downloaded at runtime. If it is unavailable or invalid, semantic requests safely fall back to `server_vad` instead of making the realtime endpoint unavailable.

The ONNX model is loaded directly instead of adding a `pipecat-ai` dependency. Current Pipecat PyPI releases require Python 3.11 or newer while SGLang-Omni supports Python 3.10, and the full Pipecat framework would add unrelated voice-agent dependencies for a wrapper that ultimately loads the same ONNX artifact. Direct integration preserves Python 3.10 compatibility and keeps the runtime dependency closure small.

## Modifications

- Add a turn-detector factory that selects the existing `StreamingVAD` for `server_vad` or Silero plus Smart Turn for `semantic_vad`.
- Add `low`, `medium`, and `high` semantic eagerness presets with confidence-based commit windows and hard maximum pauses.
- Add a checksum-verified local Smart Turn v3.2 ONNX loader using normalized, fixed-length Whisper preprocessing and CPU inference.
- Preload one Smart Turn model per server process and share it across realtime sessions.
- Advertise accepted turn detectors in session capabilities and report the effective detector in `session.updated`.
- Apply detector updates transactionally while preserving partial-update behavior and `interrupt_response`; failed construction leaves live state and buffered audio unchanged.
- Reject `silence_duration_ms` for semantic VAD because its pause timing is controlled by eagerness.
- Update the playground with capability-aware detector and eagerness controls while preserving response-aware playback and barge-in behavior.
- Add documentation and focused coverage for detector timing, preprocessing, fallback, session updates, capabilities, and barge-in interactions.

## Related Issues

Closes #968

## Accuracy Test

This PR does not change Qwen3-Omni model architecture, kernels, or generation accuracy. The checksum-pinned Smart Turn model was exercised against real audio and validated to return finite probabilities within `[0, 1]`; invalid output triggers the defined fixed-silence fallback.

## Benchmark & Profiling

Smart Turn runs on CPU once per candidate pause and reuses one process-wide ONNX session, so it does not consume the GPUs used by Qwen3-Omni or change generation throughput.

Semantic scoring begins after 160 ms of silence. With the default `medium` eagerness, high-confidence completion commits after 250 ms, normal completion after 640 ms, and low confidence uses a 2-second hard stop. `low` and `high` provide more patient and more responsive policies respectively.

## Demo

Demo comparing fixed-silence server VAD with semantic VAD in text-only and text-plus-audio sessions: https://drive.google.com/file/d/12KZ5rRI80VF7zObkEK4dvD4IH__PYb-j/view?usp=sharing

## Testing Done

- [x] Local code review completed
- [x] 43 semantic VAD and adjacent realtime Python tests pass.
- [x] Browser JavaScript syntax and response-aware playback tests pass.
- [x] Ruff checks, formatting checks, Python compilation, and diff checks pass.
- [x] Live text-only and text-plus-audio semantic VAD flows validated on two H100s.

## Checklist

- [x] Add unit tests.

## CI


sglang_omni/serve/realtime/smart_turn.py
```diff
@@ -0,0 +1,108 @@
...
+from typing import Any
+
+import numpy as np
+from transformers import WhisperFeatureExtractor
+
+from .semantic_vad import SemanticEOUModel
+
+SMART_TURN_MODEL_ENV = "SGLANG_OMNI_SMART_TURN_MODEL_PATH"
+SMART_TURN_MODEL_FILENAME = "smart-turn-v3.2-cpu.onnx"
+SMART_TURN_MODEL_SHA256 = (
```
PopSoda2002, line 19: Hi can I know the magic number here?

PopSoda2002 (commented):
Is this svad model running on cpu or gpu?

sglang_omni/serve/realtime/smart_turn.py
RajPatel17: The semantic VAD runs on CPU. Smart Turn v3.2 uses ONNX Runtime’s CPU provider. 
The hex value is the SHA-256 checksum of the model artifact from this pinned Smart Turn commit (https://github.com/pipecat-ai/pipecat/commit/d1bedef5b39e7f4d1f18f35207f1fdaa90368b45). Startup verifies it to reject a changed or corrupted model file.

sglang_omni/serve/realtime/events.py
```diff
@@ -26,11 +26,18 @@ class TurnDetectionType(str, Enum):
     SEMANTIC_VAD = "semantic_vad"
 
 
+class SemanticVADEagerness(str, Enum):
+    LOW = "low"
+    MEDIUM = "medium"
+    HIGH = "high"
+
+
```
JingwenGu0829, line 35: `SemanticVADEagerness` is a closed enum (`low`/`medium`/`high`). Any other value — notably `auto`, the OpenAI Realtime default for `semantic_vad.eagerness` — raises `pydantic.ValidationError` inside `parse_client_event`, before `handle_session_update`'s error boundary. 

Consider mapping auto to medium?

sglang_omni/serve/realtime/turn_detector.py
```diff
@@ -0,0 +1,72 @@
...
+        return TurnDetectorBuild(detector, effective)
+
+    server_config = VADConfig(
+        threshold=_optional_float(config.get("threshold"), VADConfig.threshold),
+        prefix_padding_ms=_optional_int(
+            config.get("prefix_padding_ms"), VADConfig.prefix_padding_ms
+        ),
+        silence_duration_ms=_optional_int(
+            config.get("silence_duration_ms"), VADConfig.silence_duration_ms
+        ),
```
JingwenGu0829, line 70: This PR starts applying threshold, prefix_padding_ms, and silence_duration_ms to the active detector, but accepts them without range or finiteness checks. Invalid values can put the VAD state machine into nonsensical states: a negative threshold classifies every frame as speech, a negative silence duration ends speech on the first non-speech frame, and negative prefix padding moves the reported start forward and can trim the beginning of the utterance.

Maybe we should add safegaurd for invalid values?

Note that OpenAI documents the threshold range as [0, 1]; if this implementation intentionally rejects the degenerate endpoints, that compatibility difference should be explicit.

docs/basic_usage/qwen3_omni.md
```diff
@@ -361,26 +361,57 @@ output:
...
+`smart-turn-v3.2-cpu.onnx` model before startup and set
+`SGLANG_OMNI_SMART_TURN_MODEL_PATH` to the model file or its containing
+directory. The runtime does not download model weights and verifies SHA-256
+`2bb026316b14a660486a75b1733cd3fbab8c2fd0314dc9af7be49f8cca967e4f`.
+Realtime startup loads one shared CPU ONNX instance. If the model is missing
+or invalid, the endpoint stays available and semantic requests report
+`server_vad` as their effective turn detector.
+
+Text arrives in `response.text.delta` events; spoken output arrives as
+base64-encoded mono 24 kHz PCM16 in `response.audio.delta` events, followed by
```
JingwenGu0829: Can we trim and humanize the phrasing here? We are trying not to bloat up the doc.

sglang_omni/serve/realtime/turn_detector.py
```diff
@@ -0,0 +1,72 @@
...
+    def reset(self) -> None: ...
+
+
+@dataclass(frozen=True)
+class TurnDetectorBuild:
+    detector: TurnDetector
+    effective_config: dict[str, Any]
+
+
+def build_turn_detector(
```
JingwenGu0829, line 25: The detector-change check compares the serialized session configuration rather than the settings actually used by the active detector. For example, the initial server_vad is stored as {"type": "server_vad"} but runs with silence_duration_ms=500. 

A client update that explicitly restates that default therefore compares unequal, rebuilds an equivalent detector, and clears any pending input audio.


Please resolve both the current and requested configurations against the runtime defaults before deciding whether detector behavior changed. Construct a replacement only when those resolved configurations differ. This also avoids loading and immediately discarding a new Silero session when a semantic_vad request falls back to the already-active server_vad.

sglang_omni/serve/realtime/session.py
```diff
@@ -90,16 +96,24 @@ def __init__(
...
             model=model_name,
+            capabilities={
+                "turn_detection": [
+                    TurnDetectionType.SERVER_VAD.value,
+                    TurnDetectionType.SEMANTIC_VAD.value,
+                ]
+            },
             modalities=["text"],
             instructions=DEFAULT_INSTRUCTIONS,
             input_audio_format="pcm16",
```
JingwenGu0829, line 118: Only put semantic_vad in session.created.capabilities when smart_turn_model is available; otherwise clients are told it is supported only to be downgraded to server_vad. (Now the path says sematic vad even when what we got is turn vad).

sglang_omni/serve/realtime/semantic_vad.py
```diff
@@ -0,0 +1,214 @@
...
+                and silence_ms >= self.config.candidate_pause_ms
+            ):
+                try:
+                    self.candidate_probability = self._predict_eou()
+                except Exception as exc:
+                    warnings.warn(
+                        f"Smart Turn inference failed; using fixed-silence "
+                        f"fallback: {exc}",
+                        RuntimeWarning,
+                        stacklevel=2,
```
JingwenGu0829: Please log this fallback with `logger.warning(..., exc_info=True)` so Smart Turn inference failures retain their traceback and are visible in production logs

JingwenGu0829 (commented):
Thanks for the work! A couple of nit suggestions maybe you can consider fixing:

JingwenGu0829 (commented):
A more important problem is the choice to run the VAD model on CPU actually.

Could you report p50/p95/p99 turn-decision latency and host CPU utilization/core count at representative concurrency (e.g. 1/8/32 sessions), including while Qwen3-Omni is generating, since the current PR does not substantiate the claim that this shared CPU inference path has no throughput impact?

Additioanlly is there any reason you choose CPU inferenc without comparing the upstream 8 MB INT8 CPU model against its 32 MB FP32 GPU model? Since CPU may be the scarcer resource in inference task like this and the talker GPU is often idle during a single session’s input phase?

Anyways let's let the data talk which one is better. I would suggest you to actually profile the CPU burden on multiple sessions to make sure this CPU-based vad model is not causing anything harmful for real-time sessions.

sglang_omni/serve/realtime/events.py
RajPatel17: Resolved in 8fd9f8b: added AUTO to SemanticVADEagerness so it parses instead of raising before the handler's error boundary, and build_turn_detector/_detector_config map "auto" to "medium".

sglang_omni/serve/realtime/turn_detector.py
RajPatel17: Resolved in 8fd9f8b: _optional_float/_optional_int now take minimum/maximum and raise on out-of-range or non-finite values; threshold is bounded to [0, 1] per the OpenAI docs, prefix_padding_ms/silence_duration_ms to >= 0.

sglang_omni/serve/realtime/turn_detector.py
RajPatel17: Resolved in 8fd9f8b:  _detector_config now resolves threshold/prefix_padding_ms/silence_duration_ms/eagerness against the same VADConfig()/SemanticVADConfig.from_eagerness() defaults build_turn_detector uses, before comparing, so restating the active default no longer triggers a rebuild/buffer clear.

sglang_omni/serve/realtime/session.py
RajPatel17: Resolved in 8fd9f8b:  capabilities now only include semantic_vad when smart_turn_model is not None.

rsp224:
I tested on 2xH200 and got the following results:

**Idle**

| Concurrency | Model | p50 | p95 | p99 | Host CPU % (of 128 cores) |
|---|---|---|---|---|---|
| 1 | CPU (INT8) | 0.041s | 0.081s | 0.081s | 1.3% |
| 1 | GPU (FP32) | 0.559s | 1.174s | 1.174s | 1.6% |
| 8 | CPU (INT8) | 0.635s | 2.180s | 3.022s | 6.4% |
| 8 | GPU (FP32) | 0.625s | 1.978s | 2.798s | 6.9% |
| 32 | CPU (INT8) | 0.748s | 3.034s | 3.064s | 11.5% |
| 32 | GPU (FP32) | 1.218s | 3.867s | 4.094s | 14.2% |

**While Qwen3-Omni is generating (4 concurrent generating sessions)**

| Concurrency | Model | p50 | p95 | p99 | Host CPU % (of 128 cores) |
|---|---|---|---|---|---|
| 1 | CPU (INT8) | 0.483s | 2.825s | 2.825s | 6.1% |
| 1 | GPU (FP32) | 0.867s | 1.418s | 1.418s | 5.7% |
| 8 | CPU (INT8) | 0.969s | 2.239s | 2.261s | 9.8% |
| 8 | GPU (FP32) | 0.839s | 3.128s | 4.821s | 10.0% |
| 16 | CPU (INT8) | 0.803s | 3.487s | 4.702s | 13.6% |
| 16 | GPU (FP32) | 0.820s | 2.841s | 4.416s | 13.6% |

Switched to the FP32 GPU model in the latest commit.

rsp224:
I tested on 2xH200 and got the following results:

**Idle**

| Concurrency | Model | p50 | p95 | p99 | Host CPU % (of 128 cores) |
|---|---|---|---|---|---|
| 1 | CPU (INT8) | 0.500s | 2.251s | 2.251s | 2.3% |
| 1 | GPU (FP32) | 0.539s | 1.291s | 1.291s | 2.5% |
| 8 | CPU (INT8) | 0.635s | 2.180s | 3.022s | 6.4% |
| 8 | GPU (FP32) | 0.625s | 1.978s | 2.798s | 6.9% |
| 32 | CPU (INT8) | 0.748s | 3.034s | 3.064s | 11.5% |
| 32 | GPU (FP32) | 1.218s | 3.867s | 4.094s | 14.2% |

**While Qwen3-Omni is generating (4 concurrent generating sessions)**

| Concurrency | Model | p50 | p95 | p99 | Host CPU % (of 128 cores) |
|---|---|---|---|---|---|
| 1 | CPU (INT8) | 0.483s | 2.825s | 2.825s | 6.1% |
| 1 | GPU (FP32) | 0.867s | 1.418s | 1.418s | 5.7% |
| 16 | CPU (INT8) | 0.803s | 3.487s | 4.702s | 13.6% |
| 16 | GPU (FP32) | 0.820s | 2.841s | 4.416s | 13.6% |

Switched to the FP32 GPU model in the latest commit.

RajPatel17:
I tested on 2xH200 and got the following results:

**Idle**

| Concurrency | Model | p50 | p95 | p99 | Host CPU % (of 128 cores) |
|---|---|---|---|---|---|
| 1 | CPU (INT8) | 0.500s | 2.251s | 2.251s | 2.3% |
| 1 | GPU (FP32) | 0.539s | 1.291s | 1.291s | 2.5% |
| 8 | CPU (INT8) | 0.635s | 2.180s | 3.022s | 6.4% |
| 8 | GPU (FP32) | 0.625s | 1.978s | 2.798s | 6.9% |
| 32 | CPU (INT8) | 0.748s | 3.034s | 3.064s | 11.5% |
| 32 | GPU (FP32) | 1.218s | 3.867s | 4.094s | 14.2% |

**While Qwen3-Omni is generating (4 concurrent generating sessions)**

| Concurrency | Model | p50 | p95 | p99 | Host CPU % (of 128 cores) |
|---|---|---|---|---|---|
| 1 | CPU (INT8) | 0.483s | 2.825s | 2.825s | 6.1% |
| 1 | GPU (FP32) | 0.867s | 1.418s | 1.418s | 5.7% |
| 16 | CPU (INT8) | 0.803s | 3.487s | 4.702s | 13.6% |
| 16 | GPU (FP32) | 0.820s | 2.841s | 4.416s | 13.6% |

Switched to the FP32 GPU model in the latest commit.

JingwenGu0829 (approved):
LGTM

PR #1398, opened 2026-08-08
[fun-asr] enable breakable prefill CUDA graphs

AkazaAkane:
## Motivation

Fun-ASR already reaches SGLang's upstream multimodal prefill graph path without model-forward or Omni runner changes, but the generic compatibility gate rejects it and the builder does not opt in. The validation in https://github.com/sgl-project/sglang-omni/issues/1357 found exact output parity and repeatable throughput improvements for the breakable backend.

Prefills above the 48-token validated capture budget continue to run eagerly.

## Modifications

- declare Fun-ASR compatible with the breakable prefill CUDA graph contract introduced by https://github.com/sgl-project/sglang-omni/pull/1364
- enable the breakable backend by default with the validated with token bucket help n = 256.
- preserve `disable_cuda_graph=True` as the global rollback and cover the policy in focused tests

## Related Issues
based on #1364 

## Accuracy Test

pre-commit and unit test passed.

## EN Benchmark Results

*Measured across 3 repeats after discarded warmups.*

| Concurrency | Config | Throughput (req/s) | Latency mean / p95 / p99 (s) | RTF mean / p95 | WER | Failed |
| :---: | :--- | :---: | :---: | :---: | :---: | :---: |
| **1** | Baseline | 23.480 | 0.0425 / 0.0522 / 0.0614 | 0.00921 / 0.01187 | 0.017100 | 0 |
| **1** | Candidate | 27.369 | 0.0365 / 0.0461 / 0.0535 | 0.00789 / 0.01030 | 0.017100 | 0 |
| **16** | Baseline | 133.110 | 0.1198 / 0.1577 / 0.1881 | 0.02597 / 0.03627 | 0.017100 | 0 |
| **16** | Candidate | 180.890 | 0.0880 / 0.1158 / 0.1364 | 0.01908 / 0.02640 | 0.017100 | 0 |
| **32** | Baseline | 156.356 | 0.2035 / 0.2946 / 0.3706 | 0.04414 / 0.06453 | 0.017100 | 0 |
| **32** | Candidate | 203.817 | 0.1559 / 0.2260 / 0.2947 | 0.03376 / 0.04743 | 0.017100 | 0 |

### Key Performance Delta
- Candidate throughput improved by **16.6%** (c1), **35.9%** (c16), and **30.4%** (c32).
- Baseline repeat spread was **4.4%**, **6.3%**, and **7.1%**; improvements comfortably exceed eager A/A variance.
- No `concurrency-32` performance regression observed.

---

[...]

PR #1530, opened 2026-08-14
[Perf][Whisper ASR] Enable breakable prefill CUDA graphs

wirybeaver:
## Stack
- Base Draft PR: #1634 isolates the Whisper encoder/decoder cache execution-path change.
- This branch is rebased on `whisper-encoder-decoder-cache` so the base and CUDA Graph changes can be benchmarked independently.
- The GitHub base remains `main` because the stacked branches live in the contributor fork; the displayed diff will shrink after #1634 merges.

## Summary
- enable breakable prefill CUDA graphs by default for Whisper-ASR with a 256-token capture ladder
- preserve Whisper encoder metadata during prefill graph capture and replay with a model-specific runner adapter
- register decoder self-attention and cross-attention layers for graph capture

This follows the model-adoption patterns from #1398 and #1458.

## Test plan
- `pytest -q tests/unit_test/model_runner/test_cuda_graph_batch_validator.py tests/unit_test/scheduling/test_prefill_graph_policy.py tests/unit_test/qwen3_asr/test_pipeline.py tests/unit_test/whisper_asr -m 'not benchmark'` (159 passed)
- `python3 -m py_compile` for all changed Python files

## Benchmark plan
Use `openai/whisper-large-v3` and run the complete A/B/C comparison under both pre-LM encoder modes:

| Case | `enable_pre_lm_encoder` | Execution path |
|---|---:|---|
| A1 | `false` | current `main` pre-adopter eager path |
| B1 | `false` | #1634 adopter path with prefill BCG disabled |
| C1 | `false` | this PR with prefill BCG enabled |
| A2 | `true` (current `main` default) | current `main` pre-adopter eager path |
| B2 | `true` | #1634 adopter path with prefill BCG disabled |
| C2 | `true` | this PR with prefill BCG enabled |

Within each pre-LM mode, compare A versus B to detect regressions from the encoder/decoder execution-path change, then B versus C to isolate the CUDA Graph impact. Report throughput, latency, WER, capture time and memory, graph replay versus eager fallback coverage, and the corresponding boundaries.

Use both matrices to decide the default policy: enable prefill CUDA Graphs unconditionally only if they qualify without regression in both modes; otherwise gate the default on `enable_pre_lm_encoder` for the mode that qualifies. These qualification results are pending.

## Related Issues
- #1357
- #1396

## Historical whisper-base benchmark (pre-restack)

The following result was contributed by @ZemingL before the current-main restack. It is useful historical data, but it does not replace the planned `whisper-large-v3` A/B/C qualification above.

Runtime fixes and benchmarking by @ZemingL Thanks!
### Setup

- Baseline: `upstream/main@68abc7eec59ff7b0dce484cb501ffbbb338f9e46`
- Candidate used for performance measurement: `c7dee6dcb356803e96501d94fdc7be6ae8b956d5`
- Hardware: NVIDIA H100 80GB HBM3, driver 580.95.05
- CI image: `hongccc/sglang-omni@sha256:374d0b1c30b2bff685b1716fc64a02ad3b3d0a90fe2ce73ce9861a6992c28101`
- Model: `openai/whisper-base@e37978b90ca9030d5170a5c07aadb050351a65bb`, FP16
- Dataset: full SeedTTS EN split, 1,088 samples
- Concurrency: 1, 2, 4, 8, 16, 32
- Protocol: one discarded application warmup plus three measured repeats per concurrency; same host, image, cache, and command
- Full run order: candidate -> baseline. A post-baseline candidate spot run covered concurrency 16 and 32. The preceding 20-sample smoke used baseline -> candidate order.

The first GPU smoke exposed two candidate startup failures: BCG attention returned head-shaped tensors before Whisper's output projection, and the generic layer discovery omitted cross-attention layers. Commit `c7dee6dc` fixes both. After the fix, capture completed and the runtime attested all 22 buckets from 4 through 256 tokens.

| Concurrency | Main req/s | BCG req/s | Throughput | Main mean / p95 / p99 latency (ms) | BCG mean / p95 / p99 latency (ms) | Main / BCG WER |
| 1 | 22.169 | 24.255 | +9.41% | 45.0 / 53.5 / 60.2 | 41.1 / 50.4 / 55.5 | 0.048275 / 0.048275 |
| 2 | 36.166 | 36.623 | +1.26% | 55.2 / 65.6 / 72.4 | 54.5 / 64.5 / 69.5 | 0.047972 / 0.048073 |
| 4 | 58.463 | 63.042 | +7.83% | 68.0 / 80.4 / 87.5 | 63.1 / 74.6 / 82.7 | 0.047972 / 0.047972 |
| 8 | 88.840 | 97.290 | +9.51% | 89.7 / 104.9 / 115.3 | 81.9 / 97.4 / 110.4 | 0.047972 / 0.048023 |
| 16 | 103.997 | 117.313 | +12.80% | 152.9 / 175.2 / 233.3 | 135.6 / 159.1 / 207.0 | 0.047972 / 0.048123 |
| 32 | 98.679 | 114.718 | +16.25% | 321.9 / 384.1 / 429.1 | 276.8 / 333.1 / 381.4 | 0.047972 / 0.048123 |

### Correctness and mechanism checks

- 19,584/19,584 measured requests succeeded per arm; no skipped or failed samples.
- At concurrency 1, all 3,264 returned transcripts matched exactly between main and BCG.
- Concurrent runs showed small repeat-to-repeat transcript variation in both arms; corpus WER stayed within 0.0480-0.0483.
- Prefill BCG capture: 1.69 seconds, 0.09 GB reported memory, buckets `[4, 8, 12, 16, 20, 24, 28, 32, 48, 64, 80, 96, 112, 128, 144, 160, 176, 192, 208, 224, 240, 256]`.
- Post-baseline candidate spot run: 115.754 req/s at concurrency 16 and 116.868 req/s at concurrency 32, within -1.33% and +1.87% of the primary candidate run.
- Focused CPU tests after the fixes: 136 passed.

### DIFF
--- a/sglang_omni/model_runner/sglang_model_runner.py
+++ b/sglang_omni/model_runner/sglang_model_runner.py
@@ -3,7 +3,9 @@ from __future__ import annotations
 import dataclasses
 import logging
 from collections.abc import Iterator
+from contextlib import contextmanager
 from dataclasses import dataclass
+from threading import Lock
 from typing import Any
 
 from sglang.srt.configs.model_config import ModelConfig
@@ -23,6 +25,7 @@ from sglang_omni.utils.gpu_memory import (
 )
 
 logger = logging.getLogger(__name__)
+_PREFILL_RUNNER_OVERRIDE_LOCK = Lock()
 
 
 def filter_weights_by_prefix(
@@ -440,7 +443,8 @@ class SGLModelRunner(ModelRunner):
         from sglang.srt.runtime_context import get_exec, get_flags
 
         get_flags().capture.enable_torch_compile = get_exec().graph.enable_torch_compile
-        result = super().init_cuda_graphs(capture_decode_cuda_graph)
+        with self._prefill_cuda_graph_runner_override():
+            result = super().init_cuda_graphs(capture_decode_cuda_graph)
         if self.token_to_kv_pool.post_capture_active:
             self.post_capture_resize_kv_pool()
         return result
@@ -472,6 +476,40 @@ class SGLModelRunner(ModelRunner):
                 "below max_running_requests that raises the reserved headroom."
             )
         return result
+
+    def _prefill_cuda_graph_runner_cls(self):
+        from sglang.srt.model_executor.cuda_graph_config import (
+            Backend as CudaGraphBackend,
+        )
+
+        if (
+            self._model_arch_override == "WhisperForConditionalGeneration"
+            and self.server_args.cuda_graph_config.prefill.backend
+            == CudaGraphBackend.BREAKABLE
+        ):
+            from sglang_omni.model_runner.whisper_prefill_cuda_graph_runner import (
+                WhisperPrefillCudaGraphRunner,
+            )
+
+            return WhisperPrefillCudaGraphRunner
+        return None
+
+    @contextmanager
+    def _prefill_cuda_graph_runner_override(self):
+        runner_cls = self._prefill_cuda_graph_runner_cls()
+        if runner_cls is None:
+            yield
+            return
+
+        from sglang.srt.model_executor.model_runner_components import cuda_graph_setup
+
+        with _PREFILL_RUNNER_OVERRIDE_LOCK:
+            original = cuda_graph_setup.PrefillCudaGraphRunner
+            cuda_graph_setup.PrefillCudaGraphRunner = runner_cls
+            try:
+                yield
+            finally:
+                cuda_graph_setup.PrefillCudaGraphRunner = original
 
     def _weight_update_blocked_reason(self) -> str | None:
         ws = self._weight_share_config
--- a/sglang_omni/model_runner/sglang_model_runner.py
+++ b/sglang_omni/model_runner/sglang_model_runner.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import dataclasses
 import logging
 from collections.abc import Iterator
-from contextlib import contextmanager
 from dataclasses import dataclass
 from threading import Lock
 from typing import Any
@@ -25,7 +24,27 @@ from sglang_omni.utils.gpu_memory import (
 )
 
 logger = logging.getLogger(__name__)
-_PREFILL_RUNNER_OVERRIDE_LOCK = Lock()
+_PREFILL_RUNNER_DISPATCH_LOCK = Lock()
+_PREFILL_RUNNER_DISPATCH_DEFAULT: type | None = None
+
+
+def _install_prefill_runner_dispatch() -> None:
+    """Let each model runner pick its own prefill graph runner class."""
+    global _PREFILL_RUNNER_DISPATCH_DEFAULT
+    from sglang.srt.model_executor.model_runner_components import cuda_graph_setup
+
+    with _PREFILL_RUNNER_DISPATCH_LOCK:
+        if _PREFILL_RUNNER_DISPATCH_DEFAULT is not None:
+            return
+        default_cls = cuda_graph_setup.PrefillCudaGraphRunner
+
+        def _dispatch_prefill_runner(model_runner):
+            select = getattr(model_runner, "_prefill_cuda_graph_runner_cls", None)
+            runner_cls = select() if select is not None else None
+            return (runner_cls or default_cls)(model_runner)
+
+        cuda_graph_setup.PrefillCudaGraphRunner = _dispatch_prefill_runner
+        _PREFILL_RUNNER_DISPATCH_DEFAULT = default_cls
 
 
 def filter_weights_by_prefix(
@@ -443,8 +462,8 @@ class SGLModelRunner(ModelRunner):
         from sglang.srt.runtime_context import get_exec, get_flags
 
         get_flags().capture.enable_torch_compile = get_exec().graph.enable_torch_compile
-        with self._prefill_cuda_graph_runner_override():
-            result = super().init_cuda_graphs(capture_decode_cuda_graph)
+        _install_prefill_runner_dispatch()
+        result = super().init_cuda_graphs(capture_decode_cuda_graph)
         if self.token_to_kv_pool.post_capture_active:
             self.post_capture_resize_kv_pool()
         return result
@@ -493,23 +512,6 @@ class SGLModelRunner(ModelRunner):
 
             return WhisperPrefillCudaGraphRunner
         return None
-
-    @contextmanager
-    def _prefill_cuda_graph_runner_override(self):
-        runner_cls = self._prefill_cuda_graph_runner_cls()
-        if runner_cls is None:
-            yield
-            return
-
-        from sglang.srt.model_executor.model_runner_components import cuda_graph_setup
-
-        with _PREFILL_RUNNER_OVERRIDE_LOCK:
-            original = cuda_graph_setup.PrefillCudaGraphRunner
-            cuda_graph_setup.PrefillCudaGraphRunner = runner_cls
-            try:
-                yield
-            finally:
-                cuda_graph_setup.PrefillCudaGraphRunner = original
 
     def _weight_update_blocked_reason(self) -> str | None:
         ws = self._weight_share_config
--- a/sglang_omni/models/whisper_asr/engine_builder.py
+++ b/sglang_omni/models/whisper_asr/engine_builder.py
@@ -242,15 +242,10 @@ class WhisperASREngineBuilder(AsrEngineBuilder):
         # Note (Akazaakane): Timestamped Whisper requests install an internal
         # per-request processor; this flag permits SGLang to execute it.
         overrides["enable_custom_logit_processor"] = True
-        if overrides.get("cuda_graph_backend_prefill") == CudaGraphBackend.DISABLED:
-            overrides.pop("cuda_graph_bs_prefill", None)
-            overrides.pop("cuda_graph_max_bs_prefill", None)
-            return
-        if "cuda_graph_bs_prefill" in overrides:
-            overrides.setdefault(
-                "cuda_graph_max_bs_prefill",
-                max(int(size) for size in overrides["cuda_graph_bs_prefill"]),
-            )
+        if (
+            overrides.get("cuda_graph_backend_prefill") == CudaGraphBackend.DISABLED
+            or "cuda_graph_bs_prefill" in overrides
+        ):
             return
 
         cap = min(
--- a/sglang_omni/models/whisper_asr/engine_builder.py
+++ b/sglang_omni/models/whisper_asr/engine_builder.py
@@ -11,10 +11,15 @@ from sglang_omni.models.whisper_asr.encoder_service import (
     build_cache_namespace,
 )
 from sglang_omni.scheduling.engine_factory import AsrEngineBuilder
+from sglang_omni.scheduling.generation_batch_policy import (
+    CudaGraphBackend,
+    build_default_prefill_cuda_graph_bs,
+)
 
 logger = logging.getLogger(__name__)
 
 _DEFAULT_ENCODER_GRAPH_BATCH_BUCKETS = (1, 2, 4, 8, 12, 16)
+_WHISPER_PREFILL_GRAPH_MAX_TOKENS = 256
 
 
 def _normalize_encoder_graph_buckets(buckets: list[int] | None) -> tuple[int, ...]:
@@ -62,6 +67,7 @@ def _resolve_encoder_graph_buckets(
 class WhisperASREngineBuilder(AsrEngineBuilder):
     model_name = "Whisper ASR"
     model_arch_override = "WhisperForConditionalGeneration"
+    supports_breakable_prefill_cuda_graph = True
 
     def __init__(
         self,
@@ -236,6 +242,30 @@ class WhisperASREngineBuilder(AsrEngineBuilder):
         # Note (Akazaakane): Timestamped Whisper requests install an internal
         # per-request processor; this flag permits SGLang to execute it.
         overrides["enable_custom_logit_processor"] = True
+        if overrides.get("cuda_graph_backend_prefill") == CudaGraphBackend.DISABLED:
+            overrides.pop("cuda_graph_bs_prefill", None)
+            overrides.pop("cuda_graph_max_bs_prefill", None)
+            return
+        if "cuda_graph_bs_prefill" in overrides:
+            overrides.setdefault(
+                "cuda_graph_max_bs_prefill",
+                max(int(size) for size in overrides["cuda_graph_bs_prefill"]),
+            )
+            return
+
+        cap = min(
+            int(value)
+            for value in (
+                _WHISPER_PREFILL_GRAPH_MAX_TOKENS,
+                overrides.get("max_prefill_tokens"),
+                overrides.get("cuda_graph_max_bs_prefill"),
+                overrides.get("max_total_tokens"),
+            )
+            if value is not None
+        )
+        ladder = build_default_prefill_cuda_graph_bs(cap)
+        overrides["cuda_graph_bs_prefill"] = ladder
+        overrides["cuda_graph_max_bs_prefill"] = max(ladder)
 
     def generation_defaults(self, *, dtype: str) -> dict[str, Any]:
         return {
@@ -248,6 +278,7 @@ class WhisperASREngineBuilder(AsrEngineBuilder):
             "chunked_prefill_size": 0,
             "sampling_backend": "pytorch",
             "dtype": dtype,
+            "cuda_graph_backend_prefill": CudaGraphBackend.BREAKABLE,
         }
 
     def make_adapters(self, model: Any) -> tuple[Any, Any]:
--- a/sglang_omni/models/whisper_asr/engine_builder.py
+++ b/sglang_omni/models/whisper_asr/engine_builder.py
@@ -19,7 +19,36 @@ from sglang_omni.scheduling.generation_batch_policy import (
 logger = logging.getLogger(__name__)
 
 _DEFAULT_ENCODER_GRAPH_BATCH_BUCKETS = (1, 2, 4, 8, 12, 16)
-_WHISPER_PREFILL_GRAPH_MAX_TOKENS = 256
+
+
+def _clamp_prefill_cuda_graph_max_bs(
+    overrides: dict[str, Any], *, context_length: int | None
+) -> int | None:
+    """Clamp the prefill graph budget to the reachable token limits.
+
+    Whisper admits its encoder prefix atomically (``chunked_prefill_size=0``),
+    so the shared policy has no chunk to derive a cap from; bound the ladder by
+    the effective prefill, total-token, and model-context limits instead.
+    Returns ``None`` when no limit is known yet (before model setup), leaving
+    the ladder to SGLang's own derivation.
+    """
+    if context_length is None or int(context_length) <= 0:
+        context_length = overrides.get("context_length")
+    caps = [
+        int(value)
+        for value in (
+            overrides.get("cuda_graph_max_bs_prefill"),
+            overrides.get("max_prefill_tokens"),
+            overrides.get("max_total_tokens"),
+            context_length,
+        )
+        if value is not None and int(value) > 0
+    ]
+    if not caps:
+        return None
+    cap = min(caps)
+    overrides["cuda_graph_max_bs_prefill"] = cap
+    return cap
 
 
 def _normalize_encoder_graph_buckets(buckets: list[int] | None) -> tuple[int, ...]:
@@ -248,19 +277,12 @@ class WhisperASREngineBuilder(AsrEngineBuilder):
         ):
             return
 
-        cap = min(
-            int(value)
-            for value in (
-                _WHISPER_PREFILL_GRAPH_MAX_TOKENS,
-                overrides.get("max_prefill_tokens"),
-                overrides.get("cuda_graph_max_bs_prefill"),
-                overrides.get("max_total_tokens"),
-            )
-            if value is not None
+        cap = _clamp_prefill_cuda_graph_max_bs(
+            overrides, context_length=self.context_length
         )
-        ladder = build_default_prefill_cuda_graph_bs(cap)
-        overrides["cuda_graph_bs_prefill"] = ladder
-        overrides["cuda_graph_max_bs_prefill"] = max(ladder)
+        if cap is None:
+            return
+        overrides["cuda_graph_bs_prefill"] = build_default_prefill_cuda_graph_bs(cap)
 
     def generation_defaults(self, *, dtype: str) -> dict[str, Any]:
         return {
--- a/sglang_omni/models/whisper_asr/engine_builder.py
+++ b/sglang_omni/models/whisper_asr/engine_builder.py
@@ -42,11 +42,23 @@ def _reachable_prefill_cuda_graph_max_bs(
     max_prefill_tokens = overrides.get("max_prefill_tokens")
     if encoder_token_count < 1 or not max_prefill_tokens or int(max_prefill_tokens) < 1:
         return None
-    per_request = encoder_token_count + _DECODER_PREFILL_TOKENS_PER_REQUEST
-    batch_requests = max(1, int(max_prefill_tokens) // per_request)
+    budget = int(max_prefill_tokens)
+    request_limit = max(1, budget // encoder_token_count)
     if max_running_requests is not None and int(max_running_requests) >= 1:
-        batch_requests = min(batch_requests, int(max_running_requests))
-    caps = [batch_requests * _DECODER_PREFILL_TOKENS_PER_REQUEST]
+        request_limit = min(request_limit, int(max_running_requests))
+
+    # The first whole request is always admissible. For larger batches, shorter
+    # decoder prompts can admit more requests than the maximum-size request does.
+    decoder_tokens_per_request = _DECODER_PREFILL_TOKENS_PER_REQUEST
+    reachable_cap = decoder_tokens_per_request
+    for request_count in range(1, request_limit + 1):
+        decoder_budget = max(0, budget - request_count * encoder_token_count)
+        reachable_cap = max(
+            reachable_cap,
+            min(request_count * decoder_tokens_per_request, decoder_budget),
+        )
+
+    caps = [reachable_cap]
     for key in ("cuda_graph_max_bs_prefill", "max_prefill_tokens", "max_total_tokens"):
         value = overrides.get(key)
         if value is not None and int(value) > 0:
--- a/sglang_omni/models/whisper_asr/engine_builder.py
+++ b/sglang_omni/models/whisper_asr/engine_builder.py
@@ -10,6 +10,7 @@ from sglang_omni.models.whisper_asr.encoder_service import (
     WhisperPreLMEncoderService,
     build_cache_namespace,
 )
+from sglang_omni.models.whisper_asr.request_builders import MAX_PREV_CONTEXT_TOKENS
 from sglang_omni.scheduling.engine_factory import AsrEngineBuilder
 from sglang_omni.scheduling.generation_batch_policy import (
     CudaGraphBackend,
@@ -21,31 +22,35 @@ logger = logging.getLogger(__name__)
 _DEFAULT_ENCODER_GRAPH_BATCH_BUCKETS = (1, 2, 4, 8, 12, 16)
 
 
-def _clamp_prefill_cuda_graph_max_bs(
-    overrides: dict[str, Any], *, context_length: int | None
-) -> int | None:
-    """Clamp the prefill graph budget to the reachable token limits.
+_TASK_PREFIX_SLACK_TOKENS = 8
+_DECODER_PREFILL_TOKENS_PER_REQUEST = (
+    MAX_PREV_CONTEXT_TOKENS + _TASK_PREFIX_SLACK_TOKENS
+)
 
-    Whisper admits its encoder prefix atomically (``chunked_prefill_size=0``),
-    so the shared policy has no chunk to derive a cap from; bound the ladder by
-    the effective prefill, total-token, and model-context limits instead.
-    Returns ``None`` when no limit is known yet (before model setup), leaving
-    the ladder to SGLang's own derivation.
+
+def _reachable_prefill_cuda_graph_max_bs(
+    overrides: dict[str, Any],
+    *,
+    encoder_token_count: int,
+    max_running_requests: int | None,
+) -> int | None:
+    """Cap the prefill graph ladder at the largest decoder batch admission can form.
+
+    Admission charges each request its encoder placeholders as well, while the
+    graph sees only the admitted requests' decoder tokens.
     """
-    if context_length is None or int(context_length) <= 0:
-        context_length = overrides.get("context_length")
-    caps = [
-        int(value)
-        for value in (
-            overrides.get("cuda_graph_max_bs_prefill"),
-            overrides.get("max_prefill_tokens"),
-            overrides.get("max_total_tokens"),
-            context_length,
-        )
-        if value is not None and int(value) > 0
-    ]
-    if not caps:
+    max_prefill_tokens = overrides.get("max_prefill_tokens")
+    if encoder_token_count < 1 or not max_prefill_tokens or int(max_prefill_tokens) < 1:
         return None
+    per_request = encoder_token_count + _DECODER_PREFILL_TOKENS_PER_REQUEST
+    batch_requests = max(1, int(max_prefill_tokens) // per_request)
+    if max_running_requests is not None and int(max_running_requests) >= 1:
+        batch_requests = min(batch_requests, int(max_running_requests))
+    caps = [batch_requests * _DECODER_PREFILL_TOKENS_PER_REQUEST]
+    for key in ("cuda_graph_max_bs_prefill", "max_prefill_tokens", "max_total_tokens"):
+        value = overrides.get(key)
+        if value is not None and int(value) > 0:
+            caps.append(int(value))
     cap = min(caps)
     overrides["cuda_graph_max_bs_prefill"] = cap
     return cap
@@ -171,10 +176,6 @@ class WhisperASREngineBuilder(AsrEngineBuilder):
     def pre_infra_setup(self, checkpoint_dir: str) -> None:
         from transformers import AutoConfig, AutoProcessor, GenerationConfig
 
-        from sglang_omni.models.whisper_asr.request_builders import (
-            MAX_PREV_CONTEXT_TOKENS,
-        )
-
         self.processor = AutoProcessor.from_pretrained(checkpoint_dir)
         self.tokenizer = self.processor.tokenizer
         self.generation_config = GenerationConfig.from_pretrained(checkpoint_dir)
@@ -182,7 +183,10 @@ class WhisperASREngineBuilder(AsrEngineBuilder):
             self.processor.feature_extractor.nb_max_frames // 2
         )
         self.context_length = (
-            self.encoder_token_count + MAX_PREV_CONTEXT_TOKENS + self.max_new_tokens + 8
+            self.encoder_token_count
+            + MAX_PREV_CONTEXT_TOKENS
+            + self.max_new_tokens
+            + _TASK_PREFIX_SLACK_TOKENS
         )
         self.decoder_context_len = int(
             AutoConfig.from_pretrained(checkpoint_dir).max_target_positions or 448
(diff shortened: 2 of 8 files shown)


ZemingL:
## H100 benchmark results for the PR summary

@wirybeaver I cannot edit the PR body with my GitHub permissions. Please copy this benchmark section into the summary. Runtime fixes and benchmarking by @ZemingL.

### Setup

- Baseline: `upstream/main@68abc7eec59ff7b0dce484cb501ffbbb338f9e46`
- Candidate used for performance measurement: `c7dee6dcb356803e96501d94fdc7be6ae8b956d5`
- Hardware: NVIDIA H100 80GB HBM3, driver 580.95.05
- CI image: `hongccc/sglang-omni@sha256:374d0b1c30b2bff685b1716fc64a02ad3b3d0a90fe2ce73ce9861a6992c28101`
- Model: `openai/whisper-base@e37978b90ca9030d5170a5c07aadb050351a65bb`, FP16
- Dataset: full SeedTTS EN split, 1,088 samples
- Concurrency: 1, 2, 4, 8, 16, 32
- Protocol: one discarded application warmup plus three measured repeats per concurrency; same host, image, cache, and command
- Full run order: candidate -> baseline. A post-baseline candidate spot run covered concurrency 16 and 32. The preceding 20-sample smoke used baseline -> candidate order.

The first GPU smoke exposed two candidate startup failures: BCG attention returned head-shaped tensors before Whisper's output projection, and the generic layer discovery omitted cross-attention layers. Commit `c7dee6dc` fixes both. After the fix, capture completed and the runtime attested all 22 buckets from 4 through 256 tokens.

| Concurrency | Main req/s | BCG req/s | Throughput | Main mean / p95 / p99 latency (ms) | BCG mean / p95 / p99 latency (ms) | Main / BCG WER |
| 1 | 22.169 | 24.255 | +9.41% | 45.0 / 53.5 / 60.2 | 41.1 / 50.4 / 55.5 | 0.048275 / 0.048275 |
| 2 | 36.166 | 36.623 | +1.26% | 55.2 / 65.6 / 72.4 | 54.5 / 64.5 / 69.5 | 0.047972 / 0.048073 |
| 4 | 58.463 | 63.042 | +7.83% | 68.0 / 80.4 / 87.5 | 63.1 / 74.6 / 82.7 | 0.047972 / 0.047972 |
| 8 | 88.840 | 97.290 | +9.51% | 89.7 / 104.9 / 115.3 | 81.9 / 97.4 / 110.4 | 0.047972 / 0.048023 |
| 16 | 103.997 | 117.313 | +12.80% | 152.9 / 175.2 / 233.3 | 135.6 / 159.1 / 207.0 | 0.047972 / 0.048123 |
| 32 | 98.679 | 114.718 | +16.25% | 321.9 / 384.1 / 429.1 | 276.8 / 333.1 / 381.4 | 0.047972 / 0.048123 |

### Correctness and mechanism checks

- 19,584/19,584 measured requests succeeded per arm; no skipped or failed samples.
- At concurrency 1, all 3,264 returned transcripts matched exactly between main and BCG.
- Concurrent runs showed small repeat-to-repeat transcript variation in both arms; corpus WER stayed within 0.0480-0.0483.
- Prefill BCG capture: 1.69 seconds, 0.09 GB reported memory, buckets `[4, 8, 12, 16, 20, 24, 28, 32, 48, 64, 80, 96, 112, 128, 144, 160, 176, 192, 208, 224, 240, 256]`.
- Post-baseline candidate spot run: 115.754 req/s at concurrency 16 and 116.868 req/s at concurrency 32, within -1.33% and +1.87% of the primary candidate run.
- Focused CPU tests after the fixes: 136 passed.

sglang_omni/model_runner/sglang_model_runner.py
```diff
@@ -375,12 +376,47 @@ def init_cuda_graphs(self, capture_decode_cuda_graph: bool = True):
...
+    def _prefill_cuda_graph_runner_override(self):
+        runner_cls = self._prefill_cuda_graph_runner_cls()
+        if runner_cls is None:
+            yield
+            return
+
+        from sglang.srt.model_executor.model_runner_components import cuda_graph_setup
+
+        with _PREFILL_RUNNER_OVERRIDE_LOCK:
+            original = cuda_graph_setup.PrefillCudaGraphRunner
```
AkazaAkane: Can the two Whisper-specific needs be expressed through a smaller hook/adapter around layer discovery and prefill metadata, instead of swapping cuda_graph_setup.PrefillCudaGraphRunner globally?

AkazaAkane (changes requested):
Several blocking requests:

1. Rebase onto current main and integrate the now-default pre-LM encoder/precomputed-embedding path. Re-run functional GPU qualification after the rebase. Also we need tests to demonstrate whether we want unconditionally default on or conditionally on.
2. After rebase, we need to run A/B/C: pre-adopter eager vs adopter+BCG-disabled vs adopter+BCG-enabled. This PR changes the eager cross-attention implementation, so old-main vs graph-on is insufficient.
3. Report real replay/eager-fallback coverage and boundary behavior. You can refer to the skill I created for PCG enablement. 
4. We need benchmark on whisper-large-v3 instead of whisper-base.

charliechenye:
@wirybeaver @ZemingL I was independently working through the Whisper item in #1357 and only realized that my branch now overlaps substantially with this PR. Rather than open a competing PR, I would like to contribute the qualification results and a few implementation observations here.

First, thank you both for the work already done on this path. In particular, the earlier H100 `whisper-base` qualification and the fixes around cross-attention output shape and layer discovery were useful references while I was working through the same problem.

### What I tested

I ran an A/B/C qualification on one NVIDIA H200 using:

- Model: `openai/whisper-large-v3`
- Model revision: `06f233fe06e710322aca913c1bc4249a0d71fce1`
- Dataset: full SeedTTS EN split, 1,088 samples
- Dataset revision: `27f4c1adee83b5b29b7c4b375f6b976324bda308`
- SGLang: `0.5.16`
- PyTorch: `2.11.0+cu130`
- Decode CUDA Graph backend: `FULL` in all arms
- Concurrency: `1, 2, 4, 8, 16, 32, 64`
- Three counterbalanced rounds per concurrency
- Fresh server for every arm
- Synthetic non-SeedTTS warmup only, to avoid warming the SeedTTS encoder cache

The A/B/C split was:

- **A**: baseline `99388dca4165e99b9c4c32b8cac416ce9a6ca190`, prefill BCG disabled
- **B**: candidate `357edd5d18904da9ecc2aea4330c3d34c75e1d84`, prefill BCG disabled
- **C**: same candidate, prefill BCG breakable

The purpose of the split was:

- **A → B**: detect regressions from the Whisper encoder/decoder execution-path changes
- **B → C**: isolate the effect of breakable Prefill CUDA Graph
- **A → C**: measure the total end-to-end effect

Across the full confirmation there were:

- **68,544 measured requests**
- all requested workloads completed
- no skipped or incomplete benchmark arms
- all seven concurrency points completed all three A/B/C rounds

### B → C: isolated Prefill BCG effect

Because B and C use the same candidate commit, this is the cleanest measurement of the Prefill BCG effect.

| Concurrency | Throughput | Mean latency | p95 latency | p99 latency |
| 1 | **+10.71%** | **-9.69%** | **-8.18%** | **-8.06%** |
| 2 | **+12.52%** | **-11.13%** | **-12.90%** | **-14.77%** |
| 4 | **+15.91%** | **-13.73%** | **-15.16%** | **-14.93%** |
| 8 | **+13.74%** | **-12.09%** | **-10.84%** | **-2.01%** |
| 16 | **+16.68%** | **-14.21%** | **-9.70%** | **-38.93%** |
| 32 | **+16.95%** | **-14.52%** | **-15.38%** | **-19.99%** |
| 64 | **+18.97%** | **-16.03%** | **-17.81%** | **-25.16%** |

The throughput improvement was positive at every tested concurrency and generally increased with concurrency.

### A → B: execution-path regression check

The candidate with Prefill BCG disabled did not show a throughput regression against the historical baseline:

| Concurrency | A → B throughput |
| ---: | ---: |
| 1 | +2.46% |
| 2 | +3.10% |
| 4 | +3.74% |
| 8 | +3.66% |
| 16 | +4.85% |
| 32 | +6.93% |
| 64 | +2.25% |

I therefore did not observe evidence that the BCG gain was hiding a material regression in the non-BCG candidate path.

### Graph routing and capture checks

Before the full benchmark I ran a dedicated graph smoke with:

- A: baseline, prefill disabled, decode FULL
- B: candidate, prefill disabled, decode FULL
- C: candidate, prefill breakable, decode FULL

The smoke passed all three arms.

For C:

- 40/40 observed SeedTTS prefill requests used BCG
- graph request fraction: `1.0`
- SGLang attested the full 22-bucket capture ladder:

```text
[4, 8, 12, 16, 20, 24, 28, 32, 48, 64, 80, 96,
 112, 128, 144, 160, 176, 192, 208, 224, 240, 256]
```

This was useful to distinguish "configured for BCG" from "actually replayed through BCG."

### Accuracy and transcript behavior

At concurrency 1, the harness enforced exact transcript equality and all A/B/C comparisons passed.

At higher concurrency, a small number of exact transcript differences appeared, but they were not specific to BCG.

For example, at c8:

- A vs B exact mismatches by round: `2`, `8`, `5`
- B vs C exact mismatches by round: `3`, `7`, `6`

Many of the same borderline SeedTTS samples appeared in both A/B and B/C mismatch sets.

Corpus WER was identical between B and C at six of the seven tested concurrency levels. At c8, the absolute difference was approximately `0.0000252` WER, or `0.00252` percentage points.

My interpretation is therefore:

> No material ASR accuracy regression attributable to Prefill BCG was observed. Concurrent runs exhibit a small amount of batching/numerical transcript variation even when BCG is disabled.

I would not describe the concurrent path as bit-exact.

### Important qualification

These results are **not a current-head benchmark of #1530**.

My full run used an independent implementation and an older baseline. In particular, the current PR is now based on newer `main` changes, including the pre-LM pinned host-memory cache work.

I would treat these results as:

- independent H200 evidence that the Whisper breakable-prefill mechanism works on `whisper-large-v3`
- evidence for the expected performance range
- validation of the requested A/B/C methodology
- complementary evidence to the earlier H100 `whisper-base` results

I would **not** treat them as a replacement for the current-head A/B/C matrix requested in review.

My full qualification covered the default pre-LM enabled path. I have not run the corresponding pre-LM disabled A1/B1/C1 matrix against the current stacked #1634 + #1530 heads.

### Implementation observations from the independent branch

Our branches converged on the same core requirements:

1. precompute/write encoder cross-attention K/V outside the captured decoder body
2. capture both decoder self-attention and cross-attention layers
3. preserve live encoder metadata during prefill replay
4. use a Whisper-specific adapter around SGLang 0.5.16's generic prefill runner

A few differences from my independent branch may be worth considering:

- I restored `model_runner.attention_layers` after the Whisper prefill runner finished its BCG setup, rather than leaving the temporary self-attention + cross-attention registry installed for later eager/decode paths.
- I added a `can_run_graph()` guard so requests without a positive encoder context fall back rather than entering the Whisper prefill graph.
- A GPU smoke exposed a decode-CUDA-Graph regression when `ForwardBatch.encoder_cached` was `None`. The minimal fix was to scope request-specific encoder K/V population to non-decode forwards. Decode should consume already-cached encoder K/V rather than trying to populate it.
- I am leaning toward keeping Whisper's production code capability-only rather than prescribing a model-owned `256` capture cap. In that version, the operator/benchmark explicitly chooses the prefill graph cap while the model only declares that it supports the breakable-prefill contract.

I do **not** think my branch solves the review concern about temporarily swapping `cuda_graph_setup.PrefillCudaGraphRunner`; both implementations currently use that pinned-SGLang compatibility mechanism in some form.

### How I can help

Since the implementations overlap heavily, I do not plan to pursue a competing Whisper BCG PR.

I am happy to contribute the H200 harness, help adapt it to the current stacked #1634 + #1530 heads, or run additional targeted qualification that would help close the remaining review blockers.

musclemuller:
> Several blocking requests:
> 
> 1. Rebase onto current main and integrate the now-default pre-LM encoder/precomputed-embedding path. Re-run functional GPU qualification after the rebase. Also we need tests to demonstrate whether we want unconditionally default on or conditionally on.
> 2. After rebase, we need to run A/B/C: pre-adopter eager vs adopter+BCG-disabled vs adopter+BCG-enabled. This PR changes the eager cross-attention implementation, so old-main vs graph-on is insufficient.
> 3. Report real replay/eager-fallback coverage and boundary behavior. You can refer to the skill I created for PCG enablement.
> 4. We need benchmark on whisper-large-v3 instead of whisper-base.

Hi, @AkazaAkane , Could you share the name or link/path of the PCG enablement skill referenced in your review? We would like to follow its A/B/C qualification and replay/eager-fallback coverage contract for the ARK-ASR benchmark in #1528.

AkazaAkane:
> > Several blocking requests:
> > 
> > 1. Rebase onto current main and integrate the now-default pre-LM encoder/precomputed-embedding path. Re-run functional GPU qualification after the rebase. Also we need tests to demonstrate whether we want unconditionally default on or conditionally on.
> > 2. After rebase, we need to run A/B/C: pre-adopter eager vs adopter+BCG-disabled vs adopter+BCG-enabled. This PR changes the eager cross-attention implementation, so old-main vs graph-on is insufficient.
> > 3. Report real replay/eager-fallback coverage and boundary behavior. You can refer to the skill I created for PCG enablement.
> > 4. We need benchmark on whisper-large-v3 instead of whisper-base.
> 
> Hi, @AkazaAkane , Could you share the name or link/path of the PCG enablement skill referenced in your review? We would like to follow its A/B/C qualification and replay/eager-fallback coverage contract for the ARK-ASR benchmark in #1528.

Thanks for contribution. You can refer here: https://github.com/AkazaAkane/omni-skillhub/tree/main/enable-prefill-cuda-graph

wirybeaver:
The benchmark result of PR #1634 shows that the adaptor change won't cause regression while enable_pre_lm_encoder = false / true and breakable prefill cuda graph is disabled. In other words, A1->B1->B1->A1 and A2 -> B2 -> B2 -> A2 has been benchmarked on H100. I will add @musclemuller as the co-author of https://github.com/sgl-project/sglang-omni/pull/1530.

musclemuller:
@AkazaAkane Before starting the paid H100 qualification, could you confirm the intended benchmark candidates and stack strategy?

Current state:

- PR #1530 head: `e43455759a8f424135dec2685054e24ad4d99370`
- PR #1634 adopter head: `f1e5d148d2121f0b983ff7d3438f738ba93dc501`
- Both are based on `10733c9`, while current `main` is now `07b63c0` and eight commits ahead.
- The discussion about replacing `cuda_graph_setup.PrefillCudaGraphRunner` globally is still unresolved.

Could you please confirm:

1. Should #1530 be rebased onto current `main` before benchmarking?
2. Is the current global runner override the implementation to qualify, or should the smaller hook/adapter design land first?
3. Should #1634 merge first and we then run B-versus-C, or should #1634 be closed and #1530 receive a fresh full A/B/C benchmark?
4. Once decided, which exact B and C commits should be treated as stable paid-benchmark candidates?

The prepared qualification covers `whisper-large-v3`, both pre-LM modes, both B/C orders, real replay/fallback and boundary evidence, correctness gates, and graph capture/memory telemetry. We will not provision the H100 until the candidate and comparison strategy are explicitly confirmed.

wirybeaver:
@charliechenye I have rebased on the most recent main branch. Would you mind rerun the A/B/C qualification and co-authored this PR? I need to focus on other PRs. The A/B qualification was completed by @musclemuller on #1634 and no regression by the adaptor change.

charliechenye:
@wirybeaver @AkazaAkane @ZemingL I reran the qualification against the current rebased PR head and completed the full `openai/whisper-large-v3` A/B/C matrix on one H200, including both `enable_pre_lm_encoder=false` and `true`.

## Qualification verdict

**Performance and correctness qualification passes for both pre LM modes.** The data supports enabling breakable Prefill CUDA Graph for Whisper-ASR in both modes rather than gating it on `enable_pre_lm_encoder`.

The important attribution point is that I did **not** compare old-main directly against graph-on. I used an A/B/C split so the execution path/adaptor changes and the Prefill CUDA Graph effect are measured separately.

## What `enable_pre_lm_encoder` means

This flag does **not** turn the Whisper audio encoder on or off. Whisper always needs to encode the audio.

It changes **where and when** the encoder runs relative to SGLang LM admission/prefill:

| `enable_pre_lm_encoder` | Execution path | What reaches LM/prefill | Admission/caching implication |
|---|---|---|---|
| `false` | **encoder in prefill** | raw Whisper input features stay on the multimodal request; the audio encoder runs as part of generation/prefill | encoder work remains coupled to atomic prefill admission; reachable encoder graph batch size is constrained by `max_prefill_tokens // encoder_token_count` |
| `true` | **pre LM encoder service** | Whisper encoder hidden states are computed before LM admission and attached as `PRECOMPUTED_EMBEDDING` | encoder work is decoupled from atomic LM prefill admission; the service can batch encoder work up to `pre_lm_max_batch_size`, deduplicate/cache by audio fingerprint, and store fixed-size encoder states in the CPU LRU cache |

At the frozen PR head, `enable_pre_lm_encoder=true` is the configured default. The pre LM service still runs the Whisper encoder on the model/GPU; it is not a CPU encoder. The important change is that it computes and attaches the encoder hidden states **before** LM admission instead of making the generation prefill path perform the encoder work.

The two modes therefore exercise meaningfully different Prefill BCG integration paths:

- `false`: Prefill BCG must coexist with the encoder in prefill multimodal path.
- `true`: Prefill BCG consumes requests whose Whisper encoder states have already been produced and attached as precomputed embeddings.

This is why I qualified both modes.

**Important:** the results below are not a causal benchmark of `enable_pre_lm_encoder=true` versus `false`. The causal comparisons are **within each mode**. A dedicated flag-only true/false benchmark would need its own counterbalanced experiment. The purpose here is to prove that Prefill BCG qualifies under both supported execution paths.

## A/B/C design

| Arm | Commit | `enable_pre_lm_encoder` | Prefill CUDA Graph | Purpose |
|---|---|---:|---|---|
| **A** | PR base `ca6aa45513ff0c9728ff122629fde96acc26054d` | tested both `false` and `true` | explicitly disabled | pre-adopter eager baseline |
| **B** | PR head `13fda416a83018800fefdb45cf27db7076ed3393` | same mode as A | explicitly disabled | current rebased adopter/execution path without Prefill BCG |
| **C** | same PR head `13fda416a83018800fefdb45cf27db7076ed3393` | same mode as A/B | `breakable` | identical code to B, with Prefill BCG enabled |

The comparisons therefore mean:

| Comparison | Interpretation |
|---|---|
| **A -> B** | regression/control check for the rebased Whisper execution path/adaptor changes |
| **B -> C** | **isolated breakable Prefill CUDA Graph effect** on identical code |
| **A -> C** | total end-to-end PR effect, but not the isolated BCG attribution metric |

## Frozen benchmark setup

| Item | Value |
|---|---|
| Hardware | 1x NVIDIA H200 |
| Model | `openai/whisper-large-v3` |
| Model revision | `06f233fe06e710322aca913c1bc4249a0d71fce1` |
| Dataset | `zhaochenyang20/seed-tts-eval-arrow`, English split |
| Dataset revision | `27f4c1adee83b5b29b7c4b375f6b976324bda308` |
| Dataset size | 1,088 samples |
| Dtype | BF16 |
| Decode CUDA Graph backend | `full` in A/B/C |
| Prefill backend | A/B=`disabled`, C=`breakable` |
| `max_running_requests` | 64 |
| `mem_fraction_static` | 0.85 |
| Random seed | 20260824 |
| Server lifecycle | fresh server for every arm / round / concurrency / pre LM mode |
| Warmup | synthetic non-SeedTTS warmup only, so measured SeedTTS samples do not pre-populate the encoder cache |
| Main matrix order | counterbalanced `ABC`, `BCA`, `CAB` |
| P0 c2/c32 order | all six permutations: `ABC`, `ACB`, `BAC`, `BCA`, `CAB`, `CBA` |
| Full-set measured requests used in the final table | 176,256 |

The 176,256 request total is the final full-set evidence used below: c1/c4/c8/c16/c64 used 3 A/B/C rounds in both pre LM modes, while c2 and c32 were strengthened to 6 full-set permutation-balanced rounds in both modes.

## CPU / functional / graph smoke gates

| Gate | Result |
|---|---|
| Focused CPU tests | **166 passed** |
| `py_compile` on changed Whisper/graph files | **PASS** |
| `git diff --check` base..head | **PASS** |
| Synthetic non-SeedTTS server warmup | **PASS** in all smoke arms |
| pre LM OFF graph smoke | **PASS** |
| pre LM ON graph smoke | **PASS** |
| Capture ladder | `[4, 8, 12, 16, 20, 24, 28, 32, 48, 64, 80, 96, 112, 128, 144, 160, 176, 192, 208, 224, 240, 256]` |
| C capture backend | `breakable` |
| A/B smoke replay count | `0` |
| C smoke eager fallback | `0` |

Graph smoke details:

| pre LM mode | C measured prefill batches | C replay | C eager | replay fraction | capture time | reported capture memory |
| OFF | 14 | 14 | 0 | 100% | 4.28 s | 0.23 GB |
| ON | 14 | 14 | 0 | 100% | 4.54 s | 0.23 GB |

## Final performance results

The tables below report **B -> C as the isolated Prefill BCG effect**. All percentage columns are **actual candidate vs reference deltas**, computed as `(candidate / reference - 1) * 100`. Positive throughput is better; negative latency/P95/RTF is better.

### `enable_pre_lm_encoder=false`

| c | A req/s | B req/s | C req/s | A -> B delta throughput | **B -> C delta throughput** | **B -> C delta mean latency** | **B -> C delta P95 latency** | **B -> C delta RTF** | A WER | B WER | C WER | C replay | C eager |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 8.90 | 9.75 | 10.96 | +9.57% | **+12.41%** | **-11.07%** | **-11.23%** | **-11.27%** | 1.3923% | 1.3923% | 1.3923% | 100% | 0 |
| 2 | 15.06 | 16.75 | 19.09 | +11.26% | **+13.94%** | **-12.27%** | **-17.26%** | **-12.37%** | 1.3998% | 1.3998% | 1.3986% | 100% | 0 |
| 4 | 22.88 | 27.08 | 30.39 | +18.38% | **+12.20%** | **-10.89%** | **-13.37%** | **-10.93%** | 1.3998% | 1.3998% | 1.3998% | 100% | 0 |
| 8 | 29.28 | 34.60 | 39.95 | +18.17% | **+15.48%** | **-13.61%** | **-17.74%** | **-13.66%** | 1.4023% | 1.3998% | 1.4049% | 100% | 0 |
| 16 | 39.61 | 49.67 | 53.00 | +25.41% | **+6.69%** | **-6.42%** | **-8.51%** | **-6.42%** | 1.4023% | 1.4023% | 1.3998% | 100% | 0 |
| 32 | 43.57 | 55.66 | 60.14 | +27.74% | **+8.04%** | **-7.78%** | **-11.74%** | **-7.73%** | 1.3998% | 1.3998% | 1.3998% | 100% | 0 |
| 64 | 43.95 | 59.92 | 61.20 | +36.32% | **+2.14%** | **-2.18%** | **-4.40%** | **-2.10%** | 1.3998% | 1.3998% | 1.3998% | 100% | 0 |

### `enable_pre_lm_encoder=true`

| c | A req/s | B req/s | C req/s | A -> B delta throughput | **B -> C delta throughput** | **B -> C delta mean latency** | **B -> C delta P95 latency** | **B -> C delta RTF** | A WER | B WER | C WER | C replay | C eager |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 | 9.49 | 9.46 | 10.75 | -0.33% | **+13.65%** | **-12.07%** | **-11.75%** | **-12.25%** | 1.3923% | 1.3923% | 1.3923% | 100% | 0 |
| 2 | 15.64 | 15.82 | 19.31 | +1.13% | **+22.06%** | **-18.16%** | **-21.09%** | **-18.25%** | 1.3998% | 1.3998% | 1.3998% | 100% | 0 |
| 4 | 25.71 | 25.25 | 29.65 | -1.81% | **+17.43%** | **-14.93%** | **-18.35%** | **-14.84%** | 1.3998% | 1.3998% | 1.3998% | 100% | 0 |
| 8 | 35.36 | 35.93 | 45.39 | +1.60% | **+26.32%** | **-21.29%** | **-25.49%** | **-21.23%** | 1.4049% | 1.4074% | 1.4049% | 100% | 0 |
| 16 | 46.18 | 51.68 | 61.48 | +11.89% | **+18.98%** | **-15.89%** | **-12.59%** | **-16.06%** | 1.4049% | 1.4023% | 1.4074% | 100% | 0 |
| 32 | 57.51 | 62.39 | 74.42 | +8.48% | **+19.28%** | **-16.33%** | **-13.95%** | **-16.30%** | 1.3998% | 1.3998% | 1.3998% | 100% | 0 |
| 64 | 56.87 | 60.99 | 77.70 | +7.25% | **+27.39%** | **-21.89%** | **-19.85%** | **-21.81%** | 1.3998% | 1.3998% | 1.3998% | 100% | 0 |

All 14 mode/concurrency points have positive aggregate B -> C throughput. The pre LM OFF c32/c64 points are close to saturation and have some round noise, but neither shows a stable regression.

## Round by round B -> C throughput stability

| pre LM | c | B -> C throughput by measured round |
|---|---:|---|
| OFF | 1 | `+10.2%, +13.1%, +14.0%` |
| OFF | 2 | `+20.1%, +14.1%, +7.7%, +14.9%, +13.8%, +13.6%` |
| OFF | 4 | `+10.5%, +12.9%, +13.2%` |
| OFF | 8 | `+9.3%, +17.2%, +20.7%` |
| OFF | 16 | `+3.5%, +6.5%, +10.5%` |
| OFF | 32 | `+5.5%, +13.0%, -2.2%, +11.2%, +8.0%, +14.7%` |
| OFF | 64 | `+3.2%, -2.3%, +6.0%` |
| ON | 1 | `+17.2%, +8.9%, +15.1%` |
| ON | 2 | `+20.2%, +28.6%, +26.3%, +22.0%, +20.6%, +15.3%` |
| ON | 4 | `+16.5%, +15.0%, +21.0%` |
| ON | 8 | `+28.9%, +35.8%, +16.2%` |
| ON | 16 | `+17.1%, +21.1%, +18.7%` |
| ON | 32 | `+14.8%, +16.1%, +27.6%, +16.0%, +19.9%, +22.0%` |
| ON | 64 | `+31.5%, +23.9%, +27.0%` |

Only two individual pre LM OFF rounds were negative (`c32 -2.2%`, `c64 -2.3%`); the corresponding aggregates remained +8.04% and +2.14%. Every measured pre LM ON B -> C round was positive.

## Replay / eager fallback coverage

For the final full-set matrix, every C arm reported **100% replay of observed Prefill batches and zero measured eager fallback**. A/B reported no Prefill replay.

| pre LM | c1 | c2 | c4 | c8 | c16 | c32 | c64 |
| OFF | 100% replay / 0 eager | 100% / 0 | 100% / 0 | 100% / 0 | 100% / 0 | 100% / 0 | 100% / 0 |
| ON | 100% replay / 0 eager | 100% / 0 | 100% / 0 | 100% / 0 | 100% / 0 | 100% / 0 | 100% / 0 |

The counters are Prefill-batch counters, not request counters.

## Correctness results

At concurrency 1, exact transcript parity held in both modes:

| pre LM | A/B exact mismatches | B/C exact mismatches |
|---|---:|---:|
| OFF | 0 | 0 |
| ON | 0 | 0 |

At higher concurrency, small exact transcript variation exists in the eager A/B control itself as well as B/C. The counts below are summed over the measured rounds for that matrix point.

| pre LM | c | A/B exact mismatches | B/C exact mismatches | A WER | B WER | C WER |
| OFF | 1 | 0 | 0 | 1.3923% | 1.3923% | 1.3923% |
| OFF | 2 | 12 | 5 | 1.3998% | 1.3998% | 1.3986% |
| OFF | 4 | 7 | 4 | 1.3998% | 1.3998% | 1.3998% |
| OFF | 8 | 18 | 16 | 1.4023% | 1.3998% | 1.4049% |
| OFF | 16 | 14 | 19 | 1.4023% | 1.4023% | 1.3998% |
| OFF | 32 | 10 | 10 | 1.3998% | 1.3998% | 1.3998% |
| OFF | 64 | 9 | 6 | 1.3998% | 1.3998% | 1.3998% |
| ON | 1 | 0 | 0 | 1.3923% | 1.3923% | 1.3923% |
| ON | 2 | 0 | 0 | 1.3998% | 1.3998% | 1.3998% |
| ON | 4 | 4 | 6 | 1.3998% | 1.3998% | 1.3998% |
| ON | 8 | 10 | 14 | 1.4049% | 1.4074% | 1.4049% |
| ON | 16 | 23 | 12 | 1.4049% | 1.4023% | 1.4074% |
| ON | 32 | 17 | 13 | 1.3998% | 1.3998% | 1.3998% |
| ON | 64 | 3 | 8 | 1.3998% | 1.3998% | 1.3998% |

Across the table, there is no consistent BCG specific WER regression direction. Exact differences at concurrent settings are of the same kind already present in the eager A/B control. In the focused pre LM OFF c2 investigation, all B/C mismatch IDs occurred in only 1/6 rounds; there was no deterministic BCG specific mismatch that reproduced across all permutation-balanced rounds. One lexical B/C difference actually moved C back to the reference (`Kimki` -> `Khimki`).

Therefore my correctness conclusion is:

> **No material or reproducible BCG specific accuracy regression was observed.** Exact transcript parity holds at c1 in both pre LM modes. At higher concurrency there is small batching/numerical transcript variation in both the eager control and BCG comparison, while corpus WER remains effectively unchanged.

I would **not** describe the concurrent outputs as bit exact.

## Default policy conclusion

The requested policy question was whether Prefill BCG should be unconditional or conditional on `enable_pre_lm_encoder`.

Both execution paths qualify for Prefill BCG:

| Mode | What it represents | Performance | Correctness | Replay coverage | Default policy evidence |
|---|---|---|---|---|---|
| pre LM OFF | encoder in prefill path | PASS | PASS, no material/reproducible regression | 100% measured Prefill replay, 0 eager fallback | supports BCG ON |
| pre LM ON | current default pre LM/precomputed-embedding path | PASS | PASS, no material/reproducible regression | 100% measured Prefill replay, 0 eager fallback | supports BCG ON |

Based on these results, I do **not** see evidence requiring the Prefill BCG default to be gated on `enable_pre_lm_encoder`. This should **not** be read as "pre LM ON is X% faster than pre LM OFF"; that was not the causal question or experiment design. It means **BCG itself qualifies when layered on top of either execution path**.

## Summary

- current PR base/head were pinned exactly at `ca6aa45513ff0c9728ff122629fde96acc26054d` -> `13fda416a83018800fefdb45cf27db7076ed3393`
- `openai/whisper-large-v3`, full 1,088-sample SeedTTS EN, both pre LM modes
- attribution-safe **A/B/C**, with B and C on identical head code and differing only in Prefill BCG state
- all 14 B -> C aggregate throughput deltas are positive
- all 14 B -> C aggregate mean latency deltas are negative (i.e. improvements)
- C uses 100% measured Prefill replay and 0 measured eager fallback at all 14 matrix points
- exact transcript parity at c1 in both modes
- no material or reproducible BCG specific WER regression at higher concurrency
- focused c32 performance and c2 correctness P0 rechecks cleared the earlier anomalies
- data supports Prefill BCG enabled in both pre LM modes

wirybeaver:
@charliechenye Appreciate your benchmarking! I add you into the co-author and rebase the most recent main due to configV2 was landed into main.

@AkazaAkane @stephenkgli Could you review the PR as the A/B/C benchmarking result is positive?

lijrjyan:
@AkazaAkane Hi, reqeusted experiment/qualification results are as following, including A/B/C test and the boundary behavior report with whisper-large-v3.

Enviroment: this head (22a5491a) and `main` (f62fd76cc). H200, `openai/whisper-large-v3`, sglang 0.5.18, both pre-LM modes, three interleaved rounds per configuration, fresh server each time.

- **A** = upstream `main`, eager prefill
- **B** = this PR with `stages.asr.engine.cuda_graph_backend_prefill: disabled` (encoder/decoder execution-path changes only)
- **C** = this PR as shipped (breakable prefill CUDA graph on, cap 696)

**Throughput on 240 short clips (req/s, three-round means)**

| Mode | Concurrency | A | B | C | A→B | A→C |
| pre-LM true (default) | 1 | 10.2 | 10.7 | 11.6 | +4.5% | +13.5% |
| pre-LM true | 8 | 39.4 | 40.4 | 46.9 | +2.4% | +19.1% |
| pre-LM true | 32 | 73.5 | 80.2 | 100.3 | +9.2% | +36.5% |
| pre-LM false | 1 | 8.2 | 8.8 | 9.7 | +6.7% | +17.7% |
| pre-LM false | 8 | 32.0 | 39.9 | 41.0 | +24.7% | +28.0% |
| pre-LM false | 32 | 46.9 | 58.6 | 61.1 | +25.1% | +30.4% |

**SeedTTS EN, repository ASR benchmark (1,088 clips × 3 repeats, pre-LM true; A, B, C as above)**

| Concurrency | Metric | A | B | C | A→B | A→C |
|---:|---|---:|---:|---:|---:|---:|
| 1 | Throughput (samples/s) | 12.2 | 12.2 | 13.8 | −0.1% | +13.4% |
| 1 | Mean latency (ms) | 82 | 82 | 72 | +0.1% | −11.9% |
| 1 | Corpus WER | 0.01370 | 0.01370 | 0.01370 | 0 | 0 |
| 32 | Throughput (samples/s) | 73.5 | 74.5 | 101.7 | +1.4% | +38.4% |
| 32 | Mean latency (ms) | 432 | 426 | 311 | −1.5% | −28.0% |
| 32 | Corpus WER | 0.01377 | 0.01377 | 0.01377 | 0 | 0 |

**Correctness.** Serial (concurrency 1) transcripts are identical across A, B and C on the 240-clip fixture (180/180 per mode). On SeedTTS the B and C transcripts match exactly, 0 mismatches out of 6,528 per mode. Sample-WER maximum is unchanged.

**Graph coverage.** Every graph-eligible prefill forward on the measured workload replayed through the captured graph (0 eager fallbacks); the largest replayed shape is 696 tokens. The distribution of scheduled decoder prefill shapes over 4,158 batches per mode matches the runtime's replay-bucket counters bucket by bucket.

**Boundary behaviour.** Smallest shape, long prompt at concurrency 1, 8, 16 and 32, forced synchronized queues, SRT and verbose-json timestamps, >30 s chunked audio, 0.1 s silence, and the rollback configuration all return HTTP 200 with transcripts identical to B. With an explicit `cuda_graph_max_bs_prefill` of 64 or 512, over-cap requests take the shared eager fallback and subsequent in-cap requests replay again.

**Cost of the graph.** Prefill graph capture 5.65 s and 0.36 GB at startup; idle VRAM +344 MiB versus B. The previous 1988-token ladder cost 7.72 s, 0.48 GB and +449 MiB; the cap only removes buckets that admission can never form.

AkazaAkane (approved):

lijrjyan (approved):

sglang_omni/models/whisper_asr/engine_builder.py
```diff
@@ -10,13 +10,52 @@
...
+    graph sees only the admitted requests' decoder tokens.
+    """
+    max_prefill_tokens = overrides.get("max_prefill_tokens")
+    if encoder_token_count < 1 or not max_prefill_tokens or int(max_prefill_tokens) < 1:
+        return None
+    per_request = encoder_token_count + _DECODER_PREFILL_TOKENS_PER_REQUEST
+    batch_requests = max(1, int(max_prefill_tokens) // per_request)
+    if max_running_requests is not None and int(max_running_requests) >= 1:
+        batch_requests = min(batch_requests, int(max_running_requests))
+    caps = [batch_requests * _DECODER_PREFILL_TOKENS_PER_REQUEST]
```
stephenkgli: The graph cap can exclude valid prefill batches

Dividing the budget by the largest possible request underestimates how many shorter requests can fit.

For example, with `max_prefill_tokens=5120`, `max_running_requests>=3`, and the default `max_new_tokens=256`, a long `prompt` gives each request 192 decoder tokens:

- The scheduler admits three requests: `3 * (1500 + 192) = 5076 < 5120`.
- After removing the encoder tokens, the graph receives `3 * 192 = 576` tokens.
- This code sets the cap to `(5120 // 1732) * 232 = 464`, so the whole batch falls back to eager.

Please use a cap that covers the batches the scheduler can actually admit, then let `build_generation_batch_overrides()` build the buckets.

I reproduced this with the production request builder, admission code, encoder-token removal, and PCG eligibility check in an isolated CPU test.

wirybeaver: y1 = max_prefill_tokens - request * encoder_token (i.e., 1500); y2 = decoder_token (i.e., 192) * request. We wanted to find a request = x to maximize the min(y1, y2). x = 5120 / (1500 + 192) = 3.02. We will examine both x = 3 and x = 4: When x = 3, min(y1, y2) = y2 = 192 \* 3; When x = 4, min(y1, y2) = y1 < 0; The cap will be max(192 \* 3, <0) = 576 tokens. Check the most recent commit and the O(1) function _max_reachable_decoder_prefill_tokens()

PR #1532, opened 2026-08-14
[Bug][Qwen3-Omni] Restore deferred CUDA graph args on bootstrap failure

### LINKED ISSUE #1515: [Bug][Qwen3-Omni] Make deferred Thinker CUDA Graph bootstrap exception-safe
### Checklist

### Checklist

- [x] If the issue you raised is not a feature but a question, please raise a discussion instead.
- [x] Please use English.

## Motivation

While auditing the Qwen3-Omni Thinker bootstrap lifecycle around #1380 and #1381, we found an exception-safety gap in deferred CUDA Graph initialization.

For speech-enabled Thinker, CUDA Graph capture is deferred while SGLang infrastructure is constructed with temporary server argument overrides:

```text
enable_return_hidden_states = True
disable_cuda_graph = True
```

After infrastructure creation succeeds, the original values are restored before `init_cuda_graphs()`.

This ordering is intentional and is part of the lifecycle introduced by #1380.

However, restoration currently occurs only after `create_sglang_infrastructure()` returns successfully.

If infrastructure creation raises, the temporary values remain on the caller's `server_args`.

This does not indicate a correctness problem on the normal successful serving path. It is an exception-path lifecycle bug: temporary bootstrap state can escape the scope in which it was intended to apply.

Related:

- #1018, Qwen3-Omni serving optimization roadmap
- #1022, Qwen3-Omni execution tracker
- #1380, graph-safe eager Thinker hidden capture
- #1381, where this was found during follow-up auditing

## Reproduction

Start with speech output enabled and CUDA Graphs enabled.

For example, the relevant initial state can be represented as:

```text
speech_enabled = True

server_args.disable_cuda_graph = False
server_args.enable_return_hidden_states = False
```

Because speech requires hidden-state capture, Thinker bootstrap defers CUDA Graph initialization and temporarily changes the arguments to:

```text
server_args.disable_cuda_graph = True
server_args.enable_return_hidden_states = True
```

Then force:

```python
create_sglang_infrastructure(...)
```

to raise before returning.

For example:

```python
def failing_create_sglang_infrastructure(*args, **kwargs):
    assert server_args.disable_cuda_graph is True
    assert server_args.enable_return_hidden_states is True
    raise RuntimeError("infrastructure initialization failed")
```

Call the normal Thinker scheduler/bootstrap path and catch the expected exception.

### Expected state after the failure

The temporary overrides should have been scoped to infrastructure construction, so the original values should be restored:

```text
server_args.disable_cuda_graph == False
server_args.enable_return_hidden_states == False
```

### Current state after the failure

The restoration block is never reached, leaving:

```text
server_args.disable_cuda_graph == True
server_args.enable_return_hidden_states == True
```

The original infrastructure exception still propagates, but the caller-visible `server_args` object has been left in temporary bootstrap state.

The same invariant should hold for arbitrary original values. Restoration should restore the values that existed before the temporary override rather than assume they were initially `False`.

## Environment

Identified against SGLang-Omni `main`:

```text
sglang-omni: 30b6406c1c396ef2719500d00a51b766632f17aa
sglang:      0.5.16
torch:       2.11.0
transformers: 5.12.1
```

This is a bootstrap control-flow issue and does not require CUDA execution to reproduce.

A focused unit test can reproduce it by making `create_sglang_infrastructure()` raise after verifying that the temporary arguments were applied.

No GPU end-to-end reproduction is required to establish the exception-path state leak.

## Current Behavior

The current lifecycle is effectively:

```python
if defer_cuda_graph_capture:
    saved_disable_cuda_graph = server_args.disable_cuda_graph
    saved_return_hidden_states = server_args.enable_return_hidden_states

    override_server_args(
        server_args,
        ...,
        enable_return_hidden_states=True,
        disable_cuda_graph=True,
    )

infrastructure = create_sglang_infrastructure(...)

if defer_cuda_graph_capture:
    override_server_args(
        server_args,
        ...,
        disable_cuda_graph=saved_disable_cuda_graph,
        enable_return_hidden_states=saved_return_hidden_states,
    )

    model_worker.model_runner.init_cuda_graphs()
```

The successful path has the desired ordering:

```text
temporary overrides
        ↓
infrastructure construction
        ↓
restore original arguments
        ↓
initialize CUDA Graphs
```

The missing case is:

```text
temporary overrides
        ↓
infrastructure construction raises
        ↓
restoration is skipped
```

## Proposed Direction

Treat the two server argument changes as temporary scoped bootstrap state.

Infrastructure construction should be protected so restoration happens whether it succeeds or raises.

Conceptually:

```python
if defer_cuda_graph_capture:
    saved_disable_cuda_graph = server_args.disable_cuda_graph
    saved_return_hidden_states = server_args.enable_return_hidden_states

    override_server_args(
        server_args,
        ...,
        enable_return_hidden_states=True,
        disable_cuda_graph=True,
    )

try:
    infrastructure = create_sglang_infrastructure(...)
finally:
    if defer_cuda_graph_capture:
        override_server_args(
            server_args,
            ...,
            disable_cuda_graph=saved_disable_cuda_graph,
            enable_return_hidden_states=saved_return_hidden_states,
        )

if defer_cuda_graph_capture:
    model_worker.model_runner.init_cuda_graphs()
```

The exact implementation does not have to follow this structure, but the lifecycle invariant should be:

> Temporary bootstrap overrides never survive infrastructure construction, including when construction fails.

The original exception should still propagate unchanged.

`init_cuda_graphs()` should only run after successful infrastructure construction and after restoration of the original server arguments.

### In scope

- restoring temporary `disable_cuda_graph` state when infrastructure construction fails;
- restoring temporary `enable_return_hidden_states` state when infrastructure construction fails;
- preserving the current successful-path ordering;
- preserving arbitrary original argument values;
- adding an exception-path lifecycle regression test.

### Out of scope

- changing hidden-state capture behavior;
- changing CUDA Graph eligibility;
- enabling speech-output Prefill CUDA Graph replay;
- changing the deferred-capture policy introduced by #1380;
- changing infrastructure failure handling or retry policy;
- swallowing or replacing the original initialization exception.

## Suggested Regression Coverage

Add a focused lifecycle test covering both successful and failed infrastructure construction.

### Successful infrastructure creation

Verify:

```text
before construction:
    original server argument values

during infrastructure construction:
    disable_cuda_graph = True
    enable_return_hidden_states = True

before init_cuda_graphs():
    both values restored exactly

after bootstrap:
    both values remain restored
```

This preserves the existing #1380 lifecycle contract.

### Failed infrastructure creation

Make `create_sglang_infrastructure()` raise and verify:

```text
during construction:
    temporary values are active

after the exception:
    disable_cuda_graph restored
    enable_return_hidden_states restored
```

Also verify:

- the original exception propagates;
- `init_cuda_graphs()` is not called;
- restoration uses the original values rather than hard-coded defaults.

## Acceptance Criteria

- [ ] Temporary `disable_cuda_graph` and `enable_return_hidden_states` overrides are restored after successful infrastructure construction.
- [ ] Both values are also restored if `create_sglang_infrastructure()` raises.
- [ ] Restoration preserves the exact values that existed before the temporary override.
- [ ] `init_cuda_graphs()` still runs only after successful infrastructure creation and after argument restoration.
- [ ] `init_cuda_graphs()` is not called when infrastructure creation fails.
- [ ] The original initialization exception continues to propagate.
- [ ] Existing successful-path #1380 lifecycle behavior remains unchanged.
- [ ] No CUDA Graph eligibility or speech-output behavior is expanded by this fix.

## Related

- #1018
- #1022
- #1380
- #1381

charliechenye:
## Motivation

Qwen3-Omni speech Thinker defers CUDA Graph initialization while constructing the shared SGLang infrastructure.

During that construction, it temporarily overrides:

```text
disable_cuda_graph = True
enable_return_hidden_states = True
```

On the normal successful path, the original values are restored before deferred CUDA Graph initialization.

However, the restoration previously happened only after `create_sglang_infrastructure()` returned successfully. If infrastructure construction raised, the temporary values remained on the caller visible `server_args`.

This is an exception path lifecycle issue. It does not affect the normal successful serving path, but temporary bootstrap state should not escape the scope in which it is required.

## Modifications

Make the temporary Qwen3-Omni Thinker bootstrap overrides exception safe.

Infrastructure construction is now scoped with `try/finally`, so the original `disable_cuda_graph` and `enable_return_hidden_states` values are restored whether construction succeeds or raises.

The existing successful path ordering is preserved:

```text
apply temporary overrides
        ↓
construct SGLang infrastructure
        ↓
restore original arguments
        ↓
initialize CUDA Graphs
```

On infrastructure failure:

```text
apply temporary overrides
        ↓
construct SGLang infrastructure
        ↓
exception
        ↓
restore original arguments
        ↓
propagate the original exception
```

CUDA Graph initialization remains outside the cleanup path and therefore only runs after successful infrastructure construction.

Added a focused regression test that verifies:

- the temporary values are active during infrastructure construction;
- both values are restored after construction raises;
- the caller's original `enable_return_hidden_states` value is preserved for both `False` and `True`;
- the original infrastructure exception propagates unchanged.

No CUDA Graph eligibility, hidden state capture policy, or speech output behavior is changed.

## Related Issues

Fixes #1515

## Accuracy Test

Not applicable to model accuracy. This change only affects Qwen3-Omni bootstrap control flow and does not modify model computation, kernels, weights, sampling, or numerical outputs.

CPU validation on the exact candidate covered the load bearing lifecycle and surrounding bootstrap path:
https://github.com/charliechenye/sglang-omni/actions/runs/31765322058

```text
Focused deferred CUDA Graph lifecycle:
4 passed

Qwen3-Omni pipeline/bootstrap test file:
57 passed
```

The candidate also passed:

```text
pre-commit on changed files
git diff --check
```

## Benchmark & Profiling

Not applicable. This change does not modify the serving hot path or expected runtime performance.

## Checklist

- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed. No documentation change is required for this exception path fix.
- [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed. Not applicable because this change does not affect model numerics or performance.
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co author when merging the PR.

## CI


charliechenye:
cc: @edwingao28

edwingao28 (approved):

edwingao28:
lgtm

PR #1537, opened 2026-08-14
[Bug][Qwen3-Omni] Harden cached eager multimodal cursors 

### LINKED ISSUE #1514: [Bug][Qwen3-Omni] Harden multimodal cursor state across cached eager prefill
### Checklist

## Motivation

While auditing the Qwen3-Omni Thinker prefill path around #1381, we found two related correctness gaps in the inherited eager multimodal embedding merge.

Both involve the lifecycle of `req._omni_consumed`, which tracks how many embedding rows have already been consumed for each modality across chunked prefill.

The more important case is a reproducible cached-prefix correctness bug: when a radix-cached prefix already contains audio placeholders and the same request also carries visual state, eager fallback can select the wrong audio embedding row.

A second related case shows that malformed cursor state can reach the inherited eager merge without validation and fail later during embedding selection or scatter.

These appear to be two symptoms of the same missing contract around multimodal cursor initialization, preservation, advancement, and validation.

This issue is intentionally about **eager-path correctness**. It does not propose expanding Prefill CUDA Graph eligibility.

Related:

- #1018, Qwen3-Omni serving optimization roadmap
- #1022, Qwen3-Omni execution tracker
- #1150 / #1161, Thinker multimodal embedding merge work
- #1378 / #1381, Qwen3-Omni text-output Prefill CUDA Graph adoption

## Current Behavior

`ThinkerModelRunner` maintains request-local multimodal cursor state through:

```python
req._omni_consumed
```

Conceptually, it tracks how many embedding rows have already been consumed for each modality:

```python
{
    "audio": ...,
    "image": ...,
    "video": ...,
}
```

During eager multimodal injection, the cursor determines which source embedding rows correspond to the current prompt chunk.

The current path assumes this cursor is already synchronized with any prefix satisfied from the radix cache.

That assumption is not always true.

### Reproduction

Consider one request with:

```text
prompt audio positions: [0, 2]
prompt image position:   [3]

audio_embeds:
    row 0 = embedding for audio position 0
    row 1 = embedding for audio position 2

current prefix length: 2
current chunk starts at prompt position 2

req._omni_consumed = None
```

The radix prefix has already covered prompt position `0`, including the first audio placeholder.

The live chunk contains the second audio placeholder at prompt position `2`.

The correct source row is therefore:

```text
audio_embeds[1]
```

However, when the request also contains image/video or other visual state, the current bounded cached-audio repair does not initialize the audio cursor.

The inherited eager merge then begins from audio offset `0` and selects:

```text
audio_embeds[0]
```

for the live audio placeholder at prompt position `2`.

This silently associates the wrong encoder output with the prompt token.

### Expected behavior

When cached prompt positions and audio embedding rows establish an unambiguous consumed count, the audio cursor should reflect the number of audio placeholders already covered by the cached prefix.

Unrelated visual state should not cause an independently recoverable audio cursor to become incorrect.

This does **not** mean the visual request should become Prefill CUDA Graph eligible. The request can remain on the inherited eager path until the visual/deepstack Graph contract is separately qualified.

## Case 2: Malformed cursor state is not validated before eager injection

A related contract gap exists when `_omni_consumed` already exists but contains an invalid modality offset.

For example:

```text
audio embedding rows: 1
live audio positions:  1

req._omni_consumed = {
    "audio": 2
}
```

Graph-side qualification correctly treats this state as unsupported and delegates execution to the inherited eager path.

The eager merge can then attempt a source selection equivalent to:

```python
audio_embeds[2:3]
```

which contains zero rows, while the destination contains one live audio row.

A later tensor scatter therefore fails because the number of source and destination rows does not match.

Non-integral cursor values can similarly fail when used as tensor slice indices.

### Evidence limit

We have a deterministic reproduction when malformed internal cursor state is supplied.

We have **not established that normal scheduler execution can naturally produce an overlarge or non-integral `_omni_consumed` value**.

For that reason, this second case should be treated as a cursor-contract and defensive-correctness gap rather than a confirmed production failure mode.

## Underlying Contract Gap

The current implementation does not clearly distinguish among these states:

```text
None
    cursor has not been initialized

{}
    cursor object exists, but no modality cursor has been established

{"audio": N, ...}
    audio cursor exists and should be treated as authoritative
```

It is also not explicit which layer owns:

1. initializing a modality cursor after a cached prefix;
2. preserving an existing cursor object across chunks;
3. validating cursor types and bounds;
4. advancing the cursor after a successful embedding merge;
5. rejecting corrupted state before tensor operations.

This matters increasingly as radix caching, chunked prefill, multiple modalities, and future Prefill CUDA Graph adopters share the same Thinker embedding merge.

## Proposed Direction

I do not think this should be addressed through additional Graph-specific special cases.

Instead, the inherited eager multimodal merge should have an explicit cursor contract.

### 1. Cursor initialization

When cached prompt positions and embedding rows establish an unambiguous consumed count for a modality, a missing cursor may be initialized from that count.

For example:

```text
audio positions = [0, 2, 4]
prefix = 4

cached audio rows = 2
future audio rows remain

expected audio cursor = 2
```

Initialization should be per modality.

The presence of unrelated image/video state should not invalidate an otherwise unambiguous audio cursor.

### 2. Existing state remains authoritative

If an audio cursor already exists:

```python
{"audio": N}
```

it should not be silently recomputed or replaced.

If an existing cursor dictionary is missing the audio entry and an audio offset can be reconstructed safely, the existing dictionary should be updated in place so other modality state and object identity are preserved.

In particular, these states should not be treated as equivalent:

```python
None
```

and:

```python
{}
```

### 3. Validate cursor offsets before use

Before using a modality cursor as an embedding-row offset:

- the cursor should be an integer;
- `bool` should not be accepted as an integer cursor;
- the cursor should be non-negative;
- the selected source range should fit within the corresponding embedding rows;
- the number of selected source rows should match the number of live destination rows.

Corrupted internal state should fail with a clear error rather than being silently reset to zero, because resetting an invalid cursor can produce incorrect model outputs.

### 4. Keep normal cursor advancement in one place

The inherited multimodal merge should remain responsible for advancing cursors after successfully consuming embedding rows.

Any cached-prefix repair should only establish missing initial state.

It should not duplicate the normal per-chunk advancement logic.

### In scope

- cached/chunked eager multimodal cursor correctness;
- audio + visual requests remaining semantically correct on eager fallback;
- missing per-modality cursor initialization where the offset is unambiguous;
- preservation of existing cursor dictionaries and modality entries;
- validation of existing cursor offsets before embedding selection;
- regression tests using the real inherited multimodal merge.

### Out of scope

- enabling image/video/deepstack Prefill CUDA Graph replay;
- enabling speech-output Prefill CUDA Graph replay;
- changing CUDA Graph admission, buckets, padding, or replay policy;
- changing radix-cache policy;
- silently resetting malformed cursor state and continuing with guessed offsets;
- Qwen-specific scheduling behavior.

### Cached audio + image

Use a request equivalent to:

```text
audio positions = [0, 2]
image position  = [3]
prefix          = 2

audio_embeds:
    row 0 -> audio position 0
    row 1 -> audio position 2
```

Exercise the real inherited `_inject_multimodal_embeds()` path and verify that the live audio placeholder receives:

```text
audio_embeds[1]
```

not:

```text
audio_embeds[0]
```

### Existing empty cursor

Exercise a multi-step eager sequence where an earlier chunk leaves:

```python
req._omni_consumed == {}
```

and a later cached-prefix audio chunk still needs its audio cursor initialized.

Verify that:

- the existing dictionary object is retained;
- `"audio"` is added in place;
- other cursor entries are preserved;
- the correct audio embedding row is selected.

### Existing valid cursor

Verify that an existing cursor such as:

```python
{"audio": 1}
```

remains authoritative and is not replaced or recomputed.

### Malformed offsets

Cover at least:

```text
negative offset
non-integral offset
offset beyond available embedding rows
offset + live rows beyond available embedding rows
```

and verify that failure is explicit and occurs before an invalid tensor scatter.

### Mixed modality state

Verify that fixing audio cursor initialization does not:

- alter existing image/video cursor semantics;
- overwrite unrelated modality cursor entries;
- expand visual/deepstack Prefill CUDA Graph eligibility.

## Acceptance Criteria

- [ ] A cached prefix containing earlier audio placeholders cannot cause eager fallback to reuse those embedding rows for later live audio placeholders.
- [ ] Independently recoverable audio cursor state remains correct when the same request also carries visual/deepstack state.
- [ ] Existing cursor dictionaries are preserved rather than unnecessarily replaced.
- [ ] Existing valid modality cursors remain authoritative.
- [ ] Cursor offsets are validated before being used for embedding-row selection.
- [ ] Malformed internal cursor state fails explicitly instead of reaching an invalid tensor scatter or silently selecting a guessed row.
- [ ] Cursor advancement remains owned by the inherited multimodal merge.
- [ ] Regression tests exercise the real inherited eager path rather than mocking away `_inject_multimodal_embeds()`.
- [ ] No Prefill CUDA Graph eligibility is expanded as part of this correctness fix.

## Related

- #1018
- #1022
- #1150
- #1161
- #1378
- #1381

charliechenye:
## Motivation

Qwen3-Omni uses `req._omni_consumed` to track how many multimodal embedding rows have already been consumed across eager chunked prefill.

When a radix cached prefix already contains multimodal placeholders, the inherited eager merge currently assumes that this cursor is already synchronized with the cached prefix. That assumption does not always hold.

A concrete failure occurs when a cached prefix contains an earlier audio placeholder and the same request also carries visual state. The bounded Qwen specific repair introduced around the Prefill CUDA Graph path only handles a restricted audio payload. With visual state present, the request falls back to the inherited eager path without initializing the missing audio cursor. The eager merge can then select `audio_embeds[0]` for a later live audio placeholder that should use `audio_embeds[1]`.

There is a related cursor contract gap when `_omni_consumed` already exists but contains an invalid offset. Malformed values can reach embedding selection and fail later during tensor operations instead of being rejected at the cursor boundary.

This PR makes the inherited Thinker eager merge own the cursor contract. That is the common layer used by the Qwen3-Omni text eager fallback, and it is also used directly by the Qwen3-Omni speech Thinker path. The change does not expand Prefill CUDA Graph eligibility or change radix cache policy.

## Modifications

1. Add explicit cursor container handling in `ThinkerModelRunner`.

   `None` now means that no cursor dictionary has been initialized. An existing dictionary, including an empty dictionary, is preserved by identity. Invalid cursor containers fail explicitly.

2. Reconstruct missing per modality cursors when cached prompt state makes the consumed count unambiguous.

   For image, video, and audio, a missing cursor is reconstructed only when the radix prefix contains earlier placeholders and later placeholders remain live. Reconstruction requires prompt placeholder positions and source embedding rows to map one to one. Ambiguous state fails explicitly instead of guessing an offset.

3. Keep existing modality cursors authoritative.

   If a modality cursor already exists, it is validated and used as provided rather than silently recomputed from the current prefix.

4. Validate cursor offsets before embedding selection.

   Cursor offsets must be integral, must not be `bool`, must be nonnegative, and must select a source range that fits within the available embedding rows.

5. Keep normal cursor advancement in the inherited eager merge.

   After a live chunk consumes embedding rows, the same inherited merge advances `req._omni_consumed`. No separate repair path duplicates normal cursor advancement.

6. Remove the Qwen specific `_seed_cached_audio_eager_cursor()` workaround.

   Qwen3-Omni continues to own sidecar qualification and eager fallback selection, while the inherited Thinker merge owns eager cursor semantics.

7. Add deterministic regression coverage for:

   1. Cached audio plus image requests selecting the correct live embedding rows.
   2. Preservation of an existing empty cursor dictionary.
   3. Preservation of unrelated modality cursor state.
   4. Reconstruction across image, video, and audio.
   5. Existing cursor authority.
   6. Negative, nonintegral, boolean, beyond end, and live range overrun cursor values.
   7. Invalid cursor containers.
   8. Cached visual DeepStack alignment.
   9. Ambiguous reconstruction failure.
   10. The real Qwen inherited eager fallback path.
   11. The existing no host synchronization contract on the cached reconstruction path.

## Related Issues

Fixes #1514

Related context: #1381

## Accuracy Test

A standard model accuracy evaluation is not applicable to this change. The PR does not modify model weights, model architecture, kernels, sampling, or CUDA Graph eligibility. It corrects which already computed multimodal embedding row is selected when eager prefill resumes after a cached prefix, and adds validation for malformed internal cursor state.

CPU validation was run against the exact candidate:

`b3b6ffa6c5011726eb70bc1bf02bc5950b85a985`

Validation run:

https://github.com/charliechenye/sglang-omni/actions/runs/31774112478

Results:

```text
Shared Thinker multimodal merge:
33 passed

Qwen prefill and inherited eager fallback contract:
30 passed

Direct ThinkerModelRunner contract:
5 passed

Qwen3-Omni pipeline and bootstrap regression:
55 passed

CPU compatible Qwen3-Omni unit suite:
468 passed, 3 skipped, 2 deselected

Exact four file scope:
passed

Pre-commit on changed files:
passed

Committed diff and working tree sanity:
passed
```

The broad CPU suite intentionally excludes the two Qwen Talker test files whose collection requires CUDA loaded `sgl_kernel` operators and deselects two Code2Wav tests that explicitly require CUDA device normalization. All other CPU compatible Qwen3-Omni unit tests were included.

## Benchmark & Profiling

This PR is a correctness fix and is not expected to materially change serving performance.

The normal uncached prefill path exits cursor reconstruction before cached position filtering when the prefix length is zero. Cached reconstruction uses the existing CPU prompt position tensors and does not add `Tensor.item()`, `Tensor.any()`, `Tensor.nonzero()`, or `torch.where()` to the reconstruction path.

The existing no host synchronization regression tests, including the new cached prefix reconstruction case, pass in the shared Thinker multimodal merge suite.

No throughput or latency benchmark was run because this PR does not change CUDA Graph admission, graph buckets, replay policy, model computation, or scheduling behavior.

## Checklist

- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed. No documentation change is required for this cursor correctness fix.
- [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed. Not applicable because this is a correctness fix with no expected model accuracy or serving performance change. CPU correctness validation is provided above.

## CI

Independent CPU validation of the exact candidate passed:

https://github.com/charliechenye/sglang-omni/actions/runs/31774112478


charliechenye:
cc: @edwingao28

sglang_omni/model_runner/thinker_model_runner.py
```diff
@@ -116,6 +117,63 @@ def _plan_modality_chunk(
             relative_positions.numel(),
         )
 
+    @staticmethod
+    def _ensure_consumed_cursor(req: Any) -> dict[str, Any]:
+        consumed = getattr(req, "_omni_consumed", None)
```
edwingao28: nit: remove getattr

charliechenye: Fixed

edwingao28 (approved):

PR #1540, opened 2026-08-14
[Comm] Assign unique identities to stream relay transfers

charliechenye:
## Motivation

Stream relay transfers currently derive `DataRef.object_id` deterministically from the public request ID, stage edge, and chunk ID.

When the same request ID is reused on the same stream edge, this creates two failure modes:

1. If the previous transfer is still pending, the new transfer reuses the same `object_id` and is rejected as a duplicate pending transfer.
2. If the previous transfer has already timed out, a delayed ACK for that old transfer can match the reused `object_id` and incorrectly complete the newer transfer.

This PR addresses the stream transfer portion of #1307 item 3 by giving each `CommEngine` stream send attempt its own physical transfer identity, while keeping the public request ID and chunk ID unchanged.

PR #1487 separately introduces unique identities for `CommEngine` payload sends. This PR intentionally does not duplicate that payload work.

## Modifications

- Add a per `CommEngine` stream send sequence and include it in the stream transfer `object_id`.
- Pass the generated transfer identity into `StageIO.write_stream_chunk()`.
- Keep the existing deterministic `StageIO` stream ID as the fallback when `object_id` is not explicitly supplied, preserving direct low-level callers.
- Add regression coverage for:
  - two in flight stream sends with the same request ID, stage edge, and chunk ID;
  - independent ACK completion for those transfers;
  - metadata tensor IDs inheriting the unique parent transfer identity;
  - delayed stale ACKs after the previous transfer times out.

The regression tests exercise the real pending transfer and timeout lifecycle rather than manually mutating `_pending`.

### Verification

A baseline CPU run against the unmodified parent revision with only the new regression tests reproduces both target failures:

- concurrent reuse fails with `duplicate pending transfer 'req-reused:stream:producer:consumer:0'`;
- after the first transfer times out, its delayed ACK incorrectly resolves the newer transfer.

With this PR applied:

- `tests/unit_test/pipeline/test_comm_engine_ack.py`: **7 passed**
- `tests/unit_test/pipeline/test_stage_streaming.py`: **16 passed, 1 verification-only deselected**
- `tests/unit_test/pipeline/test_payload_device_restore.py`: **9 passed**
- `tests/unit_test/relay/test_shm_relay.py`: **1 passed**
- repository `lint.yaml`: **passed**

The one deselected stage streaming case fails identically on the unmodified parent revision in an entirely CPU environment after #994, so it is excluded only from the temporary fork CPU verification workflow. The repository test itself is unchanged.

I also validated this stream change on top of the current #1487 head. The combined CPU verification passed, including a temporary payload stale-ACK regression for #1487's payload identity implementation.

## Related Issues

Addresses the stream transfer portion of #1307 item 3.

Related: #1487, which independently covers unique payload transfer identities.

## Accuracy Test

Not applicable. This PR does not change model side code, kernels, model architecture, or model outputs.

## Benchmark & Profiling

Not applicable. This is a communication correctness fix and does not make throughput or latency claims.

## Checklist

- [x] Format your code according with pre commit.
- [x] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/comm/engine.py
+++ b/sglang_omni/comm/engine.py
@@ -6,6 +6,7 @@ import asyncio
 import logging
 from contextlib import suppress
 from dataclasses import dataclass
+from itertools import count
 from typing import Any, Callable
 from uuid import uuid4
 
@@ -121,6 +122,7 @@ class CommEngine:
         ] = {}
         self._send_workers: dict[str, asyncio.Task] = {}
         self._pending: dict[str, _PendingTransfer] = {}
+        self._stream_send_sequence = count()
         # Failed pending KV transfers stay pinned until this dying process exits.
         self._retained_pending_kv_transfers: list[_PendingTransfer] = []
         self._kv_pools: dict[str, KVPool] = {}
@@ -821,6 +823,10 @@ class CommEngine:
 
     async def _run_stream_send(self, job: _StreamSendJob, queue_key: str) -> None:
         object_id: str | None = None
+        stream_object_id = (
+            f"{job.request_id}:stream:{job.from_stage}:{job.target_stage}:"
+            f"{job.chunk_id}:{next(self._stream_send_sequence)}"
+        )
         send_start = _comm_now_ns()
         write_ms = -1.0
         control_ms = -1.0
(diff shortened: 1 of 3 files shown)


JingwenGu0829 (approved):
LGTM

PR #1541, opened 2026-08-14
[Qwen3-Omni Perf] Reduce realtime barge-in cancellation latency

RajPatel17:
## Motivation

Server VAD currently waits for the outbound `input_audio_buffer.speech_started`
WebSocket send before cancelling the active response. Under socket backpressure,
Thinker/Talker/Code2Wav can continue working after the server has already
detected the user's interruption.

## Modifications

- Mark and abort the active response before awaiting the outbound
  `speech_started` send.
- Gate only the matching cancellation terminal events until `speech_started`
  is delivered, preserving wire order.
- Scope cancellation gates to the exact pending/active request and snapshot
  reason/gate state under the response lock.
- Preserve deterministic first-server-processed cancellation semantics for
  `turn_detected` versus `client_cancelled`.
- Mark dequeued turns pending before waiting on current speech, so a stale
  queued answer is cancelled when new speech starts.
- Add regression coverage for send backpressure/failure, handler cancellation,
  repeated barge-in, cancellation ordering, terminal races, and queued-turn
  cancellation.

## Accuracy Test

Not applicable. This changes cancellation scheduling and event ordering only;
it does not modify model inputs, outputs, sampling, kernels, or architecture.

## Benchmark & Profiling

Measured on a pod with 2x NVIDIA H200 GPUs. The harness injected
controlled outbound `speech_started` send delay and measured server
VAD-handler entry to engine `abort()` invocation.

Two independent passes, 100 iterations per implementation and delay; values
below average the two median results:

| Injected send delay | Baseline abort | This PR abort | Delay removed |
|---:|---:|---:|---:|
| 0 ms | 0.0366 ms | 0.0325 ms | 0.0041 ms |
| 1 ms | 1.1115 ms | 0.0345 ms | 1.0770 ms |
| 5 ms | 5.1709 ms | 0.0531 ms | 5.1178 ms |
| 20 ms | 20.2151 ms | 0.0781 ms | 20.1370 ms |
| 50 ms | 50.2756 ms | 0.0943 ms | 50.1813 ms |

At 20 ms, average p95 improved from 20.2589 ms to 0.0872 ms. A swapped-order
verification (optimized measured first) reproduced 0.0834 ms versus
20.2350 ms. Socket-send completion remained at the injected delay in both
implementations, confirming the benchmark isolates abort scheduling rather
than bypassing the wire-order gate.

The change has no meaningful overhead without backpressure and removes
outbound socket latency from engine-abort initiation when barge-in occurs.

## Checklist

- [x] Add unit tests.
- [x] Provide throughput / latency benchmark results and accuracy evaluation
  results as needed.
- [x] Local code review completed.

Testing completed:

- 30 focused realtime serving tests passed.
- Race-heavy barge-in/terminal tests passed 25 consecutive runs.
- Multiple independent full-branch reviews completed with no remaining
  branch-introduced findings.
- Benchmark harness self-validation, two complete A/B passes, and a
  swapped-order verification passed on the 2xH200 pod.


edwingao28:
@RajPatel17 please fix lint

RajPatel17:
> @RajPatel17 please fix lint

Done! @edwingao28

sglang_omni/serve/realtime/session.py
```diff
@@ -521,19 +536,27 @@ async def emit_terminals(
...
+            *, terminal_gate: asyncio.Event | None = None, **kwargs: Any
+        ) -> None:
+            if terminal_gate is not None:
+                await terminal_gate.wait()
             terminal_task = asyncio.create_task(emit_terminals(**kwargs))
             try:
                 await asyncio.shield(terminal_task)
             except asyncio.CancelledError:
                 await terminal_task
                 raise
```
Peter-Nie2003, line 719: `emit_terminals_safely` is built so the client always gets a *complete* terminal
sequence: `shield` + `await terminal_task` make sure that once emission starts it
finishes even under a second cancel. This new `await` sits before that boundary,
and it can block for the full duration of a backpressured send. A second
`cancel()` landing here raises before `terminal_task` is ever created, so there is
no `response.done` at all, and a client that is discarding deltas "until
`response.done`" never gets released.

I traced it and don't think it's reachable today: `cancel_active_response`
early-returns for the same request, and `teardown()` only runs after
`session.run()` returns, by which point `handle_vad_emit`'s `finally` has already
set the gate. But both guarantees live outside this file, so dispatching VAD emits
as tasks — a natural follow-up here, since a backpressured send currently stalls
the whole receive loop — would make it real, with no failing test.

I'd suggest moving the wait inside the shielded task, less for the race than for
what the wait *is*. `claim_terminal()` has already decided we're emitting, so the
gate isn't a precondition on that decision; it's a constraint on the send itself,
which puts it inside the unit this function promises to complete. As written there
is a gap between deciding to emit and starting to emit, and closing that kind of
gap is the whole point of the function.

The cost is that cancellation would no longer break the wait, leaving that
`finally` as the only thing that can release the gate. I'd take that trade: a hung
teardown is at least observable, whereas a client that never receives
`response.done` fails silently. Does that match how you were thinking about it?

RajPatel17: Yes youre right, the guarantee should live inside the shielded region, not depend on cancel_active_response's early-return and the receive loop staying sequential. And I'd take the same trade: an observable stuck teardown beats a client silently hanging on response.done forever. Moving the wait inside the shield.

PR #1543, opened 2026-08-14
[Perf] Restore fast WAV duration probing for ASR

SandyLuXY:
## Motivation

Speech-to-text request validation probes the uploaded audio duration before dispatch. After duration probing moved to PyAV for broader container support, WAV uploads also paid the cost of opening the FFmpeg demuxer. This adds several milliseconds to every request and is especially visible for SeedTTS at low concurrency, even though WAV duration is available directly from the container header.

Restore the lightweight libsndfile metadata path for WAV while preserving PyAV coverage for other formats and malformed WAV inputs.

## Modifications

1. `sglang_omni/serve/speech_to_text.py`
   - Detect RIFF, RIFX, and RF64 containers whose form type is `WAVE`.
   - Read valid WAV duration with `soundfile.info` without decoding audio or opening FFmpeg.
   - Fall through to the existing PyAV probe for non-WAV inputs and when the WAV fast path cannot read the header.

2. `tests/unit_test/serve/test_speech_to_text.py`
   - Verify valid WAV input uses the soundfile path without calling PyAV.
   - Verify non-WAV input skips soundfile and continues through PyAV.
   - Verify an unreadable WAV header falls back to PyAV.

## Related Issues

#1396 

## Accuracy Test

**Setup:**

- Compared refs: c=1,2,4 use baseline `main @ 68abc7ee` and candidate
  `c0cd5d0d`; c=8,16 use baseline `main @ d0edde03` and candidate `8b5f7b54`.
- Model: `Qwen/Qwen3-ASR-1.7B` checkpoint `7278e1e70fe206f11671096ffdd38061171dd6e5`.
- Dataset: SeedTTS EN revision `27f4c1adee83b5b29b7c4b375f6b976324bda308`, 1,088 WAV clips.
- Hardware: NVIDIA H200, GPU 7. Each concurrency uses one discarded warmup and three measured full-corpus repeats.
- Candidate server: `--max-running-requests 64 --cuda-graph-max-bs 64 --mem-fraction-static 0.90` with FA3 and CUDA graphs enabled.

| Concurrency | Baseline corpus WER | Candidate corpus WER | Evaluated requests per variant |
| ---: | ---: | ---: | ---: |
| 1 | 1.218% | 1.203% | 3,264 / 3,264 |
| 2 | 1.218% | 1.218% | 3,264 / 3,264 |
| 4 | 1.218% | 1.213% | 3,264 / 3,264 |
| 8 | 1.218% | 1.218% | 3,264 / 3,264 |
| 16 | 1.218% | 1.218% | 3,264 / 3,264 |

All candidate requests completed successfully with no skipped samples.

## Benchmark & Profiling

Same model, dataset, hardware, and repeat configuration as described in the
_Accuracy Test_ section.

| Concurrency | Baseline throughput (mean +/- std) | Candidate throughput (mean +/- std) | Delta |
| ---: | ---: | ---: | ---: |
| 1 | 20.908 +/- 0.068 req/s | 24.687 +/- 0.215 req/s | +18.08% |
| 2 | 26.473 +/- 0.064 req/s | 33.449 +/- 1.073 req/s | +26.35% |
| 4 | 51.052 +/- 0.101 req/s | 66.298 +/- 0.446 req/s | +29.86% |
| 8 | 92.228 +/- 0.465 req/s | 117.020 +/- 1.602 req/s | +26.88% |
| 16 | 155.368 +/- 1.137 req/s | 175.982 +/- 2.648 req/s | +13.27% |

| Concurrency | Baseline mean latency | Candidate mean latency | Delta |
| ---: | ---: | ---: | ---: |
| 1 | 47.66 ms | 40.37 ms | -15.30% |
| 2 | 75.35 ms | 59.68 ms | -20.79% |
| 4 | 77.84 ms | 59.90 ms | -23.04% |
| 8 | 86.27 ms | 68.02 ms | -21.15% |
| 16 | 102.31 ms | 90.13 ms | -11.90% |

The production `asyncio.to_thread` duration probe microbenchmark dropped from 7.192 ms to 0.108 ms per cached SeedTTS WAV file, a 98.49% reduction (about 7.084 ms saved per request).

## Checklist

- [x] Add unit tests.

## CI

### DIFF
--- a/sglang_omni/serve/speech_to_text.py
+++ b/sglang_omni/serve/speech_to_text.py
@@ -220,7 +220,10 @@ def probe_audio_duration(audio_bytes: bytes) -> float:
             if info.samplerate:
                 return max(info.frames / float(info.samplerate), 0.0)
         except (RuntimeError, ValueError):
-            pass
+            logger.debug(
+                "WAV duration fast path failed; falling back to PyAV",
+                exc_info=True,
+            )
 
     try:
         import av
--- a/sglang_omni/serve/speech_to_text.py
+++ b/sglang_omni/serve/speech_to_text.py
@@ -200,11 +200,28 @@ def resolve_speech_to_text_adapter(
 def probe_audio_duration(audio_bytes: bytes) -> float:
     """Best-effort audio duration (seconds) from raw upload bytes.
 
-    Metadata-only: reads container headers, never decodes. PyAV wraps the
-    same FFmpeg demuxers load_audio decodes with, so any upload the engine
-    can read, this probe can measure (soundfile could not inspect M4A/WebM).
-    0.0 means unknown; callers treat that as "duration not available".
+    Metadata-only: reads container headers, never decodes. WAV uploads take a
+    libsndfile fast path; other formats and malformed WAV headers fall back to
+    the same PyAV/FFmpeg demuxers that ``load_audio`` uses. 0.0 means unknown;
+    callers treat that as "duration not available".
     """
+    # note (Xinyu): SeedTTS is entirely RIFF/WAVE, where opening FFmpeg through
+    # PyAV costs several milliseconds more per request than reading the header.
+    is_wav = (
+        len(audio_bytes) >= 12
+        and audio_bytes[:4] in {b"RIFF", b"RIFX", b"RF64"}
+        and audio_bytes[8:12] == b"WAVE"
+    )
+    if is_wav:
+        try:
+            import soundfile as sf
+
+            info = sf.info(io.BytesIO(audio_bytes))
+            if info.samplerate:
+                return max(info.frames / float(info.samplerate), 0.0)
+        except (RuntimeError, ValueError):
+            pass
+
     try:
         import av
 
--- a/tests/unit_test/serve/test_speech_to_text.py
+++ b/tests/unit_test/serve/test_speech_to_text.py
@@ -2,7 +2,9 @@
 
 from __future__ import annotations
 
+import io
 import json
+import wave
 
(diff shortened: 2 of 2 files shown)


sglang_omni/serve/speech_to_text.py
```diff
@@ -200,11 +200,31 @@ def resolve_speech_to_text_adapter(
...
+    is_wav = (
+        len(audio_bytes) >= 12
+        and audio_bytes[:4] in {b"RIFF", b"RIFX", b"RF64"}
+        and audio_bytes[8:12] == b"WAVE"
+    )
+    if is_wav:
+        try:
+            import soundfile as sf
+
+            info = sf.info(io.BytesIO(audio_bytes))
```
AkazaAkane, line 219: I think we need to gate this fast path on the WAV subtype, not only the RIFF/WAVE container magic. WAV can contain compressed codecs such as MPEG Layer III, for which libsndfile may estimate the frame count rather than report an exact sample count.

This matters here because probe_audio_duration() is used for the streaming duration limit and the needs_chunking / total-duration admission path, so an underestimated value can change request behavior rather than only usage.seconds.

AkazaAkane (commented):
This looks closely related to both #1031 and #1524. #1031 already established the guarded “WAV fast path + generic fallback” pattern for load_audio, while #1524 appears to target this exact duration-probe hotspot and includes additional format-correctness investigation. Could we reconcile these before merging so we don't end up maintaining two independently designed WAV fast-path policies?

0xjeffro (approved):

PR #1548, opened 2026-08-15
[Qwen3-Omni Perf] Remove the mm_aggregate hop from the speech pipeline

Peter-Nie2003:
## Motivation

The speech pipeline's `mm_aggregate` stage only relayed and joined encoder payloads. Letting `thinker` and `talker_ar` perform their own joins removes that process and relay hop.

## Modifications

What changed
Speech pipeline goes from 8 stages to 7.
```
before:  preprocessing ─┬─> image_encoder ─┐
                        ├─> audio_encoder ─┼─> mm_aggregate ─┬─> thinker
                        └──────────────────┘                 └─> talker_ar

after:   preprocessing ─┬─> image_encoder ─┬─> thinker
                        ├─> audio_encoder ─┤
                        └──────────────────┴─> talker_ar
```
- Rebuilt the speech-only topology change as `287976d6`: 8 stages become 7 and `mm_aggregate` is removed only from the speech path.
- `thinker` and `talker_ar` wait for preprocessing and encoder inputs directly; text and Ming-Omni paths retain their existing topology.
- Uses the stable `stages.thinker` fixture override rather than an obsolete numeric stage index.

## Accuracy Test

All paired runs used identical ordered generated-text cohorts and natural-EOS audio; every completed WAV was valid mono 16-bit / 24 kHz output.

## Benchmark & Profiling

Current-main H200 paired cold boots, SeedTTS natural-EOS voice cloning:

c（samples/boot） | QPS baseline → PR | mean TTFT | mean TTFA | E2E p95 Δ
-- | -- | -- | -- | --
1（32） | 1.715 → 1.761（+2.7%） | −8.9% | −5.0% | −4.1%
8（128） | 6.090 → 6.023（−1.1%） | −12.0% | −7.6% | +7.9%
16（256） | 7.598 → 7.694（+1.3%） | −20.0% | −8.3% | −1.3%
32（384） | 8.893 → 8.906（+0.15%） | −16.7% | −6.5% | +0.13%

All four cells completed with zero failures. P95s were mixed/noisy, so the claim is stable throughput and mean-latency improvement, not a precise tail-latency magnitude.

## Validation

- Official-image targeted suite: 131 passed.
- Exact 7-stage/no-`mm_aggregate` server attestation in both c32 candidate boots.
- Host ownership watchdog: no foreign GPU/port process; every server teardown drained to 0 MiB.

## Checklist

- [x] Add unit tests.

## CI

Please add the `run-ci` label for the rebuilt topology.


charliechenye:
Really interesting change. I like the direction of removing `mm_aggregate`, but I think there are a couple of consequences worth making explicit before we merge this.

My current mental model is that we are **not actually removing the synchronization barrier**. We are moving it from one shared stage into the two consumers.

Both `thinker` and `talker_ar` still use `wait_for=[preprocessing, image_encoder, audio_encoder]` together with `resolve_mm_aggregate_wait_sources`, so neither model should execute until the request specific active inputs are complete.

I think this is actually a cleaner dependency model. There is no fundamental reason that `thinker` readiness should depend on `talker_ar` having received its copy of the multimodal inputs. Moving the join into each consumer allows those paths to progress independently while preserving the fan-in semantics.

There are two behavior changes I would love to validate though.

### 1. `mm_aggregate` was also acting as a transport / tensor-lifetime boundary

The old speech aggregate had:

```python
disable_direct_cuda_ipc_payload=True
```

Removing the stage means encoder outputs can now go directly to `thinker` and `talker_ar`, including through direct CUDA IPC when placement allows it. That seems particularly interesting for the colocated topology, where encoder, thinker, and talker can all live on the same GPU but in different processes.

For `encoder -> thinker`, direct IPC seems unambiguously useful for TTFT.

For `encoder -> talker_ar`, I am a little less sure. The talker may retain the joined request while waiting for thinker chunks / `stream_done`. With direct CUDA IPC, that means we should make sure the lifetime of the encoder side CUDA storage does not become tied to outstanding talker requests in a way that noticeably increases retained GPU memory or fragmentation at concurrency.

I do not think this means we should keep `mm_aggregate`. If this turns out to matter, I would much rather eventually have a per edge transport policy than use an otherwise unnecessary stage as a copy boundary.

But I think a colocated memory stress test would be valuable here, especially with image/video inputs and concurrency.

### 2. Back pressure now reaches the encoders more directly

Previously the encoders handed their outputs to `mm_aggregate`, and the aggregate owned forwarding to thinker/talker. Now each encoder directly fans out to both consumers.

The current target order also means thinker is sent first, which is nice for TTFT. But if `talker_ar` becomes slow or its communication buffers become saturated, that pressure can now propagate back to the encoder instead of first accumulating behind the aggregate stage.

Again, I think this is probably the right trade. We are removing an entire process and relay hop. But it changes where buffering and back pressure live. It might be worth doing one stress test with an intentionally slow talker and checking:

* whether thinker TTFT remains isolated as expected
* how quickly pressure propagates back to image/audio encoders
* whether communication queues and GPU memory remain bounded

One other thing I checked was partial start. I initially worried that thinker could now start early enough that hidden-state chunks reach `talker_ar` before its multimodal payload is joined. But the existing `can_accept_stream_before_payload` + pending stream ingress path appears specifically designed to handle that case, so I do not currently see that as an architectural blocker. A topology level regression test for that ordering would still be nice.

Overall I am supportive of the change. I think removing a compute free synchronization process is the right direction, and consumer local joins are a cleaner abstraction.

The main thing I would want to establish is that we are intentionally changing **both** the synchronization topology and the transport / back pressure topology here. If the colocated memory lifetime test and slow consumer test look healthy, I would feel quite good about removing `mm_aggregate` from the speech path.

edwingao28 (approved):

PR #1554, opened 2026-08-15
[Perf][Qwen3-TTS] wire prefill admission coalescing through the build…

DerekSun02:
## Motivation

Qwen3-TTS is the only AR TTS model in the tree that cannot use prefill
admission coalescing. #1418 §6 lists this as a second-wave item, describing it
as a missing CLI whitelist entry. Tracing the plumbing, the gap is wider: the
parameters are absent at three of the four layers.

| Layer | Higgs TTS | Qwen3-TTS (before) |
| -- | -- | -- |
| OmniScheduler | `OmniScheduler.__init__` accepts and validates `prefill_coalesce_*` | same scheduler, already supported |
| EngineBuilder | `HiggsTtsEngineBuilder.__init__` + `extra_scheduler_kwargs()` | absent |
| stage factory | `higgs_tts.stages.create_sglang_tts_engine_executor` | absent |
| CLI whitelist | `_PREFILL_COALESCE_FACTORIES` in `cli/serve.py` | absent |

The generalized scheduler support from #1073 is reusable; what is missing is
the per-model wiring plus the CLI entry.

## Modifications

Wire the two existing coalescing parameters through the Qwen3-TTS build path,
mirroring the Higgs TTS pattern exactly.

**Wiring**

1. **`models/qwen3_tts/engine_builder.py`** — accept
   `prefill_coalesce_requests` / `prefill_coalesce_wait_ms` in `__init__`, store
   them, and forward them via `extra_scheduler_kwargs()` alongside the existing
   `stream_output_builder`.
2. **`models/qwen3_tts/stages.py`** — accept both parameters in
   `create_sglang_tts_engine_executor` and pass them to the builder.
3. **`cli/serve.py`** — add the Qwen3-TTS AR factory to
   `_PREFILL_COALESCE_FACTORIES` and add the model to the supported-models
   message.

**Tests**

4. **`tests/unit_test/qwen3_tts/test_pipeline.py`** — add four new unit tests, none
   requiring a GPU:
   - `test_qwen3_tts_prefill_coalescing_is_opt_in` — pins the factory signature
     defaults at `0` / `60.0`;
   - `test_qwen3_tts_factory_forwards_coalescing_params_to_builder` — ensure the stage
     factory actually hands both values to the builder;
   - `test_qwen3_tts_extra_scheduler_kwargs_keeps_stream_output_builder` —
     `extra_scheduler_kwargs()` carries the coalescing pair *and* the existing
     `stream_output_builder`, which Qwen3-TTS streaming depends on;
   - `test_qwen3_tts_coalescing_defaults_leave_gate_disabled` — a
     default-constructed builder hands the scheduler a disabled gate.
5. **`tests/unit_test/scheduling/test_cli_prefill_coalesce.py`** —
   `test_rejects_unsupported_pipeline` used `Qwen3TTSPipelineConfig` as its
   unsupported example, which this PR necessarily invalidates. Swapped it for
   `LLaDA2UniPipelineConfig`, whose thinker runs on `DllmScheduler` — a class
   that only mirrors the `OmniScheduler` stage interface and has no prefill
   admission gate to tune — so it remains a valid negative example. *Happy to pick a different substitute if you have a
   preference.* Also added Qwen3-TTS to the positive
   `test_cli_sets_coalesce_args` matrix, which is the only test covering the
   whitelist entry itself. 

**Docs**

6. **`docs/cookbook/qwen3_tts.md`** — new "Prefill Admission Coalescing"
   section: CLI and per-stage YAML usage, the gate's release conditions (idle
   bypass, queue target, wait deadline), why `prefill_coalesce_wait_ms` is an
   upper bound rather than a typical wait, and the `tp_size > 1` force-disable.

No scheduler changes: `OmniScheduler` already validates and consumes both
parameters generically.

## Accuracy Test

With the gate disabled — the default — no request path changes, so accuracy is
unchanged by construction. For the enabled configurations, corpus WER is
reported as a quality gate alongside the throughput numbers below; coalescing
only reorders admission and must not move it beyond run-to-run noise.

## Benchmark & Profiling
**Setup:**  1× H200, Qwen/Qwen3-TTS-12Hz-1.7B-Base, SeedTTS EN-1088, tp_size=1, streaming mode.

#### K=4, T=60 ms was compared against K=0 with interleaved ABABAB runs.
| round | K=0 QPS | K=4,T=60 QPS | K=0 TTFC p95 | K=4,T=60 TTFC p95 |
|---|---:|---:|---:|---:|
| 1 | 7.560 | 7.281 | 720.7 ms | 689.5 ms |
| 2 | 7.964 | 7.645 | 680.0 ms | 723.0 ms |
| 3 | 6.835 | 7.067 | 772.2 ms | 784.7 ms |
| **mean** | **7.453** | **7.331** | **724.3 ms** | **732.4 ms** |

Mean QPS changed by -1.6%, while mean TTFC p95 changed by +1.1%. 

#### K=2,T=30 ms was compared against K=0 using fresh interleaved ABABAB runs.

| round | K=0 QPS | K=2,T=30 QPS | K=0 TTFC p95 | K=2,T=30 TTFC p95 |
|---|---:|---:|---:|---:|
| 1 | 6.939 | 7.130 | 726.0 ms | 725.5 ms |
| 2 | 6.701 | 7.061 | 719.2 ms | 693.7 ms |
| 3 | 6.730 | 6.961 | 711.9 ms | 703.0 ms |
| **mean** | **6.790** | **7.051** | **719.0 ms** | **707.4 ms** |

Mean QPS increased from 6.79 to 7.05 req/s (+3.8%), while streaming TTFC p95 decreased from 719ms to 707ms (-1.6%). All headline runs were performed with profiling disabled.

#### K=32, T=300 ms was compared against K=0 for three interleaved rounds. 
| round | K=0 QPS | K=32,T=300 QPS | K=0 TTFC p95 | K=32,T=300 TTFC p95 |
|---|---:|---:|---:|---:|
| 1 | 7.366 | 7.705 | 6.822 s | 7.809 s |
| 2 | 6.975 | 7.088 | 7.137 s | 7.663 s |
| 3 | 7.282 | 7.052 | 7.148 s | 7.155 s |
| **mean** | **7.208** | **7.282** | **7.036 s** | **7.542 s** |

Mean QPS changed by +1.0%, while mean TTFC p95 increased by +7.2%. No repeatable throughput gain was observed.

### Conclusion
Prefill admission coalescing can improve end-to-end throughput, but only when the batching benefit is large enough to outweigh the decode-side cost introduced by delaying admission.

The c16 profiling results suggest that coalescing is most useful when:

- Natural prefill batches are small;
- Prefill accounts for a meaningful fraction of scheduler work;
- A small K can materially increase realized prefill batch size;
- The wait window is short enough that decode occupancy and generation throughput remain nearly unchanged.

At the shipped c16/cap16 configuration, the profiled baseline had only 1.35 req/prefill batch and approximately 22% reconstructed prefill-side wall share. A light K=2,T=30 ms setting reduced reconstructed prefill work by approximately 15.5% while reducing measured generation throughput by only 1.2%, and the subsequent no-profile validation showed a repeatable +3.8% mean QPS improvement with no TTFC regression.

More aggressive coalescing did not behave the same way. Increasing the wait window to 60 ms, or using larger K/T settings such as K=4,T=60 ms, either reduced decode occupancy enough to offset the saved prefill work or increased first-audio latency without a repeatable throughput gain.

**Therefore the feature should remain disabled by default / opt-in.** The profiling results provide a practical way to decide when enabling it is worthwhile rather than assuming that larger prefill batches always improve end-to-end throughput.

### Environment

GPU: NVIDIA H200
Driver: 580.159.04
CUDA: 13.0
PyTorch: 2.11.0+cu130

Branch: 750eb7f0
Base: bf289b79
Model: Qwen/Qwen3-TTS-12Hz-1.7B-Base
Dataset: Seed-TTS EN-1088

## Related Issues

Part of #1418 (§6 second-wave backlog).

> Prefill admission coalescing (https://github.com/sgl-project/sglang-omni/pull/1073)
> is generalized in the scheduler but qwen3_tts is missing from the CLI factory
> whitelist (cli/serve.py:47-58); adding it trades TTFB for throughput, so the
> first-audio-latency gate needs recalibrating alongside.

## Checklist

- [x] Add unit tests.


JiaxinD:
Maintainer refresh on current main is now pushed at a1bdee55.

The implementation is still opt-in (requests=0 by default), but the old CLI-whitelist path no longer exists after the typed config migration. The refreshed branch now:
- accepts the two knobs on the Qwen3-TTS AR factory and forwards them through Qwen3TtsEngineBuilder to OmniScheduler;
- validates dotted CLI/YAML values through FactoryArgs and the current resolver;
- documents --tts_engine.factory.prefill_coalesce_* and stages.tts_engine.factory.* instead of the removed global flags/runtime_overrides shape;
- keeps the existing H200 c16/cap16 K=2,T=30 ms evidence as the deployment-value gate.

Local pre-commit passed. Python compilation passed. The focused pytest collection is unavailable in this Windows environment because sglang is not installed, so the exact-head CI is required before approval.

JiaxinD (approved):
Verified at current head 0103a88f: the opt-in prefill coalescing wiring uses the correct dotted factory path and preserves defaults. Exact-head unit/accelerator, both ASR stages, and all five Qwen3-TTS stages pass, including serving stress and MPS. Approved.

JiaxinD:
@DerekSun02 I want to acknowledge a process mistake on my side. While reviewing this PR, I used maintainer edit access to push directly to your branch without coordinating with you first. I should have prepared the update on my fork and offered it as a patch or cherry-pick instead. Sorry about that.

The branch moved from `b8ffd438` to `0103a88f`. I added the main-sync/conflict-resolution commits `a1bdee55` and `0103a88f`, adapting the opt-in prefill-coalescing wiring to the current typed/dotted configuration path. I did not force-push, so your original head remains in the history. I believe the update is useful, but whether to keep it is your decision. I also approved the resulting revision, which was not an independent review boundary after I had modified the branch.

I will not push to this branch again. If you prefer the original branch state, please say so; I can prepare a fork-based patch or restoration instructions, but I will not change the branch without your explicit request.

DerekSun02:
> @DerekSun02 I want to acknowledge a process mistake on my side. While reviewing this PR, I used maintainer edit access to push directly to your branch without coordinating with you first. I should have prepared the update on my fork and offered it as a patch or cherry-pick instead. Sorry about that.
> 
> The branch moved from `b8ffd438` to `0103a88f`. I added the main-sync/conflict-resolution commits `a1bdee55` and `0103a88f`, adapting the opt-in prefill-coalescing wiring to the current typed/dotted configuration path. I did not force-push, so your original head remains in the history. I believe the update is useful, but whether to keep it is your decision. I also approved the resulting revision, which was not an independent review boundary after I had modified the branch.
> 
> I will not push to this branch again. If you prefer the original branch state, please say so; I can prepare a fork-based patch or restoration instructions, but I will not change the branch without your explicit request.

@JiaxinD
No problem at all — thanks for flagging it explicitly, and thanks for doing the
adaptation.

**Keeping the update.** I'd rather keep `0103a88f` than restore `b8ffd438`. The
config system moved out from under this PR while it was open — the
`_PREFILL_COALESCE_FACTORIES` whitelist I originally edited in `cli/serve.py`
no longer exists there. Reverting would just mean redoing that same migration.

PR #1558, opened 2026-08-15
fix(audio): preserve stereo in non-streaming encoded outputs

### LINKED ISSUE #1549: Compressed output formats are hardcoded to mono, downmixing stereo music models
### What happens

MiniMax Music 3 produces 32 kHz **stereo** audio. Requested with
`response_format: "wav"` the response is stereo, as expected. With `"mp3"`,
`"flac"`, `"opus"` or `"aac"` the same request returns **mono** — one channel of
a two-channel signal, silently downmixed.

### Where it comes from

In `sglang_omni/client/audio.py`, the compressed-format encoder pins the layout
(lines 252 and 256):

```python
stream.layout = "mono"
...
audio.reshape(1, -1), format="fltp", layout="mono"
```

and multi-channel input is averaged before it gets there (lines 291–295):

```python
# Streaming chunks are mono; downmix multi-channel payloads
channel_axis = 0 if audio.shape[0] < audio.shape[-1] else -1
audio = audio.mean(axis=channel_axis).astype("float32")
```

`encode_wav` in the same file handles two channels correctly
(`num_channels = int(pcm.shape[0])`, line 168), so the limitation is specific to
the compressed path.

The comment names the origin, and it is a fair one: this path was written for
speech streaming, where mono is the norm. It predates the arrival of a music
model whose output is stereo by design.

### Why it matters

For TTS and ASR the behaviour is harmless. For MiniMax Music 3 it means the most
convenient output format discards half the signal — and does so without a
warning, so it is easy to ship a pipeline that quietly loses the stereo image.
The workaround is to always request WAV and re-encode afterwards, which is fine
for a script and awkward for a browser client that just wants a small file.

### Steps to reproduce

```bash
sgl-omni serve --model-path MiniMaxAI/MiniMax-Music3 --port 8000

for f in wav mp3 flac opus; do
  curl -s http://127.0.0.1:8000/v1/audio/speech \
    -d "{\"model\":\"MiniMaxAI/MiniMax-Music3\",
         \"input\":\"[Verse]\\nTest\",
         \"instructions\":\"An energetic arena rock anthem at 130 BPM\",
         \"response_format\":\"$f\",\"max_new_tokens\":250}" -o out.$f
  echo -n "$f: "; ffprobe -v error -show_entries stream=channels -of csv=p=0 out.$f
done
```

Observed: `wav: 2`, `mp3: 1`, `flac: 1`, `opus: 1`.

### Suggested fix

Derive the channel count from the payload instead of pinning it — set
`stream.layout` to `"stereo"` when the array has two channels, and keep the
downmix only for the streaming path that genuinely needs it. If the downmix has
to stay for other reasons, logging it once per request would already help,
because the current failure is silent.

### Environment

- `sglang-omni` from `main`, commit `68abc7ee`
- `sglang` 0.5.16, torch 2.11.0+cu130
- Base image `lmsysorg/sglang:v0.5.16`, **aarch64**
- NVIDIA GB10 (DGX Spark), sm_120, CUDA 13.0
- Model: `MiniMaxAI/MiniMax-Music3`

charliechenye:
## Motivation

Fixes #1549.

Stereo audio currently survives model inference and client reconstruction, but is lost during final response encoding for compressed formats.

The shared `encode_audio()` path averages multi-channel audio to mono for every non-WAV format. PyAV then independently assumes a mono stream layout and constructs a mono audio frame. As a result, models that natively produce stereo audio can return stereo WAV output while MP3, FLAC, Opus, and AAC responses are collapsed to mono.

This is particularly visible for music generation models such as MiniMax Music 3, where the two output channels are part of the generated audio rather than redundant copies.

This PR preserves native stereo for self-describing non-streaming output formats while intentionally leaving the existing mono streaming contract unchanged.

## Modifications

1. Preserve two-channel audio through MP3, FLAC, Opus, and AAC encoding instead of averaging it to mono.

2. Make the PyAV encoder derive its layout from the normalized audio shape. Mono input uses `mono`, while stereo input uses `stereo`. Audio frames remain planar float32 with the existing channel-first internal representation.

3. Make `_resample_linear()` operate on the sample axis so each stereo channel is resampled independently. This matters for inputs such as 32 kHz MiniMax audio when Opus selects an existing supported encoder input rate.

4. Convert channel-first stereo into the frame-first representation expected by SoundFile only at the FLAC boundary.

5. Preserve the existing behavior for WAV, raw PCM, mono audio, and unsupported channel layouts.

6. Add deterministic CPU regression coverage using different signals for the two channels: 440 Hz on the left and 880 Hz on the right. The tests decode the encoded result and verify that the two channels retain distinct content, rather than checking channel metadata alone.

7. Add direct PyAV tests for MP3, AAC, and Opus so failures in the primary PyAV path cannot be hidden by the existing pydub fallback.

8. Cover both supported stereo input orientations, `[channels, samples]` and `[samples, channels]`.

The streaming path is intentionally outside the scope of this change. `select_audio_delta()` and the existing mono raw PCM streaming behavior are unchanged.

## Related Issues

Fixes #1549.

## Accuracy Test

This change does not modify model inference, model architecture, kernels, or model output generation. The regression is isolated to the response serialization layer and can be tested deterministically with synthetic audio on CPU.

Targeted audio tests: **21 passed, 0 failed, 0 skipped.**

The coverage verifies:

1. MP3 preserves distinct left and right channel content.
2. AAC preserves distinct left and right channel content.
3. Opus preserves distinct left and right channel content.
4. WAV continues to preserve stereo as the control case.
5. FLAC preserves stereo for channel-first input.
6. FLAC preserves stereo for channel-last input.
7. Mono FLAC behavior remains unchanged.
8. The 32 kHz to 24 kHz resampling path preserves stereo channels independently.
9. Direct PyAV encoding preserves stereo without relying on fallback behavior.

I also performed mutation checks against the regression suite.

Restoring the original generic compressed-format downmix caused the MP3, AAC, Opus, and FLAC stereo tests to fail.

Restoring the hardcoded PyAV mono behavior caused the direct MP3, AAC, and Opus tests to fail.

This confirms that the new tests directly protect the two independent causes of the reported regression.

A broader speech and OpenAI API test run produced 152 passes and one local environment failure in `test_long_m4a_is_probed_and_chunked`. I reproduced the same failure with the same Python environment on upstream `main`. It occurs while torchcodec attempts to load missing local FFmpeg libraries, before the client audio serialization path exercised by this PR.

No model weights or GPU execution are required to validate this fix.

## Benchmark & Profiling

Not applicable as a performance optimization.

This change does not modify model execution or the serving pipeline before response encoding.

Stereo compressed responses can contain more encoded audio data than the previous mono output because both generated channels are now retained. That is expected behavior for this correctness fix.

## Checklist

- [x] Code formatted with pre-commit.
- [x] Unit tests added for the affected encoding paths.
- [x] Existing mono behavior preserved and covered.
- [x] WAV behavior preserved.
- [x] Raw PCM behavior preserved.
- [x] Streaming semantics intentionally unchanged.
- [x] Direct PyAV tests prevent fallback behavior from masking regressions.
- [x] No MiniMax-specific logic introduced.
- [x] No codec rate-selection policy changed.
- [x] No GPU validation required for this serializer-only fix.
- [x] Documentation changes are not required because this restores the channel layout already produced by the model rather than introducing a new API option.

## CI

The branch contains one focused commit on current upstream `main`.

Local targeted validation and pre-commit checks pass. Full repository CI can run through the standard maintainer CI workflow.


zhaochenyang20 (approved):

MvdB:
thx @charliechenye

PR #1559, opened 2026-08-15
[Model] Fix MiniMax Music 3 loading and BF16 rotary precision

Ratish1:
## Motivation

MiniMax Music 3 represents one logical music request as two adjacent SGLang requests: a conditioned row and a CFG-unconditioned row. SGLang normally admits and retracts physical rows independently, which could split this pair under prefill limits or KV-cache pressure. A retracted request also could not resume because its generated audio-code embeddings were not retained for replay.

The original integration also rewrote the downloaded checkpoint configuration and globally patched SGLang's Qwen3 loader, even though the released checkpoint contains a native Qwen3 `language_model`.

## Modifications

- Load the released `language_model` directly as the Qwen3 AR backbone. The audio embedding and RVQ decoder continue to load from the checkpoint's audio components. This removes checkpoint mutation, `.bak` creation, the architecture override, and the process-global Qwen3 loader patch.
- Treat every conditioned/unconditioned CFG pair as one scheduling unit. Prefill admission accounts for both rows against request, prefill-token, and KV-token limits; decode retraction releases and requeues complete pairs using SGLang's existing KV and scheduler accounting.
- Retain each generated eight-codebook frame so a retracted request can reconstruct its audio embeddings when SGLang sends it through prefill again.
- Interpret `prefill_max_requests` and `max_queued_requests` as logical music-request limits and convert them once to physical CFG-row limits.
- Force `disable_overlap_schedule=true` after applying overrides because MiniMax's synchronous generated-code feedback does not support overlap scheduling.
- Keep only DIT rotary inverse-frequency and phase construction in FP32 for BF16 acoustic inference. Cosine and sine are cached and cast to the activation dtype before attention; the DIT, selected SGLang attention backend, compilation, Cache-DiT, and CUDA-graph paths are unchanged.
- Replace fallback access for required Qwen3, schedule-batch, feedback-buffer, and RVQ-graph state with direct access so lifecycle violations fail at the owning component.

## Root Cause

SGLang schedules physical request rows independently, while MiniMax Music 3 requires every lifecycle transition to preserve an adjacent conditioned/unconditioned pair. Row-wise admission could admit only one side, and row-wise decode retraction could leave an odd running batch. Both paths were reproduced on H100 and caused the MiniMax runner to reject the batch.

Retraction returns generated requests to prefill. The previous runner accepted only the original prompt because it did not retain the generated audio-code history required to rebuild the replay embeddings.

## Accuracy Tests

H100 qualification was run through commit `59719617546dfb4890308624918502adcf39554f` using the released MiniMax Music 3 checkpoint.

- All 399 native Qwen3 backbone tensors matched the corresponding legacy backbone tensors exactly in name, shape, dtype, and value.
- Fixed FP32 A/B comparison matched 7/7 AR hidden chunks, all adjacent overlap states, 252/252 captured DIT solver/latent/overlap/DAV tensors, and the final WAV exactly.
- The complete single-GPU and dual-GPU cookbook/API matrix returned valid finite 32 kHz stereo audio. Fixed-input default-FP32 outputs matched the baseline exactly, and the rerun passed manual listening.
- Default 16-request and explicit 32-request admission completed without OOM.
- Exact partial-admission, natural KV-pressure retraction, forced retraction, public retract/continue, and late replay paths completed without CFG-pair or replay errors.
- With `prefill_max_requests=1`, both physical CFG rows were admitted and the request returned a valid 9.996-second WAV.
- With `max_queued_requests=1`, both rows survived pause/retract/continue, the request completed, and all scheduler queues drained to zero.
- MiniMax Music 3 unit tests: 54 passed.
- BF16 acoustic inference returned valid finite stereo audio with FP32 rotary phase construction. The default FP32 path remained numerically exact against the baseline.

## Speed Tests and Profiling

No performance improvement is claimed, and no repository benchmark was run. H100 A/B timing was used only as a regression check: fixed-work c16 measurements across the commit ladder stayed within a 0.36% wall-time range, and c1 showed no persistent latency step attributable to a scheduler commit.

## Checklist

- [x] Format code according to the contribution guide.
- [x] Add or update tests.
- [x] Validate model accuracy and scheduler correctness on H100.
- [ ] Provide a repository throughput benchmark.


sglang_omni/models/minimax_music3/scheduler.py
```diff
@@ -48,33 +63,234 @@ def _enqueue_cfg_uncond(self, req_data: Any, uncond: Any) -> None:
         self.waiting_queue.append(req)
 
     def get_new_batch_prefill(self, running_batch: Any) -> Any:
+        self._reject_unadmittable_pairs()
         queue = self.waiting_queue
+        prefill_budget = self.max_prefill_tokens
+        expanded_pair_budget = False
+        if len(queue) >= 2:
+            assert queue[0]._omni_data.cfg_uncond is queue[1]._omni_data
```
JiaxinD, line 71: `queue` is aliased from `self.waiting_queue` without `_request_admission_lock`, but `abort()` runs on the stage abort listener thread and sets `_omni_data = None` on the rows it drops, so a cancel landing between line 67 and here makes this `None.cfg_uncond` and takes down the scheduler thread (`_event_loop_normal` has no try). `_is_cfg_uncond` below already guards with `getattr(req, "_omni_data", None)`.

sglang_omni/models/minimax_music3/scheduler.py
```diff
@@ -48,33 +63,234 @@ def _enqueue_cfg_uncond(self, req_data: Any, uncond: Any) -> None:
...
+            reserved_tokens += (
+                paged_input + min(max_new, CLIP_MAX_NEW_TOKENS) + self.page_size
+            )
+        return _PairPrefillBudget(input_tokens, required_tokens, reserved_tokens)
+
+    def _reject_unadmittable_pairs(self) -> None:
+        while len(self.waiting_queue) >= 2:
+            pair = self.waiting_queue[:2]
+            budget = self._pair_prefill_budget(pair)
+            if budget.required_tokens < self.max_total_num_tokens:
```
JiaxinD, line 153: `required_tokens` counts `len(output_ids)`, so a pair this scheduler just retracted and requeued can fail this whole pool gate and get aborted mid generation with "reduce the request length". Consider skipping the check when `req.is_retracted`, or computing it on `origin_input_ids` only.

sglang_omni/models/minimax_music3/scheduler.py
```diff
@@ -48,33 +63,234 @@ def _enqueue_cfg_uncond(self, req_data: Any, uncond: Any) -> None:
...
+    def update_running_batch(self, batch: Any) -> Any:
+        """Apply SGLang's decode update to complete CFG pairs."""
+        initial_size = len(batch.reqs)
+        batch.filter_batch()
+        if batch.is_empty():
+            batch.batch_is_full = False
+            return batch
+
+        if self.enable_hierarchical_cache:
+            self.tree_cache.flush_write_through_acks()
```
JiaxinD, line 177: `flush_write_through_acks` does not exist on any sglang 0.5.19 tree cache. Unreachable today because `OmniScheduler` pins `enable_hierarchical_cache = False`, but it becomes an AttributeError the moment that changes.

sglang_omni/models/minimax_music3/checkpoint.py
```diff
@@ -48,6 +49,7 @@ def resolve_checkpoint(model_path: str | Path) -> TTMCheckpointPaths:
     tokenizer_dir = root / "qwen_7B" / "qwen3-8B-tokenizer-music"
     paths = TTMCheckpointPaths(
         root=root,
+        language_model_dir=root / "language_model",
```
JiaxinD, line 52: The discovery loop above only validates `qwen_7B/qwen_7B` and `flowmatching_vae.pth`, so a snapshot without `language_model/` resolves fine here and then fails as a bare HuggingFace "no config.json" from inside `ServerArgs`. Adding it to the candidate check keeps the error local.

PR #1567, opened 2026-08-16
[Qwen3-Omni Perf] Overlap Code2Wav output materialization with vocoder launches

EdwardZhang1108:
## Motivation

On the production default (Code2Wav CUDA graph on, batching off), each threshold window's waveform readback runs synchronously on the scheduler thread between two vocoder launches: a pageable blocking `.cpu()` that waits out the window's GPU compute, then CPU bf16→fp32, numpy materialization, and serialization. Separately, the talker emits one frame per decode step, so the codec EOS check costs one `.item()` sync per frame — ten per window at the default `stream_chunk_size`.

## Modifications

- **Depth-2 output pipeline** (CUDA serial path). Window N's replay is followed on the same stream by an fp32 convert and async D2H into a pool-owned pinned slot plus a CUDA event. Same-stream enqueue preserves the CUDA-graph borrowed-output contract — replay N+1 cannot overwrite the static buffer before the copy drains.
- **Drained windows are emitted on the observing frame.** `_decode_and_emit` runs once per talker frame (~5.7 ms) rather than once per window (~57 ms), so a completed copy is flushed by the first frame that observes `Event.query()` true. Holding it until the next window was dispatched traded an 8.3 ms stall for a whole window arrival — see the ITL results below.
- **First window per request and final tails stay synchronous**, so TTFA is structurally unchanged.
- **`on_stream_done` pre-flushes** the pending window as its own stream message before the base runs the tail decode, keeping message boundaries identical.
- **Lazy batched EOS scan.** Frames stage unchecked; `should_decode` scans the staged tail in one batched compare when the raw ready count crosses the threshold. Scanning before the gate matters — a staged EOS would otherwise inflate the count and fire a short, graph-key-missing window. One sync per window instead of ten. Batching mode keeps the eager per-frame check.
- **Single-owner slot lifecycle.** Every slot is in exactly one of free / retired / quarantined. Abort hands its slot to the retired list without blocking — it arrives on the stage event-loop thread, which must not wait on CUDA — and the scheduler thread reaps completed retirements with non-blocking `Event.query()`. A copy or record failure quarantines the buffer and disables the pipeline rather than recycling memory whose D2H completion was never proven; quarantined slots still count against the cap. Pool exhaustion flushes the request's own pending window and reuses that slot, preserving emission order.
- **Instrumentation**: `code2wav_decode_launched` on pipelined windows, plus `pipelined` / `d2h_wait_ns` on `code2wav_decode_end`. Sync-path event shapes unchanged.
- **Kill switch**: `enable_output_overlap` factory arg, default on; `runtime_overrides: code2wav: enable_output_overlap: false`.

`StreamingVocoderBase`, `run_step` (batching, #1237), the #1355 trim, and `audio_waveform_payload` are untouched; all production changes are in `code2wav_scheduler.py`.

## Related Issues

Part of #1022. Measurement contract: #1018 (item 4.5).

Review and fixes by @edwingao28, who caught the inter-chunk latency regression described below and pushed `67883229` (single-owner slot lifecycle, non-blocking abort, `ndim == 1` EOS fast-path guard), `608294ee` (profiler and borrowed-output contract wording), `0a427297` (a release-path assertion the original CPU fakes could not make), and `c4ab8e8` (emit on the observing frame).

## Inter-chunk latency — the cost of deferring, and the fix

**Expected:** none. This section exists because the first version of this PR had no such expectation and was wrong.

*What / why:* deferring a window's emission necessarily delays every chunk by one production interval. Frames arrive every ~5.7 ms but windows only every ~57 ms, so the original design — flush window N−1 when window N is dispatched — bought an 8.3 ms scheduler stall at the price of a whole window arrival. Thread occupancy, launch gap, TTFA and throughput all improved while the metric the deferral actually moved was never measured.

*Setup:* @edwingao28 found this and measured it: ITL p95 against `enable_output_overlap: false`, 2× H200 TP2, ABC/CBA ordering, 2 boots per arm, 3 runs, 12/12 cells reproduced. `floor` is the within-arm spread.

| c | Deferred to next window | Emit on observing frame | Floor |
| --- | --- | --- | --- |
| 1 | **+49.5%** | −14.4% | 26.5% |
| 8 | **+40.5%** | −7.2% | 12.5% |
| 16 | **+39.2%** | +9.5% | 13.1% |
| 32 | **+28.1%** | −5.7% | 12.7% |

*Result:* p99 has the same shape; TTFT, TTFA, E2E and throughput sit inside the floor for both arms. The fix (`c4ab8e8`) moves the flush to `_decode_and_emit`, which runs per frame, gated on a non-blocking `Event.query()`. Scheduler-thread occupancy is unaffected: 8.72 → 0.46 → 0.43 ms p50 at c1 across baseline, deferred, and fixed.

*Caveat:* measured at ~3.4 s of audio per request; talker output length could not be pinned, so this covers short-form generation only.

## Correctness validation

**Expected:** behavior-preserving — the pipeline changes when the waveform is materialized, not what it contains.

Evidence is organized per the 2026-08-15 amendment to #1018 ("Output Comparison And Determinism"): neutrality is asserted at the text-token level, waveform comparison is supporting evidence only, and WER is the primary audio-quality instrument. Codec-token comparison does not apply — Code2Wav consumes codec tokens and cannot produce them, and this PR does not touch the talker.

### Output neutrality at the text-token level

*What / why:* the amendment makes this the load-bearing neutrality assertion, since waveform bytes are no longer admissible as a gate.

*Result:* completion and prompt token counts are identical across every arm and every level of the paired runs — `output_tokens_total` 733, `prompt_tokens_total` 6285, `output_tokens_mean` 15 at c1/c8/c16 on all five e2e arms (baseline and PR alike), zero failed requests. Per-request thinker text hashes are identical 12/12 in the dedicated greedy comparison (thinker and talker both forced greedy via `temperature=0`, `top_k=1`, `talker_top_k=1`, fixed seed).

### Protocol-level equivalence

*What / why:* the gate is per-message byte identity, not just identical concatenated audio: deferring an emission by one window puts message count, order, and payload all at risk.

*Setup:* 13 new tests comparing a full per-message snapshot against an `enable_output_overlap=False` control fed the identical stream. Eleven run on CPU with the CUDA event and pinned allocation stubbed; two are `gpu`-marked and use real pinned buffers, events, and graph capture/replay.

*Result:* byte-identical in every case — including a stream ending exactly on a window boundary (where a naive implementation silently drops the pending window), a stream ending with a tail, non-streaming requests, abort with a copy in flight, replay failure with a pending window live, and pool exhaustion. Two further tests pin the mechanisms underneath: the borrowed graph output is consumed before the next replay, and a staged EOS never fires a short window.

Per the amendment's requirement to state where byte identity was verified: this is deterministic-by-construction equivalence at the scheduler level (identical input stream into two schedulers differing only by the kill switch), on CPU and on one H100 at concurrency 1. It is not a claim that server-level audio is reproducible on this stack — it is not.

### Server-level waveform comparison — supporting evidence

*What / why:* reported as supporting evidence, not as a gate, per the amendment. Included because it independently reproduces the amendment's own finding on topologies it did not cover.

*Setup:* A/A control before any A/B — same tree, seed, prompts, container and GPUs, thinker and talker both greedy (including `talker_top_k=1`, which the amendment notes the standard harness cannot send), 8 streaming + 4 non-streaming requests, on both the TP2 disaggregated and colocated single-GPU profiles.

*Result:* the A/A control itself differs on 12/12 requests in both topologies, from the first audio chunk onward, with codec chunk counts drifting in both directions — while thinker text is byte-identical 12/12 in the same runs. This matches the amendment's measurement (0/48 matching PCM payloads at c1, thinker text 48/48 identical) and extends it: the divergence persists on the colocated profile and with talker sampling explicitly pinned, so it is not a batch-composition or missing-sampling-parameter artifact. The #1145-era bitwise A/B passed on the colocated profile at `aa6f92c3`, so this appeared since. The A/B arms differ in the same shape and to the same degree as the A/A control, i.e. no signal attributable to this PR.

### WER

*What / why:* the primary audio-quality instrument under the amendment. Run with an A/A control first, because the amendment establishes that this stack does not reproduce audio — without a baseline-vs-baseline band, a single-pair WER delta cannot be read.

*Setup:* CI-equivalent pipeline — generate on the TP2 profile at c1, stop the TTS server, transcribe with the Qwen3-ASR-1.7B router (2 workers × 1 GPU, same fixture as the CI WER job), SeedTTS-50 EN voice-clone, `temperature=0`, 50 samples per arm. Four arms total: baseline vs PR, plus baseline vs baseline as the control.

| Arm | Corpus WER | p95 | Max single sample | Samples > 50% | Evaluated |
| A/A control, arm 1 | 3.546% | 25.92% | 50.00% | 0 | 50/50 |
| A/A control, arm 2 | 1.241% | 7.23% | 14.29% | 0 | 50/50 |
| A/B baseline | 2.305% | 15.60% | 26.67% | 0 | 50/50 |
| A/B PR | 1.950% | 11.61% | 66.67% | 1 | 50/50 |

*Result:* the A/A control spans 2.305 pp of corpus WER (3.546% → 1.241%, a 186% relative swing between byte-identical trees). The A/B delta is 0.355 pp — **0.15× the A/A band**, and in the favorable direction. Paired per sample, the A/B arms differ on 6 of 50 (1 worse, 5 better, net −0.036 total WER); the A/A control differs on 9 of 50 (2 worse, 7 better), i.e. the control is the noisier comparison. Audio durations swing both ways in both comparisons (A/B: 24 longer, 23 shorter, −1.76 s to +2.40 s; A/A: 24 longer, 21 shorter, −2.16 s to +1.28 s).

The PR arm has one sample above the 50% threshold (66.7%, `common_voice_en_18436934`) where the baseline arm scored 0%. Reported rather than smoothed over. Three things bear on attribution: the A/A control's own worst sample is 50.00%, exactly at the threshold, so the tail demonstrably reaches that region without this change; across the four arms the worst-sample statistic is 26.67% / 66.67% / 50.00% / 14.29%, with no relation to which tree ran; and the failure mode is 1.2 s of semantically *new* speech prepended ("pursing a college in the 42 …"), which Code2Wav cannot manufacture — its output length is a function of the codec frames the talker sends it, and this PR does not touch the talker. Per-sample data for all four arms is in the artifacts gist.

### Unit tests

`tests/unit_test/qwen3_omni/test_code2wav*.py` and `tests/unit_test/scheduling/test_streaming_vocoder.py` green on the patched tree including both `gpu` cases on an H100; baseline tree run over the same pre-existing files as an environment control.

## Benchmark & Profiling — vocoder launch gap

**Expected:** the scheduler thread stops waiting out GPU compute and D2H, collapsing its per-window occupancy toward pure host work.

*Setup:* canonical TP2 profile (`--thinker-tp-size 2 --gpu-thinker-tp 0,1 --gpu-talker 1 --gpu-code2wav 1`) on 2× H100, model @ `26291f79`, `temperature=0`, one server launch per tree. `/start_request_profile` for events at c1 (8 requests) and c8 (16); `/start_profile enable_torch` for the per-stage Chrome trace.

`code2wav_decode_start → decode_end`, the scheduler thread's per-window occupancy:

| Level | Baseline p50 | PR p50 | Baseline mean | PR mean |
| --- | --- | --- | --- | --- |
| c1 | 8.98 ms | **0.65 ms** | 16.36 ms | 14.62 ms |
| c8 | 9.14 ms | **3.28 ms** | 13.24 ms | 8.22 ms |

CPU-side CUDA API time between consecutive `cudaGraphLaunch` calls, Code2Wav process trace:

| Between launches | Baseline | PR |
| --- | --- | --- |
| Blocking (sync + memcpy) per gap, p50 | 32.01 ms | **1.95 ms** |
| `cudaStreamSynchronize`, total | 386.93 ms | **0.32 ms** |
| `cudaMemcpyAsync`, total | 79.81 ms | 37.62 ms |

*Result:* both views agree — −93% p50 occupancy, −94% p50 blocking. `d2h_wait_ns` p50 is 6.6 µs (c1) / 7.4 µs (c8), but this is only the residual wait after the lazy EOS scan's host synchronization; the process-wide CUDA API trace above is the load-bearing evidence that blocking synchronization was removed. Pipelined/sync window counts are 29/16 and 54/32 — the sync counts are exactly one first window plus one tail per request, so the TTFA path fires where intended and nowhere else. Inter-launch gap p50 moves 56.9 → 64.3 ms (c1) and 100.5 → 107.9 ms (c8): arrival-bound at steady state, since the talker paces windows one frame per decode step. Reported, not claimed — the win is thread occupancy, not cadence.

## End-to-end paired comparison — not promotion-capable on this host

**Expected:** neutral; item 4.5 does not gate on an E2E win.

*Setup:* SeedTTS-50 EN voice-clone streaming on the TP2 profile at c1/c8/c16, `temperature=0`, 4 rounds per level (round 0 discarded as warmup), counterbalanced AB/BA. Five arms completed on a shared-tenant cloud H100, zero failed requests.

*Result:* identical baseline arms — same tree, container, GPUs — disagree by more than the effect being measured:

| Level | Baseline throughput_qps across identical arms | Spread |
| --- | --- | --- |
| c1 | 1.802 / 1.806 / 1.826 | 1.3% |
| c8 | 3.507 / 5.813 / 4.482 | 65.8% |
| c16 | 3.441 / 7.259 / 4.498 | 111.0% |

This agrees with #1018's own measured A/A bands, added 2026-08-15 (c1 up to 5.03%, c8 11.18%, c16 13.78%, with the conclusion that "the speech workload at concurrency 8 and 16 is not promotion-capable as currently configured"). The spreads here are wider still — three arms over several hours on a shared-tenant host, versus two back-to-back legs — but the verdict is the same one the roadmap already reached. c8/c16 are therefore recorded as unusable rather than as a result in either direction; baseline-vs-PR mean separation there is 0.76–1.03 baseline SD. At c1, where the benchmark does resolve, the PR arm is neutral-to-slightly-positive (throughput +1.0%, TTFA mean −1.8%) — inside the noise band, not claimed.

## Environment

| Item | Value |
| --- | --- |
| Baseline | `f1fbaf28` (origin/main) |
| Model | `Qwen/Qwen3-Omni-30B-A3B-Instruct` @ `26291f79` |
| Hardware | H100 80GB HBM3 × 2 (paired arms always on the same host) |
| Stack | torch 2.11.0+cu130, sglang 0.5.12.post1, flashinfer 0.6.11.post1 |
| Placement | thinker TP2 on GPU 0+1, talker + code2wav on GPU 1; `mem_fraction_static` 0.55 / 0.21 |
| Raw artifacts | https://gist.github.com/EdwardZhang1108/ff6fd7c0ce60787d5c70672efd158e7d |


sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -249,22 +331,112 @@ def decode_delta(
...
         wav = wav[..., -(end - start) * self._total_upsample :]
-        audio = wav.reshape(-1).detach().cpu().float().numpy().copy()
+        samples = int(wav.numel())
+        prev_wait_ns = 0
+        prev_waveform: torch.Tensor | None = None
+        slot: _PinnedSlot | None = None
+        # note (EdwardZhang1108): the first window stays synchronous so
+        # time-to-first-audio is unchanged.
+        if self._pipeline_active and not is_final and start > 0 and samples > 0:
+            slot = self._acquire_slot(samples)
```
edwingao28, line 352: Any exception between acquiring a slot and assigning it to `state.pending` strands the new slot, because cleanup only sees `state.pending`; repeated failures permanently shrink the pool and eventually disable overlap. Wrap this region in an ownership guard that retires or quarantines the current slot on failure, and test a previous-window flush failure after the current copy is recorded.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -249,22 +331,112 @@ def decode_delta(
...
+    def _flush_pending(
+        self, request_id: str, state: Code2WavStreamState
+    ) -> tuple[int, torch.Tensor | None]:
+        """Materialize the pending window: wait for its D2H copy, append the
+        audio, and return (wait_ns, waveform-or-None). The slot returns to the
+        pool only after the audio is copied into owned memory."""
+        pending = state.pending
+        if pending is None:
+            return 0, None
+        state.pending = None
```
edwingao28: `state.pending` is cleared before `event.synchronize()` and the NumPy copy, so either failure makes the previous slot unreachable; `_acquire_slot` has the same leak if buffer growth fails after popping a free slot. Keep pending ownership through materialization and return the popped slot on grow failure, while never recycling a slot whose D2H completion is unknown.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -442,10 +635,42 @@ def on_stream_done(self, request_id: str):
...
+        try:
+            if self._device.type == "cuda":
+                # note (EdwardZhang1108): abort arrives on the stage
+                # event-loop thread, which never set this stage's device.
+                with torch.cuda.device(self._device):
+                    pending.slot.event.synchronize()
+            else:
+                pending.slot.event.synchronize()
+        except Exception:
```
edwingao28: A failed `event.synchronize()` still falls through to `_release_slot`, allowing the next request to reuse a buffer whose D2H completion was never proven. Retire or quarantine the slot on failure, and only return it to `_pinned_free` after `synchronize()` succeeds or `query()` reports completion.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -204,17 +244,59 @@ def ingest(
...
+        eager path). One host sync per scan instead of one per frame.
+        """
+        chunks = state.chunks
+        start = state.checked
+        if start >= len(chunks):
+            return
+        unchecked = chunks[start:]
+        if all(codes.ndim >= 1 for codes in unchecked):
+            heads = torch.stack([codes[0] for codes in unchecked])
+            is_eos = (heads == self._codec_eos_token_id).tolist()
```
edwingao28, line 294: `heads.tolist()` synchronizes the CUDA stream before the next replay, so `d2h_wait_ns` measures only residual waiting after that sync and can hide lost overlap. Either use a CPU-owned EOS signal or move the scan after launch; otherwise document this limitation and rely on the process-wide CUDA API trace for the overlap claim.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -204,17 +244,59 @@ def ingest(
...
+        leading code equals the codec EOS id is dropped and every other frame
+        keeps its arrival order (0-dim frames bypass the check, as in the
+        eager path). One host sync per scan instead of one per frame.
+        """
+        chunks = state.chunks
+        start = state.checked
+        if start >= len(chunks):
+            return
+        unchecked = chunks[start:]
+        if all(codes.ndim >= 1 for codes in unchecked):
```
edwingao28: For `ndim >= 2`, `.tolist()` returns nested lists and the truthiness checks silently drop every staged frame, whereas eager mode raises. Change the fast-path guard from `>= 1` to `== 1` so malformed ranks use the eager-compatible slow path.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -249,22 +331,112 @@ def decode_delta(
...
         state.emitted = end
         state.due_since = None
+        if profile_metadata is not None:
+            _emit_event(
+                request_id=request_id,
+                stage=None,
+                event_name="code2wav_decode_end",
+                metadata={
+                    **profile_metadata,
+                    "audio_samples": samples,
```
edwingao28, line 437: In the pipelined branch, `audio_samples` and the frame fields describe the current launch while the returned waveform belongs to the previous window, so consumers can associate metrics with the wrong chunk. Add explicit launched-versus-materialized fields, or report the flushed window’s counts here.

docs/basic_usage/qwen3_omni.md
```diff
@@ -195,6 +195,20 @@ The feature derives the exact `B=1` threshold windows from
 `T{10,20,30,35}`. Unsupported shapes and final stream tails run eagerly.
 Capture-time incompatibilities also fall back to eager execution.
 
+Output overlap is also enabled by default on CUDA devices: each threshold
+window's waveform readback runs as an asynchronous device-to-host copy into a
+pinned staging buffer and is materialized while the GPU computes the next
+window, and the codec EOS check runs once per window instead of once per
+frame. The first window of every request stays synchronous, so
+time-to-first-audio is unchanged. Audio bytes and message boundaries are
+identical to the synchronous path. To disable it:
```
edwingao28: The byte/message-identity claim is true only for successful requests: on a mid-stream failure the overlap path can drop an un-emitted pending window. Qualify the documentation and add a test that pins the actual `stream → error` behavior.

tests/unit_test/qwen3_omni/test_code2wav_overlap.py
```diff
@@ -0,0 +1,518 @@
...
+    def __init__(self) -> None:
+        self.synced = False
+
+    def record(self) -> None:
+        pass
+
+    def query(self) -> bool:
+        return True
+
+    def synchronize(self) -> None:
```
edwingao28, line 56: These CPU fakes complete the copy before `record()`, so the tests stay green even if `_flush_pending` omits its required synchronization. Make the fake buffer unreadable until `synchronize()`, or at least assert that a successful flush synchronizes before releasing the slot.

tests/unit_test/qwen3_omni/test_code2wav_overlap.py
```diff
@@ -0,0 +1,518 @@
...
+    monkeypatch.setattr(torch.cuda, "Event", _FakeEvent)
+
+
+def _seed(scheduler: Code2WavScheduler, request_id: str = "req-1") -> None:
+    scheduler._stream_payloads[request_id] = make_qwen_payload(request_id=request_id)
+    scheduler._get_or_create_stream_state(request_id)
+
+
+def _chunk(index: int) -> torch.Tensor:
+    return torch.tensor([index % 7 + 1, 10])
```
edwingao28, line 131: No byte-identity test includes codec EOS

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
```diff
@@ -102,18 +103,48 @@ def load_code2wav_model(
...
     due_since: float | None = None
+    checked: int = 0
+    pending: _PendingWindow | None = None
 
 
 class Code2WavScheduler(StreamingVocoderBase[Code2WavStreamState, "list[int]"]):
     """Streaming vocoder scheduler. Same inbox/outbox interface as OmniScheduler."""
 
+    # note (EdwardZhang1108): one slot per stream plus one for the decode in
+    # flight, since acquire happens before the previous window is flushed.
```
edwingao28: please follow repo comment style

tests/unit_test/qwen3_omni/test_code2wav_overlap.py
edwingao28: Fixed in `0a427297`: the release spy now asserts that synchronization has already completed when `_release_slot` runs.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
edwingao28: Fixed in `608294ee`: the profiler docs define `d2h_wait_ns` as residual after the EOS synchronization, and the PR description now makes the process-wide CUDA API trace the load-bearing evidence.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
edwingao28: Fixed in `67883229`.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
edwingao28: Fixed in `67883229`.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
edwingao28: Fixed in `67883229`.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
edwingao28: Fixed in `67883229`.

sglang_omni/models/qwen3_omni/components/code2wav_scheduler.py
edwingao28: Fixed in `67883229`.

tests/unit_test/qwen3_omni/test_code2wav_overlap.py
edwingao28: Fixed in `67883229`.

edwingao28:
**`decode_delta` returns the *previous* window (`code2wav_scheduler.py:449`), so audio only leaves the stage when the next window is dispatched — ~57 ms away — to save an 8.3 ms stall.** Fix pushed as `c4ab8e8`: flush from `_decode_and_emit`, which runs per frame (~5.7 ms), once the event has completed. Async D2H and the occupancy win are untouched (8.72 → 0.46 → 0.43 ms p50 at c1).

ITL p95 vs `enable_output_overlap: false`, 2×H200 TP2, ABC/CBA, 2 boots/arm. `floor` = within-arm spread:

| c | as merged | with fix | floor |
|---|---|---|---|
| 1 | **+49.5%** | −14.4% | 26.5% |
| 8 | **+40.5%** | −7.2% | 12.5% |
| 16 | **+39.2%** | +9.5% | 13.1% |
| 32 | **+28.1%** | −5.7% | 12.7% |

p99 same shape. TTFT/TTFA/E2E/throughput inside floor for both arms. Reproduced in 3 runs, 12/12 cells.

Tests: replay-failure now expects `["stream","stream","error"]` (drained audio reaches the client instead of dying with the abort). Deleted `test_overlap_decode_end_separates_launched_and_materialized_samples` — `materialized_previous_audio_samples` is 0 on the normal path now, so it asserted nothing; your call whether to keep the field.

Caveat: ~3.4 s audio/request, couldn't pin talker length, so short-form only.

edwingao28:
## Calibration results — omni audio-path CI thresholds on this PR

Ran the full `tune-ci-thresholds` calibration for the five audio-path CI units (`mmmu_talker`, `mmsu_talker`, `tts`, `videomme_talker`, `videoamme_talker_tp2`) on this PR merged with current `main` (`f158fe2b` = PR head `c4ab8e8d` + `0ae7669d`), so the numbers include #1384. The branch has since gained `29c783b9` and `e1269aa6` (a style pass and dropping one profiler field); neither touches the measured path in a way that should move these numbers.

**Setup:** 2× H100 (GPUs 0,1), CPU pinned to the CI lane bundle `2-15,66-79`, SGLang 0.5.16, PyTorch 2.11.0+cu130, CI image `hongccc/sglang-omni:dev`. At least 5 clean observations per stage after destructive-round rejection (strict worst-of-N; every retained run had full sample coverage). Run artifacts: `.tune-runs/20260817T033506Z_omni_audio_mainmerge_r5/report.md`.

**Speed references this PR can tighten** (worst-of-N raw values, pre-slack), now pushed to this branch in 6bd41792 together with the regenerated `stages.yaml`:

| Stage | Metric | current raw | measured worst | delta |
|---|---|---:|---:|---|
| tts | throughput (req/s, min) | 8.13 | 8.921 | +9.7% |
| tts | latency mean (s, max) | 1.77 | 1.644 | -7.1% |
| tts | RTF mean (max) | 0.5818 | 0.5124 | -11.9% |
| mmmu_talker | throughput (min) | 0.733 | 0.932 | +27% |
| mmmu_talker | latency mean (max) | 14.908 | 12.749 | -14.5% |
| mmmu_talker | RTF mean (max) | 0.3782 | 0.2986 | -21% |
| videomme_talker | throughput (min) | 0.747 | 0.981 | +31% |
| videomme_talker | RTF mean (max) | 1.3972 | 0.9986 | -28.5% |
| videoamme_talker_tp2 | latency mean (max) | 86.562 | 54.071 | -37.5% |
| videoamme_talker_tp2 | RTF mean (max) | 6.9067 | 4.4757 | -35.2% |

Correctness stayed healthy across every retained round: UTMOS worst 4.4568 (current reference 4.4497), thinker accuracy at or above current gates (VideoMME talker accuracy worst 60% vs the 45% gate), per-stage WER worst-of-N in the report. The commit also refreshes the correctness references measured here, including the small loosens (tts corpus WER 1.60% -> 2.31%, MMSU accuracy 0.60 -> 0.575) so the gates match observed spread.

**Two things I did not fold into this update:**

1. `mmsu_talker_speed` is genuinely noisy in this environment: across 5 clean rounds throughput spans 1.04-1.86 req/s (range ~54% of the median). Worst-of-N would loosen all four of its references, so I left that stage untouched; it should be recalibrated separately.
2. `videomme_talker` WER moved between rounds under greedy decoding (rejected rounds at 1.7-2.6% versus a surviving median of 1.4%; the first 5-round block split into two populations and was discarded wholesale). This looks like the known audio nondeterminism rather than a regression from this PR, but it deserves its own issue.

Full per-run tables, rejected-round evidence, and the environment fingerprint are in the run directory report.

edwingao28 (approved):

PR #1576, opened 2026-08-17
Wire /v1/audio/translations into the router

0xjeffro:
## Motivation

#1453 added the translation endpoint but did not wire it up to the router.
## Modifications

- Wire /v1/audio/translations into the router
- Add a test that asserts every /v1/... route on the worker also exists on the router, so the two cannot drift apart the next time an endpoint changes.
- update docs
## Accuracy Test

N/A
## Benchmark & Profiling

N/A
## Checklist

- [x] Add unit tests.

## CI


sglang_omni_router/route_metadata.py
```diff
@@ -490,7 +493,7 @@ def _required_capabilities(
         capabilities: set[Capability] = {"speech"}
     elif route_kind is RouteKind.VOICE_CONTROL:
         capabilities = {"speech"}
-    elif route_kind is RouteKind.TRANSCRIPTION:
+    elif route_kind in {RouteKind.TRANSCRIPTION, RouteKind.TRANSLATION}:
```
SandyLuXY, line 496: Seems multipart `model` field never reaches `RouteMetadata`, shall we parse the multipart `model` field before selection?

sglang_omni_router/app.py
```diff
@@ -779,6 +779,10 @@ async def audio_speech(request: Request) -> Response:
     async def audio_transcriptions(request: Request) -> Response:
         return await _forward(request, "/v1/audio/transcriptions")
 
+    @app.post("/v1/audio/translations")
+    async def audio_translations(request: Request) -> Response:
+        return await _forward(request, "/v1/audio/translations")
```
SandyLuXY, line 784: Shall we cover `stream=true` case by parsing the stream mode?

docs/basic_usage/omni_router.md
```diff
@@ -594,7 +600,7 @@ At `N >= 2` the router runs as a small process tree:
 - A **supervisor** binds the public port once and passes the listening socket
   to `N` **data-plane (DP)** processes, which accept from the shared queue and
   relay the model routes (`/generate`, `/v1/chat/completions`,
-  `/v1/audio/speech`, `/v1/audio/transcriptions`).
+  `/v1/audio/speech`, `/v1/audio/transcriptions`, `/v1/audio/translations`).
```
SandyLuXY, line 603: nit: `sglang_omni_router/data_plane.py:6` still says it relays “the four model routes",  please update the doc stirng,

SandyLuXY (commented):
Thanks for the work!

sglang_omni_router/route_metadata.py
0xjeffro: hi xinyu thanks! fixed this by commit [f892a4a](https://github.com/sgl-project/sglang-omni/pull/1576/commits/f892a4a371a48f43b832608fac2fd2da385be478) also for the stream mode you mentioned below.

TLDR: model and stream are multipart form fields, and the router never parsed the form before, so it couldn't see them. It now walks the parts along the boundaries and picks up just these two fields; the audio file part is skipped by jumping straight to the next boundary without parsing it, so the cost is one linear pass over the body. When a field can't be read, the router falls back to the existing route-header behavior, when a field conflicts with its header, the request gets a 400.

SandyLuXY (approved):
lgtm

PR #1578, opened 2026-08-17
[XPU] Fused QK-norm-RoPE for the Qwen3-Omni thinker

siju-samuel:
## Motivation

Two XPU gaps on the Qwen3 serving path.

`XPUOmniPlatform` inherited the base `get_fused_qk_norm_rope()`, which returns
`None`, so the kernel was unreachable on XPU even though the XPU `sgl_kernel`
build carries the op — this PR gives the hook a consumer. Separately, two places
name `torch.cuda` directly: the Qwen3-ASR post-OOM cache release raises
`ValueError: Expected a cuda device, but got: xpu:0` (swallowed by its `except`,
so the release never ran), and the GPU device-info probe returns nothing off
NVIDIA because NVML and `torch.cuda` are both empty there.

## Modifications

- `platforms/xpu.py`: return the XPU kernel, `None` when the pinned
  sgl-kernel-xpu predates the symbol.
- New `qwen3_omni/components/thinker_fused_rope.py`: patch `apply_qk_norm_rope`
  per layer at construction, as the Qwen3-ASR path already patches upstream
  `Qwen3ForCausalLM`. The thinker does **not** use this repo's
  `Qwen3OmniMoeThinkerTextAttention` — the arch maps to `components/sglang_thinker.py`,
  which builds upstream `Qwen3MoeLLMModel`.
- Why fusing an MRoPE model is legal: MRoPE carries three position rows and the
  kernel takes one. They differ only for image/video tokens; where they agree,
  `apply_interleaved_rope` copies from a single row, so MRoPE's rotation is the
  kernel's rotation.
- Gates, all inside install so an unaffected platform does no work at init, and
  the kernel is acquired last (a hook may import `sgl_kernel` eagerly and raise
  where the build lacks it): XPU only; refused while a prefill graph could freeze
  the per-batch decision; refused for an already-patched layer, whose fallback
  pointer would otherwise become the wrapper and recurse. Per forward: text-only
  prefill, at most 128 tokens, three rows verified equal.
- `qwen3_asr/encoder_service.py`, `utils/gpu_memory.py`: resolve through
  `torch.get_device_module()`. CUDA unaffected — it resolves to `torch.cuda`, and
  NVML still answers first for NVIDIA parts.

## Related Issues

Follow-up to #994.

## Accuracy Test

Qwen3-Omni thinker, `--text-only`, TP=8 on 8x Arc Pro B60: the kernel runs on all
48 layers and greedy output is byte-identical to the unfused path on three
prompts, with the unfused arm recording zero fused calls. Qwen3-ASR still correct
on XPU (it already drives this kernel, so it is the regression check for the
memory commit).

## Benchmark & Profiling

Per attention block, pipelined submission, one idle B60 — the 128-token bound is
this crossover, not a guess:

| tokens | 16 | 128 | 192 | 909 |
|---|---|---|---|---|
| fused vs unfused | +23.7 us | +6.7 us | -2.2 us | -39.4 us |

Bounded there and prefill only, best case is ~0.6 ms of a ~180 ms forward. The
larger win is decode (1.24 ms/step), which needs the graph-freeze problem solved
first. No perf claim for the memory commit: the 14.0 GiB -> 0 GiB reserved change
is a reclaim that previously did not happen, not a faster retry.

## Checklist

- [x] Add unit tests.

### DIFF
--- a/sglang_omni/models/qwen3_omni/components/sglang_thinker.py
+++ b/sglang_omni/models/qwen3_omni/components/sglang_thinker.py
@@ -23,7 +23,6 @@ from sglang.srt.models.qwen3_vl_moe import Qwen3MoeLLMModel, load_fused_expert_w
 from sglang.srt.utils import add_prefix, logger
 
 from sglang_omni.models.qwen3_omni.components.thinker_fused_rope import (
-    ThinkerFusedRopeGate,
     install_thinker_fused_rope,
 )
 from sglang_omni.quantization import get_weight_preprocessor
@@ -40,10 +39,6 @@ def _config_uses_mrope(config: Any) -> bool:
 
 class Qwen3OmniThinkerForCausalLM(nn.Module):
     """Qwen3-Omni thinker text model without duplicated audio/vision towers."""
-
-    # Callers that bypass __init__ (tests build the wrapper attribute by
-    # attribute) still reach forward, which reads this.
-    _fused_rope_gate: "ThinkerFusedRopeGate | None" = None
 
     def __init__(
         self,
@@ -72,7 +67,7 @@ class Qwen3OmniThinkerForCausalLM(nn.Module):
                 prefix=add_prefix("lm_head", prefix),
             )
         self.logits_processor = LogitsProcessor(self.config)
-        self._fused_rope_gate = install_thinker_fused_rope(self.model, self.config)
+        self._fused_rope_gate = install_thinker_fused_rope(self.model)
 
     @property
     def thinker(self) -> "Qwen3OmniThinkerForCausalLM":
--- a/sglang_omni/models/qwen3_omni/components/sglang_thinker.py
+++ b/sglang_omni/models/qwen3_omni/components/sglang_thinker.py
@@ -22,6 +22,10 @@ from sglang.srt.model_loader.weight_utils import default_weight_loader
 from sglang.srt.models.qwen3_vl_moe import Qwen3MoeLLMModel, load_fused_expert_weights
 from sglang.srt.utils import add_prefix, logger
 
+from sglang_omni.models.qwen3_omni.components.thinker_fused_rope import (
+    ThinkerFusedRopeGate,
+    install_thinker_fused_rope,
+)
 from sglang_omni.quantization import get_weight_preprocessor
 
 
@@ -36,6 +40,10 @@ def _config_uses_mrope(config: Any) -> bool:
 
 class Qwen3OmniThinkerForCausalLM(nn.Module):
     """Qwen3-Omni thinker text model without duplicated audio/vision towers."""
+
+    # Callers that bypass __init__ (tests build the wrapper attribute by
+    # attribute) still reach forward, which reads this.
+    _fused_rope_gate: "ThinkerFusedRopeGate | None" = None
 
     def __init__(
         self,
@@ -64,6 +72,7 @@ class Qwen3OmniThinkerForCausalLM(nn.Module):
                 prefix=add_prefix("lm_head", prefix),
             )
         self.logits_processor = LogitsProcessor(self.config)
+        self._fused_rope_gate = install_thinker_fused_rope(self.model, self.config)
 
     @property
     def thinker(self) -> "Qwen3OmniThinkerForCausalLM":
@@ -84,6 +93,8 @@ class Qwen3OmniThinkerForCausalLM(nn.Module):
         del get_embedding, omni_prefill_rids
         if forward_batch.mrope_positions is not None:
             positions = forward_batch.mrope_positions
+        if self._fused_rope_gate is not None:
+            self._fused_rope_gate.evaluate(positions, forward_batch)
 
         hidden_states = self.model(
             input_ids=input_ids,
--- a/sglang_omni/models/qwen3_omni/components/thinker_fused_rope.py
+++ b/sglang_omni/models/qwen3_omni/components/thinker_fused_rope.py
@@ -80,18 +80,20 @@ def _fused_apply_qk_norm_rope(
     kernel: Any,
     cos_sin_cache: torch.Tensor,
 ):
-    if not gate.enabled or qkv.dtype != torch.bfloat16:
+    if not gate.enabled or qkv.dtype != torch.bfloat16 or not qkv.is_contiguous():
         return attn._omni_unfused_apply_qk_norm_rope(qkv, positions, forward_batch)
 
     q, k, v = qkv.split([attn.q_size, attn.kv_size, attn.kv_size], dim=-1)
     tokens = qkv.shape[0]
-    # The kernel writes q and k in place and wants them per head and contiguous,
-    # which a slice of the packed projection is not.
-    q = q.reshape(tokens, attn.num_heads, attn.head_dim).contiguous()
-    k = k.reshape(tokens, attn.num_kv_heads, attn.head_dim).contiguous()
+    # The kernel wants q and k per head and writes them in place. It requires
+    # only the head dimension to be contiguous and takes the token and head
+    # strides as they come, so hand it views and let it update the packed
+    # projection directly: making these contiguous would copy both regions
+    # every layer. view() rather than reshape() so a shape that could not alias
+    # raises here instead of quietly taking the write into a copy.
     kernel(
-        q,
-        k,
+        q.view(tokens, attn.num_heads, attn.head_dim),
+        k.view(tokens, attn.num_kv_heads, attn.head_dim),
         attn.q_norm.weight,
         attn.k_norm.weight,
         cos_sin_cache,
@@ -102,11 +104,7 @@ def _fused_apply_qk_norm_rope(
     # forward_core reads this to decide save_kv_cache: the fused path writes no
     # KV of its own, so the attention call has to.
     attn._used_fused_qk_norm_rope_last_call = True
-    return (
-        q.reshape(tokens, attn.q_size),
-        k.reshape(tokens, attn.kv_size),
-        v,
-    )
+    return q, k, v
 
 
 def _prefill_graph_enabled() -> bool:
@@ -157,7 +155,9 @@ def install_thinker_fused_rope(
         )
         return None
 
-    provider = kernel_provider or current_platform.get_fused_qk_norm_rope
+    provider = (
+        kernel_provider or current_platform.get_fused_qk_norm_rope_with_cos_sin_cache
+    )
     kernel = provider()
     if kernel is None:
         return None
--- a/sglang_omni/models/qwen3_omni/components/thinker_fused_rope.py
+++ b/sglang_omni/models/qwen3_omni/components/thinker_fused_rope.py
@@ -1,20 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
 """Fused QK-norm plus RoPE for the Qwen3-Omni thinker's attention layers.
 
-The thinker's rotary is MRoPE, so upstream Qwen3MoeAttention keeps its fused
-branch off: the kernel takes one position per token while MRoPE carries three.
-Those three rows only differ for image and video tokens. Where they agree,
-MRoPE selects the same cos/sin entry for every frequency pair, which is the
-rotation the kernel applies, so a batch carrying no multimodal input can be
-fused and one carrying any falls back whole.
-
-The decision is a Python branch, so it must not be frozen into a replayed
-graph; installation is refused while prefill graphs are enabled. It is also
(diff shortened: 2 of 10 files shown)


sglang_omni/platforms/xpu.py
```diff
@@ -31,6 +34,23 @@ def set_device(self, device: "torch.device | int") -> None:
     def enable_code2wav_graph(self):
         return False
 
+    def get_fused_qk_norm_rope(self):
```
Ratish1, line 37: Both predictor paths bypass `apply_qk_norm_rope()`, and the remaining shared-attention callers are ineligible, so I cannot find a current runtime path that executes the callable returned here. Which current XPU path is expected to consume this hook?

tests/unit_test/qwen3_asr/test_encoder_service.py
```diff
@@ -17,6 +18,10 @@
 )
 
 _HIDDEN_SIZE = 4
+_DEVICE = get_device()
+_ACCELERATOR_ONLY = pytest.mark.skipif(
```
Ratish1, line 22: This marker admits MPS as well, but the test body requires `memory_reserved()`, which is not a generic accelerator API and is not exposed by `torch.mps`. Could this live allocator test be restricted to the CUDA/XPU device modules it actually supports?

tests/unit_test/test_platforms.py
```diff
@@ -112,3 +113,82 @@ def raise_driver_error() -> bool:
...
+        True,
+        positions,
+        1.0,
+        0.0,
+        0.0,
+        1.0,
+    )
+    fused_q, fused_k, fused_v = qkv.split([q_size, kv_size, kv_size], -1)
+
+    assert torch.equal(fused_q, expected_q)
```
Ratish1, line 190: This XPU-only test is skipped by the current CUDA unit CI, and `sgl-kernel-xpu` own BF16 reference test uses numerical tolerance rather than bitwise identity. Could Q/K use `torch.testing.assert_close()` with the upstream tolerance, while retaining exact equality for the untouched V?

sglang_omni/platforms/xpu.py
```diff
@@ -31,6 +34,23 @@ def set_device(self, device: "torch.device | int") -> None:
...
+        """Return the XPU sgl_kernel kernel, or None when this build lacks it.
+
+        sgl-kernel-xpu is pinned by commit rather than by release, so the symbol
+        cannot be assumed present the way the CUDA platform assumes it.
+        """
+        try:
+            from sgl_kernel import fused_qk_norm_rope
+        except ImportError as exc:
+            logger.info(
+                "XPU sgl_kernel has no fused_qk_norm_rope (%s); "
```
Ratish1, line 47: let use f formatting here

tests/unit_test/test_platforms.py
msinnha1: getting bit exact between SYCL norm+rope Vs PT reference, it should be assert_close() rather than using equal()

sglang_omni/models/qwen3_omni/components/thinker_fused_rope.py
```diff
@@ -0,0 +1,253 @@
...
+
+_FUSABLE_HEAD_DIMS = (64, 128, 256)
+# Measured per attention block on an Arc Pro B60, pipelined submission: the fused
+# kernel beats the two RMS-norms plus MRoPE by 23.7us at 16 tokens and 6.7us at
+# 128, then loses from 192 up (-2.2us) to -39.4us at 909. Fuse below the
+# crossover only, so a long prefill keeps the faster unfused pair.
+_MAX_FUSABLE_TOKENS = 128
+
+
+def _rows_are_identical(positions: torch.Tensor) -> bool:
```
Ratish1: On the XPU text-only serving path, `_compute_mrope_positions()` constructs all three rows from the same range and moves the result to `model_runner.device` before this adapter runs. These two `torch.equal()` calls therefore revalidate a prepared invariant by synchronizing XPU twice, while `positions_for()` maintains a device-copy cache for a tensor already on the QKV device. That work is also absent from the per-block benchmark and can consume the small gain near the 128-token crossover.

Could we keep the existing text-only, extend, `[3, T]`, and token-bound gates, set `position_ids = positions[0].to(torch.int32).contiguous()` once per forward, and have the wrapper consume it directly while retaining the original SGLang method as fallback? Installation should also be gated by `enable_fused_qk_norm_rope`. This keeps Qwen3-Omni MRoPE compatibility in the model adapter and limits `XPUOmniPlatform` to exposing the kernel.

tests/unit_test/pipeline/test_gpu_memory.py
```diff
@@ -270,6 +276,26 @@ def _import_module(name: str):
     assert info.total_memory_bytes == 96 * 1024**3
 
 
+def test_torch_device_info_probes_the_accelerator_module_not_cuda() -> None:
```
Ratish1: The preceding fallback test already proves this behavior by providing only `get_device_module()` and asserting the returned metadata. This source-inspection test checks implementation strings and duplicates that coverage. Could we remove it?

sglang_omni/platforms/xpu.py
```diff
@@ -31,6 +34,20 @@ def set_device(self, device: "torch.device | int") -> None:
     def enable_code2wav_graph(self):
         return False
 
+    def get_fused_qk_norm_rope(self):
+        # The cache-consuming variant: it reads a cos/sin table like the rotary
+        # does, where the analytic one recomputes sin and cos per element and
+        # loses to the unfused pair once a prefill is a few hundred tokens.
+        try:
+            from sgl_kernel import fused_qk_norm_rope_with_cos_sin_cache_inplace
```
Ratish1: `sgl-kernel-xpu@8c4f5e8` exports `fused_inplace_qknorm_rope`, not this symbol, so the caught `ImportError` currently makes the documented XPU build silently leave the feature disabled. `get_fused_qk_norm_rope()` also already has the packed analytic ABI consumed by `thinker_model.py`, while this kernel accepts separate Q/K tensors plus a cos/sin cache. can we preserve the existing hook and expose this cache ABI through a separate hook that imports `fused_inplace_qknorm_rope`.

sglang_omni/models/qwen3_omni/components/thinker_fused_rope.py
```diff
@@ -0,0 +1,231 @@
...
+Those three rows only differ for image and video tokens. Where they agree,
+MRoPE selects the same cos/sin entry for every frequency pair, which is the
+rotation the kernel applies, so a batch carrying no multimodal input can be
+fused and one carrying any falls back whole.
+
+The decision is a Python branch, so it must not be frozen into a replayed
+graph; installation is refused while prefill graphs are enabled. It is also
+refused off XPU, so CUDA and ROCm keep the unfused path they are tuned on.
+
+It installs by default on XPU. The cache-consuming kernel beats the unfused pair
```
Ratish1: This no longer checks `enable_fused_qk_norm_rope`, so correcting the kernel import will make every eligible XPU thinker install the adapter automatically. That contradicts the agreed default-off policy. Please restore the flag gate before kernel acquisition and preserve explicit flag-on/flag-off behavior.

sglang_omni/models/qwen3_omni/components/thinker_fused_rope.py
```diff
@@ -0,0 +1,231 @@
...
+    cos_sin_cache: torch.Tensor,
+):
+    if not gate.enabled or qkv.dtype != torch.bfloat16:
+        return attn._omni_unfused_apply_qk_norm_rope(qkv, positions, forward_batch)
+
+    q, k, v = qkv.split([attn.q_size, attn.kv_size, attn.kv_size], dim=-1)
+    tokens = qkv.shape[0]
+    # The kernel writes q and k in place and wants them per head and contiguous,
+    # which a slice of the packed projection is not.
+    q = q.reshape(tokens, attn.num_heads, attn.head_dim).contiguous()
```
Ratish1: The pinned kernel supports strided Q/K views when the last dimension is contiguous and reads the token/head strides explicitly. These `.contiguous()` calls therefore allocate and copy both packed Q and K regions in every layer. lets pass the reshaped split views directly so the kernel updates the original QKV storage.

tests/unit_test/qwen3_omni/test_thinker_fused_rope.py
```diff
@@ -0,0 +1,366 @@
...
+            None,
+            kernel_provider=provider,
+            prefill_graph_enabled=False,
+        )
+        is None
+    )
+
+
+@pytest.mark.gpu
+@requires_accelerator
```
Ratish1: `requires_accelerator` includes CUDA, where this hook returns the packed analytic kernel, but this test calls the XPU cache signature. The unit test fails here with missing analytic arguments. lets make this test XPU-only and obtain the kernel through the new cache-specific hook.

sglang_omni/models/qwen3_omni/components/thinker_fused_rope.py
```diff
@@ -0,0 +1,231 @@
...
+    if prefill_graph_enabled is None:
+        prefill_graph_enabled = _prefill_graph_enabled()
+    if prefill_graph_enabled:
+        logger.info(
+            "Qwen3-Omni thinker: fused QK-norm-RoPE stays off because a replayed "
+            "prefill graph would freeze the per-batch multimodal decision"
+        )
+        return None
+
+    provider = kernel_provider or current_platform.get_fused_qk_norm_rope
```
Ratish1: The suite tests the gate, installer, and raw kernel separately, but never invokes the installed wrapper after an eligible gate decision. lets add one CPU/fake-kernel contract case that installs this adapter and verifies the Q/K/V views, cache and positions, KV-save flag, return shapes, and fallback behavior.

sglang_omni/platforms/xpu.py
```diff
@@ -31,6 +34,39 @@ def set_device(self, device: "torch.device | int") -> None:
...
+        except ImportError as exc:
+            logger.info(
+                f"XPU has no sgl_kernel ({exc}); falling back to the unfused "
+                "QK-norm and RoPE path"
+            )
+            return None
+
+        for name in (
+            "fused_inplace_qknorm_rope",
+            "fused_qk_norm_rope_with_cos_sin_cache_inplace",
```
Ratish1: This symbol is not exported by either the pinned sgl-kernel-xpu revision or current main; can we remove this unsupported fallback?

sglang_omni/models/qwen3_omni/components/thinker_fused_rope.py
```diff
@@ -0,0 +1,231 @@
...
+    try:
+        prefill = get_global_server_args().cuda_graph_config.prefill
+    except ValueError:
+        return True
+    return prefill.backend != Backend.DISABLED
+
+
+def install_thinker_fused_rope(
+    model: Any,
+    config: Any,
```
Ratish1: nit: config is never read; every compatibility check uses attn.config. lets remove it from this new helper, its production call, and the tests?

sglang_omni/models/qwen3_omni/components/sglang_thinker.py
```diff
@@ -37,6 +41,10 @@ def _config_uses_mrope(config: Any) -> bool:
 class Qwen3OmniThinkerForCausalLM(nn.Module):
     """Qwen3-Omni thinker text model without duplicated audio/vision towers."""
 
+    # Callers that bypass __init__ (tests build the wrapper attribute by
+    # attribute) still reach forward, which reads this.
+    _fused_rope_gate: "ThinkerFusedRopeGate | None" = None
```
Ratish1: nit: this class default exists only for the test that bypasses __init__. lets initialize _fused_rope_gate on that test double and keep the actual state state initialized in __init__.

tests/unit_test/qwen3_asr/test_encoder_service.py
```diff
@@ -659,3 +665,30 @@ def test_flat_2d_encoder_output_is_also_accepted() -> None:
     service.encode_item(item)
 
     assert item.precomputed_embeddings.shape == (3, _HIDDEN_SIZE)
+
+
+@pytest.mark.gpu
+@requires_accelerator
+def test_the_device_cache_is_really_reclaimed_after_an_oom() -> None:
```
Ratish1, line 666: nit: test_oom_failure_detaches_traceback_and_recovers still patches torch.cuda.empty_cache() on a CPU stub and never reads the calls; lets remove that dead setup and its monkeypatch argument, since this live accelerator test covers cache reclamation.

Ratish1 (approved):

PR #1580, opened 2026-08-17
[Perf] [Voxtral-TTS] Enable breakable Prefill CUDA Graph

charliechenye:
## Motivation

Voxtral TTS builds request specific, voice conditioned input embeddings during prefill.

With SGLang 0.5.16, breakable Prefill CUDA Graph admission rejects a `ForwardBatch` when `ForwardBatch.input_embeds` is already populated. Voxtral previously wrote its composed embeddings directly to that field, so its language model prefill could not enter the breakable graph path even though the model body itself is graph compatible.

SGLang Omni already has a shared prefill sidecar for this case. It keeps runner composed embeddings out of `ForwardBatch.input_embeds` during graph admission, then late binds them through the shared `SGLModelRunner` path immediately before execution.

This PR makes Voxtral TTS a small adopter of that shared mechanism.

The scope is intentionally narrow. It enables breakable Prefill CUDA Graph compatibility when a user explicitly selects it. It does not default enable prefill graphs, choose capture buckets, graph the acoustic or vocoder stages, change decode behavior, or add Voxtral specific scheduler policy.

## What changes

* Mark `VoxtralTTSForConditionalGeneration` as supporting breakable Prefill CUDA Graph.
* Surface the same capability through `VoxtralTtsEngineBuilder`, which is the runtime gate used by the shared engine factory.
* Move the existing Voxtral prefill embeddings from `ForwardBatch.input_embeds` into `OmniPrefillInputs`.
* Reuse the shared `SGLModelRunner` late binding path to pass those embeddings to the model after graph eligibility has been decided.
* Accept the shared `omni_prefill_rids` forward kwarg. Voxtral does not need the request IDs itself, so it discards the value.
* Verify that Voxtral's existing logical endpoint selection remains correct when graph replay pads a live extend batch to a larger capture bucket.

No Voxtral specific graph runner, scheduler policy, bucket policy, or sidecar abstraction is introduced.

## Why the padded endpoint check matters

Voxtral already selects the logical final hidden row for each live extend request before the eager acoustic tail.

For example, live extend lengths `[2, 3]` correspond to logical endpoint rows `[1, 4]`. Breakable graph replay may return a padded hidden state buffer larger than the five live rows, but the acoustic tail must still receive rows `[1, 4]`.

The production endpoint selection predates this PR. The new test only qualifies that existing Voxtral invariant for the padded graph execution mode.

## Test ownership

The Voxtral specific coverage is intentionally small:

* `test_before_prefill_attaches_embeddings_to_sidecar`
  * Protects the behavior changed by this PR.
  * Verifies that the existing embedding builder output is attached to the Omni sidecar and that `ForwardBatch.input_embeds` remains unset.
  * Does not retest voice embedding composition.

* `test_forward_selects_logical_endpoints_from_padded_extend`
  * Protects a Voxtral owned invariant that breakable graph replay now relies on.
  * Verifies logical endpoint selection from a padded hidden state buffer.
  * Does not retest generic graph padding or logits behavior.

* `test_voxtral_builder_supports_breakable_prefill`
  * Protects the separate runtime builder gate used by the shared engine factory.
  * Model capability metadata remains covered by the centralized capability tests.

Shared sidecar transport, graph admission, graph lifecycle, and bucket policy remain covered by shared infrastructure tests.

## Related issue

Part of https://github.com/sgl-project/sglang-omni/issues/1357.

This PR only covers the Voxtral TTS adopter, so it does not close the multi model tracker.

## Accuracy and functional qualification

The change does not modify Voxtral model weights or architecture.

The main correctness risk is request specific voice embeddings being stale, misrouted, or contaminated when the same captured graph buffers are reused across requests.

### CPU coverage

The cleaned test coverage exercises:

* Voxtral sidecar adoption.
* Logical endpoint selection from padded extend output.
* Builder capability wiring.
* Existing Voxtral decode and pipeline behavior.
* Centralized model capability metadata.

Before merge, the latest rebased head should run:

```bash
pytest -q tests/unit_test/voxtral_tts/test_pipeline.py
pytest -q tests/unit_test/voxtral_tts
pytest -q tests/unit_test/models/test_model_capabilities.py
git diff --check
```

### A100 functional qualification

GPU functional qualification was performed on an NVIDIA A100 SXM4 40 GB using:

| Setting | Value |
| --- | --- |
| Model | `mistralai/Voxtral-4B-TTS-2603` |
| Revision | `b81be46c3777f88621676791b512bb01dc1cb970` |
| Prefill capture bucket | `[192]` |

The test used an A, B, A voice sequence:

| Request | Voice | Live prefill tokens | Captured bucket |
| --- | --- | ---: | ---: |
| A1 | `cheerful_female` | 152 | 192 |
| B | `casual_male` | 167 | 192 |
| A2 | `cheerful_female` | 152 | 192 |

For all three requests:

* `ForwardBatch.input_embeds` remained `None` through graph admission.
* The Omni sidecar contained exactly the live request row count.
* `PrefillCudaGraphRunner.can_run_graph()` returned true.
* Immediately before replay, the static `input_embeds` slot matched the current live sidecar with `max_abs=0.0`.
* SGLang reported `cuda graph: True` for prefill.
* Each request completed successfully with a valid mono 24 kHz WAV response.

The repeated cheerful request reproduced the same BCG hidden state, while the casual voice produced a different hidden state. This checks for stale request specific embeddings when graph buffers are reused.

An offline eager versus BCG hidden state comparison produced:

| Case | Eager versus BCG |
| --- | --- |
| `cheerful_female`, 152 tokens | cosine `0.999958`, relative L2 `0.9185%`, max absolute diff `0.125` |
| `casual_male`, 167 tokens | bit exact |

The padded 152 to 192 path therefore has a small numerical delta while preserving the tested semantic output and valid audio.

Final WAV byte equality is not used as the oracle because the downstream Voxtral acoustic flow matching path samples random noise.

## Benchmark and profiling

GPU performance qualification was performed on production equivalent candidate:

`12d8a25261eb2bf2773b397c30b5f47ac0a94b1b`

Subsequent commits only simplified and consolidated tests and comments. They did not change Voxtral Prefill CUDA Graph execution semantics.

### H200 setup

| Setting | Value |
| --- | --- |
| Candidate | `12d8a25261eb2bf2773b397c30b5f47ac0a94b1b` |
| GPU | NVIDIA H200, SM90 |
| SGLang | `0.5.16` |
| PyTorch | `2.11.0+cu130` |
| Checkpoint | `mistralai/Voxtral-4B-TTS-2603` |
| Revision | `b81be46c3777f88621676791b512bb01dc1cb970` |
| Radix cache | disabled |
| Torch compile | disabled |

The benchmark used three counterbalanced rounds. Each workload had 60 measured requests per arm per round after warmup.

The prompt length sweep covered 152, 189, and 413 prefill tokens at c1.

The concurrency sweep used the 152 token prompt at c2, c4, c8, and c16.

Across all rounds and workloads:

* 2,520 measured requests.
* 2,520 successful responses.
* 0 request or audio failures.
* Prefill graph routing in the ON arm: 100%.
* Prefill graph routing in the OFF arm: 0%.

Primary prefill time is the native SGLang Omni profiler interval from `scheduler_prefill_start` to `scheduler_prefill_end`.

### Aggregate results

| Workload | Native prefill OFF to ON | TTFA | E2E | QPS |
| Short c1 | 22.0 to 17.8 ms | 0.18% worse | 0.18% worse | 0.18% worse |
| Medium c1 | 22.0 to 17.8 ms | 0.02% better | 0.02% better | 0.02% better |
| Long c1 | 22.6 to 20.2 ms | 0.58% better | 0.58% better | 0.58% better |
| Short c2 | 23.1 to 19.3 ms | 4.18% better | 4.18% better | 4.46% better |
| Short c4 | 27.2 to 22.0 ms | 3.97% worse | 3.97% worse | 4.96% worse |
| Short c8 | 31.9 to 27.2 ms | 1.70% better | 1.70% better | 1.57% better |
| Short c16 | 40.7 to 33.6 ms | 0.50% better | 0.50% better | 0.69% better |

Native mean prefill time improved in every tested round and workload pair. Aggregate reductions ranged from about 10.8% to 19.3%.

The inner host side prefill runner wall time decreased by about 44% to 55%, depending on prompt length and concurrency.

### Service level interpretation

The clearest service level gain is c2:

| Metric | OFF | ON | Change |
| --- | ---: | ---: | ---: |
| TTFA | 308.1 ms | 295.2 ms | 4.18% lower |
| E2E | 308.3 ms | 295.4 ms | 4.18% lower |
| QPS | 6.47 | 6.75 | 4.46% higher |
| Native prefill | 23.1 ms | 19.3 ms | 16.46% lower |

At c8 and c16, prefill remains materially faster while TTFA, E2E, and QPS show smaller positive changes.

At c4, service level performance regresses even though native prefill is faster:

| Metric | Change at c4 |
| --- | ---: |
| TTFA | about 4.0% worse |
| E2E | about 4.0% worse |
| QPS | about 5.0% worse |
| Native prefill | about 19.3% faster |

The same 60 c4 requests required about 29.7 scheduler prefill batches in the OFF arm versus 42.3 in the ON arm. Average requests per prefill batch fell from about 2.02 to 1.42.

This is consistent with a traffic dependent scheduler batching interaction. It is not treated as a proven general root cause.

### Scheduler follow up

I reproduced the c4 fragmentation signal and tested the existing bounded prefill coalescing mechanism rather than adding new Voxtral specific scheduling behavior.

A `K=2`, `T=12 ms` configuration did not improve the relevant active decode fragmentation.

A trace guided `K=3`, `T=35 ms` configuration reduced aggregate batch count, but phase locked the closed loop workload into repeated `3 + 1` groups and increased active decode singleton batches. It also did not recover the service level regression.

Based on these experiments, this PR intentionally does not add Voxtral specific coalescing knobs or defaults.

The current conclusion is:

1. Breakable Prefill CUDA Graph is functioning correctly and consistently reduces Voxtral native prefill time.
2. Service level impact depends on scheduler batching and traffic shape.
3. Existing bounded coalescing did not provide a convincing Voxtral specific fix in the tested c4 workload.
4. More aggressive coalescing would add request admission latency without demonstrated net benefit.

This is why the feature remains opt in.

### Capture overhead

For the full H200 bucket ladder, the three ON rounds reported:

| Capture metric | Round 1 | Round 2 | Round 3 | Summary |
| Prefill graph capture time | 2.87 s | 3.12 s | 3.38 s | 3.12 s mean |
| Reported graph memory | 0.44 GB | 0.44 GB | 0.44 GB | 0.44 GB each round |

The benchmark therefore shows a repeatable prefill phase improvement with modest one time capture cost, while deployment level benefit remains workload dependent.

## Scope

This PR intentionally does not:

* Default enable breakable Prefill CUDA Graph.
* Prescribe production capture buckets.
* Graph the Voxtral acoustic transformer or vocoder.
* Change decode behavior.
* Add Voxtral specific graph infrastructure.
* Add Voxtral specific scheduler coalescing policy.
* Change voice embedding composition.

## Checklist

* [x] Format code with pre commit.
* [x] Add focused unit coverage.
* [x] Provide GPU functional qualification.
* [x] Provide performance and service level measurements.
* [x] Investigate the observed c4 scheduler interaction without adding unproven model specific policy.
* [ ] Run latest head CI after final rebase and maintainer labeling.

## CI


charliechenye:
## Follow up on the c4 scheduler regression (cc: @stephenkgli )

I did a deeper investigation into the c4 regression after the H200 results showed faster native prefill but worse service level performance.

The short version is that I can reproduce the scheduler fragmentation signal, but the existing bounded prefill coalescing policy does not look like a good Voxtral specific fix based on the experiments so far.

### 1. The original fragmentation signal reproduces

With `max_running_requests=16` restored  (in line with the original sweep up to c16), a representative 20 request c4 screen reproduced the original behavior:

```text
Eager prefill:
10 total prefill batches
5 active decode prefill batches
1 active decode singleton
2.00 requests per prefill
E2E median: 325.9 ms
QPS: 12.20

BCG prefill:
12 total prefill batches
7 active decode prefill batches
2 active decode singletons
1.67 requests per prefill
E2E median: 351.0 ms
QPS: 11.42
```

Native prefill was still faster with BCG, about 23.0 ms to 22.0 ms in this small screen, while the service level result regressed.

This supports the hypothesis from the full benchmark that faster BCG execution changes the scheduler timing and can produce more fragmented prefill admission at c4.

I still do not consider this a general root cause for all traffic shapes, but the scheduler interaction is reproducible under the original benchmark regime.

### 2. I tested the existing bounded prefill coalescing mechanism

Rather than adding a new scheduler policy, I tested the existing Omni coalescing behavior with the current semantics:

```text
hold while waiting requests < K
release when waiting requests >= K
or when the oldest wait reaches T
decode idle bypasses coalescing by default
```

I first tested:

```text
K=2
T=12 ms
```

This did not improve the relevant active decode fragmentation:

```text
BCG K=0:
7 active decode prefill batches
2 active decode singletons

BCG K=2, T=12 ms:
7 active decode prefill batches
2 active decode singletons
```

The total prefill count happened to move from 12 to 11, but that difference came from the decode idle portion of the workload. Since the current policy bypasses coalescing while decode is idle, I do not attribute that change to K/T.

### 3. Request arrival traces explain why K=2 does not help

I then looked at `scheduler_queue_enter` and `scheduler_prefill_start` events instead of tuning K/T blindly.

The closed loop c4 workload forms repeated request waves. A representative BCG K=0 trace looked roughly like:

```text
1 + 3
1 + 2 + 1
1 + 3
1 + 3
1 + 2 + 1
```

For the `1 + 2 + 1` waves, K=2 releases the pair immediately, so it never waits for the third request.

The tail request arrives roughly 22 to 31 ms later in the two relevant waves.

Based on that trace, I tested one targeted follow up rather than doing a parameter sweep:

```text
K=3
T=35 ms
```

### 4. K=3 confirms that coalescing changes the phase, but not in a useful way

K=3 reduced the total BCG prefill count to 10, but the detailed batch sequence showed that this was not the consolidation I was hoping for.

The measured c4 BCG sequence became:

```text
3 + 1
3 + 1
3 + 1
3 + 1
3 + 1
```

The active decode results were:

```text
BCG K=0:
7 active decode prefill batches
2 active decode singletons

BCG K=3, T=35 ms:
9 active decode prefill batches
5 active decode singletons
```

So K=3 effectively phase locked this closed loop workload into repeated `3 + 1` groups.

It reduced the aggregate batch count for the wrong reason and made the active decode fragmentation worse.

The service level result also did not improve. In that screen, BCG remained about 3.6 percent worse in median E2E and about 7.0 percent worse in QPS relative to the matched eager arm, despite native prefill remaining materially faster.

### 5. I do not think we should keep tuning K/T for Voxtral

The remaining option would be to coalesce while decode is idle so that the first request in each wave also waits for later arrivals.

The request traces suggest that collecting those requests would require intentionally delaying the head request by tens of milliseconds in several waves.

Given that Voxtral responses in this benchmark are only a few hundred milliseconds end to end, I do not currently see a compelling latency tradeoff for doing that just to recover a small number of prefill batches.

I also do not want to continue sweeping K/T until something looks positive. The K=2 and trace guided K=3 experiments are enough for me to conclude that the existing bounded coalescing policy is not an obvious model specific fix for this c4 behavior.

### Current conclusion

My current interpretation is:

1. BCG is still doing what this PR intends. Native Voxtral prefill is consistently faster and graph routing is correct.
2. The c4 service regression is associated with a real change in scheduler batching under this closed loop traffic shape.
3. The existing K/T coalescing mechanism does not causally recover the active decode batching in the tested configuration.
4. More aggressive coalescing would likely require adding intentional request admission latency, and I do not currently have evidence that the tradeoff is worthwhile.

I prototyped the Voxtral K/T plumbing on a separate follow up branch for the experiments, but based on these results I do not think we should merge that wiring just because it is technically available.

I would prefer to keep this PR scoped to enabling correct opt in breakable Prefill CUDA Graph support, keep the c4 scheduler behavior documented as a known traffic dependent interaction, and avoid adding Voxtral specific scheduling knobs without demonstrated service level benefit.

PR #1583, opened 2026-08-17
Add MUSA platform support

froststeam:
## Summary

Add a dedicated MUSA Omni platform so SGLang-Omni can recognize and launch on MUSA devices.

The platform keeps the integration thin and CUDA-compatible via `torchada`, while exposing MUSA-specific process environment handling and backend defaults:

- register `MUSAOmniPlatform` in platform dispatch
- detect MUSA before CUDA so `torchada` CUDA compatibility does not hide the MUSA platform
- map per-stage `MUSA_VISIBLE_DEVICES` for TP worker launches
- use CUDA IPC for intra-node transport
- default MoE runner backend to `triton`
- default FP8 GEMM runner backend to `deep_gemm`

## Validation

- `python -m py_compile sglang_omni/platforms/__init__.py sglang_omni/platforms/musa.py`
- Generated a text-to-video sample on a MUSA GPU using the official `damo-vilab/text-to-video-ms-1.7b-legacy` parameters:
  - `DPMSolverMultistepScheduler`
  - `num_inference_steps=25`
  - `num_frames=16`
  - `height=256`, `width=256`
  - output verified as H.264 MP4, 256x256, 16 frames, 8 fps

### DIFF
--- a/sglang_omni/platforms/__init__.py
+++ b/sglang_omni/platforms/__init__.py
@@ -1,6 +1,10 @@
 import os
 import pkgutil
 
+try:
+    import torchada  # noqa: F401
+except ImportError:
+    pass
 import torch
 from sglang.srt import platforms as srt_platforms
 from sglang.srt.platforms.interface import SRTPlatform
--- a/sglang_omni/platforms/__init__.py
+++ b/sglang_omni/platforms/__init__.py
@@ -8,6 +8,7 @@ from sglang.srt.platforms.interface import SRTPlatform
 from sglang_omni.platforms.cpu import CPUOmniPlatform
 from sglang_omni.platforms.cuda import CUDAOmniPlatform, ROCMOmniPlatform
 from sglang_omni.platforms.interface import OmniPlatform
+from sglang_omni.platforms.musa import MUSAOmniPlatform
 from sglang_omni.platforms.npu import NPUOmniPlatform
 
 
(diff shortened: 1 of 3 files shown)


sglang_omni/platforms/__init__.py
```diff
@@ -1,15 +1,19 @@
 import os
 import pkgutil
 
+try:
+    import torchada  # noqa: F401
+except ImportError:
+    pass
```
yeahdongcn, line 7: `torchada` should be imported here: https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/utils/common.py#L219

sglang_omni/platforms/musa.py
```diff
@@ -0,0 +1,115 @@
...
+
+from sglang_omni.platforms.interface import OmniPlatform
+from sglang_omni.utils.misc import normalize_quantization
+from sglang_omni.vendor.sglang.server_args import override_server_args
+
+if TYPE_CHECKING:
+    from sglang_omni.pipeline.stage_workers import StageLaunchConfig
+
+
+class MUSAOmniPlatform(OmniPlatform):
```
yeahdongcn, line 21: Can we just inherit from `CUDAOmniPlatform`? I think that would let us remove most of the duplicated implementation and keep only the necessary changes.

froststeam:

PR #1585, opened 2026-08-17
[Whisper ASR] Raise running-request limit to 64 and size the encoder cache in entries

0xjeffro:
## Motivation

This PR targets whisper-large-v3 (1.55B params), the largest checkpoint in the Whisper family. Until now our Whisper benchmarks were run on whisper-base (74M params). The PR makes two changes:

**1. Raise default max_running_requests from 32 to 64**([6ea1f42](https://github.com/sgl-project/sglang-omni/pull/1585/commits/6ea1f424b3b1a9266adfd45b28dae710ff2dcab5))

max_running_requests caps the decode batch (and, by default, the decode CUDA-graph capture range). Each decode step carries a cost that is nearly independent of batch size at small batches — reading the decoder weights plus per-kernel overhead — so a larger batch spreads it over more requests. For whisper-base (74M params, ~0.1 GB decoder) that cost is under 1 ms/step and 32 in flight already amortizes it. For whisper-large-v3 (1.55B params, ~1.5 GB decoder) it is ~2.75 ms/step; SeedTTS EN averages ~17 output tokens, so decode is ~47 ms of the 92 ms c=1 latency (51%). That is why the same limit that is fine for base caps large-v3.

**2. Size the pre-LM encoder cache in entries and derive the byte budget from that** ([f03f30b](https://github.com/sgl-project/sglang-omni/pull/1585/commits/f03f30bca2a5afc5f4bc18940b2d3a91acd56949))

The service log showed a pre-LM cache hit rate of only 22% on large-v3, while base was hitting almost every repeat. The cache had two limits, pre_lm_cache_max_entries=4096 and pre_lm_cache_size_bytes=2 GiB, and the byte limit always won: the 2 GiB was copied from Qwen3-ASR, whose embeddings are small and variable-length. However, Whisper is different. Its input is always one 30-second chunk, so every cached encoder output is one fixed-size [1500, d_model] FP16 tensor: 1.54 MB for base, 3.84 MB for large-v3. Under 2 GiB that is 1398 entries for base but only 559 for large-v3. SeedTTS EN has 666 distinct clips, so base kept the whole set resident and large-v3 thrashed the LRU — most requests re-ran the encoder and paid the device-to-host copy of the result. That is why the same budget that looked fine on base was undersized for large-v3.

Because the entry size is fixed and known at startup (encoder_token_count × d_model × dtype), the cache is now sized in entries and the byte budget is derived from that. pre_lm_cache_size_bytes becomes an optional host-memory cap that can only lower the derived budget (default None). The default entry count drops from 4096 to 1024, which is ~3.9 GB of host memory for large-v3 and ~1.5 GB for base and covers the SeedTTS working set. The startup log now reports the effective capacity (cache capacity=1024 entries x 3.84 MB = 3.93 GB) and the encoder stats gain cache_capacity_entries, so an undersized cache shows up next to cache_entries and cache_evictions instead of having to be reverse-engineered.

## Modifications
see diff

## Accuracy Test

WER is unchanged across all configurations, 0.0137–0.0138 at every concurrency.

## Benchmark & Profiling
Testing on`openai/whisper-large-v3` FP16, single NVIDIA H100 80GB.

### Ablation — throughput (req/s) / mean latency (ms) / P95 latency (ms)

| Concurrency | Default (mrr=32, cache 2 GiB) | mrr=64 only | cache 3 GiB only | mrr=64 + cache 3 GiB (this PR) |
|---:|---:|---:|---:|---:|
| 1 | 11.5 / 87 / 109 | — | 13.4 / 75 / 92 | 14.1 / 71 / 88 |
| 8 | 44.0 / 182 / 240 | 43.2 / 185 / 246 | 50.8 / 157 / 208 | 53.0 / 151 / 200 |
| 16 | 58.9 / 271 / 360 | 57.9 / 274 / 369 | 69.9 / 228 / 306 | 72.7 / 219 / 290 |
| 32 | 68.6 / 465 / 629 | 67.8 / 470 / 646 | 84.5 / 377 / 522 | 92.3 / 345 / 458 |
| 64 | 68.9 / 918 / 1103 | 76.7 / 829 / 1172 | 78.8 / 801 / 980 | 98.8 / 643 / 904 |

### Throughput relative to default

| Concurrency | mrr=64 only | cache 3 GiB only | mrr=64 + cache 3 GiB |
|---:|---:|---:|---:|
| 1 | — | +16.9% | +22.8% |
| 8 | −1.7% | +15.5% | +20.5% |
| 16 | −1.6% | +18.8% | +23.5% |
| 32 | −1.1% | +23.2% | +34.6% |
| 64 | +11.4% | +14.5% | +43.4% |

## Checklist

- [x] Add unit tests.

## CI


SandyLuXY (approved):
lgtm

PR #1586, opened 2026-08-17
[Fix][TTS] Rebuild Qwen3-TTS and delay Moss re-prefill after retraction

rwang5203:
## Motivation

After KV-pressure retraction, the scheduler re-prefills the prompt plus already-generated tokens. Qwen3-TTS and delay Moss previously sliced prompt-only projected buffers, so the runners could provide fewer embedding rows than the scheduler's extend length and KV allocation.

Reconstructed generated tails also need radix isolation. Both runners publish scalar scheduler IDs that do not fully identify the audio-conditioned embeddings written to KV.

## Modifications

- Rebuild Qwen3-TTS re-prefill inputs from the prompt plus recorded decode-input history, including the final queued feedback row.
- Rebuild delay-Moss inputs from `prompt_rows + output_rows` and clear feedback stranded by retraction after validating the reconstructed row count.
- Give every Qwen3-TTS and delay-Moss request lifetime a unique radix `extra_key`.
- Add focused Qwen cadence, Moss feedback-lifecycle, failure-ordering, and radix-isolation regressions.

## Related Issues

Addresses part of https://github.com/sgl-project/sglang-omni/issues/1555. Zonos2 retraction recovery remains out of scope, so this PR does not close that issue.

## Accuracy Test

```bash
pytest -q -rs tests/unit_test/moss_tts tests/unit_test/qwen3_tts
```

Result: `349 passed, 2 skipped`. Both skips require unavailable `flash_attn_varlen_func` and are unrelated to this change.

The regressions cover the observed Qwen `594`-vs-`460` mismatch, production feedback cadence, Moss reconstruction and stale-feedback lifecycle, row-mismatch failure ordering, and request-lifetime radix isolation. No model weights, logits, or sampling math change.

## Benchmark & Profiling

No benchmark was run. The request-lifetime `extra_key` disables cross-request radix-prefix sharing for Qwen3-TTS and delay Moss, while preserving same-request cache reuse across retractions. That is the conservative correctness fence. Content-addressed generated keys can recover safe cross-request sharing in a follow-up.

Issue #1555, opened 2026-08-15
[Bug] ARs can truncate re-prefill after retraction


0xtoward:
### Checklist

- [x] 1. I have searched related issues but cannot get the expected help.
- [x] 2. The bug has not been fixed in the latest version.
- [x] 3. Please note that if the bug-related issue you submitted lacks corresponding environment info and a minimal reproducible demo, it will be challenging for us to reproduce and resolve the issue, reducing the likelihood of receiving feedback.
- [x] 4. If the issue you raised is not a bug but a question, please raise a discussion at https://github.com/sgl-project/sglang-omni/discussions/new/choose Otherwise, it will be closed.
- [x] 5. Please use English, otherwise it will be closed.

## Summary

Some audio AR model runners build projected prefill embeddings by slicing a **prompt-only** buffer.

After KV-pressure retraction, the same request may be re-prefilled over:

```text
prompt + already-generated tokens
```

but the projected-embed buffer still contains only the prompt. Python slicing silently returns fewer rows than the scheduler's extend range, while `input_ids` / KV allocation still use the full length.

For `qwen3_tts`, this reproduces as a KV-write shape mismatch. In all 6 observed failures:

```text
missing projected rows == already-generated tokens
```

Examples:

| Full extend rows | Projected rows | Missing |
| ---: | ---: | ---: |
| 594 | 460 | 134 |
| 256 | 229 | 27 |
| 255 | 230 | 25 |
| 154 | 115 | 39 |
| 163 | 115 | 48 |
| 718 | 688 | 30 |

## Root cause

`qwen3_tts` slices only `prompt_input_embeds`:

```python
prompt_embeds[prefix_len : prefix_len + req_len]
```

During normal decode, projected feedback rows are consumed from `pending_feedback_queue`, but are not persisted for later replay.

So after retraction:

```text
requested re-prefill rows
= prompt tail + generated tail

available projected rows
= prompt tail only
```

`qwen3_omni` already handles this correctly by retaining decode-input history and rebuilding the generated portion during projected re-prefill.

### Verified

- `qwen3_tts`

### Same prompt-only pattern found by static audit

- `moss_tts` — slices only `prompt_rows`; generated rows live separately in `output_rows`.
- `zonos2` — also slices only `prompt_rows`, which is initialized once and never extended with generated frames.

[...]

PR #1587, opened 2026-08-17
[Fix][TTS] Retraction re-prefill for zonos2 / higgs / voxtral / fishaudio (+ black-box CI)

0xtoward:
## Motivation

After a KV-pressure retraction the scheduler re-prefills a request over
`prompt + already-generated` positions (none of these models set `Req.input_embeds`,
so sglang#14110's `output_ids` clear never fires). Runners that slice a prompt-only
structure then mishandle the generated tail — either crashing or silently dropping the
acoustic codebooks. Same family as #1586, part of #1555.

**Scope split with #1586.** #1586 (concurrent, same issue) fixes `qwen3_tts` and
delay-`moss_tts` and explicitly leaves Zonos2 out of scope. This PR stacks on it and
covers the four models it does not touch — **zonos2, higgs_tts, voxtral_tts,
fishaudio_s2_pro** — plus an end-to-end, black-box retraction CI test. (Our earlier
qwen3_tts/moss_tts commits are reverted here so the two PRs don't conflict.)

## Modifications

| Model | Bug class | Before → after | Fix |
|---|---|---|---|
| `zonos2` | whole-row + decode-state-pool + async | HTTP 500 `shape … invalid` (server-fatal under `--async-decode`) → 200 | rebuild `[G, n+1]` rows from `output_codes` with the same `embed_frames` sum the decode fed, + rehydrate the on-device decode-state pool row (generation_step / rep_hist / EOS replay) |
| `fishaudio_s2_pro` | base+overlay **shape** | HTTP 500 `mask [1,P] ≠ indexed [1,T,D]` → 200 | clamp the reference overlay to the prompt slice + rebuild the generated tail via `embed_text_dim` |
| `higgs_tts` | base+overlay **value** (drops cb1..N) | silent acoustic drift → correct | overlay the fused generated embeds on the extend tail from `output_code_buffer` |
| `voxtral_tts` | base+overlay **value** (drops cb1..N) | silent acoustic drift → correct | replay the fused decode embeds (`audio_token_embedding.sum`) on re-prefill |

All four rebuild the generated tail from host-side state (`output_codes` / code buffers)
that survives retraction. The added work runs only on a retraction re-prefill
(`if end > prompt_len`), so the fresh-prefill and decode paths are byte-identical — zero
steady-state cost. No model weights, logits, or sampling math change.

## Radix isolation (adopting #1586's insight)

#1586 adds a per-request radix `extra_key` because qwen3/moss published scalar scheduler
IDs that don't identify the audio-conditioned KV, so same-prompt requests could share a
reconstructed generated tail. We audited the four models here:

- `zonos2` content-addresses every frame's radix key via `poly_row_hash(rows)` over the
  full 10-wide frame (all 9 codebooks + text), so divergent generated tails never collide.
- `fishaudio_s2_pro` / `higgs_tts` / `voxtral_tts` already namespace the radix subtree by a
  reference-audio / voice fingerprint `extra_key` (`_ref_vq_fingerprint` hashes all
  codebooks; `_ref_audio_fingerprint`; `_voice_cache_key`).

So #1586's scalar-key cross-request collision does not reproduce here. The residual
same-reference / same-prompt corner for the three fingerprint-namespaced models can take
the same per-request `extra_key` fence as a fast-follow if maintainers prefer it over the
existing cross-request prompt sharing.

## Test

`tests/test_model/test_retract_reprefill_ci.py` — one model-agnostic, **black-box**,
end-to-end test for the whole family:

- **Trigger (unified):** admin `POST /pause_generation {"mode":"retract"}` retracts every
  running request, then `/continue_generation`; `SGLANG_RADIX_FORCE_MISS=1` forces the full
  re-prefill. The retract is paced by **generated audio, not wall-clock** — a streaming pacer
  fires it once a fixed amount of audio has been produced — so it lands at the same
  generation depth whether the model runs at 1× or 4× realtime.
- **Liveness (always, model-agnostic):** no fatal signature in the server log after the
  retract, every in-flight request 200, and a fresh post-retract request 200.
- **Content (opt-in):** transcribe the audio via `benchmarks.tasks.asr` — the same WER
  layer as the TTS CI, defaulting to its Qwen3-ASR (`Qwen/Qwen3-ASR-1.7B`; whisper-large-v3
  also works over the same path) — and assert WER ≤ a per-model budget.

Per model: `RETRACT_MODEL_PATH=… RETRACT_CI_MODEL=zonos2 RETRACT_REF_AUDIO=… pytest -s -x tests/test_model/test_retract_reprefill_ci.py`

## Validation

- `zonos2` and `fishaudio_s2_pro`: reproduced the pre-fix crash under the unified trigger
  (distinct signatures) on a real GPU, then confirmed 200 + intelligible audio after the
  fix (**WER 0.00**, dev-checked with whisper-large-v3; the committed test measures via the
  TTS-CI Qwen3-ASR layer).
- The committed CI test is verified end-to-end on a real GPU with the frame-paced retract
  trigger: it **fails on the unfixed zonos2 runner** (catches the crash) and **passes on the
  fix**.
- **Content models (higgs / voxtral) caveat:** the bug drops the acoustic codebooks cb1..N
  while the words survive, so **WER is blind to it** (≈0 before and after). The right
  detectors are the existing TTS-CI **UTMOS** and **speaker-similarity (ASV)** metrics; the
  black-box test here guards the crash family end-to-end and the WER budget for the crash
  models. The content fixes reconstruct exactly the fused embedding the decode fed.

## Related

- Issue #1555
- Stacks on / complements #1586 (qwen3_tts + delay moss + per-request radix isolation)

### DIFF
--- a/sglang_omni/models/fishaudio_s2_pro/model_runner.py
+++ b/sglang_omni/models/fishaudio_s2_pro/model_runner.py
@@ -3,6 +3,8 @@
 
 from __future__ import annotations
 
+import logging
+import os
 from typing import Any
 
 import torch
@@ -10,6 +12,8 @@ import torch
 from sglang_omni.model_runner.base import ModelRunner
 from sglang_omni.models.fishaudio_s2_pro.sglang_model import _NO_SEED
 from sglang_omni.sampling.seed import resolve_row_seed
+
+logger = logging.getLogger(__name__)
 
 
 def collect_s2pro_step_outputs(
@@ -187,7 +191,15 @@ class FishS2ProModelRunner(ModelRunner):
                 vq_mask = vq_mask.squeeze(0)
 
             prefix_len = len(req.prefix_indices)
-            mask_slice = vq_mask[prefix_len : prefix_len + req_len]
+            # vq_mask_tokens covers ONLY the prompt. A retraction re-prefill
+            # extends past it into already-generated frames, so slicing the mask
+            # by the full extend length truncates it to the prompt length while
+            # req_embeds keeps the full length -> embed_text_dim indexes a
+            # [1, extend] tensor with a [1, prompt] mask -> shape-mismatch crash.
+            # Clamp the reference overlay to the prompt slice; the generated
+            # frames are rebuilt by the generated-tail loop below.
+            ref_len = max(0, min(req_len, int(vq_mask.shape[0]) - prefix_len))
+            mask_slice = vq_mask[prefix_len : prefix_len + ref_len]
             if not bool(mask_slice.any()):
                 offset += req_len
                 continue
@@ -204,7 +216,7 @@ class FishS2ProModelRunner(ModelRunner):
             num_vq_in_slice = int(mask_slice.sum().item())
             vq_slice = vq_parts_flat[vq_before : vq_before + num_vq_in_slice]
 
-            req_embeds = text_embeds[offset : offset + req_len]
+            req_embeds = text_embeds[offset : offset + ref_len]
             vq_embeds = self.model._audio_decoder.embed_text_dim(
                 req_embeds.unsqueeze(0),
                 vq_slice,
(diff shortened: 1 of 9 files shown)


sglang_omni/models/voxtral_tts/model_runner.py
```diff
@@ -92,6 +92,28 @@ def _build_prefill_input_embeds(
...
+                    if not (0 <= emb_lo <= emb_hi <= len(gen)):
+                        raise RuntimeError(
+                            "voxtral re-prefill needs generated embeds "
+                            f"[{emb_lo}:{emb_hi}] of {len(gen)} "
+                            f"(prefix={prefix_len}, req_len={req_len}, prompt={prompt_len})"
+                        )
+                    stacked = torch.stack(gen[emb_lo:emb_hi], dim=0).to(
+                        device=input_embeds.device, dtype=input_embeds.dtype
+                    )
+                    input_embeds[offset + local_lo : offset + req_len] = stacked
```
luojiaxuan, line 116: Please clear pending_feedback_queue after replay; otherwise the first resumed decode consumes the old row again.

sglang_omni/models/higgs_tts/model_runner.py
```diff
@@ -458,8 +458,46 @@ def _build_prefill_input_embeds(
...
+        # Without it a retraction re-prefill leaves generated positions as
+        # embed_tokens(cb0) only -- dropping cb1..N -> silent audio drift. The
+        # extend always ends at the current sequence end, so the generated frames
+        # are the last min(extend_len, n_gen) rows of each request's extend,
+        # mapping 1:1 onto the output_code_buffer tail.
+        offset = 0
+        for sched_req in requests:
+            data = sched_req.data
+            ext_len = int(data.req.extend_range.length)
+            gen = self._generated_codes(data)
```
luojiaxuan, line 473: Higgs still publishes cb0 while KV depends on cb1..N. The reference key does not isolate same-reference requests; please add a request key or full-frame radix ID.

sglang_omni/models/higgs_tts/model_runner.py
```diff
@@ -458,8 +458,46 @@ def _build_prefill_input_embeds(
...
+            if gen is not None and int(gen.shape[0]) > 0 and ext_len > 0:
+                n_gen = int(gen.shape[0])
+                k = min(ext_len, n_gen)
+                sel = gen[n_gen - k : n_gen].to(device=device, dtype=torch.long)
```
luojiaxuan, line 477: This assumes every extend reaches the sequence tail; please map generated rows from extend_range.start so chunked re-prefill cannot overwrite prompt rows.

sglang_omni/models/fishaudio_s2_pro/model_runner.py
```diff
@@ -214,8 +222,47 @@ def _build_prefill_input_embeds(
...
+        # the exact same fusion (embed_text_dim == the decode formula) from the
+        # stored generated codes. The extend always ends at the sequence end, so
+        # the generated frames are the last min(extend_len, n_gen) rows, mapping
+        # 1:1 onto the output_codes tail. output_codes is populated only during
+        # decode, so this is a no-op on a fresh prefill (retract-only cost).
+        offset = 0
+        for sched_req in requests:
+            data = sched_req.data
+            ext_len = int(data.req.extend_range.length)
+            codes = self._generated_codes_tail(data, ext_len, device)
```
luojiaxuan, line 239: Please map generated rows from the absolute extend range; this tail slice is wrong for a non-final chunked re-prefill.

luojiaxuan (commented):
Please address the inline issues.

tests/test_model/test_retract_reprefill_ci.py
```diff
@@ -0,0 +1,413 @@
...
+        ]
+        # Pace the retract by generated audio, not seconds: a streaming pacer runs
+        # the same workload in lockstep with the batch; retract once it has produced
+        # RETRACT_MIN_STREAM_BYTES. Capped by RETRACT_MAX_WAIT_S so it never hangs.
+        pacer = asyncio.create_task(
+            _pace_until(s, texts[0], preset, voice_ok, RETRACT_MIN_STREAM_BYTES, fired)
+        )
+        with contextlib.suppress(asyncio.TimeoutError):
+            await asyncio.wait_for(fired.wait(), timeout=RETRACT_MAX_WAIT_S)
+        await _admin(s, "pause_generation", {"mode": "retract"})  # retract ALL running
```
luojiaxuan, line 302: This can pass without retracting any request because the pacer always fires and admin results are ignored; please assert the threshold and num_paused_requests > 0.

tests/test_model/test_retract_reprefill_ci.py
```diff
@@ -0,0 +1,413 @@
...
+            f"budget {preset.wer_max} (all={[(i, round(w, 3)) for i, w in wers]})"
+        )
+    finally:
+        if asr_proc is not None:
+            _stop(asr_proc, asr_log_f)
+
+
+# ---- the test ---------------------------------------------------------------
+def test_retraction_reprefill_survives(tmp_path):
+    if not MODEL_PATH:
```
luojiaxuan, line 363: No workflow supplies RETRACT_MODEL_PATH, so this test always skips in CI.

sglang_omni/models/fishaudio_s2_pro/model_runner.py
```diff
@@ -187,7 +187,15 @@ def _build_prefill_input_embeds(
                 vq_mask = vq_mask.squeeze(0)
 
             prefix_len = len(req.prefix_indices)
-            mask_slice = vq_mask[prefix_len : prefix_len + req_len]
+            # vq_mask_tokens covers ONLY the prompt. A retraction re-prefill
```
luojiaxuan, line 190: Please add the # note (luojiaxuan): prefix to every new comment block in this file.

sglang_omni/models/higgs_tts/model_runner.py
```diff
@@ -458,8 +458,46 @@ def _build_prefill_input_embeds(
             text_embeds[mask_idx] = embed.to(text_embeds.dtype)
             offset = end
 
+        # Rebuild ALREADY-GENERATED audio positions with their fused all-codebook
```
luojiaxuan, line 461: Please add the # note (luojiaxuan): prefix to every new comment block in this file.

sglang_omni/models/voxtral_tts/model_runner.py
```diff
@@ -92,6 +92,28 @@ def _build_prefill_input_embeds(
             req = data.req
             req_len = int(req.extend_range.length)
             prefix_len = len(req.prefix_indices)
+            # Replay the fused (all-codebook) embeds for already-generated frames
```
luojiaxuan, line 95: Please add the # note (luojiaxuan): prefix to every new comment block in this file.

sglang_omni/models/voxtral_tts/request_builders.py
```diff
@@ -23,6 +23,12 @@ class VoxtralSGLangRequestData(SGLangARRequestData):
     audio_token_id: int = 24
     output_codes: list[torch.Tensor] = field(default_factory=list)
     pending_feedback_queue: Any = field(default_factory=collections.deque)
+    # Persistent history of the fused (all-codebook) decode-input embedding fed
```
luojiaxuan, line 26: Please add the # note (luojiaxuan): prefix to every new comment block in this file.

sglang_omni/models/zonos2/model_runner.py
```diff
@@ -115,9 +123,86 @@ def _build_prefill_embeds(self, forward_batch, requests) -> torch.Tensor:
                         s.to(model.speaker_projection.weight.dtype)
                     )
                     emb[pos] = s.to(emb.dtype)
+
+            # Retraction re-prefill: the extend runs past the prompt into
```
luojiaxuan, line 127: Please add the # note (luojiaxuan): prefix to every new comment block in this file.

tests/test_model/test_retract_reprefill_ci.py
```diff
@@ -0,0 +1,413 @@
...
+        assert worst <= preset.wer_max, (
+            f"retract re-prefill drifted: req_{worst_i} WER={worst:.3f} > "
+            f"budget {preset.wer_max} (all={[(i, round(w, 3)) for i, w in wers]})"
+        )
+    finally:
+        if asr_proc is not None:
+            _stop(asr_proc, asr_log_f)
+
+
+# ---- the test ---------------------------------------------------------------
```
luojiaxuan, line 361: Please add the # note (luojiaxuan): prefix to every new comment block in this file.

PR #1591, opened 2026-08-17
Enhance pre-LM encoder caching with pinned host memory

0xjeffro:
## Motivation

This PR is part of the Whisper-large-v3 performance work.

Across our ASR models we use a pre-LM encoder service: the audio encoder runs before LM admission, on a dedicated worker thread and CUDA stream, batching across concurrent requests, with a CPU LRU cache keyed by audio fingerprint storing the encoder outputs. Repeated audio can then skip mel extraction and the encoder forward entirely, which improves throughput.

Whisper is special here: HF's WhisperFeatureExtractor pads every clip to 30 s, so the encoder output is always `[1500, d_model]` regardless of the actual audio duration. For large-v3 one cache entry is 1500 × 1280 × fp16 = 3.84 MB. Models like Qwen3-ASR scale their embeddings with audio duration, so on a short-clip corpus like SeedTTS (~4 s per clip) one entry is only ~0.2 MB. With the same caching strategy, Whisper moves roughly 20× more bytes per copy than Qwen3.

In the current implementation, both directions of the cache's GPU↔CPU traffic are synchronous and go through pageable memory:

- put (D2H): `StageOutputCache._detach_value` stores entries with `value.to("cpu")` into ordinary (pageable) memory. This is a synchronous copy on the encoder worker thread: after encoding a batch, the worker pays N serial copies before it can `set_result`, and cannot encode the next batch in the meantime.
- get (H2D): `attach_embedding` does call `.to(device, non_blocking=True)`, but because the source is pageable memory, CUDA silently ignores `non_blocking` and the copy degrades to a synchronous one, blocking the request-build worker.

## Modifications

The core idea of this PR is to back the Whisper pre-LM encoder cache with page-locked (pinned) host memory, making the GPU↔CPU cache copies truly asynchronous and improving cache performance, especially for long-audio workloads. Implementation details:

1. Add a `pin_memory` option to `StageOutputCache` (default false). When enabled, tensors put into the cache are stored in pinned memory; tensors that are already pinned are stored as-is without another copy; if pinned allocation fails, the cache falls back to pageable memory with a corresponding warning log. Since this optimization only concerns GPU↔CPU copies, it automatically degrades to a no-op on non-CUDA backends.
2. Add a `stage_host_copy` hook to the `PreLMEncoderService` base class. It is invoked per item after `split_embeddings` and before `synchronize_batch`. At that point the base class has not yet exited `_batch_context`, so the current thread's CUDA stream is still the dedicated encoder stream: a non-blocking GPU→CPU copy issued here queues onto the encoder stream and is retired by the subsequent `synchronize_batch()`. The base class collects the per-item return values into a list and, once past the barrier, passes each one to `cache_embedding(item, embedding, host_copy)` as the new third argument.
3. Following the above: the Whisper pre-LM encoder service is not yet fully abstracted, so this PR only affects Whisper. For example, in `cache_embedding(item, embedding, host_copy)`, when `host_copy` is None the cache automatically falls back to storing `embedding`; the behavior of the other ASR models is unchanged.
4. Prewarm the pinned memory pool at service start-up: allocate pinned buffers once, using the existing `cache_capacity_entries` value, so the per-allocation cost is not paid on the hot path.

## Related Issues

#1396 

## Benchmark & Profiling

| Concurrency | Config | Success | Throughput (req/s) | Mean latency (ms) | P50 (ms) | P95 (ms) | P99 (ms) | Mean RTF | P95 RTF | Corpus WER |
|---:|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 16 | main #1585 | 3264/3264 | 71.291 | 223.97 | 221.35 | 302.67 | 340.92 | 0.0485 | 0.0670 | 0.0138 |
|  | PR #1591 pinned | 3264/3264 | 74.872 | 212.87 | 211.22 | 282.08 | 326.75 | 0.0461 | 0.0633 | 0.0137 |
|  | Δ (PR / main) |  | 1.05× | 1.05× | 1.05× | 1.07× | 1.04× | 1.05× | 1.06× | −0.003 pt |
| 32 | main #1585 | 3264/3264 | 93.264 | 341.26 | 337.43 | 457.67 | 544.83 | 0.0738 | 0.1006 | 0.0137 |
|  | PR #1591 pinned | 3264/3264 | 97.782 | 325.28 | 320.67 | 433.71 | 521.07 | 0.0703 | 0.0952 | 0.0137 |
|  | Δ (PR / main) |  | 1.05× | 1.05× | 1.05× | 1.06× | 1.05× | 1.05× | 1.06× | +0.000 pt |
| 64 | main #1585 | 3264/3264 | 102.096 | 621.87 | 616.87 | 874.79 | 1027.82 | 0.1341 | 0.1836 | 0.0137 |
|  | PR #1591 pinned | 3264/3264 | 108.228 | 585.54 | 583.98 | 818.35 | 980.29 | 0.1263 | 0.1738 | 0.0137 |
|  | Δ (PR / main) |  | 1.06× | 1.06× | 1.06× | 1.07× | 1.05× | 1.06× | 1.06× | −0.003 pt |
## Checklist

- [x] Add unit tests.

## CI


sglang_omni/models/whisper_asr/encoder_service.py
```diff
@@ -386,14 +457,44 @@ def split_embeddings(
...
+        item: Any,
+        embedding: torch.Tensor,
+        host_copy: torch.Tensor | None = None,
+    ) -> None:
         key = self._cache_key(item)
-        if key is not None:
-            self._cache.put(key, embedding)
+        if key is None:
+            return
+        # host_copy is complete here (synchronize_batch ran in between) and
```
SandyLuXY: Please follow the comment style: `note (xxx): `

sglang_omni/scheduling/stage_cache.py
```diff
@@ -50,17 +77,22 @@ def __init__(
...
             raise ValueError("max_bytes must be non-negative")
         if isinstance(cache_device, str):
             cache_device = torch.device(cache_device)
+        if pin_memory and (cache_device is None or cache_device.type != "cpu"):
+            raise ValueError("pin_memory requires cache_device='cpu'")
         self._cache: OrderedDict[str, _CacheEntry] = OrderedDict()
         self.max_size = max_size
         self.max_bytes = max_bytes
         self.cache_device = cache_device
+        # Pinning needs a CUDA context to be meaningful, so on CPU-only hosts it degrades to plain pageable storage.
```
SandyLuXY: same here

tests/unit_test/scheduling/test_pre_lm_encoder.py
```diff
@@ -92,7 +101,22 @@ def test_successful_dispatch_attaches_and_caches() -> None:
...
+
+
+def test_stage_host_copy_runs_in_batch_context_and_reaches_cache() -> None:
+    service = _Service()
+    service.stage_host_copies = True
+    try:
+        future = service._submit(4)
+
+        assert future.result(timeout=2) == 8
+        # Staged inside the batch context (before its exit), delivered to
```
SandyLuXY: same here

tests/unit_test/scheduling/test_stage_cache.py
```diff
@@ -145,3 +145,70 @@ def remover() -> None:
...
+@_requires_cuda
+def test_pinned_cache_stores_device_tensors_in_page_locked_memory() -> None:
+    cache = StageOutputCache(cache_device="cpu", pin_memory=True)
+    src = torch.arange(16, dtype=torch.float16, device="cuda").reshape(4, 4)
+    cache.put("dev", src)
+    cached = cache.get("dev")
+    assert cached is not None
+    assert cached.device.type == "cpu" and cached.is_pinned()
+    assert torch.equal(cached, src.cpu())
+    # Nested containers are pinned too.
```
SandyLuXY: same here

tests/unit_test/whisper_asr/test_encoder_service.py
```diff
@@ -112,11 +113,125 @@ def _make_service(
...
+def _cached_entry(
+    service: WhisperPreLMEncoderService, fingerprint: str
+) -> torch.Tensor:
+    cached = service.lookup_cached_embedding(fingerprint, _TOKENS)
+    assert cached is not None
+    return cached
+
+
+def test_pinning_is_off_without_cuda_device() -> None:
+    # The stub model lives on CPU: there is nothing to DMA against, so the
```
SandyLuXY: same here

tests/unit_test/whisper_asr/test_encoder_service.py
```diff
@@ -112,11 +113,125 @@ def _make_service(
...
+    assert service.pin_host_memory is True
+    item = _make_item(fingerprint="pinned", fill=2.0)
+    service.encode_item(item)
+    cached = _cached_entry(service, "pinned")
+    assert cached.device.type == "cpu"
+    assert cached.is_pinned()
+    assert cached.dtype == torch.float16
+    torch.cuda.synchronize()
+    assert torch.equal(cached.to("cuda"), item.precomputed_embeddings)
+    # No fingerprint means nothing will be cached, so nothing is staged.
```
SandyLuXY: same here

tests/unit_test/whisper_asr/test_encoder_service.py
```diff
@@ -112,11 +113,125 @@ def _make_service(
...
+    service = _make_service(_cuda_model())
+    assert service.pin_host_memory is True
+
+    def _boom(_tokens: int) -> torch.Tensor:
+        raise RuntimeError("cudaHostAlloc failed")
+
+    monkeypatch.setattr(service, "_new_pinned_host", _boom)
+    item = _make_item(fingerprint="fallback", fill=4.0)
+    service.encode_item(item)
+    # The entry is still cached, just in pageable memory, and the service
```
SandyLuXY: same here

SandyLuXY (changes requested):

SandyLuXY (approved):
lgtm

PR #1553, opened 2026-08-15
[Whisper ASR] Enable async decode by default

Dayuxiaoshui:
## Motivation

Whisper currently completes every decode step synchronously, which keeps host-side result processing on the decode critical path. This PR enables the shared one-step-lookahead path for batches of two or more so GPU execution for the current step can overlap host processing for the previous step. Batch size one remains synchronous to avoid adding lookahead overhead when no overlap is available.

The serving defaults are adjusted with the decode path so Whisper has enough running-request and atomic-prefill capacity to form useful decode batches. Encoder CUDA Graph capture is also limited to batches that the configured serving path can actually reach, avoiding startup time and GPU memory for graphs that cannot replay.

## Modifications

- Enable asynchronous decode for Whisper by default with `async_decode_min_batch_size=2`, while preserving the synchronous batch-size-one path.
- Raise `max_running_requests` from 16 to 32 and the atomic `max_prefill_tokens` budget from 4,096 to 6,144, allowing up to four default 1,504-token Whisper requests in one LM prefill batch.
- Add Whisper to the shared `--decode-mode sync|async` and `--async-lookahead-min-batch-size` CLI controls and keep the Typer help text consistent with the supported-model list.
- Resolve default encoder CUDA Graph buckets from `pre_lm_max_batch_size` and `max_running_requests` when pre-LM encoding is enabled, or from `max_prefill_tokens // encoder_token_count` when it is disabled. The default pre-LM configuration captures batches 1, 2, 4, and 8.
- Cover default configuration, CLI overrides, scheduler argument forwarding, graph capture ordering, and serving-reachable bucket resolution with unit tests.
- Document the serving behavior, synchronous fallback, reproducible benchmark procedure, accuracy results, performance results, and runtime path evidence.

## Related Issues

Part of https://github.com/sgl-project/sglang-omni/issues/1396

## Accuracy Test

The end-to-end comparison used the pinned `openai/whisper-large-v3` revision `06f233fe06e710322aca913c1bc4249a0d71fce1` and the 128-sample SeedTTS EN subset at dataset revision `27f4c1adee83b5b29b7c4b375f6b976324bda308` in FP16 on one NVIDIA H200. Each decode mode ran one discarded warmup and three measured repeats at concurrencies 1, 2, 4, 8, 16, and 32.

| Result | Synchronous | Asynchronous |
|---|---:|---:|
| Measured requests | 2,304 | 2,304 |
| Successful requests | 2,304 | 2,304 |
[...]

PR #1592, opened 2026-08-18
[Perf][W-PR8] Route Whisper suppress tokens through the shared sparse suppression path

MrlixiangWE:
## Motivation

Whisper passes `generation_config.suppress_tokens` to SGLang as a `SamplingParams.logit_bias`
(`whisper_asr/request_builders.py:64`). For `openai/whisper-large-v3` that is 88 ids over a 51866
token vocabulary, and SGLang materialises it per request:

```python
# sglang 0.5.16, python/sglang/srt/sampling/sampling_batch_info.py:132
logit_bias = torch.zeros(len(reqs), vocab_size, device=device)
for i, r in enumerate(reqs):
    if r.sampling_params.logit_bias is not None:
        for key, value in r.sampling_params.logit_bias.items():
            logit_bias[i, int(key)] = value          # <- line 137
```

py-spy on the loaded ASR stage worker (1x A800 80GB, whisper-base, SeedTTS-50 EN at concurrency 32,
20 s at 50 Hz after a 25 s settle; 1231 samples, 1173 of them non-idle = 95.28838%) puts
**8.86616%** of non-idle samples on line 137 alone, plus **0.42626%** on the dense `logits.add_()`
that applies it. `filter_batch` gathers the same `[batch, vocab]` tensor whenever a request finishes
and `merge_batch` concatenates it on every prefill merge.

Timed in isolation, the construction alone costs (medians of 50 repeats x 2 rounds, arms interleaved
and reversed between rounds):

| batch | dense build | id tuple instead | saved per request |
|---:|---:|---:|---:|
| 1 | 1198.485 us | 9.899 us | 1188.586 us |
| 8 | 9429.281 us | 41.997 us | 1173.410 us |
| 32 | 37539.346 us | 152.820 us | 1168.329 us |
| 64 | 75128.059 us | 300.005 us | 1169.188 us |

So 1168.329-1188.586 us of scheduler-thread time per request, essentially flat in batch size.

The runtime already has the sparse path for this: `ModelRunner._apply_codec_suppress_tokens` caches
one device tensor of ids per (content, vocab, device) for the whole fleet. Qwen3-Omni and Qwen3-TTS
use it; Whisper was opted out with an explicit `req._codec_suppress_tokens = None`.

Routing Whisper onto that path **alone** made whisper-large-v3 slower, not faster: **-9.26706%**
req/s at concurrency 8. Chasing that produced the second commit, and it is the more valuable half of
this PR.

## Modifications

**1. `whisper_asr/request_builders.py` — use the shared path.** `_build_logit_bias` becomes
`_build_suppress_token_ids`, which builds the id tuple once per adapter and hands it to each request
by reference, so there is no per-request copy. Transcription and translation requests set
`req._codec_suppress_tokens` and pass `logit_bias=None`.

**Language identification deliberately keeps its dense `logit_bias`.** `language_token_bias` is a
*positive* +100 bias that keeps one language token ahead of the others — it is not a suppression,
and such a request samples exactly one token.

**2. `model_runner/base.py` — stop the shared path synchronizing on every sampling call.** It wrote
its `-inf` through advanced indexing with a Python scalar, which stages that scalar through a
blocking host-to-device copy. Queue a known amount of GPU work, then time the host side of each
candidate *without* synchronizing after it — an op that only enqueues returns in microseconds, an
op that blocks returns only once the queue drains (A800, `[8, 51866]` fp16, 88 ids):

| op | host time | queued GPU work | |
|---|---:|---:|---|
| `logits.add_(dense_bias)` — what main does today | 0.008 ms | 136.091 ms | enqueue only |
| `logits[:, toks] = -inf` — the path as it stood | **24.793 ms** | 25.102 ms | **blocks** |
| `logits.index_fill_(1, toks, -inf)` | 0.010 ms | 25.442 ms | enqueue only |
| `logits[rows[:,None], toks[None,:]] = -inf` | **24.739 ms** | 25.418 ms | **blocks** |
| `torch.tensor(rows, device="cuda")` — the row-index helper | **24.724 ms** | 25.205 ms | **blocks** |

So every decode step gave up one CPU/GPU overlap. The fix is `index_fill_` for the uniform case and
a per-row `index_fill_` for the mixed-set case, which also removes the row-index tensor.

This is also why the effect looked model-dependent: whisper-base's decode steps are short, so there
is little queued work to stall on and the per-request saving still won, while large-v3's steps are
long enough that the stall dominated. The large-v3 profile at concurrency 8 shows the trade directly
— the dense fill is **2.91080%** of non-idle samples on main, against **23.37553%** for
`logits[:, toks_t] = -inf` before this fix.

Qwen3-Omni and Qwen3-TTS use this path today, so they get the same fix.

## Related Issues

#1396 — W-PR8.

The suppression this moves onto is safe under async decode by construction: the note on
`ThinkerModelRunner._sample_lookahead` records that "only static suppress tokens are lag-safe",
which is exactly what this is. That matters now that #1553 made async decode the Whisper default.

## Accuracy Test

**925 tests pass** on 1x A800 80GB (python 3.12, torch 2.11.0+cu130, sglang 0.5.16):

```bash
pytest -q tests/unit_test/whisper_asr tests/unit_test/model_runner \
  tests/unit_test/qwen3_tts tests/unit_test/qwen3_omni
```

The Qwen3-TTS and Qwen3-Omni suites are included because they exercise the shared suppression path
that the second commit changes. `pre-commit run --files` passes on every changed file.

Suppression had no test coverage at all before this — every existing Whisper case built with
`suppress_tokens=None`. Six new Whisper cases cover the routing and the `>= 0` filter, the
no-suppress-set case staying inert, the id tuple being shared by reference, language identification
keeping its dense bias, an assertion that the shared cache holds **one** device tensor for four
requests, and **argmax equivalence** against the `-1e9` bias this replaces — on a mixed batch of two
suppressed rows and one language-identification row, with a suppressed id forced to be the argmax so
a silent no-op cannot pass. Five new shared-path cases cover the uniform fill, two different
suppress sets in one batch not bleeding across rows, an opted-out row staying untouched,
out-of-vocabulary ids being dropped, and cache reuse across calls.

**Corpus WER is identical in every measured cell of every run below**: `0.034226` on whisper-base
and `0.00744` on whisper-large-v3, both arms, every concurrency, every repeat.

## Benchmark & Profiling

1x A800 80GB, SeedTTS-50 EN. Both arms on the same GPU and checkpoint, interleaved, with the arm
order reversed between rounds; 5 repeats plus a discarded warmup per concurrency per round. p is a
two-sided Mann-Whitney U on the pooled per-repeat throughputs. Percentages are given to five
decimals so that no cell reads as a suspiciously round number.

**whisper-base**, 2 rounds, 10 pooled repeats per cell:

| concurrency | req/s before | req/s after | throughput | mean latency | p |
| 1 | 16.55194 | 17.16279 | **+3.69050%** | -3.55750% | 0.05878 |
| 8 | 54.62768 | 57.85167 | **+5.90174%** | -5.59943% | 0.00051 |
| 32 | 73.34744 | 81.39583 | **+10.97298%** | -10.00347% | 0.00016 |

**whisper-large-v3**, 2 rounds, 10 pooled repeats per cell:

| concurrency | req/s before | req/s after | throughput | mean latency | p |
| 8 | 24.24201 | 24.17877 | -0.26085% | +0.39662% | 0.93974 |
| 32 | 38.94003 | 39.81751 | **+2.25342%** | -2.78293% | 0.09630 |

Note that at a fixed concurrency the mean-latency column is not independent evidence: throughput x
latency is pinned by Little's law, and the product of the two ratios is 1.000017 / 0.999718 /
0.998718 for the three whisper-base cells. The two columns are one result reported two ways.

Concurrency 1 on large-v3 was the noisiest cell in the set, so I re-measured it on its own with
4 rounds, 20 pooled repeats per arm: **7.33024 -> 7.13525 req/s, -2.66005%, p=0.15954 — not
significant.** All twenty candidate repeats fall in 6.94488-7.29812 req/s. Sixteen of the twenty
baseline repeats fall in the same band, 7.06056-7.29753; the other four are one faster server
instance at 7.64044 / 7.84864 / 8.17330 / 8.47699 that pulls the baseline mean up. On medians the
gap is **-0.45919%**. Batch 1 is also where this change has least to offer, since a single dense row
is the cheapest case for the construction being removed.

The profile predicts the serving result. On whisper-base the dense construction disappears from the
head profile and the sparse write appears in its place — `sampling_batch_info.py:137` 8.86616% plus
`:297` 0.42626% before, against `model_runner/base.py:1002` 3.92478% after: a net **5.36764** points
of the worker's non-idle time, which predicts 5.36764 / (100 - 5.36764) = **+5.67405%** throughput.

Worth one line on its own: with only the first commit, the three whisper-base cells measured
**-0.34600% / +3.38524% / +5.65386%**. Removing the per-step stall roughly doubled the gain there as
well — the stall was present on whisper-base too, just smaller than the saving it was hiding behind.

## Checklist

- [x] Add unit tests.

### DIFF
--- a/sglang_omni/model_runner/base.py
+++ b/sglang_omni/model_runner/base.py
@@ -997,9 +997,17 @@ class ModelRunner:
                 row_groups[key] = (toks_t, [row_idx])
             else:
                 group[1].append(row_idx)
+        # Note (Lixiang Li): index_fill_ rather than advanced-index assignment.
+        # Assigning a Python scalar through advanced indexing stages it through a
+        # host-to-device copy and blocks until the stream drains — measured at
+        # 24.8 ms behind 25.1 ms of queued work, against 0.010 ms for
+        # index_fill_. This runs once per sampling call, so the stall cost one
+        # full CPU/GPU overlap per decode step. Same reason the mixed-set branch
+        # fills per row instead of building a row-index tensor: that
+        # torch.tensor(..., device=...) is itself a blocking host-to-device copy.
         for toks_t, rows in row_groups.values():
             if len(rows) == logits.shape[0]:
-                logits[:, toks_t] = float("-inf")
+                logits.index_fill_(1, toks_t, float("-inf"))
             else:
-                rows_t = torch.tensor(rows, dtype=torch.long, device=device)
-                logits[rows_t[:, None], toks_t[None, :]] = float("-inf")
+                for row_idx in rows:
+                    logits[row_idx].index_fill_(0, toks_t, float("-inf"))
--- a/sglang_omni/models/whisper_asr/request_builders.py
+++ b/sglang_omni/models/whisper_asr/request_builders.py
@@ -61,11 +61,14 @@ def _resolve_language(value: Any) -> str:
     return _LANGUAGE_ALIASES.get(language, language)
 
 
-def _build_logit_bias(generation_config: GenerationConfig) -> dict[str, float] | None:
+def _build_suppress_token_ids(
+    generation_config: GenerationConfig,
+) -> tuple[int, ...] | None:
     suppress_tokens = generation_config.suppress_tokens
     if not suppress_tokens:
         return None
-    return {str(int(token_id)): -1.0e9 for token_id in suppress_tokens if token_id >= 0}
+    token_ids = tuple(int(token_id) for token_id in suppress_tokens if token_id >= 0)
+    return token_ids or None
 
 
 def _build_prefix_tokens(tokenizer: Any, *, language: str, task: str) -> list[int]:
@@ -124,7 +127,13 @@ def make_whisper_scheduler_adapters(
 ) -> tuple[
     Callable[[StagePayload], WhisperASRRequestData], Callable[[Any], StagePayload]
 ]:
-    logit_bias = _build_logit_bias(generation_config)
+    # note (Lixiang Li): the suppress set is a model constant, so build the id
+    # tuple once here and let the shared sparse suppression path in
+    # ModelRunner._apply_codec_suppress_tokens cache one device tensor for the
+    # whole fleet. Passing it as a SamplingParams logit_bias instead made
+    # SGLang materialize a dense [batch, vocab] float32 tensor and fill it with
+    # one scalar store per suppressed id per request.
+    suppress_token_ids = _build_suppress_token_ids(generation_config)
     # note (Dayuxiaoshui): set_prefix_tokens mutates shared tokenizer state
     # across request-build workers.
     tokenizer_lock = Lock()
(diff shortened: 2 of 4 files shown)


MrlixiangWE:
Closing this — W-PR8 is already covered by #1589, which removes the same dense `logit_bias` construction and applies the suppress set inside `forward()` so the decode CUDA graph captures it. That is the better place for it. Sorry for the duplicate.

Issue #1593, opened 2026-08-18
[Comm] Record the failure paths, and run a soak when the comm layer changes


jzheng17:
### Summary

This issue asks for two things.

**First, trace events on the paths that fail.** The comm data plane emits eight events today. All eight record success. The layer emits nothing when a transfer fails. It emits nothing when it keeps slots after a failure. It emits nothing when it keeps a pending entry after a failure. If any of these happen in production, the logs show nothing.

We scoped this to four events on the slot-pool path when we opened the issue. Since then we read the paged KV path and found a larger gap, described below. #1595 implements the KV part.

**Second, a soak that runs only when the comm layer changes.** We ran one for 8 hours and it answered a question that no unit test answers: does the layer leak resources under sustained load? A soak costs a GPU for hours, so it must not run on every pull request. It should run when a pull request touches `sglang_omni/comm/` or `sglang_omni/relay/`.

The two parts connect. The soak reads the trace events. Without events on the failure paths, a soak can only report that nothing crashed.

Neither part changes behaviour. Neither part asks you to choose a recovery policy.

### The paged KV path has no events at all

The eight events all sit on the slot-pool path. Paged KV transfer uses pre-registered pools and page indices instead of slots, so none of them fire for it. We checked every function: thirteen functions across `comm/engine.py` and `relay/cuda_ipc.py`, about 500 lines, contain no `_comm_trace` call. `send_kv_pages` is 113 lines, `get_kv_pages` is 108, `prepare_kv_receive` is 58, and none of them record anything.

`get_kv_pages` also constructs its operation with `emit_trace=False`, so the one place where an existing event could have fired is switched off.

Prefill/decode disaggregation will run its traffic over this path. Today that traffic would produce no log line, on success or on failure.

Two smaller gaps sit next to it.

**The router never records which transport an edge selected.** An edge that falls back keeps working, so every counter still reads as healthy. We lost an eight-hour soak to this: one edge moved from `cuda_ipc` to `shm` after the first request, the run reported balanced ACK counts and flat memory, and the path under test carried no traffic. Nothing in the log said so.

**The direct-IPC stream path records only its sender.** `send_stream_chunk` and `_run_stream_send` emit; `outbound_stream` and `read_stream_chunk` do not. That path has no byte ceiling, and measuring the held bytes needs both rates, so the missing consumer event makes the growth unmeasurable.

[...]

PR #1595, opened 2026-08-18
[Comm] Trace the paged KV transfer path, transport selection, and stream reads

jzheng17:
## Motivation

The paged KV transfer path emits no trace events. Thirteen functions across `comm/engine.py` and `relay/cuda_ipc.py`, about 500 lines, contain no `_comm_trace` call. The eight events the comm layer already has all sit on the slot-pool path, and paged KV transfer uses pre-registered pools and page indices instead of slots, so none of them fire for it.

Prefill/decode disaggregation will run its traffic over this path. Today that traffic would produce no log line at all.

Two smaller gaps come from the same measurement work:

- **The router never records which transport an edge picked.** An edge that falls back keeps working, so every counter still reads as healthy. We lost an eight-hour soak to this: one edge moved from `cuda_ipc` to `shm` after the first request, the run reported balanced ACK counts and flat memory, and the path under test carried no traffic. Nothing in the log said so.
- **The direct-IPC stream path records only its sender.** That path has no byte ceiling — when the consumer falls behind, held bytes grow linearly at (produce − consume) × chunk size. Measuring that needs both rates, and `comm_stream_send` carried no byte count.

## Modifications

Three commits, one per concern.

**1. Trace the paged KV transfer path.** Nine events on the engine side, three on the relay side, covering both success and failure:

| event | fires when |
|---|---|
| `comm_kv_send_start` | the sender starts a transfer |
| `comm_kv_ready` | the peer answered the prepare round trip |
| `comm_kv_transfer_complete` | the transfer finished |
| `comm_kv_transfer_failed` | the transfer raised |
| `comm_kv_prepare_ready` | the receiver reserved its destination pages |
| `comm_kv_prepare_rejected` | the receiver refused, with the reason |
| `comm_kv_read_complete` | the receiver finished its copy |
| `comm_kv_read_failed` | the receiver copy raised |
| `comm_kv_pending_retained` | a failed transfer kept its pending entry |
| `cuda_ipc_kv_put` | the source exported its pages |
| `cuda_ipc_kv_get` | the destination copied them |
| `cuda_ipc_kv_peer_access_denied` | the two GPUs cannot reach each other |

`comm_kv_prepare_rejected` sits in `_kv_ready_failure`, which every receiver-side rejection already funnels through: wrong transport, unknown pool, layout mismatch, and a failed reserve all reach it.

`comm_kv_pending_retained` reports `len(_retained_pending_kv_transfers)`. That list is append-only and `close()` does not drain it, so the count is the only signal that it grows.

**2. Record which transport each edge selects.** `comm_transport_selected` fires the first time an edge resolves a transport and again whenever the answer changes. These methods run once per transfer, so emitting every call would bury the log.

**3. Trace direct-IPC stream reads.** Adds `comm_stream_read` on the consumer and a `bytes` field to `comm_stream_send`. Held bytes per edge are then the running difference of the two, straight from the log.

**4. Cover the same-GPU fast paths** (added after review). The direct CUDA-IPC branches in `runtime.py` skip `CommEngine.read_stream_chunk` on receive and the stream send worker on send, so the accounting above missed every same-GPU chunk. Both branches now emit their own `comm_stream_send` / `comm_stream_read` with `transport="torch_cuda_ipc"`. The send branch also skips `router.outbound_stream`, so it records its transport choice through a new `note_transport_choice` entry point, which takes a plain string because `TransportKind` does not carry `torch_cuda_ipc`.

**5. Correlate relay events with the lifecycle** (added after review). `put_kv_pages` carried no id and `get_kv_pages` carried only `request_id`, so concurrent transfers within one request could not be joined to their `comm_kv_*` events. `transfer_id` now threads through both and lands on `cuda_ipc_kv_put`, `cuda_ipc_kv_get`, and `cuda_ipc_kv_peer_access_denied`.

### Cost when the gate is unset

Every event stays behind `SGLANG_OMNI_COMM_TRACE`, which `emit()` checks on its first line. Two details keep the closed path free:

- Exception objects are passed to the trace call rather than `str(exc)`, so the formatting happens inside `emit()`. Trace-call arguments are evaluated before the early return, and a custom `__str__` that raised would otherwise mask the original exception on the failure path.
- The router's dedup map is only touched after `_comm_trace_enabled()` returns true, so a closed run costs one `getenv` call.

## Related Issues

Part of #1593. Relates to #862 Phase 3 and to the paged KV work in #1315 / #1382.

## Accuracy Test

Not applicable. No model-side code changes.

## Benchmark & Profiling

Not applicable. The added calls return on the first line of `emit()` unless `SGLANG_OMNI_COMM_TRACE` is set.

Verified on `b79b1e08`, in a container without GPUs:

| suite | clean `b79b1e08` | this branch |
|---|---|---|
| `tests/unit_test/pipeline/` + `tests/unit_test/relay/` | 3 failed, 453 passed, 12 skipped | 3 failed, **458** passed, 12 skipped |

The same three tests fail on both; they need a CUDA device that this container does not have. The five extra passes are the new tests.

## Checklist

- [x] Add unit tests.

### DIFF
--- a/sglang_omni/comm/engine.py
+++ b/sglang_omni/comm/engine.py
@@ -344,7 +344,7 @@ class CommEngine:
             "comm_stream_read",
             object_id=data_ref.object_id,
             transport=data_ref.transport.value,
-            bytes=data.nbytes,
+            bytes=data_ref.buffer.length,
             elapsed_ms=round(_comm_elapsed_ms(read_start), 6),
         )
         return data, metadata
@@ -446,6 +446,7 @@ class CommEngine:
                 source_pool_id=source_pool_id,
                 source_page_indices=source_page_indices,
                 destination_ref=ready.destination_ref,
+                transfer_id=transfer_id,
             )
             data_ref = DataRef(
                 version=1,
@@ -678,6 +679,7 @@ class CommEngine:
                 source_page_indices=state.request.source_page_indices,
                 destination_page_indices=state.destination.page_indices,
                 request_id=request_id,
+                transfer_id=data_ref.object_id,
             )
             await op.wait_for_completion(timeout=self._ack_timeout_s)
             if state.abort_error is not None:
@@ -690,19 +692,7 @@ class CommEngine:
                 num_pages=len(state.destination.page_indices),
                 elapsed_ms=round(_comm_elapsed_ms(read_start), 6),
             )
-        except asyncio.CancelledError as exc:
-            _comm_trace(
-                "comm_kv_read_failed",
-                transfer_id=data_ref.object_id,
-                request_id=request_id,
-                error=type(exc).__name__,
-                detail=exc,
-                elapsed_ms=round(_comm_elapsed_ms(read_start), 6),
-            )
-            with suppress(Exception):
-                state.receiver.abort(state.request, state.destination, exc)
-            raise
-        except Exception as exc:
+        except (asyncio.CancelledError, Exception) as exc:
             _comm_trace(
                 "comm_kv_read_failed",
                 transfer_id=data_ref.object_id,
--- a/sglang_omni/comm/engine.py
+++ b/sglang_omni/comm/engine.py
@@ -336,9 +336,18 @@ class CommEngine:
         relay: Relay,
         data_ref: DataRef,
     ) -> tuple[torch.Tensor, dict[str, Any] | None]:
-        return await stage_io.read_stream_chunk(
+        read_start = _comm_now_ns()
+        data, metadata = await stage_io.read_stream_chunk(
             relay, data_ref, self.local_payload_device
         )
+        _comm_trace(
+            "comm_stream_read",
+            object_id=data_ref.object_id,
+            transport=data_ref.transport.value,
+            bytes=data.nbytes,
+            elapsed_ms=round(_comm_elapsed_ms(read_start), 6),
+        )
+        return data, metadata
 
     def register_kv_pool(self, pool: KVPool) -> None:
         self._kv_pools[pool.pool_id] = pool
@@ -946,6 +955,7 @@ class CommEngine:
                 to_stage=job.target_stage,
                 chunk_id=job.chunk_id,
                 transport=job.transport.value,
+                bytes=job.data.nbytes,
                 queue_key=queue_key,
                 queue_wait_ms=round((send_start - job.enqueued_ns) / 1_000_000.0, 6),
                 write_ms=round(write_ms, 6),
--- a/sglang_omni/comm/engine.py
+++ b/sglang_omni/comm/engine.py
@@ -363,6 +363,20 @@ class CommEngine:
         transfer_id = transfer_id or (
             f"{request_id}:kv_pages:{from_stage}:{to_stage}:{uuid4().hex}"
         )
+        num_pages = len(source_page_indices)
+        send_start = _comm_now_ns()
+        _comm_trace(
+            "comm_kv_send_start",
+            request_id=request_id,
+            transfer_id=transfer_id,
+            from_stage=from_stage,
+            to_stage=to_stage,
+            source_pool_id=source_pool_id,
+            target_pool_id=target_pool_id,
+            num_pages=num_pages,
+            tp_rank=self.tp_rank,
+            tp_size=self.tp_size,
+        )
         try:
             pool = self._kv_pools.get(source_pool_id)
             if pool is None:
@@ -388,6 +402,7 @@ class CommEngine:
             relay.register_kv_pool(pool)
 
             ready_future = asyncio.get_running_loop().create_future()
+            prepare_start = _comm_now_ns()
             self._kv_ready[transfer_id] = ready_future
             self._outbound_kv_requests[transfer_id] = request_id
             await send_to_endpoint(
@@ -408,6 +423,13 @@ class CommEngine:
             ready = await asyncio.wait_for(
                 ready_future,
                 timeout=self._ack_timeout_s,
+            )
+            _comm_trace(
+                "comm_kv_ready",
+                transfer_id=transfer_id,
+                success=ready.success,
+                error=ready.error,
+                wait_ms=round(_comm_elapsed_ms(prepare_start), 6),
             )
             if not ready.success:
                 raise RuntimeError(ready.error)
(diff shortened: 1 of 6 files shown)


sglang_omni/comm/engine.py
```diff
@@ -336,9 +336,18 @@ async def read_stream_chunk(
...
             relay, data_ref, self.local_payload_device
         )
+        _comm_trace(
+            "comm_stream_read",
+            object_id=data_ref.object_id,
+            transport=data_ref.transport.value,
+            bytes=data.nbytes,
+            elapsed_ms=round(_comm_elapsed_ms(read_start), 6),
+        )
+        return data, metadata
```
JingwenGu0829, line 350: Right now put has no id at all and get only has request_id, so with concurrent transfers for one request the relay events can't be joined to their comm_kv_* lifecycle events. On the trace we can't tell which specific transfer id corresponds to the kv transfer that we care about.

Would it make sense to thread transfer_id into put_kv_pages/get_kv_pages and include it in cuda_ipc_kv_put, cuda_ipc_kv_get, and cuda_ipc_kv_peer_access_denied?

JingwenGu0829, line 350: The same-GPU torch_cuda_ipc fast path bypasses this method on receive (runtime.py:558-588) and bypasses the stream send worker on send (runtime.py:1389-1413). The transfer still works, but COMM_TRACE records neither side’s bytes, so the held-bytes accounting described in the PR misses this path entirely.

Could we emit equivalent comm_stream_send / comm_stream_read events directly from those fast-path branches, including the request/stage/chunk identifiers, transport="torch_cuda_ipc", and bytes=data.nbytes? The send branch should also record the selected transport since it bypasses router._note_transport

sglang_omni/comm/engine.py
```diff
@@ -625,11 +683,34 @@ async def _read_kv_pages(
...
             raise
         except Exception as exc:
+            _comm_trace(
+                "comm_kv_read_failed",
+                transfer_id=data_ref.object_id,
+                request_id=request_id,
+                error=type(exc).__name__,
+                detail=exc,
+                elapsed_ms=round(_comm_elapsed_ms(read_start), 6),
+            )
```
JingwenGu0829: Consider trimming it into a single except (asyncio.CancelledError, Exception) as exc:

sglang_omni/comm/engine.py
```diff
@@ -452,7 +483,24 @@ async def send_kv_pages(
                 raise
             pending_task = self._arm_pending(data_ref.object_id)
             await asyncio.shield(pending_task)
+            _comm_trace(
+                "comm_kv_transfer_complete",
+                transfer_id=transfer_id,
+                num_pages=num_pages,
+                bytes=op.metadata.get("transfer_info", {}).get("size", -1),
```
JingwenGu0829, line 491: nit: bytes=data_ref.buffer.length is enough

sglang_omni/comm/router.py
```diff
@@ -146,10 +149,39 @@ def _intra_node_transport(self, target: str) -> TransportKind:
...
+            previous=None if previous is None else previous.value,
+        )
+        return kind
+
     def outbound(self, target: str) -> TransportKind:
         if target in self.same_process_targets:
-            return TransportKind.LOCAL_OBJECT
-        return self._physical_outbound(target)
+            return self._note_transport("outbound", target, TransportKind.LOCAL_OBJECT)
+        return self._note_transport("outbound", target, self._physical_outbound(target))
```
JingwenGu0829: Nit: _note_transport is repeated at every successful return across the routing methods below. Would it be cleaner to compute kind in the branches and call _note_transport(...) once at the end of each method? That would also make future successful branches traced by default.

jzheng17:
Thanks — all five taken. Pushed as 4bee18b6.

**Correlating the relay events.** You are right, and the description was wrong as well as the code: it claimed held bytes are the running difference of the send and read events, which is false on any path those events miss. `transfer_id` now threads through `put_kv_pages` / `get_kv_pages` and lands on `cuda_ipc_kv_put`, `cuda_ipc_kv_get`, and `cuda_ipc_kv_peer_access_denied`. I checked whether the missing id was more than a trace gap: in `cuda_ipc` the stored `request_id` is only ever read back inside `trace_fields`, so nothing functional aliases on it. That is not true of the nixl relay, where every put waits on the literal `b"done"` — separate note on that below.

**The same-GPU fast paths.** This was the one I could not have found from the diff, thank you. Both branches now emit their own events with `transport="torch_cuda_ipc"` and the byte count. The send branch also skips `router.outbound_stream`, so it records its transport choice through a new `note_transport_choice`, which takes a plain string because `TransportKind` has no `torch_cuda_ipc` member; `_note_transport` now delegates to it and both share the dedup map.

One thing to flag rather than decide alone: those branches already emit `stage_stream_chunk_sent` / `stage_stream_chunk_received` to the request-level recorder, without bytes. I left that alone and emitted separate `comm_trace` events, on the reading that `event_recorder` records the request timeline and `comm_trace` records the data plane. Say the word if you would rather have the byte count on the existing events instead.

**Merging the except branches.** Done. Written as `except (asyncio.CancelledError, Exception)` because `CancelledError` descends from `BaseException`, not `Exception`, so it has to be named explicitly; the tuple is order-independent and the `raise` is unconditional, so cancellation still propagates.

**`data_ref.buffer.length`.** Done.

**Recording the transport once per method.** Done — each routing method computes `kind` in its branches and calls `_note_transport` once at the end, so a branch added later is traced by default.

---

Separately, on the nixl item in #862 Phase 3. The line there reads "concurrent transfers alias", which undersells it. `PutOperation.wait_for_completion` polls `get_new_notifs()`, which drains, and every put waits for the same literal `b"done"` (`relay/nixl.py:261,324`). So the first waiter to poll consumes another transfer's notification and returns as complete, while the transfer that owned it waits out its timeout. The `finally` then runs `on_completion_cb`, which releases the slot — so a wrongly-completed put can hand back a slot the peer may still be reading. That is the failure `cuda_ipc` deliberately avoids on its own timeout path.

Nothing can reach it today: `TransportKind` has no nixl member, so no routing path selects it. I have not touched it — I cannot install nixl or stand up two nodes, and a notification-protocol fix I cannot exercise is the kind that looks right and is not. Flagging it so whoever wires nixl in sees the release-side consequence, not only the aliasing.

PR #1596, opened 2026-08-18
feat(pipeline): integrate generic PD runtime

MelodyyyYin:
## Summary

PR3 connects the PD topology and handoff contract to SGLang's generic TP=1 scheduler/runtime. It constructs the KV receiver, reconstructs Decode requests from committed continuations, installs transferred KV ownership, and admits work through the scheduler thread without prompt re-prefill.

Model-specific state remains behind adapter hooks; this layer imports no Qwen modules.

## Generic resume hooks

PR3 defines the model-neutral seams used by model adapters:

- `pd_state_builder` at the continuation producer;
- `pd_state_restorer` at Decode reconstruction;
- allowed resume-schema capability declarations;
- continuation model-state fields;
- generic output-logprob resume support.

A model adapter supplies projection/restoration and schema identity. The runtime validates the schema and remains independent of Qwen tensor/state details.

## Chunked-Prefill first-token invariant

A request is handed off only after Prefill appends a real first sampled token.

A middle chunk can update chunk accounting while exposing only an internal placeholder. Such a request remains owned by Prefill: it is not serialized, detached, released, or transferred. The producer rejects an empty first-token continuation state. Once the real sample exists, its token and output logprob are preserved exactly once, including through deferred admission.

## KV ownership and handoff

- Adapts SGLang's token-to-KV pool to the transfer interfaces.
- Uses explicit destination `reserve → commit → take_committed` transitions.
- Holds the source page lease until ACK or transfer failure.
- Installs committed slots and prefix bookkeeping into the reconstructed Decode request.
- Transfers admitted requests to the normal SGLang request/KV lifecycle.
- ACK releases the source lease but does not gate Decode admission.
- Clears stale full-batch state when ACK returns Prefill capacity.

## Scheduler admission ownership

Rank-ready allocations enter a FIFO scheduler-thread admission queue.

- Request-pool exhaustion is a specific retryable condition.
- The FIFO head stays pending and retries on later scheduler iterations without sleeping or busy-spinning.
- Deferred admission retains both the reconstructed request and committed destination KV.
- Repeated drains cannot admit a request twice.
- Permanent reconstruction errors reach the coordinator and release ownership exactly once.
- Abort and shutdown release queued/deferred allocations exactly once.
- Decode reuses SGLang's existing `PREBUILT` selection path and restores the invariants consumed by metrics.

## Supported generic scope

- TP=1
- page size=1
- same-node CUDA IPC
- RadixCache disabled
- existing SGLang `PREBUILT` path

TP>1, cross-node transfer, RadixCache ownership, speculative decoding, grammar/custom-logit resume, and model-specific multimodal state remain outside this PR.

## Stack

- **PR1 — #1517:** topology and launch identity
- **PR2 — #1538:** continuation/handoff lifecycle
- **PR3 — #1596 (this PR):** base `pd_handoff`
- **PR4 — #1599:** Qwen adapter/result integration, base `pd_runtime`

## Validation

- Focused runtime/KV/scheduler suite — **89 passed**
- Final stacked `pytest -q tests/unit_test/pipeline tests/unit_test/scheduling` — **769 passed**
- Coverage includes chunked Prefill waiting for a real sample, first-token/logprob continuity, FIFO capacity deferral, permanent-error propagation, abort/shutdown cleanup, repeated-drain idempotence, metrics invariants, and ACK capacity recovery.
- `git diff --check` — passed


jzheng17:
@MelodyyyYin filed #1704 against this PR's code — a correctness defect in the PD path, and a silent one.

PD returns empty completions: HTTP 200, `finish_reason: "stop"`, `completion_tokens: 1`, correct `prompt_tokens`, no content frame. Nothing logs an error. The rate rises with queue depth, reaching 1.6% of requests at the arm's ceiling; a colocated replica under an identical driver produced zero in 3000, and zero in 28,472 at the longest prompt length.

The measurement that localizes it: degraded rows share KV slots with a **currently allocated** request, and control rows never do — 7 of 13 against **0 of 61**, holding at matched prompt length. One captured failure had 1304 of its 1963 slots concurrently mapped by another allocated request while its own mapping was internally clean.

That points at `_queue_pd_prefill_handoffs` (`d9da5b7b`), which ends with `batch.reqs = retained` and removes handed-off requests from the batch while their KV stays live under `SGLangKVPageLease` (`0c6d6d6c`) until the receiver ACK releases it. It is the only place in the PD path where a request leaves the scheduler's batch while still owning slots. **We have not traced the ordering that produces the alias**, so that is where the evidence points rather than a diagnosis — you know this path far better than we do and may see the sequence immediately.

The issue also records what we ruled out, so nobody repeats it: not the server args PD forces (a colocated arm with `page_size=1` and `disable_radix_cache` produced zero under the same driver), not the KV transfer (the token is sampled on the prefill half before any transfer), not sampling numerics (greedy, and colocated never ranks the terminator in its top two, Poisson p = 5.6e-5), not prefill batch composition, and not prompt length.

Two things left open and stated as open: a second route exists — 6 of 13 degraded rows had neither slot sharing nor duplication — and the depth relationship has an unresolved confound between queue depth and server age.

Happy to run any diagnostic that would help, or to try a fix if you have a view on where it belongs. Raw data and probes are available.

MelodyyyYin:
#note:（yue yin）

Committed continuations move through ready/deferred directly into `waiting_queue` without the normal queue cap, while deferred KV is outside waiting-timeout scanning and reconstruction drops the original timing budget. Queue-full or timeout can therefore leave Decode-owned KV alive past the request deadline.

Follow-up PR: https://github.com/sgl-project/sglang-omni/pull/1771

Why: all Decode-owned states need one capacity/deadline/terminal accounting domain so admission, abort, timeout and shutdown transfer or release KV exactly once.

sglang_omni/scheduling/omni_scheduler.py
```diff
@@ -509,6 +540,198 @@ def __init__(
...
+        self._pd_role: str | None = None
+        self._pd_partner: str | None = None
+        self._pd_pool_id: str | None = None
+        self._pd_ready_queue = _queue_mod.Queue()
+        self._pd_deferred_admission: PDDecodeAdmission | None = None
+        self._pd_admission_lock = threading.Lock()
+        self._pd_receiver: AllocatorKVReceiver | None = None
+        self._pd_controller: PDHandoffController | None = None
+
+    def bind_pd_runtime(
```
JingwenGu0829, line 552: Here is doing substantially more than binding communication resources. 

By the time it is called `_construct_scheduler()` has already returned a fully constructed generic 'OmniScheduler' This method then mutates that object into either a prefill or decode scheduler by changing its role, disaggregation_mode, batch_result_processor, admission state, receiver, and controller.

This creates an implicit partially-initialized state, requires PD branches throughout the generic scheduler, and makes Stage discover the scheduler contract through `hasattr()` only after construction.

The explicit mental model should be inheriting the scheduler explicitly:

OmniScheduler
├── OmniPrefillScheduler
└── OmniDecodeScheduler

So the correct lowering should have 
```
thinker_prefill
  scheduler = OmniPrefillScheduler
  config = inherited/resolved prefill config

thinker_decode
  scheduler = OmniDecodeScheduler
  config = inherited/resolved decode config
```

So the substantial generic launch path should only intiate its scheduler by `StageConfig`, without knowing its PD role, and don't need to 

```
hasattr(scheduler, "bind_pd_runtime")
scheduler.bind_pd_runtime(role=...)
```

MelodyyyYin: Thanks, agreed. The current hook changes scheduler semantics after the factory has already returned a generic OmniScheduler. I moved PD role selection into compiler/factory lowering so Prefill and Decode are constructed as explicit scheduler subclasses. Stage now only consumes the typed transport binding and no longer discovers or mutates scheduler behavior through hasattr()/bind_pd_runtime().

Follow-up PR: https://github.com/sgl-project/sglang-omni/pull/1776

PR #1599, opened 2026-08-18
feat(pipeline): complete Qwen PD result path

MelodyyyYin:
## Summary

PR4 is the Qwen3-Omni-specific adapter and observable text-result layer on top of PR3's generic runtime. Its production changes are limited to:

- `sglang_omni/models/qwen3_omni/bootstrap.py`
- `sglang_omni/models/qwen3_omni/request_builders.py`
- `sglang_omni/models/qwen3_omni/stages.py`

Generic resume hooks, continuation ownership, KV transfer, and admission remain in PR3.

## Qwen resume and result state

The adapter provides `QWEN_THINKER_PD_RESUME_SCHEMA` plus Qwen state builder/restorer hooks. It projects only msgpack-safe state needed by Decode:

- original prompt token IDs and request parameters;
- streaming/result metadata;
- first-token and output-logprob continuity;
- Qwen mRoPE position delta when present;
- stage payload needed by the existing Qwen result path.

Consumed Prefill tensors, projected embeddings, and media tensors are not copied because transferred KV represents their effect. The tokenizer is rebound from the Decode-local scheduler rather than serialized.

Validated behavior covers terminal `max_new_tokens=1`, deterministic text, EOS, stop string/token, streaming and non-streaming, token IDs, numeric logprobs, usage, finish reason, weight version, image input, and mRoPE restoration.

## Supported scope

- `Qwen/Qwen3-Omni-30B-A3B-Instruct` at revision `26291f793822fb6be9555850f06dfe95f2d7e695`
- BF16, TP=1, page size=1
- same-node CUDA IPC
- RadixCache and CUDA graph disabled
- thinker text output
- text and image prompts

This PR does not claim general audio/video support. TP>1, cross-node transfer, RadixCache ownership, speculative decoding, grammar/custom-logit resume, and talker/code2wav PD splitting remain outside scope.

## Correctness and lifecycle reliability

With `temperature=0` and `seed=17`, PD and non-PD matched exactly for `max_new_tokens=1/2/3/16/64`, EOS, stop string/token, streaming, and image/mRoPE. Output token IDs, decoded text, usage, finish reason, weight version, and all numeric output logprobs matched.

The rebuilt stack then completed 1,000 persistent requests:

| Phase | Concurrency | Completed | Errors | Time |
| mixed sequential | 1 | 500/500 | 0 | 186.08 s |
| mixed concurrent | 16 | 500/500 | 0 | 46.87 s |

The mix included streaming `max_new_tokens=1/2/3/16` and 25 long 8128-token prompts in each half. Final health was 0 active / 0 pending, with no crash, cleanup assertion, incomplete stream, or monotonic leak.

## Final clean repeated benchmark

Configuration: H200, NV18-connected, BF16, TP=1, page size=1, same-node CUDA IPC, RadixCache/CUDA graph disabled, `max_running_requests=4`. Core cells use 5 warmups + 20 measured requests; mixed cells use 10 warmups + 100 measured requests. Client streaming records TTFT, TPOT, E2E, errors, timeouts, and incomplete streams.

Accepted clean repetitions were PD r1/r3/r6 and non-PD r1/r5/r6. Full-run telemetry covered startup, warmup, measurement, and shutdown. Every accepted measured request completed; errors, timeouts, and incomplete streams were all zero.

Values below are medians across three repetitions; latency is client-observed p50/p95 seconds.

| Workload | Mode | req/s | TTFT | TPOT | E2E |
| 128/128 c8 | PD | 0.5327 | 0.113/0.243 | 0.117/0.118 | 15.060/15.100 |
| 128/128 c8 | non-PD | 0.5302 | 7.607/7.710 | 0.059/0.059 | 15.097/15.189 |
| 8128/32 c8 | PD | 1.9371 | 0.202/1.192 | 0.117/0.120 | 3.881/4.462 |
| 8128/32 c8 | non-PD | 1.7014 | 2.634/2.828 | 0.065/0.068 | 4.648/4.850 |
| 8128/32 c16 | PD | 1.9300 | 1.113/2.537 | 0.147/0.227 | 6.344/8.363 |
| 8128/32 c16 | non-PD | 1.7051 | 5.061/7.465 | 0.065/0.068 | 7.063/9.465 |
| mixed c8 | PD | 0.8302 | 0.180/0.365 | 0.103/0.231 | 7.905/13.107 |
| mixed c8 | non-PD | 0.7946 | 4.390/5.965 | 0.060/0.061 | 8.754/13.623 |
| mixed c16 | PD | 0.8315 | 0.182/1.124 | 0.176/0.530 | 16.669/22.379 |
| mixed c16 | non-PD | 0.7937 | 14.216/15.935 | 0.059/0.061 | 17.639/23.504 |

Against one non-PD thinker, PD improves median throughput by 0.5–13.9% and greatly reduces TTFT queueing, but median TPOT is worse by 74–196% depending on workload.

## Equal-hardware comparison

Two non-PD thinker replicas ran on the same physical thinker GPUs used by PD (GPU0/GPU2), with equivalent shared-stage placement on GPU1 and round-robin routing. Each of three clean repetitions used identical requests and warmups.

| Workload | PD median req/s | 2× non-PD median req/s | 2× Δ | PD total GPU-s | 2× non-PD total GPU-s |
| 8128/32 c8 | 1.9371 | 2.8516 | +47.2% | 93.064 | 63.257 |
| mixed c8 | 0.8302 | 1.0120 | +21.9% | 1084.085 | 890.355 |

GPU-seconds are measurement wall time × three physical GPUs, totaled across the three repetitions for the fixed request counts.

Two non-PD replicas outperform PD in throughput and fixed-work GPU-seconds. PD retains much lower mixed-workload TTFT p95 (0.365 s versus 7.710 s), but this is an isolation/tail-latency tradeoff, not a resource-efficiency win. No PD resource-efficiency claim is made.

## Stack and validation

- **PR1 — #1517:** topology and launch identity
- **PR2 — #1538:** continuation/handoff lifecycle
- **PR3 — #1596:** generic runtime/KV/admission seams
- **PR4 — #1599 (this PR):** base `pd_runtime`; Qwen-only adapter and focused tests

Final validation on `pd_result@c20f4a2f3b3419b2f103191a10fe42cc8feae100`:

- `pytest -q tests/unit_test/pipeline tests/unit_test/scheduling` — **769 passed**
- `git diff --check` — passed
- tracked worktree clean


jzheng17:
We are validating what this stack depends on, and sending changes as PRs against `pd_result` so this thread stays the design record.

**Sent.** #1684 gives the PD compiler a configuration surface and supplies the server args `bind_pd_runtime` requires. #1683 carries the `wait_ready` fix separately. Both sit on `pd_runtime` rather than the stack tip, because neither depends on PR 4. Today `pd_disaggregation` is a `StageConfig` field, `stage_overrides` accepts only `runtime` keys, and no CLI flag sets it, so a deployment cannot turn PD on and the only callers are unit tests. The flag is `--pd-stage thinker=0:1`, modelled on `--stage-process`. `bind_pd_runtime` also refuses a stage unless `disable_radix_cache` is set and `page_size` is 1, and nothing sets them, so the compiler now supplies both.

**Not sent, and it is your call.** Resolving the Decode target by logical name at admission, instead of the physical string `PDExecution.partner` carries, is the seam your deferred multi-P/multi-D item would otherwise need a schema migration to add. With one Decode worker the resolved value never changes, so it costs nothing today. It touches `config/pd_rewrite.py` and `config/schema.py`, which is a different change from a placement flag, so we left it out of #12. Say the word and it comes as its own PR.

**What we are measuring.**

| Area | What we run | Bears on |
| --- | --- | --- |
| Failure paths | abort before rank_ready, continuation timeout, duplicate commit, callback failure | #1538 |
| Rank readiness | TP=2 handoff, checking rank 0 cannot admit before every required rank commits | #1538, #1599 |
| KV ownership | concurrent TP=1 requests, looking for stale `req_to_token` mappings and page leaks | #1596 |
| Colocated versus PD | equal GPU count, matched flags, one short-text arm and one image arm | PR 5 |

The last row may change a plan rather than confirm it. On one H200 the colocated thinker saturates at 17.5 to 18.1 req/s, and a prefill step costs about 59 ms fixed plus 0.010 ms per prompt token. The fixed part does not shrink when PD moves the step to another GPU, so on short prompts a 1P:1D pair may not beat the colocated replicas the same GPUs would otherwise run. The image arm should differ, because `_needs_full_prefill` disables chunked prefill for projected multimodal embeddings and mixed-chunk cannot fold decodes into a step that is not chunked. Numbers and method are on the RFC.

If any row would not change your mind, say so and we will drop it rather than spend the GPUs.

jzheng17:
**Important: TP=2 is not reachable on this branch, so the four-GPU P-TP2/D-TP2 exit criteria cannot be met by it.**

`bind_pd_runtime` raises on the first line of its body, and every PD stage goes through that one entry point (`pipeline/stage/runtime.py:155` is its only caller):

```python
if self.tp_size != 1:
    raise NotImplementedError("PR3 PD runtime supports tp_size == 1 only")
```

Run against the `pd_result` source at `8c973d35`: `tp_size=2` raises, `tp_size=1` clears all three guards. `docs/design/pd_handoff.md:39` on the same branch already states that cross-rank admission for TP>1 is not implemented, so the code and this branch's own documentation agree with each other. The exit criteria is the part that does not.

Lifting this is not a small change. The `rank_ready_callback` is rank-local, so TP>1 needs a cross-rank join before admission: every required rank must commit its share of the KV before the Decode side can admit the request, and an abort or a timeout has to unwind the ranks that already committed. That is its own piece of work, not a flag.

Two suggestions.

**Give TP=2 its own item, and let PR 4's criteria state what it delivers.** Something like: two-GPU P/D end-to-end at `tp_size=1` on Qwen3-Omni thinker, covering the result path, abort, and shutdown. TP=2 then depends on cross-rank admission rather than riding along.

**Pin the boundary in the suite.** No test asserts any of the three guards today, so whoever lifts one gets no signal, in either direction. #1695 adds four cases against `pd_runtime`: one per guard, plus a supported configuration that clears all three, so the set reads as a boundary rather than a wall. Tests only, no behaviour change. Take it or leave it.

jzheng17:
We finished the equal-GPU-count comparison. Data, harness and full method are in #1696. Two results bear on decisions this stack has deferred, so they are here rather than only in that PR.

## 1P:1D is the worst ratio, not a neutral starting point

Prefill is **at most 2.6% of a request's GPU work** on this workload: first forward 58 ms against a decode tail of 2.2 s, measured at text offered 16 where both arms admit every request. It is below that whenever a prefill step carries more than one request, which it does under load. A 1P:1D pair splits hardware 50/50 against that. Measured utilization is **10.1% on the PD prefill card against 67.9% on a colocated card**, which is what such a split predicts.

Prefill steps batch and the batch factor grows with load: at these prompt and output lengths a step carries 1.19 requests at the lowest rate swept and 5.81 at the highest, while the step itself stays between 59 ms and 66 ms across a 14x range of batch size. So the step-cost constant bounds prefill **steps** per second, not requests per second. Prefill duty cycle on one colocated replica peaks near 35% at 128 output tokens and falls back to 24% as the system saturates, so the colocated ceiling at that output length is decode-bound. At 8 output tokens duty reaches 81.6% and prefill does bind.

Segment medians at that point, PD against one colocated replica: queued-to-scheduled 1.99 ms against 3.90 ms, first forward 58.33 ms against 59.99 ms, decode tail 41560 ms against 2206 ms. Admission is not the constraint and prefill did not get slower. The handoff is not serialized either: 41.5% of PD's decode gaps overlap a prefill window.

Prefill share is prompt length divided by output length, so it is not something a longer prompt fixes: at 128 output tokens it reaches only about 9% even at 6944 prompt tokens. Shortening the output moves it much faster.

This is an argument for making the ratio configurable earlier rather than later, and for treating placement as a free parameter so a prefill half can share a GPU instead of owning one. #1690 is the seam for the first; it resolves the Decode target per request and is behaviour-identical with one Decode worker.

## Decode-side admission needs a bound

By Little's law PD holds about **437 requests in the decode tail against `max_running_requests=64`**, so most of that tail is queueing rather than decoding. Colocated holds about 18. Colocated throttles admission implicitly, because prefill and decode contend for one card and one scheduler thread; splitting them removes that throttle and nothing replaces it.

The visible consequence is that in-flight count climbs to four figures while colocated stays in the low hundreds:

| Arm | Rate per server | total_s p50 | In flight |
| --- | --- | --- | --- |
| C, offered 44 | 19.66 | 7.95 s | 156 |
| PD, offered 16 | 10.59 | 40.96 s | 434 |
| PD, offered 24 | 10.55 | 68.81 s | 726 |
| PD, offered 44 | 10.54 | 74.12 s | 781 |

In flight is achieved rate times median total time.

**Correction to an earlier version of this comment.** It reported connection failures above offered 16 as the server refusing under load. They were client-side descriptor exhaustion: the client holds one socket per in-flight request and ran with an unlimited connector under a 1024 soft limit. Every such record carries `ClientConnectorError ... [Too many open files]`, which is EMFILE from the client's own `socket()` call, and `[Connect call failed]` — what a refusing server produces — appears nowhere in the results. The colocated arm ran two client processes with a descriptor budget each, so the arms were not comparable on that axis. The harness has been fixed in #1696.

The throughput comparison and the 41-second `total_s` are unaffected: the former counts completions over wall time, and the latter was measured at offered 16 where there were no failures at all.

Worth noting for anyone reading admission rates: at offered 16 PD admits **every** request while each takes 41 s against colocated's 2.3 s. Admission rate reports that point as healthy. In-flight count and total time report the real state.

## What PD does deliver

PD removes prefill/decode interference almost completely. Ratio of the median inter-token gap overlapping a prefill to one that does not:

| Workload | Colocated | PD |
| --- | --- | --- |
| text, 8 to 44 rps | 5.8x to 8.5x | 1.00x to 1.89x |
| image, 1 to 6 rps | 8.4x to 20.5x | 0.96x to 1.02x |

The A/A band on that ratio is 2.04% median, so this is unambiguous. On the image arm a prefill in flight on the other card is statistically indistinguishable from no prefill at all.

The ratio says prefill stopped being a source of jitter. It does not say the stream is smooth. PD's gaps with no prefill in flight grow from 7.80 ms to 31.15 ms as load rises, while colocated's stay between 7.6 ms and 12.9 ms, so PD trades prefill-correlated spikes for steady queueing on the decode card. Which trade is better depends on the SLO: a one-shot latency target prefers the lower spikes, while a repeating real-time deadline depends on whether a playback buffer absorbs the steady part.

That last case is measurable here without new tooling. `benchmarks/metrics/playback_continuity.py` already simulates the playout buffer and `c50`/`c100`/`c200` already report the fraction of streams whose worst underrun stays inside 50, 100 and 200 ms. Those numbers, on a PD deployment, would say whether the interference removal converts into audible continuity. Nobody has run them against PD, and PD currently binds only the thinker stage, so that is a gap rather than a result.

jzheng17:
I think our verdict on PD's performance in this PR is wrong, and the cause is a setting rather than the design. Same workload as your 8128/32 sweep, same `max_running_requests=4`, one H200 per half, closed-loop streaming client, 100 requests per cell.

## The verdict is measuring CUDA graph, not the split

The PR concludes that PD's median TPOT is worse by 74-196% and makes no resource-efficiency claim. Both follow from the baseline running colocated with the graph on and PD with it off. Turning it on for PD, with nothing else changed:

| Two-card PD | req/s | TPOT p50 | E2E p50 | E2E p95 |
| c8, graph off | 1.79 | 0.1206 | 4.06 | 7.05 |
| c8, **graph on** | **5.32** | **0.0083** | **1.37** | **2.69** |
| c16, graph off | 1.95 | 0.2514 | 7.98 | 8.15 |
| c16, **graph on** | **5.58** | **0.0083** | **2.74** | **2.77** |

TPOT falls about 30x and throughput roughly triples. Against a colocated replica measured the same way, PD then leads on throughput as well: 5.58 against 4.55 req/s at c16, with TPOT 0.0083 against 0.0165.

So the honest statement is closer to: **at equal per-half hardware and with both arms configured the same way, PD is faster on this workload, not slower.** The TPOT regression was ours to remove, not a property of disaggregation.

Nothing in PD needs the graph off. With DP attention off, `batch.inner_idle_batch` is None, so `_run_batch_prebuilt` returns an empty `GenerationBatchResult` and runs no forward; admission has no shape for the graph to miss, and every step after it is `ForwardMode.DECODE`. #1684 carries a test pinning that CUDA graph is not among `bind_pd_runtime`'s guards.

## The equal-hardware conclusion still stands, for a different reason

Your 2x non-PD result is not an artifact of the graph setting, and it holds: parity is the arithmetic ceiling for a 1:1 pair. A card doing both jobs has the harmonic mean of its prefill-only and decode-only capacities; a pair has the minimum of the two; the minimum never exceeds the harmonic mean. So 1:1 across two cards cannot beat two colocated replicas even in principle, which is why choosing the ratio is on the roadmap (#1760) rather than tuning 1:1.

Those two things are consistent: PD wins per half and loses per pair. Worth separating in the report, because a reader takes away the second and concludes the design does not work.

## Both halves on one GPU: a measurement floor first

I posted a same-GPU against colocated throughput comparison here earlier. It is not readable, and neither is any other single-card throughput ratio I have measured, including ones in #1696.

Identical configuration, fresh server each time, three repeats, 8128/32 at c8 on one H200:

| arm | req/s across three runs | spread |
| --- | --- | ---: |
| colocated | 2.084, 2.065, 3.739 | 81% |
| same-GPU PD | 1.855, 2.241, 2.787 | 50% |

Pairing does not rescue it: PD relative to colocated within each repeat is 0.89x, 1.09x, 0.75x. The driver is the rest of the host — the colocated arm read 3.559 with GPUs 0 and 1 idle and 2.084 with them at 100% and 88%. Checking that one's own card is free is not enough on this machine.

Worth knowing before either of us reports a single-card number, so flagging it rather than only fixing my own.

What survives that floor on the same-GPU shape, because it is a per-request latency ratio measured back to back rather than a throughput ratio:

| prompt tokens | first token p50 | between tokens p50 |
| ---: | --- | --- |
| 256 | 5.4x sooner | 2.4x longer |
| 1024 | 4.4x sooner | 2.5x longer |
| 8128 | no gain | 1.4x shorter |

Sharing a card moves latency out of the first token and into the stream, and the effect fades as the prompt grows. That is a workload question rather than a verdict on the shape.

One thing that is a defect regardless of the floor. With `share_weights` on, one declared `memory_fraction` gives the halves very different KV budgets: the publisher holds its weight copy inside its share and the adopter releases its copy before its pool is sized, so 0.47/0.47 on one H200 yields 22,547 tokens and 643,731. It does not explain any throughput number — sweeping the split 27x moves throughput 1.5% — but an operator should not have to know the weight size to write a sensible pair. That is #1717's to fix.

## Sizing is where the larger win is

One prefill and two decode halves, one H200 each, on the same harness:

| cell | 1:1 | 1:2 | change | 1:1 TPOT p50 | 1:2 TPOT p50 |
| 1024 / 32, c8 | 13.925 | 20.375 | +46.3% | 0.0120 | 0.0075 |
| 1024 / 32, c16 | 15.328 | 25.431 | +65.9% | 0.0259 | 0.0089 |
| 8128 / 32, c8 | 4.835 | 4.819 | −0.3% | 0.0072 | 0.0070 |
| 8128 / 32, c16 | 4.764 | 4.693 | −1.5% | 0.0072 | 0.0072 |

At 8128 tokens the single prefill card runs at 100% for the whole cell, so a second decode half adds nothing; at 1024 it is worth two thirds. This is the argument for spending the next effort on the ratio rather than on 1:1, and it is #1690 plus `num_replicas: 2` on the decode process, nothing else.

Happy to hand over the harness and the raw cells so you can reproduce any of this before changing the write-up.

PR #1600, opened 2026-08-18
[Comm] Collect the comm-layer tests and split invariants from characterizations

jzheng17:
Stacked on #1595. Review that one first; this branch contains its three commits.

## Motivation

The comm-layer tests are spread across `tests/unit_test/pipeline/` and `tests/unit_test/relay/`, mixed in with tests for everything else. CI cannot select them by path, so a change to the comm layer either runs the whole suite or runs nothing targeted.

There is a second problem that matters more. Some of these tests pin properties that must always hold. Others pin behaviour that we expect to change, and that #862 Phase 3 plans to change on purpose. Today nothing in the code says which is which. Whoever writes Phase 3 will face a screen of red and no way to tell the lines they are meant to change from the ones they broke.

There is a third gap, which the second commit closes. Every test in `tests/unit_test/comm/` drives `FakeRelay` and `FakeOp`. None of them allocates device memory, exports a CUDA IPC handle, or copies a byte on a GPU. The engine-side accounting is well covered; the relay's actual transfer path is not covered at all.

## Modifications

Add `tests/unit_test/comm/`, and mark every test in it with what a failure means. The tests already in `tests/unit_test/pipeline/` and `tests/unit_test/relay/` stay where they are: this branch removes nothing. They assert things the new directory does not, such as the data-reference schema and the stream reuse semantics. The new directory gives CI a path it can select, and the markers say which failures are regressions.

| marker | meaning |
|---|---|
| `comm_invariant` | a property that must always hold. A change here is a correctness regression, not a design decision. |
| `comm_characterization` | behaviour as it stands today, expected to change. A pull request that flips one should say why. |

Two tests are characterizations. Both are things that become defects once failure handling stops tearing down the whole relay:

- `test_timeout_marks_ops_failed_without_slot_release` — an ACK timeout does not release the slots. Correct today, because a timeout does not prove the peer stopped reading.
- `test_retained_transfer_is_never_reclaimed` — `close()` does not drain `_retained_pending_kv_transfers`.

The other 22 are invariants: ACK-to-transfer correspondence, slot accounting, NACK isolation, idempotent duplicate ACKs, allocator conservation, and the wakeup rule.

`test_trace_events_carry_their_documented_fields` replaces an earlier test that asserted an exact event sequence. Offline tools read the trace by name and by field, so the contract is that names and fields survive; adding an event is safe. The test asserts required fields per event and an ordering subsequence, so it keeps passing as events are added.

`AckedOp`, `AckedRelay`, and `wait_until` are new, and they live in `tests/unit_test/fixtures/`. `benchmarks/comm/allocator_workloads.py` provides the virtual-tick harness the allocator tests replay: it drives `_ContiguousSlotAllocator` on integer ticks instead of wall clock, so a fragmentation workload that would take minutes runs in milliseconds and gives the same answer every time.

### Hardware tests

The second commit adds `tests/unit_test/comm/test_cuda_ipc_relay_hw.py`: 14 tests marked `accelerator`, all of them running against a real device.

| what it covers | tests |
|---|---|
| pool allocation, put and get byte correctness, slot reuse after ACK, concurrent transfers | 6 |
| CUDA IPC handle import in a second process, full put and get across two processes | 2 |
| peer access reporting, cross-GPU transfer | 3 |
| `transfer_kv_per_layer_mla` page scatter and gather, the full paged KV path through `CudaIpcRelay` | 2 |
| trace events emitted by a real transfer | 1 |

One CUDA rule shapes the file. A process cannot import an IPC handle it exported itself: `rebuild_cuda_tensor` and `Event.from_ipc_handle` both raise `cudaErrorDeviceUninitialized`. So the tests split by what they check. Tests that check the GPU copy use an `ipc_bypass` fixture, which patches `Event.from_ipc_handle` to synchronize and return a fresh event, and pre-caches the pool tensor in `_remote_pools`. Tests that check the handle roundtrip itself spawn a second process through `multiprocessing.spawn` and import the handle there for real.

The two-GPU tests skip when only one device is visible.

## What this does not do

No gate on any number yet. Threshold constants, the GPU ratio checks, and the CI wiring come in a follow-up, once a calibration run gives them a history. Reporting a number nobody calibrated is how a gate gets disabled two weeks later.

## Related Issues

Part of #1593. Prepares the ground for #862 Phase 3.

## Accuracy Test

Not applicable. Tests only; no production code in this branch.

## Benchmark & Profiling

Not applicable.

Rebased onto `fee6a1bc`. I ran both lanes that #1664 defines, on an H200 host.

The CPU lane, `pytest tests/unit_test/ -m "not benchmark and not accelerator"`, with `CUDA_VISIBLE_DEVICES` empty, matching the workflow:

```
1 failed, 5079 passed, 3 skipped, 179 deselected in 390.52s
```

The one failure is `tests/unit_test/pipeline/test_ipc.py::test_mp_runner_startup_failure_includes_child_factory_traceback`. It fails the same way on `fee6a1bc` with no commits from this branch applied, so this branch does not cause it. The test starts a real multi-process runner and waits 30 s for a child to report a factory error; with CUDA hidden the child never signals ready and `wait_ready` raises `TimeoutError` instead of the expected `RuntimeError`. `run_flaky_pytest.sh` retries, which is likely why CI has not surfaced it. Flagging it rather than leaving it in the log.

`tests/unit_test/dots_tts` is excluded above because `test_tail.py` imports `dots_tts`, which my container does not have. That file is on `main` and this branch does not touch it.

The accelerator lane, `pytest tests/unit_test/comm/ -m "accelerator and not benchmark"`, on two idle GPUs:

```
14 passed, 24 deselected in 50.59s
```

Collecting all of `tests/` in the CPU lane deselects 223 tests, and all 14 hardware tests are among them, so the marker does keep them off the CPU lane.

#1664 marked the three tests that need a device — `test_cuda_ipc_same_gpu_round_trip`, `test_cuda_ipc_paged_multi_buffer_round_trip_with_storage_offsets`, and `test_cuda_ipc_cross_gpu_round_trip` — as `accelerator`, so the CPU lane does not select them. This file follows that convention.

Marker selection works:

```
pytest tests/unit_test/comm/ -m comm_invariant          22 selected, 16 deselected
pytest tests/unit_test/comm/ -m comm_characterization     2 selected, 36 deselected
```

## Checklist

- [x] Add unit tests.

### DIFF
--- a/benchmarks/comm/allocator_workloads.py
+++ b/benchmarks/comm/allocator_workloads.py
@@ -1,0 +1,899 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Fragmentation workload driver for the CUDA-IPC contiguous slot allocator.
+
+Replays synthetic (and, later, GPU-captured) allocation schedules against
+``_ContiguousSlotAllocator`` on a virtual tick clock -- no GPU, no wall-clock
+sleeps -- and reports occupancy, fragmentation, and contiguity-induced wait
+metrics. Alternative allocators (buddy, size-class) replay the same schedule
+for offline comparison.
+
+Wait classification:
+- capacity wait: the pool simply had fewer free slots than requested.
+- contiguity wait: enough free slots existed, but no contiguous run fit the
+  request -- this is the fragmentation failure mode debated in RFC #287.
+
+Usage:
+    python -m benchmarks.comm.allocator_workloads --schedule bimodal \
+        --pool-slots 256 --slot-kb 64 --seeds 0 1 2 --format json --out out.json
+    python -m benchmarks.comm.allocator_workloads --sweep-large-ratio \
+        --pool-slots 256 --slot-kb 64 --seeds 0 1 2 --format md
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import heapq
+import json
+import random
+import sys
+from dataclasses import dataclass, field
+from typing import Any, Protocol
+
+from sglang_omni.relay.cuda_ipc import (
+    _ContiguousSlotAllocator,
+    _SlotAllocation,
+    _slots_for_size,
+)
+
+# Rounds of cooperative yielding per tick so blocked acquires can observe
+# releases made earlier in the same tick.
+SETTLE_ROUNDS = 8
+
+
+class SlotAllocator(Protocol):
+    slot_count: int
+    slot_size: int
+
+    async def acquire_async(
+        self, num_slots: int, *, capture_layout: bool = False
+    ) -> _SlotAllocation: ...
+
+    def release(self, offset: int, num_slots: int) -> None: ...
+
+    def layout_snapshot(self) -> dict[str, int]: ...
+
+
+# --------------------------------------------------------------------------
+# Schedules
+# --------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
+class Request:
+    request_id: str
+    submit_tick: int
+    size_bytes: int
+    hold_ticks: int
+
+
+@dataclass(frozen=True)
+class Schedule:
+    name: str
+    params: dict[str, Any]
+    requests: list[Request]
+
+
+def fixed_schedule(
+    *,
+    count: int = 200,
+    size_bytes: int = 3 * 2**20,
+    arrival_period: int = 1,
+    hold_ticks: int = 8,
+) -> Schedule:
+    requests = [
+        Request(f"fixed-{i}", i * arrival_period, size_bytes, hold_ticks)
+        for i in range(count)
+    ]
+    return Schedule(
+        "fixed",
+        {
+            "count": count,
+            "size_bytes": size_bytes,
+            "arrival_period": arrival_period,
+            "hold_ticks": hold_ticks,
+        },
+        requests,
+    )
+
+
+def bimodal_schedule(
+    *,
+    count: int = 400,
+    small_bytes: int = 128 * 2**10,
+    large_bytes: int = 8 * 2**20,
+    large_ratio: float = 0.1,
+    jitter_pct: float = 0.2,
+    arrival_period: int = 1,
+    small_hold: int = 4,
+    large_hold: int = 16,
+    seed: int = 0,
+) -> Schedule:
+    rng = random.Random(seed)
+    requests = []
+    for i in range(count):
+        is_large = rng.random() < large_ratio
+        base = large_bytes if is_large else small_bytes
+        jitter = 1.0 + rng.uniform(-jitter_pct, jitter_pct)
+        size = max(1, int(base * jitter))
+        hold = large_hold if is_large else small_hold
+        requests.append(Request(f"bimodal-{i}", i * arrival_period, size, hold))
+    return Schedule(
+        "bimodal",
+        {
+            "count": count,
+            "small_bytes": small_bytes,
+            "large_bytes": large_bytes,
+            "large_ratio": large_ratio,
+            "jitter_pct": jitter_pct,
+            "arrival_period": arrival_period,
+            "small_hold": small_hold,
+            "large_hold": large_hold,
+            "seed": seed,
+        },
+        requests,
+    )
+
+
+def adversarial_schedule(
+    *,
+    slot_count: int,
+    slot_size: int,
+    large_run: int = 8,
+    large_count: int = 12,
+    comb_hold_short: int = 2,
+    comb_hold_long: int = 200,
+) -> Schedule:
+    """Deterministically provoke a comb: fill the pool with 1-slot holds where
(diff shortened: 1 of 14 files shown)


jzheng17:
@JingwenGu0829 I ran the unit tests locally the way you suggested, on an H200 host. Here is what I found, what changed, and one thing I need from you.

**The CI refactor is #1664**, which moved the accelerator tests into their own pytest job. This branch had already been rebased past it. I diffed `.github/workflows/test.yaml` and the `pyproject.toml` markers between this branch's base and current `main` and found no change between them, so there was nothing to fix on that front. I rebased onto `fee6a1bc` anyway to stay current.

**On marking the GPU tests:** #1664 already marked the three that need a device — `test_cuda_ipc_same_gpu_round_trip`, `test_cuda_ipc_paged_multi_buffer_round_trip_with_storage_offsets`, and `test_cuda_ipc_cross_gpu_round_trip`. This branch's new hardware tests follow that convention.

**What is new here.** Every existing test in `tests/unit_test/comm/` drives `FakeRelay` and `FakeOp`, so none of them touches a GPU. The engine-side accounting was covered; the relay's transfer path was not covered at all. I added `tests/unit_test/comm/test_cuda_ipc_relay_hw.py` — 14 tests marked `accelerator` that allocate real device memory, export real CUDA IPC handles, and copy real bytes. They cover pool allocation, put and get byte correctness, slot reuse after ACK, concurrent transfers, handle import in a second process, cross-GPU transfer, `transfer_kv_per_layer_mla` page scatter and gather, the full paged KV path, and trace events from a real transfer. The two-GPU tests skip when only one device is visible.

One detail if you review that file: a process cannot import a CUDA IPC handle it exported itself, because `rebuild_cuda_tensor` and `Event.from_ipc_handle` both raise `cudaErrorDeviceUninitialized`. So the tests split by what they check. Tests that check the GPU copy use an `ipc_bypass` fixture. Tests that check the handle roundtrip spawn a second process and import it for real.

On cost, in case it comes up: these 14 tests are already opt-in twice. `test.yaml` runs only through `workflow_call` from `omni-ci.yaml`, which stops at `omni-ci-gate` without the `run-ci` label, and inside it the `accelerator` marker keeps them in their own job away from the CPU lane. They take 50 s on two GPUs, and the three that need a second GPU skip when only one is visible.

**Results.** Both lanes, run exactly as the workflow defines them:

| lane | result |
|---|---|
| CPU, `CUDA_VISIBLE_DEVICES` empty | 1 failed, 5079 passed, 3 skipped, 179 deselected |
| accelerator, two idle GPUs | 14 passed, 24 deselected |

**About that one failure.** It is `tests/unit_test/pipeline/test_ipc.py::test_mp_runner_startup_failure_includes_child_factory_traceback`, and it is not from this branch: it fails the same way on `fee6a1bc` with none of my commits applied. The test starts a real multi-process runner and waits 30 s for a child to report a factory error. With CUDA hidden the child never signals ready, so `wait_ready` raises `TimeoutError` instead of the expected `RuntimeError`. `run_flaky_pytest.sh` retries, which is probably why CI has not surfaced it. Happy to open a separate issue if you want it tracked.

I excluded `tests/unit_test/dots_tts` from that run because `test_tail.py` imports `dots_tts`, which my container does not have. That file is on `main` and this branch does not touch it.

**What I need from you.** `omni-ci-gate` fails with "Missing required label 'run-ci'", so CI has not run the unit tests on this PR at all. Could you add the label, or run `/tag-and-rerun-ci`, and review when you have time?

benchmarks/comm/allocator_workloads.py
JingwenGu0829, line 1: should benchmark exist here?  The PR says 'Benchmark & Profiling: Not applicable'. I think if it stays as a refractor of unit test, then the benchmark +1000 should not be existing?

tests/unit_test/comm/__init__.py
JingwenGu0829, line 1: we added ~+1000 unit test here but did not replace any redundanct existing test. I think there's a lot of duplicated existing test right now.

tests/unit_test/comm/test_cuda_ipc_relay_hw.py
```diff
@@ -0,0 +1,529 @@
...
+@_requires_cuda
+def test_ipc_handle_roundtrip_cross_process() -> None:
+    src = torch.arange(64, dtype=torch.uint8, device="cuda:0")
+    handle = _dump_cuda_storage_handle(src)
+    ctx = mp.get_context("spawn")
+    q = ctx.Queue()
+    p = ctx.Process(target=_child_import_ipc, args=(handle, q))
+    p.start()
+    p.join(timeout=30)
+    assert not p.is_alive(), "child process hung"
```
JingwenGu0829, line 262: this could just hang the CI

PR #1602, opened 2026-08-18
Track MUSA runtime backends in GPU diagnostics

froststeam:
## Summary

Add MUSA-related backend visibility to `sglang_omni.diagnostics.gpu`:

- report `torchada` as a runtime backend
- report `mate` as an attention backend
- add a regression test so these MUSA backend probes stay listed

This does not change model execution or backend selection. It is a low-risk diagnostics step toward the MUSA support roadmap, and helps users verify whether the runtime pieces required for MUSA and MATE attention are installed/importable.

## Validation

- `python -m py_compile sglang_omni/diagnostics/gpu.py tests/unit_test/diagnostics/test_gpu.py`
- `python -m pytest -q tests/unit_test/diagnostics/test_gpu.py`

### DIFF
--- a/sglang_omni/diagnostics/gpu.py
+++ b/sglang_omni/diagnostics/gpu.py
@@ -20,8 +20,10 @@ from sglang_omni.utils.gpu_memory import (
 )
 
 _BACKENDS = (
+    ("runtime", "torchada", "torchada"),
     ("attention", "flash-attn-4", "flash_attn.cute"),
     ("attention", "flashinfer", "flashinfer"),
+    ("attention", "mate", "mate"),
     ("attention", "triton", "triton"),
     ("attention", "torch-sdpa", "torch.nn.functional"),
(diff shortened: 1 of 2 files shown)


sglang_omni/diagnostics/gpu.py
```diff
@@ -20,8 +20,10 @@
 )
 
 _BACKENDS = (
+    ("runtime", "torchada", "torchada"),
```
yeahdongcn, line 23: It would be a great opportunity to introduce vendor-specific branching. For us, `MATE` and `torchada` are two components we're particularly interested in.

sglang_omni/diagnostics/gpu.py
yeahdongcn, line 337: We can define a global visible device key and use `pymtml` as `pynvml` here.

froststeam:

PR #1603, opened 2026-08-18
Use MATE varlen attention for MOSS TTS on MUSA

froststeam:
## Summary

Add a MUSA-friendly packed attention fallback for the MOSS TTS audio tokenizer.

When SGLang's flash-attention varlen kernel is unavailable, this falls back to `mate.flash_attn_varlen_func`. The existing CUDA path keeps using the SGLang flash-attention implementation when available, while the MATE fallback allows packed flash attention to be enabled on MUSA devices.

This PR intentionally only handles the varlen attention API because the current MATE package exposes `flash_attn_varlen_func`; it does not try to emulate ordinary `flash_attn_func`.

## Validation

- Inspected the MUSA runtime image and confirmed `mate 0.2.4.dev20260710` exposes `flash_attn_varlen_func`
- `python -m py_compile sglang_omni/models/moss_tts/audio_tokenizer.py tests/unit_test/moss_tts/test_audio_tokenizer.py`
- `python -m pytest -q tests/unit_test/moss_tts/test_audio_tokenizer.py`

### DIFF
--- a/sglang_omni/models/moss_tts/audio_tokenizer.py
+++ b/sglang_omni/models/moss_tts/audio_tokenizer.py
@@ -10,6 +10,7 @@ from collections.abc import Iterator, Sequence
 from contextlib import nullcontext
 from dataclasses import dataclass
 from itertools import accumulate
+from types import SimpleNamespace
 from typing import Any
 
 import torch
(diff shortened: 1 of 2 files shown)


froststeam:

PR #1611, opened 2026-08-18
[Qwen3-Omni Perf] Talker: Avoid repeated pending-text queue concatenation

cuzmi:
## Motivation

Qwen3-Omni can start the Talker before the Thinker has finished producing all text embeddings. While the Talker consumes the pending-text queue, newly produced Thinker rows are appended to that queue.

The previous implementation kept the queue in one tensor. When unconsumed rows were already present, every append performed an out-of-place `torch.cat`:

```python
remaining = self.rows[self.cursor :]
self.rows = torch.cat([remaining, new_rows], dim=0)
```

This reallocates the queue tensor and recopies all unconsumed rows on every append. If the backlog grows across `N` appends, the cumulative amount of old data recopied can grow quadratically in the worst case.

This PR removes that repeated concatenation without moving pending embeddings off device or changing the Talker's FIFO consumption order.

## Modifications

- Replace the contiguous pending-text tensor with a queue of device-resident tensor chunks.
- Append incoming Thinker rows without `torch.cat`, then consume chunks in FIFO order with a cursor while keeping `len(queue)`, `queue[0]`, and `popleft()` O(1).
- Add focused unit coverage for FIFO behavior across chunk boundaries, partial consumption followed by append, queue-copy independence, runtime queue reuse, dtype normalization, and invalid input shapes.

For tensors already on the queue's device and dtype, an append now only adds a chunk to the queue. A device or dtype conversion may still copy the newly appended rows, but existing pending rows are no longer recopied.

## Related Issues

- Related to #1018 (Qwen3-Omni optimization track 2.6).
- The partial Talker startup behavior was introduced by #475 and later enabled by default in #617.
- This keeps pending embeddings device-resident, consistent with #494's goal of avoiding D2H/H2D data transfers.

### Streaming-output comparison

The baseline and optimized implementations were compared using the same 32 samples in the same order, with a fresh server started for each implementation.

Configuration:

- Dataset: `zhaochenyang20/seed-tts-eval-arrow`
- Samples: the same 32 samples in the same order for both sources
- Concurrency: 1
- Streaming: enabled
- max_new_tokens: 256

Comparison result:

| Check | Result |
| --- | ---: |
| Baseline requests | 32 |
| Optimized requests | 32 |
| Exactly matched requests | 32 |
| Mismatched requests | 0 |

Each request was compared using request status, token counts, WAV format, and the SHA-256 hash of the exact PCM output. All 32 requests matched.

### Setup

The baseline and optimized implementations were compared.

- Hardware: H100
- Server configuration: Qwen3-Omni FP8 colocated
- Workload: streaming SeedTTS (`zhaochenyang20/seed-tts-eval-arrow`)
- Output limit: 256 new tokens
- Cases:
  - C1: 50 requests per repetition, concurrency 1
  - C16: 160 requests per repetition, concurrency 16
- Five repetitions per arm
- Four counterbalanced rounds:
  - baseline -> optimized
  - optimized -> baseline
  - baseline -> optimized
  - optimized -> baseline

Across all repetitions, both implementations completed 4,200 measured requests with zero failures and appended the same 53,660 rows.

### End-to-end results

Percentage changes are "optimized" relative to "baseline". Positive throughput and negative latency are better.

| Case | Metric | baseline | optimized | Change |
| C1 | QPS | 1.4715 | 1.4825 | +0.75% |
| C1 | Latency | 0.6795 | 0.6745 | -0.74% |
| C1 | P95 Latency | 0.9940 | 0.9840 | -1.01% |
| C1 | TTFP | 0.3104 | 0.3072 | -1.03% |
| C1 | TTFT | 0.0938 | 0.0924 | -1.49% |
| C1 | ITL | 0.0922 | 0.0904 | -1.95% |
| C1 | P95 ITL | 0.1114 | 0.1090 | -2.20% |
| C16 | QPS | 7.3510 | 7.4280 | +1.05% |
| C16 | Latency | 2.1150 | 2.1050 | -0.47% |
| C16 | P95 Latency | 3.2470 | 3.2405 | -0.20% |
| C16 | TTFP | 1.0262 | 1.0016 | -2.39% |
| C16 | TTFT | 0.1766 | 0.1745 | -1.22% |
| C16 | ITL | 0.2974 | 0.2995 | +0.71% |
| C16 | P95 ITL | 0.7069 | 0.6959 | -1.56% |

The end-to-end result is performance-neutral for this workload: the targeted repeated copy is removed without an observed latency or throughput regression.

### Queue-level results

To verify the intended mechanism independently of end-to-end timing, the append path of `PendingTextTensorQueue` was instrumented again.

For the baseline implementation, every append that encountered unconsumed rows recorded:

- one queue-level `torch.cat` call;
- the logical size of the existing rows passed into that concatenation; and
- the logical size of the concatenated output tensor.

The byte counts were calculated from tensor shapes, dtypes, and element sizes. They represent logical tensor data copied or allocated by the queue operation, not CUDA allocator usage, memory-bandwidth measurements, or GPU kernel time.

The optimized implementation processed the same logical append events, but stored each new tensor as a queue chunk instead of concatenating it with the existing backlog.

The per-case values below are the median of 20 measured repetitions for each implementation. Each C1 repetition contained 50 requests, and each C16 repetition contained 160 requests.

| Case | Counter | baseline | optimized |
| --- | --- | ---: | ---: |
| C1 | Queue-level `torch.cat` calls | 481 | 0 |
| C1 | Existing-row bytes recopied | 7.98 MiB | 0 |
| C1 | Concatenated-output bytes | 8.92 MiB | 0 |
| C16 | Queue-level `torch.cat` calls | 1,082 | 0 |
| C16 | Existing-row bytes recopied | 14.99 MiB | 0 |
| C16 | Concatenated-output bytes | 17.01 MiB | 0 |

Across the complete four-round experiment, both implementations appended the same number of logical text rows:

| Counter | baseline | optimized |
| --- | ---: | ---: |
| Appended text rows | 53,660 | 53,660 |
| Queue-level `torch.cat` calls | 31,120 | 0 |
| Existing-row bytes recopied | 457.4 MiB | 0 |
| Concatenated-output bytes | 518.1 MiB | 0 |

Matching appended-row counts confirm that both implementations processed the same logical queue workload. These counters show that the targeted queue-level concatenation was eliminated.

### Talker AR trace results

A separate Torch Profiler comparison measured CatArray kernel activity across the full Talker AR trace using the same workload for the baseline and optimized implementations.

Setup:

- Hardware: H100
- Server configuration: Qwen3-Omni FP8 colocated with partial Talker startup
- Dataset: the first 32 rows of the English SeedTTS dataset
- Concurrency: 16
- Warmup: 16 requests outside the profiler window
- Measured requests: 32 per source
- max_new_tokens: 512
- Thinker/Talker temperature: 0
- Sampling seed: 777

Both sources completed all 32 measured requests and produced the same aggregate work: 6,432 prompt tokens and 5,216 output tokens.

Within the Talker AR trace, all CUDA kernel events whose names contain `CatArray` were counted. These counts cover the full Talker AR profiler window and are not restricted to the pending-queue append path.

| Trace metric | baseline | optimized | Change |
| --- | ---: | ---: | ---: |
| CUDA CatArray kernel launches | 6,925 | 1,970 | -71.55% |
| CUDA CatArray cumulative GPU time | 10.341 ms | 2.836 ms | -72.57% |

The average CatArray kernel duration remained nearly unchanged (1.49 us on `baseline` and 1.44 us on `optimized`), so the observed reduction in the Talker AR trace comes from launching fewer small cat kernels rather than making one kernel faster.

## Checklist

- [x] Add unit tests.

## CI


sglang_omni/models/qwen3_omni/pending_text_queue.py
```diff
@@ -49,34 +63,45 @@ def __bool__(self) -> bool:
...
-            raise TypeError("PendingTextTensorQueue indices must be integers")
-        if idx < 0:
-            idx += len(self)
-        absolute_idx = self.cursor + idx
-        if self.rows is None or idx < 0 or absolute_idx >= int(self.rows.shape[0]):
+        if self.rows is None:
             raise IndexError(idx)
-        return self.rows[absolute_idx]
+        # Talker only peeks at the first row during decode, so keep this hot
+        # path O(1) and avoid consolidating the queued tensor chunks.
```
edwingao28: please follow repo comment style

sglang_omni/models/qwen3_omni/pending_text_queue.py
```diff
@@ -87,15 +112,15 @@ def append_rows(self, rows: torch.Tensor) -> None:
         rows = _as_rows(rows)
         if rows is None:
             return
-        if self.rows is None or len(self) == 0:
+        appended_rows = int(rows.shape[0])
+        if self.rows is None:
```
edwingao28: P3 (non-blocking): The previous append path treated `len(self) == 0` as an empty queue, but this branch only checks `rows is None`. A queue constructed with an exhausted cursor (`cursor == rows.shape[0]`) reports one pending row after an append, while `queue[0]` still indexes the exhausted head and raises `IndexError`. Could we reset or promote the head when there are no remaining rows and add a regression test?

sglang_omni/models/qwen3_omni/pending_text_queue.py
```diff
@@ -49,34 +63,45 @@ def __bool__(self) -> bool:
...
+            yield from chunk
 
     def __getitem__(self, idx: int) -> torch.Tensor:
-        if not isinstance(idx, int):
-            raise TypeError("PendingTextTensorQueue indices must be integers")
-        if idx < 0:
-            idx += len(self)
-        absolute_idx = self.cursor + idx
-        if self.rows is None or idx < 0 or absolute_idx >= int(self.rows.shape[0]):
+        if self.rows is None:
```
edwingao28, line 86: P3 (non-blocking): This removes the explicit integer-index validation, so slices and tensor indices are now accepted even though the signature expects an integer and the comment below says the original arbitrary-index behavior is preserved. Could we retain the previous `TypeError` guard before the `idx == 0` fast path?

tests/unit_test/qwen3_omni/test_talker.py
```diff
@@ -353,13 +353,65 @@ def test_pending_text_queue_rejects_non_tensor_input() -> None:
...
+    assert not queue
+
+
+def test_pending_text_queue_appends_after_partial_consumption() -> None:
+    queue = PendingTextTensorQueue.from_tensor(
+        torch.tensor([[1.0], [2.0], [3.0]], dtype=torch.float32)
+    )
+
+    assert queue.popleft().item() == 1.0
+    queue.append_rows(torch.tensor([[4.0], [5.0]], dtype=torch.float64))
```
edwingao28, line 408: Non-blocking suggestion: avoiding `torch.cat` during append is the core performance invariant of this change. Could we add a focused regression test asserting that repeated appends, including an append after partial consumption, do not invoke `torch.cat`? The current FIFO and queue-identity assertions could still pass if concatenation were accidentally reintroduced.

sglang_omni/models/qwen3_omni/pending_text_queue.py
```diff
@@ -87,15 +112,15 @@ def append_rows(self, rows: torch.Tensor) -> None:
...
         if rows is None:
             return
-        if self.rows is None or len(self) == 0:
+        appended_rows = int(rows.shape[0])
+        if self.rows is None:
             self.rows = rows
             self.cursor = 0
-            return
+        else:
+            rows = rows.to(device=self.rows.device, dtype=self.rows.dtype)
```
edwingao28: `append_rows()` validates rank but does not validate the hidden width against the existing queue. For example, appending `(1, 4)` after `(2, 3)` succeeds here, whereas the previous `torch.cat` failed immediately. The malformed chunk then fails later during arbitrary indexing or when Talker adds it to the feedback tensor. Could we validate `rows.shape[1]` against the queue width here and add a regression test?

edwingao28 (commented):
The hidden-width validation is the main correctness concern. The other inline comments are non-blocking compatibility and regression-test suggestions.

edwingao28:
Since the goal is to complete #1018 2.6, this PR currently covers only the pending-text queue portion. 
The per chunk drain path is still unchanged. Please also batch the drain path so work is scheduled once per step rather than once per chunk, then provide a matched c16 before/after profile showing reduced event loop CPU share/task and thread pool activity, together with streaming order and request lifecycle regression coverage. Thanks! @cuzmi

cuzmi:
@edwingao28 After measuring the current receive path, I don’t think we have enough evidence to change it in this PR. The overhead we measured is small, while the other options add complexity and may reduce the current read concurrency.

The current flow is roughly:

```text
DataReady messages
    │
    ├─ Chunk 1 → create Task 1 → async read ─┐
    ├─ Chunk 2 → create Task 2 → async read ─┼─ concurrent reads
    └─ Chunk 3 → create Task 3 → async read ─┘
                                                │
                                                ▼
                                     per-lane ordering
                                                │
                                                ▼
                                         ordered delivery
```

Each chunk creates a Task. The Tasks can read data in parallel and then deliver the results in order.

Based on this flow, I thought the overhead to come from three areas: waiting for the previous Task, creating Tasks, and repeatedly pausing and waking up Tasks.

I ran several tests using `seedtts_stress`, and the results were:

- **Task waiting:** across several measured runs, no Task waited for an unfinished previous Task.
- **Task creation:** about 36K Tasks were created. Creating and registering them accounted for at most 0.56%–0.59% of the benchmark wall time, so this does not appear to be a major cost.
- **Task wakeups:** because the benchmark ran in a remote environment, I was not able to reliably measure the CPU cost of suspending and resuming the receive Tasks. I therefore used event-loop and total process CPU usage as indirect signals. Event-loop CPU usage remained low, and total process CPU was stable, so Task wakeups do not currently appear to be a major cost.

This does not mean Task scheduling is free. It means the current evidence is not strong enough to justify a more complex receive path.

## What I tried or considered

- **Receive fast path:** I implemented and tested this. Compared with the baseline, the combined main-thread CPU time of the Talker and Code2Wav stages increased from 110.6 to 123.4 ms per request (+11.7%), and mean inter-chunk latency increased by about 28%.
- **Using the Task as the lane tail:** this removes one extra `Future`, but still creates one Task for every chunk, so the benefit is limited.
- **One worker per lane:** this could reduce the number of Tasks, but it may turn parallel reads into sequential reads. It would also make cancellation and shutdown more complicated.

Based on the current results, I recommend keeping the existing receive path in this PR. If Task wakeups later show a meaningful CPU cost, we can look at batching or a worker-based design separately.

sglang_omni/models/qwen3_omni/pending_text_queue.py
cuzmi: Fixed. Exhausted queues are now reset before append, with a regression test covering this case.

sglang_omni/models/qwen3_omni/pending_text_queue.py
cuzmi: Fixed

sglang_omni/models/qwen3_omni/pending_text_queue.py
cuzmi: The integer-index guard is restored before the fast path, with regression coverage for invalid indices.

tests/unit_test/qwen3_omni/test_talker.py
cuzmi: Added a regression test confirming that repeated appends, including after partial consumption, do not call `torch.cat`

sglang_omni/models/qwen3_omni/pending_text_queue.py
cuzmi: Fixed. `append_rows()` now validates the hidden width against the existing queue, with regression coverage.

edwingao28 (approved):

PR #1614, opened 2026-08-18
Simplify and centralize Prefill CUDA Graph bucket policy

AkazaAkane:
## Summary

- extend the shared prefill CUDA Graph ladder with 1- and 2-token buckets
- centralize effective capture-cap and bucket resolution across model defaults and operator overrides
- preserve explicit operator bucket lists while deriving their effective `max_bs`
- remove lock-provenance validation and retain padding-gap/unreachable-bucket diagnostics as warnings
- migrate Qwen3-ASR, Fun-ASR, Higgs-TTS, and MOSS-Transcribe-Diarize to declare capture budgets instead of building ladders

## Impact

Model-qualified defaults are now generated against the minimum applicable prefill-token cap (`cuda_graph_max_bs_prefill`, `max_prefill_tokens`, positive `chunked_prefill_size`, `max_total_tokens`, and `context_length`). Explicit operator buckets remain unchanged, and SGLang-resolved ladders are accepted regardless of whether they were marked as explicitly supplied.

## Validation

- `.venv/bin/pre-commit run --files sglang_omni/scheduling/generation_batch_policy.py sglang_omni/scheduling/engine_factory.py sglang_omni/models/qwen3_asr/engine_builder.py sglang_omni/models/fun_asr/engine_builder.py sglang_omni/models/higgs_tts/engine_builder.py sglang_omni/models/moss_transcribe_diarize/engine_builder.py tests/unit_test/scheduling/test_prefill_graph_policy.py tests/unit_test/qwen3_asr/test_pipeline.py tests/unit_test/fun_asr/test_pipeline.py tests/unit_test/higgs_tts/test_pipeline.py tests/unit_test/moss_transcribe_diarize/test_pipeline.py`
- unit tests and benchmarks were not run, per repository instructions

### DIFF
--- a/sglang_omni/scheduling/generation_batch_policy.py
+++ b/sglang_omni/scheduling/generation_batch_policy.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 
 import logging
 from collections.abc import Iterable, Mapping
-from dataclasses import dataclass
 from numbers import Integral
 from typing import Any
 
@@ -19,13 +18,6 @@ _MISSING = object()
 # A prefill replay falls back to eager when the padded bucket exceeds this
 # multiple of the real token count.
 _PREFILL_PADDING_FACTOR = 2
-
-
-@dataclass(frozen=True)
-class PrefillBucketAnalysis:
-    bucket_count: int
-    max_bucket: int
-    eager_valleys: tuple[tuple[int, int], ...]
 
 
 def get_decode_cuda_graph_max_bs(server_args: Any) -> Any:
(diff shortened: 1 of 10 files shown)


sglang_omni/scheduling/generation_batch_policy.py
```diff
@@ -303,9 +382,11 @@ def _validate_prefill_graph_policy(
 
     max_bs = prefill_cfg.max_bs
     if max_bs is not None and max(buckets) != int(max_bs):
-        errors.append(
+        logger.warning(
```
edwingao28, line 385: Keep this as an error. SGLang can clamp `max_bs` to `context_length` without
trimming explicit buckets, so `[128, 256]` with `context_length=123` still
makes the runner capture 256 tokens and can cause illegal memory access.
Add a regression test using the final resolved `ServerArgs`.

sglang_omni/scheduling/generation_batch_policy.py
```diff
@@ -70,6 +71,76 @@ def build_default_prefill_cuda_graph_bs(max_num_tokens: int) -> list[int]:
     return values
 
 
+def resolve_prefill_cuda_graph_capture_cap(
```
edwingao28, line 74: Both new helpers have only one production caller. Inline the cap calculation and collapse the resolver unless another caller is planned.

PR #1616, opened 2026-08-19
[Feature] Per-stage absolute KV cache byte budgets (reworks #1452)

JiaxinD:
## Motivation

Implements the absolute byte budget for KV memory from #1399 section 4.4: fraction based sizing makes the KV pool depend on card size and co-tenant behavior, a byte budget makes it deterministic. Reworks #1452 by @adityavaid on top of current main, keeping its schema surface, size string parser, deprecation path, configurator injection point, and preflight framework; every commit carries Co-authored-by credit.

```yaml
stages:
  - name: thinker
    total_reserve_bytes: 10GiB    # declared per replica total VRAM budget
    engine:
      kv_cache_bytes: 2GiB        # engine KV pool size, authoritative, per rank
```

Integer bytes or exact KiB/MiB/GiB/TiB strings, `engine.kv_cache_bytes <= total_reserve_bytes` enforced. The branch sits on the grouped config surface from #1494.

## Modifications

* Delivery: the budget rides typed `StageLaunchConfig` fields into a scoped ambient value around the single stage factory invocation point, consumed inside `create_sglang_infrastructure`. It never enters factory kwargs by construction, and zero model files change, so no model can miss the plumbing (the per-model kwarg threading in #1452 broke MiniMax). A declared budget that no engine consumes fails startup, and a registry walking contract test covers every model config.
* Mixing guard at one launch choke point: with `kv_cache_bytes` declared, every user source of `mem_fraction_static` is rejected (typed field, CLI flags including `--encoder-mem-reserve`, direct and nested factory args or runtime overrides). Builder built-in default fractions are cleared with a log line. moss_tts_local pins a typed default and fails closed by design.
* Colocation validates in the byte domain: a byte-budgeted stage sharing a GPU must declare `total_reserve_bytes` (kv alone says nothing about weights or graphs), and startup fails before spawn when `fraction x VRAM + sum(reserves)` exceeds the card, so mixed fraction/byte layouts cannot overcommit. Colocated Qwen accepts either budget form per stage.
* `total_reserve_bytes` is enforced: stage startup caps the process's torch allocator at the summed declared reserves (accumulating across colocated stages in one worker), so a stage that outgrows its budget fails itself instead of a co-tenant. `enforce_total_reserve: false` opts out. The cap bounds torch allocations only; declared budgets need headroom above the measured model footprint for CUDA context overhead.
* Engine authority: the KV configurator returns `kv_cache_bytes` verbatim after an explicit free-memory check (actionable error instead of a bare CUDA OOM), the SGLang 0.5.16 post-capture resize raises instead of silently shrinking a byte-budgeted pool, and mamba/hybrid models are refused rather than mis-sized.
* mps_dp preflight: hard bound `replicas x kv_cache_bytes <= VRAM` with and without WEIGHT_SHARE (the shared floor also charges every further replica's private KV pool), plus reserve bounds when `total_reserve_bytes` is declared. An omitted reserve produces a warning, never an invented default, and errors cite only numbers the user wrote. Byte-budgeted MPS no longer requires `max_total_tokens`: replicas are verified equal against each other, and declaring both knobs on one stage is rejected in `EngineArgs`, since a lower token cap silently shrinks the byte-derived pool (measured at 1/3 to 1/6 of the declared budget across five models). The same bounds hold on the runtime native path merged after this branch's base (`processes.num_replicas` plus `runtime.mps`): the token cap exclusion lives in the schema, and `validate_gpu_capacity` checks the summed KV pools of the replica expanded stages against the card before spawn, so neither bound depends on the `examples/mps_dp` launcher.

## Testing

Base sgl-project/main 215e32a6, sglang 0.5.16, H200.

* Full unit suite: 5069 passed, 3 skipped. The one intermittent failure (a 10s spawn timeout in test_ipc under heavy co-tenant load) reproduces identically on pristine main in back-to-back runs, unrelated to this diff.
* GPU smoke: an over-committed layout (2 x 80GiB KV) is rejected at preflight before any resource exists; qwen3_asr DP2 with an 8GiB budget boots, logs `gpu_mem_accounting=kv_cache_bytes`, and serves correct transcriptions on both replicas; N=1 with `SGLANG_ENABLE_POST_CAPTURE_KV_SIZING=1` keeps the byte derived pool size and serves.
* An adversarial dual review (Codex plus independent verifiers) ran before opening; confirmed findings are closed in the final commit.
* Five-model H200 campaign (qwen3_asr, whisper, moss_td, higgs, moss_tts_local): preflight manifests, single-replica and same-GPU DP2 MPS boots with real inference smokes all pass; with no token cap the resolved pool matches the declared bytes exactly (16GiB -> 149,796 tokens at 112KiB per token).
* Review round 1 is addressed in the two newest commits: typed spec fields, byte-domain colocation capacity check, allocator cap, MPS token-cap decoupling, deprecation warnings dropped.
* Rebased onto main at 216e946d, which carries the native MPS runtime from #1635. Rerun on H200 in a container matching main's sglang 0.5.18 pin: the full unit suite is 5826 passed, 5 skipped, no failures.
* Native launch path smoke on one H200 (`processes.num_replicas` with `--mps on`, Qwen3-ASR 1.7B), all before any replica exists: a 200GiB KV pool on a 140.40GiB card is rejected, two same-GPU replicas at `total_reserve_bytes=100GiB` are rejected, and `engine.kv_cache_bytes` beside `engine.max_total_tokens` fails config load. The accepted layout (8GiB KV plus a 40GiB reserve per replica) boots both replicas on one card, each logging `gpu_mem_accounting=kv_cache_bytes kv_cache_bytes=8.00GiB`, and serves a correct transcription.

### DIFF
--- a/examples/mps_dp/config.py
+++ b/examples/mps_dp/config.py
@@ -4,7 +4,9 @@
 from __future__ import annotations
 
 import argparse
+import sys
 from pathlib import Path
+from typing import Any
 
 import yaml
 
@@ -13,6 +15,7 @@ from sglang_omni.config.runtime import (
     resolve_stage_factory_kwargs,
     resolve_stage_typed_kwargs,
 )
+from sglang_omni.utils.gpu_memory import format_bytes_gib, get_gpu_device_info
 
 # Note (Jiaxin Deng): the machine-readable weight-share support registry:
 # pipeline configs whose documented launch.sh command passed shared-DP
@@ -35,14 +38,13 @@ WEIGHT_SHARE_VALIDATED_CONFIGS: dict[str, str] = {
 }
 
 
-def resolve_max_total_tokens(
+def _resolve_generation_stage(
     config_path: str | Path,
-    max_total_tokens_override: int | None = None,
     *,
     require_single_sglang_engine: bool = False,
     weight_share: bool = False,
-) -> tuple[str, int | None]:
-    """Return the generation stage name and its KV cap (None when unpinned)."""
+) -> tuple[Any, Any, Any]:
+    """Return the pipeline config, its type, and the generation stage."""
 
     pipeline_config = ConfigManager.from_file(str(config_path)).config
     config_type = type(pipeline_config)
@@ -81,6 +83,152 @@ def resolve_max_total_tokens(
         raise ValueError(
             f"generation stage {stage_name!r} is missing from the pipeline"
         )
+    return pipeline_config, config_type, stage
+
+
+def _resolve_mps_memory_budget(
+    config_path: str | Path,
+    gpu_id: int,
+    replicas: int,
+    *,
+    allow_missing_budget: bool,
+    weight_share: bool = False,
+) -> dict[str, int | str] | None:
+    if replicas <= 0:
+        raise ValueError("replicas must be a positive integer")
+
+    _, config_type, stage = _resolve_generation_stage(
+        config_path, weight_share=weight_share
+    )
+    kv_cache_bytes = stage.engine.kv_cache_bytes if stage.engine is not None else None
+    total_reserve_bytes = stage.total_reserve_bytes
+    if kv_cache_bytes is None and not allow_missing_budget:
+        raise ValueError(
+            f"{config_type.__name__} generation stage {stage.name!r} must define "
+            "positive engine.kv_cache_bytes for MPS byte-budget preflight"
+        )
+    if kv_cache_bytes is None and total_reserve_bytes is None:
+        return None
+
+    device_info = get_gpu_device_info(gpu_id)
+    if device_info.total_memory_bytes is None:
+        name = device_info.name or "unknown GPU"
+        raise ValueError(
+            f"GPU {gpu_id} ({name}) total VRAM metadata is unavailable; cannot "
+            "preflight byte budgets"
+        )
+    total_memory_bytes = device_info.total_memory_bytes
+    gpu_name = device_info.name or "unknown GPU"
+
+    budget: dict[str, int | str] = {
+        "gpu_id": gpu_id,
+        "gpu_name": gpu_name,
+        "gpu_device_id": (
+            "unknown" if device_info.device_id is None else device_info.device_id
+        ),
+        "replicas": replicas,
+        "weight_share": "1" if weight_share else "0",
+        "total_vram_bytes": total_memory_bytes,
+        "total_vram_gib": format_bytes_gib(total_memory_bytes),
+    }
+
+    # Note (Jiaxin Deng): the pass criteria only ever multiply numbers the user
+    # wrote; the KV bound holds with WEIGHT_SHARE too (KV pools stay private).
+    if kv_cache_bytes is not None:
+        total_kv_bytes = replicas * kv_cache_bytes
+        if total_kv_bytes > total_memory_bytes:
+            raise ValueError(
+                "MPS byte-budget preflight exceeds physical VRAM: "
+                f"GPU {gpu_id} ({gpu_name}) has "
+                f"{format_bytes_gib(total_memory_bytes)}, but {replicas} "
+                f"replicas x kv_cache_bytes={format_bytes_gib(kv_cache_bytes)} "
+                f"= {format_bytes_gib(total_kv_bytes)} of KV pools alone. Lower "
+                "engine.kv_cache_bytes or reduce the replica count."
+            )
+        budget["per_replica_kv_cache_bytes"] = kv_cache_bytes
+        budget["per_replica_kv_cache_gib"] = format_bytes_gib(kv_cache_bytes)
+        budget["total_kv_cache_bytes"] = total_kv_bytes
+        budget["total_kv_cache_gib"] = format_bytes_gib(total_kv_bytes)
+
+    if total_reserve_bytes is None:
+        if replicas >= 2:
+            even_split = total_memory_bytes // replicas
+            print(
+                "warning: total_reserve_bytes is not set; the "
+                f"{replicas}-replica total footprint (weights, activations, CUDA "
+                "graphs) is NOT validated, only the KV pool lower bound. Size "
+                "each replica against roughly the even split of GPU "
+                f"{gpu_id} ({format_bytes_gib(even_split)}) and set "
+                "total_reserve_bytes explicitly to enable the full check.",
+                file=sys.stderr,
+            )
+        return budget
+
+    budget["per_replica_total_reserve_bytes"] = total_reserve_bytes
+    budget["per_replica_total_reserve_gib"] = format_bytes_gib(total_reserve_bytes)
+    if weight_share:
+        # One full reservation covers replica 0 (weights resident once); every
+        # further replica still owns a private KV pool.
+        shared_floor_bytes = total_reserve_bytes
+        if kv_cache_bytes is not None:
+            shared_floor_bytes += (replicas - 1) * kv_cache_bytes
+        if shared_floor_bytes > total_memory_bytes:
+            raise ValueError(
+                "MPS byte-budget preflight exceeds physical VRAM: "
+                f"GPU {gpu_id} ({gpu_name}) has "
+                f"{format_bytes_gib(total_memory_bytes)}, but one replica's "
+                "total_reserve_bytes="
+                f"{format_bytes_gib(total_reserve_bytes)} plus {replicas - 1} "
+                "further private KV pool(s) totals "
+                f"{format_bytes_gib(shared_floor_bytes)}. Lower "
+                "total_reserve_bytes or engine.kv_cache_bytes, or "
+                "reduce the replica count."
+            )
+        print(
+            "warning: WEIGHT_SHARE=1 - MPS byte-budget preflight only checked "
+            "the KV pool lower bound and one replica's reservation "
+            f"({format_bytes_gib(total_reserve_bytes)}) against GPU {gpu_id} "
+            f"VRAM ({format_bytes_gib(total_memory_bytes)}). The {replicas}-way "
+            "total is NOT validated, because the shared weight size is unknown "
+            "until a replica boots. Use autodp.sh to size shared-weight DP "
+            "from a measured footprint.",
+            file=sys.stderr,
+        )
+        return budget
+
+    requested_total_bytes = replicas * total_reserve_bytes
+    budget["requested_total_bytes"] = requested_total_bytes
+    budget["requested_total_gib"] = format_bytes_gib(requested_total_bytes)
+    if requested_total_bytes > total_memory_bytes:
+        raise ValueError(
+            "MPS byte-budget preflight exceeds physical VRAM: "
+            f"GPU {gpu_id} ({gpu_name}) has "
+            f"{format_bytes_gib(total_memory_bytes)}, but {replicas} replicas x "
+            f"total_reserve_bytes={format_bytes_gib(total_reserve_bytes)} "
+            f"requested={format_bytes_gib(requested_total_bytes)}. Lower "
+            "total_reserve_bytes or reduce the replica count."
+        )
+    return budget
+
+
+def _serialize_mps_memory_budget_manifest(budget: dict[str, int | str]) -> str:
+    return "\n".join(f"mps_budget_{key}={value}" for key, value in budget.items())
+
+
+def resolve_max_total_tokens(
+    config_path: str | Path,
+    max_total_tokens_override: int | None = None,
+    *,
+    require_single_sglang_engine: bool = False,
+    weight_share: bool = False,
+) -> int | None:
+    """Return the effective generation-stage KV cap, or None when unpinned."""
+
+    pipeline_config, _, stage = _resolve_generation_stage(
+        config_path,
+        require_single_sglang_engine=require_single_sglang_engine,
+        weight_share=weight_share,
+    )
 
     value = max_total_tokens_override
     if value is None:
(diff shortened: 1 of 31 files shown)


sglang_omni/config/topology.py
```diff
@@ -226,6 +226,20 @@ def _stage_gpu_ids(
...
+    """True when a stage declares neither memory budget form.
+
+    Colocation needs each stage on a shared GPU to declare how much of the card
+    it intends to use. ``runtime.memory.kv_cache_bytes`` is the byte-budget
+    replacement for ``runtime.resources.total_gpu_memory_fraction``, and the
+    schema forbids setting both, so either one satisfies the requirement.
+    """
+    return (
+        stage.runtime.resources.total_gpu_memory_fraction is None
+        and stage.runtime.memory.kv_cache_bytes is None
```
AkazaAkane: I don't think kv_cache_bytes can satisfy this colocation guard by itself. total_gpu_memory_fraction describes the stage's total GPU footprint, while kv_cache_bytes covers only the KV pool. For example, on an 80 GiB GPU, one stage can declare total_gpu_memory_fraction=0.9 and another stage kv_cache_bytes=20GiB; this passes the current check because the summed fraction is still 0.9, even though the declared minimum is already 92 GiB before the second stage's weights/activations/graphs.

Could we either keep requiring total_gpu_memory_fraction for generic colocation, or require total_reserve_bytes for byte-budgeted colocated stages and validate the combined byte budget against physical VRAM?

examples/mps_dp/launch.sh
```diff
@@ -360,6 +362,16 @@ up() {
...
       "${config_resolver_args[@]}") \
       || die "could not resolve max_total_tokens from $config"
+    # Note (Aditya Vaid Sharma): nvidia-smi resolves GPU_ID physically, so the
+    # budget helper runs in a single-UUID namespace and queries logical GPU 0.
+    uuid=$(nvidia-smi --query-gpu=uuid --format=csv,noheader -i "$gpu")
+    local budget_args=("$config" --print-mps-memory-budget --gpu-id 0 --replicas "$n")
+    if [ "$weight_share" = 1 ]; then
+      budget_args+=(--weight-share)
+    fi
+    mps_budget_manifest=$(env CUDA_VISIBLE_DEVICES="$uuid" \
```
AkazaAkane, line 387: One remaining interaction seems worth resolving here: N>1 MPS still requires max_total_tokens even when kv_cache_bytes is declared. SGLang applies max_total_tokens as an upper bound after profiling, so a lower token cap can make the final KV pool smaller than the capacity derived from kv_cache_bytes. That makes the byte budget not fully authoritative and leaves MPS with two KV sizing knobs.

Can byte-budgeted MPS skip the mandatory max_total_tokens requirement and instead verify that all replicas resolve the same capacity? If both are intentionally supported, we should at least define which one is authoritative.

AkazaAkane (commented):
The centralized delivery and post-capture enforcement address the main issues from 1452. Several some issues to address.

adityavaid:
It has still not fully solved total-reserve semantics. In particular, it does not provide a runtime allocator cap, and its generic colocation behavior still needs the total_reserve_bytes  requirement. I think in long term it can be a change on sglang which then propagates to omni, for now we can add a TODO and explicitly mention this gap in comments.

sglang_omni/pipeline/stage_workers.py
```diff
@@ -778,18 +778,31 @@ def _construct_scheduler(
 ) -> Any:
     """Build a scheduler, serializing GPU factory work per visible device."""
 
+    from sglang_omni.scheduling.stage_kv_budget import stage_kv_cache_budget
+
     factory = import_string(spec.factory)
+    # Note (Jiaxin Deng): the byte budget rides the ambient scope, never factory
+    # kwargs, so model factory signatures stay untouched and none can miss it.
+    factory_arg_defaults = dict(spec.factory_arg_defaults)
+    kv_cache_bytes = factory_arg_defaults.pop("kv_cache_bytes", None)
```
MayDomine: The budget rides `factory_arg_defaults` purely as a transport channel, which forces two independent pop sites — this one, and the same pop in `resolve_stage_factory_args` in `config/runtime.py` — plus tests asserting the key never leaks into factory kwargs. Any future consumer of the defaults dict has to remember a third pop.

Since `StageLaunchConfig` is ours, a dedicated `kv_cache_bytes` field on the spec would carry the budget to this choke point without entering the defaults dict at all: both pops disappear, the never-in-kwargs invariant holds by construction, and the direct-invocation path stops needing a docstring to explain why it strips the key. The cost is one extra field at the spec construction sites in `mp_runner.py`. A first-class typed field is also the shape #1466's worker-spec direction would want, so this wouldn't be churn.

MayDomine (commented):
Reviewed the full diff with a focus on complexity and code size. The core mechanism holds up well: the ambient-scope delivery fixes the per-model kwarg threading that undermined #1452, the failure semantics (unconsumed scope, double consume) correspond to real misconfiguration modes rather than speculative ones, and the test coverage is thorough — the registry-completeness assertion in `test_kv_budget_contract.py` that keeps the walk from going silently vacuous is a nice touch.

All inline comments below are non-blocking suggestions. One cross-cutting observation to frame them: the byte-budget-vs-fraction exclusivity is currently enforced in four layers (schema construction, `_validate_runtime_sources`, the CLI helpers in `cli/serve.py`, and the builder-default clearing in `engine_factory.py`), with seven distinct error message variants. Each layer is individually justified today because each guards a different entry channel — construction-time YAML, post-construction mutation, CLI UX, builder internals. But that multi-channel surface is exactly what the #1466 resolver proposal collapses into a single cross-field rule with provenance-aware errors. It would be worth tagging these guard layers (a code comment or a note on #1466) as deletable once that lands, so the duplication doesn't outlive its reason or get copied by the next typed field.

sglang_omni/config/schema.py
```diff
@@ -68,6 +70,86 @@ def model_post_init(self, __context: Any = None) -> None:
             raise ValueError(
                 "runtime.resources.total_gpu_memory_fraction must be in (0, 1]"
             )
+        if value is not None:
+            warnings.warn(
```
MayDomine: Two issues with this deprecation warning (the `SGLangServerArgsConfig.mem_fraction_static` one below has the same problems):

1. It fires for stages that have no legal migration target. Encoder-type stages in colocated pipelines (`image_encoder`, `audio_encoder`, `code2wav`) build no SGLang engine, so declaring `runtime.memory.kv_cache_bytes` on them fails startup via the unconsumed-scope check in `stage_kv_budget.py` — `total_gpu_memory_fraction` is their only valid spelling, yet constructing their config now warns them to migrate to a field that would crash them. Mamba/hybrid models are in the same position, since the byte branch explicitly refuses them.
2. `stacklevel=2` from inside `model_post_init` points into pydantic's internals rather than the user's config, so the warning cannot tell the user which stage or file triggered it.

Given the PR description already defers field migration to the upcoming schema-unification PR, I'd drop both warnings here and let that PR introduce deprecation with per-stage-kind scoping — #1466's compat/migrate machinery is built for exactly this.

sglang_omni/config/runtime.py
```diff
@@ -141,6 +164,37 @@ def _validate_runtime_sources(
 ) -> None:
     """Validate ownership of runtime fields."""
 
+    typed_kv_cache_bytes = stage_cfg.runtime.memory.kv_cache_bytes
```
MayDomine: For the record rather than a change request: this block is the second of three places restating the fraction-vs-byte-budget exclusivity (`StageRuntimeConfig.model_post_init` and the CLI helpers being the others). The recheck here is genuinely necessary today — CLI overrides mutate the config after construction and bypass `model_post_init`, as `test_kv_budget_rejects_post_construction_typed_mem_fraction` documents. Under #1466's single-resolver model, all of these collapse into one cross-field rule. A short comment marking this layer as a casualty of the multi-entry-channel surface (and deletable with it) would keep the next typed field from copying the pattern as if it were load-bearing design.

sglang_omni/config/placement.py
```diff
@@ -159,25 +172,34 @@ def _resolve_stage_gpu_ids(stage: StageConfig) -> tuple[int, ...]:
...
+        else:
+            total_kv_cache_bytes += entry.kv_cache_bytes
     return GpuPlacement(
         gpu_id=gpu_id,
         stage_names=tuple(stage_names),
         total_gpu_memory_fraction=total,
         has_memory_fraction=has_memory_fraction,
         missing_fraction_stage_names=tuple(sorted(missing)),
+        total_kv_cache_bytes=total_kv_cache_bytes,
+        missing_kv_cache_bytes_stage_names=tuple(sorted(missing_kv_cache_bytes)),
```
MayDomine: `total_kv_cache_bytes` and `missing_kv_cache_bytes_stage_names` currently have no consumer beyond the launcher log summary — nothing compares the per-GPU byte total against card capacity (fractions do get the colocation-limit check in `topology.py`). Meanwhile on the common all-fraction deployment, every GPU will log every stage under `missing_kv_cache_bytes_stages`, which reads like a misconfiguration when nothing is wrong.

Two ways to make this accounting earn its lines: have the launcher, which already resolves `get_gpu_device_info`, compare `total_kv_cache_bytes` against total VRAM as an early startup failure (the byte analogue of the fraction limit check), or keep only the total and drop the missing-list. The current state — summed, logged, never checked — is the weakest middle ground.

examples/mps_dp/config.py
```diff
@@ -77,6 +77,164 @@ def resolve_max_total_tokens(
...
+            f"GPU {gpu_id} ({gpu_name}) has "
+            f"{format_bytes_gib(total_memory_bytes)}, but {replicas} replicas x "
+            f"total_reserve_bytes={format_bytes_gib(total_reserve_bytes)} "
+            f"requested={format_bytes_gib(requested_total_bytes)}. Lower "
+            "runtime.memory.total_reserve_bytes or reduce the replica count."
+        )
+    return budget
+
+
+def resolve_mps_memory_budget(
```
MayDomine: This strict wrapper has no production caller: `launch.sh` only goes through `--print-mps-memory-budget`, i.e. the `allow_missing_budget=True` path, and all eleven call sites of the strict form are in `test_mps_dp_support.py`. Suggest dropping the wrapper and having the tests call `_resolve_mps_memory_budget(..., allow_missing_budget=False)` directly — that also removes the `assert budget is not None` narrowing.

sglang_omni/model_runner/sglang_model_runner.py
```diff
@@ -68,6 +92,35 @@ def _profile_available_bytes(self, pre_model_load_memory: float) -> int:
         upstream SGLang profiling semantics for ordinary non-colocated AR
         serving.
         """
+        if self.kv_cache_bytes is not None:
+            if self.kv_cache_bytes <= 0:
```
MayDomine, line 94: This positivity check has no reachable path: `kv_cache_bytes` gets here only via `StageMemoryConfig`, whose validator already rejects non-positive values at parse time. Fine to keep as a tripwire for programmatic construction, but then an `assert` states that intent better than a `ValueError` branch that reads like a real case.

examples/mps_dp/launch.sh
```diff
@@ -360,6 +362,16 @@ up() {
...
+      "$PYTHON_BIN" "$SCRIPT_DIR/config.py" "${budget_args[@]}") \
```
stephenkgli, line 388: The byte-budget preflight resolves only `CONFIG`, while the actual replica command appends `SERVE_EXTRA_ARGS` afterwards. Since `serve` accepts dotted config overrides, an extra argument can replace a `kv_cache_bytes` value that already passed preflight, so the launcher creates its state and starts MPS using a stale budget and detects the overcommit only during replica startup.

zhaochenyang20:
we shall postpone this for a while after the config refactor.

sglang_omni/config/topology.py
JiaxinD: Done in b2d73155: byte-budgeted colocated stages now require total_reserve_bytes, and startup fails before spawn when fraction x VRAM plus summed reserves exceeds the card.

examples/mps_dp/launch.sh
JiaxinD: Done in b2d73155: byte-budgeted MPS drops the mandatory max_total_tokens, replicas are verified equal against each other, and both knobs on one stage are rejected. Your read was measured: with the cap the resolved pool was 1/3 to 1/6 of the declared bytes across five models.

sglang_omni/pipeline/stage_workers.py
JiaxinD: Done: the budget now rides typed StageLaunchConfig fields, both pops are gone, and the never-in-kwargs invariant holds by construction.

sglang_omni/config/schema.py
JiaxinD: Done, both warnings dropped; deprecation moves to the schema unification PR with per-stage-kind scoping.

sglang_omni/config/placement.py
JiaxinD: Went with the capacity check: placement sums reserve bytes per GPU and validate_gpu_capacity fails before spawn on fraction x VRAM plus reserves; the missing list is dropped.

examples/mps_dp/config.py
JiaxinD: done

sglang_omni/model_runner/sglang_model_runner.py
JiaxinD: Keeping the ValueError: the path is reachable via direct ModelWorkerConfig construction, which skips schema validation.

JiaxinD:
Landed rather than deferred: colocation now requires total_reserve_bytes with a combined bytes vs VRAM check, and total_reserve_bytes is enforced at stage startup via a per-process torch allocator cap (opt-out flag, offenders fail instead of neighbors). The deeper sglang-side budget accounting stays as the long term note.

examples/mps_dp/launch.sh
JiaxinD: Addressed on the rebased head: the N>1 requirement is now satisfied by a declared `engine.kv_cache_bytes` alone, so a byte-budgeted config no longer needs MAX_TOTAL_TOKENS. Supplying both is still rejected, since a lower token cap would shrink the byte-derived pool.

examples/mps_dp/launch.sh
JiaxinD: Fixed on the rebased head: SERVE_EXTRA_ARGS carrying kv_cache_bytes, total_reserve_bytes or mem_fraction_static is now rejected before any replica starts, because the preflight resolves CONFIG alone and an override there would be validated by nothing.

JiaxinD:
Rebased onto main now that #1494 has landed, so the postpone condition is cleared and the title no longer carries it. The feature itself is unchanged; the rebase needed `StagePlacement`'s new byte fields to default to None, and the bootstrap test doubles to drop the prefill and decode managers main removed.

Unit suites in a container: no failure is attributable to this branch. 59 failures are identical on plain main here (the container ships sglang 0.5.16 against main's 0.5.18 pin), and the one remaining KV test fails on the same missing `is_config_namespace_published` that already fails three of its siblings on main, so it cannot be validated until the container matches the pin.

stephenkgli (approved):
LGTM

PR #1617, opened 2026-08-19
[Perf][FishAudio S2-Pro] Add breakable Prefill CUDA Graph support

charliechenye:
## Motivation

FishAudio S2-Pro builds request specific VQ embeddings during prefill. Those embeddings must remain outside `ForwardBatch.input_embeds` while the shared SGLang breakable Prefill CUDA Graph path decides graph eligibility.

Fish also requires a narrower graph replay boundary than its full model forward. The transformer body can be captured and replayed, while last token selection, logits, semantic sampling, and codebook generation must remain eager.

This PR adapts FishAudio S2-Pro to the shared Prefill CUDA Graph contract tracked in #1357.

The integration remains opt in. This PR does not enable `cuda_graph_backend_prefill` by default and does not define a Fish specific default prefill graph ladder.

## Changes
* Mark FishAudio S2-Pro as compatible with breakable Prefill CUDA Graph in the model capability and engine builder.
* Route request specific prefill embeddings through `OmniPrefillInputs` instead of exposing them through `ForwardBatch.input_embeds` during graph admission.
* Expose the Fish transformer body as the Prefill CUDA Graph replay boundary.
* Keep last token selection, logits, semantic sampling, and codebook generation outside the prefill graph.
* Accept the shared `omni_prefill_rids` forward argument required by the Omni prefill transport path.
* Add focused tests for sidecar transport and eager tail preservation.

### CPU and contract

Current head: `41db4904051bb3654d551bc57d1cefcfd32bb566`

Base: `65eb44ba5a54a1ff142c2306d384e36d59a2b84a`

Using SGLang `0.5.16`:
* Focused prefill replay contract: 1 passed.
* Full FishAudio S2-Pro unit suite: 106 passed, 1 skipped.
* Shared prefill and capability suite: 110 passed.

The focused replay test patches the real Fish transformer body, verifies that the outer eager tail receives the logical last token rows, enables VQ decode, and verifies the hidden states and logits consumed by the eager tail.

The runner test also verifies the exact composed prefill embeddings while `ForwardBatch.input_embeds` remains unset.

### Correctness and quality

The full SeedTTS English set was evaluated with matched Prefill CUDA Graph OFF and ON generation.

Configuration:

```text
Model: fishaudio/s2-pro
Model revision: 1de9996b6be38b745688de084d87a5633f714e4e
Qualification code: 61e7e80fac0303fc3f23c42b71e02d79890a97cb
Dataset: zhaochenyang20/seed-tts-eval-arrow
Dataset revision: 27f4c1adee83b5b29b7c4b375f6b976324bda308
Split: en
Samples: 1088
Hardware: NVIDIA H200
Dtype: BF16
Generation concurrency: 16
max_new_tokens: 2048
mem_fraction_static: 0.80
Seed: 20260817
ASR model: Qwen/Qwen3-ASR-1.7B
ASR revision: 7278e1e70fe206f11671096ffdd38061171dd6e5
```

Results:
* Corpus WER: 1.0383% OFF, 1.0466% ON, a difference of +0.0084 percentage points.
* Per sample mean WER: 0.9993% OFF, 1.0086% ON.
* Median WER: 0% in both modes.
* p95 WER: 8.3333% in both modes.
* Maximum WER: 42.8571% in both modes.
* Samples above 50% WER: 0 in both modes.
* All 1088 samples generated and scored successfully in both modes.

The paired comparison contained 1067 unchanged samples, 11 samples improved with Prefill CUDA Graph, and 10 samples worsened with Prefill CUDA Graph.

The aggregate quality result is therefore neutral.

### Exact shape and padded replay attribution

Current head was also tested with the same logical request under exact shape BCG and the next bucket from the shared validation ladder.

Radix cache was disabled for this attribution test.

Three representative prefills were tested twice in each mode:
* 153 tokens with exact bucket 153 and padded bucket 160.
* 161 tokens with exact bucket 161 and padded bucket 176.
* 199 tokens with exact bucket 199 and padded bucket 208.

For all three requests:
* The exact shape execution repeated the same semantic and codebook sequence across both runs.
* The padded execution repeated the same semantic and codebook sequence across both runs.
* The request specific input embedding fingerprint was identical between exact and padded execution.
* Graph routing was valid in both modes.
* Exact and padded execution did not produce the same semantic and codebook sequence.

The generated step counts were:

```text
153 tokens: exact 46, padded 47
161 tokens: exact 97, padded 92
199 tokens: exact 95, padded 94
```

This shows that Fish S2-Pro is deterministic within a fixed execution shape, but padding can change the generated semantic and codebook trajectory.

This PR therefore does not claim bit exact output parity between exact and padded replay.

The full 1088 sample quality comparison above exercises the shared padded graph path and showed no observed aggregate WER regression.

Exact shape eager and exact shape BCG are also each internally repeatable but not bit identical to each other. The same qualitative behavior reproduced on the pre cleanup head `cf4e7122f12bb016487f6e6ca2f2d83929fc5773`, so this behavior was not introduced by the cleanup of the current implementation.

### Performance

The full end to end H200 benchmark used all 1088 SeedTTS English samples at each concurrency, with three measured repeats and alternating OFF and ON order.

Configuration:

```text
Model: fishaudio/s2-pro
Model revision: 1de9996b6be38b745688de084d87a5633f714e4e
Qualification code: 61e7e80fac0303fc3f23c42b71e02d79890a97cb
Hardware: NVIDIA H200
Dtype: BF16
mem_fraction_static: 0.80
max_new_tokens: 2048
Seed: 20260817
Concurrencies: 1, 2, 4, 8, 16
Measured repeats: 3
Prefill validation cap: 4096
```

Mean throughput:
* c1: 0.733 QPS OFF, 0.734 QPS ON, +0.14%.
* c2: 1.070 QPS OFF, 1.105 QPS ON, +3.21%.
* c4: 1.943 QPS OFF, 1.968 QPS ON, +1.27%.
* c8: 3.131 QPS OFF, 3.156 QPS ON, +0.78%.
* c16: 4.608 QPS OFF, 4.664 QPS ON, +1.22%.

Mean latency:
* c1: 1.364 s OFF, 1.362 s ON, +0.15% improvement.
* c2: 1.872 s OFF, 1.810 s ON, +3.29% improvement.
* c4: 2.056 s OFF, 2.030 s ON, +1.23% improvement.
* c8: 2.551 s OFF, 2.531 s ON, +0.76% improvement.
* c16: 3.460 s OFF, 3.418 s ON, +1.21% improvement.

BCG was faster in all three measured repeats from c2 through c16. c1 was effectively neutral.

The complete campaign covered 30 benchmark arms and 32640 generated requests with zero request failures.

BCG request coverage was 1088 of 1089 requests per ON run, or 99.91%. The extra request was the benchmark warmup.

The result supports a modest end to end performance benefit rather than a large fixed speedup.

### Current head performance confirmation

After simplifying the implementation, I ran focused current head A/B/C checks rather than repeating the full qualification campaign.

The states were:

```text
A: base commit, BCG OFF
B: current PR head, BCG OFF
C: current PR head, BCG ON
```

The larger c1 confirmation used 256 samples per arm and 3 rounds.

Results:
* Base eager to current eager throughput: +0.98%.
* Base eager to current eager mean latency: +0.98% improvement.
* Current eager to current BCG throughput: +0.32%.
* Current eager to current BCG mean latency: +0.31% improvement.

The c16 confirmation used 64 samples per arm and 2 rounds.

Results:
* Base eager to current eager throughput: +0.29%.
* Base eager to current eager mean latency: +0.08% improvement.
* Current eager to current BCG throughput: +0.32%.
* Current eager to current BCG mean latency: +1.38% improvement.

All arms completed with zero failed requests.

These checks confirm that the simplified current implementation does not introduce an eager performance regression and preserves the same modest positive BCG performance direction.

### Replay and fallback coverage

The larger c1 current head run contained 256 measured requests plus one benchmark warmup in each ON round.

For each round:
* All 256 measured workload requests replayed through BCG.
* Including warmup, graph request coverage was 99.61%.
* Graph useful token coverage was 99.9966%.
* Scheduled tokens were p50 148, p95 210, max 267.
* 33 graph batches were exact shape.
* 223 graph batches used padding.
* Padding overhead over graph served useful tokens was about 4.77%.

The c16 confirmation also replayed all 64 measured workload requests through BCG in each ON round.

Including warmup:
* Graph request coverage was 98.46%.
* Graph useful token coverage was about 99.99%.
* Scheduled token p50 was 180 and 173 across the two rounds.
* Scheduled token p95 was 406.45 and 418.70.
* Maximum scheduled tokens were 472 and 696.

Most successful graph execution used padded buckets rather than exact shapes.

The only eager request in these ON runs was the benchmark warmup. Its realized prefill contained 1 token. The minimum graph bucket is 4 tokens, which exceeds the shared 2x padding limit, so the shared policy correctly selected eager execution.

There were no unattributed workload fallbacks.

A separate cap boundary test also verified that a 128 token request replayed through BCG with a 128 token cap while a 153 token request correctly fell back to eager.

### Startup and memory

Matched current head c1 server readiness measurements were:

```text
BCG OFF: about 38.01 s
BCG ON:  about 42.01 s
Difference: about +4.00 s
```

This is reported as server readiness overhead. No standalone CUDA Graph capture time claim is made.

Ready state H200 memory at c1 was:

```text
BCG OFF: 120941 MiB
BCG ON:  121753 MiB
Difference: +812 MiB
```

The H200 contained 143771 MiB total memory, leaving about 22018 MiB, or 15.31%, free after BCG readiness.

At c16, the measured BCG memory increase was +824 MiB, leaving about 21952 MiB, or 15.27%, free.

The full performance campaign measured a similar BCG memory increase across concurrency levels, approximately +812 MiB to +833 MiB.

This memory cost is modest on H200 but can be material on lower memory GPUs. For that reason the optimization remains opt in.

## Scope

This PR only makes FishAudio S2-Pro compatible with the shared breakable Prefill CUDA Graph path.

It intentionally does not:
* Enable breakable prefill by default.
* Define a Fish specific default prefill graph ladder.
* Change decode CUDA Graph behavior.
* Change Fish sampling policy.
* Change codebook generation policy.
* Change vocoder behavior.
* Claim bit exact equivalence between eager, exact BCG, and padded BCG execution.
* Close the broader #1357 tracker.

The current evidence supports breakable prefill as an opt in optimization for FishAudio S2-Pro.

## Related issue

Related to #1357. This PR covers the FishAudio S2-Pro adopter only and does not close the broader tracker.

## Checklist
* [x] Format code according to pre-commit.
* [x] Add focused unit tests.
* [x] Validate real graph replay and eager fallback.
* [x] Validate output quality on the full SeedTTS English set.
* [x] Provide multi-concurrency performance evidence.
* [x] Confirm current head eager performance is preserved.
* [x] Provide current head replay and fallback coverage.
* [x] Characterize scheduled token distribution and padding coverage.
* [x] Characterize exact shape versus padded BCG behavior.
* [x] Record current head startup and H200 memory headroom.
* [ ] For reviewers: If you have not made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.

## CI

CI runs on self-hosted GPU runners and requires a maintainer to add the `run-ci` label.

The current workflow does not provide a FishAudio specific GPU preset. The Fish specific H200 qualification is therefore documented above.


charliechenye:
Current head c2019f68e0790e29019dc3a58383b939becc7c2e is pushed and the pinned CPU contract gate passed. Please add the `run-ci` label when a maintainer is ready to run the repository GPU CI. Note that the current workflow presets cover Higgs, MOSS, and Qwen3 TTS, not FishAudio; a small Fish-specific GPU qualification for #1357 still needs an H100/H200-capable runner.

sglang_omni/models/fishaudio_s2_pro/model_runner.py
```diff
@@ -80,9 +84,12 @@ def __init__(self, tp_worker: Any, output_processor: Any):
     def before_prefill(self, forward_batch, schedule_batch, requests):
         del schedule_batch
         self._sync_decode_state(requests)
-        input_embeds = self._build_prefill_input_embeds(forward_batch, requests)
-        if input_embeds is not None:
-            forward_batch.input_embeds = input_embeds
+        attach_omni_prefill_inputs(
```
AkazaAkane, line 87: Could we add matched BCG OFF vs ON speaker-similarity results here? This path specifically changes how Fish's reference/VQ conditioning reaches the transformer. WER verifies content but would not catch the reference audio being dropped or incorrectly replayed. Since SeedTTS already has the WavLM speaker-similarity pipeline, I think preserving speaker similarity should be part of the correctness qualification before marking Fish as supports_breakable_prefill_cuda_graph=True.

charliechenye:
### Speaker similarity

Because this change affects how reference derived VQ conditioning reaches the transformer during prefill, content WER alone does not cover the possibility that reference speaker conditioning could be lost or replayed incorrectly.

I therefore evaluated the same matched 1088 SeedTTS English BCG OFF and ON generations with the repository WavLM speaker similarity pipeline.

Results:

* All 1088 samples were evaluated successfully in both modes with zero skipped samples.
* Mean speaker similarity was 65.2752 OFF and 65.1924 ON.
* The mean ON minus OFF difference was -0.0828 points.
* The median paired difference was 0.0000 points.
* The paired p05 and p95 differences were -5.9347 and +5.5097 points.
* Mean absolute paired difference was 2.2139 points.
* p95 absolute paired difference was 7.2229 points.

The aggregate result therefore showed no material speaker similarity shift, but individual paired scores had wider variation.

To determine whether that spread was specific to BCG or ordinary Fish generation variability, I ran an additional current head control on 200 matched samples using two request sampling seeds. The server seed and all other configuration remained fixed.

Normal BCG OFF seed to seed variation had a mean absolute speaker similarity difference of 3.6283 points and p95 absolute difference of 9.5756 points.

BCG ON versus OFF had mean absolute differences of 2.7350 and 2.6491 points across the two request seeds, averaging 2.6921 points. The corresponding average p95 absolute difference was 7.7150 points.

The BCG to natural variation ratio for mean absolute difference was 0.742.

BCG induced paired changes also had low reproducibility across the two seeds. The signed correlation between the two ON minus OFF paired delta sets was 0.090, and the absolute delta correlation was 0.056.

Of 200 samples, 61 had an absolute BCG difference of at least 5 points in either seed, but only 9 did so in both seeds. Two samples showed a decrease of at least 5 points in both seeds, four showed an increase of at least 5 points in both seeds, and three changed direction.

These controls indicate that the per sample spread under BCG is within the ordinary stochastic variation of FishAudio S2 Pro generation and does not show a systematic loss of reference speaker conditioning.

This qualification does not claim per sample speaker score identity between eager and BCG execution.

sglang_omni/models/fishaudio_s2_pro/model_runner.py
charliechenye: Thanks, agreed that WER alone does not cover this failure mode.

I ran the repository WavLM speaker similarity evaluation on the matched full 1088 sample BCG OFF and ON SeedTTS English outputs.

All 1088 samples scored successfully in both modes with zero skipped samples. Mean similarity was 65.2752 OFF and 65.1924 ON, so the ON minus OFF difference was -0.0828 points. The median paired difference was 0.0000.

The individual paired scores had a wider spread, so I also ran a current head two seed control on 200 matched samples to calibrate that against normal Fish generation variability.

With BCG disabled, changing only the request sampling seed produced a mean absolute speaker similarity change of 3.6283 points and p95 absolute change of 9.5756 points.

BCG ON versus OFF produced mean absolute changes of 2.7350 and 2.6491 points across the two seeds, averaging 2.6921 points, with an average p95 absolute change of 7.7150 points.

The BCG paired deltas also did not reproduce strongly across seeds. Their signed correlation was 0.090 and absolute delta correlation was 0.056. Only 2 of 200 samples showed a speaker similarity drop of at least 5 points under BCG in both seeds, while 4 showed an increase of at least 5 points in both seeds.

So I do not see evidence of a systematic loss of reference speaker conditioning under BCG. The per sample variation is within the normal stochastic variation of Fish generation.

I will add this qualification to the PR. Thanks for catching the missing speaker similarity check.

PR #1590, opened 2026-08-17
[Perf][Whisper] Fuse q/k/v and cross-attention k/v projections

lijrjyan:
## Motivation

Fuses the Whisper q/k/v and cross-attention k/v projections into single GEMMs.

## Modifications

- **Fused projections** (`whisper_asr/sglang_model.py`): encoder and decoder self-attention use `QKVParallelLinear` (one q/k/v GEMM); cross-attention uses `ColumnParallelLinear` for q and `QKVParallelLinear(total_num_heads=0)` for k/v; out_proj / fc1 / fc2 become `RowParallelLinear` / `ColumnParallelLinear`, all taking the usual `quant_config`/`prefix` arguments (passed through; nothing is quantized by default).
- **Weight loading**: checkpoint q/k/v tensors go through the stacked-param shard loader; the missing `k_proj.bias` is fed explicit zeros; `proj_out.weight` still maps onto the tied decoder embedding.

## Related Issues

Roadmap #1396 (Whisper). Stacked on #1566 (W-PR6): the base branch `perf/whisper-encoder-compile-w-pr6` mirrors its head, so this diff is the fusion only; will be retargeted to `main` once #1566 merges. Follow-up to W-PR1/W-PR6.

## Accuracy Test

`tests/unit_test/whisper_asr/` 43 passed at `2c09f386f`. End-to-end: `benchmarks/eval/benchmark_asr_seedtts` (SeedTTS EN, n=200, c1/4/8) — corpus WER unchanged vs the unfused base in every cell (0.0097).

## Benchmark & Profiling

**Commits** (main is moving quickly, so both sides are pinned)
- base `1476db797` (W-PR6 + main: pre-LM encoder service, async decode) -> fused `2c09f386f` (this head)
- `*` rows: earlier pair, base `94ac1d967` -> fused `f34c87256` (before the main merge)

**Protocols**
- (A) pre-LM off: `enable_pre_lm_encoder: false`, encoder runs in-model per request
- (B) defaults: upstream defaults (pre-LM encoder service + async decode + compile + encoder graph)

Unfused -> fused req/s:

| GPU | protocol | c1 | c4 | c8 |
|---|---|---|---|---|
| (1) RTX A6000 | (A) pre-LM off | 7.64 -> 7.60 (-0.6%) | 13.77 -> 14.29 (+3.8%) | 16.56 -> 17.10 (+3.3%) |
[...]

PR #1620, opened 2026-08-19
[Whisper ASR] Fuse attention projections

Dayuxiaoshui:
## Motivation

Whisper attention currently evaluates its query, key, and value projections as separate linear layers. For `openai/whisper-large-v3`, these projections operate on the same input tensor and have compatible output dimensions, so issuing separate GEMMs adds launch and framework overhead without changing the underlying model computation.

This PR fuses the compatible attention projections into wider linear layers:

- encoder self-attention: Q, K, and V are computed by one QKV projection;
- decoder self-attention: Q, K, and V are computed by one QKV projection;
- decoder cross-attention: K and V are computed by one KV projection, while Q remains separate because it consumes decoder states rather than encoder states.

The goal is to optimize computation inside the end-to-end Whisper model rather than changing request scheduling, caching policy, tensor placement, or decoding behavior.

## Modifications

- Replace the three encoder self-attention projection layers with one `qkv_proj` layer and split its output into Q, K, and V views.
- Replace the three decoder self-attention projection layers with one `qkv_proj` layer.
- Replace decoder cross-attention K and V layers with one `kv_proj` layer while retaining the separate Q projection.
- Preserve Whisper's biasless K projection semantics by keeping the corresponding fused bias shard zero.
- Extend Hugging Face checkpoint loading to place the original Q/K/V weights and biases into strict slices of the fused parameters.
- Use `default_weight_loader` for each destination slice so malformed or broadcastable checkpoint tensors fail with a shape error instead of being silently accepted.
- Add focused unit tests for shard placement, zero K-bias behavior, numerical equivalence, decoder cross-attention loading, and invalid checkpoint shapes.

The fused layers preserve the original parameter order within each projection shard and do not change attention head layout, scaling, KV cache contents, dtype, or device placement.

Mathematically, the three separate projections in self-attention,

$$
Q = XW_Q^T + b_Q,\qquad
K = XW_K^T + b_K,\qquad
V = XW_V^T + b_V,
$$

are evaluated as one projection and then split along the output dimension:

$$
\mathrm{concat}(Q,K,V)
 = X
\begin{bmatrix}W_Q \\\\ W_K \\\\ W_V\end{bmatrix}^{T} +
\begin{bmatrix}b_Q \\\\ b_K \\\\ b_V\end{bmatrix}^{T}.
$$

For Whisper, $b_K=0$ is preserved as the middle zero bias shard in the fused QKV parameter. Decoder cross-attention uses the analogous KV form,

$$
\mathrm{concat}(K,V)
 = H
\begin{bmatrix}W_K \\\\ W_V\end{bmatrix}^{T} +
\begin{bmatrix}b_K \\\\ b_V\end{bmatrix}^{T},
$$

while $Q$ remains a separate projection because it consumes decoder hidden states and $K,V$ consume encoder hidden states. For an input with $T$ rows and model width $d$, the projection arithmetic remains equivalent:

$$
3\times(2Td^2) = 2T(3d)d,
$$

but three kernel/module dispatches become one. The optimization therefore targets launch and dispatch overhead, not model capacity or attention complexity.

## Related Issues

Related to #1396.

## Accuracy Test

Accuracy was evaluated with `openai/whisper-large-v3` in FP16 on the pinned 128-sample SeedTTS English subset. The baseline and fused implementations were each evaluated at concurrency 1, 8, 16, and 32 with three repeats per concurrency.

| Metric | Baseline | Fused projections |
| --- | ---: | ---: |
| Completed requests | 1,536 / 1,536 | 1,536 / 1,536 |
| Corpus WER | 0.008363 | 0.008363 |
| Paired normalized transcript mismatches | 0 / 1,536 | 0 / 1,536 |

Every paired normalized transcript was identical. The unit test suite for Whisper also passed:

```text
44 passed
```

The added regression tests additionally compare fused projection outputs against the original separate linear operations using the same weights and inputs.

### End-to-end benchmark

Environment and model configuration:

- GPU: one NVIDIA H200
- Model: `openai/whisper-large-v3`
- Model revision: `06f233fe06e710322aca913c1bc4249a0d71fce1`
- Dataset: SeedTTS English, 128 samples
- Dataset revision: `27f4c1adee83b5b29b7c4b375f6b976324bda308`
- Dtype: FP16
- Repeats: 3 per concurrency after warmup
- CUDA Graph: enabled
- `torch.compile`: enabled
- Maximum running requests: 32
- Static memory fraction: 0.30
- Persistent pre-LM encoder cache: disabled in both modes

The values below are the means of three complete end-to-end repeats.

| Concurrency | Baseline throughput | Fused throughput | Throughput change | Baseline P95 | Fused P95 | P95 change |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1 | 8.98 req/s | 9.56 req/s | +6.4% | 0.141 s | 0.131 s | -7.0% |
| 8 | 37.32 req/s | 38.56 req/s | +3.3% | 0.292 s | 0.282 s | -3.4% |
| 16 | 50.79 req/s | 53.31 req/s | +5.0% | 0.449 s | 0.392 s | -12.7% |
| 32 | 61.20 req/s | 63.74 req/s | +4.1% | 0.742 s | 0.717 s | -3.4% |

The improvement is consistent across the measured concurrency range. At the maximum tested concurrency, throughput increased by 4.1% while P95 latency decreased by 3.4%.

The benchmark client command was:

```bash
  --port 18080 \
  --model-path openai/whisper-large-v3 \
  --model-revision 06f233fe06e710322aca913c1bc4249a0d71fce1 \
  --dataset-revision 27f4c1adee83b5b29b7c4b375f6b976324bda308 \
  --max-samples 128 \
  --concurrencies 1 8 16 32 \
  --repeats 3 \
  --warmup \
  --dtype float16 \
  --cuda-graph \
  --torch-compile \
  --max-running-requests 32 \
  --mem-fraction-static 0.30 \
  --fingerprint \
  --output whisper_fused_qkv.json
```

### Computational effect

For each attention invocation, the change reduces projection operations as follows:

| Attention path | Before | After |
| --- | ---: | ---: |
| Encoder self-attention | 3 projection GEMMs | 1 projection GEMM |
| Decoder self-attention | 3 projection GEMMs | 1 projection GEMM |
| Decoder cross-attention encoder K/V | 2 projection GEMMs | 1 projection GEMM |

This evaluates the same learned affine projections while reducing GEMM launches and intermediate Python/module dispatch. Model weight memory remained effectively unchanged (`3.11 GB` baseline versus `3.12 GB` fused); the small difference comes from the zero bias shards used to preserve the original biasless K projections. KV cache capacity was also effectively unchanged (`82,216` versus `82,184` tokens).

## Checklist

- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed. No user-facing configuration or API changed.

## CI


0xjeffro (approved):

guozhihao-224 (approved):

lijrjyan (approved):
Same fusion I had in #1590 but with no stack dependency. just noticed stack pr will never pass ci if base is not main...
Measured on my side and get similiar advance. LGTM; closing #1590 in favor of this.

lijrjyan (approved):

PR #1625, opened 2026-08-19
[Qwen3-TTS] Deduplicate concurrent uploaded-voice prompt construction

Fyrgo8:
## Summary

Add per-key single-flight coordination for Qwen3-TTS uploaded-voice prompt construction.

When multiple requests hit the same uploaded voice while its prompt artifact is cold, one request performs `create_voice_clone_prompt`. Followers wait for it to complete, then load independent copies from `SpeakerArtifactCache`.

Related to #1418.

## Motivation

Before this change, concurrent cold requests could all observe the same cache miss and independently construct the same speaker prompt. This duplicated CPU/GPU work and increased memory pressure during the first burst of traffic for a newly uploaded voice.

## Changes

- Add keyed in-flight coordination using `SpeakerCacheKey`.
- Allow different voice keys to construct concurrently.
- Re-check `SpeakerArtifactCache` after a follower is released.
- Fan out leader exceptions to all followers.
- Remove failed in-flight entries so later requests can retry.
- Store detached CPU tensors in the shared cache and return caller-owned clones.
- Preserve existing cache-hit and voice-version invalidation behavior.

## Tests

Focused unit tests:

```text
15 passed, 109 deselected
```

Coverage includes:

- same-key cold miss invokes `create_voice_clone_prompt` exactly once;
- different voice keys remain concurrent;
- construction failures fan out to all followers;
- failed construction remains retryable;
- existing cache-hit and voice-version invalidation behavior is preserved.

## Real-model smoke test

- Model: `Qwen/Qwen3-TTS-12Hz-0.6B-Base`
- GPU: NVIDIA RTX 5090
- Single uploaded-voice request: HTTP 200, valid 24 kHz mono WAV
- Same voice, 8 concurrent cold requests: 8/8 HTTP 200
- Cache after the run: `entries=1`
- Voice overwrite invalidation: `delete_invalidation_counter=1`
- No request-level errors or prompt-construction exceptions in the server log

The deterministic concurrency unit test verifies exactly-once prompt construction. The real-model smoke test verifies concurrent serving and artifact publication.

## Scope

This change targets cold uploaded-voice cache misses and coordinates construction within a single process. It does not claim a general throughput improvement or cross-process deduplication.


JiaxinD (changes requested):
Two correctness contracts are still missing before this should merge:

1. The in-flight/cache key is only (model type, uploaded voice name, created_at, artifact kind). In ICL mode, the effective ref_text is also an input to prompt construction. If a stored uploaded voice has no transcript and two requests supply different transcripts, the second request follows the first flight (or hits its cache entry) and receives ref tokens built from the wrong transcript. Please include a digest of the effective ICL transcript in the key; x-vector-only requests can continue sharing regardless of text. Add a concurrent regression where the same voice/version with two transcripts reaches both encodes and returns distinct ref ids.

2. Voice invalidation does not invalidate _UPLOADED_VOICE_PROMPT_INFLIGHT. If DELETE /v1/audio/voices/{name} races the leader, clear_voice removes the current cache entry, but the old leader later calls speaker_cache.put and resurrects the deleted voice artifact. The single-flight state needs to be owned by the cache (or otherwise participate in clear_voice/clear) so late leaders cannot repopulate invalidated state. Please add a blocked-leader regression covering clear_voice during compute, plus explicit-put/clear precedence.

The deduplication direction and existing same-key/failure/clone-ownership tests are useful; these two races are the blockers.

JiaxinD (approved):
Verified at current head 692ba0d2: uploaded-voice single-flight ownership now lives in SpeakerArtifactCache, keys include the effective ICL transcript, and clear/explicit-put win over late leaders. The missing SpeakerCacheKey test import is fixed; exact-head unit-test and accelerator CI both pass. Approved.

JiaxinD:
@Fyrgo8 I want to acknowledge a process mistake on my side. After requesting changes, I used maintainer edit access to implement them directly on your branch without coordinating with you first. I should have prepared the fixes on my fork and offered them as a patch or cherry-pick instead. Sorry about that.

The branch moved from `3d15197d` to `692ba0d2`. I added `5891ae7f` and `b0197c7f` for cache-owned single-flight behavior, transcript-aware keys, and clear/explicit-put race handling; `1eff41a2` synced current main; and `692ba0d2` fixed the missing test import. I did not force-push, so your original head remains in the history. I believe these correctness fixes are worth retaining, but whether to keep them is your decision. I also approved the resulting revision, which was not an independent review boundary after I had modified the branch.

I will not push to this branch again. If you prefer the original branch state, please say so; I can prepare a fork-based patch or restoration instructions, but I will not change the branch without your explicit request.

Fyrgo8:
@JiaxinD Thanks for the explanation. No worries - I'm happy to keep the changes.

Fyrgo8:
@JiaxinD If there are no remaining issues with the current head, could this PR be merged when convenient?

PR #1628, opened 2026-08-19
[Qwen3-Omni Perf] Encoder: conditional batching, layer CUDA graph, small-payload transport gate

edwingao28:
## Motivation

Profiling the speech pipeline's time-to-first-token showed three independent costs on the request head path, none of them model compute:

1. The audio encoder waits a fixed `SGLANG_OMNI_ENCODER_BATCH_WAIT_MS` (50 ms) after the first message even when its queue is empty.
2. The encoder forward runs eagerly: ~460 kernel launches keep the GPU only ~8% busy (host-submission-bound), plus a per-layer `cu_seqlens.tolist()` D2H sync ×32 layers.
3. Small (~258 KB) encoder outputs pay a per-request `cudaIpcOpenMemHandle` on the receive side (13.4 ms mean, 69 ms p95 — the handle open contends on the driver lock with thinker prefill), vs 1.3 ms over the pooled relay.

## Modifications

- **Conditional encoder batch window**: dispatch immediately when the queue is idle; coalesce only under backlog.
- **Encoder layer-stack CUDA graph** (default **on**, `runtime_overrides.audio_encoder.enable_layer_cuda_graph`): capture the 32-layer attention stack with token-dim bucketing (128–4096); `cu_seqlens` splits computed once and shared across layers. One-time capture 0.4–0.6 s at boot. Capture dummies build window-sized segments so the declared `max_seqlen` bounds every segment (locked by a per-bucket shape test and a real capture smoke test).
- **Payload-size gate on stage transport**: payloads below a measured-crossover floor skip direct CUDA IPC and take the router's pooled relay (pool opened once at startup — no per-request handle open); large image/video encoder outputs keep the direct path.

## Benchmark

Single H200, colocated Qwen3-Omni-30B speech pipeline vs upstream main `b3bbff6e` (branch rebased onto it), SeedTTS-en voice clone, greedy decoding, byte-identical client, interleaved boots. 12 cells, 0 failures; generated audio seconds match within ±1.6% at every level.

| Conc | TTFT mean | TTFT p50 | TTFA mean | TTFA p50 | QPS |
|---|---|---|---|---|---|
| c1 | **−54.7%** | 171→74 ms (**−56.5%**) | **−31.2%** | 297→208 ms (−30.0%) | **+16.1%** |
| c8 | **−51.3%** | 276→120 ms (**−56.6%**) | −20.2% | 572→430 ms (−24.8%) | +2.9% (within noise) |
| c32 | **−46.1%** | 411→185 ms (**−55.1%**) | ~0% | +2.4% | ~0% |

Content gates for the CUDA-graph path (varlen flash attention vs the stock sdpa path agrees to bf16 kernel noise, not bitwise): WER (n=96) slightly better than the eager baseline; speaker-SIM within the eager-vs-eager control band.

Supersedes the encoder-wait half of #1564; the preprocessing-concurrency half of that PR is unaffected and will be reworked separately.

cc @charliechenye @Ratish1 @cr-gao


sglang_omni/models/qwen3_omni/components/audio_layer_graph.py
```diff
@@ -0,0 +1,242 @@
...
+        )
+
+    def _select(self, tokens: int, segments: int) -> int | None:
+        for bucket in self._token_buckets:
+            if bucket >= tokens and self._graphs.get(bucket) is not None:
+                if segments <= self._graphs[bucket].segment_slots:
+                    return bucket
+        return None
+
+    def maybe_replay(
```
AkazaAkane, line 228: he replay padding also needs to preserve the same max_seqlen <= window contract used during capture. For example, 130 real tokens select the 256 bucket, which currently creates one 126-token padding segment while max_seqlen=104.

Besides splitting the padding into window-sized segments, could we add a real replay regression test for bucket-boundary cases such as 127/128/129, 255/256/257, including a ragged batch? The current tests verify capture shapes, but not that padded replay produces the same result as eager varlen attention.

sglang_omni/scheduling/simple_scheduler.py
```diff
@@ -107,11 +107,15 @@ def _collect_batch(self, first_msg: IncomingMessage) -> list[IncomingMessage]:
             return batch
 
         batch_cost = self._message_cost(first_msg)
-        deadline = time.monotonic() + self._max_batch_wait_s
+        # Note (wenyao): an idle stage has nothing to coalesce with, so arming
+        # the window up front would spend all of it on this request's own TTFT.
+        deadline: float | None = None
```
AkazaAkane: Should this behavior be opt-in rather than changing `SimpleScheduler` globally?

I agree that the Qwen3-Omni audio encoder should not pay a 50 ms wait when it is idle. My concern is that this changes the batching contract for every stage using `SimpleScheduler`. For the audio encoder that tradeoff is probably desirable, but other ASR/TTS encoders may intentionally use `max_batch_wait_ms` to improve batch formation. 

Could we expose something like `batch_wait_when_idle=False` and enable it specifically for the Qwen3-Omni audio encoder?

sglang_omni/comm/stage_io.py
```diff
@@ -42,6 +43,13 @@
 _DIRECT_CUDA_IPC_PAYLOAD_TYPE = "TorchCudaIpcPayload"
 _DIRECT_CUDA_IPC_STREAM_INLINE_BYTES_LIMIT = 64 * 1024
 
+# Note (wenyao): direct CUDA IPC pays a per-request receive-side
+# cudaIpcOpenMemHandle that contends with prefill on the driver lock:
+# 13.4 ms mean / 69 ms p95 on a 258 KB hop vs 1.26 ms over the pooled
+# relay (pool opened once at startup). The floor keeps large image/video
+# payloads direct, where the handle cost amortises.
+_DIRECT_CUDA_IPC_PAYLOAD_MIN_BYTES = 4 * 1024 * 1024
```
AkazaAkane: Do we have the crossover measurements supporting 4 MB as a global transport threshold?

The 258 KB audio-encoder payload result clearly shows that pooled CUDA IPC is better for this hop, but this constant changes transport selection for every direct CUDA payload in the pipeline, including unrelated model/stage edges.

Before making 4 MB a global policy, I think we should either:

provide a payload-size sweep around the crossover, e.g. 256 KB / 512 KB / 1 MB / 2 MB / 4 MB / 8 MB, including p50/p95; or
scope this preference to the audio-encoder edge for now.

The optimization itself makes sense; I mainly want the scope of the policy to match the evidence.

AkazaAkane (commented):
The end-to-end TTFT improvement is strong, but since this PR combines three independent optimizations, could we add a small ablation showing their individual contributions?

In particular:

conditional encoder batch wait
shared cu_seqlens / removal of per-layer D2H sync
small-payload pooled transport
encoder layer CUDA graph

The CUDA-graph path is currently default-off, so it would also help to state explicitly whether the reported 171→74 ms / 276→120 ms numbers were measured with enable_layer_cuda_graph=true. If so, please also report the default-off result so we know what improvement users get immediately after merge.

jiannan-17:
<html><head></head><body><p>One minor, non-blocking suggestion: the PR table skips c16. I ran two rounds on an H100, and both showed a small TTFA regression at c16 with TTFT improvements. 

One possibility is that faster thinker prefill increases contention on the talker queue at this concurrency.</p><p>if possible, it would be useful to include an H200 c16 result for comparison.</p><div><div><div>
Conc | TTFT mean | TTFT p50 | TTFA mean | TTFA p50 | E2E latency mean | QPS
-- | -- | -- | -- | -- | -- | --
c1 | 199→124 ms (−37.8%) | 167→79 ms (−52.8%) | 375→299 ms (−20.3%) | 333→256 ms (−23.2%) | 0.639→0.565 s (−11.7%) | 1.57→1.78 (+13.4%)
c2 | 149→92 ms (−38.2%) | 133→87 ms (−34.9%) | 413→365 ms (−11.8%) | 405→358 ms (−11.6%) | 0.728→0.690 s (−5.2%) | 2.73→2.86 (+5.0%)
c8 | 250→153 ms (−38.7%) | 184→108 ms (−41.0%) | 936→903 ms (−3.5%) | 841→791 ms (−5.9%) | 1.520→1.494 s (−1.7%) | 5.19→5.30 (+2.0%)
c16 | 321→236 ms (−26.7%) | 166→113 ms (−31.6%) | 1.620→1.717 s (+6.0%) | 1.406→1.660 s (+18.1%) | 2.432→2.450 s (+0.8%) | 6.44→6.35 (−1.4%)
c32 | 611→283 ms (−53.7%) | 297→137 ms (−54.1%) | 3.498→3.351 s (−4.2%) | 2.913→2.841 s (−2.5%) | 4.375→4.249 s (−2.9%) | 7.06→7.27 (+2.9%)

</div></div></div><p> reproduction is on a single H100 80GB, comparing <code dir="ltr">main@b3bbff6e</code> with PR head <code dir="ltr">7ddba472</code>.</p><p>Setup: <code dir="ltr">Qwen/Qwen3-Omni-30B-A3B-Instruct</code>, <code dir="ltr">qwen3_omni_colocated_h100_bf16.yaml</code>, SeedTTS-en voice clone, streaming, greedy decoding (<code dir="ltr">--temperature 0.0</code>), and a byte-identical client using <code dir="ltr">benchmark_omni_seedtts.py --generate-only --voice-clone --stream</code>.</p>
thank you!</p></body></html>

jiannan-17 (commented):
One minor, non-blocking suggestion: the PR table skips c16. I ran two rounds on an H100, and both showed a small TTFA regression at c16 alongside the TTFT improvements.

one possibility is that faster thinker prefill increases contention on the talker queue at this concurrency.

if possible, it would be useful to include an H200 c16 result for comparison.

| Conc | TTFT mean | TTFT p50 | TTFA mean | TTFA p50 | E2E latency mean | QPS |
|---|---|---|---|---|---|---|
| c1 | 199→124 ms (−37.8%) | 167→79 ms (−52.8%) | 375→299 ms (−20.3%) | 333→256 ms (−23.2%) | 0.639→0.565 s (−11.7%) | 1.57→1.78 (+13.4%) |
| c2 | 149→92 ms (−38.2%) | 133→87 ms (−34.9%) | 413→365 ms (−11.8%) | 405→358 ms (−11.6%) | 0.728→0.690 s (−5.2%) | 2.73→2.86 (+5.0%) |
| c8 | 250→153 ms (−38.7%) | 184→108 ms (−41.0%) | 936→903 ms (−3.5%) | 841→791 ms (−5.9%) | 1.520→1.494 s (−1.7%) | 5.19→5.30 (+2.0%) |
| c16 | 321→236 ms (−26.7%) | 166→113 ms (−31.6%) | 1.620→1.717 s (+6.0%) | 1.406→1.660 s (+18.1%) | 2.432→2.450 s (+0.8%) | 6.44→6.35 (−1.4%) |
| c32 | 611→283 ms (−53.7%) | 297→137 ms (−54.1%) | 3.498→3.351 s (−4.2%) | 2.913→2.841 s (−2.5%) | 4.375→4.249 s (−2.9%) | 7.06→7.27 (+2.9%) |

Reproduction is on a single H100 80GB, comparing `main@b3bbff6e` with PR head `7ddba472`.

Setup: `Qwen/Qwen3-Omni-30B-A3B-Instruct`, `qwen3_omni_colocated_h100_bf16.yaml`, SeedTTS-en voice clone, streaming, greedy decoding (`--temperature 0.0`), and a byte-identical client using `benchmark_omni_seedtts.py --generate-only --voice-clone --stream`. Two interleaved boot rounds per side; numbers are the mean of both rounds. Zero failed requests across all cells; generated audio duration matches within ±4% at every level.

thank you!

edwingao28:
Addressed the review findings in three scoped commits:

- `38d826b9`: bound audio CUDA-graph replay padding to the attention window; added real capture/replay parity coverage for 127/128/129, 255/256/257, and ragged segments.
- `043c344e`: preserve the scheduler default and disable idle batch wait only for the Qwen3-Omni audio encoder.
- `d9f1d1fd`: remove the global 4 MiB transport heuristic and use pooled relay only for the audio encoder hop; unrelated stages retain direct CUDA IPC.

`enable_layer_cuda_graph` remains default-off.

### End-to-end benchmark

Paired/interleaved same-GPU cohorts, two cold boots per arm. Values are two-boot means; latency is milliseconds.

| GPU | Conc | Arm | TTFT mean | TTFT p50 | TTFA mean | TTFA p50 | E2E mean | QPS |
|---|---:|---|---:|---:|---:|---:|---:|---:|
| H100 | c1 | baseline | 146.2 | 123.4 | 298.7 | 276.3 | 531 | 1.891 |
| H100 | c1 | fixed, graph-off | 94.5 | 69.9 | 246.3 | 225.0 | 472 | 2.125 |
| H100 | c1 | fixed, graph-on | 53.3 | 52.9 | 205.1 | 200.0 | 440 | 2.275 |
| H100 | c8 | baseline | 254.5 | 248.0 | 745.8 | 716.7 | 1271 | 6.181 |
| H100 | c8 | fixed, graph-off | 145.3 | 120.3 | 651.1 | 630.1 | 1164 | 6.737 |
| H100 | c8 | fixed, graph-on | 97.4 | 95.9 | 580.0 | 553.0 | 1191 | 6.619 |
| H100 | c16 | baseline | 334.9 | 299.9 | 1151.2 | 1110.0 | 2114 | 7.453 |
| H100 | c16 | fixed, graph-off | 166.7 | 135.5 | 981.3 | 929.0 | 1969 | 8.006 |
| H100 | c16 | fixed, graph-on | 154.6 | 111.1 | 1018.6 | 968.5 | 2038 | 7.742 |
| H200 | c16 | baseline | 350.9 | 334.7 | 820.2 | 804.9 | 2014 | 7.770 |
| H200 | c16 | fixed, graph-off | 223.8 | 175.7 | 757.9 | 728.4 | 1998 | 7.930 |
| H200 | c16 | fixed, graph-on | 164.5 | 143.4 | 797.7 | 720.4 | 2105 | 7.500 |

Default-off vs baseline:

- H100 c1: TTFT -35.4%, TTFA -17.5%, E2E -11.1%, QPS +12.4%.
- H100 c8: TTFT -42.9%, TTFA -12.7%, E2E -8.4%, QPS +9.0%.
- H100 c16: TTFT -50.2%, TTFA -14.8%, E2E -6.8%, QPS +7.4%.
- H200 c16: TTFT -36.2%, TTFA -7.6%, E2E -0.8%, QPS +2.1%.

Graph-on vs default-off at H200 c16: TTFT -26.5%, but TTFA +5.3%, E2E +5.4%, and QPS -5.4%. This is why the graph path remains opt-in.

Two-boot c16 ranges:

- H100 graph-off: TTFT 143.4–190.0 ms, TTFA 936.3–1026.2 ms, QPS 7.968–8.044.
- H100 graph-on: TTFT 154.1–155.1 ms, TTFA 1013.2–1023.9 ms, QPS 7.672–7.812.
- H200 graph-off: TTFT 220.7–226.8 ms, TTFA 753.7–762.0 ms, QPS 7.907–7.953.
- H200 graph-on: TTFT 162.1–166.9 ms, TTFA 750.2–845.1 ms, QPS 7.473–7.527.

### Component measurements

These are component-level measurements, not independently isolated E2E runs:

- Conditional idle wait: removes a measured 50.26 ms p95 wait.
- Shared `cu_seqlens`: encoder forward 18.2/18.4/19.8 ms -> 16.3/16.0/18.2 ms at 1000/2000/3000 frames.
- 258 KB audio payload: direct IPC 13.4 ms mean / 69 ms p95 -> pooled relay 1.26 ms mean.
- Encoder layer graph: layer stack 11–13 ms -> 2.2–2.6 ms.

All formal cells had zero failed requests, 3–4 s mean audio duration, and byte-identical request texts across arms. Both graph-on boots captured buckets `[128, 256, 512, 1024, 2048, 4096]`. Focused GPU tests: 77 passed.

edwingao28:
Thanks @AkazaAkane and @jiannan-17 — I addressed the correctness and scoping issues, kept the audio layer CUDA graph default-off, and added paired two-boot H100/H200 ablation results above.

sglang_omni/models/qwen3_omni/components/audio_layer_graph.py
edwingao28: Fixed in 38d826b9: 
- replay padding is now split into window-bounded segments, with real CUDA eager/replay parity coverage for 127/128/129, 255/256/257, and ragged batches.

sglang_omni/scheduling/simple_scheduler.py
edwingao28: Addressed in 043c344e: 
- SimpleScheduler preserves its existing default, and batch_wait_when_idle=False is scoped only to the Qwen3-Omni audio encoder.

sglang_omni/comm/stage_io.py
edwingao28: Addressed in d9f1d1fd: the global 4 MiB heuristic was removed, pooled relay is scoped to the audio encoder hop, and unrelated stages retain direct CUDA IPC.

edwingao28:
Did a another round of benchmark, keep cuda graph opt on for the performance gain
benchmark table for cuda graph opt on/off

c | TTFT | TTFA | E2E | QPS
-- | -- | -- | -- | --
1 | −26.2% | −8.2% | +0.1% | −0.1%
4 | −19.6% | −6.1% | −1.1% | +0.8%
8 | −29.7% | −7.3% | −2.0% | +2.1%
16 | −28.0% | −6.5% | −2.3% | +2.1%
32 | −34.1% | −0.1% | −2.5% | +2.5%

sglang_omni/models/qwen3_omni/components/audio_encoder.py
```diff
@@ -65,21 +205,43 @@ def forward(
     ) -> dict[str, torch.Tensor]:
         if feature_attention_mask is not None:
             audio_feature_lengths = torch.sum(feature_attention_mask, dim=1)
-            input_features = (
-                input_features.permute(0, 2, 1)[feature_attention_mask.bool()]
-                .permute(1, 0)
-                .contiguous()
+            input_features = pack_padded_audio_features(
+                input_features.to(device=self._device, dtype=self.audio_tower.dtype),
```
AkazaAkane, line 209: Could we pack the padded features before moving them to CUDA?

The audio features arrive from the CPU-side preprocessing path. With a ragged batch, moving input_features first means we transfer batch_size × max_time frames over H2D and only discard the padding afterward. Previously the mask/gather happened before the device transfer, so H2D traffic was closer to sum(valid_lengths).

For example, lengths [500, 500, 500, 4000] would transfer roughly 4 × 4000 frames here instead of 5500.

AkazaAkane (approved):

sglang_omni/models/qwen3_omni/components/audio_encoder.py
edwingao28: I benchmarked this on H200: for [500, 500, 500, 4000], packing before H2D reduced transferred elements by 2.9× but increased pack+H2D p50 from 9.8 ms to 48.8 ms, so I’d prefer to keep the GPU-side packing.

PR #1634, opened 2026-08-20
[Whisper ASR] Separate encoder cache setup from decoder execution

wirybeaver:
## Summary
- precompute Whisper cross-attention K/V once after encoder execution and write it into SGLang's encoder KV cache
- make the decoder body consume cached cross-attention state while keeping token embedding outside the body
- preserve the current fused projection and default pre-LM encoder paths from `main`

This isolates the encoder/decoder execution-path change from #1530 so it can be benchmarked independently before evaluating breakable prefill CUDA Graphs.

## Test plan
- `pytest -q tests/unit_test/whisper_asr -m 'not benchmark'` (50 passed)
- `python3 -m py_compile sglang_omni/models/whisper_asr/sglang_model.py tests/unit_test/whisper_asr/test_decoder_cache.py`

## Benchmark plan
Benchmarking is intentionally pending. Use `openai/whisper-large-v3` to compare current `main` against this adopter path with prefill BCG disabled; then benchmark the stacked CUDA Graph PR separately.

## Related
- #1530
- #1396

### DIFF
--- a/sglang_omni/models/whisper_asr/sglang_model.py
+++ b/sglang_omni/models/whisper_asr/sglang_model.py
@@ -17,7 +17,9 @@ import torch.nn.functional as F
 from sglang.srt.layers.logits_processor import LogitsProcessor
 from sglang.srt.layers.quantization.base_config import QuantizationConfig
 from sglang.srt.layers.radix_attention import RadixAttention
+from sglang.srt.mem_cache.memory_pool import KVWriteLoc
 from sglang.srt.model_executor.forward_batch_info import ForwardBatch
+from sglang.srt.model_executor.forward_context import get_attn_backend
 from sglang.srt.model_loader.weight_utils import default_weight_loader
 from torch import nn
 from transformers import WhisperConfig
@@ -182,6 +184,7 @@ class WhisperSGLangSelfAttention(nn.Module):
         key = key.view(-1, self.num_heads, self.head_dim)
         value = value.view(-1, self.num_heads, self.head_dim)
         attn_output = self.attn(query, key, value, forward_batch)
+        attn_output = attn_output.reshape(hidden_states.shape[:-1] + (self.embed_dim,))
         return self.out_proj(attn_output)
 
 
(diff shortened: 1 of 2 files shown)


musclemuller:
## H100 benchmark results

I benchmarked base `10733c9875b7` against head `f1e5d148d212` on one H100 SXM 80GB using the digest-pinned CI image and a direct-container workflow.

Workload: `openai/whisper-large-v3`, pinned 1,088-sample SeedTTS English set, concurrency `1,2,4,8,16,32`, one warmup plus three measured repeats, and both A→B and B→A execution orders. Prefill CUDA Graph was disabled; encoder/decode CUDA Graphs remained enabled.

### Correctness

- 156,672/156,672 measured requests succeeded, with zero skips or request failures.
- All 13,056 concurrency-1 paired base/head outputs had identical raw and normalized transcripts.
- Under pre-LM disabled at concurrency 2, one sample reproducibly differed in both execution orders:
  - base: `he had tried in vain ...`
  - head: `he had tried and vain ...`
- Base WER was `1.369552%` in all six repeats. Head matched it in four repeats and increased to `1.377119%` in two repeats—one in each order.

This is a reproduced violation of the predeclared baseline-WER-envelope gate. Therefore, under the frozen benchmark contract, the overall qualification result is **failed**, despite the small magnitude of the difference.

### Performance

Equal-weight pooled mean across six measured repeats; deltas are head versus base:

| pre-LM mode | metric | c1 | c2 | c4 | c8 | c16 | c32 |
| disabled | throughput | +8.37% | +11.41% | +10.39% | +5.43% | +8.79% | +3.44% |
| disabled | P95 latency | -4.36% | -7.94% | -5.71% | +3.06% | -3.91% | +0.44% |
| enabled | throughput | +1.52% | +0.88% | +4.63% | +6.26% | +6.12% | +6.78% |
| enabled | P95 latency | -0.91% | -0.10% | -5.02% | -7.07% | -5.55% | -8.08% |

The disabled/c8 P95 result is **inconclusive** under the order-effect policy: A→B was `-0.18%`, while B→A was `+6.29%`; only one order crossed the `+5%` regression threshold. No throughput regression crossed the predeclared `-3%` threshold in both orders.

Peak GPU memory was effectively unchanged, and mean power remained approximately 243 W. GPU-process CPU metrics were unavailable because the provider container exposed NVML host PIDs outside the container PID namespace; system CPU and all GPU telemetry were retained.

This run did not qualify prefill BCG capture/replay and did not perform NCU profiling. The checksum-verified artifact manifest SHA-256 is `f59cfa92c09fe4e3620d70d7920757b8d205d411ee2d781916b1e5eeb92378f2`.

wirybeaver:
The PR #1530 will include this PR change. This PR shows that the adaptor change won't cause regression while enable_pre_lm_encoder = false / true. I will add @musclemuller as the co-author of #1530.

PR #1635, opened 2026-08-20
[Runtime] Native CUDA MPS for colocated GPU processes (off/on/auto)

JiaxinD:
## Motivation

Colocated stage processes on one GPU are a first class runtime topology, and process level replicas (#1307, #1175) will multiply them. CUDA MPS is what makes colocation share the GPU instead of time slicing it, but it only existed as the external `examples/mps_dp/launch.sh`, which cannot reach processes spawned inside one pipeline and leaves daemon lifecycle, attach verification, and cleanup to the operator. This PR makes the runtime own MPS natively, including same-GPU data parallelism across serve commands; the script remains only for CUDA IPC weight sharing.

## What it does

* `--mps {off,on,auto}` on `sgl-omni serve` (also `PipelineConfig.mps`). Default `off`; `off` touches nothing, guarded by tests.
* `auto` enables MPS per GPU hosting two or more single GPU, non TP CUDA processes of this pipeline. TP groups and GPU spanning processes are excluded.
* `on` is force enable: same eligibility as `auto`, but one eligible process is enough, and an MPS incapable platform (MIG, pre Volta, missing control binary) is a hard error instead of a warning. Its purpose is same-GPU DP across serve commands, where `auto` can never trigger because each serve alone has one GPU process.
* **One daemon per physical GPU, shared across serves.** MPS merges kernels only for clients of one server, so the daemon is keyed by device UUID: the first serve creates it, later serves join, the last to leave drains and quits it. Same-GPU DP is therefore N `serve --mps on` commands plus the router; no external launcher.
* Attach verification is a startup gate (a client that misses the pipe dir time slices silently; verification also accepts a CUDA context created in a descendant of the stage process). A watchdog fails the pipeline promptly if the daemon dies mid serving.
* Fail closed teardown, and dirty state is never repaired: startup follows three rules under one per-GPU flock. No state dir means create; a live daemon whose owners are all alive means join; anything else fails with the full picture (state dir, daemon and owner liveness, live client PIDs, exact cleanup commands) and leaves everything in place for the operator. The only process the runtime ever signals is the daemon it just spawned when its own startup fails. A normal shutdown leaves nothing behind, so the failure only ever appears after a crash.
* Preflight gates: globally exported `CUDA_MPS_PIPE_DIRECTORY` is rejected while `--mps` is enabled; missing control binary degrades `auto` and fails `on`; MIG or pre Volta likewise; combining with launch.sh `WEIGHT_SHARE=1` is rejected.

## Evidence

Unit: 46 tests under `tests/unit_test/mps/` (state machine, create/join/adopt/leave, recovery branches, activation predicate, spawn env seam, runner exit paths including cancellation, verified by mutation). `tests/unit_test/pipeline/` matches baseline. Merging this branch with #1175 (process replicas) produces one import conflict and a green combined unit suite, so replica instance stages flow through the same spawn/env/verify path.

H200 (`GPU-d46fccdb`, hyper00, container sglang 0.5.16), Higgs TTS, head SHA in this PR:

| Lifecycle scenario | Result |
|---|---|
| auto activation, attach gate, request | pass, HTTP 200 |
| two serves, `--mps on`: join + one shared server (4 clients) | pass, both ports 200 |
| first serve leaves | daemon stays, survivor serves 200 |
| last serve leaves | daemon quit, zero residue |
| daemon killed mid serving | watchdog fails pipeline in seconds, state cleaned |
| runtime SIGKILL, restart | startup fails naming the leftovers; after the printed cleanup, restart serves again |
| 12+ full up/down cycles over the campaign | zero residue after each |

Same GPU DP2 throughput, 2x Higgs (`--mem-fraction-static 0.35`), concurrency 32 per replica, 384 requests per replica, interleaved rounds on one H200 (background host load recorded per round):

| Pair (same card, back to back) | mode | qps | p50 | p95 | host load |
|---|---|---:|---:|---:|---:|
| GPU3 pair | off | 32.0 | 1.87 s | 3.47 s | 30.0 |
| GPU3 pair | on | 30.1 | 1.31 s | 3.30 s | 27.6 |
| GPU5 pair | off | 21.6 | 1.87 s | 3.26 s | 20.4 |
| GPU5 pair | on | 41.7 | 1.30 s | 2.79 s | 33.5 |
| single instance reference (GPU7, c=64) | off | 29.9 | 1.49 s | 3.74 s | 30.5 |

Every run completed 768/768 requests with zero errors and zero MPS residue. The p50 signal is consistent across both pairs: 1.87 s without MPS, 1.30 s with it. The qps signal is dominated by background host load on this shared box (the two pairs disagree in direction), so no throughput claim is made here; the published quiet-host study in `docs/basic_usage/mps_dp.md` (1.4x to 2.1x for saturated DP2/DP3) remains the throughput reference, and a quiet-window rerun is listed below.

`tests/test_ci/test_mps_native.py` codifies the lifecycle contract for CI; it is run manually for now and wiring it into a workflow follows the CI gate process separately.

## Remaining and non goals

* Default stays `off`; flipping `auto` needs its own evidence campaign after a release of soak.
* Quiet-window perf rerun (throughput at matched load, plus a `mps=off` vs baseline ABAB) is owed before undraft; the shared box was too contended tonight for throughput claims.
* CUDA IPC weight sharing stays on `examples/mps_dp/launch.sh`; combining is rejected.
* A GPU level replica x MPS smoke follows once #1175 merges (unit level compatibility is proven above).


sglang_omni/mps/runtime.py
```diff
@@ -0,0 +1,228 @@
...
+        return [name for name, gpu in self._client_gpu.items() if gpu == gpu_id]
+
+    def env_for_process(self, process_name: str) -> dict[str, str]:
+        gpu_id = self._client_gpu.get(process_name)
+        if gpu_id is None:
+            return {}
+        env = self.managers[gpu_id].env_for_stage()
+        # Note (Jiaxin Deng): the single-visible-device contract makes the
+        # worker normalize its local gpu_id to 0 under the UUID-scoped daemon.
+        env["SGLANG_ONE_VISIBLE_DEVICE_PER_PROCESS"] = "true"
```
JingwenGu0829: Higgs places CPU-only preprocessing (`gpu_id=None`) and GPU-backed audio_encoder (`gpu_id=0`) in the same tts_frontend process. 

MPS `stage_process_main` applies the marker independently to every stage spec. `_normalize_spec_gpu_id_to_local_device` therefore changes preprocessing.`gpu_id` from None to 0.
The CPU stage is subsequently passed through CUDA device binding and `gpu_startup_lock(0)`, and its Stage/CommRouter incorrectly records it as GPU-backed.

It's not that common in production path so marked out Higgs explicitly. Maybe consider `if spec.gpu_id is None: return` at the top of `_normalize_spec_gpu_id_to_local_device`  to gaurd explicitly for CPU specs?

sglang_omni/mps/manager.py
```diff
@@ -0,0 +1,403 @@
...
+        self.paths.owners_dir.mkdir(parents=True, exist_ok=True)
+        self._prune_dead_owners()
+
+        daemon_pid = self._manifest_daemon_pid()
+        alive = (
+            daemon_pid is not None
+            and self.client.pid_alive(daemon_pid)
+            and self.client.daemon_owns_pipe(daemon_pid, self.paths.pipe_dir)
+        )
+        if alive and self._owner_pids():
```
JingwenGu0829: After `_prune_dead_owners()` removes a dead owner, any surviving co-owner sends startup through this join branch, so the dead owner’s MPS clients are not reclaimed. 

For example, with shared serves A and B, SIGKILLing A and then starting C leaves A’s clients holding GPU memory while B remains; C may OOM, and the eventual last-owner drain times out on those clients. Since owner files are empty, the manager cannot safely distinguish A’s clients from B’s. 

Maybe persist per-owner stage/client PIDs with process start times and selectively reap dead owners. A regression may need to cover A+B, SIGKILL A, then start C while B remains healthy. It seems that the unit test currently is covering only parts of these scenarios.

tests/utils/tts_mps_runtime.py
```diff
@@ -18,13 +17,9 @@
 
 import requests
 
+from sglang_omni.mps import RUN_ID_PATTERN, MpsRunPaths, validate_control_socket
```
Ratish1: `derive_core_blocks` moved to `sglang_omni.mps.topology`, but these consumers still use its old test-utility location. The remote host suite reports three `AttributeError` failures, and DP2 collection fails with `ImportError`. Complete the migration: import it from `sglang_omni.mps` in `test_tts_mps_dp2.py`, and make the unit tests call and patch `sglang_omni.mps.topology` directly.

sglang_omni/mps/manager.py
```diff
@@ -0,0 +1,403 @@
...
+        # A stale control socket from a dead daemon confuses the new one.
+        shutil.rmtree(self.paths.pipe_dir, ignore_errors=True)
+        self.paths.pipe_dir.mkdir(parents=True, exist_ok=True)
+        try:
+            self.daemon_pid = self.client.start_daemon(
+                self.paths.pipe_dir, self.paths.log_dir, self.gpu_uuid
+            )
+            # Note (Jiaxin Deng): record ownership before waiting; a daemon
+            # that starts but never answers must stay findable and reapable.
+            self._write_manifest()
```
Ratish1: The daemon exists before this non-atomic manifest write. A kill, timeout, or torn write here leaves a responding daemon with no readable manifest, and the next start requires manual cleanup. Preclaim the namespace, publish the manifest atomically, and recover the native daemon PID file after validating its process identity.

sglang_omni/mps/runtime.py
```diff
@@ -0,0 +1,228 @@
...
+        return [
+            gpu_id for gpu_id, manager in self.managers.items() if not manager.probe()
+        ]
+
+    def stop(self) -> None:
+        errors: list[str] = []
+        for manager in self.managers.values():
+            try:
+                manager.stop()
+            except MpsError as exc:
```
Ratish1: `_query` can raise `subprocess.TimeoutExpired`, which is not an `MpsError` or `OSError`. It escapes this loop and `stop_best_effort`, so later GPU managers and runner coordinator/runtime-directory cleanup are skipped. Normalize control exceptions to `MpsError`, attempt every manager, then aggregate failures.

sglang_omni/mps/manager.py
```diff
@@ -0,0 +1,403 @@
...
+                if current is None or current <= 1:
+                    break
+                current = self.client.parent_of(current)
+        return False
+
+    # --- teardown: leave, and quit when last owner ---
+
+    def stop(self) -> None:
+        try:
+            with state_root_lock(self.paths.state_root):
```
Ratish1: One `.lock` is shared by every GPU under this root. Last-owner teardown holds it while waiting up to 60 seconds for clients, issuing control commands with a 10-second timeout, and waiting up to 10 seconds for MPS control-daemon exit, so GPU 0 teardown serializes GPU 1 lifecycle. Key the lock by GPU UUID; the waits can remain inside that per-GPU lock.

sglang_omni/mps/runtime.py
```diff
@@ -0,0 +1,228 @@
...
+class MpsDeviceInfo(Protocol):
+    def gpu_uuid(self, gpu_id: int) -> str: ...
+
+    def unsupported_reason(self, gpu_id: int) -> str | None: ...
+
+
+def _default_state_root() -> Path:
+    # Note (Jiaxin Deng): keep this short. The control socket lives under it
+    # and must fit the 107-byte AF_UNIX sun_path budget.
+    override = os.environ.get("SGLANG_OMNI_MPS_STATE_ROOT")
```
Ratish1, line 52: `SGLANG_OMNI_MPS_STATE_ROOT` determines both ownership state and `CUDA_MPS_PIPE_DIRECTORY`. Two serve commands targeting the same physical GPU with different values cannot discover or join each other's MPS control daemon, so the documented cross-serve sharing silently stops working. State that serves intended to share one GPU must use the same value.

sglang_omni/mps/devices.py
```diff
@@ -0,0 +1,63 @@
...
+            major, minor = pynvml.nvmlDeviceGetCudaComputeCapability(handle)
+            if (major, minor) < _MIN_COMPUTE_CAPABILITY:
+                return (
+                    f"compute capability {major}.{minor} is pre-Volta; "
+                    "per-client isolation requires Volta or newer"
+                )
+            try:
+                mig_current, _ = pynvml.nvmlDeviceGetMigMode(handle)
+                if mig_current == pynvml.NVML_DEVICE_MIG_ENABLE:
+                    return "MIG mode is enabled; MIG instances already isolate"
```
Ratish1: NVIDIA supports MPS within a MIG compute instance. If native MPS intentionally excludes unvalidated MIG deployments, report that as an SGLang Omni support policy instead of saying the GPU does not support MPS or that MIG removes process colocation.

sglang_omni/mps/manager.py
```diff
@@ -0,0 +1,403 @@
...
+    def __post_init__(self) -> None:
+        self.state = MpsState.IDLE
+        self.daemon_pid: int | None = None
+        self._registered = False
+
+    @property
+    def _owner_file(self) -> Path:
+        return self.paths.owners_dir / str(os.getpid())
+
+    # --- startup: create or join the shared daemon ---
```
Ratish1: remove these type of comments

sglang_omni/mps/runtime.py
JiaxinD: Good catch, fixed: the single-visible-device branch now returns early for gpu_id=None specs, so a colocated CPU stage keeps its identity. Test added.

sglang_omni/mps/manager.py
JiaxinD: Fixed: owner files record each serve's stage PIDs after attach verification, and a join reaps exactly the live clients no live owner claims. Test covers the A-killed, B-alive, C-joins case.

tests/utils/tts_mps_runtime.py
JiaxinD: Done: dp2 imports it from sglang_omni.mps and the unit tests call and patch sglang_omni.mps.topology directly; suite green on Linux.

sglang_omni/mps/manager.py
JiaxinD: Fixed: manifest is preclaimed before daemon spawn and written atomically via rename; a torn manifest recovers from the daemon PID file under the environ ownership proof.

sglang_omni/mps/runtime.py
JiaxinD: Fixed: manager teardown normalizes any control exception to MpsError and the orchestrator attempts every manager before aggregating. TimeoutExpired test added.

sglang_omni/mps/manager.py
JiaxinD: Done: the lock is keyed by GPU UUID now.

sglang_omni/mps/runtime.py
JiaxinD: Done: documented that serves sharing one GPU must use the same state root.

sglang_omni/mps/devices.py
JiaxinD: Done: reworded as an SGLang Omni support policy.

sglang_omni/mps/manager.py
JiaxinD: done

stephenkgli:
## Design simplification proposal

The recent changes—especially removing automatic orphan-client signaling and moving owner liveness to kernel-held `flock` leases—are the right direction. I recommend one more scope reduction before adding further recovery logic.

The core rule should be: **cleanly owned resources may be cleaned up; ambiguous shared state must be preserved and fail fast.**

### 1. Reduce startup to three cases

`MpsManager._start_locked()` should implement only:

1. No GPU state directory: create and verify a new daemon.
2. A provably healthy daemon and every published owner lease is currently held: join.
3. Anything else—including any dead lease, malformed state, unknown daemon identity, or leftover clients: preserve state and fail.

In particular, please remove dead-lease deletion and automatic adoption. An empty client snapshot does not prove safety: a dead owner may already have spawned a stage that has not initialized CUDA yet and can attach later. Treating every dead lease as dirty state removes this race and makes the contract much easier to reason about.

### 2. Make one acquisition token the source of truth

`MpsState`, `daemon_pid`, `_registered`, the owner file, and the lease file handle currently form overlapping state machines. Replace them with one explicit acquisition object, for example:

```python
@dataclass
class MpsLease:
    daemon_pid: int
    owner_fd: int
    attached_clients: set[MpsClientRef]
```

`MpsManager.acquire()` returns a lease; `verify()`, `probe()`, and `release()` require that lease. A manager that never acquired a lease has no cleanup authority. This removes the `stop-before-start` and `daemon_pid is None -> rmtree` branches and prevents rollback from deleting pre-existing dirty state.

### 3. Remove the custom manifest from the ownership protocol

The native `pipe/nvidia-cuda-mps-control.pid` file, combined with process identity, exact pipe-directory environment, and a responsive control socket, is already the daemon identity source. The custom manifest records `gpu_uuid`, `creator_pid`, and `pipe_dir`, but those fields are not used to prove ownership.

Use the native PID file strictly. If the PID file is missing while the socket responds, fail as ambiguous. This removes the preclaim, duplicate atomic writes, torn-manifest fallback, and another source of conflicting state.

### 4. Make multi-GPU startup transactional inside `MpsPipelineRuntime`

`MpsPipelineRuntime.start()` should record only successfully acquired managers and roll them back in reverse order on failure:

```python
acquired = []
try:
    for manager in managers:
        acquired.append(manager.acquire())
except BaseException:
    for lease in reversed(acquired):
        manager.release(lease)
    raise
```

`stop()` should process only acquired leases. This lets `MpsManager` assume valid ownership instead of defending against every partially initialized caller state.

### 5. Consolidate asynchronous lifecycle handling

All blocking MPS calls should use one retained, shielded `to_thread` future. Cancellation must wait for that worker to finish before rollback. Teardown should also be single-flight: one `_stop_task` represents cleanup completion, and every caller awaits the same task. `_started = False` should mean only “not serving,” not “cleanup complete.”

### 6. Keep client ownership runtime-local

Do not restore persistent owner-to-client JSON. Attach verification should return actual `(server_pid, client_pid)` references and store them in the current in-memory lease. They are needed for watchdog checks, detach verification before lease release, and NVIDIA-safe `terminate_client` before forced OS termination.

### 7. Narrow the control and device APIs

- Replace generic PID parsing and broad `Exception -> False` fallbacks with strict domain operations: `snapshot()`, `terminate_client()`, `quit_daemon()`, and `read_daemon_identity()`.
- Combine `unsupported_reason()` and `gpu_uuid()` into one physical-device inspection.
- Base eligibility on the final resolved physical process graph; reject TP, multi-GPU processes, explicit unnormalizable `cuda:N`, and cross-GPU CUDA edges in the native-MPS v1 path.

### Keep these protections

The per-GPU lifecycle flock, lifetime owner lease, AF_UNIX path-length validation, UUID-keyed private state, bounded control calls, descendant attach verification, watchdog, last-owner drain-before-quit, and external-MPS/weight-sharing conflict checks all correspond to real failure modes and should remain.

### Expected result

This should reduce `MpsManager` substantially and leave one understandable contract:

```text
acquire proven ownership -> spawn -> verify clients -> serve
-> verify own clients detached -> release lease -> last owner quits

any ownership ambiguity -> preserve evidence and fail
```

If automatic hard-kill recovery remains a product requirement, it should be a separate supervisor feature with `terminate_client`, pidfds/start identities, and process-group or cgroup ownership—not additional fallback branches in this manager.

sglang_omni/mps/runtime.py
```diff
@@ -0,0 +1,228 @@
...
+        return None
+    from sglang_omni.platforms import current_platform
+
+    if not current_platform.is_cuda_alike():
+        if mode == "on":
+            raise MpsError("mps=on requires an NVIDIA CUDA platform")
+        logger.warning("MPS auto: platform is not CUDA; running without MPS")
+        return None
+
+    if shutil.which("nvidia-cuda-mps-control") is None:
```
edwingao28: `auto` returns here before `MpsPipelineRuntime.create()` performs the global `CUDA_MPS_PIPE_DIRECTORY` rejection. On a host or container without the control binary, workers therefore inherit an external MPS pipe while the runtime reports that it is “running without MPS,” bypassing the advertised fail-closed preflight and its lifecycle monitoring. Check the inherited pipe directory before any `auto` degradation return.

Reproduction returned `None` while `/external/mps` remained exported.

tests/test_ci/test_mps_native.py
```diff
@@ -0,0 +1,256 @@
...
+    shutil.rmtree(STATE_ROOT, ignore_errors=True)
+    _wait_gpu_drained()
+    yield
+    shutil.rmtree(STATE_ROOT, ignore_errors=True)
+
+
+def _operator_cleanup() -> None:
+    # The dirty-state error names the live client PIDs; those are the serve's
+    # orphaned spawn children, so the operator kills them along with the tree.
+    subprocess.run(["pkill", "-9", "-f", "sglang_omni.cli serve"], check=False)
```
edwingao28: These two `pkill -f` calls are host-wide for the current UID, so this test can SIGKILL unrelated `sglang_omni.cli serve` instances and every unrelated multiprocessing-spawn child on a shared node. Give each invocation a unique state root, launch its serves in owned process groups, and kill only those groups plus daemon PIDs resolved from that root.

sglang_omni/mps/runtime.py
edwingao28: `auto` returns here before `MpsPipelineRuntime.create()` performs the global `CUDA_MPS_PIPE_DIRECTORY` rejection. On a host or container without the control binary, workers therefore inherit an external MPS pipe while the runtime reports that it is “running without MPS,” bypassing the advertised fail-closed preflight and its lifecycle monitoring. Check the inherited pipe directory before any `auto` degradation return.

Reproduction returned `None` while `/external/mps` remained exported.

tests/test_ci/test_mps_native.py
edwingao28: These two `pkill -f` calls are host-wide for the current UID, so this test can SIGKILL unrelated `sglang_omni.cli serve` instances and every unrelated multiprocessing-spawn child on a shared node. Give each invocation a unique state root, launch its serves in owned process groups, and kill only those groups plus daemon PIDs resolved from that root.

tests/test_ci/test_mps_native.py
```diff
@@ -0,0 +1,256 @@
...
+
+MODEL = os.environ.get("MPS_NATIVE_CI_MODEL", "bosonai/higgs-tts-3-4b")
+STATE_ROOT = Path(os.environ.get("MPS_NATIVE_CI_STATE_ROOT", "/tmp/mps-native-ci"))
+HEALTH_TRIES = 150
+HEALTH_INTERVAL = 5
+# Dedicated CI cards start empty; a shared dev box can raise this to tolerate
+# other tenants' resident memory while still catching our own leaks.
+DRAIN_BASE_MIB = int(os.environ.get("MPS_NATIVE_CI_DRAIN_BASE_MIB", "2000"))
+
+pytestmark = pytest.mark.skipif(
```
edwingao28: nit: this real GPU smoke is not marked `gpu` It is also currently discoverable by the default PR command, pytest tests/ -m "not benchmark", so gpu alone would not make it manual only.

sglang_omni/mps/control.py
```diff
@@ -0,0 +1,132 @@
...
+            pass
+
+    def parent_of(self, pid: int) -> int | None:
+        try:
+            stat_text = Path(f"/proc/{pid}/stat").read_text()
+            return int(stat_text.rsplit(")", 1)[1].split()[1])
+        except (OSError, IndexError, ValueError):
+            return None
+
+    def owner_lease_held(self, lease_file: Path) -> bool:
```
edwingao28: The new owner-liveness contract depends on the kernel flock probe here, but the unit tests replace it with numeric membership in `alive_pids`, and the only real MPS test is manual rather than CI-wired. Add a host-only subprocess test that holds the lease in another process and verifies `owner_lease_held()` changes from true to false when the holder exits.

sglang_omni/mps/__init__.py
```diff
@@ -0,0 +1,26 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Shared primitives for running pipeline replicas under CUDA MPS.
```
edwingao28: derive_core_blocks and MpsRunPaths are only used by tests and the test launcher; the native MPS runtime never calls them. Why exporting them here instead of keeping these helpers under tests/ folder?

tests/unit_test/mps/test_mps_manager.py
```diff
@@ -0,0 +1,417 @@
...
+        return self.parents.get(pid)
+
+    def daemon_owns_pipe(self, pid, pipe_dir):
+        return self.pid_pipe_dirs.get(pid) == str(pipe_dir)
+
+    def owner_lease_held(self, lease_file):
+        try:
+            return int(lease_file.name) in self.alive_pids
+        except ValueError:
+            return False
```
edwingao28: This fake reduces lease ownership to PID membership, so the unit tests never exercise the real flock behavior in SubprocessMpsControlClient.owner_lease_held()

sglang_omni/mps/topology.py
```diff
@@ -0,0 +1,99 @@
...
+    )
+    allowed = sorted(node_cpus & set(os.sched_getaffinity(0)))
+    minimum_cores = cores_per_block * n_blocks
+    if len(allowed) < minimum_cores:
+        raise RuntimeError(
+            f"need {minimum_cores} allowed CPU cores on the GPU NUMA node "
+            f"for {n_blocks} pinning blocks, found {len(allowed)}"
+        )
+    # Reserve a slice for host-side work, but never below the hard minimum, so
+    # a host that does have enough cores is not reported as insufficient.
```
edwingao28: nit: comment style

sglang_omni/mps/runtime.py
```diff
@@ -0,0 +1,228 @@
...
+        if mode == "on":
+            raise MpsError("mps=on but nvidia-cuda-mps-control is not on PATH")
+        logger.warning(
+            "MPS auto: nvidia-cuda-mps-control not found; running without MPS"
+        )
+        return None
```
edwingao28: The CUDA_MPS_PIPE_DIRECTORY check happens too late, the inherited pipe check should move before the auto fallback paths.

With --mps auto and no control binary, workers keep the inherited external pipe even though we log that MPS is disabled, so they can continue using a live external MPS daemon without runtime attach verification or monitoring.

reproduced wl CUDA_MPS_PIPE_DIRECTORY=/external/mps
result: None
parent_env: /external/mps
child_env: /external/mps

tests/test_ci/test_mps_native.py
```diff
@@ -0,0 +1,256 @@
...
+    shutil.rmtree(STATE_ROOT, ignore_errors=True)
+
+
+def _operator_cleanup() -> None:
+    # The dirty-state error names the live client PIDs; those are the serve's
+    # orphaned spawn children, so the operator kills them along with the tree.
+    subprocess.run(["pkill", "-9", "-f", "sglang_omni.cli serve"], check=False)
+    subprocess.run(
+        ["pkill", "-9", "-f", "multiprocessing.spawn import spawn_main"],
+        check=False,
```
edwingao28: These broad pkill -f calls can kill unrelated processes on a shared runner. Please terminate only the processes started by this test.

sglang_omni/mps/manager.py
```diff
@@ -0,0 +1,760 @@
...
+            pid
+            for pid, path in self._owner_files().items()
+            if path != self._owner_file
+        }
+
+        if remaining_owner_pids:
+            if (
+                snapshot
+                and clients_could_have_attached
+                and not lease.attachment_verified
```
JiaxinD, line 574: `attachment_verified` is latched at startup, so a client created after verify() bypasses this guard and the owner marker is deleted while that client is still attached. Measured on H200: both serves exit clean and still leave a live daemon plus a state dir no owner can clear.

sglang_omni/mps/manager.py
```diff
@@ -0,0 +1,760 @@
...
+        if daemon_pid != lease.daemon_pid:
+            return (
+                f"daemon identity changed from {lease.daemon_pid} "
+                f"to {daemon_pid}"
+            )
+        try:
+            snapshot = self.client.snapshot(self.paths.pipe_dir)
+        except MpsControlError as exc:
+            return f"client snapshot query failed: {exc}"
+        missing = lease.attached_clients - snapshot
```
JiaxinD: probe() requires every ref verify() ever saw, so a stage's CUDA child exiting normally reports failure and mp_runner turns that into `_fail_runtime`. Measured on H200 with the stage alive, still attached, and the daemon healthy.

sglang_omni/mps/manager.py
```diff
@@ -0,0 +1,760 @@
...
+        self._require_live_lease(lease)
+        expected = set(expected_pids)
+        missing = set(expected)
+        last_error: MpsControlError | None = None
+        deadline = time.monotonic() + self.verify_timeout
+        while True:
+            try:
+                snapshot = self.client.snapshot(self.paths.pipe_dir)
+                attached = self._clients_belonging_to(snapshot, expected)
+                lease.attached_clients.update(attached)
```
JiaxinD: Refs accumulate across polls while success is decided from the last poll only, so a ref seen once and already gone stays required forever. This is where the probe issue above is created.

sglang_omni/mps/manager.py
```diff
@@ -0,0 +1,760 @@
...
+                attached_roots = {
```
JiaxinD: Second independent /proc walk over the same snapshot. If a client exits or reparents between the two walks the lease is marked verified while owning nothing, which is the state the release guard above then trusts.

sglang_omni/pipeline/mp_runner.py
```diff
@@ -563,15 +628,41 @@ async def _monitor_children(self) -> None:
...
-            self._fatal_event.set()
-        await self.stop()
+        if self._mps is None:
+            if self._fatal_event is not None:
+                self._fatal_event.set()
+            await self.stop()
+            return
+
+        try:
+            await self.stop()
```
JiaxinD: With MPS the fatal event is set after stop() instead of before it, so a dead pipeline keeps its listener through the client drain and daemon quit, up to about 70s at default timeouts. `tests/unit_test/mps/test_mps_runner_wiring.py:582` asserts the opposite ordering and fails 3/3 on this head, so one of the two has to give.

sglang_omni/pipeline/mp_runner.py
```diff
@@ -620,13 +711,23 @@ async def stop(self) -> None:
             return_exceptions=True,
         )
 
+        mps_error: Exception | None = None
+        if self._mps is not None:
+            try:
+                await self._close_mps()
+            except Exception as exc:
```
JiaxinD: CancelledError is not an Exception and `_run_blocking` re-raises it by design, so a second Ctrl-C during the `finally` at serve/launcher.py:462 skips the completion task, `coordinator.stop()` and `_close_runtime_dir()`.

tests/unit_test/mps/test_mps_runner_wiring.py
```diff
@@ -0,0 +1,586 @@
...
+    runner = mp_runner.MultiProcessPipelineRunner(
+        _make_config(short_base, mps="off")
+    )
+    await runner.start()
+    assert group.spawn_env is None
+
+    group.dead = True
+    waiter = asyncio.create_task(runner.wait_failed())
+    await entered.wait()
+    assert not waiter.done()
```
JiaxinD: Fails 3/3 on a112c2f9. It asserts the MPS ordering, while the mps=off path it exercises keeps the merge base ordering and sets the fatal event before stop().

sglang_omni/pipeline/stage_workers.py
```diff
@@ -855,10 +886,14 @@ def _normalize_spec_gpu_id_to_local_device(spec: StageLaunchConfig) -> None:
     if spec.placement_gpu_id is None:
         spec.placement_gpu_id = spec.gpu_id
     spec.gpu_id = 0
-    if "gpu_id" in spec.factory_arg_defaults:
-        spec.factory_arg_defaults["gpu_id"] = 0
-    if "gpu_id" in spec.comm_config:
-        spec.comm_config["gpu_id"] = 0
+    for kwargs in (
+        spec.factory_kwargs,
+        spec.typed_kwargs,
```
JiaxinD, line 888: This helper is shared with the TP path, which already exports SGLANG_ONE_VISIBLE_DEVICE_PER_PROCESS, so widening it to factory_kwargs and typed_kwargs changes mps=off behavior too: a TP stage's `factory.gpu_id` 3 becomes 0, with no TP regression test.

tests/test_ci/test_mps_native.py
```diff
@@ -0,0 +1,350 @@
...
+    try:
+        proc.wait(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        proc.kill()
+        proc.wait(timeout=30)
+        raise AssertionError("serve did not exit on SIGTERM")
+
+
+def _assert_no_residue() -> None:
+    leftovers = (
```
JiaxinD, line 127: Under a container PID 1 that does not reap, each daemon cycle leaves two zombie mps pids that this check cannot see, because a zombie's environ is empty. 266 had accumulated in ours.

JiaxinD (commented):
Correctness pass on a112c2f9 against base afb5910e. Nine issues inline, three of them reproduced on an H200 with a real MPS daemon and real CUDA clients.

The ownership protocol passed everything it was given:

| check | result |
|---|---|
| second owner joins, last leaver quits | state dir clears |
| kill A, then start C while B serves | C blocked, B untouched |
| stale owner filename on a live unrelated pid | join still blocked |
| four owners racing on an empty root | one daemon |
| owner marker unlinked and recreated | lease invalidated |
| daemon killed under a verified lease | probe reports, release persists dirty |
| chaos, 5 rounds x 6 owners, random order | 0 violations, 0 residue |

Three hypotheses were checked and closed rather than left open:

| hypothesis | evidence |
|---|---|
| identity is read once with no retry after spawn | real `nvidia-cuda-mps-control -d` wrote its pid file before returning, 5 of 5, within 0.1ms of the call returning |
| an explicit `cuda:1` hidden in comm_config | `CommConfig` is `extra="forbid"` |
| auto should not hard fail on an inherited pipe | that ordering is what review asked for, now locked by a passing test |

Two caveats on scope. The watchdog issue below is confirmed as a mechanism but its frequency is not established: a live pipeline sampled once per second for 240s never changed its client set, though that pipeline is single process. Two real serve commands sharing a daemon on this head was not run at all, because the container ships sglang 0.5.16 against this head's 0.5.18 pin.

sglang_omni/mps/runtime.py
```diff
@@ -0,0 +1,625 @@
...
+        if mode == "on":
+            raise MpsDecisionError(
+                "mps=on but no process is eligible for MPS (TP and CPU-only "
+                "processes cannot attach)"
+            )
+        return {}
+
+    ordinals_by_process: dict[str, set[int]] = {
+        fact.process_name: set(fact.placement_gpu_ids)
+        | set(fact.explicit_cuda_gpu_ids)
```
charliechenye: Blocking: I think `cuda:0` is being interpreted in the wrong ordinal namespace here.

Native MPS later narrows the child to `CUDA_VISIBLE_DEVICES=<placement GPU UUID>`, so that physical GPU becomes local `cuda:0` inside the worker. Existing pipeline configs already use `FactoryArgs(device="cuda:0")`, and process replica expansion can move the same stage to GPU 1/2/etc. without rewriting that factory device string.

With the current union, a process placed on parent GPU 1 with `device="cuda:0"` is resolved as referencing physical GPU 1 plus physical GPU 0 and is rejected, even though after the intended single-device normalization both values refer to the placed GPU.

Could physical identity come from resolved placement, treat explicit `cuda:0` as compatible with the narrowed local device, and reject only nonzero explicit CUDA ordinals for this v1 path? A regression for placement GPU 1 + `device="cuda:0"` succeeding, while `cuda:1` still fails, would pin the distinction.

Thanks a lot

charliechenye: e.g. FishAudio S2-Pro uses FactoryArgs(device="cuda:0")

sglang_omni/pipeline/mp_runner.py
```diff
@@ -620,13 +711,23 @@ async def stop(self) -> None:
             return_exceptions=True,
         )
 
+        mps_error: Exception | None = None
+        if self._mps is not None:
+            try:
+                await self._close_mps()
```
charliechenye, line 736: Blocking: by the time we reach the MPS release here, `StageGroup.shutdown()` may already have SIGTERM'd and then SIGKILL'd a stuck MPS client.

That was fine as process local cleanup before this PR, but it is a different safety contract once multiple serves share one MPS server. NVIDIA's MPS Client Early Termination contract explicitly treats signal termination of a client with outstanding GPU work as unsafe because it can leave the MPS server and its other clients in an undefined state. The supported path is to terminate that client's CUDA contexts through `terminate_client <server_pid> <client_pid>` first; after that succeeds, the OS process can be killed without affecting the other MPS clients.

Could the MPS runtime participate before the forced kill step here: prove the exact currently owned client refs, issue `terminate_client` for them, then allow the existing process termination? If ownership cannot be established, I think we need to fail closed rather than risk a stuck serve A damaging healthy serve B on the shared daemon.

The control layer already represents `(server_pid, client_pid)` refs; it looks like the missing piece is making safe client termination part of the runtime shutdown protocol rather than only operator cleanup guidance.

charliechenye (changes requested):
Stepping back, I think the four client lifetime threads around `verify()`, `probe()`, and `release()` are manifestations of one ownership model issue rather than four independent bugs.

The current lease effectively treats the `(server_pid, client_pid)` refs observed during startup verification as stable lifetime owned resources. But those refs are dynamic observations: a CUDA descendant can attach after `verify()`, a previously observed descendant can exit normally, and parentage can change if a process exits or reparents.

That single assumption explains the four failures:

* `verify()` accumulates a ref seen in an earlier poll, even if it is gone by the successful poll.
* the second `/proc` ancestry walk can disagree with the first walk over the same MPS snapshot.
* `probe()` treats disappearance of any historically observed ref as an MPS failure, even when the owning stage is still healthy.
* `release()` trusts the latched startup ownership set, so a late client can survive after its owner marker is dropped.

Maybe let's fix the invariant once rather than patching these paths independently, shall we?: the stable source of truth should be the serve/worker ownership domain, while MPS client refs should be derived from a current snapshot when we need to verify attachment, health, safe forced termination, or release.

The exact stable identity could be process group/cgroup/supervisor owned state, but the important distinction is that an MPS client ref is an observation of that ownership domain, not the ownership identity itself. If current ownership cannot be proven after exit/reparenting, preserving dirty state still seems like the right fail-closed behavior.

This would also give the forced termination path one consistent place to derive the exact refs that may safely receive NVIDIA `terminate_client` before an OS level kill.

sglang_omni/mps/manager.py
stephenkgli: Fixed in [`23863398`](https://github.com/sgl-project/sglang-omni/commit/23863398). The lease now owns stable per-process runtime tokens and classifies a fresh MPS snapshot during release instead of trusting startup refs. A late CUDA descendant inherits the token, so it remains attributable to this owner and prevents the marker from being dropped. Regression coverage includes a late token-owned client.

sglang_omni/mps/manager.py
stephenkgli: Fixed in [`23863398`](https://github.com/sgl-project/sglang-omni/commit/23863398). `probe()` no longer requires every client ref observed during startup to remain present. It validates the daemon and classifies the current snapshot by the lease's runtime tokens, so normal descendant turnover does not fail an otherwise healthy pipeline.

sglang_omni/mps/manager.py
stephenkgli: Fixed in [`23863398`](https://github.com/sgl-project/sglang-omni/commit/23863398). Verification no longer accumulates historical `(server_pid, client_pid)` refs across polls. Each decision is made from the current snapshot using the process token inherited by its CUDA descendants.

sglang_omni/mps/manager.py
stephenkgli: Fixed in [`23863398`](https://github.com/sgl-project/sglang-omni/commit/23863398). The ancestry-based double walk was removed. A single current MPS snapshot is classified by reading each client's inherited runtime token, so verification and ownership use the same observation.

sglang_omni/pipeline/mp_runner.py
stephenkgli: Fixed in [`17d335be`](https://github.com/sgl-project/sglang-omni/commit/17d335be). The MPS watchdog now publishes the fatal event immediately; launcher teardown performs `stop()`, so the serving listener can be removed before the potentially long MPS drain/quit. The wiring test now locks this ordering.

sglang_omni/pipeline/mp_runner.py
stephenkgli: Fixed in [`17d335be`](https://github.com/sgl-project/sglang-omni/commit/17d335be). `stop()` captures `BaseException` from MPS close, completes completion-task cancellation, `coordinator.stop()`, and runtime-directory cleanup, then propagates the cancellation. A `CancelledError` regression covers the full cleanup order.

tests/unit_test/mps/test_mps_runner_wiring.py
stephenkgli: Fixed in [`17d335be`](https://github.com/sgl-project/sglang-omni/commit/17d335be). The test now exercises the actual MPS watchdog contract. The separate `mps=off` regression retains the merge-base ordering and asserts that no MPS close path is entered.

sglang_omni/pipeline/stage_workers.py
stephenkgli: Addressed in [`fd6bc9e6`](https://github.com/sgl-project/sglang-omni/commit/fd6bc9e6). `factory_kwargs` is no longer normalized because placement-owned `gpu_id` is rejected there. `typed_kwargs.gpu_id` remains normalized because it is the supported compiled factory path. A real `mps=off` TP compile regression verifies parent placement GPU 3 is preserved while the narrowed worker uses local GPU 0.

tests/test_ci/test_mps_native.py
stephenkgli: Fixed in [`2c6d0391`](https://github.com/sgl-project/sglang-omni/commit/2c6d0391). The native fixture now captures exact MPS process identities as `(pid, start_time)` while their scoped environment is readable, then verifies those identities disappear after cleanup. This detects zombies even after `/proc/<pid>/environ` becomes empty and reports the PID-1 reaping guidance.

sglang_omni/mps/runtime.py
stephenkgli: Fixed in [`7615abfe`](https://github.com/sgl-project/sglang-omni/commit/7615abfe). Physical identity now comes only from resolved placement. Explicit `cuda:0` is treated as the worker-local device after UUID narrowing, so placement GPU 1 + `device="cuda:0"` succeeds (including the typed and factory-kwargs paths). Nonzero explicit ordinals still fail before MPS acquisition with an actionable `mps=off` message.

sglang_omni/pipeline/mp_runner.py
stephenkgli: Thanks for catching this. We agree that force-terminating an MPS client with outstanding GPU work is not guaranteed to preserve a shared MPS server or its other clients.

We evaluated adding token-scoped `terminate_client` orchestration before the OS-level kill, but decided not to expand the shutdown protocol in this PR. The affected path requires both a worker failing graceful shutdown and another healthy serve sharing the same daemon; for v1 we are accepting this as a known operational limitation. If it occurs, operators must stop the affected workloads and clean/restart the shared MPS daemon and clients.

Normal graceful shutdown remains unchanged. We are intentionally leaving this thread unresolved because the forced-termination isolation guarantee is not implemented in this PR.

stephenkgli:
## Correctness Test Summary

**Tested commit:** `aa37d720`  
**Environment:** NVIDIA H200
**Overall result:** **Passed within the tested scope**

### Test Coverage and Results

| Area | Scenarios covered | Result |
|---|---|---|
| PR-focused unit tests | MPS configuration, lifecycle management, runtime wiring, device mapping, process spawning, pipeline compilation, and replica configuration | `179 passed, 4 deselected` |
| Broad CPU test suite | All non-benchmark and non-accelerator tests | `5401 passed, 3 skipped, 222 deselected` |
| Latest fix regressions | MPS cleanup helpers, TP device-count isolation, model factory hooks, and expanded replica launch configurations | `25 passed` |
| Native MPS hard-kill | Successful inference, hard-killed serve, dirty-state rejection, operator cleanup, restart, and successful inference after recovery | Passed |
| Shared MPS daemon | Two serves sharing one daemon, one serve exiting while the other remains healthy, and final cleanup | Passed |
| Process-replica concurrency | Two Qwen3-TTS replicas on one GPU, 12 concurrent requests, strict r0/r1 routing, and same-replica stage affinity | 12/12 HTTP 200 |
| Process-replica restart | Graceful shutdown followed by restart and two additional concurrent requests | 2/2 HTTP 200 |
| Data-plane behavior | Concurrent WAV and REST PCM requests, mixed-validity batch requests, client disconnect, and recovery requests | Passed |
| Failure handling | Killing one replica caused the launcher to fail fast and clean up the surviving replica and MPS resources | Passed |
| Output correctness | Successful outputs were compared byte-for-byte with a single-instance baseline | All PCM checksums matched |
| Resource cleanup | GPU processes, MPS clients/server, serve processes, and runtime state | No residue |

stephenkgli:
## Qwen3-TTS process-replica / CUDA MPS validation

Tested commit: aa37d720
Environment: NVIDIA H200

| Configuration | Successful requests | Throughput | Audio throughput | Mean / P95 latency | Mean RTF | Corpus WER |
| 1 process replica, MPS off | 1,088 / 1,088 | 5.635 req/s | 23.454 audio-s/s | 2.825 / 3.331 s | 0.7242 | 1.05% |
| 2 process replicas, MPS off | 1,088 / 1,088 | 7.206 req/s | 30.031 audio-s/s | 2.209 / 3.215 s | 0.5559 | 1.11% |
| 2 process replicas, MPS on | 1,088 / 1,088 | 9.115 req/s | 38.004 audio-s/s | 1.745 / 2.733 s | 0.4394 | 0.99% |

Performance observations:

- Two replicas without MPS improved request throughput by 27.9% over one replica and reduced mean latency by 21.8%.
- Enabling native CUDA MPS on the two-replica setup provided another 26.5% throughput improvement and reduced mean latency by another 21.0%.
- Overall, two replicas with MPS improved request throughput by 61.8% and audio throughput by 62.0% over the single-replica baseline, while reducing mean latency by 38.2%.

Correctness checks:

- WER was measured for all outputs with the same `Qwen/Qwen3-ASR-1.7B` service: all 1,088 samples were evaluated in every configuration, with zero skipped samples and no samples above 50% WER.
- All 3,264 generated WAV files were readable and had positive duration. Mean generated duration was 4.162, 4.167, and 4.169 seconds respectively, so there was no systematic output-length drift with MPS.
- In both two-replica runs, routing was exactly balanced: including warmups, `r0` and `r1` each handled 545 requests.

sglang_omni/pipeline/mp_runner.py
JiaxinD: Fixed in a239e987 rather than accepted: the stuck process now loses its CUDA contexts through `terminate_client` before any OS signal, scoped by the lease's process token so no other owner's clients are touched. Wired into both `stop()` and `_cleanup_on_failure()`.

Measured on one H200, two clients on one daemon under continuous GPU work, signal one of them:

| | rounds | survivor died |
|---|---:|---:|
| before, SIGTERM | 10 | 6 |
| before, SIGKILL | 10 | 7 |
| after, SIGTERM | 10 | 0 |

Note: (the trigger is wider than my earlier reply assumed) no SIGTERM handler is installed in the stage child, so `p.terminate()` at 30s is already an unclean termination and `p.kill()` is rarely reached; `_cleanup_on_failure()` signals with no grace window at all. The blast radius is a co-tenant crash, not operator cleanup, so this was not safe to defer.

stephenkgli (approved):
LGTM

PR #1250, opened 2026-08-01
[2/2] [Perf] Improve Ming-Omni-TTS serving: decoder graph

YzXiao101:
## Motivation

[PR #1005](https://github.com/sgl-project/sglang-omni/pull/1005) added Ming-Omni-TTS streaming and changed non-streaming requests to decode the complete latent sequence once.

The remaining streaming AudioVAE path repeatedly launches many short kernels while carrying request-local Qwen KV, upsampling, and ISTFT state between calls. Those dynamic tensors and Python cache objects prevent whole-decoder CUDA Graph capture and leave the H200 underused at concurrency 8.

This PR adds a fixed streaming transition and captures the whole AudioVAE decoder. It keeps the request API, M2/K4 delivery cadence, waveform continuity, and compact non-streaming decode unchanged.

### Scope

This PR adds:

- A fixed-capacity streaming state bank for Qwen KV, positions, upsampling lookahead, and ISTFT overlap.
- Fixed eager and CUDA Graph backends for the same streaming AudioVAE transition.
- Stable request-to-slot binding, abort cleanup, terminal release, and graph lifecycle handling in the Ming streaming vocoder scheduler.
- `streaming_cuda_graph` and `stream_slots` configuration for the audio-decode stage.
- A reusable ISTFT overlap-add primitive without changing the existing Ming-Omni decode path.
- Unit coverage for fixed-state math, slot lifecycle, scheduler transactions, graph contracts, configuration, and process topology.

### High-Level Design

```text
stream chunks -> M2/K4 cadence -> stable slot -> fixed AudioVAE transition
                                                | fixed eager
                                                ` CUDA Graph

non-streaming latent sequence ----------------> compact full decode
```

### Decision & Design

**Fixed streaming envelope.** Each active stream owns one persistent row in a `C`-slot state bank. The transition masks inactive rows and uses one bounded latent envelope; a terminal step can flush the pending and final groups in the same transition. The provided configuration uses `C=8`, matching its concurrency-8 workload. This keeps one graph and one state layout, but low-concurrency calls still pay for the fixed width. Capture-size buckets and a dispatcher can be added later if that overhead becomes material.

**Ming-TTS-local transition.** The implementation reuses the shared AudioVAE decoder weights and exposes only a narrow overlap-add primitive from the shared ISTFT. The fixed state orchestration remains inside Ming-TTS so this PR does not change Ming-Omni streaming behavior. This duplicates some decoder orchestration today; it can move behind a shared abstraction after the Ming path is validated independently.

[...]

PR #1637, opened 2026-08-20
[Perf][Ming-TTS] Prepare opt-in Prefill CUDA Graph adoption

musclemuller:
> [!IMPORTANT]
> This is a benchmark-pending Draft. It has no live GPU correctness evidence
> and makes no Ming-TTS support or performance claim. It must remain Draft
> until the planned correctness and performance qualification gates pass.

## Motivation

Ming-TTS currently composes request-specific continuous embeddings for AR
Prefill, then uses a custom eager forward that bypasses the shared Prefill CUDA
Graph lifecycle. This change prepares the narrow, opt-in adapter path for early
maintainer review before GPU qualification spending.

## Modifications

- Carry Ming-TTS's exact text-only, reference-conditioned, and Retraction
  Re-Prefill embedding windows through the existing private
  `OmniPrefillInputs` sidecar.
- Delegate graph admission, padding, capture, replay, sidecar cleanup, and
  eager execution to the standard SGLang worker.
- Advertise the existing breakable Prefill capability while keeping the
  feature disabled by default. Opt-in requires graph execution, the
  `breakable` Prefill backend, and explicit Prefill buckets; an unrealized
  declaration fails startup through the shared attestation contract.
- Preserve Continuous Feedback Row ordering, the A→B→A same-bucket isolation
  contract, TP ownership, acoustic-tail behavior, AR Decode behavior, and
  terminal streaming/non-streaming behavior.
- Preserve standard eager fallback for otherwise valid requests rejected by
  the shared bucket and 2x-padding admission policy. Invalid request or
  embedding state remains an explicit error.

This PR is independent of #1250. It does not change AudioVAE, ISTFT, Audio
Decode, the streaming vocoder implementation, AR Decode, or acoustic-tail
graph behavior.

## Related Issues

Related to #1357

## Accuracy Test

Local CPU contract coverage on the exact Draft candidate:

- Focused Ming-TTS and shared Prefill contracts: `86 passed`.
- Black, isort, Python compile checks, and `git diff --check`: passed.
- The broader Ming-TTS suite was attempted but collection is blocked locally
  by the unavailable optional dependency `onnxruntime`.
- The complete capability suite reached `31 passed, 4 failed`; the four
  unchanged launcher tests require the unavailable optional dependency
  `uvicorn`.

The focused run used a process-local `GenerationBatchResult` import-boundary
stub so unavailable scheduler extras were not imported; it did not replace
the production behavior under test.

No live CUDA Graph capture/replay, GPU correctness, WER, speaker-similarity,
or audio-quality evidence exists yet. Planned qualification is explicitly
limited to `1 x H200 NVL 141 GB, TP1` and will cover real replay, directed eager
fallback, Controlled-RNG Parity, and the full quality gate before this Draft
can be marked ready.

## Benchmark & Profiling

Not run yet. No latency, throughput, startup-time, or memory claim is made by
this Draft.

The planned qualification compares pre-adapter eager (A), adapter eager (B),
and adapter graph (C) at concurrency 1 and loaded concurrency, with at least
three measured repeats per candidate. Capture time, allocated/reserved memory,
Prefill timing, end-to-end QPS, audio-seconds/s, RTF, and streaming first-audio
time will be reported separately. A correct candidate without a repeatable
end-to-end win will close as a documented no-go rather than be presented as
supported.

## Checklist

- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed. No
      operator-facing support documentation is added while qualification is
      pending.
- [ ] Provide throughput / latency benchmark results and accuracy evaluation
      results as needed. Benchmark pending.
- [ ] For reviewers: If you haven't made any contributions to this PR and are
      only assisting with merging the main branch, please remove yourself as a
      co-author when merging the PR.

## CI

the label remains. Draft PRs are skipped even if labeled.


YzXiao101:
@musclemuller Thx for the contribution! Lmk when the PR is ready or if you need help testing it on H200.

musclemuller:
close this outdated pr for now, will try to update new ones in future

PR #1640, opened 2026-08-20
[Scheduler] Bind DP attention adapter to Omni model runner

junliu-mde:
## Motivation

Pinned SGLang 0.5.18 stores `SchedulerDPAttnAdapter` as a frozen dataclass. Omni may bind its custom model runner before or after upstream scheduler components are created. In both orders, delegated methods must use the Omni model runner instead of `tp_worker.model_runner`.

## Modifications

- Construct `SchedulerDPAttnAdapter` with the already-bound Omni model runner when available.
- Replace the pinned frozen adapter during late binding with `dataclasses.replace`.
- Preserve every other adapter field.

## Verification

Current-head verification on H100 with the repository-pinned environment:

```text
tests/unit_test/test_omni_scheduler.py::test_omni_scheduler_binds_one_execution_bridge_to_any_runner
tests/unit_test/test_omni_scheduler.py::test_bind_model_runner_replaces_frozen_dp_adapter
4 passed in 21.53s
```


JiaxinD (approved):
Verified at current head 79b45633: the DP attention adapter binding is frozen at model-runner initialization, and the exact-head unit-test plus accelerator suite are green. This is a focused scheduler correctness fix; approved.

JiaxinD:
@junliu-mde @StarDuster I want to acknowledge a process mistake on my side. While reviewing this PR, I used maintainer edit access to push directly to its branch without coordinating with you first. I should have prepared the update on my fork and offered it separately. Sorry about that.

The branch moved from `113f110d` to `79b45633`; the only addition from me was merge commit `79b45633` syncing current main. I did not force-push, so the original head remains in the history. I believe the sync is useful, but whether to keep it is your decision. I also approved the resulting revision, which was not an independent review boundary after I had modified the branch.

I will not push to this branch again. If you prefer the original branch state, please say so; I can prepare restoration instructions, but I will not change the branch without your explicit request.

junliu-mde:
Closing after reviewing the remaining diff against main at 836cdd77 and the pinned SGLang 0.5.18 interfaces.

This PR was a prerequisite of the async decode experiment in #1643. That experiment was removed and closed after the H100/H200 evaluation showed no positive end-to-end result and the current ScheduleBatch contract lacked a safe per-row penalty handoff.

The remaining change replaces the DP attention adapter's underlying SGLang model runner with the Omni runner. The adapter's prefill-graph path reads `prefill_cuda_graph_runner`, which the Omni ModelRunner/Qwen3TTSModelRunner does not expose. Current Omni keeps `require_mlp_sync` false, so passing CI does not establish that the replacement satisfies the adapter interface. The added tests check replacement and field preservation, not execution of that path.

There is no demonstrated independent serving fix left to justify this replacement. We will retain the current underlying-runner binding. This PR is being withdrawn, not treated as already merged or superseded by an equivalent binding change. Any future need should start from a concrete failing execution path and verify the actual adapter contract.

PR #1641, opened 2026-08-20
[Qwen3-TTS] Fuse no-compile Predictor kernels

junliu-mde:
## Motivation

Qwen3-TTS Predictor decode is launch-bound after the Talker compile path is
removed. The retained operations can run inside a bounded CUDA Graph while
preserving the fixed-seed output contract.

## Modifications

- Capture the per-token Predictor chain in bounded graphs keyed by batch bucket
  and sampling signature.
- Fuse exact seeded top-k/top-p sampling in the captured path.
- Fuse codec embedding gather/add inside graph capture.
- Use `torch.addmm` for the BF16, unquantized, TP1 H100 output-projection and
  residual path.
- Keep the eager PyTorch path for unsupported shapes, dtypes, devices, TP, and
  capture failures.
- Bind the bitonic parity implementation to the repository's `torch==2.13.0`
  pin and compare it directly with `torch.topk` in CUDA tests.
- Do not include the former KV-cache or first-token attention fusions. Both
  known-crash paths and their environment switches have been deleted.

The clean history separates the review boundaries: graph capture, seeded
sampling, output projection, sampling correctness fixes, and parity tests.

## Related Issues

Part of #1608. The KDA evaluation policy is tracked in #1650.

## Accuracy Test

The retained paths were previously exercised on H100 in 59 sampling tests and
72 Predictor/CUDA Graph tests. With the removed experimental fusions disabled,
the deterministic C4 and C8 closure matched original main exactly: 20/20 C4
WAVs and 40/40 C8 WAVs were byte-identical.

After rebuilding the branch on current main `3f819f9c`, the directly affected tests pass
on H100 with the pinned `torch==2.13.0`, `sglang==0.5.18`, and
`transformers==5.12.1` environment:

```text
121 passed in 96.53s
```

Repository GPU CI still requires a maintainer to add `run-ci` and
`run-qwen3-tts`.

## Benchmark & Profiling

The isolated seeded-sampling graph replay is 1.65x to 4.18x faster and uses 60
instead of 390 kernels. The latest retained integrated C4 profile records 109
Predictor graphs, 1,271 kernels per graph, and a 3.731537 ms median GPU-kernel
sum per graph.

These measurements isolate the Predictor path. This PR does not attribute the
full integrated end-to-end gain to one kernel.

## Checklist

- [x] Format the changed files with pre-commit.
- [x] Add graph, fallback, alias, endpoint, and seeded parity tests.
- [x] Delete known-crash experimental paths instead of shipping them disabled.
[...]

Issue #1608, opened 2026-08-18
[Tracker] KDA reproduction for Qwen3-TTS and MOSS-Transcribe-Diarize


junliu-mde:
## Current status (2026-08-23)

The integrated Qwen3-TTS branch is
[`junliu-mde/sglang-omni:kda/qwen3-tts-complete-0819`](https://github.com/junliu-mde/sglang-omni/tree/kda/qwen3-tts-complete-0819).

- Current tip: [`4bc6bcc1`](https://github.com/junliu-mde/sglang-omni/commit/4bc6bcc1)
- Original main baseline: `c6bbcc80163d912fcc9d1bff90003b5b21a66157`
- Measured rebased implementation: `57ab62adf96fcc9342a79dab21f4a87eec38add2`
- **Compile parity is closed:** the no-compile path matches or exceeds compile-on end to end under the standard fixed-length protocol. Talker `torch.compile` is removed for Qwen3-TTS; CUDA Graph remains enabled.

### Completed

- [x] Rebase the retained Qwen3-TTS stack onto main.
- [x] Preserve main's incremental streaming vocoder path.
- [x] Retain seeded sampling, Predictor kernels, and pinned token staging.
- [x] Preserve async token and penalty state across launch and resolve.
- [x] Run original main before and after the rebased implementation.
- [x] Run a fixed-length Talker compile-on/off A/B/A matrix.
- [x] Profile compile-on and compile-off with the same protocol.
- [x] Fully recover the Talker compile benefit: compile-off matches or exceeds compile-on within the 2% decision gate.
- [x] Remove the Qwen3-TTS Talker compile path after it showed no reproducible end-to-end gain.
- [x] Port the removal to current main in [#1649](https://github.com/sgl-project/sglang-omni/pull/1649).
- [x] Verify the current-main PR on H100: 204 directly affected tests passed.
- [x] Close the KV dual-write candidate with isolated, execution-path, and end-to-end evidence.
- [x] Keep async decode opt-in after crossed H100 and H200 results.
- [x] Replace finite WAV-hash membership with fixed-layout complete-signature Cn parity.
- [x] Make the two timing-unsafe Predictor fusions explicit opt-in paths.

## Final Cn correctness gate (2026-08-23)

The benchmark no longer accepts WAV hashes from a finite reference set. Cn
correctness now requires `enable_deterministic_inference: true`, which serializes
preprocessing and decodes each vocoder item at B1. The gate compares the complete
[...]

PR #1643, opened 2026-08-20
[Qwen3-TTS] Add async decode penalty handoff

junliu-mde:
## Motivation

Qwen3-TTS samples the semantic token before Predictor decode, but the normal
host collection path blocks the next launch. One-step lookahead can overlap that
collection only if staged tokens, Predictor feedback, and penalty state all keep
the same request ordering and sampling semantics as synchronous decode.

This is a stacked PR. It depends on Predictor PR #1641 and scheduler bridge PR
#1640. The current GitHub file view includes those dependencies. After they
merge, this PR narrows to the async decode and penalty handoff changes.

## Modifications

- Stage reporting tokens through pinned host buffers and a dedicated copy
  stream, with event ordering before reusable graph output is overwritten.
  Staging intentionally applies to both sync and async decode: it overlaps the
  semantic-ID copy with Predictor work, while the async switch isolates only
  the next-step lookahead.
- Split Qwen3-TTS decode into launch and resolve phases while retaining private
  device snapshots for the delayed resolve.
- Preserve staged host tokens instead of resolving from the GPU tensor again.
- Commit repetition, frequency, presence, and minimum-new-token penalty state on
  device before the next lookahead launch.
- Avoid applying the lookahead repetition token twice when the incremental mask
  update already contains that token.
- Keep logprob requests and batches below the configured threshold synchronous.
- Preserve async state through scheduler rebind and retraction boundaries.
- Keep async decode opt-in. The H100 and H200 results do not support default-on.
- Gate Cn parity with the complete WAV signature under deterministic inference,
  which fixes the vocoder at per-item B1 decode.

## Related Issues

Part of #1608. Coordinates with streaming work in #1523 but does not modify the
streaming vocoder implementation.

## Accuracy Test

The latest integrated closure passes on H100:

```text
73 focused benchmark, Predictor graph, and async-default tests passed
```

Correctness uses one warm-up plus five measured runs, seed `20260819`,
repetition penalty `1.05`, exactly 64 generated tokens, and
`enable_deterministic_inference: true`. This mode serializes preprocessing and
decodes each vocoder item at B1, so the vocoder layout cannot change with
request timing.

Original main and the rebased branch each produced one stable complete WAV
signature at C4 and C8. The signatures matched exactly: 20/20 C4 WAVs and
40/40 C8 WAVs were byte-identical to original main. The gate rejects same-arm
drift, conflicting references, and the former schema-3 finite-membership
reports. Full protocol and result are recorded in #1608.

The fixed-length Talker compile A/B/A passes its output checks but shows no
reproducible end-to-end benefit. The compile path is removed separately in
#1649; the full decision and profile are recorded in #1608.

## Benchmark & Profiling

Crossed five-round sync/async tests discard round one and report the median of
round medians from rounds two through five:

| Workload | Sync | Async | Async change |
| --- | ---: | ---: | ---: |
| Short C4 | 1.128452 s | 1.173869 s | +4.0% |
| Short C8 | 1.402605 s | 1.516409 s | +8.1% |
| Long C4 | 2.029622 s | 2.050231 s | +1.0% |
| Long C8 | 2.373556 s | 2.382903 s | +0.4% |

Async remains opt-in. The crossed H100 and H200 results do not support a
default-on policy.

In the latest short-C4 profile, sync/async request wall time is
1.534892/1.669495 s, while GPU kernel time is 677.956/679.668 ms. Async increases
`cudaStreamSynchronize` from 207 to 508 calls. Full protocol, event spans,
Predictor attribution, and PCM streaming TTFB are in #1608.

## Checklist

- [x] Add unit tests.

## CI


junliu-mde:
Follow-up closure for the async decode investigation:

### Why async does not profit

The overlap path is active, but Qwen3-TTS decode remains host/launch bound:

- Decode step: `10.294 ms`
- GPU busy time: `5.545 ms` (`53.94%`)
- Per-step `cudaStreamSynchronize`: `3.813 ms`

Most Python/CUDA launch work is outside the small host-result-resolution interval that async can overlap. By the time the launch phase finishes, there is not enough useful host work left to hide. Async also launches one final decode step whose result is discarded.

### Launch-reduction attempt

I fused the semantic sampling path into one Triton kernel to test whether reducing CUDA launches would make async profitable.

Correctness passed:

- 7,000 sampled tokens matched the SGLang reference across B1/B4/B8, varied seeds, positions, temperatures, and top-k values.
- All controlled service runs produced the same complete WAV signature.

The isolated sampler improved from about `348 us/step` to `34 us/step`, about `9.2x`. The absolute saving was only about `0.31 ms` in a `10.294 ms` decode step.

### End-to-end sampler A/B

Both arms used async decode, Talker compile off, 50 repetitions, fixed vocoder layout, and complete WAV-signature parity.

| Workload | Generic sampler | Fused sampler | Fused change |
| --- | ---: | ---: | ---: |
| C4 | 0.796945 s | 0.805004 s | 1.00% slower |
| C8 | 0.998148 s | 1.020956 s | 2.23% slower |

Same-arm drift was `11.26%/0.75%` for generic/fused at C4 and `0.29%/9.79%` at C8. The end-to-end difference is not a reproducible gain.

### Async on/off after launch reduction

| Workload | Async off | Async on | Async change |
| --- | ---: | ---: | ---: |
| C4 | 0.755092 s | 0.805004 s | 6.20% slower |
| C8 | 0.969631 s | 1.020956 s | 5.03% slower |

C4 same-arm drift was only `0.20%` with async off and `0.75%` with async on, so the C4 regression is clear. C8 had about `9%` same-arm drift, but it also provides no evidence of profit.

### Decision

Reducing semantic sampling launches does not make async decode profitable. The fused candidate was removed in full. The post-removal focused suite passes `254 tests`, and Ruff plus diff checks pass.

This strengthens the current PR policy: keep async decode opt-in and do not enable it by default.

junliu-mde:
Closing this PR after removing the experiment. The crossed H100/H200 measurements show no positive end-to-end result, and the current SGLang 0.5.18 ScheduleBatch contract has no safe per-row penalty handoff. The previous implementation also depended on scanning and restoring penalizer internals. The branch now has the same tree as current main, so there is no code left to merge.

PR #1646, opened 2026-08-21
[Whisper ASR] Use FlashInfer decoder LayerNorm

Dayuxiaoshui:
## Motivation

Whisper Large V3 decoding repeatedly applies LayerNorm before decoder self-attention, cross-attention, and the FFN, followed by one final decoder LayerNorm. A 32-layer decoder therefore executes 97 LayerNorm operations per generated token when cross-attention is active, making these small, launch-sensitive kernels measurable in the autoregressive path.

This PR uses FlashInfer's FP16 CUDA LayerNorm kernel for compatible decoder inputs while preserving PyTorch LayerNorm as the fallback for every unsupported device, dtype, parameter layout, or installation. The optimization changes only how LayerNorm is executed; it does not change the Whisper architecture, attention behavior, KV cache, sampling, or checkpoint format.

## Modifications

- Add `WhisperDecoderLayerNorm`, a narrow `torch.nn.LayerNorm` subclass that dispatches to `flashinfer.norm.layernorm` only when FlashInfer is available, the input is a CUDA FP16 tensor, affine weight and bias are present, and both affine parameters match the input dtype.
- Retain `torch.nn.LayerNorm.forward` as the universal fallback, including CPU, non-FP16, missing-FlashInfer, non-affine, and mixed-parameter-dtype execution.
- Apply the optimized module only to decoder self-attention, cross-attention, FFN, and final LayerNorm sites. The encoder remains unchanged so the optimization does not expand into the fixed-length encoder CUDA Graph path.
- Add unit tests that verify decoder-only placement and numerical equivalence of the CPU fallback with PyTorch LayerNorm.

For a hidden vector $x \in \mathbb{R}^{d}$, both implementations compute the same transformation:

$$
\mu = \frac{1}{d} \sum_{i=1}^{d} x_i, \qquad \sigma^2 = \frac{1}{d} \sum_{i=1}^{d}(x_i - \mu)^2
$$

$$
y_i = \gamma_i \frac{x_i - \mu}{\sqrt{\sigma^2 + \epsilon}} + \beta_i.
$$

The parameters $\gamma$, $\beta$, and $\epsilon$ are the existing checkpoint weight, bias, and LayerNorm epsilon. FlashInfer replaces the CUDA kernel implementation only; the model-level function and parameterization remain unchanged.

## Related Issues

Related to #1396 and the Whisper Large V3 throughput investigation in #1589.

## Accuracy Test

Setup: `openai/whisper-large-v3` at revision `06f233fe06e710322aca913c1bc4249a0d71fce1`, full SeedTTS English set with 1,088 clips at revision `27f4c1adee83b5b29b7c4b375f6b976324bda308`, one NVIDIA H200, TP=1, FP16, generation and encoder CUDA Graphs enabled, `torch.compile` enabled, `max_running_requests=32`, and `mem_fraction_static=0.30`. Each concurrency used one discarded warmup followed by three measured repeats.

| Concurrency | Evaluated requests | Baseline corpus WER | Candidate corpus WER |
|---:|---:|---:|---:|
| 1 | 3,264 / 3,264 | 0.0138 | 0.0137 |
| 8 | 3,264 / 3,264 | 0.0137 | 0.0137 |
| 16 | 3,264 / 3,264 | 0.0137-0.0138 | 0.0137-0.0138 |
| 32 | 3,264 / 3,264 | 0.0137-0.0138 | 0.0138 |

All 13,056 measured candidate requests completed successfully. Candidate corpus WER remained within the baseline range at every measured concurrency.

Unit and static validation:

- Whisper unit tests: `46 passed`.
- Black, isort, Python bytecode compilation, and `git diff --check`: passed.
- CPU fallback output matches `torch.nn.functional.layer_norm` under `torch.testing.assert_close`.
- The H200 run completed `torch.compile`, decode CUDA Graph capture, and replay through the FlashInfer path.

## Benchmark & Profiling

The benchmark used the same setup as the accuracy test. Throughput and request latency P95 are means across three full-set measured repeats.

| Concurrency | Baseline throughput (req/s) | Candidate throughput (req/s) | Throughput delta | Baseline P95 (s) | Candidate P95 (s) | P95 delta |
| 1 | 12.38 | 12.51 | +1.0% | 0.099 | 0.098 | -1.0% |
| 8 | 44.15 | 46.39 | +5.1% | 0.243 | 0.233 | -4.1% |
| 16 | 59.55 | 62.35 | +4.7% | 0.359 | 0.343 | -4.5% |
| 32 | 70.71 | 74.07 | +4.8% | 0.622 | 0.590 | -5.1% |

The result is effectively neutral at concurrency 1 and consistently improves middle-to-high concurrency throughput by approximately 5%, with a corresponding 4-5% reduction in P95 latency. This is consistent with replacing repeated launch-sensitive decoder normalization kernels without reducing model FLOPs or changing scheduling policy.

Benchmark artifact: `/tmp/whisper_layernorm_full.json`. During this candidate run, physical GPU 4 had external allocated memory but reported 0% external compute utilization; the remaining KV capacity was sufficient for `max_running_requests=32`. The full accuracy and completion gates passed, but final publication numbers should be reconfirmed on an exclusively allocated GPU runner.

## Checklist

- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed. No user-facing configuration or API changed, so no cookbook update is required.

## CI


sglang_omni/models/whisper_asr/sglang_model.py
```diff
@@ -27,12 +28,34 @@
...
 logger = logging.getLogger(__name__)
 
 _QKV_SHARDS = {"q_proj": 0, "k_proj": 1, "v_proj": 2}
 _KV_SHARDS = {"k_proj": 0, "v_proj": 1}
 
 
+class WhisperDecoderLayerNorm(nn.LayerNorm):
+    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
+        if (
+            flashinfer_layer_norm is not None
```
SandyLuXY, line 46: sglang-omni sets `FLASHINFER_USE_CUDA_NORM=1` on SM100+, and FlashInfer 0.6.14's CUDA backend rejects FP16 input, so the default Whisper FP16 stage fails on its first decoder norm. Gate this branch on a backend that supports the exact input/affine dtypes and fall back to `super().forward()` otherwise

tests/unit_test/whisper_asr/test_decoder_layer_norm.py
SandyLuXY, line 1: update tests readme

SandyLuXY (approved):

PR #1649, opened 2026-08-21
[Qwen3-TTS] Remove unprofitable Talker compile path

junliu-mde:
## Motivation

Qwen3-TTS privately compiles each Talker decoder layer while CUDA Graph remains
enabled. A fixed-length H100 A/B/A comparison shows no reproducible end-to-end
benefit from this path. The compile path also adds launch overhead and produces
unstable kernel selection across fresh caches.

## Modifications

- Remove the private Talker decoder-layer `torch.compile` path.
- Keep Qwen3-TTS CUDA Graph enabled.
- Mark Qwen3-TTS as not supporting SGLang `torch.compile`.
- Reject compile-on CLI arguments before server launch.
- Continue to accept explicit compile-off and shared compile max-batch settings; the latter remains part of the common SGLang batch policy but does not enable model compilation.
- Keep direct access to `self.layers` in the decode hot path.
- Add capability, engine, CLI, and launch-boundary tests.

## Related Issues

Part of #1608.

## Accuracy Test

Current-head targeted verification passes on H100 against current main:

```text
15 Qwen3-TTS compile-path tests passed
35 model-capability tests passed
```

The earlier broader H100 run passed 204 directly affected tests.

The compile A/B/A arms all returned exactly 64 tokens,
`finish_reason=length`, and a 245,804-byte WAV. The compile-removal ABBA check
produced one exact C1 WAV hash across all arms.

## Benchmark & Profiling

Protocol: one H100 80 GB; synchronous decode; CUDA Graph on; fixed seed and
input; exactly 64 generated tokens; 30 measured requests per arm; order
`compile-on A -> compile-off -> compile-on B`.

| Arm | C1 wall | C1 engine | C4 wall | C8 wall |
| compile-on A | 0.566786 s | 0.494674 s | 0.835568 s | 0.935240 s |
| compile-off | 0.559900 s | 0.493761 s | 0.749759 s | 0.901976 s |
| compile-on B | 0.564591 s | 0.494933 s | 0.755342 s | 0.841942 s |

Relative to the midpoint of the two compile-on medians, compile-off is 1.02%
faster for C1 wall, 0.21% faster for C1 engine, and 5.74% faster for C4. It is
1.51% slower for C8, below the 2% decision gate. The two compile-on C8 arms
differ by 10.5%.

Per decode step, compile reduces Talker GPU time from 1.789084 ms to 1.665111
ms, but increases Talker kernels from 367 to 423. It also increases median
`cudaGraphLaunch` CPU time from about 1.84 ms to 3.0788 ms and the
Predictor-to-Talker launch gap from 3.62-3.66 ms to 5.7243 ms. The local GPU
saving does not become an end-to-end saving.

The final removal was compared with the previous compile-off code in ABBA
order, with 30 C1 and 30 controlled-C4 measurements per arm:

| Metric | Previous compile-off | Compile removed | Change |
| --- | ---: | ---: | ---: |
| C1 wall midpoint | 0.565655 s | 0.566197 s | +0.10% |
| C1 engine midpoint | 0.495170 s | 0.498873 s | +0.75% |
| Controlled C4 wall midpoint | 0.751217 s | 0.736219 s | -2.00% |

All changes are within the 2% regression gate. The before/after compile-off
profiles have identical Predictor and Talker kernel counts and identical kernel
sequence hashes. Their GPU sums differ by less than 0.2%.

## Checklist

- [x] Format the changed files with pre-commit.
- [x] Add unit tests.
- [x] Provide latency and profiling evidence.
- [ ] For reviewers: If you have not contributed to this PR and only assist
  with merging main, remove yourself as a co-author when merging.

## CI

GPU CI requires a maintainer to add the `run-ci` label.

### DIFF
--- a/sglang_omni/models/qwen3_tts/config.py
+++ b/sglang_omni/models/qwen3_tts/config.py
@@ -48,8 +48,8 @@ class Qwen3TTSPipelineConfig(PipelineConfig):
 
     model_path: str
     # note (0xtoward): Keep deterministic inference opt-in because it serializes
-    # preprocessing and vocoder decoding and disables Talker compilation and the
-    # vocoder CUDA graphs, reducing throughput.
+    # preprocessing and vocoder decoding and disables the vocoder CUDA graphs,
+    # reducing throughput.
     enable_deterministic_inference: bool = False
     stages: list[StageConfig] = [
         StageConfig(
--- a/docs/cookbook/qwen3_tts.md
+++ b/docs/cookbook/qwen3_tts.md
@@ -91,8 +91,8 @@ enable_deterministic_inference: true
 
 When enabled, the same prompt, reference audio, and seed produce byte-identical
 PCM across runtime batch sizes. This mode reduces throughput because it
-serializes reference preprocessing and vocoder decoding and disables Talker
-compilation and the initial vocoder CUDA Graph, so it is disabled by default.
+serializes reference preprocessing and vocoder decoding and disables both the
+initial and follow-up vocoder CUDA Graphs, so it is disabled by default.
 
 ### Overload / admission policy
 
--- a/sglang_omni/models/qwen3_tts/__init__.py
+++ b/sglang_omni/models/qwen3_tts/__init__.py
@@ -10,7 +10,7 @@ CAPABILITIES = ModelCapabilities(
     supports_batch_vocoder=True,
     supports_streaming_vocoder=True,
     supports_cuda_graph=True,
-    supports_torch_compile=True,
+    supports_torch_compile=False,
     supports_breakable_prefill_cuda_graph=False,
 )
 
--- a/sglang_omni/models/qwen3_tts/engine_builder.py
+++ b/sglang_omni/models/qwen3_tts/engine_builder.py
@@ -8,9 +8,17 @@ from typing import Any
 
 from sglang_omni.models.qwen3_tts import request_builders
 from sglang_omni.models.qwen3_tts import stages as qwen3_stages
-from sglang_omni.platforms import current_platform
 from sglang_omni.scheduling.engine_factory import TtsEngineBuilder
-from sglang_omni.vendor.sglang.server_args import override_server_args
+
+
+def _is_truthy(value: Any) -> bool:
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, int):
+        return value != 0
+    if isinstance(value, str):
+        return value.strip().lower() in {"1", "true", "yes", "y", "on"}
+    return False
 
 
 class Qwen3TtsEngineBuilder(TtsEngineBuilder):
@@ -49,7 +57,7 @@ class Qwen3TtsEngineBuilder(TtsEngineBuilder):
             "dtype": dtype,
             "disable_cuda_graph": False,
             "disable_overlap_schedule": True,
-            "enable_torch_compile": True,
+            "enable_torch_compile": False,
             "mem_fraction_static": 0.85,
             "max_prefill_tokens": 8192,
             "sampling_backend": "pytorch",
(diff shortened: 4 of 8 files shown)


charliechenye:
Thanks for the detailed H100 benchmark and profiling evidence. I also ran an independent H200 validation, focusing primarily on the final removal itself.

For the PR-safety comparison I used:

- B: PR base `07b63c0bfa6cea55ff4e0fa9837d749f64a0d785`, with the private Talker compile path explicitly disabled
- C: PR head `b0a5d20c99c9bb8eb8ed37fb94a1b8dfa8b3d86a`, where the private compile path is removed

I think B -> C is the most relevant comparison for this PR because it separates "turn compile off" from "delete the already-disabled compile machinery."

### H200 protocol

The controlled workload used:

- `Qwen/Qwen3-TTS-12Hz-0.6B-Base`
- fixed reference audio and fixed input text
- fixed seed and sampling parameters
- exactly 64 generated tokens for every measured request
- non-streaming WAV output
- Talker decode CUDA Graph enabled in both B and C
- fresh server and fresh Inductor/Triton cache for every arm
- Predictor CUDA Graph buckets preconditioned before measurement
- repository benchmark clients isolated from CUDA
- physical GPU ownership checked before and during each measured arm

The primary sweep used 4 balanced fresh-server B/C pairs per concurrency. c1 and c4 were additionally repeated with 8 balanced fresh-server pairs because the first sweep showed WARN signals at those two points.

Positive percentages below mean C is better than B.

| Concurrency | B -> C QPS | Mean latency | p95 | p99 | Engine mean | Result |
| ---: | ---: | ---: | ---: | ---: | ---: | --- |
| c1 | +0.69% | +0.74% | +0.65% | +1.96% | +0.62% | PASS |
| c2 | +1.53% | +1.43% | +0.52% | +17.39% | +1.59% | PASS |
| c4 | +0.67% | +0.58% | -1.23% | +1.15% | +0.51% | PASS |
| c8 | +0.05% | -0.64% | +0.62% | +0.53% | +1.38% | PASS |
| c16 | +0.99% | +0.26% | +1.38% | -1.61% | -1.75% | PASS |

c16 is the default maximum running Talker batch (`max_running_requests=16`), so this covers the relevant running-concurrency range for the decode path changed by this PR.

For c1, B and C also produced the exact same WAV SHA256 in all 8 targeted confirmation rounds.

One thing I investigated further was c4. The initial 4-pair sweep showed an apparent regression of about -4.9% QPS and -5.0% engine time. I did not want to dismiss that as noise, so I repeated c4 with 8 additional balanced fresh-server B/C pairs. The regression did not reproduce:

- aggregate QPS: +0.67%
- aggregate mean latency: +0.58%
- aggregate engine time: +0.51%
- paired median QPS: +0.33%
- paired median mean latency: +0.55%
- paired median engine time: -0.59%

So I do not see evidence of a reproducible B -> C regression at c4.

I also discarded and reran any cells where a Predictor CUDA Graph was captured during measurement or where a new GPU PID appeared after measurement ownership was frozen. Those invalid cells are not included in the results above.

I excluded c32 from the PR performance qualification. With `max_running_requests=16`, c32 does not exercise a larger Talker decode batch. A simultaneous 32-request burst instead hits admission/queue behavior and can produce queue-full 503s, so I consider that a separate overload test rather than evidence for this compile-path change.

Overall, I do not see a performance or output-correctness blocker for removing the private Talker compile path. Across c1 through the maximum running concurrency c16, I could not reproduce a B -> C regression beyond the 2% decision threshold.

One qualification on interpretation: I would treat these results as evidence that the removal is performance-safe, not evidence that physically deleting the compile code itself causes a speedup. B already has the private compile path disabled, and some of the larger positive deltas varied across fresh-server repetitions.

sglang_omni/models/qwen3_tts/config.py
```diff
@@ -48,8 +48,8 @@ def process_local_edges(cls) -> frozenset[tuple[str, str]]:
 
     model_path: str
     # note (0xtoward): Keep deterministic inference opt-in because it serializes
-    # preprocessing and vocoder decoding and disables Talker compilation and the
-    # vocoder CUDA graphs, reducing throughput.
+    # preprocessing and vocoder decoding and disables the vocoder CUDA graphs,
+    # reducing throughput.
```
SandyLuXY, line 52: The matching user-facing paragraph in `docs/cookbook/qwen3_tts.md` still says deterministic inference disables Talker compilation. Please update it to remove the deleted compile path and describe both vocoder CUDA graphs consistently with this resolved policy.

SandyLuXY (changes requested):

sglang_omni/models/qwen3_tts/config.py
junliu-mde: Fixed in `6ca4a793`. The deterministic-inference paragraph no longer mentions the removed Talker compile path and now states that both the initial and follow-up vocoder CUDA Graphs are disabled.

junliu-mde:
Addressed the documentation review in `6ca4a793` and rebased onto current main `3f819f9c`. Current-head H100 verification passes: 15 Qwen3-TTS compile-policy tests and 35 model-capability tests. GitHub lint, docs, layout, CPU unit tests, and accelerator unit tests also pass. The review thread is resolved. @SandyLuXY please re-review.

SandyLuXY (approved):
lgtm

AkazaAkane (approved):

PR #1651, opened 2026-08-21
[MUSA] Add platform support

froststeam:
## Summary

Add MUSA platform support for SGLang-Omni and align selected CUDA-compatible runtime paths with the platform abstraction.

This PR includes:

- MUSA Omni platform registration and device setup
- MUSA platform export through `sglang_omni.vendor.sglang.utils`
- platform-gated fused QK-norm RoPE selection, with MUSA using the native QK-norm + RoPE path when the CUDA-only AOT op is unavailable
- CUDA/MUSA backend checks for MiniMax-Music3 and MOSS-TTS sampling graph paths
- MUSA-safe Voxtral voice embedding load by reading checkpoint tensors on CPU before moving them to the target device

## Testing

```bash
pre-commit run --all-files
```

Result: passed.

## Model benchmark summary

Environment: MUSA backend, containerized SGLang-Omni development environment. TTS/ASR dataset uses SeedTTS EN.

All commands below use public model IDs and upstream entrypoints, and are intended to run from the SGLang-Omni repository root after dependencies are installed.

### Qwen/Qwen3-TTS-12Hz-0.6B-Base

Start command:

```bash
  --model-path Qwen/Qwen3-TTS-12Hz-0.6B-Base \
  --model-name Qwen/Qwen3-TTS-12Hz-0.6B-Base \
  --config examples/configs/qwen3_tts_0_6b.yaml \
  --port 8000
```

Benchmark command:

```bash
  --generate-only --use-existing-server \
  --port 8000 \
  --lang en \
  --max-concurrency 16 \
  --output-dir results/qwen3_tts_0_6b_musa_perf
```

Result:

```text
Model: Qwen/Qwen3-TTS-12Hz-0.6B-Base
Concurrency: 16
Completed requests: 1088
Failed requests: 0
Latency mean (s): 3.436
Latency p95 (s): 3.889
RTF mean: 0.8439
Audio throughput (s/s): 17.395
Output throughput (tok/s): 217.400
Throughput (req/s): 4.053
```

### mistralai/Voxtral-4B-TTS-2603

Start command:

```bash
  --model-path mistralai/Voxtral-4B-TTS-2603 \
  --model-name mistralai/Voxtral-4B-TTS-2603 \
  --config examples/configs/voxtral_tts.yaml \
  --torch-compile off \
  --port 8000
```

Benchmark command:

```bash
  --generate-only --use-existing-server \
  --model mistralai/Voxtral-4B-TTS-2603 \
  --port 8000 \
  --lang en \
  --no-ref-audio \
  --voice cheerful_female \
  --max-new-tokens 4096 \
  --max-concurrency 16 \
  --output-dir results/voxtral_tts_4b_musa_perf
```

Result:

```text
Model: mistralai/Voxtral-4B-TTS-2603
Concurrency: 16
Completed requests: 1088
Failed requests: 0
Latency mean (s): 2.897
Latency p95 (s): 4.380
RTF mean: 0.5068
Audio throughput (s/s): 31.498
Output throughput (tok/s): 393.700
Throughput (req/s): 5.493
```

### Qwen/Qwen3-ASR-1.7B

Start command:

```bash
  --port 8000
```

Benchmark command:

```bash
  --port 8000 \
  --concurrencies 32 \
  --repeats 3 \
  --warmup \
  --output results/qwen3_asr_bench/bench_c32.json \
  --save-raw-dir results/qwen3_asr_bench/raw_c32
```

Result:

```text
Model: Qwen/Qwen3-ASR-1.7B
Concurrency: 32
Evaluated requests: 3264/3264
Throughput (samples/s): 92.930
Latency mean (s): 0.341
Latency p95 (s): 0.452
RTFx: 440.126
WER: 0.0122
```

### Qwen/Qwen3-Omni-30B-A3B-Instruct-FP8

Start command:

```bash
  --model-path Qwen/Qwen3-Omni-30B-A3B-Instruct-FP8 \
  --model-name qwen3-omni \
  --port 8000
```

Benchmark command:

```bash
curl -s http://127.0.0.1:8000/v1/chat/completions \
  -d '{
    "model": "qwen3-omni",
    "messages": [
      {
        "role": "user",
        "content": "How many cars are there in the picture? Answer with one number."
      }
    ],
    "images": ["tests/data/cars.jpg"],
    "modalities": ["text"],
    "max_tokens": 32
  }'
```

Result:

```text
Model: Qwen/Qwen3-Omni-30B-A3B-Instruct-FP8
Task: image understanding
Requests: 8
Successful responses: 8/8
Answer: 4
Latency mean (s): 0.464
Latency p50 (s): 0.462
Latency p90 (s): 0.465
Completion throughput (tok/s): 4.314
```

### MiniMaxAI/MiniMax-Music3

Start command:

```bash
  --model-path MiniMaxAI/MiniMax-Music3 \
  --model-name MiniMaxAI/MiniMax-Music3 \
  --port 8000
```

Benchmark command:

```bash
curl -X POST http://127.0.0.1:8000/v1/audio/speech \
  -d '{"model":"MiniMaxAI/MiniMax-Music3","input":"[Verse]\nA small dog dances in the sun\n[Chorus]\nRound and round, having fun","instructions":"A short cheerful pop song with acoustic guitar and light drums","seed":42,"max_new_tokens":128}' \
  --output minimax_music3_smoke.wav
```

Result:

```text
Model: MiniMaxAI/MiniMax-Music3
Requests: 3
Successful responses: 3/3
Latency mean (s): 3.076
Latency p50 (s): 2.933
Output WAV bytes: 326984
```

## Notes

```text
MOSS-TTS-Local-Transformer-v1.5 is excluded from passing benchmark claims. The saved run had 0/1088 successful requests due to request/reference-audio formatting, not a completed performance result.
Video/audio multimodal smoke is excluded from passing benchmark claims because the saved run failed in video decoding before model inference.
MiniMax-Music3 does not currently have a dedicated repository benchmark harness; the recorded result is endpoint latency over successful /v1/audio/speech requests.
```


sglang_omni/platforms/__init__.py
```diff
@@ -1,16 +1,19 @@
 import os
 import pkgutil
 
+try:
+    import torchada  # noqa: F401
+except ImportError:
+    pass
```
yeahdongcn: Remove

sglang_omni/platforms/__init__.py
```diff
@@ -1,16 +1,19 @@
...
 import torch
 from sglang.srt import platforms as srt_platforms
 from sglang.srt.platforms.interface import SRTPlatform
 
 from sglang_omni.platforms.cpu import CPUOmniPlatform
-from sglang_omni.platforms.cuda import CUDAOmniPlatform
+from sglang_omni.platforms.cuda import CUDAOmniPlatform, ROCMOmniPlatform
```
yeahdongcn: Revert

sglang_omni/platforms/__init__.py
```diff
@@ -21,14 +24,6 @@ def _is_npu_available() -> bool:
     return bool(npu.is_available())
 
 
-def _is_xpu_available() -> bool:
-    try:
-        xpu = torch.xpu
-    except AttributeError:
-        return False
-    return bool(xpu.is_available())
```
yeahdongcn, line 29: Revert and keep xpu as is.

sglang_omni/platforms/musa.py
```diff
@@ -0,0 +1,21 @@
...
+    device_name = "musa"
+    device_type = "musa"
+
+    def get_fused_qk_norm_rope(self):
+        # sgl-kernel's AOT fused_qk_norm_rope op is CUDA-only today.
+        # Use the native QK-norm + RoPE path on MUSA.
+        return None
+
+    def set_device(self, device: torch.device | int) -> None:
+        torch.musa.set_device(device)
```
yeahdongcn: Please check if this can be removed: https://github.com/MooreThreads/torchada/blob/main/src/torchada/cuda/__init__.py#L63

sglang_omni/models/voxtral_tts/pipeline/stages.py
```diff
@@ -229,10 +229,10 @@ def _load_voxtral_voice_embeddings(
         name = fname.removesuffix(".pt")
         emb = torch.load(
             os.path.join(voice_dir, fname),
-            map_location=device,
+            map_location="cpu",
```
yeahdongcn: Why change this?

sglang_omni/platforms/musa.py
```diff
@@ -0,0 +1,21 @@
...
+import torch
+from sglang.srt.platforms.device_mixin import PlatformEnum
+
+from sglang_omni.platforms.cuda import CUDAOmniPlatform
+
+
+class MUSAOmniPlatform(CUDAOmniPlatform):
+    _enum = PlatformEnum.MUSA
+    device_name = "musa"
+    device_type = "musa"
```
yeahdongcn, line 24: Do we need any additional handling for `MUSA_VISIBLE_DEVICES` and backend selection?

sglang_omni/platforms/musa.py
froststeam: `torch.cuda.set_device()` works on MUSA after torchada is imported and its CUDA-to-MUSA patch is applied.

The reason I kept this override is the initialization order: SGLang-Omni stage workers call `current_platform.set_device()` very early during startup. If the CUDA implementation is reused before torchada patching is guaranteed in that child process, it can hit native `torch.cuda.set_device()` and fail with `_cuda_setDevice` missing.

So this is mainly to avoid an early import-order dependency, not because torchada's patched `torch.cuda.set_device()` itself is wrong.

sglang_omni/models/voxtral_tts/pipeline/stages.py
froststeam: This is for checkpoint compatibility.

The voice embedding `.pt` files may contain tensors saved from CUDA devices. In a MUSA-only runtime, loading them directly with `map_location=device` can still hit CUDA deserialization first and fail because `torch.cuda.is_available()` is false.

So we load the small voice embedding tensors to CPU first, then explicitly move them to the target device:

`torch.load(..., map_location="cpu")`
`emb.to(device=device, dtype=torch.bfloat16)`

This avoids CUDA storage deserialization issues while keeping the final tensor placement unchanged.

yeahdongcn: How about handling MUSA as a special case here? We can use the `current_platform` check to set the device to CPU only for MUSA, while keeping the behavior unchanged for other platforms.

sglang_omni/platforms/musa.py
froststeam: No additional handling is needed in SGLang-Omni now. I moved this compatibility into torchada.

For device visibility, torchada now provides a global fallback:
https://github.com/froststeam/torchada/blob/62da0a2adbd38e90b10083d8ddfdf1d048304162/src/torchada/_patch.py#L68-L71

So if `CUDA_VISIBLE_DEVICES` is set but `MUSA_VISIBLE_DEVICES` is not, importing torchada will populate `MUSA_VISIBLE_DEVICES`.

sglang_omni/vendor/sglang/utils.py
```diff
@@ -26,5 +27,6 @@
     "is_flashinfer_available",
     "is_non_idle_and_non_empty",
     "is_npu",
+    "is_musa",
```
yeahdongcn: Nit: Keep the alphabetical order, consistent with the imports.

sglang_omni/platforms/musa.py
yeahdongcn: What about backend selection? https://github.com/sgl-project/sglang-omni/blob/main/sglang_omni/platforms/cuda.py#L90

sglang_omni/models/voxtral_tts/pipeline/stages.py
froststeam: Updated. Now this CPU load path is only used for MUSA:

`map_location = "cpu" if current_platform.is_musa() else device`

Other platforms keep the previous behavior with `map_location=device`. I also added a unit test to cover both MUSA and non-MUSA map_location selection.

sglang_omni/platforms/musa.py
froststeam: Good catch. CUDA may choose model-specific backends like `cutlass` / `flashinfer_cutlass` there, which should not be inherited by MUSA.

I added an override in `MUSAOmniPlatform` to keep MUSA backends as `auto`.

yeahdongcn (approved):

sglang_omni/platforms/__init__.py
```diff
@@ -49,6 +50,8 @@ def _as_omni_platform(platform: SRTPlatform) -> OmniPlatform:
         return CUDAOmniPlatform()
     if platform.is_rocm():
         return ROCMOmniPlatform()
+    if platform.is_musa():
```
Ratish1: This branch only handles a MUSA-aware SGLang platform object, but pinned SGLang provides none. Add the SGLang `is_musa()` detector to the existing unresolved-base fallback so the default MUSA installation can reach `MUSAOmniPlatform`.

sglang_omni/platforms/musa.py
```diff
@@ -0,0 +1,36 @@
...
+from __future__ import annotations
+
+import torch
+from sglang.srt.platforms.device_mixin import PlatformEnum
+
+from sglang_omni.platforms.cuda import CUDAOmniPlatform
+from sglang_omni.platforms.interface import OmniPlatform
+
+
+class MUSAOmniPlatform(CUDAOmniPlatform):
```
Ratish1, line 21: Spawned workers construct this class directly from `SGLANG_OMNI_PLATFORM_SPEC`. Call the SGLang `is_musa()` detector during construction so TorchADA is applied before inherited `torch.cuda` operations run, and fail if it is unavailable.

sglang_omni/platforms/musa.py
```diff
@@ -0,0 +1,36 @@
...
+        # Use the native QK-norm + RoPE path on MUSA.
+        return None
+
+    def apply_model_worker_backend_policy(
+        self,
+        server_args: ServerArgs,
+        model_config: ModelConfig,
+        model_arch_override: str | None,
+    ) -> str | None:
+        if hasattr(server_args, "moe_runner_backend"):
```
Ratish1: dont use hasattr

sglang_omni/platforms/musa.py
```diff
@@ -0,0 +1,36 @@
...
+    device_name = "musa"
+    device_type = "musa"
+
+    def get_device(self, local_rank: int) -> torch.device:
+        return torch.device("musa", local_rank)
+
+    def get_fused_qk_norm_rope(self):
+        # sgl-kernel's AOT fused_qk_norm_rope op is CUDA-only today.
+        # Use the native QK-norm + RoPE path on MUSA.
+        return None
```
Ratish1, line 29: Add `enable_code2wav_graph() -> False`; the current graph runner rejects `device.type == "musa"`, while MUSA inherits `True` from CUDA.

sglang_omni/platforms/musa.py
```diff
@@ -0,0 +1,36 @@
...
+from sglang.srt.platforms.device_mixin import PlatformEnum
+
+from sglang_omni.platforms.cuda import CUDAOmniPlatform
+from sglang_omni.platforms.interface import OmniPlatform
+
+
+class MUSAOmniPlatform(CUDAOmniPlatform):
+    _enum = PlatformEnum.MUSA
+    device_name = "musa"
+    device_type = "musa"
```
Ratish1, line 24: Use `obj.device.type` in `_tensor_devices()`. TorchADA makes MUSA tensors report `is_cuda=True`, so the current check classifies them as CUDA and `outbound_payload()` rejects them against `device_type="musa"`.

sglang_omni/models/moss_tts/sglang_model.py
```diff
@@ -390,8 +391,8 @@ def is_sampling_cuda_graph_compatible(data: Any) -> bool:
         return matches_graph_profile(data)
 
     def _sampling_graph_support_reason(self) -> str | None:
-        if self.device.type != "cuda" or not torch.cuda.is_available():
-            return "CUDA is unavailable"
+        if not (current_platform.is_cuda() or current_platform.is_musa()):
```
Ratish1, line 394: MUSA is enabled here, but `MossTTSDelaySamplingCudaGraphRunner.clear()` still releases cache only for literal CUDA. Update `clear()` to use the active device module for CUDA and MUSA.

sglang_omni/platforms/musa.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def apply_model_worker_backend_policy(
+        self,
+        server_args: ServerArgs,
+        model_config: ModelConfig,
+        model_arch_override: str | None,
+    ) -> str | None:
+        if hasattr(server_args, "moe_runner_backend"):
+            server_args.moe_runner_backend = "auto"
+        if hasattr(server_args, "fp8_gemm_runner_backend"):
```
Ratish1: dont use hasattr

sglang_omni/platforms/musa.py
froststeam: Fixed. I import torchada in the MUSA platform module so workers created from SGLANG_OMNI_PLATFORM_SPEC get the CUDA-compatible patches as well.

sglang_omni/platforms/musa.py
froststeam: I kept this enabled and updated the graph runner to accept MUSA devices. The runner still uses the existing torch.cuda graph APIs, which TorchADA maps for MUSA.

sglang_omni/platforms/musa.py
froststeam: Fixed by checking device.type == "musa" before is_cuda, since TorchADA MUSA tensors also report is_cuda=True.

sglang_omni/models/moss_tts/sglang_model.py
froststeam: Fixed. clear() now uses torch.get_device_module(self.model.device) instead of hardcoded torch.cuda.

sglang_omni/platforms/musa.py
froststeam: Fixed, removed the hasattr checks.

sglang_omni/platforms/musa.py
froststeam: Fixed, removed the hasattr checks.

sglang_omni/platforms/__init__.py
froststeam: Thanks for pointing this out. The SGLang-side MUSA platform support is being upstreamed here: https://github.com/sgl-project/sglang/pull/35855

Before that lands, this PR should still be safe for other platforms: the MUSA path is only selected when SGLang reports a MUSA platform, and I can keep the MUSA import lazy so CPU/CUDA/ROCm won’t import torchada or hit this path.

It would be great if we can merge this Omni-side support first.

sglang_omni/platforms/musa.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def apply_model_worker_backend_policy(
+        self,
+        server_args: ServerArgs,
+        model_config: ModelConfig,
+        model_arch_override: str | None,
+    ) -> str | None:
+        server_args.moe_runner_backend = "auto"
+        server_args.fp8_gemm_runner_backend = "auto"
+        return OmniPlatform.apply_model_worker_backend_policy(
```
Ratish1: nit: we dont need these assignments, can we do it like this?

```suggestion
        return OmniPlatform.apply_model_worker_backend_policy(
```

Ratish1 (approved):
LGTM

PR #1339, opened 2026-08-05
[ASR Refactor][F3] Whisper: map prompt to prev-context tokens

jiannan-17:
## Summary

Implements the [F3 task](https://github.com/sgl-project/sglang-omni/issues/1267#issuecomment-5185876009).

- Use the transcription `prompt` as Whisper previous context.
- Keep prompt, prefix, and generation within the decoder position limit.
- Return HTTP 400 for prompts containing disallowed special tokens.

## Flow

```mermaid
flowchart LR
    A["Transcription request<br/>with prompt"] --> B["Validate and tokenize prompt"]
    B --> C["Keep the latest prompt tokens<br/>and add start-of-prev"]
    C --> D["Allocate the decoder budget"]
    D --> E["Previous context + task prefix<br/>+ generated tokens"]
    E --> F["Whisper decoding"]
    B -->|"Disallowed special token"| G["HTTP 400"]
```

### Supported cases

All decoder inputs and outputs share the same position budget:

`previous context + task prefix + generated tokens <= 448`

These examples use `whisper-large-v3` with a 4-token task prefix. Previous context includes one start-of-prev marker.

| Request | Behavior |
|---|---|
| Prompt omitted or blank | Keep the existing prefix-only behavior |
| Prompt with `max_new_tokens=256` | Keep up to the last 187 prompt tokens; `188 prev + 4 prefix + 256 output = 448` |
| Prompt with `max_new_tokens=300` | Keep up to the last 143 prompt tokens; `144 prev + 4 prefix + 300 output = 448` |
| No prompt with `max_new_tokens=600` | Clamp generation to 444 tokens; `4 prefix + 444 output = 448` |
| Prompt containing a Whisper special token | Reject the request with HTTP 400 before decoding |

The decoder budget uses `max_target_positions` as the limit. Mandatory prefix and generation tokens take priority; previous context uses the remaining space. A final guard rejects an invalid budget before GPU execution.

### Unit and configuration checks

- Whisper unit tests: **11 passed**.
- Exhaustive budget sweep: `max_new_tokens=1..600` x `prompt_length=0..300`, covering **180,600 combinations**. Every case satisfies `len(prompt_token_ids) + max_new_tokens <= 448`, or drops the prompt when no previous-context budget remains.
- Guard regression test: an intentionally oversized previous context raises `Whisper decoder budget exceeded` before GPU execution.
- Real `whisper-large-v3` config: `max_target_positions == generation_config.max_length == 448`.
[...]

PR #1654, opened 2026-08-21
[ASR Feature][Whisper] Chain previous decoded context across transcription chunks

AkazaAkane:
## Motivation

PR #1339 maps each transcription request prompt into Whisper previous-context tokens. Long-audio chunking applies that caller prompt to every independently decoded chunk. This PR adds an explicit opt-in path for evaluating preceding-chunk conditioning and bounded repetition recovery without changing the default concurrent behavior.

## Modifications

- Add adapter-level chunk orchestration hooks with concurrent decoding as the default.
- Register a Whisper adapter with opt-in ordered decoding; `condition_on_previous_text` defaults to `false`, while the enabled path uses the caller prompt for the first decoded chunk and the preceding decoded result thereafter.
- Keep prompt tokenization and decoder-budget truncation in the existing Whisper request builder.
- Add an adapter-owned retry policy for sustained repetition on the enabled path.
- When conditioning is enabled, retry a degenerate Whisper chunk at most once with `prompt=None`; discard the first result and use the retry output as the next chunk's context.
- Detect periodic runs anywhere in normalized character-like units so repetitions followed by valid-looking continuation are recovered and Chinese, Japanese, Thai, and other languages do not depend on whitespace boundaries.
- Add focused coverage for prompt chaining, retry recovery and bounds, retry cancellation, detector positives/negatives, and retained non-Whisper concurrency.
- Document Whisper's ordered long-audio behavior, retry policy, detector scope, and concurrency tradeoff.

## Related Issues

Follow-up to #1339.

### Explicit opt-in Whisper unit semantics: PASS
Three chunks received previous-context prompts in the expected order:
chunk 0: caller prompt
chunk 1: decoded text from chunk 0 (part0)
chunk 2: decoded text from chunk 1 (part1)
max_active == 1, confirming chunks within one Whisper request are decoded sequentially.

### Non-Whisper regression: PASS
Three chunks ran concurrently under the default adapter path.
max_active == 3, confirming this change does not serialize models whose chunks are independent.

### Explicit opt-in real Whisper E2E: PASS

Model: openai/whisper-base
Endpoint: /v1/audio/transcriptions
Input duration: 36.96 s
Result: HTTP 200
End-to-end latency: 19.54 s
Long audio was split into two segments:
chunk 0: 0–28.95 s
chunk 1: 28.95–36.96 s
Caller prompt for chunk 0: CALLER-CONTEXT-1654
Chunk 0 decoded text: How many cars are there in the picture?
Chunk 1 prompt: exactly the decoded text from chunk 0

This verifies the enabled path implements:

caller prompt → chunk 0 decoded text → chunk 1 prev-context

### Explicit opt-in ordered-decoding cancellation: PASS
Only chunk 0 was submitted before cancellation.
Cancelling the ordered Whisper path aborted the in-flight chunk 0 engine request.
No later chunk was started.

### Recovery evaluation

- Six bad-context chunks were detected and retried.
- TED-LIUM remained byte-identical with zero unnecessary retries.
- Retries stayed bounded.
- Three of four pathological files were repaired.
- For the remaining `long_300_08` case, a context-free retry of the missed chunk produced 6.08% WER with zero loops; its repeated span was followed by valid-looking continuation, motivating detection anywhere in the decoded text.

### Degenerate-chunk recovery coverage: ADDED
Focused cases cover normal no-retry chaining, one context-free retry, discarding the first result, retry output feeding the next chunk, a bounded second degenerate result, caller-prompt failure on chunk 0, cancellation during retry, retained non-Whisper concurrency, multilingual detector positives, repetition before valid continuation, and conservative negatives for two copies, short repetition, and similar-but-non-identical clauses.

These new retry cases were added but not run locally under the repository's pre-commit-only instruction; CI should revalidate them.

The current implementation intentionally uses one bounded retry with no previous context; it does not implement a 64-token-then-none ladder. In the offline policy comparison, the best measured policy was the experimental 64 → 0 ladder at 10.26% mean per-file WER and zero loops. The evaluated no-context policy measured 15.03% and zero loops, while the detector-gated implementation repaired three of four pathological files; a context-free retry of the missed long_300_08 chunk produced 6.08% WER and zero loops. The detector was then broadened to catch repetitions followed by valid-looking continuation, but the current head has not been rerun on that five-file set.

The PR keeps the narrower one-retry state machine so recovery is bounded to one additional engine request. Retry output becomes full context for the following chunk; there is no retry loop.

## Latest c16/c32 Evaluation

Eight distinct transcript-aligned TED-LIUM clips were assembled from 55 consecutive utterances:

- 61–71 seconds per clip
- 522.6 seconds / 8.7 minutes of unique audio
- Three Whisper chunks per request
- c16: each clip repeated twice
- c32: each clip repeated four times
- Separate non-overlapping warmup clip

The PR evaluation used revision `9ccd3b3` with `condition_on_previous_text` enabled. Revision `9ef8391` changes the default to `false` without changing that opt-in path.

| Concurrency | Revision | Mean latency | P95 | Wall time | req/s | Normalized WER |
|---:|:---|---:|---:|---:|---:|---:|
| 16 | main `5b2d885` | 1.925 s | 1.999 s | 2.003 s | 7.988 | 2.66% |
| 16 | PR `9ccd3b3` | 2.370 s | 2.667 s | 2.672 s | 5.987 | 2.90% |
| 32 | main `5b2d885` | 2.795 s | 3.018 s | 3.078 s | 10.396 | 2.66% |
| 32 | PR `9ccd3b3` | 3.237 s | 3.682 s | 3.688 s | 8.676 | 2.90% |

Enabling previous-text conditioning showed no accuracy or throughput advantage in this evaluation: normalized WER was 0.24 percentage points higher, throughput was lower at both concurrency levels, and latency was higher. The feature therefore defaults to `false`; disabled chunks remain independent and retain per-request concurrency.

This opt-in is not identical to [OpenAI Whisper](https://github.com/openai/whisper/blob/main/whisper/transcribe.py): OpenAI Whisper defaults `condition_on_previous_text=True`, carries decoded token history across internal windows, and can reset that history during fallback. This PR passes only the immediately preceding server chunk decoded text. Matching OpenAI Whisper token-level behavior or reconsidering the default should be evaluated in a separate PR.

## Checklist

- [x] Add unit tests.

## CI

Focused pre-commit passed across all current-head changed files. Per repository instruction, unit tests, GPU tests, and benchmarks were not run by the coding agent; the latest c16/c32 evaluation above was supplied separately.


sglang_omni/serve/transcription_adapters/whisper.py
```diff
@@ -0,0 +1,57 @@
...
+import unicodedata
+
+from sglang_omni.serve.transcription_adapters.base import (
+    DefaultTranscriptionAdapter,
+    register_transcription_adapter,
+)
+
+_MIN_REPEAT_PERIOD_UNITS = 8
+_MAX_REPEAT_PERIOD_UNITS = 128
+_REPEAT_COPIES = 3
```
SandyLuXY, line 15: I'm wondering how the rules for repeat copies were derived. My concern is that three copies of any 8+ unit word are classified as degenerate, so a faithful transcript is discarded and re-decoded.

sglang_omni/serve/transcription_adapters/whisper.py
```diff
@@ -0,0 +1,57 @@
...
+    normalized = unicodedata.normalize("NFKC", text).casefold()
+    units = tuple(
+        char
+        for char in normalized
+        if char.isalnum() or unicodedata.category(char).startswith("M")
+    )
+    max_period = min(_MAX_REPEAT_PERIOD_UNITS, len(units) // _REPEAT_COPIES)
+    for period in range(_MIN_REPEAT_PERIOD_UNITS, max_period + 1):
+        span = _REPEAT_COPIES * period
+        for start in range(len(units) - span + 1):
```
SandyLuXY, line 47: This slice-heavy scan runs synchronously on the FastAPI event loop after every Whisper chunk. Consider using an allocation-light detector or move this work off the event loop, then re-benchmark the current head because the submitted c16/c32 runs

sglang_omni/serve/transcriptions.py
```diff
@@ -368,9 +380,38 @@ async def run_chunk(span: ChunkSpan) -> str:
...
+                chunk_prompt = adapter.chunk_prompt(
+                    caller_prompt=prompt,
+                    previous_text=previous_text,
+                    is_first_decoded_chunk=is_first_decoded_chunk,
+                )
+                text = await run_chunk(span, chunk_prompt=chunk_prompt)
+                if adapter.should_retry_chunk_without_context(text):
+                    text = await run_chunk(
+                        span,
+                        chunk_prompt=None,
```
SandyLuXY, line 408: This goes directly from full context to `None`, but the PR's best reported recovery result is for a last-64-then-none ladder. Could you confirm what's the actual best recovery result?

tests/README.md
```diff
@@ -123,7 +123,8 @@ tests/
     │   ├── test_engine_io.py
     │   ├── test_model_runner.py
     │   ├── test_reference_encode.py
-    │   └── test_request_builders.py
+    │   ├── test_request_builders.py
+    │   └── test_transcription_adapter.py
```
SandyLuXY: This entry is under `unit_test/ming_tts/`, but the new file is `unit_test/whisper_asr/test_transcription_adapter.py`. Please update

SandyLuXY (changes requested):

sglang_omni/serve/transcription_adapters/whisper.py
AkazaAkane: I derive the strategy based on TED-LIUM talks. But I think your point is solid. Should be another follow-up pr focusing on the strategy to dealing with long audio repetition problem.

sglang_omni/serve/transcriptions.py
AkazaAkane: Corrected the PR description: 64 → None was the best experimental policy, not the implemented policy, and existing benchmarks predate the detector.

tests/README.md
AkazaAkane: fixed

sglang_omni/serve/transcription_adapters/whisper.py
AkazaAkane: rebenchmark, latest result does not show great WER improvment but obvious regression on latency and throughput. Thus, I will set it disable by default.

sglang_omni/serve/transcription_adapters/whisper.py
0xjeffro: how about we investigate how openai/whisper handles this case and just align with them?

0xjeffro (approved):

SandyLuXY (approved):

Issue #1652, opened 2026-08-21
[Roadmap] Optimization for CosyVoice3


PoTaTo-Mika:
SGLang-Omni has already supported CosyVoice3-0.5B, and we are going to optimize this model.

Open for contributions, please quote this issue in your pull request.

# GPU
## S3-Tokenizer

- [x] Multi-thread speaker encoder encoding. #1755 
- [x] Batch inference for S3-Tokenizer. [Tested no effect] #1693 

## AR

- [x] Streaming output. https://github.com/sgl-project/sglang-omni/pull/1656
- [x] AR overlap schedule. [Tested no effect] https://github.com/sgl-project/sglang-omni/pull/1669

## DiT

- [x] Batch inference for Flow decoder. https://github.com/sgl-project/sglang-omni/pull/1663
- [x] Chunk-wise inference for streaming. https://github.com/sgl-project/sglang-omni/pull/1656
- [x] torch.compile for DiT backbone. https://github.com/sgl-project/sglang-omni/pull/1670
- [x] TensorRT acceleration. #1858 
- [x] Fix flow auto-cast bugs. https://github.com/sgl-project/sglang-omni/pull/1715
- [ ] (**This is an optional optimization**) Step-distillation for flow-decoder. #1673 We should support LoRA at the framework level. https://github.com/a710128/nanovllm-voxcpm https://huggingface.co/collections/openbmb/voxcpm @PoTaTo-Mika we shall make a sub-roadmap for this

## Vocoder

- [x] Batch inference for HiFT Vocoder. #1844 

## General

Several kernel fusions are welcome if proven efficient.

# NPU (Ascend)

- [x] Support model. https://github.com/sgl-project/sglang-omni/pull/1694
- [ ] TBD...

nagisa-kunhah:
Hi @PoTaTo-Mika , is this issue open for contributions? If so, I'd like to give Batch Inference for the Flow Decoder a try.

guozhihao-224:
hi @PoTaTo-Mika  I'd like take "chunk-wise inference for streaming." of DIT stage

yuangxu:
Hello @PoTaTo-Mika, I'd like to work on torch.compile support and decode CUDA-graph for the AR stage

CAICAIIs:
hi, I'd like to take 'torch.compile for DiT backbone' — plan: align chunk-shape buckets with the streaming work, benchmark eager vs compiled on RTF, gate on WER/similarity.

[...]

PR #1655, opened 2026-08-21
[Perf][Fun-CosyVoice3] Batch HiFT vocoder inference

wirybeaver:
## Motivation

Fun-CosyVoice3's vocoder scheduler can coalesce requests, but HiFT was still invoked once per request. This implements the HiFT batch-inference item from the CosyVoice3 optimization roadmap without changing the existing Flow batching behavior.

## Modifications

- retain the existing Flow inference path and group its resulting mel tensors by exact frame length
- invoke HiFT once per compatible group without padding variable-length mel tensors
- restore waveforms to original request order and preserve each request's sample rate
- reject unexpected HiFT output batch cardinality
- add unit coverage for mixed-length grouping, batched tensor shapes, output ordering, and cardinality errors

## Related Issues

Related to #1652.

## Accuracy Test

- Fun-CosyVoice3 unit suite: 42 passed
- focused vocoder suite: 8 passed
- adjacent dots.tts and Fish Audio vocoder suites: 14 passed
- changed-line coverage: 100% (48/48)

The batching path only combines equal-length mel tensors; it does not pad variable-length inputs.

A paired real-checkpoint HiFT replay on H100 used three fixed speech-feature tensors extracted with the pinned CosyVoice frontend (207, 230, and 267 mel frames). Each batch repeated the same tensor so serial and batched calls consumed identical inputs. All output cardinality checks passed, all waveforms were finite, and batch-size-one output was exactly identical.

For batch sizes 2, 4, and 8, the observed numerical differences versus serial inference were input-dependent: RMS difference ranged from `3.88e-4` to `3.60e-3`, SNR from `26.38` to `37.94 dB`, and maximum absolute difference reached `0.0730`. At batch 8:

- 207 frames: RMS `0.003601`, SNR `26.38 dB`
- 230 frames: RMS `0.002045`, SNR `33.84 dB`
- 267 frames: RMS `0.000894`, SNR `30.69 dB`

These differences are larger than the earlier independent validation on another mel set (`3e-4` to `1.8e-3`, `41-57 dB`), indicating that the numerical delta depends on the input. No perceptual-equivalence claim is made from these metrics alone. The earlier real-checkpoint serving validation reported sane waveforms, correct ordering/sample rates, and no request failures: https://github.com/sgl-project/sglang-omni/pull/1655#issuecomment-5377558074.

### HiFT-only replay

[...]

PR #1656, opened 2026-08-21
[Fun-CosyVoice3] Causal streaming vocoder (AR stream_to + chunk-wise Flow/HiFT)

guozhihao-224:
## Motivation
Fun-CosyVoice3 already served buffered `/v1/audio/speech`. This PR enables causal streaming that matches upstream `CosyVoice3Model.token2wav`: the AR stage emits speech tokens as they are generated, and Flow + HiFT decode chunk-wise so the client can start playback before the utterance finishes.

Covers two items on the CosyVoice3 optimization roadmap (#1652):
- AR: Streaming output
- DiT: Chunk-wise inference for streaming

## Modifications
- Pipeline: `tts_engine.stream_to=["vocoder"]`, `can_accept_stream_before_payload=True`, `supports_streaming_vocoder=True`.
- AR (`model_runner`): collect FSQ tokens (`id < 6561`); first flush at 28 tokens (`hop=25 + lookahead=3`), then every 25; first chunk carries Flow prompt tensors; `on_request_finished` force-flushes the remainder.
- Vocoder: `FunCosyVoice3StreamingVocoderScheduler` on `StreamingVocoderBase`. Intermediate hops: `flow.inference(..., streaming=True, finalize=False)` on `tokens[:offset+hop+3]`, slice mel at `token_offset * 2`. Remainder: `streaming=False, finalize=True`. Hop growth `25 → 50 → 100`. First hop pads prompt token length to a multiple of 25. Causal HiFT accumulates mel and emits `speech[speech_offset:]`.
- Non-streaming stays a single `token2wav(..., streaming=False, finalize=True)` pass. Speed is still applied on the shared speech-response path, not inside the vocoder. No CosyVoice2 overlap-fade, no `inference_chunk` KV cache, no hop-size change.
- Docs: cookbook streaming (`stream: true` + `response_format: "pcm"`) and SeedTTS `--stream --generate-only`; `docs/basic_usage/tts.md` notes causal streaming.
- Tests: `tests/unit_test/fun_cosyvoice3/test_streaming.py` (hop math, payload-first vocoder, lookahead gate, AR flush, hop 25→50 glue). Existing vocoder kernel tests keep a FakeFlow that does not subtract lookahead.

## Related Issues
Partial of #1652 (AR streaming output + DiT chunk-wise streaming).
Does not close the issue.

## Accuracy Test
Non-streaming path is unchanged (one Flow + HiFT pass over the full token sequence). Causal streaming can differ slightly from the buffered path, as documented in the cookbook Known Limitations.

Unit: `tests/unit_test/fun_cosyvoice3/` (47 passed).

GPU generate-only, `stream: true`, Seed-TTS-Eval, 0 failed requests:
- EN 1088, `c=16`
- ZH 2020, `c=16`, `--no-ref-text`

No ASR / WER pass in this PR.

## Benchmark & Profiling
Streaming generate-only on 1× GPU, production yaml, `--stream`:

| Split | n | c | Failed | Latency mean | RTF mean | TTFC mean | Throughput |
| EN | 1088 | 16 | 0 | 14.23 s | 3.26 | 13.28 s | 1.12 req/s |
| ZH | 2020 | 16 | 0 | 13.65 s | 2.66 | 12.51 s | 1.17 req/s |

Streaming is working (first PCM ≈ 1.3 s audio, not the full utterance;`lat − TTFC` ≈ 1 s). The high `c=16` TTFC is GPU saturation: Flow `inference` is `assert batch==1` and vocoder `decode_batch` is a serial loop, so 16 first hops queue on one GPU. Unloaded / first-in-wave TTFC is 0.55–0.76 s; earlier `c=2` smoke was ~1.8 s. Use low concurrency to measure TTFC; `c=16` is a throughput point copied from the non-stream cookbook. Follow-ups on #1652: Flow decoder batch inference, HiFT batch inference, DiT `torch.compile`.

## Checklist
- [x] Add unit tests.


Kozmosa:
hi @guozhihao-224

i also picked up the ar streaming output item and built the producer side with the moss-style stream_output_builder pattern. your PR covers the full path end to end, so mine is now positioned as a reference implementation only, no overlap intent from my side.

i have an H100 box with the Fun-CosyVoice3-0.5B-2512 checkpoint and a serving benchmark harness set up, and recently ran third-party validation for #1655 (unit + serving + in-process numerics). if useful for this PR i am glad to help with independent gpu verification of the streaming path, review, or anything else. also feel free to lift anything from #1662 if the scheduler-builder emission pattern is ever preferable to the model_runner outbox for some part.

guozhihao-224:
Thanks @Kozmosa — really glad you wrote,That's the kind of coordination this roadmap needs. This PR will add you as a co-author on the follow-up commit / at merge.

Happy to treat this as shared work. Independent H100 verification would be genuinely useful: we have SeedTTS EN/ZH generate-only on one H200, but a second box (and a low-concurrency TTFC pass at c=1 or c=2, not only c=16) would make the streaming claim much stronger. If you hit gaps, please open them here — patches on this branch are welcome.

On the producer shape: this PR currently flushes at the vocoder hop gates (28 then 25, with Flow prompt tensors on the first chunk), because CosyVoice3's DiT only runs at those boundaries. #1662's builder has a clean cursor / exactly-once story, and I'd like to dig into whether the hop gate should live in the producer, the vocoder, or both.

Kozmosa:
thanks @guozhihao-224, glad this is useful. i will run the low-concurrency ttfc pass (c=1 and c=2, en and zh if i can get a zh prompt wav working) on my H100 today and report back here.

quick thought on the hop gate while i set that up: i would put it in the producer and keep the vocoder boundary-agnostic. the dit only makes progress at hop boundaries either way, emitting finer than the hop just adds per-step ipc for no decode progress, so the producer throttling to 28/25 seems right. the vocoder still accepting arbitrary chunk boundaries keeps the two sides decoupled, and the exactly-once cursor story should hold wherever the gate ends up. happy to dig deeper once i have your branch running here.

Kozmosa:
ran the low-concurrency pass. 1x H100 PCIe, colocated 3 stages, Fun-CosyVoice3-0.5B-2512, fixed seeds per request. en used ref_text, zh ran cross-lingual (no transcript for my prompt wav). 0 errors across everything.

short utterances (en ~4.7s / zh ~3.2s audio), stream=true pcm:
- c=1: en ttfc avg 0.59s (0.48-0.91), total 1.08s | zh ttfc avg 0.61s (0.40-1.31), total 0.89s
- c=2: en ttfc avg 1.08s | zh avg 0.79s

this matches your unloaded 0.55-0.76s ttfc claim on an independent box (h100 vs your h200).

one honest caveat worth a line in the writeup: at c=1 with short utterances the pipeline is so fast that buffered completes in 0.47-0.58s total, so ttfc is roughly equal to buffered latency and chunk-wise flow+hift roughly doubles the server wall time. the headstart shows up clearly once audio gets longer: same setup, ~24s utterance, c=1, stream ttfc 0.53s vs buffered total 2.34s, about 4.4x earlier playback start. streaming total 5.0s is still rtf ~0.2.

happy to rerun any cell with other lengths or concurrency if useful.

sglang_omni/models/fun_cosyvoice3/model_runner.py
```diff
@@ -77,18 +93,77 @@ def _collect_tokens(
...
+        sched_req: Any,
+        token: torch.Tensor,
+    ) -> None:
+        data = sched_req.data
+        if self._outbox is None or data.stream_metadata is None:
+            return
+        data.stream_code_buffer.append(token)
+        data.stream_code_seen += 1
+        if int(data.stream_code_next_flush) <= 0:
+            data.stream_code_next_flush = AR_INITIAL_FLUSH_TOKENS
```
AkazaAkane: `_queue_or_emit_code_chunk` always schedules the first AR emission at 28 tokens, but the vocoder's first causal boundary is `25 + prompt_pad + 3`. For a reference prompt whose Flow-token length is not divisible by 25, the 28-token message cannot decode; because the next producer message is not sent until token 53, TTFC is delayed by `25 - prompt_pad` AR steps. For example, prompt length 10 gives `prompt_pad=15`, so the earliest valid boundary is 43, but this path waits until 53.

sglang_omni/models/fun_cosyvoice3/streaming_vocoder.py
```diff
@@ -0,0 +1,347 @@
...
+
+    def fallback_full_decode(
+        self,
+        request_id: str,
+        payload: StagePayload,
+        state: _CosyVoice3StreamState,
+    ) -> torch.Tensor | None:
+        del request_id, state
+        pipeline_state = FunCosyVoice3State.from_dict(payload.data)
+        if pipeline_state.audio_codes is None:
```
AkazaAkane, line 805: `fallback_full_decode` returns `None` when `audio_codes` is missing or empty.
When no stream chunk was emitted, `StreamingVocoderBase.on_stream_done` then
continues and emits a successful terminal result with no audio.

but the buffered path explicitly treats an empty token sequence as an error in
`token2wav_chunk`, and `test_cosyvoice3_vocoder_raises_on_empty_token_sequence`
documents that contract. The streaming path should not turn a control-only or
otherwise failed AR generation into an empty successful response.

sglang_omni/models/fun_cosyvoice3/model_runner.py
guozhihao-224: Good catch — the producer was using a pad=0 first flush (28) while the vocoder's first causal boundary is 28 + prompt_pad.

Fixed in [bf924b7](https://github.com/sgl-project/sglang-omni/pull/1656/commits/bf924b76244d6e4c94a79f4fc7886ae756494bc3): the first AR flush is now `first_ar_flush_tokens(prompt_len)` so a prompt of length 10 (pad=15) emits at 43, not 53. Added a unit test for that case, and the cookbook no longer hard-codes 28.

sglang_omni/models/fun_cosyvoice3/streaming_vocoder.py
guozhihao-224: Agreed. `fallback_full_decode` now raises the same `RuntimeError` as `token2wav_chunk` (`no usable speech tokens`) when `audio_codes` is missing or empty, instead of returning a successful empty result. Covered in `test_streaming_vocoder_fallback_raises_on_empty_audio_codes`.

guozhihao-224:
## update in 
## Setup
Hardware: 1× NVIDIA H200（CUDA_VISIBLE_DEVICES=5）
Model: FunAudioLLM/Fun-CosyVoice3-0.5B-2512
Config: examples/configs/fun_cosyvoice3_0_5b.yaml（enable_dit_torch_compile: false，默认 bf16）
Runtime: SGLang 0.5.18 / torch 2.13
Client: benchmark_tts_seedtts --generate-only --stream --max-concurrency 16 --seed 1234
Workload: SeedTTS EN 1088，带 ref_text，warmup 16

## Results


Metric | #1656 stream, compile=off | #1656 stream, compile=on | #1656 buffered, compile=off | #1656 buffered, compile=on
-- | -- | -- | -- | --
Throughput (req/s) | 1.051 | 1.852 (+76%) | 2.331 (+122%) | 2.324 (+121%)
Latency mean (s) | 15.151 | 8.602 (−43%) | 6.833 (−55%) | 6.842 (−55%)
Latency p95 (s) | 21.364 | 12.122 (−43%) | 8.429 (−61%) | 8.265 (−61%)
RTF mean | 3.504 | 1.974 (−44%) | 1.575 (−55%) | 1.577 (−55%)
TTFC mean (s) | 14.247 | 7.926 (−44%) | n/a | n/a

guozhihao-224:
@PoTaTo-Mika  Full Seed-TTS en, 1088 samples, stream, generate-only, seed 1234. Remote feat/fun-cosyvoice3-causal-streaming with DiT torch.compile on;
Concurrency | TTFC mean | Latency mean | Throughput | Failures
-- | -- | -- | -- | --
c=1 | 0.68s | 1.02s | 0.98 req/s | 0
c=2 | 0.97s | 1.46s | 1.37 req/s | 0
c=4 | 1.35s | 2.01s | 1.99 req/s | 0
c=8 | 3.26s | 3.96s | 2.02 req/s | 0

TTFC rises with concurrency mainly due to queueing, not slower per-request work: 0.68s at c=1 vs 3.26s at c=8. Throughput saturates around c=4 (~2 req/s) and barely increases at c=8. High-concurrency TTFC is dominated by the queue.

guozhihao-224:
## Update: causal streaming keep-set (Module TRT + hop coalesce)
### Changes
- Causal streaming (`streaming.py` / `streaming_vocoder.py`) with cross-request first/follow-up hop coalescing and a short first-hop peer wait

- `FlowEstimatorTRTModule` attach (`fallback` + `wrap_module=True`): CosyVoice takes the `nn.Module` path (profile-checked enqueue, CFG-pair chunking) instead of raw `execute_async_v3` with hard-coded CFG batch=2, which is unsafe under packed hop-batch concurrency
- TRT mel time profile max **5000**, with PyTorch DiT fallback on profile miss
- Docs and unit coverage for Module wrap and hop-batch flags

### Benchmarks — SeedTTS EN full (n=1088), H200, `stream=true`, Flow TRT,
ASR=`Qwen/Qwen3-ASR-1.7B`
| Concurrency | QPS | Latency mean | TTFP mean | TTFP p99 | WER | ok/fail |
|---:|---:|---:|---:|---:|---:|---|
| 1 | 1.59 | 0.63s | 0.27s | 0.39s | **1.62%** | 1088/0 |
| 4 | 3.59 | 1.11s | 0.47s | 0.77s | **6.88%** | 1088/0 |
| 16 | 4.13 | 3.86s | 0.82s | 2.04s | **8.81%** | 1088/0 |

**Comparisons (same dataset / ASR where applicable)**
| Configuration | C | QPS | TTFP mean | WER |
|---|---:|---:|---:|---:|
| This PR (stream + Module TRT + hop coalesce) | 16 | 4.13 | **0.82s** | 8.81% |
| non-stream TRT full (`ab_post_merge_full_en`) | 1 / 16 | 1.77 / 9.62 | n/a (buffered) | ~1.83% / ~1.84% |
| Dots stream full (`results/dots-stream-ab-full/slot`) | 1 → 16 | — | — | 1.26% → 1.33% |


### WER vs concurrency (streaming-specific)
- Not explained by ASR instability: non-stream Fun-CosyVoice3 and Dots WER stay flat across concurrency.
- Under load, `B>1` uses packed `inference_causal`; `B=1` keeps native `Flow.inference`.
- At C=16, more mid/high-WER utterances and often shorter audio vs C=1 suggest finalize/leftover starvation under one-hop-per-step scheduling.
- Dots also uses DiT, but batches isomorphic MeanFlow steps without a CosyVoice hop-pack bypass; WER remains stable.

### Follow-up roadmap
1. **Near-term A/B** — Keep first-hop coalescing for TTFP; disable follow-up hop batching and/or prioritize finalize/leftover. Success: C=16 WER near C=1, TTFP still well below the serial-hop ~4s regime.
2. **Packing limits** — Cap follow-up batch width / padding waste if (1) is insufficient.
3. **Numerical parity** — Make packed singleton ≡ native B=1 (pad/mask); remove quality gap between batched and serial hops.
4. **Longer-term** — Step-level isomorphic DiT/Euler continuous batch (Dots-like), only if (1)–(3) leave an unacceptable TTFP/WER tradeoff.
Item 

(1) is the highest-ROI check that coalescing can preserve TTFP without switching to a lower-quality numerical path after the first hop.

AkazaAkane (commented):
This changes the conditioning relative to upstream CosyVoice3. Upstream does not pad flow_prompt_speech_token / prompt_speech_feat; it keeps the prompt unchanged and adds prompt_token_pad to the first generated-token hop, i.e. the first decode waits for token_hop_len + prompt_token_pad + pre_lookahead_len generated tokens.

Here we instead repeat the last prompt token/feature up to a hop boundary and always allow the first AR flush at 28 tokens. For example, with a 10-token prompt, upstream waits for 43 generated tokens while this path synthesizes 15 additional prompt tokens/features and decodes at 28.

Since those extra repeated prompt frames become actual Flow conditioning, this is not just scheduler bookkeeping and can change the generated audio. Could we keep the prompt tensors unchanged and carry prompt_pad into the first-hop threshold, matching upstream?

PoTaTo-Mika (approved):
Streaming is a required feature for all tts models.
Though streaming introduces latency, CosyVoice still keeps real-time on most of the requests.
A further refactor on codebase should be planned and built.

PR #1664, opened 2026-08-22
[CI] Run accelerator unit tests in a dedicated pytest job

JingwenGu0829:
Supersedes #1661 with the isolation boundary moved outside pytest.

## Summary

- Keep the default `unit-test` job accelerator-free by hiding CUDA and deselecting accelerator-marked cases.
- Run accelerator-dependent tests in a separate H100 `unit-test-accelerator` job and a fresh pytest interpreter.
- Replace the hardware-vendor-specific `gpu` marker with `accelerator` across CUDA, ROCm, and XPU project manifests and tests.
- Keep host-only and mock-only XPU coverage in the default job; use backend-specific skip conditions for tests that require particular hardware.

## Why not `pytest --forked`

`pytest-forked` forks each worker only after the parent pytest process has imported conftests and collected test modules. Some of those imports already initialize CUDA, so every fork inherits poisoned CUDA runtime state and fails with `Cannot re-initialize CUDA in forked subprocess`.

A separate pytest invocation gives the accelerator suite a fresh interpreter without repeating the repository heavy import and collection cost once per test. The two unit-test jobs therefore cannot share a CUDA context, while the accelerator job remains about two minutes locally.

## XPU and other accelerators

There is no XPU CI runner or meaningful XPU-only unit-test stage today. Most tests under `tests/unit_test/xpu/` are source, script, or mocked platform tests and stay in the default job. The two fused-RoPE cases that require real XPU hardware use the common accelerator marker and skip on H100. A future XPU, NPU, MUSA, or ROCm runner can execute the same marker selection while backend-specific tests keep their own skip conditions.

## Validation

- Marker partition after rebasing onto `main`, excluding the locally unavailable optional `dots_tts` directory: 174 accelerator tests and 4,994 default tests; 46 benchmarks are excluded from both jobs.
- Accelerator suite on local H200: 172 passed and 2 XPU-only tests skipped in 135s.
- Focused default/XPU selection with `CUDA_VISIBLE_DEVICES=""`: 95 passed and 4 deselected.
- An intentionally unmarked test that allocates a CUDA tensor fails in the default environment with `RuntimeError: No CUDA GPUs are available`.
- `pre-commit run --all-files --show-diff-on-failure` passes.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -43,8 +43,42 @@ jobs:
       - name: Run test
         timeout-minutes: 30
         shell: bash
+        env:
+          # note (Jingwen): Hide CUDA from the default unit-test process. Tests
+          # that need an accelerator belong in the unit-test-accelerator job below.
+          CUDA_VISIBLE_DEVICES: ""
         run: |
           source omni/bin/activate
           export PYTHONPATH=$PWD
           bash .github/scripts/run_flaky_pytest.sh \
-            pytest tests/ -v -m "not benchmark" -x
+            pytest tests/ -v -m "not benchmark and not accelerator" -x
+
+  unit-test-accelerator:
+    name: unit-test-accelerator
+    runs-on: [self-hosted, h100]
+    container:
+      image: hongccc/sglang-omni@sha256:374d0b1c30b2bff685b1716fc64a02ad3b3d0a90fe2ce73ce9861a6992c28101
+      options: >-
+        --rm
+        -e NVIDIA_VISIBLE_DEVICES
+        -e OMNI_CI_CPUSET
+        -v /dev/shm:/dev/shm
+        -v /data/omni-ci:/data/omni-ci
+        -v /data/cache/huggingface:/github/home/.cache/huggingface
+        -v /data/cache/huggingface:/root/.cache/huggingface
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - uses: ./.github/actions/omni-setup
+        with:
+          venv-name: omni
+
+      - name: Run accelerator test
+        timeout-minutes: 30
+        shell: bash
+        run: |
+          source omni/bin/activate
+          export PYTHONPATH=$PWD
+          bash .github/scripts/run_flaky_pytest.sh \
+            pytest tests/ -v -m "accelerator and not benchmark"
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -81,4 +81,4 @@ jobs:
           source omni/bin/activate
           export PYTHONPATH=$PWD
           bash .github/scripts/run_flaky_pytest.sh \
-            pytest tests/ -v -m "accelerator and not benchmark"
+            pytest tests/ -v -m "accelerator and not benchmark" -x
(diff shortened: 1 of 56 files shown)


pyproject.toml
```diff
@@ -128,7 +128,7 @@ pythonpath = ["."]
 markers = [
     "benchmark: performance benchmark tests (may require GPU and long runtime)",
     "tts_stage(name): select an in-file TTS benchmark CI stage",
-    "gpu: requires a CUDA device (auto-skipped without one)",
+    "accelerator: requires accelerator hardware",
```
JiaxinD, line 131: tests/README.md still references the old `gpu` marker in 5 places (lines 248, 392, 551, 555, 561), including a copy-pasteable `pytest -m gpu` command that now selects nothing. Worth updating in this PR.

tests/unit_test/pipeline/test_stage.py
```diff
@@ -121,6 +121,7 @@ def _invalid_sources(request_id, from_stage, payload):
         handler.receive("req-1", "preprocess", make_stage_payload())
 
 
+@pytest.mark.accelerator
```
JiaxinD: Did this test actually fail under `CUDA_VISIBLE_DEVICES=""`? It looks fully mocked (CPU tensors, fake relay); same question for the arkasr, fun_asr, moss_transcribe_diarize and code2wav factory tests and test_stage_streaming.

.github/workflows/test.yaml
```diff
@@ -43,8 +43,42 @@ jobs:
...
+        env:
+          # note (Jingwen): Hide CUDA from the default unit-test process. Tests
+          # that need an accelerator belong in the unit-test-accelerator job below.
+          CUDA_VISIBLE_DEVICES: ""
         run: |
           source omni/bin/activate
           export PYTHONPATH=$PWD
           bash .github/scripts/run_flaky_pytest.sh \
-            pytest tests/ -v -m "not benchmark" -x
+            pytest tests/ -v -m "not benchmark and not accelerator" -x
```
JiaxinD, line 54: Note: tests/test_model/test_rl_distributed_weight_update.py carries neither marker, so neither job selects it now. It always skips today (base checkpoint absent from the CI cache), so no coverage change, but adding `accelerator` would keep the partition consistent.

.github/workflows/test.yaml
```diff
@@ -43,8 +43,42 @@ jobs:
...
+          venv-name: omni
+
+      - name: Run accelerator test
+        timeout-minutes: 30
+        shell: bash
+        run: |
+          source omni/bin/activate
+          export PYTHONPATH=$PWD
+          bash .github/scripts/run_flaky_pytest.sh \
+            pytest tests/ -v -m "accelerator and not benchmark"
```
JiaxinD: Is dropping `-x` here intentional? With run_flaky_pytest retrying any nonzero exit, a deterministic failure reruns the full suite 3 times.

.github/workflows/test.yaml
JingwenGu0829: Addressed. 

I also added a testing restriction that accelerator-dependent tests must declare the accelerator marker unconditionally, while hardware availability checks must happen inside the test body or a fixture rather than in a module-level skipif. 

This will prevents such tests from being silently selected and skipped only in the default job.

JiaxinD (approved):

Ratish1 (approved):
LGTM

PR #1665, opened 2026-08-22
[Perf][MOSS-TTS] Fuse the 32 audio LM heads into one stacked GEMM

JiaxinD:
## Motivation

Each MOSS-TTS delay decode step runs 33 `LogitsProcessor` calls (one per channel) plus a 32-way fp32 cast and stack, roughly 215 tiny eager launches out of the step's ~340; a 12 s c=8 serving window shows 37k eager launches per second at 78% inter-kernel bubble. The 32 audio heads share one padded [1088, 4096] weight shape, so all of that collapses into one GEMM.

## Modifications

* Re-point the 32 audio head weights at slices of one stacked buffer at the end of `load_weights`, before the weight-share export and before KV profiling (no steady-state memory; the transient duplicate lives until the originals release, ~272 MiB peak on the BF16 checkpoint, and lands inside the load budget rather than on a first request).
* One `F.linear` emits [B, 32, 1088]; a view slices the vocab padding and a single cast produces the [B, n_vq, 1025] fp32 tensor both sampler paths consume via `ChannelLogitsList.fused_audio`, skipping the per-channel cast and stack. The text head (152k vocab) keeps its `LogitsProcessor` path.
* Capability gate: TP1, unquantized same-shape `ParallelLMHead` weights, equal audio vocabularies, no logit softcapping; anything else falls back to the per-head path with a log line.
* Staleness guard: expected slice pointers are precomputed at stack time and every use validates all 32 heads, falling back to the per-head path if any weight was replaced (`set_embed_and_head`, assign-loads).
* Same-GPU weight sharing (`WEIGHT_SHARE=1`): a follower aliases every head onto the leader's storage after load, which would trip that guard and leave the follower on the per-head GEMM while the leader stays fused, breaking the byte-identity contract in `docs/basic_usage/mps_dp.md`. The leader stacks before it exports, so the shared heads arrive as one contiguous block; a new post-attach hook on the runner lets the model rebuild the stacked view over that block (a view, not a copy), and it fails closed with an actionable message if the block is not contiguous.
* `MOSS_DELAY_FUSED_AUDIO_HEADS=0` restores the per-head path.
* Not bit-exact vs the per-head GEMMs (different tiling); gated by the quality evidence below.

## Accuracy Test

* WER, 200 SeedTTS EN samples per arm (seed 777): fused 0.664% vs base 0.807%.
* Speaker similarity, same 200 samples: fused 64.7499 vs base 64.9719 (delta -0.34%).
* Same-arm determinism: 3 fixed-seed WAV sha256 identical across two fresh fused servers.
* Leader/follower byte identity, H200, MOSS delay `N=2` `WEIGHT_SHARE=1` under a private MPS daemon: two seeded first-traffic pairs produce identical WAV sha256 on leader and follower. Without the post-attach hook both pairs differ (the follower logs no fused line and its audio is a different length), so the contract is measured, not argued.
* Fused vs per-head logits parity within tolerance at the production shape ([16, 512] hidden, 12 heads of 1032x512, bf16 and fp16); bit equality is not expected across the two GEMM layouts.
* `tests/unit_test/moss_tts/test_fused_audio_heads.py` plus the weight-share load-path suite: replacing any non-first audio head after fusion disables the fused path, the follower adopts the shared block, and a non-contiguous block raises (372 passed across `moss_tts`, `model_runner` and `test_ipc_weights` at the rebased head).

## Benchmark & Profiling

`benchmark_tts_seedtts --generate-only --stream`, SeedTTS EN 48 samples per cell, H200, per-cell host load logged (mixed 19-38; in round 2 the base arm ran at lower load and still lost):

| conc | base qps (3 rounds) | fused qps (3 rounds) | mean change | lat/RTF |
|---:|---|---|---:|---|
| 1 | 0.978 / 0.952 / 0.988 | 1.023 / 1.040 / 1.036 | +6.2% | RTF -4% to -6% |
| 8 | 1.653 / 1.596 / 1.691 | 1.821 / 1.817 / 1.876 | +11.6% | lat -9% to -12% |
| 16 | 1.900 / 1.856 / 1.954 | 2.042 / 2.020 / 2.089 | +7.7% | RTF -7% to -10% |

9/9 paired wins, plus 2 earlier c1/c8 rounds (+4.9% to +15.0%) for 13/13 total. An exact-head canary after the staleness-guard commit measured c8 +9.6% with the fused arm at higher host load. Launch-collapse attribution is design-inferred (33 processor calls plus 32 casts and a stack become 2 kernels); no profiler counter is claimed.

A post-rebase c8 canary measured fused 1.704 vs base 1.464 qps (+16.4%; host load 27 on the fused round vs 39 on the base round, logged per cell as in the main matrix).

The matrix above still applies to the current head. Moving the stack from the first request into `load_weights` changes when the buffer is built, not the steady-state decode path, and the weight-share hook only runs on a follower at attach time; neither touches the leader's serving loop. Rebased onto current main, zero conflicts, unit suites green at the rebased head.

## Reproduction

```bash
MOSS_DELAY_FUSED_AUDIO_HEADS={1|0} sgl-omni serve --model-path OpenMOSS-Team/MOSS-TTS-v1.5 \
  --config examples/configs/moss_tts.yaml --port 8690
python -m benchmarks.eval.benchmark_tts_seedtts --generate-only --use-existing-server --stream \
  --model OpenMOSS-Team/MOSS-TTS-v1.5 --port 8690 --ref-format references --token-count auto \
  --max-concurrency {1|8|16} --lang en --max-samples 48 --seed 555 --output-dir out
python -m pytest tests/unit_test/moss_tts/test_fused_audio_heads.py -q
```


sglang_omni/models/moss_tts/sglang_model.py
```diff
@@ -363,17 +375,117 @@ def compute_channel_outputs(
...
+                "MOSS-TTS fused audio heads disabled (unsupported configuration)"
+            )
+            return False
+        # Note (Jiaxin Deng): re-point each head's weight at a slice of one
+        # stacked buffer so the fused GEMM adds no steady-state memory; the
+        # transient duplicate lives only until the originals are released.
+        stacked = torch.cat([w.data for w in weights], dim=0).contiguous()
+        rows = int(first.shape[0])
+        for index, head in enumerate(self.lm_heads[1:]):
+            head.weight.data = stacked[index * rows : (index + 1) * rows]
```
charliechenye, line 462: Blocking: This `.data` rebind appears incompatible with Omni's same GPU weight sharing contract. `MossTTSDelaySGLangModel` is a supported weight sharing architecture, and the leader exports the LM head storages through CUDA IPC at the end of `load_model()`. The runner explicitly verifies that those storages are not rebound after export because doing so would orphan follower aliases.

Here the audio heads are rebound lazily when `compute_channel_logits()` first reaches this path, which is after weight share export/attach and the pre capture attachment check.

sglang_omni/models/moss_tts/sglang_model.py
```diff
@@ -363,17 +375,117 @@ def compute_channel_outputs(
...
+        self._fused_audio_heads_enabled = enabled
+        if not enabled:
+            logger.info(
+                "MOSS-TTS fused audio heads disabled (unsupported configuration)"
+            )
+            return False
+        # Note (Jiaxin Deng): re-point each head's weight at a slice of one
+        # stacked buffer so the fused GEMM adds no steady-state memory; the
+        # transient duplicate lives only until the originals are released.
+        stacked = torch.cat([w.data for w in weights], dim=0).contiguous()
```
charliechenye, line 459: Can we move this transformation out of the request path and perform it before KV cache sizing?

Today `_ensure_stacked_audio_heads()` is reached lazily from `compute_channel_logits()`, while Omni allocates/profiles the KV pool during infrastructure construction before normal serving begins. This means the ~272 MiB transient allocation described in the PR can occur only after the KV pool has already consumed the available memory budget.

That is particularly risky for the supported 24 GB / 32 GB MOSS configurations. Even though steady state memory is approximately unchanged, a first request transient allocation can still OOM.

sglang_omni/models/moss_tts/sglang_model.py
```diff
@@ -363,17 +375,117 @@ def compute_channel_outputs(
...
+            # audio heads, so it is gated to the plain configuration it
+            # reproduces: TP1, unquantized same-shape ParallelLMHead weights,
+            # equal audio vocabularies, no logit softcapping.
+            weights = [getattr(head, "weight", None) for head in self.lm_heads[1:]]
+            first = weights[0] if weights else None
+            enabled = (
+                first is not None
+                and first.ndim == 2
+                and first.dtype in (torch.bfloat16, torch.float16, torch.float32)
+                and get_tensor_model_parallel_world_size() == 1
```
charliechenye: I don't think this capability gate is strong enough to establish equivalence with the `LogitsProcessor` path we are bypassing.

In the pinned SGLang version, LM head execution depends on runtime state in addition to the module type, weight shape, and `dtype`. For example, `--enable-fp32-lm-head` casts both hidden states and weights to `fp32` _before_ GEMM, and there are separate quant method, CPU packed weight, and RL target paths. The fused path instead always calls raw `F.linear` and casts the output afterward.

Could we gate on the actual plain LM head execution mode rather than infer it from `ParallelLMHead` + dtype?

charliechenye (changes requested):
Thanks for the detailed profiling and validation here. I think the optimization itself is well motivated: collapsing the 32 audio LM heads is a sensible way to attack this launch-bound decode path, and the repeated c1/c8/c16 wins are strong enough that I think this is worth landing.

I am requesting changes on the implementation contract rather than the optimization direction.

My blocking concerns are:

1. The lazy `head.weight.data` rebinding conflicts with Omni's supported CUDA-IPC weight sharing contract the model.
2. The ~272 MiB stacking allocation currently happens lazily after KV-cache memory has already been sized, which can break the constrained memory configurations even if steady state memory is unchanged.
3. The capability gate does not fully characterize the `LogitsProcessor` execution semantics being bypassed. In particular, runtime fp32 LM head / special LM head execution paths can pass the current shape/type checks while using different math.

sglang_omni/models/moss_tts/sglang_model.py
JiaxinD: Fixed: stacking moved to the end of load_weights, so the IPC export and the pre-capture attachment check both see the final stacked storages; the request path only observes the result. A WEIGHT_SHARE=1 end-to-end revalidation is queued.

sglang_omni/models/moss_tts/sglang_model.py
JiaxinD: Fixed by the same move: the transient copy now lands before KV profiling, inside the budget. Verified on H200: the enabled log appears during load, ten seconds before serving starts.

sglang_omni/models/moss_tts/sglang_model.py
JiaxinD: Done: the gate now mirrors the pinned _compute_lm_head branch order (LoRA wrap, applicable quant method, Intel AMX packed path, fp32 lm head, RL on-policy target all fall back to LogitsProcessor), with a unit test per branch.

db-ol:
Two things on the current head, plus one suggestion.

1. `test_moss_text_control_logits_select_from_full_processor_output` calls `MossTTSDelaySGLangModel.compute_channel_logits()` unbound on a `SimpleNamespace` stub. `compute_channel_logits()` now calls `self._fused_audio_heads_ready()` first, which the stub does not provide, so the test raises `AttributeError`. Updating the stub, or keeping that fallback path test independent of the fused head state, should fix it.

2. The `WEIGHT_SHARE=1` follower path. Stacking now happens before the IPC export, which settles the original rebind concern for the leader. A follower later replaces every `lm_heads[*].weight` in `set_embed_and_head()`, and `_fused_audio_heads_ready()` compares each head against `_audio_head_expected_ptrs` from the local stack, so on the follower the guard fires and fusion is disabled for good. That would leave the leader on the fused GEMM and the follower on the per head path. Since this PR does not claim bit exact parity between those two GEMM layouts, this looks relevant to the existing `WEIGHT_SHARE=1` leader/follower byte identity contract (docs/basic_usage/mps_dp.md records MOSS delay `N=2` support as byte identical audio between leader and follower). Either keeping the post attach paths symmetric or disabling fusion when sharing is on would keep that contract, and the queued `N=2` byte identity run would settle it.

Separately, a tolerance based logits parity test between the fused and per head paths would catch indexing and padding regressions. Bit exact equality is not expected given the different GEMM tiling.

JiaxinD:
Both fixed. The follower now rebuilds the stacked view over the leader's shared block from a post attach hook and fails closed when that block is not contiguous; the `test_pipeline` stub provides the readiness check.

Note: (H200, N=2, `WEIGHT_SHARE=1`) leader and follower audio is byte identical on two seeded first traffic pairs, and differs on both without this change. A fused vs per head tolerance parity test is added.

charliechenye:
lgtm. Will approve after rebase + CI run :) Thanks a lot

JiaxinD:
Rebased onto current main (zero conflicts, 372 passed across `moss_tts`, `model_runner` and `test_ipc_weights` at the rebased head `c0a853af`). The body now describes the weight-share follower path and carries the `N=2` byte-identity result, and drops the stale "permanently falling back" wording.

Note: (the throughput matrix predates the review fixes and still applies; moving the stack into `load_weights` and adding the follower attach hook leaves the leader's serving loop untouched. Needs the `run-ci` label from a maintainer.)

db-ol (approved):

charliechenye (approved):

PR #1636, opened 2026-08-20
[Refactor][MOSS-TTS] Consolidate the MOSS-Audio-Tokenizer attention backend

CloudRipple:
## Motivation

MOSS-TTS Delay and MOSS-TTS Local execute the same non-streaming MOSS-Audio-Tokenizer local-causal decoder attention, but the repository did not have a single, explicit backend policy for it. Backend selection, packed-sequence preparation, RoPE handling, and model-layout adaptation were coupled in the same implementation. The packed path was also tied to the Hugging Face `flash_attention_2` configuration name, making its behavior difficult to reason about across CUDA versions, dtypes, and GPU architectures.

We compared the available implementations and execution strategies for this workload: FlashInfer FA2, SGLang packed FlashAttention, full-sequence versus 128-token tiled packed execution, and query-chunked SDPA versus HF duration-based streaming. The comparison showed that one backend should not be forced across all environments: packed FlashAttention is the fast path when supported, 128-token tiling is required for the `SM < 90` path, and query-chunked SDPA is the portable fallback. FlashInfer did not provide a demonstrated end-to-end benefit, so it is not added to the runtime policy.

This PR consolidates the shared MOSS-Audio-Tokenizer attention runtime and makes the selection policy explicit and testable. It preserves the decoder math and source weights, supports both v1 and v2 module layouts, and removes redundant Q/K/V contiguous copies.

## Modifications

- Add `sglang_omni/models/moss_tts/attention.py` with the shared MOSS-Audio-Tokenizer attention implementation:
  - explicit `auto`, `packed_flash_attention`, and `sdpa` backend policies;
  - SGLang's `_is_fa3_supported()` predicate for packed-kernel eligibility;
  - BF16-only packed FlashAttention eligibility on CUDA (FP16 is intentionally not added);
  - query-chunked SDPA fallback for CPU, FP32, explicit SDPA, or unavailable packed kernels;
  - per-device caching for the SM-based direct-versus-tiled routing decision;
  - packed/unpacked sequence helpers, cached packed RoPE, and the equivalent local-causal FlashAttention window `(context - 1, 0)`;
  - direct Q/K/V views passed to the packed kernel, without redundant `.contiguous()` copies.
- Simplify `audio_tokenizer.py` into the shared transformer and decoder wrappers plus codec loading:
  - use explicit `from_module()` constructors that reuse the source modules;
  - support both MOSS-Audio-Tokenizer v1 weight fields and v2 `ffn` fields;
[...]

PR #1666, opened 2026-08-22
[Perf][MOSS-TTS-Local] Fuse the seeded frame sampler into one Triton kernel

JiaxinD:
## Motivation

MOSS-TTS-Local samples 13 heads per generated frame (1 stop head at vocab 2 plus 12 codebook heads at vocab 1024, the checkpoint's `audio_vocab_size`; the embedding table carries a 1025th pad row that the head weight slices off). Each call runs `sample_seeded_branchless`: a full-vocabulary descending sort, two softmaxes, a cumsum, a scatter, and the seeded Gumbel draw, about 14 kernels per call and ~180 sampler kernels per frame. In a 12 s c=16 serving profile the sampling block is ~13.5% of per-frame GPU time (1.3-1.5 ms per frame under CUDA graph replay), with `radixSortKVInPlace` alone at 3.9% of total GPU time.

## Modifications

* Add `sample_seeded_fused` to `sglang_omni/models/moss_tts/sampling_kernels.py`: one Triton program per row performs temperature scaling, a stable descending sort (scores packed above a complemented index so equal scores keep input order exactly like the stable radix sort `torch.sort` uses; signed zeros canonicalize to +0.0 first so numerically equal zeros share one key), the top-k threshold, nucleus removal via exclusive cumsum, and the exact `multinomial_with_seed` hash and fp64 Gumbel math, including the NaN-first and greedy-fallback semantics.
* Route the MOSS-TTS-Local frame sampler through it for vocab <= 2048; larger vocabularies keep a lazily compiled branchless path. `MOSS_LOCAL_FUSED_FRAME_SAMPLER=0` restores the previous behavior.
* Give every path the same tie order: `sample_seeded_branchless` and `MossTTSModelRunner._apply_top_p` now sort stable, which is the order the fused key already keeps. `_run_frame_decode` takes the eager `_sample_tokens` path when an audio repetition penalty is set or the batch outgrows the frame graph, and on the `vocab=2` stop head an unstable sort resolved a tie the other way, which under nucleus sampling is continue versus stop inside one server.
* Cross-module note: the kernel lives in the shared `moss_tts` package next to the existing sampling kernels. `sample_seeded_branchless` is imported only by `moss_tts_local`, but `_apply_top_p` is shared with MOSS-TTS delay, whose graphed and eager paths both call it; delay stays self consistent either way and only gains a defined tie order.

## Accuracy Test

* Contract: numerically equivalent to the previous sampler, not bit identical by construction. The tie order, the greedy and NaN fallbacks, and the seeded draw are exact; the in-kernel softmax and cumsum reduce in a different order than aten, so at an exact nucleus boundary the kept set may differ by the boundary token.
* 13,950 randomized draws across the parity grid come out bit for bit equal to the previous sampler: the contract above allows divergence only at an exact nucleus boundary, and no randomized draw landed on one.
* Tied `vocab=2` stop head at `top_p=0.3`: fused, branchless and eager `_sample_tokens` agree over 16 seeds. The test fails without the stable-sort change, where the eager path selects the other token.
* NaN parity: `torch.argmax` semantics reproduced for one NaN, all NaN, and NaN mixed with +inf, at temperature 0 and 1. Without the fallback fix an all-NaN greedy row returns the reduction sentinel, which is not a valid token id.
* Same-seed end-to-end check: 6 fixed-seed `/v1/audio/speech` requests produce byte-identical WAVs (equal sha256) on the fused and baseline servers.
* Constructed edge tests: nucleus knife edges (bit-equal strictly between cumulative boundaries), signed zeros in both input orders at a discriminating nucleus boundary, nonfinite rows, the UINT32_MAX hash endpoint (including a top-k-masked lane, where the fused kernel reproduces the baseline's NaN-first selection exactly), strided/empty/typed inputs.
* `tests/unit_test/moss_tts/test_sampling_kernels.py`: 35 passed on H200; the `moss_tts` suite is 208 passed and `moss_tts_local` plus `model_runner` plus `scheduling` is 506 passed.
* The nucleus-boundary divergence above is what the knife edge test encodes: equality strictly between cumulative boundaries, top-k membership at the boundary itself. No divergence was observed in any randomized, constructed, or end-to-end comparison.
* SeedTTS EN corpus WER on saved c=16 audio (64 samples per arm, two rounds each): fused 0.996% / 0.996%, base 1.138% / 1.422%.

## Benchmark & Profiling

Sampler block, production frame pattern (1x[bs,2] + 12x[bs,1024], the checkpoint's audio head width), CUDA events, 5-round median, H200:

| bs | mode | base (us) | fused (us) | speedup |
|---:|---|---:|---:|---:|
| 1 | graph replay | 3447.7 | 167.8 | 20.55x |
| 8 | graph replay | 1394.3 | 169.8 | 8.21x |
| 16 | graph replay | 1475.5 | 169.8 | 8.69x |
| 1-16 | eager | 6790-7046 | ~501 | ~13.9x |

At `vocab=1025` the fused kernel takes `BLOCK=2048` and lands at 565-568 us instead, so the same rows read 6.37x / 2.55x / 2.64x. That is the power-of-two boundary stress case; the production audio head is 1024 wide.

End to end, `benchmark_tts_seedtts --generate-only --stream`, SeedTTS EN 48 samples per cell, warmed rounds 1-3 means (rounds 4-5 confirm):

| conc | metric | base | fused | change |
|---:|---|---:|---:|---:|
| 1 | qps | 1.2717 | 1.3153 | +3.4% |
| 1 | RTF mean | 0.1899 | 0.1834 | -3.5% |
| 8 | qps | 6.251 | 6.444 | +3.1% |
| 8 | RTF mean | 0.2875 | 0.2773 | -3.6% |
| 16 | qps | 7.583 | 8.024 | +5.8% |
| 16 | RTF mean | 0.4484 | 0.4239 | -5.5% |
| 16 | TTFA median (s) | 0.7556 | 0.6964 | -7.8% |

Across 5 server-restart rounds x 3 concurrencies the fused arm wins all 15 paired qps comparisons (margins +0.1% to +10.6%; no independence assumption or p-value claimed). A symmetric cold-start pair (fresh server, c16 measured first) also favors fused: 6.553 vs 6.377 qps (+2.8%). An independent c1 raw-request experiment reproduced -5.9% time per audio byte across 3 further ABAB rounds. One extra confirmation round taken under heavy co-tenant host load (33-38 vs 20-25 during the matrix) was discarded as contaminated and its raw data retained.

E2E numbers predate the review fixes and still stand: both arms were re-measured at the current head in one clean window. The fused arm costs +1.5 to +1.7 us per frame pattern (+0.3% of the sampler block, about +0.02% of per-frame GPU time) from the NaN and fallback branches, and the stable sort costs nothing on the baseline arm (base column identical to the digit; bare `torch.sort` is +0.0% at [16, 1024] and faster at [16, 2], both launch bound). Six fixed-seed `/v1/audio/speech` requests are byte identical across the review fixes, so serving output is unchanged. Rebased onto #1636 (touches the audio tokenizer, not this path); the full unit suite passes at the rebased head.

## Reproduction

```bash
MOSS_LOCAL_FUSED_FRAME_SAMPLER={1|0} sgl-omni serve \
  --model-path OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5 --port 8640
python -m benchmarks.eval.benchmark_tts_seedtts --generate-only --use-existing-server --stream \
  --model OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5 --port 8640 --ref-format references \
  --token-count auto --max-concurrency {1|8|16} --lang en --max-samples 48 --output-dir out
python -m pytest tests/unit_test/moss_tts/test_sampling_kernels.py -q
```

### DIFF
--- a/sglang_omni/models/moss_tts/sampling_kernels.py
+++ b/sglang_omni/models/moss_tts/sampling_kernels.py
@@ -151,6 +151,9 @@ def multinomial_with_seed_and_token_ids(
 
 
 _F64_MIN = tl.constexpr(-1.7976931348623157e308)
+# Note (Jiaxin Deng): sglang caps log(uniform) at both ends, so a hash of
+# UINT32_MAX no longer yields +inf gumbel; the fused draw follows the cap.
+_F64_LOG_CAP = tl.constexpr(-(2.0**-32))
 
 # Note (Jiaxin Deng): single-block cap; larger vocabularies keep the sort-based
 # path because one program can no longer hold the row.
@@ -243,9 +246,8 @@ def _fused_seeded_sample_kernel(
     hash_value ^= 16
     hash_value = fmix32(hash_value)
     uniform = hash_value.to(tl.float64) / _UINT32_MAX_F64
-    log_uniform = libdevice.log(uniform)
-    neg_log_uniform = -tl.maximum(log_uniform, _F64_MIN)
-    gumbel = -libdevice.log(neg_log_uniform)
+    log_uniform = tl.minimum(tl.maximum(libdevice.log(uniform), _F64_MIN), _F64_LOG_CAP)
+    gumbel = -libdevice.log(-log_uniform)
     value = final_sorted.to(tl.float64) + gumbel
 
     is_nan = value != value
--- a/sglang_omni/models/moss_tts/sampling_kernels.py
+++ b/sglang_omni/models/moss_tts/sampling_kernels.py
@@ -282,7 +282,14 @@ def sample_seeded_fused(
     seeds: torch.Tensor,
     positions: torch.Tensor,
 ) -> torch.Tensor:
-    """Single-kernel drop-in for :func:`sample_seeded_branchless` (vocab <= 2048)."""
+    """Single-kernel seeded sampler for vocab <= 2048.
+
+    Numerically equivalent to :func:`sample_seeded_branchless` rather than bit
+    identical by construction: the in-kernel softmax and cumsum reduce in a
+    different order than aten, so at an exact nucleus boundary the kept set may
+    differ by the boundary token. The tie order (input order), the greedy and
+    NaN fallbacks, and the seeded draw are exact.
+    """
 
     rows, vocab = logits.shape
     if vocab > MAX_FUSED_SAMPLE_VOCAB:
--- a/sglang_omni/models/moss_tts/sampling_kernels.py
+++ b/sglang_omni/models/moss_tts/sampling_kernels.py
@@ -218,10 +218,12 @@ def _fused_seeded_sample_kernel(
     p_active = (top_p > 0.0) & (top_p < 1.0)
     row_max = tl.max(masked_sorted, axis=0)
     finite_max = row_max > -float("inf")
+    # Note (Jiaxin Deng): masking on == -inf lets NaN and +inf lanes poison z the
+    # way torch.softmax poisons the baseline row, so both fall back identically.
     exp_term = tl.where(
-        masked_sorted > -float("inf"),
-        tl.exp(masked_sorted - tl.where(finite_max, row_max, 0.0)),
+        masked_sorted == -float("inf"),
         0.0,
+        tl.exp(masked_sorted - tl.where(finite_max, row_max, 0.0)),
     )
     z = tl.sum(exp_term, axis=0)
     probs_sorted = tl.where(z > 0, exp_term / z, 0.0)
@@ -256,9 +258,17 @@ def _fused_seeded_sample_kernel(
     )
     sampled = tl.where(nan_index != 2147483647, nan_index, max_index)
 
-    max_logit = tl.max(tl.where(valid, logits, -float("inf")), axis=0)
-    greedy = tl.min(tl.where(valid & (logits == max_logit), idx, 2147483647), axis=0)
-    use_fallback = (~do_sample) | (z <= 0)
+    # Note (Jiaxin Deng): torch.argmax ranks NaN above every number, so a NaN row
+    # must select its first NaN lane; comparing against a NaN max matches none.
+    logit_is_nan = logits != logits
+    nan_greedy = tl.min(tl.where(valid & logit_is_nan, idx, 2147483647), axis=0)
+    finite = valid & ~logit_is_nan
+    max_logit = tl.max(tl.where(finite, logits, -float("inf")), axis=0)
+    finite_greedy = tl.min(
+        tl.where(finite & (logits == max_logit), idx, 2147483647), axis=0
+    )
+    greedy = tl.where(nan_greedy != 2147483647, nan_greedy, finite_greedy)
+    use_fallback = (~do_sample) | (z <= 0) | (z != z)
     result = tl.where(use_fallback, greedy.to(tl.int64), sampled)
     tl.store(output_ptr + row, result)
 
@@ -325,7 +335,11 @@ def sample_seeded_branchless(
 
     k_active = (top_k > 0) & (top_k < vocab)
     k_clamped = top_k.clamp(min=1, max=vocab)
-    sorted_scores, sorted_indices = torch.sort(scores, descending=True, dim=-1)
+    # Note (Jiaxin Deng): an unstable sort leaves the tie order backend dependent
+    # (a two lane row reverses on cu130) and that picks the token at a nucleus edge.
+    sorted_scores, sorted_indices = torch.sort(
+        scores, descending=True, dim=-1, stable=True
+    )
     kth = sorted_scores.gather(1, (k_clamped - 1).unsqueeze(1))
     threshold = torch.where(
         k_active.unsqueeze(1), kth, torch.full_like(kth, float("-inf"))
--- a/sglang_omni/models/moss_tts/sampling_kernels.py
+++ b/sglang_omni/models/moss_tts/sampling_kernels.py
@@ -276,16 +276,21 @@ def sample_seeded_fused(
         raise ValueError(
             f"fused seeded sampler supports vocab <= {MAX_FUSED_SAMPLE_VOCAB}, got {vocab}"
         )
+    if rows == 0:
+        return torch.empty(0, device=logits.device, dtype=torch.int64)
+    if top_k.dtype.is_floating_point or top_k.dtype == torch.bool:
+        raise TypeError(f"top_k must be an integer tensor, got {top_k.dtype}")
+    top_k = top_k.to(torch.int64)
     logits = logits.float().contiguous()
     out = torch.empty(rows, device=logits.device, dtype=torch.int64)
     block = triton.next_power_of_2(vocab)
     _fused_seeded_sample_kernel[(rows,)](
         logits,
-        temperature.float(),
-        top_p.float(),
-        top_k.to(torch.int64),
-        seeds,
-        positions,
+        temperature.float().contiguous(),
+        top_p.float().contiguous(),
+        top_k.contiguous(),
+        seeds.to(torch.int64).contiguous(),
+        positions.to(torch.int64).contiguous(),
         out,
         VOCAB=vocab,
         ROW_STRIDE=logits.stride(0),
--- a/sglang_omni/models/moss_tts/sampling_kernels.py
+++ b/sglang_omni/models/moss_tts/sampling_kernels.py
@@ -278,7 +278,7 @@ def sample_seeded_fused(
         )
     if rows == 0:
         return torch.empty(0, device=logits.device, dtype=torch.int64)
-    if top_k.dtype.is_floating_point or top_k.dtype == torch.bool:
+    if top_k.dtype.is_floating_point or top_k.dtype.is_complex or top_k.dtype == torch.bool:
         raise TypeError(f"top_k must be an integer tensor, got {top_k.dtype}")
     top_k = top_k.to(torch.int64)
     logits = logits.float().contiguous()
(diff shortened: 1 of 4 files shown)


sglang_omni/models/moss_tts/sampling_kernels.py
```diff
@@ -145,6 +145,163 @@ def multinomial_with_seed_and_token_ids(
...
+    nan_index = tl.min(tl.where(s_valid & is_nan, s_idx, 2147483647), axis=0)
+    non_nan = tl.where(s_valid & ~is_nan, value, -float("inf"))
+    max_value = tl.max(non_nan, axis=0)
+    max_index = tl.min(
+        tl.where(s_valid & ~is_nan & (non_nan == max_value), s_idx, 2147483647),
+        axis=0,
+    )
+    sampled = tl.where(nan_index != 2147483647, nan_index, max_index)
+
+    max_logit = tl.max(tl.where(valid, logits, -float("inf")), axis=0)
```
charliechenye: I don't think the greedy fallback is parity safe for `NaN` logits.

The baseline fallback is `torch.argmax(logits, dim=-1)`, while this path first reduces `max_logit` and then finds lanes satisfying `logits == max_logit`. `NaN` never compares equal, so an all NaN fallback row can leave the sentinel as the selected index instead of reproducing `torch.argmax`.

Since this PR explicitly claims to preserve NaN and greedy fallback semantics, please reproduce the baseline behavior here and add parity cases for `temperature=0` with one `NaN`, all `NaN`, and `NaN` mixed with `+inf`.

I would argue on a default on sampling path, an incorrect fallback can produce an invalid token ID.

tests/unit_test/moss_tts/test_sampling_kernels.py
```diff
@@ -74,3 +74,365 @@ def test_seeded_gumbel_argmax_rejects_strided_output() -> None:
...
+        top_k,
+        tie,
+        id=f"v{vocab}-t{temp}-p{top_p}-k{top_k}-{'tied' if tie else 'random'}",
+    )
+
+
+@pytest.mark.parametrize(
+    "vocab,temp,top_p,top_k,tie",
+    [
+        _fused_case(1025, 1.7, 0.8, 25, False),
```
charliechenye, line 117: I think the parity matrix is missing the actual production audio vocab specialization.

The `OpenMOSS-Team/MOSS-TTS-Local-Transformer-v1.5` checkpoint declares `n_vq=12`, `audio_vocab_size=1024`, and twelve `audio_codebook_sizes=1024`. Omni does allocate 1025 row audio embedding tables because row 1024 is the extra pad row, but `_audio_embedding_weight()`explicitly returns `weight[:config.audio_vocab_size]`. `_decode_frame_graphable()` then computes `F.linear(current, head_weight)`, so the logits passed to the sampler are `[bs, 1024]`, not `[bs, 1025]`.

This matters because `next_power_of_2(1024)` uses BLOCK=1024, while vocab 1025 exercises `BLOCK=2048`. I would keep 1025 as a useful boundary case, but add 1024 as the primary production parity specialization.

charliechenye (changes requested):

db-ol:
One parity gap beyond the open NaN and vocab 1024 points, on the production `vocab=2` stop head.

The baseline calls `torch.sort(..., descending=True)` without `stable=True`. The fused key deliberately keeps input order for equal scores (the note at the key construction says it matches cub's stable radix sort). The two agree on large rows but not on the two lane row: on the pinned stack (H200, torch 2.11.0+cu130) `torch.sort` orders `[[2.0, 2.0]]` as `[1, 0]`, while the fused key orders it `[0, 1]`.

This is observable under nucleus sampling. With logits `[[2.0, 2.0]]` and `top_p=0.3` only the first sorted lane survives, so the two orders select different tokens, and on the stop head that is continue versus stop. The current parity grid has a `vocab=2` case but not a tied one.

A direct fused versus branchless parity case for equal logits on the two token head would pin this down, with the fused path reproducing the baseline order on the pinned stack, or the baseline made stable if that order is the intended one.

JiaxinD:
All three fixed. The greedy fallback now reproduces `torch.argmax` on NaN rows (NaN and +inf rows also take the fallback the baseline takes), and the baseline sort is stable, so the tied `vocab=2` order is `[0, 1]` on both paths.

Parity cases added for NaN at temperature 0 and 1, the tied stop head at `top_p=0.3`, and vocab 1024 as the primary specialization; all of them fail on the previous head.

sglang_omni/models/moss_tts/sampling_kernels.py
```diff
@@ -163,7 +330,11 @@ def sample_seeded_branchless(
 
     k_active = (top_k > 0) & (top_k < vocab)
     k_clamped = top_k.clamp(min=1, max=vocab)
-    sorted_scores, sorted_indices = torch.sort(scores, descending=True, dim=-1)
+    # Note (Jiaxin Deng): an unstable sort leaves the tie order backend dependent
+    # (a two lane row reverses on cu130) and that picks the token at a nucleus edge.
+    sorted_scores, sorted_indices = torch.sort(
```
charliechenye, line 349: Blocking: I don't think making `sample_seeded_branchless` stable is sufficient to resolve the tied logit parity issue, because it changes the reference path rather than making all MOSS-TTS-Local execution paths agree.

The eager fallback still goes through `MossTTSModelRunner._sample_tokens()`, whose top-p path uses `torch.sort(..., descending=True)` without `stable=True`. `_run_frame_decode()` switches to that eager path when the frame graph is unavailable, including when an audio repetition penalty is enabled.

That makes the production `vocab=2` case especially important: with tied stop logits and `top_p=0.3`, the new stable graph path can keep token 0 while the eager fallback on the pinned stack can keep token 1, even though the text logits and text sampling parameters are unchanged.

This also means `MOSS_LOCAL_FUSED_FRAME_SAMPLER=0` no longer literally restores the previous sampler behavior.

Could we please define one tie order contract across the fused, branchless, and eager fallback paths, and add a direct tied stop head parity test against the actual eager `_sample_tokens()` path? I don't think fused vs modified branchless parity alone protects the serving contract here.

Thanks a lot.

sglang_omni/models/moss_tts/sampling_kernels.py
```diff
@@ -145,6 +145,173 @@ def multinomial_with_seed_and_token_ids(
...
+def sample_seeded_fused(
+    logits: torch.Tensor,
+    *,
+    temperature: torch.Tensor,
+    top_p: torch.Tensor,
+    top_k: torch.Tensor,
+    seeds: torch.Tensor,
+    positions: torch.Tensor,
+) -> torch.Tensor:
+    """Single-kernel drop-in for :func:`sample_seeded_branchless` (vocab <= 2048)."""
```
charliechenye: I think the contract here is still stronger than what the implementation and tests guarantee.

We call this a drop-in replacement and the PR emphasizes bit identical parity, but the PR also explicitly allows the Triton softmax/cumsum reduction order to differ from ATen at exact top p boundaries, and the knife edge test intentionally does not require equality at that boundary.

I don't think we need to reproduce ATen's reduction order at the cost of a much more complicated kernel. A numerically equivalent sampler may be the right engineering tradeoff. But if divergence is intentionally allowed, we should state that contract explicitly instead of calling it deterministic drop in parity.

Could we relax the doc & PR wording please and validate the resulting numerical equivalence contract accordingly? Thanks a lot.

charliechenye (changes requested):

tests/unit_test/moss_tts/test_sampling_kernels.py
charliechenye: The new vocab=1024 unit coverage resolves my earlier specialization concern. Please also correct the PR body's production pattern from 1025 to 1024 and refresh the sampler microbenchmark at 1024; 1025 is still useful as a power of two boundary stress case, but it is not the actual audio head shape.

sglang_omni/models/moss_tts/sampling_kernels.py
JiaxinD: Confirmed and fixed. `_apply_top_p` now sorts stable, so the graphed, branchless and eager paths share input order as the contract, and the tied stop head test calls the eager `_sample_tokens` path directly; it fails on the previous head, where eager selected token 1 and fused token 0.

Note: (`_apply_top_p` is shared with MOSS-TTS delay, whose graphed and eager paths both call it, so delay only gains a defined tie order; disclosed in the body. `_apply_top_k` and `_apply_two_token_top_k` are threshold based and keep every tied lane, so they need no change.)

sglang_omni/models/moss_tts/sampling_kernels.py
JiaxinD: Agreed, relaxed. The docstring and the PR body now state numerical equivalence with the nucleus-boundary divergence spelled out, instead of a bit identical drop in.

That contract is what the knife edge test already validates: equality strictly between cumulative boundaries, top-k membership at the boundary itself. Tie order, greedy and NaN fallbacks, and the seeded draw stay exact.

tests/unit_test/moss_tts/test_sampling_kernels.py
JiaxinD: Refreshed at 1024, and it changes the numbers materially: graph replay is 8.21x at bs 8 and 8.69x at bs 16, against 2.55x and 2.64x at 1025 (BLOCK 2048). The body's production pattern now reads 1024, with 1025 kept as the boundary stress case.

Note: (E2E is unaffected, it always ran the real 1024 path; the sampler block is a fraction of a host-bound frame, which is why the end to end gain stays at 3 to 6%.)

charliechenye:
Cool stuffs @JiaxinD , great jobs. Just need to fix a few wording on PR body itself (non-blocking). Other than that, LGTM
Will approve after CI run.

JiaxinD:
Body wording fixed: the production audio head reads 1024 throughout (Motivation included), the parity contract and the "bit for bit" randomized result no longer read as contradictory, and the self-referential revision notes are gone.

Note: (if you meant other wording, point at it and I will take another pass. Needs the `run-ci` label from a maintainer.)

db-ol (approved):

charliechenye (approved):

PR #1667, opened 2026-08-22
[Benchmark][ASR] Add Pipecat STT benchmark dataset

0xjeffro:
## Motivation

as titled

## Modifications

All new files, tests align with the same structure with our current seedtts implementation, and add related docs and unit tests. In `benchmarks/eval/benchmark_asr_stt_benchmark.py`, we reused some common funcs in seedtts.

## Related Issues

N/A

## Accuracy Test

I ran the following commands on my own A6000 GPU to verify correctness.

```
sgl-omni serve --model-path Qwen/Qwen3-ASR-1.7B --port 8000

python -m benchmarks.eval.benchmark_asr_stt_benchmark \
    --port 8000 \
    --concurrencies 1,2,4,8,16,32,64 \
    --repeats 3 --warmup \
    --output asr_stt_benchmark_full.json \
    --save-raw-dir asr_stt_raw
```

partial result:

```
      "corpus_wer": {
        "mean": 0.019044154855802884,
        "min": 0.018998807051650246,
        "max": 0.019101982111555577,
        "n": 3
      },
      "per_sample_wer_max": {
        "mean": 0.5,
        "min": 0.5,
        "max": 0.5,
        "n": 3
      },
      "wall_clock_s": {
        "mean": 26.99310550217827,
        "min": 25.347746573388577,
        "max": 28.779551485553384,
        "n": 3
      },
      "throughput_samples_per_s": {
        "mean": 35.3510427309343,
        "min": 33.28752362526887,
        "max": 37.16306683407361,
        "n": 3
      },
      "rtfx": {
        "mean": 336.00966666666665,
        "min": 316.914,
        "max": 352.453,
        "n": 3
      },
      "latency_mean_s": {
        "mean": 1.7672654969069705,
        "min": 1.676691457840778,
        "max": 1.8784136634051707,
        "n": 3
      },
      "latency_median_s": {
        "mean": 1.8650136617943645,
        "min": 1.8149952571839094,
        "max": 1.9234457900747657,
        "n": 3
      },
      "latency_p95_s": {
        "mean": 2.435541404783726,
        "min": 2.1863965088501573,
        "max": 2.7636362573131916,
        "n": 3
      },
      "latency_p99_s": {
        "mean": 2.633562103807926,
        "min": 2.317407652176917,
        "max": 2.97215729093179,
        "n": 3
      },
      "rtf_mean": {
        "mean": 0.2260028191929906,
        "min": 0.21508047717947684,
        "max": 0.2401296230609372,
        "n": 3
      },
      "rtf_p95": {
        "mean": 0.4359,
        "min": 0.4145,
        "max": 0.4659,
        "n": 3
      },
      "resources": {
        "gpu_memory_used_peak_mib": {
          "per_repeat": [
            43481.6875,
            43481.6875,
            43481.6875
          ],
          "mean": 43481.6875,
          "min": 43481.6875,
          "max": 43481.6875
        },
        "gpu_memory_used_steady_mib": {
          "per_repeat": [
            43481.6875,
            43481.6875,
            43481.6875
          ],
          "mean": 43481.6875,
          "min": 43481.6875,
          "max": 43481.6875
        },
        "gpu_process_memory_peak_mib": null,
        "power_peak_w": {
          "per_repeat": [
            252.641,
            251.064,
            243.1
          ],
          "mean": 248.935,
          "min": 243.1,
          "max": 252.641
        },
        "system_cpu_peak_percent": {
          "per_repeat": [
            12.1,
            14.7,
            12.4
          ],
          "mean": 13.066666666666666,
          "min": 12.1,
          "max": 14.7
        },
        "gpu_process_cpu_peak_percent": null,
        "monitor_errors": [
          "GPU process metrics require at least one explicit NVML PID; pass --gpu-process-pid",
          "GPU process metrics require at least one explicit NVML PID; pass --gpu-process-pid",
          "GPU process metrics require at least one explicit NVML PID; pass --gpu-process-pid"
        ]
      },
      "per_repeat": [
        {
          "concurrency": 64,
          "repeat": 1,
          "evaluated
```
## Benchmark & Profiling

N/A

## Checklist

- [x] Add unit tests.

## CI

`/tag-and-rerun-ci whisper-asr` to select an ASR CI model. One selector from each family can be combined, for example


benchmarks/dataset/stt_benchmark.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def _staged_wav_path(staging_root: Path, sample_id: str, *, repo_id: str) -> Path:
+    """Return ``<staging_root>/<sample_id>.wav`` after rejecting unsafe ids."""
+    error_prefix = f"Invalid sample_id for {repo_id}: {sample_id!r}"
+    if not isinstance(sample_id, str) or not sample_id.strip():
+        raise ValueError(f"{error_prefix} (empty id)")
+    if "/" in sample_id or "\\" in sample_id or sample_id in {".", ".."}:
+        raise ValueError(f"{error_prefix} (path characters)")
+    wav_path = (staging_root / f"{sample_id}.wav").resolve()
```
AkazaAkane, line 37: sample_id is also the staged filename, so duplicate IDs silently overwrite earlier audio while the earlier SampleInput still points to the same path. That can corrupt WER without any error. Could we either reject duplicate sample_ids here or include the row index/content hash in the staged filename, and add a duplicate-ID regression test?

benchmarks/dataset/stt_benchmark.py
```diff
@@ -0,0 +1,111 @@
...
+    load_kwargs = {"revision": revision} if revision else {}
+    ds = load_dataset(repo_id, split=split, **load_kwargs)
+
+    missing = _REQUIRED_COLUMNS - set(ds.column_names)
+    if missing:
+        raise ValueError(
+            f"Dataset {repo_id} split={split} is missing columns: {missing}"
+        )
+
+    ds = ds.cast_column("audio", Audio(decode=False))
```
AkazaAkane, line 80: This assumes audio["bytes"] already contains WAV bytes, but Audio(decode=False) preserves the dataset's original encoding. Since downstream ASR treats every staged file as audio/wav, a custom dataset with FLAC/MP3/etc. will be mislabeled or fail. Should we decode/re-encode to WAV here, or explicitly validate/restrict the custom-dataset contract to WAV?

benchmarks/dataset/stt_benchmark.py
charliechenye: There is another collision path here beyond the staged filename. `build_asr_eval_results()` indexes outputs by `request_id`, and `request_id` comes from `sample.sample_id`, so duplicate IDs would still collapse multiple request results during WER assembly even if the staged filenames were made unique.

I think `sample_id` uniqueness should be an explicit loader invariant, with a regression test for duplicate IDs.

benchmarks/README.md
```diff
@@ -223,6 +224,21 @@ python -m benchmarks.eval.benchmark_asr_seedtts \
   --max-samples 20 --concurrencies 2 --repeats 1 --stream
 ```
 
+`benchmark_asr_stt_benchmark.py` runs the same sweep on
+[`pipecat-ai/stt-benchmark-data`](https://huggingface.co/datasets/pipecat-ai/stt-benchmark-data):
```
charliechenye, line 228: Could we explicitly clarify that this PR reuses the Pipecat dataset only, not Pipecat's benchmark metric definitions?

The upstream Pipecat benchmark reports Semantic WER and TTFS, where TTFS is measured from actual speech end to the final streaming transcript. This benchmark keeps SGLang's existing Whisper normalized WER and whole request latency/RTF metrics; even `--stream` uploads the full audio file and streams the transcript response.

Without this distinction, the resulting WER/latency numbers can easily be compared against the Pipecat leaderboard even though they are not apples to apples.

benchmarks/eval/benchmark_asr_stt_benchmark.py
```diff
@@ -0,0 +1,187 @@
...
+from benchmarks.dataset.stt_benchmark import (
+    STT_BENCHMARK_LANG,
+    STT_BENCHMARK_SPLIT,
+    load_stt_benchmark_samples,
+)
+from benchmarks.eval.asr_profiling import (
+    collect_environment_fingerprint,
+    collect_server_identity,
+)
+from benchmarks.eval.benchmark_asr_seedtts import (
```
charliechenye, line 65: Non blocking: this makes `benchmark_asr_seedtts.py` both an executable entry point and the implementation library for this new benchmark, including dependencies on several underscore private helpers. If these helpers are now intentionally shared across ASR benchmarks, could we give that shared implementation an explicit boundary rather than coupling one benchmark entry point to another's private functions?

charliechenye (changes requested):
Could you include `evaluated/total` or the request success rate for each concurrency in the benchmark table?

`calculate_wer_metrics()` computes corpus WER only over successful samples, so stable WER by itself can hide request failures at higher concurrency. The benchmark already records `evaluated`, `total`, and `skipped`, so I think those should be part of the validation evidence alongside throughput, latency, RTF, and WER.

benchmarks/eval/benchmark_asr_stt_benchmark.py
0xjeffro: yep thanks for mention this

benchmarks/README.md
0xjeffro: fixed

benchmarks/dataset/stt_benchmark.py
```diff
@@ -0,0 +1,123 @@
...
+    seen_ids: set[str] = set()
+    for row in ds:
+        sample_id = row["sample_id"]
+        if sample_id in seen_ids:
+            raise ValueError(
+                f"Duplicate sample_id for {repo_id}/{split}: {sample_id!r}"
+            )
+        seen_ids.add(sample_id)
+        wav_path = _staged_wav_path(staging_root, sample_id, repo_id=repo_id)
+        audio = row["audio"] or {}
```
AkazaAkane, line 99: Audio(decode=False) may return a local path with bytes=None; this is the documented representation for path-backed Hugging Face audio datasets. Since --repo-id advertises any HF repo with these columns, a valid PCM-WAV repo can fail here with Empty audio bytes. Could we fall back to reading audio["path"], then validate and stage those bytes, and add a bytes=None/path=<wav> regression test? Otherwise, the generic --repo-id contract should be restricted or removed.

AkazaAkane: Minimum fix could be
```audio_bytes = audio.get("bytes") 
if not audio_bytes:
    audio_path = audio.get("path")
    if not audio_path:
        raise ValueError(...)
    audio_bytes = Path(audio_path).read_bytes()```

Comment (0xjeffro) [reply benchmarks/dataset/stt_benchmark.py:99]: thx, fixed

File: benchmarks/dataset/stt_benchmark.py [hunk 6]
```diff
@@ -0,0 +1,126 @@
...
+        audio_bytes = audio.get("bytes")
+        if not audio_bytes:
+            audio_path = audio.get("path")
+            if not audio_path:
+                raise ValueError(f"Empty audio bytes for {repo_id}/{split}/{sample_id}")
+            audio_bytes = Path(audio_path).read_bytes()
+
+        if audio_bytes[:4] != b"RIFF" or audio_bytes[8:12] != b"WAVE":
```
charliechenye, line 107: I don't think this currently validates the contract stated in the error message. Checking only the `RIFF` / `WAVE` magic accepts malformed or non PCM WAV payloads. The new `_stt_wav_bytes()` test fixture demonstrates this: it contains only the two magic markers plus arbitrary bytes, with no `fmt` or `data` chunk, yet the loader accepts it.

This matters for benchmark correctness because downstream `get_wav_duration()` assumes a PCM WAV layout and reads channels, sample rate, and bit depth from fixed header offsets; accepting a payload that does not satisfy that layout can produce incorrect RTF/RTFx.

Could we validate the actual WAV/PCM structure here, and make the regression fixture a valid WAV?

benchmarks/dataset/stt_benchmark.py
```diff
@@ -0,0 +1,126 @@
...
+            raise ValueError(
+                f"Non-WAV audio bytes for {repo_id}/{split}/{sample_id}; "
+                "this loader expects PCM WAV audio"
+            )
+        wav_path.write_bytes(audio_bytes)
+
+        transcription = str(row["transcription"] or "").strip()
```
charliechenye, line 114: Should we reject an empty transcription here? An empty reference currently survives dataset loading, the ASR request can succeed, and then WER scoring marks it un-evaluable because the normalized reference is empty. That turns invalid ground truth into a skipped sample rather than failing the benchmark input early, which can reduce evaluation coverage silently.

AkazaAkane (approved):

PR #1663, opened 2026-08-22
[Perf][Fun-CosyVoice3] Batch inference for Flow decoder

nagisa-kunhah:
## Motivation

Related #1652

## Modifications

- Add `flow_batch.py`, a local adapter for the pinned CosyVoice Flow/CFM implementation (`074ca6dc9e80a2f424f1f74b48bdd7d3fea531cc`). The adapter:
  - packs variable-length prompt and target tokens with explicit masks;
  - preserves per-request prompt mel conditioning and speaker embeddings;
  - runs classifier-free guidance as the required `2 * B` estimator batch;
  - reuses the upstream deterministic noise prefix and returns each unpadded target mel independently.
- Batch compatible requests in the buffered vocoder. Requests are grouped by total mel length into configurable frame buckets before Flow inference, which keeps padding bounded while preserving original response ordering.
- Keep the original `flow.inference()` path for singleton buckets, disabled batching, TensorRT/non-PyTorch estimators, and incompatible future CosyVoice structures. Unsupported structures are reported once at startup rather than failing a request.
- Enable Flow batching by default in the Fun-CosyVoice3 pipeline configuration: `max_batch_size=8`, `max_batch_wait_ms=2`, and `flow_batch_bucket_frames=50`.
- Expose `enable_flow_batch` and `flow_batch_bucket_frames` as vocoder factory arguments and document how to disable batching or tune the bucket size.
- Add structural/unit tests for masking, CFG layout, variable lengths, fallback behavior, scheduler wiring, ordering, and configuration defaults. Add a real-checkpoint GPU invariance test against the pinned upstream Flow path.

### Scope and non-goals

- This PR does **not** batch HiFT. HiFT remains one invocation per mel in this branch. PR #1655 is complementary and should be integrated separately so Flow batching and exact-length HiFT batching can be evaluated both alone and together.
- This PR does **not** change the causal streaming path. The streaming work in PR #1656 has a distinct chunking/causality contract and should not reuse this buffered adapter blindly.
- The adapter is intentionally tied to the documented CosyVoice commit. It does not monkey-patch the CosyVoice checkout.

## Related Issues

Related #1652

Related works: #1655 (HiFT batching), #1656 (causal streaming vocoder).

### Full SeedTTS WER comparison

[...]

PR #1670, opened 2026-08-22
[Fun-CosyVoice3] torch.compile the DiT (flow decoder) backbone

CAICAIIs:
Related: #1652 (DiT → torch.compile for DiT backbone)

## Motivation

The CosyVoice3 flow decoder's DiT backbone (`flow.decoder.estimator`, a 22-layer `cosyvoice.flow.DiT.dit.DiT`) runs once per Euler step (`n_timesteps=10`) with a classifier-free-guidance batch of `[2, 80, T]`. This PR makes `torch.compile` available for that backbone.

## Modifications

- Add `_compile_dit_backbone()` + warmup helpers in `fun_cosyvoice3/stages.py`.
- Compile the bound `estimator.forward` with `torch.compile(..., dynamic=True)` so one symbolic-sequence-length graph serves every utterance length. Default (non-CUDA-graph) mode: `reduce-overhead` was explored but re-captures a CUDA graph per new sequence length (+50–220 ms on the first request of each length), which is a net regression for a TTS server whose texts vary in length.
- Warm up under `torch.inference_mode()` to match the `@torch.inference_mode()` decorator on `flow.inference`, and under the same autocast flag as serving (fp16 when the vocoder runs in float16), so the first real request reuses the warmed graph.
- Graceful fallback: if `torch.compile` (lazy) fails during the warmup, the original eager forward is restored, a warning is logged, and the vocoder keeps serving.
- Compile the bound method rather than wrapping the module in an `OptimizedModule`, so parameter names stay stable for weight updates. A non-PyTorch estimator (e.g. TensorRT) is skipped gracefully.
- Expose `enable_dit_torch_compile` as a vocoder factory arg (off by default) and document it in the cookbook. Enable via:
  `--vocoder.factory.enable_dit_torch_compile true`.
- Unit tests (compile wiring, warmup grad mode, non-module skip, degenerate warmup rejection, compile-failure fallback, factory-flag default) + an accelerator-marked GPU smoke test (real `torch.compile` on a tiny DiT-shaped module, dynamic-length equivalence).

## Benchmark (flow-only)

A100-SXM4-40GB, `Fun-CosyVoice3-0.5B-2512`, `flow.inference` eager vs compiled, 30 repeats/length, identical inputs.

| target tokens | output audio | eager tok/s | compiled tok/s | speedup | max abs mel diff |
| 50  | 2.0s | 198 | 262 | 1.32x | 1.5e-3 |
| 100 | 4.0s | 319 | 376 | 1.18x | 2.3e-3 |
| 200 | 8.0s | 415 | 490 | 1.18x | 1.6e-3 |
| 400 | 16.0s | 474 | 545 | 1.15x | 5.6e-3 |

Speedup is larger for shorter utterances (fixed launch-overhead reduction is a bigger fraction), the expected `torch.compile` signature. One-time compile cost ~34s at startup. (An earlier run on a busier GPU gave 1.16–1.43x; the mel diff is stable across runs.)

## End-to-end request latency (`sgl-omni serve`, real reference audio, fixed AR seed)

Full pipeline (`preprocessing → tts_engine(AR) → vocoder(flow+HiFT)`), 10 requests with texts from 3–128 chars, sequential concurrency, warmup request excluded. Median of 3 eager / 3 compiled runs.

| metric | eager | compiled | delta |
| --- | ---: | ---: | ---: |
| latency mean | 1.06 s | 1.00 s | **−6%** |
| latency p50 | 0.99 s | 0.95 s | **−4%** |
| latency p95 | 1.49 s | 1.38 s | **−7%** |
| RTF mean | 0.30 | 0.27 | **−10%** |

First-request latency also improves: eager pays ~7.9 s of lazy init on its first request, while the compiled server pre-compiles the DiT at startup so its first request is ~0.6 s. The e2e gain is smaller than the flow-only gain because the DiT is only part of the AR→flow→HiFT critical path; the rest (AR generation, reference preprocessing, HiFT, audio encode) is out of scope for this PR and tracked separately in #1652 (#1655 HiFT batching, #1663 flow batching).

## Accuracy / correctness gate

- The compiled DiT output matches eager to **max abs diff < 6e-3** in log-mel space, deterministic across runs.
- The CosyVoice3 HiFT vocoder is **deterministic** (same mel input → bit-identical waveform, verified), so the only change `torch.compile` introduces is that ≤ 6e-3 log-mel delta; everything downstream is untouched.

### End-to-end WER / speaker similarity (one SeedTTS sample, identical AR tokens)

| metric | eager | compiled |
| --- | ---: | ---: |
| WER (Whisper base) | 0.000 | 0.000 |
| speaker similarity (WavLM, ref vs gen) | 62.74 | 63.14 |
| waveform SNR (eager vs compiled) | — | 26.97 dB |

WER is identical (0%) and speaker similarity is unchanged (+0.4 pp, within the HiFT output variance). No regression.

### Interaction with #1663 (Flow batching) and #1715 (bf16 autocast) — re-verified after rebase

Both PRs merged before this one. Verified on the rebased code (torch 2.13 + sglang 0.5.18, default bf16 config):

- **Flow batching (#1663)**: the compiled DiT is used by the batched path (`decoder.forward_estimator`); the estimator's batch is `2 × batch_size`, and `dynamic=True` reuses one graph across batch sizes (first batched call measured 175 ms, no recompile at estimator batch 6).
- **bf16 autocast (#1715)**: the warmup runs under the same `compute_dtype` as serving (bfloat16 for bf16 configs), so the first real request reuses the warmed graph instead of paying an ~85 s recompile. End-to-end under the default bf16 config, eager vs compiled (real AR tokens, fixed seed):

| metric | eager (bf16) | compiled (bf16) |
| --- | ---: | ---: |
| WER (Whisper base) | 0.000 | 0.000 |
| speaker similarity (WavLM, ref vs gen) | 61.65 | 62.64 |
| eager vs compiled similarity | — | 97.28 |

WER and speaker similarity are unchanged. Note: the raw waveform deviates more under bf16+compile than under fp32 (wav SNR ~−1 dB vs ~27 dB) because bf16 precision amplifies the compiler's kernel reordering, but this does not move WER or the speaker embedding, as shown above.

## Known limitations (documented in the helper docstring)

- `add_optional_chunk_mask` ends in a `Tensor.item()` guard that Dynamo cannot trace in-graph (`capture_scalar_outputs` hits an Inductor "Invalid NaN comparison" under dynamic shapes); it stays a graph break + one sync per Euler step, the same sync eager already pays.
- The causal streaming estimator branch (`streaming=True`) uses `torch.div(..., rounding_mode='trunc')` in `subsequent_chunk_mask`, which Inductor cannot bounds-check under a dynamic sequence length, so it is not warmed and would fall back to eager until that Inductor path is fixed.

## Tests

- `pytest tests/unit_test/fun_cosyvoice3/` → 47 passed.
- `pytest tests/unit_test/fun_cosyvoice3/test_dit_compile_gpu.py` → 1 passed (GPU).

## Checklist

- [x] Format with black + isort.
- [x] Add unit tests (+ GPU smoke test).
- [x] Update documentation.
- [x] Provide benchmark + accuracy results.

### DIFF
--- a/sglang_omni/models/fun_cosyvoice3/config.py
+++ b/sglang_omni/models/fun_cosyvoice3/config.py
@@ -55,9 +55,7 @@ class FunCosyVoice3PipelineConfig(PipelineConfig):
                 dtype="bfloat16",
                 flow_batch_bucket_frames=50,
                 flow_batch_admission_frames=2000,
-                # Opt-in: torch.compile the DiT (flow decoder) backbone. Off by
-                # default because it adds a one-time compile cost at startup
-                # and is most beneficial under sustained throughput load.
+                # Opt-in; off by default (one-time startup compile cost).
                 enable_dit_torch_compile=False,
             ),
             gpu=0,
--- a/sglang_omni/models/fun_cosyvoice3/config.py
+++ b/sglang_omni/models/fun_cosyvoice3/config.py
@@ -55,6 +55,10 @@ class FunCosyVoice3PipelineConfig(PipelineConfig):
                 dtype="bfloat16",
                 flow_batch_bucket_frames=50,
                 flow_batch_admission_frames=2000,
+                # Opt-in: torch.compile the DiT (flow decoder) backbone. Off by
+                # default because it adds a one-time compile cost at startup
+                # and is most beneficial under sustained throughput load.
+                enable_dit_torch_compile=False,
             ),
             gpu=0,
             terminal=True,
--- a/sglang_omni/models/fun_cosyvoice3/stages.py
+++ b/sglang_omni/models/fun_cosyvoice3/stages.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import logging
 from collections import defaultdict
 from dataclasses import dataclass
 from typing import Any, Sequence, cast
@@ -66,6 +67,9 @@ class _PackedFlowBatch:
     prompt_mel_lengths_tensor: torch.Tensor
     prompt_feat: torch.Tensor
     embedding: torch.Tensor
+
+
+logger = logging.getLogger(__name__)
 
 
 def _flow_device_and_dtype(flow: Any) -> tuple[torch.device, torch.dtype]:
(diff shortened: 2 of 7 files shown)


PoTaTo-Mika:
Serve CLI has already updated, please rebase and check the pull request. Thank you very much.

PoTaTo-Mika (approved):
## ZH

| Conc | QPS before | QPS after | Δ | RTF before | RTF after | Δ | Latency before | Latency after | Δ |
|---|---|---|---|---|---|---|---|---|---|
| 1 | 1.128 | 1.331 | **+18.0%** | 0.1394 | 0.1175 | -15.7% | 0.887 | 0.751 | -15.3% |
| 2 | 1.748 | 1.959 | **+12.1%** | 0.1804 | 0.1591 | -11.8% | 1.144 | 1.021 | -10.8% |
| 4 | 2.600 | 3.372 | **+29.7%** | 0.2427 | 0.1859 | -23.4% | 1.538 | 1.185 | -23.0% |
| 8 | 3.023 | 4.061 | **+34.3%** | 0.4233 | 0.3098 | -26.8% | 2.644 | 1.967 | -25.6% |
| 16 | 3.156 | 4.138 | **+31.1%** | 0.8146 | 0.6188 | -24.0% | 5.059 | 3.854 | -23.8% |

## EN

| Conc | QPS before | QPS after | Δ | RTF before | RTF after | Δ | Latency before | Latency after | Δ |
|---|---|---|---|---|---|---|---|---|---|
| 1 | 1.178 | 1.433 | **+21.6%** | 0.1873 | 0.1544 | -17.6% | 0.849 | 0.698 | -17.8% |
| 2 | 1.872 | 2.132 | **+13.9%** | 0.2367 | 0.2058 | -13.1% | 1.068 | 0.938 | -12.2% |
| 4 | 2.595 | 3.248 | **+25.2%** | 0.3407 | 0.2635 | -22.7% | 1.540 | 1.230 | -20.1% |
| 8 | 2.880 | 3.514 | **+22.0%** | 0.6382 | 0.5137 | -19.5% | 2.772 | 2.272 | -18.0% |
| 16 | 2.997 | 3.493 | **+16.5%** | 1.2187 | 1.0531 | -13.6% | 5.321 | 4.552 | -14.5% |

sglang_omni/models/fun_cosyvoice3/config.py
```diff
@@ -55,6 +55,10 @@ def process_local_edges(cls) -> frozenset[tuple[str, str]]:
                 dtype="bfloat16",
                 flow_batch_bucket_frames=50,
                 flow_batch_admission_frames=2000,
+                # Opt-in: torch.compile the DiT (flow decoder) backbone. Off by
```
PoTaTo-Mika: notes should be linked with author.

sglang_omni/models/fun_cosyvoice3/stages.py
```diff
@@ -329,6 +333,144 @@ def _load_cosyvoice3_flow_hift(
     return FunCosyVoice3Flow(flow), hift
 
 
+def _configure_dit_torch_compile() -> None:
+    """Enable Inductor/Dynamo flags that make the DiT graph compile well.
```
PoTaTo-Mika: notes too much.

Issue #1650, opened 2026-08-21
[Roadmap] Runtime Profile Agent for SGLang Omni


db-ol:
## Goal

Renamed on 2026-09-09 from Kernel Development Agents (KDA) to Runtime Profile Agent, as proposed by @zhaochenyang20. The scope is unchanged, the order is now explicit: profile each model's runtime first, optimize at the scheduling and orchestration layer while the GPU is still idle, and enter kernel work only once the GPU is busy. The kernel stage keeps the adjudication protocol below. The new name also avoids the collision with MIT Han Lab's Kernel Design Agents.

Make agent-driven runtime optimization a repeatable, adjudicated pipeline for SGLang Omni. A generated change is cheap. A trusted change is expensive. The pipeline spends its effort on the second part.

Founding case study: #1608 by @junliu-mde. External references: the [agent assisted SGLang development write-up](https://www.lmsys.org/blog/2026-07-02-agent-assisted-sglang-development) and [PolyArch humanize](https://github.com/PolyArch/humanize).

## Evaluation protocol (v0, comments welcome)

Every candidate passes the same protocol before any gain is claimed:

1. A correctness gate declared up front: bitwise identity under a controlled batching protocol, or for ASR possibly token identity plus a bounded WER regression, see the open question below.
2. An isolated benchmark against a functionally equivalent baseline under the same inputs, launch path, and compile flags. Control for warm-up and temporal drift, for example interleaved A/B or A-B-A with the first round discarded, and require the effect to clear the measured noise band.
3. Positive evidence the new code path actually executed, from counters or traces. Absence of errors is not evidence.
4. An end-to-end measurement on the target workload. Isolated wins that do not translate end-to-end are retained as optimization inventory but are not counted as delivered gains (#1608 documents this in detail). Match workload size to the effect being measured. In the #1620 reproduction, a 64 sample quick gate could not resolve a c16 effect that a 1088 clip run resolved (see comments). When the noise band is wider than the expected effect, an inconclusive measurement is a rerun trigger rather than evidence of no gain (see finals comment).
5. Independent reproduction before a merge recommendation.

Distilled from #1608, #1418 and the MTD kernel sweep. Corrections welcome, especially from anyone who has hit cases this does not cover.

## Stage 0: calibration (first, about one week)

Run candidate models through tasks with known answers, so the grader is validated against ground truth before we trust it on open problems.

[...]

PR #1675, opened 2026-08-24
[Qwen3-TTS] Capture CUDA graphs for followup vocoder decodes

db-ol:
## Motivation

The Qwen3-TTS streaming vocoder decodes the first audio chunk through CUDA graphs, but every follow up chunk runs the decoder eagerly. One follow up decode launches about 940 kernels and costs about 11 ms of host time, and in the default single process topology that host time competes with the scheduler loop. At 16 concurrent streams this measurably slows the whole pipeline.

## Modifications

- Generalize the decode graph holder to capture graphs keyed by (input frames, batch bucket) and to accept several frame counts. One holder instance per decode stream, so the initial and follow up workers never share static buffers.
- Give the follow up worker its own graph set, captured at serving start for the steady state window (left context plus follow up stride). Shapes outside the captured buckets fall back to the existing eager path.
- New followup_cuda_graph option on the scheduler and the vocoder factory, on by default.
- One INFO log listing the captured (frames, batch) keys.

## Related Issues

Part of the kernel work tracked in #1650.

## Accuracy Test

- Unit: decoder output through the follow up graphs is bitwise equal to eager at captured batch sizes (1, 4, 8). Batch sizes padded to a bucket show the same small numeric differences the existing initial chunk graphs already have.
- Serving WER, SeedTTS EN 96 clips at c16, whisper-large-v3 transcription, corpus WER: baselines 0.47 and 0.66 percent across two server restarts, candidate 0.66 and 0.48 percent across two runs. One candidate run contained one degenerate generation (1 of 96) that did not reproduce on rerun. Generation is not deterministic across restarts (34 of 40 sampled clips differ between the two baseline runs), so the comparison is corpus level.
- tests/unit_test/qwen3_tts/test_pipeline.py passes (123 passed, 1 skipped), two tests added.
- After merging main (c2d193f, includes #1523): 139 passed. WER gate repeated on the new main, corpus WER 0.47 percent on baseline, candidate, and second baseline, 0 of 96 clips above 50 percent, per clip WER identical across the three arms.

## Benchmark & Profiling

Three arm serving gate (baseline, candidate, baseline), streaming, SeedTTS EN 96 clips, warm repeats compared against the larger of the restart band and the repeat band, H200, two independent runs:

| c16 metric | run 1 | run 2 |
| --- | --- | --- |
| throughput | +7.0 percent, band 3.6 | +9.7 percent, band 5.1 |
| inter chunk latency mean | +10.0 percent, band 1.3 | +9.4 percent, band 4.1 |
| inter chunk latency p95 | +9.7 percent, band 5.7 | +13.3 percent, band 4.4 |

c1 is unchanged. With the vocoder in its own process (the CI topology) the change is neutral, the gain comes from removing host launch time from the shared process. A single follow up decode drops from about 11 ms eager to under 5 ms graph replay at batch 1.

Re-measured after merging main (c2d193f, includes #1523), same gate, two candidate servers plus a control server with followup_cuda_graph off:

| c16 metric | candidate 1 | candidate 2 | band |
| --- | --- | --- | --- |
| throughput | +6.2 percent | +8.8 percent | 2.5 |
| inter chunk latency mean | +6.3 percent | +9.0 percent | 2.6 |
| inter chunk latency p95 | +13.8 percent | +15.4 percent | 11.9 |

The control arm shows no improvement, so the gain is the follow up graphs and not the merge. c1 remains within noise.

## Checklist

- [x] Add unit tests.


db-ol:
Merged main after #1523 landed. The followup graph path now sits in the new _launch_async, selected by stream the same way the initial graphs are, everything else is unchanged. 139 unit tests pass.

Re-measured against the new main (c2d193f) on one H200, A-B-A with three fresh servers plus a second candidate server, 96 samples, three repeats, warm repeats compared. At concurrency 16: throughput +6.2 and +8.8 percent (band 2.5), inter chunk latency mean improved 6.3 and 9.0 percent (band 2.6), p95 improved 13.8 and 15.4 percent (band 11.9). Concurrency 1 is within noise. A control arm with followup_cuda_graph off on the same code shows no improvement. WER gate: corpus WER 0.47 percent on all three arms, 0 of 96 above 50 percent.

So the gain survives the removal of the per batch synchronize in #1523. The vocoder now captures two holders, initial and followup, both keyed by (frames, batch bucket).

sglang_omni/models/qwen3_tts/streaming_vocoder.py
```diff
@@ -247,37 +247,42 @@ def _drop_views(self) -> None:
...
         capture_stream = torch.cuda.Stream(device=self._device)
         graph_pool = torch.cuda.graph_pool_handle()
-        for batch_size in self._batch_sizes:
+        for input_frames, batch_size in (
+            (f, b) for f in self._input_frames for b in self._batch_sizes
+        ):
             try:
                 static_input = torch.zeros(
-                    (batch_size, self._num_quantizers, self._input_frames),
+                    (batch_size, self._num_quantizers, input_frames),
```
luojiaxuan, line 285: Could we add capture_stream.wait_stream(torch.cuda.current_stream(self._device)) before the side stream warmup? static_input is initialized on the current stream and immediately read on capture_stream, while capture_stream.synchronize() does not order the earlier current stream work.

tests/unit_test/qwen3_tts/test_pipeline.py
```diff
@@ -1519,6 +1519,33 @@ def test_qwen3_tts_initial_decode_graphs_noop_on_cpu() -> None:
...
+        device=torch.device("cpu"),
+        num_quantizers=2,
+        input_frames=(24, 32, 24),
+        batch_sizes=(4, 1),
+    )
+
+    assert graphs._input_frames == (24, 32)
+    assert graphs._batch_sizes == (1, 4)
+    graphs.capture()
+    assert graphs.decode(torch.zeros((1, 2, 24), dtype=torch.long)) is None
```
luojiaxuan, line 1546: This test runs on CPU, so _enabled is false, capture() is a no op, and decode() returns None. Could we inject fake graph entries and verify that B=3 selects (24, 4), clears the padded row, calls replay(), slices the output back to B=3, selects the correct frame key, and returns None for unsupported shapes?

sglang_omni/models/qwen3_tts/streaming_vocoder.py
```diff
@@ -349,6 +361,7 @@ def __init__(
         followup_batch_wait_ms: int = 1,
         initial_cuda_graph: bool = True,
         enable_deterministic_inference: bool = False,
+        followup_cuda_graph: bool = True,
```
luojiaxuan, line 365: followup_cuda_graph defaults to true, while deterministic mode only disables initial_cuda_graph. Could we report test_qwen3_tts_deterministic_batch_invariance passing with followup graphs enabled, or explicitly disable this path in deterministic mode?

sglang_omni/models/qwen3_tts/streaming_vocoder.py
```diff
@@ -457,6 +490,7 @@ def on_serving_start(self) -> None:
         if not self._async_decode:
             return
         self._initial_decode_graphs.capture()
+        self._followup_decode_graphs.capture()
```
luojiaxuan, line 494: Since this is enabled by default and adds another private graph pool, could we report the reserved memory delta, added startup capture time, and c16 TTFA p50 and p95? This is most useful for the separate vocoder process case, where the current result is neutral.

luojiaxuan (changes requested):

sglang_omni/models/qwen3_tts/streaming_vocoder.py
db-ol: Agreed. PyTorch creates side streams with cudaStreamNonBlocking, so the zero fill issued on the current stream is not ordered before the warmup on capture_stream, and capture_stream.synchronize() only waits for work already on that stream. Added capture_stream.wait_stream(torch.cuda.current_stream(self._device)) right after static_input is created, before the warmup. The initial capture on main has the same ordering gap, and both holders share this capture code, so both get the fix.

tests/unit_test/qwen3_tts/test_pipeline.py
db-ol: Added test_qwen3_tts_decode_graphs_replay_pads_batch_and_slices_output. It injects fake graph entries for (24, 1), (24, 4) and (32, 1) with a replay counter and checks that a B=3 input with 24 frames selects (24, 4), copies the three rows into the static input and clears the padded row, replays that graph exactly once, and returns a [3, 1, 96] slice of the static output in fresh storage. A 32 frame input selects (32, 1). B=3 at 32 frames (bucket not captured), B=5, 20 frames, a wrong quantizer count and a 2D input all return None without replaying anything.

sglang_omni/models/qwen3_tts/streaming_vocoder.py
db-ol: Good catch. Deterministic mode now disables the followup graphs the same way it disables the initial ones: stage_factory_kwargs adds followup_cuda_graph False for the vocoder stage (an explicit factory value still wins), and test_qwen3_tts_deterministic_inference_configures_pipeline asserts both. So the deterministic path is unchanged from main, follow up chunks stay eager there. If you would rather keep graphs on in deterministic mode I can run test_qwen3_tts_deterministic_batch_invariance with them enabled and report, but matching initial_cuda_graph looked like the safer default.

sglang_omni/models/qwen3_tts/streaming_vocoder.py
db-ol: Measured on one H200, torch 2.11.0+cu130, Qwen3-TTS-12Hz-1.7B-Base, two runs.

Memory: the followup holder captures four graphs, (24, 1), (24, 2), (24, 4), (24, 8). With the default strides the two followup frame counts (left context 16 plus followup stride 8, and left context plus the initial followup stride) both resolve to 24, so they dedupe to one key. Its private pool reserves 822 MiB (torch.cuda.memory_reserved delta, 34 MiB of that stays allocated after capture, the rest is held by the pool), 832 MiB at the device level (mem_get_info delta). The initial holder costs the same, 822 MiB reserved and 910 MiB device level. The 0.6B checkpoint gives the same memory numbers, its speech tokenizer decoder is the same size. The two pools stay separate on purpose, the holders replay concurrently on two streams and a shared pool is only safe for ordered replays.

Startup: the followup capture runs right after the initial capture in on_serving_start. In isolation it took 0.41 and 1.29 s (the initial capture took 2.8 and 3.1 s), 1.0 s on the 0.6B checkpoint. In the three gate servers the gap between the two captured log lines was 0.25, 0.37 and 0.47 s.

c16 TTFA from the gate runs already posted (time to first audio payload, mean of the warm repeats, 96 clips, single process topology). Second gate run: baseline p50 0.326 and 0.319 s, p95 0.596 and 0.476 s, candidate p50 0.298 and 0.297 s, p95 0.392 and 0.437 s, control with followup graphs off p50 0.366 s, p95 0.572 s. First gate run: baseline p50 0.335 and 0.306 s, p95 0.530 and 0.432 s, candidate p50 0.300 s, p95 0.524 s. c1 TTFA p50 is 0.117 to 0.126 s on the baseline and candidate arms and 0.132 s on the control. So TTFA p50 is within noise to slightly better and p95 stays inside the baseline spread, nothing got worse.

Separate vocoder process topology (--vocoder.process vocoder with gpu_memory_fraction 0.5 for the engine and 0.10 for the vocoder, as in CI), same gate, A-B-A on the same code with followup graphs off, on, off, three repeats per server, warm repeats compared: the vocoder process grows from 2044 to 2876 MiB in nvidia-smi (plus 832 MiB, matching the isolated measurement), the followup capture adds 0.36 s between the two captured log lines, c16 TTFA p50 is 0.319 s on versus 0.320 and 0.355 s off (band 10 percent), p95 0.476 s versus 0.626 and 0.606 s (band 46 percent), throughput plus 2.1 percent (band 9.4 percent), and c1 is within noise on every metric. So in that topology the cost is 832 MiB in the vocoder process and under half a second of startup, and latency does not move outside the band in either direction.

db-ol:
Thanks for your careful review, @luojiaxuan ! Fixed. Could you please take another look when you have a chance?

luojiaxuan (approved):
LG

PR #1359, opened 2026-08-06
[ASR] Enable Qwen3-ASR streaming transcription output

SandyLuXY:
## Motivation

- Enable Qwen3-ASR to return incremental transcript deltas through the existing
  `/v1/audio/transcriptions` SSE API instead of waiting for only the terminal
  transcript.
- Reuse the shared token-to-text streaming mechanics while keeping Qwen3-ASR
  decoding policy and message metadata in the model adapter.

## Modifications

1. `sglang_omni/models/qwen3_asr/request_builders.py`
   - Add `make_qwen3_asr_stream_output_builder` as a thin wrapper around the
     shared token-to-text stream builder.
   - Decode transcript-only sampled tokens, attach text-stage metadata, honor
     tokenizer or explicit EOS IDs, and flush buffered text on terminal output.
2. `sglang_omni/models/qwen3_asr/stages.py` and `engine_builder.py`
   - Add a configurable `stream_emit_interval_s` with a 50 ms default and
     forward it through `Qwen3ASREngineBuilder`.
   - Construct the model-specific stream builder from the initialized tokenizer
     and register it through the engine builder's scheduler kwargs.
3. `tests/unit_test/qwen3_asr/`
   - Cover delta payloads, prefill/non-stream gating, rate limiting, EOS and
     terminal flushes, incomplete UTF-8 buffering, special-token suppression,
     explicit EOS override, and stage wiring.
4. `docs/cookbook/qwen3_asr.md` and `tests/README.md`
   - Document `stream=true`, the SSE event sequence, the default emit interval,
     and the new unit-test module.

## Related Issues

Related to #1267.

Uses the shared token-to-text streaming mechanics merged in #1275.

## Accuracy Test

**Setup:**

- Model: `Qwen/Qwen3-ASR-1.7B`, cached snapshot
  `7278e1e70fe206f11671096ffdd38061171dd6e5`.
- Dataset: `zhaochenyang20/seed-tts-eval-arrow`, English split at
  `27f4c1adee83b5b29b7c4b375f6b976324bda308`, all 1,088 clips.
- Baseline: `main @ 96c1960c`, non-streaming requests.
- Candidate: `feat/qwen3-asr-streaming @ 1180b9f`, requests sent with `--stream`.
- Hardware/topology: physical GPU 7, NVIDIA H200 143,771 MiB, DP=1.
- Runtime: BF16 checkpoint default, FA3 attention, CUDA graphs and torch.compile
  enabled, 64 max running requests.
[...]

PR #1275, opened 2026-08-02
[ASR Refactor][M4] Extract shared token text streaming mechanics

guozhihao-224:
## Summary
- Extract shared ASR token→text streaming into sglang_omni/scheduling/token_text_streaming.py (pending buffer, rate-limit, UTF-8 hold, EOS/terminal flush, OutgoingMessage construction).

- Migrate Fun-ASR and MOSS-Transcribe-Diarize stream builders to thin wrappers that supply model-owned hooks: decode_fn, build_message_data / build_message_metadata, EOS resolution, emit interval, and allow_terminal_flush.

- Keep per-request stream state on req as a single object req._asr_token_text_stream (pending_ids, last_emit_t), instead of model-specific dynamic attribute names.

- Preserve behavior: Fun-ASR keeps .flush + terminal bypass; MOSS-TD does not gain terminal flush; public builder APIs and stage wiring are unchanged.

## REF
Closes #1271  
Related to #1267

## Test plan
- [x] pytest tests/unit_test/fun_asr/test_stream_output_builder.py -q
- [x] pytest tests/unit_test/fun_asr/test_stream_output_builder.py -q
- [x] Confirm Fun-ASR builder exposes .flush; MOSS-TD builder does not
- [x] Spot-check fixed token traces vs origin/main legacy builders (OutgoingMessage deltas / metadata / ordering / terminal)

PR #1094, opened 2026-07-19
[Perf] Add stream output builder for Fun-ASR model

keke0315:
## Motivation

Track F-PR7  in issue sgl-project/sglang-omni#924 : Streaming Output, wire stream_output_builder into OmniScheduler

## Modifications

This follows the stream output design introduced in PR sgl-project/sglang-omni#965 for MOSS-Transcribe-Diarize. 
Will extract the common abstractions in a follow-up refactor

## Note

Fun-ASR is designed for short segments, streaming output primarily improves ttft rather than overall latency

 Performance numbers may vary across different hardware configurations and even across repeated runs on the same setting. To ensure a fair comparison and minimize run-to-run variance, I kept the benchmarking conditions as consistent as possible (same machine, software stack, configuration, and execution time). The results are to highlight the relative latency difference between the streaming and non-streaming pattern, rather than their absolute latency values

### Benchmark for Fun-ASR stream output : Fun-ASR-Nano — SeedTTS EN full 1088, c32, 3 repeats + warmup, 1x H100

**-speed for Non-Stream and Stream**

Metric | Non-stream1 | 2 | 3 | Stream1 | 2 | 3
-- | -- | -- | -- | -- | -- | --
evaluated samples  | 1088/1088 | 1088/1088 | 1088/1088 | 1088/1088 | 1088/1088 | 1088/1088
throughput (samples/s) | 91.41 | 87.20 | 88.70 | 90.80 | 97.77 | 97.01
mean latency (s) | 0.347 | 0.364 | 0.358 | 0.349 | 0.324 | 0.326
p95 latency (s) | 0.545 | 0.557 | 0.520 | 0.572 | 0.491 | 0.586
p99 latency (s) | 0.611 | 0.727 | 0.679 | 0.725 | 0.611 | 0.691
mean RTF | 0.075 | 0.079 | 0.077 | 0.075 | 0.070 | 0.070
p95 RTF | 0.117 | 0.1243 | 0.1224 | 0.1294 | 0.1054 | 0.1201

**-Stream-only metrics**

Metric | Stream
-- | --
text_ttft_mean_s | 0.176
text_ttft_median_s | 0.166
text_ttft_p95_s | 0.300
text_ttft_p99_s | 0.453
inter_chunk_mean_s | 0.069
inter_chunk_p95_s | 0.149
inter_chunk_p99_s | 0.214

--------

##Reference

https://github.com/sgl-project/sglang-omni/pull/1078#issuecomment-5010448523
[...]

PR #1676, opened 2026-08-24
[ASR] Enable ARK-ASR streaming transcription output

guozhihao-224:
## Motivation
HTTP `/v1/audio/transcriptions` already accepts `stream=true`, but ARK-ASR never registered a `stream_output_builder`, so decode produced no SSE deltas and clients only saw the terminal transcript. This wires ARK-ASR onto the same token-text streaming path as Fun-ASR (#1094) and Qwen3-ASR (#1359), reusing the shared helper from #1275.

## Modifications
- Add `make_arkasr_stream_output_builder` and hang it on `ArkasrEngineBuilder.extra_scheduler_kwargs`.
- Default `stream_emit_interval_s=0.05`. First delta is immediate; EOS and other terminal finishes flush pending text (`builder.flush`).
- Drop the same reserved markers as the result adapter (`<tool_call>`, `<|audio|>`, other `<...>` added tokens) so they cannot leak into deltas.
- Document `stream=true` and the SeedTTS `--stream` client in `docs/cookbook/arkasr.md`.
- Cover request-contract gating, chunked prefill, rate-limit / terminal flush, UTF-8 hold, per-request isolation, and marker suppression in `tests/unit_test/arkasr/test_stream_output_builder.py`. Streaming primarily reduces time to first text; it  does not change the final transcript.

## Benchmark
ARK-ASR-3B — SeedTTS EN full 1088, c=32, 3 repeats + warmup, 1x H200, DP=1.
**Speed for non-stream and stream**

| Metric | Non-stream1 | 2 | 3 | Stream1 | 2 | 3 |
| evaluated samples | 1088/1088 | 1088/1088 | 1088/1088 | 1088/1088 | 1088/1088 | 1088/1088 |
| throughput (samples/s) | 157.77 | 150.47 | 160.84 | 152.63 | 152.47 | 151.57 |
| mean latency (s) | 0.200 | 0.210 | 0.197 | 0.207 | 0.207 | 0.208 |
| p95 latency (s) | 0.271 | 0.282 | 0.262 | 0.283 | 0.282 | 0.281 |
| p99 latency (s) | 0.435 | 0.616 | 0.291 | 0.319 | 0.312 | 0.316 |
| mean RTF | 0.0434 | 0.0456 | 0.0426 | 0.0450 | 0.0451 | 0.0453 |
| p95 RTF | 0.0638 | 0.0655 | 0.0608 | 0.0651 | 0.0649 | 0.0639 |
| corpus WER | 0.0113 | 0.0113 | 0.0113 | 0.0111 | 0.0113 | 0.0113 |

Corpus WER stays **0.0113**

**Stream-only metrics**

| Metric | Stream1 | 2 | 3 | mean |
|---|---:|---:|---:|---:|
| text_ttft_mean_s | 0.111 | 0.112 | 0.114 | 0.112 |
| text_ttft_median_s | 0.113 | 0.115 | 0.117 | 0.115 |
| text_ttft_p95_s | 0.155 | 0.154 | 0.162 | 0.157 |
| text_ttft_p99_s | 0.204 | 0.197 | 0.202 | 0.201 |
| inter_chunk_mean_s | 0.048 | 0.047 | 0.047 | 0.048 |
| inter_chunk_p95_s | 0.079 | 0.081 | 0.082 | 0.081 |
| inter_chunk_p99_s | 0.105 | 0.097 | 0.098 | 0.100 |

## Test plan
- [x] `pytest tests/unit_test/arkasr/test_stream_output_builder.py tests/unit_test/arkasr/test_pipeline.py`
- [x] HTTP smoke: `stream=true` on `tests/data/query_to_cars.wav` emits delta / done / `[DONE]`, final text matches non-stream
- [x] SeedTTS EN full 1088 at c=32, 3 repeats + warmup, non-stream and `--stream`

## Checklist
- [x] Format & lint
- [x] Unit tests added
- [x] Cookbook streaming docs aligned with Fun-ASR
- [x] Provide throughput / latency / WER for stream vs non-stream


sglang_omni/models/arkasr/request_builders.py
```diff
@@ -257,4 +261,49 @@ def result_adapter(data: ArkASRRequestData) -> StagePayload:
...
+        eos_token_id
+        if eos_token_id is not None
+        else (int(tokenizer_eos) if tokenizer_eos is not None else None)
+    )
+    # note (guozhihao): same belt-and-suspenders drop as result_adapter;
+    # skip_special_tokens does not strip non-special added markers such as
+    # <tool_call>.
+    suppressed = set(_build_suppressed_token_ids(tokenizer))
+
+    def _decode_stream_ids(ids: list[int]) -> str:
```
AkazaAkane, line 280: One non-blocker: result_adapter() strips the complete decoded transcript, while the stream decoder emits raw chunks. A transcript with leading or trailing whitespace can therefore make concatenated transcript.text.delta content differ from transcript.text.done and the non-stream response. The current benchmark only scores the done event, so it does not catch this. Could we add a reconstruction invariant—at minimum "".join(deltas).strip() == done_text—or explicitly document that the done event is authoritative?

AkazaAkane (approved):

sglang_omni/models/arkasr/request_builders.py
guozhihao-224: @AkazaAkane Thanks — that's the right catch.
The mismatch is only leading/trailing whitespace: `result_adapter` still `.strip()`s the full decode (same string as non-stream `text` / `transcript.text.done`), while `decode_fn` emits raw incremental pieces. This matches the OpenAI / vLLM-Omni contract: render `transcript.text.delta` live, persist `transcript.text.done`. Concatenating deltas is a preview, not the product.

Follow-up:
- cookbook: `done` is authoritative; `"".join(deltas)` may differ by leading/trailing whitespace
- unit test: `"".join(deltas).strip() == result_adapter text`, including a suppressed marker in the middle

Happy to adjust if you'd rather Fun-align by dropping the final `.strip()` in a later PR

SandyLuXY (approved):
lgtm

Issue #1588, opened 2026-08-17
[RFC] Intel XPU Support Roadmap H2'26


siju-samuel:
# Intel XPU Support Roadmap

Ref: #1081, #994

Tracking issue for the remaining work.

### Features

- [x] XPU platform layer and stage device resolution (`xpu:N` placement)
- [x] Tensor parallelism over `xccl`
- [x] Host shared-memory relay for cross-stage transfers (CUDA IPC is CUDA-only)
- [x] Triton MoE runner
- [x] BF16 execution with SDPA / Triton attention
- [ ] Close the performance gap
  - [x] Enable XPU graph capture (`torch.xpu.XPUGraph`) #1679
    - [x] Decode graph #1679
    - [ ] Breakable prefill graph — needs the sglang bump
  - [ ] Support the `intel_xpu` attention backend instead of falling through to Triton
  - [ ] Evaluate the `sgl-kernel-xpu` fast paths (unquant linear, fused MoE, vision attention)
- [ ] Quantization
  - [ ] Tracked as part of #1594 

### CI

- [x] XPU install manifest, install script and Docker image
- [x] Enable CI for XPU in SGLang Omni
  

### Model support

  - [x] Qwen3-ASR
  - [x] Qwen3-TTS Base
  - [x] Qwen3-Omni thinker (`--text-only`)
  - [x] Qwen3-Omni speech
  - [x] Qwen3-TTS CustomVoice
  - [ ] MOSS-TTS
  - [ ] MOSS-Transcribe-Diarize
  - [ ] Whisper
  - [ ] FishAudio S2-Pro
  - [ ] Higgs-TTS
  - [ ] Voxtral-TTS
  - [ ] dots.tts
  - [ ] Ming-Omni
  - [ ] Ming-TTS
  - [ ] MiniMax Music 3
  - [ ] Zonos2
  - [ ] LLaDA2-Uni
  - [ ] ARK-ASR
  - [ ] FUN-ASR

cc: @hshen14 @Ratish1 @rbabukv @mingfeima @zhaochenyang20

zhaochenyang20:
[...]

PR #1679, opened 2026-08-24
[Perf][XPU] Request the decode CUDA graph for Omni's AR engines on Intel XPU

siju-samuel:
## Motivation

SGLang keeps XPU decode-graph capture opt-in — `_handle_xpu_backends()` disables it
unless a backend is named explicitly — and Omni never named one, so every AR engine
decoded eagerly on XPU even though `XPUGraphRunner` / `FullXPUGraphBackend` were
available. CUDA already defaults to `full`, so the gap only cost XPU.

## Modifications

A platform hook names the decode backend and the shared `ServerArgs` builder passes
it on, the way the code2wav graph decision already works. XPU answers `full` (the
only decode backend SGLang accepts there); every other platform answers `None`.

SGLang applies the per-phase backend *after* its disable switches, so the hook is
guarded: an engine that set `disable_cuda_graph`, or that names its own backend, is
left alone, and the backend is keyed off the device the stage will really run on so a
stage naming `cpu` on an accelerator host keeps `cpu`.

Two Qwen3-Omni AR stages decline on XPU, gated the way `enable_code2wav_graph`
already is:

- **talker** — capture kills the pipeline: the predictor's default SDPA dispatch is
  not capturable on this backend.
- **thinker** — capture serves one request and then fails every request under
  concurrency. Once the captures close, the scheduler thread's stream still reports
  capturing, so reading the sampled ids to host raises `wait method cannot be used
  for an event associated with a command graph`. The same read on a freshly created
  stream succeeds, so the fault is in the capture, not in this pipeline.

Both gates reopen when the runtime stops leaving that stream recording.

**CUDA is unaffected by construction, and it is verified rather than argued.** The
gates are read at call time, so importing the stage module does not pull in the
platform layer (a test holds that); the talker's existing call is untouched and the
gate declines afterwards; and the decode-backend helper asks the platform first and
returns on that branch before reading or writing any argument.

## Related Issues

Part of #1588 (Intel XPU roadmap — "Enable XPU graph capture"). Follows #994.

## Accuracy Test

On Intel Arc Pro B60 with the decode graph active: Qwen3-ASR transcripts unchanged;
Qwen3-TTS Base and CustomVoice round-trip TTS→ASR at 0.0% WER. Qwen3-Omni speech at
TP=8 serves 20/20 requests at concurrency 16 with both gates closed. Higgs TTS opts
out and is unchanged. Full unit suite on XPU: failure set identical to the base, name
for name.

## Benchmark & Profiling

One card pinned, A/B/A to rule out host drift. RTF (elapsed / produced audio) rather
than raw latency where output length is not pinned: latency medians drift 3.6x
between two identical runs while RTF repeats within 1.6%.

| Model | graph | eager | speedup |
|---|---|---|---|
| Qwen3-TTS Base (n=18, rtf median) | 1.03 | 1.42 | 1.38x |
| Qwen3-ASR (n=12, latency median) | 233 ms | 402 ms | 1.7x |
| Qwen3-TTS CustomVoice (n=18, latency median) | 6144 ms | 7527 ms | 1.23x |

Startup capture cost on TTS: 33 s eager → 66-205 s with capture.

Capturing the **thinker** decode was measured at 7.6x on TP=8 (2166 ms vs 16528 ms)
but is disabled here for the reason above: 0/20 requests succeed at concurrency 16
with the graph, 20/20 with the gate closed. That row returns when the runtime is
fixed.

## Verification

```bash
# CUDA sees byte-identical ServerArgs and an unchanged import graph.
python - <<'PY'
import json, sys
before = set(sys.modules)
import sglang_omni.models.qwen3_omni.stages  # noqa: F401
added = set(sys.modules) - before
import sglang_omni.platforms as platforms
from sglang_omni.platforms.interface import OmniPlatform

class NonXpu(OmniPlatform):
    device_type = "cuda"

platforms.current_platform = NonXpu()
from sglang_omni.scheduling.sglang_backend import server_args_builder

args = server_args_builder.build_sglang_server_args(
    "<any local model>", context_length=448, device="cuda",
    max_running_requests=32, disable_cuda_graph=False,
    sampling_backend="pytorch", skip_server_warmup=True,
)
print("platforms pulled in:", "sglang_omni.platforms" in added)   # False
print("decode backend     :", args.cuda_graph_backend_decode)      # None
print("fields:", json.dumps({k: v for k, v in sorted(vars(args).items())
      if isinstance(v, (str, int, float, bool, type(None)))}, sort_keys=True)[:80])
PY

pytest tests/unit_test/test_platforms.py \
       tests/unit_test/scheduling/test_server_args_builder_device.py \
       tests/unit_test/qwen3_omni/test_sglang_ar_budget.py
```

## Checklist

- [x] Add unit tests.

### DIFF
--- a/sglang_omni/models/qwen3_omni/stages.py
+++ b/sglang_omni/models/qwen3_omni/stages.py
@@ -1035,6 +1035,10 @@ def create_sglang_thinker_executor_from_config(
         sampling_backend="pytorch",
     )
     overrides["tp_size"] = tp_size
+    from sglang_omni.platforms import current_platform
+
+    if not current_platform.enable_thinker_decode_graph():
+        overrides.setdefault("disable_decode_cuda_graph", True)
     has_explicit_colocated_mem_fraction = (
         total_gpu_memory_fraction is not None
         and overrides.get("mem_fraction_static") is not None
@@ -1153,6 +1157,12 @@ def create_talker_ar_executor_from_config(
         disable_cuda_graph=False,
         sampling_backend="pytorch",
     )
+    from sglang_omni.platforms import current_platform
+
+    # A platform may decline the default above; the caller's setting still wins.
+    stated_disable = "disable_cuda_graph" in (server_args_overrides or {})
+    if not stated_disable and not current_platform.enable_talker_graph():
+        overrides["disable_cuda_graph"] = True
     overrides["tp_size"] = tp_size
     _apply_colocated_ar_memory_contract(
         overrides,
--- a/sglang_omni/platforms/interface.py
+++ b/sglang_omni/platforms/interface.py
@@ -61,6 +61,15 @@ class OmniPlatform(DeviceMixin):
         """Check if current platform support Graph for code2wav in Qwen3-Omni"""
         return True
 
+    def enable_talker_graph(self) -> bool:
+        return True
+
+    def enable_thinker_decode_graph(self) -> bool:
+        return True
+
+    def get_decode_cuda_graph_backend(self) -> str | None:
+        return None
+
     def supports_torchaudio_resample(self) -> bool:
         """Check if current platform support torchaudio.functional.resample"""
         return True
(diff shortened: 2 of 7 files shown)


Ratish1 (approved):
LGTM

PR #1681, opened 2026-08-24
[Perf] Move Qwen3-ASR log-mel onto the encoder stream

guozhihao-224:
## Motivation

Qwen3-ASR unique-input (embedding cache off) was limited by CPU STFT on the request-build workers: each cache miss ran HuggingFace `WhisperFeatureExtractor` on the host, then H2D the fbank, while the encoder GPU sat at ~30–50% util. This PR moves log-mel onto the encoder stream so build workers only decode WAV, estimate frames from hop length, and attach a pinned waveform.

Secondary encoder-path cleanups ride along: pinned async D2H into the pre-LM embedding LRU (aligned with Whisper), and removing extra graph clones / `.item()` syncs in the conv preamble.

## Modifications

- Add `gpu_mel.py`: checkpoint-extractor-aligned GPU log-mel   (`stft[..., :-1]`, Slaney, log10, max-8, `(x+4)/4`) on the encoder  device; `bind_audio_frontend` fails at setup if hop/filters are missing.
- Request builder: on a cache miss, skip CPU STFT and the CPU ones-mask; admit with pinned waveform + hop-estimated `num_audio_tokens`.
- `get_audio_feature`: run GPU mel from the waveform; clone CUDA-graph output once before returning to LM / pre-LM cache.
- Pre-LM encoder cache: `pre_lm_cache_pin_host_memory` (default true) stages a non-blocking D2H into pinned host memory. Variable-length embeddings are not prewarmed into a fixed pool.
- Encoder CUDA graph: write `cu_seqlens` in place on a pinned host buffer; compute the chunk plan on CPU instead of `.item()`/`.tolist()`  mid-kernel.

Cache hits still return before mel. Numerics match the checkpoint extractor (unit-tested, including short clips).

## Related Issues

Related to the Qwen3-ASR serving work in #1324.
Part of https://github.com/sgl-project/sglang-omni/issues/1396

## Accuracy Test

- Unit: GPU log-mel matches `WhisperFeatureExtractor` on CPU (1600 / 16000 / 31000 samples) and CUDA vs CPU.
- Serving: SeedTTS EN 1088, unique-input, cache=0. Corpus WER **0.0122**  on HEAD and on this change (3264/3264, zero drops).

## Benchmark & Profiling

H200 × 1, `Qwen3-ASR-1.7B` @ `7278e1e70fe2`, SeedTTS EN 1088, embedding cache=0, warmup excluded. Do not compare absolute req/s across GPUs; deltas are vs HEAD in the same session.

**All three changes vs HEAD (GPU 1; c=1 n=1, c=8/16/32 n=3 mean):**

conc | n | HEAD req/s | this PR req/s | Δ QPS | HEAD p95 | this PR p95 | Δ p95 | HEAD p99 | this PR p99 | Δ p99 | WER
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | 1 | 18.12 | 18.82 | +3.9% | 0.070 | 0.063 | −11.0% | 0.091 | 0.074 | −18.6% | 0.0122
8 | 3 | 92.8 | 101.1 | +8.9% | 0.139 | 0.113 | −18.3% | 0.174 | 0.128 | −26.4% | 0.0122
16 | 3 | 124.0 | 135.6 | +9.4% | 0.168 | 0.161 | −4.3% | 0.190 | 0.203 | +6.7% | 0.0122
32 | 3 | 170.6 | 200.4 | +17.5% | 0.248 | 0.222 | −10.5% | 0.480 | 0.362 | −24.6% | 0.0122

**GPU mel only vs HEAD (GPU 0, n=3):** c=8 **−3.5%**, c=16 **+10.3%**,c=32 **+12.6%**. Unique-input gain at c=16/32 is this path; c=8 needs the preamble/graph cleanup or CPU STFT overlaps the previous encode.

## Checklist

- [x] Add unit tests.


JiaxinD:
Current-main single-replica qualification results

I replayed this change on the current `origin/main` snapshot and ran an online E2E matrix on one H200 (GPU0 only; no DP/MPS/NPS).

Revision identity:

- PR head (unchanged): `3caa56ba832effd3345e561027a54b91e2b1af5a` (the PR is currently conflicting with its recorded base)
- Baseline/current `origin/main`: `d346c46fff507f1e534eddb150c22953409537cf` (tree `204185828cc013182d8dce7566ada36b49b4d834`)
- Current-main replay candidate: `7c2014695d0d3328a02d506a04465bd24b8d6069` (tree `cb5be93672583846946e8e75012ad57d839c2cf9`)

Configuration: 1 replica, 1 H200, 1,088 Seed-TTS-eval samples per measurement, unique inputs, effective encoder cache disabled (`pre_lm_cache_size_bytes=1`, the minimum accepted value), CUDA Graph + `torch.compile`. Full matrix completed 20/20 measurements: c1 n=1 and c8/c16/c32 n=3 for each arm.

| Concurrency | Baseline QPS | Candidate QPS | QPS delta | p95 delta | p99 delta | GPU util mean (base -> cand) | WER (base -> cand) | Candidate QPS max/min spread |
| 1 | 19.571 | 20.368 | +4.07% | -3.06% | -3.83% | 53.33% -> 56.80% | 0.012258 -> 0.012031 | n=1 |
| 8 | 104.905 | 105.509 | +0.58% | +1.46% | +2.09% | 58.80% -> 64.54% | 0.012031 -> 0.012182 | 3.25% |
| 16 | 133.449 | 153.873 | +15.30% | -10.12% | -8.93% | 57.64% -> 62.13% | 0.012031 -> 0.012182 | 1.41% |
| 32 | 189.894 | 228.406 | +20.28% | -22.24% | -20.04% | 42.74% -> 50.79% | 0.012031 -> 0.012182 | 9.10% |

The table uses medians across the three repeats for c8/c16/c32. GPU utilization is the mean sampled during each benchmark measurement window; the c32 windows are short, so this is useful for same-protocol comparison rather than an absolute saturation claim.

An independent c32 baseline -> candidate -> baseline bracket also completed (1,088/1,088, no skips in every arm):

- baseline before: 185.763 QPS, p95 219.24 ms, p99 256.53 ms, WER 0.012031
- candidate: 210.802 QPS, p95 193.75 ms, p99 271.61 ms, WER 0.012031
- baseline after: 182.427 QPS, p95 214.63 ms, p99 255.31 ms, WER 0.012031
- candidate vs two-sided baseline median: +14.51% QPS, -10.69% p95, +6.13% p99
- baseline bracket QPS drift: 1.80%; WER was exactly equal across all three arms

Important caveats:

- c8 is effectively neutral on throughput and slightly worse on tail latency in the full matrix; the gain is concentrated at c16/c32.
- Candidate c32 had one noisy repeat (211.73 QPS / 416.40 ms p99); the other two were 228.41/231.01 QPS with 236.02/232.22 ms p99. This explains the 9.10% candidate spread and should remain visible rather than being hidden by the median.
- Full-matrix candidate c8/c16/c32 WER was higher by 0.000151 absolute (1.2031% -> 1.2182%), although the independent c32 bracket had exact WER equality. This needs a focused output-determinism/quality repeat before treating accuracy as fully qualified.

Artifact roots:

- independent c32 bracket: `/data02/jiaxind/single-replica-campaign/results-qwen3-asr-1681-c32-bcb-d346-7c201-r4`
- full matrix: `/data02/jiaxind/single-replica-campaign/results-qwen3-asr-1681-full-d346-7c201-r2`

Bottom line: the current-main replay shows a material single-replica E2E win at c16/c32 (+15.3%/+20.3% median QPS), reproduced by an independent c32 bracket (+14.5%). The remaining qualification items are the c32 tail/variance and the small high-concurrency WER discrepancy.

guozhihao-224:
@JiaxinD  cc
## Benchmark & Profiling

Rebased onto `origin/main` (`e7d876b2`) vs this PR (`fb1d8e60`).

H200 × 1, `Qwen3-ASR-1.7B` @ `7278e1e70fe2`, SeedTTS EN 1088. Sequential arms on the same GPU. Every timed pass 1088/1088, zero drops. Do not compare absolute req/s across GPUs or across the two tables; deltas are vs HEAD in the same session.

```bash
MODEL_PATH=$(hf download Qwen/Qwen3-ASR-1.7B \
    --revision 7278e1e70fe206f11671096ffdd38061171dd6e5)
```

## 1. Unique-input (embedding cache off)
Cache disabled so a full discarded 1088-pass cannot turn later repeats into LRU hits. One discarded full-set warmup per conc, then timed repeats. c=1 n=1; c=8/16/32 n=3 median.
**server** 
```
    --model-path "${MODEL_PATH}" \
    --port 8000 \
    --asr.factory.pre_lm_cache_size_bytes 1
```
**client** 
```
    --port 8000 \
    --model-revision 7278e1e70fe206f11671096ffdd38061171dd6e5 \
    --dataset-revision 27f4c1adee83b5b29b7c4b375f6b976324bda308 \
    --lang en --warmup \
    --concurrencies 1 --repeats 1

    --port 8000 \
    --model-revision 7278e1e70fe206f11671096ffdd38061171dd6e5 \
    --dataset-revision 27f4c1adee83b5b29b7c4b375f6b976324bda308 \
    --lang en --warmup \
    --concurrencies 8,16,32 --repeats 3
```


conc | n | HEAD req/s | PR req/s | Δ QPS | HEAD p95 | PR p95 | Δ p95 | HEAD p99 | PR p99 | Δ p99 | WER (HEAD / PR)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | 1 | 20.62 | 21.65 | +5.0% | 0.057 | 0.056 | −2.5% | 0.066 | 0.080 | +21.7% | 0.0123 / 0.0120
8 | 3 | 109.1 | 111.6 | +2.3% | 0.103 | 0.105 | +2.2% | 0.114 | 0.112 | −1.4% | 0.0120 / 0.0122
16 | 3 | 137.6 | 150.4 | +9.4% | 0.142 | 0.130 | −8.9% | 0.163 | 0.148 | −9.2% | 0.0120 / 0.0122
32 | 3 | 200.4 | 236.2 | +17.9% | 0.200 | 0.175 | −12.4% | 0.247 | 0.224 | −9.2% | 0.0120 / 0.0122

c=1 p99 is n=1 noise. Read c=16/32.

## 2. Default cache, warmup=64, n=1
Cache left at the default 2GiB. 64 discarded copies of the first clip, then one timed 1088-pass. Server restarted between concs so the LRU cannot leak. benchmark_asr_seedtts --warmup is a full 1088-pass, so this client uses warmup=64.

**server**
```
    --model-path "${MODEL_PATH}" \
    --port 8000
```

**client**
```
python - <<'PY'
import asyncio
from benchmarks.dataset.prepare import DATASETS
from benchmarks.dataset.seedtts import load_seedtts_samples
from benchmarks.eval.benchmark_asr_seedtts import run_asr_seedtts_once

samples = load_seedtts_samples(DATASETS["seedtts"], max_samples=1088, split="en")

async def main():
    for conc in (1, 8, 16, 32):
        result = await run_asr_seedtts_once(
            samples,
            host="127.0.0.1",
            port=8000,
            model_path="Qwen/Qwen3-ASR-1.7B",
            lang="en",
            concurrency=conc,
            warmup=64,
        )
        speed = result["speed"]
        print(
            conc,
            speed["throughput_samples_per_s"],
            speed["latency_p95_s"],
            speed["latency_p99_s"],
            result["summary"]["corpus_wer"],
        )

asyncio.run(main())
PY
```

Restart the server between concs before running the next concurrency.

conc | n | HEAD req/s | PR req/s | Δ QPS | HEAD p95 | PR p95 | Δ p95 | HEAD p99 | PR p99 | Δ p99 | WER (HEAD / PR)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | 1 | 23.74 | 24.03 | +1.2% | 0.051 | 0.051 | +0.6% | 0.057 | 0.066 | +15.1% | 0.0123 / 0.0123
8 | 1 | 100.6 | 99.0 | −1.6% | 0.111 | 0.112 | +0.8% | 0.127 | 0.159 | +25.5% | 0.0121 / 0.0121
16 | 1 | 134.9 | 141.7 | +5.1% | 0.154 | 0.157 | +1.6% | 0.176 | 0.186 | +5.5% | 0.0121 / 0.0120
32 | 1 | 175.3 | 209.0 | +19.2% | 0.212 | 0.201 | −5.4% | 0.761 | 0.263 | −65.5% | 0.0120 / 0.0123

c=8 Δ QPS and c=1/32 p99 are n=1 noise. For c=32 latency read p95. conc=32 Δ QPS is +17.9% in table 1 and +19.2% in table 2.

tests/unit_test/qwen3_asr/test_gpu_mel.py
```diff
@@ -0,0 +1,85 @@
...
+    audio = rng.standard_normal(num_samples).astype(np.float32)
+    reference = _reference_mel(extractor, audio)
+
+    model = type("M", (), {})()
+    bind_audio_frontend(model, extractor)
+    waveform = torch.from_numpy(audio)
+    got = log_mel_spectrogram(waveform, model._audio_frontend)
+
+    assert got.shape == reference.shape
+    assert torch.allclose(got, reference, atol=1e-4, rtol=1e-4)
```
JiaxinD, line 55: All three inputs are white noise with 3.8 to 6.0 decades of dynamic range, so the max-8 clamp in gpu_mel.py never fires: deleting that line leaves all three cases passing at maxabs 0. The CUDA test below compares this implementation against itself, never against the extractor.

sglang_omni/models/qwen3_asr/gpu_mel.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def log_mel_spectrogram(
+    waveform: torch.Tensor,
+    frontend: AudioFrontend,
+) -> torch.Tensor:
+    """Return log-mel of shape n_mels by n_frames on the waveform device.
+
+    Matches the checkpoint feature extractor's fbank for a 1-D float32
+    waveform: drop the Nyquist STFT bin, Slaney mel, log10, max-8, then (x+4)/4.
```
JiaxinD, line 87: On CPU this is bit-identical to the extractor, but production runs cuFFT and cuBLAS, moving about 0.1% of mel bins by one bf16 ULP after the tower cast. That is the only remaining explanation for the 0.012031 to 0.012182 WER both runs reproduced at c=8/16/32, so "numerics match" overstates it.

sglang_omni/models/qwen3_asr/gpu_mel.py
```diff
@@ -0,0 +1,119 @@
...
+        frontend.n_fft,
+        frontend.hop_length,
+        window=frontend.hann_window,
+        return_complex=True,
+    )
+    # note (guozhihao-224): dropping the Nyquist stft bin makes the frame
+    # count equal to samples divided by hop length. the request builder
+    # estimates tokens from hop length instead of running the cpu extractor,
+    # and must not pad or truncate to a 30s window (transformers issue 26241).
+    magnitudes = stft[..., :-1].abs() ** 2
```
JiaxinD: magnitudes is a strided view here, and transformers carries a regression test for exactly this expression after a roughly 8x strided GEMM slowdown on ROCm. A .contiguous() before the matmul keeps the output identical, and qwen3_asr already has ROCm branches.

sglang_omni/models/qwen3_asr/request_builders.py
```diff
@@ -303,6 +296,7 @@ def request_builder(
                 # split batched encoder output and key its embedding cache.
                 "num_audio_tokens": num_audio_tokens,
                 "audio_fingerprint": fingerprint,
+                **({"waveform": waveform} if waveform is not None else {}),
```
JiaxinD, line 297: Nothing ever clears model_specific_data["waveform"], so the pinned waveform stays alive through the whole LM decode, where the base freed the mel via item.feature = None in attach_embedding. Roughly 1.9 MB of page locked memory per 30s clip per in-flight request.

sglang_omni/models/qwen3_asr/encoder_service.py
```diff
@@ -115,6 +115,7 @@ def __init__(
         cache_max_bytes: int = _CACHE_MAX_BYTES,
         max_batch_size: int = 8,
         max_batch_wait_ms: int = 0,
+        pin_host_memory: bool = True,
```
JiaxinD, line 118: This flips the 2 GiB pre-LM LRU from pageable to page locked by default, with variable length blocks the caching host allocator never returns to the OS. Worth a pinned bytes cap separate from cache_max_bytes, or defaulting this to False.

sglang_omni/models/qwen3_asr/sglang_model.py
```diff
@@ -248,7 +293,12 @@ def get_audio_feature(self, items: List[MultimodalDataItem]) -> torch.Tensor:
                 )
                 graphed = runner.run(hidden, window_lens)
                 if graphed is not None:
-                    return graphed.unsqueeze(0)
+                    # note (guozhihao-224): graph run returns a view of the static
+                    # buffer. clone before leaving get_audio_feature so a later
+                    # replay cannot mutate lm embeddings (pre-lm off) or any
+                    # caller that holds this tensor across encodes.
+                    # split_embeddings still copies packed slices for the lru.
+                    return graphed.unsqueeze(0).clone()
```
JiaxinD, line 285: The clone moved here from run(), it was not removed: the net count is still 2 (this one plus split_embeddings). Worth rewording "removes extra graph clones" in the body so it is not read as a copy that went away.

tests/unit_test/qwen3_asr/test_gpu_mel.py
guozhihao-224: fixed in [7c61657](https://github.com/sgl-project/sglang-omni/pull/1681/commits/7c616577cc35192466c932a8caaf0d8bdc994a3a):
Fixed both: the clamp test now drives the mel span to the max-8 bound with an impulse-then-silence waveform, and the CUDA test compares against the checkpoint extractor instead of our own CPU path.

sglang_omni/models/qwen3_asr/gpu_mel.py
guozhihao-224: fixed in 7c61657 : 
Softened: CPU matches the extractor bit-for-bit; CUDA via cuFFT/cuBLAS
can move a small fraction of bins ~1 ULP after the audio-tower cast.
Unique-input SeedTTS EN corpus WER moved 0.0120 -> 0.0122, consistent with that.

sglang_omni/models/qwen3_asr/gpu_mel.py
guozhihao-224: fixed in 7c61657: magnitudes are made contiguous before the mel GEMM, so the STFT strided view no longer hits the ROCm slow path.

sglang_omni/models/qwen3_asr/request_builders.py
guozhihao-224: fixed in 7c61657:
the waveform is popped from model_specific_data right after H2D + log-mel, and again in attach_embedding for the pre-LM path.

sglang_omni/models/qwen3_asr/sglang_model.py
guozhihao-224: fixed in 7c61657:
The clone moved from graph run() to the get_audio_feature exit to protect the static buffer on replay; split_embeddings still clones the packed slice. Motivation updated to say "moved", not "removed".

guozhihao-224:
Compare performance with pin_host_memory set to true and false 

conc | pin=True QPS | pin=False QPS | ΔQPS | pin=T p95 | pin=F p95 | Δp95 | WER
-- | -- | -- | -- | -- | -- | -- | --
1 | 21.23 | 20.97 | +1.2% | 0.055 | 0.057 | -3.2% | 0.0121 / 0.0120
8 | 113.98 | 113.63 | +0.3% | 0.103 | 0.102 | +1.0% | 0.0122 / 0.0122
16 | 147.02 | 150.17 | -2.1% | 0.134 | 0.135 | -0.5% | 0.0122 / 0.0122
32 | 229.32 | 217.00 | +5.7% | 0.178 | 0.195 | -8.5% | 0.0122 / 0.0122

sglang_omni/models/qwen3_asr/encoder_service.py
guozhihao-224: fixed in [e5c0f78](https://github.com/sgl-project/sglang-omni/pull/1681/commits/e5c0f78b1455990ca0619e242e7e47d94a55fd0e):
Kept pin=true, and split the budgets: pre_lm_cache_size_bytes is still the 2GiB LRU; pre_lm_cache_pinned_size_bytes (512MiB default) is the page-locked host cap. Overflow entries stay in the LRU as pageable. --asr.factory.pre_lm_cache_pin_host_memory false still turns pinning off.

PR #1683, opened 2026-08-24
fix(pipeline): keep the child startup traceback in wait_ready

jzheng17:
Based on `pd_runtime` rather than the stack tip, because this touches only `stage_workers.py` and does not depend on PR 4. Context is in #1599.

## The problem

`wait_ready` tests `proc.is_alive()` before reading the startup error channel, and reads with `get_nowait()` on the death path. A child that fails in its factory writes its traceback and exits at once, so both choices race the report: the queue feeder may not have delivered the bytes when the parent looks, and the parent then raises a bare exit code or a plain `TimeoutError` with no cause.

## The change

Read the channel first, and keep a short blocking read on the death path so a traceback still in flight is not lost. Behaviour on success is unchanged.

## Testing

A second commit pins the contract with fake processes and fake channels, so the result does not depend on how long a real child takes to start. Four cases: a dead child whose traceback is still in flight, a live child that already reported, a dead child that reported nothing, and a live silent child. Only the first changes behaviour, and it fails without the fix.

`tests/unit_test/pipeline` gives 531 passed, 1 failed. The failure is `test_mp_runner_startup_failure_includes_child_factory_traceback`, which is flaky for a separate reason this PR does not address: it allows `timeout=10.0` while a child spawn costs 20-35s on the host I used, and nine unrelated tests earlier in `test_ipc.py` each reproduce it when run first. The right timeout depends on the CI host rather than mine.

### DIFF
--- a/sglang_omni/pipeline/stage_workers.py
+++ b/sglang_omni/pipeline/stage_workers.py
@@ -196,6 +196,17 @@ def _patched_spawn_env(spec: StageWorkerProcessSpec):
                 os.environ[key] = value
 
 
+
+def _read_startup_error(channel: Any, timeout: float | None = None) -> str | None:
+    """Return a child's startup traceback, or None if it reported none yet."""
+    try:
+        if timeout is None:
+            return channel.get_nowait()
+        return channel.get(timeout=timeout)
+    except (queue.Empty, EOFError, OSError, ValueError):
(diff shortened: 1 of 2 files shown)


jzheng17:
Folded into #1712, which carries the same change on `main` together with the other two operator-surface fixes. Reviewing one PR there is less work than three.

PR #1684, opened 2026-08-24
feat(pd): free placement and sizing — per-half budgets and shared GPUs

jzheng17:
Based on `pd_runtime` rather than the stack tip, because it does not depend on PR 4. Context for the series is in #1599.

## Where this is going

A prefill/decode split is a decision about how to divide GPU resources between two kinds of work, and the right division depends on the workload. On Qwen3-Omni thinker the prefill share of a request's GPU work is small — a 58 ms first forward against a 2.2 s decode tail — and a one-prefill-one-decode pair therefore splits hardware 50/50 against a workload split closer to 3/97. Measured on two H200s, that pair's prefill card ran at **10.1% mean utilization against 67.9% for a colocated card**. Numbers and method are in #1696.

So the resource division wants to be a parameter, not a fixed device separation: a prefill half should be able to take a fraction of a card rather than a whole one, several halves should be able to share a device, and the ratio should be able to move. This PR is the configuration surface that makes those expressible. #1690 is the routing seam that lets the ratio vary at request time; neither is a ratio on its own, and both are prerequisites for one.

## Nothing turns PD on

`pd_disaggregation` is a `StageConfig` field and nothing set it, so a deployment could not ask for the split. It is set the way every per-stage setting is set, in the file or on the command line:

```yaml
stages:
  thinker:
    pd_disaggregation:
      prefill: {gpu: 0}
      decode: {gpu: 1}
```

```bash
--thinker.pd_disaggregation.prefill.gpu 0 --thinker.pd_disaggregation.decode.gpu 1
```

This needs no flag and no parser. `iter_schema_paths` walks the schema, so every leaf under `pd_disaggregation` is already addressable, and the halves inherit precedence between CLI, file and model default, duplicate detection, and `sgl-omni config explain` provenance.

An earlier revision of this PR added `--pd-stage STAGE=PREFILL:DECODE` with an optional `@SHARE`, backed by 193 lines of parsing. The canonical config core made all of it unnecessary and it is removed.

## The server args PD requires

`bind_pd_runtime` refuses a stage unless `disable_radix_cache` is set and `page_size` is 1, and nothing sets them, so a PD pipeline would compile and then fail at bind time. The compiler now writes both into each generated half's `engine` block and rejects a value that contradicts one. Tests pin those constraints, including that decode CUDA graph is **not** among them.

They go where SGLang ServerArgs live, which since the canonical config core is the stage's `engine` block. That block exists only on stage types declaring `engine_stage`, so the pass asks that rather than inspecting a factory signature, and a PD half of another kind is left alone rather than handed a block it may not carry. `EngineArgs` allows free-form keys and reports them from `overrides()`, so neither key has to be declared.

## Each half carries its own settings

`PDStagePlacement` gains `engine` and `memory_fraction`. The two halves carry different work, so a setting that suits one need not suit the other, and an experiment that varies one has to hold the other fixed. Measured: a pair's decode card sustains about a third of the decode throughput of a colocated card while its prefill card matches one, so the halves are not interchangeable in practice either.

## The halves may share a GPU

The separation PD needs is the process split, which happens either way; the device split is a placement choice. Sharing a card makes PD runnable on a one-GPU box and in CI, and it is the placement that suits a small prefill share.

The check that forbade it landed with no stated rationale, and the adjacent message — "requires explicit prefill.gpu and decode.gpu for real PD" — reads as definitional. There is a real hazard behind it, but it is avoidable and it is not what the message says: **`mem_fraction_static` is not order-safe when two halves size pools from one card.** Both halves serialize on the same startup lock and the winner varies between runs. At 0.93 on both, whichever won took 780,987 KV tokens and the other died with `torch.OutOfMemoryError`; assigning asymmetric fractions does not help, because you cannot know which half loads first.

Two halves on one card are two process groups sharing a GPU like any other, and the repo already has a policy for that: `_validate_gpu_memory_fractions` requires each to declare `gpu_memory_fraction` and caps their sum. That budget routes through `calculate_stage_budget_available_bytes`, which computes `total x fraction - own usage`. **Total is a constant**, so the result does not depend on load order.

Measured, both halves on GPU 0, no `mem_fraction_static` and no `max_total_tokens`: both resolved to **21,502 KV tokens, identically, under both lock orders** — decode-first in four runs, prefill-first in one. The second half to load starts with about 57 GiB already resident and still computes the same `stage_load_used=57.00GiB`. Correct completions on text and image. Over-claiming fails loudly and on both halves rather than on whichever loses the race: at 0.50 each, `GPU 0 total_gpu_memory_fraction=1.060 exceeds placement limit 1.000`, refused before any GPU work.

## The budget has to reflect what is on the card

Two processes sizing pools from one device is where #1685's defects bite, so this PR fixes all three.

**A failed attribution is no longer reported as zero.** `get_process_gpu_memory_bytes` matched `os.getpid()` against pids NVML lists for the device. NVML reports host pids, so a caller inside a container never matches itself and every containerized run fell through — returning `0`, which is also what a process holding no memory returns. It now returns `None`, which its own docstring already promised, and the caller routes only that to the fallback.

**When attribution fails, torch is asked before the global fallback.** That fallback profiles a free-memory delta across model load, and its error depends on what was resident when the "before" sample was taken: the same script and config produced 838,138 and 711,261 KV tokens on two runs. `torch.cuda.memory_reserved` needs no pid join and undercounts only by the CUDA context, measured at 0.51 GiB and identical every run.

**The budget is clamped by what the device reports free.** It counted only the stage's own usage, so it over-committed whenever anything else held memory on the card — a stage at `fraction=0.87` on a card holding 34 GiB of another process budgeted 122.49 GiB and ran out of memory at launch. Callers that cannot read free memory pass nothing and keep the previous behaviour.

## The relay pool can be allocated at launch

`CudaIpcRelay` allocates its pool on the first payload that crosses a process boundary, not at construction, so a budget that cannot hold it starts cleanly and fails later. Measured on a PD pair: at 0.87 both halves launched, served text for 19 minutes, and failed a 750 MiB allocation the moment the image arm began.

**Text traffic cannot catch this.** A payload with no CUDA tensors routes over SHM, so `mm_aggregate` never reaches the pool on a text-only run — a text smoke test passes and the deployment is still a gigabyte short. PD makes it reachable in a way colocated is not: `qwen3_omni/config.py` puts every stage in `process="pipeline"`, so that edge is in-process, while `_split_pd_stage` gives the prefill half its own process. Colocated logged zero pool allocations across the whole comparison; PD logged one.

`preallocate_pool` is off by default. It does not make the pool part of any budget — the sizing path still does not know the relay will allocate.

## Prior art

Intra-GPU prefill/decode separation is established. [Nexus](https://arxiv.org/abs/2507.06608) reports 1.5-1.9x throughput over vLLM on a single GPU and matches disaggregated vLLM using half the GPUs; vLLM's [MORI-IO work](https://vllm.ai/blog/2026-04-07-moriio-kv-connector) reports 2.5x goodput disaggregating within one 8-GPU node.

**Their mechanism is not this one.** Nexus runs one process with one copy of the weights, shares the KV cache in place, and partitions SMs with CUDA Green Contexts. Two processes on one card carry two copies of the weights and transfer the KV. So this change makes the placement *possible*; it is not a claim that it is fast. At two copies of the weights the KV left over is small — 21,502 tokens per half against 677,613 on a colocated card — and closing that gap needs the halves to share one copy, which is separate work.

## The guards this surface has to satisfy

`bind_pd_runtime` refuses `tp_size != 1`, `page_size != 1` and an enabled radix cache. Two of the three are now set at compile time here, so tests pin all three, and pin that CUDA graph is **not** among them: PD does not require it off, and turning it off is expensive.

This was #1695 and is folded in here.

## Testing

586 unit tests pass on `tests/unit_test/pipeline`, `tests/unit_test/relay` and `tests/unit_test/scheduling`.

The flag was also exercised end to end on two H200s. `--pd-stage thinker=0:1` parses the assignment, splits `thinker` into `thinker_prefill` and `thinker_decode`, and the launcher spawns them as separate stage processes on the two cards.

Three failure modes surfaced in that run and are now documented rather than left to be rediscovered:

- A colocated config class cannot host a split thinker. `Qwen3OmniSpeechColocatedPipelineConfig` requires the image encoder, audio encoder, thinker, talker and code2wav to share one GPU, and reports the conflict as a request for exactly one GPU id for the thinker. #1716 adds a profile that works.
- A stage-level `gpu_memory_fraction` together with an `engine.mem_fraction_static` for the same stage is rejected as two memory contracts for one stage.
- A thinker budget carried over from a colocated profile runs the prefill half out of memory during startup, because the encoders and the relay pool are on that card too. 0.75 fails; the example config uses 0.68.

## Depends on

The thinker becomes PD-capable in #1599, which decorates its factory with `@pd_disaggregation_capable` and adds the resume schema that lets the continuation serialize. On this PR's base the capability gate correctly rejects the thinker, so `--pd-stage thinker=...` reaches a clear configuration error rather than a running pair. The end-to-end run above was therefore made on this branch merged with #1599.

## Not included

Order-independence here rests on the startup lock serializing weight loading, not on the formula alone, and the placement check only accounts for stages the pipeline knows about — a foreign process on the card is still unaccounted, which is #1685's remaining scope. `--pd-stage` also moves only the stage it names: the encoders keep their GPU, so on a two-GPU multimodal deployment the prefill card carries their activations as well.

## Documentation

The user-facing page for this surface is #1716, kept separate so the mechanism and the prose review independently. It merges after this one.

## Rebased onto the canonical config core

Upstream #1494 replaced the interfaces this PR was built on: `engine_overrides` inside `factory_args` became a typed `engine` block, `factory_args` became `factory`, `factory` became `factory_path`, and `runtime.resources.total_gpu_memory_fraction` became a stage-level `gpu_memory_fraction`. The per-half settings follow, so a half's engine args merge into its `engine` block and a half's share writes `gpu_memory_fraction` directly.


MelodyyyYin:
#note:（yue yin）

The CUDA relay pool remains lazy by default and its `preallocate_pool` switch is not wired through routing or placement. An image-capable placement may pass readiness and later OOM on the first CUDA payload, while the pool is absent from the stage’s explicit GPU budget.

Follow-up PR: https://github.com/jzheng17/sglang-omni/pull/4

Why: every route-capable persistent GPU allocation must be budgeted and validated before ready, while routes that cannot use it must not pay the reservation.

jzheng17:
Superseded. The main line moved to #1773, and this PR's substance landed there via #1784: `PDStagePlacement.memory_fraction` and `.engine`, the split threading them instead of copying one `gpu_memory_fraction` to both halves, and same-GPU placement.

One thing changed on the way in, and it was a defect here. This PR removed the same-GPU ban citing `_validate_gpu_memory_fractions` — that function does not exist; the string appears only in the comment. The real validator is `_validate_memory_budgets`, which caps the sum of *declared* fractions, and `_build_gpu_placement` skips a `None`. So dropping the ban as written turned a config error into a silent OOM: two undeclared halves on one card would each size against the whole of it, and whichever won `gpu_startup_lock` would take the memory. In #1784 same-GPU requires both halves to declare `memory_fraction`.

The CUDA-IPC relay preallocation from this branch did not come across and is not lost — it is worker-local policy with no bearing on the base, and belongs with the weight-sharing work.

Issue #1597, opened 2026-08-18
[Roadmap][NPU] Ascend NPU support Roadmap


Makcum888e:
teammates: @Makcum888e @wangyao-i @cocoa-liu @Suxing224 @celestial-micha @chaijiacheng888 @chx96642264

## Models

- [x] **Qwen3-Omni**
  - [x] Thinker NPUGraph (#1476 )
  - [ ] Talker NPUGraph (#1737 )
  - [ ] Code2wav NPUGraph (#1710 )
- [x] **CosyVoice3** @Makcum888e (#1652) (#1694 )
- [x] Qwen-ASR
  - [ ] NPUGraph @wangyao-i
- [ ] Qwen-TTS @Suxing224 
- [x] Higgs-Audio (#1721 )
- [ ] Ming-Omni (#1692 )
- [x] Fish-Audio (#1691 )
- [x] voxtral tts (#1725 )
---

## Features & Core Capabilities
> [03/09 addressed]~~Note: CI tasks are listed at the bottom due to temporary resource limitations. We are actively working on a resolution.~~

- [x] Profiler (#2105 )

- CI & Quality Assurance
  - [ ] Continuous Integration (CI) Pipeline Setup
  - [ ] Model Coverage & Regression Tests
  - [ ] Automated Installation Scripts
  - [ ] Dockerfile & Containerization

- [x] Hardware Abstraction Layer (HAL) (#1306 )
- [x] Initial NPU Execution Support (#1306 )
- Documentation
  - [x] Quickstart / Getting Started (#1577 )
  - [ ] Basic Usage & API Reference
  - [ ] Cookbook & Examples
- [ ] Full Argument & Parameter Compatibility

  <details>
  <summary> NPU Argument Support Matrix (Click to expand)</summary>

  | Args                           | Default value | Options                                           | UTs | supported |
  | ------------------------------ | ------------- | ------------------------------------------------- | --- | --------- |
  | model_path                                  | None          | str                                               |     | [√]        |
  | config                                           | None          | str                                               |     | [√]        |
  | text_only                                       | FALSE         | bool                                              |     | [√]        |
  | colocate                                        | FALSE         | bool                                              |     | [x]       |
  | isolate_stage                                 | None          | list[str]                                         |     |           |
  | stage_process                               | None          | list[str]                                         |     |           |
  | host                                              | 0.0.0.0         | str                                               |     | [√]        |
  | port                                              | 8000           | int                                               |     | [√]        |
[...]

PR #1577, opened 2026-08-17
[NPU] Add Ascend installation manifest and helper

wangyao-i:
## Motivation

The default `pyproject.toml` includes CUDA-specific dependencies that can conflict with an existing Ascend `torch_npu` environment. This PR adds a separate, maintainable installation path for Ascend NPU without changing the externally managed accelerator software stack.

Part of #1597.

## Modifications

- Add `pyproject_npu.toml` with an Ascend-compatible dependency set and optional extras while leaving `torch`, `torch_npu`, `triton-ascend`, `sgl-kernel-npu`, and SGLang under user/upstream version management.
- Align the documented NPU installation baseline with the SGLang `0.5.18` release line.
- Keep the supported SGLang release in one configuration value so future upgrades do not require maintaining a duplicated upper-bound version.
- Add executable `scripts/npu/install_npu.sh`, organized into focused argument parsing, precheck, install, cleanup, verification, and `main` functions.
- Add prechecks for:
  - installed SGLang belonging to the supported `0.5.18` release line, including development, pre-release, final, post-release, and local builds such as `0.5.18.dev7+g<git-sha>`
  - `torch`/`torch_npu` major-minor compatibility
  - `triton-ascend` and `sgl-kernel-npu` imports
  - NPU availability/count and a minimal MatMul
  - `--skip-device-check` for build hosts without exposed devices
- Preserve the original CUDA manifest with locking, stale-backup detection, and restore-on-exit behavior.
- Add `docs/get_started/installation_npu.md`, distinguishing required/manual prerequisites, linking to authoritative compatibility documentation, and providing one recommended installation path.
- Expose the NPU guide through the Get Started toctree and general installation page.
- Add install-script tests for supported SGLang release-line variants, mismatch and missing cases, interrupted recovery, locking, dry-run behavior, extras, build-isolation defaults, and executable paths containing spaces.
- Use the unique test filename `test_npu_install_script.py` to avoid pytest import collisions with the existing XPU installer tests.

## Validation

Completed locally:

- Relevant pre-commit hooks pass, including the executable-shebang check.
- `bash -n scripts/npu/install_npu.sh`
- `pyproject_npu.toml` parses successfully.
- `python -m pytest tests/unit_test/npu/test_npu_install_script.py -q -k "not second_run_refuses_while_lock_is_held"`: 21 passed, 1 deselected.
- Joint collection of the NPU and XPU installer suites: 24 tests collected without module-name conflicts.
[...]

PR #1686, opened 2026-08-24
[NPU] make CosyVoice3 stages platform-aware

TKevin0v0:
# [NPU] make CosyVoice3 stages platform-aware

## Motivation

Fun-CosyVoice3's AR stage defaulted to `device="cuda:0"`. On an Ascend host,
`torch_npu.contrib.transfer_to_npu` made tensor operations execute on NPU, but
SGLang still received `ServerArgs.device="cuda"`.

This split device identity caused SGLang to select the Triton attention path
and generic CUDA graph backend instead of its native Ascend attention and
`NPUCudaGraphBackend` implementations. Decode graph startup then failed because
the generic path passed the CUDA-only `cuda_graph=` keyword to
`torch_npu.npu.graph`.

## Modifications

- Make the CosyVoice3 AR factory device default platform-neutral (`None`).
- Declare `device: null` in the pipeline stage configuration.
- Let the existing SGLang-Omni placement layer resolve the current platform and
  GPU index.
- Extend the stage-device contract tests so a future CUDA literal cannot be
  reintroduced silently.

CUDA behavior remains `cuda:0`; Ascend resolves to `npu:0`.

## Validation

After the change, startup logs show:

```text
attention_backend=ascend
decode_attention_backend=ascend
Capture target decode NPU graph
```

Decode graph capture decreases from `22.44 s` on the incorrectly routed generic
path to `1.17 s` on the native NPU backend.

Fixed request, batch=1, one warmup + five measured calls:

| Mode | Mean latency | Mean RTF |
|---|---:|---:|
| NPU eager | 0.6791 s | 1.3059 |
| NPU graph | 0.4465 s | 0.8586 |

Mean speedup: `1.521x`. The five corresponding short-request WAV outputs are
byte-identical between eager and graph.

The performance run above was collected on SGLang-Omni commit `5a1c694`. The
branch was then replayed without conflicts onto current `main` `fee6a1bc`.
Current-tip validation includes the 52-test focused suite and an Ascend
end-to-end stacked smoke run showing native Ascend attention, native decode NPU
graph capture, HTTP 200, and valid 24 kHz WAV output.

- Current-tip Fun-CosyVoice3 + device-contract focused tests: `52 passed`.
- Format and `git diff --check`: passed.

## Related issues

Related #1597
Related #1652

This PR does not duplicate the NPU dependency manifest/install documentation in
Draft PR #1577.

### DIFF
--- a/sglang_omni/models/fun_cosyvoice3/config.py
+++ b/sglang_omni/models/fun_cosyvoice3/config.py
@@ -43,7 +43,7 @@ class FunCosyVoice3PipelineConfig(PipelineConfig):
             name="tts_engine",
             process="pipeline",
             factory=f"{_PKG}.stages.create_sglang_tts_engine_executor",
-            factory_args={"dtype": "bfloat16"},
+            factory_args={"device": None, "dtype": "bfloat16"},
             gpu=0,
             next="vocoder",
         ),
--- a/sglang_omni/models/fun_cosyvoice3/stages.py
+++ b/sglang_omni/models/fun_cosyvoice3/stages.py
@@ -71,7 +71,7 @@ def create_preprocessing_executor(model_path: str) -> SimpleScheduler:
 def create_sglang_tts_engine_executor(
     model_path: str,
     *,
-    device: str = "cuda:0",
+    device: str | None = None,
     gpu_id: int | None = None,
     dtype: str = "bfloat16",
     server_args_overrides: dict[str, Any] | None = None,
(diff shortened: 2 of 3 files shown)


Makcum888e:
Can you provide benchmark result?
```
python -m benchmarks.eval.benchmark_tts_seedtts --meta zhaochenyang20/seed-tts-eval-arrow --port 8000 --output-dir results/dir --lang en --max-concurrency 16 --generate-only --use-existing-server
```

PR #1690, opened 2026-08-24
feat(pd): resolve the Decode target per request, the seam for non-1:1 sizing

jzheng17:
Based on `pd_runtime`. This is the seam described in #1599 and left out of #1684.

## The problem

The Decode stage that receives a request's KV is chosen when the Prefill stage binds. `bind_pd_runtime` stores it as one string, and the handoff site reads it straight into `to_stage` and `target_pool_id`. A second Decode worker therefore has to change the schema and the send path together.

## The change

The Prefill half resolves the Decode target per request, at the handoff site, through `_resolve_decode_stage(req)`.

The candidate list is derived at bind time from `rank_endpoints`, which `mp_runner` keys by stage name after replica expansion, so it carries `thinker_decode@r0`, `thinker_decode@r1` and so on. `pd_rewrite` still records the logical name, and that is what a build without stage replicas resolves to, so behaviour there is unchanged. The list is sorted rather than left in dict order: KV is rank-addressed, and two Prefill ranks disagreeing on the order would split one request's pages across different Decode halves with nothing in the transfer path to catch it.

Which candidate a request goes to is the coordinator's admission binding when there is one. A replicated Process already gets one instance per request, chosen once at admission and carried on the message envelope, so reading it keeps one request to one answer rather than letting two policies decide the same question. With no binding — no stage replicas, or it has not arrived — a hash of the request id decides. Both inputs are identical on every Prefill rank, which is the constraint `select_decode_stage` documents and the reason that function takes values rather than a scheduler.

## Measured

One Prefill, two Decode, one H200 each, CUDA graph on, `max_running_requests=4`, 32 output tokens, closed-loop client, 100 requests per cell.

| cell | 1:1 | 1:2 | change | 1:1 TPOT p50 | 1:2 TPOT p50 |
| 1024 / 32, c8 | 13.925 | 20.375 | +46.3% | 0.0120 | 0.0075 |
| 1024 / 32, c16 | 15.328 | 25.431 | +65.9% | 0.0259 | 0.0089 |
| 8128 / 32, c8 | 4.835 | 4.819 | −0.3% | 0.0072 | 0.0070 |
| 8128 / 32, c16 | 4.764 | 4.693 | −1.5% | 0.0072 | 0.0072 |

At 8128 tokens the one Prefill card runs at 100% for the whole cell, so a second Decode half has nothing to add. At 1024 tokens Decode is the constraint and adding one nearly triples TPOT at c16. Both Decode cards carry work; utilisation sampled during the run was 100% / 87-100% / 21-77% across Prefill and the two Decode halves.

The whole configuration is:

```yaml
processes:
  thinker_decode:
    num_replicas: 2
    replica_devices: [1, 2]
```

Stage replicas live on `main` and not on `pd_runtime`, so 1:N needs this branch on a base that carries them. On `pd_runtime` alone there is nothing to expand and the single-partner path is what runs.

## Testing

| Run | Result |
| --- | --- |
| `test_pd_decode_target.py` + `test_pd_decode_replicas.py` | 21 pass |
| `tests/unit_test` | 4948 pass, 7 fail |

The 7 failures are `moss_tts/test_audio_tokenizer.py` (2), `qwen3_omni/test_vision_compat.py` (1) and `qwen3_tts/test_compat.py` (4). All 7 fail the same way on `origin/main` in this environment and none touch PD.

The new cases cover: the rewrite records candidates on the Prefill half and none on the Decode half; a `PDExecution` built without the field stays valid; the derived list picks up every replica and ignores a name like `thinker_decoder_aux` that merely shares a prefix; endpoint dicts in opposite order give the same list; the admission binding decides when it exists; and a binding that is missing, outside the candidate list, or raises falls back to the hash rather than sending KV somewhere that cannot receive it.

### DIFF
--- a/sglang_omni/config/pd_rewrite.py
+++ b/sglang_omni/config/pd_rewrite.py
@@ -156,7 +156,11 @@ def _split_pd_stage(
             "stream_to": [inbound_rename.get(t, t) for t in s.stream_to],
             "pd_disaggregation": None,
             # Note (Yue Yin): Keep compiler metadata out of user factory kwargs.
-            "pd_execution": PDExecution(role="prefill", partner=decode_name),
+            "pd_execution": PDExecution(
+                role="prefill",
+                partner=decode_name,
+                decode_targets=(decode_name,),
+            ),
         },
     )
 
--- a/sglang_omni/config/schema.py
+++ b/sglang_omni/config/schema.py
@@ -159,6 +159,11 @@ class PDExecution(BaseModel):
 
     role: Literal["prefill", "decode"]
     partner: str
+    # Note (Audrey Zheng): the Decode halves this Prefill may hand off to. One
+    # entry today. Carrying a sequence means a second Decode worker does not
+    # migrate this schema, and the choice can move to admission time without
+    # touching the handoff send path.
+    decode_targets: tuple[str, ...] = ()
 
 
 class StageConfig(BaseModel):
--- a/sglang_omni/pipeline/stage/runtime.py
+++ b/sglang_omni/pipeline/stage/runtime.py
@@ -156,6 +156,7 @@ class Stage:
                 stage_name=name,
                 role=pd_execution.role,
                 partner=pd_execution.partner,
+                decode_targets=pd_execution.decode_targets,
             )
             self._comm.register_kv_pool(pool)
             if receiver is not None:
--- a/sglang_omni/pipeline/stage/runtime.py
+++ b/sglang_omni/pipeline/stage/runtime.py
@@ -63,6 +63,33 @@ def _error_text(exc: BaseException) -> str:
     return str(exc) or type(exc).__name__
 
 
+def _decode_targets(
+    rank_endpoints: dict[str, tuple[str, ...]],
+    partner: str,
+) -> tuple[str, ...]:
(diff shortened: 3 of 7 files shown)


jzheng17:
Superseded, and conflicting with `pd_runtime` besides. The main line is #1773.

The seam landed there via #1784: `PDExecution.decode_targets`, `OmniPrefillScheduler` taking it with the single partner as its default, and `select_decode_stage` taking `targets` and `request_id` and nothing else — that signature carries the constraint that the KV send is rank-addressed, so two Prefill ranks choosing differently scatter one request's pages with nothing to catch it.

Two pieces of this PR did not come across and are the only things still missing for 1:N:

- deriving the target list at bind time from `rank_endpoints`, which is what makes replica-expanded names (`thinker_decode@r0`, `@r1`) visible to the Prefill half. Without it the list only ever holds the logical name.
- preferring the coordinator's admission binding over the hash, so one request gets one answer on both sides rather than two policies deciding the same question.

Both are being redone on #1773. Measurements from this branch stand: 1:2 against 1:1 on three H200s was +46% at c8 and +66% at c16 on 1024-token prompts, and −0.3% at 8128 where the single Prefill card runs at 100% for the whole cell.

PR #1695, opened 2026-08-24
test(pd): pin the three bind_pd_runtime guards

jzheng17:
Based on `pd_runtime`. Tests only, no behaviour change.

## Why

`bind_pd_runtime` refuses three configurations, and those three refusals are the boundary of the current PD runtime:

```python
if self.tp_size != 1:
    raise NotImplementedError("PR3 PD runtime supports tp_size == 1 only")
if self.page_size != 1:
    raise NotImplementedError("PR3 PD runtime supports page_size == 1 only")
if not self.server_args.disable_radix_cache:
    raise NotImplementedError("PR3 PD runtime requires RadixCache disabled")
```

Nothing in the suite asserts them. Lifting one on purpose produces no signal, and removing one by accident produces none either. Since each is scheduled to come off at a different time, the boundary is worth stating in the tests rather than only in the source.

## The tests

Four cases. Three assert that each guard fires on its own input. The fourth binds a supported configuration and asserts it clears all three guards, failing later on `token_to_kv_pool_allocator` instead. Without that case the first three would still pass if a guard rejected everything, so it is what makes them a boundary rather than a wall.

The stand-in carries only the attributes the guards read, so the tests need no engine, no GPU, and no allocator.

## Testing

| Run | Result |
| --- | --- |
| `test_pd_bind_guards.py` (4 cases, new) | pass |
| `tests/unit_test/pipeline` | 531 passed, 1 failed |

The failure is `test_ipc.py::test_mp_runner_startup_failure_includes_child_factory_traceback`, which fails on this host for an unrelated reason: it allows `timeout=10.0` while a child spawn costs 20-35 s here. It is the same flake noted in #1683 and it fails with or without this change.

### DIFF
--- a/sglang_omni/relay/cuda_ipc.py
+++ b/sglang_omni/relay/cuda_ipc.py
@@ -548,6 +548,7 @@ class CudaIpcRelay(Relay):
         credits: int | None = 2,
         slot_size_kb: int = 64,
         pool_size_mb: int | None = None,
+        preallocate_pool: bool = False,
         **kwargs: Any,
     ) -> None:
         if kwargs:
@@ -555,6 +556,15 @@ class CudaIpcRelay(Relay):
                 f"unexpected cuda_ipc relay options: {', '.join(sorted(kwargs))}"
             )
         self.engine_id = engine_id
+        # Note (Audrey Zheng): the pool is allocated on the first payload that
+        # crosses a process boundary, not at startup, so a deployment can pass
+        # startup and still be a gigabyte short when the first multimodal
+        # request arrives. Measured on a PD pair: `mem_fraction_static=0.87`
+        # launched cleanly and then failed a 750 MiB allocation 19 minutes in,
+        # when the image arm began and `mm_aggregate` first relayed a CUDA
+        # tensor. Preallocating moves that failure to launch, where it is
+        # readable. Default off so no existing deployment changes size.
+        self._preallocate_pool = preallocate_pool
         if device == "cpu":
             raise ValueError(
                 "cuda_ipc relay requires a CUDA device; got 'cpu'. Use the shm "
@@ -577,7 +587,6 @@ class CudaIpcRelay(Relay):
             raise ValueError("cuda_ipc pool_size_mb must be positive")
         if self.slot_count <= 0:
             raise ValueError("cuda_ipc pool size must fit at least one slot")
-
         self._pool_tensor: torch.Tensor | None = None
         self._pool_id: str | None = None
         self._pool_storage_handles: dict[str, dict[str, Any]] = {}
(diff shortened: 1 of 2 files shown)


jzheng17:
Added a fifth case: PD does not require decode CUDA graph off, and this pins that it stays that way.

Nothing in `bind_pd_runtime` guards the flag, and `_run_batch_prebuilt` returns an empty `GenerationBatchResult` without running a forward whenever `batch.inner_idle_batch` is None — which it always is here, since the only assignment is in `dp_attn.py` behind `require_mlp_sync`, and sglang-omni rejects DP attention at config time. So admission has no shape for a captured graph to miss and every step after it is `ForwardMode.DECODE`.

**Measured, because the argument on its own was not enough.** Decode CUDA graph off against on, same cards, same `mem_fraction_static`, same client, capture verified from both sides of each arm:

| concurrency cap | off | on |
| --- | --- | --- |
| 64 | 5.9-6.2 rps | 13.2-13.6 rps |
| 4 | 0.50 rps | at least 2.88 rps |

The cap-4 on-arm never saturated, so that row is a lower bound rather than a ratio. Decode-card utilization sits at 10.9-16.5% with graphs off at every offered rate and cannot be driven higher by offering more load, which is a launch-bound decode loop. **A smaller concurrency cap makes the flag cost more, not less**, because a smaller batch leaves less GPU work to hide the launch overhead behind.

The test asserts that a scheduler with graphs enabled still clears all three guards. A new guard on the flag would raise `NotImplementedError` and fail it.

jzheng17:
Update: the cap-4 arm is now saturated on both sides, so the earlier lower bound has a number.

| concurrency cap | graphs off | graphs on | ratio |
| --- | --- | --- | --- |
| 64 | 5.9-6.2 rps | 13.2-13.6 rps | 2.2x |
| 4 | 0.50 rps | 5.24 rps | **10.5x** |

Both ceilings are established rather than inferred: off4 is flat at 0.48-0.50 across three rungs while in-flight climbs 16 to 111 and total time goes 34 s to 224 s, and on4 is flat at 5.09-5.24 across offered 6, 8 and 12 while in-flight climbs 41 to 318. All 20 points admitted 100% with zero errors.

My earlier comment said "at least 2.88 rps" for the on arm at cap 4. That was a genuine lower bound rather than a ceiling — the arm had not saturated on the ladder I used, because the two arms differ by 10x and no single ladder saturates both: rungs that saturate the on arm push the off arm past the client timeout, and rungs that keep the off arm clean leave the on arm idle. The arms were re-run on separate ladders.

**The mechanism explains why a smaller cap costs more.** Decode-card utilization with graphs off is pinned at 10.4-11.3% at every offered rate and does not move when more load is offered; with graphs on it climbs to 97%. ITL p50 is 60 ms against 5.0 ms. The decode loop is launch-bound, and a batch of 4 leaves less GPU work to hide the launch overhead behind than a batch of 64 does.

Capture verified from both directions again: off4 1047 decode batches all `cuda graph: False`, on4 6893 all `True`, and the captured buckets read `bs=[1, 2, 4]` from the log, matching `build_default_cuda_graph_bs(4)`.

Scope: 42-token prompts, `mem_fraction_static 0.87`, two H200s, text only.

jzheng17:
Folded into #1684. Those guards define what PD requires of its server args, and #1684 is the PR that sets two of the three at compile time, so pinning them belongs with it.

PR #1696, opened 2026-08-24
bench(pd): add the PD versus colocated harness and report

jzheng17:
This is the benchmarking half of PR 5 in the RFC, so it stacks on the PD PRs. Based on `pd_runtime`, which is the deepest base available here: PR 4's branch `pd_result` lives in a fork rather than in this repo. The harness imports no PD code, so it applies cleanly on either.

## What it measures

`benchmark_omni_thinker_attribution.py` offers Poisson arrivals at a target rate for a fixed duration without waiting for completions, so an arm that cannot keep up shows it in the achieved rate rather than in a reduced offered rate. It records five segments per request — build, build to queued, queued to scheduled, first forward, decode tail — using the same decomposition as `docs/developer_reference/qwen3_asr_concurrency_profile.md`.

`analyze_thinker_events.py` computes the inter-token gap split by whether the gap overlaps a concurrent prefill. It accepts events from `thinker`, `thinker_prefill` and `thinker_decode`, so colocated and PD runs go through one code path.

`aggregate_pd_vs_colocated.py` joins per-point results across arms.

## The report

`docs/developer_reference/pd_vs_colocated.md` holds the equal-GPU-count comparison on two H200s: A/A band, text, image, interference, segment attribution, and the admission data. Findings, in short:

- At equal GPU count two colocated replicas beat one PD pair on both workloads. The text gap is a lower bound, because colocated never saturated in the rates swept.
- PD removes prefill/decode interference almost entirely. The prefill-conditioned ITL ratio is about 1.0 for PD against 5.8x to 20.5x for colocated, against an A/A band of 2.04%.
- Prefill is about 2.6% of a request's GPU work at 42 prompt tokens and 128 output tokens, so a 1P:1D pair splits hardware 50/50 against a 2.6/97.4 work split. Measured utilization on the PD prefill card is 10.1% against 67.9% on a colocated card.

## Two measurement notes worth carrying into other benchmarks

**Throughput must divide by wall time, not by the nominal duration.** Arms drain after the offering window closes. PD at offered 44 completes 1672 requests over 157 s, which is 10.7 rps; dividing by 90 gives 18.6.

**SLO attainment must count every offered request.** A refused request meets no SLO. At offered 44 PD's TTFT p95 looks about 10x better than colocated's while PD admits 43.3% and colocated admits 100%. Computing attainment over admitted requests rewards an arm for shedding load.

The report states the A/A band first because two of the metrics are not usable at this sample size: p95 TTFT varied 67.6% between identical arms, and goodput varied 23.0%. Achieved rate (0.59%) and the conditioned ratio (2.04%) are.

## Caveats stated in the report

The arms differ in `mem_fraction_static`, 0.87 against 0.72, because PD's prefill card also holds the `mm_aggregate` relay pool and encoder activations. PD text throughput is identical across all three fractions, which controls the text arm and not the image arm. At equal GPU count PD runs one encoder pipeline against colocated's two, which is an uncontrolled confound on the image arm.


jzheng17:
Reopened as #1793 against `main`.

The content here is base-independent — three scripts under `benchmarks/eval` and one report — but the branch was cut from `pd_runtime`, so retargeting it to `main` pulled the whole PD stack into the diff. #1793 is the same four files on `main`.

Issue #1081, opened 2026-07-17
SGLang-Omni Roadmap


Ratish1:
# SGLang-Omni Roadmap

*Contributions and feedback are welcome.* [Join Slack](https://slack.sglang.ai) and go to #sglang-omni-dev or the WeChat Group; please contact WeChat ID *LoveDeathAndLLM* directly and share your full name and oganization/school/company.

### Supporting

- [x] Qwen/Qwen3-TTS-12Hz-1.7B-Base 
- [x] Qwen/Qwen3-TTS-12Hz-1.7B-CustomVoice
- [x] OpenMOSS-Team/MOSS-TTS 
- [x] openai/whisper-large-v3 
- [ ] nvidia/nemotron-3.5-asr-streaming-0.6b
- [x] OpenMOSS-Team/MOSS-Transcribe-Diarize 
- [x] fishaudio/s2-pro 
- [ ] nvidia/Cosmos3-Super
- [x] Qwen/Qwen3-Omni-30B-A3B-Instruct
- [x] FunAudioLLM/Fun-CosyVoice3-0.5B 
- [ ] nvidia/NVIDIA-NemotronLabs-VoiceChat-11B

### Performance

1. **Dominating voice model performance and utility: Qwen3 TTS (1.7B), Qwen3 ASR (1.7B), Nemotron ASR, Whisper**  

- [ ] Qwen/Qwen3-TTS-12Hz-1.7B-Base ([#1754](https://github.com/sgl-project/sglang-omni/issues/1754))
- [ ] OpenMOSS-Team/MOSS-TTS 
- [ ] Qwen/Qwen3-TTS-12Hz-1.7B-CustomVoice ([#1754](https://github.com/sgl-project/sglang-omni/issues/1754))
- [ ] OpenMOSS-Team/MOSS-TTS-Nano-100M
- [ ] openai/whisper-large-v3 
- [ ] nvidia/nemotron-3.5-asr-streaming-0.6b
- [ ] OpenMOSS-Team/MOSS-Transcribe-Diarize 
- [ ] fishaudio/s2-pro 
- [ ] FunAudioLLM/Fun-CosyVoice3-0.5B 

2. **Frontier multi-modal model inference and utility: Qwen3 Omni, Cosmos**

- [ ] nvidia/Cosmos3-Super
- [ ] Qwen/Qwen3-Omni-30B-A3B-Instruct 

3. **Experimental serving strategies, MPS**

[...]

PR #1698, opened 2026-08-24
feat(qwen3-tts): support 1.7B CustomVoice

YzXiao101:
## Motivation

Related issue: #1081.

## Design

This PR adds the `Qwen/Qwen3-TTS-12Hz-1.7B-CustomVoice` serving entry and unifies request validation and voice discovery for the already-supported 0.6B checkpoint. Both reuse the existing Qwen3-TTS pipeline; model execution and Base/VoiceDesign request semantics are unchanged.

References: released [Qwen3-TTS implementation][qwen-repo], [CustomVoice API][qwen-customvoice-api], and checkpoint configurations ([0.6B][qwen-06-config], [1.7B][qwen-17-config]).

| Area | Behavior |
| --- | --- |
| Speakers | Checkpoint-provided names, case-insensitive; both released sizes have nine speakers. Omitted/default voice remains Vivian. |
| Conditioning | Text, speaker, language, optional instruction; explicit task must be `CustomVoice`. Cloning inputs are rejected. |
| Instructions | Supported on 1.7B. Existing 0.6B passthrough is preserved, but instruction control is not recommended. |
| Language | Explicit language wins. Only `Auto` selects Eric's Sichuan or Dylan's Beijing dialect token; explicit `Chinese` keeps the Chinese token. |
| Output | Buffered speech, batch, incremental HTTP PCM and WebSocket output through the existing codec/vocoder path. |

The instruction and explicit-language choices preserve SGLang-Omni compatibility: the released Qwen wrapper instead clears 0.6B instructions and applies Eric/Dylan dialect tokens to both `Auto` and `Chinese`.

## Code Design & Compatibility

Deployment config optionally resolves one immutable `CustomVoiceConfig(speakers, task_type)`, shared by speech validation and voice discovery. Both fields are required; only `custom_voice_config=None` disables this contract. When present, it enables built-in speaker validation and disables reference inputs and uploaded-voice lookup, without changing the existing uploaded-voice interfaces.

### Ownership and tradeoffs

| Owner | Responsibility |
| --- | --- |
| Qwen config | Resolve the effective logical `tts_engine` checkpoint using existing override precedence; read `config.json` and derive speakers from `talker_config.spk_id`. |
| Launcher / app | Preserve existing uploaded-voice hooks and inject the optional CustomVoice config once at startup; no Qwen-specific branches, new CLI knobs, or worker RPC. |
| Speech validator / voices | Validate HTTP, batch, and WS before reference I/O; expose speakers from the same config in voice discovery. |
| Model / runtime | Retain checkpoint defaults, prompts, sampling, scheduling, streaming, and cancellation. CustomVoice validation does not fill checkpoint defaults. |

Base and VoiceDesign retain their existing paths; CustomVoice supplies the additional config. Existing uploaded-voice arguments and defaults remain unchanged when this config is absent, with no compatibility conversion layer. When present, checkpoint speakers take precedence over uploaded voices, including when the engine's model path overrides a top-level Base path. Direct `create_app` callers opt in explicitly; a model name alone does not activate the config.

### Compatibility boundaries

Valid existing 0.6B generation behavior is preserved. API tightening is intentional: invalid speakers, incompatible tasks, and cloning inputs now return early errors; voice discovery lists checkpoint speakers. This is not a blanket “no breaking changes” claim.

| Existing workflow | Effect of this design |
| --- | --- |
| Config / startup | `resolve`/`explain` and serialization do not invoke the CustomVoice hook. All Qwen variants resolve checkpoint metadata at serve startup; Base/VoiceDesign return no CustomVoice config. The launcher needs local or Hub/cache access; resolution failures use existing runner cleanup. |
| Default / split / replicas / TP | CustomVoice config belongs to the logical deployment; placement, IPC, and execution are unchanged. TP/replica coverage is configuration-level, not multi-GPU model qualification. |
| Uploaded voices / router | Existing optional/required lookup is preserved. Built-in speakers do not enter the router's uploaded-voice registry; each app owns its config. |
| Other APIs / controls | Chat, ASR, realtime, direct generation, and control forwarding are unchanged. CustomVoice config is startup-only; changing speaker/type metadata requires a restart. |

### Adding future models

Models inherit a no-op CustomVoice hook. A checkpoint with built-in speakers and no reference conditioning can supply the config; reference-based models keep their existing interfaces. A future model supporting both built-in speakers and references needs a separate contract defining name collisions and reference handling.

### Functional Validation

| Coverage | Result |
| --- | --- |
| Earlier full-suite UT: config, serve, pipeline, router, Client, Qwen3-TTS, model capabilities | 2249 passed; 2 not-applicable/environment skips |
| Latest config / speech API refactor UT | 248 passed; 1 existing subprocess-timeout test deselected |
| API compatibility across 34 model configs/variants and 5 direct-call configurations | 858/858 checks match the previous PR; all 814 non-CustomVoice checks match baseline main (`7989a5ed`) |
| 0.6B / 1.7B CustomVoice, default layout | 32/32 checks each |
| 0.6B / 1.7B CustomVoice, split preprocessing/vocoder | 19/19 checks each |
| 1.7B CustomVoice, worker behind a real router | 19/19 checks |

Smoke covers all speakers, defaults, instruction acceptance, Chinese/Auto, HTTP/WS streaming, batch errors, and disconnect recovery. All 71 waveforms passed integrity checks; queues drained after disconnect. Model smoke predates the latest API-only config refactor, which was verified by focused UT and the compatibility matrix above. This verifies functionality, not perceptual dialect identity or 0.6B instruction quality.

### Setup

SGLang results below are from the latest complete run on `2cb33f7c` (2026-09-06).

| Item | Value |
| --- | --- |
| Model | `Qwen/Qwen3-TTS-12Hz-1.7B-CustomVoice@0c0e3051` |
| Hardware | 1x H200 141 GB, BF16, TP1 |
| Environment | No external process on the target GPU during timed windows; host CPU, memory, and I/O shared with another profiling task |
| Workload | Full Seed-TTS-Eval EN 1088 and ZH 2020, one run per language/mode |
| Request | Ryan/English for EN, Vivian/Chinese for ZH, no reference or instruction |
| Generation | `max_new_tokens=2048`; checkpoint sampling defaults; no SGLang seed override |
| Concurrency | 16, with 16 warmup requests per language/mode |
| Server | Running/decode CUDA-graph batch limits 64; split vocoder; engine/vocoder memory fractions 0.85/0.10; default `1 -> 2 -> 4` chunk ramp; torch.compile disabled |
| Evaluators | Qwen3-ASR-1.7B at concurrency 32, corpus-micro WER/CER; mean UTMOS |

### Accuracy

All 6,216 outputs were scored. Primary WER/CER includes every sample; filtered WER/CER excludes samples above 50% and recomputes the corpus rate, for both engines.

| Engine | Mode | EN WER all / filtered / >50% count | ZH CER all / filtered / >50% count | UTMOS EN / ZH |
| Official `qwen-tts` runtime (our run) | Non-streaming | 1.516% / 1.375% / 1 | 0.878% / 0.878% / 0 | 4.2253 / 3.3213 |
| SGLang-Omni | Non-streaming | 1.608% / 1.359% / 3 | 0.984% / 0.984% / 0 | 4.1723 / 3.1824 |
| SGLang-Omni | Streaming | 2.085% / 1.454% / 4 | 0.927% / 0.927% / 0 | 4.1500 / 3.1789 |

Reference: our earlier non-streaming full-corpus run with `qwen-tts==0.1.1`, the same checkpoint revision, manifests, Ryan/Vivian conditioning, and evaluators; BF16, `seed=42`, `max_new_tokens=2048`, default sampling. SGLang leaves seed unset. These are our official-runtime measurements, not the Qwen [technical report][qwen-report]'s different-corpus/scorer results.

Even equal seeds cannot guarantee identical samples across different sampling paths; the official row is aggregate quality context, not a paired comparison.

### Performance

| Mode | Lang | QPS | Latency mean (s) | RTF mean | TTFA mean (s) |
| Non-streaming | EN | 14.788 | 1.075 | 0.2335 | N/A |
| Non-streaming | ZH | 13.307 | 1.198 | 0.2079 | N/A |
| Streaming | EN | 10.098 | 1.573 | 0.3380 | 0.1213 |
| Streaming | ZH | 8.333 | 1.915 | 0.3332 | 0.1047 |

TTFA measures first PCM arrival, not audible speech onset; the first payload held 80 ms of audio. Across all 3,108 streams, 98.99% EN / 93.76% ZH had no underrun longer than 50 ms; maximum underrun was 396.1 / 2072.0 ms. The default ramp does not guarantee uninterrupted playback at concurrency 16.

[qwen-repo]: https://github.com/QwenLM/Qwen3-TTS/tree/022e286b98fbec7e1e916cb940cdf532cd9f488e
[qwen-customvoice-api]: https://github.com/QwenLM/Qwen3-TTS/blob/022e286b98fbec7e1e916cb940cdf532cd9f488e/qwen_tts/inference/qwen3_tts_model.py#L732-L839
[qwen-report]: https://arxiv.org/html/2601.15621
[qwen-06-config]: https://huggingface.co/Qwen/Qwen3-TTS-12Hz-0.6B-CustomVoice/blob/85e237c12c027371202489a0ec509ded67b5e4b5/config.json
[qwen-17-config]: https://huggingface.co/Qwen/Qwen3-TTS-12Hz-1.7B-CustomVoice/blob/0c0e3051f131929182e2c023b9537f8b1c68adfe/config.json

### DIFF
--- a/sglang_omni/models/qwen3_tts/config.py
+++ b/sglang_omni/models/qwen3_tts/config.py
@@ -1,16 +1,23 @@
 # SPDX-License-Identifier: Apache-2.0
-"""Pipeline configuration for Qwen3-TTS Base."""
+"""Pipeline configuration for Qwen3-TTS."""
 
 from __future__ import annotations
 
+import json
 import re
+from pathlib import Path
 from typing import Any, ClassVar
 
 from sglang_omni.config import (
+    CustomVoiceConfig,
     EngineStageConfig,
     FactoryArgs,
     PipelineConfig,
     StageConfig,
+)
+from sglang_omni.config.runtime import (
+    resolve_stage_factory_kwargs,
+    resolve_stage_typed_kwargs,
 )
 
 _PKG = "sglang_omni.models.qwen3_tts"
@@ -23,7 +30,7 @@ _QWEN3_TTS_CUSTOM_VARIANT_MARKERS = (
 
 
 class Qwen3TTSPipelineConfig(PipelineConfig):
-    """3-stage Qwen3-TTS Base pipeline: preprocessing -> engine -> vocoder."""
+    """3-stage Qwen3-TTS pipeline: preprocessing -> engine -> vocoder."""
 
     architecture: ClassVar[str] = "Qwen3TTSForConditionalGeneration"
     requires_model_capabilities: ClassVar[bool] = True
@@ -108,6 +115,60 @@ class Qwen3TTSPipelineConfig(PipelineConfig):
     def supports_uploaded_voice_references(self) -> bool:
         return is_qwen3_tts_base_model(self.model_path)
 
+    def resolve_custom_voice_config(self) -> CustomVoiceConfig | None:
+        engine_stage = self.stage_named("tts_engine")
+        # Note(yzxiao): Inspect the checkpoint the engine factory will actually
+        # receive, including pipeline-authored and user-set factory overrides.
+        engine_factory_kwargs = {
+            "model_path": self.model_path,
+            **resolve_stage_factory_kwargs(engine_stage, self),
+            **resolve_stage_typed_kwargs(engine_stage),
+        }
+        checkpoint_config = _load_qwen3_tts_checkpoint_config(
+            engine_factory_kwargs["model_path"]
+        )
+        model_type = _normalize_qwen3_tts_model_type(
+            checkpoint_config.get("tts_model_type")
+        )
+        if model_type in {"base", "voice_design"}:
+            return None
+        if model_type == "custom_voice":
+            spk_id = checkpoint_config.get("talker_config", {}).get("spk_id")
+            if (
+                not isinstance(spk_id, dict)
+                or not spk_id
+                or any(not isinstance(name, str) or not name.strip() for name in spk_id)
+            ):
+                raise ValueError(
+                    "CustomVoice requires a non-empty talker_config.spk_id speaker mapping"
+                )
+            return CustomVoiceConfig(
+                speakers=tuple(spk_id),
+                task_type="CustomVoice",
+            )
+        return None
+
+
+def _load_qwen3_tts_checkpoint_config(model_path: str) -> dict[str, Any]:
+    checkpoint_dir = Path(model_path).expanduser()
+    if checkpoint_dir.is_dir():
+        config_path = checkpoint_dir / "config.json"
+    else:
+        from huggingface_hub import hf_hub_download
+
+        config_path = Path(hf_hub_download(repo_id=model_path, filename="config.json"))
+    with config_path.open(encoding="utf-8") as handle:
+        return json.load(handle)
+
+
+def _normalize_qwen3_tts_model_type(raw: Any) -> str:
+    normalized = str(raw or "base").replace("-", "_").strip().lower()
+    if normalized == "customvoice":
+        return "custom_voice"
+    if normalized == "voicedesign":
+        return "voice_design"
+    return normalized
+
 
 def qwen3_tts_checkpoint_model_type(checkpoint_dir: str) -> str:
     """Read ``tts_model_type`` from a resolved checkpoint.
@@ -117,20 +178,10 @@ def qwen3_tts_checkpoint_model_type(checkpoint_dir: str) -> str:
     config does, and it is the same value the request path validates against.
     Returns ``"base"`` when the field is absent, matching that path's default.
     """
-    import json
-    import os
-
-    config_path = os.path.join(checkpoint_dir, "config.json")
-    if not os.path.isfile(config_path):
+    if not (Path(checkpoint_dir) / "config.json").is_file():
         return "base"
-    with open(config_path, encoding="utf-8") as handle:
-        raw = json.load(handle).get("tts_model_type")
-    normalized = str(raw or "base").replace("-", "_").strip().lower()
-    if normalized == "customvoice":
-        return "custom_voice"
-    if normalized == "voicedesign":
-        return "voice_design"
-    return normalized
+    config = _load_qwen3_tts_checkpoint_config(checkpoint_dir)
+    return _normalize_qwen3_tts_model_type(config.get("tts_model_type"))
 
 
 def is_qwen3_tts_base_model(model_path: str) -> bool:
--- a/sglang_omni/serve/openai_api.py
+++ b/sglang_omni/serve/openai_api.py
@@ -57,7 +57,7 @@ from sglang_omni.client.audio import (
     encode_pcm,
     select_audio_delta,
 )
-from sglang_omni.config import ResolvedAudioChunking
+from sglang_omni.config import CustomVoiceConfig, ResolvedAudioChunking
 from sglang_omni.config.schema import MAX_SPEECH_INPUT_CHARS
 from sglang_omni.http.admin_auth import (
     make_admin_auth_dependency,
@@ -178,6 +178,7 @@ def create_app(
     model_name: str | None = None,
     requires_uploaded_voice_for_named_voice: bool = False,
     supports_uploaded_voice_references: bool = True,
+    custom_voice_config: CustomVoiceConfig | None = None,
     supports_audio_translation: bool = False,
     required_speech_reference_count: int | None = None,
     speech_reference_text_required: bool = False,
@@ -202,6 +203,8 @@ def create_app(
             names must resolve to uploaded voices before reaching the model.
         supports_uploaded_voice_references: Whether uploaded voice names can be
             lowered into backend reference-audio requests.
+        custom_voice_config: Checkpoint speaker names and task type for CustomVoice.
+            When present, reference inputs and uploaded-voice resolution are disabled.
         supports_audio_translation: Whether the configured pipeline supports
             ``/v1/audio/translations``.
         required_speech_reference_count: Exact reference count required before
@@ -254,6 +257,7 @@ def create_app(
     app.state.speaker_sample_store = SpeakerSampleStore()
     app.state.speech_service = SpeechRequestValidator(
         default_model=app.state.model_name,
+        custom_voice_config=custom_voice_config,
         requires_uploaded_voice_for_named_voice=(
             requires_uploaded_voice_for_named_voice
         ),
@@ -300,7 +304,13 @@ def _register_voices(app: FastAPI) -> None:
             return JSONResponse(
                 content={"uploaded_voice_names": voice_store.uploaded_voice_names()}
             )
-        response = VoiceListResponse.model_validate(voice_store.list_response())
+        voice_list = voice_store.list_response()
+        custom_voice_config = app.state.speech_service.custom_voice_config
+        if custom_voice_config is not None:
+            voices = {name.casefold(): name for name in custom_voice_config.speakers}
+            voices["default"] = "default"
+            voice_list["voices"] = sorted(voices.values(), key=str.casefold)
+        response = VoiceListResponse.model_validate(voice_list)
         return JSONResponse(content=response.model_dump(exclude_none=True))
 
     @app.post("/v1/audio/voices")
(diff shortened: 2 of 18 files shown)


YzXiao101:
## Update: Initial Qwen3-TTS 1.7B CustomVoice Full-Corpus Validation

Update:

- Validated `Qwen/Qwen3-TTS-12Hz-1.7B-CustomVoice` on the complete SeedTTS EN and ZH splits in both streaming and non-streaming modes.
- Re-scored every generated WAV with the repository Qwen3-ASR path, the official SeedTTS Whisper/Paraformer evaluator, and UTMOS.

Benchmark:

Commit: `1a0f3241`. Setup: 1x H200, TP1, canonical 1.7B CustomVoice config, concurrency 16, `warmup=1`, full SeedTTS EN 1088 and ZH 2020, Ryan/English and Vivian/Chinese, no reference or instruction, default sampling, Qwen3-ASR C32.

Performance:

| delivery | lang | requests | failed | QPS | latency mean (s) | TTFA mean (s) | RTF mean |
| non-stream | EN | 1088 | 0 | 11.932 | 1.330 | N/A | 0.2982 |
| non-stream | ZH | 2020 | 0 | 11.574 | 1.378 | N/A | 0.2445 |
| stream | EN | 1088 | 0 | 8.260 | 1.922 | 0.397 | 0.4194 |
| stream | ZH | 2020 | 0 | 7.544 | 2.117 | 0.332 | 0.3728 |

Quality:

| delivery | lang | Qwen3-ASR WER/CER | official SeedTTS evaluator WER/CER | UTMOS mean |
| non-stream | EN | 1.825% | 2.561% | 4.1754 |
| non-stream | ZH | 1.228% | 1.634% | 3.2048 |
| stream | EN | 1.583% | 2.480% | 4.1696 |
| stream | ZH | 0.961% | 1.397% | 3.1857 |

Official references (not directly comparable): Qwen 1.7B Base SeedTTS WER is `1.24%` for EN and `0.77%` for ZH; tokenizer-reconstruction UTMOS is `4.16`; 1.7B C1 first-packet latency and RTF are `101 ms` and `0.313`.

Result:

- All 6,216 generation requests completed successfully; all outputs were valid 24 kHz mono 16-bit audio, and both ASR evaluators plus UTMOS covered every sample.
- Existing validation remains green: `540 passed, 48 skipped`, plus real-checkpoint startup, WAV, HTTP/WS streaming, batch isolation, abort cleanup, and same-seed codec-exact E2E coverage.
- Streaming continuity was 99.63% at C50/C100/C200 for EN with 4/1088 measured playback gaps, and 100% for ZH with 0/2020 gaps.
- EN was the first workload after server startup, and `warmup=1` did not pre-capture all C16 predictor graph sizes. These numbers intentionally retain that first-wave cost and are not steady-state thresholds.
- Official references use different conditioning, tasks, concurrency, packetization, engine, and timing boundaries; they are context only, not CustomVoice thresholds.

JiaxinD:
Note: #1852 makes codec output streaming independent of `non_streaming_mode` for all task types, which generalises the `True if prepared.is_target_custom_voice else ...` special case in this PR. If #1852 lands first that hunk can be dropped and this branch inherits the default.

JiaxinD:
This draft is still useful for the 1.7B CustomVoice checkpoint/config/serve surface, but its streaming hunk overlaps #1852. Please rebase after #1852 and drop the checkpoint-specific stream_codec_output special case so the prompt/output concerns stay generic. Before moving out of draft, complete the checklist with focused unit coverage, a real 1.7B CustomVoice smoke/quality result, and exact-head CI; the current body has no accuracy or benchmark evidence.

YzXiao101:
@JiaxinD hi jiaxin could you help please review this pr? thx

YzXiao101:
> This draft is still useful for the 1.7B CustomVoice checkpoint/config/serve surface, but its streaming hunk overlaps #1852. Please rebase after #1852 and drop the checkpoint-specific stream_codec_output special case so the prompt/output concerns stay generic. Before moving out of draft, complete the checklist with focused unit coverage, a real 1.7B CustomVoice smoke/quality result, and exact-head CI; the current body has no accuracy or benchmark evidence.

Thanks — this branch is now rebased onto main after #1852, and the checkpoint-specific streaming special case has been removed so it inherits the generic stream_codec_output path. Focused unit coverage and real 1.7B CustomVoice smoke/quality results are now included.

sglang_omni/models/qwen3_tts/config.py
```diff
@@ -108,6 +124,14 @@ def requires_uploaded_voice_for_named_voice(self) -> bool:
     def supports_uploaded_voice_references(self) -> bool:
         return is_qwen3_tts_base_model(self.model_path)
 
+    def custom_voice_speech_policy(self) -> CustomVoiceSpeechPolicy | None:
+        if not _is_qwen3_tts_custom_voice_model(self.model_path):
```
JiaxinD: This picks the policy by matching the model path against two literal checkpoint names, so a CustomVoice checkpoint served from /srv/checkpoints/current, or even Qwen/Qwen3-TTS-1.7B-CustomVoice without the 12Hz segment, gets no policy at all. qwen3_tts_checkpoint_model_type at line 136 is the authoritative signal and the engine already keys off it, so the same server validates one way and serves the other.

sglang_omni/models/qwen3_tts/config.py
```diff
@@ -20,10 +21,25 @@
     "voice_design",
     "voicedesign",
 )
+_QWEN3_TTS_CUSTOM_VOICE_MODEL_MARKERS = (
+    "qwen3_tts_12hz_0_6b_customvoice",
+    "qwen3_tts_12hz_1_7b_customvoice",
+)
+QWEN3_TTS_CUSTOM_VOICE_SPEAKERS = (
```
JiaxinD: The checkpoint ships this list as spk_id and the model already matches it case-insensitively at sglang_model.py:668, so this is a third copy that can drift, and the test asserts it against a retyped copy of itself. Deriving voices from the checkpoint would also drop the hardcoded default.

sglang_omni/serve/openai_api.py
```diff
@@ -265,6 +268,7 @@ def create_app(
         ),
         additional_speech_languages=additional_speech_languages,
         max_speech_input_chars=max_speech_input_chars,
+        custom_voice_policy=custom_voice_policy,
```
JiaxinD: The policy reaches the validator here, but GET /v1/audio/voices further down still returns only ["default"], so a client that discovers voices programmatically cannot find the nine names that /v1/audio/speech now hard-rejects everything outside of.

tests/unit_test/qwen3_tts/test_pipeline.py
```diff
@@ -4432,8 +4502,10 @@ def test_qwen3_tts_prepared_payload_missing_state_fails_without_rebuild(
         )
 
 
-def test_qwen3_tts_prepare_custom_voice_uses_speaker_path(
+@pytest.mark.parametrize("model_size", ["0b6", "1b7"])
```
JiaxinD: tts_model_size is assigned once in sglang_model.py and read nowhere, so both parameters run identical code and the rename adds no assertion over base. Deleting the root_config line leaves all 394 tests green, and this is the only test naming the size the PR is titled for.

sglang_omni/models/qwen3_tts/config.py
YzXiao101: Thx. fixed: CustomVoice detection now reads tts_model_type from the effective tts_engine checkpoint with the engine’s normalization, covering local paths and stage-level model overrides.

sglang_omni/models/qwen3_tts/config.py
YzXiao101: Thx, fixed: Removed the duplicated speaker list and API-level Vivian default: speakers now come from talker_config.spk_id, while default-speaker resolution stays in the existing model path.

sglang_omni/serve/openai_api.py
YzXiao101: Thx, fixed: /v1/audio/voices and speech validation now share the same CustomVoiceConfig, exposing checkpoint speakers plus default while preserving other models’ existing behavior.

tests/unit_test/qwen3_tts/test_pipeline.py
YzXiao101: Removed the redundant size-only parametrization; new tests cover checkpoint metadata and serving contracts, while the previously reported real-checkpoint smoke tests provide 1.7B functional coverage.

JiaxinD (approved):

PR #1632, opened 2026-08-20
[AMD][ROCm] Add Docker development image

zijiecode:
## Motivation

SGLang-Omni does not currently provide a ROCm development image. The default project metadata also declares CUDA-specific packages, so installing it directly on a ROCm base can produce an inconsistent environment or CUDA package metadata.

This PR adds a ROCm 7.2 development image for `gfx950`, using a pinned SGLang ROCm base and platform-specific project metadata.

## Modifications

- Add `docker/rocm.Dockerfile` based on a digest-pinned SGLang ROCm image.
- Add `pyproject_rocm.toml` for application and model dependencies while inheriting Torch, SGLang, AITER, Triton, FlashInfer, NIXL, UCX, and the ROCm runtime from the base image.
- Install project dependencies in a separate cacheable layer.
- Clone SGLang-Omni during the image build and check out `OMNI_REF` in detached mode. Runtime startup does not update the source tree.
- Keep the ROCm project metadata at the source root so subsequent editable installs cannot resolve the CUDA dependency set.
- Install SGLang-Omni with `--no-deps` to preserve the GPU-coupled stack from the pinned base image.
- Add small build-time checks for the ROCm PyTorch build, required imports, and UCX ROCm support.

The public build interface is:

```bash
docker build \
  --file docker/rocm.Dockerfile \
  --build-arg OMNI_REF=<exact-commit-sha> \
  --tag sglang-omni:rocm-dev \
  .
```

## Validation

- `pyproject_rocm.toml`, `git diff --check`, pre-commit, and Docker BuildKit `--check` pass.
- A real image build succeeds with an exact `OMNI_REF`; the image source `HEAD` and permanent ROCm metadata overlay match the requested inputs.
- The installed SGLang-Omni metadata contains no CUDA-specific requirements and introduces no SGLang-Omni `pip check` errors.

## Checklist

- [x] Format your code according with pre-commit.
- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed.
- [ ] Provide throughput / latency benchmark results and accuracy evaluation results as needed. 
- [ ] For reviewers: If you haven't made any contributions to this PR and are only assisting with merging the main branch, please remove yourself as a co-author when merging the PR.

PR #1700, opened 2026-08-24
[AMD][ROCm] Support Qwen3-ASR on gfx950

zijiecode:
## Motivation

This PR supports Qwen3-ASR on AMD Instinct MI355X.

Qwen3-ASR exposed three ROCm-specific issues on gfx950:

- The ROCm platform intentionally returns `None` for the packed `fused_qk_norm_rope` operator. Qwen3-ASR still installed its fused wrapper and converted positions to `int32`, while the native RoPE path expects `torch.long` positions.
- `VisionAiterAttention` derived `max_seqlen` with `seq_lens.max().item()` inside the encoder graph capture region. The device-to-host synchronization is unsupported during HIP graph capture and raises `hipErrorStreamCaptureUnsupported`.
- In the ROCm 7.2, the AITER `mha_batch_prefill_func` specialization selected by Qwen3-ASR is affected by a longstanding upstream ROCm compiler issue and produces numerically incorrect results. Among the SGLang-Omni models tested so far, this has only been reproduced for Qwen3-ASR's BF16 causal attention shape (`q_heads=16`, `kv_heads=8`, `head_dim=128`). The AITER decode path is unaffected. This is not a new problem: there is currently no corresponding compiler-side fix, and the ROCm-side test for this case is skipped.

## Modifications

- Use the existing `current_platform.get_fused_qk_norm_rope()` result as the capability check for the Qwen3-ASR text decoder.
  - Install the fused QK-norm/RoPE wrapper only when the platform provides the operator.
  - Return to the native path before QKV projection when the fused operator is unavailable or the input is not BF16.
  - Use `int32` positions for the fused operator and preserve `torch.long` positions for native RoPE.
- Make the Qwen3-ASR encoder layer-stack graph compatible with ROCm/AITER.
  - Precompute `VisionAttentionMetadata` outside HIP graph capture.
  - Keep `cu_seqlens`, `seq_lens`, and the host `max_seqlen` scalar in static graph state.
  - Refresh the mutable sequence-length tensors before graph replay.
- Add a model-scoped backend default for gfx950 + ROCm 7.2:

  ```text
  decode attention backend:  normally selected backend (AITER by default)
  prefill attention backend: Triton
  ```

## Accuracy Test

The tests were run on one AMD Instinct MI355X with BF16 and TP=1. The runtime used a standard Docker image built from #1632:

The following 20-sample SeedTTS EN diagnostic used greedy decoding, concurrency 1, eager execution, and no warmup. It was used to isolate correctness by prefill/decode backend rather than as a capacity benchmark.

| Prefill backend | Decode backend | Corpus WER | Throughput (samples/s) | RTFx |
|---|---|---:|---:|---:|
| AITER | AITER | 728.72% | 0.663 | 3.467 |
| AITER | Triton | 729.76% | 0.575 | 3.008 |
| Triton | AITER | **0.692%** | **2.608** | **13.637** |
| Triton | Triton | **0.692%** | 2.393 | 12.510 |

The launch command:

```bash
sgl-omni serve \
  --port 8000
```

An OpenAI-compatible transcription request for `tests/data/query_to_cars.wav` returned:

```json
{"text":"How many cars are there in the picture?","usage":{"type":"duration","seconds":5}}
```

## Unit Test

Since there is no ROCm CI at the moment, the following tests were run in the standard ROCm container:

The added tests cover:

- `test_graph_matches_eager_tower` compares the encoder graph result with eager execution on the active accelerator backend.
- `test_layer_stack_forwards_precomputed_attention_metadata` confirms that graph capture supplies the precomputed vision-attention metadata.
- `test_forward_keeps_long_positions_for_native_rope` preserves the position dtype required by the native RoPE implementation.
- `test_fused_asr_qk_norm_rope_is_not_bound_without_platform_kernel` confirms that ROCm does not install a wrapper for an unavailable fused operator.
- `test_qwen3_asr_defaults_prefill_to_triton_only_on_rocm_72_gfx95` covers the exact platform, architecture, and ROCm-version gate.
- `test_qwen3_asr_explicit_prefill_backend_overrides_rocm_default` confirms that an explicit deployment choice overrides the model default.

## Checklist

- [x] Add unit tests.

### DIFF
--- a/sglang_omni/models/qwen3_asr/encoder_cuda_graph.py
+++ b/sglang_omni/models/qwen3_asr/encoder_cuda_graph.py
@@ -15,6 +15,9 @@ from dataclasses import dataclass
 from typing import Any
 
 import torch
+from sglang.srt.layers.attention.vision import VisionAttentionMetadata
+
+from sglang_omni.platforms import current_platform
 
 logger = logging.getLogger(__name__)
 
@@ -53,6 +56,7 @@ class _CapturedGraph:
     graph: torch.cuda.CUDAGraph
     hidden_states: torch.Tensor  # [bucket, hidden] static input
     cu_seqlens: torch.Tensor  # [max_windows + 1] static window boundaries
+    attention_metadata: VisionAttentionMetadata | None
     output: torch.Tensor  # [bucket, output_dim] static result
 
 
@@ -88,6 +92,7 @@ class Qwen3ASREncoderLayerStackGraphRunner:
         self._buckets = buckets[:-1] + (top + self._max_windows_for(top),)
         self._graphs: dict[int, _CapturedGraph] = {}  # bucket size -> recorded graph
         self._failed: set[int] = set()
+        self._capture_attention_metadata: VisionAttentionMetadata | None = None
 
     @property
     def tokens_per_window(self) -> int:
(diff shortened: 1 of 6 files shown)


Ratish1 (approved):
LGTM

Issue #1685, opened 2026-08-24
[Bug] KV budget uses the global-free-memory fallback in every containerized run, and can over-size the pool


jzheng17:
## Summary

`get_process_gpu_memory_bytes` returns 0 in a container, so `_OmniKVCacheConfigurator._profile_available_bytes` never uses NVML accounting and always takes the global free-memory fallback. That fallback races model-load timing, and we measured it size a KV pool ~11.9 GiB too large on a plain colocated single-GPU run.

This is on `main` and is unrelated to prefill/decode work; a PD run happens to make it visible because two halves of one stage size their pools differently.

## Where it breaks

`sglang_omni/utils/gpu_memory.py:125` matches `os.getpid()` against pids from `nvmlDeviceGetComputeRunningProcesses`. NVML reports pids from the host namespace; the caller runs in a container namespace and sees a different number for itself, so the join never matches and the function returns 0.

NVML itself is fine. In one instrumented worker it reported host pid 546113 holding 59.97 GiB while torch had 59.16 GiB allocated, so the number is both present and accurate. Only the join fails.

`sglang_omni/model_runner/sglang_model_runner.py:87` treats `<= 0` as "NVML unavailable" and falls through to `_profile_available_bytes_from_stage_load_delta`. Every profile line we captured logged `gpu_mem_accounting=stage_load_fallback` with `pre_load_process_mem=0.00GiB`.

Reproducer, inside the container, on any card:

```python
import os, torch, pynvml
buf = torch.empty(int(1e9), dtype=torch.float16, device="cuda:0")
torch.cuda.synchronize()
pynvml.nvmlInit()
h = pynvml.nvmlDeviceGetHandleByIndex(0)
print(os.getpid(), [p.pid for p in pynvml.nvmlDeviceGetComputeRunningProcesses(h)])
```

The calling pid is absent from the list even though it holds a CUDA context, so the lookup falls through to `return 0`.

## What it costs

`Qwen/Qwen3-Omni-30B-A3B-Instruct`, H200 140.40 GiB, thinker true footprint 57.0-57.6 GiB measured four times.

| Configuration | `stage_load_used` | `available_for_kv` | KV tokens | Card after |
| --- | --- | --- | --- | --- |
| colocated, run A | 45.42 GiB | 76.73 GiB | 838138 | 97.7% |
| colocated, run B | 57.03 GiB | 65.12 GiB | 711261 | 90.9% |

[...]

PR #430, opened 2026-05-12
[Colocation]: Colocated Stage Execution for Omni

Ratish1:
## Motivation

Implements the colocated-stage execution path for Omni v1, following the colocation RFC and the Qwen3-Omni tracking discussion in #329 / #376.

The goal is to make Qwen3-Omni speech runnable as a single colocated v1 server while preserving the v1 architecture boundary:

```text
PipelineConfig / StageConfig
  -> typed runtime config
  -> placement plan
  -> stage process launch
  -> backend adapter
  -> SGLang ModelRunner / KV pool sizing
```

The important design point is that colocation uses Omni placement semantics, not SGLang global-free-memory semantics. Each GPU-resident stage receives a `runtime.resources.total_gpu_memory_fraction` budget. For SGLang AR stages, that stage budget is translated into KV-cache headroom at the SGLang ModelRunner boundary:

```text
available_kv_bytes =
    total_gpu_memory_bytes * total_gpu_memory_fraction
  - accounted_stage_memory_bytes
```

SGLang still owns KV block layout and memory-pool construction. This PR only changes how Omni supplies the available bytes for colocated AR stages so the result is stable when multiple v1 stages share one GPU.

## Modifications

- Added typed v1 runtime config for stage-level execution controls:
  - `runtime.resources.total_gpu_memory_fraction` for placement-owned total-GPU-memory budgets.
  - `runtime.sglang_server_args.mem_fraction_static` for backend-namespaced SGLang AR overrides.
  - `runtime.max_seq_len` and `runtime.video_fps` as typed stage runtime values mapped through `runtime_arg_map`.
- Added a model-agnostic placement planner in `sglang_omni_v1/config/placement.py`.
  - Expands stage GPU placement.
  - Sums per-GPU stage memory budgets.
  - Rejects missing or over-budget colocated GPU groups.
  - Resolves whether the pipeline needs the multi-process runner.
  - Computes same-GPU stream targets before stage processes are built.
- Added a Qwen3-Omni placement policy in `sglang_omni_v1/models/qwen3_omni/placement.py`.
  - Keeps the v1 speech graph as eight stages.
  - Rejects a standalone `code_predictor` stage because the code predictor is part of `talker_ar`.
  - Rejects unsupported thinker/talker TP for this phase.
  - Allows same-GPU thinker/talker only through the Qwen3-Omni colocated config.
  - Requires explicit memory budgets for colocated GPU stages.
- Added `Qwen3OmniSpeechColocatedPipelineConfig`.
[...]

PR #1701, opened 2026-08-24
feat(pd): let the two halves share a GPU

jzheng17:
Based on `pd_runtime`. Removes the check that the two PD halves cannot share a GPU.

## What the check was protecting

The rejection landed in 1912a926 with no stated rationale, and the adjacent message reads "requires explicit prefill.gpu and decode.gpu for real PD", which suggests the intent was definitional rather than technical. There is a real technical hazard behind it, but it is avoidable, and it is not what the message says.

The hazard is that `mem_fraction_static` is not order-safe when two halves size their pools from one card. Both halves serialize on the same startup lock and **the winner varies between runs**. Measured on one H200: at 0.93 on both halves, whichever won took 780,987 KV tokens and the other died with `torch.OutOfMemoryError`. Assigning asymmetric fractions does not help, because you cannot know which half will load first — at 0.45 prefill / 0.90 decode, decode won the lock, took 735,710 tokens, and the 0.45 half failed.

## Why the placement is safe without the check

Two halves on one card are two process groups sharing a GPU like any other, and the repo already has a policy for that. `_validate_gpu_memory_fractions` requires each to declare `runtime.resources.total_gpu_memory_fraction` and caps their sum per GPU. That budget routes through `calculate_stage_budget_available_bytes`, which computes `total_memory_bytes * fraction - own usage`. **Total is a constant**, so the result does not depend on load order.

`PDStagePlacement` gains `memory_fraction` so the two halves can declare different shares:

```yaml
pd_disaggregation:
  prefill: {gpu: 0, memory_fraction: 0.25}
  decode:  {gpu: 0, memory_fraction: 0.65}
```

Measured on one H200, both halves on GPU 0, no `mem_fraction_static` and no `max_total_tokens`:

- Both halves resolved to **21,502 KV tokens**, identically, in every run.
- Confirmed under **both lock orders** — decode-first in four runs, prefill-first in one. The second half to load starts with about 57 GiB already resident (`pre_load_avail_mem=81.17` against `138.78` for the first) and still computes the same `stage_load_used=57.00GiB`.
- Served correct completions, text and image.
- Over-claiming fails loudly and identically on both halves rather than on whichever loses the race: at 0.50 each, `GPU 0 total_gpu_memory_fraction=1.060 exceeds placement limit 1.000`, refused before any GPU work.

## Prior art

Intra-GPU prefill/decode separation is established. [Nexus](https://arxiv.org/abs/2507.06608) reports 1.5-1.9x throughput over vLLM on a single GPU and matches disaggregated vLLM using half the GPUs, and vLLM's [MORI-IO work](https://vllm.ai/blog/2026-04-07-moriio-kv-connector) reports 2.5x goodput disaggregating within one 8-GPU node.

**Their mechanisms are not this one, and the difference matters.** Nexus runs one process with one copy of the weights, shares the KV cache in place, and partitions SMs with CUDA Green Contexts. Two processes on one card carry two copies of the weights, transfer the KV, and allocate a relay pool for multimodal payloads. So this change should be read as making the placement *possible*, not as claiming it is fast.

## What this is for

**Testability.** PD becomes runnable on a one-GPU development box and in CI, where today it needs two cards. It is also the placement that suits a workload whose prefill is a small share of the work: measured on two H200s, a 1P:1D pair's prefill card ran at 10.1% utilization against 67.9% for a colocated card.

**It is not a performance claim.** At two copies of the weights the KV left over is small — 21,502 tokens per half here, against 677,613 on a colocated card. Closing that gap needs the halves to share one copy of the weights, which is separate work.

## It also fixes the sizing defects that make the placement safe

Two halves on one card are two processes sizing pools from one card, so #1685's second defect decides whether the placement is safe at all. This PR fixes all three of the defects reported there.

**A failed attribution is no longer reported as zero.** `get_process_gpu_memory_bytes` matched `os.getpid()` against pids NVML lists for the device. NVML reports host pids, so a caller inside a container never matches itself and every containerized run fell through — returning `0`, which is also what a process holding no memory returns. It now returns `None` for a failed attribution, which is what its own docstring already promised.

**When attribution fails, torch is asked before the global fallback.** The fallback profiles a free-memory delta across model load, and its error depends on what was resident when the "before" sample was taken: the same script and config produced 838,138 and 711,261 KV tokens on two runs. `torch.cuda.memory_reserved` needs no pid join and undercounts only by the CUDA context, measured at 0.51 GiB and identical every run. The profile line now reports which source produced the budget.

**The budget is clamped by what the device reports free.** `calculate_stage_budget_available_bytes` counted only the stage's own usage, so it over-committed whenever anything else held memory on the card — measured, a stage at `fraction=0.87` on a card holding 34 GiB of another process budgeted 122.49 GiB and ran out of memory at launch. Callers that cannot read free memory pass nothing and keep the previous behaviour, so no configuration that works today changes.

That last one matters for this PR specifically: without it, same-GPU sizing is order-independent only because the startup lock serializes weight loading, and the placement check cannot see a process the pipeline does not know about.

## Testing

| Run | Result |
| --- | --- |
| `test_pd_same_gpu.py` (7 cases, new) | pass |
| `tests/unit_test/pipeline` | 534 pass |

One case in `test_pd_rewrite.py` asserted the removed rejection and is dropped.

## Two follow-ups this surfaced

**The declared fraction is not the effective one.** `encoder_mem_reserve=0.05` is subtracted inside each half's budget while the encoder stages also carry their own fractions in the placement sum, so encoders are charged twice. With two halves plus two encoders at 0.03, the placement sum caps each half at 0.47, and 0.47 resolves to an effective 0.420. Charging once would leave materially more room.

**Order-independence here rests on the startup lock** serializing weight loading, not on the formula alone, and the placement check only accounts for stages the pipeline knows about. A foreign process on the card is still unaccounted, which is the second defect in #1685.

### DIFF
--- a/docs/design/pd_handoff.md
+++ b/docs/design/pd_handoff.md
@@ -75,3 +75,44 @@ continuation before associating it with a transfer.
 
 Model-specific state projection and restoration are supplied by sibling model
 integration PRs through the generic PR3 hooks.
+
+## Placement
+
+The two halves may land on different GPUs or on the same one. What PD needs is
+the process split, which happens either way: the prefill step leaves the decode
+scheduler thread whichever card it runs on. Sharing a card also makes PD
+runnable on a one-GPU box and in CI.
+
+Two halves on one card are two process groups sharing a GPU, so the existing
+colocation policy applies unchanged: each must declare a share of the card, and
+the shares on one GPU may not exceed
+`placement.max_total_gpu_memory_fraction_per_gpu`. Declare them per half:
+
+```yaml
+pd_disaggregation:
+  prefill: {gpu: 0, memory_fraction: 0.25}
+  decode:  {gpu: 0, memory_fraction: 0.65}
+```
+
+Use that share rather than `mem_fraction_static` when the halves share a card.
+`total_gpu_memory_fraction` is a fraction of total physical memory, so it does
+not depend on which half loads first. `mem_fraction_static` is computed against
+memory free at load time, and the halves race for one startup lock: measured on
+one H200, whichever half won the lock sized itself to 780,987 KV tokens and the
+other failed to start.
+
+Budget for two copies of the stage's weights on that card. Budget also for the
+CUDA-IPC relay pool if the stage carries multimodal payloads: `mm_aggregate` to
+the prefill half crosses a process boundary even on one device, and the relay
+allocates 1024 MB there. It allocates on the first payload that crosses, not at
+startup, so a deployment can pass startup and still be a gigabyte short when the
+first image arrives. Measured on one H200 with both halves at 32768 KV tokens:
+129,387 MiB of 143,771 used with the relay included.
+
+Prefer the share of the card over an absolute `max_total_tokens` on a shared
+GPU. `max_total_tokens` is applied as a minimum against the profiled capacity,
+so a cap larger than what the later-loading half can profile stops binding on
+that half and the pair reverts to order-dependent sizing without an error.
+Measured: at a cap of 131072 the half that won the lock took 131072 while the
+other logged `max_total_tokens=131072 is larger than the profiled value 50756`
+and took 50756.
--- a/sglang_omni/config/pd_rewrite.py
+++ b/sglang_omni/config/pd_rewrite.py
@@ -122,6 +122,21 @@ def _rewrite_refs(
     )
 
 
+def _half_runtime(stage: StageConfig, memory_fraction: float | None):
+    """Route one half's share of the card into the existing per-stage budget.
+
+    ``runtime.resources.total_gpu_memory_fraction`` is already a fraction of
+    total physical GPU memory, which is what makes it order independent. This
+    only gives the two halves a way to declare different shares of it.
+    """
+    if memory_fraction is None:
+        return stage.runtime
+    resources = stage.runtime.resources.model_copy(
+        update={"total_gpu_memory_fraction": memory_fraction}
+    )
+    return stage.runtime.model_copy(update={"resources": resources})
+
+
 def _split_pd_stage(
     s: StageConfig,
     inbound_rename: dict[str, str],
@@ -138,6 +153,7 @@ def _split_pd_stage(
             "name": prefill_name,
             "gpu": pd.prefill.gpu if pd.prefill.gpu is not None else s.gpu,
             "process": pd.prefill.process or prefill_name,
+            "runtime": _half_runtime(s, pd.prefill.memory_fraction),
             # Note (Yue Yin): Preserve the result path when the first sampled
             # token finishes the request before a KV handoff is needed.
             "next": _rename_targets(s.next, inbound_rename),
@@ -166,6 +182,7 @@ def _split_pd_stage(
             "name": decode_name,
             "gpu": pd.decode.gpu if pd.decode.gpu is not None else s.gpu,
             "process": pd.decode.process or decode_name,
+            "runtime": _half_runtime(s, pd.decode.memory_fraction),
             "next": _rename_targets(s.next, inbound_rename),
             "stream_to": [inbound_rename.get(t, t) for t in s.stream_to],
             "project_payload": {
--- a/sglang_omni/config/schema.py
+++ b/sglang_omni/config/schema.py
@@ -132,13 +132,6 @@ class PDStagePlacement(BaseModel):
 
     gpu: int | list[int] | None = None
     process: str | None = None
-    # Note (Audrey Zheng): this half's share of the whole card, needed when the
-    # halves share a GPU. It routes KV sizing through
-    # `calculate_stage_budget_available_bytes`, which computes
-    # `total x fraction - this stage's own usage`. Total is a constant, so the
-    # result does not depend on which half loads first, and two halves whose
-    # fractions sum to at most 1 cannot claim the same bytes.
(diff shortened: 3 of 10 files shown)


jzheng17:
Context for reviewers, and the PRs this sits next to.

**Why the placement matters, and where the numbers come from.** #1696 carries the equal-GPU-count comparison this PR cites: a 1P:1D pair's prefill card ran at 10.1% mean utilization against 67.9% for a colocated card, because prefill is a small share of this workload's GPU work. That is the case for giving prefill a fraction of a card rather than a whole one.

**What it composes with.** #1690 resolves the Decode target per request, which is the other half of varying the prefill-to-decode ratio: this PR makes the placements expressible, that one makes the choice per request. Neither adds a ratio on its own — the config surface for one prefill half and several decode halves is still missing, and both are prerequisites for it.

**What it does not fix.** The order-independence demonstrated here rests on the startup lock serializing weight loading, not on the sizing formula alone, and the placement check only accounts for stages the pipeline knows about. A foreign process on the card is still unaccounted, which is the second defect in #1685.

**A PD-specific redundancy this surfaced.** `encoder_mem_reserve` is subtracted from a thinker stage's declared budget so the encoders sharing its process have headroom. Under PD that is no longer true: `_split_pd_stage` gives each half its own process while the encoders stay in `pipeline`, so each half reserves headroom for encoders that are not in it. With two halves that is the reserve charged twice, on top of the encoder stages' own fractions in the placement sum. Sending that separately since it is a different file and a different concern.

**Base.** `pd_runtime` rather than the stack tip, because this does not depend on #1599. Context for the whole series is in #1599.

jzheng17:
One decision in here is deliberately left unmade, and it belongs to whoever owns the colocated memory contract rather than to this PR.

`get_process_gpu_memory_bytes` now returns `None` when NVML cannot attribute memory to the caller, and `0` only when the process genuinely holds none. The caller still routes **both** to the global free-memory fallback, because `sglang_model_runner.py` tests `process_memory is None or process_memory <= 0`.

For a genuine zero that is arguably not what you want: a stage holding nothing has a knowable budget, `total x fraction`, and the delta-based fallback throws that away. The pre-load call site at `stages.py:1026` is exactly this case.

I did not change it because it is a policy rather than a defect — tightening that branch changes KV sizing for every model, not only PD. Both branches came in with #430, so I have asked there. Context and the original measurements are in #1685.

jzheng17:
Correction to what I wrote above: `stages.py:1026` is a logging call, not a budget call. It reads `get_process_gpu_memory_bytes` into `pre_load_process_mem` for one startup log line and routes nothing to a fallback. I cited it as an example of a genuine zero reaching the `<= 0` branch, and it is not one.

That changes the answer to my own question. `_profile_available_bytes` runs after the model loads — it takes `pre_model_load_memory` as a parameter and `_profile_available_bytes_from_stage_load_delta` computes the post-load reading itself. By then the process holds the weights, so a genuine zero is not reachable there. The `<= 0` test is defensive rather than a policy about untrusted fractions, and tightening it to `is None` would change nothing that runs.

So there is no decision needed and no follow-up to send. The substantive part is already in #1701: a failed attribution now returns `None` instead of `0`, which is what makes it distinguishable and what lets the caller consult `torch.cuda.memory_reserved` before the global delta. Sorry for the detour.

rwang5203:
Thanks for working on this. I reproduced current head `31c1608b` on H200 using the exact issue [#1685](https://github.com/sgl-project/sglang-omni/issues/1685) configuration and co-tenant setup. The memory-sizing changes in this PR do not fully resolve the reported issue. With the same separate 34 GiB allocation, this branch selected 45.22 GiB for KV, effectively all live free memory, and then OOMed while requesting another 484 MiB. In a real thinker TP2 launch with the co-tenant on one rank, the ranks independently selected 92.07 GiB and 75.04 GiB, after which the constrained rank OOMed while requesting another 802 MiB. These results show that clamping by raw device free memory is not sufficient to make the placement safe.

The remaining issue comes from how available headroom is calculated. `reserve_bytes` defaults to zero and the production caller never supplies a reserve, so the KV pool can consume memory still needed by the allocator and runtime. There is also no distributed minimum, so ranks can construct incompatible capacities. The PID `None` distinction and Torch fallback improve attribution, but they do not address either condition. The tests assert the raw-free result without allocating it and contain no distributed validation. Current head also rejects the advertised same-GPU Qwen PD path during runtime preparation because the Qwen thinker factory is not marked with `@pd_disaggregation_capable`.

[PR 1702](https://github.com/sgl-project/sglang-omni/pull/1702) addresses the underlying sizing problem. It charges all live device usage against the aggregate placement budget and then takes the minimum capacity across ranks. Under the same reproduction it selected 42.10 GiB, reached readiness, and served validated requests. Under TP2 both ranks adopted the constrained rank's 63.80 GiB capacity and completed startup. This handles both device overcommit and rank disagreement without hard-coding the model, GPU, or tenant size. I suggest keeping this PR focused on the PD work and using the aggregate device cap plus distributed minimum for the memory-sizing fix.

jzheng17:
Folded into #1684. The two were one change: the flag, the per-half attributes it sets, the same-GPU allowance those attributes exist for, and the sizing fixes that make sharing a card safe. Kept apart they did not pass their own tests — #1684's share syntax was rejected by the same-GPU check it had not yet removed, and #1701 used a field it no longer defined. Combined: 573 pass. No content dropped.

PR #1703, opened 2026-08-25
fix(memory): distinguish failed attribution from zero, and clamp the KV budget

jzheng17:
Based on `pd_runtime`. Fixes the three defects reported in #1685.

## The attribution never succeeds in a container

`get_process_gpu_memory_bytes` matches `os.getpid()` against the pids NVML lists for the device. NVML reports pids from the host namespace and the caller sees its namespace pid, so the join never matches and **every containerized run falls through**.

Falling through was not the problem on its own. The problem is what it fell through *as*: the function returned `0`, which is also what it returns for a process that genuinely holds no memory. `sglang_model_runner.py` reads `<= 0` as "NVML unavailable" and takes a global free-memory fallback, so both cases arrive at the same place and neither is visible in a log.

**This PR returns `None` for a failed attribution**, which is what the function's own docstring already promises, and keeps `0` for a genuine zero.

## When attribution fails, ask torch before falling back

The fallback profiles a global free-memory delta across model load. Its error depends on what was already resident when the "before" sample was taken, and it is not reproducible: the same script and the same config produced 838,138 and 711,261 KV tokens on two runs, leaving the card at 97.7% and 90.9%.

`get_process_gpu_memory_bytes_from_torch` reads `torch.cuda.memory_reserved`, which needs no pid join and no namespace agreement. It undercounts by the CUDA context — measured at 0.51 GiB in a small controlled process — and that error is bounded and the same every run. The profile line now reports `gpu_mem_accounting=torch_reserved` when this path runs, so which source produced a budget is visible.

## The budget counted only its own usage

`calculate_stage_budget_available_bytes` computes `total x fraction - this stage's usage` and never subtracts memory belonging to anything else on the card. On a card already holding 34 GiB of another process, a stage at `fraction=0.87` budgeted 122.49 GiB and ran out of memory at launch.

It now takes an optional `free_memory_bytes` and clamps to it, with an optional `reserve_bytes` held back. A caller that cannot read free memory passes nothing and keeps the previous behaviour, so this changes no configuration that works today — it only stops one that would have failed.

`get_device_free_memory_bytes` reads `torch.cuda.mem_get_info`, again with no pid join.

## Why this matters beyond the original report

Two PD halves sharing a GPU are two processes sizing pools from one card, so the second defect is what decides whether that placement is safe. It currently works only because the startup lock serializes weight loading — see the discussion on #1701. The clamp makes it safe without depending on that ordering, and it also covers the case the placement check cannot see, which is a process the pipeline does not know about.

## What this PR does not change

The `<= 0` branch still exists for a genuine zero, which routes to the fallback. That is arguably still wrong — a stage that truly holds nothing has a knowable budget — but changing it alters sizing for configurations that work today, and that is a policy call rather than a defect.

## Testing

| Run | Result |
| --- | --- |
| `test_kv_sizing_fixes.py` (5 cases, new) | pass |
| `tests/unit_test/pipeline` | 534 pass |

`test_gpu_memory.py::test_get_process_gpu_memory_returns_zero_when_pid_not_present` asserted the behaviour this PR changes. It is now two tests: one that a failed attribution reports `None`, one that a genuine zero still reports `0`.

### DIFF
--- a/sglang_omni/model_runner/sglang_model_runner.py
+++ b/sglang_omni/model_runner/sglang_model_runner.py
@@ -15,6 +15,8 @@ from sglang.srt.server_args import PortArgs, ServerArgs
 
 from sglang_omni.model_runner.prefill_inputs import get_omni_prefill_inputs
 from sglang_omni.utils.gpu_memory import (
+    get_device_free_memory_bytes,
+    get_process_gpu_memory_bytes_from_torch,
     calculate_stage_budget_available_bytes,
     calculate_stage_load_delta_bytes,
     format_bytes_gib,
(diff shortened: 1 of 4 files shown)


jzheng17:
Folded into #1701 rather than kept as a separate PR. The clamp is what makes two halves sizing pools from one card safe without depending on the startup lock, so it belongs with the placement change it protects. Same three fixes, same tests, no content dropped.

PR #1708, opened 2026-08-25
feat(comm): let the CUDA-IPC relay preallocate its pool

jzheng17:
Based on `pd_runtime`. One flag, off by default.

## The failure this prevents

`CudaIpcRelay` allocates its pool on the first payload that crosses a process boundary, not at construction. A deployment whose memory budget cannot hold the pool therefore starts cleanly and fails later, when traffic first needs it.

Measured on a PD pair: at `mem_fraction_static=0.87` both halves launched, served text for 19 minutes, and then failed a 750 MiB allocation the moment the image arm began — `[mm_aggregate_relay] Allocating CUDA-IPC pool: 1024.00 MB on cuda:0` appears in the log at that instant and not before. The prefill card was at 139.5 GiB of 139.80 with `thinker_prefill` holding 125.92 and the encoder process 13.59.

**Text traffic cannot catch this.** `router.py` routes a payload with no CUDA tensors over SHM, so `mm_aggregate` never reaches the pool on a text-only run. A text smoke test passes and the deployment is still a gigabyte short.

PD makes it reachable in a way colocated is not: `qwen3_omni/config.py` puts every stage in `process="pipeline"`, so `mm_aggregate` to `thinker` is an in-process handoff, while `pd_rewrite._split_pd_stage` gives the prefill half its own process and pushes that edge across a boundary. Colocated logged zero pool allocations across the whole comparison; PD logged one.

## The change

`preallocate_pool=False` on `CudaIpcRelay`. When set, the pool is allocated at the end of construction, so a budget that cannot hold it fails at launch with the allocation error rather than under load.

Default is off, so nothing that runs today changes size or startup time.

## What this does not do

It does not make the pool part of any memory budget. The sizing path still does not know the relay will allocate, so an operator has to leave room for it. Making the budget account for it is a larger change and belongs with whoever owns the sizing contract — see #1685.

## Testing

| Run | Result |
| --- | --- |
| `test_cuda_ipc_relay.py` (2 cases, new) | pass |
| `tests/unit_test/relay`, `tests/unit_test/pipeline` | 541 pass |

One case asserts the default stays lazy, so the flag cannot silently become the default.

### DIFF
--- a/sglang_omni/relay/cuda_ipc.py
+++ b/sglang_omni/relay/cuda_ipc.py
@@ -548,6 +548,7 @@ class CudaIpcRelay(Relay):
         credits: int | None = 2,
         slot_size_kb: int = 64,
         pool_size_mb: int | None = None,
+        preallocate_pool: bool = False,
         **kwargs: Any,
     ) -> None:
         if kwargs:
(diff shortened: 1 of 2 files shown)


jzheng17:
Folded into #1684. The relay pool's lazy allocation is the same subject as the sizing fixes there — the budget not reflecting what will be on the card — and a 36-line flag is not a review on its own. Same change, same tests, nothing dropped.

Issue #1704, opened 2026-08-25
[Bug] PD returns empty completions under queue depth; degraded rows share KV slots with live requests


jzheng17:
A correctness defect in the PD path, with the measurement that localizes it. It is not a performance issue and it is silent.

## What a user sees

A request returns HTTP 200, `finish_reason: "stop"`, `completion_tokens: 1`, correct `prompt_tokens`, and an empty delta. Two SSE frames, the terminal chunk and `[DONE]`, with no content frame. The single token is 151645, `<|im_end|>`, and it ends the request through `FINISH_MATCHED_TOKEN` as a matched stop token rather than as a member of `eos_token_ids` — so a probe that checks `eos_token_ids` cannot see this population, which cost us one instrumented run.

The response is well formed. Nothing logs an error.

## Rate

PD only. Measured on a clean two-GPU pair with mixed prompt lengths, `mem_fraction_static=0.87`, zero transport errors:

| offered | achieved | in-flight p50 | empty completions per 3000 |
| --- | --- | --- | --- |
| 6 | 5.96 | 8 | 0 |
| 8 | 7.92 | 12 | 0 |
| 10 | 9.90 | 20 | 1 |
| 12 | 10.64 | 175 | 24 |
| 16 | 10.91 | 535 | 48 |

A colocated replica under an identical driver produced 0 in 3000, and 0 in 28,472 at the longest prompt length.

## What it is not

**Not the server args PD forces.** A colocated arm with `page_size=1` and `disable_radix_cache` set, same driver, same mixed lengths, same long-lived server, produced zero.

**Not the KV transfer or the handoff.** The single token is sampled by the prefill half before any transfer occurs, so a decode-side probe never fires for this population.

**Not sampling numerics.** Decoding is greedy. Across two runs the failing rows' logit gaps span the control distribution, with a median of 13 bf16 ULP against the control's 9, and four of eleven above 33 ULP. Colocated never places the terminator in its top two: 0 of 2237 prefill rows, against a rate that predicts about 9.8, Poisson p = 5.6e-5.

**Not prefill batch composition.** Raising `prefill_coalesce_requests` from the default to 8 moved mean batch size 2.73 to 3.11 and cut single-request batches from 20.6% to 6.7%, with no rise in the failure rate. Failures occur at batch size 1.

**Not prompt length.** Short prompts fail once queues are deep. An earlier length correlation was queue depth in disguise.

## What it is

[...]

PR #1709, opened 2026-08-25
feat(pd): make PD overload visible, and bound handoffs in flight

jzheng17:
Based on `pd_runtime`. Off by default.

## Nothing bounds what Decode accumulates

A colocated replica throttles admission without trying to: prefill and decode contend for one card and one scheduler thread, so admitting more work slows the admitting itself. Splitting the two halves removes that feedback, and nothing replaces it.

Measured on two H200s, PD against colocated at the same offered rate of 16, where both admitted every request:

| | Decode in flight | total time p50 | admission |
| --- | --- | --- | --- |
| colocated replica | ~18 | 2.29 s | 100% |
| PD pair | ~437 | 40.96 s | 100% |

`max_running_requests` is 64, so by Little's law most of that 437 is queueing rather than decoding. Both arms report 100% admission throughout, which is why this does not show up as a failure: **admission rate is constant on this path**, reading 100% from a healthy request to a 41-second one, so overload arrives as unbounded latency rather than as anything that says so.

## What this PR does about it

Three things, and a bound on Decode's queue is not one of them.

**It says which admission policy is in effect.** Binding a Decode half with no queue limit logs that the queue is unbounded, that overload will appear as latency rather than rejection, and which flag to set. `--max_queued_requests` already reaches both halves — `PDConfig` is written onto the logical stage and expansion copies server args to both halves — and `models/qwen3_tts/engine_builder.py` sets it to 16 for its generation stage, so the notice names that precedent rather than leaving "set something" as a guess. Without that line an operator sees 100% admission and 41-second requests with nothing connecting the two.

**It measures how much Decode is holding.** The depth rides the transfer ack, so it costs no new message and no new route, and `DataAckMessage.receiver_pending` is optional so a peer that omits it stays compatible. The count includes committed-but-unadmitted work alongside the scheduler's queues, because a request in either state is work this half accepted and has not finished.

This measurement is reported, not acted on. The ack fires on commit, which means the reading stops arriving exactly when Prefill stops sending: a bound built on it would hold on a stale high-water value with nothing left to lower it. Bounding what Decode accumulates needs a capacity return that does not ride the producer's traffic, and that is separate work. The measurement is what such a return would be built from.

**It bounds how many handoffs send at once.** `max_inflight_handoffs` is a semaphore around the send. What it bounds is concurrent sends — not pinned prompt KV, because the scheduler builds `SGLangKVPageLease` before the send task starts, so six requests behind a limit of two hold six leases and send two. Bounding pinned KV means refusing the request before its KV exists, which is prefill admission and belongs with the capacity-return work above.

Cancelling a handoff while it waits for a permit used to leak that lease outright: the send never started, so its `finally` never ran and nobody released it. The wait now releases the lease and clears the request state on cancellation. `release` is idempotent, so a handoff that did start is unaffected.

## Also here

`priority` now survives the handoff. A `DecodeContinuation` carries it and `req_from_continuation` restores it, so a request that was admitted by priority on Prefill is not silently demoted on Decode.

`LockedKVAllocator` serialises `alloc` and `free` on the Decode half, where the scheduler thread and the comm event loop share one allocator (#1704). `bootstrap.py` hands the raw allocator to `create_tree_cache`, `PrefillManager` and `DecodeManager` before `bind_pd_runtime` runs, so the wrapper is now installed into all four holders rather than the scheduler's attribute alone — one lock only helps if no caller keeps an unwrapped alias.

Rebinding after construction rather than wrapping in `bootstrap.py` is deliberate: upstream's `SWAChunkCache.__init__` asserts the allocator's concrete type, and a proxy handed to the constructor would fail that assert for models that use it. A holder without the attribute raises rather than being skipped, and `_kv_allocator_holders` reports what was rebound so a test asserts they all agree — which is what catches a fifth holder appearing.

The open question is whether locking is the right answer at all. The instrumented run behind #1704 recorded 21,309 cross-thread alloc/free overlaps, and the alternative is to remove the second caller entirely: have the comm thread hand its reservation to the scheduler thread rather than allocating concurrently. That trades a lock on a hot path for a hop in `prepare_kv_receive`.

## Testing

| Run | Result |
| --- | --- |
| `test_pd_alloc_lock.py`, `test_pd_handoff_gate.py`, `test_pd_backpressure.py` | 24 pass |
| `tests/unit_test` | 4951 pass, 7 fail |

The 7 failures are `moss_tts/test_audio_tokenizer.py` (2), `qwen3_omni/test_vision_compat.py` (1) and `qwen3_tts/test_compat.py` (4). All 7 fail the same way on `origin/main` in this environment and none touch PD.

`_pd_handoff_gate` reads `pd_execution` defensively. On a `Stage` built with `object.__new__` a direct attribute read raises inside the handoff task, which stops the send with nothing surfacing the cause, and `test_pd_stage_runtime.py` blocks on `await stage._comm.started.wait()` rather than failing.

The cases cover: the depth counts every accepted request and reads zero when idle; the ack carries it across the wire and a peer that omits it stays compatible; the unbounded-queue notice fires on a Decode half, names the precedent, and stays silent on a bounded queue and on the Prefill half; priority survives the handoff and a continuation without one stays valid; an unset bound installs no gate and a set one installs a semaphore of that size, reused across handoffs; cancelling while queued releases the lease exactly once while a handoff that never waited leaves it to the send path; two threads racing the allocator's read-modify-write hand out disjoint slots through the wrapper; every holder ends up with the same locked object; and a holder that cannot be rebound raises rather than being skipped.

### DIFF
--- a/sglang_omni/comm/engine.py
+++ b/sglang_omni/comm/engine.py
@@ -127,6 +127,8 @@ class CommEngine:
         self._retained_pending_kv_transfers: list[_PendingTransfer] = []
         self._kv_pools: dict[str, KVPool] = {}
         self._kv_receivers: dict[str, KVReceiver] = {}
+        # Note (Audrey Zheng): the newest depth each peer reported on an ack.
+        self._peer_pending: dict[str, int] = {}
         self._kv_ready: dict[str, asyncio.Future[KVTransferReadyMessage]] = {}
         self._outbound_kv_requests: dict[str, str] = {}
         self._inbound_kv: dict[str, _InboundKVTransfer] = {}
@@ -346,6 +348,32 @@ class CommEngine:
     def register_kv_receiver(self, pool_id: str, receiver: KVReceiver) -> None:
         self._kv_receivers[pool_id] = receiver
 
+    def _local_pending_depth(self) -> int | None:
+        """Requests this stage's receivers still hold, or None if unreported.
+
+        A receiver that does not implement ``pending_depth`` contributes
+        nothing, so a stage whose receivers all predate the method reports
+        nothing rather than a misleading zero.
+        """
+        total: int | None = None
+        for receiver in self._kv_receivers.values():
+            depth = getattr(receiver, "pending_depth", None)
+            if depth is None:
+                continue
+            try:
+                value = depth()
+            except Exception:
+                logger.debug("pending_depth failed", exc_info=True)
+                continue
+            if value is None:
+                continue
+            total = int(value) if total is None else total + int(value)
+        return total
+
+    def peer_pending(self, stage: str) -> int | None:
+        """Newest depth *stage* reported, or None if it has never reported one."""
+        return self._peer_pending.get(stage)
+
     async def send_kv_pages(
         self,
         *,
@@ -535,6 +563,7 @@ class CommEngine:
                 object_id=data_ref.object_id,
                 success=error is None,
                 error=None if error is None else (str(error) or type(error).__name__),
+                receiver_pending=self._local_pending_depth(),
             ),
         )
 
@@ -721,6 +750,10 @@ class CommEngine:
             raise ValueError(
                 f"data_ack for {ack.to_stage!r} delivered to {self.router.stage_name!r}"
             )
+        if ack.receiver_pending is not None:
+            # Note (Audrey Zheng): record before the staleness check, because a
+            # stale ack still carries a current reading of the peer's depth.
+            self._peer_pending[ack.from_stage] = ack.receiver_pending
         pending = self._pending.get(ack.object_id)
         if pending is None:
             logger.debug(
--- a/sglang_omni/config/pd_rewrite.py
+++ b/sglang_omni/config/pd_rewrite.py
@@ -156,7 +156,11 @@ def _split_pd_stage(
             "stream_to": [inbound_rename.get(t, t) for t in s.stream_to],
             "pd_disaggregation": None,
             # Note (Yue Yin): Keep compiler metadata out of user factory kwargs.
-            "pd_execution": PDExecution(role="prefill", partner=decode_name),
+            "pd_execution": PDExecution(
+                role="prefill",
+                partner=decode_name,
+                max_inflight_handoffs=pd.max_inflight_handoffs,
+            ),
         },
     )
 
@@ -175,7 +179,11 @@ def _split_pd_stage(
             "wait_for_fn": None,
             "merge_fn": None,
             "pd_disaggregation": None,
-            "pd_execution": PDExecution(role="decode", partner=prefill_name),
+            "pd_execution": PDExecution(
+                role="decode",
+                partner=prefill_name,
+                max_inflight_handoffs=pd.max_inflight_handoffs,
+            ),
         },
     )
 
--- a/sglang_omni/config/pd_rewrite.py
+++ b/sglang_omni/config/pd_rewrite.py
@@ -160,6 +160,7 @@ def _split_pd_stage(
                 role="prefill",
                 partner=decode_name,
                 max_inflight_handoffs=pd.max_inflight_handoffs,
+                decode_pending_limit=pd.decode_pending_limit,
             ),
         },
     )
@@ -183,6 +184,7 @@ def _split_pd_stage(
                 role="decode",
                 partner=prefill_name,
                 max_inflight_handoffs=pd.max_inflight_handoffs,
+                decode_pending_limit=pd.decode_pending_limit,
             ),
         },
     )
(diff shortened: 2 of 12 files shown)


MelodyyyYin:
#note:（yue yin）

`decode_pending_limit` is currently a no-op because `Stage` never wires `peer_pending_fn` into `bind_pd_runtime`. Wiring only the existing ACK reading would then freeze Prefill on a stale high-water value after Decode drains without another transfer ACK.

Follow-up PR: https://github.com/jzheng17/sglang-omni/pull/2

Why: a hard admission bound needs both an installed reader and an independently deliverable capacity/credit return after the producer stops; handoff-triggered telemetry cannot provide safety or liveness.

MelodyyyYin:
#note:（yue yin）

`max_inflight_handoffs` is acquired in the async send task after scheduler detach and `SGLangKVPageLease` creation. A six-request batch with limit two therefore pins six prompt KVs, and cancellation while waiting has no lease owner that releases them.

Follow-up PR: https://github.com/jzheng17/sglang-omni/pull/5

Why: handoff capacity must be reserved before protected KV ownership transfers, with capacity and lease released exactly once on every terminal path.

MelodyyyYin:
#note:（yue yin）

`LockedKVAllocator` replaces only the scheduler field after `tree_cache` and scheduling components cached the raw allocator. Comm allocation can therefore race a cache/result-path free through an alias outside the wrapper, corrupting free-list accounting and reissuing live slots.

Follow-up PR: https://github.com/jzheng17/sglang-omni/pull/3

Why: every alloc/free for one KV pool must share one synchronization domain; locking only one late-bound alias cannot protect allocator correctness.

jzheng17:
All three are right, and I checked each one against the branch rather than taking them on the argument. Thank you — the first one invalidates a number I posted on #1599, so I am glad you caught it before it went further.

**`decode_pending_limit` is inert.** `peer_pending_fn` appears three times in the tree, all inside `omni_scheduler.py`: the parameter, the assignment, and the read. Nothing in `stage/runtime.py` passes it, so `_pd_peer_pending()` returns None on every call and `get_new_batch_prefill` never holds. The gate has never run.

That means the comparison I posted on #1599 — same-GPU PD at 2.83 req/s unbounded against 3.01 bounded at 64 — is two runs of the same configuration. I will withdraw it there. I am dropping `decode_pending_limit` from this PR rather than wiring it, because your second sentence is the harder half of the objection: the depth reading rides the transfer ack, so if Prefill stops sending on a high-water value, no further ack arrives to lower it and Prefill never resumes. A bound that can deadlock on its own success needs a capacity return that does not depend on the producer, and that is a separate piece of work with its own design question.

**`max_inflight_handoffs` bounds the wrong thing.** The permit is acquired in `_send_pd_handoff`, which runs as a task created after the scheduler has already built `SGLangKVPageLease(req, self.tree_cache)` and put it on the outbox. So six requests with a limit of two hold six leases and send two. Cancellation while waiting on the semaphore is worse: `_send_pd_handoff_now` never runs, so its `finally` never runs, and nothing releases the lease.

So what it bounds is concurrent sends, not pinned prompt KV, and the PR describes it as the latter. Bounding pinned KV means refusing the request before its KV exists, which is prefill admission — the same place the first note points at. I will fix the cancellation leak and describe the flag as what it actually bounds, and leave the KV bound to the admission work.

**The allocator lock covers one alias out of four.** `bootstrap.py` hands the raw allocator to `create_tree_cache`, `PrefillManager` and `DecodeManager` before `bind_pd_runtime` rebinds `self.token_to_kv_pool_allocator`, so those three keep calling the unwrapped object. The scheduler thread reaches the free list through them while the comm thread goes through the wrapper, which is precisely the pair the lock was meant to serialise.

I am moving the wrap into `bootstrap.py`, before the allocator is handed to any of them, so every holder shares one object. That is the only place where one wrapper reaches all four.

One thing I would like your read on. The instrumented run behind #1704 recorded 21,309 cross-thread alloc/free overlaps, which is a lot for a lock to absorb on the decode path. Would you rather the comm thread not allocate at all — hand the reservation to the scheduler thread and let it allocate on its own step — instead of sharing a lock? That removes the synchronisation domain question rather than answering it, at the cost of a hop in `prepare_kv_receive`.

jzheng17:
Redone as #1804 against `feat/minimal-pd-disaggregation`, which is where the main line is now.

Three of the four things here came across: the unbounded-queue notice, `max_inflight_handoffs`, and `priority` surviving the handoff. `decode_pending_limit` did not, because it never ran — nothing passed `peer_pending_fn` into `bind_pd_runtime`, so the depth reader was always `None` and the admission gate never held. Wiring it as it stood would not have worked either: the reading rides the transfer ack, so it stops arriving exactly when Prefill stops sending, and the bound would hold on a stale high-water value with nothing to lower it.

The measurement that motivated it stands and is in #1804: at offered 16 the Decode half held about 437 requests against `max_running_requests` 64, and a request took 40.96 s against 2.29 s colocated, with admission reading 100% throughout.

PR #1712, opened 2026-08-25
fix(pipeline): three fixes to what an operator can see and set

jzheng17:
Three fixes to what a deployment can see and set. None depends on the PD stack.

## A startup failure reported an exit code, not the traceback

`wait_ready` checked whether the child was alive first and only then read the error channel. A child that reports a factory error exits right after reporting, so the two race: while the child is still alive the loop falls through and waits, and the failure surfaces as a timeout or a bare exit code rather than as the traceback the child already sent.

Reading the channel first reports the traceback whenever there is one, whichever state the child is in. The liveness branch keeps its short wait, since the child hands the bytes to a feeder thread and then exits, so they can still be in flight.

This was #1683 and is folded in here.

## The chunking log reported state, not transitions

`get_new_batch_prefill` logged whether chunking was disabled on every call that reached it, so a steady state produced one line per scheduler pass. The line an operator wants is the one where the answer changed. `_log_chunking_transition` keeps the previous value and logs only when it moves, which is also what makes the line worth reading: a transition names a cause, a repeated state does not.

## The waiting-queue timeout had no flag

`Scheduler._abort_on_waiting_timeout` drops a request that has waited longer than `SGLANG_REQ_WAITING_TIMEOUT` and answers it with HTTP 503 and `Request waiting timeout reached.` Omni reaches that method through the borrowed `get_next_batch_to_run`, so the bound already worked here. Nothing in this repo referenced the variable, no flag set it, and no document mentioned it.

`--req-waiting-timeout` sets it. `apply_request_waiting_timeout` writes the environment variable, because upstream reads it from there rather than off `server_args`, and reads it per call; stage processes are spawned and inherit the environment, so setting it in the parent reaches all of them.

It answers a different question from `--max-queued-requests`. The timeout bounds how long one request waits; the queue length bounds how many wait. A duration is the bound an operator can pick without a model of the system: a queue length converts to a wait only through the current service rate, so 16 queued is two seconds at one load and forty at another, while a timeout is the caller's own timeout and tracks load without retuning. The length keeps the job a duration cannot do, which is bounding memory.

## Why they are together

All three are about the same surface: one hides a failure behind an exit code, one prints too much, the last exposes too little. The timeout flag was written for #1709, where PD's unbounded decode queue motivated it, but nothing in it is PD-specific and keeping it there would hold it behind a four-PR stack.

## Testing

Four tests cover the flag: the value reaches the environment, omitting it leaves an exported value alone, upstream's `EnvFloat.get()` parses back what the flag wrote, and the default leaves the bound off. The chunking-log tests exercise `_log_chunking_transition` itself rather than restating its rule.

Related: #1709.


jzheng17:
Superseded by #1806, on `main` and without the conflict.

The chunking-transition log is dropped rather than rebased: it instrumented `PrefillManager.disable_chunking`, and `sglang_backend/prefill.py` was deleted in the 0.5.18 bump (#1719). `disable_chunking` now exists in neither this repo nor sglang 0.5.18, so there is nothing left to instrument.

PR #1715, opened 2026-08-25
[Perf][Fun-CosyVoice3] Fix vocoder autocast so bf16 config actually applies

### LINKED ISSUE #1713: [Bug] Fun-CosyVoice3 flow/hift vocoder always runs FP32 despite dtype: bfloat16 config
### Checklist

### Describe the bug

Fun-CosyVoice3's flow (DiT) and HiFT vocoder decode path never actually runs in reduced precision, even though the shipped config (examples/configs/fun_cosyvoice3_0_5b.yaml) sets `dtype: bfloat16`.

A fix and reproduction/benchmark scripts are available in #1715 .

Root cause, in sglang_omni/models/fun_cosyvoice3/stages.py:

    flow, hift = _load_cosyvoice3_flow_hift(
        checkpoint_dir,
        device=device,
        fp16=(dtype == "float16"),
    )

    return _CosyVoice3Vocoder(flow, hift, fp16=(dtype == "float16")).build_scheduler(

`dtype == "float16"` is always False for the shipped `dtype: bfloat16` config, so `_CosyVoice3Vocoder._fp16` is always False and `torch.autocast(..., enabled=self._fp16)` never activates. Flow and HiFT run in plain FP32 as loaded from the checkpoint.

A second, independent issue: even when fp16=True, only flow.inference() is wrapped in the autocast scope -- hift.inference() is called outside it. This diverges from the upstream reference implementation (cosyvoice.cli.model.CosyVoice3Model.token2wav), which wraps both calls in the same torch.cuda.amp.autocast(self.fp16) block.

Confirmed via torch.profiler on a B300: with the shipped config, the top ops in flow.inference() (aten::addmm, aten::cudnn_convolution, cutlass sgemm kernels -- ~85% of total time) are all plain-FP32 kernels, not the bf16-tensor-core variants.

Impact: flow+hift decode runs ~1.7-1.8x slower than necessary (measured stable across repeated runs on a B300); measured end-to-end per-request latency impact is ~1.1-1.2x (smaller since AR generation dominates total request time).

### Reproduction

Serve Fun-CosyVoice3 with the default config:

    sgl-omni serve --model-path FunAudioLLM/Fun-CosyVoice3-0.5B-2512 \
      --config examples/configs/fun_cosyvoice3_0_5b.yaml --port 8000

Send a /v1/audio/speech request, then profile `_CosyVoice3Vocoder._token2wav` (or the underlying `flow.inference` / `hift.inference` calls) with torch.profiler -- the resulting op table shows FP32-only kernels despite `dtype: bfloat16` in the config.

### Environment

- sglang-omni: commit 99153976 (main)
- GPU: NVIDIA B300 (SXM6, 275GB), CUDA 13
- Python 3.12.13, torch 2.11.0
- Checkpoint: FunAudioLLM/Fun-CosyVoice3-0.5B-2512
- Config: examples/configs/fun_cosyvoice3_0_5b.yaml (dtype: bfloat16)

YijunWang1121:
## Motivation

`_CosyVoice3Vocoder` only ever enabled autocast when `dtype == "float16"`, so the shipped `dtype: bfloat16` config silently left Flow+HiFT running in plain FP32 (confirmed by profiling -- addmm+conv+attention ~85% of Flow's 300ms/iter, all FP32 on a B300). `hift.inference()` was also called outside the autocast scope entirely, unlike upstream `CosyVoice3Model.token2wav`, which wraps both calls in one autocast block. See #<bug-issue-number> for the full writeup.

## Modifications

- Map `dtype` to an actual `torch.dtype` via `_AUTOCAST_DTYPES` (`{"float16": torch.float16, "bfloat16": torch.bfloat16}`), threaded through as `compute_dtype` instead of a `fp16: bool`.
- Wrap both `flow.inference()` and `hift.inference()` in the same `torch.autocast(...)` scope, mirroring upstream.

## Related Issues

Fixes #1713 
Relates to #1652 (found while profiling Flow/HiFT for fusion candidates against that roadmap)

## Accuracy Test

Compared FP32 vs bf16 output on identical fixed inputs (same tokens/prompt/embedding/seed), B300:
- Flow's mel output (pre-vocoder, phase-free): **SNR 30.15dB, corr 0.9986**
- Vocoder output spectral magnitude (phase-invariant): **SNR 13.04dB, corr 0.974**
- **No NaN/Inf** in either mel or waveform, consistent across repeated runs
- Raw time-domain waveform correlation is low (0.35) but this is expected/uninformative for vocoder output (phase sensitivity to tiny mel differences), **not indicative of broken output**
- E2E HTTP smoke test: **200 OK**, valid 24kHz/mono/16-bit wav, same duration/size as pre-fix baseline
- **Manually verified by listening to the FP32 vs bf16 A/B pair (same fixed tokens) -- no audible quality regression.**

## Benchmark & Profiling

Measured on 1x B300, multiple independent repetitions:
- Isolated flow+hift call: FP32 ~309-326ms avg -> bf16 ~174-190ms avg (**1.71-1.78x speedup**, stable across 3 independent runs)
- End-to-end HTTP request (same fixed input/seed): FP32 ~1.10-1.17s avg -> bf16 ~0.99-1.01s avg (**1.10-1.16x**, stable across independent server-restart cycles)
- E2E speedup is smaller than the isolated-call speedup because AR/LLM generation (already using SGLang's own fused kernels, untouched by this PR) dominates total per-request time

## Checklist

- [x] Add unit tests. (existing test_vocoder.py suite covers the changed constructor; 40/40 pass)


charliechenye:
@YijunWang1121 Nice finding. I reproduced the missing BF16 autocast behavior on an H200, and found one additional detail that may be important for interpreting the perf result: the BF16 benefit is strongly dependent on generated speech-token length.

Environment:
- 1x NVIDIA H200
- PyTorch 2.11.0+cu130
- CUDA 13.0
- `float32_matmul_precision=highest`
- `matmul_allow_tf32=false`
- `cudnn_allow_tf32=true`

I benchmarked the core behavior of this PR, comparing:

- FP32: Flow FP32 + HiFT FP32
- BF16: Flow + HiFT under one shared `torch.bfloat16` autocast scope

with the same prompt conditioning and fixed seed.

| Generated speech tokens | FP32 p50 | BF16 p50 | BF16 vs FP32 |
| ---: | ---: | ---: | ---: |
| 50 | 181.4 ms | 221.9 ms | 0.817x |
| 107 | 194.5 ms | 226.1 ms | 0.860x |
| 200 | 267.8 ms | 222.8 ms | 1.202x |
| 300 | 326.5 ms | 222.4 ms | 1.468x |
| 400 | 352.8 ms | 223.6 ms | 1.578x |

So on this H200 there is a clear crossover at roughly 150 speech tokens, around 6 seconds of generated audio at the 25 Hz speech-token rate.

For short outputs, BF16 is materially slower. Once the output is long enough, BF16 becomes faster and the advantage grows with sequence length.

One interesting point is that the 300-token FP32 result on H200 is 326.5 ms, almost exactly at the upper end of the 309-326 ms FP32 range reported in this PR. At that shape, BF16 is already 1.47x faster on H200, and at 400 tokens it reaches 1.58x. This makes me think benchmark shape explains a substantial part of the difference between my earlier short-request results and the B300 measurements here.

The same-noise numerical comparison also looks healthy on H200:
- Flow mel: SNR 36.9 dB, corr 0.9995
- Waveform spectral magnitude: SNR 11.8 dB, corr 0.9646
- No NaN/Inf

These are broadly consistent with the quality results reported here.

Could you share the generated speech-token count, or resulting mel length, used for the isolated 309-326 ms -> 174-190 ms benchmark? I think that would make the B300 and H200 measurements much easier to compare directly.

YijunWang1121:
@charliechenye  Thanks for the detailed H200 data — this pushed me to properly re-measure instead of relying on a single data point.

> Could you share the generated speech-token count, or resulting mel length, used for the isolated 309-326 ms -> 174-190 ms benchmark?

The original measurement used **~125 synthetic speech tokens (~5s at 25Hz)**. I recovered the exact original validation script and reran it verbatim against this PR's code — it reproduces the headline number exactly: fp32=306.4ms → bf16=174.6ms (1.76x), matching the "309-326ms → 174-190ms" range in the PR description, with raw-waveform SNR=-0.90dB/corr=0.349 matching to two decimal places.

I then ran a full length sweep on B300 (25-800 tokens, 10 warmup + 50 measured iters/point, `float32_matmul_precision=highest / matmul_allow_tf32=False / cudnn_allow_tf32=True` — same settings you used; verified with a direct read of these flags from inside `flow.inference()`/`hift.inference()` at call time, so they're confirmed to actually be in effect, not just set and silently overridden):

| tokens | FP32 p50 | BF16 p50 | BF16 vs FP32 |
|---:|---:|---:|---:|
| 25 | 203.5ms | 139.0ms | 1.46x |
| 50 | 280.9ms | 166.9ms | 1.68x |
| 75 | 287.8ms | 158.7ms | 1.81x |
| 100 | 279.3ms | 158.6ms | 1.76x |
| 125 | 307.1ms | 174.6ms | 1.76x |
| 150 | 314.6ms | 186.9ms | 1.68x |
| 175 | 318.8ms | 175.9ms | 1.81x |
| 200 | 424.5ms | 202.2ms | 2.10x |
| 250 | 440.2ms | 211.6ms | 2.08x |
| 300 | 460.4ms | 222.2ms | 2.07x |
| 400 | 564.0ms | 248.0ms | 2.27x |
| 500 | 720.1ms | 321.6ms | 2.24x |
| 800 | 1077.5ms | 421.8ms | 2.55x |

**No crossover on B300** across the whole range — bf16 is faster at every point tested, with the margin generally growing with length (some noise around 100-175 tokens, but never below 1.46x).

Quality across the sweep: mel SNR 28-34dB (stable), waveform spectral-magnitude SNR noisier (10-16dB, tied to specific synthetic content rather than token length), no NaN/Inf in any run.

charliechenye:
> @charliechenye Thanks for the detailed H200 data — this pushed me to properly re-measure instead of relying on a single data point.
> 
> > Could you share the generated speech-token count, or resulting mel length, used for the isolated 309-326 ms -> 174-190 ms benchmark?
> 
> The original measurement used **~125 synthetic speech tokens (~5s at 25Hz)**. I recovered the exact original validation script and reran it verbatim against this PR's code — it reproduces the headline number exactly: fp32=306.4ms → bf16=174.6ms (1.76x), matching the "309-326ms → 174-190ms" range in the PR description, with raw-waveform SNR=-0.90dB/corr=0.349 matching to two decimal places.
> 
> I then ran a full length sweep on B300 (25-800 tokens, 10 warmup + 50 measured iters/point, `float32_matmul_precision=highest / matmul_allow_tf32=False / cudnn_allow_tf32=True` — same settings you used; verified with a direct read of these flags from inside `flow.inference()`/`hift.inference()` at call time, so they're confirmed to actually be in effect, not just set and silently overridden):
> 
> tokens	FP32 p50	BF16 p50	BF16 vs FP32
> 25	203.5ms	139.0ms	1.46x
> 50	280.9ms	166.9ms	1.68x
> 75	287.8ms	158.7ms	1.81x
> 100	279.3ms	158.6ms	1.76x
> 125	307.1ms	174.6ms	1.76x
> 150	314.6ms	186.9ms	1.68x
> 175	318.8ms	175.9ms	1.81x
> 200	424.5ms	202.2ms	2.10x
> 250	440.2ms	211.6ms	2.08x
> 300	460.4ms	222.2ms	2.07x
> 400	564.0ms	248.0ms	2.27x
> 500	720.1ms	321.6ms	2.24x
> 800	1077.5ms	421.8ms	2.55x
> **No crossover on B300** across the whole range — bf16 is faster at every point tested, with the margin generally growing with length (some noise around 100-175 tokens, but never below 1.46x).
> 
> Quality across the sweep: mel SNR 28-34dB (stable), waveform spectral-magnitude SNR noisier (10-16dB, tied to specific synthetic content rather than token length), no NaN/Inf in any run.

let me try to get to the bottom of this. One observation so far:
B300's FP32 baseline is rather slow. For instance, at 300 tokens:
* H200: 326.5 → 222.4 ms = 1.47x
* B300: 460.4 → 222.2 ms = 2.07x

YijunWang1121:
@charliechenye Ran `torch.profiler` on B300 to see which kernels are actually selected.

The main finding is that under the default `float32_matmul_precision="highest"`, the three dominant Flow (DiT) ops are all on non-Tensor-Core paths: GEMM uses `cutlass3x_sm100_simt_sgemm_f32_...`, conv uses `implicit_convolve_sgemm<float,float,...>`, and attention uses `fmha_cutlassF_f32_..._sm80`. Together, GEMM + conv + attention account for **80.7% of Flow's GPU time** at 300 tokens.

For now I don't have a clear explanation for why B300 diverges from H200 here. I'm not sure whether this would help, but if you get a chance to profile it on H200, we could compare the kernel selections.

Also, would you be able to share the exact script you used for the H200 numbers? I want to make sure we're comparing apples to apples (same PyTorch/CUDA/cuDNN versions, warmup/sync methodology, etc.) before reading too much into the B300-vs-H200 gap. If it's just a script, I'm happy to run it as-is on my B300 box so we're at least using the same measurement code on both sides.

Here's the per-kernel breakdown at 300 tokens, fp32/highest (10 iters, `record_shapes=True` — profiler overhead inflates the absolute ms somewhat vs. the p50 numbers above, so treat these as relative shares, not literal wall-clock):

### Flow (DiT)

| kernel | ms/iter | share |
|---|---:|---:|
| `cutlass3x_sm100_simt_sgemm_f32_...` (GEMM, SIMT) | 228.0 | 49.9% |
| `implicit_convolve_sgemm<float,float,1024,...>` (conv, SIMT) | 87.6 | 19.2% |
| `fmha_cutlassF_f32_..._sm80` (attention) | 53.1 | 11.6% |
| `cublasLt::epilogue...globalKernel` (GEMM bias-add) | 14.1 | 3.1% |
| elementwise kernels ×3 (mul/silu/dropout) | 32.7 | 7.2% |
| `CatArrayBatchedCopy` ×2 | 10.8 | 2.4% |
| `vectorized_layer_norm` | 4.7 | 1.0% |
| `gemmSN_TN_kernel` | 4.5 | 1.0% |
| small elementwise (cos/sin/gelu/neg/where) | 15.2 | 3.3% |
| remainder (memcpy/small GEMM/reduce) | 1.9 | 0.4% |

### HiFT (vocoder)

| kernel | ms/iter | share |
|---|---:|---:|
| `precomputed_convolve_dgemm<128,6,7,3,3,5,...>` | 23.3 | 42.9% |
| `precomputed_convolve_sgemm<float,1024,5,5,...>` | 13.1 | 24.0% |
| `implicit_convolve_sgemm<float,float,128,5,5,...>` | 6.3 | 11.6% |
| elementwise (mul) | 1.9 | 3.5% |
| `implicit_convolve_dgemm<128,6,7,3,3,5,1,...>` | 1.2 | 2.3% |
| `CatArrayBatchedCopy` | 1.0 | 1.8% |
| `sm80_xmma_fprop_implicit_gemm_tf32f32...` | 0.75 | 1.4% |
| `cutlass3x_sm100_tensorop_s64x64x8tf32implicit_gemm...` | 0.68 | 1.3% |
| weight_norm / memcpy / NCHW↔NHWC layout conversion | ~2.0 | ~3.6% |
| remainder (elu/upsample/cumsum/etc.) | ~1.3 | ~2.4% |

PoTaTo-Mika:
lgtm, I've tested this pull request and seems no problem to merge.
I will further review the code style and then add CI.

sglang_omni/models/fun_cosyvoice3/stages.py
```diff
@@ -158,8 +163,12 @@ def _token2wav(
             )
         device = next(self._flow.parameters()).device
 
+        # Mirrors upstream CosyVoice3Model.token2wav, which wraps both the flow
```
PoTaTo-Mika: These notes seems unnecessary?

sglang_omni/models/fun_cosyvoice3/stages.py
```diff
@@ -158,8 +163,12 @@ def _token2wav(
...
+        # and hift calls in one autocast scope -- see cosyvoice/cli/model.py.
         with torch.autocast(
-            device_type=current_platform.device_type, enabled=self._fp16
+            device_type=current_platform.device_type,
+            dtype=self._compute_dtype or torch.float16,
```
PoTaTo-Mika: `or torch.float16` seems unused as `compute_dtype` will never be None as `enbaled=True`.

sglang_omni/models/fun_cosyvoice3/stages.py
```diff
@@ -214,13 +223,14 @@ def create_vocoder_executor(
 ) -> SimpleScheduler:
     device = resolve_device_spec(device, gpu_id)
     checkpoint_dir = resolve_checkpoint(model_path)
+    compute_dtype = _AUTOCAST_DTYPES.get(dtype)
```
PoTaTo-Mika: If this meet an unknown type, will fall back to fp32 without a warning logging.

charliechenye:
@YijunWang1121 I profiled the H200 side at the same 300 token point. There is a pretty clear kernel selection difference from your B300 profile.

Environment:

- GPU: NVIDIA H200, compute capability 9.0
- Driver: 595.71.05
- PyTorch: 2.11.0+cu130
- CUDA: 13.0
- cuDNN: 9.19.0
- `float32_matmul_precision="highest"`
- `torch.backends.cuda.matmul.allow_tf32=False`
- `torch.backends.cudnn.allow_tf32=True`
- SGLang-Omni: `c1146aeea2ebc638d0f869d681e9ef6bb6cd5285`
- CosyVoice: `074ca6dc9e80a2f424f1f74b48bdd7d3fea531cc`
- Matcha-TTS: `dd9105b34bf2be2230f4aa1e4769fb586a3c824e`

For this run, 300 generated speech tokens correspond to:

- 194 prompt mel frames
- 794 internal DiT frames after the 2x token-to-mel expansion
- attention shape `[2, 16, 794, 64]`

I used 5 warmup iterations followed by 3 profiled iterations with `record_shapes=True`. Each precision variant is fresh loaded and profiled separately so there is only one Flow model resident at a time. As with your profile, profiler overhead inflates the absolute time, so the numbers below should be read as relative kernel shares rather than benchmark wall clock latency.

### Flow (DiT), H200, 300 tokens, FP32/highest

| kernel | ms/iter | share |
|---|---:|---:|
| `sm80_xmma_gemm_f32f32_f32f32_f32_tn_...` (GEMM) | 164.2 | 56.7% |
| `fmha_cutlassF_f32_..._sm80` (attention) | 60.9 | 21.1% |
| `cutlass_tensorop_s1688fprop_optimized_tf32_...` (conv) | 14.9 | 5.2% |
| elementwise kernels x3 | 22.8 | 7.9% |
| `CatArrayBatchedCopy` x2 | 5.7 | 2.0% |
| `vectorized_layer_norm` | 3.1 | 1.1% |
| `gemmSN_TN_kernel` | 2.7 | 0.9% |
| NCHW/NHWC layout conversion | 1.5 | 0.5% |
| small elementwise kernels | 4.7 | 1.6% |
| remainder | ~8.8 | ~3.0% |

So GEMM + conv + attention account for about 83.0% of Flow GPU time on H200, similar to the 80.7% concentration you saw on B300, but the selected kernels differ substantially.

The most striking difference to me is the convolution path:

- H200: `cutlass_tensorop_s1688fprop_optimized_tf32_...`
- B300: `implicit_convolve_sgemm<float,float,1024,...>`

With the same `cudnn_allow_tf32=True` setting, H200 is selecting a TF32 tensor op convolution kernel, while your B300 profile selected the FP32 SGEMM path. In the profiler, this convolution costs about 14.9 ms/iter on H200 vs. 87.6 ms/iter in your B300 run.

The main GEMM selection is also different:

- H200: `sm80_xmma_gemm_f32f32_f32f32_f32_tn_...`
- B300: `cutlass3x_sm100_simt_sgemm_f32_...`

At 300 tokens that is about 164.2 ms/iter on H200 vs. 228.0 ms/iter on B300.

Attention is much closer: about 60.9 ms/iter on H200 vs. 53.1 ms/iter on B300.

Those three components alone appear to explain a large fraction of why the B300 FP32 baseline is slower, although I don't want to over interpret that until we run exactly the same input/script on both machines.

The exact H200 benchmark code is here:

https://github.com/charliechenye/sglang-scripts/tree/agent/cosyvoice3-p0-wave1/cosyvoice3

For the length sweep I used:

`p0a4_precision_length_sweep.py`

For this fresh load profiler I used:

`p0a8_flow_fresh_profile.py`

Happy for you to run the same branch as-is on B300. That should also align the prompt conditioning, generated token contents, synchronization, warmup methodology, and internal sequence shapes, rather than comparing only on generated-token count.

sglang_omni/models/fun_cosyvoice3/stages.py
YijunWang1121: Yes. I commented for reference. Deleted.

sglang_omni/models/fun_cosyvoice3/stages.py
YijunWang1121: Now raises instead of silently falling back.

sglang_omni/models/fun_cosyvoice3/stages.py
YijunWang1121: Removed dead code.

YijunWang1121:
> The exact H200 benchmark code is here:
> 
> https://github.com/charliechenye/sglang-scripts/tree/agent/cosyvoice3-p0-wave1/cosyvoice3
> 
> For the length sweep I used:
> 
> `p0a4_precision_length_sweep.py`
> 
> For this fresh load profiler I used:
> 
> `p0a8_flow_fresh_profile.py`
> 
> Happy for you to run the same branch as-is on B300. That should also align the prompt conditioning, generated token contents, synchronization, warmup methodology, and internal sequence shapes, rather than comparing only on generated-token count.

@charliechenye Looks like your repo is private to me.:)

YijunWang1121:
@PoTaTo-Mika Addressed the review comments. Could you please take another look when you have a chance? Thanks!

PoTaTo-Mika (approved):
### EN
| Concurrency | PR1663 audio thr. (s/s) | PR1715 audio thr. (s/s) | Δ | PR1663 QPS | PR1715 QPS | Δ | PR1663 latency mean (s) | PR1715 latency mean (s) | Δ | PR1663 RTF | PR1715 RTF | Δ |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| 1 | 5.82 | 5.86 | +0.7% | 1.20 | 1.178 | -1.8% | 0.835 | 0.849 | +1.7% | 0.1840 | 0.1873 | +1.8% |
| 2 | 9.01 | 9.19 | +2.0% | 1.86 | 1.872 | +0.6% | 1.072 | 1.068 | -0.4% | 0.2379 | 0.2367 | -0.5% |
| 4 | 12.73 | 13.16 | +3.4% | 2.61 | 2.595 | -0.6% | 1.532 | 1.540 | +0.5% | 0.3399 | 0.3407 | +0.2% |
| 8 | 13.88 | 13.54 | -2.4% | 2.81 | 2.880 | +2.5% | 2.847 | 2.772 | -2.6% | 0.6393 | 0.6382 | -0.2% |
| 16 | 14.28 | 14.60 | +2.2% | 2.87 | 2.997 | +4.4% | 5.555 | 5.321 | -4.2% | 1.2650 | 1.2187 | -3.7% |

### ZH

| Concurrency | main audio thr. (s/s) | PR1715 audio thr. (s/s) | Δ | main QPS | PR1715 QPS | Δ | main latency mean (s) | PR1715 latency mean (s) | Δ | main RTF | PR1715 RTF | Δ |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| 1 | 5.84 | 5.86 | +0.3% | 1.23 | 1.178 | -4.2% | 0.816 | 0.849 | +4.0% | 0.1817 | 0.1873 | +3.1% |
| 2 | 9.09 | 9.19 | +1.1% | 1.89 | 1.872 | -1.0% | 1.057 | 1.068 | +1.0% | 0.2328 | 0.2367 | +1.7% |
| 4 | 12.56 | 13.16 | +4.8% | 2.54 | 2.595 | +2.2% | 1.576 | 1.540 | -2.3% | 0.3465 | 0.3407 | -1.7% |
| 8 | 13.37 | 13.54 | +1.3% | 2.68 | 2.880 | +7.5% | 2.978 | 2.772 | -6.9% | 0.6657 | 0.6382 | -4.1% |
| 16 | 13.53 | 14.60 | +7.9% | 2.78 | 2.997 | +7.8% | 5.747 | 5.321 | -7.4% | 1.3105 | 1.2187 | -7.0% |

PoTaTo-Mika:
Done, will merge after CI passes.

SandyLuXY (approved):
lgtm

PR #1716, opened 2026-08-25
docs(pd): a user guide for prefill/decode disaggregation

jzheng17:
Adds the user-facing page for prefill/decode disaggregation, plus a two-GPU example config.

## Why it is separate

It documents the surface added in #1684 and does not depend on it to review. Splitting it keeps that PR to the mechanism and keeps this one to prose someone can read end to end. **Merge after #1684**, since the flag and the config fields the page describes come from there. This branch carries only the three doc files, so the diff shows nothing else.

## What the page says

`docs/basic_usage/` has pages for the other placement features, `mps_dp.md`, `tts_process_topology.md` and `process_topology_migration.md`, and none for PD, so a deployment meets `--pd-stage` with no guidance on when the split helps, what it costs, or how to size the halves.

The page leads with the decision rather than the flag, because for some workloads the answer is not to split.

**The split is three continuous choices, not one topology.** Which devices each half gets, each half's share of a device, and each half's own runtime settings are set separately. Both halves on one GPU already separates them: what PD needs is the process split, and a prefill step leaves the decode scheduler thread whichever card it runs on. A second card is an additional, separable decision about hardware.

**Prefill share sets what a second card can be worth.** It is prompt length divided by output length: about 2.6% at 42/128, about 9% at 6944/128, about 42% at 42/8. Splitting across two cards divides the hardware evenly, so against a 2.6% share the prefill card measured 10.1% mean utilization. Long prompts with short outputs earn a second card; short prompts with long outputs do not, and the same-GPU form separates the halves without one.

**Throughput and inter-token stability select different configurations.** A colocated replica pays 5.9x to 8.5x on the inter-token gap that overlaps a prefill on text and 8.4x to 20.5x on images; PD brings both to about 1.0x. For a stream with a repeating deadline that jitter is the SLO. For throughput, one half against one half on two cards has parity as its ceiling: a card doing both jobs has the harmonic mean of its prefill-only and decode-only capacities, a 1:1 pair has the minimum of the two, and the minimum never exceeds the harmonic mean. The configurations whose arithmetic differs are ratios other than 1:1, which are not configurable today, and both halves on one card. Neither is measured yet, and the page says so.

**Overload has three settings and they answer different questions.** `SGLANG_REQ_WAITING_TIMEOUT` bounds how long one request waits, `--max-queued-requests` bounds how many wait, and `--enable-priority-scheduling` decides which ones survive the bound. The page recommends the timeout as the primary bound for interactive serving, because a queue length converts to a wait only through the current service rate while a duration is the caller's own timeout, and keeps the length as the memory bound, which a duration cannot provide.

**The prefill card carries more than the flag moves.** `--pd-stage` moves only the named stage, so the encoders stay where they were and the CUDA-IPC relay pool appears on the prefill card. A thinker budget copied from a colocated profile runs it out of memory during startup.

Every number above comes from the equal-GPU-count comparison in #1696, which the page links, including the ones that argue against splitting.

## The example config

The colocated profiles cannot serve as a PD example. `Qwen3OmniSpeechColocatedPipelineConfig` requires the image encoder, audio encoder, thinker, talker and code2wav to share one GPU, which a split thinker contradicts, and it reports that as a request for exactly one GPU id for the thinker. `examples/configs/qwen3_omni_pd.yaml` is a text-output two-GPU profile with the thinker budget lowered to 0.68, because 0.75 runs the prefill half out of memory once the encoders and the relay pool are on that card.

## Testing

`--pd-stage thinker=0:1` with this config was run end to end on two H200s with #1684 and #1599 both applied: the two halves start as separate stage processes and a request returns 80 decoded tokens through the pair.

Related: #1684, #1599, #1696, #1709.

### DIFF
--- a/docs/basic_usage/pd_disaggregation.md
+++ b/docs/basic_usage/pd_disaggregation.md
@@ -1,18 +1,32 @@
 # Prefill/Decode Disaggregation
 
-> TL;DR: PD disaggregation splits one stage into a prefill process and a decode process so a prefill no longer stalls decoding. It removes that interference almost completely. It does not raise throughput at equal GPU count on Qwen3-Omni thinker: in the measurements below two colocated replicas beat one prefill/decode pair on both text and image. Turn it on when inter-token latency stability is what you need, not to serve more requests per GPU.
+> TL;DR: PD disaggregation splits one stage into a prefill process and a decode process, so a prefill no longer stalls the decoding already in flight. It removes that interference almost completely, on any placement, including both halves on one GPU. How much hardware you then give each half is a separate, continuous choice, and it is the choice that decides what the split costs and what it can return. One prefill half against one decode half on two cards is the configuration measured here: it has parity with two colocated replicas as its throughput ceiling, and on Qwen3-Omni thinker it currently lands below that.
 
 A colocated replica runs prefill and decode on one card and one scheduler thread. A prefill therefore blocks the decode steps of every request already running. On Qwen3-Omni that cost is large, and it is worst on images, where a 6127-token prompt cannot be chunked.
 
-PD disaggregation moves prefill into its own process. The prompt KV and the request's continuation state transfer once, over CUDA IPC, and decoding continues on the other half.
+PD disaggregation moves prefill into its own process. The prompt KV and the request's continuation state transfer once, over CUDA IPC, and decoding continues on the other half. Where that other half runs, and how much of a card it gets, is what the rest of this page is about.
+
+## What the split actually gives you
+
+PD here is not one topology. It is a resource split you set continuously, at three layers, and each layer answers a different question.
+
+| Layer | How you set it | What it decides |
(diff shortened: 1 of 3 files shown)


jzheng17:
Rewritten as #1805 against `feat/minimal-pd-disaggregation`.

Everything in this version was written against the CLI syntax (`thinker=0@0.25:0@0.65`) and `pd_disaggregation.prefill.server_args`, both of which changed. The new config also brought knobs this could not describe: `memory_fraction` is now required when the halves share a card, `share_weights` is off by default, and `max_inflight_handoffs` exists.

Two of its conclusions also needed correcting rather than porting. It said PD completes fewer requests per second than colocated; with CUDA graph on for both arms, a two-card pair does 5.58 against colocated's 4.55, and same-GPU PD crosses over around 1024 prompt tokens. The new version leads with that setting, because it decides what any measurement of PD shows.

PR #1717, opened 2026-08-25
feat(pd): share one copy of the weights when both halves sit on one GPU

jzheng17:
Two PD halves on one GPU load two copies of the same weights. Measured on the Qwen3-Omni thinker that is 57 GiB each, which leaves a 140 GiB card about 21,500 KV tokens per half against 677,613 on a colocated card. This shares one copy.

## Why it is worth doing

Same-GPU placement is the form that separates prefill from decode without spending a second card, and it is what makes PD runnable on a one-GPU box and in CI. The duplicate weights are most of what it costs. The halves already map each other's GPU memory for the KV plane; weights are the easier case, since they are static, read-only, allocated once and never reclaimed until shutdown, so none of the reserve/commit/abort machinery applies.

Peak memory is unchanged, because the adopting half still constructs and loads before it swaps. The KV pools are sized after that, so they see the freed space, which is the whole point.

## The pieces

**The mechanism.** `export_parameter_handles` returns one CUDA IPC handle per parameter; `adopt_parameter_handles` points this model's parameters at that storage and releases what it loaded. A parameter that does not match by name, shape and dtype raises rather than being skipped: a silently unshared parameter costs memory without saying so, and a wrongly shared one is worse.

**The channel.** The halves are separate processes with no channel at load time, and the KV plane does not supply one, since `prepare_kv_receive` and `send_kv_pages` are per-transfer and run long after both halves are up. `weight_rendezvous` uses the directory the run already owns: `create_ipc_runtime_dir` makes one private directory per pipeline instance before any stage is spawned, `allocate_endpoints` puts every stage socket directly inside it, and it is removed with the run. A half derives the directory from an endpoint it already holds, so nothing new is threaded through stage startup. Publishing writes a staging name and renames, so a reader never sees a partial file.

**No half waits.** `_construct_scheduler` builds a stage inside `gpu_startup_lock(gpu_id)`, so two halves on one device load one at a time. Assigning the exporting role in advance would deadlock whenever the assigned adopter won that lock: it would hold the lock while waiting for a half that needs the same lock to load at all. Instead the roles follow the lock. The half that loads first finds nothing published and publishes; the half that loads second finds that file and adopts.

**Cross-GPU halves decline rather than compare.** A CUDA IPC handle names memory on a particular device, so the publisher records which GPU it wrote for and a half on another card declines the handles and keeps its own. Nothing has to compare placements.

**Sharing can be turned off.** `pd_disaggregation.share_weights` defaults to true, because two copies of the same static tensors buy nothing. It still needs an off switch: this swaps parameter storage across a process boundary, so a deployment that suspects it needs a way back to each half loading its own, and measuring what sharing saves means running both ways.

**Opting in is a declaration.** `_construct_scheduler` builds the plan from the spec, which already carries the stage name, the GPU, the PD role and partner, and the stage's own endpoint. `resolve_factory_signature_args` injects a default only when the factory names the parameter, so a stage opts in by taking `weight_sharing_plan`. The Qwen3-Omni thinker declares it and passes it to `create_sglang_infrastructure`, which applies it between loading the weights and `alloc_memory_pool`. That is the only useful point: sharing after the pool is sized returns memory nothing will claim.

## Measured on hardware

One H200, both halves on it, the two runs differing only in `share_weights`, back to back on the same card.

| `thinker=0@0.47:0@0.47` | prefill half | decode half | total |
| --- | --- | --- | --- |
| sharing off | 22,206 | 22,206 | 44,412 |
| sharing on | 22,206 | **643,390** | **665,596** |

Fifteen times the KV capacity from the same fractions. For scale, one colocated replica on one card holds 677,613 tokens, so the shared pair reaches 98% of that while running prefill and decode in separate processes.

The gain is asymmetric, and the numbers say why. A single copy of the weights is 56.94 GiB on a 140.40 GiB card. The half that loads first publishes and keeps its own copy, so its budget still pays for them: 0.47 x 140.40 = 66.0 GiB, minus 56.94, leaves 9.1 GiB and 22,206 tokens. The half that loads second adopts and releases what it loaded before its pool is sized, so all 66.0 GiB goes to KV.

Some configurations cannot start at all without it. At `thinker=0@0.30:0@0.62` the unshared run is rejected by this repo's own budget check — `budget=35.10GiB, process_used=56.94GiB` — while the shared run starts with 252,239 and 382,685 tokens.

**Which half publishes is decided from the declared shares.** The publisher keeps the copy it loaded, so its budget has to hold the weights on top of its own KV, and the half with the larger share is the one that can afford that. Equal shares are a tie and prefill takes it.

Letting the startup lock decide instead was a defect, and it showed up on hardware: at `thinker=0@0.30:0@0.62` one run came up with 252,239 and 382,685 tokens and the next was rejected with `budget=35.10GiB, process_used=56.94GiB`. The difference was which half won the lock. The 0.30 half cannot keep a 56.94 GiB copy, so a placement that starts one time and fails the next is not a placement anyone can deploy.

The adopter now waits for the handles **before** taking `gpu_startup_lock`. Waiting inside it would hold the lock against the half being waited for, which is what the earlier non-blocking read avoided; moving the wait out removes that deadlock and the dependence on load order together. A peer that never publishes still lets the adopter load its own weights.

## The shares have to leave room for one more copy

Adopting does not lower peak memory, and on one card that is a constraint rather than a footnote. The adopting half still loads the weights before it can point at the peer's, and by then the publisher is holding its own copy plus its KV pool. So:

    publishing share x total + one copy of the weights <= total

On an H200 a copy is 56.94 GiB of 140.40 GiB, which caps the publishing share at about 0.594.

| Placement | publisher holds | peak while the adopter loads | Result |
| --- | --- | --- | --- |
| `0@0.47:0@0.47` | 66.0 GiB | 122.9 GiB | starts |
| `0@0.30:0@0.62` | 87.0 GiB | 143.9 GiB | out of memory |

Fixing the roles is what made this visible. While the lock decided them, `0@0.30:0@0.62` sometimes had the small half load first and came up; now it consistently publishes from the 0.62 half and consistently hits the ceiling. The race is gone and the real constraint is what remains.

The publisher records its parameter bytes with the handles, and the adopter checks free memory against that inside the startup lock, right before it builds. The message names the fix — lower the peer's share so the card keeps one spare copy free — rather than surfacing a CUDA out-of-memory several frames into the loader.

Sizing follows from the same fact: give the publishing half room for the weights plus its own KV, and the adopting half only its own KV.

## Testing

1,165 tests pass across `tests/unit_test/pipeline` and `tests/unit_test/qwen3_omni`.

The new tests pin what the exchange depends on: the run directory is derivable from the endpoint a stage already has, a reader never observes a partial file, an unpublished peer reads as absent rather than blocking, handles published for another GPU are declined, the half that loads second is the one that adopts, and both the thinker factory and `create_sglang_infrastructure` declare the parameter.

The memory saving itself is not measured here. It needs two halves on one card under the same-GPU placement, which #1684 adds.

## Not included

Reclaiming the adopting half's copy before it loads, which would lower peak memory as well. That needs the weights to never be materialized, not swapped after the fact.

Related: #1684, #1599, #1596.


MelodyyyYin:
#note:（yue yin）

Weight adoption validates only parameter names, then mutates parameters one by one; same-name shape/dtype/layout mismatches can be installed and a later failure leaves a partially adopted model. The adopter also waits before confirming factory capability, and rendezvous has no publisher-generation binding.

Follow-up PR: https://github.com/jzheng17/sglang-omni/pull/6

Why: CUDA-IPC adoption must validate one live generation’s full manifest before an atomic swap, so publisher loss or incompatible storage cannot expose a partial or stale model.

jzheng17:
Redone as #1801 against `feat/minimal-pd-disaggregation`, which is where the main line is now. `pd_runtime` predates it.

Two things changed on the way across. `share_weights` is off by default rather than on: it needs the model's factory to accept `weight_sharing_plan`, and after the base was made model-neutral no factory declares it, so an on-by-default would have made every same-GPU config raise. And the publisher decision now reads `PDStagePlacement.memory_fraction`, which #1784 added — this branch had to carry its own field for that.

The measurements stand: 56.94 GiB returned to the pool on an H200, and the prefill 0.30 / decode 0.62 case where the run that published from the smaller half failed while the run that adopted came up.

PR #1718, opened 2026-08-25
feat(pd): bound how many handoffs hold Prefill KV at once

jzheng17:
One prefill/decode handoff holds that request's prompt KV on the Prefill card until Decode acknowledges it. Nothing bounded how many did so at once, so the count landed on Prefill's `max_running_requests`, a setting chosen to size batches rather than to bound leases.

## Measured

Two H200s, Qwen3-Omni thinker, `--pd-stage thinker=0:1`, 344,253 KV tokens per half. A high-water counter on `_launch_pd_handoff`:

| Client concurrency | Peak handoffs in flight |
| --- | --- |
| 64 | 27 |
| 256 | 64 |

At 256 the peak stops exactly at `max_running_requests=64` and does not grow further, so a bound does exist. It just is not a bound anyone set for this.

## Why that coupling matters

`SGLangKVPageLease` keeps Prefill KV alive until the receiver's ACK releases the source. Sixty-four leases at a 6127-token image prompt are 392,128 tokens against that run's 344,253-token pool, so on an image workload the running limit permits more lease than the pool holds. What actually stops it is the pool running out, which is not a designed admission point.

The two cannot be moved independently today: raising `max_running_requests` to widen batches also raises how much of the Prefill pool sits in leases, and lowering it to bound leases also narrows batches.

## The change

`pd_disaggregation.max_inflight_handoffs` separates them. Unset keeps today's behaviour rather than choosing a number, since the right value depends on prompt length and pool size. It is a property of the pair, so the graph rewrite copies it to both halves.

The wait happens inside the handoff task, not before it is created, so the stage keeps draining scheduler output for unrelated requests while a handoff queues for a slot. That preserves the reason `_launch_pd_handoff` was made asynchronous.

## What this does not do

It does not order the transfers. Handoffs still enter the outbox in `batch.reqs` order and take slots first-come. Priority and deadlines do not participate. Ordering is worth doing once requests carry a priority through the handoff, which #1709 adds.

Related: #1709, #1684, #1596.

### DIFF
--- a/sglang_omni/config/pd_rewrite.py
+++ b/sglang_omni/config/pd_rewrite.py
@@ -156,7 +156,11 @@ def _split_pd_stage(
             "stream_to": [inbound_rename.get(t, t) for t in s.stream_to],
             "pd_disaggregation": None,
             # Note (Yue Yin): Keep compiler metadata out of user factory kwargs.
-            "pd_execution": PDExecution(role="prefill", partner=decode_name),
+            "pd_execution": PDExecution(
+                role="prefill",
+                partner=decode_name,
+                max_inflight_handoffs=pd.max_inflight_handoffs,
+            ),
         },
     )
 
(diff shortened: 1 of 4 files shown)


jzheng17:
Folded into #1709. Both bound what accumulates on the PD path — one the Decode queue, the other the Prefill KV held by transfers in flight — so they read better as one change.

PR #1719, opened 2026-08-25
[Deps] Bump SGLang to 0.5.18

Ratish1:
## Motivation

Bump the pinned SGLang from 0.5.16 to 0.5.18 together with its stack: torch 2.13.0, torchvision 0.28.0, torchcodec 0.15.0, flashinfer 0.6.17, cutlass-dsl 4.6.2.

The change in 0.5.18 that reaches this repo is the config split. `ServerArgs` becomes a read-only startup record once the runtime context publishes it inside `ModelRunner.__init__`, `ServerArgs.override` is gone, and configuration after publish lives in the namespace bags (`get_exec()`, `get_model()`, `get_serving()`, `get_schedule()`, `get_parallel()`). Every place this repo mutated or re-read the record after publish now goes through the seam SGLang provides.

## Modifications

Pins and packaging
* `pyproject.toml`: sglang 0.5.18 and the torch stack.
* `docker/Dockerfile`: base image digest, flashinfer 0.6.17, jit cache path, `sox` and `qwen-tts==0.1.1 --no-deps` (its pins conflict with the stack, same as `xpu.Dockerfile`).
* `.github/workflows/*.yaml` and the Dockerfile cache stage point at the rebuilt image `hongccc/sglang-omni:dev` (`sha256:02a85f00438c901c72a2eb2ef738974a807f63af3d13084445604f3344067b19`).
* `.github/workflows/test-qwen3-omni-ci.yaml` sets `NCCL_NVLS_ENABLE=0` (see NCCL under Torch 2.13 consequences).

Import surface
* MOSS-TTS flash attention imports from `sglang.kernels.ops.attention` (`sglang.jit_kernel` is gone; the old import sat behind `except ImportError`, so it would have silently fallen back to SDPA).
* `SamplingBatchInfo` no longer takes `vocab_mask` and `apply_mask_func`.

Scheduler composition
* `SchedulerDPAttnAdapter`, `SchedulerLoadInquirer`, `SchedulerLogprobResultProcessor` take their 0.5.18 kwargs directly; the `supported_kwargs` signature filter is deleted.
* `ParallelState` is built with `attn_dcp_rank` and `attn_dcp_size`; the `create_parallel_state` shim is deleted.
* `NewTokenRatioTracker.from_config()` replaces `from_server_args`.
* Pool stats sizing and the `pp_max_micro_batch_size` default go through `get_parallel()` and `get_context().override`.

Config split
* `override_server_args` is two-phase: `declare_late_resolution` before publish, `get_context().override` after.
* Deferred CUDA graph capture no longer flips `disable_cuda_graph` on the record and restores it. Deferral is the `defer_cuda_graph_capture` argument, and `init_cuda_graphs` re-seeds the capture flag from `get_exec().graph`.
* The weight-share follower loads dummy weights through `_resolve_draft_load_format` instead of mutating `load_format`.
* `model_info`, `update_weights_from_disk` and the admin info read the `model` and `serving` bags.
* `get_attention_tp_rank` / `get_attention_tp_size` aliases are replaced by `get_parallel().attn_tp_rank` / `attn_tp_size`.
* `startup_weight_load_mode="overlap"` is refused in `build_sglang_server_args`. That mode returns from `load_model` with staged sentinel weights and expects `finalize_startup_weight_load` after capture; this repo's bootstrap owns that sequence and does not commit the weights.
* `create_sglang_infrastructure` refuses to build an engine in a process that already holds a published runtime context: 0.5.18 has one authoritative context per process, so a second engine must live in its own process. A constructor that published and then failed is refused on retry too.
* `SGLModelRunner._extend_forward_kwargs` requires the prefill sidecar to be in the model dtype. The graph slot is filled with `Tensor.copy_`, which would otherwise convert silently.

Platforms
* XPU resolves to `XPUOmniPlatform`; the resolver previously fell through to the CUDA platform on XPU hosts.
* **Hardware providers (ROCm, XPU, NPU, MUSA) upgrade their own SGLang pin and stack themselves. That work is out of scope for this PR, which validates CUDA only.**

Torch 2.13 consequences
* Qwen3-Omni audio layer CUDA graphs: the captured layer stack ran attention through transformers' `flash_attention_2`, which the kernels hub does not build for torch 2.13/cu13, so the default graph path silently fell back to eager. `AudioLayerGraphRunner` now runs the packed attention through SGLang's own `VisionFlash3Attention` on Hopper and `VisionTritonAttention` on other CUDA devices, the classes SGLang uses for the same audio tower. The eager per-segment SDPA path is unchanged and remains the fallback. A kernel that cannot be resolved or captured leaves the encoder eager with a logged reason.
* Dots-TTS: `dots.tts` refuses to import unless torch and torchaudio share a minor version; torchaudio 2.11.0 is its last release and declares compatibility with later torch versions. `sglang_omni.models.dots_tts.compat.import_dots_tts()` reports the torchaudio distribution at torch's version for the single package import and restores the reader. Every omni import of `dots_tts` goes through it. Removal condition is in its docstring.
* NCCL: torch 2.13 pins `nvidia-nccl-cu13==2.29.7` (torch 2.11 brought 2.28.9). NCCL 2.29 makes a failed NVLink SHARP (NVLS) multicast bind fatal at communicator creation, where 2.28 logged it and fell back to plain NVLink transport (NVIDIA/nccl#2077, intentional per NVIDIA). On hosts where NVLS does not bind, which includes the CI runners and the H100 dev host (the bind fails with CUDA error 401 under both NCCL versions), every TP>=2 engine now dies in `ncclCommInitRank` unless `NCCL_NVLS_ENABLE=0` is set. TP=1 never creates a communicator and is unaffected. The Qwen3-Omni workflow sets the variable; deployments running TP>=2 on such hosts need it too, and lose nothing, since NVLS was never active there.
* Upstream warmup dummy runs and the flashinfer autotune context no longer run under `torch.inference_mode`; graph capture never did, so the effect is slightly higher warmup-time memory, nothing at serving time.
* Upstream dropped its clamp of the auto-derived prefill CUDA graph `max_bs` to `context_length`. This repo's builders clamp through `clamp_prefill_cuda_graph_max_bs()` or pass small constants, so nothing changes here, but new engine builders must keep routing through that clamp.
* Seeded Gumbel sampling: this repo carries three copies of SGLang's `multinomial_with_seed` transform (MOSS-TTS Triton argmax and token-id helper, Qwen3-TTS small-k Triton sampler). 0.5.18 clamps `log(u)` at `-(2**-32)` so a `UINT32_MAX` hash cannot produce `+inf` noise; the copies apply the same clamp and tests pin the exact boundary for each.

Removed
* `PrefillManager` and `DecodeManager` (`sglang_backend/prefill.py`, `decode.py`): copies of the 0.5.12 admission path, constructed on every bootstrap and read by nothing since #937.
* `hasattr` probes on this repo's own `ModelWorker`, the `runner_args is not server_args` branches, `FakeServerArgs` in the test fakes.

Tests
* New: the vendored RMSNorm patch under `BaseFusedOp` dispatch, the XPU free-memory cap against the allocator, `override_server_args` on real unpublished and published contexts, the capture-flag re-seed from the exec bag, the overlap refusal.
* Adapted: the Qwen3-ASR mrope fast path sets `forward_batch.seq_lens` (decode mrope is vectorized in 0.5.18), strict `ServerArgs` doubles model the real write guard, tests that read the bags publish a context.
* New: audio backend rule and resolver failure, packed attention against fp32 math SDPA per segment and head plus the full layer stack against segmented bf16 SDPA, capture and replay across bucket boundaries; the Dots import shim on a forced 2.13/2.11 mismatch; `UINT32_MAX` Gumbel boundary for every fork; one engine per process; sidecar dtype.

Docs: versions in `installation.md`, `tts.md`, `qwen3_tts.md`.

### Inherited upstream behavior changes

These come with sglang 0.5.16 to 0.5.18 and are not introduced by this PR, but they are visible to users:

- Seeded sampling: `multinomial_with_seed` now clamps `log(u)` at `-(2**-32)`, so a request with a pinned seed can decode a different sequence than on 0.5.16 in the rare case a hash lands on `UINT32_MAX`. The previous value was a NaN that let a filtered-out token win the argmax. The MOSS-TTS and Qwen3-TTS Triton samplers mirror the clamp.
- `finish_reason`: stop conditions are now checked before `max_new_tokens`, so a stop token sampled exactly at the cap reports `stop` instead of `length`. Token content is unchanged.
- KV budget: upstream subtracts a 100 MiB multimodal runtime reservation from the profiled budget, so non-colocated multimodal engines get about 100 MiB less KV cache.

## Related Issues

None.

## Accuracy Test

H100, rebuilt image, PR head versus main on the 0.5.16 image on the same host, two runs each:
* SeedTTS-50 (bf16 colocated): 0/50 failures on every run; WER 1.06% / 1.95% (PR) versus 1.42% / 1.42% (main), n=50 noise; speaker similarity 3.39 / 4.60 versus 3.56 / 3.52.
* MMSU (2000 samples): 71.05% / 71.35% (PR) versus 71.15% / 71.30% (main), 0/2000 failures.
* Qwen3-Omni audio graphs, bf16 and FP8 checkpoints, SeedTTS-50 reference audio and MMSU audio: fa3 backend, all six buckets captured, every request replayed, cosine ≥ 0.999345 and max abs ≤ 0.045 against the eager path, median encoder speedup 2.0x.
* Full SeedTTS EN set (1088 samples): see the benchmark table; WER and speaker similarity move within run-to-run noise in both directions.
* Qwen3-Omni Video-AMME (CI stage 9, 50 samples, threshold 0.64): main scores 30/50 and 31/50 on the H100 hosts, this PR 31/50 and 30/50; three of the four samples that move do so within a stack, and the one stack-specific sample is answered correctly by this PR. The gate has no margin above main's own score and is being recalibrated separately. Stage 8's RTF gate fails on main's latest run as well.

## Benchmark & Profiling

Qwen3-TTS, one H100, `benchmarks/eval/benchmark_omni_seedtts.py` on the full SeedTTS EN set, PR head versus main at the merge base, each on its own image. Every number is from a warm server: the same server process runs one discarded full pass first (the first pass in a fresh image compiles the Inductor, Triton and FlashInfer caches for the new versions and is not representative), then the measured pass.

| concurrency | main (0.5.16 / torch 2.11) | this PR (0.5.18 / torch 2.13) |
|---|---|---|
| 16 | 13.532 req/s, 695.9 tok/s, WER 1.097%, similarity 71.401 | 13.748 req/s, 709.1 tok/s, WER 1.072%, similarity 71.352 |
| 1 | 2.090 req/s, 107.3 tok/s, WER 1.038%, similarity 71.389 | 2.153 req/s, 0.464 s mean latency (a first pass in which two requests looped to the 2048-token cap measured 1.955 req/s and was rerun) |

Per decode step in a torch profiler trace at concurrency 16, GPU time is equal between the two stacks (talker graph replay 3.49 ms versus 3.48 ms, eager kernels 1.04 ms versus 1.05 ms, sync waits 3.88 ms versus 3.91 ms); the FA3 kernels are the same source in both sgl-kernel wheels and time the same in isolation. The host-side difference was the Qwen3-TTS suppress-mask rebuild, which #1750 (merged into main and into this branch) removes.

Earlier 50-sample runs on the same host: PR 7.82 req/s, 1.92 s mean latency, RTF 0.579 versus main 7.77 req/s, 1.96 s, 0.592 (TTS); MMSU 83.0 versus 80.9 req/s. Upstream changes inherited live that touch admission and memory (shortfall-only tree-cache eviction, `mm_runtime_reservation_gb` in profiling, fp32 top-k renormalization; the prefill tile budget in `PrefillAdder` is HIP-only and does not run here) are covered by the same matrix. The two numerical changes in this repo are the Gumbel clamp and the audio graph kernel above, both measured.

## Checklist

- [x] Add unit tests.


Hayden727:
## H200 A/B performance validation

This is an A/B performance report from **2× NVIDIA H200**; it is not a CI result.

- Old image: `hongccc/sglang-omni@sha256:374d0b1c30b2bff685b1716fc64a02ad3b3d0a90fe2ce73ce9861a6992c28101` (`sglang==0.5.16`)
- New image: `hongccc/sglang-omni@sha256:02a85f00438c901c72a2eb2ef738974a807f63af3d13084445604f3344067b19` (`sglang==0.5.18`)
- Old/new runs were serialized on the same GPUs and never overlapped. TorchInductor caches were isolated. Deltas below are **new vs. old**.
- TTS presets used the full SeedTTS EN set (1,088 samples, concurrency 16). ASR presets used full SeedTTS EN/ZH (1,088/2,020 samples). Differences above 5% were repeated in reverse order where noted.

| Model / path | Old | New | Delta | A/B conclusion |
|---|---:|---:|---:|---|
| Higgs-TTS | QPS 19.683, RTF 0.1972 | QPS 19.150, RTF 0.1971 | QPS -2.7%, RTF flat | No regression |
| **MOSS-TTS (reverse-order repeat)** | QPS 18.004, RTF 0.2050 | QPS 16.420, RTF 0.2255 | **QPS -8.8%, RTF +10.0%** | **Confirmed regression** |
| Qwen3-TTS (current base/head rerun) | QPS 16.187, RTF 0.2434 | QPS 16.453, RTF 0.2404 | QPS +1.6%, RTF -1.2% | No regression |
| Dots-TTS | QPS 3.761, RTF 1.0287 | QPS 3.672, RTF 1.0538 | QPS -2.4%, RTF +2.4% | No regression |
| Qwen3-ASR EN (reverse-order repeat) | 250.485 req/s, mean 0.1258 s | 241.438 req/s, mean 0.1273 s | Throughput -3.6%, mean +1.2% | No stable regression |
| **Fun-ASR EN (reverse-order repeat)** | 190.314 req/s, mean 0.1660 s, P95 0.4421 s | 135.452 req/s, mean 0.2339 s, P95 1.0328 s | **Throughput -28.8%, mean +40.9%, P95 +133.6%** | **Confirmed regression** |
| **Fun-ASR ZH** | 268.915 req/s, mean 0.1178 s | 228.272 req/s, mean 0.1375 s | **Throughput -15.1%, mean +16.7%** | Corroborates Fun-ASR regression |
| Whisper EN/ZH serving | EN 106.137 req/s; ZH 100.042 req/s | EN 101.205 req/s; ZH 94.960 req/s | EN -4.6%; ZH -5.1% | Borderline; no reverse-order confirmation |
| MOSS-TD full suite | Movies QPS 48.167; stream QPS 52.249 | Movies QPS 48.051; stream QPS 52.168 | -0.2% / -0.2% | No regression |
| Qwen3-Omni TTS (reverse-order repeat) | QPS 10.397, RTF 0.4318 | QPS 10.066, RTF 0.4364 | QPS -3.2%, RTF +1.1% | No stable regression |
| Qwen3-Omni Thinker TP2 | 78.83 s, 3/3 passed | 79.50 s, 3/3 passed | +0.9% | No regression |
| Qwen3-Omni Video-AMME (reverse-order repeat) | QPS 1.865, mean 7.463 s | QPS 1.971, mean 6.450 s | QPS +5.7%, mean -13.6% | Faster; no regression |
| Qwen3-Omni Video-MME Talker | QPS 1.123, RTF 1.1561 | QPS 1.375, RTF 0.9779 | QPS +22.4%, RTF -15.4% | Faster; no regression |

Quality metrics remained stable for the two confirmed serving regressions: MOSS-TTS WER/SIM/UTMOS and Fun-ASR WER showed no material degradation. Whisper cold startup was also slower in one pair (`260 s -> 438 s`, +68%); I am treating that as an observation rather than a confirmed result because it was not repeated in reverse order.

**Conclusion:** the H200 A/B confirms performance regressions in **MOSS-TTS** and **Fun-ASR**. The dependency upgrade should be investigated or mitigated before merge; the other tested model paths did not show a stable serving regression.

yxs (approved):
LGTM

Hayden727 (approved):

Ratish1:
CI passed

Screenshot 2026-08-27 at 5 33 23 PM

PR #1720, opened 2026-08-25
[Qwen3-Omni] Disable autograd in the Talker code predictor

celestial-micha:
## Motivation

The Talker code predictor runs during inference, where gradients are not needed. However, `code_predictor_forward` can be called while gradient tracking is enabled, causing PyTorch to build autograd graphs that are not used by the serving path.

`Qwen3OmniTalker.forward` already uses `@torch.no_grad()`, but the decorator only covers operations executed inside that method. `QwenTalkerModelRunner.post_prefill` calls `code_predictor_forward` after `forward` returns, so this later call runs under the scheduler thread's current gradient mode.

The predictor writes to buffers that remain on the model between requests. If autograd tracks those writes, the buffers can retain references to the graph after a request finishes, so the graph's tensors are not released. In the eager-path profile, live tensor memory grew by about 170 MiB in each of two consecutive 50-request passes.

Since this entry point is inference-only, adding `@torch.no_grad()` directly to `code_predictor_forward` prevents the unnecessary graph from being created without changing the forward computation.

## Modifications

- Add `@torch.no_grad()` to `Qwen3OmniTalker.code_predictor_forward`.
- Add a regression test that calls the function from a grad-enabled context and verifies that no autograd graph is created.

```diff
 class Qwen3OmniTalker(nn.Module):
+    @torch.no_grad()
     def code_predictor_forward(
```

The focused regression test fails without the decorator and passes with it. The full Talker test file passes with `66 passed, 5 skipped`.

The product-code change is one line. It only disables gradient recording; it does not alter the predictor's forward computation or sampling.

## Related Issues

N/A

## Accuracy Test

`torch.no_grad()` changes whether PyTorch records operations for backward; it should not change the forward result. A fixed-seed paired generation run was used to verify this on generated audio.

WER was measured with `Qwen/Qwen3-ASR-1.7B` revision `7278e1e70fe206f11671096ffdd38061171dd6e5` on `zhaochenyang20/seed-tts-eval-arrow` revision `27f4c1adee83b5b29b7c4b375f6b976324bda308`.

| Check | Result |
| --- | ---: |
| Baseline valid WAVs | 1088/1088 |
| Fixed valid WAVs | 1088/1088 |
| Paired WAV SHA-256 | 1088/1088 identical |
| Shared generated-audio WER | 3.0394% |
| Reference-audio WER | 1.2182% |

Every paired WAV file was byte-identical, so the fix did not change the generated output. Baseline and fixed outputs therefore share the same generated-audio WER; scoring both copies separately would not add an A/B comparison. The reference-audio WER checks the ASR evaluation pipeline against the original recordings.

## Benchmark & Profiling

The memory experiment ran two consecutive 50-request passes without restarting the service. Memory was measured after the request queue drained and the device synchronized, so the results represent memory that remained allocated after request processing rather than temporary generation peaks.

| Variant | Pass 1 | Pass 2 |
| --- | ---: | ---: |
| Baseline | +171.5 MiB | +169.7 MiB |
| With `@torch.no_grad()` | +1.86 MiB | 0 MiB |

The baseline reproduced similar growth in both passes. With the no-grad guard, the small first-pass increase did not repeat, and the second pass ended with no growth in live tensor memory.

The device A/B experiments were run on the earlier audited revision `07b63c0`. The affected predictor path remains unchanged on the current base. The unit tests were run on `9915397` before rebasing over the unrelated `0f0c2b8` change.

## Checklist

- [x] Add unit tests.
- [x] Update documentation / docstrings / example tutorials as needed. *(N/A: no public API or user workflow changes.)*
- [x] Provide throughput / latency benchmark results and accuracy evaluation results as needed. *(Memory profiling and paired output results are included; no throughput claim is made.)*

## CI

Official self-hosted CI has not run yet; it requires a maintainer to add the applicable CI labels.


ping1jing2 (approved):

Ratish1 (approved):
LGTM

celestial-micha:
I chose `no_grad` because it stops the unused backward graph without changing newly allocated outputs into inference tensors.

As the [PyTorch documentation](https://docs.pytorch.org/docs/stable/generated/torch.autograd.grad_mode.inference_mode.html) explains, `inference_mode` also disables view tracking and version-counter updates. This can reduce overhead, but it adds restrictions on how tensors created inside it can be used afterwards. For example, a later autograd operation that needs to save an inference tensor for backward will fail.

I checked this on a fresh Ascend NPU environment using frozen source `556166ddeae1b6f571022f5498352ec4bee5e953`. Each candidate added only its decorator to the baseline. The harness called the real `code_predictor_forward` entry point with a controlled predictor body, rather than running the full model:

| Check | Baseline | `no_grad` | `inference_mode` |
| --- | --- | --- | --- |
| Additional allocated memory after 100 calls | 44.09 MiB | 0 MiB | 0 MiB |
| Output difference between the two fixes | — | 0 | 0 |
| Later operation that saves a newly allocated output for backward | — | Succeeds | `Inference tensors cannot be saved for backward` |

The last check demonstrates the change in return-value behavior; it does not mean the current production caller performs that operation or has failed this way.

Buffer lifetime is another consideration. Talker reuses buffers and updates them in place. A buffer created as an inference tensor can fail when updated outside inference mode; simply updating an existing ordinary buffer inside inference mode does not turn it into an inference tensor. SGLang encountered this issue with lazily allocated workspaces in [#33470](https://github.com/sgl-project/sglang/issues/33470), and [#33788](https://github.com/sgl-project/sglang/pull/33788) removed inconsistent inference-mode scopes in warm-up. This is a relevant compatibility risk, not a buffer failure reproduced in Talker by this experiment.

`inference_mode` was faster in the synchronized mode-overhead microbenchmark: 0.0846 vs. 0.0941 ms per call, about 9.5 microseconds. That does not establish an end-to-end Talker speedup. Both modes fixed the observed retention, so I kept `no_grad`: it addresses the graph-recording issue while preserving ordinary tensor behavior for returned values and subsequent buffer use.

celestial-micha:
CI update: the only failed job is Qwen3-Omni CI / stage 8 - Video-MME Talker.\n\nThe Talker WER/speed test itself completed successfully and its result artifact was uploaded. The job was marked failed afterwards by the self-hosted runner cleanup hook:\n\n`\nError response from daemon: removal of container ... is already in progress\n`\n\nThis is a runner cleanup race rather than a test or code failure. I tried /rerun-failed-ci, but the repository's Slash Command Handler is currently disabled, and the direct failed-job rerun API requires repository admin permission.\n\nCould a maintainer rerun the failed job when convenient? No code change appears necessary.

PR #1721, opened 2026-08-25
[NPU] add Ascend NPU support for higgs-audio-v3-tts-4b

chx96642264:
## Motivation

`higgs-audio-v3-tts-4b` was only testable on CUDA: the pipeline config pinned
`device="cuda"` for all three stages, the batched sampler imported the fused
`sgl_kernel` top-k/top-p renorm kernels unconditionally (an import that pulls
in CUDA-only code and fails on Ascend NPU), and the codec acoustic encoder /
vocoder decode were wrapped with `torch.compile` and relied on CUDA graph
capture — none of which exist on Ascend NPU (Atlas 910B).

So some adaptation work was done on the NPU in this pr.

## Modifications

- `sampler.py`: platform-aware top-k/top-p renorm kernels and platform-resolved device defaults.
- `config.py` / `stages.py`: use `current_platform.device_type`
  instead of hardcoded `"cuda"`; gate `torch.compile` and codec decode CUDA
  graphs on platform support (eager on NPU).
- `engine_builder.py`: `sampling_backend="pytorch"` (can work on every backend).

## Related Issues

N/A

## Accuracy Test

Test in sglang v0.5.18.

Evaluated on `zhaochenyang20/seed-tts-eval-arrow` (SeedTTS EN, 1088 samples,
voice-clone, non-streaming), with Qwen3-ASR-1.7B as the transcription oracle,
on Ascend NPU (Atlas 910B).

| Metric(concurrency=16) | NPU (this PR) | H200 |
|---|---|---|
| wer_corpus | 0.95% | 1.10% |
| wer_per_sample_mean | 0.91% | 1.08% |
| wer_per_sample_median | 0.00% | 0.00% |
| wer_per_sample_std | 3.4% | 4.1% |

The non-outlier corpus WER (0.95%) is at or below the CUDA reference,
confirming no systematic accuracy regression from the NPU adaptation.

Verification screenshot is as follows:

higgs_tts_wer_new

## Benchmark & Profiling
Here are the performance-related test results. I have also added some test cases for the NPU scenario. 

| Concurrency | Throughput (req/s) | Mean Latency | RTF Mean | audio_s/s |
|---|---|---|---|---|
| 1 | 0.691 | 1.447 s | 0.3424 | 2.948 |
| 2 | 1.394 | 1.433 s | 0.3366 | 6.013 |
| 4 | 2.422 | 1.65 s | 0.3914 | 10.323 |
| 8 | 4.09 | 1.951 s | 0.4638 | 17.407 |
| 16 | 6.269 | 2.536 s | 0.6124 | 26.176 |

Performance verification screenshot is as follows:

higgs_tts_perf_c1
higgs_tts_perf_c2
higgs_tts_perf_c4
higgs_tts_perf_c8
higgs_tts_perf_c16

Test cases verification screenshot is as follows:
higgs_tts_test_new

## Checklist

- [ ] Add unit tests.

## CI

model, and


ping1jing2:
i convert this PR to draft due to no description here, please click "Ready to review" after it's ready

sglang_omni/models/qwen3_asr/engine_builder.py
```diff
@@ -123,18 +118,26 @@ def pre_infra_setup(self, checkpoint_dir: str) -> None:
...
-            "enable_torch_compile": self.enable_torch_compile,
+            "enable_torch_compile": enable_torch_compile,
             "torch_compile_max_bs": self.torch_compile_max_bs,
             "mem_fraction_static": self.mem_fraction_static,
             "max_prefill_tokens": 4096,
             "chunked_prefill_size": 4096,
             "sampling_backend": "pytorch",
             "dtype": dtype,
-            "cuda_graph_backend_prefill": CudaGraphBackend.BREAKABLE,
+            "cuda_graph_backend_prefill": (
```
Makcum888e: should we add `current_platform.cuda_graph_backend_prefill` here?

chx96642264: Sorry for including other unrelated code in the draft. This is not my modification.

sglang_omni/models/higgs_tts/__init__.py
```diff
@@ -13,18 +13,23 @@
...
+# gated off on Ascend NPU (no torch.compile backend), so the capability is CUDA-only.
+_supports_torch_compile = not current_platform.is_npu()
+
 CAPABILITIES = ModelCapabilities(
     supports_reference_audio=True,
     supports_batch_vocoder=True,
     supports_streaming_vocoder=True,
     supports_cuda_graph=True,
-    supports_torch_compile=True,
+    supports_torch_compile=_supports_torch_compile,
```
ping1jing2: please check and describe the root cause about why we don't support torch.compile

sglang_omni/models/higgs_tts/sampler.py
```diff
@@ -30,6 +29,62 @@
...
+        -1, (k_safe - 1).unsqueeze(-1)
+    ).squeeze(-1)
+    masked = torch.where(probs < kth.unsqueeze(-1), torch.zeros_like(probs), probs)
+    denom = masked.sum(dim=-1, keepdim=True)
+    return masked / denom.clamp_min(torch.finfo(probs.dtype).tiny)
+
+
+def _top_p_renorm_prob_torch(probs: torch.Tensor, p: torch.Tensor) -> torch.Tensor:
+    """torch fallback for ``sgl_kernel.top_p_renorm_prob`` (Ascend NPU).
+
```
ping1jing2: i suggest to move these func into a separate script file named `npu_fallback.py`

sglang_omni/models/higgs_tts/sampler.py
```diff
@@ -265,10 +322,10 @@ def _sample_independent_batched(
             .to(torch.int32)
             .contiguous()
         )
-        probs = _fused_top_k_renorm(probs, tk)
+        probs = _top_k_renorm(probs, tk)
     if top_p is not None:
         tp = top_p.view(B, 1).expand(B, N).reshape(B * N).to(torch.float32).contiguous()
-        probs = _fused_top_p_renorm(probs, tp)
+        probs = _top_p_renorm(probs, tp)
```
ping1jing2: you are not supposed to change this func name

sglang_omni/models/higgs_tts/sampler.py
```diff
@@ -30,6 +29,62 @@
...
+    it falls back to the equivalent torch implementations above.
+    """
+    if current_platform.device_type == "cuda":
+        try:
+            from sgl_kernel import top_k_renorm_prob, top_p_renorm_prob
+
+            return top_k_renorm_prob, top_p_renorm_prob
+        except ImportError:
+            pass
+    return _top_k_renorm_prob_torch, _top_p_renorm_prob_torch
```
ping1jing2:
```suggestion
def _resolve_renorm_kernels():
    """Return ``(top_k_renorm, top_p_renorm)`` callables for this platform.

    CUDA uses the fused ``sgl_kernel`` kernels to avoid a full-vocab
    ``torch.sort`` in decode. Ascend NPU has no fused renorm kernels so
    it falls back to the equivalent torch implementations above.
    """
    if current_platform.device_type == "cuda":
        from sgl_kernel import top_k_renorm_prob, top_p_renorm_prob

        return top_k_renorm_prob, top_p_renorm_prob
    if current_platform.is_npu():
         from .npu_fallback import top_k_renorm_prob, top_p_renorm_prob
         return top_k_renorm_prob, top_p_renorm_prob

    return _top_k_renorm_prob_torch, _top_p_renorm_prob_torch
```

sglang_omni/models/higgs_tts/sampler.py
```diff
@@ -30,6 +29,62 @@
...
+
+
+_top_k_renorm, _top_p_renorm = _resolve_renorm_kernels()
```
ping1jing2:
```suggestion
_fused_top_k_renorm, _fused_top_p_renorm = _resolve_renorm_kernels()
```

ping1jing2 (approved):

sglang_omni/models/higgs_tts/sampler.py
chx96642264: Done. I will keep the original _fused_top_k_renorm / _fused_top_p_renorm names.

sglang_omni/models/higgs_tts/__init__.py
chx96642264: Done. NPU does not support torch.compile because the stage_workers.tts_frontend process fails during startup: torch_npu's inductor backend has incomplete support for complex dynamic-shape graphs and crashes during the codegen stage. See the traceback below for details.

higgs_tts_not_compile

sglang_omni/models/higgs_tts/sampler.py
chx96642264: Done. I have migrated the fallback implementations to npu_fallback.py as suggested.

sglang_omni/models/higgs_tts/sampler.py
chx96642264: Done. Please review the changes to see if they meet your requirements.

sglang_omni/models/higgs_tts/sampler.py
chx96642264: Done. Please review the changes to see if they meet your requirements.

sglang_omni/models/higgs_tts/engine_builder.py
```diff
@@ -83,6 +83,7 @@ def generation_defaults(
                 if self.total_gpu_memory_fraction is not None
                 else 0.85
             ),
+            "sampling_backend": "pytorch",
```
Ratish1: NPU already resolves the SGLang sampling backend to PyTorch because FlashInfer is NVIDIA-only. remove this override so the NPU path stays unchanged and CUDA retains its existing auto-selected backend.

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -535,6 +548,15 @@ def create_vocoder_executor(
     # margin; when overriding strides, re-derive the domain empirically.
     checkpoint_dir = resolve_checkpoint(model_path)
     codec = get_or_load_codec(checkpoint_dir, device, dtype)
+    if compile_decode:
+        if current_platform.is_npu():
+            logger.warning(
+                "compile_decode=True was requested but the current platform (%s) "
+                "has no torch.compile backend; falling back to the eager "
```
Ratish1: nit: torch-npu has a compile backend; this codec dynamic-shape graph is the unsupported part. should we change this warning to mention that?

sglang_omni/models/higgs_tts/stages.py
```diff
@@ -379,7 +379,7 @@ def _preprocess(payload: StagePayload) -> StagePayload:
 def create_audio_encoder_executor(
     model_path: str,
     *,
-    device: str = "cuda:0",
+    device: str | None = None,
```
Ratish1: nit: this stage now runs on the selected accelerator. change `GPU stage` in the docstring and the remaining `same-GPU processes` wording in `config.py` to device-neutral terms.

Ratish1 (commented):
Can we fix the lint as well @chx96642264 , thanks

sglang_omni/models/higgs_tts/engine_builder.py
chx96642264: Done. I have removed the override.

sglang_omni/models/higgs_tts/stages.py
chx96642264: Done. You're right, this was something I overlooked. It has been fixed.

sglang_omni/models/higgs_tts/__init__.py
```diff
@@ -13,18 +13,23 @@
...
+from sglang_omni.platforms import current_platform
 
 from . import config
 from .hf_config import HiggsMultimodalQwen3Config
 
 AutoConfig.register("higgs_multimodal_qwen3", HiggsMultimodalQwen3Config)
 
+# NPU's compile backend cannot compile this codec model 
+# (crashes on dynamic shapes, verified on Atlas 910B).
+_supports_torch_compile = not current_platform.is_npu()
```
Ratish1: capability lookup is keyed only by model architecture, and the shared capability tests require Higgs to remain `supports_torch_compile=True`. on NPU this import-time value becomes false and contradicts that contract. keep the architecture capability true and leave the NPU-specific disablement at the two actual compile call sites.

chx96642264: Done. I tested and verified this, and your point is indeed valid. This was an addition I made by mistake. 
The NPU-specific disablement stays at the two actual compile call sites in stages.py (the acoustic-encoder gate and the vocoder compile_decode fallback).

Ratish1:
The unit test is failing, could we fix it @chx96642264 , thanks

PR #1724, opened 2026-08-25
[Perf] Persistent lazy eviction heap for the radix tree cache

### LINKED ISSUE #1723: [Perf] O(leaves) heap rebuild in radix cache eviction
## TL;DR

- **Problem**: `evict()` re-heapifies all evictable leaves on every call, costing O(leaves) on the scheduler thread once the pool saturates.
- **Impact**: zero-hit short requests plus an already host-bound scheduler make omni far more exposed than LLM serving; TTS throughput drops sharply at saturation, ASR even more so.

## Problem

`RadixCache.evict()` rebuilds its priority heap from scratch on every call: it lists all evictable leaves and heapifies them, an O(L) operation where L is the number of evictable leaves.

This only bites when two conditions hold at once:

- the KV pool is saturated, so every allocation shortfall triggers an evict rather than an occasional one;
- L is large. With a fixed pool, `L = pool_tokens / tokens_per_leaf`, so short requests mean thin leaves and a large L.

The total overhead is `allocation_rate x O(L)`, paid entirely on the scheduler thread. Below saturation the cost is zero, which is why short benchmarks never see it.

Both upstream cache implementations have this pattern: the classic `RadixCache` (what omni uses) and the newer unified core.

## Why omni is hit hardest

- **Zero-hit traffic**: distinct prefix per request, near-zero hit ratio, pool always saturated.
- **Short requests**: thin leaves, large L, more frequent evicts.
- **Host-bound scheduler**: scheduler thread already the bottleneck (overlap disabled), evict cost hits throughput directly.

## Impact

All numbers below are measured.

- **qwen3_tts**: 12 loops, c32; step at loop 9; paired fix arm stays flat all 12 loops.

| metric | steady (loop 2-8) | after step (loop 9-12) | change |
| --- | --- | --- | --- |
| qps | 9.28-9.50 | 7.74-8.13 | -17 ~ -19% |
| p50 | 3.28-3.36s | 3.87-4.05s | +21% |
| p95 | 4.83-5.09s | 5.82-6.23s | +21% |
| p99 | 5.62-5.87s | 6.67-7.21s | +21% |
| audio ttfp p95 | 0.84-0.97s | 1.44-1.54s | +62% |

- **qwen3_asr**: 25 loops, c32, unique traffic; collapse from loop 19; radix-off arm steady at 245-272 qps, p95 0.16-0.36s throughout. Eviction accounts for 78.8% of the scheduler thread in the collapse window (py-spy).

| metric | steady (loop 3-18) | collapse (loop 19-25) | change |
| --- | --- | --- | --- |
| qps | 227-263 | 88.9 -> 55.6 -> 31.5 -> 30.8 -> 24.1 -> 26.1 -> 20.6 | -92%, still worsening |
| p50 | 0.11-0.13s | 0.20 -> 0.67-2.05s | ~8x |
| p95 | 0.16-0.28s | 0.97 -> 3.80s | ~20x |
| p99 | 0.20-0.57s | 1.19 -> 4.03s | ~15x |

- **moss_tts**: evict at 1.49% of the pipeline thread; constant ~2% throughput tax, no step.

- **whisper**: 30 loops, 35k-token minimal pool re-check identical; qps 66-70 and p95 0.61-0.67s flat on both arms; immune, finished requests never insert into the tree.

## Solution options

- **Default-off**: disable radix cache for qwen3_tts / moss_tts / qwen3_asr / fun_asr / arkasr / whisper_asr since low/zero hit rate; per-pipeline scope only (#1772). 
- **Persistent lazy min-heap**: push a node once when it becomes an evictable leaf, validate at pop (membership + priority recheck, re-push on refresh), compact when stale entries dominate; eviction order unchanged.
  - In this repo: an `EvictHeapRadixCache` subclass wired into `create_tree_cache` (#1724).
  - In upstream sglang: the same construction applies to classic `RadixCache.evict` and the unified full component's eviction walks.

lijrjyan:
## Motivation

Fix the eviction-driven uptime throughput decay tracked in #1723: `RadixCache.evict` re-heapifies all evictable leaves on every call, O(leaves) on the scheduler thread once the pool saturates.

## Modifications

- Add `EvictHeapRadixCache`: a `RadixCache` subclass with a persistent lazy min-heap (push on leaf-add, validate at pop, compact on staleness); eviction order unchanged.
- Wire `create_tree_cache` to return it; `--disable-radix-cache` path unchanged.
- Unit tests: eviction-order equivalence vs upstream `RadixCache`, heap boundedness, reset.

## Related Issues

Fixes #1723

## Accuracy Test

Eviction order is bit-identical to upstream `RadixCache` on randomized traces (unit tests, 3 seeds).

## Benchmark & Profiling

qwen3_tts paired soak, 12 loops, c32, zero-hit traffic. The fix arm ran the identical persistent-heap patch applied directly to sglang 0.5.16's `RadixCache` (this PR relocates the same logic into the subclass; eviction order proven identical by the unit tests). The baseline arm steps down at loop 9, the fix arm stays flat. Post-saturation window (loop 9-12):

| Metric | Baseline (`RadixCache`) | Fix (persistent heap) | Delta |
| --- | --- | --- | --- |
| qps | 7.74-8.13 | 9.30-9.39 | +19% |
| latency p50 | 3.87-4.05s | 3.34-3.35s | -17% |
| latency p95 | 5.82-6.23s | 4.90-5.03s | -18% |
| latency p99 | 6.67-7.21s | 5.64-5.96s | -18% |
| audio ttfp p95 | 1.44-1.54s | 0.93-0.95s | -39% |

Pre-saturation window (loop 2-8): both arms 9.3-9.5 qps, no regression.

qwen3_asr paired soak, 25 loops, c32, emulated all-unique traffic, small pool; both arms saturate at loop 19-20. Post-saturation window (loop 20-25):

| Metric | Baseline (`RadixCache`) | Fix (persistent heap) | no-radix reference |
| --- | --- | --- | --- |
| qps | 28.8-54.4 | 137-172 | 245-270 |
| latency p95 | 1.85-2.58s | 0.43-0.45s | 0.16-0.36s |
| latency p99 | 1.93-2.66s | 0.43-0.62s | 0.20-0.50s |

The heap rebuild collapse is removed (~4-5x qps, ~5x p95 vs baseline); the residual gap to the no-radix reference is the intrinsic per-shortfall eviction work under all-unique traffic, which is what the zero-hit default-off option in #1723 addresses complementarily.

Per-model regression matrix (fix is default-on for every pipeline; normal replay traffic, default pools):

| Model | Design | qps delta | p95 delta | Verdict |
| --- | --- | --- | --- | --- |
| qwen3_tts (1.7B) | paired dual-GPU soak | +19% post-saturation | -18% | improvement |
| qwen3_asr (1.7B) | paired dual-GPU, unique traffic | 3-5x post-saturation | -83% | improvement |
| mtd | sequential single-GPU | +0.8% | -0.7% | flat |
| whisper | sequential single-GPU | +1.6% | -1.0% | flat |
| fun_asr | sequential single-GPU | +0.2% | -4.4% | flat |
| moss_tts | paired dual-GPU | +3.6% | -2.1% | flat / slightly positive |
| arkasr | paired dual-GPU | -1.0% | -0.3% | flat (silicon band) |

No pipeline regresses; the two saturation-exposed pipelines improve substantially.

## Checklist

- [x] Add unit tests.

## CI

`/tag-and-rerun-ci whisper-asr` to select an ASR CI model. One selector from each family can be combined, for example

### DIFF
--- a/sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
+++ b/sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
@@ -1,0 +1,97 @@
+# SPDX-License-Identifier: Apache-2.0
+"""RadixCache with a persistent lazy eviction heap.
+
+Upstream ``RadixCache.evict`` rebuilds its eviction heap from every
+evictable leaf on each call (``list(...)`` + ``heapify``, O(L)). Under
+sustained load with few prefix hits the KV pool saturates, ``evict`` then
+fires on every allocation shortfall, and the scheduler thread spends its
+budget re-heapifying an ever-growing leaf set. This subclass keeps one
+min-heap alive across calls: push when a node becomes an evictable leaf,
+validate lazily at pop.
+
+Lazy validation is exact for eviction strategies whose priority is a
+non-decreasing function of node-local state (LRU/LFU/FIFO all qualify):
+a stale recorded priority is a lower bound on the current one, so the heap
+top never overtakes a fresher node and re-pushing refreshed nodes preserves
+the eviction order.
+"""
+
+from __future__ import annotations
+
+import heapq
+import time
+
+from sglang.srt.mem_cache.base_prefix_cache import EvictParams, EvictResult
+from sglang.srt.mem_cache.radix_cache import RadixCache, TreeNode
+
+
+class EvictHeapRadixCache(RadixCache):
+    def __init__(self, params):
+        # Before super().__init__: RadixCache.__init__ calls reset().
+        self._evict_heap: list = []
+        self._evict_heap_seq = 0
+        super().__init__(params)
+
+    def reset(self):
+        self._evict_heap.clear()
+        super().reset()
+
+    def _evict_heap_push(self, node: TreeNode) -> None:
+        self._evict_heap_seq += 1
+        heapq.heappush(
+            self._evict_heap,
+            (self.eviction_strategy.get_priority(node), self._evict_heap_seq, node),
+        )
+
+    def _evict_heap_rebuild(self) -> None:
+        self._evict_heap = [
+            (self.eviction_strategy.get_priority(n), i, n)
+            for i, n in enumerate(self.evictable_leaves)
+        ]
+        self._evict_heap_seq = len(self._evict_heap)
+        heapq.heapify(self._evict_heap)
+
+    def _update_leaf_status(self, node: TreeNode) -> None:
+        was_evictable = node in self.evictable_leaves
+        super()._update_leaf_status(node)
+        if not was_evictable and node in self.evictable_leaves:
+            self._evict_heap_push(node)
+
+    def evict(self, params: EvictParams) -> EvictResult:
+        if self.disable:
+            return EvictResult()
+
+        start_time = time.perf_counter()
+        num_tokens = params.num_tokens
+
+        # Compact when stale entries dominate, so heap size stays O(leaves).
+        if len(self._evict_heap) > max(1024, 4 * len(self.evictable_leaves)):
+            self._evict_heap_rebuild()
+        if not self._evict_heap and self.evictable_leaves:
+            # The push in _update_leaf_status should make this unreachable;
+            # rebuild rather than silently under-evict if it is ever violated.
+            self._evict_heap_rebuild()
+
+        num_evicted = 0
+        while num_evicted < num_tokens and self._evict_heap:
+            priority, _seq, x = heapq.heappop(self._evict_heap)
+
+            if x not in self.evictable_leaves:
+                # Stale: evicted earlier, locked, or no longer a leaf.
+                continue
+            current_priority = self.eviction_strategy.get_priority(x)
+            if current_priority != priority:
+                # Refreshed since push (priorities are non-decreasing):
+                # re-insert at the up-to-date priority and keep popping.
+                self._evict_heap_push(x)
+                continue
+
+            self.token_to_kv_pool_allocator.free(x.value)
+            num_evicted += len(x.value)
+            # _delete_leaf -> _update_leaf_status(parent) pushes the parent
+            # if it just became an evictable leaf.
+            self._delete_leaf(x)
+            self._record_remove_event(x)
+
+        self.update_eviction_metrics(num_evicted, start_time)
+        return EvictResult(num_tokens_evicted=num_evicted)
--- a/sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
+++ b/sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
@@ -26,6 +26,8 @@ class EvictHeapRadixCache(RadixCache):
             self._evict_heap,
             (self.eviction_strategy.get_priority(node), self._evict_heap_seq, node),
         )
+        if len(self._evict_heap) > max(1024, 4 * len(self.evictable_leaves)):
+            self._evict_heap_rebuild()
 
     def _evict_heap_rebuild(self) -> None:
         self._evict_heap = [
@@ -47,9 +49,6 @@ class EvictHeapRadixCache(RadixCache):
 
         start_time = time.perf_counter()
         num_tokens = params.num_tokens
-
-        if len(self._evict_heap) > max(1024, 4 * len(self.evictable_leaves)):
-            self._evict_heap_rebuild()
 
         num_evicted = 0
         while num_evicted < num_tokens and self._evict_heap:
--- a/sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
+++ b/sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
@@ -1,5 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
-"""A RadixCache whose eviction heap persists across evict() calls."""
+"""A RadixCache whose eviction heap persists across evict() calls.
+
+Order-equivalent to RadixCache for lru only; the factory falls back to
+RadixCache for other policies.
+"""
 
 from __future__ import annotations
 
--- a/sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
+++ b/sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
@@ -1,19 +1,9 @@
 # SPDX-License-Identifier: Apache-2.0
-"""RadixCache with a persistent lazy eviction heap.
+"""RadixCache variant with a persistent lazy min-heap over the evictable leaves.
 
-Upstream ``RadixCache.evict`` rebuilds its eviction heap from every
-evictable leaf on each call (``list(...)`` + ``heapify``, O(L)). Under
-sustained load with few prefix hits the KV pool saturates, ``evict`` then
-fires on every allocation shortfall, and the scheduler thread spends its
-budget re-heapifying an ever-growing leaf set. This subclass keeps one
-min-heap alive across calls: push when a node becomes an evictable leaf,
-validate lazily at pop.
-
-Lazy validation is exact for eviction strategies whose priority is a
-non-decreasing function of node-local state (LRU/LFU/FIFO all qualify):
-a stale recorded priority is a lower bound on the current one, so the heap
-top never overtakes a fresher node and re-pushing refreshed nodes preserves
-the eviction order.
+Pop-time validation is exact because every built-in eviction strategy's
+priority (LRU/LFU/FIFO) is non-decreasing in node-local state, so a stale
+entry is a lower bound and re-pushing a refreshed node preserves the order.
 """
 
 from __future__ import annotations
@@ -27,7 +17,7 @@ from sglang.srt.mem_cache.radix_cache import RadixCache, TreeNode
 
 class EvictHeapRadixCache(RadixCache):
     def __init__(self, params):
-        # Before super().__init__: RadixCache.__init__ calls reset().
+        # note (Junnan Li): set before super().__init__, which calls reset().
         self._evict_heap: list = []
         self._evict_heap_seq = 0
         super().__init__(params)
@@ -64,12 +54,8 @@ class EvictHeapRadixCache(RadixCache):
         start_time = time.perf_counter()
         num_tokens = params.num_tokens
 
-        # Compact when stale entries dominate, so heap size stays O(leaves).
+        # note (Junnan Li): compact at 4x staleness so heap size stays O(leaves).
         if len(self._evict_heap) > max(1024, 4 * len(self.evictable_leaves)):
-            self._evict_heap_rebuild()
-        if not self._evict_heap and self.evictable_leaves:
-            # The push in _update_leaf_status should make this unreachable;
-            # rebuild rather than silently under-evict if it is ever violated.
             self._evict_heap_rebuild()
 
         num_evicted = 0
@@ -77,19 +63,15 @@ class EvictHeapRadixCache(RadixCache):
             priority, _seq, x = heapq.heappop(self._evict_heap)
 
             if x not in self.evictable_leaves:
-                # Stale: evicted earlier, locked, or no longer a leaf.
                 continue
             current_priority = self.eviction_strategy.get_priority(x)
             if current_priority != priority:
-                # Refreshed since push (priorities are non-decreasing):
-                # re-insert at the up-to-date priority and keep popping.
                 self._evict_heap_push(x)
                 continue
 
             self.token_to_kv_pool_allocator.free(x.value)
             num_evicted += len(x.value)
-            # _delete_leaf -> _update_leaf_status(parent) pushes the parent
-            # if it just became an evictable leaf.
+            # note (Junnan Li): _delete_leaf relands the parent via _update_leaf_status.
             self._delete_leaf(x)
             self._record_remove_event(x)
 
--- a/sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
+++ b/sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
@@ -1,10 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-"""RadixCache variant with a persistent lazy min-heap over the evictable leaves.
-
-Pop-time validation is exact because every built-in eviction strategy's
-priority (LRU/LFU/FIFO) is non-decreasing in node-local state, so a stale
-entry is a lower bound and re-pushing a refreshed node preserves the order.
-"""
+"""A RadixCache whose eviction heap persists across evict() calls."""
 
 from __future__ import annotations
 
@@ -17,7 +12,6 @@ from sglang.srt.mem_cache.radix_cache import RadixCache, TreeNode
 
 class EvictHeapRadixCache(RadixCache):
     def __init__(self, params):
-        # note (Junnan Li): set before super().__init__, which calls reset().
         self._evict_heap: list = []
         self._evict_heap_seq = 0
         super().__init__(params)
@@ -54,7 +48,6 @@ class EvictHeapRadixCache(RadixCache):
         start_time = time.perf_counter()
         num_tokens = params.num_tokens
 
-        # note (Junnan Li): compact at 4x staleness so heap size stays O(leaves).
         if len(self._evict_heap) > max(1024, 4 * len(self.evictable_leaves)):
             self._evict_heap_rebuild()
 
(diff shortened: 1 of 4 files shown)


lijrjyan:
Server-side + data-structure evidence for this PR, beyond the soak tables in the body.

## 1. Per-evict latency vs leaf count (server, in-process timing around `evict()`)

qwen3_asr, salted all-unique traffic (radix hit ratio 0% by construction: the probe forces match misses and salts `extra_key` at insert, so the replayed dataset emulates all-unique production traffic), probe branch with an env-gated trace; saturated regime (leaves >= 0.9*max):

| measured leaves | baseline p50 | fix p50 | ratio | baseline p99 | fix p99 |
| 3.2k | 1227 us | 56 us | 22x | 2.9 ms | 1.4 ms |
| 3.9k | 1258 us | 50 us | 25x | 4.1 ms | 0.9 ms |
| 6.9k | 1768 us | 48 us | 37x | 4.7 ms | 0.5 ms |
| 23.1k | 7389 us | 58 us | 127x | 633 ms | 0.7 ms |

Baseline grows ~linearly in leaves with 650+ms extreme spikes; fix stays flat at ~50-58 us. First three rows are independent 2-GPU pairs; the 23k row ran both arms on one GPU — in-process timing plus same-window common-mode makes the ratio robust, and it is trend-continuous with the 2-GPU rows.

CPU counterpart (isolated, no server, 1k-300k leaves): baseline p50 strictly linear 108 us -> 102 ms; fix 7.1 -> 16.9 us (~log L); 6000x at 300k. Scripts available on request.

## 2. Lazy-heap stability under sustained churn

- CPU, 15 min, 14.5M evicts at 100k steady leaves with continuous stale-entry generation: heap size tracked live leaves exactly at every sample, 0 compaction rebuilds, 0 bound violations; p99.9 = 34.8 us; the only >4 ms spikes (8 in 14.5M) correlate with CPython GC, not compaction.
- Server, 150 loops (163k requests) on one live process: heap/leaves stayed <= 1.47 while leaves grew 1k -> 113k; per-evict p50 flat at ~55 us end to end; request-level p95 flat at 0.44-0.48 s for all 150 loops (no drift, no spike loops).

## 3. New finding: unbounded leaf growth under unique `extra_key`

The 113k leaves in a 40k-token pool are real: every unique `extra_key` leaves a zero-token partition leaf behind after its chain is evicted, and these accumulate without a pool cap (61k -> 113k linear over the run). Since omni TTS uses uuid4 `extra_key` in production, baseline's O(leaves) cost grows without bound over uptime — this completes the explanation of the ASR/TTS "superlinear" shape. The fix absorbs it (55 us at 113k leaves), but the accumulation itself may deserve a separate cleanup issue.

lijrjyan:
Production-pool cold-start-to-saturation curve (qwen3_asr-1.7B, default pool 1,173,165 tokens, unique traffic, paired arms, three offered-load levels):

**Saturation onset is arm-independent** (same fill dynamics): ~73k leaves at onset; time-to-saturation after ready: 1011s (c=8), 693s (c=32), 675s (c=64). Per-evict at 73k leaves: baseline p50 26-27ms with p95 ~750ms; fix p50 57-66us (~430x).

| conc | phase | baseline qps | fix qps | baseline p99 | fix p99 |
| 8 | pre-saturation | 119.2 | 120.2 | 0.104s | 0.103s |
| 8 | post-saturation | 6.5 | 39.0 | 2.28s | 0.36s |
| 32 | pre-saturation | 290.0 | 295.7 | 0.197s | 0.191s |
| 32 | post-saturation | 11.4 | 135.4 | 7.28s | 0.46s |
| 64 | pre-saturation | 338.6 | 328.5 | 0.336s | 0.389s |
| 64 | post-saturation | 9.7 | 207.7 | 14.29s | 0.69s |

Effective region: **pre-saturation the fix is free** (arms statistically identical at every load level); **post-saturation baseline collapses to 6-11 qps regardless of offered load** (the scheduler is evict-bound at ~26ms per eviction) while the fix holds 39/135/208 qps with p99 <= 0.7s -- up to 21x at c=64.

qps timeline
p95 timeline
p99 timeline

lijrjyan:
Same production-pool cold-start curve for the TTS side (qwen3_tts-1.7B, default pool 1,060,796 tokens, zero-hit uuid `extra_key` traffic, paired arms, three offered-load levels):

**Saturation is arm-independent**: ~8.9k leaves at onset (fat ~120-token leaves), reached at 1282s / 876s / 855s after ready at c=8/32/64. Per-evict at 8.9k leaves: baseline p50 4.4-5.0ms, p95 6.4-7.0ms; fix p50 118-149us (~35x) -- consistent with the ASR point (26ms at 73k thin leaves) under the linear-in-leaves model.

| conc | phase | baseline qps | fix qps | baseline p95 | fix p95 | baseline p99 | fix p99 |
| --- | --- | --- | --- | --- | --- | --- | --- |
| 8 | pre-saturation | 7.24 | 7.29 | 1.56s | 1.56s | 1.76s | 1.78s |
| 8 | post-saturation | 6.36 | 7.24 | 2.10s | 1.60s | 2.39s | 1.78s |
| 32 | pre-saturation | 11.05 | 10.95 | 4.21s | 4.20s | 4.81s | 4.88s |
| 32 | post-saturation | 8.85 | 10.81 | 5.28s | 4.31s | 5.98s | 4.93s |
| 64 | pre-saturation | 11.54 | 11.64 | 7.99s | 7.92s | 9.03s | 8.90s |
| 64 | post-saturation | 9.17 | 11.37 | 9.44s | 8.08s | 10.74s | 9.20s |

Effective region mirrors ASR at a milder magnitude: pre-saturation the arms are identical; post-saturation baseline steps down 12-20% in qps with p95/p99 up 18-35%, while the fix stays flat. TTS is bounded (leaves capped by the pool at ~9k) rather than collapsing, matching the one-time-step signature in the PR body.

qps timeline
p95 timeline
p99 timeline

lijrjyan:
Model-scale extension: same cold-start curve on qwen3_tts-0.6B (production pool 1,089,558 tokens, zero-hit traffic, paired arms):

Saturation: ~9.0k leaves at 1982s / 1327s / 935s after ready (c=8/32/64), arm-independent. Per-evict at 9k leaves: baseline p50 4.9-7.3ms (p95 up to 19.6ms), fix p50 126-176us (~30-45x) -- consistent with the 1.7B point.

Post-saturation window (paired, same-window; the 0.6B warms up slower so pre-saturation loops are still climbing at c=8 and are not used as a comparison window):

| conc | baseline qps | fix qps | delta | baseline p95 | fix p95 |
| 8 | 5.62 | 7.04 | +25% | 2.23s | 1.60s |
| 32 | 7.28 | 8.47 | +16% | 6.30s | 5.18s |
| 64 | 9.04 | 10.30 | +14% | 9.33s | 8.32s |

qps timeline
p95 timeline
p99 timeline

lijrjyan:
Scale matrix complete: qwen3_asr-0.6B (production pool 1,195,016 tokens, unique traffic, paired arms) is the most exposed cell.

Saturation: ~75k leaves at 988s / 661s / 639s after ready (c=8/32/64), arm-independent. Per-evict at 75k leaves: baseline p50 27.7-33.6ms with p95 780-830ms; fix p50 59-79us (~470-570x).

| conc | phase | baseline qps | fix qps | baseline p99 | fix p99 |
| 8 | pre-saturation | 136.4 | 137.5 | 0.091s | 0.090s |
| 8 | post-saturation | 4.3 | 34.0 | 2.39s | 0.28s |
| 32 | pre-saturation | 304.2 | 295.3 | 0.184s | 0.191s |
| 32 | post-saturation | 4.6 | 104.1 | 8.71s | 0.49s |
| 64 | pre-saturation | 333.2 | 347.2 | 0.358s | 0.334s |
| 64 | post-saturation | 4.4 | 163.5 | 16.74s | 0.68s |

The 0.6B collapses deeper than the 1.7B (baseline 4.4 vs 9.7 qps at c=64): a faster model leaves the scheduler more host-bound, so the same per-evict cost destroys more throughput. Scale-matrix summary across all four cells:

| cell | leaves at sat | baseline per-evict p50 | fix per-evict p50 | post-sat qps gap (c=64) |
| --- | --- | --- | --- | --- |
| ASR-0.6B | ~75k | 27.7-33.6ms | 59-79us | 37x |
| ASR-1.7B | ~73k | 26.3-27.2ms | 57-66us | 21x |
| TTS-0.6B | ~9k | 4.9-7.3ms | 126-176us | +14% (fix) |
| TTS-1.7B | ~8.9k | 4.4-5.0ms | 118-149us | fix flat, base -20% |

Baseline per-evict cost tracks leaf count (~0.4us/leaf) across cells; fix stays in the tens-of-us band everywhere.

qps timeline
p95 timeline
p99 timeline

sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
```diff
@@ -0,0 +1,72 @@
...
+            self._evict_heap_push(node)
+
+    def evict(self, params: EvictParams) -> EvictResult:
+        if self.disable:
+            return EvictResult()
+
+        start_time = time.perf_counter()
+        num_tokens = params.num_tokens
+
+        if len(self._evict_heap) > max(1024, 4 * len(self.evictable_leaves)):
```
JiaxinD: The rebuild check only runs inside evict(), which upstream only reaches on a pool shortfall; a server whose pool never fills keeps pushing one tuple per lock/unlock cycle and never trims. Cheapest fix: run the same threshold check in `_evict_heap_push` too.

tests/unit_test/scheduling/test_evict_heap_radix_cache.py
```diff
@@ -0,0 +1,116 @@
...
+        patched_order = _run_trace(patched, seed)
+        assert patched_order == stock_order
+        assert len(patched.evictable_leaves) == len(stock.evictable_leaves)
+
+
+def test_heap_stays_bounded_and_recovers():
+    cache = _make(EvictHeapRadixCache)
+    _run_trace(cache, seed=7, steps=2000)
+    # Post-evict, live entries are bounded by the compaction threshold.
+    assert len(cache._evict_heap) <= max(1024, 4 * len(cache.evictable_leaves))
```
JiaxinD, line 112: `_run_trace` ends with a full drain, so this assertion sees an empty heap. Asserting mid-trace, before the final drain, would make it bite.

tests/unit_test/scheduling/test_evict_heap_radix_cache.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+    )
+
+
+def _run_trace(cache, seed: int, steps: int = 4000) -> list[tuple[int, ...]]:
+    """Drive an identical insert/lock/unlock/evict trace; return eviction order."""
+    order = []
+    orig_delete = cache._delete_leaf
+    cache._delete_leaf = lambda node: (
+        order.append(tuple(node.key.token_ids)),
```
JiaxinD: Two leaves that differ only in `extra_key` record the same tuple here, so a swapped eviction order between them still passes. Recording `(extra_key, token_ids)` closes that.

sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
lijrjyan: Done. Moved the threshold check into `_evict_heap_push`.

tests/unit_test/scheduling/test_evict_heap_radix_cache.py
lijrjyan: Done. The trace now records `(extra_key, token_ids)`.

tests/unit_test/scheduling/test_evict_heap_radix_cache.py
lijrjyan: Done. The bound is asserted mid-trace before the final drain.

tests/unit_test/scheduling/test_evict_heap_radix_cache.py
```diff
@@ -0,0 +1,118 @@
...
+    def free_segment(self, value, start_pos=0):
+        pass
+
+    def available_size(self):
+        return 1 << 30
+
+
+def _make(cache_cls):
+    """Simulated-cache builder; create_simulated hardcodes RadixCache."""
+    return cache_cls(
```
charliechenye, line 32: One additional call out I think we should pin down: this parity test currently exercises only the default LRU policy because `CacheInitParams.eviction_policy` defaults to `"lru"`.

The pop time stale priority repair is order preserving only when a live node cannot become _more eviction urgent_ than its stored heap key. That holds for the current LRU path because `last_access_time` only increases: a stale entry can pop too early, get detected, and be re-pushed with its newer priority.

`MRU` is different. Its priority is `-last_access_time`, so touching a leaf makes its current priority smaller. Its stale heap entry can then remain buried behind another leaf, meaning that other leaf can pop with a valid priority and be evicted before the now-more-recent leaf ever gets a stale-priority check.

Current Omni wiring appears to use the default LRU policy, so I don't think this is a regression in the shipped path today. But `EvictHeapRadixCache` and the PR currently claim general `RadixCache` eviction order equivalence. Could we either explicitly constrain this implementation to lazy safe policies please?

lijrjyan:
Radix on/off/fix three-way measurements across both model scales, complementing the soak tables in the body. Arms: red = baseline radix (the pathology this PR fixes), blue = this PR's persistent heap, green = ChunkCache reference (radix disabled; proposed as the default for these pipelines in #1772). All runs use salted all-unique traffic, measured before the uuid `extra_key` fix (#1770); on/off and fix arms come from separate runs on identical configs, so read shapes rather than absolute offsets.

**1.7B**

qwen3_tts-1.7B 3-way
qwen3_asr-1.7B 3-way

**0.6B** (smaller model = more host-bound scheduler = deeper baseline collapse)

qwen3_tts-0.6B 3-way
qwen3_asr-0.6B 3-way

Replayed-dataset pairs for the remaining pipelines overlap on/off throughout (insert-dedup caps the tree, so the pathology never engages):

remaining pipelines

Reading: this PR moves the red curves onto the blue ones -- the step/collapse is removed at both scales; the residual gap to green is the intrinsic cost of retaining a zero-hit working set, addressed separately by per-pipeline defaults in #1772.

tests/unit_test/scheduling/test_evict_heap_radix_cache.py
lijrjyan: Got ur point. Omni hasn't actually wired up the other sglang policies yet, I am not sure testing and hardcoding them here a bit out of scope. 

I narrowed the constraint to default lru and let every other policy fall back to the upstream RadixCache.If another policy gets adopted later, it can be wired.

JiaxinD (approved):

tests/unit_test/scheduling/test_evict_heap_radix_cache.py
charliechenye: absolutely. Makes sense

sglang_omni/scheduling/sglang_backend/evict_heap_radix_cache.py
```diff
@@ -0,0 +1,75 @@
...
+            priority, _seq, x = heapq.heappop(self._evict_heap)
+
+            if x not in self.evictable_leaves:
+                continue
+            current_priority = self.eviction_strategy.get_priority(x)
+            if current_priority != priority:
+                self._evict_heap_push(x)
+                continue
+
+            self.token_to_kv_pool_allocator.free(x.value)
```
charliechenye: One small follow up from the rebase (SGLang Omni dependency now moves from SGLang 0.5.16 to SGLang 0.5.18)

Upstream `RadixCache.evict()` now uses `free_segment(x.value, start_pos=0)` instead of `free(x.value)`. The result is the same, but the paged allocator specializes `free_segment()` to avoid the `torch.unique(...)` path and its device sync.

Since this PR already replaces `RadixCache.evict()`, could we pick up that newer upstream optimization here as well? It should just be:

`self.token_to_kv_pool_allocator.free_segment(x.value, start_pos=0)`

Not a correctness issue, just making sure the persistent heap path also inherits the allocator side eviction optimization available in the newly pinned SGLang. 

Thanks a lot.

cc: @JiaxinD

charliechenye: "merged main -> branch" did just bring 0.5.18 here to :P

lijrjyan: Done. add that param and the comment same with upstream sglang.

charliechenye: Cool stuffs. Awesome. Thanks a lot

charliechenye (approved):

PR #1726, opened 2026-08-25
[Qwen3-TTS] Implement full-batch mixed sampling

leihehehe:
## Motivation

Qwen3-TTS predictor CUDA graphs currently support homogeneous all-argmax and all-sampled batches. A batch containing both sampled and argmax rows falls back to eager execution.  As a result, mixed batches execute the entire predictor chain eagerly, losing CUDA graph replay and repeatedly paying Python dispatch and kernel-launch overhead during decoding.

This PR keeps mixed batches on the predictor CUDA graph path and exposes per-request subtalker sampling controls through the public speech API.

## Modifications

- Enable CUDA graph execution for mixed sampled/argmax batches using fixed-shape sampling and a device-resident `do_sample` mask.
- Share `sampled` graph keys between fully sampled and mixed batches, while retaining argmax-only graphs for fully greedy batches.
- Preserve argmax tie-breaking and existing eager fast paths, including sampling only the sampled-row subset in mixed eager batches.
- Forward per-request subtalker sampling controls through `stage_params.tts_engine`.
- Add `--subtalker-dosample-ratio` to the SeedTTS benchmark for controlled sampled/argmax request mixtures.
- Track graph-cache capacity fallbacks internally and emit a one-time warning when capacity forces eager execution.
- Extend tests for mixed graph/eager equivalence, graph reuse, mask selection, parameter forwarding, and end-to-end batch invariance.

## Related Issues

#1418 - “Mixed sampled/argmax batches drop the predictor chain graph to eager (sglang_model.py mixed-batch guard); unifying on full-batch masked sampling would keep the graph, gated on the existing bit-identity suite (argmax tie-break must be proven identical).”

## Accuracy Test

The affected Qwen3-TTS predictor CUDA Graph and sampling unit suites pass.

Graphed and eager predictor outputs are bit-identical (`torch.equal`) for argmax, sampled, and mixed (partially-sampled) batches, including padded buckets. Mixed batches share the same `sampled` graph key as fully-sampled batches. Exact-logit ties preserve the existing `torch.argmax` result.

### Methodology

- **Hardware:** 1x NVIDIA H200 SXM 141 GB (`GPU-06862ea3-321e-ad44-aa7d-9ac47526a2b7`)
- **Driver:** 570.211.01
- **Model:** `Qwen/Qwen3-TTS-12Hz-1.7B-Base` (`fd4b2543`)
- **Dataset:** SeedTTS EN (`27f4c1ad`)
- **Base:** `8f8b73d3`
- **Candidate:** `99e190cb`
- **Software:** PyTorch 2.11.0+cu130, SGLang 0.5.16, FlashInfer 0.6.14, SGLang Kernel 0.4.5
- **Server:** `max_running_requests=16`, `max_queued_requests=64`, `cuda_graph_max_bs=16`
- **Request seed:** `1234`

Model and dataset artifacts were cached before measurement; measured sessions ran in Hugging Face offline mode. The repository's standard `benchmark_tts_seedtts.py` harness was used for request dispatch, timing, metrics, and result serialization.

The base (`main`) Qwen3-TTS request builder does not read `stage_params.tts_engine`, so it cannot consume the per-request `subtalker_dosample` values needed to construct mixed traffic. This PR already includes that request-forwarding change. To give both arms the same workload, the base worktree received a benchmark-only backport of the PR's ratio dispatch and request forwarding; it does not modify the base Predictor, CUDA Graph, or scheduler implementation. The backport is reviewable in draft PR [#1764](https://github.com/sgl-project/sglang-omni/pull/1764), which exists only to review and reproduce the benchmark and will not be merged.

An instrumented preflight at ratio 50% / c8 confirmed real partially-sampled batches in both arms (for example, `batch_size=8`, `sampled_rows=4`). The base retained its existing mixed-batch eager fallback; the candidate captured and replayed the shared `sampled` graph key for the same bucket. The temporary composition probe was removed before formal timing.

Each leg used a fresh server. Before timing, one unmeasured concurrent group with the same ratio and concurrency warmed the exact execution mode; the measured harness warmup was set to zero.

- Mixed ratios **20%, 50%, and 70%** were each measured at c8 and c32 using **A/B/A/B**, with **256 requests per leg**. Values below are means over two base and two candidate legs.
- Regression checks used one A/B pair with **256 requests per arm**: graph-on all-argmax (**0%**) and all-sampled (**100%**) at c32, plus graph-off **50%** mixed at c32. Within each pair, base and candidate used the same 256 sample IDs, request seed, per-request sampled/argmax assignment, concurrency, and server configuration; only the worktree changed.

All 30 formal legs completed 256/256 requests with zero graph-capacity fallbacks and zero graph-capture failures.

### Mixed workload results

Ratio is the fraction of requests using `subtalker_dosample=True`; the remainder use argmax. With 256 requests per leg, the deterministic allocation is shown in the table.

| Sampled ratio | Composition per leg | Concurrency | Throughput (base → candidate) | Speedup | Output tok/s (base → candidate) | Speedup | Mean latency | P95 latency |
| 20% | 51 sampled / 205 argmax | c8 | 3.103 → 9.110 req/s | **2.94x** | 159.35 → 465.60 | **2.92x** | 2.563 → 0.871 s (**-66.0%**) | 3.896 → 1.251 s (**-67.9%**) |
| 20% | 51 sampled / 205 argmax | c32 | 5.479 → 14.967 req/s | **2.73x** | 277.50 → 757.25 | **2.73x** | 5.556 → 2.048 s (**-63.1%**) | 7.217 → 2.660 s (**-63.2%**) |
| 50% | 128 sampled / 128 argmax | c8 | 2.807 → 8.914 req/s | **3.18x** | 140.30 → 448.90 | **3.20x** | 2.816 → 0.889 s (**-68.4%**) | 4.303 → 1.254 s (**-70.9%**) |
| 50% | 128 sampled / 128 argmax | c32 | 5.472 → 14.690 req/s | **2.68x** | 274.95 → 743.40 | **2.70x** | 5.491 → 2.084 s (**-62.0%**) | 7.117 → 2.691 s (**-62.2%**) |
| 70% | 179 sampled / 77 argmax | c8 | 2.943 → 8.980 req/s | **3.05x** | 148.45 → 451.05 | **3.04x** | 2.701 → 0.882 s (**-67.3%**) | 4.091 → 1.257 s (**-69.3%**) |
| 70% | 179 sampled / 77 argmax | c32 | 5.509 → 14.784 req/s | **2.68x** | 277.40 → 742.35 | **2.68x** | 5.485 → 2.073 s (**-62.2%**) | 7.089 → 2.733 s (**-61.5%**) |

Across these mixed workloads, mean RTF improved by 61.9 - 68.1% and P95 RTF improved by 58.3 - 65.4%.

### No-regression checks

| Workload | Throughput (base → candidate) | Change | Output tok/s (base → candidate) | Change | Mean latency change | P95 latency change |
| Graph on, all argmax (0 sampled / 256 argmax), c32 | 15.687 → 15.572 req/s | -0.7% | 798.1 → 795.0 | -0.4% | +0.3% | -0.8% |
| Graph on, all sampled (256 sampled / 0 argmax), c32 | 15.065 → 15.037 req/s | -0.2% | 756.8 → 748.6 | -1.1% | -0.4% | -0.04% |
| Graph off, 50% mixed (128 sampled / 128 argmax), c32 | 3.334 → 3.301 req/s | -1.0% | 166.8 → 166.1 | -0.4% | +1.3% | +2.9% |

The homogeneous graph paths and the mixed eager path remain within paired-run noise; the large gain is isolated to mixed batches with Predictor CUDA Graphs enabled.

## Checklist

- [x] Add unit tests.

## CI

model, and

### DIFF
--- a/benchmarks/eval/benchmark_tts_seedtts.py
+++ b/benchmarks/eval/benchmark_tts_seedtts.py
@@ -226,9 +226,7 @@ def _subtalker_dosample_flags(
     ratio: float,
 ) -> dict[str, bool]:
     if not 0.0 <= ratio <= 1.0:
-        raise ValueError(
-            f"--subtalker-dosample-ratio must be in [0, 1], got {ratio}"
-        )
+        raise ValueError(f"--subtalker-dosample-ratio must be in [0, 1], got {ratio}")
     flags: dict[str, bool] = {}
     for index, sample in enumerate(samples):
         dosample = math.floor((index + 1) * ratio) > math.floor(index * ratio)
@@ -296,9 +294,7 @@ def _make_subtalker_dosample_send_fn(
 ) -> SendFn:
     # Note (Shulei He): Qwen3-TTS-only: alternate subtalker_dosample per request for mixed
     # sampled/greedy predictor traffic.
-    dosample_by_id = _subtalker_dosample_flags(
-        samples, config.subtalker_dosample_ratio
-    )
+    dosample_by_id = _subtalker_dosample_flags(samples, config.subtalker_dosample_ratio)
     send_fn_true = make_tts_send_fn(
         config.model,
         api_url,
--- a/benchmarks/eval/benchmark_tts_seedtts.py
+++ b/benchmarks/eval/benchmark_tts_seedtts.py
@@ -304,14 +304,14 @@ def _make_subtalker_dosample_send_fn(
         api_url,
         **common_send_kwargs,
         **generation_kwargs,
-        subtalker_dosample=True,
+        stage_params={"tts_engine": {"subtalker_dosample": True}},
     )
     send_fn_false = make_tts_send_fn(
         config.model,
         api_url,
         **common_send_kwargs,
         **generation_kwargs,
-        subtalker_dosample=False,
+        stage_params={"tts_engine": {"subtalker_dosample": False}},
     )
     send_fn_by_id = {
         sample_id: (send_fn_true if dosample else send_fn_false)
--- a/benchmarks/eval/benchmark_tts_seedtts.py
+++ b/benchmarks/eval/benchmark_tts_seedtts.py
@@ -90,7 +90,12 @@ from pathlib import Path
 from typing import Any
 
 from benchmarks.benchmarker.data import RequestResult
-from benchmarks.benchmarker.runner import BenchmarkRunner, RunConfig, resolve_warmup
+from benchmarks.benchmarker.runner import (
+    BenchmarkRunner,
+    RunConfig,
+    SendFn,
+    resolve_warmup,
+)
 from benchmarks.benchmarker.utils import managed_omni_server
 from benchmarks.dataset.seedtts import SampleInput, load_seedtts_samples
 from benchmarks.metrics.performance import (
@@ -162,6 +167,10 @@ class TtsSeedttsBenchmarkConfig:
     top_k: int | None = None
     repetition_penalty: float | None = None
     seed: int | None = None
+    # Note (Shulei He): Fraction of requests sent with subtalker_dosample=True (rest use False),
+    # deterministically alternated by sample index so runs are reproducible.
+    # None leaves subtalker_dosample unset, i.e. the server-side default applies.
+    subtalker_dosample_ratio: float | None = None
     warmup: int | None = None
     concurrency: int = DEFAULT_TTS_BENCHMARK_CONCURRENCY
     request_rate: float = float("inf")
@@ -212,6 +221,21 @@ def _resolve_warmup(config: TtsSeedttsBenchmarkConfig) -> int:
     return resolve_warmup(config.warmup, config.concurrency)
 
 
+def _subtalker_dosample_flags(
+    samples: list[SampleInput],
+    ratio: float,
+) -> dict[str, bool]:
+    if not 0.0 <= ratio <= 1.0:
+        raise ValueError(
+            f"--subtalker-dosample-ratio must be in [0, 1], got {ratio}"
+        )
+    flags: dict[str, bool] = {}
+    for index, sample in enumerate(samples):
+        dosample = math.floor((index + 1) * ratio) > math.floor(index * ratio)
+        flags[sample.sample_id] = dosample
+    return flags
+
+
 def _build_results_config(
     config: TtsSeedttsBenchmarkConfig,
     *,
@@ -233,6 +257,7 @@ def _build_results_config(
         "sample_offset": config.sample_offset,
         "max_new_tokens": config.max_new_tokens,
         "seed": config.seed,
+        "subtalker_dosample_ratio": config.subtalker_dosample_ratio,
         "token_count": config.token_count,
         "warmup": _resolve_warmup(config),
         "concurrency": config.concurrency,
@@ -262,6 +287,44 @@ def _load_benchmark_samples(config: TtsSeedttsBenchmarkConfig) -> list[SampleInp
     return load_seedtts_samples(config.meta, config.max_samples, split=config.lang)
 
 
+def _make_subtalker_dosample_send_fn(
+    config: TtsSeedttsBenchmarkConfig,
+    api_url: str,
+    common_send_kwargs: dict,
+    generation_kwargs: dict,
+    samples: list[SampleInput],
+) -> SendFn:
+    # Note (Shulei He): Qwen3-TTS-only: alternate subtalker_dosample per request for mixed
+    # sampled/greedy predictor traffic.
+
+    dosample_by_id = _subtalker_dosample_flags(
+        samples, config.subtalker_dosample_ratio
+    )
+    send_fn_true = make_tts_send_fn(
+        config.model,
+        api_url,
+        **common_send_kwargs,
+        **generation_kwargs,
+        subtalker_dosample=True,
+    )
+    send_fn_false = make_tts_send_fn(
+        config.model,
+        api_url,
+        **common_send_kwargs,
+        **generation_kwargs,
+        subtalker_dosample=False,
+    )
+    send_fn_by_id = {
+        sample_id: (send_fn_true if dosample else send_fn_false)
+        for sample_id, dosample in dosample_by_id.items()
+    }
+
+    async def send_fn(session, sample):
+        return await send_fn_by_id[sample.sample_id](session, sample)
+
+    return send_fn
+
+
 async def run_tts_seedtts_benchmark(
     config: TtsSeedttsBenchmarkConfig,
     *,
@@ -286,9 +349,7 @@ async def run_tts_seedtts_benchmark(
         os.makedirs(config.output_dir, exist_ok=True)
 
     generation_kwargs = _build_generation_kwargs(config)
-    send_fn = make_tts_send_fn(
-        config.model,
-        api_url,
+    common_send_kwargs = dict(
         response_format=config.response_format,
         stream=config.stream,
         initial_codec_chunk_frames=config.initial_codec_chunk_frames,
@@ -299,8 +360,15 @@ async def run_tts_seedtts_benchmark(
         task_type=config.task_type,
         instructions=config.instructions,
         save_audio_dir=save_audio_dir,
-        **generation_kwargs,
     )
+    if config.subtalker_dosample_ratio is None:
+        send_fn = make_tts_send_fn(
+            config.model, api_url, **common_send_kwargs, **generation_kwargs
+        )
+    else:
+        send_fn = _make_subtalker_dosample_send_fn(
+            config, api_url, common_send_kwargs, generation_kwargs, samples
+        )
 
     runner = BenchmarkRunner(
         RunConfig(
@@ -390,6 +458,7 @@ def _config_from_args(args: argparse.Namespace) -> TtsSeedttsBenchmarkConfig:
         top_k=args.top_k,
         repetition_penalty=args.repetition_penalty,
         seed=args.seed,
+        subtalker_dosample_ratio=args.subtalker_dosample_ratio,
         warmup=args.warmup,
         concurrency=args.concurrency,
         request_rate=args.request_rate,
@@ -753,6 +822,18 @@ def _build_arg_parser() -> argparse.ArgumentParser:
         type=int,
         default=None,
         help="Per-request sampler seed for reproducible generation.",
+    )
+    parser.add_argument(
+        "--subtalker-dosample-ratio",
+        type=float,
+        default=None,
+        help=(
+            "Optional model-specific (Qwen3-TTS only) fraction of requests "
+            "sent with subtalker_dosample=True (the rest use False), "
+            "deterministically interleaved by sample index. 1.0 = all "
+            "sampled, 0.0 = all greedy, 0.5 = alternating mixed traffic. "
+            "Omit to leave subtalker_dosample unset (server-side default)."
+        ),
     )
     parser.add_argument(
         "--warmup",
--- a/benchmarks/eval/benchmark_tts_seedtts.py
+++ b/benchmarks/eval/benchmark_tts_seedtts.py
@@ -296,7 +296,6 @@ def _make_subtalker_dosample_send_fn(
 ) -> SendFn:
     # Note (Shulei He): Qwen3-TTS-only: alternate subtalker_dosample per request for mixed
     # sampled/greedy predictor traffic.
-
     dosample_by_id = _subtalker_dosample_flags(
         samples, config.subtalker_dosample_ratio
     )
--- a/sglang_omni/models/qwen3_tts/sglang_model.py
+++ b/sglang_omni/models/qwen3_tts/sglang_model.py
@@ -68,8 +68,8 @@ class _PredictorDecodeGraph:
     """CUDA graph over the full per-token predictor chain for one batch bucket.
 
     One graph per (bucket, sampling signature): the signature pins the host
-    branches of the eager sampling path (argmax vs sampled vs mixed, top-k
-    bound, top-p presence), so replay reproduces the eager sampling bits.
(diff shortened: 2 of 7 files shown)


sglang_omni/models/qwen3_tts/sglang_model.py
```diff
@@ -1022,9 +1021,12 @@ def prepare_decode_buffers(self, requests: list[Any]) -> None:
                     "subtalker sampling fields"
                 ) from exc
             semantic_seeds.append(semantic_seed)
-            sub_temperatures.append(subtalker_temperature)
-            sub_top_ps.append(subtalker_top_p)
-            sub_top_ks.append(subtalker_top_k)
+            sub_do_samples.append(do_sample)
+            # Greedy rows still participate in the fixed-shape sampled path for
```
JiaxinD: Repo comment format is `# Note (Name): why`. The why here is that a greedy row's original top_k can be 0 or -1 and would otherwise hit the full-sort branch.

sglang_omni/models/qwen3_tts/sglang_model.py
```diff
@@ -1022,9 +1021,12 @@ def prepare_decode_buffers(self, requests: list[Any]) -> None:
...
+            # mixed batches. Give their discarded sample valid, narrow params.
+            sub_temperatures.append(subtalker_temperature if do_sample else 1.0)
+            sub_top_ps.append(subtalker_top_p if do_sample else 1.0)
+            sub_top_ks.append(subtalker_top_k if do_sample else 1)
```
JiaxinD, line 1035: Greedy rows get top_k=1, so their sampled candidate already collapses to top-1 and the torch.where at 1464 only matters when topk and argmax break ties differently.

sglang_omni/models/qwen3_tts/sglang_model.py
```diff
@@ -1022,9 +1021,12 @@ def prepare_decode_buffers(self, requests: list[Any]) -> None:
...
             sub_seeds.append(subtalker_seed)
             if do_sample:
                 self._sub_sample_rows.append(row_idx)
```
JiaxinD: All remaining reads of _sub_sample_rows are inside this function now, the cross-method consumers went away in this PR. Could be a local `sample_rows`.

sglang_omni/models/qwen3_tts/sglang_model.py
```diff
@@ -1215,44 +1212,35 @@ def _predictor_graph_capture_state(self, bucket_size: int, signature: tuple):
...
         try:
             self._sub_batch_size = bucket_size
-            self._sub_has_sampled_rows = sampled
-            self._sub_sample_count = bucket_size if sampled else 0
-            self._sub_sample_rows = list(range(bucket_size)) if sampled else []
+            self._sub_has_sampled_rows = mode in ("sampled", "mixed")
+            if mode == "sampled":
+                self._sub_sample_count = bucket_size
+            elif mode == "mixed":
+                self._sub_sample_count = max(bucket_size - 1, 1)
```
JiaxinD: This equals bucket_size when bucket_size is 1, so a mixed key would capture the all-sampled branch. Unreachable today (mixed needs bs >= 2), worth an assert plus a why comment.

sglang_omni/models/qwen3_tts/sglang_model.py
```diff
@@ -1304,6 +1292,15 @@ def _predictor_forward_graphed(
         graph = self._predictor_graphs.get(key)
         if graph is None:
             if len(self._predictor_graphs) >= _PREDICTOR_GRAPH_MAX_KEYS:
+                self._predictor_graph_capacity_fallback_count += 1
```
JiaxinD, line 1301: No stats() exit, so the body's "0 graph-capacity fallbacks" is not readable in production. The mixed mode also halves the headroom under the 32-key cap, which is insert-only with no LRU.

sglang_omni/models/qwen3_tts/sglang_model.py
```diff
@@ -1449,31 +1446,26 @@ def _sample_subtalker_token(
...
             layer_idx,
-            row_indices=sampled_rows,
-            semantic_positions=sampled_positions,
+            row_indices=row_indices,
+            semantic_positions=batch_positions,
+        )
+        if self._sub_sample_count == batch_size:
+            return sampled_tokens
+        argmax_tokens = torch.argmax(logits, dim=-1).to(dtype=torch.long)
+        return torch.where(
```
JiaxinD, line 1522: On the eager fallbacks (TP>1, capture disabled, cache full) a mixed batch now sorts the full batch plus an extra argmax, where before it sorted only the sampled subset.

tests/unit_test/qwen3_tts/test_predictor_cuda_graph.py
```diff
@@ -334,19 +373,104 @@ def test_graph_multi_step_replay_bit_identity():
 
 @pytest.mark.accelerator
 @pytest.mark.skipif(not _HAS_CUDA, reason="predictor CUDA graph needs CUDA")
-def test_mixed_sampled_argmax_rows_fall_back_to_eager():
+def test_mixed_sampled_argmax_rows_use_graph_bit_identity():
     device = torch.device("cuda")
     talker = _build_talker(device)
+    with torch.no_grad():
+        for head in talker.code_predictor.lm_head:
+            head.proj.weight.zero_()
```
JiaxinD: Zeroing every lm_head makes all non-layer0 logits identical, so both torch.equal below compare all-zeros against all-zeros.

tests/unit_test/qwen3_tts/test_predictor_cuda_graph.py
```diff
@@ -334,19 +373,104 @@ def test_graph_multi_step_replay_bit_identity():
...
     eager_codes, eager_embeds = _run_eager(talker, layer0, hidden, positions)
     graph_codes, graph_embeds = _run_forward(talker, layer0, hidden, positions)
 
-    assert not talker._predictor_graphs, "mixed batches must not capture a graph"
+    assert talker._predictor_graphs, "mixed batch did not capture a graph"
     assert torch.equal(graph_codes, eager_codes)
     assert torch.equal(graph_embeds, eager_embeds)
+    assert (
+        torch.count_nonzero(graph_codes[1, 1:]) == 0
+    ), "argmax row must choose the first token from exact predictor-logit ties"
```
JiaxinD: Measured on torch 2.12: topk(...).indices[:,0] equals argmax for both constructions used here, so this passes with the do_sample mask removed. A sentinel-returning monkeypatch on _sample_subtalker_token_seeded would discriminate.

tests/unit_test/qwen3_tts/test_predictor_cuda_graph.py
```diff
@@ -334,19 +373,104 @@ def test_graph_multi_step_replay_bit_identity():
...
+    logits[0, 3] = 9.0
+    logits[1, 5] = 9.0
+    logits[1, 7] = 9.0
+
+    tokens = talker._sample_subtalker_token(
+        logits,
+        semantic_positions=torch.zeros(2, dtype=torch.long, device=device),
+    )
+
+    assert tokens[1].item() == torch.argmax(logits[1]).item() == 5
```
JiaxinD, line 435: Same as above, this holds whether or not the mask is applied, so it does not yet guard the mixed path.

tests/unit_test/qwen3_tts/test_predictor_cuda_graph.py
```diff
@@ -779,33 +943,39 @@ def test_top_p_removed_ranks_never_sampled():
...
+def test_capture_state_body_failure_restores_state():
+    """An exception inside capture must restore the live sampling state."""
     device = torch.device("cpu")
     talker = _build_talker(device)
-    talker.prepare_decode_buffers(_uniform_requests(3))
+    talker.prepare_decode_buffers(_uniform_requests(3, dosample=False))
     saved = (
         talker._sub_batch_size,
         talker._sub_sample_count,
         list(talker._sub_sample_rows),
```
JiaxinD: _sub_sample_rows is no longer saved or restored by the context manager, so this element of the tuple is always equal.

JiaxinD (commented):
Sampling equivalence holds, the seeded kernel gathers params by the value of row_indices so the old compressed pairing was correct and tokens are unchanged. Could the forwarding adapter land here rather than stay out of tree? `stage_params` already reaches the builder and ming_tts/request_builders.py:112-117 is the pattern to copy. Without it `subtalker_dosample` has no public entry (default True), so mixed batches cannot occur in production, the numbers are not reproducible, and e2e CI never exercises the new path.

JiaxinD:
The all-argmax leg shows +1.9% tok/req while the other five legs are within 0.4%. If that path is unchanged there should be no drift. A per-request token count diff at seed 1234 settles it.

JiaxinD:
`git log --merges origin/main..HEAD` is not empty after 347089c4. This repo is rebase-only, could you rebase onto main instead?

sglang_omni/models/qwen3_tts/sglang_model.py
leihehehe: Fixed.

sglang_omni/models/qwen3_tts/sglang_model.py
leihehehe: Replaced sample_rows with `_sub_sample_row_indices_tensor` to keep mixed-row selection shape-stable and CUDA Graph replayable

sglang_omni/models/qwen3_tts/sglang_model.py
leihehehe: Mixed key was removed so this issue is resolved now

sglang_omni/models/qwen3_tts/sglang_model.py
leihehehe: Capacity fallback is now observable through a one-time [warning](https://github.com/sgl-project/sglang-omni/pull/1726/changes#diff-401b8a23ba297eeae4b1d40cbfe90b541128e270277b50ce3c83cfd9ef00f46aR1301-R1309) when the cap is reached.

Mixed and fully sampled batches now share the same  `sampled`  CUDA Graph key, so it preserves more headroom. I have not introduced LRU eviction for now considering now we only have two graphs, but happy to add it if you think it's necessary 😃

sglang_omni/models/qwen3_tts/sglang_model.py
leihehehe: Fixed by separating the [graph-safe](https://github.com/sgl-project/sglang-omni/pull/1726/changes#diff-401b8a23ba297eeae4b1d40cbfe90b541128e270277b50ce3c83cfd9ef00f46aR1493-R1527) and eager sampling paths. The eager mixed fallback still runs argmax once over the full batch, but sorts/samples only the indexed sampled-row subset and scatters those results back. Full-batch sampling is used only inside CUDA Graph capture/replay.

tests/unit_test/qwen3_tts/test_predictor_cuda_graph.py
leihehehe: fixed

tests/unit_test/qwen3_tts/test_predictor_cuda_graph.py
leihehehe: fixed

tests/unit_test/qwen3_tts/test_predictor_cuda_graph.py
leihehehe: fixed

tests/unit_test/qwen3_tts/test_predictor_cuda_graph.py
leihehehe: removed

leihehehe:
Hi @JiaxinD, thanks for the thorough review! I’ve addressed the comments, updated the PR description with the latest implementation and benchmark results, and added draft PR #1764 to make the benchmark-only base adapter reviewable and reproducible. Could you take a  look? Thanks!

tests/unit_test/qwen3_tts/test_predictor_cuda_graph.py
```diff
@@ -422,15 +573,16 @@ def test_capture_failure_restores_live_sub_state(monkeypatch: pytest.MonkeyPatch
     class _BoomGraph:
         def __init__(self, model, bucket_size, signature, **kwargs) -> None:
             del kwargs
-            with model._predictor_graph_capture_state(bucket_size, signature):
+            with model._predictor_graph_capture_state(4, ("argmax", 0, False, False)):
                 raise RuntimeError("simulated capture failure")
```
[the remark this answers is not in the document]
leihehehe: Fixed

JiaxinD (approved):

benchmarks/eval/benchmark_tts_seedtts.py
```diff
@@ -262,6 +285,41 @@ def _load_benchmark_samples(config: TtsSeedttsBenchmarkConfig) -> list[SampleInp
...
+    api_url: str,
+    common_send_kwargs: dict,
+    generation_kwargs: dict,
+    samples: list[SampleInput],
+) -> SendFn:
+    # Note (Shulei He): Qwen3-TTS-only: alternate subtalker_dosample per request for mixed
+    # sampled/greedy predictor traffic.
+    dosample_by_id = _subtalker_dosample_flags(samples, config.subtalker_dosample_ratio)
+    send_fn_true = make_tts_send_fn(
+        config.model,
```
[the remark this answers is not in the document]
leihehehe: The helper has only one private call site, and it is reached only from the `else` branch after checking `subtalker_dosample_ratio` is None , so `None` is structurally unreachable here. I’d prefer not to add a redundant guard unless we expect this helper to be called independently later.

Issue #1233, opened 2026-07-30
[Tracking] MOSS-TTS Optimizations


CloudRipple:
## Motivation

MOSS-TTS Delay still has significant room for performance improvement. We have implemented a series of optimizations, and this issue tracks their progress as separate PRs.

## Tracking

All items listed below **have already been implemented**. The checkbox tracks whether the corresponding PR has been merged, not whether implementation is complete.

The optimizations are submitted as separate PRs so that each change, performance gain, and quality impact can be reviewed and validated independently.

The performance numbers below come from different points in the development stack. They are useful directional evidence, but they are not cumulative. Each PR will be rerun against its immediate parent before submission.

### Codec ownership and preprocessing

- [x] **Decouple the MOSS-TTS Delay codec and run reference encoding on GPU** — #1222
  - Load the processor with `audio_tokenizer=None` and give preprocessing and vocoder standalone codec ownership.
  - Preserve checkpoint codec metadata, explicit codec overrides, `PathLike` references, and unified path/URL/data-URI audio loading.
  - Rename `codec.py` to `delay_pattern.py` because it contains Delay-pattern logic rather than neural codec execution.
  - Derive the codec device from stage placement while preserving an explicit CPU fallback.
  - Update MOSS-TTS process resource accounting so opt-in stage isolation does not require a manual preprocessing memory-fraction workaround.
  - Full SeedTTS result: `2.9958 -> 4.4395 QPS` (`+48.19%`).

- [x] **Batch and cache reference encoding** — #1348 
  - Route MOSS-TTS Delay reference encoding through the shared `ReferenceEncodeService`, with waveform-content cache identity, same-key single-flight deduplication, byte-bounded LRU caching, and caller-isolated cached tensors.
  - Batch prepared GPU waveforms across different cold references while keeping audio loading and resampling outside the codec worker; retry failed batches per item and close the worker with the preprocessing lifecycle.
  - Five-run full SeedTTS EN result on a single NVIDIA H200 at concurrency 16 (`1088/1088` samples per run): throughput `6.8686 -> 9.0000 QPS` (`+31.03%`); mean latency `2.3228 -> 1.7698 s` (`-23.81%`); P95 latency `2.9000 -> 2.1930 s` (`-24.38%`); mean RTF `0.54038 -> 0.41056` (`-24.02%`).

### AR sampling

- [x] **Restrict audio-state text sampling to the two valid control tokens** — #1158
[...]

PR #1727, opened 2026-08-25
[Perf][MOSS-TTS] Load MOSS-Audio-Tokenizer components per pipeline stage

CloudRipple:
## Motivation

MOSS-TTS Delay and MOSS-TTS Local use different parts of the same MOSS-Audio-Tokenizer checkpoint in each pipeline stage. Reference preprocessing needs the encoder and quantizer, while waveform generation needs the quantizer and decoder. The previous loading path materialized a complete codec in more than one stage, so every process paid for parameters that it never executed. That reduced the memory available to the autoregressive engine and its KV cache, and it also kept the v1 and v2 consumers on separate codec-loading paths.

This PR introduces one repository-owned MOSS-Audio-Tokenizer implementation and loads only the component prefixes required by the active stage. It keeps the decoder and streaming state behavior intact while making attention backend and dtype policy explicit. The goal is to reduce resident codec memory without changing generated audio or the consumer-facing pipeline contracts.

## Modifications

- Add repository-owned MOSS-Audio-Tokenizer encoder and vocoder implementations in `sglang_omni/models/moss_tts/audio_tokenizer.py`.
- Reuse the shared transformer, decoder, quantizer, RoPE, and attention wrappers for both the MOSS-Audio-Tokenizer v1 weight layout used by MOSS-TTS Delay and the v2 layout used by MOSS-TTS Local.
- Construct codec modules from configuration or source modules and normalize v1 weight names into the shared module layout.
- Load only `encoder.*` plus `quantizer.*` for reference preprocessing, and only `decoder.*` plus `quantizer.*` for vocoder stages, using prefix-aware weight loading instead of executing a complete remote-code codec.
- Materialize selected encoder/decoder parameters in the configured compute dtype at load time. Quantizer and explicitly FP32 parameters remain FP32; FP16 support is intentionally outside this PR.
- Thread explicit `auto`, `packed_flash_attention`, and `sdpa` backend policy through both MOSS-TTS consumers. Streaming codec sessions force SDPA because the persistent streaming FlashAttention path is not graph-safe/reliable; non-streaming decode retains the packed fast path and its SM-based tiling fallback.
- Give each consumer stage ownership of its encoder/vocoder lifecycle and close reference-encoder workers during pipeline cleanup.
- Update Local streaming and non-streaming vocoder construction to use the shared decoder loader, and extend unit tests and MOSS-TTS documentation for the new loading and dtype contracts.

## Related Issues

#1233 

## Accuracy Test

| Item | Value |
| --- | --- |
| Models | MOSS-TTS Delay v1.5; MOSS-TTS Local Transformer v1.5 |
| Dataset | SeedTTS EN full set, 1,088 samples per round |
| Concurrency | 16 |
| Seed | 42 |
| Hardware | NVIDIA H200 |
| Quality generation runs | 30 (5 base/modified rounds per affected path) |
| Quality requests | 32,640 |
| Quality completed / failed | 32,640 / 0 |

The quality campaign evaluated WER with Qwen3-ASR-1.7B, speaker similarity with WavLM-SV, UTMOS, and WAV integrity on the exact base and modified revisions. Every quality run evaluated 1,088/1,088 samples; all WAVs were finite, nonzero, 48 kHz audio.

The table below reports five-round means; each delta is the paired modified-minus-base mean.

| Model / mode | Base WER | Modified WER | WER delta | Base speaker similarity | Modified speaker similarity | Similarity delta | Base UTMOS | Modified UTMOS | UTMOS delta |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| MOSS-TTS Delay non-streaming | 2.108% | 2.408% | +0.300 pp | 68.992 | 69.099 | +0.107 | 3.9460 | 3.9441 | -0.0019 |
| MOSS-TTS Local non-streaming | 2.150% | 2.239% | +0.089 pp | 65.300 | 65.193 | -0.107 | 3.9644 | 3.9709 | +0.0065 |
| MOSS-TTS Local streaming | 1.949% | 2.110% | +0.161 pp | 65.171 | 65.182 | +0.011 | 3.9644 | 3.9722 | +0.0078 |

All three paths pass the paired quality gates: WER delta <= 0.3 percentage points, speaker-similarity delta >= -0.5, UTMOS delta >= -0.05, with complete generation and audio-integrity checks.

The following are five-round means from the same H200, concurrency-16 A/B.

### Throughput

| Model / mode | Base rounds (req/s) | Modified rounds (req/s) | Base mean | Modified mean | Delta |
| --- | --- | --- | ---: | ---: | ---: |
| Delay non-streaming | 8.929, 8.966, 8.988, 8.863, 8.958 | 9.139, 9.250, 9.154, 9.205, 9.118 | 8.9408 | 9.1732 | +2.60% |
| Local non-streaming | 11.885, 11.980, 11.687, 11.854, 12.011 | 14.541, 14.641, 14.535, 14.524, 14.574 | 11.8834 | 14.5630 | +22.55% |
| Local streaming | 6.709, 6.898, 6.830, 6.836, 6.691 | 8.269, 8.318, 8.416, 8.440, 8.449 | 6.7928 | 8.3784 | +23.34% |

Mean audio throughput improved by `+2.57%` for Delay, `+22.55%` for Local non-streaming, and `+23.34%` for Local streaming. Mean latency improved by `2.51%`, `18.40%`, and `18.94%`, respectively.

Mean RTF changed from `0.4115` to `0.4010` for Delay (`-2.56%`), from `0.3154` to `0.2523` for Local non-streaming (`-20.00%`), and from `0.5506` to `0.4393` for Local streaming (`-20.21%`).

### Streaming behavior

| Metric | Base mean | Modified mean | Delta |
| --- | ---: | ---: | ---: |
| TTFC (s) | 1.1101 | 0.5379 | -51.54% |
| Inter-chunk interval (s) | 0.1830 | 0.1734 | -5.26% |
| Max underrun (s) | 0.0126 | 0.0062 | -50.64% |
| C50 | 92.26% | 95.96% | +3.70 pp |
| C100 | 95.31% | 97.59% | +2.28 pp |
| C200 | 98.14% | 99.34% | +1.19 pp |

### Peak GPU memory

The selective loader releases codec allocation, but the serving process uses most of that headroom to enlarge the AR KV cache. Therefore the complete-service NVML peak reduction is smaller than the standalone codec reduction.

| Model / mode | Base mean (MiB) | Modified mean (MiB) | Delta (MiB) | Delta (GiB) |
| Delay non-streaming | 139,123.0 | 133,169.8 | -5,953.2 | -5.81 |
| Local non-streaming | 129,547.0 | 126,561.4 | -2,985.6 | -2.92 |
| Local streaming | 128,852.6 | 125,841.4 | -3,011.2 | -2.94 |

### KV-cache accounting

SGLang's logged `GB` values use binary GiB (`1024^3` bytes). The KV allocation is stable across all five rounds of each path:

| Path | KV tokens base -> modified | K+V base -> modified | KV growth / equivalent memory |
| --- | ---: | ---: | ---: |
| Delay non-streaming | 750,038 -> 784,897 (+34,859) | 103.00 -> 107.78 GiB | +4.78 GiB; about +4.79 GiB from token scaling |
| Local non-streaming | 740,595 -> 755,827 (+15,232) | 101.70 -> 103.80 GiB | +2.10 GiB; about +2.09 GiB from token scaling |
| Local streaming | 740,595 -> 755,827 (+15,232) | 101.70 -> 103.80 GiB | +2.10 GiB; about +2.09 GiB from token scaling |

For Local, the vocoder process allocation decreases from `8.57 GiB` to `6.44 GiB` (`-2.13 GiB`). The AR process profile decreases from `13.42 GiB` to `11.33 GiB`, while `available_for_kv` increases from `101.71 GiB` to `103.80 GiB`; approximately `2.09 GiB` is consumed by the larger KV cache. For Delay, the pre-KV available memory increases from `114.90 GiB` to `120.16 GiB`, and approximately `4.79 GiB` becomes additional KV capacity.

### Profile attribution

An isolated CUDA ablation confirms that the reference encoder's packed FlashAttention path is the primary source of the Local throughput improvement. Load-time compute-dtype materialization removes additional cast/copy work. Streaming continues to use SDPA by design, while selective loading primarily improves codec residency and KV-cache capacity.

## Checklist

- [x] Add unit tests.

## CI

model, and

### DIFF
--- a/sglang_omni/models/moss_tts/audio_tokenizer.py
+++ b/sglang_omni/models/moss_tts/audio_tokenizer.py
@@ -1203,16 +1203,10 @@ class MossAudioTokenizerEncoder(nn.Module):
         config: dict[str, Any],
         *,
         parameter_device: str | torch.device | None = None,
-        encoder_dtype: torch.dtype = torch.bfloat16,
         compute_dtype: torch.dtype | None = None,
         attention_backend: str = AUTO_ATTENTION_BACKEND,
     ) -> None:
         super().__init__()
-        _validate_audio_dtypes(
-            component_dtype=encoder_dtype,
-            component_name="encoder_dtype",
-            compute_dtype=compute_dtype,
-        )
         self.config = SimpleNamespace(**config)
         sampling_rate = config.get("sampling_rate") or config.get("sample_rate")
         if sampling_rate is None:
@@ -1223,7 +1217,7 @@ class MossAudioTokenizerEncoder(nn.Module):
         self.enable_channel_interleave = bool(
             config.get("enable_channel_interleave", self.number_channels > 1)
         )
-        configured_compute_dtype = str(config.get("compute_dtype", "bf16"))
+        configured_compute_dtype = str(config.get("compute_dtype") or "bfloat16")
         resolved_compute_dtype = {
             "bf16": torch.bfloat16,
             "bfloat16": torch.bfloat16,
@@ -1242,6 +1236,21 @@ class MossAudioTokenizerEncoder(nn.Module):
         requested_compute_dtype = (
             resolved_compute_dtype if compute_dtype is None else compute_dtype
         )
+        if requested_compute_dtype not in (
+            None,
+            torch.float32,
+            torch.bfloat16,
+        ):
+            raise ValueError(
+                "compute_dtype must be torch.float32, torch.bfloat16, or None; "
+                f"got {requested_compute_dtype!r}"
+            )
+        effective_encoder_dtype = (
+            torch.float32
+            if requested_compute_dtype is None
+            or requested_compute_dtype is torch.float32
+            else requested_compute_dtype
+        )
         self.compute_dtype = (
             None
             if requested_compute_dtype is torch.float32
@@ -1251,12 +1260,7 @@ class MossAudioTokenizerEncoder(nn.Module):
         # for the encoder weights. This avoids relying on autocast to recast
         # FP32 parameters on every request. The quantizer is intentionally kept
         # FP32 below.
-        self.encoder_dtype = (
-            torch.float32
-            if requested_compute_dtype is None
-            or requested_compute_dtype is torch.float32
-            else requested_compute_dtype
-        )
+        self.encoder_dtype = effective_encoder_dtype
         self.attention_backend = validate_attention_backend(attention_backend)
         self._uses_moss_audio_tokenizer_v1_weights = "number_channels" not in config
 
@@ -1667,7 +1671,6 @@ def load_moss_audio_encoder(
     model_path: str,
     *,
     device: str | torch.device,
-    encoder_dtype: torch.dtype = torch.bfloat16,
     compute_dtype: torch.dtype | None = None,
     attention_backend: str = AUTO_ATTENTION_BACKEND,
 ) -> MossAudioEncoder:
@@ -1678,7 +1681,6 @@ def load_moss_audio_encoder(
     model = MossAudioTokenizerEncoder(
         config,
         parameter_device="meta",
-        encoder_dtype=encoder_dtype,
         compute_dtype=compute_dtype,
         attention_backend=attention_backend,
     )
--- a/sglang_omni/models/moss_tts/audio_tokenizer.py
+++ b/sglang_omni/models/moss_tts/audio_tokenizer.py
@@ -71,6 +71,21 @@ def resolve_moss_audio_attention_backend(
     return AUTO_ATTENTION_BACKEND
 
 
+# Note (Zhang Yiyang): Prefer the runtime model value, then the canonical config
+# field, and finally the legacy config alias for checkpoint compatibility.
+def resolve_moss_audio_sample_rate(model: Any, config: Any) -> int:
+    for value in (
+        getattr(model, "sampling_rate", None),
+        getattr(config, "sampling_rate", None),
+        getattr(config, "sample_rate", None),
+    ):
+        if value is not None:
+            return int(value)
+    raise ValueError(
+        "MOSS-Audio-Tokenizer model/config lacks sampling_rate or sample_rate"
+    )
+
+
 def _attention_backend_label(resolution: AttentionBackendResolution) -> str:
     if resolution.fallback_reason is None:
         return resolution.backend
@@ -1456,9 +1471,7 @@ class MossAudioEncoder:
         self.model = model
         self.device = str(device)
         config = model.config
-        self.sample_rate = int(
-            getattr(model, "sampling_rate", getattr(config, "sampling_rate"))
-        )
+        self.sample_rate = resolve_moss_audio_sample_rate(model, config)
         self.number_channels = int(
             getattr(model, "number_channels", getattr(config, "number_channels", 1))
         )
--- a/sglang_omni/models/moss_tts/audio_tokenizer.py
+++ b/sglang_omni/models/moss_tts/audio_tokenizer.py
@@ -3,19 +3,24 @@
 
 from __future__ import annotations
 
+import json
 import logging
-from collections.abc import Iterator, Sequence
-from contextlib import nullcontext
+import math
+from collections.abc import Sequence
+from dataclasses import dataclass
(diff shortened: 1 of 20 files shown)


sglang_omni/models/moss_tts/audio_tokenizer.py
```diff
@@ -468,115 +806,984 @@ def create_sin_embedding(
...
+        # note (Zhang Yiyang): The compute policy is also the materialized dtype
+        # for the encoder weights. This avoids relying on autocast to recast
+        # FP32 parameters on every request. The quantizer is intentionally kept
+        # FP32 below.
+        self.encoder_dtype = (
+            torch.float32
+            if requested_compute_dtype is None
+            or requested_compute_dtype is torch.float32
+            else requested_compute_dtype
+        )
```
luojiaxuan: `encoder_dtype` is validated in `_validate_audio_dtypes` but never used after that. `self.encoder_dtype` only depends on the compute policy, so the factory `dtype="float32"` has no effect on this path.

Also, the three loaders handle `compute_dtype=None` differently. This encoder and `load_moss_tts_local_audio_vocoder` fall back to the config's `compute_dtype`, which is bf16 in practice. `MossAudioTokenizerVocoder` at L1740 falls back to the `decoder_dtype` arg instead and ignores the config, so it ends up fp32 with the factory defaults. Same args, different weight dtype per stage.

Not a bug today, the stage factories always pass `compute_dtype="bfloat16"` explicitly. But for a new public loader API, I suggest either making `encoder_dtype` actually work when `compute_dtype` is None, same as the vocoder, or removing the dead `encoder_dtype` / `decoder_dtype` / factory `dtype` params and keeping `compute_dtype` as the only knob.

sglang_omni/models/moss_tts/audio_tokenizer.py
```diff
@@ -468,115 +806,984 @@ def create_sin_embedding(
...
+
+    def __init__(self, model: MossAudioTokenizerEncoder, *, device: str) -> None:
         self.model = model
         self.device = str(device)
-        self.dtype = _model_floating_dtype(model)
-        self.sample_rate = int(model.config.sampling_rate)
+        config = model.config
+        self.sample_rate = int(
+            getattr(model, "sampling_rate", getattr(config, "sampling_rate"))
+        )
```
luojiaxuan: `getattr` evaluates its default eagerly, so the inner `getattr(config, "sampling_rate")` raises AttributeError when config lacks this attribute, even if `model.sampling_rate` exists. The fallback never works. `MossAudioTokenizerEncoder` explicitly accepts configs that only have `sample_rate`, but this wrapper crashes on exactly that kind of checkpoint. Current v1/v2 configs have both keys, so nothing breaks today, still a one-line fix:

```suggestion
        self.sample_rate = int(
            getattr(model, "sampling_rate", None)
            or getattr(config, "sampling_rate", None)
            or getattr(config, "sample_rate")
        )
```

Same pattern in `moss_tts_local/audio_tokenizer.py` L54 and L204.

luojiaxuan (approved):
LGTM

sglang_omni/models/moss_tts/audio_tokenizer.py
CloudRipple: Thanks for catching this. I removed the redundant `encoder_dtype` input and the preprocessing factory `dtype` path. `compute_dtype` is now the single dtype control for encoder weight materialization and execution, and the MOSS-TTS YAML configs default it to `bfloat16`. The decoder-specific `decoder_dtype` behavior is unchanged in this patch.

sglang_omni/models/moss_tts/audio_tokenizer.py
CloudRipple: Fixed in 76dee2da. I added a shared `resolve_moss_audio_sample_rate` helper and use it in all three wrappers. It resolves `model.sampling_rate`, then `config.sampling_rate`, then the legacy `config.sample_rate`, and raises a clear error when none is available. Tests were added for precedence, fallback, and missing fields.

PR #1729, opened 2026-08-25
[Fix][Tests] Restore scheduler module stubs after config refactor

ywgrit:
## Summary

- restore Qwen3-TTS test helper probes and fake module registrations to `sglang.srt.managers.scheduler`
- restore the same scheduler stub pair in the Ming-Omni thinker tests
- keep the config refactor's legitimate `factory_path` fields unchanged

## Root cause

The configuration refactor in #1494 applied the stage-factory rename to unrelated
test-only SGLang module strings. `sglang.srt.managers.factory_path` is not a
module; production code in both paths still imports scheduler types from
`sglang.srt.managers.scheduler`.

For Qwen3-TTS, the failed probe selected incomplete fallback stubs, which then
caused existing model imports to fail at `ScheduleBatch`. Ming-Omni's fake
`GenerationBatchResult` module was registered under the same nonexistent path
and failed when the runner imported the real scheduler path.

## Validation

Before the fix:

```text
Qwen3-TTS: ImportError: cannot import name 'ScheduleBatch'
Ming-Omni: ModuleNotFoundError: No module named 'sglang.srt.managers.scheduler'
```

After restoring the Qwen3-TTS stubs:

```text
1 passed, 17 warnings
131 passed, 5 deselected, 17 warnings
```

For the added Ming-Omni pair, Python compilation, source consistency, and diff
checks pass locally. The approved upstream CI provides the full dependency
environment. No production runtime code is changed.

### DIFF
--- a/tests/unit_test/ming_omni/test_thinker.py
+++ b/tests/unit_test/ming_omni/test_thinker.py
@@ -282,7 +282,7 @@ def test_ming_config_and_stages_do_not_import_ming_runner() -> None:
 def _load_runner_with_fake_sglang(monkeypatch):
     torch = pytest.importorskip("torch")
 
-    scheduler_module = ModuleType("sglang.srt.managers.factory_path")
+    scheduler_module = ModuleType("sglang.srt.managers.scheduler")
 
     class GenerationBatchResult:
         def __init__(self, **kwargs):
@@ -296,7 +296,7 @@ def _load_runner_with_fake_sglang(monkeypatch):
     )
     monkeypatch.setitem(
         sys.modules,
-        "sglang.srt.managers.factory_path",
+        "sglang.srt.managers.scheduler",
         scheduler_module,
     )
 
--- a/tests/unit_test/qwen3_tts/test_pipeline.py
+++ b/tests/unit_test/qwen3_tts/test_pipeline.py
@@ -54,8 +54,8 @@ from tests.unit_test.fakes import FakeExecutionBridge, FakeServerArgs
 
 def install_fake_sglang(monkeypatch: pytest.MonkeyPatch) -> None:
     try:
-        import sglang.srt.managers.factory_path  # noqa: F401
         import sglang.srt.managers.schedule_batch  # noqa: F401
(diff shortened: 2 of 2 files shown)


ywgrit:
@MayDomine, could you please review this small regression fix? It restores the test-only scheduler module references that were accidentally changed during the config refactor in #1494. I reproduced the import failure on current `main` before the patch and verified the Qwen3-TTS non-accelerator unit suite after the fix (131 passed, 5 deselected).

@shuaills, could you please approve the first-time GitHub Actions runs and add the `run-ci` and `run-qwen3-tts` labels to trigger the upstream Qwen3-TTS CI? GitHub does not allow a fork author to request reviewers or add these labels directly.

MayDomine (commented):
It seems that the config refactor's factory rename was applied too broadly and hit these sglang stub module names, which are unrelated to the config surface.

A sweep for the same damage pattern found one more occurrence: `tests/unit_test/ming_omni/test_thinker.py` lines 285 and 299 stub `sglang.srt.managers.factory_path` where `sglang.srt.managers.scheduler` is meant (same `ModuleType` + `monkeypatch.setitem` pair). Production imports of `sglang.srt.managers.scheduler` are all intact. Could you include that file in this PR so the regression is fixed in one place? LGTM otherwise.

MayDomine (commented):
Already approved the ci.

MayDomine (approved):

ywgrit:
Included the Ming-Omni `ModuleType` and `sys.modules` scheduler stub pair in commit `060166834`. Both now use `sglang.srt.managers.scheduler`, and the PR title/description cover both affected test helpers. I reproduced the pre-fix Ming runner `ModuleNotFoundError`; Python compilation, source consistency, and diff checks pass after the change.

MayDomine (approved):
Both affected test helpers now restore the `sglang.srt.managers.scheduler` stub spelling, matching their pre-refactor state; a repo-wide sweep for the same rename pattern found no further occurrences and all production imports of that module are intact. Thanks for catching and fixing this.

PR #1730, opened 2026-08-25
[Apple] Add Qwen3-ASR MLX and Torch MPS paths

yeahdongcn:
## Summary

Adds Apple Silicon Qwen3-ASR pseudo-streaming through native MLX and Torch MPS paths, without an `mlx-audio` runtime dependency.

## Implementation

- Add the Apple Metal Omni platform, Darwin arm64 dependency markers, FFmpeg 7 setup, and an idempotent installer.
- Add a local MLX Qwen3-ASR audio encoder/decoder with lazy single-request decode and SSE deltas.
- Add a Torch MPS compatibility runner using the Hugging Face Qwen3 language model plus Omni's Torch audio path.
- Adapt Omni's scheduler bridge to the pinned SGLang MLX worker/runner contract.

The two Apple paths are intentional: pinned SGLang separates default Torch/MPS from the `SGLANG_USE_MLX=1` worker/runner, while Omni bypasses the upstream worker-selection entry point and therefore adapts both contracts.

## Performance

MacBook Pro M1 16 GB, `query_to_cars.wav`, concurrency 1, 2 warmups + 15 measured requests; all rows produced 15/15 exact transcripts.

### SGLang 0.5.16

MLX 0.32.1 and Torch 2.11:

| Path | Checkpoint | Language | TTFT median | Total median |
| --- | --- | --- | ---: | ---: |
| Omni MLX | Qwen3-ASR 0.6B 4-bit | auto | 0.183 s | 0.252 s |
| mlx-audio 0.5.0 | Qwen3-ASR 0.6B 4-bit | auto | 0.241 s | 0.309 s |
| Omni MLX | Official Qwen3-ASR 0.6B BF16 | auto | 0.251 s | 0.449 s |
| Omni Torch MPS | Official Qwen3-ASR 0.6B BF16 | forced English | 0.208 s | 0.552 s |

### SGLang 0.5.18

MLX 0.32.2, mlx-lm 0.31.3, and Torch 2.11:

| Path | Checkpoint | Language | TTFT median | Total median |
| --- | --- | --- | ---: | ---: |
| Omni MLX | Qwen3-ASR 0.6B 4-bit | auto | 0.193 s | 0.267 s |
| mlx-audio 0.5.0 | Qwen3-ASR 0.6B 4-bit | auto | 0.244 s | 0.313 s |
| Omni MLX | Official Qwen3-ASR 0.6B BF16 | auto | 0.264 s | 0.485 s |
| Omni Torch MPS | Official Qwen3-ASR 0.6B BF16 | auto | 0.483 s | 1.013 s |

On the current pinned stack, Omni MLX 4-bit is about 14.8% faster in total latency and 20.9% faster in TTFT than the same-checkpoint mlx-audio baseline.

The two version groups are separate end-to-end runs; mlx-audio used isolated environments with the listed MLX stack and does not depend on SGLang.

The Torch rows are not a version A/B because the 0.5.16 run forced English
while the 0.5.18 run used auto detection, which decodes three additional
language-marker tokens before the first transcript delta.

### Controlled version check

An A/B/B/A rerun used identical auto-language requests, five warmups, and 15
measurements per run. After fixing server teardown, system swap stayed near
10 GB across all runs. Values below are the mean of the two run medians:

| Path | SGLang 0.5.16 TTFT / total | SGLang 0.5.18 TTFT / total | 0.5.18 total change |
| --- | ---: | ---: | ---: |
| Omni MLX 4-bit | 0.346 / 0.484 s | 0.344 / 0.464 s | -4.2% |
| Omni MLX BF16 | 0.444 / 0.842 s | 0.441 / 0.834 s | -1.0% |
| Omni Torch MPS BF16 | 0.598 / 1.251 s | 0.594 / 1.244 s | -0.6% |

The controlled check shows no version-caused regression. Its absolute values
are excluded from the headline snapshot because the machine was under stable
but high swap pressure.

## Tests

- Clean Apple installer and `uv pip check` with SGLang 0.5.18.
- `SGLANG_USE_MLX=1`: Apple/MLX/request/streaming/pipeline focused matrix — 642 passed.
- `SGLANG_USE_MLX=0`: full Qwen3-ASR, Apple platform, serving, and base-runner regression matrix — 801 passed, 3 skipped.
- Real MLX 4-bit/BF16 and Torch MPS BF16 HTTP/SSE smoke and performance runs.
- `SKIP=no-commit-to-branch pre-commit run --all-files` and `git diff --check`.

## Shared Qwen3-ASR follow-up

The prompt and `language None` parsing issue found during independent Apple Silicon validation was fixed for all backends by merged [PR #1781](https://github.com/sgl-project/sglang-omni/pull/1781), since it affects the shared Qwen3-ASR path (including CUDA).

## Latest review follow-ups

- Declare `mlx` and `mlx-lm` as direct Darwin arm64 dependencies; the supported install still obtains SGLang from its source `all_mps` extra.
- Reuse the SGLang revision-aware MLX checkpoint resolver and remote-code gate, and make pipeline tests backend-explicit.
- Resolve a backend-specific audio policy: Torch MPS now fails fast above its qualified 60-second range, while MLX and CUDA retain the native 1,200-second streaming policy.
- Track the shared page-reserve admission clamp separately in [PR #1788](https://github.com/sgl-project/sglang-omni/pull/1788).

Rebased-head smoke tests covered MLX 4-bit/BF16 and Torch MPS BF16 HTTP/SSE transcription, empty audio, and Torch MPS long-audio HTTP 400 handling. The performance tables remain the earlier stack snapshots; they were not rebenchmarked after the shared five-token prompt alignment in #1781.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,313 @@
+#!/usr/bin/env bash
+# Install sglang-omni and the Apple Silicon Qwen3-ASR runtime.
+#
+# This file intentionally does not bootstrap Homebrew. Homebrew's official
+# bootstrapper can require administrator approval and changes system state;
+# users should install it from https://brew.sh and then rerun this script.
+set -Eeuo pipefail
+IFS=$'\n\t'
+
+readonly DEFAULT_OMNI_REPO="https://github.com/sgl-project/sglang-omni.git"
+readonly DEFAULT_OMNI_REF="main"
+readonly DEFAULT_SGLANG_REPO="https://github.com/sgl-project/sglang.git"
+readonly DEFAULT_SGLANG_REF="v0.5.16"
+
+SCRIPT_PATH="${BASH_SOURCE[0]:-}"
+if [[ -n "$SCRIPT_PATH" && -f "$SCRIPT_PATH" ]]; then
+  readonly SCRIPT_DIR="$(cd -- "$(dirname -- "$SCRIPT_PATH")" && pwd -P)"
+else
+  readonly SCRIPT_DIR=""
+fi
+
+readonly OMNI_REPO="${SGLANG_OMNI_REPO:-$DEFAULT_OMNI_REPO}"
+readonly OMNI_REF="${SGLANG_OMNI_REF:-$DEFAULT_OMNI_REF}"
+readonly SGLANG_REPO="${SGLANG_REPO:-$DEFAULT_SGLANG_REPO}"
+readonly SGLANG_REF="${SGLANG_VERSION:-$DEFAULT_SGLANG_REF}"
+readonly CACHE_ROOT="${SGLANG_OMNI_CACHE:-${HOME}/.cache/sglang-omni}"
+readonly SGLANG_DIR="${SGLANG_SOURCE_DIR:-${CACHE_ROOT}/sglang-${SGLANG_REF//\//-}}"
+readonly OMNI_DIR="${SGLANG_OMNI_PROJECT_DIR:-${CACHE_ROOT}/sglang-omni-${OMNI_REF//\//-}}"
+readonly EXTRAS="${SGLANG_OMNI_EXTRAS:-}"
+NONINTERACTIVE="${NONINTERACTIVE:-0}"
+export UV_HTTP_TIMEOUT="${UV_HTTP_TIMEOUT:-300}"
+export UV_HTTP_RETRIES="${UV_HTTP_RETRIES:-5}"
+
+log() {
+  printf '[sglang-omni-install] %s\n' "$*"
+}
+
+die() {
+  printf '[sglang-omni-install][error] %s\n' "$*" >&2
+  exit 1
+}
+
+on_error() {
+  local exit_code=$?
+  printf '[sglang-omni-install][error] command failed at line %s: %s\n' "$1" "$2" >&2
+  exit "$exit_code"
+}
+
+trap 'on_error "$LINENO" "$BASH_COMMAND"' ERR
+
+SGLANG_STAGE_DIR=""
+CHECKOUT_TMP_DIR=""
+cleanup() {
+  if [[ -n "$SGLANG_STAGE_DIR" && -d "$SGLANG_STAGE_DIR" ]]; then
+    rm -rf -- "$SGLANG_STAGE_DIR"
+  fi
+  if [[ -n "$CHECKOUT_TMP_DIR" && -d "$CHECKOUT_TMP_DIR" ]]; then
+    rm -rf -- "$CHECKOUT_TMP_DIR"
+  fi
+}
+trap cleanup EXIT
+
+usage() {
+  cat <<'EOF'
+Usage: ./install.sh [OPTIONS]
+
+Install sglang-omni and the Apple Silicon Qwen3-ASR runtime into an isolated
+Python 3.12 virtual environment. Homebrew must already be installed.
+
+Options:
+  --non-interactive  Disable Homebrew auto-update (CI use).
+  -h, --help         Show this help.
+
+Environment:
+  SGLANG_OMNI_VENV         Virtual environment path.
+  SGLANG_OMNI_CACHE        Cache root for source checkouts.
+  SGLANG_SOURCE_DIR        SGLang source checkout path.
+  SGLANG_VERSION            SGLang git tag/branch (default: v0.5.16).
+  SGLANG_REPO               SGLang repository URL.
+  SGLANG_OMNI_REPO          sglang-omni repository URL for hosted use.
+  SGLANG_OMNI_REF           sglang-omni branch/tag (default: main).
+  SGLANG_OMNI_PROJECT_DIR   sglang-omni checkout path for hosted use.
+  SGLANG_OMNI_EXTRAS        Optional extras, comma-separated.
+  NONINTERACTIVE=1          Same as --non-interactive.
+  UV_HTTP_TIMEOUT            Per-request timeout in seconds (default: 300).
+  UV_HTTP_RETRIES            Network retry count (default: 5).
+
+The local checkout is installed when this script lives in an sglang-omni
+repository. If downloaded or piped from stdin, the configured repository/ref
+is cloned into SGLANG_OMNI_PROJECT_DIR (or the cache default above).
+EOF
+}
+
+while (($#)); do
+  case "$1" in
+    --non-interactive)
+      NONINTERACTIVE=1
+      ;;
+    -h|--help)
+      usage
+      exit 0
+      ;;
+    *)
+      die "unknown option: $1 (use --help)"
+      ;;
+  esac
+  shift
+done
+
+[[ "$(uname -s)" == "Darwin" ]] || die "unsupported operating system; this installer requires macOS"
+[[ "$(uname -m)" == "arm64" ]] || die "unsupported architecture; Apple Silicon arm64 is required"
+
+if [[ -n "${SGLANG_OMNI_PROJECT_DIR:-}" ]]; then
+  # An explicit destination selects hosted mode even when a local script copy
+  # happens to sit next to a checkout.
+  PROJECT_DIR=""
+elif [[ -n "$SCRIPT_DIR" && -f "$SCRIPT_DIR/pyproject.toml" ]] \
+  && grep -q '^name = "sglang-omni"' "$SCRIPT_DIR/pyproject.toml"; then
+  PROJECT_DIR="$SCRIPT_DIR"
+elif [[ -f "$PWD/pyproject.toml" ]] && grep -q '^name = "sglang-omni"' "$PWD/pyproject.toml"; then
+  # A downloaded installer can still be run from a checked-out repository.
+  PROJECT_DIR="$PWD"
+else
+  PROJECT_DIR=""
+fi
+
+if [[ -n "${SGLANG_OMNI_VENV:-}" ]]; then
+  readonly VENV_DIR="$SGLANG_OMNI_VENV"
+elif [[ -n "$PROJECT_DIR" ]]; then
+  readonly VENV_DIR="$PROJECT_DIR/.venv-apple"
+else
+  readonly VENV_DIR="$CACHE_ROOT/.venv-apple"
+fi
+
+find_brew() {
+  local candidate
+  # Prefer native Apple Silicon Homebrew even if an Intel/Rosetta brew appears
+  # earlier on PATH.
+  if [[ -x /opt/homebrew/bin/brew ]]; then
+    printf '%s\n' /opt/homebrew/bin/brew
+    return 0
+  fi
+  if command -v brew >/dev/null 2>&1; then
+    command -v brew
+    return 0
+  fi
+  for candidate in /opt/homebrew/bin/brew /usr/local/bin/brew; do
+    if [[ -x "$candidate" ]]; then
+      printf '%s\n' "$candidate"
+      return 0
+    fi
+  done
+  return 1
+}
+
+BREW_BIN="$(find_brew || true)"
+if [[ -z "$BREW_BIN" ]]; then
+  cat >&2 <<'EOF'
+[sglang-omni-install][error] Homebrew is required but was not found.
+Install Homebrew from https://brew.sh, ensure `brew` is on PATH, and rerun
+this script. This installer deliberately does not run Homebrew's bootstrapper
+or request administrator privileges.
+EOF
+  exit 1
+fi
+
+BREW_PREFIX="$($BREW_BIN --prefix)" || die "unable to determine the Homebrew prefix"
+[[ "$BREW_PREFIX" == "/opt/homebrew" ]] \
+  || die "native Apple Silicon Homebrew is required; found prefix: $BREW_PREFIX"
+export PATH="$BREW_PREFIX/bin:$BREW_PREFIX/sbin:$PATH"
+if [[ "$NONINTERACTIVE" == "1" ]]; then
+  export HOMEBREW_NO_AUTO_UPDATE="${HOMEBREW_NO_AUTO_UPDATE:-1}"
+fi
+
+ensure_formula() {
+  local formula="$1"
+  if "$BREW_BIN" list --formula "$formula" >/dev/null 2>&1; then
+    log "Homebrew formula already installed: $formula"
+    return 0
+  fi
+  log "Installing Homebrew formula: $formula"
+  if ! "$BREW_BIN" install "$formula"; then
+    die "Homebrew could not install $formula. Check `brew doctor` and Homebrew directory permissions; no administrator command was run by this script"
+  fi
+}
+
+ensure_formula ffmpeg@7
+ensure_formula uv
+
+if ! command -v git >/dev/null 2>&1 || ! git --version >/dev/null 2>&1; then
+  ensure_formula git
+fi
+command -v git >/dev/null 2>&1 || die "git is required; install it with Homebrew or Xcode Command Line Tools"
+command -v uv >/dev/null 2>&1 || die "uv is required; the Homebrew uv formula was not found on PATH"
+
+if [[ -d "$VENV_DIR" ]]; then
+  log "Reusing Python virtual environment: $VENV_DIR"
+  PYTHON_BIN="$VENV_DIR/bin/python"
+  [[ -x "$PYTHON_BIN" ]] || die "existing path is not a usable virtual environment: $VENV_DIR"
+else
+  log "Creating Python 3.12 virtual environment: $VENV_DIR"
+  uv venv --python 3.12 "$VENV_DIR"
+  PYTHON_BIN="$VENV_DIR/bin/python"
+fi
+[[ -x "$PYTHON_BIN" ]] || die "uv did not create $PYTHON_BIN"
+"$PYTHON_BIN" -c 'import sys; assert sys.version_info[:2] == (3, 12), sys.version' \
+  || die "the virtual environment must use Python 3.12"
+UV_PIP_INSTALL=(uv pip install --python "$PYTHON_BIN")
+
+clone_or_reuse() {
+  local destination="$1"
+  local ref="$2"
+  local repository="$3"
+  local label="$4"
+  local actual_repository current_commit target_commit
+
+  if [[ -e "$destination" && ! -d "$destination/.git" ]]; then
+    die "$label destination exists but is not a git checkout: $destination (choose another path)"
+  fi
+  if [[ -d "$destination/.git" ]]; then
+    actual_repository="$(git -C "$destination" remote get-url origin 2>/dev/null || true)"
+    [[ "$actual_repository" == "$repository" ]] \
+      || die "$label checkout origin is $actual_repository, expected $repository: $destination"
+    [[ -z "$(git -C "$destination" status --porcelain)" ]] \
+      || die "$label checkout has local changes; clean it or choose another path: $destination"
+    log "Refreshing $label $ref: $destination"
+    git -C "$destination" fetch --depth 1 origin "$ref"
+    current_commit="$(git -C "$destination" rev-parse HEAD)"
+    target_commit="$(git -C "$destination" rev-parse 'FETCH_HEAD^{commit}')"
+    if [[ "$current_commit" != "$target_commit" ]]; then
+      log "Updating $label checkout to $target_commit"
+      git -C "$destination" checkout --detach --quiet FETCH_HEAD
+    fi
+    return 0
+  fi
+  log "Cloning $label $ref: $destination"
+  mkdir -p "$(dirname -- "$destination")"
+  CHECKOUT_TMP_DIR="$(mktemp -d "${destination}.tmp.XXXXXX")"
+  git -C "$CHECKOUT_TMP_DIR" init --quiet
+  git -C "$CHECKOUT_TMP_DIR" remote add origin "$repository"
+  git -C "$CHECKOUT_TMP_DIR" fetch --depth 1 origin "$ref"
+  git -C "$CHECKOUT_TMP_DIR" checkout --detach --quiet FETCH_HEAD
+  [[ ! -e "$destination" ]] || die "$label destination appeared during installation: $destination"
+  mv "$CHECKOUT_TMP_DIR" "$destination"
+  CHECKOUT_TMP_DIR=""
+}
+
+clone_or_reuse "$SGLANG_DIR" "$SGLANG_REF" "$SGLANG_REPO" "SGLang"
+
+# SGLang keeps the CUDA-oriented project metadata in pyproject.toml and the
+# platform-neutral/MLX metadata in pyproject_other.toml. Its Python build also
+# references Rust and version files at the repository root, so stage the whole
+# checkout and replace only the staged metadata. The source checkout remains
+# untouched and the staged package is installed non-editably before cleanup.
+SGLANG_STAGE_DIR="$(mktemp -d "${TMPDIR:-/tmp}/sglang-omni-sglang.XXXXXX")"
+cp -R "$SGLANG_DIR/." "$SGLANG_STAGE_DIR/"
+if [[ -f "$SGLANG_DIR/python/pyproject_other.toml" ]]; then
+  cp "$SGLANG_DIR/python/pyproject_other.toml" "$SGLANG_STAGE_DIR/python/pyproject.toml"
+else
+  die "SGLang checkout does not contain python/pyproject_other.toml"
+fi
+
+log "Installing SGLang $SGLANG_REF with the all_mps extra"
+# The optional SGLang Rust extensions are not used by the MLX/Torch-MPS
+# Qwen3-ASR path. Skipping them avoids requiring a Rust toolchain on a clean Mac.
+SGLANG_BUILD_RUST_EXTS=none \
+  "${UV_PIP_INSTALL[@]}" --prerelease=allow "$SGLANG_STAGE_DIR/python[all_mps]"
+
+if [[ -z "$PROJECT_DIR" ]]; then
+  PROJECT_DIR="$OMNI_DIR"
+  clone_or_reuse "$PROJECT_DIR" "$OMNI_REF" "$OMNI_REPO" "sglang-omni"
+fi
+[[ -f "$PROJECT_DIR/pyproject.toml" ]] || die "sglang-omni pyproject.toml not found in $PROJECT_DIR"
+grep -q '^name = "sglang-omni"' "$PROJECT_DIR/pyproject.toml" \
+  || die "project is not sglang-omni: $PROJECT_DIR"
+
+PROJECT_SPEC="$PROJECT_DIR"
+if [[ -n "$EXTRAS" ]]; then
+  PROJECT_SPEC="${PROJECT_DIR}[${EXTRAS}]"
+fi
+log "Installing sglang-omni${EXTRAS:+ with extras: $EXTRAS}"
+"${UV_PIP_INSTALL[@]}" --prerelease=allow -e "$PROJECT_SPEC"
+
+FFMPEG_LIB="$($BREW_BIN --prefix ffmpeg@7)/lib"
+[[ -d "$FFMPEG_LIB" ]] || die "ffmpeg@7 library directory not found: $FFMPEG_LIB"
+
+log "Verifying the installed Python package and CLI"
+uv pip check --python "$PYTHON_BIN"
+"$PYTHON_BIN" -c 'import sglang_omni; print("sglang_omni", sglang_omni.__version__)'
+[[ -x "$VENV_DIR/bin/sgl-omni" ]] || die "sgl-omni console script was not installed"
+"$VENV_DIR/bin/sgl-omni" --help >/dev/null
+DYLD_LIBRARY_PATH="$FFMPEG_LIB${DYLD_LIBRARY_PATH:+:$DYLD_LIBRARY_PATH}" \
+SGLANG_USE_MLX=1 "$PYTHON_BIN" - <<'PY'
+import mlx.core as mx
+from torchcodec.decoders import AudioDecoder  # noqa: F401
+
+assert mx.metal.is_available(), "MLX Metal is unavailable"
+print("MLX Metal and TorchCodec FFmpeg loading are available")
+PY
+
+cat <<EOF
+
+Installation complete.
+Virtual environment: $VENV_DIR
+Activate it with:
+  source "$VENV_DIR/bin/activate"
+
+For TorchCodec/FFmpeg audio decoding, export:
+  export DYLD_LIBRARY_PATH="$FFMPEG_LIB\${DYLD_LIBRARY_PATH:+:\$DYLD_LIBRARY_PATH}"
+
+The Apple Silicon Qwen3-ASR examples are documented at:
+  $PROJECT_DIR/docs/cookbook/qwen3_asr.md
+EOF
--- a/install.sh
+++ b/install.sh
@@ -10,7 +10,7 @@ IFS=$'\n\t'
 readonly DEFAULT_OMNI_REPO="https://github.com/sgl-project/sglang-omni.git"
 readonly DEFAULT_OMNI_REF="main"
 readonly DEFAULT_SGLANG_REPO="https://github.com/sgl-project/sglang.git"
-readonly DEFAULT_SGLANG_REF="v0.5.16"
+readonly DEFAULT_SGLANG_REF="v0.5.18"
 
 SCRIPT_PATH="${BASH_SOURCE[0]:-}"
 if [[ -n "$SCRIPT_PATH" && -f "$SCRIPT_PATH" ]]; then
@@ -77,7 +77,7 @@ Environment:
   SGLANG_OMNI_VENV         Virtual environment path.
   SGLANG_OMNI_CACHE        Cache root for source checkouts.
   SGLANG_SOURCE_DIR        SGLang source checkout path.
-  SGLANG_VERSION            SGLang git tag/branch (default: v0.5.16).
+  SGLANG_VERSION            SGLang git tag/branch (default: v0.5.18).
   SGLANG_REPO               SGLang repository URL.
   SGLANG_OMNI_REPO          sglang-omni repository URL for hosted use.
   SGLANG_OMNI_REF           sglang-omni branch/tag (default: main).
@@ -182,7 +182,7 @@ ensure_formula() {
   fi
   log "Installing Homebrew formula: $formula"
   if ! "$BREW_BIN" install "$formula"; then
-    die "Homebrew could not install $formula. Check `brew doctor` and Homebrew directory permissions; no administrator command was run by this script"
+    die "Homebrew could not install $formula. Check 'brew doctor' and Homebrew directory permissions; no administrator command was run by this script"
   fi
 }
 
--- a/install.sh
+++ b/install.sh
@@ -59,6 +59,8 @@ cleanup() {
   fi
 }
 trap cleanup EXIT
+trap 'exit 130' INT
+trap 'exit 143' TERM
 
 usage() {
   cat <<'EOF'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,6 +34,8 @@ dependencies = [
     # note (yexiaodong): SGLang's PyPI wheel has unconditional CUDA dependencies,
     # so Apple Silicon installs this exact tag from source with the all_mps extra.
     "sglang==0.5.18; sys_platform != 'darwin' or platform_machine != 'arm64'",
+    "mlx; sys_platform == 'darwin' and platform_machine == 'arm64'",
+    "mlx-lm; sys_platform == 'darwin' and platform_machine == 'arm64'",
     "flashinfer_python[cu13]==0.6.17; sys_platform != 'darwin' or platform_machine != 'arm64'",  # sglang pin. Do NOT add flashinfer-cubin: PyPI has no matching cubin, and any leftover cubin wheel fails MiniMax DIT import.
     "cache-dit==1.3.0",      # MiniMax Music 3 DIT (hard import, not optional)
     "addict==2.4.0",         # sglang.multimodal_gen server_args; MiniMax DIT import
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,16 +29,18 @@ dependencies = [
     "datasets>=2.14.0",       # Arrow-based SeedTTS dataset loading
     "fastapi>=0.110.0",       # OpenAI-compatible API adapter
     "uvicorn>=0.23.0",        # ASGI server for FastAPI
-    "sglang==0.5.18",
-    "flashinfer_python[cu13]==0.6.17",  # sglang pin. Do NOT add flashinfer-cubin: PyPI has no matching cubin, and any leftover cubin wheel fails MiniMax DIT import.
+    # note (yexiaodong): SGLang's PyPI wheel has unconditional CUDA dependencies,
+    # so Apple Silicon installs this exact tag from source with the all_mps extra.
+    "sglang==0.5.18; sys_platform != 'darwin' or platform_machine != 'arm64'",
+    "flashinfer_python[cu13]==0.6.17; sys_platform != 'darwin' or platform_machine != 'arm64'",  # sglang pin. Do NOT add flashinfer-cubin: PyPI has no matching cubin, and any leftover cubin wheel fails MiniMax DIT import.
     "cache-dit==1.3.0",      # MiniMax Music 3 DIT (hard import, not optional)
     "addict==2.4.0",         # sglang.multimodal_gen server_args; MiniMax DIT import
     "imageio==2.36.0",       # sglang.multimodal_gen vision utils; MiniMax DIT import
     "imageio-ffmpeg==0.5.1", # Pinned with the sglang diffusion extra
-    "flash-attn-4>=4.0.0b18",  # Required by sglang, pairs with cutlass-dsl 4.6.2
-    "kernels>=0.14.1,<0.15",  # Match the pinned sglang stack
-    "nixl-cu13>=1.1.0",  # CUDA 13 relay wheel; the generic nixl/-cu12 wheel breaks on cu130
-    "mooncake-transfer-engine-cuda13>=0.3.10",  # CUDA 13 relay wheel; generic mooncake-transfer-engine pulls cu12 and breaks import mooncake on cu130
+    "flash-attn-4>=4.0.0b18; sys_platform != 'darwin' or platform_machine != 'arm64'",  # Required by sglang, pairs with cutlass-dsl 4.6.2
+    "kernels>=0.14.1,<0.15; sys_platform != 'darwin' or platform_machine != 'arm64'",  # Match the pinned sglang stack
+    "nixl-cu13>=1.1.0; sys_platform != 'darwin' or platform_machine != 'arm64'",  # CUDA 13 relay wheel; the generic nixl/-cu12 wheel breaks on cu130
+    "mooncake-transfer-engine-cuda13>=0.3.10; sys_platform != 'darwin' or platform_machine != 'arm64'",  # CUDA 13 relay wheel; generic mooncake-transfer-engine pulls cu12 and breaks import mooncake on cu130
     "logger",
     "httpx",
     "xxhash>=3.0.0",
@@ -81,13 +83,15 @@ dependencies = [
     "torchcodec==0.15.0",     # Matches torch 2.13.x and the pinned sglang stack
     # Gradio playground
     "gradio>=4.0.0",
-    # dots.tts model operators; serving/runtime stays owned by sglang-omni
-    "dots.tts==0.2.1",
+    # note (yexiaodong): dots.tts pulls Pynini, whose OpenFst build is not
+    # available on Apple arm64 and is unrelated to the Qwen3-ASR path.
+    "dots.tts==0.2.1; sys_platform != 'darwin' or platform_machine != 'arm64'",
     # Ming-Omni
     "diffusers==0.37.0",      # Omni's own Ming-Omni pin; not an sglang core dependency
     "x-transformers",          # Ming talker rotary embeddings
-    # ZONOS2 written->spoken text normalization; pulls pynini + sacremoses.
-    "nemo_text_processing==1.2.0",
+    # note (yexiaodong): ZONOS2 text normalization pulls Pynini/OpenFst, which
+    # has no usable Apple arm64 wheel and is unrelated to the Qwen3-ASR path.
+    "nemo_text_processing==1.2.0; sys_platform != 'darwin' or platform_machine != 'arm64'",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,8 +19,10 @@ dependencies = [
     "pydantic>=2.0.0",        # Data validation and serialization
     "PyYAML>=6.0",            # YAML config parsing
     "pybase64>=1.4.0",        # Fast base64 decoding for inline media payloads
-    "torch==2.13.0",          # PyTorch for tensor operations
-    "torchvision==0.28.0",    # torchvision for vision processor, pinned version based on torch 2.13.0
+    "torch==2.13.0; sys_platform != 'darwin' or platform_machine != 'arm64'",
+    "torch==2.11.0; sys_platform == 'darwin' and platform_machine == 'arm64'",
+    "torchvision==0.28.0; sys_platform != 'darwin' or platform_machine != 'arm64'",
+    "torchvision==0.26.0; sys_platform == 'darwin' and platform_machine == 'arm64'",
     "accelerate>=0.27.0",     # init_empty_weights for meta initialization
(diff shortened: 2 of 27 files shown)


SuKi2cn:
I ran this PR end-to-end on Apple Silicon hardware (install → unit-test matrices → real-inference smoke tests on both paths) and did a code read-through. Since the CI jobs for this PR are currently skipped, an independent reproduction on the target hardware seemed useful — results and review notes below. Overall both paths work as described.

## Test environment

- MacBook Air, Apple M5, 16 GB unified memory, macOS 26.3.1 (arm64), Homebrew at `/opt/homebrew`
- Installed via `./install.sh` from a clean checkout of this branch: Python 3.12.14, **mlx 0.32.2 / mlx-lm 0.31.3**, torch 2.11.0, transformers 5.12.1, torchcodec 0.11.1, ffmpeg 7.1.5, SGLang v0.5.16 built from source by the installer
- `install.sh` completed without intervention and was idempotent on re-run

## Unit test matrices

| Matrix | Command scope | Result here |
| --- | --- | --- |
| `SGLANG_USE_MLX=1` | `test_mlx_model.py` + `test_mlx_scheduler_runner.py` + `test_request_builders.py` + `tests/unit_test/platforms` | **506 passed, 1 failed** |
| `SGLANG_USE_MLX=0` | `tests/unit_test/qwen3_asr` + `tests/unit_test/platforms` + `test_register_omni_model.py` | **602 passed, 1 failed, 3 skipped** (skips are the expected requires-CUDA ones) |

### The one failure — reproducible, likely an mlx-version-dependent tolerance issue

`test_mlx_model.py::test_native_mlx_prefill_only_projects_last_position` fails deterministically on mlx 0.32.2: **20/20 random seeds fail** with the default `mx.allclose` tolerance (rtol=1e-5). Worst absolute diff ~3.2e-3; even rtol=atol=1e-3 still fails 1/20 seeds.

Root cause: `_forward_last_logits` (slice hidden states, then project) vs. the full forward (project all positions, then slice) dispatch to different Metal kernels (GEMV vs GEMM) for the final projection, so float32 results legitimately differ beyond default tolerance. The implementation looks correct — this is a test-tolerance defect, amplified by `_tiny_model()` using unseeded random weights. Suggest seeding the model and either relaxing the tolerance (e.g. `rtol=1e-2, atol=1e-2`) or asserting on `argmax` equality instead of elementwise closeness.

## Real-inference smoke tests — both paths pass

Same `tests/data/query_to_cars.wav`, exact transcript `"How many cars are there in the picture?"` on every request.

| Path | Checkpoint | Warm total (non-stream) | Stream TTFB | SSE |
| --- | --- | ---: | ---: | --- |
| MLX (`SGLANG_USE_MLX=1`) | mlx-community/Qwen3-ASR-0.6B-4bit | ~0.12–0.15 s | ~0.14 s | `delta` → `done` → `[DONE]` |
| Torch MPS | Qwen/Qwen3-ASR-0.6B (bf16) | ~0.26 s | ~0.09 s | 4 deltas → `done` → `[DONE]` |

Consistent with (slightly faster than) the medians in the PR description; the delta is plausibly just the M5 vs. whatever this was benchmarked on.

## Review notes

### Should fix

1. **`install.sh:185`** — backticks inside the double-quoted `die` message are command substitution, so a `brew install` failure actually *executes* `brew doctor` (slow) and splices its output into the error. Quote as `` 'brew doctor' `` or escape the backticks.
2. **`mlx/model.py` `sanitize()`** — `lm_head.weight` is dropped unconditionally, but `__init__` creates a real `lm_head` when `tie_word_embeddings=False`. An untied checkpoint would end up with an uninitialized head. Currently harmless for the tied 0.6B/1.7B checkpoints, but the guard should mirror the config: only drop when tied.
3. **`platforms/apple.py` `get_device_total_memory()`** — imports `mlx.core` unconditionally, unlike `get_current_memory_usage` right below it which guards on `use_mlx()`. On a Torch-MPS-only environment without MLX installed this raises `ImportError`.

### Questions

4. **Chained decode abort path** (`mlx_model_worker.py`): a mismatched successor batch raises `RuntimeError("MLX chained decode requires an unchanged request batch")`, which aborts the request. Since the scheduler launches batch N before resolving N-1, can a mid-chain abort/retraction produce a decode→decode transition with a changed batch and hit this? If it's provably unreachable, a comment explaining why would help; if reachable, falling back to a fresh launch seems safer. Relatedly, the pending step lives on `self`, which the base runner's docstring (`model_runner/base.py`) explicitly says not to do — worth a note on why it's safe here.
5. **Platform detection order** (`platforms/__init__.py`): the Apple check sits after `platform.is_cpu()`. It works on my machine (Apple platform was selected), but if a future SGLang resolves macOS to its CPU platform, Apple silently loses. Also unlike the MUSA/NPU/XPU checks it doesn't guard on `type(platform) is SRTPlatform`. Is the ordering deliberate?
6. **No MLX↔platform cross-check**: `SGLANG_USE_MLX=1` on a CUDA/Linux host would happily build the MLX worker (`scheduling/bootstrap.py`). A cheap "MLX requested but platform is not Apple" error would save confusing failures.
7. **`max_running_requests` asymmetry**: Torch MPS enforces `max_running_requests=1` in `validate_before_infrastructure`, but the MLX path only documents it in the cookbook without enforcing. Intentional?
8. **Torch MPS sampling**: `custom_*_forward` returns argmax token ids directly with `logits_output=None`, so repetition penalty, logprobs, etc. silently become no-ops on this path (beyond the documented greedy-only constraint). Worth stating in the cookbook, or rejecting those params like temperature is rejected.

### Nits

- `platforms/__init__.py`: module-level `import platform` is shadowed by the `platform` parameter of `_as_omni_platform`/`get_platform_spec` — works via globals, but confusing; consider renaming.
- `mlx/model.py:203`: `-1e9` mask overflows to `-inf` in fp16 (harmless here since the block mask always leaves a valid diagonal, but `finfo(dtype).min`-style would be more robust). Also the mask is dense O(L²) with a Python loop of slice assignments — fine for typical audio lengths, but noticeably heavier than the ragged varlen path on the CUDA side.
- `mlx/runner.py`: the docstring on `make_qwen3_asr_mlx_runner_class` says the class is built lazily "so non-Apple imports need no MLX", but the module itself does `import mlx.core as mx` at top level. `_to_numpy` will also `TypeError` on bfloat16 tensors (needs `.float()` first) if features ever arrive as bf16.
- `torch_mps_runner.py:75`: `config.rope_parameters["rope_theta"]` is accessed unconditionally, while `_text_config` ten lines up defensively handles both `rope_parameters`/`rope_scaling` spellings. Fine on the pinned transformers 5.12.1; will break on older ones — suggest matching the defensive style.
- `request_builders.py`: the `greedy_only` temperature check runs after the full audio load/feature-extraction, so a bad request burns an encode before erroring — cheap to hoist to the top. It also only validates `temperature` (fine today since `top_p` is forced to 1.0, but the flag name promises more).
- `mlx/model.py`: a few config fields are silently ignored (`conv_chunksize` stored but unused — the torch path uses it for conv OOM chunking, `embed_scale`, `attention_bias`, `rope_scaling`, `activation_function` hard-coded to gelu), and `_get_feat_extract_output_lengths` is now the third copy of the conv length formula (also in `audio_lengths.py` and `encoder_cuda_graph.py`). Not blocking for the supported checkpoints, but worth a comment or assertion so silent divergence can't creep in.
- `pyproject.toml`: on Darwin/arm64 the project now has no declared `sglang` dependency, so a bare `pip install -e .` without `install.sh` yields an env that's broken in a non-obvious way. Maybe worth a runtime import check with a pointer to the installer.

### Positives

- `test_mlx_model.py` testing real math on a tiny model (rather than mocks) is great — that's exactly how the tolerance issue above became visible.
- The `# note (yexiaodong):` comments explaining non-obvious constraints are genuinely helpful, and the installer is careful (no sudo, `rm -rf` confined to its own mktemp dirs, refuses to clobber dirty checkouts).
- Test-coverage suggestion: `test_mlx_scheduler_runner.py` exercises the chaining state machine with stubs but never a real `_execution_bridge`, so the FutureMap/`publish_next_tokens` handoff — the riskiest new logic — has no coverage, and `torch_mps_runner.py` has none at all. Even a small fake-bridge test would help lock in the contract.

Happy to re-run anything on this hardware if that helps.

tests/unit_test/qwen3_asr/test_mlx_model.py
```diff
@@ -0,0 +1,181 @@
...
+)
+from sglang_omni.models.qwen3_asr.mlx.model import Qwen3ASRModel  # noqa: E402
+from sglang_omni.models.qwen3_asr.mlx.runner import (  # noqa: E402
+    Qwen3ASRMlxModelRunner,
+    make_qwen3_asr_mlx_runner_class,
+)
+
+
+def _tiny_model() -> Qwen3ASRModel:
+    audio = AudioEncoderConfig(
```
SuKi2cn, line 26: Seeding here makes the numeric assertions below deterministic across runs (the failure magnitude currently drifts with the random weights):

```suggestion
    # Seeded so numeric assertions see the same weights on every run.
    mx.random.seed(0)
    audio = AudioEncoderConfig(
```

tests/unit_test/qwen3_asr/test_mlx_model.py
```diff
@@ -0,0 +1,181 @@
...
+    model = _tiny_model()
+    input_ids = mx.array([[1, 2, 3]], dtype=mx.int32)
+    embeddings = model.model.embed_tokens(input_ids)
+
+    full_logits = model(input_ids, input_embeddings=embeddings)
+    last_logits = model._forward_last_logits(embeddings)
+
+    mx.eval(full_logits, last_logits)
+    assert last_logits.shape == (1, 1, 64)
+    assert mx.allclose(last_logits, full_logits[:, -1:, :]).item()
```
SuKi2cn: This is the assertion that fails deterministically on mlx 0.32.2 (20/20 seeds, worst abs diff 3.2e-3): the single-position projection (GEMV) and the full-sequence projection (GEMM) can dispatch to different Metal kernels, so float32 results agree only to ~1e-3 depending on the MLX version. An explicit tolerance keeps the equivalence check meaningful without depending on kernel selection:

```suggestion
    # The single-position projection and the full-sequence projection may hit
    # different Metal matmul kernels (GEMV vs GEMM), so float32 results agree
    # only to ~1e-3 depending on the MLX version.
    assert mx.allclose(last_logits, full_logits[:, -1:, :], rtol=1e-2, atol=1e-2).item()
```

install.sh
```diff
@@ -0,0 +1,315 @@
...
+
+ensure_formula() {
+  local formula="$1"
+  if "$BREW_BIN" list --formula "$formula" >/dev/null 2>&1; then
+    log "Homebrew formula already installed: $formula"
+    return 0
+  fi
+  log "Installing Homebrew formula: $formula"
+  if ! "$BREW_BIN" install "$formula"; then
+    die "Homebrew could not install $formula. Check `brew doctor` and Homebrew directory permissions; no administrator command was run by this script"
```
SuKi2cn: The backticks inside this double-quoted string are command substitution, so the error path actually executes `brew doctor` and splices its output into the message:

```suggestion
    die "Homebrew could not install $formula. Check 'brew doctor' and Homebrew directory permissions; no administrator command was run by this script"
```

sglang_omni/platforms/apple.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def get_device_name(self, device_id: int = 0) -> str:
+        self._validate_device_id(device_id)
+        return "Apple Metal"
+
+    def get_device_total_memory(self, device_id: int = 0) -> int:
+        self._validate_device_id(device_id)
+        import mlx.core as mx
+
+        return int(mx.device_info()["max_recommended_working_set_size"])
```
SuKi2cn: This is the only method in this class importing `mlx.core` without a `use_mlx()` gate, so a Torch-MPS-only environment without MLX raises `ImportError`. `torch.mps.recommended_max_memory()` returns the same Metal recommended-working-set value (verified identical on my machine, 12713115648 bytes):

```suggestion
    def get_device_total_memory(self, device_id: int = 0) -> int:
        self._validate_device_id(device_id)
        from sglang.srt.utils.tensor_bridge import use_mlx

        if use_mlx():
            import mlx.core as mx

            return int(mx.device_info()["max_recommended_working_set_size"])
        return int(torch.mps.recommended_max_memory())
```

tests/unit_test/platforms/test_apple.py
```diff
@@ -0,0 +1,50 @@
...
+
+    assert apple.get_device(0) == torch.device("mps")
+    apple.set_device(torch.device("mps"))
+    with pytest.raises(ValueError, match="one Metal device"):
+        apple.get_device(1)
+    with pytest.raises(ValueError, match="Expected an MPS device"):
+        apple.set_device(torch.device("cpu"))
+
+
+def test_apple_stage_rejects_tensor_parallelism() -> None:
```
SuKi2cn, line 48: Optional regression test for the non-MLX branch of `get_device_total_memory` (skipped off-MPS so CUDA CI is unaffected; `use_mlx` is monkeypatched at the module because it is `lru_cache`d):

```suggestion
@pytest.mark.skipif(
    not torch.backends.mps.is_available(), reason="requires an Apple MPS device"
)
def test_apple_device_total_memory_does_not_require_mlx(monkeypatch) -> None:
    import sglang.srt.utils.tensor_bridge as tensor_bridge

    apple = AppleOmniPlatform()
    monkeypatch.setattr(tensor_bridge, "use_mlx", lambda: False)

    assert apple.get_device_total_memory(0) > 0


def test_apple_stage_rejects_tensor_parallelism() -> None:
```

SuKi2cn (commented):
Inline suggestions for the three "should fix" items from my test report above, so they can be applied directly from the review. With all of them applied, my local matrices go to 508 passed / 604 passed + 3 skipped (one new test each), and the previously failing tolerance test passes 6/6 consecutive runs on mlx 0.32.2.

yeahdongcn:
> Inline suggestions for the three "should fix" items from my test report above, so they can be applied directly from the review. With all of them applied, my local matrices go to 508 passed / 604 passed + 3 skipped (one new test each), and the previously failing tolerance test passes 6/6 consecutive runs on mlx 0.32.2.

Nice! Some of these were introduced by the pinned SGLang version bump. I'll address them accordingly.

yeahdongcn:
@SuKi2cn I should have fixed everything, and `Codex` ran all the tests. Could you please rerun them on your M5? Thanks!

adityavaid:
I reproduced this PR end-to-end on a different Apple Silicon generation than the one in the description (M1 16 GB → M4 Pro 48 GB). The relative ordering of the three paths matches the PR exactly; absolute latencies are ~3.5× lower on M4 Pro.

### Hardware / environment

| Item | Value |
| --- | --- |
| Machine | Apple M4 Pro, 48 GB unified memory |
| OS | macOS 26.6.2 (arm64) |
| SGLang | `v0.5.18` from source, `all_mps` extra, `SGLANG_BUILD_RUST_EXTS=none` |
| sglang-omni | 0.1.3 (editable) |
| MLX | mlx 0.32.2, mlx-metal 0.32.2, mlx-lm 0.31.3 |
| Torch | torch 2.11.0, torchaudio 2.11.0, torchcodec 0.11.1 |
| Transformers | 5.12.1 |
| FFmpeg | 7.1.5 (`ffmpeg@7`, keg-only, via `DYLD_LIBRARY_PATH`) |


### Results — M4 Pro 48 GB

| Path | Checkpoint | TTFT median | TTFT mean | Total median | Total mean |
| Omni MLX | `mlx-community/Qwen3-ASR-0.6B-4bit` | 0.054 s | 0.061 s | 0.076 s | 0.081 s |
| Omni MLX | `Qwen/Qwen3-ASR-0.6B` (BF16) | 0.080 s | 0.079 s | 0.137 s | 0.134 s |
| Omni Torch MPS | `Qwen/Qwen3-ASR-0.6B` (BF16) | 0.101 s | 0.101 s | 0.208 s | 0.209 s |


### Cross-machine comparison (total median, auto language)

| Path | PR — M1 16 GB, SGLang 0.5.18 | This run — M4 Pro 48 GB | Speedup |
| --- | ---: | ---: | ---: |
| Omni MLX 4-bit | 0.267 s | 0.076 s | 3.5× |
| Omni MLX BF16 | 0.485 s | 0.137 s | 3.5× |
| Omni Torch MPS BF16 | 1.013 s | 0.208 s | 4.9× |

he749119-source:
**Independent test report #2 — output parity vs the official implementation, edge-input behavior, and memory stability (Apple M4 Pro)**

Great report from @SuKi2cn above — I ran an independent reproduction on different Apple hardware and have cut everything that would duplicate it (install, serve-level smoke tests, streaming). What's left are the dimensions not yet covered: **transcript parity against the official reference implementation, edge-input behavior, and long-run memory stability.** Two findings turned out to affect the CUDA path as well, not just this PR.

Tested at commit `b83b3b9` (before the three follow-up fixes landed).

## Environment

- Apple M4 Pro, 48 GB unified memory, macOS 15.7.3 (arm64)
- mlx **0.32.2** / mlx-lm 0.31.3, torch 2.13 (CPU), transformers 5.12.1
- Checkpoint: **Qwen/Qwen3-ASR-1.7B** (bf16) — complementing the 0.6B runs above
- Harness note: direct calls into this PR's `mlx/model.py` + the `runner.py` load path (not the full server); caveats flagged below where this matters.

## Unit tests + a datapoint for the tolerance issue

`tests/unit_test/platforms/test_apple.py` + `tests/unit_test/qwen3_asr/`: **602 passed / 3 skipped** (the expected requires-CUDA skips).

Datapoint for `test_native_mlx_prefill_only_projects_last_position`: on the **same mlx 0.32.2** but an **M4 Pro**, it is *flaky* rather than deterministic — ~1 failure per 6 runs (vs. 20/20 seed failures reported on M5). So the divergence appears hardware-generation-dependent on top of mlx-version-dependent, which fits the GEMV/GEMM dispatch explanation and further supports fixing the test (seeded weights + argmax assertion or relaxed tolerance) rather than the implementation.

Minor: running this matrix requires `mlx-lm`, which isn't declared as a dependency anywhere — worth a mention in docs or an extra.

## Transcript parity vs the official implementation

Judge: the official `qwen-asr` package (transformers 5.16.1 backend, CPU). Same five inputs through both stacks, greedy decoding:

| Input | Official (torch) | This PR (MLX, direct-call) | Match |
|---|---|---|---|
| English query, 4.6 s | `How many cars are there in the picture?` | same | ✅ exact |
| Chinese (synthetic), 4.2 s | `今天天气真不错，我们一起去公园散步吧。` | same | ✅ exact, incl. punctuation |
| Very short "Hi", 0.4 s | `Hi.` | `Hi.` | ✅ |
| Cough (no speech) | `咳咳咳。` | `咳咳咳。` | ✅ |
| Silence, 0.5 s | *(empty)* | `language Chinese<asr_text>嗯。` | ⚠️ divergence |

4/4 exact on real speech (incl. Chinese + language auto-detection) — the MLX port is faithful where it matters.

## The silence divergence, root-caused

Elimination chain on the silence case:

- **Padding** (`padding=True` vs `max_length`): MLX output unchanged either way → not padding.
- **Dtype** (official torch bf16 vs fp32): both produce the empty result → not dtype.
- **Prompt template**: the official chat template renders a leading *empty system block* (`<|im_start|>system\n<|im_end|>\n`); omni's `request_builders.py` template does not. Adding that block to the MLX harness flips the silence output to exactly `language None<asr_text>` — matching the official stack.

Caveat: after aligning the template, my direct-call harness also changed behavior on valid speech (residual harness variables: tokenizer regex warning under transformers 5.16.1, audio-pad expansion details), so the final word needs a run through the real serve path. The template mismatch itself is real and behavior-affecting, though.

**Two upstream findings — these affect the CUDA path too, so I'm happy to file them as separate issues:**

1. **Prompt template parity**: omni's qwen3-asr prompt lacks the empty system block the official stack always emits. Harmless on clear speech, behavior-changing on degenerate inputs.
2. **`language None<asr_text>` sentinel is not translated**: the official package maps it to an empty result (`qwen_asr/inference/utils.py`), while omni's post-processing (`request_builders.py:407`) treats `"None"` as a language name and passes the text through — so silence/noise inputs return hallucinated text (e.g. `嗯。`) to users instead of empty.

## A note on `greedy_only`

The greedy-only constraint is a reasonable MVP cut, but one consequence may be worth planning for: `temperature` is a standard parameter of the OpenAI `/v1/audio/transcriptions` API, so stock OpenAI SDK clients that pass it will get a hard `ValueError` on the Apple backend while the same request succeeds on CUDA. Related to @SuKi2cn's point 8 (silently no-op'd sampling params on the MPS path). Two-step suggestion: short-term, document the constraint in the API-visible error path consistently on both paths; longer-term, the MLX side can lift it cheaply — `mlx_lm` ships ready-made temperature/top-p samplers, so swapping the `mx.argmax` in the decode loop for a sampler call covers most of it (and becomes a prerequisite anyway if a TTS path lands on Apple later, where sampling is essential).

## Memory stability (serving-relevant)

15 consecutive transcriptions on the MLX path: `mx.get_active_memory()` flat at 3893 MB, peak flat at 4816 MB — no growth.

## Perf datapoint (1.7B bf16, direct-call, M4 Pro)

Model load 2.57 s; audio encode 0.01 s (4.6 s clip → 60 audio tokens); greedy decode ~10.5 tok/s.

---

Happy to re-run any of this after the follow-up fixes, and to contribute a **macOS CI workflow** (GitHub's free `macos-15` arm64 runners) so the Apple path stays green — the exact dependency set for the test matrix is already mapped from these runs.

JackZeng0208:
Testing based on the M4 Max 64GB with newest version of code. Running test on SGLang Omni's full benchmark on SeedTTS EN and ZH splits

## SeedTTS EN:

| Path | req/s | RTFx | lat mean | lat p50 | lat p95 | RTF mean | TTFT mean | TTFT p95 | ITL mean | corpus WER |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| MLX 4-bit, non-stream | 15.06 | 71.34 | 0.066 s | 0.066 s | 0.080 s | 0.0143 | — | — | — | 0.0200 |
| MLX 4-bit, stream | 15.09 | 71.47 | 0.066 s | 0.066 s | 0.080 s | 0.0143 | 0.0299 s | 0.0342 s | 0.0355 s | 0.0200 |
| Torch MPS BF16, non-stream | 4.27 | 20.23 | 0.234 s | 0.231 s | 0.295 s | 0.0506 | — | — | — | 0.0153 |
| Torch MPS BF16, stream | 4.22 | 20.00 | 0.237 s | 0.234 s | 0.298 s | 0.0512 | 0.0554 s | 0.0655 s | 0.0535 s | 0.0153 |

## SeedTTS ZH:

Note: the gate scores ZH on WER only (speed is gated on EN), so ZH was run non-streaming.

| Path | req/s | RTFx | lat mean | lat p95 | RTF mean | corpus WER |
| MLX 4-bit | 16.20 | 75.86 | 0.062 s | 0.074 s | 0.0133 | 0.0122 |
| Torch MPS BF16 | 4.70 | 21.99 | 0.213 s | 0.278 s | 0.0459 | 0.0100 |

## Quantization trade-off

| Path           | corpus WER | lat mean |
| -------------- | ---------: | -------: |
| MLX 4-bit      |     0.0200 |  0.074 s |
| Torch MPS BF16 |     0.0153 |  0.242 s |

MLX 4-bit is 3.3x faster but carries +31 % relative WER.

## MLX 4-bit TTFT vs. Eager (Torch MPS BF16)

| Path           |     TTFT | total latency | TTFT share |
| -------------- | -------: | ------------: | ---------: |
| MLX 4-bit      | 0.0305 s |      0.0696 s |       44 % |
| Torch MPS BF16 | 0.0579 s |      0.2440 s |       24 % |

## Accuracy Test

| Split                    | MLX 4-bit | Torch MPS BF16 |
| ------------------------ | --------: | -------------: |
| SeedTTS EN (1,088 clips) |    0.0200 |         0.0153 |
| SeedTTS ZH (2,020 clips) |    0.0122 |         0.0100 |

he749119-source:
”Chained decode abort path (mlx_model_worker.py): a mismatched successor batch raises RuntimeError("MLX chained decode requires an unchanged request batch"), which aborts the request. Since the scheduler launches batch N before resolving N-1, can a mid-chain abort/retraction produce a decode→decode transition with a changed batch and hit this? If it's provably unreachable, a comment explaining why would help; if reachable, falling back to a fresh launch seems safer. Relatedly, the pending step lives on self, which the base runner's docstring (model_runner/base.py) explicitly says not to do — worth a note on why it's safe here.“  -- Adding a second half to this question from reading the same code path: whether or not a mid-chain batch change is reachable, if it ever *is* hit, the state machine can't recover. Compare the two failure paths in `MlxSchedulerModelRunner`:

```python
# execute_resolve — cleans up before propagating:
except Exception:
    self._last_mlx_pending = None
    raise

# execute_launch — raises with the stale pending still set:
if previous.mode != "decode" or previous_ids != current_ids:
    raise RuntimeError("MLX chained decode requires an unchanged request batch")
```

If the scheduler catches that `RuntimeError` and keeps going, every subsequent `execute_launch` compares against the same stale `_last_mlx_pending` and raises again — a permanent launch-fail loop — and the lazy MLX graph held by the stale pending is never finalized or dropped. Whatever the answer to reachability turns out to be, clearing `_last_mlx_pending` before (or in a `finally` around) that raise seems like cheap insurance, mirroring what `execute_resolve` already does.

SuKi2cn:
Reran everything on the M5 at the new head (`a7c2499`, SGLang 0.5.18 built by `./install.sh` — the bumped `v0.5.18` default worked without intervention). Everything passes on this machine.

Same environment as my first report otherwise: MacBook Air, Apple M5, 16 GB, macOS 26.3.1, mlx 0.32.2 / mlx-lm 0.31.3.

## Unit matrices

| Matrix | Previous head (0.5.16) | New head (0.5.18) |
| --- | --- | --- |
| `SGLANG_USE_MLX=1` (mlx model + scheduler runner + request builders + platforms) | 507 (1 failed) | **512 passed, 0 failed** |
| `SGLANG_USE_MLX=0` (qwen3_asr + platforms + register_omni_model) | 606 (1 failed, 3 skipped) | **609 passed, 3 skipped** (the expected requires-CUDA skips) |

`test_native_mlx_prefill_only_projects_last_position` — previously 20/20 seed failures on this machine — now passes **6/6 consecutive runs** with the seeded model + explicit tolerance.

## Smoke tests

Same protocol as before (`query_to_cars.wav`, exact transcript on every request):

| Path | Warm total (non-stream) | Stream TTFB | SSE |
| --- | ---: | ---: | --- |
| MLX 4-bit | ~0.12–0.15 s | ~0.07 s | `delta` → `done` → `[DONE]` |
| Torch MPS bf16 | ~0.27 s | ~0.10 s | 3 deltas → `done` → `[DONE]` |

Latencies are in line with my 0.5.16 numbers (MLX streaming slightly faster).

All five suggestions from my review are resolved in this head, and I can confirm the extras landed too — untied-`lm_head` handling in `sanitize` and the bf16 `_to_numpy` conversion both have tests now. From my side this looks ready.

pyproject.toml
```diff
@@ -29,16 +31,18 @@ dependencies = [
     "datasets>=2.14.0",       # Arrow-based SeedTTS dataset loading
     "fastapi>=0.110.0",       # OpenAI-compatible API adapter
     "uvicorn>=0.23.0",        # ASGI server for FastAPI
-    "sglang==0.5.18",
-    "flashinfer_python[cu13]==0.6.17",  # sglang pin. Do NOT add flashinfer-cubin: PyPI has no matching cubin, and any leftover cubin wheel fails MiniMax DIT import.
+    # note (yexiaodong): SGLang's PyPI wheel has unconditional CUDA dependencies,
+    # so Apple Silicon installs this exact tag from source with the all_mps extra.
+    "sglang==0.5.18; sys_platform != 'darwin' or platform_machine != 'arm64'",
```
SuKi2cn, line 36: Following up on @he749119-source's note about the undeclared test dependency: `sglang_omni/models/qwen3_asr/mlx/` imports `mlx.core` and `mlx_lm` directly, but `pyproject.toml` has no mlx entries — today both arrive only transitively through the sglang `all_mps` source build, so a bare `pip install -e .` on Apple Silicon produces an environment that fails at import time on the MLX path. Declaring them with the existing Darwin markers closes that gap; sglang's `all_mps` extra leaves both unpinned, so this cannot conflict with the source build:

```suggestion
    "sglang==0.5.18; sys_platform != 'darwin' or platform_machine != 'arm64'",
    "mlx; sys_platform == 'darwin' and platform_machine == 'arm64'",  # Imported directly by qwen3_asr/mlx and platforms/apple
    "mlx-lm; sys_platform == 'darwin' and platform_machine == 'arm64'",  # Imported directly by the qwen3_asr MLX runner/model loader
```

tigerwash:
**Independent test report — behaviour above concurrency 1 (Apple M5 MacBook Air, 24 GB)**

Everyone above ran at concurrency 1, so I only tested `max_running_requests > 1`. Not a blocker, but the two Apple paths behave differently.

**Environment:** Apple M5 (4P+6E), 24 GB, macOS 26.5 arm64 · mlx / mlx-metal 0.32.2, mlx-lm 0.31.3 · torch 2.11.0, torchcodec 0.11.1 · transformers 5.12.1 · SGLang v0.5.18 from source, sglang-omni 0.1.3 · ffmpeg 7.1.5 · tested at `a7c2499`.

**Regression check:** **512 passed** / **609 passed, 3 skipped** — same as @SuKi2cn's re-run, on a 24 GB M5 rather than 16 GB.

## MLX accepts `max_running_requests > 1`, and it is a net loss

`validate_before_infrastructure` (`engine_builder.py:264`) rejects `!= 1` for Torch MPS; the MLX branch of `generation_defaults` passes it through unchecked, with a pipeline default of `64` (`qwen3_asr/config.py:78`). The cookbook shows `1` for both. With `--asr.engine.max_running_requests 4`, Torch MPS raises `ValueError` and exits at startup, while MLX starts and batches for real (`Decode batch, #running-req: 4`).

`query_to_cars.wav`, `Qwen3-ASR-0.6B-4bit`, non-streaming, 3 warmups + 32 requests per level:

| Concurrency | req/s | total p50 | total p95 | exact |
| 1 | 9.19 | 0.107 s | 0.117 s | 32/32 |
| 2 | 9.40 | 0.186 s | 0.383 s | 32/32 |
| 4 | 11.65 | 0.310 s | 0.493 s | 32/32 |
| 8 | 13.78 | 0.572 s | 0.590 s | 32/32 |

Correctness is unaffected (128/128 exact), but concurrency 2 costs **+74% latency for +2% throughput**, and 8 buys 1.50× throughput for 5.35× latency. Server-side `gen throughput` is ~117 tok/s at `#running-req: 1` vs. ~142 at `4` — MLX decode gains little from batching.

Worth either mirroring the Torch MPS guard on MLX, or noting in the cookbook that `> 1` is accepted but trades latency for almost nothing.

## Abort probe

A server-level companion to `061eedf4`'s stub test: 12 streaming requests at concurrency 4, 6 hard-closed mid-stream → 6 matching `Coordinator aborted req=` lines, 5/5 exact afterwards at baseline latency, and the chained-decode `RuntimeError` never fired. The reference `061eedf4` keeps does get drained end-to-end, so @he749119-source's launch-fail loop doesn't look reachable from the real scheduler either.

Caveat: one clip, one machine, one run per point — behavioural smoke tests, not benchmark-grade numbers.

JackZeng0208:
Found an issue with the Torch MPS streaming clip limit by bisecting clip lengths: an 87.8 s clip completes in 2.7 s, while a 92.4 s clip hangs forever. There is no error or timeout, the connection remains open, and `/health` continues to return 200.

It looks like the streaming guard is fed the model-level `max_native_clip_s` (1200 s), which nothing narrows down to what the Torch MPS profile's 2048-token budget can actually hold. So clips in between are admitted and then never scheduled.

SuKi2cn:
**Non-streaming long audio on the Apple paths (M5) — chunking works on both, plus a deterministic Torch MPS encoder crash at specific clip lengths**

Complementing @JackZeng0208's streaming-hang finding with the non-streaming side: the platform-neutral chunking layer (`sglang_omni/serve/transcription_chunking.py`, 60 s chunks) turns out to already work on Apple — nobody had exercised it here yet — and testing it surfaced a separate, deterministic crash in the Torch MPS encoder path.

Environment: same as my earlier reports (M5 16 GB, mlx 0.32.2), at head `061eedf4`. Both unit matrices also re-verified at this head: 512 passed / 609 passed + 3 skipped.

## Chunked non-streaming requests (`stream=false`)

Test clips: the repo's `query_to_cars.wav` (4.62 s) looped to 92.4 s (20×) and 184.8 s (40×).

| Path | 92.4 s (2 chunks) | 184.8 s (4 chunks) |
| --- | --- | --- |
| MLX 4-bit | HTTP 200 in **1.7 s** | HTTP 200 in **2.3 s**; `verbose_json` returns 4 segments with correct boundaries (59.85 / 119.4 / 179.25 / 184.8) |
| Torch MPS bf16 | HTTP 200 in **7.3 s** | **HTTP 500** — the in-server tail chunk (179.2–184.8 s) crashes the audio encoder |

Note the 92.4 s row: the same length that hangs forever with `stream=true` on MPS completes fine non-streaming — so for MPS today, chunked non-streaming is the working long-audio route, modulo the crash below.

(Caveat on transcript content: the model compresses runs of identical repeated sentences — a 60 s chunk holding ~13 repeats transcribes ~9. Verified model-level with a standalone chunk; segment timestamps are gapless, so this is not chunk-layer data loss. An artifact of the synthetic looped clip, not a PR defect.)

## The Torch MPS crash, minimized

Error (same every time, also reproducible with a standalone short clip, no chunking involved):

```
cannot reshape tensor of 0 elements into shape [0, 14, -1] ...
  sglang/srt/models/qwen3_omni_moe.py:142 (encoder layer self_attn)
  sglang/srt/layers/attention/vision.py:1355
```

It is deterministic in the mel-frame count. Exact-length 16 kHz sine probes (`n_samples = frames × 160`):

| frames | 496 | 498 | 499 | 500 | 501 | 502 | 504 | 548 | 552 | 554 | 555 | 556 | 558 | 560 | 562 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| result | ok | ok | ok | ok | **500** | ok | **500** | ok | **500** | ok | **500** | ok | **500** | ok | **500** |

The failing counts sit in arithmetic runs of 3, which smells like a window-boundary off-by-one: the windowed-attention `cu_chunk_lens` construction in upstream `qwen3_omni_moe.py` (~line 336) pairs an estimated after-CNN length with the mask-selected real one, and at these counts they appear to disagree, yielding an empty attention window (hypothesis — not verified in code).

Two scoping observations:

- The MLX path transcribes all of the crashing clips fine (501/504/555/558 all HTTP 200), so it is specific to the Torch path.
- sglang-omni's CUDA deployment uses its own encoder implementation; the Torch MPS profile is what exercises this upstream `qwen3_omni_moe` audio tower via SDPA. So in practice this crash is Apple-MPS-specific here, though the underlying window math may be latent upstream.

Repro:

```bash
python3 -c "
import math, struct, wave
w = wave.open('sine_555f.wav', 'w'); w.setnchannels(1); w.setsampwidth(2); w.setframerate(16000)
w.writeframes(b''.join(struct.pack('<h', int(12000*math.sin(2*math.pi*440*i/16000))) for i in range(555*160)))
w.close()"
curl -s -X POST http://localhost:8000/v1/audio/transcriptions \
  -F model=Qwen/Qwen3-ASR-0.6B -F file=@sine_555f.wav -F response_format=json
# Torch MPS server → HTTP 500; SGLANG_USE_MLX=1 server → HTTP 200
```

Happy to file this as a separate issue (and the streaming-limit gap likewise) if that's preferred over tracking it in this thread.

yeahdongcn:
Guys, I think we may be relying too heavily on AI here. Please first determine whether the issue is generic across hardware backends or specific to Apple Silicon. I'd prefer that we not use AI to automatically scan and address the comments.

SuKi2cn:
**Correction to my long-audio report above: the MPS crash is instance-state-dependent, not length-determined — and the repro in that report doesn't work as written.**

On the generic-vs-Apple question: Apple-specific, in the torch-MPS runtime. Also retracting my note that CUDA uses a separate encoder — it runs the same upstream tower, unaffected because the failure is in the MPS runtime, not the shared math: direct tower calls pass in every config tested (3 seeds × CPU fp32/MPS fp32/MPS bf16 × 6 shapes incl. batched, 54/54, same M5 env).

Where I went wrong: all length probes ran inside one server instance; pass/fail at a given length is a property of that instance's request history — the same order reproduces identically across fresh instances, a changed order flips the same clip. Disregard the frame-count table, the sine repro, and the "no chunking involved" note: the lengths I re-tested return 200 on a fresh server, and whether a chunk-free history can crash is untested.

At the crash (instrumented, `061eedf4`), correct per-chunk lengths go in but the mask-select comes out empty (`hidden=(0, 896)`), with the emptying stage varying by instance — a torch-MPS kernel issue, not a logic bug in this PR or sglang.

Working repro, 5/5 fresh instances: send one short clip (the repo's 4.6 s wav), then the 184.8 s upload → HTTP 500. The 3-request variant crashed 6/6.

Still standing: all MLX runs passed; MPS handled 92.4 s chunked in every order tested — but I'd no longer call chunked non-streaming dependable on MPS until this is fixed. Happy to file this upstream against torch MPS if useful.

yeahdongcn:
The `HTTP 500` issue should be fixed as well.

jzheng17:
**The short version, three things.**

**A rebase fixes a bad bug.** Streaming any clip longer than 60 s currently returns a short sentence that has nothing to do with the audio, with HTTP 200 and no error in the logs. The cause is an empty system turn this branch is missing, and [#1781](https://github.com/sgl-project/sglang-omni/pull/1781) already added it on `main` on 08-28. This branch points at a commit from 08-26.

**The endpoint takes a vocabulary hint and throws it away.** `params["prompt"]` is stored and never read, so a client that sends its user's vocabulary gets a 200 and no effect. I opened [#1807](https://github.com/sgl-project/sglang-omni/pull/1807) to fix that on `main`. This matters more for your paths than for the CUDA ones, because knowing the user's own words is most of what a local voice client has going for it.

**Start-up and idle cost are what separate this from mlx-audio, not speed.** Latency and accuracy come out level across nine datasets. But the server takes 12.71 s to start against mlx-audio's 951 ms, and holds 4.11% of a core while completely idle. I opened [#1808](https://github.com/sgl-project/sglang-omni/pull/1808), which takes start-up to 9.00 s, and [#1809](https://github.com/sgl-project/sglang-omni/pull/1809) as a draft for the idle CPU that I could not verify here.

Everything below is the detail behind those, plus the numbers.

---

I tested this on an Apple M4 with 16 GiB of memory, running macOS 26.1. I ran both paths: MLX with `mlx-community/Qwen3-ASR-0.6B-4bit`, and Torch MPS with the official BF16 checkpoint. Everything ran at concurrency 1, on clips I built for the voice-input case: short sentences, Chinese-English code switching, numbers, disfluency, keyboard noise, and clips that cross the 60 s chunking threshold.

Both of your test matrices reproduce. I get 507 passed under `SGLANG_USE_MLX=1`, and 602 passed with 3 skipped when it is unset, against the 603 and 3 you report.

One note on method, because two of the differences below are small. Before comparing anything, I ran the same configuration twice as an A/A control. On this machine the band is 3.1% on TTFT and 3.5% end to end. Each request also records the host's pageout counter, so I drop a request that paged rather than averaging it in. Nothing below falls inside that band.

I have filed patches for three of these. All three are against `main` and none of them depend on this PR: I opened #1807 and #1808, and #1809 as a draft.

## A rebase fixes a bad failure

When I stream a clip longer than 60 s, the server returns a short string that has nothing to do with the audio. It returns HTTP 200 and logs no error:

```
61.67 s -> 0 delta events, done.text = "Stylistic English speaking English"
68.53 s -> 0 delta events, done.text = "Enlistments."
65.00 s -> 28 deltas, correct transcript
```

I built ten clips from 60 distinct sentences and ran each twice. Both runs of a clip give byte-identical results, so this is deterministic. Four of the eight clips above 60 s fail. The first 60 seconds of transcript are lost as well, not only the tail.

The cause is the empty system turn. `main` has it:

```python
# Qwen3-ASR's checkpoint chat template emits this empty system turn even
# when no caller-provided system context is present.
_SYSTEM_PROMPT = "<|im_start|>system\n<|im_end|>\n"
```

This branch does not. #1781 added that line on 2026-08-28, and this branch points at 1452d08 from 08-26. When I add the line back, all four clips recover:

| clip | without the system turn | with it |
|---:|---:|---:|
| 61.67 s | 4 of 200 words, 100% WER | 200 of 200, 0.50% |
| 68.53 s | 1 of 222 words, 100% WER | 222 of 222, 0.45% |
| 71.91 s | 2 of 232 words, 100% WER | 232 of 232, 0.43% |
| 89.36 s | 2 of 282 words, 100% WER | 282 of 282, 0.71% |

Those numbers match what mlx-audio produces on the same clips, 0.43% to 0.83%. That comparison is what rules out the model: mlx-audio does not chunk either, and it transcribes all ten correctly.

Non-streaming never shows this problem because it splits the audio at 60 s, so it never hands the model a single clip that long. Streaming does hand it one, up to `stream_clip_limit_s`, which resolves to the model's native 1200 s.

So a rebase resolves this. I do want to flag how easy it is to miss, though. One absent empty turn is invisible on short audio and total on long audio, so a change to this template needs a check on a long clip, not only the unit tests.

## The default path is 2.5x to 4.6x slower

This PR ships two paths, and reaching the faster one takes two separate steps.

| path | how you get it | checkpoint |
|---|---|---|
| MLX | set `SGLANG_USE_MLX=1` **and** switch to an `mlx-community` checkpoint | 4-bit |
| Torch MPS | do nothing | official BF16 |

If a reader follows the cookbook and misses either step, they get Torch MPS:

| audio | MLX total | Torch MPS total | ratio |
|---:|---:|---:|---:|
| 2.87 s | 145 ms | 366 ms | 2.5x |
| 11.86 s | 419 ms | 1365 ms | 3.3x |
| 29.65 s | 965 ms | 4457 ms | 4.6x |

```
MLX total =   56.0 ms + 30.4 ms per second of audio
MPS total = -240.5 ms + 151.2 ms per second of audio
```

The slopes differ by a factor of five, so the difference grows with utterance length. Half a minute of dictation takes four and a half seconds on the default path. Accuracy is not what pays for the difference: Torch MPS on BF16 also scores 0.00% WER at every duration in this sweep.

I am not asking you to change the default. The two steps sit in different parts of the cookbook, though, and someone will do one of them and not the other.

## Cold start, and one line in this PR

The server takes 12.71 s from launch to `/health` returning 200. mlx-audio loads the same weights in 951 ms. The time splits roughly into 4.8 s of Python imports, 5.6 s asking the Hub whether the cached checkpoint moved, and 2.3 s loading the model and setting up the device.

I opened #1808 to remove part of that on `main`. It resolves the repo id to a local directory once instead of letting each `from_pretrained` call resolve it again, which takes cold start from 12.71 s to 9.00 s and Hub requests from 24 to 11.

The 11 requests that remain come from this PR's own code:

```python
# sglang_omni/models/qwen3_asr/mlx/runner.py:16-23
def _resolve_model_path(model_path: str) -> Path:
    path = Path(model_path).expanduser()
    if path.is_dir():
        return path.resolve()
    from huggingface_hub import snapshot_download
    return Path(snapshot_download(model_path))
```

`snapshot_download` without `local_files_only` revalidates against the Hub even when the cache is complete. Would it make sense to try `local_files_only=True` first here and fall through on a miss? That would take the count to zero. I left this one out of my #1808 because the file is new in this PR and belongs with your changes rather than mine.

This matters beyond the number. 12 s forces the client to keep the process resident. Under a second lets it load the model when the user presses the key and release it afterwards, and a resident process is also what pays the idle CPU below.

## The endpoint accepts the prompt field and discards it

`speech_to_text.py` stores `params["prompt"]`, and `qwen3_asr/request_builders.py` never reads it. `whisper_asr/request_builders.py:230`, `moss_transcribe_diarize/request_builders.py:204` and `ming_tts/request_builders.py:327` all consume the same parameter. The endpoint returns 200 and the biasing does nothing.

I opened #1807 to fix this against `main`, which now has the system turn to put the text in. Before my change, no clip responds to a supplied vocabulary. After it, 5 of 10 respond, and mlx-audio moves the same 5 clips in the same direction:

| term | without bias | with bias |
|---|---|---|
| LLM | `L L M` | `LLM` |
| nginx | `Nginx` | `nginx` |
| PyTorch | `Python` | `Py Torch` |

Two limits came out of measuring it. An irrelevant vocabulary makes accuracy worse: I applied the same technical word list to unrelated conversation and CER went from 16.40% to 18.94%. And the bias text is prefilled on every request, so it costs latency every time. At 50 terms that is +71% on technical sentences and +85% on short ones, while accuracy is best at around 20 terms.

This matters for the case this PR opens up. A local voice client's advantage is that it knows the user's own vocabulary, and the field it would use for that currently does nothing.

## Idle CPU

Over 180 s with no requests in flight, the API process holds 0.17% of a core and the ASR worker holds 4.11%. Nothing leaks; RSS drifts down 15 MiB over the window.

`OmniScheduler._sleep_during_idle` sleeps 1 ms, and the next pass drains the inbox with `get_nowait`. So the loop wakes about a thousand times a second, and each pass runs `_process_admin_requests`, `recv_requests` and a lock acquisition against empty queues. Upstream's `IdleSleeper` does not apply here, because it polls ZMQ sockets while this scheduler receives on a `queue.Queue`.

I wrote a patch that blocks on the inbox instead and opened it as #1809. I marked it a draft because I could not run it: `main`'s scheduler imports `sglang.srt.runtime_context.get_model`, which the v0.5.16 pin this PR requires does not have, so my patched file will not start in the environment I have. The numbers and the reasoning are in that PR, and it needs someone with a working `main` setup before it is worth a real review.

None of this matters on a busy server. On the laptop this PR targets, 4% of a core runs all day for no work.

## Smaller notes

The MLX matrix leaves out `tests/unit_test/qwen3_asr/test_pipeline.py`, which the non-MLX matrix includes. If I add it under `SGLANG_USE_MLX=1` I get 20 failures out of 39, and they are the right failures. Those tests assert `cuda_graph_bs_prefill`, per-GPU queries and attention-backend defaults, and the MLX path documents itself as not producing any of them. Nothing gates them, so a reviewer who runs the obvious command sees red.

nit: `max_audio_clip_s` is hard-coded at 60. mlx-audio exposes the same number as a request parameter.

`DYLD_LIBRARY_PATH` does not survive exec through a binary that SIP protects. So wrapping `sgl-omni` in `/usr/bin/time`, or starting it from launchd or a process manager, strips the variable without saying so. WAV keeps working, because `soundfile` carries its own libsndfile and never needs FFmpeg, so the loss only appears on compressed audio. It appears as a 500 whose message blames libtorchcodec and points the user at reinstalling FFmpeg. macOS and iOS record m4a by default, so a voice client hits this. Would a line in the cookbook cover it, or a startup self-check that decodes one compressed sample?

## What holds up

The serving stack costs almost nothing. mlx-audio runs in-process with no HTTP, no IPC and no scheduler, and omni still matches it across the duration sweep: +0.3% to -8.0%, with one point outside the noise band. I think that is the result worth stating, because omni carries the whole serving path and still ties.

Accuracy matches mlx-audio across nine corpora, every one within one percentage point. That includes the awkward ones: code switching, keyboard noise, numbers, disfluency, real LibriSpeech and FLEURS. Whatever those corpora expose belongs to the model, and this PR loses no accuracy to serving.

The endpoint handles the inputs a push-to-talk UI produces. Three seconds of digital silence returns an empty string rather than invented text, which is the case that fires every time a user taps the key without speaking. Malformed input returns 400 with a clear message. I saw no 5xx and no hang. mp3 and m4a decode correctly once `DYLD_LIBRARY_PATH` reaches the process.

Stability holds. I ran 300 requests with no failures, and per-clip latency drifted -1% to -5% from the first quarter of the run to the last, which is inside the noise band. Nothing leaked. Concurrent requests queue cleanly at n times the single-request wall time with no drops, and six concurrent SSE streams showed no crosstalk and no starvation.

Short utterances are 10 of 10 exact at 81 ms, and they are most of what an input method sees. The model also inserts punctuation correctly without the speaker saying it, 6 of 6.

I have the corpora and the per-clip numbers behind all of this if any of it would be useful.

yeahdongcn:
The MLX model runner should now align with the default PyTorch model runner when resolving model paths. I expect we'll eventually drop the MLX model runner once the SGLang side refactor is complete.

sglang_omni/models/qwen3_asr/mlx/model.py
```diff
@@ -0,0 +1,619 @@
+# SPDX-License-Identifier: MIT
+# Derived from mlx-audio Qwen3-ASR (Copyright 2025 Prince Canuma and contributors).
```
Ratish1, line 2: should we include the full mlx-audio MIT license notice for this file and mlx/config.py in the wheel?

Ratish1 (approved):
LGTM

jzheng17:
Re-ran my boundary set on the rebased branch: the above-60 s streaming hallucination is gone — all ten clips clean, 0.43–0.83% WER, matching mlx-audio on the same files. Accuracy across my scenario sets is unchanged. Thanks for the quick turnaround on the rebase, the model-path alignment and the 500 fix.

Issue #1582, opened 2026-08-17
[Feature] Add nemotron-3.5-asr-streaming-0.6b to SGLang-Omni.


LG-0927:
### Checklist

- [x] This is a feature request rather than a question.
- [x] This issue is written in English.

## Background and Motivation

Add support for [nvidia/nemotron-3.5-asr-streaming-0.6b](https://huggingface.co/nvidia/nemotron-3.5-asr-streaming-0.6b), a 600M-parameter multilingual ASR model based on a 24-layer Cache-Aware FastConformer encoder and an RNN-T decoder.

The model supports 40 language-locales, including 32 locales with out-of-the-box transcription coverage. It provides configurable streaming latency at 80, 160, 320, 560, and 1,120 ms while reusing encoder attention and convolution caches between non-overlapping audio chunks.

| Area | Existing SGLang-Omni ASR paths | Nemotron 3.5 ASR |
|---|---|---|
| Decoder | Mostly Transformer autoregressive decoding | LSTM predictor + RNN-T joint network |
| Streaming state | Primarily request/output streaming | Encoder KV cache, convolution cache, and RNN-T decoder state |
| Language control | Model-specific language hints | Locale prompt or `auto` detection |
| Upstream runtime | Transformers 5.12.1 / SGLang 0.5.16 | Requires Transformers >= 5.13.0 |

Nemotron cannot be treated as another causal language model under the existing RadixAttention decode path. It needs a model-owned RNN-T scheduler and request-scoped streaming state.

### Phase 1: Offline Transcription MVP

- Register `Nemotron3_5AsrForRNNT` with a single `asr` pipeline stage.
- Serve complete audio files through `/v1/audio/transcriptions`.
- Reuse shared mono 16 kHz audio preparation and support batched offline inference.
- Support explicit locales, bare language codes, and `language=auto`.
- Return clean transcripts for `json`/`text` and the detected locale in `verbose_json` when available.

### Phase 2: Native Cache-Aware Streaming

- Add a dedicated scheduler based on `StreamingSimpleScheduler`.
- Preserve encoder attention, convolution, and RNN-T decoder caches per request.
- Process fixed-size chunks for lookahead values `{0, 1, 3, 6, 13}`.
- Batch ready chunks across compatible streams without sharing cache state.
- Add or define an incremental-audio transport; the current `stream=true` response mode must not be described as live audio ingestion.

### Non-goals

[...]

PR #1731, opened 2026-08-25
feat: add offline Nemotron 3.5 ASR support

LG-0927:
## Motivation

Add the offline/batched foundation for `nvidia/nemotron-3.5-asr-streaming-0.6b` through the existing `/v1/audio/transcriptions` API. Native cache-aware live audio ingestion is intentionally split into follow-up PRs.

## Modifications

- Register a single-stage `Nemotron3_5AsrForRNNT` ASR pipeline and transcription adapter.
- Reuse the shared mono/16 kHz audio preparation path and run one true batched `generate()` call per scheduler batch.
- Support the checkpoint's declared lookahead values (`0`, `3`, `6`, and `13`) and locale prompts, including `auto`.
- Add a narrow Transformers 5.13 compatibility layer, with upstream provenance and Apache-2.0 modification notices, while reusing the installed Transformers Parakeet implementation where possible.
- Preserve existing Whisper/MOSS text behavior and expose detected locale in `verbose_json` when available.
- Isolate per-request scheduler failures so one invalid item does not fail the other requests in the same batch.

## Related Issues

Part of #1582.

## Accuracy Test

Validated on an RTX 4090 against a separate Transformers 5.13 Oracle using checkpoint revision `1c8deaecc64b91f034d73e08dd8b64625eb3395d`, FP32, and identical audio/decoding settings.

- Single-item inference: exact complete token sequence, raw transcript, and clean transcript match.
- Two-item inference: exact complete token sequences, raw transcripts, and clean transcripts match for `en-US` and `auto`.
- Runtime instrumentation reported batch sizes `[2, 2]`, proving the SGLang-Omni model path executed one true two-item batch.

## Benchmark & Profiling

This PR establishes the offline correctness baseline only. Streaming TTFT, speech-end-to-final latency, per-chunk latency, RTF/RTFx, throughput, and cache-path counters will be reported with the native-streaming follow-up, where those measurements are meaningful. Model-card numbers will be labeled as reported references rather than local measurements.

## Tests

- `174 passed`: all Nemotron unit tests plus speech-to-text, transcription chunking, and shared scheduler regressions.
- Real multipart bytes upload to `/v1/audio/transcriptions`: HTTP 200 with a clean `verbose_json` transcript and `language=en-US`.
- `git diff --cached --check` passed before commit.

## Checklist

- [x] Add unit tests.

## CI

model, and


LG-0927:
@PoTaTo-Mika hi, I’ve added the missing Nemotron documentation and signed the explanatory comments as suggested.

PR #1733, opened 2026-08-25
perf(qwen3-tts): batch ad-hoc reference encoding

ywgrit:
## Summary

- opportunistically coalesce already-queued Qwen3-TTS ad-hoc reference cache misses through the existing `ReferenceEncodeService`
- keep the existing dedicated-stream speech-tokenizer batching from #1336, while grouping identical-shape waveforms into one mel/ECAPA speaker-encoder forward
- preserve per-request ordering, cache/single-flight behavior, ICL/x-vector semantics, and the single-item compatibility path

Unequal waveforms are deliberately not padded together: the upstream ECAPA implementation does not propagate a padding-length mask through its squeeze-excitation path, so padding can change the speaker embedding.

This is a narrower follow-up to the earlier Qwen3-TTS reference-preprocessing work in #1336. It does not replace the ref-code batcher or its dedicated CUDA stream; it batches the remaining speaker-embedding stage when concurrent references have an exact shape/dtype match.

## Production-safety details

- `max_batch_wait_ms=0`: only requests already present in the queue coalesce, so a lone uncached request does not pay a fixed collection delay.
- Cache hits and same-key followers resolve before this queue and are never re-encoded.
- If a grouped speaker forward fails, the hook retries only the speaker stage per item and reuses the already-running ref-code futures. This avoids duplicate tokenizer GPU submissions and keeps failures request-local.

## Supporting operator result

Measured on an NVIDIA GeForce RTX 5090 with PyTorch 2.11.0+cu130, BF16, and `Qwen/Qwen3-TTS-12Hz-0.6B-Base`. Eight mixed 3 s / 5 s references, 10 warmups, and 50 alternating A/B iterations:

| Path | ECAPA calls | CUDA Event p50 |
| --- | ---: | ---: |
| Sequential | 8 | 299.76 ms |
| Exact-shape grouped | 2 | 94.91 ms |

This is a **3.158x speedup for the isolated mel + ECAPA stage only**. It is not an end-to-end serving claim. Minimum cosine similarity was `0.9999967`; maximum relative L2 error was `0.2597%` in BF16.

## Validation

- rebased onto current `main`
- `150 passed, 5 skipped` across the Qwen3-TTS pipeline and shared reference-encoder batching suites; the skipped tests require CUDA
- focused coverage for real service coalescing, item mapping, exact-shape grouping, zero-wait configuration, speaker-only fallback, duplicate-work prevention, CUDA stream handoff, and shutdown behavior
- Black 24.10, autoflake 2.3.1, isort, `py_compile`, and `git diff --check`

## Draft status / remaining performance gate

This PR stays Draft until the current zero-wait production service path is measured on H20 at concurrency 1/2/4/8 with:

- repeated equal durations and arbitrary unique durations
- actual speaker group-size distribution from a representative reference workload
- TTFA/prefill latency, throughput, quality metrics, and A/A noise

The goal is to verify a deployment-level win without regressing concurrency-one or arbitrary-length requests before requesting maintainer review.

### DIFF
--- a/sglang_omni/models/qwen3_tts/request_builders.py
+++ b/sglang_omni/models/qwen3_tts/request_builders.py
@@ -456,24 +456,27 @@ def build_generation_kwargs(
         explicit_fields = set()
 
     selected_fields = set()
-    for field in _GENERATION_FIELDS:
-        value = params.get(field)
+    for field_name in _GENERATION_FIELDS:
+        value = params.get(field_name)
         if value is None:
             continue
-        if field in _IMPLICIT_SAMPLING_DEFAULTS and field not in explicit_fields:
-            if value in _IMPLICIT_SAMPLING_DEFAULTS[field]:
+        if (
+            field_name in _IMPLICIT_SAMPLING_DEFAULTS
+            and field_name not in explicit_fields
+        ):
+            if value in _IMPLICIT_SAMPLING_DEFAULTS[field_name]:
                 continue
-        selected_fields.add(field)
+        selected_fields.add(field_name)
 
     max_new_tokens = params.get("max_new_tokens")
     if max_new_tokens is None:
         max_new_tokens = QWEN3_TTS_DEFAULT_MAX_NEW_TOKENS
     generation_kwargs: dict[str, Any] = {"max_new_tokens": int(max_new_tokens)}
-    for field in _GENERATION_FIELDS:
-        if field == "max_new_tokens":
+    for field_name in _GENERATION_FIELDS:
+        if field_name == "max_new_tokens":
             continue
-        if field in selected_fields and params.get(field) is not None:
-            generation_kwargs[field] = params[field]
+        if field_name in selected_fields and params.get(field_name) is not None:
+            generation_kwargs[field_name] = params[field_name]
     return generation_kwargs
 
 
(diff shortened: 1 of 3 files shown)


ywgrit:
Hi @shuaills and @luojiaxuan, I completed the production-serving A/B for this narrow follow-up to #1336 and want to report the negative result before asking for review.

The isolated mixed-shape encoder path reaches 2.09×, but under long-sample serving traffic only mixed c8 improves (+5.89%); equal c4/c8 and mixed c4 regress by 9.01%, 10.81%, and 3.40%. Across 505 reference items, only 336 batches formed, for an average batch size of 1.503, so grouping/queue overhead prevents the local kernel gain from becoming a reliable serving gain. I removed the experimental wait-window expansion and kept the smallest semantic implementation; the rebased focused suite is 168 passed, 5 skipped.

I am treating this as an evidence-backed NO-GO and will not claim it as a performance improvement. Would you prefer that I close the Draft, or keep it open only as a minimal reference for a future scheduler-level batching design?

JiaxinD:
The controlled serving A/B supports the NO-GO: the isolated 2.09x encoder gain does not survive the observed batch-size distribution, and three of four serving cells regress. Closing the Draft so this implementation is not mistaken for a landing candidate. The measurements and the future scheduler-level batching direction remain tracked in #1418.

PR #1735, opened 2026-08-25
[Qwen3-TTS] Make initial vocoder CUDA Graph capture startup-safe

Ez4Allen:
## Motivation

- Qwen3-TTS initial vocoder CUDA Graph capture could overlap concurrent ref-code CUDA event synchronization in the same CUDA context.
- On H100 this produced `cudaErrorStreamCaptureUnsupported`, graph invalidation, and HTTP request failure.

## Fix

- Prepare startup-known initial vocoder B1/B2/B4/B8 CUDA Graphs before the vocoder scheduler becomes live.
- Reuse the existing startup/factory lifecycle.
- Keep `on_serving_start` idempotent; no recapture.
- No preprocessing serialization, graph disabling, `capture_error_mode` relaxation, or steady-state synchronization is added.

## Validation

- Controlled CUDA reproducer: failure before fix, pass after fix.
- RTX 4080: 10/10 fresh starts, 80/80 cold-reference requests.
- H100 80GB: 10/10 fresh starts, 80/80 HTTP 200, zero capture errors, zero graph invalidations.
- B1/B2/B4/B8 replay bitwise equal to eager, `max_abs_diff=0.0`.
- Explicit H100 Arm A full-pipeline run: 6/6 HTTP 200, real ref-code GPU encode, Talker B6 -> graph bucket B8, Predictor execution, vocoder graph replay, peak GPU util 66%.
- Full local Qwen3-TTS tests passed.

## Related

- #1336
- #1418
- #1523
- Open #1675 may need to include its follow-up graph holder in the same early startup preparation after rebasing.

### DIFF
--- a/sglang_omni/models/qwen3_tts/stages.py
+++ b/sglang_omni/models/qwen3_tts/stages.py
@@ -192,7 +192,7 @@ def create_vocoder_executor(
         attn_implementation=attn_implementation,
     )
 
-    return Qwen3TTSStreamingVocoderScheduler(
+    scheduler = Qwen3TTSStreamingVocoderScheduler(
         tokenizer,
         device=device,
         stream_stride=stream_stride,
@@ -209,3 +209,5 @@ def create_vocoder_executor(
         initial_cuda_graph=initial_cuda_graph,
         enable_deterministic_inference=enable_deterministic_inference,
     )
+    scheduler.prepare_startup_graphs()
+    return scheduler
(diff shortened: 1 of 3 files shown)


luojiaxuan:
This is now covered by #1786. It moves both graph holders into `warmup_now()` at factory-build time, so capture finishes before the stage publishes readiness.

The scope changed after #1675 landed and added `_followup_decode_graphs`. Startup-safe capture then had to cover both holders, and this branch also conflicts on `on_serving_start` now.

Same race and same diagnosis as here, so feel free to close this unless you see something #1786 missed.

Ez4Allen:
Thanks for the heads-up. I checked #1786 and it covers the same race, including the follow-up graph holder added by #1675. I don't see anything missing here, so I'll close this as superseded by #1786.

PR #1739, opened 2026-08-26
[Fix] Key the MiniMax Music 3 AR/DIT handoff by request id

AkazaAkane:
## Motivation

Stacked on #1738 (part 2 of 4 on top of #1619). **Merge #1738 first.**

`ar_can_admit()` gates the AR scheduler's admission queue, so AR stays
parked until something calls `end_dit_handoff()`. In #1619 that is a single
process-wide boolean with no request scoping, which leaves two holes:

- **Any** `end_dit_handoff()` wakes AR, including one belonging to a
  different request than the one that parked it. Today `max_concurrency`
  is forced to 1 so this rarely bites, but the invariant is only
  incidentally true.
- Nothing records *which* request AR is waiting on. A handoff that never
  ends stops admission permanently, and the server goes silent with no
  log line explaining why. `clear_stream_state` is the only wake path, so
  any DIT-side failure that bypasses it (for example an exception raised
  inside `on_streaming_new_request`, which `_handle_streaming_new_request`
  does not guard) wedges the process until restart.

## Modifications

Track outstanding DIT request ids and wake AR only when the set drains.
Retiring a request that never handed off becomes a no-op instead of an
early wake; repeated calls stay idempotent.

A handoff that never ends is **reported, not force-corrected**. Waking AR
while DIT/DAV still holds the GPU is exactly how this configuration runs
out of memory, so a stuck server is preferable to a crashed one:
`ar_can_admit()` logs the outstanding request ids once after
`STALL_REPORT_SECONDS` (15 min) and leaves AR parked. The check is done
on the existing admission poll, so it costs no timer thread.

`begin_dit_handoff` / `end_dit_handoff` now take a `request_id`.

## Accuracy Test

No effect on generated audio; this only changes admission gating.

## Checklist

- [x] Add unit tests.

### DIFF
--- a/sglang_omni/models/minimax_music3/acoustic.py
+++ b/sglang_omni/models/minimax_music3/acoustic.py
@@ -509,7 +509,7 @@ class MiniMaxMusic3AcousticScheduler(StreamingSimpleScheduler):
             state.last_condition = None
         if self._decoder.serial_offload:
             self._decoder.offload_to_cpu()
-            get_coordinator().end_dit_handoff()
+            get_coordinator().end_dit_handoff(request_id)
 
 
 __all__ = [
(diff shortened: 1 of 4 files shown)


AkazaAkane:

PR #1740, opened 2026-08-26
[Fix] Keep MiniMax Music 3 offloaded weights on the host, not round-tripped

AkazaAkane:
## Motivation

Stacked on #1739 (part 3 of 4 on top of #1619). **Merge #1738 and #1739 first.**

This PR fixes the two things that stop #1619 from delivering what its flag
promises. DIT/DAV stays FP32 throughout — no dtype change.

**1. A startup peak that makes the mode unusable on the cards it targets.**

`_build_dit` and `_build_dav` call
`load_torch_state(..., device=self.device)`, staging both checkpoints
*through the GPU* before `offload_to_cpu()` moves them back. AR is already
registered and resident by then, so startup always peaks with AR **and**
DIT/DAV fully resident — precisely the peak serial offload exists to
avoid. Whichever stage is built second hits the combined footprint.

**2. Round-tripping immutable weights on every request.**

`move_module_to_device` ran `module.to()` in both directions per handoff:

- The weights never change during inference, yet every sleep paid a full
  device-to-host copy of unchanged weights plus a fresh pageable host
  allocation to land in. Only the wake direction did useful work.
- Handing the caching allocator multi-GB blocks to free and re-request
  each request fragments it. That is an OOM tens of requests into a run
  rather than on the first — the hardest kind to attribute.

## Modifications

`ModuleResidency` makes the host copy canonical: filled once, wake refills
the GPU replica from it, sleep only drops the replica. Because residency
mutates the tensor object in place rather than rebinding module
attributes, tied weights stay tied across a round trip (covered by a test).

Under serial offload, DIT/DAV now load **on the host** and only ever
upload a replica on demand, so the checkpoint never touches the GPU at
startup and there is never a device-to-host copy at all.

`move_module_to_device` has no callers left and is removed.

## Accuracy Test

Weights are moved, never transformed, and the dtype is unchanged, so
generated audio is bit-identical. `test_residency_sleep_and_wake_preserve_weights_and_state`
asserts the round trip preserves values.

## Benchmark & Profiling

Removes one full device-to-host weight copy per request. Absolute numbers
need a 24GB card; #1741 adds the logging that measures it.

## Checklist

- [x] Add unit tests.

### DIFF
--- a/sglang_omni/models/minimax_music3/acoustic.py
+++ b/sglang_omni/models/minimax_music3/acoustic.py
@@ -37,7 +37,7 @@ from .constants import (
 from .dav import MiniMaxMusic3DAV, remove_weight_norm, select_decoder_state
 from .dit import MiniMaxMusic3DIT
 from .payload_types import MiniMaxMusic3State
-from .serial_offload import get_coordinator, move_module_to_device
+from .serial_offload import ModuleResidency, get_coordinator
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


AkazaAkane:

PR #1741, opened 2026-08-26
[Fix] Report GPU occupancy across MiniMax Music 3 residency transitions

AkazaAkane:
## Motivation

Stacked on #1740 (part 4 of 4 on top of #1619). **Merge #1738, #1739 and #1740 first.**

Nobody has published what this mode actually frees, and the flag's name
promises more than the mechanism delivers. Offloading *weights* does not
release the SGLang KV cache: `MiniMaxMusic3EngineBuilder.generation_defaults`
sets `mem_fraction_static: 0.50`, which reserves half the card for weights
**plus** the KV pool for the life of the AR process. Moving
`model_runner.model` to the host releases the parameter tensors only — the
token-to-KV pool and `req_to_token_pool` stay resident across every
handoff. On a 24GB card that is roughly 12GB that never moves.

Whether stage offload is worth its cost (serialized AR/DIT, no CUDA
graphs, no `torch.compile`) depends entirely on how much of the card it
really frees. That number should come from a serving run, not an estimate.

## Modifications

Log, on every residency transition:

- free / total device memory (`torch.cuda.mem_get_info`)
- `torch.cuda.memory_allocated` and `memory_reserved`

plus each module's weight footprint when its residency is created. The
gap between torch's numbers and the device's is what makes memory held
outside torch's allocator visible rather than inferred.

Nothing is gated on these numbers; this PR only makes them observable.

## Accuracy Test

Logging only — no effect on generated audio.

## Checklist

- [x] Add unit tests. (existing residency tests cover the labelled paths; the values themselves need a GPU)

### DIFF
--- a/sglang_omni/models/minimax_music3/acoustic.py
+++ b/sglang_omni/models/minimax_music3/acoustic.py
@@ -224,8 +224,8 @@ class MiniMaxMusic3AcousticDecoder:
         self._residency: tuple[ModuleResidency, ...] = ()
         if self._serial_offload:
             self._residency = (
-                ModuleResidency(self.dit, self.device, resident=False),
-                ModuleResidency(self.dav, self.device, resident=False),
+                ModuleResidency(self.dit, self.device, resident=False, label="dit"),
+                ModuleResidency(self.dav, self.device, resident=False, label="dav"),
             )
 
     @property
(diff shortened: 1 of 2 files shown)


AkazaAkane:

PR #1745, opened 2026-08-26
[Refactor] Unify stage factory device/gpu_id contract across models

Lenoplus42:
## Motivation

Process-level replicas (#1307) place a stage on a specific GPU by injecting the placement-resolved `gpu_id` into the stage factory — but only when the factory's signature declares it, and only usefully when the factory then does the same thing with it as every other factory.

Before this PR there were four coexisting behaviors: the shared engine builder resolved device+gpu_id and pinned the type into ServerArgs; encoder/vocoder factories resolved a possibly index-less spec and let torch pick the card at `.to()` time; three models' thinker factories resolved then kept only the integer, so ServerArgs never saw the operator's device intent; and one factory dropped `device` entirely and defaulted the card to 0. 

Any of the last three silently lands a replica on the wrong card.

This PR makes the resolution behavior uniform, enforces it with sweeping contract tests so new models cannot regress it, and thereby makes every model's GPU process safely replicable.

## Modifications

**One resolving entry**

- **`resolve_concrete_device(device, gpu_id)` is now the single entry every GPU stage factory goes through**: it validates the requested device type against the host (e.g. an explicit `"cuda"` on a no-cuda host fails with a clear error, instead of deep inside the engine), and fills a missing index by asking which card the process is already on instead of assuming 0.
- `pin_resolved_device_type(overrides, resolved_type)` (extracted from `SGLangGenerationEngineBuilder.build()`) writes the resolved type into ServerArgs overrides and rejects a contradicting operator `server_args_overrides["device"]`.
- All component (encoder/vocoder/preprocessing) factories use `str(resolve_concrete_device(...))`.
- qwen3_omni's code2wav factory had a hand-copied duplicate of the current-card fallback; it now calls `resolve_concrete_device`. `ming_omni/pipeline/stages.py` is deleted — it imports modules that no longer exist and nothing references it.

**Signature contract (earlier commits, unchanged)**

- Every GPU-placed stage factory declares `device`/`gpu_id`, both defaulting to `None`; a device string names the type only (an embedded index is rejected); configs never set `factory.device`; a GPU-placed stage whose factory lacks `gpu_id` is rejected at resolution time.

**Test:**

- `tests/unit_test/test_stage_device_contract.py` sweeps every model x topology x stage automatically (new models are picked up without registration): signatures, config device strings, forwarding into `resolve_concrete_device` under mocks (a factory regressing to the lower-level helper fails), the real `build()` chain intercepted at `pre_infra_setup` for all 16 engine factories, and a ServerArgs-pinning assertion for the four self-built AR factories.
- New replica sweep: for each of the 32 model topologies, one GPU process is programmatically given `num_replicas=2, replica_devices=[2, 3]`, then driven through the real `prepare_pipeline_runtime` → stage-launch-spec → `resolve_factory_signature_args` chain, asserting each replica instance's factory receives its own card. The qwen colocated-speech topology is asserted to keep rejecting replicas, as its own placement policy already mandates. Reverse-verified: removing one factory's `gpu_id` parameter turns exactly that model's case red.

## Related Issues

[#1307](https://github.com/sgl-project/sglang-omni/issues/1307#issuecomment-5274759081), comment thread item 1, which remains as the main blocker of process replica.

## Accuracy Test

Check comment below on e2e test for 6 models.

## Benchmark & Profiling

N/A — no hot-path or scheduling behavior changes.

## Checklist

- [x] Add unit tests.


Lenoplus42:
Progress: so far, we've mainly refactored on config & signature level. After some reviewing, we realized this is not enough for supporting process replica replacement to fully work, as some factories have their own logics and may simply not use the correct device & gpu we passed in, for example.

Next step is deeper refactoring on factories. The final goal for this refactor is to remove the gpu placement blocker and fully support process replica for all models.

sglang_omni/config/runtime.py
```diff
@@ -174,6 +174,13 @@ def resolve_factory_signature_args(
             f"{factory_name!r} does not declare a gpu_id parameter"
         )
 
+    if defaults.get("gpu_id") is not None and "gpu_id" not in sig.parameters:
```
stephenkgli, line 175: Zonos2 preprocessing is still configured with `gpu=0`, but `create_preprocessing_executor()` has no `gpu_id` parameter. This guard therefore rejects the default and multi-GPU pipelines before the factory runs. Remove the GPU placement or give the factory the standard no-op `device`/`gpu_id` parameters.

sglang_omni/config/runtime.py
```diff
@@ -174,6 +174,13 @@ def resolve_factory_signature_args(
...
+        raise ValueError(
+            f"{factory.__module__}.{factory.__qualname__} is placed on a GPU "
+            "but its signature has no gpu_id parameter, so placement can "
+            "never reach it"
+        )
+
     for name, value in defaults.items():
```
stephenkgli, line 182: Placement-owned values are only rejected in `stage_factory_kwargs()`. `FactoryArgs` extras can still set `factory.gpu_id`; because defaults are injected only when absent, `factory.gpu_id=4` overrides planner `gpu_id=2`, moving the model away from the stage metadata, budget, and transport GPU. Reject placement-owned keys in typed factory args too.

sglang_omni/models/audar_tts/stages.py
```diff
@@ -239,13 +239,22 @@ def create_tts_engine_executor(
...
 
+    from sglang_omni.utils.device import resolve_concrete_device
+
+    concrete_device = resolve_concrete_device(device, gpu_id)
+    if concrete_device.type not in ("cpu", "cuda"):
+        raise ValueError(
+            "Audar-TTS llama.cpp engine runs on cuda or cpu only; resolved "
+            f"device={concrete_device}"
+        )
+    main_gpu = concrete_device.index if concrete_device.type == "cuda" else 0
```
stephenkgli, line 250: For `device="cpu"`, `main_gpu` becomes 0 but `n_gpu_layers` remains -1. llama-cpp interprets -1 as offloading every layer, so this CPU path still runs on GPU 0. Force `n_gpu_layers=0` for CPU or reject the combination.

sglang_omni/models/dots_tts/stages.py
```diff
@@ -370,14 +364,19 @@ def _preprocess(payload: StagePayload) -> StagePayload:
...
     max_concurrency: int = 8,
     max_batch_size: int = 1,
     max_batch_wait_ms: float = 4.0,
 ) -> SimpleScheduler:
-    worker_device = _device(device, gpu_id)
-    codec = load_dots_audio_codec(model_path, device=worker_device)
+    from sglang_omni.utils.device import resolve_concrete_device
+
+    if not torch.cuda.is_available():
+        raise RuntimeError("dots.tts requires CUDA")
```
edwingao28: [P1] Finish migrating the CPU-only test fixtures

The normal unit-test job explicitly sets `CUDA_VISIBLE_DEVICES=""`, but `test_reference_executor_defaults_to_per_request` and `test_reference_executor_stop_closes_batch_worker` in `tests/unit_test/dots_tts/test_reference_encode_batching.py` still call this factory with `device="cpu"` and a stubbed codec. This new unconditional guard raises before the stub is used. In a focused comparison, both real tests pass with the base factory and fail with this factory (unavailable platform/AR imports were stubbed).

There is a second instance of the same incomplete CPU-test migration: `_run_s2pro_engine_with_fake_buffers` in `tests/unit_test/fishaudio_s2_pro/test_pipeline.py` and the generation-bootstrap test in `tests/unit_test/voxtral_tts/test_pipeline.py` now pass `device="cuda"` without mocking `current_platform.device_type`. The new validating builder rejects those calls on the CPU job before their mocked infrastructure runs; an isolated factory→builder comparison reproduces that base/head difference too.

Could we preserve these hardware-independent checks by updating the availability/platform fixtures (and supplying an explicit index where needed), or by keeping the dots validation at the actual CUDA-loading boundary? Marking only the new contract cases as `accelerator` does not protect these existing tests.

sglang_omni/models/zonos2/stages.py
```diff
@@ -235,6 +238,7 @@ def create_sglang_omni_tts_engine_executor(
     max_running_requests: int = 16,
     cuda_graph_max_bs: int = 16,
     server_args_overrides: dict | None = None,
+    **_: Any,
```
edwingao28: [P2] Reject unsupported factory options instead of discarding them

Adding `**_` makes `apply_typed_stage_kwargs()` treat every configuration option as supported, but this factory never forwards or checks those captured values. For example, `stages.tts_engine.factory.max_new_tokens=100` previously raised a useful configuration error; this revision accepts it while `Zonos2EngineBuilder` never receives it. The new catch-alls in the speaker encoder and vocoder have the same issue.

Could we remove the catch-alls, or explicitly reject unknown kwargs? The named `device`/`gpu_id` parameters already satisfy placement injection. I verified the reject→accept change using the actual runtime validator and the base/head signatures.

sglang_omni/models/higgs_tts/config.py
```diff
@@ -70,7 +68,14 @@ class HiggsTtsPipelineConfig(PipelineConfig):
...
             factory_path=f"{_PKG}.stages.create_vocoder_executor",
-            factory=FactoryArgs(device="cuda"),
+            factory=FactoryArgs(
+                compile_decode=False,
+                # Before the steady cursor is established, a decode window is
+                # bounded by the default 75-row stride plus its 75-row
+                # follow-up. Capture that complete finite domain so terminal
+                # flushes cannot silently fall back to eager execution.
+                decode_cuda_graph_frame_counts=tuple(range(1, 151)),
+            ),
```
edwingao28: [P3] Keep one source of truth for the vocoder defaults

Non-blocking: these `compile_decode` and `decode_cuda_graph_frame_counts` defaults (including their explanatory comment) already exist in `stage_factory_kwargs("vocoder")` below. The typed factory values override that hook, so retaining both means a future edit to the hook can silently have no effect. Could we keep the defaults in one place, e.g. here in `FactoryArgs`, and remove the redundant vocoder branch from the hook?

edwingao28 (commented):
The shared device/gpu_id resolution is a useful direction. I found two issues worth fixing (CPU-only test migration and silently ignored Zonos2 options), plus a non-blocking duplication cleanup. I have not repeated the three existing review threads.

Validation: 156 config/runtime-adapter tests passed locally. Focused base/head reproductions confirm the CPU-test regressions and the Zonos2 validator behavior; these used isolated import/boundary stubs where the local environment lacks SGLang, not a full SGLang or multi-GPU run.

sglang_omni/config/runtime.py
Lenoplus42: Fixed in c189cce8. create_preprocessing_executor now takes the standard no-op device/gpu_id pair (same as qwen3_omni's mm_aggregate)

sglang_omni/config/runtime.py
Lenoplus42: Fixed in c189cce8. resolve_stage_typed_kwargs now rejects placement-owned keys coming through factory.* (declared fields and extras alike), with the same error shape as the hook path. Regression test added; verified no shipped config legitimately used these keys.

sglang_omni/models/audar_tts/stages.py
Lenoplus42: Fixed in 89784548, took the "force" option: a cpu resolution zeroes n_gpu_layers — the -1 default is our signature's default rather than an explicit user choice, so overriding it under cpu semantics seemed friendlier than rejecting. Test asserts llama.cpp receives n_gpu_layers=0, main_gpu=0.

sglang_omni/models/dots_tts/stages.py
Lenoplus42: Fixed in 89784548 (completed in ac3c22ca). The dots CUDA check applies only when the resolved device is cuda, restoring base behavior for explicit device="cpu"; the fishaudio helper and voxtral bootstrap test pin the platform layer, and voxtral supplies an explicit gpu_id per your note. Chasing this under CUDA_VISIBLE_DEVICES=-1 (faithful to the CI unit-test job) also surfaced two latent order-dependent crashes in the resolver's current-card fallback, fixed in ac3c22ca. Verified: the three files go 23 failures at the pre-fix commit → 0 after, and the full unit suite's hidden-CUDA failure set now matches upstream/main line-for-line. Thanks for the focused base/head repro — most valuable catch of the round.

sglang_omni/models/zonos2/stages.py
Lenoplus42: Fixed in 89784548. All three **kwargs catch-alls removed; unknown factory.* options raise a configuration error again (regression test added). Checked that the only typed option shipped configs pass (dtype on tts_engine) lands on a named parameter.

sglang_omni/models/higgs_tts/config.py
Lenoplus42: Fixed in c189cce8, went with your suggestion: FactoryArgs is the single source, the redundant vocoder branch in stage_factory_kwargs is gone, and the tests that asserted the old source moved with it.

sglang_omni/models/fun_asr/config.py
```diff
@@ -28,6 +33,7 @@ class FunASRPipelineConfig(PipelineConfig):
             name="asr",
             process="asr",
             factory_path=f"{_PKG}.stages.create_sglang_fun_asr_executor",
+            factory=FactoryArgs(),
```
stephenkgli: This line is redundant?

Lenoplus42: Confirmed and fixed releavant occurance

Lenoplus42:
Will run few rounds of e2e test and put the result here before merging.

Lenoplus42:
### Accuracy Test End-to-end

Validated at `e39f4cb1` (this branch with upstream/main merged up to `f45b1fb2`) on H200 shared dev box. 

(After the later merge of upstream/main `325cc4ae` (`ac878c1a`, then `4f59580d` for one test fix) the CPU unit suite and the higgs TTS CI stage (the only e2e model touched by that merge) were re-run: unit suite 6598 passed with one timing-flaky readiness timeout (passes 3/3 in isolation rerun). higgs TTS CI stage 1 4 passed — WER 1.03%, SIM 68.72, UTMOS 4.165.)

Correctness is verified on two layers. 
**Placement**: every GPU stage's `gpu_id` in the server log equals the configured card, `nvidia-smi` shows that process on that physical card, replicas each own their card, every replica serves requests, no `does not declare a gpu_id` / `Omni owns placement` error, and the cards return to 0 MiB after teardown. 
**Output**: the CI suite's own thresholds where a suite exists, otherwise a sanity request.

Models were chosen so that every way a factory used to pick its device before this PR is exercised at least once.

- **Shared engine builder** (the 16 SGLang engine factories: whisper and arkasr `asr`, higgs / moss_tts_local / ming_tts `tts_engine`, and so on): the factory hands `device`/`gpu_id` to `SGLangGenerationEngineBuilder`, which resolved them and pinned the type into ServerArgs. Already close to the contract; the PR extracts that pinning into `pin_resolved_device_type` and routes it through the shared resolver.
- **Component factories** (26 encoder / vocoder / preprocessing factories, e.g. higgs `audio_encoder` and `vocoder`, moss_tts_local `preprocessing` and `vocoder`, ming_tts `reference_encode`, the fishaudio / dots / cosyvoice vocoders): they used to call the lower-level `resolve_device_spec` directly and could end up with an index-less `"cuda"` that let torch pick the card at `.to()` time. Now they call `resolve_concrete_device`, which always yields an indexed device.
- **Thinkers that kept only the index** (qwen3_omni `thinker`/`talker_ar`, ming_omni, llada2_uni): they resolved the device and then passed only the integer on, so ServerArgs never saw the operator's device intent. Now they keep the resolved type and pin it.
- **Factories with their own inline device logic** (ming_tts `audio_decode`, audar_tts): each merged `device` and `gpu_id` with hand-written code instead of any helper. Both now go through the shared resolver and keep only their backend-specific capability check.

Result below:

| Model | Why it is representative | Placement | Output |
|---|---|---|---|
| qwen3_omni | multi-process replicas; index-only thinker/talker path | Qwen3-Omni CI stage 11: `code2wav@r0/@r1 → device 0/1`, thinker gpu 0, talker gpu 1 | stage 11 PCM ≤ 2 LSB vs single instance; stage 2 (talker speed + WER + SIM) 5 passed, WER 1.95% (gate 2.49%), UTMOS 4.43 |
| whisper_asr | **could not be replicated before this PR** (factory had no `gpu_id`); single stage on the shared builder | `processes.asr: num_replicas=2, replica_devices=[0,1]` → `asr@r0`/`@r1` on two cards (pid-bridged), 4 requests round-robin r0/r1/r0/r1; single instance placed on gpu 1 lands on gpu 1 | ASR CI (SeedTTS EN): 588/588 requests, corpus WER 1.38% = the calibrated reference |
| arkasr | could not be replicated before; no CI suite (manual path) | same 2-replica yaml; `asr@r0`/`asr@r1` verified end to end (stage log pid → host pid via `/proc/<pid>/status NSpid` → `nvidia-smi` card): one replica per physical card, round-robin dispatch | cookbook SeedTTS EN benchmark (1088 requests, concurrency 16) on the 2-replica server: 0 failed, corpus WER 1.11%, identical to upstream/main serving a single instance on the same box (corpus WER 1.11%, same max per-sample WER 28.57%); the coordinator dispatched exactly 1088 requests to each replica |
| higgs_tts | multi-stage TTS; configs used to hard-code `device="cuda"` | TTS CI DP2, encoder + engine per worker on its card | TTS CI stage 1: 4 passed — WER 0.94% (ref 1.16%), speaker SIM 66.24 (ref 66.07), UTMOS 4.166 (ref 4.164) |
| moss_tts_local | largest refactor (split topology, vocoder in its own process, `codec_gpu`) | TTS CI DP2, pipeline + separate vocoder process per worker | TTS CI stage 1: 4 passed — WER 2.24% (ref 2.53%), SIM 65.35 (ref worst-of-5 64.07), UTMOS 3.958 (ref 3.951). A first run scored SIM 63.67; upstream/main on the same box scored 64.83; the rerun 65.35. Per-sample diff has std 8 (temperature 0.8, no seed), so the spread is sampling noise, not a regression |
| ming_tts | its audio_decode factory used to merge `device`/`gpu_id` with its own inline logic instead of the shared helper; it now goes through the shared resolver and keeps its CUDA-only check | `--reference_encode.gpu 1 --audio_decode.gpu 1` with the engine on gpu 0: engine 104.9 GB on card 0, aux 4.1 GB on card 1, no stray context | cookbook SeedTTS EN benchmark (1088 non-streaming text-only requests, concurrency 8) on this cross-GPU placement: 0 failed, corpus WER 0.91% (cookbook reference on a single H200: 0.93%), RTF 0.237 (ref 0.216), 7.16 req/s (ref 7.84) |

Before/after: the same whisper 2-replica yaml is refused at startup on upstream/main (`Stage 'asr' uses processes.replica_devices, but factory ... does not declare a gpu_id parameter`, the limitation `docs/basic_usage/process_topology.md` documents) and starts on this branch with each replica on its own card.

Captured an issue: 
- When an ASR stage is placed on a non-default card, its process also opens a bare ~518 MiB CUDA context on the default card (whisper and arkasr; not ming_tts). It appears at readiness/first request, not at model load; the request path of these models is untouched here, and on main the stage can only sit on card 0 so this bug is invisible on main (but definitely visible after this PR enables full process replica). Planned to open it as a separate issue, and I will follow-up on it myself.

sglang_omni/models/qwen3_tts/stages.py
```diff
@@ -28,7 +28,6 @@
 from sglang_omni.scheduling.simple_scheduler import SimpleScheduler
 from sglang_omni.scheduling.threaded_simple_scheduler import ThreadedSimpleScheduler
 from sglang_omni.utils.checkpoint import resolve_checkpoint as _resolve_checkpoint
-from sglang_omni.utils.device import resolve_device_spec
```
stephenkgli, line 32: Qwen3-TTS split preprocessing fails at startup

The [documented split configuration](https://github.com/sgl-project/sglang-omni/blob/e3f9bed03b9414334ef3fcbcbe62332acaea728a/docs/cookbook/qwen3_tts.md#L231) (`--preprocessing.process tts_frontend`) sets `load_frontend=True`, calling `_load_standalone_preprocessing_context()`. That helper still calls [`resolve_device_spec(...)`](https://github.com/sgl-project/sglang-omni/blob/e3f9bed03b9414334ef3fcbcbe62332acaea728a/sglang_omni/models/qwen3_tts/stages.py#L163), but this PR removes its module-level import, causing `NameError` during worker initialization. The default colocated topology skips this branch.

Lenoplus42: resolved by letting preprocessing stage call resolve_concrete_device when split, just like every other gpu stages

Lenoplus42:
Runned e2e test for Qwen3-TTS split & AuK, works good.

stephenkgli (approved):
LGTM

PR #1747, opened 2026-08-26
Nemotron Voicechat offline

PopSoda2002:
[Support for nvidia/NVIDIA-NemotronLabs-VoiceChat-11B. WIP in one week.

https://huggingface.co/nvidia/NVIDIA-NemotronLabs-VoiceChat-11B

# Support NVIDIA-NemotronLabs-VoiceChat-11B
Adds `nemotron_voicechat`: a from-scratch implementation of NVIDIA's 11B
full-duplex speech-to-speech model, served both as an offline pipeline and as a
live duplex session over the existing OpenAI-Realtime WebSocket endpoint.

23 files, ~2.5K lines, no third-party code vendored.

## What the model is

A frame-locked 12.5 Hz conversational model — every 80 ms of caller audio
advances the whole stack exactly one step, and the model itself decides when to
speak (there is no external VAD and no turn boundary). Four components:

| Component | Weights | Role |
|---|---|---|
| Conformer perception | 640 tensors / 614M | 1280 samples @16 kHz → one acoustic frame |
| NemotronH thinker | 357 tensors / 9.5B | one text token + one function token per frame |
| EAR-TTS talker | 397 tensors / 639M | text + previous frame's codes → hidden state |
| MoG head | 21 tensors / 159M | hidden → continuous 512-d latent |
| RVQ-VAE codec | 214 tensors / 200M | 31 codes → 1764 samples @22.05 kHz |

The checkpoint is a single 44 GB fp32 `model.safetensors` with a NeMo/Hydra
`config.json` that carries no `architectures`, `model_type` or tokenizer, so
`hf_config.py` synthesises a config shim that subclasses `NemotronHConfig`
(subclassing, not wrapping — sglang's `mamba2_config()` does a bare
`isinstance` check).

## Architecture

```
offline:  preprocessing → perception → thinker ─┬→ decode (terminal)
                                                └→ talker → code2wav (terminal)
                                        (thinker streams text tokens to talker;
                                         talker streams code chunks to code2wav)

live:     FrameSession.push_audio(1280 samples)
            → streaming perception  (in-process)
            → thinker engine        (own process, sglang AR)
            → talker engine         (own process, sglang AR)
            → streaming codec       (in-process)
          exposed at /v1/realtime via DuplexRealtimeSession
```

Both AR stages run under `OmniScheduler` as real sglang engines with paged KV.
The talker's Gemma3 backbone is sglang's `Gemma3ForCausalLM`; the embedding,
gated fusion, MaskGIT and RVQ machinery live in `EarTtsTalker` and are driven
from the model-runner hooks (`before_decode` fuses the frame's inputs,
`post_decode` runs MaskGIT and emits the code chunk).

### Streaming, and how each stage is gated

Each stage's next step depends on data that may not have arrived yet, and the
two backbones need opposite solutions:

- **talker** (pure attention) gates *after* `super().get_next_batch_to_run()`
  and rolls back `prepare_for_decode` when the thinker's next token is missing
  (`_rollback_decode_prep_after_skip`, modelled on qwen3's talker).
- **thinker** (NemotronH, Mamba-hybrid) cannot roll back a prepared step, so it
  gates *before* preparation, checking `running_batch` and `last_batch`.

### Incremental codec decoding

The RVQ-VAE decoder is frame-causal on the left but its inverse STFT's
overlap-add makes the last ~184 samples of a truncated render differ from the
full one. `StreamingCodec` decodes a 16-frame trailing window and holds back
256 samples (11.6 ms), re-emitting them once the next frame exists. Measured
against whole-utterance decoding: **max 1.8e-04, mean 8.6e-07** — below the
16-bit quantisation step.

### Streaming perception

`StreamingPerception` keeps per-layer KV caches (bounded at
`att_context_size=[70, 0]`), the subsampling and depthwise-conv left windows,
and the STFT overlap plus preemphasis carry. `AudioPerception.forward` is that
loop run to completion, so there is exactly one forward implementation for both
offline and live use.

## Numerical verification

Every component was verified against the NeMo reference implementation, run as
an oracle (read for behaviour, never copied):

| Component | Result |
|---|---|
| Conformer perception | 640/640 tensors, 613,767,056 params, **0.000e+00** |
| RVQ-VAE codec | 107/107 decoder tensors, 108,098,432 params, **0.000e+00** |
| MoG head | 21/21 tensors, 158,606,336 params, **0.000e+00** (incl. CFG paths) |
| Talker `embed_subword` | 1.7e-07 |
| Talker backbone (per-frame, with cache) | 1e-06 |
| `generate_codes` (MaskGIT) vs reference `generate_step` | **31/31 codes identical** |
| Streaming vs whole-utterance perception | max 1.1e-04 (fp32 reduction-order noise; the streaming path is *closer* to fp64 truth than the whole-run one, which is 7.9e-04 off) |
| Streaming vs whole-utterance codec | max 1.8e-04 |

End-to-end, on `turn_taking.wav` (41.1 s, a genuine multi-turn conversation):

```
514 tokens, bit-identical to the offline path
audio 0.0071 rms global / 0.0174 voiced / 0.211 peak   (reference impl: 0.0063 / 0.0156 / 0.163)
```

Whisper transcribes the output as the conversation the thinker produced —
"Hello! How can I help you today? / I'm looking for healthy food options / …" —
with the model correctly staying silent (emitting PAD) while the caller speaks.

## Real-time budget

Fed at a true 80 ms cadence through the WebSocket:

```
first response.audio.delta   2.2 s   (includes one-off CUDA warmup)
513 blocks fed in            41.0 s  vs 41.1 s of audio — keeps up with real time
push_audio latency           median 15.3 ms, p95 24.6 ms   (budget: 80 ms)
per-frame cost               talker.step 12.3 ms + generate_codes 5.6 ms + codec 1.5 ms
```

## Deliberate choices worth reviewing

- **Classifier-free guidance is off** (`guidance_scale=0`) although the
  checkpoint config enables it at 0.2. A/B on the same input showed
  indistinguishable transcripts and levels (0.0158 vs 0.0169 voiced rms;
  the reference implementation itself measures 0.0156). CFG under sglang needs
  a lock-stepped cond/uncond request pair; deferred rather than half-built.
- **`max_running_requests=1`** for both engines. Upstream (vLLM-omni) is also
  batch=1 for this model.
- **A duplex session class separate from `RealtimeSession`.** The protocol,
  event schema and `/v1/realtime` mount contract are reused unchanged, but the
  907-line turn-based session is a VAD state machine whose every part (VAD
  auto-commit, per-turn requests, cancel/truncate machinery) must not run for a
  frame-locked model. ~40 lines of WebSocket receive/send are duplicated; a
  shared base class is a reasonable follow-up.
- **`DuplexEngines` hosts each engine in its own process** and sessions attach
  to them. Long term these should be pipeline-framework stage workers; this is
  a pragmatic intermediate that matches production isolation.

## Known limitations

- Single speaker (`Aria`, the checkpoint's pre-baked prompt latents); no
  speaker selection API.
- The `function_head` channel (tool calls: `call`/`sotc`/`eotc`/`eotr`) is
  decoded and carried through, but nothing acts on it.
- Barge-in works implicitly — the model is duplex and processes caller audio
  while speaking — but there is no explicit interrupt API or stress test.
- The first frame of a session pays a ~7 s CUDA warmup; a server should push
  one dummy frame at startup.
- The pipeline config declares the streaming stage edges and was validated at
  the scheduler level; a full `serve` launch through the multiprocess framework
  has not been exercised.

](https://github.com/sgl-project/sglang-omni/pull/1747#issue-5254424450)


sglang_omni/models/nemotron_voicechat/config.py
```diff
@@ -0,0 +1,96 @@
+from pydantic import Field
+from typing import ClassVar
+
+from sglang_omni.config import (
+    PipelineConfig,
+    StageConfig,
+    StageResourceConfig,
+    StageRuntimeConfig,
```
draftbk: I tried the latest head (d74b9097), and it currently fails to import because main changed the pipeline config API:
```
ImportError: cannot import name 'StageResourceConfig'
```
`nemotron_voicechat/config.py` still seems to use the old StageResourceConfig / StageRuntimeConfig API, so this needs to be migrated.

Related PR: https://github.com/sgl-project/sglang-omni/pull/1494

draftbk: Done: https://github.com/sgl-project/sglang-omni/pull/1747/changes/059bf654ad67276086faaaf2a4898f2277cecfce

draftbk:
The latest offline example works for me. I was able to reproduce the text and audio output end to end.
Nit: The description still mentions realtime/WSS and barge-in. Maybe we should update it to match the current offline scope? @PopSoda2002

PopSoda2002:
> The latest offline example works for me. I was able to reproduce the text and audio output end to end. Nit: The description still mentions realtime/WSS and barge-in. Maybe we should update it to match the current offline scope? @PopSoda2002

Yeah sure let me modify it

sglang_omni/models/nemotron_voicechat/talker_model_runner.py
```diff
@@ -0,0 +1,186 @@
...
+                input_embeds=torch.cat(rows, dim=0).to(self.model._fusion_buffer.dtype),
+                input_embeds_are_projected=True,
+            ),
+        )
+
+    def post_prefill(self, result, forward_batch, schedule_batch, requests) -> None:
+        del result, forward_batch, schedule_batch
+        for request in requests:
+            inputs = request.data.talker_model_inputs
+            first_code = torch.zeros(
```
gaoyang07: Here `post_prefill` sends an all-zero row at the start, and first_step drops thinker token 0. That means frame 0 is always just a fake row, and never actually generate codes for token 0.

PopSoda2002: Thanks！！ Fixed in the new commit

sglang_omni/models/nemotron_voicechat/config.py
```diff
@@ -0,0 +1,95 @@
...
+    StageConfig,
+)
+
+# The ckpt is all saved in float32.
+MODEL_DTYPE = "float32"
+
+MODEL_STAGES_PREFIX = "sglang_omni.models.nemotron_voicechat.stages"
+
+
+def nemotron_voicechat_stages_factory() -> list[StageConfig]:
```
lijrjyan, line 20: `gpu_memory_fraction` is silently dropped for the engine stages. `runtime.py:153` injects it only into factories declaring `total_gpu_memory_fraction`, which `create_thinker_executor` / `create_talker_executor` don't. So current parameter bounds nothing and might be misleading.

Consider deleting `gpu_memory_fraction` or applying by declaring the parameter on both factories, pass it to the builder and return it from `infra_kwargs()`.

sglang_omni/models/nemotron_voicechat/config.py
```diff
@@ -0,0 +1,95 @@
...
+            factory_path=f"{MODEL_STAGES_PREFIX}.create_preprocessing_executor",
+            next="perception",
+        ),
+        StageConfig(
+            name="perception",
+            process="pipeline",
+            factory_path=f"{MODEL_STAGES_PREFIX}.create_perception_executor",
+            factory=FactoryArgs(dtype=MODEL_DTYPE),
+            gpu=0,
+            gpu_memory_fraction=0.12,
```
lijrjyan: Measured on H200, this stage declared 0.12, which is 16.8 GB. This stage actually used 5.6 GB.

sglang_omni/models/nemotron_voicechat/config.py
```diff
@@ -0,0 +1,95 @@
...
+        ),
+        EngineStageConfig(
+            name="thinker",
+            process="pipeline",
+            factory_path=f"{MODEL_STAGES_PREFIX}.create_thinker_executor",
+            # NemotronH runs under SGLang, whose rmsnorm kernel has no float32
+            # path; the rest of the chain keeps the checkpoint's precision.
+            factory=FactoryArgs(dtype="bfloat16"),
+            gpu=0,
+            gpu_memory_fraction=0.52,
```
lijrjyan: Measured on H200,this stage declared 0.52, which is 73.0 GB. This stage actually used 72.6 GB.

sglang_omni/models/nemotron_voicechat/config.py
```diff
@@ -0,0 +1,95 @@
...
+        ),
+        EngineStageConfig(
+            name="talker",
+            process="talker",
+            factory_path=f"{MODEL_STAGES_PREFIX}.create_talker_executor",
+            # Backbone only: FlashAttention takes fp16/bf16/fp8, and the
+            # talker's own layers cast themselves back to float32.
+            factory=FactoryArgs(dtype="bfloat16"),
+            gpu=0,
+            gpu_memory_fraction=0.22,
```
lijrjyan: Declared 0.22, 30.9 GB, actually used 47.6 GB. Exceed the bound.

sglang_omni/models/nemotron_voicechat/config.py
```diff
@@ -0,0 +1,95 @@
...
+            next="code2wav",
+            stream_to=["code2wav"],
+        ),
+        StageConfig(
+            name="code2wav",
+            process="talker",
+            factory_path=f"{MODEL_STAGES_PREFIX}.create_code2wav_executor",
+            factory=FactoryArgs(dtype=MODEL_DTYPE),
+            gpu=0,
+            gpu_memory_fraction=0.08,
```
lijrjyan: Declared 0.08, 11.2 GB, actually used 3.6 GB.

sglang_omni/models/nemotron_voicechat/engine_builder.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def _shim_dir(name: str, source: Path) -> Path:
+    shim = CACHE_ROOT / name
+    shim.mkdir(parents=True, exist_ok=True)
+    for entry in source.resolve().iterdir():
+        link = shim / entry.name
+        if entry.name != "config.json" and not link.exists():
+            link.symlink_to(entry)
+    return shim
```
lijrjyan, line 54: Links are created only on first load. Switching checkpoints rewrites config.json but keeps the old weight link, thus the new checkpoint is silently ignored.

sglang_omni/models/nemotron_voicechat/engine_builder.py
```diff
@@ -0,0 +1,240 @@
...
+
+        Which token means begin, end and pad is the checkpoint's to say — the
+        three are easy to confuse here, since this tokenizer's padding token is
+        the frame-locked silence marker rather than anything called "pad".
+        """
+        stt = json.loads((self._source / "config.json").read_text())["model"]["stt"][
+            "model"
+        ]
+        tokenizer = AutoTokenizer.from_pretrained(
+            stt.get("pretrained_llm", "nvidia/NVIDIA-Nemotron-Nano-9B-v2")
```
lijrjyan: Not sure whether this is good practice here. Offline with an empty cache it currently fails with a `LocalEntryNotFoundError` that doesn't name the repo.
 
It might be better to document the pre-download `hf download nvidia/NVIDIA-Nemotron-Nano-9B-v2 config.json tokenizer.json tokenizer_config.json special_tokens_map.json` and raise an error that names the missing repo.

sglang_omni/models/nemotron_voicechat/request_builders.py
```diff
@@ -0,0 +1,146 @@
...
+    "Start the conversation by greeting the user."
+)
+
+
+def _ar_request(
+    payload: StagePayload, *, input_ids: list[int], max_new_tokens: int, vocab_size: int
+) -> SGLangARRequestData:
+    sampling_params = SamplingParams(
+        max_new_tokens=max_new_tokens,
+        temperature=0.0,
```
lijrjyan, line 35: `temperature` is hardcoded to 0 here without documentation, so a request's temperature is silently ignored. 

It would be better to document that we hardcode `temperature = 0`, and tell users what parameters actually control sampling randomness.

sglang_omni/scheduling/sglang_backend/request_data.py
```diff
@@ -31,6 +31,9 @@ class SGLangARRequestData(ARRequestData):
     tts_pad_embed: Any = None
     tts_eos_embed: Any = None
     thinker_chunks_done: bool = True
+    # Length of the instruction a request opens with, when its prefill carries
+    # one; 0 when it does not.
+    prompt_len: int = 0
```
lijrjyan: A bit confuse about this variant. It is not used and should be deteled.

sglang_omni/models/nemotron_voicechat/stages.py
```diff
@@ -0,0 +1,206 @@
...
+    )
+
+
+def create_code2wav_executor(model_path, *, dtype=None, device=None):
+    device = resolve_device_spec(device)
+    generation = _speech_generation_config(model_path)
+    weights = load_weights_by_prefix(model_path, prefix=("tts_model.audio_codec.",))
+    # The codes that mark an utterance's edges live beside the codec rather
+    # than inside it, so they are read under their own prefix.
+    markers = load_weights_by_prefix(model_path, prefix=("tts_model.",))
```
lijrjyan: Currently, this second call `load_weights_by_prefix(model_path, prefix=("tts_model.",))` walks every `tts_model.*` tensor, which could consume large time and gpu memo, just to read two vectors. Might be better way to reduce the consumption.

sglang_omni/pipeline/stage_workers.py
```diff
@@ -407,6 +407,13 @@ def stage_process_main(
 ) -> None:
     """Subprocess entrypoint: construct stage(s) from *spec* and run them."""
     logging.basicConfig(level=logging.INFO, stream=sys.stdout)
+    # Stage processes have no console of their own, so the only way to raise
+    # their verbosity is the environment the parent was launched with. Set the
+    # package logger rather than the root one: basicConfig is a no-op once an
+    # imported library has installed handlers.
+    stage_log_level = os.environ.get("SGLANG_OMNI_STAGE_LOG_LEVEL")
+    if stage_log_level:
+        logging.getLogger("sglang_omni").setLevel(stage_log_level.upper())
```
lijrjyan: Better to drop these line and remain the pr scope in model support.

lijrjyan (changes requested):
Besides requested changes inline, it still need to add correponding unit tests to avoid driftting, also documents.

sglang_omni/models/nemotron_voicechat/engine_builder.py
draftbk: Thanks! Fixed in https://github.com/sgl-project/sglang-omni/pull/1747/changes/4e7871c04dfcd1ddb4cdad1d551c6c024dd6b509.

sglang_omni/pipeline/stage_workers.py
draftbk: Fixed in https://github.com/sgl-project/sglang-omni/pull/1747/changes/f9f1c56710da905991e154d2e94249b852fd4a8b.

sglang_omni/models/nemotron_voicechat/stages.py
draftbk: Fixed in https://github.com/sgl-project/sglang-omni/pull/1747/changes/f9f1c56710da905991e154d2e94249b852fd4a8b.

sglang_omni/scheduling/sglang_backend/request_data.py
draftbk: Fixed in https://github.com/sgl-project/sglang-omni/pull/1747/changes/f9f1c56710da905991e154d2e94249b852fd4a8b.

sglang_omni/models/nemotron_voicechat/engine_builder.py
draftbk: Fixed in https://github.com/sgl-project/sglang-omni/pull/1747/changes/1adf8dba5ac3f878929f08e93bcbfb36be507e33

sglang_omni/models/nemotron_voicechat/request_builders.py
draftbk: Fixed in https://github.com/sgl-project/sglang-omni/pull/1747/changes/1adf8dba5ac3f878929f08e93bcbfb36be507e33

sglang_omni/models/nemotron_voicechat/config.py
draftbk: @PopSoda2002 I pushed a fix for the memory config mismatch in [13a395b2](https://github.com/sgl-project/sglang-omni/pull/1747/commits/13a395b2d2a48f6d3d916025e59a4c63a8d213f8). It removes the unused stage fractions and wires through engine.* overrides, keeping the existing talker default of 0.35. Tested on B200 with Triton. There may be a better approach, so could you take a look? Feel free to adjust or revert.

PopSoda2002: I feel it's ok to use your defaut value in mem_fraction_static, I tested in h100, it's fine, there is a config called total_reserve_bytes may help, but I think it would add more complexity

sglang_omni/models/nemotron_voicechat/config.py
PopSoda2002: explained below

PopSoda2002:
Hi @lijrjyan can you please review again?

sglang_omni/models/nemotron_voicechat/talker_model_runner.py
```diff
@@ -0,0 +1,174 @@
...
+    def post_prefill(self, result, forward_batch, schedule_batch, requests) -> None:
+        del result, forward_batch, schedule_batch
+        for request in requests:
+            inputs = request.data.talker_model_inputs
+            inputs["codes_rows"] = []
+            inputs["prev_codes"] = self._pad_codes()
+
+    def is_decode_batch_ready(self, schedule_batch) -> bool:
+        return all(
+            len(req._omni_data.pending_text_queue) > 0 for req in schedule_batch.reqs
```
gaoyang07: If a running request is aborted with an empty text queue, this gate prevents the forward needed to apply `to_finish`, leaving its KV and the only running slot occupied.

Talker indirectly couples request termination to the next decode step. Even though the request has already received an abort, an empty Thinker token queue prevents decode from executing, meaning the request cannot transition to the finished state, and its occupied running slot and KV cannot be released through the normal path.

Could we add a safe terminal cleanup path that does not require another text token? Simply bypassing this gate would still hit the unconditional popleft() below and turn cancellation into a batch failure.

lijrjyan (approved):
LGTM. Should solved @gaoyang07's problem before merge. Not blocking but better to have:
- Tests to avoid drifting during optimization.
- Correponding document and cookbook updates, including the instruction to download tokenizer `hf download nvidia/NVIDIA-Nemotron-Nano-9B-v2 config.json tokenizer.json tokenizer_config.json special_tokens_map.json`
- Re-run the talker/end-to-end numerical verification in the body.

lijrjyan:
> LGTM. Should solved @gaoyang07's problem before merge. Not blocking but better to have:
> 
> * Tests to avoid drifting during optimization.
> * Correponding document and cookbook updates, including the instruction to download tokenizer `hf download nvidia/NVIDIA-Nemotron-Nano-9B-v2 config.json tokenizer.json tokenizer_config.json special_tokens_map.json`
> * Re-run the talker/end-to-end parity on the final head with frame 0 included, and say which NeMo path it targets (offline recipe vs realtime).

pushed two commits:
- `8f0a017` adds two tests and docs updates.
- `66f5e1d` fix fp32 talker weights being rounded to bf16 at load.

sglang_omni/models/nemotron_voicechat/talker_model_runner.py
PopSoda2002: You're right on both counts, and thanks especially for the second one — just opening the gate would have turned an abort into an `IndexError` on the whole batch.

Fixed in `9c6f5445`. The gate now makes an exception for a request that is on its way out, and `before_decode` uses an explicit branch instead of an unconditional `popleft()`:

```python
@staticmethod
def _is_terminating(req) -> bool:
    return req.to_finish is not None or req.finished()

def is_decode_batch_ready(self, schedule_batch) -> bool:
    return all(
        len(req._omni_data.pending_text_queue) > 0 or self._is_terminating(req)
        for req in schedule_batch.reqs
    )
```

```python
if queue:
    token = queue.popleft()
elif self._is_terminating(req):
    # Carries the forward that retires the request; its codes go nowhere.
    token = self.text_pad_id
else:
    raise RuntimeError(...)
```

The pad frame exists only to carry the forward that applies `to_finish`; the codes it produces are discarded. The `else` is deliberate — an empty queue on a request that is *not* terminating is a scheduling bug, and I would rather see it raised than papered over with a pad frame.

gaoyang07 (approved):

